PR #1619, opened 2023-04-20
Warn if RPC server handlers are not exiting fast enough during stopping time

xemul:
When rpc::server stops it kicks accept abort, stops connections and waits for all of them to resolve relevant stopping futures. One of the stopping promises is that the `_reply_gate` is closed. Inside this gate server runs responses to messages, but also the gate enter-exit spans over the message handler callback to resolve.

While all the kicked-and-waited stuff is internal to RPC, handlers are external and seastar has no way to ensure they finish in a timely manner, it can only assume that. The PR is supposed to facilitate debugging this assumption by counting handler callback entrances and exits (i.e. future resolutions) and warn in server stop doesn't finish within a minute printing the number of currently not-exited handlers.

Stuck server.stop() with zero running handlers is still an internal issue to result.

PR #1671, opened 2023-05-25
Split rpc::server stop into two parts

xemul:
This is to give users the ability to isolate the node from the network without waiting for the internal RPC activity to wrap up. When shutdown()-ed the rpc::server aborts all the connections and accepting socket and returns.

### DIFF
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -616,6 +616,13 @@ public:
     server(protocol_base* proto, server_socket, resource_limits memory_limit = resource_limits(), server_options opts = server_options{});
     server(protocol_base* proto, server_options opts, server_socket, resource_limits memory_limit = resource_limits());
     void accept();
+    /**
+     * Stops the server.
+     *
+     * It makes sure that no new rpcs are admitted, no rpc handlers issued on this
+     * connection are running any longer and no replies on the previously running
+     * handlers will be sent.
+     */
     future<> stop();
     template<typename Func>
     void foreach_connection(Func&& f) {
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -608,6 +608,7 @@ private:
     promise<> _ss_stopped;
     gate _reply_gate;
     server_options _options;
+    bool _shutdown = false;
     uint64_t _next_client_id = 1;
 
 public:
@@ -624,6 +625,17 @@ public:
      * handlers will be sent.
      */
     future<> stop();
+    /**
+     * Shuts down the server.
+     *
+     * Light version of the stop, that just makes sure the server is not visible
+     * by remote clients, i.e. -- no new rpcs are admitted and no replies on the
+     * previously running handlers will be sent. Currently running handlers may
+     * still run.
+     *
+     * Caller of shutdown() mush wait for it to resolve before calling stop.
+     */
+    future<> shutdown();
     template<typename Func>
     void foreach_connection(Func&& f) {
         for (auto c : _conns) {
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -1181,8 +1181,7 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
           _servers.erase(*_options.streaming_domain);
       }
       return _ss_stopped.get_future().then([this] {
-      return
-          parallel_for_each(_conns | boost::adaptors::map_values, [] (shared_ptr<connection> conn) {
+          return parallel_for_each(_conns | boost::adaptors::map_values, [] (shared_ptr<connection> conn) {
               return conn->stop();
           });
       }).finally([this] {
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -1176,12 +1176,14 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
       if (_options.streaming_domain) {
           _servers.erase(*_options.streaming_domain);
       }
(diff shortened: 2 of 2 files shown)


src/rpc/rpc.cc
```diff
@@ -1170,16 +1170,27 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
...
+          return make_ready_future<>();
+      }
+
       _ss.abort_accept();
       _resources_available.broken();
       if (_options.streaming_domain) {
           _servers.erase(*_options.streaming_domain);
       }
-      return when_all(_ss_stopped.get_future(),
+      return
```
gleb-cloudius: But you still wait for connections to top here. From the commit message I was expecting to see `void shutdown()` that only kicks the process and all the waiting happens in `stop()`.

xemul: I thought the same, but then changed my mind. The main goal here is to make `messaging_service::shutdown()` stop doing networking communications, just shutting down all sockets (with a syscall) doesn't look sufficient, it should also wait for packets to stop traveling and that's what connection::stop() does. The server::stop() is to destroy the server completely.

I admit that some parts of connection::stop() are better _not_ to be waited in shutdown. If there are some then we'd need `future<> connection::shutdown()` too, but, again, so far I don't see how it will differ from `connection::stop()`.

Does this make sense to you?

gleb-cloudius: I need to understand what guaranties you need from the `shutdown()`. If you want to make sure that no rpc handlers issued on this connection are running any longer `connection->stop()` is not enough. What it guaranties is that no new rpcs will be admitted (but they are may be some running already) and no replies will be sent to already admitted once. Is this what you want? What is the use case?

xemul: > If you want to make sure that no rpc handlers issued on this connection are running any longer connection->stop() is not enough. 

No, that's not the goal here.

> What it guaranties is that no new rpcs will be admitted (but they are may be some running already) and no replies will be sent to already admitted once. Is this what you want?

Yup, that's what I need.

gleb-cloudius: > 
> > What it guaranties is that no new rpcs will be admitted (but they are may be some running already) and no replies will be sent to already admitted once. Is this what you want?
> 
> Yup, that's what I need.

Then I can confirm it is what you get with this implementation :) But what is the use case may I ask?

xemul: > Then I can confirm it is what you get with this implementation :) 

:champagne: 

> But what is the use case may I ask?

Sure! The use-case is described in scylladb/scylla#14031 . The thing is that `storage_service::do_drain()` eventually calls `messaging_service::shutdown()` which is a bit heavyweight -- it, among other things, waits for all RPC handlers to stop running with the help of `_reply_gate` (I added debug for it in #1619 , you're welcome to review and commit it too).  I want to relax `messaging_service::shutdown()` to only isolate the service from the networking -- i.e. call this PR's `rpc::server::shutdown()` -- and do full stopping later, with `rpc::server::stop()`. It's going to look like this -- https://github.com/xemul/scylla/tree/br-messaging-service-split-drain-and-stop (no PR for this branch yet)

gleb-cloudius: But should isolation also makes sure that no connection are accepted any longer? I think there is a bug in the current implementation that spilled into this patch as well. Before we stop all connections we need to wait for accept fiber to complete otherwise it may add more connections later. Current code waits for accept in parallel with connection closing and after your patch it does it much later.

include/seastar/rpc/rpc.hh
```diff
@@ -616,6 +617,7 @@ public:
     server(protocol_base* proto, server_socket, resource_limits memory_limit = resource_limits(), server_options opts = server_options{});
     server(protocol_base* proto, server_options opts, server_socket, resource_limits memory_limit = resource_limits());
     void accept();
+    future<> shutdown();
     future<> stop();
```
avikivity, line 621: Please add documentation. Yes, rpc is undocumented and the policy is to only update documented interfaces, but stop() and shutdown() are too similar and must be distinguished by an explanation.

src/rpc/rpc.cc
```diff
@@ -1170,16 +1170,27 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
...
               return conn->stop();
-          }),
+          }).finally([this] {
+              _shutdown = true;
+          });
+  }
+
+  future<> server::stop() {
+      return when_all(_ss_stopped.get_future(),
+          shutdown(),
```
avikivity, line 1192: Can we not have two copies of shutdown() running? Or is the user required to wait for shutdown first (sensible)?

avikivity, line 1192: Can we not have two copies of shutdown() running? Or is the user required to wait for shutdown first (sensible)?

xemul: The latter -- user is required to wait for shutdown first. I'll mention it in the doc comment

src/rpc/rpc.cc
xemul: Hm... The accept looks like

```
    _ss.accept().then([] (auto sk) {
        ... // no co_await-s here
        _conns.emplace(sk);
    });
```
so if `accept()` wasn't aborted and resolved there's no way shutdown could start killing connections before this new one gets emplaced into `_conns`. What am I missing?

xemul:
upd:
- documented server::stop() and server::shutdown()
- wait for server socket to stop before killing connections

src/rpc/rpc.cc
xemul: Ah, it can smp::submit_to connections between shards...

gleb-cloudius (approved):

PR #1657, opened 2023-05-17
Improve http::reply re-allocations and copying in client

xemul:
The http::client and http::connection code handle the parsed reply with unnecessary allocations and movements. It's possible to reduce both and simplify the parser a bit.

PR #1672, opened 2023-05-26
HTTP headers in http_response are now case insensitive

kfirgollan:
Previously to this commit only the headers in the http requests were case insensitive. This failed to handled a scenario in which a server returns headers with different cases.
The most impactful case is `content-type` instead of `Content-Type`. In case a server uses lower case, seastar would ignore the content and return an empty reply.

To address this issue the same case-insensitive logic was reused. Move the case_insensitive_cmp/case_insensitive_hash to a utility module.

---------------------------------
As the test suggests, I noticed this issue while working with [uvicorn](https://www.uvicorn.org/), which is a popular web server for python. I noticed that seastar was ignoring the data of the reply. After digging a bit I found that it is caused by the fact that uvicorn uses `content-length` and seastar expected `Content-Length` in a case sensitive way.

Note: According the the HTTP RFC header field names are case insensitive.
[StackOverflow reference](https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive)

---------------------------------
Note: Since I moved `case_insensitive_cmp/case_insensitive_hash` from `request.hh` to the new `string_utils.hh` I took the license comment from the original file.

@xemul - I noticed that you worked on `http` lately, would love a review :)

### DIFF
--- a/include/seastar/util/string_utils.hh
+++ b/include/seastar/util/string_utils.hh
@@ -1,0 +1,68 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright 2015 Cloudius Systems
+ */
+
+//
+// request.hpp
+// ~~~~~~~~~~~
+//
+// Copyright (c) 2003-2013 Christopher M. Kohlhoff (chris at kohlhoff dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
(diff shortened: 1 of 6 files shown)


tests/unit/httpd_test.cc
```diff
@@ -1245,6 +1245,18 @@ SEASTAR_TEST_CASE(case_insensitive_header) {
     return make_ready_future<>();
 }
 
+SEASTAR_TEST_CASE(case_insensitive_header_reply) {
+    seastar::http_response resp{};
```
xemul, line 1249: I'm killing this thing with #1657 . So please, either try to avoid using it, or let's wait for that commit to settle and then rebase

include/seastar/util/string_utils.hh
```diff
@@ -0,0 +1,63 @@
...
+
+#ifndef SEASTAR_MODULE
+#include <cstring>
+#include <stdio.h>
+#endif
+
+#include <seastar/util/modules.hh>
+#include <seastar/core/sstring.hh>
+
+namespace seastar {
```
xemul, line 41: Plz, also wrap it with `namespace internal`. Or put the whole thing into http/common.{cc|hh}

tests/unit/httpd_test.cc
kfirgollan: @xemul - sounds good, you can take care of the case sensitivity issue in your PR as well then?

include/seastar/util/string_utils.hh
kfirgollan: Interesting, why do you think a utility like this should be marked `internal`? any implementation detail of seastar to hide here?
Also adding some guidance on that in `contributing.md` would be great I guess.

xemul: No details to hide. If we add anything to non-internal API it should be
- documented
- maintained in backward compatible manner
The former is not implemented here, the latter is something unlikely to cause any troubles, but still

kfirgollan:
@xemul took a while but applied the fixes eventually, please review

PR #1675, opened 2023-05-29
Add a stall detector histogram

### LINKED ISSUE #1674: create a histogram for stall detection
Not all stalls are the same, it would be helpful to have a histogram of the stalls, we can report a single histogram per node

amnonh:
Seastar reports stalls above some threshold. It is helpful to follow the  distribution of the stalls. 

This series adds a time_estimated_histogram to the stall detector.
The series takes the estimated_histogram definition from the ScyllaDB project.
    
Whenever a stall is found, it is added to the histogram.
The stall histogram is reported once per node.
    
Fixes #1674

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -959,6 +959,7 @@ reactor::task_queue::set_shares(float shares) noexcept {
 void
 reactor::account_runtime(task_queue& tq, sched_clock::duration runtime) {
     if (runtime > (2 * _task_quota)) {
+        _stalls_histogram.add(runtime);
         tq._time_spent_on_task_quota_violations += runtime - _task_quota;
     }
     tq._vruntime += tq.to_vruntime(runtime);
@@ -2527,6 +2528,7 @@ void reactor::register_metrics() {
             sm::make_total_bytes("aio_bytes_write", _io_stats.aio_write_bytes, sm::description("Total aio-writes bytes")),
             sm::make_counter("aio_outsizes", _io_stats.aio_outsizes, sm::description("Total number of aio operations that exceed IO limit")),
             sm::make_counter("aio_errors", _io_stats.aio_errors, sm::description("Total aio errors")),
+            sm::make_histogram("stalls", sm::description("A histogram of reactor stall durations"), [this] {return _stalls_histogram.to_metrics_histogram();}).aggregate({seastar::metrics::shard_label}).set_skip_when_empty(),
             // total_operations value:DERIVE:0:U
             sm::make_counter("fsyncs", _fsyncs, sm::description("Total number of fsync operations")),
             // total_operations value:DERIVE:0:U
--- a/tests/unit/metrics_test.cc
+++ b/tests/unit/metrics_test.cc
@@ -31,6 +31,7 @@
 #include <seastar/core/do_with.hh>
 #include <seastar/core/io_queue.hh>
 #include <seastar/core/loop.hh>
+#include <seastar/core/internal/estimated_histogram.hh>
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
 #include <seastar/testing/test_runner.hh>
@@ -318,4 +319,62 @@ SEASTAR_THREAD_TEST_CASE(test_relabel_enable_disable_skip_when_empty) {
         return mi.should_skip_when_empty == sm::skip_when_empty::yes;
     }), 0);
     sm::set_relabel_configs({}).get();
+}
+
+SEASTAR_THREAD_TEST_CASE(test_estimated_histogram) {
+    using namespace seastar::metrics;
+    using namespace std::chrono_literals;
+    internal::time_estimated_histogram histogram1;
+    internal::time_estimated_histogram histogram2;
+    // The number of linearly-spaced buckets between consecutive powers of 2 in time_estimated_histogram.
+    constexpr int PRECISION = 4;
+
+    // The lower bound of time_estimated_histogram is 512 us.
+    std::chrono::steady_clock::duration min = std::chrono::microseconds(512);
+    std::chrono::steady_clock::duration next = min*2;
+
+    for (size_t i = 0; i < 16; i++) {
+        auto delta = (next - min)/PRECISION;
+        for (size_t j = 0; j< PRECISION; j++) {
+            histogram1.add(min + delta*j);
+        }
+        min = next;
+        next *= 2;
+    }
+    BOOST_CHECK_EQUAL(histogram1.count(), 64);
+    for (size_t i = 0; i < 64; i++) {
+        BOOST_CHECK_EQUAL(histogram1.get(i), 1);
+    }
+    min = std::chrono::microseconds(512);
+    next = min*2;
+    for (size_t i = 0; i < 8; i++) {
+        auto delta = (next - min)/PRECISION;
+        for (size_t j = 0; j< PRECISION; j++) {
+            histogram2.add(min + delta*j);
+        }
+        min = next;
+        next *= 2;
+    }
+    BOOST_CHECK_EQUAL(histogram2.count(), 32);
+    for (size_t i = 0; i < 32; i++) {
+        BOOST_CHECK_EQUAL(histogram2.get(i), 1);
+    }
+    for (size_t i = 33; i < 64; i++) {
+        BOOST_CHECK_EQUAL(histogram2.get(i), 0);
+    }
+    histogram1.merge(histogram2);
+    BOOST_CHECK_EQUAL(histogram1.count(), 96);
+    for (size_t i = 0; i < 32; i++) {
+        BOOST_CHECK_EQUAL(histogram1.get(i), 2);
+    }
+    for (size_t i = 33; i < 64; i++) {
+        BOOST_CHECK_EQUAL(histogram1.get(i), 1);
+    }
+    auto mh = histogram1.to_metrics_histogram();
+    for (size_t i = 0; i < 32; i++) {
+        BOOST_CHECK_EQUAL(mh.buckets[i].count, 2 + i*2);
+    }
+    for (size_t i = 33; i < 64; i++) {
+        BOOST_CHECK_EQUAL(mh.buckets[i].count, 33 + i);
+    }
 }
--- a/include/seastar/core/internal/estimated_histogram.hh
+++ b/include/seastar/core/internal/estimated_histogram.hh
@@ -1,0 +1,381 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2023 ScyllaDB.
+ */
+
+#pragma once
+
+#include <cmath>
+#include <algorithm>
+#include <vector>
+#include <chrono>
+#include <seastar/core/metrics_types.hh>
+#include <seastar/core/print.hh>
+#include <seastar/core/bitops.hh>
+#include <limits>
+#include <array>
+
+namespace seastar::metrics::internal {
+/**
+ * This is a pseudo-exponential implementation of an estimated histogram.
+ *
+ * An exponential-histogram with coefficient 'coef', is a histogram where for bucket 'i'
+ * the lower limit is coef^i and the higher limit is coef^(i+1).
+ *
+ * A pseudo-exponential is similar but the bucket limits are an approximation.
+ *
+ * The approximate_exponential_histogram is an efficient pseudo-exponential implementation.
+ *
+ * The histogram is defined by a Min and Max value limits, and a Precision (all should be power of 2
+ * and will be explained).
+ *
+ * When adding a value to a histogram:
+ * All values lower than Min will be included in the first bucket (the assumption is that it's
+ * not suppose to happen but it is ok if it does).
+ *
+ * All values higher than Max will be included in the last bucket that serves as the
+ * infinity bucket (the assumption is that it can happen but it is rare).
+ *
+ * Note the difference between the first and last buckets.
+ * The first bucket is just like a regular bucket but has a second roll to collect unexpected low values.
+ * The last bucket, also known as the infinity bucket, collect all values that passes the defined Max,
+ * it only collect those values.
+ *
+ * Buckets Distribution (limits)
+ * =============================
+ * The buckets limits in the histogram are defined similar to a floating-point representation.
+ *
+ * Buckets limits have an exponent part and a linear part.
+ *
+ * The exponential part is a power of 2. Each power-of-2 range [2^n..2^n+1)
+ * is split linearly to 'Precision' number of buckets.
+ *
+ * The total number of buckets is:
+ *   NUM_BUCKETS = log2(Max/Min)*Precision +1
+ *
+ * For example, if the Min value is 128, the Max is 1024 and the Precision is 4, the number of buckets is 13.
+ *
+ * Anything below 160 will be in the bucket 0, anything above 1024 will be in bucket 13.
+ * Note that the first bucket will include all values below Min.
+ *
+ * the range [128, 1024) will be split into log2(1024/128) = 3 ranges:
+ * 128, 256, 512, 1024
+ * Or more mathematically: [128, 256), [256, 512), [512,1024)
+ *
+ * Each range is split into 4 (The Precision).
+ * 128            | 256            | 512            | 1024
+ * 128 160 192 224| 256 320 384 448| 512 640 768 896|
+ *
+ * To get the exponential part of an index you divide by the Precision.
+ * The linear part of the index is Modulus the precision.
+ *
+ * Calculating the bucket lower limit of bucket i:
+ * The exponential part: exp_part = 2^floor(i/Precision)* Min
+ *    with the above example 2^floor(i/4)*128
+ * The linear part: (i%Precision) * (exp_part/Precision)
+ *    With the example: (i%4) * (exp_part/4)
+ *
+ * So the lower limit of bucket 6:
+ *    2^floor(6/4)* 128  = 256
+ *    (6%4) * 256/4 = 128
+ *    lower-limit   = 384
+ *
+ * How to find a bucket index for a value
+ * =======================================
+ * The bucket index consist of two parts:
+ * higher bits (exponential part) are based on log2(value/min)
+ *
+ * lower bits (linear part) are based on the 'n' MSB (ignoring the leading 1) where n=log2(Precision).
+ * Continuing with the example where the number of precision bits: PRECISION_BITS = log2(4) = 2
+ *
+ * for example: 330 (101001010)
+ * The number of precision_bits: PRECISION_BITS = log2(4) = 2
+ * higher bits: log2(330/128) = 1
+ * MSB: 01 (the highest two bits following the leading 1)
+ * So the index: 101 = 5
+ *
+ * About the Min, Max and Precision
+ * ================================
+ * For Min, Max and Precision, choose numbers that are a power of 2.
+ *
+ * Limitation: You must set the MIN value to be higher or equal to the Precision.
+ *
+ * \param Min - The lowest expected value. A value below that will be set to the Min value.
+ * \param Max - The highest number to consider. A value above that is considered infinite.
+ * \param Precision - The number of buckets in each range (where a range is the i^2..i^(2+1) )
+ */
+template<uint64_t Min, uint64_t Max, size_t Precision>
+class approximate_exponential_histogram {
+    static constexpr unsigned NUM_EXP_RANGES = log2floor(Max/Min);
+    static constexpr size_t NUM_BUCKETS = NUM_EXP_RANGES * Precision + 1;
+    static constexpr unsigned PRECISION_BITS = log2floor(Precision);
+    static constexpr unsigned BASESHIFT = log2floor(Min);
+    static constexpr uint64_t LOWER_BITS_MASK = Precision - 1;
+    std::array<uint64_t, NUM_BUCKETS> _buckets;
+public:
+    approximate_exponential_histogram() {
+        clear();
+    }
+
+    /*!
+     * \brief Returns the bucket lower limit given the bucket id.
+     * The first and last bucket will always return the MIN and MAX respectively.
+     *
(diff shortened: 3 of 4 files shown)


src/core/reactor.cc
```diff
@@ -1111,7 +1111,9 @@ cpu_stall_detector::cpu_stall_detector(cpu_stall_detector_config cfg)
     namespace sm = seastar::metrics;
 
     _metrics.add_group("stall_detector", {
-            sm::make_counter("reported", _total_reported, sm::description("Total number of reported stalls, look in the traces for the exact reason"))});
+            sm::make_counter("reported", _total_reported, sm::description("Total number of reported stalls, look in the traces for the exact reason")),
+            sm::make_histogram("stalls", sm::description("A histogram stalls"), [this] {return _stalls_histogram.to_metrics_histogram();}).aggregate({seastar::metrics::shard_label}).set_skip_when_empty()
+    });
```
raphaelsc, line 1116: I think we can easily cache seastar::metrics::histogram, and only recompute it once the histogram is updated.

raphaelsc: Updating the histogram will be rare, as it only happens once we have a stall in a shard, therefore, computing the metric histogram will be cheap most of the time.

amnonh: While this is true, it's an unneeded optimization. A calculation is done only when scrapping a metric, typically once every 20-30 seconds.

Reducing that calculation is neglectable.

raphaelsc: A large number of small incidents can add up to a disaster.

Users won't be happy with spikes every 20-30 seconds.

Not claiming this specific metric is the root of all evil, but I think the rule # 1 should be make every metric computation as fast as possible. Expose some metric precomputed. E.g. per-table disk usage. Will I compute the disk usage by iterating through all sstables when scraping the metric, or I will take advantage that I can precompute it once updating the sstable set? The latter.

to_metrics_histogram() allocates some memory, iterates through each bucket, and aggregate some stats from each.

amnonh: but we have one such metric per shard

amnonh: @raphaelsc not sure if you missed something or if  missed something.
once every scrap interval (lets say 20s) the metric layer collect the value.
it takes a copy of the data (it's preamptable so things can change while reporting).
We don't save much if we hold that copy (we're talking about 60*float of data).

We have many more histograms in the system, so saving this one will have zero impact.
In the future I'll be moving to Prometheus native histograms, that's were more optimization would have effect.

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+ * So the index: 101 = 5
+ *
+ * About the Min, Max and Precision
+ * ================================
+ * For Min, Max and Precision, choose numbers that are a power of 2.
+ *
+ * Limitation: You must set the MIN value to be higher or equal to the Precision.
+ *
+ */
+template<uint64_t Min, uint64_t Max, size_t Precision>
```
avikivity, line 121: Please add \tparam annotations to explain the parameters directly.

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+class approx_exponential_histogram {
```
avikivity, line 122: s/approx/approximate/

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+public:
+
+    static constexpr unsigned NUM_EXP_RANGES = log2floor(Max/Min);
+    static constexpr size_t NUM_BUCKETS = NUM_EXP_RANGES * Precision + 1;
+    static constexpr unsigned PRECISION_BITS = log2floor(Precision);
+    static constexpr unsigned BASESHIFT = log2floor(Min);
+    static constexpr uint64_t LOWER_BITS_MASK = Precision - 1;
```
avikivity, line 129: Can these be made private?

include/seastar/core/internal/stall_detector.hh
```diff
@@ -59,6 +60,7 @@ protected:
     std::atomic<uint64_t> _stall_detector_missed_ticks = { 0 };
     unsigned _reported = 0;
     unsigned _total_reported = 0;
+    metrics::time_estimated_histogram _stalls_histogram;
```
avikivity, line 63: You're not using approx_blah_historgram here.

amnonh: @avikivity right, but time_estimated_histogram uses it.

I've took it as-is from scyllaDB repository, in the future we can decide to keep only this copy

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+#include <vector>
+#include <chrono>
+#include <seastar/core/metrics_types.hh>
+#include <seastar/core/print.hh>
+#include <seastar/core/bitops.hh>
+#include <limits>
+#include <array>
+
+namespace seastar {
+namespace metrics {
```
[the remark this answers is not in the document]
amnonh: it's public for users of the metrics layer

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+class approx_exponential_histogram {
+public:
+
+    static constexpr unsigned NUM_EXP_RANGES = log2floor(Max/Min);
+    static constexpr size_t NUM_BUCKETS = NUM_EXP_RANGES * Precision + 1;
+    static constexpr unsigned PRECISION_BITS = log2floor(Precision);
+    static constexpr unsigned BASESHIFT = log2floor(Min);
+    static constexpr uint64_t LOWER_BITS_MASK = Precision - 1;
+private:
+    std::array<uint64_t, NUM_BUCKETS> _buckets;
```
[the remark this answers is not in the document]
amnonh: I didn't understnad this comment

amnonh:
@avikivity V2 had the following changes:
const expr are now private
add \param for the the template parameters
add a unit test for the histogram

include/seastar/core/estimated_histogram.hh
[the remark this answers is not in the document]
tchaikov: nit, could instead initialize `_buckets` using aggregation initialization, like:
```c++
std::array<uint64_t, NUM_BUCKETS> _buckets = {};
```
instead of calling `clear()` in the constructor. IMHO, it'd be more clear that to initialize member variable where they are defined.

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+     * It will return 0 if the histogram is empty.
+     */
+    uint64_t mean() const {
+        uint64_t elements = 0;
+        double sum = 0;
+        for (size_t i = 0; i < NUM_BUCKETS - 1; i++) {
+            elements += _buckets[i];
+            sum += _buckets[i] * get_bucket_lower_limit(i);
+        }
+        return (elements) ?  sum / elements : 0;
```
tchaikov, line 295: nit, might want to drop the parenthesis  around `elements` .

include/seastar/core/estimated_histogram.hh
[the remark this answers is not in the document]
tchaikov: we can use `namespace seastar::metrics { ... }` since C++17, IMHO, it's more readable than the nested notation. see syntax (8) in https://en.cppreference.com/w/cpp/language/namespace.

amnonh:
> @avikivity V2 had the following changes: const expr are now private add \param for the the template parameters add a unit test for the histogram

@avikivity can you take a look at V2?

src/core/reactor.cc
```diff
@@ -1401,6 +1403,7 @@ void cpu_stall_detector::generate_trace() {
     auto delta = reactor::now() - _run_started_at;
 
     _total_reported++;
+    _stalls_histogram.add(delta);
```
michoecho, line 1406: This is the wrong place to put the histogram increment.

 `generate_trace()` is rate limited. We have to ratelimit the logs out of necessity, but it would be undesirable to ratelimit histogram increments.

Also, `cpu_stall_detector` reports every long stall multiple times (as its millisecond duration crosses consecutive power of 2), which would render the histogram confusing at least.

I strongly suggest moving this line to the `if` inside `reactor::account_runtime`:
```
void
reactor::account_runtime(task_queue& tq, sched_clock::duration runtime) {
    if (runtime > (2 * _task_quota)) {
        tq._time_spent_on_task_quota_violations += runtime - _task_quota;
    }
    tq._vruntime += tq.to_vruntime(runtime);
    tq._runtime += runtime;
}
```

This way we both avoid the two problems I mentioned above (rate limiting and accounting the same stall multiple times), and we make the histogram count all stalls (as defined by `task_quota_ms`, not just the *logged* "stalls" as defined by `blocked-reactor-notify-ms`, which is much more useful.

michoecho (changes requested):
The PR looks good to me, and it implements a desirable feature.

But I have a change request, because I think you picked the wrong place to increment the histogram. You increment it in `generate_trace()`, which is called when stalls are *logged*. But not all stalls are logged, and some stalls are logged multiple times. It would be much better to move this histogram increment to the source — the reactor — and account all stalls, not only the logged ones.

src/core/reactor.cc
michoecho: @raphaelsc Your line of thinking is dangerous, because it leads to a positive feedback loop of performance problems.

You are basically saying: "running this performance diagnostic code might cause performance problems, so let's run it only when performance problems happen". That's the kind of thing that invites a disaster. If there really is a performance problem in diagnostic code, it's better if it happens every 20s (so we can catch it with any performance test and react), rather than have it happen conditionally during other performance problems and amplify them.

That aside, as @amnonh says, outputting a duration histogram to wire every 20s should be nowhere near a performance problem. (And if it is a performance problem, the entire metrics framework deserves to be burned).

raphaelsc: > @raphaelsc Your line of thinking is dangerous, because it leads to a positive feedback loop of performance problems.
> 
> You are basically saying: "running this performance diagnostic code might cause performance problems, so let's run it only when performance problems happen". _That's_ the kind of thing that invites a disaster. If there really is a performance problem in diagnostic code, it's better if it happens unconditionally every 20s (so we can catch it with any performance test and react), rather than have it happen conditionally during other performance problems and amplify them.
> 

Not at all. You're thinking I was suggest hiding the dirt under the carpet, a performance issue that comes from to_metrics_histogram() for example, but that's not what I meant. I was only suggesting that as a rule of thumb, we want a metric implementation to be as fast as possible. Otherwise, we'll be unnecessarily burning cpu cycles which could be devouted elsewhere (like exposing more metrics during the scraping). For example, if I want to expose some metric from sstable set (e.g. compaction backlog), will I iterate through sstables and perform the calculation or will I take advantage of the fact that the metric can be precomputed when updating the set? Exposing a metric is not free, and there's a real overhead that can affect the system. There's a reason why we have real issues with monitoring when enabling per-table metrics.

amnonh: @raphaelsc the problem we have with per-table metrics is because we have histograms per table per shard, that's alot, this is one histogram per node. one.
It's neglectable. There's a future work of making histogram reporting faster, but that would have impact one you're talking about many histogram, for a single histogram it will have no effect

michoecho: > I was only suggesting that as a rule of thumb, we want a metric implementation to be as fast as possible.

@raphaelsc But if I understand correctly, you aren't suggesting to speed it up — you are suggesting to cache it. This will save you some CPU cycles, yes, but only in the case when there are no stalls during the scrape period, which is a condition that we don't control.

This is an opportunistic optimization — one that speeds the system up, but only in lucky/happy conditions.
In a latency-sensitive system, such optimizations are risky in general because they make the behavior of the system harder to predict. But they are extra risky if their conditions depend on performance, because that causes a feedback loop.

In this particular case, we could claim the following: "If there are frequent stalls, the proposed change has no effect. If there are no stalls, the proposed change speeds things up. So the change is a clear improvement".
But the dual of that is: "After the change, when the system has performance problems (which is likely to cause stalls), producing the metric is more expensive than usual, so the system has even more performance problems".

Another example of this trope in Scylla is the row cache. It doesn't have a fixed size, but balloons to fill all unused memory in the system. In the usual case this makes the cache bigger, which makes requests cheaper, which means that the peak throughput of the system is greater.
But the dual of that is: if the system gets overloaded, requests and maintenance work might start to queue up, which will steal memory from the cache, which will make the cache smaller, which will make requests more expensive to serve. And in the end, this design causes the system to respond to overload by *slowing down*.

Do you see what I mean? The work saved by opportunistic optimizations is not truly gone — it's hidden under the carpet, ready to come back when that would be the most problematic.

raphaelsc: > > I was only suggesting that as a rule of thumb, we want a metric implementation to be as fast as possible.
> 
> @raphaelsc But if I understand correctly, you aren't suggesting to speed it up — you are suggesting to cache it. This will save you some CPU cycles, yes, but only in the case when there are no stalls during the scrape period, which is a condition that we don't control.
> 
> This is an opportunistic optimization — one that speeds the system up, but only in lucky/happy conditions. In a latency-sensitive system, such optimizations are risky in general because they make the behavior of the system harder to predict. But they are extra risky if their conditions depend on performance, because that causes a feedback loop.
> 
> In this particular case, we could claim the following: "If there are frequent stalls, the proposed change has no effect. If there are no stalls, the proposed change speeds things up. So the change is a clear improvement". But the dual of that is: "After the change, when the system has performance problems (which is likely to cause stalls), producing the metric is more expensive than usual, so the system has even more performance problems".
> 
> Another example of this trope in Scylla is the row cache. It doesn't have a fixed size, but balloons to fill all unused memory in the system. In the usual case this makes the cache bigger, which makes requests cheaper, which means that the peak throughput of the system is greater. But the dual of that is: if the system gets overloaded, requests and maintenance work might start to queue up, which will steal memory from the cache, which will make the cache smaller, which will make requests more expensive to serve. And in the end, this design causes the system to respond to overload by _slowing down_.
> 
> Do you see what I mean? The work saved by opportunistic optimizations is not truly gone — it's hidden under the carpet, ready to come back when that would be the most problematic.

What you wrote above makes total sense, and I agree completely with you.

src/core/reactor.cc
amnonh: @michoecho thanks for the suggestion, I just forced push v3 with the changes

include/seastar/core/estimated_histogram.hh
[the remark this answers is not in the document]
avikivity: I know it's public, but should it be? Making it public means we can't change it later.

include/seastar/core/estimated_histogram.hh
avikivity: Seastar naming convention is to use full names. It shouldn't be affected by the donor repository.

If ScyllaDB wants to use it, it can use a type synonym, or update the mentions.

include/seastar/core/estimated_histogram.hh
avikivity: ?

include/seastar/core/estimated_histogram.hh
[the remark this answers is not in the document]
avikivity: If you use std::array, you should include \<array> and not rely on it getting here somehow, because some day it won't.

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+
+    void add(const duration& latency) {
+        add_micro(std::chrono::duration_cast<std::chrono::microseconds>(latency).count());
+    }
+};
+
+inline time_estimated_histogram time_estimated_histogram_merge(time_estimated_histogram a, const time_estimated_histogram& b) {
+    return a.merge(b);
+}
+
```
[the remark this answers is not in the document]
avikivity: ?

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+/*!
+ * \brief estimated histogram for duration values
+ * time_estimated_histogram is used for short task timing.
+ * It covers the range of 0.5ms to 33s with a precision of 4.
+ *
+ * 512us, 640us, 768us, 896us, 1024us, 1280us, 1536us, 1792us...16s, 20s, 25s, 29s, 33s (33554432us)
+ */
+class time_estimated_histogram : public approx_exponential_histogram<512, 33554432, 4> {
+public:
```
avikivity, line 363: Why is time_estimated_histogram choosing these values? Different measurements will need different parameters. Best to keep this a template.

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+    using clock = std::chrono::steady_clock;
+    using duration = clock::duration;
+    time_estimated_histogram& merge(const time_estimated_histogram& b) {
+        approx_exponential_histogram<512, 33554432, 4>::merge(b);
+        return *this;
+    }
+
+    void add_micro(uint64_t n) {
+        approx_exponential_histogram<512, 33554432, 4>::add(n);
```
avikivity, line 372:
```
template <...>
add(std::chrono::duration<>
```

include/seastar/core/estimated_histogram.hh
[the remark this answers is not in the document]
amnonh: it's included

include/seastar/core/estimated_histogram.hh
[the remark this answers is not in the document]
amnonh: I've added unit test

include/seastar/core/estimated_histogram.hh
amnonh: It's taken from Scylla As-Is, I took it with as minimal changes as possible, I can link to the discussion when it was added back then.

When I'll do more changes to the histograms mechanism, I'll update it here, but i think we should be consisten between the implementations.

michoecho: @avikivity Are you aware that all histogram code in this PR was copied as-is from the Scylla repository? Focusing the review on years-old code, which @amnonh isn't an author of, seems counterproductive.

include/seastar/core/estimated_histogram.hh
amnonh: They are now private

include/seastar/core/estimated_histogram.hh
amnonh: I can template the add(duration)

tests/unit/metrics_test.cc
```diff
@@ -319,3 +320,57 @@ SEASTAR_THREAD_TEST_CASE(test_relabel_enable_disable_skip_when_empty) {
     }), 0);
     sm::set_relabel_configs({}).get();
 }
+
+SEASTAR_THREAD_TEST_CASE(test_estimated_histogram) {
+    using namespace seastar::metrics;
+    using namespace std::chrono_literals;
+    time_estimated_histogram histogram1;
+    time_estimated_histogram histogram2;
+    time_estimated_histogram::duration min = std::chrono::microseconds(512);
```
michoecho, line 329: nit: magic number. It should have a comment, like:
```
// The lower bound of time_estimated_histogram is 512 us.
time_estimated_histogram::duration min = std::chrono::microseconds(512);
```

tests/unit/metrics_test.cc
```diff
@@ -319,3 +320,57 @@ SEASTAR_THREAD_TEST_CASE(test_relabel_enable_disable_skip_when_empty) {
...
+    time_estimated_histogram::duration next = min*2;
+
+    for (size_t i = 0; i < 16; i++) {
+        auto delta = (next - min)/4;
```
michoecho, line 333: nit: magic number. It would be better to have something like:
```
// The number of linearly-spaced buckets between consecutive powers of 2 in time_estimated_histogram.
constexpr int PRECISION = 4;
...
auto delta = (next - min) / PRECISION;
```

src/core/reactor.cc
```diff
@@ -2519,6 +2520,7 @@ void reactor::register_metrics() {
             sm::make_total_bytes("aio_bytes_write", _io_stats.aio_write_bytes, sm::description("Total aio-writes bytes")),
             sm::make_counter("aio_outsizes", _io_stats.aio_outsizes, sm::description("Total number of aio operations that exceed IO limit")),
             sm::make_counter("aio_errors", _io_stats.aio_errors, sm::description("Total aio errors")),
+            sm::make_histogram("stalls", sm::description("A histogram of reactor stalls duration"), [this] {return _stalls_histogram.to_metrics_histogram();}).aggregate({seastar::metrics::shard_label}).set_skip_when_empty(),
```
michoecho, line 2523: typo: reactor stalls duration -> reactor stall durations

michoecho (changes requested):
Looks good to me, but I think we should make the histograms private/internal (by moving them to different folders and or to the `internal` namespace), because we don't want them used by users of Seastar.

I also added a few very minor nitpicks.

I didn't review the histogram code, because I think that's out of scope here.

tests/unit/metrics_test.cc
amnonh: It's a test. Those are arbitrary numbers. I chose the same numbers I used for time_estimated_histogram, but it could have been anything (as long as it's  power of 2)

amnonh:
> Looks good to me, but I think we should make the histograms private/internal (by moving them to different folders and or to the `internal` namespace), because we don't want them used by users of Seastar.
> 

Yes, following @avikivity comment I've moved them into impl name space, compiling and testing it now

tests/unit/metrics_test.cc
michoecho: > Those are arbitrary numbers. I chose the same numbers I used for time_estimated_histogram, but it could have been anything (as long as it's power of 2)

What do you mean? If you changed the 512 here to any other number, the test would fail. Likewise, if any constant in time_estimated_histogram changed, the test would fail.

amnonh:
@avikivity @michoecho 
I've forced push V4
rename to approximate_exponential_histogram
histograms are now under impl namespace.
I've used a template member function for adding a duration
In the unit test, I've replaced numbers with consexpr
fixed the typo in the metric definition.

michoecho (approved):

amnonh:
@avikivity ping

include/seastar/core/estimated_histogram.hh
```diff
@@ -0,0 +1,385 @@
...
+#include <chrono>
+#include <seastar/core/metrics_types.hh>
+#include <seastar/core/print.hh>
+#include <seastar/core/bitops.hh>
+#include <limits>
+#include <array>
+
+namespace seastar {
+namespace metrics {
+namespace impl {
```
avikivity, line 36: Seastar uses `internal` for internal stuff.

You can write `namespace seastar::metrics::internal {` instead of nesting three statements.

Please move the file to core/internal so it's not used by users.

include/seastar/core/estimated_histogram.hh
avikivity: Taking 7 year old mistakes and enshrining them as an API is not a way to bypass review.

We bypassed review in another way, by making things internal, but copy/paste in general is not an excuse for bad code.

amnonh:
@avikivity I've moved to internal and pull rebased and forced push an updated version

PR #1676, opened 2023-05-29
gate: track holders

bhalevy:
Keep a (boost intrusive) list of holders in debug mode
so in case `gate.close()` doesn't finish in timely manner on shutdown, we can follow this list and track the continuation chain that holding the `gate::holder` and blocking `close()`

Closes #1676

### DIFF
--- a/include/seastar/core/gate.hh
+++ b/include/seastar/core/gate.hh
@@ -305,8 +305,8 @@ namespace internal {
 template <typename Func>
 inline
 auto
-invoke_func_with_gate(gate& g, Func&& func) noexcept {
-    return futurize_invoke(std::forward<Func>(func)).finally([&g] { g.leave(); });
+invoke_func_with_gate(gate::holder&& gh, Func&& func) noexcept {
+    return futurize_invoke(std::forward<Func>(func)).finally([gh = std::forward<gate::holder>(gh)] {});
 }
 
 } // namespace internal
@@ -324,8 +324,7 @@ template <typename Func>
 inline
 auto
 with_gate(gate& g, Func&& func) {
-    g.enter();
-    return internal::invoke_func_with_gate(g, std::forward<Func>(func));
+    return internal::invoke_func_with_gate(g.hold(), std::forward<Func>(func));
 }
 
 /// Executes the function \c func if the gate \c g can be entered
@@ -344,11 +343,11 @@ template <typename Func>
 inline
 auto
 try_with_gate(gate& g, Func&& func) noexcept {
-    if (!g.try_enter()) {
+    if (g.is_closed()) {
         using futurator = futurize<std::invoke_result_t<Func>>;
         return futurator::make_exception_future(gate_closed_exception());
     }
-    return internal::invoke_func_with_gate(g, std::forward<Func>(func));
+    return internal::invoke_func_with_gate(g.hold(), std::forward<Func>(func));
 }
 /// @}
 
--- a/include/seastar/core/gate.hh
+++ b/include/seastar/core/gate.hh
@@ -22,6 +22,8 @@
 #pragma once
 
 #ifndef SEASTAR_MODULE
+#include <boost/intrusive/list.hpp>
+
 #include <seastar/core/future.hh>
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/modules.hh>
@@ -61,21 +63,16 @@ class gate {
     std::optional<promise<>> _stopped;
 
 #ifdef SEASTAR_GATE_HOLDER_DEBUG
-    size_t _holders = 0;
-
-    void assert_not_held_when_moved() const noexcept {
-        assert(!_holders && "gate moved with outstanding holders");
-    }
-    void assert_not_held_when_destroyed() const noexcept {
-        assert(!_holders && "gate destroyed with outstanding holders");
-    }
+    void assert_not_held_when_moved() const noexcept;
+    void assert_not_held_when_destroyed() const noexcept;
 #else   // SEASTAR_GATE_HOLDER_DEBUG
     void assert_not_held_when_moved() const noexcept {}
     void assert_not_held_when_destroyed() const noexcept {}
 #endif  // SEASTAR_GATE_HOLDER_DEBUG
 
 public:
-    gate() = default;
+    // Implemented to force noexcept due to boost:intrusive::list
+    gate() noexcept {};
     gate(const gate&) = delete;
     gate(gate&& x) noexcept
         : _count(std::exchange(x._count, 0)), _stopped(std::exchange(x._stopped, std::nullopt)) {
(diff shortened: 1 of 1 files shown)


include/seastar/core/gate.hh
```diff
@@ -61,21 +63,16 @@ class gate {
...
 #ifdef SEASTAR_GATE_HOLDER_DEBUG
-    size_t _holders = 0;
-
-    void assert_not_held_when_moved() const noexcept {
-        assert(!_holders && "gate moved with outstanding holders");
-    }
-    void assert_not_held_when_destroyed() const noexcept {
-        assert(!_holders && "gate destroyed with outstanding holders");
-    }
+        void assert_not_held_when_moved() const noexcept;
```
xemul, line 66: Why did indentation shifted one level right?

include/seastar/core/gate.hh
```diff
@@ -61,21 +63,16 @@ class gate {
...
+        void assert_not_held_when_destroyed() const noexcept;
 #else   // SEASTAR_GATE_HOLDER_DEBUG
-    void assert_not_held_when_moved() const noexcept {}
-    void assert_not_held_when_destroyed() const noexcept {}
+        void assert_not_held_when_moved() const noexcept {}
```
xemul, line 69: And here too

include/seastar/core/gate.hh
```diff
@@ -201,7 +214,9 @@ public:
         /// Construct a \ref holder by moving another \c holder.
         /// The referenced \ref gate is unaffected, and so the
         /// move-constructor must never throw.
-        holder(holder&& x) noexcept : _g(std::exchange(x._g, nullptr)) { }
+        holder(holder&& x) noexcept : _g(std::exchange(x._g, nullptr)) {
+            debug_hold_gate();
```
xemul, line 218: The `x` should be unlinked from the list. Probably with the help of container_type::node_algorithms::replace_node?

include/seastar/core/gate.hh
```diff
@@ -246,29 +262,35 @@ public:
...
     /// the gate when constructed and \ref leave "leave()s" it when destroyed.
     holder hold() {
         return holder(*this);
     }
+
+private:
+#ifdef SEASTAR_GATE_HOLDER_DEBUG
+    using holders_list_t = boost::intrusive::list<holder,
+        boost::intrusive::member_hook<holder, holder::member_hook_t,
+        &holder::_link>,
```
xemul, line 277: nit: this `&holder::_link>,` should be on previous line

include/seastar/core/gate.hh
```diff
@@ -315,8 +315,7 @@ template <typename Func>
 inline
 auto
 with_gate(gate& g, Func&& func) {
```
xemul, line 317: Maybe deprecate the with_gate() altogether? Holder provide better way of not-forgetting-to-leave the gate

include/seastar/core/gate.hh
bhalevy: not sure. Will fix.

include/seastar/core/gate.hh
bhalevy: The boost intrusive hook has `auto_unlink` mode so x would get unlinked when destroyed.

include/seastar/core/gate.hh
bhalevy: why? having each param on it's own line is quite clear.
maybe the sin here is using single indent rather than double :)

include/seastar/core/gate.hh
bhalevy: We have similar with_* functions like with_closeable, with_lock, etc. that I like for providing scoped and protected access while using some resource, and this works great for continuations.
The holder concept that is more open ended and serves RAII usage in coroutines, which aren't even enabled by default in seastar.
I think that when we enable coroutines by default it'd be a good time to consider deprecating the with_* wrappers.

include/seastar/core/gate.hh
xemul: Well, yes, so it should be either

```
    using holders_list_t = boost::intrusive::list<holder,
        boost::intrusive::member_hook<holder, holder::member_hook_t, &holder::_link>,
        boost::intrusive::constant_time_size<false>>;
```

or

```
    using holders_list_t = boost::intrusive::list<holder,
        boost::intrusive::member_hook<holder,
            holder::member_hook_t,
            &holder::_link>,
        boost::intrusive::constant_time_size<false>>;

```

bhalevy: ah, got your point.
clang-format to the rescue! :)

include/seastar/core/gate.hh
xemul: Yes, technically you're perfectly correct, but then `x` happens in a "state" when it's linked, but `x._g == nullptr` which is weird (and probably puts a landmine for future)

include/seastar/core/gate.hh
bhalevy: fixed in v2

include/seastar/core/gate.hh
bhalevy: fixed in v2

include/seastar/core/gate.hh
bhalevy: fixed in v2 by adding a private release_gate function that extracts the gate member and if not null, calls `debug_release_gate()` that calls boost `unlink()`.

include/seastar/core/gate.hh
bhalevy: fixed in v2

dotnwat:
This is cool.

> so in case gate.close() doesn't finish in timely manner on shutdown, we can follow this list and track the continuation

There is the related problem too of being able to figure out which gate is holding up the shutdown. Did you consider maybe putting each gate itself on a per-reactor list so that on shutdown there could be an entry point for finding stuck gates?

avikivity:
We have the tasktrace thing.

include/seastar/core/gate.hh
```diff
@@ -233,42 +258,46 @@ public:
         holder& operator=(holder&& x) noexcept {
             if (&x != this) {
                 release();
-                _g = std::exchange(x._g, nullptr);
+                _g = std::move(x).release_gate();
```
avikivity, line 261: Why std::move() here? It's not harmful, just odd.

bhalevy: It's there to show that x is going away.
We can have `release_gate` operate on a lvalue but the fact it doesn't leave the gate, like `release` make it more suitable to operate on a rvalue since the released gate needs to be picked up and be tracked.

bhalevy: That's the way I started, but then I can't call it in `release()`.
Since it's private we can get away with just documenting any gotchas
and keep the call sites minimal. WDYT?

bhalevy:
In v3 ([5309ddf](https://github.com/scylladb/seastar/commit/5309ddf22d4661fb32502f0a36f778543df0b987)):
- Clarified comment for `release_gate`
- Call `release_gate` without `std:move`ing the holder.
- Rebased

include/seastar/core/gate.hh
xemul: One more presumably invalid corner case detected, but it worked before this patch, so I'll describe.
If you move-constuct or move-assign a released or (weird, but) moved-from gate then you call debug_hold_gate() with this->_g == nullptr and crash. Before this patch this would "just work"

bhalevy: > One more presumably invalid corner case detected, but it worked before this patch, so I'll describe. If you move-constuct or move-assign a released or (weird, but) moved-from gate then you call debug_hold_gate() with this->_g == nullptr and crash. Before this patch this would "just work"

fixed in v3 by linked the holder to the gate only when the `_g` gate* is not null.

bhalevy:
In v3 ([ec603b1](https://github.com/scylladb/seastar/commit/ec603b138cf414125b97d0c9cadc984910e28401)):
- rebased
- debug_hold_gate: link holder to gate only if `_g` is not null.

avikivity (approved):

include/seastar/core/gate.hh
xemul: The swap-node-ish move ctor would help to avoid this extra if, but since it's debug-mode that's ok

PR #1678, opened 2023-05-31
Cleanup RPC request/response frames maintenance

xemul:
The RPC request/response encoding/decoding code uses a bunch of magic hard-coded constants denoting sizes of request/response headers. The PR main intent is to replace them with named constants and respective math and checks around it.

While at it -- drop unused method and keep negotiation steps in one method.

### DIFF
--- a/include/seastar/rpc/rpc_impl.hh
+++ b/include/seastar/rpc/rpc_impl.hh
@@ -469,15 +469,10 @@ auto send_helper(MsgType xt, signature<Ret (InArgs...)> xsig) {
             // send message
             auto msg_id = dst.next_message_id();
             snd_buf data = marshall(dst.template serializer<Serializer>(), 28, args...);
-            static_assert(snd_buf::chunk_size >= 28, "send buffer chunk size is too small");
-            auto p = data.front().get_write() + 8; // 8 extra bytes for expiration timer
-            write_le<uint64_t>(p, uint64_t(t));
-            write_le<int64_t>(p + 8, msg_id);
-            write_le<uint32_t>(p + 16, data.size - 28);
 
             // prepare reply handler, if return type is now_wait_type this does nothing, since no reply will be sent
             using wait = wait_signature_t<Ret>;
-            return when_all(dst.send(std::move(data), timeout, cancel), wait_for_reply<Serializer>(wait(), timeout, cancel, dst, msg_id, sig)).then([] (auto r) {
+            return when_all(dst.request(uint64_t(t), msg_id, std::move(data), timeout, cancel), wait_for_reply<Serializer>(wait(), timeout, cancel, dst, msg_id, sig)).then([] (auto r) {
                     std::get<0>(r).ignore_ready_future();
                     return std::move(std::get<1>(r)); // return future of wait_for_reply
             });
--- a/include/seastar/rpc/rpc_impl.hh
+++ b/include/seastar/rpc/rpc_impl.hh
@@ -494,6 +494,9 @@ auto send_helper(MsgType xt, signature<Ret (InArgs...)> xsig) {
     return shelper{xt, xsig};
 }
 
+// Refer to struct response_frame for more details
+static constexpr size_t response_frame_headroom = 12;
+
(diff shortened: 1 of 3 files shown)


src/rpc/rpc.cc
```diff
@@ -643,6 +706,12 @@ namespace rpc {
           auto size = read_le<uint32_t>(ptr + 8);
           return std::make_tuple(msgid, size);
       }
+      static void encode_header(int64_t msg_id, snd_buf& data) {
+          static_assert(snd_buf::chunk_size >= raw_header_size, "send buffer chunk size is too small");
+          auto p = data.front().get_write();
+          write_le<int64_t>(p, msg_id);
+          write_le<uint32_t>(p + 8, data.size - raw_header_size);
```
tchaikov, line 713: shall we replace 8 with `sizeof(int64_t)` here? or, can we update `write_le<>`, so it looks something like:
```c++
template <typename T>
inline
char*
write_le(char* p, T datum) noexcept {
    datum = cpu_to_le(datum);
    return std::copy_n(reinterpret_cast<const char*>(&datum), sizeof(T), p);
}
```

so we can have something like:
```c++
auto p = data.front().get_write();
p = write_le<int64_t>(p, msg_id);
p = write_le<uint32_t>(p, data.size - raw_header_size);
```

?

include/seastar/rpc/rpc_impl.hh
```diff
@@ -451,6 +451,8 @@ relative_timeout_to_absolute(rpc_clock_type::duration relative) {
     return now + std::min(relative, rpc_clock_type::time_point::max() - now);
 }
 
+static constexpr size_t request_frame_headroom = 28;
```
tchaikov, line 454: IIUC, 28 is `request_frame_with_timeout::raw_header_size`. maybe we can have a comment explaining the connection between 28 and `request_frame_with_timeout`?

src/rpc/rpc.cc
```diff
@@ -626,14 +632,15 @@ namespace rpc {
           std::get<0>(h) = read_le<uint64_t>(ptr);
           return h;
       }
+      static void encode_header(uint64_t type, int64_t msg_id, snd_buf& buf) {
+          static_assert(snd_buf::chunk_size >= 28, "send buffer chunk size is too small");
```
tchaikov: could s/28/request_frame_headroom/ ?

src/rpc/rpc.cc
```diff
@@ -588,6 +588,46 @@ namespace rpc {
       return it->second;
   }
 
+  struct request_frame {
```
tchaikov, line 591: could you please add a couple words in the commit message explaining the motivation of this change?

avikivity:
@gleb-cloudius please review

src/rpc/rpc.cc
xemul: That's to make request_frame declared before `future<> client::request(...)` for the next patch. Will add

src/rpc/rpc.cc
```diff
@@ -593,8 +593,10 @@ namespace rpc {
       using header_and_buffer_type = std::tuple<std::optional<uint64_t>, uint64_t, int64_t, opt_buf_type>;
       using return_type = future<header_and_buffer_type>;
       using header_type = std::tuple<std::optional<uint64_t>, uint64_t, int64_t, uint32_t>;
+      static constexpr size_t raw_header_size = sizeof(uint64_t) + sizeof(int64_t) + sizeof(uint32_t);
```
gleb-cloudius, line 596: While we are at it may be we should document here what each of those is for?

gleb-cloudius (approved):

src/rpc/rpc.cc
xemul: Sure. I'll push v2 with more comments

xemul:
upd:
- added more comments about why sizes are such

gleb-cloudius (approved):

xemul:
@scylladb/seastar-maint , please review/merge

xemul:
@scylladb/seastar-maint , please review/merge

PR #1680, opened 2023-06-03
sharded: avoid #include <seastar/core/reactor.hh> for run_in_backgrou…

avikivity:
…nd()

reactor.hh is a heavy header and we should avoid including it, especially in common headers. sharded.hh includes it for the logger and for reactor::run_in_background().

Drop it by forward-declaring seastar_logger and adding a helper internal::run_in_background() which does not need reactor.hh.

### DIFF
--- a/include/seastar/core/sharded.hh
+++ b/include/seastar/core/sharded.hh
@@ -24,12 +24,12 @@
 #include <seastar/core/smp.hh>
 #include <seastar/core/loop.hh>
 #include <seastar/core/map_reduce.hh>
+#include <seastar/core/internal/run_in_background.hh>
 #include <seastar/util/is_smart_ptr.hh>
 #include <seastar/util/tuple_utils.hh>
 #include <seastar/core/do_with.hh>
 #include <seastar/util/concepts.hh>
 #include <seastar/util/log.hh>
-#include <seastar/core/reactor.hh>
 #include <seastar/util/modules.hh>
 
 #ifndef SEASTAR_MODULE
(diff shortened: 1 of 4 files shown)


include/seastar/core/sharded.hh
```diff
@@ -52,6 +52,8 @@
 
 namespace seastar {
 
+extern logger seastar_logger;
```
tchaikov, line 55: why? i can still compile seastar without this declaration.

nyh: I agree - sharded.hh doesn't need seastar_logger for itself. If existing code needed seastar_logger and got it accidentally (because reactor.hh), I don't think we need to support this - because the same can be true for basically anything which is defined in reactor.hh, not just seastar_logger.

By the way, I a natural place to put that "extern logger" thing is  <seastar/util/log.hh>. Since log.hh is still included in sharded.hh, it will also solve the problem for the user of sharded.hh who assumed that seastar_logger is defined.

avikivity: > I agree - sharded.hh doesn't need seastar_logger for itself. If existing code needed seastar_logger and got it accidentally (because reactor.hh), I don't think we need to support this - because the same can be true for basically anything which is defined in reactor.hh, not just seastar_logger.

I had a failure, will check again.

> 
> By the way, I a natural place to put that "extern logger" thing is <seastar/util/log.hh>. Since log.hh is still included in sharded.hh, it will also solve the problem for the user of sharded.hh who assumed that seastar_logger is defined.

No, log.hh is for people who want to use their own loggers.  seastar_logger is internal.

avikivity:
v2: drop logger declaration in shareded.hh, add to foreign_ptr_test.cc instead.

nyh (approved):

PR #1649, opened 2023-05-08
build: build ingredients using user-specified C compiler

tchaikov:
in this series, instead of using the default C compiler, we build liburing and DPDK using user-specified C compiler. this has two benefits:

1. be more consistent on fulfill user's configuration
2. silence the warning of 
    ```
    CMake Warning:
      Manually-specified variables were not used by the project:
    
        CMAKE_C_COMPILER
    ```
    it could be misleading sometimes, most of the times, it's just annoying.

PR #1682, opened 2023-06-05
Support ccache in ./configure.py

travisdowns:
By passing --ccache to configure.py, the CMAKE_C_COMPILER_LAUCHER (and equivalent CXX property) will be set to ccache, so that ccache will be used to cache the compilation results.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -70,6 +70,8 @@ arg_parser.add_argument('--compiler', action = 'store', dest = 'cxx', default = 
                         help = 'C++ compiler path')
 arg_parser.add_argument('--c-compiler', action='store', dest='cc', default='gcc',
                         help = 'C compiler path (for bundled libraries such as dpdk)')
+arg_parser.add_argument('--ccache', nargs='?', const='ccache', default='', metavar='CCACHE_BINARY_PATH',
+                        help = 'Use ccache to cache compilation (and optionally provide a path to ccache binary)')
 arg_parser.add_argument('--c++-standard', action='store', dest='cpp_standard', default='',
                         help='C++ standard to build with [default: %(default)s]')
 arg_parser.add_argument('--cook', action='append', dest='cook', default=[],
@@ -189,6 +191,7 @@ def configure_mode(mode):
         '-DCMAKE_C_COMPILER={}'.format(args.cc),
         '-DCMAKE_CXX_COMPILER={}'.format(args.cxx),
         '-DCMAKE_CXX_STANDARD={}'.format(args.cpp_standard),
+        '-DCMAKE_CXX_COMPILER_LAUNCHER={}'.format(args.ccache),
         '-DCMAKE_INSTALL_PREFIX={}'.format(args.install_prefix),
         '-DCMAKE_EXPORT_COMPILE_COMMANDS={}'.format('yes' if args.cc_json else 'no'),
         '-DBUILD_SHARED_LIBS={}'.format('yes' if mode in ('debug', 'dev') else 'no'),


travisdowns:
Tested locally with and without `--cache` and also that reconfiguring from with/without to the other works w/o needing to nuke the CMakeCache.txt.

configure.py
```diff
@@ -189,6 +191,8 @@ def configure_mode(mode):
         '-DCMAKE_C_COMPILER={}'.format(args.cc),
         '-DCMAKE_CXX_COMPILER={}'.format(args.cxx),
         '-DCMAKE_CXX_STANDARD={}'.format(args.cpp_standard),
+        '-DCMAKE_C_COMPILER_LAUNCHER={}'.format(args.ccache),
```
tchaikov, line 194: as a C++ project, seastar does not use the C compiler by itself, currently, the only consumer of the `CMAKE_C_COMPILER` is the `Cooking.cmake` script generated by `cooking.sh` . this CMake script is used for building the ingredients. and `CMAKE_C_COMPILER` is currently used for building CMake based ingredients. `fmt` and `yaml-cpp` is the only two in this category. but neither of them is a C project.

so without #1649, i am afraid that we cannot benefit from this setting. but even with #1649, we still need to populate this setting down to the C projects like c-ares, dpdk, liburing and lz4, when calling `cooking_ingredient()`.

xemul:
But doesn't ccache substitute itself transparently against pretty much every compiler out there so you don't need to do _anything_ in order to make it work?

tchaikov:
> But doesn't ccache substitute itself transparently against pretty much every compiler out there so you don't need to do _anything_ in order to make it work?

typically yes. by "typically", it means user need to 

1. use a shell which sources `/etc/profile` at start up. so that `/etc/profile` can source all files under `/etc/profile.d`. sure, Bash does..
1. install ccache into a directory where the `/etc/profile` can pick up the `ccache.sh` installed by ccache
2. run CMake in this shell, so that `/usr/lib64/ccache` is listed before, for instance `/usr/bin` in `$PATH`

IMHO, this change has its value to those who happens to not meet these assumptions. for instance, what if he/she use fish, and do not install CMake to the expected path?

travisdowns:
Other option, which I've sometimes wanted, is just a general CMake flag escape hatch, like "I know what I'm doing, please add -DFOO_BAR=blah on the cmake configure command line". It's already there for cxxflags and cflags so in some sense the precedent is set.

Then `configure.py` won't have to grow every time someone wants a new tweak to the cmake.

avikivity:
Ah, I see CMAKE_C_COMPILER_LAUNCHER is standard.

Yes, I agree that a way to pass cmake definitions is best. And in general automation should use cmake directly. ./configure.py is there to insulate normal human beings from that monstrosity.

xemul:
> > But doesn't ccache substitute itself transparently against pretty much every compiler out there so you don't need to do _anything_ in order to make it work?
> 
> typically yes. by "typically", it means user needs to
> 
>     1. use a shell which sources `/etc/profile` at start up. so that `/etc/profile` can source all files under `/etc/profile.d`. sure, Bash does..
> 
>     2. install ccache into a directory where the `/etc/profile` can pick up the `ccache.sh` installed by ccache
> 
>     3. run CMake in this shell, so that `/usr/lib64/ccache` is listed before, for instance `/usr/bin` in `$PATH`
> 
> 
> IMHO, this change has its value to those who happen to not meet these assumptions. for instance, what if he/she use fish, and do not install CMake to the expected path?

I must be too old-schooled, but 

```
$ echo $PATH
/usr/lib64/ccache:...
$ which g++
/usr/lib64/ccache/g++
```

so unless someone (who?) launches compiler with `execl()` giving it full path it should just work

travisdowns:
> so unless someone (who?) launches compiler with `execl()` giving it full path it should just work

The problem is the approach works for launching `g++` with `ccache` but not `/opt/something/gcc-trunk/bin/g++`, though I suppose I could also add `/opt/something/gcc-trunk/bin` to the front of `PATH` and just call it like `g++`.

However calling cmake with a modified PATH is IMO sort of an anti-pattern: it will break if you run the build (e..g, `ninja`, not talking about cmake) without the same PATH in the future: it detects the change and tries to reconfigure itself and pick up different compilers.

travisdowns:
Am I correctly reading between the lines that this change for cmake-style ccache is not going to make it in, but that a more general loophole to forward args to make would work?

xemul:
@tchaikov ?

tchaikov:
i am for this change in general, as it is the standard way to enable CMake to use ccache instead of relying on the default `$PATH`. the only thing i am not sure is https://github.com/scylladb/seastar/pull/1682/files#r1217402431

configure.py
travisdowns: @tchaikov - thanks for the background. I added `CMAKE_C_COMPILER_LAUNCHER` for symmetry with the C++ side of things and because this script does already set up other the CMAKE_C properties, though not for direct use of seastar as you point out.

Personally, I find that only the C++ launcher is really needed to use ccache usefully, as I'm very rarely rebuilding and of the decencies. So where does this leave us? Would you prefer I remove the C_COMPILER_LAUNCHER? I fine either way.

tchaikov: @travisdowns

hi Travis,

tl;dr, i'd like to drop it.

long story: as it adds another variable which is never picked up by CMake. it's just confusing from a picky tool (CMake actually complains at seeing unused variables) or a human developer's perspective.

and after digging deeper in the dependency building machinery, i realized that `Cooking.cmake` does not rebuild the ingredients at all. because, it adds the ingredients using `ExternalProject_add()`, which does not rebuild the external projects unless `BUILD_ALWAYS` is specified, see https://cmake.org/cmake/help/latest/module/ExternalProject.html . and presumably, like you are, neither do other Seastar application developers want to hack and rebuild the dependencies once they are built. so, to avoid addressing the non-existing needs, we don't even need to populate this flag down to `Cooking.cmake` via `cooking_recipe.cmake`

travisdowns: @tchaikov - sure, no problem. I've pushed an update which only sets `CMAKE_CXX_COMPILER_LAUNCHER`.

https://github.com/scylladb/seastar/commit/f4e36e91b5ef8da984cd2483296f265d354dfda6

tchaikov (approved):
lgtm

tchaikov:
@scylladb/seastar-maint hello maintainers, could you please help review and merge this change?

Issue #1645, opened 2023-05-03
flaky unit test httpd_test :: test_100_continue


nyh:
The test test_100_continue in tests/unit/httpd_test.cc failed twice today, in two circleci runs:

https://app.circleci.com/pipelines/github/avikivity/seastar/252/workflows/7fa40da2-16d2-4b2d-88a2-dfe266a083b4/jobs/857
https://app.circleci.com/pipelines/github/avikivity/seastar/252/workflows/7fa40da2-16d2-4b2d-88a2-dfe266a083b4/jobs/855

The error is the case in both:
```
/home/circleci/project/tests/unit/httpd_test.cc(810): fatal error: in "test_100_continue": critical check std::string(resp.get(), resp.size()).find("100 Continue") != std::string::npos has failed [18446744073709551615 == 18446744073709551615]
```

I don't know how often this test fails (I don't think circleci keeps a log of per-test failures?) but I found another - different - failure of the same test in https://github.com/scylladb/seastar/issues/1190. 

I am not familiar enough with this test to guess if this is a test bug or a real bug discovered by the test. @wmitros, you wrote this test (and the feature it tests), can you please take a look?

xemul:
The root-cause is that test writes several requests in a loop and reads back whatever arrives from server with a single `input_strea<>::read()` call. So at some point server's response is written in two parts so test loop next iteration instead of reading new server reply continues reading previous reply's 2nd part

PR #1685, opened 2023-06-06
dns: handle exception in do_sendv for udp

andijcr:
a std::system_error in udp.channel.send(...) can go unhandled, if the future is not available immediately. For example, ECONNABORTED can go uncaught and be shown in the log as an Error.
This commit adds a handler to log it as a warning, since the protocol already handles the case of lost udp packets

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -37,6 +37,7 @@
 #include <seastar/core/reactor.hh>
 #include <seastar/core/gate.hh>
 #include <seastar/core/print.hh>
+#include <system_error>
 
 namespace seastar::net {
 
@@ -845,16 +846,24 @@ private:
                     }).finally([fd, me = shared_from_this()] {
                         me->release(fd);
                     });
-                    // if we have a fast-fail, give error.
-                    if (e.udp.f.failed()) {
-                        try {
-                            e.udp.f.get();
-                        } catch (std::system_error& e) {
-                            errno = e.code().value();
-                        } catch (...) {
+
+                    if (e.udp.f.available()) {
+                        // if we have a fast-fail, give error.
+                        if (e.udp.f.failed()) {
+                            try {
+                                e.udp.f.get();
+                            } catch (std::system_error& e) {
+                                errno = e.code().value();
+                            } catch (...) {
+                            }
+                            e.udp.f = make_ready_future<>();
+                            return -1;
                         }
-                        e.udp.f = make_ready_future<>();
-                        return -1;
+                    } else {
+                        // ensure that no exception from channel.send is left uncaught
+                        e.udp.f = e.udp.f.handle_exception_type([](std::system_error const& e){
+                            dns_log.warn("UDP send exception: {}", e.what());
+                        });
                     }
                     // c-ares does _not_ use non-blocking retry for udp sockets. We just pretend
                     // all is fine even though we have no idea. Barring stack/adapter failure it


src/net/dns.cc
```diff
@@ -856,6 +857,11 @@ class net::dns_resolver::impl
                         e.udp.f = make_ready_future<>();
                         return -1;
                     }
+
+                    // ensure that no exception from channel.send is left uncaught
+                    e.udp.f = e.udp.f.handle_exception_type([](std::system_error const& e){
+                        dns_log.warn("UDP send exception: {}", e.what());
```
avikivity, line 863: Better to do this after the sendv(). Change finally to then_wrapped so you can handle it there. Fewer extra continuations, and more local error handling.

@elcallio please review too.

andijcr: But in that case, I have to remove the fast-fail branch on line 849, right?

dotnwat: It sorta looks like you can move the fast-fail case directly into the last `finally` (but switched to then_wrapped) without sacrificing the optimization. It's either going to be available at that point (finally is just a fancy then_wrapped) or you'd be waiting on it to become available (the case for the code you are adding now).

but yeh i didn't look in complete detail at what's going on.

elcallio: @avikivity `then_wrapped` solution will not be able to fast-fail (because we have no notion of fast/continued there). If you want to minimize continuations, why not just wrap with another/same check, i.e. `f.available()`, if, and failed -> fast-fail, if success -> success, if not available(), add this handle_exception. Since we will fast-succeed in 90% of cases anyway, it should not add much overhead.

andijcr: I'm trying the then_wrapped solution, but it gets somewhat awkward to make it work with the fast fail, in contrast, I think @elcallio suggestion is simpler to get right

andijcr:
force push: conditionally add handle_exeception_type only if the future is not immediately available

src/net/dns.cc
```diff
@@ -845,21 +846,31 @@ class net::dns_resolver::impl
...
-                    if (e.udp.f.failed()) {
-                        try {
-                            e.udp.f.get();
-                        } catch (std::system_error& e) {
-                            errno = e.code().value();
-                        } catch (...) {
+
+                    if (e.upd.f.available()) {
+                        // if we have a fast-fail, give error.
+                        if (!e.udp.f.failed()) {
```
elcallio, line 852: Code becomes shorter (and imho more readable) if you make it:

```
                    if (e.upd.f.available()) {
                        // if we have a fast-fail, give error.
                        if (e.udp.f.failed()) {
                            try {
                                e.udp.f.get();
                            } catch (std::system_error& e) {
                                errno = e.code().value();
                            } catch (...) {
                            }
                            e.udp.f = make_ready_future<>();
                            return -1;
                        }
                    } else {
                        // ensure that no exception from channel.send is left uncaught
                        e.udp.f = e.udp.f.handle_exception_type([](std::system_error const& e){
                            dns_log.warn("UDP send exception: {}", e.what());
                        });
                    }
                    // c-ares does _not_ use non-blocking retry for udp sockets. We just pretend
                    // all is fine even though we have no idea. Barring stack/adapter failure it
                    // is close to the same guarantee a "normal" message send would have anyway.
                    return bytes;
                default:
                    return -1;

File: src/net/dns.cc [hunk 3]
```diff
@@ -845,21 +846,31 @@ class net::dns_resolver::impl
...
                     });
-                    // if we have a fast-fail, give error.
-                    if (e.udp.f.failed()) {
-                        try {
-                            e.udp.f.get();
-                        } catch (std::system_error& e) {
-                            errno = e.code().value();
-                        } catch (...) {
+
+                    if (e.upd.f.available()) {
```
andijcr, line 850: silly mistake upd<->udp

src/net/dns.cc
andijcr: agreed, done

src/net/dns.cc
dotnwat: @elcallio i'm a little curious about this statement, but perhaps i'm about to learn something new!

> then_wrapped solution will not be able to fast-fail (because we have no notion of fast/continued there)

in the follow proposed change, if `e.udp.f` is available, then it implies that the full chain above the available check, including the finally blocks were fully evaluated (otherwise, how would it be available)?

```c++
                    e.udp.f = e.udp.f.finally([&e, p = std::move(p)]() mutable {
                        return e.udp.channel.send(e.udp.dst, std::move(p));;
                    }).finally([fd, me = shared_from_this()] {
                        me->release(fd);
                    });
                    if (e.udp.f.available()) {
                        // if we have a fast-fail, give error.
                        if (e.udp.f.failed()) {
```

if that's the case, then checking inside an attached `then_wrapped` seems like it would be equivalent in cost (because a `then_wrapped` is already there in the form of `finally`).

andijcr: @dotnwat The problem I encountered is that fast-fail needs to see the exception, but if then_wrapped absorbs the exception, we lose that data channel. I tried to change from future<> to future<std::error_code>, but the change would be bigger and It would have to deal with the future destruction done by me->release(fd)

dotnwat: Thanks @elcallio for confirming my understanding. You're right, of course about the return code! I realize now in the back of my mind I thinking about capturing a stack variable reference, but that's massively problematic :)

andijcr:
Failure is a known issue https://github.com/scylladb/seastar/issues/1645

dotnwat:
@elcallio gentle ping here. looks like known failure in CI ?

elcallio:
@avikivity ?

dotnwat:
@elcallio @avikivity is there something more that needs to be done for this?

avikivity:
Looks okay to me, but would like review by @elcallio .

avikivity:
I see there's a "?", but it doesn't count for review.

PR #1688, opened 2023-06-07
Add file::list_directory() that co_yields entries

xemul:
Current implementation of list_directory() requires a callback function that's called with entries. That's not very convenient for coroutines-enabled code. E.g. Scylla has the convenience emulation of this approach, see scylladb/scylladb@207174c6927

### DIFF
--- a/include/seastar/core/file.hh
+++ b/include/seastar/core/file.hh
@@ -173,9 +173,7 @@ public:
     virtual std::unique_ptr<file_handle_impl> dup();
     virtual subscription<directory_entry> list_directory(std::function<future<> (directory_entry de)> next) = 0;
 #ifdef SEASTAR_COROUTINES_ENABLED
-    virtual coroutine::experimental::generator<directory_entry, dir_entry_buffer> experimental_list_directory() {
-        throw std::runtime_error("not implemented");
-    }
+    virtual coroutine::experimental::generator<directory_entry, dir_entry_buffer> experimental_list_directory();
 #endif
 
     friend class reactor;
--- a/include/seastar/core/file.hh
+++ b/include/seastar/core/file.hh
@@ -21,6 +21,8 @@
 
 #pragma once
 
+#include <seastar/core/coroutine.hh>
+#include <seastar/coroutine/generator.hh>
 #include <seastar/core/do_with.hh>
 #include <seastar/core/stream.hh>
 #include <seastar/core/sstring.hh>
@@ -28,6 +30,7 @@
 #include <seastar/core/align.hh>
 #include <seastar/core/io_priority_class.hh>
 #include <seastar/core/file-types.hh>
+#include <seastar/core/circular_buffer.hh>
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/modules.hh>
 #ifndef SEASTAR_MODULE
@@ -110,6 +113,9 @@ public:
     virtual shared_ptr<file_impl> to_file() && = 0;
 };
 
+template <typename T>
+using dir_entry_buffer = circular_buffer<T>;
+
 class file_impl {
     friend class file;
 protected:
@@ -166,6 +172,11 @@ public:
     virtual future<> close() = 0;
     virtual std::unique_ptr<file_handle_impl> dup();
     virtual subscription<directory_entry> list_directory(std::function<future<> (directory_entry de)> next) = 0;
+#ifdef SEASTAR_COROUTINES_ENABLED
+    virtual coroutine::experimental::generator<directory_entry, dir_entry_buffer> experimental_list_directory() {
+        throw std::runtime_error("not implemented");
+    }
+#endif
 
     friend class reactor;
 };
@@ -680,6 +691,11 @@ public:
 
     /// Returns a directory listing, given that this file object is a directory.
     subscription<directory_entry> list_directory(std::function<future<> (directory_entry de)> next);
+
+#ifdef SEASTAR_COROUTINES_ENABLED
+    /// Returns a directory listing, given that this file object is a directory.
+    coroutine::experimental::generator<directory_entry, dir_entry_buffer> experimental_list_directory();
+#endif
 
 #if SEASTAR_API_LEVEL < 7
     /**
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -395,6 +395,37 @@ static std::optional<directory_entry_type> dirent_type(const linux_dirent64& de)
     return type;
 }
 
+#ifdef SEASTAR_COROUTINES_ENABLED
+static coroutine::experimental::generator<directory_entry, dir_entry_buffer> make_list_directory_generator(coroutine::experimental::buffer_size_t buffer_size, thread_pool& pool, int fd) {
+    temporary_buffer<char> buf(8192);
+
+    while (true) {
+        auto size = co_await engine().read_directory(fd, buf.get_write(), buf.size());
+        if (size == 0) {
+            co_return;
+        }
+
+        for (const char* b = buf.get(); b < buf.get() + size; ) {
+            const auto de = reinterpret_cast<const linux_dirent64*>(b);
+            b += de->d_reclen;
+            sstring name(de->d_name);
+            if (name == "." || name == "..") {
+                continue;
+            }
+            std::optional<directory_entry_type> type = dirent_type(*de);
+            directory_entry ret(std::move(name), type);
+            co_yield ret;
+        }
+    }
+}
+
+coroutine::experimental::generator<directory_entry, dir_entry_buffer> posix_file_impl::experimental_list_directory() {
+    // Keep 8 entries. The sizeof(directory_entry) is 24 bytes, the name itself 
+    // is allocated out of this buffer, so the buffer would grow up to ~200 bytes
+    return make_list_directory_generator(coroutine::experimental::buffer_size_t{8}, *engine()._thread_pool, _fd);
+}
+#endif
+
 subscription<directory_entry>
 posix_file_impl::list_directory(std::function<future<> (directory_entry de)> next) {
     static constexpr size_t buffer_size = 8192;
@@ -1127,6 +1158,12 @@ subscription<directory_entry>
 file::list_directory(std::function<future<>(directory_entry de)> next) {
     return _file_impl->list_directory(std::move(next));
 }
+
+#ifdef SEASTAR_COROUTINES_ENABLED
+coroutine::experimental::generator<directory_entry, dir_entry_buffer> file::experimental_list_directory() {
+    return _file_impl->experimental_list_directory();
+}
+#endif
 
 future<int> file::ioctl(uint64_t cmd, void* argp) noexcept {
     return _file_impl->ioctl(cmd, argp);
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -364,6 +364,37 @@ blockdev_file_impl::size() noexcept {
     });
 }
 
+static std::optional<directory_entry_type> dirent_type(const linux_dirent64& de) {
+    std::optional<directory_entry_type> type;
+    switch (de.d_type) {
+    case DT_BLK:
+        type = directory_entry_type::block_device;
+        break;
+    case DT_CHR:
+        type = directory_entry_type::char_device;
+        break;
+    case DT_DIR:
+        type = directory_entry_type::directory;
+        break;
+    case DT_FIFO:
+        type = directory_entry_type::fifo;
+        break;
+    case DT_REG:
+        type = directory_entry_type::regular;
+        break;
+    case DT_LNK:
+        type = directory_entry_type::link;
+        break;
+    case DT_SOCK:
+        type = directory_entry_type::socket;
+        break;
+    default:
+        // unknown, ignore
+        ;
+    }
+    return type;
+}
+
 subscription<directory_entry>
 posix_file_impl::list_directory(std::function<future<> (directory_entry de)> next) {
     static constexpr size_t buffer_size = 8192;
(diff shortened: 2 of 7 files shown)


include/seastar/core/file.hh
```diff
@@ -166,6 +172,11 @@ public:
     virtual future<> close() = 0;
     virtual std::unique_ptr<file_handle_impl> dup();
     virtual subscription<directory_entry> list_directory(std::function<future<> (directory_entry de)> next) = 0;
+#ifdef SEASTAR_COROUTINES_ENABLED
+    virtual coroutine::experimental::generator<directory_entry, dir_entry_buffer> list_directory() {
+        throw std::runtime_error("not implemented");
```
avikivity: It's a little dangerous to have a formal API return an experimental thing, no?

src/core/file.cc
```diff
@@ -383,6 +383,40 @@ std::optional<directory_entry_type> dirent_type(const linux_dirent64& de) {
...
 }
 
+#ifdef SEASTAR_COROUTINES_ENABLED
+coroutine::experimental::generator<directory_entry, dir_entry_buffer> make_list_directory_generator(coroutine::experimental::buffer_size_t buffer_size, thread_pool& pool, int fd) {
+    temporary_buffer<char> buf(8192);
+
+    while (true) {
+        auto ret = co_await pool.submit<syscall_result<long>>([fd, &buf] {
+            auto ret = ::syscall(__NR_getdents64, fd, reinterpret_cast<linux_dirent64*>(buf.get_write()), buf.size());
+            return wrap_syscall(ret);
```
avikivity, line 394: Can't it use the existing APIs? Why reimplement the syscall issue/parse?

xemul: It's possible to make the new API appear on the file level only and it will re-use existing file_impl-s facilities. There's `make_list_directory_fallback_generator` in the last patch of this PR that does it. In the end of the day it needs seastar::queue on top what this implementation has.

include/seastar/core/file.hh
xemul: It is, but the API extension thus becomes experimental itself. Those who care about stability can stick to legacy list_directory() and avoid using new list_directory() until it returns non-experimental generator.

avikivity: It's too tricky. No one will notice that it's experimental.

src/core/file.cc
avikivity: Yes, but why not call the existing API here? Instead of duplicating the implementation?

include/seastar/core/file.hh
tchaikov: how about using `[[[deprecated](https://en.cppreference.com/w/cpp/language/attributes/deprecated)]]` ? i understand that this attribute has "deprecated" in its name but as the standard notes

> The [attribute-token](https://eel.is/c++draft/dcl.attr.grammar#nt:attribute-token) deprecated can be used to mark names and entities whose use is still allowed, but is discouraged for some reason[.](https://eel.is/c++draft/dcl.attr.deprecated#1.sentence-1)

see https://eel.is/c++draft/dcl.attr.deprecated. so we can add an appropriate warning message here. and once this API is proved mature enough. we can promote it and the underlying generator so they won't be marked experimental.

src/core/file.cc
tchaikov: @xemul i guess what Avi suggests is to make the generator a wrapper of the existing `list_directory()`?

xemul: Yes, I got Avi's point, but my plan was to encourage file_impl::list_directory() switch to generator-style too. If implementing `make_list_directory_fallback_generator` on file level, it will keep all file_impls old-style for pretty much forever.

include/seastar/core/file.hh
xemul: Marking new API with deprecated is too tricky too :) We want the `[[experimental]]` attribute here

xemul:
update:
- name new method `experimental_list_directory()`. It will need to be renamed regardless of the way it's marked as such (unless we get `[[experimental]]` attribute)
- detach call to `sys_getdents(fd, ...)` via thread-pool to reactor method. Other than giving us more code reuse, this also relaxes `friend`ship between reactor and posix_file_impl used _only_ for the latter to access reactor::_thread_pool

include/seastar/core/file.hh
xemul: Named the new method explicitly as `experimental_list_directory` in v2. Without attribute callers will need to rename something on their side anyway, be it `file fd; fd.experimental.list_directory();` or `fd.list_directory(experimental_tag{})` or `experimental(fd).list_directory();` or anything else that come to my mind :(

src/core/reactor.cc
```diff
@@ -1959,6 +1959,16 @@ reactor::fstat(int fd) noexcept {
     });
 }
 
+future<size_t> reactor::fill_dirent_buffer(int fd, char* buffer, size_t buffer_size) {
```
tchaikov, line 1962: nit, i'd suggest rename this function to `get_dir_entries()` for better readability, and filling a buffer just but how this function is exposed to its caller, just like how the `read(2)` is exposed to the libc users. i understand that `read(2)` is named "read" for a handful reasons, for instance, its inventor was in favor of very short names back in the old days, but if we compare them we can see why `read()` is a better name than `fill_read_buffer()`.

src/core/file.cc
```diff
@@ -1346,6 +1347,32 @@ file_impl::dup() {
     throw std::runtime_error("this file type cannot be duplicated");
 }
 
+#ifdef SEASTAR_COROUTINES_ENABLED
+coroutine::experimental::generator<directory_entry, dir_entry_buffer> make_list_directory_fallback_generator(coroutine::experimental::buffer_size_t buffer_size, file_impl& me) {
```
tchaikov, line 1351: might want to mark this function `static` or move it into an anonymous namespace.

tests/unit/directory_test.cc
```diff
@@ -74,13 +76,18 @@ int main(int ac, char** av) {
             });
         }
     };
+    fmt::print("--- Regular lister test ---\n");
```
tchaikov, line 79: note to myself, i see the opportunity to transform this test into a `SEASTAR_THREAD_TEST_CASE` based tests.

src/core/reactor.cc
xemul: I'd then named it after the syscall -- getdents() or readdir(). WDYT, @tchaikov ?

src/core/file.cc
xemul: Good point. I had this habit to mark .cc-local methods `static` from my C past, but then abandoned it for some reason

xemul:
upd:
- renamed reactor::fill_dirent_buffer() into reactor::readdir()
- marked make...lister_generator() helpers static

include/seastar/core/file.hh
michoecho: > Marking new API with deprecated is too tricky too :) We want the [[experimental]] attribute here

Yet another reason to fork the compiler.

src/core/reactor.cc
tchaikov: yeah. `readdir()` is better. i'd prefer using a name which is more consistent with other names in reactor. please note, we already have names like

- `make_directory`
- `touch_directory`

so probably `get_directory_entries()` or `read_directory()` is a better alternative.

tchaikov (commented):
left a couple comments on the naming and a magic number.

src/core/file.cc
```diff
@@ -381,6 +381,35 @@ std::optional<directory_entry_type> dirent_type(const linux_dirent64& de) {
...
+            }
+            std::optional<directory_entry_type> type = dirent_type(*de);
+            directory_entry ret(std::move(name), type);
+            co_yield ret;
+        }
+    }
+}
+
+coroutine::experimental::generator<directory_entry, dir_entry_buffer> posix_file_impl::experimental_list_directory() {
+    return make_list_directory_generator(coroutine::experimental::buffer_size_t{128 / sizeof(directory_entry)}, *engine()._thread_pool, _fd);
```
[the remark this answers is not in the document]
xemul: > sizeof(seastar::directory_entry) is 24. i am not able to find any clues on where 128 comes from.

The buffer-size argument here is the number of entries, not the number of bytes, isn't it?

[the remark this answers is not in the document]
tchaikov: yeah, number of elements, the former.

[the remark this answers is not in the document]
tchaikov: probably i failed to explain my question. let me rephrase: where does the 128 in "128 / sizeof(directory_entry)" come from?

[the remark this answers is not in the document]
xemul: From ... dunno ... pretty much nowhere, just a buffer that's larger than a single entry, but still small enough. And the power-of-two of course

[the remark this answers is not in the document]
xemul: BTW, name on directory entry is sstring, not online buffer, so it doesn't contribute to generator buffer capacity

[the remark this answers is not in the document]
michoecho: > From ... dunno ... pretty much nowhere, just a buffer that's larger than a single entry, but still small enough. And the power-of-two of course

I think @tchaikov's point is that it's a magic number, not that it's a wrong number.
Whatever is the reason for choosing 128, it should be explained with a comment it the code.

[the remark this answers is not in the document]
tchaikov: Both, I also suggest another magic number in the first comment in this thread.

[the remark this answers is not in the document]
tchaikov: > BTW, name on directory entry is sstring, not online buffer, so it doesn't contribute to generator buffer capacity

Hmm, the name string is returned by the provided buffer. I double checked by reading the Linux source.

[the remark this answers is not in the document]
xemul: The generator buffer is used to keep seastar dirents in, not linux dirents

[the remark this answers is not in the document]
xemul: > Whatever is the reason for choosing 128, it should be explained with a comment in the code.
> ...
> Both, I also suggest another magic number in the first comment in this thread.

Magic of 128 is the same as magic of 8192 to me :) From the "no magic" perspective it better be a generator with std::optional<> buffer type feeding one dirent at a time. No magic here, but not nice either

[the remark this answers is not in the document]
xemul: I think that the source of confusion is that there are _two_ buffers.

One is used to get linux dirents from the kernel. This buffer is 8k in size and it's just the same as in legacy list_directory() method. This buffers fits in dirents _and_ their names. It's "that large" to make less (pretty expensive) system calls.

The other one is the generator buffer used by co_yield code to accumulate several yielded elements before switching between coroutines. It's 128 bytes in size, not 8k, because it's used to keep _only_ 24-bytes seastar dirents, since their names are sstrings and are allocated _elsewhere_. This buffer doesn't need to be 8k as it doesn't lead to more system calls, only coroutines switches.

tchaikov (approved):

src/core/file.cc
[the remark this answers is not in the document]
tchaikov: i see. sorry for the noise. and thank you for the detailed explanations.

xemul:
@avikivity , regarding [your concern](https://github.com/scylladb/seastar/pull/1688#discussion_r1221860402), does the below justification sounds reasonable to you?

> I still want to keep explicit old and new listing calls in file_impl. It's merely in the number of "moving parts" if short-circuiting the new one via the old one.
> 
> In independent implementation yielding code has
> 
> *  temporary buffer for syscall
> *  generator buffer
> 
> If yielding call calls legacy one it will have
> 
>  * temporary buffer for syscall (in old code)
>  * seastar::stream (in old code) to push dentries to callback function
>  * seastar::queue to be able to exchange them between callback function and experimental_list_directory body
>  * generator buffer

If so, is there's anything else that keeps you from merging it?

tchaikov:
@avikivity could you please take another look?

tchaikov:
@avikivity ping?

xemul:
upd:
- rebased (had conflicts)

@avikivity , take another look, please

avikivity:
looking

src/core/file.cc
```diff
@@ -420,16 +420,12 @@ posix_file_impl::list_directory(std::function<future<> (directory_entry de)> nex
...
         return do_until(eofcond, [w, this] {
             if (w->current == w->total) {
-                return engine()._thread_pool->submit<syscall_result<long>>([w , this] () {
-                    auto ret = ::syscall(__NR_getdents64, _fd, reinterpret_cast<linux_dirent64*>(w->buffer), buffer_size);
-                    return wrap_syscall(ret);
-                }).then([w] (syscall_result<long> ret) {
-                    ret.throw_if_error();
-                    if (ret.result == 0) {
+                return engine().read_directory(_fd, w->buffer, buffer_size).then([w] (auto size) {
+                    if (size == 0) {
```
avikivity, line 456: This also takes us one step closer to implementing it with io_uring (not that it's important)

tests/unit/directory_test.cc
```diff
@@ -74,13 +76,18 @@ int main(int ac, char** av) {
...
+    return engine().open_directory(".").then([] (file f) {
+        return do_with(lister(std::move(f)), [] (lister& l) {
+          return l.done();
```
avikivity, line 82: The test is fairly bad since nothing guarantees it will test multiple calls into readdir.

avikivity: (but of course a preexisting problem)

include/seastar/core/file.hh
avikivity: btw this will allow us to gain experience with the interface an unmark it as experimental. We should also compare it with std::generator and see that it matches as closely as possible.

src/core/file.cc
```diff
@@ -395,6 +395,37 @@ static std::optional<directory_entry_type> dirent_type(const linux_dirent64& de)
...
+            directory_entry ret(std::move(name), type);
+            co_yield ret;
+        }
+    }
+}
+
+coroutine::experimental::generator<directory_entry, dir_entry_buffer> posix_file_impl::experimental_list_directory() {
+    // Keep 8 entries. The sizeof(directory_entry) is 24 bytes, the name itself 
+    // is allocated out of this buffer, so the buffer would grow up to ~200 bytes
+    return make_list_directory_generator(coroutine::experimental::buffer_size_t{8}, *engine()._thread_pool, _fd);
```
avikivity, line 426: Too little IMO, we should amortize the syscall more.

avikivity:
Looks good. I left a comment about the number "8".

src/core/file.cc
xemul: This constant doesn't amortize the syscall, it amortizes generator buffer draining. The one to amortize syscall is here:

```c++
static coroutine::experimental::generator<directory_entry, dir_entry_buffer> make_list_directory_generator(coroutine::experimental::buffer_size_t buffer_size, thread_pool& pool, int fd) {
    temporary_buffer<char> buf(8192);

    while (true) {
        auto size = co_await engine().read_directory(fd, buf.get_write(), buf.size());
        ...
```

copied from current non-generator implementation

xemul:
> Looks good. I left a comment about the number "8".

@avikivity , reminding (I commented about the number 8 concern)

Issue #1691, opened 2023-06-08
debug builds: warn that -m parameters is not applied


gusev-p:
Debug builds use `ASAN` allocator and ignore this parameter. Currently there are no clear signs that it's not applied, which causes misunderstanding and wasted time. For example, it took some time to get why [this test](https://github.com/scylladb/scylladb/blob/5acb137c2e9b14d15e8147d8c071671446873e09/test/cql-pytest/test_shedding.py#L76) test was failing in debug builds. It made some assumptions about available memory, which break in debug builds. Recently, @alecco stumbled upon the same problem.

denesb:
@gusev-p for some reason this was not closed even though https://github.com/scylladb/seastar/pull/1693 was merged. Can we close this?

nyh:
According to https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue, one has to write "Fixes #1691" - adding a colon "Fixes: #1691" is not ok :-(

gusev-p:
> According to https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue, one has to write "Fixes #1691" - adding a colon "Fixes: #1691" is not ok :-(

thanks

PR #1693, opened 2023-06-09
smp: warn if --memory parameter is not supported

gusev-p:
This parameter is ignored if `SEASTAR_DEFAULT_ALLOCATOR` build option is set. For Scylla this is true for debug builds to enable `ASAN` allocator. There was no clear indication that it was not applied before, which could cause misunderstanding and wasted time during development.

`--memory` is one of the options in `smp_options` and `reactor_options` which won't have effect if Seastar allocator was not compiled in. We emit a warning only for --`memory` since other options are
way less common and impactful.

Fixes: #1691

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4213,6 +4213,9 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     }
     std::vector<reactor*> reactors(smp::count);
     if (smp_opts.memory) {
+#ifdef SEASTAR_DEFAULT_ALLOCATOR
+        seastar_logger.warn("Seastar compiled with default allocator, --memory option won't take effect");
+#endif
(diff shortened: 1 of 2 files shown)


src/core/reactor.cc
```diff
@@ -4139,6 +4139,12 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     }
 #endif
 
+#ifdef SEASTAR_DEFAULT_ALLOCATOR
+    if (!smp_opts.memory.defaulted()) {
+        seastar_logger.warn("Seastar compiled with default allocator, --memory limit is not supported");
+    }
+#endif
```
denesb, line 4146: I think this should be in the `if (smp_opts.memory) {` scope below.

gusev-p: done

denesb (approved):

PR #1694, opened 2023-06-09
deleter: Delete move and copy constructors

xemul:
It looks like this small helper object has them defaulted which likely leads to double ::free() of memory. It doesn't happen yet, but it's better to be on the safer side.

### DIFF
--- a/include/seastar/core/deleter.hh
+++ b/include/seastar/core/deleter.hh
@@ -193,6 +193,8 @@ SEASTAR_MODULE_EXPORT_END
 struct free_deleter_impl final : deleter::impl {
     void* obj;
     free_deleter_impl(void* obj) : impl(deleter()), obj(obj) {}
+    free_deleter_impl(const free_deleter_impl&) = delete;
+    free_deleter_impl(free_deleter_impl&&) = delete;
     virtual ~free_deleter_impl() override { std::free(obj); }
 };
 /// \endcond


tchaikov:
i'd suggest be more explicit in the commit message, as otherwise, it might be considered as fix. but IMHO, it's more like a comment for this class.

include/seastar/core/deleter.hh
```diff
@@ -193,6 +193,8 @@ SEASTAR_MODULE_EXPORT_END
 struct free_deleter_impl final : deleter::impl {
     void* obj;
     free_deleter_impl(void* obj) : impl(deleter()), obj(obj) {}
+    free_deleter_impl(const free_deleter_impl&) = delete;
+    free_deleter_impl(free_deleter_impl&&) = delete;
     virtual ~free_deleter_impl() override { std::free(obj); }
```
avikivity, line 198: As soon as you have a virtual destructor, the genereated copy/move stuff is disabled.

xemul: https://i.stack.imgur.com/b2VBV.png

Declared destructor disables only move-stuff

xemul:
> if we have any customized constructor defined, compiler won't generate the defaulted copy ctor for us.
> 
> if we have a customized dtor, the compiler won't generate the defaulted move ctor for us.
> 
> but yeah, better to be more explicit.

https://i.stack.imgur.com/b2VBV.png

Any destructor declared doesn't delete copy-stuff

tchaikov:
> > if we have any customized constructor defined, compiler won't generate the defaulted copy ctor for us.
> > if we have a customized dtor, the compiler won't generate the defaulted move ctor for us.
> > but yeah, better to be more explicit.
> 
> https://i.stack.imgur.com/b2VBV.png
> 
> Any ~de~constructor declared doesn't delete copy-stuff

ah, right. thank you for correcting my remarks. for nun-human-readable proof, see also https://eel.is/c++draft/class.default.ctor

include/seastar/core/deleter.hh
avikivity: Oops! I see there was a proposal to deprecate copy construction/assignment, if a destructor is defined, but nothing done.

PR #1695, opened 2023-06-11
file: call close() without the syscall thread

### LINKED ISSUE #1689: close() of read-only files needlessly uses syscall thread
```
11:34:15.362633 read(70, "\1\0\0\0\0\0\0\0", 8) = 8 <0.317622>
11:34:15.680493 rename("/var/lib/scylla/commitlog/Recycled-CommitLog-2-108086411005585196.log", "/var/lib/scylla/commitlog/CommitLog-2-108086411005585284.log") = 0 <0.000100>
11:34:15.680720 read(70, "\1\0\0\0\0\0\0\0", 8) = 8 <0.000016>
11:34:15.680798 openat(AT_FDCWD, "/var/lib/scylla/commitlog", O_RDONLY|O_CLOEXEC|O_DIRECTORY) = 29544 <0.000040>
11:34:15.680886 read(70, "\1\0\0\0\0\0\0\0", 8) = 8 <0.000073>
11:34:15.681235 read(70, "\1\0\0\0\0\0\0\0", 8) = 8 <0.000260>
11:34:15.681555 close(29544)            = 0 <0.000022>
```

close() of files opened for read-only, or directories, could be done without the syscall thread.

_Originally posted by @avikivity in https://github.com/scylladb/scylladb/issues/13753#issuecomment-1543843104_

tchaikov:
file: call close() without the syscall thread

the close() system call

1. flushes the data if any
2. reclaims the fd

so far, most popular local filesystems do not implement
the "flush" op, but quite a few network filesystems
and layered filesystems implement it. for instance, cifs,
f2fs, fuse, nfs implement the "flush" op. so to be on the
safe side, let's call close() in reactor only when the
file is opened read-only. presumably, the fs implementation
mutate the its state only if the file is opened in writeable
flag enabled. this behavior is confirmed after auditing
all filesystems with flush op in linux v6.3.

Fixes #1689

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -318,17 +318,16 @@ posix_file_impl::close() noexcept {
             return make_ready_future<syscall_result<int>>(wrap_syscall<int>(::close(fd)));
         }();
     } else {
-    // TODO: fix the indent
-    closed = [fd] () noexcept {
-        try {
-            return engine()._thread_pool->submit<syscall_result<int>>([fd] {
-                return wrap_syscall<int>(::close(fd));
-            });
-        } catch (...) {
-            report_exception("Running ::close() in reactor thread, submission failed with exception", std::current_exception());
-            return make_ready_future<syscall_result<int>>(wrap_syscall<int>(::close(fd)));
-        }
-    }();
+        closed = [fd] () noexcept {
+            try {
+                return engine()._thread_pool->submit<syscall_result<int>>([fd] {
+                    return wrap_syscall<int>(::close(fd));
+                });
+            } catch (...) {
+                report_exception("Running ::close() in reactor thread, submission failed with exception", std::current_exception());
+                return make_ready_future<syscall_result<int>>(wrap_syscall<int>(::close(fd)));
+            }
+        }();
     }
     return closed.then([] (syscall_result<int> sr) {
       try {
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -312,7 +312,14 @@ posix_file_impl::close() noexcept {
     }
     delete _refcount;
     _refcount = nullptr;
-    auto closed = [fd] () noexcept {
+    auto closed = make_ready_future<syscall_result<int>>(0, 0);
+    if ((_open_flags & open_flags::ro) != open_flags{}) {
+        closed = [fd] () noexcept {
+            return make_ready_future<syscall_result<int>>(wrap_syscall<int>(::close(fd)));
+        }();
+    } else {
+    // TODO: fix the indent
+    closed = [fd] () noexcept {
         try {
             return engine()._thread_pool->submit<syscall_result<int>>([fd] {
                 return wrap_syscall<int>(::close(fd));
@@ -322,6 +329,7 @@ posix_file_impl::close() noexcept {
             return make_ready_future<syscall_result<int>>(wrap_syscall<int>(::close(fd)));
         }
     }();
+    }
     return closed.then([] (syscall_result<int> sr) {
       try {
         sr.throw_if_error();


raphaelsc:
close syscall syncs, and we had significant reactor stalls before due to syncs. I don't think it's a good idea to stop using syscall thread for files opened for write. note eventually we'll be able to push close syscalls through iouring submission queue.

seastar applications care a lot about latency, and I don't think users would be happy with their cpu being idle while linux waits for an outstanding I/O before proceeding with sync (they're serialized) on close. you cannot assume either that sync will complete fast. relatively faster is not an argument for allowing a blocking sync, even one which takes < 1ms. the reactor is a like a factory pipeline, if you block the cpu, you're preventing tasks from performing computation on the data previously retrieved through I/O, making the factory way less efficient.

tchaikov:
> close syscall syncs, and we had significant reactor stalls before due to syncs.

@raphaelsc could you provide more details or any documents, references / quotations on this, so I can read to understand your remarks on it?

raphaelsc:
> > close syscall syncs, and we had significant reactor stalls before due to syncs.
> 
> @raphaelsc could you provide more details or any documents, references / quotations on this, so I can read to understand your remarks on it?

we don't issue blocking syscalls in the reactor thread. that's our most basic guideline with our reactor model. stalls are our evil enemy. the framework is latency oriented, and users will not be happy if they see some spike in latency whenever a file is closed that was not explicitly synced before. you cannot make assumptions on the sync latency whatsoever.

tchaikov:
I am not sure how do you define blocking call in this context. Actually, I am trying to understand why you think close() syncs. That is a more specific thing.

raphaelsc:
> I am not sure how do you define blocking call in this context. Actually, I am trying to understand why you think close() syncs. That is a more specific thing.

you're right about sync, it's done only optionally like on NFS.

__fput(), called by close sycall, calls might_sleep(), which detones a function that might sleep.

in __fput(), we see:

```
	if (file->f_op->release)
		file->f_op->release(inode, file);
```
		
which in turn calls xfs's release entrypoint

xfs's release may trigger I/O under certain conditions.

note it's a very common scenario for a file to be closed after unlinked, so it's possible that release will trigger I/O to reclaim blocks from an inode once its refcount drops to zero. even if it's scheduled in background, I think xfs may persist the intent into a log.

and let's remember this is all file system specific. there's a reason why open syscall interface may return EIO, which is because it may trigger I/O, and again, we don't issue blocking syscalls in the reactor thread. I doubt this patch will go in as is. I see that even @avikivity suggested dropping the syscall thread usage only for read-only files, and that's what we should do. we must be conservative here, @tchaikov. please proceed as suggested.

tchaikov:
okay, when it comes to the network filesystems, both nfs and cifs do implement the `flush` op. so i think that's probably what Avi and you are talking about. if we really want to play safe, then yes. 

and yes, i do understand we don't do block io in reactor thread.

tchaikov:
> > I am not sure how do you define blocking call in this context. Actually, I am trying to understand why you think close() syncs. That is a more specific thing.
> 
> you're right about sync, it's done only optionally like on NFS.
> 
> __fput(), called by close sycall, calls might_sleep(), which detones a function that might sleep.
> 
> in __fput(), we see:
> 
> ```
> 	if (file->f_op->release)
> 		file->f_op->release(inode, file);
> ```
> 

i am not 100% sure i am correct. but i think `__fput()` is scheduled in background.

avikivity:
We've seen stalls on NFS, I think that's the only problematic one.

tchaikov:
@raphaelsc @avikivity , hi Raphael and Avi, thank you very much for your insights! just read the `flush` implementation of all the fs driver in linux v6.3, and confirmed:
```console
$ git describe --tags
v6.3
$ git grep -w \\.flush | cut -d':' -f1 | uniq
cifs/cifsfs.c
cifs/smb1ops.c
cifs/smb2ops.c
ecryptfs/file.c
f2fs/file.c
fuse/file.c
nfs/file.c
nfs/nfs4file.c
nfsd/export.c
orangefs/file.c
overlayfs/file.c
```

src/core/file.cc
```diff
@@ -318,17 +318,16 @@ posix_file_impl::close() noexcept {
...
+            try {
+                return engine()._thread_pool->submit<syscall_result<int>>([fd] {
+                    return wrap_syscall<int>(::close(fd));
+                });
+            } catch (...) {
+                report_exception("Running ::close() in reactor thread, submission failed with exception", std::current_exception());
+                return make_ready_future<syscall_result<int>>(wrap_syscall<int>(::close(fd)));
+            }
+        }();
     }
```
raphaelsc, line 331: we should consider switching to std::invoke() later for improved readability and also make @avikivity happy.

raphaelsc (approved):
looks very good. thanks @tchaikov

src/core/file.cc
tchaikov: sure. #1697

PR #1696, opened 2023-06-11
reactor: use s/::free()/::io_uring_free_probe()/

### LINKED ISSUE #1692: free(): invalid pointer abort with liburing 2.4 (works with liburing 2.3!)
After updating liburing to version 2.4, Seastar started failing to startup the io_uring reactor backend.
I did some manual debugging with GDB to figure out that that was the root cause of the issue, and then downgraded liburing to 2.3, and Seastar applications started to work with the io_uring backend again.

This is the specific line where SIGABRT is being sent: https://github.com/scylladb/seastar/blob/b4db95380887f1e744b4fda2edbd33e8dc8071e0/src/core/memory.cc#L998

### System Description:
uname -a: `Linux _hostname_ 6.3.6-arch1-1 '#'1 SMP PREEMPT_DYNAMIC Mon, 05 Jun 2023 15:12:57 +0000 x86_64 GNU/Linux

The issue happened when using any clang version from 15 to 16 and trunk (17).

I also tested building from a few recent and older Seastar commits, and all gave the same issue.

### Steps to reproduce:
Build Seastar from source with liburing 2.4 installed on your system, then try running any Seastar demo apps.

tchaikov:
use ::io_uring_free_probe() instead of ::free() to free the memory chunk pointed by `probe`. io_uring allocates the memory chunk by itself. so there are chances that the precompiled liburing is using the vanilla malloc from the libc, while we free the memory chunk using the overridden version which works with the customized allocator.

in other words, we cannot assume which malloc() was used. in this change, ::io_uring_free_probe() is used, as this function is provided by liburing, so we can ensure that the free() used by it always matches with the malloc() called by io_uring_get_probe_ring().

this should address the issue reported by ASan when the tree is compiled with liburing 2.4, like:

AddressSanitizer: attempting free on address which was not malloc()-ed

Fixes #1692

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -1239,7 +1239,7 @@ try_create_uring(unsigned queue_len, bool throw_on_error) {
         maybe_throw(std::runtime_error("unable to create io_uring probe"));
         return std::nullopt;
     }
-    auto free_probe = defer([&] () noexcept { ::free(probe); });
+    auto free_probe = defer([&] () noexcept { ::io_uring_free_probe(probe); });
 
     for (auto op : required_ops) {
         if (!io_uring_opcode_supported(probe, op)) {


tchaikov:
this change was tested with seastar master HEAD + liburing 2.4 + linux 6.1.0.

PR #1697, opened 2023-06-12
file: call lambda with std::invoke()

tchaikov:
for better readability. as it would be immediately clear to the developer that we are calling this lambda, not defining a variable holding it, as he/she does not have to parse the last token in this expression to get this information.


raphaelsc (approved):
looks good.

PR #1702, opened 2023-06-16
Prefault memory when --lock-memory 1 is specified

avikivity:
To avoid latency spikes due to page faults on anonymous memory,
we mlockall() all memory. This prevents memory from being swapped out,
but we can still see a latency spike when faulting in the memory the first time
it is touched.

It's rare for this to be a problem, as faulting in an anonymous memory page
on first use is cheap - the kernel just has to zero it. However, things are complicated
by transparent hugepages as the kernel first has to defragment memory. This can
take quite a while, as was observed in [1].

The solution is to launch background threads that will attempt to fault-in the
memory ahead of the application. We need to be careful so that these threads
themselves don't compete with the application and cause latency spikes themselves,
so we take the following steps:
1. We launch just one thread per NUMA node, reducing lock contention
2. We place the threads in the SCHED_IDLE class, so the kernel will favor application threads
3. We let the thread affinity float over the entire NUMA node, so it can find the least contended core

Tested on an old Intel E5 E5-2697. It was able to fault 200GB in 50 cpu seconds
(25 wall-clock seconds) for a rate of 8 GB/s. This shows that even a large-memory application
will be able to fault in memory faster than it will need it.

[1] https://github.com/scylladb/scylladb/issues/8828

### DIFF
--- a/include/seastar/core/smp.hh
+++ b/include/seastar/core/smp.hh
@@ -70,6 +70,14 @@ inline shard_id* this_shard_id_ptr() noexcept {
 }
 #endif
 
+class memory_prefaulter;
+
+}
+
+namespace memory::internal {
+
+struct numa_layout;
+
 }
 
 SEASTAR_MODULE_EXPORT_BEGIN
@@ -322,6 +330,7 @@ class smp : public std::enable_shared_from_this<smp> {
     std::vector<posix_thread> _threads;
     std::vector<std::function<void ()>> _thread_loops; // for dpdk
     std::optional<boost::barrier> _all_event_loops_done;
+    std::unique_ptr<internal::memory_prefaulter> _prefaulter;
     struct qs_deleter {
       void operator()(smp_message_queue** qs) const;
     };
@@ -334,8 +343,11 @@ class smp : public std::enable_shared_from_this<smp> {
     using returns_future = is_future<std::invoke_result_t<Func>>;
     template <typename Func>
     using returns_void = std::is_same<std::invoke_result_t<Func>, void>;
+private:
+    void setup_prefaulter(const seastar::resource::resources& res, seastar::memory::internal::numa_layout layout);
 public:
-    explicit smp(alien::instance& alien) : _alien(alien) {}
+    explicit smp(alien::instance& alien);
+    ~smp();
     void configure(const smp_options& smp_opts, const reactor_options& reactor_opts);
     void cleanup() noexcept;
     void cleanup_cpu();
--- a/src/core/resource.cc
+++ b/src/core/resource.cc
@@ -43,6 +43,11 @@ module seastar;
 #include <seastar/util/log.hh>
 #include <seastar/core/io_queue.hh>
 #include "cgroup.hh"
+
+#if SEASTAR_HAVE_HWLOC
+#include <hwloc/glibc-sched.h>
+#endif
+
 #endif
 
 namespace seastar {
@@ -497,6 +502,26 @@ hwloc_topology_t topology_holder::get() {
 
 } // namespace hwloc::internal
 
+static
+std::unordered_map<unsigned, cpuset>
+numa_node_id_to_cpuset(hwloc_topology_t topo) {
+    auto ret = std::unordered_map<unsigned, cpuset>();
+    for (auto numa_node = hwloc_get_next_obj_by_type(topo, HWLOC_OBJ_NUMANODE, NULL);
+            numa_node;
+            numa_node = hwloc_get_next_obj_by_type(topo, HWLOC_OBJ_NUMANODE, numa_node)) {
+        auto parent = numa_node->parent;
+        auto cpuset = parent->cpuset;
+        cpu_set_t os_cpuset;
+        hwloc_cpuset_to_glibc_sched_affinity(topo, cpuset, &os_cpuset, sizeof(os_cpuset));
+        for (unsigned idx = 0; idx < CPU_SETSIZE; ++idx) {
+            if (CPU_ISSET(idx, &os_cpuset)) {
+                ret[numa_node->os_index].insert(idx);
+            }
+        }
+    }
+    return ret;
+}
+
 resources allocate(configuration& c) {
     auto topology = c.topology.get();
     auto bm = hwloc_bitmap_alloc();
@@ -650,6 +675,9 @@ resources allocate(configuration& c) {
     for (auto devid : c.devices) {
         ret.ioq_topology.emplace(devid, allocate_io_queues(topology, ret.cpus, cpu_to_node, c.num_io_groups, last_node_idx));
     }
+
+    ret.numa_node_id_to_cpuset = numa_node_id_to_cpuset(topology);
+
     return ret;
 }
 
--- a/src/core/smp.cc
+++ b/src/core/smp.cc
@@ -29,10 +29,12 @@ module;
 module seastar;
 #else
 #include <seastar/core/smp.hh>
+#include <seastar/core/resource.hh>
 #include <seastar/core/loop.hh>
 #include <seastar/core/semaphore.hh>
 #include <seastar/core/print.hh>
 #include <seastar/core/on_internal_error.hh>
+#include "prefault.hh"
 #endif
 
 namespace seastar {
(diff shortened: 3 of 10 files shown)


avikivity:
@xemul please review

xemul:
@avikivity , on it today (was on vacation last two weeks)

src/core/resource.cc
```diff
@@ -259,6 +264,33 @@ io_queue_topology::io_queue_topology(io_queue_topology&& o)
 
 }
 
+std::optional<std::vector<unsigned>>
+resource::internal::cpus_for_numa_node_id(hwloc::internal::topology_holder& topo_holder, unsigned numa_node_id) {
```
xemul, line 268: The resource::allocate() already calculates the numa topology, it's worth returning it in the `resource::resources`

include/seastar/core/smp.hh
```diff
@@ -334,8 +343,11 @@ class smp : public std::enable_shared_from_this<smp> {
     using returns_future = is_future<std::invoke_result_t<Func>>;
     template <typename Func>
     using returns_void = std::is_same<std::invoke_result_t<Func>, void>;
+private:
+    void setup_prefaulter(seastar::resource::hwloc::internal::topology_holder& topo, seastar::memory::internal::numa_layout layout);
 public:
-    explicit smp(alien::instance& alien) : _alien(alien) {}
+    explicit smp(alien::instance& alien);
+    ~smp();
```
xemul, line 350: nit: It doesn't look like this change belongs to this patch

src/core/smp.cc
```diff
@@ -148,4 +150,89 @@ smp_service_group_semaphore& get_smp_service_groups_semaphore(unsigned ssg_id, s
...
+
+internal::memory_prefaulter::memory_prefaulter(resource::hwloc::internal::topology_holder& topo, memory::internal::numa_layout layout) {
+    for (auto& range : layout.ranges) {
+        _layout_by_node_id[range.numa_node_id].push_back(std::move(range));
+    }
+    auto page_size = getpagesize();
+    for (auto&& [numa_node_id, ranges] : _layout_by_node_id) {
+        posix_thread::attr a;
+        auto cpus_for_numa_node = resource::internal::cpus_for_numa_node_id(topo, numa_node_id);
+        if (cpus_for_numa_node) {
```
xemul, line 176: Does it have rights to work in the "else" case at all?

src/core/smp.cc
```diff
@@ -148,4 +150,89 @@ smp_service_group_semaphore& get_smp_service_groups_semaphore(unsigned ssg_id, s
...
+        // copy eveything into locals, or the optimizer will worry about them due
+        // to the cast below and not optimize anything
+        auto start = range.start;
+        auto end = range.end;
+        for (size_t i = 0; i < batch_size && start < end; ++i) {
+            // atomic_ref would be better, but alas C++20 only
+            auto p = reinterpret_cast<volatile std::atomic<char>*>(start);
+            // Touch the page for write, but be sure not to modify anything
+            p->fetch_or(0, std::memory_order_relaxed);
+            start += page_size;
```
xemul, line 215: What's the point in touching the whole huge-page with page-size steps? If the first (well, zeroth) touch allocates the huge-page, the we're touching the remaining 511 steps in vain. If it doesn't it means that we're touching small pages and delay huge-page split into kernel's background anyway

mykaul: How do we know the compiler wouldn't outsmart us and since we do not actually do anything there, it won't just eliminate this? Something like https://owasp.org/www-community/vulnerabilities/Insecure_Compiler_Optimization ?

src/core/resource.cc
```diff
@@ -259,6 +264,33 @@ io_queue_topology::io_queue_topology(io_queue_topology&& o)
...
+    if (!numa_node) {
+        return std::nullopt; // ??
+    }
+    auto parent = numa_node->parent;
+    auto cpuset = parent->cpuset;
+    cpu_set_t os_cpuset;
+    hwloc_cpuset_to_glibc_sched_affinity(topo, cpuset, &os_cpuset, sizeof(os_cpuset));
+    std::vector<unsigned> ret;
+    for (unsigned idx = 0; idx < CPU_SETSIZE; ++idx) {
+        if (CPU_ISSET(idx, &os_cpuset)) {
```
xemul, line 286: The `cpu_set_t` is converted into `std::vector<int>` here and then back in fa47229a47f6e7564420180c08fa0d9fc5223075. Why not make `cpus_for_numa_node_id` return cpu_set_t right at once? (Or keep it in `resource::resources` as I suggested in above comment

src/core/smp.cc
xemul: It's marked as `volatile` while casting, so it shouldn't

include/seastar/core/smp.hh
avikivity: Probably leftover from try { patch } catch { do something else }

src/core/smp.cc
avikivity: I tried to make it conservative. If it doesn't know what to do, at least don't cause harm. If we don't have a good idea where to run the prefaulter, give up.

src/core/smp.cc
avikivity: > What's the point in touching the whole huge-page with page-size steps? If the first (well, zeroth) touch allocates the huge-page, the we're touching the remaining 511 steps in vain. If it doesn't it means that we're touching small pages and delay huge-page split into kernel's background anyway

The thinking was that we don't know if huge pages work or not, it could be disabled in the kernel.

If huge pages work, then 511 extra atomics are meaningless compared to the page faults (~25 cycle for an atomic compared to at least 4000 cycles for the page fault + lots more cycles for the zeroing + L1/L2 caches destroyed).

If they don't work, then the 511 extra atomics prevent scylladb/scylladb#8828.

btw, I saw patches to use non-temporal stores to zero huge pages, but they didn't make it in.

src/core/resource.cc
avikivity: I think I considered it, but forgot the details. Will look again.

avikivity: Ok, I considered something slightly different - to return the actual NUMA layout. But this can't be done until memory::configure() is run and can associate addresses.

I could return cpus_to_numa_node_id in resources, but I don't think that's its role. hwloc is responsible for answering random questions about topology, and since we want to abstract it, we put the helpers in resources.cc. Should resource::resources be enlarged each time we have a new query for topology?

xemul: > Ok, I considered something slightly different - to return the actual NUMA layout. But this can't be done until memory::configure() is run and can associate addresses.

But you don't need addresses association here, only cpu:numa_node map

> I could return cpus_to_numa_node_id in resources, but I don't think that's its role. ... Should resource::resources be enlarged each time we have a new query for topology?

There are two options here. First -- make resources.cc be a bunch of transparent wrappers over hwloc so that we could ask it anything about topology, it would consult with hwloc behind the scenes, then return back some internal (btw, resources is not in internal) data type. Second -- make resources::allocate() talk to hwloc and return back anything that reactor would ever need to start. So far the second approach is in use.

avikivity: I refreshed my memory. I don't think this fits resources::allocate(). resources::allocate() says "look at system resources and tell me how they should be allocated to shards", not "look at system resources and tell me how they should be allocated to shards plus some random information".

src/core/resource.cc
avikivity: There is also a resources::cpuset data type.

src/core/resource.cc
xemul: Then io-topology allocation deserves being removed from this method either (definitely not in this PR, unless you feel enough courage)

avikivity: I guess it's better to avoid churn, and define allocate() as "look at system resources and tell me how they should be allocated to shards, and give me any supporting topology information that can help me later".

In the end it's an internal API, even though we forgot to put it in the internal namespace.

include/seastar/core/smp.hh
avikivity: It's necessary since memory_prefaulter is forward declared, so its destructor is not available here. I'll enhance the changelog.

avikivity:
(retesting)

avikivity:
I checked on i4i.metal to see numa binding works:

```
     9,188,146,520      mem-stores                                                              (28.64%)
                 0      mem-loads                                                               (42.92%)
   529,639,148,735      cycles                                                                  (57.21%)
        70,076,138      node-loads                                                              (57.21%)
     2,167,152,006      node-stores                                                             (71.51%)
        18,071,442      ocr.itom.l3_miss_local                                                  (28.55%)
                77      ocr.itom.remote                                                         (14.28%)
```

"itom" == "initialize to memory" (I presume, write an entire cache line without reading it first) only shows local.

avikivity:
v2:
 - retested
 - dropped unnecessary semicolon
 - explained why smp's destructor had to be deinlined
 - changed method of conveying NUMA binding from separate helper to a field in resource::resources

raphaelsc:
> (retesting)

https://godbolt.org/z/G93dhsdMa

am I missing anything or clang is incorrectly optimizing out the code even though atomic was volatile qualified?

avikivity:
v3: dropped use of unordered_map::contains which isn't available on all toolchains used

raphaelsc:
> v3: dropped use of unordered_map::contains which isn't available on all toolchains used



> > (retesting)
> 
> https://godbolt.org/z/G93dhsdMa
> 
> am I missing anything or clang is incorrectly optimizing out the code even though atomic was volatile qualified?

seastar uses gcc by default IIRC, which is probably why you didn't bump into it during development

raphaelsc:
> Looks like clang is dropping the optimization.

I tried sprinkling volatile keyword all over, but mr Clang wants to optimize out that no matter what. Perhaps an attribute which can force that function to -O0?

avikivity:
Good catch, I was testing with gcc only.

avikivity:
v4: switched to inline asm (still need to test aaaarch64)

src/core/smp.cc
```diff
@@ -148,4 +150,100 @@ smp_service_group_semaphore& get_smp_service_groups_semaphore(unsigned ssg_id, s
...
+#elif defined(__aarch64__)
+        int byte; // ldxrb likes 32-bit registers
+        int need_loop;
+        asm volatile ("1: ldxrb %w0, %2; stxrb %w1, %w0, %2; cbnz %w1, 1b"
+                : "=&r"(byte), "=&r"(need_loop), "+Q"(*p));
+#else
+        // atomic_ref would be better, but alas C++20 only
+        auto p1 = reinterpret_cast<volatile std::atomic<char>*>(p);
+        p->fetch_or(0, std::memory_order_relaxed);
+#endif
```
raphaelsc, line 218: nice

avikivity:
v5:
 - tested on aarch64 (no change)
 - drop atomic_flag, test() method is C++20 only; use atomic\<bool> instead.

avikivity:
v6: (sigh)
 - use consistent struct/class tag to shut up stupid clang error

raphaelsc:
> Good catch, I was testing with gcc only.

Filed an issue to LLVM: https://github.com/llvm/llvm-project/issues/63692

avikivity:
> > Good catch, I was testing with gcc only.
> 
> Filed an issue to LLVM: [llvm/llvm-project#63692](https://github.com/llvm/llvm-project/issues/63692)

Likely to get rejected.

avikivity:
v8:
 - fix for non-hwloc environments (again)

avikivity:
v9:
 - eliminate capturing structured binding which some clangs dislike
 - eliminate unused member in memory_prefauter
 - eliminate unused lambda capture

avikivity:
v10:
 - only start if --lock-memory is specified
 - fix recent use-after-stack-deallocate regression
 - retested locally with all CI variations
 - fixed reactor backend debug-mode bug https://github.com/scylladb/seastar/pull/1721
 - fixed build failure generating openssl certificates https://github.com/scylladb/seastar/pull/1725

raphaelsc:
> > > Good catch, I was testing with gcc only.
> > 
> > 
> > Filed an issue to LLVM: [llvm/llvm-project#63692](https://github.com/llvm/llvm-project/issues/63692)
> 
> Likely to get rejected.

https://reviews.llvm.org/D154609

avikivity:
> > > > Good catch, I was testing with gcc only.
> > > 
> > > 
> > > Filed an issue to LLVM: [llvm/llvm-project#63692](https://github.com/llvm/llvm-project/issues/63692)
> > 
> > 
> > Likely to get rejected.
> 
> https://reviews.llvm.org/D154609

Happy to be wrong! Please see if you can get it backported, too.

michoecho:
> > > > Good catch, I was testing with gcc only.
> > > 
> > > 
> > > Filed an issue to LLVM: [llvm/llvm-project#63692](https://github.com/llvm/llvm-project/issues/63692)
> > 
> > 
> > Likely to get rejected.
> 
> https://reviews.llvm.org/D154609

Nice.

So they introduced the bug by adding a special case optimization for `fetch_or(0)`. I wonder what software benefits from that optimization.

src/core/resource.cc
```diff
@@ -650,6 +675,9 @@ resources allocate(configuration& c) {
     for (auto devid : c.devices) {
         ret.ioq_topology.emplace(devid, allocate_io_queues(topology, ret.cpus, cpu_to_node, c.num_io_groups, last_node_idx));
     }
+
+    ret.numa_node_id_to_cpuset = numa_node_id_to_cpuset(topology);
```
xemul, line 679: The `std::unordered_map<unsigned, hwloc_obj_t> cpu_to_node` in `resource::allocate()` already contains `cpu -> numa_node` map, so the `numa_node_id_to_cpuset` can be evaluated by turning the `cpu_to_node` inside out. Is the newly introduced `numa_node_id_to_cpuset()` produces some different result?

avikivity: They aren't the same. For example, with --smp 1 cpu_to_node will be a map (cpu)0 -> (node)0, amd numa_node_id_to_cpuset() will return a map (node)0 -> (cpu)0..15. (node)1 -> (cpu)16..31 (say).

If we use the first for the prefault thread, it will bind the prefault thread to cpu 0 (worst place since it's a reactor thread), while with the second we bind to 0..15.

PR #1703, opened 2023-06-17
tests: fail spawn_test if output is empty

tchaikov:
before this change, test_spawn_echo passes even if the output is empty, as `_matched` is set as long as the captured output match with `_expected`, but if an empty output is returned `_matched` is also assigned to `true`, and the coroutine of `stdout.consume()` returns. this leaves us with a `true` `matched` even if the "hello world" is not fully matched.

so, in this change, `_matched` is set only if the the whole expected string is matched or any mismatch is identified.


tchaikov:
@scylladb/seastar-maint hi maintainers, could you help review this change?

PR #1706, opened 2023-06-19
reactor: change shares for default IO class from 1 to 200

michoecho:
The number of shares is supposed to be set in the 1-1000 range, and user-defined IO classes have shares on the order of hundreds in practice.

Therefore, 1 share is not a sane default, because the default class can very easily be starved by accident.
We have seen in practice for example in scylladb/scylladb#13753.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1716,7 +1716,7 @@ const io_priority_class& default_priority_class() {
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
     static thread_local auto shard_default_class = [] {
-        return io_priority_class::register_one("default", 1);
+        return io_priority_class::register_one("default", 200);
     }();
     return shard_default_class;
 #pragma GCC diagnostic pop


mykaul:
There's no 'backport' label here, but I assume our intention is to pick this up for stable versions.

michoecho:
@avikivity I'm not sure if we want this change in seastar master (this could be a backwards-incompatible change for some users), or only in Scylla's stable forks of Seastar. What's your judgement?

michoecho:
> There's no 'backport' label here, but I assume our intention is to pick this up for stable versions.

Yes.

dorlaor:
Is it a regression or shares==1 has always been there?

michoecho:
> Is it a regression or shares==1 has always been there?

Since 97f418a4f2c146a2ac9d56cc673aea9a13541c4c (so, since forever).

michoecho:
@avikivity @xemul Bump. Do we want this change in seastar master (this could be a backwards-incompatible change for some users), or only in `scylla-seastar`?

avikivity:
This code is obsolete in API level 7. So let's go for minimum disruption and only apply the change to ScyllaDB.

michoecho:
> This code is obsolete in API level 7. So let's go for minimum disruption and only apply the change to ScyllaDB.

Then please create branch-5.3 in scylla-seastar, so we can backport seastar fixes to 5.3.

avikivity:
Done.

michoecho:
Transferred to https://github.com/scylladb/scylla-seastar/pull/3.

PR #1583, opened 2023-03-28
[RFC] reactor: optionally print decoded stack traces on abort

michoecho:
Decoding stack traces requires access to all binaries appearing in the backtrace, and a manual invocation of seastar-addr2line on the backtrace cop-pasted from logs. This is very inconvenient, especially since seastar became a shared library in debug mode, which makes the main executable alone insufficient to decode a backtrace.

In Scylla, we have backtrace.scylladb.com to automate this, but it is only applicable to public builds, not to local or CI builds.

This series adds an option, implemented using libbacktrace, which prints a decoded backtrace in addition to the regular backtrace.
It is only used on aborts, because the decoding can cause huge stalls.

PR #1707, opened 2023-06-21
util: print_safe: fix hex print functions

tchaikov:
print_zero_padded_hex_safe was never supposed to be for signed integers. The concept was accidentally changed from std::integral to std::signed_integral instead of std::unsigned_integral in a recent refactor in 274f22a8c53c472feb1d110409a80d66df15b101. Fix that.

convert_zero_padded_hex_safe is using a non-existent one-argument variant of convert_hex_safe. Fix that.


tchaikov:
@michoecho hi Michał, i took the liberty to extract this change from your #1583 . as i think this is a bug that we need to address. and this change is less controversial. so it has better chance to be accepted.

michoecho (approved):

PR #1708, opened 2023-06-21
websocket: simplify payload length and mask parsing.

voutilad:
Re-use the byteorder functions instead of pointer derefs. I found that while building something on an arm64 system, for various reasons the pointer derefs were happening on non-aligned addresses causing runtime errors from the address sanitizer when using a Debug build. This simplifies the code a bit.


tchaikov (approved):
lgtm

voutilad:
@avikivity any interest?

PR #1710, opened 2023-06-21
Prevent under-reads and data loss of websocket payloads.

voutilad:
The current websocket payload handling has the potential to produce misaligned reads, corrupting future results, as well as under-reads with lost data.

In the current logic, a presented buffer (`data`) that passes through the header parsing but doesn't contain enough remaining capacity to extract the full, expected payload, will produce an under read and the websocket handler's `input_stream<char>` will receive a truncated payload.

In the case enough data _is_ present, any additional data that may have been over-read will be truncated and lost. This will corrupt subsequent reads for websocket frames from the stream, breaking the parser.

I'm not too in love with using a growing `seastar::sstring` as a buffer like in the header parsing, but this is the only way I found to make sure data is not lost. I'm open to alternatives to reduce copying because I'm not too familiar with the consumer abstraction since I've only started hacking on Seastar 😄 .

The benefit to this approach is that on the consumer side, i.e. in the websocket handler function, reads of the `input_stream<char>` will be presented as whole payloads...assuming continuation frames aren't used since they don't appear to be supported yet.


xemul:
@voutilad , can you cook an artificial unit test that could emulate this bug?

voutilad:
Looks like I need to change my test code to fix those CI errors.

voutilad:
@xemul Just nudging you.

xemul:
LGTM, but have some cosmetic comments

1. Patches 3, 4 and 5 all patch the code that was introduced in patch 1. That's excessive, you can just squash them all
2. Patch 2 uses tabs as indentation. Seastar coding-style assumes that you use 4-spaces as single indent step. This is somewhat critical, because mixed tabs and spaces produces unreadable result in the screen

voutilad:
@xemul:
> 1. Patches 3, 4 and 5 all patch the code that was introduced in patch 1. That's excessive, you can just squash them all

Squashed. Now there are 2 commits: the addition of the unit test and then the addition of the changes to resolve the issue.

> 2. Patch 2 uses tabs as indentation. Seastar coding-style assumes that you use 4-spaces as single indent step. This is somewhat critical, because mixed tabs and spaces produces unreadable result in the screen

Sorry about that. Should be better now.

Tests are mostly failing (all clang builds?) though as they seem to be broken on `master` at the moment.

src/websocket/server.cc
```diff
@@ -254,18 +256,23 @@ future<websocket_parser::consumption_result_t> websocket_parser::operator()(
...
-            _result = std::move(data);
-            return websocket_parser::stop(buff_t(0));
+        if (_buffer.length() + data.size() >= _payload_length) {
+            if (_buffer.length() < _payload_length) {
+                size_t blen = _buffer.length();
+                _buffer.append(data.get(), _payload_length - blen);
+                data.trim_front(_payload_length - blen);
+
+                // We've completed a payload, so reset state and set result.
+                _result = buff_t{_buffer.data(), _buffer.length()};
```
tchaikov, line 266: i'd suggest
```c++
_result = buff_t{_payload_length};
auto result = std::copy(_buffer.cbegin(), _buffer.cend(), _result.get_write());
_buffer.trim(0);

size_t to_consume = _payload_length - blen;
std::copy_n(data.get(), , to_consume, result);
data.trim_front(_payload_length - blen);

// We've completed a payload, so reset state and set result.
remove_mask(_result, _result.size());
_payload_length = 0;
_state = parsing_state::flags_and_payload_data;
return websocket_parser::stop(std::move(data));
```

for couple reasons:

1. avoid copying the payload twice
2. do not free `_buffer`. we can just reuse it for the next payload
3. name `_payload_length - blen` for better readability

src/websocket/server.cc
```diff
@@ -58,7 +58,9 @@ opcodes websocket_parser::opcode() const {
 }
 
 websocket_parser::buff_t websocket_parser::result() {
-    return std::move(_result);
+    auto r = std::move(_result);
+    _result = buff_t{};
+    return r;
```
tchaikov, line 63: might want to
```c++
return std::exchange(_result, {});
```

simpler this way.

p12tic:
@voutilad I think that this PR can be closed. 

The fixes have been independently landed in https://github.com/scylladb/seastar/pull/2535. Unfortunately I didn't think to look into the PR list before spending time fixing issues myself.

I used this PR for test improvement ideas when working on this refactoring https://github.com/scylladb/seastar/pull/2562. So this PR was useful in the end, even though it wasn't merged directly. Thanks a lot!

PR #1715, opened 2023-07-02
shared_ptr: deprecate lw_shared_ptr operator=(T&&)

### LINKED ISSUE #1621: Proposal to remove `shared_ptr::operator=(T&&)`
This overload was added 9 years ago in https://github.com/scylladb/seastar/commit/319ae0b530107dce510c27f45d21b4ad696ef07d.

```
    shared_ptr& operator=(T&& x) {
        this->~shared_ptr();
        new (this) shared_ptr(new data(std::move(x)));
        return *this;
    }
```

This recently burned me when I had mistakenly typed `p = std::string();` but what I meant was `*p = std::string()` expecting holders of copies of `p` to see the update. FWIW std::shared_ptr won't allow such mistakes to compile.

I guess I'm interested in there is some history here for this, and if it should be removed.

tchaikov:
unlike std::shared_ptr, seastar::shared_ptr allows us to move a
value into a shared_ptr with operator=(). the corresponding change in
seastar was introduced by
https://github.com/scylladb/seastar/commit/319ae0b530107dce510c27f45d21b4ad696ef07d. despite that this helper is convenient, it is confusing. as the behavior of a shared_ptr should look like a pointer instead the value pointed by  it. and this could be error-prune, because a developer could by accident write something like:
```c++
p = std::string();
```
and expect that the value pointed by `p` is cleared. and all copies of this shared_ptr are updated accordingly. what he/she really wants is:
```c++
*p = std::string();
```
but the code compiles. while the outcome of the statement is that the destructor `p` is called, and `p` now points to a new instance of string with a new address. the copies of this instance of shared_ptr still hold the old value.

apparently, this behavior is not expected.

so let's deprecate this operator and remove it in future.

Fixes #1621

### DIFF
--- a/include/seastar/core/shared_ptr.hh
+++ b/include/seastar/core/shared_ptr.hh
@@ -343,6 +343,7 @@ public:
     lw_shared_ptr& operator=(std::nullptr_t) noexcept {
         return *this = lw_shared_ptr();
     }
+    [[deprecated("call make_lw_shared<> and assign the result instead")]]
     lw_shared_ptr& operator=(T&& x) noexcept {
         this->~lw_shared_ptr();
         new (this) lw_shared_ptr(make_lw_shared<T>(std::move(x)));


tchaikov:
@tgrabiec hi Tomasz, could you help review this change?

nyh:
@tchaikov your example is convincing - we don't want to encourage people to write code that is confusing to read and accepts typos, and the argument that std::shared_ptr doesn't allow this is also a good one.

@tgrabiec can you please try to remember why you wanted to allow this `p = std::string()` to be shorthand for `p = make_shared<std::string>(std::string())`? Was it just less typing, or something else?

tchaikov:
v2:

- correct the commit message: the pointee of `p` is not necessarily destructed. its `p` 's destructor gets called.

include/seastar/core/shared_ptr.hh
```diff
@@ -343,6 +343,7 @@ public:
     lw_shared_ptr& operator=(std::nullptr_t) noexcept {
         return *this = lw_shared_ptr();
     }
+    [[deprecated("Move into the pointee instead")]]
```
avikivity, line 346: Moving into the pointee has a different effect - all other sharers will see the new value.

The real replacement is to call make_lw_shared\<> and assign the result.

tchaikov:
v2

- incorporate the suggestion from Avi.

tgrabiec:
> @tgrabiec can you please try to remember why you wanted to allow this `p = std::string()` to be shorthand for `p = make_shared<std::string>(std::string())`? Was it just less typing, or something else?

I don't remember.

bhalevy:
What about the equivalent operator for shared_ptr?

```
    template <typename U, typename = std::enable_if_t<std::is_base_of<T, U>::value>>
    shared_ptr& operator=(shared_ptr<U>&& x) noexcept {
```

tchaikov:
it assigns a pointer to another pointer by moving the former. so i think it should be fine. the problem i was trying to address in this PR was assigning a value to a pointer.

bhalevy:
> it assigns a pointer to another pointer by moving the former. so i think it should be fine. the problem i was trying to address in this PR was assigning a variable of type `T` to a pointer pointing to `T`.

makes sense, thanks.

PR #1718, opened 2023-07-03
http/client: Add "total new connections" metrics

xemul:
The metrics is the monotonic counter that gets ++ed every time the factory is kicked for new connection.

The recommended usage is to get the time-derivative from the counter and check _this_ value. If it's large, it means that the existing connections are not getting back to pool due to connection errors or unexpected server status codes. This, in turn, can be used to explain large delays in requests, as each time a request is run over a new connection it doesn't benefit from connections keep-alive and gets extra latency.

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -168,6 +168,7 @@ class client {
     std::unique_ptr<connection_factory> _new_connections;
     unsigned _nr_connections = 0;
     unsigned _max_connections;
+    unsigned long _total_new_connections = 0;
     condition_variable _wait_con;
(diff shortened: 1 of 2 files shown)


xemul:
upd:
- rebased
- ++ the counter before connect to account for failures too

xemul:
@nyh , please review

xemul:
@nyh , review ping

nyh (approved):

nyh:
Strange, why didn't the CI run on this PR?
Anyway, the patch is simple enough that I'll just compile-check and merge.

Issue #1641, opened 2023-05-02
IO sched native throttler is probably too throttly


xemul:
There are three places where dispatch loop is throttled

1. self-throttling with token bucket according to math model
2. per-shard one-tick threshold
3. 2-bucket approach when tokens are replenished only after they are released from disk

Experiments and observations show that in many cases IO latency becomes _better_ if configuring the scheduler with higher limits thus pushing the limits for all three cases up. This means that current dispatch queue throttling is too tight and can be relaxed, but need to understand what's wrong now

xemul:
It's "capped release" feature of the token bucket that can slow down the token bucket dispatch rate heavily.
Found with the latest version of #1724:

Created token-bucket 1000000.0 t/s
perf_pure_context.sleeping_throughput_with_hog: 966646.1 t/s
perf_capped_context.sleeping_throughput: 838035.2 t/s
perf_pure_context.sleeping_throughput_with_hog: 317685.3 t/s

xemul:
From the test (#1724) -- the tokens releasing (ceil rover) is constantly "behind" even the grabbed tokens (tail rover) although it grows at the same rate. This slows down the grabbing loop, thus pushing the released rover even more behind until it finds its balance at ~50% of the configured speed :(

In simple words -- the ceil feedback works poorly, far from what it was expected

xemul:
Advancing the ceil in the attempt to compensate for that doesn't help :(

Here's the image with X-axis is the "compensation" -- ceil is forwarded this amount of tokens on start. Y-axis is the resulting rate of the TB. Pink line is no-capped t.b., blue line -- capped. Values of compensation of 100...1000 are chosen "proportional" to the t.b. limit, which is 500 in this test. So even if compensation is 2x-times the limit the resulting rate is still not 1M as it should be

![image](https://github.com/scylladb/seastar/assets/4498177/57c40b6c-eafe-4747-b259-185bf1df2a21)

xemul:
Math behind the slowdown.

Let's assume that a some point scheduler dispatched `N_d` requests. It means that it was able to grab `N_d` tokens in `T_d` duration, the rate of dispatch is `R_d = N_d/T_d`. The requests are to be completed by the reactor next tick. Let's assume it takes `T_c` time until reactor gets there and it completes `N_c` requests. The rate of completion is thus `R_c = N_c/T_c`. Apparently, `N_c <= N_d`, because kernel cannot complete more requests that it was queued.

[...]

PR #1722, opened 2023-07-06
shared_token_bucket: Fix duration_for() underflow

xemul:
This helper calculates the duration that it takes to replenish the specified amount of tokens. When the number of provided tokens is less than than the configured rate the integer division gives zero result which immediatelly converts to empty duration. That's not correct, the duration type is double and can be calculated with more precision.

Typically users of token bucket call this method when they see a deficiency after grabbing a token. In that case they sleep or arm a timer for the provided duration and then call replenish() on the bucket. If the deficiency is small and the calculated duration is zero so the caller doesn't sleep. This leads to unneeded busy-looping around the token bucket.

Another consequence of not sleeping on small deficiency is that replenish() converts time delta to tokens potentially losing precision (delta is floating point, tokens are integers). The conversion rounds the fractional tokens in a hope that rounding happens in both directions and compensates each other on average. This might not be the case when caller doesn't sleep and calls replenish() with tiny durations in between -- rounding goes upper-bound more often than lower-bound thus resulting in higher rate than needed.

refs: #1641

### DIFF
--- a/include/seastar/util/shared_token_bucket.hh
+++ b/include/seastar/util/shared_token_bucket.hh
@@ -200,7 +200,7 @@ public:
     // Estimated time to process the given amount of tokens
     // (peer of accumulated_in helper)
     rate_resolution duration_for(T tokens) const noexcept {
-        return rate_resolution(tokens / _replenish_rate);
+        return rate_resolution(double(tokens) / _replenish_rate);
     }
 
     T rate() const noexcept { return _replenish_rate; }


xemul:
@avikivity , please review

xemul:
@avikivity , please review

include/seastar/util/shared_token_bucket.hh
```diff
@@ -200,7 +200,7 @@ public:
     // Estimated time to process the given amount of tokens
     // (peer of accumulated_in helper)
     rate_resolution duration_for(T tokens) const noexcept {
-        return rate_resolution(tokens / _replenish_rate);
+        return rate_resolution(double(tokens) / _replenish_rate);
```
tchaikov, line 203: how about:

```suggestion
        return rate_resolution{tokens} / _replenish_rate;
```

so we don't need to hardwire to double after defining `rate_resolution`, which intends to abstract the representation of the rate resolution.

avikivity: I'd say that _replenish_rate needs to be a double (and then maybe we can make Period = 1).

I'll merge this for now as a minimal fix and we can improve it later.

Issue #1083, opened 2022-06-01
IO capacity balancing is not well balanced


xemul:
As seen on i3.large node (2 cores) in scylladb/scylla#10704

vladzcloudius:
@tomer-sandler FYI

tomer-sandler:
@DoronArazii please help push this FW, this issue is effecting Disney, but may very well effect many others upon upgrade to 2022.x.y.
As I mentioned in our call today ~85% of our customers are still NOT running 2022.x.y.
The OSS [issue](https://github.com/scylladb/scylladb/issues/12562) @xemul mentioned is targeted for OSS 5.3, but Ent customers need it much sooner.

avikivity:
(deleted comment mentioning customers)

dorlaor:
How complex is to fix this?

xemul:
I've a patch that has two problems
- it's not confirmed on anything but a io-tester-based reproducer
- it has a "configuration parameter" that should be somehow configured by hand and there's no good ideas (yet) how to select it automatically

mykaul:
What's the latest status of this issue? (seeing if it'll make it to Scylla 5.4)

xemul:
It's in lower prio, because there's a "workaround" -- one need to configure more io-groups than it's auto-detected by seastar to make groups' size smaller and thus reduce the per-group imbalance. Avi thinks it should be the default behavior.

bhalevy:
> It's in lower prio, because there's a "workaround" -- one need to configure more io-groups than it's auto-detected by seastar to make groups' size smaller and thus reduce the per-group imbalance. Avi thinks it should be the default behavior.

@xemul so are we making this the default behavior?

xemul:
> @xemul so are we making this the default behavior?

I lean towards it, but I've no good ideas how to calculate/estimate which amount of shards in a group is good enough

PR #1724, opened 2023-07-06
tests: Add perf test for shard_token_bucket

xemul:
The test checks if the token-bucket "rate" is held under various circumstances:

- when shards sleep between grabbing tokens
- when shards poll the t.b. frequently
- when shards are disturbed with CPU hogs

So far the test shows four problems:

- With few shards tokens deficiency produces zero sleep time, so the "good" user that sleeps between grabs effectively converts into a polling ("bad") user (fixed by #1722)

- Sometimes replenishing rounding errors accumulate and render lower resulting rate than configured (fixed by #1723)

- When run with CPU hogs the individual shard's rates may differ too much (see #1083). E.g. the bucket configured with the rate of 100k tokens/sec, 48 shards, run 4 seconds.

  "Slowest" shard vs "fastest" shards get this amount of tokens:

    no hog:   6931 ... 9631
    with hog: 2135 ... 29412

  (sum rate is 100k with the aforementioned fixes)

- With "capped-release" token bucket and token releasing by-timer with
  the configured rate and hogs the resulting throughput can be as low as
  30% of the configured (see https://github.com/scylladb/seastar/issues/1641)

  Created token-bucket 1000000.0 t/s
  perf_pure_context.sleeping_throughput_with_hog:   966646.1 t/s
  perf_capped_context.sleeping_throughput:          838035.2 t/s
  perf_capped_context.sleeping_throughput_with_hog:   317685.3 t/s

### DIFF
--- a/tests/perf/shared_token_bucket.cc
+++ b/tests/perf/shared_token_bucket.cc
@@ -1,0 +1,363 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2023 ScyllaDB Ltd.
+ */
+
+
+#include <random>
+#include <seastar/testing/perf_tests.hh>
+#include <seastar/testing/random.hh>
+#include <seastar/core/sharded.hh>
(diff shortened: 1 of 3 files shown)


xemul:
@avikivity , please review

xemul:
upd:
- added capped-release test that finally shows a real problem

xemul:
upd:
- removed some unused stats
- less hard-coded constants
- relaxed hogs not to burn CPU throughout the whole task-quota
- improved releaser to work like real reap_kernel_completion -- it releases tokens proportional to real pause duration

tests/perf/shared_token_bucket.cc
```diff
@@ -0,0 +1,295 @@
...
+        });
+    }
+};
+
+struct perf_capped_context : public context<capped_token_bucket_t> {};
+struct perf_pure_context : public context<pure_token_bucket_t> {};
+
+PERF_TEST_F(perf_capped_context, sleeping_throughput) { return test_sleeping(); }
+PERF_TEST_F(perf_capped_context, yielding_throughput) { return test_yielding(); }
+PERF_TEST_F(perf_capped_context, sleeping_throughput_with_hog) { return test_sleeping_with_hog(); }
```
avikivity, line 291: Please add lots of comments explaining the test scenario and documenting what a good result is (and why). I'm sufficiently detached from the internals that it's hard for me to figure it out directly.

xemul: Yup, my bad. Added

PR #1725, opened 2023-07-06
build: serialize calls to openssl certificate generation

avikivity:
During build, we generate a few SSL certificates for testing. The generation process as a .srl file containing a serial number, which is both read and written. It appears openssl doesn't bother to lock it, so if ninja happens to run two such processes concurrently, one of them can fail when reading a partially written file.

Fix this by placing these build steps into a job pool with a concurrency of one.

Also add the implicit dependency into the command declarations this doesn't help much, since the dependency is an input/output file which can't be expressed by the build system, but helps make the file visible.

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -526,6 +526,9 @@ function(seastar_add_certgen name)
 
   find_program(OPENSSL openssl)
 
+  # Some openssl commands read and write the serial number file (.srl). These can't be executed in parallel.
+  set_property(GLOBAL PROPERTY JOB_POOLS openssl_serial=1)
+
   add_custom_command(OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${CERT_PRIVKEY}"
     COMMAND ${OPENSSL} genpkey -quiet -out ${CERT_PRIVKEY} -algorithm ${CERT_ALG} ${CERT_ALG_OPTS}
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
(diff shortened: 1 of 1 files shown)


nyh:
Your patch is correct and I will commit it, but it's really disappointing to see, yet again, how something which is trivial to do in a shell script (just run 3 commands one after another) becomes a 30-line mess complete with concurrency bugs.

nyh:
> Your patch is correct and I will commit it, but it's really disappointing to see, yet again, how something which is trivial to do in a shell script (just run 3 commands one after another) becomes a 30-line mess complete with concurrency bugs.

I forgot to put the blame where the blame is due - it becomes **in cmake** a 30-line mess. We've been using this tool for a few years now, and I still can't for the life of me think of anything positive to say about it. Sad.

avikivity:
I dislike cmake as much as the next person, and possible also the previous person, but here you're incorrect. It's easy to write a three line shell script to do three things one after the other (modulo error handling and whitespace treatment), but build systems use a declarative system to allow for concurrent builds. It takes more effort but rewards people with 16-core machines.

nyh:
I think my rant didn't come across the way I meant it. I was referring to **make**, which recently celebrated its 47th birthday. It (not the shell itself) had the declarative system for doing concurrent builds, but if you need to run three commands in sequence because openssl can't tolerate concurrency, you just write a 3-line command and it runs in a subshell, sequentially. You don't need 30 lines to say how to run these commands concurrently in some incomprehensible language, and then a few more lines to say "but please don't run it concurrently".

tchaikov (commented):
lgtm. just wanted to note job_pool is a ninja only feature. probably most of us are using ninja when building Seastar, but still..

avikivity:
> lgtm. just wanted to note job_pool is a ninja only feature. probably most of us are using ninja when building Seastar, but still..

That's sad. Maybe it's possible to use -set_serial and drop -CAcreateserial to avoid the whole thing.

avikivity:
> I think my rant didn't come across the way I meant it. I was referring to **make**, which recently celebrated its 47th birthday. It (not the shell itself) had the declarative system for doing concurrent builds, but if you need to run three commands in sequence because openssl can't tolerate concurrency, you just write a 3-line command and it runs in a subshell, sequentially. You don't need 30 lines to say how to run these commands concurrently in some incomprehensible language, and then a few more lines to say "but please don't run it concurrently".

I hate to be defending cmake, but you could do the same in cmake. Instead of separate targets for each certificate, create just one target that generates all the certificates.

If anything, make hasn't managed to properly support multi-output targets despite having 47 years to do so.

PR #1727, opened 2023-07-08
treewide: use fmtlib's fmt::format_string in format() 

tchaikov:
this series enables `seastar::format()` to use compile-time format check.

using `fmt::format_string` would allow us to perform compile-time
format checking. so in this change, we switch to `fmt::format_string`
when fmtlib's version is greater or equal to 8.0, which introduced
the compile-time checking.

also, because the `fmt::format_string` can only be instantiated with
a compile-time constexpr, the fmt string specified in test is changed
to a constexpr accordingly.

please note, this is a breaking change which might require caller to use
`seastar::format()` if the tree is compiled with libstdc++ 13 or a standard
library with `std::formt()` implemented.

### DIFF
--- a/include/seastar/core/print.hh
+++ b/include/seastar/core/print.hh
@@ -140,9 +140,9 @@ log(A&&... a) {
  */
 template <typename... A>
 sstring
-format(const char* fmt, A&&... a) {
+format(fmt::format_string<A...> fmt, A&&... a) {
     fmt::memory_buffer out;
-    fmt::format_to(fmt::appender(out), fmt::runtime(fmt), std::forward<A>(a)...);
+    fmt::format_to(fmt::appender(out), fmt, std::forward<A>(a)...);
     return sstring{out.data(), out.size()};
 }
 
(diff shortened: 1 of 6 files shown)


tchaikov:
i am still working on getting scylladb compiled with this change.

tchaikov:
i think i hit a bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=110602

tchaikov:

if we have concerns on the backward compatibility like we did in 228bf07db98f34357c5c0d552f7abc610986f180, i will add a macro which plays a similar role of `SEASTAR_LOGGER_COMPILE_TIME_FMT`, so that the existing Seastar application can opt out of this compile-time formatter. as, because the way how ADT works, applications will have to explicitly call either `std::format()` or `seastar::format()`. see also https://github.com/scylladb/seastar/pull/1727#issuecomment-1627588115

avikivity:
Most reasonable applications don't use `using namespace seastar`.

include/seastar/core/print.hh
```diff
@@ -140,9 +140,9 @@ log(A&&... a) {
  */
 template <typename... A>
 sstring
-format(const char* fmt, A&&... a) {
+format(fmt::format_string<A...> fmt, A&&... a) {
```
avikivity, line 143: Do we need another overload for fmt::runtime, or will this work out of the box?

avikivity: Ah, runtime_format_string is convertible to basic_format_string.

PR #1734, opened 2023-07-13
build, net: disable implicit fallthough

tchaikov:
this series disables implicit fallthough and add [[fallthough]] when appropriate

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -765,6 +765,7 @@ set (Seastar_PRIVATE_CXX_FLAGS
   -UNDEBUG
   -Wall
   -Werror
+  -Wimplicit-fallthrough
   -Wno-array-bounds # Disabled because of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93437
   -Wno-error=deprecated-declarations)
 if (NOT BUILD_SHARED_LIBS)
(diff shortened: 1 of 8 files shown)


tchaikov:
in the same spirit of https://github.com/scylladb/scylladb/pull/14607

tchaikov:
this changeset includes #1735 to address the build failure. despite that #1735 is not relevant to this changeset.

cmake/CheckHeaders.cmake
```diff
@@ -84,6 +84,11 @@ function (seastar_check_self_contained library)
   list (REMOVE_DUPLICATES includes)
   target_include_directories (${check_lib}
     PRIVATE ${includes})
+  get_target_property (system_includes ${library} INTERFACE_SYSTEM_INCLUDE_DIRECTORIES)
+  if (system_includes)
+    target_include_directories (${check_lib} SYSTEM
+      PRIVATE ${system_includes})
```
avikivity, line 90: Maybe we can disable those warnings with #pragma instead, to avoid silencing all warnings.

tchaikov: yeah, even better.

tchaikov:
v2

- use `#pragma` instead of using `-isystem <dir>` compiler command line option to silence warnings

@avikivity changed as suggested. could you take another look?

tchaikov:
@avikivity ping again?

tchaikov:
@avikivity hi Avi, yet another ping?

PR #1735, opened 2023-07-13
circular_buffer_fixed_capacity_test: enable randomize test to reverse

tchaikov:
before this change, the randomized test accepts a RNG which generates numbers in the range of [0, 6], and translate the generated numbers to corresponding operations, then apply them to the `std::deque` and `circular_buffer_fixed_capacity<>` under test.

it seems that we intent to support numbers in the range of [0, 7], as we also handle 7, and reverse the dequeu and buffer respectively. but we failed generate 7 with the RNG.

so, in this change,

* enlarge the maximum value of of the uniform distribution from 6 to 7
* do not fallthrough in the "case 7" block, otherwise the test will abort() at seeing "7" generated from RNG.

this change should improve the coverage of the test. and enable us to pass `-Wimplicit-fallthrough` command line option to the compiler for better compile-time check.

### DIFF
--- a/tests/unit/circular_buffer_fixed_capacity_test.cc
+++ b/tests/unit/circular_buffer_fixed_capacity_test.cc
@@ -106,7 +106,7 @@ using deque = std::deque<int>;
 
 BOOST_AUTO_TEST_CASE(test_random_walk) {
     auto rand = std::default_random_engine();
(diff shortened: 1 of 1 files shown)


tchaikov:
@avikivity hi Avi, could you help review this change?

PR #1737, opened 2023-07-13
Fix server-side RPC stream shutdown

xemul:
When RPC server stops or shutdowns it leaves stream connections alive until they get closed "naturally" by TCP. Also there's a stream socket registration race that can leave stream socket alive.

refs: scylladb/scylladb#13100


xemul:
@avikivity , @gleb-cloudius review ping

include/seastar/rpc/rpc.hh
```diff
@@ -595,6 +595,7 @@ public:
             return _info.server;
         }
         future<> deregister_this_stream();
+        future<> abort_all_streams();
```
gleb-cloudius, line 598: > This mostly comes unnoticed because when server stops and aborts all its
connections the connections' peers on other nodes get closed too by TCP.
Client-side connection closing does aborts all associated stream which,
in turn, makes server-side stream connections closed by TCP. However,
it's not nice to rely on two network round-trips to abort a stream, it
should better be aborted instantly.

What problem does it fix? The current behavior is by design. The design may be wrong, but I want to understand why exactly.

xemul: When `rpc::server` shuts down it calls `stop()` on all linked `connection`s and waits for them to resolve `_stopped` futures. When the whole `rpc::server::shutdown()` resolves there can be streaming ends still alive, because server-side stream sockets are closed ... indirectly(?) and asynchronously via the following chain of events:

- server-side RPC socket is eventually closed and sends FIN to the peer
- client-side RPC socket receives FIN and makes client::connection::process() to resolve and "destroy" the socket, in particular, `client::connection::abort_all_streams()` is called
- client-side stream socket is eventually closed and sends FIN to the peer (back to server)
- server-side stream socket receives FIN and makes server::connection::process to resolve and "destroy" the socket

This makes stream sockets out-survive the messaging_service.shutdown() call for some time. I think it's better to close server-side stream ends immediately without waiting for two round-trips.

gleb-cloudius: > This makes stream sockets out-survive the messaging_service.shutdown() call for some time. I think it's better to close server-side stream ends immediately without waiting for two round-trips.

I do not mind round trips (you will still messaging sent on tcp level either because of bi-directional close or RSTs because socket is closed already), but out-surviving messaging_service.shutdown() is indeed something we better avoid. I misread the code and thought that stream connections are also in `_conns` which would prevent this from happening, but now I see that they are not.

mykaul:
@avikivity - please review.

bhalevy:
@avikivity ping review

include/seastar/rpc/rpc.hh
avikivity: If the remote is dead, there won't be any round trips, and we'll wait until TCP figures out it needs to shut down the connection. Maybe not too bad if keepalive is enabled, but let's not rely on that.

src/rpc/rpc.cc
```diff
@@ -1112,7 +1112,7 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
               if (is_stream()) {
                   return deregister_this_stream();
               } else {
-                  return make_ready_future<>();
+                  return abort_all_streams();
```
avikivity, line 1118: Let's ratify C++23 so we can convert this code to coroutines

src/rpc/rpc.cc
```diff
@@ -1143,6 +1143,16 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
...
   }
 
+  future<> server::connection::abort_all_streams() {
+      return parallel_for_each(_streams | boost::adaptors::map_values, [] (xshard_connection_ptr s) {
+          return smp::submit_to(s->get_owner_shard(), [s] {
+              s->get()->abort();
+          });
+      }).then([this] {
+          _streams.clear();
+      });
```
avikivity, line 1156: Could there be some strange race here?
1. some operation begins that would eventually add an entry to _streams
2. we stop the server
3. the operation completes just before _streams.clear()

I think it's still better than what we have now, even with the race.

avikivity:
Please add a unit test to tests/unit/rpc_test.cc

DoronArazii:
> upd:
> 
> * extended rpc stream tests to shutdown server on the fly

@avikivity need merge

DoronArazii:
@avikivity ping

PR #1528, opened 2023-03-05
net, websocket: replace Crypto++ calls with GnuTLS

tchaikov:
since we already depend on GnuTLS which provides the functions which we use Crypto++ for. so let's rewrite the Crypto++ function calls using the ones provided by GnuTLS for less dependencies.

Fixes #680
Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>

PR #1740, opened 2023-07-14
net: use designated initialization when appropriate

tchaikov:
GCC encourages us to initialize all fields of a plain C struct,
which does not provide default value(s) for its member variables,
and hence does not have a default constructor.

and it warns us like:
```
/home/kefu/dev/seastar/src/net/dns.cc:152:36: error: missing field 'timeout' initializer [-Werror,-Wmissing-field-initializers]
  152 |         ares_options a_opts = { 0, };
      |                                    ^
1 error generated.
```

and per C++ standard (draft), we can just initialize fields in the order
of their declaration, and zero-initialize the missing fields. see https://eel.is/c++draft/dcl.init#list . which
defines how the fields in an object can be initialized with the
designated initializer, and if the field(s) are not in the list, they
are value-initialized, which is equivalent to zero-initialization in
this case.

so in this change, we use designated initialization to initialize the
C structs instead of initializing them using the form like `{0, }`.

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -149,7 +149,7 @@ public:
         // do this ourselves, and instead set ares options
         // here, but it seems more error prone (me parsing
         // resolv.conf -> hah!)
-        ares_options a_opts = { 0, };
+        ares_options a_opts = {};
 
         // For now, use the default "fb" query order
         // (set explicitly lest we forget).
--- a/src/net/posix-stack.cc
+++ b/src/net/posix-stack.cc
@@ -942,10 +942,11 @@ std::vector<network_interface> posix_network_stack::network_interfaces() {
 
         auto pid = ::getpid();
 
-        sockaddr_nl local = { 0, };
-        local.nl_family = AF_NETLINK;
-        local.nl_pid = pid;
-        local.nl_groups = RTMGRP_IPV6_IFADDR|RTMGRP_IPV4_IFADDR;
+        sockaddr_nl local = {
+          .nl_family = AF_NETLINK,
+          .nl_pid = static_cast<unsigned int>(pid),
+          .nl_groups = RTMGRP_IPV6_IFADDR|RTMGRP_IPV4_IFADDR,
+        };
 
         throw_system_error_on(bind(fd, (struct sockaddr *) &local, sizeof(local)) < 0, "could not bind netlink socket");
 
@@ -960,12 +961,7 @@ std::vector<network_interface> posix_network_stack::network_interfaces() {
                     rtgenmsg gen;
                     ifaddrmsg addr; 
                 }; 
-            } req = { {0}, };
-
-            sockaddr_nl kernel = { 0, }; 
-            msghdr rtnl_msg = { 0, };
-    
-            kernel.nl_family = AF_NETLINK; /* fill-in kernel address (destination of our message) */
+            } req = {};
 
             req.hdr.nlmsg_len = NLMSG_LENGTH(sizeof(struct rtgenmsg));
             req.hdr.nlmsg_type = msg;
@@ -979,15 +975,19 @@ std::vector<network_interface> posix_network_stack::network_interfaces() {
                 req.addr.ifa_family = AF_UNSPEC;
             }
 
-            iovec io;
-
-            io.iov_base = &req;
-            io.iov_len = req.hdr.nlmsg_len;
-
-            rtnl_msg.msg_iov = &io;
-            rtnl_msg.msg_iovlen = 1;
-            rtnl_msg.msg_name = &kernel;
-            rtnl_msg.msg_namelen = sizeof(kernel);
+            sockaddr_nl kernel = {
+              .nl_family = AF_NETLINK, /* fill-in kernel address (destination of our message) */
+            };
+            iovec io = {
+              .iov_base = &req,
+              .iov_len = req.hdr.nlmsg_len,
+            };
+            msghdr rtnl_msg = {
+              .msg_name = &kernel,
+              .msg_namelen = sizeof(kernel),
+              .msg_iov = &io,
+              .msg_iovlen = 1,
+            };
 
             throw_system_error_on(::sendmsg(fd, (struct msghdr *) &rtnl_msg, 0) < 0, "could not send netlink request");
             /* parse reply */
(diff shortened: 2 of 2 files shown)


src/net/posix-stack.cc
```diff
@@ -998,8 +998,8 @@ std::vector<network_interface> posix_network_stack::network_interfaces() {
             bool done = false;
 
             while (!done) {
-                msghdr rtnl_reply = { 0, };
-                iovec io_reply = { 0, };
+                msghdr rtnl_reply = {};
+                iovec io_reply = {};
```
xemul, line 1002: Both rtnl_reply and io_reply are initialized with some values few lines below, I think it's worth just initialize both with those values here instead of zeroifying

src/net/posix-stack.cc
```diff
@@ -960,10 +960,10 @@ std::vector<network_interface> posix_network_stack::network_interfaces() {
                     rtgenmsg gen;
                     ifaddrmsg addr; 
                 }; 
-            } req = { {0}, };
+            } req = {};
 
-            sockaddr_nl kernel = { 0, }; 
-            msghdr rtnl_msg = { 0, };
+            sockaddr_nl kernel = {};
+            msghdr rtnl_msg = {};
```
xemul, line 966: This thing is initialized way below this place, so it make sense to move it and initialize declaration-time instead of zeoifying

src/net/posix-stack.cc
```diff
@@ -960,10 +960,10 @@ std::vector<network_interface> posix_network_stack::network_interfaces() {
                     rtgenmsg gen;
                     ifaddrmsg addr; 
                 }; 
-            } req = { {0}, };
+            } req = {};
 
-            sockaddr_nl kernel = { 0, }; 
-            msghdr rtnl_msg = { 0, };
+            sockaddr_nl kernel = {};
```
xemul, line 965: Note, that this sockaddr_nl is initialized by the kernel, so zeroifying it makes no practical sense

src/net/dns.cc
```diff
@@ -149,7 +149,7 @@ class net::dns_resolver::impl
         // do this ourselves, and instead set ares options
         // here, but it seems more error prone (me parsing
         // resolv.conf -> hah!)
-        ares_options a_opts = { 0, };
+        ares_options a_opts = {};
```
xemul, line 152: This a_opts is initialized below. I think it's better to declare it uninitialized, but likely compiler should be smart enough not to _re_-initialize it later

tchaikov: yeah, but we don't initialize all its fields. but i think we are obliged to initialize every single bit in this struct. so, probably we should still use `{}` here.

src/net/posix-stack.cc
tchaikov: right. but, again, we don't initialize every field in this struct. for instance, we don't initialize [msg_control](https://elixir.bootlin.com/glibc/latest/source/bits/latest/C/ident/msg_control) . so i think we still need to keep this value-initialization.

src/net/posix-stack.cc
tchaikov: i concur with you on dropping the zero-initialization of  `io_reply`. but i am not quite sure about `rtnl_reply`. i agree that the syscall does not read `[msg_control](https://docs.huihoo.com/doxygen/linux/kernel/3.7/structmsghdr.html#ae0bcd626259df3c2f039cee9353150c6)` or [msg_controllen](https://docs.huihoo.com/doxygen/linux/kernel/3.7/structmsghdr.html#a9d26eb73cb060c84f34a9e1aa58e631e) fields in this case, but i don't want to change the existing behavior in this refactor. -- especially, i still don't have an environment for testing the DPDK application at this moment. 

regarding `io_reply`, i created another commit to use designated initialization. so that the first commit can focus on the preserving the existing code structure  in addition to its behavior.

src/net/posix-stack.cc
tchaikov: i agree, but i'd prefer leave this to another change. to avoid initializing fields in an output parameter is not in the scope of this PR. it created to address the compiling warning. we can have another PR to do this. and we would need to test it. again, i don't have the environment for testing DPDK application. FWIW, that's also one of reasons why https://github.com/scylladb/seastar/pull/1528 is stalled.

tchaikov:
v2: 

- add a commit to use designed initialization to avoid the unnecessary zero initialization

@xemul thank you for your review. could you take another look?

src/net/posix-stack.cc
xemul: But if you initialize part of the struct with designated initializers, the missing parts are zeroified by compiler, aren't they?

src/net/posix-stack.cc
xemul: The `rtnl_reply` is initialized below:

```
                 rtnl_reply.msg_iov = &io_reply;
                 rtnl_reply.msg_iovlen = 1;
                 rtnl_reply.msg_name = &kernel;
                ...
```

you can patch it to be initialized with designated initializers instantly

src/net/dns.cc
xemul: Yes, I agree with this place

src/net/posix-stack.cc
tchaikov: yes, i can. but we have to initialize the members *in the order of* their declarations. if what you suggest is to initialize `msg_control` and `msg_controllen` as well. well, that's an option also. i don't see the merit of doing so in comparison with the zero initialization. but if you still think we should go this way, i will implement in the next revision.

src/net/posix-stack.cc
tchaikov: hmm, right. will implement this in the next revision.

tchaikov:
v3

- use designated initialization in place of zero initialization. simpler this way.

@xemul thank you for your suggestion. i thought that the designated initialization does not initialize the missing fields unless they have their default ctor. but i realized that i was wrong after reading the standard. as you mentioned, they are zero initialized.

could you take yet another look?

src/net/posix-stack.cc
tchaikov: switched to designated initializer anyway. i still agree with you that zerofying it does not make sense. but i would leave it to another change.

src/net/posix-stack.cc
```diff
@@ -942,7 +942,7 @@ std::vector<network_interface> posix_network_stack::network_interfaces() {
 
         auto pid = ::getpid();
 
-        sockaddr_nl local = { 0, };
+        sockaddr_nl local = {};
```
xemul, line 945: Why not designate-initialize `local` instantly as well?

tchaikov: ahh, indeed. changed.

tchaikov:
v4:

- designate-initialize `local` as well.

tchaikov:
@xemul hi Pavel, thank you for reviews. could you please take another look?

PR #1741, opened 2023-07-14
metrics: change metadata vector to chunked_fifo

NyaliaLui:
Upon resizing, a std::vector will allocate a large piece of contiguous memory which may become fragmented over time. This can lead to oversized allocations on the metadata vector occurring at the call to vector.emplace_back():

```cpp
void impl::update_metrics_if_needed() {
if (_dirty) {
   _metadata = ::seastar::make_shared<metric_metadata>();

   auto mt_ref = ::seastar::make_shared<metric_metadata>();
   auto &mt = *(mt_ref.get());
   mt.reserve(_value_map.size());
   _current_metrics.resize(_value_map.size());
   size_t i = 0;
   for (auto&& mf : _value_map) {
       metric_metadata_vector metrics;
       _current_metrics[i].clear();
       for (auto&& m : mf.second) {
           if (m.second && m.second->is_enabled()) {
               metrics.emplace_back(m.second->info()); // Backtrace shows overalloc occurs here
               _current_metrics[i].emplace_back(m.second->get_function());
           }
       }
       if (!metrics.empty()) {
           mt.emplace_back(metric_family_metadata{mf.second.info(), std::move(metrics)});
           i++;
       }
   }
   _current_metrics.resize(i);
   _metadata = mt_ref;
   _dirty = false;
}
}
```

Therefore, this PR switches the metadata vector with a chunked_fifo.

Fixes: https://github.com/scylladb/seastar/issues/1844

Changes from force-push `d8662a3`:
- Split the patches into two commits
- Add more explanation to why `chunked_fifo::basic_iterator` needs a default ctor

Changes from force-push `55d6663`:
- Reduce more potential overallocs in member `_current_metrics` by switching `std::vector<std::vector<metric_function>>` to `std::vector<chunked_fifo<metric_function>>`
- Reduce more potential overallocs in `values_copy::values` by changing `metrics_values` to a `std::deque` 

Changes from force-push `be869b7`:
- Commit https://github.com/scylladb/seastar/commit/3e9321cb46d30adf0d12ff24e401d718e619de0b added a default ctor to `chunked_fifo::basic_iterator`, so this force-push dropped the commit in this patch that added the same default ctor.

Changes from force-push `8a06f57`:
- Changed `chunked_fifo` to a `std::deque`

### DIFF
--- a/include/seastar/core/metrics_api.hh
+++ b/include/seastar/core/metrics_api.hh
@@ -29,6 +29,8 @@
 #include <boost/functional/hash.hpp>
 #endif
 
+#include <deque>
+
 /*!
  * \file metrics_api.hh
  * \brief header file for metric API layer (like prometheus or collectd)
@@ -322,7 +324,7 @@ public:
 
 using value_map = std::map<sstring, metric_family>;
 
-using metric_metadata_vector = std::vector<metric_info>;
+using metric_metadata_fifo = std::deque<metric_info>;
 
 /*!
  * \brief holds a metric family metadata
@@ -333,7 +335,7 @@ using metric_metadata_vector = std::vector<metric_info>;
  */
 struct metric_family_metadata {
     metric_family_info mf;
-    metric_metadata_vector metrics;
+    metric_metadata_fifo metrics;
 };
 
 using value_vector = std::vector<metric_value>;
@@ -355,7 +357,7 @@ class impl {
     bool _dirty = true;
     shared_ptr<metric_metadata> _metadata;
     std::set<sstring> _labels;
-    std::vector<std::vector<metric_function>> _current_metrics;
+    std::vector<std::deque<metric_function>> _current_metrics;
     std::vector<relabel_config> _relabel_configs;
 public:
     value_map& get_value_map() {
@@ -380,7 +382,7 @@ public:
 
     shared_ptr<metric_metadata> metadata();
 
-    std::vector<std::vector<metric_function>>& functions();
+    std::vector<std::deque<metric_function>>& functions();
 
     void update_metrics_if_needed();
 
--- a/include/seastar/core/metrics_api.hh
+++ b/include/seastar/core/metrics_api.hh
@@ -340,7 +340,7 @@ struct metric_family_metadata {
 
 using value_vector = std::vector<metric_value>;
 using metric_metadata = std::vector<metric_family_metadata>;
-using metric_values = std::vector<value_vector>;
+using metric_values = std::deque<value_vector>;
 
 struct values_copy {
     shared_ptr<metric_metadata> metadata;
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -366,7 +366,6 @@ foreign_ptr<values_reference> get_values() {
     auto& mv = res.values;
     res.metadata = get_local_impl()->metadata();
     auto & functions = get_local_impl()->functions();
-    mv.reserve(functions.size());
     for (auto&& i : functions) {
         value_vector values;
         values.reserve(i.size());
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -399,7 +399,7 @@ void impl::update_metrics_if_needed() {
         _current_metrics.resize(_value_map.size());
         size_t i = 0;
         for (auto&& mf : _value_map) {
-            metric_metadata_vector metrics;
+            metric_metadata_fifo metrics;
             _current_metrics[i].clear();
             for (auto&& m : mf.second) {
                 if (m.second && m.second->is_enabled()) {
@@ -426,7 +426,7 @@ shared_ptr<metric_metadata> impl::metadata() {
     return _metadata;
 }
 
-std::vector<std::vector<metric_function>>& impl::functions() {
+std::vector<std::deque<metric_function>>& impl::functions() {
     update_metrics_if_needed();
     return _current_metrics;
 }
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -392,7 +392,7 @@ public:
             // does not exist, because everything is sorted by metric family name, this is fine.
             if (metadata.mf.name == name()) {
                 const mi::value_vector& values = metric_family->values[pos_in_metric_per_shard];
-                const mi::metric_metadata_vector& metrics_metadata = metadata.metrics;
+                const mi::metric_metadata_fifo& metrics_metadata = metadata.metrics;
                 for (auto&& vm : boost::combine(values, metrics_metadata)) {
                     auto& value = boost::get<0>(vm);
                     auto& metric_metadata = boost::get<1>(vm);
--- a/src/core/scollectd.cc
+++ b/src/core/scollectd.cc
@@ -412,7 +412,7 @@ void impl::arm() {
 void impl::run() {
     typedef size_t metric_family_id;
     typedef seastar::metrics::impl::value_vector::iterator value_iterator;
-    typedef seastar::metrics::impl::metric_metadata_vector::iterator metadata_iterator;
+    typedef seastar::metrics::impl::metric_metadata_fifo::iterator metadata_iterator;
     typedef std::tuple<metric_family_id, metadata_iterator, value_iterator, type_id, cpwriter> context;
 
     auto ctxt = make_lw_shared<context>();


amnonh:
The change looks safe LGTM

nyh (approved):
Looks good to me, although I would have preferred it to be split to two separate patches (one improving chunked_fifo, the other the metrics).

include/seastar/core/chunked_fifo.hh
```diff
@@ -146,6 +146,7 @@ private:
         size_t _item_index = 0;
 
     protected:
+        inline basic_iterator() noexcept;
```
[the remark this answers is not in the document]
tchaikov: the reason we need to have an uninitialized iterator is that we need to create an empty `context` at https://github.com/scylladb/seastar/blob/e908cfce2d4488976f3b91906954f435aaa99238/src/core/scollectd.cc#L418 , and the `context` has an `seastar::metrics::impl::metric_metadata_fifo::iterator` after this change. without a default ctor, `make_lw_shared<context>()` won't compile.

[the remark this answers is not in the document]
tchaikov: i concur with @nyh that it'd be better to explain why we need a default ctor for the `basic_iterator`. as it might be otherwise not self-explained.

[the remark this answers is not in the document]
nyh: Yes, I guessed that some other code must have used an uninitialized iterator and initialize it later, and since this works with std::vector (and I didn't check, probably also with std::dequeue) then it should also work with chunked_fifo, so it's a good change.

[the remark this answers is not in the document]
NyaliaLui: Yes, I added the default ctor to the iterator because of compilation issues with `make_lw_shared<context>()`. I will split up the commits and explain that further in the commit message.

xemul:
Isn't the same jumbo-vector problem stands for `_current_metrics` and, consequentially, for inhabitants of `styruct values_copy`? @NyaliaLui

NyaliaLui:
> Isn't the same jumbo-vector problem stands for `_current_metrics` and, consequentially, for inhabitants of `styruct values_copy`? @NyaliaLui

Apologies for late reply, I see how `_current_metrics` could overalloc but I don't see how `values_copy` are at risk. Could you explain a bit more?

Redpanda ran into this overalloc only with `metric_metadata_vector` so I changed it to `metric_metadata_fifo`. @xemul

include/seastar/core/metrics_api.hh
```diff
@@ -323,7 +324,7 @@ public:
 
 using value_map = std::map<sstring, metric_family>;
 
-using metric_metadata_vector = std::vector<metric_info>;
+using metric_metadata_fifo = chunked_fifo<metric_info>;
```
avikivity, line 327: Isn't std::deque more suitable here?

NyaliaLui:
> > Isn't the same jumbo-vector problem stands for `_current_metrics` and, consequentially, for inhabitants of `styruct values_copy`? @NyaliaLui
> 
> Apologies for late reply, I see how `_current_metrics` could overalloc but I don't see how `values_copy` are at risk. Could you explain a bit more?
> 
> Redpanda ran into this overalloc only with `metric_metadata_vector` so I changed it to `metric_metadata_fifo`. @xemul

I see `values_copy::values` overallocing since it gets the same size as `_current_metrics` in `src/core/metrics.cc`:
```cpp
foreign_ptr<values_reference> get_values() {
    shared_ptr<values_copy> res_ref = ::seastar::make_shared<values_copy>();
    auto& res = *(res_ref.get());
    auto& mv = res.values;
    res.metadata = get_local_impl()->metadata();
    auto & functions = get_local_impl()->functions();
    mv.reserve(functions.size()); // Potentially here
    for (auto&& i : functions) {
        value_vector values;
        values.reserve(i.size());
        for (auto&& v : i) {
            values.emplace_back(v());
        }
        mv.emplace_back(std::move(values));
    }
    return res_ref;
}
```

include/seastar/core/metrics_api.hh
NyaliaLui: I argue for `chunked_fifo` because `std::deque` still does some contiguous allocations since it uses chunks of arrays. The problem that we were trying to address here was an overallocation issue on contiguous allocations.

May I ask, what is your concern? @avikivity

NyaliaLui: > I argue for `chunked_fifo` because `std::deque` still does some contiguous allocations since it uses chunks of arrays. The problem that we were trying to address here was an overallocation issue on contiguous allocations.
> 
> May I ask, what is your concern? @avikivity

After posting I realized that `std::deque` is a good substitute for `values_copy::values` because it needs random access at locations [scollectd.cc:429](https://github.com/scylladb/seastar/blob/28daa8f50ccd5737b1779a2656d4330473d53d59/src/core/scollectd.cc#L429), [scollectd.cc:450](https://github.com/scylladb/seastar/blob/28daa8f50ccd5737b1779a2656d4330473d53d59/src/core/scollectd.cc#L450), [scollectd.cc:471](https://github.com/scylladb/seastar/blob/28daa8f50ccd5737b1779a2656d4330473d53d59/src/core/scollectd.cc#L471) and `chunked_fifo` does not support random access.

avikivity: Mostly that std::queue is a standard type and more similar to std::vector.

NyaliaLui: > Mostly that std::queue is a standard type and more similar to std::vector.

Thank you for the response. I am a little confused though, are there more items to discuss? cc: @avikivity

NyaliaLui: > Mostly that std::queue is a standard type and more similar to std::vector.

These types are similar and they do have a design/interface that is unique compared to `chunked_fifo`. However, the variables `_current_metrics` and `metric_metadata_vector` do not seem to rely on those differences. A `chucked_fifo` also decreases the potential of memory fragmentation. For these reasons, I believe a `chunked_fifo` is OK.

I also believe a `std::deque` is OK for those places that could use improved memory management but still need the interfaces from `std`,  (e.g., `values_copy::values`).

NyaliaLui:
@nyh @tchaikov I resolved conflicts and this patch is ready for re-review again.

NyaliaLui:
@tchaikov @avikivity @amnonh 
Wondering if any other changes are needed here?

Also, @nyh it looks like github glitched and did not dismiss your review from July 16th

nyh (approved):
The patch looks ok to me, but I think the commit messages can be improved. In particular, both commit messages talk about "over allocations", but it isn't clear to me what this means. std::vector doesn't allocate **more** than seastar::chunked_fifo. I guess by "over allocation" you meant "oversized allocation", i.e., a single contiguous allocation that is very big, and Seastar warns about because in highly-fragmented memory it may become impossible to do this allocation and it will crash the application.

In general, it is always good to start by opening a github issue describing a **problem** (e.g., did you see an "oversized allocation" warning? did you run out of memory? In what case - did you add a lot of metrics? Or what?), and then your commit message can say it solves that issue (give a short summary and then say "Fixes #nnn").




@amnonh please review too, you wrote most of the code that this patch modifies.

tchaikov (approved):
lgtm

NyaliaLui:
@nyh What you say makes sense, I created https://github.com/scylladb/seastar/issues/1844, linked this PR to that issue, and amended the PR description. It seems that changing the commit message is a non-blocker, so I'll leave that alone for now.

tchaikov:

include/seastar/core/metrics_api.hh
avikivity: I think deque is more suitable, chunked_fifo is a specialized type while deque is general-purpose.

NyaliaLui: > I think deque is more suitable, chunked_fifo is a specialized type while deque is general-purpose.

@avikivity I am a little confused, are you requesting changes here?

avikivity: Yes, please switch to deque.

NyaliaLui: Made the change.

NyaliaLui:
Pinging in case this got lost somewhere. CC @avikivity @tchaikov @xemul 

All requested changes are submitted.

PR #1742, opened 2023-07-15
Cleanup class metrics::registered_metrics

xemul:
It had changed since it was first introduced and can be brushed up a little bit:

- make it non-virtual class
- mark value-function getter const
- remove now unused operator()

### DIFF
--- a/include/seastar/core/metrics_api.hh
+++ b/include/seastar/core/metrics_api.hh
@@ -210,14 +210,13 @@ public:
 
 class impl;
 
-class registered_metric {
+class registered_metric final {
     metric_info _info;
     metric_function _f;
     shared_ptr<impl> _impl;
 public:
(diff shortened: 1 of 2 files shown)


amnonh:
LGTM

xemul:
@scylladb/seastar-maint , please merge

nyh (approved):
Looks good to me, but I'm not familiar enough with this code to agree or disagree with your guesses on what this code was supposed to be doing, so @amnonh can you please review this?

nyh:
> Looks good to me, but I'm not familiar enough with this code to agree or disagree with your guesses on what this code was supposed to be doing, so @amnonh can you please review this?

Oh, sorry, @amnonh I see you already reviewed this. Good. I'll merge it.

PR #1744, opened 2023-07-16
util: add formatter using IEC/SI units

tchaikov:
to ease the printing the size of information content, like the bandwidth and size of file. a data_size type and its formatter is defined.

### DIFF
--- a/include/seastar/util/conversions.hh
+++ b/include/seastar/util/conversions.hh
@@ -25,6 +25,7 @@
 #include <cstdlib>
 #include <string_view>
 #include <vector>
+#include <fmt/format.h>
 #endif
 #include <seastar/util/modules.hh>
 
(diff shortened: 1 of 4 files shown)


nyh:
This looks like a nice utility function, but I have doubts why it belongs in Seastar. Seastar isn't supposed to be like Boost a collection of random useful utility functions. Do you plan to use this utility inside Seastar somewhere? Do you suspect that Seastar applications have a special use for this ability to print file sizes that isn't needed by non-Seastar applications?

tchaikov:
> This looks like a nice utility function, but I have doubts why it belongs in Seastar. Seastar isn't supposed to be like Boost a collection of random useful utility functions. Do you plan to use this utility inside Seastar somewhere? Do you suspect that Seastar applications have a special use for this ability to print file sizes that isn't needed by non-Seastar applications?

i was asking myself the same questions when pushing this change. on one hand i've been taking efforts to ditch the generic formatters in seastar, on the other hand i am adding formatter for general purpose to this library. that's kind of weird...

tchaikov:
i will close this pull request. probably a better home of it is scylladb.

PR #1747, opened 2023-07-17
resource: take kernel min_free_kbytes into account when allocating memory

avikivity:
The kernel declares OOM when memory dips below /proc/sys/vm/min_free_kbytes. On normal systems, this is a reasonable value (69MB on my 128GB desktop) and is far below the 7% we already reserve for the kernel. But on ARM systems with 64kB pages and transparent hugepages enabled, this becomes very large (~800MB on a 16GB system), presumably so there is space to collect a 512MB hugepage. This reservation is already 5% of total system memory, and accounting for other things, the system will OOM as soon as Seastar allocates all of its memory. With prefault, this is within a few seconds of starting, as seen in [1].

Fix this by subtracting min_free_kbytes from available memory. On sane systems this has negligible effect, and on systems with very large hugepage sizes, this will allow them to avoid running out of memory.

[1] https://github.com/scylladb/scylladb/issues/14721

### DIFF
--- a/src/core/resource.cc
+++ b/src/core/resource.cc
@@ -223,7 +223,24 @@ optional<T> read_setting_V1V2_as(std::string cg1_path, std::string cg2_fname) {
 
 namespace resource {
 
+static
+size_t
+kernel_memory_reservation() {
+    try {
+        return read_first_line_as<size_t>("/proc/sys/vm/min_free_kbytes") * 1024;
+    } catch (...) {
+        return 0;
+    }
+}
+
 size_t calculate_memory(const configuration& c, size_t available_memory, float panic_factor = 1) {
+    auto kernel_reservation = kernel_memory_reservation();
+    if (kernel_reservation >= 200'000'000) {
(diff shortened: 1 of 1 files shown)


src/core/resource.cc
```diff
@@ -223,7 +223,18 @@ optional<T> read_setting_V1V2_as(std::string cg1_path, std::string cg2_fname) {
...
+kernel_memory_reservation() {
+    try {
+        return read_first_line_as<size_t>("/proc/sys/vm/min_free_kbytes") * 1024;
+    } catch (...) {
+        return 0;
+    }
+}
+
 size_t calculate_memory(const configuration& c, size_t available_memory, float panic_factor = 1) {
+    available_memory -= kernel_memory_reservation();
```
xemul, line 237: Please, add some logging in case the obtained value doesn't meet the expectations described in the commit message.

Or -- at the end of the `calculate_memory` print info about collected+configured numbers and the evaluated result.

avikivity: Unconditional printing is too noise. I'll print something if the value is unexpected.

avikivity:
v2: warn if the kernel reservation is unexpectedly high

Issue #1428, opened 2023-01-18
Extend ioinfo to show shat's usually printed in startup logs


xemul:
These include:
- effective latency goal
- per-request size capacities
- RL scheduler parameters

PR #1754, opened 2023-07-21
Print more IO info with ioinfo app

xemul:
On start reactor and io-queue do some math calculating IO-queue's configuration. For the ease of debugging all this math was printed into log with info level. Now it's more-or-less stable and predictable, so the printing can be dropped not to make too many noise into logs. To still have the ability to get the math results there's ioinfo app out there that's extended by this PR to print all internal IO scheduler configuration

fixes: #1428


xemul:
@avikivity , please review. This set greatly reduces the noise in start-up logs

xemul:
upd:
- rebased

xemul:
@avikivity , please review

xemul:
@avikivity , please review

Issue #634, opened 2019-04-24
Drop seastar::sstring type, use standard std::string instead


nyh:
The reason why seastar has its own "sstring" type instead of using std::string is historical, due to C++ library implementation when the Seastar projects started: In gcc before version 5, std::string used to use atomic operations which are bad for many-core applications like Seastar. As the nice blog post https://shaharmike.com/cpp/std-string/ explains why this was the case in pre-C++-11 and why gcc for ABI backward-compatibility reasons left this implementation long after 2011. But since gcc 5, this old implementation is gone. Today's std::string implementations in gcc and clang use exactly the same techniques as seastar::sstring does: no atomic operations, and in-structure storage for small strings (a.k.a. "small-string optimization") - the aforementioned blog points out that clang's implementation is particularly clever.

For this reason, seastar::sstring is no longer needed and we can switch to using std::string.

Since Seastar does not care about ABI compatibility (we do not yet support a Seastar shared library - see issue #467), and only about source-code compatibility, it should be easy to change Seastar so that seastar::sstring is merely an alias to std::string, without breaking compatibility for existing applications. Applications can then start gradually changing their own code to use std::string instead of seastar::sstring.

nyh:
@avikivity pointed out in the mailing list, "I don't think it's simple. The interfaces are not identical, sstring has several features beyond std::string.".

One example I found is the null-termination template parameter introduced in commit 7328d17337523ae1e64dbd07cc5aa14628a022f0 which std::basic_string doesn't support, but seastar::basic_sstring does.

nyh:
Another difference is that `seastar::sstring` has a constructor taking `sstring::initialized_later()` and a size, which `std::string` doesn't have. This constructor allows creating an uninitialized string of a certain length, and only then fill its data. We also have a convenience function `make_sstring()` to take a bunch of arguments and concatenate them into one new string. The closest constructor that `std::string` has is one taking an iterator pair. It's not as general as `make_sstring()` allowing parameters of different type, but I don't think this is a critical feature (of course there's the issue of backward compatibility, though....).

Yet another feature we have and std::string doesn't is `sstring::release()`,  introduced in commit a2ca5568369c2f782477de890b468b0650ac2a33. This allows taking out the internal buffer held by sstring so it can be held by another owner without copying.

nyh:
[...]

PR #1755, opened 2023-07-24
sstring: add more accessors

tchaikov:
in this series, more accessors are added to `seastar::sstring`. the goal is to make sstring more standard compliant, so that it can be used in lieu of `std::string` when appropriate.

### DIFF
--- a/include/seastar/core/sstring.hh
+++ b/include/seastar/core/sstring.hh
@@ -596,6 +596,18 @@ public:
         }
     }
 
+    constexpr bool starts_with(std::basic_string_view<char_type, traits_type> sv) const noexcept {
+        return size() > sv.size() && compare(0, sv.size(), sv) == 0;
+    }
+
+    constexpr bool starts_with(char_type c) const noexcept {
+        return !empty() && traits_type::eq(front(), c);
+    }
+
+    constexpr bool starts_with(const char_type* s) const noexcept {
+        return starts_with(std::basic_string_view<char_type, traits_type>(s));
+    }
+
     void swap(basic_sstring& x) noexcept {
         contents tmp;
         tmp = x.u;
--- a/include/seastar/core/sstring.hh
+++ b/include/seastar/core/sstring.hh
@@ -256,20 +256,18 @@ public:
         return npos;
     }
 
-    size_t find(const basic_sstring& s, size_t pos = 0) const noexcept {
-        const char_type* it = str() + pos;
-        const char_type* end = str() + size();
-        const char_type* c_str = s.str();
-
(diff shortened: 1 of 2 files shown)


tchaikov:
this change was inspired by https://github.com/scylladb/scylladb/pull/14800/files#diff-7b428ff10eee5bec5a6f74223c8dee8d61085bbaf290538bd3ae6e8c4c606efdR1139 where the author uses 
```c++
 if (!fn_name.find("castas")) {
     // something we should do when fn_name starts with "castas" 
 }
```

IMHO, this is not quite readable.

include/seastar/core/sstring.hh
```diff
@@ -296,6 +294,23 @@ public:
...
+    constexpr size_t find(const char_type* s, size_t pos = 0) const noexcept {
+        return find(s, pos, traits_type::length(s));
+    }
+
+    size_t find(const basic_sstring& s, size_t pos = 0) const noexcept {
+        return find(s.str(), pos, s.size());
+    }
+
+    template<class StringViewLike,
+             std::enable_if_t<std::is_convertible_v<StringViewLike,
```
xemul, line 306: Sanity check: usage of SEASTAR_CONCEPT is now not mandatory, isn't it?

tchaikov: Yeah, I think these helper are also handy to C++17 users, so chose to avoid using C++20.

nyh:
> in this series, more accessors are added to `seastar::sstring`. the goal is to make sstring more standard compliant, so that it can be used in lieu of `std::string` when appropriate.

I agree that seastar::sstring should be more compatible with std::string, but I don't agree with your wish to use seastar::sstring in lieu of std::string - I think it should be the other way around, Seastar applications should avoid seastar::sstring as much as possible - see https://github.com/scylladb/seastar/issues/634

tchaikov:
sorry for the misunderstanding. i think, i, well, again, sent the wrong message to the reviewer and greater community. i am not a big fan of `sstring` or homebrew `string` solutions. on the contrary, i want to reduce the pain of switching over to `std::string` or `std::string_view`. the goal of adding more accessors to `sstring` is not to make it "better", but to encourage developers to replace it with the string-alike class provided by the standard library.

but @xemul mentioned at https://github.com/scylladb/seastar/blob/ba1e39dce7d4938d846c9c4ca104035deafbe2c7/include/seastar/core/sstring.hh#L56-L59 that there was a "a small performance degradation". i need to verify this.

BTW, the initiative to remove the formatters for standard library containers in Seastar and ScyllaDB is of the same spirit, despite it is difficult in the sense of the amount of work.

PR #1756, opened 2023-07-24
rpc: Generalize get_stats_internal() method

xemul:
It exists on client and server connecitons, but the value it returns sits on their common ancestor -- the rpc::connection class. So the method can exist in the same place

### DIFF
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -348,6 +348,9 @@ public:
     connection_id get_connection_id() const noexcept {
         return _id;
     }
+    stats& get_stats_internal() noexcept {
+        return _stats;
+    }
     xshard_connection_ptr get_stream(connection_id id) const;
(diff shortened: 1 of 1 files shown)


tchaikov (approved):

include/seastar/rpc/rpc.hh
```diff
@@ -348,6 +348,9 @@ public:
     connection_id get_connection_id() const noexcept {
         return _id;
     }
+    stats& get_stats_internal() noexcept {
```
tchaikov, line 351: thought it was a private, but it's actually a `public`...

xemul: It's for `wait_for_reply` templatized helper

tchaikov: i see. no wonder it's would be more convoluted if we wanted to mark it as a `friend`.

xemul:
@scylladb/seastar-maint , merge ping

Issue #1732, opened 2023-07-12
Server-side streams are not deregistered on parent socket abort


xemul:
In client connection:
```
          return stop_send_loop(ep).then_wrapped([this] (future<> f) {
              f.ignore_ready_future();
              _outstanding.clear();
              if (is_stream()) {
                  deregister_this_stream();
              } else {
                  abort_all_streams();
              }
```
and the `abort_all_streams()` aborts the whole `_streams` list

But in server connection

```
          return stop_send_loop(ep).then_wrapped([this] (future<> f) {
              f.ignore_ready_future();
              get_server()._conns.erase(get_connection_id());
              if (is_stream()) {
                  return deregister_this_stream();
              } else {
                  return make_ready_future<>();
              }
```

and the `_streams` list remains intact

@gleb-cloudius , are server-side streams supposed to clean themselves somehow else, or is it a bug?

gleb-cloudius:
> @gleb-cloudius , are server-side streams supposed to clean themselves somehow else, or is it a bug?

What is the point to do a cleanup from both sides (client and server)? If client connection fails it is enough to abort the streams on a client side (the fact that parent connection failure causes streams created through it to be aborted as well is out artificial requirement anyway).

Shouldn't such things be discussed on the mailing list before an issue is created?

mykaul:
@xemul - what's the next step here?

xemul:
Review and merge #1757 , @mykaul

PR #1757, opened 2023-07-24
rpc: Fix client-side stream registration race

xemul:
Client-side stream socket is made by client::make_stream_sink() call. The stream creation consists of

- wait for "this" to finish negotiation
- make new client (this issues connect() and negotiation in the background)
- wait for the new socket negotiation to finish
- .then([] { register new socket to "this" socket })

The last step is racy. After the new socket negotiation succeeds the parent socket (called "this" above) may receive "peer closed" from the kernel and start demolishing the client. The latter means aborting all streams, that happens in one non-preemptive and non-waiting loop. So in case the new stream wasn't _yet_ registered, it won't be aborted and will wait for the server-side stream to terminate (which also may never happen, see #1732)

### DIFF
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -519,6 +519,9 @@ public:
             c->_parent = this->weak_from_this();
             c->_is_stream = true;
             return c->await_connection().then([c, this] {
+                if (_error) {
+                    throw closed_error();
+                }
                 xshard_connection_ptr s = make_lw_shared(make_foreign(static_pointer_cast<rpc::connection>(c)));
                 this->register_stream(c->get_connection_id(), s);
                 return sink<Out...>(make_shared<sink_impl<Serializer, Out...>>(std::move(s)));


include/seastar/rpc/rpc.hh
```diff
@@ -519,6 +519,9 @@ public:
             c->_parent = this->weak_from_this();
             c->_is_stream = true;
             return c->await_connection().then([c, this] {
+                if (_error) {
+                    throw std::runtime_error("Parent socket is closing");
```
gleb-cloudius, line 523: Shouldn't we throw `rpc::closed_error` instead?

xemul: Yes, why not. Fixed in v2

gleb-cloudius (approved):

mykaul:
@gleb-cloudius - can you please re-review?
@avikivity - can you review?

gleb-cloudius:
> @gleb-cloudius - can you please re-review?

I do not see any changes after I approved it last week.

bhalevy:
@avikivity ping review

mykaul:
@avikivity - re-ping.

PR #1758, opened 2023-07-25
build: do not mark mtls_ca.srl as DEPENDS

tchaikov:
after 18f43a0c7b, we always generate the certs created by `mtls_certs` target for testing even if they already exist. this is annoying. and can be misleading sometimes. as in general, a build system does not need to rebuild its targets if the targets are considered up-to-date.

in 18f43a0c7b, we add mtls_ca.srl as a DEPENDS of the `add_custom_command()` calls which signs the x.509 certs. quoting the part of the commit message:

> Also add the implicit dependency into the command declarations
> this doesn't help much, since the dependency is an input/output
> file which can't be expressed by the build system, but helps
> make the file visible.

and quoting the relative part in OPENSSL-X509(1):

> -CAcreateserial
>     With this option and the -CA option the CA serial number file
>     is created if it does not exist.  A random number is generated,
>     used for the certificate, and saved into the serial number file
>     determined as described above.

so strictly speaking the `openssl x509` command does not depend on this serial number file in the sense that it can complete without it.

in this this change,

* these new DEPENDS are removed from the the `add_custom_command()` calls.
* to preserve the connection in between the serial number file, the commands which update/generate it and the job pool which enforce the serialization of these commands calls, the commend above the `set_property()` call adding the `openssl_serial` pool is revised.


tests/unit/CMakeLists.txt
```diff
@@ -566,14 +566,16 @@ function(seastar_gen_mtls_certs)
...
+  # They read and write the serial number file (mtls_ca.srl in this case).
+  # These can't be executed in parallel.
   set_property(GLOBAL APPEND PROPERTY JOB_POOLS openssl_serial=1)
 
   add_custom_command(OUTPUT mtls_ca.key
     COMMAND ${OPENSSL} ecparam -name prime256v1 -genkey -noout -out mtls_ca.key
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
   )
-  add_custom_command(OUTPUT mtls_ca.crt mtls_ca.srl
+  add_custom_command(OUTPUT mtls_ca.crt
```
avikivity, line 578: Maybe this is right and maybe this is wrong, one thing I'm certain is that I don't know how to review it.

How about dropping -CAcreateserial and using -set_serial nnn instead? We can hardcode 1, 2, 3 for the files we create.

avikivity: We can drop the job pool too.

tchaikov: indeed, much better!

tchaikov:
v2:

- add a helper for signing/creating certs
- drop the "openssl_serial" pool

PR #1615, opened 2023-04-14
Make seastar options HUP-updateable

xemul:
Currently all seastar options are immutable after seastar app starts. That's not very convenient as it takes full app restart to pick up new options if needed. Some options, however, allow live-updating because they either just one-to-one map to some simple reactor variable or because live-update can be implemented as a callable.

Seastar can read its options from ini-style seastar.conf file, so what this PR does is makes seastar re-read this file upon receiving HUP signal (*). Options should be marked as live-updateable explicitly by providing a noncopyable function on construction. When config file is re-read all those functions are called with new values arguments.

As an example the --unsafe-bypass-fsync is turned into live-updateable option.

(*) Scylla re-installs HUP handler, so it'll have to call reread_config() by hand.

PR #1761, opened 2023-07-28
app-template: allow reserving -h for the application

denesb:
Normally, -h is the short-form of --help. Some applications however may wish to reserve -h as the short-form of another option. Allow this via a new option: app_template::seastar_options::reserve_dash_h. When set to true, seastar will only register --help, without a short-form.

### DIFF
--- a/include/seastar/core/app-template.hh
+++ b/include/seastar/core/app-template.hh
@@ -106,6 +106,13 @@ public:
         /// You can adjust the behavior of SIGINT/SIGTERM by installing signal handlers
         /// via reactor::handle_signal().
         bool auto_handle_sigint_sigterm = true;
+        /// \brief Reserve -h for the application
+        ///
+        /// Normally, -h is the short-form for --help.
+        /// Some applications however might want to reserve -h for their own use.
+        /// Set this option to true, to reserve -h for the application. In this
+        /// case, --help will not have a short-form option.
+        bool reserve_dash_h = false;
         /// Configuration options for the reactor.
         reactor_options reactor_opts;
         /// Configuration for the metrics sub-system.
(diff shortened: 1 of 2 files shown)


include/seastar/core/app-template.hh
```diff
@@ -106,6 +106,13 @@ public:
         /// You can adjust the behavior of SIGINT/SIGTERM by installing signal handlers
         /// via reactor::handle_signal().
         bool auto_handle_sigint_sigterm = true;
+        /// \brief Reserve -h for the application
+        ///
+        /// Normally, -h is the short-form for --help.
+        /// Some applications however might want to reserve -h for their own use.
+        /// Set this option to true, to reserve -h for the application. In this
+        /// case, --help will not have a short-form option.
+        bool reserve_dash_h = false;
```
xemul, line 115: This flag supports [Avi's concern](https://github.com/scylladb/seastar/pull/1615#issuecomment-1653089501) about design error regarding options parsing by seastar itself. Maybe the better direction is in all-or-nothing approach when app-template refuses to set up any built-in options and lets the application do it

denesb: Right, this approach doesn't scale. What if another app wants `-c`. Or another wants `-m`.  I think the right approach is another overload of `app_template::run()` which doesn't take `argc` and `argv`, and instead leaves arg parsing to the application.

avikivity: Well, app_template should continue to support the current status-quo (we need it for internal applications like httpd, and it is an easy way to start using). We should also expose the current bpo support as building blocks so an application that wants to continue torturing itself with bpo (or has simply realized there is no good alternative) can continue to do so. But it should be driven by the application, instead of driven by the framework with lots of APIs to adjust the behavior.

PR #1762, opened 2023-07-31
[RFC] core/app-template: add run() overloads which do no option parsing

denesb:
This PR adds two new run() variants (one with future<int>, another with future<> return type), which are inteded to be used for applications which want full control on how command line options are exposed, parsed, handled and propagated to seastar. Seastar will not do any command-line option exposing or parsing, nor will it look for config files.


denesb:
RFC because I only compile-tested this so far. Also, I want to see the reactions to this "radical" approach.

denesb:
I opened this in place of https://github.com/scylladb/seastar/pull/1761, as a much more complete approach.

avikivity:
We should deprecate all the other approaches and keep only this one.

avikivity:
Explanation: Seastar shouldn't be in the business of parsing command line options. It's only a library. The applications should parse the options, fill config structs, and pass them to Seastar.

Now that raises the barrier to entry to writing a seastar application by quite a lot, so we can have a helper function that, given boost::program_options stuff, adds Seastar options and spits out computed config structs. So there's a one-liner that adds all the standard Seastar options if you're willing to live with boost::program_options.

xemul:
> RFC because I only compile-tested this so far. Also, I want to see the reactions to this "radical" approach.

That's the way to go from my pov, but not only argv parsing should be eliminated, but also the seastar.conf reading.
And there's also the `--io-properties-file` option that makes seastar parse one more config file. It's worth considering too.

include/seastar/core/app-template.hh
```diff
@@ -172,6 +172,25 @@ public:
     // with exit code 0 when the future returned by func resolves
     // successfully.
     int run(int ac, char ** av, std::function<future<> ()>&& func) noexcept;
+
+    // A variant of run(), which leaves all option parsing entirely to the application.
+    //
+    // When this overload is used, seastar won't parse the config file, nor will
+    // it expose or any command-line option of its own. All confiuration is left
```
avikivity, line 179: spelling: configuration

denesb:
> Now that raises the barrier to entry to writing a seastar application by quite a lot, so we can have a helper function that, given boost::program_options stuff, adds Seastar options and spits out computed config structs. So there's a one-liner that adds all the standard Seastar options if you're willing to live with boost::program_options.

We already have this helper code in seastar:
* config struct -> bpo description
* bpo value map -> config struct

We just didn't make them public because we didn't want to make seastar a config parser lib either. Maybe we can move these to a demo app, so the code can be copy-pasted, fast-tracking the creation of a new seastar app.

src/core/app-template.cc
```diff
@@ -270,4 +276,12 @@ app_template::run(int ac, char ** av, std::function<future<int> ()>&& func) noex
     return exit_code;
 }
 
+int app_template::run(std::string_view argv0, std::function<future<>()>&& func) noexcept {
```
avikivity, line 279: We should move this argv0 to the config struct too, it's not special.

avikivity:
> > RFC because I only compile-tested this so far. Also, I want to see the reactions to this "radical" approach.
> 
> That's the way to go from my pov, but not only argv parsing should be eliminated, but also the seastar.conf reading. And there's also the `--io-properties-file` option that makes seastar parse one more config file. It's worth considering too.

Agree, but we should provide helpers for everything.

Framework: does everything for you, needs intrusive hooks for any change
Library: does nothing for you, need to call library-supplied helper if you're lazy

denesb:
> > That's the way to go from my pov, but not only argv parsing should be eliminated, but also the seastar.conf reading.
> 
> This is already eliminated by this PR.
> 
> > And there's also the `--io-properties-file` option that makes seastar parse one more config file. It's worth considering too.
> 
> I forgot about this. I will have to check how difficult this is to eliminate.

```c++
  3     /// Path to a YAML file describing the characteristics of the I/O Subsystem.
  2     program_options::value<std::string> io_properties_file;
  1     /// A YAML string describing the characteristics of the I/O Subsystem.
73      program_options::value<std::string> io_properties;
```

We have these two options, so the application has the option to bypass the file parsing already.

xemul:
> We have these two options, so the application has the option to bypass the file parsing already.

The 2nd option is YAML file encoded into a string, so it's an option to bypass reading a file, but not bypass YAML parsing. It's all in `disk_config_params::parse_config()`

xemul:
> > > RFC because I only compile-tested this so far. Also, I want to see the reactions to this "radical" approach.
> > 
> > 
> > That's the way to go from my pov, but not only argv parsing should be eliminated, but also the seastar.conf reading. And there's also the `--io-properties-file` option that makes seastar parse one more config file. It's worth considering too.
> 
> Agree, but we should provide helpers for everything.
> 
> Framework: does everything for you, needs intrusive hooks for any change Library: does nothing for you, need to call library-supplied helper if you're lazy

Screenshot from the https://seastar.io homepage

image

avikivity:
> > > > RFC because I only compile-tested this so far. Also, I want to see the reactions to this "radical" approach.
> > > 
> > > 
> > > That's the way to go from my pov, but not only argv parsing should be eliminated, but also the seastar.conf reading. And there's also the `--io-properties-file` option that makes seastar parse one more config file. It's worth considering too.
> > 
> > 
> > Agree, but we should provide helpers for everything.
> > Framework: does everything for you, needs intrusive hooks for any change Library: does nothing for you, need to call library-supplied helper if you're lazy
> 
> Screenshot from the https://seastar.io homepage
> 
> image

Yes, and it will continue to be a framework. It forces you to write all your code in its style.

It will be a little less framework-y and a little more librarylike, but still overall a framework. But at least it will require less coordinated patching of the application and framework in some areas.

PR #1764, opened 2023-08-02
install-dependencies: add centos9 dependencies

Matan-B:
gcc-toolset-11 is not available in centos9.

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -131,6 +131,17 @@ centos8_packages=(
     gcc-toolset-11-libatomic-devel
 )
 
+centos9_packages=(
+    "${redhat_packages[@]}"
+    ninja-build
+    ragel
+    gcc-toolset-13-gcc
+    gcc-toolset-13-gcc-c++
+    gcc-toolset-13-libubsan-devel
+    gcc-toolset-13-libasan-devel
+    gcc-toolset-13-libatomic-devel
+)
+
 # 1) glibc 2.30-3 has sys/sdt.h (systemtap include)
 #    some old containers may contain glibc older,
 #    so enforce update on that one.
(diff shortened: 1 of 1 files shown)


mykaul:
I thought we were looking at GCC 13, which is available at CentOS Stream 9. @tchaikov ?

tchaikov:
Yeah, GTS-13 was already packaged for CentOS 9 stream. see https://centos.pkgs.org/9-stream/centos-appstream-x86_64/gcc-toolset-13-13.0-1.el9.x86_64.rpm.html 

@Matan-B hi Matan, why do you want to install GCC-12 instead of GCC-13?

Matan-B:
> Yeah, GTS-13 was already packaged for CentOS 9 stream. see https://centos.pkgs.org/9-stream/centos-appstream-x86_64/gcc-toolset-13-13.0-1.el9.x86_64.rpm.html
> 
> @Matan-B hi Matan, why do you want to install GCC-12 instead of GCC-13?

I was simply not aware :)
Updated, thanks!

install-dependencies.sh
```diff
@@ -131,6 +131,17 @@ centos8_packages=(
     gcc-toolset-11-libatomic-devel
 )
 
+centos9_packages=(
+    "${redhat_packages[@]}"
+    ninja-build
+    ragel
+    gcc-toolset-13-gcc
+    gcc-toolset-13-gcc-c++
```
tchaikov, line 139: i think `gcc-toolset-13-gcc-c++` depends on [gcc-toolset-13-gcc](https://pkgs.org/download/gcc-toolset-13-gcc). and as a C++ framework, Seastar should be built with C++. any reason to install gcc explicitly?

Matan-B: The same is true with [gcc-toolset-11-gcc-c++](https://almalinux.pkgs.org/8/almalinux-appstream-aarch64/gcc-toolset-11-gcc-c++-11.2.1-9.1.el8.aarch64.rpm.html). `centos8_packages` should be edited as well then, right?

tchaikov: we can include that change in another PR.

Hi @JunHe77 i just realized that you added `gcc-toolset-9-gcc` back in https://github.com/scylladb/seastar/commit/627775dfa94bb4fa8e277437239b33a06dedf946#diff-4852d2a5a482159af418383e38099c1321b69c6272981405ed32d629a19de73dR123 . do you happen to remember why that package was added?

PR #1765, opened 2023-08-02
future: add forward_to(shared_promise&) helpers

bhalevy:
Provide helpers to forward a future value/exception to a shared_promise.

Otherwise the caller needs to handle the ready_future
in a then_wrapped continuation and call set_value
or set_exception appropriately.

### DIFF
--- a/include/seastar/core/shared_future.hh
+++ b/include/seastar/core/shared_future.hh
@@ -370,6 +370,35 @@ public:
     }
 };
 
+SEASTAR_MODULE_EXPORT
+template <typename T>
+template <typename... U>
+SEASTAR_CONCEPT( requires (!std::same_as<T, void> && std::same_as<std::tuple<std::remove_cv_t<U>...>, std::tuple<T>>) )
+std::enable_if_t<!std::is_same_v<T, void> && std::is_same_v<std::tuple<std::remove_cv_t<U>...>, std::tuple<T>>, void> future<T>::forward_to(shared_promise<U...>& pr) noexcept {
+    // Discard future. Caller must await pr.get_shared_future();
+    (void)then_wrapped([&pr] (auto f) {
+        if (f.failed()) {
+            pr.set_exception(f.get_exception());
+        } else {
+            pr.set_value(f.get());
+        }
+    });
+}
+
+template <typename T>
+template <typename... U>
+SEASTAR_CONCEPT( requires (std::same_as<T, void> && std::same_as<std::tuple<std::remove_cv_t<U>...>, std::tuple<>>) )
+std::enable_if_t<std::is_same_v<T, void> && std::is_same_v<std::tuple<std::remove_cv_t<U>...>, std::tuple<>>, void> future<T>::forward_to(shared_promise<U...>& pr) noexcept {
+    // Discard future. Caller must await pr.get_shared_future();
+    (void)then_wrapped([&pr] (auto f) {
+        if (f.failed()) {
+            pr.set_exception(f.get_exception());
+        } else {
+            pr.set_value();
+        }
+    });
+}
+
 /// @}
 
 }
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -155,6 +155,9 @@ class future;
 template <typename... T>
 class shared_future;
 
+template <typename... T>
+class shared_promise;
+
 struct future_state_base;
 
 /// \brief Creates a \ref future in an available, value state.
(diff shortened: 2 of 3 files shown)


bhalevy:
cc @Deexie

include/seastar/core/shared_future.hh
```diff
@@ -370,6 +370,35 @@ public:
...
+template <typename... U>
+SEASTAR_CONCEPT( requires (!std::same_as<T, void> && std::same_as<std::tuple<std::remove_cv_t<U>...>, std::tuple<T>>) )
+std::enable_if_t<!std::is_same_v<T, void> && std::is_same_v<std::tuple<std::remove_cv_t<U>...>, std::tuple<T>>, void> future<T>::forward_to(shared_promise<U...>& pr) noexcept {
+    // Discard future. Caller must await pr.get_shared_future();
+    (void)then_wrapped([&pr] (auto f) {
+        if (f.failed()) {
+            pr.set_exception(f.get_exception());
+        } else {
+            pr.set_value(f.get());
+        }
```
denesb, line 384: Regular future's `forward_to()` has a much more efficient implementation, can we reuse the same ideas here?

bhalevy: Unfortunately, I don't think it's possible (or easy enough) since the `shared_state` isn't compatible with `future_state_base`.

denesb: Ok, I still think the patch provides value.

denesb (approved):

bhalevy:
Ref https://github.com/scylladb/seastar/issues/1645 for the CI failure

bhalevy:
@denesb how to respin CI?

avikivity:
Some time ago I reached the conclusion that shared_promise isn't really necessary and shared_future is all that's needed. Since shared_future can be copied, and can be converted to a future, it acts as a promise too.

denesb:
> @denesb how to respin CI?

I have no idea.

xemul:
@bhalevy , where do you want to use it? I bet the code is going to be better if replacing shared_promise with regular promise, rather than utilizing forward_to(shared _promise)

bhalevy:
> @bhalevy , where do you want to use it? I bet the code is going to be better if replacing shared_promise with regular promise, rather than utilizing forward_to(shared _promise)

@xemul see https://github.com/scylladb/scylladb/pull/14913 for a use case.  There it's for replacing a shared_future with a shared_promise.

bhalevy:
> I see. Well, the pattern
> 
> ```
>     create an object and get "done" shared future from it
>     ...
>     start executing something later and forward the result into "done"
> ```
> 
> is a bit dangerous, because while the `...` thing happens something can go wrong and execution never starts, so we shouldn't forget to resolve the waiters' futures. And it looks like that PR is missing it

This PR is dealing only with "start executing something later and forward the result into "done""
it's the caller responsibility to set an exception on the shared promise if "..." failed.
Why would a low-level mechanism like `forward_to` want to deal with that?

xemul:
> Why would a low-level mechanism like forward_to want to deal with that?

It shouldn't, of course. My point is that this mechanism is ... so neat and handy to be used for that pattern, that the user is "encouraged" to forget about this pitfall.

The shared_promise<> is a wrapper of promise<> and shared_future<> in one pack. Were these two used explicitly separately, without the help of shared_promise<>, the need to abort the promise part on error would be obvious

bhalevy:
> > Why would a low-level mechanism like forward_to want to deal with that?
> 
> It shouldn't, of course. My point is that this mechanism is ... so neat and handy to be used for that pattern, that the user is "encouraged" to forget about this pitfall.
> 
> The shared_promise<> is a wrapper of promise<> and shared_future<> in one pack. Were these two used explicitly separately, without the help of shared_promise<>, the need to abort the promise part on error would be obvious

That's a valid point.
It looks like if we expose `promise_type& shared_promise::promise()` we could use forward_to directly to that promise.

bhalevy:
> It looks like if we expose `promise_type& shared_promise::promise()` we could use forward_to directly to that

No, sorry, `future::forward_to` requires a `promise&&`

xemul:
No, this one https://groups.google.com/g/seastar-dev/c/ijHOqieWHxI/m/ZFK3dlbWBgAJ

xemul:
And this -- https://groups.google.com/g/seastar-dev/c/SEs26Rpw5uw/m/jXqHKeTIAwAJ -- an attempt to "auto-convert" shared promises into regular

bhalevy:
Closing this PR as instead of using a shared_promise in scylla's use case we kept the shared_future as used of regular promise (with `forward_to(std::move(_promise))`) to forward the result of the compaction task we wanted to track.

PR #1492, opened 2023-02-10
Add "IO tick violation time" metrics

xemul:
IO scheduler math is built around the assumption that reactor polls it once per io-latency-goal or more frequently. If reactor fails to do it, disk may become underloaded, IO queues would accumulate over time. Another nasty side effect is that rare polls in seastar non-preemptive model result in phantom jams in IO requests flow.

The new metrics is supposed to help observing this effect takes place.

refs: #1311

Issue #1311, opened 2022-11-28
Disk stall detector


xemul:
Similarly to cpu stall detector that prints warnings and calltraces when reactor doesn't tick when it needs, it would be nice to have IO stall detector that warns us about disk being slow.

`scylla_io_queue_total_exec_sec` metrics exists, but short spikes still can come under the radar

bhalevy:
@xemul what else needs to be done here other than #1492?

xemul:
I've the branch already :) Need to polish it and push

mykaul:
With the issues we've seen with some customer setup, I wonder if this could be helpful? If so, @xemul - please try to push this (so it'll get into 5.3).

xemul:
> With the issues we've seen with some customer setup, I wonder if this could be helpful?

This was implemented this way _exactly_ after looking at the issue with "some customer"'s cluster

PR #1766, opened 2023-08-02
Slowdown IO scheduler based on dispatched/completed ratio

xemul:
There are three places where IO dispatch loop is throttled

  * self-throttling with token bucket according to math model
  * per-shard one-tick threshold
  * 2-bucket approach when tokens are replenished only after they are released from disk

This PR removes the last one, because it leads to self-slowdown in case of reactor stalls. This back-link was introduced to catch the case when the disk suddenly slows down to stop dispatched to over-load it with requests, but effectively this back-link measures not the real disk dispatch rate, but the disk+kernel+reactor dispatch rate. Despite the "kernel" part is tiny, the reactor part can grow large triggering the self slow-down effect.

Here's some math.

Let's assume that a some point scheduler dispatched N_d requests. It means that it was able to grab N_d tokens in T_d duration, the rate of dispatch is R_d = N_d/T_d. The requests are to be completed by the reactor next tick. Let's assume it takes T_c time until reactor gets there and it completes N_c requests. The rate of completion is thus R_c = N_c/T_c. Apparently, N_c <= N_d, because kernel cannot complete more requests that it was queued.

In case reactor experiences a stall during the completion tick, T_c > T_d and since N_c <= N_d consequentially N_d/T_d > N_c/T_c. In case reactor doesn't stall, the number of requests that will complete N_c = N_d/T_d * T_c, because this is how dispatch rate is defined. This is equivalent to N_c/T_c = N_d/T_d.

Finally: R_d >= R_c i.e. the dispatch rate is equal of greater than the completion rate where the "equal" part is less likely and is only if reactor clockworks and doesn't stall.

The mentioned back-link makes sure that R_d <= R_c, coupled with the stalls (even the small ones) this drives the R_d down each tick, causing the R_c to go down as well, then again.

The removed fuse is replaced with the flow-monitor based on dispatch-to-completion rate. Normally, the number of requests dispatched for a certain duration divided by the number of requests completed for the same duration must be 1.0. Otherwise that would mean that requests accumulate in disk. However, this ratio cannot be such immediately and in the longer run it tends to be slightly greater that 1.0, because if reactor polls kernel for IO completions more often, it won't get more requests that it was dispatched. But even a small delay in polling would make Nr_completed / duration less because of the larger denominator value.

Having said that, the new backlink is based on the flow-ratio. When the "average" value of dispatched/completed rates exceeds some threshold (configurable, 1.5 by default) the "cost" of individual requests increases thus reducing the dispatch rate.

The main difference from the current implementation is that the new backlink is not "immediate". The averaging is the exponential moving average filter with 100ms updates and 0.95 smoothing factor. Current backlink is immediate in a sense that delay to deliver a completion immediately slows down the next tick dispatch thus accumulating spontaneous reactor micro-stalls.

This can be reproduced by the test introduced in #1724 . It's not (yet) in the PR, but making the tokens release loop artificially release ~1% more tokens fixes this case, which also supports the theory of reduced completion rate being the culprit. BTW, it cannot be the fix, because the ... over-release factor is not constant and it's hard to calculate it.

fixes: #1641 
refs: #1311 
refs: #1492 (*) in fact, _this_ is the metrics that correlates with the flow ratio to grow above 1.0, but this metrics is sort of look at quota-violation from the IO angle
refs: #1774 this PR has attached metrics screenshots demonstrating the effect on stressed scylla

### DIFF
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -114,6 +114,8 @@ private:
     // decoupling and is temporary
     size_t _queued_requests = 0;
     size_t _requests_executing = 0;
+    uint64_t _requests_dispatched = 0;
+    uint64_t _requests_completed = 0;
 public:
 
     using clock_type = std::chrono::steady_clock;
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -157,6 +157,7 @@ public:
         size_t block_count_limit_min = 1;
         unsigned flow_ratio_ticks = 100;
         double flow_ratio_ema_factor = 0.95;
+        double flow_ratio_backpressure_threshold = 1.1;
     };
 
     io_queue(io_group_ptr group, internal::io_sink& sink);
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -33,6 +33,7 @@
 #include <seastar/core/metrics_registration.hh>
 #include <seastar/core/future.hh>
 #include <seastar/core/internal/io_request.hh>
+#include <seastar/core/lowres_clock.hh>
 #include <seastar/util/spinlock.hh>
 #include <seastar/util/modules.hh>
 
@@ -116,6 +117,15 @@ private:
     size_t _requests_executing = 0;
     uint64_t _requests_dispatched = 0;
     uint64_t _requests_completed = 0;
+
+    // Flow monitor
+    uint64_t _prev_dispatched = 0;
+    uint64_t _prev_completed = 0;
+    double _flow_ratio = 1.0;
+    timer<lowres_clock> _flow_ratio_update;
+
+    void update_flow_ratio() noexcept;
+
 public:
 
     using clock_type = std::chrono::steady_clock;
@@ -144,6 +154,8 @@ public:
         float rate_factor = 1.0;
         std::chrono::duration<double> rate_limit_duration = std::chrono::milliseconds(1);
         size_t block_count_limit_min = 1;
+        unsigned flow_ratio_ticks = 100;
+        double flow_ratio_ema_factor = 0.95;
     };
 
     io_queue(io_group_ptr group, internal::io_sink& sink);
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -126,6 +126,7 @@ private:
 
     void update_flow_ratio() noexcept;
 
+    metrics::metric_groups _metric_groups;
 public:
 
     using clock_type = std::chrono::steady_clock;
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -204,6 +204,8 @@ public:
     ~io_group();
     struct priority_class_data;
 
+    std::chrono::duration<double> io_latency_goal() const noexcept;
+
 private:
     friend class io_queue;
     friend struct ::io_queue_for_tests;
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -539,6 +539,7 @@ const fair_group& get_fair_group(const io_queue& ioq, unsigned stream) {
 void
 io_queue::complete_request(io_desc_read_write& desc) noexcept {
     _requests_executing--;
+    _requests_completed++;
     _streams[desc.stream()].notify_request_finished(desc.capacity());
 }
 
@@ -1008,6 +1009,7 @@ void io_queue::poll_io_queue() {
 void io_queue::submit_request(io_desc_read_write* desc, internal::io_request req) noexcept {
     _queued_requests--;
     _requests_executing++;
+    _requests_dispatched++;
     _sink.submit(desc, std::move(req));
 }
 
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -578,10 +578,8 @@ fair_group::config io_group::make_fair_group_config(const io_queue::config& qcfg
     return cfg;
 }
 
-static void maybe_warn_latency_goal_auto_adjust(const fair_group& fg, const io_queue::config& cfg) noexcept {
-    auto goal = fg.rate_limit_duration();
-    auto lvl = goal > 1.1 * cfg.rate_limit_duration ? log_level::warn : log_level::debug;
-    seastar_logger.log(lvl, "IO queue uses {:.2f}ms latency goal for device {}", goal.count() * 1000, cfg.devid);
+std::chrono::duration<double> io_group::io_latency_goal() const noexcept {
+    return _fgs.front()->rate_limit_duration();
 }
 
 io_group::io_group(io_queue::config io_cfg, unsigned nr_queues)
@@ -590,11 +588,13 @@ io_group::io_group(io_queue::config io_cfg, unsigned nr_queues)
 {
     auto fg_cfg = make_fair_group_config(_config);
     _fgs.push_back(std::make_unique<fair_group>(fg_cfg, nr_queues));
-    maybe_warn_latency_goal_auto_adjust(*_fgs.back(), io_cfg);
     if (_config.duplex) {
         _fgs.push_back(std::make_unique<fair_group>(fg_cfg, nr_queues));
-        maybe_warn_latency_goal_auto_adjust(*_fgs.back(), io_cfg);
     }
+
+    auto goal = io_latency_goal();
+    auto lvl = goal > 1.1 * io_cfg.rate_limit_duration ? log_level::warn : log_level::debug;
+    seastar_logger.log(lvl, "IO queue uses {:.2f}ms latency goal for device {}", goal.count() * 1000, io_cfg.devid);
 
     /*
      * The maximum request size shouldn't result in the capacity that would
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -920,7 +920,16 @@ double internal::request_tokens(io_direction_and_length dnl, const io_queue::con
 }
 
 fair_queue_entry::capacity_t io_queue::request_capacity(io_direction_and_length dnl) const noexcept {
-    return _streams[request_stream(dnl)].tokens_capacity(internal::request_tokens(dnl, get_config()));
+    const auto& cfg = get_config();
+    auto tokens = internal::request_tokens(dnl, cfg);
+    if (_flow_ratio <= cfg.flow_ratio_backpressure_threshold) {
+        return _streams[request_stream(dnl)].tokens_capacity(tokens);
+    }
+
+    auto stream = request_stream(dnl);
+    auto cap = _streams[stream].tokens_capacity(tokens * _flow_ratio);
+    auto max_cap = _streams[stream].maximum_capacity();
+    return std::min(cap, max_cap);
 }
 
 io_queue::request_limits io_queue::get_request_limits() const noexcept {
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -536,6 +536,20 @@ const fair_group& get_fair_group(const io_queue& ioq, unsigned stream) {
 
 } // internal namespace
 
+template <typename T>
+void update_moving_average(T& result, T value, double factor) noexcept {
+    result = result * factor + value * (1.0 - factor);
+}
+
+void io_queue::update_flow_ratio() noexcept {
+    if (_requests_completed > _prev_completed) {
+        auto instant = double(_requests_dispatched - _prev_dispatched) / double(_requests_completed - _prev_completed);
+        update_moving_average(_flow_ratio, instant, get_config().flow_ratio_ema_factor);
+        _prev_dispatched = _requests_dispatched;
+        _prev_completed = _requests_completed;
+    }
+}
+
 void
 io_queue::complete_request(io_desc_read_write& desc) noexcept {
     _requests_executing--;
@@ -553,6 +567,7 @@ io_queue::io_queue(io_group_ptr group, internal::io_sink& sink)
     : _priority_classes()
     , _group(std::move(group))
     , _sink(sink)
+    , _flow_ratio_update([this] { update_flow_ratio(); })
 {
     auto& cfg = get_config();
     if (cfg.duplex) {
@@ -563,6 +578,7 @@ io_queue::io_queue(io_group_ptr group, internal::io_sink& sink)
     } else {
         _streams.emplace_back(*_group->_fgs[0], make_fair_queue_config(cfg, "rw"));
     }
+    _flow_ratio_update.arm_periodic(std::chrono::duration_cast<std::chrono::milliseconds>(group->io_latency_goal() * cfg.flow_ratio_ticks));
 }
 
 fair_group::config io_group::make_fair_group_config(const io_queue::config& qcfg) noexcept {
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -579,6 +579,16 @@ io_queue::io_queue(io_group_ptr group, internal::io_sink& sink)
         _streams.emplace_back(*_group->_fgs[0], make_fair_queue_config(cfg, "rw"));
     }
     _flow_ratio_update.arm_periodic(std::chrono::duration_cast<std::chrono::milliseconds>(group->io_latency_goal() * cfg.flow_ratio_ticks));
+
+    namespace sm = seastar::metrics;
+    auto owner_l = sm::shard_label(this_shard_id());
+    auto mnt_l = sm::label("mountpoint")(mountpoint());
+    auto group_l = sm::label("iogroup")(to_sstring(_group->_allocated_on));
+    _metric_groups.add_group("io_queue", {
+        sm::make_gauge("flow_ratio", [this] { return _flow_ratio; },
+                sm::description("Ratio of dispatch rate to completion rate. Is expected to be 1.0+ growing larger on reactor stalls or (!) disk problems"),
+                { owner_l, mnt_l, group_l }),
+    });
 }
 
 fair_group::config io_group::make_fair_group_config(const io_queue::config& qcfg) noexcept {
--- a/doc/io-scheduler.md
+++ b/doc/io-scheduler.md
@@ -1,0 +1,117 @@
+IO scheduler uses rate-limiter to throttle the amount of data it dispatches into the disk.
+
+# Basic math
+
+The scheduler's main equation that models the disk behavior is
+
+$$ \frac {bandwidth_r} {bandwidth_{r_{max}}} +
+   \frac {bandwidth_w} {bandwidth_{w_{max}}} +
+   \frac {iops_r} {iops_{r_{max}}} +
+   \frac {iops_w} {iops_{w_{max}}} \le 1.0 $$
+
+Let's say that
+
+$$ m_b = \frac {bandwidth_{r_{max}}} {bandwidth_{w_{max}}} \ \ m_o = \frac {iops_{r_{max}}} {iops_{w_{max}}} $$
+
+Since
+
+$$ bandwidth_x = \frac {d}{dt} bytes_x \ \ iops_x = \frac {d}{dt} ios_x $$
+
+The main equation turns into
+
+$$ \frac {d}{dt} \( \frac {bytes_r + m_b \times bytes_w} {bandwidth_{r_{max}}} + \frac {ios_r + m_o \times ios_w} {iops_{r_{max}}} \) \le 1.0 $$
+
+Requests are asigned a 2d value of _{1, bytes}_ for reads and _{m<sub>o</sub>, m<sub>b</sub> * bytes}_ for writes called "tickets"
+
+The "normalization" operation is defined as
+
+$$ N(ticket) = \frac {ticket_0}{iops_{r_{max}}} + \frac {ticket_1}{bandwidth_{r_{max}}} $$
+
+With that the main equation turns into 
+
+$$ \frac {d}{dt} \sum_{ticket} N(ticket) \le 1.0 $$
+
+The time-derivative limitation is then implemented with the token bucket algorithm
+
+# Token bucket
+
+The algorithm creates token bucket with the refill rate of _1.0_ and each request wants to carry the fractional token value of _tokens = N(ticket)_ with _ticket_ defined above
+
+The bucket additionally requires the _limit_ parameter which is the maximum number of tokens the bucket may hold. This value is calculated using the _io_latency_goal_ parameter, to be the amount of tokens accumulated for the _io_latency_goal_ duration
+
+# Slowing down the flow
+
+Let's assume we need to reduce the rate of request run-time. This means that we want to
+
+$$ \frac {d}{dt} \( \frac {bytes_r + m_b \times bytes_w} {bandwidth_{r_{max}} \times \alpha} + \frac {ios_r + m_o \times ios_w} {iops_{r_{max}} \times \beta} \) \le 1.0 \ \  \alpha \le 1.0 \ \beta \le 1.0 $$
+
+There are many ways of selecting which of IOPS or bandwidth to reduce, the "general" slowing down may assume that
+
+$$ \alpha = \beta = \frac {1}{\gamma} , \ \ \ \gamma \ge 1.0 $$
+
+The main equation then turns into
+
+$$ \frac {d}{dt} \sum_{ticket} \gamma \times N(ticket) \le 1.0 $$
+
+which in turn means, that we can just multiply the request cost (in tokens) by some number above 1.0
+
+# Detecting the slowdown
+
+Let's say that
+
+_d<sub>i</sub>_ -- the amount of requests dispatched at tick _i_
+_p<sub>i</sub>_ -- the amount of requests processed by disk at tick _i_
+_c<sub>i</sub>_ -- the amount of requests completed by reactor loop at tick _i_
+
+We can observe _d<sub>i</sub>_ and _c<sub>i</sub>_ in the dispatcher, but _not_ the _p<sub>i</sub>_, because we don't have direct access to disks' queues
+
+After _n_ ticks we have 
+
+_D<sub>n</sub>_ -- total amount of requests dispatched, 
+_P<sub>n</sub>_ -- total amount of requests processed, 
+_C<sub>n</sub>_ -- total amount of requests completed, 
+
+$$ D_n = \sum_{i=0}^n d_i $$
+
+$$ P_n = \sum_{i=0}^n p_i $$
+
+$$ C_n = \sum_{i=0}^n c_i $$
+
+* Disk cannot process more than it was dispatched, but it can process less "accumulating" a queue
+
+$$ d_i \ge p_i   \Rightarrow  D_n \ge P_n$$
+
+* Reactor cannot complete more than it was processed by disk either
+
+$$ p_i \ge c_i \Rightarrow P_n \ge C_n $$
+
+
+Note that _D<sub>n</sub> > P<sub>n</sub>_  means that disk is falling behind. Our goal is to make sure the disk doesn't do it and doesn't accumulate the queue, i.e. _D<sub>n</sub> = P<sub>n</sub>_, but we cannot observe _P<sub>n</sub>_ directly, only _C<sub>n</sub>_.
+
+Next
+
+$$ D_n - P_n = Qd_n \ge 0 $$
+
+$$ P_n - C_n = Qc_n \ge 0 $$
+
+$$ D_n - C_N = Qd_n + Qc_n \ge 0 $$
+
+_Qd<sub>n</sub>_ is the queue accumulated in disk. We try to avoid it. _Qc<sub>n</sub>_ is the accumulated delta between processed (by disk) and completed (by reactor) requests. We cannot reliably say that it goes to zero over time, because there's always some unknown amount of processed but not yet completed requests. So the above formula contains two unknowns and we cannot solve it. Let's try the other way
+
+$$ \frac {D_n} {P_n} = Rd_n \ge 1 $$
+
+$$ \frac {P_n} {C_n} = Rc_n \ge 1 $$
+
+$$ \frac {D_n} {C_n} = Rd_n \times Rc_n \ge 1 $$
+
+_Rd<sub>n</sub>_ is the ratio of dispatched to processed requests. If it's 1, we're OK, if it's greater, disk is accumulating the queue, we try to avoid it. _Rc<sub>n</sub>_ is the ratio between processed (by disk) and completed (by reactor) requests. It has a not immediately apparent, but very pretty property
+
+$$ \lim_{n\to\infty} Rc_n = \lim_{n\to\infty} \frac {P_n} {C_n} = \lim_{n\to\infty} \frac {C_n + Qc_n} { C_n} = \lim_{n\to\infty} \(1 +  \frac {Qc_n} {C_n} \) = 1 + \lim_{n\to\infty} \frac {Qc_n} {C_n}$$
+
+The _Qc<sub>n</sub>_ doesn't grow over time. It's non-zero, but it's upper bound by some value. Respectively
+
+$$ \lim_{n\to\infty} Rc_n = 1 $$
+
+$$ \lim_{n\to\infty} \frac {D_n} {C_n} = \lim_{n\to\infty} \( Rd_n \times Rc_n \) =  \lim_{n\to\infty} Rd_n  $$
+
+IOW -- we can say if the disk is accumulating the queue or not by observing the dispatched-to-completed (to _completed_, not _processed_) over a long enough time 
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -191,7 +191,7 @@ public:
 
     static constexpr float fixed_point_factor = float(1 << 24);
     using rate_resolution = std::milli;
-    using token_bucket_t = internal::shared_token_bucket<capacity_t, rate_resolution, internal::capped_release::yes>;
+    using token_bucket_t = internal::shared_token_bucket<capacity_t, rate_resolution, internal::capped_release::no>;
 
 private:
 
@@ -255,7 +255,6 @@ public:
     capacity_t per_tick_grab_threshold() const noexcept { return _per_tick_threshold; }
     capacity_t grab_capacity(capacity_t cap) noexcept;
     clock_type::time_point replenished_ts() const noexcept { return _token_bucket.replenished_ts(); }
-    void release_capacity(capacity_t cap) noexcept;
     void replenish_capacity(clock_type::time_point now) noexcept;
     void maybe_replenish_capacity(clock_type::time_point& local_ts) noexcept;
 
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -393,6 +393,10 @@ public:
         return _group.tokens_capacity(tokens);
     }
 
+    capacity_t maximum_capacity() const noexcept {
+        return _group.maximum_capacity();
+    }
+
     /// Queue the entry \c ent through this class' \ref fair_queue
     ///
     /// The user of this interface is supposed to call \ref notify_requests_finished when the
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -69,6 +69,14 @@ struct reactor_options : public program_options::option_group {
     ///
     /// Default: 1.5 * task_quota_ms value
     program_options::value<double> io_latency_goal_ms;
+    /// \bried Dispatch rate to completion rate ratio threshold
+    ///
+    /// Describes the worst ratio at which seastar reactor is allowed to delay
+    /// IO requests completion. If exceeded, the scheduler will consider it's
+    /// disk that's the reason for completion slow-down and will scale down
+    ///
+    /// Default: 1.1
+    program_options::value<double> io_flow_ratio_threshold;
     /// \brief Maximum number of task backlog to allow.
     ///
     /// When the number of tasks grow above this, we stop polling (e.g. I/O)
--- a/src/core/fair_queue.cc
+++ b/src/core/fair_queue.cc
@@ -123,10 +123,6 @@ auto fair_group::grab_capacity(capacity_t cap) noexcept -> capacity_t {
     return _token_bucket.grab(cap);
 }
 
-void fair_group::release_capacity(capacity_t cap) noexcept {
-    _token_bucket.release(cap);
-}
-
 void fair_group::replenish_capacity(clock_type::time_point now) noexcept {
     _token_bucket.replenish(now);
 }
@@ -261,10 +257,6 @@ auto fair_queue::grab_pending_capacity(const fair_queue_entry& ent) noexcept -> 
         return grab_result::cant_preempt;
     }
 
-    if (cap < _pending->cap) {
-        _group.release_capacity(_pending->cap - cap); // FIXME -- replenish right at once?
-    }
-
     _pending.reset();
     return grab_result::grabbed;
 }
@@ -330,7 +322,6 @@ void fair_queue::queue(class_id id, fair_queue_entry& ent) noexcept {
 }
 
 void fair_queue::notify_request_finished(fair_queue_entry::capacity_t cap) noexcept {
-    _group.release_capacity(cap);
 }
 
 void fair_queue::notify_request_cancelled(fair_queue_entry& ent) noexcept {
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3818,6 +3818,7 @@ reactor_options::reactor_options(program_options::option_group* parent_group)
                 "busy-poll for disk I/O (reduces latency and increases throughput)")
     , task_quota_ms(*this, "task-quota-ms", 0.5, "Max time (ms) between polls")
     , io_latency_goal_ms(*this, "io-latency-goal-ms", {}, "Max time (ms) io operations must take (1.5 * task-quota-ms if not set)")
+    , io_flow_ratio_threshold(*this, "io-flow-rate-threshold", 1.1, "Dispatch rate to completion rate threshold")
     , max_task_backlog(*this, "max-task-backlog", 1000, "Maximum number of task backlog to allow; above this we ignore I/O")
     , blocked_reactor_notify_ms(*this, "blocked-reactor-notify-ms", 25, "threshold in miliseconds over which the reactor is considered blocked if no progress is made")
     , blocked_reactor_reports_per_minute(*this, "blocked-reactor-reports-per-minute", 5, "Maximum number of backtraces reported by stall detector per minute")
@@ -4047,6 +4048,7 @@ private:
     unsigned _num_io_groups = 0;
     std::unordered_map<dev_t, mountpoint_params> _mountpoints;
     std::chrono::duration<double> _latency_goal;
+    double _flow_ratio_backpressure_threshold;
 
 public:
     explicit disk_config_params(unsigned max_queues) noexcept
@@ -4073,6 +4075,8 @@ public:
         seastar_logger.debug("smp::count: {}", smp::count);
         _latency_goal = std::chrono::duration_cast<std::chrono::duration<double>>(latency_goal_opt(reactor_opts) * 1ms);
         seastar_logger.debug("latency_goal: {}", latency_goal().count());
+        _flow_ratio_backpressure_threshold = reactor_opts.io_flow_ratio_threshold.get_value();
+        seastar_logger.debug("flow-ratio threshold: {}", _flow_ratio_backpressure_threshold);
 
         if (smp_opts.num_io_groups) {
             _num_io_groups = smp_opts.num_io_groups.get_value();
@@ -4157,6 +4161,7 @@ public:
         cfg.duplex = p.duplex;
         cfg.rate_factor = p.rate_factor;
         cfg.rate_limit_duration = latency_goal();
+        cfg.flow_ratio_backpressure_threshold = _flow_ratio_backpressure_threshold;
         // Block count limit should not be less than the minimal IO size on the device
         // On the other hand, even this is not good enough -- in the worst case the
         // scheduler will self-tune to allow for the single 64k request, while it would


avikivity:
I don't understand. What if the disk is 5% slower than the setting in io_properties.yaml. Won't the queue keep growing and latency grow with it?

xemul:
> I don't understand this either. After a reactor stall, all pending requests will complete, because the disk hasn't stalled. The scheduler is in pristine state.

No it isn't. The scheduler assumes that N submitted requests will complete in T duration. They are, but disk, reactor my add "t" duration on top. The resulting completion rate is `N / (T + t) < N / T`

xemul:
> But it could be compensated by increasing io_latency_goal (if throughput is more important than latency) or by fixing the stalls.

When I tell "stall" I don't mean "stalls that are reported". I mean -- reactor poll completion notably _later_ than disk completes the request.

Let's take a disk of 1Gb/s. Per 0.75ms (default io latency goal) it should do 1Gb * 0.75 / 1000 = 786kb/s. Now the scheduler submitted them all in one poll. The expectation is that the token bucket will replenish all 786k tokens in 0.75ms. Now the reactor completes requests in 0.80ms, just 50usec later. The token bucket will replenish ~736k tokens, because _this_ is the completion rate Next tick the scheduler will be able to submit only 736k tokens, which is ~958Mb/s, not 1Gb/s. Next tick it will lose 7% more.

avikivity:
> Yes, it will. My point is that currently we cannot tell "disk is 5% slower than the setting" from "reactor deliver completions a bit later and this looks like disk is 5% slower".

We have metrics for quota violations.

And "a bit" doesn't hurt as long as the io latency goal is large enough.

> And the attempt to slow down the dispatch rate leads to lower completion rate (literally, because less requests are dispatched) which, in turn, slows down the dispatch rate even further.

I don't understand. We should stuff enough requests to fill up the latency goal. Small stalls shouldn't affect it.

xemul:
> I don't understand. We should stuff enough requests to fill up the latency goal. 

We do, but "enough" == according to the number of tokens in a bucket.

> Small stalls shouldn't affect it.

Small stalls make just a bit less tokens into a bucket during replenishment. This bits accumulate and eventually the replenishment rate is notably lower.

avikivity:
> > But it could be compensated by increasing io_latency_goal (if throughput is more important than latency) or by fixing the stalls.
> 
> When I tell "stall" I don't mean "stalls that are reported". I mean -- reactor poll completion notably _later_ than disk completes the request.

That's almost always true. Small reads take ~0.1ms. The task quota is 0.5ms.

> 
> Let's take a disk of 1Gb/s. Per 0.75ms (default io latency goal) it should do 1Gb * 0.75 / 1000 = 786kb/s. Now the scheduler submitted them all in one poll. The expectation is that the token bucket will replenish all 786k tokens in 0.75ms. Now the reactor completes requests in 0.80ms, just 50usec later. The token bucket will replenish ~736k tokens, because _this_ is the completion rate Next tick the scheduler will be able to submit only 736k tokens, which is ~958Mb/s, not 1Gb/s. Next tick it will lose 7% more.

If we regularly have a 0.3ms stall (regularly = almost every poll period), that's bad. If we have such a hiccup every 100 periods, it's unnoticable.

Which is it?

We should increase the io latency goal. 0.75ms is too close to the task quota period.

xemul:
> If we regularly have a 0.3ms stall (regularly = almost every poll period), that's bad. If we have such a hiccup every 100 periods, it's unnoticable.
> 
> Which is it?

It doesn't really matter how frequent 0.3ms stalls are (in fact, the bad clusters I've seen all had ~10% of task-quota violations). If we replace 50usec with 5usec in the above math, the result will be the same, but the scheduler will lose 0.7% of tokens every tick, not 7%. The result is the same, but slower.

> We should increase the io latency goal. 0.75ms is too close to the task quota period.

Every time a request is completed X% later then the math thinks it will, token bucket loses ~X% of tokens in it. The %-s accumulate over time.

avikivity:
> > If we regularly have a 0.3ms stall (regularly = almost every poll period), that's bad. If we have such a hiccup every 100 periods, it's unnoticable.
> > Which is it?
> 
> It doesn't really matter how frequent 0.3ms stalls are (in fact, the bad clusters I've seen all had ~10% of task-quota violations). If we replace 50usec with 5usec in the above math, the result will be the same, but the scheduler will lose 0.7% of tokens every tick, not 7%. The result is the same, but slower.


How can  we lose a token? We take it from the top bucket, wait for completion, put it in the bottom bucket, move it again. It's fully recycled.

> 
> > We should increase the io latency goal. 0.75ms is too close to the task quota period.
> 
> Every time a request is completed X% later then the math thinks it will, token bucket loses ~X% of tokens in it. The %-s accumulate over time.

Well that's broken, but I don't understand why.

xemul:
> I don't understand. A stall will give the disk time to catch up and complete all requests.

Yes. Disk will complete all the requests. Then the replenisher will try to put _more_ tokens into bucket, because _more_ time had passed since lat time, but it won't be able to do it, because the N_requests / Duration is < N_requests / IO_quota

> After we recover from the stall, the buckets are fully replenished.

Provided nobody grabs from it, yes. If the IO stops for a while, the bucket is fully replenished.

xemul:
> How can we lose a token? We take it from the top bucket, wait for completion, put it in the bottom bucket, move it again. It's fully recycled.

`Token != request`
`Token == request / duration`
We take `request / task_quota`, put back `request / (task_quota + extra_delay)`

avikivity:
Sorry, I'm really confused. We probably have different mental models of how this works. Since you wrote it, your model is probably correct, but it doesn't help me understand what's going on.

avikivity:
> > How can we lose a token? We take it from the top bucket, wait for completion, put it in the bottom bucket, move it again. It's fully recycled.
> 
> `Token != request` `Token == request / duration` We take `request / task_quota`, put back `request / (task_quota + extra_delay)`

Ah. That's wrong.

Token should be "normalized disk ability to do something". If you take it, you have to put it back.

 Pulling something from the token bucket at a steady rate = queuing work at a disk at a steady rate.

xemul:
The "ability" nowadays means IOPS and bandwidth, which are speeds, not raw "number of requests" and "number of bytes" metrics. One cannot take "1 operation per second" thing from a bucket, instead, it takes an "1 operation" from a bucket, but it appears the re once every second. If it appears once every 1.01 seconds, then one will take "0.99 operations per second", which, in turn, will make it appear back in 1.02 seconds, and so on. This effect disappears in case token bucket is left idle for some time to fully replenish

xemul:
Let me try to demonstrate it with some .. weird artificial numbers just to take a look at it from the different angle.

Let's take a t.b. with 10 token / second rate. And the input flow into the scheduler at the same rate -- 10 requests / second. Tick is every second.

tick-1: 10 requests in the queue, 10 tokens in a bucket. All can be and are dispatched
tick-2: 10 requests complete, 10 tokens are put into the 'replenishing bucket'. And again, 10 requests in a queue, 10 tokens replenished, all can be and all are dispatched
tick-N: the same

This is how it should work. Now let's imagine that tick-X takes 1.1 seconds.

tick-x: 10 requests completed, 10 tokens are put into replenishing bucket. 11 requests in a queue. 11 tokens are to be replenished (because 1.1 seconds passed since last replenish), but (!) only 10 can be, 1 token is lost only 10 are replenished and only 10 (!) requests are submitted. Queue grew up 1 request

None of the next ticks will drain this queue, because each second only 10 requests are completed and only 10 tokens are replenished. We've "lost" 1 token and can only get it back if only 9 or less requests are queued, no other ways. Even if tick-y takes 0.9 seconds it won't help, because we'll have 9 requests completed, 9 queued and 8 tokens replenished.

xemul:
> We drain tokens at some rate. 

Yes, exactly. And we replenish tokens at some rate and in order for it to work in-sunk the rate of filling the 2nd bucket should be at least the same, or faster. But since it's _a rate_, delay in completing a request renders us larger denominator of the "work / duration" fraction and _lower_ replenishing rate.

> So if a token represents work, draining at some rate represents work per unit time.

xemul:
> Why do we want to replenish 11 and not 10?

Because we replenish "the amount of tokens for 1.1 second" which is time * rate = 1.1 * 10 = 11

avikivity:
> > Why do we want to replenish 11 and not 10?
> 
> Because we replenish "the amount of tokens for 1.1 second" which is time * rate = 1.1 * 10 = 11

We should replenish the amount of tokens we got.

xemul:
> We should replenish the amount of tokens we got.

This is going to be advanced `--max-io-requests` thing, not the rate-limiter based on ... cyan-purple diskplorer plots that ties _rates_ of {requests, bytes} x {read, write} with each other.

xemul:
> We should replenish the amount of tokens we got.

By the way, we _got_ 10 tokens in the above example. 11 is the amount of tokens we _want_ as per time-replenishment

xemul:
One more example of how the back-link makes things worse.

Let's assume that we've 5000 requests to dispatch and each tick a scheduler can push 500 as per the rate-limiter. And (!) each tick delays completion of 5% of the dispatched requests. In case dispatch loop just does what it needs -- dispatches 500 requests at a time, the ticks would look like this:

```
tick       |   1 |   2 |   3 |   4 |   5 |   6 |   7 |   8 |   9 |  10 |  11
dispatched | 500 | 500 | 500 | 500 | 500 | 500 | 500 | 500 | 500 | 500 |  
complated  | 475 | 475 | 475 | 475 | 475 | 475 | 475 | 475 | 475 | 475 | 250
```

It takes 11 ticks to process and one extra tick to pick up delayed completions. Not a big deal. And the dispatch average rate of 5000/10=500 req/sec is almost on par with the completion rate of 5000/11=454 req/sec.

Now if adding a "back-link", i.e. -- each tick the scheduler can dispatch min(500, completed_on_previous_tick) requests, the result is notably worse

```
tick       |   1 |   2 |   3 |   4 |   5 |   6 |   7 |   8 |   9 |  10 |  11 |  12 |  13 |  14 | 15
dispatched | 500 | 475 | 451 | 429 | 407 | 387 | 368 | 349 | 332 | 315 | 299 | 284 | 270 | 134 |  
completed  | 475 | 451 | 429 | 407 | 387 | 368 | 349 | 332 | 315 | 299 | 284 | 270 | 257 | 127 | 250
```

It takes 15 ticks, and still one unnoticeable extra tick to pick up those 250 delayed requests. The dispatch rate is 357 req/sec and completion rate is 333 req/sec. 30% of the throughput ability is lost

This model has a flaw in a sense that in the latter case delayed requests will get completed two/three/... ticks later, yes, but it's a simplification. For more ... real-ish numbers I've created a test in #1724 , it shows very similar effect, it takes more "ticks" to reveal itself, but since each tick is 0.5ms, in real time it will show up in seconds.

avikivity:
> One more example of how the back-link makes things worse.
> 
> Let's assume that we've 5000 requests to dispatch and each tick a scheduler can push 500 as per the rate-limiter. And (!) each tick delays completion of 5% of the dispatched requests. In case dispatch loop just does what it needs -- dispatches 500 requests at a time, the ticks would look like this:
> 
> ```
> tick       |   1 |   2 |   3 |   4 |   5 |   6 |   7 |   8 |   9 |  10 |  11
> dispatched | 500 | 500 | 500 | 500 | 500 | 500 | 500 | 500 | 500 | 500 |  
> complated  | 475 | 475 | 475 | 475 | 475 | 475 | 475 | 475 | 475 | 475 | 250
> ```
> 
> It takes 11 ticks to process and one extra tick to pick up delayed completions. Not a big deal. And the dispatch average rate of 5000/10=500 req/sec is almost on par with the completion rate of 5000/11=454 req/sec.

This is fine and after some time we'll dispatch and complete 475 each tick. But this corresponds to the current implementation as I understand it.


> 
> Now if adding a "back-link", i.e. -- each tick the scheduler can dispatch min(500, completed_on_previous_tick) requests, the result is notably worse

Well, why would we do that?


> 
> ```
> tick       |   1 |   2 |   3 |   4 |   5 |   6 |   7 |   8 |   9 |  10 |  11 |  12 |  13 |  14 | 15
> dispatched | 500 | 475 | 451 | 429 | 407 | 387 | 368 | 349 | 332 | 315 | 299 | 284 | 270 | 134 |  
> completed  | 475 | 451 | 429 | 407 | 387 | 368 | 349 | 332 | 315 | 299 | 284 | 270 | 257 | 127 | 250
> ```
> 
> It takes 15 ticks, and still one unnoticeable extra tick to pick up those 250 delayed requests. The dispatch rate is 357 req/sec and completion rate is 333 req/sec. 30% of the throughput ability is lost
> 
> This model has a flaw in a sense that in the latter case delayed requests will get completed two/three/... ticks later, yes, but it's a simplification. For more ... real-ish numbers I've created a test in #1724 , it shows very similar effect, it takes more "ticks" to reveal itself, but since each tick is 0.5ms, in real time it will show up in seconds.

xemul:
> Well, why would we do that?

Because that's how the 2-buckets work. A bucket doesn't have more tokens than there were returned upon completion. I.e. if we dispatched 20, then 19 were completed and we replenish t.b., even if time-based replenisher tells we have 20, we'd still replenish 19

avikivity:
> > Well, why would we do that?
> 
> Because that's how the 2-buckets work. A bucket doesn't have more tokens than there were returned upon completion. I.e. if we dispatched 20, then 19 were completed and we replenish t.b., even if time-based replenisher tells we have 20, we'd still replenish 19

Yes, but the 1 that we lost would be gained back eventually. We don't dispatch min(500, completed_on_previous_tick), we can dispatch everything that was completed on even previous-er ticks.

xemul:
> The tokens represent work.

That's the point, yes. But token is not a request, token is a request/duration. If we dispatch N requests in D==io_latency_goal, then we have consumed N/D tokens, but if completion of those tokens takes D+d duration, then we return N/(D+d) tokens, i.e. -- less than was taken.

> ... we can dispatch everything that was completed on even previous-er ticks.

Yes, we do, but when it happens it happens "too late already". Similarly to my above comments (with tables) -- the last tick compensates for the 10+-ticks of slowdown, but it doesn't help, the ops/tick rate is already low.

> But on the other hand if we introduce tokens into the system (not recycling them from completions), we're giving ourselves credit for work that the disk can't perform.

The idea of 2-bucket replenisher came from this, yes. However this backlink works with the 2-fold assumption:

a) scheduler increases the dispatch rate and at some point the completion rate goes down, because disk gets overloaded
b) scheduler decreases the dispatch rate and at some point the completion rate goes up (goes back to normal), because disk can finally cope with the load

This assumption doesn't always work as expected with 2-buckets in its current implementation. On one hand, the dispatch rate increases and at the same time reactor IO poll rate slows down (because of "time spent in task quota violation") so according to part a) of the assumption the t.b. decreases its dispatch rate expecting the part b) to execute. But it doesn't  instead, "time spent in task quota violation" makes those fewer requests, that had been dispatched, complete later as well. So the completion _rate_ doesn't get back to normal. And instead of getting b) to execute, t.b. gets a) again and decrease the rate again.

I'll show another backlink later today, it needs some work on the fair-queue-ticket resurrected.

avikivity:
> > The tokens represent work.
> 
> That's the point, yes. But token is not a request, token is a request/duration. If we dispatch N requests in D==io_latency_goal, then we have consumed N/D tokens, but if completion of those tokens takes D+d duration, then we return N/(D+d) tokens, i.e. -- less than was taken.

I'm really confused. In the classical application of token bucket, networking, tokens represent bytes, not bytes/sec. bytes are dripped into the bucket at the maximum bandwidth and taken out by packets being sent. So the tokens represent work, not work/duration.

How can you use work/duration to control requests that you don't know how long will take?


> 
> > ... we can dispatch everything that was completed on even previous-er ticks.
> 
> Yes, we do, but when it happens it happens "too late already". Similarly to my above comments (with tables) -- the last tick compensates for the 10+-ticks of slowdown, but it doesn't help, the ops/tick rate is already low.

If the disk completes work that corresponds to 475 tokens every tick, we should issue new work that corresponds to 475 tokens every tick. This is natural is tokens represent work, not work/duration, and works if the disk is lower than iotune says.


> 
> > But on the other hand if we introduce tokens into the system (not recycling them from completions), we're giving ourselves credit for work that the disk can't perform.
> 
> The idea of 2-bucket replenisher came from this, yes. However this backlink works with the 2-fold assumption:
> 
> a) scheduler increases the dispatch rate and at some point the completion rate goes down, because disk gets overloaded b) scheduler decreases the dispatch rate and at some point the completion rate goes up (goes back to normal), because disk can finally cope with the load
> 
> This assumption doesn't always work as expected with 2-buckets in its current implementation. On one hand, the dispatch rate increases and at the same time reactor IO poll rate slows down (because of "time spent in task quota violation") so according to part a) of the assumption the t.b. decreases its dispatch rate expecting the part b) to execute. But it doesn't instead, "time spent in task quota violation" makes those fewer requests, that had been dispatched, complete later as well. So the completion _rate_ doesn't get back to normal. And instead of getting b) to execute, t.b. gets a) again and decrease the rate again.


So long as the task quota + task quota violation is significantly slower than the io latency goal, it's not a problem, because we gave the disk enough work to carry over the stall.

If the stall is larger than io_latency_goal, then yes, disk bandwidth will suffer. It's not a problem if these large stalls are rare. If we want bandwidth to remain despite such large stalls, we should increase io_latency_goal to be larger than recent stalls. Note it still won't work well since while we stall we won't generate I/O for the disk. I don't propose doing any of that, it's better to just fix the stalls.

> 
> I'll show another backlink later today, it needs some work on the fair-queue-ticket resurrected.

avikivity:
> > I'm really confused. In the classical application of token bucket, networking, tokens represent bytes, not bytes/sec. bytes are dripped into the bucket at the maximum bandwidth and taken out by packets being sent. So the tokens represent work, not work/duration.
> 
> Well, they do represent work, but in a very restricted sense. E.g. tokens cannot be "cached" or withdrawn "in advance", once a token is taken, it should be immediately utilized by sending the packet/request to its destination. Nor it can be "reused" later, i.e. if a packet is sent the corresponding token is sort of lost forever, it cannot be taken back by next packet in the queue.


For networking, yes. A network has a fixed bandwidth, no more (and only very rarely less). The rate of token drip represents the bandwidth limit.

The disk doesn't have fixed bandwidth. So we recycle tokens to prevent accumulation.

total_token_count / disk_bandwidth_in_tokens_per_sec =~ io_latency_goal

If the real disk bandwidth is different than the iotune estimation, the result is a different effective io_latency_goal, not ever increasing latency.

> 
> > How can you use work/duration to control requests that you don't know how long will take?
> 
> But I don't. Token buckets control the rate of "going out requests" only.
> 
> > If the disk completes work that corresponds to 475 tokens every tick, we should issue new work that corresponds to 475 tokens every tick. This is natural is tokens represent work, not work/duration, and works if the disk is lower than iotune says.
> 
> Not "every tick" that's the point. Completions have longer "tick" compared to the initially configured dispatching tokenb bucket replenishing tick.
> 
> The 2nd bucket with completing requests can be thought of a token bucket too, where tokens fill it not by timer, but by completion events. And this token bucket filling process would also have its "rate" (average, of course, because completions are not uniform). And there are two links between those rates. Completing bucket filling rate is less or equal that the dispatching bucket rate, naturally, disk cannot complete more requests than there were dispatched. Dispatching bucket has less or equal rate because of that back-link. Now there are two buckets with rates that are only allowed to become smaller.

No. Those requests will complete on later ticks and will be recycled.

Let's say we dispatch one token and the request doesn't complete. So yes on some ticks we have 0 completions < 1 dispatched. But on later ticks we have 1 completion > 0 dispatched.

It can impact bandwidth if (average completion time) > (total number of tokens) / (token drip rate) (that's just Little's Law). In fact "average" can be replaced by some high percentile. But if it happens it just means the io_latency_goal is unrealistic.



> 
> > So long as the task quota + task quota violation is significantly slower than the io latency goal, it's not a problem, because we gave the disk enough work to carry over the stall.
> > If the stall is larger than io_latency_goal, then yes, disk bandwidth will suffer. It's not a problem if these large stalls are rare. If we want bandwidth to remain despite such large stalls, we should increase io_latency_goal to be larger than recent stalls. Note it still won't work well since while we stall we won't generate I/O for the disk. I don't propose doing any of that, it's better to just fix the stalls.
> 
> Your point it valid, but it's different. If stalls are above io-latency goal then we're having a ["phantom jam"](https://www.scylladb.com/2022/04/19/exploring-phantom-jams-in-your-data-flow/). This thing happens at another level. [Here](https://github.com/scylladb/seastar/pull/1774#issuecomment-1669435751) I collected the 100ms-averaged `dispatched_rate / completed_rate` gauge for a node under pretty mild stress. "Mild" -- there were no quota-violations and cpu utilization was below 30%. The raio was _above_ 1.0 most of the time. It's above by less than 1%, but those drops accumulate in the long run and pull the rate down.

I don't know if dispatched_rate/completed_rate is meaningful and can be averaged. For a point in time, dispatched_rate refers to future work, and completed_rate refers to past work. Let's say in second t we dispatched 100 requests (and completed none). In second t+1 we dispatched none (and completed 100).

xemul:
> No. Those requests will complete on later ticks and will be recycled.
> 
> Let's say we dispatch one token and the request doesn't complete. So yes on some ticks we have 0 completions < 1 > dispatched. But on later ticks we have 1 completion > 0 dispatched.

This silently assumes that we have just one request in the queue and _maybe_ one more will arrive later. In that assumption no problems happen indeed, but that's not the case here. Neither I'm talking about 1 request per tick rate, the rate drop-down I'm observing in the test and metrics finds its equilibrium point at much higher rates.

But let's take this 1/tick rate, but (!) assume that a new request gets queued _each_ tick. If "on some ticks we have 0 completions" this means that the request queued this tick will _not_ be submitted, even though the previous one might have been completed by the disk. So after 2 ticks we have 2 requests queued, 1 submitted and 1 completed, while it should be 2 queued, 2 submitted and 1 completed. 50% loss. 

I agree that next+N such ticks we'll have N queued, N-1 completed and N-1 submitted. In such _small_ scale this will show up as 1/N deviation, with N->~~inf~~even "few tens" this becomes negligible. At higher rates this will be ~R/N loss, so it will "last longer" and while it "lasts" the queue will be larger than 1 request (it won't grow infinitely though) and the average request latency will be proportional to that queue size.

> I don't know if dispatched_rate/completed_rate is meaningful and can be averaged. For a point in time, dispatched_rate refers to future work, and completed_rate refers to past work.

Yes, and with the backlink in question we're limiting the future work with the amount of past work and the latter is _less_, so the future work also goes down.

> Let's say in second t we dispatched 100 requests (and completed none). In second t+1 we dispatched none (and completed 100).

Of course, of course, that's the idea behind this measurement. It's normally slightly above 1.0, but once submission rate exceeds _disk_ ability to process them to higher it will grow. Given disk is not overloaded the delay seastar puts on top vanishes.

xemul:
> io_latency_goal is represented by the size of the upper bucket.

Which on is upper? I assume that "upper'" is the one into which completed requests put their tokens. In that case "lower" is the one into which "upper" pours and from which scheduler gets tokens to dispatch requests with. Correct?

xemul:
>  There's just no way to have both high throughput and low latency. If you want high throughput you need to keep stuffing requests, but if the disk isn't keeping up the latency will grow.

I agree with that. But I'm not talking about higher throughput at the cost of lower latency :) I'm describing the case when the rate-limiter is configured at rate X, but when it starts working it generates Y << X (much less than) and its incoming queue is far from being empty. And the question at this point -- "WTF, it ~~can do~~ is configured for _much_ more, why doesn't it dispatch more requests?"

avikivity:
> > io_latency_goal is represented by the size of the upper bucket.
> 
> Which on is upper? I assume that "upper'" is the one into which completed requests put their tokens. In that case "lower" is the one into which "upper" pours and from which scheduler gets tokens to dispatch requests with. Correct?

Now I confused myself. Let's have a call, it will be faster.

avikivity:
> > There's just no way to have both high throughput and low latency. If you want high throughput you need to keep stuffing requests, but if the disk isn't keeping up the latency will grow.
> 
> I agree with that. But I'm not talking about higher throughput at the cost of lower latency :) I'm describing the case when the rate-limiter is configured at rate X, but when it starts working it generates Y << X (much less than) and its incoming queue is far from being empty. And the question at this point -- "WTF, it ~can do~ is configured for _much_ more, why doesn't it dispatch more requests?"

Well, why doesn't it?

xemul:
Avi, here's a spreadsheet simulation :)

Columns:
- tick length -- 500 usec each
- disk exec -- number of requests that were in disk this tick. It equals "dispatch" from previous tick
- disk complete -- number of requests completed by disk by the end of the tick. It's min("tick length", "disk exec"), because disk completes min(as much as it has, as much time it had to). The latter is considered to be on-par with the replenish limit
- reactor complete -- number of requests picked by by reactor and delivered to io scheduler. I set it to "disk complete" * 0.95 + (!) "completed by disk, but lost by reactor" from previous tick. This is to simulate this idea of reactor delaying completions and (!) picking up previously lost deliveries
- upper bucket -- the number of tokens put from completion -- equals "reactor complete"
- timer replenish -- the number of tokens that timer wants to put into bucket -- equals "tick length"
- lower bucket -- the number of tokens available for dispatch -- equals min("upper bucket", "timer replenish"). This is that notorious back-link thing
- dispatch -- how many requests are dispatched for the next tick. Equals min("lower bucket", 750), the 750 is io-latency-goal which is 1.5 * standard tick, but it cannot dispatch more than there are tokens in a bucket

Here's how it goes over time (screenshot):

image

xemul:
Correction. The "reactor complete" shouldn't substract previous tick if previous tick picked up everything (spoiler: doesn't change much)

image

avikivity:
Well, it's obviously bad, but why does it happen? It should stabilize on a steady state if the disk is able to complete at a steady state.

xemul:
> Well, it's obviously bad, but why does it happen? 

Because number of tokens in the upper bucket is on average less than the number of time-driven tokens. This results in less tokens in the lower bucket, than it should be. The fewer tokens in the lower bucket results in less requests in disk. Less requests in disk, again, results in less tokens in the upper bucket.

xemul:
updated:
- rebased
- merged flow-rate backlink (#1774)

@avikivity , reminder ping :)

avikivity:
I don't agree we can remove the second bucket. If the presumed disk parameters are higher than actual disk bandwidth, then latency will go through the roof.

Let's say configured throughput is 1GB/s and real throughput is 95% of that. After 1 second, there will be 50MB in flight in the kernel. After 2 seconds, 100MB in flight. Pretty soon all of the I/O will be kernel and nothing in user queues. It just kills scheduling.

xemul:
That's why the bda1287fe02ddc6366a9c3613ae4e44dfe8198e7 is here. If the disk happens to be too slow, the dispatch rate to completion rate ratio would grow, the cost of requests would grow as well, the dispatch rate would go down

src/core/io_queue.cc
```diff
@@ -536,6 +536,48 @@ const fair_group& get_fair_group(const io_queue& ioq, unsigned stream) {
 
 } // internal namespace
 
+template <typename T>
+void update_moving_average(T& result, const T& value, double factor) noexcept {
```
avikivity, line 540: Some constraint on T (with SEASTAR_CONCEPT)?

include/seastar/core/io_queue.hh
```diff
@@ -116,6 +116,9 @@ private:
     size_t _requests_executing = 0;
     uint64_t _requests_dispatched = 0;
     uint64_t _requests_completed = 0;
+    struct flow_monitor;
+    std::unique_ptr<flow_monitor> _flow_mon;
+
```
avikivity, line 121: Looks like unique_ptr is unnecessary here, flow_monitor is small and better inlined. I don't completely get why we need it _yet_.

avikivity:
Sorry, I still lack a basic understanding of the problem. I don't agree that R_c <= R_d. It can be greater, if the disk/kernel is completes requests dispatched from multiple ticks.

Maybe the problem is that the second bucket also has a fixed rate "hole"? The second bucket should have an infinitely-sized hole. Maybe we shouldn't say it's a bucket.

First bucket: from classical token-bucket algorithm. Tokens that leak from it are fed into the disk.
Second bucket: collects completions, feed back to first bucket.

xemul:
> Sorry, I still lack a basic understanding of the problem. I don't agree that R_c <= R_d. It can be greater, if the disk/kernel is completes requests dispatched from multiple ticks.

The essence of the problem is that both buckets' filling rate is less-or-equal to its peer's filling rate. The two-buckets system thus has a way to slow down, but doesn't have a way to speed up. Statistically, in the long run, tiny slow-down occurrences pile up. In order for the system to restore either bucket needs to get filled at higher rate than it's peer.

The latter can happen in several ways.

First, once IO queue gets empty the whole thing very quickly restores, because no tokens are grabbed from the lower bucket so after enough time passes they just replenish everything from disk. The problem is that once IO starts again it only takes tens-to-hundreds of ticks ([like this](https://github.com/scylladb/seastar/pull/1766#issuecomment-1672923672)) to lose most of its tokens. Tens-to-hundreds of ticks is less than a second, on the metrics dashboards we just see it's constantly low bandwidth

Second, if we disconnect lower bucket filling rate from the upper bucket filling rate, so it can grow larger. Or, some better wording -- the upper bucket filling rate can become _lower_ sometimes. This is what I do in this PR. In the long run there's the "flow monitor" thing that compares both rates in the longer run (EMA for 100ms duration) and once upper bucket falls behind, it slows down the lower bucket either.

> Maybe the problem is that the second bucket also has a fixed rate "hole"? The second bucket should have an infinitely-sized hole. Maybe we shouldn't say it's a bucket.

Upper bucket cannot have infinitely-sized hole, it's flow-in rate is limited with disk+kernel+seastar ability to complete requests.

> First bucket: from classical token-bucket algorithm. Tokens that leak from it are fed into the disk.
> Second bucket: collects completions, feed back to first bucket.

Both work like you described already. [Still](https://github.com/scylladb/seastar/pull/1766#issuecomment-1672923672), happens according to "filling rate of each bucket is less-or-equal to its peer" model

include/seastar/core/io_queue.hh
xemul: I just wanted to keep flow_monitor definition out of .hh file. But I'll move it back, it's small enough indeed

avikivity:
> > Sorry, I still lack a basic understanding of the problem. I don't agree that R_c <= R_d. It can be greater, if the disk/kernel is completes requests dispatched from multiple ticks.
> 
> The essence of the problem is that both buckets' filling rate is less-or-equal to its peer's filling rate. The two-buckets system thus has a way to slow down, but doesn't have a way to speed up. Statistically, in the long run, tiny slow-down occurrences pile up. In order for the system to restore either bucket needs to get filled at higher rate than it's peer.


Ok. Agree and understand so far.

> 
> The latter can happen in several ways.
> 
> First, once IO queue gets empty the whole thing very quickly restores, because no tokens are grabbed from the lower bucket 

Let's find absolute names for the buckets because I keep getting confused. Let's call one bucket the "classical token-bucket" and the other the "completion bucket".

I think I see another source of confusion for me - the classical token bucket gets refilled at a constant rate, it isn't drained at a constant rate. So now I understand the rate-lowering spiral too.

> so after enough time passes they just replenish everything from disk. The problem is that once IO starts again it only takes tens-to-hundreds of ticks ([like this](https://github.com/scylladb/seastar/pull/1766#issuecomment-1672923672)) to lose most of its tokens. Tens-to-hundreds of ticks is less than a second, on the metrics dashboards we just see it's constantly low bandwidth
> 
> Second, if we disconnect lower bucket filling rate from the upper bucket filling rate, so it can grow larger. Or, some better wording -- the upper bucket filling rate can become _lower_ sometimes. This is what I do in this PR. In the long run there's the "flow monitor" thing that compares both rates in the longer run (EMA for 100ms duration) and once upper bucket falls behind, it slows down the lower bucket either.
> 
> > Maybe the problem is that the second bucket also has a fixed rate "hole"? The second bucket should have an infinitely-sized hole. Maybe we shouldn't say it's a bucket.
> 
> Upper bucket cannot have infinitely-sized hole, it's flow-in rate is limited with disk+kernel+seastar ability to complete requests.
> 
> > First bucket: from classical token-bucket algorithm. Tokens that leak from it are fed into the disk.
> > Second bucket: collects completions, feed back to first bucket.
> 
> Both work like you described already. [Still](https://github.com/scylladb/seastar/pull/1766#issuecomment-1672923672), happens according to "filling rate of each bucket is less-or-equal to its peer" model

My description was actually incorrect. Tokens leak into the classical token-bucket at a fixed rate. They are drained from it at an infinite rate. This gives a limited burst (because we can drain the entire bucket at once), maximum bandwidth (because tokens replenish into it at a fixed rate) and a lot of back-and-forth conversation (because I misunderstood it).

Now that a deep misunderstanding is clear, I can try to re-read previous explanations and see if I understand them better.

avikivity:
I re-read the cover letter but I need some graphics to understand the goal.

xemul:
> Let's find absolute names for the buckets because I keep getting confused. Let's call one bucket the "classical token-bucket" and the other the "completion bucket".

In previous discussions we used "upper bucket" and "lower bucket". The "lower bucket" is the one from which scheduler grabs tokens for requests to submit them into the disk. This is "classical token-bucket". The "upper bucket" was the one into which completed requests put their tokens back. This is "completion bucket".

> I think I see another source of confusion for me - the classical token bucket gets refilled at a constant rate, it isn't drained at a constant rate. So now I understand the rate-lowering spiral too.
> My description was actually incorrect. Tokens leak into the classical token-bucket at a fixed rate. They are drained from it at an infinite rate. This gives a limited burst (because we can drain the entire bucket at once), maximum bandwidth (because tokens replenish into it at a fixed rate) and a lot of back-and-forth conversation (because I misunderstood it).

Tokens leak into the classical (lower) bucket at the fixed rate, but they are additionally limited with the amount of tokens in the completion (upper) bucket. They are drained from the classical (lower) bucket at an infinite rate, yes. The burst (bucket size)  is configured to be "as many requests that a disk claims to serve in one io-latency-goal duration"

> I re-read the cover letter but I need some graphics to understand the goal.

Sure thing, but what kind of graphics? Diagram of buckets/disk/tokens/etc., X-Y plane with plots showing the token conversion rates, or something else?

There's a table out of which I can make an X-Y plane with tokens [here](https://user-images.githubusercontent.com/4498177/259702362-ad99d40b-3fc3-4a14-a140-fef4039c3a10.png), but it's pretty artificially modeled. Also I stressed scylla with and without this PR [here](https://github.com/scylladb/seastar/pull/1774), last two comments have graphana plots for the "flow ratio" thing and the statement class IO queue delay + IO queue length (unfortunately I lost all VMs because I forgot about over-night killer)

avikivity:
A block diagram / flow chart thing. There was just so much talk about it, and ambiguous terms, and my own misunderstanding, that I can't form a clear picture without help.

avikivity:
What is threshold?

I don't understand why you're decreasing the cost if  the disk is slow.

avikivity:
btw, it's worth adding this to the documentation. github can render text representations: 

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams
http://mermaid.js.org/intro/

xemul:
> What is threshold?

Threshold is the maximum dispatch/completion ratio we consider to be "normal". In me stressing this value wasn't growing above 1.01, but in the PR I left it configurable

> I don't understand why you're decreasing the cost if the disk is slow.

Why decreasing? I multiply the cost by the ratio. If reactor dispatches more then disk manages to complete, the ratio grows above 1.0, the cost is multiplied by it and also goes up

xemul:
@avikivity , reminder ping (/me still remembers the request to add rendered diagrams to docs)

avikivity:
Quadratic behavior detected: each time I lose the thread I have to reread it from the beginning to restore the state.

avikivity:
I don't feel comfortable with the approach. How do you prove that it stabilizes? And what's the motivation for increasing cost by flow_ratio?

With the previous model. I had a mental picture of what's going on (maybe a wrong one, since it doesn't work well), but with the new model, I can't make the leap from equations to what's happening, how the system responds to (whatever it's supposed to respond to).

avikivity:
I guess it stabilizes because the adjustment is equal when taking tokens out of the bucket, and when returning them. So the total number of tokens cannot change.

avikivity:
No, that's not true, because we don't have the completion bucket.

xemul:
> How do you prove that it stabilizes? And what's the motivation for increasing cost by flow_ratio?

The rate-limited main equation is

$$ (\frac{iops}{iops_{max}} + \frac{bandwidth}{bandwidth_{max}}) <= 1$$

or

$$ \frac{d}{dt} \( \frac{ops}{iops_{max}} + \frac{bytes}{bandwidthbytes_{max}} \) <= 1$$

We implement it via token bucket with rate of `1.0` and the cost of individual request (cost == amount of tokens it needs to carry) being

$$ cost = \frac{1}{iops_{max}} + \frac{size}{bandiwdth_{max}}$$

Now let's assume that at some point the rate should be (I'll get back to how we assume it below) lowered, because disk doesn't keep up. It means that in teh frst formula we want to _decrease_ `iops_max` and `bandwidth_max` denominators.

$$ \frac{d}{dt} \( \frac{ops}{iops_{max} * \alpha} + \frac{bytes}{bandwidthbytes_{max} * \beta} \) <= 1 , \alpha < 1.0 , \beta < 1.0$$

In this PR I say that

$$ \alpha = \beta = \frac {1}{\gamma} , \gamma > 1.0$$

because we don't know which IOPS or bandwidth was overwhelmed for the disk (spoiler: actually most likely it's the latter, but still), which means new token bucket equation is

$$  \frac{d}{dt} \( \gamma * \( \frac{ops}{iops_{max}} + \frac{bytes}{bandwidthbytes_{max}} \) \) <= 1, \gamma > 1.0 $$

The last equation means, that in order to make scheduler work with _lower_ IOPS and bandwidth, we may multiply individual request cost by some number larger than 1.0.

Now how do we "assume that IOPS and bandwidth should be lowered"? There are also many ways to assume that, but in this PR I say that: the IOPS and bandwidth should be lowered in case scheduler submits more requests per second than the kernel completes requests per second in the long run. This is where the flow ratio came from.

Summary:
- in our case decreasing t.b. rate can be achieved (though not equivalent) to multiplying the request cost by some value larger than 1.0
- the sign of "disk is overloaded" is when it completes requests at lower rate than we dispatch

avikivity:
+2 points for inline equations


> The last equation means, that in order to make scheduler work with _lower_ IOPS and bandwidth, we may multiply individual request cost by some number larger than 1.0.

That's pretty simple, now that you've explained it so. Another way to look at it is that the we make consumption rate smaller than alpha (alpha<1) instead of smaller than 1.


> 
> Now how do we "assume that IOPS and bandwidth should be lowered"? There are also many ways to assume that, but in this PR I say that: the IOPS and bandwidth should be lowered in case scheduler submits more requests per second than the kernel completes requests per second in the long run. This is where the flow ratio came from.

But measured in consumption units, yes? especially if the bandwidth is the root cause.

But in the really long run, submits have to equal completes, no? So the flow ratio measures if we're int the process of moving from one equilibrium (there submits equal completes) to another (where they are also equal, but latency is much higher). I guess in the really long run, we achieve equilibrium by running out of concurrency, whereas we want to achieve equilibrium by limiting latency.

> 
> Summary:
> 
> * in our case decreasing t.b. rate can be achieved (though not equivalent) to multiplying the request cost by some value larger than 1.0
> * the sign of "disk is overloaded" is when it completes requests at lower rate than we dispatch
> 
> (upd: s/bandwidthbytes/bandwidth/ in some formulas)


Is the algorithm stable? suppose we increased the cost and achieved a flow ratio of 1. We must have, we can't have a flow ratio > 1 for long periods. Then we jump back to gamma=1. Will we fluctuate between the two modes?


One way to test it is to run a test with iotune numbers adjusted upwards by 20%, so we know the disk is slower than what the scheduler thinks it is.

Maybe we should monitor latency directly, though that's a very noisy meaurement. If moving_average(latency) > latency_goal, increase cost.

We also need to parametrize the algorithm. 100ms works for SSD (maybe too large), won't work for HDD, and then I have long conversations with @horschi.

xemul:
> Another way to look at it is that the we make consumption rate smaller than alpha (alpha<1) instead of smaller than 1.

Possible, but 1.0 on the right side of the equation is token bucket filling rate. It's "shared" between shards and tuning it in share-nothing approach is more challenging that letting each shard increase its requests' cost

> But measured in consumption units, yes? especially if the bandwidth is the root cause.

No, measured in the number of requests

> But in the really long run, submits have to equal completes, no? So the flow ratio measures if we're int the process of moving from one equilibrium (there submits equal completes) to another (where they are also equal, but latency is much higher). I guess in the really long run, we achieve equilibrium by running out of concurrency, whereas we want to achieve equilibrium by limiting latency.

In the _really_ long run -- probably. But If we push more requests into ~~disk~~ kernel than the disk can handle what will happen? AFAIK (but I can be wrong) the amount of data disk _and_kernel_ can queue-up internal is pretty large if not "infinite". With 10Gb/s disk and 100ms averaging in case disk just _stops_ kernel will need to queue just 1Gb, if disk is "suddenly" back to normal is will drain withing the next 100ms

> Is the algorithm stable? suppose we increased the cost and achieved a flow ratio of 1. We must have, we can't have a flow ratio > 1 for long periods. Then we jump back to gamma=1. Will we fluctuate between the two modes?

Once every 100ms, it is possible in theory, yes. But why is that a problem? Is your point that "we better settle at gamma of, say, 1.01, rather than switching between 1.0 and, say, 1.1"?

> Maybe we should monitor latency directly, though that's a very noisy meaurement. If moving_average(latency) > latency_goal, increase cost.

I had this idea, but requests of different size and direction would have different latencies, say 4k write is 50us, 128k read can be 200us. If the former's latency goes up 3x times it's still below latency goal, but disk is likely in trouble already.

> We also need to parametrize the algorithm. 100ms works for SSD (maybe too large), won't work for HDD

It is all configurable already, but I just noticed that I forgot to expose the averaging period as an option, only as io-queue config parameter (the cost-increase threshold _is_ an option in this PR)

avikivity:
> > Another way to look at it is that the we make consumption rate smaller than alpha (alpha<1) instead of smaller than 1.
> 
> Possible, but 1.0 on the right side of the equation is token bucket filling rate. It's "shared" between shards and tuning it in share-nothing approach is more challenging that letting each shard increase its requests' cost

Sure, I was more concerned in explaining it to myself than suggesting a code change.


> 
> > But measured in consumption units, yes? especially if the bandwidth is the root cause.
> 
> No, measured in the number of requests
> 
> > But in the really long run, submits have to equal completes, no? So the flow ratio measures if we're int the process of moving from one equilibrium (there submits equal completes) to another (where they are also equal, but latency is much higher). I guess in the really long run, we achieve equilibrium by running out of concurrency, whereas we want to achieve equilibrium by limiting latency.
> 
> In the _really_ long run -- probably. But If we push more requests into ~disk~ kernel than the disk can handle what will happen? AFAIK (but I can be wrong) the amount of data disk _and_kernel_ can queue-up internal is pretty large if not "infinite". With 10Gb/s disk and 100ms averaging in case disk just _stops_ kernel will need to queue just 1Gb, if disk is "suddenly" back to normal is will drain withing the next 100ms

The kernel queue is really the size of the aio context we allocate (summed across all shards). So we can control submit rate until we fill up all aio slots. Once we reach the limit, submit rate becomes equal to the completion rate.


> 
> > Is the algorithm stable? suppose we increased the cost and achieved a flow ratio of 1. We must have, we can't have a flow ratio > 1 for long periods. Then we jump back to gamma=1. Will we fluctuate between the two modes?
> 
> Once every 100ms, it is possible in theory, yes. But why is that a problem? Is your point that "we better settle at gamma of, say, 1.01, rather than switching between 1.0 and, say, 1.1"?
> 
> > Maybe we should monitor latency directly, though that's a very noisy meaurement. If moving_average(latency) > latency_goal, increase cost.
> 
> I had this idea, but requests of different size and direction would have different latencies, say 4k write is 50us, 128k read can be 200us. If the former's latency goes up 3x times it's still below latency goal, but disk is likely in trouble already.

Why do we care?  We have a latency goal.

Note if we have small requests, we also send very many of them, so their latency will also grow.


> 
> > We also need to parametrize the algorithm. 100ms works for SSD (maybe too large), won't work for HDD
> 
> It is all configurable already, but I just noticed that I forgot to expose the averaging period as an option, only as io-queue config parameter (the cost-increase threshold _is_ an option in this PR)


I'd still like to see a test with artificially raised io_properties.yaml. Will it survive?

xemul:
> The kernel queue is really the size of the aio context we allocate (summed across all shards). So we can control submit rate until we fill up all aio slots. Once we reach the limit, submit rate becomes equal to the completion rate.

Yes, that's true

> Why do we care? We have a latency goal.
> Note if we have small requests, we also send very many of them, so their latency will also grow.

Because one-shot measure is not reliable. E.g. with latencies we don't measure disk latency, we measure disk + seastar latency, so even if one-time measurement goes up it doesn't mean disk gets slow. This is _exactly_ what the current approach has troubles with -- it notices that the short-term completion rate gets lower (because reactor is a bit late in completing request). For per-request latencies we need some long-term aggregate, but what can it be? The "expected" latency is not a single number we can get e.g. some average on.

avikivity:
Perhaps we should "learn" the actual disk capacity using a long-term moving average. But wait - that's effectively what you do. But I don't see why the flow ratio is the right trigger.

xemul:
The math behind flow-ratio.

_d<sub>i</sub>_ -- the amount of requests dispatched at tick _i_
_p<sub>i</sub>_ -- the amount of requests processed by disk at tick _i_
_c<sub>i</sub>_ -- the amount of requests completed by reactor loop at tick _i_

We can observe _d<sub>i</sub>_ and _c<sub>i</sub>_ in the dispatcher, but _not_ the _p<sub>i</sub>_, because we don't have direct access to disks' queues

After _n_ ticks we have 

_D<sub>n</sub>_ -- total amount of requests dispatched, 
_P<sub>n</sub>_ -- total amount of requests processed, 
_C<sub>n</sub>_ -- total amount of requests completed, 

$$ D_n = \sum_{i=0}^n d_i $$

$$ P_n = \sum_{i=0}^n p_i $$

$$ C_n = \sum_{i=0}^n c_i $$

* Disk cannot process more than it was dispatched, but it can process less "accumulating" a queue

$$ d_i \ge p_i   \Rightarrow  D_n \ge P_n$$

* Reactor cannot complete more than it was processed by disk either

$$ p_i \ge c_i \Rightarrow P_n \ge C_n $$


Note that _D<sub>n</sub> > P<sub>n</sub>_  means that disk is falling behind. Our goal is to make sure the disk doesn't do it and doesn't accumulate the queue, i.e. _D<sub>n</sub> = P<sub>n</sub>_, but we cannot observe _P<sub>n</sub>_ directly, only _C<sub>n</sub>_.

Next

$$ D_n - P_n = Qd_n \ge 0 $$

$$ P_n - C_n = Qc_n \ge 0 $$

$$ D_n - C_N = Qd_n + Qc_n \ge 0 $$

_Qd<sub>n</sub>_ is the queue accumulated in disk. We try to avoid it. _Qc<sub>n</sub>_ is the accumulated delta between processed (by disk) and completed (by reactor) requests. We cannot reliably say that it goes to zero over time, because there's always some unknown amount of processed but not yet completed requests. So the above formula contains two unknowns and we cannot solve it. Let's try the other way

$$ \frac {D_n} {P_n} = Rd_n \ge 1 $$

$$ \frac {P_n} {C_n} = Rc_n \ge 1 $$

$$ \frac {D_n} {C_n} = Rd_n \times Rc_n \ge 1 $$

_Rd<sub>n</sub>_ is the ratio of dispatched to processed requests. If it's 1, we're OK, if it's greater, disk is accumulating the queue, we try to avoid it. _Rc<sub>n</sub>_ is the ratio between processed (by disk) and completed (by reactor) requests. It has a not immediately apparent, but very pretty property

$$ \lim_{n\to\infty} Rc_n = \lim_{n\to\infty} \frac {P_n} {C_n} = \lim_{n\to\infty} \frac {C_n + Qc_n} { C_n} = \lim_{n\to\infty} \(1 +  \frac {Qc_n} {C_n} \) = 1 + \lim_{n\to\infty} \frac {Qc_n} {C_n}$$

The _Qc<sub>n</sub>_ doesn't grow over time. It's non-zero, but it's upper bound by some value. Respectively

$$ \lim_{n\to\infty} Rc_n = 1 $$

$$ \lim_{n\to\infty} \frac {D_n} {C_n} = \lim_{n\to\infty} \( Rd_n \times Rc_n \) =  \lim_{n\to\infty} Rd_n  $$

IOW -- we can say if the disk is accumulating the queue or not by observing the dispatched-to-completed (to _completed_, not _processed_) over a long enough time

xemul:
upd:
- make flow-ratio duration depend on io_latency_goal
- reduce slow-down threshold to 1.1
- added doc file describing math behind the model

PR #1767, opened 2023-08-04
Add word count demo.

rishvin:


rishvin:
This change is still a draft.

PR #1768, opened 2023-08-04
Make make_file_impl() less yielding

xemul:
The helper in question makes a file object out of a given file descriptor and options+flags. In order to proceed it first calls `engine().fstat(fd)` to get the file mode and device. The `reactor::fstat()` yields to thread-pool. However, the make_file_impl()'s callers get the file descriptor via thread pool either, so they can call fstat on it instantly thus avoiding second yield

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -206,7 +206,14 @@ posix_file_impl::flush() noexcept {
 
 future<struct stat>
 posix_file_impl::stat() noexcept {
-    return engine().fstat(_fd);
+    return engine()._thread_pool->submit<syscall_result_extra<struct stat>>([fd = _fd] {
+        struct stat st;
+        auto ret = ::fstat(fd, &st);
+        return wrap_syscall(ret, st);
+    }).then([] (syscall_result_extra<struct stat> ret) {
+        ret.throw_if_error();
+        return make_ready_future<struct stat>(ret.extra);
+    });
 }
 
 future<>
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -1024,84 +1024,85 @@ xfs_concurrency_from_kernel_version() {
 
 future<shared_ptr<file_impl>>
 make_file_impl(int fd, file_open_options options, int flags, struct stat st) noexcept {
-        auto st_dev = st.st_dev;
-
-        if (S_ISBLK(st.st_mode)) {
-            size_t block_size;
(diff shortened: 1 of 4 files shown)


avikivity:
Note that we can consider fstat() a non-blocking call since it's just metadata on an open file. I'm not sure it's correct for nfs through with its limited metadata caching.

src/core/reactor.cc
```diff
@@ -1874,9 +1875,13 @@ reactor::open_file_dma(std::string_view nameref, open_flags flags, file_open_opt
                     ::ioctl(fd, XFS_IOC_FSSETXATTR, &attr);
                 }
             }
+            r = ::fstat(fd, &st);
+            if (r == -1) {
+                return wrap_syscall(r, st);
+            }
```
avikivity, line 1881: Thought experiment: with io_uring this translates to IOURING_OPEN_AT + IOSQE_FIXED_FILE, followed by IORING_OP_STATX + IOSQE_FIXED_FILE. They can then be done by the same io_uring_enter call (though likely with a workqueue to help). /cc  @tchaikov

avikivity: (it would have been nice of io_uring to be able to say, take the fd from the linked openat call)

src/core/file.cc
```diff
@@ -206,7 +206,14 @@ posix_file_impl::flush() noexcept {
 
 future<struct stat>
 posix_file_impl::stat() noexcept {
-    return engine().fstat(_fd);
+    return engine()._thread_pool->submit<syscall_result_extra<struct stat>>([fd = _fd] {
+        struct stat st;
+        auto ret = ::fstat(fd, &st);
```
avikivity, line 211: I expect we'll need to move it back, as reactor::open_and_stat(), once we do this via io_uring. We need an abstraction in reactor once it can be done in more than one way.

src/core/reactor.cc
mykaul: IO uring has this file descriptors 'pre-registration' which helps with reusing fd's across calls. Are we using it?

mykaul: (https://man7.org/linux/man-pages/man3/io_uring_register_files.3.html )

avikivity: That's what IOSQE_FIXED_FILE uses (IIUC). We're not using it yet.

PR #1770, opened 2023-08-05
fair_queue: Do not re-evaluate request capacity twice

xemul:
On dispatch the request capacity is calculated into req_cap variable, so it can be reused everywhere below


tchaikov (approved):
just a nit in the title of the commit message: s/twise/twice/

avikivity (approved):

xemul:
> just a nit in the title of the commit message: s/twise/twice/

Thanks! Fixed

tchaikov (approved):
lgtm

PR #1773, opened 2023-08-08
Keep capacity on fair_queue_entry

xemul:
Current implementation of fair-queue uses two distinct data types to "measure" a request -- tickets and capacities. The former is a 2d value that represents request weight and size and the latter is a fixed-point number that represents the request cost that's used by cross-classes balancing and capacity throttling. The capacity is calculated from the ticket each time f.q. needs to know the entry's capacity, which is not very efficient.

Other than generating extra CPU waste, this detail makes it hard to auto-adjust the dispatch rate. This is because the entry's (and a entry represents an IO request) capacity is its ticket normalized by the capacity "cost" ticket. If the queue needs to lower or increase the dispatch rate, it has two options: re-configure the token bucket, or change the "cost" value so that each request becomes more weighty or lighter. The 2nd option is not possible now, because the "cost" dispatch-time and the "cost" completion-time cannot differ. This patch fixes it by pinning the entry's cost construction time, so the rate can be altered at any time.

Tested with:
- test.py
- comparing ioinfo output for unconfigured and iotune-configured disk (capacities changed tiny bit due to rounding shifts)
- stressing scylla build with this set applied

### DIFF
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -243,11 +243,10 @@ public:
          * must accept those as large as the latter pair (but it can accept
          * even larger values, of course)
          */
-        unsigned limit_min_weight = 0;
-        unsigned limit_min_size = 0;
         unsigned long weight_rate;
         unsigned long size_rate;
         double min_tokens = 0.0;
+        double limit_min_tokens = 0.0;
         float rate_factor = 1.0;
         std::chrono::duration<double> rate_limit_duration = std::chrono::milliseconds(1);
     };
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -214,7 +214,6 @@ private:
      * into more tokens in the bucket.
      */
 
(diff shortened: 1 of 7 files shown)


include/seastar/core/fair_queue.hh
```diff
@@ -116,7 +119,7 @@ public:
             bi::cache_last<true>,
             bi::member_hook<fair_queue_entry, bi::slist_member_hook<>, &fair_queue_entry::_hook>>;
 
-    fair_queue_ticket ticket() const noexcept { return _ticket; }
+    fair_queue_ticket capacity() const noexcept { return _ticket; }
```
avikivity: capacity is the wrong term here, the disk has capacity (capability to do something), this is something that must fit in capacity. I don't mind if it's renamed post-merge, but we need a more accurate name.

xemul: It could be "cost", but "cost" is busy with cross-classes balancing: `request.cost = request.capacity / class.shares` and then each class is charged with this "cost", scheduler picks a class with the lowest charge for the next dispatch

PR #1774, opened 2023-08-08
Slowdown IO scheduler based on dispatched/completed ratio

xemul:
The essence of the PR is last commit (b24ff95fd895e08f7efd77e79c7054a03da4d9b6). It also contains PRs #1766, #1773 and #1754 it depends on

As described in #1766, current attempt to make IO queue slowdown if disk cannot keep up with the load is problematic. The main problem is that the backlink relies on requests completions arriving in a timely manner thus replenishing tokens into the bucket. It doesn't work well, because the case when "disk is overloaded and delays completion of requests" is     indistinguishable from "reactor is overloaded and cannot poll completions in a timely manner". As a result, lowering the rate of  dispatch on softly stalling reactor doesn't necessarily help making the disk more happy. Reactor would continue delaying completions and the  backling would drop the rate even further.

This PR makes the backlink based on dispatch-to-completion rate. Normally, the number of requests dispatched for a certain duration divided by the number of requests completed for the same duration must be 1.0. Otherwise that would mean that requests accumulate in disk. However, this ratio cannot be such immediately and in the longer run it tends to be slightly greater that 1.0, because if reactor polls kernel for IO completions more often, it won't get _more_ requests that it was dispatched. But even a small delay in polling would make Nr_completed / duration _less_ because of the larger denominator value.

Having said that, the backlink is based on the flow-ratio intruduced in #1766 . When the "average" value of dispatched/completed rates exceeds some threshold (configurable, 1.5 by default) the "cost" of individual requests increases thus reducing the dispatch rate.

The main difference from the current implementation is that the new backlink is not "immediate". The averaging is the exponential moving average filter with 100ms updates and 0.95 smoothing factor. Current backlink is immediate in a sense that delay to deliver a completion immediately slows down the next tick dispatch thus accumulating spontaneous reactor micro-stalls.

### DIFF
--- a/apps/io_tester/ioinfo.cc
+++ b/apps/io_tester/ioinfo.cc
@@ -21,6 +21,7 @@
 #include <seastar/core/app-template.hh>
 #include <seastar/core/thread.hh>
 #include <seastar/core/reactor.hh>
+#include <seastar/core/io_queue.hh>
 #include <seastar/util/closeable.hh>
 #include <yaml-cpp/yaml.h>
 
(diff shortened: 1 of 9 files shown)


xemul:
Testing it on i3en.3xl with some cstress reproducer of @vladzcloudius 's latency spike from left the past

The flow ratio introduced in #1766 is mostly above 1.0 under load, as was described in that PR. Good new is that it mostly stays well below 1.1, so the slow-down threshold from here can be restricted by default. OTOH, the CPU load in this test is ~30%, no time-spent-in-task-quota-violations :\

image

xemul:
image

Left part -- with this PR. Right part -- current master. The "backpressure" doesn't work in this experiment, because flow ratio is ~1.01 with the threshold of 1.5. But 2-bucket capping is OFF in this PR and this allows statement class to render shorter queue with smaller latencies

Issue #1150, opened 2022-07-22
Obfuscating batch flushes


xemul:
The `output_stream::flush()` looks like this

```c++
template <typename CharType>
future<> output_stream<CharType>::flush() noexcept {
    if (!_batch_flushes) {
        return do_sync_flush(_fd);
    } else {
        if (_ex) {
            // flush is a good time to deliver outstanding errors
            return make_exception_future<>(std::move(_ex));
        } else {
            _flush = true;
            if (!_in_batch) {
                add_to_flush_poller(this);
                _in_batch = promise<>();
            }
        }
    }
    return make_ready_future<>();
}
```

so if batch-flushing takes place the `stream.flush()` resolves immediately without exception. Later, the `.poll_flush()` may put and exception from `_fd.flush()`/`_fd.put()` int stream's `.ex` field, but nothing more.

Next, consider an RPC-like message exchange:

```c++
    while (true) {
        co_await client_stream.write();
        co_await client_stream.flush();
        co_await client_stream.recv();
        co_await handle();
    }
```

if sending a message in batch-flush fails no place here ever receives this exception:
* `.flush()` resolves without exception as was shown above
* `.recv()` hangs waiting for the response that may never arrive, because request sending failed
* `.handle()` won't get called
* `.write()` on next loop iteration _may_ notice the exception, but it won't be called either

xemul:
Found in #1146

PR #1776, opened 2023-08-08
http client: check for exception when closing streams

pgoron:
By not checking exceptions, we can end with abandoned exceptional futures.
Situation observed with https connections when tls handshake failed (expired certs for example), closing out stream will rethrow exception seen during last flush.

Ignoring exception during close seems ok to me. Alternatively
we could use when_all_succeed instead of when_all if we
want to report errors to client.

No UT added because http client doesn't seem to be covered yet.

### DIFF
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -171,11 +171,15 @@ input_stream<char> connection::in(reply& rep) {
 }
 
 future<> connection::close() {
-    return when_all(_read_buf.close(), _write_buf.close()).discard_result().then([this] {
-        auto la = _fd.local_address();
-        return std::move(_closed).then([la = std::move(la)] {
-            http_log.trace("destroyed connection {}", la);
-        });
+    return when_all(
+        _read_buf.close().handle_exception([] (auto ignored) {}),
+        _write_buf.close().handle_exception([] (auto ignored) {}))
+        .discard_result()
+        .then([this] {
+            auto la = _fd.local_address();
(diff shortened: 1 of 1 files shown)


src/http/client.cc
```diff
@@ -171,11 +171,15 @@ input_stream<char> connection::in(reply& rep) {
...
 
 future<> connection::close() {
-    return when_all(_read_buf.close(), _write_buf.close()).discard_result().then([this] {
-        auto la = _fd.local_address();
-        return std::move(_closed).then([la = std::move(la)] {
-            http_log.trace("destroyed connection {}", la);
-        });
+    return when_all(
+        _read_buf.close().handle_exception([] (auto ignored) {}),
+        _write_buf.close().handle_exception([] (auto ignored) {}))
```
xemul, line 176: That's the only place that can produce exception, isn't it? I mean that `input_stream::close()` is not resolving into exceptions. (#1150)

In fact, we're trying to make all close()-s not to resolve into exceptions, so it probably should be output_stream::close()'s fix, not http client's, all the more so rpc clients also close write streams without checking for exceptions

Cc @bhalevy , WDYT?

bhalevy: We definitely want read resource close to never return an error.
Historically, for files, we typically do flush in close, and flush may fail, but it's still the caller's responsibility to invoke flush/fsync before close if it is interested in its completion status, so I think it's OK to handle all exceptions in close and log a warning when ignoring them.

@avikivity are you okay with that?

PR #1775, opened 2023-08-08
tests: perf: shard_token_bucket: avoid capturing unused variables in lambdas

avikivity:
Clang complains.

Fixes 7051efb9d36d9ff219c41cf8904883b6c66d70ae.

PR #1782, opened 2023-08-09
tests: do not capture unused variables

tchaikov:
`this` is not used in some of these lambdas, so drop them. this should address the warnings like

```
/home/circleci/project/tests/perf/shared_token_bucket.cc:227:22: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
        return work([this] (std::chrono::duration<double> d) {
                     ^
```


BenPope:
I think this is a dupe of https://github.com/scylladb/seastar/pull/1775

tchaikov:
indeed. Thank you again =)

PR #1786, opened 2023-08-13
install-dependencies.sh: re-add protobuf

avikivity:
In preparation for re-introducing protobuf support to Prometheus, add the supporting libraries and compiler to the dependencies.

The testing toolchain is regenerated.


tchaikov (approved):
lgtm. will drop corresponding change in my change.

tchaikov:
> lgtm. will drop corresponding change in my change.

dropped.

PR #1787, opened 2023-08-15
smp: make service management semaphore thread local

dotnwat:
This PR is in support of running multiple instances of Seastar in a process. The smp service group management semaphore is only used on core 0. However, in a multi-instance scenario core 0 corresponds to different threads, one in each instance of seastar.

### DIFF
--- a/src/core/smp.cc
+++ b/src/core/smp.cc
@@ -59,7 +59,7 @@ struct smp_service_group_impl {
 #endif
 };
 
-static smp_service_group_semaphore smp_service_group_management_sem{1, named_semaphore_exception_factory{"smp_service_group_management_sem"}};
+static thread_local smp_service_group_semaphore smp_service_group_management_sem{1, named_semaphore_exception_factory{"smp_service_group_management_sem"}};
 static thread_local std::vector<smp_service_group_impl> smp_service_groups;
 
 static named_semaphore_exception_factory make_service_group_semaphore_exception_factory(unsigned id, shard_id client_cpu, shard_id this_cpu, std::optional<sstring> smp_group_name) {


src/core/smp.cc
```diff
@@ -59,7 +59,7 @@ struct smp_service_group_impl {
 #endif
 };
 
-static smp_service_group_semaphore smp_service_group_management_sem{1, named_semaphore_exception_factory{"smp_service_group_management_sem"}};
+static thread_local smp_service_group_semaphore smp_service_group_management_sem{1, named_semaphore_exception_factory{"smp_service_group_management_sem"}};
```
tchaikov, line 62: this create n more `smp_service_group_semaphore` instances in a seastar application where n is `smp::count - 1`, and these `smp_service_group_semaphore` instances are never used. see https://eel.is/c++draft/stmt.dcl#3 

so probably a better approach is to have a `static` function which returns the reference to an instance of `static thread_local` defined in that function?

avikivity: Those semaphores don't take up a significant amount of space, we can live with it.

Logically they belong in the smp instance, but it's not worth the trouble.

dotnwat: @tchaikov was reading that reference at the same time that avi merged (thanks!) but i'm curious about it still. are you saying that for global thread_locals that new threads automatically have the variable allocated for them without interaction, but with block scoped thread local the interaction must occur?

tchaikov: @dotnwat hi Noah, my reading of

> the first time control passes through its declaration

was that the "control" automatically passes all global variables in unspecified order, as per https://eel.is/c++draft/basic.start.dynamic#1. but i misinterpreted, as i thought "control" was equivalent to the definitions. i think in this context, "control" is somewhat like "odr-use". and i missed another point:

> It is implementation-defined whether the dynamic initialization of a non-block non-inline variable with static storage duration is sequenced before the first statement of main or is deferred[.](https://eel.is/c++draft/basic.start.dynamic#5.sentence-1)
> If it is deferred, it strongly happens before any non-initialization odr-use of any non-inline function or non-inline variable defined in the same translation unit as the variable to be initialized[.](https://eel.is/c++draft/basic.start.dynamic#5.sentence-2)[40](https://eel.is/c++draft/basic.start.dynamic#footnote-40)
> It is implementation-defined in which threads and at which points in the program such deferred dynamic initialization occurs[.](https://eel.is/c++draft/basic.start.dynamic#5.sentence-3)

see https://eel.is/c++draft/basic.start.dynamic#5

so, a standard compliant compiler *can* choose to initialize a `static` variable *after* `main()` and before its odr-use! i created a minimal reproducer to demonstrate this, see https://godbolt.org/z/WE5MWMGMs . per the experiment, both Clang and GCC choose to initialize the `static` thread_local variable lazily (after main and before odr-use).

thank you for pointing this out.

PR #1788, opened 2023-08-15
aio_general_context: flush: return short flush on error

### LINKED ISSUE #1784: Crash/assert in seastar::aio_general_context::flush(): Assertion `clock::now() < retry_until' failed.
Hi folks, 

in Redpanda we experience the following assert/crash under heavy load:

```
seastar::aio_general_context::flush(): Assertion `clock::now() < retry_until' failed.
```

https://github.com/scylladb/seastar/blob/master/src/core/reactor_backend.cc#L349-L356

```
[Backtrace #0]
void seastar::backtrace<seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}>(seastar::backtrace_buffer::append_backtrace()::{lambda(seastar::frame)#1}&&) at /v/build/v_deps_build/seastar-prefix/src/seastar/include/seastar/util/backtrace.hh:60
 (inlined by) seastar::backtrace_buffer::append_backtrace() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:786
 (inlined by) seastar::print_with_backtrace(seastar::backtrace_buffer&, bool) at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:816
 (inlined by) seastar::print_with_backtrace(char const*, bool) at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:828
seastar::sigabrt_action() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:4065
 (inlined by) operator() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:4041
 (inlined by) __invoke at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:4037
/opt/redpanda/lib/libc.so.6: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=6e7b96dfb83f0bdcb6a410469b82f86415e5ada3, for GNU/Linux 3.2.0, stripped

__sigaction at ??:?
pthread_key_delete at ??:?
raise at ??:?
abort at ??:?
?? ??:0
__assert_fail at ??:?
seastar::aio_general_context::flush() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor_backend.cc:338
 (inlined by) seastar::reactor_backend_aio::kernel_submit_work() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor_backend.cc:522
seastar::reactor::poll_once() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:3407
 (inlined by) operator() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:3300
 (inlined by) seastar::reactor::do_run() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:3328
operator() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/reactor.cc:4522
 (inlined by) decltype ((std::declval<seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_2&>())()) std::__1::__invoke[abi:v160004]<seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_2&>(seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_2&) at /vectorized/llvm/bin/../include/c++/v1/__functional/invoke.h:394
 (inlined by) void std::__1::__invoke_void_return_wrapper<void, true>::__call<seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_2&>(seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_2&) at /vectorized/llvm/bin/../include/c++/v1/__functional/invoke.h:487
 (inlined by) std::__1::__function::__alloc_func<seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_2, std::__1::allocator<seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_2>, void ()>::operator()[abi:v160004]() at /vectorized/llvm/bin/../include/c++/v1/__functional/function.h:185
 (inlined by) std::__1::__function::__func<seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_2, std::__1::allocator<seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_2>, void ()>::operator()() at /vectorized/llvm/bin/../include/c++/v1/__functional/function.h:356
std::__1::__function::__value_func<void ()>::operator()[abi:v160004]() const at /vectorized/llvm/bin/../include/c++/v1/__functional/function.h:510
 (inlined by) std::__1::function<void ()>::operator()() const at /vectorized/llvm/bin/../include/c++/v1/__functional/function.h:1156
 (inlined by) seastar::posix_thread::start_routine(void*) at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/posix.cc:73
pthread_condattr_setpshared at ??:?
__xmknodat at ??:?
```

I don't have an easy reproducer but I can reliably reproduce it with one of our internal load tests if further debugging is needed.

This bit of code seems to have a bit of history and was already changed a couple times:
 - https://github.com/scylladb/seastar/commit/1fb42e2776239341935efd7700336ec5eb9bc253
 - https://github.com/scylladb/seastar/commit/528762adbaf2c6dde857957b920606378959e8fb

I am not sure I fully understand the code. The comment mentions an "infinite loop" but I guess what it means is a longer stall in `flush`? I assume the 1 second was chosen arbitrarily? IMO stalling for 1 (or 5 or 10) seconds is better than crashing?

cc @bhalevy 

Thanks,
Stephan

bhalevy:
As seen in https://github.com/scylladb/seastar/issues/1784, having flush retry `io_submit` in a loop might be futile if the kernel internal queue is full and we need to reap completions to free up space for more iocb:s.

This change breaks out of the loop as soon as
`io_submit` fails and just moves the remaining entries in the iocbs queue into the front so more entries can be queued (otherwise, we'd need a circular queue that will penalize the error-free path for the unlikely event of an error)

Fixes scylladb/seastar#1784

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -450,7 +450,14 @@ size_t aio_general_context::flush() {
             begin += r;
             continue;
         }
-        // errno == EAGAIN is expected here. We don't explicitly assert that
+        SEASTAR_ASSERT(r < 0);
+        // EAGAIN is expected here when "Insufficient resources are available to queue any iocbs" (see io_submit(2)).
+        // This indicates overload on the kernel internal queue.
+        // Returning early below will allow us to reap completions to free up resources for further iocb:s.
+        // Abort on any other error, as those indicate an internal error on our side.
+        if (errno != EAGAIN) {
+            on_fatal_internal_error(seastar_logger, format("aio_general_context::flush: io_submit failed with unexpected system error: {}", errno));
+        }
         // just adjust the queue and return the number of iocbs we were able to submit.
         std::move(begin, last, iocbs.get());
         break;
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -444,9 +444,6 @@ void aio_general_context::queue(linux_abi::iocb* iocb) {
 
 size_t aio_general_context::flush() {
     auto begin = iocbs.get();
-    using clock = std::chrono::steady_clock;
-    constexpr clock::time_point no_time_point = clock::time_point(clock::duration(0));
-    auto retry_until = no_time_point;
     while (begin != last) {
         auto r = io_submit(io_context, last - begin, begin);
         if (__builtin_expect(r > 0, true)) {
@@ -454,16 +451,12 @@ size_t aio_general_context::flush() {
             continue;
         }
         // errno == EAGAIN is expected here. We don't explicitly assert that
-        // since the assert below prevents an endless loop for any reason.
-        if (retry_until == no_time_point) {
-            // allow retrying for 1 second
-            retry_until = clock::now() + 1s;
-        } else {
-            SEASTAR_ASSERT(clock::now() < retry_until);
-        }
+        // just adjust the queue and return the number of iocbs we were able to submit.
+        std::move(begin, last, iocbs.get());
+        break;
     }
-    auto nr = last - iocbs.get();
-    last = iocbs.get();
+    auto nr = begin - iocbs.get();
+    last -= nr;
     return nr;
 }
 
@@ -543,6 +536,11 @@ void preempt_io_context::request_preemption() {
     // monitored.
     _hrtimer_aio_completion.maybe_queue(_context);
     _context.flush();
+    // A short flush leaves the hrtimer iocb queued in userspace, so the aio
+    // ring cannot make need_preempt() true yet.
+    if (_context.has_pending()) {
+        return;
+    }
 
     // The kernel is not obliged to deliver the completion immediately, so wait for it
     while (!need_preempt()) {
@@ -655,7 +653,7 @@ bool reactor_backend_aio::kernel_submit_work() {
 }
 
 bool reactor_backend_aio::kernel_events_can_sleep() const {
-    return _storage_context.can_sleep();
+    return !_polling_io.has_pending() && _storage_context.can_sleep();
 }
 
 void reactor_backend_aio::wait_and_process_events(const sigset_t* active_sigmask) {
@@ -668,6 +666,9 @@ void reactor_backend_aio::wait_and_process_events(const sigset_t* active_sigmask
     _hrtimer_poll_completion.maybe_queue(_polling_io);
     _smp_wakeup_aio_completion.maybe_queue(_polling_io);
     _polling_io.flush();
+    if (_polling_io.has_pending()) {
+        timeout = 0;
+    }
     await_events(timeout, active_sigmask);
     _preempting_io.service_preempting_io(); // clear task quota timer
 }
--- a/src/core/reactor_backend.hh
+++ b/src/core/reactor_backend.hh
@@ -70,8 +70,12 @@ struct aio_general_context {
     internal::linux_abi::iocb** last;
     internal::linux_abi::iocb** const end;
     void queue(internal::linux_abi::iocb* iocb);
-    // submit all queued iocbs and return their count.
+    // Submit queued iocbs until the queue is drained or io_submit() reports
+    // EAGAIN. Returns the number submitted; unsubmitted iocbs remain queued.
     size_t flush();
+    bool has_pending() const {
+        return last != iocbs.get();
+    }
 };
 
 class aio_storage_context {


src/core/reactor_backend.cc
```diff
@@ -337,26 +337,18 @@ void aio_general_context::queue(linux_abi::iocb* iocb) {
...
-        // since the assert below prevents an endless loop for any reason.
-        if (retry_until == no_time_point) {
-            // allow retrying for 1 second
-            retry_until = clock::now() + 1s;
-        } else {
-            assert(clock::now() < retry_until);
+        if (__builtin_expect(r <= 0, false)) {
+            // errno == EAGAIN is expected here, but we don't explicitly assert that.
+            // just adjust the queue and return the number of iocbs we were able to submit
+            std::memmove(iocbs.get(), begin, (last - begin) * sizeof(iocb*));
```
tchaikov, line 345: nit,
```c++
std::move(begin, last, iocbs.get());
```

src/core/reactor_backend.cc
```diff
@@ -337,26 +337,18 @@ void aio_general_context::queue(linux_abi::iocb* iocb) {
...
-            continue;
-        }
-        // errno == EAGAIN is expected here. We don't explicitly assert that
-        // since the assert below prevents an endless loop for any reason.
-        if (retry_until == no_time_point) {
-            // allow retrying for 1 second
-            retry_until = clock::now() + 1s;
-        } else {
-            assert(clock::now() < retry_until);
+        if (__builtin_expect(r <= 0, false)) {
```
StephanDollberg, line 342: Would still make sense to check/assert/retry for errors other than EAGAIN? 

From what I read from the man page most of those are probably logic errors.

tchaikov: i'd vote for `assert()` . it's guarded by `__builtin_expect(r <= 0, false)`, so presumably this should not hurt the performance that much, considering that we always undefine `NDEBUG`.

bhalevy: Can do.

bhalevy:
In v2 ([2cbb4cc](https://github.com/scylladb/seastar/commit/2cbb4cccec913f9a74c09e7bca88483175384cf4)):
- use `std::move` rather than `memmove`
- verify `errno` and abort using `on_fatal_internal_error` if it is not `EAGAIN.`

tchaikov (approved):

src/core/reactor_backend.cc
travisdowns: r==0 should be impossible (only occurs if we try to submit 0 iocbs, which we don't don't). So perhaps we should include it in the assert case or have a separate assert case.

bhalevy:
Updated the branch after review feedback.

Changes since the previous push:
- folded the pending-IOCB sleep guard into the short-flush patch so the reactor does not enter a blocking wait while polling IOCBs remain queued in userspace
- folded the preemption-context guard into the same patch so request_preemption() does not spin waiting for an hrtimer IOCB that was not submitted
- kept the unexpected-error handling as the second patch
- updated the short-flush commit message to document both hazards

Local verification:
- git diff --check origin/master..HEAD
- ninja -C build/dev CMakeFiles/seastar.dir/src/core/reactor_backend.cc.o

src/core/reactor_backend.cc
bhalevy: > r==0 should be impossible (only occurs if we try to submit 0 iocbs, which we don't don't). So perhaps we should include it in the assert case or have a separate assert case.

Done in [2cbb4cc](https://github.com/scylladb/seastar/commit/2cbb4cccec913f9a74c09e7bca88483175384cf4)

bhalevy:
@xemul please review

StephanDollberg:
We have been running with the two original patches for a couple years now. Once you have those two in you'll quickly need this third one on top as the asserts just shift: https://github.com/redpanda-data/seastar/commit/f922a4bb5b0c06dc7bde48b14dfe2842efc03c81

Note there is a problem with all these patches that seems inherent to the way the reactor is built and how the aio/iocb stuff works.

These three patches handle the case well where there is a short temporary spike in IOCB demand. Instead of crashing we just churn through all the IOCBs.

However, what this patch series still doesn't handle is when there is constant IOCB demand higher than `max-networking-io-control-blocks` (many concurrent pending reads or wait_input_shutdown). In that case there is basically a constant oustanding queue which possibly only moves slowly. In all likely hood this results in a degraded system as important things like node heartbeats are queue behind other less important connections. Observability into this state is also very poor. This also combines poorly with `max-networking-io-control-blocks` silently getting truncated when max-aio-nr is low or smp count is high.

Nevertheless because the first scenario is not uncommon and it's much better to just absorb these spikes than crash we are still running with these patches. But it's not a fix-all state.

avikivity:
> We have been running with the two original patches for a couple years now. Once you have those two in you'll quickly need this third one on top as the asserts just shift: [redpanda-data@f922a4b](https://github.com/redpanda-data/seastar/commit/f922a4bb5b0c06dc7bde48b14dfe2842efc03c81)
> 
> Note there is a problem with all these patches that seems inherent to the way the reactor is built and how the aio/iocb stuff works.
> 
> These three patches handle the case well where there is a short temporary spike in IOCB demand. Instead of crashing we just churn through all the IOCBs.
> 
> However, what this patch series still doesn't handle is when there is constant IOCB demand higher than `max-networking-io-control-blocks` (many concurrent pending reads or wait_input_shutdown). In that case there is basically a constant oustanding queue which possibly only moves slowly. In all likely hood this results in a degraded system as important things like node heartbeats are queue behind other less important connections. Observability into this state is also very poor. This also combines poorly with `max-networking-io-control-blocks` silently getting truncated when max-aio-nr is low or smp count is high.

My recommendation is to try out the assymmetric io_uring backend [*]. It should not have any of these problems.

> 
> Nevertheless because the first scenario is not uncommon and it's much better to just absorb these spikes than crash we are still running with these patches. But it's not a fix-all state.


[*] the secret agenda is that you hit all the bugs first

bhalevy:
@avikivity what do you suggest doing with this change?

travisdowns:
> This also combines poorly with `max-networking-io-control-blocks` silently getting truncated when max-aio-nr is low or smp count is high.
> 
> Nevertheless because the first scenario is not uncommon and it's much better to just absorb these spikes than crash we are still running with these patches. But it's not a fix-all state.

Yeah, short of "use a different backend" there doesn't seem to be a great solution here, even with "unlimited budget" for changes. The solution space seems to be:

1) Current (master) approach: crash randomly when the number of in-flight iocbs happens to tips over the high water mark.

2) This fix (+ Stephan patches). Slow down (sometimes dramatically) when the number of in-flight iocbs is over the HWM in a sustained matter (but blips are fine). Arguably a strictly better behavior than the current one unless maybe you want to avoid grey-failures which this can allow.

3) Once we know how many iocbs we got, cap up front the number of network connections allowed at a conservative amount that will never exceed the iocb count, which basically means iocb count / 3, (or iocb count / 1.5 if you want to rely on the existing kernel behavior which effectively doubles the count internally). Connections above that can fail, this doesn't need to be an assert, it could be an exception. This can also lead to a type of grey failure where a server is up, but rejecting some fraction of connections, but it's a better more than (2) I think. The main downside is this limit is fairly conservative because at runtime many applications never reach anywhere close to connections * 3 IOCBs, so you might cut the effective connection handling capability (at the limit) down by 10x or something if you implement this. Though it's OK to ask the user to just increase the kernel values, I think. The downside seems low.

> My recommendation is to try out the assymmetric io_uring backend [*]. It should not have any of these problems.

Right. I guess io_uring just doesn't have this issue at all, since it doens't consume a cb (or similar resource) for polling a socket at all. Furthermore, the asymmetric version also avoids the problem with increasing `--smp` where we statically partition the cb pool among shards which means each shard gets a small slice which means problems sooner where there is inter-shard variance.

src/core/reactor_backend.cc
```diff
@@ -444,26 +444,19 @@ void aio_general_context::queue(linux_abi::iocb* iocb) {
 
 size_t aio_general_context::flush() {
```
avikivity, line 445: > As seen in https://github.com/scylladb/seastar/issues/1784, having flush retry `io_submit` in a loop might be futile if the kernel internal queue is full and we need to reap completions to free up space for more iocb:s.

How is it possible? The number of IOCBs is equal to the ring size.

avikivity:
> @avikivity what do you suggest doing with this change?

Explain it better, I don't understand what it is fixing.

travisdowns:
@avikivity wrote:

> Explain it better, I don't understand what it is fixing.

The reactor is crashing because the number of requests in flight plus the number we are trying to submit in this poll is larger than the number of available aio kernel slots/cq size (these two are bound together). So `io_submit` returns EAGAIN but we don't do anything to resolve the condition. This happens because we set the CQ size based on "max networking iocbs" but there is nothing in seastar which actually limits the iocbs to that number.

So you set max networking iocbs to 1000, we allocate 1000 CQ (there is some more math in here, rounding up to page size, etc), then you poll read on 2000 connections and you will crash around connection 1000 because there are no more slots in the kernel/CQ, so io_submit will return EGAIN and we get the associated assert.

travisdowns:
This reproduces trivially BTW, by opening and reading socket pairs, no need for a load test.

travisdowns:
I, Claude, am posting the math and a self-contained reproducer here since I think it makes a strong case for this PR's approach — in particular it shows that once this state is reached, the current 1-second retry loop in `flush()` can *never* succeed, so the assert is a deterministic abort rather than a "we were briefly slow" back-stop.

## Why io_submit returns EAGAIN

In `fs/aio.c`, every in-flight iocb reserves one completion-ring slot at `io_submit` time (`get_reqs_available`), and a slot is only released when userspace *consumes* the completion, i.e. when the ring head advances (`refill_reqs_available` subtracts events still sitting in the ring). Completion alone does not free a slot.

An `IOCB_CMD_POLL` on an idle socket never completes, so every parked network poll holds its slot indefinitely. The polling context's ring is sized from `max-networking-io-control-blocks`, but nothing bounds the number of in-flight polls to that value: each `aio_pollable_fd_state` embeds up to 3 iocbs and every idle connection parks at least one POLLIN poll, so in-flight polls scale with open fds, not with the configured budget. (Contrast the storage context, whose `iocb_pool` caps in-flight iocbs at `max_aio` and stops draining the sink when the pool is empty — the networking side has no equivalent accounting, so the first "enforcement" of the option is the kernel's EAGAIN.)

## Ring sizing math

```
io_setup(nr)  →  ioctx_alloc:     nr' = max(nr, 4 × num_possible_cpus) × 2
                 aio_setup_ring:  nr' += 2; capacity = (pages(32 + 32·nr')·4096 − 32) / 32
                 usable = capacity − 1     (reqs_available is initialized to nr_events − 1)
```

Measured on a 6.17 kernel, 32 possible CPUs (by reading `ring->nr` from the mmap'd ring header and submitting idle polls until EAGAIN):

| requested nr | ring capacity | EAGAIN at (in-flight polls) |
|---|---|---|
| 50 | 383 | 382 |
| 1000 | 2047 | 2040 |
| 10000 (default) | 20095 | 19819 |

(The shortfall below capacity−1 on larger rings is per-CPU `req_batch` stranding.)

So with the default configuration, one shard aborts once it accumulates ~19.8k parked polls — roughly 10-20k idle-ish connections depending on how many fds have both read- and write-side polls parked, which is exactly the profile of the crashes we see.

## Reproducer

Gist with two programs: https://gist.github.com/travisdowns/33ac15cc61e31675782cb969090c322f

1. A raw C program (no seastar) that demonstrates the kernel behaviour and measures the table above, including retrying `io_submit` for 1.5 s after the first EAGAIN (~7M attempts): every single one fails.
2. A seastar app that opens idle loopback connections, each parking two never-completing POLLIN polls (client fd + server fd):

```
./aio_eagain_repro --smp 1 --reactor-backend linux-aio \
    --max-networking-io-control-blocks 50 --conns 2000
```

The crash point is predictable to within a couple of connections: usable slots = 382, permanent holders = 2N parked polls + 3 housekeeping polls (listener accept, hrtimer, smp-wakeup eventfd), so it should die opening connection ~190 — observed abort right after connection 189:

```
opened 189 connections (380 parked polls)
src/core/reactor_backend.cc:462: size_t seastar::aio_general_context::flush():
    Assertion `clock::now() < retry_until` failed.
Aborting on shard 0 ...
```

With `--max-networking-io-control-blocks 1000` (usable = 2046, predicted N ≈ 1021) it dies at connection 1019, the last few slots going to `req_batch` stranding.

## Why the 1s retry can never succeed

Once `flush()` enters its retry loop it never calls `io_getevents`, so the ring head never advances and no slot can ever be freed. Even completions from *busy* fds don't help: they are appended to the ring but keep holding their slots until reaped, and the kernel's `user_refill_reqs_available` only reclaims events userspace has already consumed. So after the first EAGAIN with a ring full of parked polls, the abort 1 s later is guaranteed — there is no load level or timing that escapes it, which also means bumping the retry window would only convert the crash into a 100% reactor stall of the same duration.

Given that, returning early from `flush()` and keeping the unsubmitted iocbs queued in userspace (plus the no-sleep-while-pending guard) seems like the right minimal fix; the longer-term alternative would be real accounting of in-flight network iocbs against the configured budget, the way the storage context's `iocb_pool` already does.

src/core/reactor_backend.cc
```diff
@@ -444,26 +444,26 @@ void aio_general_context::queue(linux_abi::iocb* iocb) {
...
+        // EAGAIN is expected here when "Insufficient resources are available to queue any iocbs" (see io_submit(2)).
+        // This indicates overload on the kernel internal queue.
+        // Returning early below will allow us to reap completions to free up resources for further iocb:s.
+        // Abort on any other error, as those indicate an internal error on our side.
+        if (errno != EAGAIN) {
+            on_fatal_internal_error(seastar_logger, format("aio_general_context::flush: io_submit failed with unexpected system error: {}", errno));
         }
+        // just adjust the queue and return the number of iocbs we were able to submit.
+        std::move(begin, last, iocbs.get());
+        break;
```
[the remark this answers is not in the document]
travisdowns: not sure where you are getting your info, `std::move` is safe for left shift

PR #1792, opened 2023-08-21
test: add fixture support to seastar::testing

tchaikov:
before this change, because of the missing support of test fixture,
some tests in scylladb are using lambdas or homebrew test helper
class for hosting the helpers and for creating the test fixture,
so they can be shared in a test case or across test cases. this
hurts the readability. the better practice would be to have smaller
tests which share the same fixture explicitly.

so, in this change, 

* the `const` specifier is removed from `seastar_test::run_test_case()`
* two more macros are added so tests can be defined with fixtures:
  - SEASTAR_FIXTURE_TEST_CASE
  - SEASTAR_THREAD_FIXTURE_TEST_CASE
* tls_test.cc is updated to use `SEASTAR_THREAD_FIXTURE_TEST_CASE`

### DIFF
--- a/include/seastar/testing/test_case.hh
+++ b/include/seastar/testing/test_case.hh
@@ -33,14 +33,14 @@
 #define SEASTAR_TEST_CASE_WITH_DECO(name, decorators)               \
     struct name : public seastar::testing::seastar_test {           \
         using seastar::testing::seastar_test::seastar_test;         \
-        seastar::future<> run_test_case() const override;           \
+        seastar::future<> run_test_case() override;                 \
     };                                                              \
     static const name name ## _instance(                            \
         #name,                                                      \
         __FILE__,                                                   \
         __LINE__,                                                   \
         decorators); /* NOLINT(cert-err58-cpp) */                   \
-    seastar::future<> name::run_test_case() const
+    seastar::future<> name::run_test_case()
 
 #define SEASTAR_TEST_CASE_WITHOUT_DECO(name)                        \
     SEASTAR_TEST_CASE_WITH_DECO(                                    \
--- a/include/seastar/testing/test_case.hh
+++ b/include/seastar/testing/test_case.hh
@@ -29,6 +29,7 @@
 #include <seastar/core/future.hh>
 
 #include <seastar/testing/seastar_test.hh>
+#include <utility>
 
 #define SEASTAR_TEST_CASE_WITH_DECO(name, decorators)               \
     struct name : public seastar::testing::seastar_test {           \
@@ -53,4 +54,77 @@
             BOOST_PP_EQUAL(BOOST_PP_VARIADIC_SIZE(__VA_ARGS__), 1), \
             SEASTAR_TEST_CASE_WITHOUT_DECO,                         \
             SEASTAR_TEST_CASE_WITH_DECO),                           \
+        __VA_ARGS__)
+
+namespace seastar::testing::internal {
+    template<typename T, typename = void>
+    struct has_setup : std::false_type {};
+    template<typename T>
+    struct has_setup<T, std::void_t<decltype(std::declval<T>().setup())>> : std::true_type {};
+    template<typename T>
+    seastar::future<> setup_conditional(T& t) {
+        if constexpr(has_setup<T>::value) {
+            return t.setup();
+        }
+        return seastar::make_ready_future();
+    }
+
+    template<typename T, typename = void>
+    struct has_teardown : std::false_type {};
+    template<typename T>
+    struct has_teardown<T, std::void_t<decltype(std::declval<T>().teardown())>> : std::true_type {};
+    template<typename T>
+    seastar::future<> teardown_conditional(T& t) {
+        if constexpr(has_setup<T>::value) {
+            return t.teardown();
+        }
+        return seastar::make_ready_future();
+    }
+}
+
+#define SEASTAR_FIXTURE_TEST_CASE_WITH_DECO(name, F, decorators)        \
+    struct name : public seastar::testing::seastar_test,                \
+                  private F {                                           \
+        using seastar::testing::seastar_test::seastar_test;             \
+        seastar::future<> run_test_case() override;                     \
+    private:                                                            \
+        seastar::future<> do_run_test_case() const;                     \
+    };                                                                  \
+    static const name name ## _instance(                                \
+        #name,                                                          \
+        __FILE__,                                                       \
+        __LINE__,                                                       \
+        decorators); /* NOLINT(cert-err58-cpp) */                       \
+    seastar::future<> name::run_test_case() {                           \
+        BOOST_TEST_CHECKPOINT('"' << #name << "\" fixture ctor");       \
+        BOOST_TEST_CHECKPOINT('"' << #name << "\" fixture setup");      \
+        /* seastar_test does not have  adefault ctor,  */               \
+        /* but setup_conditional needs to construct the fixture  */     \
+        /* to check if setup() is available, so cast it to F */         \
+        F& fixture = *this;                                             \
+        return testing::internal::setup_conditional(fixture).then([this] { \
+            BOOST_TEST_CHECKPOINT('"' << #name << "\" test entry");     \
+            return do_run_test_case();                                  \
+        }).finally([this] {                                             \
+            BOOST_TEST_CHECKPOINT('"' << #name << "\" fixture teardown"); \
+            F& fixture = *this;                                         \
+            return testing::internal::teardown_conditional(fixture);    \
+        });                                                             \
+        BOOST_TEST_CHECKPOINT('"' << #name << "\" fixture dtor");       \
+        return seastar::make_ready_future();                            \
+    }                                                                   \
+    seastar::future<> name::do_run_test_case() const
+
+
+#define SEASTAR_FIXTURE_TEST_CASE_WITHOUT_DECO(name, F)             \
+    SEASTAR_FIXTURE_TEST_CASE_WITH_DECO(                            \
+        name, F,                                                    \
+        boost::unit_test::decorator::collector_t::instance())
+
+#define SEASTAR_FIXTURE_TEST_CASE(...)                              \
+    SEASTAR_TEST_INVOKE(                                            \
+        BOOST_PP_IIF(                                               \
+            BOOST_PP_EQUAL(BOOST_PP_VARIADIC_SIZE(__VA_ARGS__), 2), \
+            SEASTAR_FIXTURE_TEST_CASE_WITHOUT_DECO,                 \
+            SEASTAR_FIXTURE_TEST_CASE_WITH_DECO),                   \
         __VA_ARGS__)
--- a/tests/unit/tls_test.cc
+++ b/tests/unit/tls_test.cc
@@ -22,6 +22,7 @@
 
 #include <iostream>
 
+#include <memory>
 #include <seastar/core/do_with.hh>
 #include <seastar/core/sstring.hh>
 #include <seastar/core/reactor.hh>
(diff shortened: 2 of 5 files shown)


tchaikov:
inspired by https://github.com/scylladb/scylladb/blob/23be6f03360ba4943da55ac5128c7f6398264476/test/boost/sstable_compaction_test.cc#L939-L1007

tests/unit/tls_test.cc
```diff
@@ -268,36 +269,43 @@ class https_server {
...
 
-SEASTAR_THREAD_TEST_CASE(test_simple_x509_client) {
+SEASTAR_THREAD_FIXTURE_TEST_CASE(test_simple_x509_client, with_https_server) {
     auto certs = ::make_shared<tls::certificate_credentials>();
-    https_server server;
-    certs->set_x509_trust_file(server.cert(), tls::x509_crt_format::PEM).get();
-    connect_to_ssl_addr(certs, server.addr(), server.name()).get();
+    certs->set_x509_trust_file(server->cert(), tls::x509_crt_format::PEM).get();
+    connect_to_ssl_addr(certs, server->addr(), server->name()).get();
 }
```
avikivity, line 288: I understand the motivation, but the macro proliferation makes it harder to understand the role of the inputs to SEASTAR_THREAD_FIXURE_TEST_CASE.

If we stick to macro-less C++, we'd write `with_https_server_thread(std::function<void (https_server&)>). ` and use that in the test case. Isn't that easier to read?


/cc @nyh

tchaikov: i see, you are following the route of `test_env::do_with_async()`.

the reason to change `tls_test.cc` is just to demonstrate how `SEASTAR_THREAD_FIXURE_TEST_CASE` works. but it might not be the best place for this. i can drop this change if it hurts the readability in this very case.

what i really want is to split https://github.com/scylladb/scylladb/blob/23be6f03360ba4943da55ac5128c7f6398264476/test/boost/sstable_compaction_test.cc#L939-L1007 into smaller tests. i think there are two places where we can improve the test:

1. a fixture which provides the `test_env`. we already have a `test_env` . despite that the way how we customize it is a little bit awkward -- to pass a 2nd parameter after the huge body of the test.  
2. a helper class which provides all the helper functions. these helper functions are now implemented as lambdas. 

by introducing these macros to Seastar, we should be able to use them in other tests as well. i understand that the fixture macro might hurt the readability, but this pattern is not invented here. but it depends on if developers are used to the fixture macros in other testing framework. AFAICT, it is already widely used by popular test frameworks like [GoogleTest](http://google.github.io/googletest/reference/testing.html#TEST_F) and [Boost.Test](https://live.boost.org/doc/libs/1_83_0/libs/test/doc/html/boost_test/tests_organization/fixtures/case.html). so as long as developers are familiar with these test frameworks, they can apply what they've been using when writing tests for Seastar applications.

include/seastar/testing/test_case.hh
```diff
@@ -54,3 +55,76 @@
...
+        BOOST_TEST_CHECKPOINT('"' << #name << "\" fixture setup");      \
+        /* seastar_test does not have  adefault ctor,  */               \
+        /* but setup_conditional needs to construct the fixture  */     \
+        /* to check if setup() is available, so cast it to F */         \
+        F& fixture = *this;                                             \
+        return testing::internal::setup_conditional(fixture).then([this] { \
+            BOOST_TEST_CHECKPOINT('"' << #name << "\" test entry");     \
+            return do_run_test_case();                                  \
+        }).finally([this] {                                             \
+            BOOST_TEST_CHECKPOINT('"' << #name << "\" fixture teardown"); \
```
nyh, line 109: If I understand correctly, these fixtures are per-test-function fixtures, similar to pytest's "function-scoped" fixtures. They aren't shared by multiple test function. In that case, I honestly don't understand what's the point. In C++, you have RAII to do that - you can begin a test with a declaration `with_http_server h;` which will cause this type's constructor to run in the beginning at the test, and its destructor when the test ends. What does your fixture idea gain beyond this, or make more convenient?

tests/unit/tls_test.cc
tchaikov: > what i really want is to split https://github.com/scylladb/scylladb/blob/23be6f03360ba4943da55ac5128c7f6398264476/test/boost/sstable_compaction_test.cc#L939-L1007 into smaller tests. i think there are two places where we can improve the test:
> 

i think we are talking about different things. could you please take a look my reply above?

nyh: My point is that if you have specific goals in mind why you think fixtures are useful, please demonstrate **these** goals, and not a non-interesting example. You said it helps with test_env but I didn't understand how exactly. Is the main goal of the fixtures to reduce one (or several) layers of indentation with ugly lambda? If that's the case, you should show an example that does this (your example didn't).
The "helper functions" thing doesn't make sense to me. C++ already has plenty of features to do this - you can have file-static functions (this old-school approach is underappreciated today, I don't know why). You can have classes. You can instantiate classes with RAII. I still don't understand how fixtures help for that.

In general, my main point is that your example should demonstrate what you consider the best (or just "good") use of fixtures. Not some use case that doesn't appear useful.

avikivity: Maybe the fixures help with reuse? Since they advertise what object the test wants, the test framework can provide a pre-existing object.

But this should be written in C++, not macros.

tchaikov: > In general, I always ask that people who add test-framework improvements do this together with an actual use for these improvements. In this example, I don't see anything in the post-fixture version of the example you used that was better than we had before (at least, I didn't understand where it is better). Can you perhaps use a fixture in a different test which make it more obvious how wonderful this feature is and how much better/shorter/clearer/safer/whatever it makes tests?

i explained in previous comment that 

> the reason to change tls_test.cc is just to demonstrate how SEASTAR_THREAD_FIXURE_TEST_CASE works. but it might not be the best place for this. i can drop this change if it hurts the readability in this very case.

i haven't found a test in Seastar which replicate the patter of https://github.com/scylladb/scylladb/blob/23be6f03360ba4943da55ac5128c7f6398264476/test/boost/sstable_compaction_test.cc#L939-L1007, probably i didn't try hard enough.

> 
> From my experience with pytest, the most important feature of fixtures is the ability to share them between multiple tests - e.g., set up a table once and then let 10 different test functions use the same table. This allows the developer to create many very small tests, without paying overhead (like creating the same table again and again 10 times). ‎The result is that instead of having huge test functions that test 20 different things (like happens often in dtest), you can write very small test functions that each tests just one small feature - and when you have a failure you can see immediately where it failed. If I understand correctly, the fixture feature you added here - in the way you added it - doesn't support this ability. These fixtures get created and destroyed separately for each test. My question then is why are they any better than RAII, which already allows C++ code to create something in a beginning of a test and ensure it gets destroyed when the test ends no matter how it gets destroyed.

you are talking about a fixture with session or suite scope. this is not the pain point i am trying to address in this change.

tchaikov: > Maybe the fixures help with reuse? Since they advertise what object the test wants, the test framework can provide a pre-existing object.

exactly. to be more specific, by "reuse",  we are not reusing the same object across different test cases. but to reuse the code.

tchaikov:
@nyh and @avikivity thank you for your inputs and reviews. i am closing this PR for following reasons

1. Nadav is more interested in the session / suite scoped fixture . i am afraid that it's non trivial to add this to the existing structure how Seastar registers and launches the test cases. and this is not in the scope of this PR.
2. i haven't found a great demo apart from the one i found in scylladb. so i might not be able to convince the review the usage of the change in the desired approach. and yes, i can always convert the test in scylladb to use a helper class. probably it's just me who is missing the suites/fixtures macros in Boost.Test / Google Test.
3. Avi wants less macro and more C++.  i concur with him. but i need more time to make this happen.

tests/unit/tls_test.cc
nyh: > you are talking about a fixture with session or suite scope. this is not the pain point i am trying to address in this change.

I see. So it's still not clear to me what pain point you are solving that RAII didn't. But in any case, since you're inventing a new fixture API, and since shared ("session scope") fixtures are part of the fixture feature in every test framework I know, shouldn't you at least think ahead how your API will fit that case as well?

And I agree with Avi, let's try to think before we invent new macros whether the same thing can be implemented using C++ features such as RII, inheritance or templates.

nyh: > > Maybe the fixures help with reuse? Since they advertise what object the test wants, the test framework can provide a pre-existing object.
> 
> exactly. to be more specific, by "reuse", we are not reusing the same object across different test cases. but to reuse the code.

But to reuse code, why isn't it enough to have global functions (not lambdas) or classes with methods, and so on? I still don't understand the use case, I'm afraid.

nyh:
>     1. Nadav is more interested in the session / suite scoped fixture . i am afraid that it's non trivial to add this to the existing structure how Seastar registers and launches the test cases. and this is not in the scope of this PR.

Imagine that you have fixture which is **a class** xyz_fixture, and the code does this:

```
test_case_1(){
    xyz_fixture xyz;
   ... do something with xyz
}

test_case_1(){
    xyz_fixture xyz;
    ...do something with xyz
}
```

Now, `xyz_fixture` can be RAII (running in a seastar::thread!!) which when constructed creates some data, and when destructed (automatically at the end of the test) destroys it.
Moreover later (we don't need to do it now), xyz_fixture can become shared across all tests by adding some common code (e.g., inherit from some common class suite_fixture) which doesn't call the teardown() function at the end of the test and instead saves the object it in some global cache where it can be reused - and only frees stuff in this cache after all tests).

>     2. i haven't found a great demo apart from the one i found in scylladb. so i might not be able to convince the reviewers the value of the change in the desired approach. and yes, i can always convert the test in scylladb to use a helper class. probably it's just me who is missing the suites/fixtures macros in Boost.Test / Google Test.

You can show us an example in ScyllaDB.

>     3. Avi wants less macro and more C++.  i concur with him. but i need more time to make this happen.

tests/unit/tls_test.cc
tchaikov: yeah, i agree. the global functions and helper classes can also achieve this goal. and fixture, probably, should provide data or runtime facilities, i.e., fixtures, not the code which manipulates them.

i've been using fixtures for both purposes in previous projects. so i wanted to keep using fixtures in scylladb the same way. alas, probably, i should go with a helper class/function as you suggest. and leave C++ implementation for another day, when we really need the fixture support in Seastar.

tchaikov: > I see. So it's still not clear to me what pain point you are solving that RAII didn't. But in any case, since you're inventing a new fixture API, and since shared ("session scope") fixtures are part of the fixture feature in every test framework I know, shouldn't you at least think ahead how your API will fit that case as well?

i did think about this. i even read related code and document in Boost.Test. but i didn't dump what i had thought or researched in the commit message or the code if it does not help. i don't really want to make the cover letter looks like the first part of a thesis. it is not closely related to the PR and it does not contribute to the design decision. sorry for repeating my comment, i just didn't think it's in the scope of this PR.

tchaikov:
> > ```
> > 1. Nadav is more interested in the session / suite scoped fixture . i am afraid that it's non trivial to add this to the existing structure how Seastar registers and launches the test cases. and this is not in the scope of this PR.
> > ```
> 
> 
> Imagine that you have fixture which is **a class** xyz_fixture, and the code does this:
> 
> 
> ```
> test_case_1(){
>     xyz_fixture xyz;
>    ... do something with xyz
> }
> 
> test_case_1(){
>     xyz_fixture xyz;
>     ...do something with xyz
> }
> ```
> 

this is exactly how the tls tests work.

> 
>     
>   
> 
> Now, `xyz_fixture` can be RAII (running in a seastar::thread!!) which when constructed creates some data, and when destructed (automatically at the end of the test) destroys it.
> Moreover later (we don't need to do it now), xyz_fixture can become shared across all tests by adding some common code (e.g., inherit from some common class suite_fixture) which doesn't call the teardown() function at the end of the test and instead saves the object it in some global cache where it can be reused - and only frees stuff in this cache after all tests).

you are proposing a solution for adding suite-scoped fixture, right? this solution can be based on the change proposed by this PR.  just not really like the global part. and i wanted to support the test decorators, not sure how it plays in this design though..

anyhow, let's leave it for future.

nyh:
> > Imagine that you have fixture which is **a class** xyz_fixture, and the code does this:
> > ```
> > test_case_1(){
> >     xyz_fixture xyz;
> >    ... do something with xyz
> > }
> > 
> > test_case_1(){
> >     xyz_fixture xyz;
> >     ...do something with xyz
> > }
> > ```
> 
> this is exactly how the tls tests work.

So isn't this good? Instead of a macro, we are using a familiar C++ idiom (RAII) and it well-understood, readable, standard, etc. What's the downside of this approach?

> > Now, `xyz_fixture` can be RAII (running in a seastar::thread!!) which when constructed creates some data, and when destructed (automatically at the end of the test) destroys it.
> > Moreover later (we don't need to do it now), xyz_fixture can become shared across all tests by adding some common code (e.g., inherit from some common class suite_fixture) which doesn't call the teardown() function at the end of the test and instead saves the object it in some global cache where it can be reused - and only frees stuff in this cache after all tests).
> 
> you are proposing a solution for adding suite-scoped fixture, right? this solution can be based on the change proposed by this PR. just not really like the global part. and i wanted to support the test decorators, not sure how it plays in this design though..

Yes, I thought how my RAII-based suggestion could have the RAII object just a pointer (like a PIMPL singleton pattern) to a cached object. I wasn't saying we must do this right away, but I was saying there is a natural way to do this without changing the xyz_fixture call sights (which still look like RAII). I was asking if the same could be said to your macro-based "decorators". Maybe the answer is yes, but I just didn't understand them enough to tell.

> anyhow, let's leave it for future.

Right.

tchaikov:
> > > Imagine that you have fixture which is **a class** xyz_fixture, and the code does this:
> > > ```
> > > test_case_1(){
> > >     xyz_fixture xyz;
> > >    ... do something with xyz
> > > }
> > > 
> > > test_case_1(){
> > >     xyz_fixture xyz;
> > >     ...do something with xyz
> > > }
> > > ```
> > 
> > 
> > this is exactly how the tls tests work.
> 
> So isn't this good? Instead of a macro, we are using a familiar C++ idiom (RAII) and it well-understood, readable, standard, etc. What's the downside of this approach?

sorry for the confusion. i never claim that it's bad. actually, it's me who added the `https_server` thing to `tls_test.cc` .

> 
> > > Now, `xyz_fixture` can be RAII (running in a seastar::thread!!) which when constructed creates some data, and when destructed (automatically at the end of the test) destroys it.
> > > Moreover later (we don't need to do it now), xyz_fixture can become shared across all tests by adding some common code (e.g., inherit from some common class suite_fixture) which doesn't call the teardown() function at the end of the test and instead saves the object it in some global cache where it can be reused - and only frees stuff in this cache after all tests).
> > 
> > 
> > you are proposing a solution for adding suite-scoped fixture, right? this solution can be based on the change proposed by this PR. just not really like the global part. and i wanted to support the test decorators, not sure how it plays in this design though..
> 
> Yes, I thought how my RAII-based suggestion could have the RAII object just a pointer (like a PIMPL singleton pattern) to a cached object. I wasn't saying we must do this right away, but I was saying there is a natural way to do this without changing the xyz_fixture call sights (which still look like RAII). I was asking if the same could be said to your macro-based "decorators". Maybe the answer is yes, but I just didn't understand them enough to tell.
> 

probably you missed my point. i was reusing the design in Boost.Test which supports decorators. see https://www.boost.org/doc/libs/1_83_0/libs/test/doc/html/boost_test/tests_organization/decorators.html

> > anyhow, let's leave it for future.
> 
> Right.

Issue #1793, opened 2023-08-21
ss::chunked_fifo::iterator is not a regular iterator


andijcr:
https://github.com/scylladb/seastar/blob/f8cf9a1060b0971346d13c6f8e1eda071de6691d/include/seastar/core/chunked_fifo.hh#L162-L172

ss::chunked_fifo::iterator and const_iterator are not regular iterators, because they derive from a base class to gain operator++, but the return type is base_iterator. 

this makes some usually idiomatic code not compile.

example code in compiler-explorer: https://gcc.godbolt.org/z/ehW4WEv9h

the solution could be to use crtp to implement those classes

PR #1794, opened 2023-08-21
make chunked_fifo a regular range

andijcr:
this Pr modifies chunked_fifo::basic_iterator to make chunked_fifo a range,
and makes `static_assert(std::ranges::range<chunked_fifo<int>>);` compile.

logically there are two changes to accomplish this
1. make operator++ return the correct type by using CRTP
2. make basic_iterator default constructible, needed for the end iterator

The unit tests are static_assert on the specific issues

Fixes https://github.com/scylladb/seastar/issues/1793

### DIFF
--- a/include/seastar/core/chunked_fifo.hh
+++ b/include/seastar/core/chunked_fifo.hh
@@ -149,6 +149,7 @@ private:
         size_t _item_index = 0;
 
     protected:
+        inline basic_iterator() = default;
         inline explicit basic_iterator(chunk* c) noexcept;
         inline basic_iterator(chunk* c, size_t item_index) noexcept;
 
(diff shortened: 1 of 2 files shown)


tchaikov:
@andijcr hello Andrea, seems we were working in parallel on this issue. i like both your solution and mine. as your change are equipped with a test, but mine is a little bit simpler. probably you could include my implementation of iterator in your PR or i could have your permission on including your test?

andijcr:
@tchaikov 
go ahead, use my test. i just wanted to make std::views::filter compile :D

tchaikov:
> @tchaikov
> go ahead, use my test. i just wanted to make std::views::filter compile :D

molte grazie!

PR #1797, opened 2023-08-23
tls: Optionally skip client EOF wait

VladLazar:
This commit extends the `tls::wrap_client` interface with an optional param which indicates to the sessions that it should not wait for an EOF after sending the BYE message.

Badly behaved servers will not send the expected EOF message. Clients that call `wait_input_shutdown` on the connected socket will hang for 10 seconds in that case (the tls session `close` method spawns a background fiber with a 10s hardcoded timeout). I've tested this against the wacky server with both Seastar's experimental HTTP client and Redpanda's own HTTP client and the result was the same.

The previous default behaviour has been preserved and clients should always wait for EOF if the intend to re-use the connection.

### DIFF
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -309,6 +309,8 @@ namespace tls {
     struct tls_options {
         /// \brief whether to wait for EOF from server on session termination
         bool wait_for_eof_on_shutdown = true;
+        /// \brief server name to be used for the SNI TLS extension
+        sstring server_name = {};
     };
 
     /**
@@ -317,11 +319,28 @@ namespace tls {
      * Typically these should contain enough information
      * to validate the remote certificate (i.e. trust info).
      *
-     * \param name An optional expected server name for the remote end point
+     * ATTN: The method is going to be deprecated
+     *
+     * \param name The expected server name for the remote end point
      */
     /// @{
-    future<connected_socket> connect(shared_ptr<certificate_credentials>, socket_address, sstring name = {});
-    future<connected_socket> connect(shared_ptr<certificate_credentials>, socket_address, socket_address local, sstring name = {});
+    [[deprecated("Use overload with tls_options parameter")]]
+    future<connected_socket> connect(shared_ptr<certificate_credentials>, socket_address, sstring name);
+    [[deprecated("Use overload with tls_options parameter")]]
+    future<connected_socket> connect(shared_ptr<certificate_credentials>, socket_address, socket_address local, sstring name);
+    /// @}
+
+    /**
+     * Creates a TLS client connection using the default network stack and the
+     * supplied credentials.
+     * Typically these should contain enough information
+     * to validate the remote certificate (i.e. trust info).
+     *
+     * \param options Optional additional session configuration
+     */
+    /// @{
+    future<connected_socket> connect(shared_ptr<certificate_credentials>, socket_address, tls_options option = {});
+    future<connected_socket> connect(shared_ptr<certificate_credentials>, socket_address, socket_address local, tls_options options = {});
     /// @}
 
     /**
@@ -330,10 +349,38 @@ namespace tls {
      * Typically these should contain enough information
      * to validate the remote certificate (i.e. trust info).
      *
-     * \param name An optional expected server name for the remote end point
+     * ATTN: The method is going to be deprecated
+     *
+     * \param name The expected server name for the remote end point
      */
     /// @{
-    ::seastar::socket socket(shared_ptr<certificate_credentials>, sstring name = {});
+    [[deprecated("Use overload with tls_options parameter")]]
+    ::seastar::socket socket(shared_ptr<certificate_credentials>, sstring name);
+    /// @}
+
+    /**
+     * Creates a socket through which a TLS client connection can be created,
+     * using the default network stack and the supplied credentials.
+     * Typically these should contain enough information
+     * to validate the remote certificate (i.e. trust info).
+     *
+     * \param options Optional additional session configuration
+     */
+    /// @{
+    ::seastar::socket socket(shared_ptr<certificate_credentials>, tls_options options = {});
+    /// @}
+
+    /**
+     * Wraps an existing connection in SSL/TLS.
+     *
+     * ATTN: The method is going to be deprecated
+     *
+     * \param name The expected server name for the remote end point
+     */
+    /// @{
+    [[deprecated("Use overload with tls_options parameter")]]
+    future<connected_socket> wrap_client(shared_ptr<certificate_credentials>, connected_socket&&, sstring name);
+    future<connected_socket> wrap_server(shared_ptr<server_credentials>, connected_socket&&);
     /// @}
 
     /**
@@ -342,8 +389,7 @@ namespace tls {
      * \param options Optional additional session configuration
      */
     /// @{
-    future<connected_socket> wrap_client(shared_ptr<certificate_credentials>, connected_socket&&, sstring name = {}, tls_options options = {});
-    future<connected_socket> wrap_server(shared_ptr<server_credentials>, connected_socket&&);
+    future<connected_socket> wrap_client(shared_ptr<certificate_credentials>, connected_socket&&, tls_options options = {});
     /// @}
 
     /**
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -305,6 +305,12 @@ namespace tls {
         sstring _priority;
     };
 
+    /// TLS configuration options
+    struct tls_options {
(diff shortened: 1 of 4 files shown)


VladLazar:
cc. @elcallio as the original author

include/seastar/net/tls.hh
```diff
@@ -330,9 +330,14 @@ namespace tls {
...
-    /** Wraps an existing connection in SSL/TLS. */
+    /** 
+     * Wraps an existing connection in SSL/TLS.
+     *
+     * \param wait_for_eof Optionally indicate to the client if it should wait for an EOF from the server
+     * after sending the BYE message. By default, the wait is hardcoded to 10 seconds.
+     */
     /// @{
-    future<connected_socket> wrap_client(shared_ptr<certificate_credentials>, connected_socket&&, sstring name = {});
+    future<connected_socket> wrap_client(shared_ptr<certificate_credentials>, connected_socket&&, sstring name = {}, bool wait_for_eof = true);
```
elcallio, line 340: While perhaps overengineering, I think this should be something like:
```
struct tls_options {
   bool wait_for_eof_on_shutdown = true;
   // add as needed.
};

 future<connected_socket> wrap_client(shared_ptr<certificate_credentials>, connected_socket&&, sstring name = {}, std::optional<tls_options> = {});

```
and from there to initing a tls_options to this or defaults in session.

It would make it easier to extend with any further required options ahead without breaking API.

elcallio (changes requested):

include/seastar/net/tls.hh
VladLazar: Good shout. I kept the change super minimal to begin with as I was unsure what you'd think about it. Updating the PR.

VladLazar: Re-tested and force pushed the update.

include/seastar/net/tls.hh
```diff
@@ -305,6 +305,12 @@ namespace tls {
         sstring _priority;
     };
 
+    struct tls_options {
+        // Optionally indicate to the client if it should wait for an EOF from the server
+        // after sending the BYE message. By default, the wait is hardcoded to 10 seconds.
+        bool wait_for_eof_on_shutdown = true;
```
avikivity, line 311: Please use doxygen syntax (and verify it renders correctly locally).

include/seastar/net/tls.hh
```diff
@@ -330,9 +336,13 @@ namespace tls {
...
 
-    /** Wraps an existing connection in SSL/TLS. */
+    /**
+     * Wraps an existing connection in SSL/TLS.
+     *
+     * \param options Optional additional session configuration
+     */
     /// @{
-    future<connected_socket> wrap_client(shared_ptr<certificate_credentials>, connected_socket&&, sstring name = {});
+    future<connected_socket> wrap_client(shared_ptr<certificate_credentials>, connected_socket&&, sstring name = {}, std::optional<tls_options> options = std::nullopt);
```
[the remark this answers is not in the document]
avikivity: But why do we care? The caller sends options, either the default ones, or their own carefully crafted struct.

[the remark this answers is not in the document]
elcallio: No particular reason. I tend to think it more clear/easily eye-parsed when used. More or less just stylistic on my part. Feel free to override.

include/seastar/net/tls.hh
VladLazar: Done.

include/seastar/net/tls.hh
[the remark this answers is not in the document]
avikivity: I do want non-optional options, I think it's redunant. We use this style (without optional) everywhere.

VladLazar:
Looks like `test_100_continue` in `httdp_test.cc` [failed](https://app.circleci.com/pipelines/github/scylladb/seastar/2132/workflows/b28ab3d0-bdf1-428d-863f-bd25e57c4fd8/jobs/12337). I had a look through the previous runs and noticed the same failure occurred previously ([here](https://app.circleci.com/pipelines/github/scylladb/seastar/2090/workflows/4660a294-c781-47c0-b148-932a82a48fee/jobs/11826) & [here](https://app.circleci.com/pipelines/github/scylladb/seastar/2119/workflows/4f442e37-da0c-47a4-aa76-ea4862c5d5b6/jobs/12182)). Perhaps it's a touch flaky?

include/seastar/net/tls.hh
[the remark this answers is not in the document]
VladLazar: Sounds good. Will change to what you're suggesting.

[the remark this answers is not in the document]
VladLazar: Done.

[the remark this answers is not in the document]
VladLazar: `name` is the [Server Name Indication](https://www.cloudflare.com/en-gb/learning/ssl/what-is-sni/) which is a tls extension. I think it's recommended for clients to use it. To me moving `name` into the options would be a bit confusing since we'd need two overloads with different ways of specifying the SNI (one with a name param and another with options). Without an extra overload the change would break library users when they upgrade.

nyh:
> Looks like `test_100_continue` in `httdp_test.cc` [failed](https://app.circleci.com/pipelines/github/scylladb/seastar/2132/workflows/b28ab3d0-bdf1-428d-863f-bd25e57c4fd8/jobs/12337). I had a look through the previous runs and noticed the same failure occurred previously ([here](https://app.circleci.com/pipelines/github/scylladb/seastar/2090/workflows/4660a294-c781-47c0-b148-932a82a48fee/jobs/11826) & [here](https://app.circleci.com/pipelines/github/scylladb/seastar/2119/workflows/4f442e37-da0c-47a4-aa76-ea4862c5d5b6/jobs/12182)). Perhaps it's a touch flaky?

It does indeed appear flaky - see https://github.com/scylladb/seastar/issues/1645.

include/seastar/net/tls.hh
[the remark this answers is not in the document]
avikivity: The trick is to deprecate the old overload, then to remove it in a few months. Then we don't have some of the options as positional parameters, and some in option structs. See babac6de088ff8f4603fec0ca91baf3274225066 for an example.

include/seastar/net/tls.hh
```diff
@@ -305,6 +305,12 @@ namespace tls {
         sstring _priority;
     };
 
+    /// TLS configuration options
+    struct tls_options {
+        /// \brief whtether to wait for EOF from server on session termination
```
tchaikov, line 310: typo.

include/seastar/net/tls.hh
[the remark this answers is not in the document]
VladLazar: Makes sense. If we go down this route we'll have to change the other client APIs in the same way: `tls::connect` and `tls::socket` for the sake of consistency (new overloads and deprecate later). Not against doing it, but let's make sure we're in agreement first.

include/seastar/net/tls.hh
VladLazar: Thank you. Fixed

include/seastar/net/tls.hh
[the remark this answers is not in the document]
VladLazar: @elcallio Added a new commit which changes the API as discussed.

include/seastar/net/tls.hh
```diff
@@ -317,11 +319,28 @@ namespace tls {
...
+     * Creates a TLS client connection using the default network stack and the
+     * supplied credentials.
+     * Typically these should contain enough information
+     * to validate the remote certificate (i.e. trust info).
+     *
+     * \param options Optional additional session configuration
+     */
+    /// @{
+    future<connected_socket> connect(shared_ptr<certificate_credentials>, socket_address, tls_options option = {});
+    future<connected_socket> connect(shared_ptr<certificate_credentials>, socket_address, socket_address local, tls_options options = {});
```
avikivity, line 343: Sorry for not noticing earlier. Should we also move the "local" parameter to options?

I think not, "local" is too basic for tcp.

avikivity: No, it's basic for listening, for connecting it's an advanced option.

So maybe we should move it into the option struct too.

elcallio: Local is part of the "normal" socket API, so it should remain where it is for symmetry me thinks.

elcallio (approved):

include/seastar/net/tls.hh
VladLazar: I considered moving local too when making the change, but reached the same conclusion as @elcallio. "local" reads like a config for the socket and not for the tls layer.

avikivity: ok

PR #1799, opened 2023-08-28
fix compilation for musl

### LINKED ISSUE #1743: make backtrace optional
execinfo.h was remove from alpine linux here:
https://gitlab.alpinelinux.org/alpine/aports/-/commit/50795a14dee639ce2dcc836e2b2baca9bad4a1b1

In seastar commit https://github.com/scylladb/seastar/commit/a48fe69724492ffec64625a081f6dea54a50e470 changed from using libunwind to execinfo.h

alpine linux still has libunwind but the above link makes it sound like it would not give useful information with musl

a16bitsysop:
musl lacks execinfo.h
lfs64 definitions are now deprecated and will be removed in the next release of musl

Fixes #1743

### DIFF
--- a/include/seastar/util/backtrace.hh
+++ b/include/seastar/util/backtrace.hh
@@ -28,7 +28,10 @@
 #include <seastar/util/modules.hh>
 
 #ifndef SEASTAR_MODULE
+#if __has_include(<execinfo.h>)
 #include <execinfo.h>
+#define HAVE_EXECINFO
+#endif
 #include <iosfwd>
 #include <memory>
 #include <variant>
@@ -59,6 +62,7 @@ frame decorate(uintptr_t addr) noexcept;
 SEASTAR_MODULE_EXPORT
 template<typename Func>
 void backtrace(Func&& func) noexcept(noexcept(func(frame()))) {
+#ifdef HAVE_EXECINFO
     constexpr size_t max_backtrace = 100;
     void* buffer[max_backtrace];
     int n = ::backtrace(buffer, max_backtrace);
@@ -66,6 +70,10 @@ void backtrace(Func&& func) noexcept(noexcept(func(frame()))) {
         auto ip = reinterpret_cast<uintptr_t>(buffer[i]);
         func(decorate(ip - 1));
     }
+#else
+// Not implemented yet
+#define SEASTAR_BACKTRACE_UNIMPLEMENTED
+#endif
 }
 
 // Represents a call stack of a single thread.
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -382,9 +382,13 @@ posix_file_impl::list_directory(std::function<future<> (directory_entry de)> nex
     // instead.
 
     // From getdents(2):
+    // check for 64-bit inode number
+    static_assert(sizeof(ino_t) == 8, "large file support not enabled");
+    static_assert(sizeof(off_t) == 8, "large file support not enabled");
+
     struct linux_dirent64 {
-        ino64_t        d_ino;    /* 64-bit inode number */
-        off64_t        d_off;    /* 64-bit offset to next structure */
+        ino_t          d_ino;    /* 64-bit inode number */
+        off_t          d_off;    /* 64-bit offset to next structure */
         unsigned short d_reclen; /* Size of this dirent */
         unsigned char  d_type;   /* File type */
         char           d_name[]; /* Filename (null-terminated) */
--- a/src/util/backtrace.cc
+++ b/src/util/backtrace.cc
@@ -22,7 +22,6 @@
 module;
 #endif
 
-#include <execinfo.h>
 #include <link.h>
 #include <sys/types.h>
 #include <unistd.h>
--- a/tests/unit/exception_logging_test.cc
+++ b/tests/unit/exception_logging_test.cc
@@ -200,9 +200,11 @@ BOOST_AUTO_TEST_CASE(throw_with_backtrace_exception_logging) {
         log_msg << std::current_exception();
     }
 
+#ifndef SEASTAR_BACKTRACE_UNIMPLEMENTED
     auto regex_str = "backtraced<std::runtime_error> \\(throw_with_backtrace_exception_logging Backtrace:(\\s+(\\S+\\+)?0x[0-9a-f]+)+\\)";
     std::regex expected_msg_re(regex_str, std::regex_constants::ECMAScript | std::regex_constants::icase);
     BOOST_REQUIRE(std::regex_search(log_msg.str(), expected_msg_re));
+#endif
 }
 
 BOOST_AUTO_TEST_CASE(throw_with_backtrace_nested_exception_logging) {
@@ -217,9 +219,11 @@ BOOST_AUTO_TEST_CASE(throw_with_backtrace_nested_exception_logging) {
         }
     }
 
+#ifndef SEASTAR_BACKTRACE_UNIMPLEMENTED
     auto regex_str = "std::_Nested_exception<unknown_obj>.*backtraced<std::runtime_error> \\(outer Backtrace:(\\s+(\\S+\\+)?0x[0-9a-f]+)+\\)";
     std::regex expected_msg_re(regex_str, std::regex_constants::ECMAScript | std::regex_constants::icase);
     BOOST_REQUIRE(std::regex_search(log_msg.str(), expected_msg_re));
+#endif
 }
 
 BOOST_AUTO_TEST_CASE(throw_with_backtrace_seastar_nested_exception_logging) {
@@ -240,10 +244,12 @@ BOOST_AUTO_TEST_CASE(throw_with_backtrace_seastar_nested_exception_logging) {
         }
     }
 
+#ifndef SEASTAR_BACKTRACE_UNIMPLEMENTED
     auto regex_str = "seastar::nested_exception:.*backtraced<std::runtime_error> \\(inner Backtrace:(\\s+(\\S+\\+)?0x[0-9a-f]+)+\\)"
             " \\(while cleaning up after unknown_obj\\)";
     std::regex expected_msg_re(regex_str, std::regex_constants::ECMAScript | std::regex_constants::icase);
     BOOST_REQUIRE(std::regex_search(log_msg.str(), expected_msg_re));
+#endif
 }
 
 BOOST_AUTO_TEST_CASE(double_throw_with_backtrace_seastar_nested_exception_logging) {
@@ -264,8 +270,10 @@ BOOST_AUTO_TEST_CASE(double_throw_with_backtrace_seastar_nested_exception_loggin
         }
     }
 
+#ifndef SEASTAR_BACKTRACE_UNIMPLEMENTED
     auto regex_str = "seastar::nested_exception:.*backtraced<std::runtime_error> \\(inner Backtrace:(\\s+(\\S+\\+)?0x[0-9a-f]+)+\\)"
             " \\(while cleaning up after .*backtraced<std::runtime_error> \\(outer Backtrace:(\\s+(\\S+\\+)?0x[0-9a-f]+)+\\)\\)";
     std::regex expected_msg_re(regex_str, std::regex_constants::ECMAScript | std::regex_constants::icase);
     BOOST_REQUIRE(std::regex_search(log_msg.str(), expected_msg_re));
+#endif
 }
--- a/tests/unit/unwind_test.cc
+++ b/tests/unit/unwind_test.cc
@@ -50,12 +50,14 @@ BOOST_AUTO_TEST_CASE(test_signal_mask_is_preserved_on_unwinding) {
         // ignore
     }
 
-    // Check backtrace()
+    // Check backtrace() if execinfo.h is present
+#ifndef SEASTAR_BACKTRACE_UNIMPLEMENTED
     {
         size_t count = 0;
         backtrace([&count] (auto) { ++count; });
         BOOST_REQUIRE(count > 0);
     }
+#endif
 
     {
         sigset_t mask2;


a16bitsysop:
not sure if the config.hh creation is correct and also if `__USE_FILE_OFFSET64` is correct

a16bitsysop:
removed adding `ucontext` to `target_link_libraries` for `seastar` as fails to link on alpine/musl without it, but fails on glibc with it.  Guess it needs a check to decide if needed?

tchaikov (commented):
@a16bitsysop hi Duncan, much appreciated!

i'd suggest split it into smaller commits, each of which is accompanied with its own explanation. this would be non-trivial.  i can help on this. but i am not able to allocate large chunk of time on this. so if you are fine with this, we can divide and conquer this problem -- i can extract part of this into a separated PR. and leave the rest to you. and if you don't have enough bandwidth, i can go on with the rest. but if you do, that'd be great!

CMakeLists.txt
```diff
@@ -408,6 +408,10 @@ include (CheckIncludeFileCXX)
 check_include_file_cxx (execinfo.h SEASTAR_EXECINFO)
 check_include_file_cxx (sys/sdt.h SEASTAR_SDT_H)
 configure_file (config.hh.in ${CMAKE_CURRENT_SOURCE_DIR}/include/seastar/config.hh @ONLY)
+pkg_check_modules (UCONTEXT IMPORTED_TARGET libucontext)
+if (UCONTEXT_FOUND)
+  set (UCONTEXT_LIB PkgConfig::UCONTEXT)
+endif ()
```
tchaikov, line 414: i'd extract this into a separate `Finducontext.cmake`. it would resemble `Findrt.cmake`.

a16bitsysop:
> @a16bitsysop hi Duncan, much appreciated!
> 
> i'd suggest split it into smaller commits, each of which is accompanied with its own explanation. this would be non-trivial. i can help on this. but i am not able to allocate large chunk of time on this. so if you are fine with this, we can divide and conquer this problem -- i can extract part of this into a separated PR. and leave the rest to you. and if you don't have enough bandwidth, i can go on with the rest. but if you do, that'd be great!

Okay, will split them up.  I could create a separate PR from the original patches by @neopskose with the stderr stuff as that is alot of changes

CMakeLists.txt
a16bitsysop: okay will have a look

a16bitsysop: I will push with the changes in new PR removed (I rebased it onto this branch)

src/core/file.cc
```diff
@@ -383,8 +383,14 @@ posix_file_impl::list_directory(std::function<future<> (directory_entry de)> nex
...
     struct linux_dirent64 {
-        ino64_t        d_ino;    /* 64-bit inode number */
-        off64_t        d_off;    /* 64-bit offset to next structure */
+        /* 64-bit inode number */
+#ifdef __USE_FILE_OFFSET64
+        ino64_t        d_ino; /* 64-bit inode number */
+	off64_t        d_off; /* 64-bit offset to next structure */
+#else
+        ino_t          d_ino; /* 64-bit inode number */
+        off_t          d_off; /* 64-bit offset to next structure */
```
tchaikov, line 392: i think musl libc has not dropped the legacy types for large file support. see https://git.musl-libc.org/cgit/musl/tree/include/dirent.h#n59 . they are not provided by `_GNU_SOURCE` feature macro anymore, but under `_LARGEFILE64_SOURCE`. but yeah, they will be eventually removed. see https://musl.libc.org/releases.html for the release note for v1.2.4.

so, in the light of that `ino64_t` and `off64_t` are glibc extensions, and they are defined when `__USE_LARGEFILE64` is defined. and glibc also has type aliases of `ino_t` and `off_t` for these types if `_FILE_OFFSET_BITS` is 64. this macro is set by user.

`__USE_FILE_OFFSET64` is used by the glibc internally, it is defined when `_FILE_OFFSET_BITS` is 64.  if `__USE_FILE_OFFSET64` is defined, glibc defines 64bit APIs as default. which means the types like `ino_t` would be 64-bit, and `readdir()` would be 64-bit as well. one don't have to use the `readdir64()` variant explicitly.

so,

1. we should avoid using a macro used by glibc internally as a part of its public interface.
2. and the guarded statements are not quite relevant. as we should be able to use `ino_t` in the "64-bit" mode. as we want to use 64-bit inode number and offset.

as IMHO, the 64-bit inode number and 64-bit offset should be the default in this context as Seastar is designed to be used to build high-performance services, and they should be default under most circumstances. the existence their 32-bit cousins is only for the 32-bit systems -- only for embedded and some customized systems.

so, i think it'd be better to use `ino_t` and `off_t` directly. to achieve this goal, we can just put
```c++
static_assert(sizeof(ino_t) == 8, "large file support not enabled");
static_assert(sizeof(off_t) == 8, "large file support not enabled");
```

as by default, if we compile the tree with gcc or clang, `_GNU_SOURCE` is always defined.  and libstdc++ actually needs this, see https://gcc.gnu.org/onlinedocs/libstdc++/faq.html#faq.predefined. and glibc would define `_LARGEFILE64_SOURCE` if this macro is defined. and musl libc just use the 64-bit `ino_t` and `off_t`. so we should be safe in both cases without defining `_FILE_OFFSET_BITS=64` explicitly.

CMakeLists.txt
tchaikov: also, the introduction of `config.hh` is an important design decision. currently, we are using macros to persist various settings, and to pass them down to the Seastar applications. i think both of them have pros and cons. personally, i am in favor of `config.hh`. but i think we need more inputs on this regard. as this would be a tree-wide change if we want to be consistent on persisting / populating the configurations.

a16bitsysop: okay, I wasn’t sure how to pass the defines down into the source code from cmake

src/core/file.cc
a16bitsysop: okay will chenage

a16bitsysop:
updated and dropped `sys/sdt.h` check for the minute as that needs to check for systemtap, changed passing execinfo status to target_compile_definitions

ucontext check is copied from rt check, but it doesn’t work/run and I can’t see why

CMakeLists.txt
```diff
@@ -796,6 +800,7 @@ target_link_libraries (seastar
     StdAtomic::atomic
     lksctp-tools::lksctp-tools
     rt::rt
+    ucontext:ucontext
```
tchaikov, line 803: @a16bitsysop you might need to put `ucontext::ucontext` (two colons). but i created #1799, as i think there is no need to use a dedicated source file for checking the availability of the `getcontext()` function. could you please give it a try ?

tchaikov:
> updated and dropped `sys/sdt.h` check for the minute as that needs to check for systemtap, changed passing execinfo status to target_compile_definitions
> 

i created #1823 to address this. could you help review / test it ?

CMakeLists.txt
a16bitsysop: thanks, dropped that commit and tested yours

include/seastar/util/backtrace.hh
```diff
@@ -59,13 +61,16 @@ frame decorate(uintptr_t addr) noexcept;
 SEASTAR_MODULE_EXPORT
 template<typename Func>
 void backtrace(Func&& func) noexcept(noexcept(func(frame()))) {
+    #ifdef SEASTAR_HAVE_EXECINFO
```
tchaikov, line 64: we don't indent `#ifdef` in this project elsewhere. so probably we don't need to do so here?

include/seastar/util/backtrace.hh
```diff
@@ -59,13 +61,16 @@ frame decorate(uintptr_t addr) noexcept;
...
     constexpr size_t max_backtrace = 100;
     void* buffer[max_backtrace];
     int n = ::backtrace(buffer, max_backtrace);
     for (int i = 0; i < n; ++i) {
         auto ip = reinterpret_cast<uintptr_t>(buffer[i]);
         func(decorate(ip - 1));
     }
+    #endif
+    return;
```
tchaikov, line 73: this `return` is not necessary.

src/util/backtrace.cc
```diff
@@ -22,7 +22,9 @@
 module;
 #endif
 
+#ifdef SEASTAR_HAVE_EXECINFO
 #include <execinfo.h>
+#endif
```
tchaikov, line 27: i think we can drop this `#include`. as this source file does not reference the symbols exposed by `execinfo.h`

src/util/backtrace.cc
```diff
@@ -95,11 +97,13 @@ frame decorate(uintptr_t addr) noexcept {
 
 simple_backtrace current_backtrace_tasklocal() noexcept {
     simple_backtrace::vector_type v;
+    #ifdef SEASTAR_HAVE_EXECINFO
```
tchaikov, line 100: no need to `#ifdef` here, i think. as `backtrace()` would be a no-op if execinfo is not available.

CMakeLists.txt
```diff
@@ -870,6 +874,11 @@ if (Seastar_SSTRING)
     PUBLIC SEASTAR_SSTRING)
 endif ()
 
+if (SEASTAR_HAVE_EXECINFO)
```
tchaikov, line 877: i'd move the `include()` and `check_include_file_cxx ()` call closer to where `SEASTAR_HAVE_EXECINFO` is used.

CMakeLists.txt
```diff
@@ -870,6 +874,11 @@ if (Seastar_SSTRING)
...
+target_compile_definitions (seastar                                      
+	PUBLIC SEASTAR_HAVE_EXECINFO)
```
tchaikov, line 879: could you use two spaces for indent here?

CMakeLists.txt
tchaikov: also, we tend to use `Seastar_HAVE_EXECINFO`  as variable name in CMake.

CMakeLists.txt
```diff
@@ -405,6 +405,10 @@ find_package (Threads REQUIRED)
 find_package (PthreadSetName REQUIRED)
 find_package (Valgrind REQUIRED)
 
+include (CheckIncludeFileCXX)
+check_include_file_cxx (execinfo.h SEASTAR_HAVE_EXECINFO)
+#configure_file (config.hh.in ${CMAKE_CURRENT_SOURCE_DIR}/include/seastar/config.hh @ONLY)
```
tchaikov, line 410: might want to drop commented out code.

include/seastar/util/backtrace.hh
a16bitsysop: okay changed

include/seastar/util/backtrace.hh
a16bitsysop: yes works without

src/util/backtrace.cc
a16bitsysop: dropped it

src/util/backtrace.cc
a16bitsysop: changed

CMakeLists.txt
a16bitsysop: okay

CMakeLists.txt
a16bitsysop: changed

CMakeLists.txt
```diff
@@ -750,6 +753,11 @@ add_library (seastar
 
 add_library (Seastar::seastar ALIAS seastar)
 
+if (Seastar_HAVE_EXECINFO)
+target_compile_definitions (seastar
+  PUBLIC Seastar_HAVE_EXECINFO)
+endif()
```
tchaikov, line 759:
```suggestion
if (Seastar_HAVE_EXECINFO)
  target_compile_definitions (seastar
    PUBLIC Seastar_HAVE_EXECINFO)
endif()
```

CMakeLists.txt
tchaikov: > i'd move the `include()` and `check_include_file_cxx ()` call closer to where `SEASTAR_HAVE_EXECINFO` is used.

@a16bitsysop could you reconsider this suggestion?

CMakeLists.txt
```diff
@@ -754,9 +754,9 @@ add_library (seastar
 add_library (Seastar::seastar ALIAS seastar)
 
 if (Seastar_HAVE_EXECINFO)
-target_compile_definitions (seastar
-  PUBLIC Seastar_HAVE_EXECINFO)
-endif()
+  target_compile_definitions (seastar
+    PUBLIC Seastar_HAVE_EXECINFO)
+endif ()
```
tchaikov, line 759: could you please squash this change into the one which introduced `Seastar_HAVE_EXECINFO`?

CMakeLists.txt
a16bitsysop: I moved after seastar target is declared, it can’t go before as otherwise target doesn’t exist

CMakeLists.txt
a16bitsysop: okay, squashed

src/core/file.cc
```diff
@@ -382,9 +382,13 @@ posix_file_impl::list_directory(std::function<future<> (directory_entry de)> nex
     // instead.
 
     // From getdents(2):
+    /* check for 64-bit inode number */
```
tchaikov, line 385: nit, might want to put
```c++
// check for 64-bit inode number
```

tchaikov (approved):
lgtm.

tests/unit/unwind_test.cc
```diff
@@ -50,12 +50,14 @@ BOOST_AUTO_TEST_CASE(test_signal_mask_is_preserved_on_unwinding) {
         // ignore
     }
 
-    // Check backtrace()
+    // Check backtrace() unless execinfo.h is not present
```
tchaikov, line 53:
```suggestion
    // Check backtrace() unless execinfo.h is present
```

tests/unit/unwind_test.cc
```diff
@@ -50,7 +50,7 @@ BOOST_AUTO_TEST_CASE(test_signal_mask_is_preserved_on_unwinding) {
         // ignore
     }
 
-    // Check backtrace() unless execinfo.h is not present
+    // Check backtrace() if execinfo.h is present
```
tchaikov, line 53: could you please squash this change as well?

a16bitsysop: done it

tchaikov (approved):
lgtm

tchaikov:

include/seastar/util/backtrace.hh
```diff
@@ -28,7 +28,9 @@
 #include <seastar/util/modules.hh>
 
 #ifndef SEASTAR_MODULE
+#ifdef Seastar_HAVE_EXECINFO
```
nyh, line 31: This is so much cmake boilerplate code just to check if <execinfo.h> exists in nothing more. Since C++17, it's no longer needed - "__has_include" exists. Can we just use that and avoid all cmake changes?
since you need to know also below if execinfo exists, you can just do something like
```
#if __has_include(<execinfo.h>)
#include <execinfo.h>
#define HAVE_EXECINFO
#endif
```
and below check HAVE_EXECINFO.

include/seastar/util/backtrace.hh
```diff
@@ -59,13 +61,15 @@ frame decorate(uintptr_t addr) noexcept;
...
 void backtrace(Func&& func) noexcept(noexcept(func(frame()))) {
+#ifdef Seastar_HAVE_EXECINFO
     constexpr size_t max_backtrace = 100;
     void* buffer[max_backtrace];
     int n = ::backtrace(buffer, max_backtrace);
     for (int i = 0; i < n; ++i) {
         auto ip = reinterpret_cast<uintptr_t>(buffer[i]);
         func(decorate(ip - 1));
     }
+#endif
```
nyh, line 72: I wonder if we shouldn't in an `#else` return some sort of string saying backtrace() is unimplemented. Or at least leave a "FIXME" comment. One day we'll need to fix that, right?

nyh: By the way, a question about the preexisting code: Why is backtrace() defined in the header file? Do end users use it directly, or things implemented with it? Even if end-users do use it directly, it should switch to taking std::function (or seastar::noncopyable_function) instead of being a template - and then the implementation can be moded to the source file. It's not important for performance.

tests/unit/exception_logging_test.cc
```diff
@@ -200,9 +200,11 @@ BOOST_AUTO_TEST_CASE(throw_with_backtrace_exception_logging) {
         log_msg << std::current_exception();
     }
 
+#ifdef Seastar_HAVE_EXECINFO
```
nyh, line 203: In the long run, this ifdef is wrong: Eventually we'll want to implement a backtrace correctly even if HAVE_EXECINFO is false, using some other method.
One way you can solve this is to do in backtrace.hh something like:
```c++
#if __has_include(<execinfo.h>)
#include <execinfo.h>
#define HAVE_EXECINFO
#endif
...
backtrace() {
#ifdef HAVE_EXECINFO
... some real definition
#else
   // Not implemented yet
#define SEASTAR_BACKTRACE_UNIMPLEMENTED
#endif
```

Then in the test you can check for `#ifndef SEASTAR_BACKTRACE_UNIMPLEMENTED`. In the future, we can implement it, and remove this macro, even when execinfo.h doesn't exist.

nyh (commented):
Looks good, but I had some ideas on how to make the backtrace stuff simpler and more forward-looking to when we have backtrace() impelemented in any case. For your consideration.

tests/unit/exception_logging_test.cc
a16bitsysop: Yes that is simpler

nyh (approved):

PR #1800, opened 2023-08-28
Fix -Wunused-variable warnings

andijcr:
unused params in net::ethernet_address::adjust_endianess,
net::arp_protocol::forward and net::ip_protocol::forward trigger -Wunused-parameter 

This PR fixes this by either removing the name of the parameter or by void-casting

### DIFF
--- a/include/seastar/net/arp.hh
+++ b/include/seastar/net/arp.hh
@@ -47,7 +47,12 @@ public:
     arp_for_protocol(arp& a, uint16_t proto_num);
     virtual ~arp_for_protocol();
     virtual future<> received(packet p) = 0;
-    virtual bool forward(forward_hash& out_hash_data, packet& p, size_t off) { return false; }
+    virtual bool forward(forward_hash& out_hash_data, packet& p, size_t off) {
+      std::ignore = out_hash_data;
+      std::ignore = p;
+      std::ignore = off;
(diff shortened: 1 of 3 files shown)


include/seastar/net/arp.hh
```diff
@@ -47,7 +47,12 @@ public:
     arp_for_protocol(arp& a, uint16_t proto_num);
     virtual ~arp_for_protocol();
     virtual future<> received(packet p) = 0;
-    virtual bool forward(forward_hash& out_hash_data, packet& p, size_t off) { return false; }
+    virtual bool forward(forward_hash& out_hash_data, packet& p, size_t off) {
+      (void)out_hash_data;
```
tchaikov, line 51: instead of using the good-old C-style cast, i'd suggest use `std::ignore`. just for better readability, so developer is aware that this value is intentionally *ignored*.

andijcr: TIL

andijcr:
> and might want to change the title of this PR to something like "Fix -Wunused-variable warnings". as "Fix/warning/unused" is not quite parsable, IMHO.

done, forgot to change the default

tchaikov (approved):
lgtm

tchaikov:

PR #1802, opened 2023-08-29
Fix missing argument in tutorial

theidexisted:


doc/tutorial.md
```diff
@@ -451,7 +451,7 @@ make_dishes(coroutine::experimental::buffer_size_t max_dishes_on_table,
 
 seastar::future<> have_a_dinner(unsigned max_dishes_on_table) {
     Ingredients ingredients;
-    auto dishes = make_dishes(std::move(ingredients));
+    auto dishes = make_dishes(max_dishes_on_table, std::move(ingredients));
```
tchaikov, line 454: thank you for fixing the buggy example. but this does not compile. C++ does convert an integer or even the underlying type to a (scoped) enum automatically, despite the fact that one can initialize the enum with an integer without narrowing.

could you change it to somethling like
```c++
auto dishes = make_dishes(coroutine::experimental::buffer_size_t{max_dishes_on_table},
                          std::move(ingredients));
```

theidexisted: Thanks for the good catch. If it weren't for your reminder, I wouldn't have known that buffer_size_t is an `enum` type.

tchaikov:
@theidexisted thank you for addressing the comment. could you please squash these two commits into a single one? and update the title of the commit message to something like:

> tutorial: add a missing argument in code example

more specific this way, and so that one can tell that this change only touches the document when skimming through the commit logs.

doc/tutorial.md
tchaikov: yeah, i wanted to differentiate it from a regular integer. probably we can change it. but the idea is that it can be identified and consumed by the `promise_type` of the generator.

theidexisted: Thanks for the explanation.

tchaikov (approved):
lgtm!

tchaikov:

PR #1803, opened 2023-08-29
Fix compilation for non glibc platforms like alpine using musl libc

a16bitsysop:
Currently seastar fails to compile on non glibc platforms like ones using musl libc due to the use of some internal glibc macros and glibc extensions of functions, so change to portable ones or perform checks to make sure the correct one is used for the target platform.

These changes include:
Changing includes of sys/poll.h and sys/signal.h to the posix location without sys
Replacing the internal __mode_t with mode_t
Fix strerror_r usage which was a glibc specific extension
Use pthread_setaffinity_np instead of pthread_attr_setaffinity_np
Access sigevent.sigev_notify_thread_id with a macro

### DIFF
--- a/src/core/linux-aio.cc
+++ b/src/core/linux-aio.cc
@@ -33,6 +33,7 @@ module;
 #include <unistd.h>
 #include <sys/syscall.h>
 #include <valgrind/valgrind.h>
+#include <features.h>
 
 #ifdef SEASTAR_MODULE
 module seastar;
@@ -171,7 +172,11 @@ void setup_aio_context(size_t nr, linux_abi::aio_context_t* io_context) {
     auto r = io_setup(nr, io_context);
     if (r < 0) {
         char buf[1024];
-        char *msg = strerror_r(errno, buf, sizeof(buf));
+#ifdef __GLIBC__
+        const char *msg = strerror_r(errno, buf, sizeof(buf));
+#else
+        const char *msg = strerror_r(errno, buf, sizeof(buf)) ? "unknown error" : buf;
+#endif
         if (errno == EAGAIN) {
             auto aio_max_nr = read_first_line_as<unsigned>("/proc/sys/fs/aio-max-nr");
             throw std::runtime_error(
--- a/src/core/memory.cc
+++ b/src/core/memory.cc
@@ -69,6 +69,7 @@ module;
 #include <boost/container/static_vector.hpp>
 
 #include <dlfcn.h>
+#include <features.h>
 
 #ifndef SEASTAR_DEFAULT_ALLOCATOR
 #include <new>
@@ -1608,7 +1609,11 @@ configure(std::vector<resource::memory> m, bool mbind,
 
             if (r == -1) {
                 char err[1000] = {};
-                char *msg = strerror_r(errno, err, sizeof(err));
+#ifdef __GLIBC__
+                const char *msg = strerror_r(errno, err, sizeof(err));
+#else
+                const char *msg = strerror_r(errno, err, sizeof(err)) ? "unknown error" : buf;
+#endif
                 std::cerr << "WARNING: unable to mbind shard memory; performance may suffer: "
                         << msg << std::endl;
             }
--- a/src/core/posix.cc
+++ b/src/core/posix.cc
@@ -122,15 +122,15 @@ posix_thread::posix_thread(attr a, std::function<void ()> func)
     }
 #endif
 
-    if (a._affinity) {
-        auto& cpuset = *a._affinity;
-        pthread_attr_setaffinity_np(&pa, sizeof(cpuset), &cpuset);
-    }
-
     r = pthread_create(&_pthread, &pa,
                 &posix_thread::start_routine, _func.get());
     if (r) {
         throw std::system_error(r, std::system_category());
+    }
+
+    if (a._affinity) {
+        auto& cpuset = *a._affinity;
+        pthread_setaffinity_np(_pthread, sizeof(cpuset), &cpuset);
     }
 }
 
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -48,7 +48,7 @@ module;
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/eventfd.h>
-#include <sys/poll.h>
+#include <poll.h>
 #include <netinet/in.h>
 #include <boost/lexical_cast.hpp>
 #include <boost/thread/barrier.hpp>
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1950,7 +1950,7 @@ reactor::chmod(std::string_view name, file_permissions permissions) noexcept {
     });
 }
 
-directory_entry_type stat_to_entry_type(__mode_t type) {
+directory_entry_type stat_to_entry_type(mode_t type) {
     if (S_ISDIR(type)) {
         return directory_entry_type::directory;
     }
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1176,7 +1176,10 @@ cpu_stall_detector_posix_timer::cpu_stall_detector_posix_timer(cpu_stall_detecto
     struct sigevent sev = {};
     sev.sigev_notify = SIGEV_THREAD_ID;
     sev.sigev_signo = signal_number();
-    sev._sigev_un._tid = syscall(SYS_gettid);
+#ifndef sigev_notify_thread_id
+#define sigev_notify_thread_id _sigev_un._tid
+#endif
+    sev.sigev_notify_thread_id = syscall(SYS_gettid);
     int err = timer_create(CLOCK_THREAD_CPUTIME_ID, &sev, &_timer);
     if (err) {
         throw std::system_error(std::error_code(err, std::system_category()));
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -32,7 +32,7 @@ module;
 #include <fcntl.h>
 #include <signal.h>
 #include <sys/epoll.h>
-#include <sys/poll.h>
+#include <poll.h>
 #include <sys/syscall.h>
 #include <sys/resource.h>
 #include <boost/container/small_vector.hpp>
--- a/tests/unit/thread_test.cc
+++ b/tests/unit/thread_test.cc
@@ -30,7 +30,7 @@
 #include <seastar/core/loop.hh>
 #include <seastar/core/sleep.hh>
 #include <sys/mman.h>
-#include <sys/signal.h>
+#include <signal.h>
 
 #include <valgrind/valgrind.h>
 


src/core/linux-aio.cc
```diff
@@ -171,9 +171,9 @@ void setup_aio_context(size_t nr, linux_abi::aio_context_t* io_context) {
     auto r = io_setup(nr, io_context);
     if (r < 0) {
         char buf[1024];
-        char *msg = strerror_r(errno, buf, sizeof(buf));
+        strerror_r(errno, buf, sizeof(buf));
```
tchaikov, line 174: this could be wrong. as per the manapge

>        The  GNU-specific  strerror_r()  returns  a pointer to a string containing the error message.  This may be either a pointer to a string that the function stores in buf, or a pointer to some (immutable) static string (in which
>        case buf is unused).  If the function stores a string in buf, then at most buflen bytes are stored (the string may be truncated if buflen is too small and errnum is unknown).  The string always  includes  a  terminating  null
>        byte ('\0').

actually, glibc does not use the `buf` passed to it if `_GNU_SOURCE` is defined. see also https://godbolt.org/z/nv8qzMK8K

probably we can check for the existence of glibc's extension by checking the `_GNU_SOURCE` macro
   ```c++
   char buf[1024] = "unknown error";
   #ifdef _GNU_SOURCE
       char *msg = strerror_r(error, buf, 0);
   #else
       char *msg = strerror_r(error, buf, sizeof(buf));
   #endif
   ```

also, i'd suggest put more details in the commit message to explain that what we are using now is a glibc extension of `strerror_r()`, which does not exists in, for instance, musl libc.

include/seastar/util/log.hh
```diff
@@ -504,8 +504,8 @@ enum class logger_timestamp_style {
 /// \brief Output stream to use for logging.
 enum class logger_ostream_type {
     none,
-    stdout,
-    stderr,
+    std_out,
+    std_err,
```
tchaikov, line 508: could you please fill more details on why we should replace `stdout` with `std_out` ?

> fix stdout stderr usage for musl 

does not provide enough context for the reviewer to tell if the fix is the right one, or even it address the failure. and if it does, what the failure is ?

tchaikov (commented):
i am still reviewing. some of the commits fail to include explanations in the commit messages. would be great if you could fill more details there.

include/seastar/util/log.hh
a16bitsysop: The problem is stdout/stderr/stdin are permitted to be macros, so musl defines them as such.

 Will add that to the description, I can get the full error when it builds with fmt 10.1.1

src/core/linux-aio.cc
a16bitsysop: okay thanks, will change it

include/seastar/util/log.hh
tchaikov: the FTBFS w/ fmt v10.1.1 should be fixed by #1808

src/core/posix.cc
```diff
@@ -122,16 +122,16 @@ posix_thread::posix_thread(attr a, std::function<void ()> func)
...
-        auto& cpuset = *a._affinity;
-        pthread_attr_setaffinity_np(&pa, sizeof(cpuset), &cpuset);
-    }
-
     r = pthread_create(&_pthread, &pa,
                 &posix_thread::start_routine, _func.get());
     if (r) {
         throw std::system_error(r, std::system_category());
     }
+
```
tchaikov, line 130: might want to explain why we need to move this `if` clause down in the commit message. also, we have a tradition of adding prefix in the title of commit message, so the title should look like:

> posix: ...

and the title is not accurate, this commit does not move a function.

src/core/reactor.cc
```diff
@@ -1946,7 +1946,7 @@ reactor::chmod(std::string_view name, file_permissions permissions) noexcept {
     });
 }
 
-directory_entry_type stat_to_entry_type(__mode_t type) {
+directory_entry_type stat_to_entry_type(mode_t type) {
```
tchaikov, line 1952: might want to add a prefix like

> reactor: replace __mode_t with mode_t

and explain the rationales in the body of commit message

> as __mode_t is glib specific, while mode_t is specified by POSIX, see https://man7.org/linux/man-pages/man3/mode_t.3type.html

src/core/reactor.cc
```diff
@@ -1172,7 +1172,10 @@ cpu_stall_detector_posix_timer::cpu_stall_detector_posix_timer(cpu_stall_detecto
     struct sigevent sev = {};
     sev.sigev_notify = SIGEV_THREAD_ID;
     sev.sigev_signo = signal_number();
-    sev._sigev_un._tid = syscall(SYS_gettid);
+#ifndef sigev_notify_thread_id
+#define sigev_notify_thread_id _sigev_un._tid
+#endif
```
tchaikov, line 1177: a little bit background in the commit message would be great, something like

> _sigev_un._tid is an internal glibc field and is not available on
musl libc. 
> see https://sourceware.org/bugzilla/show_bug.cgi?id=27417

tchaikov:
i left a few comment. the title of this PR can be improved. currently, it reads

> split parts of #1799 into separate commits

but without the context, it might not be very helpful. so probably could rewrite it to something like:

> fix build failures with musl libc

include/seastar/util/log.hh
a16bitsysop: Thanks, it fixes on alpine apart from for aarch64 which now has a test failure:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1099806

/builds/a16bitsysop/aports/testing/seastar/src/seastar-1ddb6fe061aea521aeeafe3ad6ab926b803ab51d/tests/unit/stall_detector_test.cc(107): fatal error: in "simple_stalls": critical check reports == 5 has failed [4 != 5]

src/core/posix.cc
a16bitsysop: I am not sure why, @neopsykose wrote it.  I will build without that patch and see what happens.

a16bitsysop:
Thanks for the review, will add more info into the commits and PR title

include/seastar/util/log.hh
a16bitsysop: the error is like:
```
include/seastar/util/log.hh:518:5: error: 'logger_ostream_type' does not name a type
  518 |     logger_ostream_type logger_ostream = logger_ostream_type::stderr;
```
Full pipeline:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1100306

tchaikov: @a16bitsysop , thank you Duncun! so the problem is that musl libc actually implement `stderr` and `stdout` as C macros like https://git.musl-libc.org/cgit/musl/tree/include/stdio.h#n66. so we cannot use them as symbol names anymore. furthermore, this is allowed by the C++ standard. see https://eel.is/c%2B%2Bdraft/cstdio.syn .

the change per-se looks good to me. actually, i'd prefer using names like `cout` and `cerr`. but i have a bigger concern: Seastar is a library, so it is important to us to keep backward compatibility at some degree. `logger_ostream_type` is part of the public interface. so what if users are already using `logger_ostream_type::stdout` in their application? one might argue their applications do not compile with musl libc before and now they can. but what if they only have glibc and do not plan to switch over to musl libc? their applications are compiling with glibc, but after this change, they don't compile anymore.

so, i'd suggest deprecate `stderr` and `stdout` like
```c++
    stdout __attribute__ ((deprecated ("use cout instead"))) = 1,
    stderr __attribute__ ((deprecated ("use cerr instead"))) = 2,
    cout = 1,
    cerr = 2,
}
```

as `[[deprecated]]` is now allowed to be applied to a member of enum so i am using the compiler extension here. fortunately, both GCC and Clang support this.

if you agree with me, i'd suggest add more context and rationales in the corresponding commit message.

a16bitsysop: Yes that seems fine to me

src/core/posix.cc
a16bitsysop: I tried without that patch and the error is:

```
src/core/posix.cc:127:9: error: 'pthread_attr_setaffinity_np' was not declared in this scope; did you mean 'pthread_setaffinity_np'?
  127 |         pthread_attr_setaffinity_np(&pa, sizeof(cpuset), &cpuset);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      |         pthread_setaffinity_np
```

a16bitsysop: I don’t think that one is musl specific?

src/core/linux-aio.cc
a16bitsysop: Tried that and it looks like alpine has _GNU_SOURCE defined somewhere or maybe in boost or another header?

```
src/seastar-1ddb6fe061aea521aeeafe3ad6ab926b803ab51d/src/core/linux-aio.cc:175:31: error: invalid conversion from 'int' to 'char*' [-fpermissive]
  175 |         char *msg = strerror_r(errno, buf, 0);
      |                     ~~~~~~~~~~^~~~~~~~~~~~~~~
      |                               |
      |                               int
```

Also the musl version just returns and `int` https://git.musl-libc.org/cgit/musl/tree/src/string/strerror_r.c

This works but needs `features.h` to get `__GLIBC__`
```
#ifdef __GLIBC__
        char *msg = strerror_r(errno, buf, sizeof(buf));
 #else
        char *msg = buf;
       strerror_r(errno, buf, sizeof(buf));
 #endif
```

Another solution is to implement with `snprintf` http://www.club.cc.cmu.edu/~cmccabe/blog_strerror.html

tchaikov: > Tried that and it looks like alpine has _GNU_SOURCE defined somewhere or maybe in boost or another header?
> 
> ```
> src/seastar-1ddb6fe061aea521aeeafe3ad6ab926b803ab51d/src/core/linux-aio.cc:175:31: error: invalid conversion from 'int' to 'char*' [-fpermissive]
>   175 |         char *msg = strerror_r(errno, buf, 0);
>       |                     ~~~~~~~~~~^~~~~~~~~~~~~~~
>       |                               |
>       |                               int
> ```
> 
> Also the musl version just returns and `int` https://git.musl-libc.org/cgit/musl/tree/src/string/strerror_r.c
> 
> This works but needs `features.h` to get `__GLIBC__`

how about: 
```c++
char buf[1024];
#if (_POSIX_C_SOURCE >= 200112L) && !  _GNU_SOURCE
    const char *msg = strerror_r(error, buf, sizeof(buf)) ? "unknown error" : buf;
#else
    const char * msg = strerror_r(error, buf, 0);
#endif

File: src/core/posix.cc (hunk 3 above)
Comment (tchaikov) [reply src/core/posix.cc:130]: strictly speaking, none of the changes in this PR is. they are the consequence of the differences of glibc and non-glibc, which in this case, musl libc implementations.

in this very case, as the name of `pthread_attr_setaffinity_np` suggests, this API is not portable, but a GNU extension. without which, we cannot set the affinity when creating a pthread.

but without another extension `pthread_setaffinity_np`, we can set the CPU affinity **after** the thread is created. hence this change. since both glibc and musl libc implements this non-portable extension, we can use it here as an alternative of `pthread_attr_setaffinity_np`.

if you are with me. could you please put more context and the rationales in the corresponding commit message?

File: src/core/posix.cc [hunk 6]
```diff
@@ -122,16 +122,16 @@ posix_thread::posix_thread(attr a, std::function<void ()> func)
...
+    if (a._affinity) {
+        auto& cpuset = *a._affinity;
+        pthread_setaffinity_np(_pthread, sizeof(cpuset), &cpuset);
+    }
```
tchaikov, line 134: note to myself, we can have a minor refactor here. like
```c++
if (auto* cpuset = a._affinity; cpuset) {
    pthread_setaffinity_np(_pthread, sizeof(*cpuset), cpuset);
}
```

src/core/linux-aio.cc
a16bitsysop: the musl/posix version returns an int https://linux.die.net/man/3/strerror_r

include/seastar/util/log.hh
a16bitsysop: this gives me the error:
```
/include/seastar/util/log.hh:505:32: not
e: to match this '{'                                                                                                               
  505 | enum class logger_ostream_type {                                                                                           
      |                                ^                                                                                           /home/dunk/gits/aports/testing/seastar/src/seastar-1ddb6fe061aea521aeeafe3ad6ab926b803ab51d/include/seastar/util/log.hh:507:64: err
or: cannot convert 'int' to 'seastar::logger_ostream_type' in initialization                                                       
  507 |     stdout __attribute__ ((deprecated ("use cout instead"))) = 1,                                                          
      |                                                                ^                                                           
      |                                                                |                                                                 |                                                                int                                                         /home/dunk/gits/aports/testing/seastar/src/seastar-1ddb6fe061aea521aeeafe3ad6ab926b803ab51d/include/seastar/util/log.hh:508:64: err
or: cannot convert 'int' to 'seastar::logger_ostream_type' in initialization                                                       
  508 |     stderr __attribute__ ((deprecated ("use cerr instead"))) = 2,                                                          
      |                                                                ^                                                           
      |                                                                |                                                                 |                                                                int                                                         
/home/dunk/gits/aports/testing/seastar/src/seastar-1ddb6fe061aea521aeeafe3ad6ab926b803ab51d/include/seastar/util/log.hh:509:12: err
or: cannot convert 'int' to 'seastar::logger_ostream_type' in initialization                                                       
  509 |     cout = 1,                                                                                                              
      |            ^                                                                                                               
      |            |                                                                                                               
      |            int                                                                                                             
/home/dunk/gits/aports/testing/seastar/src/seastar-1ddb6fe061aea521aeeafe3ad6ab926b803ab51d/include/seastar/util/log.hh:510:12: err
or: cannot convert 'int' to 'seastar::logger_ostream_type' in initialization
  510 |     cerr = 2,           
      |            ^            
      |            |
      |            int
```

a16bitsysop:
Added more info now, the deprected enum didn’t work for me when assigning a number.
posix strerror_r returns an int

a16bitsysop:
i seemed to have accidentally included some header file changes, I am not sure if these are useful or not.  They give warnings on musl, but are redirected to other location.

src/core/linux-aio.cc
tchaikov: > the musl/posix version returns an int https://linux.die.net/man/3/strerror_r

i understand. and the first branch does check the `int` return value. am i missing anything?

a16bitsysop: yes, you suggestion has both cases with `const char *msg = strerror_r`

which fails on musl as it returns an int

tchaikov: @a16bitsysop please reread the statement:
```c++
const char *msg = strerror_r(error, buf, sizeof(buf)) ? "unknown error" : buf;
```
which checks the return value of `strerror_r()`, if it returns non-zero, "unknown error" is returned, `buf` otherwise.

a16bitsysop: ah, I see.  The `_GNU_SOURCE` check wasn’t working as its already set by one of the dependency headers I think.

nyh:
I haven't reviewed the actual patches yet, but a request about the commit message: 
Please write a cover letter (the first comment in this PR) which describes what this PR is about, and also update the name of the PR - it's not a generic "non-glibc platforms" - if understand correctly, this PR is about compilation with Musl, so the title of the PR and its description should mention it, and the description should summarize what kinds of problems you encountered and will be solved by the following patches.
Thanks!

include/seastar/util/log.hh
a16bitsysop: thanks, it failed.  Will add link in new PR

a16bitsysop:
okay will do a cover letter description

src/core/linux-aio.cc
tchaikov: > ah, I see. The `_GNU_SOURCE` check wasn’t working as its already set by one of the dependency headers I think.

hmm, i tested locally last week end. will retry later this week.

src/core/posix.cc
```diff
@@ -122,16 +122,16 @@ posix_thread::posix_thread(attr a, std::function<void ()> func)
     }
 #endif
 
-    if (a._affinity) {
```
tchaikov, line 125: >  core/posix.cc: pthread_attr_setaffinity_np is glibc specific and not portable

this reads like a part of bug report. so i think the title of this commit message can be improve further. ideally, we should use the imperative mood in the title. in this case, we could use something like:

> posix: use pthread_setaffinity_np instead of pthread_attr_setaffinity_np
>
> pthread_attr_setaffinity_np is glibc specific and not portable. without which, 
> we cannot set the affinity when creating a pthread.but without another
> extension pthread_setaffinity_np, we can set the CPU affinity after the thread
> is created. hence this change. since both glibc and musl libc implements this
> non-portable extension, we can use it here as an alternative of 
> pthread_attr_setaffinity_np.

please note,  the original title is now included as part of the commit message.

src/core/posix.cc
```diff
@@ -122,16 +122,16 @@ posix_thread::posix_thread(attr a, std::function<void ()> func)
...
-
     r = pthread_create(&_pthread, &pa,
                 &posix_thread::start_routine, _func.get());
     if (r) {
         throw std::system_error(r, std::system_category());
     }
+
+    if (a._affinity) {
+        auto& cpuset = *a._affinity;
+        pthread_setaffinity_np(_pthread, sizeof(cpuset), &cpuset);
```
nyh, line 133: The patch looks good:
One (like me) might wonder if creating a new thread with the old affinity and only then changing the affinity doesn't have some small performance implications - like the thread being created on one CPU and then having to move. But in reality, the Linux clone() system call does **not** allow the user to control the new thread's cpu affinity, and one needs to first clone() the thread and then call sched_setaffinity. This is also what the glibc implementation does (I took a look) for pthread_attr_setaffinity_np - it just remembers it, and applies it after creating the new thread with clone(). So the fact you're doing this explicitly is perfectly fine.

src/core/reactor.cc
```diff
@@ -1172,7 +1172,10 @@ cpu_stall_detector_posix_timer::cpu_stall_detector_posix_timer(cpu_stall_detecto
     struct sigevent sev = {};
     sev.sigev_notify = SIGEV_THREAD_ID;
     sev.sigev_signo = signal_number();
-    sev._sigev_un._tid = syscall(SYS_gettid);
+#ifndef sigev_notify_thread_id
```
nyh, line 1175: You forgot to mention in the commit message or a comment that what Musl (and FreeBSD, and the Linux kernel) have a sigev_notify_thread_id macro. Only glibc is missing it (https://sourceware.org/bugzilla/show_bug.cgi?id=27417) so we used that internal thing.

include/seastar/util/log.hh
```diff
@@ -504,8 +504,8 @@ enum class logger_timestamp_style {
 /// \brief Output stream to use for logging.
 enum class logger_ostream_type {
     none,
-    stdout,
-    stderr,
+    cout,
+    cerr,
```
tchaikov, line 508: might want to drop this commit and rebase onto the master HEAD.

src/core/reactor.cc
tchaikov, line 1175: could you please use the imperative mood in the title, and put the explanations in the body of the commit message? the title could look like :

> reactor: access sigevent.sigev_notify_thread_id with a macro

and might want to note that the name macro of `sigev_notify_thread_id` comes from musl libc, which defines a macro with the same name for accessing the corresponding field in `sigevent`.

src/core/linux-aio.cc
```diff
@@ -170,8 +171,12 @@ int io_pgetevents(aio_context_t io_context, long min_nr, long nr, io_event* even
 void setup_aio_context(size_t nr, linux_abi::aio_context_t* io_context) {
     auto r = io_setup(nr, io_context);
     if (r < 0) {
-        char buf[1024];
-        char *msg = strerror_r(errno, buf, sizeof(buf));
+	char buf[1024];
+#ifdef __GLIBC__
+        const char *msg = strerror_r(errno, buf, 0);
```
nyh, line 176: Why did you stop using sizeof(buf) and changed it to 0?
I think that if errno has some unrecognized number, the buffer *will* be used (to contain a string like "unknown error 1234") so better safe than sorry, do pass the correct sizeof(buf) and not 0.
(as an aside, if you only needed 1024 bytes for the Posix variant, please move it to the else).

nyh (changes requested):
Looks mostly good, I left just some minor comments. Please take out the patch that I think already went in by @tchaikov.

src/core/linux-aio.cc
tchaikov: > Why did you stop using sizeof(buf) and changed it to 0? I think that if errno has some unrecognized number, the buffer _will_ be used (to contain a string like "unknown error 1234") so better safe than sorry,
> do pass the correct sizeof(buf) and not 0. (as an aside, if you only needed 1024 bytes for the Posix variant, please move it to the else).

i think it was my idea. after a second thought and cross-checking the document by reading the [glibc code](https://github.com/bminor/glibc/blob/6333a6014f32c7ced36ced610d5a5cecbafba6c3/string/_strerror.c#L24-L35), i agree with you. we should respect glibc's manpage.

> his may be either a pointer to a string that the function stores in buf, or a pointer to some (immutable) static string  (in which case buf is unused).  If the function stores a string in buf, then at most buflen bytes are stored (the string may be truncated if buflen is too small and errnum is unknown).  The string always  includes  a  terminating  null         byte ('\0').

tchaikov: note to myself, 128 bytes would be more than enough. it's not that important though.

src/core/posix.cc
a16bitsysop: okay will update

include/seastar/util/log.hh
a16bitsysop: yes will drop and rebase

src/core/reactor.cc
a16bitsysop: okay will modify description

src/core/posix.cc
a16bitsysop: okay

src/core/reactor.cc
a16bitsysop: okay, will add that

src/core/linux-aio.cc
a16bitsysop: okay will change

a16bitsysop:
Is that okay now?

src/core/linux-aio.cc
```diff
@@ -170,8 +171,12 @@ int io_pgetevents(aio_context_t io_context, long min_nr, long nr, io_event* even
 void setup_aio_context(size_t nr, linux_abi::aio_context_t* io_context) {
     auto r = io_setup(nr, io_context);
     if (r < 0) {
-        char buf[1024];
-        char *msg = strerror_r(errno, buf, sizeof(buf));
+	char buf[1024];
```
tchaikov, line 174: please replace <tab> with 4 spaces. see https://github.com/scylladb/seastar/blob/master/coding-style.md#whitespace

src/core/reactor.cc
tchaikov: this comment is not addressed yet. could you please reconsider it?

src/core/posix.cc
tchaikov: this comment is not addressed yet. could you please reconsider it?

src/core/linux-aio.cc
tchaikov: okay. just tested. my approach indeed does work. =(

src/core/posix.cc
a16bitsysop: should be done now

src/core/linux-aio.cc
a16bitsysop: done it

a16bitsysop:
missed that, changed it now.On 8 Sep 2023, at 03:34, Kefu Chai ***@***.***> wrote:﻿
@a16bitsysop hi Duncun, we are close. just one thing: could you please take a look at #1803 (comment) ?

src/core/memory.cc
```diff
@@ -1608,7 +1609,11 @@ configure(std::vector<resource::memory> m, bool mbind,
 
             if (r == -1) {
                 char err[1000] = {};
-                char *msg = strerror_r(errno, err, sizeof(err));
+#ifdef __GLIBC__
+                const char *msg = strerror_r(errno, err, sizeof(buf));
```
tchaikov, line 1613: copy-n-pasta.

a16bitsysop: yes just noticed as well

tchaikov:
> I haven't reviewed the actual patches yet, but a request about the commit message:
> Please write a cover letter (the first comment in this PR) which describes what this PR is about, and also update the name of the PR - it's not a generic "non-glibc platforms" - if understand correctly, this PR is about compilation with Musl, so the title of the PR and its description should mention it, and the description should summarize what kinds of problems you encountered and will be solved by the following patches.
> Thanks!

@a16bitsysop i am quoting @nyh 's comment above. personally, i think the changes are not specific to musl libc. we are more like doing something

```c
#if GLIBC
use_glibc_extension();
#else
use_posix_or_some_other_standard_alternative(); // or well, yet another extension
#endif
```

so, it might be helpful to at least mention musl libc in the cover letter. this would help posterity to understand the original intention of the changes introduced by the merge commit. please note, we use the cover letter in the merge commit, as this changeset includes multiple commits.

a16bitsysop:
okay, will change

tchaikov (approved):
thanks! lgtm.

tchaikov:
@nyh hi Nadav, could you please take another look? i think your concerns have been addressed in the latest revision.

tchaikov:

tchaikov:
@nyh hi Nadav, could you please take another look? i think your concerns have been addressed in the latest revision.

nyh:
> @a16bitsysop i am quoting @nyh 's comment above. personally, i think the changes are not specific to musl libc. as we are more like doing something

My opinion is different, but this discussion over the commit message text isn't important enough to delay the PR over.
In my opinion, this PR contains a specific combination of changes needed to support **Musl**, specifically. it is unlikely that any other "non glibc" library will get supported by this specific change - it won't start working on Solaris or FreeBSD or Windows thanks to this PR, and its author didn't check any of those other "non glibc" libraries.

I agree that each individual patch should mention the reason for each individual change, and this can be, indeed, that something is a "non-standard glibc" extension. Even then, we should mention the immediate motivation - which other library (i.e., Musl) doesn't support this extension. After all, the PR author - or you - did not volunteer to embark on an all-out effort to eradicate from Seastar anything not supported by some antique Posix standard, the latest version of it was written 7 years ago.  The PR author took on himself the task of finding why Seastar doesn't build on Musl, so the commit message should be clear on something.

> ```
> #if GLIBC
> use_glibc_extension();
>````

In my opinion, we shouldn't check "if glibc" unless we don't have another option. It is very likely that other non-glibc "clones" will sooner or later want to support everything supported by the popular glibc, so it's better to detect the existence of an individual feature - not the entire library. This is what autoconf, and now cmake, was supposed to be doing (whenever other in-language alternatives like __has_include() aren't enough).

nyh (approved):

tchaikov:
> In my opinion, we shouldn't check "if glibc" unless we don't have another option. It is very likely that other non-glibc "clones" will sooner or later want to support everything supported by the popular glibc, so it's better to detect the existence of an individual feature - not the entire library. This is what autoconf, and now cmake, was supposed to be doing (whenever other in-language alternatives like __has_include() aren't enough).

agreed. lemme create a patch to drop this kind of check.

tchaikov:
> > In my opinion, we shouldn't check "if glibc" unless we don't have another option. It is very likely that other non-glibc "clones" will sooner or later want to support everything supported by the popular glibc, so it's better to detect the existence of an individual feature - not the entire library. This is what autoconf, and now cmake, was supposed to be doing (whenever other in-language alternatives like __has_include() aren't enough).
> 
> agreed. lemme create a patch to drop this kind of check.

created #1852

PR #1805, opened 2023-08-30
coroutine: Add a necessary include in seastar/core/generator.hh

dawmd:
The types used in the header file rely on types from \<coroutine>, but that header does not seem to be included.

### DIFF
--- a/include/seastar/coroutine/generator.hh
+++ b/include/seastar/coroutine/generator.hh
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <cassert>
+#include <coroutine>
 #include <optional>
 #include <utility>
 #include <seastar/core/future.hh>


include/seastar/coroutine/generator.hh
```diff
@@ -26,6 +26,13 @@
 #include <utility>
 #include <seastar/core/future.hh>
 
+
+#ifndef SEASTAR_COROUTINES_ENABLED
+#error Coroutines support disabled.
+#else
+#include <coroutine>
+#endif
```
tchaikov, line 34: i'd suggest just
```c++
#include <coroutine>
```

as the error message would be self-explained.

tchaikov: or, if we want to mirror the way how `coroutine.hh` checks for the availability using `std-compat.hh`, could put
```c++
#include <seastar/util/std-compat.hh>
#ifndef SEASTAR_COROUTINES_ENABLED
#error Coroutines support disabled.
#endif
#include <coroutine>
```
as the macro is offered by `seastar/util/std-compat.hh`, we should not rely on ``seastar/core/future.h` to do this for us. but ... personally, i still prefer `#include <coroutine>` directly. as we don't have "error handling" here.

dawmd: Updated

tchaikov (changes requested):
please squash the two commits into a single one. the second commit message does not really help, IMHO.

tchaikov (commented):
just a nit, could you add the "coroutine: "  prefix in the title of the commit message?

dawmd:
Done

tchaikov:
> just a nit, could you add the "coroutine: " prefix in the title of the commit message?

please note, the cover letter and the title of the PR is not the same thing as the commit of a certain commit in the PR. you might need to edit the commit locally and `git push --force` to update the remote branch.

dawmd:
Corrected. Thank you for the help

tchaikov (commented):
could you please s/Coroutine/coroutine/ in the commit message?

tchaikov (approved):
lgtm.

nyh (approved):

PR #1806, opened 2023-08-30
[cmake] portable build: only link with rt where it's available

leonid-s-usov:
This PR is in support of https://github.com/ceph/ceph/pull/52629

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -795,7 +795,7 @@ target_link_libraries (seastar
     GnuTLS::gnutls
     StdAtomic::atomic
     lksctp-tools::lksctp-tools
-    rt::rt
+    $<$<AND:$<BOOL:UNIX>,$<NOT:$<BOOL:APPLE>>>:rt::rt>
     yaml-cpp::yaml-cpp
     "$<BUILD_INTERFACE:Valgrind::valgrind>"
     Threads::Threads)


CMakeLists.txt
```diff
@@ -795,7 +795,7 @@ target_link_libraries (seastar
     GnuTLS::gnutls
     StdAtomic::atomic
     lksctp-tools::lksctp-tools
-    rt::rt
+    $<$<AND:$<BOOL:UNIX>,$<NOT:$<BOOL:APPLE>>>:rt::rt>
```
tchaikov, line 798: how come seastar can work without `timer_create()`  if it is built on Apple on UNIX ?

leonid-s-usov: @tchaikov thanks, good point! I investigated this and turns out, I don't have seastar compiled. It's OFF by default 
```
option(WITH_SEASTAR "Build seastar components")
```

Apparently, when I hit that linking issue I was still trying to compile with seastar ON, and later found out it wasn't working, but already had this change committed.
I'll now try to compile again, if it fails even with this fix - I'll close the PR.

Thanks again!

leonid-s-usov:
It appears that building seastar on a Mac is a bigger challenge than just this one tweak. Due to a misconfiguration, I was under the wrong impression that seastar was being built. Closing this PR for now; if I get to a point where seastar is required for my work I'll pick it up and try to make it work

Issue #1809, opened 2023-09-03
semaphore wait_list expiry handler is not updated when semaphore is moved


bhalevy:
The semaphore `_wait_list` is and `abortable_fifo` with an `expiry_handler` that keeps a reference to the semaphore:
https://github.com/scylladb/seastar/blob/fdcb43e39bda90acbe2e7ee6a75327c8c6c36180/include/seastar/core/semaphore.hh#L169-L170

The `expiry_handler` is constructed along with the semaphore:
https://github.com/scylladb/seastar/blob/fdcb43e39bda90acbe2e7ee6a75327c8c6c36180/include/seastar/core/semaphore.hh#L243

But, if the semaphore is moved, the reference to the semaphore is invalidated, and e.g. aborting a wait might end up with use-after-free or stack-use-after-return.

For example, consider the following test that creates a semaphore on the stack
which is moved to the heap.
```c++
SEASTAR_THREAD_TEST_CASE(test_abort_after_move) {
    auto sem = std::make_unique<semaphore>([] { return semaphore(0); }());
    abort_source as;
    as.request_abort();

    BOOST_REQUIRE_THROW(get_units(*sem, 1, as).get(), abort_requested_exception);
}
```

```
==4116189==ERROR: AddressSanitizer: stack-use-after-scope on address 0x7f23796b52d8 at pc 0x0000004af2ac bp 0x7f23798b6510 sp 0x7f23798b6508
READ of size 8 at 0x7f23796b52d8 thread T1
...
    #0 0x4af2ab in std::__exception_ptr::exception_ptr::operator bool() const /usr/include/c++/13/bits/exception_ptr.h:163
    #1 0x535e75 in seastar::basic_semaphore<seastar::semaphore_default_exception_factory, std::chrono::_V2::steady_clock>::expiry_handler::operator()(seastar::basic_semaphore<seastar::semaphore_default_exception_factory, std::chrono::_V2::steady_clock>::entry&) /home/bhalevy/dev/seastar/include/seastar/core/semaphore.hh:178
[...]

PR #1810, opened 2023-09-03
semaphore: initialize wait_list in move ctor

bhalevy:
The default move-ctor naively moves the wait list
expiry_handler while the latter keeps a reference
to the moved semaphore.

Using it after the move might cause use-after-free or stack-use-after-return.

Fixes scylladb/seastar#1809
Refs scylladb/scylladb#15253

### DIFF
--- a/include/seastar/core/semaphore.hh
+++ b/include/seastar/core/semaphore.hh
@@ -249,6 +249,43 @@ public:
     {
         static_assert(std::is_nothrow_move_constructible_v<expiry_handler>);
     }
+
+    /// Move-constructs a semaphore object from a moved-from semaphore object,
+    /// inheriting the number of units from the moved-from semaphore.
+    /// The moved-from semaphore must be unused.
+    ///
+    /// \param other the moved-from semaphore object.
+    basic_semaphore(basic_semaphore&& other) noexcept(std::is_nothrow_move_constructible_v<exception_factory>)
+        : exception_factory(other)
+        , _count(other._count)
+        , _ex(std::exchange(other._ex, std::exception_ptr()))
+        , _wait_list(expiry_handler{*this})
+        , _used(std::move(other._used))
+    {
+        // semaphore cannot be moved with non-empty waiting list
+        assert(other._wait_list.empty());
+    }
+
+    /// Move-assigns a semaphore object from a moved-from semaphore object,
+    /// inheriting the number of units from the moved-from semaphore.
+    /// The number of units of the assigned semaphore is overwritten by the
+    /// the moved-from number of units.
+    /// Both the moved-to and moved-from semaphores must be unused.
+    ///
+    /// \param other the moved-from semaphore object.
+    basic_semaphore& operator=(basic_semaphore&& other) noexcept(std::is_nothrow_move_assignable_v<exception_factory>) {
+        // semaphore cannot be moved with non-empty waiting list
+        assert(_wait_list.empty());
+        assert(other._wait_list.empty());
+        if (this != &other) {
+            exception_factory::operator=(other);
+            _count = other._count;
+            _ex = std::exchange(other._ex, std::exception_ptr());
+            _used = std::move(other._used);
+        }
+        return *this;
+    }
+
     /// Waits until at least a specific number of units are available in the
     /// counter, and reduces the counter by that amount of units.
     ///
--- a/tests/unit/semaphore_test.cc
+++ b/tests/unit/semaphore_test.cc
@@ -452,3 +452,58 @@ SEASTAR_THREAD_TEST_CASE(test_reassigned_units_are_returned) {
     wait.get();
     t.cancel();
 }
+
+SEASTAR_THREAD_TEST_CASE(test_get_units_after_move) {
+    auto sem = std::make_unique<semaphore>([] { return semaphore(0); }());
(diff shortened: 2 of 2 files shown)


include/seastar/core/semaphore.hh
```diff
@@ -249,6 +249,26 @@ public:
     {
         static_assert(std::is_nothrow_move_constructible_v<expiry_handler>);
     }
+    basic_semaphore(basic_semaphore&& x) noexcept(std::is_nothrow_move_constructible_v<exception_factory>)
+        : exception_factory(std::move(x))
+        , _count(x._count)
+        , _wait_list(expiry_handler{*this})
+    {
```
avikivity, line 256: Should we move/assign _ex too? moves should happen before _ex is assigned anything, but better to be safe.

bhalevy: yeah, makes sense.

include/seastar/core/semaphore.hh
```diff
@@ -249,6 +249,26 @@ public:
...
+        // semaphore cannot be moved with non-empty waiting list
+        assert(x._wait_list.empty());
```
nyh, line 258: You know what I think about assert()... This could have thrown instead of assert.
But, can't we change the waiting list's pointer, even if it's not empty?

include/seastar/core/semaphore.hh
```diff
@@ -249,6 +249,26 @@ public:
...
+    }
+
+    basic_semaphore& operator=(basic_semaphore&& x) noexcept(std::is_nothrow_move_assignable_v<exception_factory>) {
+        // semaphore cannot be moved with non-empty waiting list
+        assert(_wait_list.empty());
+        assert(x._wait_list.empty());
+        if (this != &x) {
+            exception_factory::operator=(x);
+            _count = x._count;
```
nyh, line 267: Maybe I'm missing something here, but don't you have to reset _wait_list? If you don't, then why insist that it's empty?

tests/unit/semaphore_test.cc
```diff
@@ -452,3 +452,58 @@ SEASTAR_THREAD_TEST_CASE(test_reassigned_units_are_returned) {
     wait.get();
     t.cancel();
 }
+
+SEASTAR_THREAD_TEST_CASE(test_get_units_after_move) {
```
nyh, line 456: just curious - did *all* these tests crash before the fix?

nyh (approved):

include/seastar/core/semaphore.hh
bhalevy: done in next version.
plus, moved also the used flag.

include/seastar/core/semaphore.hh
bhalevy: In 9b6e181e424fd18b780ade4df59e2c0c3a806742 we already restricted the move of the semaphore
after it was used (even if there were waiters in the past and they were already satisfied and no longer on the wait list).  The reason was to simplify move while there are waiters on the wait list.
So this patch follows the same lines.

include/seastar/core/semaphore.hh
bhalevy: this->_wait_list remains intact, ready for business for new waiters, therefore being empty is good enough that we can toss away the other.wait_list.

tests/unit/semaphore_test.cc
bhalevy: yes, they all fail due to the same root cause from slightly different paths.

bhalevy:
In [4dd7986](https://github.com/scylladb/seastar/commit/4dd7986c281aa6e09588c1892d894a60a4f9b149):
- moved also `_ex`, and `_used`
- added doxygen comments

bhalevy:
- reference scylladb/scylladb#15253 in commit message

include/seastar/core/semaphore.hh
avikivity: You can't throw in a noexecpt move constructor. And marking move constructors as throwing generally decreases stability since it adds exceptional paths that are never tested.

nyh: Right.

include/seastar/core/semaphore.hh
nyh: I understood why you want to check that x._wait_list.empty() because it' being tossed away, but why assert that this->wait_list is empty as well? Why do we care if there are waiters already, since the pointer to this remains value?

bhalevy: I'd like to avoid assigning to a used semaphore. That's even broader than non-empty wait list as it means there could be outstanding semaphore units that relate to the pre-assigned state of the semaphore and when returned, would increment the count that may now be changed as a side effect of the assignment.
I think we better restrict the usage to a safe model that is easy to remember than to deal with weird corner cases even if we can technically support that.

bhalevy: I'd like to avoid assigning to a used semaphore. That's even broader than non-empty wait list as it means there could be outstanding semaphore units that relate to the pre-assigned state of the semaphore and when returned, would increment the count that may now be changed as a side effect of the assignment.
I think we better restrict the usage to a safe model that is easy to remember than to deal with weird corner cases even if we can technically support that.

PR #1812, opened 2023-09-04
build, process: avoid using stdout or stderr as C++ identifiers

tchaikov:
musl libc defines `stdout` like
```c++
#define stdout (stdout)
```
while glibc defines `stdout` like
```c++
#define stdout stdout
```
the same applies to `stderr`. both these definitions comply to the the C++ standard, but we reuse `stdout` and `stderr` when defining an `enum class` or when defining a member function . so, for instance, after preprocessing, the definition from musl libc is expanded into
```c++
enum class logger_ostream_type {
  none,
  (stdout),
  (stderr),
};
```
which does not compile. so, in order to be more standard compliant, and does not rely on a specific libc implementation, while preserve the backward compatibility. let's check if `stdout` and `stderr` can be used as an identifier. also, since `stdout` and `stderr` are required to be defined as macros by the standard library. let's mark them deprecated in hope to remove them in a future version in favor of `cout` and `cerr` which are guranteed to be valid identifiers by the C++ standard.

but as the `[[deprecated]]` specifier cannnot be applied to a certain enumerator, let's use the ` __attribute__ ((deprecated...))` instead. fortunately, it is supported by both Clang and GCC.

this should address one of the FTBFS with musl libc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1033,6 +1033,18 @@ if (Seastar_DEBUG_SHARED_PTR IN_LIST True_STRING_VALUES OR
       SEASTAR_DEBUG_SHARED_PTR)
 endif ()
 
+include (CheckCXXSourceCompiles)
+file (READ ${CMAKE_CURRENT_LIST_DIR}/cmake/code_tests/stdout_test.cc _stdout_test_code)
+check_cxx_source_compiles ("${_stdout_test_code}" Stdout_Can_Be_Used_As_Identifier)
+if (Stdout_Can_Be_Used_As_Identifier)
+  # "stdout" is defined as a macro by the C++ standard, so we cannot assume
+  # that the macro is always expanded into an identifier which can be re-used
+  # to name a enumerator in the declaration of an enumeration.
+  target_compile_definitions (seastar
+    PUBLIC
+      SEASTAR_LOGGER_TYPE_STDOUT)
+endif ()
+
 set (Seastar_STACK_GUARD_MODES "Debug" "Sanitize" "Dev")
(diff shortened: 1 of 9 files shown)


a16bitsysop:
This fails on musl with error:

```
tests/unit/spawn_test.cc:67:31: error: expected unqualified-id before '(' token
   67 |         auto stdout = process.stdout();
      |                               ^~~~~~

```

Check for stdout?

```
-- Performing Test Stdout_Can_Be_Used_As_Identifier
-- Performing Test Stdout_Can_Be_Used_As_Identifier - Failed
```

full pipeline:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1102434

tchaikov:
@a16bitsysop thank you for testing! i missed the non-enum use cases. will add them as a dedicated commit in this PR. so it'd be easier for reviewers to review.

tchaikov:
@a16bitsysop hi Duncan, should be better now.

tchaikov:
v2:

- add a commit to rename the `stdout()` member function to `cout()` in the `process` class. 
- update the cover letter to cover the new commit

nyh (approved):

nyh:
Looks good to me, but @a16bitsysop can you please confirm it works with musl?

a16bitsysop:
Yes it works fine on musl replacing the stdout patch from my pr, also now passes for aarch64 again

full x86_64 pipeline:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1102950

tchaikov:
@nyh @a16bitsysop thank your review and test, Nadav and Duncan!

PR #1813, opened 2023-09-05
http/client: pass tls_options to tls::connect()

tchaikov:
in e7e51612, we moved the `server_name` into `tls_options`, and deprecated the overloads of `tls::connect()` which accept `server_name` in favor of the ones which accept `tls::tls_options`. but we failed to update the caller of the deprecated function in src/htpp/client.cc.

to use the new API and to address the deprecation warning, let's update this caller as well.

Issue #602, opened 2019-02-12
Can't deprecate a class


avikivity:
If I add [[deprecated]], the build fails because of -Werror and no -Wno-error-deprecated (or whatever). This is a regression from the pre-cmake days.

avikivity:
@hakuch example trigger: https://github.com/avikivity/seastar/commit/d62bc3865c20ff35a736a2461c1a3d7bb529b36d

hakuch:
The solution (if we wish to keep `-Werror` enabled) is to add `-Wno-deprecated` to `Seastar_DEVELOPMENT_FLAGS`.

Do you have any objection to that, @avikivity?

avikivity:
That means we won't get warnings for deprecated names.

There's -Wno-error=deprecated-declarations that emits warnings but does not upgrade them to errors.

hakuch:
Let's do that, then!

PR #1817, opened 2023-09-07
build: reenable -Warray-bound and -Wdeprecated-declarations

tchaikov:
-Warray-bound was disabled in bdbce89e, by then we had false alarms from GCC 9, but now that the fix of the false alarm was merged in GCC 10, and GCC 13 was already released. since Seastar supports up to two major releases of GCC, we can safely drop this workaround.

-Wno-error=deprecated-declarations was added in 861da6e3. we wanted to be consistent with the behavior before switching to CMake. but, it would be better to encourage the developer to address the places where the deprecated thing are used when he/she adds the `[[deprecated]]` attribute.

Refs #602
Refs #1813

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -767,9 +767,7 @@ set (Seastar_PRIVATE_CXX_FLAGS
   -UNDEBUG
   -Wall
   -Werror
-  -Wimplicit-fallthrough
-  -Wno-array-bounds # Disabled because of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93437
-  -Wno-error=deprecated-declarations)
+  -Wimplicit-fallthrough)
 if (NOT BUILD_SHARED_LIBS)
   list (APPEND Seastar_PRIVATE_CXX_FLAGS -fvisibility=hidden)
 endif ()


avikivity:
How can we error on a using a deprecated function? Sometimes the implementation of one deprecated API will need to call another, or testing will.

Also, we shouldn't export this to the enclosing project (but maybe that's already the case, I don't know how cmake does that).

tchaikov:
@avikivity 

> How can we error on a using a deprecated function? Sometimes the implementation of one deprecated API will need to call another, or testing will.

if "we" means Seastar, i think we can, and we should. for instance, if Seastar deprecates some of its own functions, Seastar should be the first one who stop using them. if "we" means application / libraries using Seastar, see below. the option helps us to identify them. but if we have to, for instance, test the deprecated API, we should either restructure the code so that the deprecated API calls into the one which replaces it, and test the underlying function. if we have to, i think we can disable the warning temporarily just for the test.

> 
> Also, we shouldn't export this to the enclosing project (but maybe that's already the case, I don't know how cmake does that).

we are not exporting this to the enclosing project. typically, a project are using Seastar in one of following three ways:

- a project using CMake can include Seastar's CMake directly. in this case, `Seastar_PRIVATE_CXX_FLAGS` is not populated back to its parent project. this is how scylladb's CMake based building system works.
- a project using CMake can import Seastar's `SeastarConfig.cmake`. similarly, `Seastar_PRIVATE_CXX_FLAGS` is not included by it. this is how Redpand uses Seastar.
- a project using `pkgconf` to find Seastar. `seastar.pc` does not include `Seastar_PRIVATE_CXX_FLAGS` in its `seastar_cflags`. this is how scylladb's `configure.py` based building system works.

tchaikov:
i extracted a subset of this change into #1819 in hope to get the latter merged sooner.

avikivity:
It can happen that Seastar itself uses deprecated classes/functions to implement other deprecated classes/functions, or to test them.

tchaikov:
> It can happen that Seastar itself uses deprecated classes/functions to implement other deprecated classes/functions, or to test them.

if we want to allow Seastar to use the deprecated classes/functions. then this pull request is not relevant anymore. personally, i'd avoid using deprecated facilities if possible, even for the purpose to implement other deprecated classes/functions, as explained at https://github.com/scylladb/seastar/pull/1817#issuecomment-1710987833 .

PR #1518, opened 2023-02-26
treewide: bring back prometheus protobuf support

tchaikov:
This reverts commit 8ed9771ae98130d5adbccdaea4ecbac738170939. also adapts to the latest master HEAD of Seastar

quote from https://prometheus.io/docs/instrumenting/exposition_formats/#historical-versions

> However, beginning with Prometheus v2.40, there is experimental
> support for native histograms, which – at least in its initial
> experimental state – utilizes the old Protobuf format (with some
> newer additions) again. Therefore, a very recent Prometheus
> server with the experimental native histogram support enabled,
> will again be able to ingest the Protobuf format. This support
> is experimental and might get removed again.

Fixes #1510
Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>

PR #1818, opened 2023-09-07
Bringing back Prometheus protobuf support

amnonh:
This PR is based on PR #1518 and includes an additional commit that adds a flag to the Prometheus configuration object disabling protobuf by default.

The original PR #1518  resurrected the protobuf support that was removed in 8ed9771ae98130d5adbccdaea4ecbac738170939.

The motivation is the experimental Prometheus protobuf with native histogram support. After this series, it will be possible to adopt Prometheus protobuf support to the newer feature of the text representation and add Prometheus native histogram support.

### DIFF
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -800,19 +800,19 @@ public:
 
     future<std::unique_ptr<http::reply>> handle(const sstring& path,
         std::unique_ptr<http::request> req, std::unique_ptr<http::reply> rep) override {
-        auto text = is_accept_text(req->get_header("Accept"));
+        auto is_text_format = !_ctx.allow_protobuf || is_accept_text(req->get_header("Accept"));
         sstring metric_family_name = req->get_query_param("__name__");
         bool prefix = trim_asterisk(metric_family_name);
         bool show_help = req->get_query_param("__help__") != "false";
         std::function<bool(const mi::labels_type&)> filter = make_filter(*req);
 
-        rep->write_body((text) ? "txt" : "proto", [this, text, metric_family_name, prefix, show_help, filter] (output_stream<char>&& s) {
+        rep->write_body(is_text_format ? "txt" : "proto", [this, is_text_format, metric_family_name, prefix, show_help, filter] (output_stream<char>&& s) {
             return do_with(metrics_families_per_shard(), output_stream<char>(std::move(s)),
-                    [this, text, prefix, &metric_family_name, show_help, filter] (metrics_families_per_shard& families, output_stream<char>& s) mutable {
-                return get_map_value(families).then([&s, &families, this, text, prefix, &metric_family_name, show_help, filter]() mutable {
+                    [this, is_text_format, prefix, &metric_family_name, show_help, filter] (metrics_families_per_shard& families, output_stream<char>& s) mutable {
+                return get_map_value(families).then([&s, &families, this, is_text_format, prefix, &metric_family_name, show_help, filter]() mutable {
                     return do_with(get_range(families, metric_family_name, prefix),
-                            [&s, this, text, show_help, filter](metric_family_range& m) {
-                        return (text) ? write_text_representation(s, _ctx, m, show_help, filter) :
+                            [&s, this, is_text_format, show_help, filter](metric_family_range& m) {
+                        return (is_text_format) ? write_text_representation(s, _ctx, m, show_help, filter) :
                                 write_protobuf_representation(s, _ctx, m);
                     });
                 }).finally([&s] () mutable {
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -20,6 +20,9 @@
  */
 
 #include <seastar/core/prometheus.hh>
+#include <google/protobuf/io/coded_stream.h>
+#include <google/protobuf/io/zero_copy_stream_impl_lite.h>
+#include "proto/metrics2.pb.h"
 #include <sstream>
 
 #include <seastar/core/metrics_api.hh>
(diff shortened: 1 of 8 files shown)


amnonh:
@avikivity this is @tchaikov PR with the additional configuration option to disable protobuf by default

tchaikov:
@amnonh, despite that the failures should not be relevant, could you please repush your changes to retrigger the CI, see if the problem persists?

amnonh:
@tchaikov Rebased and forced push

tchaikov:
thank you @amnonh . it's a new issue. filed at #1824

amnonh:
@tchaikov I see that there are conflicts now

tchaikov:
> @tchaikov I see that there are conflicts now

@amnonh could you try to rebase your change onto https://github.com/tchaikov/seastar/tree/protobuf and repush?

amnonh:
hi @tchaikov can you take a look?

amnonh:
@tchaikov ping

src/core/prometheus.cc
```diff
@@ -806,7 +806,7 @@ class metrics_handler : public httpd::handler_base  {
         bool show_help = req->get_query_param("__help__") != "false";
         std::function<bool(const mi::labels_type&)> filter = make_filter(*req);
 
-        rep->write_body((text) ? "txt" : "proto", [this, text, metric_family_name, prefix, show_help, filter] (output_stream<char>&& s) {
+        rep->write_body((text || !_ctx.allow_protobuf) ? "txt" : "proto", [this, text, metric_family_name, prefix, show_help, filter] (output_stream<char>&& s) {
```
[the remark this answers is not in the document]
tchaikov: we are using `text` at line 815 as the source of truth to tell if we should return the text representation or its protobuf representation. so i'd suggest mutate the returned `text` based on if protobuf support is enabled. also the variable name of `text` could be misleading. so i'd suggest change to something like:
```c++
bool is_text_accepted(const std::string& accept, bool allows_protobuf) {
    /// throw an error with not_acceptable if we are not able to handle 
    /// the request
}

bool accepts_text = is_text_accepted(req->get_header("Accept"), 
    _ctx.allow_protobuf);
sstring content_type{accepts_text ? "text" : "proto"};
rep->write_body(content_type, ///
  ///...
      if (accepts_text) {
          return write_text_representation(...);
      } else {
          return write_protobuf_representation(...);
      }
  }
```

tchaikov (commented):
@amnonh hi Amnon, thank you for your patience. i left a couple comment.

src/core/prometheus.cc
```diff
@@ -632,6 +712,38 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx,
...
+            fill_metric(mtf, value, value_info.id, ctx);
+        });
+        if (!write_delimited_to(mtf, &os)) {
+            seastar_logger.warn("Failed to write protobuf metrics");
+        }
+        return out.write(s);
+    });
+}
+
+bool is_accept_text(const std::string& accept) {
```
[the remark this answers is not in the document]
amnonh: We can revert the logic, right now it return if text accepted.

src/core/prometheus.cc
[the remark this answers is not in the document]
amnonh: We are returning the functionality that was there (with an additional configuration flag parameter). What we want, that after the change unless someone really wants protobuf representation, they will get text representation.

In the current situation, we should return text representation even if someone asks for protobuf, if the allow_protobuf is set to false, this is alligned with the current Seastar implementation that always return text.

amnonh:
@tchaikov I've push an additional commit that revert the logic of is_accept_text to is_accept_protobuf, I've also changed it to static and change the text parameter to text_format.

I cannot compile it locally because of issues with the cmake, so I would appriciate if you can take a look

src/core/prometheus.cc
```diff
@@ -800,19 +800,19 @@ class metrics_handler : public httpd::handler_base  {
 
     future<std::unique_ptr<http::reply>> handle(const sstring& path,
         std::unique_ptr<http::request> req, std::unique_ptr<http::reply> rep) override {
-        auto text = is_accept_text(req->get_header("Accept"));
+        auto text_format = !_ctx.allow_protobuf || !is_accept_protobuf(req->get_header("Accept"));
```
tchaikov, line 803: i'd suggest revert this condition as well:
```c++
bool accepts_protobuf = _ctx.allow_protobuf && is_accept_protobuf(req->get_header("Accept"));
```

src/core/prometheus.cc
```diff
@@ -800,19 +800,19 @@ class metrics_handler : public httpd::handler_base  {
...
         sstring metric_family_name = req->get_query_param("__name__");
         bool prefix = trim_asterisk(metric_family_name);
         bool show_help = req->get_query_param("__help__") != "false";
         std::function<bool(const mi::labels_type&)> filter = make_filter(*req);
 
-        rep->write_body((text || !_ctx.allow_protobuf) ? "txt" : "proto", [this, text, metric_family_name, prefix, show_help, filter] (output_stream<char>&& s) {
+        rep->write_body((text_format) ? "txt" : "proto", [this, text_format, metric_family_name, prefix, show_help, filter] (output_stream<char>&& s) {
```
tchaikov, line 809: could take this opportunity to drop the parenthesis around `text_format`. again, the variable name could be confusing. as it does not indicate the `text_format`.

src/core/prometheus.cc
```diff
@@ -800,19 +800,19 @@ class metrics_handler : public httpd::handler_base  {
...
             return do_with(metrics_families_per_shard(), output_stream<char>(std::move(s)),
-                    [this, text, prefix, &metric_family_name, show_help, filter] (metrics_families_per_shard& families, output_stream<char>& s) mutable {
-                return get_map_value(families).then([&s, &families, this, text, prefix, &metric_family_name, show_help, filter]() mutable {
+                    [this, text_format, prefix, &metric_family_name, show_help, filter] (metrics_families_per_shard& families, output_stream<char>& s) mutable {
+                return get_map_value(families).then([&s, &families, this, text_format, prefix, &metric_family_name, show_help, filter]() mutable {
                     return do_with(get_range(families, metric_family_name, prefix),
-                            [&s, this, text, show_help, filter](metric_family_range& m) {
-                        return (text) ? write_text_representation(s, _ctx, m, show_help, filter) :
+                            [&s, this, text_format, show_help, filter](metric_family_range& m) {
+                        return (text_format) ? write_text_representation(s, _ctx, m, show_help, filter) :
```
tchaikov, line 815: ditto.

tchaikov, line 815: i'd suggest include this very change in the previous one. as it addresses a bug in it.

tchaikov:
> @tchaikov I've push an additional commit that revert the logic of is_accept_text to is_accept_protobuf, I've also changed it to static and change the text parameter to text_format.
> 
> I cannot compile it locally because of issues with the cmake, so I would appriciate if you can take a look

you mean take a look at the issues with the cmake, right? may i learn what kind of the issues you're facing?

src/core/prometheus.cc
amnonh: But It does, text_format means we are using text format

src/core/prometheus.cc
amnonh: I can squas the changes, I tried to keep the original as close as possible to make it easier for maintainer to understand what was changed.

amnonh:
> 
> you mean take a look at the issues with the cmake, right? may i learn what kind of the issues you're facing?

I sort local compilation, the issue with the broken tests remains

src/core/prometheus.cc
tchaikov: sorry for the confusion. i mean, it does not contain the information about the "text_format", but about if the "text" content type is used. that's why i suggest to use name like "accepts_text" or "is_text_accepted" in previous comment.

src/core/prometheus.cc
tchaikov: the bug was introduced in the commit of "[Prometheus: Disable Prometheus protobuf with a configuration](https://github.com/scylladb/seastar/pull/1818/commits/f3d8620cdfe5798b643c323a135cf109142a65f5)", so you are fixing the issue in the very same PR which introduces it.

tchaikov:
> > you mean take a look at the issues with the cmake, right? may i learn what kind of the issues you're facing?
> 
> I sort local compilation, the issue with the broken tests remains

i am confused. how are the test failures related to cmake?

amnonh:
@tchaikov I've forced push a minimal change that would keep the original code as much as possible. Reverting the logic will be done in a follow-up patch and would be done explicitly to make it clear what is changed and why

tchaikov (approved):
lgtm. probably @avikivity might want to take a final look.

tchaikov:
@amnonh JFYI, i resurrected #1518 to see if the test failures can be reproduced at least in another pull request.

tchaikov:
the same tests in #1518 also are failing. i tested locally by repeating the same steps in `.circleci` but with a newer kernel 6.5.6, the tests passed with the dev build. will try with the same kernel later.

amnonh:
@tchaikov how can we progress?

tchaikov:
> @tchaikov how can we progress?

@amnonh i'd suggest fix #1824 before moving on. as the test failures are reproduced consistently. without fixing it, we would lose the test coverage previously offered by the dev builds.

amnonh:
@avikivity finally whatever caused the test to failed (probably #1824) was solved by rebase.
Can you merge the PR? to recap, it's the original protobuf code with an extra parameter that disable it (so even if someone would try and ask for protobuf they would get the text format)

avikivity:
Please change the subject and cover letter to something meaningful.

src/core/prometheus.cc
```diff
@@ -800,19 +800,19 @@ class metrics_handler : public httpd::handler_base  {
 
     future<std::unique_ptr<http::reply>> handle(const sstring& path,
         std::unique_ptr<http::request> req, std::unique_ptr<http::reply> rep) override {
-        auto text = is_accept_text(req->get_header("Accept"));
+        auto is_text_format = !_ctx.allow_protobuf || is_accept_text(req->get_header("Accept"));
```
avikivity, line 803: I don't understand the logic. Suppose the header only accepts protobuf. We should respond with an error of protobuf is not configured, instead we respond with text.

amnonh: We are keeping the current behavior, right now, Prometheus does not check the header and will return text format regardless of what the client asks.
I believe this is an ok behavior.

If we want to change the current behavior, it should be done with an explicit issue and a warning that theoretically it could be a breaking change.

avikivity: ok, I see is_accept_text() is really is_not_accept_protobuf().

amnonh: @avikivity I've updated the title and cover

PR #1819, opened 2023-09-08
build: reenable -Warray-bound compiling option

tchaikov:
-Warray-bound was disabled in bdbce89e, by then we had false alarms from GCC 9, but now that the fix of the false alarm was merged in GCC 10, and GCC 13 was already released. since Seastar supports up to two major releases of GCC, we can safely drop this workaround.


tchaikov:
this change is a subset of #1817 . it was created to have smaller and more atomic changes. also, this change should be less controversial than #1817.

PR #1820, opened 2023-09-09
Demo application for metrics.

kfirgollan:
Simple demo application that showcases how to use metrics. It provides a simple class that exports metrics with example usage.


demos/metrics_demo.cc
```diff
@@ -0,0 +1,137 @@
...
+#include <fmt/format.h>
+
+using namespace seastar;
+using namespace std::chrono_literals;
+namespace sm = seastar::metrics;
+namespace bpo = boost::program_options;
+
+logger applog("app");
+
+class MetricsExample {
```
xemul, line 40: According to [naming guide](https://github.com/scylladb/seastar/blob/master/coding-style.md#naming) it should be _-separated

demos/metrics_demo.cc
```diff
@@ -0,0 +1,137 @@
...
+            PrometheusExample server(address, pport, prefix);
+
+            applog.info("Starting with metrics 1");
+            {
+                MetricsExample metrics(1);
+                for (int i = 0; i < 10; i++) {
+                    sleep(1s).get();
+                    metrics.increase(5);
+                }
+            }
```
xemul, line 121: I don't quite get what the above block is for. It will just wipe everything without a trace in 10 seconds

kfirgollan: The point is to demonstrate that when the containing object gets out of scope then the metrics are no longer reported.
While iterating it on it I simply used `curl http://localhost:9180/metrics`, to see the data live.
My thoughts were on to put an emphasis that its most likely preferable to keep the object alive as long as some monitoring system cares about it.

Do you think this point is clear without this part? Any alternative approach to explain this? or eliminate it all together?

xemul: I would eliminate it altogether and made the demo just start the server "forever" and printing a message to the user saying something like

```
Now try `curl http://127.0.0.1:123/metrics` to see how metrics are exported
```

demos/metrics_demo.cc
```diff
@@ -0,0 +1,137 @@
...
+private:
+    sm::metric_groups _metrics;
+    uint64_t _counter;
+    uint64_t _triggered;
+
+public:
+    metrics_example(int instance) : _counter(0) {
+        // Note that these metrics are only available as long at the _metrics instance
+        // is valid.
+        _metrics.add_group(fmt::format("metrics_demo_{}", instance), {
```
amnonh, line 50: If it was a real code, I would say instane should be a label, I know it's a demo, but maybe we should
be correct.

demos/metrics_demo.cc
```diff
@@ -0,0 +1,137 @@
...
+            // Access the member each time the metric is read.
+            // Note that counter is expected to be monotonically increasing.
+            sm::make_counter("counter", _counter, sm::description("A counter that counts.")),
+            // A function that can return any integer
+            sm::make_counter("function", [this]() {applog.info("triggered"); return ++_triggered;}, sm::description("A function which returns a value.")),
+            // A gauge is a counter that can go up and down, can be used for rates
+            sm::make_gauge("gauge", []() {return 3;}, sm::description("random number"))
```
amnonh, line 57: It's not random, it's arbitrary

demos/metrics_demo.cc
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * In order to access the metrics we need to use one of the builtin mechanisms.
+ * For the sake of this example we are using prometheus.
+ * To view the specific metrics of the demo use something in the lines of
+ *  curl http://localhost:9180/metrics | grep "metrics_demo"
+ * 
+ */
+int main(int argc, char** argv) {
+    seastar::app_template app;
+    app.add_options()("prometheus_port", bpo::value<uint16_t>()->default_value(9180), "Prometheus port. Set to zero in order to disable.");
```
amnonh, line 102: You can use metrics relabel to turn off all metrics besides the one used for Demo, not sure if it makes things clearer or not.

kfirgollan: Could you give an example for how the relabel process?

From the config docs, it seems like its not intended to be used, so I guess shouldn't be included in the example usage?
```cpp
/*!
 * Holds prometheus related configuration
 */
struct config {
    sstring metric_help; //!< Default help message for the returned metrics
    sstring hostname; //!< hostname is deprecated, use label instead
    std::optional<metrics::label_instance> label; //!< A label that will be added to all metrics, we advice not to use it and set it on the prometheus server
    sstring prefix = "seastar"; //!< a prefix that will be added to metric names
};
```

amnonh:
LGTM but please add a capture of the result metrics (e.g. the result from the ```curl```)

kfirg-cetu:
Any updates here?

PR #1822, opened 2023-09-10
build: check if libucontext is necessary for using ucontext functions

tchaikov:
musl libc does not offer the symbols of getcontext() and friends. it merely provides the header file of `ucontext.h` with the declarations of these functions so that "gcc with dwarf2 unwinding" can be built without errors. these functions were removed from POSIX. fortunately, libucontext provides a portable implementation.

in this change, if the function of `getcontext()` can not be compiled or linked, we fall back to check libucontext, and use it as a fallback.

this should improve portability of Seastar, so it can compile on platforms where glibc is not available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -401,6 +401,7 @@ seastar_find_dependencies ()
 
 # Private build dependencies not visible to consumers
 find_package (ragel 6.10 REQUIRED)
+find_package (ucontext REQUIRED)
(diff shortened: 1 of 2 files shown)


a16bitsysop:
this compiles file for alpine: `-- Found ucontext: /lib/libucontext.so`

full pipeline: https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1108479

PR #1823, opened 2023-09-10
build: enable systemtap only if it is available

### LINKED ISSUE #557: cmake should check for `sys/sdt.h` offered by `systemtap-sdt-dev`
otherwise `make` could fail like

```
/var/ceph/ceph-ubuntu-16.04-kefu/src/seastar/src/core/reactor.cc:100:10: fatal error: sys/sdt.h: No such file or directory
 #include <sys/sdt.h>
          ^~~~~~~~~~~
compilation terminated.
CMakeFiles/seastar.dir/build.make:412: recipe for target 'CMakeFiles/seastar.dir/src/core/reactor.cc.o' failed
make[2]: *** [CMakeFiles/seastar.dir/src/core/reactor.cc.o] Error 1
```

tchaikov:
systemtap is not available in musl libc. and we should detect at the configure time.

in this change, we check for the existence of `sys/sdt.h`, and define a placeholder for `STAP_PROBE()` macro if the header is not available.

Fixes #557

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -404,6 +404,7 @@ find_package (ragel 6.10 REQUIRED)
 find_package (Threads REQUIRED)
 find_package (PthreadSetName REQUIRED)
 find_package (Valgrind REQUIRED)
+find_package (SystemTap-SDT)
 
 #
 # Code generation helpers.
@@ -998,6 +999,13 @@ if (Seastar_NUMA)
 
   target_link_libraries (seastar
     PRIVATE numactl::numactl)
+endif ()
+
+if (SystemTap-SDT_FOUND)
+  list (APPEND Seastar_PRIVATE_COMPILE_DEFINITIONS SEASTAR_HAVE_SYSTEMTAP_SDT)
+
+  target_link_libraries (seastar
+    PRIVATE SystemTap::SDT)
 endif ()
 
 seastar_supports_flag ("-Werror=unused-result" ErrorUnused_FOUND)
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -95,7 +95,11 @@ module;
 #include <sys/mman.h>
 #include <sys/utsname.h>
 #include <linux/falloc.h>
+#ifdef SEASTAR_HAVE_SYSTEMTAP_SDT
 #include <sys/sdt.h>
+#else
+#define STAP_PROBE(provider, name)
+#endif
 
 #ifdef HAVE_OSV
 #include <osv/newpoll.hh>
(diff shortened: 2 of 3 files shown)


src/core/reactor.cc
```diff
@@ -95,7 +95,11 @@ module;
 #include <sys/mman.h>
 #include <sys/utsname.h>
 #include <linux/falloc.h>
+#ifndef SEASTAR_HAVE_SYSTEMTAP_SDT
```
a16bitsysop, line 98: this needs to be #ifdef or it includes sys/sdt.h if not found

tchaikov: @a16bitsysop ahh, right. fixed.

tchaikov:
v2:

- should `#include <sys/sdt.h>`,  when `SEASTAR_HAVE_SYSTEMTAP_SDT` is defined. not the other way around.

a16bitsysop:
that works `-- Could NOT find SystemTap-SDT (missing: SystemTap-SDT_INCLUDE_DIR)`

full pipeline: https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1108506

CMakeLists.txt
```diff
@@ -1000,6 +1001,13 @@ if (Seastar_NUMA)
     PRIVATE numactl::numactl)
 endif ()
 
+if (SystemTap-SDT_FOUND)
+  list (APPEND Seastar_PRIVATE_COMPILE_DEFINITIONS SEASTAR_HAVE_SYSTEMTAP_SDT)
+
+  target_link_libraries (seastar
+    PRIVATE SystemTap::SDT)
+endif ()
```
avikivity, line 1009: What library is needed here? AFAIK none.

src/core/reactor.cc
```diff
@@ -95,7 +95,11 @@ module;
 #include <sys/mman.h>
 #include <sys/utsname.h>
 #include <linux/falloc.h>
+#ifdef SEASTAR_HAVE_SYSTEMTAP_SDT
 #include <sys/sdt.h>
```
avikivity, line 99: Maybe it's enough to use __has_include.

CMakeLists.txt
tchaikov: the library is but the header provided by systemtap-sdt-dev. normally it's installed into `/usr/include`. IMHO, to use `find_package()` is just a way to 

1. make it explicit that the library is not a part of libc or system (linux) header
2. `find_path()` searches with a set of predefined rules for finding the header file. the rules allows use to customize the place where the header to be found without changing the building system. for instance, he/she can define `<PackageName>_ROOT` and `find_path()` will search the header in it.

src/core/reactor.cc
tchaikov: @avikivity  that'd work also. actually i considered it. but again, i'd prefer searching non-standard library using `find_path()`. if you insist, i will use `__has_include`. could please let me know that you think.

avikivity: ok, if we use cmake above then we might as well use it here.

Issue #1824, opened 2023-09-12
segfault found in dev build


tchaikov:
this is spotted in our CI. and is found on all 4 combinations of {clang-15, gcc-12} and c++standard 17, c++standard 20
```
          4 - Seastar.unit.alien (Failed)
         16 - Seastar.unit.directory (Failed)
         31 - Seastar.unit.httpd (Failed)
         51 - Seastar.unit.smp (Failed)
         58 - Seastar.unit.thread_context_switch (Failed)
         59 - Seastar.unit.timer (Failed)
         73 - Seastar.app.memcached.memcached (Failed)
```
```
[0/1] cd /home/circleci/project/build/dev/tests/unit && /usr/bin/cmake -E env ASAN_OPTIONS=disable_coredump=0:abort_on_error=1:detect_stack_use_after_return=1:verify_asan_link_order=0 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1 BOOST_TEST_CATCH_SYSTEM_ERRORS=no /home/circleci/project/build/dev/tests/unit/timer_test -c 2
INFO  2023-09-11 12:56:29,828 seastar - Reactor backend: linux-aio
Segmentation fault on shard 0.
Backtrace:
  /home/circleci/project/build/dev/libseastar.so+0x30b2dc
  /home/circleci/project/build/dev/libseastar.so+0x3375d6
  /lib/x86_64-linux-gnu/libc.so.6+0x3bcef
  /home/circleci/project/build/dev/libseastar.so+0x286a5f
  /home/circleci/project/build/dev/libseastar.so+0x272000
  /home/circleci/project/build/dev/libseastar.so+0x267f19
  /home/circleci/project/build/dev/libseastar.so+0x267d63
  /home/circleci/project/build/dev/libseastar.so+0x488972
  /home/circleci/project/build/dev/libseastar.so+0x267c69
  /home/circleci/project/build/dev/libseastar.so+0x308eff
  /home/circleci/project/build/dev/libseastar.so+0x308100
  /home/circleci/project/build/dev/libseastar.so+0x3236ad
  /home/circleci/project/build/dev/libseastar.so+0x3255d8
  /home/circleci/project/build/dev/libseastar.so+0x28f584
  0x590a
  /lib/x86_64-linux-gnu/libc.so.6+0x2350f
  /lib/x86_64-linux-gnu/libc.so.6+0x235c8
  0x5404
Segmentation fault
[...]

PR #1825, opened 2023-09-13
linux-aio: better error handling when io_setup() fails

tchaikov:
we can raise exception with more accurate error message according to `errno`. so in this series, let's fill `std::runtime_error` with the more specific error message.

### DIFF
--- a/src/core/linux-aio.cc
+++ b/src/core/linux-aio.cc
@@ -172,8 +172,16 @@ void setup_aio_context(size_t nr, linux_abi::aio_context_t* io_context) {
     if (r < 0) {
         char buf[1024];
         char *msg = strerror_r(errno, buf, sizeof(buf));
-        auto aio_max_nr = read_first_line_as<unsigned>("/proc/sys/fs/aio-max-nr");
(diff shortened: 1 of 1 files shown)


tchaikov:
inspired by https://github.com/scylladb/seastar/issues/1824#issuecomment-1716073479, where `errno` was actually `ENOSYS`. my hunch is that the alpine's CI was testing a cross compiled seastar. so the `SYS_io_setup` does not match.

a16bitsysop:
The log file I attached was from tests running in docker qemu from X86_64 host, but it seems to match failures on alpine CI which is arch64 hardware.  But alpine CI won’t be a privileged container.  I think that would be the same for running tests in other Distros CI that these tests would fail if they need a privileged container.  But it does pass in the alpine CI for x86_64.

It looks like it might be possible to add an expected failure Regex to the tests to catch this?

tchaikov:
@a16bitsysop i still suggest create a dedicated issue for tracking this issue instead commenting on this PR.

a16bitsysop:
Okay will do

PR #1827, opened 2023-09-13
scripts: Fix escaping of regexes

StephanDollberg:
The existing regexes didn't escape backslash correctly:
using say '\S' rather than '\\S' or r'\S'. This seems to mostly
have been working by luck as if an escape character isn't
recognized by python it will pass through the invalid
escape backslash and all.


tchaikov (approved):
lgtm.

yeah, these are actually invalid escape sequences. FWIW, `python -Wdefault` or `python -X dev` would complain at seeing them.

tchaikov:

PR #1828, opened 2023-09-13
core: add sleep_abortable instantiation for manual_clock

dotnwat:
This is useful for unit testing classes which use
sleep_abortable: all the types work, we just
weren't creating the symbol in the .cc file.


tchaikov (approved):
lgtm

PR #1830, opened 2023-09-15
scripts: use raw string to avoid invalid escape sequences

tchaikov:
we use, for instance '\s' and '\S' in regular expressions, in which '\s' is a two-char string, but not one of the escaped sequences defined by
https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

despite that CPython just considers "\s" as "\\s" like a two-char string. but if we run the script in developer mode using `python3 -Xdev`, it'd warn us like:

```
DeprecationWarning: invalid escape sequence '\S'
```

this change

* use `r` prefix when appropriate
* escapes the "\" manually otherwise.

### DIFF
--- a/scripts/perftune.py
+++ b/scripts/perftune.py
@@ -171,21 +171,22 @@ def restart_irqbalance(banned_irqs):
     perftune_print("Restarting irqbalance: going to ban the following IRQ numbers: {} ...".format(", ".join(banned_irqs_list)))
 
     # Search for the original options line
-    opt_lines = list(filter(lambda line : re.search("^\s*{}".format(options_key), line), cfile_lines))
+    opt_lines = list(filter(lambda line : re.search(r"^\s*{}".format(options_key), line), cfile_lines))
     if not opt_lines:
         new_options = "{}=\"".format(options_key)
     elif len(opt_lines) == 1:
         # cut the last "
-        new_options = re.sub("\"\s*$", "", opt_lines[0].rstrip())
+        new_options = re.sub(r'"\s*$', "", opt_lines[0].rstrip())
         opt_lines = opt_lines[0].strip()
     else:
         raise Exception("Invalid format in {}: more than one lines with {} key".format(config_file, options_key))
 
     for irq in banned_irqs_list:
         # prevent duplicate "ban" entries for the same IRQ
-        patt_str = "\-\-banirq\={}\Z|\-\-banirq\={}\s".format(irq, irq)
+        opt = f"--banirq={irq}"
+        patt_str = rf"{opt}\Z|{opt}\s"
         if not re.search(patt_str, new_options):
-            new_options += " --banirq={}".format(irq)
+            new_options += f" {opt}"
 
     new_options += "\""
 
@@ -196,7 +197,7 @@ def restart_irqbalance(banned_irqs):
     else:
         with open(config_file, 'w') as cfile:
             for line in cfile_lines:
-                if not re.search("^\s*{}".format(options_key), line):
+                if not re.search(r"^\s*{}".format(options_key), line):
                     cfile.write(line)
 
             cfile.write(new_options + "\n")


tchaikov:
this change is inspired by #1827

tchaikov:
@vladzcloudius hi Vlad, this is but a tiny cleanup. could you help review it?

scripts/perftune.py
```diff
@@ -171,19 +171,19 @@ def restart_irqbalance(banned_irqs):
...
 
     # Search for the original options line
-    opt_lines = list(filter(lambda line : re.search("^\s*{}".format(options_key), line), cfile_lines))
+    opt_lines = list(filter(lambda line : re.search(r"^\s*{}".format(options_key), line), cfile_lines))
     if not opt_lines:
         new_options = "{}=\"".format(options_key)
     elif len(opt_lines) == 1:
         # cut the last "
-        new_options = re.sub("\"\s*$", "", opt_lines[0].rstrip())
+        new_options = re.sub("\"\\s*$", "", opt_lines[0].rstrip())
```
bhalevy, line 179: This is a bit confusing.
How about using `r"\"\s*$"`?

vladzcloudius: This change looks wrong in general, @tchaikov.
Following the patch description what you want here is `r'"\s*$'`

scripts/perftune.py
```diff
@@ -171,19 +171,19 @@ def restart_irqbalance(banned_irqs):
...
         opt_lines = opt_lines[0].strip()
     else:
         raise Exception("Invalid format in {}: more than one lines with {} key".format(config_file, options_key))
 
     for irq in banned_irqs_list:
         # prevent duplicate "ban" entries for the same IRQ
-        patt_str = "\-\-banirq\={}\Z|\-\-banirq\={}\s".format(irq, irq)
+        patt_str = r"\-\-banirq\={}\Z|\-\-banirq\={}\s".format(irq, irq)
```
vladzcloudius, line 186: `r"--banirq={}\Z|--banirq={}\s".format(irq, irq)`

vladzcloudius (changes requested):

scripts/perftune.py
tchaikov: thank you Benny and Vlad for pointing this out. indeed, it should follow the patch description. just wanted to explain why it is right, not in the sense it matches with the description, but in the sense that it does the job:
```console
In [1]: import re

In [2]: line = 'hello"   '

In [3]: re.sub("\"\\s*$", "",  line)
Out[3]: 'hello'

In [4]: re.sub(r'"\s*$', "",  line)
Out[4]: 'hello'
```

will use `r'"\s*$'`, as single quotes makes the regex much simpler to read from developer's point of view.

scripts/perftune.py
tchaikov: thank you, Vlad. after a second thought, i went further by replacing `"--banirq"` with a variable.
with my change, it looks like
```py
        opt = f"--banirq={irq}"
        patt_str = rf"{opt}\Z|{opt}\s"
        if not re.search(patt_str, new_options):
            new_options += f" {opt}"
```
does it look better now?

tchaikov:
v2:

- use r-string to "cut the last `"`"
- do not escape "`=`" and "`-`" when using raw string.
- use f-string to reduce the duplication.

bhalevy (approved):
LGTM

scripts/perftune.py
vladzcloudius: > just wanted to explain why it is right, not in the sense that it matches with the description, but in the sense that it does the job

That's strange but I confirm that your initial version gives the same match as the original (before the patch) one.
That's very weird because I was expecting that "\\s" regex would match a two char "\s" string (backslash + 's'). 

My apologies and I'm taking my "This change looks wrong in general" back.

scripts/perftune.py
vladzcloudius: It definitely does!
I intentionally didn't suggest you using f-strings in my initial suggestion because there rest of file uses the old-school formatting all other the place.  But if you are up to using them - I won't mind.

scripts/perftune.py
```diff
@@ -171,21 +171,22 @@ def restart_irqbalance(banned_irqs):
...
-        new_options = re.sub("\"\s*$", "", opt_lines[0].rstrip())
+        new_options = re.sub(r'"\s*$', "", opt_lines[0].rstrip())
         opt_lines = opt_lines[0].strip()
     else:
         raise Exception("Invalid format in {}: more than one lines with {} key".format(config_file, options_key))
 
     for irq in banned_irqs_list:
         # prevent duplicate "ban" entries for the same IRQ
-        patt_str = "\-\-banirq\={}\Z|\-\-banirq\={}\s".format(irq, irq)
+        opt = "--banirq"
```
vladzcloudius, line 186: Why not to do what you suggested: `opt = f"--banirq={irq}"` ?

scripts/perftune.py
```diff
@@ -171,21 +171,22 @@ def restart_irqbalance(banned_irqs):
...
+        patt_str = rf"{opt}={irq}\Z|{opt}={irq}\s"
```
vladzcloudius, line 187: ...and then here it would be `patt_str = rf"{opt}\Z|{opt}\s"`

scripts/perftune.py
```diff
@@ -171,21 +171,22 @@ def restart_irqbalance(banned_irqs):
...
         if not re.search(patt_str, new_options):
-            new_options += " --banirq={}".format(irq)
+            new_options += f" {opt}={irq}"
```
vladzcloudius, line 189: ...and here - `new_options += f" {opt}"`

vladzcloudius (changes requested):

scripts/perftune.py
tchaikov: `\s` is special in at least python's re, i think. i am quoting CPython's document at https://docs.python.org/3/library/re.html#index-32

> \s
> For Unicode (str) patterns:
> Matches Unicode whitespace characters (which includes [ \t\n\r\f\v], and also many other characters, for example the non-breaking spaces mandated by typography rules in many languages).
> 
> Matches [ \t\n\r\f\v] if the [ASCII](https://docs.python.org/3/library/re.html#re.ASCII) flag is used.
> 
> For 8-bit (bytes) patterns:
> Matches characters considered whitespace in the ASCII character set; this is equivalent to [ \t\n\r\f\v].

scripts/perftune.py
tchaikov: thank you Vlad. i forgot to push before leaving for my dentist appointment! corrected.

scripts/perftune.py
tchaikov: i see. there are indeed a lot of places in this script, where  we could use f-string for better readability. yeah, i defend my changes with "readability" quite often =) but i hold my hands back when they tried to do the refactor. as, it might be better to change it piecemeal for reducing the risk of change, and reducing review's workload for reviewing each PR.

tchaikov:
v2:

- drop the change included in #2160
- apply the `opt = f"--banirq={irq}"` change, and use f-string to interpolate this variable in regex so we don't repeat this command line option in 3 places.

vladzcloudius (approved):
Looks good. Thanks!

tchaikov:

scripts/perftune.py
```diff
@@ -171,21 +171,22 @@ def restart_irqbalance(banned_irqs):
...
+        new_options = re.sub(r'"\s*$', "", opt_lines[0].rstrip())
         opt_lines = opt_lines[0].strip()
     else:
         raise Exception("Invalid format in {}: more than one lines with {} key".format(config_file, options_key))
 
     for irq in banned_irqs_list:
         # prevent duplicate "ban" entries for the same IRQ
-        patt_str = "\-\-banirq\={}\Z|\-\-banirq\={}\s".format(irq, irq)
+        opt = f"--banirq={irq}"
+        patt_str = rf"{opt}\Z|{opt}\s"
```
avikivity, line 187: Perhaps better written as {opt}(\z|\s). Not suggesting to change.

PR #1832, opened 2023-09-17
build: build with dpdk v23.07

### LINKED ISSUE #1831: update dpdk to newer version
The update to a newer dpdk was reverted in https://github.com/scylladb/seastar/commit/225bee3e38b2bbeb3b7507e8ac1f2b946c6b34ca

Alot of distros have a newer version: https://repology.org/project/dpdk/versions

It looks like dpdk 19 was EOL December 2022 and there is no planned next version https://core.dpdk.org/roadmap/

tchaikov:
the previous version of change was reverted by 225bee3e38b2bbeb3b7507e8ac1f2b946c6b34ca.

back then, the reverted change didn't include dpdk *into* `libseastar.a`, it just exposed the static libraries of dpdk via a public interface library. and this broke the workflow of

1. build seastar with dpdk enabled
2. install it or redistribute it
3. build a seastar application with the installed / redistributed seastar

so, in order to address this problem, in this change, dpdk libraries are now collected into a single relocatable .o object file as before. and this object file is exposed as a library target to ease the usage on the consumer side. the corresponding dpdk is updated to the latest stable version, v23.07, with two patches (explained below) applied, see https://github.com/tchaikov/dpdk/tree/wip-seastar-v3

* pick up dpdk v23.07 with following patches applied
  - ixgbevf: ixgbevf_dev_info_get(): return default tx_rs_thresh=1
  - build: add meson options of max_memseg_lists
  the patches above are not upstreamed yet.
  before this change, the dpdk submodule was pointing to v19.05 with
  following patches applied:
  - d42481c4 ixgbe_pmd: forbid tx_rs_thresh above 1 for all NICs but 82598
    this change has been fixed by 9f5d2352fe. it is included by v21.08
  - 6a70dd05 ixgbevf: ixgbevf_dev_info_get(): return default tx_rs_thresh=1
    this change was cherry-picked on top of v21.08 in the new dpdk
    commit
  - 7c29bbc8 net: i40e: add VLAN tag size to RXMAX
    ditto
  - 71db32d7 mk: avoid combining -r and -export-dynamic linker options
    it has been included as 5a352a3a98 in v21.08
  - 0f486994 pmdinfogen: fix build with gcc 11
    it has been included as 5a75cc3c88 in v21.08
* build dpdk using meson and ninja
* override the options using meson options
* disable the build of drivers using meson options
* update the libraries name to be in sync with the latest dpdk
* dpdk
  - s/ether_addr/rte_ether_addr/
    ether_addr was renamed to ether_addr in 6d13ea8e8e49ab957deae2bba5ecf4a4bfe747d1,
    see https://github.com/DPDK/dpdk/commit/6d13ea8e8e49ab957deae2bba5ecf4a4bfe747d1
    the first release including this change was v19.08. so let's
    update accordingly
  - s/ether_hdr/rte_ether_hdr/
    the same applies to "ether_hdr"
  - do not set TOEPLITZ as RSS hash
    this change pratically reverts https://github.com/scylladb/seastar/commit/eec89a57a43c070df9d3e29fa649df7237be9f39,
    because DPDK v21.08 includes the fix of
    ef4c16fd9148215897abadf8e8a965488c82ba03 and c725221d09113bde89faa1e3c468e805fb335939
    which set the RSS hash function properly for i40e.
    so let's drop this fix to set TOEPLITZ as RSS hash for
    unconditionally.
  - s/rte_vfio_container_dma_map()/rte_vfio_dma_map()/
    the latter was deprecated in v19.11. see
    dpdk/doc/guides/rel_notes/release_19_11.rst
  - replace PKT_TX_* with RTE_MBUF_F_*, the former was deprecated.
  - replace RTE_* with RTE_ETH_*, the former was removed in 22.11
  - replace DEV_TX_* with RTE_ETH_TX_*, the former was removed in 22.11
* install-dependencies: add meson and python3-pyelftoolsto the
  packages list. because we need them to build newer DPDK.

Fixes #1831


a16bitsysop:
It finds the installed dpdk 23.07 fine on alpine, but I get the following warning which is treated as an error by the default flags

```
include/seastar/core/dpdk_rte.hh:34:2: error: #warning "CONFIG_RTE_MBUF_REFCNT_ATOMIC should be disabled in DPDK's " "config/common_linuxapp" [-Werror=cpp]
   34 | #warning "CONFIG_RTE_MBUF_REFCNT_ATOMIC should be disabled in DPDK's " \
      |  ^~~~~~~
cc1plus: all warnings being treated as errors
```

tchaikov:
@a16bitsysop yeah, that's expected. the warning is not a bug. for better performance on seastar, `CONFIG_RTE_MBUF_REFCNT_ATOMIC` should be disabled. but if the dpdk installed in system is compiled for general purpose dpdk applications, we have to enable it. in that case, you can always disable the warning by, for instance, passing `-DCMAKE_CXX_FLAGS=-Wno-cpp` to `cmake` .

tchaikov:
> It finds the installed dpdk 23.07 fine on alpine, but I get the following warning which is treated as an error by the default flags
> 
> ```
> include/seastar/core/dpdk_rte.hh:34:2: error: #warning "CONFIG_RTE_MBUF_REFCNT_ATOMIC should be disabled in DPDK's " "config/common_linuxapp" [-Werror=cpp]
>    34 | #warning "CONFIG_RTE_MBUF_REFCNT_ATOMIC should be disabled in DPDK's " \
>       |  ^~~~~~~
> cc1plus: all warnings being treated as errors
> ```

or, we can make it a configure-time warning. but that's not in the scope of this change. if you think we should improve this in seastar, could you please file an issue for it?

a16bitsysop:
okay, I added the flag and compiled locally and it looked like it failed all the tests.

so I tried in the alpine CI and it passed, but doesn't look like it is linked to any dpdk libraries on x86_64:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1116102#L1267 aarch64 is the same

full pipeline:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1116102

the test for dpdk on ppc64le fails, so I guess one of the modules it needs isn't available for ppc64le: https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1116105

tchaikov:
> okay, I added the flag and compiled locally and it looked like it failed all the tests.

without the error messages or the steps to reproduce the problem, i can hardly figure out what's going wrong =(

> 
> so I tried in the alpine CI and it passed, but doesn't look like it is linked to any dpdk libraries on x86_64: https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1116102#L1267 aarch64 is the same
> 
> full pipeline: https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1116102
> 
> the test for dpdk on ppc64le fails, so I guess one of the modules it needs isn't available for ppc64le: https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1116105

i see. i was focused on building with the static libraries. as that's the preferred way to build and use Seastar. i just revised the build script to cater the needs of building Seastar with the shared dpdk libraries shipped by the system.

a16bitsysop:
Thanks, will try again.  I couldn't see the errors either as my scrollback wasn't working somehow.  Will try new version.

dpdk_config
```diff
@@ -1,24 +0,0 @@
...
-CONFIG_RTE_LIBRTE_DISTRIBUTOR=n
-CONFIG_RTE_LIBRTE_PMD_CRYPTO_SCHEDULER=n
-CONFIG_RTE_LIBRTE_REORDER=n
-CONFIG_RTE_LIBRTE_PORT=n
-CONFIG_RTE_LIBRTE_TABLE=n
-CONFIG_RTE_LIBRTE_PIPELINE=n
-CONFIG_RTE_LIBRTE_FLOW_CLASSIFY=n
-CONFIG_RTE_LIBRTE_BPF=n
-CONFIG_RTE_LIBRTE_EFD=n
-CONFIG_RTE_LIBRTE_MEMBER=n
```
avikivity, line 24: Looks like this was truncated, should it be deleted instead?

src/core/reactor.cc
```diff
@@ -4516,7 +4516,7 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
 #ifdef SEASTAR_HAVE_DPDK
     if (_using_dpdk) {
         auto it = _thread_loops.begin();
-        RTE_LCORE_FOREACH_SLAVE(i) {
+        RTE_LCORE_FOREACH_WORKER(i) {
```
avikivity, line 4519: Let's defer this patch until the thing is merged.

dpdk_config
tchaikov: it is deleted. see https://github.com/tchaikov/seastar/tree/wip-dpdk-v3/dpdk_config

src/core/reactor.cc
tchaikov: sure. dropped.

tchaikov:
v2:

- dropped the commit which s/RTE_LCORE_FOREACH_SLAVE/RTE_LCORE_FOREACH_WORKER/

avikivity:
I pushed the dpdk branch to https://github.com/scylladb/dpdk/commits/local-23.07

avikivity:
btw it's a good simplification for us, I disliked that wart.

tchaikov:
> Please test it with scylladb in release mode to make sure there are no loose ends.
> 
> In particular, there's risk that architecture selection has changed. In older versions we converted uarch names like "westmere" to Intel acronyms like "wsm". Is that still there?

yeah, it's now changed in two ways:

1. the option is configured using `meson setup -Dcpu_instruction_set=${Seastar_DPDK_MACHINE}`
2. the value of `cpu_instruction_set` is passed to `-march` on amd64 and aarch64.

i dropped that translation in the latest revision.

avikivity:
Ah, I see we used the gcc names in configure.py/cmake, not the dpdk name. This keeps them isolated.

We can drop Seastar_DPDK_MACHINE AFAICT.

dpdk_config
avikivity: ok, don't know why I thought otherwise.

avikivity:
Looks good, let me know when you've tested with ScyllaDB and I'll commit.

a16bitsysop:
I split out the static libraries into a sub package for alpine and without static libraries installed it links to:

```
	so:librte_eal.so.23
	so:librte_ethdev.so.23
	so:librte_mbuf.so.23
	so:librte_mempool.so.23
```

If static libraries are installed it must link to those instead as the dependency checker doesn't find them, no really a massive issue I think

tchaikov:
> If static libraries are installed it must link to those instead 

@a16bitsysop i am not sure i completely understand what you mean. do you mean that, if the dpdk static libraries are installed, seastar will link to the static libraries? if that's the case, i think it's the expected behavior. 

> as the dependency checker doesn't find them, no really a massive issue I think

IIUC, the dependency checker only checks the runtime dependencies, no? if it is a compile time behavior. i think what we can do is to make sure the packaging recipe has all ingredients included.

tchaikov:
> Ah, I see we used the gcc names in configure.py/cmake, not the dpdk name. This keeps them isolated.
> 
> We can drop Seastar_DPDK_MACHINE AFAICT.

`Seastar_DPDK_MACHINE` is a convenient way allowing us to explicitly specify the `-march` for DPDK. without which, DPDK will go with "native", this could be wrong if the building host uses a different architecture than that of the target machine, so it could pick an (extension) instruction set not supported by the target machine, even we don't do cross-compiling. or, the "native" could end with a much smaller subset of instruction set, so we won't be able to take the full advantage of the target machine. so, probably the safe approach is to keep using `Seastar_DPDK_MACHINE`, we could use a better name though. but that's another story.

tchaikov:
v3:

- merge s/RTE_LCORE_FOREACH_SLAVE/RTE_LCORE_FOREACH_WORKER/ change into the first commit. as the `RTE_LCORE_FOREACH_SLAVE` macro has been removed in DPDK.
- include `-Wno-volatile` in cflags only if it is supported. as clang does not support this compiling option.

tchaikov:
@avikivity hi Avi, tested by building both debug and release builds. i am pasting the info of the release build:

```console
$ ninja build/release/scylla
[1/3] NINJA build/release/seastar/libseastar.a
ninja: Entering directory `build/release/seastar'
ninja: no work to do.
[2/3] NINJA build/release/seastar/libseastar_testing.a
ninja: Entering directory `build/release/seastar'
ninja: no work to do.
$ grep dpdk build.ninja | head -n1
configure_args = --mode release --enable-dpdk --compiler clang++ --c-compiler clang --with scylla
$ build/release/scylla --version
5.4.0~dev-0.20230919.55308781a66a
$ find . -name 'dpdk.o'
./build/release/seastar/dpdk.o
./build/debug/seastar/dpdk.o
$ nm build/release/seastar/libseastar.a|grep rte_hexdump
0000000000018710 T rte_hexdump
$ nm -D build/debug/seastar/libseastar.so|grep rte_hexdump
00000000040f2540 T rte_hexdump
```

a16bitsysop:
> > If static libraries are installed it must link to those instead
> 
> @a16bitsysop i am not sure i completely understand what you mean. do you mean that, if the dpdk static libraries are installed, seastar will link to the static libraries? if that's the case, i think it's the expected behavior.
> 

Yes thats it, seastar links to the static dpdk libraries if they are installed.

> > as the dependency checker doesn't find them, no really a massive issue I think
> 
> IIUC, the dependency checker only checks the runtime dependencies, no? if it is a compile time behavior. i think what we can do is to make sure the packaging recipe has all ingredients included.

Yes that is correct as well.

src/core/reactor.cc
tchaikov: turns out i have to include it to get the change build with dpdk v23.07. so merged it into the first commit.

tchaikov:
Thank you Duncan, just wanted to make sure there was nothing missing in this change.

avikivity:
Thanks a lot for this work.

PR #1833, opened 2023-09-17
app-template: add support to specify seastar.conf

syuu1228:
On Scylla, we want to use seastar.conf to store configured parameters from setup scripts, but we do not want to place it under ~/.config/seastar, we want to use /etc instead.
Let's add SEASTAR_CONF_FILE environment variable to specify the file path.

Related scylladb/scylladb#14674

### DIFF
--- a/src/core/app-template.cc
+++ b/src/core/app-template.cc
@@ -117,6 +117,17 @@ const app_template::seastar_options& app_template::options() const {
 
 app_template::configuration_reader app_template::get_default_configuration_reader() {
     return [this] (bpo::variables_map& configuration) {
+        auto seastar_conf = std::getenv("SEASTAR_CONF");
+        if (seastar_conf) {
+            std::ifstream ifs {std::string(seastar_conf)};
+            if (ifs) {
+                bpo::store(bpo::parse_config_file(ifs, _opts_conf_file), configuration);
+            } else {
+                seastar_logger.error("Failed to open {}", seastar_conf);
+                _exit(1);
+            }
+            return;
+        }
         auto home = std::getenv("HOME");
         if (home) {
             std::ifstream ifs(std::string(home) + "/.config/seastar/seastar.conf");


src/core/app-template.cc
```diff
@@ -117,6 +117,17 @@ const app_template::seastar_options& app_template::options() const {
 
 app_template::configuration_reader app_template::get_default_configuration_reader() {
     return [this] (bpo::variables_map& configuration) {
+        auto seastar_conf = std::getenv("SEASTAR_CONF");
```
tchaikov, line 120: how about changing scylla instead:

```c++
app.set_configuration_reader([cfg, &app](bpo::variables_map& configuration) {
    auto seastar_conf = cfg->get_conf_sub("seastar.conf");
    if (std::ifstream ifs(seastar_conf); ifs) {
        bpo::store(bpo::parse_config_file(ifs, app.get_conf_file_options_description()), configuration);
    }
});
```

in `scylla_main()`. probably we need use the facilities and mechinaries provided by seastar, instead of arming it with them to implement the policies.

xemul:
> Having Seastar be responsible for reading options was a design mistake IMO. The app should be responsible for options, and of it needs to send some to Seastar that it should do so via app_template::config. Then, the names of the parameters, the format and location of the configuration file, and responsibility for reloading are all up to the application.
> 
> So, we need to choose. Do we continue to improve option processing, or do we deprecate them in favor of moving everything to the application?

https://github.com/scylladb/seastar/pull/1615#issuecomment-1653089501

syuu1228:
> how about changing scylla instead:
> 
> ```c++
> app.set_configuration_reader([cfg, &app](bpo::variables_map& configuration) {
>     auto seastar_conf = cfg->get_conf_sub("seastar.conf");
>     if (std::ifstream ifs(seastar_conf); ifs) {
>         bpo::store(bpo::parse_config_file(ifs, app.get_conf_file_options_description()), configuration);
>     }
> });
> ```
> 
> in `scylla_main()`. probably we could instead use the facilities and mechinaries provided by seastar, instead of arming it with them to implement the desired policies.

This worked, thanks.
So I do not need to merge this PR, closing.

PR #1835, opened 2023-09-18
scripts: addr2line: use yield instead defining a class

tchaikov:
it is kind of overkill to define a class to implement an iterator. less just use `yield` here. simpler this way, and less indent.

as we expect the resolver prints all decoded backtrace in a bulk, we still materialize the input backtrace using `list()`, so the input and output are not interleaved.

### DIFF
--- a/scripts/seastar-addr2line
+++ b/scripts/seastar-addr2line
@@ -32,28 +32,24 @@ class StdinBacktraceIterator(object):
     Enter twice. Altough reading char-by-char is slow this won't be a
     problem here as backtraces shouldn't be huge.
     """
-    def __iter__(self):
-        linefeeds = 0
-        lines = []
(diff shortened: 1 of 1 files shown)


bhalevy:
Looks good to me.
Any morivation besides cleaning up the code?
(If so, it might not the best use of our time...)

bhalevy (approved):

tchaikov:
> Looks good to me.
> Any morivation besides cleaning up the code?

no, it's just a cleanup.

> (If so, it might not the best use of our time...)

let's discuss this offline.

tchaikov:
this introduces a regression, it should have included a part of #1834 in this change. please create a follow-up change ASAP.

tchaikov:
#1846

PR #1837, opened 2023-09-19
Add alternate method for file system detection

### LINKED ISSUE #251: reactor::file_system_at() does not able to detect ext3/ext4 correctly
Since EXT2_SUPER_MAGIC == EXT3_SUPER_MAGIC == EXT4_SUPER_MAGIC, reactor::file_system_at() always returns fs_type::ext2.
Perhaps we can use getmntent() instead of statfs().

travisdowns:
This attempts to resolve #251 by adding an alternative detection method which parses `/proc/mounts` rather than relying on superblock magic checks (as suggested in 251).

I did not modify the existing method, since I'm not entirely sure how robust this method is: rather it is added as an alternative method with the same API.

### DIFF
--- a/include/seastar/core/file-types.hh
+++ b/include/seastar/core/file-types.hh
@@ -88,6 +88,9 @@ enum class fs_type {
     btrfs,
     hfs,
     tmpfs,
+
+    first_type = other,
+    last_type = tmpfs   // sentinels for iteration
 };
 
 // Access flags for files/directories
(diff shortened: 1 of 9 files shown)


travisdowns:
@xemul given the discussion in https://github.com/scylladb/seastar/issues/251, I am guessing this PR has a low likelihood to be accepting so I'm going to close it now.

If I guessed wrong feel free to re-open.

PR #1839, opened 2023-09-19
reactor: don't interact with the memory.cc when not using seastar the allocator

denesb:
In particular:
* don't attempt to set reclaim hooks
* don't create poller for draining cross-cpu free list

Either of these will cause a SIGSEGV when seastar is configured to use the standard memory allocator.

This patch is a follow-up to 2cec417426621b604ee0278f71d1e634254cb150. For some reason the crashes only reproduce when --smp > 1. That is why it remained hidden so long

Fixes: scylladb/scylladb#12019


src/core/reactor.cc
```diff
@@ -4369,6 +4373,7 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     reactor_config reactor_cfg;
     reactor_cfg.auto_handle_sigint_sigterm = reactor_opts._auto_handle_sigint_sigterm;
     reactor_cfg.max_networking_aio_io_control_blocks = adjust_max_networking_aio_io_control_blocks(reactor_opts.max_networking_io_control_blocks.get_value());
+    reactor_cfg.using_seastar_allocator = (smp_opts.memory_allocator == memory_allocator::seastar);
```
avikivity, line 4376: In debug mode we don't use the seastar allocator, does smp_opts.memory_allocator reflect that?

avikivity: ok. but all CI checks failed.


I guess the cause is, once you call anything in memory.cc, it initializes thread-local variables and breaks things.

PR #1840, opened 2023-09-19
Added ability to configure different credentials per HTTP listeners

mmaslankaprv:
For some deployments it may be required to configure different set of
credentials for different http server listeners. Added ability to
set/override default http server credentials with listener specific set
of credentials.

### DIFF
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -249,7 +249,9 @@ public:
     future<> stop();
     future<> set_routes(std::function<void(routes& r)> fun);
     future<> listen(socket_address addr);
+    future<> listen(socket_address addr, shared_ptr<seastar::tls::server_credentials> credentials);
     future<> listen(socket_address addr, listen_options lo);
+    future<> listen(socket_address addr, listen_options lo, shared_ptr<seastar::tls::server_credentials> credentials);
     distributed<http_server>& server();
 };
 SEASTAR_MODULE_EXPORT_END
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -86,27 +86,30 @@ class connection : public boost::intrusive::list_base_hook<> {
     // null element marks eof
     queue<std::unique_ptr<http::reply>> _replies { 10 };
     bool _done = false;
+    const bool _tls;
 public:
-    [[deprecated("use connection(http_server&, connected_socket&&)")]]
-    connection(http_server& server, connected_socket&& fd,
-            socket_address) : connection(server, std::move(fd)) {}
-    connection(http_server& server, connected_socket&& fd)
+    [[deprecated("use connection(http_server&, connected_socket&&, bool tls)")]]
+    connection(http_server& server, connected_socket&& fd, socket_address, bool tls) 
+            : connection(server, std::move(fd), tls) {}
+    connection(http_server& server, connected_socket&& fd, bool tls)
             : _server(server)
             , _fd(std::move(fd))
             , _read_buf(_fd.input())
             , _write_buf(_fd.output())
             , _client_addr(_fd.remote_address())
-            , _server_addr(_fd.local_address()) {
+            , _server_addr(_fd.local_address())
+            , _tls(tls) {
         on_new_connection();
     }
     connection(http_server& server, connected_socket&& fd,
-            socket_address client_addr, socket_address server_addr)
+            socket_address client_addr, socket_address server_addr, bool tls)
             : _server(server)
             , _fd(std::move(fd))
             , _read_buf(_fd.input())
             , _write_buf(_fd.output())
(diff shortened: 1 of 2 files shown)


include/seastar/http/request.hh
```diff
@@ -87,6 +87,7 @@ struct request {
     std::unordered_map<sstring, sstring> chunk_extensions;
     sstring protocol_name = "http";
     noncopyable_function<future<>(output_stream<char>&&)> body_writer; // for client
+    int listener_idx;
```
xemul, line 90: @mmaslankaprv , there's a PR #1851 that adds remote and (!) local address to http::request. Would the knowledge of local address be a good replacement for the listener_idx thing?

src/http/httpd.cc
```diff
@@ -370,14 +370,27 @@ void http_server::set_content_streaming(bool b) {
...
+        _listeners.push_back(seastar::tls::listen(listener_credentials, addr, lo));
     } else {
         _listeners.push_back(seastar::listen(addr, lo));
     }
-    return do_accepts(_listeners.size() - 1);
+    return do_accepts(_listeners.size() - 1, listener_credentials != nullptr);
 }
+
+future<> http_server::listen(socket_address addr, listen_options lo) {
+    return listen(addr, lo, _credentials);
```
xemul, line 385: It seems to me that `set_tls_credentials()` should be deprecated in favor of the new `listen(..., creds)` API

include/seastar/http/httpd.hh
```diff
@@ -166,11 +167,13 @@ public:
...
     void set_content_streaming(bool b);
 
+    future<> listen(socket_address addr, shared_ptr<seastar::tls::server_credentials> credentials);
+    future<> listen(socket_address addr, listen_options lo, shared_ptr<seastar::tls::server_credentials> credentials);
     future<> listen(socket_address addr, listen_options lo);
     future<> listen(socket_address addr);
     future<> stop();
 
-    future<> do_accepts(int which);
+    future<> do_accepts(int which, bool tls = false);
```
xemul, line 177: The default value seems incorrect. It should be `tls = _credentials != nullptr` instead, no?

include/seastar/http/request.hh
mmaslankaprv: yes, it would be even better, i will defer the update till #1851 is merged, thank you for this suggestion

xemul: @mmaslankaprv #1851 is merged (FYI)

src/http/httpd.cc
mmaslankaprv: done

include/seastar/http/httpd.hh
mmaslankaprv: done

src/http/httpd.cc
```diff
@@ -389,14 +389,27 @@ void http_server::set_content_streaming(bool b) {
...
-    if (_credentials) {
-        _listeners.push_back(seastar::tls::listen(_credentials, addr, lo));
+future<> http_server::listen(socket_address addr, listen_options lo,
+            shared_ptr<seastar::tls::server_credentials> listener_credentials) {
+    if (listener_credentials) {
+        _listeners.push_back(seastar::tls::listen(listener_credentials, addr, lo));
     } else {
         _listeners.push_back(seastar::listen(addr, lo));
     }
     return do_accepts(_listeners.size() - 1);
```
xemul, line 399: Is that correct? If I call `lister(addr, lo, creds)` it will add tls-enabled listener and will call `do_accepts(nr)`. The latter, in turn, would call `do_accepts(nr, _credentials != nullptr)` which is wrong -- `_credentials` shouldn't be used in this particular case

mmaslankaprv: you are right, sorry for this, it should actually decide based on the present of current listener credentials, fixed

include/seastar/http/httpd.hh
```diff
@@ -86,27 +86,30 @@ class connection : public boost::intrusive::list_base_hook<> {
     // null element marks eof
     queue<std::unique_ptr<http::reply>> _replies { 10 };
     bool _done = false;
+    bool _tls;
```
xemul, line 89: Please, mark as const

include/seastar/http/httpd.hh
```diff
@@ -179,6 +182,7 @@ public:
         }).get();
      *
      */
+    [[deprecated("use listen(socket_address addr, shared_ptr<seastar::tls::server_credentials> credentials)")]]
     void set_tls_credentials(shared_ptr<seastar::tls::server_credentials> credentials);
```
xemul, line 186: The `shared_ptr<seastar::tls::server_credentials>` appears a lot after this PR, so it deserved a `using server_credentials_ptr` alias in the class scope

xemul (approved):
LGTM, please fix two cosmetics

PR #1841, opened 2023-09-19
Proactively close non-keepalive http connections

xemul:
Some (spoiler: AWS S3) http servers may send back the `Connection: close` header indicating that its going to close the connection after sending back the response. Not handling it is problematic for http client due to batch-flushes being explicitly used.

After client handles such a response, it will put back the connection into the pool. In case there's another request to be served, the same connection will be taken back and the request will be sent. However, write()-ing the request into the connection and even flush()-ing it won't report back _any_ exception because the batch-flush is turned on on socket by default. So the fact that the server had already closed the connection will only be found out at the attempt to read back and parse the response. Although in this case the client must re-connect and re-send the request, that's too late already. Even if the request object itself can be preserved, there's no guarantee that the body writer is able to "restart" and write the body again.

So the best that can be done here is to note that there's the `Connection: close` header in the response and not caching the connection in pool after it. This gracefully handles the typical scenario when AWS S3 server decides to stop keep-alive-ing a connection.

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -76,6 +76,10 @@ class connection : public enable_shared_from_this<connection> {
     hook_t _hook;
     future<> _closed;
     internal::client_ref _ref;
+    // Client sends HTTP-1.1 version and assumes the server is 1.1-compatible
+    // too and thus the connection will be persistent by default. If the server
+    // responds with older version, this flag will be dropped (see recv_reply())
+    bool _persistent = true;
 
 public:
     /**
@@ -175,7 +179,7 @@ class client {
     using connection_ptr = seastar::shared_ptr<connection>;
 
     future<connection_ptr> get_connection();
-    future<> put_connection(connection_ptr con, bool can_cache);
+    future<> put_connection(connection_ptr con);
     future<> shrink_connections();
 
     template <typename Fn>
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -125,7 +125,7 @@ future<connection::reply_ptr> connection::recv_reply() {
     http_response_parser parser;
     return do_with(std::move(parser), [this] (auto& parser) {
         parser.init();
-        return _read_buf.consume(parser).then([&parser] {
+        return _read_buf.consume(parser).then([this, &parser] {
             if (parser.eof()) {
                 throw std::runtime_error("Invalid response");
             }
@@ -133,6 +133,9 @@ future<connection::reply_ptr> connection::recv_reply() {
             auto resp = parser.get_parsed_response();
             sstring length_header = resp->get_header("Content-Length");
             resp->content_length = strtol(length_header.c_str(), nullptr, 10);
+            if (resp->_version != "1.1") {
+                _persistent = false;
+            }
             return make_ready_future<reply_ptr>(std::move(resp));
         });
     });
@@ -245,8 +248,8 @@ future<client::connection_ptr> client::get_connection() {
     });
 }
 
-future<> client::put_connection(connection_ptr con, bool can_cache) {
-    if (can_cache && (_nr_connections <= _max_connections)) {
+future<> client::put_connection(connection_ptr con) {
+    if (con->_persistent && (_nr_connections <= _max_connections)) {
         http_log.trace("push http connection {} to pool", con->_fd.local_address());
         _pool.push_back(*con);
         _wait_con.signal();
@@ -290,10 +293,8 @@ template <typename Fn>
 SEASTAR_CONCEPT( requires std::invocable<Fn, connection&> )
 auto client::with_connection(Fn&& fn) {
     return get_connection().then([this, fn = std::move(fn)] (connection_ptr con) mutable {
-        return fn(*con).then_wrapped([this, con = std::move(con)] (auto f) mutable {
-            return put_connection(std::move(con), !f.failed()).then([f = std::move(f)] () mutable {
-                return std::move(f);
-            });
+        return fn(*con).finally([this, con = std::move(con)] () mutable {
+            return put_connection(std::move(con));
         });
     });
 }
@@ -316,6 +317,9 @@ future<> client::make_request(request req, reply_handler handle, reply::status_t
             }
 
             return handle(rep, con.in(rep)).finally([reply = std::move(reply)] {});
+        }).handle_exception([&con] (auto ex) mutable {
+            con._persistent = false;
+            return make_exception_future<>(std::move(ex));
         });
     });
 }
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -133,7 +133,7 @@ future<connection::reply_ptr> connection::recv_reply() {
             auto resp = parser.get_parsed_response();
             sstring length_header = resp->get_header("Content-Length");
             resp->content_length = strtol(length_header.c_str(), nullptr, 10);
-            if (resp->_version != "1.1") {
+            if ((resp->_version != "1.1") || seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "close")) {
                 _persistent = false;
             }
             return make_ready_future<reply_ptr>(std::move(resp));


src/http/client.cc
```diff
@@ -125,14 +125,15 @@ future<connection::reply_ptr> connection::recv_reply() {
...
-        return _read_buf.consume(parser).then([&parser] {
+        return _read_buf.consume(parser).then([this, &parser] {
             if (parser.eof()) {
                 throw std::runtime_error("Invalid response");
             }
 
             auto resp = parser.get_parsed_response();
             sstring length_header = resp->get_header("Content-Length");
             resp->content_length = strtol(length_header.c_str(), nullptr, 10);
+            _keep &= resp->get_header("Connection") != "close";
```
nyh, line 136: Please compare the string in a case-insensitive manner. According to the RFC - see https://www.rfc-editor.org/rfc/rfc7230#section-6.1:

> Connection options are case-insensitive.

nyh (changes requested):
Look good except one comparison needs to be case-insensitive.
There's another issue to consider, which I'm too lazy to chase through the RFCs.

IIRC, if the client asks for HTTP 1.1 but the server responds with HTTP 1.0 (or 0.9...), the defaults are reversed: no "Connection" header in the response means default to close, and if you see a "Connection: Keep-Alive" in the response, then it means not to closed. For 100% correct code, you should handle this possibility too, although HTTP 1.0 servers are probably extinct by now...

src/http/client.cc
xemul: Fixed

xemul:
upd:
- use case-insensitive comparison to check "connection" header value

tchaikov:
> Look good except one comparison needs to be case-insensitive. There's another issue to consider, which I'm too lazy to chase through the RFCs.
> 
> IIRC, if the client asks for HTTP 1.1 but the server responds with HTTP 1.0 (or 0.9...), the defaults are reversed: no "Connection" header in the response means default to close, and if you see a "Connection: Keep-Alive" in the response, then it means not to closed. For 100% correct code, you should handle this possibility too, although HTTP 1.0 servers are probably extinct by now...

yeah, see https://www.rfc-editor.org/rfc/rfc2616#section-8.1.2.1. the default value of `_keep` in this changeset only applies to HTTP/1.1 .

include/seastar/http/client.hh
```diff
@@ -76,6 +76,7 @@ class connection : public enable_shared_from_this<connection> {
     hook_t _hook;
     future<> _closed;
     internal::client_ref _ref;
+    bool _keep = true;
```
tchaikov, line 79: how about using the terminology in the HTTP/1.1 standard, say, s/_keep/_persistent/?

xemul:
upd:
- renamed _keep -> _persistent
- added comment about http-1.0 and below incompatibility

tchaikov (approved):
lgtm

src/http/client.cc
```diff
@@ -125,14 +125,16 @@ future<connection::reply_ptr> connection::recv_reply() {
...
+        return _read_buf.consume(parser).then([this, &parser] {
             if (parser.eof()) {
                 throw std::runtime_error("Invalid response");
             }
 
             auto resp = parser.get_parsed_response();
             sstring length_header = resp->get_header("Content-Length");
             resp->content_length = strtol(length_header.c_str(), nullptr, 10);
+            // HTTP-1.1 and above only, see comment about this field for details
+            _persistent &= !seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "close");
```
nyh, line 137: "&=" is not the right operator, it's the bitwise operator :-(
I know some people claim it works (https://stackoverflow.com/questions/2488406/why-doesnt-c-have-or-for-booleans) but I think it's wrong.

nyh (changes requested):

include/seastar/http/client.hh
```diff
@@ -76,6 +76,10 @@ class connection : public enable_shared_from_this<connection> {
     hook_t _hook;
     future<> _closed;
     internal::client_ref _ref;
+    // Connections are persistent by default in HTTP-1.1 and above, so it's
+    // not 100% compatible with older versions. Since such servers are very
+    // likely not in use nowadays, keep it like this for simplicity
```
nyh, line 81: Since Seastar is a library, personally I would have liked to see 100%-accurate implementations for all cases (even obscure ones) rather than comments about not being 100%, especially since we're likely to ever review this area of the code again in the future, but I don't insist - I agree that the chance of finding an HTTP-1.0-only server in 2023 is 0%, and probably was 0% already a decade ago.

src/http/client.cc
nyh: By the way, "&=" also can't short-circuit - even if _persistent is false, it will still do (unnecessarily) the case_insensitive_cmp. I agree this has zero importance, but seeing "&=" still looks to me like a bug and any future reader will suspect it to be a bug, so let's not do it. Personally, I think the clearest thing to do is:
```
if (_presistent) {
    _persistent = !seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "close");
}
```

Or if you want to support HTTP 1.0, assuming we have _http11 from parsing the HTTP response:
```
 // If the connection is already not to be persistent (e.g., handler failed), leave it like that.
// But if it could be persisted, the "Connection:" header in the server's response may force
// us to cancel the persistence, i.e., close the connection.
if (_presistent) {
    if (_http11) {
        _persistent = !seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "close");
    } else {
         // HTTP 1.0 or 0.9
        _persistent = seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "keep-alive");
     }
}
```

xemul:
upd:
- always drop the `_persistent` bit if the server version is not 1.1 (client advertises 1.1) thus making it work with 1.0 (and 0.9) servers

include/seastar/http/client.hh
xemul: OK, updated it like that

src/http/client.cc
nyh: This is true - in HTTP 1.0, if the client didn't ask for keep-alive, there is no way the server will set it.
There's always the risk that in the future we'll change the part of the code that sends the request, and forgot to change the part that parses the response, but I agree - all of this is NOT important and nobody will ever have an HTTP 1.0 server anyway.

nyh (approved):

src/http/client.cc
```diff
@@ -125,14 +125,17 @@ future<connection::reply_ptr> connection::recv_reply() {
...
+        return _read_buf.consume(parser).then([this, &parser] {
             if (parser.eof()) {
                 throw std::runtime_error("Invalid response");
             }
 
             auto resp = parser.get_parsed_response();
             sstring length_header = resp->get_header("Content-Length");
             resp->content_length = strtol(length_header.c_str(), nullptr, 10);
+            if ((resp->_version != "1.1") || seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "close")) {
+                _persistent = false;
```
tchaikov, line 137: if we are aiming at a standard compliant implementation, then we need to check for "Keep-Alive" in header. in a gist, the check would look like:
```c++
if (resp->_version == "1.1") {
    // stay open unless explicitly closed
    if (seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "close")) {
        _persistent = false;
    }
} else {
    // for older HTTP, Keep-Alive must be explicitly negotiated
    _persistent = false;
    if (resp->get_header("Keep-Alive").empty()) {
        _persistent = true;
    } else if (seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "keep-alive")) {
        _persistent = true;
    }
}
````
see also https://www.rfc-editor.org/rfc/rfc2068#section-19.7.1.1

Comment (nyh) [reply src/http/client.cc:137]: Are you sure about that? I didn't remember this header so I went and look at documentation, and it appears that this "Keep-Alive" header is just used as an unofficial (not defined in any RFC)  way to add parameters like timeout to the keepalive.

You suggested that the "Keep-alive:" header had special meaning in HTTP 1.0. Can you find a reference to that?

I think it was "Connection: Keep-Alive", not "Keep-Alive: ...", which mattered in HTTP 1.0. But it was so long ago my memory is hazy ;-)

Comment (xemul) [reply src/http/client.cc:137]: For older servers the keep-alive is set by the server _only_ if the client does so. Seastar client doesn't so there's no chance it will see keep-alive in 1.0 server response

Comment (xemul) [reply src/http/client.cc:137]:
```
} else {
    // for older HTTP, Keep-Alive must be explicitly negotiated
    _persistent = false;
    if (resp->get_header("Keep-Alive").empty()) {
        _persistent = true;
    } else if (seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "keep-alive")) {
        _persistent = true;
    }
}
```

:point_up: if keep-alive header is missing, why _persistent = true? It's older server that is not persistent by default

nyh: > For older servers the keep-alive is set by the server _only_ if the client does so. Seastar client doesn't so there's no chance it will see keep-alive in 1.0 server response

This is true (we discussed this above). In theory it's best not to couple the sending part with the response-parsing part of the code, but it's fine that we do - if we ever want to change one to support keep-alive on http 1.0 (and we won't...) we'll just need to change the other.
In any case, none of this explains why @tchaikov asked to look at the "Keep-Alive" header. The so-called "Keep-Alive Connection token" mentioned in the RFC is the "Connection: Keep-Alive" header, not "Keep-Alive: ...".

tchaikov: > Are you sure about that? 

i referenced the related section of RFC. does it explain the rationales?

> For older servers the keep-alive is set by the server only if the client does so. Seastar client doesn't so there's no chance it will see keep-alive in 1.0 server response

ahh, i see. so it's determined by the (faulty) client side. but there is a, well hypothesized use case, where Seastar is used to implement a proxy which proxies the traffic from an ancient client. and we are supposed to cache the connections, as that's what proxy does. and now that we do proxy, we might want to use the same HTTP protocol as the ancient client does. in other words, we do what the client does.

tchaikov: > I think it was "Connection: Keep-Alive", not "Keep-Alive: ...", which mattered in HTTP 1.0. But it was so long ago my memory is hazy ;-)

okay, i think i misinterpreted the section.

> 
>    When the Keep-Alive connection-token has been transmitted with a
>    request or a response, a Keep-Alive header field MAY also be
>    included. The Keep-Alive header field takes the following form:

so we don't need to check it if we already check the one in "Connection".

nyh: > okay, i think i misinterpreted the following paragraph:
> 
> > When the Keep-Alive connection-token has been transmitted with a
> > request or a response, a Keep-Alive header field MAY also be
> > included. The Keep-Alive header field takes the following form:
> 
> so we don't need to check it if we already check the one in "Connection".

Right. If you look more carefully (I agree the wording is confusing), the phrase "Keep-Alive connection-token" refers to "Connection: Keep-Alive". It's not related to the "Keep-Alive:" header which is documented in https://datatracker.ietf.org/doc/html/rfc2068#section-19.7.1.1 and has no standard semantics so I think there's absolutely nothing we need to do with this header. (only if we were a proxy, we should have taken care to copy it, or something).

tchaikov: based on our discussion above, i revised the proposed check:
```c++
if (resp->_version == "1.1") {
    // stay open unless explicitly closed
    _persistent = !seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "close");
} else {
    // for older HTTP, Keep-Alive must be explicitly negotiated
    _persistent = seastar::internal::case_insensitive_cmp()(resp->get_header("Connection"), "keep-alive");
}
````

nyh: This was exactly my proposal above (https://github.com/scylladb/seastar/pull/1841#discussion_r1331404362).
But @xemul rightly noted that according to the protocol, the 1.0 server can never respond with "Connection: Keep-Alive" if the client didn't have the same header in its request, and since we know that we didn't (we're talking about the client code here), the server also can't. So in the 1.0 case, "_persistent = false" is correct in our case. At least until we change our code to send "Connection: Keep-Alive", which I'm pretty sure we never will because nobody cares about the performance on 1.0 servers (which don't exist any more).

PR #1843, opened 2023-09-20
http/exception: add formatter for httpd::base_exception

tchaikov:
to ease the debugging of http related failures.

### DIFF
--- a/include/seastar/http/exception.hh
+++ b/include/seastar/http/exception.hh
@@ -163,3 +163,12 @@ SEASTAR_MODULE_EXPORT_END
 }
 
 }
+
+SEASTAR_MODULE_EXPORT
+template <>
+struct fmt::formatter<seastar::httpd::base_exception> {
+    constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+    auto format(const seastar::httpd::base_exception& e, fmt::format_context& ctx) const {
+        return fmt::format_to(ctx.out(), "{} ({})", e.what(), e.status());
+    }
+};


include/seastar/http/exception.hh
```diff
@@ -163,3 +163,12 @@ SEASTAR_MODULE_EXPORT_END
 }
 
 }
+
+SEASTAR_MODULE_EXPORT
+template <>
+struct fmt::formatter<seastar::httpd::base_exception> {
+    constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+    auto format(const seastar::httpd::base_exception& e, fmt::format_context& ctx) const {
+        return fmt::format_to(ctx.out(), "{} ({})", e.what(), e.status());
```
xemul, line 172: Aren't std::exception-s formattable by default with the e.what() printing?

tchaikov: it does not print the `e.status()`, that's what i am missing here.

xemul:
@tchaikov , regarding this formatter. This is what printed in logs when scylla's s3 client test fails:

```
[Exception] - seastar::httpd::unexpected_status_error: Unexpected reply status
 == [File] - test/boost/s3_test.cc
 == [Line] - 149
```

When an exception flies away far enough and stops being the httpd::base_exception, we still don't know the status. There's #1931 for that, but my question here is -- what's the point in this formatter if it only works when we _know_ that the exception is httpd one ?

tchaikov:
> @tchaikov , regarding this formatter. This is what printed in logs when scylla's s3 client test fails:
> 
> ```
> [Exception] - seastar::httpd::unexpected_status_error: Unexpected reply status
>  == [File] - test/boost/s3_test.cc
>  == [Line] - 149
> ```
> 
> When an exception flies away far enough and stops being the httpd::base_exception, we still don't know the status. There's #1931 for that, but my question here is -- what's the point in this formatter if it only works when we _know_ that the exception is httpd one ?

it's *handy*

1. in the PR which was later repurpose to print logging message in the tests. see https://github.com/scylladb/scylladb/pull/15477
2. in https://github.com/scylladb/seastar/blob/cba38d6f0e064b8f09d495aa0ea9260edb63f0a9/src/util/log.cc#L664-L670 if we add `seastar::httpd::unexpected_status_error` to the list. and just print it with `formatter`.  please note, i imagine that we can just print the exceptions with their formatters without calling their specific method for inspecting them individually.

in short, it's handy. but not a must have in the context, where a virtual method like `what()` is expected.

PR #1845, opened 2023-09-20
http/client: Add authorization callback

xemul:
Currently client assumes that the request being sent already has all the necessary headers set up including the authorization header(s). However, this can play bad joke with the user.

The thing is that before sending a request client may wait for a while until some connection (from pool) becomes available or creating a new connection. This may happen long enough to make the request's authorization header stale.

For example, S3 requires the request date and time to be included into the auth header. Despite the maximum time skew is set to be whole 15 minutes, even this can surprisingly be not enough. Scylla's S3 client uploads objects with 5Mb parts. In the worst (observed) case each part may require new connection, connect() may take more than a minute. The upload speed may also drop down to megabytes-per-second making things worse (a little). Said that, rare enough, but long queue of uploading parts may have its trailing requests' auth header time-out.

The solution is to authorize request right before it goes onto the wire.

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -161,6 +161,10 @@ public:
  */
 
 class client {
+public:
+    using authorizer = noncopyable_function<void(request&)>;
+
+private:
     friend class http::internal::client_ref;
     using connections_list_t = bi::list<connection, bi::member_hook<connection, typename connection::hook_t, &connection::_hook>, bi::constant_time_size<false>>;
     static constexpr unsigned default_max_connections = 100;
@@ -171,6 +175,7 @@ class client {
     unsigned long _total_new_connections = 0;
     condition_variable _wait_con;
     connections_list_t _pool;
+    authorizer _authorize;
 
     using connection_ptr = seastar::shared_ptr<connection>;
 
@@ -244,6 +249,15 @@ public:
      * \param nr -- the new limit on the number of connections
      */
     future<> set_maximum_connections(unsigned nr);
+
+    /**
+     * \brief Sets request authorizer
+     *
+     * The authorizer is called right before the client is about to send the request onto
+     * the wire. The callback is free to set up any additional headers, but changing any
+     * other request parameter may not have its effect and must not happen
+     */
+    void set_authorizer(authorizer);
 
     /**
      * \brief Closes the client
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -220,6 +220,7 @@ client::client(socket_address addr, shared_ptr<tls::certificate_credentials> cre
 client::client(std::unique_ptr<connection_factory> f, unsigned max_connections)
         : _new_connections(std::move(f))
         , _max_connections(max_connections)
+        , _authorize([] (request&) { /* do nothing */ })
 {
 }
 
@@ -286,6 +287,10 @@ future<> client::set_maximum_connections(unsigned nr) {
     return shrink_connections();
 }
 
+void client::set_authorizer(authorizer auth) {
+    _authorize = std::move(auth);
+}
+
 template <typename Fn>
 SEASTAR_CONCEPT( requires std::invocable<Fn, connection&> )
 auto client::with_connection(Fn&& fn) {
@@ -299,7 +304,8 @@ auto client::with_connection(Fn&& fn) {
 }
 
 future<> client::make_request(request req, reply_handler handle, reply::status_type expected) {
-    return with_connection([req = std::move(req), handle = std::move(handle), expected] (connection& con) mutable {
+    return with_connection([this, req = std::move(req), handle = std::move(handle), expected] (connection& con) mutable {
+        _authorize(req);
         return con.do_make_request(std::move(req)).then([&con, expected, handle = std::move(handle)] (connection::reply_ptr reply) mutable {
             auto& rep = *reply;
             if (rep._status != expected) {


include/seastar/http/client.hh
```diff
@@ -171,6 +175,7 @@ class client {
     unsigned long _total_new_connections = 0;
     condition_variable _wait_con;
     connections_list_t _pool;
+    authorizer _authorize;
```
tchaikov, line 178: nit, might want to initialize it like
```c++
authorizer _authorize{[] (request&) { /* do nothing */ }};
```

tchaikov (approved):
lgtm. just left a couple comments.

xemul:
upd:
- make set_authorizer's argument rvalue ref

nyh:
Hi @xemul there's something I don't understand about this API. We're doing this in the context of Seastar's HTTP client - not any specific user like s3. HTTP doesn't have the concept of an "authorizer", or authorization headers, so I think it's strange that our API will add such a concept.

Morever, I don't understand why this concept is needed. There's always a close connection between the headers and the body of the request - for example the headers might mention the time of the requests, cookies that need to go with the request, the size of the request, a signature of the entire request, and so on. I don't think it ever makes sense to generate headers and then one hour later - generate the body. So why is this even an issue? Why don't we generate the headers (all of them, not just "authorizer") and the body - at least the beginning of the body - at the same time?

xemul:
> Hi @xemul there's something I don't understand about this API. We're doing this in the context of Seastar's HTTP client - not any specific user like s3. HTTP doesn't have the concept of an "authorizer", or authorization headers, so I think it's strange that our API will add such a concept.

You're not quite right, it [has](https://datatracker.ietf.org/doc/html/rfc7235#page-8)

> Morever, I don't understand why this concept is needed. There's always a close connection between the headers and the body of the request - for example the headers might mention the time of the requests, cookies that need to go with the request, the size of the request, a signature of the entire request, and so on. I don't think it ever makes sense to generate headers and then one hour later - generate the body. So why is this even an issue? Why don't we generate the headers (all of them, not just "authorizer") and the body - at least the beginning of the body - at the same time?

Maybe it's not the best way possible indeed, but currently the API of making the request is
- caller prepares the `class http::request` object with all data on-board
- caller passes the request by value into the `http::client::make_request(...)` method

From now on, there's no way for the client to change anything on the request other than the contents of the body and even the latter is only possible _provided_ the body is generated using the callable.

Said that, there should be a way for the client to ... refresh (?) the request object before http client code puts it on a wire.

The need to refresh can arise for another reason other than the described above. E.g., from what I've seen -- when sending a request into the socket server can, for whatever reason, just close its end without reading it up to the end (I'm seeing it once every several minutes of uploading large S3 object -- AWS S3 just drops the connection and the client gets EPIPE or ECONABORTED depending on which part he's in -- writing request or reading reply). And it's not right to just bail out with an exception letting the caller do something with it. There's a "standard practice" that failed request is tried again after a while in a hope that there was a temporary networking glitch. And this is what our http client currently doesn't do, but it should. In that case, re-authorizing the request is also required on an existing `http::request` object.

avikivity:
It's a little "special" but I don't have a better (more generic) idea.

nyh:
>     * caller prepares the `class http::request` object with all data on-board
> 
>     * caller passes the request by value into the `http::client::make_request(...)` method
> 
> 
> From now on, there's no way for the client to change anything on the request other than the contents of the body and even the latter is only possible _provided_ the body is generated using the callable.

So, maybe the solution should be to allow not just the body, but also the headers, to be generated by a callable?

> Said that, there should be a way for the client to ... refresh (?) the request object before http client code puts it on a wire.
> The need to refresh can arise for another reason other than the described above. E.g., from what I've seen -- when sending a request into the socket server can, for whatever reason, just close its end without reading it up to the end (I'm seeing it once every several minutes of uploading large S3 object -- AWS S3 just drops the connection and the client gets EPIPE or ECONABORTED depending on which part he's in -- writing request or reading reply). And it's not right to just bail out with an exception letting the caller do something with it. There's a "standard practice" that failed request is tried again after a while in a hope that there was a temporary networking glitch. And this is what our http client currently doesn't do, but it should. In that case, re-authorizing the request is also required on an existing `http::request` object.

If we had a callable generating the headers, just like we have (optionally) a callable generating the body, then we can easily call these things again if we need to resend the request. Note that the body's callable might even generate different content (if, theoretically, the body of the request contains data that has changed since the previous send) so we need to calculate new headers anyway. I'm just suggesting maybe this recalculation needs to be something general for all "headers", not just for "authorization"?

xemul:
> So, maybe the solution should be to allow not just the body, but also the headers, to be generated by a callable?

And, eventually, query parameters too :)

>  I'm just suggesting maybe this recalculation needs to be something general for all "headers", not just for "authorization"?

Well, yes, the way this call works allows the caller to reset any header, not just the Authorization one. How about renaming the `_authorize` into generic `_refresh` then?

tchaikov:
if it is a normality that the header or body vary based on when the requests is actually sent, probably we should use something like `prepare`?

nyh:
Another option (which I think is relevant to your intended use case of s3 in scylladb) if for the lowest level of the http client library not to do any retry: If an HTTP request fails, or cannot be done in a timely manner, the library will return an error to the caller and not retry. The caller can then decide if it can retry exactly the same query, or needs to modify the query or headers before retrying. The library can then add an option or function that does retries in a specific way, which might not be appropriate for all applications.

xemul:
@nyh , you're correct about interaction with retries. Let's then first decide with the #1847 one that adds the restarting ability

xemul:
Nor relevant any longer, authorization is solved by other means in the caller code (scylla S3 client)

PR #1847, opened 2023-09-22
Make http client restart requests (backoff-retry)

xemul:
Right now if an http request fails for whatever reason, the exception is propagated back to caller. However, if the failure is due to connection problems, often the only option to handle it is to try the same request again a bit later. This PR adds this ability to the client.

Tested with scylla S3 client over "real" S3 server

closes: #1883 (this PR is another implementation of restarts)

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -166,6 +166,10 @@ public:
  */
 
 class client {
+public:
+    using reply_handler = noncopyable_function<future<>(const reply&, input_stream<char>&& body)>;
+
+private:
     friend class http::internal::client_ref;
     using connections_list_t = bi::list<connection, bi::member_hook<connection, typename connection::hook_t, &connection::_hook>, bi::constant_time_size<false>>;
     static constexpr unsigned default_max_connections = 100;
@@ -187,8 +191,9 @@ class client {
     SEASTAR_CONCEPT( requires std::invocable<Fn, connection&> )
     auto with_connection(Fn&& fn);
 
+    future<> do_make_request(request req, reply_handler handle, reply::status_type expected);
+
 public:
-    using reply_handler = noncopyable_function<future<>(const reply&, input_stream<char>&& body)>;
     /**
      * \brief Construct a simple client
      *
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -191,7 +191,7 @@ private:
     SEASTAR_CONCEPT( requires std::invocable<Fn, connection&> )
     auto with_connection(Fn&& fn);
 
-    future<> do_make_request(request& req, reply_handler handle, reply::status_type expected);
+    future<> do_make_request(request& req, reply_handler& handle, reply::status_type expected);
 
 public:
     /**
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -126,7 +126,7 @@ public:
     future<> close();
 
 private:
-    future<reply_ptr> do_make_request(request rq);
+    future<reply_ptr> do_make_request(request& rq);
     void setup_request(request& rq);
     future<> send_request_head(const request& rq);
     future<reply_ptr> maybe_wait_for_continue(const request& req);
@@ -191,7 +191,7 @@ private:
     SEASTAR_CONCEPT( requires std::invocable<Fn, connection&> )
     auto with_connection(Fn&& fn);
 
-    future<> do_make_request(request req, reply_handler handle, reply::status_type expected);
+    future<> do_make_request(request& req, reply_handler handle, reply::status_type expected);
 
 public:
     /**
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -155,6 +155,23 @@ public:
 };
 
 /**
+ * \brief Parameters for retrying failed requests
+ *
+ * In case server closes the connection in the middle of client writing request into it or
+ * reading the request back, the client may retry the request on the caller's bahalf and
+ * this param structure controls the way these attempts are made
+ *
+ * \param max_count -- the maximum number of retry attempts. Default is zero, so no
+ *                     retries would happen
+ * \param pause_base -- each new attempt waits 2x times more than the previous one before
+ *                      retrying, this is the base value for the sequence. Default is 100ms
+ */
+struct retry_param {
+    unsigned max_count = 0;
+    std::chrono::milliseconds pause_base = std::chrono::milliseconds(100);
+};
+
+/**
  * \brief Class client wraps communications using HTTP protocol
  *
  * The class allows making HTTP requests and handling replies. It's up to the caller to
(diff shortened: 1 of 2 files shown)


xemul:
upd:
- rebased

src/http/client.cc
```diff
@@ -164,13 +163,14 @@ future<connection::reply_ptr> connection::do_make_request(request req) {
                 });
             });
         });
-    });
 }
 
 future<reply> connection::make_request(request req) {
-    return do_make_request(std::move(req)).then([] (reply_ptr rep) {
+  return do_with(std::move(req), [this] (auto& req) {
```
tchaikov: could you please extract this into its own commit? as it actually prolongs the lifecycle of `req` so that it is alive until `do_make_request(req)` is resolved ? this addresses the problem we discussed the other day on why `s3::client::get_object_tagging()` does not crash. and with this change, i think we will be able to drop the tricky hacks like https://github.com/scylladb/scylladb/blob/2a81b2e49af8ca01a058675dd4b468c6145f9aad/utils/s3/client.cc#L271 . as `req` as the captured variables will be alive after the lambda resumes after it is suspended, even we don't capture it as a local variable and force the coroutine runtime to save it to the coroutine frame.

include/seastar/http/client.hh
```diff
@@ -154,6 +154,23 @@ public:
...
+ * \brief Parameters for retrying failed requests
+ *
+ * In case request fails due to transport errors, the client may retry it on the caller's
+ * bahalf and this param structure controls the way these attempts are made
+ *
+ * \param max_count -- the maximum number of retry attempts. Default is zero, so no
+ *                     retries would happen
+ * \param pause_base -- client uses exponentially increasing delays between retry attempts
+ *                      and this is the base value for the exponent. First attempt would
+ *                      happen after this duration. Default is 100ms
```
nyh, line 167: Since the exponent is not configurable, please at least say what it is (I looked at the code and saw it's 2).
To be honest, I'm not sure why exponential backoff makes sense for the case of a **transport error**. Presumably, this happens in the middle of a long request. Does it make sense to wait 0.1 seconds and if it happens again 10 seconds later, wait 0.2 seconds, and if it happens again later, wait 0.4 seconds? I'm not sure how this is helpful.

Please also say what a "transport error" is. If the remote host is down (either timeout in SYN, or ICMP host-not-reachable, or I don't know what) - will the request be retried or not?

src/http/client.cc
```diff
@@ -307,9 +308,32 @@ auto client::with_connection(Fn&& fn) {
...
 }
 
-future<> client::make_request(request req, reply_handler handle, reply::status_type expected) {
-    return do_with(std::move(req), std::move(handle), [this, expected] (request& req, reply_handler& handle) mutable {
-        return do_make_request(req, handle, expected);
+future<> client::make_request(request req, reply_handler handle, reply::status_type expected, retry_param retry) {
+    return do_with(std::move(req), std::move(handle), 0u, [this, expected, retry] (request& req, auto& handle, unsigned& attempt) mutable {
+        // Default fast-path
+        if (retry.max_count == 0) {
+            return do_make_request(req, handle, expected);
```
nyh, line 315: ‎There's a delicate issue here that should be documented:
When the connection breaks in the specific way (EPIPE or ECONNABORTED), the same handler is reused for the second connection. If the handler keeps some sort of internal state, it must reset it on every call - NOT in its constructor. 
We once had in Alternator a bug exactly like that - a supposedly used-once object that had some of its state reset only in the constructor, and then unexpectedly, it was used twice and we didn't get this reset before the second call.

nyh (changes requested):

include/seastar/http/client.hh
xemul: Updated comment

Well, this is what happens with S3 server (as I see it from logs, unfortunately I didn't collect strong statistics)

1. client accumulates 5 connections in pool
2. user sends a request, client picks a connection from pool and starts writing request into it
3. if writing request _or_ reading response fails with, respectively, EPIPE or ECONRESET -> goto 5
4. return success
5. client waits `0.1 << attempt` sec, then picks up another connection from pool -> goto step 2


Some heuristics:

- S3 server is pretty unstable in terms of how actively it kills the connections runtime. I could run a test for minutes using hundreds of connections and uploading gigs of data, no EPIPE/ECONRESET. But at times server could close every 2nd connection, so retry code was pretty active
- In most of the cases _single_ retry (i.e. -- using two connections) after 100ms helped
- Rarely server fails the first retry attempt (after failing the initial attempt), so client uses 3 connections to succeed
- Using first retry timeout of 10ms increases the chance of client draining the whole pool and proceeding with fresh new connection. Note, that I used 4 retries count, so 10ms + 20ms + 40ms + 80ms = 150ms.

The last observation makes me think that with S3 100ms is some sane duration after which server heals itself

src/http/client.cc
xemul: Good point. Added

include/seastar/http/client.hh
xemul: Another retry option that also worked, was -- once connection from pool closes in the middle, just go ahead and establish new one even if there are other connections in the pool

pros:
- the logic is more straightforward and doesn't imply magic constatnts

cons:
- initiates connect()ion storm at times
- the resulting patch is little bit more hairy

src/http/client.cc
tchaikov: so my impression was that there two things going on here:

1. change the parameter type
2. use `do_with()`

include/seastar/http/client.hh
xemul: @nyh , the force-new-connection options from above is in the https://github.com/xemul/seastar/tree/br-http-restart-requests-a branch. If you think it's more appropriate than the exponential-backoff-retry approach, let me know, I'll force push that branch here for review

src/http/client.cc
tchaikov: i see. could you please emphasis the problem addressed by this change, and explain how the problem is addressed?

> Now the method in question gets the request value and runs its code inside the do_with(std::move(request), ...) context

is not obvious to me that "now ... gets ..." is the change made in this very commit. it is just not explicit enough, i think.

and the problem addressed by keeping the `request` alive using `do_with` , and the change of `do_with()` is missing.

would you mind addressing these concerns?

xemul: Well, the "problem" is: there are two methods:

- `connection::make_request()`
- `connection::do_make_request()`

The former _moves_ request into the latter. Also, there's `client::make_request()` which also _moves_ request into `connection::do_make_request()`. 

Next, this PR tries to teach `client::make_request()` restart requests. Restarting means that `client::make_request()` calls `connection::do_make_request()`. If the latter fails, the former needs to do something and (!) call it _again_. But when calling it again it should have the request at hand to pass it as an argument. In current approach `client::make_request()` won't have the request, because it had _moved_ it into the `connection::do_make_request()` on the first try. So I need to keep request object in `client::make_request()` so that it does _not_ move it into `connection::do_make_request()` but passes there a reference. Reference can be passed several times (and #1848 makes sure request remains intact between retries)

tchaikov: thank you for the detailed explanation. even the change is a necessity of the retry implementation, it still addresses the problem we discussed couple weeks ago. personally, i would want to ensure that i can drop the hackish local variables in the caller, after reading the commit message. but i cannot find it in the commit message. so i am wondering if it is possible to note it in the commit message.

and i think, my concern is the impaction of the change, which _happens_ to address the problem we discussed the other day. not quite related to the "problem" you explained in your previous reply.

PR #1848, opened 2023-09-22
Make sure http client doesn't change request while sending

xemul:
The confidence comes from marking the request reference const throughout the sending methods. There are only two places that are allowed to change the request, and after this patch they become explicit and, thus, controllable.

refs: #1847 (this pr is prerequisite for that one)
fixes: scylladb/scylladb#15509

### DIFF
--- a/include/seastar/http/request.hh
+++ b/include/seastar/http/request.hh
@@ -259,7 +259,7 @@ struct request {
 private:
     void add_param(const std::string_view& param);
     sstring request_line() const;
-    future<> write_request_headers(output_stream<char>& out);
+    future<> write_request_headers(output_stream<char>& out) const;
     friend class experimental::connection;
 };
 
--- a/include/seastar/http/request.hh
+++ b/include/seastar/http/request.hh
@@ -60,6 +60,7 @@ struct request {
     sstring _version;
     ctclass content_type_class;
     size_t content_length = 0;
+    size_t _bytes_written = 0;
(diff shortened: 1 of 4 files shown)


include/seastar/http/request.hh
```diff
@@ -60,6 +60,7 @@ struct request {
     sstring _version;
     ctclass content_type_class;
     size_t content_length = 0;
+    mutable size_t _real_content_length = 0;
```
tchaikov, line 63: i'd suggest change the variable name to `_bytes_written`. because it is not the length of the request's content. instead, it is used to tell the write ptr's position. yeah, it is also used for passing the length to the data sink. so it's both an input parameter **and** an output parameter.

this is both clever and confusing. this is more related to the design of `http_content_length_data_sink_impl` , and is not in the scope of this change. but i think we can make it less confusing with a better name.

tchaikov: i created #1849 to explain what i am thinking.

xemul: OK, renamed

xemul:
upd:
- rebase
- rename _real_content_length -> _bytes_written

include/seastar/http/request.hh
tchaikov, line 1: could you please update the title of this commit message, and also the body of it. they are still referencing "content length" instead of "bytes written".

xemul: Yikes :( Updated, thanks for catching

tchaikov (approved):
lgtm

tchaikov:

xemul:
@nyh , please review

PR #1849, opened 2023-09-22
http: do not use req.content_length as both input parameter

tchaikov:
before this change, req.content_length is used as both input and output parameter. but it would be clearer if we only use it for tracking the progress of writing the content.

### DIFF
--- a/src/http/common.cc
+++ b/src/http/common.cc
@@ -152,11 +152,13 @@ class http_content_length_data_sink_impl : public data_sink_impl {
     size_t& _bytes_written;
 
 public:
-    http_content_length_data_sink_impl(output_stream<char>& out, size_t& len)
+    http_content_length_data_sink_impl(output_stream<char>& out, size_t total_len, size_t& bytes_written)
             : _out(out)
-            , _limit(std::exchange(len, 0))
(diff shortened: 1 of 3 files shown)


src/http/common.cc
```diff
@@ -152,10 +152,10 @@ class http_content_length_data_sink_impl : public data_sink_impl {
     size_t& _bytes_written;
 
 public:
-    http_content_length_data_sink_impl(output_stream<char>& out, size_t& len)
+    http_content_length_data_sink_impl(output_stream<char>& out, size_t total_len, size_t& bytes_written)
             : _out(out)
-            , _limit(std::exchange(len, 0))
```
xemul, line 157: I think there's a bug here. This line used to put 0 into the `req._bytes_written` before the sink started to work. After this PR the `req._bytes_written` remains unchanged and will re-use the value left from previous request (connections are long-living and may serve several requests one after another)

tchaikov: ahh, indeed. we should always reset the `_bytes_written` to 0 in the ctor. fixed and repushed.

tchaikov:
v2:

- reset `_bytes_written` in ctor.

Issue #1850, opened 2023-09-27
Add network connection information to `http::request` structure 


michael-redpanda:
Handlers for HTTP requests may need to know the IP address and port of the client that generated the HTTP request.  This information could be added to the `http::request` structure that is passed to the handlers.

PR #1851, opened 2023-09-27
httpd: Added server and client addresses to request structure

michael-redpanda:
Adds the client and servers network address information to the `seastar::http::request` structure and populates it when it is created.

Fixes: #1850

### DIFF
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -76,6 +76,8 @@ class connection : public boost::intrusive::list_base_hook<> {
     connected_socket _fd;
     input_stream<char> _read_buf;
     output_stream<char> _write_buf;
+    socket_address _client_addr;
+    socket_address _server_addr;
     static constexpr size_t limit = 4096;
     using tmp_buf = temporary_buffer<char>;
     http_request_parser _parser;
@@ -89,8 +91,22 @@ public:
     connection(http_server& server, connected_socket&& fd,
             socket_address) : connection(server, std::move(fd)) {}
     connection(http_server& server, connected_socket&& fd)
-            : _server(server), _fd(std::move(fd)), _read_buf(_fd.input()), _write_buf(
-                    _fd.output()) {
+            : _server(server)
+            , _fd(std::move(fd))
+            , _read_buf(_fd.input())
+            , _write_buf(_fd.output())
+            , _client_addr(_fd.remote_address())
+            , _server_addr(_fd.local_address()) {
+        on_new_connection();
+    }
+    connection(http_server& server, connected_socket&& fd,
+            socket_address client_addr, socket_address server_addr)
+            : _server(server)
+            , _fd(std::move(fd))
+            , _read_buf(_fd.input())
+            , _write_buf(_fd.output())
+            , _client_addr(std::move(client_addr))
+            , _server_addr(std::move(server_addr)) {
         on_new_connection();
     }
     ~connection();
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -220,6 +220,10 @@ future<> connection::read_one() {
         }
         ++_server._requests_served;
         std::unique_ptr<http::request> req = _parser.get_parsed_request();
+
+        req->_server_address = this->_server_addr;
+        req->_client_address = this->_client_addr;
+
         if (_server._credentials) {
             req->protocol_name = "https";
         }
@@ -423,7 +427,9 @@ future<> http_server::do_accepts(int which) {
 
 future<> http_server::do_accept_one(int which) {
     return _listeners[which].accept().then([this] (accept_result ar) mutable {
-        auto conn = std::make_unique<connection>(*this, std::move(ar.connection), std::move(ar.remote_address));
+        auto local_address = ar.connection.local_address();
+        auto conn = std::make_unique<connection>(*this,
+            std::move(ar.connection), std::move(ar.remote_address), std::move(local_address));
         (void)try_with_gate(_task_gate, [conn = std::move(conn)]() mutable {
             return conn->process().handle_exception([conn = std::move(conn)] (std::exception_ptr ex) {
                 hlogger.error("request error: {}", ex);
--- a/include/seastar/core/posix.hh
+++ b/include/seastar/core/posix.hh
@@ -295,6 +295,12 @@ public:
         throw_system_error_on(r == -1, "getsockname");
         return addr;
     }
+    socket_address get_remote_address() {
+        socket_address addr;
+        auto r = ::getpeername(_fd, &addr.u.sa, &addr.addr_length);
+        throw_system_error_on(r == -1, "getpeername");
+        return addr;
+    }
     void listen(int backlog) {
         auto fd = ::listen(_fd, backlog);
         throw_system_error_on(fd == -1, "listen");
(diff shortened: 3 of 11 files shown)


xemul:
@michael-redpanda , please, split the patch into two
1. add `get_remote_address` to sockets API
2. equip http request with local and remote addresses (the essence of this PR)

michael-redpanda:
> @michael-redpanda , please, split the patch into two
> 
> 1. add `get_remote_address` to sockets API
> 2. equip http request with local and remote addresses (the essence of this PR)

done

src/http/httpd.cc
```diff
@@ -220,6 +220,10 @@ future<> connection::read_one() {
         }
         ++_server._requests_served;
         std::unique_ptr<http::request> req = _parser.get_parsed_request();
+
+        req->_server_address = _fd.local_address();
+        req->_client_address = _fd.remote_address();
```
xemul, line 225: This happens per every request sent by the client, and both `::getpeername()` and `::getsockname()` are system calls that (!) can yield in the kernel. We typically try to avoid both in seastar. So it's definitely better for `connection::read_one()` to get the local and remote addresses in advance, keep them on connection itself and just copy to request, instead of talking to the kernel every time.

Another thing to note here, is that remote address can be known from the `accept_result` when the new connection is accepted, so the new `remote_address()` call is good to have, but likely not required here.

michael-redpanda: Thanks for the review!  Appreciate the feedback.

> Another thing to note here, is that remote address can be known from the accept_result when the new connection is accepted, so the new remote_address() call is good to have, but likely not required here.

I noticed that one of the constructors for `http::connection` would accept a `socket_address` but was marked deprecated: https://github.com/scylladb/seastar/blob/f11f1b4370ed4414af552e5269bb173c721617fa/include/seastar/http/httpd.hh#L88-L90

Can I un-deprecate this in order to pass the address to the constructor?  I'm not sure what historical reason there is as to why it was deprecated.

xemul: But you'll need to construct it with _two_ socket_address-es :) So I would say it's not going to be undeprecation. Instead, there will be another constructor with local and remote addresses, and the existing one ... would need to call `_fd.get_local_address()` and `_fd.get_remote_address()` for those values

src/http/httpd.cc
```diff
@@ -423,7 +427,9 @@ future<> http_server::do_accepts(int which) {
 
 future<> http_server::do_accept_one(int which) {
     return _listeners[which].accept().then([this] (accept_result ar) mutable {
-        auto conn = std::make_unique<connection>(*this, std::move(ar.connection), std::move(ar.remote_address));
+        auto conn = std::make_unique<connection>(*this,
+            std::move(ar.remote_address), ar.connection.local_address(),
+                    std::move(ar.connection));
```
xemul, line 432: This is problematic -- the order of `ar.connection.local_address()` and `std::move(ar.connection)` is undefined, so it may happen that the `.local_address()` is called on moved-from `ar.connection`

src/http/httpd.cc
```diff
@@ -220,6 +220,10 @@ future<> connection::read_one() {
         }
         ++_server._requests_served;
         std::unique_ptr<http::request> req = _parser.get_parsed_request();
+
+        req->_server_address = _fd.local_address();
```
xemul, line 224: Why `_fd.local_address()`? It should be `this->_server_address`. Same for `_client_address` below

include/seastar/http/httpd.hh
```diff
@@ -87,10 +89,18 @@ class connection : public boost::intrusive::list_base_hook<> {
 public:
     [[deprecated("use connection(http_server&, connected_socket&&)")]]
     connection(http_server& server, connected_socket&& fd,
-            socket_address) : connection(server, std::move(fd)) {}
+            socket_address client_addr)
+            : connection(server, std::move(client_addr), 
+                    fd.local_address(), 
+                    std::move(fd)) {}
     connection(http_server& server, connected_socket&& fd)
+            : connection(server, fd.remote_address(),
```
xemul, line 97: Trailing space

include/seastar/http/httpd.hh
```diff
@@ -87,10 +89,18 @@ class connection : public boost::intrusive::list_base_hook<> {
...
     [[deprecated("use connection(http_server&, connected_socket&&)")]]
     connection(http_server& server, connected_socket&& fd,
-            socket_address) : connection(server, std::move(fd)) {}
+            socket_address client_addr)
+            : connection(server, std::move(client_addr), 
+                    fd.local_address(), 
+                    std::move(fd)) {}
     connection(http_server& server, connected_socket&& fd)
+            : connection(server, fd.remote_address(), 
+                    fd.local_address(), std::move(fd)) {}
```
xemul, line 98: The order of `fd.<foo>_address()` and `std::move(fd)` is undefined

include/seastar/http/httpd.hh
```diff
@@ -87,10 +89,18 @@ class connection : public boost::intrusive::list_base_hook<> {
...
+    connection(http_server& server, socket_address client_addr,
+            socket_address server_addr, connected_socket&& fd) 
             : _server(server), _fd(std::move(fd)), _read_buf(_fd.input()), _write_buf(
-                    _fd.output()) {
+                    _fd.output()), _client_addr(std::move(client_addr)),
+                    _server_addr(std::move(server_addr)) {
```
xemul, line 103: nit: Please, reformat the initializer list so that it looks like

```c++
    connection(...)
        : _server(server)
        , _fd(std::move(fd))
        , _read_buf(_fd.input())
        ...
```

src/http/httpd.cc
```diff
@@ -423,7 +427,8 @@ future<> http_server::do_accepts(int which) {
 
 future<> http_server::do_accept_one(int which) {
     return _listeners[which].accept().then([this] (accept_result ar) mutable {
-        auto conn = std::make_unique<connection>(*this, std::move(ar.connection), std::move(ar.remote_address));
+        auto conn = std::make_unique<connection>(*this,
+            std::move(ar.connection), std::move(ar.remote_address));
```
xemul, line 431: This calls deprecated `connection(server, fd, address)` constructor which ignores the `remote_address` argument and goes calling methods on fd.

@tchaikov , something is wrong with 3841bbdd19e8fb879c266d02a6f1ffbf3cb6f8ee -- this place in httpd.cc still uses the deprecated constructor (even before patching), but compiler silently allows that

michael-redpanda: Except for the white space I erroneously added, this was how `connection` was constructed before my PR

michael-redpanda: I can change it to call the new constructor

xemul: > I can change it to call the new constructor

The plan was to 
- get remote address from accept_result
- in case connection is constructed without accept_result at hand -- call the fd.remote_address()

tchaikov: > This calls deprecated `connection(server, fd, address)` constructor which ignores the `remote_address` argument and goes calling methods on fd.
> 
> @tchaikov , something is wrong with [3841bbd](https://github.com/scylladb/seastar/commit/3841bbdd19e8fb879c266d02a6f1ffbf3cb6f8ee) -- this place in httpd.cc still uses the deprecated constructor (even before patching), but compiler silently allows that

see #1817 . i wanted to add the compile option to let compiler complains when compiling the caller sites of deprecated symbols. but alas, Avi does not agree with me on this.

PR #1853, opened 2023-09-28
build: pass target name to seastar_check_self_contained

tchaikov:
instead of hard-wiring the target for "checkheaders" to "checkheaders", pass the target name as a parameter. also, this change does not add a header check helper target if the specified library target does not contain any headers.

this allows us to reuse the `seastar_check_self_contained()` function in the parent project.


tchaikov:
because scylladb also provides targets named `{mode}-headers`. to enable it to reuse the `seastar_check_self_contained`, let's generalize this function a little bit.

PR #1854, opened 2023-09-29
net/tls: Expose certificate internals

oleiman:
Adds `tls::get_impl` to expose the gnutls objects underlying `tls::certificate_credentials`. In particular, we want to access the trust list, cert expiry, and cert serial number, primarily for monitoring purposes.

Also adds an additional reload callback type which includes a reference to the loaded `certificate_credentials` object.


avikivity:
Please configure your circleci stuff so it runs the checks.

include/seastar/net/tls.hh
```diff
@@ -152,6 +152,10 @@ namespace tls {
      */
     using dn_callback = noncopyable_function<void(session_type type, sstring subject, sstring issuer)>;
 
+    class certificate_credentials;
+    void* get_impl(certificate_credentials const&);
+
```
avikivity, line 157: No documentation?

src/net/tls.cc
```diff
@@ -1980,6 +1981,9 @@ std::ostream& tls::operator<<(std::ostream& os, const subject_alt_name& a) {
     return os << a.type << "=" << a.value;
 }
 
+void* tls::get_impl(const tls::certificate_credentials &creds) {
+    return creds._impl->_creds;
```
avikivity, line 2002: Why return void*? What is the caller supposed to do with it?

include/seastar/net/tls.hh
```diff
@@ -301,7 +302,9 @@ namespace tls {
         // same as above, but any files used for certs/keys etc will be watched
         // for modification and reloaded if changed
         future<shared_ptr<certificate_credentials>> build_reloadable_certificate_credentials(reload_callback = {}, std::optional<std::chrono::milliseconds> tolerance = {}) const;
+        future<shared_ptr<certificate_credentials>> build_reloadable_certificate_credentials(reload_callback_with_creds, std::optional<std::chrono::milliseconds> tolerance = {}) const;
         future<shared_ptr<server_credentials>> build_reloadable_server_credentials(reload_callback = {}, std::optional<std::chrono::milliseconds> tolerance = {}) const;
+        future<shared_ptr<server_credentials>> build_reloadable_server_credentials(reload_callback_with_creds, std::optional<std::chrono::milliseconds> tolerance = {}) const;
```
avikivity, line 307: What's the motivation for this patch?

oleiman:
@avikivity @elcallio  Thanks for the review! I will rework this a bit and circle back.

src/net/tls.cc
avikivity: operator is too magical, I prefer the explicit `as_gnutls_certificate_credentials_t`.

about abstracted accessors, I agree in principle, but if this a shallow-but-wide interface (large number of simple getters) it may be difficult to maintain.

avikivity: (is it a shallow-but-wide interface?)

oleiman: > is it a shallow-but-wide interface?

Not really. The information we want is:
- serial number and expiration time of each loaded cert
- serial number and expiration time of each CA in the trust list

I have a working branch that adds the following interface to `net::tls::certificate_credentials`:

```c++
std::optional<std::vector<cert_info>> get_cert_info() const noexcept;
std::optional<std::vector<cert_info>> get_trust_list_info() const noexcept;
```

oleiman: If that looks agreeable, I'll get my branch cleaned up resubmitted here.

For reference, `cert_info` looks like this:
```
struct cert_info {
    sstring serial;
    time_t expiry;
};
```

avikivity: Sure it makes sense. Presumably, the same interface would work for openssl?

oleiman: I assume so. The gnutls bits are handled inside `::impl`, but I don't know the corresponding openssl incantations offhand.

src/net/tls.cc
```diff
@@ -887,8 +897,14 @@ class tls::reloadable_credentials_base {
             }
         }
         void do_callback(std::exception_ptr ep = {}) {
-            if (_cb && !_files.empty()) {
```
elcallio, line 890: This check still seems relevant. If we have no files, we have nothing to report, because we did not in fact reload anything.

src/net/tls.cc
```diff
@@ -887,8 +897,14 @@ class tls::reloadable_credentials_base {
...
-                _cb(boost::copy_range<std::unordered_set<sstring>>(_files | boost::adaptors::map_keys), std::move(ep));
+            if (_cb) {
```
elcallio, line 981: If ep != nullptr, you should not do this. You potentially hide an exception (either by succeeding or by generating a new exception  - the latter more likely)

src/net/tls.cc
```diff
@@ -887,8 +897,14 @@ class tls::reloadable_credentials_base {
...
+                shared_ptr<const tls::certificate_credentials> creds = nullptr;
```
elcallio, line 982: This creates a new certificate set. Why not add a virtual method to reloadable_credentials_base, `virtual tls::certificate_credentials& as_certificate_credentials() const = 0`, that simply returns `*this`, and change your callback to take a reference instead?
This would also ensure we don't generate extra overhead for callbacks that don't care about them, as well as avoid having to potentially generate an extra exception that was not part of the reload cycle until now (race).

elcallio (changes requested):

src/net/tls.cc
oleiman: Great suggestion. Done.

src/net/tls.cc
```diff
@@ -887,8 +896,9 @@ class tls::reloadable_credentials_base {
             }
         }
         void do_callback(std::exception_ptr ep = {}) {
-            if (_cb && !_files.empty()) {
-                _cb(boost::copy_range<std::unordered_set<sstring>>(_files | boost::adaptors::map_keys), std::move(ep));
+            if (_cb && !_files.empty() && _creds) {
+                const auto &creds = _creds->as_certificate_credentials();
+                _cb(boost::copy_range<std::unordered_set<sstring>>(_files | boost::adaptors::map_keys), creds, std::move(ep));
```
elcallio, line 982: How is the set of files relevant for the receiver of the call? Given that we provide no relation to the actual files used by the resulting credentials, it seems a receiver must know them anyway to match them to anything, no?
Here it is not even changed files, but all watched. Seems like unneeded overhead...

oleiman: I hadn't given it much thought to be honest. Intention was for this to be a non-breaking change. Is your suggestion to remove this now that the credentials are included with the callback?

elcallio (approved):

src/net/tls.cc
oleiman: Good stuff, thanks :)

oleiman:
force push after rebase to pull in `rpc_test` fix

include/seastar/net/tls.hh
```diff
@@ -263,6 +263,7 @@ namespace tls {
     class reloadable_credentials_base;
 
     using reload_callback = std::function<void(const std::unordered_set<sstring>&, std::exception_ptr)>;
+    using reload_callback_with_creds = std::function<void(const std::unordered_set<sstring> &, const certificate_credentials&, std::exception_ptr)>;
```
avikivity, line 284: Will we have reload_callback_with_creds_and_something_else? If there's any likelihood of that, let's replace the parameter with a struct that can be expanded. The struct can also serve to document the parameters.

src/net/tls.cc
```diff
@@ -962,20 +975,38 @@ void tls::reloadable_credentials<tls::certificate_credentials>::rebuild(const cr
     this->_impl = std::move(tmp->_impl);
 }
 
+template <>
+const tls::certificate_credentials &tls::reloadable_credentials<tls::certificate_credentials>::as_certificate_credentials() const noexcept {
+    return *this;
```
avikivity, line 1061: Our coding style attaches the reference sigil to the type, not the function name.

include/seastar/net/api.hh
```diff
@@ -169,6 +169,12 @@ struct session_dn {
     sstring issuer;
 };
 
+  /// Information about a certificate
```
avikivity, line 172: Please add some motivation to the changelog.

src/net/tls.cc
```diff
@@ -387,6 +397,51 @@ class tls::certificate_credentials::impl: public gnutlsobj {
                 gnutls_certificate_set_x509_simple_pkcs12_mem(_creds, &w,
                         gnutls_x509_crt_fmt_t(fmt), password.c_str()));
     }
+    std::vector<cert_info> get_x509_info() const {
```
avikivity, line 400: These can't be used?

include/seastar/net/tls.hh
elcallio: Is it worth that to break the API (will require scylla modification). Or new, struct + old?
Strictly, the callback arg should be a variant of { ep, <cert_data>+ whatnot }

src/net/tls.cc
```diff
@@ -387,6 +397,51 @@ class tls::certificate_credentials::impl: public gnutlsobj {
...
+        unsigned int crt_list_size{};
+        gtls_chk(
+          gnutls_certificate_get_x509_crt(*this, 0, &crt_list, &crt_list_size));
+        // auto certs = std::vector<gnutls_x509_crt_t>(crt_list, crt_list + crt_list_size);
+        auto cleanup = defer([&crt_list, crt_list_size]() noexcept {
+          for (unsigned int i = 0; i < crt_list_size; ++i) {
+            gnutls_x509_crt_deinit(crt_list[i]);
+          }
+          gnutls_free(crt_list);
+        });
```
avikivity, line 411: We use 4-space indents.

include/seastar/net/tls.hh
oleiman: >will we have reload_callback_with_creds_and_something_else?

From the redpanda side, I doubt it. If it seems likely/desirable in a general sense, sure I can add it into this PR.

include/seastar/net/api.hh
oleiman: In the commit message, yeah? Or is there a separate changelog somewhere?

src/net/tls.cc
oleiman: Sorry, could you elaborate? Used directly by clients you mean?

include/seastar/net/tls.hh
avikivity: I'm asking you, not about redpanda requirements, but whether there's more such auxiliary information that's likely to be added in the future. I'm not an expert about tls.

include/seastar/net/api.hh
avikivity: The commit message.

src/net/tls.cc
avikivity: They look orphaned, in a .cc file without a header. Later I saw the user in the next patch.

include/seastar/net/tls.hh
elcallio: Given that the callback is on a certificate_credentials, giving the object itself back is probably as much info as can be provided. I.e. I cannot see anything else that could be provided...

src/net/tls.cc
oleiman: Gotcha, thx. Squashed the two commits as suggested.

oleiman:
Hi @avikivity and @elcallio  - thanks again for the reviews. Any remaining concerns on this PR?

oleiman:
[force push](https://github.com/scylladb/seastar/compare/6b7304eeba4ee416b352d6a68d870b2f3b38a067..ae37fcb2a7e15102fb7dabdd495f7951f4d611fe) to express `cert_info::serial` as more explicitly a buffer of `uint8_t`.

include/seastar/net/api.hh
```diff
@@ -169,6 +169,14 @@ struct session_dn {
...
+struct cert_info {
+    static constexpr size_t bytes_inline_size = 31;
+    using bytes = basic_sstring<uint8_t, uint32_t, bytes_inline_size, false>;
```
avikivity, line 175: Why introduce a special type? Yes it saves a byte, but it forces a new instantiation of basic_sstring for a weak reason.

sstring should be fine.

include/seastar/net/tls.hh
```diff
@@ -220,6 +220,24 @@ namespace tls {
          */
         void set_dn_verification_callback(dn_callback);
 
+        /**
+         * Retrieve information about loaded certificates.
+         *
+         * Returns a std::vector of cert_info, each extracted from a loaded
+         * certificate, std::nullopt if the impl does not exist or we detect
+         * an error during extraction.
```
avikivity, line 228: Shouldn't we throw on error? It's dangerous to just swallow errors, esp. in security. The caller can always ignore the exception if they want.

include/seastar/net/api.hh
oleiman: Less to save a memory and more to express that this is an owned buffer of raw bytes as opposed to a legible string. Felt relevant here because one could reasonably expect either representation for a serial number. 

Anyway, if `sstring` is the preferred type for this usage, I'm happy to roll it back, perhaps expanding the doc comment.

include/seastar/net/tls.hh
oleiman: Fair point. I was thinking that an exception might be overkill since these are effectively just accessors on a fully constructed credentials struct, but I suppose more information is better.

oleiman:
[force push](https://github.com/scylladb/seastar/compare/ae37fcb2a7e15102fb7dabdd495f7951f4d611fe..1611333694a1d04d77870a27b2a7811f87c298e2) with review comments

- cert_info accessors now throw
- revert to `sstring` for delivering serial number bytes

include/seastar/net/api.hh
avikivity: Another alternative is std::vector\<std::byte>. I recognize the basic_sstring specialization, but I prefer not to introduce ad-hoc types. It would be different if seastar had a bytes type (well it does, temporary_buffer\<char>, but it's not suitable here).

src/net/tls.cc
```diff
@@ -987,7 +1064,7 @@ void tls::reloadable_credentials<tls::server_credentials>::rebuild(const credent
 }
 
 template <>
-const tls::certificate_credentials &tls::reloadable_credentials<tls::server_credentials>::as_certificate_credentials() const noexcept{
+const tls::certificate_credentials& tls::reloadable_credentials<tls::server_credentials>::as_certificate_credentials() const noexcept{
```
avikivity, line 1067: Move the fix to the patch that introduced the problem please.

include/seastar/net/api.hh
oleiman: >std::vector<std::byte>

Yeah, I like that better. I considered temporary_buffer as well, but as noted not a great fit.

src/net/tls.cc
oleiman: Hm, ya my mistake

oleiman:
[force pushes](https://github.com/scylladb/seastar/compare/e384b4a7ef004b402a47ae1b01e6333b9c38624a..ed2d02125672732258ff55968139644400855041) contents:
- replace `sstring` with `std::vector<std::byte>` for raw serial number representation
- move previous review ask to correct commit

PR #1857, opened 2023-10-03
core/posix return old behaviour using pthread_attr_setaffinity_np

a16bitsysop:
if it is present

as noted in the review of: https://github.com/scylladb/seastar/pull/1803/

and the commit
https://github.com/scylladb/seastar/commit/982690e8ead7d8cf30a71746ea0d55ee74f582a5#r1315429863

using the portable version could lead to a performance hit of some instructions running on the wrong core while the affinity was set after creation

### DIFF
--- a/cmake/CheckLibc.cmake
+++ b/cmake/CheckLibc.cmake
@@ -29,3 +29,13 @@ if (Strerror_R_Returns_Char_P)
     PRIVATE
       SEASTAR_STRERROR_R_CHAR_P)
 endif ()
+
+include (CheckFunctionExists)
+
+check_function_exists (pthread_attr_setaffinity_np
+  Pthread_Attr_Setaffinity_Np)
+if (Pthread_Attr_Setaffinity_Np)
+  target_compile_definitions (seastar
+    PRIVATE
+    SEASTAR_PTHREAD_ATTR_SETAFFINITY_NP)
+endif ()
--- a/src/core/posix.cc
+++ b/src/core/posix.cc
@@ -122,16 +122,25 @@ posix_thread::posix_thread(attr a, std::function<void ()> func)
     }
 #endif
 
+#ifdef SEASTAR_PTHREAD_ATTR_SETAFFINITY_NP
+    if (a._affinity) {
+        auto& cpuset = *a._affinity;
+        pthread_attr_setaffinity_np(&pa, sizeof(cpuset), &cpuset);
+    }
+#endif
+
     r = pthread_create(&_pthread, &pa,
                 &posix_thread::start_routine, _func.get());
     if (r) {
         throw std::system_error(r, std::system_category());
     }
 
+#ifndef SEASTAR_PTHREAD_ATTR_SETAFFINITY_NP
     if (a._affinity) {
         auto& cpuset = *a._affinity;
         pthread_setaffinity_np(_pthread, sizeof(cpuset), &cpuset);
     }
+#endif
 }
 
 posix_thread::posix_thread(posix_thread&& x)


a16bitsysop:
test isn’t working when `pthread_attr_setaffinity_np` exists

tchaikov:
please reference the exact commit sha1 in the commit message.

and did you create this change based on the comment of https://github.com/scylladb/seastar/pull/1803/commits/982690e8ead7d8cf30a71746ea0d55ee74f582a5#r1315429863 from @nyh ? or based on observation?

a16bitsysop:
yes it was based on the comment by @nyh not observation, searching for cases where this has been changed in other packages they just change to the call as in https://github.com/scylladb/seastar/commit/982690e8ead7d8cf30a71746ea0d55ee74f582a5#r1315429863

okay will add commit sha1 in commit message

tchaikov (changes requested):
i think the existing implementation is better.

>  using the portable version could lead to a performance hit of some instructions running on the wrong core while
the affinity was set after creation

i disagree. and i guess you misinterpreted Nadav's comment at https://github.com/scylladb/seastar/pull/1803#discussion_r1315429863 . let me quote his comment for our discussion

> The patch looks good:
> One (like me) might wonder if creating a new thread with the old affinity and only then changing the affinity doesn't have some small performance implications - like the thread being created on one CPU and then having to move. But in reality, the Linux clone() system call does not allow the user to control the new thread's cpu affinity, and one needs to first clone() the thread and then call sched_setaffinity. This is also what the glibc implementation does (I took a look) for pthread_attr_setaffinity_np - it just remembers it, and applies it after creating the new thread with clone(). So the fact you're doing this explicitly is perfectly fine.
> 

i think what he meant to say is that, under the hood, both `pthread_attr_setaffinity_np` and `pthread_setaffinity_np` set the affinity using `sched_setaffinity` syscall. and due to the limit of  `clone()` syscall, `pthread_attr_setaffinity_np` also calls `sched_setaffinity()` after the child thread is created. it's practically identical to what we have after your change. i also checked glibc's source code, see https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_create.c;h=6a41d50109cb27ecc09e56ced4745b7f83c06e07;hb=HEAD#l297 and https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_create.c;h=6a41d50109cb27ecc09e56ced4745b7f83c06e07;hb=HEAD#l313 .  and i concur with Nadav's remarks.

nyh:
@tchaikov please note that in later comments I went back to the glibc source code and noticed that they actually found a way to avoid running the user's code in the wrong affinity for even a picosecond: What they do is to start new new thread running glibc code which waits on a condition variable **before** calling the user's given thread function. Glibc then sets the affinity on this sleeping thread, and only then - when it has the right affinity - does it awake the condition variable and the new thread goes on to run the user's code.

We can actually do exactly this and get a perfect implementation even when pthread_attr_setaffinity_np doesn't exist. But @avikivity said he doesn't insist we bother to do it, but at least, if pthread_attr_setaffinity_np **does** exist, we can use it and not the incomplete phread_create-and-then-pthread_setaffinity_np approach _without_ that extra condition variable.

cmake/CheckLibc.cmake
```diff
@@ -29,3 +29,23 @@ if (Strerror_R_Returns_Char_P)
...
+int main() {
+    pthread_attr_t pa;
+    pthread_attr_init(&pa);
+    cpu_set_t cpuset;
+    pthread_attr_setaffinity_np(&pa, sizeof(cpuset), &cpuset);
+}"
+  Pthread_Attr_Setaffinity_Np)
+if (Pthread_Attr_Setaffinity_Np)
+  # define Pthread_Attr_Setaffinity_Np  if provided as is GNU-specific version,
+  # np is none portable.
```
nyh, line 47: "non-portable".

nyh (approved):

tchaikov:
> @tchaikov please note that in later comments I went back to the glibc source code and noticed that they actually found a way to avoid running the user's code in the wrong affinity for even a picosecond: What they do is to start new new thread running glibc code which waits on a condition variable **before** calling the user's given thread function. Glibc then sets the affinity on this sleeping thread, and only then - when it has the right affinity - does it awake the condition variable and the new thread goes on to run the user's code.
> 
> We can actually do exactly this and get a perfect implementation even when pthread_attr_setaffinity_np doesn't exist. But @avikivity said he doesn't insist we bother to do it, but at least, if pthread_attr_setaffinity_np **does** exist, we can use it and not the incomplete phread_create-and-then-pthread_setaffinity_np approach _without_ that extra condition variable.

thanks for the explanations. yeah, that's indeed better in the sense that the user function won't be running on the wrong core for even a jiffy.

cmake/CheckLibc.cmake
```diff
@@ -29,3 +29,23 @@ if (Strerror_R_Returns_Char_P)
     PRIVATE
       SEASTAR_STRERROR_R_CHAR_P)
 endif ()
+
+check_cxx_source_compiles ("
```
tchaikov, line 33: instead of using an application, we could just use
```cmake
include (CheckFunctionExists)

check_function_exists (pthread_attr_setaffinity_np
  Pthread_Attr_Setaffinity_Np)
```

simpler this way.

nyh:
@tchaikov any reason not to commit this patch?

It leaves the non-glibc situation imperfect, with the user code possibly running for a short while before the affinity changing (and I explained above how glibc, which also needs to use a post-creation sched_setaffinity(), solves this with a condition variable), but it returns the glibc version to be the way it was, without any downsides. I think it's better than what we have now.

cmake/CheckLibc.cmake
```diff
@@ -29,3 +29,23 @@ if (Strerror_R_Returns_Char_P)
...
+
+int main() {
+    pthread_attr_t pa;
+    pthread_attr_init(&pa);
+    cpu_set_t cpuset;
+    pthread_attr_setaffinity_np(&pa, sizeof(cpuset), &cpuset);
+}"
+  Pthread_Attr_Setaffinity_Np)
+if (Pthread_Attr_Setaffinity_Np)
+  # define Pthread_Attr_Setaffinity_Np  if provided as is GNU-specific version,
```
tchaikov, line 46: what is defined is `SEASTAR_PTHREAD_ATTR_SETAFFINITY_NP`, so might want to update the comment accordingly. but probably we can drop the comment. as the difference is less subtle than that between the two variants of `strerror_r()` -- they have different function signatures and usage.

tchaikov:
> @tchaikov any reason not to commit this patch?
> 
> It leaves the non-glibc situation imperfect, with the user code possibly running for a short while before the affinity changing (and I explained above how glibc, which also needs to use a post-creation sched_setaffinity(), solves this with a condition variable), but it returns the glibc version to be the way it was, without any downsides. I think it's better than what we have now.

see the comment above. i agree it's an improvement overall. and i can always fix the imperfection later on. if you think we should merge this change at it is, please go ahead.

nyh:
> see the comment above. i agree it's an improvement overall. and i can always fix the imperfection later on. if you think we should merge this change at it is, please go ahead.

I see you had other requests regarding simplifying the cmake code, so I'll wait for those. Thanks.

cmake/CheckLibc.cmake
a16bitsysop: okay changed it

cmake/CheckLibc.cmake
a16bitsysop: dropped the comment

tchaikov (approved):
lgtm. the test failures should have been addressed by eb37891dab13571742af58f6f06c669f8b9f28ff

avikivity:
@a16bitsysop please rebase and push to get the tests to pass

avikivity:
Patch itself looks fine.

a16bitsysop:
okay, pushed rebase commit

PR #1862, opened 2023-10-08
http/exception: stop using dynamic exception specification

tchaikov:
it has been removed since C++14, see
https://eel.is/c++draft/diff.cpp14.except . so let's replace it with `noexcept` as suggested by Clang.

```
In file included from /home/kefu/dev/seastar/src/http/file_handler.cc:39:
/home/kefu/dev/seastar/include/seastar/http/exception.hh:46:38: warning: dynamic exception specifications are deprecated [-Wdeprecated-dynamic-exception-spec]
    virtual const char* what() const throw () {
                                     ^~~~~~~~
/home/kefu/dev/seastar/include/seastar/http/exception.hh:46:38: note: use 'noexcept' instead
    virtual const char* what() const throw () {
                                     ^~~~~~~~
                                     noexcept
1 warning generated.
```

### DIFF
--- a/include/seastar/http/exception.hh
+++ b/include/seastar/http/exception.hh
@@ -43,7 +43,7 @@ public:
             : _msg(msg), _status(status) {
     }
 
-    virtual const char* what() const throw () {
+    virtual const char* what() const noexcept {
         return _msg.c_str();
     }
 


nyh (approved):

nyh:
Sadly many of the CI runs are failing (recognizing some silly leak...) but this seems to have nothing to do with this specific patch and seems to be happening for other patches as well. So I'll merge despite the CI failure, and we'll need to fix the CI failures separately (maybe we already have a patch for that, I'm behind in my reviews...).

Issue #1863, opened 2023-10-08
enable -Wdeprecated compiler option


tchaikov:
this option is an umbrella option which enables a bunch of different `-Wdeprecated-*` options. for the Clang option, see https://clang.llvm.org/docs/DiagnosticsReference.html#wdeprecated.

these options help us to identify potentially dangerous code or non-standard compliant implementation, which in turn, can also be dangerous in practice.

avikivity:
If we enable it, it should be as a warning, not an error, since newer C++ versions can deprecate some features while they're still supported on others.

PR #1865, opened 2023-10-08
reactor: do not define dtor of kernel_completion

tchaikov:
there is no need to define its destructor explicitly. if we want to prevent applications from instantiate this class, it already has a pure virtual member function, so this never happens.

this silences the warning like:
```
[19/85] Building CXX object CMakeFiles/seastar.dir/src/core/reactor_backend.cc.o
In file included from /home/kefu/dev/seastar/src/core/reactor_backend.cc:51:
In file included from /home/kefu/dev/seastar/src/core/reactor_backend.hh:26:
In file included from /home/kefu/dev/seastar/include/seastar/core/internal/pollable_fd.hh:26:
/home/kefu/dev/seastar/include/seastar/core/internal/io_desc.hh:31:5: warning: definition of implicit copy assignment operator for 'kernel_completion' is deprecated because it has a user-declared destructor [-Wdeprecated-copy-with-dtor]
    ~kernel_completion() = default;
    ^
/home/kefu/dev/seastar/src/core/reactor_backend.cc:70:7: note: in implicit copy assignment operator for 'seastar::kernel_completion' first required here
class pollable_fd_state_completion : public kernel_completion {
      ^
/home/kefu/dev/seastar/src/core/reactor_backend.cc:611:15: note: in implicit move assignment operator for 'seastar::pollable_fd_state_completion' first required here
        *desc = pollable_fd_state_completion{};
              ^
1 warning generated.
```

Refs #1863

### DIFF
--- a/include/seastar/core/internal/io_desc.hh
+++ b/include/seastar/core/internal/io_desc.hh
@@ -27,8 +27,6 @@
 namespace seastar {
 
 class kernel_completion {
-protected:
-    ~kernel_completion() = default;
 public:
     virtual void complete_with(ssize_t res) = 0;
 };


nyh:
Sadly many of the CI runs are failing (recognizing some silly leak...) but this seems to have nothing to do with this specific patch and seems to be happening for other patches as well. So I'll merge despite the CI failure, and we'll need to fix the CI failures separately (maybe we already have a patch for that, I'm behind in my reviews...).

nyh (approved):

tchaikov:
> But what if someone tries to delete a derived class through the base pointer? The derived destructor won't be invoked.
> 


I think, instead of deleting the dtor here, I'm deleting the user defined one, which annoys the compiler. It still implicitly generates the default one for us.

nyh:
> But what if someone tries to delete a derived class through the base pointer? The derived destructor won't be invoked.

A class with virtual functions should also have a *virtual* destructor. The code that @tchaikov removed had a non-virtual destructor, which was wrong, so it was good to delete it. However I agree that maybe the more correct thing to do is to add a **virtual** destructor.  
This would only matter if indeed someone deletes the object through a base-class pointer, which we probably don't do, but maybe it's better for completeness.

avikivity:
A non-virtual protected destructor ensures that the object is not deleted via a pointer to the base.

A public destructor (explicit or implicit) doesn't ensure that the object is not deleted via a pointer to the base.

The patch deleted two things: the destructor definition, and the protected access specifier. The second one is important.

nyh:
> A non-virtual protected destructor ensures that the object is not deleted via a pointer to the base.

This is true. I missed that it was protected. Also, if that was the intention, we should have had a comment saying that - it's too delicate an issue to assume people will understand.

> A public destructor (explicit or implicit) doesn't ensure that the object is not deleted via a pointer to the base.
> 
> The patch deleted two things: the destructor definition, and the protected access specifier. The second one is important.

Ok. I can revert this patch, but @tchaikov maybe you want to produce a followup patch instead that avoids the warning in a different way?

tchaikov:
> A non-virtual protected destructor ensures that the object is not deleted via a pointer to the base.
> 
> A public destructor (explicit or implicit) doesn't ensure that the object is not deleted via a pointer to the base.
> 
> The patch deleted two things: the destructor definition, and the protected access specifier. The second one is important.

@nyh and @avikivity thank you for your inputs, i will create a follow-up change early tomorrow.

tchaikov:
created at #1870

PR #1866, opened 2023-10-08
sstring, util: do not define constexpr static out-of-line

tchaikov:
it was deprecated in C++17, see
https://eel.is/c++draft/depr.static.constexpr, so let's just keep the inline definition.

this silences the following warning from the Clang compiler:
```
FAILED: CMakeFiles/seastar.dir/src/core/memory.cc.o
/home/kefu/.local/bin/clang++ -DBOOST_NO_CXX98_FUNCTION_BASE -DFMT_SHARED -DSEASTAR_API_LEVEL=7 -DSEASTAR_BUILD_SHARED_LIBS -DSEASTAR_DEBUG -DSEASTAR_DEBUG_SHARED_PTR -DSEASTAR_DEFAULT_ALLOCATOR -DSEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_NUMA -DSEASTAR_HAVE_SYSTEMTAP_SDT -DSEASTAR_HAVE_URING -DSEASTAR_LOGGER_TYPE_STDOUT -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_SHUFFLE_TASK_QUEUE -DSEASTAR_SSTRING -DSEASTAR_STRERROR_R_CHAR_P -DSEASTAR_THREAD_STACK_GUARDS -DSEASTAR_TYPE_ERASE_MORE -Dseastar_EXPORTS -I/home/kefu/dev/seastar/include -I/home/kefu/dev/seastar/build/debug/gen/include -I/home/kefu/dev/seastar/src -isystem /home/kefu/dev/seastar/build/debug/_cooking/installed/include -g -std=gnu++20 -fPIC -U_FORTIFY_SOURCE -Wno-error=unused-result "-Wno-error=#warnings" -fstack-clash-protection -UNDEBUG -Wall -Werror -Wdeprecated -Wimplicit-fallthrough -gz -fsanitize=address -fsanitize=undefined -fno-sanitize=vptr -MD -MT CMakeFiles/seastar.dir/src/core/memory.cc.o -MF CMakeFiles/seastar.dir/src/core/memory.cc.o.d -o CMakeFiles/seastar.dir/src/core/memory.cc.o -c /home/kefu/dev/seastar/src/core/memory.cc
[...]

PR #1867, opened 2023-10-08
build: enable -Wdeprecated compiler option

### LINKED ISSUE #1863: enable -Wdeprecated compiler option

tchaikov:
so that for instance, we are aware that which pieces of code are

* using deprecated stuff with "deprecated" attribute attached to it
* using deprecated C++ feature

please note, this will enable a bunch of "-Wdeprecated-*" compiler
warning options, which are previously disabled. and these warning
will be taken as errors with the exception of "deprecated-declarations".
These warnings are from the `[[deprecated]]` attribute.

So this change requires us to address all of the other "-Wdeprecated-*"
warnings in Seastar. but this does not force the applications using
Seastar to do the same, as the "-Wdeprecated" option is a private CXX
flag which is only used when compiling Seastar itself.

Fixes #1863

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -765,7 +765,8 @@ set (Seastar_PRIVATE_CXX_FLAGS
   -Wall
   -Werror
   -Wimplicit-fallthrough
-  -Wno-error=deprecated-declarations)
+  -Wdeprecated
+  -Wno-error=deprecated)
 if (NOT BUILD_SHARED_LIBS)
   list (APPEND Seastar_PRIVATE_CXX_FLAGS -fvisibility=hidden)
 endif ()


tchaikov:
this change depends on the PRs referencing #1863.

tchaikov:
- [ ] #1862
- [ ] #1864
- [ ] #1865
- [ ] #1866

nyh:
> this change depends on the PRs referencing #1863.

Please rebase later to get the CI working.

tchaikov:
> > this change depends on the PRs referencing #1863.
> 
> Please rebase later to get the CI working.

@nyh hi Nadav, thank you for your review. rebased and repushed.

avikivity:
> so that for instance, we are aware that which pieces of code are
> 
> * using deprecated stuff with "deprecated" attribute attached to it
> * using deprecated C++ feature
> 


Suppose C++(N+1) deprecates feature X and introduces feature Y as a replacement.  This means we can't support both C++N and C++(N+1), without using conditional compilation for X and Y.

tchaikov:
> > so that for instance, we are aware that which pieces of code are
> > 
> > * using deprecated stuff with "deprecated" attribute attached to it
> > * using deprecated C++ feature
> 
> Suppose C++(N+1) deprecates feature X and introduces feature Y as a replacement. This means we can't support both C++N and C++(N+1), without using conditional compilation for X and Y.

i see. i just relaxed the requirement by adding `-Wno-error=deprecated`. so all deprecated warnings are tolerated.

FWIW, i also created #1873 to verify that we don't use any deprecated features in Seastar. fortunately, the feature Ys used in the fixes are still supported by C++17.

Issue #1868, opened 2023-10-08
Seastar.unit.rpc leaks


tchaikov:
```
      Start 46: Seastar.unit.rpc
46/74 Test #46: Seastar.unit.rpc ..............................***Failed   56.69 sec
[0/1] cd /home/circleci/project/build/debug/tests/unit && /usr/bin/cmake -E env ASAN_OPTIONS=disable_coredump=0:abort_on_error=1:detect_stack_use_after_return=1:verify_asan_link_order=0 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1 BOOST_TEST_CATCH_SYSTEM_ERRORS=no /home/circleci/project/build/debug/tests/unit/rpc_test -- -c 2
Running 32 test cases...
WARNING: debug mode. Not for benchmarking or production
INFO  2023-10-08 03:52:13,519 seastar - Reactor backend: io_uring
INFO  2023-10-08 03:52:13,520 [shard 0:n/a ] seastar - Perf-based stall detector creation failed (EACCESS), try setting /proc/sys/kernel/perf_event_paranoid to 1 or less to enable kernel backtraces: falling back to posix timer.
random-seed=1391530320
==444==WARNING: ASan doesn't fully support makecontext/swapcontext functions and may produce false positives in some cases!
Reactor stalled for 36 ms on shard 0. Backtrace: 0x529c7a 0x31d74e4 0x31d69ec 0x301da13 0x301850f 0x3017fb7 0x30188d4 0x301ef0a 0x3bcef 0x587f6c 0x571705 0x88ef3a 0x88eedc 0x88ee7c 0x88eda5 0x88ec21 0x3904492 0x38f8414 0x3d7bf58 0x3d7b60d 0x5c11fd 0x5c5bb0 0x90b14f 0x90b0e0 0x90b080 0x90b024 0x90afba 0x90aed9 0x90ad8a 0x90abb8 0x2cf9acd 0x353dcd7
[...]

PR #1871, opened 2023-10-09
treewide: explicitly define ctor and assignment operator

tchaikov:
there are quite a few places where we define a base class with virtual destructor so that we can destroy the instances of its derived classes via a pointer to the base class. despite that the destructor is marked "default", compiler still considers it as a user-provided destructor. so the related guardrails defined by the C++ standard apply: the implicit copy assignment operator and the implicit copy constructor are deprecated for these base classes, because they are considered dangerous. the idea is: now that developer customizes the way how the instance is teared down, the way how the instance is copied/assigned is very likely to be customized as well. see
https://eel.is/c++draft/depr.impldec

to make our intention explicit and to assure the compiler that we are satisified with the default-generated copy constructor and assignement operator, we need to define them explicitly.

but if we define the user-provided copy constructor by "default"ing it explicitly, we prevent the compiler from creating the implicitly default-generated default constructor. so we have to define the default constructor as well, otherwise the derived classes would have trouble to get a default constructor.

this change fixes the leftover of

- 2148122db8f2d5a3b9749500e1442bd16bf09303
- eb3d703a36948af9ea28668fb1f4dc7f23e846a8

this change should address the warning like:
```
include/seastar/util/backtrace.hh:146:5: error: definition of implicit copy assignment operator for 'tasktrace' is deprecated because it has a user-provided destructor [-Werror,-Wdeprecated-copy-with-user-provided-dtor]
    ~tasktrace();
    ^
/home/circleci/project/src/core/memory.cc:858:30: note: in implicit copy assignment operator for 'seastar::tasktrace' first required here
    new_alloc_site.backtrace = get_backtrace();
                             ^
```
and
```
include/seastar/util/backtrace.hh:147:5: error: definition of implicit copy constructor for 'tasktrace' is deprecated because it has a user-provided destructor [-Werror,-Wdeprecated-copy-with-user-provided-dtor]
    ~tasktrace();
    ^
/home/circleci/project/src/core/memory.cc:187:8: note: in implicit copy constructor for 'seastar::tasktrace' first required here
struct allocation_site {
       ^
```

Refs #1868

### DIFF
--- a/include/seastar/core/condition-variable.hh
+++ b/include/seastar/core/condition-variable.hh
@@ -76,6 +76,10 @@ private:
     // the base for queue waiters. looks complicated, but this is
     // to make it transparent once we add non-promise based nodes
     struct waiter : public boost::intrusive::list_base_hook<boost::intrusive::link_mode<boost::intrusive::auto_unlink>> {
+        waiter() = default;
+        waiter(waiter&&) = default;
+        waiter(const waiter&) = delete;
+        waiter& operator=(const waiter&) = delete;
         virtual ~waiter() = default;
         void timeout() noexcept;
 
--- a/include/seastar/core/internal/io_desc.hh
+++ b/include/seastar/core/internal/io_desc.hh
@@ -29,7 +29,8 @@ namespace seastar {
 class kernel_completion {
 protected:
      kernel_completion() = default;
-     kernel_completion(const kernel_completion&) = default;
+     kernel_completion(kernel_completion&&) = default;
+     kernel_completion& operator=(kernel_completion&&) = default;
      ~kernel_completion() = default;
 public:
     virtual void complete_with(ssize_t res) = 0;
--- a/include/seastar/http/file_handler.hh
+++ b/include/seastar/http/file_handler.hh
@@ -51,7 +51,8 @@ public:
      */
     virtual output_stream<char> transform(std::unique_ptr<http::request> req,
             const sstring& extension, output_stream<char>&& s) = 0;
-
+    file_transformer() = default;
+    file_transformer(file_transformer&&) = default;
     virtual ~file_transformer() = default;
 };
 
--- a/include/seastar/json/json_elements.hh
+++ b/include/seastar/json/json_elements.hh
@@ -194,6 +194,9 @@ public:
 
 class jsonable {
 public:
+    jsonable() = default;
+    jsonable(const jsonable&) = default;
+    jsonable& operator=(const jsonable&) = default;
     virtual ~jsonable() = default;
     /**
      * create a foramted string of the object.
--- a/include/seastar/net/stack.hh
+++ b/include/seastar/net/stack.hh
@@ -52,6 +52,9 @@ public:
 
 class socket_impl {
 public:
+    socket_impl() = default;
+    socket_impl(const socket_impl&) = delete;
+    socket_impl(socket_impl&&) = default;
     virtual ~socket_impl() {}
     virtual future<connected_socket> connect(socket_address sa, socket_address local, transport proto = transport::TCP) = 0;
     virtual void set_reuseaddr(bool reuseaddr) = 0;
--- a/include/seastar/util/backtrace.hh
+++ b/include/seastar/util/backtrace.hh
@@ -143,6 +143,8 @@ private:
 public:
     tasktrace() = default;
     tasktrace(simple_backtrace main, vector_type prev, size_t prev_hash, scheduling_group sg) noexcept;
+    tasktrace(const tasktrace&) = default;
+    tasktrace& operator=(const tasktrace&) = default;
     ~tasktrace();
 
     size_t hash() const noexcept { return _hash; }
--- a/tests/unit/distributed_test.cc
+++ b/tests/unit/distributed_test.cc
@@ -135,6 +135,8 @@ SEASTAR_TEST_CASE(test_map_reduce) {
 SEASTAR_TEST_CASE(test_map_reduce_lifetime) {
     struct map {
         bool destroyed = false;
+        map() = default;
+        map(const map&) = default;
         ~map() {
             destroyed = true;
         }
@@ -148,6 +150,9 @@ SEASTAR_TEST_CASE(test_map_reduce_lifetime) {
     struct reduce {
         long& res;
         bool destroyed = false;
+        reduce(long& result)
+            : res{result} {}
+        reduce(const reduce&) = default;
         ~reduce() {
             destroyed = true;
         }
@@ -174,6 +179,8 @@ SEASTAR_TEST_CASE(test_map_reduce_lifetime) {
 SEASTAR_TEST_CASE(test_map_reduce0_lifetime) {
     struct map {
         bool destroyed = false;
+        map() = default;
+        map(const map&) = default;
         ~map() {
             destroyed = true;
         }
@@ -186,6 +193,8 @@ SEASTAR_TEST_CASE(test_map_reduce0_lifetime) {
     };
     struct reduce {
         bool destroyed = false;
+        reduce() = default;
+        reduce(const reduce&) = default;
         ~reduce() {
             destroyed = true;
         }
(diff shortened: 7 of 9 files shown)


include/seastar/core/condition-variable.hh
```diff
@@ -76,6 +76,8 @@ private:
     // the base for queue waiters. looks complicated, but this is
     // to make it transparent once we add non-promise based nodes
     struct waiter : public boost::intrusive::list_base_hook<boost::intrusive::link_mode<boost::intrusive::auto_unlink>> {
+        waiter() = default;
+        waiter(const waiter&) = default;
```
xemul, line 80: It seems to me that waiters are not supposed to be copyable at all. Maybe the copy constructor should be explicitly deleted instead?

tchaikov: indeed, i am adding the defaulted move constructor. it is what we need.

tchaikov:
v2:

- instead of adding a copy ctor for `waiter`, define a move ctor for it. the former would be a bug, but the latter is exactly what we need.

tchaikov:
@xemul hi Pavel, thank you for your review. could you please take another look?

include/seastar/core/condition-variable.hh
```diff
@@ -76,6 +76,10 @@ private:
     // the base for queue waiters. looks complicated, but this is
     // to make it transparent once we add non-promise based nodes
     struct waiter : public boost::intrusive::list_base_hook<boost::intrusive::link_mode<boost::intrusive::auto_unlink>> {
+        waiter() = default;
+        waiter(waiter&&) = default;
+        waiter(const waiter&) = delete;
+        waiter& operator=(const waiter&) = delete;
```
xemul, line 82: Sorry for being very picky, but does definition of the move constructor (even being =default one) make compiler delete the copy- one on its own?

include/seastar/core/internal/io_desc.hh
```diff
@@ -30,6 +30,7 @@ class kernel_completion {
 protected:
      kernel_completion() = default;
      kernel_completion(const kernel_completion&) = default;
+     kernel_completion& operator=(const kernel_completion&) = default;
```
xemul, line 33: completions should be move-able. Won't compiler auto-delete the move constructor for it after it finds the user-defined copy-constructor?

include/seastar/net/stack.hh
```diff
@@ -52,6 +52,8 @@ public:
 
 class socket_impl {
 public:
+    socket_impl() = default;
+    socket_impl(const socket_impl&) = default;
```
xemul, line 56: socket_impl-s are _not_ supposed to be movable and copyable, methinks. These objects are pimpl "backends" for socket-s, and socket-s just keep unique_ptr-s on them. So I'd delete both move- and copy-constructors

(the above statement is not compile-checked nonetheless :) )

include/seastar/core/condition-variable.hh
tchaikov: i think so. i am fine removing this. i made this very change based on your comment of 

> Maybe the copy constructor should be explicitly deleted instead?

include/seastar/core/internal/io_desc.hh
tchaikov: ahh, i think so. fixed.

include/seastar/net/stack.hh
tchaikov: no, it's supposed to be moveable, as we use `do_with()` on it.

tchaikov:
v3:

- dropped the copy ctor of `socket_impl` and `kernel_completion`

include/seastar/net/stack.hh
tchaikov: but yeah, copy ctor is not necessary and not used. dropped.

include/seastar/http/file_handler.hh
```diff
@@ -51,7 +51,8 @@ public:
      */
     virtual output_stream<char> transform(std::unique_ptr<http::request> req,
             const sstring& extension, output_stream<char>&& s) = 0;
-
+    file_transformer() = default;
+    file_transformer(const file_transformer&) = default;
```
xemul, line 55: AFAIU, file_transformers are sugar-objects to push into route's handler. The way routes maintain handlers is weird by itself, but it seems to me that handlers are non-copyable too. Maybe move-able

include/seastar/json/json_elements.hh
```diff
@@ -194,6 +194,9 @@ public:
 
 class jsonable {
 public:
+    jsonable() = default;
+    jsonable(const jsonable&) = default;
+    jsonable& operator=(const jsonable&) = default;
```
xemul, line 199: I know nothing about jsonable thing, but I'd try to be on the safe-side and start with allowing it to be move-able and delete the copy-constructor (explicitly or implicitly) WDYT?

tchaikov: good idea! but tried. failed. it is actually used and required
```
/home/kefu/.local/bin/clang++ -DBOOST_NO_CXX98_FUNCTION_BASE -DFMT_SHARED -DSEASTAR_API_LEVEL=7 -DSEASTAR_BUILD_SHARED_LIBS -DSEASTAR_DEBUG -DSEASTAR_DEBUG_SHARED_PTR -DSEASTAR_DEFAULT_ALLOCATOR -DSEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_NUMA -DSEASTAR_HAVE_SYSTEMTAP_SDT -DSEASTAR_HAVE_URING -DSEASTAR_LOGGER_TYPE_STDOUT -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_SHUFFLE_TASK_QUEUE -DSEASTAR_SSTRING -DSEASTAR_TYPE_ERASE_MORE -I/home/kefu/dev/seastar/apps/httpd -I/home/kefu/dev/seastar/build/debug/apps/httpd -I/home/kefu/dev/seastar/include -I/home/kefu/dev/seastar/build/debug/gen/include -g -std=gnu++20 -UNDEBUG -Wall -Werror -Wimplicit-fallthrough -Wdeprecated -gz -U_FORTIFY_SOURCE -Wno-error=unused-result "-Wno-error=#warnings" -fstack-clash-protection -fsanitize=address -fsanitize=undefined -fno-sanitize=vptr -MD -MT apps/httpd/CMakeFiles/app_httpd.dir/demo.json.cc.o -MF apps/httpd/CMakeFiles/app_httpd.dir/demo.json.cc.o.d -o apps/httpd/CMakeFiles/app_httpd.dir/demo.json.cc.o -c /home/kefu/dev/seastar/build/debug/apps/httpd/demo.json.cc
In file included from /home/kefu/dev/seastar/build/debug/apps/httpd/demo.json.cc:1:
In file included from /home/kefu/dev/seastar/build/debug/apps/httpd/demo.json.hh:19:
/usr/include/boost/range/irange.hpp:211:39: error: call to implicitly-deleted copy constructor of 'seastar::httpd::demo_json::my_object::enum_var_wrapper'
        return integer_range<Integer>(first, last);
                                      ^~~~~
/home/kefu/dev/seastar/build/debug/apps/httpd/demo.json.hh:97:27: note: in instantiation of function template specialization 'boost::irange<seastar::httpd::demo_json::my_object::enum_var_wrapper>' requested here
            return boost::irange(begin(), end());
                          ^
/home/kefu/dev/seastar/build/debug/apps/httpd/demo.json.hh:45:31: note: copy constructor of 'enum_var_wrapper' is implicitly deleted because base class 'json::jsonable' has a deleted copy constructor
    struct enum_var_wrapper : public json::jsonable {
                              ^
/home/kefu/dev/seastar/include/seastar/json/json_elements.hh:198:5: note: copy constructor is implicitly deleted because 'jsonable' has a user-declared move constructor
    jsonable(jsonable&&) = default;
    ^
/usr/include/boost/range/irange.hpp:186:31: note: passing argument to parameter 'first' here
        integer_range(Integer first, Integer last)
                              ^
1 error generated.
```

include/seastar/http/file_handler.hh
tchaikov: thank you for pointing this out! yeah, de-copyablized and moveablized.

tchaikov:
v4:

- trade `file_transformer`'s copy ctor for its move ctor.

PR #1874, opened 2023-10-09
loop: Fix iterator_range_estimate_vector_capacity for random iters

StephanDollberg:
The implementation uses tag dispatch to call out to `std::distance` for
forward iterators and below.

However this doesn't work as expected for random iterators. They will
actually get dispatched to the default template version.

Hence, functions that use `iterator_range_estimate_vector_capacity`
don't pre-reserve anything for `std::vector` or similar classes.

This causes lots of reallocations when using `seastar::when_all` or
similar.

Change the implementation to also call out to std::distance for derived
types.

### DIFF
--- a/include/seastar/core/loop.hh
+++ b/include/seastar/core/loop.hh
@@ -503,17 +503,14 @@ struct has_iterator_category<T, std::void_t<typename std::iterator_traits<T>::it
 template <typename Iterator, typename Sentinel, typename IteratorCategory>
 inline
 size_t
-iterator_range_estimate_vector_capacity(Iterator const&, Sentinel const&, IteratorCategory) {
+iterator_range_estimate_vector_capacity(Iterator begin, Sentinel end, IteratorCategory) {
+    // May be linear time below random_access_iterator_tag, but still better than reallocation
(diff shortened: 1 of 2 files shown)


avikivity:
Thanks a lot for this fix. Is there a way to add a test for the actual bug fixed? Perhaps by wrapping std::vector with some wrapper that counts calls to reserve().

StephanDollberg:
> Thanks a lot for this fix. Is there a way to add a test for the actual bug fixed? Perhaps by wrapping std::vector with some wrapper that counts calls to reserve().

Hmm, the vector is instantiated inside the when_all so a bit hard to override that. 

I could possible do something based on counting the allocations using the allocator stats but that would be a bit hacky?

avikivity:
I think it's okay. You'll have to disable the test if !SEASTAR_ALLOCATOR but it's good not to have this regress.

avikivity:
Oh and you can just test estimate_vector_capacity since it receives the container as a parameter.

StephanDollberg:
> Oh and you can just test estimate_vector_capacity since it receives the container as a parameter.

This I already added a test for or do you mean something else?

Issue #16, opened 2015-01-22
DPDK support TX queue for each cpu irrespective of RX queue type


slivne:
On intel cards we are limited by rss to use 16 rx queues, however we can use additional TX queues - if they are available.

For example on bare metal we can have 128 queues and 24 cpus
- 16 cpus will use physical rx queues and distribute the traffix to the 24 cpus
- 8 cpus will use software rx queues
- 24 cpus will use a physical tx queue

PR #1877, opened 2023-10-13
memory: use `err` for the error string

tchaikov:
in cf0c3ee0df1d64bb6e162d5b47edbba53c62215d, we detected for the GNU variant of `strerror_r()` and falled back to the XSI variant if the GNU version was not available. but we failed to handle the case where `strerror_r()` succeeds in the XSI branch -- we should use the buffer passed to `strerror_r()` for the returned `msg`. and `buf` is not defined there. it was a copy-pasta error.

in this change, the error is corrected by s/err/buf/

### DIFF
--- a/src/core/memory.cc
+++ b/src/core/memory.cc
@@ -1611,7 +1611,7 @@ configure(std::vector<resource::memory> m, bool mbind,
 #ifdef SEASTAR_STRERROR_R_CHAR_P
                 const char *msg = strerror_r(errno, err, sizeof(err));
 #else
-                const char *msg = strerror_r(errno, err, sizeof(err)) ? "unknown error" : buf;
+                const char *msg = strerror_r(errno, err, sizeof(err)) ? "unknown error" : err;
 #endif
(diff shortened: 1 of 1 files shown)


tchaikov:
@a16bitsysop hi Duncan, could you help verify if this change builds?

a16bitsysop:
Sure, must have missed that one.  Will give it a test build

a16bitsysop:
From the latest commit with this patch it build fines and passes the same tests on x86_64 aarch64
Here is the x86_64 pipeline: https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1143105

I am getting a dns error for s390x:
`Test #16: Seastar.unit.dns ..............................***Timeout 300.01 sec`

And 8 test failures for ppc64le.

But both failures happen with the same commit sha they passed on before in the alpine CI, so they can't be related to any change in seastar.

tchaikov:
@a16bitsysop hi Duncan, thank you for testing the change!

Issue #1876, opened 2023-10-11
iotune causes assertion error when the environment does not have enough memory


syuu1228:
I found that iotune causes following assertion error when the environment does not have enough memory:
```
$ iotune --fs-check --evaluation-directory . --default-log-level info
iotune: seastar/src/core/resource.cc:636: seastar::resource::resources seastar::resource::allocate(seastar::resource::configuration &): Assertion `!remain' failed.
Aborting.
Backtrace:
  0x53a4c8
  0x56ea42
  /opt/scylladb/libreloc/libc.so.6+0x3cb1f
  /opt/scylladb/libreloc/libc.so.6+0x8ce5b
  /opt/scylladb/libreloc/libc.so.6+0x3ca75
  /opt/scylladb/libreloc/libc.so.6+0x267fb
  /opt/scylladb/libreloc/libc.so.6+0x2671a
  /opt/scylladb/libreloc/libc.so.6+0x35655
  0x59c990
  0x555de5
  0x4fda88
  0x4fd1e8
  0x4af5ae
  /opt/scylladb/libreloc/libc.so.6+0x2750f
  /opt/scylladb/libreloc/libc.so.6+0x275c8
  0x4ad1f4
Aborted (core dumped)
```

iotune is used from Scylla setup script, and users may not understand what is the problem with this message.
It should have better error message.

syuu1228:
Seems like not just iotune but all seastar based programs, since it is caused at core/resource.cc.
I also got same error from Scylla:
```
Scylla version 5.3.0~rc0-0.20230523.3c3621db072d with build-id 78cb9caaf1cdd0b37c3a38fa83548e6c4e3d4ab4 starting ...
command used: "/opt/scylladb/bin/scylla"
parsed command line options: []
scylla: seastar/src/core/resource.cc:636: seastar::resource::resources seastar::resource::allocate(seastar::resource::configuration &): Assertion `!remain' failed.
Aborting.
Backtrace:
  0x5634538
  0x5668ab2
  /opt/scylladb/libreloc/libc.so.6+0x3cb1f
  /opt/scylladb/libreloc/libc.so.6+0x8ce5b
  /opt/scylladb/libreloc/libc.so.6+0x3ca75
  /opt/scylladb/libreloc/libc.so.6+0x267fb
  /opt/scylladb/libreloc/libc.so.6+0x2671a
  /opt/scylladb/libreloc/libc.so.6+0x35655
  0x58aa210
  0x564fe55
[...]

PR #1880, opened 2023-10-18
resource: move variable into smaller lexical scope

tchaikov:
this series includes two cleanups 

- resource: use structured binding when appropriate
- resource: move variable into smaller lexical scope


tchaikov:
this change was inspired when investigating #1876. it includes #1879 so that the test can pass. will drop it from this PR once #1879 lands.

PR #1652, opened 2023-05-10
Limit the number of http client opened connections

xemul:
In its current shape http::client generates as many connections as it takes to serve all make_request() calls. Eventually the connections settle in the client's internal pool from which client re-uses them, but whenever the pool is empty the new connection is made.

This set allows configuring the limit on the total number of connections -- both active and idle (in pool) -- a client may have. If the pool is empty but there are too many active connections, the new one isn't spawned, instead the caller waits for active one to either get back into the pool, or to get closed.

Also on top of it two stats calls are added to return total number of connections and the amount of idling ones.

PR #1883, opened 2023-10-18
Make http client restart requests

xemul:
There's an issue with maintaining keep-alive sockets by http client.

Some prior art first.

The HTTP client uses seastar connections to send requests over and receive responses back. Typically the connection is a TCP socket (or TLS connection which is a wrapper over TCP socket too). Once request-response cycle is over the connection used for that is kept by a client in a "pool". Making another http request may then pick up the existing connection from the pool thus avoiding the extra latency of establishing new connection. Pool may thus accumulate more than one connection if the client user sends several requests in parallel.

```mermaid
flowchart TD
  A[ Start ] --> B{ pool is not empty }
  B -->| y | C[ pop connection from pool ]
  B -->| n | D[ establish new connection ]
  C --> E[ talk to server ]
  D --> E
  E --> F{ got error? }
  F --> | n | G[ push connection to pool ]
  F --> | y | H[ Done ]
  G --> H
```

HTTP servers may sometimes want to terminate the connections it keeps. This can happen in one of several ways.

The "gentle" way is when server adds the "connection: close" header to its response. In that case client would handle the response and will not put the connection back to pool, but instead would just close it. So next request would either pick some other connection from pool or would need to establish a new one.

Less gentle way a server may terminate a connection is by closing it, so the underlying TCP stack would communicate regular TCP FIN-s. On the client side there's connected_socket::wait_input_shutdown() method that returns a future that gets resolved when kernel terminates the connection by peers request. In case client's connection is kept in pool it will be closed and removed from the pool behind the scenes. Next request won't even notice that -- it will either pick some other connection from pool, or will establish a new one.

Sometimes more unpleasant situation happens. It can be either a race or deliberate server's "abort". In the former case, server closes the connection and TCP starts communicating FIN-s in parallel with client picking up a connection for its new request. In that case, even if kernel resolves "input-shutdown" event described above, due to seastar event loop and batch flusher, client would see that the connection is closed _after_ it had picked it from pool and had chance to put some data into it. In the latter case server closes the connection in the middle of reading the request from the client or even in the middle of writing back the response. This is very unlikely, but still happens from time to time.

Having said the above, the problem -- when user calls `client::make_request()` and client steps on the "unpleasant" server-side connection closure, the request making resolves with exception and user has to do something about it. There are two options here -- handle the exception somehow or ask client to make the same request again (restart the request). This PR suggest that the latter choice can be implemented in the HTTP client code. If user doesn't want to restart, it may ask client not to do it, the new API allows for that.

First (a side node) -- to tell the "server closed its socket" from other errors the exception from scoket IO is checked to be the system error with EPIPE or ECONABORTED code in it. EPIPE is reported when it comes from writing the request, ECONABORTED is reported when it comes from reading the response.

Next, there's only one way to replay the request -- client should get new connection somehow and repeat the request-response cycle. There are two options where to get new connection from -- from pool (if it's there) or establish a new one. While experimenting with HTTP client I noticed that picking up connection from pool often results in several "transport exception"-s in a row, as if server was closing connections in batches. So this PR makes a shortcut for restarts it always establishes new connection to server (which, after request-response cycle, is put back into pool normally).

```mermaid
flowchart TD
  A[ Start ] --> B{ pool is not empty }
  B -->| y | C[ pop connection from pool ]
  B -->| n | D[ establish new connection ]
  C --> E[ talk to server ]
  D --> E
  E --> F{ got error? }
  F --> | n | G[ push connection to pool ]
  F --> | y | J{ should restart }
  J --> | n | H[ Done ]
  J --> | y | D
  G --> H
```

And the last, but not least, restart only happens once, because the chance of observing "transport exception" from newly established connection is considered to be low enough not to care. Respectively, if a request is made over new connection (not over a connection from pool) and "transport exception" pops up it's not restarted.

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -220,7 +220,14 @@ public:
      * may re-use the sockets on its own
      *
      * \param f -- the factory pointer
+     * \param max_connections -- maximum number of connection a client is allowed to maintain
+     * (both active and cached in pool)
      *
+     * The client uses connections provided by factory to send requests over and receive responses
+     * back. Once request-response cycle is over the connection used for that is kept by a client
+     * in a "pool". Making another http request may then pick up the existing connection from the
+     * pool thus avoiding the extra latency of establishing new connection. Pool may thus accumulate
+     * more than one connection if user sends several requests in parallel.
      */
     explicit client(std::unique_ptr<connection_factory> f, unsigned max_connections = default_max_connections);
 
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -126,7 +126,7 @@ public:
     future<> close();
 
 private:
-    future<reply_ptr> do_make_request(request rq);
+    future<reply_ptr> do_make_request(request& rq);
     void setup_request(request& rq);
     future<> send_request_head(const request& rq);
     future<reply_ptr> maybe_wait_for_continue(const request& req);
@@ -166,6 +166,10 @@ public:
  */
 
 class client {
+public:
+    using reply_handler = noncopyable_function<future<>(const reply&, input_stream<char>&& body)>;
+
+private:
     friend class http::internal::client_ref;
     using connections_list_t = bi::list<connection, bi::member_hook<connection, typename connection::hook_t, &connection::_hook>, bi::constant_time_size<false>>;
     static constexpr unsigned default_max_connections = 100;
@@ -191,8 +195,9 @@ class client {
     requires std::invocable<Fn, connection&>
     auto with_new_connection(Fn&& fn);
 
+    future<> do_make_request(connection& con, request& req, reply_handler& handle, std::optional<reply::status_type> expected);
+
 public:
-    using reply_handler = noncopyable_function<future<>(const reply&, input_stream<char>&& body)>;
     /**
      * \brief Construct a simple client
      *
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -168,6 +168,7 @@ public:
 class client {
 public:
     using reply_handler = noncopyable_function<future<>(const reply&, input_stream<char>&& body)>;
+    using retry_requests = bool_class<struct retry_requests_tag>;
 
 private:
     friend class http::internal::client_ref;
@@ -178,6 +179,7 @@ private:
     unsigned _nr_connections = 0;
     unsigned _max_connections;
     unsigned long _total_new_connections = 0;
+    const retry_requests _retry;
     condition_variable _wait_con;
     connections_list_t _pool;
 
@@ -232,14 +234,35 @@ public:
      * \param f -- the factory pointer
      * \param max_connections -- maximum number of connection a client is allowed to maintain
      * (both active and cached in pool)
+     * \param retry -- whether or not to retry requests on connection IO errors
      *
      * The client uses connections provided by factory to send requests over and receive responses
      * back. Once request-response cycle is over the connection used for that is kept by a client
      * in a "pool". Making another http request may then pick up the existing connection from the
      * pool thus avoiding the extra latency of establishing new connection. Pool may thus accumulate
      * more than one connection if user sends several requests in parallel.
+     *
+     * HTTP servers may sometimes want to terminate the connections it keeps. This can happen in
+     * one of several ways.
+     *
+     * The "gentle" way is when server adds the "connection: close" header to its response. In that
+     * case client would handle the response and will just close the connection without putting it
+     * to pool.
+     *
+     * Less gentle way a server may terminate a connection is by closing it, so the underlying TCP
+     * stack would communicate regular TCP FIN-s. If the connection happens to be in pool when it
+     * happens the client would just clean the connection from pool in the background.
+     *
+     * Sometimes the least gentle closing occurs when server closes the connection on the fly and
+     * TCP starts communicating FIN-s in parallel with client using it. In that case, user would
+     * receive exception from the \ref make_request() call and will have to do something about it.
+     * Client provides a transparent way of handling it called "retry".
+     *
+     * When enabled, it makes client catch the transport error, close the broken connection, open
+     * another one and retry the very same request one more time over this new connection. If the
+     * second attempt fails, this error is reported back to user.
      */
-    explicit client(std::unique_ptr<connection_factory> f, unsigned max_connections = default_max_connections);
+    explicit client(std::unique_ptr<connection_factory> f, unsigned max_connections = default_max_connections, retry_requests retry = retry_requests::no);
 
     /**
      * \brief Send the request and handle the response
@@ -253,6 +276,9 @@ public:
      * \param handle -- the response handler
      * \param expected -- the optional expected reply status code, default is std::nullopt
      *
+     * Note that the handle callback should be prepared to be called more than once, because
+     * client may restart the whole request processing in case server closes the connection
+     * in the middle of operation
      */
     future<> make_request(request req, reply_handler handle, std::optional<reply::status_type> expected = std::nullopt);
 
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -180,11 +180,16 @@ class client {
     using connection_ptr = seastar::shared_ptr<connection>;
 
     future<connection_ptr> get_connection();
+    future<connection_ptr> make_connection();
     future<> put_connection(connection_ptr con);
     future<> shrink_connections();
 
     template <std::invocable<connection&> Fn>
     auto with_connection(Fn&& fn);
+
+    template <typename Fn>
+    requires std::invocable<Fn, connection&>
+    auto with_new_connection(Fn&& fn);
 
 public:
     using reply_handler = noncopyable_function<future<>(const reply&, input_stream<char>&& body)>;
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -157,28 +157,28 @@ future<connection::reply_ptr> connection::recv_reply() {
 }
 
 future<connection::reply_ptr> connection::do_make_request(request& req) {
-        setup_request(req);
-        return send_request_head(req).then([this, &req] {
-            return maybe_wait_for_continue(req).then([this, &req] (reply_ptr cont) {
-                if (cont) {
-                    return make_ready_future<reply_ptr>(std::move(cont));
-                }
+    setup_request(req);
+    return send_request_head(req).then([this, &req] {
+        return maybe_wait_for_continue(req).then([this, &req] (reply_ptr cont) {
+            if (cont) {
+                return make_ready_future<reply_ptr>(std::move(cont));
+            }
 
-                return write_body(req).then([this] {
-                    return _write_buf.flush().then([this] {
-                        return recv_reply();
-                    });
+            return write_body(req).then([this] {
+                return _write_buf.flush().then([this] {
+                    return recv_reply();
                 });
             });
         });
+    });
 }
 
 future<reply> connection::make_request(request req) {
-  return do_with(std::move(req), [this] (auto& req) {
-    return do_make_request(req).then([] (reply_ptr rep) {
-        return make_ready_future<reply>(std::move(*rep));
+    return do_with(std::move(req), [this] (auto& req) {
+        return do_make_request(req).then([] (reply_ptr rep) {
+            return make_ready_future<reply>(std::move(*rep));
+        });
     });
-  });
 }
 
 input_stream<char> connection::in(reply& rep) {
@@ -337,26 +337,26 @@ future<> client::make_request(request req, reply_handler handle, std::optional<r
 }
 
 future<> client::do_make_request(connection& con, request& req, reply_handler& handle, std::optional<reply::status_type> expected) {
-        return con.do_make_request(req).then([&con, &handle, expected] (connection::reply_ptr reply) mutable {
-            auto& rep = *reply;
-            if (expected.has_value() && rep._status != expected.value()) {
-                if (!http_log.is_enabled(log_level::debug)) {
-                    return make_exception_future<>(httpd::unexpected_status_error(rep._status));
-                }
-
-                return do_with(con.in(rep), [reply = std::move(reply)] (auto& in) mutable {
-                    return util::read_entire_stream_contiguous(in).then([reply = std::move(reply)] (auto message) {
-                        http_log.debug("request finished with {}: {}", reply->_status, message);
-                        return make_exception_future<>(httpd::unexpected_status_error(reply->_status));
-                    });
-                });
+    return con.do_make_request(req).then([&con, &handle, expected] (connection::reply_ptr reply) mutable {
+        auto& rep = *reply;
+        if (expected.has_value() && rep._status != expected.value()) {
+            if (!http_log.is_enabled(log_level::debug)) {
+                return make_exception_future<>(httpd::unexpected_status_error(rep._status));
             }
 
-            return handle(rep, con.in(rep)).finally([reply = std::move(reply)] {});
-        }).handle_exception([&con] (auto ex) mutable {
-            con._persistent = false;
-            return make_exception_future<>(std::move(ex));
-        });
+            return do_with(con.in(rep), [reply = std::move(reply)] (auto& in) mutable {
+                return util::read_entire_stream_contiguous(in).then([reply = std::move(reply)] (auto message) {
+                    http_log.debug("request finished with {}: {}", reply->_status, message);
+                    return make_exception_future<>(httpd::unexpected_status_error(reply->_status));
+                });
+            });
+        }
+
+        return handle(rep, con.in(rep)).finally([reply = std::move(reply)] {});
+    }).handle_exception([&con] (auto ex) mutable {
+        con._persistent = false;
+        return make_exception_future<>(std::move(ex));
+    });
 }
 
 future<> client::close() {
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -137,7 +137,12 @@ future<connection::reply_ptr> connection::recv_reply() {
         parser.init();
         return _read_buf.consume(parser).then([this, &parser] {
             if (parser.eof()) {
-                throw std::runtime_error("Invalid response");
+                http_log.trace("Parsing response EOFed");
+                throw std::system_error(ECONNABORTED, std::system_category());
+            }
+            if (parser.failed()) {
+                http_log.trace("Parsing response failed");
+                throw std::runtime_error("Invalid http server response");
             }
 
             auto resp = parser.get_parsed_response();
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -256,6 +256,10 @@ future<client::connection_ptr> client::get_connection() {
         });
     }
 
+    return make_connection();
+}
+
+future<client::connection_ptr> client::make_connection() {
     _total_new_connections++;
     return _new_connections->make().then([cr = internal::client_ref(this)] (connected_socket cs) mutable {
         http_log.trace("created new http connection {}", cs.local_address());
@@ -308,6 +312,16 @@ future<> client::set_maximum_connections(unsigned nr) {
 template <std::invocable<connection&> Fn>
 auto client::with_connection(Fn&& fn) {
     return get_connection().then([this, fn = std::move(fn)] (connection_ptr con) mutable {
+        return fn(*con).finally([this, con = std::move(con)] () mutable {
+            return put_connection(std::move(con));
+        });
+    });
+}
+
+template <typename Fn>
+requires std::invocable<Fn, connection&>
+auto client::with_new_connection(Fn&& fn) {
+    return make_connection().then([this, fn = std::move(fn)] (connection_ptr con) mutable {
         return fn(*con).finally([this, con = std::move(con)] () mutable {
             return put_connection(std::move(con));
         });
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -236,9 +236,10 @@ client::client(socket_address addr, shared_ptr<tls::certificate_credentials> cre
 {
 }
 
-client::client(std::unique_ptr<connection_factory> f, unsigned max_connections)
+client::client(std::unique_ptr<connection_factory> f, unsigned max_connections, retry_requests retry)
         : _new_connections(std::move(f))
         , _max_connections(max_connections)
+        , _retry(retry)
 {
 }
 
@@ -330,9 +331,27 @@ auto client::with_new_connection(Fn&& fn) {
 
 future<> client::make_request(request req, reply_handler handle, std::optional<reply::status_type> expected) {
     return do_with(std::move(req), std::move(handle), [this, expected] (request& req, reply_handler& handle) mutable {
-        return with_connection([this, &req, &handle, expected] (connection& con) {
+        auto f = with_connection([this, &req, &handle, expected] (connection& con) {
             return do_make_request(con, req, handle, expected);
         });
+
+        if (_retry) {
+            f = f.handle_exception_type([this, &req, &handle, expected] (const std::system_error& ex) {
+                auto code = ex.code().value();
+                if ((code != EPIPE) && (code != ECONNABORTED)) {
+                    return make_exception_future<>(ex);
+                }
+
+                // The 'con' connection may not yet be freed, so the total connection
+                // count still account for it and with_new_connection() may temporarily
+                // break the limit. That's OK, the 'con' will be closed really soon
+                return with_new_connection([this, &req, &handle, expected] (connection& con) {
+                    return do_make_request(con, req, handle, expected);
+                });
+            });
+        }
+
+        return f;
     });
 }
 
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -156,8 +156,7 @@ future<connection::reply_ptr> connection::recv_reply() {
     });
 }
 
-future<connection::reply_ptr> connection::do_make_request(request req) {
-    return do_with(std::move(req), [this] (auto& req) {
+future<connection::reply_ptr> connection::do_make_request(request& req) {
         setup_request(req);
         return send_request_head(req).then([this, &req] {
             return maybe_wait_for_continue(req).then([this, &req] (reply_ptr cont) {
@@ -172,13 +171,14 @@ future<connection::reply_ptr> connection::do_make_request(request req) {
                 });
             });
         });
-    });
 }
 
 future<reply> connection::make_request(request req) {
-    return do_make_request(std::move(req)).then([] (reply_ptr rep) {
+  return do_with(std::move(req), [this] (auto& req) {
+    return do_make_request(req).then([] (reply_ptr rep) {
         return make_ready_future<reply>(std::move(*rep));
     });
+  });
 }
 
 input_stream<char> connection::in(reply& rep) {
@@ -329,8 +329,15 @@ auto client::with_new_connection(Fn&& fn) {
 }
 
 future<> client::make_request(request req, reply_handler handle, std::optional<reply::status_type> expected) {
-    return with_connection([req = std::move(req), handle = std::move(handle), expected] (connection& con) mutable {
-        return con.do_make_request(std::move(req)).then([&con, expected, handle = std::move(handle)] (connection::reply_ptr reply) mutable {
+    return do_with(std::move(req), std::move(handle), [this, expected] (request& req, reply_handler& handle) mutable {
+        return with_connection([this, &req, &handle, expected] (connection& con) {
+            return do_make_request(con, req, handle, expected);
+        });
+    });
+}
+
+future<> client::do_make_request(connection& con, request& req, reply_handler& handle, std::optional<reply::status_type> expected) {
+        return con.do_make_request(req).then([&con, &handle, expected] (connection::reply_ptr reply) mutable {
             auto& rep = *reply;
             if (expected.has_value() && rep._status != expected.value()) {
                 if (!http_log.is_enabled(log_level::debug)) {
@@ -350,7 +357,6 @@ future<> client::make_request(request req, reply_handler handle, std::optional<r
             con._persistent = false;
             return make_exception_future<>(std::move(ex));
         });
-    });
 }
 
 future<> client::close() {
--- a/tests/unit/loopback_socket.hh
+++ b/tests/unit/loopback_socket.hh
@@ -56,7 +56,7 @@ private:
     queue<temporary_buffer<char>> _q{1};
     loopback_error_injector* _error_injector;
     type _type;
-    promise<> _shutdown;
+    std::optional<promise<>> _shutdown;
 public:
     loopback_buffer(loopback_error_injector* error_injection, type t) : _error_injector(error_injection), _type(t) {}
     future<> push(temporary_buffer<char>&& b) {
@@ -92,16 +92,22 @@ public:
         return _q.pop_eventually();
     }
     void abort() noexcept {
-        if (!_aborted) {
-            // it can be called by both -- reader and writer socket impls
-            _shutdown.set_value();
-        }
+        shutdown();
         _aborted = true;
         _q.abort(std::make_exception_ptr(std::system_error(EPIPE, std::system_category())));
     }
+    void shutdown() noexcept {
+        // it can be called by both -- reader and writer socket impls
+        if (_shutdown.has_value()) {
+            _shutdown->set_value();
+            _shutdown.reset();
+        }
+    }
 
     future<> wait_input_shutdown() {
-        return _shutdown.get_future();
+        assert(!_shutdown.has_value());
+        _shutdown.emplace();
+        return _shutdown->get_future();
     }
 };
 
@@ -160,6 +166,8 @@ public:
     future<> close() override {
         if (!_eof) {
             _buffer->abort();
+        } else {
+            _buffer->shutdown();
         }
         return make_ready_future<>();
     }
--- a/tests/unit/loopback_socket.hh
+++ b/tests/unit/loopback_socket.hh
@@ -69,7 +69,7 @@ public:
                 return make_exception_future<>(std::runtime_error("test injected glitch on send"));
             }
             if (error == loopback_error_injector::error::abort) {
-                shutdown();
+                abort();
                 return make_exception_future<>(std::runtime_error("test injected error on send"));
             }
         }
@@ -85,13 +85,13 @@ public:
                 return make_exception_future<temporary_buffer<char>>(std::runtime_error("test injected glitch on receive"));
             }
             if (error == loopback_error_injector::error::abort) {
-                shutdown();
+                abort();
                 return make_exception_future<temporary_buffer<char>>(std::runtime_error("test injected error on receive"));
             }
         }
         return _q.pop_eventually();
     }
-    void shutdown() noexcept {
+    void abort() noexcept {
         if (!_aborted) {
             // it can be called by both -- reader and writer socket impls
             _shutdown.set_value();
@@ -159,7 +159,7 @@ public:
     }
     future<> close() override {
         if (!_eof) {
-            _buffer->shutdown();
+            _buffer->abort();
         }
         return make_ready_future<>();
     }
@@ -180,11 +180,11 @@ public:
         return data_sink(std::make_unique<loopback_data_sink_impl>(_tx, [this] { shutdown_input(); }));
     }
     void shutdown_input() override {
-        _rx->shutdown();
+        _rx->abort();
     }
     void shutdown_output() override {
         (void)smp::submit_to(_tx->get_owner_shard(), [tx = _tx] {
-            (*tx)->shutdown();
+            (*tx)->abort();
         });
     }
     void set_nodelay(bool nodelay) override {
@@ -330,9 +330,9 @@ public:
             _connect_abort->set_exception(std::make_exception_ptr(std::system_error(ECONNABORTED, std::system_category())));
             _connect_abort = std::nullopt;
         } else {
-            _b1->shutdown();
+            _b1->abort();
             (void)smp::submit_to(_b2.get_owner_shard(), [b2 = std::move(_b2)] {
-                b2->shutdown();
+                b2->abort();
             });
         }
     }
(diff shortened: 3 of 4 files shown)


xemul:
@nyh , @tchaikov this is the restart implementation without magic numbers (but more hairy) as I mentioned [here](https://github.com/scylladb/seastar/pull/1847#discussion_r1361782978). Mind taking a look. Patches 1 through 5 are identical in both PRs.

I do not yet have strong preference to either approach, both help S3 client recover from synchronously dropped connections

xemul:
upd:
- returned back runaway comment

tchaikov:
sorry for the latency, will do in this week.

src/http/client.cc
```diff
@@ -312,6 +312,15 @@ auto client::with_connection(Fn&& fn) {
     });
 }
 
+struct transport_error : public std::exception {
+    unsigned requests_submitted;
+    std::system_error lower;
+    transport_error(unsigned cs, const std::system_error& err) noexcept : requests_submitted(cs), lower(err) {}
```
tchaikov, line 319: i am curious what "cs" is short for?

src/http/client.cc
```diff
@@ -129,7 +129,12 @@ future<connection::reply_ptr> connection::recv_reply() {
         parser.init();
         return _read_buf.consume(parser).then([this, &parser] {
             if (parser.eof()) {
-                throw std::runtime_error("Invalid response");
+                http_log.trace("Parsing responce EOFed");
```
tchaikov, line 132: s/responce/response/

src/http/client.cc
```diff
@@ -312,6 +312,15 @@ auto client::with_connection(Fn&& fn) {
     });
 }
 
+struct transport_error : public std::exception {
+    unsigned requests_submitted;
+    std::system_error lower;
```
tchaikov, line 318: and may i learn what is "lower" is for? lower than what?

src/http/client.cc
```diff
@@ -312,6 +312,15 @@ auto client::with_connection(Fn&& fn) {
     });
 }
 
+struct transport_error : public std::exception {
```
tchaikov, line 316: how about renaming it to `transient_transport_error` ?

include/seastar/http/client.hh
```diff
@@ -180,15 +185,20 @@ class client {
     using connection_ptr = seastar::shared_ptr<connection>;
 
     future<connection_ptr> get_connection();
+    future<connection_ptr> make_connection();
     future<> put_connection(connection_ptr con);
     future<> shrink_connections();
 
+    struct force_new_tag {};
+    using force_new_connection = bool_class<force_new_tag>;
```
tchaikov, line 193:
```suggestion
    using force_new_connection = bool_class<struct force_new_tag>;
```

src/http/client.cc
tchaikov: also, nit, this `exception` is part of the internal implementation of the retry machinery. it will never be caught as an `exception`. so no need to inherit from `exception` and no need to override `what`.

src/http/client.cc
xemul: :facepalm: it must be "reCwuests Submitted"

src/http/client.cc
xemul: Yup. It's the original exception that triggered this one. "Nested" is good term, but it's not really using the nested exceptions from std:: so I named it somehow else

src/http/client.cc
xemul: Not quite. The `do_make_request()` is called two times by `make_request` and the `transport_error` is handled only by the first call. If the second one ends up with the same exception, it will be propagated back to the caller, so it must be the `std::exception` with `what()` too

include/seastar/http/client.hh
xemul: Mm, yes, looks shorter. Is it c++17 compatible?

xemul:
@tchaikov , thank you for the review. I'd also like you to share your thoughts about which restarting mechanism you like more -- this one (that replays the request once over fresh new connection) or the #1847 approach (that implements backoff-retry and that may potentially use other connections from pool)

include/seastar/http/client.hh
tchaikov: see https://eel.is/c++draft/basic.def#4

tchaikov:
sure. will do in this week.

xemul:
@nyh , reminder ping. Please tell which approach sounds more reasonable to you -- this or #1847

tchaikov:
> @tchaikov , thank you for the review. I'd also like you to share your thoughts about which restarting mechanism you like more -- this one (that replays the request once over fresh new connection) or the #1847 approach (that implements backoff-retry and that may potentially use other connections from pool)

i'd prefer retrying with backoff. and i don't have strong preference without good reasoning. i'd go with a `retry_strategy` class, if i were you. and the `retry_strategy` would need to determine 

1. if we should retry given a certain error if we've retried for n times.
2. how long shall we wait before the next retry if we've retried n times.

a default implementation of `retry_strategy` is provided for the backoff and retry policy. so the magic numbers can have a better home without losing the flexibility at the expense of introducing more complexity.

xemul:
The #1847 has the `retry_param` to configure the way requests are retried. But anyway, thank you for the feedback.
@nyh , please, let's decide which way to go -- backoff-retry or force-reconnect

xemul:
@nyh , please, need your feedback on which http client retry approach we adopt -- force-reconnect (this PR) or backoff-retry (#1847 ). Or both :)

nyh:
ּSorry, I'll review it now.

nyh (changes requested):
Looks fairly good, but I have some comments and requests.
One bigger thing that bothers me is why the "automatic restart" of a request is not optional. I'm pretty sure that not all users of Seastar's HTTP client API will always have an idempotent use case - some will want to know when a request failed in the middle and not restart the same request - and instead create a new, potentially different, request.
I think you yourself told me once about a use case where if the request fails after a long time (e.g., imagine uploading a large file failed), the application needs to know this to sign a new request with the new current time - it can't just replay exactly the same request.

src/http/client.cc
```diff
@@ -312,6 +312,15 @@ auto client::with_connection(Fn&& fn) {
...
+    unsigned requests_submitted;
```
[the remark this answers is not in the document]
xemul: The catcher of this exception checks how many requests were submitted via the connection it tried to use. If it was an connection with more than one request, it means that the connection was picked up from cache and it make sense to replay the request via fresh new one. If the connection tried to serve just one request, it means that it was _not_ a connection from pool, but a new connection. In that case it doesn't make sense to retry request via another new connection.

src/http/client.cc
```diff
@@ -323,11 +324,24 @@ struct transport_error : public std::exception {
...
+            if (ex.requests_submitted > 1 /* cached connection */) {
+                // This forces one more connection without checking the limit, but
+                // since the old connection is closing it compensates the increase.
+                //
+                // However, in case new make_request() manages to creep in between
+                // it will temporarily create one extra connection, but it will not
+                // be held in pool anyway. Still, this is unlikely, _nr_connections
+                // decrease will happen in a waiting task and this place is chained
+                // directly after the put_connection()
+                return do_make_request(req, handle, expected, force_new_connection::yes);
```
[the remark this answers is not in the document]
xemul: > But why does it need to be a new connection? Why can't we reuse an existing connection already lying in the pool and it must be a new connection? Why will this new connection "not be held in the pool anyway" - we'll only use it once?

That's _exactly_ the question we're trying to resolve. There are currently two implementation of "retry failed http requests". This PR and the #1847 one. This PR retries request over fresh new connection. This approach is good, because it doesn't have any magic parameters from the caller (well, maybe a boolean flag whether to do it at all or not) -- it _just_ retries the request once over new connection and that's it. THe #1847 implements request retrying in the "backoff-retry" manner. I.e. if a request fails it waits for a while, then picks up another connection (and in that case this another connection can be from pool or can be a new one, depending on the pool state) and tries request again. And does this several times. That approach needs configuration from the caller -- the number of retrying attempts and the pause duration between retries.

> Also, I don't understand the discussion about concurrent requests and limit. If a request is made, and there is no open connection, we have to create a new connection - we can't "limit" this number. 

We don't

> We can limit the number of unused connections we cache, but how is it relevant here?

It just says that when the do_make_request() failed the connection that served that request is still alive at this place and it will only get closed few continuations after that. By the time this connection gets destroyed, request retrying would likely create a _new_ connection _without_ checking the limit thus breaking this limit for a short period of time. It's not critical, just a clarification comment.

include/seastar/http/client.hh
```diff
@@ -189,11 +189,14 @@ private:
...
+
     template <typename Fn>
     SEASTAR_CONCEPT( requires std::invocable<Fn, connection&> )
-    auto with_connection(Fn&& fn);
+    auto with_connection(Fn&& fn, force_new_connection force_new);
```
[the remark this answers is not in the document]
xemul: As I wrote there, there's another approach to restarting requests (#1847 ) in the backoff-retry manner. In it you [asked](https://github.com/scylladb/seastar/pull/1847#discussion_r1349685480) why backoff-retry makes any sense, so here's another approach that just creates a new connection and retries via it.

Both approaches work, but in #1847 retrying a request via cached connection may fail several times in a row, so there's "maximum number of retry attempts" magic configuration constant. This approach doesn't need it. So now we need to decide which way to go -- this or that (or some other third one)

src/http/client.cc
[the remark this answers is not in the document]
xemul: > I still don't understand both points:
> 
>     1. If the connection was from the pool, why do we need "a new connection" and can't take another connection from the pool?

That's what #1847 does -- it picks another connection from whatever it is -- from the pool or makes a new one. That approach also works, but needs configuration from caller -- how many times make retry attempts. This, in turn, is because retrying once over connection from poll doesn't really work -- retrying from the connection from poll fails again quite often and needs 2nd and sometimes 3rd retry.

This PR retries once, because in my experiments retrying over fresh connection works in 99.99% of the cases.

>     2. If this was the first request on this connection, how does it's prove it's not from the pool? Can't it be the first request made on a connection that is / will be in the pool?

The connection usage logic is like this:

```
    con = pop_from_pool()
    if (con == nullptr)
        con = make_new_con()
    con.requests_submitted++
    con.make_request()
    push_to_pool(con)
```

in case con.make_request() throws and con.request == 1 it means that the connection came from make_new_con() and if con.request == 2 it means that the connection came from pool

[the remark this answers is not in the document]
xemul: >  ... and if con.request == 2 it means that the connection came from pool

fix: ... and if con.request >= 2 ...

include/seastar/http/client.hh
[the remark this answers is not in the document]
nyh: Giving two options, this or #1847, is a https://en.wikipedia.org/wiki/False_dilemma - let's try to think if anything needs improving in **this** approach, instead of discussing whether it is better as a whole than some other as-a-whole approach.

The question of whether we need to retry just once or multiple times is, I think, completely orthogonal to where the connection comes from... A retried request may fail for a second time on a recycled connection or a brand-new connection, just the same.

But I think I finally understood now. Perhaps your point is not to complicate this API with the notion of a "pool" at all. So this API only gets a single connection, and if this connection fails, we need to create a new one. It doesn't have any notion of pool it can use.
If this is the explanation, I think the code could be simplified: It doesn't need a "force_new_connection" flag. Instead, before the recursive call to do_make_request() on retry, the caller can open the new connection - and hand the function the fresh connection - the callee doesn't need to care if it was called on a fresh connection or not.

src/http/client.cc
```diff
@@ -304,8 +304,9 @@ future<> client::set_maximum_connections(unsigned nr) {
 
 template <typename Fn>
 SEASTAR_CONCEPT( requires std::invocable<Fn, connection&> )
-auto client::with_connection(Fn&& fn) {
-    return get_connection().then([this, fn = std::move(fn)] (connection_ptr con) mutable {
+auto client::with_connection(Fn&& fn, force_new_connection force_new) {
+    return (__builtin_expect(force_new == force_new_connection::no, true) ?
```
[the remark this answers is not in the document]
xemul: Removed in next version

src/http/client.cc
xemul: Renamed to `cause`

src/http/client.cc
xemul: Renamed :)

include/seastar/http/client.hh
```diff
@@ -80,6 +80,7 @@ class connection : public enable_shared_from_this<connection> {
     // too and thus the connection will be persistent by default. If the server
     // responds with older version, this flag will be dropped (see recv_reply())
     bool _persistent = true;
+    unsigned _requests_submitted = 0;
```
nyh, line 83: I'll be honest - I still don't understand why you need this count. Why doesn't your code know if it's a retry (which requires a new connection) or the original request? Why does it need to keep this flag as a counter inside the connection?

src/http/client.cc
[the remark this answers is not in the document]
nyh: I still don't understand this. After you pop_from_pool(), the connection - no matter where it came from - is no longer in the pool. Why is it important to you to remember if it came from the pool or didn't? How is this knowledge relevant for deciding what to do on retry? On retry you need to close the old connection (it's broken...) and create a new one. Unconditionally... Who cares whether the original connection came from the pool or not?

src/http/client.cc
```diff
@@ -340,6 +340,22 @@ future<> client::make_request(request req, reply_handler handle, reply::status_t
     return do_with(std::move(req), std::move(handle), [this, expected] (request& req, reply_handler& handle) mutable {
         return with_connection([this, &req, &handle, expected] (connection& con) {
             return do_make_request(con, req, handle, expected);
+        }).handle_exception_type([this, &req, &handle, expected] (const transport_error& ex) {
```
nyh, line 343: I still think this auto-retry should be an option, not mandatory (and perhaps not even the default). I'm worried that not all Seastar users will think this behavior is appropriate.

src/http/client.cc
```diff
@@ -340,6 +340,22 @@ future<> client::make_request(request req, reply_handler handle, reply::status_t
...
+            if (ex.requests_submitted == 1) {
+                // It wasn't connection from pool, no need in making new one again
+                return make_exception_future<>(ex.cause);
```
nyh, line 346: Why?
I don't understand :-(
If I understood your code correctly, when the pool was empty you created a new connection and we have requests_submitted==1. But what if this connection breaks in the middle - which is very possible? Why should we not retry it? Why do you only retry if the connection was recycled from the pool?

src/http/client.cc
```diff
@@ -340,6 +340,22 @@ future<> client::make_request(request req, reply_handler handle, reply::status_t
...
+            }
+            // This creates one more connection without checking the limit, but
+            // since the old connection is closing it compensates the increase.
+            //
+            // However, in case new make_request() manages to creep in between
+            // it will temporarily create one extra connection, but it will not
+            // be held in pool anyway. Still, this is unlikely, _nr_connections
+            // decrease will happen in a waiting task and this place is chained
+            // directly after the put_connection()
```
nyh, line 355: I still don't understand what this comment is saying. I think I just didn't understand how your pool works.

include/seastar/http/client.hh
xemul: OK, so first here's basically how a pool works:

```
    make_request(req):
        if (pool.empty())
            con = make_new_connection()
        else
            con = pool.pop_front()

        err = con.sen_request_and_recv_reply(req)

        if (!err)
            pool.push_back(con)
```

In simple words -- after successful serving a request (serving means sending the request, getting the reply and handling it) keep the connection in the pool so that next make_request could re-use already established connection rather than making a new one. It's, of course, a bit more complex as it takes into account keep-alive headers and maximum number of simultaneously opened connections, but basically it's like that.

Now, why do I need the counter. Let's assume that the retry is implemented in a straightforward manner, i.e. -- if con.send_request_and_recv_reply(req) failed just go ahead and make another attempt over new connection

```
    make_request(req):
        if (pool.empty())
            con = make_new_connection()
        else
            con = pool.pop_front()

        err = con.send_request_and_recv_reply(req)
        if (err) {
            con = make_new_connection()
            err = con.send_request_and_recv_reply(req)
        }

        if (!err)
            pool.push_back(con)
```

In that case if the pool was empty on entry, we effectively execute this sequence of steps:

```
            con = make_new_connection()
            err = con.send_request_and_recv_reply(req)
            if (err) {
                con = make_new_connection()
                err = con.send_request_and_recv_reply(req)
            }

```

but this makes little sense, because we already tried a request over new connection once, it failed, what's the point in trying it one more time?

Opposite to that, in case first attempt was made over connection from pool, retrying it has much more sense (more on that below) so the retrying code turns into

```
    make_request(req):
        if (pool.empty()) {
            con = make_new_connection()
            from_pool = false
        } else {
            con = pool.pop_front()
            from_pool = true
        }

        err = con.send_request_and_recv_reply(req)
        if (err && from_pool) {
            con = make_new_connection()
            err = con.send_request_and_recv_reply(req)
        }

        if (!err)
            pool.push_back(con)
```

But in case of http client code, carrying the from_pool boolean all the way from client::make_connection() is a bit problematic, it's simpler just to keep this "bit" on the connection itself. I could make it a `bool was_in_pool_ever` on the connection instead of a counter, but counting the number of requests served is potentially nice metrics, while that bool is not


Now why retrying a request after it failed over new connection doesn't make sense, and retrying it if it failed over connection from pool makes sense.

The thing is that from my experiments I see that AWS S3 server never closes its connections when serving the very 1st request over it. Maybe I ran the test for not long enough, but still -- it simply never happened. Often S3 server behaves in a polite way and sends back `connection: close` header in the reply to indicate that this request is fulfilled, but the server s no longer going to server more requests over it. And sometimes S3 server just closes the connection "in the middle of operation". From the client perspective this "in the middle" is two fold. First, the connection can get closed server-side while client is sending the request (header or body). This happens more often and I can explain it to myself that server just times-out the idling connection (idling, because it may have not yet seen the request on his side). Second, server can close connection while sending back the response. I can't explain this to myself, but it still happens.

src/http/client.cc
xemul: OK, I'll add bool-flag off by default. I don't have any strong arguments for having it always retry

xemul:
upd:
- rebased
- report `bool connection_from_pool` in transport exception
- rephrased the comment describing retry vs connection limits benign race
- made retry/not-retry configurable (off by default)

src/http/client.cc
xemul: Renamed ex.requests_submitted into ex.connection_from_pool and tuned up the comment
Also there's more elaborated description of this place above, so I'm resolving this particular comment

src/http/client.cc
xemul: Added construction-time on/off switch (off by default)

include/seastar/http/client.hh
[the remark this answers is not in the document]
nyh: > > The question of whether we need to retry just once or multiple times is, I think, completely orthogonal to where the connection comes from... A retried request may fail for a second time on a recycled connection or a brand-new connection, just the same.
> 
> That's not so. At least from the statistical point of view. In my testing of S3 client, when a request fails in a connection from pool (i.e. this connection had served at least one successful requests before the failing one) chances that the next pooled connection fails too are pretty high. Neither 0, nor 1, but somewhere in between. Opposite to that, retrying a request over a fresh new connection never failed in my tests.

I think I understand what might explain this statistical observation: Basically what you're saying that if you have in your pool connection A and B to the same server, and connection B broke in the middle, there is a chance that connection A broke too - e.g., perhaps because some load balancer in front of the server crashed and caused many different connections to fail.

I have two problems with that:
1. It's very strange as a general (Seastar) feature, I don't think this mode of failure is something you'd expect to see in all use cases.
2. But, if you do want to handle this case (assume that maybe all prior connections broke), please write about this - in a comment and commit message - explicitly. E.g., maybe something like my paragraph above. Otherwise, it's impossible to guess why you did all this request counting, pool special casing, and so on.

Finally, I have another question: If you have 100 connections in your pool and one broke, you open a new one for the retry because you think you can't trust any of the connections in the pool. But... You still have 99 other connections in the pool, and you don't trust them... So you didn't do the retry on one of these connections, but 99 future requests may try to use these connections and fail. So maybe, if you decided that the pool can't be trusted after a failure, maybe you need to close all of the connections? Or, alternatively, if you think it's too drastic to not reuse all these 99 connections, why was it important to not **retry** on one of these connections? Why are we worried that another retry will fail (and will be re-retried again!), more than we are worried about a new original request failing?


> 
> > But I think I finally understood now. Perhaps your point is not to complicate this API with the notion of a "pool" at all.
> 
> The make_request() API doesn't reveal pool concept to the caller. And the force_new_connection thing is private type of the client class. My goal in this PR is not to complicate the make_request() API with "retry config" thing. If doing backoff-retry retrying once doesn't help in pretty large amount of cases, in my tests I had to take 1, 2 or sometimes 3 retries to make S3 uploading test pass. Retrying infinitely is also not good, so there _must_ be some magic "maximum number of retries" configuration parameter.

I didn't understand - are you saying that this code retries only once, that it retries infinitely, or what?

And please don't refer me to other versions of this PR and what they do - I'm confused already as is :-) Let's discuss the features of **this** PR and whether it does the right thing or not, and if not how it can be improved incrementally, not replacing it by a completely different implementation.

>However, I think you're right that we probably can get rid of the force_new_connection flag.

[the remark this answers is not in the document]
xemul: > I think I understand what might explain this statistical observation: Basically what you're saying that if you have in your pool connection A and B to the same server, and connection B broke in the middle, there is a chance that connection A broke too - e.g., perhaps because some load balancer in front of the server crashed and caused many different connections to fail.

Yes.
And important thing here -- "there's a chance". I'll get back to why this is important a bit later.

> Finally, I have another question: If you have 100 connections in your pool and one broke, you open a new one for the retry because you think you can't trust any of the connections in the pool. But... You still have 99 other connections in the pool, and you don't trust them...

Not quite. As I write above -- "there's a chance", so ... (see below)

> So you didn't do the retry on one of these connections, but 99 future requests may try to use these connections and fail. So maybe, if you decided that the pool can't be trusted after a failure, maybe you need to close all of the connections? Or, alternatively, if you think it's too drastic to not reuse all these 99 connections, why was it important to not retry on one of these connections? Why are we worried that another retry will fail (and will be re-retried again!), more than we are worried about a new original request failing?

So :) I still trust the rest of the pool. In its first version this "restart http requests on transport error" PR had  another retry logic -- it retried requests over connections from pool, but in that case the 1st retry was sometimes enough and sometimes it wasn't. In the end of the day it could take 1, 2, 3, ... $pool_size retries to succeed (I didn't collect the historgram, sorry) and sometimes the pool became empty and a new connection was created. In other words -- retrying a request over some other connection from pool has non-zero chance to fail and non-zero chance to succeed and it takes time to find it out.

> I didn't understand - are you saying that this code retries only once, that it retries infinitely, or what?

_This_ code retries only once over new connection.

Previous version retried several times over connections from pool (or, if the pool exhausts, over new connection). In order not to retry infinitely I broke the look explicitly after N retries. There could be another loop breakage condition -- once the pool is over and we retry over new connection and it fails -- stop retrying.

> And please don't refer me to other versions of this PR and what they do - I'm confused already as is :-)

I can't :) You're asking question like "what if we ..." and my answer is "we already did, and ...". But I'll try to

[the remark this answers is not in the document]
xemul: @nyh , reminder ping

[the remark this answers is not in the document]
nyh: I don't know what to do about this PR (let's try to get a second reviewer). My problem is that I am still not convinced that this approach is more than an ad-hoc band-aid for what you saw in S3 and really belongs in a general-purpose HTTP library in Seastar.
Moreover, I'm not sure I understand exactly what this PR is trying to do (you know the "it's hard to see the forest from the trees" problem) - can you please rewrite the cover letter to include a full explanation of what this implementation does, in a high level? Only now I understood for the first time that "this code retries only once over new connection" - I thought that this can happen more than once. As a mathematician who learned probability, I find it a bit odd (if there's a 1% chance that a request fails, then there's a %0.01 that the second request will fail too!). But I guess I can understand this simplification if it is documented well. But... I still don't understand the entire pool, statics, and so on, business. It sounds to me like like this code takes one connection an returns one connection - perhaps the same one, or perhaps a different one (if it closed a broken connection and opened a new one). Why does it need to keep counts?

Again, I suggest that instead of trying to answer my specific question, please write a longer and self-contained cover letter explaining what this PR does, exactly (by "exactly" I don't mean referring to code, but exactly what the algorithm is doing; By "self-contained" I mean assuming readers don't know exactly what other pieces like the pool do).

And/or, please get a second reviewer (@tchaikov) with a fresh perspective who maybe will understand this PR better.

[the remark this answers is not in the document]
xemul: OK, I wrote a longer story in the description. I cannot avoid mentioning pool in it, because effectively the restart handles an issue that pops up on pooled connections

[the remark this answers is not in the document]
xemul: > As a mathematician who learned probability, I find it a bit odd (if there's a 1% chance that a request fails, then there's a %0.01 that the second request will fail too!)

This implies that the probability of all events is the same, but it's not. The probability of "request fails over connection from pool is 1%", while probability of "request fails over new connection" is different. Tying to ["measure"](https://en.wikipedia.org/wiki/Monte_Carlo_method) it gives 0, which is, of course, a nonsense. Still it appears to be MUCH lower than for "... over connection from pool", provided there's no infrastructure problems

include/seastar/http/client.hh
xemul: > And another physically broken scene is that server returned 503 and this can only be handled in the s3 client. 

Why? Doesn't http client sees 503 reply status?

> When 503 returned, retry should be made on a new connection with a different IP other than from the connection pool. 

Pool doesn't have IP at hand, it just has a bunch of sockets. And it's up to the client's factory whether all those sockets are connected to the same IP or different

xemul: > Therefore, it is crucial to close idle connections promptly. 

Agree, but "promptly" is not something measurable. AFAIK client doesn't have any hint on whether to terminate idling connection. Neither it's good to create new connection for every request, keep-alive should work here too. And once it's ON, server closing its side in the middle of request can happen as I described in the PR top comment

nyh: @xemul 
> but this makes little sense, because we already tried a request over new connection once, it failed, what's the point in trying it one more time?

Well, to me it makes every sense in the world :-( When some hardware in the server cluster fails, whatever connection we have might break in the middle. This can be some long-standing connection we took from a pool, or some shiny-new connection we initiated just 1 second ago - both can break. Why should we **not** retry the shiny-new connection that broke in the middle? It seems completely arbitrary to me.

What worries me about completely arbitrary rules is not the performance impact (like keeping an extra counter) but rather the code complexity - I think you are adding logic that a year from now (even a week from now) developers will not understand, and don't know how to test, and everything will just work the same without this logic, and the developer who wants to change something in the implementation won't know if it it's safe to change or not.

xemul: > Why should we not retry the shiny-new connection that broke in the middle?

Because "shiny-new connection that broke in the middle" can be one of two -- this can be new connection that was created to serve the request because there were no ready to use connections in the pool (I'll call it A below), or this is the new connection that we created to retry current request, that failed once (I'll call it B below). The "why should we not retry" question applies to both. And there are two answers to it.

1. We don't retry the shiny-new connection
2. We do retry the shiny-new connection

You propose to go the 2nd route and in that case we should not only retry case B (above), but case A too, shouldn't we? And if we do, then we can retry infinitely, because A reproduces itself if connection fails. And if that's the case we should at some point stop retrying to avoid infinite loops.


Returning to my initial comment in this thread. We're converting the

```
    make_request(req):
        if (pool.empty())
            con = make_new_connection()
        else
            con = pool.pop_front()

        err = con.sen_request_and_recv_reply(req)

        if (!err)
            pool.push_back(con)
```

code into

```
    make_request(req):
        if (pool.empty()) {
            con = make_new_connection()
            from_pool = false
        } else {
            con = pool.pop_front()
            from_pool = true
        }

        err = con.send_request_and_recv_reply(req)
        if (err && from_pool) {
            con = make_new_connection()
            err = con.send_request_and_recv_reply(req)
        }

        if (!err)
            pool.push_back(con)
```

You find the `from_pool` concept excessive, I don't disagree with that, so we can go with the

```
    make_request(req):
        if (pool.empty())
            con = make_new_connection()
        else
            con = pool.pop_front()

        err = con.send_request_and_recv_reply(req)
        if (err) {
            con = make_new_connection()
            err = con.send_request_and_recv_reply(req)
        }

        if (!err)
            pool.push_back(con)
```

code.

In that case we might also apply "why don't we restart shiny-new connection" logic to the 2nd connection and make it look like

```
    make_request(req):
        attempts = N
again:
        if (pool.empty())
            con = make_new_connection()
        else
            con = pool.pop_front()

        err = con.send_request_and_recv_reply(req)
        if (err && --attempts > 0) {
            goto again;
        }

        if (!err)
            pool.push_back(con)
```

?

nyh: I'm not sure I follow all the different options you're suggesting, or how it relates from what's in the code.
You told me earlier that you don't want to retry several times - just retry ONCE. I am not sure I fully understood why, but that was your decision and it does make some sense if the probability of a single failure is low (and you said this is why you don't want to reuse from the pool and start a new connection).

So, if you only retry ONCE, you can still do this retry no matter where you got the original connection from - if it was a recycled connection from the pool or a new connection because the pool happened to be empty.

Yes, you can start doing retries in a loop, and so on (and probably have some sort of limit on the number or duration of retries, etc.), but I think that's orthogonal to the question of whether or not to do the first (or only) retry.

lovio: > And it's up to the client's factory whether all those sockets are connected to the same IP or different

That's true, but the client cannot decide to make requests on a new connection.

xemul:
upd:
Removed the per-connection tracking of number of requests served and simplified the retry logic -- now it retries over new connection regardless of what. Respectively, there's no longer need in dedicated transport_error exception so it's also removed

include/seastar/http/client.hh
xemul: @nyh , OK, maybe the tracking of per-connection requests and retry based on that was some overkill that doesn't really worth it. I removed it in the updated version, now retry just goes ahead and creates new connection

src/http/client.cc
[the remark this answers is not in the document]
xemul: This logic is removed, so resolving the conversation

include/seastar/http/client.hh
xemul: This logic is removed, resolving the conversation

xemul:
upd:
- rebased

xemul:
upd:
- compilation fix (`SEASTAR_CONCEPT() macro is now gone)

nyh:
> @nyh , let's try to resurrect this PR. I rebased it, tuned up the description and resolved all the controversial comments from the previous iterations ([the last one left](https://github.com/scylladb/seastar/pull/1883#discussion_r1426480383) is about the approach itself)

Ok, I'll try to to fairly re-review this PR from scratch. Earlier in https://github.com/scylladb/seastar/pull/1883#discussion_r1426480383 I suggested that a different reviewer should be brought in, with a fresh view,  but at this point I don't remember what I said and what I didn't like so maybe my own view will be fresh as well ;-)

nyh:
Before reading the code, I want to point out that the new commit message was very helpful, and I feel like I finally understand what you're doing. It even makes a lot of sense if the only kind of error you want to handle is the case of the server **closing an inactive connection** without the client noticing, and in that case you convinced me that what had looked to me like arbitrary rules actually makes sense. Plus I like the fact that there seems to be fewer arbitrary rules here (no "first request on connection" special handling, etc.)

What still bothers me a bit that this is _not really_ the only kind of error that exists in the wild for an HTTP client. In a long request (e.g., some very long upload), there is a non-zero probability that the connection will break in the middle of the connection - not because we missed a closing (the scenario you considered), but really because of a server problem. So if there's a p=0.001% chance that a 1GB transfer will fail in the middle, there is a p^2= 0.000001% chance both the original transfer and its retry will also fail - which is small but not zero, so the decision to retry only once is less justified for this case. But I do agree that if the API can retry once **and** also respond with a failure and let the client retry even more times, that's fine too. Other APIs also have such limitations - for example the boto3 API for DynamoDB retries a temporary failure, by default, 3 times, without returning to the user. Why 3 and not 7? It doesn't matter. If it fails 3 retries it will return the failure to the user, the the user can try 3 more times if he wants.

Now I'll start reading the code.

nyh (changes requested):
Looks good to me, I only have minor requests about some code comments.

Another thing I'm missing is any sort of testing (this PR seems to include none). You added new exception for the failure case, but didn't add a test for it. You're adding an elaborate retry mechanism, but don't have a test for it.
If you believe it's too hard to test in a unit test (this is not good, but sometimes it's the honest truth), I think you'll at least need to explain how you tested this new feature manually. The code looks correct to me, but who knows, does it actually work when encountering an error, or there is something missing that is hard to catch in a code review?

xemul:
upd:
- added tests for incorrect server response and incomplete server response
- added documentation to newly introduced "retry" option
- added test for request retrying on server transport error

src/http/client.cc
```diff
@@ -137,7 +137,12 @@ future<connection::reply_ptr> connection::recv_reply() {
         parser.init();
         return _read_buf.consume(parser).then([this, &parser] {
             if (parser.eof()) {
-                throw std::runtime_error("Invalid response");
+                http_log.trace("Parsing response EOFed");
+                throw std::system_error(ECONNABORTED, std::system_category());
+            }
+            if (parser.failed()) {
+                http_log.trace("Parsing response failed");
+                throw std::runtime_error("Invalid http server response");
```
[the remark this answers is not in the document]
xemul: Fixed and added tests

include/seastar/http/client.hh
```diff
@@ -232,7 +234,7 @@ public:
      * \param f -- the factory pointer
      *
      */
-    explicit client(std::unique_ptr<connection_factory> f, unsigned max_connections = default_max_connections);
+    explicit client(std::unique_ptr<connection_factory> f, unsigned max_connections = default_max_connections, retry_requests retry = retry_requests::no);
```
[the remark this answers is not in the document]
xemul: OK, copied the essential part of the PR description and added to code comment. Also added the documentation for pre-existing `max_connections` argument that happened not to have one

xemul:
upd:
- fix loopback socket double-end shutdown

include/seastar/http/client.hh
```diff
@@ -220,7 +220,14 @@ public:
      * may re-use the sockets on its own
      *
      * \param f -- the factory pointer
+     * \param max_connections -- maximum number of connection a client is allowed to maintain
```
nyh, line 235: I don't understand why we wanted such a feature... What happens if the user starts another request? It waits until one of the older request finishes? Why?

Anyway, you're just documenting a feature we already have, so ok.

I can understand a maximum number of connections **cached** in the pool, that's useful. But why do we want to artificially limit the total number of connections?

Although I guess you're just documenting this now,

nyh:
The new version looks really good (and well-tested), thanks. I just wrote a few comments for your consideration (I tentatively "approved" it).

tests/unit/loopback_socket.hh
```diff
@@ -85,13 +85,13 @@ public:
...
             if (error == loopback_error_injector::error::abort) {
-                shutdown();
+                abort();
                 return make_exception_future<temporary_buffer<char>>(std::runtime_error("test injected error on receive"));
             }
         }
         return _q.pop_eventually();
     }
-    void shutdown() noexcept {
+    void abort() noexcept {
```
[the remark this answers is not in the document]
xemul: It's not public API, it's a helper class for tests :)
Yes, this shutdown is not quite shutdown(2) :(

tests/unit/loopback_socket.hh
```diff
@@ -92,16 +92,22 @@ public:
...
     void abort() noexcept {
-        if (!_aborted) {
-            // it can be called by both -- reader and writer socket impls
-            _shutdown.set_value();
-        }
+        shutdown();
         _aborted = true;
         _q.abort(std::make_exception_ptr(std::system_error(EPIPE, std::system_category())));
     }
+    void shutdown() noexcept {
```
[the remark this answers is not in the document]
xemul: Well, yes, I'm introducing another shutdown(). Old shutdown (that got renamed to abort) now calls new shutdown _and_ aborts the inner queue

tests/unit/loopback_socket.hh
```diff
@@ -92,16 +92,22 @@ public:
...
+        // it can be called by both -- reader and writer socket impls
+        if (_shutdown.has_value()) {
+            _shutdown->set_value();
+            _shutdown.reset();
+        }
+    }
 
     future<> wait_input_shutdown() {
-        return _shutdown.get_future();
+        assert(!_shutdown.has_value());
```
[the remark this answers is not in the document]
xemul: Well, I'd prefer assert here. It's for tests only, and no tests are supposed to call wair_input_shutdown() twice. It's like resolving a promise<> two times in a row, we don't have on_internal_error() there

include/seastar/http/client.hh
xemul: > What happens if the user starts another request? It waits until one of the older request finishes? Why?

If `max_connections` is already serving other requests (and pool is empty) then yes. Because this knob limits the number of _sockets_ opened simultaneously. Limiting only the number of connections in pool is also useful, but I think concurrency control is also required for active connections (even though they are supposedly short-living)

xemul: And yes, it's the pre-existing feature from #1652  (you approved it back then :wink:  )

tests/unit/loopback_socket.hh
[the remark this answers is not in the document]
nyh: Oh, I missed the directory path which clearly says "tests" :-)

tests/unit/loopback_socket.hh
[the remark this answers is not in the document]
nyh: Again I missed that it's a test :-) Assert is perfectly fine.

PR #1884, opened 2023-10-18
circleci: use conditional steps

tchaikov:
it seems that the checkheaders target is never built even if the parameters.mode is "dev". so let's use the conditional step instead to conditional built "checkheaders".


tchaikov:
tested at https://app.circleci.com/pipelines/github/tchaikov/seastar/874/workflows/0bcc6f8e-74af-4d8a-ba0f-cd648c0b357a/jobs/10497

PR #1885, opened 2023-10-19
reactor: use submit_and_wait_timeout if possible

tchaikov:
to save the overhead of an extra syscall.

tested using http + wrk:

the httpd is started using
```
./build/release/apps/httpd/httpd --cpuset 0  --smp 1
```
and wrk is started using
```
taskset --cpu-list 1-15 ./wrk -t 10 -d 10 -c 20  http://localhost:10000
```
#### before this change:
```
Running 10s test @ http://localhost:10000
  10 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   136.70us  106.61us   3.93ms   97.60%
    Req/Sec    15.74k     1.29k   17.78k    70.89%
  1582180 requests in 10.10s, 206.72MB read
Requests/sec: 156651.55
Transfer/sec:     20.47MB
```
#### after this change:
```
Running 10s test @ http://localhost:10000
  10 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   142.00us  121.43us   2.77ms   96.71%
    Req/Sec    15.64k     1.44k   18.50k    75.42%
  1569784 requests in 10.10s, 205.10MB read
Requests/sec: 155432.98
Transfer/sec:     20.31MB

due to the high stdev, i am not sure the test result is statistically significant.

### CODE REVIEW

File: src/core/reactor_backend.cc [hunk 1]
```diff
@@ -1559,16 +1559,16 @@ class reactor_backend_uring final : public reactor_backend {
     virtual void wait_and_process_events(const sigset_t* active_sigmask) override {
         _smp_wakeup_completion.maybe_rearm(*this);
         _hrtimer_completion.maybe_rearm(*this);
-        ::io_uring_submit(&_uring);
         bool did_work = false;
         did_work |= _preempt_io_context.service_preempting_io();
         did_work |= std::exchange(_did_work_while_getting_sqe, false);
         if (did_work) {
+            ::io_uring_submit(&_uring);
```
avikivity, line 1566: error handling?

src/core/reactor_backend.cc
```diff
@@ -1559,16 +1559,16 @@ class reactor_backend_uring final : public reactor_backend {
...
             return;
         }
         struct ::io_uring_cqe* cqe = nullptr;
         sigset_t sigs = *active_sigmask; // io_uring_wait_cqes() wants non-const
-        auto r = ::io_uring_wait_cqes(&_uring, &cqe, 1, nullptr, &sigs);
+        auto r = ::io_uring_submit_and_wait_timeout(&_uring, &cqe, 1, nullptr, &sigs);
```
avikivity, line 1571: This returns -ETIME on timeout, but you don't handle it. How did it work?

src/core/reactor_backend.cc
avikivity: ah, there wasn't before.

avikivity: how can you tell what io_uring_submit does? the documentation just says it submits events.

src/core/reactor_backend.cc
tchaikov: `-ETIME` is returned only if a `ts` is specified. see https://github.com/axboe/liburing/blob/45ed5c68150e1438b6145e3390af2c3b17e90f4a/src/queue.c#L349 , where `__io_uring_get_cqe()` gets called. and `__io_uring_get_cqe()` passes a `get_data` instance to [`_io_uring_get_cqe()`](https://github.com/axboe/liburing/blob/45ed5c68150e1438b6145e3390af2c3b17e90f4a/src/queue.c#L62) without setting its `has_ts`, see https://github.com/axboe/liburing/blob/45ed5c68150e1438b6145e3390af2c3b17e90f4a/src/queue.c#L135 . and `-ETIME` is returned only if `data->has_ts` is set. see https://github.com/axboe/liburing/blob/45ed5c68150e1438b6145e3390af2c3b17e90f4a/src/queue.c#L107 . and this function calls into `io_uring_enter2` syscall, which is an alias of `io_uring_enter`. but this syscall does not return `-ETIME`. see https://github.com/axboe/liburing/blob/master/man/io_uring_enter.2

src/core/reactor_backend.cc
tchaikov: > ah, there wasn't before.

will add the error handling in a separated PR.

tchaikov: > how can you tell what io_uring_submit does? the documentation just says it submits events.

i don't understand your question. is it related to this change? or do you propose that we should *only* handle the events instead of trying to submit the requests as well. as this function is called when the reactor has no more work to do, and is going to sleep / sleeping.

tchaikov: > > ah, there wasn't before.
> 
> will add the error handling in a separated PR.

#1888

PR #1886, opened 2023-10-19
reactor: merge pollfn on I/O paths into a single one

tchaikov:
this change consolidates the 3 pollfn classes into a single one.
this allows the backend to customize the way how the requests
are submitted, and how responses events are collected. for instance,
io_uring backend can submit and wait for events with a single syscall,
so a single place to do polling would make this possible.

in a follow-up change, we will move the part in the poller closer to
backend into backend.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1784,9 +1784,26 @@ steady_clock_type::time_point reactor::next_pending_aio() const noexcept {
     return next;
 }
 
-bool
-reactor::reap_kernel_completions() {
-    return _backend->reap_kernel_completions();
+bool reactor::poll_all_events() {
+    bool work = false;
+    work |= _backend->reap_kernel_completions();
+    work |= flush_pending_aio();
+#ifndef HAVE_OSV
+    work |= _backend->kernel_submit_work();
+#endif
+    work |= _backend->reap_kernel_completions();
+    return work;
+}
+
+bool reactor::poll_any_event() {
+    return (
+        // actually performs work, but triggers no user continuations, so okay
+        _backend->reap_kernel_completions() ||
+        flush_pending_aio() ||
+#ifndef HAVE_OSV
+        _backend->kernel_submit_work() ||
+#endif
+        _backend->reap_kernel_completions());
 }
 
 #if SEASTAR_API_LEVEL < 7
@@ -2797,24 +2814,10 @@ class reactor::kernel_events_pollfn final : public reactor::pollfn {
 public:
     kernel_events_pollfn(reactor& r) : _r(r) {}
     bool poll() final {
-        bool work = false;
-        work |= _r.reap_kernel_completions();
-        work |= _r.flush_pending_aio();
-#ifndef HAVE_OSV
-        work |= _r._backend->kernel_submit_work();
-#endif
-        work |= _r.reap_kernel_completions();
-        return work;
+        return _r.poll_all_events();
     }
     bool pure_poll() final {
-       return (
-           // actually performs work, but triggers no user continuations, so okay
-           _r.reap_kernel_completions() ||
-           _r.flush_pending_aio() ||
-#ifndef HAVE_OSV
-           _r._backend->kernel_submit_work() ||
-#endif
-           _r.reap_kernel_completions());
+        return _r.poll_any_event();
     }
     bool try_enter_interrupt_mode() final {
         if (!_r._backend->kernel_events_can_sleep()) {
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2784,8 +2784,7 @@ class reactor::kernel_events_pollfn final : public reactor::pollfn {
     // Wake-up the reactor with highres timer when the io-queue
     // decides to delay dispatching until some time point in
     // the future
-    bool _armed = false;
-    timer<> _nearest_wakeup { [this] { _armed = false; } };
+    timer<> _nearest_wakeup { [] { } };
     bool maybe_set_alarm_for_aio() {
         auto next = _r.next_pending_aio();
         auto now = steady_clock_type::now();
@@ -2793,7 +2792,6 @@ class reactor::kernel_events_pollfn final : public reactor::pollfn {
             return false;
         }
         _nearest_wakeup.arm(next);
-        _armed = true;
         return true;
     }
 public:
@@ -2834,9 +2832,8 @@ public:
     }
 
     void exit_interrupt_mode() final {
-        if (_armed) {
+        if (_nearest_wakeup.armed()) {
             _nearest_wakeup.cancel();
-            _armed = false;
         }
     }
 };
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2741,19 +2741,6 @@ reactor::do_check_lowres_timers() const noexcept {
     return lowres_clock::now() > _lowres_next_timeout;
 }
 
-#ifndef HAVE_OSV
-
-class reactor::kernel_submit_work_pollfn final : public simple_pollfn<true> {
-    reactor& _r;
-public:
-    kernel_submit_work_pollfn(reactor& r) : _r(r) {}
-    virtual bool poll() override final {
-        return _r._backend->kernel_submit_work();
-    }
-};
-
-#endif
-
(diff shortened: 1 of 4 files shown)


avikivity:
```c++
    poller reap_kernel_completions_poller(std::make_unique<reap_kernel_completions_pollfn>(*this));
    poller io_queue_submission_poller(std::make_unique<io_queue_submission_pollfn>(*this));
    poller kernel_submit_work_poller(std::make_unique<kernel_submit_work_pollfn>(*this));
    poller final_real_kernel_completions_poller(std::make_unique<reap_kernel_completions_pollfn>(*this));
```

Perhaps we should merge those pollers. They made sense when disk and networking were separated, but now they are not.

So instead of bending io_uring to the old model, let's bend the old model to io_uring. And have just one call into the backend for all of them (which translates to one call into io_uring_enter for io_uring, and multiple io_submit/epoll_whatever for the others).

src/core/reactor.cc
```diff
@@ -2834,7 +2796,42 @@ class reactor::io_queue_submission_pollfn final : public reactor::pollfn {
...
+    kernel_events_pollfn(reactor& r) : _r(r) {}
+    bool poll() final {
+        bool work = false;
+        work |= _r.reap_kernel_completions();
+        work |= _r.flush_pending_aio();
+#ifndef HAVE_OSV
+        work |= _r._backend->kernel_submit_work();
+#endif
+        work |= _r.reap_kernel_completions();
+        return work;
```
avikivity, line 2809: We should merge these into a single call, but better to wait with that until later and proceed in small steps.

tchaikov:
v2:

- merge 3 pollerfn into a single one.

@avikivity indeed. that's even better.

src/core/reactor.cc
```diff
@@ -2834,7 +2796,42 @@ class reactor::io_queue_submission_pollfn final : public reactor::pollfn {
...
+    }
+    bool try_enter_interrupt_mode() final {
+        if (!_r._backend->kernel_events_can_sleep()) {
+            return false;
+        }
+        if (!maybe_set_alarm_for_aio()) {
+            return false;
+        }
+        if (!_r._backend->kernel_events_can_sleep()) {
+            return false;
```
avikivity, line 2829: You have to undo maybe_set_alarm, no?

avikivity (commented):
I'd like to see tests for all backends.

avikivity:
> I'd like to see tests for all backends.

I mean results, not new test cases.

src/core/reactor.cc
tchaikov: yes. added a `defer` and repushed.

tchaikov:
> > I'd like to see tests for all backends.
> 
> I mean results, not new test cases.

i will test locally and paste the results. the CI is now using io_uring backend. probably we should paramterize the backend in our CI with for instance the release + clang only, to avoid the combination explosion.

tchaikov:
v3:

- unset the alarm for waking up reactor if fails to enter interrupt mode.

tchaikov:
### `--reactor-backend linux-aio`
```
...
FAILED: tests/unit/CMakeFiles/test_unit_dns_run /var/ssd/seastar/build/debug/tests/unit/CMakeFiles/test_unit_dns_run
cd /var/ssd/seastar/build/debug/tests/unit && /usr/bin/cmake -E env ASAN_OPTIONS=disable_coredump=0:abort_on_error=1:detect_stack_use_after_return=1:verify_asan_link_order=0 UBSAN_OPTIONS=halt_on_error=1:abort_on_e
rror=1 BOOST_TEST_CATCH_SYSTEM_ERRORS=no /var/ssd/seastar/build/debug/tests/unit/dns_test -- -c 2 --reactor-backend linux-aio
...
99% tests passed, 1 tests failed out of 74

Total Test time (real) = 430.70 sec

The following tests FAILED:
         18 - Seastar.unit.dns (Failed)
```

### `--reactor-backend epoll`
```
...
FAILED: tests/unit/CMakeFiles/test_unit_dns_run /var/ssd/seastar/build/debug/tests/unit/CMakeFiles/test_unit_dns_run
cd /var/ssd/seastar/build/debug/tests/unit && /usr/bin/cmake -E env ASAN_OPTIONS=disable_coredump=0:abort_on_error=1:detect_stack_use_after_return=1:verify_asan_link_order=0 UBSAN_OPTIONS=halt_on_error=1:abort_on_e
rror=1 BOOST_TEST_CATCH_SYSTEM_ERRORS=no /var/ssd/seastar/build/debug/tests/unit/dns_test -- -c 2 --reactor-backend epoll
...
99% tests passed, 1 tests failed out of 74

Total Test time (real) = 424.96 sec

The following tests FAILED:
         18 - Seastar.unit.dns (Failed)
```

### `--reactor-backend io_uring`
```
99% tests passed, 1 tests failed out of 74

Total Test time (real) = 411.41 sec

The following tests FAILED:
         18 - Seastar.unit.dns (Failed)
```
please note, due to my spotty network environment. i am not able to access `www.google.com`, so the DNS test is failing in my environment.

tchaikov:
> probably we should paramterize the backend in our CI with for instance the release + clang only, to avoid the combination explosion.

just took a look it seems non-trivial -- i don't want recompile the tests to achieve this. fun fact, CMake even has an [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/20470) for this. but the way how we add a test is somehow more convoluted.

tchaikov:
@avikivity ☝️

avikivity:
> > probably we should paramterize the backend in our CI with for instance the release + clang only, to avoid the combination explosion.
> 
> just took a look at this. it seems non-trivial -- i don't want recompile the tests to achieve this. fun fact, CMake even has an [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/20470) for this. but the way how we add a test is somehow more convoluted.

I don't think it's so important - changes to the backends are rare enough, and usually bugs are too subtle for unit tests.

src/core/reactor.cc
```diff
@@ -1800,10 +1797,7 @@ bool reactor::poll_any_event() {
         // actually performs work, but triggers no user continuations, so okay
         _backend->reap_kernel_completions() ||
         flush_pending_aio() ||
-#ifndef HAVE_OSV
-        _backend->kernel_submit_work() ||
-#endif
-        _backend->reap_kernel_completions());
+        _backend->submit_and_get_completions());
```
avikivity, line 1800: Well this moves in a baby step, since we can't eliminate reap_kernel_completions as it has another call site.

I suppose it can be further merged later, this is already a big change.

tchaikov:
i experimented a little bit with `io_uring_submit_and_get_events` and `io_uring_submit_and_wait` last weekend.

- `io_uring_submit_and_get_events` helps with the case when CQ entries overflows (can be checked with `io_uring_cq_has_overflow`: currently, we use `s_queue_len` = 200 for SQ entries, and CQ entries is twice as many as SQ entries by default, so it is 400. so far, using `io_uring_submit_and_get_events` does not improve the latency significantly. but i am using a single machine for testing. so it might not be able to be representative the real world setup.
- the idea of  `io_uring_submit_and_wait` is to wait for at least one CQ entry, but in our case, unless we are sure that we don't have **anything** pending tasks in the queue, we cannot afford to **wait**. we would need to differentiate this by checking  `have_more_tasks()` though.
- io_uring_submit_and_wait_timeout: i haven't got a chance to test it.

src/core/reactor.cc
tchaikov: yeah, let's leave it for another change. so far, the goal is to push the implementation of poller down to `reactor_backend`. this will allow us to either customize them or further merge them.

but thinking forward, i am not quite if we can merge `reap_kernel_comletions()` with `submit_and_get_completions()`, as `flush_pending_aio()` is a oddball. it mutates `reactor::_io_queues`, which does not belong to `reactor_backend`. and we cannot reorder these three steps per the comment at https://github.com/scylladb/seastar/blob/17183ed4e475849e5c624c876cf37f6eabc64220/src/core/reactor.cc#L3186

tchaikov:
@avikivity ping? at this moment this change is a refactory, it removes the unnecessary abstraction -- multiple poolers, and consolidate them into a single one.

tchaikov:
@avikivity ping?

PR #1889, opened 2023-10-20
condition-variable: add support for `abort_source`

kbr-scylla:
Add variants of `wait` and `when` member functions that accept an
`abort_source`, similar to the `timeout` variants.

Extend tests.


kbr-scylla:
```
/home/circleci/project/include/seastar/core/condition-variable.hh:304:37: error: no member named 'bind_front' in namespace 'std'
        w->_sub = as.subscribe(std::bind_front(&waiter::abort, w.get()));
```
huh?

Hm, I guess this is a C++20 feature...
But a `std::bind` version will not be `nothrow_invocable` which is required by `abort_source::subscribe` :(

Guess I'll have to use a lambda

michoecho (changes requested):

include/seastar/core/condition-variable.hh
```diff
@@ -58,6 +59,14 @@ public:
     virtual const char* what() const noexcept;
 };
 
+/// Exception thrown when wait() operation is aborted
+/// \ref condition_variable::wait(abort_source&).
+class condition_variable_aborted : public std::exception {
```
gleb-cloudius, line 64: Lets inherit from `abort_requested_exception` like we already do for some other abort exceptions.

gleb-cloudius (changes requested):
Left a comment.

elcallio (approved):
Agree with @gleb-cloudius and @michoecho , other than that it looks fine.

include/seastar/core/condition-variable.hh
```diff
@@ -58,6 +59,14 @@ public:
...
+public:
+    /// Reports the exception reason.
+    virtual const char* what() const noexcept;
```
tchaikov, line 67: nit, instead of using `virtual` i'd suggest use `override`, like
```c++
const char* what() const noexcept override;
```

tests/unit/condition_variable_test.cc
```diff
@@ -128,6 +129,40 @@ SEASTAR_THREAD_TEST_CASE(test_condition_variable_timeout) {
...
 
+SEASTAR_THREAD_TEST_CASE(test_condition_variable_abort) {
+    abort_source as;
+    condition_variable cv1;
+
+    {
+        auto f = cv1.wait(as);
+        cv1.signal();
+        sleep(10ms).get();
+        BOOST_REQUIRE_EQUAL(f.available(), true);
```
tchaikov, line 140: nit, how about just put `BOOST_REQUIRE(f.available())` ?

tests/unit/condition_variable_test.cc
```diff
@@ -128,6 +129,40 @@ SEASTAR_THREAD_TEST_CASE(test_condition_variable_timeout) {
...
+        f.get();
+    }
+
+    auto f1 = cv1.wait(as);
+    BOOST_REQUIRE_EQUAL(f1.available(), false);
```
tchaikov, line 145: nit, how about just put `BOOST_REQUIRE(!f1.available())` ?

include/seastar/core/condition-variable.hh
```diff
@@ -159,6 +169,31 @@ private:
...
+            awaiter::signal();
+        }
+        void set_exception(std::exception_ptr ep) noexcept override {
+            _sub = std::nullopt;
+            awaiter::set_exception(std::move(ep));
+        }
+        auto operator co_await() {
+            if (!_sub) {
+                return ::seastar::internal::awaiter<false, void>(
+                        make_exception_future<>(std::make_exception_ptr(condition_variable_aborted())));
```
tchaikov, line 191: could also just put
```c++
eturn ::seastar::internal::awaiter<false, void>(
                        make_exception_future<>(condition_variable_aborted()));
```

as `std::make_exception_ptr()` has an templated overload.

kbr-scylla:
v2:
- use `this->abort()` in `abortable_awaiter`'s `abort_source` subscription
- remove redundant `std::make_exception_ptr`
- inherit `condition_variable_aborted` from `abort_requested_exception`
- use `override` instead of `virtual` (added new refactoring commit so the code is consistent)
- use `BOOST_REQUIRE` instead of `BOOST_REQUIRE_EQUAL` (added new refactoring commit so the code is consistent)


range-diff:
```
1:  ea7edbe8 = 1:  ea7edbe8 condition-variable: make `waiter`s non-moveable
2:  3937af49 = 2:  3937af49 condition_variable_test: use Boost macros for exceptions
-:  -------- > 3:  ff723e99 condition_variable_test: use `BOOST_REQUIRE` instead of `BOOST_REQUIRE_EQUAL`
-:  -------- > 4:  4ba510eb condition-variable: use `override` instead of `virtual`
3:  fc4f70bc ! 5:  4a41ebf4 condition-variable: add support for `abort_source`
    @@ include/seastar/core/condition-variable.hh
      #   include <seastar/core/coroutine.hh>
     @@ include/seastar/core/condition-variable.hh: class condition_variable_timed_out : public std::exception {
    -     virtual const char* what() const noexcept;
    +     const char* what() const noexcept override;
      };
      
     +/// Exception thrown when wait() operation is aborted
     +/// \ref condition_variable::wait(abort_source&).
    -+class condition_variable_aborted : public std::exception {
    ++class condition_variable_aborted : public abort_requested_exception {
     +public:
     +    /// Reports the exception reason.
     +    virtual const char* what() const noexcept;
    @@ include/seastar/core/condition-variable.hh: class condition_variable {
     +
     +        abortable_awaiter(condition_variable* cv, abort_source& as)
     +            : awaiter(cv)
    -+            , _sub(as.subscribe([this] () noexcept { abort(); }))
    ++            , _sub(as.subscribe([this] () noexcept { this->abort(); }))
     +        {}
     +
     +        void signal() noexcept override {
    @@ include/seastar/core/condition-variable.hh: class condition_variable {
     +        auto operator co_await() {
     +            if (!_sub) {
     +                return ::seastar::internal::awaiter<false, void>(
    -+                        make_exception_future<>(std::make_exception_ptr(condition_variable_aborted())));
    ++                        make_exception_future<>(condition_variable_aborted()));
     +            }
     +            return awaiter::operator co_await();
     +        }
    @@ tests/unit/condition_variable_test.cc: SEASTAR_THREAD_TEST_CASE(test_condition_v
     +        auto f = cv1.wait(as);
     +        cv1.signal();
     +        sleep(10ms).get();
    -+        BOOST_REQUIRE_EQUAL(f.available(), true);
    ++        BOOST_REQUIRE(f.available());
     +        f.get();
     +    }
     +
     +    auto f1 = cv1.wait(as);
    -+    BOOST_REQUIRE_EQUAL(f1.available(), false);
    ++    BOOST_REQUIRE(!f1.available());
     +
     +    sleep(10ms).get();
    -+    BOOST_REQUIRE_EQUAL(f1.available(), false);
    ++    BOOST_REQUIRE(!f1.available());
     +
     +    condition_variable cv2;
     +
     +    as.request_abort();
     +    auto f2 = cv2.wait(as);
     +
    -+    BOOST_REQUIRE_EQUAL(f1.available(), true);
    -+    BOOST_REQUIRE_EQUAL(f2.available(), true);
    ++    BOOST_REQUIRE(f1.available());
    ++    BOOST_REQUIRE(f2.available());
     +
     +    BOOST_REQUIRE_THROW(
     +        f1.get(),
    @@ tests/unit/condition_variable_test.cc: SEASTAR_TEST_CASE(test_condition_variable
     @@ tests/unit/condition_variable_test.cc: SEASTAR_TEST_CASE(test_condition_variable_when_signal) {
          co_await cv.when();
          // ensure we did not resume before timer ran fully
    -     BOOST_REQUIRE_EQUAL(ready, true);
    +     BOOST_REQUIRE(ready);
     +
     +    ready = false;
     +    abort_source as;
    @@ tests/unit/condition_variable_test.cc: SEASTAR_TEST_CASE(test_condition_variable
     +    t.arm(100ms);
     +
     +    co_await cv.when(as);
    -+    BOOST_REQUIRE_EQUAL(ready, true);
    ++    BOOST_REQUIRE(ready);
     +
     +    as.request_abort();
     +    cv.signal();
```

kbr-scylla:
@avikivity please review/merge

avikivity:
You can accomplish waiting without explicit support.

```c++

    auto check_aborted = _as.subscribe([&] { _cv.broadcast(); });
    while (!_as.check() && !other_conditions) {
          co_await _cv.when();
          // handle abort_source or other_conditions
    }
```

condition_variable is a general class for waiting, we can't add specialized support for every other class that can be signaled, we'll have a combinatorial explosion. Should we have an overload for two abort_source:s?

kbr-scylla:
@avikivity You could argue the same about the timeout variants. Why do we have them?

Regarding this:
```
    auto check_aborted = _as.subscribe([&] { _cv.broadcast(); });
```
It has a performance problem. `broadcast()` has to iterate over all waiters.
Suppose there are N waiters, all of them run this code.
If all of them / most of them get aborted, you get N^2 iterations.

This is an actual consideration I made when trying to solve a problem in Scylla, that's when I decided it's best to have an abortable `wait` / `when`.

> Should we have an overload for two abort_source:s?

I think we should find a reasonable balance.

Two abort sources can easily be combined into a single one (and we could implement a proper abstraction for easy combining) in a performant way, so there's no point in having an overload for two abort sources.

`abort_source` is our go-to abstraction for aborting, so we should support it, and if we were to choose between supporting timeout variant or supporting `abort_source` variant, the `abort_source` one would be more important IMO, especially that we can easily turn a timeout into an abort using `abort_on_expiry`.

avikivity:
> @avikivity You could argue the same about the timeout variants. Why do we have them?
> 

Mostly because it's part of the traditional condition variable API.


> Regarding this:
> 
> ```
>     auto check_aborted = _as.subscribe([&] { _cv.broadcast(); });
> ```
> 
> It has a performance problem. `broadcast()` has to iterate over all waiters. Suppose there are N waiters, all of them run this code. If all of them / most of them get aborted, you get N^2 iterations.

You'd place the subscription in the same scope as the abort source, so there's be only one. In any case, aborts are rare.


> 
> This is an actual consideration I made when trying to solve a problem in Scylla, that's when I decided it's best to have an abortable `wait` / `when`.
> 
> > Should we have an overload for two abort_source:s?
> 
> I think we should find a reasonable balance.
> 
> Two abort sources can easily be combined into a single one (and we could implement a proper abstraction for easy combining) in a performant way, so there's no point in having an overload for two abort sources.
> 
> `abort_source` is our go-to abstraction for aborting, so we should support it, and if we were to choose between supporting timeout variant or supporting `abort_source` variant, the `abort_source` one would be more important IMO, especially that we can easily turn a timeout into an abort using `abort_on_expiry`.

This is the road to combinatorial explosion.

We could kill it by introducing a layer of indirection.

```
auto when(condition_variable_awaitable auto... awaitables)
```

So you can mix any combination of predicates, timers, timeouts, abort_sources, or anything else that can be signalled.

kbr-scylla:
> You'd place the subscription in the same scope as the abort source, so there's be only one.

In my use case there is a global condition variable signaled by schema changes. And there are waiters on the data path waiting for schema to be synced. If schema sync takes too long, all waiters will get aborted. If you include a `broadcast()` on every waiter for that global cv, you get N^2 iterations on waiters.

> This is the road to combinatorial explosion.

It is not combinatorial. Nobody proposes `wait(abort_source&, timeout)`, because there is no need to do such thing.

We support abort_source already in other primitives like semaphore or shared_future.

What is special about condition_variable that you want to prevent it from getting abort_source support, while other primitives got extended without such resistance? See e.g. 1d81c8e5aaa793ba41d374febe438cca9006babf, 721819dfb3b7dbc76c55037f79eb2e81f3db7b63, **merged by you** @avikivity .

condition_variable missing abort_source support is a hole in our APIs.

kbr-scylla:
Look, `shared_promise` API:

```cpp

    /// \brief Gets new future associated with this promise.
    /// If the promise is not resolved before timeout the returned future will resolve with \ref timed_out_error.
    /// This instance doesn't have to be kept alive until the returned future resolves.
    future_type get_shared_future(time_point timeout = time_point::max()) const noexcept {
        return _shared_future.get_future(timeout);
    }

    /// \brief Gets new future associated with this promise.
    /// If the promise is not resolved before abort source is triggered the returned future will
    /// resolve with \ref abort_requests_exception.
    /// This instance doesn't have to be kept alive until the returned future resolves.
    future_type get_shared_future(abort_source& as) const noexcept {
        return _shared_future.get_future(as);
    }
```

it has both timeout and abort_source variants! And you merged the patch which added the new variant.
Why do it for some primitives but not for others? It's a glaring inconsistency and simply annoying for the user.

gleb-cloudius:
> What is special about condition_variable that you want to prevent it from getting abort_source support, while other primitives got extended without such resistance? See e.g. [1d81c8e](https://github.com/scylladb/seastar/commit/1d81c8e5aaa793ba41d374febe438cca9006babf), [721819d](https://github.com/scylladb/seastar/commit/721819dfb3b7dbc76c55037f79eb2e81f3db7b63), **merged by you** @avikivity .

To be fair the same argument was made for at least one of those. FWIW I agree with Kamil. If we want only one way to abort the waiter to prevent explosion it should have been `abort_source` since it is easy to make a timer that triggers it if needed.

kbr-scylla:
An alternative and more generic solution would be to expose the waiter objects as a public interface, and expose a generic "signal()" method on the waiters. This would allow waking up the waiters directly, not just through condition variable.

This could be then used to implement both the abort_source variant and timeout variant, even as free functions outside Seastar.

However, this might turn out much more complicated to do. For example, `wait()` returns a future; it would additionally have to return something like a shared pointer to the waiter which can then be signaled by the caller of `wait()` if necessary. The coroutine version, `when()`, already returns an awaiter object, but now it's supposed to be `co_await`ed immediately, allowing it to get `co_await`ed at some later step complicates the situation (what if we yield in between and so on...)

avikivity:
The question to be answered is what are the primitives here and what are high-level building blocks? I view condition_variable as a low-level primitive and abort_source as a higher level construct, but maybe it's a wrong way to look at things. And yes, I did resist adding abort_source support everywhere, but maybe that's wrong.

kbr-scylla:
Maybe I should look into your proposal anyway:
```
auto when(condition_variable_awaitable auto... awaitables)
```

to provide a generic "wake up the waiter bypassing the condition variable", and then provide abort_source wrapper inside Scylla  itself.

avikivity:
> Maybe I should look into your proposal anyway:
> 
> ```
> auto when(condition_variable_awaitable auto... awaitables)
> ```
> 
> to provide a generic "wake up the waiter bypassing the condition variable", and then provide abort_source wrapper inside Scylla itself.

There's no need for ScyllaDB to link between two Seastar primitives.

But, if we accept the picture of condition_variable being a low-level primitive, then abort_source could provide that glue.

michoecho:
@avikivity A waiting primitive needs *some* way to cancel the wait, otherwise it's a bad primitive. The fact that `condition_variable` doesn't have a way to cancel the wait is a hole in the API.
If anything, the `timeout` variants are the superfluous ones, because a timeout can be implemented with a timer that cancels the wait. So I agree with @kbr-scylla and @gleb-cloudius. 

That said, `abort_source` isn't the minimal possible API for cancellation. The minimal API would be:
```
class condition_variable::cancel_callback {
    condition_variable::waiter* _waiter;
public:
    void operator()(std::exception_ptr& reason);
};

std::pair<future<>, cancel_callback> condition_variable::wait();
// Or, equivalently, but with extra convenience:
future<> condition_variable::wait(std::invocable<cancel_callback> auto f);
```

`abort_source` isn't a minimal primitive, and that's a bad thing. But the difference isn't big enough to matter. The more important thing is that there is *some* way to cancel the wait.
  
> You'd place the subscription in the same scope as the abort source, so there's be only one. In any case, aborts are rare.

"abort" in this context means any cancellation. Server shutdown isn't the only occasion where you might want to cancel a wait. The API should allow any wait to be cancelled independently. Your proposed method of cancellation is viable only if we only ever want to cancel all waiters at once.

PR #1900, opened 2023-10-26
build: add GLOB parameter to seastar_check_self_contained()

tchaikov:
sometimes, parent project would like to check header's self-containness without listing them in the CMake script, the project might want to specify a glob pattern instead.

this enables the parent projects to add header files by specifying a glob pattern, so they can add the header files in a more convenient way.

### DIFF
--- a/cmake/CheckHeaders.cmake
+++ b/cmake/CheckHeaders.cmake
@@ -33,10 +33,18 @@ function (seastar_check_self_contained target library)
     parsed_args
     ""
     ""
-    "EXCLUDE;INCLUDE"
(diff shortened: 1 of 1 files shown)


avikivity:
Why would the parent process want to check Seastar headers? It can't do anything if the check fails.

tchaikov:
i plan to let scylla to use this function for checking its headers' self-containness. so i don't need to reimplement it again in scylla.

avikivity:
But then it becomes a stable API and we have to avoid breaking compatibility.

tchaikov:
fair enough. i don't feel strong either way. will replicate this function on scylla's side then.

PR #1903, opened 2023-10-27
coroutine: fix use-after-free in parallel_for_each

### LINKED ISSUE #1902: coroutine::parallel_for_each leaves a dangling reactor::_current_task
When reactor resumes a task, it sets `reactor::_current_task` to it. This is needed to implement `current_tasktrace()`.

For this to stay correct, every `task` object has to stay alive since it's resumed until control returns to the reactor.

`coroutine::parallel_for_each` breaks this rule: when it finishes after being resumed, it resumes the parent coroutine and dies. The reactor is left pointing to the now-dead `parallel_for_each`. If `current_tasktrace()` is called in this state, undefined behavior occurs.

Refs https://github.com/scylladb/seastar/issues/1599 and https://github.com/scylladb/seastar/issues/1759. This is the same kind of bug.

michoecho:
After coroutine::parallel_for_each resumes its caller coroutine and dies, reactor::_current_task might be left dangling.

To fix this without allocations, we have to set _current_task to the coroutine before resuming it.

Fixes #1902

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -579,6 +579,11 @@ public:
     }
 
     task* current_task() const { return _current_task; }
+    // If a task wants to resume a different task instead of returning control to the reactor,
+    // it should set _current_task to the resumed task.
+    // In particular, this is mandatory if the task is going to die before control is returned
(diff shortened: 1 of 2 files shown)


tchaikov (approved):
lgtm

PR #1904, opened 2023-10-27
Add sampling memory profiler mode

travisdowns:
This series changes the existing heap profiling to support a sampling mode, added
by @StephanDollberg and includes also some optimizations of the allocation fast paths.

The following description of the sampling profiler comes from Stephan:

Instead of tracking every allocation, it will track a sampled subset of allocations. This
is useful for profiling in production systems where the overhead of tracking every
allocation is too high.

A sample is taken at a set sample rate which feeds into an exponential
distribution. Hence we avoid falling into patterns that would potentially make
the profile biased otherwise. This approach is inspired by the android heapprofd
memory profiler.

By adjusting the sample rate it's possible to tune the overhead of the memory
profiler. The idea is to make this an always-on production profiler that can
give insight into allocations in OOM scenarios with negliable overhead.

Starting with the latest release of Redpanda we are using this in a default
always-on mode with a sampling period of 3MB. This is a rather conservative
setting and results in no measurable overhead while still giving good resolution
for the top allocation contributors.

We also add an API to retrieve the current heap profile programmatically. In
Redpanda we use this to print the top sites on OOM using the additional
diagnostics API. Further it allows easier retrieval of the data in a
dev/profiling fashion.

As today heap profiler can be compiled in or not (using `--heap-prof` at 
./configure.py time) and can also be enabled or disabled at runtime if it
has been compiled in.

The individual commits contain a more detailed description of the respective
changes. Note the heap profiler and all added codepaths still require explicit
opt in via the extra define.

This series also includes some optimizations that especially improve
small allocations and hence we end up faster than the current state even with
the profiler on, under the assumption that small allocations (which here means
<= 16 KiB) are considerably more frequent than larger allocations, which is 
overwhelmingly the case for most applications.

For example, a new/delete pair for a small allocation prior to this series take on
average ~104 instructions, versus ~73 after _with heap profiling on_ or ~67 with
it compiled out. Performance measured in time rather than instructions generally
follows the same improvement factors as time, though with more variance.

To aid with analysis of the above we also added a set of microbenchmarks. Below
you will find the selected key numbers from this benchmark comparing current
master and this patch set with the profiler compiled in and
out. 

Note profiler-on numbers are taken from a run with very high sample period
to avoid actually taking any samples. Sample creation is fully dominated by
backtrace creation so this avoids noise.

With the profiler enabled at compile time the runtime performance is identical 
between "compiler turned off at runtime" and "very large sample period such
that zero samples are actually taken". That is, there is no additional check for
sampler enablement on the fast path, only the single check to see if the sampling
period has been reached.

There are a lot of profiling results, so I'll share first this summary of a single 
test, which uses `new` to allocate many small pointers and then deletes them.

The configurations show are:

*Baseline:* Current `master` without any of the changes in this series.
*New-OFF:* This series but with heap profiling disabled at compile time.
*New-ON:* This series and with heap profiling enabled at compile time.

First the instruction counts:

| Config | Compiler | test | instructions |
| - | - | - | -: | 
| Baseline | gcc-11   | alloc_bench.op_new_delete                | 97.6 |
| New-OFF  | gcc-11   | alloc_bench.op_new_delete                | 74.6 |
| New-ON   | gcc-11   | alloc_bench.op_new_delete                | 82.6 |
| Baseline | clang-16 | alloc_bench.op_new_delete                | 104.6 |
| New-OFF  | clang-16 | alloc_bench.op_new_delete                | 67.5|
| New-ON   | clang-16 |alloc_bench.op_new_delete                |  73.5 |

Next, runtimes on Skylake and Zen 4, these are from separate runs than the instruction counts above as I found that enabling instruction count measurement disturbed (i.e., increased) the timings by a significant amount, probably because some part of the perf syscall cost ends up timed.

| Config | Arch | Compiler | test                                     |  iterations |      median |         mad |         min |         max |
|-|-|-| -                                        |           - |           -: |           -: |           -: |           -: |
| Baseline | Skylake | clang-16  | alloc_bench.op_new_delete                |   134081000 |     7.376ns |     0.005ns |     7.370ns |     7.534ns |
| New-OFF  | Skylake | clang-16  | alloc_bench.op_new_delete                |   146836000 |     6.586ns |     0.008ns |     6.577ns |     6.765ns |
| New-ON   | Skylake | clang-16  | alloc_bench.op_new_delete                |   193241000 |     5.113ns |     0.004ns |     5.110ns |     5.246ns |
||
| Baseline | Zen4 | gcc-11   | alloc_bench.op_new_delete                |   207025000 |     4.796ns |     0.000ns |     4.795ns |     4.797ns |
| New-OFF  | Zen4 | gcc-11   | alloc_bench.op_new_delete                |   305103000 |     3.241ns |     0.000ns |     3.240ns |     3.241ns |
| New-ON   | Zen4 | gcc-11   | alloc_bench.op_new_delete                |   302127000 |     3.271ns |     0.000ns |     3.271ns |     3.273ns |
||
| Baseline | Zen4 | clang-16 | alloc_bench.op_new_delete                |   212207000 |     4.663ns |     0.000ns |     4.661ns |     4.663ns |
| New-OFF  | Zen4 | clang-16 | alloc_bench.op_new_delete                |   358628000 |     2.742ns |     0.001ns |     2.740ns |     2.745ns |
| New-ON   | Zen4 | clang-16 | alloc_bench.op_new_delete                |   347325000 |     2.833ns |     0.003ns |     2.830ns |     2.836ns |

The primary anomalous result is New-OFF on clang-16 on Skylake (also shows up on Alder Lake, not shown). The runtime is considerably slower than the instruction count would suggest (i.e., it is lower IPC than other configurations) and I don't see any problem in a manual inspection of the assembly. The same function is fast on Zen 2 and Zen 4. I believe it is probably an Intel-specific alignment effect. I did move the bottom of the test loop around to avoid the jcc erratum (results not shown) which improved the results but the result was still slower than expected. In any case, even this configuration is faster than the baseline and I expect the issue to go away at some point.

Large allocations are a bit slower, very slightly when heap profiling is off at compile time and more when they are enabled, e.g., here with 32,000 byte allocations:

*Baseline*
```
test                                      iterations      median         mad         min         max      allocs       tasks        inst
alloc_bench.alloc_free_large                72444000    13.125ns     0.010ns    13.073ns    13.140ns       1.000       0.000       336.1
```

*New-OFF*
```
test                                      iterations      median         mad         min         max      allocs       tasks        inst
alloc_bench.alloc_free_large                71368000    13.334ns     0.004ns    13.287ns    13.338ns       1.000       0.000       343.0
```

*New-ON*
```
test                                      iterations      median         mad         min         max      allocs       tasks        inst
alloc_bench.alloc_free_large                65224000    14.623ns     0.002ns    14.604ns    14.628ns       1.000       0.000       399.5
```

Such allocations are rare compared to small allocations, however, and in any case the relative cost of allocation goes down compared to the work done on the allocated memory itself (e.g., writing into it) and that ratio becomes ~negligible except for an application which allocates large regions but then does not use them.

We have many micro-benchmarks but this description is already quite long: let me know what you think would be useful.

### DIFF
--- a/include/seastar/core/memory.hh
+++ b/include/seastar/core/memory.hh
@@ -414,8 +414,13 @@ public:
 struct allocation_site {
     mutable size_t count = 0; /// number of live objects allocated at backtrace.
     mutable size_t size = 0; /// amount of bytes in live objects allocated at backtrace.
-    mutable const allocation_site* next = nullptr; // TODO: remove
     simple_backtrace backtrace; /// call site for this allocation
+
+    // All allocation sites are linked to each other. This can be used for easy
+    // iteration across them in gdb scripts where it's difficult to work with
+    // the C++ data structures.
+    mutable const allocation_site* next = nullptr; // next allocation site in the chain
+    mutable const allocation_site* prev = nullptr; // previous allocation site in the chain
 
     bool operator==(const allocation_site& o) const {
         return backtrace == o.backtrace;
--- a/include/seastar/core/memory.hh
+++ b/include/seastar/core/memory.hh
@@ -25,6 +25,7 @@
 #include <seastar/core/bitops.hh>
 #include <seastar/util/backtrace.hh>
 #include <seastar/util/modules.hh>
+#include <seastar/util/sampler.hh>
 #ifndef SEASTAR_MODULE
 #include <new>
 #include <cstdint>
@@ -91,8 +92,9 @@ namespace seastar {
 /// ### Heap profiling
 ///
 /// Heap profiling allows finding out how memory is used by your application, by
-/// recording the stacktrace of all allocations. See:
-/// * \ref set_heap_profiling_enabled()
+/// recording the stacktrace of a sampled subset (or all) allocations. See:
+/// * \ref set_heap_profiling_sampling_rate()
+/// * \ref sampled_memory_profile()
 /// * \ref scoped_heap_profiling
 ///
 /// ### Abort on allocation failure
@@ -177,10 +179,9 @@ public:
 // Can be nested, in which case the profiling is re-enabled when all
 // the objects go out of scope.
 class disable_backtrace_temporarily {
-    bool _old;
+    sampler::disable_sampling_temporarily _disable_sampling;
 public:
     disable_backtrace_temporarily();
-    ~disable_backtrace_temporarily();
 };
 
 enum class reclaiming_result {
@@ -410,7 +411,10 @@ public:
 ///
 /// The location is identified by its backtrace. One allocation_site can
 /// represent many allocations at the same location. `count` and `size`
-/// represent the cumulative sum of all allocations at the location.
+/// represent the cumulative sum of all allocations at the location. Note the
+/// size represents an extrapolated size and not the sampled one, i.e.: when
+/// looking at the total size of all allocation sites it will approximate the
+/// total memory usage
 struct allocation_site {
     mutable size_t count = 0; /// number of live objects allocated at backtrace.
     mutable size_t size = 0; /// amount of bytes in live objects allocated at backtrace.
@@ -431,9 +435,13 @@ struct allocation_site {
     }
 };
 
-/// @brief If memory profiling is on returns the current memory live set
+/// @brief If memory sampling is on returns the current sampled memory live set
+///
+/// If there is tracked allocations (because heap profiling was on earlier)
+/// these will still be returned if heap profiling is now off
+///
 /// @return a vector of \ref allocation_site
-std::vector<allocation_site> memory_profile();
+std::vector<allocation_site> sampled_memory_profile();
 
 /// @brief Copies the current sampled set of allocation_sites into the
 /// array pointed to by the output parameter
@@ -448,31 +456,43 @@ std::vector<allocation_site> memory_profile();
 /// @param output array to copy the allocation sites to
 /// @param size the size of the array pointed to by \p output
 /// @return number of \ref allocation_site copied to the vector
-size_t memory_profile(allocation_site* output, size_t size);
+size_t sampled_memory_profile(allocation_site* output, size_t size);
 
-/// Enable heap profiling
+/// @brief Enable sampled heap profiling by setting a sample rate
+///
+/// @param sample_rate the sample rate to use. Disable heap profiling by setting
+/// the sample rate to 0
 ///
 /// In order to use heap profiling you have to define
 /// `SEASTAR_HEAPPROF`.
+///
+/// Use \ref sampled_memory_profile for API access to profiling data
+///
+/// Note: Changing the sampling rate while previously sampled allocations are
+/// still alive can lead to inconsistent results of their reported size (i.e.:
+/// their size will be over or under reported). Undefined behavior or memory
+/// corruption will not occur.
 ///
 /// For an example script that makes use of the heap profiling data
 /// see [scylla-gdb.py] (https://github.com/scylladb/scylla/blob/e1b22b6a4c56b4f1d0adf65d1a11db4bcb51fe7d/scylla-gdb.py#L1439)
 /// This script can generate either textual representation of the data,
 /// or a zoomable flame graph ([flame graph generation instructions](https://github.com/scylladb/scylla/wiki/Seastar-heap-profiler),
 /// [example flame graph](https://user-images.githubusercontent.com/1389273/72920437-f0cf8a80-3d51-11ea-92f0-f3dbeb698871.png)).
-void set_heap_profiling_enabled(bool);
+void set_heap_profiling_sampling_rate(size_t sample_rate);
 
-/// Checks whether heap profiling is currently enabled
-/// @return true if heap profiling is enabled, false otherwise
-bool get_heap_profiling_enabled();
+/// @brief Returns the current heap profiling sampling rate (0 means off)
+/// @return the current heap profiling sampling rate
+size_t get_heap_profiling_sample_rate();
 
 /// @brief Enable heap profiling for the duration of the scope.
 ///
+/// Note: Nesting different sample rates is currently not supported.
+///
 /// For more information about heap profiling see
-/// \ref set_heap_profiling_enabled().
+/// \ref set_heap_profiling_sampling_rate().
 class scoped_heap_profiling {
 public:
-    scoped_heap_profiling() noexcept;
+    scoped_heap_profiling(size_t) noexcept;
     ~scoped_heap_profiling();
 };
 
--- a/include/seastar/core/memory.hh
+++ b/include/seastar/core/memory.hh
@@ -23,6 +23,7 @@
 
 #include <seastar/core/resource.hh>
 #include <seastar/core/bitops.hh>
+#include <seastar/util/backtrace.hh>
 #include <seastar/util/modules.hh>
 #ifndef SEASTAR_MODULE
 #include <new>
@@ -405,13 +406,50 @@ public:
     void operator=(scoped_large_allocation_warning_disable&&) = delete;
 };
 
-/// Enable/disable heap profiling.
+/// @brief Describes an allocation location in the code.
+///
+/// The location is identified by its backtrace. One allocation_site can
+/// represent many allocations at the same location. `count` and `size`
+/// represent the cumulative sum of all allocations at the location.
+struct allocation_site {
+    mutable size_t count = 0; /// number of live objects allocated at backtrace.
+    mutable size_t size = 0; /// amount of bytes in live objects allocated at backtrace.
+    mutable const allocation_site* next = nullptr; // TODO: remove
+    simple_backtrace backtrace; /// call site for this allocation
+
+    bool operator==(const allocation_site& o) const {
+        return backtrace == o.backtrace;
+    }
+
+    bool operator!=(const allocation_site& o) const {
+        return !(*this == o);
+    }
+};
+
+/// @brief If memory profiling is on returns the current memory live set
+/// @return a vector of \ref allocation_site
+std::vector<allocation_site> memory_profile();
+
+/// @brief Copies the current sampled set of allocation_sites into the
+/// array pointed to by the output parameter
+///
+/// Copies up to \p size elements of the current sampled set of allocation
+/// sites into the output array, which must have length at least \p size.
+///
+/// Returns amount of copied elements. This method does not allocate so it
+/// is a useful alternative to \ref sampled_memory_profile() when one wants
+/// to avoid allocating (e.g.: under OOM conditions).
+///
+/// @param output array to copy the allocation sites to
+/// @param size the size of the array pointed to by \p output
+/// @return number of \ref allocation_site copied to the vector
+size_t memory_profile(allocation_site* output, size_t size);
+
+/// Enable heap profiling
 ///
 /// In order to use heap profiling you have to define
 /// `SEASTAR_HEAPPROF`.
-/// Heap profiling data is not currently exposed via an API for
-/// inspection, instead it was designed to be inspected from a
-/// debugger.
+///
 /// For an example script that makes use of the heap profiling data
 /// see [scylla-gdb.py] (https://github.com/scylladb/scylla/blob/e1b22b6a4c56b4f1d0adf65d1a11db4bcb51fe7d/scylla-gdb.py#L1439)
 /// This script can generate either textual representation of the data,
@@ -419,7 +457,11 @@ public:
 /// [example flame graph](https://user-images.githubusercontent.com/1389273/72920437-f0cf8a80-3d51-11ea-92f0-f3dbeb698871.png)).
 void set_heap_profiling_enabled(bool);
 
-/// Enable heap profiling for the duration of the scope.
+/// Checks whether heap profiling is currently enabled
+/// @return true if heap profiling is enabled, false otherwise
+bool get_heap_profiling_enabled();
+
+/// @brief Enable heap profiling for the duration of the scope.
 ///
 /// For more information about heap profiling see
 /// \ref set_heap_profiling_enabled().
@@ -432,4 +474,15 @@ public:
 SEASTAR_MODULE_EXPORT_END
 
 }
+}
+
+namespace std {
+
+template<>
+struct hash<seastar::memory::allocation_site> {
+    size_t operator()(const seastar::memory::allocation_site& bi) const {
+        return std::hash<seastar::simple_backtrace>()(bi.backtrace);
+    }
+};
+
 }
--- a/include/seastar/util/sampler.hh
+++ b/include/seastar/util/sampler.hh
@@ -1,0 +1,95 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef SRC_PROFILING_MEMORY_SAMPLER_H_
+#define SRC_PROFILING_MEMORY_SAMPLER_H_
+
+#include <stdint.h>
+
+#include <atomic>
+#include <random>
+
+#include "perfetto/ext/base/utils.h"
+
+namespace perfetto {
+namespace profiling {
+
+constexpr uint64_t kSamplerSeed = 1;
+
+std::default_random_engine& GetGlobalRandomEngineLocked();
+
+// Poisson sampler for memory allocations. We apply sampling individually to
+// each byte. The whole allocation gets accounted as often as the number of
+// sampled bytes it contains.
+//
+// The algorithm is inspired by the Chromium sampling algorithm at
+// https://cs.chromium.org/search/?q=f:cc+symbol:AllocatorShimLogAlloc+package:%5Echromium$&type=cs
+// Googlers: see go/chrome-shp for more details.
+//
+// NB: not thread-safe, requires external synchronization.
+class Sampler {
+ public:
+  void SetSamplingInterval(uint64_t sampling_interval) {
+    sampling_interval_ = sampling_interval;
+    sampling_rate_ = 1.0 / static_cast<double>(sampling_interval_);
+    interval_to_next_sample_ = NextSampleInterval();
+  }
+
+  // Returns number of bytes that should be be attributed to the sample.
+  // If returned size is 0, the allocation should not be sampled.
+  //
+  // Due to how the poission sampling works, some samples should be accounted
+  // multiple times.
+  size_t SampleSize(size_t alloc_sz) {
+    if (PERFETTO_UNLIKELY(alloc_sz >= sampling_interval_))
+      return alloc_sz;
+    return static_cast<size_t>(sampling_interval_ * NumberOfSamples(alloc_sz));
+  }
+
+  uint64_t sampling_interval() const { return sampling_interval_; }
+
+ private:
+  int64_t NextSampleInterval() {
+    std::exponential_distribution<double> dist(sampling_rate_);
+    int64_t next = static_cast<int64_t>(dist(GetGlobalRandomEngineLocked()));
+    // We approximate the geometric distribution using an exponential
+    // distribution.
+    // We need to add 1 because that gives us the number of failures before
+    // the next success, while our interval includes the next success.
+    return next + 1;
+  }
+
+  // Returns number of times a sample should be accounted. Due to how the
+  // poission sampling works, some samples should be accounted multiple times.
+  size_t NumberOfSamples(size_t alloc_sz) {
+    interval_to_next_sample_ -= alloc_sz;
+    size_t num_samples = 0;
+    while (PERFETTO_UNLIKELY(interval_to_next_sample_ <= 0)) {
+      interval_to_next_sample_ += NextSampleInterval();
+      ++num_samples;
+    }
+    return num_samples;
+  }
+
+  uint64_t sampling_interval_;
+  double sampling_rate_;
+  int64_t interval_to_next_sample_;
+};
+
+}  // namespace profiling
+}  // namespace perfetto
+
+#endif  // SRC_PROFILING_MEMORY_SAMPLER_H_
--- a/include/seastar/util/sampler.hh
+++ b/include/seastar/util/sampler.hh
@@ -14,82 +14,148 @@
  * limitations under the License.
  */
 
-#ifndef SRC_PROFILING_MEMORY_SAMPLER_H_
-#define SRC_PROFILING_MEMORY_SAMPLER_H_
+// This file has been originally been imported from:
+// https://cs.android.com/android/platform/superproject/+/013901367630d3ec71c9f2bb3f3077bd11585301:external/perfetto/src/profiling/memory/sampler.h
 
-#include <stdint.h>
+//
+// The code has been modified as follows:
+//
+//  - Integrated into seastar and adapted to coding style
+//  - Right now we don't account for samples multiple times (in case we have
+//  multiple loops of drawing from the exp distribution). The reason is that
+//  in our memory sampler we would have to store the weight in addition to the
+//  alloation site ptr as on free we need to know how much a sample accounted
+//  for. Hence, for now we simply always use the sampling interval.
+//  - Sampler can be turned "off" with a 0 sampling rate
+//  - The fast path is more optimized (as a consequence of the first point)
+//  - Provide a way to temporarily pause sampling
+//
+// Changes Copyright (C) 2023 ScyllaDB
 
-#include <atomic>
+#pragma once
+
 #include <random>
 
-#include "perfetto/ext/base/utils.h"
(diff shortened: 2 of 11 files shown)


avikivity:
Looks very promising from reading the cover letter, but as this is delicate material I'd like multiple reviewers. Enlisting @michoecho

tests/perf/allocator_perf.cc
```diff
@@ -0,0 +1,139 @@
...
+struct allocator_test {
+    static constexpr size_t COUNT = 1000;
+    std::array<void *, COUNT> _pointers{};
+};
+
+struct allocator_test_split : public allocator_test {
+    enum alloc_test_flags {
+      MEASURE_ALLOC = 1 << 0,
+      MEASURE_FREE = 2 << 0,
+    };
```
avikivity, line 54: please use 4-space indents throughout

avikivity: also 1 << 0, 2 << 0? Did you mean 1<< 0, 1 << 1?

tests/perf/allocator_perf.cc
```diff
@@ -0,0 +1,139 @@
...
+
+    template <int M> struct maybe_measure {
```
avikivity, line 60: bool? what's M?

include/seastar/core/memory.hh
```diff
@@ -405,20 +406,48 @@ public:
...
+/// allocations at the location.
+struct allocation_site {
+    mutable size_t count = 0; // number of live objects allocated at backtrace.
+    mutable size_t size = 0; // amount of bytes in live objects allocated at backtrace.
+    mutable const allocation_site* next = nullptr; // TODO: remove
+    simple_backtrace backtrace; // call site for this allocation
+
+    bool operator==(const allocation_site& o) const {
+        return backtrace == o.backtrace;
+    }
```
avikivity, line 428: It's not good for operator== to only compare part of the struct, but I see it's a preexisting problem so no need to fix it in the series.

src/core/memory.cc
```diff
@@ -504,29 +474,35 @@ class small_pool_array {
...
     // For page-aligned sizes, allocation_site* lives in page::alloc_site, not with the object.
     static_assert(is_page_aligned(max_small_allocation), "assuming that max_small_allocation is page aligned so that we"
             " don't need to add allocation_site_ptr to objects of size close to it");
-    size_t next_page_aligned_size = next_page_aligned(size);
-    if (next_page_aligned_size - size > sizeof(allocation_site_ptr)) {
-        size += sizeof(allocation_site_ptr);
-    } else {
-        return next_page_aligned_size;
+    if (is_sampled)
+    {
```
avikivity: Please keep to the coding style which puts the brackets on the same line as the if.

src/core/memory.cc
```diff
@@ -408,14 +408,21 @@ class small_pool {
     unsigned _min_free;
     unsigned _max_free;
     unsigned _pages_in_use = 0;
+    // Flag to indicate whether this pool stores sampled allocations.
+    // When freeing small allocations this flag is checked to see whether an
+    // allocation site pointer is part of the object and the allocation needs
+    // removal from the allocation_site tracking
+    bool _sampled_pool = false;
```
avikivity, line 444: #ifdef HEAPPROF

include/seastar/util/sampler.hh
```diff
@@ -0,0 +1,95 @@
...
+
+#ifndef SRC_PROFILING_MEMORY_SAMPLER_H_
+#define SRC_PROFILING_MEMORY_SAMPLER_H_
+
+#include <stdint.h>
+
+#include <atomic>
+#include <random>
+
+#include "perfetto/ext/base/utils.h"
```
avikivity: This won't work

src/core/memory.cc
```diff
@@ -1121,20 +1121,19 @@ cpu_pages::try_free_fastpath(void* ptr) {
...
 /// The size parameter will be either no_size tag type or size_t, and most
 /// of the implementation can be shared, using constexpr if or other dispatch
 /// in the places where there should be a difference of behavior.
-template<typename T>
-concept SizeOrNot = std::same_as<T, size_t> || std::same_as<T, no_size>;
+struct no_size {};
 
+template <typename S>
+SEASTAR_CONCEPT(requires std::same_as<S, size_t> || std::same_as<S, no_size>)
 [[gnu::noinline]]
```
avikivity, line 1132: Please fold into the patch that introduces the problem.

avikivity:
Looks good. Please make newly introduced public APIs doxygen documented and verify the rendering.

include/seastar/util/sampler.hh
StephanDollberg: You have probably realized at this point but I just want to clarify the approach we took here. 

We got advised that when importing only a part of a library it's best to make it clear what the original imported work is and what changes were made on top. 

Hence, we split the import & change into two. The first commit (what you commented on and what will definitely not compile) just imports the original file as is. The second commit then does the modifications.

If you prefer any other way of doing this we are of course very happy to change.

tests/perf/allocator_perf.cc
travisdowns: Let's say this was an easter egg to ensure reviewers were awake which will be removed on a subsequent push :).

src/core/memory.cc
travisdowns: This is addressed in https://github.com/scylladb/seastar/commit/9d5d2445547b0eb47feef73a708dc9aeb1fbe005.

tests/perf/allocator_perf.cc
travisdowns: This is addressed (fix and offending commit folded) in https://github.com/scylladb/seastar/commit/9d5d2445547b0eb47feef73a708dc9aeb1fbe005.

include/seastar/util/sampler.hh
avikivity: No, it's fine and a good approach.

tests/perf/allocator_perf.cc
avikivity: oh come on, the only thing that triggers me more is trailing whitespace

tests/perf/allocator_perf.cc
avikivity: No please change the name M to something meaningful. I know templates are supposed to be difficult but they do allow multi-character names.

And you could narrow the int to bool by writing something != 0.

include/seastar/util/sampler.hh
avikivity: A line in the patch changelog would help (for reviewers that read the changelog).

StephanDollberg: Have amended the commit text

src/core/memory.cc
StephanDollberg: Added

src/core/memory.cc
StephanDollberg: Fixed a bunch of those.

StephanDollberg:
> Looks good. Please make newly introduced public APIs doxygen documented and verify the rendering.

I have added more of the doxygen parameters and checked the generated html.

travisdowns:
@michoecho have you had a chance to take a peek at this one? All of @avikivity's feedback should be addressed.

tests/perf/allocator_perf.cc
travisdowns: After another look at this I was writing 20 lines to save 5 (numbers approximate), so I've refactored this a bit and the NTTP has disappeared and I think it's cleaner overall. Thanks for the eyes on this.

tests/perf/allocator_perf.cc
travisdowns: This reminded me to dust off my .clang-format patch for seastar. Get ready for some vigorous debate!

michoecho (approved):
Sorry for the delay, I forgot about this.

Looks good. I left a few comments, but feel free to ignore the nitpicks.

src/core/memory.cc
```diff
@@ -438,7 +445,7 @@ class small_pool {
 public:
     explicit small_pool(unsigned object_size, bool is_sampled) noexcept;
     ~small_pool();
-    void* allocate();
+    inline void* allocate();
```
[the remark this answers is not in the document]
travisdowns: Yes, it declares `small_pool::allocate` to be inline, which increases the chance of it being inlined and also prevents an out-of-line definition from being included in the object file and binary.

include/seastar/util/sampler.hh
```diff
@@ -145,14 +152,12 @@ private:
         int64_t next = static_cast<int64_t>(dist(random_gen));
         // We approximate the geometric distribution using an exponential
         // distribution.
-        // We need to add 1 because that gives us the number of failures before
-        // the next success, while our interval includes the next success.
-        return next + 1;
+        return next;
```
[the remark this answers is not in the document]
travisdowns: The important part is that `maybe_sample` should use a `< 0` check instead of a `<= 0` check since this allows the whole decrement and compare to 0 operation which is on the hot path of every allocation to be serviced by a single RMW like `sub [interval_to_next_sample_], rax` followed by a comparison using the carry flag. IIRC that saved 3 instructions on the hot path though the net improvement is less than this would suggest as an RMW is more expensive than your average instruction (though less expensive than the replacement sequence).

To make that change correct, I had to decrement the initial value of `interval_to_next_sample_` by 1 which is why the  +1 disappeared.

Note that the change is from `next + 1` to `next` not the other way around (your comment suggests the other way around).

include/seastar/util/sampler.hh
```diff
@@ -14,82 +14,145 @@
...
+        }
+    }
 
-}  // namespace profiling
-}  // namespace perfetto
+    int64_t next_sampling_interval() {
+        std::exponential_distribution<double> dist(sampling_rate_);
+        int64_t next = static_cast<int64_t>(dist(random_gen));
+        // We approximate the geometric distribution using an exponential
+        // distribution.
```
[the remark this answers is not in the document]
travisdowns: Interesting question, I'm not sure. This is probably from the original source the sampler was drawn from. @StephanDollberg do you know why this was used?

It wouldn't surprise me if exponential was faster as continuous distributions with closed formed floating point formulas often end up faster than integer formulas which need to calculate integer powers, etc.

include/seastar/core/memory.hh
```diff
@@ -431,43 +435,59 @@ struct allocation_site {
...
 ///
 /// In order to use heap profiling you have to define
 /// `SEASTAR_HEAPPROF`.
 ///
+/// Use \ref sampled_memory_profile for API access to profiling data
+///
+/// Note: Changing the sampling rate is currently not supported.
+/// Re-enabling heap profiling with a different sample rate to previous is fine
+/// to do if and only if all allocations are freeed before heap profiling is
+/// turned back on.
```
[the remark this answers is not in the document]
travisdowns: Yeah possibly. What would you suggest if the check fails though? `abort()`?

tests/perf/allocator_perf.cc
```diff
@@ -0,0 +1,126 @@
...
+};
+
+PERF_TEST_F(allocator_test_split, alloc_only) { return alloc_test<8, MEASURE_ALLOC>(); }
+
+PERF_TEST_F(allocator_test_split, free_only) { return alloc_test<8, MEASURE_FREE>(); }
+
+PERF_TEST_F(allocator_test_split, alloc_free) { return alloc_test<8>(); }
+
+// this test doesn't serve much value. It should take about 10 times as the
+// single alloc test above. If not, something is wrong.
```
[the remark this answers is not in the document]
travisdowns: "Single" in the sense that the tests above report the average cost of a single allocation (as they return the allocation count so the test can normalize by that value).

tests/unit/alloc_test.cc
```diff
@@ -331,35 +331,114 @@ SEASTAR_TEST_CASE(test_diagnostics_allocation) {
...
+    // NB: the test framework allocate
+    seastar::memory::set_heap_profiling_enabled(0);
+
+    {
+        auto stats = seastar::memory::memory_profile();
+        BOOST_REQUIRE_EQUAL(stats.size(), 2);
+        BOOST_REQUIRE_EQUAL(stats[0].size, stats[0].count * 1000000);
+    }
+
+    seastar::memory::set_heap_profiling_enabled(1000000);
```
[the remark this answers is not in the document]
travisdowns: Good eye. I accept your offer to ignore this nitpick however as rewriting it to fix this will take a few minutes I could spend doing something like getting a the sesatar codebase auto-formatted. However if there is a concern about bisectability I can fix it.

src/core/memory.cc
```diff
@@ -1494,18 +1500,24 @@ void* allocate(size_t size) {
     if (size <= sizeof(free_object)) {
         size = sizeof(free_object);
     }
+
+#ifdef SEASTAR_HEAPPROF
+    bool should_sample = get_cpu_mem().should_sample(size);
+#else
+    bool should_sample = get_cpu_mem().should_sample(size);
+#endif
```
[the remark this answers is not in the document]
travisdowns: Good eye. I accept your offer to ignore this nitpick however as rewriting it to fix this will take a few minutes I could spend doing something like improving sesatar compile times.

include/seastar/util/sampler.hh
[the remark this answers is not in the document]
StephanDollberg: Yes exactly, the perfetto blog post has some numbers on that:

```
STL Distributions
Benchmarking of the STL distributions on a Pixel 4 reinforces our approach of estimating the geometric distribution using an exponential distribution, as its performance is ~8x better ([full results](https://gist.github.com/fmayer/3aafcaf58f8db09714ba09aa4ac2b1ac)).

Draw sample from exponential distribution with p = 1 / 32000:

ARM64: 21.3ns (sd 0.066ns, negligibly small, smaller than a CPU cycle)

ARM32: 33.2ns (sd 0.011ns, negligibly small, smaller than a CPU cycle)

Draw sample from geometric distribution with p = 1 / 32000:

ARM64: 169ns (sd 0.023ns, negligibly small, smaller than a CPU cycle) (7.93x)

ARM32: 222ns (sd 0.118ns, negligibly small, smaller than a CPU cycle) (6.69x)
```

source: https://perfetto.dev/docs/design-docs/heapprofd-sampling#performance-considerations

Not sure how this would look on beefier x86 processors.

tests/perf/allocator_perf.cc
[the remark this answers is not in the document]
michoecho: I see.

include/seastar/core/memory.hh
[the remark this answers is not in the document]
michoecho: Abort, yes.

If we think aborting would be too dangerous/unexpected to the poor user who attempts to use it in production without testing it, then we could put a warning in the one place a user is guaranteed to read — the name of the function. Like `assert_there_are_no_allocations_and_set_heap_profiling_sampling_rate.`

@avikivity What's your opinion on this?

tests/unit/alloc_test.cc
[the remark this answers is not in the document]
michoecho: It probably is a bisectability concern, yes. But then again, if we really cared about bisectability, we would make the CI test all commits. Who knows how many bad commits we have out there.

@avikivity How much do we care about the bisectability here?

src/core/memory.cc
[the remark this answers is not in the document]
michoecho: > improving sesatar compile times

Modules will save us. R-right?

include/seastar/util/sampler.hh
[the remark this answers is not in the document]
michoecho: If we go with this logic, then I think you should also change the inequality `while (interval_to_next_sample_ <= 0)` to `< 0`. (Otherwise, e.g. when you sample allocations of size 1 with sampling interval 1, a sampled allocation will always be followed by an unsampled one).

(Yes, of course this doesn't matter for any practical purpose).

include/seastar/core/memory.hh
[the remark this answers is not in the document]
travisdowns: Perhaps the comment also makes it sound scarier than it is. I should clarify that re-enabling sampling with a different rate doesn't cause any crash, corruption or undefined behavior: rather you are simply mixing samples with different frequencies and so your the aggregate samples will be biased, i.e., the samples which were taken when the sampling rate were lower should have a higher weight in principle but nothing like that is implement.

So the comment there from @StephanDollberg is indicating what you would need to do to avoid this bias (not have any allocations, which would also mean no samples), which of course is not really feasible in any case.

My recommendation would be that we soften the comment to detail the "problems" with mixing sampling of different rates but that we don't actually block it because it would still be useful in some situations. FWIW, the way we use the profiler currently in Redpanda is that is always on, and at a fixed rate.

WDYT?

src/core/memory.cc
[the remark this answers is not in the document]
travisdowns: I will play with modules from the retirement home I'll be in I guess.

include/seastar/util/sampler.hh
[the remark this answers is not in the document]
travisdowns: I guess it depends heavily on the stdlib since with libstdc++ I measure them to have almost identical performance of ~66 ns for exponential and 70 ns for geometric. Looking and their implementations they are almost identical: both internally taking a floating point log of a random value.

In any case they should produce virtually identical results here?

tests/unit/alloc_test.cc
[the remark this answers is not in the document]
travisdowns: @michoecho - in the force push https://github.com/scylladb/seastar/commit/a7d7f691117f8971d240937bb789cd3e9d8c87ed I have re-written this commit to move the test changes into a later commit and checked that the re-written commits all compile and pass the alloc test.

include/seastar/util/sampler.hh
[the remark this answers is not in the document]
travisdowns: Good call! In force push https://github.com/scylladb/seastar/commit/a7d7f691117f8971d240937bb789cd3e9d8c87ed I have updated that loop to use `< 0`.

cc @StephanDollberg

src/core/memory.cc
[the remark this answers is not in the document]
travisdowns: > Does it? I was under the impression that `inline` is only for linkage nowadays.

It's the conventional wisdom in some sense that "it doesn't matter" but at least gcc definitely still uses it as input to its inlining heuristic: easy to show a case where it inlines only with inline.

Usually we get "implicit inline" by declaring the body inside the class but for whatever reason here the functions are not declared in the class body, but later on in the file so they lost the implicit inline.

include/seastar/core/memory.hh
```diff
@@ -405,21 +406,57 @@ public:
...
+
+/// @brief Copies the current set of allocation_sites in the output parameter
+/// (up to size of output vector)
+///
+/// Returns amount of copied elements. Useful if one wants to avoid allocating
+/// an output vector (e.g.: under OOM conditions)
+///
+/// @param output vector to copy the allocation sites to
+/// @return number of \ref allocation_site copied to the vector
+size_t memory_profile(std::vector<allocation_site>& output);
```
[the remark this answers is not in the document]
travisdowns: @michoecho thanks for the feedback. I've adjusted this to use a raw pointer + length as you've suggested here: https://github.com/scylladb/seastar/pull/1904/commits/b0f4726151790d0d7adc9daaca80473413737ff4.

I've also added some testing of the raw interface.

travisdowns:
@michoecho - I think all the feedback so far as been addressed, have a look when you have some time.

tests/unit/alloc_test.cc
```diff
@@ -311,6 +353,181 @@ SEASTAR_TEST_CASE(test_diagnostics_allocation) {
...
+
+    seastar::memory::set_heap_profiling_sampling_rate(0);
+
+    {
+        auto stats = seastar::memory::sampled_memory_profile();
+        BOOST_REQUIRE_EQUAL(stats.size(), 1000);
+    }
+
+    for (auto ptr : ptrs)
+    {
```
tchaikov, line 522:
```suggestion
    for (auto ptr : ptrs) {
```

tchaikov (commented):
i didn't get a chance to review this changeset properly. just left some comments on formatting.

travisdowns:
Force https://github.com/scylladb/seastar/commit/03e9966729c727d850795cc5facaa49f716cb1e3 was to update also the `sampled_memory_profile` stub method to the raw interface.

tests/unit/alloc_test.cc
StephanDollberg: Fixed

tests/unit/alloc_test.cc
```diff
@@ -19,6 +19,7 @@
  * Copyright (C) 2015 Cloudius Systems, Ltd.
  */
 
+#include "seastar/testing/perf_tests.hh"
```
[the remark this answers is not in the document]
StephanDollberg: Fixed, this particular one was actually unneeded.

include/seastar/core/memory.hh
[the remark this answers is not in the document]
StephanDollberg: I have added said check and updated the comment to be more clear. I have also added a test covering this scenario.

travisdowns:
@tchaikov and @michoecho I think the most recent round of feedback has been addressed. If you can let us know if there is anything outstanding that we haven't addressed it would be much appreciated!

src/core/memory.cc
```diff
@@ -1509,11 +1541,28 @@ void* allocate_from_small_pool(size_t size)
...
+ * Common handling code when a pointer (possibly null) has
+ * been returned by any allocation path.
+ */
+[[gnu::always_inline]]
+static inline void* finish_allocation(void* ptr, size_t size) {
+    if (!ptr) {
+        on_allocation_failure(size);
+    } else {
+#ifdef SEASTAR_DEBUG_ALLOCATIONS
+    std::memset(ptr, debug_allocation_pattern, size);
```
tchaikov, line 1626: nit, wrong indent.

tchaikov:
> @tchaikov and @michoecho I think the most recent round of feedback has been addressed. If you can let us know if there is anything outstanding that we haven't addressed it would be much appreciated!

@travisdowns hi Travis, i think we are pending on a maintainer (hey @avikivity !) to review and merge this change. since maintainers are consistently been preempted by other tasks, guess a visible latency is expected =(

avikivity:
I'm waiting for @michoecho's blessing

michoecho:
> I'm waiting for @michoecho's blessing

Ouch, I didn't think you would wait for me again. The first review was already an approval, after all. I only left some nitpicks, they didn't deserve to block anything in the first place.

I can't see the changes since the last review (since github already GCd the hash I reviewed...), but I see that my comments have been addressed, so — assuming that no regression has sneaked in in the meantime — I (still) approve.

Sorry for stalling this PR for so long. I have no excuse — I should keep better track of review requests.

tchaikov:
@avikivity hi Avi, ping?

avikivity:
> > I'm waiting for @michoecho's blessing
> 
> Ouch, I didn't think you would wait for me again. The first review was already an approval, after all. I only left some nitpicks, they didn't deserve to block anything in the first place.
> 
> I can't see the changes since the last review (since github already GCd the hash I reviewed...), but I see that my comments have been addressed, so — assuming that no regression has sneaked in in the meantime — I (still) approve.
> 
> Sorry for stalling this PR for so long. I have no excuse — I should keep better track of review requests.

The trick is to push the blame on the contributor. @travisdowns please ping vigorously if your patches don't make progress.

include/seastar/core/memory.hh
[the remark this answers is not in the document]
avikivity: Looks like it's still using vector&.

We don't generally include a bug/deficiency and fix it in a later patch in the same set, but fold them together.

src/core/memory.cc
avikivity: preexisting

include/seastar/core/memory.hh
```diff
@@ -443,15 +443,20 @@ struct allocation_site {
...
+///
+/// Returns amount of copied elements. This method does not allocation to
+/// it is useful alternative to \ref sampled_memory_profile() when one wants
+/// to avoid allocating (e.g.: under OOM conditions).
+///
+/// @param output array to copy the allocation sites to
+/// @param size the size of the array pointed to by \p output
 /// @return number of \ref allocation_site copied to the vector
-size_t sampled_memory_profile(std::vector<allocation_site>& output);
+size_t sampled_memory_profile(allocation_site* output, size_t size);
```
avikivity, line 459: Please fold this into the original patch.

travisdowns:
> The trick is to push the blame on the contributor.

:rofl: 

> @travisdowns please ping vigorously if your patches don't make progress.

Ack.

include/seastar/core/memory.hh
travisdowns: This is now folded in force push: https://github.com/scylladb/seastar/compare/58e4eb2dd3d67da55c0b76a186f045cd47ac3996..7a97ddfda2fec056975f00f20a8859c25bac41d9

travisdowns:
@avikivity your feedback is addressed.

I've also pulled into one very small edge-case fix we've applied downstream as: https://github.com/scylladb/seastar/pull/1904/commits/a3548bcbbfa3ec93c04e7d8a65646072c8a35173

travisdowns:
A big thank you for all the reviews and collaboration here. Hope this feature proves useful to you: it has already paid off nicely in Redpanda (the key is to emit some stacks in the OOM error message, which is up to the application currently).

src/core/memory.cc
tchaikov: fixed at #2025

avikivity:
This series was, in fact, mentioned a few ours ago on one of our OOM issues.

PR #1905, opened 2023-10-28
build: do not provide zlib as an ingredient

### LINKED ISSUE #1901: drop zlib from cooking_recipe.cmake
zlib is used as the software implementation of [compressdev](https://doc.dpdk.org/guides/prog_guide/compressdev.html), and its hardware accelerated counterpart is QAT.

probably we can drop it as we are not using it. or encourage the upstream to switch to zlib-ng if we plan to use this API in our application, but i doubt that we are going in this direction.

tchaikov:
zlib is used as the software implementation of SPDK's compressdev, and its hardware accelerated counterpart is QAT.

we can drop it as we are not using it, neither do we plan to use this DPDK API in our, so let's stop providing it as an ingredient.

Fixes #1901

### DIFF
--- a/cooking_recipe.cmake
+++ b/cooking_recipe.cmake
@@ -119,14 +119,6 @@ cooking_ingredient (numactl
     BUILD_COMMAND <DISABLE>
     INSTALL_COMMAND ${make_command} install)
 
-cooking_ingredient (zlib
-  EXTERNAL_PROJECT_ARGS
-    URL https://github.com/madler/zlib/releases/download/v1.2.13/zlib-1.2.13.tar.gz
-    URL_MD5 9b8aa094c4e5765dabf4da391f00d15c
-    CONFIGURE_COMMAND <SOURCE_DIR>/configure --prefix=<INSTALL_DIR>
-    BUILD_COMMAND <DISABLE>
-    INSTALL_COMMAND ${make_command} install)
-
 ##
 ## Private and private/public dependencies.
 ##


tchaikov:
tested with following steps:
```console
$ sudo dnf remove zlib-devel
$ ./configure.py --enable-dpdk --mode debug
$ ninja -C build/debug/ seastar
```

nyh (approved):
I approve this patch, but I have to wonder about the "cooking" whole thing (which is what you're testing): Is anyone still using this approach? Did you check that before - or after - your patch it the "cooking.sh" script actually works?

tchaikov:
> I approve this patch, but I have to wonder about the "cooking" whole thing (which is what you're testing): Is anyone still using this approach? Did you check that before - or after - your patch it the "cooking.sh" script actually works?

Regarding cooking the dpdk, i tested it when migrating to a newer version dpdk. And yes, I also tested it before and after this very change.

mykaul:
Strange, on my laptop the configure part fails consistently:
```
-- Configuring done (11.4s)
-- Generating done (0.2s)
-- Build files have been written to: /home/ykaul/github/scylladb/build/debug/seastar
Traceback (most recent call last):
  File "/home/ykaul/github/scylladb/./configure.py", line 2384, in <module>
    generate_compdb('compile_commands.json', ninja, buildfile, selected_modes)
  File "/home/ykaul/github/scylladb/./configure.py", line 302, in generate_compdb
    os.symlink(compdb_target, compdb)
FileExistsError: [Errno 17] File exists: 'build/debug/compile_commands.json' -> 'compile_commands.json'
```

(of course I've deleted 'build' dir before the run)

tchaikov:
@mykaul Yaniv, i think this it not quite related to this change. it's more relevant to scylladb. but i don't have theory explaining this, as we always check if `compile_commands.json`, and return if it already exists. could you check if your `compile_commands.json` exits in your current directory? if this problem persists, i'd suggest file an issue against scylladb/scylladb with the reproducing steps.

nyh:
>  but i don't have a theory explaining this, as we always check if `compile_commands.json`, and return if it already exists. could you check if your `compile_commands.json` exits in your current directory? if this problem persists, i'd suggest file an issue against scylladb/scylladb with the reproducing steps.

A common culprit causing such messages (but I didn't try to debug this one - I never had this problem) is concurrent compilation: Ninja compiles two targets in parallel, both of them depend on compile_commands.json (?), so two processes are started to create it, and one will complain the file already exists. But as I said, I don't know if this actually happens - it's just a theory.

tchaikov:
yeah, this is an interesting direction to explore. but `compile_commands.json` is not part of the building system in `build.ninja`, it is only created by `configure.py`. unless Yaniv runs multiple instances of `configured.py` in parallel, i don't think they can step on each other due to this racing condition.

nyh:
Note that the error quoted by @mykaul  is from Scylla's build, not Seastar, file reported, `build/debug/compile_commands.json`, isn't in Seastar's build directory, so all of this might be happening in Scylla's build, not Seastar's. Or maybe worse - these two build systems might be trying to create a file with the same name in the same place?

mykaul:
> Note that the error quoted by @mykaul is from Scylla's build, not Seastar, file reported, `build/debug/compile_commands.json`, isn't in Seastar's build directory, so all of this might be happening in Scylla's build, not Seastar's. Or maybe worse - these two build systems might be trying to create a file with the same name in the same place?

That was my suspicion as well.

tchaikov:
created https://github.com/scylladb/scylladb/pull/15870 to address this issue.

tchaikov:
> Note that the error quoted by @mykaul is from Scylla's build, not Seastar, file reported, `build/debug/compile_commands.json`, isn't in Seastar's build directory, so all of this might be happening in Scylla's build, not Seastar's. Or maybe worse - these two build systems might be trying to create a file with the same name in the same place?

i don't think so. seastar's `configure.py` creates `compile_commands.json` under cmake's (and ninja's) build directory for seastar, like`build/{mode}/seastar/compile_commands.json`, and it does not create a symlink for it. while scylladb's `configure.py` creates this file under ninja's build directory, like `build/{mode}/compile_commands.json`, and create a symlink from current source directory to one of `build/*/compile_commands.json`.

PR #1908, opened 2023-10-30
fair_queue::unregister_priority_class:fix assertion

### LINKED ISSUE #1907: Assertion in fair_queue::unregister_priority_class fails when read ahead is enabled
Following program:
```
#include <seastar/core/fstream.hh>
#include <seastar/core/loop.hh>
#include <iostream>

using namespace std;
using namespace seastar;

    app_template app;
    app.run(argc, argv, [&app] {
        return with_file_close_on_failure(
            open_file_dma("random_input_file", open_flags::ro),
            [] (file f) {
                auto stream = make_file_input_stream(std::move(f), {1024, 2});
                auto istream = make_lw_shared(std::move(stream));
                return repeat([istream] () {
                    return istream->read().then([] (auto&& buf) {
                        return stop_iteration{buf.empty()};
                    });
                }).finally([istream = std::move(istream)] {
                    return istream->close().then([istream] {});
                });
            }
        );
    });
}
```
compiled with:
```
g++ $prog_name.cc \
    $(pkg-config --libs --cflags --static $seastar_dir/build/release/seastar.pc) \
    -o $prog_name
```
and run with:
```
./$prog_name -c 1
```
on file generated by:
```
dd if=/dev/urandom of=random_input_file count=1 bs=1024
```
fails with:
```
seastar/src/core/fair_queue.cc:293: void seastar::fair_queue::unregister_priority_class(class_id): Assertion `pclass && pclass->_queue.empty()' failed.
```

It seems that this is caused by cancelled requests that are not finished yet when the program exits.

haaawk:
It is possible that the priority class is unregistered before all cancelled requests in that class have ended.

Following program that reads a file with read ahead causes current assertion to fail:

```

using namespace std;
using namespace seastar;

    app_template app;
    app.run(argc, argv, [&app] {
        return with_file_close_on_failure(
            open_file_dma("random_input_file", open_flags::ro),
            [] (file f) {
                auto stream = make_file_input_stream(std::move(f), {size_t{1}<<10, 2});
                auto istream = make_lw_shared(std::move(stream));
                return repeat([istream] () {
                    return istream->read().then([] (auto&& buf) {
                        return stop_iteration{buf.empty()};
                    });
                }).finally([istream = std::move(istream)] {
                    return istream->close().then([istream] {});
                });
            }
        );
    });
}
```

This patch excludes cancelled requests from the assertion check.

Fixes #1907

### DIFF
--- a/src/core/fair_queue.cc
+++ b/src/core/fair_queue.cc
@@ -290,7 +290,7 @@ void fair_queue::register_priority_class(class_id id, uint32_t shares) {
 
 void fair_queue::unregister_priority_class(class_id id) {
     auto& pclass = _priority_classes[id];
-    assert(pclass && pclass->_queue.empty());
+    assert(pclass);
     pclass.reset();
     _nr_classes--;
 }
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -661,6 +661,7 @@ io_queue::~io_queue() {
     //
     // And that will happen only when there are no more fibers to run. If we ever change
     // that, then this has to change.
+    assert(_queued_requests == 0);
     for (auto&& pc_data : _priority_classes) {
         if (pc_data) {
             for (auto&& s : _streams) {


haaawk:
@xemul Would you mind having a look please?

src/core/fair_queue.cc
```diff
@@ -159,6 +159,14 @@ class fair_queue::priority_class_data {
     void update_shares(uint32_t shares) noexcept {
         _shares = (std::max(shares, 1u));
     }
+
+    bool has_noncancelled_entry() const {
+        bool result = false;
+        for (auto& ent : _queue) {
+            result |= ent._capacity > 0;
+        }
+        return result;
```
tchaikov, line 168:
```suggestion
        auto found = boost::range::find_if(_queue, [] (const auto& ent) {
            return ent._capacity > 0;
        });
        return found != _queue.end();
```
no need to continue if one of the non-cancelled entry is found.

tchaikov (commented):
might want to add this test as a unit test.

xemul:
> It is possible that the priority class is unregistered before all cancelled requests in that class have ended.

Why does "cancelled" matter here? I mean -- classes are unregistered on app exit, in ~io_queue, so the same bug should happen if there are non-cancelled and non-waited-upon requests in the queue.

xemul:
There's even comment inside io_queue destructor

```
io_queue::~io_queue() {
    // It is illegal to stop the I/O queue with pending requests.
    // Technically we would use a gate to guarantee that. But here, it is not
    // needed since this is expected to be destroyed only after the reactor is destroyed.
    //     
    // And that will happen only when there are no more fibers to run. If we ever change
    // that, then this has to change.
```

apparently requests cancellation broke that, but (my bad) I don't get why :(

src/core/fair_queue.cc
haaawk: That was my initial version but it makes error case faster by making correct/happy/usual path slower by introducing branching inside the loop. I think it's better to have faster happy path instead of faster error path which ends in assertion crashing anyway.

haaawk:
> might want to add this test as a unit test.

I tried that but didn't manage to reproduce it using SEASTAR_TEST :(. Probably testing machinery adds enough delay at shutdown to make it impossible to reproduce.

haaawk:
> > It is possible that the priority class is unregistered before all cancelled requests in that class have ended.
> 
> Why does "cancelled" matter here? I mean -- classes are unregistered on app exit, in ~io_queue, so the same bug should happen if there are non-cancelled and non-waited-upon requests in the queue.

Correct code never leaves ongoing requests at exit but cancelled requests don't account as ongoing.

haaawk:
I think cancelled requests are fine @xemul.
Just the assertion was wrong/old.
When @asias introduced the assertion there was no concept of cancellation I believe.

xemul:
> Correct code never leaves ongoing requests at exit but cancelled requests don't account as ongoing.

You mean that any place that submitted dma_read/dma_write must .get() the returned future by itself?

haaawk:
> > Correct code never leaves ongoing requests at exit but cancelled requests don't account as ongoing.
> 
> You mean that any place that submitted dma_read/dma_write must .get() the returned future by itself?

.get() or sth like this https://github.com/scylladb/seastar/blob/master/src/core/fstream.cc#L209

src/core/fair_queue.cc
```diff
@@ -159,6 +159,14 @@ class fair_queue::priority_class_data {
     void update_shares(uint32_t shares) noexcept {
         _shares = (std::max(shares, 1u));
     }
+
+    bool has_noncancelled_entry() const {
+        bool result = false;
+        for (auto& ent : _queue) {
+            result |= ent._capacity > 0;
```
xemul, line 166: There's an opinion that `|=` doesn't work for bool types:
https://github.com/scylladb/seastar/pull/1841#discussion_r1331386016
https://stackoverflow.com/questions/2488406/why-doesnt-c-have-or-for-booleans

src/core/fair_queue.cc
```diff
@@ -290,7 +298,7 @@ void fair_queue::register_priority_class(class_id id, uint32_t shares) {
 
 void fair_queue::unregister_priority_class(class_id id) {
     auto& pclass = _priority_classes[id];
-    assert(pclass && pclass->_queue.empty());
+    assert(pclass && !pclass->has_noncancelled_entry());
```
xemul, line 301: I think that this line deserves a comment summing up recent discussion here -- despite it's technically OK to destruct the class with queued requests, we still want to make sure that the rest of the code picked up the requests it queued previously

src/core/fair_queue.cc
xemul: If we aim at speed and think that there can be canceled requests in the queue, then it's worth checking `io_queue::_queued_requests` to become 0 in `~io_queue()` and drop the queue being empty part of this assertion

src/core/fair_queue.cc
haaawk: I think |= works as long as everything is bool, no?

> A bool may only be true or false in C++. As such, using &= and |= is relatively safe (even though I don’t particularly like the notation). True, they will perform bit operations rather than logical operations (and thus they won’t short-circuit) but these bit operations follow a well-defined mapping, which is effectively equivalent to the logical operations, as long as both operands are of type bool.1

src/core/fair_queue.cc
haaawk: Added a comment.

src/core/fair_queue.cc
xemul: I don't know:

> "&=" is not the right operator, it's the bitwise operator :-(
> I know some people claim it works (https://stackoverflow.com/questions/2488406/why-doesnt-c-have-or-for-booleans) but I think it's wrong.

For me it works, but Nadav (in comment) correctly says that |= is not logical, but a bitwise operator. The correct one here should be ||=, but there's no such in C++

haaawk: there's a big difference between &= and |=. &= won't work for bool_var &= 3 but |= always works.

xemul: I don't see the problem even with &= in the first place. Compiler knows its bool values and can handle it correctly. However, if we try to find interpretation that leads to broken behavior then even for `|= 3` things can go wrong
```
bool x = true; // x = 0x1
x |= 3; // x = 0x3
x == true; // 0x3 == 0x1 -> can be either true or false
```

xemul: So I'll leave it up to our compilers expert (@tchaikov ). Kefu, do you think the problem is invented out of nowhere or should indeed avoid |= here?

haaawk: I don't think `x == true` can ever be false if x is true. No matter what the underlying value of x is. This is bool comparison so all values have to be casted to bool otherwise this wouldn't work either:
```
bool a = 1;
bool b = 2;
assert(a == b);
```

tchaikov: if i were a C++17 compiler implemented using static single assignment, and am requested to compile following source list:
```c++
bool x = true;
x |= 3;
x == true;
```
i would first translate it to something like
```c++
bool x = true;
int x1 = static_cast<int>(x); // x1 = 0x1, as x should be promoted. so that it can participate as the LHS of `|` operation.
int x2 = x1|3; // x2 = 0x1 | 0x3 = 3
bool x3 = static_cast<bool>(x2); // x3 = true. as it casted back to the type of the LHS
bool y = x3 == true; // x3 == true
```

see also https://eel.is/c++draft/expr.ass#6 and https://eel.is/c++draft/expr#conv.bool

probably we can just sum up the `ent._capacity` so we can avoid the branching in the tight loop, if the sum  does not overflow.

haaawk:
Added assertion to ~io_queue and removed the check from unregister_priority_class.

src/core/fair_queue.cc
tchaikov: > I don't think `x == true` can ever be false if x is true. No matter what the underlying value of x is. This is bool comparison so all values have to be casted to bool otherwise this wouldn't work either:
> 
> ```
> bool a = 1;
> bool b = 2;
> assert(a == b);
> ```

yeah, you beat me to it. i concur with your point. just one thing the "underlying value of x", from the compiler's point of view should always be `true` or `false`. but from the machine's perspective, it's always 1 or 0.

Issue #1677, opened 2023-05-29
coroutine generator use-after-free


xemul:
Consider a code (can be added as a test case to coroutines_test.cc):

```c++
struct wrapped_string {
    sstring s;
};

coroutine::experimental::generator<wrapped_string, buffered_container> generate_wrapped_strings(coroutine::experimental::buffer_size_t size, int nr) {
    for (int i = 0; i < nr; i++) {
        sstring x = format("abcdefghijklmnopqrstuvwxyz-{}", i);
        co_yield wrapped_string(std::move(x));
    }
}

SEASTAR_TEST_CASE(test_generator_lref) {
    auto sgen = generate_wrapped_strings(coroutine::experimental::buffer_size_t{8}, 42);
    while (auto ws = co_await sgen()) {
        fmt::print("{}\n", ws->s);
    }
}
```

First, it's important that `sstring x` is _that_ long -- it must be external.

The bug -- if run in debug mode it crashes with heap use-after-free message:

* the memory was allocated where `sstring x` variable was declared -- that is sstring's external memory buffer.
* the memory was freed at the `generate_wrapped_string()`'s closing brace -- that is when the `x` went out of scope of the generator
* the use-after-free happens when `ws->s` is printed in the test case function

It seems that `ws->s` should be eventually copy-constructed or move-constructed from `x`, but in fact `ws` and `x` share the pointer to the external buffer for some reason.

There are two options how to work around that bug.

1. Declare constructor for `wrapped_string`

```c++
struct wrapped_string {
    sstring s;
    wrapped_string(sstring s_) : s(std::move(s_)) {}
};
```

2. Make `wrapped_string` local variable in generator

```c++
    for (int i = 0; i < nr; i++) {
        sstring x = format("abcdefghijklmnopqrstuvwxyz-{}", i);
        auto ws = wrapped_string(std::move(x));
        co_yield ws;
    }
```

cc @avikivity  @tchaikov

avikivity:
Looks like a miscompile, no? with what clang do you see it?

xemul:
> Looks like a miscompile, no? 

Yes. `ws` and `x` have different `this` values, but strings inside point to the same malloc-ed buffer. It all can be explained by "compiler memcpy-ed `x` into `ws` instead of copy/move constructing it"

> with what clang do you see it?

```
$ clang++ --version
clang version 15.0.7 (Fedora 15.0.7-2.fc37)
```

avikivity:
[...]

PR #1909, opened 2023-10-31
file: construct directory_entry with aggregated ctor

tchaikov:
despite that `directory_entry` can be default constructed, it does not have a contructor which accepts two parameters. so, in order to construct an instance of it, we either need to use the aggregated contructor, or the list-initialization.

in this change, in order to appease the compiler which does not accept the form of construct, let's use the aggregated constructor.

this should address the FTBFS with Clang-15 + C++20
```
/home/circleci/project/src/core/file.cc:417:29: error: no matching constructor for initialization of 'seastar::directory_entry'
            directory_entry ret(std::move(name), type);
                            ^   ~~~~~~~~~~~~~~~~~~~~~
/home/circleci/project/include/seastar/core/file.hh:58:8: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 2 were provided
struct directory_entry {
       ^
/home/circleci/project/include/seastar/core/file.hh:58:8: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 2 were provided
/home/circleci/project/include/seastar/core/file.hh:58:8: note: candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 2 were provided
1 error generated.
```

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -414,7 +414,8 @@ static coroutine::experimental::generator<directory_entry, dir_entry_buffer> mak
                 continue;
             }
             std::optional<directory_entry_type> type = dirent_type(*de);
-            directory_entry ret(std::move(name), type);
+            // See: https://github.com/scylladb/seastar/issues/1677
+            directory_entry ret{std::move(name), type};
             co_yield ret;
         }
     }


tchaikov:
this should address the FTBFS found at https://app.circleci.com/pipelines/github/scylladb/seastar/2350/workflows/1c51deb8-c610-4b95-897d-8e709e68ea73/jobs/14941

xemul:
@tchaikov , this code looks this way because I stepped on #1677 with my compiler back then. It's just FYI, if supported compilers work correctly, I will be happy to merge this PR (so far CI fails as far as I can see)

tchaikov:
> @tchaikov , this code looks this way because I stepped on #1677 with my compiler back then. It's just FYI, if supported compilers work correctly, I will be happy to merge this PR (so far CI fails as far as I can see)

@xemul hi Pavel, i see. just to be on the safe side, could you please help review the other two changes: #1910 and #1911?

tchaikov:
turns out #1911 alone does not help to address the FTBFS, i am including it here to see if these two changes together can cure the build.

tchaikov:
> > @tchaikov , this code looks this way because I stepped on #1677 with my compiler back then. It's just FYI, if supported compilers work correctly, I will be happy to merge this PR (so far CI fails as far as I can see)
> 
> @xemul hi Pavel, i see. just to be on the safe side, could you please help review the other two changes: ~#1910~ and #1911? once they are reviewed and merged. if this change still breaks the build or test, i will go with your route: create a temporary object and return it.

yeah, it is reproduciable. see https://app.circleci.com/pipelines/github/tchaikov/seastar/906/workflows/c03e7522-8271-4db3-85ef-d70f42cf5128/jobs/10883

i reverted to your workaround.

xemul:
Hm.. CI still crashes on the use-after-free in directory listing test

PR #1911, opened 2023-10-31
file: s/ino64_t/ino_t/, s/off64_t/off_t/

tchaikov:
turns out 9e37a6fc overwrote the changes made by 2638c51cc3, so lets resurrect them.


tchaikov:
this change is a subset of #1909, so that we can verify that it helps to address the FTBFS.

PR #1893, opened 2023-10-24
io_tester: disable -Wuninitialized when including boost.accumulators

tchaikov:
io_tester: disable -Wuninitialized when including boost.accumulators

before this change, when compiling boost.accumulators headers, if
the library does not include the fix for
https://github.com/boostorg/accumulators/pull/54. and we are building
the release build with GCC-13, we'd have following -Wuninitialized build
failure:

```
In copy constructor ‘constexpr boost::accumulators::impl::extended_p_square_quantile_impl<double, boost::accumulators::unweighted, boost::accumulators::quadratic>::extended_p_square_quantile_impl(const boost::accumulators::impl::extended_p_square_quantile_impl<double, boost::accumulators::unweighted, boost::accumulators::quadratic>&)’,
    inlined from ‘boost::accumulators::detail::accumulator_wrapper<Accumulator, Feature>::accumulator_wrapper(const boost::accumulators::detail::accumulator_wrapper<Accumulator, Feature>&) [with Accumulator = boost::accumulators::impl::extended_p_square_quantile_impl<double, boost::accumulators::unweighted, boost::accumulators::quadratic>; Feature = boost::accumulators::tag::extended_p_square_quantile_quadratic]’ at /usr/include/boost/accumulators/framework/depends_on.hpp:320:69,
    inlined from ‘constexpr boost::fusion::cons<Car, Cdr>::cons(typename boost::fusion::detail::call_param<Car>::type, typename boost::fusion::detail::call_param<Cdr>::type) [with Car = boost::accumulators::detail::accumulator_wrapper<boost::accumulators::impl::extended_p_square_quantile_impl<double, boost::accumulators::unweighted, boost::accumulators::quadratic>, boost::accumulators::tag::extended_p_square_quantile_quadratic>; Cdr = boost::fusion::cons<boost::accumulators::detail::accumulator_wrapper<boost::accumulators::impl::sum_impl<double, boost::accumulators::tag::sample>, boost::accumulators::tag::sum>, boost::fusion::cons<boost::accumulators::detail::accumulator_wrapper<boost::accumulators::impl::mean_impl<double, boost::accumulators::tag::sum>, boost::accumulators::tag::mean>, boost::fusion::cons<boost::accumulators::detail::accumulator_wrapper<boost::accumulators::impl::max_impl<double>, boost::accumulators::tag::max>, boost::fusion::nil_> > >]’ at /usr/include/boost/fusion/container/list/cons.hpp:66:15,
[...]

PR #1912, opened 2023-10-31
rpc_tester: Disable -Wuninitialized when including boost.accumulator

xemul:
The same fix was applied to io_tester some time ago -- #1893. Now its rpc_tester turn as it includes the very same set of headers.


tchaikov (approved):
lgtm

PR #1630, opened 2023-04-26
docker: bump up to clang {16,17} and gcc {12,13}

tchaikov:
since we only support the latest two major releases of compilers. at the moment of writing, Clang just released v16, and the latest major release of GCC is v13.

so we should install them respectively.

because ubuntu kinetic does not ship clang-16. we need to bump up the base image from ubuntu:kinetic to ubuntu:lunar despite that lunar is not an LTS release. we should use ubuntu 24.04 once it's out.

see also 80969ef9ffc10bb219bd3ef83ab76c2c536de7ec, which bumped the compilers also.

Issue #1913, opened 2023-11-01
directory_test is flaky


tchaikov:
it fails due to one of following failures:

`co_yield` gets called again even if the previous returned `yield_awaiter` is not ready, and that not-ready awaiter has not been resumed:

```
directory_test: /home/kefu/dev/seastar/include/seastar/coroutine/generator.hh:186: yield_awaiter<T, Container> seastar::coroutine::experimental::internal::generator_buffered_promise<seastar::directory_entry, seastar::dir_entry_buffer>::yield_value(U &&) [T = seastar::directory_entry, Container = seastar::dir_entry_buffer, U = seastar::directory_entry &]: Assertion `!_wait_for_free_space' failed.
```

a task / promise is run even after it is deleted. this can be reproduced with `-c 1`:
```
==2432222==ERROR: AddressSanitizer: heap-use-after-free on address 0x617000000e90 at pc 0x5584c4fb79b5 bp 0x7ffc9418e070 sp 0x7ffc9418e068                                                                                                    
READ of size 8 at 0x617000000e90 thread T0                                                                                                                                                                                                    
    #0 0x5584c4fb79b4 in seastar::reactor::run_tasks(seastar::reactor::task_queue&) /home/kefu/dev/seastar/src/core/reactor.cc:2661:14                                                                                                        
    #1 0x5584c4fbfb3e in seastar::reactor::run_some_tasks() /home/kefu/dev/seastar/src/core/reactor.cc:3124:9
    #2 0x5584c4fc3a1d in seastar::reactor::do_run() /home/kefu/dev/seastar/src/core/reactor.cc:3293:9
    #3 0x5584c4fc16f3 in seastar::reactor::run() /home/kefu/dev/seastar/src/core/reactor.cc:3176:16
    #4 0x5584c4d6fb2a in seastar::app_template::run_deprecated(int, char**, std::function<void ()>&&) /home/kefu/dev/seastar/src/core/app-template.cc:276:31
    #5 0x5584c4d6d0ca in seastar::app_template::run(int, char**, std::function<seastar::future<int> ()>&&) /home/kefu/dev/seastar/src/core/app-template.cc:167:12
    #6 0x5584c4d701ba in seastar::app_template::run(int, char**, std::function<seastar::future<void> ()>&&) /home/kefu/dev/seastar/src/core/app-template.cc:182:12
    #7 0x5584c4b496d7 in main /home/kefu/dev/seastar/tests/unit/directory_test.cc:145:27
[...]

Issue #1716, opened 2023-07-03
Seastar cannot be compiled with gcc 13.1 and fmt lib version 10


raphaelsc:
```
In static member function ‘static constexpr _Up* std::__copy_move<_IsMove, true, std::random_access_iterator_tag>::__copy_m(_Tp*, _Tp*, _Up*) [with _Tp = unsigned int; _Up = unsigned int; bool _IsMove = false]’,
    inlined from ‘constexpr _OI std::__copy_move_a2(_II, _II, _OI) [with bool _IsMove = false; _II = unsigned int*; _OI = unsigned int*]’ at /usr/include/c++/13/bits/stl_algobase.h:506:30,
    inlined from ‘constexpr _OI std::__copy_move_a1(_II, _II, _OI) [with bool _IsMove = false; _II = unsigned int*; _OI = unsigned int*]’ at /usr/include/c++/13/bits/stl_algobase.h:533:42,
    inlined from ‘constexpr _OI std::__copy_move_a(_II, _II, _OI) [with bool _IsMove = false; _II = unsigned int*; _OI = unsigned int*]’ at /usr/include/c++/13/bits/stl_algobase.h:540:31,
    inlined from ‘constexpr _OI std::copy(_II, _II, _OI) [with _II = unsigned int*; _OI = unsigned int*]’ at /usr/include/c++/13/bits/stl_algobase.h:633:7,
    inlined from ‘static _ForwardIterator std::__uninitialized_copy<true>::__uninit_copy(_InputIterator, _InputIterator, _ForwardIterator) [with _InputIterator = unsigned int*; _ForwardIterator = unsigned int*]’ at /usr/include/c++/13/bits/stl_uninitialized.h:147:27,
    inlined from ‘_ForwardIterator std::uninitialized_copy(_InputIterator, _InputIterator, _ForwardIterator) [with _InputIterator = unsigned int*; _ForwardIterator = unsigned int*]’ at /usr/include/c++/13/bits/stl_uninitialized.h:185:15,
    inlined from ‘constexpr void fmt::v9::basic_memory_buffer<T, SIZE, Allocator>::grow(size_t) [with T = unsigned int; long unsigned int SIZE = 32; Allocator = std::allocator<unsigned int>]’ at /usr/include/fmt/format.h:925:26,
    inlined from ‘constexpr void fmt::v9::detail::buffer<T>::try_reserve(size_t) [with T = unsigned int]’ at /usr/include/fmt/core.h:928:39,
[...]

PR #1915, opened 2023-11-02
docker: bump up to clang {15,16} and gcc {12,13}

avikivity:
since we only support the latest two major releases of compilers. at the moment of writing, Clang just released v16, and the latest major release of GCC is v13.

so we should install them respectively.

because ubuntu kinetic does not ship clang-16. we need to bump up the base image from ubuntu:kinetic to ubuntu:lunar despite that lunar is not an LTS release. we should use ubuntu 24.04 once it's out.

see also 80969ef9ffc10bb219bd3ef83ab76c2c536de7ec, which bumped the compilers also.

Closes scylladb/seastar#1630

[avi: update toolchain]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ jobs:
       - checkout
       - run: git submodule sync
       - run: git submodule update --init
-      - run: echo 'docker run --pids-limit -1 --security-opt seccomp=unconfined --network host --user "$(id -u):$(id -g)" --rm -v $PWD:$PWD -w $PWD  docker.io/scylladb/seastar-toolchain:2023-08-13 "$@"' > run; chmod +x run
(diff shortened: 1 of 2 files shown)


avikivity:
@xemul @tchaikov tests are failing with the new toolchain. Are those flaky tests or did the new toolchain expose bugs?

mykaul:
Isn't clang 17 already available?
https://discourse.llvm.org/t/llvm-17-0-4-released/74548

tchaikov:
@avikivity it's a known issue. see https://github.com/scylladb/seastar/issues/1913 and https://github.com/scylladb/seastar/issues/1917.

 the GCC warning is tracked by https://github.com/scylladb/seastar/issues/1716. i will create a PR to silence it.

tchaikov:
> Isn't clang 17 already available? https://discourse.llvm.org/t/llvm-17-0-4-released/74548

ahh, right. the original PR #1630 was created half a year ago. i just updated it to use ubuntu:mantic and clang-17.

avikivity:
But please check the failures.

tchaikov:
> But please check the failures.

i did. see https://github.com/scylladb/seastar/pull/1915#issuecomment-1791719028

PR #1919, opened 2023-11-04
build: generate config-specific rules using generator expressions

tchaikov:
in this series, we start to use  generator expressions to set config-specific rules instead of checking `CMAKE_BUILD_TYPE`. if we check `CMAKE_BUILD_TYPE` for setting the compiling options, we won't be able to support multi-config generator.

another missing puzzle to support Multi-Config is to generate different options or different `seastar.pc`, it is addressed by #1929.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -919,10 +919,12 @@ endif ()
 # as many compilation units as possible.  So while these flags don't
 # have to be public, we don't expect anyone to want to build seastar
 # with them and some client code without.
-if (Seastar_SPLIT_DWARF AND (NOT (CMAKE_BUILD_TYPE STREQUAL "Dev")))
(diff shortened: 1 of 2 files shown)


tchaikov:
@denesb hi Botond, could you please help review this change? i'd like to implement the Multi-Config generator support in baby steps. the next step is to address the .pc and/or .cmake file problem, as with Multi-Config generators, we need to generate different .pc file for each of the enabled build mode.

tchaikov:

tchaikov:
@nyh hi Nadav, is this a blocker for merging this change? it'd be great if we can have it in seastar sooner. as i have a couple pending changes on scylla's end which depend on this one, if we can have it. i would be able to progress in another front.

PR #1924, opened 2023-11-06
perftune.py: skip discovering IRQs for iSCSI disks

syuu1228:
Currently, when we use iSCSI disk for the data volume, we get following error while running "--tune disks":

 ERROR: [Errno 2] No such file or directory: '/sys/devices/platform/modalias'. Your system can't be tuned until the issue is fixed.

This is because iSCSI is virtual device, it does not pointed to /sys/devices/pciXXX, and does not have IRQs informations.

We should just skip discovering IRQs for such devices, instead of calling learn_all_irqs_one().

Fixes scylladb/scylladb#10367


tchaikov (approved):
lgtm. see also https://github.com/open-iscsi/open-iscsi/blob/master/usr/iscsi_sysfs.c

avikivity:
@xemul please fix the failing tests

Issue #1899, opened 2023-10-26
Assertion `pc._plugged && pc._queued' failed


xemul:
There's a bug in unplugging IO priority class in duplex mode

```c++
void fair_queue::unplug_priority_class(priority_class_data& pc) noexcept {
    assert(pc._plugged);
    if (pc._queued) {
        pop_priority_class(pc);
    }
    pc._plugged = false;
}
```

the thing is that `pop_...` thing only works on the class that is on top of the handles queue. Usually this constraint is satisfied, because unplug happens from throttle, which in turn happens when dispatching a requests and dispatch happens from top class only. However, in duplex mode throttling tries to unplug classes from two streams, and the "other" one may happen not to be on top of its handles

xemul:
Conf file for reproducing with io-tester (io-properties must exist too with duplex=true)

```
- name: a
  shards: all
  type: seqwrite
  data_size: 1GB
  shard_info:
    parallelism: 1
    reqsize: 256kB
    shares: 10
    bandwidth: 10MB

- name: b
  shards: all
  type: randread
  data_size: 1GB
  shard_info:
    parallelism: 1
    reqsize: 4kB
    shares: 100
    bandwidth: 10MB

- name: c
  shards: all
  type: randread
  data_size: 1GB
  shard_info:
    parallelism: 1
    reqsize: 4kB
    class: a

- name: d
  shards: all
  type: seqwrite
  data_size: 1GB
  shard_info:
    parallelism: 1
    reqsize: 128kB
    class: b
    bandwidth: 10MB
```

PR #1925, opened 2023-11-06
fair_queue: Do not pop unplugged class immediately

xemul:
A fair queue's priority class have two bits -- queued and plugged.

Queued means that the class is on the priority queue and dispatching loop will eventually pick it up from there. When false this means that the class is not on the handles list. The bit protects the class from being enqueued more than once.

Dequeueing (called popping) has a trickery -- it can only happen on the class that's on top of the dispatch queue. That's because the dispatch queue queue is std::priority_queue which doesn't have random access eraser call, but only have the .pop() one.

Plugged bit is controlled by the bandwidth throttler. When the class runs out of its throttling tokens, it's unplugged. When it replenishes the necessary amount of tokens backm, it's plugged back. Currently, unplugging the class also dequeues it from the dispatch list.

The aforementioned dequeuing limitation works with unplugging most of the time, because unplug can only happen from inside the dispatch loop and and the dispatch loop first picks the class from the top of the queue. So unplugged class is automagically the top one.

However, if there's duplex mode turned on, then each class is present in two fair queues (called streams) and unplugging a class means unplugging it from both fair queues. And even if the priority class being dispatched is indeed the top one, its couterpart from another stream is not, and popping it while unplugging is wrong and leads to broken pop -- the correct class is maked as unplugged and dequeued, but it's _not_ removed from the dispatch list, instead some other class is pooped from it. Later dispatch loop would find this other class in the dispatch queue makred unplugged and dequeued and would crash on assertion.

The fix is to pop classes lazily. It means that unplugged classes are just marked as such and that's it. Later, when the dispatch loop picks up the unplugged class is will pop it instantly and proceed.

fixes: #1899

Tested with 10 minutes of io-tester with the reproducer config attached to the aforementioned issue.

### DIFF
--- a/src/core/fair_queue.cc
+++ b/src/core/fair_queue.cc
@@ -215,14 +215,15 @@ void fair_queue::push_priority_class_from_idle(priority_class_data& pc) noexcept
     }
 }
 
+// ATTN: This can only be called on pc that is from _handles.top()
 void fair_queue::pop_priority_class(priority_class_data& pc) noexcept {
-    assert(pc._plugged && pc._queued);
+    assert(pc._queued);
     pc._queued = false;
     _handles.pop();
 }
 
 void fair_queue::plug_priority_class(priority_class_data& pc) noexcept {
-    assert(!pc._plugged && !pc._queued);
+    assert(!pc._plugged);
     pc._plugged = true;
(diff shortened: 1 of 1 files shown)


mykaul:
@avikivity - please review.

PR #1927, opened 2023-11-07
gate: correct the typo in doxygen comment

tchaikov:
gate::get_holder never exists, it should be gate::hold.


tchaikov:
@bhalevy hi Benny, could you please help review this change?

bhalevy (approved):
yes, thanks.

IIRC it existed in an early draft, but was indeed replaced by `hold()`

Issue #1926, opened 2023-11-06
shared_promise is not available immediately after set_value/exception if it has outstanding shared futures


bhalevy:
This issue is derived from https://github.com/scylladb/scylladb/issues/15956

The following example fails:
```c++
SEASTAR_THREAD_TEST_CASE(test_shared_promise_with_outstanding_future_is_immediately_available) {
    shared_promise<> pr;
    auto f = pr.get_shared_future();
    pr.set_value();
    BOOST_REQUIRE(pr.available());
    BOOST_REQUIRE_NO_THROW(f.get());
}
```

while, as @piodul noticed, adding a `yield()` after `set_value` allows the state to be set properly:
```c++
SEASTAR_THREAD_TEST_CASE(test_shared_promise_with_outstanding_future_is_immediately_available) {
    shared_promise<> pr;
    auto f = pr.get_shared_future();
    pr.set_value();
    yield().get();
    BOOST_REQUIRE(pr.available());
    BOOST_REQUIRE_NO_THROW(f.get());
}
```

See @piodul's comment https://github.com/scylladb/scylladb/issues/15956#issuecomment-1795332929:
> The only missing detail is that `shared_state::resolve` is called *in a continuation* which is attached to `_original_future`. The `_original_future` waits on a seastar promise inside `shared_promise`, and `shared_promise::set_value` sets that promise. However, setting a seastar promise does not cause the continuation attached to the future to run immediately, it is only `seastar::schedule`d to run later, at the end of the queue of the current scheduling group. Therefore, `available()` does not become true immediately after `set_value`.
>
> If, after the task which calls `set_value`, another task which calls `available()` is already scheduled, then the other task will also see the shared future as unavailable.

bhalevy:
Reopened by 584d2cd80af3f5ddbc8b04b847ff183b3c7d33d1

Issue #1917, opened 2023-11-02
Use-after-free with list_directory() that co_yields entries


bhalevy:
Seen in https://app.circleci.com/pipelines/github/bhalevy/seastar/169/workflows/a90389f6-be44-404a-a306-66185082077b/jobs/1496
```
./run ./configure.py --compiler clang++-15 --c++-standard 20
...
./run ./test.py --mode=debug
...
16/74 Test #16: Seastar.unit.directory ........................***Failed    0.86 sec
[0/1] cd /home/circleci/project/build/debug/tests/unit && /usr/bin/cmake -E env ASAN_OPTIONS=disable_coredump=0:abort_on_error=1:detect_stack_use_after_return=1:verify_asan_link_order=0 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1 BOOST_TEST_CATCH_SYSTEM_ERRORS=no /home/circleci/project/build/debug/tests/unit/directory_test -c 2
WARNING: debug mode. Not for benchmarking or production
INFO  2023-11-02 18:06:53,528 seastar - Reactor backend: io_uring
INFO  2023-11-02 18:06:53,529 [shard 0:n/a ] seastar - Perf-based stall detector creation failed (EACCESS), try setting /proc/sys/kernel/perf_event_paranoid to 1 or less to enable kernel backtraces: falling back to posix timer.
--- Regular lister test ---
io_queue_test (type=regular)
mtls_server.key (type=regular)
https-server.py (type=regular)
closeable_test (type=regular)
other.crt (type=regular)
alien_test (type=regular)
signal_test (type=regular)
lowres_clock_test (type=regular)
ipv6_test (type=regular)
CMakeFiles (type=directory)
simple_stream_test (type=regular)
test.key (type=regular)
weak_ptr_test (type=regular)
file_utils_test (type=regular)
fsnotifier_test (type=regular)
allocator_test (type=regular)
caother.pem (type=regular)
thread_test (type=regular)
request_parser_test (type=regular)
mtls_ca.crt (type=regular)
other.csr (type=regular)
source_location_test (type=regular)
CTestTestfile.cmake (type=regular)
file_io_test (type=regular)
socket_test (type=regular)
shared_token_bucket_test (type=regular)
chunk_parsers_test (type=regular)
mtls_client1.key (type=regular)
mtls_ca.key (type=regular)
json_formatter_test (type=regular)
pipe_test (type=regular)
tls-ca-bundle.pem (type=regular)
[...]

PR #1928, opened 2023-11-07
shared_future: make available() immediate after set_value()

piodul:
A shared future takes a regular future on construction and resolves after the original future resolves. This is implemented by attaching a `then_wrapped` continuation on the original future which, when run, forwards the result of the original future to the currently registered waiters.

The original future is kept in the `_original_future` field. The continuation is attached on the first `shared_future::get_future()` call - which causes the `_original_future` to become invalid - and it is reassigned again at the end of the continuation.

This implementation detail has an effect on `shared_future::available()`, which is implemented via `_original_future.available()`. When the promise associated with the original future is set, the continuation is scheduled for execution but does not run immediately. Because of that, `_original_future` is not updated immediately after `promise::set_value` and `shared_future::available()` keeps returning false until the continuation runs.

This is especially confusing for `shared_promise`, which are essentially a `promise` + `shared_future` pair and have both `set_value` and `available` methods: setting the shared promise does not cause it to become available.

Fix this by getting rid of the `_original_future` field and by changing the internal `shared_state` to derive from `continuation_base`. By being a continuation, the shared state now always has access to the current `future_state` (no reassigning is necessary), so the `available()` method now works properly.

Fixes: scylladb/seastar#1926

### DIFF
--- a/include/seastar/core/shared_future.hh
+++ b/include/seastar/core/shared_future.hh
@@ -53,6 +53,7 @@ struct future_option_traits<with_clock<Clock>, T> {
 
     using future_type = future<T>;
     using promise_type = promise<T>;
+    using continuation_base_type = continuation_base<T>;
 };
 
 template <typename Clock>
@@ -61,6 +62,7 @@ struct future_option_traits<with_clock<Clock>> {
 
     using future_type = future<>;
     using promise_type = promise<>;
+    using continuation_base_type = continuation_base<>;
 };
 
 template <typename T>
@@ -109,6 +111,7 @@ template<typename... T>
 class shared_future {
     template <typename... U> friend class shared_promise;
     using options = future_option_traits<T...>;
+    using continuation_base_type = typename options::continuation_base_type;
 public:
     using clock = typename options::clock_type;
     using time_point = typename clock::time_point;
@@ -117,8 +120,9 @@ public:
     using value_tuple_type = typename future_type::tuple_type;
 private:
     /// \cond internal
-    class shared_state : public enable_lw_shared_from_this<shared_state> {
-        future_type _original_future;
+    class shared_state final : public enable_lw_shared_from_this<shared_state>, public continuation_base_type {
+        // Ensures that shared_state is alive until run_and_dispose() runs.
+        lw_shared_ptr<shared_state> _keepaliver;
         struct entry {
             promise_type pr;
             std::optional<abort_on_expiry<clock>> timer;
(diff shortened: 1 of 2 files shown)


piodul:
Looks like the tests failed on https://github.com/scylladb/seastar/issues/1917

include/seastar/core/shared_future.hh
```diff
@@ -141,41 +145,47 @@ private:
...
                     try {
-                        p.set_value(state.get_value());
+                        p.set_value(this->_state.get_value());
                     } catch (...) {
                         p.set_exception(std::current_exception());
                     }
                     _peers.pop_front();
                 }
             }
+            _keepaliver.release();
```
avikivity, line 177: This is nice

include/seastar/core/shared_future.hh
```diff
@@ -184,53 +194,41 @@ private:
...
-                return future_type(exception_future_marker(), std::exception_ptr(_original_future._state.get_exception()));
+            } else if (this->_state.failed()) {
+                return future_type(exception_future_marker(), std::exception_ptr(this->_state.get_exception()));
             } else {
-                return future_type(ready_future_marker(), _original_future._state.get_value());
+                return future_type(ready_future_marker(), this->_state.get_value());
             }
         }
 
         future_type get_future(abort_source& as) noexcept {
```
avikivity, line 205: and this is why it's a mistake to push abort_source everywhere @kbr-scylla

kbr-scylla: Please elaborate. You mean the code duplication?

include/seastar/core/shared_future.hh
tchaikov: what if the `shared_state` is never scheduled? it would lead to leak of it.

tchaikov: #1937 was created to track this regression.

piodul: Why wouldn't it be scheduled? It is scheduled here, only if the result of the future is not available yet:

```c++
        explicit shared_state(future_type f) noexcept
                : _keepaliver(f.available() ? nullptr : this->shared_from_this()) {
            if (f.available()) {
                this->set_state(f.get_available_state_ref());
            } else {
                internal::set_callback(std::move(f), this);
            }
        }
```

How is it different from e.g. a continuation created by `.then()`?

tchaikov: @piodul hi Piotr, the "not-scheduled" case is but a hypothesis. what i really want to address is #1937. maybe, just maybe, the corresponding promise is never fulfilled?

PR #1929, opened 2023-11-08
build: use different names for .pc files for each build mode

tchaikov:
before this change, we always generate a single seasetar.pc for seastar, this works fine for the CMake generators which only generate a single configuration. but when it comes to Multi-Config generators, which generate multiple configurations for each available build configuration, this model does not work anymore. as, for instance, the cflags in .pc file are different in Debug build and in Release build.

so, in order to cater the needs to support Multi-Config generators, we use different library names for each of the build configurations, for instance, use seastar_Release.pc for the release build of Seastar. and keep using the original path name of seastar.pc if the non-multi-config generator is used, to preserve the compatibility of the existing seastar.pc.

since the typical use case of Multi-Config generator of Seastar is currently for building parent projects with also use Multi-Config generator in the build-only environment, but not for installing Seastar, so we don't provide a wrapper .pc file which defaults to, for instance, the release build of .pc file.


denesb (approved):

tchaikov:

PR #1935, opened 2023-11-10
defer, closeable: do not use [[nodiscard(str)]]

### LINKED ISSUE #1932: defer.hh:46:51: error: wrong number of arguments specified for ‘nodiscard’ attribute compile error
help me

tchaikov:
`[[nodiscard]]` does not accept an argument until C++20, but we still support C++17, so let's avoid using the C++20 feature before dropping the support of C++17.

Fixes #1932


tchaikov:
#1936 was created to bring the string parameters back in a post-C++17 era.

PR #1939, opened 2023-11-11
coroutine: generator: resume caller coroutine only by next_awaiter 

tchaikov:
this series addresses two issues in the buffered generator.

1. resume the caller coroutine only on the getter side
2. reset `_wait_for_next_value` before `set_value()`

Fixes https://github.com/scylladb/seastar/issues/1913

### DIFF
--- a/include/seastar/coroutine/generator.hh
+++ b/include/seastar/coroutine/generator.hh
@@ -87,8 +87,7 @@ public:
         assert(_generator);
         _generator->put_next_value(std::forward<U>(value));
         assert(_wait_for_next_value);
-        _wait_for_next_value->set_value();
-        _wait_for_next_value = {};
+        std::exchange(_wait_for_next_value, std::nullopt)->set_value();
         return {};
     }
 
@@ -166,9 +165,8 @@ public:
     generator_buffered_promise(const generator_buffered_promise&) = delete;
     ~generator_buffered_promise() = default;
     void return_void() noexcept {
-        if (_wait_for_next_value) {
-            _wait_for_next_value->set_value();
-            _wait_for_next_value = {};
+        if (auto wait_for_next_value = std::exchange(_wait_for_next_value, std::nullopt)) {
+            wait_for_next_value->set_value();
         }
     }
     void unhandled_exception() noexcept;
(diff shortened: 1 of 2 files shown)


include/seastar/coroutine/generator.hh
```diff
@@ -131,17 +131,20 @@ class generator_buffered_promise;
...
+    yield_awaiter(bool wait_for_free_space) noexcept
+        : _wait_for_free_space{wait_for_free_space} {}
 
     bool await_ready() noexcept {
-        return _future.available();
+        return _wait_for_free_space;
+    }
+
+    coroutine_handle<> await_suspend(coroutine_handle<promise_type>) noexcept {
+        return std::noop_coroutine();
```
avikivity, line 144: I don't understand this. Why not return `true`?  Why resume some do-nothing coroutine?

avikivity: (or void)

include/seastar/coroutine/generator.hh
```diff
@@ -174,18 +176,11 @@ public:
...
             _wait_for_next_value = {};
-            return {make_ready_future()};
-        }
-        if (ready) {
-            return {make_ready_future()};
-        } else {
-            assert(!_wait_for_free_space);
-            return {_wait_for_free_space.emplace().get_future()};
         }
+        return {ready};
```
avikivity, line 180: Might as well delete yield_awaiter and return ready directly, no? The new yield_awaiter does nothing.

avikivity:
Please add a test that exposes the problem. The fact that the directory lister exposed it means the regular tests are insufficient.

include/seastar/coroutine/generator.hh
tchaikov: i don't want to return to the caller coroutine.

include/seastar/coroutine/generator.hh
tchaikov: again, i don't want to return to the caller coroutine. we can define `yield_awaiter` right in `generator_buffered_promise` if you prefer this way.

tchaikov:
somehow, the existing tests pass. but the new tests added to verify the fix are now failing. looking.

PR #1941, opened 2023-11-14
build: do not define SEASTAR_TYPE_ERASE_MORE on all builds

tchaikov:
this change addresses the regression introduced by 3536f55d.

in 3536f55d, we should have defined SEASTAR_TYPE_ERASE_MORE only for the enabled build modes. but in that commit, this macro was defined for all builds modes. this introduces a regression which erases the coroutine type for release build. this pratically prevents us from checking a coroutine task by looking at its symbol name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1043,8 +1043,7 @@ endif ()
 
 target_compile_definitions (seastar
   PUBLIC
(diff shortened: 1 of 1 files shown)


tchaikov:
@denesb hi Botond, could you help review this change. sorry for introducing this regression.

denesb (approved):

tchaikov:
@scylladb/seastar-maint hello maintainers could you help merge this change?

PR #1943, opened 2023-11-15
build: use tri_state_option() to link against Sanitizers

tchaikov:
this change is a leftover of fc0189f9, where we extracted tri_state_option() out. but it turns out Seastar_SANITIZE also falls into this category.

so, in this change, we use `tri_state_option()` to link against Sanitizer libraries. this allows us to set the compiling options for all available configurations configured by multi-config generator.


denesb (approved):

tchaikov:

Issue #1785, opened 2023-08-11
epoll backend: Aborted (EPOLLERR|EPOLLHUP) connections cause 100% reactor util until the respective socket is closed


StephanDollberg:
Hi,

using the epoll backend connections that error out and are not closed in timely fashion can cause 100% reactor util due to spinning on `reactor_backend_epoll::wait_and_process`. Examples are RSTed server connections or client connections that failed to connect.

**How to reproduce:**

We originally saw this on a Redpanda cluster that showed 100% reactor utilization on shards even though no load was applied at the same time we were busy polling:

![image](https://github.com/scylladb/seastar/assets/2904723/b5c7a7c6-d745-47f9-8c07-b418e413eca7)

I have since then reproduced it with either the uploaded code from https://github.com/scylladb/seastar/issues/762 and also with the following small modification to the sctp clients demo apps (all using latest seastar):

``` cpp
diff --git a/demos/tcp_sctp_client_demo.cc b/demos/tcp_sctp_client_demo.cc
index 3de72776..2ab8b5e1 100644
--- a/demos/tcp_sctp_client_demo.cc
+++ b/demos/tcp_sctp_client_demo.cc
@@ -43,7 +43,7 @@ transport protocol = transport::TCP;

 class client {
 private:
-    static constexpr unsigned _pings_per_connection = 10000;
+    static constexpr unsigned _pings_per_connection = 160000; // gives more time to kill the client
     unsigned _total_pings;
     unsigned _concurrent_connections;
     ipv4_addr _server_addr;
diff --git a/demos/tcp_sctp_server_demo.cc b/demos/tcp_sctp_server_demo.cc
index 6918ab2e..9d942ee9 100644
--- a/demos/tcp_sctp_server_demo.cc
+++ b/demos/tcp_sctp_server_demo.cc
@@ -19,6 +19,7 @@
  * Copyright 2014 Cloudius Systems
  */

+#include "seastar/core/sleep.hh"
 #include <seastar/core/reactor.hh>
 #include <seastar/core/app-template.hh>
 #include <seastar/core/temporary_buffer.hh>
@@ -77,13 +78,16 @@ class tcp_server {
             connected_socket fd = std::move(ar.connection);
             socket_address addr = std::move(ar.remote_address);
             auto conn = new connection(*this, std::move(fd), addr);
-            (void)conn->process().then_wrapped([conn] (auto&& f) {
+            (void)conn->process().then_wrapped([conn] (auto&& f) -> future<void> {
[...]

PR #1945, opened 2023-11-15
epoll: Avoid spinning on aborted connections

StephanDollberg:
Using the epoll backend and having a non-closed socket that got terminated
with either EPOLLHUP|EPOLLERR under the hood but isn't closed yet can lead to
busy spinning in the reactor at 100%.

It's caused by the following scenario:

 - epoll_pwait returns an event for a socket with EPOLLIN|EPOLLHUP (or
   EPOLLOUT in a different respective case)
 - We enter the error branch and set evt.events. This doesn't make much
   difference at this point as pfd->events_requested is still equal to
   EPOLLIN
 - events_to_remove is calculated to 0
 - complete_epoll_event is called setting pfd->events_requested to 0
 - The loop concludes without removing the socket from the epoll set as
   we calculate events_to_remove before changing it. I think this is
   intended to avoid unneeded syscalls in the good case (so usually only
   receiving an event when not waiting for it will remove the socket from
   the set)
 - Given we run in level triggered mode epoll_wait again returns with
   EPOLLIN|EPOLLHUP for our socket
 - This time around in the error branch we set evt.events to 0
 - This causes events_to_remove again to be calculated to 0 even though
   pfd->events_requested is 0 as well
 - Above continues until the socket is eventually closed and hence
   implicitly removed from the epoll set

This patch prevents this from happening by pro-actively removing the fd
from the epoll set once we have received an error and are no longer
waiting for any events on the socket.

Note this should be relatively safe in any case. In case we ever get
interested in events again we would just add it back.

Fixes https://github.com/scylladb/seastar/issues/1785

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -857,13 +857,14 @@ reactor_backend_epoll::wait_and_process(int timeout, const sigset_t* active_sigm
             _steady_clock_timer_deadline = {};
             continue;
         }
-        if (evt.events & (EPOLLHUP | EPOLLERR)) {
+        bool has_error = evt.events & (EPOLLHUP | EPOLLERR);
+        if (has_error) {
             // treat the events as required events when error occurs, let
             // send/recv/accept/connect handle the specific error.
             evt.events = pfd->events_requested;
         }
         auto events = evt.events & (EPOLLIN | EPOLLOUT | EPOLLRDHUP);
-        auto events_to_remove = events & ~pfd->events_requested;
+        auto events_to_remove = has_error ? pfd->events_requested : events & ~pfd->events_requested;
         complete_epoll_event(*pfd, events, EPOLLRDHUP);
         if (pfd->events_rw) {
             // accept() signals normal completions via EPOLLIN, but errors (due to shutdown())


avikivity:
@michoecho please review

StephanDollberg:
Test failure seems to be https://github.com/scylladb/seastar/issues/1913

avikivity:
> Test failure seems to be #1913

Yes we can ignore it, there's a pending fix.

src/core/reactor_backend.cc
```diff
@@ -882,6 +884,18 @@ reactor_backend_epoll::wait_and_process(int timeout, const sigset_t* active_sigm
...
+            // In case we got an error and we are not interested in any events
+            // anymore, remove the fd from the epoll set. Otherwise due to the
+            // above logic of setting evt.events to pfd->events_requested (0)
+            // and subsequent calculation of events_to_remove, we would never
+            // remove the fd from the epoll set. Due to running in level
+            // triggered mode this would then cause repeated wake ups and
+            // spinning until the fd is closed and thereby removed from the
+            // epoll set.
+            pfd->events_epoll = 0;
+            ::epoll_ctl(_epollfd.get(), EPOLL_CTL_DEL, pfd->fd.get(), nullptr);
```
avikivity, line 897: This affects both the read end and the write end. What if the write end is still open?

Perhaps we should add EPOLLIN to events_to_remove and let the if (events_to_remove) branch choose between EPOLL_CTL_MOD and EPOLL_CTL_ADD.

src/core/reactor_backend.cc
```diff
@@ -882,6 +884,18 @@ reactor_backend_epoll::wait_and_process(int timeout, const sigset_t* active_sigm
             auto op = evt.events ? EPOLL_CTL_MOD : EPOLL_CTL_DEL;
             ::epoll_ctl(_epollfd.get(), op, pfd->fd.get(), &evt);
         }
+        else if (has_error && pfd->events_requested == 0) {
```
avikivity, line 887: Also, the coding style places the else on the same line as the closing bracket.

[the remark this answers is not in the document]
StephanDollberg: Yes you are right it should not be possible. I added it as a kind of a safety check to make sure we are not removing the fd from the epoll set in case we are still interested in anything. 

Given this bit of code is already a bit complicated with the different bit manipulations and their order mattering I wanted to make it resistent to any codes change that could potentially break this assumption.

src/core/reactor_backend.cc
StephanDollberg: > This affects both the read end and the write end. What if the write end is still open?

By write end do you actually mean our write site i.e.: a possible registration of EPOLLOUT or the remote end?

Re the former, as currently implemented we would only take this path if we are not interested in any events anymore because of the `pfd->events_requested == 0` check in the if.

However, as per the other comment given how the whole handling is currently implemented the moment we receive an error we will remove everything from `pfd->events_requested` anyway and hence would ignore any further events in any case.

Re the later, I don't think that should matter in that regard. As stated above if we received any further events then we would not get notified of them anymore. However given `events_requested` is 0 now even if we got notified we would not deliver them anywhere.

Want to state that this can happen on both EPOLLIN and EPOLLOUT but that doesn't really change your point or what's written above (just that any handling specific to either of the two would have to handle both cases).

avikivity: > > This affects both the read end and the write end. What if the write end is still open?
> 
> By write end do you actually mean our write site i.e.: a possible registration of EPOLLOUT or the remote end?


A possible registration of EPOLLOUT.

> 
> Re the former, as currently implemented we would only take this path if we are not interested in any events anymore because of the `pfd->events_requested == 0` check in the if.

Having re-read the code I agree, but from a higher-level point of view: the code prefers to generate a spurious wakeup (complete_epoll_event()). If the wakeup was wanted, great. If not wanted, the waiter will re-register.


> 
> However, as per the other comment given how the whole handling is currently implemented the moment we receive an error we will remove everything from `pfd->events_requested` anyway and hence would ignore any further events in any case.
> 
> Re the later, I don't think that should matter in that regard. As stated above if we received any further events then we would not get notified of them anymore. However given `events_requested` is 0 now even if we got notified we would not deliver them anywhere.
> 
> Want to state that this can happen on both EPOLLIN and EPOLLOUT but that doesn't really change your point or what's written above (just that any handling specific to either of the two would have to handle both cases).

Yes. What about reducing complexity by using events_to_remove to de-register? Right now both has_error and events_to_remove do the same thing.

StephanDollberg: > Yes. What about reducing complexity by using events_to_remove to de-register? Right now both has_error and events_to_remove do the same thing.

I have pushed a change which uses events_to_remove. 

I am not sure it removes complexity but I'll let you guys be the judge. As touched on in the other comment I originally tried to basically leave the existing code as is and just add additional handling so as not to infer with the already existing complexity of how events and events_to_remove is calculated.

However this change should work equally well. If you prefer the new version I will squash the commits.

cc @michoecho

avikivity: I think it's simpler. Could be simplified a bit further by calculating events_to_remove early, then overriding it the error branch, but we can leave it for later. I prefer not having a maze of conditionals.

Please squash the commits.

StephanDollberg: Commits are squashed btw

michoecho:
@avikivity Ping.

michoecho:
> @avikivity Ping.

@avikivity Ping.

PR #1946, opened 2023-11-16
memory: do not use variable length array

tchaikov:
vla (variable length array) is an extension in GCC and Clang. and it is not part of the C++ standard.

so let's avoid using it if possible, for better standard compliant and to silence the warning from Clang 18:

```
/home/kefu/.local/bin/clang++ -DBOOST_NO_CXX98_FUNCTION_BASE -DFMT_DEPRECATED_OSTREAM -DFMT_SHARED -DSCYLLA_BUILD_MODE=release -DSEASTAR_API_LEVEL=7 -DSEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_NUMA -DSEASTAR_HAVE_SYSTEMTAP_SDT -DSEASTAR_HAVE_URING -DSEASTAR_LOGGER_TYPE_STDOUT -DSEASTAR_PTHREAD_ATTR_SETAFFINITY_NP -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_SSTRING -DSEASTAR_STRERROR_R_CHAR_P -DCMAKE_INTDIR=\"Release\" -I/home/kefu/dev/scylladb/seastar/include -I/home/kefu/dev/scylladb/build/seastar/gen/include -I/home/kefu/dev/scylladb/seastar/src -ffunction-sections -fdata-sections -O3 -g -gz -std=gnu++20 -fvisibility=hidden -Wall -Werror -Wextra -Wno-error=deprecated-declarations -Wimplicit-fallthrough -Wno-c++11-narrowing -Wno-mismatched-tags -Wno-overloaded-virtual -Wno-unsupported-friend -Wno-unused-parameter -Wno-missing-field-initializers -Wno-deprecated-copy -Wno-ignored-qualifiers -ffile-prefix-map=/home/kefu/dev/scylladb=. -march=westmere -mllvm -inline-threshold=2500 -fno-slp-vectorize -U_FORTIFY_SOURCE -Werror=unused-result "-Wno-error=#warnings" -UNDEBUG -Wdeprecated -Wno-error=deprecated -fvisibility=hidden -gz -MD -MT seastar/CMakeFiles/seastar.dir/Release/src/core/memory.cc.o -MF seastar/CMakeFiles/seastar.dir/Release/src/core/memory.cc.o.d -o seastar/CMakeFiles/seastar.dir/Release/src/core/memory.cc.o -c /home/kefu/dev/scylladb/seastar/src/core/memory.cc
/home/kefu/dev/scylladb/seastar/src/core/memory.cc:1152:54: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]
 1152 |     auto new_page_array_pages = align_up(sizeof(page[new_pages + 1]), page_size) / page_size;
      |                                                      ^~~~~~~~~~~~~
```

### DIFF
--- a/src/core/memory.cc
+++ b/src/core/memory.cc
@@ -1149,7 +1149,7 @@ void cpu_pages::do_resize(size_t new_size, allocate_system_memory_fn alloc_sys_m
     mem.release();
     maybe_enable_transparent_hugepages(mmap_start, mmap_size);
     // one past last page structure is a sentinel
-    auto new_page_array_pages = align_up(sizeof(page[new_pages + 1]), page_size) / page_size;
+    auto new_page_array_pages = align_up(sizeof(page) * (new_pages + 1), page_size) / page_size;
     auto new_page_array
         = reinterpret_cast<page*>(allocate_large(new_page_array_pages));
     if (!new_page_array) {
@@ -1160,7 +1160,7 @@ void cpu_pages::do_resize(size_t new_size, allocate_system_memory_fn alloc_sys_m
     new_page_array[new_pages].free = false;
     auto old_pages = reinterpret_cast<char*>(pages);
     auto old_nr_pages = nr_pages;
-    auto old_pages_size = align_up(sizeof(page[nr_pages + 1]), page_size);
+    auto old_pages_size = align_up(sizeof(page) * (nr_pages + 1), page_size);
     old_pages_size = size_t(1) << log2ceil(old_pages_size);
     pages = new_page_array;
     nr_pages = new_pages;
--- a/tests/perf/fstream_perf.cc
+++ b/tests/perf/fstream_perf.cc
@@ -26,6 +26,7 @@
 #include <seastar/core/do_with.hh>
 #include <seastar/core/loop.hh>
 #include <fmt/printf.h>
+#include <string>
 
 using namespace seastar;
 using namespace std::chrono_literals;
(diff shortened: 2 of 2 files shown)


mykaul:
Apart from not being part of the standard, what's the downside of using it? I feel it has value.

tchaikov:
> Apart from not being part of the standard, what's the downside of using it? I feel it has value.

there are lots of discussion around this topic. my put is that it's dangerous. just for the fun of reading: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3810.pdf

nyh:
Although there is long arguments for and against VLAs (see for example https://stackoverflow.com/questions/1887097/why-arent-variable-length-arrays-part-of-the-c-standard), but the fact of that matter is that it has been in standard C (not C++!) for a quarter century now (since C99), and in gcc (and then clang) for that long as well - in both C and C++ modes. So I don't see any urgent (or non urgent) reason to stop using them.

That being said, if we only use them in one place, and it's easy to get rid of it without tradeoffs, then I'm not against.

src/core/memory.cc
```diff
@@ -1149,7 +1149,7 @@ void cpu_pages::do_resize(size_t new_size, allocate_system_memory_fn alloc_sys_m
     mem.release();
     maybe_enable_transparent_hugepages(mmap_start, mmap_size);
     // one past last page structure is a sentinel
-    auto new_page_array_pages = align_up(sizeof(page[new_pages + 1]), page_size) / page_size;
+    auto new_page_array_pages = align_up(sizeof(page) * (new_pages + 1), page_size) / page_size;
```
nyh, line 1152: Good fix. This wasn't even a VLA - no VLA was ever created here. It was just a weird way to do multiplication - it's like doing `sizeof(int[n])` instead of doing `sizeof(int)*n`. Good riddance.

nyh (approved):

nyh:
Your patch looks good, and I would merge it, but can you please look into the CI failures? Thanks.

avikivity:
Those are known failures for which we're preparing a fix.

Issue #1947, opened 2023-11-16
reactor fails to run when submitting request to NFS


tchaikov:
when running a seastar application and it sends io request to files located on a filesystem which does not support `RWF_NOWAIT`, `io_submit()` fails with `ENOTSUPP`:

```
ERROR 2023-11-11 01:16:53,219 [shard 0] seastar - io_submit: Operation not supported `
exception running reactor main loop on shard 0.
Backtrace:
  0x4ff41f8
  0x5006883
  0x50050b8
  0x4fabaf5
  0x4faac68
  0x1148d34
  0x114a840
  0x114746a
  /opt/scylladb/libreloc/libc.so.6+0x2750f
  /opt/scylladb/libreloc/libc.so.6+0x275c8
  0x1145624
2023-11-11 01:16:53,315 INFO exited: scylla (exit status 1; not expected)
```

it turns out quite a few (if not all) networking filesystem do not support nowait aio, so in addition to nfs, this problem also applies to 9p, cephfs, cifs and ocfs2.

we could print a more informative error message instead of 

> `io_submit: Operation not supported`

and a backtrace

hiqsociety:
1. possible to make seastar enable this?
2. what are the networking filesystems currently supported by seastar?

tchaikov:
> 1. possible to make seastar enable this?

as Nadav put in another thread. it might be possible. but this is not our priority at this moment.

> 2. what are the networking filesystems currently supported by seastar?

as the description of this issue put, AFAICT, no known network filesystem or stacked filesystem are supported. at least the ones in linux 6.3 are not supported.

hiqsociety:
@tchaikov allow me to thank you for this reply. 
sorry for the late response, can i please ask for the version of seastar support that works on nfs and with least amount of bug or caveats to look out for?

i'm using scylladb and if i cant use nfs, it's a no-go for my deployment.

can you provide the scylladb version (with least bug) for production use using nfs as storage backend? i understand the settings needing intr options for nfs 4.1 etc. or maybe you can just provide your expert opinion on how to get this done.

basically i'm prepared to use legacy version of scylladb just for the NFS backend and definitely hope to get your input on what to look out for for production use.


p.s. : i helped operate the english telegram group of scylladb and i'm quite sure there are a few people who would like to know about this too.

appreciate this prompt reply. thx!

hiqsociety:
@tchaikov any feedback on this? what legacy version of scylladb will work with nfs for production use?

[...]

PR #1949, opened 2023-11-16
reactor: print more informative error when io_submit fails

tchaikov:
when io_submit() fails with errno of ENOTSUP, there are good chances that the application is sending i/o requests to a network filesystem or a stacked filesystem. instead of printing

> seastar - io_submit: Operation not supported

let's print

> seastar - could not submit io. this happens when accessing filesystem not support asynchronous direct I/O

Refs #1947

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -179,6 +179,18 @@ aio_storage_context::handle_aio_error(linux_abi::iocb* iocb, int ec) {
             // we will only remove it from _pending_io and try again.
             return 1;
         }
+        case EINVAL:
+            // happens when the filesystem does not implement aio read or write
+            [[fallthrough]];
+        case ENOTSUP: {
+            seastar_logger.error("io_submit failed: this happens when "
+                                 "accessing filesystem which does not supports "
+                                 "asynchronous direct I/O");
+            auto desc = get_user_data<kernel_completion>(*iocb);
+            _iocb_pool.put_one(iocb);
+            desc->complete_with(-ENOTSUP);
+            return 1;
+        }
         default:
             ++_r._io_stats.aio_errors;
             throw std::system_error(ec, std::system_category(), "io_submit");


src/core/reactor_backend.cc
```diff
@@ -182,6 +182,11 @@ aio_storage_context::handle_aio_error(linux_abi::iocb* iocb, int ec) {
             // we will only remove it from _pending_io and try again.
             return 1;
         }
+        case ENOTSUP:
+            throw_system_error_on(true,
+                                  "could not submit io. this happens when accessing "
+                                  "filesystem not support asynchronous direct I/O");
+            abort();
```
avikivity, line 189: why this abort?

avikivity: also, why throw? it can't be handled

tchaikov: @avikivity i was following the approach of the `default` branch.  as i thought we should just abort the application in this case, as we might not a good way to handle this issue.

but after a second thought, it might be desirable to return this error to the caller as an `std::system_error`.

tchaikov:

src/core/reactor_backend.cc
```diff
@@ -182,6 +182,15 @@ aio_storage_context::handle_aio_error(linux_abi::iocb* iocb, int ec) {
             // we will only remove it from _pending_io and try again.
             return 1;
         }
+        case ENOTSUP: {
```
xemul, line 185: I guess `case EOPNOTSUPP` should also come here, I found this in ext4 code:

```
static ssize_t ext4_buffered_write_iter(struct kiocb *iocb,
                    struct iov_iter *from)
{
    ssize_t ret;
    struct inode *inode = file_inode(iocb->ki_filp);

    if (iocb->ki_flags & IOCB_NOWAIT)
        return -EOPNOTSUPP;
```

it's buffered, but ext4 sometimes falls back to it even if O_DIRECT is requested

xemul: Also check aio_read() code:

```
    if (unlikely(!file->f_op->read_iter))
        return -EINVAL;
```

some filesystems may bail out early with EINVAL for any AIO

tchaikov: > Also check aio_read() code:
> 
> ```
>     if (unlikely(!file->f_op->read_iter))
>         return -EINVAL;
> ```
> 
> some filesystems may bail out early with EINVAL for any AIO

yeah, if the aio read or write op is not implemented at all.

tchaikov:
v2:

- handle `EINVAL` as well.

tchaikov:
@xemul hi Pavel, thanks for your insights. could you take another look?

PR #1952, opened 2023-11-17
build: do not return a non-false condition if the option is off 

tchaikov:
this series addresses two bugs in tri_state_option() function. these bugs break the dev build of scylla by enabling the santizer in the dev build.


tchaikov:
@denesb hi Botond, could you please help review this change as well?

tchaikov:
thank you for your review and approval, Botond!

the multi-config generator thing is also new to me. so i am in a process of learning by doing. =)

tchaikov:
@scylladb/seastar-maint hello maintainers, could you help merge this change? it addresses a regression in the building system.

PR #1954, opened 2023-11-18
Fix `fmt` lib version check

jbajic:
Fix `fmt` lib version check.

PR #1955, opened 2023-11-22
sstring_test: include fmt/std.h only if fmtlib >= 10.0.0

tisonkun:
Fixup https://github.com/scylladb/seastar/pull/1897 where we miss one zero.

See https://github.com/fmtlib/fmt/blob/c13753a70cc55f3b1c99fb8f8395e78e5f9cae43/include/fmt/core.h#L20-L21


tisonkun:
Duplicate of https://github.com/scylladb/seastar/pull/1954

PR #1956, opened 2023-11-22
shared_ptr: add fmt::formatter for shared_ptr types

tchaikov:
fmtlib v10.0.0 stopped defining the formatter for the types with operator<<() even with `FMT_DEPRECATED_OSTREAM` defined. so let's define the formatters for shared_ptr types.


tchaikov:
@avikivity instead adding them in `seastarx.hh`, i think the better home for them is seastar.

include/seastar/core/shared_ptr.hh
```diff
@@ -926,20 +927,40 @@ struct hash<seastar::shared_ptr<T>> : private hash<T*> {
...
 }
 
-SEASTAR_MODULE_EXPORT
 template<typename T>
 const void* ptr(const seastar::shared_ptr<T>& p) {
     return p.get();
 }
 
+template <typename T>
+struct formatter<seastar::shared_ptr<T>> {
```
xemul, line 944: The 6786ad732c5c299e31879fa8074f2c18b95b460d says:

> so we are able to format, for instance, seastar::shared_ptr<T>,  with
>     
>    seastar::shared_ptr<int> p;
>    auto s = fmt::format("{}", fmt::ptr(p));

Is `std::unique_ptr<T>` default-formattable? No, it needs to be wrapped with fmt::ptr and in this case it prints pointer value, not what it points to. I think it should be the case for seastar shared pointers too. All the more so there's lazy-deref thing that's printable

tchaikov: > The [6786ad7](https://github.com/scylladb/seastar/commit/6786ad732c5c299e31879fa8074f2c18b95b460d) says:
> 
> > so we are able to format, for instance, seastar::shared_ptr,  with
> > seastar::shared_ptr p;
> > auto s = fmt::format("{}", fmt::ptr(p));
> 
> Is `std::unique_ptr<T>` default-formattable? No, it needs to be wrapped with fmt::ptr and in this case it prints pointer value, not what it points to. 

yes. and this is not quite related to this change, actually. 

> I think it should be the case for seastar shared pointers too. All the more so there's lazy-deref thing that's printable

i am not sure, the goal is to preserve the existing behavior in fmt v10 or without `FMT_DEPRECATED_OSTREAM` defined. the "existing behavior" is we have when `FMT_DEPRECATED_OSTREAM` is defined and to get seastar applications compilable with fmt v10 with minimum surprise. if we want to have the support of `fmt::ptr()` , it's already there as you already noticed. but it does not help with the fmt v10 transition.

xemul: I see. OK, then

PR #1958, opened 2023-11-22
shared_future: make available() immediate after set_value()

piodul:
A shared future takes a regular future on construction and resolves after the original future resolves. This is implemented by attaching a `then_wrapped` continuation on the original future which, when run, forwards the result of the original future to the currently registered waiters.

The original future is kept in the `_original_future` field. The continuation is attached on the first `shared_future::get_future()` call - which causes the `_original_future` to become invalid - and it is reassigned again at the end of the continuation.

This implementation detail has an effect on `shared_future::available()`, which is implemented via `_original_future.available()`. When the promise associated with the original future is set, the continuation is scheduled for execution but does not run immediately. Because of that, `_original_future` is not updated immediately after `promise::set_value` and `shared_future::available()` keeps returning false until the continuation runs.

This is especially confusing for `shared_promise`, which are essentially a `promise` + `shared_future` pair and have both `set_value` and `available` methods: setting the shared promise does not cause it to become available.

Fix this by implementing the `task` interface for the shared state. Instead of creating a continuation via `.then_wrapped`, the shared state is attached directly as a task to the future. There is no longer need to move the `_original_future` back and forth so the `available()` method now works correctly.

This is a reworked version of the earlier fix in 9f561da which got reverted. It avoids the memory leak issue by scheduling the task only in the same circumstances as the current implementation. A regression test is included.

Fixes: https://github.com/scylladb/seastar/issues/1926

### DIFF
--- a/include/seastar/core/shared_future.hh
+++ b/include/seastar/core/shared_future.hh
@@ -239,6 +239,11 @@ private:
         bool failed() const noexcept {
             return _original_future.failed();
         }
+
+        // Used only in tests (see shared_future_tester in futures_test.cc)
+        bool has_scheduled_task() const noexcept {
+            return _keepaliver != nullptr;
+        }
     };
     /// \endcond
     lw_shared_ptr<shared_state> _state;
@@ -296,6 +301,10 @@ public:
     bool valid() const noexcept {
         return bool(_state);
     }
+
+    /// \cond internal
+    friend class shared_future_tester;
+    /// \endcond
 };
 
 /// \brief Like \ref promise except that its counterpart is \ref shared_future instead of \ref future
--- a/include/seastar/core/shared_future.hh
+++ b/include/seastar/core/shared_future.hh
@@ -117,8 +117,10 @@ public:
     using value_tuple_type = typename future_type::tuple_type;
 private:
     /// \cond internal
-    class shared_state : public enable_lw_shared_from_this<shared_state> {
+    class shared_state final : public enable_lw_shared_from_this<shared_state>, public task {
         future_type _original_future;
+        // Ensures that shared_state is alive until run_and_dispose() runs (if the task was scheduled)
+        lw_shared_ptr<shared_state> _keepaliver;
         struct entry {
             promise_type pr;
             std::optional<abort_on_expiry<clock>> timer;
@@ -145,9 +147,11 @@ private:
                 _original_future.ignore_ready_future();
             }
         }
-        explicit shared_state(future_type f) noexcept : _original_future(std::move(f)) { }
-        void resolve(future_type&& f) noexcept {
-            _original_future = std::move(f);
+        explicit shared_state(future_type f) noexcept
+                : task(default_scheduling_group()) // SG is set later, when the task is scheduled
+                , _original_future(std::move(f)) {
+        }
+        void run_and_dispose() noexcept override {
             auto& state = _original_future._state;
             if (_original_future.failed()) {
                 while (_peers) {
@@ -165,6 +169,11 @@ private:
                     _peers.pop_front();
                 }
             }
+            _keepaliver.release();
+        }
+
+        task* waiting_task() noexcept override {
+            return nullptr;
         }
 
         future_type get_future(time_point timeout = time_point::max()) noexcept {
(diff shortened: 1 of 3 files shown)


piodul:
I believe the cause of the test failures is the same as in the previous version of the PR. It is unrelated to the current PR and is already tracked by https://github.com/scylladb/seastar/issues/1913

tchaikov:
@piodul hi Piotr, i will review this change early next week.

tchaikov:
i am still reviewing this change..

include/seastar/core/shared_future.hh
```diff
@@ -38,6 +38,9 @@ namespace seastar {
 /// \addtogroup future-module
 /// @{
 
+template <typename... T>
+class shared_future;
```
tchaikov, line 42: this forward declaration is not necessary.

include/seastar/core/shared_future.hh
```diff
@@ -296,6 +310,11 @@ public:
     bool valid() const noexcept {
         return bool(_state);
     }
+
+    /// \cond internal
+    template<typename... U>
+    friend bool internal::has_scheduled_task(const shared_future<U...>& p) noexcept;
+    /// \endcond
```
tchaikov, line 317: i'd suggest put all these tester probes into `futures_test.cc`, and only leave a friend declaration. and instead of put the helper into `internal` namespace, put it into `seastar`, so that we can just put
```c++
friend class shared_future_tester;
```
in `shared_future`. and have
```c++
namespace seastar {
class shared_future_tester {
public:
    template <typename... T>
    static bool has_scheduled_task(const shared_future<T...>& f) noexcept {
        return f._state->has_scheduled_task();
    }
};
}
``` 
in `futures_test.cc`. 

for couple reasons:

1. simpler this way. as it avoids the forward declarations
2. less distraction. so we put less test related stuff in the design under test
3. more obvious that the probe is a test-only facility
4. and ensures that the test-only facility can only be used by test.

tchaikov (commented):
i left a couple comments.

include/seastar/core/shared_future.hh
```diff
@@ -145,9 +147,11 @@ private:
                 _original_future.ignore_ready_future();
             }
         }
-        explicit shared_state(future_type f) noexcept : _original_future(std::move(f)) { }
-        void resolve(future_type&& f) noexcept {
-            _original_future = std::move(f);
+        explicit shared_state(future_type f) noexcept
+                : task(default_scheduling_group()) // SG is set later, when the task is scheduled
```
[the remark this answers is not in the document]
piodul: `current_scheduling_group()` needs to access a thread-local variable, while `default_scheduling_group()` returns a constant. I suppose that the latter should be a little bit faster.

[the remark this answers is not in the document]
tchaikov: thank you. it makes sense to me now.

include/seastar/core/shared_future.hh
piodul: The `shared_future_tester` is a definitely a good idea, I'll do it.

include/seastar/core/shared_future.hh
```diff
@@ -145,9 +147,11 @@ private:
...
+                , _original_future(std::move(f)) {
+        }
+        virtual void run_and_dispose() noexcept override {
```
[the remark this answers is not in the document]
piodul: Done, I also did it for `waiting_task()`

include/seastar/core/shared_future.hh
```diff
@@ -239,6 +239,10 @@ private:
         bool failed() const noexcept {
             return _original_future.failed();
         }
+
```
tchaikov, line 242: could add a comment to explain that this helper is only provided for testing.

tchaikov (approved):
lgtm modulo a nit. it's not a blocker anyway.

include/seastar/core/shared_future.hh
piodul: Done

include/seastar/core/shared_future.hh
```diff
@@ -165,6 +169,11 @@ private:
                     _peers.pop_front();
                 }
             }
+            _keepaliver.release();
+        }
+
+        task* waiting_task() noexcept override {
+            return nullptr;
```
avikivity, line 176: This is inaccurate, is it not? All of _peers are waiting.

But we don't have an interface to expose multiple waiters. Still, we could have exposed one.

PR #1962, opened 2023-11-23
make reactor::_task_queue init before the reactor::_cpu_stall_detector

hurricane1026:
because if kernel.perf_event_paranoid = 2, the seastar_logger will be used, and it need check reactor::_task_queues if is empty. In release, the reactor::_task_queue was not initialized, and it was not empty, but just can get a null unique_ptr, the programm will be crash.

BTW. my environment is clang17.0.4 and fedora 39

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -297,7 +297,6 @@ private:
     uint64_t _global_tasks_processed = 0;
     uint64_t _polls = 0;
     metrics::internal::time_estimated_histogram _stalls_histogram;
-    std::unique_ptr<internal::cpu_stall_detector> _cpu_stall_detector;
 
     unsigned _max_task_backlog = 1000;
     timer_set<timer<>, &timer<>::_link> _timers;
@@ -340,6 +339,7 @@ private:
     };
 
     boost::container::static_vector<std::unique_ptr<task_queue>, max_scheduling_groups()> _task_queues;
+    std::unique_ptr<internal::cpu_stall_detector> _cpu_stall_detector;
     internal::scheduling_group_specific_thread_local_data _scheduling_group_specific_data;
     int64_t _last_vruntime = 0;
     task_queue_list _active_task_queues;
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1034,6 +1034,7 @@ reactor::reactor(std::shared_ptr<smp> smp, alien::instance& alien, unsigned id, 
     , _engine_thread(sched::thread::current())
 #endif
     , _cpu_started(0)
+    , _task_queues()
     , _cpu_stall_detector(internal::make_cpu_stall_detector())
     , _reuseport(posix_reuseport_detect())
     , _thread_pool(std::make_unique<thread_pool>(*this, seastar::format("syscall-{}", id))) {


xemul:
> because if kernel.perf_event_paranoid = 2, the seastar_logger will be used, and it need check reactor::_task_queues if is empty

Please be more specific here -- what's the call sequence that leads to a crash

hurricane1026:
> > because if kernel.perf_event_paranoid = 2, the seastar_logger will be used, and it need check reactor::_task_queues if is empty
> 
> Please be more specific here -- what's the call sequence that leads to a crash

crash detail:
here:
[](https://github.com/scylladb/seastar/blob/f0aefcfcaa1b83115e8d77838ff0043d0088a325/src/core/reactor.cc#L1450C120-L1450C144)
when I debug into, then found:
[](https://github.com/scylladb/seastar/blob/f0aefcfcaa1b83115e8d77838ff0043d0088a325/src/core/reactor.cc#L4924)

this line will crash in release, and works fine in debug.

I try to add a log, then found:
`task_queue's size =  105553117609216`

then check the reactor's constructor func:
```
reactor::reactor(std::shared_ptr<smp> smp, alien::instance& alien, unsigned id, reactor_backend_selector rbs, reactor_config cfg)
: _smp(std::move(smp))
, _alien(alien)
, _cfg(cfg)
, _notify_eventfd(file_desc::eventfd(0, EFD_CLOEXEC))
, _task_quota_timer(file_desc::timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC))
, _id(id)
#ifdef HAVE_OSV
, _timer_thread(
    [&] { timer_thread_func(); }, sched::thread::attr().stack(4096).name("timer_thread").pin(sched::cpu::current()))
, _engine_thread(sched::thread::current())
#endif
, _cpu_started(0)
, _cpu_stall_detector(internal::make_cpu_stall_detector())
, _reuseport(posix_reuseport_detect())
, _thread_pool(std::make_unique<thread_pool>(*this, seastar::format("syscall-{}", id))) {
/*
 * The _backend assignment is here, not on the initialization list as
 * the chosen backend constructor may want to handle signals and thus
 * needs the _signals._signal_handlers map to be initialized.
 */
_backend = rbs.create(*this);
*internal::get_scheduling_group_specific_thread_local_data_ptr() = &_scheduling_group_specific_data;
_task_queues.push_back(std::make_unique<task_queue>(0, "main", "main", 1000));
_task_queues.push_back(std::make_unique<task_queue>(1, "atexit", "exit", 1000));
_at_destroy_tasks = _task_queues.back().get();
set_need_preempt_var(&_preemption_monitor);
seastar::thread_impl::init();
_backend->start_tick();
#ifdef HAVE_OSV
_timer_thread.start();
#else
sigset_t mask;
sigemptyset(&mask);
sigaddset(&mask, internal::cpu_stall_detector::signal_number());
auto r = ::pthread_sigmask(SIG_UNBLOCK, &mask, NULL);
assert(r == 0);
#endif
memory::set_reclaim_hook([this] (std::function<void ()> reclaim_fn) {
    add_high_priority_task(make_task(default_scheduling_group(), [fn = std::move(reclaim_fn)] {
        fn();
    }));
});
}
```
the _cpu_stall_detector is inited before the push_back  actions. so the _task_queue was not *correctly* initialized.

tchaikov:
@hurricane1026 i see now. but i think, instead of rearranging the member variable, we should avoid setting `local_engine` when it is still being initialized. so i propose an alternative fix. see #1974. could you help test and review it?

hurricane1026:
> @hurricane1026 i see now. but i think, instead of rearranging the member variables, we should avoid setting `local_engine` when it is still being initialized. so i propose an alternative fix. see #1974. could you help test and review it?

ok， assign it to me.

Issue #1931, opened 2023-11-08
http client exceptions are not very informative


xemul:
When client gets unexpected reply status, it throws `unexpected_reply_status` error with "Unexpected reply status" message and the status itself. However, the `.what()` just returns the message itslef, and doesn't include status into it

PR #1963, opened 2023-11-23
http/exception: Make unexpected status message more informative

xemul:
The exception in question initializes its base's message with useless "Unexpected reply status" string. This message is what returned from the virtual exception::what() method and when printed in logs given no glue to what the problem was.

Fix it by putting the status' string explanation into base string. When printed, the "unexpected status" would be clear from the exception real data type, and the what() would then reveral the actual problem.

fixes: #1931

### DIFF
--- a/include/seastar/http/exception.hh
+++ b/include/seastar/http/exception.hh
@@ -155,7 +155,7 @@ private:
 class unexpected_status_error : public base_exception {
 public:
     unexpected_status_error(http::reply::status_type st)
-        : base_exception("Unexpected reply status", st)
+        : base_exception(fmt::to_string(st), st)
     {}
 };
 
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -1407,3 +1407,12 @@ SEASTAR_TEST_CASE(test_url_param_encode_decode) {
 
     return make_ready_future<>();
 }
+
+SEASTAR_TEST_CASE(test_unexpected_exception_format) {
+    try {
+        throw httpd::unexpected_status_error(http::reply::status_type::see_other);
+    } catch (const std::exception& ex) {
+        BOOST_REQUIRE_EQUAL(sstring(ex.what()), format("{}", http::reply::status_type::see_other));
+    }
+    return make_ready_future<>();
+}


include/seastar/http/reply.hh
```diff
@@ -228,6 +228,10 @@ private:
 
 std::ostream& operator<<(std::ostream& os, reply::status_type st);
 
+namespace status_strings {
+const sstring& to_string(reply::status_type status);
```
tchaikov, line 232: instead of exposing `to_string()`, i'd suggest add a `fmt::formatter<>` for it. or if we has no strong interests in the compatibility of fmtlib v9, could just go ahead and implement `format_as()`. so we can kill two birds with one stone:

- expose this helper to a wider audience like `unexpected_status_error` 's constructor
- and include fmtlib in the audience

xemul:
upd:
- format status type into string w/o explicitly exporting the to_string helper

xemul:
upd:
- and add test

include/seastar/http/exception.hh
```diff
@@ -155,7 +155,7 @@ private:
 class unexpected_status_error : public base_exception {
 public:
     unexpected_status_error(http::reply::status_type st)
-        : base_exception("Unexpected reply status", st)
+        : base_exception(fmt::format("{}", st),st)
```
tchaikov, line 158:
```suggestion
        : base_exception(fmt::format("{}", st), st)
```

or `fmt::to_string(st)`. simpler this way.

xemul:
upt:
- use fmt::to_string

tchaikov (approved):
lgtm.

lovio:
Inside the ﻿`unexpected_status_error`, there is only the status. However, the response body sometimes provides more information and currently, the response body is only logged at the debug level. IMHO, it would be better to include the response body in the ﻿unexpected_status_error.

xemul:
That's good point, but response body can be quite large. Currently I've no good ideas how to deal with that

lovio:
Perhaps it's okay to pass an additional ﻿`reply_handler err_handle` to `﻿client::make_request`

xemul:
Or attach `.handle_exception_type([] (unexpected_status_error&) {})` continuation

lovio:
> Or attach `.handle_exception_type([] (unexpected_status_error&) {})` continuation

But the problem is that  `unexpected_status_error` only includes a `status`. How to get response body?

nyh:
> That's good point, but response body can be quite large. Currently I've no good ideas how to deal with that

In Alternator, we have for this purpose a function `truncated_content_view` (see alternator/server.cc) which prints for debugging purposes the first 1024 bytes of the content.

include/seastar/http/exception.hh
```diff
@@ -155,7 +155,7 @@ private:
 class unexpected_status_error : public base_exception {
 public:
     unexpected_status_error(http::reply::status_type st)
-        : base_exception("Unexpected reply status", st)
+        : base_exception(fmt::to_string(st), st)
```
nyh, line 158: So now instead of "Unexpected reply status" you see "continue" (or whatever) - is this clearer? Maybe you need to have both the words "unexpected reply status" and "continue"?
Or maybe you're counting on the printout of the exception's **type** to provide some context? Is the exception type always printed clearly?

xemul:
> But the problem is that unexpected_status_error only includes a status. How to get response body?

Ah, true :(

> In Alternator, we have for this purpose a function truncated_content_view (see alternator/server.cc) which prints for debugging purposes the first 1024 bytes of the content.

But it looks like it only works in case tracing log-level is ON or request tracing is ON. In http client there's body printing on debug log level :)

include/seastar/http/exception.hh
xemul: > So now instead of "Unexpected reply status" you see "continue" (or whatever) - is this clearer? Maybe you need to have both the words "unexpected reply status" and "continue"? Or maybe you're counting on the printout of the exception's **type** to provide some context? Is the exception type always printed clearly?

Yes, I rely on exception type. So far it's been shown in logs, e.g. scylladb/scylladb#16133

image

nyh:
> > In Alternator, we have for this purpose a function truncated_content_view (see alternator/server.cc) which prints for debugging purposes the first 1024 bytes of the content.
> 
> But it looks like it only works in case tracing log-level is ON or request tracing is ON. In http client there's body printing on debug log level :)

Sure, I indeed used it in Alternator for tracing - I was just pointing out to you this function, and this idea - that you can print the beginning of the body, which will be the entire body for short replies, and a truncated one for long replies.

I don't know if it's even a good idea to do this (why should the exception contain the body? normally exceptions don't contain the whole input that caused them!), but I pointed out this function just because you said you didn't know what you could do about a long body.

lovio:
> why should the exception contain the body? normally exceptions don't contain the whole input that caused them!

When using the S3 API, you need to parse the complete XML-formatted response body along with the status code to obtain specific errors. The unexpected reply status provides very little information for future error handling. IMHO, error_handle is necessary to conditionally parse the response body and return an custom exception.

lovio:
Sounds good

xemul:
@nyh , I'd like to merge it as is. The problem with unexpected request body now has two other solutions that suit all the parties :)

xemul:
@nyh , reminder ping

PR #1965, opened 2023-11-24
tls: linearize small packets on send

StephanDollberg:
In Redpanda we were seeing surprisingly large overhead from running with
TLS enabled.

After some debugging we found that this is partly caused by doing an
excessively larger amount of `sendmsg` syscalls compared to when using
plain TCP.

This is caused by the implementation of `tls::session::put` which loops
over each fragment in the passed `net::packet` and treats it separately.
This will result in a call to `gnutls_record_send` for each fragment in
the packet (and consequently a call to `sendmsg` for each fragment via
the vector push function). Besides the `sendmsg` overhead this also
avoids the other associated costs and the extra scheduling that is
caused by `do_for_each`.

This can cause excessive overhead for `net::packets` which represent a
"packet" wrapped by header, trailer etc. In redpanda we were seeing
about 3-4 fragments per packet and hence the same factor of overhead in
terms of `sendmsg` syscalls.

To avoid this issue this patch linearizes the `net::packet` if it's
below the TLS max record size. This forces a single pass through the
whole machinery for smallish packets. Note there is many variations one
could do here but this seems like the safest to avoid running into edge
cases where linearization would actually make things worse.

There certainly is room for further improvement in the TLS
implementation. For example one could batch in the push function instead
of forwarding data to `sendmsg` unconditionally. However at larger send
sizes the syscall overhead should be less of an issue.

This patch improves performance quite measurably.

Below is data for a Redpanda OMB benchmark @ 512MB/s with 50 producers
and 50 consumers (~11KB batch size). This is on 3xi3en.6xlarge brokers.

Aggregated e2e latency changes (numbers are fairly stable compared to
the changes):

p50: 7.2ms -> 5.6ms
p99: 35ms -> 15ms
p999: 74ms -> 43ms

At the same time reactor utilization goes down by ~3%. Note due to how
batching/lingering works in Kafka clients this is while doing ~5% more
RPS (with decresed batch size).

We can further confirm that the amount of `sendmsg` syscalls we do is
down by a factor of ~3 (before vs. with this patch):

```
ubuntu@ip-172-31-16-29:~$ sudo perf stat -e 'syscalls:sys_enter_sendmsg' -p $(pidof redpanda) -r 5 -- sleep 10

 Performance counter stats for process id '14046' (5 runs):

           2790664      syscalls:sys_enter_sendmsg                                     ( +-  0.18% )

ubuntu@ip-172-31-16-29:~$ sudo perf stat -e 'syscalls:sys_enter_sendmsg' -p $(pidof redpanda) -r 5 -- sleep 10

 Performance counter stats for process id '17479' (5 runs):

            989158      syscalls:sys_enter_sendmsg                                     ( +-  0.21% )
```


StephanDollberg:
Below some graphs from the benchmark mentioned above:

image
image

src/net/tls.cc
```diff
@@ -1369,6 +1369,18 @@ class session : public enable_lw_shared_from_this<session> {
...
         }
+
+        // We want to make sure that we call gnutls_record_send with as large
+        // packets as possible. This is because each call to gnutls_record_send
+        // translates to a sendmsg syscall. Further it results in larger TLS
+        // records which makes encryption/decryption faster. Hence to avoid
+        // cases where we would do an extra syscall for something like a 100
+        // bytes header we linearize the packet if it's below the max TLS record
+        // size.
+        if (p.nr_frags() > 1 && p.len() <= 16384) {
```
elcallio, line 1380: Nit, 16k really should be a constant somewhere, but the comment more or less makes up for it. :-)

elcallio (changes requested):
Looks ok to me. However, CI failed ASAN checks. It does not look to be related, but you need to verify this.

StephanDollberg:
> Looks ok to me. However, CI failed ASAN checks. It does not look to be related, but you need to verify this.

Yeah those are https://github.com/scylladb/seastar/issues/1913

elcallio:
> Yeah those are #1913

Good.

src/net/tls.cc
avikivity: 16k isn't small, but maybe for TLS it's reasonable.

Is it possible to use a vectored version of gnutls_record_send? I couldn't find one.

Another option is to change vec_push() to queue the data locally, then send it all at once. But that keeps the extra TLS framing, so it's not a good option.

avikivity: I see gnutls_record_cork() and gnutls_record_uncork(). But I suspect it will generate extra frames, or perhaps it will linearize internally.

avikivity: Looked at the code, cork linearizes. So as far as I can tell this code is the best we can do.

avikivity: re the constant, since it's used once adding a name doesn't really help.

avikivity: (but using one from gnutls will)

StephanDollberg: > Is it possible to use a vectored version of gnutls_record_send? I couldn't find one.

Yeah don't think there is.

> Another option is to change vec_push() to queue the data locally, then send it all at once. But that keeps the extra TLS framing, so it's not a good option.

Yes that is another further improvement but is a bit of larger change. This is being complicated by how do_put and the push function currently interact via `wait_for_output`. 

> I see gnutls_record_cork() and gnutls_record_uncork(). But I suspect it will generate extra frames, or perhaps it will linearize internally.

Yeah corking does linearize internally so it's fairly similar to what this patch does. However I think the solution in this patch is easier as it again avoids any complications introduced by corking in relation to wait_for_output. 

> (but using one from gnutls will)

Done now. I changed it so that we use the actual session such that in the unlikely case of a smaller value being negotiated that is being respected now.

StephanDollberg:
Rebased on master, all tests should be passing now.

PR #1969, opened 2023-11-29
coroutine: generator: specialize promise instead of generator

tchaikov:
the intention of this change was to address #1913, and it turns out the problem is in the implementation of the buffered generator. as it is much more complicated than the unbuffered one. for instance, the latter does not need to handle the case where the generator coroutine is done even before the consumer coroutine is still suspended, or the case where the generator still has values yet consumed, while the consumer coroutine is done.

so, i think the first step is to cleanup the implementation, in order to have a shallow and clean water, so that the bugs in it are more visible.

in this change:

* move the state of promise from the generator into the corresponding promise. so that, instead of looking at generator and its promise, we can just focus on promise without being distracted. and there is no need to expose the accessors for accessing state tracked by the generator. simpler this way.
* consolidate the implementation of generators into a single class, now that the generator does not carry different states for different types of promise, we are able to reuse it. so, `generator_base` is introduced to consolidate these two implementations. strictly speaking, we don't need a common interface for them. it's just convenience to expose the same set of interfaces.
* since the state is managed by promise itself, we can specialize the handling further. so, to communicating with the coroutine waiting for the value, generator_unbuffered_promise does not set the state via generator anymore, instead, it just set the promise, as it is a lazily-started generator coroutine, we can always assume that the caller is expecting the value. this is more straightforward, and more efficient. the same applies to the generator_buffered_promise.
* since the two promises now set the seastar promise with the result directly, next_awaiter is able to retrieve the result from the corresponding future.
* define an enum of state to track the state, this is not a must. as we can always check the corresponding value. but it helps with the readability and debugability. once we have more confidence about the correctness of the implementation, we could remove it.
* let final_suspend() return suspend_always. this allows us to properly destroy the corotine handle in generator's destructor, otherwise, the handle is deleted directly when the generator coroutine returns.
* instead of using a sequence of

  1. _wait_for_next_value->set_value();
  2. _wait_for_next_value = {};

  let's just do this in one-go, so we reset it before set_value() on
  it, this ensures that the coroutine(s) waiting on this promise, to
  see an empty `_wait_for_next_value`. they don't have to assert
  this, but it would help improve the readability by making the
  invariant more clear.

Refs #1913


xemul:
CI decided that this is 'fixes #1913', not 'refs ...' :) Is it?

tchaikov:
if i put the new tests demanded by Avi at https://github.com/scylladb/seastar/pull/1939#issuecomment-1807199947 into the PR, the new tests would fail. and i did root cause the failures . just don't have a satisfying fix for the failures yet..

tchaikov:
this change is a superset of #1939.

tchaikov:
as this change is quite different from #1939. i am closing the latter.

quote from https://github.com/scylladb/seastar/pull/1939#issuecomment-1807199947

> Please add a test that exposes the problem. The fact that the directory lister exposed it means the regular tests are insufficient.

i will come up a a real fix later on based on this change.

tchaikov:

tchaikov:

tchaikov:

tchaikov:

PR #1971, opened 2023-11-30
http: make_request accept optional expected status

lovio:
The original API requires passing the expected status into the `﻿make_request` function. The ﻿reply_handle function will only be called if the expected status matches. Otherwise, an ﻿`httpd::unexpected_status_error` will be returned and there is no way to handle the response body, which contains additional information. With the `httpd::unexpected_status_error`, `reply->status` is not enough. This was discussed in detail on https://github.com/scylladb/seastar/pull/1963

This PR patches the API in a backward incompatible way. If the expected status is ﻿std::nullopt, the reply_handler will handle any reply without missing anything.

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -229,16 +229,16 @@ public:
      * \brief Send the request and handle the response
      *
      * Sends the provided request to the server and calls the provided callback to handle
-     * the response when it arrives. If the reply's status code is not equals the expected
-     * value, the handler is not called and the method resolves with exceptional future.
-     * Otherwise returns the handler's future
+     * the response when it arrives. If the expected status is specified and the response's
+     * status is not the expected one, the handler is not called and the method resolves
+     * with exceptional future. Otherwise returns the handler's future
      *
      * \param req -- request to be sent
      * \param handle -- the response handler
-     * \param expected -- the expected reply status code
+     * \param expected -- the optional expected reply status code, default is std::nullopt
      *
      */
-    future<> make_request(request req, reply_handler handle, reply::status_type expected = reply::status_type::ok);
+    future<> make_request(request req, reply_handler handle, std::optional<reply::status_type> expected = std::nullopt);
 
     /**
      * \brief Updates the maximum number of connections a client may have
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -810,10 +810,23 @@ SEASTAR_TEST_CASE(test_client_unexpected_reply_status) {
                 }
             };
             auto cln = http::experimental::client(std::make_unique<connection_factory>(lcf));
-            auto req = http::request::make("GET", "test", "/test");
-            BOOST_REQUIRE_THROW(cln.make_request(std::move(req), [] (const http::reply& rep, input_stream<char>&& in) {
-                return make_ready_future<>();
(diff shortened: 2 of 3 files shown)


tests/unit/httpd_test.cc
```diff
@@ -810,10 +810,23 @@ SEASTAR_TEST_CASE(test_client_unexpected_reply_status) {
...
+            {
+                auto req = http::request::make("GET", "test", "/test");
+                BOOST_REQUIRE_THROW(cln.make_request(std::move(req), [] (const http::reply& rep, input_stream<char>&& in) {
+                    BOOST_REQUIRE(false); // should throw before handling response
+                    return make_ready_future<>();
+                }, http::reply::status_type::ok).get0(), httpd::unexpected_status_error);
+            }
+            {
+                auto req = http::request::make("GET", "test", "/test");
+                http::reply::status_type status;
```
xemul, line 822: I'd make it `std::optional<status_type>` and ...

tests/unit/httpd_test.cc
```diff
@@ -810,10 +810,23 @@ SEASTAR_TEST_CASE(test_client_unexpected_reply_status) {
...
+                cln.make_request(std::move(req), [&status] (const http::reply& rep, input_stream<char>&& in) {
+                    BOOST_REQUIRE_EQUAL(rep._status, http::reply::status_type::internal_server_error);
```
xemul, line 824: ... dropped this REQUIRE and ...

tests/unit/httpd_test.cc
```diff
@@ -810,10 +810,23 @@ SEASTAR_TEST_CASE(test_client_unexpected_reply_status) {
...
+                    status = rep._status;
+                    return make_ready_future<>();
+                }).get();
+                BOOST_REQUIRE_EQUAL(status, http::reply::status_type::internal_server_error);
```
xemul, line 828: ... put two REQUIREs here:
```
    BOOST_REQUIRE(status.has_value())
    BOOST_REQUIRE_EQUAL(status.value(), status_type::internal_server_error)
```

xemul (changes requested):
LGTM, but please tune up the test a bit

include/seastar/http/client.hh
```diff
@@ -229,16 +229,16 @@ public:
      * \brief Send the request and handle the response
      *
      * Sends the provided request to the server and calls the provided callback to handle
-     * the response when it arrives. If the reply's status code is not equals the expected
-     * value, the handler is not called and the method resolves with exceptional future.
-     * Otherwise returns the handler's future
+     * the response when it arrives. If the expected status is present and not equals the
```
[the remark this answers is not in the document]
lovio: updated

tests/unit/httpd_test.cc
lovio: done

tests/unit/httpd_test.cc
lovio: done

tests/unit/httpd_test.cc
lovio: done

tchaikov:
> http: make_requests accept optional expected

your commit message only has a title. could you please put more details in it? you could reuse the content of the cover letter.

lovio:
> you could reuse the content of the cover letter.

updated

PR #1972, opened 2023-11-30
io: Apply rate-factor early

xemul:
Here's the sched throttling math:

  bw/bw_max + iops/iops_max <= C

The "rate_factor" is the C and it's usually set to be 1.0, but there was a reservation for convex/concave io profiles where the C would be set to lower or higher values to adjust the limit.

The same goal can be achieved by scaling ..._max values on the right side of the equation, like this

  bw/(bw_max*C) + iops/(iopx_max*C) <= 1.0

This makes the fair-queue code a bit simpler and, as a nice side effect, adds sanity check for the rate-factor not been too low.

### DIFF
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -244,7 +244,6 @@ public:
          */
         double min_tokens = 0.0;
         double limit_min_tokens = 0.0;
-        float rate_factor = 1.0;
         std::chrono::duration<double> rate_limit_duration = std::chrono::milliseconds(1);
     };
 
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4131,6 +4131,12 @@ public:
                         throw std::runtime_error(fmt::format("Configured number of queues {} is larger than the maximum {}",
                                                  _mountpoints.size(), _max_queues));
                     }
+
+                    d.read_bytes_rate *= d.rate_factor;
+                    d.write_bytes_rate *= d.rate_factor;
+                    d.read_req_rate *= d.rate_factor;
+                    d.write_req_rate *= d.rate_factor;
+
                     if (d.read_bytes_rate == 0 || d.write_bytes_rate == 0 ||
                             d.read_req_rate == 0 || d.write_req_rate == 0) {
                         throw std::runtime_error(fmt::format("R/W bytes and req rates must not be zero"));
@@ -4170,7 +4176,6 @@ public:
         }
         cfg.mountpoint = p.mountpoint;
         cfg.duplex = p.duplex;
-        cfg.rate_factor = p.rate_factor;
         cfg.rate_limit_duration = latency_goal();
         cfg.flow_ratio_backpressure_threshold = _flow_ratio_backpressure_threshold;
         // Block count limit should not be less than the minimal IO size on the device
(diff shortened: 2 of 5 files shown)


tchaikov:
> Here's the sched throttling math:
> 
> $\frac{bw}{bw_{max}} + \frac{iops}{iops_{max}} <= C $
> 
> The "rate_factor" is the C and it's usually set to be 1.0, but there was a reservation for convex/concave io profiles where the C would be set to lower or higher values to adjust the limit.
> 
> The same goal can be achieved by scaling ..._max values on the right side of the equation, like this
> 
> $\frac{bw}{bw_{max} \times C} + \frac{iops}{iops_{max} \times C} <= 1.0 $
> $\frac{bw}{bw_{max\_c} } + \frac{iops}{iops_{max\_c}} <= 1.0 $
> 
> This makes the fair-queue code a bit simpler and, as a nice side effect, adds sanity check for the rate-factor not been too low.

i reformatted the cover letter and add an extra equation as above for better readability.

src/core/reactor.cc
```diff
@@ -4131,6 +4131,12 @@ class disk_config_params {
                         throw std::runtime_error(fmt::format("Configured number of queues {} is larger than the maximum {}",
                                                  _mountpoints.size(), _max_queues));
                     }
+
+                    d.read_bytes_rate = d.read_bytes_rate * d.rate_factor;
```
tchaikov, line 4135:
```suggestion
                    d.read_bytes_rate *= d.rate_factor;
```

tchaikov: also, if the only sources of `mountpoint_params` are 

- its default constructor if the mount pointer is not configured
- the yaml settings

can we instead have a tiny helper method exposed by `mountpoint_params` like:
```c++
struct mountpoint_params {
    uint64_t read_bytes_rate = std::numeric_limits<uint64_t>::max();
    // ...
    void apply_rate(float rate) {
        read_bytes_rate *= rate;
        write_bytes_rate *= rate;
        // ...
    }
};

namespace YAML {
template<>
struct convert<seastar::mountpoint_params> {
    static bool decode(const Node& node, seastar::mountpoint_params& mp) {
        // ...
        if (node["rate_factor"]) {
            mp.apply_rate(node["rate_factor"].as<float>());
        }
        /// btw, can even fail the conversion if `rate_factor` is 0.
        return true;
    }
};
```

xemul:
> i reformatted the cover letter and add an extra equation as above for better readability.

Appreciated, but most of the time I read commit messages in a terminal, that doesn't have latex format facilities :(

include/seastar/core/fair_queue.hh
```diff
@@ -244,7 +244,6 @@ public:
          */
         double min_tokens = 0.0;
         double limit_min_tokens = 0.0;
-        float rate_factor = 1.0;
```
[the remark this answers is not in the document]
xemul: Well, fair_group is not public interface. It should be in internal namespace, but everybody is lazy to move it there

tchaikov:
> > i reformatted the cover letter and add an extra equation as above for better readability.
> 
> Appreciated, but most of the time I read commit messages in a terminal, that doesn't have latex format facilities :(

yeah, just for the reviewers of this PR.

src/core/reactor.cc
xemul: I think the first suggestion to use `*=` is better. Applied one

xemul:
@avikivity , please review

xemul:
@avikivity , please review

xemul:
@avikivity , please review

avikivity:
sorry, viewing now

Issue #1563, opened 2023-03-20
Setting promise value on a foreign shard in the background runs continuation on the foreign shard


bhalevy:
The following test demonstrates this issue:
```c++
SEASTAR_THREAD_TEST_CASE(foreign_promise_set_in_background_test) {
    promise<int> pr;
    auto origin_shard = this_shard_id();
    auto other_shard = origin_shard ^ 1;

    (void)smp::submit_to(other_shard, [&] {
        auto shard = this_shard_id();
        BOOST_REQUIRE_EQUAL(shard, other_shard);
        pr.set_value(shard);
    });

    auto res = pr.get_future().get();
    BOOST_REQUIRE_EQUAL(res, other_shard);
    BOOST_REQUIRE_EQUAL(this_shard_id(), origin_shard);
}
```

test output in debug mode also shows that the finally continuation in:
https://github.com/scylladb/seastar/blob/1eea95e0201eb676f860fcf3b4faf4f03666b217/src/core/app-template.cc#L161-L163
also runs on the wrong shard:

```
/home/bhalevy/dev/seastar/tests/unit/futures_test.cc(2047): fatal error: in "foreign_promise_set_in_background_test": critical check this_shard_id() == origin_shard has failed [1 != 0]

*** 1 failure is detected in the test module "Master Test Suite"
[...]

PR #1975, opened 2023-12-01
future: future_state_base: assert owner shard in debug mode

bhalevy:
Assert that the future state is accessed only
on the shard on which it was created as the future
and promise objects are not designed to be accessed
across shards.

For example, currenty, if promise.set_value() (or
set_exception) is called on a promise from a foreign
shard the promise.get_future() continuation would
run on the foreign shard.

Fixes scylladb/seastar#1563

### DIFF
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -36,6 +36,7 @@
 #include <seastar/core/task.hh>
 #include <seastar/core/thread_impl.hh>
 #include <seastar/core/function_traits.hh>
+#include <seastar/core/shard_id.hh>
 #include <seastar/util/critical_alloc_section.hh>
 #include <seastar/util/concepts.hh>
 #include <seastar/util/noncopyable_function.hh>
@@ -767,6 +768,20 @@ protected:
     future_state_base* _state;
 
     task* _task = nullptr;
+#ifdef SEASTAR_DEBUG_PROMISE
+    int _task_shard = -1;
+
+    void set_task(task* task) noexcept {
+        _task = task;
+        _task_shard = this_shard_id();
+    }
+    void assert_task_shard() const noexcept;
+#else
+    void set_task(task* task) noexcept {
+        _task = task;
+    }
+    void assert_task_shard() const noexcept { }
+#endif
 
     promise_base(const promise_base&) = delete;
     promise_base(future_state_base* state) noexcept : _state(state) {}
@@ -1118,7 +1133,7 @@ protected:
     void schedule(task* tws, future_state_base* state) noexcept {
         promise_base* p = detach_promise();
         p->_state = state;
-        p->_task = tws;
+        p->set_task(tws);
     }
 
     void do_wait() noexcept;
@@ -1747,7 +1762,7 @@ public:
 private:
     void set_task(task& t) noexcept {
         assert(_promise);
-        _promise->_task = &t;
+        _promise->set_task(&t);
     }
 
     void set_callback(continuation_base<T>* callback) noexcept {
--- a/src/core/future.cc
+++ b/src/core/future.cc
@@ -64,6 +64,9 @@ void promise_base::move_it(promise_base&& x) noexcept {
     // if &x == this.
     _task = x._task;
     x._task = nullptr;
+#ifdef SEASTAR_DEBUG_PROMISE
+    _task_shard = x._task_shard;
+#endif
     _state = x._state;
     x._state = nullptr;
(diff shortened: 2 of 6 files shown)


bhalevy:
- fixed CI failures (hopefully all of them)

include/seastar/core/sharded.hh
```diff
@@ -180,14 +180,26 @@ class sharded {
...
+            // do not wait for instance to be deleted since it is not going to notify us
+            service_deleted();
+        }
+
+        void track_deletion(std::true_type) noexcept {
+            service->_delete_cb = std::bind(std::mem_fn(&entry::service_deleted), this);
+        }
     };
-    std::vector<entry> _instances;
+    std::vector<std::unique_ptr<entry>> _instances;
```
avikivity, line 199: It's really minor, but I can't like the double indirection (unique_ptr->shared_ptr).

How about this:
 1. move `freed` into `async_sharded_service` (which is allocated on the shard)
 2. use `constexpr if` or other tricks to avoid referring to `freed` when `Service` isn't an `async_sharded_service`. After all, in those cases deletion is synchronous.

include/seastar/core/future.hh
```diff
@@ -521,7 +521,11 @@ public:
         assert(_u.st == state::future);
         _u.set_exception(std::move(ex));
     }
-    future_state_base& operator=(future_state_base&& x) noexcept = default;
+    future_state_base& operator=(future_state_base&& x) noexcept {
+        // Let any move-assignment operator handle self-move
+        _u = std::move(x._u);
+        return *this;
```
avikivity, line 547: But, won't the default implementation also call move assignment on every member?

include/seastar/core/future.hh
```diff
@@ -490,11 +491,25 @@ struct future_state_base {
...
+
+    void foreign_access_error() const noexcept;
+    void assert_shard_owner() const noexcept {
+        if (_owner_shard != this_shard_id()) {
+            foreign_access_error();
+        }
+    }
+#else
+    void assert_shard_owner() const noexcept { }
+#endif
```
avikivity, line 506: It's not clear that it's illegal in all cases. One might pass a ready future to a different shard (as a container of an object|exception), or similarly pass an unattached promise to a different shard. It only becomes problematic once a future/promise pair is created.

If we were starting Seastar now, we could say it's illegal, but at this time this change could break existing applications. I think we need to relax it to only apply when there's a connection between the future and the promise.

include/seastar/core/sharded.hh
bhalevy: I'll submit this patch separately since it's worthwhile regardless of the way we handle https://github.com/scylladb/seastar/issues/1563

bhalevy: > I'll submit this patch separately since it's worthwhile regardless of the way we handle https://github.com/scylladb/seastar/issues/1563

Sorry, I've given it another go and sent a new version including the revised version of this patch.
If this version is not agreed upon I'll send the first patch independently.

include/seastar/core/future.hh
bhalevy: This patch was dropped from next version

include/seastar/core/future.hh
bhalevy: Addressed in next version

include/seastar/core/sharded.hh
avikivity: Please send it independently regardless, it's easier to review it in its own context.

bhalevy:
Draft until https://github.com/scylladb/seastar/pull/1976

bhalevy:
rebased

src/core/future.cc
```diff
@@ -257,14 +271,14 @@ void internal::future_base::do_wait() noexcept {
...
-    _promise->_task = &wake_task;
+    _promise->set_task(&wake_task);
     thread_impl::switch_out(thread);
 }
 
 #ifdef SEASTAR_COROUTINES_ENABLED
 void internal::future_base::set_coroutine(task& coroutine) noexcept {
     assert(_promise);
-    _promise->_task = &coroutine;
+    _promise->set_task(&coroutine);
```
avikivity, line 281: Hmm, what about coroutine::switch_to?

bhalevy:
In v4 ([12a8196](https://github.com/scylladb/seastar/commit/12a819664f74e0d09ac44f06ea86526f2d29d133)):
- call `_promise->set_task` from `future::set_task()`

PR #1977, opened 2023-12-04
test: network_interface_test: add tests for format and parse

tchaikov:
to ensure that we adhere to the related RFC. because seastar::inet_address does not have its own test suite, let's colocate it with the tests for network_interfaces() at this moment. we can extract them out once there are more of them.

Refs https://github.com/scylladb/scylladb/issues/16039


elcallio (approved):
Should probably expand the test set, but it is a start. Would be great to pick a test set from other source and use here.

PR #1979, opened 2023-12-04
dpdk: adjust for poller in internal namespace

avikivity:
In be5fe12ebda1eacdf64cc5ee95664dd43f4b7e4d, we moved the poller class to the internal namespace, but did not adjust dpdk.cc. This corrects the problem.

### DIFF
--- a/src/net/dpdk.cc
+++ b/src/net/dpdk.cc
@@ -55,6 +55,7 @@ module seastar;
 #include <seastar/core/sstring.hh>
 #include <seastar/core/memory.hh>
 #include <seastar/core/metrics.hh>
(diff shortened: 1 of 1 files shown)


src/net/dpdk.cc
```diff
@@ -1422,11 +1423,11 @@ class dpdk_qp : public net::qp {
     std::vector<fragment> _frags;
     std::vector<char*> _bufs;
     size_t _num_rx_free_segs = 0;
-    reactor::poller _rx_gc_poller;
+    internal::poller _rx_gc_poller;
```
xemul, line 1426: reactor::poller alias would still work

PR #1980, opened 2023-12-04
scripts: Add ability to add extra configure arguments

xemul:
The build.sh script performs full bulding cycle inside seastar-dev container but generates configure.py arguments out of script's args, no extras (like --enable-dpdk) are allowed. This patch fixes that

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -27,6 +27,7 @@ MODE=$1
 COMPILER=$2
 VERSION=$3
 STANDARD=$4
+CONF_EXTRA=$5
 
 CONFIGURE="--mode=$MODE"
 if [ ! -z "$COMPILER" ]; then
@@ -50,5 +51,5 @@ if [ ! -z "$COMPILER" ]; then
     fi
 fi
 
-./configure.py $CONFIGURE
+./configure.py $CONFIGURE $CONF_EXTRA
 ninja -C build/$MODE


scripts/build.sh
```diff
@@ -27,6 +27,7 @@ MODE=$1
 COMPILER=$2
 VERSION=$3
 STANDARD=$4
+CONF_EXTRA=$5
```
tchaikov, line 30: is this able to capture multiple arguments?  the title claims that it is, but i am not sure.

tchaikov: actually, i'd suggest dropping all these positional arguments, and keep `$MODE` only.

xemul: > is this able to capture multiple arguments? the title claims that it is, but i am not sure.

Yes, it's possible, just put them in quotes in the command line

> actually, i'd suggest dropping all these positional arguments, and keep $MODE only.

I want to use it with and without `--enable-dpdk` and don't want to introduce yet another special-purpose argument to this script

tchaikov: > > is this able to capture multiple arguments? the title claims that it is, but i am not sure.
> 
> Yes, it's possible, just put them in quotes in the command line
> 
> > actually, i'd suggest dropping all these positional arguments, and keep $MODE only.
> 
> I want to use it with and without `--enable-dpdk` and don't want to introduce yet another special-purpose argument to this script

yeah, but the purpose of the comment is to remove all the special-purpose arguments.

xemul: > yeah, but the purpose of the comment is to remove all the special-purpose arguments.

But how?

* `mode` -- you propose to keep one, OK
* `compiler` -- script supports two, gcc and clang and must chose one
* `version` -- in theory it can be merged with `compiler`, but script treats them independently, so will need to parse
* `standard` -- it's used as argument to configure.py, so could be merged with `mode`, but `mode` itself is used independently, so will need to parse

tchaikov: how about 
```bash
MODE=$1
shift
./configure.py --mode=$MODE $*
ninja -C build/$MODE
```
and let the caller to worry about the compiler etc. for instance, if one want to build the `debug` build with clang-19:
```console
$ ./scripts/build.sh debug --compiler /the/home/of/clang-19
```

xemul: Taking into account I've no idea where /the/home/of/clang-19 is :) The
`$ ./scripts/build.sh debug --compiler /the/home/of/clang-19`
vs
`$ ./scripts/build.sh debug clang`
doesn't look appealing

tchaikov: sorry for being obscure. actually, i am talking about a use case where `/usr/bin/${COMPILER}-${VERSION}` does not work. the positional arguments are better when the we can deduce what they are. like the use case of

```console
$ ls /foo/bar
```
where `/foo/bar` is positional argument. as it's obvious. but unlike the paths passed to `ls`, the ordering of compiler, version, standard is not intuitive.

xemul: Well, this script was not supposed to be super intuitive :) The plan was to pack it into Jenkins job with hard-coded compiler version, C++ standard and mode. Eventually this was replaced with circle-CI config and the script remained in its current form

xemul: @tchaikov , so what do you want to do about it?
Options
1. merge this PR
2. drop the script altogether
3. rework the script to parse argument somehow else
4. some other decision

tchaikov: @xemul i think it depends on how we are going to use it. if 

> performs full bulding cycle inside seastar-dev container but generates configure.py arguments out of script's args

still holds, and you want to enable others to use it, then i'd go with option 4. and my suggestion at https://github.com/scylladb/seastar/pull/1980#discussion_r1415157390 still applies. as i'd avoid the "parse argument" part in option 3. i think it could be the `tools/toolchain/dbuild` in seastar.

if we don't intend to improve it other than patching it to occasionally, then i'd suggest drop it.

xemul:
closing in favor of #2011

PR #1985, opened 2023-12-06
build: use a different *_CMAKE_API for CMake 3.27

tchaikov:
CMAKE_EXPERIMENTAL_CXX_MODULE_CMAKE_API is a CMake variable used for testing the experimental C++20 modules support before it is officially available. fortunately, the official support landed on CMake 3.28. but before that we need to enable the C++ module support by setting this variable with a UUID. but this UUID varies in CMake 3.26 and 3.27. before this change, we only encode the UUID for CMake 3.26, since CMake 3.27 and 3.28 have been released. let's update the UUID accordingly. also, because
CMAKE_EXPERIMENTAL_CXX_SCANDEP_SOURCE and
CMAKE_EXPERIMENTAL_CXX_MODULE_DYNDEP are set properly in CMake 3.27, we don't need to set them when building with this version.

and since CMake 3.28 has the official support of C++20 modules, we can dispense with these settings.

this change adds the support for C++20 modules when building with CMake 3.27 and CMake 3.28 and up.

also, since the base image of our CI is now using ubuntu:mantic, which comes with CMake 3.27.4, this change prepare us for performing an extra test with C++20 modules enabled.


tchaikov:
the test failure is a known one, see #1645

avikivity:
cmake is terrible

PR #1984, opened 2023-12-06
build: cleanups configure.py to be more PEP8 compatible

tchaikov:
to make `configure.py` more readable, clean it up a little bit.

PR #1983, opened 2023-12-05
circleci: build with dpdk enabled

tchaikov:
add one more build with the combination of "clang++16", C++20, release with DPDK enabled. so that we can at least test with DPDK build without doubling the size of the test matrix.

PR #1986, opened 2023-12-06
build: add "--enable-cxx-modules" option to configure.py

tchaikov:
so that we can configure CMake to enable C++20 modules support with `configure.py`.

this enables us to build and test with C++20 modules support in our CI, which utilizes configure.py to drive cmake.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -90,6 +90,11 @@ add_tristate(
     help = 'DPDK support')
 add_tristate(
     arg_parser,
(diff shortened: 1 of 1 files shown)


tchaikov:
this change might conflict with #1984, which ever gets merged first. i will rebase the other one.

tchaikov:
once this change gets merged. i will add a change to build C++20 module in our CI after #1983 lands.

PR #1988, opened 2023-12-07
treewide: include used headers

tchaikov:
* core/smp.cc: in 63d39d0698d0e0baeb1413f50710e86c37c9a677, we use `std::make_unique` in this source file, so need to include the corresponding "<memory>"
* net/net.cc: std::map is used, so include "<map>".
* core/scollectd.cc: include sys/socket.h for accessing AF_INET
* net/inet_address.cc: include <algorithm> for using std::all_of()
* core/file.cc: include <coroutine>, as `co_await` requires the `std::coroutine_traits` type. and when compiling C++20 modules, the .cc files are supposed to be self-contained. our own headers does not expose the symbols in standard library, so the .cc file does not get automatically the access to them by putting "export module seastar;" or "module seastar;" like '#include <seastar/foobar.hh>', where foobar.hh includes the C++ standard library headers. in other words, the .cc files have to include the used standard library headers explicitly.

otherwise, we have FTBFS with Seastar_MODULE enabled.

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -25,6 +25,9 @@ module;
 
 #include <algorithm>
 #include <atomic>
+#ifdef SEASTAR_COROUTINES_ENABLED
+#include <coroutine>
+#endif
 #include <deque>
 #include <filesystem>
 #include <functional>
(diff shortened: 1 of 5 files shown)


tchaikov:
@scylladb/seastar-maint hello maintainers, could you help review this change? after addressing the bitrot problem of C++20 modules support, i will add the CI for testing it. the sooner the better.

PR #1991, opened 2023-12-07
build: silence the waring of -Winclude-angled-in-module-purview

tchaikov:
-Winclude-angled-in-module-purview was introduced in clang 18 to complain if the compiler spot something like `"#include <filename>"` in the module purview. as it is a signal that the purview is exposing something which does not belong to it.

but in Seastar, we include the public Seastar headers using angled bracket also, so this warning does not apply to us. as an exception, we _could_ include headers using quotes instead of angled brackets in seastar.cc, as we don't redistribute this source file as a header, so `#include "seastar/foobar.hh"` does not give user an impression that we are including an in-project file. but this does not address the whole problem, because we have other headers including headers using angled bracket in their module purview.

so, let's silence this warning at this moment before we have a better solution.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -98,6 +98,13 @@ target_compile_definitions (seastar-module
 target_compile_options (seastar-module
   PUBLIC
     -U_FORTIFY_SOURCE)
+include (CheckCXXCompilerFlag)
+check_cxx_compiler_flag ("-Winclude-angled-in-module-purview"
+    Warning_IncludeAngledInModulePurview_FOUND)
+if (Warning_IncludeAngledInModulePurview_FOUND)
+  target_compile_options (seastar-module
+    PRIVATE "-Wno-include-angled-in-module-purview")
+endif ()
 
 target_link_libraries (seastar-module
   PUBLIC


tchaikov:

tchaikov:

avikivity:
I don't mind merging, but what are people supposed to do?

I guess there's no problem if people build module-only libraries without support for #include.

Additional option: have a separate layer of #include that are only used from traditional C++ and #includes the private headers used by modules. I don't propose we do this.

tchaikov:
> I don't mind merging, but what are people supposed to do?

i think, they should *not* `#include` system headers in the global purview. it's kind of like `using namespace std` in a header file. but practically, we don't violate this principal, we just choose to use `#include <seastar/foo.hh>` in our headers instead of `#include "seastar/foo.hh"`, in both header files and source files.

> 
> I guess there's no problem if people build module-only libraries without support for #include.

yeah, in a module-only world. developers do not need `#include`, not mention the `#include <foo/bar.h>`. but this is not a world this warning tries to persuade us into. it just don't like the `#include <string>` in the global purview, as we don't own the standard library, and we should not expose it as a part of our public interface.

> 
> Additional option: have a separate layer of #include that are only used from traditional C++ and #includes the private headers used by modules. I don't propose we do this.

`seastar.cc` is kind of using this approach, if i understand what you are proposing.

avikivity:
So maybe there should be -ftreat-include-prefix-as-module-local=seastar/ to allow people to write dual module/traditional libraries.

avikivity:
(the warning is valuable, for the N-1 people who don't know what a global module preview is)

tchaikov:
i concur with you. it's really helpful, i'd say.

PR #1992, opened 2023-12-07
circleci: do not set disable dpdk explicitly

tchaikov:
the job's with_dpdk is off by default, so no need to specify it explicitly.


tchaikov:

PR #1994, opened 2023-12-08
rpc/rpc_types: add formatter for rpc::optional<T>

tchaikov:
{fmt} v10 is less tolerant than {fmt} v9, so let's add a proper formatter for rpc::optional<T>


tchaikov:

PR #1569, opened 2023-03-23
http: more flexible exceptional responses

jcsp:
It is convenient for HTTP request handlers to be able to throw an exceptions with arbitrary responses:
- for general exceptions, setting the response content type and providing the full body.
- for redirects, specify what kind of redirect status should be used

An example of needing to set content type + body in Redpanda is where we have a configuration PUT handler that validates the request, and on invalid input returns a 400 status with a JSON body containing a dict of config key to the validation error.  This wasn't possible when the HTTP exception classes would always assume the content of a `httpd::base_exception` was a plain string, and wrap it in a generic exception JSON doc.

An example of setting a custom status on a redirect is where we redirect certain HTTP requests to raft leader nodes, where the default `moved_permanently` status doesn't make sense, and we use `temporary_redirect` instead.

PR #1995, opened 2023-12-08
httpd: support temporary redirect from inside async reply

dotnwat:
- The redirect_exception is hard coded for permanent move status. In Redpanda we also have cases where we need to use temporary redirect, which has the same Location header but different http status. This PR allows the status code to be customized for the redirect_exception.

- Additionally, this PR handles redirect for thrown from future returned from handler. Previous the redirect was only handled when returned from the synchronous code inside handler.

- Adds test for redirect_exception

### DIFF
--- a/src/http/routes.cc
+++ b/src/http/routes.cc
@@ -77,6 +77,9 @@ std::unique_ptr<http::reply> routes::exception_reply(std::exception_ptr eptr) {
             }
         }
         std::rethrow_exception(eptr);
+    } catch (const redirect_exception& _e) {
+       rep.reset(new http::reply());
+       rep->add_header("Location", _e.url).set_status(_e.status());
     } catch (const base_exception& e) {
         rep->set_status(e.status(), json_exception(e).to_json());
     } catch (...) {
@@ -98,10 +101,6 @@ future<std::unique_ptr<http::reply> > routes::handle(const sstring& path, std::u
             }
             auto r =  handler->handle(path, std::move(req), std::move(rep));
             return r.handle_exception(_general_handler);
-        } catch (const redirect_exception& _e) {
-            rep.reset(new http::reply());
-            rep->add_header("Location", _e.url).set_status(_e.status()).done(
-                    "json");
         } catch (...) {
             rep = exception_reply(std::current_exception());
         }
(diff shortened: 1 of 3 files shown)


src/http/routes.cc
```diff
@@ -77,6 +77,9 @@ std::unique_ptr<http::reply> routes::exception_reply(std::exception_ptr eptr) {
             }
         }
         std::rethrow_exception(eptr);
+    } catch (const redirect_exception& _e) {
+       rep.reset(new http::reply());
+       rep->add_header("Location", _e.url).set_status(_e.status());
```
dotnwat, line 82: Comment from previous PR cc @nyh.

> I think it's better to put this code in exception_reply(), which already handles all the other types of official exception, instead of splitting the exception-handling code between here and exception_reply()

https://github.com/scylladb/seastar/pull/1569#discussion_r1145927726

src/http/routes.cc
```diff
@@ -98,10 +101,6 @@ future<std::unique_ptr<http::reply> > routes::handle(const sstring& path, std::u
             }
             auto r =  handler->handle(path, std::move(req), std::move(rep));
             return r.handle_exception(_general_handler);
-        } catch (const redirect_exception& _e) {
-            rep.reset(new http::reply());
-            rep->add_header("Location", _e.url).set_status(_e.status()).done(
-                    "json");
```
dotnwat, line 104: Note to reviewers: by removing this redirect exception handling we delegate handling to `exception_reply` which handles `redirect_exception` in exactly the same way. However, in `exception_reply` registered exception handlers (via `routes::register_exception_handler`) take precedence. So in principle this could break some users? It seems low risk, but I'm happy to add it back if there is concern.

dotnwat:
cc @nyh @amnonh. this PR is partially replacing https://github.com/scylladb/seastar/pull/1569 which I left a note on and I think it can be closed.

src/http/routes.cc
nyh: I think this is fine and even better (if someone set an exception handler for redirect_exception, they wanted it to be used, not ignored). I don't see how any user will want to rely on a situation where if he set this handler explicitly, it will be ignored. But please correct me if I'm wrong and it's somehow more serious.

@amnonh maybe you can review too.

nyh (approved):

nyh:
@amnonh I am planning to merge this, so please review if you want to not merge it yet.

tchaikov:

PR #1999, opened 2023-12-11
docker: install clang-tools-17

tchaikov:
to build with C++20 modules support, CMake needs to use clang-scan-deps, which is packaged by clang-tools-17, see
https://packages.debian.org/sid/amd64/clang-tools-17/filelist and
https://packages.ubuntu.com/mantic/amd64/clang-tools-17/filelist, so we have to install it in our CI image.


tchaikov:
Probably we can switch to GitHub workflow and use something like https://github.com/aminya/setup-cpp?

tchaikov:
if this change is approved. we might want to redo the image along with the change of #1997

tchaikov:
@avikivity hi Avi, could you please help review this change? and redo the docker image?

PR #2001, opened 2023-12-14
core: enable seastar to run multiple times in a single process

tchaikov:
this series enables us to run multiple `seastar::app_template` instances sequentially in a single process. this is a false requirement in production, as we don't launch seastar app multiple times in a single process. but this is handy if we want to test, for instance, app_template in a single test process, where we need to start multiple app_template instances one after another.

### DIFF
--- a/src/core/exception_hacks.cc
+++ b/src/core/exception_hacks.cc
@@ -90,6 +90,10 @@ static dl_iterate_fn dl_iterate_phdr_org() {
 static std::vector<dl_phdr_info> *phdrs_cache = nullptr;
 
 void init_phdr_cache() {
+    // this process started reactor before, no need to fill the cache
+    if (phdrs_cache) {
+        return;
+    }
     // Fill out elf header cache for access without locking.
     // This assumes no dynamic object loading/unloading after this point
     phdrs_cache = new std::vector<dl_phdr_info>();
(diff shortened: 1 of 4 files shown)


tchaikov:
i tested this change with the debug build of scylla:
```console
$ JAVA_HOME=/usr/lib/jvm/jre-11 pytest  -c $PWD/pytest.ini --cassandra-dir $HOME/dev/scylladb/build/Debug scylla_sstable_test.py
...
=========================================================================================== 3 passed, 8 warnings in 53.08s ===========================================================================================
```

avikivity:
Please add a test that shows two app_templates, perhaps communicating with each other via seastar::alien.

tchaikov:
v2:

- added a test run two instances of `app_template` in sequentially in a single process.

@avikivity the goal of this changeset is not to run multiple `app_template` instances in parallel, but to run them sequentially.

avikivity:
> v2:
> 
> * added a test run two instances of `app_template` in sequentially in a single process.
> 
> @avikivity the goal of this changeset is not to run multiple `app_template` instances in parallel, but to run them sequentially.

Ah ok. I got excited prematurely.

tchaikov:
v3:

- rebase against master HEAD
- disable `app_standard_memory_allocator` in C++ source file instead of using CMake.

PR #2007, opened 2023-12-19
reactor: poll less frequently in debug mode

avikivity:
Usually, seastar loops will try to run in one task, until the task quota expires. This avoids the overhead of breaking out into the reactor and running run_some_tasks(). To avoid going over the task quota and generating high latency, these loops check the need_preempt() function to see whether the task quota expired.

Since many loops often do not preempt, the code paths where they do are a source of bugs. To root them out, need_preempt() is defined in debug mode as always true. As a result, every loop will immediately break into the scheduler, which also checks need_preempt() to see if new I/O has arrived and needs servicing (perhaps by a high priority scheduling_group). The end result is that every loop body execution is accompanies by a reactor poll, which is an expensive operation designed to happen every 0.5ms, not on every vector element (or whatever is being processed in the loop).

To fix this, we split need_preempt() into two functions: the original need_preempt(), which tells loops whether they need to preempt, and a new internal::scheduler_need_preempt(), which tells the scheduler whether it needs to poll for I/O. Outside debug mode, they are the same, since in both cases the goal is to limit latency to the task quota. In debug mode, need_preempt() always returns true (to check the preemption code paths in loops), while scheduler_need_preempt() returns true every 64 calls (to allow for some I/O polling without destroying loop efficiency).

This improves debug mode performance by more than 2X in some tests [1].

[1] https://github.com/scylladb/scylladb/issues/16470

### DIFF
--- a/include/seastar/core/execution_stage.hh
+++ b/include/seastar/core/execution_stage.hh
@@ -250,7 +250,7 @@ private:
             futurize<ReturnType>::apply(_function, unwrap(std::move(wi_in))).forward_to(std::move(wi_ready));
             _stats.function_calls_executed++;
 
-            if (need_preempt()) {
+            if (internal::scheduler_need_preempt()) {
                 _stats.tasks_preempted++;
                 break;
             }
--- a/include/seastar/core/preempt.hh
+++ b/include/seastar/core/preempt.hh
@@ -69,4 +69,29 @@ inline bool need_preempt() noexcept {
 #endif
 }
 
+namespace internal {
+
+
+
+// Same as need_preempt(), but for the scheduler's use. Outside debug
+// mode they have the same meaning - the task quota expired and we need
+// to check for I/O.
+inline
+bool
+scheduler_need_preempt() {
+#ifndef SEASTAR_DEBUG
+    return need_preempt();
+#else
+    // Within the scheduler, preempting all the time (as need_preempt()
+    // does in debug mode) reduces performance drastically since we check
+    // for I/O after every task. Since we don't care about latency in debug
+    // mode, run some random-but-bounded number of tasks instead. Latency
+    // will be high if those tasks are slow, but this is debug mode anyway.
+    static thread_local unsigned counter = 0;
+    return ++counter % 64 == 0;
+#endif
+}
+
+}
+
 }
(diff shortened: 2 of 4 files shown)


avikivity:
Note: I'm checking this with scylladb's test suite too.

avikivity:
v2:
 - reverted call to scheduler_need_preempt() back to original in reactor::run_tasks(). With the previous version, an infinite loop that uses yield() will never poll, since if the call to scheduler_need_preempt() in the inner run_some_tasks() returns true, then the call to scheduler_need_preempt() in the outer reactor::run_tasks() will return false. We need the calls to be consistent.
 - tested with ScyllaDB's test suite, passes.

avikivity:
v3: disable stall detector tests in debug mode, they rely on need_preempt() being at least as reliable as the timer tick, and we made it less so

avikivity:
Entire test suite in debug mode:

before:


CPU utilization: 67.5%

real	36m43.400s
user	719m21.630s
sys	76m9.223s


after:

CPU utilization: 61.2%

real	18m34.403s
user	292m45.508s
sys	72m38.922s


So, 2X improvement in real time, 2.2X improvement in CPU time.

nyh (approved):
Very nice. 

I wonder if even the user-level need_preempt() should really continue to preempt on *every* call - maybe this too is a bit excessive? Maybe it could also be done only one in two (or some other number) calls and speed up debug runs as well - and maybe this can also give us a performance benefit? After all, we want the debug run to jostle some rare code paths - but not necessary take them each and every time if this slows us down.

nyh:
> Note: I'm checking this with scylladb's test suite too.

Please say when you are comfortable with the testing of this patch and want it to be committed.

avikivity:
> Very nice.
> 
> I wonder if even the user-level need_preempt() should really continue to preempt on _every_ call - maybe this too is a bit excessive? Maybe it could also be done only one in two (or some other number) calls and speed up debug runs as well - and maybe this can also give us a performance benefit? After all, we want the debug run to jostle some rare code paths - but not necessary take them each and every time if this slows us down.

I expected this question. Let's say we have a do_for_each(replicas) in a metadata path that sets up a cluster that does not properly capture a value. It works if there's no preemption and fails if there is. So we do want to preempt frequently, otherwise the test that triggers it will be seen as flaky.

Maybe a compromise can be reached, say 1-in-3 with some randomness, but it has to be considered carefully.

avikivity:
> > Note: I'm checking this with scylladb's test suite too.
> 
> Please say when you are comfortable with the testing of this patch and want it to be committed.

I'm happy with the patch.

PR #2013, opened 2023-12-23
condition_variable: add support for wait with abort_source

nvartolomei:
Hi Seastar developers 👋 

I've come across a use-case where I need to abort waiting on a condition variable (e.g. when client aborts the connection).

I've come up with the following patch. I've used sleep_abortable and abortable_fifo as an inspiration. The tests are pretty match copy & paste of the "wait with timeout" ones.

Haven't added "native" coroutine (`when(...)`) support as I couldn't figure out where is most reasonable to check the status of the abort source and to create the subscription. I.e. `constructor` vs `operator co_await`.

Please let me know what you think about this change and whether "native" coroutine support is a requirement for merging (given that there are no other objections).

Also, I've done best-effort for code formatting. Couldn't figure out how to automatically format it and match the codebase style.

### DIFF
--- a/include/seastar/core/condition-variable.hh
+++ b/include/seastar/core/condition-variable.hh
@@ -28,6 +28,7 @@
 #include <functional>
 #endif
 
+#include "seastar/core/abort_source.hh"
 #include <seastar/core/timer.hh>
 #ifdef SEASTAR_COROUTINES_ENABLED
 #   include <seastar/core/coroutine.hh>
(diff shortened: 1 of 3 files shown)


tchaikov:
JFYI, we had a similar change at #1889.

avikivity:
I'm trying to avoid a matrix of every primitive interacting with every other primitive. See https://github.com/scylladb/seastar/pull/1889#issuecomment-1777915851 for how the two can be glues together.

PR #2016, opened 2023-12-26
treewide: s/format/fmt::format/ when appropriate

tchaikov:
this change replace some `format()` calls with `fmt::format()`.

before this change, we use `format()` when formatting to an sstring and when formatting to an std::string.

the upside of this design is that we don't need to differentiate the calls where an sstring is expected from those where an std::string is expected.

but the downside is that we create an `std::string` by deep copy the content of sstring after formatting the parameters and then drop the `sstring` on the floor in the second use case, where only an std::string is expected.

another potential problem is that there are actually three `format()` functions. they are from three different namespaces: `seastar`, `std` and `fmt` in C++20. currently, the `seastar::format()` one wins in the function loopup. because the other two are templated format string, while `seastar::format()` accepts a plain `const char*`, and we always open the `seastar` namespace in this project.

and in future, once we switch over to the compile-time format check, the function signature of `seastar::format()` will be changed from `format(const char* fmt, A&&... a)` to
`format(fmt::format_string<A...> fmt, A&&... a)`, the compiler would have trouble when performing the function lookup because of the ambiguity.

so, to address the second problem we have to change all call sites of this function to one of these three format functions.

but before that, just to prepare for the change, let's change some of them to `fmt::format()` to address both the first problem and the second one.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3818,7 +3818,7 @@ static program_options::selection_value<network_stack_factory> create_network_st
     }
 
     return program_options::selection_value<network_stack_factory>(zis, "network-stack", std::move(candidates), default_stack,
-            format("select network stack (valid values: {})", format_separated(net_stack_names.begin(), net_stack_names.end(), ", ")));
+            fmt::format("select network stack (valid values: {})", format_separated(net_stack_names.begin(), net_stack_names.end(), ", ")));
 }
 
 static program_options::selection_value<reactor_backend_selector>::candidates backend_selector_candidates() {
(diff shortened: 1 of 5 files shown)


tchaikov:
this change is extracted out of a large patch which migrates to the compile-time format. it is submitted in hope to implement the related changes in an iterative manner.

tchaikov:
this change conflicts with #2017. will rebase this change once the other gets merged.

tchaikov:

tchaikov:

PR #2017, opened 2023-12-26
print: deprecate format_separated() 

tchaikov:
`fmt::join()` is a better alternative than `format_separated()`. because Seastar should focus on what it is good at instead of trying to help resolving all the problems.

in this changeset, we replace the only caller site of this function with `fmt::join()`, and mark `format_separated()` deprecated.


tchaikov:
this change conflicts with #2016. will rebase this change once the other gets merged.

tchaikov:

tchaikov:

PR #2018, opened 2023-12-26
Prometheus protobuf add the enhanced functionality from the text based and Native histogram support

amnonh:
While the Prometheus protobuf support was removed, the metrics layer, and specifically the Prometheus text representation, kept evolving.
This series addresses multiple issues:
1. Prometheus protobuf definition (metrics2.proto) uses the later protobuf version with additional definitions towards native histogram support.
2. Prometheus text representation has new features: Label-based filtering, aggregate by label, and skip_when_empty, which were added while the protobuf was not supported.
3. Prometheus native histogram is an alternative representation that supports high resolution with lower resource cost.
4. Seastar approximate_exponential_histogram uses similar logic to the native histogram, which allows Prometheus to choose to send it as a native histogram
5. Prometheus chooses which representation format to use when answering a request based on configuration and the accept header. The default is text representation. The code needs to reflect that logic clearly. 

The series is around the Prometheus code, but each set of changes has its own commit.

### DIFF
--- a/include/seastar/core/internal/estimated_histogram.hh
+++ b/include/seastar/core/internal/estimated_histogram.hh
@@ -127,6 +127,7 @@ class approximate_exponential_histogram {
     static constexpr unsigned PRECISION_BITS = log2floor(Precision);
     static constexpr unsigned BASESHIFT = log2floor(Min);
     static constexpr uint64_t LOWER_BITS_MASK = Precision - 1;
+    static constexpr size_t MIN_ID = log2ceil(Min) * Precision + 1; // id0 (x,1]
     std::array<uint64_t, NUM_BUCKETS> _buckets;
 public:
     approximate_exponential_histogram() {
@@ -325,12 +326,38 @@ public:
     uint64_t& operator[](size_t b) noexcept {
         return _buckets[b];
     }
+    /*!
+     * \brief get_schema the schema of a native histogram
+     *
+     * Native histograms (also known as sparse histograms), are an experimental Prometheus feature.
+     *
+     * schema defines the bucket schema. Currently, valid numbers are -4 <= n <= 8.
+     * They are all for base-2 bucket schemas, where 1 is a bucket boundary in each case, and
+     * then each power of two is divided into 2^n logarithmic buckets.
+     * Or in other words, each bucket boundary is the previous boundary times 2^(2^-n).
+     */
+    int32_t get_schema() const noexcept {
+        return PRECISION_BITS;
+    }
+
+    /*!
+     * \brief return the histogram min as a native histogram ID
+     *
+     * Native histograms (also known as sparse histograms), are an experimental Prometheus feature.
+     *
+     * Native histograms always starts from 1, while approximate_exponential_histogram have a min first bucket
+     * This function returns approximate_exponential_histogram min value as the bucket id of native histogram.
+     */
+    int32_t min_as_native_histogram_id() const noexcept {
+        return MIN_ID;
+    }
 
     seastar::metrics::histogram to_metrics_histogram() const noexcept {
         seastar::metrics::histogram res;
         res.buckets.resize(size() - 1);
         uint64_t cummulative_count = 0;
         res.sample_sum = 0;
+        res.native_histogram = seastar::metrics::native_histogram_info{get_schema(), min_as_native_histogram_id()};
 
         for (size_t i = 0; i < NUM_BUCKETS - 1; i++) {
             auto& v = res.buckets[i];
--- a/include/seastar/core/metrics_types.hh
+++ b/include/seastar/core/metrics_types.hh
@@ -25,6 +25,7 @@
 #include <cstdint>
 #include <vector>
 #include <seastar/util/modules.hh>
+#include <optional>
 #endif
 
 namespace seastar {
@@ -42,7 +43,21 @@ struct histogram_bucket {
     uint64_t count = 0; // number of events.
     double upper_bound = 0;      // Inclusive.
 };
-
+/*!
+ * \brief native histogram specific information
+ *
+ * Native histograms (also known as sparse histograms) are an experimental Prometheus feature.
+ */
+struct native_histogram_info {
+    // schema defines the bucket schema. Currently, valid numbers are -4 <= n <= 8.
+    // They are all for base-2 bucket schemas, where 1 is a bucket boundary in each case, and
+    // then each power of two is divided into 2^n logarithmic buckets.
+    // Or in other words, each bucket boundary is the previous boundary times 2^(2^-n).
+    // In the future, more bucket schemas may be added using numbers < -4 or > 8.
+    int32_t schema;
+    // min_id is the first bucket id of a given schema.
+    int32_t min_id;
+};
 
 /*!
  * \brief Histogram data type
@@ -82,6 +97,8 @@ struct histogram {
      */
     histogram operator+(histogram&& h) const;
 
+    // Native histograms are an experimental Prometheus feature.
+    std::optional<native_histogram_info> native_histogram;
 };
 
 SEASTAR_MODULE_EXPORT_END
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -824,16 +824,16 @@ future<> write_protobuf_representation(output_stream<char>& out, const config& c
     });
 }
 
-bool is_accept_text(const std::string& accept) {
+bool is_accept_protobuf(const std::string& accept) {
     std::vector<std::string> strs;
     boost::split(strs, accept, boost::is_any_of(","));
     for (auto i : strs) {
         boost::trim(i);
         if (boost::starts_with(i, "application/vnd.google.protobuf;")) {
-            return false;
+            return true;
         }
     }
-    return true;
+    return false;
 }
 
 class metrics_handler : public httpd::handler_base  {
@@ -892,20 +892,19 @@ public:
 
     future<std::unique_ptr<http::reply>> handle(const sstring& path,
         std::unique_ptr<http::request> req, std::unique_ptr<http::reply> rep) override {
-        auto is_text_format = !_ctx.allow_protobuf || is_accept_text(req->get_header("Accept"));
+        auto is_protobuf_format = _ctx.allow_protobuf && is_accept_protobuf(req->get_header("Accept"));
         sstring metric_family_name = req->get_query_param("__name__");
         bool prefix = trim_asterisk(metric_family_name);
         bool show_help = req->get_query_param("__help__") != "false";
         std::function<bool(const mi::labels_type&)> filter = make_filter(*req);
-
-        rep->write_body(is_text_format ? "txt" : "proto", [this, is_text_format, metric_family_name, prefix, show_help, filter] (output_stream<char>&& s) {
+        rep->write_body(is_protobuf_format ? "proto" : "txt", [this, is_protobuf_format, metric_family_name, prefix, show_help, filter] (output_stream<char>&& s) {
             return do_with(metrics_families_per_shard(), output_stream<char>(std::move(s)),
-                    [this, is_text_format, prefix, &metric_family_name, show_help, filter] (metrics_families_per_shard& families, output_stream<char>& s) mutable {
-                return get_map_value(families).then([&s, &families, this, is_text_format, prefix, &metric_family_name, show_help, filter]() mutable {
+                    [this, is_protobuf_format, prefix, &metric_family_name, show_help, filter] (metrics_families_per_shard& families, output_stream<char>& s) mutable {
+                return get_map_value(families).then([&s, &families, this, is_protobuf_format, prefix, &metric_family_name, show_help, filter]() mutable {
                     return do_with(get_range(families, metric_family_name, prefix),
-                            [&s, this, is_text_format, show_help, filter](metric_family_range& m) {
-                        return (is_text_format) ? write_text_representation(s, _ctx, m, show_help, filter) :
-                                write_protobuf_representation(s, _ctx, m, filter);
+                            [&s, this, is_protobuf_format, show_help, filter](metric_family_range& m) {
+                        return (is_protobuf_format) ?  write_protobuf_representation(s, _ctx, m, filter) :
+                                write_text_representation(s, _ctx, m, show_help, filter);
                     });
                 }).finally([&s] () mutable {
                     return s.close();
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -74,14 +74,14 @@ static bool write_delimited_to(const google::protobuf::MessageLite& message,
     return true;
 }
 
-static pm::Metric* add_label(pm::Metric* mt, const metrics::impl::metric_id & id, const config& ctx) {
-    mt->mutable_label()->Reserve(id.labels().size() + 1);
+static pm::Metric* add_label(pm::Metric* mt, const metrics::impl::labels_type & id, const config& ctx) {
+    mt->mutable_label()->Reserve(id.size() + 1);
     if (ctx.label) {
         auto label = mt->add_label();
         label->set_name(ctx.label->key());
         label->set_value(ctx.label->value());
     }
-    for (auto &&i : id.labels()) {
+    for (auto &&i : id) {
         auto label = mt->add_label();
         label->set_name(i.first);
         label->set_value(i.second);
@@ -90,17 +90,28 @@ static pm::Metric* add_label(pm::Metric* mt, const metrics::impl::metric_id & id
 }
 
 static void fill_metric(pm::MetricFamily& mf, const metrics::impl::metric_value& c,
-        const metrics::impl::metric_id & id, const config& ctx) {
+        const metrics::impl::labels_type & id, const config& ctx) {
     switch (c.type()) {
     case scollectd::data_type::GAUGE:
         add_label(mf.add_metric(), id, ctx)->mutable_gauge()->set_value(c.d());
         mf.set_type(pm::MetricType::GAUGE);
         break;
-    case scollectd::data_type::SUMMARY:
-        [[fallthrough]];
+    case scollectd::data_type::SUMMARY: {
+        auto& h = c.get_histogram();
+        auto mh = add_label(mf.add_metric(), id,ctx)->mutable_summary();
+        mh->set_sample_count(h.sample_count);
+        mh->set_sample_sum(h.sample_sum);
+        for (auto b : h.buckets) {
+            auto bc = mh->add_quantile();
+            bc->set_value(b.count);
+            bc->set_quantile(b.upper_bound);
+        }
+        mf.set_type(pm::MetricType::SUMMARY);
+        break;
+    }
     case scollectd::data_type::HISTOGRAM:
     {
-        auto h = c.get_histogram();
+        auto& h = c.get_histogram();
         auto mh = add_label(mf.add_metric(), id,ctx)->mutable_histogram();
         mh->set_sample_count(h.sample_count);
         mh->set_sample_sum(h.sample_sum);
@@ -115,7 +126,7 @@ static void fill_metric(pm::MetricFamily& mf, const metrics::impl::metric_value&
     case scollectd::data_type::REAL_COUNTER:
         [[fallthrough]];
     case scollectd::data_type::COUNTER:
-        add_label(mf.add_metric(), id, ctx)->mutable_counter()->set_value(c.ui());
+        add_label(mf.add_metric(), id, ctx)->mutable_counter()->set_value(c.d());
         mf.set_type(pm::MetricType::COUNTER);
         break;
     }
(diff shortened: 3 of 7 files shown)


src/core/prometheus.cc
```diff
@@ -712,23 +723,46 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx,
...
         auto& name = metric_family.name();
         pm::MetricFamily mtf;
-
+        bool empty_metric = true;
         mtf.set_name(ctx.prefix + "_" + name);
         mtf.mutable_metric()->Reserve(metric_family.size());
-        metric_family.foreach_metric([&mtf, &ctx](auto value, auto value_info) {
-            fill_metric(mtf, value, value_info.id, ctx);
+        metric_family.foreach_metric([&mtf, &ctx, &name, &filter, &aggregated_values, &empty_metric, should_aggregate](auto value, auto value_info) {
+            fill_metric(mtf, value, value_info.id.labels(), ctx);
```
tchaikov, line 789: shall we skip this `fill_metric()` call based on the policies used in the next line? in other words, shall we just drop this line?

src/core/prometheus.cc
```diff
@@ -712,23 +723,46 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx,
...
+                return;
+            }
+            try {
+                if (should_aggregate) {
+                    aggregated_values.add(value, value_info.id.labels());
+                } else {
+                    fill_metric(mtf, value, value_info.id.labels(), ctx);
+                    empty_metric = false;
+                }
+            } catch (const std::range_error& e) {
```
tchaikov, line 800: under what circumstances, `fill_metric()` throws a `std::range_error`?

src/core/prometheus.cc
```diff
@@ -712,23 +723,46 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx,
...
+                seastar_logger.warn("Failed fill_metrics {}", name);
+            }
         });
-        if (!write_delimited_to(mtf, &os)) {
-            seastar_logger.warn("Failed to write protobuf metrics");
+        if (!aggregated_values.empty()) {
+            for (auto&& h : aggregated_values.get_values()) {
+                fill_metric(mtf, h.second, h.first, ctx);
+                empty_metric = false;
+            }
```
tchaikov, line 808: can we just drop the `if` check ? as the net result would be identical, i think.

src/proto/metrics2.proto
```diff
@@ -11,72 +11,137 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-syntax = "proto2";
+// This is copied and lightly edited from
+// github.com/prometheus/client_model/io/prometheus/client/metrics.proto
```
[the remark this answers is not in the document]
amnonh: It's confusing that text is taken from the original file, I'll change the commit message to reflect that

src/core/prometheus.cc
amnonh: it was a rebase problem, it shouldn't have been left there, and will be removed

src/core/prometheus.cc
amnonh: It was solved in a different way and this will be removed

src/core/prometheus.cc
amnonh: I will

src/core/prometheus.cc
```diff
@@ -90,14 +90,25 @@ static pm::Metric* add_label(pm::Metric* mt, const metrics::impl::metric_id & id
...
+        const metrics::impl::labels_type & id, const config& ctx) {
     switch (c.type()) {
     case scollectd::data_type::GAUGE:
         add_label(mf.add_metric(), id, ctx)->mutable_gauge()->set_value(c.d());
         mf.set_type(pm::MetricType::GAUGE);
         break;
-    case scollectd::data_type::SUMMARY:
-        [[fallthrough]];
+    case scollectd::data_type::SUMMARY: {
+        auto h = c.get_histogram();
```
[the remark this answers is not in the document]
amnonh: Changing it const ref explicitely

src/core/prometheus.cc
```diff
@@ -712,23 +723,46 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx,
     });
 }
 
-future<> write_protobuf_representation(output_stream<char>& out, const config& ctx, metric_family_range& m) {
-    return do_for_each(m, [&ctx, &out](metric_family& metric_family) mutable {
+future<> write_protobuf_representation(output_stream<char>& out, const config& ctx, metric_family_range& m, std::function<bool(const mi::labels_type&)> filter) {
+    return do_for_each(m, [&ctx, &out, filter](metric_family& metric_family) mutable {
```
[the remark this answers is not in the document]
amnonh: done

src/proto/metrics2.proto
```diff
@@ -11,72 +11,137 @@
...
+// and finally converted to proto3 syntax to make it usable for the
+// gogo-protobuf approach taken within prometheus/prometheus.
+
+syntax = "proto3";
 
 package io.prometheus.client;
-option java_package = "io.prometheus.client";
+option go_package = "io_prometheus_client";
+
+//import "google/protobuf/gogo.proto";
```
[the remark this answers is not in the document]
amnonh: The file is taken from:
https://raw.githubusercontent.com/prometheus/prometheus/release-2.44/prompb/io/prometheus/client/metrics.proto

amnonh:
@tchaikov Thanks for the review, I've forced pushed V2

src/proto/metrics2.proto
```diff
@@ -11,72 +11,137 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-syntax = "proto2";
+// This is copied and lightly edited from
+// github.com/prometheus/client_model/io/prometheus/client/metrics.proto
+// and finally converted to proto3 syntax to make it usable for the
```
tchaikov, line 16: just out of curiosity, why not use the latest release. namely v2.48.1?  because you didn't want to include https://github.com/prometheus/prometheus/commit/644cb3be701fcfebbb639711d5cdc0c79946efd3 ?

src/core/prometheus.cc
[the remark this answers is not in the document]
tchaikov: @amnonh it's not changed.

src/core/prometheus.cc
```diff
@@ -90,17 +90,28 @@ static pm::Metric* add_label(pm::Metric* mt, const metrics::impl::metric_id & id
...
+        const metrics::impl::labels_type & id, const config& ctx) {
     switch (c.type()) {
     case scollectd::data_type::GAUGE:
         add_label(mf.add_metric(), id, ctx)->mutable_gauge()->set_value(c.d());
         mf.set_type(pm::MetricType::GAUGE);
         break;
-    case scollectd::data_type::SUMMARY:
-        [[fallthrough]];
+    case scollectd::data_type::SUMMARY: {
+        auto h = c.get_histogram();
```
tchaikov, line 151: comment not addressed.

src/core/prometheus.cc
```diff
@@ -90,17 +90,28 @@ static pm::Metric* add_label(pm::Metric* mt, const metrics::impl::metric_id & id
...
+            bc->set_value(b.count);
+            bc->set_quantile(b.upper_bound);
+        }
+        mf.set_type(pm::MetricType::SUMMARY);
+        break;
+    }
     case scollectd::data_type::HISTOGRAM:
     {
-        auto h = c.get_histogram();
+        const seastar::metrics::histogram& h = c.get_histogram();
```
tchaikov, line 165: nit, could use `auto&` here.

src/core/prometheus.cc
```diff
@@ -712,23 +723,39 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx,
...
+                empty_metric = false;
+            }
         });
-        if (!write_delimited_to(mtf, &os)) {
-            seastar_logger.warn("Failed to write protobuf metrics");
+        for (auto&& h : aggregated_values.get_values()) {
+            fill_metric(mtf, h.second, h.first, ctx);
+            empty_metric = false;
+        }
+        if (!empty_metric) {
```
tchaikov, line 803: nit, could early return `if (empty_metric)`. less indent this way.

src/core/prometheus.cc
```diff
@@ -89,6 +89,57 @@ static pm::Metric* add_label(pm::Metric* mt, const metrics::impl::labels_type &
...
 
+static void fill_old_type_histogram(const metrics::histogram& h, ::io::prometheus::client::Histogram* mh) {
+    mh->set_sample_count(h.sample_count);
+    mh->set_sample_sum(h.sample_sum);
+    for (auto b : h.buckets) {
+        auto bc = mh->add_bucket();
+        bc->set_cumulative_count(b.count);
+        bc->set_upper_bound(b.upper_bound);
+    }
+
```
tchaikov, line 100: could you please remove this empty line?

include/seastar/core/metrics_types.hh
```diff
@@ -54,6 +55,8 @@ struct histogram_bucket {
 struct histogram {
     uint64_t sample_count = 0;
     double sample_sum = 0;
+    std::optional<int32_t> schema;
```
tchaikov, line 58: might want to add comment to explain how this is connected to protobuf's `Histogram` message definition? otherwise one can hardly figure out what `schema` implies.

include/seastar/core/metrics_types.hh
```diff
@@ -54,6 +55,8 @@ struct histogram_bucket {
...
+    std::optional<int32_t> min_id;
```
tchaikov, line 59: ditto.

src/core/prometheus.cc
```diff
@@ -89,6 +89,57 @@ static pm::Metric* add_label(pm::Metric* mt, const metrics::impl::labels_type &
...
+                bc = nullptr;
+                last_bucket_id = id;
+            }
+        }
+        count = b.count;
+        id++;
+    }
+    if (bc) {
+        bc->set_length(length);
+        bc = nullptr;
```
tchaikov, line 138: what is this for? `bc` is not referenced anymore from this point.

src/core/prometheus.cc
```diff
@@ -89,6 +89,57 @@ static pm::Metric* add_label(pm::Metric* mt, const metrics::impl::labels_type &
...
+    mh->set_sample_sum(h.sample_sum);
+    size_t id = h.min_id.value();
+
+    mh->set_schema(h.schema.value());
+    double last_bucket = 0;
+    double count = 0;
+
+    size_t length = 0;
+    size_t last_bucket_id = 0;
+    ::io::prometheus::client::BucketSpan* bc = nullptr;
```
tchaikov, line 115: could you add a comment explain what this `bc` for? BTW, i think `bc` deserves a better name. what `bc` stands for?

include/seastar/core/internal/estimated_histogram.hh
```diff
@@ -326,12 +327,21 @@ public:
         return _buckets[b];
     }
 
+    size_t get_schema() const noexcept {
+            return PRECISION_BITS;
```
tchaikov, line 331: wrong indent.

include/seastar/core/internal/estimated_histogram.hh
```diff
@@ -326,12 +327,21 @@ public:
...
+    }
+
+    size_t min_as_spars_id() const noexcept {
```
tchaikov, line 334: could you explain what "spars" stands for?

src/core/prometheus.cc
amnonh: replaced here also

include/seastar/core/metrics_types.hh
amnonh: done

src/core/prometheus.cc
amnonh: Added documentation and renamed for clarity

include/seastar/core/internal/estimated_histogram.hh
amnonh: done

src/proto/metrics2.proto
amnonh: Not really. It's just that the PR took so much time to be released, and they kept updating the file. I'm not rushing to add more data on the messages if we don't have to, so it wasn't urgent to keep updating the metrics file if we do not support those extra features.

src/core/prometheus.cc
amnonh: done

amnonh:
@tchaikov Thanks for the comments and suggestions, Just forced push V3, I hope I didn't miss anyone and address them all.

include/seastar/core/internal/estimated_histogram.hh
amnonh: renamed to make it clearer

tchaikov (approved):
lgtm

src/proto/metrics2.proto
tchaikov: i see. was asking because IIUC, the native histogram support in prometheus is still in a flux, so **might** want to use the latest proto definition. if the latest versions did not change the message struct since v2.44 , fine by me.

src/proto/metrics2.proto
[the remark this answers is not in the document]
tchaikov: thanks. with the revise the commit message. it's all good now.

avikivity:
I don't see any tests

src/proto/metrics2.proto
avikivity: This code is protected by a runtime flag, yes?

src/proto/metrics2.proto
```diff
@@ -11,72 +11,137 @@
...
+  double value = 1;
 }
 
 message Histogram {
-  optional uint64 sample_count = 1;
-  optional double sample_sum   = 2;
-  repeated Bucket bucket       = 3; // Ordered in increasing order of upper_bound, +Inf bucket is optional.
+  uint64 sample_count       = 1;
+  double sample_count_float = 4; // Overrides sample_count if > 0.
+  double sample_sum         = 2;
```
avikivity, line 74: Will messages sent using this histogram type be compatible with old servers that used the old histogram type? Not that I care, it's up to Prometheus to maintain compatibility, just curious.

src/core/prometheus.cc
```diff
@@ -712,19 +723,35 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx,
...
         mtf.set_name(ctx.prefix + "_" + name);
         mtf.mutable_metric()->Reserve(metric_family.size());
-        metric_family.foreach_metric([&mtf, &ctx](auto value, auto value_info) {
-            fill_metric(mtf, value, value_info.id, ctx);
+        metric_family.foreach_metric([&mtf, &ctx, &filter, &aggregated_values, &empty_metric, should_aggregate](auto value, auto value_info) {
+            if ((value_info.should_skip_when_empty && value.is_empty()) || !filter(value_info.id.labels())) {
+                return;
+            }
+            if (should_aggregate) {
+                aggregated_values.add(value, value_info.id.labels());
```
avikivity, line 807: Shouldn't it say by which labels? Or is it all-or-nothing?

src/core/prometheus.cc
```diff
@@ -813,13 +840,16 @@ class metrics_handler : public httpd::handler_base  {
                     return do_with(get_range(families, metric_family_name, prefix),
                             [&s, this, is_text_format, show_help, filter](metric_family_range& m) {
                         return (is_text_format) ? write_text_representation(s, _ctx, m, show_help, filter) :
-                                write_protobuf_representation(s, _ctx, m);
+                                write_protobuf_representation(s, _ctx, m, filter);
                     });
+                }).handle_exception([] (auto ep) mutable {
+                    seastar_logger.warn("exception in hande: Failed to write protobuf metrics {}", ep);
                 }).finally([&s] () mutable {
```
avikivity, line 911: Seems unrelated. Also, use std::exception_ptr for the type.

include/seastar/core/metrics_types.hh
```diff
@@ -82,6 +83,17 @@ struct histogram {
...
+    // Native histograms are an experimental Prometheus feature.
+
+    // schema defines the bucket schema. Currently, valid numbers are -4 <= n <= 8.
+    // They are all for base-2 bucket schemas, where 1 is a bucket boundary in each case, and
+    // then each power of two is divided into 2^n logarithmic buckets.
+    // Or in other words, each bucket boundary is the previous boundary times 2^(2^-n).
+    // In the future, more bucket schemas may be added using numbers < -4 or > 8.
+    std::optional<int32_t> schema;
+    // min_id is the first bucket id of a given schema.
+    std::optional<int32_t> min_id;
```
avikivity, line 96: suggest: struct histogram_schema, and make the whole struct member optional.  This avoids ambiguity if one is set and the other is missing.

amnonh:
> I don't see any tests

You are right, and tests are an issue. We want to test that Prometheus agrees with what we are sending. To ensure that if Prometheus changes something in their implementation, our test will fail.

We cannot do that with a unit test. My thinking (which will be done in a separate patch) is to create a demo app that registers multiple metrics with known values and set up a test that starts this app with a Prometheus container and queries Prometheus to validate the value.

include/seastar/core/metrics_types.hh
amnonh: I like the idea, will change

src/proto/metrics2.proto
amnonh: They will not be because of the change in the metrics definitions. This means that if someone enables the protobof on the seastar side, they should use Prometheus 2.0 version or higher.

src/core/prometheus.cc
amnonh: ok

src/proto/metrics2.proto
```diff
@@ -11,72 +11,137 @@
...
+
+  // Everything below here is for native histograms (also known as sparse histograms).
+  // Native histograms are an experimental feature without stability guarantees.
+
+  // schema defines the bucket schema. Currently, valid numbers are -4 <= n <= 8.
+  // They are all for base-2 bucket schemas, where 1 is a bucket boundary in each case, and
+  // then each power of two is divided into 2^n logarithmic buckets.
+  // Or in other words, each bucket boundary is the previous boundary times 2^(2^-n).
+  // In the future, more bucket schemas may be added using numbers < -4 or > 8.
+  sint32 schema           = 5;
```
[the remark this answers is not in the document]
amnonh: I didn't understand the comment

avikivity:
A container will be too clunky. Maybe we can just read the protobuf stuff and see expected values?

src/proto/metrics2.proto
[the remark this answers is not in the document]
avikivity: I thought it was familiar with our floating point like thing, but it's actually different.

amnonh:
> A container will be too clunky. Maybe we can just read the protobuf stuff and see expected values?

That would work,  but it may lead to a situation where the test passes, but in practice, we get the wrong results in Prometheus because they change something on their end.

src/proto/metrics2.proto
tchaikov: you mean the `std::optional<>` ? yes, wanted to make sure we are using the latest schema used by the protected code, even if this piece of code is protected by a flag.

amnonh: @tchaikov, not the optional, the configuration flag that disables the protobuf code altogether

amnonh:
@avikivity I've forced pushed V3 with your comments. I'm working on a test app that will be either a new series on top of this one, or more commits.

amnonh:
@avikivity I've added a commit with a simplified test app. It's not intended to perform an actual test. It simply enables protobuf support and registers two metrics (one of them a histogram with known values), which allows running Prometheus and see that we get the result in a native-histogram format.

image

include/seastar/core/metrics_types.hh
```diff
@@ -42,7 +43,21 @@ struct histogram_bucket {
...
+ *
+ * Native histograms (also known as sparse histograms) are an experimental Prometheus feature.
+ */
+struct native_histogram_info {
+    // schema defines the bucket schema. Currently, valid numbers are -4 <= n <= 8.
+    // They are all for base-2 bucket schemas, where 1 is a bucket boundary in each case, and
+    // then each power of two is divided into 2^n logarithmic buckets.
+    // Or in other words, each bucket boundary is the previous boundary times 2^(2^-n).
+    // In the future, more bucket schemas may be added using numbers < -4 or > 8.
+    int32_t schema;
```
nyh, line 57: As if we didn't have enough confusing meanings for the word "schema" :-(

nyh (approved):
Looks fine to me, although I won't pretend to understand all (or even half) of the details.
I think the metric-writing test is a good start, but you'll also need to add the metric-reading half - i.e., ask the user to manually run Prometheus, run the metric-writing part and also in code, read the metrics back from Prometheus and check that they can be read correctly. We can start with just the writing part, but let's not end this way - a year from now this code can bitrot and nobody will be able to check if this happened.

(Note that I'm not even suggesting full automated CI running this test, prometheus-in-docker or anything of this sort - I'm just suggesting that if a developer wants to run this test a year from now, manually, they should be able to, without knowing what are the expected results).

nyh:
> A container will be too clunky. Maybe we can just read the protobuf stuff and see expected values?

What Avi is suggesting here is an "approval test" in the sense of https://approvaltests.com/ - just checking that the output did not change at all from the previous run, without knowing whether it was correct in the first place or not. I'm not a big fan of this approach (which @kostja used in ScyllaDB's test/cql), but it does work and in this case can quickly give you some sort of regression testing that you don't have yet.

PR #2020, opened 2023-12-26
net/tcp: generate a random src_port hashed to current shard if smp::count > 1

sqddk:
I have fixed an issue in the TCP protocol stack related to the "connect" phase. Now, it will be modified to determine whether to randomly generate a port that can be hashed to the current core based on smp::count. This resolves the TCP connect failure issue occurring in my environment.

### DIFF
--- a/include/seastar/net/tcp.hh
+++ b/include/seastar/net/tcp.hh
@@ -829,18 +829,19 @@ auto tcp<InetTraits>::listen(uint16_t port, size_t queue_length) -> listener {
 
 template <typename InetTraits>
 auto tcp<InetTraits>::connect(socket_address sa) -> connection {
-    uint16_t src_port;
     connid id;
     auto src_ip = _inet._inet.host_address();
     auto dst_ip = ipv4_address(sa);
     auto dst_port = net::ntoh(sa.u.in.sin_port);
 
-    do {
-        src_port = _port_dist(_e);
-        id = connid{src_ip, dst_ip, src_port, dst_port};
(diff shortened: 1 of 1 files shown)


avikivity:
The fix looks good, but please move the explanation into the patch changelog, so it will be visible in the git tree.

sqddk:
> The fix looks good, but please move the explanation into the patch changelog, so it will be visible in the git tree.

no problem

tchaikov:
also, please use a more descriptive title for the commit. i'd suggest use something like:

> net/tcp: generate a random src_port hashed to current shard if smp::count > 1

sqddk:
> also, please use a more descriptive title for the commit. i'd suggest use something like:
> 
> > net/tcp: generate a random src_port hashed to current shard if smp::count > 1

Received! Benefited a lot.

tchaikov:
you might need to revise the commit locally and force push to update it. the cover letter is not used in a single-commit merge.

tchaikov:
> > also, please use a more descriptive title for the commit. i'd suggest use something like:
> > > net/tcp: generate a random src_port hashed to current shard if smp::count > 1
> 
> Received! Benefited a lot.

could you reconsider this suggestion?

and i'd suggest only use ASCII in the commit message. see also https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

tchaikov:
please replace "fix:" with "net/tcp:".

tchaikov (approved):
lgtm

PR #2022, opened 2023-12-29
seastar.cc: replace cryptopp with gnutls when building seastar modules

tchaikov:
in 6bdef1e4319fcc3af078844c2c2ed57f7a6debf3, Crypto++ was dropped, and later the updated build image removed it as well. but it still exists in seastar.cc and src/CMakeLists.txt, let's drop them as well.


tchaikov:
this change is extracted out of #2022 in hope to be got merged sooner.

PR #2024, opened 2024-01-02
util/log: drop unused function

tchaikov:
despite that it is a public function, the removed function is an internal function as per its comment suggest. since it's not used anywhere, let's drop it.


tchaikov:
i tested this change by compiling scylladb after removing this function in its seastar submodule.

PR #2028, opened 2024-01-02
http/client: include used header files

tchaikov:
we failed to include http/client.cc in the seastar module, so let's build it and also include the user headers in that source file.


tchaikov:

PR #2030, opened 2024-01-04
Added documentation for procedure to close a hanging socket

arkocetu:
Added documentation for how to use shared_future/shared_promise

### DIFF
--- a/doc/hanging-socket.md
+++ b/doc/hanging-socket.md
@@ -1,0 +1,99 @@
+Seastar socket peer failure handling
+------------------------------------
+Due to the nature of Seastar & TCP/IP stack, there are cases where reading/writing can hang, in multiple cases
+such as, when a client isn't sending any data, or in the case there is a network failure, that doesn't allow to any data to pass other than Syn.
+
+In such cases reading from a socket might never return.
+
+If we look at the next example, we have an infinite read for any amount of data, but in a case of a Denial of Server multiple connections, can cause high pressure on the system. 
+```cpp
+seastar::future<> handle_connection(seastar::connected_socket s) {
+    auto out = s.output();
+    auto in = s.input();
+
+    return do_with(std::move(s), std::move(out), std::move(in),
+            [] (auto& s, auto& out, auto& in) {
+        return seastar::repeat([&out, &in] {
+            return in.read().then([&out] (auto buf) {
+                if (buf) {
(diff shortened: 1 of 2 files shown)


doc/hanging-socket.md
```diff
@@ -0,0 +1,99 @@
...
+            });
+        }).then([&out] {
+            return out.close();
+        });
+    });
+}
+```
+In this example, the next step after read, will never happen, and we'll never reach the "end" state.
+
+A simple solution for this case might be to add a timeout on the read operation, in the case the client doesn't do any progress, for instance
```
avikivity, line 37: Please mention that it's wrong up-front, before people (or copilot) starts copy'n'pasting it.

doc/hanging-socket.md
```diff
@@ -0,0 +1,99 @@
...
+                    });
+                } else {
+                    return seastar::make_ready_future<seastar::stop_iteration>(
+                            seastar::stop_iteration::yes);
+                }
+            });
+        }).then([&out] {
+            return out.close();
+        });
+    });
```
avikivity, line 61: Better to rewrite the examples are coroutines, nobody should be using continuations in 2024.

doc/hanging-socket.md
```diff
@@ -0,0 +1,99 @@
...
+                        return seastar::stop_iteration::no;
+                    });
+                } else {
+                    return seastar::make_ready_future<seastar::stop_iteration>(
+                            seastar::stop_iteration::yes);
+                }
+            }).handle_exception_type([&s] (const seastar::timed_out_error &err) {
+                // Will resolve the currently waiting futures, with an exception
+                s.shutdown_input();
+                // Allows to wait untill the inputs are actually closed
```
avikivity, line 90: Well this is also wrong, since the original continuation escapes. Maybe it works for the example, but for the general case, it's problematic.

The correct approach is to have a separate monitoring fiber issue the shutdown when it wants. Then the original write() just terminates with an exception. It's also  more efficient as we don't wrap everything with with_timeout wrappers.

doc/hanging-socket.md
arkocetu: Added note about this.

doc/hanging-socket.md
arkocetu: All examples are written with coroutine.

PR #2033, opened 2024-01-08
shared_future: shared_state::run_and_dispose(): release reserve of _peers

denesb:
run_and_dispose() clears _peers but if it was non empty at some point, it might still hold on to some reserve memory after it was cleared. After run_and_dispose() was run, it won't ever contain members again, so this is just a waste, kept around until the shard_state is finally destroyed. Since _peers is a chunked_fifo under the hood, this memory amount might be significant too.
Prevent this waste by resetting _peers to default state at the end of said method.

Refs: scylladb/scylladb#16493

### DIFF
--- a/include/seastar/core/shared_future.hh
+++ b/include/seastar/core/shared_future.hh
@@ -168,6 +168,9 @@ private:
                     _peers.pop_front();
                 }
             }
+            // _peer is now empty, but let's also make sure it releases any
+            // memory it might hold in reserve.
+            _peers = {};
             _keepaliver.release();
         }
 


nyh (approved):
I think this patch is good, but also raises a broader question: are we fine that when _peers is *not* empty, it takes so much memory (you said in another thread 32KB)?  Should we use perhaps a chunked_fifo with a smaller chunk?

nyh:
@denesb there are a few CI failures which seem related to your patch. I wonder if it didn't expose a pre-existing use-after-free bug?

denesb:
> @denesb there are a few CI failures which seem related to your patch. I wonder if it didn't expose a pre-existing use-after-free bug?

No, it is my patch introducing a use-after-free, in the case, where `_keepaliver.release()` is the last reference to `this`. in this case, the object is destroyed by the time the new line I added is executed. I have to swap the order.

avikivity:
> > I think this patch is good, but also raises a broader question: are we fine that when _peers is _not_ empty, it takes so much memory (you said in another thread 32KB)? Should we use perhaps a chunked_fifo with a smaller chunk?
> 
> I checked and it turns out that abortable_fifo has a special optimization for the one-item case. For two items or more, we will use a whole chunk which in chunked_fifo defaults (and abortable_fifo doesn't change it) to 128 items, which apparently in the shared_future case takes 32KB (so 256 bytes per entry). I don't know if this 128-item chunk makes sense, but I'm happy that we also have the 1-item case.

I'm guessing 127 will be much better than 128.

PR #2035, opened 2024-01-10
build: expose flag like -std=c++20 via seastar.pc

tchaikov:
before this change, we don't expose compiling option like `-std=c++20` if Seastar is configured with `-DCMAKE_CXX_STANDARD=20`, and the Seastar application which consumes Seastar via seastar.pc should specify its compiling option by itself. while the CMake config file of Seastar exposes the C++ standard using

```
target_compile_features(seastar
  PUBLIC
    cxx_std_${CMAKE_CXX_STANDARD})
```

this allows the Seastar application consuming Seastar via its CMake config file to use the compile flags exposed by this setting.

after this change, users of seastar.pc will be able to have the `-std=c++20` included the cflags exposed by
`pkgconfig --cflags seastar`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1262,6 +1262,12 @@ if (Seastar_INSTALL)
     set (Seastar_PC ".pc")
   endif ()
 
+  if(CMAKE_CXX_EXTENSIONS)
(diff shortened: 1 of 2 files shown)


denesb (approved):

avikivity:
Tests failed.

tchaikov:
> Tests failed.

should have been fixed in master. lemme rebase the change.

PR #2037, opened 2024-01-10
rpc: fix a deadlock in connection::send()

michoecho:
If send_entry() throws (e.g. because compression is unable to allocate an output buffer), then the intended error handling never gets to run. In particular, the connection isn't aborted, and the promise of outgoing_entry is never completed -- the connection is stuck.

The fix is to extend the exception catching back to the send_entry() call.

### DIFF
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -298,7 +298,7 @@ namespace rpc {
               }
 
               p->uncancellable();
-              return send_entry(*p).then_wrapped([this, p = std::move(p)] (auto f) mutable {
+              return futurize_invoke([this, &pp = *p] { return send_entry(pp); }).then_wrapped([this, p = std::move(p)] (auto f) mutable {
                   if (f.failed()) {
                       f.ignore_ready_future();
                       abort();


src/rpc/rpc.cc
```diff
@@ -298,7 +298,7 @@ namespace rpc {
               }
 
               p->uncancellable();
-              return send_entry(*p).then_wrapped([this, p = std::move(p)] (auto f) mutable {
+              return futurize_invoke([this, &pp = *p] { return send_entry(pp); }).then_wrapped([this, p = std::move(p)] (auto f) mutable {
```
tchaikov, line 301: instead of fixing the caller of `connection::send_entry()`, does it make sense to fix `connection::send_entry()` ? as it would reduce the chance of such kind of errors. coroutines should always return a failed future rather than throwing an actual exception.

xemul: +1 It's better to make send_entry() be noexcept (and maybe even mark it as such)

michoecho: Makes sense, I will move the fix to the callee.

avikivity:
@gleb-cloudius @xemul please review

src/rpc/rpc.cc
michoecho: Done.

michoecho:
Fixed the nitpick. Please re-review and merge.

PR #2038, opened 2024-01-10
rpc: add a way for compressors to send something to their peer

michoecho:
In some situations, we might want to re-negotiate some parameters of RPC compression. For example, a fancy streaming compressor might want to shrink the compression context of the connection to reclaim some memory for other, more important connection that just came up.

But for that, there needs to be some way for the two compressors to communicate with each other, in finite time. Currently, there is no such way. Compressors can send messages to each other by simply prepending them to the actual RPC frames, but there never is a guarantee that an actual RPC frame will be sent anytime soon.

This patch adds a mechanism which allows compressors to communicate. For each RPC connection, it spawns a worker fiber which listens on a condition_variable (passed to the compressor's constructor), and sends an empty RPC frame (eventually/soon) whenever the variable is signaled. A compressor can communicate something to the peer in finite time by requesting an empty frame to be sent, and prepending its message to the next sent frame.

### DIFF
--- a/include/seastar/rpc/rpc_types.hh
+++ b/include/seastar/rpc/rpc_types.hh
@@ -283,6 +283,7 @@ public:
     // decompress data
     virtual rcv_buf decompress(rcv_buf data) = 0;
     virtual sstring name() const = 0;
+    virtual future<> close() noexcept { return make_ready_future<>(); };
     
     // factory to create compressor for a connection
     class factory {
@@ -291,6 +292,12 @@ public:
         // return feature string that will be sent as part of protocol negotiation
         virtual const sstring& supported() const = 0;
         // negotiate compress algorithm
+        // send_empty_frame() requests an empty frame to be sent to the peer compressor on the other side of the connection. 
+        // By attaching a header to this empty frame, the compressor can communicate somthing to the peer,
+        // send_empty_frame() mustn't be called from inside compress() or decompress().
+        virtual std::unique_ptr<compressor> negotiate(sstring feature, bool is_server, std::function<future<>()> send_empty_frame) const {
+            return negotiate(feature, is_server);
+        }
         virtual std::unique_ptr<compressor> negotiate(sstring feature, bool is_server) const = 0;
     };
 };
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -134,7 +134,7 @@ namespace rpc {
 
   future<> connection::send_entry(outgoing_entry& d) noexcept {
     return futurize_invoke([this, &d] {
-      if (_propagate_timeout) {
+      if (d.buf.size && _propagate_timeout) {
           static_assert(snd_buf::chunk_size >= sizeof(uint64_t), "send buffer chunk size is too small");
           if (_timeout_negotiated) {
               auto expire = d.t.get_timeout();
@@ -671,7 +671,7 @@ namespace rpc {
           // supported features go here
           case protocol_features::COMPRESS:
               if (_options.compressor_factory) {
-                  _compressor = _options.compressor_factory->negotiate(e.second, false);
+                  _compressor = _options.compressor_factory->negotiate(e.second, false, [this] { return send({}); });
               }
               if (!_compressor) {
                   throw std::runtime_error(format("RPC server responded with compression {} - unsupported", e.second));
@@ -982,6 +982,8 @@ namespace rpc {
               } else {
                   abort_all_streams();
               }
+          }).finally([this] {
+              return _compressor ? _compressor->close() : make_ready_future();
           }).finally([this]{
               _stopped.set_value();
           });
@@ -1012,7 +1014,7 @@ namespace rpc {
           // supported features go here
           case protocol_features::COMPRESS: {
               if (get_server()._options.compressor_factory) {
-                  _compressor = get_server()._options.compressor_factory->negotiate(e.second, true);
+                  _compressor = get_server()._options.compressor_factory->negotiate(e.second, true, [this] { return send({}); });
                   if (_compressor) {
                        ret[protocol_features::COMPRESS] = _compressor->name();
                   }
@@ -1191,6 +1193,8 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
               } else {
                   return abort_all_streams();
               }
+          }).finally([this] {
+              return _compressor ? _compressor->close() : make_ready_future();
           }).finally([this] {
               _stopped.set_value();
           });
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -461,12 +461,18 @@ namespace rpc {
               }
               auto ptr = compress_header.get();
               auto size = read_le<uint32_t>(ptr);
-              return read_rcv_buf(in, size).then([this, size, &compressor, info] (rcv_buf compressed_data) {
+              return read_rcv_buf(in, size).then([this, size, &compressor, info, &in] (rcv_buf compressed_data) {
                   if (compressed_data.size != size) {
                       _logger(info, format("unexpected eof on a {} while reading compressed data: expected {:d} got {:d}", FrameType::role(), size, compressed_data.size));
                       return FrameType::empty_value();
                   }
                   auto eb = compressor->decompress(std::move(compressed_data));
+                  if (eb.size == 0) {
+                      // Empty frames might be sent as means of communication between the compressors, and should be skipped by the RPC layer.
+                      // We skip the empty frame here. We recursively restart the function, as if the empty frame didn't happen.
+                      // The yield() is here to limit the stack depth of the recursion to 1.
+                      return yield().then([this, info, &in, &compressor] { return read_frame_compressed<FrameType>(info, compressor, in); });
+                  }
                   net::packet p;
                   auto* one = std::get_if<temporary_buffer<char>>(&eb.bufs);
                   if (one) {
--- a/doc/rpc.md
+++ b/doc/rpc.md
@@ -90,7 +90,13 @@ Actual negotiation looks like this:
     uint32_t len
     uint8_t compressed_data[len]
 
-    after compressed_data is uncompressed it becomes regular request, response or streaming frame 
+    After compressed_data is uncompressed, it becomes a regular request, response or streaming frame.
+
+    As a special case, it is allowed to send a compressed frame of size 0 (pre-compression).
+    Such a frame will be a no-op on the receiver.
+    (This can be used as a means of communication between the compressors themselves.
+    If a compressor wants to send some metadata to its peer, it can send a no-op frame,
+    and prepend the metadata as a compressor-specific header.)
 
 ## Request frame format
     uint64_t timeout_in_ms - only present if timeout propagation is negotiated
--- a/include/seastar/rpc/multi_algo_compressor_factory.hh
+++ b/include/seastar/rpc/multi_algo_compressor_factory.hh
@@ -46,18 +46,21 @@ public:
         multi_algo_compressor_factory(std::vector<const rpc::compressor::factory*>(std::move(factories))) {}
     multi_algo_compressor_factory(const rpc::compressor::factory* factory) : multi_algo_compressor_factory({factory}) {}
     // return feature string that will be sent as part of protocol negotiation
-    virtual const sstring& supported() const {
+    const sstring& supported() const override {
         return _features;
     }
     // negotiate compress algorithm
-    virtual std::unique_ptr<compressor> negotiate(sstring feature, bool is_server) const {
+    std::unique_ptr<compressor> negotiate(sstring feature, bool is_server) const override {
+        return negotiate(feature, is_server, nullptr);
+    }
+    std::unique_ptr<compressor> negotiate(sstring feature, bool is_server, std::function<future<>()> send_empty_frame) const override {
         std::vector<sstring> names;
         boost::split(names, feature, boost::is_any_of(","));
         std::unique_ptr<compressor> c;
         if (is_server) {
             for (auto&& n : names) {
                 for (auto&& f : _factories) {
-                    if ((c = f->negotiate(n, is_server))) {
+                    if ((c = f->negotiate(n, is_server, send_empty_frame))) {
                         return c;
                     }
                 }
@@ -65,7 +68,7 @@ public:
         } else {
             for (auto&& f : _factories) {
                 for (auto&& n : names) {
-                    if ((c = f->negotiate(n, is_server))) {
+                    if ((c = f->negotiate(n, is_server, send_empty_frame))) {
                         return c;
                     }
                 }
--- a/tests/unit/rpc_test.cc
+++ b/tests/unit/rpc_test.cc
@@ -20,6 +20,8 @@
  */
 
 #include "loopback_socket.hh"
+#include "seastar/core/condition-variable.hh"
+#include "seastar/core/temporary_buffer.hh"
 #include <seastar/rpc/rpc.hh>
 #include <seastar/rpc/rpc_types.hh>
 #include <seastar/rpc/lz4_compressor.hh>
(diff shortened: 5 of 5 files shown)


michoecho:
Needed by (the current shape of) https://github.com/scylladb/scylla-enterprise/pull/3718.

I expect this series to be controversial, it's rather awkward.

bhalevy:
@avikivity ping, please review.

michoecho:
> Presumably this is needed by streaming RPC connections only?

No, why? You might also want to renegotiate compression params in regular RPC connections.

xemul:
> No, why? 

Because for non-streaming socket you can just send a standard message. Introduce a service verb, a service handler that's registered once connection is created, and use it as any other verbs with handlers

> Each stream client has _parent pointer on the non-stream RPC connection serves messages. So maybe for a stream connection to re-negotiate something one may send a message over the stream's parent client?

So this suggestion still stands

michoecho:
> Because for non-streaming socket you can just send a standard message. Introduce a service verb, a service handler that's registered once connection is created, and use it as any other verbs with handlers

As far as I can see, you can't send a message from the server to the client. A server can only respond. And I want both sides to be able to trigger the renegotiation.

xemul:
> As far as I can see, you can't send a message from the server to the client. A server can only respond. And I want both sides to be able to trigger the renegotiation.

Absolutely, and this is what this PR is about -- bumper calls send() even being on the server-side. My point is that you can leave streaming sockets alone and only equip verb sockets (is there a better name for those?) with bi-directional service messages. Client->server would likely mostly re-use existing facility, server->client would need this bumper hack or similar (though I think that condvar will probably not required too, instantly calling send() looks better)

avikivity:
> But for that, there needs to be some way for the two compressors to communicate with each other, in finite time. Currently, there is no such way. Compressors can send messages to each other by simply prepending them to the actual RPC frames, but there never is a guarantee that an actual RPC frame will be sent anytime soon.


Is re-negotiation really needed? One can open new connection, negotiate whatever they want, and quiesce the old connection in the background.

michoecho:
> Is re-negotiation really needed? One can open new connection, negotiate whatever they want, and quiesce the old connection in the background.

The problem is that the server has no way to initiate that. If something changes on the server, which warrants a re-negotiation (e.g. the server observed a new compression dictionary), AFAIU the server currently can't communicate that to the client in any way other than by terminating the connection, which is too disruptive.

michoecho:
> Is re-negotiation really needed?

I mean, nothing is "really needed", but if you don't have the ability to send a message from the server to the client, the algorithms you can design are more limited than if you do. My current design for dictionary compressors needs that.

michoecho:
In any case, we don't have to commit to this change. We can consider it experimental, and revert it later if we aren't satisfied with it.

avikivity:
> > Is re-negotiation really needed?
> 
> I mean, nothing is "really needed", but if you don't have the ability to send a message from the server to the client, the algorithms you can design are more limited than if you do. My current design for dictionary compressors needs that.

I meant it in a practical rather than absolute sense.

avikivity:
> In any case, we don't have to commit to this change. We can consider it experimental, and revert it later if we aren't satisfied with it.

I'm not against the change if it's needed (in a practical way), I'm only against it if the requirement can be satisfied with existing facilities.

avikivity:
Here's an alternative: the client can send a "please tell me when you want to change compression" rpc that the server can respond to when it has something to say.

It might not be a good fit - rpcs talk to the application behind the server, not the rpc server itself or the compression algorithm. But maybe it works for your need.

include/seastar/rpc/rpc.hh
```diff
@@ -383,6 +385,23 @@ public:
...
+    // something to the other end of the connection (e.g. to re-negotiate
+    // some parameters of compression, such as a dictionary).
+    future<> spawn_compressor_bumper() {
+        return seastar::repeat([this] {
+            return _compressor_needs_progress.wait().then([this] () {
+                return send({})
+                    .handle_exception([] (const auto&) {})
+                    .then([] () { return stop_iteration::no; });
+            });
+        }).handle_exception_type([] (const broken_condition_variable&) {});
```
avikivity, line 403: Note: since 5d3ee980730b49c2379a2728d69698ee68f5dbeb it's permissible (and encouraged) to use coroutines where they make sense.

michoecho:
> It does miss an update of the protocol documentation (doc/rpc.md), maybe it will look worse there.

The protocol change is: after the change, both sides are allowed to send empty RPC frames. Empty frames are no-ops.

src/rpc/rpc.cc
```diff
@@ -465,6 +465,9 @@ namespace rpc {
                       return FrameType::empty_value();
                   }
                   auto eb = compressor->decompress(std::move(compressed_data));
+                  if (eb.size == 0) {
```
gleb-cloudius, line 468: What happens if a new code sends empty frame to the old one?

avikivity:
> > Here's an alternative: the client can send a "please tell me when you want to change compression" rpc that the server can respond to when it has something to say.
> > It might not be a good fit - rpcs talk to the application behind the server, not the rpc server itself or the compression algorithm. But maybe it works for your need.
> 
> It would be technically possible to work with that, but I don't like it. It makes for a spaghetti where RPC depends on compression and compression depends on RPC.

Agree

avikivity:
> > It does miss an update of the protocol documentation (doc/rpc.md), maybe it will look worse there.
> 
> The protocol change is: after the change, both sides are allowed to send empty RPC frames. Empty frames are no-ops.
> 
> I'll update the formal documentation after the overall idea is approved.

It's sort of capturing a scarce resource (we won't be able to use empty frames for other functions).

We could dedicate bit 31 of the frame size to indicate a special frame (and have 2 billion special frame types). But we can also do that if we ever need another special frame type, no need to start now.

src/rpc/rpc.cc
michoecho: The receiver will log an error and drop the connection.

michoecho:
> It's sort of capturing a scarce resource (we won't be able to use empty frames for other functions).

But is there a real chance that we could want empty frames to do be something else than no-ops?

src/rpc/rpc.cc
```diff
@@ -459,7 +459,7 @@ namespace rpc {
               }
               auto ptr = compress_header.get();
               auto size = read_le<uint32_t>(ptr);
-              return read_rcv_buf(in, size).then([this, size, &compressor, info] (rcv_buf compressed_data) {
+              return read_rcv_buf(in, size).then([this, size, &compressor, info, &in] (rcv_buf compressed_data) {
```
gleb-cloudius, line 462: What do we capture `&in` here for?

include/seastar/rpc/rpc_types.hh
```diff
@@ -290,7 +290,7 @@ public:
         // return feature string that will be sent as part of protocol negotiation
         virtual const sstring& supported() const = 0;
         // negotiate compress algorithm
-        virtual std::unique_ptr<compressor> negotiate(sstring feature, bool is_server, condition_variable* compressor_needs_progress) {
+        virtual std::unique_ptr<compressor> negotiate(sstring feature, bool is_server, condition_variable* compressor_needs_progress) const {
```
gleb-cloudius, line 293: Why not squash with the previous patch?

src/rpc/rpc.cc
```diff
@@ -977,6 +978,9 @@ namespace rpc {
               } else {
                   abort_all_streams();
               }
+          }).finally([this] {
+            _compressor_needs_progress.broken();
+            return std::move(_compressor_bumper_exit);
```
gleb-cloudius, line 983: Don't you need to add the same for the server loop?

include/seastar/rpc/rpc.hh
```diff
@@ -383,6 +385,23 @@ public:
...
+    // is to eventually send an empty RPC frame over the connection whenever
+    // the compressor demands it (by signalling a condition variable).
+    //
+    // This is a mechanism which allows the compressor to communicate
+    // something to the other end of the connection (e.g. to re-negotiate
+    // some parameters of compression, such as a dictionary).
+    future<> spawn_compressor_bumper() {
+        return seastar::repeat([this] {
+            return _compressor_needs_progress.wait().then([this] () {
+                return send({})
```
gleb-cloudius, line 399: I do not see the usage in the series, but why can't whoever suppose to signal  `_compressor_needs_progress` simply call `send({})`?

src/rpc/rpc.cc
michoecho: Right, I do. Good catch.

src/rpc/rpc.cc
gleb-cloudius: The capture should be in the previous patch than because otherwise it will not compile by itself.

michoecho: Right. I first wrote the entire thing and then split it into the patches, and didn't check that each commit compiles.

include/seastar/rpc/rpc.hh
michoecho: The purpose of all this is to let the compressor send some information over the connection. So `_compressor_needs_progress` is supposed to be signaled by compressors.

So the problems with calling `send({})` directly are:
1. `send({})` is a private interface of `connection`.
2. Compressors are synchronous. If something calls `send({})`, it should await the future, but compressors have no way to await anything.

You could shuffle things around (e.g. add `future<> close();` to compressors and await it when closing a connection) to make calling `send({})` directly possible, but what you get at the end is going to be an equivalent of this variant with `condition_variable`.

include/seastar/rpc/rpc_types.hh
michoecho: Why do? Usually we like to split patches into smaller ones when possible.

gleb-cloudius: Because you introduced this function in the previous patch and add const to it now. Why not add it from the beginning?

michoecho: Ah, you meant only the `const`.

Right, this is a leftover from an earlier revision, I'll fold the `const` into the previous patch.

include/seastar/rpc/rpc.hh
gleb-cloudius: So the alternative can be:
1. Making connection compressor's friend
2. Pass the connection to a compressor
3. add `future<> compressor::close()`
4. call the `close()` where we call `_compressor_needs_progress.broken()` now

I am not sure this is worse than fiber (which we start regardless if  compression is enabled or not?). @avikivity ?

michoecho: Yes, that's what the alternative would look like. They are pretty much equivalent.

The indirect variant with `condition_variable` (or, say, `std::function`) has the benefit of being easier to write tests with, though. It allows you to create and test compressors without dealing with `connection`s.

michoecho:
Force-push:
- Rebased over a conflict.
- Fixed the unawaited fiber in server's send loop.
- Fixed the bisectability problem and checked that libseastar.a compiles at every commit.
- Squashed the signature change of factory::negotiate into the previous commit.

include/seastar/rpc/rpc.hh
gleb-cloudius: I, personally, would prefer `std::function` + `future<> compressor::close()` to a fiber.

src/rpc/rpc.cc
michoecho: Fixed.

src/rpc/rpc.cc
michoecho: Fixed.

avikivity:
> > It's sort of capturing a scarce resource (we won't be able to use empty frames for other functions).
> 
> But is there a real chance that we could want empty frames to do be something else than no-ops?

No.

include/seastar/rpc/rpc.hh
michoecho: @avikivity Do you have an opinion about this?

avikivity: I agree that std::function is nicer rather than exposing an internal.

michoecho: > I agree that std::function is nicer rather than exposing an internal.

That we all agree with, but that wasn't the question.

Anyway, to avoid dragging the discussion, I'm going to follow Gleb's preference and move the fiber out of `connection`.

michoecho:
@gleb-cloudius I pushed a new version without the fiber, where the connection passes `send({})` wrapped into a `std::function` to the compressor. Check if you like this version more.

I don't like the fact that this version is easier to mess up (e.g. because the compressor implementation has to take care not to call the std::function from within `compress` or `decompress`), but it doesn't really matter.

avikivity:
Please also add tests.

bhalevy:
https://circleci.com/api/v1.1/project/github/scylladb/seastar/18566/output/107/0?file=true&allocation-id=65cfb09da413bc0574563a0d-0-build%2FABCDEFGH
```
46/75 Test #46: Seastar.unit.rpc ..............................***Failed    0.28 sec
[0/1] cd /home/circleci/project/build/debug/tests/unit && /usr/bin/cmake -E env ASAN_OPTIONS=disable_coredump=0:abort_on_error=1:detect_stack_use_after_return=1:verify_asan_link_order=0 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1 BOOST_TEST_CATCH_SYSTEM_ERRORS=no /home/circleci/project/build/debug/tests/unit/rpc_test -- -c 2
Running 34 test cases...
WARNING: debug mode. Not for benchmarking or production
INFO  2024-02-16 19:11:08,371 seastar - Reactor backend: io_uring
INFO  2024-02-16 19:11:08,371 seastar - Perf-based stall detector creation failed (EACCESS), try setting /proc/sys/kernel/perf_event_paranoid to 1 or less to enable kernel backtraces: falling back to posix timer.
random-seed=1436643140
==445==WARNING: ASan doesn't fully support makecontext/swapcontext functions and may produce false positives in some cases!
/home/circleci/project/src/rpc/rpc.cc:986:35: runtime error: member call on null pointer of type 'seastar::rpc::compressor'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/circleci/project/src/rpc/rpc.cc:986:35 
```

michoecho:
> https://circleci.com/api/v1.1/project/github/scylladb/seastar/18566/output/107/0?file=true&allocation-id=65cfb09da413bc0574563a0d-0-build%2FABCDEFGH

In the newest revision I called `_compressor->close()` unconditionally, even though `_compressor` can be null. Fixed.

> Please also add tests.

I'll add tests and update the documentation after the base is approved (so I don't have to fix tests later if the base has to change again). @gleb-cloudius Please comment on the new revision.

src/rpc/rpc.cc
gleb-cloudius: It means that when Scylla start to use this feature we need to remember to introduce a feature flag that guards it.

src/rpc/rpc.cc
```diff
@@ -134,7 +134,7 @@ namespace rpc {
 
   future<> connection::send_entry(outgoing_entry& d) noexcept {
     return futurize_invoke([this, &d] {
-      if (_propagate_timeout) {
+      if (d.buf.size && _propagate_timeout) {
```
gleb-cloudius, line 137: Shouldn't this be part of previous patch?

include/seastar/rpc/rpc_types.hh
```diff
@@ -283,6 +283,7 @@ public:
     // decompress data
     virtual rcv_buf decompress(rcv_buf data) = 0;
     virtual sstring name() const = 0;
```
gleb-cloudius, line 285: > For every connection, a condition_variable is created

You need to update the commit message.

gleb-cloudius (commented):
Need to update one commit message, but otherwise looks fine. A test will help to see how it is actually used. I do not like that we use `negotiate()` argument to pass parameters to compressor creation code, but do not see a better alternative.

src/rpc/rpc.cc
michoecho: We don't need that, because compressors already have their own negotiation mechanism.

If some Scylla version doesn't have the Seastar patch, it also won't have any compressor which uses the empty frames.

I.e. if a new Scylla version includes a new compressor `MY_CUSTOM_COMPRESSOR` which sends empty frames, but it connects to an old Scylla version which doesn't have this compressor, the compression will just be negotiated to `LZ4_FRAGMENTED` and no empty frames will be sent to the old node.

gleb-cloudius: Good point. We are not going to extend existing compressors with the feature, but add new once.

src/rpc/rpc.cc
michoecho: No, because the previous patch is for the reader side, while this patch is for the writer side, and this line belongs to the latter.

avikivity:
Looks ok.

include/seastar/rpc/rpc_types.hh
michoecho: Fixed.

michoecho:
I added a test (new commit). Take a look, is it satisfactory?

gleb-cloudius (approved):

PR #2039, opened 2024-01-11
Adding a Metrics tester app

amnonh:
The challenge in validating the metrics protobuf representation in general and the experimental Prometheus Native-histogram, in particular, is that an actual test involves validating what data the Prometheus server holds.

This series makes such validation easier.
It extends the metrics_tester app to read the metrics it registers from a configuration file.
It adds a utility that reads from both the metrics_tester and a Prometheus server and validates that the value it gets back equals the value taken from the configuration.

### DIFF
--- a/apps/metrics_tester/CMakeLists.txt
+++ b/apps/metrics_tester/CMakeLists.txt
@@ -23,4 +23,5 @@
 seastar_add_app (metrics_tester
   SOURCES metrics_tester.cc)
 
-target_link_libraries (app_metrics_tester)
+target_link_libraries (app_metrics_tester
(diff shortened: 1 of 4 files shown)


amnonh:
@nyh ping, can you take a look

nyh (approved):
Thanks, I'll merge.
A nitpick (for your future consideration for a followup, if you want):
I would have moved (or duplicated) much of the explanations you put in the commit message about what test_metrics.py is and how to use it into a comment at the head of the test_metrics.py script. A year from now, nobody will remember what this script is for or how to use it. Commit messages are fine on the first commit, but if the script is later modified or moved, it becomes hard to find the right commit - and a comment in the code just stays with it.

Issue #2036, opened 2024-01-10
Assertion failure when HTTP server asks to listen to a taken port, with multiple cores


nyh:
This bug was discovered in https://github.com/scylladb/scylladb/issues/16720. The applications runs a Seastar HTTP server by calling something like:
```
    _http_server.listen(socket_address(addr, port));
```
On all cores.

The bug is what happens if we try to listen on an already-taken port.
If there is just one core, this `listen()` throws good exception:
```
std::system_error (error system:98, posix_listen failed for address 127.0.0.1:9042: Address already in use)
```

But if there are multiple cores, while core 0 throws this good exception, all other cores have an assertion failure:
```
scylla: seastar/src/net/posix-stack.cc:559: virtual future<accept_result> seastar::net::posix_ap_server_socket_impl::accept(): Assertion `i.second' failed.
Aborting on shard 1.
```

The backtrace:
```
__assert_fail at ??:?
seastar::net::posix_ap_server_socket_impl::accept() at ??:?
seastar::server_socket::accept() at ??:?
seastar::httpd::http_server::do_accept_one(int, bool) at ??:?
seastar::httpd::http_server::do_accepts(int, bool) at ??:?
seastar::httpd::http_server::listen(seastar::socket_address, seastar::listen_options, seastar::shared_ptr<seastar::tls::server_credentials>) at ??:?
seastar::httpd::http_server::listen(seastar::socket_address) at ??:?
```

The code in `src/net/posix-stack.cc` has a strange `assert(i.second)`, with no explanation what it means, and this is the assert that fails.

This assert was introduced by @gleb-cloudius almost a decade ago, in commit 19addf6e6491923a2d005343c9e08b896211912f. The commit message explains that "_only one thread opens listen socket and runs accept(). Other threads emulate listen()/accept() by waiting for connected socket from the main thread._". I'm not 100% sure that this commit message is actually up to date, because I was under the impression that today all cores accept() connections - and send them to other cores. But I'm not sure.

This also raises another question: Is it important that http_server::listen() be called **first** on core 0, and only then on other cores? Right now I call all of them in parallel, and luckily it does run on core 0 first, but what if it runs first on core 1, before core 0?

nyh:
[...]

PR #2040, opened 2024-01-12
net/posix: Track ap-server ports conflict

xemul:
If creating two listening posix sockets with conflicting ports on non-main thread, the listen() call succeeds, but subsequent server_socket::accept() steps on assertion.

The thing is that posix_ap_server_socket_impl() accept() may try to emplace a promise into thread-local sockets map which is used to wait for the main thread to accept() the connection for real and then send it to this map. The map uses server socket as a key in the map, and when emplaced, the code asserts that insertion did took place, not the existing object was found. This is because listening can only be done once, if there's no listened, main thread sends ready sockets via another thread-local container.

When two ap server sockets bound to the same port call accept() only one of them them manages to _insert_ the promise -- the other one will find that the promise is already there and assertion fires.

The fix is in tracking ports for ap stack and throwing EADDRINUSE in case of conflict early.

fixes: #2036

### DIFF
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -140,15 +140,18 @@ class posix_ap_server_socket_impl : public server_socket_impl {
         conntrack::handle connection_tracking_handle;
         connection(pollable_fd xfd, socket_address xaddr, conntrack::handle cth) : fd(std::move(xfd)), addr(xaddr), connection_tracking_handle(std::move(cth)) {}
     };
+    using port_map_t = std::unordered_set<protocol_and_socket_address>;
     using sockets_map_t = std::unordered_map<protocol_and_socket_address, promise<accept_result>>;
     using conn_map_t = std::unordered_multimap<protocol_and_socket_address, connection>;
+    static thread_local port_map_t ports;
     static thread_local sockets_map_t sockets;
     static thread_local conn_map_t conn_q;
     int _protocol;
     socket_address _sa;
     std::pmr::polymorphic_allocator<char>* _allocator;
 public:
-    explicit posix_ap_server_socket_impl(int protocol, socket_address sa, std::pmr::polymorphic_allocator<char>* allocator = memory::malloc_allocator) : _protocol(protocol), _sa(sa), _allocator(allocator) {}
+    explicit posix_ap_server_socket_impl(int protocol, socket_address sa, std::pmr::polymorphic_allocator<char>* allocator = memory::malloc_allocator);
+    ~posix_ap_server_socket_impl();
     virtual future<accept_result> accept() override;
     virtual void abort_accept() override;
     socket_address local_address() const override {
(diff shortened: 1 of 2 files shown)


gleb-cloudius (approved):

nyh (approved):
Thanks, looks good, but I'll test the original Scylla symptoms before merging.

nyh:
I don't see CI has run on this patch, how come? @benipeled 
I'll test it myself, but it will be on just one type of compiler, etc., not everything usually tested by CI.

benipeled:
> I don't see CI has run on this patch, how come? @benipeled

@xemul's PRs don't trigger the CI [0], someone with circleci write-permissions needs to manually trigger it,
The long-run solution should be migrating the CI to github-actions/Jenkins

[0] https://support.circleci.com/hc/en-us/articles/360043679453-CircleCI-Terms-of-Service-Violation-Sanctioned-Country

nyh:
> > I don't see CI has run on this patch, how come? @benipeled
> 
> @xemul's PRs don't trigger the CI [0]

Wow, this is surprising. I didn't expect that the problem was his **country**. Now that you mention it, I think I might have been aware of this in the past but forgot :-(

> someone with circleci write-permissions needs to manually trigger it, 

How do I "manually trigger" this test? @gleb-cloudius also asked me the same question because he sends patches via mail (not PRs) and wanted to know how to trigger these tests.

> The long-run solution should be migrating the CI to github-actions/Jenkins

It's certainly not helpful to ScyllaDB employees that we have two different CI systems with different quirks, policies and processes. On the other hand, if we do use ScyllaDB's Jenkins, we need to make sure that the Seastar runs are completely open to the general public, not just to ScyllaDB employees (we did this in the past for the OSv projects).

benipeled:
> How do I "manually trigger" this test? @gleb-cloudius also asked me the same question because he sends patches via mail (not PRs) and wanted to know how to trigger these tests.

https://app.circleci.com/pipelines/github/scylladb/seastar + icons-intuition
If you don't have write permissions, AFAIR @bhalevy / @avikivity can help here 

I'm not sure we can run it against patches

nyh:
Unfortunately, it turns out not running CI because of the author's nationality comes with a price: This patch really did fail on other compilers, as the post-merging tests showed.

On G++13 and clang++17:
```
ect/build/dev/checkheaders/include/seastar/net/posix-stack.hh.cc
/home/circleci/project/build/dev/checkheaders/include/seastar/net/posix-stack.hh.cc:143:29: error: 'unordered_set' in namespace 'std' does not name a template type
  143 |     using port_map_t = std::unordered_set<protocol_and_socket_address>;
      |                             ^~~~~~~~~~~~~
/home/circleci/project/build/dev/checkheaders/include/seastar/net/posix-stack.hh.cc:29:1: note: 'std::unordered_set' is defined in header '<unordered_set>'; did you forget to '#include <unordered_set>'?
```

I'll prepare a fix for this myself.

PR #2042, opened 2024-01-13
build: do not install FindProtobuf.cmake

### LINKED ISSUE #2041: It seems that cmake directory doesn't contains FindProtobuf.cmake
When I attempt to install the latest branch using `ninja -C build/release install`, it gives me an error stating that it cannot find `cmake/FindProtobuf.cmake`. Even though I have copied the file with the same name from `/usr/local/share/cmake-3.28/Modules`, it still shows me error. Clearly, the recently introduced Protobuf dependency is preventing the master branch from being easily installed into `/usr/local`.

tchaikov:
263cd37d introduced a regression which tried to install a non-existant file -- cmake/FindProtobuf.cmake.

since this FindProtobuf.cmake is shipped by CMake since 3.13, and our minimal required version of CMake 3.13, there is no need to vendor this file.

so, in this change, we don't install this file.

Fixes #2041

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1359,7 +1359,6 @@ if (Seastar_INSTALL)
     FILES
       ${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindGnuTLS.cmake
       ${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindLinuxMembarrier.cmake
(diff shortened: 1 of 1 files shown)


amnonh:
As long as we can still compile protobuf, LGTM

tchaikov:

avikivity:
We could add a unit test that tests installation and the generated .pc files.

avikivity:
I meant integration test in .circleci

PR #2043, opened 2024-01-14
net: add missing include

nyh:
A recent change to net/posix-stack.hh added a use of std::unordered_set but forgot to include the header file <unordered_set>. This was fine on some versions of the standard library, but ended up in a missing the include in other versions. So let's add the missing include.


tchaikov (approved):
lgtm.

PR #2044, opened 2024-01-15
rpc: extend the use_gate until request processing is finished

gusev-p:
Before the patch we released the `use_gate` in `server::connection::process` when the future, returned by handler `func` is complete. This doesn't mean though that the request processing is finished, it continues in the background and can cause segfaults during node shutdown when `unregister_handler` is called.

To resolve this we extend the `use_gate` until all request related background activity is finished.

The ticket in Scylla: scylladb/scylladb#16382

### DIFF
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -685,7 +685,7 @@ public:
 };
 
 using rpc_handler_func = std::function<future<> (shared_ptr<server::connection>, std::optional<rpc_clock_type::time_point> timeout, int64_t msgid,
-                                                 rcv_buf data)>;
+                                                 rcv_buf data, gate::holder guard)>;
 
 struct rpc_handler {
     scheduling_group sg;
@@ -700,8 +700,11 @@ public:
 protected:
     friend class server;
 
-    virtual rpc_handler* get_handler(uint64_t msg_id) = 0;
-    virtual void put_handler(rpc_handler*) = 0;
+    struct handler_with_holder {
+        rpc_handler& handler;
+        gate::holder holder;
+    };
+    virtual std::optional<handler_with_holder> get_handler(uint64_t msg_id) = 0;
 };
 
 /// \addtogroup rpc
@@ -940,8 +943,7 @@ public:
     }
 
 private:
-    rpc_handler* get_handler(uint64_t msg_id) override;
-    void put_handler(rpc_handler*) override;
+    std::optional<handler_with_holder> get_handler(uint64_t msg_id) override;
 
     template<typename Ret, typename... In>
     auto make_client(signature<Ret(In...)> sig, MsgType t);
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -1157,12 +1157,9 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
 
                       // If the new method of per-connection scheduling group was used, honor it.
                       // Otherwise, use the old per-handler scheduling group.
-                      auto sg = _isolation_config ? _isolation_config->sched_group : h->sg;
-                      return with_scheduling_group(sg, [this, timeout, msg_id, h, data = std::move(data.value())] () mutable {
-                          return h->func(shared_from_this(), timeout, msg_id, std::move(data)).finally([this, h] {
-                              // If anything between get_handler() and here throws, we leak put_handler
-                              get_server()._proto.put_handler(h);
-                          });
+                      auto sg = _isolation_config ? _isolation_config->sched_group : h->handler.sg;
+                      return with_scheduling_group(sg, [this, timeout, msg_id, &h = h->handler, data = std::move(data.value()), guard = std::move(h->holder)] () mutable {
+                          return h.func(shared_from_this(), timeout, msg_id, std::move(data), std::move(guard));
                       });
                   }
               });
--- a/include/seastar/rpc/rpc_impl.hh
+++ b/include/seastar/rpc/rpc_impl.hh
@@ -569,7 +569,8 @@ auto recv_helper(signature<Ret (InArgs...)> sig, Func&& func, WantClientInfo, Wa
     return [func = lref_to_cref(std::forward<Func>(func))](shared_ptr<server::connection> client,
                                                            std::optional<rpc_clock_type::time_point> timeout,
                                                            int64_t msg_id,
-                                                           rcv_buf data) mutable {
+                                                           rcv_buf data,
+                                                           gate::holder guard) mutable {
         auto memory_consumed = client->estimate_request_size(data.size);
         if (memory_consumed > client->max_request_size()) {
             auto err = format("request size {:d} large than memory limit {:d}", memory_consumed, client->max_request_size());
@@ -583,7 +584,7 @@ auto recv_helper(signature<Ret (InArgs...)> sig, Func&& func, WantClientInfo, Wa
             return make_ready_future();
         }
         // note: apply is executed asynchronously with regards to networking so we cannot chain futures here by doing "return apply()"
-        auto f = client->wait_for_resources(memory_consumed, timeout).then([client, timeout, msg_id, data = std::move(data), &func] (auto permit) mutable {
+        auto f = client->wait_for_resources(memory_consumed, timeout).then([client, timeout, msg_id, data = std::move(data), &func, g = std::move(guard)] (auto permit) mutable {
                 // FIXME: future is discarded
                 (void)try_with_gate(client->get_server().reply_gate(), [client, timeout, msg_id, data = std::move(data), permit = std::move(permit), &func] () mutable {
                     try {
@@ -597,7 +598,7 @@ auto recv_helper(signature<Ret (InArgs...)> sig, Func&& func, WantClientInfo, Wa
                         client->get_logger()(client->info(), msg_id, format("caught exception while processing a message: {}", std::current_exception()));
                         return make_ready_future();
                     }
-                }).handle_exception_type([] (gate_closed_exception&) {/* ignore */});
+                }).handle_exception_type([g = std::move(g)] (gate_closed_exception&) {/* ignore */});
         });
 
         if (timeout) {
@@ -701,23 +702,16 @@ bool protocol<Serializer, MsgType>::has_handler(MsgType msg_id) {
 }
 
 template<typename Serializer, typename MsgType>
-rpc_handler* protocol<Serializer, MsgType>::get_handler(uint64_t msg_id) {
-    rpc_handler* h = nullptr;
-    auto it = _handlers.find(MsgType(msg_id));
+std::optional<protocol_base::handler_with_holder> protocol<Serializer, MsgType>::get_handler(uint64_t msg_id) {
+    const auto it = _handlers.find(MsgType(msg_id));
     if (it != _handlers.end()) {
         try {
-            it->second.use_gate.enter();
-            h = &it->second;
+            return handler_with_holder{it->second, it->second.use_gate.hold()};
         } catch (gate_closed_exception&) {
             // unregistered, just ignore
         }
     }
-    return h;
-}
-
-template<typename Serializer, typename MsgType>
-void protocol<Serializer, MsgType>::put_handler(rpc_handler* h) {
-    h->use_gate.leave();
+    return std::nullopt;
 }
 
 template<typename T> T make_shard_local_buffer_copy(foreign_ptr<std::unique_ptr<T>> org);
--- a/tests/unit/rpc_test.cc
+++ b/tests/unit/rpc_test.cc
@@ -1493,6 +1493,29 @@ SEASTAR_TEST_CASE(test_unregister_handler) {
             std::cerr << "call failed in an unexpected way: " << std::current_exception() << std::endl;
             BOOST_REQUIRE(false);
         }
+
+        // verify that unregister_handler waits for pending requests to finish
+        {
+            promise<> handler_reached_promise;
+            promise<> handler_go_promise;
+            sstring value_to_return = "before_unregister";
+            env.register_handler(1, [&]() -> future<sstring> {
+                handler_reached_promise.set_value();
+                return handler_go_promise.get_future().then([&] { return value_to_return; });
+            }).get();
+            auto f = env.proto().make_client<future<sstring>()>(1);
+            auto response_future = f(c1);
+            handler_reached_promise.get_future().get0();
+            auto unregister_future = env.unregister_handler(1).then([&] {
+                value_to_return = "after_unregister";
+            });
+            BOOST_REQUIRE(!unregister_future.available());
+            sleep(1ms).get();
+            BOOST_REQUIRE(!unregister_future.available());
+            handler_go_promise.set_value();
+            unregister_future.get0();
+            BOOST_REQUIRE_EQUAL(response_future.get0(), "before_unregister");
+        }
     });
 }
 


src/rpc/rpc.cc
```diff
@@ -1155,14 +1155,18 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
                           return send_unknown_verb_reply(timeout, msg_id, type);
                       }
 
+                      gate::holder guard;
+                      try {
+                          guard = h->use_gate.hold();
+                      } catch (const gate_closed_exception&) {
+                          return send_unknown_verb_reply(timeout, msg_id, type);
```
gleb-cloudius, line 1162: I would have made `get_handler` return `std::pair<rpc_handler*, gate::holder>` to not repeat error handling here, but this is fine as well.

gleb-cloudius (approved):

src/rpc/rpc.cc
gusev-p: ok, I'll update. I also forgot to add a test to `rpc_test.cc`

kbr-scylla:
> fixes https://github.com/scylladb/scylladb/issues/16382

only after seastar submodule update is merged to Scylla

tchaikov:
@gusev-p Petr, it seems that the build failure was caused by an already-fixed issue. could you rebase for a green test result?

gusev-p:
new version:
  * rebased on master;
  * `get_handler -> handler_with_holder`;
  * improved the `rpc_test/test_unregister_handler`.

gleb-cloudius (approved):

gusev-p:
gentle ping

xemul:
> . This doesn't mean though that the request processing is finished, it continues in the background and can cause segfaults during node shutdown when unregister_handler is called.

Why should rpc layer care about background processing? The handler's future is resolved, processing is finished, if any background activity is still alive there should be another gate/semaphore/future/whatever that's closed/broken/waited/whatever on stop after or in-parallel with handler's unregistration. No?

gleb-cloudius:
> > . This doesn't mean though that the request processing is finished, it continues in the background and can cause segfaults during node shutdown when unregister_handler is called.
> 
> Why should rpc layer care about background processing?

Because it is going to unregister the handler. Theoretically (this is probably not our case, but still)  after unregistering the handler may be destroyed completely. What if it is still runnig?

> The handler's future is resolved, processing is finished, if any background activity is still alive there should be another gate/semaphore/future/whatever that's closed/broken/waited/whatever on stop after or in-parallel with handler's unregistration. No?

Looking at the commit that introduced the gate I see:

```
    This extends the api.  Previously unregister_handler returned void,
    so callers will now be warned of discarding the returned future
    if they don't wait on.  Callers should consider doing that since
    it may be unsafe to proceed after unregister_handler when there
    could be asynchronous users using the handler.
```

So the intention was to return only when there is no async users. TBH if only sync callers are considered the gate is not even needed.

gusev-p:
> Why should rpc layer care about background processing? The handler's future is resolved, processing is finished, if any background activity is still alive there should be another gate/semaphore/future/whatever that's closed/broken/waited/whatever on stop after or in-parallel with handler's unregistration. No?

The resolved future here (`h->func` result) is just an RPC implementation detail, it's not what the user provided callback originally returned. This callback is invoked inside `recv_helper` (`apply(func, ...)`), and my fix extends the `use_gate`'s lifetime to contain the invocation.

xemul:
> Because it is going to unregister the handler. Theoretically (this is probably not our case, but still) after unregistering the handler may be destroyed completely. What if it is still runnig?

Does this "still running" thing continue referencing/pointing to some data that would be freed by the unregistered handler?

xemul:
> > Why should rpc layer care about background processing? The handler's future is resolved, processing is finished, if any background activity is still alive there should be another gate/semaphore/future/whatever that's closed/broken/waited/whatever on stop after or in-parallel with handler's unregistration. No?
> 
> The resolved future here (`h->func` result) is just an RPC implementation detail, it's not what the user provided callback originally returned. This callback is invoked inside `recv_helper` (`apply(func, ...)`), and my fix extends the `use_gate`'s lifetime to contain the invocation.

Yes, that's clear. But it doesn't answer my question -- why should RPC layer care about anything forked into background by the handler function? [For example](https://github.com/scylladb/seastar/pull/2044#issuecomment-1896158052)

> Does this "still running" thing continue referencing/pointing to some data that would be freed by the unregistered handler?

If it doesn't then what does it protect? The only thing that is described here is -- there's an intent to make `unregister_handler()` resolve only after this background processing is finished. My point is that it's not something RPC should deal with, background fibers of that kind are to be collected by the code/subsystem that (un)registers that handler.

gusev-p:
> Does this "still running" thing continue referencing/pointing to some data that would be freed by the unregistered handler?

"still running" thing == user callback + serialisation to socket. The user callback might keep a strong ptr to something, then trying to access it after `unregister_handler` will crash. In my case the problem was different, the `ser::storage_proxy_rpc_verbs::unregister(&_ms)` wasn't waiting for the handlers and `storage_proxy::stop_remote` destroyed `storage_proxy::remote` early.

> But it doesn't answer my question -- why should RPC layer care about anything forked into background by the handler function?

This was the initial intent of `unregister_handler` returning a future, otherwise it should be void. I just fixed a bug.

> My point is that it's not something RPC should deal with, background fibers of that kind are to be collected by the code/subsystem that (un)registers that handler.

This is an overly general statement. Why can't a subsystem that consumes callbacks provide an API to wait for all those callbacks to finish?
  * this allows clients to reuse this API;
  * this API can wait for serialisation to finish, which can't be handled by user code.

xemul:
> "still running" thing == user callback + serialisation to socket.

How come? The serialization of reply back to socket is already running with `use_gate` held. "Still running" is the background fiber only, and such fibers simply cannot serialize anything back to socket.

> The user callback might keep a strong ptr to something, then trying to access it after unregister_handler will crash. In my case the problem was different, the ser::storage_proxy_rpc_verbs::unregister(&_ms) wasn't waiting for the handlers and storage_proxy::stop_remote destroyed storage_proxy::remote early.

OK, but `storage_proxy::remote` can and should protect itself from being used-after-free by background fibers without RPC help.

> This was the initial intent of unregister_handler returning a future, otherwise it should be void. I just fixed a bug.

I don't agree. The original intent of unregister_handler() seems to be in waiting for _synchronous_ part of the callback _and_ serialization back to socket. Because while synchronous part of the handler is running the socket should be kept alive in order to serialize the response back into it. This is what it does already. Anything spawned by the handler is on its own, _unless_ it holds reference/pointer to some RPC object. Does it? What is it? In that case there should be protection on the RPC level.

> Why can't a subsystem that consumes callbacks provide an API to wait for all those callbacks to finish?

It's also overly general statement. Some abstract subsystem can do anything, but we're talking about specific RPC one and so far I only see the intent to ~~ab~~use RPC internal gate to solve external object liveness issue.

xemul:
>  The user callback might keep a strong ptr to something, then trying to access it after unregister_handler will crash

More info about this part, please -- which callback holds a strong ptr on what?

gusev-p:
> The serialization of reply back to socket is already running with `use_gate` held.

You mean `reply_gate()`, not `use_gate`, it covers deserialisation, the callback itself and serialisation of the response.

> OK, but storage_proxy::remote can and should protect itself from being used-after-free by background fibers without RPC help.

I don't know why.

> I don't agree. The original intent of unregister_handler() seems to be in waiting for _synchronous_ part of the callback _and_ serialization back to socket. Because while synchronous part of the handler is running the socket should be kept alive in order to serialize the response back into it. This is what it does already.

There are two gates,  `use_gate` and `reply_gate()`. I'm fixing the first one. Before my patch it was released even before the user provided callback started running.

> Anything spawned by the handler is on its own

Agree.

xemul:
> > The serialization of reply back to socket is already running with `use_gate` held.
> 
> You mean `reply_gate()`, not `use_gate`, it covers deserialisation, the callback itself and serialisation of the response.

```c++
server::connection::process() {
    ...
                      auto h = get_server()._proto.get_handler(type);
                      if (!h) {
                          return send_unknown_verb_reply(timeout, msg_id, type);
                      }

                      // If the new method of per-connection scheduling group was used, honor it.
                      // Otherwise, use the old per-handler scheduling group.
                      auto sg = _isolation_config ? _isolation_config->sched_group : h->sg;
                      return with_scheduling_group(sg, [this, timeout, msg_id, h, data = std::move(data.value())] () mutable {
                          return h->func(shared_from_this(), timeout, msg_id, std::move(data)).finally([this, h] {
                              // If anything between get_handler() and here throws, we leak put_handler
                              get_server()._proto.put_handler(h);
                          });
                      });
```

The get_handler() enters use_gate and put_handler() leaves it. The h->func() thing calls user function and `reply()` that serializes the response.

gusev-p:
> > > The serialization of reply back to socket is already running with `use_gate` held.
> > 
> > 
> > You mean `reply_gate()`, not `use_gate`, it covers deserialisation, the callback itself and serialisation of the response.
> 
> ```c++
> server::connection::process() {
>     ...
>                       auto h = get_server()._proto.get_handler(type);
>                       if (!h) {
>                           return send_unknown_verb_reply(timeout, msg_id, type);
>                       }
> 
>                       // If the new method of per-connection scheduling group was used, honor it.
>                       // Otherwise, use the old per-handler scheduling group.
>                       auto sg = _isolation_config ? _isolation_config->sched_group : h->sg;
>                       return with_scheduling_group(sg, [this, timeout, msg_id, h, data = std::move(data.value())] () mutable {
>                           return h->func(shared_from_this(), timeout, msg_id, std::move(data)).finally([this, h] {
>                               // If anything between get_handler() and here throws, we leak put_handler
>                               get_server()._proto.put_handler(h);
>                           });
>                       });
> ```
> 
> The get_handler() enters use_gate and put_handler() leaves it. The h->func() thing calls user function and `reply()` that serializes the response.

No. `func` is bound to `recv_helper`, the future it returns resolves when `wait_for_resources` finishes, but it doesn't wait for user callback.

gleb-cloudius:
> The get_handler() enters use_gate and put_handler() leaves it. The h->func() thing calls user function and `reply()` that serializes the response.

h->func() calls user function asynchronously. There is no point in the gate as it is used now. What does it protect exactly?

xemul:
> h->func() calls user function asynchronously.

Ah, right, I missed that it (void)-s the (user func + reply)'s future :(

include/seastar/rpc/rpc.hh
```diff
@@ -700,8 +700,11 @@ public:
 protected:
     friend class server;
 
-    virtual rpc_handler* get_handler(uint64_t msg_id) = 0;
-    virtual void put_handler(rpc_handler*) = 0;
+    struct handler_with_holder {
```
xemul, line 703: I think we don't need this wrapper. Just return the rpc_handler to the caller and make _it_ enter the gate and propagate the holder down the stack

gusev-p: This was discussed in [this comment](https://github.com/scylladb/seastar/pull/2044#discussion_r1452073859) - this wrap allows for less code repetition and preserve the current code structure.

gleb-cloudius:
> > h->func() calls user function asynchronously.
> 
> Ah, right, I missed that it (void)-s the (user func + reply)'s future :(

No need to be sad about it, we do not want to execute RPCs one by on :)

xemul:
> No need to be sad about it, we do not want to execute RPCs one by on :)

Maybe we do -- no concurrency, no races :stuck_out_tongue:

include/seastar/rpc/rpc.hh
xemul: >  ... and preserve the current code structure.

It's not a benefit, but anyway, since everybody else prefers yet-another-wrapper I'm fine too

kbr-scylla:
I got this segfault stacktrace today with new Seastar (when trying to execute a schema change):
```
seastar::rpc::server::connection::process()::$_0::operator()()::{lambda()#1}::operator()() const::{lambda()#2}::operator()()::{lambda(std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >)#1}::operator()(std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >) const::{lambda()#1}::operator()() at rpc.cc:?
_ZN7seastar20noncopyable_functionIFNS_6futureIvEEOSt5tupleIJSt8optionalImEmlS4_INS_3rpc7rcv_bufEEEEEE17direct_vtable_forIZNS1_IS9_E4thenIZZZZNS6_6server10connection7processEvEN3$_0clEvENKUlvE_clEvENUlvE0_clEvEUlS9_E_S2_EET0_OT_EUlDpOT_E_E4callEPKSC_SA_ at rpc.cc:?
void seastar::futurize<seastar::future<void> >::satisfy_with_result_of<seastar::future<std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> > >::then_impl_nrvo<seastar::noncopyable_function<seastar::future<void> (std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >&&)>, seastar::future<void> >(seastar::noncopyable_function<seastar::future<void> (std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >&&)>&&)::{lambda(seastar::internal::promise_base_with_type<void>&&, seastar::noncopyable_function<seastar::future<void> (std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >&&)>&, seastar::future_state<std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> > >&&)#1}::operator()(seastar::internal::promise_base_with_type<void>&&, seastar::noncopyable_function<seastar::future<void> (std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >&&)>&, seastar::future_state<std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> > >&&) const::{lambda()#1}>(seastar::internal::promise_base_with_type<void>&&, seastar::noncopyable_function<seastar::future<void> (std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >&&)>&&) at ??:?
seastar::continuation<seastar::internal::promise_base_with_type<void>, seastar::noncopyable_function<seastar::future<void> (std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >&&)>, seastar::future<std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> > >::then_impl_nrvo<seastar::noncopyable_function<seastar::future<void> (std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >&&)>, seastar::future<void> >(seastar::noncopyable_function<seastar::future<void> (std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >&&)>&&)::{lambda(seastar::internal::promise_base_with_type<void>&&, seastar::noncopyable_function<seastar::future<void> (std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> >&&)>&, seastar::future_state<std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> > >&&)#1}, std::tuple<std::optional<unsigned long>, unsigned long, long, std::optional<seastar::rpc::rcv_buf> > >::run_and_dispose() at ??:?
seastar::reactor::run_some_tasks() at ??:?

```


If I'm deciphering this correctly, the segfault happened in the lambda passed to `read_request_frame_compressed(_read_buf).then`. Could it be related to this change?
```cpp
-                      auto sg = _isolation_config ? _isolation_config->sched_group : h->sg;
-                      return with_scheduling_group(sg, [this, timeout, msg_id, h, data = std::move(data.value())] () mutable {
-                          return h->func(shared_from_this(), timeout, msg_id, std::move(data)).finally([this, h] {
-                              // If anything between get_handler() and here throws, we leak put_handler
-                              get_server()._proto.put_handler(h);
-                          });
+                      auto sg = _isolation_config ? _isolation_config->sched_group : h->handler.sg;
+                      return with_scheduling_group(sg, [this, timeout, msg_id, &h = h->handler, data = std::move(data.value()), guard = std::move(h->holder)] () mutable {
+                          return h.func(shared_from_this(), timeout, msg_id, std::move(data), std::move(guard));
                       });
```

gleb-cloudius:
> I got this segfault stacktrace today with new Seastar (when trying to execute a schema change)

Was it during shutdown?

kbr-scylla:
> Was it during shutdown?

No

gleb-cloudius:
> > Was it during shutdown?
> 
> No

In that case the patch should not make any difference at all.

gleb-cloudius:
> > > Was it during shutdown?
> > 
> > 
> > No
> 
> In that case the patch should not make any difference at all.

Though this `&h = h->handler` looks suspicious.

kbr-scylla:
After recompilation it stopped reproducing :( I think I didn't change anything, but can't be sure. We'll see if it happens again

Issue #734, opened 2020-04-01
Add HTTP/2 support to the HTTP server


nyh:
HTTP/2 added a number of interesting features beyond HTTP 1.1, such as request multiplexing and header compression. It would be good if Seastar's builtin HTTP server implementation, which is currently limited to HTTP 1.1, also implemented HTTP/2. We probably want this support for both plain-text and encrypted (TLS) connections.

Seastar-using projects are already waiting for this - e.g., https://github.com/scylladb/scylla/issues/6113.

jewertow:
Hello @nyh, could I work on this issue?

nyh:
I don't think anyone else is working on this issue at the moment, so the answer is probably yes.
@wmitros has been working recently on improving Seastar's HTTP 1.1 support (fixing various parsing bugs, supporting chunked-encoding, etc.) but I don't know if he was planning to "expand" to HTTP 2 in the near future.

wmitros:
I was thinking about expanding to HTTP 2 at some unspecified time in the future, but I don't think there's anything that should be preventing you from taking it up right now @jewertow

jewertow:
Thank you for the answer. I need few weeks to investigate and implement this feature.

dimaqq:
I wonder if Jacek is still working on this or if he gave up 😅

dimaqq:
Poke

jewertow:
No, I didn't work on that, because the project for which it was needed, used the proxygen library to handle http/2 and didn't use seastar at all. I'm sorry, I shouldn't have declared that I will work on it without being sure.

mykaul:
I'd skip and go directly to HTTP/3.

nyh:
> I'd skip and go directly to HTTP/3.

I agree that implementing https://datatracker.ietf.org/doc/html/rfc9000 and https://datatracker.ietf.org/doc/html/rfc9114 will be cool.

Just note that HTTP/3 is much newer and support for it in various languages (including my favorite playground, Python) is much less mature than support for HTTP/2, so it may be harder to write clients to the HTTP/3 server. But I'm sure that the client availability will improve in the next few years.

muthu90tech:
Hi @mykaul / @nyh , I would like to work on the HTTP/3 support in that case, would like to add incremental changes
I can repurpose the http2 frame parsing to http3 frame parsing: https://github.com/scylladb/seastar/pull/2886
do you recommend I work and commit to a dedicated branch in your repo? thank you

[...]

PR #2046, opened 2024-01-15
http_v2: add enum and trie support

sqddk:
Unfortunately, during the usage, I have discovered that the current HTTP framework still has some issues.
Next, I hope to develop an HTTP_v2 version, providing users with better support for the HTTP protocol framework. What has been committed so far is a small part of the foundational work. I believe splitting it into multiple commits would be beneficial for code review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -770,7 +770,17 @@ add_library (seastar
   src/util/tmp_file.cc
   src/util/short_streams.cc
   src/websocket/server.cc
-  )
+
+  include/seastar/http_v2/request.hh
+  include/seastar/http_v2/response.hh
+  include/seastar/http_v2/common.hh
+  include/seastar/http_v2/parse.hh
+  include/seastar/util/trie.hh
+  src/http_v2/common.cc
+  src/http_v2/request.cc
+  src/http_v2/response.cc
+  src/http_v2/parse.cc
+)
 
 add_library (Seastar::seastar ALIAS seastar)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -773,10 +773,10 @@ add_library (seastar
 
   include/seastar/http_v2/request.hh
   include/seastar/http_v2/response.hh
-  include/seastar/http_v2/common.hh
+  include/seastar/http_v2/enum.hh
   include/seastar/http_v2/parse.hh
   include/seastar/util/trie.hh
-  src/http_v2/common.cc
+  src/http_v2/enum.cc
   src/http_v2/request.cc
   src/http_v2/response.cc
   src/http_v2/parse.cc
--- a/include/seastar/http_v2/common.hh
+++ b/include/seastar/http_v2/common.hh
@@ -1,0 +1,59 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <seastar/core/sstring.hh>
+#include <seastar/util/trie.hh>
+#include <utility>
+
+namespace seastar::http_v2 {
+
+/**
+ * Enum Base Class
+ */
+class enum_object {
+    sstring name;
+public:
+    explicit enum_object(sstring name);
+    inline sstring& get_name();
+};
+
+/**
+ * http method enums
+ */
+class method : public enum_object {
+private:
+    explicit method(const sstring& name);
+public:
+    const static method GET, HEAD, POST, PUT, DELETE, CONNECT, OPTIONS, TRACE;
+    static method* find_by_name(sstring& name);
+};
+
+/**
+ * http protocol version enums
+ */
+class version : public enum_object {
+private:
+    explicit version(const sstring& name);
+public:
+    const static version HTTP1_0, HTTP1_1;
+    static version* find_by_name(sstring& name);
+};
+
+}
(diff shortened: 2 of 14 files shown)


nyh:
I see you closed this pull request, so maybe you didn't intend anyone to review it, but I have two general requests:

1. Please be more concrete about what "http v2" means. I din't understand if this is about HTTP 2 (https://en.wikipedia.org/wiki/HTTP/2) or just a new implementation of the http. It looks like the latter, but why? You say the current implementation has "some issues", but what are these? Do we have issues in the bug tracker? Can't these issues be solved without a rewrite? You'll need to explain all of this in the pull request.
2. If you want to start working on a big component, please start with creating an issue or design document or mailing list thread about what you are thinking about, and why. If you start with sending utility classes like trie, they are unlikely to go in - we have no idea why you want this, or why it needs to be in Seastar. To be honest, I'm quite familiar with both tries and HTTP server implementations - and I can't figure out why you might need a try and std::unordered_map or std::map does not suffice.
3. Continuing point 2, if you're working on a completely separate component which doesn't share code with anything else, please consider working on it in a separate branch (you can make it public, talk about it in the mailing list, etc., but still be separate) until it's ready. If you haven't finished this work, and certainly if you don't have a watertight design document, it is quite possible that a lot will change in your implementation before it's ready, and all these non-functional middle stages will not be useful for other Seastar users or developers.

sqddk:
@nyh I'm sorry, I didn't express it clearly. The reason I didn't elaborate on the issues with the existing HTTP implementation is that I haven't thoroughly read the source code of that part yet, and I cannot criticize many designs from a more macro and accurate perspective.
In my limited understanding, let me provide a few examples. Firstly, the request object is too cumbersome, containing many variables that could be omitted. For instance, client_address and _server_address can be obtained from the connected socket, and query_parameters can be acquired through method parsing. Moreover, it does not support streaming data read and write. To implement these new features, I find myself struggling with the difficult-to-maintain legacy code.
Another issue is that the current httpclient is too challenging to use. Why does it have such a peculiar design of obtaining the response in a callback function? I prefer getting a `future<response>` instead.
It seems like you are currently developing a gateway and facing challenges such as forwarding received requests to other services. Additionally, you need to consider the possibility of HTTP body sizes exceeding the available memory, which requires implementing streaming for body read and write. Seastar, in its current state, doesn't address these pain points. Therefore, I am exploring the development of a better HTTP framework that allows users to set routes, handle requests in callback functions, and provides more flexibility in obtaining, constructing, and manipulating request/response as well as connection instances at a lower level.
The choice of using a trie is motivated by the consideration of potential illegal requests/responses that might have a very long initial line, where a trie exhibits a clear advantage. Its maximum overhead is determined by the length of the longest string used as a key. In the case of HTTP methods enumeration, the longest string is typically OPTIONS. While it's possible to use a map, since the trie has already been developed, there's a rationale for leveraging it instead of reinventing the wheel with a map.
The decision to close the pull request was driven by my realization of the points you mentioned. I acknowledge that the portion I wrote is of little benefit to others until a truly usable framework is developed. It makes more sense to submit a pull request after HTTPv2 has matured to a certain extent.

nyh:
> @nyh I'm sorry, I didn't express it clearly. The reason I didn't elaborate on the issues with the existing HTTP implementation is that I haven't thoroughly read the source code of that part yet, and I cannot criticize many designs from a more macro and accurate perspective.

I can't say I'm particularly fond of the existing HTTP server design, but to completely replace it with a new design, it's not enough to write something else that looks like correct code, you need to convince the other developers that:
1. Your design is significantly better than the old one in some important way - more usable, more performant, more featureful, or something, that makes the hassle of throwing the old one and using a new one worth it.
2. Your implementation cannot break existing applications - it can't have fewer features, and you should supply a wrapper that will allow an application using the old API (even if you think it's a bad API) to continue working.
3. The existing HTTP server is fairly trusted, and has been used quite heavily and is known to work correct in the use cases it was used. You will need to demonstrate using tests that your implementation is trustworthy as well. I don't know how good our unit/regression tests are for the HTTP servers, if they are not good, you'll need to write new ones.

Please think twice if you really need to do the entire HTTP server from scratch, or you can make **incremental** changes to the existing HTTP server to improve things incrementally, fixing small things that bother you, without a one-off complete rewrite.

By the way, something which you can do that will likely convince us to pick up your implementation is to implement HTTP/2 (https://github.com/scylladb/seastar/issues/734) - this something the current HTTP server implementation lacks, so if you create an implementation that supports it, it will be a big improvement.

> In my limited understanding, let me provide a few examples. Firstly, the request object is too cumbersome, containing many variables that could be omitted. For instance, client_address and _server_address can be obtained from the connected socket, and query_parameters can be acquired through method parsing. Moreover, it does not support streaming data read and write.

It does support streaming write, although in somewhat obscure ways. I think it also supports streaming read of the request, but I don't remember and and didn't check now.

> To implement these new features, I find myself struggling with the difficult-to-maintain legacy code. Another issue is that the current httpclient is too challenging to use. Why does it have such a peculiar design of obtaining the response in a callback function? I prefer getting a `future<response>` instead.

To be honest, I don't remember. I didn't design the current server, nor was involved in its design, and when I used it quite heavily in the ScyllaDB's "Alternator" system (a DynamoDB-compatible API, based on HTTP) I remember I had some struggles but at the end, got everything to work. We could have a more informed discussion if you wrote down in some design document or issue or mailing list thread a more detailed argument, with examples, of what the current callback looks like vs what a future would look like.

Please note that the whole idea of "streaming write" (which the current HTTP server) **does** support) is that you can't prepare the entire "response" up-front and return it: The response object needs to contain some sort of function or something which knows how to write the full body into the output socket.

> I am currently developing a gateway and facing challenges such as forwarding received requests to other services. Additionally, I need to consider the possibility of HTTP body sizes exceeding the available memory, which requires implementing streaming for body read and write. Seastar, in its current state, doesn't address these pain points. Therefore, I am exploring the development of a better HTTP framework that allows users to set routes, handle requests in callback functions, and provides more flexibility in obtaining, constructing, and manipulating request/response as well as connection instances at a lower level. The choice of using a trie is motivated by the consideration of potential illegal requests/responses that might have a very long initial line, where a trie exhibits a clear advantage. Its maximum overhead is determined by the length of the longest string used as a key. In the case of HTTP methods enumeration, the longest string is typically OPTIONS. While it's possible to use a map, since the trie has already been developed, why not use it? The decision to close the pull request was driven by my realization of the points you mentioned. I acknowledge that the portion I wrote is of little benefit to others until a truly usable framework is developed. It makes more sense to submit a pull request after HTTPv2 has matured to a certain extent.

Please also note that although if you come up with a great Seastar HTTP server replacement, it would be great to contribute it back to the Seastar project, you **don't have to** - you can maintain your HTTP server on top of Seastar without it being _inside_ Seastar. If there are low-level changes you need to make to Seastar, you can contribute those, but the entire HTTP server implementation can live outside Seastar until the day it is ready to be contributed to Seastar as a ready component.

PR #2048, opened 2024-01-20
net/posix: include used header

tchaikov:
in 8b9ae36b, we started to use `std::unordered_set` in `include/seastar/net/posix-stack.hh`, but we failed to include the used header. this breaks the `checkheaders` check.

so, in this change, let's include the userd header.


tchaikov:
@xemul hi Pavel, could you please help review this change?

PR #2051, opened 2024-01-22
net/posix: add a way to change file permissions of unix domain socket

margdoc:
Currently, Unix domain sockets are created with default permissions 755.
Add a new attribute to the listen_options struct to enable changing these permissions.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1651,6 +1651,18 @@ reactor::posix_listen(socket_address sa, listen_options opts) {
 
     try {
         fd.bind(sa.u.sa, sa.length());
+
+        if (sa.is_af_unix() && opts.unix_domain_socket_permissions) {
+            // After bind the socket is created in the file system.
+            mode_t mode = static_cast<mode_t>(opts.unix_domain_socket_permissions.value());
+            auto result = ::chmod(sa.u.un.sun_path, mode);
+            if (result < 0) {
+                auto errno_copy = errno;
+                ::unlink(sa.u.un.sun_path);
+                throw std::system_error(errno_copy, std::system_category(), "chmod failed");
+            }
+        }
+
         fd.listen(opts.listen_backlog);
     } catch (const std::system_error& s) {
         throw std::system_error(s.code(), fmt::format("posix_listen failed for address {}", sa));
--- a/include/seastar/net/api.hh
+++ b/include/seastar/net/api.hh
@@ -35,6 +35,7 @@
 #include <seastar/net/packet.hh>
 #include <seastar/core/internal/api-level.hh>
 #include <seastar/core/temporary_buffer.hh>
+#include <seastar/core/file-types.hh>
 #include <seastar/core/iostream.hh>
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/program-options.hh>
@@ -394,6 +395,7 @@ struct listen_options {
     transport proto = transport::TCP;
     int listen_backlog = 100;
     unsigned fixed_cpu = 0u;
+    std::optional<file_permissions> unix_domain_socket_permissions;
     void set_fixed_cpu(unsigned cpu) {
         lba = server_socket::load_balancing_algorithm::fixed;
         fixed_cpu = cpu;


kbr-scylla:
Please ask someone familiar with this code to review it

src/core/reactor.cc
```diff
@@ -1651,6 +1651,13 @@ reactor::posix_listen(socket_address sa, listen_options opts) {
 
     try {
         fd.bind(sa.u.sa, sa.length());
+
+        if (sa.is_af_unix() && opts.unix_domain_socket_permissions) {
+            auto mode = static_cast<mode_t>(*opts.unix_domain_socket_permissions);
+            auto syscall_result = wrap_syscall<int>(::chmod(sa.u.un.sun_path, mode));
+            syscall_result.throw_if_error();
```
avikivity, line 1658: Should unlink on failure.

Add a comment explaining why ::chmod() doesn't cause a stall.

avikivity:
Maybe we should defer this to the application? After all, the posix API which we follow does too.

margdoc:
> Maybe we should defer this to the application? After all, the posix API which we follow does too.

Seastar calls `bind` and `listen` one by one. I wanted to call `chmod` before `listen` to not make a time window where  permission are other than the one we want. If this time window is not a problem, we can do it in the application.

margdoc:
> After all, the posix API which we follow does too.

Yes, but it also splits `socket`, `bind` and `listen` from each other. In `reactor::posix_listen` we call all of them. I think we should do it in seastar.

src/core/reactor.cc
margdoc: I found out about `fchmod` system call. Here it should be more suitable than `chmod` since we have a file descriptor for the socket. I will change the code to use `fchmod` and then unlinking on failure won't be necessary (`fchmod` will be called before `bind`, so there is nothing to unlink).

nyh:
The traditional way of controlling the permissions of new Unix-domain sockets, and also new files, isn't to create them first and then chmod (something which is vulnerable to race conditions) but rather to use **umask(2)** before creating the socket (i.e., bind).
The permissions you mentioned 755, is simply what is left when you mask the full permissions, with your system's umask which I guess is 022 (which is the default on many setups).

Can you please explain if there is any downside to using umask() instead of chmod()?
Can you please explain how you plan to avoid the race involved with creating a socket with the wrong permission and only then reducing it (this can be solved by having a separate directory to hold the socket, but I don't see this in the code).

src/core/reactor.cc
nyh: Very nice, I didn't know that fchmod **before** the bind was even possible, but I see multiple references to this trick online. This nicely solves the race condition of bind-and-then-chmod. I didn't know this works, and apparently it's a relatively new Linux trick - on non-Linux Unix-like systems you'll get an EINVAL (POSIX https://pubs.opengroup.org/onlinepubs/009696699/functions/fchmod.html says it's "unspecified"). 

Please note that umask still applies (I believe), so whatever permissions you'll try to set, actually 022 will be removed from it (your umask).

nyh: fchmod does work, but it *also* removes the umask bits.
The question is why is it bad? open(2) and creat(2) also lets you choose the "mode" bits on file creation - but neither of them takes your given mode as-is, both subtract umask from it. So it is expected by users.

Note that if you use chmod() after the bind to just **increase** permissions, there is no security risk. So if you really need to override the umask and readd these bits, you can. I just don't think anybody needs to do this. What do you plan to use this feature for?

Finally, please note that umask() is shared by all threads of a process. So changing it is a rather nasty thing to do, and should be done by the application once on initialization - not before some bind() call. I wasn't suggesting that you should change umask in each call - I was thinking more along the line of accepting the current umask(), or asking the user to change it once in their application.

margdoc:
v2:
* use `::umask` instead of `::chmod`

avikivity:
> > Maybe we should defer this to the application? After all, the posix API which we follow does too.
> 
> Seastar calls `bind` and `listen` one by one. I wanted to call `chmod` before `listen` to not make a time window where permission are other than the one we want. If this time window is not a problem, we can do it in the application.

Good point. And it makes sense to split the calls - one can do additional configuration.

So I accept the change is necessary.

nyh:
> v2:
> 
>     * use `::umask` instead of `::chmod`

Sorry for not being clear enough, but my idea was not to call umask() every time you want to call bind(). umask() affects the entire process (all threads) so should not be called in a multi-threaded program without some sort of nasty lock. My idea was more of assuming that the application would set umask() on its own, once, during the initialization, and this would control the permissions of new sockets - as well as new files, instead of the default 0755 which you get when the umask is 022. E.g., if you use umask of 077, all new sockets will be writable and readable only by the owner - not others. Then you don't need to pass the permissions to every socket creations.

But maybe I'm missing the intended use case - why did you want to add this permissions option? Do you really intend to pass different permissions on different calls?

avikivity:
Now I have a feeling we'll need optional chown (to change the group ownership)

nyh:
> Seastar calls `bind` and `listen` one by one. I wanted to call `chmod` before `listen` to not make a time window where permission are other than the one we want.

If this is true, it means the race condition that worried me does not exist - the attacker cannot actually open() the socket file after it's created by bind(), and can only connect() to it and that requires listen() - which you only do after the chmod().

The open(2) manual page suggests that open() on a unix domain socket will result in ENXIO, and connect(2) suggests that it will result in a ECONNREFUSED if called before the listen().

So maybe your chmod code was correct after all! So sorry for misleading you.

nyh:
> Now I have a feeling we'll need optional chown (to change the group ownership)

As for umask, I don't understand the motivation of controlling the group or mode bits of each individual socket - as opposed to just creating all new sockets (as well as regular files, probably) with the same permissions and group. We should have a motivation in mind before adding a lot of complicated options that have no corresponding parallels in the familiar Linux APIs.

avikivity:
> > Now I have a feeling we'll need optional chown (to change the group ownership)
> 
> As for umask, I don't understand the motivation of controlling the group or mode bits of each individual socket - as opposed to just creating all new sockets (as well as regular files, probably) with the same permissions and group. We should have a motivation in mind before adding a lot of complicated options that have no corresponding parallels in the familiar Linux APIs.

The parallel is bind()/chmod()/chown()/listen(), which we can't do because bind() and listen() are atomic in Seastar.

The motivating use case is a privileged socket for ScyllaDB administrators. We want the socket to be available to the scyllaadm group, but not others.

margdoc:
> Now I have a feeling we'll need optional chown (to change the group ownership)

Do you mean to add the next option to `listen_options` with `gid_t`?
Then how in the scylla we should get gid of scyllaadm group? By `getgrnam`?

margdoc:
@avikivity 
Can we just do `bind()/chmod()/chown()/listen()` or do we have to do sth else?

avikivity:
Yes, add things like std::optional<gid_t> and std:optional<uid_t> (the latter will be useless since only root can change the uid, but better have it for completeness).

Doesn't have to be in this series.

The whole thing mimics posix_spawn() (which needs to set all options directly) vs fork/change options via syscalls/exec.

margdoc:
v3:
* move back to `chmod`
* add unlink on failure

src/core/reactor.cc
```diff
@@ -1651,6 +1651,17 @@ reactor::posix_listen(socket_address sa, listen_options opts) {
 
     try {
         fd.bind(sa.u.sa, sa.length());
+
+        if (sa.is_af_unix() && opts.unix_domain_socket_permissions) {
+            // After bind the socket is created in the file system.
+            mode_t mode = static_cast<mode_t>(opts.unix_domain_socket_permissions.value());
+            auto result = ::chmod(sa.u.un.sun_path, mode);
+            if (result < 0) {
+                ::unlink(sa.u.un.sun_path);
```
avikivity, line 1660: If unlink also fails for some reason (shouldn't but could), errno is overwritten.

Need to capture errno before the unlink.

margdoc:
v4:
* in case of failure, use errno from `chmod` instead of `unlink`

kbr-scylla:
@avikivity ping can this be merged?

Issue #1772, opened 2023-08-08
seastar does not build with clang-17 + libstdc++ + C++23


tchaikov:
```shell
cmake -Bbuild/cxx23 -GNinja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_STANDARD=23
```
and the build fails with errors like
```
FAILED: CMakeFiles/seastar.dir/src/core/reactor.cc.o 
/home/kefu/.local/bin/clang++ -DBOOST_NO_CXX98_FUNCTION_BASE -DFMT_SHARED -DSEASTAR_API_LEVEL=7 -DSEASTAR_DEBUG -DSEASTAR_DEBUG_SHARED_PTR -DSEASTAR_DEFAULT_ALLOCATOR -DSEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_NUMA -DSEASTAR_HAVE_URING -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_SHUFFLE_TASK_QUEUE -DSEASTAR_THREAD_STACK_GUARDS -DSEASTAR_TYPE_ERASE_MORE -I/home/kefu/dev/seastar/include -I/home/kefu/dev/seastar/build/debug/gen/include -I/home/kefu/dev/seastar/src -g -std=gnu++23 -U_FORTIFY_SOURCE -DSEASTAR_SSTRING -Wno-error=unused-result "-Wno-error=#warnings" -fstack-clash-protection -UNDEBUG -Wall -Werror -Wimplicit-fallthrough -Wno-array-bounds -Wno-error=deprecated-declarations -fvisibility=hidden -gz -fsanitize=address -fsanitize=undefined -fno-sanitize=vptr -MD -MT CMakeFiles/seastar.dir/src/core/reactor.cc.o -MF CMakeFiles/seastar.dir/src/core/reactor.cc.o.d -o CMakeFiles/seastar.dir/src/core/reactor.cc.o -c /home/kefu/dev/seastar/src/core/reactor.cc
In file included from /home/kefu/dev/seastar/src/core/reactor.cc:29:
[...]

PR #2053, opened 2024-01-24
Update supported C++ standards to C++23 and C++20 (dropping C++17)

avikivity:
C++23 has passed all the ballots [1]. We can therefore support it now formally. This means that C++17 is no longer supported, and C++20 is supported until C++26 is released.

The primary benefit of C++23 support is that we can now use coroutines in core Seastar code, as all supported C++ versions have them.

[1] https://lists.isocpp.org/std-proposals/2024/01/8880.php

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ workflows:
           matrix:
             parameters:
               compiler: ["clang++-17", "g++-13"]
-              standard: ["20", "17"]
+              standard: ["23", "20"]
               mode: ["dev", "debug", "release"]
       # only build this combination with dpdk enabled, so we don't double
       # the size of the test matrix, and can at least test the build with
@@ -80,7 +80,7 @@ workflows:
           matrix:
             parameters:
               compiler: ["clang++-17"]
-              standard: ["20"]
+              standard: ["23"]
               mode: ["release"]
               with_dpdk: [ "enable-dpdk" ]
       # only build this combination with C++20 moduels enabled, so we don't double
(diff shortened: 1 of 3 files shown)


avikivity:
Probably https://github.com/llvm/llvm-project/issues/17042

tchaikov (approved):
yeah. it's around the corner. strictly speaking it's not out yet. see https://www.iso.org/standard/83626.html

tchaikov:
now we are blocked by #1772

tchaikov:
before the bug on Clang-17 is addressed, can we temporarily disable the test for the combination of of clang and C++23?

avikivity:
I saw one bug was fixed, but then using a newer libstdc++ triggered another.

tchaikov:
> I saw one bug was fixed, but then using a newer libstdc++ triggered another.

FWIW, clang-18 (230c13d59d0843c3b738920b85c341cc78a61fa9) + libstdc++ (from 13.2.1) + C++23 works fine.

tchaikov:
now that clang-18 has the commit of 128b3b61fe6768c724975fd1df2be0abec848cf6, probably we can bump up the clang in our CI build image? see #2058

tchaikov (commented):
to address the FTBFS with clang-17, after redo the image with the change of #2058, we might need to s/clang++-17/clang++-18/ in this change .

tchaikov:
> to address the FTBFS with clang-17, after redo the image with the change of #2058, we might need to s/clang++-17/clang++-18/ in this change .

i applied the proposed change, and posted #2091.

PR #2054, opened 2024-01-24
circular_buffer_fixed_capacity: use std::uninitialized_move() instead of open-coding

avikivity:
The code even contains a comment about it, so standardize the code.


tchaikov (approved):
lgtm

PR #2056, opened 2024-01-24
future: remove tuples from get0_return_type

avikivity:
get0_return_type uses some complicated type manipulations to translate void types; these date from the days when we supported variadic futures.

We don't any more, so simplify.

### DIFF
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -259,15 +259,15 @@ template <typename T>
 struct get0_return_type;
 
 template <>
-struct get0_return_type<std::tuple<>> {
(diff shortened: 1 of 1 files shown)


nyh (approved):
Looks good, but can you please remind me why we even have get0() any more now that we don't have variadic futures? Or, if we need to keep get0() for backward compatibility, why is it not 100% identical to get() and is slightly different?

avikivity:
> Looks good, but can you please remind me why we even have get0() any more now that we don't have variadic futures? Or, if we need to keep get0() for backward compatibility, why is it not 100% identical to get() and is slightly different?

I think it makes sense to deprecate get0().

PR #2058, opened 2024-01-26
docker: install clang-tools-18

tchaikov:
to address the build failure caused by
https://github.com/llvm/llvm-project/issues/59827, and to test the build with C++23, we need to use clang-18. which will be released in March 2023.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
       - checkout
       - run: git submodule sync
       - run: git submodule update --init
-      - run: echo 'docker run --pids-limit -1 --security-opt seccomp=unconfined --network host --user "$(id -u):$(id -g)" --rm -v $PWD:$PWD -w $PWD  docker.io/scylladb/seastar-toolchain:2023-12-28 "$@"' > run; chmod +x run
+      - run: echo 'docker run --pids-limit -1 --security-opt seccomp=unconfined --network host --user "$(id -u):$(id -g)" --rm -v $PWD:$PWD -w $PWD  docker.io/scylladb/seastar-toolchain:2024-02-01 "$@"' > run; chmod +x run
       - when:
           condition:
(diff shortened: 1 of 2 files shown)


tchaikov:
tested using 
```console
$ podman build -t seastar-dev -f ./docker/dev/Dockerfile  .
...
STEP 10/10: CMD /bin/bash
COMMIT seastar-dev
--> c4401e6db9d7
Successfully tagged localhost/seastar-dev:latest
c4401e6db9d7057e6ab910c49a1a2b2bdee5239953c012748cb3406a898739c8
```

tchaikov:
@avikivity could you help review this change?

tchaikov:
@avikivity could you help review this change?

avikivity:
Why install an unreleased compiler?  Is the fix not backported to clang 17?

tchaikov:
> Why install an unreleased compiler? Is the fix not backported to clang 17?

no, clang-17 won't get any backport anymore. i prepared the backport on f38, f39 and rawhide, see https://src.fedoraproject.org/rpms/clang/pull-request/226 and https://src.fedoraproject.org/rpms/clang/pull-request/225. but i am too lazy to file a ticket on BTS and on launchpad.

tchaikov:
v2:

- rebased against master HEAD

tchaikov:
@avikivity hi Avi, mind taking another look?

tchaikov:
@avikivity ping?

avikivity:
Is this the same build script as was used to generate the image referenced in .circleci? I don't remember the gpg stuff.

tchaikov:
> Is this the same build script as was used to generate the image referenced in .circleci? I don't remember the gpg stuff.

@avikivity it is. please compare it with #2070 . without installing GnuPG, it's impossible to import LLVM's keyring into the system, and we would not be able to install their debian packages without bypassing the signature check.

PR #2059, opened 2024-01-28
future: deprecate future::get0()

avikivity:
In the days when futures carried tuples, it was useful to get the first/only element. But it's now a synonym for get(), so deprecate it.

### DIFF
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -1355,6 +1355,7 @@ public:
     ///
     /// Equivalent to: \c std::get<0>(f.get()).
     using get0_return_type = typename future_state::get0_return_type;
+    [[deprecated("Use get() instead")]]
     get0_return_type get0() {
         return (get0_return_type)get();
     }


nyh (changes requested):
I'm in favor of this deprecation, but I wonder if it wouldn't cause unnecessary headaches to Seastar projects which will now need to change all the calls to get0() or face hundreds of warning messages.
Could we make "get0()" an exact synonym of get()" - with less code (no separate return type, etc.) and then we won't mind just leaving it there for years?

The reason I'm asking for "Request changes" is that I think you forgot to through Seastar itself and replace all calls to get0() by get() - we have many of those in various helper classes, in example applications, and so on. Didn't you get a lot of deprecation warnings when building Seastar with this patch?

nyh:
> Didn't you get a lot of deprecation warnings when building Seastar with this patch?

I see in the CI that you did :-)

tchaikov:
@avikivity now that #2069 has been merged, could you rebase this change? i think it's should be ready now. i tested it locally. the tree built just fine after applying this change.

tchaikov:
> @avikivity now that #2069 has been merged, could you rebase this change? i think it's should be ready now. i tested it locally. the tree built just fine after applying this change.

i took the liberty to rebase this change at #2188

tchaikov:
closing, since #2188 has been merged.

PR #2063, opened 2024-01-30
build: do not expose -Wno-error=#warnings

tchaikov:
the option of `-Wno-error=#warnings` is added to address the FTBFS when the compiler runs into `#warning` preprocessor, and fails because it takes warning as errors. but we only use `#warning` in the internal header/source files. we even filter out `-Wno-error=#warnings` in 83339edb. this is another proof that this compile option is not necessary to be a public of the public interface of Seastar.

but we still expose this option when the Seastar application is still consuming Seastar via CMake. this causes discrepancies between the public compiling options of Seastar applications.

so, in this change, we

* mark `-Wno-error=#warnings` one of the private compile option instead.
* revert 83339edb

with this change, we

* reduce the differences between the compiling options exposed by seastar.pc and CMake config module,
* and avoid exposing unnecessary compiling options to Seastar applications.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1045,7 +1045,7 @@ endif ()
 check_cxx_compiler_flag ("-Wno-error=#warnings" ErrorWarnings_FOUND)
 if (ErrorWarnings_FOUND)
   target_compile_options (seastar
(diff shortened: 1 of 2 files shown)


tchaikov:

PR #2064, opened 2024-01-30
util/spinlock: use #warning instead of #warn

tchaikov:
`#warning` is a preprocessor macro in C/C++, while `#warn` is not. the reason we haven't run into the build failure caused by this is that we are not building on an architecture other than amd64, i386, PowerPC and aarch64 yet.

### DIFF
--- a/include/seastar/util/spinlock.hh
+++ b/include/seastar/util/spinlock.hh
@@ -71,7 +71,7 @@ inline void cpu_relax() {
 
 [[gnu::always_inline]]
 inline void cpu_relax() {}
-#warn "Using an empty cpu_relax() for this architecture"
+#warning "Using an empty cpu_relax() for this architecture"
 
 #endif
 


vladzcloudius (changes requested):
Looks good but there is another place (exceptions.hh) where we use #warn - please, patch it too.
Also, please, clarify what actually happened when you apparently tried to compile the original code on a platform different from those that you mentioned.
My guess is that you've got an compilation error of some kind. If "yes", please, provide more details on that either in the patch description or in the GH issue that this PR will close (and then you'd reference it from the patch description).

tchaikov:
> Looks good but there is another place (exceptions.hh) where we use #warn - please, patch it too. 

there is only one `exceptions.hh` in the tree, which is https://github.com/scylladb/seastar/blob/master/include/seastar/util/exceptions.hh. i searched it for `#warn` , no luck.

```console
$ git grep -w '#warning' && echo found
include/seastar/core/dpdk_rte.hh:#warning "CONFIG_RTE_MBUF_REFCNT_ATOMIC should be disabled in DPDK's " \
include/seastar/util/spinlock.hh:#warning "Using an empty cpu_relax() for this architecture"
src/core/thread.cc:    #warning "Backtracing from seastar threads may be broken"
found
$ git grep -w '#warn' && echo found
$
```

> Also, please, clarify what actually happened when you apparently tried to compile the original code on a platform different from those that you mentioned.

i explained why the FTBFS didn't surface before. but i didn't try to compile the code on an architecture which could trigger the FTBFS. the reason why i noticed it was #2063. i was auditing the code to see if any of the public header files is using `#warning`.

> My guess is that you've got a compilation error of some kind. If "yes", please, provide more details on that either in the patch description or in the GH issue that this PR will close (and then you'd reference it from the patch description).

the answer is "no".

vladzcloudius:
> > Looks good but there is another place (exceptions.hh) where we use #warn - please, patch it too.
> 
> there is only one `exceptions.hh` in the tree, which is https://github.com/scylladb/seastar/blob/master/include/seastar/util/exceptions.hh. i searched it for `#warn` , no luck. could you point me to the place?

My bad. I was referring an `exceptions.hh` from a `scylla` project.
Please, disregard then.

vladzcloudius (approved):

tchaikov:
> > > Looks good but there is another place (exceptions.hh) where we use #warn - please, patch it too.
> > 
> > 
> > there is only one `exceptions.hh` in the tree, which is https://github.com/scylladb/seastar/blob/master/include/seastar/util/exceptions.hh. i searched it for `#warn` , no luck. could you point me to the place?
> 
> My bad. I was referring an `exceptions.hh` from a `scylla` project. Please, disregard then.

Vlad, thank you for your review and approval! created https://github.com/scylladb/scylladb/pull/17074 to address it.

PR #2065, opened 2024-01-30
core/future: add as_ready_future utility

rockwotj:
seastar::make_ready_future requires a template parameter when
constructing being used due to the default of `void` and supporting
in place construction. Common usage (at least in Redpanda) is
to create a fully formed future from an already constructed struct.

Example with `seastar::make_ready_future`:

```
seastar::future<ss::sstring> foo() {
  ss::sstring str = compute_foo();
  return seastar::make_ready_future<ss::sstring>(std::move(str));
}
```

Using `seastar::as_ready_future`:

```
seastar::future<ss::sstring> foo() {
  ss::sstring str = compute_foo();
  return seastar::as_ready_future(std::move(str));
}
```

### DIFF
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -166,6 +166,15 @@ struct future_state_base;
 template <typename T = void, typename... A>
 future<T> make_ready_future(A&&... value) noexcept;
 
+
+/// \brief Returns a ready \ref future that is already resolved.
+template<typename T>
+inline
+future<std::remove_cv_t<std::remove_reference_t<T>>> as_ready_future(T&& v) noexcept {
+    return make_ready_future<std::remove_cv_t<std::remove_reference_t<T>>>(
+        std::forward<T>(v));
+}
+
 /// \brief Creates a \ref future in an available, failed state.
 ///
 /// Creates a \ref future object that is already resolved in a failed


include/seastar/util/later.hh
```diff
@@ -36,6 +36,14 @@ future<> now() {
     return make_ready_future<>();
 }
 
+/// \brief Returns a ready future that is already resolved.
+template<typename T>
+inline
+future<std::remove_cv_t<std::remove_reference_t<T>>> now(T&& v) noexcept {
+    return make_ready_future<std::remove_cv_t<std::remove_reference_t<T>>>(
```
rockwotj, line 43: I would use std::remove_cvref_t but that requires C++20

include/seastar/util/later.hh
```diff
@@ -36,6 +36,14 @@ future<> now() {
     return make_ready_future<>();
 }
 
+/// \brief Returns a ready future that is already resolved.
+template<typename T>
+inline
+future<std::remove_cv_t<std::remove_reference_t<T>>> now(T&& v) noexcept {
```
BenPope, line 42: If the type isn't movable it could still throw.

Might be worth having an `now<T>(std::in_place, Args&&... args)`?

rockwotj: Can you elaborate? I would think you should just use `make_ready_future` if you want in place construction.

travisdowns: I don't think there's much value having `noexcept` on a small standalone function with visible body, so yeah it probably makes sense to remove it. The alternative, I guess, is metaprogramming to try to propagate the no-exceptness of the invoked constructor, seems annoying.

travisdowns: > If the type isn't movable it could still throw.

Even _if_ is moveable it could throw, right? Also even if a type is "moveable" we will only invoke the move constructor here if we are passed an rvalue, which will probably often not be the case (sometimes in error, i.e., `std::move` was not called where it should have been).

rockwotj: Removed the `noexcept` clause.

BenPope: > Can you elaborate? I would think you should just use `make_ready_future` if you want in place construction. Are you suggesting to remove `noexcept` then?

Basically, currently you have to write:
```c++
ss::make_ready_future<T>(T{args...});
```
Honestly, it's a minor point without much of a current demand.

> Removed the `noexcept` clause, thanks all.

I wasn't necessarily arguing against `noexcept` or it's utility, just noting that a copy is possible, and therefore allocs. I don't know if compilers use it as an optimisation, but it might be a `constexpr` limitation.

rockwotj: > Basically, currently you have to write:

This seems to work for me:

```c++
struct foo {
    std::string bar;
    size_t qux;
};

int main() {
    auto fut = seastar::make_ready_future<foo>(foo{"nice", 42});
    fut = seastar::now(foo{"nice", 42});
    return 1;
}
```


https://godbolt.org/z/qTsTjfqTj

BenPope: That's what I said? `std::string` is movable.

rockwotj: Ah I was misunderstanding. Yeah I think that at least the move constructor is required:

https://github.com/scylladb/seastar/blob/938a4e603ea438d2785eaf6ba1778e6728548e15/include/seastar/core/future.hh#L565-L568

BenPope: https://clang.llvm.org/extra/clang-tidy/checks/performance/noexcept-move-constructor.html

And `ss::make_ready_future` is `noexcept`. So I say leave `noexcewpt` in, maybe at the concept check, but I'm like 2/10 on it.

rockwotj: I put `noexcept` back but I not going to mess with the `SEASTAR_CONCEPT` stuff for now.

avikivity:
now() is a poor name for this. Nobody can guess that the return value is a future.

avikivity:
also, it can live in future.hh, it's too simple to deserve its own header file.

rockwotj:
> now() is a poor name for this.

I renamed it to `future_of`. Alternatives I considered are: `ready_future_of`, `immediate_future`, `make_future`, or making it a static method on future: `seastar::future::ready`. Please advise if there is another name preferred.

> also, it can live in future.hh, it's too simple to deserve its own header file.

Done.

include/seastar/util/later.hh
travisdowns: You should ignore my comments above. However I have new better comments. 

I had missed the fact that `ss::make_ready_future` is itself `noexcept` and this is possible because ss::future has a noexecpt constructor (regardless what constructor is called, whether it throws or not), because exceptions during construction are caught and transformed into an exceptional future: see `future_state::future_state(ready_future_marker, A&&... a)` implementation. 

So @BenPope I don't actually see how @rockwotj 's method can throw? In particular, I don't see where an allocation can occur and nothing else that I can see would throw.

travisdowns: (so think this should definitely have an unqualified `noexcept` as it does)

nyh:
I agree that a feature like this will be useful, and every time I type the long `make_ready_future<some_long_type_name>(std::move(something))` I feel like I'm programming in Java :-)

But in the best bikeshedding tradition, I want to talk about the name of the function.
If I understand correctly, the following two lines have the same "goal":
```
return seastar::make_ready_future<ss::sstring>(std::move(str));
return seastar::future_of(std::move(str));
```
Just that the way they work is the first one **constructs** ("make"s) a future with the parameter str given to the constructor (in this case, the move constructor), and the second one can't construct a new object, it just moves it into the future. But why did we lose the word "ready" in the function name? It's a ready future, just like in make_ready_future, and I think this word should remain in its name.

So how about calling the new function `ready_future` - same as `make_read_future` without the "make" to emphasize it doesn't construct the object?

It would be even nicer, but I don't know if that's even possible, if for the original name of `make_ready_future` to be overloaded to also also work in this case. I.e., when make_ready_future receives an rvalue reference to a type T, it could try to use the move constructor to create a future<T> - with this T being automatically deduced. I don't know how easy this is to do, if at all possible. Wouldn't it be terrific if we could do this? (again, I don't know if it's possible)
```
// this overload needs to create an sstring from "hello" so the the type sstring needs to be explicit:
return seastar::make_ready_future<sstring>("hello"); 

// this overload gets an sstring&&, so it deduces automatically that it returns a future<sstring>
return seastar::make_ready_future(std::move(str));
```

avikivity:
It's probably quite ugly, because in one case the arguments are independent from the the return type, and in the other not. So there would be some work to disambiguate it.

Name suggestions: to_ready_future, as_ready_future.

rockwotj:
Updated to `as_ready_future`.

nyh:
> It's probably quite ugly, because in one case the arguments are independent from the the return type, and in the other not. So there would be some work to disambiguate it.

Ugly to implement, or ugly to use?

> Name suggestions: to_ready_future, as_ready_future.

So why not just "ready_future"? Three less characters to type? :-)

travisdowns:
Well since the bike shed doors have officially been opened, my 2c is that conciseness over-self documenting names is increasingly called for as you move from "rarely called" to "frequently called".

`make_ready_future` is near the top of the  "frequently called" leaderboards (though `.then` has about 2x as many calls), so I my personal preference is a short name. Anyone spending a few minutes in the code base will just has this idiom as muscle memory.

Furthermore, `seastar::now()` already exists (and I love it) in later.hh: why should `void` futures have all the fun?

Since a bikeshed isn't complete without adding another suggestion to the pile: It it needs to express better the "ready" asked, how about `seastar::ready()`?

nyh:
`seastar::ready()` looks good to me, but I don't know if it's not overdoing things (we can also call it "r" to make it even shorter ;-)) - I'll let others comment.
By the way, it would be nice for `ready()` (or whatever we'll call it) - without an argument - to work as a shortcut for `make_ready_future<>()` . The latter is used often (I counted 1000 instances in the ScyllaDB source code ;-)) and maybe it makes sense to make it shorter.

rockwotj:
> By the way, it would be nice for `ready()` (or whatever we'll call it) - without an argument - to work as a shortcut for `make_ready_future<>()` . The latter is used often (I counted 1000 instances in the ScyllaDB source code ;-)) and maybe it makes sense to make it shorter.

Thats what `seastar::now` is in util/later.hh

rockwotj:
I think users of seastar can alias this to a shorter name if desired, we will probably do this in Redpanda. I think erroring on side of readability is the right decision.

PR #2067, opened 2024-02-01
io_tester: Allow running on non-XFS fs

xemul:
Currently io-tester just exits in this case, but it's a testing app, it doesn't need to be smarter than whoever uses it, so a WARNING should be enough.

### DIFF
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -968,7 +968,7 @@ int main(int ac, char** av) {
             if (*st_type == directory_entry_type::directory) {
                 auto fs = file_system_at(storage).get0();
                 if (fs != fs_type::xfs) {
-                    throw std::runtime_error(format("This is a performance test. {} is not on XFS", storage));
+                    std::cout << "WARNING!!! This is a performance test. " << storage << " is not on XFS" << std::endl;
                 }
             }
 


apps/io_tester/io_tester.cc
```diff
@@ -968,7 +968,7 @@ int main(int ac, char** av) {
             if (*st_type == directory_entry_type::directory) {
                 auto fs = file_system_at(storage).get0();
                 if (fs != fs_type::xfs) {
-                    throw std::runtime_error(format("This is a performance test. {} is not on XFS", storage));
+                    std::cout << "WARNING!!! This is a performance test. " << storage << " is not on XFS" << std::endl;
```
tchaikov, line 971: does it make sense to send this message to `std::cerr` ? and to use
```c++
fmt::print(cerr, "This is a performance test. {} is not on XFS\n", storage);
```
less quotes this way.

tchaikov: bonus, could even use something like
```c++
fmt::print(cerr, fmt::emphasis::bold | fg(fmt::color::red), "WARNING!!!");
fmt::print(cerr, "This is a performance test. {} is not on XFS\n", storage);
```

xemul: The whole io-tester uses old-style cout<< printing. I think it deserves its own PR to convert everything into fmt

tchaikov (approved):
lgtm

apps/io_tester/io_tester.cc
xemul: I've my /home ext4-formatted and when I want to quickly validate some io-related change, I run a short and stupid io-tester work. And I'm sick and tired of need to work this check around :(

tchaikov: i see. btrfs formatted here. so i am already in your shoes =)

avikivity:
Even a warning is unnecessary IMO.

PR #2068, opened 2024-02-01
io_tester: implement request_type::unlink

pwrobelse:
This change introduces request_type::unlink as well
as unlink_class_data to io_tester. It also extends
the creation of operations to be executed to recognize
the new request type.

The purpose of this change is to enable io_tester to
analyze the impact of unlink operations on read and
write operations.

unlink_class_data creates a given number of files
during startup. When requests are issued it calls
unlink on the created files.

Refs: scylladb#1299

### DIFF
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -35,7 +35,10 @@
 #include <seastar/core/io_intent.hh>
 #include <seastar/util/later.hh>
 #include <chrono>
+#include <optional>
+#include <utility>
 #include <unordered_set>
+#include <vector>
 #include <boost/range/irange.hpp>
 #include <boost/algorithm/string.hpp>
 
@@ -65,7 +68,7 @@ static auto random_seed = std::chrono::duration_cast<std::chrono::microseconds>(
 static thread_local std::default_random_engine random_generator(random_seed);
 
 class context;
-enum class request_type { seqread, seqwrite, randread, randwrite, append, cpu };
+enum class request_type { seqread, seqwrite, randread, randwrite, append, cpu, unlink };
 
 namespace std {
 
@@ -86,6 +89,39 @@ auto allocate_and_fill_buffer(size_t buffer_size) {
     memset(buffer.get(), fill(random_generator), buffer_size);
 
     return buffer;
+}
+
+future<std::pair<file, uint64_t>> create_and_fill_file(sstring name, uint64_t fsize, open_flags flags, file_open_options options) {
+    return open_file_dma(name, flags, options).then([fsize] (auto f) mutable {
+        return do_with(std::move(f), [fsize] (auto& f) {
+            return f.size().then([f, fsize] (uint64_t pre_truncate_size) mutable {
+                return f.truncate(fsize).then([f, fsize, pre_truncate_size] () mutable {
+                    if (pre_truncate_size >= fsize) {
+                        return make_ready_future<std::pair<file, uint64_t>>(std::pair{f, 0u});
+                    }
+
+                    const uint64_t buffer_size{256ul << 10};
+                    const uint64_t buffers_count{static_cast<uint64_t>(fsize / buffer_size) + 1u};
+                    const uint64_t last_buffer_id = (buffers_count - 1u);
+                    const uint64_t last_write_position = buffer_size * last_buffer_id;
+
+                    return do_with(boost::irange(UINT64_C(0), buffers_count), [f, buffer_size] (auto& buffers_range) mutable {
+                        return max_concurrent_for_each(buffers_range.begin(), buffers_range.end(), 64, [f, buffer_size] (auto buffer_id) mutable {
+                            auto source_buffer = allocate_and_fill_buffer(buffer_size);
+                            auto write_position = buffer_id * buffer_size;
+                            return do_with(std::move(source_buffer), [f, write_position, buffer_size] (const auto& buffer) mutable {
+                                return f.dma_write(write_position, buffer.get(), buffer_size).discard_result();
+                            });
+                        });
+                    }).then([f]() mutable {
+                        return f.flush();
+                    }).then([f, last_write_position]() {
+                        return make_ready_future<std::pair<file, uint64_t>>(std::pair{f, last_write_position});
+                    });
+                });
+            });
+        });
+    });
 }
 
 future<> busyloop_sleep(std::chrono::steady_clock::time_point until, std::chrono::steady_clock::time_point now) {
@@ -207,8 +243,12 @@ struct job_config {
     ::options options;
     // size of each individual file. Every class and every shard have its file, so in a normal
     // system with many shards we'll naturally have many files and that will push the data out
-    // of the disk's cache
+    // of the disk's cache. An exception to that rule is unlink_class_data, that creates files_count
+    // files with file_size/files_count.
     uint64_t file_size;
+    // the number of files to create and unlink by unlink_class_data per shard
+    // remaining operations utilize only one file per shard
+    std::optional<uint64_t> files_count;
     uint64_t offset_in_bdev;
     std::unique_ptr<class_data> gen_class_data();
 };
@@ -218,6 +258,10 @@ static bool keep_files = false;
 
 future<> maybe_remove_file(sstring fname) {
     return keep_files ? make_ready_future<>() : remove_file(fname);
+}
+
+future<> maybe_close_file(file& f) {
+    return f ? f.close() : make_ready_future<>();
 }
 
 class class_data {
@@ -359,15 +403,16 @@ public:
             return make_ready_future<>();
         }
     }
-    // Generate the test file for reads and writes alike. It is much simpler to just generate one file per job instead of expecting
-    // job dependencies between creators and consumers. So every job (a class in a shard) will have its own file and will operate
-    // this file differently depending on the type:
+    // Generate the test file(s) for reads and writes alike. It is much simpler to just generate one file per job instead of expecting
+    // job dependencies between creators and consumers. Removal of files is an exception - it creates multiple files during startup to
+    // unlink them. So every job (a class in a shard) will have its own file(s) and will operate differently depending on the type:
     //
     // sequential reads  : will read the file from pos = 0 onwards, back to 0 on EOF
     // sequential writes : will write the file from pos = 0 onwards, back to 0 on EOF
     // random reads      : will read the file at random positions, between 0 and EOF
     // random writes     : will overwrite the file at a random position, between 0 and EOF
     // append            : will write to the file from pos = EOF onwards, always appending to the end.
+    // unlink            : will unlink files created at the beginning of the execution
     // cpu               : CPU-only load, file is not created.
     future<> start(sstring dir, directory_entry_type type) {
         return do_start(dir, type).then([this] {
@@ -380,10 +425,9 @@ public:
     }
 
     future<> stop() {
-        if (_file) {
-            return _file.close();
-        }
-        return make_ready_future<>();
+        return stop_hook().finally([this] {
+            return maybe_close_file(_file);
+        });
     }
 
     const sstring name() const {
@@ -399,6 +443,7 @@ protected:
             { request_type::randwrite, "RAND WRITE" },
             { request_type::append , "APPEND" },
             { request_type::cpu , "CPU" },
+            { request_type::unlink, "UNLINK" },
         }[_config.type];;
     }
 
@@ -491,6 +536,9 @@ protected:
 
 public:
     virtual void emit_results(YAML::Emitter& out) = 0;
+    virtual future<> stop_hook() {
+        return make_ready_future<>();
+    }
 };
 
 class io_class_data : public class_data {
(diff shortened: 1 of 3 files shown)


pwrobelse:
This change is only a draft. It needs to be properly tested. Also, further description of `unlink_class_data` will be added to the commit message.

apps/io_tester/io_tester.cc
```diff
@@ -657,6 +665,96 @@ class write_io_class_data : public io_class_data {
...
+
+                            const uint64_t buffer_size{256ul << 10};
+                            const uint64_t buffers_count{static_cast<uint64_t>(fsize / buffer_size) + 1u};
+
+                            return do_with(boost::irange(UINT64_C(0), buffers_count), [f, buffer_size] (auto& buffers_range) mutable {
+                                return max_concurrent_for_each(buffers_range.begin(), buffers_range.end(), 64, [f, buffer_size] (auto buffer_id) mutable {
+                                    auto source_buffer = allocate_and_fill_buffer(buffer_size);
+                                    auto write_position = buffer_id * buffer_size;
+
+                                    return f.dma_write(write_position, source_buffer.get(), buffer_size).discard_result();
```
xemul, line 745: Why it _this_ here? The job is supposed to just unlink files, not unlink and then write something to them

apps/io_tester/io_tester.cc
```diff
@@ -657,6 +665,96 @@ class write_io_class_data : public io_class_data {
...
+        return format("{}/test-{}-shard-{:d}-file-{}", _dir_path, name(), this_shard_id(), file_id);
+    }
+
+    future<> do_start_on_directory(sstring path) {
+        _dir_path = std::move(path);
+
+        return seastar::parallel_for_each(boost::irange(UINT64_C(0), files_count()), [this] (uint64_t file_id) {
+            const auto fname = get_filename(file_id);
+            const auto fsize = _config.file_size;
+            const auto flags = open_flags::rw | open_flags::create;
```
xemul, line 723: This is not "unlink" job :) It's create-and-unlink one. I think files should be pre-created before measurement starts

apps/io_tester/io_tester.cc
```diff
@@ -657,6 +665,96 @@ class write_io_class_data : public io_class_data {
...
+    }
+
+    sstring get_filename(uint64_t file_id) const {
+        return format("{}/test-{}-shard-{:d}-file-{}", _dir_path, name(), this_shard_id(), file_id);
+    }
+
+    future<> do_start_on_directory(sstring path) {
+        _dir_path = std::move(path);
+
+        return seastar::parallel_for_each(boost::irange(UINT64_C(0), files_count()), [this] (uint64_t file_id) {
```
xemul, line 720: This makes unlink work at its maximum ~~speed~~ rate. However, more interesting case is when rate is configured and unlink tries to unlink "one file at N <x>seconds"

apps/io_tester/io_tester.cc
pwrobelse: Isn't it like that right now? In main() I can see:
```
std::cout << "Creating initial files..." << std::endl;
ctx.invoke_on_all([] (auto& c) {
    return c.start();
}).get();

std::cout << "Starting evaluation..." << std::endl;
ctx.invoke_on_all([] (auto& c) {
    return c.issue_requests();
}).get();
```

Your comment is related to `do_start_on_directory()` function, that is called via `start()`. I thought, that it is completed before the evaluation starts. Does the comment in main() lie? Or maybe I have mistaken something?

apps/io_tester/io_tester.cc
pwrobelse: Unlink is called in `future<size_t> issue_request(char *buf, io_intent* intent)` in line 687. Your comment refers to `do_start_on_directory()` which is a function called during startup to prepare files. Or at least I thought that it is called only during startup :sweat_smile: 

I agree, that calling `unlink()` could be configurable. I will try to implement the proposed comment.

However, I do not understand this particular statement - here `seastar::parallel_for_each()` is used to create files during startup and to ensure that they are created and filled at the maximum rate. These files are later unlinked in `issue_request()`.

This function is not run during evaluation (or I mistaken the way how main() works). However, from the code I would assume that `start()` finishes before starting the evaluation.

```
std::cout << "Creating initial files..." << std::endl;
ctx.invoke_on_all([] (auto& c) {
    return c.start();
}).get();

std::cout << "Starting evaluation..." << std::endl;
ctx.invoke_on_all([] (auto& c) {
    return c.issue_requests();
}).get();
```

apps/io_tester/io_tester.cc
pwrobelse: `maybe_remove_file()` is used to preserve files from previous run if they exist. Your comment is related to code that is executed during startup - it only prepares files and fills them. So, its logic is as follows:

1. Create N files via parallel_for:
1.1. If file exists and the user specified `preserve_files=true`, then ensure that its size is sufficient.
1.2. If the file exists, but user specified `preserve_files=false` then remove that file and create new one.
1.3. Ensure that either preserved or created file has the desired size and fill it with dummy data.

As far as I understand main(), it is not a part of evaluation. During evaluation only unlink is called.

```
    future<size_t> issue_request(char *buf, io_intent* intent) override {
        if (all_files_removed()) {
            return make_ready_future<size_t>(1u);
        }

        const auto fname = get_filename(_file_id_to_remove);
        ++_file_id_to_remove;

        return remove_file(fname).then([]{
            return make_ready_future<size_t>(1u);
        });
    }
```

pwrobelse: Well, I feel that I mistaken the semantic of  `maybe_remove_file()` in `io_class_data` - I tried to follow the same approach. The creation of files here will be adjusted.

pwrobelse: The PR has been adjusted.

This is a part of `do_start_on_directory()` that is called via `start()`. It creates and fills files (if needed), that will be unlinked during the evaluation.

apps/io_tester/io_tester.cc
```diff
@@ -657,6 +665,99 @@ class write_io_class_data : public io_class_data {
...
+                    return f.size().then([f, fsize] (uint64_t pre_truncate_size) mutable {
+                        return f.truncate(fsize).then([f, fsize, pre_truncate_size] () {
+                            if (pre_truncate_size >= fsize) {
+                                return make_ready_future<>();
+                            }
+
+                            const uint64_t buffer_size{256ul << 10};
+                            const uint64_t buffers_count{static_cast<uint64_t>(fsize / buffer_size) + 1u};
+
+                            return do_with(boost::irange(UINT64_C(0), buffers_count), [f, buffer_size] (auto& buffers_range) mutable {
```
xemul, line 743: Please, generalize this code with io_class_data::do_start_on_directory()

apps/io_tester/io_tester.cc
```diff
@@ -657,6 +665,99 @@ class write_io_class_data : public io_class_data {
...
+                                return max_concurrent_for_each(buffers_range.begin(), buffers_range.end(), 64, [f, buffer_size] (auto buffer_id) mutable {
+                                    auto source_buffer = allocate_and_fill_buffer(buffer_size);
+                                    auto write_position = buffer_id * buffer_size;
+
+                                    return f.dma_write(write_position, source_buffer.get(), buffer_size).discard_result();
+                                });
+                            }).then([f]() mutable {
+                                return f.flush();
```
xemul, line 751: File should also be close()d

apps/io_tester/io_tester.cc
```diff
@@ -207,8 +209,12 @@ struct job_config {
     ::options options;
     // size of each individual file. Every class and every shard have its file, so in a normal
     // system with many shards we'll naturally have many files and that will push the data out
-    // of the disk's cache
+    // of the disk's cache. An exception to that rule is unlink_class_data, that creates files_count
+    // files with file_size/files_count.
```
xemul, line 213: Why this exception?

apps/io_tester/io_tester.cc
```diff
@@ -657,6 +665,99 @@ class write_io_class_data : public io_class_data {
...
+    future<> do_start(sstring path, directory_entry_type type) override {
+        if (type == directory_entry_type::directory) {
+            return do_start_on_directory(path);
+        }
+        throw std::runtime_error(format("Unsupported storage. {} should be directory", path));
+    }
+
+    future<size_t> issue_request(char *buf, io_intent* intent) override {
+        if (all_files_removed()) {
+            return make_ready_future<size_t>(1u);
```
xemul, line 689: That's not really nice. The result would be silently skewed if all files are removed, but other workload continues to run. It should at least print a warning here, so that test runner knows that something dodn't go the way it was expected to.

Also, what if returning 0 here?

apps/io_tester/io_tester.cc
pwrobelse: Okay, a call to `close()` has been added. It will be uploaded in the next patch-set.

I did not add it previously, because I thought that the file is closed when the last object holding the reference to FD is destroyed. In this particular case I thought that the destruction of an object that is held in `do_with()` closes the file.

From `seastar/src/core/file.cc`:

```
posix_file_handle_impl::~posix_file_handle_impl() {
    if (_refcount && _refcount->fetch_add(-1, std::memory_order_relaxed) == 1) {
        ::close(_fd);
        delete _refcount;
    }
}
```

apps/io_tester/io_tester.cc
pwrobelse: The following new warning has been added:

```
fmt::print("[WARNING]: Cannot fulfill issue in unlink_class_data! All files have been removed for shard_id={}\n"
           "[WARNING]: Please create more files or adjust the frequency of unlinks.", this_shard_id());
```

Regarding returning `1u` here: I followed the approach from `cpu_class_data` that also does not perform any read or write operations.

In general, the returned value is accumulated in `class_data::_data` and is used only when throughput is calculated and emitted. However, in `unlink_class_data` throughput is not emitted. Therefore, usage of 0 is possible and it does not cause any difference in behavior of `unlink_class_data`.

apps/io_tester/io_tester.cc
pwrobelse: In `struct convert<job_config>::decode()` we have the following lines:

```
// The data_size is used to divide the available (and effectively
// constant) disk space between workloads. Each shard inside the
// workload thus uses its portion of the assigned space.
if (node["data_size"]) {
    cl.file_size = node["data_size"].as<byte_size>().size / smp::count;
} else {
    cl.file_size = 1ull << 30; // 1G by default
}
```

If `unlink_class_data` created `files_count` files with `file_size` then the assumption about `data_size` config field would be wrong.

As far as I understand `data_size`, it divides disk space between workloads. Each shard receives only a piece of that `data_size`. To ensure, that `unlink_class_data` uses only that piece, the calculation of a single file in that job is performed as follows:
`const auto fsize = _config.file_size / files_count();`

That's the reason why the comment in `struct job_config` was adjusted to:

```
// size of each individual file. Every class and every shard have its file, so in a normal
// system with many shards we'll naturally have many files and that will push the data out
// of the disk's cache. An exception to that rule is unlink_class_data, that creates files_count
// files with file_size/files_count.
uint64_t file_size;
```

Is it wrong? Which part should be corrected? Should the comment explain the origin of that exception or should this exception be removed altogether? Or maybe the calculation of `file_size` in `struct convert<job_config>::decode()` should be adjusted?

apps/io_tester/io_tester.cc
pwrobelse: Sure. Extracted the common part to a free function called `create_and_fill_file()`.

This function is invoked by `io_class_data` and `unlink_class_data` during `do_start_on_directory()`. It returns the pair: handle to file and the offset to the last write position - the latter is needed in case of `request_type::append`.

apps/io_tester/io_tester.cc
xemul: Yes, it does, but explicit is better than implicit

apps/io_tester/io_tester.cc
xemul: Actually, reporting back accumulated "throughput" or "IOPS" for unlink class would also be nice -- we'll see if regular IO running in parallel affects unlinkage rate or not

apps/io_tester/io_tester.cc
xemul: Ah, you're right. I forgot that `file_size` here is actually the `data_size` from yaml :( Let's keep this assumption for unlink class too

apps/io_tester/io_tester.cc
pwrobelse: @xemul: is this conversation resolved? `unlink_class_data` is derived from `class_data`, which allows it to use `rpm` and `think_time`.

According to `io-tester.md`:
- **rps**: the requests-per-second rate to apply to sending fibers. If unset, each fiber sends new request as soon as previous one completes
- **think_time**: how long to wait before submitting another request in this job once one finishes.

The manual tests show that these parameters enable users to configure `request_type::unlink` to be aligned with your comment.

apps/io_tester/io_tester.cc
pwrobelse: Okay. The following results have been added to `unlink_class_data`:
- IOPS
- latencies: [average, max]
- stats: [total_requests]

I am not sure how to calculate the bandwidth - shall `unlink_class_data` report the size of file that was removed or maybe we would like to report `1u` like `cpu_data_class`?

pwrobelse:
Hi @xemul, please find the adjustments introduced by the new patch-set:
1. The common part related to file creation has been extracted from `io_data_class` and `unlink_data_class` to the new free function called `create_and_fill_file()`. It returns file handle and last write position.
2. A warning is now printed by `unlink_class_data::issue_request()` when all files have been removed. It means, that the request cannot be fulfilled.
3. More fields are now printed via `unlink_data_class::emit_results()` including IOPS, average/max latencies and total number of requests.
4. A new member function called `stop_hook()` has been added to `class_data`. By default it is empty, but the derived classes may override it. Such new member function is used to inject a cleanup routine, that is specific for the derived class.
5. Added `stop_hook()` implementation for `unlink_data_class` to remove files, if any of them remain, when `keep_files == false`. This way the tested directory is clean when the execution finishes.

pwrobelse:
> Can you also put here some measurement results, e.g. -- read workload run on its own vs read workload run in parallel with unlink one to see how unlinking affects read latency

Hi @xemul, please find some measurements.

| Environment property | Value|
| --------------| --------- |
| Machine  | i3.xlarge |
| OS | Ubuntu 22.04 | 
| Test duration | 20s |
| SMP | 4 |


**Compared configurations:**
| | (1) randread | (2) randread+unlink_v1 | (3) randread+unlink_v2 |
| - |------------------ | --------------------------------- | ---------------------------------- |
| read data_size | 20GB |  20GB |  20GB |
| read reqsize | 512 | 512 | 512 |
| read parallelism | 1 | 1 | 1 |
| read shares | 100 | 100 | 100 |
| unlink file size | N/A | 1MB | 1MB |
| unlink think_time | N/A | 100us | 100us |
| unlink parallelism | N/A | 5 | 10 |


Results (shard0):
| | (1) randread | (2) randread+unlink_v1 | (3) randread+unlink_v2 |
| - |------------------ | --------------------------------- | ---------------------------------- |
| | | |
| | | |

PR #2069, opened 2024-02-01
treewide: replace calls to future::get0() by calls to future::get()

avikivity:
get0() is a remnant of the days of variadic futures.

To prepare for its deprecation, stop using it internally.


tchaikov (approved):
lgtm

PR #2070, opened 2024-02-01
docker: install clang-tools-18

avikivity:
to address the build failure caused by
https://github.com/llvm/llvm-project/issues/59827, and to test the build with C++23, we need to use clang-18. which will be released in March 2023.

Closes scylladb/seastar#2058

[avi: regenerate toolchain]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
       - checkout
       - run: git submodule sync
       - run: git submodule update --init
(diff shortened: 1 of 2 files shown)


avikivity:
@tchaikov tests fail

tchaikov:
@avikivity thank you for redo the image. because `clang-scan-deps` is packaged in `clang-tools-18` package. we need to update circleci to use clang-18 for building C++20 module. i updated #2058 to use `docker.io/scylladb/seastar-toolchain:2024-02-01`.

avikivity:
It still fails.

tchaikov:
> It still fails.

@avikivity after applying #2071 , the build should be green now .

tchaikov:
i am closing this change. as 233b48c152de353ff0b2e8cac8c57f71c2019ebc has landed on master.

PR #2072, opened 2024-02-01
circular_buffer: make the iterator default constructible

andijcr:
this pr makes circular_buffer::iterator default constructible, 
It is needed to make circular_buffer a regular range.

this check is added to circular_buffer_test.cc as a compile-time test:
```cpp
static_assert(std::ranges::range<circular_buffer<int>>);
```

### DIFF
--- a/include/seastar/core/circular_buffer.hh
+++ b/include/seastar/core/circular_buffer.hh
@@ -194,9 +194,10 @@ private:
        difference_type operator-(const cbiterator<CB, ValueType>& rhs) const noexcept {
             return idx - rhs.idx;
         }
+        cbiterator() = default;
     private:
-        CB* cb;
-        size_t idx;
+        CB* cb{nullptr};
+        size_t idx{0};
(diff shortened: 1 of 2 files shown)


tchaikov (approved):
`std::sentinel_for` requires `std::semiregular` which in turn requires `std::default_initializable`, so to make `circular_buffer` satisfy `std::ranges::range`, we have to make `circular_buff::iterator` constructible with `circular_buff::iterator{}`, `circular_buff::iterator it` and `circular_buff::iterator()`.

include/seastar/core/circular_buffer.hh
```diff
@@ -194,9 +194,10 @@ private:
        difference_type operator-(const cbiterator<CB, ValueType>& rhs) const noexcept {
             return idx - rhs.idx;
         }
+        cbiterator() = default;
     private:
-        CB* cb;
-        size_t idx;
+        CB* cb{nullptr};
+        size_t idx{0};
```
avikivity, line 200: These are unnecessary (default constructed iterator is undefined), but I guess doesn't hurt.

PR #2073, opened 2024-02-02
Avoid some in-code ifdefs in reactor.cc

xemul:
Preprocessor conditionals (#if, #ifdef) in the middle of a function body make code harder to read and logic harder to follow. This patch re-ifdefs some of them to produce specific or stub implementations for different if/else cases. The compiler is good at not generating any code for stub calls.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4069,6 +4069,14 @@ static void sigabrt_action() noexcept {
     reraise_signal(SIGABRT);
 }
 
+// We don't need to handle SIGSEGV when asan is enabled.
+#ifdef SEASTAR_ASAN_ENABLED
+template<>
+void install_oneshot_signal_handler<SIGSEGV, sigsegv_action>() {
+    (void)sigsegv_action;
+}
+#endif
+
 void smp::qs_deleter::operator()(smp_message_queue** qs) const {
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -891,6 +892,13 @@ static decltype(auto) install_signal_handler_stack() {
         }
     });
 }
+#else
+// SIGSTKSZ is too small when using asan. We also don't need to
+// handle SIGSEGV ourselves when using asan, so just don't install
+// a signal handler stack.
+struct nothing { ~nothing() {} };
+nothing install_signal_handler_stack() { return nothing{}; }
```
tchaikov, line 900: nit, could move `nothing` into the body of `install_signal_handler_stack()`, and make the function to return `auto`, so the symbol of `nothing` is not exposed, and we don't need to move it into an anonymous namespace.

xemul: Hm, good point, will try

xemul: Yup, it did the trick, thanks!

xemul:
Upd:
- moved the `struct nothing` stub into a function scope not to pollute file namespace

tchaikov (approved):
lgtm

PR #2076, opened 2024-02-04
include/seastar/util: fix compilation error

Matan-B:
Clang 14.0.0
Full Log: https://jenkins.ceph.com/job/ceph-pull-requests/128658/consoleText

---

Revert "util/source_location-compat: mark `source_location::current() consteval"

```
../src/seastar/include/seastar/util/log.hh:191:51: error: cannot take address of consteval function 'current' outside of an immediate invocation
        format_info(compat::source_location loc = compat::source_location::current()) noexcept
                                                  ^
../src/seastar/include/seastar/util/source_location-compat.hh:50:21: note: declared here
    source_location current(const char* file = __builtin_FILE(), const char* func = __builtin_FUNCTION(), int line = __builtin_LINE(), int col = 0) noexcept {
```

This reverts commit 550da0e6750a625a171934240c9f7842ab05bc58.

### DIFF
--- a/include/seastar/util/source_location-compat.hh
+++ b/include/seastar/util/source_location-compat.hh
@@ -43,11 +43,8 @@ public:
         , _line(0)
         , _col(0)
     { }
-    static
-#if __cplusplus >= 202002L
-    consteval
-#endif
-    source_location current(const char* file = __builtin_FILE(), const char* func = __builtin_FUNCTION(), int line = __builtin_LINE(), int col = 0) noexcept {
+
+    static source_location current(const char* file = __builtin_FILE(), const char* func = __builtin_FUNCTION(), int line = __builtin_LINE(), int col = 0) noexcept {
         return source_location(file, func, line, col);
     }
 


tchaikov:
@Matan-B hi Matan, we only support the latest two major releases of GCC and Clang. at the time of writing, Clang has released 17, and 18 has branched. and Clang-18 will be released in early March, see https://discourse.llvm.org/t/llvm-18-release-schedule/76175 . so i would suggest install a newer version of Clang in Ceph's CI.

avikivity:
I don't mind accepting contributions to support more versions;  the two-version limit means we won't do any special work, but we can accept fixes.

tchaikov:
@avikivity @Matan-B `consteval` was added for the fmt compile-time check. if we want to consider fix the build with clang-14, we should

tchaikov:
FWIW, it's a known issue reported at https://bugs.llvm.org/show_bug.cgi?id=48886, and a minimal reproducer can be found at https://godbolt.org/z/novhK13fE

avikivity:
We shouldn't just revert it. @Matan-B if you want to fix something for an older compiler, you need to retain the changes made for newer compilers.

avikivity:
And really, upgrading the compiler is better.

tchaikov:
> @avikivity @Matan-B this change reverts [550da0e](https://github.com/scylladb/seastar/commit/550da0e6750a625a171934240c9f7842ab05bc58). `consteval` was added for the fmt compile-time check and to be more C++20 compatible. if we want to consider fix the build with clang-14, we need to reconsider the original fix.

we don't mark `format_info::format_info()` with `consteval` anymore, because when working on the compile-time check of format in the logging subsystem, it's not a reason anymore for keeping the `consteval` here. but it's still nice to be C++20 compatible -- to have `consteval` for `std::source_location::current()`.  see https://en.cppreference.com/w/cpp/utility/source_location/current

Matan-B:
@tchaikov,  Clang 14 compiles successfully w/ https://github.com/scylladb/seastar/pull/2077. Should we close this one?

tchaikov:
@Matan-B hi Matan, that's a great news. thank you for the update. please feel free to close this PR unless any of us wants to work on it. IMO, it's a workaround of a known bug of the Clang compiler.

Matan-B:
Closing as the issue no longer persists.
Attaching a successful clang 14 log: https://jenkins.ceph.com/job/ceph-pull-requests/128790/consoleText

@tchaikov @avikivity, Thank you for taking a look!

PR #2077, opened 2024-02-05
util/source_location-compat: use __cpp_consteval

tchaikov:
we should use `__cpp_consteval` for checking the availability of `consteval` instead of using `__cplusplus`. the former is more specific than the later.


tchaikov:
while looking at #2076, i realized that i should have used a better macro for checking the feature.

tchaikov:
```
The following tests FAILED:
         31 - Seastar.unit.httpd (Failed)
```
known failure #1645

PR #2050, opened 2024-01-22
github/workflow: add a github workflow for running tests

tchaikov:
before this change, contributor is required to sign on CircleCI for triggering the tests for his/her pull request, but not all contributrs register this service.

after this change, contributors are able to run tests as long as the pull request is created on GitHub,

PR #2081, opened 2024-02-06
test/http: Read full server response before sending next

xemul:
The test_100_continue test cases sends a series of requests to the test server with different headers and bodies in order to check how server responds to it. The problem of this test is that it reads back from server with single input_stream<>::read() calls. Most of the time it works OK, but from time to time server sends headers then preempts and appends the response body later. If this happens client reads response headers, parses what it wants and proceeds with the next query. But reading back the response for this next query results in reading back body from the previous query, thus failing the test.

fixes: #1645

### DIFF
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -853,24 +853,43 @@ SEASTAR_TEST_CASE(test_100_continue) {
             for (auto version : {sstring("1.0"), sstring("1.1")}) {
                 for (auto content : {sstring(""), sstring("xxxxxxxxxxx")}) {
                     for (auto expect : {sstring(""), sstring("Expect: 100-continue\r\n"), sstring("Expect: 100-cOnTInUE\r\n")}) {
+                        bool need_cont = (version == "1.1" && expect.length());
+                        bool body_sent = false;
                         auto content_len = content.empty() ? sstring("") : (sstring("Content-Length: ") + to_sstring(content.length()) + sstring("\r\n"));
                         sstring req = sstring("GET /test HTTP/") + version + sstring("\r\nHost: test\r\nConnection: Keep-Alive\r\n") + content_len + expect + sstring("\r\n");
+                        sstring resp;
+
                         output.write(req).get();
                         output.flush().get();
-                        bool already_ok = false;
-                        if (version == "1.1" && expect.length()) {
-                            auto resp = input.read().get();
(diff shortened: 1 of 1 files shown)


tests/unit/httpd_test.cc
```diff
@@ -853,24 +853,42 @@ SEASTAR_TEST_CASE(test_100_continue) {
...
                             output.write(content).get();
                             output.flush().get();
-                            auto resp = input.read().get();
-                            BOOST_REQUIRE_EQUAL(std::string(resp.get(), resp.size()).find("100 Continue"), std::string::npos);
-                            BOOST_REQUIRE_NE(std::string(resp.get(), resp.size()).find("200 OK"), std::string::npos);
+                            body_sent = true;
+                        }
+
+                        while (true) {
+                            auto r = input.read().get();
```
tchaikov, line 872: shall we check if `r` is empty? as it indicates eof,  if we run into eof before reading the response containing "hello", probably we should fail the test, instead of waiting forever in a tight loop.

nyh (approved):
Looks good. I agree with @tchaikov that it's wise to stop the test immediately on eof instead of going into a busy loop.

tests/unit/httpd_test.cc
xemul: Good point, fixed

tests/unit/httpd_test.cc
```diff
@@ -853,24 +853,42 @@ SEASTAR_TEST_CASE(test_100_continue) {
             for (auto version : {sstring("1.0"), sstring("1.1")}) {
                 for (auto content : {sstring(""), sstring("xxxxxxxxxxx")}) {
                     for (auto expect : {sstring(""), sstring("Expect: 100-continue\r\n"), sstring("Expect: 100-cOnTInUE\r\n")}) {
+                        bool need_cont = (version == "1.1" && expect.length());
+                        bool body_sent = false;
```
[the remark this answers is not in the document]
xemul: Actually, I put the need_cont closer to version and expect since it's evaluated from those, for better readability :)

xemul:
upd:
- check for non-EOF while reading server response

tchaikov (approved):
lgtm

nyh:
Merged. Note that CI did not run for political reasons, so I only checked the build on one compiler.

tchaikov:
> Merged. Note that CI did not run for political reasons, so I only checked the build on one compiler.

i am trying to address this with #2050

Issue #2082, opened 2024-02-07
`httpd::routes::handle()` should verify http param's data type


tchaikov:
if an http param's data type is specified, we should verify it. currently, `routes::handle()` just check if the required parameters exist in the request. but their types are not checked at all. the application is responsible to verify the type of parameters.

would be great if `httpd::routes` can verify the parameter and even return the parameters using an interface like 
```c++
http::request::get_query_param_as<int>(std::string_view) const
```

PR #2084, opened 2024-02-07
http: add handler_base::verify_mandatory_params()

tchaikov:
before this change, this member function was a free function called
by `routes::handle()`, but since `handler_base` is closer to the
`_mandatory_param` and in the following change, we will store
the expected param types in `_mandatory_param`, this makes
`handler_base` a better home for the verification function. so to
prepare for the change, let's move this free function into handler_base.

also, let's mark `_mandatory_param` private, allows us to store the
typing of the params into `_mandatory_param`. as this variable should
not be visible from the outer world. and the typing information should
be opaque from the caller.

Refs #2082

### DIFF
--- a/include/seastar/http/handlers.hh
+++ b/include/seastar/http/handlers.hh
@@ -23,6 +23,7 @@
 
 #include <seastar/http/request.hh>
 #include <seastar/http/common.hh>
+#include <seastar/http/exception.hh>
 #include <seastar/http/reply.hh>
 
 #include <unordered_map>
@@ -40,6 +41,7 @@ typedef const http::request& const_req;
  *
  */
 class handler_base {
+    std::vector<sstring> _mandatory_param;
 protected:
     handler_base() = default;
     handler_base(const handler_base&) = default;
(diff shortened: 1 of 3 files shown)


include/seastar/http/handlers.hh
```diff
@@ -66,8 +67,7 @@ public:
         return *this;
     }
 
-    std::vector<sstring> _mandatory_param;
-
+    friend class routes;
```
xemul, line 70: This caught my eye. I think it's better to get const getter for _mandatory_param reference, _or_ move the verify_param() helper to be handler_base's method

tchaikov: i plan to stuff move stuff into `_mandatory_param`, so we can verify the type of the params right in seastar, instead of in the higher application level. so the const getter is a no-go, as it *exposes* the internals to the public. so i am repurposing this PR to go with the 2nd route.

tchaikov:
v2:

- "move the verify_param() helper to be handler_base's method"

PR #2085, opened 2024-02-08
coroutine/exception: document return_exception_ptr()

tchaikov:
before this change, the doxygen comment above `return_exception_ptr()` is for `return_exception()`, which was marked deprecated. so the document does not apply anymore. so, in this change:

* move the comment for `return_exception()` down to where it belongs
* add the comment for `return_exception_ptr()`, and add a real-life example for it.
* replace the references to "return_exception" with "return_exception_ptr". as the former has been deprecated in favor of the latter.


bhalevy (approved):
LGTM, thanks for fixing

PR #2087, opened 2024-02-09
fix the behavour of sstring::starts_with and ends_with for two equal strings

hurricane1026:
In c++20, 

```
std::string same_str{"seastar"};
auto eq_str{same_str};
assert(same_str.starts_with(eq_str));
assert(same_str.ends_with(eq_str));
```

but, seastar::sstring::starts_with and ends_withs can not support same string.

### DIFF
--- a/include/seastar/core/sstring.hh
+++ b/include/seastar/core/sstring.hh
@@ -597,7 +597,7 @@ public:
     }
 
     constexpr bool starts_with(std::basic_string_view<char_type, traits_type> sv) const noexcept {
-        return size() > sv.size() && compare(0, sv.size(), sv) == 0;
+        return size() >= sv.size() && compare(0, sv.size(), sv) == 0;
     }
 
     constexpr bool starts_with(char_type c) const noexcept {
@@ -609,7 +609,7 @@ public:
     }
 
     constexpr bool ends_with(std::basic_string_view<char_type, traits_type> sv) const noexcept {
-        return size() > sv.size() && compare(size() - sv.size(), npos, sv) == 0;
+        return size() >= sv.size() && compare(size() - sv.size(), npos, sv) == 0;
     }
 
     constexpr bool ends_with(char_type c) const noexcept {
--- a/tests/unit/sstring_test.cc
+++ b/tests/unit/sstring_test.cc
@@ -113,6 +113,8 @@ BOOST_AUTO_TEST_CASE(test_str_not_find_sstring) {
 }
 
 BOOST_AUTO_TEST_CASE(test_str_starts_with) {
+    BOOST_CHECK(std::string("abcdefg").starts_with("abcdefg"));
+    BOOST_CHECK(sstring("abcdefg").starts_with("abcdefg"));
     BOOST_CHECK(sstring("abcdefg").starts_with("ab"sv));
     BOOST_CHECK(sstring("abcde").starts_with('a'));
     BOOST_CHECK(sstring("abcde").starts_with("ab"));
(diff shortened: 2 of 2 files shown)


xemul:
Please, extend the tests/unit/sstring_test.cc to validate this works

nyh:
> Please, extend the tests/unit/sstring_test.cc to validate this works

Right. A good example is the test function `test_find_sstring_compatible` - note how it checks the same thing in both seastar::sstring and std::string to prove the test is correct. In other words, if the stated point of the patch is to make seastar::sstring be more compatible with std::string, the test should verify this instead of showing only how seastar::sstring works, and hoping you correctly checked how std::string works.

Also, please note that in a single-patch pull request, the cover letter (the first message on the pull request) is lost. You need to put the explanations into the commit itself (right now the commit only has a title).

nyh:
Anyway, thanks for finding this bug - it's amazing we haven't noticed this before, I guess these new C++20 functions aren't too popular, and even if used, not to compare two equal strings (I just reviewed the uses in ScyllaDB, and none are vulnerable to this bug).

I'll be happy to merge this once you add a test as suggested above. Thanks.

hurricane1026:
> Please, extend the tests/unit/sstring_test.cc to validate this works

ok, I will do that tomorrow.

hurricane1026:
> Anyway, thanks for finding this bug - it's amazing we haven't noticed this before, I guess these new C++20 functions aren't too popular, and even if used, not to compare two equal strings (I just reviewed the uses in ScyllaDB, and none are vulnerable to this bug).
> 
> I'll be happy to merge this once you add a test as suggested above. Thanks.

hi, nyh, I just add few simple test code to verify the code change, if you have any question about this PR, I was glad to get future update.

xemul (approved):

tchaikov (approved):
lgtm. thanks!

tchaikov:
the title reads: "sstring starts_with and ends_with should behaviour same as std::string". this looks like a title for a bug report. please summarize the change in the title. and put the rationales in the commit message.

nyh:
> the title reads: "sstring starts_with and ends_with should behaviour same as std::string". this looks like a title for a bug report. please summarize the change in the title. and put the rationales in the commit message.

Right, the title is also too general. You're solving one specific problem - not "same behavior as std::string" in general. How about the following title?

"Fix behavior of sstring::starts_with() and ends_with() for two equal strings"

include/seastar/core/sstring.hh
```diff
@@ -597,7 +597,7 @@ public:
     }
 
     constexpr bool starts_with(std::basic_string_view<char_type, traits_type> sv) const noexcept {
-        return size() > sv.size() && compare(0, sv.size(), sv) == 0;
+        return size() >= sv.size() && compare(0, sv.size(), sv) == 0;
```
nyh, line 600: As in the cover letter, please also rename this patch, and add some text explaining what you did (i.e., fix the behavior of these functions when given two identical strings - x.starts_with(x) should be true, not false.
"same as std::string" is much too general.

nyh (changes requested):
Looks good, just please improve the commit messages as @tchaikov asked (and I elaborated), and we're good to go. Thanks.

include/seastar/core/sstring.hh
hurricane1026: ok, have re-pushed and changed the PR's title.

nyh: Thanks.

nyh (approved):

PR #2088, opened 2024-02-09
file: Remove unused thread-pool arg from directory lister

xemul:
It's an argument lost by 95a7ff9989 from one of its draft implementations


tchaikov (approved):
lgtm

PR #2091, opened 2024-02-11
Update supported C++ standards to C++23 and C++20 (dropping C++17)

tchaikov:
C++23 has passed all the ballots [1]. We can therefore support it now formally. This means that C++17 is no longer supported, and C++20 is supported until C++26 is released.

The primary benefit of C++23 support is that we can now use coroutines in core Seastar code, as all supported C++ versions have them.

[1] https://lists.isocpp.org/std-proposals/2024/01/8880.php

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,8 +70,8 @@ workflows:
       - build_and_test:
           matrix:
             parameters:
-              compiler: ["clang++-17", "g++-13"]
-              standard: ["20", "17"]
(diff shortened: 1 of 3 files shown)


tchaikov:
this change is based on #2053. on top of which, it also replaces `clang++-17` with `clang++-18` in `.circleci/config.yml`.

tchaikov:
@avikivity Hi Avi, i took the liberty to rebase on the master, and sent this change. this should enable us to start coroutinizing Seastar.

Issue #202, opened 2016-10-07
input_stream::read_exactly_part() uses recursion which may overflow the stack for large sizes


tgrabiec:
Refs #195

niekbouman:
We seem to have hit this issue recently.

@avikivity 
If I'd like to fix this and submit a PR or patch, can I do so by transforming `read_exactly_part` into a coroutine?

(given that you just switched to c++20/c++23 standards? 
https://github.com/scylladb/seastar/commit/5d3ee980730b49c2379a2728d69698ee68f5dbeb )

avikivity:
Yes

niekbouman:
> Yes

thanks
please see https://github.com/scylladb/seastar/pull/2093

PR #2093, opened 2024-02-11
Replace recursion by iteration in input_stream::read_exactly_part

niekbouman:
Currently `read_exactly_part` (found in the iostream-impl.hh header) uses recursion, which can lead to stack overflow. This commit rewrites `read_exactly_part` into a coroutine that does not use recursion. Fixes https://github.com/scylladb/seastar/issues/202

### DIFF
--- a/include/seastar/core/iostream-impl.hh
+++ b/include/seastar/core/iostream-impl.hh
@@ -152,7 +152,9 @@ future<> output_stream<CharType>::write(temporary_buffer<CharType> p) noexcept {
 
 template <typename CharType>
 future<temporary_buffer<CharType>>
-input_stream<CharType>::read_exactly_part(size_t n, tmp_buf out, size_t completed) noexcept {
+input_stream<CharType>::read_exactly_part(size_t n) noexcept {
+    temporary_buffer<CharType> out(n);
+    size_t completed{0U};
     while (completed < n) {
         size_t avail = available();
         if (avail) {
@@ -160,10 +162,9 @@ input_stream<CharType>::read_exactly_part(size_t n, tmp_buf out, size_t complete
             std::copy_n(_buf.get(), now, out.get_write() + completed);
             _buf.trim_front(now);
             completed += now;
-        }
-
-        if (completed == n) {
-            break;
+            if (completed == n) {
+                break;
+            }
         }
 
         // _buf is now empty
@@ -200,13 +201,8 @@ input_stream<CharType>::read_exactly(size_t n) noexcept {
             return this->read_exactly(n);
         });
     } else {
-      try {
         // buffer too small: start copy/read loop
-        tmp_buf b(n);
-        return read_exactly_part(n, std::move(b), 0);
-      } catch (...) {
-        return current_exception_as_future<tmp_buf>();
-      }
+        return read_exactly_part(n);
     }
 }
 
--- a/include/seastar/core/iostream-impl.hh
+++ b/include/seastar/core/iostream-impl.hh
@@ -23,6 +23,7 @@
 
 #pragma once
 
+#include <seastar/core/coroutine.hh>
 #include <seastar/core/do_with.hh>
 #include <seastar/core/loop.hh>
 #include <seastar/net/packet.hh>
@@ -152,26 +153,29 @@ future<> output_stream<CharType>::write(temporary_buffer<CharType> p) noexcept {
 template <typename CharType>
 future<temporary_buffer<CharType>>
 input_stream<CharType>::read_exactly_part(size_t n, tmp_buf out, size_t completed) noexcept {
-    if (available()) {
-        auto now = std::min(n - completed, available());
-        std::copy(_buf.get(), _buf.get() + now, out.get_write() + completed);
-        _buf.trim_front(now);
-        completed += now;
-    }
-    if (completed == n) {
-        return make_ready_future<tmp_buf>(std::move(out));
-    }
+    while (completed < n) {
+        size_t avail = available();
+        if (avail) {
+            auto now = std::min(n - completed, avail);
+            std::copy_n(_buf.get(), now, out.get_write() + completed);
+            _buf.trim_front(now);
+            completed += now;
+        }
 
-    // _buf is now empty
-    return _fd.get().then([this, n, out = std::move(out), completed] (auto buf) mutable {
+        if (completed == n) {
+            break;
+        }
+
+        // _buf is now empty
+        temporary_buffer<CharType> buf = co_await _fd.get();
         if (buf.size() == 0) {
             _eof = true;
             out.trim(completed);
-            return make_ready_future<tmp_buf>(std::move(out));
+            break;
         }
         _buf = std::move(buf);
-        return this->read_exactly_part(n, std::move(out), completed);
-    });
+    }
+    co_return out;
 }
 
 template <typename CharType>
--- a/include/seastar/core/iostream.hh
+++ b/include/seastar/core/iostream.hh
@@ -355,7 +355,7 @@ public:
     /// \returns the data_source
     data_source detach() &&;
 private:
-    future<temporary_buffer<CharType>> read_exactly_part(size_t n, tmp_buf buf, size_t completed) noexcept;
+    future<temporary_buffer<CharType>> read_exactly_part(size_t n) noexcept;
 };
 
 struct output_stream_options {


include/seastar/core/iostream-impl.hh
```diff
@@ -152,26 +153,28 @@ future<> output_stream<CharType>::write(temporary_buffer<CharType> p) noexcept {
...
+            _buf.trim_front(now);
+            completed += now;
+        }
+    
+        if (completed == n) {
+            co_return out;
+        }
+    
+        // _buf is now empty
+        temporary_buffer<CharType> buf = co_await _fd.get();
```
niekbouman, line 170: If I understand correctly, even though `co_await` could throw, this will be translated back to an exceptional future,  so that `read_exactly_part` being `noexcept` does not lead to trouble, right?

tchaikov: right. `noexcept` is literally `noexcept`, which has nothing to do the coroutine abstraction. the exception "thrown" by `_fd.get()` would be propagated via `promise_type::unhandled_exception()`.

tchaikov:
also, i'd suggest use a prefix in the title of the commit message. you could use `git log include/seastar/core/iostream-impl.hh` for some examples. and would be great if you could put `Fixes #202` on a separate line.

include/seastar/core/iostream-impl.hh
```diff
@@ -152,26 +153,28 @@ future<> output_stream<CharType>::write(temporary_buffer<CharType> p) noexcept {
...
-        _buf.trim_front(now);
-        completed += now;
-    }
-    if (completed == n) {
-        return make_ready_future<tmp_buf>(std::move(out));
-    }
-
-    // _buf is now empty
-    return _fd.get().then([this, n, out = std::move(out), completed] (auto buf) mutable {
+    while(true) {
```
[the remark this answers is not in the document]
niekbouman: agreed, see updated commit

niekbouman:
> also, i'd suggest use a prefix in the title of the commit message. you could use `git log include/seastar/core/iostream-impl.hh` for some examples. and would be great if you could put `Fixes #202` on a separate line.

done.

include/seastar/core/iostream-impl.hh
```diff
@@ -152,26 +153,28 @@ future<> output_stream<CharType>::write(temporary_buffer<CharType> p) noexcept {
...
+            std::copy_n(_buf.get(), now, out.get_write() + completed);
+            _buf.trim_front(now);
+            completed += now;
+        }
 
-    // _buf is now empty
-    return _fd.get().then([this, n, out = std::move(out), completed] (auto buf) mutable {
+        if (completed == n) {
+            co_return out;
+        }
```
tchaikov, line 167: how about just `break`?

include/seastar/core/iostream-impl.hh
```diff
@@ -152,26 +153,28 @@ future<> output_stream<CharType>::write(temporary_buffer<CharType> p) noexcept {
...
+
+        // _buf is now empty
+        temporary_buffer<CharType> buf = co_await _fd.get();
         if (buf.size() == 0) {
             _eof = true;
             out.trim(completed);
-            return make_ready_future<tmp_buf>(std::move(out));
+            co_return out;
```
tchaikov, line 174: and `break` here? so that we can put `co_return` after the `while ()` loop.

include/seastar/core/iostream-impl.hh
```diff
@@ -152,26 +153,28 @@ future<> output_stream<CharType>::write(temporary_buffer<CharType> p) noexcept {
...
-    if (available()) {
-        auto now = std::min(n - completed, available());
-        std::copy(_buf.get(), _buf.get() + now, out.get_write() + completed);
-        _buf.trim_front(now);
-        completed += now;
-    }
-    if (completed == n) {
-        return make_ready_future<tmp_buf>(std::move(out));
-    }
+    while (completed < n) {
```
tchaikov, line 156: the reason why we passed `completed` to this function was that we wanted to call this function recursively. i think since we have the luxury of using coroutine, why not just drop the `completed` parameter?

niekbouman: What's the opinion on such breaking API changes? I guess there's a tradeoff between making code cleaner and breaking api.

include/seastar/core/iostream-impl.hh
```diff
@@ -152,26 +153,28 @@ future<> output_stream<CharType>::write(temporary_buffer<CharType> p) noexcept {
...
 input_stream<CharType>::read_exactly_part(size_t n, tmp_buf out, size_t completed) noexcept {
-    if (available()) {
-        auto now = std::min(n - completed, available());
-        std::copy(_buf.get(), _buf.get() + now, out.get_write() + completed);
-        _buf.trim_front(now);
-        completed += now;
-    }
-    if (completed == n) {
-        return make_ready_future<tmp_buf>(std::move(out));
-    }
```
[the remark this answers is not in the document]
niekbouman: will try that.

include/seastar/core/iostream-impl.hh
tchaikov: it does not break the public API. it is but a private function.

include/seastar/core/iostream-impl.hh
tchaikov: it's more of a matter of style. i made this suggestion just because the build failed. and thought it might be more straightforward to address the build failure.

niekbouman: Ok. I was too rushed and therefore sloppy (and introduced the build failure). Will fix this.

include/seastar/core/iostream-impl.hh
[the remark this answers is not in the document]
niekbouman: fast path is now inside `read_exactly` (modulo some minor code duplication)

include/seastar/core/iostream-impl.hh
niekbouman: ultimately decided against removing this parameter, due to Avi's idea about fast path vs slow path.

include/seastar/core/iostream-impl.hh
```diff
@@ -198,8 +202,21 @@ input_stream<CharType>::read_exactly(size_t n) noexcept {
     } else {
       try {
         // buffer too small: start copy/read loop
-        tmp_buf b(n);
-        return read_exactly_part(n, std::move(b), 0);
+        tmp_buf buf(n);
+        size_t count {0U};
+        size_t avail = available();
+        if (avail) {
```
niekbouman, line 208: is this branch actually good to have?
(we can also simply compute `count`, and call copy_n and trim_front, even when `count == 0`.)

avikivity: I think I gave you bad advice re the split - all the no-allocation cases are covered in read_exactly(). If we ever drop into read_exactly_part(), that means we'll allocate for sure (e.g. we won't hit `if (completed == n)` in the first iteration. So this patch is unnecessary.

Please verify and drop if I'm right in this second iteration.

niekbouman: I've dropped the unnecessary commit. Added a new commit where I do some more minor refactorings.

I've left the `if(_buf.size() == 0)` as a follow up, because in `read_exactly_part` there is at the beginning the allocation of the tmp_buf of size `n`, which would be redundant in the `if(_buf.size() == 0)` branch in `read_exactly` in case of EOF. Does that make sense?

include/seastar/core/iostream-impl.hh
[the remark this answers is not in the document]
niekbouman: we concluded that the split was not a good idea after all, see other conversation.

include/seastar/core/iostream-impl.hh
```diff
@@ -151,27 +152,31 @@ future<> output_stream<CharType>::write(temporary_buffer<CharType> p) noexcept {
...
+    temporary_buffer<CharType> out(n);
+    size_t completed{0U};
+    while (completed < n) {
+        size_t avail = available();
+        if (avail) {
+            auto now = std::min(n - completed, avail);
+            std::copy_n(_buf.get(), now, out.get_write() + completed);
+            _buf.trim_front(now);
+            completed += now;
+            if (completed == n) {
```
niekbouman, line 165: note that I moved this `if (completed == n)` branch inside the `if (avail)` branch

include/seastar/core/iostream-impl.hh
```diff
@@ -196,13 +201,8 @@ input_stream<CharType>::read_exactly(size_t n) noexcept {
             return this->read_exactly(n);
         });
     } else {
-      try {
         // buffer too small: start copy/read loop
-        tmp_buf b(n);
-        return read_exactly_part(n, std::move(b), 0);
-      } catch (...) {
```
niekbouman, line 203: this exception-handling machinery can be removed, right? (If I understand correctly, it serves no purpose now that the allocation of the tmp_buf has been moved into `read_exactly_part`.

avikivity: Right, and since coroutines automatically wrap exceptions in futures.

PR #2094, opened 2024-02-12
util/log: add formatter for log_level

tchaikov:
before this change, we rely on the default-generated fmt::formatter
created from operator<<, but fmt v10 dropped the default-generated
formatter.

in this change, we define formatters for `seastar::log_level`, its
operator<< is preserved for backward compatibility.

Refs https://github.com/scylladb/scylladb/issues/13245

### DIFF
--- a/src/util/log.cc
+++ b/src/util/log.cc
@@ -72,6 +72,14 @@ struct wrapped_log_level {
     seastar::log_level level;
 };
 
+static const std::map<seastar::log_level, std::string_view> log_level_names = {
+        { seastar::log_level::trace, "trace" },
+        { seastar::log_level::debug, "debug" },
+        { seastar::log_level::info, "info" },
+        { seastar::log_level::warn, "warn" },
+        { seastar::log_level::error, "error" },
+};
+
 namespace fmt {
 template <> struct formatter<wrapped_log_level> {
     using log_level = seastar::log_level;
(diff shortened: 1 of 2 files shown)


src/util/log.cc
```diff
@@ -112,6 +112,19 @@ template <> struct formatter<wrapped_log_level> {
...
 
+static const std::map<seastar::log_level, std::string_view> log_level_names = {
+        { seastar::log_level::trace, "trace" },
+        { seastar::log_level::debug, "debug" },
+        { seastar::log_level::info, "info" },
+        { seastar::log_level::warn, "warn" },
+        { seastar::log_level::error, "error" },
+};
+
+auto fmt::formatter<seastar::log_level>::format(seastar::log_level level, fmt::format_context& ctx) const
```
xemul, line 123: Why not put the formatter next to where log_level_names currently is? Instead of moving the log_level_names to new location

tchaikov: @xemul i didn't want to open/close the namespace of `seastar::internal` namespace in a single file. please imagine something like this:
```c++
namespace seastar {
namespace internal {
  // log_buf stuff
static const std::map<seastar::log_level, std::string_view> log_level_names = {
        { seastar::log_level::trace, "trace" },
        { seastar::log_level::debug, "debug" },
        { seastar::log_level::info, "info" },
        { seastar::log_level::warn, "warn" },
        { seastar::log_level::error, "error" },
};
}
}

auto fmt::formatter<seastar::log_level>::format(seastar::log_level level, fmt::format_context& ctx) const
    -> decltype(ctx.out()) {
    return fmt::format_to(ctx.out(), "{}", log_level_names.at(level));
}

namespace seastar {
namespace internal {
  // log_level stuff
}
```

i think the level of evilness is lower than that of the above. IMHO, this compromise is similar to 032c2cd11625b26da0e6e4ea2eb6649a729f65cf . what do you think?

xemul: Ah, I see. In that case I suggest to move log_level_names even _higher_, there's already the namespace fmt up there with formatter<wrapped_log_level> so we can re-use this namespace and pack all formatters together. WDYT?

PR #2095, opened 2024-02-12
test/unit: do not check if __cplusplus > 201703L

tchaikov:
now that we've dropped C++17 support in 01257e6058, there is no need to check `__cplusplus > 201703L`, as it's always true now.


tchaikov:

PR #2103, opened 2024-02-13
Cleanup rpc FrameType:: API

xemul:
There are several frame types declared in rpc.cc, each with a certain API that's used by read helpers. This set makes the API a bit shorter as preparation to RPC round-trip-time measurement and as continuation of #1687


xemul:
@avikivity , please, review

avikivity:
@gleb-cloudius please review

avikivity:
> @gleb-cloudius please review

Ah he's away.

PR #2106, opened 2024-02-17
sstring: implement push_back()

### LINKED ISSUE #2104: basic_sstring should have`push_back()`
we need to implement `constexpr void push_back(char_type ch);` for `basic_sstring<..>`.

tchaikov:
to be more compatible with basic_string<>
this is also what we expect from a sequence-alike STL containers.

Fixes #2104


include/seastar/core/sstring.hh
```diff
@@ -346,6 +346,29 @@ public:
...
+    /**
+     * Append the given character to the end of the string
+     * @param ch  The character to append
+     * @note unlike @c std::basic_string, sstring does not preallocate, so
+     *       this call always lead to reallocation if external storage is
+     *       used.
+     */
+    constexpr void push_back(char_type ch) {
+        if (is_internal()) {
+            if (static_cast<Size>(u.internal.size) < max_size) {
```
avikivity, line 358: Should add padding() here, see the constructor from initialized_later.

tests/unit/sstring_test.cc
```diff
@@ -199,6 +199,27 @@ BOOST_AUTO_TEST_CASE(test_append) {
     BOOST_REQUIRE_EQUAL(sstring("aba").append("1234", 0), "aba");
 }
 
+BOOST_AUTO_TEST_CASE(test_push_back) {
```
xemul, line 202: nit: push_back() against empty string looks like good corner case test

include/seastar/core/sstring.hh
```diff
@@ -346,6 +346,29 @@ public:
...
+                u.internal.str[u.internal.size++] = ch;
+                if (NulTerminate) {
+                    u.internal.str[u.internal.size] = '\0';
```
nyh, line 361: What if at this point u.internal.size already reached max_size?

include/seastar/core/sstring.hh
```diff
@@ -346,6 +346,29 @@ public:
...
+                }
+                return;
+            }
+        }
+        basic_sstring new_str(initialized_later(), u.external.size + 1);
+        std::copy(begin(), end(), new_str.begin());
+        new_str.u.external.str[new_str.u.external.size - 1] = ch;
+        *this = std::move(new_str);
```
nyh, line 369: I am not familiar with NulTerminate but don't we need to do it here as well?

include/seastar/core/sstring.hh
```diff
@@ -346,6 +346,29 @@ public:
...
+        if (is_internal()) {
+            if (static_cast<Size>(u.internal.size) < max_size) {
+                u.internal.str[u.internal.size++] = ch;
+                if (NulTerminate) {
+                    u.internal.str[u.internal.size] = '\0';
+                }
+                return;
+            }
+        }
+        basic_sstring new_str(initialized_later(), u.external.size + 1);
```
nyh, line 366: What if we reached here despite is_internal() (because we reached max size)? Isn't u.external.size wrong?
Please make sure you tests check the case of an empty string growing and growing until passing the internal size and see it still works.

tests/unit/sstring_test.cc
```diff
@@ -199,6 +199,27 @@ BOOST_AUTO_TEST_CASE(test_append) {
...
+    {
+        // append in internal storage
+        basic_sstring<char, uint32_t, 15> s("0123456789");
+        s.push_back('a');
+        BOOST_REQUIRE_EQUAL(s, "0123456789a");
+    }
+    {
+        // append causing spilling to external storage
```
nyh, line 210: I would do this test in a loop, for many sizes, not just 15 - e.g. from size 0 to 50. I don't know if 15 exactly is the cutoff point, or maybe 14 or 16? I'm worried that the above code has multiple bugs (I pointed out my suspicions above) and this test doesn't make me confident.
Also please check the NulTerminate option, whatever it is.

include/seastar/core/sstring.hh
```diff
@@ -346,6 +346,29 @@ public:
...
+        std::copy(begin(), end(), new_str.begin());
```
nyh, line 367: The documentation https://en.cppreference.com/w/cpp/string/basic_string/push_back says that push_back() should have "amortized constant" complexity. It means you cannot do this copy on every push_back() call - it would make it O(n) complexity. I think you need to multiply the size of the string by something (1.1 or 2 or whatever) and until you reach the allocated size, just copy one character and not the entire string.

nyh (changes requested):

include/seastar/core/sstring.hh
avikivity: We don't have unused capacity, so can't do that.

I think push_back is bad since it can't be implemented efficiently and will lead to accidentally quadratic users.

nyh: If we can't do that, we shouldn't implement this function. Its complexity is part of its API.

tchaikov: how about trading the implementation for a comment, noting that we don't intend to implement this because the performance issues due to the inherent design decision made by `sstring`?

avikivity: Nobody will read a comment on push_back. Everyone already knows how it works.

tchaikov: @avikivity hi Avi, i didn't mean to add a comment near by this function. i wanted to put a comment **in the place of it** explaining why we don't have it. to explain that it is a design decision.

PR #2110, opened 2024-02-21
shared_ptr: Generalize shared_ptr/make_shared to accept class with throwing dtor

niekbouman:
Currently, `seastar::make_shared` cannot be used with a polymorphic type whose base class' destructor is marked as `noexcept(false)` :

```cpp
#include <seastar/core/shared_ptr.hh>

class my_base {
public:
    virtual ~my_base() noexcept(false) = default;
};

class my_derived : public my_base {
public:
};

int main()
{
    auto x = seastar::make_shared<my_derived>();
    // currently fails
}
```
This commit suggests a fix for this problem.

Question is whether such a generalization of seastar::shared_ptr would be desirable in the first place.

### DIFF
--- a/include/seastar/core/shared_ptr.hh
+++ b/include/seastar/core/shared_ptr.hh
@@ -121,8 +121,31 @@ struct lw_shared_ptr_accessors_esft;
 template <class T>
 struct lw_shared_ptr_accessors_no_esft;
 
+template <typename T>
+struct is_complete_helper {
+    template <typename U>
+    static auto test(U*)  -> std::integral_constant<bool, sizeof(U) == sizeof(U)>;
+    static auto test(...) -> std::false_type;
+    using type = decltype(test((T*)0));
+};
+
+// test whether a type is complete
+template <typename T>
+struct is_complete : is_complete_helper<T>::type {};
 }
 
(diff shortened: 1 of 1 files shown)


niekbouman:
my implementation is broken, need to fix that first

avikivity:
What's the use case for a throwing destructor?

niekbouman:
> What's the use case for a throwing destructor?

Good question. The issue came up by trying to using a type (with a possibly throwing destructor) from a third party library (Cap'n Proto) with `seastar::shared_ptr`. The author of Cap'n Proto (Kenton Varda, also the main guy behind Cloudflare Workers) has an opinion on thowing destructors:

https://github.com/capnproto/capnproto/blob/master/style-guide.md#exceptions-can-happen-anywhere-including-destructors

Anyway, I now found a relevant [issue](https://github.com/capnproto/capnproto/issues/553) (on capnproto's github), that basically says that the C++ standard requires a type to have a nothrow-dtor for use with shared_ptr:

https://eel.is/c++draft/util.smartptr.shared.const#5

So having found/learned that, I will close this PR, and will use the 'black-box' DestructorCatcher solution instead.

PR #2115, opened 2024-02-24
circleci: quote literal in condition with `"`

tchaikov:
this should bring the checkheader back to the CI

the regression was introduced by e3fce16653


tchaikov:
no, checkheaders has been built.

PR #2118, opened 2024-02-25
src/core/reactor.cc: makes the metrics registration generic

amnonh:
This patch holds no functionality changes. It aligns how the metrics in reactor.cc are registered with the generic way (using namespace alias).

This is important when grepping for a metric.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1325,7 +1325,7 @@ cpu_stall_detector_linux_perf_event::arm_timer() {
     auto period = _threshold * _report_at + _slack;
     uint64_t ns =  period / 1ns;
     _next_signal_time = reactor::now() + period;
-    
+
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -2649,30 +2649,31 @@ void reactor::register_metrics() {
             sm::make_counter("abandoned_failed_futures", _abandoned_failed_futures, sm::description("Total number of abandoned failed futures, futures destroyed while still containing an exception")),
     });
 
-    using namespace seastar::metrics;
+
```
tchaikov, line 2652: might want to remove this empty line.

also

> makes the metrics registration generic

could be confusing, as generic could mean different things in different contexts. could you use something like 

> qualify metric function calls with "sm::"

?

amnonh: forced push v2 with the changes

tchaikov (approved):
lgtm

PR #2120, opened 2024-02-26
seastar-json2code: extract Parameter class and cleanups

tchaikov:
this change paves the road to https://github.com/scylladb/seastar/issues/2082. so that we can store more
properties in the Parameter class in this python script and then
translate them into a C++ class which is used to verify and expose
the content and schema of a parameter. so that they can be used
in `handler_base::verify_mandatory_params()` in a follow-up
change.

Refs https://github.com/scylladb/seastar/issues/2082

### DIFF
--- a/scripts/seastar-json2code.py
+++ b/scripts/seastar-json2code.py
@@ -164,12 +164,39 @@ def clear_path_ending(path):
         return path
     return path[0:-1]
 
-# check if a parameter is query required.
-# It will return true if the required flag is set
-# and if it is a query parameter, both swagger 1.2 'paramType' and swagger 2.0 'in' attributes
-# are supported
-def is_required_query_param(param):
-    return "required" in param and param["required"] and ("paramType" in param and param["paramType"] == "query" or "in" in param and param["in"] == "query")
+
+class Parameter:
+    '''represents a parameter
+
+    TODO: return an instance of Parameter from get_parameter_by_name()
+    '''
+    def __init__(self, definition):
+        self.definition = definition
+
+    @property
+    def name(self):
+        return self.definition['name']
+
+    @property
+    def is_required(self):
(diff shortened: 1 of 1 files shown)


tchaikov:
@amnonh hi Amnon, could you help review this change?

tchaikov:
@amnonh hi Amnon, could you help review this change?

scripts/seastar-json2code.py
```diff
@@ -316,18 +343,19 @@ def add_operation(hfile, ccfile, path, oper):
...
-        if is_required_query_param(param):
-            required_query_params.append(param["name"])
-        if "enum" in param:
-            enum_decl, parse_func = generate_code_from_enum(oper["nickname"],
-                                                            param["name"],
-                                                            param["enum"])
+        query_param = Parameter(param)
+        if query_param.is_required:
+            required_query_params.append(query_param)
+        if query_param.enum is not None:
```
amnonh, line 349: why the is not none?

amnonh (approved):
beside the nitpick about the if statement, it looks good

scripts/seastar-json2code.py
tchaikov: @amnonh i am able to parse your question. do you mean "when it is not None"? or do you mean "why shall we check if `query_param.enum` is not None"?

amnonh: why not:
if query_param.enum:

tchaikov: i see. i wanted to be very specific. as this property is `None` or "enum" is in its definition, or a list of string otherwise. an empty list should probably be handled in `generate_code_from_enum()`, and an error should be reported instead of being ignored. if i had more bandwidth, i would like to add typing annotations to this script.

tchaikov:

Issue #725, opened 2020-02-23
http server is missing support for percent decoding


nyh:
When an HTTP client encodes a URL, it usually encodes each of its components and subcomponents - **separately** - using so-called "percent encoding" (see https://en.wikipedia.org/wiki/Percent-encoding or the detailed definition in https://tools.ietf.org/html/rfc3986#section-2.1). Each character inside one of the subcomponents (e.g. one of the query parameters) which isn't alphanumeric, -, ., _ or ~ - is usually replaced by "%" and the two-byte hexadecimal representation of that character.

Seastar's HTTP server should do the reverse transformation, percent decoding, before passing the URL components (the path, query parameters, etc.) to the application code. But Seastar's HTTP server is missing this decoding part.

As an example of bugs this causes, see the Scylla bug https://github.com/scylladb/scylla/issues/5883. In that bug, the client fetched the URL "/column_family/metrics/write_latency/moving_average_histogram/a#usertable:usertable", which was encoded: The client changed the "#" into "%23", as it should (it could have changed also the ":" but decided not too). The Scylla code got it as "a%23usertable:usertable"  and reported that it couldn't find the table "a%23usertable:usertable" - which of course doesn't exist, the real table name is "a#usertable:usertable".

Another Scylla bug caused by this is https://github.com/scylladb/scylla/issues/9614.

fruch:
here's this example as a unit test:
```
diff --git a/tests/unit/httpd_test.cc b/tests/unit/httpd_test.cc
index aeeaa304..2991b7df 100644
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -72,6 +72,10 @@ SEASTAR_TEST_CASE(test_match_rule)
     res = mr.get("/hell/val1", param);
     httpd::handler_base* nl = nullptr;
     BOOST_REQUIRE_EQUAL(res, nl);
+
+    res = mr.get("/hello/a%23usertable", param);
+    BOOST_REQUIRE_EQUAL(param["param"], "a#usertable");
+
     return make_ready_future<>();
 }
```

would using https://github.com/cpp-netlib/url is solid enough ? 
seems like easy enough contribution for seastar, at least as my first one...

nyh:
[...]

PR #2125, opened 2024-03-01
http: path parameter parsing convenience methods

pgellert:
Some REST APIs of Redpanda were incorrectly using the `seastar::http::internal::url_decode` method to url decode path parameters, ignoring the subtle difference between the rules of query parameter decoding and path parameter decoding. This caused issues for Redpanda's clients as they were unable to delete users that had a '+' in the username.

As part of solving this decoding problem, we introduced a `path_decode` method similar to seastar's `url_decode`, and it would be most convenient if this method lived next to the existing `url_decode` utility method that seastar provides both to enable code reuse and to bring it to users' attention that they should make a conscious choice between the two decode methods.

To make this even easier, this PR also adds the `get_path_param` convenience method to the http request object, similar to `get_query_param`, to expose the decoded path parameters of the http request.

Related PR: https://github.com/redpanda-data/seastar/pull/97
Fixes https://github.com/scylladb/seastar/issues/725

### DIFF
--- a/include/seastar/http/common.hh
+++ b/include/seastar/http/common.hh
@@ -27,6 +27,7 @@
 
 #include <seastar/core/sstring.hh>
 #include <seastar/core/iostream.hh>
+#include <seastar/http/url.hh>
 
 namespace seastar {
 
@@ -45,6 +46,8 @@ namespace httpd {
 SEASTAR_MODULE_EXPORT_BEGIN
 
 class parameters {
+    // Note: the path matcher adds parameters with the '/' prefix into the params map (eg. "/param1"), and some getters
+    // remove this '/' to return just the path parameter value
     std::unordered_map<sstring, sstring> params;
 public:
     const sstring& path(const sstring& key) const {
@@ -57,6 +60,20 @@ public:
 
     const sstring& at(const sstring& key) const {
         return path(key);
+    }
+
+    sstring get_decoded_param(const sstring& key) const {
+        auto res = params.find(key);
+        if (res == params.end()) {
+            return "";
+        }
+        auto raw_path_param = res->second.substr(1);
+        auto decoded_path_param = sstring{};
+        auto ok = seastar::http::internal::path_decode(raw_path_param, decoded_path_param);
+        if (!ok) {
+            return "";
+        }
+        return decoded_path_param;
     }
 
     bool exists(const sstring& key) const {
--- a/include/seastar/http/common.hh
+++ b/include/seastar/http/common.hh
@@ -54,6 +54,7 @@ public:
         return params.at(key);
     }
 
+    [[deprecated("Use request::get_path_param() instead.")]]
     sstring operator[](const sstring& key) const {
         return params.at(key).substr(1);
     }
--- a/include/seastar/http/request.hh
+++ b/include/seastar/http/request.hh
@@ -109,12 +109,12 @@ struct request {
     }
 
     /**
-     * Search for the first header of a given name
-     * @param name the header name
-     * @return a pointer to the header value, if it exists or empty string
+     * Search for the last query parameter of a given key
+     * @param key the query paramerter key
+     * @return the query parameter value, if it exists or empty string
      */
-    sstring get_query_param(const sstring& name) const {
-        auto res = query_parameters.find(name);
+    sstring get_query_param(const sstring& key) const {
+        auto res = query_parameters.find(key);
         if (res == query_parameters.end()) {
             return "";
         }
@@ -277,7 +277,7 @@ struct request {
     static request make(httpd::operation_type type, sstring host, sstring path);
 
 private:
-    void add_param(const std::string_view& param);
+    void add_query_param(std::string_view param);
     sstring request_line() const;
     future<> write_request_headers(output_stream<char>& out) const;
     friend class experimental::connection;
--- a/include/seastar/http/request.hh
+++ b/include/seastar/http/request.hh
@@ -122,6 +122,16 @@ struct request {
     }
 
     /**
+     * Search for the last path parameter of a given key
+     * @param key the path paramerter key
+     * @return the unescaped path parameter value, if it exists and can be path decoded successfully, otherwise it
+     *  returns an empty string
+     */
+    sstring get_path_param(const sstring& key) const {
+        return param.get_decoded_param(key);
+    }
+
+    /**
      * Get the request protocol name. Can be either "http" or "https".
      */
     sstring get_protocol_name() const {
--- a/src/http/url.cc
+++ b/src/http/url.cc
@@ -49,7 +49,7 @@ short hex_to_byte(char c) {
 /**
  * Convert a hex encoded 2 bytes substring to char
  */
-char hexstr_to_char(const std::string_view& in, size_t from) {
+char hexstr_to_char(std::string_view in, size_t from) {
 
     return static_cast<char>(hex_to_byte(in[from]) * 16 + hex_to_byte(in[from + 1]));
 }
@@ -67,9 +67,7 @@ inline char char_to_hex(unsigned char val) {
     return "0123456789ABCDEF"[val];
 }
 
-}
-
-bool url_decode(const std::string_view& in, sstring& out) {
+bool decode(bool replace_plus, std::string_view in, sstring& out) {
     size_t pos = 0;
     sstring buff(in.length(), 0);
     for (size_t i = 0; i < in.length(); ++i) {
@@ -80,7 +78,7 @@ bool url_decode(const std::string_view& in, sstring& out) {
             } else {
                 return false;
             }
-        } else if (in[i] == '+') {
+        } else if (replace_plus && in[i] == '+') {
             buff[pos++] = ' ';
         } else {
             buff[pos++] = in[i];
@@ -91,7 +89,18 @@ bool url_decode(const std::string_view& in, sstring& out) {
     return true;
 }
 
-sstring url_encode(const std::string_view& in) {
+}
+
+bool url_decode(std::string_view in, sstring& out) {
+    return decode(true, in, out);
+}
+
+bool path_decode(std::string_view in, sstring& out) {
+    return decode(false, in, out);
+}
+
+
+sstring url_encode(std::string_view in) {
     size_t encodable_chars = 0;
     for (size_t i = 0; i < in.length(); i++) {
         if (should_encode(in[i])) {
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -65,7 +65,7 @@ SEASTAR_TEST_CASE(test_param_matcher)
     parameters param;
     BOOST_REQUIRE_EQUAL(m.match("/abc/hello", 4, param), 10u);
     BOOST_REQUIRE_EQUAL(param.path("param"), "/hello");
-    BOOST_REQUIRE_EQUAL(param["param"], "hello");
+    BOOST_REQUIRE_EQUAL(param.get_decoded_param("param"), "hello");
     return make_ready_future<>();
 }
 
@@ -78,7 +78,7 @@ SEASTAR_TEST_CASE(test_match_rule)
     mr.add_str("/hello").add_param("param");
     httpd::handler_base* res = mr.get("/hello/val1", param);
     BOOST_REQUIRE_EQUAL(res, h);
-    BOOST_REQUIRE_EQUAL(param["param"], "val1");
+    BOOST_REQUIRE_EQUAL(param.get_decoded_param("param"), "val1");
     res = mr.get("/hell/val1", param);
     httpd::handler_base* nl = nullptr;
     BOOST_REQUIRE_EQUAL(res, nl);
@@ -295,27 +295,22 @@ SEASTAR_TEST_CASE(test_json_path) {
 
     path1.set(*route, [res1] (const_req req) {
         (*res1) = true;
-        BOOST_REQUIRE_EQUAL(req.param["param1"], "value1");
         BOOST_REQUIRE_EQUAL(req.get_path_param("param1"), "value1");
         return "";
     });
 
     path2.set(*route, [res2] (const_req req) {
         (*res2) = true;
-        BOOST_REQUIRE_EQUAL(req.param["param1"], "value4+value4%20value4");
         BOOST_REQUIRE_EQUAL(req.get_path_param("param1"), "value4+value4 value4");
 
-        BOOST_REQUIRE_EQUAL(req.param["param2"], "text4%2Btext4");
         BOOST_REQUIRE_EQUAL(req.get_path_param("param2"), "text4+text4");
         return "";
     });
 
     path3.set(*route, [res3] (const_req req) {
         (*res3) = true;
-        BOOST_REQUIRE_EQUAL(req.param["param1"], "value3");
         BOOST_REQUIRE_EQUAL(req.get_path_param("param1"), "value3");
 
-        BOOST_REQUIRE_EQUAL(req.param["param2"], "text2/text3");
         BOOST_REQUIRE_EQUAL(req.get_path_param("param2"), "text2/text3");
         return "";
     });
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -16,6 +16,7 @@
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
 #include "loopback_socket.hh"
+#include "seastar/http/reply.hh"
 #include <boost/algorithm/string.hpp>
 #include <seastar/core/thread.hh>
 #include <seastar/util/noncopyable_function.hh>
@@ -81,6 +82,12 @@ SEASTAR_TEST_CASE(test_match_rule)
     res = mr.get("/hell/val1", param);
     httpd::handler_base* nl = nullptr;
     BOOST_REQUIRE_EQUAL(res, nl);
+
+    // Test that URL decoding happens correctly on the path part of
+    // the URL. Reproduces issue #725.
+    res = mr.get("/hello/hello%21hi", param);
+    BOOST_REQUIRE_EQUAL(param.get_decoded_param("param"), "hello!hi");
+
     return make_ready_future<>();
 }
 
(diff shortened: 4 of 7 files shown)


include/seastar/http/request.hh
```diff
@@ -277,7 +277,7 @@ struct request {
     static request make(httpd::operation_type type, sstring host, sstring path);
 
 private:
-    void add_param(const std::string_view& param);
+    void add_query_param(const std::string_view& param);
```
xemul, line 301: Can do `s/const std::string_view&/std::string_view/` in the same change as well

src/http/url.cc
```diff
@@ -67,8 +67,7 @@ inline char char_to_hex(unsigned char val) {
     return "0123456789ABCDEF"[val];
 }
 
-template<bool ReplacePlus>
-bool decode(const std::string_view& in, sstring& out) {
+bool decode(bool replace_plus, const std::string_view& in, sstring& out) {
```
xemul, line 70: This function was made a template in this PR, so either keep it as template, or make it accept `replace_plus` boolean argument in the 1st patch (the one that made it a template)

tests/unit/httpd_test.cc
```diff
@@ -205,6 +205,24 @@ SEASTAR_TEST_CASE(test_decode_url) {
     return make_ready_future<>();
 }
 
+SEASTAR_TEST_CASE(test_decode_path) {
```
xemul, line 208: I would say that's not enough. There's a `test_json_path` test that validates how parameters matching works, I think _that_ test should be extended to validate the new change

include/seastar/http/request.hh
```diff
@@ -121,6 +122,26 @@ struct request {
         return res->second;
     }
 
+    /**
+     * Search for the last path parameter of a given key
+     * @param key the path paramerter key
+     * @return the unescaped path parameter value, if it exists and can be path decoded successfully, otherwise it
+     *  returns an empty string
+     */
+    sstring get_path_param(const sstring& key) const {
```
xemul, line 131: This method looks to better belong to `class parameters` rather than to request itself

xemul:
Some commits in this PR contain references to other commits that don't belong to this repo (e.g. the `(cherry picked from commit 51b0584)` one). Please, drop them

xemul (changes requested):
Had some review comments

pgellert:
Thanks for the quick review @xemul. I now [force-pushed](https://github.com/scylladb/seastar/compare/04827cd55bef9b6a1192674b4066ae7d45b920fd..79c509e748f20e52726dbb0eb8ef83b1621c24a9) with the following changes based on feedback:
* Squashed the first 2 commits
* Removed the cherry-picking comments
* Moved ahead the change to use the boolean replace_plus argument instead of using the template argument to be consistent across the PR
* Moved the get_decoded_param implementation into common.hh
* Adapted the test_json_path unit tests to test get_decoded_param as well
* __The path matcher adds arguments into the internal map of params with a ‘/‘ prefix, so I fixed get_decoded_param to remove this from the return value to match the behaviour of operator[].__

tests/unit/httpd_test.cc
pgellert: Thanks, this made me realise that the path matcher adds arguments into the internal map of params with a ‘/‘ prefix, so I fixed get_decoded_param to remove this from the return value to match the behaviour of operator[].

I've updated the `test_json_path` unit test, let me know if there's anything else you had in mind here.

include/seastar/http/request.hh
xemul: This suggestion is still not applied

include/seastar/http/common.hh
```diff
@@ -59,6 +62,20 @@ public:
         return path(key);
     }
 
+    sstring get_decoded_param(const sstring& key) const {
+        auto res = params.find(key);
+        if (res == params.end()) {
+            return "";
+        }
+        auto raw_path_param = res->second.substr(1);
```
xemul, line 70: Why substr(1)?

pgellert: The path matcher adds arguments into the internal map of params with a ‘/‘ prefix, so I made `get_decoded_param` remove this from the return value to match the behaviour of operator[] above on line 55.

I expect that in most cases people use `params["param-key"]` and expect `params.get_decoded_param("param-key")` to return the same parameter value, decoded, without the `/` at the front. I am wondering though whether we should replace the first 5 lines of `get_decoded_param`:
```
auto res = params.find(key);
if (res == params.end()) {
    return "";
}
auto raw_path_param = res->second.substr(1);
```
with 
```auto raw_path_param = operator[](key);```
to match the error behaviour in the case that the key doesn't exist.

File: include/seastar/http/request.hh (hunk 1 above)
Comment (pgellert) [reply include/seastar/http/request.hh]: Sorry, I first thought that you meant to ask for me to squash the first 2 commits, the one introducing the `path_decode` function and the one that did the `s/const std::string_view&/std::string_view/` for `url.hh` and `url.cc`.

But I think you meant that I should do the same `s/const std::string_view&/std::string_view/` in `request.cc` and `request.hh` as well. I've done that now (see [force-push](https://github.com/scylladb/seastar/compare/79c509e748f20e52726dbb0eb8ef83b1621c24a9..d03443606044ecb0978a290461aaa9766f0cd01b)).

**Comment by pgellert** [comment]
[Force-pushed](https://github.com/scylladb/seastar/compare/79c509e748f20e52726dbb0eb8ef83b1621c24a9..d03443606044ecb0978a290461aaa9766f0cd01b) to do `s/const std::string_view&/std::string_view/` for `request.hh` and `request.cc` as well.

**Comment by pgellert** [comment]
@xemul I'd appreciate another look at this when you have a chance. I would still like to get this merged if possible.

File: include/seastar/http/common.hh (hunk 5 above)
Comment (xemul) [reply include/seastar/http/common.hh:70]: The more I look at this get_decoded_param() the less I like it.

Why would someone want to _avoid_ using this method and grab req.params directly? Isn't it more correct for the param matcher to decode the paths while parsing and put req.params already parsed?

Comment (pgellert) [reply include/seastar/http/common.hh:70]: > But how do we know it? :)

Whoever owns the client and server implementation pair knows it. :) Seastar's http router doesn't know it on its own. 

> The only problem will happen in case client sends a path that needs to be en/de-coded, the path arrives encoded, but server handler explicitly wants un-decoded path parameter. Am I missing something else?

Decoding more than necessary can also be a problem. Two examples I can think of are:
(a) Let's say the API endpoint is `example.com/labels/{some-custom-label}` where `some-custom-label` is a path parameter that we expect to be an arbitrary string defined by some end user. Let's also say that today the client doesn't path encode the parameter and the server just uses the raw `req.param` to get the label. We could have one of the custom labels be `DISCOUNT%20` that some end-user defined. Now, suppose we change seastar to decode path parameters by default. In that case, it silently changes the behaviour in the server as it will now consider this label as `DISCOUNT ` instead of `DISCOUNT%20` as the customer intended.
(b) Similarly, you could have the same API but in this case the client *does* path encode and the server also path decodes `req.param` because seastar had not path decoded `req.param` in the past. Now again, if the end-user gives the label `DISCOUNT%20` to the client, the client will path encode it and call the endpoint with `example.com/labels/DISCOUNT%2520`. If seastar now path decodes the parameter, it will set `req.param` with `DISCOUNT%20`. After that, the server will still do its path decoding leading to `DISCOUNT ` again.

So that is why I think it would be safer to keep the current behaviour of `req.param` and expose path decoded parameters as a separate `get_path_param` method.

Comment (xemul) [reply include/seastar/http/common.hh:70]: > Let's also say that today the client doesn't path encode the parameter ...

I think it can be treated as buggy client, and we can assume that paths coming from client are always encoded. I checked the `request::parse_query_param()` -- it assumes that query parameters are encoded and decodes them inside `add_query_param()`, but it keeps path as-is

> We could have one of the custom labels be DISCOUNT%20 that some end-user defined. Now, suppose we change seastar to decode path parameters by default. In that case, it silently changes the behaviour in the server as it will now consider this label as DISCOUNT instead of DISCOUNT%20 as the customer intended.

Wait, wait, the `DISCOUNT%20` is the as-is string passed by the client? In that case, it must be a "buggy client" right?

> (b) Similarly, you could have the same API but in this case the client does path encode and the server also path decodes req.param because seastar had not path decoded req.param in the past. Now again, if the end-user gives the label DISCOUNT%20 to the client, the client will path encode it and call the endpoint with example.com/labels/DISCOUNT%2520. If seastar now path decodes the parameter, it will set req.param with DISCOUNT%20. After that, the server will still do its path decoding leading to DISCOUNT again.

Yeah, double-decoding can be very surprising... In that case we need to think how to deprecate the old behavior. I mean:
- current API for creating/registering path_descriptions is marked as deprecated
- new way to create path appears
- when path is registered with new API, `parse_query_param()` always decodes it before processing

Comment (nyh) [reply include/seastar/http/common.hh:70]: I completely missed this PR... While rediscovering the missing path decoding, I also discovered the '+' bug - it appears somebody incorrectly added this HTML-only extension to the generic decoding function, which isn't a great idea.
I'm not working today so I can't review this PR today, but I'll do it on Sunday. Please review mine also.

Comment (nyh) [reply include/seastar/http/common.hh:70]: please also refer to the existing open Seastar issue.

Comment (nyh) [reply include/seastar/http/common.hh:70]: Namely, https://github.com/scylladb/seastar/issues/725

File: include/seastar/http/common.hh [hunk 6]
```diff
@@ -45,6 +46,8 @@ namespace httpd {
 SEASTAR_MODULE_EXPORT_BEGIN
 
 class parameters {
+    // Note: the path matcher adds parameters with the '/' prefix into the params map (eg. "/param1"), and some getters
```
nyh, line 49: I'm afraid I don't understand this patch at all. As I also noticed and fixed in my own "version" of your patch, https://github.com/scylladb/seastar/pull/2125 (I didn't see your PR when I wrote mine), we're indeed missing the decoding of the path. So my patch just added the missing decoding call, and that's it!
Why did you patch become so much more complicated?
If your PR adds new functionality, please put it in a separate PR or at least a separate patch in the same PR, with justification for the new functionality. This patch should have just had the call to decode_path() in the place (actually two places) it was previously missing.
Or maybe I just misunderstood this patch?

include/seastar/http/common.hh
nyh: I also don't understand the need for the new function or the talk about "double decoding". The old API for query parameters (NOT the path) worked, why do we need to change it? What am I missing?

tests/unit/httpd_test.cc
nyh: Please take a look at the tests I wrote in https://github.com/scylladb/seastar/pull/2125 and consider adopting them (they are almost trivial, I think there's one case your test doesn't cover).
It's "too easy" to write a super-low-level unit test that calls your new function and checks it does what you wrote in it. It's more interesting to use encoded parameters in the higher-up APIs and see that the correct thing happens to them.
I also have a very-high-level test in ScyllaDB that uses the API and checks it works as it should, https://github.com/scylladb/scylladb/pull/18091, but that will be a test that I will run later - it can't be included in this PR obviously. I hope you also have your own high-level tests in Redpanda which reproduced the original bug and show that your fix is correct.

tests/unit/httpd_test.cc
```diff
@@ -268,28 +286,35 @@ SEASTAR_TEST_CASE(test_json_path) {
...
 
     path2.set(*route, [res2] (const_req req) {
         (*res2) = true;
-        BOOST_REQUIRE_EQUAL(req.param["param1"], "value2");
-        BOOST_REQUIRE_EQUAL(req.param["param2"], "text1");
+        BOOST_REQUIRE_EQUAL(req.param["param1"], "value4+value4%20value4");
+        BOOST_REQUIRE_EQUAL(req.get_path_param("param1"), "value4+value4 value4");
+
+        BOOST_REQUIRE_EQUAL(req.param["param2"], "text4%2Btext4");
+        BOOST_REQUIRE_EQUAL(req.get_path_param("param2"), "text4+text4");
```
nyh, line 299: So I don't understand why this is useful. Unlike my patch in https://github.com/scylladb/seastar/pull/2162, here you decided not to fix `param["param2"]` and instead provide a new, fixed, function get_path_param. Why? When will it **ever** make sense to use the unfixed one?

nyh (changes requested):

tests/unit/httpd_test.cc
nyh: In particular, IIRC (I'll need to confirm this), this broken param[] API is exactly what ScyllaDB used and led to the bug https://github.com/scylladb/scylladb/pull/18091 where an undecoded path was used. So in this patch you leave the broken API and only users who remember to switch to the new API will get the fix. We will see this bug 5 years from now as new users will use the old API and wonder why it's broken. So, why should we leave a broken API and not fix existing software already using the old API and unknowingly have this bug?

pgellert: I have now deprecated the old API and adopted some of the tests from your PR.

However, I did not port the path decoding inside `parse_query_param` for now. This changes the url seastar uses for routing to be the decoded url. I am curious if you think that we should fix this in place (changing the existing http routing behaviour) or if we should keep that as is. If we do path decode the return value of `parse_query_param`, some further changes are needed to pass both the raw and the decoded urls into `routes::handle` to allow us to keep the raw path parameters in `req->param`. Let me know what you think.

Also, I haven't yet made optimized for the decoded (non-buggy) use case yet. `request::get_path_param` is still suboptimal compared to `param[]` as it requires path decoding on each access. For this, I was thinking that we could either introduce a second `httpd::parameters decoded_params;` inside `http::request` or have a second `unordered_map` inside `http::parameters`. Let me know if you have a preference for either.

pgellert: Also, I left this comment on your PR which is related to why I haven't yet implemented the path decoding inside `parse_query_param`: https://github.com/scylladb/seastar/pull/2162#discussion_r1551811108

include/seastar/http/common.hh
nyh: I just realized that operator[] also has that funky `substr(1)`. I still don't understand why it's done like this, but at least it's not some mistake - this is how it always was :-(

tests/unit/httpd_test.cc
nyh: I don't remember now if I changed that routing code because it was exercised in a test, or because I needed in the Scylla use case. I'll test your patch later with my Scylla reproducer.
I think that most Web servers when they get the URL `/some/dir/file%20name` will treat it just like `/some/dir/file name` and it should also work for "routing" - if it doesn't the first user to try to use this feature with encoded characters will be suprised :-(
I guess you're worried about backward compatibility with old code that might want the old unencoded path?

nyh (commented):
Thanks, looks good, but I'm afraid I am not sure what to do about this routing thing, I want to at least verify that even without it, the Scylla bug (that caused me to look into this issue) is fixed.
I'm also a bit worried that if we leave the routing thing unfixed, a few years down the line, someone is bound to run into that bug too... Or am I misunderstanding someting?

tests/unit/httpd_test.cc
pgellert: > I guess you're worried about backward compatibility with old code that might want the old unencoded path?

Yes, I am unsure if I can have such a breaking change in a patch version change or if it needs to be gated with `SEASTAR_API_LEVEL` or similar. But you know better than me about this, so if you think we can fix this in place, I am happy to make that change as well.

But for redpanda, no, I do not think routing on decoded urls will cause any issues after looking through the API we expose.

However, if we want to keep exposing undecoded path parameters with `param[]` while starting to route based on the decoded path, the implementation will be a bit more complex as both the decoded and undecoded url will have to be passed down from `connection::generate_reply` all the way to `param_matcher::match`. But perhaps that's the least of our concerns here.

mykaul:
What's the next step here?

pgellert:
@nyh have you had a chance to test this change with the ScyllaDB bug you mentioned? Let me know if there is anything else I can help with.

nyh:
> @nyh have you had a chance to test this change with the ScyllaDB bug you mentioned? Let me know if there is anything else I can help with.

Unlike my own patch which just fixed the problem for existing code (assuming applications didn't try to fix the same problem on their own), your patch resulted in a dozen Scylla source files getting deprecation warnings and still (as expected, deliberately) the bug doesn't get fixed. I'll try to change all those callers and check if it helps.

nyh:
@pgellert did you deliberately want the user to have to convert `req.param["name"]` to `req.param.get_decoded_param("name")` with the word "param" twice? Wouldn't it make more sense to call the method `req.get_decoded_param()`, not `req.param.get_decoded_param()`?

nyh:
> @pgellert did you deliberately want the user to have to convert `req.param["name"]` to `req.param.get_decoded_param("name")` with the word "param" twice? Wouldn't it make more sense to call the method `req.get_decoded_param()`, not `req.param.get_decoded_param()`?

In particular, here is a snippet from Scylla code after your patch:

```
        auto keyspace = req->param.get_decoded_param("keyspace");
        auto table = req->get_query_param("table");
 ```

It is unclear what's the difference between get_query_param and get_decoded_param, and why one operates on req and the other on req->param.

**Comment by nyh** [comment]
> @nyh have you had a chance to test this change with the ScyllaDB bug you mentioned? Let me know if there is anything else I can help with.

The good news is that your fatch fixes the bug.
The bad news is that I had to fix 104 (!!) callers in ScyllaDB from using `req.param["something"]` to `req.param.get_decoded_param("something")` to fix the bug (and remove deprecation warnings) in all callers.

I guess we can do that. But please give a thought to the naming question I raised above -  why we have `req.get_query_param()` and don't call the other one `req.something()` (perhaps `req.get_path_param()`?) and instead need to use param explicitly in `req.param.get_decoded_param()` (and isn't this "decoded param" specific for path params?).

**Comment by nyh** [comment]
> I guess we can do that. But please give a thought to the naming question I raised above - why we have `req.get_query_param()` and don't call the other one `req.something()` (perhaps `req.get_path_param()`?) and instead need to use param explicitly in `req.param.get_decoded_param()` (and isn't this "decoded param" specific for path params?).

Wow, it turns out you did exactly what I wished for - req.get_path_param() :-) I'll start using that. Sorry about the spam.

**Review by nyh — APPROVED** [review]
Ok, I'm "approving" this approach. I have a request about the deprecation message which misled me, and it will give a lot of headache to users who will need to resolve all these deprecations (in the Scylla, there were 104 of them...), but if we don't want any chance of backward incompatibility, I guess we need to do this.

**Comment by nyh** [comment]
Also please add "Fixes #725" to your commit message to have this Seastar issue be closed when your PR is merged.

File: include/seastar/http/common.hh [hunk 8]
```diff
@@ -54,6 +54,7 @@ public:
         return params.at(key);
     }
 
+    [[deprecated("Use get_decoded_param() instead.")]]
```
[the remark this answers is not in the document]
pgellert: Makes sense :) I've now updated the deprecation warning: https://github.com/scylladb/seastar/compare/9907c372439ad4e7130bc4b7d93f849129fd79cb..d15c6a3862c0af82c622d802cc9b46b45473d5b6

nyh (approved):

PR #2126, opened 2024-03-01
reactor: Remove write-only task_queue._current

xemul:
It's set when reactor start running tasksk from that queue and unset when reactor stops doing it. The bit is unly used by scylla-gdb script, but active/inactive status of the queue can be found out with the help of current_scheduling_group_ptr


avikivity:
Since 98630451a2b97ef37e95895ae2f4a43fa7017d92

PR #2129, opened 2024-03-03
src/core: revise some grammar warnings

sqddk:
clang-tidy has raised syntax warnings for some existing code, such as using pass-by-value where pass-by-reference could be used, or renaming local variables. I have carefully revised these sections.

### DIFF
--- a/include/seastar/core/resource.hh
+++ b/include/seastar/core/resource.hh
@@ -117,7 +117,7 @@ struct io_queue_topology {
 
     io_queue_topology();
     io_queue_topology(const io_queue_topology&) = delete;
-    io_queue_topology(io_queue_topology&&);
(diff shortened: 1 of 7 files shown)


nyh:
I see you closed this PR, why? If you plan to send a new version, you don't need to close the PR - you can just force-push a new version of your github branch, and the PR will automatically pick up the new version.

Anyway, if you do plan to send a new version it will be helpful if you don't bunch together 5 different fixes just because the same tool found all of them, and if possible please split the changes into several patches (in the same PR), with an explanation on each one. For example, I noticed one of the changes is to not bother to use "std::move()" on a pointer, because it is both unnecessary and misleading (it makes it sound like the move does something special). You can bunch all the changes of the same type into one patch, and explain it.

PR #2130, opened 2024-03-05
metrics: support relabel aggregate

qiuqiu-007:
inorder to aggregate metrics to node (or others) level, here add an optional to action, which is "aggregate_label". the usage is similar to action "drop_label". For example:
```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '[ \
   { \
     "source_labels": [ \
       "__name__" \
     ], \
     "action": "aggregate_label", \
     "target_label": "shard", \
     "regex": "column_family_live_sstable" \
   } \
 ]' 'http://localhost:10000/v2/metrics-config/'
```
Signed-off-by: qiulijuan2<qiulijuan2_yewu@cmss.chinamobile.com>

### DIFF
--- a/include/seastar/core/metrics_api.hh
+++ b/include/seastar/core/metrics_api.hh
@@ -192,6 +192,7 @@ struct metric_info {
     labels_type original_labels;
     bool enabled;
     skip_when_empty should_skip_when_empty;
+    std::vector<std::string> aggregate_labels;
 };
 
 
@@ -216,7 +217,7 @@ class registered_metric final {
     metric_function _f;
     shared_ptr<impl> _impl;
 public:
-    registered_metric(metric_id id, metric_function f, bool enabled=true, skip_when_empty skip=skip_when_empty::no);
+    registered_metric(metric_id id, metric_function f, bool enabled=true, skip_when_empty skip=skip_when_empty::no, std::vector<std::string> aggregate_labels = {});
     metric_value operator()() const {
         return _f();
     }
--- a/include/seastar/core/relabel_config.hh
+++ b/include/seastar/core/relabel_config.hh
@@ -89,7 +89,7 @@ public:
  *
  */
 struct relabel_config {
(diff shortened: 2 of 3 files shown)


amnonh:
I'm not sure if aggregate should be done on a metric level or on a metric family level, see my pr #2121

qiuqiu-007:
> I'm not sure if aggregate should be done on a metric level or on a metric family level, see my pr #2121

In my opinion, first of all, this way is convenient to set a single metric or a group metrics to be aggregated, so just add a function to relabel_config, secondly, it changes the minimum range of code.

amnonh:
Most of the time, aggregation is on the metric family level, a single instance cannot be aggregated, which which metrics?
As far as I can tell, the code itself just adds a list of aggregate labels to a metric, but without changing the prometheus code it does not add any functionality.

Please do not merge

qiuqiu-007:
In prometheus.cc, it seems that aggregate operation is done during http request ? And in registration time, it just sends the aggregate_labels arg to stored in  map , right? @amnonh

amnonh:
Prometheus takes the relabel from the metrics family, that's why it make more sense to put the configuration at that level

qiuqiu-007:
> Prometheus takes the relabel from the metrics family, that's why it make more sense to put the configuration at that level

Do u mean that add the aggregate_labels to struct of metric_info makes no sense, even thought send the aggregate_labels to the metric_family_info ?

qiuqiu-007:
-Here are two points I'm interested in :
-  if `update_aggregate_labels` will be release in newer Scylla version, if yes, which version ? 
- `skip_when_empty` is added to the relabel_config function, then why this `aggregate_labels` can not be an info added to metric_info? Both of them can be a member of metric label? What is the point to group metric level and metric_family level?
@amnonh

qiuqiu-007:
ping @amnonh I am eager to receive your reply.

Issue #2113, opened 2024-02-22
"undefined reference to `archive_read_new"


mickey298:
### os
fresh new vm
```bash
uname -a
Linux archlinuxtest 6.7.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 17 Feb 2024 14:02:33 +0000 x86_64 GNU/Linux
```
### seastar commit
f0b2971a24ab7716933a720572207fa9a1df38bd

---
configure and build
```bash
./configure.py --mode=release --enable-dpdk --prefix=/usr

ninja -C build/release
```
 error:
```bash
ninja: Entering directory `build/release'
[104/351] Linking CXX executable tests/perf/fstream_perf
FAILED: tests/perf/fstream_perf 
: && /usr/bin/g++ -O2 -g -DNDEBUG  tests/perf/CMakeFiles/test_perf_fstream.dir/fstream_perf.cc.o -o tests/perf/fstream_perf  libseastar.a  /usr/lib/libboost_program_options.so  /usr/lib/libboost_thread.so  /usr/lib/libboost_chrono.so  /usr/lib/libboost_atomic.so  /usr/lib/libcares.so  /usr/lib/libfmt.so.10.2.0  /usr/lib/liblz4.so  -ldl  /usr/lib/libgnutls.so  -latomic  /usr/lib/libsctp.so  /usr/lib/libprotobuf.so  /usr/lib/libyaml-cpp.so  /usr/lib/libhwloc.so  /usr/lib/liburing.so  /usr/lib/libnuma.so && :
/usr/bin/ld: libseastar.a(dpdk.o): in function `firmware_read':
eal_firmware.c:(.text+0x2c6f2): undefined reference to `archive_read_new'
/usr/bin/ld: eal_firmware.c:(.text+0x2c706): undefined reference to `archive_read_support_format_raw'
/usr/bin/ld: eal_firmware.c:(.text+0x2c716): undefined reference to `archive_read_support_filter_xz'
/usr/bin/ld: eal_firmware.c:(.text+0x2c72e): undefined reference to `archive_read_open_filename'
/usr/bin/ld: eal_firmware.c:(.text+0x2c741): undefined reference to `archive_read_next_header'
/usr/bin/ld: eal_firmware.c:(.text+0x2c795): undefined reference to `archive_read_data'
/usr/bin/ld: eal_firmware.c:(.text+0x2c7c2): undefined reference to `archive_read_free'
[...]

PR #2132, opened 2024-03-06
build: relax the version constraint for Protobuf

tchaikov:
as higher versions of Protobuf does not claim to be compatible with
v2.5.0, so we should not specify a version number which is acceptable
by Seastar as a parameter of `find_package()`. and instead, we just
check the version number after calling `find_package()` in config mode.
if we find the package using config mode, we search it again, for
printing out the version number to the stdout, otherwise we search
using its C++ API version.

Refs ##2113

### DIFF
--- a/cmake/SeastarDependencies.cmake
+++ b/cmake/SeastarDependencies.cmake
@@ -159,14 +159,11 @@ macro (seastar_find_dependencies)
   # ProtobufConfig.cmake provided by protobuf defines this linkage. so we try
   # the CMake package configuration file first, and fall back to CMake's
   # FindProtobuf module.
-  find_package (Protobuf 2.5.0
-    QUIET CONFIG)
-  if (Protobuf_FOUND)
+  find_package (Protobuf QUIET CONFIG)
+  if (Protobuf_FOUND AND Protobuf_VERSION VERSION_GREATER_EQUAL 2.5.0)
     # do it again, so the message is printed when the package is found
-    find_package(Protobuf 2.5.0
-      CONFIG REQUIRED)
+    find_package(Protobuf CONFIG REQUIRED)
   else ()
-    find_package(Protobuf 2.5.0
-      REQUIRED)
+    find_package(Protobuf 2.5.0 REQUIRED)
   endif ()
 endmacro ()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -417,10 +417,11 @@ function (seastar_generate_protobuf)
   set (source_out ${args_OUT_DIR}/${in_file_name}.pb.cc)
 
   add_custom_command (
-    DEPENDS ${args_IN_FILE}
+    DEPENDS ${args_IN_FILE} protobuf::protoc
     OUTPUT ${header_out} ${source_out}
     COMMAND ${CMAKE_COMMAND} -E make_directory ${args_OUT_DIR}
-    COMMAND ${Protobuf_PROTOC_EXECUTABLE} --cpp_out=${args_OUT_DIR} -I${in_file_dir} ${args_IN_FILE})
+    COMMAND protobuf::protoc
+    ARGS --cpp_out=${args_OUT_DIR} -I${in_file_dir} ${args_IN_FILE})
 
   add_custom_target (${args_TARGET}
     DEPENDS


cmake/SeastarDependencies.cmake
```diff
@@ -159,11 +159,11 @@ macro (seastar_find_dependencies)
   # ProtobufConfig.cmake provided by protobuf defines this linkage. so we try
   # the CMake package configuration file first, and fall back to CMake's
   # FindProtobuf module.
-  find_package (Protobuf 2.5.0
+  find_package (Protobuf 22.0...25.3
```
avikivity, line 162: I don't understand how 2.5.0 worked before and now it's 22.0..22.3. I understood there are two versioning systems, but you don't change versioning systems here.

avikivity: Also, does 26 not work?

mykaul:
Is it somehow aligned with the version Prometheus uses (CC @amnonh ) ?

cmake/SeastarDependencies.cmake
tchaikov: @avikivity 

> I don't understand how 2.5.0 worked before and now it's 22.0..22.3. I understood there are two versioning systems, but you don't change versioning systems here.

22.0 is the first version which uses abseil. so i wanted to make this explicit. and i didn't want to mix two versioning systems in a single condition, like `find_package (Protobuf 2.5...25.3)` even if it would yield the expected outcome. but it's just weird..

> Also, does 26 not work?

yeah, i asked myself when writing these numbers. but i wanted to avoid adding versions in future and wanted to avoid pretending that we support versions we haven't even compiled with. that's why it looks this way. but it didn't pan out. 

anyway, i am pushing an updated version.

tchaikov:
v2:

- do not enforce version constraints when finding protobuf in config mode, as it turns out protobuf's cmake configuration does not claim the compatibility with other versions part from its exact version.
- check the minimal required version of `2.5.0` instead of the minimal version which introduced abseil libraries linkage. simpler and more consistent this way.

tchaikov:
> Is it somehow aligned with the version Prometheus uses (CC @amnonh ) ?

@mykaul hi Yaniv, this change is intended to address the FTBFS on a very bleeding edge distro which

1. packages a very new protobuf
2. packages protobuf's CMake configuration file.

the metrics' on-wire protocol is defined by https://buf.build/gogo/protobuf/file/4df00b267f944190a229ce3695781e99:gogoproto/gogo.proto and https://github.com/prometheus/prometheus/blob/main/prompb/io/prometheus/client/metrics.proto. when it comes to the C++ language binding. i think protobuf compiler *should* be backward compatible in this regards.

tchaikov:
per @mickey298 's test at https://github.com/scylladb/seastar/issues/2113#issuecomment-1981213761 . the change does address the FTBFS.

tchaikov:
@scylladb/seastar-maint hello maintainers, could you review this change?

cmake/SeastarDependencies.cmake
avikivity: What I didn't understand is the version jump from 2 to 22. Was the numbering changed? Or do the version numbers refer to different things?

tchaikov: please note, in the latest revision i stopped using 22 as the version number. and sorry, i failed to explain my reasoning on the previous revision in an understandable way. let me rephrase it.

more like it refer to a different thing. 

actually i didn't bumped the version from 2 to 22. it was because, the CMake configuration file shipped by Protobuf uses the library's version, i.e., 22 and 23..., while the `FindProtobuf` module uses the language binding's version, like 2.5, 3.12, 3.21, 4.22, ....  .  the cutting edge distros are more likely to use the CMake configuration file shipped by Protobuf, so i intended to use the library's version when finding the package using the config mode, and i need to pick a library version . instead of trying to map 2.5 to "protobuf v5.0" which does not exst at all. i guess the reason is that it predates the version when protobuf developers decided to using this new versioning approach. so, i used v22, which is the first [version](https://github.com/protocolbuffers/protobuf/releases/tag/v22.0) which uses abseil. so that i can assure that the change works.

tchaikov: @avikivity ^

avikivity: ah, thanks.

PR #2134, opened 2024-03-10
util: add formatter for optimized_optional<>

tchaikov:
before this change, we rely on the default-generated fmt::formatter created from operator<<, but fmt v10 dropped the default-generated formatter.

in this change, we define formatters for `seastar::optimized_optional`, its operator<< is preserved for backward compatibility.

Refs scylladb/scylladb#13245

### DIFF
--- a/include/seastar/util/optimized_optional.hh
+++ b/include/seastar/util/optimized_optional.hh
@@ -28,6 +28,7 @@
 #include <iostream>
 #include <memory>
 #include <type_traits>
+#include <fmt/core.h>
 #endif
 
 namespace seastar {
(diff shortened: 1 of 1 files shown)


nyh (approved):
Looks good.

I see you didn't remove the old operator<<, is it because we have other instances of operator<<  that still use it?
I guess that when this whole fmt "project" is over, we can go over the entire source code and remove all the instances of operator<<.

nyh:
> I see you didn't remove the old operator<<, is it because we have other instances of operator<< that still use it? I guess that when this whole fmt "project" is over, we can go over the entire source code and remove all the instances of operator<<.

Hmm, sorry, I missed the **project** this PR is in. In **Seastar**, we need to support users who really want to use `operator<<`, not just those who want to use fmt. So we do need both.

tchaikov:
yeah, i don't dare to break the backward compatibility of Seastar unless i have the luxury of adding another API level.

PR #2135, opened 2024-03-10
Add zlib back to cooking ingredients

lixin-wei:
This reverts commit <https://github.com/scylladb/seastar/commit/03cf327720055a7d325a891d7bdc28ac7c060978>.

In that commit zlib was dropped. But Protobuf still depends on it. Which will make cooking.sh error.

This commit adds zlib back.

### DIFF
--- a/cooking_recipe.cmake
+++ b/cooking_recipe.cmake
@@ -119,6 +119,14 @@ cooking_ingredient (numactl
     BUILD_COMMAND <DISABLE>
     INSTALL_COMMAND ${make_command} install)
 
+cooking_ingredient (zlib
+  EXTERNAL_PROJECT_ARGS
+    URL https://github.com/madler/zlib/releases/download/v1.2.13/zlib-1.2.13.tar.gz
+    URL_MD5 9b8aa094c4e5765dabf4da391f00d15c
+    CONFIGURE_COMMAND <SOURCE_DIR>/configure --prefix=<INSTALL_DIR>
+    BUILD_COMMAND <DISABLE>
+    INSTALL_COMMAND ${make_command} install)
+
 ##
 ## Private and private/public dependencies.
 ##


cooking_recipe.cmake
```diff
@@ -194,6 +194,11 @@ cooking_ingredient (GnuTLS
     BUILD_COMMAND <DISABLE>
     INSTALL_COMMAND ${make_command} install)
 
+cooking_ingredient (zlib
+  EXTERNAL_PROJECT_ARGS
+    URL https://github.com/madler/zlib/releases/download/v1.3.1/zlib-1.3.1.tar.gz
+    URL_MD5 9855b6d802d7fe5b7bd5b196a2271655)
```
tchaikov, line 200: i dropped zlib in 03cf327720055a7d325a891d7bdc28ac7c060978, which introduced this regression. as a fix, you might want to revert it. and mention the offending commit in the commit message. and if you think it'd be nice to bump up the version. i think it's fine to implement it in the same commit. but please explain the reason why 1.3.1 is picked.

lixin-wei: Thanks for the review. I changed the version back. And mentionded the offending commit in the PR's description.

Would you please have a look again?

tchaikov:
please squash these two commits into a single one -- you could just use `git revert 03cf327720055a7d325a891d7bdc28ac7c060978` on top of master HEAD

and, please amend the commit message to explain why we should revert that change. please note, the cover letter of a pull request is not the same thing of a commit in that pull request.

lixin-wei:
@tchaikov Get it. All done 😀 Would you please have a look again?

BTW, why not just squash merging this PR, and set the commit message to PR's description? [This chrome plugin](https://chromewebstore.google.com/detail/squashed-merge-message/kgabfelbdpeifcemndligpdfddhbbmfm?pli=1) might be handy to automatically set the commit message to PR's description.

tchaikov (approved):
lgtm

tchaikov:
the way how the cover letter is included in the merged commit depends on how we merge a pull request. and it varies from one project to another. IIUC, unlike some other projects, seastar does not squash commits before merging their PR.

lixin-wei:
I see. Thanks for the explanation.

tchaikov:

nyh:
> @scylladb/seastar-maint hello maintainers, could you help review this change?

I can't review this patch - I still have no idea why we introduced or recommend this "cooking" technique, and also don't understand why Seastar requires protobuf (which should have been optional, no?) or zlib to be compiled. But since you (@tchaikov) reviewed it, and the patch is a revert of your earlier patch, I trust your review and merged the patch. Thanks.

lixin-wei:
@nyh It's for hermetic building. We have a Bazel build cluster which serves many projects. So I can't install dependency system wide, or other projects will be affected.

nyh:
> @nyh It's for hermetic building in my case. We have a Bazel build cluster which serves many projects. So I can't install dependency system wide, or other projects will be affected.

You can still download pre-compiled packages from your distro and open them to a non-root directory - or the more "modern" version of this is Docker. You don't have to compile all these packages from source. But anyway, if people like it, I'm not against - I just can't review the correctness of this because I don't know how to run it. But apparently @tchaikov knows and reviewed this patch, so good.

lixin-wei:
@nyh Yeah you are right. But our company want to make everything source-dependent😂so we can easily edit anything.

BTW, does Seastar willing to adopt another build system? I mean Bazel. It will unlock all Bazel users to Seastar.

I'm working on it and want to contribute it. Just wonder whether you're willing to accept it.😃

nyh:
> BTW, is Seastar willing to adopt another build system? I mean Bazel.
> 
> I'm working on it and want to contribute it. Just wonder whether you're willing to accept it.😃It will unlock all Bazel users to Seastar.

You are asking the wrong person. I personally believe it has been a big mistake by the community at large to think that there is any reason or any major benefit to everybody using the same build system. In the past it was considered the norm that some libraries were built with make, some with ninja, some cmake, some autotools, some imake, and more - and the different projects cooperated by building each one separately, and each finding the other project's build results. A good way to tie all those different build systems together was pkg-config, introduced in 2000. But this idea seems to no longer be "fashionable". Now the fashion is that cmake users believe that they can expect all the packages in the world to have been built with cmake, and bazel users expect all the backages to be built by Bazel :-(

Seastar actually switched to cmake not because of cmake's beauty (which I can't see, and I'm not sure who does...) but because of the belief that everybody was switching to cmake, and we must switch to too so that all the cmake users to be be able use Seastar. Now you are saying the same thing about Bazel :-)

In any case, if we ever want to switch Seastar's build system, the person who will be making this decision will be @avikivity.

lixin-wei:
@nyh Thanks for the detailed reply with the funny history 😄

To clarify I'm not trying to 'swich' the build system to Bazel. I just want to add another set of build scripts, alongside with cmake. You can still use cmake. @avikivity Are you comfortable with this?😃 If OK I'll open a PR to contribute.

nyh:
> To clarify I'm not trying to 'switch' the build system to Bazel. I just want to add another set of build scripts, alongside with cmake. You can still use cmake. @avikivity Are you comfortable with this?😃 If OK I'll open a PR to contribute.

Will this extra set of build scripts duplicate everything already in cmake? I'm afraid that if that will be the case, we'll need to remember (and likely will not remember) to update both of them every time we change one. We have something similar in the ScyllaDB project - where we have a cmake-based build system and one not based on cmake - and the two always get out of sync :-(

lixin-wei:
Yes they'll get out of sync. But has a 'used to work' version is better than nothing. If user find It's broken he can fix it. That's one charming thing of the open source community - user will help themselves.

PR #2138, opened 2024-03-11
memory.cc: Fix static init fiasco in system malloc override

### LINKED ISSUE #2137: Dependent library init can cause memory (re)allocation failures when test are run w/o `SEASTAR_DEFAULT_ALLOCATOR`
When `SEASTAR_DEFAULT_ALLOCATOR` is not active, we override `malloc, free, realloc` etc. 

If an application, say a database, loads dependent libraries and does some `dlinit` time initializations that cause memory to be allocated, then later in the lifetime of the process, do something else, which causes said memory to be free:d/realloced or whatnot,
we can run into angry asserts that we are on the wrong CPU if we don't ensure to retain all usage of said external libraries to shard 0. 

So far so good. 
But when running unit tests, `dlinit` will happen in test runner thread, not actual reactor thread 0. Even though reactor init is delayed into the test runner thread (effectively thread 1). 

The problem is that `cpu_mem`, which contains all info. is a thread local structure. Fair enough. But the `cpu_id` counter in it is assigned from an atomic counter, incremented on each init. This means that we end up with:

Test runner : cpu_mem{ cpu_id = 0 }
Reactor-0: cpu_mem{ cpu_id = 1 }
Reactor-1: cpu_mem{ cpu_id = 2 }

etc. Again, would not be a problem, except for the fact that the only way we have to ensure allocation from a third party that might combine/mess with data from a static init is handled on the same CPU, and we cannot in any meaningful way send this to the originating thread (which hogs cpu_id 0).

We do have a guard on malloc, checking if calling context is a reactor thread or not, and if not, we call back to iriginal libc malloc instead. _But_ the detection of such memory in subsequent realloc is apparently not fool proof. Because we get past if sometimes, when alloc is made before `original_malloc_func` is init:ed

Provoked via openssl usage, which exhibits this (not great) pattern for certain algo resolutions. (See `ossl_provider_set_operation_bit`). 
However, most dependencies, libc, boost etc allocs on dlinit.

elcallio:
Fixes #2137

When running without SEASTAR_DEFAULT_ALLOCATOR, i.e. when all allocs are overridden to seastar malloc etc, we try to detect allocations comming from outside reactor threads and redirect them to system alloc. However, allocs coming from dependency libraries during dlinit (i.e. static init) might preceed our resolving of said system alloc functions. This fiasco only really shows iff a process main thread is separate from normal reactor thread 0 (shard 0), since we tend to push all external dependecy allocs/changes to here.

But when running things like boost test, our own harness will run on thread 0, and reactor shard 0 will actualy be thread 1. This causes us to try to handle memory allocated via a `cpu_mem` thread local for a different thread -> crash.

While this does not handle any of the cases where external allocs are legitimately free:d/realloced on different threads, we can at least try to plug the static init fiasco so we have uniform behaviour regardless of reordered dlinit time allocations.

Fixed by double checking external function bindings once in malloc override.

### DIFF
--- a/src/core/memory.cc
+++ b/src/core/memory.cc
@@ -98,6 +98,7 @@ module;
 #include <mutex>
 #include <functional>
 #include <cstring>
+#include <utility>
 #include <boost/intrusive/list.hpp>
 #include <sys/mman.h>
 
@@ -1639,6 +1640,30 @@ void *allocate_slowpath(size_t size) {
             alloc_stats::increment(alloc_stats::types::foreign_mallocs);
             return original_malloc_func(size);
         }
+
+        // #2137 - static init fiasco for fallback functions. 
+        // If dependent libraries do malloc _before_ the above declaration inits are run,
+        // we end up here with nowhere to go. Add a second check and attempt the full init
+        // already. If we find the functions, all is good. Otherwise, trudge along and probably
+        // crash terribly later.
+        // Note: this is only relevant if we're in dlinit phase, in which case we are single
+        // threaded. Or there is no external func to find. No need for atomics or fences.
+        static bool double_checked = false;
+        if (!double_checked) {
+            double_checked = true;
+
+            original_malloc_func = reinterpret_cast<malloc_func_type>(dlsym(RTLD_NEXT, "malloc"));
+            original_free_func = reinterpret_cast<free_func_type>(dlsym(RTLD_NEXT, "free"));
+            original_realloc_func = reinterpret_cast<realloc_func_type>(dlsym(RTLD_NEXT, "realloc"));
+            original_aligned_alloc_func = reinterpret_cast<aligned_alloc_type>(dlsym(RTLD_NEXT, "aligned_alloc"));
+            original_malloc_trim_func = reinterpret_cast<malloc_trim_type>(dlsym(RTLD_NEXT, "malloc_trim"));
+            original_malloc_usable_size_func = reinterpret_cast<malloc_usable_size_type>(dlsym(RTLD_NEXT, "malloc_usable_size"));
+
+            if (original_malloc_func) {
+                return allocate_slowpath(size);
+            }
+        }
+
         // original_malloc_func might be null for allocations before main
         // in constructors before original_malloc_func ctor is called
         init_cpu_mem();


src/core/memory.cc
```diff
@@ -1639,6 +1639,26 @@ void *allocate_slowpath(size_t size) {
...
             return original_malloc_func(size);
         }
+
+        // #2137 - static init fiasco for fallback functions. 
+        // If dependent libraries do malloc _before_ the above declaration inits are run,
+        // we end up here with nowhere to go. Add a second check and attempt the full init
+        // already. If we find the functions, all is good. Otherwise, trudge along and probably
+        // crash terribly later.
+        static bool double_checked = false;
+        if (!std::exchange(double_checked, true)) {
```
avikivity, line 1649: This can be multi-threaded, so needs barriers.

I think we can only hit it when we are single-threaded, but at least move the write to within the if.

elcallio: Within the if? We are inside the if. 
We can only get here during dlinit phase, i.e. single threaded (assuming not loading seastar as a dynamic library, and in that case local allocator would not be permissible).
Seems pointless overkill to no either fence or atomics here.

avikivity: std::exchange() is executed outside the if, its store is unconditional.

xemul: Why is this double_checked needed at all? If it's false, we initialize original_... stuff, if it's true _and_ original_... is null, then why not re-initialize it again?

elcallio:
ping?

xemul:
I still don't get the point of double_checked protection, but otherwise looks good

elcallio:
Because the check `if (original_malloc_func)` exists and is followed by `init_cpu_mem()` and then falling back to allocating using seastar allocator here, we must assume that it is a valid linkage permutation that we do in fact _fail_ looking up `malloc` et al from libc. In that case, we should not repeat the dlsym calls over and over. 

It is beyond pointless to respect such a weird situation, but if there is one thing I know it is how to be pointless.

avikivity:
I'm good with the patch, @xemul please commit when satisfied

PR #2141, opened 2024-03-12
io_tester: implement unlinking via blocks discarding

pwrobelse:
This change is intended to implement a possibility
to select the implementation of unlink operation
used by unlink_request_type.

It will be used to conduct measurements related to
impact of unlinking on read and write operations.

### DIFF
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -70,6 +70,8 @@ static thread_local std::default_random_engine random_generator(random_seed);
 class context;
 enum class request_type { seqread, seqwrite, randread, randwrite, append, cpu, unlink };
 
+enum class unlink_type { ordinary, discard_blocks };
+
 namespace std {
 
 template <>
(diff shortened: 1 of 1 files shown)


avikivity:
Experimentation can be done with fio, it's just playing with configuration files.

apps/io_tester/io_tester.cc
```diff
@@ -742,6 +753,76 @@ class unlink_class_data : public class_data {
...
+
+    future<size_t> issue_blocks_discarding_unlink() {
+        return ensure_file_to_discard_open().then([this]() {
+            return discard_next_block();
+        }).then([this]() {
+            return close_file_if_all_blocks_discarded();
+        }).then([]() {
+            return make_ready_future<size_t>(0u);
+        });
+    }
```
pwrobelse, line 777: Sadly, the behavior of these lines of code differs from the behavior of `fallocate --punch-hole` tool.

When the workload is inspected via `blktrace`, the trim requests are issued only after the flush completes via `xfsaild`. 

I also tested calling unlink() when all blocks are discarded. No difference.

When I added a call to `fsync()` after each discard is issued, then I saw trim requests issued with certain frequency.

robertbindar (approved):
This code works as intended, I ran it in various scenarios to reproduce some of Patryk's experiments and it works alright. Code looks alright.
Regarding Patryk's concern on trim requests getting queued until they're picked up by `xfsaild`, fsyncing the file after each block discard solves the issue and there was no measurable impact on parallel workflow latencies, we can add an option in the future to force an fsync if it becomes necessary.

PR #2143, opened 2024-03-13
membarrier: cooperatively serialize calls to sys_membarrier

avikivity:
In [1], Linux started serializing sys_membarrier calls with a mutex, citing slowdowns. In fact we observed such slowdowns ([2]) on AWS d3en instances.

While the reactor calling sys_membarrier is going to sleep and wouldn't mind a slowdown, this can cause a large increase in wakeup latency. If many reactors are simulataneously trying to sleep, they will have to wait until they get to acquire the mutex before ever getting a chance to poll for new events.

Fix this by only allowing one reactor to attempt to generate a barrier at a time; if it's racing with another reactor, we'll just return false and go back to polling. This returns our wakeup latency to normal.

In fact, we already had this lock before - this is a partial revert of [3]. It's different in that it now applies to both madvise() based barriers and membarrier() based barriers, whereas before it only applied to mprotect() based barriers (a predecessor of madvise() barriers). Since madvise() barriers are only used on ancient kernels, the effort to check whether locking is advisable is not worthwhile.

[1] https://github.com/torvalds/linux/commit/944d5fe50f3f03daacfea16300e656a1691c4a23
[2] https://github.com/scylladb/scylladb/issues/17207
[3] https://github.com/scylladb/seastar/commit/77a58e4dc020233f66fccb8d9e8f7a8b7f9210c4

### DIFF
--- a/src/core/systemwide_memory_barrier.cc
+++ b/src/core/systemwide_memory_barrier.cc
@@ -41,6 +41,7 @@ module seastar;
 #include <seastar/core/systemwide_memory_barrier.hh>
 #include <seastar/core/cacheline.hh>
 #include <seastar/util/log.hh>
+#include <seastar/util/defer.hh>
 #endif
 
 namespace seastar {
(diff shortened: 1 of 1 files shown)


avikivity:
v2: don't use `[] noexcept` lambdas, they're C++23 only. Use `[] () noexcept`.

PR #2146, opened 2024-03-16
smp: allocate hugepages eagerly when kernel support is available

avikivity:
Instead of deferring merging pages into hugepages to the transparent
hugepage scanner, advise the kernel to do so immediately using the new
MADV_POPULATE_WRITE and MADV_COLLAPSE advices.

Refactor the prefaulter to attempt first to use MAP_POPULATE_WRITE
to fault in a whole hugepage's worth of memory. This should fault
the range as a hugepage but for good measure use MADV_COLLAPSE too
(which would be a no-op if the work was done in MADV_POPULATE_WRITE).

### DIFF
--- a/src/core/smp.cc
+++ b/src/core/smp.cc
@@ -25,6 +25,10 @@ module;
 #include <boost/range/algorithm/find_if.hpp>
 #include <memory>
 #include <vector>
+#include <regex>
+#include <sys/mman.h>
+#include <unistd.h>
+#include <fcntl.h>
 
 #ifdef SEASTAR_MODULE
 module seastar;
@@ -35,6 +39,8 @@ module seastar;
 #include <seastar/core/semaphore.hh>
 #include <seastar/core/print.hh>
 #include <seastar/core/on_internal_error.hh>
+#include <seastar/core/posix.hh>
+#include <seastar/core/align.hh>
 #include "prefault.hh"
 #endif
 
@@ -173,11 +179,32 @@ smp::setup_prefaulter(const seastar::resource::resources& res, seastar::memory::
 #endif
 }
 
+static
+std::optional<size_t>
+get_huge_page_size() {
+    auto meminfo_fd = file_desc::open("/proc/meminfo", O_RDONLY | O_CLOEXEC);
+    std::string meminfo;
+    char buf[4096];
+    while (auto size_opt = meminfo_fd.read(buf, sizeof(buf))) {
+        if (!*size_opt) {
+            break;
+        }
+        meminfo.append(buf, *size_opt);
+    }
+    static std::regex re(R"(Hugepagesize:\s*(\d+) kB)");
+    auto m = std::smatch{};
+    if (std::regex_search(meminfo, m, re)) {
+        return std::stoi(m[1]) * size_t(1024);
+    }
+    return std::nullopt;
+}
+
 internal::memory_prefaulter::memory_prefaulter(const resource::resources& res, memory::internal::numa_layout layout) {
     for (auto& range : layout.ranges) {
         _layout_by_node_id[range.numa_node_id].push_back(std::move(range));
     }
     auto page_size = getpagesize();
+    auto huge_page_size_opt = get_huge_page_size();
     for (auto& numa_node_id_and_ranges : _layout_by_node_id) {
         auto& numa_node_id = numa_node_id_and_ranges.first;
         auto& ranges = numa_node_id_and_ranges.second;
@@ -191,8 +218,8 @@ internal::memory_prefaulter::memory_prefaulter(const resource::resources& res, m
             }
             a.set(cpuset);
         }
-        _worker_threads.emplace_back(a, [this, &ranges, page_size] {
-            work(ranges, page_size);
+        _worker_threads.emplace_back(a, [this, &ranges, page_size, huge_page_size_opt] {
+            work(ranges, page_size, huge_page_size_opt);
         });
     }
 }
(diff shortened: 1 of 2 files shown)


avikivity:
@xemul please review

xemul:
Side note -- https://lwn.net/Articles/846501/
There's MADV_POPULATE flag that "prefaults" memory for us. Do you think it's worthwhile? Should be present since 5.14

src/core/smp.cc
```diff
@@ -218,6 +245,18 @@ internal::memory_prefaulter::work(std::vector<memory::internal::memory_range>& r
...
         auto p1 = reinterpret_cast<volatile std::atomic<char>*>(p);
         p1->fetch_or(0, std::memory_order_relaxed);
+#endif
+#ifdef MADV_COLLAPSE
+    if (huge_page_size_opt) {
+        auto addr = reinterpret_cast<uintptr_t>(p);
+        auto aligned = addr & ~(page_size - 1);
+        auto offset_in_hugepage = aligned & (*huge_page_size_opt - 1);
+        auto first_page_offset = forwards ? 0 : *huge_page_size_opt - page_size;
+        auto hugepage_addr = aligned - offset_in_hugepage;
```
xemul, line 255: Why not just `hugepage_addr = addr & ~(huge_page_size_opt - 1)` ?

avikivity:
MADV_POPULATE_WRITE is probably better than MADV_COLLAPSE. We could just issue both.

1. If MADV_POPULATE_WRITE is available, issue it, followed by MADV_COLLAPSE which is likely a no-op
2. Otherwise do the funky stuff with manually faulting in pages

src/core/smp.cc
xemul: @avikivity , 

> Why not just hugepage_addr = addr & ~(huge_page_size_opt - 1) ?

avikivity: That just happened to be the way I wrote it. I'll reconsider, but more important is to use MAP_POPULATE_WRITE instead.

avikivity:
v2: rewrite; use MADV_POPULATE_WRITE if available to have the kernel do the faulting. Add new helper populate_memory() to work on a hugepage's worth of memory

src/core/smp.cc
```diff
@@ -173,11 +179,32 @@ smp::setup_prefaulter(const seastar::resource::resources& res, seastar::memory::
...
+    auto meminfo_fd = file_desc::open("/proc/meminfo", O_RDONLY | O_CLOEXEC);
+    std::string meminfo;
+    char buf[4096];
+    while (auto size_opt = meminfo_fd.read(buf, sizeof(buf))) {
+        if (!*size_opt) {
+            break;
+        }
+        meminfo.append(buf, *size_opt);
+    }
+    static std::regex re(R"(Hugepagesize:\s*(\d+) kB)");
```
nyh, line 194: Nitpick: I don't think the outer parentheses are needed.

By the way, I hope nobody will ever change the kernel to print any other units except kB in /proc/meminfo :-) I believe it's hard-coded to kB right now, so it's fine.

src/core/smp.cc
```diff
@@ -173,11 +179,32 @@ smp::setup_prefaulter(const seastar::resource::resources& res, seastar::memory::
...
+    auto m = std::smatch{};
+    if (std::regex_search(meminfo, m, re)) {
+        return std::stoi(m[1]) * size_t(1024);
```
nyh, line 197: Is it m[1] and not m[0] because of that extra outer parentheses in the regular expression?

src/core/smp.cc
```diff
@@ -205,12 +232,33 @@ internal::memory_prefaulter::~memory_prefaulter() {
...
+        std::optional<size_t> huge_page_size_opt) {
     sched_param param = { .sched_priority = 0 };
     // SCHED_IDLE doesn't work via thread attributes
     pthread_setschedparam(pthread_self(), SCHED_IDLE, &param);
     size_t current_range = 0;
-    const size_t batch_size = 512; // happens to match huge page size on x86, through not critical
+    const size_t batch_size = huge_page_size_opt.value_or(512*4096);
+    auto populate_memory_madvise = [works = true] (char* start, char* end) mutable {
+#ifdef MADV_POPULATE_WRITE
+        if (works) {
```
nyh, line 244: since works is always set to true above, what's the point of this if()?
Or what does it help to set works in various places below, if you don't use it afterwards? (I expected to see something like "return works", but you didn't do it...).

src/core/smp.cc
```diff
@@ -227,25 +275,29 @@ internal::memory_prefaulter::work(std::vector<memory::internal::memory_range>& r
...
-        range.start = start;
-        range.end = end;
-        if (start >= end) {
+        auto batch_start = std::max(align_down(range.end - 1, batch_size), range.start);
+        populate_memory(batch_start, range.end);
+        range.end = batch_start;
+
+        if (range.start >= range.end) {
             ranges.erase(ranges.begin() + current_range);
             current_range = 0;
```
nyh, line 302: This is pre-existing code, but I still wonder - why do we have this current_range at all? It seems it's set to 0 initially, and then always remains 0 or set to 0 again. Where is it set to anything else?

nyh (changes requested):

src/core/smp.cc
```diff
@@ -173,11 +179,32 @@ smp::setup_prefaulter(const seastar::resource::resources& res, seastar::memory::
...
 }
 
+static
+std::optional<size_t>
+get_huge_page_size() {
+    auto meminfo_fd = file_desc::open("/proc/meminfo", O_RDONLY | O_CLOEXEC);
+    std::string meminfo;
+    char buf[4096];
+    while (auto size_opt = meminfo_fd.read(buf, sizeof(buf))) {
+        if (!*size_opt) {
```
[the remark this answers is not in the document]
avikivity: size_opt cannot be disengaged here, since we'd exit the while loop.

src/core/smp.cc
avikivity: The outer parentheses are needed. The middle parentheses are part of the `R"` literal syntax.

src/core/smp.cc
avikivity: No

src/core/smp.cc
avikivity: works is changed to `false` if it doesn't work.

src/core/smp.cc
avikivity: It's incremented below.

src/core/smp.cc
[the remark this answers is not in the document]
nyh: Oh, I missed the fact it's the while's condition. Ok.

src/core/smp.cc
nyh: Wow, I never even heard of this crazy syntax, and now it turns out it already existed in C++11. This language is a disaster :-(

src/core/smp.cc
nyh: So why is this not m[0]?

src/core/smp.cc
nyh: Oh, I see now - you call the same lambda multiple times and it changes for the next call.

nyh (approved):

src/core/smp.cc
avikivity: m[0] is the entire match, not a subexpression.

https://en.cppreference.com/w/cpp/regex/match_results/operator_at

PR #2149, opened 2024-03-19
memory: Ensure thread locals etc are minimally initialized even with non-seastar reactor options for alloc

### LINKED ISSUE #2148: Running without `SEASTAR_USE_DEFAULT_ALLOCATOR` but smp_opts.memory_allocator != memory_allocator::seastar breaks
Found in https://github.com/scylladb/scylladb/issues/17880.

When using the seastar allocator, there are numerous implicit reliances on `cpu_mem_ptr` having been set up. Previously, this just happened to be true, due to #2137. With that fixed, suddenly it is not, unless `smp::configure` is invoked with `memory_allocator == memory_allocator::seastar`. 

It not being so is rare, and the test we have for it is only run for dpdk builds (why?). And even so, does not provoke the problem, since it normally never hits a reactor poller or similar. Adding a small sleep to the test fixes this -> crash.

elcallio:
Fixes #2148

When using the seastar allocator, there are numerous implicit reliances on cpu_mem_ptr having been set up. Previously, this just happened to be true, due to #2137.

Fixed by adding a minimal memory setup call to smp::configure, even if using non-seastar allocator reactor options (not compile flags).

Fixed the test for the feature (allocator options) so that it 
* Actually runs. regardlerss of dpdk
* Actually catches the issue

### DIFF
--- a/src/core/memory.cc
+++ b/src/core/memory.cc
@@ -1641,6 +1641,13 @@ void *allocate_slowpath(size_t size) {
             return original_malloc_func(size);
         }
 
+        // original_malloc_func might be null for allocations before main
+        // in constructors before original_malloc_func ctor is called
+        // Note on #2137: Moved to here, because there is lots of code 
+        // that implicitly relies on the static init fiasco below to have occurred, and thus
+        // cpu_mem_ptr being available and inited. This is not great.
+        init_cpu_mem();
+
         // #2137 - static init fiasco for fallback functions. 
         // If dependent libraries do malloc _before_ the above declaration inits are run,
         // we end up here with nowhere to go. Add a second check and attempt the full init
@@ -1663,10 +1670,6 @@ void *allocate_slowpath(size_t size) {
                 return allocate_slowpath(size);
             }
         }
-
-        // original_malloc_func might be null for allocations before main
-        // in constructors before original_malloc_func ctor is called
-        init_cpu_mem();
     }
     // On the fast path we've already called maybe_sample, except in the case
     // of !is_reactor_thread (we don't sample such alloctions).
@@ -1823,6 +1826,10 @@ size_t get_large_allocation_warning_threshold() {
 
 void disable_large_allocation_warning() {
     get_cpu_mem().large_allocation_warning_threshold = std::numeric_limits<size_t>::max();
+}
+
+void configure_minimal() {
+    init_cpu_mem();
 }
 
 internal::numa_layout
@@ -2633,6 +2640,9 @@ configure(std::vector<resource::memory> m, bool mbind,
         std::optional<std::string> hugepages_path) {
     return {};
 }
+
+void configure_minimal() 
+{}
 
 statistics stats() {
     return statistics{0, 0, 0, 1 << 30, 1 << 30, 0, 0, 0, 0, 0, 0};
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4085,7 +4085,8 @@ void smp::qs_deleter::operator()(smp_message_queue** qs) const {
         for (unsigned j = 0; j < smp::count; j++) {
             qs[i][j].~smp_message_queue();
         }
-        ::operator delete[](qs[i]);
+        ::operator delete[](qs[i], std::align_val_t(alignof(smp_message_queue))
+        );
     }
     delete[](qs);
 }
@@ -4414,6 +4415,11 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     std::optional<memory::internal::numa_layout> layout;
     if (smp_opts.memory_allocator == memory_allocator::seastar) {
         layout = memory::configure(allocations[0].mem, mbind, use_transparent_hugepages, hugepages_path);
+    } else {
+        // #2148 - if running seastar allocator but options that contradict this, we still need to
+        // init memory at least minimally, otherwise a bunch of stuff breaks.
+        // Previously, we got away wth this by accident due to #2137.
+        memory::configure_minimal();
     }
 
     if (reactor_opts.abort_on_seastar_bad_alloc) {
@@ -4527,6 +4533,9 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
                 auto another_layout = memory::configure(allocation.mem, mbind, use_transparent_hugepages, hugepages_path);
                 auto guard = std::lock_guard(mtx);
                 *layout = memory::internal::merge(std::move(*layout), std::move(another_layout));
+            } else {
+                // See comment above (shard 0)
+                memory::configure_minimal();
             }
             if (heapprof_sampling_rate) {
                 memory::set_heap_profiling_sampling_rate(heapprof_sampling_rate);
@@ -4584,7 +4593,14 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     _qs_owner = decltype(smp::_qs_owner){new smp_message_queue* [smp::count], qs_deleter{}};
     _qs = _qs_owner.get();
     for(unsigned i = 0; i < smp::count; i++) {
-        smp::_qs_owner[i] = reinterpret_cast<smp_message_queue*>(operator new[] (sizeof(smp_message_queue) * smp::count));
+        smp::_qs_owner[i] = reinterpret_cast<smp_message_queue*>(operator new[] (sizeof(smp_message_queue) * smp::count
+        // smp_message_queue has members with hefty alignment requirements.
+        // if we are reactor thread, or not running with dpdk, doing this
+        // new default aligned seemingly works, as does reordering
+        // dlinit dependencies (ugh). But we should enforce calling out to
+        // aligned_alloc, instead of pure malloc, if possible.
+            , std::align_val_t(alignof(smp_message_queue))
+        ));
         for (unsigned j = 0; j < smp::count; ++j) {
             new (&smp::_qs_owner[i][j]) smp_message_queue(reactors[j], reactors[i]);
         }
--- a/tests/unit/app-template_test.cc
+++ b/tests/unit/app-template_test.cc
@@ -24,17 +24,13 @@
 #include <string>
 #include <boost/test/unit_test.hpp>
 #include <seastar/core/app-template.hh>
+#include <seastar/core/sleep.hh>
 
 using namespace seastar;
+using namespace std::chrono_literals;
 
-#ifdef Seastar_DPDK
-constexpr bool with_dpdk = true;
-#else
-constexpr bool with_dpdk = false;
-#endif
-
-BOOST_AUTO_TEST_CASE(app_standard_memory_allocator,
-                     *boost::unit_test::enable_if<with_dpdk>()) {
+// #2148 - always run this.
+BOOST_AUTO_TEST_CASE(app_standard_memory_allocator) {
     // by default, use conservative settings instead of maxing out the performance
     // for testing app_template and underlying reactor's handling of different
     // settings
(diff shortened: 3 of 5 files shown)


src/core/reactor.cc
```diff
@@ -4414,6 +4414,12 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     std::optional<memory::internal::numa_layout> layout;
     if (smp_opts.memory_allocator == memory_allocator::seastar) {
         layout = memory::configure(allocations[0].mem, mbind, use_transparent_hugepages, hugepages_path);
+    } else {
+        // #2148 - if running seastar allocator but options that contradict this, we still need to
+        // init memory at least minimally, otherwise a bunch of stuff breaks.
+        // Previously, we got away wth this by accident due to #2137.
+        // Calling memory::configure without any memory etc will fix up internal structures enough.
+        memory::configure({}, mbind, false, std::nullopt);
```
denesb, line 4422: We should passe `false` as `mbind` as well. Our docs claim that `--mbind` is ignored when non-seastar allocator is used, so we should ignore it.

elcallio: fair enough.

src/core/reactor.cc
```diff
@@ -4414,6 +4414,12 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     std::optional<memory::internal::numa_layout> layout;
     if (smp_opts.memory_allocator == memory_allocator::seastar) {
         layout = memory::configure(allocations[0].mem, mbind, use_transparent_hugepages, hugepages_path);
+    } else {
+        // #2148 - if running seastar allocator but options that contradict this, we still need to
+        // init memory at least minimally, otherwise a bunch of stuff breaks.
+        // Previously, we got away wth this by accident due to #2137.
+        // Calling memory::configure without any memory etc will fix up internal structures enough.
+        memory::configure({}, false, false, std::nullopt);
```
denesb, line 4422: I just checked and this will set `is_reactor_thread=true`, which means that the seastar allocator will start serving allocation requests. This effectively breaks `smp_options::memory_allocator = memory_allocator::standard`, which scylla tools use.

elcallio: An alternative option is adding "config_minimal" that only does init_cpu_mem().

elcallio: Changed to the above.

denesb (approved):

avikivity:
Where do we rely on cpu_mem? Shouldn't they be guarded by is_reactor_thread?

elcallio:
No. A bunch of calls, like `seastar::memory::get_memory_layout()` rely on cpu_mem being inited, and are called from scylla both before, and regardless of whether, reactor has been created. Or whether said reactor is created with, or without seastar allocation enabled. 
There are a bunch of assumptions and interdependencies in(scylla) tests as well, which test features that more or less implicitly depend on this without a reactor init. Not sure why those did not fail for me on the initial SIF fix (pretty sure I ran a full scylla test.py on it).

elcallio:
As to "should", one could indeed wish...

tests/unit/app-template_test.cc
```diff
@@ -53,7 +52,11 @@ BOOST_AUTO_TEST_CASE(app_standard_memory_allocator,
     int actual_status = app.run(
         std::size(args), std::data(args),
         [expected_status] {
-            return make_ready_future<int>(expected_status);
+            // #2148 - add a small sleep to ensure the reactor does
+            // some of its background stuff, pollers etc for example.
+            return seastar::sleep(2s).then([expected_status] {
+                return make_ready_future<int>(expected_status);
```
avikivity, line 58: New flaky test!

src/core/memory.cc
```diff
@@ -1641,6 +1641,13 @@ void *allocate_slowpath(size_t size) {
             return original_malloc_func(size);
         }
 
+        // original_malloc_func might be null for allocations before main
+        // in constructors before original_malloc_func ctor is called
+        // Note on #2137: Moved to here, because there is lots of code 
+        // that implicitly relies on the SIF below to have occurred, and thus
```
avikivity, line 1647: What is SIF?

tchaikov: guess it's static init fiasco .

tests/unit/app-template_test.cc
elcallio: Should not be, as we will crash as soon as pollers are invoked. Considered waiting for something, but that is just the same thing. Need to ensure we simply end up in task queue, and a reactor poll is done.

src/core/memory.cc
elcallio: static init fiasco.

src/core/reactor.cc
```diff
@@ -4584,7 +4596,16 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     _qs_owner = decltype(smp::_qs_owner){new smp_message_queue* [smp::count], qs_deleter{}};
     _qs = _qs_owner.get();
     for(unsigned i = 0; i < smp::count; i++) {
-        smp::_qs_owner[i] = reinterpret_cast<smp_message_queue*>(operator new[] (sizeof(smp_message_queue) * smp::count));
+        smp::_qs_owner[i] = reinterpret_cast<smp_message_queue*>(operator new[] (sizeof(smp_message_queue) * smp::count
+        // smp_message_queue has members with hefty alignment requirements.
+        // if we are reactor thread, or not running with dpdk, doing this
+        // new default aligned seemingly works, as does reordering
+        // dlinit dependencies (ugh). But we should enforce calling out to
+        // aligned_alloc, instead of pure malloc, if possible.
```
tchaikov, line 4604: i don't quite understand this part.. 

i am not questioning the correctness of this change. as `__STDCPP_DEFAULT_NEW_ALIGNMENT__` is 16. and `sizeof(smp_message_queue)` is greater than 16. so this change should enforce the alignment of the allocated storage.

but how is this related to if we are seastar reactor or using a dpdk reactor? do you mean, we could perform unaligned access when using dpdk?

tests/unit/app-template_test.cc
```diff
@@ -24,22 +24,21 @@
...
-                     *boost::unit_test::enable_if<with_dpdk>()) {
+// #2148 - always run this.
+BOOST_AUTO_TEST_CASE(app_standard_memory_allocator) {
     // by default, use conservative settings instead of maxing out the performance
     // for testing app_template and underlying reactor's handling of different
     // settings
     app_template::seastar_options opts;
-    opts.smp_opts.thread_affinity.set_value(false);
+    // Here was a `opts.smp_opts.thread_affinity.set_value(false);`.
+    // This breaks the subsequent tests below (which run non-smp-1 config).
```
tchaikov, line 39: i am not following you. yes, we do have non-smp-1 tests in this source file. but why does your change break them?

src/core/reactor.cc
elcallio: When building with dpdk, we get a slightly different init sequence, which, when fixing the original SIF such that non-reactor allocation goes to libc malloc from the get-go, causes the above new/alloc to end up in different memory blocks -> without SIF fix we get nice, page-level memory. With it, we end up in normal malloc and get 16/8-aligned block. I can only assume because in the former case, we've done a bunch of library allocs actually via seastar allocator, which in turn means we've stolen a bunch of pages and thus impact where malloc, once called, will place things. 

Now, I assume that normally the allocation here should compensate for this unless calling a new(aligned) alloc, and over-alloc + adjust pointers + make the end result object messy internally (finding the original memory etc), but either due to the raw memory alloc + alloc in place, or something else I am missing, this does not happen. 

Honestly, I would love to have a more exact and detailed explanation, but my tolerance for single-instructioning myself through libc etc only goes so far. Should also be observed that the failure does not happen if running in gdb either, possibly because some slight .so loading difference there as well. 

In any case, what I could observe is alignment of memory we actually get in the two variants, and when if works we get properly aligned memory, even though the new goes to a simple malloc. Luck?

tests/unit/app-template_test.cc
tchaikov: i understand that pollers are also queued like other tasks. but why does it matter? what exactly are we waiting? what happens if we do not wait long enough? could you please elaborate it a little bit?

tests/unit/app-template_test.cc
elcallio: It does not. But running the tests with and without it does. Something gets confused with some thread-local somewhere, such that the latter tests cannot init, because smp::invoke_all fails to complete.

src/core/memory.cc
tchaikov: probably, would be better to repeat "static init fiasco" here, before "SIF" becomes a well-known abbreviation, like, for instance, OOM.

tests/unit/app-template_test.cc
elcallio: In fact, the service_group semaphores gets confused.

tests/unit/app-template_test.cc
elcallio: Pollers are run when we have nothing to run/after a task iteration. Even 2s here is probably on the high side. Just as long as we're pushed into the task queue, we should provoke the issue.

src/core/memory.cc
elcallio: Sure.

tests/unit/app-template_test.cc
elcallio: I changed my mind. Better to fix the issue.

src/core/reactor.cc
avikivity: The seastar allocator has a power-of-two thing so it gets naturally aligned, and I guess libc does not, so it makes sense.

avikivity: And, the original code was just broken - it bypassed the new-expression to call operator new directly (maybe to avoid allocating the array size), but "forgot" to align properly.

elcallio: Note however that we could get past this (by luck?) even when running with such options that the new call was always served by glibc malloc. Otoh, even then, we'd mostly have smp_count > 1 (i.e. 8 or so), and thus at least fall into larger alloc domain.

tests/unit/app-template_test.cc
elcallio: I've updated the comment to be more explicit about what we try to do.

src/core/reactor.cc
```diff
@@ -4584,7 +4596,16 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
...
+#ifdef __cpp_aligned_new
+            , std::align_val_t(alignof(smp_message_queue))
+#endif
+        ));
```
avikivity, line 4608: __cpp_aligned_new is from 2018 (gcc and clang), we can avoid the #ifdef

elcallio: Just being consistent with the definition of the aligned alloc functions in memory.cc.

src/core/reactor.cc
tchaikov: thank you very much Calle, all clear!

tests/unit/app-template_test.cc
tchaikov: thanks!

src/core/reactor.cc
tchaikov: probably we can drop them all together in another change?

tchaikov (approved):
lgtm. thanks!

tchaikov:
@avikivity @scylladb/seastar-maint hello maintainers, could you help review this change?

denesb (approved):

src/core/memory.cc
nyh: Arguably, even the pharse "static init fiasco" doesn't mean anything to the uninitiated.

src/core/reactor.cc
tchaikov: FWIW, Avi is removing them in #2150 .

elcallio: Ok, removed.

PR #2150, opened 2024-03-20
memory: drop support for compilers that don't support aligned new

avikivity:
Aligned new (__cpp_aligned_new) was introduced in C++17 and compilers support it since 2018. Remove the conditional compilation and refuse to build on compilers that are too old to support it (we require C++20+ anyway).


tchaikov (approved):
lgtm

avikivity:
@xemul please review

PR #2155, opened 2024-03-25
TLS: Add support for TLS1.3 session tickets

### LINKED ISSUE #2154: Add TLS1.3 session ticket support
TLS handshakes have a fairly significant overhead, esp. certificate checks etc. 
Many client libraries used against popular seastar applications support session resume of various kinds. 

While TLS1.2 session handling would be somewhat tricky (for sharding reasons etc), TLS1.3 session tickets place the full responsibility on the client side, bypassing many obstacles for a seastar based server (and even a client really). 
And encouraging people to use TLS1.3 is good in itself. 

Since GNUTLS has fairly transparent and easy API for session tickets, this should be easy to add.

elcallio:
Fixes #2154
    
TLS1.3 session tickets are a client-storage solution for session resume, bypassing some of the more expensive parts of a normal handshake.

Implemented by adding an optional session data parameter to tls_options, options to certificates for server keys, and query
functions for state/session data on the client side.

Note that the feature requires both client and server to handle tls1.3.

Server session keys are kept in the certificates object, and can be rotated by simply re-enabling the feature on the object.

Reloadable certificates will rotate keys on cert reload.

Simple test included.

### DIFF
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -242,6 +242,14 @@ namespace tls {
     };
 
     /**
+     * Session resumption support. 
+     * We only support TLS1.3 session tickets.
+    */
+    enum class session_resume_mode {
+        NONE, TLS13_SESSION_TICKET
+    };
+
+    /**
      * Extending certificates and keys for server usage.
      * More probably goes in here...
      */
@@ -258,6 +266,14 @@ namespace tls {
         server_credentials& operator=(const server_credentials&) = delete;
 
         void set_client_auth(client_auth);
+
+        /**
+         * Sets session resume mode.
+         * If session resumption is set to TLS13 session tickets, 
+         * calling this also functions as key rotation, i.e. creates
+         * a new window of TLS session keys.
+        */
+        void set_session_resume_mode(session_resume_mode);
     };
 
     class reloadable_credentials_base;
@@ -291,6 +307,7 @@ namespace tls {
         future<> set_system_trust();
         void set_client_auth(client_auth);
         void set_priority_string(const sstring&);
+        void set_session_resume_mode(session_resume_mode);
 
         void apply_to(certificate_credentials&) const;
 
@@ -306,8 +323,11 @@ namespace tls {
 
         std::multimap<sstring, boost::any> _blobs;
         client_auth _client_auth = client_auth::NONE;
+        session_resume_mode _session_resume_mode = session_resume_mode::NONE;
         sstring _priority;
     };
+
+    using session_data = std::vector<uint8_t>;
 
     /// TLS configuration options
     struct tls_options {
@@ -315,6 +335,10 @@ namespace tls {
         bool wait_for_eof_on_shutdown = true;
         /// \brief server name to be used for the SNI TLS extension
         sstring server_name = {};
+
+        /// \brief Optional session resume data. Must be retrieved via 
+        /// get_session_resume_data below.
+        session_data session_resume_data;
     };
 
     /**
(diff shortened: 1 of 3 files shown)


xemul (approved):

elcallio:
Update: Added small extra safety check/fix in `state_checked_access`, preserving previous recursive behaviour.

elcallio:
ping?

xemul:
@avikivity , @bhalevy , please review too

src/net/tls.cc
```diff
@@ -332,6 +332,19 @@ future<tls::x509_cert> tls::x509_cert::from_file(
...
+struct gnutls_datum : public gnutls_datum_t {
+    gnutls_datum() {
+        data = nullptr;
+        size = 0;
+    }
+    ~gnutls_datum() {
+        if (data != nullptr) {
+            ::gnutls_free(data);
+        }
+    }
```
avikivity, line 345: Somewhat subject to slicing.

avikivity:
Looks good to me, I asked some questions to improve my knowledge.

include/seastar/net/tls.hh
```diff
@@ -452,6 +476,28 @@ namespace tls {
...
+     * 
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
+     * If no session resumption data is available, returns empty buffer.
+     * 
+     * Note: TLS13 session tickets most of the time require data to have been transferred
+     * between client/server. To ensure getting the session data, it is advisable to 
+     * delay this call to sometime before shutting down/closing the socket.
+    */
+    future<session_data> get_session_resume_data(connected_socket&);
```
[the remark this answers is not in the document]
elcallio: Yes. See above (and test)

src/net/tls.cc
elcallio: ? What do you mean?

elcallio: It has no destructor whatsoever. This is just a simple struct to ensure we can call to gnutls functions that allocate gnutls_datum_t:s and still have some nice, safe RAII handling of freeing the data, while maintaining nice error checks etc...
The "virtual" part of datum destruction is in `gnutls_free`, which will handle whatever might be in it. Our side is responsible for calling it, nothing more.

tchaikov: > This is just a simple struct to ensure we can call to gnutls functions that allocate gnutls_datum_t:s and still have some nice, safe RAII handling of freeing the data, while maintaining nice error checks etc...

yeah, i see. just wanted to understand Avi's concern. and what data slicing would imply.

include/seastar/net/tls.hh
```diff
@@ -452,6 +476,28 @@ namespace tls {
...
     future<std::vector<subject_alt_name>> get_alt_name_information(connected_socket& socket, std::unordered_set<subject_alt_name_type> types = {});
 
+    /**
+     * Checks if the socket was connected using session resume.
+     * Will force handshake if not already done. 
+     * 
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
+    */
+    future<bool> check_session_is_resumed(connected_socket& socket);
```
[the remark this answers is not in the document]
avikivity: Would merging it with get_session_resume_data be helpful? return optional\<session_data> to save a step.

src/net/tls.cc
avikivity: If gnutls_datum as move-assigned to gnutls_datum_t, it will destroy prematurely.

It's not a big concern since the type doesn't escape this file.

elcallio: Not meant to be moved, no. If you really need it, I can add move-constructor etc, but it seems a bit overzealous...

PR #2156, opened 2024-03-26
http: Move parsed params and urls

xemul:
The add_param() method decodes parameter and then places it into the params map. The "places" can move local variable into map, not copy.

Same for url_decode() itself -- at the end it places the parsed result into its output parameter and can move it there instead of copying.

### DIFF
--- a/src/http/request.cc
+++ b/src/http/request.cc
@@ -76,7 +76,7 @@ void request::add_param(const std::string_view& param) {
         sstring value;
         if (http::internal::url_decode(param.substr(0,split), key)
                 && http::internal::url_decode(param.substr(split + 1), value)) {
-            query_parameters[key] = value;
+            query_parameters[key] = std::move(value);
         }
     }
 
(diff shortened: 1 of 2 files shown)


xemul:
@nyh , please review

nyh:
As usual, CI silently ignored this PR because of political reason, and as usual I don't remember how to force the CI to run manually, so I didn't :-(
Can't the political CI at least do us the courtesy and print a message saying it didn't run and how to run it manually? CC @benipeled 

So before merging I just verified the build on one case (my laptop) - but nothing more.

benipeled:
> As usual, CI silently ignored this PR because of political reason, and as usual I don't remember how to force the CI to run manually, so I didn't :-( Can't the political CI at least do us the courtesy and print a message saying it didn't run and how to run it manually? CC @benipeled

I doubt,
It's a tiny workflow, we should move it to gihtub-actions - last time I tried we had an issue with the slowness-machines - I'll try checking it again

PR #2157, opened 2024-03-27
net: Modify the listen_backlog parameter, because this parameter is s…

tangcheng999:
Modify the listen_backlog parameter, because this parameter is small and causes a connect timeout error on the client.


xemul:
The correct way of fixing it is to patch scylla and make this parameter configurable, not hard-code another heuristic value

Issue #2158, opened 2024-03-27
`rpc_test.cc` does not compile with libstdc++ shipped along with GCC-14


tchaikov:
```
FAILED: tests/unit/CMakeFiles/test_unit_rpc.dir/rpc_test.cc.o
/usr/bin/g++-14 -DBOOST_ALL_DYN_LINK -DFMT_SHARED -DSEASTAR_API_LEVEL=7 -DSEASTAR_BUILD_SHARED_LIBS -DSEASTAR_DEBUG -DSEASTAR_DEBUG_PROMISE -DSEASTAR_DEBUG_SHARED_PTR -DSEASTAR_DEFAULT_ALLOCATOR -DSEASTAR_DEFERRED_
ACTION_REQUIRE_NOEXCEPT -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_NUMA -DSEASTAR_HAVE_SYSTEMTAP_SDT -DSEASTAR_HAVE_URING -DSEASTAR_LOGGER_COMPILE_TIME_FMT -DSEAS
TAR_LOGGER_TYPE_STDOUT -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_SHUFFLE_TASK_QUEUE -DSEASTAR_SSTRING -DSEASTAR_TESTING_MAIN -DSEASTAR_TESTING_WITH_NETWORKING=1 -DSEASTAR_TYPE_ERASE_MORE -I/var/ssd/seastar/tes
ts/unit -I/var/ssd/seastar/src -I/var/ssd/seastar/include -I/var/ssd/seastar/build/debug/gen/include -I/var/ssd/seastar/build/debug/gen/src -g -std=gnu++23 -U_FORTIFY_SOURCE -Wno-maybe-uninitialized -Wno-error=unus
ed-result -fstack-clash-protection -fsanitize=address -fsanitize=undefined -fno-sanitize=vptr -UNDEBUG -Wall -Werror -Wimplicit-fallthrough -Wdeprecated -Wno-error=deprecated -Wno-error=stringop-overflow -Wno-error
[...]

PR #2159, opened 2024-03-27
rpc: cast rpc::tuple to std::tuple when passing it to std::apply()

tchaikov:
before this change, we implement rpc::tuple by inheriting it from
std::tuple, and following the tuple-like protocol, by implementing
`tuple_size<seastar::rpc::tuple<T...>>` and
`tuple_element<I,seastar::rpc::tuple<T...>>`. but in C++23, the
tuple-like constraints are tightened, quote from libstdc++ shipped
along with GCC-14

```c++
  template<typename _Tp>
    inline constexpr bool __is_tuple_v = false;

  template<typename... _Ts>
    inline constexpr bool __is_tuple_v<tuple<_Ts...>> = true;

  // TODO: Reuse __is_tuple_like from <type_traits>?
  template<typename _Tp>
    inline constexpr bool __is_tuple_like_v = false;

  template<typename... _Elements>
    inline constexpr bool __is_tuple_like_v<tuple<_Elements...>> = true;

  template<typename _T1, typename _T2>
    inline constexpr bool __is_tuple_like_v<pair<_T1, _T2>> = true;

  template<typename _Tp, size_t _Nm>
    inline constexpr bool __is_tuple_like_v<array<_Tp, _Nm>> = true;
// ...
```

which is a loyal translation of the C++23 standard.

apparently, `rpc::tuple` cannot fulfill this constraint, hence
`rpc_test.cc` fails to compile with GCC-14 and C++23. fortunately,
`rpc::tuple` is derived from `std::tuple`.

so, in this change, we cast `rpc::tuple` to `std::tuple` when passing
it to `std::apply` to satisfy the more picky critera in C++23.

Fixes https://github.com/scylladb/seastar/issues/2158

### DIFF
--- a/include/seastar/rpc/rpc_impl.hh
+++ b/include/seastar/rpc/rpc_impl.hh
@@ -241,7 +241,11 @@ struct marshall_one {
             auto do_do_marshall = [&serializer, &out] (const auto&... args) {
                 do_marshall(serializer, out, args...);
             };
-            std::apply(do_do_marshall, arg);
+            // since C++23, std::apply() only accepts tuple-like types, while
+            // rpc::tuple is not a tuple-like type from the tuple-like C++
+            // concept's perspective. so we have to cast it to std::tuple to
+            // appease std::apply()
+            std::apply(do_do_marshall, static_cast<const std::tuple<T...>&>(arg));
         }
     };
 };


tchaikov:
to be more future-proof and standard compliant, either we need to avoid using `std::apply()` or to deprecate the non-alias version along with its deduction guide.

tchaikov:

nyh (approved):

tchaikov:
thank you for your approval, Nadav. my guess is that the deduction guide cannot be applied to a template alias. but it's handy to use -- `make_tuple` is shorter than `tuple`.

gleb-cloudius:
The context is here e51a1a8ed955030774cf3ae406d725613e4a099b

tchaikov:
@gleb-cloudius thank you Gleb. seems i need to implement a `std::apply()` which works with `rpc::tuple`.

gleb-cloudius:
But if rpc::tuple is just an alias and not a separate type how can it be implemented?

tchaikov:
@gleb-cloudius i intend to respect the original idea

> I chose to introduce rpc::tuple instead of using std::tuple, since there may be existing users of std::tuple in rpc, with its own serialization.

so, will keep it as a derived class of `std::tuple`. and instead implement our own `apply()` which does not enforce the type constraints.

gleb-cloudius:
> @gleb-cloudius i intend to respect the original idea
> 
> > I chose to introduce rpc::tuple instead of using std::tuple, since there may be existing users of std::tuple in rpc, with its own serialization.
> 
> so, will keep it as a derived class of `std::tuple`. and instead implement our own `apply()` which does not enforce the type constraints.

Ah. I misunderstood. Sounds fine.

tchaikov:
tested with clang-19 + c++23.
```console
$ grep CMAKE_CXX_STANDARD build/debug/CMakeCache.txt
CMAKE_CXX_STANDARD:STRING=23
$ cmake --build /var/ssd/seastar/build/debug --target test_unit_rpc
 [1/7] Building CXX object CMakeFiles/seastar.dir/src/rpc/lz4_fragmented_compressor.cc.o
 [2/7] Building CXX object CMakeFiles/seastar.dir/src/rpc/lz4_compressor.cc.o
 [3/7] Building CXX object CMakeFiles/seastar.dir/src/rpc/rpc.cc.o
 [4/7] Linking CXX shared library libseastar.so
 [5/7] Linking CXX shared library libseastar_testing.so
 [6/7] Building CXX object tests/unit/CMakeFiles/test_unit_rpc.dir/rpc_test.cc.o
 [7/7] Linking CXX executable tests/unit/rpc_test
```

gleb-cloudius (approved):

tchaikov:
v2:

- s/apply/rpc::apply/  to address the build failure like `/home/circleci/project/include/seastar/rpc/rpc_impl.hh:261:13: error: call to 'apply' is ambiguous` when compiling with older libstdc++.

avikivity:
Alternatively: model tuple-like (https://en.cppreference.com/w/cpp/utility/tuple/tuple-like). These need to overload std::tuple_size, std::tuple_element, and std::get.

tchaikov:
> Alternatively: model tuple-like (https://en.cppreference.com/w/cpp/utility/tuple/tuple-like). These need to overload std::tuple_size, std::tuple_element, and std::get.

@avikivity it's a no-go. we cannot model tuple-like. please take a closer look at the cover letter and the explanation in the link above actually renders it impossible. std::tuple_size, std::tuple_element, and std::get helps with the structured-binding, but they does not get a homebrew `tuple` a ticket into the tuple-alike club.

tchaikov:
@nyh hi Nadav, per Gleb's comment, i am using another approach to address the build failure on C++23, could you take another look?

tchaikov:
@nyh could please you take a look?

tchaikov:

tchaikov:

include/seastar/rpc/rpc_impl.hh
```diff
@@ -209,6 +209,23 @@ struct serialize_helper<true> {
...
+        std::forward<Fn>(fn),
+        std::get<Index>(std::forward<Tuple>(t))...);
+}
+
+template <typename Fn, typename Tuple>
+constexpr decltype(auto) apply(Fn&& fn, Tuple&& t) {
+     return apply_impl(
+         std::forward<Fn>(fn),
+         std::forward<Tuple>(t),
+         std::make_index_sequence<std::tuple_size_v<std::remove_reference_t<Tuple>>>{});
```
avikivity, line 226: Couldn't we static_cast to std::tuple and forward to std::apply?

tchaikov: ahh, we can! and probably we should . changed accordingly.

tchaikov:
v3:

- instead of duplicating `std::apply`, just cast `rpc::tuple` to `std::tuple`.

tchaikov:
@avikivity thank you for your suggestion. incorporated and it works great. could you take another look?

tchaikov:
@avikivity could you take another look?

PR #2160, opened 2024-03-28
scripts/stall-analyser: resolve string escape sequence warning

bhalevy:
Use a raw string for the following regexe:
```
/home/bhalevy/dev/scylla/seastar/scripts/stall-analyser.py:329: SyntaxWarning: invalid escape sequence '\s'
  comment = re.compile('^\s*#')
```


tchaikov (approved):
lgtm. BTW, it's a subset of #1830

PR #2162, opened 2024-03-28
http: add missing decoding of path part of URL

### LINKED ISSUE #725: http server is missing support for percent decoding

nyh:
Seastar's HTTP server implementation correctly does percent decoding (e.g., "%21" to "!") on the individual pieces of the query part of the URL (i.e., after the "?"). However, we forgot to do the same decoding also on the path part of the URL (before the "?"). This patch fixes this oversight.

The patch fixes two places: The more important place is the code which splits the URL to the path and the different query parameters. This is the fix most users will need. The other place is the code which allows a rule matching a specific path to also match an encoded version. This patch also adds simple test cases for both.

I also verified that this fix fixes a bug for one user of this HTTP server (the ScyllaDB project), which did not get correctly-decoded paths before this patch, and works as expected afterwards.

Fixes #725

### DIFF
--- a/src/http/matcher.cc
+++ b/src/http/matcher.cc
@@ -29,6 +29,7 @@ module;
 module seastar;
 #else
 #include <seastar/http/matcher.hh>
+#include <seastar/http/url.hh>
 #endif
 
 namespace seastar {
@@ -65,7 +66,12 @@ size_t param_matcher::match(const sstring& url, size_t ind, parameters& param) {
         }
         return sstring::npos;
     }
-    param.set(_name, url.substr(ind, last - ind));
+    sstring path;
+    if (http::internal::url_decode(url.substr(ind, last - ind), path)) {
+        param.set(_name, path);
+    } else {
+        param.set(_name, url.substr(ind, last - ind));
+    }
     return last;
 }
 
--- a/src/http/request.cc
+++ b/src/http/request.cc
@@ -82,10 +82,19 @@ void request::add_param(const std::string_view& param) {
 
 }
 
+static sstring decode_path(std::string_view path) {
+    sstring out;
+    if (http::internal::url_decode(path, out)) {
+        return out;
+    } else {
+        return sstring(path);
+    }
+}
+
 sstring request::parse_query_param() {
     size_t pos = _url.find('?');
     if (pos == sstring::npos) {
-        return _url;
+        return decode_path(_url);
     }
     size_t curr = pos + 1;
     size_t end_param;
@@ -95,7 +104,7 @@ sstring request::parse_query_param() {
         curr = end_param + 1;
     }
     add_param(url.substr(curr));
-    return _url.substr(0, pos);
+    return decode_path(_url.substr(0, pos));
 }
 
 void request::write_body(const sstring& content_type, sstring content) {
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -81,6 +81,12 @@ SEASTAR_TEST_CASE(test_match_rule)
     res = mr.get("/hell/val1", param);
     httpd::handler_base* nl = nullptr;
     BOOST_REQUIRE_EQUAL(res, nl);
+
+    // Test that URL decoding happens correctly on the path part of
+    // the URL. Reproduces issue #725.
+    res = mr.get("/hello/hello%21hi", param);
+    BOOST_REQUIRE_EQUAL(param["param"], "hello!hi");
+
     return make_ready_future<>();
 }
 
@@ -202,6 +208,12 @@ SEASTAR_TEST_CASE(test_decode_url) {
     req.parse_query_param();
     BOOST_REQUIRE_EQUAL(req.get_query_param("a"), "#$#");
     BOOST_REQUIRE_EQUAL(req.get_query_param("b"), "\"&\"");
+    // Test that URL decoding happens correctly on the path part of
+    // the URL too. Reproduces issue #725.
+    req._url = "/a%21b?q=%23%24%23";
+    url = req.parse_query_param();
+    BOOST_REQUIRE_EQUAL(url, "/a!b");
+    BOOST_REQUIRE_EQUAL(req.get_query_param("q"), "#$#");
     return make_ready_future<>();
 }
 


fruch:
Should it also be done on the host name ? Or I'm missing something here....

src/http/matcher.cc
```diff
@@ -65,7 +66,12 @@ size_t param_matcher::match(const sstring& url, size_t ind, parameters& param) {
         }
         return sstring::npos;
     }
-    param.set(_name, url.substr(ind, last - ind));
+    sstring path;
+    if (http::internal::url_decode(url.substr(ind, last - ind), path)) {
```
xemul, line 70: @nyh , `url_decode` doesn't do path decoding correctly, see #2125 (https://github.com/scylladb/seastar/pull/2125/commits/8b7f8a3badbeeec0096d24ab637c87311c9d4e16)

nyh: Yes, I noticed the incorrect "+" decoding added by @bhalevy in 799a9aacb53db4f13c35caecf5071cf078b37159. I somehow accepted this when I reviewed it (https://github.com/scylladb/seastar/pull/1531) - saying that I "am not happy" didn't prevent me from merging it :-(
I thought that this was a pre-existing problem that my patch didn't need to solve, but perhaps you're right it should - especially when it comes to the path when that HTML-only "+" decoding should **never** be done.

nyh:
This is another, different, fix for the same issue in https://github.com/scylladb/seastar/pull/2125. Only one of the two can go in.

src/http/matcher.cc
```diff
@@ -65,7 +66,12 @@ size_t param_matcher::match(const sstring& url, size_t ind, parameters& param) {
...
+        param.set(_name, path);
```
avikivity, line 71: std::move()?

src/http/request.cc
```diff
@@ -82,10 +82,19 @@ void request::add_param(const std::string_view& param) {
...
 }
 
+static sstring decode_path(std::string_view path) {
+    sstring out;
+    if (http::internal::url_decode(path, out)) {
+        return out;
+    } else {
+        return sstring(path);
+    }
+}
```
avikivity, line 92: Adds an allocation.

src/http/matcher.cc
pgellert, line 71: This is a breaking change because servers using this code (eg. redpanda here: https://github.com/redpanda-data/redpanda/blob/29be93c2b379e5b72afeb07eeed9db93cb2978df/src/v/redpanda/admin/security.cc#L269) that are currently assuming that param[] returns a raw undecoded path parameter are doing their own path decoding of the result of param[]. Without the server being aware of this breaking change, this could lead to double-decoding of path parameters, which is a problem because path decoding is not idempotent.

For more context, see these two comments:
* https://github.com/scylladb/seastar/pull/2125#discussion_r1547811750
* https://github.com/scylladb/seastar/pull/2125#discussion_r1542606401

nyh: The situation here is very depressing. CC @avikivity @amnonh:

Basically, we have a library, Seastar HTTPD, with a serious bug - unlike what the HTTP RFC says a server needs to do, it does URL decoding in one case but not in other cases. This bug was known for more than 4 years (https://github.com/scylladb/seastar/issues/725) and caused application bugs, and yet the two major (only?) users of the library decided to work around the bug instead of just fixing it:
1. Scylla avoided using characters like "#" that can expose this bug (https://github.com/scylladb/scylladb/issues/5883)
2. Redpanda worked around the bug by doing URL decoding on the result that the library forgot to do it.

Now that someone (me) got fed up with this bug and wanted to fix it, he is told: No, we need to keep the bug because users now rely on the buggy behavior. We can't fix the bug and will need to keep it forever.

I don't want to accept that. If the consensus is that we can't fix param[], then we need to deprecate it and introduce a new, bug-free, API. We shouldn't keep both as good alternatives - the buggy one isn't good - it's buggy.

amnonh: Specifically for the query params, I vote for deprecating and using an alternative correct implementation. When doing so, changing its API, the current implementation returns a single result per param (which is easy to use), but query parameters can support multiple values, if we are fixing things, we can fix this as well.

nyh: The splitting of the "query" part of the different keys and values is defined by the HTML standard (not HTTP or URL standards), and needs to happen before the URL decoding (yes, it's a big mess...). I'm not aware of similar encoding for the "path" part. The path part is normally just a single string with slashes in it, and doesn't have key-value pairs. Or maybe you meant to separate it into a vector, according to the un-encoded slashes? That could be done, but I don't know if anyone will need it, and it will complicate the lives of current users who get the entire path part unsplit (unless I'm missing something).

nyh:
> Should it also be done on the host name ? Or I'm missing something here....

In the HTTP standard, the host name comes separately in the "Host:" header.  Headers do not get encoded with "URL encoding" and are simply ASCII. Unicode characters in the domain name can be encoding using so-called punicode (https://datatracker.ietf.org/doc/html/rfc3492). But I think this is beyond the scope of this discussion.

nyh:
Because of the ongoing discussion here and in the different fix for the same issue in https://github.com/scylladb/seastar/pull/2125, I have downgraded my PR to draft mode.

src/http/matcher.cc
pgellert, line 70: Also note that with the path decoding already done in `parse_query_param`, the argument `url` here is already decoded, so there is no need to decode the path component a second time. (Let me know if I am missing something.)

nyh:
Closing this PR in favor of https://github.com/scylladb/seastar/pull/2125.

PR #2165, opened 2024-03-29
stall-analyser: use raw string to avoid invalid escaping sequence

tchaikov:
"\s" is not a valid escaping sequence, while "\\s" is. in this use case, what we want is a literal of "\s". so let's use raw string.


tchaikov:
oh my. we have 3 PRs for this tiny issue. i even forgot it i had one, and you have one as well.

PR #2164, opened 2024-03-29
stall-analyser: cleanups to address some formatting issues

tchaikov:
to improve the readability. and to be more PEP8 compliant.

PR #2166, opened 2024-03-29
stall-analyser: use 0 for the default value of --width

tchaikov:
instead using None for its value and checking for None later on, let's just use 0, simpler this way.


tchaikov:
this change conflicts with #2164. i will rebase another one if either of them lands first.

bhalevy (approved):

PR #2172, opened 2024-04-02
TLS: Move background BYE handshake to engine::run_in_background

### LINKED ISSUE #2171: TLS close can cause quickly existing programs to leave dangling objects (ASAN errors)
See https://github.com/scylladb/scylla-enterprise/issues/3913

If a program, say a test, closes a TLS connection to somewhere with slow-ish network ping time, the BYE handshake can take longer than shutting down, in which case reactor will not wait for the pending handshake -> potential error due to "leaked" memory. 

Potential solution (suggested by @bhalevy): Move the background handshake to (relatively) new `engine::run_in_background`, which will simply ensure that reactor shutdown waits for the handshake to finish or time out (in either case release the session object). 

Potential downside is that we might add as much as 10s to shutdown on network inactivity. But should not be an issue in normal cases.

elcallio:
Fixes #2171

If a program, say a test, closes a TLS connection to somewhere with slow-ish network ping time, the BYE handshake can take longer than shutting down, in which case reactor will not wait for the pending handshake -> potential error due to "leaked" memory.

Fix by moving the background BYE handshake to run_in_background. This forces reactor to wait for this until done or timed out, at which point at least session object will be released.

Potential downside is that a bad network can cause shutdown delays, but real world usage should not be affected.

### DIFF
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -1523,7 +1523,7 @@ public:
         if (!std::exchange(_shutdown, true)) {
             auto me = shared_from_this();
             // running in background. try to bye-handshake us nicely, but after 10s we forcefully close.
-            (void)with_timeout(timer<>::clock::now() + std::chrono::seconds(10), shutdown()).finally([this] {
+            engine().run_in_background(with_timeout(timer<>::clock::now() + std::chrono::seconds(10), shutdown()).finally([this] {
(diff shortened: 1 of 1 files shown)


bhalevy (approved):
LGTM

tchaikov:

PR #2173, opened 2024-04-02
docs: show a coroutine-based implementation of the echo server in the tutorial

Dobiasd:
Hi,

this pull request is meant mainly as a proposal. For me, it was eye-opening to convert the continuation-based implementation of the TCP-echo server to coroutines. So I thought seeing this might help others too. But I don't know if it makes sense to have it in the tutorial. So, if you don't see fit, I'd fully understand if you simply close this PR without merging. :relaxed:

### DIFF
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -1778,6 +1778,40 @@ It is often a mistake to silently ignore an exception, so if the future we're ig
 
 The ```handle_connection()``` function itself is straightforward --- it repeatedly calls ```read()``` read on the input stream, to receive a ```temporary_buffer``` with some data, and then moves this temporary buffer into a ```write()``` call on the output stream. The buffer will eventually be freed, automatically, when the ```write()``` is done with it. When ```read()``` eventually returns an empty buffer signifying the end of input, we stop ```repeat```'s iteration by returning a ```stop_iteration::yes```.
 
+Re-written using C++20's coroutines, the above becomes this:
+
+```cpp
+seastar::future<> handle_connection(seastar::connected_socket s) {
+    try {
+        auto out = s.output();
+        auto in = s.input();
+        while (true) {
+            auto buf = co_await in.read();
+            if (buf) {
+                co_await out.write(std::move(buf));
+                co_await out.flush();
+            } else {
+                break;
+            }
+        }
+        co_await out.close();
+    }
+    catch (const std::exception &ex) {
+        fmt::print(stderr, "Could not handle connection: {}\n", ex);
+    }
+}
+
+seastar::future<> service_loop_3() {
+    seastar::listen_options lo;
+    lo.reuse_address = true;
+    auto listener = seastar::listen(seastar::make_ipv4_address({1234}), lo);
+    while (true) {
+        auto res = co_await listener.accept();
+        (void) handle_connection(std::move(res.connection));
+    }
+}
+```
+
 # Sharded services
 
 In the previous section we saw that a Seastar application usually needs to run its code on all available CPU cores. We saw that the `seastar::smp::submit_to()` function allows the main function, which initially runs only on the first core, to start the server's code on all `seastar::smp::count` cores.


doc/tutorial.md
```diff
@@ -1778,6 +1778,40 @@ It is often a mistake to silently ignore an exception, so if the future we're ig
 
 The ```handle_connection()``` function itself is straightforward --- it repeatedly calls ```read()``` read on the input stream, to receive a ```temporary_buffer``` with some data, and then moves this temporary buffer into a ```write()``` call on the output stream. The buffer will eventually be freed, automatically, when the ```write()``` is done with it. When ```read()``` eventually returns an empty buffer signifying the end of input, we stop ```repeat```'s iteration by returning a ```stop_iteration::yes```.
 
+Re-written using C++20's coroutines, the above becomes this:
+
```
avikivity, line 1782: IMO it should replace the original. We encourage coroutines now (except in one case - when the function usually resolves with a ready future and it is very performance sensitive). The original is very outdated, no one should use keep_doing().

doc/tutorial.md
```diff
@@ -1778,6 +1778,40 @@ It is often a mistake to silently ignore an exception, so if the future we're ig
...
+    }
+}
+
+seastar::future<> service_loop_3() {
+    seastar::listen_options lo;
+    lo.reuse_address = true;
+    auto listener = seastar::listen(seastar::make_ipv4_address({1234}), lo);
+    while (true) {
+        auto res = co_await listener.accept();
+        (void) handle_connection(std::move(res.connection));
```
avikivity, line 1810: Ignoring futures is bad practice in a real server it will lead to running out of memory, or to problems during shutdown. I'll accept it since it's so in the original, but at least add comments about it.

doc/tutorial.md
nyh: I think eventually we need to go over the tutorial and reorganize/rewrite it in a way that gives more explanations and examples using coroutines - and only later in a separate section introduces the finer points of continuations, how to use them, when to use them, and so on. In the continuations section we'll need some examples, including of keep_doing, but it doesn't need to be this specific server.

doc/tutorial.md
Dobiasd: Thanks for the remark. Just [added the comment](https://github.com/scylladb/seastar/pull/2173/commits/3aa1710d06e1752cb9f0edfda5a65ab010afb830). :heavy_check_mark:

doc/tutorial.md
```diff
@@ -1778,6 +1778,40 @@ It is often a mistake to silently ignore an exception, so if the future we're ig
...
+Re-written using C++20's coroutines, the above becomes this:
+
+```cpp
+seastar::future<> handle_connection(seastar::connected_socket s) {
+    try {
+        auto out = s.output();
+        auto in = s.input();
+        while (true) {
+            auto buf = co_await in.read();
+            if (buf) {
```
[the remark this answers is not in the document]
Dobiasd: Oh, what a great idea! :heart:
I just [committed this change](https://github.com/scylladb/seastar/pull/2173/commits/ea5b7a3b5e98b9ede805bd72dea50045cc35db25). :heavy_check_mark:

doc/tutorial.md
Dobiasd: Given the following goals:
- A: Remove the continuation-based version of the echo server and adjust the explanation to match the coroutine-based version.
- B: Convert the whole tutorial from continuations to coroutines.

I guess we have the following options for this PR here:
- Discard this PR (close unmerged).
- Merge it as it is (keep A and B for later)
- I do A in this PR.

(Doing B currently feels way above my understanding/skills.)

Dobiasd:
If you have any suggestions on what I can do to help move this PR forward, please let me know. In case it's currently not wanted, and you'd prefer to close this unmerged, that's also ok for me. :relaxed:

avikivity:
> If you have any suggestions on what I can do to help move this PR forward, please let me know. In case it's currently not wanted, and you'd prefer to close this unmerged, that's also ok for me. ☺️

Replace the continuations based example with the coroutine. There's no reason to show both.

Dobiasd:
Thanks! I just pushed [this change](https://github.com/scylladb/seastar/pull/2173/commits/8ba25cf17326fd293c0327b3259eac8367474e76) which also adjusts the explanation text accordingly. :heavy_check_mark:

Dobiasd:
If there is any way I can help move this forward, please let me know.

PR #2174, opened 2024-04-02
stall-analyser: add output support for flamegraph

tchaikov:
a typical session is like:

```console
$ ./stall-analyser.py --format trace -e scylla/libexec/scylla \
  reactor_stalled_231d31df_2022_2_twcs_3h.log  > out.folded
$ flamegraph.pl out.folded > /tmp/folded.svg
$ xdg-open /tmp/folded.svg
```

### DIFF
--- a/scripts/stall-analyser.py
+++ b/scripts/stall-analyser.py
@@ -5,8 +5,9 @@ import sys
 import re
 
 import addr2line
-from itertools import dropwhile
-from typing import Self
+from collections import defaultdict
+from itertools import chain, dropwhile
+from typing import Iterator, Self
 
 
 def get_command_line_parser():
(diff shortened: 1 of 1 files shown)


bhalevy:
Can you please upload an example of the result?

tchaikov:
the log file and executable come from https://github.com/scylladb/scylladb/issues/11971
```console
$ ./stall-analyser.py --format trace -e scylla/libexec/scylla \
  reactor_stalled_231d31df_2022_2_twcs_3h.log | \
  flamegraph.pl --title 'Reactor Stalls' --countname "ms" > /tmp/folded.svg
$ xdg-open /tmp/folded.svg
```
[folded.log](https://github.com/scylladb/seastar/files/14845814/folded.log)
folded

tchaikov:
@bhalevy hi Benny, if the general direction is sane. we could render an SVG on the fly without using the flamegraph, this way, we could customize the blocks further. but before that i will try to work on our in-house web service to hook up with this tool.

bhalevy:
It looks nice visually, but I don't see how it helps understanding the stack traces any better.
Seems distracting to me.

tchaikov:
@bhalevy hi Benny, thank you for your review. i think it does not provide more information, even worse, it provide less information than the text output. but the reason i added this output is not just for offering an eye candy. 

with a graph we can identify the the hotspot with a quick glance. then we can continue looking at the text output and the logging messages. it serves the same purpose of https://www.brendangregg.com/flamegraphs.html 

i am extracting the refactor part out into #2178. would be great to make stall-analyser an importable module in future.

PR #2179, opened 2024-04-12
file: Add directory_entry::is_hidden() helper

xemul:
It checks if the file is "hidden" i.e. has its name starting with a dot. Scylla code checks that in some places and this helper is just for verbosity of those

### DIFF
--- a/include/seastar/core/file.hh
+++ b/include/seastar/core/file.hh
@@ -61,6 +61,9 @@ struct directory_entry {
     sstring name;
     /// Type of the directory entry, if known.
     std::optional<directory_entry_type> type;
+
+    /// Checks if the file is hidden or not
+    bool is_hidden() const noexcept { return name.starts_with("."); }
 };
 
 /// Filesystem object stat information


include/seastar/core/file.hh
```diff
@@ -61,6 +61,9 @@ struct directory_entry {
     sstring name;
     /// Type of the directory entry, if known.
     std::optional<directory_entry_type> type;
+
+    /// Checks if the file is hidden or not
+    bool is_hidden() const noexcept { return name[0] == '.'; }
```
tchaikov, line 66: probably could use `name.starts_with(".")`, so it can survive the test like
```cxx
directory_entry e;
e.is_hidden();
```

xemul: Fair point. Fixed

xemul:

tchaikov (approved):
lgtm

include/seastar/core/file.hh
```diff
@@ -61,6 +61,9 @@ struct directory_entry {
     sstring name;
     /// Type of the directory entry, if known.
     std::optional<directory_entry_type> type;
+
+    /// Checks if the file is hidden or not
+    bool is_hidden() const noexcept { return name.starts_with("."); }
 };
```
avikivity, line 67: If we're converting it from a struct to struct+member functions, we need stronger reasons.

nyh:
I have a sacrilegious question: Why do we need a function to encapsulate the super-complicated code `name.starts_with('.')` ?

It's not a coincidence that the standard C library never had a function is_hidden. The idea that files whose name begin in a dot are "hidden" is a convention of the "ls" tool and very little else: "find" shows them just any other file, so does boost's directory iterators and almost everything else. I just checked, and even the Gnome file managers shows all those files and directories beginning with dot (although, strangely, it seems to make a special effort to move them to the end of the alphabetized list instead of the top of the list where ASCII would put them).  I personally don't see any reason why Seastar needs to officially call these files "hidden", especially since open-coding `starts_with('.')` is so easy.

xemul:
> I have a sacrilegious question: Why do we need a function to encapsulate the super-complicated code name.starts_with('.') ?

Because it's simply less keystrokes
```
de.name.starts_with(".")
de.name[0] == '.'
de.is_hidden()
```
the thing in the middle is how Scylla checks if a file is hidden

> The idea that files whose name begin in a dot are "hidden" is a convention of the "ls" tool and very little else

Apparently Scylla is in this very little list of exceptions for some reason :D

> I personally don't see any reason why Seastar needs to officially call these files "hidden", especially since open-coding starts_with('.') is so easy.

OK

Issue #1298, opened 2022-11-17
with_file_input_stream doesn't close file


xemul:
as per fe8fcd6d2fded7c94bd3e8fc3e489e3d9ba060ab
```c++
auto with_file_input_stream(const std::filesystem::path& path, Func func, file_open_options file_opts = {}, file_input_stream_options input_stream_opts = {}) {
    static_assert(std::is_nothrow_move_constructible_v<Func>);
    return open_file_dma(path.native(), open_flags::ro, std::move(file_opts)).then(
            [func = std::move(func), input_stream_opts = std::move(input_stream_opts)] (file f) mutable {
        return do_with(make_file_input_stream(std::move(f), std::move(input_stream_opts)),
                [func = std::move(func)] (input_stream<char>& in) mutable {
            return futurize_invoke(std::move(func), in).finally([&in] {
                return in.close();
            });
        });
    }); 
}
```

the file is move-d into input stream, but closing the stream doesn't close the file, respectively file is not closed (however, file_impl-s are supposed to close the underlying descriptor in destructor)

cc @bhalevy

bhalevy:
Hmm, it seems like file_data_source_impl::close should close its _file member.
Is there a reproducer for this issue?
Best if we had a unit test for it.

xemul:
> Hmm, it seems like file_data_source_impl::close should close its _file member.

Of course, the stream itself doesn't have file on board, the issue is rather about data source

> Is there a reproducer for this issue?

Hmm, I don't think there is one. But what kind of "reproducer" should it be? Is there any landmine we can step on if leaving a file non-closed?

bhalevy:
> > Hmm, it seems like file_data_source_impl::close should close its _file member.
> 
> Of course, the stream itself doesn't have file on board, the issue is rather about data source
> 
> > Is there a reproducer for this issue?
> 
> Hmm, I don't think there is one. But what kind of "reproducer" should it be? Is there any landmine we can step on if leaving a file non-closed?

The critical part is cancelling outstanding reads and waiting for them to resolve before destroying the file.
Otherwise, we risk use after free.

niekbouman:
> [bhalevy]: Hmm, it seems like file_data_source_impl::close should close its _file member.

Could it be that you were confusing `file_data_sink_impl::close` and `file_data_source_impl::close`?

`file_data_sink_impl` does close the file in its `close` virtual method:
https://github.com/scylladb/seastar/blob/master/src/core/fstream.cc#L492

[...]

PR #2186, opened 2024-04-14
util: with_file_input_stream: always close file

bhalevy:
The file opened in `with_file_input_stream` is currently moved to `make_file_input_stream`, but the latter doesn't close it when the input_stream is closed.

This change turns `with_file_input_stream` into a
coroutine to simplify the close logic, as well
amortize allocations, and then make sure that
the opened file is always closed, on the error
path as well as the error-free path.

Fixes scylladb/seastar#1298

### DIFF
--- a/include/seastar/util/file.hh
+++ b/include/seastar/util/file.hh
@@ -27,6 +27,8 @@
 #include <seastar/core/future.hh>
 #include <seastar/core/fstream.hh>
 #include <seastar/core/sstring.hh>
+#include <seastar/core/coroutine.hh>
+#include <seastar/coroutine/as_future.hh>
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/short_streams.hh>
 #include <seastar/util/modules.hh>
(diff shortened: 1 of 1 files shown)


include/seastar/util/file.hh
```diff
@@ -64,20 +66,26 @@ namespace util {
...
-            [func = std::move(func), input_stream_opts = std::move(input_stream_opts)] (file f) mutable {
-        return do_with(make_file_input_stream(std::move(f), std::move(input_stream_opts)),
-                [func = std::move(func)] (input_stream<char>& in) mutable {
-            return futurize_invoke(std::move(func), in).finally([&in] {
-                return in.close();
-            });
-        });
-    });
+    auto f = co_await open_file_dma(path.native(), open_flags::ro, std::move(file_opts));
+    input_stream<char> in;
```
xemul, line 73: nit: It's a pity that input_stream<> is default-constructed, then move-assigned. Hopefully compiler can optimize it

bhalevy: We can contain it in an optional but I'm not sure it's worth it.

include/seastar/util/file.hh
```diff
@@ -64,20 +66,26 @@ namespace util {
...
+    }
+    if (ex) {
+        co_await f.close();
+        co_await coroutine::return_exception_ptr(std::move(ex));
+    }
+
+    auto res = co_await coroutine::as_future(futurize_invoke(std::move(func), in));
+    co_await in.close();
+    co_await f.close();
+    co_return co_await std::move(res);
```
xemul, line 88: Not a blocked for this patch, but a general question -- what's the difference between `co_await future` and `co_await std::move(future)`? I checked assembly for similar yet different code, namely `return fut.finally(...);` vs `return std::move(fut).finally(...)` , but it was identical

bhalevy: Can you co_await future without moving it to the co_awaiter object?

bhalevy: > I checked assembly for similar yet different code, namely return fut.finally(...); vs return std::move(fut).finally(...) , but it was identical

They are since you return the future returned by .finally, not the first future in the continuation chain, so it doesn't matter if it's moved or not.

include/seastar/util/file.hh
```diff
@@ -64,20 +66,26 @@ namespace util {
...
+    } catch (...) {
+        ex = std::current_exception();
+    }
+    if (ex) {
+        co_await f.close();
+        co_await coroutine::return_exception_ptr(std::move(ex));
+    }
+
+    auto res = co_await coroutine::as_future(futurize_invoke(std::move(func), in));
+    co_await in.close();
```
raphaelsc, line 86: read aheads are emitted without yielding. if we get an exception before future is returned, wouldn't we miss close() call above? which could lead to use-after-free.

bhalevy: futurize_invoke would return an exceptional future in this case and the exception will be returned by the co_return below.

raphaelsc (approved):

PR #2187, opened 2024-04-15
tests: coroutines_test: add reproducers for #1913

tchaikov:
again even if the previous returned `yield_awaiter` is not ready, and that not-ready awaiter has not been resumed. another problem is that, the awaiter in generator coroutine fails to keep the async buffer alive when caller of async generator either drains the generator or returns without draining the generator. either way, some test cases are added in this test for two purposes:

1. as reproducers, they can be used to verify their fixes
2. these test cases should help to prevent future regressions after the fixes get included.

Refs #1913


tchaikov:
superseded by #2218

PR #2188, opened 2024-04-15
future: deprecate future::get0()

tchaikov:
In the days when futures carried tuples, it was useful to get
the first/only element. But it's now a synonym for get(), so
deprecate it.

Also, all future::get0() calls were dropped in place of future::get()
in 1504dc0e, so Seastar should be free of deprecation warnings
due to this change.

Kefu: amend the comment to highlight the difference between `get0()`
and `get()`, to help developers to switch from `get0()` to `get()`.

### DIFF
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -1354,12 +1354,25 @@ public:
     /// Gets the value returned by the computation.
     ///
     /// Similar to \ref get(), but instead of returning a
-    /// tuple, returns the first value of the tuple.  This is
-    /// useful for the common case of a \c future<T> with exactly
-    /// one type parameter.
+    /// \c T&&, returns \c T.
     ///
-    /// Equivalent to: \c std::get<0>(f.get()).
+    /// \note The \c get0() method is deprecated. It's a remnant from older
+    /// versions of Seastar that supported variadic futures, capable of
+    /// returning multiple values through a tuple. Back then, \c get0() served
+    /// the purpose of retrieving the first (and usually the only) value.
+    /// Today, the \ref get() method accomplishes the same task. However,
+    /// there's a subtle difference in return types: \c get0() returned
+    /// \c T, while \ref get() returns \c T&& (an rvalue reference to
+    /// \c T). This distinction typically won't cause issues when switching
+    /// from \c get0() to \ref get(). However, in specific metaprogramming
+    /// scenarios, especially when the code expects type \c T, you'll need
+    /// to use \c std::remove_reference_t<decltype(fut.get())> to extract
+    /// the underlying type \c T.
+    /// For new code that utilizes \c future<tuple<...>>, employ
+    /// \c std::get<0>(fut.get()) to access the first element of the tuple,
+    /// rather than the deprecated \ref get0().
     using get0_return_type = typename future_state::get0_return_type;
+    [[deprecated("Use get() instead")]]
     get0_return_type get0() {
         return (get0_return_type)get();
     }


tchaikov:
this change is actually #2059 rebased against master HEAD. it is replicated in order to have a passing CI, and to get the change merged into master sooner.

nyh:
I asked the following questions in my original review, maybe you can answer them here:

1. I'm in favor of this deprecation, but I wonder if it wouldn't cause unnecessary headaches to Seastar projects which will now need to change all the calls to get0() or face hundreds of warning messages. Could we make "get0()" an exact synonym of get()" - with less code (no separate return type, etc.) and then we won't mind just leaving it there for years, un-deprecated but with a comment deterring new users?

2.  I think you forgot to through Seastar itself and replace all calls to get0() by get() - we have many of those in various helper classes, in example applications, and so on. Didn't you get a lot of deprecation warnings when building Seastar with this patch? Shouldn't this be fixed?

tchaikov:
> I asked the following questions in my original review, maybe you can answer them here:
> 
>     1. I'm in favor of this deprecation, but I wonder if it wouldn't cause unnecessary headaches to Seastar projects which will now need to change all the calls to get0() or face hundreds of warning messages.

yeah, that's indeed a very valid concern. and scylladb has gone through this. see https://github.com/scylladb/scylladb/pull/17130 . and, unfortunately, despite that it is a mechanical replacing process under most circumstances, that's not always the case.

> Could we make "get0()" an exact synonym of get()" - with less code (no separate return type, etc.) and then we won't mind just leaving it there for years, un-deprecated but with a comment deterring new users?

i like the result of "less code". but i am afraid that we cannot get there by making `get0()` an alias of `get()`. it's worse than deprecation because it would be a breaking change -- there are some subtle differences between `get0()` and `get()`. that's why https://github.com/scylladb/scylladb/pull/17130 has 3 commits instead of one. so i am still in favor of doing the dance of deprecation and removal.

> 
>     2. I think you forgot to through Seastar itself and replace all calls to get0() by get() - we have many of those in various helper classes, in example applications, and so on. Didn't you get a lot of deprecation warnings when building Seastar with this patch? Shouldn't this be fixed?

Avi addressed this in a separate PR, see #2069. without which, this PR would face the same build failure as #2059 did. see https://app.circleci.com/pipelines/github/avikivity/seastar/295/workflows/c4f1323d-4c0b-4c35-88c0-c14a73e8fd60/jobs/1463 for an example.

nyh:
Thanks. I wasn't aware that #2069 already went it (it didn't when I reviewed Avi's patch) - please mention it in the commit message. Indeed, "git grep get0" in the Seastar source tree returns nothing except future.hh.
Also, you mentioned above some "subtle differences between get0() and get()" - please explain these differences and give examples in a comment next to the deprecation of get0(), so that users encountering these deprecation warnings would not do what I would have intuitively done - remove the "0" everywhere and assume or hope that would solve everything.

tchaikov:
v2:

- reference the commit of 1504dc0ee2805a6b86e4c2d9d865fe305fd34c34 which replaced `get()` to `get0()`.
- add a note in comment to explain the subtle difference between `get()` and `get0()`

tchaikov:
@nyh hi Nadav, thank you for your review and suggestions. could you take another look?

include/seastar/core/future.hh
```diff
@@ -1354,12 +1354,15 @@ public:
...
     ///
     /// Similar to \ref get(), but instead of returning a
-    /// tuple, returns the first value of the tuple.  This is
-    /// useful for the common case of a \c future<T> with exactly
-    /// one type parameter.
+    /// \c T&&, returns \c T.
     ///
-    /// Equivalent to: \c std::get<0>(f.get()).
+    /// \note Under most circumstances, it is equivalent to:
+    /// \c std::get<0>(f.get()). But when performing meta programming,
```
nyh, line 1360: I'm not sure what you're saying here. In "most circumstances", the future's type T won't be a tuple at all, so `get<0>(f.get())` won't even work. I'm guessing this example worked in the past, but not today?

Maybe it's better to more explicitly say something like the following: (it's just something I quickly wrote, feel free to completely reword):

The get0() method is deprecated, left over from earlier versions of Scylla that had variadic futures - futures that returned multiple values. Back then, get0() was useful for getting the first (and usually, only) value. Today just get() does the same thing, although because of the different return type (get0() returned T, get() returns T&&) some metaprogramming will need to use (... insert details here). In new code using a future<tuple<...>>, use `std::get<0>(f.get())` instead of `f.get0()` to get the first component of the tuple.

tchaikov:
@nyh hi Nadav, could you please take another look?

nyh (approved):
Thanks, the comment is now very clear and (I hope) helpful.

Issue #2180, opened 2024-04-13
The link interface of target "Seastar::seastar" contains: Valgrind::valgrind but the target was not found.


robinchrist:
I'm getting the following error when building b74a027c819ee5b9bfcee4f32edf8f37017bba12 and cd8a9133d2c02f63dbd578d882cf7333a427e194 (only tried those two):

```
cmake -DCMAKE_PREFIX_PATH="/data/dev/Kataklysmos/external/seastar/build/release;/data/dev/Kataklysmos/external/seastar/build/release/_cooking/installed" -DCMAKE_MODULE_PATH="/data/dev/Kataklysmos/external/seastar/cmake" -DCMAKE_BUILD_TYPE="Release" -GNinja /data/dev/Kataklysmos
-- The C compiler identification is GNU 13.2.0
-- The CXX compiler identification is GNU 13.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Boost: /usr/include (found suitable version "1.74.0", minimum required is "1.73.0")  
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found Boost: /usr/include (found suitable version "1.74.0", minimum required is "1.73.0") found components: filesystem program_options thread unit_test_framework chrono date_time atomic 
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.8.1") 
-- Found c-ares: /usr/lib/x86_64-linux-gnu/libcares.so (found suitable version "1.19.1", minimum required is "1.13") 
-- Checking for module 'libdpdk'
--   Found libdpdk, version 23.07.0
-- Found dpdk: /data/dev/Kataklysmos/external/seastar/build/release/_cooking/installed/include  
-- Performing Test _warning_supported_volatile
-- Performing Test _warning_supported_volatile - Success
-- Found lz4: /usr/lib/x86_64-linux-gnu/liblz4.so (found suitable version "1.9.4", minimum required is "1.7.3") 
[...]

PR #2189, opened 2024-04-15
build: do not export valgrind with export()

tchaikov:
in Seastar, valgrind library is only used at the build time, and is used internally for accessing its header files when buildings its internal .cc files. so we don't expect to expose it via `export()`, for exposing it in `SeastarTargets.cmake`.

but somehow when building Seastar as a static library, valgrind is still exposed in Seastar's CMake configuration file. if a CMake based project consume Seastar using its CMake configuration file, it runs into following failure when CMake generates the building system:

```
-- Configuring done (2.5s)
CMake Error at external/seastar/build/release/SeastarTargets.cmake:52 (set_target_properties):
  The link interface of target "Seastar::seastar" contains:

    Valgrind::valgrind

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

  external/seastar/build/release/SeastarConfig.cmake:36 (include)
  CMakeLists.txt:6 (find_package)
```

fortunately, in CMake 3.26, `BUILD_LOCAL_INTERFACE` was introduced this problem in specific, so that the linked library is not included by `export()` anymore.

in this change, we include `Valgrind::valgrind` using this generator expression if the CMake's version is greater or equal to 3.26, so that `Valgrind` is not required for building Seastar applications.

as this generator expression is not available on CMake versions lower than 3.26, and the minimal required CMake version for building Seastar is 3.13, and Seastar does support static build. The issue is still not fully addressed even with this change with older CMake and if Seastar is built as a static library.

Refs #2180


tchaikov:
@robinchrist hi Robin, could you help test this change?

tchaikov:

avikivity:
We should consider dropping valgrind. Is there anything it does that asan/ubsan does not?

robinchrist:
Yep I can confirm this fix is working! Thanks a lot @tchaikov

nyh:
> We should consider dropping valgrind. Is there anything it does that asan/ubsan does not?

Four years ago, after someone who was a valgrind fan decided that Seastar must depend on it, I wrote a patch in https://groups.google.com/g/seastar-dev/c/f0WfaK0KgZ0/m/qK2WlDMrCAAJ making this dependency optional - i.e., if you (a Seastar developer) already have valgrind installed, it will be compiled in a way that you can use valgrind on it. But if you don't have valgrind installed? That's fine, and you can still compile without it.

But there was fierce opposition to my patch and it never got in.

Maybe it's time to reconsider - my original patch or something else with the same effect of making valgrind optional.

tchaikov:
> > We should consider dropping valgrind. Is there anything it does that asan/ubsan does not?
> 
> Four years ago, after someone who was a valgrind fan decided that Seastar must depend on it, I wrote a patch in https://groups.google.com/g/seastar-dev/c/f0WfaK0KgZ0/m/qK2WlDMrCAAJ making this dependency optional - i.e., if you (a Seastar developer) already have valgrind installed, it will be compiled in a way that you can use valgrind on it. But if you don't have valgrind installed? That's fine, and you can still compile without it.
> 
> But there was fierce opposition to my patch and it never got in.
> 
> Maybe it's time to reconsider - my original patch or something else with the same effect of making valgrind optional.

if it's desirable to make valgrind optional, i can add a `Seastar_VALGRIND` option. and i concur with Raphael and Pekka on `#if __has_include()` is not the cure of this issue (if it is).

PR #2193, opened 2024-04-16
Add Safety to the metrics layer

### LINKED ISSUE #2191: metrics: trying to add summaries is unsafe
It does not make sense to aggregate a summary, but if some have a reason to do it, the code should not throw as it does today.
Instead, it should act like a histogram

amnonh:
This series addresses two corner cases in the metrics layer:
1. use the addition operator on a Summary; while this does not make sense, a user may still choose to do that; it should not throw an exception and treat the summary as a histogram.
2. There could be a situation when calling set_metric_family_configs while the _metadata object does not exist, either because it was not created yet or because it was deleted and not re-created. In this situation, we should leave it to the next time the _metadata will be created.

Fixes #2191

### DIFF
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -226,6 +226,7 @@ metric_value metric_value::operator+(const metric_value& c) {
     metric_value res(*this);
     switch (_type) {
     case data_type::HISTOGRAM:
+    case data_type::SUMMARY:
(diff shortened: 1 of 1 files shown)


nyh (approved):
The patches look good to me, even though I'm not sure why it's good to implement addition that you think doesn't make sense, it's at least better than having the wrong exception.

Issue #123, opened 2016-03-10
Add hold() method for futures


nyh:
This suggestion is due to Gleb, raised in this thread: https://groups.google.com/d/msg/seastar-dev/7fjd7kJiIOQ/cV3tTGGpIwAJ

We commonly use finally() in the following idiom:

``` c++
   fut.finally([obj = std::move(obj)] {});
```

to hold an object alive until a future is concluded.
This idiom is common enough that we started to worry that finally is inefficient in this case (see issue #92).

There are times when we need the full finally(), of course, but Gleb's idea was that for this specific use case, it would be better to have its own method, hold(T&&):

``` c++
    fut.hold(std::move(obj));
```

In a mail in the above thread, I tried to enumerate the benefits of this hold() over the existing finally()-based idiom, and another proposal of a "with_shared()" function:

The hold() syntax,

``` c++
    auto fut = async_call(obj);
    fut.hold(std::move(obj));
```

is cleaner and more natural than the baroque syntax we currently recommend:

``` c++
    auto fut = async_call(obj);
    fut.finally([obj = std::move(obj)] { });
```

Avi's alternative suggestion of "with_shared" is even simpler in this example:

``` c++
    with_shared(obj, async_call);
```

However, the with_shared syntax become much uglier if obj is not the only parameter you need to pass to async_call - e.g., consider:

``` c++
     async_call(i, j, k, obj).hold(std::move(obj));
```

while with with_shared, you find yourself having to write this:

``` c++
     with_shared(ptr, [i, j, k] (auto& obj ) { return async_call(i, j, k, obj); });
```

It should be carefully documented that "hold" has a limitation on what kind of objects it can hold: For example, imagine you try to do:

``` c++
    future<> async_call(T&);
    T obj;
    return async_call(obj).hold(std::move(obj));
```

Is this safe? It is _NOT_ safe if obj and async_call are such that moving the object after calling async_call(obj) can cause async_call to later (as part of some later async processing) access obj again, after it was moved. It _is_ safe for all T which are "smart pointers and similar" because with smart pointers, async_call remembers the pointed-to object object, and never tries to look at the content of the pointer object itself after the function returns.
I don't know how we can formalize this requirement from T.
In particular, consider this question can T be "class file"? Can someone do:

``` c++
    file f;
    return f.close().hold(std::move(f));
```

[...]

PR #2194, opened 2024-04-16
core/sleep: Use more raii-sh aproach to maintain sleeper

xemul:
The sleeper helper struct is allocated and deleted by hand. Using unique_ptr is the same.

Also replace .then() with .finally(). The future in question never fails, but finally is more idiomatic to guard object lifetime

### DIFF
--- a/include/seastar/core/sleep.hh
+++ b/include/seastar/core/sleep.hh
@@ -55,9 +55,9 @@ future<> sleep(std::chrono::duration<Rep, Period> dur) {
             tmr.arm(dur);
         }
     };
-    sleeper *s = new sleeper(dur);
+    auto s = std::make_unique<sleeper>(dur);
     future<> fut = s->done.get_future();
-    return fut.then([s] { delete s; });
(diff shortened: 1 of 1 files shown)


tchaikov (approved):
lgtm

nyh (approved):
Looks correct, so I'm not against this patch, but maybe you can say what was your motivation to even touch this ancient code. Was it only style improvements or were you worried about something else?

I do agree that finally() has become the "idiomatic" way to hold on to something while a future is running (I proposed https://github.com/scylladb/seastar/issues/123, but it wasn't accepted). Also "new" is really old-style C++ that we don't want to see in code. So I approve of this patch.

xemul:
> Was it only style improvements or were you worried about something else?

Well, initially I started to look at the ability to inject random delays into continuation chains (in debug mode, like there's runqueue shuffling). Eventually I came across this sleep() helper and new/delete pair caught my eye because vim highlights it by default. So it's just a style improvement

PR #2196, opened 2024-04-17
shared_mutex: Add RAII locks

dawmd:
In this PR, we introduce RAII locks for `seastar::shared_mutex`, similar to `seastar::semaphore_units` and `seastar::gate::holder`.

### DIFF
--- a/include/seastar/core/shared_mutex.hh
+++ b/include/seastar/core/shared_mutex.hh
@@ -22,11 +22,14 @@
 #pragma once
 
 #ifndef SEASTAR_MODULE
+#include <seastar/core/coroutine.hh>
 #include <seastar/core/future.hh>
 #include <seastar/core/chunked_fifo.hh>
 #include <seastar/util/modules.hh>
 #include <cassert>
+#include <mutex>
 #include <utility>
+#include <shared_mutex>
 #endif
 
 namespace seastar {
@@ -231,6 +234,69 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
         });
     } catch (...) {
         return futurize<std::invoke_result_t<Func>>::current_exception_as_future();
+    }
+}
+
+namespace internal {
+
+/// Seastar analogue of the `BasicLockable` named requirement, see:
+///   [https://eel.is/c++draft/thread.req.lockable.basic#def:Cpp17BasicLockable]
+template <typename T>
+concept FutureBasicLockable = requires (T& t) {
+    { t.lock() } -> std::same_as<future<>>;
+    { t.unlock() } noexcept -> std::same_as<void>;
+};
+
+/// Seastar analogue of the `SharedLockable` named requirement, see:
+///   [https://eel.is/c++draft/thread.req.lockable.shared#def:Cpp17SharedLockable]
+/// It's called `basic` because the concept doesn't require the type
+/// to define `try_lock_shared()` method to satisfy the constraint, similarly
+/// to the `BasicLockable` named requirement.
+template <typename T>
+concept FutureBasicSharedLockable = requires (T& t) {
+    { t.lock_shared() } -> std::same_as<future<>>;
+    { t.unlock_shared() } noexcept -> std::same_as<void>;
+};
+
+} // namespace internal
+
+/// \brief Construct a RAII-based shared lock corresponding to a given object.
+///
+/// Since constructors cannot be exectued asynchronously, use this function
+/// to shared lock the passed object and construct an std::shared_lock
+/// that will unlock it when the lock is being destroyed.
+///
+/// The caller is responsible for keeping the passed object
+/// alive at least until the returned lock is destroyed.
+template <internal::FutureBasicSharedLockable T>
+future<std::shared_lock<T>> get_shared_lock(T& t) {
+    co_await t.lock_shared();
+
+    try {
+        co_return std::shared_lock<T>{t, std::adopt_lock_t{}};
+    } catch (...) {
+        t.unlock_shared();
+        throw;
+    }
+}
+
+/// \brief Construct a RAII-based unique lock corresponding to a given object.
+///
+/// Since constructors cannot be exectued asynchronously, use this function
+/// to lock the passed object and construct an std::unique_lock
+/// that will unlock it when the lock is being destroyed.
+///
+/// The caller is responsible for keeping the passed object
+/// alive at least until the returned lock is destroyed.
+template <internal::FutureBasicLockable T>
+future<std::unique_lock<T>> get_unique_lock(T& t) {
+    co_await t.lock();
+
+    try {
+        co_return std::unique_lock<T>{t, std::adopt_lock_t{}};
+    } catch (...) {
+        t.unlock();
+        throw;
     }
 }
 
--- a/tests/unit/locking_test.cc
+++ b/tests/unit/locking_test.cc
@@ -25,6 +25,8 @@
 #include <exception>
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
+#include <seastar/core/coroutine.hh>
+#include <seastar/coroutine/parallel_for_each.hh>
 #include <seastar/core/do_with.hh>
 #include <seastar/core/loop.hh>
 #include <seastar/core/sleep.hh>
@@ -419,4 +421,73 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
             BOOST_FAIL(format("Unexpected exception type: {}", e.what()));
         }
     }
+}
+
+SEASTAR_TEST_CASE(test_shared_mutex_locks) {
+    shared_mutex sm;
+
+    {
+        const auto ulock = co_await get_unique_lock(sm);
+        BOOST_REQUIRE(!sm.try_lock());
+        BOOST_REQUIRE(!sm.try_lock_shared());
+    }
+
+    BOOST_REQUIRE(sm.try_lock());
+    sm.unlock();
+
+    {
+        const auto slock1 = co_await get_shared_lock(sm);
+        BOOST_REQUIRE(!sm.try_lock());
+        BOOST_REQUIRE(sm.try_lock_shared());
+
+        const auto slock2 = co_await get_shared_lock(sm);
+
+        // This balances out the call to `try_lock_shared()` above.
+        sm.unlock_shared();
+    }
+
+    BOOST_REQUIRE(sm.try_lock());
+    sm.unlock();
+}
+
+SEASTAR_TEST_CASE(test_shared_mutex_exclusive_locks) {
+    shared_mutex sm{};
+    unsigned counter = 0;
+
+    co_await coroutine::parallel_for_each(boost::irange(0, 10), coroutine::lambda([&sm, &counter] (auto&&) -> future<> {
+        const auto ulock = co_await get_unique_lock(sm);
+        BOOST_REQUIRE_EQUAL(counter, 0u);
+        ++counter;
+        co_await sleep(1ms);
+        --counter;
+        BOOST_REQUIRE_EQUAL(counter, 0u);
+    }));
+}
+
+SEASTAR_TEST_CASE(test_shared_mutex_exception_locks) {
+    shared_mutex sm;
+
+    // Verify that the shared_mutex is unlocked when an exception is thrown.
+    try {
+        const auto slock = co_await get_shared_lock(sm);
+        throw std::runtime_error("injected");
+    } catch (const std::runtime_error&) {
+        BOOST_REQUIRE(sm.try_lock());
+        sm.unlock();
+    } catch (...) {
+        BOOST_FAIL(format("Unexpected exception type: {}", std::current_exception()));
+    }
+
+    try {
+        const auto ulock = co_await get_unique_lock(sm);
+        throw std::runtime_error("injected");
+    } catch (const std::runtime_error&) {
+        BOOST_REQUIRE(sm.try_lock());
+        sm.unlock();
+    } catch (...) {
+        BOOST_FAIL(format("Unexpected exception type: {}", std::current_exception()));
+    }
+
+    BOOST_REQUIRE(sm.try_lock());
+    sm.unlock();
 }


include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +234,122 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+/// happens, the destructor should eventually be called.
+///
+/// The corresponding \ref shared_mutex object MUST be kept alive
+/// throughout the lifetime of this object.
+///
+/// The class is NOT responsible for shared locking a \ref shared_mutex.
+/// \see get_shared_lock(shared_mutex&)
+/// \see unique_lock
+class shared_lock {
+private:
```
bhalevy, line 253: too many "private:"s :)
why do we need all (or any) of them?

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +234,122 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+/// locked and to unconditionally release it because no matter what
+/// happens, the destructor should eventually be called.
+///
+/// The corresponding \ref shared_mutex object MUST be kept alive
+/// throughout the lifetime of this object.
+///
+/// The class is NOT responsible for locking a \ref shared_mutex.
+/// \see get_unique_lock(shared_mutex&)
+/// \see shared_lock
+class unique_lock {
```
bhalevy, line 310: The counterpart of shared lock is typically an exclusive lock, not unique.

Let's stay consistent with:
```
    /// Lock the \c shared_mutex for exclusive access
    ///
    /// \return a future that becomes ready when no access, shared or exclusive
    ///         is granted to anyone.
    future<> lock() noexcept {
```

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +234,122 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+private:
+    friend future<unique_lock> get_unique_lock(shared_mutex& mutex);
+private:
+    shared_mutex* _mutex;
+
+private:
```
bhalevy, line 316: Here too, we can consolidate all the "private:"s.

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +234,122 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+    unique_lock(shared_mutex& mutex) noexcept : _mutex{std::addressof(mutex)} {}
+public:
+    unique_lock(unique_lock&& other) noexcept : _mutex{std::exchange(other._mutex, nullptr)} {}
+    unique_lock& operator=(unique_lock&& other) noexcept {
```
bhalevy, line 320: I'm not sure this is really needed.

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +234,122 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+private:
+    friend future<shared_lock> get_shared_lock(shared_mutex& mutex);
+private:
+    shared_mutex* _mutex;
+
+private:
+    shared_lock(shared_mutex& mutex) noexcept : _mutex{std::addressof(mutex)} {}
+public:
+    shared_lock(shared_lock&& other) noexcept : _mutex{std::exchange(other._mutex, nullptr)} {}
+    shared_lock& operator=(shared_lock&& other) noexcept {
```
bhalevy, line 262: I'm not sure that move-assignment is really needed

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +234,122 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+            this->~shared_lock();
+            new (this) shared_lock{std::move(other)};
+        }
+        return *this;
+    }
+
+    shared_lock(const shared_lock&) = delete;
+    shared_lock& operator=(const shared_lock&) = delete;
+
+    ~shared_lock() noexcept {
```
bhalevy, line 273: destructors are noexcept by default, no need for the specifier.

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +234,122 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+        if (_mutex) {
+            _mutex->unlock_shared();
+            _mutex = nullptr;
```
bhalevy, line 276: not really needed in the destructor,
better keep it as is so that the sanitizer can detect use after free if somehow double destruction happens.

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +234,122 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+        return *this;
+    }
+
+    unique_lock(const unique_lock&) = delete;
+    unique_lock& operator=(const unique_lock&) = delete;
+
+    ~unique_lock() noexcept {
+        if (_mutex) {
+            _mutex->unlock();
+            _mutex = nullptr;
```
bhalevy, line 334: ditto

tests/unit/locking_test.cc
```diff
@@ -420,3 +420,171 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
         }
     }
 }
+
+SEASTAR_THREAD_TEST_CASE(test_shared_mutex_locks) {
+    shared_mutex sm;
+
+    {
+        const auto ulock = get_unique_lock(sm).get();
```
bhalevy, line 428: s/unique/exclusive/ everywhere

tests/unit/locking_test.cc
```diff
@@ -420,3 +420,171 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
...
+    BOOST_REQUIRE(sm.try_lock());
+    sm.unlock();
+
+    {
+        const auto slock1 = get_shared_lock(sm).get();
+        BOOST_REQUIRE(!sm.try_lock());
+        BOOST_REQUIRE(sm.try_lock_shared());
+
+        const auto slock2 = get_shared_lock(sm).get();
+        sm.unlock_shared();
```
bhalevy, line 442: please add comment that this balances the successful `try_lock_shared()` call above

tests/unit/locking_test.cc
```diff
@@ -420,3 +420,171 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
...
+    }
+
+    BOOST_REQUIRE(sm.try_lock());
+    sm.unlock();
+}
+
+SEASTAR_TEST_CASE(test_shared_mutex_exclusive_locks) {
+    return do_with(shared_mutex(), unsigned(0), [] (shared_mutex& sm, unsigned& counter) {
```
bhalevy, line 450: Now that we dropped support for c++17 the test can just use coroutines

tests/unit/locking_test.cc
```diff
@@ -420,3 +420,171 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
...
+                    --counter;
+                    BOOST_REQUIRE_EQUAL(counter, 0u);
+                });
+            });
+        });
+    });
+}
+
+SEASTAR_TEST_CASE(test_shared_mutex_shared_locks) {
+    return do_with(shared_mutex(), unsigned(0), unsigned(0), [] (shared_mutex& sm, unsigned& counter, unsigned& max) {
```
bhalevy, line 465: coroutine / thread

bhalevy (changes requested):
see comments inline

xemul:
I think it's worth having SEASTAR_SEMAPHORE_DEBUG analogue for this holder from the very beginning

nyh:
Before I review the specific code, I want to ask whether we could use the standard C++ (C++14) classes `std::shared_lock<seasatar::shared_mutex>` and `std::unique_lock<seastar::shared_mutex>`, instead of implementing our own classes with exactly the same name (shared_lock and unique_lock) to hold the locks.

dawmd:
> Before I review the specific code, I want to ask whether we could use the standard C++ (C++14) classes `std::shared_lock<seasatar::shared_mutex>` and `std::unique_lock<seastar::shared_mutex>`, instead of implementing our own classes with exactly the same name (shared_lock and unique_lock) to hold the locks.

I don't think the APIs of `std::unique_lock` and `std::shared_lock` allow for that, at least not in a simple enough way to prefer that to implementing our own locks. But it's also quite likely I Just don't see how to do it

nyh:
> > Before I review the specific code, I want to ask whether we could use the standard C++ (C++14) classes `std::shared_lock<seasatar::shared_mutex>` and `std::unique_lock<seastar::shared_mutex>`, instead of implementing our own classes with exactly the same name (shared_lock and unique_lock) to hold the locks.
> 
> I don't think the APIs of `std::unique_lock` and `std::shared_lock` allow for that, at least not in a simple enough way to prefer that to implementing our own locks. But it's also quite likely I Just don't see how to do it

std::shared_lock and std::unique_lock should work on any SharedLockable (see https://en.cppreference.com/w/cpp/named_req/SharedLockable). Doesn't our seastar::shared_mutex fit that constraint? If not, why not? What happens when you try?

dawmd:
> std::shared_lock and std::unique_lock should work on any SharedLockable (see https://en.cppreference.com/w/cpp/named_req/SharedLockable). Doesn't our seastar::shared_mutex fit that constraint? If not, why not? What happens when you try?

I don't think it satisfies that constraint. `seastar::shared_mutex::{lock(), lock_shared()}` doesn't block until the mutex can be obtained; it immediately returns a future and only when that future resolves can the mutex be obtained. The type thus only satisifes [the Lockable requirement](https://en.cppreference.com/w/cpp/named_req/Lockable) in my opinion.

I think it might be possible to somehow make it work, but I'm not knowledgeable enough to argue for or against the existence of a solution. Perhaps C++20's coroutines have somehow changed the situation?

Nonetheless, I attempted to make it work, but even this test
```cpp
SEASTAR_THREAD_TEST_CASE(test_standard_unique_lock) {
    shared_mutex sm{};
    unsigned counter = 0;

    parallel_for_each(boost::irange(0, 10), [&] (int idx) -> future<> {
        return async([&] {
            const auto lock = std::unique_lock(sm);
            BOOST_REQUIRE_EQUAL(counter, 0u);
            ++counter;
            sleep(1ms).get();
            --counter;
            BOOST_REQUIRE_EQUAL(counter, 0u);
        });
    }).get();
}
```
fails with the following repeated error:
```
[...]/src/seastar/tests/unit/locking_test.cc(473): fatal error: in "test_standard_unique_lock2": critical check counter == 0u has failed [1 != 0]
```
We could defer the locking when constructing the lock and only then attempt to do it, but `std::shared_lock<Mutex>::{lock(), shared_lock()}` returns a void, so we lose any access to the underlying future we should await.

include/seastar/core/shared_mutex.hh
nyh: The C++ standard library has std::shared_lock and std::unique_lock, and I assume this is wheret his choice of names comes from. The questions I raised in my review was why can't we use std::shared_lock and std::unique_lock themselves, and not just their names.

dawmd:
> @dawmd yes, surely
> 
> ```
> const auto lock = std::unique_lock(sm);
> ```
> 
> would not work, but does something like this work?
> 
> ```
> co_await sm.lock();
> return std::unique_lock(sm, std::adopt_lock);
> ```
> 
> In this code we only use the std::unique_lock() to wrap an already taken lock - not to actually take it. You'll still have a new function to get the lock - but it will return an std::unique_lock object and not a seastar::unique_lock.
> 
> If it doesn't work I'll give up, and review your code :-)

Thank you. I didn't understand what you meant at first, but you're right.

Version 2:
* Got rid of the redundant lock classes. Switched to `std::unique_lock` and `std::shared_lock` instead.
* The name of the function – `get_unique_lock` – was preserved since we return an `std::unique_lock` now.

tests/unit/locking_test.cc
bhalevy: Not needed with the use of the std structures

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +237,44 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+/// to shared lock a \ref shared_mutex and construct an std::shared_lock object
+/// that will unlock the mutex when it's being destroyed.
+///
+/// The caller is responsible for keeping the corresponding \ref shared_mutex object
+/// alive at least until the returned lock is destroyed.
+inline future<std::shared_lock<shared_mutex>> get_shared_lock(shared_mutex& mutex) {
+    co_await mutex.lock_shared();
+
+    try {
+        co_return std::shared_lock<shared_mutex>{mutex, std::adopt_lock_t{}};
```
bhalevy, line 252: std::shared_lock is copyable, calling the murex.lock_shared, but in seastar case, this returns a future.  We should prevent this.

bhalevy (changes requested):

include/seastar/core/shared_mutex.hh
dawmd: It seems it's actually not. Fortunately the standard specifies that both the copy constructor and the copy assignment operator are both deleted ([link](https://eel.is/c++draft/thread.lock.shared#general))

dawmd:
Version 3:
* I forgot to coroutinize the tests I was adding. It's done now.
* Some of the new tests were redundant as they were testing the same scenarios others already did. They are removed in this version.
* I updated the commit message.

include/seastar/core/shared_mutex.hh
bhalevy: yes, sorry, I confused it with another ctor.

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +237,44 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
 
+/// \brief Construct a RAII-based shared lock corresponding to a given \ref shared_mutex object.
+///
+/// Since constructors cannot be exectued asynchronously, use this function
+/// to shared lock a \ref shared_mutex and construct an std::shared_lock object
+/// that will unlock the mutex when it's being destroyed.
+///
+/// The caller is responsible for keeping the corresponding \ref shared_mutex object
+/// alive at least until the returned lock is destroyed.
+inline future<std::shared_lock<shared_mutex>> get_shared_lock(shared_mutex& mutex) {
```
bhalevy, line 248: We should consider making the mutex type a template parameter so it could be extended/overridden, even for testing.

dawmd:
Version 4:
* CI failures fixed: when the macro `SEASTAR_ENABLE_ALLOC_FAILURE_INJECTION` was resolved to 0, some of our functions didn't return or await anything causing compilation errors.

dawmd:
Version 5:
* Turned the functions into function templates taking the mutex type as their parameter.

bhalevy (approved):

piodul:
@nyh, would you like to review? You said that you would like to do so.

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +237,46 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
 
+/// \brief Construct a RAII-based shared lock corresponding to a given Mutex object.
+///
+/// Since constructors cannot be exectued asynchronously, use this function
+/// to shared lock a Mutex and construct an std::shared_lock object
+/// that will unlock the mutex when it's being destroyed.
+///
+/// The caller is responsible for keeping the corresponding Mutex object
+/// alive at least until the returned lock is destroyed.
+template <typename Mutex>
```
nyh, line 248: nitpick: I'm not sure the name "Mutex" really represents the type of object this is. It needs to have exactly the right methods lock_shared(), unlock_shared(). Would have been nice to create a concept like https://en.cppreference.com/w/cpp/named_req/SharedLockable (just with a different return type for lock_shared()) and use that concept.
But it's a nitpick, not critical.

include/seastar/core/shared_mutex.hh
```diff
@@ -234,6 +237,46 @@ with_lock(shared_mutex& sm, Func&& func) noexcept {
...
+future<std::shared_lock<Mutex>> get_shared_lock(Mutex& mutex) {
+    co_await mutex.lock_shared();
+
+    try {
+        co_return std::shared_lock<Mutex>{mutex, std::adopt_lock_t{}};
+    } catch (...) {
```
nyh, line 254: Do you have an idea how the std::shared_lock{..., adopt_lock} could possibly throw? Maybe it's a bug in the standard that they forgot the "noexcept" for this constructor?

tests/unit/locking_test.cc
```diff
@@ -420,3 +422,170 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
         }
     }
 }
+
+SEASTAR_TEST_CASE(test_shared_mutex_locks) {
+    shared_mutex sm;
+
+    {
+        const auto ulock = co_await get_unique_lock(sm);
```
nyh, line 430: An idea: wouldn't it be more "OO-like" and easier to use if you also added a method, sm.get_unique_lock() (or maybe name it differently, to be more like the rest of the Seastar API)?
Of course it will be implemented using the get_unique_lock() template, but it will still be useful to have methods. Or so I think.

tests/unit/locking_test.cc
```diff
@@ -420,3 +422,170 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
...
+            coroutine::lambda([&sm, &counter, &max] (auto&&) -> future<> {
+        const auto slock = co_await get_shared_lock(sm);
+        ++counter;
+        max = std::max(max, counter);
+        co_await sleep(1ms);
+        --counter;
+    }));
+
+    BOOST_REQUIRE_EQUAL(counter, 0u);
+    BOOST_REQUIRE_NE(max, 0u);
```
nyh, line 482: I don't think this last check checks anything - obviously max can't be 0 because it will always be at least 1 (you increment counter, so it's at least 1, and then set max to at least that).

I'm not sure you really needed to check all this parallelization stuff, I think test_shared_mutex_locks above already checks the same thing more simply (it tries to take the same lock 3 times, and it succeeds).

tests/unit/locking_test.cc
```diff
@@ -420,3 +422,170 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
...
+
+    // If l.lock() fails, neither the function, nor l.unlock()
+    // should be called.
+    const auto ulock = co_await get_unique_lock(sm);
+
+    seastar::memory::local_failure_injector().fail_after(0);
+    BOOST_REQUIRE_THROW(co_await std::invoke([&] () -> future<> {
+        const auto slock = co_await get_shared_lock(sm);
+        BOOST_REQUIRE(false);
+    }), std::bad_alloc);
```
nyh, line 526: Can you please help me understand what we're trying to test here? That the coroutine functions need to allocate and if they can't, they fail before even starting? Isn't this obvious for all coroutine functions? Is this really what you hoped to test here?

tests/unit/locking_test.cc
```diff
@@ -420,3 +422,170 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
...
+    seastar::memory::local_failure_injector().fail_after(0);
+    BOOST_REQUIRE_THROW(co_await std::invoke([&] () -> future<> {
+        const auto ulock = co_await get_unique_lock(sm);
+        BOOST_REQUIRE(false);
+    }), std::bad_alloc);
+
+    seastar::memory::local_failure_injector().cancel();
+#endif // SEASTAR_ENABLE_ALLOC_FAILURE_INJECTION
+
+    // If the macro resolves to 0, we need this to prevent compilation errors
```
nyh, line 537: since you used ifdef and not if, I assume you mean if SEASTAR_ENABLE_ALLOC_FAILURE_INJECTION is not defined -  not if it "resolves to 0".

tests/unit/locking_test.cc
```diff
@@ -420,3 +422,170 @@ SEASTAR_THREAD_TEST_CASE(test_with_lock_typed_return_throwing_move_func) {
...
+    auto res = co_await std::invoke([&sm, &expected] () -> future<int> {
+        const auto slock = co_await get_shared_lock(sm);
+        co_return expected;
+    });
+    BOOST_REQUIRE_EQUAL(res, expected);
+
+    try {
+        const auto slock = co_await get_shared_lock(sm);
+        if (expected == 42) {
+            throw expected_exception(expected);
```
nyh, line 583: In test_shared_mutex_failed_func_locks above you already tested that an exception correctly unlocks the mutex. This seems to test the same thing, but not as well (you didn't check that sm got unlocked!). So why do we need this test? Why does this test's name have the strings "return_nothrow_move" - what does this refer to?

nyh (changes requested):
Looks good, and almost ready for merge, but I do have a few comments and questions I'd like you to please consider.

include/seastar/core/shared_mutex.hh
dawmd: It surprised me too and I don't see how it would happen either. I think it's still better to abide by the standard in case there are reasons for that. I highly doubt it's a mistake considering the type has been part of the language for 10 years

tests/unit/locking_test.cc
dawmd: I think that's a tricky case because on one hand I agree that it would be more convenient and natural to have it as a method, but on the other there are two arguments against it:
1. code duplication
2. inconsistency with Seastar (e.g. semaphores don't have an equivalent method of `seastar::get_units`) and STL (neither `std::mutex`, nor `std::shared_mutex` has a method returning a lock. Though it might be because of point 1)

include/seastar/core/shared_mutex.hh
nyh: The only guess I have is that because adopt_lock_t has undefined behavior if called on an _unlocked_ mutex, maybe they wanted to allow the possibility that this behavior might be throwing an exception...
In any case, what you did to work around this problem seems fine, so I'm resolving this discussion.

tests/unit/locking_test.cc
nyh: Ok. We can always add more methods later if we want.

tests/unit/locking_test.cc
nyh: Can you please answer this?
It's not a critical issue (who cares if a test checks a silly condition that can never be false), but it is a hint that maybe you missed the fact that the test isn't testing what you think it checks - so maybe you forgot to test what you really wanted to test, or something. So please just have a look.

nyh (changes requested):
The code looks good, and I resolved some of my old comments, but I still have some unanswered questions about the test, so can you please answer them? Thanks.

tests/unit/locking_test.cc
dawmd: I'm sorry for the late response. Most of the tests I added are analogues of the existing tests with using locks and coroutines being the only difference. I think you're right it's not necessary to check those things again. You're definitely right that comparing `max` against 0 is redundant. It would make more sense if Seastar were preemptive. I'll remove the test.

tests/unit/locking_test.cc
dawmd: As in https://github.com/scylladb/seastar/pull/2196#discussion_r1606063265, this test is an analogue of the existing one, adjusted to using locks and coroutinised. The way I understand the logic here is we want to check what happens in case of an allocation failure: we enter a coroutine and await a lock. However, the lock will never be returned because the allocation of the coroutine awaiting it will fail, throwing an `std::bad_alloc` (alternatively, the underlying allocation of a promise performed by `seastar::shared_mutex` will fail. Either way, there should be an allocation failure). However, now that I look at it, it seems overly complex and probably unnecessary; chances are I also don't understand it correctly. I can remove the test.

tests/unit/locking_test.cc
dawmd: Good catch. I forgot the rename the test. It was originally doing something else (and as most of the other tests I added, it was based on the existing ones). But you're right that it's unnecessary. I'll remove it.

include/seastar/core/shared_mutex.hh
dawmd: That's a good idea. I'll add the concepts

dawmd:
Version 2:
* rebased the PR on top of the current master branch (because of C-ares errors regarding deprecated functions)
* added concepts imposing constraints on the types accepted by `get_shared_lock()` and `get_unique_lock()`. Adjusted the comments accordingly
* Renamed the test `test_shared_mutex_failed_func_locks` to `test_shared_mutex_exception_locks` to better express what it verifies
* Removed the tests:
  * `test_shared_mutex_shared_locks`
  * `test_shared_mutex_failed_lock_locks`
  * `test_shared_mutex_failed_lock_mixed_locks`
  * `test_with_shared_typed_return_nothrow_move_func_locks`

nyh (approved):

PR #2203, opened 2024-04-25
treewide: inherit from formatter<string_view> not formatter<std::string_view>

tchaikov:
before {fmt} v10, it provides the specialization of `fmt::formatter<..>` for `std::string_view` as well as the specialization of `fmt::formatter<..>` for `fmt::string_view` which is an implementation builtin in {fmt} for pre-C++17 compatibility. but this type is used even if the code is compiled with C++ stadandard greater or equal to C++17. also, before {fmt} v10, `fmt::formatter<std::string_view>::format()` is defined so it accepts `std::string_view`. that's why we can define a `fmt::formatter<Type>` for a custom type by inheriting from `fmt::formatter<std::string_view>`, and call the parent class's `format()` method by passing a variable convertible to `std::string_view`.

after v10, `fmt::formatter<std::string_view>` still exists, but it is now defined using a macro named `FMT_FORMAT_AS()`, which delegates the `format()` call to `fmt::formatter<fmt::string_view, Char>`. so, `fmt::formatter<std::string_view>::format()` does not accept `std::string_view` anymore, instead, it now accepts `fmt::string_view`. because the `fmt::string_view` can be converted to `fmt::string_view` automatically, so this arrangement still works for `fmt::formatter<std::string_view>`. however, this does not for the custom types' formatter implemented with the above technique, the compiler fails to compile them like:

```
/usr/include/fmt/core.h:2593:45: error: implicit instantiation of undefined template 'fmt::detail::type_is_unformattable_for<seastar::basic_sstring<char, unsigned int, 15>, char>'
 2593 |     type_is_unformattable_for<T, char_type> _;
      |                                             ^
/usr/include/fmt/core.h:2656:23: note: in instantiation of function template specialization 'fmt::detail::parse_format_specs<seastar::basic_sstring<char, unsigned int, 15>, fmt::detail::compile_parse_context<char>>' requested here
 2656 |         parse_funcs_{&parse_format_specs<Args, parse_context_type>...} {}
      |                       ^                                                                                                                                                                                                               /usr/include/fmt/core.h:2787:47: note: in instantiation of member function 'fmt::detail::format_string_checker<char, seastar::basic_sstring<char, unsigned int, 15>, seastar::basic_sstring<char, unsigned int, 15>>::format_string_checker' r
equested here
 2787 |       detail::parse_format_string<true>(str_, checker(s));
      |                                               ^                                                                                                                                                                                       /home/kefu/dev/scylladb/master/mutation/mutation.cc:292:31: note: in instantiation of function template specialization 'fmt::basic_format_string<char, const seastar::basic_sstring<char, unsigned int, 15> &, const seastar::basic_sstring<ch
ar, unsigned int, 15> &>::basic_format_string<char[26], 0>' requested here
  292 |     out = fmt::format_to(out, "{{table: '{}.{}', key: {{", s.ks_name(), s.cf_name());
      |                               ^
/usr/include/fmt/core.h:1578:45: note: template is declared here
 1578 | template <typename T, typename Char> struct type_is_unformattable_for;
      |                                             ^
```

in this change, instead of inheriting from `fmt::formatter<std::string_view>`, we inherit from `fmt::formatter<string_view>`. this practice follows the document at  https://fmt.dev/latest/api.html#formatting-user-defined-types . and since there is less indirection under the hood -- we do not use the specialization created by FMT_FORMAT_AS which inherits from `formatter<fmt::string_view>`, hopefully this can improve the compilation speed a little bit. also, this change addresses the build failure with {fmt} v10.

### DIFF
--- a/include/seastar/core/abort_source.hh
+++ b/include/seastar/core/abort_source.hh
@@ -201,7 +201,7 @@ SEASTAR_MODULE_EXPORT_END
 #if FMT_VERSION < 100000
 // fmt v10 introduced formatter for std::exception
 template <>
-struct fmt::formatter<seastar::abort_requested_exception> : fmt::formatter<std::string_view> {
+struct fmt::formatter<seastar::abort_requested_exception> : fmt::formatter<string_view> {
     auto format(const seastar::abort_requested_exception& e, fmt::format_context& ctx) const {
         return fmt::format_to(ctx.out(), "{}", e.what());
     }
--- a/include/seastar/core/sstring.hh
+++ b/include/seastar/core/sstring.hh
@@ -882,12 +882,12 @@ std::ostream& operator<<(std::ostream& os, const std::unordered_map<Key, T, Hash
 SEASTAR_MODULE_EXPORT
 template <typename char_type, typename Size, Size max_size, bool NulTerminate>
 struct fmt::formatter<seastar::basic_sstring<char_type, Size, max_size, NulTerminate>>
-    : public fmt::formatter<std::basic_string_view<char_type>> {
-    using format_as_t = std::basic_string_view<char_type>;
+    : public fmt::formatter<basic_string_view<char_type>> {
+    using format_as_t = basic_string_view<char_type>;
     using base = fmt::formatter<format_as_t>;
     template <typename FormatContext>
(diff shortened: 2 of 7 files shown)


tchaikov:

PR #2204, opened 2024-04-25
iotune: allow user to set buffer size for random IO

pwrobelse:
In some cases the buffer size used by iotune was
incorrect. Because of that, the reported IOPS
for random write was lower than expected.

This change adds '--random-io-buffer-size <SIZE>'
parameter that enables users to force the buffer
size used in measurements of random read and
random write IOPS.

### DIFF
--- a/apps/iotune/iotune.cc
+++ b/apps/iotune/iotune.cc
@@ -417,6 +417,7 @@ private:
     fs::path _dirpath;
     uint64_t _file_size;
     file _file;
+    uint64_t _forced_random_io_buffer_size;
 
     std::unique_ptr<position_generator> get_position_generator(size_t buffer_size, pattern access_pattern) {
         if (access_pattern == pattern::sequential) {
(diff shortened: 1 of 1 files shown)


apps/iotune/iotune.cc
```diff
@@ -138,6 +141,10 @@ struct evaluation_directory {
                 _min_data_transfer_size = std::max(_min_data_transfer_size, disk_min_io_size);
                 _max_iodepth += read_first_line_as<uint64_t>(queue_dir / "nr_requests");
                 _disks_per_array++;
+
+                _physical_block_size = read_first_line_as<uint64_t>(queue_dir / "physical_block_size");
```
xemul, line 146: What if sysfs file is missing?

pwrobelse: Adjusted the PR according to the offline discussion. Instead of querying values I added a parameter that allows users to supply the desired buffer size for measurements of random read and random write IOPS.

apps/iotune/iotune.cc
```diff
@@ -549,6 +565,12 @@ class iotune_multi_shard_context {
         });
     }
 
+    future<> force_buffer_size(unsigned buffer_size) {
```
xemul, line 568: I think it makes sense to route the buffer size via iotune_multi_shard_context constructor, next to evaluation_directory argument

pwrobelse: Done

Issue #1698, opened 2023-06-13
iotune: Random IO buffer size is not always correct


xemul:
From https://github.com/scylladb/scylladb/issues/13477 : doing random writes test with auto-selected 512 bytes (or even 1k) makes i4i instance's drive ro read-modify-write thus dropping down the resulting IOPS rate. Need to make random writes block size larger ... somehow

avikivity:
There's the physical block size in /sys/block/nvme0n1/queue/physical_block_size.

Note: we still want to write the commitlog with logical block size, there's hope it avoids RMW since it's a stream.

mykaul:
Finally found what I was looking for. From https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/storage-optimized-instances.html#storage-instances-diskperf : 
```
This decrease in performance is even larger if the write operations are not in multiples of 4,096 bytes or not aligned to a 4,096-byte boundary. If you write a smaller amount of bytes or bytes that are not aligned, the SSD controller must read the surrounding data and store the result in a new location. This pattern results in significantly increased write amplification, increased latency, and dramatically reduced I/O performance.
```

mykaul:
Reviving this issue, in the hope it'll make it to Scylla 6.0.
Ping @xemul , @avikivity

pwrobelse:
Hello, please find some experiments with `iotune` related to buffer size of random IO. I used `i4i.4xlarge` instance type and `AMI` with `ScyllaDB 5.4.6`. The measurements are a follow-up to [issue#13477](https://github.com/scylladb/scylladb/issues/13477).

To prepare the machine the following steps were performed:
- I opened [ScyllaDB website](https://opensource.docs.scylladb.com/stable/reference/aws-images.html) with available AMIs of open source version and used version 5.4.6.
- An instance of `i4i.4xlarge` with `ScyllaDB 5.4.6 AMI` was launched via EC2 console.
- I ensured that setup of ScyllaDB had finished and then I performed the steps described in the [issue#13477](https://github.com/scylladb/scylladb/issues/13477) to reproduce low IOPS of random write - the logs from setup can be found below.

### Step 1: stop scylla-server and verify that it is not running.
```
scyllaadm@ip-172-31-46-190:~$ sudo systemctl stop scylla-server 
scyllaadm@ip-172-31-46-190:~$ ps aux | grep scylla  
[...]

PR #2205, opened 2024-04-25
iotune: Print file alignments too

xemul:
When performing measurements, iotune uses IO block size that is not-less than the DMA alignment reported by file. This in turn is what XFS reports (spiced with io-properties.yaml numbers, but given it's iotune, these numbers are not making any difference).

Without knowing those values any reasoning on the produced measurements is furtune-telling :(

refs: #1698

### DIFF
--- a/apps/iotune/iotune.cc
+++ b/apps/iotune/iotune.cc
@@ -439,6 +439,9 @@ public:
             options.extent_allocation_size_hint = _file_size;
             return open_file_dma(testfile.string(), open_flags::rw | open_flags::create, std::move(options)).then([this, testfile] (file file) {
                 _file = file;
+                if (this_shard_id() == 0) {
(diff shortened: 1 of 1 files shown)


pwrobelse (approved):

apps/iotune/iotune.cc
```diff
@@ -439,6 +439,9 @@ class test_file {
             options.extent_allocation_size_hint = _file_size;
             return open_file_dma(testfile.string(), open_flags::rw | open_flags::create, std::move(options)).then([this, testfile] (file file) {
                 _file = file;
+                if (this_shard_id() == 0) {
+                    iotune_logger.info("Filsystem parameters: read alignment {}, write alignment {}", _file.disk_read_dma_alignment(), _file.disk_write_dma_alignment());
```
pwrobelse, line 443: A typo: `Filsystem` -> `Filesystem`

xemul: Thanks, fixed

xemul:
@avikivity , please review

PR #2206, opened 2024-04-25
http: `to_reply()` from `redirect_exception`

WillemKauf:
This PR adds `redirect_exception::to_reply()` for construction of a `http::reply` from a redirect case, with a `Location` header and optionally a `Retry-After` header.

### DIFF
--- a/include/seastar/http/exception.hh
+++ b/include/seastar/http/exception.hh
@@ -24,6 +24,9 @@
 #include <seastar/util/log.hh>
 #include <seastar/http/reply.hh>
 #include <seastar/json/json_elements.hh>
+#include <initializer_list>
+#include <utility>
+#include <vector>
 
 namespace seastar {
 
@@ -59,14 +62,33 @@ private:
 };
 
 /**
- * Throwing this exception will result in a redirect to the given url
+ * Throwing this exception will result in a redirect to the given url.
+ * Optional extra headers can be included in the redirect response.
  */
 class redirect_exception : public base_exception {
 public:
-    redirect_exception(const std::string& url, http::reply::status_type status = http::reply::status_type::moved_permanently)
-            : base_exception("", status), url(url) {
+    using header_list = std::initializer_list<std::pair<std::string, std::string>>;
+
+    redirect_exception(const std::string& url,
+                       http::reply::status_type status = http::reply::status_type::moved_permanently,
+                       header_list extra_headers = {})
+            : base_exception("", status), url(url), _extra_headers(extra_headers) {
     }
+
+    /// Construct an http::reply for this redirect, with Location header and status set.
+    http::reply to_reply() const {
+        http::reply reply{};
+        reply.add_header("Location", url);
+        for (const auto& [name, value] : _extra_headers) {
+            reply.add_header(name, value);
+        }
+        reply.set_status(status());
+        return reply;
+    }
+
     std::string url;
+private:
+    std::vector<std::pair<std::string, std::string>> _extra_headers;
 };
 
 /**
--- a/src/http/routes.cc
+++ b/src/http/routes.cc
@@ -75,8 +75,7 @@ std::unique_ptr<http::reply> routes::exception_reply(std::exception_ptr eptr) {
         }
         std::rethrow_exception(eptr);
     } catch (const redirect_exception& _e) {
-       rep.reset(new http::reply());
-       rep->add_header("Location", _e.url).set_status(_e.status());
+      *rep = _e.to_reply();
     } catch (const base_exception& e) {
         rep->set_status(e.status(), internal::to_json(e));
     } catch (...) {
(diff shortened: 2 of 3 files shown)


xemul:
@nyh , please review

include/seastar/http/exception.hh
```diff
@@ -65,10 +65,21 @@ private:
  */
 class redirect_exception : public base_exception {
 public:
-    redirect_exception(const std::string& url, http::reply::status_type status = http::reply::status_type::moved_permanently)
-            : base_exception("", status), url(url) {
+    redirect_exception(const std::string& url, http::reply::status_type status = http::reply::status_type::moved_permanently, const std::optional<int>& retry_after = std::nullopt)
+            : base_exception("", status), url(url), retry_after(retry_after) {
```
nyh, line 69: You should mention in the commit message (and perhaps even split the commit into two) that one of the things that this patch does is to add to add Retry-After header (see https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3) support for redirections.
It is separate feature from the to_reply() function you are adding in this patch.

src/http/routes.cc
```diff
@@ -94,6 +93,9 @@ future<std::unique_ptr<http::reply> > routes::handle(const sstring& path, std::u
             handler->verify_mandatory_params(*req);
             auto r =  handler->handle(path, std::move(req), std::move(rep));
             return r.handle_exception(_general_handler);
+        } catch (const redirect_exception& _e) {
+            *rep = _e.to_reply();
+            rep->done("json");
```
nyh, line 98: This catch was missing previously. It looks like it fixes a bug - can you explain in the commit message what bug this was? Can you perhaps write a unit test for it?

nyh:
The commit message made it sound like your patch only added a new API function, but as I noted in the review above, it actually included two more separate changes:

1. It adds support for Retry-After (https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3), at least the part for 3xx (redirection), which was missing before. However, I want to ask something (because my command of the RFC is quite rusty) Is this the really only header that makes sense to put on a redirection reply? Shouldn't a redirection allow a more general set of headers to be given, and not specifically Retry-After (which isn't a particularly useful or commonly-used header, to be honest).
2. It adds a missing catch of redirect_exception. I am guessing there was a reason to adding this, and it's a bug fixed, but it needs to be explained and justified

travisdowns:
Rebased onto current master (723f319d7885) — commit is rebase-only, no functional changes. New HEAD: 1b8f69434.

travisdowns:
Apologies that this sat unaddressed for so long — I'm helping move it along.

Updated commit (17fcc7179) addresses some review feedback:

**General headers instead of Retry-After-specific** (@nyh's question about whether Retry-After is really the right choice): agreed, it is too narrow. The `optional<int> retry_after` parameter has been replaced with a general `header_list extra_headers` (`std::initializer_list<std::pair<std::string, std::string>>`), so callers can pass any headers they need, e.g.:
```cpp
throw redirect_exception("/new", status_type::moved_temporarily, {{"Retry-After", "120"}});
```

**Missing catch in `routes::handle()`** (@nyh's question about what bug this fixes): the PR's original catch block had a latent null-dereference. When a handler throws `redirect_exception` synchronously (rather than via `make_exception_future`), `rep` has already been moved into the handler by the time the catch fires, leaving the local `unique_ptr` null. The original code did `*rep = _e.to_reply()` which would dereference null. Fixed by constructing a fresh `unique_ptr` instead. A new integration test (`test_redirect_exception_sync_throw`) covers this path end-to-end; the existing `test_redirect_exception` only exercised the async path.

**Tests**: also added `test_redirect_exception_to_reply`, a simple unit test that verifies `to_reply()` sets the Location header, status, and any extra headers correctly without going through the full server stack.

WillemKauf:
> Apologies that this sat unaddressed for so long — I'm helping move it along.

Yes, thank you Travis! :) sorry I let this rot.

src/http/routes.cc
```diff
@@ -96,6 +95,10 @@ future<std::unique_ptr<http::reply> > routes::handle(const sstring& path, std::u
             handler->verify_mandatory_params(*req);
             auto r =  handler->handle(path, std::move(req), std::move(rep));
             return r.handle_exception(_general_handler);
+        } catch (const redirect_exception& _e) {
+            // rep may have been moved into handler->handle(), so create a fresh one
+            rep = std::make_unique<http::reply>(_e.to_reply());
+            rep->set_content_type("json");
```
nyh, line 101: I see this set_content_type("json") is preexisting, so it's ok that you did it too, but I wonder who/why decided that Seastar's error messages should be in JSON. I think it should have been the application's choice.

But in any case, I am guessing most existing applications already bypass this code, as it's really not general enough.

nyh (approved):

PR #2208, opened 2024-04-26
tls: add assignment operator for gnutls_datum

tchaikov:
when compiling the tree with GCC, it warns like:
```
[77/92] Building CXX object CMakeFiles/seastar.dir/src/net/tls.cc.o
/var/ssd/seastar/src/net/tls.cc:342:5: warning: definition of implicit copy assignment operator for 'gnutls_datum' is deprecated because it has a user-provided destructor [-Wdeprecated-copy-with-user-provided-dtor]
  342 |     ~gnutls_datum() {
      |     ^
/var/ssd/seastar/src/net/tls.cc:425:33: note: in implicit copy assignment operator for 'seastar::gnutls_datum' first required here
  425 |             _session_resume_key = {};
      |                                 ^
```

it rightly pointed out that we are using an implicitly defined copy assignment operator of `gnutls_datum`. since the lhs operator of the assignment operator could contain a resource which is supposed to be freed using `gnutls_free()` if it is reset by the instance by the default constructor, and we don't explicitly define one, so this would result in a leakage.

in this change

- define the move assignment operator, so that the copy constructor and assignment operators are not generated.
- explicit disable the copy constructor. this is not necessary, because the move assignment operator is already defined. but it makes it more obvious that this class is not copyable, and is only moveable.

Refs c224fe09f055a7c500a5fa22a8e115e8d1d64641

### DIFF
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -339,6 +339,18 @@ struct gnutls_datum : public gnutls_datum_t {
         data = nullptr;
         size = 0;
     }
+    gnutls_datum(const gnutls_datum&) = delete;
+    gnutls_datum& operator=(gnutls_datum&& other) {
+        if (this == &other) {
+            return *this;
+        }
+        if (data != nullptr) {
+            ::gnutls_free(data);
+        }
+        data = std::exchange(other.data, nullptr);
+        size = std::exchange(other.size, 0);
+        return *this;
+    }
     ~gnutls_datum() {
         if (data != nullptr) {
             ::gnutls_free(data);


tchaikov:
the build failure should be addressed by #2209

src/net/tls.cc
```diff
@@ -339,6 +339,18 @@ struct gnutls_datum : public gnutls_datum_t {
...
+        if (this == &other) {
+            return *this;
+        }
+        if (data != nullptr) {
+            ::gnutls_free(data);
+        }
+        data = std::exchange(other.data, nullptr);
+        size = std::exchange(other.size, 0);
+        return *this;
+    }
```
elcallio, line 353: The type is never moved. The warning is somewhat aggressive, since it only addresses a potential move. I suggest simply adding a delete for both move and copy constructors. Should never be needed.

elcallio (changes requested):

src/net/tls.cc
tchaikov: @elcallio hi Calle, thank you for your review. do you imply that at https://github.com/scylladb/seastar/blob/a62c9dbc9219b41ebeeab822b2389c56d7638648/src/net/tls.cc#L425 , `_session_resume_key` is always empty?

elcallio: No, sorry, missed/forgot that assignment. You are of course right.

tchaikov: @elcallio thank =) may i consider your reply an approval of this change?

elcallio: Yes, though I think perhaps you could get the same effect by just defining a move constructor, no? (Fewer lines).

elcallio (approved):

src/net/tls.cc
avikivity: A move constructor won't help, this is an assignment.

avikivity:
Looks good but all the tests fail.

tchaikov:
v2:

- rebased against master HEAD.

tchaikov:

avikivity:
@elcallio isn't this an example of slicing? Without this move assignment operator, after the right-hand-side is destroyed, it will call ::gnutls_free even though it no longer owns the data.

elcallio:
Yes and no. The "move" in question in the code was just a lazy "clear()". But a move is more general, so...

elcallio:
Yeah, sorry, I am a bit woosy today... The above move operator is really a move constructor in disguise... Any move from a non-empty source will leak. That is not good.

avikivity:
Moral: proper RAIIing is important even for local classes.

elcallio:
No, wait, now I can't read github diffs (hiding lines). There is a free there, so the move is 100% fine.

elcallio:
I will still send a "cleanup", following our "usual" move+init pattern instead, just to make sure none of us keep misreading this. Not urgent, but might help old, broken eyes like mine.

avikivity:
> I'll send a fix.

Comment was about pre-commit state.

PR #2211, opened 2024-04-26
CMakeLists.txt: Apply CXX deprecated flags conditionally

dawmd:
In ef941e5, we replaced the `-Wdeprecated-declarations` flag by `-Wdepracted` and` -Wno-error=deprecated`. Unfortunately, GCC and Clang interpret `-Wdeprecated` differently, compare:

* https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Warning-Options.html#index-Wno-deprecated
* https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Deprecated-Features.html

and

* https://clang.llvm.org/docs/DiagnosticsReference.html#wdeprecated

As a result, `-Wno-error=deprecated` doesn't prevent GCC from turning warnings related to using functions, variables, or types marked as deprecated into errors. Clang, however, does that.

In this commit, we conditionally enable the intended options depending on which compiler is used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -805,7 +805,10 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     list (APPEND Seastar_PRIVATE_CXX_FLAGS
       -Wno-error=stringop-overflow
       -Wno-error=array-bounds)
-  endif()
+  endif ()
+  list (APPEND Seastar_PRIVATE_CXX_FLAGS
+    -Wdeprecated-declarations
+    -Wno-error=deprecated-declarations)
 endif ()
 
 if (NOT BUILD_SHARED_LIBS)


tchaikov (approved):
thanks! lgtm.

CMakeLists.txt
```diff
@@ -795,17 +795,24 @@ set (Seastar_PRIVATE_CXX_FLAGS
...
+  endif ()
+  list (APPEND Seastar_PRIVATE_CXX_FLAGS
+    -Wdeprecated-declarations
+    -Wno-error=deprecated-declarations)
+endif ()
+
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+  list (APPEND Seastar_PRIVATE_CXX_FLAGS
+    -Wdeprecated
+    -Wno-error=deprecated)
```
avikivity, line 815: Shouldn't this be unconditional? We do want to warn about using deprecated features.

CMakeLists.txt
```diff
@@ -795,17 +795,24 @@ set (Seastar_PRIVATE_CXX_FLAGS
   -UNDEBUG
   -Wall
```
avikivity, line 796: > In https://github.com/scylladb/seastar/pull/1867, we replaced the `-Wdeprecated-declarations`

Please use commit hash references, not pull requests, to make the commit log less tied to github.

CMakeLists.txt
dawmd: I think it's undocumented because its behaviour is already described by `-Wno-deprecated` – if that option disables warnings about using deprecated features, `-Wdeprecated` makes the compiler warn about them. The manual says this:

> Many options have long names starting with `-f` or with `-W`---for example, `-fmove-loop-invariants`, `-Wformat` and so on.  Most of these have both positive and negative forms; the negative form of `-ffoo` is `-fno-foo`. This manual documents only one of these two forms, whichever one is not the default.

while [the documentation](https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Warning-Options.html) this:

> You can request many specific warnings with options beginning with ‘`-W`’, for example `-Wimplicit` to request warnings on implicit declarations. Each of these specific warning options also has a negative form beginning ‘`-Wno-`’ to turn off warnings; for example, `-Wno-implicit`. This manual lists only one of the two forms, whichever is not the default.

I also think enabling those options unconditionally would be a good idea, or at least `-Wno-error=deprecated` as `-Wdeprecated` is enabled by default.

dawmd:
Version 2:
* Enable `-Wdeprecated` and `-Wno-error=deprecated` unconditionally.
* Commit hash instead of the issue number in the commit message.

PR #2212, opened 2024-04-27
cmake: suppress warnings from ragel generated headers

tchaikov:
as ragel could generate headers which does not meet the criteria
enforced by the project, so the compiler may warn and fail the
build like:

```
/home/runner/work/scylladb/scylladb/build/gen/redis/protocol_parser.hh:110:1: error: unannotated fall-through between switch labels [clang-diagnostic-implicit-fallthrough]
  110 | case 1:
      | ^
/home/runner/work/scylladb/scylladb/build/gen/redis/protocol_parser.hh:110:1: note: insert 'FMT_FALLTHROUGH;' to silence this warning
  110 | case 1:
      | ^
      | FMT_FALLTHROUGH;
```

but the seastar application cannot control the behavior of ragel. if it is a CMake-based project, what it can do is to mark the headers as SYSTEM headers.

both Seastar and Seastar applications suffer from this. but Seastar can address by marking the exposed target with SYSTEM header. if the Seastar project chooses to include Seastar by including its CMakeLists.txt and use `seastar_generate_protobuf()` to generate its own headers using ragel, it can benefit from this as well.

so, in this series, we specify expose the generated headers as an INTERFACE target and specify `SYSTEM` for this exposed target, so that the compiler does not warn at seeing, for instance, the above switch labels.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -447,7 +447,8 @@ function (seastar_generate_ragel)
   target_sources(${args_TARGET}
     PRIVATE
       ${args_OUT_FILE})
-  target_include_directories(${args_TARGET} INTERFACE "${out_dir}")
+  target_include_directories(${args_TARGET} SYSTEM
+    INTERFACE "${out_dir}")
 
   set (${args_VAR} ${args_OUT_FILE} PARENT_SCOPE)
 endfunction ()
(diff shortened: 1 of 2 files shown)


tchaikov:
```
CMake Error: install(EXPORT "seastar-export" ...) includes target "seastar" which requires target "seastar_http_chunk_parsers" that is not in any export set.
CMake Error: install(EXPORT "seastar-export" ...) includes target "seastar" which requires target "seastar_http_request_parser" that is not in any export set.
CMake Error: install(EXPORT "seastar-export" ...) includes target "seastar" which requires target "seastar_http_response_parser" that is not in any export set.
CMake Error in CMakeLists.txt:
  export called with target "seastar" which requires target
  "seastar_http_chunk_parsers" that is not in any export set.


CMake Error in CMakeLists.txt:
  export called with target "seastar" which requires target
  "seastar_http_request_parser" that is not in any export set.


CMake Error in CMakeLists.txt:
  export called with target "seastar" which requires target
  "seastar_http_response_parser" that is not in any export set.
```

PR #2213, opened 2024-04-30
core/app_template: return status code 0 for --help

tchaikov:
if user queries the usage of an seastar application using `--help`, `--help-seastar` or `--help-loggers`, he/she should not be greeted with a non-zero status code, which implies an error in general. but in this case, there is no errors. so we should return 0 as the status code.


xemul:
:crossed_fingers: it won't break existing users, this `return 1` seems to exist since the very beginning of seastar

tchaikov:
🤞

PR #2214, opened 2024-04-30
tls: Restructure gnutls_datum move operator, adding construstor

elcallio:
Refs c224fe0
Refs #2208

Makes the type more RAII complete, and following usual pattern of destroy + construct in place -> old eyes can read easier.

### DIFF
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -340,15 +340,16 @@ struct gnutls_datum : public gnutls_datum_t {
         size = 0;
     }
     gnutls_datum(const gnutls_datum&) = delete;
+    gnutls_datum(gnutls_datum&& d) 
+    {
+        data = std::exchange(d.data, nullptr);
+        size = std::exchange(d.size, 0);
+    }
     gnutls_datum& operator=(gnutls_datum&& other) {
-        if (this == &other) {
-            return *this;
+        if (this != &other) {
+            this->~gnutls_datum();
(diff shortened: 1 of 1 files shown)


src/net/tls.cc
```diff
@@ -340,15 +340,16 @@ struct gnutls_datum : public gnutls_datum_t {
...
+    }
     gnutls_datum& operator=(gnutls_datum&& other) {
-        if (this == &other) {
-            return *this;
-        }
-        if (data != nullptr) {
-            ::gnutls_free(data);
+        if (this != &other) {
+            this->~gnutls_datum();
+            new (this) gnutls_datum(std::move(other));
```
tchaikov, line 351: yeah, i understand the motivation. but, personally, i am not in favor of calling destructor and using placement new unless really unnecessary.  they hurts the readability. and the new move-ctor is never used. i'd defer the review to seastar maintainers.

elcallio: Only motivation is that we tend to use this pattern elsewhere in seastar etc. Plus it is delegating operations. :-)

tchaikov: i see. i just don't really like this pattern, despite that it is repeating itself in seastar =( . and yeah, it reuses the dtor in a  very *obvious* way.. but still.. i just cannot convince myself that i like it.

avikivity: I'm at fault for introducing it but I now regret it, it's likely not totally safe wrt the standard.

PR #2215, opened 2024-05-01
perftune.py: fix systemd detection does not executed on some condition

syuu1228:
Currently, systemd detection is only executed when /etc/sysconfig/irqbalance or /etc/conf.d/irqbalance available. It causes error on Debian variants when systemd not available, such as container image.

To fix this, we should run systemd detection unconditionally.

Fixes scylladb/scylladb#17787


yaronkaikov (approved):

vladzcloudius (changes requested):
I don't understand the motivation for this patch nor how it fixes any problem.
You are not supposed to be needing systemctl in any context apart from irqbalance related tweaking.

On top of that, please, see my comment in the corresponding GH issue.

syuu1228:
I think this was incorrect way to fix the problem, closing.
Described more specifically at : https://github.com/scylladb/scylladb/issues/17787#issuecomment-2098046638

PR #2219, opened 2024-05-04
prometheus: support disabling aggregation at query time

nvartolomei:
In Seastar, metrics can be defined with implicit aggregation by specific labels, which occurs at query time. This feature is useful, for instance to define metrics per shard or even more finely grained per an application-defined entity while reporting them in a more aggregated manner, such as sum or histogram per server.

However, there are times when it is necessary to inspect the fine-grained metrics. This can be achieved by adding `__aggregate__=false` to the query string. For example: `http://localhost:9180/metrics?__aggregate__=false`

### DIFF
--- a/doc/prometheus.md
+++ b/doc/prometheus.md
@@ -54,6 +54,16 @@ To remove the help lines set `__help__=false`
 for example:
 `http://localhost:9180/metrics?__help__=false`
 
+## Aggregation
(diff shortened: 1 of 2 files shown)


nvartolomei:
cc @amnonh who added the original aggregation capabilities.

amnonh (approved):
I like the approach, LGTM

Issue #2216, opened 2024-05-02
Additional data when segfaulting


nvartolomei:
Currently a segfault is reported as:

```
Segmentation fault on shard 0.
Backtrace:
  ...
```

I believe we could add additional data about the context in which the segfault happens to aid debugging. I believe we should have access to the current scheduling group name, maybe tasktrace too. Anything else?

PR #2221, opened 2024-05-04
reactor: print scheduling group along with backtrace

nvartolomei:
reactor: print scheduling group along with backtrace

Backtraces are printed in at least 2 cases:
  1) during segfaults,
  2) during reactor stalls.
Extra context is always helpful in identifying the exact circumstances
during which the above happen. E.g. in a server application where user
requests are processed in one (or more) scheduling group(s), background
processes are split between different scheduling groups as well, knowing
the scheduling group narrows the search space.

Prior to this commit:

```
Segmentation fault on shard 0.
Backtrace:
  ...
```

With this commit:

```
Segmentation fault on shard 0, in scheduling group main.
Backtrace:
  ...
```

Ref https://github.com/scylladb/seastar/issues/2216

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -143,6 +143,7 @@ module seastar;
 #include <seastar/core/reactor.hh>
 #include <seastar/core/report_exception.hh>
 #include <seastar/core/resource.hh>
+#include <seastar/core/scheduling.hh>
 #include <seastar/core/scheduling_specific.hh>
 #include <seastar/core/sleep.hh>
 #include <seastar/core/smp_options.hh>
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -850,6 +851,9 @@ static void print_with_backtrace(backtrace_buffer& buf, bool oneline) noexcept {
         buf.append_decimal(this_shard_id());
     }
 
+    buf.append(", in scheduling group ");
```
xemul, line 854: It should happen inside `if (local_engine)` check above

nvartolomei: Updated.

PR #2227, opened 2024-05-08
github/workflow: remove the confusing comment

tchaikov:
based on the feedback of reviewer,

> the phrase "only build this combination when dpdk is enable" confusing
> - the phrase "in addition to the above" says the opposite, that we did
> build this combination without dpdk but now we'll build it again with
> dpdk.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -21,19 +21,47 @@ jobs:
         standard: [20, 23]
         mode: [dev, debug, release]
         include:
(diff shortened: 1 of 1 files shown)


nyh (approved):

nyh (approved):
Having to unroll these loops is ugly, but better ugly than wrong and confusing :-)

tchaikov:
there are still 12 builds. i need to find anther way to fix the matrix.

PR #2223, opened 2024-05-07
seastar-json2code: fix the path param handling

tchaikov:
this change addresses a regression introduced by
3f534c77, which failed to differentiate the path param from the non-path param, and always prefixed the param with "/".

Fixes 3f534c77d7b47398e582610e03ad08214fce32be
Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>

PR #2229, opened 2024-05-08
tests/unit: add a unit test for json2code

### LINKED ISSUE #2225: add a unit test for json2code
to add a unit test for the json2code based on `apps/httpd/main.cc`. in the test, we can reuse the same approach as the one we are using in `tests/unit/tls_test.cc`:

1. launch this httpd server, and
2. send some http requests to it, then
3. check their responses

tchaikov:
before this change, we don't have a test for
scripts/seastar-json2code.py. to verify the behavior of this script, let's add an integration test which exercise an http server which uses the source code generated with this script.

Fixes #2225


tchaikov:
closing in favor of #2223

PR #2232, opened 2024-05-09
prometheus: revert the condition for enabling aggregation

tchaikov:
in d7e16bbca0, we added the query time configuration for disabling aggregation, but the condition for disabling aggregation was wrong. before that change, we always enable the aggregation, but after that change, the aggregation is disabled when the value of query parameter of "__aggregate__" is not "false". in other words,

- if this query parameter is not set, the aggregation is disabled. this changes the existing behavior.
- if this query parameter's value is "false", the aggregation is enabeld. this is not the expected behavior.

so, in this change

- change the parameter from `disable_aggregation` to `enable_aggregation`. as the positive condition is always easier to understand.
- and update the handling accodingly.

Fixes d7e16bbca0d81e2d3b43886c932c242eeffb6e96

### DIFF
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -735,15 +735,15 @@ std::string get_value_as_string(std::stringstream& s, const mi::metric_value& va
     return value_str;
 }
 
-future<> write_text_representation(output_stream<char>& out, const config& ctx, const metric_family_range& m, bool show_help, bool disable_aggregation, std::function<bool(const mi::labels_type&)> filter) {
-    return seastar::async([&ctx, &out, &m, show_help, disable_aggregation, filter] () mutable {
+future<> write_text_representation(output_stream<char>& out, const config& ctx, const metric_family_range& m, bool show_help, bool enable_aggregation, std::function<bool(const mi::labels_type&)> filter) {
+    return seastar::async([&ctx, &out, &m, show_help, enable_aggregation, filter] () mutable {
         bool found = false;
         std::stringstream s;
         for (metric_family& metric_family : m) {
             auto name = ctx.prefix + "_" + metric_family.name();
             found = false;
             metric_aggregate_by_labels aggregated_values(metric_family.metadata().aggregate_labels);
-            bool should_aggregate = !disable_aggregation && !metric_family.metadata().aggregate_labels.empty();
+            bool should_aggregate = enable_aggregation && !metric_family.metadata().aggregate_labels.empty();
             metric_family.foreach_metric([&s, &out, &ctx, &found, &name, &metric_family, &aggregated_values, should_aggregate, show_help, &filter](const mi::metric_value& value, const mi::metric_info& value_info) mutable {
                 s.clear();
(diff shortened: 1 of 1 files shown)


tchaikov:
@scylladb/seastar-maint could you help review this change? it addresses a regression. and is required for https://github.com/scylladb/scylladb/issues/18554
and unblocking tablet testing based on https://github.com/scylladb/scylla-ccm/pull/574.

tchaikov:
cc @bhalevy

tchaikov:
with this change, scylladb master is able to pass `topology_experimental_raft/test_tablets`.

tchaikov:
cc @nvartolomei

nyh:
> in [d7e16bb](https://github.com/scylladb/seastar/commit/d7e16bbca0d81e2d3b43886c932c242eeffb6e96), we added the query time configuration for disabling aggregation, but the condition for disabling aggregation was wrong. before that change, we always enable the aggregation, but after that change, the aggregation is disabled when the value of query parameter of "**aggregate**" is not "false". in other words,
> 
>     * if this query parameter is not set, the aggregation is disabled. this changes the existing behavior.
> 
>     * if this query parameter's value is "false", the aggregation is enabeld. this is not the expected behavior.

These last two things don't look like "in other words" of the statement before them, but rather a failure to implement what the previous statement said.


> so, in this change
> 
>     * change the parameter from `disable_aggregation` to `enable_aggregation`. as the positive condition is always easier to understand.

You are overloading the word "parameter". The user-visible parameter remains with the name "aggregate", no change to that. You are just changing the name of some _variable_ in the code, to make it easier for you to understand.

I'll take a look at the actual change now :-)

src/core/prometheus.cc
```diff
@@ -896,16 +896,16 @@ class metrics_handler : public httpd::handler_base  {
         sstring metric_family_name = req->get_query_param("__name__");
         bool prefix = trim_asterisk(metric_family_name);
         bool show_help = req->get_query_param("__help__") != "false";
-        bool disable_aggregation = req->get_query_param("__aggregate__") != "false";
+        bool enable_aggregation = req->get_query_param("__aggregate__") != "false";
```
nyh, line 899: Please note that you could have fixed this bug by a one-line change - changing != to ==, and that's it. The renaming (and flipping) the variable wasn't necessary. But I agree it looks better.

nyh:
> These last two things don't look like "in other words" of the statement before them, but rather a failure to implement what the previous statement said.

I just now realized that the previous statement had a wrong "not" in the statement, so indeed the following statements are equivalent to that first wrong statement. This is a very confusing way to describe the bug. But I think I finally understood.

nyh (approved):

src/core/prometheus.cc
tchaikov: i knew. but i am not after a single-liner fix.

nvartolomei:
🤦‍♂️  sorry for the bad PR. Appreciate the fix!

PR #2256, opened 2024-05-21
sstring: deprecate formatters for vector and unordered_map

### LINKED ISSUE #1544: Deprecate public range print helpers
Seastar exposes two print helpers in https://github.com/scylladb/seastar/blob/58a4e90046da539adb969d26958095660d11bb5d/include/seastar/core/sstring.hh#L749-L782
and a bunch of `to_sstring` helpers in https://github.com/scylladb/seastar/blob/58a4e90046da539adb969d26958095660d11bb5d/include/seastar/core/sstring.hh#L720-L747

This is out of scope for the seastar library.

In particular, the implementation for `operator<<(std::ostream&, const std::unordered_map)` is unusual and conflicts with similar helpers in scylla utils/to_string.hh

cc @avikivity @nyh @tchaikov

tchaikov:
Seastar is an event-driven framework, not a collection of libraries for multiple purposes. also, when migrating to a {fmt} only formatting solution, the operator<< based printers can actually makes our lives more difficult. for instance, Boost.test expects operator<< and fall back to `boost_test_print_type()`, so even if we provide a templated `boost_test_print_type()` which accepts all types which can be formatted with {fmt}, if Boost.test is able to find the operator<<-based formatter for a std::vector, it just picks it, and then hits the brick wall, as the elements in the vector does not provide an operator<<-based formatter.

instead of enabling these operator<<:s to print the element with fmt::formatter support, let's just disable them on user's request.

in this change, a new option is added. so that user can disable these formatter on request. and these two formatters are marked deprecated. so in future, we can remove them when we bump up the API level or just completely drop them.

Fixes #1544

### DIFF
--- a/include/seastar/core/sstring.hh
+++ b/include/seastar/core/sstring.hh
@@ -836,10 +836,13 @@ string_type to_sstring(T value) {
 }
 }
 
+#ifdef SEASTAR_DEPRECATED_OSTREAM_FORMATTERS
+
 namespace std {
 
 SEASTAR_MODULE_EXPORT
 template <typename T>
+[[deprecated("Use {fmt} instead")]]
 inline
 std::ostream& operator<<(std::ostream& os, const std::vector<T>& v) {
     bool first = true;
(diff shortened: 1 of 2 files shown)


include/seastar/core/sstring.hh
```diff
@@ -858,6 +861,7 @@ std::ostream& operator<<(std::ostream& os, const std::vector<T>& v) {
 
 SEASTAR_MODULE_EXPORT
 template <typename Key, typename T, typename Hash, typename KeyEqual, typename Allocator>
+[[deprecated("Use {fmt} instead")]]
```
avikivity, line 864: (or write your own), but it's good advice.

avikivity (approved):

PR #2257, opened 2024-05-21
std-compat: remove #include:s which were added for pre C++17

tchaikov:
these `#include`s were introduced by 0bbcbbae and 79e846e9, for backward compatibility with pre-C++17 era. since we've dropped C++17 support, and now only support C++20 and up. let's remove these cruft.


avikivity (approved):

tchaikov:
v2:

- include the headers directly previously included by `std-compat.hh`

PR #2263, opened 2024-05-22
abort_source: add method to get exception pointer

Deexie:
Currently to check the exact type of an exception in abort_source, we need to rethrow it.

Add method to get abort_source exception_ptr.

### DIFF
--- a/include/seastar/core/abort_source.hh
+++ b/include/seastar/core/abort_source.hh
@@ -185,6 +185,11 @@ public:
         }
     }
 
+    /// Returns an exception with which an abort was requested.
+    const std::exception_ptr& abort_requested_exception_ptr() const noexcept {
+        return _ex;
+    }
+
     /// Returns the default exception type (\ref abort_requested_exception) for this abort source.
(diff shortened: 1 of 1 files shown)


include/seastar/core/abort_source.hh
```diff
@@ -185,6 +185,11 @@ public:
         }
     }
 
+    const std::exception_ptr& get_exception_ptr() const noexcept {
```
bhalevy, line 188: 1. the name is not entirely clear to the user IMO
2. Need to add doxygen comments to document the API

How about calling it `abort_requested_exception_ptr`?
This way it's clearer that it provides the same logical answer as `abort_requested`, just returning a different type, with additional information about the abort exception.

A hackish way might be to change `abort_requested` to return the exception_ptr instead of converting it to a `bool`, but let's be conservative and add a method instead.

bhalevy (changes requested):
See inline comments

Deexie:
- rename method to `abort_requested_exception_ptr`
- add comment

PR #2267, opened 2024-05-24
github: s/clang++-18/clang++/

tchaikov:
in 28b05551, we switched from ubuntu:jammy to fedora:40 as the building environment of our CI workflow. and we replaced the input parameter "clang++-18" in test.yml with "clang++", because in fedora:40, the clang++ executable does not have the postfix as the clang-18 installed by the setup-cpp action.

but we failed to update all of them in that change, that's why the "check header" step was not performed after that change.

in this change, we replace that last "clang++-18" in the workflow to "clang++".

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -86,7 +86,7 @@ jobs:
         run: cmake --build build/${{inputs.mode}}
 
       - name: Check Header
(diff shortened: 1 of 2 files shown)


tchaikov:
this change depends on #2268

PR #2269, opened 2024-05-25
Unfriend reactor from timer

xemul:
Reactor messes with private timer fields, but it doesn't really need to. All the code that does so naturally belongs to timer_set class, not reactor itself.


xemul:
@avikivity , please review

xemul:
@avikivity , please review

xemul:
@avikivity , please review

PR #2274, opened 2024-05-28
Fix compilation failure on Ubuntu 22.04

pwrobelse:
Fix compilation failure on Ubuntu 22.04
    
When the code was built according to the
README.md on Ubuntu 22.04 it failed due
to missing includes of `<optional>` and
`<variant>`.
    
Commit b822c94b2bb5860003995ff3fa197709dfc46c2c
seemed to be the first for which the issue
occurred.
    
This change adds missing includes and removes
usage of "std-compat.hh" in modified files when
possible.

### DIFF
--- a/include/seastar/core/queue.hh
+++ b/include/seastar/core/queue.hh
@@ -23,9 +23,9 @@
 
 #include <seastar/core/circular_buffer.hh>
 #include <seastar/core/future.hh>
-#include <seastar/util/std-compat.hh>
 #include <seastar/util/modules.hh>
 #ifndef SEASTAR_MODULE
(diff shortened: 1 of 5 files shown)


tchaikov (approved):
lgtm

tchaikov:
@xemul hi Pavel, could you please take a look? i would like to prepare a PR to bump up the seastar submodule so that we can be avoid the potential build failures where some of the source files fail to include `filesystem`, `variant` or `optional` after updating the seastar submodule. for instance, the one fixed by https://github.com/scylladb/scylladb/pull/18921 .

PR #2275, opened 2024-05-28
Unfriend reactor from scheduling_group_key

xemul:
The scheduling_group_key is assumed to be only copy-able and move-able by its users. However, reactor wants to mess with its ID for its maintenance purposes, but there's dedicated internal:: helper for that. Also rector wants to construct a temporary key object out of id, but that's just the way how internal reactor helper works, it can go with plain integral id value.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -409,7 +409,7 @@ private:
     void insert_activating_task_queues();
     void account_runtime(task_queue& tq, sched_clock::duration runtime);
     void account_idle(sched_clock::duration idletime);
-    void allocate_scheduling_group_specific_data(scheduling_group sg, scheduling_group_key key);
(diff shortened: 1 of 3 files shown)


include/seastar/core/reactor.hh
```diff
@@ -409,7 +409,7 @@ private:
     void insert_activating_task_queues();
     void account_runtime(task_queue& tq, sched_clock::duration runtime);
     void account_idle(sched_clock::duration idletime);
-    void allocate_scheduling_group_specific_data(scheduling_group sg, scheduling_group_key key);
+    void allocate_scheduling_group_specific_data(scheduling_group sg, unsigned long key_id);
     future<> rename_scheduling_group_specific_data(scheduling_group sg);
```
avikivity, line 413: This is a small reduction in readability, but okay.

PR #2277, opened 2024-05-30
net/tls: Added additional error codes

michael-redpanda:
In order to not have to reference actual GnuTLS error codes in the user application, expose some additional error codes via seastar::tls::ERROR_* variables.


tchaikov:
@elcallio hi Calle, could you take a look?

elcallio (approved):

nyh (approved):

PR #2282, opened 2024-06-05
docker: introduce devcontainers

pavelbezpravel:
lint dockerfile:
- reduce image size
- reduce layers amount
- prevent caching issues related to apt-get commands usage

introduce [devcontainers](https://containers.dev/)

sort packages


xemul:
The title of this PR doesn't match its content. It's in fact two PRs -- one that introduces devcontainers, and another one that cleans existing scripts and dockerfile. Please, split

.devcontainer/Dockerfile
```diff
@@ -2,9 +2,8 @@
 
 FROM ubuntu:mantic
 
-COPY install-dependencies.sh /tmp/
-
-RUN apt-get update && apt-get install -y \
+RUN --mount=type=bind,source=./install-dependencies.sh,target=./install-dependencies.sh \
```
xemul, line 5: The patch comment lacks motivation of this change

docker/dev/Dockerfile
```diff
@@ -1,23 +1,37 @@
...
-    && update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-18 18
-COPY install-dependencies.sh /tmp/
-RUN bash /tmp/install-dependencies.sh
-CMD /bin/bash
+    && update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-18 18 \
+    && bash /tmp/install-dependencies.sh \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+CMD ["/bin/bash"]
```
xemul: nit: This change is missing in (already laconic enough) patch comment

.devcontainer/Dockerfile
mykaul: I thought it reduces layers - since we do not copy the file over to the container.

pavelbezpravel: > I thought it reduces layers - since we do not copy the file over to the container.

yep, it's right

docker/dev/Dockerfile
pavelbezpravel: I found it with [hadolint](https://hadolint.github.io/hadolint/). it helps to pass OS signals to executable in a correct way (https://github.com/hadolint/hadolint/wiki/DL3025#rationale)

I've commented about it in a new one [PR](https://github.com/scylladb/seastar/pull/2283)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
avikivity: And why it's good for us.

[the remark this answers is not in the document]
pavelbezpravel: I've added instructions and hopefully this patch will help developers simplify their development process

PR #2284, opened 2024-06-06
github: add a workflow for building scylla with seastar

### LINKED ISSUE #2239: add an integration test with scylla
we could schedule a cronjob which runs scylla unit tests with the latest scylla master HEAD every (couple) days and the latest seastar master HEAD , so we can have better a test coverage of seastar.

tchaikov:
in order to identify the fail-to-build-from-source failures early. the next step is to run unit tests.

Fixes #2239

### DIFF
--- a/.github/workflows/scylla.yaml
+++ b/.github/workflows/scylla.yaml
@@ -1,0 +1,56 @@
+name: Build Scylla
+
+on:
+  schedule:
+    # 5AM everyday
+    - cron: '0 5 * * *'
+
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  BUILD_TYPE: RelWithDebInfo
+  BUILD_DIR: build
+
+jobs:
+  build-scylla:
+    runs-on: ubuntu-latest
+    container: fedora:40
+    strategy:
+      matrix:
+        build_type:
+          - Debug
+          - RelWithDebInfo
+          - Dev
+    steps:
+      - run: |
+          sudo dnf -y install git
+      - uses: actions/checkout@v4
+        with:
+          repository: scylladb/scylladb
+          submodules: true
+      - name: Update seastar submodule
+        run: |
+          cd seastar
+          git config --global --add safe.directory '*'
+          git fetch origin master
+          git reset --hard origin/master
+          git submodule sync
+          git submodule update --init --force --recursive
+          cd ..
+      - run: |
+          sudo ./install-dependencies.sh
+      - name: Generate the building system
+        run: |
+          cmake                                         \
+            -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+            -DCMAKE_C_COMPILER=clang                    \
+            -DCMAKE_CXX_COMPILER=clang++                \
+            -G Ninja                                    \
+            -B $BUILD_DIR                               \
+            -S .
+      - run: |
+          cmake --build $BUILD_DIR --target scylla


tchaikov:
cc @bhalevy

tchaikov:
see https://github.com/tchaikov/seastar/actions/runs/9392998585/job/25868266020 for the workflow in action.

.github/workflows/scylla.yaml
```diff
@@ -0,0 +1,46 @@
...
+        run: |
+          cmake                                         \
+            -DCMAKE_BUILD_TYPE=$BUILD_TYPE              \
+            -DCMAKE_C_COMPILER=clang                    \
+            -DCMAKE_CXX_COMPILER=clang++                \
+            -G Ninja                                    \
+            -B $BUILD_DIR                               \
+            -S .
+      - run: |
+          cmake --build $BUILD_DIR --target scylla
```
bhalevy, line 46: other tagets depends on seastar too, like unit tests

.github/workflows/scylla.yaml
```diff
@@ -0,0 +1,46 @@
+name: Build Scylla
+
+on:
+  schedule:
+    # only at 5AM Saturday
```
bhalevy, line 5: why not daily?

.github/workflows/scylla.yaml
bhalevy: also, better cover more build modes.
release, dev, and debug to begin with

.github/workflows/scylla.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - run: |
+          sudo dnf -y install git
+      - uses: actions/checkout@v4
+        with:
+          repository: scylladb/scylladb
+          submodules: true
+      - name: Update seastar submodule
+        run: |
+          git config --global --add safe.directory '*'
+          git fetch origin master
```
bhalevy, line 32: don't we need to change directory into `seastar`?

.github/workflows/scylla.yaml
```diff
@@ -0,0 +1,46 @@
...
+          git reset --hard origin/master
```
bhalevy, line 33: how about sync and update of seastar's submodules (e.g. dpdk), recuresively?
they might have changed in seastar/master

tchaikov:
v2:

- schedule the job on daily basis
- `cd seastar` !
- update submodules of seastar
- cover more modes
- build all tests (including scylla executable)

tchaikov:
running at https://github.com/tchaikov/seastar/actions/runs/9395901191

.github/workflows/scylla.yaml
tchaikov: indeed!

.github/workflows/scylla.yaml
tchaikov: ack.

.github/workflows/scylla.yaml
tchaikov: yeah, good point.

.github/workflows/scylla.yaml
tchaikov: sure. added.

tchaikov:
@bhalevy the runner was out of space. probably we can start with "scylla" instead of all unit tests?

avikivity:
It's not reasonable, scylla is downstream from seastar and can introduce failures by misusing seastar (for example, using deprecated symbols which are then removed).

.github/workflows/scylla.yaml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          repository: scylladb/scylladb
+          submodules: true
+      - name: Update seastar submodule
+        run: |
+          cd seastar
+          git config --global --add safe.directory '*'
+          git fetch origin master
+          git reset --hard origin/master
+          git submodule update --init --force --recursive
```
bhalevy, line 41: I think you also need `git submodule sync --recursive` before update in case new submodules were added or removed (it happens very rarely, but still...)

.github/workflows/scylla.yaml
bhalevy: > sure. added.

I don't see other targets in [`cbc81b9` (#2284)](https://github.com/scylladb/seastar/pull/2284/commits/cbc81b922317daa7b1160f3266e127e18f617c2b)

tchaikov:
> It's not reasonable, scylla is downstream from seastar and can introduce failures by misusing seastar (for example, using deprecated symbols which are then removed).

@avikivity probably we don't need to use this workflow as a gating test, but just as a reference? like, if the workflow fails, we might want to take a look and at least to be prepared for it?

bhalevy:
> It's not reasonable, scylla is downstream from seastar and can introduce failures by misusing seastar (for example, using deprecated symbols which are then removed).

Would having a similar pipeline in scylla make more sense?
I don't like the fact it's reactive (i.e. it will detect issues after stuff was already merged into seastar/master), but better detect those integration issue as early as possible, and not only when we attempt to update the seastar submodule in scylla.

.github/workflows/scylla.yaml
tchaikov: fixed.

tchaikov:
v4:

- run `git submodule sync --recursive`

tchaikov:
closing in favor of https://github.com/scylladb/scylladb/pull/19135

tchaikov:
> As a reference it makes sense. But it should be part of ScyllaDB's CI.
> 
> (I also want to build and test ScyllaDB with llvm.git, this is similar)

i will add it later on.

tchaikov:
> As a reference it makes sense. But it should be part of ScyllaDB's CI.
> 
> (I also want to build and test ScyllaDB with llvm.git, this is similar)

created at https://github.com/scylladb/scylladb/pull/19164

PR #2287, opened 2024-06-07
Minor improvements to the json subsystem

StephanDollberg:
Some improvements to the json subsystem:

 - Fix a lifetime issue in stream_range_as_array
 - Make formatter::write(const jsonable&) recursively call `write` and not `to_json`. This allows for recursive streamable encoding and hence avoids some potentially very large string allocations
 - Adds some tests for the above

Note that because whitespace bahaviour is inconsistent across the different ways of encoding things (`write`, `to_json`, vector encoding, map encoding, ...) some of the tests need adapting as there is now less whitespace.

### DIFF
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -735,7 +735,7 @@ std::string get_value(int size) {
 /*
  * A helper object that map to a big json string
  * in the format of:
- * {"valu": "aaa....aa", "valu": "aaa....aa", "valu": "aaa....aa"...}
+ * {"valu":"aaa....aa","valu":"aaa....aa","valu":"aaa....aa"...}
  *
  * The object can have an arbitrary size in multiplication of 10000 bytes
  *  */
@@ -743,10 +743,10 @@ struct extra_big_object : public json::json_base {
     json::json_element<sstring>* value;
     extra_big_object(size_t size) {
         value = new json::json_element<sstring>;
-        // size = brackets + (name + ": " + get_value) * n + ", " * (n-1)
-        // size = 2 + (name + 6 + get_value) * n - 2
+        // size = brackets + ("name" + ":" + get_value()) * n + "," * (n-1)
+        // size = 2 + (valu + 4 + get_value) * n - 1
         value->_name = "valu";
-        *value = get_value(9990);
+        *value = get_value(9992);
         for (size_t i = 0; i < size/10000; i++) {
             _elements.emplace_back(value);
         }
@@ -769,7 +769,9 @@ struct extra_big_object : public json::json_base {
 SEASTAR_TEST_CASE(json_stream) {
     std::vector<extra_big_object> vec;
     size_t num_objects = 1000;
-    size_t total_size = num_objects * 1000001 + 1;
+    // each object is 1000001 bytes plus a comma for all but the last and plus
+    // two for the [] backets
+    size_t total_size = num_objects * 1000002 + 1;
     for (size_t i = 0; i < num_objects; i++) {
         vec.emplace_back(1000000);
     }
--- a/include/seastar/json/formatter.hh
+++ b/include/seastar/json/formatter.hh
@@ -330,9 +330,12 @@ public:
      * @param obj the date_time to format
      * @return the given json object in a json format
      */
-    static future<> write(output_stream<char>& s, const jsonable& obj) {
-      return s.write(to_json(obj));
-     }
+    template <std::derived_from<jsonable> Jsonable>
+    static future<> write(output_stream<char>& s, Jsonable obj) {
+        return do_with(std::move(obj), [&s] (const auto& obj) {
+            return obj.write(s);
+        });
+    }
 
     /**
      * return a json formatted unsigned long
--- a/include/seastar/json/formatter.hh
+++ b/include/seastar/json/formatter.hh
@@ -302,18 +302,24 @@ public:
      * @return the given vector in a json format
      */
     template<typename... Args>
-    static future<> write(output_stream<char>& s, const std::vector<Args...>& vec) {
-        return write(s, state::array, vec.begin(), vec.end());
+    static future<> write(output_stream<char>& s, std::vector<Args...> vec) {
+        return do_with(std::move(vec), [&s] (const auto& vec) {
+            return write(s, state::array, vec.begin(), vec.end());
+        });
     }
 
     template<typename... Args>
-    static future<> write(output_stream<char>& s, const std::map<Args...>& map) {
-        return write(s, state::map, map.begin(), map.end());
+    static future<> write(output_stream<char>& s, std::map<Args...> map) {
+        return do_with(std::move(map), [&s] (const auto& map) {
+            return write(s, state::map, map.begin(), map.end());
+        });
     }
 
     template<typename... Args>
-    static future<> write(output_stream<char>& s, const std::unordered_map<Args...>& map) {
-        return write(s, state::map, map.begin(), map.end());
+    static future<> write(output_stream<char>& s, std::unordered_map<Args...> map) {
+        return do_with(std::move(map), [&s] (const auto& map) {
+            return write(s, state::map, map.begin(), map.end());
+        });
     }
 
     /**
--- a/tests/unit/json_formatter_test.cc
+++ b/tests/unit/json_formatter_test.cc
@@ -24,8 +24,10 @@
 #include <seastar/core/do_with.hh>
 #include <seastar/testing/test_case.hh>
 #include <seastar/core/sstring.hh>
-#include <seastar/core/do_with.hh>
+#include <seastar/core/vector-data-sink.hh>
 #include <seastar/json/formatter.hh>
+#include <seastar/json/json_elements.hh>
+#include <seastar/testing/thread_test_case.hh>
 
 using namespace seastar;
 using namespace json;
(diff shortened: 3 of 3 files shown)


include/seastar/json/json_elements.hh
```diff
@@ -338,7 +338,9 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
                     auto fut = first ? make_ready_future<>() : s.write(", ");
                     first = false;
                     return fut.then([&s, &f, &v]() {
-                        return formatter::write(s, f(v));
+                        return do_with(f(v), [&s] (const auto& val) {
+                            return formatter::write(s, val);
```
xemul, line 342: The formatter::write() may get the retult as value or push it into `to_json()` first, so it's not necessarily that val needs a stable storage. Where exactly do you observe use after free?

StephanDollberg: I guess right now it's an implementation detail that it isn't an issue in practice because:
 
 - `jsonable` (or other non value types) are passed a const reference to `formatter::write`
 - Then they get passed to `to_json` which creates another temporary string
 - That temporary string will then be passed to the async `output_stream::write`. This internally copies the string so it works right now. If the implementation didn't do this copy things would already break.

However with my change to use `jsonable::write` in `formatter::write` there is now more async functions in between that don't directly copy their arguments so the issue gets exposed.

tests/unit/httpd_test.cc
```diff
@@ -769,7 +769,9 @@ struct extra_big_object : public json::json_base {
 SEASTAR_TEST_CASE(json_stream) {
     std::vector<extra_big_object> vec;
     size_t num_objects = 1000;
-    size_t total_size = num_objects * 1000001 + 1;
+    // each object is 1000001 bytes plus a comma for all but the last and plus
+    // two for the [] backets
+    size_t total_size = num_objects * 1000002 + 1;
```
xemul, line 774: So, serializing got changed by this patch?

include/seastar/json/json_elements.hh
xemul: > Also the issue already exists if you return a `std::vector/unordered/map` from the mapping functions as those don't do `to_json` directly.



> Also the issue already exists if you return a `std::vector/unordered/map` from the mapping functions as those don't do `to_json` directly.



> Also the issue already exists if you return a `std::vector/unordered/map` from the mapping functions as those don't do `to_json` directly.

Doesn't the `do_with(... Container ...), [] { <this code>` that wraps the whole loop prevents collection elements from being use-after-free-d?

xemul: > That temporary string will then be passed to the async output_stream::write. This internally copies the string so it works right now. If the implementation didn't do this copy things would already break.

I wouldn't call it "implementation detail that might change", bur rather the part of the iostream API

tests/unit/httpd_test.cc
StephanDollberg: Yes, the serialization is quite inconsistent in regards to spaces after `:` and `,`. `to_json`, `formatter::write`, `vector/map/pair` serialization all handle this differently and hence the change above is required.

include/seastar/json/json_elements.hh
StephanDollberg: > Doesn't the do_with(... Container ...), [] { <this code> that wraps the whole loop prevents collection elements from being use-after-free-d?

No because the mapper function doesn't necessarily return references to elements in the container but possibly newly created objects/temporaries.

xemul: > > Doesn't the do_with(... Container ...), [] {  that wraps the whole loop prevents collection elements from being use-after-free-d?
> 
> No because the mapper function doesn't necessarily return references to elements in the container but possibly newly created objects/temporaries.

So what does the mapper function do?

> ... return a std::vector/unordered/map from the mapping functions ...

or

> ... the mapper function doesn't necessarily return references to elements in the container but possibly newly created objects/temporaries ...

?

tests/unit/httpd_test.cc
xemul: Ah, so it's just spaces

src/json/formatter.cc
```diff
@@ -181,6 +181,10 @@ sstring formatter::to_json(unsigned long l) {
     return to_string(l);
 }
 
+future<> formatter::write(output_stream<char>& s, const jsonable& obj) {
+    return obj.write(s);
```
xemul, line 185: That's some awesome change, but doesn't it create liveness issues described [here](https://github.com/scylladb/seastar/pull/2287#discussion_r1633145921)? So from now on the caller of formatter::weite(s, obj) needs to maintain the lifetime of `obj`, while prior to it it didn't have to? You fixed it in one place, but can there be others?

include/seastar/json/json_elements.hh
StephanDollberg: Sorry I wasn't clear above and possibly confusing so let me give some examples:

```
stream_range_as_array(some_std_map_int_to_int, [] (const auto& kv) {
    return kv.first;
});
```

Completely fine as we just a return an integer so there is no lifetime issues.

```
stream_range_as_array(some_std_map_int_to_int, [] (const auto& kv) {
    some_jsonable_derived_type ret;

    ret.val_a = kv.first;
    ret.val_b = kv.second;

    return ret;
});
```

Also fine because of what we discussed above, we return a temporary on which we call `to_json` and then call `output_stream::write` with a string which doesn't really involve any async functions.

```
stream_range_as_array(some_std_map_int_to_int, [] (const auto& kv) {
    std::vector<int> ret;
    ret.push_back(kv.first);
    ret.push_back(kv.second);
    return ret;
});
```

Not fine, we call `write(output_stream<char>& s, const std::vector<Args...>&)` which in turn calls `write(output_stream<char>& stream, state s, Iter i, Iter e)`. This function is truly async and we might actually suspend and at which point the temporary will have already gone out of scope/stack.

In all cases the lifetime of the outer container (`some_std_map_int_to_int`) needs to be guaranteed to survive the whole thing by the caller but the actual issue stems from the internal object returned by the mapper function.

Does this make sense?

src/json/formatter.cc
StephanDollberg: > but doesn't it create liveness issues described https://github.com/scylladb/seastar/pull/2287#discussion_r1633145921

It goes back to the other comment thread. It's not clearly stated anywhere but in theory any user of `formatter::write` already had to guarantee this before because the same issue already exists on some of the other overloads like `write(output_stream<char>& s, const std::vector<Args...>&)`. 

My default assumption with async functions is that I have to make sure of lifetime issues unless the argument is by value. But not sure that's the correct default assumption.

include/seastar/json/json_elements.hh
xemul: It does, thank you. However, in the last example you map int:int _element_ into a vector (not map<int, int> into a vector), which is valid, but strange. Why would anyone want to do it? The mapper function is supposed to extract a single element out of a map and vector<> is not a "single element" thing. My main concern is that the stream_range_as_array() routine is equipped with extra allocation, that would hurt all the "non-weird" use-cases that don't need it to support a "weird" one that does.

xemul: Once I wrote it I thought that maybe you meant this

```
stream_range_as_array(std::map<int, std::vector<int>>, [] (const auto& kv) {
    return some_other_vector_made_out_of(kv.second);
});
```

?

StephanDollberg: > Once I wrote it I thought that maybe you meant this

Yes I guess that would be a better example.

> My main concern is that the stream_range_as_array() routine is equipped with extra allocation, that would hurt all the "non-weird" use-cases that don't need it to support a "weird" one that does.

That's a fair point. I am wondering whether instead of making the change to `stream_range_as_array` we'd rather move it into `formatter::write` where it's definitely needed like so:

``` diff
diff --git a/include/seastar/json/formatter.hh b/include/seastar/json/formatter.hh
index b9b6b3da..a7cb9405 100644
--- a/include/seastar/json/formatter.hh
+++ b/include/seastar/json/formatter.hh
@@ -302,18 +302,24 @@ class formatter {
      * @return the given vector in a json format
      */
     template<typename... Args>
-    static future<> write(output_stream<char>& s, const std::vector<Args...>& vec) {
-        return write(s, state::array, vec.begin(), vec.end());
+    static future<> write(output_stream<char>& s, std::vector<Args...> vec) {
+        return do_with(std::move(vec), [&s] (const auto& vec) {
+            return write(s, state::array, vec.begin(), vec.end());
+        });
     }

     template<typename... Args>
-    static future<> write(output_stream<char>& s, const std::map<Args...>& map) {
-        return write(s, state::map, map.begin(), map.end());
+    static future<> write(output_stream<char>& s, std::map<Args...> map) {
+        return do_with(std::move(map), [&s] (const auto& map) {
+            return write(s, state::map, map.begin(), map.end());
+        });
     }

     template<typename... Args>
-    static future<> write(output_stream<char>& s, const std::unordered_map<Args...>& map) {
-        return write(s, state::map, map.begin(), map.end());
+    static future<> write(output_stream<char>& s, std::unordered_map<Args...> map) {
+        return do_with(std::move(map), [&s] (const auto& map) {
+            return write(s, state::map, map.begin(), map.end());
+        });
     }

     /**
@@ -330,7 +336,12 @@ class formatter {
      * @param obj the date_time to format
      * @return the given json object in a json format
      */
-    static future<> write(output_stream<char>& s, const jsonable& obj);
+    template <std::derived_from<jsonable> Jsonable>
+    static future<> write(output_stream<char>& s, Jsonable obj) {
+        return do_with(std::move(obj), [&s] (const auto& obj) {
+            return obj.write(s);
+        });
+    }

     /**
      * return a json formatted unsigned long
```

I guess that would also address your other comment.

xemul: Well, almost. I was rather thinking about overloading formatter::write()-s with rvalues, like

```patch
--- a/include/seastar/json/formatter.hh
+++ b/include/seastar/json/formatter.hh
@@ -306,6 +306,13 @@ class formatter {
         return write(s, state::array, vec.begin(), vec.end());
     }
 
+    template<typename... Args>
+    static future<> write(output_stream<char>& s, std::vector<Args...>&& vec) {
+        return do_with(std::move(vec), [] (const auto& vec) {
+            return write(s, vec);
+        });
+    }
+
     template<typename... Args>
     static future<> write(output_stream<char>& s, const std::map<Args...>& map) {
         return write(s, state::map, map.begin(), map.end());
```

so that only temporary vectors are do_with()-ed, while those that already have stable storage -- don't. The latter can be done e.g. by

```c++
stream_range_as_array(std::map<int, std::vector<int>>, [] (const auto& kv) {
    return kv.second;
});
```

i.e. -- directory returning a reference to an element that lives in container's memory

avikivity:
@amnonh please review

include/seastar/json/json_elements.hh
xemul: > Just a side note as it doesn't really matter to your overall point, your example would actually still create a temporary/hit the rref overload as the deduced return type is not a reference in the lambda. That would need an explicit return type.

True :(

xemul: OK, so it looks like the problem is around formatter::write(). It silently assumes that the argument is in stable storage, but doesn't really uses this fact.

I'd call for API level raise, but it doesn't worth the effort. And it's questionable which is worse -- move object to stable storage or allocate a string for its json representation. Let's switch to [this](https://github.com/scylladb/seastar/pull/2287#discussion_r1634774243) suggestion. It will be slower, but it's not performance critical path (for us). It should address [the other liveness issue](https://github.com/scylladb/seastar/pull/2287#discussion_r1633456825) as well, should it?

StephanDollberg: > Let's switch to https://github.com/scylladb/seastar/pull/2287#discussion_r1634774243 suggestion. It will be slower, but it's not performance critical path (for us). It should address https://github.com/scylladb/seastar/pull/2287#discussion_r1633456825 as well, should it?

Ok I will make the change. Yes it will address that issue as well.

xemul (approved):

amnonh (approved):
LGTM

Issue #1430, opened 2023-01-20
Cross-shard preemption is needed


xemul:
As seen in, e.g. scylladb/scylla#12562, high-prio class with low-concurrency workload is in trouble.

- all shards have all low-prio queues full, disk is fully dispatched, shards line-up on shared token bucket
- high-prio request pops up on some shard
- scheduler decides that the new class should preempt everything, but the best it can do is preempt everything _on_ _this_ _shard_ _only_

As a result, the high-prio request happens in the middle of the global token queue and have chances to get dispatched only after all the preceding other shards complete their IO

Can be reproduced with io_tester:
`io_tester --conf jobs.yaml --storage /dev/null  --duration 5`
and jobs.yaml being
```
- name: big_writes
  shards: all
  type: seqwrite
  data_size: 1GB
  shard_info:
    rps: 8000
    parallelism: 2
    reqsize: 128kB
    shares: 80

- name: latency_reads
  shards: all
  type: randread
  data_size: 1GB
  shard_info:
    rps: 250
    parallelism: 1
    reqsize: 512
    shares: 1000
  options:
    pause_distribution: poisson
```

avikivity:
I think reducing the group sizes will help, and also tightening the scheduler's idea of what is an idle class.

xemul:
> I think reducing the group sizes will help, 

Agree

> and also tightening the scheduler's idea of what is an idle class.

Maybe tightening the preemption criteria? Because "idle class" notion is as simple as "zero requests in it".

...

While I was writing the above question I got an idea. We could inject "empty" requests into the global token queue for classes that are idling for too little (i.e. -- had a request dispatched recently and didn't yet have a new one queued). So that when there pops up a request for real, it could "preempt" that empty one thus getting closer to the dispatcheable future.

avikivity:
> > I think reducing the group sizes will help, 
> 
> Agree
> 
> > and also tightening the scheduler's idea of what is an idle class.
> 
> Maybe tightening the preemption criteria? Because "idle class" notion is as simple as "zero requests in it".
> 


Zero requests for some time? 


> ...
> 
> While I was writing the above question I got an idea. We could inject "empty" requests into the global token queue for classes that are idling for too little (i.e. -- had a request dispatched recently and didn't yet have a new one queued). So that when there pops up a request for real, it could "preempt" that empty one thus getting closer to the dispatcheable future.

Aha. A sort of reservation.

Issue #1017, opened 2022-02-17
smp::submit_to performance problems


xemul:
As shown in 0c2228a4916dc2dc66e9d587343ce436b92b7bef, on a busy target shard the `smp::submit_to` can take ~50ms to complete. Partially this is because a task that arrives from a smp message queue is put at the tail of the scheduler queue which is full of busyloops. However, in the example from the mentioned commit message the maximum time to traverse the queue should be `nr_tasks * task_latency = 100 * 50us =  5ms`.

avikivity:
Can't it take an arbitrary amount of time? If the task group in question is busy, it will be delayed.

xemul:
Well, the test in question uses the single task group -- the default one -- and it _is_ constantly busy. But my point is that since the arrived smp job is queued at the tail of the task group's queue, reactor should get to it once it scans this list at least once, i.e. in 5ms.

xemul:
Well, the test in question uses the single task group -- the default one -- and it _is_ constantly busy. But my point is that since the arrived smp job is queued at the tail of the task group's queue, reactor should get to it once it scans this list at least once, i.e. in 5ms.

avikivity:
That's true for non-submit_to tasks as well, no? We have FIFO order usually.

xemul:
Yes, but 5ms to process the task list should result in 1s/5ms=200 messages/s, but the result is 20 messages/s

avikivity:
What's the relationship to this issue?

PR #2289, opened 2024-06-07
[RFC] Scatter IO request among shards evenly

xemul:
Current IO queues design assumes that IO workload is very uniform on different shards in a sense that -- all classes are more-or-less equally loaded on different shards. In reality that's not true and some shards can easily get more requests in one of its class than the others.

This leads to a very nasty consequence. Consider a corner case -- two classes, A and B, with shares 100 and 1000 respectively, two shards. Class A is active on shard-0 only, while class B is active on shard-1 only. We expect, that they share disk ~~bandwidth~~ capacity in 1:10 proportion, but in reality it's going to be 1:1, because cross-shard queue doesn't preempt (it doesn't, because load is expected to be even on all shards).

The solution below is far from being ideal, but it demonstrates (and proves) one of the approaches -- when requests get queued into a class, scatter them evenly across all shards on the node in plain round robin manner. This brings the aforementioned assumption back true, and the described corner case start working as expected.

One of the immediate problem with it -- intents (IO cancellation) stops working
Another problem -- smp::submit_to() can slow things down #1017 . This has chance to be solved by introducing more sophisticated logic around scattering, that would e.g. take into account imbalance of class queues on different shards.

fixes: #1430 
fixes: #1083

Tested with io_tester on 2-shards setup

```
- name: shard_0
  shard_info:
    parallelism: 32
    reqsize: 4kB
    shares: 100
  shards:
  - '0'
  type: randread
- name: shard_1
  shard_info:
    parallelism: 32
    reqsize: 4kB
    shares: 800
  shards:
  - '1'
  type: randread
```

Without the patch the result is

```
---
- shard: 0
  shard_0:
    throughput: 354812.719  # kB/s
    IOPS: 88703.375
    latencies:  # usec
      average: 359
      p0.5: 358
      p0.95: 387
      p0.99: 408
      p0.999: 455
      max: 2704
    stats:
      total_requests: 887065
      io_queue_total_exec_sec: 50.380561289002046
      io_queue_total_delay_sec: 268.07659966000335
      io_queue_total_operations: 887097
      io_queue_starvation_time_sec: 8.5391000000000023e-05
      io_queue_consumption: 4.9869152042269711
      io_queue_adjusted_consumption: 0.051574149906635287
- shard: 1
  shard_1:
    throughput: 356593.031  # kB/s
    IOPS: 89148.4531
    latencies:  # usec
      average: 358
      p0.5: 357
      p0.95: 385
      p0.99: 406
      p0.999: 444
      max: 2768
    stats:
      total_requests: 891511
      io_queue_total_exec_sec: 50.404540082000388
      io_queue_total_delay_sec: 268.12554203199409
      io_queue_total_operations: 891543
      io_queue_starvation_time_sec: 5.6404000000000004e-05
      io_queue_consumption: 5.0119088914990426
      io_queue_adjusted_consumption: 0.0079740712642669669
...
```

With it the result is

```
---
- shard: 0
  shard_0:
    throughput: 79122.1875  # kB/s
    IOPS: 19780.5469
    latencies:  # usec
      average: 1616
      p0.5: 671
      p0.95: 3134
      p0.99: 3164
      p0.999: 3214
      max: 11612
    stats:
      total_requests: 197817
      io_queue_total_exec_sec: 5.8829173029999913
      io_queue_total_delay_sec: 16.473880899999962
      io_queue_total_operations: 98925
      io_queue_starvation_time_sec: 5.0897572340000377
      io_queue_consumption: 0.55611797422170639
      io_queue_adjusted_consumption: 0.0072732244133949277
- shard: 1
  shard_1:
    throughput: 632085.812  # kB/s
    IOPS: 158021.547
    latencies:  # usec
      average: 201
      p0.5: 232
      p0.95: 334
      p0.99: 359
      p0.999: 400
      max: 8257
    stats:
      total_requests: 1580247
      io_queue_total_exec_sec: 46.343612493998329
      io_queue_total_delay_sec: 199.64613740399747
      io_queue_total_operations: 790139
      io_queue_starvation_time_sec: 0.0068174949999999963
      io_queue_consumption: 4.4418549409508703
      io_queue_adjusted_consumption: 0.0072718314528465271
...
```

### DIFF
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -105,9 +105,12 @@ private:
     friend struct ::io_queue_for_tests;
     friend const fair_group& internal::get_fair_group(const io_queue& ioq, unsigned stream);
 
+    unsigned get_request_target(const internal::io_request&) const noexcept;
+
     priority_class_data& find_or_create_class(internal::priority_class pc);
     future<size_t> queue_request(internal::priority_class pc, internal::io_direction_and_length dnl, internal::io_request req, io_intent* intent, iovec_keeper iovs) noexcept;
     future<size_t> queue_one_request(internal::priority_class pc, internal::io_direction_and_length dnl, internal::io_request req, io_intent* intent, iovec_keeper iovs) noexcept;
+    future<size_t> queue_one_request_locally(internal::priority_class pc, internal::io_direction_and_length dnl, internal::io_request req, io_intent* intent, iovec_keeper iovs) noexcept;
 
     // The fields below are going away, they are just here so we can implement deprecated
     // functions that used to be provided by the fair_queue and are going away (from both
@@ -157,6 +160,7 @@ public:
         unsigned flow_ratio_ticks = 100;
         double flow_ratio_ema_factor = 0.95;
         double flow_ratio_backpressure_threshold = 1.1;
+        unsigned request_fanout_block_bits = 24; // 16MB
     };
 
     io_queue(io_group_ptr group, internal::io_sink& sink);
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -232,6 +232,7 @@ private:
     std::vector<std::unique_ptr<priority_class_data>> _priority_classes;
     util::spinlock _lock;
     const shard_id _allocated_on;
+    std::vector<io_queue*> _io_queues;
 
     static fair_group::config make_fair_group_config(const io_queue::config& qcfg) noexcept;
     priority_class_data& find_or_create_class(internal::priority_class pc);
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -569,6 +569,7 @@ io_queue::io_queue(io_group_ptr group, internal::io_sink& sink)
     , _sink(sink)
     , _flow_ratio_update([this] { update_flow_ratio(); })
 {
+    _group->_io_queues[this_shard_id()] = this;
     auto& cfg = get_config();
     if (cfg.duplex) {
         static_assert(internal::io_direction_and_length::write_idx == 0);
@@ -611,6 +612,7 @@ std::chrono::duration<double> io_group::io_latency_goal() const noexcept {
 io_group::io_group(io_queue::config io_cfg, unsigned nr_queues)
     : _config(std::move(io_cfg))
     , _allocated_on(this_shard_id())
+    , _io_queues(smp::count, nullptr)
 {
     auto fg_cfg = make_fair_group_config(_config);
     _fgs.emplace_back(fg_cfg, nr_queues);
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -941,7 +941,7 @@ io_queue::request_limits io_queue::get_request_limits() const noexcept {
     return l;
 }
 
-future<size_t> io_queue::queue_one_request(internal::priority_class pc, io_direction_and_length dnl, internal::io_request req, io_intent* intent, iovec_keeper iovs) noexcept {
+future<size_t> io_queue::queue_one_request_locally(internal::priority_class pc, io_direction_and_length dnl, internal::io_request req, io_intent* intent, iovec_keeper iovs) noexcept {
     return futurize_invoke([pc = std::move(pc), dnl = std::move(dnl), req = std::move(req), this, intent, iovs = std::move(iovs)] () mutable {
         // First time will hit here, and then we create the class. It is important
         // that we create the shared pointer in the same shard it will be used at later.
@@ -959,6 +959,41 @@ future<size_t> io_queue::queue_one_request(internal::priority_class pc, io_direc
         pclass.on_queue();
         _queued_requests++;
         return fut;
+    });
+}
+
+unsigned io_queue::get_request_target(const internal::io_request& req) const noexcept {
+    unsigned pos = 0;
+
+    switch (req.opcode()) {
+    case internal::io_request::operation::read:
+        pos = req.as<internal::io_request::operation::read>().pos;
+        break;
+    case internal::io_request::operation::write:
+        pos = req.as<internal::io_request::operation::write>().pos;
+        break;
+    case internal::io_request::operation::readv:
+        pos = req.as<internal::io_request::operation::readv>().pos;
+        break;
+    case internal::io_request::operation::writev:
+        pos = req.as<internal::io_request::operation::writev>().pos;
+        break;
+    default:
+        // cannot happen, but let's stay on the safe side
+        break;
+    }
+
+    return (this_shard_id() + (pos >> get_config().request_fanout_block_bits)) % smp::count;
+}
+
+future<size_t> io_queue::queue_one_request(internal::priority_class pc, io_direction_and_length dnl, internal::io_request req, io_intent* intent, iovec_keeper iovs) noexcept {
+    unsigned target = get_request_target(req);
+    if (target == this_shard_id()) {
+        return queue_one_request_locally(std::move(pc), std::move(dnl), std::move(req), intent, std::move(iovs));
+    }
+
+    return smp::submit_to(target, [pc = std::move(pc), dnl = std::move(dnl), req = std::move(req), iovs = std::move(iovs), g = _group] () mutable {
+        return g->_io_queues[this_shard_id()]->queue_one_request_locally(std::move(pc), std::move(dnl), std::move(req), nullptr, std::move(iovs));
     });
 }
 


vladzcloudius:
I like this idea.
We can merge this algo as optional and have a configurable to enable it.

avikivity:
This is likely to add a lot of latency, as well as cause bouncing in the kernel - our files are shard-local and now concurrent requests to the same file will be dispatched from different CPUs.

avikivity:
Maybe we should encode the priority-weighted consumption into the shared structure used to allocate bandwidth to different shards. Not that I know how.

xemul:
> This is likely to add a lot of latency, as well as cause bouncing in the kernel - our files are shard-local and now concurrent requests to the same file will be dispatched from different CPUs.

Yes, that's the goal -- on file is IO-ed by different cores, but what bouncing do you refer to?

Maybe we can try scattering from the different angle -- spread the IO based on its position in the file, so that close IO happens on the same core (pushed the updated version, it helps the io-tester just as good)

xemul:
> Maybe we should encode the priority-weighted consumption into the shared structure used to allocate bandwidth to different shards. Not that I know how.

That's the biggest problem. And it's two-fold.

On one hand, this shared queue size is limited with the amount of capacity a disk can chew in one io-tick. So it cannot grow infinitely. On the other hand, even if we try to over-subscribe this queue, i.e. -- put more data into it than on tick allows, this won't make any sense without cross-shard preemption.

travisdowns:
> but what bouncing do you refer to?

Not Avi, but at least all the fs-side metadata for the file, some of which is mutated (at least for writes), like unwritten extent conversion and mtime updates, so that would suffer from cross-core bouncing. 

I wonder what this looks like in a high throughput scenario with lots of writes?

xemul:
Presumably the `append` workflow is of much interest. Here's the config

```
- data_size: 5000MB
  name: shard_0
  shard_info:
    parallelism: 4
    reqsize: 128kB
    shares: 100
  shards:
  - '0'
  type: append
- data_size: 5000MB
  name: shard_1
  shard_info:
    parallelism: 4
    reqsize: 128kB
    shares: 200
  shards:
  - '1'
  type: append
```

master (without this patch)

```
---
- shard: 0
  shard_0:
    throughput: 657452.75  # kB/s
    IOPS: 5136.34961
    latencies:  # usec
      average: 777
      p0.5: 723
      p0.95: 1226
      p0.99: 1987
      p0.999: 6468
      max: 32319
    stats:
      total_requests: 154092
      io_queue_total_exec_sec: 13.292341552999675
      io_queue_total_delay_sec: 48.041072028000031
      io_queue_total_operations: 77040
      io_queue_starvation_time_sec: 8.7738936659999442
      io_queue_consumption: 6.9746361637115477
      io_queue_adjusted_consumption: 0.0714553416967392
- shard: 1
  shard_1:
    throughput: 723878.375  # kB/s
    IOPS: 5655.2998
    latencies:  # usec
      average: 706
      p0.5: 697
      p0.95: 895
      p0.99: 2628
      p0.999: 4402
      max: 32494
    stats:
      total_requests: 169662
      io_queue_total_exec_sec: 8.6915708470000403
      io_queue_total_delay_sec: 49.563282467999166
      io_queue_total_operations: 84833
      io_queue_starvation_time_sec: 8.7950991439998774
      io_queue_consumption: 7.6801571868658067
      io_queue_adjusted_consumption: 0.071407152295112616
...
```

this patch

```
---
- shard: 0
  shard_0:
    throughput: 645738.25  # kB/s
    IOPS: 5044.83008
    latencies:  # usec
      average: 791
      p0.5: 723
      p0.95: 1216
      p0.99: 2304
      p0.999: 9165
      max: 116479
    stats:
      total_requests: 151348
      io_queue_total_exec_sec: 15.342357325000025
      io_queue_total_delay_sec: 46.433791281999888
      io_queue_total_operations: 75704
      io_queue_starvation_time_sec: 8.4504402329998971
      io_queue_consumption: 6.8536845293045046
      io_queue_adjusted_consumption: 0.068532964706420896
- shard: 1
  shard_1:
    throughput: 707869.688  # kB/s
    IOPS: 5530.23193
    latencies:  # usec
      average: 722
      p0.5: 714
      p0.95: 1020
      p0.99: 2702
      p0.999: 8967
      max: 123073
    stats:
      total_requests: 165910
      io_queue_total_exec_sec: 10.552503203999859
      io_queue_total_delay_sec: 48.367674382000011
      io_queue_total_operations: 82970
      io_queue_starvation_time_sec: 8.6282275269999218
      io_queue_consumption: 7.5114948403835298
      io_queue_adjusted_consumption: 0.068475198030471807
...
```

the map of extents is the same in both cases

```
/home/xfs/test-shard_0-0:
	0: [0..16776959]: 181456..16958415
	1: [16776960..30719999]: 16958416..30901455
	2: [30720000..44568319]: 33555864..47404183
/home/xfs/test-shard_1-1:
	0: [0..16776959]: 52428848..69205807
	1: [16776960..33553919]: 69205808..85982767
	2: [33553920..48554239]: 85982768..100983087
```

xemul:
Actually, write workload is not the main goal of this PR, so if we worry about what happens when writing a file, this fanout can be turned off for writes. Like

```patch
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -970,13 +970,11 @@ unsigned io_queue::get_request_target(const internal::io_request& req) const noe
         pos = req.as<internal::io_request::operation::read>().pos;
         break;
     case internal::io_request::operation::write:
-        pos = req.as<internal::io_request::operation::write>().pos;
         break;
     case internal::io_request::operation::readv:
         pos = req.as<internal::io_request::operation::readv>().pos;
         break;
     case internal::io_request::operation::writev:
-        pos = req.as<internal::io_request::operation::writev>().pos;
         break;
     default:
         // cannot happen, but let's stay on the safe side
```

xemul:
Seqwrite io-tester workload of two parallel produced the same result

xemul:
It's also good to make this fanout conditional -- if all shards off-load their traffic to each other, then there's no sense in that, obviously. But I don't yet have any good ideas how to implement it :(

vladzcloudius:
> Actually, write workload is not the main goal of this PR, so if we worry about what happens when writing a file, this fanout can be turned off for writes. Like
> 
> ```diff
> --- a/src/core/io_queue.cc
> +++ b/src/core/io_queue.cc
> @@ -970,13 +970,11 @@ unsigned io_queue::get_request_target(const internal::io_request& req) const noe
>          pos = req.as<internal::io_request::operation::read>().pos;
>          break;
>      case internal::io_request::operation::write:
> -        pos = req.as<internal::io_request::operation::write>().pos;
>          break;
>      case internal::io_request::operation::readv:
>          pos = req.as<internal::io_request::operation::readv>().pos;
>          break;
>      case internal::io_request::operation::writev:
> -        pos = req.as<internal::io_request::operation::writev>().pos;
>          break;
>      default:
>          // cannot happen, but let's stay on the safe side
> ```

IMO the motivation of this PR relates to writes the same way as to reads.
So it does to mixed workload too.

BTW, @xemul, your test results above don't show the expected ratio (1:2). Too low `parallelizm` maybe?

>  if all shards off-load their traffic to each other, then there's no sense in that, obviously.

If by doing so you achieve the correct shares-based arbitration all the time (!) then it makes all the sense to do so.

xemul:
> > Actually, write workload is not the main goal of this PR, so if we worry about what happens when writing a file, this fanout can be turned off for writes. 
> 
> IMO the motivation of this PR relates to writes the same way as to reads. So it does to mixed workload too.
> 
> BTW, @xemul, your test results above don't show the expected ratio (1:2). Too low `parallelizm` maybe?

Yes, the queue was not saturated most of the time

> > if all shards off-load their traffic to each other, then there's no sense in that, obviously.
> 
> If by doing so you achieve the correct shares-based arbitration all the time (!) then it makes all the sense to do so.

Yes, that's the goal

avikivity:
> > This is likely to add a lot of latency, as well as cause bouncing in the kernel - our files are shard-local and now concurrent requests to the same file will be dispatched from different CPUs.
> 
> Yes, that's the goal -- on file is IO-ed by different cores, but what bouncing do you refer to?

Cache-line and lock bouncing in the kernel. Two writes to the same file sent to random vcpus. Each will lock it. Also we will get reordering. While we tolerate reordering to some extent, large-scale reordering will confuse XFS and the disk. They really want sequential writes.

> 
> Maybe we can try scattering from the different angle -- spread the IO based on its position in the file, so that close IO happens on the same core (pushed the updated version, it helps the io-tester just as good)

The best would be not to scatter, I still don't see why we can't do another solution. Better to take the synchronization hit at the higher level.

avikivity:
> Actually, write workload is not the main goal of this PR, so if we worry about what happens when writing a file, this fanout can be turned off for writes. Like


Write suffers more, yes, but read also suffers. All the file metadata has to be loaded, the inode locked (shared, but lock-shared is a write). Sequential reads get reordered, possibly confusing the disk's read-ahead stuff.

avikivity:
> > Maybe we should encode the priority-weighted consumption into the shared structure used to allocate bandwidth to different shards. Not that I know how.
> 
> That's the biggest problem. And it's two-fold.
> 
> On one hand, this shared queue size is limited with the amount of capacity a disk can chew in one io-tick. So it cannot grow infinitely. On the other hand, even if we try to over-subscribe this queue, i.e. -- put more data into it than on tick allows, this won't make any sense without cross-shard preemption.

We should reserve instead of preempt. Each shard+scheduling_group should look back at its demand for the past 10 io_latency_goal ticks and predict (using a conservative function like max) what its expected demand will be. When its time comes it takes from its reservation, returns what it didn't consume, and updates its reservation.

It means we'll sometimes under-consume, but it's better to under-consume than to have high latency.

xemul:
> We should reserve instead of preempt. ... When its time comes it takes from its reservation, returns what it didn't consume, and updates its reservation.

It's token bucket, where tokens mean "time" not "availability tickets" or something.

Let's take the above scenario -- shard-0 with 1000-shares class, shard-1 with 100-shares class, infinite queues on both (for saturation). To get 10:1 proportion shard-0 should queue 10x more requests "on average". Provided the queue is infinite, the token bucket is always drained so shards get their tokens one-by-one -- the claim for extra tokens, bucket says -- later, so shards "reserve" one slot in the shared queue for that "later".

Now shard-0 should "reserve" more tokens. It means that it should occupy 10 slots in the shared queue. This has several limitations

1. (major) The shared queue is limited with the amount of tokens that can be processed by disk in io-latency-goal duration. There can be no room for 10 slots at all (in #1522 I oversubscribed for some small fixed amount of slots due to this)
2. (moderate) If another 1000-shares class pops up on shard-1 it will _not_ be able to queue itself in front of _any_ shard-0's entry in the shared queue, so it will need to wait for all reserved "later"-s. If it's reservation for a single shard, then, probably, not that much, if it's reservations for 30 shards -- much worse _or_ each shard cannot reserve for more than tick/30 (as described in previous item)
3. (minor) If shard-0 makes a mistake and reserves 10 slots, but really only needs 6, there's no way to return 4 unused, we learned it the hard way in #1766

xemul:
> All the file metadata has to be loaded, 

Is it per-cpu in the kernel???

> the inode locked (shared, but lock-shared is a write).

Yeah, that's unfortunate :( Is it inode lock (which is really a mutex) or something else?

> Sequential reads get reordered, possibly confusing the disk's read-ahead stuff.

The /var/lib/scylla typically sit on a raid0 with 1MB chunk size, so disk-read-ahead already needs to do something with the fact that every 1Mb sequential access is interrupted.

In the last version of this PR I changed the fanout approach -- instead of round-robin each and every request, I split the file in 16Mb chunks and each chunk is always served by a single shard, the mapping is `(this_shard_id() + pos / 16Mb) % smp::count`. Does this help?

avikivity:
> > All the file metadata has to be loaded,
> 
> Is it per-cpu in the kernel???


It's per-file. Our files are per-cpu so our file's metadata is also per-cpu. But if we spread the files around, they're not per-cpu any more.

> 
> > the inode locked (shared, but lock-shared is a write).
> 
> Yeah, that's unfortunate :( Is it inode lock (which is really a mutex) or something else?

I meant the inode mutex, but for sure there are more. We start with incrementing the fd reference count, that's already bad (with io_uring registered files you can avoid it, but still).

> 
> > Sequential reads get reordered, possibly confusing the disk's read-ahead stuff.
> 
> The /var/lib/scylla typically sit on a raid0 with 1MB chunk size, so disk-read-ahead already needs to do something with the fact that every 1Mb sequential access is interrupted.

Yes, but having working read-ahead (or write-behind) for 1MB is better than nothing.

I do want to increase the chunk size, or even move to JBOD, to avoid these issues.

> 
> In the last version of this PR I changed the fanout approach -- instead of round-robin each and every request, I split the file in 16Mb chunks and each chunk is always served by a single shard, the mapping is `(this_shard_id() + pos / 16Mb) % smp::count`. Does this help?

It helps sequential I/O, but doesn't help random.

xemul:
> But does the plan need to be limited to the next io_lateny_goal? It can be longer. Let's assume we have complete visibility into the queues and predictions. 

That's pretty strong assumption. In the context when shard IO happens we have neither future/promises/coroutines, no corss-shard locks (well, we _can_ have them, but it's not going to be reactor friendly), only atomics and lockfree stuff.

> Can't we lay out a plan for the next 10 io_latency_goal ticks? Then something with 100 shares gets every 10th tick. Of course we won't implement it like that (what about 1 share?) but conceptually.

In theory -- yes. This is what a single shard does -- it prepares a plan, then picks several requests from its head each tick. We want to make similar plan ~~node~~ io_group-wide with the restrictions described above :smile:

xemul:
> > > Sequential reads get reordered, possibly confusing the disk's read-ahead stuff.
> > 
> > 
> > The /var/lib/scylla typically sit on a raid0 with 1MB chunk size, so disk-read-ahead already needs to do something with the fact that every 1Mb sequential access is interrupted.
> 
> Yes, but having working read-ahead (or write-behind) for 1MB is better than nothing.
> 
> I do want to increase the chunk size, or even move to JBOD, to avoid these issues.
> 
> > In the last version of this PR I changed the fanout approach -- instead of round-robin each and every request, I split the file in 16Mb chunks and each chunk is always served by a single shard, the mapping is `(this_shard_id() + pos / 16Mb) % smp::count`. Does this help?
> 
> It helps sequential I/O, but doesn't help random.

But how does 16Mb fanout chunk makes things worse given 1Mb raid chunk?

avikivity:
> > But does the plan need to be limited to the next io_lateny_goal? It can be longer. Let's assume we have complete visibility into the queues and predictions.
> 
> That's pretty strong assumption. In the context when shard IO happens we have neither future/promises/coroutines, no corss-shard locks (well, we _can_ have them, but it's not going to be reactor friendly), only atomics and lockfree stuff.


I'm trying to see what's the best we can do. Assuming full locking, or perhaps seqlock style to copy the data out coherently and do some work on it. Nothing bad happens if the plan is slightly outdated.

I want to ignore implementation barriers to see what's the best we can do, then add them back.

> 
> > Can't we lay out a plan for the next 10 io_latency_goal ticks? Then something with 100 shares gets every 10th tick. Of course we won't implement it like that (what about 1 share?) but conceptually.
> 
> In theory -- yes. This is what a single shard does -- it prepares a plan, then picks several requests from its head each tick. We want to make similar plan ~node~ io_group-wide with the restrictions described above 😄

The amount of contention isn't huge. We can access shared state once per tick, it's negligible compared the lock contention that shared-everything architectures do. Let's say we have 48 shards per NUMA node, at 2kHz that's an access every 10 usec on average, or 20,000 cycles. And we can have two groups per NUMA node if we want to reduce contention.

submit_to isn't free, it does amortize well but it's still k*48^2 cache misses every tick, plus lots of latency, plus misses 
 for the data and result, plus the misses we induce in the kernel.

avikivity:
> > > > Sequential reads get reordered, possibly confusing the disk's read-ahead stuff.
> > > 
> > > 
> > > The /var/lib/scylla typically sit on a raid0 with 1MB chunk size, so disk-read-ahead already needs to do something with the fact that every 1Mb sequential access is interrupted.
> > 
> > 
> > Yes, but having working read-ahead (or write-behind) for 1MB is better than nothing.
> > I do want to increase the chunk size, or even move to JBOD, to avoid these issues.
> > > In the last version of this PR I changed the fanout approach -- instead of round-robin each and every request, I split the file in 16Mb chunks and each chunk is always served by a single shard, the mapping is `(this_shard_id() + pos / 16Mb) % smp::count`. Does this help?
> > 
> > 
> > It helps sequential I/O, but doesn't help random.
> 
> But how does 16Mb fanout chunk makes things worse given 1Mb raid chunk?

I didn't say it makes it worse. It helps sequential and doesn't help random.

xemul:
OK, let's try with #2294 then

PR #2290, opened 2024-06-10
build: add pyyaml to install-dependencies.sh

tchaikov:
in 6de08b079d, we introduced the dependency of pyyaml, but we failed to update `install-dependencies.sh` to install it.

in this change, we add this dependency to `install-dependencies.sh`.

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -56,6 +56,7 @@ debian_packages=(
     protobuf-compiler
     python3
     python3-pyelftools
+    python3-yaml
     ragel
     stow
     systemtap-sdt-dev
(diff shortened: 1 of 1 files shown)


tchaikov:
please note, scylladb already has this dependency listed in its `install-dependencies.sh`, see https://github.com/scylladb/scylladb/blob/55ed18db073e106d7651f44c3fe6c69fa6cb684a/install-dependencies.sh#L127

avikivity:
Do I need to regenerate the docker image?

tchaikov:
> Do I need to regenerate the docker image?

@avikivity no need anymore. the CI does not use the docker image at this moment. it uses the vanilla "fedora:40" image.

PR #2292, opened 2024-06-12
Update tutorial.md to reflect updated preemption methods

travisdowns:
I was actually surprised to see that `then()` no longer does a preemption check (and this certainly leads to better code generation!).

I checked the tutorial and notice that it refers a different mechanism entirely, a 256-continuation counter but I can't find any trace of this in the code, so I assume it preceeded the time-based preemption check which has subsequently been removed. 

So:

 - Rip out mention of the 256-counter
 - Talk about preemption which isn't really covered in the tutorial and add that then() is not a suspension point (but co_await is, by default)

### DIFF
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -316,9 +316,7 @@ The function `slow()` deserves more explanation. As usual, this function returns
 This example begins to show the convenience of the futures programming model, which allows the programmer to neatly encapsulate complex asynchronous operations. `slow()` might involve a complex asynchronous operation requiring multiple steps, but its user can use it just as easily as a simple `sleep()`, and Seastar's engine takes care of running the continuations whose futures have become ready at the right time.
 
 ## Ready futures
-A future value might already be ready when `then()` is called to chain a continuation to it. This important case is optimized, and *usually* the continuation is run immediately instead of being registered to run later in the next iteration of the event loop.
-
-This optimization is done *usually*, though sometimes it is avoided: The implementation of `then()` holds a counter of such immediate continuations, and after many continuations have been run immediately without returning to the event loop (currently the limit is 256), the next continuation is deferred to the event loop in any case. This is important because in some cases (such as future loops, discussed later) we could find that each ready continuation spawns a new one, and without this limit we can starve the event loop. It is important not to starve the event loop, as this would starve continuations of futures that weren't ready but have since become ready, and also starve the important **polling** done by the event loop (e.g., checking whether there is new activity on the network card).
+A future value might already be ready when `then()` is called to chain a continuation to it. This important case is optimized, and the continuation is run immediately instead of being registered to run later in the next iteration of the event loop.
 
 `make_ready_future<>` can be used to return a future which is already ready. The following example is identical to the previous one, except the promise function `fast()` returns a future which is already ready, and not one which will be ready in a second as in the previous example. The nice thing is that the consumer of the future does not care, and uses the future in the same way in both cases.
 
@@ -336,6 +334,16 @@ seastar::future<> f() {
     });
 }
 ```
+
+## Preemption and Task Quota
+
+As described above, an existing fiber of execution will yield back to the event loop when it performs a blocking operation such as IO or sleeping, as it has no more work to do until this blocking operation completes. Should a fiber have a lot of CPU bound work to do without any intervening blocking operations, however, it is important that execution is still yielded back to the event loop periodically.
+
+This is implemented via _preemption_: which can only occur at specific preemption points. At these points the fiber's remaining _task quota_ is checked and it has been exceeded the fiber yields. The task quota is a measure of how long tasks should be allowed to run before yielding to the event loop, and is set to 500 µs by default.
+
+It is important not to starve the event loop, as this would starve continuations of futures that weren't ready but have since become ready, and also starve the important **polling** done by the event loop (e.g., checking whether there is new activity on the network card). For example, iterating over a large container while doing CPU-bound work without any suspension points could starve the reactor and cause a _reactor stall_, which refers to a substantial period of time (e.g., more than 20 milliseconds) during which a task does not yield.
+
+Many seastar constructs such as looping constructs have built-in preemption points. You may also insert your own preemption points by calling `seastar::maybe_yield`, which performs a preemption check. Coroutines will also perform a preemption check at each `co_await`. Note that there is _not_ a preemption check between continuations attached to a future with `then()`, so a recursive future loop without explicit preemption checks may starve the reactor.
 
 # Coroutines
 
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -432,7 +432,7 @@ need to:
 
 But there is an exception, if the buffer's size is one, we assume that the programmer is likely to use `std::optional`
 for the bounded buffer, so it's not required to pass the maximum size of the buffer as the first parameter in this case.
-But if a coroutine uses `std::optional` as its buffer, and its function sigature still lists the size as its first
+But if a coroutine uses `std::optional` as its buffer, and its function signature still lists the size as its first
 parameter, it will not break anything. As this parameter will just be ignored by the underlying implementation.
 
 Following is an example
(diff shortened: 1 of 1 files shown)


doc/tutorial.md
```diff
@@ -316,9 +316,7 @@ The function `slow()` deserves more explanation. As usual, this function returns
...
-
-This optimization is done *usually*, though sometimes it is avoided: The implementation of `then()` holds a counter of such immediate continuations, and after many continuations have been run immediately without returning to the event loop (currently the limit is 256), the next continuation is deferred to the event loop in any case. This is important because in some cases (such as future loops, discussed later) we could find that each ready continuation spawns a new one, and without this limit we can starve the event loop. It is important not to starve the event loop, as this would starve continuations of futures that weren't ready but have since become ready, and also starve the important **polling** done by the event loop (e.g., checking whether there is new activity on the network card).
+A future value might already be ready when `then()` is called to chain a continuation to it. This important case is optimized, and the continuation is run immediately instead of being registered to run later in the next iteration of the event loop.
```
tchaikov, line 319: FWIW, the limit was killed in ab74536d0fc312ad56facfe7be54c56f0dceb441 .

doc/tutorial.md
```diff
@@ -337,6 +335,16 @@ seastar::future<> f() {
...
+
+This is implemented via _preemption_: which can only occur at specific premption points. At these points the fiber's remaining _task quota_ is checked and it has been exceeded the fiber yields. The task quota is a measure of how long tasks should be allowed to run before yielding to the event loop, and is set to 500 µs by default.
+
+It is important not to starve the event loop, as this would starve continuations of futures that weren't ready but have since become ready, and also starve the important **polling** done by the event loop (e.g., checking whether there is new activity on the network card).
+
+Many seastar constructs such as looping constructs have built-in preemption points. You may also insert your own preemption points by calling `seastar::maybe_yield`, which performs a preemption check. Couroutines will also perform a preemption check at each `co_await`. Note that there is _not_ a preemption check between continuations attached to a future with `then()`, so a recursive future loop without explicit preemption checks may starve the reactor.
```
tchaikov, line 346: another good example is probably a tight loop iterating in a container with many elements.

tchaikov:
@nyh hi Nadav, could you please take a look?

doc/tutorial.md
travisdowns: Heh, in 2016, so better late than never to reflect this in the tutorial :).

doc/tutorial.md
travisdowns: I added this example in push https://github.com/scylladb/seastar/commit/a81fbc74ca5f443f3127b65a92620cd750cde24f.

travisdowns:
Push https://github.com/scylladb/seastar/commit/a81fbc74ca5f443f3127b65a92620cd750cde24f fixes a couple of typos in the new text and adds a note about CPU-bound loops as suggested by @tchaikov .

travisdowns:
https://github.com/scylladb/seastar/pull/2292/commits/755a43131d00b16c180fa78b2fa8a8ea83c719d0 fixes a few existing typos as well.

tchaikov (approved):
lgtm. thanks!

doc/tutorial.md
nyh: @travisdowns @tchaikov actually, the commit that @tchaikov found didn't cancel preemption in then()... The change back then was not to do this based on counting (every 256 continuations) but based on the "preemption timer" (need_preempt()). The change you were looking for was f58515a95cbf304ea48e0f5d8d046b7ab4357789 and was done in 2020.

Thanks for looking into this issue, it's indeed very important to document this clearly and correctly, because it's hard to guess what needs to be done to avoid "stalls" (long periods without preemption). We need to explain that then() does **not** do preemptions these days, but our loop functions do (by "doing preemption" I mean only if need_preempt() - which needs to be explained too). And most importantly for modern code based on coroutines: Our "co_await()" **does** (unless I'm misremembering?) check for need_preempt(). Why did we decide in 2020 we need to remove the need_preempt() check from then(), only to bring it back into co_await()? Maybe @avikivity remembers, I don't :-(
I see you started discussing this below. Good.

doc/tutorial.md
```diff
@@ -337,6 +335,16 @@ seastar::future<> f() {
 }
 ```
 
+## Preemption and Task Quota
+
+As described above, an existing fiber of execution will yield back to the event loop when it performs a blocking operation such as IO or sleeping, as it has no more work to do until this blocking operation completes. Should a fiber have a lot of CPU bound work to do without any intervening blocking operations, however, it is important that execution is still yielded back to the event loop periodically.
+
+This is implemented via _preemption_: which can only occur at specific preemption points. At these points the fiber's remaining _task quota_ is checked and it has been exceeded the fiber yields. The task quota is a measure of how long tasks should be allowed to run before yielding to the event loop, and is set to 500 µs by default.
```
nyh, line 342: Maybe we should mention that this is a configurable parameter and how to configure it.

doc/tutorial.md
```diff
@@ -337,6 +335,16 @@ seastar::future<> f() {
...
+
+It is important not to starve the event loop, as this would starve continuations of futures that weren't ready but have since become ready, and also starve the important **polling** done by the event loop (e.g., checking whether there is new activity on the network card). For example, iterating over a large container while doing CPU-bound work without any suspension points could starve the reactor and cause a _reactor stall_, which refers to a substantial period of time (e.g., more than 20 milliseconds) during which a task does not yield.
+
+Many seastar constructs such as looping constructs have built-in preemption points. You may also insert your own preemption points by calling `seastar::maybe_yield`, which performs a preemption check. Coroutines will also perform a preemption check at each `co_await`. Note that there is _not_ a preemption check between continuations attached to a future with `then()`, so a recursive future loop without explicit preemption checks may starve the reactor.
```
nyh, line 346: There is no `seastar::maybe_yield` unless I'm missing something... There's seastar::thread::maybe_yield which only works inside a seastar::thread, and there's seastar::coroutine::maybe_yield()  (which you should use in a long loop in a coroutined function). For continuation-based code, you should use need_preempt() in a more elaborate manner (I'm not even sure what example to give for this).

nyh (approved):

travisdowns:
@nyh wrote:

> @travisdowns @tchaikov actually, the commit that @tchaikov found didn't cancel preemption in then()... The change back then was not to do this based on counting (every 256 continuations) but based on the "preemption timer" (need_preempt()). The change you were looking for was https://github.com/scylladb/seastar/commit/f58515a95cbf304ea48e0f5d8d046b7ab4357789 and was done in 2020.

Right, I was interested in _both_ when the 256 -> need_preempt() switch happened and also when the need_preempt() -> nothing switch happened so @tchaikov was responding to my comment about the first switch (as the tutorial still mentioned the 256 counter approach). 

> Why did we decide in 2020 we need to remove the need_preempt() check from then(), only to bring it back into co_await()? Maybe @avikivity remembers, I don't :-(
I see you started discussing this below. Good.

This asymmetry also surprised me. One justification is that it's actually quite hard to stall just due to the lack of then()-preemption, because you can't write loops only based on this construct, unless you use recursion or something else strange: so you have looping constructs and they all do check for preemption. OTOH, loops work just fine with co_await and they are natural way to write things, and so we would expect that looping are basically unused in pure coroutines code. So having the preemption check in co_await prevents these loops from stalling (of course, nothing will save you if your loop has no co_await or other preemption check at all).

Currently I am seeing that for small continuations, using `then()` can be much more efficient than coroutines, and the lack of preemption check is part of the reason (yes, there is a way to turn this off for `co_await` also).

doc/tutorial.md
travisdowns: There is a vanilla version `seastar::maybe_yield` too, here: https://github.com/scylladb/seastar/blob/d7bd2e7178f55db4e5731b7a4168e22d509d7776/include/seastar/util/later.hh#L57

It works as you expect: effectively returning the `yield()` future if the task quota has expired, or a ready future otherwise.

doc/tutorial.md
travisdowns: I am happy to add this type of thing, but I should interpret your merge as indicating that it is not necessary?

Issue #323, opened 2017-08-18
rpc client/server metrics


avikivity:
Need metrics for rpc:

connections established/dropped (and current = established - dropped)
messages sent
bytes sent

Would be nice to make it per-peer, but would mean a huge number of metrics, so no.

Issue #1067, opened 2022-05-19
RPC send path is prone to producing extra latency


xemul:
(originated from scylladb/scylla#9505)

When sending an RPC message it goes via a dedicated send fiber:

```c++
void connection::send_loop() {
    _send_loop_stopped = do_until([this] { return _error; }, [this] {
        return _outgoing_queue_cond.wait([this] { return !_outgoing_queue.empty(); }).then([this] {
            ...
            auto d = std::move(_outgoing_queue.front());
            _outgoing_queue.pop_front();
            ...
}

future<> connection::send(snd_buf buf, std::optional<rpc_clock_type::time_point> timeout, cancellable* cancel) {
    ...
    _outgoing_queue.emplace_back(std::move(buf));
    ...
    _outgoing_queue_cond.signal();
}
```

The fiber, queue and condvar in question were added by e536555081cde5c8b6eac5f70f691d1fc3da7a8a
Since `condition_variable.signal()` queued the send-loop's fiber for execution, the task in question may happen at the tail of a very long execution queue thus rendering unwanted extra latency for no reason.

The solution can be to replace send fiber with semaphore-guarded continuation chain. One thing to care about it the ability to cancel requests from the middle of the queue. The most important consequence of cancelling should be releasing of the message memory,

DoronArazii:
@xemul can you please consider backport ^^

xemul:
AFAIK we don't backport new features, only fixes. And perf optimization is like a new feature for me

PR #2293, opened 2024-06-12
RPC delay metrics

xemul:
The new metrics is total time it takes an RPC message to travel back and forth between verb caller and server. It implicitly consists of the time request and response messages spend in seastar and kernel queues on both sides and classical network RTT.

Calculated for verbs that require response only. First, client negotiates with the server that response frame includes the "handler duration" value, which is the handler callback execution time. Then client notices duration between sending request and receiving response minus handler execution time received from server. This is called "delay sample". Total number of samples and their duration summary are exported as counters in the metrics domain.

refs: #323 
refs: #1753 (for what metrics domain is)

### DIFF
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -118,6 +118,7 @@ struct client_options {
     /// \see resource_limits::isolate_connection
     sstring isolation_cookie;
     sstring metrics_domain = "default";
+    bool send_handler_duration = true;
 };
 
 /// @}
@@ -179,6 +180,7 @@ enum class protocol_features : uint32_t {
     CONNECTION_ID = 2,
     STREAM_PARENT = 3,
     ISOLATION = 4,
+    HANDLER_DURATION = 5,
 };
 
 // internal representation of feature data
@@ -285,6 +287,7 @@ protected:
     std::unique_ptr<compressor> _compressor;
     bool _propagate_timeout = false;
     bool _timeout_negotiated = false;
+    bool _handler_duration_negotiated = false;
     // stream related fields
     bool _is_stream = false;
     connection_id _id = invalid_connection_id;
@@ -487,7 +490,11 @@ private:
 private:
     future<> negotiate_protocol(feature_map map);
     void negotiate(feature_map server_features);
-    future<std::tuple<int64_t, std::optional<rcv_buf>>>
+    // Returned future is
+    // - message id
+    // - optional server-side handler duration
+    // - message payload
+    future<std::tuple<int64_t, std::optional<uint32_t>, std::optional<rcv_buf>>>
     read_response_frame_compressed(input_stream<char>& in);
 public:
     /**
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -430,6 +430,7 @@ class client : public rpc::connection, public weakly_referencable<client> {
     struct reply_handler_base {
         timer<rpc_clock_type> t;
         cancellable* pcancel = nullptr;
+        rpc_clock_type::time_point start;
         virtual void operator()(client&, id_type, rcv_buf data) = 0;
         virtual void timeout() {}
         virtual void cancel() {}
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -592,7 +592,7 @@ public:
     public:
         connection(server& s, connected_socket&& fd, socket_address&& addr, const logger& l, void* seralizer, connection_id id);
         future<> process();
-        future<> respond(int64_t msg_id, snd_buf&& data, std::optional<rpc_clock_type::time_point> timeout);
+        future<> respond(int64_t msg_id, snd_buf&& data, std::optional<rpc_clock_type::time_point> timeout, std::optional<rpc_clock_type::duration> handler_duration);
         client_info& info() { return _info; }
         const client_info& info() const { return _info; }
         stats get_stats() const {
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -667,6 +667,9 @@ namespace rpc {
           case protocol_features::TIMEOUT:
               _timeout_negotiated = true;
               break;
+          case protocol_features::HANDLER_DURATION:
+              _handler_duration_negotiated = true;
+              break;
           case protocol_features::CONNECTION_ID: {
               _id = deserialize_connection_id(e.second);
               break;
@@ -692,8 +695,8 @@ namespace rpc {
   //   ...  payload
   struct response_frame {
       using opt_buf_type = std::optional<rcv_buf>;
-      using return_type = std::tuple<int64_t, opt_buf_type>;
-      using header_type = std::tuple<int64_t>;
+      using return_type = std::tuple<int64_t, std::optional<uint32_t>, opt_buf_type>;
+      using header_type = std::tuple<int64_t, std::optional<uint32_t>>;
       static constexpr size_t raw_header_size = sizeof(int64_t) + sizeof(uint32_t);
       static size_t header_size() {
           static_assert(response_frame_headroom >= raw_header_size);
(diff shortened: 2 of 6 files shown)


xemul:

nyh:
Can you write a bit about the motivation of this specific metric? Is it something you have seen in some other API and found useful (if so, a reference would be nice)? What do you want to use this new feature for?

What bothers me is that the subtraction you're doing - latency measured by client minus latency measured by server - isn't exactly the RTT of anything: It isn't the RTT of the full request (because that would include the handling time, not subtracting it), and it's not the "classic" network RTT because it includes includes queueing time on both client and server.

xemul:
When looking for sources of network latency in scylladb/scylladb#9505 it was good to know how much time it takes for RPC frame to pass through seastar(send)-kernel(send)-network-kernel(recv)-seastar(recv)-and_back (for the record, one of the problematic places found was in seastar(send) -- #1067), so the idea to collect it as metrics popped up.

> What bothers me is that the subtraction you're doing - latency measured by client minus latency measured by server - isn't exactly the RTT of anything

Why not?

1. (cln) send_entry()
2. (cln) request is queued into client outgoing queue
3. (cln) request is written into the socket
5. network processing, whatever it means
6. (srv) request is read from socket
7. (srv) call handler
8. (srv) response is queued into client outgoing queue
9. (srv) response is written into the socket
10. network processing again
11. (cln) response is read from socket
12. (cln) find and wake so called "outstanding" entry
13. (cln) response is parsed and returned back

If I measure timestamps of point 1 and 13, their difference is the total time it took request to process. If I measure the time step 7 took, and subtract it from that calculated difference, I get exactly the time it took the request to travel through seastars and kernels on both sides and the network in between.

> and it's not the "classic" network RTT because it includes includes queueing time on both client and server.

But it includes the queuing time on both -- client and server, unless I mis-read the code and placed the timestamps not in points 1 and 13 as described above

include/seastar/rpc/rpc.hh
```diff
@@ -487,7 +490,7 @@ private:
 private:
     future<> negotiate_protocol(feature_map map);
     void negotiate(feature_map server_features);
-    future<std::tuple<int64_t, std::optional<rcv_buf>>>
+    future<std::tuple<int64_t, std::optional<uint32_t>, std::optional<rcv_buf>>>
```
avikivity, line 494: Better to document what the optional means.

Note: microseconds can overflow 32 bits for an rpc call.

src/rpc/rpc.cc
```diff
@@ -667,6 +667,9 @@ namespace rpc {
           case protocol_features::TIMEOUT:
               _timeout_negotiated = true;
               break;
+          case protocol_features::HANDLER_DURATION:
+              _handler_duration_negotiated = true;
+              break;
```
avikivity, line 672: Please update rpc.md.

include/seastar/rpc/rpc.hh
```diff
@@ -430,6 +430,7 @@ class client : public rpc::connection, public weakly_referencable<client> {
     struct reply_handler_base {
         timer<rpc_clock_type> t;
```
avikivity, line 431: > Now the client has two durations using which it can estimate the
rount-trip time of a packet -- the duration between sending request and
receiving response and the time it too handler to execute on the server
side. Difference is the RTT value.


What about time waiting in the send queue (usually none) or the receive queue? Or waiting for the rpc semaphore?

src/rpc/rpc.cc
```diff
@@ -877,6 +877,15 @@ namespace rpc {
...
                         sm::description("Total number of timeout responses"), { domain_l }).set_skip_when_empty(),
+                sm::make_counter("rtt_samples", std::bind(&domain::count_all, this, &stats::rtt_samples),
+                        sm::description("Total number of RTT samples"), { domain_l }),
+                sm::make_counter("rtt_total", [this] () -> double {
+                            std::chrono::duration<double> res(0);
+                            for (const auto& m : list) {
+                                res += m._c._stats.rtt_total;
+                            }
+                            return res.count();
+                        }, sm::description("Total RTT time in seconds"), { domain_l }),
```
avikivity, line 888: What does this mean? It will be an average of this-AZ RPC + other-AZ RPC + other-region RPC. Let's say the average is 10ms. What does it mean?

include/seastar/rpc/rpc.hh
xemul: It's ~1 hour. Are there handlers that last more?

include/seastar/rpc/rpc.hh
xemul: This new metrics include all of these.
Or does your question really mean "lets adding those metrics too"?

src/rpc/rpc.cc
xemul: No, no, it won't be averaged between different DCs. The metrics is per metrics domain, Scylla configures metrics domains per-DC and per-verb-group-index

include/seastar/rpc/rpc.hh
avikivity: I mean that the different is not RTT, it's RTT + waits in queues we don't (and can't) measure.

src/rpc/rpc.cc
xemul: That's how: scylladb/scylladb#15785

> The domain as this patch defines it includes two strings:
> 
> First, the datacenter the server lives in. This is because grouping metrics for connections to different datacenters makes little sense for several reasons. For example -- packet delays will differ for local-DC vs cross-DC traffic and mixing those latencies together is pointless. Another example -- the amount of traffic may also differ for local- vs cross-DC connections e.g. because of different usage of enryption and/or compression.
> 
> Second, each verb-idx gets its own domain. That's to be able to analyze e.g. query-related traffic from gossiper one. For that the existing isolation cookie is taken as is.

avikivity: Ah wow

include/seastar/rpc/rpc.hh
avikivity: RTT is the network time without any queues (which ping measures if it doesn't have a bad day).

Do you subtract service time from the measured round trip?

src/rpc/rpc.cc
nyh: I share Avi's wow - I didn't even know a concept of "metric domains" existed. As the number of Scylla and Seastar developers grows, we need to try to be more diligent about documenting things that other developers  will need to know.  When I wrote Scylla's docs/dev/metrics.md, I spent many hours trying to figure out all the details of how metrics were used in Scylla and explain them all (I didn't write any of the metrics code, so I had to dig and talk to Amnon to find the details). But clearly, in the 6 years that have passed (time flies...) things have changed. We should update this document - and document more of Scylla's basic utilities that every developer needs to know. Moreover, we should perhaps think of documenting the metrics feature in Seastar's documentation, not just in Scylla's...

include/seastar/rpc/rpc.hh
xemul: > RTT is the network time without any queues (which ping measures if it doesn't have a bad day).

The "without any queues" is pretty optimistic requirement, I would say. I admit that outgoing is likely to take queue-less paths even in the kernel, but incoming all consists of queues.

> Do you subtract service time from the measured round trip?

Depends on what "service time" is. If it's duration of the verb handler, then of course, that's the point (https://github.com/scylladb/seastar/commit/9bbe38c522ceab6f7d46775006e09a542fc37e2b)

```patch
@@ -953,8 +964,13 @@ namespace rpc {
                           _error = true;
                       } else if (it != _outstanding.end()) {
                           auto handler = std::move(it->second);
+                          auto ht = std::get<1>(msg_id_and_data); // ht stands for "handler time"
                           _outstanding.erase(it);
                           (*handler)(*this, msg_id, std::move(data.value()));
+                          if (ht) {
+                              _stats.rtt_samples++;
+                              _stats.rtt_total += (rpc_clock_type::now() - handler->start) - std::chrono::microseconds(*ht);
+                          }
                       } else if (msg_id < 0) {
                           try {
                               std::rethrow_exception(unmarshal_exception(data.value()));
```

src/rpc/rpc.cc
nyh: It would have been nice to have in Scylla metrics.md become a "book" that documents in length and in depth the meaning of every single metric (such documentation is sorely missing, I can rarely understand anything in the metrics because of the opaque one-sentence descriptions of metric), but what I focused in the current metrics.md is in the overall structure of all metrics and how to use them. For example, I described the metric labels "shard" and "instance" that every metric has. If recently we added per-scheduling-group labels, and per-"domain" (whatever that is) labels that applies to many different metrics, it would be nice to explain them. If Seastar brings its own set of RPC metrics, would be nice to explain it.
In general, I wouldn't complain if more documentation gets written :-)

PR #2294, opened 2024-06-13
Implement cross-shard consumption fairness

### LINKED ISSUE #1083: IO capacity balancing is not well balanced

xemul:
Current IO queues design assumes that IO workload is very uniform on different shards in a sense that -- all classes are more-or-less equally loaded on different shards. In reality that's not true and some shards can easily get more requests in one of its class than the others.

This leads to a very nasty consequence. Consider a corner case -- two classes, A and B, with shares 100 and 1000 respectively, two shards. Class A is active on shard-0 only, while class B is active on shard-1 only. We expect, that they share disk bandwidth capacity in 1:10 proportion, but in reality it's going to be 1:1, because cross-shard queue doesn't preempt (it doesn't, because load is expected to be even on all shards).

The solution here is to implement the cross-class fairness approach on the shard-level. For that, each fair_queue accumulates its total amount of requests costs dispatched (cost = request.capacity / class.shares), and on every poll -- check its accumulator against those of other shards. If the local value is "somewhat ahead of" the others, the poll is skipped until later.

refs: https://github.com/scylladb/seastar/issues/1430
fixes: https://github.com/scylladb/seastar/issues/1083

Tested with io_tester on 2-shards setup from #2289

```
- name: shard_0
  shard_info:
    parallelism: 32
    reqsize: 4kB
    shares: 100
  shards:
  - '0'
  type: randread
- name: shard_1
  shard_info:
    parallelism: 32
    reqsize: 4kB
    shares: 800
  shards:
  - '1'
  type: randread
```

The result is

```
---
- shard: 0
  shard_0:
    throughput: 89627.8828  # kB/s
    IOPS: 22407.0703
    latencies:  # usec
      average: 1427
      p0.5: 1627
      p0.95: 1653
      p0.99: 1674
      p0.999: 1776
      max: 73161
    stats:
      total_requests: 672223
      io_queue_total_exec_sec: 38.740494826001402
      io_queue_total_delay_sec: 920.08525591999751
      io_queue_total_operations: 672255
      io_queue_starvation_time_sec: 0.86344176700005526
      io_queue_consumption: 3.7791568234562876
      io_queue_adjusted_consumption: 0.038457549512386321
- shard: 1
  shard_1:
    throughput: 621912.375  # kB/s
    IOPS: 155478.797
    latencies:  # usec
      average: 205
      p0.5: 199
      p0.95: 227
      p0.99: 271
      p0.999: 391
      max: 1470
    stats:
      total_requests: 4664398
      io_queue_total_exec_sec: 278.93417872600429
      io_queue_total_delay_sec: 665.67777301317381
      io_queue_total_operations: 4664430
      io_queue_starvation_time_sec: 5.5724000000000004e-05
      io_queue_consumption: 26.221615996956825
      io_queue_adjusted_consumption: 0.03324427980184555
...
```

### DIFF
--- a/apps/io_tester/ioinfo.cc
+++ b/apps/io_tester/ioinfo.cc
@@ -75,7 +75,6 @@ int main(int ac, char** av) {
                             out << YAML::EndMap;
 
                             const auto& fg = internal::get_fair_group(ioq, internal::io_direction_and_length::write_idx);
-                            out << YAML::Key << "per_tick_grab_threshold" << YAML::Value << fg.per_tick_grab_threshold();
 
                             const auto& tb = fg.token_bucket();
                             out << YAML::Key << "token_bucket" << YAML::BeginMap;
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -26,6 +26,7 @@
 #include <seastar/core/shared_ptr.hh>
 #include <seastar/core/circular_buffer.hh>
 #include <seastar/core/metrics_registration.hh>
+#include <seastar/core/lowres_clock.hh>
 #include <seastar/util/shared_token_bucket.hh>
 
 #include <chrono>
@@ -347,6 +348,14 @@ private:
     size_t _nr_classes = 0;
     capacity_t _last_accumulated = 0;
     capacity_t _total_accumulated = 0;
+
+    // Amortize balance checking by assuming that once balance achieved,
+    // it would remain such for the "quiscent" duration. Increase this
+    // duration every time the assumption keeps true, but not more than
+    // tau. When the balance is lost, reset back to frequent checks.
+    static constexpr auto minimal_quiscent_duration = std::chrono::microseconds(100);
+    std::chrono::microseconds _balance_quiscent_duration = minimal_quiscent_duration;
+    timer<lowres_clock> _balance_timer;
     // Maximum capacity that a queue can stay behind other shards
     //
     // This is similar to priority classes fall-back deviation and it's
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -217,7 +217,6 @@ private:
      */
 
     token_bucket_t _token_bucket;
-    const capacity_t _per_tick_threshold;
 
     // Capacities accumulated by queues in this group. Each queue tries not
     // to run too far ahead of the others, if it does -- it skips dispatch
@@ -265,7 +264,6 @@ public:
     fair_group(fair_group&&) = delete;
 
     capacity_t maximum_capacity() const noexcept { return _token_bucket.limit(); }
-    capacity_t per_tick_grab_threshold() const noexcept { return _per_tick_threshold; }
     capacity_t grab_capacity(capacity_t cap) noexcept;
     clock_type::time_point replenished_ts() const noexcept { return _token_bucket.replenished_ts(); }
     void replenish_capacity(clock_type::time_point now) noexcept;
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -219,7 +219,19 @@ private:
     token_bucket_t _token_bucket;
     const capacity_t _per_tick_threshold;
 
+    // Capacities accumulated by queues in this group. Each queue tries not
+    // to run too far ahead of the others, if it does -- it skips dispatch
+    // loop until next tick in the hope that other shards would grab the
+    // unused disk capacity and will move their counters forward.
+    std::vector<capacity_t> _balance;
+
 public:
+
+    // Maximum value the _balance entry can get
+    // It's also set when a queue goes idle and doesn't need to participate
+    // in accumulated races. This value is still suitable for comparisons
+    // of "active" queues
+    static constexpr capacity_t max_balance = std::numeric_limits<signed_capacity_t>::max();
 
     // Convert internal capacity value back into the real token
     static double capacity_tokens(capacity_t cap) noexcept {
@@ -267,6 +279,10 @@ public:
     }
 
     const token_bucket_t& token_bucket() const noexcept { return _token_bucket; }
+
+    capacity_t current_balance() const noexcept;
+    void update_balance(capacity_t) noexcept;
+    void reset_balance() noexcept;
 };
 
 /// \brief Fair queuing class
@@ -332,6 +348,15 @@ private:
     std::vector<std::unique_ptr<priority_class_data>> _priority_classes;
     size_t _nr_classes = 0;
     capacity_t _last_accumulated = 0;
+    capacity_t _total_accumulated = 0;
+    // Maximum capacity that a queue can stay behind other shards
+    //
+    // This is similar to priority classes fall-back deviation and it's
+    // calculated as the number of capacity points a group with 1 share
+    // accumulates over tau
+    //
+    // Check max_deviation math in push_priority_class_from_idle())
+    const capacity_t _max_imbalance;
 
     /*
      * When the shared capacity os over the local queue delays
@@ -362,6 +387,8 @@ private:
     enum class grab_result { grabbed, cant_preempt, pending };
     grab_result grab_capacity(const fair_queue_entry& ent) noexcept;
     grab_result grab_pending_capacity(const fair_queue_entry& ent) noexcept;
+
+    bool balanced() noexcept;
 public:
     /// Constructs a fair queue with configuration parameters \c cfg.
     ///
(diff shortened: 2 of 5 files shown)


avikivity:
> This leads to a very nasty consequence. Consider a corner case -- two classes, A and B, with shares 100 and 1000 respectively, two shards. Class A is active on shard-0 only, while class B is active on shard-1 only. We expect, that they share disk bandwidth capacity in 1:10 proportion, but in reality it's going to be 1:1, because cross-shard queue doesn't preempt (it doesn't, because load is expected to be even on all shards).

I don't mean to ignore the problem, but I think "very nasty" is an exaggeration. A/0 could have expected 0.44 share of the capacity (1000/2200) and got 0.5. B/1 could expect 0.04 share of the capacity, and got 0.5. So neither has anything to complain about.

I suppose with a larger number of shards and only a small number of sg/shard combinations active the problem is more interesting. Especially when the disk isn't too fast.

avikivity:
Looks reasonable (I don't claim to understand it deeply).

How can we assure this doesn't cause regressions?

xemul:
> I don't mean to ignore the problem, but I think "very nasty" is an exaggeration.

Probably

> A/0 could have expected 0.44 share of the capacity (1000/2200) and got 0.5. B/1 could expect 0.04 share of the capacity, and got 0.5. So neither has anything to complain about.

:smile: 

Not quite. Provided A is query class and B is compaction class, 0.44 reads vs 0.04 writes is not the same as 0.5 reads vs 0.5 writes, from [diskplorer plots](https://github.com/scylladb/diskplorer/tree/master/latency-matrix-results) only i4i can handle it.

> How can we assure this doesn't cause regressions?

I will :( Currently, if one shard gets stuck, other shards just continue moving. With this patch, they all will get stuck.

avikivity:
> > I don't mean to ignore the problem, but I think "very nasty" is an exaggeration.
> 
> Probably
> 
> > A/0 could have expected 0.44 share of the capacity (1000/2200) and got 0.5. B/1 could expect 0.04 share of the capacity, and got 0.5. So neither has anything to complain about.
> 
> 😄
> 
> Not quite. Provided A is query class and B is compaction class, 0.44 reads vs 0.04 writes is not the same as 0.5 reads vs 0.5 writes, from [diskplorer plots](https://github.com/scylladb/diskplorer/tree/master/latency-matrix-results) only i4i can handle it.

But isn't the 0.5 normalized? So if the disk can do 6GB/s read and 1GB/s write, 0.5r / 0.5w = 3GB/s reads and 0.5GB/s writes, which should work according to the model.

xemul:
> So if the disk can do 6GB/s read and 1GB/s write, 0.5r / 0.5w = 3GB/s reads and 0.5GB/s writes, which should work according to the model.

Yes, from the pure model perspective it's all OK, but in reality some plots are convex, so the 0.5/0.5 point happens to be "more purple" (larger read latency), while 0.4/0.04 is "more cyan" (smaller read latency)

model

xemul:
upd:
- scale tokens 8 times to avoid overflows
- fix f.q. tests (hopefully) to remember that fair queue does time-based dispatching

xemul:
upd:
- lost unused variable caused compilation to fail

xemul:
upd:
- add manual (quite long) test for check cross-shard fairness

xemul:
upd:
- rebased to check CI passes

avikivity:
> > So if the disk can do 6GB/s read and 1GB/s write, 0.5r / 0.5w = 3GB/s reads and 0.5GB/s writes, which should work according to the model.
> 
> Yes, from the pure model perspective it's all OK, but in reality some plots are convex, so the 0.5/0.5 point happens to be "more purple" (larger read latency), while 0.4/0.04 is "more cyan" (smaller read latency)

Then we should either adjust the model (and teach iotune to learn it) or add a safety factor that pushes the entire line down.

We should match the scheduler to the model and the model to the disk. If we skip a step we'll get lost.

xemul:
> Then we should either adjust the model (and teach iotune to learn it) or add a safety factor that pushes the entire line down.

It's already there -- the `rate_factor` parameter in io-properties.yaml is this K parameter from the model equation 

```
     * Throttling formula:
     *
     *    Bw/Bw_max + Br/Br_max + Ow/Ow_max + Or/Or_max <= K
```

> We should match the scheduler to the model and the model to the disk. If we skip a step we'll get lost.

I don't disagree. My "not quite" observation wasn't the justification of this PR, as it doesn't touch the model

xemul:
updated the PR description _not_ to fix #1430 , as it still doesn't help high-prio requests to advance in the queue in case of symmetrical (across shards) workload -- shards forward their accumulators equally not giving any advantage to each other.

However, the workload from #1430 clearly renders ~10% smaller tail latency for high-prio interactive class (4 shards)

On master

```
    IOPS: 244.480713
      p0.5: 318
      p0.95: 532
      p0.99: 1873
      p0.999: 4863
    IOPS: 252.025391
      p0.5: 318
      p0.95: 511
      p0.99: 1882
      p0.999: 4186
    IOPS: 250.350845
      p0.5: 319
      p0.95: 537
      p0.99: 2584
      p0.999: 4942
    IOPS: 245.260681
      p0.5: 315
      p0.95: 555
      p0.99: 2164
      p0.999: 4563
```

this PR

```
    IOPS: 245.461426
      p0.5: 323
      p0.95: 488
      p0.99: 1555
      p0.999: 3950
    IOPS: 244.498688
      p0.5: 320
      p0.95: 526
      p0.99: 1562
      p0.999: 3649
    IOPS: 250.401276
      p0.5: 321
      p0.95: 501
      p0.99: 2013
      p0.999: 4083
    IOPS: 247.366104
      p0.5: 318
      p0.95: 513
      p0.99: 2093
      p0.999: 4005
```

Possible explanation -- high-prio class is low-rate (250 RPS) and the queue is empty, so when one shard emits a request, it forwards its accumulator thus reducing its chance on dispatching requests in the next tick, so the sink has more room for other shards.

Still, that's not true preemption #1430 implies

swasik:
@xemul What is the status here? Are we just waiting for the review as indicated in PR or is something else left to be done?

bhalevy:
@xemul where do we stand here, other then rebasing to resolve conflicts, what else is missing to advance this change forward?

xemul:
The patch is proven to work with a small test, to have some strong argument "for" it we should come up with a "real-life" test that would demonstrate that things really get better.

Also, [there's](https://github.com/scylladb/seastar/pull/2294#issuecomment-2166473403) a problem that this patch introduces -- with it a shard that slows down slows other shards' IO too. But it's sort of two sides of a coin -- it _intentionally_ doesn't let shards run away from each other in order to keep their progress "balanced"

vladzcloudius:
> The patch is proven to work with a small test, to have some strong argument "for" it we should come up with a "real-life" test that would demonstrate that things really get better.
> 
> Also, [there's](https://github.com/scylladb/seastar/pull/2294#issuecomment-2166473403) a problem that this patch introduces -- with it a shard that slows down slows other shards' IO too. But it's sort of two sides of a coin -- it _intentionally_ doesn't let shards run away from each other in order to keep their progress "balanced"

This means that the system is going to work as fast as its slowest shard which is not ideal to say the least.
(Note that the shard can get slow not only because I/O gets slow - it can simply be that the task that waited for I/O simply doesn't get CPU because some other task hogs it).

xemul:
> This means that the system is going to work as fast as its slowest shard which is not ideal to say the least.

This statement is the same as "a shard works as fast as the slowest sched class", but we're fine with it. It's unavoidable, if you make two entities share something in a fair manner, you link them with each other, and make one work dependent on the other

vladzcloudius:
> > This means that the system is going to work as fast as its slowest shard which is not ideal to say the least.
> 
> This statement is the same as "a shard works as fast as the slowest sched class", but we're fine with it. 

We are very much **not** fine with that either. Slowest scheduling class is not supposed to slow down other classes beyond the desired balance defined by shards allocation to each of the classes.

This can be broken if there are "stalls" in any of the tasks, e.g. there is a long not-preempting processing that continues longer than a task quota. However every such stall should and is treated as a bug.

> It's unavoidable, if you make two entities share something in a fair manner, you link them with each other, and make one work dependent on the other

Maybe it's unavoidable in the context of this PR but it doesn't make it ok in `seastar` which is all about processing on shards being independent (a.k.a. "share nothing").

And while we could be "fine" (which we are not!) with a specific shard being "slow" because of something going on on it - we should not be "fine" at all when one shard starts slowing down other shards.
This would effectively cancel all "share nothing" paradigm.

xemul:
Closing one in favor of #3295

PR #2295, opened 2024-06-14
linux_perf_event: exclude_idle only on x86_64

bhalevy:
Commit 93f19d39526c68defac6860a15c77160da273606
added `exclude_idle = 1` to `linux_perf_event`.
However, as reported in
https://github.com/scylladb/scylladb/issues/19227#issuecomment-2165577271, `exclude_idle` is not supported on ARM platforms.

This change sets `exclude_idle` only on known-to-work architectures
(presently, it's only x86_64), assuming it is initialized to 0 as all
other unset bitfields in `perf_event_attr`.

Fixes scylladb/seastar#2298

### DIFF
--- a/tests/perf/linux_perf_event.cc
+++ b/tests/perf/linux_perf_event.cc
@@ -90,7 +90,11 @@ make_linux_perf_event(unsigned config, pid_t pid = 0, int cpu = -1, int group_fd
             .disabled = 1,
             .exclude_kernel = 1,
             .exclude_hv = 1,
+#if defined(__x86_64__)
+            // exclude_idle is not supported on all architectures (e.g. aarch64)
+            // so enable it selectively only on architectures that support it.
             .exclude_idle = 1,
+#endif
             }, pid, cpu, group_fd, flags);
 }
 
--- a/tests/perf/linux_perf_event.cc
+++ b/tests/perf/linux_perf_event.cc
@@ -81,28 +81,25 @@ linux_perf_event::disable() {
     ::ioctl(_fd, PERF_EVENT_IOC_DISABLE, 0);
 }
 
-linux_perf_event
-linux_perf_event::user_instructions_retired() {
+static linux_perf_event
+make_linux_perf_event(unsigned config, pid_t pid = 0, int cpu = -1, int group_fd = -1, unsigned long flags = 0) {
     return linux_perf_event(perf_event_attr{
             .type = PERF_TYPE_HARDWARE,
             .size = sizeof(struct perf_event_attr),
-            .config = PERF_COUNT_HW_INSTRUCTIONS,
+            .config = config,
             .disabled = 1,
             .exclude_kernel = 1,
             .exclude_hv = 1,
             .exclude_idle = 1,
-            }, 0, -1, -1, 0);
+            }, pid, cpu, group_fd, flags);
+}
+
+linux_perf_event
+linux_perf_event::user_instructions_retired() {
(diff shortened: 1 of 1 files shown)


michoecho (approved):

bhalevy:

avikivity:
Doesn't it fix an open bug?

tests/perf/linux_perf_event.cc
```diff
@@ -90,7 +90,9 @@ linux_perf_event::user_instructions_retired() {
             .disabled = 1,
             .exclude_kernel = 1,
             .exclude_hv = 1,
+#if !defined(__aarch64__)
             .exclude_idle = 1,
+#endif
```
avikivity, line 95: Where is it documented? Perhaps we should opt-in where it is supported?

avikivity: Also, need some comment, it looks completely obscure.

bhalevy: https://man7.org/linux/man-pages/man2/perf_event_open.2.html

bhalevy:
```
       EPERM  Returned on many (but not all) architectures when an
              unsupported exclude_hv, exclude_idle, exclude_user, or
              exclude_kernel setting is specified.
```

bhalevy: Since we've lived without it until https://github.com/scylladb/seastar/commit/93f19d39526c68defac6860a15c77160da273606 we can opt-in and include it only on x86_64 where we know it works (and add a comment about that)

bhalevy:
> Doesn't it fix an open bug?

The bug was opened in scylladb (https://github.com/scylladb/scylladb/issues/19227) so it updating the seastar submodule should be marked as fixing it.
I can open a corresponding seastar bug if you think it's needed.

bhalevy:
- rebased (checkheaders issue seems to be fixed already in latest master)

fruch:
@bhalevy what's holding this one from getting merge ?

bhalevy:
> @bhalevy what's holding this one from getting merge ?

Nothing, now that CI has passed.

@scylladb/seastar-maint please merge

tests/perf/linux_perf_event.cc
```diff
@@ -81,28 +81,25 @@ linux_perf_event::disable() {
...
     return linux_perf_event(perf_event_attr{
             .type = PERF_TYPE_HARDWARE,
             .size = sizeof(struct perf_event_attr),
-            .config = PERF_COUNT_HW_INSTRUCTIONS,
+            .config = config,
             .disabled = 1,
             .exclude_kernel = 1,
             .exclude_hv = 1,
             .exclude_idle = 1,
-            }, 0, -1, -1, 0);
```
avikivity, line 94: You're replacing an easy-to-read list of settings with a function that takes a large number of anonymous int parameters with lots of defaults, why?

bhalevy: Because it's practically common code.
Other than the `config` field, all other members are the same,
so maintaining 2 open-coded copies require remembering to sync the changes (if needed).

Also, why do you consider specifying `pid_t pid = 0, int cpu = -1, int group_fd = -1, unsigned long flags = 0` in the interface as `anonymous int parameters` while today we pass `0, -1, -1, 0` which is completely cryptic?

avikivity: You're right about pid/cpu/group_fd. I was going to say that the commonality is accidental and the parameters can change independently, but something like disabled/exclude_kernel/exclude_hv isn't so likely to change.

PR #2296, opened 2024-06-17
timer-set: Add missing sanity headers

xemul:
The checkheaders target somehow stepped on recently merged #2269, while the PR itself passed CI when it was created

```
[...]

PR #2297, opened 2024-06-17
Simplify fs-info getting in make_file_impl()

xemul:
The helper looks up in the map of fs-info-s twice and returns the chain of ready futures on the common paths. This PR makes things simpler.

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -1132,11 +1132,11 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
         });
     }
 
-        const internal::fs_info& fsi = s_fstype[st_dev];
-        if (!fsi.append_challenged || options.append_is_unlikely || ((flags & O_ACCMODE) == O_RDONLY)) {
-            return make_ready_future<shared_ptr<file_impl>>(make_shared<posix_file_real_impl>(fd, open_flags(flags), std::move(options), fsi, st_dev));
-        }
-        return make_ready_future<shared_ptr<file_impl>>(make_shared<append_challenged_posix_file_impl>(fd, open_flags(flags), std::move(options), fsi, st_dev));
+    const internal::fs_info& fsi = s_fstype[st_dev];
+    if (!fsi.append_challenged || options.append_is_unlikely || ((flags & O_ACCMODE) == O_RDONLY)) {
+        return make_ready_future<shared_ptr<file_impl>>(make_shared<posix_file_real_impl>(fd, open_flags(flags), std::move(options), fsi, st_dev));
+    }
+    return make_ready_future<shared_ptr<file_impl>>(make_shared<append_challenged_posix_file_impl>(fd, open_flags(flags), std::move(options), fsi, st_dev));
 }
 
 file::file(seastar::file_handle&& handle) noexcept
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -1078,8 +1078,8 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
     auto st_dev = st.st_dev;
     static thread_local std::unordered_map<decltype(st_dev), internal::fs_info> s_fstype;
 
-    future<> get_fs_info = s_fstype.count(st_dev) ? make_ready_future<>() :
-        engine().fstatfs(fd).then([fd, st_dev] (struct statfs sfs) {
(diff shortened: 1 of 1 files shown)


xemul:
CI fails due to #2296

src/core/file.cc
```diff
@@ -1078,8 +1078,8 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
     auto st_dev = st.st_dev;
     static thread_local std::unordered_map<decltype(st_dev), internal::fs_info> s_fstype;
 
-    future<> get_fs_info = s_fstype.count(st_dev) ? make_ready_future<>() :
-        engine().fstatfs(fd).then([fd, st_dev] (struct statfs sfs) {
+    if (s_fstype.count(st_dev) == 0) {
```
tchaikov: could also use 
```c++
if (!s_fstype.contains(st_dev)) {
```

since we dropped the support of C++17.

src/core/file.cc
```diff
@@ -1078,7 +1078,8 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
     auto st_dev = st.st_dev;
     static thread_local std::unordered_map<decltype(st_dev), internal::fs_info> s_fstype;
 
-    if (s_fstype.count(st_dev) == 0) {
+    auto i = s_fstype.find(st_dev);
+    if (__builtin_expect(i == s_fstype.end(), false)) {
```
tchaikov, line 1082: how about using `[[unlikely]]`, since we have the luxury of using C++20.

src/core/file.cc
xemul: Yeah, this patch keeps finding method as is, next patch fixes it

xemul:
upd:
- use [[unlikely]]
- rebased to make CI pass

xemul:
upd:
- spelling

tchaikov (approved):
lgtm

xemul:

src/core/file.cc
```diff
@@ -1126,15 +1126,17 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
                 fsi.fsync_is_exclusive = true;
                 fsi.nowait_works = false;
             }
+            auto st_dev = st.st_dev;
             s_fstype[st_dev] = std::move(fsi);
+            return make_file_impl(fd, std::move(options), flags, std::move(st));
```
nyh, line 1131: I tried, but failed, to understand this change, maybe you can help me understand it...
You have a cache s_fstype of fstatfs results. Before this patch, the complicated code below, using the result of the fstatfs to decide how to open the file, was used regardless of whether the fsstatfs results were previously cached or not. After this patch, unless I'm misreading it something (it's really hard to understand the indentation in github), it seems that when the information isn't cached, you cache it, but then don't use what you cached and unconditionlay use the non-append-challenged variant. Why?

xemul: > Before this patch, the complicated code below, using the result of the fstatfs to decide how to open the file, was used regardless of whether the fsstatfs results were previously cached or not.

No, if the cached result is there it doesn't call fstat but gets the make_ready_future<>() instantly

```c++
    future<> get_fs_info = s_fstype.count(st_dev) ? make_ready_future<>() :
        engine().fstatfs(fd).then([fd, st_dev] (struct statfs sfs) {
            ...
            s_fstype[st_dev] = std::move(fsi);
        });
```

> After this patch, unless I'm misreading it something (it's really hard to understand the indentation in github), it seems that when the information isn't cached, you cache it, but then don't use what you cached and unconditionlay use the non-append-challenged variant.

I do exactly the same, but without make_ready_future<>. If the result is not there I do

```c++
    if (s_fstype.count(st_dev) == 0) {
        return engine().fstatfs(fd).then([fd, options = std::move(options), flags, st = std::move(st)] (struct statfs sfs) {
            ...
            return ... // "recursive" call to the current function
        });
```

and if the result is not there it just returns the future with file impl _without_ .then()-ing it to the make_ready_future<>()

```c++
    return make_ready_future<shared_ptr<file_impl>>(...);
```

nyh: Oh, I missed the **recursive** call to the current function. I misread it as a call to another function.

nyh: Nitpick (not important): I don't think moving "struct stat" actually does anything.

nyh (approved):

PR #2299, opened 2024-06-18
Unfriend reactor from message queues

xemul:
The alien and smp ones need _sleeping and _stopped. Both can be patched to be part of the public API.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -368,6 +368,8 @@ public:
     /// Register a user-defined signal handler
     void handle_signal(int signo, noncopyable_function<void ()>&& handler);
     void wakeup();
+    /// @private
+    bool stopped() const noexcept { return _stopped; }
 
 private:
     class signals {
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -97,7 +97,6 @@ namespace seastar {
 using shard_id = unsigned;
 
 namespace alien {
-class message_queue;
 class instance;
 }
 SEASTAR_MODULE_EXPORT
@@ -632,7 +631,6 @@ private:
 
     future<> run_exit_tasks();
     void stop();
-    friend class alien::message_queue;
     friend class pollable_fd;
     friend class pollable_fd_state;
     friend class posix_file_impl;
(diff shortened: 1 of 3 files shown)


xemul:
CI -- #2296

avikivity:
> The alien and smp ones need _sleeping and _stopped. Both can be patched to be part of the public API.

Why should they be part of the public API? We don't want people to rely on them without a good reason.

xemul:
The `_sleeping` doesn't become such, I just move it to existing `.wakeup()` one. I would be better to hide the latter in internal API indeed, but that's separate task. 

The `_stopped` ... yes, it appears to be a new method, but it's `const` and lightweight. So maybe it's OK until we hide it together with the `wakeup()`?

xemul:
upd:
- marked the new `reactor::stopped()` with `/// @private` not to appear in docs
- rebased to pick up CI fix

xemul:
> The `_sleeping` doesn't become such, I just move it to existing `.wakeup()` one. I would be better to hide the latter in internal API indeed, but that's separate task.
> 
> The `_stopped` ... yes, it appears to be a new method, but it's `const` and lightweight. So maybe it's OK until we hide it together with the `wakeup()`?

@avikivity , does this justify the change from you perspective?

xemul:
@avikivity , reminder ping

avikivity:
> upd:
> 
> * marked the new `reactor::stopped()` with `/// @private` not to appear in docs
> * rebased to pick up CI fix

I don't understand, you want stopped() to be @private, but not private:?

Reducing friendships is good, but it should cause us to break data hiding. If we want two classes to talk via a private channel, they can be friends.

xemul:
> I don't understand, you want stopped() to be @Private, but not private:?

Yes. We need some way to tell that a specific method is public: but not intended to be used by any code outside of seastar itself. The @private marking looks like good enough approach.

> Reducing friendships is good, but it should cause us to break data hiding. If we want two classes to talk via a private  channel, they can be friends.

I agree, but friendship is pretty heavy hammer for reactor which is overloaded with functionality.

avikivity:
I'm not totally happy about it, but we're splitting hairs, so okay.

In the past I sued friend free functions in the internal namespace.

Issue #2302, opened 2024-06-25
Aborting socket::connect() doesn't work with iouring backend


xemul:
There's a new test #2301 that passes on linux-aio and fails on io-uring

tchaikov:
> There's a new test #2301 that passes on linux-aio and fails on io-uring

this reminds me the discussion in https://github.com/scylladb/seastar/pull/1633#discussion_r1182440883

ndragazis:
@xemul could you check if you can still reproduce it? It's been a year and it works fine on my laptop (6.11.0-26-generic kernel).

xemul:
Seems to work indeed.  Kernel 6.14.5-200.fc41.x86_64. It's worth blacklisting "too old" kernels for uring, but I can't tell for sure which version to use exactly.

PR #2304, opened 2024-06-26
Teach http::client::make_request() to timeout

xemul:
This PR adds the send_timeout parameter to client::make_request() method. When set, client will _not_ send request after the given time-point instead resolving with timeout_error. This parameter does _not_ affect the time client waits for the server to respond, it only makes sure that the request is sent within given time, and everything that happens after it can last as long as it want (we can add reply_timeout later if we need).

Now the justification for this timeout. The make_request() method gets request parameter by value and requires is to be fully set including query arguments, headers and body writer (or body itself). Before sending the request client may wait for a connection to become available -- it can be a newly established connection, or some preexisting connection that's now serving some other request. This wait may happen to be "too long" in some cases.

One of the example of what "too long" means is S3 requests signing. Signed requests include, among other headers, the one with date and time when the request was signed and server imposes 15 minutes timeout on it. If a request is signed but arrives later than that duration, it's rejected. 15 minutes seem to be long enough, but sometimes it's not.

A very rare, but still real-life case -- each request establishes new TCP connection because server actively closes the existing ones. TCP connect() may take more than 1 minute, so 15+ requests in the queue can be enough to invalidate the signature of the tail ones.

Another risky case is Scylla uploading sstables. It happens in 5Mb chunk and sending 5Mb body to real S3 sometimes takes seconds and increases the chance to get connection closed by server. With high parallelism it again may invalidate the signature of last requests in the queue.

This PR allows sender to bail out making request early enough to let caller re-sign request and re-send it again.

closes: #1845 (it was another attempt to overcome this, but this solution looks more generic)


xemul:
upd:
- fix (maybe, I cannot run this build locally) --with-modules compilation

nyh:
> A very rare, but still real-life case -- each request establishes new TCP connection because server actively closes the existing ones. TCP connect() may take more than 1 minute, so 15+ requests in the queue can be enough to invalidate the signature of the tail ones.

I don't understand this scenario, at all all :-( First, why would TCP connect() take "more than 1 minute"? This is not a "real life case". Second, if you need to make 15 requests why would you do them one after another instead of in parallel? I remember when NCSA Mosaic introduced the feature of sending multiple requests (I think it was 4?) to the same server instead of sending them one by one. I think it was 1994 :-)

I think these hypothetical S3 examples are not helping convince reviewers like me that this (and similar) patches are worthwhile. First, because Seastar HTTP client is not an S3 client specificially. Second, because you aren't explaining why these hypothetical examples actually happen in S3 (or any other HTTP server).

I think a more convincing argument would be that TCP connect() is not inherently limited in time, so it can take a millisecond or retry for 24 hours - which is why every connect() implementation needs the ability to specify a timeout. So Seastar too needs a connect_timeout option too.

It makes sense for a connect_timeout to be separate than a total timeout, because the total timeout will usually be proportional to the amount of data and to the network's bandwidth (you can't expect downloading 1GB over a modem to take the same amount of time as downloading 1KB over LAN), but the connect_timeout is much less reliant on these things and can usually be set to a constant.


> Another risky case is Scylla uploading sstables. It happens in 5Mb chunk and sending 5Mb body to real S3 sometimes takes seconds and increases the chance to get connection closed by server. With high parallelism it again may invalidate the signature of last requests in the queue.

Again, the example doesn't make sense to me. If the caller has "high parallelism", they expect all these requests to be done in parallel, not be queued. If complete requests are queued for more than 15 minutes, where do we get the memory for this huge queue?

Anyway, as I said above, I agree that a "connection_timeout" option for a function making an HTTP request makes a lot of sense. I'm just not convinced by the specific examples :-)

include/seastar/http/client.hh
```diff
@@ -169,6 +170,7 @@ class client {
 public:
     using reply_handler = noncopyable_function<future<>(const reply&, input_stream<char>&& body)>;
     using retry_requests = bool_class<struct retry_requests_tag>;
+    using clock_type = lowres_clock;
```
nyh, line 173: I don't like this name "clock_type". It's very unclear what this clock is used for. Future developers may start using it for the wrong things (e.g., if you want to measure a latency histogram, you may not want to use lowres_clock to do this).
I think you should name it "timeout_clock" instead.

src/http/client.cc
```diff
@@ -262,9 +270,17 @@ future<client::connection_ptr> client::get_connection(clock_type::time_point tim
...
+    return _new_connections->make().then([this, timeout, cr = internal::client_ref(this)] (connected_socket cs) mutable {
         http_log.trace("created new http connection {}", cs.local_address());
         auto con = seastar::make_shared<connection>(std::move(cs), std::move(cr));
+        if (clock_type::now() > timeout) {
+            // Factory made new connection, though it's too late already. Don't throw
+            // this effort out, but don't use it to serve current request either.
+            return put_connection(std::move(con)).then_wrapped([] (auto f) {
+                f.ignore_ready_future();
+                return make_exception_future<connection_ptr>(httpd::timeout_error());
+            });
```
nyh, line 282: I don't understand why we need this super-rare special case. If you set a 1 minute connection timeout, and the connection was made in 59.999 seconds but then when you want to start sending it's already 60.001 seconds - why do we need to treat it specially? Remember that if we start sending in 59.999, then the timeout is no longer relevant - we could be sending 1 TB of data and the whole thing will succeed.

So I would simply remove this piece of code. If the make_connection() said it finished by the timeout, that's fine - we don't need to second-guess its decision here.

xemul:
> First, why would TCP connect() take "more than 1 minute"? 

IIRC with default Linux sysctls SYN-s are sent with 4x linear timeouts of 1s plus 6x exponential timeouts which gives 4 + 64s = 68s to connect.

> Second, if you need to make 15 requests why would you do them one after another instead of in parallel?

Yes, you're right, we can do them in parallel. I wasn't very clear, let me rephrase.

Client has "max_connections" parameter which user can configure, but it's not guaranteed that the real parallelism will be less-or-equal to that number. I think that's the key point here.

E.g. if max connections is, say 10, and client is asked to do 20 requests, then 10 will be executing and 10 will be in the queue. So my point is -- _if_ it happens that we have enough requests in the queue _and_ those running in parallel are slow enough, then the last one times out. 15 requests in queue and the remaining doing slow connect is just an example. Another example was below -- some requests in queue and the remaining ones doing slow read from S3.

> I think a more convincing argument would be that TCP connect() is not inherently limited in time, so it can take a millisecond or retry for 24 hours - which is why every connect() implementation needs the ability to specify a timeout.

Erm... Default Linux sysctl configuration doesn't let connect() hang without end. What do you mean by "is not inherently limited in time"?

> If the caller has "high parallelism", they expect all these requests to be done in parallel, not be queued.

Well, it _might_ expect that, but given clients connection pool is limited, it's not guaranteed that _all_ the requests run in parallel. Some will be queued waiting for any connection to become free.

> If complete requests are queued for more than 15 minutes, where do we get the memory for this huge queue?

It's not that much memory. E.g. if you do a lot of GET-s to serve sstables read from S3 storage, then each GET request is IIRC ~100 bytes, so even 100 requests in the queue is just 10k.

tests/unit/httpd_test.cc
```diff
@@ -908,10 +914,7 @@ SEASTAR_TEST_CASE(test_client_response_eof) {
 
         future<> client = seastar::async([&lcf] {
             auto cln = http::experimental::client(std::make_unique<loopback_http_factory>(lcf));
-            auto req = http::request::make("GET", "test", "/test");
-            BOOST_REQUIRE_EXCEPTION(cln.make_request(std::move(req), [] (const http::reply& rep, input_stream<char>&& in) {
-                return make_exception_future<>(std::runtime_error("Shouldn't happen"));
-            }, http::reply::status_type::ok).get(), std::system_error, [] (auto& ex) {
+            BOOST_REQUIRE_EXCEPTION(make_failing_http_request(cln).get(), std::system_error, [] (auto& ex) {
```
nyh, line 917: I don't understand what I'm seeing here. As far as I know, BOOST_REQUIRE_EXCEPTION already validates that the specific exception type you specified (std::system_error) is the one thrown - and fails the test if the function doesn't throw at all. It looks to me like this whole "shouldn't happen" thing isn't needed at all, and your new function make_failing_http_request shouldn't be needed either.
What am I missing?

tests/unit/httpd_test.cc
```diff
@@ -954,6 +954,74 @@ SEASTAR_TEST_CASE(test_client_response_parse_error) {
...
+                sleep(std::chrono::seconds(1)).get();
+                output_stream<char> out = sk.output();
+                out.close().get();
+            });
+        });
+
+        future<> client = seastar::async([&lcf, &server_started] {
+            auto cln = http::experimental::client(std::make_unique<loopback_http_factory>(lcf), 1 /* max connections */);
+            // this request gets handled by server and ...
+            auto f1 = make_failing_http_request(cln);
```
nyh, line 976: I don't understand why you want to make a *failing* request? What is this trying to do?

tests/unit/httpd_test.cc
```diff
@@ -954,6 +954,74 @@ SEASTAR_TEST_CASE(test_client_response_parse_error) {
...
+    return seastar::async([] {
+        loopback_connection_factory lcf(1);
+        auto ss = lcf.get_server_socket();
+        promise<> server_started;
+        future<> server = ss.accept().then([&] (accept_result ar) {
+            server_started.set_value();
+            return seastar::async([sk = std::move(ar.connection)] () mutable {
+                input_stream<char> in = sk.input();
+                read_simple_http_request(in);
+                sleep(std::chrono::seconds(1)).get();
```
nyh, line 967: Nitpick: Although "1 second should be enough for everyone" we always regret this later... Can this be some sort of condition variable or something, to tell this server to just hang forever until released and not for 1 second?

tests/unit/httpd_test.cc
```diff
@@ -954,6 +954,74 @@ SEASTAR_TEST_CASE(test_client_response_parse_error) {
...
+
+        future<> client = seastar::async([&lcf, &server_started] {
+            auto cln = http::experimental::client(std::make_unique<loopback_http_factory>(lcf), 1 /* max connections */);
+            // this request gets handled by server and ...
+            auto f1 = make_failing_http_request(cln);
+            server_started.get_future().get();
+            // ... this should hang waiting for cached connection
+            auto f2 = cln.make_request(http::request::make("GET", "test", "/test"), [] (const auto& rep, auto&& in) {
+                return make_exception_future<>(std::runtime_error("Shouldn't happen"));
+            }, http::reply::status_type::ok, http::experimental::client::clock_type::now() + std::chrono::milliseconds(100));
```
nyh, line 981: Again, you are assuming that if the server hangs the previous request for 1 second, then a timeout of 100ms will not be enough to make a new connection. But this is the stuff flaky tests are made of. Would be better to use a condition variable instead of these numbers, or if it's not possible, I think we need this to be more than 1 second sleep.

tests/unit/httpd_test.cc
```diff
@@ -954,6 +954,74 @@ SEASTAR_TEST_CASE(test_client_response_parse_error) {
...
+        });
+
+        future<> client = seastar::async([&lcf, &server_started] {
+            auto cln = http::experimental::client(std::make_unique<loopback_http_factory>(lcf), 1 /* max connections */);
+            // this request gets handled by server and ...
+            auto f1 = make_failing_http_request(cln);
+            server_started.get_future().get();
+            // ... this should hang waiting for cached connection
+            auto f2 = cln.make_request(http::request::make("GET", "test", "/test"), [] (const auto& rep, auto&& in) {
+                return make_exception_future<>(std::runtime_error("Shouldn't happen"));
```
nyh, line 980: Again, I don't understand why this "shouldn't happen" thing is needed. BOOST_REQUIRE_EXCEPTION already requires an exception, and if there is no exception, it will fail. So you don't need to replace the no exception by a silly "shouldn't happen" exception.

tests/unit/httpd_test.cc
```diff
@@ -954,6 +954,74 @@ SEASTAR_TEST_CASE(test_client_response_parse_error) {
     });
 }
 
+SEASTAR_TEST_CASE(test_client_request_send_timeout_cached_conn) {
+    return seastar::async([] {
+        loopback_connection_factory lcf(1);
```
nyh, line 959: It took me a while to understand why this test is about a "cached connection".
Now I understand: The difference between this test and the next test is that this test waits for the existing requests to finish and release the concurrency limit - i.e., it waits on a condvar - while the next test really starts a TCP connect() and waits on that.

tests/unit/httpd_test.cc
```diff
@@ -954,6 +954,74 @@ SEASTAR_TEST_CASE(test_client_response_parse_error) {
...
+        auto ss = lcf.get_server_socket();
+        future<> server = ss.accept().discard_result();
+
+        class delayed_factory : public loopback_http_factory {
+        public:
+            explicit delayed_factory(loopback_connection_factory& f) : loopback_http_factory(f) {}
+            virtual future<connected_socket> make() override {
+                return sleep(std::chrono::seconds(1)).then([this] {
+                    return loopback_http_factory::make();
```
nyh, line 1006: If I understand correctly, what you're doing here isn't causing the TCP connect to timeout, but for the connection factor to add an artificial delay. So you are NOT testing the ability of this code to timeout on a very long connect() attempt, but rather testing the extra "check now() after the factory returned" which I said in the previous test you should remove.
So I think this test is wrong and you are not testing the really important feature - the ability to break a TCP connect().

nyh (changes requested):
In general, I'm in favor of the feature you are adding, but I have many comments about the implementation that I think needs to be fixed.

include/seastar/http/client.hh
```diff
@@ -275,12 +277,13 @@ public:
      * \param req -- request to be sent
      * \param handle -- the response handler
      * \param expected -- the optional expected reply status code, default is std::nullopt
+     * \param send_timeout -- time point at which make_request will stop waiting for transport
```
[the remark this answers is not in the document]
xemul: I'm fine to rename it, but I'd like to emphasize that connect timeout doesn't fit well either. As you noticed below, this timeout also applies to the cond. variable wait() which waits for existing connection (from pool) to become available, and it doesn't imply connecting by any means. From the code perspective, this timeout separates the time client spends obtaining a connection (new one or from poll) and the process of sending request head. So this is sort of "request head sending timeout", but I've no better idea for its name that that.

src/http/client.cc
```diff
@@ -252,15 +252,15 @@ future<client::connection_ptr> client::get_connection() {
     }
 
     if (_nr_connections >= _max_connections) {
-        return _wait_con.wait().then([this] {
-            return get_connection();
+        return _wait_con.wait().then([this, timeout] {
```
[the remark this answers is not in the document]
xemul: I disagree with "with no hint" credit :D here's the phrase from this commit description:

> Next patch will make the code obey the provided timeout.

include/seastar/http/exception.hh
```diff
@@ -158,6 +158,17 @@ public:
     {}
 };
 
+/**
+ * Client-side exception to report that making request timed out
+ * prior to communicating to server
+ */
+class timeout_error : public base_exception {
```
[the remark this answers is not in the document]
xemul: My bad, I looked at the last exception in this header which is `unex pected_status_error` and named the new class with _error suffix. Will rename.

include/seastar/http/client.hh
xemul: Fair point. I followed the rpc_clock_type "naming convention" for no real reason

src/http/client.cc
xemul: Factory doesn't know this deadline and it can generate new connection in much more than 59.999 seconds easily and there are two reasons for that.

First -- the call to make new connection can happen after client tried serving request over polled one (#1883). E.g. -- timeout is 1 minute, client waits 30 seconds for pooled connection, then starts sending data over it and in 15 seconds server closes its end. Client asks for a new connection and it takes 1 minute to establish it. Boom -- 105 seconds had passed.

Second --  we may equip factory::make() with timeout parameter too, but connect() is not always abortable nowadays (#2302). It's a BUG, but still http client has means to handle it.

tests/unit/httpd_test.cc
xemul: The make request accepts lambda that returns a future. I do need to return some future from it.
Also note, that BOOST_REQUIRE_EXCEPTION wants system_error with ECONNABORTED code, while "shouldn't happen" exception is of runtime_error type, they are not related at all. The latter, as I wrote, is just because some future should be returned from there.

tests/unit/httpd_test.cc
xemul: Because server in this test doesn't really send any responses. This make_failing_request is here just to make client create a connection. Server then delays responding into it (and delays closing its end) so the connection is in pool, but is not available for the 2nd request. Thus the 2nd request starts waiting for it on the client's condition variable, times out and test happily notices it.

tests/unit/httpd_test.cc
xemul: Yes, good point. Maybe in this place we can do it.

tests/unit/httpd_test.cc
xemul: Client doesn't attempt to make new connection, the max_connections here is 1, so it's waiting for the previous one to return to pool. I think you're right and we can go with explicit server kick.

tests/unit/httpd_test.cc
xemul: It's just because I need to return some future from this lambda. It could as well be make_ready_future<>(). On the other hand, if something goes wrong and server sends the response and this place gets erroneously called ... I don't know, exceptional future from here is a fuse against silly mistakes in the future. Methinks

tests/unit/httpd_test.cc
xemul: Yes

tests/unit/httpd_test.cc
xemul: Yes, that's correct, but this patch is not about breaking TCP connect() , because it's not always working (#2302), and client doesn't mess with it at all, it uses external factory to generate new connections.

xemul: ... and tests here use test-only loopback sockets which are not TCP, but are in-memory shared queue behind connected_socket API, they don't have connect() timeout/abort. Even if they had, it wouldn't be TCP connect()

xemul: While we're here, @nyh , I invite you to review a test that checks the ability to break TCP connect() -- #2301

include/seastar/http/client.hh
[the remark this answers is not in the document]
nyh: It is a timeout waiting for a connection - not necessarily a **new** connection!

This distinction can be explained in text. But the name "send_timeout" suggests it's a timeout until the entire request is sent. This is not what you are doing. Moreover, if the request is very large (e.g., uploading a file), it is very far from what you're doing.

src/http/client.cc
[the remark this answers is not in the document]
nyh: I didn't understand this. I thought I saw some of the things already obeying the timeout.

tests/unit/httpd_test.cc
nyh: Yes, I know the "shouldn't happen" works because it returns the wrong type of exception so causes the test to fail. But I said that **no exception** would also make the test fail.
I don't know about the future thing and how to make this least ugly. But I still feel that boilerplate code which creates weird exceptions can't be the simplest way to do it.

tests/unit/httpd_test.cc
nyh: @xemul I thought the whole purpose of this patch is to allow timing out connect() (either an actual Linux connect() or something from a pool). I'm starting to understand this was **not** the point of this PR.

src/http/client.cc
nyh: Ok, I'm starting to think that this specific if() that I didn't like is actually the **main part** of your patch. Apparently, you really care if the connection attempt takes more than the given amount of time (and you even said in reponses to my questions, but not in the cover letter, that you can't control the timeout in a real system connect()). All you care is that if the connection attempt takes more than the given timeout - the request will not be sent because it is already considered stale.

So this patch is apparently about not sending stale requests, more about the ability to time out. I now understand that you consider it fine to wait 10 minutes instead of the 1 minute timeout, as long as after that 10 minutes we don't send the request.

include/seastar/http/client.hh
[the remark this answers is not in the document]
xemul: > It is a timeout waiting for a connection - not necessarily a new connection!

:rofl: that's 100% perfect explanation, all the more so the object that client waits for is indeed `class connection`. I'll rename

src/http/client.cc
xemul: Yes. The thing is that client doesn't establishes new connections itself, it calls connection_factory::make() and doesn't have any control over it. Of course, we can equip connection_factory::make() with timeout argument and assume that it doesn't ignore it (the whole client is still experimental, so we're relatively safe in changing the API like that), but should client fulfill its API contract and not send a request even if factory didn't fulfill its and missed the timeout?

tests/unit/httpd_test.cc
xemul: > Yes, I know the "shouldn't happen" works because it returns the wrong type of exception so causes the test to fail. But I said that no exception would also make the test fail.

The test passes. This indirectly proves that this "shouldn't happen" is indeed never called, so it can be assert(false) or return make_ready_future<>() as well.

xemul: Maybe I'll pass default-initialized noncopyable_function there. If called would just segfault , but since it's not called it would work

PR #2305, opened 2024-06-27
stall-analyser: gracefully handle empty input

bhalevy:
Currently, the tools prints an obscure error
if the input is empty:
```
Traceback (most recent call last):
  File "/home/bhalevy/dev/scylla_s3_reloc_server/./seastar//scripts/stall-analyser.py", line 363, in <module>
    print_stats(tally, tmin)
                       ^^^^
NameError: name 'tmin' is not defined. Did you mean: 'min'?
```

Instead, print a meaningful error and direct the user to run stall-analyser.py --help for usage instructions.

Ref https://github.com/scylladb/scylla_s3_reloc_server/pull/103

### DIFF
--- a/scripts/stall-analyser.py
+++ b/scripts/stall-analyser.py
@@ -126,9 +126,15 @@ class Graph:
         # Each node in the tree contains:
         self.count = 0
         self.total = 0
-        self.nodes = {}
+        self.nodes = dict[str, Node]()
         self.tail = Node('')
         self.head = Node('')
+
+    def empty(self):
+        return not self.nodes
+
+    def __bool__(self):
+        return not self.empty()
 
     def process_trace(self, trace: list[str], t: int) -> None:
         # process each backtrace and insert it to the tree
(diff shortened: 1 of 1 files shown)


scripts/stall-analyser.py
```diff
@@ -368,11 +375,13 @@ def main():
...
         if address_threshold:
             trace = list(dropwhile(lambda addr: int(addr, 0) >= address_threshold, trace))
-        tmin = args.minimum or 0
         if t >= tmin:
             graph.process_trace(trace, t)
 
     try:
+        if not graph:
+            print("No input data found. Please run `stall-analyser.py --help` for usage instruction")
+            exit(0)
```
tchaikov, line 384: i'd suggest use `sys.exit()` instead. see https://docs.python.org/3/library/constants.html#constants-added-by-the-site-module . i am quoting the remarks in the document.

> They are useful for the interactive interpreter shell and should not be used in programs.

bhalevy: done

tchaikov (approved):
lgtm

bhalevy:
@scylladb/seastar-maint please merge

PR #2306, opened 2024-06-27
stall-analyser: use 0 for default value of --minimum

tchaikov:
this change is a cleanup. it:

- uses a better name for the `--minimum` option, `args.minimum`
  is way too general.
- uses the default value of 0 instead of `None`, simpler this way
- just references `arg.tmin`, simpler this way


xemul:
Presumably obsoleted by #2305 ?

bhalevy (approved):
lgtm

bhalevy:

PR #2313, opened 2024-06-29
reactor: fix crash during metrics gathering

brain5lug:
The crash is caused by statistics request while scheduling_group is removed. The tasks_pending metric gathering triggers reactor::pending_task_count() call. This call causes null pointer derefence as the task queue pointer has been reset in reactor::destroy_scheduling_group()

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2545,7 +2545,9 @@ uint64_t
 reactor::pending_task_count() const {
     uint64_t ret = 0;
     for (auto&& tq : _task_queues) {
-        ret += tq->_q.size();
+        if (tq) {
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -2545,7 +2545,9 @@ uint64_t
 reactor::pending_task_count() const {
     uint64_t ret = 0;
     for (auto&& tq : _task_queues) {
-        ret += tq->_q.size();
+        if(tq) {
```
avikivity, line 2548: Coding style: space after `if`.

brain5lug: @avikivity the space has been added, the branch is updated

PR #2314, opened 2024-06-30
API to get group information and to change file ownership

muthu90tech:
The scylladb codebase is bypassing seastar api to make syscalls directly, instead add seastar APIs for those syscall and use it in the transport/controller.cc in scylla
see issue: https://github.com/scylladb/scylladb/issues/17443

### DIFF
--- a/include/seastar/core/file.hh
+++ b/include/seastar/core/file.hh
@@ -61,6 +61,14 @@ struct directory_entry {
     std::optional<directory_entry_type> type;
 };
 
+/// Group details from the system group database
+struct group_details {
+    sstring group_name;
+    sstring group_passwd;
+    __gid_t group_id;
+    std::vector<sstring> group_members;
+};
+
 /// Filesystem object stat information
 struct stat_data {
     uint64_t  device_id;      // ID of device containing file
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -493,6 +493,8 @@ public:
     future<> touch_directory(std::string_view name, file_permissions permissions = file_permissions::default_dir_permissions) noexcept;
     future<std::optional<directory_entry_type>>  file_type(std::string_view name, follow_symlink = follow_symlink::yes) noexcept;
     future<stat_data> file_stat(std::string_view pathname, follow_symlink) noexcept;
+    future<> chown(std::string_view filepath, uid_t owner, gid_t group);
+    future<std::optional<struct group_details>> getgrnam(std::string_view name);
     future<uint64_t> file_size(std::string_view pathname) noexcept;
     future<bool> file_accessible(std::string_view pathname, access_flags flags) noexcept;
     future<bool> file_exists(std::string_view pathname) noexcept {
--- a/include/seastar/core/seastar.hh
+++ b/include/seastar/core/seastar.hh
@@ -359,6 +359,21 @@ using follow_symlink = bool_class<follow_symlink_tag>;
 /// with follow_symlink::yes, or for the link itself, with follow_symlink::no.
 future<stat_data> file_stat(std::string_view name, follow_symlink fs = follow_symlink::yes) noexcept;
 
+/// Wrapper around getgrnam_r.
+/// If the provided group name does not exist in the group database, this call will return an empty optional.
+/// If the provided group name exists in the group database, the optional returned will contain the struct group_details information. 
+/// When an unexpected error is thrown by the getgrnam_r libc call, this function throws std::system_error with std::error_code.
+/// \param groupname name of the group
+///
+/// \return optional struct group_details of the group identified by name. struct group_details has details of the group from the group database.
+future<std::optional<struct group_details>> getgrnam(std::string_view name);
+
+/// Change the owner and group of file. This is a wrapper around chown syscall.
+/// The function throws std::system_error, when the chown syscall fails.
+/// \param filepath
+/// \param owner
+/// \param group
+future<> chown(std::string_view filepath, uid_t owner, gid_t group);
 /// Return the size of a file.
 ///
 /// \param name name of the file to return the size
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -27,12 +27,14 @@ module;
 #include <cassert>
 #include <chrono>
 #include <cmath>
+#include <coroutine> 
 #include <exception>
 #include <filesystem>
 #include <fstream>
 #include <regex>
 #include <thread>
 
+#include <grp.h>
 #include <spawn.h>
 #include <sys/syscall.h>
 #include <sys/vfs.h>
@@ -2198,6 +2200,47 @@ future<int> reactor::waitpid(pid_t pid) {
 void reactor::kill(pid_t pid, int sig) {
     auto ret = wrap_syscall<int>(::kill(pid, sig));
     ret.throw_if_error();
+}
+
+future<std::optional<struct group_details>> reactor::getgrnam(std::string_view name) {
+    syscall_result_extra<std::optional<struct group_details>> sr = co_await _thread_pool->submit<syscall_result_extra<std::optional<struct group_details>>>(
+        [name = sstring(name)] {
+            struct group grp;
+            struct group *result;
+            memset(&grp, 0, sizeof(struct group));
+            char buf[1024];
+            errno = 0;
+            int ret = ::getgrnam_r(name.c_str(), &grp, buf, sizeof(buf), &result);
+            if (!result) {
+                return wrap_syscall(ret, std::optional<struct group_details>(std::nullopt));
+            }
+
+            group_details gd;
+            gd.group_name = sstring(grp.gr_name);
+            gd.group_passwd = sstring(grp.gr_passwd);
+            gd.group_id = grp.gr_gid;
+            for (char **members = grp.gr_mem; *members != nullptr; ++members) {
+                gd.group_members.emplace_back(sstring(*members));
+            }
+            return wrap_syscall(ret, std::optional<struct group_details>(gd));
+        });
+
+    if (sr.result != 0) {
+        throw std::system_error(sr.ec());
+    }
+
+    co_return sr.extra;
+}
+
+future<> reactor::chown(std::string_view filepath, uid_t owner, gid_t group) {
+    syscall_result<int> sr = co_await _thread_pool->submit<syscall_result<int>>(
+        [filepath = sstring(filepath), owner, group] {
+            int ret = ::chown(filepath.c_str(), owner, group);
+            return wrap_syscall(ret);
+        });
+
+    sr.throw_if_error();
+    co_return;
 }
 
 future<stat_data>
--- a/src/core/syscall_result.hh
+++ b/src/core/syscall_result.hh
@@ -55,7 +55,7 @@ struct syscall_result {
             throw_fs_exception(reason, fs::path(path1), fs::path(path2));
         }
     }
-protected:
+
     std::error_code ec() const {
         return std::error_code(error, std::system_category());
     }
--- a/src/util/file.cc
+++ b/src/util/file.cc
@@ -132,6 +132,14 @@ future<stat_data> file_stat(std::string_view name, follow_symlink follow) noexce
     return engine().file_stat(name, follow);
 }
 
+future<std::optional<struct group_details>> getgrnam(std::string_view name) {
+    return engine().getgrnam(name);
+}
+
+future<> chown(std::string_view filepath, uid_t owner, gid_t group) {
+    return engine().chown(filepath, owner, group);
+}
+
 future<uint64_t> file_size(std::string_view name) noexcept {
     return engine().file_size(name);
 }
--- a/tests/unit/libc_wrapper_test.cc
+++ b/tests/unit/libc_wrapper_test.cc
@@ -1,0 +1,44 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Copyright (C) 2024 ScyllaDB.
+ */
+#include <coroutine>
+#include <iostream>
+
+#include <seastar/core/coroutine.hh>
+#include <seastar/core/reactor.hh>
+#include <seastar/core/seastar.hh>
+
+#include <seastar/testing/test_case.hh>
+
+using namespace seastar;
+
+SEASTAR_TEST_CASE(getgrnam_group_name_does_not_exist_test) {
+    // A better approach would be to use a test setup and teardown to create a fake group
+    // with members in it and in the teardown delete the fake group.
+    std::optional<struct group_details> grp = co_await getgrnam("roo");
+    BOOST_REQUIRE(!grp.has_value());
+}
+
+SEASTAR_TEST_CASE(getgrnam_group_name_exists_test) {
+    std::optional<struct group_details> grp = co_await getgrnam("root");
+    BOOST_REQUIRE(grp.has_value());
+    BOOST_REQUIRE_EQUAL(grp.value().group_name.c_str(), "root");
+}
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -354,6 +354,9 @@ seastar_add_test (network_interface
 seastar_add_test (json_formatter
   SOURCES json_formatter_test.cc)
 
+seastar_add_test (libc_wrapper
+  SOURCES libc_wrapper_test.cc)
+
 seastar_add_test (locking
   SOURCES locking_test.cc)
 


src/core/reactor.cc
```diff
@@ -2218,6 +2219,45 @@ void reactor::kill(pid_t pid, int sig) {
...
+future<group_details> reactor::grp_detail(std::string_view groupname) noexcept {
+    return futurize_invoke([groupname, this] {
+        return _thread_pool->submit<syscall_result_extra<struct group>>([groupname = sstring(groupname)] {
+            struct group *gg = ::getgrnam(groupname.c_str());
+            if (!gg) {
+                return wrap_syscall(-1, *gg); // Simulating failure
+            }
+            return wrap_syscall(0, *gg);
+        }).then([groupname = sstring(groupname)] (syscall_result_extra<struct group> gr) {
+            gr.throw_fs_exception_if_error("group failed", groupname);
```
tchaikov, line 2231: why you are throwing an `filesystem_error` if `getgrnam()` fails or the matching entry is not found, and `groupname`  does not evaluate to a pathname? probably a `system_error` is better here?

and, this design implies an assumption: if an entry is not found, an exception is thrown with an unknown `errno`. IMO, this is not ideal:

- no matched entry should not be considered as an exception. it's one of the outcomes the caller should handle, and the caller should not handle it in the error handling path.
- throwing an exception with an unknown `errno` is not acceptable.

both of these arguments could be controversial. but i think the bottom line is to get back to the caller with the correct message.

src/core/reactor.cc
```diff
@@ -2218,6 +2219,45 @@ void reactor::kill(pid_t pid, int sig) {
     ret.throw_if_error();
 }
 
+future<group_details> reactor::grp_detail(std::string_view groupname) noexcept {
+    return futurize_invoke([groupname, this] {
+        return _thread_pool->submit<syscall_result_extra<struct group>>([groupname = sstring(groupname)] {
+            struct group *gg = ::getgrnam(groupname.c_str());
```
tchaikov, line 2225: please set `errno` to zero before calling this API. as we are going to check `errno` in `wrap_syscall()` if the retval is -1, see https://man7.org/linux/man-pages/man3/getgrnam.3.html#RETURN_VALUE

include/seastar/core/reactor.hh
```diff
@@ -502,6 +502,8 @@ public:
     future<> touch_directory(std::string_view name, file_permissions permissions = file_permissions::default_dir_permissions) noexcept;
     future<std::optional<directory_entry_type>>  file_type(std::string_view name, follow_symlink = follow_symlink::yes) noexcept;
     future<stat_data> file_stat(std::string_view pathname, follow_symlink) noexcept;
+    future<> change_ownership_of_file(std::string_view filepath, uint64_t groupid) noexcept;
```
tchaikov, line 505: i'd suggest change the name to `chown()` to be consistent with its cousin of `chmod()`.

tchaikov: and, in addition to the gid, how about exposing the uid to the caller?

src/core/reactor.cc
```diff
@@ -2218,6 +2219,45 @@ void reactor::kill(pid_t pid, int sig) {
...
+}
+
+future<> reactor::change_ownership_of_file(std::string_view filepath, uint64_t groupid) noexcept {
+    return futurize_invoke([filepath, groupid, this] {
+        return _thread_pool->submit<syscall_result<int>>([filepath = sstring(filepath), groupid] {
+            int chown_result = ::chown(filepath.c_str(), ::geteuid(), groupid);
+            return wrap_syscall(chown_result);
+        }).then([filepath = sstring(filepath), groupid] (syscall_result<int> sr) {
+            if (sr.result < 0) {
+                if (sr.error == EPERM) {
```
tchaikov, line 2250: IMO, we should leave it to the caller to throw the user friendly and length message. as the caller understand what is the best way to communicate with the user, seastar should focus on returning the accurate error. so,
```c++
throw_fs_exception_if_error("chown failed", filepath);
```
would suffice.

src/util/file.cc
```diff
@@ -132,6 +132,14 @@ future<stat_data> file_stat(std::string_view name, follow_symlink follow) noexce
     return engine().file_stat(name, follow);
 }
 
+future<group_details> grp_detail(std::string_view groupname) noexcept {
+    return engine().grp_detail(groupname);
+}
+
+future<> change_ownership_of_file(std::string_view filepath, uint64_t groupgid) noexcept {
```
tchaikov, line 139: s/filepath/pathname/ to be more consistent.

why not using `gid_t` for the type of gid?

i'd suggest rename `groupgid` to `group` , as the "group" in `groupgid` is redundant.

include/seastar/core/reactor.hh
muthu90tech: yea, that sounds good.

src/core/reactor.cc
muthu90tech: Agree, how about create a structure something like:
```
struct group_details_value {
    char *group_name;
    char *group_password;
    uint64_t group_gid;
    char **group_member;
};

struct group_details {
    int error;
    struct group_details_value details;
};
```
And populate errno values to group_details.error and can let the caller handle the values when non zero, if not the caller can look for fields in group_details.details ? Thanks

tchaikov: i'd suggest 

1. return an `optional<group_details>` 
2. reset `errno` before calling this API. this suggestion is proposed in another comment.

BTW, might want to rename `group_details` to `group` or `group_info`.

src/util/file.cc
```diff
@@ -129,7 +129,6 @@ future<uint64_t> fs_free(std::string_view name) noexcept {
 }
 
 future<stat_data> file_stat(std::string_view name, follow_symlink follow) noexcept {
-    std::cout<< "@@@file_stat reached" << std::endl;
```
tchaikov: also, please squash this commit into the previous one. we don't squash commits when merging a pull request, so that each commit is preserved.

include/seastar/core/reactor.hh
nyh: I agree, it's already bad enough that Seastar has its own versions of syscalls and users need to learn how to use them and how their parameters and behavior differ from the traditional Unix ones - let's not make it even more difficult by inventing new names for them. If this is a thin wrapper over chown(), let's call it chown.

include/seastar/core/reactor.hh
```diff
@@ -502,6 +502,8 @@ public:
...
+    future<group_details> grp_detail(std::string_view groupname) noexcept;
```
nyh, line 506: I guess this is getgrnam() so I would call it that name.
But I have to wonder - this isn't even a system call - this is just one of a bazillion libc functions, do we really need a version of each and every one of them in Seastar? 
Maybe at some point we should stop adding all these random functions and add a conventient API for running (via _thread_pool) *any* potentially-sleeping library function?

include/seastar/core/file.hh
```diff
@@ -100,6 +100,13 @@ struct file_open_options {
     static constexpr uint32_t min_extent_size_hint_alignment{128u << 10}; // 128KB
 };
 
+struct group_details {
+    char *group_name;
+    char *group_password;
+    uint64_t group_gid;
+    char **group_member;
+};
```
nyh, line 108: This is just a copy of "struct group" from <grp.h>, so why not use the original?

include/seastar/core/seastar.hh
```diff
@@ -359,6 +360,17 @@ using follow_symlink = bool_class<follow_symlink_tag>;
 /// with follow_symlink::yes, or for the link itself, with follow_symlink::no.
 future<stat_data> file_stat(std::string_view name, follow_symlink fs = follow_symlink::yes) noexcept;
 
+/// Returns details of the group from the group database.
+/// \param groupname name of the group
+///
+/// \return group_details of the group identified by name.
+future<group_details> grp_detail(std::string_view groupname) noexcept;
```
nyh, line 367: Why noexcept? getgrnam() definitely can have errors - something as benign like a group that can't be found, to IO error reading /etc/group.

include/seastar/core/seastar.hh
```diff
@@ -359,6 +360,17 @@ using follow_symlink = bool_class<follow_symlink_tag>;
...
+
+/// Change the owner and group of file. The owner is value computed after calling geteuid.
+/// \param filepath
+/// \param groupid
+///
+future<> change_ownership_of_file(std::string_view filepath, uint64_t groupid) noexcept;
```
nyh, line 373: I understand now that you decided to implement a subset of chown(), that can only change the group (and handles the uid strangely - why geteuid and not getting the existing group of the file?). In that case I guess calling this subset chown() is not a good idea so you can call it chgrp(). But why not implement the "regular" chown instead?

src/core/reactor.cc
nyh: Please use the thread-safe getgrnam_r() instead of getgrnam().

src/util/file.cc
nyh: Right. <grp.h> already has git_t and struct group (which getgrnam returns), no need to invent our own copies of the same thing.

src/core/reactor.cc
muthu90tech: I looked into using `getgrnam_r()` instead of `getgrname` , the docs for `getgrnam_r` says
```
On success, getgrnam_r() and getgrgid_r() return zero, and set *result to grp. If no matching group record was found, these functions return 0 and store NULL in *result. In case of error, an error number is returned, and NULL is stored in *result.
```
I was going to take the following approach here:

```
future<std::optional<struct group>> reactor::getgrnam(std::string_view name, char *buf, size_t buflen) noexcept {
    return futurize_invoke([name, buf, buflen, this] {
        return _thread_pool->submit<syscall_result_extra<std::optional<struct group>>>([name = sstring(name), buf, buflen] {
            struct group grp;
            struct group *result;
            int ret = ::getgrnam_r(name.c_str(), &grp, buf, buflen, &result);
            if (!result) {
                return wrap_syscall(ret, std::optional<struct group>(std::nullopt));
            }
            return wrap_syscall(ret, std::optional<struct group>(*result));
        }).then([name = sstring(name)] (syscall_result_extra<std::optional<struct group>> gr) {
            if (gr.result != 0) {
                throw std::system_error(gr.result, std::system_category());
            }
            return make_ready_future<std::optional<struct group>>(std::optional<struct group>(std::move(gr.extra)));
        });
    });
}
```

since I am returning an optional\<group\> to the user, to communicate the system error number to the caller, `throw std::system_error(gr.result, std::system_category())` that way the user can handle these errors on their end
```
ENOENT or ESRCH or EBADF or EPERM ...
```

Thoughts..?

tchaikov: couple things:

1. you missed https://github.com/scylladb/seastar/pull/2314#discussion_r1660393108
2. please check `ret` first
3. might want to s/gr/sr/ or just s/gr/ret/ as `group` is implied here
4. please throw `std::error_code(gr.error, std::system_category())`

tchaikov: and, please just send the updated pull request instead of asking for confirmation from reviewers.

include/seastar/core/file.hh
```diff
@@ -100,6 +100,13 @@ struct file_open_options {
     static constexpr uint32_t min_extent_size_hint_alignment{128u << 10}; // 128KB
 };
```
avikivity, line 101: > Add APIs to get group details and to change ownership of file. This w…

Please have just one sentence in the subject, use `git log` to see the style of our commitlogs.

include/seastar/core/file.hh
avikivity: It's actually bad to use the original (or a copy) since they're pointers into nowhere. We need a struct with values, not pointers of questionable lifetime.

avikivity: Plus documentation.

include/seastar/core/reactor.hh
avikivity: We have a tradition of using more readable names.

include/seastar/core/seastar.hh
avikivity: The error can/should be returned in the future rather than having two ways to throw.

src/core/reactor.cc
```diff
@@ -2218,6 +2219,45 @@ void reactor::kill(pid_t pid, int sig) {
     ret.throw_if_error();
 }
 
+future<group_details> reactor::grp_detail(std::string_view groupname) noexcept {
+    return futurize_invoke([groupname, this] {
```
avikivity, line 2223: Better to use coroutines. Most/all of the Seastar codebase doesn't use coroutines because until recently we supported C++17, but that's no longer the case and coroutines would be more readable here.

muthu90tech: For this issue, i was following the pattern already defined in the `src/core/reactor.cc`, could adopting coroutines be handled as a separate issue migrating all of the existing usage to the coroutines ?

muthu90tech: Hi @avikivity , does this patch `[scylladb-dev] [PATCH v1 00/11] co-routinize most of the CQL transport code` make this ticket: obsolete ? Thanks

src/core/reactor.cc
```diff
@@ -2218,6 +2219,41 @@ void reactor::kill(pid_t pid, int sig) {
...
 
+future<std::optional<struct group>> reactor::getgrnam(std::string_view name, char *buf, size_t buflen) {
+    return futurize_invoke([name, buf, buflen, this] {
+        return _thread_pool->submit<syscall_result_extra<std::optional<struct group>>>([name = sstring(name), buf, buflen] {
+            errno = 0;
+            struct group grp;
+            struct group *result;
+            memset(&grp, 0, sizeof(struct group));
+            int ret = ::getgrnam_r(name.c_str(), &grp, buf, buflen, &result);
+            if (!result) {
```
tchaikov, line 2230: quote from `getgrnam(3)`

> On  success, getgrnam_r() and getgrgid_r() return zero, and set *result to grp.  If no matching group record was found, these functions return 0 and store NULL in *result.

so i'd suggest check `ret` instead of `result` here.

src/core/reactor.cc
```diff
@@ -2218,6 +2219,41 @@ void reactor::kill(pid_t pid, int sig) {
...
+                return wrap_syscall(ret, std::optional<struct group>(std::nullopt));
+            }
+            return wrap_syscall(ret, std::optional<struct group>(*result));
```
tchaikov, line 2233: or, just 

```suggestion
            return wrap_syscall(ret, result ? std::optional<struct group>(*result) : std::nullopt);
```

src/core/reactor.cc
```diff
@@ -2218,6 +2219,41 @@ void reactor::kill(pid_t pid, int sig) {
...
+        }).then([] (syscall_result_extra<std::optional<struct group>> sr) {
+            if (sr.result != 0) {
+                // getgrnam_r, in case of error, an error number is returned (here it is stored in result), 
+                // which is the same as the error code in errno
```
tchaikov, line 2237: probably this comment is not necessary. as this follows the common pattern of other wrappers.

src/core/reactor.cc
```diff
@@ -2218,6 +2219,41 @@ void reactor::kill(pid_t pid, int sig) {
...
+                throw std::error_code(sr.result, std::system_category());
+            }
+            return make_ready_future<std::optional<struct group>>(std::optional<struct group>(std::move(sr.extra)));
```
tchaikov, line 2240:
```suggestion
            return make_ready_future<std::optional<struct group>>(std::move(sr.extra));
```

include/seastar/core/reactor.hh
avikivity: We have seastar::alien already.

src/core/reactor.cc
muthu90tech: Got it, I picked this issue as it was labeled as a `good first issue` , this was a good beginner experience for me to clone the repo and get around tools. I would like to get more familiar with `coroutines` and interested in working on porting this code to coroutines. Thank you

include/seastar/core/reactor.hh
muthu90tech: Looks like there are other libc functions, which will need to be cleaned up as well, can work on this as a followup.

src/core/reactor.cc
tchaikov: this comment is not resolved yet.

include/seastar/core/seastar.hh
```diff
@@ -359,6 +360,23 @@ using follow_symlink = bool_class<follow_symlink_tag>;
 /// with follow_symlink::yes, or for the link itself, with follow_symlink::no.
 future<stat_data> file_stat(std::string_view name, follow_symlink fs = follow_symlink::yes) noexcept;
 
+/// Returns details of the group from the group database. Wrapper around getgrnam_t.
+/// If the group is found, the std::optional contains the struct group information; otherwise, it is empty.
```
tchaikov, line 364: i'd suggest move this remark to the `\return` section.

include/seastar/core/seastar.hh
```diff
@@ -359,6 +360,23 @@ using follow_symlink = bool_class<follow_symlink_tag>;
...
+///
+/// \return optional group_info of the group identified by name.
+future<std::optional<struct group>> getgrnam(std::string_view name, char *buf, size_t buflen);
+
+/// Change the owner and group of file. This is a wrapper around chown syscall.
+/// The function throws std::system_error, when the chown syscall fails.
+/// \param filepath
+/// \param owner
+/// \param group
+///
```
tchaikov, line 378: might want to remove this empty line.

src/core/reactor.cc
```diff
@@ -2218,6 +2219,36 @@ void reactor::kill(pid_t pid, int sig) {
     ret.throw_if_error();
 }
 
+future<std::optional<struct group>> reactor::getgrnam(std::string_view name, char *buf, size_t buflen) {
+    return futurize_invoke([name, buf, buflen, this] {
```
tchaikov, line 2223: https://github.com/scylladb/seastar/pull/2314#discussion_r1665946683 is not addressed. could you reconsider it?

src/core/reactor.cc
```diff
@@ -2218,6 +2219,36 @@ void reactor::kill(pid_t pid, int sig) {
...
+        return _thread_pool->submit<syscall_result_extra<std::optional<struct group>>>([name = sstring(name), buf, buflen] {
+            errno = 0;
```
tchaikov, line 2225: nit, might want to move this line down right before `getgrnam_r()` call, like:
```
errno = 0;
int ret = ::getgrnam_r(name.c_str(), &grp, buf, buflen, &result);
```

src/core/reactor.cc
```diff
@@ -2218,6 +2219,36 @@ void reactor::kill(pid_t pid, int sig) {
...
+            }
+            return make_ready_future<std::optional<struct group>>(std::move(sr.extra));
+        });
+    });
+}
+
+future<> reactor::chown(std::string_view filepath, uid_t owner, gid_t group) {
+    return futurize_invoke([filepath, owner, group, this] {
+        return _thread_pool->submit<syscall_result<int>>([filepath = sstring(filepath), owner, group] {
+            int chown_result = ::chown(filepath.c_str(), owner, group);
```
tchaikov, line 2243:
```suggestion
            int ret = ::chown(filepath.c_str(), owner, group);
```

to be consistent with other places in this source file.

muthu90tech (commented):
test with c++ 20 failed , because of not including co_routine header file. and the struct group , I think should be addressed as struct ::group, addressing these now

src/util/file.cc
```diff
@@ -132,6 +132,14 @@ future<stat_data> file_stat(std::string_view name, follow_symlink follow) noexce
     return engine().file_stat(name, follow);
 }
 
+future<std::optional<struct group>> getgrnam(std::string_view name, char *buf, size_t buflen) {
```
tchaikov, line 135: please `#include <grp.h>` in this source file so that it has access the complete definition of `group`. you could add `#include <grp.h>` right before `#include <sys/statvfs.h>`.

src/core/reactor.cc
```diff
@@ -2218,6 +2219,35 @@ void reactor::kill(pid_t pid, int sig) {
     ret.throw_if_error();
 }
 
+future<std::optional<struct group>> reactor::getgrnam(std::string_view name, char *buf, size_t buflen) {
+    syscall_result_extra<std::optional<struct group>> sr = co_await _thread_pool->submit<syscall_result_extra<std::optional<struct group>>>(
```
tchaikov, line 2223: for using `co_await` operator, please `#include <coroutine>`. please put it right before `#include <exception>`. so that they are sorted in alphabetical order.

tchaikov (changes requested):
i think we are close. in addition to addressing the build failure with C++ modules, please fold the commits into a single one.

include/seastar/core/seastar.hh
```diff
@@ -47,6 +47,7 @@
 /// View the [Seastar compatibility statement](./md_compatibility.html) for
 /// information about library evolution.
 
+#include <grp.h>
```
tchaikov, line 50: we prefer not putting the 3rd-party headers in the purview when C++ modules are enabled. as the seastar C++module is not supposed to export the symbols provided by the 3rd-party headers . so could you please move this `#include` statement down into the `#ifndef SEASTAR_MODULE` block. you could put it after `#include <string_view>`, as `<string_view>` is a header from the standard library, and in `src/core/reactor.cc`, we put the 3rd party (the system library) headers after the headers from the standard library.

tchaikov (commented):
please remove the merge commit from the changeset.

include/seastar/core/seastar.hh
muthu90tech: Thanks for the explanation, when I added the <grp.h> within the `#ifndef SEASTAR_MODULE`, the compile using cxx modules was complaining about `struct group` being considered as `struct seastar::group`. I added a forward declaration `struct group` before the `namespace seastar` in this file and the compile using cxx modules passed.

tchaikov:
@muthu90tech please resolve the conflicts and repush.

muthu90tech:
> @muthu90tech please resolve the conflicts and repush.

rebased instead of merge commit.

include/seastar/core/seastar.hh
```diff
@@ -60,8 +60,11 @@
 #include <filesystem>
 #include <optional>
 #include <string_view>
+#include <grp.h>
```
tchaikov, line 63: we can drop this. as a full definition of the `group` is not necessary, and we already have its forward declaration in place. see the other comment.

include/seastar/core/seastar.hh
```diff
@@ -60,8 +60,11 @@
...
 #endif
 
+struct group;
```
tchaikov, line 66: this is not necessary.

the `group` struct is only used once in this header. in the following function declaration:
```c++
future<std::optional<struct group>> getgrnam(std::string_view name, char *buf, size_t buflen);
```

in this context, `struct group` serves as a forward declaration, making a separate forward declaration unnecessary.

but we do need to have 
```diff
diff --git a/src/seastar.cc b/src/seastar.cc
index ba3a1d44..6ffdb9b4 100644
--- a/src/seastar.cc
+++ b/src/seastar.cc
@@ -142,6 +142,7 @@ module;
 #include <sys/un.h>
 #include <execinfo.h>
 #include <fcntl.h>
+#include <grp.h>
 #include <malloc.h>
 #include <pthread.h>
 #include <setjmp.h>
```

otherwise when compiling the seastar C++ module, we'd have following build failure:
```
/home/kefu/.local/bin/clang++ -DFMT_SHARED -DSEASTAR_API_LEVEL=7 -DSEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_NUMA -DSEASTAR_HAVE_SYSTEMTAP_SDT -DSEASTAR_HAVE_URING -DSEASTAR_MODULE -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_SSTRING -Dseastar_module_EXPORTS -I/home/kefu/dev/seastar/include -I/home/kefu/dev/seastar/build/debug/gen/include -I/home/kefu/dev/seastar/src -g -std=c++23 -fPIC -U_FORTIFY_SOURCE -Wno-include-angled-in-module-purview -MD -MT src/CMakeFiles/seastar-module.dir/util/file.cc.o -MF src/CMakeFiles/seastar-module.dir/util/file.cc.o.d @src/CMakeFiles/seastar-module.dir/util/file.cc.o.modmap -o src/CMakeFiles/seastar-module.dir/util/file.cc.o -c /home/kefu/dev/seastar/src/util/file.cc
In file included from /home/kefu/dev/seastar/src/util/file.cc:28:
In file included from /usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/deque:62:
In file included from /usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/bits/stl_algobase.h:64:
In file included from /usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/bits/stl_pair.h:60:
/usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/type_traits:3413:7: error: incomplete type 'seastar::group' used in type trait expression
 3413 |     = __is_trivially_constructible(_Tp, __add_lval_ref_t<const _Tp>);
      |       ^
/usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/optional:506:12: note: in instantiation of variable template specialization 'std::is_trivially_copy_constructible_v<seastar::group>' requested here
  506 |            bool = is_trivially_copy_constructible_v<_Tp>,
      |                   ^
/usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/optional:704:15: note: in instantiation of default argument for '_Optional_base<seastar::group>' required here
  704 |     : private _Optional_base<_Tp>,
      |               ^~~~~~~~~~~~~~~~~~~
/usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/type_traits:3435:63: note: in instantiation of template class 'std::optional<seastar::group>' requested here
 3435 |   requires (!is_reference_v<_Tp>) && requires (_Tp& __t) { __t.~_Tp(); }
      |                                                               ^
/usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/type_traits:3435:60: note: in instantiation of requirement here
 3435 |   requires (!is_reference_v<_Tp>) && requires (_Tp& __t) { __t.~_Tp(); }
      |                                                            ^~~~~~~~~~
/usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/type_traits:3435:38: note: while substituting template arguments into constraint expression here
 3435 |   requires (!is_reference_v<_Tp>) && requires (_Tp& __t) { __t.~_Tp(); }
      |                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/kefu/dev/seastar/include/seastar/core/future.hh:354:15: note: (skipping 1 context in backtrace; use -ftemplate-backtrace-limit=0 to see all)
  354 |         (std::is_trivially_destructible_v<T> && std::is_trivially_constructible_v<T> &&
      |               ^
/home/kefu/dev/seastar/include/seastar/core/future.hh:354:15: note: during template argument deduction for variable template partial specialization 'is_trivially_destructible_v<_Tp>' [with _Tp = std::optional<seastar::group>]
/home/kefu/dev/seastar/include/seastar/core/future.hh:361:44: note: in instantiation of variable template specialization 'seastar::internal::can_inherit<std::optional<seastar::group>>' requested here
  361 |     : public uninitialized_wrapper_base<T, can_inherit<T>> {};
      |                                            ^
/home/kefu/dev/seastar/include/seastar/core/future.hh:571:58: note: in instantiation of template class 'seastar::internal::uninitialized_wrapper<std::optional<seastar::group>>' requested here
  571 | struct future_state :  public future_state_base, private internal::uninitialized_wrapper<T> {
      |                                                          ^
/home/kefu/dev/seastar/include/seastar/core/future.hh:1242:18: note: in instantiation of template class 'seastar::future_state<std::optional<seastar::group>>' requested here
 1242 |     future_state _state;
      |                  ^
/home/kefu/dev/seastar/src/util/file.cc:136:37: note: in instantiation of template class 'seastar::future<std::optional<seastar::group>>' requested here
  136 | future<std::optional<struct group>> getgrnam(std::string_view name, char *buf, size_t buflen) {
      |                                     ^
/home/kefu/dev/seastar/include/seastar/core/seastar.hh:373:29: note: forward declaration of 'seastar::group'
  373 | future<std::optional<struct group>> getgrnam(std::string_view name, char *buf, size_t buflen);
      |                             ^
In file included from /home/kefu/dev/seastar/src/util/file.cc:28:
```

tchaikov (changes requested):
we are closer =) 

left a couple comments.

include/seastar/core/seastar.hh
muthu90tech: Ok thanks, I noticed I also had to remove the `#include <grp.h>` from the `seastar/include/seastar/core/reactor.hh` or else the compile failure was 

```
...
...
In file included from /home/raj/db/scylla/seastar/include/seastar/core/reactor.hh:77:
/usr/include/grp.h:165:11: error: reference to 'group' is ambiguous
  165 |                         struct group *__restrict __resultbuf,
      |                                ^
/usr/include/grp.h:42:8: note: candidate found by name lookup is 'group'
   42 | struct group
      |        ^
/home/raj/db/scylla/seastar/include/seastar/core/seastar.hh:370:29: note: candidate found by name lookup is 'seastar::group'
  370 | future<std::optional<struct group>> getgrnam(std::string_view name, char *buf, size_t buflen);
      |                             ^
In file included from /home/raj/db/scylla/seastar/tests/perf/future_util_perf.cc:27:
In file included from /home/raj/db/scylla/seastar/include/seastar/coroutine/parallel_for_each.hh:30:
In file included from /home/raj/db/scylla/seastar/include/seastar/core/reactor.hh:77:
/usr/include/grp.h:167:11: error: reference to 'group' is ambiguous
  167 |                         struct group **__restrict __result)
      |                                ^
/usr/include/grp.h:42:8: note: candidate found by name lookup is 'group'
   42 | struct group
      |        ^
/home/raj/db/scylla/seastar/include/seastar/core/seastar.hh:370:29: note: candidate found by name lookup is 'seastar::group'
  370 | future<std::optional<struct group>> getgrnam(std::string_view name, char *buf, size_t buflen);
      |                             ^
13 errors generated.

```

tchaikov (approved):
lgtm

tchaikov:

src/core/reactor.cc
```diff
@@ -2200,6 +2202,35 @@ void reactor::kill(pid_t pid, int sig) {
...
+        [name = sstring(name), buf, buflen] {
```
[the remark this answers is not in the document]
muthu90tech: @nyh , I went with the copy approach, since most of the other function calls were also copying to new sstring.

And when I avoid copying and just use `name`, I see this compile error:

```
[1/5] Building CXX object CMakeFiles/seastar.dir/src/core/reactor.cc.o
FAILED: CMakeFiles/seastar.dir/src/core/reactor.cc.o 
/usr/local/bin/clang++ -DFMT_SHARED -DSEASTAR_API_LEVEL=7 -DSEASTAR_BUILD_SHARED_LIBS -DSEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT -DSEASTAR_DEPRECATED_OSTREAM_FORMATTERS -DSEASTAR_ENABLE_ALLOC_FAILURE_INJECTION -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_NUMA -DSEASTAR_HAVE_SYSTEMTAP_SDT -DSEASTAR_HAVE_URING -DSEASTAR_LOGGER_COMPILE_TIME_FMT -DSEASTAR_LOGGER_TYPE_STDOUT -DSEASTAR_PTHREAD_ATTR_SETAFFINITY_NP -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_SSTRING -DSEASTAR_STRERROR_R_CHAR_P -DSEASTAR_THREAD_STACK_GUARDS -DSEASTAR_TYPE_ERASE_MORE -Dseastar_EXPORTS -I/home/raj/db/scylla/seastar/include -I/home/raj/db/scylla/seastar/build/dev/gen/include -I/home/raj/db/scylla/seastar/build/dev/gen/src -I/home/raj/db/scylla/seastar/src -O2 -std=gnu++23 -fPIC -U_FORTIFY_SOURCE -Wno-error=unused-result "-Wno-error=#warnings" -fstack-clash-protection -UNDEBUG -Wall -Werror -Wimplicit-fallthrough -Wdeprecated -Wno-error=deprecated -ftls-model=initial-exec -gz -MD -MT CMakeFiles/seastar.dir/src/core/reactor.cc.o -MF CMakeFiles/seastar.dir/src/core/reactor.cc.o.d -o CMakeFiles/seastar.dir/src/core/reactor.cc.o -c /home/raj/db/scylla/seastar/src/core/reactor.cc
/home/raj/db/scylla/seastar/src/core/reactor.cc:2212:41: error: no member named 'c_str' in 'std::basic_string_view<char>'
 2212 |             int ret = ::getgrnam_r(name.c_str(), &grp, buf, buflen, &result);
      |                                    ~~~~ ^
In file included from /home/raj/db/scylla/seastar/src/core/reactor.cc:181:
/home/raj/db/scylla/seastar/src/core/thread_pool.hh:44:32: error: no matching member function for call to 'submit'
   44 |         return inter_thread_wq.submit<T>(std::move(func));
      |                ~~~~~~~~~~~~~~~~^~~~~~~~~
/home/raj/db/scylla/seastar/src/core/reactor.cc:2206:83: note: in instantiation of function template specialization 'seastar::thread_pool::submit<seastar::syscall_result_extra<std::optional<group>>, (lambda at /home/raj/db/scylla/seastar/src/core/reactor.cc:2207:9)>' requested here
 2206 |     syscall_result_extra<std::optional<struct group>> sr = co_await _thread_pool->submit<syscall_result_extra<std::optional<struct group>>>(
      |                                                                                   ^
/home/raj/db/scylla/seastar/src/core/syscall_work_queue.hh:65:15: note: candidate function template not viable: no known conversion from 'typename std::remove_reference<(lambda at /home/raj/db/scylla/seastar/src/core/reactor.cc:2207:9) &>::type' (aka '(lambda at /home/raj/db/scylla/seastar/src/core/reactor.cc:2207:9)') to 'noncopyable_function<syscall_result_extra<optional<group>> ()>' for 1st argument
   65 |     future<T> submit(noncopyable_function<T ()> func) noexcept {
      |               ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.

```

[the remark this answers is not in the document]
nyh: Ok, now I understand why these copies were done. std::string_view does not have a c_str() method because it is not guaranteed to be null-terminated. To get a null-terminated string, which we need for the C library function, the easiest thing is to copy to a std::string or seastar::sstring and use its c_str().
So your code is fine and I'm resolving this issue.

nyh:
@muthu90tech when you push a new version and ask for a review, please write a comment saying what you changed from the previous version.

src/core/syscall_result.hh
```diff
@@ -68,6 +68,10 @@ struct syscall_result_extra : public syscall_result<int> {
     Extra extra;
     syscall_result_extra(int result, int error, Extra e) : syscall_result<int>{result, error}, extra{std::move(e)} {
     }
+    
+    std::error_code error_code() const {
+        return this->ec();
+    }
```
nyh, line 74: Why do you need another alias for ec()? I don't know why ec() was made protected in a struct where absolutely everything else is public (this code comes from 02722823c0, six years ago), but if you just remove the "protected" it will become visible and you don't need a second name for it.

tests/unit/libc_wrapper_test.cc
```diff
@@ -0,0 +1,55 @@
...
+
+#include <seastar/testing/test_case.hh>
+
+using namespace seastar;
+
+SEASTAR_TEST_CASE(getgrnam_error_test) {
+    size_t buflen = 1;
+    char buf[1];
+    memset(&buf, 0, buflen);
+    BOOST_REQUIRE_THROW(co_await getgrnam("root", buf, buflen), std::system_error);
```
nyh, line 38: Nice, this is what I hoped for.
Is it possible to check not only that the error is a std::system_error, but also that its code is ERANGE?

tests/unit/libc_wrapper_test.cc
```diff
@@ -0,0 +1,55 @@
...
+#include <seastar/core/seastar.hh>
+
+#include <seastar/testing/test_case.hh>
+
+using namespace seastar;
+
+SEASTAR_TEST_CASE(getgrnam_error_test) {
+    size_t buflen = 1;
+    char buf[1];
+    memset(&buf, 0, buflen);
```
nyh, line 37: I don't think this memset is needed - I think getrnam() only writes to the buffer, it doesn't read it or expect it to be zeroed. But please correct me if I'm wrong.

tests/unit/libc_wrapper_test.cc
```diff
@@ -0,0 +1,55 @@
...
+#include <seastar/core/coroutine.hh>
+#include <seastar/core/reactor.hh>
+#include <seastar/core/seastar.hh>
+
+#include <seastar/testing/test_case.hh>
+
+using namespace seastar;
+
+SEASTAR_TEST_CASE(getgrnam_error_test) {
+    size_t buflen = 1;
```
nyh, line 35: nitpick: you can use `sizeof(buf)` below instead of having another variable.

tests/unit/libc_wrapper_test.cc
```diff
@@ -0,0 +1,55 @@
...
+    char buf[1];
+    memset(&buf, 0, buflen);
+    BOOST_REQUIRE_THROW(co_await getgrnam("root", buf, buflen), std::system_error);
+}
+
+SEASTAR_TEST_CASE(getgrnam_group_name_does_not_exist_test) {
+    size_t buflen = 1024;
+    char buf[1024];
+    memset(&buf, 0, buflen);
+    std::optional<struct group> grp = co_await getgrnam("roo", buf, buflen);
```
nyh, line 45: I hope nobody will run this test on Christopher Robin's computer :-)

tests/unit/libc_wrapper_test.cc
```diff
@@ -0,0 +1,55 @@
...
+    size_t buflen = 1;
+    char buf[1];
+    memset(&buf, 0, buflen);
+    BOOST_REQUIRE_THROW(co_await getgrnam("root", buf, buflen), std::system_error);
+}
+
+SEASTAR_TEST_CASE(getgrnam_group_name_does_not_exist_test) {
+    size_t buflen = 1024;
+    char buf[1024];
+    memset(&buf, 0, buflen);
```
nyh, line 44: Again, I think all the memset() calls in this file are unnecessary, but please correct me if I'm wrong.

nyh (changes requested):

tests/unit/libc_wrapper_test.cc
muthu90tech: :)

muthu90tech (commented):
changes made:
1) Made the ec() method public in seastar/src/core/syscall_result.hh > syscall_result
2) Modified the test, to assert on error code as well
3) the memset on buf not needed as getgrnam only writes to it.
4) removed unnecessary var to hold buffer length.

Commit Sep 15:
@nyh 
After reading comments from @avikivity , I have created a separate seastar specific group struct and named it group_details

include/seastar/core/seastar.hh
```diff
@@ -359,6 +359,22 @@ using follow_symlink = bool_class<follow_symlink_tag>;
...
 
+/// Wrapper around getgrnam_t.
+/// If the group is found, the std::optional contains the struct group information; otherwise, it is empty.
+/// The function throws std::error_code, when the getgrnam_t syscall fails.
+/// \param groupname name of the group
+/// \param buf buffer to store the string fields pointed to by the members of group structure.
+/// \param buflen size of the buffer
+///
+/// \return optional struct group of the group identified by name. struct group has details of the group from the group database.
+future<std::optional<struct group>> getgrnam(std::string_view name, char *buf, size_t buflen);
```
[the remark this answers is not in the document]
avikivity: I think if we do go to the effort of wrapping it, we can supply our own group struct.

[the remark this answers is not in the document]
avikivity: (using the unix getgrnam correctly appears very difficult at first shot)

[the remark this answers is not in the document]
muthu90tech: Thanks, I have made a seastar specific group struct and named it `group_details` also with the latest commit avoiding the user-supplied-buffer. @nyh , @avikivity

src/seastar.cc
```diff
@@ -142,6 +142,7 @@ module;
 #include <sys/un.h>
 #include <execinfo.h>
 #include <fcntl.h>
+#include <grp.h>
```
muthu90tech, line 145: should remove this

src/util/file.cc
```diff
@@ -30,6 +30,7 @@ module;
 #include <iostream>
 #include <list>
 #include <vector>
+#include <grp.h>
```
muthu90tech, line 33: should remove this

avikivity (approved):

src/core/reactor.cc
```diff
@@ -2200,6 +2202,48 @@ void reactor::kill(pid_t pid, int sig) {
...
+future<std::optional<struct group_details>> reactor::getgrnam(std::string_view name) {
+    syscall_result_extra<std::optional<struct group>> sr = co_await _thread_pool->submit<syscall_result_extra<std::optional<struct group>>>(
+        [name = sstring(name)] {
+            struct group grp;
+            struct group *result;
+            memset(&grp, 0, sizeof(struct group));
+            char buf[1024];
+            errno = 0;
+            int ret = ::getgrnam_r(name.c_str(), &grp, buf, sizeof(buf), &result);
+            return wrap_syscall(ret, result ? std::optional<struct group>(*result) : std::nullopt);
```
muthu90tech, line 2214: I think there is a problem here, buf memory which holds the results (group info) is on the stack, and can go out of scope after the lambda.
The copty to group_details must be made here, instead of outside the lambda ?

avikivity: You can allocate it on the coroutine's stack, where the coroutine will save to its frame.

include/seastar/core/file.hh
```diff
@@ -61,6 +61,14 @@ struct directory_entry {
     std::optional<directory_entry_type> type;
 };
```
avikivity, line 63: > Add APIs to get group details and to change ownership of file.

Please enrish the patch changelog with the details.

muthu90tech: > > Add APIs to get group details and to change ownership of file.
> 
> Please enrich the patch changelog with the details.

Thanks, I have added details to the commit log

avikivity:
Please drop the artificial merge commit.

muthu90tech:
> Please drop the artificial merge commit.

@avikivity , apologies, syncing my fork resulted in this, corrected it now.

PR #2315, opened 2024-07-01
gate: add try_hold

bhalevy:
Add a method to try acquiring a gate holder
that gracefully returns nullopt if the gate
is already closed, rather than throwing an exception.


bhalevy:
@raphaelsc please review

PR #2316, opened 2024-07-01
github: s/ubuntu-latest/ubuntu-24.04/

tchaikov:
Ubuntu 24.04 is now available, see
https://github.com/actions/runner-images/issues/9848

so instead of building and running the tests in a fedora 40 container, let's do the CI with the shinny ubuntu runner image, which preinstalls some of the necessary building dependencies, so in theory, it should be relatively faster than installing every package on fedora 40.

and since we are now using an image with git pre-installed, there is no need to install `git` anymore. so, in this change, we drop the step to install `git`.

because we are switching to a debian derivative distro, let's use apt-get instead of dnf to install packages.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -32,16 +32,14 @@ on:
 jobs:
   test:
     timeout-minutes: 40
-    runs-on: ubuntu-latest
-    container: fedora:40
+    runs-on: ubuntu-24.04
     steps:
-      - name: Install Git
-        run: |
-          sudo dnf -y install git
-
       - uses: actions/checkout@v4
         with:
           submodules: "${{ contains(inputs.enables, 'dpdk') }}"
+
+      - run: |
+          sudo apt-get update
 
       - name: Install build dependencies
         run: |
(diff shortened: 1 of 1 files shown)


tchaikov:
the workflows are failing due to https://github.com/scylladb/seastar/issues/2302

mykaul:
I'm not sure I understand the benefit here. Would be good to compare runtimes.

tchaikov:
> I'm not sure I understand the benefit here. Would be good to compare runtimes.

@mykaul take the "Test (clang++, C++, dev) / test as an example:

|  |initialize container | install git | install dependencies | install clang | install ccache | build | test | stop container | total |  
| ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
| f40 | 4s | 11s | 31s | 14s |1s | 7s | 3m 6s | 1s | 4m 32s |
| ubuntun 24.04 | 0 | 0 | 25s | 3s | 4s | 9s | 2m 50s | 0 | 3m 44s |

before this change, see https://github.com/scylladb/seastar/actions/runs/9741537615/job/26880954875

after this change, see https://github.com/scylladb/seastar/actions/runs/9741786724/job/26881705458

mykaul:
45 secs saving is noise level (IMHO) for CI.

tchaikov:
@mykaul i am not able to provide a statistically signifcant report, but i think to skip following steps

- initialize container
- install git
- install clang
- stop container

and to drop the unnecessary step in the workflow already deserve the change.

tchaikov:
@scylladb/seastar-maint hello maintainers, could you help review this change? i am seeing timeout when installing the dependencies recently. see https://github.com/scylladb/seastar/actions/runs/9817554876/job/27109045950#step:5:529

.github/workflows/test.yaml
```diff
@@ -50,16 +45,16 @@ jobs:
       - name: Install clang++
         if: ${{ inputs.compiler == 'clang++' }}
         run: |
-          sudo dnf -y install clang
+          sudo apt-get -y install clang
```
avikivity, line 48: What about apt-get update?

tchaikov: sure. added.

tchaikov:
v2:

- run `apt-get update` before installing dependencies.

mykaul:
ping @avikivity

PR #2317, opened 2024-07-01
build: run socket test with linux-aio reactor backend

tchaikov:
to workaround #2302, before it is fixed.

Refs #2302

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -420,7 +420,9 @@ seastar_add_test (app-template
   SOURCES app-template_test.cc)
 
 seastar_add_test (socket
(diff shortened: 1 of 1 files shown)


tchaikov:
@xemul hi Pavel, could you help review this change?

xemul:
How is backend selected when it's run on CI? Automatically by seastar itself, as it it would select it for any other application?

tchaikov:
> How is backend selected when it's run on CI? Automatically by seastar itself, as it it would select it for any other application?

@xemul yeah, when running on CI, tests selects the seastar backend automatically if we don't pick one for it. i think that's also why all of a sudden, the tests are failing after switching to the new image. see #2316

PR #2318, opened 2024-07-01
timer: fix cancel in periodic timer callback

brain5lug:
Fix new timers unable to set system timer callback
if there was removed timer with low new expire
time which was removed in their seastar callback

Please comment bad index calculated of timer then timer
is removed.

### DIFF
--- a/apps/memcached/memcache.cc
+++ b/apps/memcached/memcache.cc
@@ -124,7 +124,7 @@ struct expiration {
         return _time != never_expire_timepoint;
     }
 
-    time_point to_time_point() {
+    time_point to_time_point() const {
         return _time;
     }
 };
@@ -176,7 +176,7 @@ public:
     item(const item&) = delete;
     item(item&&) = delete;
 
-    clock_type::time_point get_timeout() {
+    clock_type::time_point get_timeout() const {
         return _expiry.to_time_point();
     }
 


include/seastar/core/timer-set.hh
```diff
@@ -151,15 +160,28 @@ public:
      * Postconditions:
      *  - timer is no longer in the active set.
      *  - this object will no longer hold any references to this timer.
+     *
+     * get_next_timeout() will return correct next time point after removing
+     * timer with earliest timeout.
      */
     void remove(Timer& timer) noexcept
```
tgrabiec, line 167: What is the incorrect behavior you're observing exactly, what do you mean by "unable to set system timer callback". E.g. new timers never fire?

include/seastar/core/timer-set.hh
```diff
@@ -151,15 +160,28 @@ public:
      * Postconditions:
      *  - timer is no longer in the active set.
      *  - this object will no longer hold any references to this timer.
+     *
+     * get_next_timeout() will return correct next time point after removing
+     * timer with earliest timeout.
```
tgrabiec, line 165: What was "incorrect" about the currently returned value? 

The doc about get_next_timeout() says:
```
    /**
     * Returns a time point at which expire() should be called
     * in order to ensure timers are expired in a timely manner.
     *
     * Returned values are monotonically increasing.
     */
```
So it is the time to call next expire(), not the time of the nearest expiring timer. After you call expire(), next timeout will be recalculated based on currently active timers.

include/seastar/core/timer-set.hh
brain5lug: The case is cased by timer self cancel in periodic timer callback. Sleep time for reactor were calculated badlly (in past). That caused reactor never come from sleep.

include/seastar/core/timer-set.hh
```diff
@@ -66,7 +66,7 @@ private:
         return _time_point.time_since_epoch().count();
     }
 
-    static timestamp_t get_timestamp(Timer& timer) noexcept
+    static timestamp_t get_timestamp(Timer const& timer) noexcept
```
nyh, line 69: `Timer const&` is equivalent to `const Timer&`, and we usually use the latter format. I also personally find it much more "obvious" and readable.

include/seastar/core/timer-set.hh
```diff
@@ -151,15 +160,28 @@ public:
...
+     *
+     * get_next_timeout() will return correct next time point after removing
+     * timer with earliest timeout.
      */
     void remove(Timer& timer) noexcept
     {
+        // The code below very likely to have a bug in detemining correct bucket
+        // actual bucket shall be in range index - bucket_n
+        // timer might be not found in real storing bucket, _non_empty_buckets
+        // might have invalid state
```
nyh, line 172: I don't understand this comment. If it has a bug, can you fix it? If it doesn't, we shouldn't say it is "very likely to have a bug"...

tests/unit/timer_test.cc
```diff
@@ -92,8 +99,29 @@ struct timer_test {
...
+            pr3.set_value();
+        });
+        t7.set_callback([this] { OK(); pr4.set_value(); });
+        t6.arm_periodic(10ms);
+        (void)pr3.get_future().then([this]{
+            t7.arm(100ms);
+        });
+
+        //some long 3rd party unstoppable calculation
+        std::this_thread::sleep_for(100ms);
```
nyh, line 117: I don't understand this comment. Why do you need the sleep_for()? the waiting for p4's future will wait for more than 100ms anyway.
Or maybe what you're saying is that somehow the cancel() broke this sleep_for() and before this patch, the sleep_for would never finish? If this is the case, please a comment saying this, and not some statement about "long 3rd party unstoppable calculation", whatever that means.

include/seastar/core/timer-set.hh
brain5lug: @nyh I didn't wanted to merge this as it is. As proper and effective fix seems not to be really straight forward. At first I wanted some comment from the support side.

nyh:
> Please add a test that demonstrates the bug (fails before the fix, passes after the fix).

I see he did add a test, but never explicitly said that this test reproduces the bug - fails before the fix, and passes afterwards.
I would suggest to please:
1. Open a github **issue** about the bug. Explaining exactly what the bug is and what its symptoms were (but not how to fix it).
2. Above the test, add a comment explaing that it reproduces the issue (mentioning the issue number).
3. Add to the pull request a "Fixes #..." line - which will cause the issue to be automatically closed when the issue is merged.

Thanks.

brain5lug:
> Please add a test that demonstrates the bug (fails before the fix, passes after the fix).

The test has been already added. Probably it requires some improvements. Currently the test hangs forever (as reactor hangs) without the fix and passes the test if fix is applied. The fix is for one bug. I suspect another bug. I can prepare branch with assert which shall prove it exists.

include/seastar/core/timer-set.hh
brain5lug: removed all comment

include/seastar/core/timer-set.hh
brain5lug: fixed

include/seastar/core/timer-set.hh
brain5lug: Comment has been removed. Will add separated branch with asserts to highlight issue if exists.

include/seastar/core/timer-set.hh
brain5lug: The confusing comment has been removed.

tests/unit/timer_test.cc
brain5lug: Added explicit comment that this line is for reactor delay

brain5lug:
I made some investigations about second possible issue located at following lines:
```
auto index = get_index(timer);
auto& list = _buckets[index];
list.erase(list.iterator_to(timer));
if (list.empty()) {
    _non_empty_buckets[index] = false;
}
```
This issue would be quite difficult to observe. The issue might appear only when max bit of time overlaps.
- The timer must be scheduled before the time overlaps it's maximum bit
- The timer must be canceled after time overlaps it's maximum bit (and after expire(time_point now) noticed it)
- The timer must be last in the bucket? (`list.erase(list.iterator_to(timer)); ` might work fine even for bad bucket because it's intrusive)

Current time 21804433186261 on my machine does not match linux epoch (1073741824). Verification for this would require custom time source which I don't have time for today to provide.

PR #2319, opened 2024-07-02
io_queue: make queued I/O requests more generic

pwrobelse:
Before this patch 'io_queue' was strongly oriented on read and write operations.
The top-level class called 'queued_io_request' depended on implementation details related
to 'io_sink' and 'internal::io_request' -- there was an assumption that all requests are routed
via the mentioned sink.

The interface of 'io_queue' contained four functions related to managing the execution of I/O requests.
1. 'void cancel_request(queued_io_request& req)'
2. 'void complete_cancelled_request(queued_io_request& req)'
3. 'void submit_request(io_desc_read_write* desc, internal::io_request req)'
4. 'void complete_request(io_desc_read_write& desc)'

Although the names of mentioned functions suggest that 'request' is a parameter, in two cases
an implementation detail related to 'io_sink' that is derived from 'io_completion' was used -- 'io_desc_read_write'.

The idea of this series of changes is to make 'io_queue' more generic to allow routing discard requests via it.
Unfortunately, discard requests cannot be routed via 'io_sink', because 'fallocate()' is not supported by Linux AIO.

This pull request:
- Removes the implementation details related to read and write operations from
'queued_io_request' class and makes it an abstract base for more specialized request types.
- Cleans the interface of 'io_queue' from the usage of types derived from 'io_completion'. A request
type may not use 'io_sink'. Therefore, 'io_queue' should not force the request types to use 'io_completion'
to finish the execution of a request.
- Allows request types derived from 'queued_io_request' to select the means to submit the
actual work to be performed - the assumption about routing each request via 'io_sink' is removed.
- Introduces a new type derived from 'queued_io_request' that is called 'queued_read_write_request'.
It encapsulates the logic required to handle read and write I/O requests.
- Moves the code of 'io_desc_read_write' to 'queued_read_write_request' to allow exclusive usage of
'queue_io_request' in the interface of 'io_queue'.

Refs: https://github.com/scylladb/scylladb/issues/14940

### DIFF
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -163,7 +163,7 @@ public:
     future<size_t> submit_io_write(internal::priority_class priority_class,
             size_t len, internal::io_request req, io_intent* intent, iovec_keeper iovs = {}) noexcept;
 
-    void submit_request(io_desc_read_write* desc, internal::io_request req) noexcept;
+    void request_dispatched() noexcept;
     void cancel_request(queued_io_request& req) noexcept;
     void complete_cancelled_request(queued_io_request& req) noexcept;
     void complete_request(io_desc_read_write& desc) noexcept;
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -65,7 +65,6 @@ class io_sink;
 using shard_id = unsigned;
 using stream_id = unsigned;
 
-class io_desc_read_write;
 class queued_io_request;
 class io_group;
 
@@ -166,7 +165,7 @@ public:
     void request_dispatched() noexcept;
     void cancel_request(queued_io_request& req) noexcept;
     void complete_cancelled_request(queued_io_request& req) noexcept;
-    void complete_request(io_desc_read_write& desc) noexcept;
+    void complete_request(queued_io_request& req) noexcept;
 
     [[deprecated("I/O queue users should not track individual requests, but resources (weight, size) passing through the queue")]]
     size_t queued_requests() const {
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -47,7 +47,6 @@ module seastar;
 #include <seastar/core/reactor.hh>
 #include <seastar/core/when_all.hh>
 #include <seastar/core/metrics.hh>
-#include <seastar/core/internal/io_desc.hh>
 #include <seastar/core/internal/io_sink.hh>
 #include <seastar/core/io_priority_class.hh>
 #include <seastar/util/log.hh>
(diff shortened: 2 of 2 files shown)


src/core/io_queue.cc
```diff
@@ -348,14 +348,22 @@ class queued_io_request {
...
 
 public:
-    queued_read_write_request(internal::io_request req, io_queue& q, fair_queue_entry::capacity_t cap, io_queue::priority_class_data& pc, io_direction_and_length dnl, iovec_keeper iovs)
+    queued_read_write_request(internal::io_sink& sink,
+                              internal::io_request req,
+                              io_queue& q,
+                              fair_queue_entry::capacity_t cap,
+                              io_queue::priority_class_data& pc,
+                              io_direction_and_length dnl,
+                              iovec_keeper iovs)
```
xemul, line 325: So now the read-write needs to reference
- io-queue
- io-sink
- proprity-class-data

:( Too messy already, but I don't suggest to do anything about it right now

src/core/io_queue.cc
```diff
@@ -365,34 +326,79 @@ class queued_read_write_request : public queued_io_request {
...
+
+        io_log.trace("dev {} : req {} submit", _ioq.dev_id(), fmt::ptr(this));
+        auto now = io_queue::clock_type::now();
+        _pclass.on_dispatch(_dnl, std::chrono::duration_cast<std::chrono::duration<double>>(now - _ts));
+        _ts = now;
+
         _ioq.request_dispatched();
-        _sink.submit(_desc.release(), std::move(_io_request));
-        delete this;
+        _sink.submit(this, std::move(_io_request));
```
xemul, line 365: The _io_request part is not duplicated inside sink. First, it lives on read-write request here (in moved-from form, but it still occupies memory), second as a part of pending_io_request

src/core/io_queue.cc
```diff
@@ -365,34 +326,79 @@ class queued_read_write_request : public queued_io_request {
...
+    // Implements the interface required by queued_io_request base class.
+    // Returns the future for the submitter of the request.
     future<size_t> get_future() noexcept override {
-        return _desc->get_future();
+        return _pr.get_future();
+    }
+
+    // An error callaback from io_completion interface. Finalizes the execution via io_queue and
+    // forwards the error to the submitter. At the end it destroys 'this' object - it is the last reference.
+    void set_exception(std::exception_ptr eptr) noexcept override {
```
xemul, line 385: Probably can be private

xemul: Ah, no, it's overridden method of io_completion

include/seastar/core/io_queue.hh
```diff
@@ -172,7 +171,7 @@ public:
     void request_dispatched() noexcept;
     void cancel_request(queued_io_request& req) noexcept;
     void complete_cancelled_request(queued_io_request& req) noexcept;
-    void complete_request(io_desc_read_write& desc) noexcept;
+    void complete_request(queued_io_request& req) noexcept;
```
xemul, line 174: That's because upon completion fair-queue needs to be notified about completed request. And this, in turn, is now empty but used to be needed because fair-queue erroneously contained the output throttler on board. In fact, the throttler should be on its own and referenced by io_queue directly, it's not the part of fair queue

pwrobelse: Is `fair_queue::notify_request_finished()` needed then? As you said it does nothing. Do you think that the empty function can be safely removed?

If it was removed, then we could rename `io_queue::complete_request(queued_io_request& req)` to `io_queue::request_completed()` -- the function would only be responsible for accounting -- just like `io_queue::request_dispatched()`.


Moreover, I was wondering if we want to rename (or somehow rework) `io_queue::queue_request()` and `io_queue::queue_one_request()` . Both functions are specific to read and write requests -- they explicitly create `queued_read_write_request`.

I just think out loud -- perhaps, the discards request would not need to take `internal::io_request` parameter -- it is needed by `io_sink`.

src/core/io_queue.cc
pwrobelse: This PR introduced the dependency to `io_sink` -- now the request knows how it wants to be dispatched.

The remaining dependencies are preserved from the baseline.

src/core/io_queue.cc
pwrobelse: Could you elaborate a little bit?

Previously `queued_io_request` was derived from `internal::io_request`. Private inheritance was used -- it models `has-a` relationship that is almost like having a member variable.

Because we neither expose any members of this base class via using directive nor implement empty-base optimization idiom I concluded that having private inheritance in code is a little bit distracting for a reader -- using member variable was much simpler to understand.

The statement about memory usage is true. Previously, `queued_io_request` was destroyed after it was dispatched. Thus, its `internal::io_request` sub-object was removed. Currently, it is not like that - we do not remove the object right after the dispatch (if it was submitted to io_sink).

Do you propose to allocate `_io_request` on heap? This way the object would be freed after the dispatch. In such case `_cancelled` variable could be removed -- we could use the pointer to `_io_request` to check whether the request can be dispatched.


Also one note about moved-from state of `internal::io_request` -- this type is a tagged union that contains only POD structs -- thus as far as I understand move is just a copy -- we cannot "steal" any resources from the original object.

xemul: There are currently 4 objects involved in IO:
- internal::io_request -- it's an API thing that users of IO queue use to tell what IO they want to do. BTW, I think you need to extend it and add operation::discard into it
- queued_io_request -- it's used to track a request that sits in seastar queue. It has fair_queue entry to track its queueing, it has internal::io_request itself, because this is what this request is doing and it has io_desc_read_write (see below)
- pending_io_request -- it's used to track a request that sits in the sink, i.e. -- ready to be sent to the kernel. It has internal::io_request itfels, because this is what this request is doing, and (via io_completion) the io_desc_read_write (see below)
- io_desc_read_write -- it's a completion tracker that's implicitly attached to internal::io_request once it gets into io-queue and that's used to inform all the interested parties after the request completes

My point is that queued_io_request and pending_io_request has their well defined role and life-time -- they help internal::io_request and io_completion (backed by io_desc_read_write) travel through io stack -- the former is there while the request is queued the latter is there while the request is being dispatched (very short). Also queued_io_request is cancellable, while pending_io_request isn't.

This change breaks this harmony by making queued_io_request combine several roles and extending its lifetime. It's not bad on its own, my point is mostly about pending_io_request -- after the above change it's no longer clear what it's needed for.

xemul: > Also one note about moved-from state of internal::io_request -- this type is a tagged union that contains only POD structs -- thus as far as I understand move is just a copy -- we cannot "steal" any resources from the original object.

Yes, of course, my concern was that "it still occupies memory"

pwrobelse: Okay, thank you for the feedback :+1: 

>internal::io_request -- it's an API thing that users of IO queue use to tell what IO they want to do. BTW, I think you need to extend it and add operation::discard into it

Sure, I have just prepared a PR with the suggested change -- please find https://github.com/scylladb/seastar/pull/2330

Regarding the adjustments of `io_queue` -- I will rework the code according to your comments to meet the required criteria. Also, locally I tried to move forward with `queued_discard_request` implementation. I ended up with a class, that it is very similar to `queued_read_write_request` -- majority of the functionality was identical (except the fact that it was called from different places).

I will rework this PR.

pwrobelse:
I am closing this PR in favor of:

1. Moving `throttling` to a separate class that brings `virtual void dispatch()` to `queued_io_request` (implemented by Pavel in https://github.com/scylladb/seastar/pull/2332 )
2. Introducing discard request type to `internal::io_request` (implemented by me via https://github.com/scylladb/seastar/pull/2330 )
3. Renaming `io_desc_read_write` to `queued_io_request_completion` (implemented by me via https://github.com/scylladb/seastar/pull/2339 )

Rationale:

- When I tried to implement `queued_discard_request` on top of the current PR (#2319), then I ended up with a lot of code duplication -- although discard requests were not submitted to `io_sink` they needed almost identical code to complete the execution via `io_queue`.
- Also, Pavel's change brings the possibility to overload dispatch function of `queued_io_request` -- it should be sufficient to implement discard requests that are not submitted to `io_sink`

PR #2320, opened 2024-07-02
code-cleanup: explicitly depend on io_desc.hh

pwrobelse:
The mentioned header file contains 'kernel_completion' interface.
Multiple classes implement it. However, the files in which the
derived classes are defined very often do not explicitly include
the header file in which the base class is defined. Therefore,
they can be easily broken.

In some places like reactor.hh the dependency was implicit.
The mentioned header pulled 'kernel_completion' via 'pollable_fd.hh'.
The file with the definition of an interface was included implicitly
-- by another header file. Therefore, if anybody removed 'io_desc.hh'
include from 'pollable_fd.hh', then 'reactor.hh' would stop to compile.

This change ensures that the header which contains the definition of
'kernel_completion' interface is explicilty included in places that need
it. This patch also removes the includes of the mentioned header from
files that do not need it.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -31,6 +31,7 @@
 #include <seastar/core/file.hh>
 #include <seastar/core/future.hh>
 #include <seastar/core/idle_cpu_handler.hh>
+#include <seastar/core/internal/io_desc.hh>
 #include <seastar/core/internal/io_request.hh>
 #include <seastar/core/internal/io_sink.hh>
 #include <seastar/core/iostream.hh>
@@ -140,7 +141,6 @@ void increase_thrown_exceptions_counter() noexcept;
 
 }
 
-class kernel_completion;
 class io_queue;
 SEASTAR_MODULE_EXPORT
 class io_intent;
--- a/include/seastar/core/internal/io_request.hh
+++ b/include/seastar/core/internal/io_request.hh
@@ -22,7 +22,6 @@
 #pragma once
 
 #include <seastar/core/sstring.hh>
-#include <seastar/core/internal/io_desc.hh>
 #include <seastar/core/on_internal_error.hh>
 #include <cassert>
 #include <cstdint>
--- a/include/seastar/core/internal/pollable_fd.hh
+++ b/include/seastar/core/internal/pollable_fd.hh
@@ -23,7 +23,6 @@
 
 #include <seastar/core/future.hh>
 #include <seastar/core/posix.hh>
-#include <seastar/core/internal/io_desc.hh>
 #include <seastar/util/bool_class.hh>
 #include <seastar/util/modules.hh>
 #ifndef SEASTAR_MODULE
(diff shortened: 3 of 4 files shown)


include/seastar/core/reactor.hh
```diff
@@ -142,7 +143,6 @@ void increase_thrown_exceptions_counter() noexcept;
 
 }
 
-class kernel_completion;
```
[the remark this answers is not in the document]
pwrobelse: Previously the full definition of `kernel_completion` was implicitly included via `include/seastar/core/internal/pollable_fd.hh`

include/seastar/core/reactor.hh
```diff
@@ -31,6 +31,7 @@
 #include <seastar/core/file.hh>
 #include <seastar/core/future.hh>
 #include <seastar/core/idle_cpu_handler.hh>
+#include <seastar/core/internal/io_desc.hh>
```
xemul, line 34: So you depopulated internal header at the cost of bloating core public one. That's not very nice, reactor.hh should be put on a diet instead

pwrobelse: No, that's not the case. Please let me explain and attach some references.

Firstly, `reactor.hh` was already bloated and it was easy to break in unexpected way. I will start from the part related to `breaking in unexpected way`.

Let's assume that you are a programmer and you open `pollable_fd.hh`. You see, that nothing in this file needs `kernel_completion` that is the only type defined in `io_desc.hh`. This means that the inclusion of that header is redundant. Normally, because of programming "scout rule" you want to leave the file cleaner when you leave it, so you remove the unneeded include to remove the dependency that is in fact redundant.

What happens after the removal? The project stops to compile, because `reactor.hh` was not self-contained [[1]](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf11-header-files-should-be-self-contained). According to cppcoreguidlines it is a good rule of thumb to make all headers self-contained.

In `reactor.hh` there was a forward declaration of `kernel_completion`  and then we tried to derive a type from that forward declaration. This is illegal [[2]](https://godbolt.org/z/7aK8aGsW5).

What's more important, `reactor.hh` was already bloated, because of the way how `#include` directive works. It is just a simple copy-paste tool. Therefore, if anybody included `reactor.hh` he/she implicitly received `io_desc.hh`, because `reactor.hh` includes `pollable_fd.hh` and it includes `io_desc.hh`.

What is wrong is the fact, that any file that includes `pollable_fd.hh` implicitly includes `io_desc.hh` header, that is not needed by `pollable_fd.hh`. If it is not needed also by the client of `pollable_fd.hh` then we created unneeded dependency.

Is this explanation enough? Honestly, I do not understand the meaning of `the cost of bloating core public one` -- the public one is already bloated on master and is easy to break due to invalid usage of forward declaration and it compiles just by luck and implicit include of `io_desc.hh` via `pollable_fd.hh` that in fact should not include it.

xemul: I see. OK, this makes sense. Please update the description to mention that reactor.hh pulled kernel_completion via pollable_fd.hh. And this PR cannot now be merged due to conflict, so please rebase as well

pwrobelse: Sure, fixed. Thanks :+1:

PR #2321, opened 2024-07-02
scripts/perftune.py: aRFS should be disabled by default in non-MQ mode

vladzcloudius:
Commit 661367668f1e459bfd2f82450057d3944da7e20 introduced a regression by enabling ntuples by default for a configuration when the applicaiton CPUs and IRQ cpus are not the same.

Let's fix that.

### DIFF
--- a/scripts/perftune.py
+++ b/scripts/perftune.py
@@ -694,7 +694,7 @@ class NetPerfTuner(PerfTunerBase):
         op = "Enable"
         value = 'on'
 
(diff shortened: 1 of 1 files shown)


bhalevy (approved):
LGTM, based on the code documentation and the backward compatibility with
```
        # If no explicit configuration has been requested enable ntuple (and thereby aRFS) only in MQ mode.
```

So in MQ mode we expect `self.irqs_cpu_mask == self.cpu_mask`, and so we want to disable arfs by default as this patch does, if `self.irqs_cpu_mask != self.cpu_mask`.

PR #2332, opened 2024-07-05
Move output IO throttler to IO queue level

xemul:
IO queue is responsible for two things:
- cross-class shares-based fair balancing
- throttling the flow of dispatching requests to obey disk model

The former is done with the help of fair_queue that implements rather simple "virtual capacity" model. The latter is done with the help of shared token bucket, but historically the throttling code was implemented as a part of fair_queue. That's not correct, fair queue has nothing to do with output flow control. This PR moves the throttling code out of fair_queue and makes IO-queue own it and use.

### DIFF
--- a/include/seastar/core/internal/throttle.hh
+++ b/include/seastar/core/internal/throttle.hh
@@ -1,0 +1,171 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Copyright (C) 2016 ScyllaDB
+ */
+#pragma once
+
+#include <seastar/core/sstring.hh>
+#include <seastar/util/shared_token_bucket.hh>
+
+namespace seastar {
+
+/// \brief Group of queues class
+///
+/// This is a fair group. It's attached by one or mode fair queues. On machines having the
+/// big* amount of shards, queues use the group to borrow/lend the needed capacity for
+/// requests dispatching.
+///
+/// * Big means that when all shards sumbit requests alltogether the disk is unable to
+/// dispatch them efficiently. The inability can be of two kinds -- either disk cannot
+/// cope with the number of arriving requests, or the total size of the data withing
+/// the given time frame exceeds the disk throughput.
+class shared_throttle {
+public:
+    using capacity_t = uint64_t; // XXX -- this could be a template parameter
+    using clock_type = std::chrono::steady_clock;
+
+    /*
+     * tldr; The math
+     *
+     *    Bw, Br -- write/read bandwidth (bytes per second)
+     *    Ow, Or -- write/read iops (ops per second)
+     *
+     *    xx_max -- their maximum values (configured)
+     *
+     * Throttling formula:
+     *
+     *    Bw/Bw_max + Br/Br_max + Ow/Ow_max + Or/Or_max <= K
+     *
+     * where K is the scalar value <= 1.0 (also configured)
+     *
+     * Bandwidth is bytes time derivatite, iops is ops time derivative, i.e.
+     * Bx = d(bx)/dt, Ox = d(ox)/dt. Then the formula turns into
+     *
+     *   d(bw/Bw_max + br/Br_max + ow/Ow_max + or/Or_max)/dt <= K
+     *
+     * Fair queue tickets are {w, s} weight-size pairs that are
+     *
+     *   s = read_base_count * br, for reads
+     *       Br_max/Bw_max * read_base_count * bw, for writes
+     *
+     *   w = read_base_count, for reads
+     *       Or_max/Ow_max * read_base_count, for writes
+     *
+     * Thus the formula turns into
+     *
+     *   d(sum(w/W + s/S))/dr <= K
+     *
+     * where {w, s} is the ticket value if a request and sum summarizes the
+     * ticket values from all the requests seen so far, {W, S} is the ticket
+     * value that corresonds to a virtual summary of Or_max requests of
+     * Br_max size total.
+     */
+
+    /*
+     * The normalization results in a float of the 2^-30 seconds order of
+     * magnitude. Not to invent float point atomic arithmetics, the result
+     * is converted to an integer by multiplying by a factor that's large
+     * enough to turn these values into a non-zero integer.
+     *
+     * Also, the rates in bytes/sec when adjusted by io-queue according to
+     * multipliers become too large to be stored in 32-bit ticket value.
+     * Thus the rate resolution is applied. The t.bucket is configured with a
+     * time period for which the speeds from F (in above formula) are taken.
+     */
+
+    static constexpr float fixed_point_factor = float(1 << 24);
+    using rate_resolution = std::milli;
+    using token_bucket_t = internal::shared_token_bucket<capacity_t, rate_resolution, internal::capped_release::no>;
+
+private:
+
+    /*
+     * The dF/dt <= K limitation is managed by the modified token bucket
+     * algo where tokens are ticket.normalize(cost_capacity), the refill
+     * rate is K.
+     *
+     * The token bucket algo must have the limit on the number of tokens
+     * accumulated. Here it's configured so that it accumulates for the
+     * latency_goal duration.
+     *
+     * The replenish threshold is the minimal number of tokens to put back.
+     * It's reserved for future use to reduce the load on the replenish
+     * timestamp.
+     *
+     * The timestamp, in turn, is the time when the bucket was replenished
+     * last. Every time a shard tries to get tokens from bucket it first
+     * tries to convert the time that had passed since this timestamp
+     * into more tokens in the bucket.
+     */
+
+    token_bucket_t _token_bucket;
+    const capacity_t _per_tick_threshold;
+
+public:
+
+    // Convert internal capacity value back into the real token
+    static double capacity_tokens(capacity_t cap) noexcept {
+        return (double)cap / fixed_point_factor / token_bucket_t::rate_cast(std::chrono::seconds(1)).count();
+    }
+
+    // Convert floating-point tokens into the token bucket capacity
+    static capacity_t tokens_capacity(double tokens) noexcept {
+        return tokens * token_bucket_t::rate_cast(std::chrono::seconds(1)).count() * fixed_point_factor;
+    }
+
+    auto capacity_duration(capacity_t cap) const noexcept {
+        return _token_bucket.duration_for(cap);
+    }
+
+    struct config {
+        sstring label = "";
+        /*
+         * There are two "min" values that can be configured. The former one
+         * is the minimal weight:size pair that the upper layer is going to
+         * submit. However, it can submit _larger_ values, and the fair queue
+         * must accept those as large as the latter pair (but it can accept
+         * even larger values, of course)
+         */
+        double min_tokens = 0.0;
+        double limit_min_tokens = 0.0;
+        std::chrono::duration<double> rate_limit_duration = std::chrono::milliseconds(1);
+    };
+
+    explicit shared_throttle(config cfg, unsigned nr_queues);
+    shared_throttle(shared_throttle&&) = delete;
+
+    capacity_t maximum_capacity() const noexcept { return _token_bucket.limit(); }
+    capacity_t per_tick_grab_threshold() const noexcept { return _per_tick_threshold; }
+    capacity_t grab_capacity(capacity_t cap) noexcept;
+    clock_type::time_point replenished_ts() const noexcept { return _token_bucket.replenished_ts(); }
+    void replenish_capacity(clock_type::time_point now) noexcept;
+    void maybe_replenish_capacity(clock_type::time_point& local_ts) noexcept;
+
+    capacity_t capacity_deficiency(capacity_t from) const noexcept;
+
+    std::chrono::duration<double> rate_limit_duration() const noexcept {
+        std::chrono::duration<double, rate_resolution> dur((double)_token_bucket.limit() / _token_bucket.rate());
+        return std::chrono::duration_cast<std::chrono::duration<double>>(dur);
+    }
+
+    const token_bucket_t& token_bucket() const noexcept { return _token_bucket; }
+};
+
+} // seastar namespace
--- a/include/seastar/core/internal/throttle.hh
+++ b/include/seastar/core/internal/throttle.hh
@@ -168,4 +168,53 @@ public:
     const token_bucket_t& token_bucket() const noexcept { return _token_bucket; }
 };
 
+class throttle {
+    using clock_type = std::chrono::steady_clock;
+
+    /*
+     * When the shared capacity os over the local queue delays
+     * further dispatching untill better times
+     *
+     * \head  -- the value group head rover is expected to cross
+     * \cap   -- the capacity that's accounted on the group
+     *
+     * The last field is needed to "rearm" the wait in case
+     * queue decides that it wants to dispatch another capacity
+     * in the middle of the waiting
+     */
+    struct pending {
+        shared_throttle::capacity_t head;
+        shared_throttle::capacity_t cap;
+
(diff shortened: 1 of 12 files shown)


src/core/io_queue.cc
```diff
@@ -297,6 +298,10 @@ class queued_io_request final : private internal::io_request, public fair_queue_
     queued_io_request(queued_io_request&&) = delete;
     ~queued_io_request() = default;
 
+    virtual throttle::grab_result can_dispatch() const noexcept {
```
pwrobelse, line 301: It seems that the compiler on CI complains that it does not use `override` keyword.

```
io_queue.cc:301:35: error: 'can_dispatch' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
  301 |     virtual throttle::grab_result can_dispatch() const noexcept {
```

src/core/fair_queue.cc
```diff
@@ -243,28 +242,24 @@ void fair_queue::notify_request_cancelled(fair_queue_entry& ent) noexcept {
...
 
-fair_queue::clock_type::time_point fair_queue::next_pending_aio() const noexcept {
-    return _throttle.next_pending();
-}
-
 void fair_queue::dispatch_requests() {
     capacity_t dispatched = 0;
     boost::container::small_vector<priority_class_ptr, 2> preempt;
 
-    while (!_handles.empty() && (dispatched < _throttle.per_tick_grab_threshold())) {
```
pwrobelse: Minor: `dispatched` variable becomes unused after the removal. The compiler on CI complains about it:

```
fair_queue.cc:238:16: error: variable 'dispatched' set but not used [-Werror,-Wunused-but-set-variable]
  238 |     capacity_t dispatched = 0;
```

src/core/io_queue.cc
```diff
@@ -602,17 +602,17 @@ shared_throttle::config io_group::make_throttle_config(const io_queue::config& q
...
 std::chrono::duration<double> io_group::io_latency_goal() const noexcept {
-    return _fgs.front().rate_limit_duration();
+    return _throttle.front().rate_limit_duration();
 }
 
 io_group::io_group(io_queue::config io_cfg, unsigned nr_queues)
     : _config(std::move(io_cfg))
     , _allocated_on(this_shard_id())
 {
     auto fg_cfg = make_throttle_config(_config);
```
pwrobelse, line 615: Minor: maybe we could also adjust the variable name to `throttle_cfg`?

tests/unit/fair_queue_test.cc
```diff
@@ -55,15 +55,15 @@ struct request {
...
+    shared_throttle _fg;
     fair_queue _fq;
     std::vector<int> _results;
     std::vector<std::vector<std::exception_ptr>> _exceptions;
     fair_queue::class_id _nr_classes = 0;
     std::vector<request> _inflight;
 
-    static fair_group::config fg_config(unsigned cap) {
-        fair_group::config cfg;
+    static shared_throttle::config fg_config(unsigned cap) {
```
pwrobelse: Minor: maybe we could rename also the function to `throttle_config()`?

tests/unit/fair_queue_test.cc
```diff
@@ -55,15 +55,15 @@ struct request {
 };
 
 class test_env {
-    fair_group _fg;
+    shared_throttle _fg;
```
pwrobelse: Minor: in the next patch we rename `_fgs` from io_queue to `_throttle` and then adjust io_queue's tests. I know that this particular test is related to fair_queue and not io_queue. I am just wondering if `_fg` here should also be adjusted.

src/core/fair_queue.cc
```diff
@@ -161,7 +161,7 @@ bool fair_queue::class_compare::operator() (const priority_class_ptr& lhs, const
     return lhs->_accumulated > rhs->_accumulated;
 }
 
-fair_queue::fair_queue(fair_group& group, config cfg)
+fair_queue::fair_queue(shared_throttle& group, config cfg)
```
pwrobelse: Minor: in `fair_queue.hh` the names of parameters of the constructor do not match the ones that are used here. In the header there is `fair_queue(shared_throttle& shared, config cfg)`.

Do we want to keep both places consistent?

tests/perf/fair_queue_perf.cc
```diff
@@ -32,19 +32,19 @@
...
 struct local_fq_and_class {
-    seastar::fair_group fg;
+    seastar::shared_throttle fg;
     seastar::fair_queue fq;
     seastar::fair_queue sfq;
     unsigned executed = 0;
 
-    static fair_group::config fg_config() {
-        fair_group::config cfg;
+    static shared_throttle::config fg_config() {
```
pwrobelse: Minor: I am just wondering -- maybe the function name could be adjusted too and changed to `throttle_config()`?

tests/perf/fair_queue_perf.cc
```diff
@@ -32,19 +32,19 @@
 static constexpr fair_queue::class_id cid = 0;
 
 struct local_fq_and_class {
-    seastar::fair_group fg;
+    seastar::shared_throttle fg;
```
pwrobelse: Minor: In the next patch we change the name of `_fgs` to `_throttle` in io_queue. I am wondering -- do we want to adjust the names also in fair_queue's tests?

tests/perf/fair_queue_perf.cc
```diff
@@ -75,10 +75,10 @@ struct perf_fair_queue {
...
 
-    seastar::fair_group shared_fg;
+    seastar::shared_throttle shared_fg;
 
-    static fair_group::config fg_config() {
-        fair_group::config cfg;
+    static shared_throttle::config fg_config() {
+        shared_throttle::config cfg;
         return cfg;
     }
```
pwrobelse: Minor: the same question about changing `fgs` -> `throttle` in the next patch and usage of `fg` name in this file.

include/seastar/core/internal/throttle.hh
```diff
@@ -0,0 +1,171 @@
...
+#include <seastar/core/sstring.hh>
+#include <seastar/util/shared_token_bucket.hh>
+
+namespace seastar {
+
+/// \brief Group of queues class
+///
+/// This is a fair group. It's attached by one or mode fair queues. On machines having the
+/// big* amount of shards, queues use the group to borrow/lend the needed capacity for
+/// requests dispatching.
```
pwrobelse, line 33: Minor: just a short question about the comment with description. In previous patches we renamed `fair_group` --> `shared_throttle`. Are any changes needed in the case of the description? It uses the old nomenclature.

src/core/throttle.cc
```diff
@@ -0,0 +1,66 @@
...
+
+shared_throttle::shared_throttle(config cfg, unsigned nr_queues)
+        : _token_bucket(fixed_point_factor,
+                        std::max<capacity_t>(fixed_point_factor * token_bucket_t::rate_cast(cfg.rate_limit_duration).count(), tokens_capacity(cfg.limit_min_tokens)),
+                        tokens_capacity(cfg.min_tokens)
+                       )
+        , _per_tick_threshold(_token_bucket.limit() / nr_queues)
+{
+    if (tokens_capacity(cfg.min_tokens) > _token_bucket.threshold()) {
+        throw std::runtime_error("Fair-group replenisher limit is lower than threshold");
```
pwrobelse, line 38: Minor: we renamed `fair_group` to `shared_throttle`. Do we want to update also this error message?

src/core/throttle.cc
```diff
@@ -0,0 +1,66 @@
...
+ */
+/*
+ * Copyright 2024 ScyllaDB
+ */
+
+#ifdef SEASTAR_MODULE
+module;
+#endif
+
+#include <seastar/core/internal/throttle.hh>
```
pwrobelse, line 26: A general question related to the build with modules -- as far as I understand `./src/seastar.cc` includes all headers to provide the declarations and definitions when we build the code with modules.

The source files (e.g. `fair_queue.cc`, `io_queue.cc` and `reactor.cc`) use ifndef e.g.:
```
#ifdef SEASTAR_MODULE
module seastar;
#else
// include the headers related to seastar
#endif
```

The comment from seastar.cc:
```
// include all declaration and definitions to be exported in the the module
// purview
```

My question is as follows -- do we want to follow the same approach in `throttle.cc`? Also, is there an explicit rule in seastar project related to inclusion of files and `SEASTAR_MODULE` macro?

src/core/fair_queue.cc
pwrobelse: I am not entirely sure why the compiler raises error -- below there is `dispatched += req_cap;`

tests/perf/fair_queue_perf.cc
pwrobelse: Okay, please ignore the comment -- I see that at the end of patches we remove it completely.

tests/perf/fair_queue_perf.cc
pwrobelse: Okay, please ignore the comment -- I see that at the end of patches we remove it completely.

tests/perf/fair_queue_perf.cc
```diff
@@ -103,10 +92,9 @@ future<> perf_fair_queue::test(bool loc) {
 
     auto invokers = local_fq.invoke_on_all([loc] (local_fq_and_class& local) {
         return parallel_for_each(boost::irange(0u, requests_to_dispatch), [&local, loc] (unsigned dummy) {
-            auto cap = local.queue(loc).tokens_capacity(double(1) / std::numeric_limits<int>::max() + double(1) / std::numeric_limits<int>::max());
-            auto req = std::make_unique<local_fq_entry>(cap, [&local, loc, cap] {
+            auto req = std::make_unique<local_fq_entry>(100, [&local, loc] {
```
pwrobelse, line 95: Just to double-check -- this patch replaces the calculation with hard-coded value of 100. Was the value the same before? Does it matter for the test?

tests/unit/fair_queue_test.cc
pwrobelse: Okay, please ignore the comment -- I see that at the end of patches we remove it completely.

tests/perf/fair_queue_perf.cc
pwrobelse: Do I understand correctly that due to returning `throttle::grab_result::grabbed` from `can_dispatch()` the capacity does not matter to the test? Because we always can dispatch (because that's how the test is now) ?

tests/unit/fair_queue_test.cc
```diff
@@ -101,7 +90,6 @@ class test_env {
     unsigned tick(unsigned n = 0) {
         unsigned processed = 0;
         while (true) {
-            _fg.replenish_capacity(_fg.replenished_ts() + std::chrono::microseconds(1));
             _fq.dispatch_requests();
```
pwrobelse, line 93: Do I understand correctly that after defining:
```
    virtual throttle::grab_result can_dispatch() const noexcept override {
        return throttle::grab_result::grabbed;
    }
```

in `request` class, then this line will dispatch all request that were queued before? When reading the code of `dispatch_requests()` I had the impression, that it is the case.


If so, are we leaving the code as it was (mainly `while(true)` part) just in case if anybody wanted to define another request type that do not allow to be dispatched immediately? Is this `while(true)` needed (of course if I did not misunderstood the code) ?

src/core/io_queue.cc
```diff
@@ -551,7 +555,6 @@ void
 io_queue::complete_request(io_desc_read_write& desc) noexcept {
     _requests_executing--;
     _requests_completed++;
-    _streams[desc.stream()].notify_request_finished(desc.capacity());
```
pwrobelse, line 554: It seems, that the function does not need to take a parameter -- it is just responsible for accounting.

src/core/io_queue.cc
```diff
@@ -211,6 +211,11 @@ class io_queue::priority_class_data {
     metrics::metric_groups metric_groups;
 };
 
+io_queue::stream::stream(shared_throttle& st, fair_queue::config cfg)
+        : fq(std::move(cfg))
+        , thr(st)
+{ }
+
 class io_desc_read_write final : public io_completion {
```
pwrobelse, line 219: Because of the removal of the need to notify about completion, this class does not need to keep `_stream` and ` fq_capacity` members.

src/core/fair_queue.cc
```diff
@@ -303,51 +235,24 @@ void fair_queue::queue(class_id id, fair_queue_entry& ent) noexcept {
...
-
-    return std::chrono::steady_clock::time_point::max();
-}
-
-void fair_queue::dispatch_requests(std::function<void(fair_queue_entry&)> cb) {
+void fair_queue::dispatch_requests() {
     capacity_t dispatched = 0;
     boost::container::small_vector<priority_class_ptr, 2> preempt;
 
-    while (!_handles.empty() && (dispatched < _group.per_tick_grab_threshold())) {
```
pwrobelse, line 337: We removed here the check related to `per_tick_grab_threshold()` -- when I search for that function call in this PR then we did not introduce its usage anywhere.

Is it expected?

tests/unit/fair_queue_test.cc
pwrobelse, line 1: I pulled the change to my local repository and `fair_queue_test.cc` fails.

tests/unit/fair_queue_test.cc
pwrobelse: I suspect, that this is the root cause of the failure of the test.

Let's take `test_fair_queue_longer_run_different_shares` as an example.

We firstly queue a lot of ops:
```
    for (int i = 0; i < 20000; ++i) {
        env.do_op(a, 1);
        env.do_op(b, 0);
    }
```

And then we want to let them in over time:

```
   for (int i = 0; i < 1000; ++i) {
        sleep(1ms).get();
        env.tick(3);
    }
```

However, because all requests can be dispatched (new implementation of `can_dispatch()`) and the first call to `tick(3)` invokes `dispatch_requests()` that submits all requests at once, then the second call to `tick(3)` is stuck in infinite loop.

When I added a debug print in the loop that calls `tick(3)`, then I saw:

```
23: Call for i = 0
23: Call for i = 1
<reactor stall errors>
```

src/core/fair_queue.cc
xemul: Because this variable becomes write-only, which effectively means unused. This place is obsoleted with #2294 anyway :(

src/core/io_queue.cc
xemul: Good point

include/seastar/core/internal/throttle.hh
xemul: Ah, yes. I fixed this comment in one of the rebases and lost it along the way

src/core/throttle.cc
xemul: Yup, good catch

tests/perf/fair_queue_perf.cc
xemul: Yes, this capacity can be any, since can_dispatch() always says "yes you can"

tests/unit/fair_queue_test.cc
xemul: This test fails because there's no (easy) way to tell dispatch loop to stop dispatching. So I marked this PR as draft to fix it later

src/core/io_queue.cc
xemul: True

src/core/fair_queue.cc
xemul: Yes. This per-tick threshold is completely removed with #2294

tests/unit/fair_queue_test.cc
xemul: ACK, fails for me too

src/core/throttle.cc
xemul: I'm not sure if there's any set of rules on how to manipulate SEASTAR_MODULE and related. @tchaikov works on modules support, he can shed more light on this

src/core/fair_queue.cc
xemul: Yes, I will need to update header name too

src/core/throttle.cc
tchaikov: @pwrobelse hi Patryk, thank you for bringing this up.

`seastar.cc` includes both the global module fragment and the purview of the `seastar` module. we put the implementation into the module implementation units which starts with `module seastar`, for instance, see `fair_queue.cc` as you pointed out.

when it comes to `throttle.cc`, it needs to access the seastar module, so i think we would need to start it with `module seastar;` instead of `#include <seastar/core/internal/throttle.hh>` when C++20 modules is enabled, which is checked using `#ifdef SEASTAR_MODULE` in general.

> I'm not sure if there's any set of rules on how to manipulate SEASTAR_MODULE and related.

@xemul good point. i will note down this. could you suggest which document i should put this in? as its target audience will be library developer, instead of library user, the closest one i can find is `coding-style.md`. but this document focuses on coding style, it's more about taste, instead of the guideline on the general design or a perspective of the code structure. shall i create another document on C++20 modules?

xemul:
Closing in favor of #2820

PR #2334, opened 2024-07-06
scripts/perftune.py: get rid of "SyntaxWarning: invalid escape sequence"

vladzcloudius:
The interpreter was complaining about the unnecessary escaping in regular expressions. Let's remove that escaping and switch to using raw strings.


tchaikov (approved):
lgtm

PR #2337, opened 2024-07-06
Sanitize reactor configuration on start

xemul:
Reactor initialization is pretty dispersed, which is natural taking its complexity. However, the way reactor::configure() works has some room for improvement. In particular -- this method initializes a bunch of on-board simple bool/int/duration/etc. members soon after the reactor is constructed. And some of such members are initialized construction-time with the help of reactor_config.

This PR removes most of the re-initialization happening in reactor::configure() and de-bloats the reactor class itself. For that, those simple members onto reactor_config, the instance of reactor_config is initialized once and is copied into reactor constructor.

### DIFF
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -29,6 +29,7 @@ namespace seastar {
 
 /// \cond internal
 struct reactor_config {
+    bool handle_sigint = true;
     bool auto_handle_sigint_sigterm = true;
     unsigned max_networking_aio_io_control_blocks = 10000;
 };
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -24,11 +24,13 @@
 #include <seastar/util/program-options.hh>
 #include <seastar/util/memory_diagnostics.hh>
 #include <seastar/util/modules.hh>
+#include <seastar/core/scheduling.hh>
 
 namespace seastar {
 
 /// \cond internal
 struct reactor_config {
+    sched_clock::duration task_quota;
     bool handle_sigint = true;
     bool auto_handle_sigint_sigterm = true;
     unsigned max_networking_aio_io_control_blocks = 10000;
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -31,6 +31,7 @@ namespace seastar {
 /// \cond internal
 struct reactor_config {
     sched_clock::duration task_quota;
+    std::chrono::nanoseconds max_poll_time;
     bool handle_sigint = true;
     bool auto_handle_sigint_sigterm = true;
     unsigned max_networking_aio_io_control_blocks = 10000;
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -32,6 +32,7 @@ struct reactor_config {
     bool handle_sigint = true;
     bool auto_handle_sigint_sigterm = true;
     unsigned max_networking_aio_io_control_blocks = 10000;
+    bool force_io_getevents_syscall = false;
     bool kernel_page_cache = false;
     bool have_aio_fsync = false;
 };
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -36,6 +36,7 @@ struct reactor_config {
     bool kernel_page_cache = false;
     bool have_aio_fsync = false;
     unsigned max_task_backlog = 1000;
+    bool strict_o_direct = true;
     bool bypass_fsync = false;
 };
 /// \endcond
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -33,6 +33,7 @@ struct reactor_config {
     bool auto_handle_sigint_sigterm = true;
     unsigned max_networking_aio_io_control_blocks = 10000;
     bool kernel_page_cache = false;
+    bool have_aio_fsync = false;
 };
 /// \endcond
 
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -35,6 +35,7 @@ struct reactor_config {
     bool force_io_getevents_syscall = false;
     bool kernel_page_cache = false;
     bool have_aio_fsync = false;
+    unsigned max_task_backlog = 1000;
 };
 /// \endcond
 
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -36,6 +36,7 @@ struct reactor_config {
     bool kernel_page_cache = false;
     bool have_aio_fsync = false;
     unsigned max_task_backlog = 1000;
+    bool bypass_fsync = false;
 };
 /// \endcond
 
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -41,6 +41,7 @@ struct reactor_config {
     unsigned max_task_backlog = 1000;
     bool strict_o_direct = true;
     bool bypass_fsync = false;
+    bool no_poll_aio = false;
 };
 /// \endcond
 
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -32,6 +32,7 @@ struct reactor_config {
     bool handle_sigint = true;
     bool auto_handle_sigint_sigterm = true;
     unsigned max_networking_aio_io_control_blocks = 10000;
+    bool kernel_page_cache = false;
 };
 /// \endcond
 
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -330,7 +330,6 @@ private:
     std::atomic<bool> _sleeping alignas(seastar::cache_line_size){0};
     pthread_t _thread_id alignas(seastar::cache_line_size) = pthread_self();
     bool _strict_o_direct = true;
-    bool _bypass_fsync = false;
     std::atomic<bool> _dying{false};
     gate _background_gate;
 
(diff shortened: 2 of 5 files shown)


xemul:
upd:
- mark reactor::_config const

tchaikov:
will take a look later today.

include/seastar/core/reactor_config.hh
```diff
@@ -29,6 +29,7 @@ namespace seastar {
 
 /// \cond internal
 struct reactor_config {
+    bool handle_sigint = true;
     bool auto_handle_sigint_sigterm = true;
```
avikivity, line 36: What's the difference between handle_sigint and auto_handle_sigint_sigterm?

The patch is just code movement, but it's moving into a public interface, we it's better if we understand the difference.

include/seastar/core/reactor_config.hh
```diff
@@ -32,6 +32,7 @@ struct reactor_config {
     bool handle_sigint = true;
     bool auto_handle_sigint_sigterm = true;
     unsigned max_networking_aio_io_control_blocks = 10000;
+    bool kernel_page_cache = false;
```
avikivity, line 39: It's good to move these configurations to reactor_config, this will allow applications to drop the use of Seastar provided boost::program_options eventually, or use them as just helpers.

include/seastar/core/reactor_config.hh
```diff
@@ -33,6 +33,7 @@ struct reactor_config {
     bool auto_handle_sigint_sigterm = true;
     unsigned max_networking_aio_io_control_blocks = 10000;
     bool kernel_page_cache = false;
+    bool have_aio_fsync = false;
```
avikivity, line 40: This isn't reactor configuration because the calculation includes kernel_supports_aio_fsync().

The configuration is whether we want to use aio fsync if it's available or not. have_aio_fsync is the result of aio_fsync && kernel_supports, this isn't configuration the user can provide.

avikivity: Though I see reactor_config is marked internal. But is it really?

avikivity: Ah, maybe it's okay. We have app_template::config from which reactor_config is derived.

avikivity:
I don't like the const/mutable bits. const+mutable is useful where the object is logically const but needs some physically mutable internals to function, like a cache for an expensive pure calculation. But here the object is not logically const. If one passes it by value rather than by reference, the result will change.

include/seastar/core/reactor_config.hh
xemul: > What's the difference between handle_sigint and auto_handle_sigint_sigterm?

The former is sub for the latter:

```c++
    if (_id == 0 && _cfg.auto_handle_sigint_sigterm) {
       if (_cfg.handle_sigint) {
          _signals.handle_signal_once(SIGINT, [this] { stop(); });
       }
       _signals.handle_signal_once(SIGTERM, [this] { stop(); });
    }
```

> The patch is just code movement, but it's moving into a public interface, we it's better if we understand the difference.

The reactor_config is not a public interface, reactor_options is

xemul:
> I don't like the const/mutable bits. const+mutable is useful where the object is logically const but needs some physically mutable internals to function, like a cache for an expensive pure calculation. But here the object is not logically const. If one passes it by value rather than by reference, the result will change.

The reactor_config is used (by reactor) like this

```c++
class reactor {
    ...
    const reactor_config _cfg;
    ...
};
```

which is exactly how it's used -- once set, configuration doesn't change. There are two exceptions -- strict_o_direct and bypass_fsync. There are API calls that allow tweaking it runtime.

xemul:
upd:
- reverted config [const-marking](https://github.com/scylladb/seastar/pull/2337#issuecomment-2216759894)

include/seastar/core/reactor_config.hh
tchaikov: > Though I see reactor_config is marked internal. But is it really?

i think it is, unless a seastar application creates an instance of `reactor` manually, instead of using `app_template`.

avikivity: It's not possible to create reactors directly.  The entry point is smp::configure but it's too difficult to use.

bhalevy:
@xemul it looks like this change is wrecking havoc in scylla when updating the seastar submodule:
For example, with memtable_test (in debug mode):
```
/home/bhalevy/dev/scylla/seastar/include/seastar/core/reactor.hh:680:12: runtime error: reference binding to null pointer of type 'reactor'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/bhalevy/dev/scylla/seastar/include/seastar/core/reactor.hh:680:12 
Aborting.
Backtrace:
  0x27b10b9
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x35a3ec7
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x35a3b55
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x34048fb
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x3435a85
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x3517b39
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x3517e5b
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x3517c8e
  /lib64/libc.so.6+0x40cff
  /lib64/libc.so.6+0x994a3
  /lib64/libc.so.6+0x40c4d
  /lib64/libc.so.6+0x28901
  0x282dc0a
  0x282bd24
  0x2840e0b
  0x2841836
  0x284147d
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x2e7c224
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x3408873
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x33f5cac
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x3450aad
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x3519e26
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x35189c0
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x35188b0
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x351801c
  /home/bhalevy/dev/scylla/build/debug/seastar/libseastar.so+0x3205818
  0x28037ac
  /lib64/libc.so.6+0x97506
  /lib64/libc.so.6+0x11b40b
/home/bhalevy/.local/bin/scylla_unit_test: line 113: 3112560 Aborted                 (core dumped) ${cmd}
```

Decoded:
```
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, 
    no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007f04f24a8513 in __pthread_kill_internal (threadid=<optimized out>, signo=6)
    at pthread_kill.c:78
#2  0x00007f04f244fc4e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f04f2437902 in __GI_abort () at abort.c:79
#4  0x000000000282dc0b in __sanitizer::Abort() ()
#5  0x000000000282bd25 in __sanitizer::Die() ()
#6  0x0000000002840e0c in __ubsan::ScopedReport::~ScopedReport() ()
#7  0x0000000002841837 in handleTypeMismatchImpl(__ubsan::TypeMismatchData*, unsigned long, __ubsan::ReportOptions) ()
#8  0x000000000284147e in __ubsan_handle_type_mismatch_v1 ()
#9  0x00007f04f747c225 in seastar::engine () at ./seastar/include/seastar/core/reactor.hh:680
#10 0x00007f04f7a08874 in seastar::pollable_fd::pollable_fd (this=0x7f04eac09d90, fd=..., 
    speculate=...) at ./build/debug/seastar/./seastar/src/core/reactor.cc:1670
#11 0x00007f04f79f5cad in seastar::reactor::reactor (this=0x524000022100, 
    smp=std::shared_ptr<seastar::smp> (empty) = {...}, alien=..., id=1, rbs=..., cfg=...)
    at ./build/debug/seastar/./seastar/src/core/reactor.cc:1053
#12 0x00007f04f7a50aae in seastar::smp::allocate_reactor (this=0x51200000c050, id=1, rbs=..., 
    cfg=...) at ./build/debug/seastar/./seastar/src/core/reactor.cc:3947
#13 0x00007f04f7b19e27 in seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_0::operator()() const (this=0x51200000c940)
    at ./build/debug/seastar/./seastar/src/core/reactor.cc:4516
```

(This goes back to our CI.  If we would test seastar PRs against scylla head-of-line we would find out about those regressions before merging to seastar/master)

tchaikov:
> (This goes back to our CI. If we would test seastar PRs against scylla head-of-line we would find out about those regressions before merging to seastar/master)

yeah, FWIW, currently we have a regular test for building latest seastar master HEAD with scylla, but we don't run the tests. because otherwise, we would be running out of the space on the free runner provided by github.

avikivity:
@xemul please revert if you can't come up with a quick fix.

bhalevy:
> @xemul please revert if you can't come up with a quick fix.

ping @xemul please tend to this ASAP so we can update the seastar submodule in scylla.

tchaikov:
> > @xemul please revert if you can't come up with a quick fix.
> 
> ping @xemul please tend to this ASAP so we can update the seastar submodule in scylla.

@bhalevy seems @xemul created #2368 to address this regression.

bhalevy:
> > > @xemul please revert if you can't come up with a quick fix.
> > 
> > 
> > ping @xemul please tend to this ASAP so we can update the seastar submodule in scylla.
> 
> @bhalevy seems @xemul created #2368 to address this regression.

Ah, it would have helped if it referenced this PR

PR #2339, opened 2024-07-08
io_queue: rename io_desc_read_write to queued_io_request_completion

pwrobelse:
This patch is a part of the preparation to route discard requests via I/O scheduler.

The mentioned 'io_desc_read_write' is used to track the completion of read and write requests submitted to 'io_sink'. It implements 'io_completion' interface -- its functions are invoked when the processing of a request via 'reactor_backend' is finished.

In spite of the fact that 'fallocate()' is not supported by Linux AIO and therefore, discards will not be submitted to 'io_sink' and processed by 'reactor_backend', discard requests will need to notify 'io_queue' about completion in the same way that reads and writes do. Thus, to avoid code duplication 'io_desc_read_write' is renamed to 'queued_io_request_completion' as it will be used by discard requests.

### DIFF
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -65,7 +65,7 @@ class io_sink;
 using shard_id = unsigned;
 using stream_id = unsigned;
 
-class io_desc_read_write;
+class queued_io_request_completion;
 class queued_io_request;
 class io_group;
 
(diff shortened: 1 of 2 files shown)


xemul:
In #2319 you merged this class with queued_io_request. Since it's not yet clear what the final solution is going to look like, let's postpone with this rename for a while

pwrobelse:
> In #2319 you merged this class with queued_io_request. Since it's not yet clear what the final solution is going to look like, let's postpone with this rename for a while

Hi, I have just closed that PR -- merging those classes was not required. Also, you pointed out some problems with it. I added rationale to the comment that closes the PR#2319.

pwrobelse:
Closing in favor of: https://github.com/scylladb/seastar/pull/2386

Issue #2238, opened 2024-05-11
High lowres clock drift > 1s in debug mode


nvartolomei:
In redpanda we have a [test](https://github.com/redpanda-data/redpanda/issues/13275) which began failing very often in debug mode since latest seastar upgrade (CPU is busy with other processes too)

The test gets the current time using the lowres clock and sets a deadline +1 second for establishing a tcp connection. Because of high lowres clock drift, although the time it takes to run the tasks is in milliseconds we do hit the 1s deadline.

I believe this change in behavior is caused by this reactor change #2007.

We preempt every 64 calls to scheduler_need_preempt (every 64 tasks?) under the argument that "latency is not important in debug mode" but that seems to be too much in some cases which use the lowres clock since it is only updated when we break out of the reactor loop.

While we are guilty to have some reactor stalls in this testing code ~100ms, I don't believe that should be an excuse to let lowres clock drift to 1s+. If we were in a seastar::loop with each iteration taking 25ms (stall warning threshold), 64 calls are already 1.6s.

I see a few possible options:

* Switch to std::chrono clocks;
* Change user code to have very lax expectations on lowres clock everywhere even though it is advertised to have ~task quota granularity;
* Change how lowres clock is updated in debug mode;
* Change the approach introduced in #2007 to break out of the reactor task queue execution if any of the following conditions are true:
    * 64 (or any other N) calls to scheduler_need_preempt
    * need_preempt() == true as it is available in release mode (aka np.head != np.tail) so that it respects the task quota even
* ???

Appreciate any thoughts or ideas on the above 🙏

nvartolomei:
Maybe something like

```diff
diff --git a/include/seastar/core/preempt.hh b/include/seastar/core/preempt.hh
index 537be5b5..21882aa9 100644
--- a/include/seastar/core/preempt.hh
+++ b/include/seastar/core/preempt.hh
@@ -49,11 +49,9 @@ inline const preemption_monitor*& get_need_preempt_var() {

 void set_need_preempt_var(const preemption_monitor* pm);

-}
-
-SEASTAR_MODULE_EXPORT
-inline bool need_preempt() noexcept {
-#ifndef SEASTAR_DEBUG
+inline
+bool
+monitor_need_preempt() {
     // prevent compiler from eliminating loads in a loop
     std::atomic_signal_fence(std::memory_order_seq_cst);
     auto np = internal::get_need_preempt_var();
[...]

PR #2340, opened 2024-07-09
reactor: respect preempt requests in debug mode

nvartolomei:
In [e0d515b6 reactor: poll less frequently in debug mode][1] reactor preemption logic in debug mode was changed to improve performance. It also introduced a problem causing `seastar::lowres_clock` to drift unbounded.

The code doc states that `sestar::lowres_clock` "is a monotonic clock with a granularity of ~task_quota". We want that claim to be respected.

To achieve that we change the `scheduler_need_preempt` to respect the preemption requests which are triggered e.g. when task quota is up.

[1]: https://github.com/scylladb/seastar/commit/e0d515b6cf96129cebdcbb1b031df90eb815e7fa

Closes https://github.com/scylladb/seastar/issues/2238

### DIFF
--- a/tests/unit/lowres_clock_test.cc
+++ b/tests/unit/lowres_clock_test.cc
@@ -21,16 +21,17 @@
 
 #include <seastar/testing/test_case.hh>
 
+#include <seastar/core/coroutine.hh>
 #include <seastar/core/do_with.hh>
 #include <seastar/core/lowres_clock.hh>
 #include <seastar/core/sleep.hh>
 #include <seastar/core/loop.hh>
+#include <seastar/util/later.hh>
 
 #include <ctime>
 
-#include <algorithm>
-#include <array>
 #include <chrono>
+#include <thread>
 
 using namespace seastar;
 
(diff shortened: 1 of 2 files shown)


tests/unit/lowres_clock_test.cc
```diff
@@ -52,6 +52,30 @@ SEASTAR_TEST_CASE(steady_clock_sanity) {
...
+        return ::seastar::yield().then([&t1] {
+            auto const elapsed = lowres_clock::now() - t1;
+            auto const minimum_elapsed = 0.9 * sleep_duration;
+
+            BOOST_REQUIRE(elapsed >= minimum_elapsed);
+
+            return make_ready_future<>();
+        });
+    });
+}
```
avikivity, line 77: Please write the test as a coroutine (or a Seastar thread), now that C++20 and above is mandatory there's no reason for introducing new code in continuation style.

nvartolomei: Fixed.

PR #2341, opened 2024-07-09
abort_source: subscription: allow calling on_abort explicitly

bhalevy:
Today, when calling `subscribe` after abort has alredy been requested on an `abort_source` we return a default- initialized subscription that signifies that subscription failed.

With this change, `subscribe` would return an unlinked subscription, that still holds the required callback. `bool(subscription)` stil elaluates to `false` in the same way, since the subscription is unlinked, but, it supports `on_abort`, that calls the subscribed
callback, at a later time, the same way as if
abort is requested at a later time.

Thise allows the users of this interface to
implement a unified abort path that can either be called by the abort_source, if subscribe happened before
abort was requested, or called by the user `on_abort` call, if abort weas never requested, or if it was requested before `subscribe` was called.

### DIFF
--- a/include/seastar/core/abort_source.hh
+++ b/include/seastar/core/abort_source.hh
@@ -66,21 +66,35 @@ public:
     class subscription : public bi::list_base_hook<bi::link_mode<bi::auto_unlink>> {
         friend class abort_source;
 
-        subscription_callback_type _target;
+        subscription_callback_type _target = noop_handler;
 
         explicit subscription(abort_source& as, subscription_callback_type target)
                 : _target(std::move(target)) {
+          if (!as.abort_requested()) {
             as._subscriptions.push_back(*this);
+          }
         }
 
         struct naive_cb_tag {}; // to disambiguate constructors
         explicit subscription(naive_cb_tag, abort_source& as, naive_subscription_callback_type naive_cb)
                 : _target([cb = std::move(naive_cb)] (const std::optional<std::exception_ptr>&) noexcept { cb(); }) {
+          if (!as.abort_requested()) {
             as._subscriptions.push_back(*this);
+          }
         }
 
+    public:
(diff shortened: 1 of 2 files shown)


include/seastar/core/abort_source.hh
```diff
@@ -140,9 +149,6 @@ public:
                   std::is_nothrow_invocable_r_v<void, Func>)
     [[nodiscard]]
     optimized_optional<subscription> subscribe(Func&& f) {
-        if (abort_requested()) {
-            return { };
-        }
         if constexpr (std::is_invocable_v<Func, std::exception_ptr>) {
```
avikivity, line 152: Isn't the documentation no longer correct?

bhalevy: Good catch. I'll address that in the next version.

xemul (approved):

include/seastar/core/abort_source.hh
bhalevy: fixed in v2

bhalevy:
v2: improve documentation

bhalevy:
@xemul is there no CI?

xemul:
> @xemul is there no CI?

@bhalevy , I don't follow. Looks like PR's CI was started with delay, or do you mean something else?

bhalevy:
> > @xemul is there no CI?
> 
> @bhalevy , I don't follow. Looks like PR's CI was started with delay, or do you mean something else?

Ah no.  I looked after force pushing a new version, so it probably erased the previous status and showed the new CI invocation with a delay of few minutes, as you mentioned.
I was concerned that CI didn't run even for v1...
The bottom line is that it's running now! :)

Deexie (approved):

bhalevy:
@avikivity / @scylladb/seastar-maint can you please merge this PR?

tchaikov:
@avikivity / @scylladb/seastar-maint  ping?

PR #2344, opened 2024-07-10
Undefine `_FORTIFY_SOURCE` only on what's needed

dotnwat:
Seastar cmake build contains

```
target_compile_options (seastar
  PUBLIC
    -U_FORTIFY_SOURCE)
```

Which disables FORTIFY_SOURCE on all Seastar targets and is inherited by anything depending on Seastar.

This PR undefines this only on src/core/thread.cc which is apparently the only thing that needs it, and avoids leaking this out to dependents.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -778,6 +778,10 @@ add_library (seastar
   src/websocket/server.cc
   )
 
+# We disable _FORTIFY_SOURCE because it generates false positives with longjmp() (src/core/thread.cc)
+set_source_files_properties(src/core/thread.cc
+  PROPERTIES COMPILE_FLAGS -U_FORTIFY_SOURCE)
+
 add_library (Seastar::seastar ALIAS seastar)
 
 add_dependencies (seastar
@@ -898,12 +902,6 @@ include (CTest)
 #
 # To disable -Werror, pass -Wno-error to Seastar_CXX_FLAGS.
 #
-# We disable _FORTIFY_SOURCE because it generates false positives with longjmp() (src/core/thread.cc)
-#
-
-target_compile_options (seastar
-  PUBLIC
-    -U_FORTIFY_SOURCE)
 
 target_compile_definitions(seastar
   PUBLIC


CMakeLists.txt
```diff
@@ -505,6 +505,11 @@ seastar_generate_protobuf (
   IN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/src/proto/metrics2.proto
   OUT_DIR ${Seastar_GEN_BINARY_DIR}/src/proto)
 
+set(FORTIFY_BUILD_TYPES "Release;RelWithDebInfo;Dev")
+if(CMAKE_BUILD_TYPE IN_LIST FORTIFY_BUILD_TYPES)
+  add_definitions(-D_FORTIFY_SOURCE=2)
+endif()
```
dotnwat, line 511: This is added in a second commit that we can drop. It's only real purpose here is to demonstrate that the tests pass with `_FORTIFY_SOURCE=2` when it's undefined on `src/core/thread.cc`.

CMakeLists.txt
```diff
@@ -898,12 +902,6 @@ include (CTest)
 #
 # To disable -Werror, pass -Wno-error to Seastar_CXX_FLAGS.
 #
-# We disable _FORTIFY_SOURCE because it generates false positives with longjmp() (src/core/thread.cc)
-#
-
-target_compile_options (seastar
```
xemul, line 904: To educate myself.
Before the patch FORTIFY_SOURCE was off for the whole seastar. Now it's off for thread.cc only, is that correct?
If so, do we expect any performance penalties once it's enabled for most of the seastar?

tchaikov: > To educate myself. Before the patch FORTIFY_SOURCE was off for the whole seastar. 

yes, it's undefined for seastar, and also for the projects which use seastar via its .pc file and those which use seastar via its CMake config files.

> Now it's off for thread.cc only, is that correct?

correct.

> If so, do we expect any performance penalties once it's enabled for most of the seastar?

i think so. as the fortified version of glibc function do perform extra checks. i think, the next question would be "how much?". as an application can always override the level of the fortification level at the compile time using something like `-U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3`. so the answer might depend on what the compile options are used.

if the overhead is not negligible, or we don't have a benchmark to justify the change, i think the safe move is probably to 

1. keep this `-U_FORTIFY_SOURCE` private and apply it only to `src/core/thread.cc` as in this commit
2.  add this option back globally in Seastar as a private option when compiled in Debug or Sanitize modes.

tchaikov: sorry, i haven't read the 2nd commit. so it has been implemented. thanks!

xemul: 2nd commit does this

```patch
+set(FORTIFY_BUILD_TYPES "Release;RelWithDebInfo;Dev")
+if(CMAKE_BUILD_TYPE IN_LIST FORTIFY_BUILD_TYPES)
+  add_definitions(-D_FORTIFY_SOURCE=2)
+endif()
+
```

I see the final result as -- set FORTIFY_SOURCE=2 for Release build. Is level 2 considered to be "no penalty"?

CMakeLists.txt
tchaikov: this works if we use a single-configuration generator. but it does not work if we use a multi-config generator to generate the rules. if we use 
```
cmake -G "Ninja Multi-Config" 
```
to generate `build.ninja`, we cannot get the "current" build mode via `CMAKE_BUILD_TYPE` anymore. as this variable is determined at the configure time. but with multi-config generator, one can pick which configuration at build time, like:
```
cmake --build build --config Dev --target seastar
```

so, i'd suggest use the generator expression, which is expanded to different value specific to each build configuration, so we could use something like:
```cmake
  PRIVATE
    $<$<IN_LIST:$<CONFIG>,${FORTIFY_BUILD_TYPES}>:_FORTIFY_SOURCE=2)
```

in which,

1. use `PRIVATE` so we don't force the parent project's to use the same fortification level as seastar does
2. instead of using `add_definitions()` use `target_compile_definitions()`, more specific this way
3. use generator expression as explained above.

CMakeLists.txt
tchaikov: no, i don't think so. it incurs less penalty than level 3. to be on the safe side, i think we should undefine this macro for `Release` and `RelWithDebInfo`. as to Dev, we can enable level 2 or even level 3.

CMakeLists.txt
dotnwat: @tchaikov i'm not suggesting that we leave this here, it is merely for testing. do you want fortify sources enabled?

CMakeLists.txt
avikivity: I agree Seastar should leave it to the application and not change defaults.

Can we just add #undef (plus comment explaining why) to the source file?

dotnwat:
@tchaikov yes, thanks for the ping. I can push an update tomorrow. But please feel free to cherry-pick the first commit if you want.

dotnwat:
> @dotnwat hi Noah, do you plan to continue working on this PR? i still think `_FORTIFY_SOURCE` is a nice feature to identify runtime issues. and the first commit would suffice. seastar applications can opt out of this feature if desired.

Dropped the second commit.

tchaikov:
thank you Noah!

tchaikov (approved):
lgtm

tchaikov:

PR #2345, opened 2024-07-10
when_all: stop using deprecated std::aligned_union_t

tchaikov:
since [P1413R3](https://wg21.link/p1413r3) was accepted by C++23, std::aligned_union_t was deprecated in this standard. so let's be prepared for this. as suggested by the proposal, we are replacing `std::aligned_union_t` with an array of `std::byte`.


tchaikov:
two of the CI workflows failed due to infra issue.

PR #2346, opened 2024-07-10
metrics: Use this_shard_id unconditionally

StephanDollberg:
https://github.com/scylladb/seastar/commit/a57b17c8f4cba96aa8ac09741f0ca8f07a208e0a changed the way `local_engine`
gets set. It's now only being set after the reactor has been fully
constructed.

This breaks a bunch of metrics that are being registered from the
reactor constructor (main scheduling group metrics, stall detector,
possibly more) as the metrics subsystem uses `engine_is_ready` which
internally checks whether `local_engine` is set. With this not being the
case it would set the shard id to 0 for all shards.

Remove the extra check to `engine_is_ready` as the shard id is
initialized before the reactor is being constructed so this is safe to
use directly.

### DIFF
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -392,11 +392,7 @@ foreign_ptr<values_reference> get_values() {
 
 
 instance_id_type shard() {
-    if (engine_is_ready()) {
-        return to_sstring(this_shard_id());
-    }
-
-    return sstring("0");
+    return to_sstring(this_shard_id());
 }
 
 void impl::update_metrics_if_needed() {


StephanDollberg:
@tchaikov did you have a chance to look at this?

tchaikov:
> @tchaikov did you have a chance to look at this?

not yet, but will do early tomorrow.

tchaikov (approved):
lgtm.

tchaikov:
@vladzcloudius hi Vlad, this change practically reverts your change of ff098c84ea7a285400a27dddeeeeb1ec55e5024b. do  you mind taking a look ?

avikivity (approved):

vladzcloudius (approved):

StephanDollberg:
Thanks folks, could someone hit the merge button?

tchaikov:

PR #2347, opened 2024-07-10
src/core: remove unnecessary includes

tchaikov:
these unused includes are identified by clangd. see https://clangd.llvm.org/guides/include-cleaner#unused-include-warning for more details on the "Unused include" warning.


include/seastar/core/future.hh
```diff
@@ -23,7 +23,6 @@
 
 #ifndef SEASTAR_MODULE
 #include <cassert>
-#include <concepts>
 #include <cstdlib>
```
avikivity, line 26: How is this not required? We have 

```c++
    template <typename Func, typename Result = typename internal::future_result<Func, T>::future_type>
    requires std::invocable<Func, T>
                 || (std::same_as<void, T> && std::invocable<Func>)
```

Below.

tchaikov:
v2:

- add `#include <concepts>` back to `future.hh`  and `memory.cc`
- add `#include <array>` to `io_queue.cc`
- add `#include <unordered_map>` and `#include <vector>` to `prefault,hh`

avikivity:
How do we know the includes are right? I don't want to review the output of some tool. Can we trust the tool?

tchaikov:
> How do we know the includes are right?

in the first version, i just trusted the output of clangd. and applied the suggestions, and then adjusted so that the CI passed

in the latest revision, i 

1. applied the suggestions, 
2. audited the changes by searching for symbols provided by the removed includes
3. preserved the ones which provide the used symbols / declarations
4. and make sure the CI passes

> I don't want to review the output of some tool. 

it's not just output of this tool anymore.

> Can we trust the tool?

no. but i tried to selectively apply its suggestions.

PR #2348, opened 2024-07-11
perftune: tune tcp_mem

avikivity:
tcp_mem defaults to 9% of memory, while Seastar defaults to 93% of memory. These sum to 102% which leads to OOM.

Fix by tuning tcp_mem, by default to 3% of memory. The low/pressure/max ratios are the same as those selected by the kernel.

### DIFF
--- a/scripts/perftune.py
+++ b/scripts/perftune.py
@@ -21,6 +21,8 @@ import urllib.request
 import yaml
 import platform
 import shlex
+import psutil
+import mmap
 
 dry_run_mode = False
 def perftune_print(log_msg, *args, **kwargs):
@@ -605,6 +607,17 @@ class NetPerfTuner(PerfTunerBase):
         # Increase the maximum number of remembered connection requests, which are still
         # did not receive an acknowledgment from connecting client.
         fwriteln_and_log('/proc/sys/net/ipv4/tcp_max_syn_backlog', '4096')
+
+        self.tune_tcp_mem()
+
+    def tune_tcp_mem(self):
+        page_size = mmap.PAGESIZE
+        total_mem = psutil.virtual_memory().total
+        # We only tune for physical memory since tcp_mem is virtualized
+        def to_pages(bytes):
+            return math.ceil(bytes / page_size)
+        max = total_mem * self.args.tcp_mem_fraction
+        fwriteln_and_log('/proc/sys/net/ipv4/tcp_mem', f"{to_pages(max / 2)} {to_pages(max * 2/3)} {to_pages(max)}")
 
     def nic_is_bond_iface(self, nic):
         return self.__nic_is_bond_iface.get(nic, False)
@@ -1565,6 +1578,10 @@ class TuneModes(enum.Enum):
     def names():
         return list(TuneModes.__members__.keys())
 
+# Seastar defaults to allocating 93% of physical memory. The kernel's default allocation for TCP is ~9%. This adds up
+# to 102%. Reduce the TCP allocation to 3% to avoid OOM.
+default_tcp_mem_fraction = 0.03
+
 argp = argparse.ArgumentParser(description = 'Configure various system parameters in order to improve the seastar application performance.', formatter_class=argparse.RawDescriptionHelpFormatter,
                                epilog=
 '''
@@ -1627,6 +1644,7 @@ argp.add_argument('--irq-core-auto-detection-ratio', help="Use a given ratio for
                                                           "CPU cores out of available according to a 'cpu_mask' value."
                                                           "Default is 16",
                   type=int, default=16, dest='cores_per_irq_core')
+argp.add_argument('--tcp-mem-fraction', default=default_tcp_mem_fraction, type=float, help="Fraction of total memory to allocate for TCP buffers")
 
 def parse_cpu_mask_from_yaml(y, field_name, fname):
     hex_32bit_pattern='0x[0-9a-fA-F]{1,8}'


vladzcloudius:
On my U24.04 machine:

```
$ cat /proc/sys/net/ipv4/tcp_mem                                                                                                                                                                                            
765579	1020775	1531158
$ free                                                                                                                                                                                                                      
               total        used        free      shared  buff/cache   available
Mem:        65599452    21581832    35377072     3084664    12456680    44017620
Swap:       67108860           0    67108860
```

| Name | value |
|----------|----------|
|tcp_mem max |            1531158|
|9% of RAM| 5903951|

@avikivity What made you believe that tcp_mem defaults to 9% of RAM?

scripts/perftune.py
```diff
@@ -1627,6 +1638,7 @@ def names():
                                                           "CPU cores out of available according to a 'cpu_mask' value."
                                                           "Default is 16",
                   type=int, default=16, dest='cores_per_irq_core')
+argp.add_argument('--tcp-mem-fraction', default=0.03, type=float, help="Fraction of total memory to allocate for TCP buffers")
```
tchaikov, line 1641: 0.03 looks like a magic number to me, how about adding a comment to reference `resource::calculate_memory()`, to explain that seastar reserves 7% for the rest of the system, and to be on the safe side, we use a number less than 7%?

vladzcloudius:
@avikivity doesn't seastar override `malloc`, `calloc` and `new` syscall making corresponding sockets buffers (those created from inside a seastar application) allocations go via the seastar allocator?

scripts/perftune.py
avikivity: yes

avikivity:
> @avikivity doesn't seastar override `malloc`, `calloc` and `new` syscall making corresponding sockets buffers (those created from inside a seastar application) allocations go via the seastar allocator?

Seastar does override malloc/new, but the kernel uses its own allocators.

avikivity:
v2: document the 3% choice

scripts/perftune.py
```diff
@@ -606,6 +606,18 @@ def tune(self):
         # did not receive an acknowledgment from connecting client.
         fwriteln_and_log('/proc/sys/net/ipv4/tcp_max_syn_backlog', '4096')
 
+        self.tune_tcp_mem()
+
+    def tune_tcp_mem(self):
+        page_size = os.sysconf("SC_PAGE_SIZE")
+        phys_mem = os.sysconf("SC_PAGE_SIZE") * page_size
```
vladzcloudius, line 613: This is a strange construct: you are effectively making `phys_mem == os.sysconf("SC_PAGE_SIZE")^2`.

I think you meant to multiply the PAGE_SIZE by NUM_PAGES.

However I don't understand why even this is needed since `tcp_mem` is configured in PAGEes anyway.

vladzcloudius (changes requested):

scripts/perftune.py
vladzcloudius: One way to do that is using `psutil.virtual_memory().total` (total memory in bytes):

```
phys_mem = psutil.virtual_memory().total`
```

And our currently shipped relocatable Python3.11 ships psutil.

The rest of your code would remain the same.

vladzcloudius (changes requested):

scripts/perftune.py
vladzcloudius: A nitpick:
If you decide to use `psutil` as above I'd also recommend changing the variable name to something else because otherwise a construct "phys_mem = ... virtual_mem.total` is confusing.

Something like "total_mem" would be less  contradictory IMO.

avikivity:
> > > @avikivity doesn't seastar override `malloc`, `calloc` and `new` syscall making corresponding sockets buffers (those created from inside a seastar application) allocations go via the seastar allocator?
> > 
> > 
> > Seastar does override malloc/new, but the kernel uses its own allocators.
> 
> Right. My bad - I assumed (dunno why) that the socket buffer lives in the userspace.

I guess you're used to our userspace tcp stack.

avikivity:
v2: replace os.sysconf by nicer alternatives

scripts/perftune.py
```diff
@@ -606,6 +608,18 @@ def tune(self):
...
+        self.tune_tcp_mem()
+
+    def tune_tcp_mem(self):
+        page_size = mmap.PAGESIZE
+        total_mem = psutil.virtual_memory().total
+        # We only tune for physical memory since tcp_mem is virtualized
+        def to_pages(bytes):
+            return math.ceil(bytes / page_size)
+        max = total_mem * self.args.tcp_mem_fraction
+        with open('/proc/sys/net/ipv4/tcp_mem', 'w') as f:
```
xemul, line 620: Other sysctls are tuned using `fwriteln_and_log()` helper, it's worth using one here as well

avikivity: Yes. It's also a way to prepare an "undo file" which can be applied on service shutdown. @vladzcloudius

avikivity:
v3: use fwriteln_and_log() instead of open-coding; retested

nyh:
I have a question, I hope it's not too stupid:
If by default Seastar allocates 93% of memory and the kernel allocates 9% of memory, why is the better solution to reduce the kernel's choice 3-fold instead of reducing a few percent from Seastar's allocation?

Also, changing perftune instead of Seastar means that users that do not run the perftune script will still have this problem.

So I have an idea: What if Seastar, on boot, calculated the amount of available memory _after_ reducing tcp_mem and used that as the amount of available memory? If we do this, Seastar will be able to handle whatever tcp_mem is set to. The default 9% might still be too much, and a waste, but at least we won't run out of memory. And perftune will be really about better tuning - not something necessary.

avikivity:
> I have a question, I hope it's not too stupid: If by default Seastar allocates 93% of memory and the kernel allocates 9% of memory, why is the better solution to reduce the kernel's choice 3-fold instead of reducing a few percent from Seastar's allocation?

The kernel's choice was was made in 2015 (torvalds/linux@b66e91ccbc34ebd5a2f90f9e1bc1597e2924a500), halving from the previous 19% as a first step to sanity. The reduction to 3% is the next step to sanity.

Dedicating 9% of memory to tcp is out of proportion. When a Seastar application uses it, it can be used for caching user data.

In regular applications it doesn't matter so much since if the reservation isn't used, it can be used by the kernel page cache or other caches, but that's not the case for Seastar applications.

> 
> Also, changing perftune instead of Seastar means that users that do not run the perftune script will still have this problem.

The solution is easy: run the script.

> 
> So I have an idea: What if Seastar, on boot, calculated the amount of available memory _after_ reducing tcp_mem and used that as the amount of available memory? If we do this, Seastar will be able to handle whatever tcp_mem is set to. The default 9% might still be too much, and a waste, but at least we won't run out of memory. And perftune will be really about better tuning - not something necessary.

That's a good idea, we already take into account some kernel-related information in the calculation, we could also do that. Then we'd change the formula from 93% of total memory to 96% of (total memory - tcp reservation) or something.

vladzcloudius (approved):
lgtm

scripts/perftune.py
vladzcloudius: That was a good point, @xemul. I forgot about that wrapper.
It's used by `--dry-run`.

nyh:
I merged this PR because I agree that we believe that 3% memory is enough for TCP on modern machines, we should set it so in perftune.py.

But I do believe that in the long run, it's better to make Seastar handle any setting of tcp_mem gracefully. The sin of not running perftune.py should not carry a death penalty (for the Seastar application) :-)

vladzcloudius:
@avikivity unfortunately this PR is incomplete: the new feature must be configurable via a configuration file (YAML) as well.

The functions that need to be patched are:
* `dump_config`
* `parse_options_file`

Please, consider sending a follow up.

PR #2349, opened 2024-07-12
coding style: allow lines up to 160 chars in length

lersek:
Avi writes,

> We promote long, explanatory names instead of acronyms. If we fold too
> much we end up taking a lot of vertical space and functions no longer
> fit on a screen.

Suggested-by: Avi Kivity <avi@scylladb.com>

### DIFF
--- a/coding-style.md
+++ b/coding-style.md
@@ -127,7 +127,7 @@ Whitespace around operators should match their precedence: high precedence = no 
 
 ## Long lines
 
-If a line becomes excessively long (>120 characters?), or is just complicated, break it into two or more lines.  The second (and succeeding lines) are _continuation lines_, and have a double indent:
+If a line becomes excessively long (>160 characters?), or is just complicated, break it into two or more lines.  The second (and succeeding lines) are _continuation lines_, and have a double indent:
 
 ```c++
     if ((some_condition && some_other_condition)


coding-style.md
```diff
@@ -127,7 +127,7 @@ Whitespace around operators should match their precedence: high precedence = no
 
 ## Long lines
 
-If a line becomes excessively long (>120 characters?), or is just complicated, break it into two or more lines.  The second (and succeeding lines) are _continuation lines_, and have a double indent:
+If a line becomes excessively long (>160 characters?), or is just complicated, break it into two or more lines.  The second (and succeeding lines) are _continuation lines_, and have a double indent:
```
tchaikov, line 130: is it possible to have a clang-format workflow to enforce this policy using `ColumnLimit`?

lersek: I'd be happy to investigate that, but I think several people on the team wouldn't appreciate it. In the internal thread, I asked *Why is the line limit guideline not consistently enforced?*, and several answers indicated it was more like a rule of thumb, not a hard rule.

dotnwat: We use clang-format on our entire tree, and enforce width limit of 80. I think you can set it to whatever width you want.

https://github.com/redpanda-data/redpanda/blob/dev/src/v/cloud_roles/request_response_helpers.cc#L95-L98

Then IndentWidth takes care of the indenting over on the wrapped lines.

PR #2352, opened 2024-07-16
sstring: prevent fmt from formatting sstring as a sequence

tchaikov:
since fmt 11, if fmt/ranges.h is included, it tries to format a sequence-like container as a sequence. this confuses the compiler, and prevents fmt from considering sstring a formattable, as this type now has two formatters:

- one is the specialization of fmt::formatter provided by sstring.hh
- another is the one provided by fmt/ranges.h

```
/home/kefu/.local/bin/clang++ -DFMT_SHARED -DSEASTAR_API_LEVEL=7 -DSEASTAR_BUILD_SHARED_LIBS -DSEASTAR_DEBUG -DSEASTAR_DEBUG_PROMISE -DSEASTAR_DEBUG_SHARED_PTR -DSEASTAR_DEFAULT_ALLOCATOR -DSEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT -DSEAST
AR_DEPRECATED_OSTREAM_FORMATTERS -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_NUMA -DSEASTAR_HAVE_SYSTEMTAP_SDT -DSEASTAR_HAVE_URING -DSEASTAR_LOGGER_COMPILE_TIME_FMT -DSEASTAR_LOGGER_TYPE
_STDOUT -DSEASTAR_PTHREAD_ATTR_SETAFFINITY_NP -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_SHUFFLE_TASK_QUEUE -DSEASTAR_SSTRING -DSEASTAR_STRERROR_R_CHAR_P -DSEASTAR_THREAD_STACK_GUARDS -DSEASTAR_TYPE_ERASE_MORE -Dseastar_EXPORTS -I/hom
e/kefu/dev/seastar/include -I/home/kefu/dev/seastar/build/debug/gen/include -I/home/kefu/dev/seastar/build/debug/gen/src -I/home/kefu/dev/seastar/src -g -std=gnu++23 -fPIC -U_FORTIFY_SOURCE -Wno-error=unused-result "-Wno-error=#warnings"
-fstack-clash-protection -UNDEBUG -Wall -Werror -Wimplicit-fallthrough -Wdeprecated -Wno-error=deprecated -gz -fsanitize=address -fsanitize=undefined -fno-sanitize=vptr -MD -MT CMakeFiles/seastar.dir/src/core/execution_stage.cc.o -MF CMak
eFiles/seastar.dir/src/core/execution_stage.cc.o.d -o CMakeFiles/seastar.dir/src/core/execution_stage.cc.o -c /home/kefu/dev/seastar/src/core/execution_stage.cc
In file included from /home/kefu/dev/seastar/src/core/execution_stage.cc:28:
In file included from /home/kefu/dev/seastar/include/seastar/core/execution_stage.hh:24:
In file included from /home/kefu/dev/seastar/include/seastar/core/future.hh:28:
In file included from /usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/functional:49:
In file included from /usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/bits/stl_function.h:60:
In file included from /usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/bits/move.h:37:
/usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/type_traits:1118:21: error: static assertion failed due to requirement 'std::__is_complete_or_unbounded(std::__type_identity<fmt::formatter<seastar::basic_sstring<char, unsign
ed int, 15, true>, char, void>>{})': template argument must be a complete class or an unbounded array
 1118 |       static_assert(std::__is_complete_or_unbounded(__type_identity<_Tp>{}),
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/fmt/base.h:1509:39: note: in instantiation of template class 'std::is_constructible<fmt::formatter<seastar::basic_sstring<char, unsigned int, 15>>>' requested here
 1509 |                                      (has_formatter<U, Context>::value &&
      |                                       ^
/usr/include/fmt/base.h:1516:40: note: in instantiation of template class 'fmt::detail::arg_mapper<fmt::context>::formattable<const seastar::basic_sstring<char, unsigned int, 15>>' requested here
 1516 |   template <typename T, FMT_ENABLE_IF(!formattable<T>::value)>
      |                                        ^
/usr/include/fmt/base.h:1517:20: note: while substituting prior template arguments into non-type template parameter [with T = const seastar::basic_sstring<char, unsigned int, 15>]
 1517 |   FMT_MAP_API auto do_map(T&) -> unformattable {
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1518 |     return {};
      |     ~~~~~~~~~~
 1519 |   }
      |   ~
/usr/include/fmt/base.h:1529:62: note: while substituting deduced template arguments into function template 'do_map' [with T = const seastar::basic_sstring<char, unsigned int, 15>, $1 = (no value)]
 1529 |   FMT_MAP_API auto map(T& val) -> decltype(FMT_DECLTYPE_THIS do_map(val)) {
      |                                                              ^
/usr/include/fmt/base.h:1545:50: note: while substituting deduced template arguments into function template 'map' [with T = const seastar::basic_sstring<char, unsigned int, 15>, U = (no value), $2 = (no value)]
 1545 |     type_constant<decltype(arg_mapper<Context>().map(std::declval<const T&>())),
      |                                                  ^
/usr/include/fmt/base.h:2718:16: note: in instantiation of template type alias 'mapped_type_constant' requested here
 2718 |       : types_{mapped_type_constant<Args, buffered_context<Char>>::value...},
      |                ^
/usr/include/fmt/base.h:2869:47: note: in instantiation of member function 'fmt::detail::format_string_checker<char, seastar::basic_sstring<char, unsigned int, 15>, seastar::basic_sstring<char, unsigned int, 15>>::format_string_checker' r
equested here
 2869 |       detail::parse_format_string<true>(str_, checker(s));
      |                                               ^
/home/kefu/dev/seastar/include/seastar/core/execution_stage.hh:342:33: note: in instantiation of function template specialization 'fmt::basic_format_string<char, seastar::basic_sstring<char, unsigned int, 15> &, const seastar::basic_sstri
ng<char, unsigned int, 15> &>::basic_format_string<char[6], 0>' requested here
  342 |         auto name = fmt::format("{}.{}", _name, sg.name());
      |                                 ^
```

in this change, we check for the FMT_VERSION, and explicitly opt out of range-based formatter.

since fmt 11 was release only two weeks ago at the time of writing, quite a few of existing user of fmt have not started using fmt 11, we do not add this template specialization for fmt < 11, even if this does not hurt, but it could slightly slow down the compilation due to inclusion of fmt/ranges.h and processing the template.

### DIFF
--- a/include/seastar/core/sstring.hh
+++ b/include/seastar/core/sstring.hh
@@ -38,6 +38,9 @@
 #include <functional>
 #include <type_traits>
 #include <fmt/format.h>
+#if FMT_VERSION >= 110000
+#include <fmt/ranges.h>
+#endif
 #endif
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/modules.hh>
@@ -877,6 +880,14 @@ std::ostream& operator<<(std::ostream& os, const std::unordered_map<Key, T, Hash
     return os;
 }
 }
+
+#endif
+
+#if FMT_VERSION >= 110000
+
+template <typename char_type, typename Size, Size max_size, bool NulTerminate>
+struct fmt::range_format_kind<seastar::basic_sstring<char_type, Size, max_size, NulTerminate>, char_type> : std::integral_constant<fmt::range_format, fmt::range_format::disabled>
+{};
 
 #endif
 


avikivity:
I'm doubly surprised by the error.

1. Isn't the formatter for basic_sstring more specialized than the formatter for ranges, and is therefore a better candidate? maybe the rules only consider something\<T> to be more specialized than T, but something\<T> and something\<else\<T>> are equally specialized.
2. how aren't we hit by this in fmt 10?

I don't have any objections to the patch (and likely every translation unit includes \<fmt/ranges.h>, but would like to understand more.

tchaikov:
> I'm doubly surprised by the error.
> 
> 1. Isn't the formatter for basic_sstring more specialized than the formatter for ranges, and is therefore a better candidate? maybe the rules only consider something<T> to be more specialized than T, but something<T> and something<else<T>> are equally specialized.

both formatters are partial specialization

fmt/range's :
```c++
template <typename R, typename Char>
struct formatter<
    R, Char,
    enable_if_t<conjunction<
        bool_constant<
            range_format_kind<R, Char>::value != range_format::disabled &&
            range_format_kind<R, Char>::value != range_format::map &&
            range_format_kind<R, Char>::value != range_format::string &&
            range_format_kind<R, Char>::value != range_format::debug_string>>::value>> {
    // ...   
};
```

sstring's
```c++
template <typename char_type, typename Size, Size max_size, bool NulTerminate>
struct fmt::formatter<seastar::basic_sstring<char_type, Size, max_size, NulTerminate>>
    : public fmt::formatter<basic_string_view<char_type>> {
    // ...
};
```

so from the compiler's perspective, both of them win.

> 2. how aren't we hit by this in fmt 10?

my guess is that fmt 11 introduced "Added support for character range formatting". see https://github.com/fmtlib/fmt/releases/tag/11.0.0 . i could bisect fmt to identify the exact commit which introduced the issue, but would leave this for a rainy night.

> 
> I don't have any objections to the patch (and likely every translation unit includes <fmt/ranges.h>, but would like to understand more.

mykaul:
General question - do we get any value from the move from fmt9 to fmt10 and now 11? Apart from job security and refactoring, what's the real value? I'm looking at https://github.com/fmtlib/fmt/blob/master/ChangeLog.md#1100---2024-07-01 which does seem to provide some value - perf. and other improvements.

nyh:
> General question - do we get any value from the move from fmt9 to fmt10 and now 11?

This is what happens when you use an external library - you are **forced** to support the newer versions. Because even if we can stay on fmt9 and not move to fmt10 and fmt11, what will happen when fmt75 comes out with a really important fix (e.g., it supports C++36 and fmt9 fails to compile on C++36)  and we are still stuck with fmt9?

By the way, I think the fmt developers **think** they are backward compatible and upgrading should be effortless, but clearly in C++, this backward-compatibility doesn't quite work even if you hope very much that it does.

I wonder if this situation will be slightly (but only slightly) better if we used C++ std::format instead of the external fmtlib (we can still be hit with problems when upgrading the standard library, though).

mykaul:
> > General question - do we get any value from the move from fmt9 to fmt10 and now 11?
> 
> This is what happens when you use an external library - you are **forced** to support the newer versions. Because even if we can stay on fmt9 and not move to fmt10 and fmt11, what will happen when fmt75 comes out with a really important fix (e.g., it supports C++36 and fmt9 fails to compile on C++36) and we are still stuck with fmt9?

I think there should be a balance we need to do between jumping on every n+1 vs. from time to time doing the hop from n to n+m. It should be based on complexity, risk, but also value.
(Python 3.11 -> 3.12 is another good example).

nyh:
> I think there should be a balance we need to do between jumping on every n+1 vs. from time to time doing the hop from n to n+m. It should be based on complexity, risk, but also value. (Python 3.11 -> 3.12 is another good example).

In my personal experience, there is no such balance to be had: Between every version jump, on average K things need to be fixed (K<1 if we're lucky). If you upgrade every version, you need to fix K things every time, or NK things after N versions. If you jump N versions together, you just find yourself having to fix NK things at the same time - and it's harder to figure out all the separate problems. Worse: if you delay the upgrade more and more, there's bigger and bigger temptation to stay with the antique version, until 5 years down the line something _forces_ you to upgrade and you need to spend a very long time figuring out what happened in the last 5 years.

tchaikov:
this discussion reminds me the fmt migration in scylladb from fmt v8 to fmt v9 and v10. 

a little bit background, fedora rawhide just [upgraded](https://bodhi.fedoraproject.org/updates/FEDORA-2024-bcf368c2e9) to fmt 11, so i wanted to be able to compile scylla on my system and to help the wide community to use seastar. hence this change. but this does not imply that scylladb is moving to fmt 11.

avikivity:
fmt is being standardized, so eventually we'll replace it with \<format>.

Issue #2355, opened 2024-07-18
io_sink pending IOs can result in large allocations


travisdowns:
If there are many pending ios, the io_sink `_pending_ios` container, which is a `circular_buffer` may grow large and cause a large contiguous allocation which can fail due to fragmentation.

I'm interesting in fixing this upstream, before I proceed I'd like the gauge the interest in either of the following solutions:

1) Replace circular_buffer with chunked_fifo for this use. It looks doable, but mising `erase_front` method would have to be added to chunked_fifo, and made more efficient than the current behavior of `clear()`. Performance between these two structures looks comparable but not identical.

2) Offer a method/configuration on the reactor to just reserve the buffers on all shards to a large value, so we can do this once at startup instead of later when memory is fragmented.

travisdowns:
Example stack for the larger allocation:

```
[Backtrace #0]
void seastar::backtrace<seastar::current_backtrace_tasklocal()::$_0>(seastar::current_backtrace_tasklocal()::$_0&&) at /v/build/v_deps_build/seastar-prefix/src/seastar/include/seastar/util/backtrace.hh:68
 (inlined by) seastar::current_backtrace_tasklocal() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/util/backtrace.cc:97
 (inlined by) seastar::current_tasktrace() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/util/backtrace.cc:148
 (inlined by) seastar::current_backtrace() at /v/build/v_deps_build/seastar-prefix/src/seastar/src/util/backtrace.cc:181
seastar::memory::cpu_pages::warn_large_allocation(unsigned long) at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/memory.cc:849
 (inlined by) seastar::memory::cpu_pages::check_large_allocation(unsigned long) at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/memory.cc:912
 (inlined by) seastar::memory::cpu_pages::allocate_large(unsigned int, bool) at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/memory.cc:919
 (inlined by) seastar::memory::allocate_large(unsigned long, bool) at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/memory.cc:1542
 (inlined by) seastar::memory::allocate_slowpath(unsigned long) at /v/build/v_deps_build/seastar-prefix/src/seastar/src/core/memory.cc:1688
[...]

Issue #70, opened 2015-10-21
io_submit blocks the reactor when device's request queue fills up


tgrabiec:
On systems with slow disks it's possible that block device queue will fill up and `io_submit` will block inside `get_request` blocking the reactor thread. This manifests itself with high iowait time.

It can be remedied by increasing `/sys/block/$DEV/queue/nr_requests` to match concurrency level but this has a downside of increasing request latency while not resulting in improved disk utilization. Better would be to avoid overflowing the queue on seastar level by applying back-pressure.

tgrabiec:
@glommer Is this issue fixed completely by ioqueues?

glommer:
Theoretically yes, since we now control how many requests are in flight. It
can probably still happen if we run with a high iodepth but I don't think
this case is worth fixing

travisdowns:
>  Is this issue fixed completely by ioqueues?

Pretty sure the answer is "no". At least in the case where the disk is not performing worse than the io-properties suggest (this seems relatively common at least in short bursts for both local and network attached disks), the IO scheduler will still let many requests into the disk and we can get a high concurrency and can exceed `nr_requests`, causing the reactor to block.

I'm not sure if this was better before the changes in https://github.com/scylladb/seastar/pull/1766, since in principle the back link responded very quickly before, whereas now it takes a while to respond and by that time you have already hit `nr_requests`.

Since we know that hitting `nr_requests` is just a death sentence for the reactor, maybe we should have another hard cap just below that value such that the IO scheduler doesn't exceed this value? It does sound a lot like the old 2 bucket system (though it works in units of "requests" not the cost units the rest of the scheduler deals in) though I don't know if it could somehow be done more simply than that.

travisdowns:
Using io_uring may or may not solve this, related question: https://github.com/axboe/liburing/issues/1184

avikivity:
Do you actually hit nr_request limits? I've seen it with spinning disks, but not SSD/NVMe.

travisdowns:
> Do you actually hit nr_request limits? I've seen it with spinning disks, but not SSD/NVMe.

Yes, but because the disk (EBS in this case, but it also happens with local SSDs) suffers a temporary slowdown, e.g., dropping to 1% of it's usual throughput for a few 100ms. During this hiccup we will quickly exceed `nr_requests` (63 per device) since we are ~3000 IO/s per second. These are background writes so this would be OK (i.e., the world wouldn't stop) except that due to the reactor stall the world does stop.

[...]

PR #2357, opened 2024-07-19
Change pending_io container to chunked_fifo

travisdowns:
The end goal of this series is to swap out circular_buffer for chunked_fifo in the `io_sink::_pending_io` member. 

This is to avoid the large contiguous allocation done by
circular_buffer when there many pending IOs, as such allocations can
fail on a long running system due to fragmentation.

While we are in there, we also optimize the size of the io_request structure.

See https://github.com/scylladb/seastar/issues/2355 and individual changes for more context.

### DIFF
--- a/include/seastar/core/chunked_fifo.hh
+++ b/include/seastar/core/chunked_fifo.hh
@@ -101,6 +101,11 @@ class chunked_fifo {
         // begin and end interpreted mod items_per_chunk
         unsigned begin;
         unsigned end;
+
+        // the number of elements in this chunk
+        size_t size() const {
+            return end - begin;
+        }
     };
     // We pop from the chunk at _front_chunk. This chunk is then linked to
     // the following chunks via the "next" link. _back_chunk points to the
@@ -184,6 +189,10 @@ public:
     inline void pop_front() noexcept;
     inline bool empty() const noexcept;
     inline size_t size() const noexcept;
+    // Pop the first n elements from the fifo. Equivalent to calling pop_front()
+    // n times, though likely to be faster for n greater than 1. The fifo must
+    // contain at least n elements or the behavior is undefined.
+    void pop_front_n(size_t n) noexcept;
     void clear() noexcept;
     // reserve(n) ensures that at least (n - size()) further push() calls can
     // be served without needing new memory allocation.
@@ -368,6 +377,29 @@ void chunked_fifo<T, items_per_chunk>::clear() noexcept {
     _nchunks = 0;
 #endif
 }
+
+template <typename T, size_t items_per_chunk>
+void chunked_fifo<T, items_per_chunk>::pop_front_n(size_t n) noexcept {
+    while (n) {
+        assert(_front_chunk && "pop_front_n n too large");
+
+        auto target = _front_chunk;
+        unsigned delete_count = std::min(target->size(), n);
+
+        for (auto i = target->begin, e = i + delete_count; i != e; i++) {
+            target->items[mask(i)].data.~T();
+        }
+
+        target->begin += delete_count;
+        n -= delete_count;
+
+        if (target->size() == 0) {
+            front_chunk_delete();
+        }
+    }
+}
+
+
 
 template <typename T, size_t items_per_chunk> void
 chunked_fifo<T, items_per_chunk>::shrink_to_fit() noexcept {
--- a/include/seastar/core/chunked_fifo.hh
+++ b/include/seastar/core/chunked_fifo.hh
@@ -335,47 +335,7 @@ chunked_fifo<T, items_per_chunk>::size() const noexcept{
 
 template <typename T, size_t items_per_chunk>
 void chunked_fifo<T, items_per_chunk>::clear() noexcept {
-#if 1
-    while (!empty()) {
-        pop_front();
-    }
-#else
-    // This is specialized code to free the contents of all the chunks and the
-    // chunks themselves. but since destroying a very full queue is not an
-    // important use case to optimize, the simple loop above is preferable.
-    if (!_front_chunk) {
-        // Empty, nothing to do
-        return;
-    }
-    // Delete front chunk (partially filled)
-    for (auto i = _front_chunk->begin; i != _front_chunk->end; ++i) {
-        _front_chunk->items[mask(i)].data.~T();
-    }
-    chunk *p = _front_chunk->next;
-    delete _front_chunk;
-    // Delete all the middle chunks (all completely filled)
-    if (p) {
-        while (p != _back_chunk) {
-            // These are full chunks
-            chunk *nextp = p->next;
-            for (auto i = 0; i != items_per_chunk; ++i) {
-                // Note we delete out of order (we don't start with p->begin).
-                // That should be fine..
-                p->items[i].data.~T();
-        }
-            delete p;
-            p = nextp;
-        }
-        // Finally delete back chunk (partially filled)
-        for (auto i = _back_chunk->begin; i != _back_chunk->end; ++i) {
-            _back_chunk->items[mask(i)].data.~T();
-        }
-        delete _back_chunk;
-    }
-    _front_chunk = nullptr;
-    _back_chunk = nullptr;
-    _nchunks = 0;
-#endif
+    pop_front_n(size());
 }
 
 template <typename T, size_t items_per_chunk>
--- a/include/seastar/core/internal/io_request.hh
+++ b/include/seastar/core/internal/io_request.hh
@@ -36,35 +36,38 @@ namespace internal {
 
 class io_request {
 public:
-    enum class operation { read, readv, write, writev, fdatasync, recv, recvmsg, send, sendmsg, accept, connect, poll_add, poll_remove, cancel };
+    enum class operation : char { read, readv, write, writev, fdatasync, recv, recvmsg, send, sendmsg, accept, connect, poll_add, poll_remove, cancel };
 private:
-    operation _op;
     // the upper layers give us void pointers, but storing void pointers here is just
     // dangerous. The constructors seem to be happy to convert other pointers to void*,
     // even if they are marked as explicit, and then you end up losing approximately 3 hours
     // and 15 minutes (hypothetically, of course), trying to chase the weirdest bug.
     // Let's store a char* for safety, and cast it back to void* in the accessor.
     struct read_op {
+        operation op;
+        bool nowait_works;
         int fd;
         uint64_t pos;
         char* addr;
         size_t size;
-        bool nowait_works;
     };
     struct readv_op {
+        operation op;
+        bool nowait_works;
         int fd;
         uint64_t pos;
         ::iovec* iovec;
         size_t iov_len;
-        bool nowait_works;
     };
     struct recv_op {
+        operation op;
         int fd;
         char* addr;
         size_t size;
         int flags;
     };
     struct recvmsg_op {
+        operation op;
         int fd;
         ::msghdr* msghdr;
         int flags;
@@ -74,31 +77,38 @@ private:
     using write_op = read_op;
     using writev_op = readv_op;
     struct fdatasync_op {
+        operation op;
         int fd;
     };
     struct accept_op {
+        operation op;
         int fd;
         ::sockaddr* sockaddr;
         socklen_t* socklen_ptr;
         int flags;
     };
     struct connect_op {
+        operation op;
         int fd;
         ::sockaddr* sockaddr;
         socklen_t socklen;
     };
     struct poll_add_op {
+        operation op;
         int fd;
         int events;
     };
     struct poll_remove_op {
+        operation op;
         int fd;
         char* addr;
     };
     struct cancel_op {
+        operation op;
         int fd;
         char* addr;
     };
+
     union {
         read_op _read;
         readv_op _readv;
@@ -119,34 +129,34 @@ private:
 public:
     static io_request make_read(int fd, uint64_t pos, void* address, size_t size, bool nowait_works) {
         io_request req;
-        req._op = operation::read;
         req._read = {
+          .op = operation::read,
+          .nowait_works = nowait_works,
           .fd = fd,
           .pos = pos,
           .addr = reinterpret_cast<char*>(address),
           .size = size,
-          .nowait_works = nowait_works,
         };
         return req;
     }
 
     static io_request make_readv(int fd, uint64_t pos, std::vector<iovec>& iov, bool nowait_works) {
         io_request req;
-        req._op = operation::readv;
         req._readv = {
+          .op = operation::readv,
+          .nowait_works = nowait_works,
           .fd = fd,
           .pos = pos,
           .iovec = iov.data(),
           .iov_len = iov.size(),
-          .nowait_works = nowait_works,
         };
         return req;
     }
 
     static io_request make_recv(int fd, void* address, size_t size, int flags) {
         io_request req;
-        req._op = operation::recv;
         req._recv = {
+          .op = operation::recv,
           .fd = fd,
           .addr = reinterpret_cast<char*>(address),
           .size = size,
@@ -157,8 +167,8 @@ public:
 
     static io_request make_recvmsg(int fd, ::msghdr* msg, int flags) {
         io_request req;
-        req._op = operation::recvmsg;
         req._recvmsg = {
+          .op = operation::recvmsg,
           .fd = fd,
           .msghdr = msg,
           .flags = flags,
@@ -168,8 +178,8 @@ public:
 
     static io_request make_send(int fd, const void* address, size_t size, int flags) {
         io_request req;
-        req._op = operation::send;
         req._send = {
+          .op = operation::send,
           .fd = fd,
           .addr = const_cast<char*>(reinterpret_cast<const char*>(address)),
           .size = size,
@@ -180,8 +190,8 @@ public:
 
     static io_request make_sendmsg(int fd, ::msghdr* msg, int flags) {
         io_request req;
-        req._op = operation::sendmsg;
         req._sendmsg = {
+          .op = operation::sendmsg,
           .fd = fd,
           .msghdr = msg,
           .flags = flags,
@@ -191,34 +201,34 @@ public:
 
     static io_request make_write(int fd, uint64_t pos, const void* address, size_t size, bool nowait_works) {
         io_request req;
-        req._op = operation::write;
         req._write = {
+          .op = operation::write,
+          .nowait_works = nowait_works,
           .fd = fd,
           .pos = pos,
           .addr = const_cast<char*>(reinterpret_cast<const char*>(address)),
           .size = size,
-          .nowait_works = nowait_works,
         };
         return req;
     }
 
     static io_request make_writev(int fd, uint64_t pos, std::vector<iovec>& iov, bool nowait_works) {
         io_request req;
-        req._op = operation::writev;
         req._writev = {
+          .op = operation::writev,  
+          .nowait_works = nowait_works,
           .fd = fd,
           .pos = pos,
           .iovec = iov.data(),
           .iov_len = iov.size(),
-          .nowait_works = nowait_works,
         };
         return req;
     }
 
     static io_request make_fdatasync(int fd) {
         io_request req;
-        req._op = operation::fdatasync;
         req._fdatasync = {
+          .op = operation::fdatasync,
           .fd = fd,
         };
         return req;
@@ -226,8 +236,8 @@ public:
 
     static io_request make_accept(int fd, struct sockaddr* addr, socklen_t* addrlen, int flags) {
         io_request req;
-        req._op = operation::accept;
         req._accept = {
+          .op = operation::accept,
           .fd = fd,
           .sockaddr = addr,
           .socklen_ptr = addrlen,
@@ -238,8 +248,8 @@ public:
 
     static io_request make_connect(int fd, struct sockaddr* addr, socklen_t addrlen) {
         io_request req;
-        req._op = operation::connect;
         req._connect = {
+          .op = operation::connect,
           .fd = fd,
           .sockaddr = addr,
           .socklen = addrlen,
@@ -249,8 +259,8 @@ public:
 
     static io_request make_poll_add(int fd, int events) {
         io_request req;
-        req._op = operation::poll_add;
         req._poll_add = {
+          .op = operation::poll_add,
           .fd = fd,
           .events = events,
         };
@@ -259,8 +269,8 @@ public:
 
     static io_request make_poll_remove(int fd, void *addr) {
         io_request req;
-        req._op = operation::poll_remove;
         req._poll_remove = {
+          .op = operation::poll_remove,
           .fd = fd,
           .addr = reinterpret_cast<char*>(addr),
         };
@@ -269,8 +279,8 @@ public:
 
     static io_request make_cancel(int fd, void *addr) {
         io_request req;
-        req._op = operation::cancel;
         req._cancel = {
+          .op = operation::cancel,
           .fd = fd,
           .addr = reinterpret_cast<char*>(addr),
         };
@@ -278,7 +288,7 @@ public:
     }
 
     bool is_read() const {
-        switch (_op) {
+        switch (opcode()) {
         case operation::read:
         case operation::readv:
         case operation::recvmsg:
@@ -290,7 +300,7 @@ public:
     }
 
     bool is_write() const {
-        switch (_op) {
+        switch (opcode()) {
         case operation::write:
         case operation::writev:
         case operation::send:
@@ -303,8 +313,11 @@ public:
 
     sstring opname() const;
 
+    // All operation variants are tagged unions with an operation as
+    // the first member, which we read through the _read union member
+    // (chosen arbitrarily) which is allowed by the common-initial-subsequence rule.
     operation opcode() const {
-        return _op;
+        return _read.op;
     }
 
     template <operation Op>
@@ -359,17 +372,17 @@ public:
 private:
     io_request sub_req_buffer(size_t pos, size_t len) const {
         io_request sub_req;
-        sub_req._op = _op;
         // read_op and write_op share the same layout, so we don't handle
         // them separately
         auto& op = _read;
         auto& sub_op = sub_req._read;
         sub_op = {
+          .op = op.op,
+          .nowait_works = op.nowait_works,
           .fd = op.fd,
           .pos = op.pos + pos,
           .addr = op.addr + pos,
           .size = len,
-          .nowait_works = op.nowait_works,
         };
         return sub_req;
     }
@@ -377,17 +390,17 @@ private:
 
     io_request sub_req_iovec(size_t pos, std::vector<iovec>& iov) const {
         io_request sub_req;
-        sub_req._op = _op;
         // readv_op and writev_op share the same layout, so we don't handle
         // them separately
         auto& op = _readv;
         auto& sub_op = sub_req._readv;
         sub_op = {
+          .op = op.op,
+          .nowait_works = op.nowait_works,
           .fd = op.fd,
           .pos = op.pos + pos,
           .iovec = iov.data(),
           .iov_len = iov.size(),
-          .nowait_works = op.nowait_works,
         };
         return sub_req;
     }
--- a/include/seastar/core/internal/io_sink.hh
+++ b/include/seastar/core/internal/io_sink.hh
@@ -21,7 +21,7 @@
 
 #pragma once
 
-#include <seastar/core/circular_buffer.hh>
+#include <seastar/core/chunked_fifo.hh>
 #include <seastar/core/internal/io_request.hh>
 
 #include <concepts>
@@ -49,7 +49,7 @@ public:
 };
 
 class io_sink {
-    circular_buffer<pending_io_request> _pending_io;
+    chunked_fifo<pending_io_request> _pending_io;
 public:
     void submit(io_completion* desc, internal::io_request req) noexcept;
 
@@ -67,7 +67,7 @@ public:
             drained++;
         }
 
-        _pending_io.erase(_pending_io.begin(), it);
+        _pending_io.pop_front_n(drained);
         return drained;
     }
 };
--- a/include/seastar/core/internal/io_sink.hh
+++ b/include/seastar/core/internal/io_sink.hh
@@ -58,19 +58,16 @@ public:
     // draining should try to drain more
     requires std::is_invocable_r<bool, Fn, internal::io_request&, io_completion*>::value
     size_t drain(Fn&& consume) {
-        size_t pending = _pending_io.size();
         size_t drained = 0;
 
-        while (pending > drained) {
-            pending_io_request& req = _pending_io[drained];
-
+        for (auto& req : _pending_io) {
             if (!consume(req, req._completion)) {
                 break;
             }
             drained++;
         }
 
-        _pending_io.erase(_pending_io.begin(), _pending_io.begin() + drained);
+        _pending_io.erase(_pending_io.begin(), it);
         return drained;
     }
 };
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -415,14 +415,15 @@ void io_sink::submit(io_completion* desc, io_request req) noexcept {
 }
 
 std::vector<io_request::part> io_request::split(size_t max_length) {
-    if (_op == operation::read || _op == operation::write) {
+    auto op = opcode();
+    if (op == operation::read || op == operation::write) {
         return split_buffer(max_length);
     }
-    if (_op == operation::readv || _op == operation::writev) {
+    if (op == operation::readv || op == operation::writev) {
         return split_iovec(max_length);
     }
 
-    seastar_logger.error("Invalid operation for split: {}", static_cast<int>(_op));
+    seastar_logger.error("Invalid operation for split: {}", static_cast<int>(op));
     std::abort();
 }
 
@@ -492,7 +493,7 @@ std::vector<io_request::part> io_request::split_iovec(size_t max_length) {
 }
 
 sstring io_request::opname() const {
-    switch (_op) {
+    switch (opcode()) {
     case io_request::operation::fdatasync:
         return "fdatasync";
     case io_request::operation::write:
--- a/tests/perf/CMakeLists.txt
+++ b/tests/perf/CMakeLists.txt
@@ -85,3 +85,6 @@ seastar_add_test (coroutine
 
 seastar_add_test (allocator
   SOURCES allocator_perf.cc)
+
+seastar_add_test (container
+  SOURCES container_perf.cc)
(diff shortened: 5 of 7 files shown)


include/seastar/core/internal/io_sink.hh
```diff
@@ -58,19 +58,17 @@ public:
...
     requires std::is_invocable_r<bool, Fn, internal::io_request&, io_completion*>::value
     size_t drain(Fn&& consume) {
-        size_t pending = _pending_io.size();
         size_t drained = 0;
+        auto it = _pending_io.begin();
 
-        while (pending > drained) {
-            pending_io_request& req = _pending_io[drained];
-
+        for (auto end = _pending_io.end(); it != end; ++it, ++drained) {
```
xemul, line 64: This took some time to parse this loop :) Why not make it classical

```
    for (auto it = _pending_io.begin(); it != _pending_io.end(); it++) {
        ...
```

?

include/seastar/core/chunked_fifo.hh
```diff
@@ -184,6 +189,10 @@ public:
     inline void pop_front() noexcept;
     inline bool empty() const noexcept;
     inline size_t size() const noexcept;
+    // Pop the first n elements from the fifo. Equivalent to calling pop_front()
+    // n times, though likely to be faster for n greater than 1. The fifo must
+    // contain at least n elements or the behavior is undefined.
+    void pop_front_n(size_t n) noexcept;
```
xemul, line 195: Would `void pop_front(size_t n = 1) noexcept` look nicer?

include/seastar/core/chunked_fifo.hh
```diff
@@ -369,6 +378,29 @@ void chunked_fifo<T, items_per_chunk>::clear() noexcept {
 #endif
 }
 
+template <typename T, size_t items_per_chunk>
+void chunked_fifo<T, items_per_chunk>::pop_front_n(size_t n) noexcept {
+    while (n) {
```
xemul, line 343: This method is going to be used by io_sink like this:

```c++
    size_t drained = 0;
    for (auto& rq : _pending_io) {
        if (!consume(rq)) {
            break;
        }
        drained++;
    }
    _pending_io.pop_front_n(drained);
```

IOW -- it first walks as many elements as it can (up to the whole container), then pop_front_n them in batch.

Would more classical approach be much worse:

```c++
    while (!_pending_io.empty()) {
        if (!consume(_pending_io.front()) {
            break;
        }
        _pending_io.pop_front();
    }
```

?

AFAICS complexity is going to be the same in big-O sense, the only difference seem to be that chunk->begin is incremented step-by-step in the latter case and "in batch" in the former

include/seastar/core/internal/io_request.hh
```diff
@@ -303,8 +313,11 @@ public:
 
     sstring opname() const;
 
+    // All operation variants are tagged unions with an operation as
+    // the first member, which we read through the _read union member
+    // (chosen arbitrarily) which is allowed by the common-initial-subsequence rule.
     operation opcode() const {
-        return _op;
+        return _read.op;
```
avikivity, line 320: Ah, good to see this is formally allowed.

include/seastar/core/internal/io_sink.hh
```diff
@@ -21,7 +21,7 @@
 
 #pragma once
 
-#include <seastar/core/circular_buffer.hh>
+#include "seastar/core/chunked_fifo.hh"
 #include <seastar/core/internal/io_request.hh>
```
avikivity, line 25: > This is to avoid the large contiguous allocation done by
> circular_buffer when there many pending IOs, as such allocations can
> fail on a long running system due to fragmentation.

How many pending ios do you have?


ScyllaDB works hard to ensure a 128kB allocation is "easy". 128kB/32 = 4096, hard to see how a single shard should have 4096 concurrent I/Os with any reasonable latency. Maybe your contiguous allocation threshold is lower?

StephanDollberg: As a side note, with the uring backend network IO also goes through the io_sink so this becomes more likely to happen.

avikivity: That's a good point. Still, I'd like to understand the context.

avikivity: btw, chunked_fifo could be optimized to bump items_per_chunk upward so that sizeof(chunk) approaches a power of two, to avoid wasted space.

nyh: chunked_fifo is currently written to assume that items_per_chunk is a power of two, so sizeof(chunk) will be a power of two if and only if sizeof(T) is a power of two.

We could allow items_per_chunk be non-power-of-two, at the cost of slightly more expensive (but probably negligible) index calculation.

avikivity: I think it can be brought down to zero, we don't have random access.

travisdowns: Thanks @avikivity for your comments.

> Maybe your contiguous allocation threshold is lower?

We also use 128K as our maximum "allowed" allocation size.

The number of IOs was in the range 4000-8000, enough to cause a ~450K allocation in this code. Evidently to exceed 128K would be a smaller number but still more than 1,000.

> hard to see how a single shard should have 4096 concurrent I/Os with any reasonable latency. 

I do not think the latency is reasonable at this point. The specific issue I'm thinking of occurs in a stress test so there are many outstanding IOs. 

Note also that "large IO backlog" does not necessary mean poor latency for all requests. E.g., there may be a bulk background operation creating lots of IO, but at the same time latency sensitive operations which arrive randomly and the IO priority stuff should allow the latency sensitive stuff to jump the queue. However, your point stands for the specific case of `io_sink` since this is a single per-shard queue which is after the scheduling decision has been made. So a large io_sink does imply poor latency for all requests.

The strongest argument to "allowing" may IOs though is that we don't entirely control the disk behavior. Even if we have io properties configured property and play by the rules, there is nothing stopping the disk from changing its performance characteristics suddenly for 100s of millseconds. Such events are fairly common for both attached EBS disks and local SSDs on the usual instance types in EC2 (this is just the cloud I know the best, I don't think it's wildly different anywhere else).

In that case, when we are doing 3,000 IO/s on one shard it only takes about 300 ms to accumulate > 1000 IOs in flight, and I don't think the current IO scheduler will prevent it.

include/seastar/core/internal/io_sink.hh
travisdowns: Well, inside baseball: there was an intermediate version where I used `boost::deque` and where then after the loop `erase(begin(), it)` was called so `it` needed to survive the loop. Evidently, that's no longer the case as we use the count to do the erase, so I'm fixing that. In fact, this can just be a ranged-for.

include/seastar/core/internal/io_sink.hh
travisdowns: @nyh wrote:

> We could allow items_per_chunk be non-power-of-two, at the cost of slightly more expensive (but probably negligible) index calculation.

On the Redpanda side we have `chunked_vector` which offers something like `std::vector`, including `operator[]` and we decided to use power-of-two size in bytes rather than power-of-two size in elements, to save memory at the cost of more expensive indexing. The indexing cost for the constant-divisor `/` is not all that negligible though and we do see regressions in some cases where `std::vector` was replaced with this class. In many cases we just have to switch to iteration rather than `operator[]` indexing as iteration is faster.

A useful API is an inverted iterator like `container::visit_all(Fn& f)` where you pass in the visitor and the container calls it for every element. This can ~remove the cost of the more complicated iterators since it can do a simple iteration over the entire chunk.

include/seastar/core/chunked_fifo.hh
travisdowns: Complexity is the same yes. As you say, however, I expect there is a difference in constant factor. The existing code was written using the bulk erase at the end, so I followed that pattern as I expected "it might be slower" to be the primary concern here and I wanted to minimize that. 

Here are some relevant results from the benchmark:

```
test                                             iterations      median         mad         min         max      allocs       tasks        inst      cycles
container_perf.pop_all_chunked_fifo               338190000     1.406ns     0.030ns     1.366ns     1.475ns       0.000       0.000        12.3         6.0
container_perf.pop_all_cicular_buffer             516880000     1.290ns     0.015ns     1.272ns     1.315ns       0.000       0.000        10.0         5.6
container_perf.erase_trivial_chunked_fifo         299160000     0.182ns     0.005ns     0.178ns     0.201ns       0.000       0.000         0.5         0.2
container_perf.erase_trivial_circular_buffer      715410000     0.143ns     0.005ns     0.137ns     0.153ns       0.000       0.000         0.1         0.1
container_perf.iter_big_chunked_fifo              640540000     0.822ns     0.003ns     0.814ns     0.825ns       0.000       0.000         9.0         3.6
container_perf.iter_big_circular_buffer          1240200000     0.479ns     0.002ns     0.477ns     0.483ns       0.000       0.000         8.0         2.1
```

`pop_all` is the approach you suggested with an "empty" consume, like so:

```
template <typename Traits, typename T>
size_t pop_all_bench(size_t size) {
    auto c = make_n<Traits, T>(size);

    perf_tests::start_measuring_time();
    while (!c.empty()) {
        perf_tests::do_not_optimize(c.front());
        c.pop_front();
    }
    perf_tests::stop_measuring_time();

    perf_tests::do_not_optimize(c);

    return size;
}
```

erase_trivial erases elements using either `pop_front_n` or in the case of `cicular_buffer`, `erase(b, e)`. 

All times are "per element". So `pop_front()` approach is 1.4 ns, whereas we expect the existing one to be something like the iteration time of 0.8 + erase time of 0.2, so ~1.0, pretty close.

Maybe simpler is better? Then the question is if `pop_front_n` should just be removed.

include/seastar/core/chunked_fifo.hh
travisdowns: I think it would pessimize `pop_front()` which has to pay for the extra complexity it doesn't need but it could be tested, but we should wait for the resolution of this [other discussion](https://github.com/scylladb/seastar/pull/2357/files#r1688243424) since it impacts it.

include/seastar/core/internal/io_sink.hh
xemul: > The strongest argument to "allowing" may IOs though is that we don't entirely control the disk behavior. Even if we have io properties configured property and play by the rules, there is nothing stopping the disk from changing its performance characteristics suddenly for 100s of millseconds. Such events are fairly common for both attached EBS disks and local SSDs on the usual instance types in EC2 (this is just the cloud I know the best, I don't think it's wildly different anywhere else).

Why does disk behavior matter? The rate at which IO scheduler dispatches requests is only controlled by the io-properties file, it doesn't care if disk completes them at some different rate (in fact, it sometimes cares if disk completes them slower than expected, but it's short-term and can only slow down the dispatch rate)

xemul: > The number of IOs was in the range 4000-8000

With default io-latency-goal of 0.75ms? That's about 10M operations/sec for a single shard

avikivity: > Thanks @avikivity for your comments.
> 
> > Maybe your contiguous allocation threshold is lower?
> 
> We also use 128K as our maximum "allowed" allocation size.
> 
> The number of IOs was in the range 4000-8000, enough to cause a ~450K allocation in this code. Evidently to exceed 128K would be a smaller number but still more than 1,000.
> 
> > hard to see how a single shard should have 4096 concurrent I/Os with any reasonable latency.
> 
> I do not think the latency is reasonable at this point. The specific issue I'm thinking of occurs in a stress test so there are many outstanding IOs.
> 
> Note also that "large IO backlog" does not necessary mean poor latency for all requests. E.g., there may be a bulk background operation creating lots of IO, but at the same time latency sensitive operations which arrive randomly and the IO priority stuff should allow the latency sensitive stuff to jump the queue. However, your point stands for the specific case of `io_sink` since this is a single per-shard queue which is after the scheduling decision has been made. So a large io_sink does imply poor latency for all requests.
> 
> The strongest argument to "allowing" may IOs though is that we don't entirely control the disk behavior. Even if we have io properties configured property and play by the rules, there is nothing stopping the disk from changing its performance characteristics suddenly for 100s of millseconds. Such events are fairly common for both attached EBS disks and local SSDs on the usual instance types in EC2 (this is just the cloud I know the best, I don't think it's wildly different anywhere else).

The I/O scheduler will globally issue as many ops as it thinks the disk needs to reach --io-latency-goal. If you have one shard and you issue 1000 ops, then on a 100 usec disk it means your latency goal was 100 ms.


> 
> In that case, when we are doing 3,000 IO/s on one shard it only takes about 300 ms to accumulate > 1000 IOs in flight, and I don't think the current IO scheduler will prevent it.

It should.

> 
> A latency regression is not avoidable when the disks do that, but this much preferable to crashing due to a large allocation.

I don't disagree, but what you describe should not have happened.

travisdowns: > Why does disk behavior matter? The rate at which IO scheduler dispatches requests is only controlled by the io-properties file, it doesn't care if disk completes them at some different rate (in fact, it sometimes cares if disk completes them slower than expected, but it's short-term and can only slow down the dispatch rate)

Sorry this was confusion on my end, I was mixing together this issue and #70 which is related to the number of the requests in the disk and so disk completion rate matters. You should mostly ignore anything related in my reply.

travisdowns: > With default io-latency-goal of 0.75ms? That's about 10M operations/sec for a single shard

Sorry I mean the number of IOs in the `io_sink` was that amount, which is just reverse engineered from the large allocation: i.e., the `_pending_ios` buffer was 450K so we can conclude there are 4000-8000 IOs in it.

Yes, we are using the default io-latency goal.

xemul: And what about io_properties.yaml? Maybe it happens without it, so scheduler is not limited in the amount of requests it can dispatch in one tick?

travisdowns: @xemul wrote:

> And what about io_properties.yaml? Maybe it happens without it, so scheduler is not limited in the amount of requests it can dispatch in one tick?

This case happens without any specific io properties set, yes.

However, though the scheduler may limit the size of the io_sink in _some_ cases (and it may in fact do it in this case), I don't think it's a guarantee:

 - Some operations bypass the scheduler/io-queue and just jump directly to the io_sink, like `fsync`. In Redpanda we often have a 1:1 ratio of write to fsync, so there may be a lot of fsync in flight.
 - With io_uring, additional operations end up in the io_sink, which also bypass the IO scheduler, such as socket operations.
 - There is no guarantee that all operations are flushed out of io_sink every tick: e.g., we have only 1K iocbs to absorb requests coming out of the io_sink, if those fill up the io_sink will carry over operations to the next tick.
 
I think all three of these were pointed out to me by @StephanDollberg.

@xemul wrote:

> Why does disk behavior matter? The rate at which IO scheduler dispatches requests is only controlled by the io-properties file, it doesn't care if disk completes them at some different rate

You are right, I was in part mixing this up with a reactor stall issue (tags exhausted at block layer) which does rely on IO completion. This _mostly_ doesn't rely on io completion behavior (though it does a bit per the last item in the list above: the io_sink may carry over ops from tick-to-tick depending on iocb exhaustion and that relates to IO completion, since it is completion that frees iocbs).

@avikivity wrote:

> The I/O scheduler will globally issue as many ops as it thinks the disk needs to reach --io-latency-goal. If you have one shard and you issue 1000 ops, then on a 100 usec disk it means your latency goal was 100 ms.

Right, and we were not using io properties here (we are changing that), though per above it I don't think it is a comprehensive solution.

avikivity: > @xemul wrote:
> 
> > And what about io_properties.yaml? Maybe it happens without it, so scheduler is not limited in the amount of requests it can dispatch in one tick?
> 
> This case happens without any specific io properties set, yes.
> 
> However, though the scheduler may limit the size of the io_sink in _some_ cases (and it may in fact do it in this case, if a suitable io properties were set), I don't think it's a guarantee:
> 
> * Some operations bypass the scheduler/io-queue and just jump directly to the io_sink, like `fsync`. In Redpanda we often have a 1:1 ratio of write to fsync, so there may be a lot of fsync in flight.

You should switch to O_DSYNC. See the ScyllaDB commitlog code, and all our internal discussions around it, since it's quite tricky. On disks with queue/write_cache = write through it's much much much better.



> * With io_uring, additional operations end up in the io_sink, which also bypass the IO scheduler, such as socket operations.

I think this is compelling.

> * There is no guarantee that all operations are flushed out of io_sink every tick: e.g., we have only 1K iocbs to absorb requests coming out of the io_sink, if those fill up the io_sink will carry over operations to the next tick.

That's an abnormal situation that should be fixed. You're adding 0.5ms latency (or maybe more) because we didn't reserve enough buffer space.

Perhaps we should preempt (add_urgent_task) if _pending_tasks grows beyond a reasonable number (100). Then we never reach those crazy counts. 


> 
> I think all three of these were pointed out to me by @StephanDollberg.
> 
> @xemul wrote:
> 
> > Why does disk behavior matter? The rate at which IO scheduler dispatches requests is only controlled by the io-properties file, it doesn't care if disk completes them at some different rate
> 
> You are right, I was in part mixing this up with a reactor stall issue (tags exhausted at block layer) which does rely on IO completion. This _mostly_ doesn't rely on io completion behavior (though it does a bit per the last item in the list above: the io_sink may carry over ops from tick-to-tick depending on iocb exhaustion and that relates to IO completion, since it is completion that frees iocbs).
> 
> @avikivity wrote:
> 
> > The I/O scheduler will globally issue as many ops as it thinks the disk needs to reach --io-latency-goal. If you have one shard and you issue 1000 ops, then on a 100 usec disk it means your latency goal was 100 ms.
> 
> Right, and we were not using io properties here (we are changing that), though per above it I don't think it is a comprehensive solution.

I believe the correct solution is to schedule a reactor poll when the queue is large enough. The reason for queuing in the first place is to amortize the system call / cache thrashing / etc over many requests, but it we already have many requests, let's flush the queue.

travisdowns:
@avikivity @xemul - the above discussion is very informative, but I think it has gone a bit into the weeds around exactly how normal it is to have a "big" io sink.

To bring it back to a higher level and summarize, I think only needs to agree that:

 - Sometimes the io_sink can get large and there is no particular bound on its size in _some_ scenario
 - This can lead to a large allocation, which may crash (or throw which is often worse) due to fragmentation
 - The above change is more or less performance neutral and/or adds basically no additional complexity (depending on whether we use `pop_front`)
 
So it sort of feels like we don't need to isolate exactly when this occurs to move forward with this change?

include/seastar/core/internal/io_sink.hh
travisdowns: > That's an abnormal situation that should be fixed. You're adding 0.5ms latency (or maybe more) because we didn't reserve enough buffer space.
> 
> Perhaps we should preempt (add_urgent_task) if _pending_tasks grows beyond a reasonable number (100). Then we never reach those crazy counts.

I think it is abnormal, yes. We only see this the large allocation in quite specific cases and while we have actually seen the 1K iocbs exhaustion too, it is again a special case, often transient. I'm just trying to be somewhat exhaustive in the cases where this can occur currently.

travisdowns: > I believe the correct solution is to schedule a reactor poll when the queue is large enough. The reason for queuing in the first place is to amortize the system call / cache thrashing / etc over many requests, but it we already have many requests, let's flush the queue.

This seems reasonable to me, but it's a much larger change in terms of behavior and is optimizing what I think you are arguing is an edge case. Also, even if we do signal preemption in this case, it doesn't really guarantee anything about the size of the io queue/sink because an application thread may still blast a lot of stuff into the queue without suspending, or do large reads/write which get split up by the io queue/sink to conform to the max io size which also results in many ops.

See also my attempt to [bypass a lot of this discussion here](https://github.com/scylladb/seastar/pull/2357#issuecomment-2263754754).

avikivity:
> @avikivity @xemul - the above discussion is very informative, and thanks for the thorough review so far, but I do think it has gone a bit into the weeds around exactly how normal it is to have a "big" io sink.
> 
> To bring it back to a higher level and summarize, I think only needs to agree that:
> 
> * In at least _one_ scenario sometimes the io_sink can get large and there is no particular bound on its size

I agree it can happen now, but I think it shouldn't happen in reasonable cases. It indicates that we failed to guarantee low latency.

> * This can lead to a large allocation, which may crash (or throw which is often worse) due to fragmentation
> * The above change is more or less performance neutral and/or adds basically no additional complexity (depending on whether we use `pop_front`)
> 
> So it sort of feels like we don't need to isolate exactly when this occurs to move forward with this change?

I think it's papering over the real problems.

The real problems are the mismatch between size of the queue connecting Seastar and the kernel, and the number of in-flight work we can have (this didn't exist with linux-aio, but does exist with io_uring), and thinking we can queue tens of thousands of I/O requests without impact. The patch downgrades the impact from "crash" to "bad latency + slightly lower efficiency in the normal case". That's not a good tradeoff, if we can change it to "better behavior in extreme cases + no degradation in the normal case".

travisdowns:
> I think it's papering over the real problems.
> 
> The real problems are the mismatch between size of the queue connecting Seastar and the kernel, and the number of in-flight work we can have (this didn't exist with linux-aio, but does exist with io_uring), and thinking we can queue tens of thousands of I/O requests without impact. The patch downgrades the impact from "crash" to "bad latency + slightly lower efficiency in the normal case". That's not a good tradeoff, if we can change it to "better behavior in extreme cases + no degradation in the normal case".

Seastar doesn't prevent the application from submitting many IOs. If the application submits enough IOs, some kind of global guarantee is out the window anyway: if the number of IOs submitted over some period of time exceeds the device IOPS, something has to give. It's only really a question of whether we can preserve some other, lesser guarantees, e.g., whether some subset of IOs, sorted by IO priority can still maintain the latency guarantees (at the cost of the other subset, which must necessarily fail those).

The way those lesser guarantees are implemented is the IO scheduler, but this assumes you have set up io properties correctly and also you have divided your IOs into suitable priority groups. It's still worth taking about what happens when we don't have those in place.

> and thinking we can queue tens of thousands of I/O requests without impact. 

Who is "we"? The application, or seastar itself? I don't think the application in this case is under any illusion that we can queue so many IOs without impact, but we would rather not crash. If it is seastar, then OK, but per above, the last-and-final gate against this is the io scheduler and this doesn't provide _a guarantee_ against this particular structure growing large in every scenario, and if the application doesn't provide good input (io properties, especially) its hands are tied.

Just to note, however, to exceed the desired allocation size of 128K requires only 2,800 queued IOs not tens of thousands. In our case, this occurs because we do a few 10s of MBs of IOs in a burst which internally get written as 16K chunks which ends up 1000s of IOs. This actually does not _usually_ cause the large allocation, but does occasionally, so I interpret this as some additional factor which needs to occur to cause the issue, possibly x00 ms disk throughput dropout (which causes a spike in IOs when it recovers).

> The patch downgrades the impact from "crash" to "bad latency + slightly lower efficiency in the normal case".

In what sense is it slightly lower efficiency in the normal case?

Because `chunked_fifo` operations are slightly slower than their `circular_buffer` counterparts? That is true, based on my micro-benchmarks (also in this PR), but the gap is quite small. If that's a concern we can use an inverted iteration pattern for the consume loop (i.e., the iteration happens by passing a functor in to a `chunked_fifo` method which does the iteration in a chunk-aware away which is faster that `circular_buffer` iteration. Another option is to use `boost::deque` which has faster iteration than `circular_buffer` also (and meets other requirements).

Then the comparison looks different, just "crash" vs "bad latency + slightly higher efficiency in the normal case".

travisdowns:
I would also propose another solution, which is even less invasive: provide some `#define` or `::internal::set_io_pending_size` methods which allows to set the size of this buffer explicitly (via `reserve()`) so we can do this at startup with plenty of contiguous memory and bypass this large allocation.

avikivity:
> > I think it's papering over the real problems.
> > The real problems are the mismatch between size of the queue connecting Seastar and the kernel, and the number of in-flight work we can have (this didn't exist with linux-aio, but does exist with io_uring), and thinking we can queue tens of thousands of I/O requests without impact. The patch downgrades the impact from "crash" to "bad latency + slightly lower efficiency in the normal case". That's not a good tradeoff, if we can change it to "better behavior in extreme cases + no degradation in the normal case".
> 
> Seastar doesn't prevent the application from submitting many IOs. If the application submits enough IOs, some kind of global guarantee is out the window anyway: if the number of IOs submitted over some period of time exceeds the device IOPS, something has to give. It's only really a question of whether we can preserve some other, lesser guarantees, e.g., whether some subset of IOs, sorted by IO priority can still maintain the latency guarantees (at the cost of the other subset, which must necessarily fail those).
> 
> The way those lesser guarantees are implemented is the IO scheduler, but this assumes you have set up io properties correctly and also you have divided your IOs into suitable priority groups. It's still worth taking about what happens when we don't have those in place, and also what happens in cases that have nothing to do with io scheduler, like socket ops when reactor backend is io_uring.


Sure, and we are talking about them. I proposed scheduling an urgent task to flush the queue when it rises about a certain size. I claim it's better than switching to a different data structure, since it not only solves the large allocation, but also avoids stalls while flushing a large queue.


> 
> > and thinking we can queue tens of thousands of I/O requests without impact.
> 
> Who is "we"? The application, or seastar itself? I don't think the application in this case is under any illusion that we can queue so many IOs without impact, but we would rather not crash. If it is seastar, then OK, but per above, the last-and-final gate against globally bad latency is the io scheduler and this doesn't provide _a guarantee_ against this particular structure growing large in every scenario, and if the application doesn't provide good input (io properties, especially) its hands are tied.
> 
> Just to note, however, to exceed the desired allocation size of 128K requires only 2,800 queued IOs not tens of thousands. In our case, this occurs because we do a few 10s of MBs of IOs in a burst which internally get written as 16K chunks which ends up 1000s of IOs. This actually does not _usually_ cause the large allocation, but does occasionally, so I interpret this as some additional factor which needs to occur to cause the issue, possibly x00 ms disk throughput dropout (which causes a spike in IOs when it recovers).
> 
> > The patch downgrades the impact from "crash" to "bad latency + slightly lower efficiency in the normal case".
> 
> In what sense is it slightly lower efficiency in the normal case?
> 
> Because `chunked_fifo` operations are slightly slower than their `circular_buffer` counterparts? That is true, based on my micro-benchmarks (also in this PR), but the gap is quite small. If that's a concern we can use an inverted iteration pattern for the consume loop (i.e., the iteration happens by passing a functor in to a `chunked_fifo` method which does the iteration in a chunk-aware away which is faster that `circular_buffer` iteration. Another option is to use `boost::deque` which has faster iteration than `circular_buffer` also (and meets other requirements).
> 
> Then the comparison looks different, just "crash" vs "bad latency + slightly higher efficiency in the normal case".

I'm not digging my heels against chunked_fifo. I am digging my heels against ignoring the other proposal.

travisdowns:
> Sure, and we are talking about them. I proposed scheduling an urgent task to flush the queue when it rises about a certain size. I claim it's better than switching to a different data structure, since it not only solves the large allocation, but also avoids stalls while flushing a large queue.

Let me try to understand this proposal a bit better.

To summarize the current state. Per shard, there are:

1. Per-iogroup (per-scheduler group in APV v7) _io queues_, and when a user task does an IO it often ends up with an io pushed into one of these io queues (network IO is a quite different through).
2. When the pollers run, there is an `io_queue_submission_pollfn` which drains the io queues into the per-shard _io sink_.
3. Right after the above, there is a `kernel_submit_work_poller` which tries to drain the io_sink to `io_submit` or whatever the backend target is.

The large allocation in step 2: when draining the io queues into the io sink. It is the io sink that becomes too large. The io queue is probably _also_ large but it does not cause a large allocation as it is a linked list (though it is also possible that the io sink ends up much larger than any io queue since many io queues drain into a single io sink).

So there is no opportunity to "schedule a task" at (2): we are running in a sync context in the pollers. We are already in the code we want to schedule.

The only time that it might make sense to schedule a task is in (1), in a user task every time we submit IO we could check if there is "too much" IO queued and then schedule an urgent task to do something. However, this doesn't seem to directly address the problem because here (a) we don't know anything about the io sink, only the io queues, so we can't easily control the size of the io_sink this way other than making some conservative assumptions (b) some things bypass the io queues and go directly to the io sink (fsync, network ops) so we can't limit them in this way and (c) this requires the application to cooperate and trigger a suspension point soon enough (and for our quota not to be exhausted, since in that case I don't think we would run the urgent task before proceeding to polling).

So I may have misunderstood something but I'm not seeing it.

cc also @xemul to check my work here.

avikivity:
> > Sure, and we are talking about them. I proposed scheduling an urgent task to flush the queue when it rises about a certain size. I claim it's better than switching to a different data structure, since it not only solves the large allocation, but also avoids stalls while flushing a large queue.
> 
> Let me try to understand this proposal a bit better.
> 
> To summarize the current state. Per shard, there are:
> 
> 1. Per-iogroup (per-scheduler group in API v7) _io queues_, and when a user task does an IO it often ends up with an io pushed into one of these io queues (network IO is quite different through).
> 2. When the pollers run, there is an `io_queue_submission_pollfn` which drains the io queues into the per-shard _io sink_.
> 3. Right after the above, there is a `kernel_submit_work_poller` which tries to drain the io_sink to `io_submit` or whatever the backend target is.
> 
> The large allocation in step 2: when draining the io queues into the io sink. It is the io sink that becomes too large. The io queue(s) are possibly _also_ large but this does not cause a large allocation as it is a linked list (though it is also possible that the io sink ends up much larger than any io queue since many io queues drain into a single io sink).

Note 1: the io queues are candidates for being chunked fifos. The amount time of a request spends on an io queue can be far longer (seconds) than the amount of time it spend on an io sink (task quota), and linked lists -> chunked fifo is an improvement compared to vector -> chunked fifo which is a regression

Note 2: io_sink is a temporary used to hold requests for the ring buffer between userspace and the kernel (linux-aio or io_uring). The reason it's there is to improve amortization. We could copy requests directly to the ring buffer. Could be worse, but could also be better. If we use IORING_SETUP_SQPOLL, it can cause lots of cacheline bouncing, but also reduce latency. On the other hand with io_uring, if we overflow the queue, 

> 
> So there is no opportunity to "schedule a task" at (2): we are running in a sync context in the pollers. We are already in the code we want to schedule.
> 
> The only time that it might make sense to schedule a task is in (1), in a user task every time we submit IO we could check if there is "too much" IO queued and then schedule an urgent task to do something. However, this doesn't seem to directly address the problem because here (a) we don't know anything about the io sink, only the io queues, so we can't easily control the size of the io_sink this way other than making some conservative assumptions (b) some things bypass the io queues and go directly to the io sink (fsync, network ops) so we can't limit them in this way and (c) this requires the application to cooperate and trigger a suspension point soon enough (and for our quota not to be exhausted, since in that case I don't think we would run the urgent task before proceeding to polling).
> 
> So I may have misunderstood something but I'm not seeing it.

No, I think you are right.

include/seastar/core/internal/io_sink.hh
```diff
@@ -21,7 +21,7 @@
 
 #pragma once
 
-#include <seastar/core/circular_buffer.hh>
+#include "seastar/core/chunked_fifo.hh"
```
avikivity, line 24: Please use \<> includes like the rest of the code.

travisdowns:
> Note 1: the io queues are candidates for being chunked fifos [instead of linked lists].

I didn't examine it in detail but it is a non-owning boost::slist (intrusive linked list) so it's not a straight swap because things which are owned elsewhere are pushed and popped into the queues (so also the performance impact is lower, since there is no per-element allocation). See `fair_queue::queue` for example (this is what underlies "push into" an io queue).

> compared to vector -> chunked fifo which is a regression

Just to clear, this change swaps from `ss::circular_buffer` (not vector) to `ss::chunked_fifo` which is less of a regression. In particular, circular buffer already has "complex iterators" (iterators need to check two conditions during iteration) and also does not have many of the lower level optimizations that vector has.

Circular buffer has another more subtle perf issue: it never shrinks and it does not reset it's head/tail pointers on clear/erase-all so even if we are pushing and then popping all elements (a modest number) we will still use all the memory in the circular buffer over time (i.e., slots 0 to 100 on tick 0, 101 to 200 on tick 2, etc), so it has poor locality in that sense. `chunked_fifo` is both better and worse here: it also creeps over over the chunk in the same way, but this is limited to the size of a chunk, not the maximum size the structure has reached in its lifetime. This could be more or less locality depending on the peak circular buffer size in practice.

It is difficult to evaluate the impact of these issues. In general the absolute cost off all this seems very small compared to the end-to-end cost of an IO, so I didn't sweat it too much. There is much lower hanging performance fruit in seastar on a per-IO basis. I'm also hoping the re-packing of the struct from 48 down to 32 bytes is enough to cancel out any impact in a macro sense.

> Note 2: io_sink is a temporary used to hold requests for the ring buffer between userspace and the kernel (linux-aio or io_uring). The reason it's there is to improve amortization.

Hmm, is it? Only between IOs sent to different groups in the same tick I think. It is the merge point between the io queues. The io queues themselves already provide batching for IOs from the same group sent in the same tick. It is also there to absorb ops which bypass the IO queues.

In general amortization is an interesting topic because we are often more interesting in efficiency than latency, at least when the latency increase is on the order of "a few" ms. E.g., we would gladly pay 1 ms of latency to reduce the CPU cost of IO by 50% say. One could imagine we could do this by increasing the task quota, but in general this does not work, since since is only a best-effort _upper bound_ on inter-poll period. As soon as there are no application tasks we run the pollers (and do IO) anyway. What would be interesting is a lower bound ("linger") on the IO poller interval so they don't run more frequently than say every xxx μs.

> We could copy requests directly to the ring buffer. 

Like "inline" in the IO calls done by the application? This seems sort of incompatible with the IO scheduler, which wants to look at the totality of all IOs during a tick. Once the IO is copied to the ring buffer it is not eligible for sesatar IO scheduling any more. Of course, one could envision a different IO scheduling mechanism which applies logic to decide inline if an IO can go to the ring or needs to be set aside. This would be terrible for `libaio` backend though because going to the ring is `io_submit` syscall which costs a lot. So the current mechanism seems reasonable. Furthermore, the io_sink provides a destination for unscheduled IO ops.

include/seastar/core/chunked_fifo.hh
avikivity: I think it's fine either way. In the end the kernel is going to incur a large number of cache misses / branch mispredicts / spaghetti code issuing the queued up requests.

We should focus on higher level optimizations, like polling more frequently when we have many items, switching to io_uring (currently a small regression), and using IORING_SETUP_IOPOLL/IORING_SETUP_SQPOLL (and the other minor micro-optimizations) to run permanently without interrupts.

travisdowns: > We should focus on higher level optimizations, like polling more frequently when we have many items, switching to io_uring (currently a small regression), and using IORING_SETUP_IOPOLL/IORING_SETUP_SQPOLL (and the other minor micro-optimizations) to run permanently without interrupts.

Thanks for the guidance, it makes a lot of sense and I agree with this approach.

PR #2359, opened 2024-07-23
pollable_fd_state: Mark destructor protected and make non-virtual

xemul:
The pollable_fd_state is base class for backend-specific fd states. Those are created and destroyed via backend virtual calls. In that case, base class destructor doesn't need to be virtual, deleting an object via derived pointer would call correct destructor on its own. Also, the base class destructor is marked protected to explicitly disallow deletion via base-class.

### DIFF
--- a/include/seastar/core/internal/pollable_fd.hh
+++ b/include/seastar/core/internal/pollable_fd.hh
@@ -62,7 +62,6 @@ using pollable_fd_state_ptr = boost::intrusive_ptr<pollable_fd_state>;
 class pollable_fd_state {
     unsigned _refs = 0;
 public:
-    virtual ~pollable_fd_state() {}
     struct speculation {
         int events = 0;
         explicit speculation(int epoll_events_guessed = 0) : events(epoll_events_guessed) {}
@@ -122,6 +121,7 @@ public:
 protected:
     explicit pollable_fd_state(file_desc fd, speculation speculate = speculation())
         : fd(std::move(fd)), events_known(speculate.events) {}
+    ~pollable_fd_state() {};
 private:
     void maybe_no_more_recv();
     void maybe_no_more_send();


xemul:
```
/usr/bin/ld: src/libseastar-module.so: undefined reference to `seastar::pollable_fd_state@seastar::~pollable_fd_state()'
```

xemul:
Hm... C++20 modules enabled [didn't like](https://github.com/scylladb/seastar/pull/2359#issuecomment-2245356539) the `~pollable_fd_state() = default`. Now it's `~pollable_fd_state() {}` and it compiles and links just fine (@tchaikov )

avikivity:
> ```
> /usr/bin/ld: src/libseastar-module.so: undefined reference to `seastar::pollable_fd_state@seastar::~pollable_fd_state()'
> ```

Looks like a gcc bug.

tchaikov (approved):
lgtm.

> Hm... C++20 modules enabled [didn't like](https://github.com/scylladb/seastar/pull/2359#issuecomment-2245356539) the `~pollable_fd_state() = default`. Now it's `~pollable_fd_state() {}` and it compiles and links just fine (@tchaikov )

i am not able to reproduce [this failure](https://github.com/scylladb/seastar/actions/runs/10060110953/job/27806902229#step:11:872) locally with clang 18.1.6, which is the version used in our CI.  

but the failure is indeed reproduciable , see also https://github.com/scylladb/seastar/actions/runs/10060110953/job/27806902229#step:11:873

tchaikov:
> > ```
> > /usr/bin/ld: src/libseastar-module.so: undefined reference to `seastar::pollable_fd_state@seastar::~pollable_fd_state()'
> > ```
> 
> Looks like a gcc bug.

could be a clang bug. or a bug in ccache?

tchaikov:
> Hm... C++20 modules enabled [didn't like](https://github.com/scylladb/seastar/pull/2359#issuecomment-2245356539) the `~pollable_fd_state() = default`. Now it's `~pollable_fd_state() {}` and it compiles and links just fine (@tchaikov )

@xemul could you please file a bug? a very short one with a link to the action run would suffice. my hunch is that it's ccache. but i need to test my hypothesis using my fork first.

PR #2360, opened 2024-07-23
abort_source: subscription: keep callback function alive after abort

bhalevy:
Change 470b539b1c1dd2b080eb13cc77b17eec5a83c21f causes a regression in scylla when the subscription callback function is destroyed after being called when abort is requested.

The app expects this function to remain alive throughout the life time of the subscription object holding it, so seastar should not breask this assumption.

This commit keeps it around by adding a boolean
`_aborted` member to subscription used to make
sure the callback is called at most once,
but otherwise, the function is not destroyed
until the whole sunscription is destroyed.

Added unit test to check that.

### DIFF
--- a/include/seastar/core/abort_source.hh
+++ b/include/seastar/core/abort_source.hh
@@ -66,7 +66,8 @@ public:
     class subscription : public bi::list_base_hook<bi::link_mode<bi::auto_unlink>> {
         friend class abort_source;
 
-        subscription_callback_type _target = noop_handler;
+        subscription_callback_type _target;
+        bool _aborted = false;
 
         explicit subscription(abort_source& as, subscription_callback_type target)
                 : _target(std::move(target)) {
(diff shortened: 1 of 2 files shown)


bhalevy:
> > The app expects this function to remain alive throughout the life time of the subscription object holding it ...
> 
> Why is that?

Well, this is how the code in scylla is currently implemented.
And since we didn't document the lifetime guarantees of the callback function before, I think it's a fair assumption.

xemul:
Compilation with modules failed with
```
/usr/bin/ld: src/libseastar-module.so: undefined reference to `seastar::abort_source@seastar::subscription::noop_handler(std::optional<std::__exception_ptr::exception_ptr> const&)'
```

it looks extremely similar to https://github.com/scylladb/seastar/pull/2359#issuecomment-2245356539 (Cc @tchaikov )

xemul (approved):

Issue #2362, opened 2024-07-23
Compilation with modules support sometimes drops symbols


xemul:
Seen twice

#2359 with `~pollable_fd_state() = default` constructor, the error was

```
/usr/bin/ld: src/libseastar-module.so: undefined reference to `seastar::pollable_fd_state@seastar::~pollable_fd_state()'
```

#2360, this time it tries to find deleted static noop_handler () method

```
/usr/bin/ld: src/libseastar-module.so: undefined reference to `seastar::abort_source@seastar::subscription::noop_handler(std::optional<std::__exception_ptr::exception_ptr> const&)'
```

PR #2365, opened 2024-07-24
github: disable ccache when building with C++ modules 

tchaikov:
ccache does not support C++20 modules yet, see https://github.com/ccache/ccache/issues/1252.

in this change, ccache is disabled when building with C++20 module
support.

Fixes https://github.com/scylladb/seastar/issues/2362

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -23,6 +23,11 @@ on:
         type: string
         default: ''
         required: false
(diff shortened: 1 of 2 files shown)


tchaikov:
tested at https://github.com/tchaikov/seastar/actions/runs/10068505543

tchaikov:
@xemul hi Pavel, could you please help review this change?

PR #2368, opened 2024-07-24
reactor: Initialize _aio_pollfd later

xemul:
Commit 3d654527 (reactor: Add .no_poll_aio to reactor_config) moved initialization of it into constructor's initializer list, but when created pollfd tries to access engine() which is not yet set. Move the pollfd initialization back to reactor::configure().


bhalevy:
change looks good to me.
testing it against latest scylladb/master (scylladb/scylladb@27b305b9d1)

bhalevy:
> change looks good to me. testing it against latest scylladb/master ([scylladb/scylladb@27b305b9d1](https://github.com/scylladb/scylladb/commit/27b305b9d1))

unit tests passed

bhalevy:
https://github.com/scylladb/scylladb/pull/19912

PR #2371, opened 2024-07-26
Detect and report slow IO requests

xemul:
Scheduler goal is to make sure that dispatched requests complete not later than after io-latency-goal duration (which is defaulted to 1.5 times reactor CPU latency goal). In cases when disk or kernel slow down, there's flow-ratio guard that slows down the dispatch rate as well [1]. However, sometimes it may not be enough and requests delay their execution even more. Slowly executing requests are indirectly reported via toal-disk-delay metrics [2], but this metrics accumulates several requests into one counter potentially smoothing spikes by good requests. This detector is aimed at detecting individual slow requests and logging them along with some related statistics that should help to understand what's going on.

refs: #1766 [1] (eefa83771e) [2] (0238d255d8)
fixes: #1311 
closes: #1609

### DIFF
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -116,7 +116,8 @@ private:
     uint64_t _prev_dispatched = 0;
     uint64_t _prev_completed = 0;
     double _flow_ratio = 1.0;
-    timer<lowres_clock> _flow_ratio_update;
+
+    timer<lowres_clock> _averaging_decay_timer;
 
     void update_flow_ratio() noexcept;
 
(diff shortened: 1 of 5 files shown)


xemul:
@avikivity , please review

avikivity:
Looks good. But I'd like it to be optional and opt-in by the application, and the lower threshold configured from the application (like the CPU stall detector). Part of making it more of a library and less of a framework.

xemul:
upd:
- rebased (no conflicts)
- fixed formatting of request execution delay
- made threshold configurable

PR #2380, opened 2024-07-30
reactor: Reset one-shot signal to DFL before handling

xemul:
While handling SEGV/ABRT signal yet another one (of the same kind) may pop up. In that case it will call the same handler again, dead-locking on the guard lock.


tchaikov (approved):
lgtm

PR #2382, opened 2024-08-01
reactor: add missing includes of noncopyable_function.hh

tomershafir:
Add missing include of noncopyable_function.hh in reactor.hh and reactor.cc, and fix lexical order.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -167,6 +167,7 @@ module seastar;
 #include <seastar/util/defer.hh>
 #include <seastar/util/log.hh>
 #include <seastar/util/memory_diagnostics.hh>
+#include <seastar/util/noncopyable_function.hh>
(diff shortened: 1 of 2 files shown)


src/core/reactor.cc
```diff
@@ -167,6 +167,7 @@ module seastar;
 #include <seastar/util/defer.hh>
 #include <seastar/util/log.hh>
 #include <seastar/util/memory_diagnostics.hh>
+#include <seastar/util/noncopyable_function.hh>
```
tchaikov, line 170: why? presumably `reactor.cc` always include `reactor.hh`.

tomershafir: `reactor.cc` also uses this header regardless of `reactor.hh` in `network_stack_factory` impl.

tchaikov (approved):
lgtm

Issue #261, opened 2017-03-12
Seastar shouldn't always traps SIGINT and SIGTERM


nyh:
I already raised this issue several months ago on the mailing list (in a thread titled "How do we handle SIGINT?" with various comments from several Seastar developers) but creating an issue now so we don't forget it.

If you try one of the network server examples in the tutorial, one thing you'll quickly realize is that it's hard to kill this server: control-C (SIGINT) is ignored, and so is "kill" with the default SIGTERM (obviously, "kill -9" does work). This is annoying: It's good that Seastar allows you to capture SIGINT et al. and turn them into asynchronous events that work safely with the rest of the Seastar programming module, but this should be an optional feature, and code that did not explicitly ask to capture these events should probably let them retain the default behavior of killing the program.

303248153:
Is anyone looking at this problem?
I found replace "app.run" with "app.run_deprecated" will solve this problem, because "app.run" will register an exit function and waiting for the first future resolved.
But I think "app.run" does the right thing, just I don't know is there a clean way to make "seastar::keep_doing" stop after SIGINT is received.

Also I found the example apps. like [memcached](https://github.com/scylladb/seastar/blob/master/apps/memcached/memcache.cc#L1426) and [httpd](https://github.com/scylladb/seastar/blob/master/apps/httpd/main.cc#L70), uses run_deprecated, this is a bit awkward...

tgrabiec:
@303248153 You can stop your loops like this:

```
bool stopped = false;
return app.run(argc, argv, [&] {
    engine().at_exit([&] {
        stopped = true;
        return make_ready_future();
    });
    return do_until([&] { return stopped; }, [] {
          ... // main loop
    });
});
```

nyh:
There are two issues here:
1. When SIGINT / SIGTERM is caught, it needs to activate some sort of notification mechanism which the application can use to trigger its clean shutdown. Apparently there is already such mechanism as @tgrabiec mentioned in the previous comment, but this mechanism needs to be explained in the tutorial.
2. By default, if the application does not specifically use the above mechanism to handle SIGINT / SIGTERM, Seastar shouldn't capture this signal at all, and it should trigger an immediate exit - because this is the Unix/Linux tradition. And because it's good enough for many applications (and certainly for test applications).

avikivity:
We should really change those APIs.

Something like

```c++
    bool stopped = false;
    // FIXME: do_with or a seastar::thread
[...]

PR #2384, opened 2024-08-04
Add seastar signal api

tomershafir:
fixes: https://github.com/scylladb/seastar/issues/261.

The PR adds an object based interface for signals on seastar namespace.

- Adds `signal.hh` header implemented by `reactor.cc`.
- Adds `doc/signal.md`.
- Deprecates `reactor::handle_signal`.
- Supports C++20 modules.

Future possibilities:
- Add signal handler unset logic, and unset signal handler on destructor.
- Consider to add a reliable test case for `once` (should require reactor changes to expose an indicator).

### DIFF
--- a/doc/signal.md
+++ b/doc/signal.md
@@ -1,0 +1,30 @@
+# Signals
+
+Seastar provides an interface to handle signals natively and safely as asynchronous tasks.
+
+It provides a dedicated header called [seastar/core/signal.hh](../include/seastar/core/signal.hh).
+
+## Usage
+
+You should call `seastar::handle_signal` procedure in order to register the provided signal handler for the specified signal based on the configuration params.
+
+The procedure must be called inside the `app.run()` lambda, otherwise it's UB.
+
+### Examples
+
+```C++
+#include <seastar/core/app-template.hh>
+#include <seastar/core/signal.hh>
+
+int main(int argc, char** argv) {
+    seastar::app_template app;
+    return app.run(argc, argv, [] {
+        seastar::handle_signal(SIGINT, [&] {
+            std::cout << "caught sigint\n";
+        }, true);
+    });
+}
+```
+
+- [tests/unit/signal_test.cc](../tests/unit/signal_test.cc)
+- [apps/lib/stop_signal.hh](../apps/lib/stop_signal.hh)
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -354,6 +354,7 @@ private:
     bool pure_poll_once();
 public:
     /// Register a user-defined signal handler
+    [[deprecated("Use seastar::handle_signal(signo, handler, once); instead")]]
     void handle_signal(int signo, noncopyable_function<void ()>&& handler);
     void wakeup();
     /// @private
@@ -385,6 +386,8 @@ private:
     friend class thread_pool;
     friend class thread_context;
     friend class internal::cpu_stall_detector;
+
+    friend void handle_signal(int signo, noncopyable_function<void ()>&& handler, bool once);
 
     uint64_t pending_task_count() const;
     void run_tasks(task_queue& tq);
--- a/src/core/signal.cc
+++ b/src/core/signal.cc
@@ -15,34 +15,29 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-/*
- * Copyright (C) 2014 Cloudius Systems, Ltd.
- */
 
+#ifdef SEASTAR_MODULE
+module;
+#endif
+
+#include <stdexcept>
+
+#ifdef SEASTAR_MODULE
+module seastar;
+#else
 #include <seastar/core/signal.hh>
-#include <seastar/core/shared_ptr.hh>
-#include <seastar/core/do_with.hh>
-#include <seastar/testing/test_case.hh>
+#include <seastar/core/reactor.hh>
+#endif
 
-using namespace seastar;
+namespace seastar {
 
-extern "C" {
-#include <signal.h>
-#include <sys/types.h>
-#include <unistd.h>
+void handle_signal(int signo, noncopyable_function<void ()>&& handler, bool once) {
+    auto& r = engine();
+    if (once) {
+        r._signals.handle_signal_once(signo, std::move(handler));
+    } else {
+        r._signals.handle_signal(signo, std::move(handler));
+    }
 }
 
-SEASTAR_TEST_CASE(test_sighup) {
-    return do_with(make_lw_shared<promise<>>(), false, [](auto const& p, bool& signaled) {
-        seastar::handle_signal(SIGHUP, [p, &signaled] {
-            signaled = true;
-            p->set_value();
-        });
-
-        kill(getpid(), SIGHUP);
-
-        return p->get_future().then([&] {
-            BOOST_REQUIRE_EQUAL(signaled, true);
-        });
-    });
-} 
+}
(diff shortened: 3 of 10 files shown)


xemul:
@tomershafir , please fix the commits sequence, it contains some "Merge branch 'master'" and a post-fixlet, both are not welcome

doc/signal.md
```diff
@@ -0,0 +1,24 @@
...
+
+Seastar provides an interface to handle signals natively and safely as asynchronous tasks.
+
+It provides a dedicated header called [seastar/core/signal.hh](../include/seastar/core/signal.hh).
+
+## Usage
+
+You need to construct a `signal_handler` instance that will register the provided signal handler for the specified signal based on the configuration params.
+
+At the moment, the signal handler will remain past object destruction.
```
xemul, line 11: That's extremely confusing. I would expect that desctruction of an object unregisters the handler

include/seastar/core/reactor.hh
```diff
@@ -354,13 +354,23 @@ private:
...
     /// Register a user-defined signal handler
+    [[deprecated("Use seastar::signal_handler sh(signo, handler); instead")]]
     void handle_signal(int signo, noncopyable_function<void ()>&& handler);
     void wakeup();
     /// @private
     bool stopped() const noexcept { return _stopped; }
 
 private:
     class signals {
+        struct _signal_handler {
```
xemul, line 365: Leading underscore is for private members, class names shouldn't have it

include/seastar/core/reactor.hh
```diff
@@ -371,20 +381,15 @@ private:
...
     };
-
+    
     signals _signals;
     std::unique_ptr<thread_pool> _thread_pool;
+    
     friend class thread_pool;
     friend class thread_context;
     friend class internal::cpu_stall_detector;
+    friend struct signal_handler;
```
xemul, line 392: New friends for reactor class are not welcome. It's large enough already

tomershafir: The other ways would be:

1. Add public `reactor::handle_signal_once` and undo deprecation of `reactor::handle_signal`. Users will be able to call them and I think it inverts the purpose of changing the api. Also, it adds more code to the reactor.
2. Refactor signal logic out of the reactor. It seems like the blast radius is big and it's not an easy task.

Thus I would prefer the `friend struct` solution at the moment. Let me know wdyt

xemul: We already have nested class signal, public nested class signal_handler would be nice replacement for explicit friend. But as I wrote above, installing handler in constructor smells like RAII and implies unsetting it in destructor

tomershafir: My goal has been to add a clean seastar namespace level api, not through the reactor.

doc/signal.md
tomershafir: Ill add unregister logic to this PR

tomershafir:
@xemul Regarding destructor logic, it seems to me like the RAII approach here may be less suitable. As the POSIX api doesn't fit RAII naturally, adapting it can introduce complexity and confusion. I assume that multiple objects can be constructed for a single `signo`, like POSIX.

For example:
```C++
seastar::signal_handler signal_handler1(SIGINT, [this] { signaled(); });
...
seastar::signal_handler signal_handler2(SIGINT, [this] { signaled(); });
...
```
What should happen if `signal_handler1` is destroyed after `signal_handler2` is constructed and before it's destroyed?
What should happen if `signal_handler1` is destroyed after `signal_handler2` is destroyed?
The possible solutions to these problems don't seem natural for RAII to me.
 
I think I should re-target this PR to introduce a clean seastar namespace level api, and do it via procedures rather than objects. Wdyt?

xemul:
> I think I should re-target this PR to introduce a clean seastar namespace level api, and do it via procedures rather than objects. Wdyt?

Yes, this sounds good

tomershafir:
@xemul I force pushed the changes

tests/unit/signal_test.cc
```diff
@@ -20,6 +20,7 @@
  */
 
 #include <seastar/core/reactor.hh>
```
xemul, line 22: Is this header still needed?

include/seastar/core/reactor.hh
```diff
@@ -371,21 +380,17 @@ private:
...
-            noncopyable_function<void ()> _handler;
-        };
-        std::atomic<uint64_t> _pending_signals;
-        std::unordered_map<int, signal_handler> _signal_handlers;
     };
-
+    
     signals _signals;
     std::unique_ptr<thread_pool> _thread_pool;
+
```
xemul, line 387: Unrelated change

include/seastar/core/reactor.hh
```diff
@@ -371,21 +380,17 @@ private:
...
-    private:
-        struct signal_handler {
-            signal_handler(int signo, noncopyable_function<void ()>&& handler);
-            noncopyable_function<void ()> _handler;
-        };
-        std::atomic<uint64_t> _pending_signals;
-        std::unordered_map<int, signal_handler> _signal_handlers;
     };
-
+
```
xemul, line 384: Unwanted indentation

include/seastar/core/reactor.hh
```diff
@@ -371,21 +380,17 @@ private:
         void handle_signal_once(int signo, noncopyable_function<void ()>&& handler);
         static void action(int signo, siginfo_t* siginfo, void* ignore);
         static void failed_to_handle(int signo);
-    private:
-        struct signal_handler {
-            signal_handler(int signo, noncopyable_function<void ()>&& handler);
-            noncopyable_function<void ()> _handler;
-        };
-        std::atomic<uint64_t> _pending_signals;
-        std::unordered_map<int, signal_handler> _signal_handlers;
```
xemul, line 380: Why is this block moved upwards?

src/core/signal.cc
```diff
@@ -0,0 +1,50 @@
...
+module seastar;
+#else
+#include <seastar/core/signal.hh>
+#include <seastar/core/reactor.hh>
+#endif
+
+namespace seastar {
+
+static reactor& engine_or_throw() {
+    if (!engine_is_ready()) {
```
xemul, line 35: I think that's impossible. No tasks can be run until engine is ready

include/seastar/core/reactor.hh
tomershafir: To remove redundant `private` in a class

tomershafir: Ill extract it into a different PR

tests/unit/signal_test.cc
tomershafir: No, will remove

src/core/signal.cc
tomershafir: Its possible, if a user calls the procedure before `app.run()`, like:

```C++
seastar::handle_signal(SIGINT, [] {});
return app.run(argc, argv, [&app] -> seastar::future<> {
...
```

I dont think its obvious that it must be called inside `run()`, so I think the logic provides a more controlled and understandable error handling.

xemul: In this case the "reactor must be initialized" can be confusing message, because it's not obvious that "app.run()" initializes the reactor. I would just mention in the method doc, that it's only valid to use one from inside app.run() lambda, otherwise it's UB

src/core/signal.cc
```diff
@@ -31,15 +31,8 @@ module seastar;
 
 namespace seastar {
 
-static reactor& engine_or_throw() {
```
xemul: Please, squash it with the previous commit

tomershafir: sorry, done

tomershafir:
@xemul close https://github.com/scylladb/seastar/issues/261?

PR #2386, opened 2024-08-08
WIP: route discards via the I/O queue

pwrobelse:
This set of changes is an another attempt to route discard requests
via the I/O queue to ensure, that they are issued to the drive in a
controlled way.

This PR:
- adds discard type to `internal::io_request`
- renames `io_desc_read_write` to `queued_io_request_completion` -- discards will use it too
- adds `discard_to_write_ratio` to `io-properties.yaml`
- adds `discard` direction to `io_direction_and_length`
- routes discard requests via the I/O queue
- introduces `remove_file_via_blocks_discarding()` function that can be used to remove files block by block

### DIFF
--- a/include/seastar/core/internal/io_request.hh
+++ b/include/seastar/core/internal/io_request.hh
@@ -425,6 +425,12 @@ private:
         return sub_req;
     }
     std::vector<part> split_iovec(size_t max_length);
+
+    io_request sub_req_discard(size_t relative_offset, size_t new_length) const {
+        auto& op = _discard;
+        return make_discard(op.fd, op.offset + relative_offset, new_length);
+    }
+    std::vector<part> split_discard(size_t max_length);
 };
 
 struct io_request::part {
--- a/include/seastar/core/internal/io_request.hh
+++ b/include/seastar/core/internal/io_request.hh
@@ -36,7 +36,7 @@ namespace internal {
 
 class io_request {
 public:
-    enum class operation : char { read, readv, write, writev, fdatasync, recv, recvmsg, send, sendmsg, accept, connect, poll_add, poll_remove, cancel };
+    enum class operation : char { read, readv, write, writev, fdatasync, recv, recvmsg, send, sendmsg, accept, connect, poll_add, poll_remove, cancel, discard };
 private:
     // the upper layers give us void pointers, but storing void pointers here is just
     // dangerous. The constructors seem to be happy to convert other pointers to void*,
@@ -108,7 +108,12 @@ private:
         int fd;
         char* addr;
     };
-
+    struct discard_op {
+        operation op;
+        int fd;
+        uint64_t offset;
+        uint64_t length;
+    };
     union {
         read_op _read;
         readv_op _readv;
@@ -124,6 +129,7 @@ private:
         poll_add_op _poll_add;
         poll_remove_op _poll_remove;
         cancel_op _cancel;
+        discard_op _discard;
     };
 
 public:
@@ -287,6 +293,17 @@ public:
         return req;
     }
 
+    static io_request make_discard(int fd, uint64_t offset, uint64_t length) {
+        io_request req;
+        req._discard = {
+            .op = operation::discard,
+            .fd = fd,
+            .offset = offset,
+            .length = length
+        };
+        return req;
+    }
+
     bool is_read() const {
         switch (opcode()) {
         case operation::read:
@@ -363,6 +380,9 @@ public:
         }
         if constexpr (Op == operation::cancel) {
             return _cancel;
+        }
+        if constexpr (Op == operation::discard) {
+            return _discard;
         }
     }
 
--- a/include/seastar/core/internal/io_request.hh
+++ b/include/seastar/core/internal/io_request.hh
@@ -435,18 +435,23 @@ struct io_request::part {
 
 // Helper pair of IO direction and length
 struct io_direction_and_length {
-    size_t _directed_length; // bit 0 is R/W flag
+    size_t _directed_length; // bits 0 and 1 depict R/W/D flags
+
+    static constexpr int direction_bits_count = 2;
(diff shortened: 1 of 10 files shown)


src/core/io_queue.cc
```diff
@@ -1055,7 +1059,26 @@ void io_queue::submit_request(queued_io_request_completion* desc, internal::io_r
...
+        submit_blocks_discarding(desc, std::move(req));
+    }
+}
+
+void io_queue::submit_blocks_discarding(queued_io_request_completion* desc, internal::io_request req) noexcept {
+    auto& discard_req = req.as<internal::io_request::operation::discard>();
+
+    // Caller synchronizes using the future returned from `desc->get_future()`.
+    // The attached continuation resolves that future when `punch_hole()` finishes.
+    (void) engine().punch_hole(discard_req.fd, discard_req.offset, discard_req.length).then_wrapped([desc, len = discard_req.length] (future<> f) {
```
xemul, line 1075: nit: There's neat `future::forward_to(promise<>)` method that can be used to link punch_hole's future with desc's promise directly

include/seastar/core/internal/io_request.hh
```diff
@@ -422,18 +422,23 @@ struct io_request::part {
...
+    static constexpr int direction_bits_count = 2;
+    static constexpr int direction_bits_mask = (1 << direction_bits_count) - 1;
 
 public:
-    size_t length() const noexcept { return _directed_length >> 1; }
-    int rw_idx() const noexcept { return _directed_length & 0x1; }
+    size_t length() const noexcept { return _directed_length >> direction_bits_count; }
+    int rwd_idx() const noexcept { return _directed_length & direction_bits_mask; }
+
+    static constexpr int discard_idx = 2;
```
xemul, line 434: I think that discard request should carry write_idx direction and whether or not to submit it via sink or via punch_hole() should be decided based on request::op()

include/seastar/core/seastar.hh
```diff
@@ -336,6 +336,18 @@ future<> sync_directory(std::string_view name) noexcept;
...
+///
+/// \param name name of the file to remove.
+///
+/// \note
+/// The removal is processed by the I/O queue according to its configuration.
+/// The removal may be delayed when the bandwidth is not available. The user
+/// must ensure, that the file is not used until the removal finishes.
+///
+/// Effectively: opens a file, unlinks it and punches holes until nothing is left.
+future<> remove_file_via_blocks_discarding(std::string_view name) noexcept;
```
xemul, line 349: seastar.hh is too generic. The include/seastar/util/file.hh seems to be better place

include/seastar/core/internal/io_request.hh
pwrobelse: Sure, I can revert the change and use `io_direction_and_length::write` for discards. The new direction does not look right in the case of `duplex` mode -- I would not expect three directions in duplex mode.

However, if there is no `discard` direction, I see the following three issues that are described below.

Firstly, if the write direction is used, then the calculation in `internal::request_tokens()` will need to contain a new if statement. It is going to distinguish between true write requests and discard requests (based on operation type), because the coefficients are different for write/discard operations.

Secondly, I am not sure about the usage of `_group->_max_request_length[]`. If I understand correctly, the maximum request length is different for discards and for writes. So, simple usage of `_group->_max_request_length[direction]` will be invalid. Somehow, I feel that it would end up with creation of a new private function that would look similar to the function below:

```
size_t get_max_request_length(io_direction_and_length dnl, const internal::io_request& req) {
    if (req.opcode() != internal::io_request::operation::discard) {
    	return _group->_max_request_length[dnl.rw_idx()];
    }
    
    return _group->_max_discard_request_length;
}
```

Lastly, the way how statistics are accumulated in `io_queue::priority_class_data::on_dispatch()` will mix discards and writes. The code looks as follows: `_rwstat[dnl.rwd_idx()].add(dnl.length());`.

If the three mentioned points do not cause a problem, then I will introduce the change.

src/core/io_queue.cc
pwrobelse: I tried to use it, but the following two issues occurred.

Firstly, the type of the future returned from `punch_hole()` is different than the type of future produced by the promise from `desc`.
```
/repo/scylladb/seastar/seastar/include/seastar/core/future.hh:1610:34: note:   no known conversion for argument 1 from ‘seastar::promise<long unsigned int>’ to ‘seastar::promise<void>&&’
```

Secondly, `desc->set_exception()` and `desc->complete()` are part of reasoning about the lifetime of `queued_io_request_completion`. They delete `desc` object -- the memory is freed. Moreover, they update `priority_class_data` and `io_queue`.

src/core/io_queue.cc
```diff
@@ -1049,6 +1062,10 @@ future<size_t> io_queue::submit_io_write(internal::priority_class pc, size_t len
     return queue_request(std::move(pc), io_direction_and_length(io_direction_write, len), std::move(req), intent, std::move(iovs));
 }
 
+future<size_t> io_queue::submit_io_discard(internal::priority_class pc, size_t len, internal::io_request req, io_intent* intent) noexcept {
+    return queue_request(std::move(pc), io_direction_and_length(io_direction_discard, len), std::move(req), intent, iovec_keeper{});
```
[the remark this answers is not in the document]
pwrobelse: Fixed

src/core/io_queue.cc
xemul: > Firstly, the type of the future returned from punch_hole() is different than the type of future produced by the promise from desc.

There's future::ignore_result() that converts future<T> into future<>

mykaul:
CC @regevran

tgrabiec:
@xemul Did this effort go anywhere? I followed various draft PRs and found a dead-end here.

robertbindar:
> @xemul Did this effort go anywhere? I followed various draft PRs and found a dead-end here.

nope, it did not.

PR #2390, opened 2024-08-15
Address non-monotonicity of steal time and other issues

travisdowns:
This series primarily addresses the problem that on systems with low amounts of steal, steal time appears negative (i.e., the cumulative steal time counter goes down from sample to sample).

This wrong on the face of it and also causes serious problems as a metric in prometheus since the counter contract (monotonic increase) is violated. This causes spurious "counter reset" detection in prometheus and hence bogus very large or very small steal time results in `rate` (or similar) queries.

This is addressed in two ways:
 - We make the sleep time calculation more accurate, which is the underlying reason for negative steal which reduces the error (and so "negativeness") of steal by a couple orders of magnitude. After this change steal time is often 0 when rounded to the nearest ms where it wasn't before.
 - Because the reduction above still does not prevent small negative steal completely, we change the implementation of the metric to essentially cap steal from below 0 in periods where steal was negative.

The individual changes have further details. I am open to splitting commits that may be less popular or require more discussion into a different PR if it makes sense.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -553,6 +553,8 @@ public:
 
     steady_clock_type::duration total_idle_time();
     steady_clock_type::duration total_busy_time();
+    steady_clock_type::duration total_awake_time() const;
+    std::chrono::nanoseconds total_cpu_time() const;
     std::chrono::nanoseconds total_steal_time();
 
     const io_stats& get_io_stats() const { return _io_stats; }
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -549,7 +549,7 @@ public:
     [[deprecated("Use this_shard_id")]]
     shard_id cpu_id() const;
 
-    void sleep();
+    void try_sleep();
 
     steady_clock_type::duration total_idle_time();
     steady_clock_type::duration total_busy_time();
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -382,7 +382,6 @@ private:
     task_queue* pop_active_task_queue(sched_clock::time_point now);
     void insert_activating_task_queues();
     void account_runtime(task_queue& tq, sched_clock::duration runtime);
-    void account_idle(sched_clock::duration idletime);
     void allocate_scheduling_group_specific_data(scheduling_group sg, unsigned long key_id);
     future<> rename_scheduling_group_specific_data(scheduling_group sg);
     future<> init_scheduling_group(scheduling_group sg, sstring name, sstring shortname, float shares);
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -302,6 +302,15 @@ private:
     const bool _reuseport;
     circular_buffer<double> _loads;
     double _load = 0;
+    // Next two fields are required to enforce the monotonicity of total_steal_time()
+    // see that method for details.
+
+    // Last measured accumulated steal time, i.e., the simple difference of accumulated
+    // awake time and consumed thread CPU time.
+    sched_clock::duration _last_true_steal{0};
+    // Accumulated steal time forced to be monotinic by rejecting any updates that would
+    // decrease it. See total_steal_time() for details.
+    sched_clock::duration _last_mono_steal{0};
     sched_clock::duration _total_idle{0};
     sched_clock::duration _total_sleep{0};
     sched_clock::time_point _start_time = now();
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -303,7 +303,7 @@ private:
     circular_buffer<double> _loads;
     double _load = 0;
     sched_clock::duration _total_idle{0};
-    sched_clock::duration _total_sleep;
+    sched_clock::duration _total_sleep{0};
     sched_clock::time_point _start_time = now();
     output_stream<char>::batch_flush_list_t _flush_batching;
     std::atomic<bool> _sleeping alignas(seastar::cache_line_size){0};
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4811,7 +4811,33 @@ std::chrono::nanoseconds reactor::total_steal_time() {
     // unexpectedly blocked (since CPU time won't tick up when that occurs).
     //
     // But what we have here should be good enough and at least has a well defined meaning.
-    return total_awake_time() - total_cpu_time();
+    //
+    // Because we calculate sleep time with timestamps around polling methods that may sleep, like
+    // io_getevents, we systematically over-count sleep time, since there is CPU usage within the
+    // period timed as sleep, before and after an actual sleep occurs (and no sleep may occur at all,
+    // e.g., if there are events immediately available). Over-counting sleep means we under-count the 
+    // wall-clock awake time, and so if there is no "true" steal, we will generally have a small
+    // *negative* steal time, because we under-count awake wall clock time while thread CPU time does
+    // not have a corresponding error.
+    //
+    // Becuase we claim "steal" is a counter, we must ensure that it never deceases, because PromQL
+    // functions which use counters will produce non-sensical results if they do. Therefore we clamp
+    // the output such that it never decreases.
+    //
+    // Finally, we don't just clamp difference of awake and CPU time since proces start at 0, but
+    // take the last value we returned from this function and then calculate the incremental steal
+    // time since that measurement, clamped to 0. This means that as soon as steal time becomes 
+    // positive, it will be reflected in the measurement, rather than needing to "consume" all the
+    // accumulated negative steal time before positive steal times start showing up.
+
+
+    auto true_steal = total_awake_time() - total_cpu_time();
+    auto mono_steal = _last_mono_steal + std::max(true_steal - _last_true_steal, 0ns);
+
+    _last_true_steal = true_steal;
+    _last_mono_steal = mono_steal;
+
+    return mono_steal;
 }
 
 static std::atomic<unsigned long> s_used_scheduling_group_ids_bitmap{3}; // 0=main, 1=atexit
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1006,11 +1006,6 @@ reactor::account_runtime(task_queue& tq, sched_clock::duration runtime) {
     tq._runtime += runtime;
 }
 
-void
-reactor::account_idle(sched_clock::duration runtime) {
-    // anything to do here?
-}
-
 struct reactor::task_queue::indirect_compare {
     bool operator()(const task_queue* tq1, const task_queue* tq2) const {
         return tq1->_vruntime < tq2->_vruntime;
@@ -3252,7 +3247,6 @@ int reactor::do_run() {
         if (check_for_work()) {
             if (idle) {
                 _total_idle += idle_end - idle_start;
-                account_idle(idle_end - idle_start);
                 idle_start = idle_end;
                 idle = false;
             }
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3278,13 +3278,11 @@ int reactor::do_run() {
                     // Turn off the task quota timer to avoid spurious wakeups
                     struct itimerspec zero_itimerspec = {};
                     _task_quota_timer.timerfd_settime(0, zero_itimerspec);
-                    auto start_sleep = now();
                     _cpu_stall_detector->start_sleep();
-                    sleep();
+                    try_sleep();
                     _cpu_stall_detector->end_sleep();
                     // We may have slept for a while, so freshen idle_end
                     idle_end = now();
-                    _total_sleep += idle_end - start_sleep;
                     _task_quota_timer.timerfd_settime(0, task_quote_itimerspec);
                 }
             } else {
@@ -3302,8 +3300,9 @@ int reactor::do_run() {
     return _return;
 }
 
+
 void
-reactor::sleep() {
+reactor::try_sleep() {
     for (auto i = _pollers.begin(); i != _pollers.end(); ++i) {
         auto ok = (*i)->try_enter_interrupt_mode();
         if (!ok) {
@@ -4812,6 +4811,10 @@ std::chrono::nanoseconds reactor::total_steal_time() {
     // Because this is totally in userspace we can miss some events. For instance, if the seastar
     // process is ready to run but the kernel hasn't scheduled us yet, that would be technically
     // steal time but we have no ways to account it.
+    //
+    // Furthermore, not all steal is from other processes: time used by the syscall thread and any
+    // alien threads will show up as steal as well as any time spent in a system call that
+    // unexpectedly blocked (since CPU time won't tick up when that occurs).
     //
     // But what we have here should be good enough and at least has a well defined meaning.
     return total_awake_time() - total_cpu_time();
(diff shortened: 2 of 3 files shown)


include/seastar/core/reactor.hh
```diff
@@ -578,6 +578,8 @@ public:
 
     steady_clock_type::duration total_idle_time();
```
avikivity, line 587: > awake-busy
> awake-idle
> sleeping

Most propitious, we were just discussing this on our own customer issue.

src/core/reactor_backend.cc
```diff
@@ -483,7 +483,12 @@ bool reactor_backend_aio::await_events(int timeout, const sigset_t* active_sigma
     bool did_work = false;
     int r;
     do {
+        const bool may_sleep = !tsp || (tsp->tv_nsec + tsp->tv_sec > 0);
+        const auto before_getevents = may_sleep ? sched_clock::now() : sched_clock::time_point{};
         r = io_pgetevents(_polling_io.io_context, 1, batch_size, batch, tsp, active_sigmask);
+        if (may_sleep) {
+            _r._total_sleep += sched_clock::now() - before_getevents;
+        }
```
avikivity, line 491: How come only one backend gets this adjustment?

avikivity:
The changes look reasonable, but isn't steal time terminally broken?

thread_cputime_clock measures both kernel and user time. (Wall time) - (sleep time) overlaps that, since some of the sleep time is spent in the sleeping system call. For epoll/linux-aio this would usually be tiny, but for io_uring it could be quite large since the same call is used both to push I/O requests and to sleep. Perhaps we should adjust the io_uring backend to only sleep if it has no events to push.

src/core/reactor_backend.cc
travisdowns: Probably because it's the one we use currently. This change is from a year or more ago and was dusted off. Let me check the other backends.

travisdowns:
> thread_cputime_clock measures both kernel and user time. (Wall time) - (sleep time) overlaps that, since some of the sleep time is spent in the sleeping system call. For epoll/linux-aio this would usually be tiny

Yes exactly, it has that flow. Another "bug" is that we can't actually even tell if we are going to sleep in the system call: `may_sleep` can be true, but we can find IO right away in which case the `getevents` call will return and we did not sleep at all but all this work got counted as sleep and this problem can repeat at high frequency for some IO patterns. However, in practice steal has been quite useful for us in a variety of incidents, even in its flawed state.

However, for aio specifically this provided good results in practice. I don't have good numbers here since this is an old change but based on my long-ago testing pushing down the sleep calculation reduced negative steal to a very small amount, often rounding to 0 (in ms) over the sample period. If there was almost any "true" steal it is now almost always enough to overwhelm the negative steal part. In practice under load we almost always see significant "true" steal even if nothing else is running due to the syscall thread which is constantly working to handle EAGAIN IOs :(.

So this is why we need the second part of this change which simply discards negative steal periods (note that is is different from simply preventing the steal counter from going down).

> but for io_uring it could be quite large since the same call is used both to push I/O requests and to sleep. Perhaps we should adjust the io_uring backend to only sleep if it has no events to push.

True, though I don't think we will be worse off after this change. We could also do as you say, but this is making extra syscalls for the purpose of better steal time tracking. Not sure about that tradeoff.

src/core/reactor_backend.cc
travisdowns: Right, it's slowly coming back: updating the other reactor backends was part of why I never upstreamed this in the first place. Doing that now.

travisdowns:
Here are some results for a really bad case for the existing steal time calculation. This is running `io_tester` with [this patch](https://gist.github.com/travisdowns/740b78b2c59f2282b7e42aa1e88afcab) which prints out steal time metrics periodically and this config:

```
- name: steal_printer
  shards: all
  type: steal
  shard_info:
    parallelism: 1
    think_time: 10ms

- name: big_writes
  shards: all
  type: seqwrite
  shard_info:
    parallelism: 1
    reqsize: 1kB
    # shares: 10
    think_time: 0us
```

and running it like:

```
ninja io_tester && apps/io_tester/io_tester --duration=10 --conf apps/io_tester/conf.yaml --smp=2 --reactor-backend=linux-aio --idle-poll-time-us=0 --poll-aio=1
```

Prior to this change, the output is:

```
Starting evaluation...
0 steal: -42.983 ms awake: 56.688 cpu: 99.671
1 steal: -41.848 ms awake: 57.782 cpu: 99.630
0 steal: -42.842 ms awake: 56.878 cpu: 99.721
1 steal: -43.470 ms awake: 56.266 cpu: 99.737
1 steal: -43.937 ms awake: 55.791 cpu: 99.727
0 steal: -44.238 ms awake: 55.531 cpu: 99.769
1 steal: -42.527 ms awake: 57.136 cpu: 99.663
```

I.e., every 100 ms, we see ~40 ms of steal, because we think we are awake only ~60ms. Of course, we know that this workload with 0 think time is actually awake 100% of the time because as soon as any IO finishes we submit another one.

After the fix, the output is:

```
Starting evaluation...
0 steal:  0.445 ms awake: 100.129 cpu: 99.684
1 steal:  0.415 ms awake: 100.097 cpu: 99.682
0 steal:  0.374 ms awake: 100.135 cpu: 99.761
1 steal:  0.396 ms awake: 100.126 cpu: 99.729
0 steal:  0.419 ms awake: 100.147 cpu: 99.728
1 steal:  0.414 ms awake: 100.121 cpu: 99.707
0 steal:  0.428 ms awake: 100.151 cpu: 99.723
1 steal:  0.413 ms awake: 100.133 cpu: 99.720
0 steal:  0.442 ms awake: 100.152 cpu: 99.710
1 steal:  0.420 ms awake: 100.119 cpu: 99.699
...
```

I.e., we correctly identify that we are awake all the team and steal is low and positive (this is probably real steal, it's running on my noisy laptop).

When I run a `stress -c 0` thread using 100% on shard 0, we get output like (after the fix):

```
Starting evaluation...
1 steal:  0.424 ms awake: 100.137 cpu: 99.713
0 steal: 63.933 ms awake: 120.246 cpu: 56.314
1 steal:  0.595 ms awake: 100.165 cpu: 99.570
0 steal: 48.330 ms awake: 106.006 cpu: 57.676
1 steal:  0.619 ms awake: 100.153 cpu: 99.534
0 steal: 63.599 ms awake: 118.001 cpu: 54.402
1 steal:  0.445 ms awake: 100.135 cpu: 99.690
0 steal: 48.325 ms awake: 105.988 cpu: 57.663
1 steal:  0.578 ms awake: 100.156 cpu: 99.577
0 steal: 63.855 ms awake: 117.987 cpu: 54.132
1 steal:  0.528 ms awake: 100.166 cpu: 99.638
0 steal: 48.104 ms awake: 106.009 cpu: 57.906
1 steal:  0.496 ms awake: 100.147 cpu: 99.651
0 steal: 63.986 ms awake: 117.964 cpu: 53.978
```

I.e., it is correctly showing 50% steal on shard 0, as we'd expect from a single competing 100% CPU thread. Pre fix, it varies from 20% to 40% steal.

travisdowns:
For `io_uring` the benefit is much less, probably because don't have this `might_sleep` logic, where we conditionally pass a timeout or not: we pass a timeout unconditionally so we always count the `io_uring_wait_cqes` time as sleeping.

Here's the same output for `io_uring` (with 1000ms output interval now to provide some more averaging as the numbers bounce around a lot):

pre-fix:
```
0 steal: -69.508 ms awake: 95.746 cpu: 165.254
1 steal: -82.224 ms awake: 117.205 cpu: 199.429
0 steal: -69.770 ms awake: 97.488 cpu: 167.258
1 steal: -82.169 ms awake: 116.266 cpu: 198.435
0 steal: -70.218 ms awake: 96.821 cpu: 167.039
1 steal: -74.251 ms awake: 103.012 cpu: 177.262
0 steal: -70.684 ms awake: 99.314 cpu: 169.998
1 steal: -66.691 ms awake: 90.682 cpu: 157.372
0 steal: -68.950 ms awake: 95.269 cpu: 164.219
1 steal: -69.599 ms awake: 94.950 cpu: 164.549
0 steal: -73.325 ms awake: 104.437 cpu: 177.763
1 steal: -69.886 ms awake: 95.770 cpu: 165.656
0 steal: -72.801 ms awake: 105.222 cpu: 178.023
1 steal: -71.757 ms awake: 101.039 cpu: 172.796
0 steal: -69.816 ms awake: 100.048 cpu: 169.864
1 steal: -73.340 ms awake: 105.095 cpu: 178.435
0 steal: -69.226 ms awake: 97.243 cpu: 166.469
1 steal: -71.744 ms awake: 98.021 cpu: 169.765
```

post-fix:

```
Starting evaluation...
0 steal: -36.019 ms awake: 136.256 cpu: 172.275
1 steal: -38.107 ms awake: 141.260 cpu: 179.367
0 steal: -34.019 ms awake: 128.660 cpu: 162.679
1 steal: -33.000 ms awake: 123.472 cpu: 156.472
1 steal: -33.379 ms awake: 124.124 cpu: 157.503
0 steal: -37.188 ms awake: 134.169 cpu: 171.357
1 steal: -40.808 ms awake: 141.996 cpu: 182.804
0 steal: -33.975 ms awake: 126.294 cpu: 160.270
1 steal: -41.464 ms awake: 143.828 cpu: 185.292
0 steal: -33.172 ms awake: 123.730 cpu: 156.902
1 steal: -34.853 ms awake: 121.107 cpu: 155.960
0 steal: -35.433 ms awake: 130.999 cpu: 166.432
1 steal: -34.941 ms awake: 128.392 cpu: 163.333
0 steal: -35.278 ms awake: 130.750 cpu: 166.028
1 steal: -33.433 ms awake: 126.931 cpu: 160.364
0 steal: -33.842 ms awake: 129.780 cpu: 163.622
1 steal: -31.123 ms awake: 119.651 cpu: 150.773
0 steal: -36.534 ms awake: 135.890 cpu: 172.424
```

So there is a definite reduction in false steal, from about 6-7% to about 3.5% but it's still significant for that backend.

Furthermore, this isn't as much of a slam dunk for epoll or io_uring backends since we need to take a timestamp where we didn't before, which has some cost.

The negative steal is about the same for `io_uring` even w/o the IO (or without the `--idle-poll-time-us=0`) arg, and the improvement in this fix is about the same.

src/core/reactor_backend.cc
travisdowns: In the last push this is fixed for all 3 reactor backends.

travisdowns:
I wanted to add that for all backends we will in some cases (the case where we get to the "sleep" function) make 2 timestamp calls instead of 1, since before the "after" timestamp was shared with the idle_end logic, but now it is not.

However, in other cases we will take 0 timestamps instead of 1, because in `try_sleep` we abandon trying to sleep in many cases and don't ever get to the sleep location. For aio I expect this latter case to be very common since we don't go to sleep if there is any outstanding IO in the default `--poll-aio=1` mode. For other backends it less clear.

Both of these cases are on the "idle" or "about to sleep" paths though so probably less performance sensitive.

travisdowns:
@avikivity or @xemul let me know if there's anything I can do to help move this along or also if it's DOA etc.

avikivity:
> @avikivity or @xemul let me know if there's anything I can do to help move this along or also if it's DOA etc.

@xemul can you handle it?

travisdowns:
@xemul any hope that you can take a peek at this one? In our experience this is a big improvement to the accuracy and hence usefulness of the steal time metric.

xemul:
Sorry for losing track of this one :( I'll review it shortly

src/core/reactor.cc
```diff
@@ -2494,8 +2494,14 @@ void reactor::register_metrics() {
             sm::make_gauge("utilization", [this] { return (1-_load)  * 100; }, sm::description("CPU utilization")),
             sm::make_counter("cpu_busy_ms", [this] () -> int64_t { return total_busy_time() / 1ms; },
                     sm::description("Total cpu busy time in milliseconds")),
+            sm::make_counter("sleep_time_ms_total", [this] () -> int64_t { return _total_sleep / 1ms; },
+                    sm::description("Total reactor sleep time (wall clock)")),
+            sm::make_counter("awake_time_ms_total", [this] () -> int64_t { return total_awake_time() / 1ms; },
+                    sm::description("Total reactor awake time (wall_clock)")),
+            sm::make_counter("cpu_used_time_ms", [this] () -> int64_t { return total_cpu_time() / 1ms; },
+                    sm::description("Total reactor thread CPU time (from CLOCK_THREAD_CPUTIME)")),
```
xemul, line 2497: OK, so in these terms:

> awake-busy
> awake-idle
> sleeping

The (existing) cpu_busy_time is awake_busy, the (new) awake_time_ms_total is awake_body + awake_idle. Is that correct?

src/core/reactor.cc
```diff
@@ -4748,9 +4747,12 @@ std::chrono::nanoseconds reactor::total_steal_time() {
...
     // steal time but we have no ways to account it.
     //
+    // Furthermore, not all steal is from other processes: time used by the syscall thread and any
+    // alien threads will show up as steal as well as any time spent in a system call that
+    // unexpectedly blocked (since CPU time won't tick up when that occurs).
+    //
     // But what we have here should be good enough and at least has a well defined meaning.
-    return std::chrono::duration_cast<std::chrono::nanoseconds>(now() - _start_time - _total_sleep) -
-           std::chrono::duration_cast<std::chrono::nanoseconds>(thread_cputime_clock::now().time_since_epoch());
+    return total_awake_time() - total_cpu_time();
```
xemul: I'd really appreciate if this change went to previous (or separate, whichever is simpler for you) patch. It has nothing to do with updating the way _total_sleep is calculated, but is just re-using the helper methods introduced by previous change

src/core/reactor.cc
```diff
@@ -4746,7 +4746,33 @@ std::chrono::nanoseconds reactor::total_steal_time() {
...
+    // time since that measurement, clamped to 0. This means that as soon as steal time becomes 
+    // positive, it will be reflected in the measurement, rather than needing to "consume" all the
+    // accumulated negative steal time before positive steal times start showing up.
+
+
+    auto true_steal = total_awake_time() - total_cpu_time();
+    auto mono_steal = _last_mono_steal + std::max(true_steal - _last_true_steal, 0ns);
+
+    _last_true_steal = true_steal;
+    _last_mono_steal = mono_steal;
```
xemul, line 4773: In other words, the mono_steal is the sum of all positive changes in true_steal, and the "change" is calculated only at times when this function is called, right?

travisdowns: Yes, exactly. It does introduce an "observation effect" in some sense: the results may be different depending on how often you observe the metric, but it's the best I could come up with and seems like it works well in practice since at the observation points are where you care about this.

src/core/reactor.cc
travisdowns: Yes, exactly.

src/core/reactor.cc
travisdowns: Will do.

travisdowns:
https://github.com/scylladb/seastar/commit/b6fac76315e4ddf0dac9ba43823ee2e908bc1ea4 is purely a rebase to the tip of master.

src/core/reactor.cc
travisdowns: I have moved this to the prior commit in the most recent push.

src/core/reactor.cc
xemul: That's unfortunate :( Let's try to keep this "update rate" under some control. I can imagine two ways, none is perfect either.

First -- together with this place, update steal time (or maybe mono-steal part) when reactor stops sleeping.
Second -- arm a (lowres?) timer that will update the steal time, this getter only reports what had been accumulated so far.

travisdowns:
@xemul all feedback addressed or at least responded to.

travisdowns:
Thanks for the review @xemul !

PR #2393, opened 2024-08-20
remove vestigial OSv support

### LINKED ISSUE #2392: The future of OSv support
Will seastar continue to support OSv? 

Currently I believe this backend does not compile, e.g., this function:

```
void
reactor_backend_osv::wait_and_process_events(const sigset_t* sigset) {
    return process_events_nowait();
}
```

calls `process_events_nowait` which does not seem to exist. 

I ask in part because when making changes to reactor backends it is not clear if OSv should be included, and if it is I am not sure how to test or even compile those changes.

travisdowns:
Per #2392 this support is obsolete and can be removed.

This change removes all the code in the HAVE_OSV blocks, the README associated with OSV, updates one comment listing osv as one of the backends, and removes an asm linker directive which was OSv specific.

It also trims some spurious trailing whitespace in the modified files.

Fixes #2392.


src/net/virtio.cc
```diff
@@ -1033,10 +922,3 @@ std::unique_ptr<net::device> create_virtio_net_device(const virtio_options& opts
 }
 
 }
-
-// Locks the shared object in memory and forces on-load function resolution.
-// Needed if the function passed to enable_interrupt() is run at interrupt
-// time.
-// TODO: Instead of doing this, _virtio.enable_interrupt() could take a
-// pollable to wake instead of a function, then this won't be needed.
-asm(".pushsection .note.osv-mlock, \"a\"; .long 0, 0, 0; .popsection");
```
travisdowns, line 1042: This is the only bit I'm 95% sure about instead of 100%. This was being compiled even without HAVE_OSV, adding a note section to the binary and indeed I can find it in the redpanda binary:

```
[ 3] .note.osv-mlock   NOTE             000000000000038c  0000038c
```

However, my very brief research indicates it is OSv-only. Can I get a confirmation?

PR #2400, opened 2024-08-23
prometheus: sanitize label value for text protocol

pgellert:
This PR fixes a bug in the prometheus text-based endpoint where a label value with a special character (backslash (\), double-quote (") or line feed (\n)) makes the whole metrics output unparseable by prometheus.

The prometheus text protocol requires that the label value to be escaped. Note that this is only needed for the text protocol and not for the protobuf-based prometheus protocol.

> label_value can be any sequence of UTF-8 characters, but the backslash (\), double-quote ("), and line feed (\n) characters have to be escaped as \\, \", and \n, respectively.

From https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md

This also adds a new test (`prometheus_http_test.cc`) to validate the output of the text-based prometheus response. I've introduced this because it was simpler to make explicit checks on the text output than modifying `prometheus_test.py` and `metrics_tester.cc` to do that. I am open to give a go at working this into `prometheus_test.py` if you prefer that.

Finally, we switch to using `fmt::print` to format the metric value and the metric type instead of the originally used `std::to_string`. `std::to_string` is going to change its formatting behaviour in C++26 and so we prefer a more deterministic formatter here. The formatting matters because if the floating point numbers are converted to scientific notation, they can lose precision.

As an added benefit, using `fmt::print` writes these values to the output stream directly which means that we avoid a string copy when writing these values to the prometheus text-based response.

### DIFF
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -19,6 +19,7 @@
  * Copyright (C) 2016 ScyllaDB
  */
 
+#include <fmt/core.h>
 #include <seastar/core/prometheus.hh>
 #include <google/protobuf/io/coded_stream.h>
 #include <google/protobuf/io/zero_copy_stream_impl_lite.h>
@@ -198,33 +199,35 @@ static void fill_metric(pm::MetricFamily& mf, const metrics::impl::metric_value&
     }
 }
 
-static std::string to_str(seastar::metrics::impl::data_type dt) {
+static std::ostream& operator<<(std::ostream& os, seastar::metrics::impl::data_type dt) {
     switch (dt) {
     case seastar::metrics::impl::data_type::GAUGE:
-        return "gauge";
+        return os << "gauge";
     case seastar::metrics::impl::data_type::COUNTER:
     case seastar::metrics::impl::data_type::REAL_COUNTER:
-        return "counter";
+        return os << "counter";
     case seastar::metrics::impl::data_type::HISTOGRAM:
-        return "histogram";
+        return os << "histogram";
     case seastar::metrics::impl::data_type::SUMMARY:
-        return "summary";
+        return os << "summary";
     }
-    return "untyped";
+    return os << "untyped";
 }
 
-static std::string to_str(const seastar::metrics::impl::metric_value& v) {
+static std::ostream& operator<<(std::ostream& os, const seastar::metrics::impl::metric_value& v) {
     switch (v.type()) {
     case seastar::metrics::impl::data_type::GAUGE:
     case seastar::metrics::impl::data_type::REAL_COUNTER:
-        return std::to_string(v.d());
+        fmt::print(os, "{:.6f}", v.d());
+        break;
     case seastar::metrics::impl::data_type::COUNTER:
-        return std::to_string(v.i());
+        fmt::print(os, "{}", v.i());
+        break;
     case seastar::metrics::impl::data_type::HISTOGRAM:
     case seastar::metrics::impl::data_type::SUMMARY:
         break;
     }
-    return ""; // we should never get here but it makes the compiler happy
+    return os;
 }
 
 /*
@@ -740,20 +743,19 @@ public:
     }
 };
 
-std::string get_value_as_string(std::stringstream& s, const mi::metric_value& value) noexcept {
+void write_value_as_string(std::stringstream& s, const mi::metric_value& value) noexcept {
     std::string value_str;
     try {
-        value_str = to_str(value);
+        s << value;
     } catch (const std::range_error& e) {
-        seastar_logger.debug("prometheus: get_value_as_string: {}: {}", s.str(), e.what());
-        value_str = "NaN";
+        seastar_logger.debug("prometheus: write_value_as_string: {}: {}", s.str(), e.what());
+        s << "NaN";
     } catch (...) {
         auto ex = std::current_exception();
         // print this error as it's ignored later on by `connection::start_response`
-        seastar_logger.error("prometheus: get_value_as_string: {}: {}", s.str(), ex);
+        seastar_logger.error("prometheus: write_value_as_string: {}: {}", s.str(), ex);
         std::rethrow_exception(std::move(ex));
     }
-    return value_str;
 }
 
 future<> write_text_representation(output_stream<char>& out, const config& ctx, const metric_family_range& m, bool show_help, bool enable_aggregation, std::function<bool(const mi::labels_type&)> filter) {
@@ -775,7 +777,7 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx, 
                     if (show_help && metric_family.metadata().d.str() != "") {
                         s << "# HELP " << name << " " <<  metric_family.metadata().d.str() << '\n';
                     }
-                    s << "# TYPE " << name << " " << to_str(metric_family.metadata().type) << '\n';
+                    s << "# TYPE " << name << " " << metric_family.metadata().type << '\n';
                     found = true;
                 }
                 if (should_aggregate) {
@@ -786,7 +788,8 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx, 
                     write_histogram(s, ctx, name, value.get_histogram(), value_info.id.labels());
                 } else {
                     add_name(s, name, value_info.id.labels(), ctx);
-                    s << get_value_as_string(s, value) << '\n';
+                    write_value_as_string(s, value);
+                    s << '\n';
                 }
                 out.write(s.str()).get();
                 thread::maybe_yield();
@@ -799,7 +802,8 @@ future<> write_text_representation(output_stream<char>& out, const config& ctx, 
                         write_histogram(s, ctx, name, h.second.get_histogram(), h.first);
                     } else {
                         add_name(s, name, h.first, ctx);
-                        s << get_value_as_string(s, h.second) << '\n';
+                        write_value_as_string(s, h.second);
+                        s << '\n';
                     }
                     out.write(s.str()).get();
                     thread::maybe_yield();
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -35,6 +35,7 @@
 #include <seastar/core/thread.hh>
 #include <seastar/core/loop.hh>
 #include <regex>
+#include <string_view>
 
 namespace seastar {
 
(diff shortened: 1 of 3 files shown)


tests/unit/prometheus_http_test.cc
```diff
@@ -0,0 +1,92 @@
...
+    }
+};
+
+future<> test_prometheus_metrics_body() {
+    test_metrics metrics;
+    metrics.setup_metrics();
+
+    co_await seastar::async([] {
+        loopback_connection_factory lcf(1);
+        http_server server("test");
```
tchaikov, line 58: nit, could use 
```c++
auto stop_server = deferred_stop(server);
```

tchaikov (approved):
just some nits, lgtm.

tchaikov:
@amnonh hey Amnon, would you please take a look?

amnonh:
@pgellert I like the idea and the optimization.  It will be better to perform the label clean up on creation, so it can be done once. I would even convider throwing an exception in that case

pgellert:
[Force-push](https://github.com/scylladb/seastar/compare/d156991b7af964b21029d15474ac341b200971bd..6bf59588c547be84d35ae49aeb58a33425f97f97): improve test code based on code review feedback

pgellert:
@tchaikov @amnonh thank you for the reviews. 

> It will be better to perform the label clean up on creation, so it can be done once. I would even convider throwing an exception in that case

This one is trickier to implement than it first seems. The main complication is that the escaping must apply to the label values only when serialized with the text protocol, but not when serialized with protobuf. I wanted to add a new field `_escaped_value` to `label_instance` but it's not that simple, the label instances associated with a metric are converted into a `labels_type` and it's difficult to trace which label representations are used at what points on the text-based serialization path and the protobuf-based serialization path. I am also not sure the tests would catch if I made a mistake during the refactoring necessary to make that change, so I am tempted to keep the change simpler and hold off with this optimization. Let me know what you think.

tests/unit/prometheus_http_test.cc
```diff
@@ -0,0 +1,92 @@
...
+        future<> client = seastar::async([&lsi] {
+            connected_socket c_socket = lsi.connect(socket_address(ipv4_addr()), socket_address(ipv4_addr())).get();
+            input_stream<char> input(c_socket.input());
+            output_stream<char> output(c_socket.output());
+
+            output.write(sstring("GET /metrics HTTP/1.1\r\nHost: test\r\n\r\n")).get();
+            output.flush().get();
+            auto resp = input.read().get();
+            auto resp_str = std::string(resp.get(), resp.size());
+            BOOST_REQUIRE_NE(resp_str.find("200 OK"), std::string::npos);
```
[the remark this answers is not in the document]
pgellert: Nice, done

tests/unit/prometheus_http_test.cc
pgellert: I couldn't use this one because the `stop()` method of `http_server` is not `noexcept` so it doesn't satisfy the `stoppable` concept required for `deferred_stop`.

amnonh:
I disagree, protobuf and text should be the same, I would seriously
concider throwing on non standard labels

mykaul:
I think we should do it for both - we do not want to be in the situation where one decides to switch from one to another and then be surprised by the result.

src/core/prometheus.cc
```diff
@@ -198,33 +199,35 @@ static void fill_metric(pm::MetricFamily& mf, const metrics::impl::metric_value&
...
+    return os << "untyped";
 }
 
-static std::string to_str(const seastar::metrics::impl::metric_value& v) {
+static std::ostream& operator<<(std::ostream& os, const seastar::metrics::impl::metric_value& v) {
     switch (v.type()) {
     case seastar::metrics::impl::data_type::GAUGE:
     case seastar::metrics::impl::data_type::REAL_COUNTER:
-        return std::to_string(v.d());
+        fmt::print(os, "{:.6f}", v.d());
```
avikivity, line 221: I don't think .6f is a good format. For small values, it loses precision. For large values, (beyond 10^(std::numeric_limit::digits10-6)), it has precision without accuracy.

I expect '{g}' will work and provide maximum accuracy, but haven't tested it.

src/core/prometheus.cc
```diff
@@ -198,33 +199,35 @@ static void fill_metric(pm::MetricFamily& mf, const metrics::impl::metric_value&
...
     case seastar::metrics::impl::data_type::COUNTER:
-        return std::to_string(v.i());
+        fmt::print(os, "{}", v.i());
+        break;
     case seastar::metrics::impl::data_type::HISTOGRAM:
     case seastar::metrics::impl::data_type::SUMMARY:
         break;
     }
-    return ""; // we should never get here but it makes the compiler happy
+    return os; // we should never get here but it makes the compiler happy
```
avikivity, line 230: Comment is now incorrect.

src/core/prometheus.cc
nyh: Should changing the value's precision be part of a patch that's supposedly about the labels?

I had in the past some bad experience (https://github.com/scylladb/scylladb/issues/19081#issuecomment-2147839659) with how Seastar's metrics convert double-precision to 6 digits of precision in obscure places, causing surprising effects. We need to be very careful about this. Note that using "to_string()" is even worse, because nobody knows what it really does... If I remember correctly (but I can't find a reference now), we even had a case where in the past some to-string function used to print 15 digits of precision but one day when we changed the way we use `fmt` suddenly it because 6 digits.

Ideally, we'd want to have regression tests for these precision issues, so we won't be surprised if one day the precision of some metric format regresses.

P.S. @avikivity , actually the default precision of printf's `%g` is 6 digits, similar to `%f` - the "g" is about scientific notation for large exponents - not about precision. But I don't know if fmt inherited printf's idea of what "g" is or invented its own - it's worth checking.

avikivity: This patch is already changing the precision, by changing to_string() to fmt::print("{:.6f").

If plain {g} doesn't work, then {:19g} (or whatever) can be substituted. Or even start with {} to maintain the status quo and improve it later.

nyh: @avikivity this is exactly my question - why does a patch for "sanitize label value for text protocol" need to change the formatting of numbers.

pgellert: > This patch is already changing the precision

My intention with this PR is to keep the existing formatting of floating point numbers, keeping this PR purely a refactoring when it comes to the formatting of the metric values. The goal is to avoid the coming changes to `std::to_string`. `std::to_string` effectively used `{:.6f}` for the formatting of `double`s and this is what I intend to preserve with this change.

From https://en.cppreference.com/w/cpp/string/basic_string/to_string:

> 7,8) Converts a floating point value to a string as if by std::sprintf(buf, "%f", value). (until C++26)

Then from https://en.cppreference.com/w/cpp/io/c/fprintf:

> f/F: Converts floating-point number to the decimal notation in the style [-]ddd.ddd.
Precision specifies the exact number of digits to appear after the decimal point character. The default precision is 6. In the alternative implementation decimal point character is written even if no digits follow it. For infinity and not-a-number conversion style see notes.

I'm happy to move this commit to another PR if you prefer.

I think keeping the metric formatting as the original `std::to_string`-like format (`{:.6f}`) would be best. An important edge case for which `{:.6f}` works but `{g}` or `{}` would be problematic is gauges that have large integer values (eg. a gauge with a unix timestamp as its value). Changing the formatting to scientific notation could make these integer values less accurate which would be confusing.

pgellert:
[Force-push](https://github.com/scylladb/seastar/compare/6bf59588c547be84d35ae49aeb58a33425f97f97..6e2f629c7ae927eb3be55e6c84d504eaccde8e7f): remove outdated comment

src/core/prometheus.cc
pgellert: Fixed

mykaul:
> @amnonh @mykaul:
> 
> > I disagree, protobuf and text should be the same, I would seriously concider throwing on non standard labels
> > Protobuf and text should be the same
> 
> Could you please elaborate in what sense they should be the same? Just to clarify, if a user of the library intends to create a label value with the string `X`, prometheus expects this label value to be serialised as `escaped(X)` on the text protocol whereas it expects it to be serialized as the string field set to `X` for protobuf. These are the rules of the two independent serialisation protocols. If we set the protobuf field to `escaped(X)`, prometheus would think that the user set this field to `escaped(X)` and not `X`. So to show the user-intended label value on the prometheus UI and other metrics frontends the label value needs to be correctly serialized according to the protocol, escaped on the text output but not escaped on the protobuf output.
> 
> Prometheus supports label values with special characters so I don't see why seastar should restrict the range of label values that it supports.

Since users are free to switch between the protocols, we do not want the end user experience to be different when changing protocol. Unless I misunderstand the end result - would the user see the same label value eventually (say, in a query in Grafana) when switching from native to text?

amnonh:
> @amnonh @mykaul:
> 
> Prometheus supports label values with special characters so I don't see why seastar should restrict the range of label values that it supports.

Label and their values are static on creation, now, we'll need to call escape each time we send them, it's a shame. Was there an actual demand for special characters?

pgellert:
I've changed the label escaping now to only happen once, on creation. Let me know if the changes look good, and if the changes in the header files are within the source compatibility requirements of seastar.

> Was there an actual demand for special characters?

Yes, in redpanda we have labels that could contain these special characters (eg. consumer group names).

src/core/prometheus.cc
avikivity: > > This patch is already changing the precision
> 
> My intention with this PR is to keep the existing formatting of floating point numbers, keeping this PR purely a refactoring when it comes to the formatting of the metric values. The goal is to avoid the coming changes to `std::to_string`. `std::to_string` effectively used `{:.6f}` for the formatting of `double`s and this is what I intend to preserve with this change.


Absolutely the correct approach.

> 
> From https://en.cppreference.com/w/cpp/string/basic_string/to_string:
> 
> > 7,8) Converts a floating point value to a string as if by std::sprintf(buf, "%f", value). (until C++26)
> 
> Then from https://en.cppreference.com/w/cpp/io/c/fprintf:
> 
> > f/F: Converts floating-point number to the decimal notation in the style [-]ddd.ddd.
> > Precision specifies the exact number of digits to appear after the decimal point character. The default precision is 6. In the alternative implementation decimal point character is written even if no digits follow it. For infinity and not-a-number conversion style see notes.
> 
> I'm happy to move this commit to another PR if you prefer.
> 
> I think keeping the metric formatting as the original `std::to_string`-like format (`{:.6f}`) would be best. An important edge case for which `{:.6f}` works but `{g}` or `{}` would be problematic is gauges that have large integer values (eg. a gauge with a unix timestamp as its value). Changing the formatting to scientific notation could make these integer values less accurate which would be confusing.

Let's plan a follow-up to improve floating-point formatting. .6f is pretty bad. It's okay for a gauge, but as soon as you take a derivative of a small slowly-changing metric, you get poor results.

include/seastar/core/metrics_api.hh
```diff
@@ -41,7 +41,7 @@ namespace seastar {
 namespace metrics {
 namespace impl {
 
-using labels_type = std::map<sstring, sstring>;
+using labels_type = std::map<sstring, label_value>;
```
avikivity, line 44: I believe that's exposed as an API and will break scylladb unit tests that verify metrics.

pgellert:
I've pushed some more changes to try to keep the type alias `labels_type` around in its original form.

Note that there are still API-incompatible changes in the last commit, namely:
 * The public field `metric_definition_impl::labels` changes its type. I intentionally didn't duplicate this field because it is public and mutable. I would rather have users fail to compile their code than silently break their code if the user only modifies `metric_definition_impl::labels` but doesn't modify a new field `metric_definition_impl::full_labels`.
 * The same reasoning applies for `metric_id::labels()` which exposes `metric_id::_labels` and allows its mutation externally.

@amnonh @avikivity could you please re-review and advise on whether the optimization in the last commit is worth doing or if I should drop it since the blast radius of the change is quite large and there are still API-breaking changes in it? I would like to get this PR in if possible.

include/seastar/core/metrics_api.hh
pgellert: I attempted to avoid changing this in place now: https://github.com/scylladb/seastar/pull/2400#issuecomment-2340443138

avikivity:
@pgellert let's split the last patch off (I haven't reviewed it yet). The fixes are more important and can be queued immediately, we can consider the optimization after that.

pgellert:
[Force-push](https://github.com/scylladb/seastar/compare/3ed894fa3880d76056a7a37fae3a47248258f5e9..6e2f629c7ae927eb3be55e6c84d504eaccde8e7f): dropped the last commit based on @avikivity's feedback above

PR #2406, opened 2024-08-26
Add ThreadSanitizer support

devDDen:
ThreadSanitizer helps to detect data races. The PR adds TSan support to seastar and a build mode where it is enabled. It allows to run seastar applications with ThreadSanitizer.

During testing, the sanitizer reported data races in the following places:
- already fixed 48c6888
- [exception_hacks.cc](https://github.com/scylladb/seastar/compare/master...devDDen:add-tsan-support?expand=1#diff-d2b16640353e324e5def6b7a674bb6f511f26cb1000521b35c6a211c68cb9a15) in `_Unwind_RaiseException` (fixed here)
- in `tests/unit/*` in `BOOST_REQUIRE` (fixed here) due to Boost.Test macros are not thread safe. To get an accurate stack trace, the Boost.Test library must be built with ThreadSanitizer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,11 +182,45 @@ mark_as_advanced (
     CMAKE_EXE_LINKER_FLAGS_SANITIZE
     CMAKE_SHARED_LINKER_FLAGS_SANITIZE)
 
+set (CMAKE_CXX_FLAGS_SANITIZETHREAD
+  "-Os -g"
+  CACHE
+  STRING
+  "Flags used by the C++ compiler during sanitize-thread builds."
+  FORCE)
+
+set (CMAKE_C_FLAGS_SANITIZETHREAD
+  "-Os -g"
+  CACHE
+  STRING
+  "Flags used by the C compiler during sanitize-thread builds."
+  FORCE)
+
+set (CMAKE_EXE_LINKER_FLAGS_SANITIZETHREAD
+  ""
+  CACHE
+  STRING
+  "Flags used for linking binaries during sanitize-thread builds."
+  FORCE)
+
+set (CMAKE_SHARED_LINKER_FLAGS_SANITIZETHREAD
+  ""
+  CACHE
+  STRING
+  "Flags used by the shared libraries linker during sanitize-thread builds."
+  FORCE)
+
+mark_as_advanced (
+    CMAKE_CXX_FLAGS_SANITIZETHREAD
+    CMAKE_C_FLAGS_SANITIZETHREAD
+    CMAKE_EXE_LINKER_FLAGS_SANITIZETHREAD
+    CMAKE_SHARED_LINKER_FLAGS_SANITIZETHREAD)
+
 set (CMAKE_BUILD_TYPE
   "${CMAKE_BUILD_TYPE}"
   CACHE
   STRING
-  "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel Dev Sanitize."
+  "Choose the type of build, options are: None Debug Release RelWithDebInfo Dev Sanitize SanitizeThread."
   FORCE)
 
 if (NOT CMAKE_BUILD_TYPE)
@@ -356,7 +390,7 @@ endif ()
 
 
 set (Seastar_TEST_ENVIRONMENT
-  "ASAN_OPTIONS=${Seastar_ASAN_OPTIONS};UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1;BOOST_TEST_CATCH_SYSTEM_ERRORS=no"
+  "ASAN_OPTIONS=${Seastar_ASAN_OPTIONS};UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1;TSAN_OPTIONS=halt_on_error=1;BOOST_TEST_CATCH_SYSTEM_ERRORS=no"
   CACHE
   STRING
   "Environment variables for running tests")
@@ -376,6 +410,12 @@ set (Seastar_SANITIZE
   CACHE
   STRING
   "Enable ASAN and UBSAN. Can be ON, OFF or DEFAULT (which enables it for Debug and Sanitize)")
+
+set (Seastar_SANITIZE_THREAD
+  "DEFAULT"
+  CACHE
+  STRING
+  "Enable TSAN. Can be ON, OFF or DEFAULT (which enables it for SanitizeThread)")
 
 set (Seastar_DEBUG_SHARED_PTR
   "DEFAULT"
@@ -888,6 +928,26 @@ if (condition)
       $<${condition}:Sanitizers::undefined_behavior>)
 endif ()
 
+tri_state_option (${Seastar_SANITIZE_THREAD}
+  DEFAULT_BUILD_TYPES "SanitizeThread"
+  CONDITION condition)
+if (condition)
+  if (NOT ThreadSanitizer_FOUND)
+    message (FATAL_ERROR "Thread sanitizer not found!")
+  endif ()
+  set (Seastar_Sanitizers_OPTIONS ${ThreadSanitizer_COMPILER_OPTIONS})
+  target_link_libraries (seastar
+    PUBLIC
+      $<${condition}:Sanitizers::thread>)
+  target_compile_definitions(seastar
+    PUBLIC
+      SEASTAR_NO_EXCEPTION_HACK)
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    # Disabled because tsan doesn't support std::atomic_thread_fence
+    list (APPEND Seastar_PRIVATE_CXX_FLAGS -Wno-error=tsan)
+  endif ()
+endif ()
+
 # We only need valgrind to find uninitialized memory uses, so disable
 # the leak sanitizer.
 # To test with valgrind run "ctest -T memcheck"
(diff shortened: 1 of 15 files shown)


avikivity:
Looks good, @tchaikov please review, especially the cmake parts.

cmake/FindSanitizers.cmake
```diff
@@ -45,13 +52,21 @@ file (READ ${CMAKE_CURRENT_LIST_DIR}/code_tests/Sanitizers_fiber_test.cc _saniti
...
 include (FindPackageHandleStandardArgs)
 
 find_package_handle_standard_args (Sanitizers
   REQUIRED_VARS
     Sanitizers_ADDRESS_COMPILER_OPTIONS
     Sanitizers_UNDEFINED_BEHAVIOR_COMPILER_OPTIONS)
 
+find_package_handle_standard_args (ThreadSanitizer
+  REQUIRED_VARS
+    ThreadSanitizer_COMPILER_OPTIONS)
```
tchaikov, line 68: in general, `FindSanitizers.cmake` should only check only for a package named "Sanitizers", not packages with other name. so i think a better alternative is to support the "COMPONENT" parameter. #2408 was created to address this.

configure.py
```diff
@@ -169,7 +169,7 @@ def identify_best_standard(cpp_standards, compiler):
 # For convenience.
 tr = seastar_cmake.translate_arg
 
-MODE_TO_CMAKE_BUILD_TYPE = {'release': 'RelWithDebInfo', 'debug': 'Debug', 'dev': 'Dev', 'sanitize': 'Sanitize' }
+MODE_TO_CMAKE_BUILD_TYPE = {'release': 'RelWithDebInfo', 'debug': 'Debug', 'dev': 'Dev', 'sanitize': 'Sanitize', 'sanitize-thread' : 'SanitizeThread'}
```
tchaikov, line 172: i'd suggest to add an option to `configure.py ` and cmake to list the enabled sanitizers. by default, if "Sanitize" is specified, both UBSan and ASan are enabled, but one can override the option by specifying it explicitly to use TSan, like 
```console
./configure.py --mode sanitize --sanitizer UndefinedBehavior --sanitizer Thread
./configure.py --mode sanitize --sanitizer Address
# -DCMAKE_BUILD_TYPE is but a convenient way to set a bunch of 
# preconfigured settings, but one can fine-tune the build with 
# these lower-level options
# just enable these two sanitizers without any predefined configurations
cmake -DCMAKE_BUILD_TYPE=None -DSeastar_SANITIZERS=Address;UndefinedBehavior
# inherit the settings of Sanitize, but customize the enabled Sanitizers
cmake -DCMAKE_BUILD_TYPE=Sanitize -DSeastar_SANITIZERS=Address;UndefinedBehavior
```

src/core/exception_hacks.cc
```diff
@@ -71,6 +71,8 @@ module seastar;
 #endif
 
 namespace seastar {
+
+#ifndef SEASTAR_NO_EXCEPTION_HACK
```
tchaikov, line 75: > restore SEASTAR_NO_EXCEPTION_HACK behavior

i'd suggest extract this into its own commit. and explain why we want to bring it back. and how this change helps.

CMakeLists.txt
```diff
@@ -963,6 +1023,10 @@ if (Sanitizers_FIBER_SUPPORT)
   list (APPEND Seastar_PRIVATE_COMPILE_DEFINITIONS SEASTAR_HAVE_ASAN_FIBER_SUPPORT)
 endif ()
 
+if (ThreadSanitizer_FIBER_SUPPORT)
+  list (APPEND Seastar_PRIVATE_COMPILE_DEFINITIONS SEASTAR_HAVE_TSAN_FIBER_SUPPORT)
```
tchaikov, line 1027: if this definition only applies to ThreadSanitizer, i'd suggest add this definition only if ThreadSanitizer is enabled. so we don't need to check both of them in the C++ code.

cmake/FindSanitizers.cmake
```diff
@@ -45,13 +52,21 @@ file (READ ${CMAKE_CURRENT_LIST_DIR}/code_tests/Sanitizers_fiber_test.cc _saniti
 set (CMAKE_REQUIRED_FLAGS ${Sanitizers_COMPILER_OPTIONS})
 check_cxx_source_compiles ("${_sanitizers_fiber_test_code}" Sanitizers_FIBER_SUPPORT)
 
+file (READ ${CMAKE_CURRENT_LIST_DIR}/code_tests/ThreadSanitizer_fiber_test.cc _thread_sanitizer_fiber_test_code)
+set (CMAKE_REQUIRED_FLAGS ${ThreadSanitizer_COMPILER_OPTIONS})
+check_cxx_source_compiles ("${_thread_sanitizer_fiber_test_code}" ThreadSanitizer_FIBER_SUPPORT)
```
tchaikov, line 57: as explained, i'd suggest define this variable only if ThreadSanitizer is listed in the COMPONENT parameter.

tchaikov (commented):
left some comments.

tests/unit/distributed_test.cc
```diff
@@ -142,7 +142,7 @@ SEASTAR_TEST_CASE(test_map_reduce_lifetime) {
         }
         auto operator()(const X& x) {
             return yield().then([this, &x] {
-                BOOST_REQUIRE(!destroyed);
+                assert(!destroyed);
```
tchaikov, line 145: > BOOST_REQUIRE macros are not thread safe, so replace them with
asserts in multithreaded environment

could you share the impact of "not thread safe" ?

src/core/thread.cc
```diff
@@ -48,28 +48,41 @@ namespace seastar {
 thread_local jmp_buf_link g_unthreaded_context;
 thread_local jmp_buf_link* g_current_context;
 
-#ifdef SEASTAR_ASAN_ENABLED
+#if defined(SEASTAR_ASAN_ENABLED) || defined(SEASTAR_TSAN_ENABLED)
 
 namespace {
 
-#ifdef SEASTAR_HAVE_ASAN_FIBER_SUPPORT
+#if defined(SEASTAR_ASAN_ENABLED) && defined(SEASTAR_HAVE_ASAN_FIBER_SUPPORT)
```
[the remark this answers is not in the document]
devDDen: because `#if` at higher level has changed
```diff
- #ifdef SEASTAR_ASAN_ENABLED
+ #if defined(SEASTAR_ASAN_ENABLED) || defined(SEASTAR_TSAN_ENABLED)
```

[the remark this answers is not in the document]
devDDen: I see the comment below about it

tests/unit/distributed_test.cc
devDDen: 1. ThreadSanitizer warns
2. [Boost.Test docs](https://live.boost.org/doc/libs/master/libs/test/doc/html/boost_test/section_faq.html) say about it

devDDen:
> @devDDen #2408 has been merged. do you plan to continue working on this change?

Yes, at first I plan to do https://github.com/scylladb/seastar/pull/2406#discussion_r1732215158 soon in a separate PR

tests/unit/distributed_test.cc
nyh: Normal Seastar tests run on a single reactor thread. But this specific test deliberately starts the same code on multiple threads. And then, supposedly, you're not supposed to use BOOST_REQUIRE from multiple threads.

However, I don't think replacing BOOST_REQUIRE by assert() is the right thing to do - it will replace a clean test failure of a single test function, by a crash of the entire executable... :-(

I think it more sense to have an (atomic-variable) counter of errors, and then BOOST_REQUIRE that the counter be 0 when the do_with_distributed ends (on a single thread).

nyh: BTW, I find it really shocking - that in 2024 Boost.Test doesn't support multiple threads out of the box. It's trivial to support (especially when performance is not even an issue). It reduces whatever little confidence I had in this test framework :-(

CMakeLists.txt
```diff
@@ -182,11 +182,45 @@ mark_as_advanced (
...
+  CACHE
+  STRING
+  "Flags used by the shared libraries linker during sanitize-thread builds."
+  FORCE)
+
+mark_as_advanced (
+    CMAKE_CXX_FLAGS_SANITIZETHREAD
+    CMAKE_C_FLAGS_SANITIZETHREAD
+    CMAKE_EXE_LINKER_FLAGS_SANITIZETHREAD
+    CMAKE_SHARED_LINKER_FLAGS_SANITIZETHREAD)
```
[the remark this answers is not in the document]
nyh: @tchaikov if you're thinking of enabling this sanitizer by default when sanitizers are enabled - please try to measure the performance impact of this sanitizer on some "typical" (however you define that) Seastar test code. The majority of the Seastar code (and code using Seastar) does *not* use share data between threads, so although catching every bug is great, I doubt this sanitizer will help catch many bugs so we need some justification to turn it on by

tchaikov:
> > @devDDen #2408 has been merged. do you plan to continue working on this change?
> 
> Yes, at first I plan to do [#2406 (comment)](https://github.com/scylladb/seastar/pull/2406#discussion_r1732215158) soon in a separate PR

thanks. please let me know if i can help.

CMakeLists.txt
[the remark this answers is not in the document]
tchaikov: > @tchaikov if you're thinking of enabling this sanitizer by default when sanitizers are enabled 

that's not my plan. my plan is to keep the existing behavior.

> - please try to measure the performance impact of this sanitizer on some "typical" (however you define that) Seastar test code. The majority of the Seastar code (and code using Seastar) does _not_ use share data between threads, so although catching every bug is great, I doubt this sanitizer will help catch many bugs so we need some justification to turn it on by

PR #2408, opened 2024-08-27
build: check the combination of Sanitizers

tchaikov:
we are going to add a build with ThreadSanitizer enabled, but ThreadSanitizer cannot be used along with AddressSanitizer. the existing sanitizer detection checks each sanitizer separately, so even if the sanitizers combination does not work, the detection still passes.

also, we hardwire two sanitizers in `FindSanitizers.cmake`, this is not extensible to the use case where we only need to selectively find a certain (combination) of sanitizers.

in order to address these problems, in this change

* find specified component(s) in FindSanitizers.cmake, to prepare for the change which selectively specifies a subset of sanitizers
* check if the compiler supports the combination of compile options required by all specified sanitizers
* explicitly specify the used sanitizers in `SeastarDependencies.cmake`.

### DIFF
--- a/cmake/FindSanitizers.cmake
+++ b/cmake/FindSanitizers.cmake
@@ -20,54 +20,63 @@
 # Copyright (C) 2018 Scylladb, Ltd.
 #
 
-include (CheckCXXSourceCompiles)
+if(NOT Sanitizers_FIND_COMPONENTS)
+  set(Sanitizers_FIND_COMPONENTS
+    address
+    undefined_behavior)
+endif()
 
-set (CMAKE_REQUIRED_FLAGS -fsanitize=address)
-check_cxx_source_compiles ("int main() {}" Sanitizers_ADDRESS_FOUND)
+foreach (component ${Sanitizers_FIND_COMPONENTS})
+  string (TOUPPER ${component} COMPONENT)
+  set (compile_options "Sanitizers_${COMPONENT}_COMPILE_OPTIONS")
+  if (component STREQUAL "address")
+    list (APPEND ${compile_options} -fsanitize=address)
+  elseif (component STREQUAL "undefined_behavior")
+    list (APPEND ${compile_options} -fsanitize=undefined)
+    # Disable vptr because of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88684
+    list (APPEND ${compile_options} -fno-sanitize=vptr)
+  else ()
+    message (FATAL_ERROR "Unsupported sanitizer: ${component}")
+  endif ()
+  list(APPEND Sanitizers_COMPILE_OPTIONS "${${compile_options}}")
+endforeach ()
 
-if (Sanitizers_ADDRESS_FOUND)
-  set (Sanitizers_ADDRESS_COMPILER_OPTIONS -fsanitize=address)
+include(CheckCXXSourceCompiles)
+include(CMakePushCheckState)
+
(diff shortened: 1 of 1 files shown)


cmake/FindSanitizers.cmake
```diff
@@ -20,54 +20,63 @@
 # Copyright (C) 2018 Scylladb, Ltd.
 #
 
-include (CheckCXXSourceCompiles)
+if(NOT Sanitizers_FIND_COMPONENTS)
+  set(Sanitizers_FIND_COMPONENTS
+    address
+    undefined_behavior)
+endif()
```
tchaikov, line 27: this preserves the backward compatibility if some parent projects use `find_package(Sanitizers)` without specifying the `COMPONENTS` parameter.

cmake/FindSanitizers.cmake
```diff
@@ -20,54 +20,63 @@
...
-endif ()
+foreach (component ${Sanitizers_FIND_COMPONENTS})
+  string (TOUPPER ${component} COMPONENT)
+  set (compile_options "Sanitizers_${COMPONENT}_COMPILE_OPTIONS")
+  if (component STREQUAL "address")
+    list (APPEND ${compile_options} -fsanitize=address)
+  elseif (component STREQUAL "undefined_behavior")
+    list (APPEND ${compile_options} -fsanitize=undefined)
+    # Disable vptr because of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88684
+    list (APPEND ${compile_options} -fno-sanitize=vptr)
```
avikivity, line 37: Note: this was fixed in gcc 10 (or perhaps 9?) - https://github.com/gcc-mirror/gcc/commit/c24847a54abe6b93139f92318f254d1952315ba2 so we can drop it later.

cmake/SeastarDependencies.cmake
```diff
@@ -141,6 +141,10 @@ macro (seastar_find_dependencies)
   seastar_set_dep_args (rt REQUIRED)
   seastar_set_dep_args (numactl
     OPTION ${Seastar_NUMA})
+  seastar_set_dep_args (Sanitizers
+    COMPONENTS
+      address
+      undefined_behavior)
   seastar_set_dep_args (ucontext REQUIRED)
```
avikivity, line 148: Is this complexity typical for projects that use sanitizers? I don't mind merging it, but isn't cmake supposed to do this for us?

The code is incomprehensible to me.

tchaikov: i wanted to make it explicit. yes, it's taken care of. see https://github.com/scylladb/seastar/pull/2408#discussion_r1734976089 . but with #2406, we will need to specify it anyway.

tchaikov: anyway, i dropped this change in the latest revision.

tchaikov:
@avikivity mind taking another look?

tchaikov:
v2:

- do not specify the "COMPONENTS" for Sanitizers in `cmake/SeastarDependencies.cmake`

cmake/FindSanitizers.cmake
tchaikov: sure. will do in a separate PR.

tchaikov:
@scylladb/seastar-maint and @avikivity could you help review this change?

cmake/FindSanitizers.cmake
tchaikov: created #2430

Issue #2409, opened 2024-08-27
Make http::client::make_request() abortable


xemul:
Currently the method can block in one of 4 places:
- connecting to server
- waiting for a connection from pool
- writing request to socket
- reading response from server

Either of the steps should be abortable. One of the use-cases is to time-out the request (see #2304). Another -- abort of large file upload in case this upload by user request (e.g. cancelling backup process for whatever reason)

ndragazis:
@xemul the first bullet was never fully implemented, right? In https://github.com/scylladb/seastar/commit/35badf56f1d145de520c6e1aa3f4526492026b88 you mention that it will be done as a followup. Do we have an issue about this?

xemul:
@ndragazis , yes, connect() still blocks. We don't have special issue for it, but we have known issue _with_ it: https://github.com/scylladb/seastar/issues/2302

PR #2410, opened 2024-08-27
Make http::client::make_request() abortable

xemul:
This PR adds abort_source& argument to client::make_request() method. With it, calling abort_source::request_abort() will break the request making process wherever it is and resolve the make_request() with abort_requested_exception exception.

One of the use-cases is to let client impose a timeout on the make_request(). For that, a timer should be armed that would request abort on the abort source passed to the method. In #2304 there was an attempt to put a timeout on connect() only, but this approach is more generic and allows aborting request at any stage, not only connecting.

Another scenario is stopping long operations on user request. There can be several examples of what "long" operation can be. One of them is simple connect() to the server -- TCP times out after 1 minute by default, and breaking this earlier can be useful. Another example of long operation is S3 file uploading. It may happen in "parts" and part can be of any size, so simple writing of the request body into the socket may take time.

fixes: #2409 
closes: #2304

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -132,6 +132,8 @@ private:
     future<reply_ptr> maybe_wait_for_continue(const request& req);
     future<> write_body(const request& rq);
     future<reply_ptr> recv_reply();
+
+    void shutdown() noexcept;
 };
 
 /**
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -185,19 +185,20 @@ private:
 
     using connection_ptr = seastar::shared_ptr<connection>;
 
-    future<connection_ptr> get_connection();
-    future<connection_ptr> make_connection();
+    future<connection_ptr> get_connection(abort_source* as);
+    future<connection_ptr> make_connection(abort_source* as);
     future<> put_connection(connection_ptr con);
     future<> shrink_connections();
 
     template <std::invocable<connection&> Fn>
-    auto with_connection(Fn&& fn);
+    auto with_connection(Fn&& fn, abort_source*);
 
     template <typename Fn>
     requires std::invocable<Fn, connection&>
-    auto with_new_connection(Fn&& fn);
+    auto with_new_connection(Fn&& fn, abort_source*);
 
-    future<> do_make_request(connection& con, request& req, reply_handler& handle, std::optional<reply::status_type> expected);
+    future<> do_make_request(request req, reply_handler handle, abort_source*, std::optional<reply::status_type> expected);
+    future<> do_make_request(connection& con, request& req, reply_handler& handle, abort_source*, std::optional<reply::status_type> expected);
 
 public:
     /**
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -286,6 +286,18 @@ public:
     future<> make_request(request req, reply_handler handle, std::optional<reply::status_type> expected = std::nullopt);
 
     /**
+     * \brief Send the request and handle the response (abortable)
+     *
+     * Same as previous method, but aborts the request upon as.request_abort() call
+     *
+     * \param req -- request to be sent
+     * \param handle -- the response handler
+     * \param as -- abort source that aborts the request
+     * \param expected -- the optional expected reply status code, default is std::nullopt
+     */
+    future<> make_request(request req, reply_handler handle, abort_source& as, std::optional<reply::status_type> expected = std::nullopt);
+
+    /**
      * \brief Updates the maximum number of connections a client may have
      *
      * If the new limit is less than the amount of connections a client has, they will be
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -150,7 +150,7 @@ public:
      * The implementations of this method should return ready-to-use socket that will
      * be used by \ref client as transport for its http connections
      */
-    virtual future<connected_socket> make() = 0;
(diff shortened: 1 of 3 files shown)


xemul:
upd:
- fix c++20 compilation

src/http/client.cc
```diff
@@ -349,9 +350,6 @@ future<> client::make_request(request req, reply_handler handle, std::optional<r
                     return do_make_request(con, req, handle, expected);
                 });
             });
-        }
-
-        return f;
```
avikivity, line 354: A coroutine would be much better here

src/http/client.cc
```diff
@@ -346,14 +350,14 @@ future<> client::make_request(request req, reply_handler handle, std::optional<r
...
+            return with_new_connection([this, &req, &handle, as, expected] (connection& con) {
+                return do_make_request(con, req, handle, as, expected);
+            }, as);
         });
     });
 }
 
-future<> client::do_make_request(connection& con, request& req, reply_handler& handle, std::optional<reply::status_type> expected) {
+future<> client::do_make_request(connection& con, request& req, reply_handler& handle, abort_source* as, std::optional<reply::status_type> expected) {
     return con.do_make_request(req).then([&con, &handle, expected] (connection::reply_ptr reply) mutable {
```
avikivity, line 379: Does it not make sense to share the abort_source in the client class?

Why does each request need a separate abort_source?

src/http/client.cc
xemul: True

src/http/client.cc
xemul: Because the intent is to abort individual make_request, not the whole client. Scylla uses [single client for configured S3 endpoint](https://github.com/scylladb/scylladb/pull/13859). If there's a [keyspace with storage=s3](https://github.com/scylladb/scylladb/pull/10144) and [backup activity](https://github.com/scylladb/scylladb/pull/19890) towards the same endpoint, if user wants to abort only backup, it should abort only some subset of in-flight requests, not the whole client.

xemul:
upd:
- rebased (no conflicts)
- fix test c++20 compilation

src/http/client.cc
avikivity: Why not create a client per {set of related requests}?

xemul: Then we'll need to implement resource control in a different manner. Resources in the scope of s3 client are -- memory buffers (we flush sstables in 5Mb parts) and the number of sockets in client pool. Or leave different {sets of related requests} consume resources independently

xemul: @avikivity , reminder ping

avikivity: Sorry. Could we extract the resource controller out? So multiple clients can share the same resource pool?

Though it's not out of the question to want to abort a particular request.

include/seastar/http/client.hh
```diff
@@ -150,7 +152,7 @@ public:
      * The implementations of this method should return ready-to-use socket that will
      * be used by \ref client as transport for its http connections
      */
-    virtual future<connected_socket> make() = 0;
+    virtual future<connected_socket> make(abort_source*) = 0;
```
tchaikov, line 155: this broke the build of scylla
```
In file included from /home/kefu/dev/scylladb/tools/scylla-nodetool.cc:52:
/home/kefu/dev/scylladb/utils/http.hh:67:38: error: 'make' marked 'override' but does not override any member functions
   67 |     virtual future<connected_socket> make() override {
In file included from /home/kefu/dev/scylladb/tools/scylla-nodetool.cc:11:
In file included from /usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/chrono:49:
In file included from /usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/bits/shared_ptr.h:53:
In file included from /usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/bits/shared_ptr_base.h:59:
/usr/lib/gcc/x86_64-redhat-linux/14/../../../../include/c++/14/bits/unique_ptr.h:1076:34: error: allocating an object of abstract class type 'utils::http::dns_connection_factory'
 1076 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
      |                                  ^
/home/kefu/dev/scylladb/tools/scylla-nodetool.cc:185:28: note: in instantiation of function template specialization 'std::make_unique<utils::http::dns_connection_factory, seastar::basic_sstring<char, unsigned int, 15> &, unsigned short &, bool, seastar::logger &>' requested here
  185 |         , _api_client(std::make_unique<utils::http::dns_connection_factory>(_host, _port, false, nlog), 1)
      |                            ^
/home/kefu/dev/scylladb/seastar/include/seastar/http/client.hh:155:38: note: unimplemented pure virtual method 'make' in 'dns_connection_factory'
  155 |     virtual future<connected_socket> make(abort_source*) = 0;
2 errors generated.
```

@xemul

tchaikov: i am preparing a quick fix.

tchaikov: #2428

src/http/client.cc
xemul: > Sorry. Could we extract the resource controller out? So multiple clients can share the same resource pool?

Yes, it's possible. However, http::client is all about socket pool management, so sharing sockets pool is the same as sharing the http client itself.

> Though it's not out of the question to want to abort a particular request.

Yes. Also, putting abort source on client will not change much in the code, just instead of abort_source& argument there will be this->abort_source

PR #2412, opened 2024-08-28
cmake: Depend on libbsd if DPDK depends on it

### LINKED ISSUE #2411: Current master fails to build with DPDK on Debian stable (bookworm) if libbsd-dev is installed
Most likely caused by the following code in `./dpdk/lib/eal/include/rte_string_fns.h`:

```
#ifdef RTE_USE_LIBBSD
#include <bsd/string.h>

#else /* no BSD header files, create own */
#define strlcpy(dst, src, size) rte_strlcpy(dst, src, size)
#define strlcat(dst, src, size) rte_strlcat(dst, src, size)

#endif /* RTE_USE_LIBBSD */
#endif /* FREEBSD */
```

Reproduction steps:

```
sudo docker run -it debian:bookworm /bin/bash
```

Inside:
```
apt update
apt install git g++ cmake ninja-build
git clone https://github.com/scylladb/seastar
cd seastar
git submodule update --init
./install-dependencies.sh
apt install libbsd-dev
./configure.py --mode=release --enable-dpdk
ninja -C build/release
```

Errors such as the following are observed:

```
[294/357] Linking CXX executable apps/httpd/httpd
FAILED: apps/httpd/httpd 
: && /usr/bin/g++ -O2 -g -DNDEBUG  apps/httpd/CMakeFiles/app_httpd.dir/demo.json.cc.o apps/httpd/CMakeFiles/app_httpd.dir/main.cc.o -o apps/httpd/httpd  libseastar.a  /usr/lib/x86_64-linux-gnu/libboost_program_options.so  /usr/lib/x86_64-linux-gnu/libboost_thread.so  /usr/lib/x86_64-linux-gnu/libboost_chrono.so  /usr/lib/x86_64-linux-gnu/libboost_date_time.so  /usr/lib/x86_64-linux-gnu/libboost_atomic.so  /usr/lib/x86_64-linux-gnu/libcares.so  /usr/lib/x86_64-linux-gnu/libfmt.so.9.1.0  /usr/lib/x86_64-linux-gnu/liblz4.so  -ldl  /usr/lib/x86_64-linux-gnu/libgnutls.so  -latomic  /usr/lib/x86_64-linux-gnu/libsctp.so  /usr/lib/x86_64-linux-gnu/libprotobuf.so  /usr/lib/x86_64-linux-gnu/libyaml-cpp.so  /usr/lib/x86_64-linux-gnu/libhwloc.so  /usr/lib/x86_64-linux-gnu/liburing.so  /usr/lib/x86_64-linux-gnu/libnuma.so && :
/usr/bin/ld: libseastar.a(dpdk.o): in function `rte_pci_scan':
(.text+0x1a6e): undefined reference to `strlcpy'
/usr/bin/ld: libseastar.a(dpdk.o): in function `pci_uio_alloc_resource':
(.text+0x2831): undefined reference to `strlcpy'
/usr/bin/ld: libseastar.a(dpdk.o): in function `insert_vdev':
vdev.c:(.text+0x48d6): undefined reference to `strlcpy'
/usr/bin/ld: libseastar.a(dpdk.o): in function `vdev_action':
vdev.c:(.text+0x4a5c): undefined reference to `strlcpy'
/usr/bin/ld: libseastar.a(dpdk.o): in function `vdev_scan':
vdev.c:(.text+0x4f4f): undefined reference to `strlcpy'
...
```

p12tic:
Fixes #2411.

If libbsd-dev is installed during build time, DPDK will depend on it without a way to turn this off.

config/meson.build has the following:

```
libbsd = dependency('libbsd', required: false, method: 'pkg-config') if libbsd.found()
    dpdk_conf.set('RTE_USE_LIBBSD', 1)
endif
```

And lib/eal/include/rte_string_fns.h contains:

```
#ifdef RTE_USE_LIBBSD
#include <bsd/string.h>

#else /* no BSD header files, create own */
#define strlcpy(dst, src, size) rte_strlcpy(dst, src, size)
#define strlcat(dst, src, size) rte_strlcat(dst, src, size)

#endif /* RTE_USE_LIBBSD */
#endif /* FREEBSD */
```

Accordingly the only way to address this inflexibility is to ensure that dpdk dependencies are appropriately transferred to INTERFACE_LINK_LIBRARIES of the dpdk library.

### DIFF
--- a/cmake/Finddpdk.cmake
+++ b/cmake/Finddpdk.cmake
@@ -129,6 +129,13 @@ find_package_handle_standard_args (dpdk
   REQUIRED_VARS
     ${dpdk_REQUIRED})
 
+# Depending on whether libbsd-dev exists at build time, DPDK may be built with a dependency on
+# libbsd. Note that other libraries in dpdk_PC_LIBRARIES are handled using separate logic
+# (see rte_libs above), thus the additional dependencies must be handled on a case by case basis.
+if ("bsd" IN_LIST dpdk_PC_LIBRARIES)
+  list (APPEND dpdk_dependencies "bsd")
+endif ()
+
 if (dpdk_FOUND AND NOT (TARGET dpdk))
   get_filename_component (library_suffix "${dpdk_EAL_LIBRARY}" LAST_EXT)
   # strictly speaking, we should have being using check_c_compiler_flag()
@@ -163,6 +170,7 @@ if (dpdk_FOUND AND NOT (TARGET dpdk))
     set_target_properties (dpdk
       PROPERTIES
         INTERFACE_INCLUDE_DIRECTORIES ${dpdk_INCLUDE_DIR}
+        INTERFACE_LINK_LIBRARIES "${dpdk_dependencies}"
         IMPORTED_OBJECTS ${dpdk_object_path}
         ${compile_options})
     # we include dpdk in seastar already, so no need to expose it with
@@ -175,7 +183,7 @@ if (dpdk_FOUND AND NOT (TARGET dpdk))
     set_target_properties (DPDK::dpdk
       PROPERTIES
         INTERFACE_INCLUDE_DIRECTORIES "${dpdk_PC_INCLUDE_DIRS}"
-        INTERFACE_LINK_LIBRARIES "${_dpdk_libraries}"
+        INTERFACE_LINK_LIBRARIES "${_dpdk_libraries};${dpdk_dependencies}"
         ${compile_options})
   endif()
 endif ()


cmake/Finddpdk.cmake
```diff
@@ -84,6 +84,14 @@ endif ()
 list (APPEND dpdk_REQUIRED
   dpdk_INCLUDE_DIR)
 
+# Depending on whether libbsd-dev exists at build time, DPDK may be built with a dependency on
+# libbsd. Note that other libraries in dpdk_PC_LIBRARIES are handled using separate logic
+# (see rte_libs above), thus the additional dependencies must be handled on a case by case basis.
+set (dpdk_dependencies "")
```
tchaikov, line 90: no need to initialize `dpdk_dependencies` . after we `list(APPEND v "foo")` when `v` is not initialized at all, the value of `v` is `"foo"`.

cmake/Finddpdk.cmake
```diff
@@ -84,6 +84,14 @@ endif ()
 list (APPEND dpdk_REQUIRED
   dpdk_INCLUDE_DIR)
 
+# Depending on whether libbsd-dev exists at build time, DPDK may be built with a dependency on
+# libbsd. Note that other libraries in dpdk_PC_LIBRARIES are handled using separate logic
+# (see rte_libs above), thus the additional dependencies must be handled on a case by case basis.
```
tchaikov, line 89: thanks for the comment. i admit that we should improve the code to either consume the `dpdk_PC_LIBRARIES` in a better way instead of hardwiring them.

cmake/Finddpdk.cmake
```diff
@@ -84,6 +84,14 @@ endif ()
...
+set (dpdk_dependencies "")
+if ("bsd" IN_LIST dpdk_PC_LIBRARIES)
+  list (APPEND dpdk_dependencies "bsd")
+endif ()
```
tchaikov, line 93: i'd suggest move this to a place 

- either closer to where it is consumed: right before `if (dpdk_FOUND ...)`
- or closer to where `dpdk_PC_*` variables are initialized: right after `pkg_check_modules (dpdk_PC ...)` 

so that the data dependency are more clear to human readers.

cmake/Finddpdk.cmake
```diff
@@ -175,7 +184,7 @@ if (dpdk_FOUND AND NOT (TARGET dpdk))
     set_target_properties (DPDK::dpdk
       PROPERTIES
         INTERFACE_INCLUDE_DIRECTORIES "${dpdk_PC_INCLUDE_DIRS}"
-        INTERFACE_LINK_LIBRARIES "${_dpdk_libraries}"
+        INTERFACE_LINK_LIBRARIES "${_dpdk_libraries} ${dpdk_dependencies}"
```
tchaikov, line 187: i think the argument for `INTERFACE_LINK_LIBRARIES` should be a list instead of space-separated values. see https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES.html and in CMake, a list is semicolon-separated. see https://cmake.org/cmake/help/latest/command/list.html, so probably, we should put something like:
```cmake
INTERFACE_LINK_LIBRARIES "${_dpdk_libraries};${dpdk_dependencies}"
```
?

tchaikov:
@p12tic hi Povilas, do you plan to continue working on this pull request?

cmake/Finddpdk.cmake
p12tic: Thanks for noticing this, I wonder how it worked on my side. Fixed now.

cmake/Finddpdk.cmake
p12tic: Agreed, fixed now.

cmake/Finddpdk.cmake
p12tic: Fixed now.

p12tic:
@tchaikov Sorry for the delay in responding, I was away for a bit.

I have applied all your comments, the PR should be ready to go.

cmake/Finddpdk.cmake
p12tic: Thanks for mentioning this. I will check ways to improve this in a subsequent PR.

tchaikov (approved):
lgtm

tchaikov:
@scylladb/seastar-maint hello maintainers, could you help approve for the workflow and probably review this change as well? so we can test this change?

PR #2414, opened 2024-08-28
httpd: rm unused prom structs

tomershafir:


apps/httpd/main.cc
```diff
@@ -73,8 +73,6 @@ void set_routes(routes& r) {
 }
```
avikivity, line 74: > httpd: rm unused prom structs

prometheus

tchaikov: also please consider using "remove" instead of "rm" for better readability. For example:

> httpd: remove unused prometheus structs

while brevity is appreciated, full words often enhance clarity in commit messages without significantly impacting title length.

Issue #2403, opened 2024-08-24
Need IO record and replay facility


xemul:
When checking IO-related issues on prod. nodes we only have metrics and some Linux tooling at hand. Metrics show averaged over several minutes counters, while Linux tooling shows the _result_ of IO scheduler's work. To get better understanding of what's going on with seastar IO stack, we need to understand how the requests are submitted by uppoer code (read -- Scylla code) to seastar, not by seastar to Linux/disk.

One of the tools to help that can be IO tracer, that collects info about requests that are queued to IO scheduler. Very important parameter here is submission timestamp, because currently IO scheduler is built with the idea of uniform workloads in mind. Non-uniform input is handled, but the understanding of how exactly "non-uniform" it is is missing.

xemul:
> Isn't the input to the I/O scheduler dependent on the output?

Well, in math sense yes, scheduler defines `output = sched_fn(input)` so in theory we could deduce a reverse function for `input = ~sched_fn(output)`, but I don't have good ideas how to do it.

> For example, a sequential workload's submit timestamps will depend on when the requests are completed (even with read-ahead or write-behind).

No, to get submit timestamp from complete timestamp, we need execution time, but it's not cut in stone. Quite opposite, we measure execution from from submit and complete timestamps.

PR #2417, opened 2024-08-30
Binary event tracer

xemul:
This patch adds ability to capture info about frequent events (such as queued/dispatched/completed IO requests) into a buffer in memory and write it into a file. The goal is to collect a lot of events in relatively short period of time (e.g. -- seconds to minutes) thus getting what _exactly_ is happening in the system, instead of averaged statistical metrics. Events are collected in binary form, so a decoder is included.

In this PR the tracer is put on io-queue to collect IO events. Example of the decoded output:

```
00000000:000.000002 --- buffer (skipped 0) ---
00000009:- OPENING: shard=0 text=0:main,1:exit,2:lr  
00000021:000.000105 IO Q 25180 r class 2 512
0000002c:000.000106 IO Q 24d80 r class 2 512
00000037:000.000116 IO D 25180
0000003e:000.000137 IO D 24d80
00000045:000.000190 IO C 24d80
0000004c:000.000190 IO C 25180
00000053:000.000304 IO Q 25180 r class 2 512
0000005e:000.000305 IO Q 24d80 r class 2 512
00000069:000.000306 IO D 25180
00000070:000.000306 IO D 24d80
00000077:000.000361 IO C 25180
0000007e:000.000371 IO C 24d80
...
```

The OPENING message includes shard id and id:name mapping for registered sched classes.
The IO event format above is
`<offset in file>:<timestamp in usec> <event name Queue/Dispatch/Complete> <request id>`
and for Queued event also -- `<direction r/w> <class ID> <length>`

For IO there's a set of io_log.trace() per-request calls, but it's not very efficient. First it takes time to format the message and forward it to syslogd (in prod servers). Next, each request traces three messages:

```
TRACE 2024-08-30 14:02:59,640 [shard 2:lr  ] io - dev 64771 : req 0x60200011ec80 queue  len 512 capacity 80612
TRACE 2024-08-30 14:02:59,640 [shard 2:main] io - dev 64771 : req 0x60200011ec80 submit
TRACE 2024-08-30 14:02:59,640 [shard 2:main] io - dev 64771 : req 0x60200011ec80 complete
```
that's 290 bytes. Contrary to this, this tracer generates 25 bytes per request. Also note, that timestamp precision is msec, while IO events need at least usec resolution.

refs: #2403


avikivity:
We should use statically defined tracepoints (\<sys/sdt.h>), like blktrace/blkparse. The challenge is getting the information out in a performant way. I tried systemtap dyninst and failed, several times. Please check if ltt-ng works.

/cc @michoecho

michoecho:
> We should use statically defined tracepoints (<sys/sdt.h>) instead of an ad-hoc format/protocol/parser, like blktrace/blkparse. The challenge is getting the information out in a performant way. I tried systemtap dyninst and failed, several times. Please check if ltt-ng works.
> 
> /cc @michoecho

@xemul In other words: Avi has banned custom tracing patches until it is proven that existing tracing frameworks aren't good enough. (I had a custom tracing patch denied too, once.)

If you want to experiment with lttng, you can take https://github.com/scylladb/scylla-enterprise-seastar/commit/f0b870368f733bde6fd54cfbec6052080b5f2db5#diff-ceb4cd71fc94af90f390646cde7f7cc8bc673cbd7fd77208609b42fc634d0fd1 as a starting point. (But to make it compile, you will probably have to prepend `LLTNG_UST_`(/`lltng_ust_`) to all lttng identifiers —  there was an API change in lltng between v2.12 and v2.13).

xemul:
> @xemul In other words: Avi has banned custom tracing patches until it is proven that existing tracing frameworks aren't good enough. 

Yeah, Avi made it crystal clear :D

> (I had a custom tracing patch denied too, once.)

There must be denied-binary-tracing club somewhere, optionally we could found one

> If you want to experiment with lttng, you can take https://github.com/scylladb/scylla-enterprise-seastar/commit/f0b870368f733bde6fd54cfbec6052080b5f2db5#diff-ceb4cd71fc94af90f390646cde7f7cc8bc673cbd7fd77208609b42fc634d0fd1 as a starting point.

Thanks!

xemul:
Closing in favor of #3449

PR #2419, opened 2024-09-04
prometheus: add server setup documentation

tomershafir:
Add missing Prometheus server setup documentation to `doc/prometheus.md`.

### DIFF
--- a/doc/prometheus.md
+++ b/doc/prometheus.md
@@ -15,6 +15,39 @@ Seastar would reply based on the content type header, so pointing your browser t
 Starting from Prometheus 2.0, the binary protocol is no longer supported.
 While seastar still supports the binary protocol, it would be deprecated in a future release.
 
+## Server Setup
+
+The Seastar application needs to manually start the Prometheus server, using Seastar API provided in [seastar/core/prometheus.hh](../include/seastar/core/prometheus.hh).
+
+### Examples
+
+```C++
+#include <seastar/core/prometheus.hh>
+#include <seastar/http/httpd.hh>
+#include <seastar/net/inet_address.hh>
+
+int main(int argc, char** argv) {
+    seastar::app_template app;
+    return app.run(argc, argv, [] {
+      httpd::http_server_control prometheus_server;
+      
+      prometheus::config pctx;
+      pctx.metric_help = "seastar app statistics";
+      pctx.prefix = "app";
+
+      prometheus_server.start("prometheus").get();
+      prometheus::start(prometheus_server, pctx).get();
+
+      net::inet_address addr("127.0.0.1");
+      uint16_t port = 9180;
+      prometheus_server.listen(socket_address{addr, port}).get();
+    });
+}
+```
+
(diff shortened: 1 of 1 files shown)


amnonh (approved):

amnonh:
LGTM

doc/prometheus.md
```diff
@@ -15,6 +15,42 @@ Seastar would reply based on the content type header, so pointing your browser t
...
+
+      prometheus_server.start("prometheus").get();
+      prometheus::start(prometheus_server, pctx).get();
+
+      net::inet_address addr("127.0.0.1");
+      uint16_t port = 9180;
+      prometheus_server.listen(socket_address{addr, port}).handle_exception([addr, port] (auto e) {
+        std::cerr << seastar::format("Could not start Prometheus API server on {}:{}: {}\n", addr, port, e);
+        return make_exception_future<>(e);
+      }).get();
```
tchaikov, line 46: i'd suggest drop the error handling. because the point of the code list is for explaining that
```c++
prometheus::start(prometheus_server, pctx)
```
the other part is for providing the context. not for showing how to write a proper seastar application. BTW, i'd suggest use `fmt::print()` directly here, instead of using `seastar::format()`.

tomershafir: Removed error handling. 

If you prefer doxygen:

1. I think it should be at the header level because it aims to present complete examples.
2. I think the entire document should be migrated.

But thats not so easy Q probably. Do you have feedback from users on docs? Do they like `doc/*.md` files? I think doxygen can be less discoverable in this case. Maybe we can move content and have `prometheus.md` contain only a link to doxygen, or have a pure link.

tchaikov: > Removed error handling.
> 
> If you prefer doxygen:
> 
>     1. I think it should be at the header level because it aims to present complete examples.

i disagree.  i think we should just highlight that what `start()` is for. the "complete" example is necessary if the document is not enough, and a code snippet is not good enough. but i think the document would suffice even without the code snippet.
> 
>     2. I think the entire document should be migrated.
> 

the document is served as a tutorial instead of a reference. they serve for different purposes

> 
> But thats not so easy Q probably. Do you have feedback from users on docs? Do they like `doc/*.md` files? 

i don't have feedback from users part from myself. as i am also a user of seastar. if i want to use an API, i read its header. if i am still confused, i read the document, demos and even tests.

> I think doxygen can be less discoverable in this case. 

doxygen as a rendered document is not that discoverable as header is . but since we use doxygen to format the header document, probably we can continue using it. the content matters more than the format, IMHO.

> Maybe we can move content and have `prometheus.md` contain only a link to doxygen, or have a pure link.

that's a different story. again, i think, they serve for different purposes.

tomershafir: Being a tutorial, I think a new user would read `doc/prometheus.md` and would not easily understand what header he should use and how to setup a prometheus server without this commit.

tomershafir:
@xemul Can we merge this?

PR #2420, opened 2024-09-04
stall-detector: Try hard not to crash while collecting backtrace

xemul:
Sometimes stall-detector signal comes in the middle of exception handling. If the stall is detected, stack unwiding starts to collect the stalled backtrace. Since exception handling means unwiding the stack as well, those two unwinders need to cooperate carefully, which is not guaranteed (spoiler: they don't cooperate carefully). In unlucky case, segmentation fault happens, the app is killed with SEGV.

This patch helps stall detector to bail out in case of SEGV arrival while collecting the backtrace with minimally possible yet detailed enough stall report.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -34,6 +34,7 @@ module;
 #include <regex>
 #include <thread>
 #include <unordered_set>
+#include <setjmp.h>
 
 #include <grp.h>
 #include <spawn.h>
@@ -836,7 +837,23 @@ public:
     }
 };
 
+static thread_local jmp_buf stall_detector_env;
+static thread_local bool in_stall_detector = false;
+static thread_local bool crash_collecting_backtrace = false;
+
+inline void maybe_crash_for_test() noexcept {
+    if (crash_collecting_backtrace) [[unlikely]] {
+        *(volatile int *)nullptr = 0;
+    }
+}
+
 static void print_with_backtrace(backtrace_buffer& buf, bool oneline) noexcept {
+    if (sigsetjmp(stall_detector_env, 0)) {
+        buf.append(" ¯\\_(ツ)_/¯\n");
+        goto out;
+    }
+    in_stall_detector = true;
+
     if (local_engine) {
         buf.append(" on shard ");
         buf.append_decimal(this_shard_id());
@@ -847,12 +864,18 @@ static void print_with_backtrace(backtrace_buffer& buf, bool oneline) noexcept {
 
   if (!oneline) {
     buf.append(".\nBacktrace:\n");
+    maybe_crash_for_test();
     buf.append_backtrace();
   } else {
     buf.append(". Backtrace:");
+    maybe_crash_for_test();
     buf.append_backtrace_oneline();
     buf.append("\n");
   }
+
+out:
+    in_stall_detector = false;
+    buf.flush();
 }
 
 // Print the current backtrace to stdout with the given cause.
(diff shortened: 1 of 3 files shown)


src/core/reactor.cc
```diff
@@ -842,7 +842,23 @@ class backtrace_buffer {
...
+
+inline void maybe_crash_for_test() noexcept {
+    if (crash_collecting_backtrace) [[unlikely]] {
+        *(volatile int *)nullptr = 0;
+    }
+}
+
 static void print_with_backtrace(backtrace_buffer& buf, bool oneline) noexcept {
+    if (sigsetjmp(stall_detector_env, 0)) {
+        buf.append(" ¯\\_(ツ)_/¯\n");
```
fee-mendes, line 857: +1

michoecho:
Doesn't solve the problem entirely, since SIGSEGV isn't the only possible symptom (you could get an infinite loop for example, why not), but I guess it prevents a crash in the cases it's enough (which is probably a great majority of cases), and doesn't hurt in the others, so why not.

src/core/reactor.cc
```diff
@@ -842,7 +843,23 @@ class backtrace_buffer {
...
+        goto out;
+    }
+    in_stall_detector = true;
+
```
avikivity, line 862: To be technically correct, we need an std::atomic_signal_fence(std::memory_order_relaxed). This prevents a magical compiler from delaying the write to memory because no one reads it.

tests/unit/stall_detector_test.cc
```diff
@@ -175,6 +175,11 @@ SEASTAR_THREAD_TEST_CASE(spin_in_kernel) {
     test_spin_with_body("kernel", [] { mmap_populate(128 * 1024); });
 }
 
+SEASTAR_THREAD_TEST_CASE(crash_collecting_backtrace) {
+    reactor::test::set_stall_detector_crash_collecting_backtrace();
+    engine().update_blocked_reactor_notify_ms(100ms);
+    spin(500ms);
+}
```
avikivity, line 182: Did you also reproduce the crash during unwinding? It's not given that siglongjmp is a safe way to unwind. If the unwinder takes a lock, it will leak it (though I'm guessing it doesn't).

xemul: > Did you also reproduce the crash during unwinding?

In labs -- unfortunately, no :( 

> It's not given that siglongjmp is a safe way to unwind.

Yes, sure, at this point the situation is already screwed up, and it's questionable whether these tricks are making things even worse or not

avikivity: Perhaps we can override __cxa_throw and whatever function it uses to exist unwinding (but maybe there isn't one), and call them via RTLD_NEXT. Then we can set flags when unwinding is in progress, and just avoid going into the stall detector again (or perhaps: ask the stall detector to run on the exit path of __cxa_throw).

avikivity: I don't think it will work.

Also, tracing exception throwers is important.

michoecho: > Perhaps we can override __cxa_throw and whatever function it uses to exist unwinding (but maybe there isn't one)

There isn't one.

avikivity: Maybe have a blacklist of functions that are known to crash. Every time we see a crash, add the triggering function to the blacklist. In a few short years we'll have a robust filter.

travisdowns:
It's worth noting there is a reproducer now, for at least one type of crash, see https://github.com/scylladb/seastar/issues/2697 and https://github.com/scylladb/seastar/pull/2714. So perhaps it is worth revisiting this PR as part of the reason it has stalled seemed to be the lack of repro?

xemul:
upd: rebased to check #2714

xemul:
closing in favor of #2714

PR #2421, opened 2024-09-06
app-template: add ability to specify options_description to be used for --help

avikivity:
Hiding options from --help output is useful when deprecating options.

When using boost::program_options, the way to hide options from display is to provide an alternative options_description object and print that. So this patch adds a way to provide a help-specific options_description.

While this entangles us even more with boost::program_options, I don't see a simple alternative.

### DIFF
--- a/include/seastar/core/app-template.hh
+++ b/include/seastar/core/app-template.hh
@@ -129,6 +129,7 @@ private:
     std::shared_ptr<smp> _smp;
     seastar_options _opts;
     boost::program_options::options_description _app_opts;
+    boost::program_options::options_description* _app_visible_opts_for_help = nullptr;
     boost::program_options::options_description _seastar_opts;
     boost::program_options::options_description _opts_conf_file;
     boost::program_options::positional_options_description _pos_opts;
@@ -136,6 +137,7 @@ private:
     configuration_reader _conf_reader;
 
     configuration_reader get_default_configuration_reader();
+    void add_extra_options(boost::program_options::options_description& opts);
 public:
     struct positional_option {
         const char* name;
@@ -153,6 +155,7 @@ public:
     boost::program_options::options_description& get_options_description();
     boost::program_options::options_description& get_conf_file_options_description();
     boost::program_options::options_description_easy_init add_options();
+    void set_app_visible_options_for_help(boost::program_options::options_description* opts);
     void add_positional_options(std::initializer_list<positional_option> options);
     boost::program_options::variables_map& configuration();
     int run_deprecated(int ac, char ** av, std::function<void ()>&& func) noexcept;
--- a/src/core/app-template.cc
+++ b/src/core/app-template.cc
@@ -85,16 +85,7 @@ app_template::app_template(app_template::seastar_options opts)
         if (!alien::internal::default_instance) {
             alien::internal::default_instance = _alien.get();
         }
-        _app_opts.add_options()
-                ("help,h", "show help message")
-                ;
-        _app_opts.add_options()
-                ("help-seastar", "show help message about seastar options")
-                ;
-        _app_opts.add_options()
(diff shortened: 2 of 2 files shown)


avikivity:
User: https://github.com/scylladb/scylladb/pull/20480

include/seastar/core/app-template.hh
```diff
@@ -129,13 +129,15 @@ private:
     std::shared_ptr<smp> _smp;
     seastar_options _opts;
     boost::program_options::options_description _app_opts;
+    boost::program_options::options_description* _app_visible_opts_for_help = nullptr;
```
tchaikov, line 132: i'd suggest use `&_app_opts` as the default value of `_app_visible_opts_for_help`, like:
```c++
boost::program_options::options_description* _app_visible_opts_for_help = &_app_opts;
```

this would help with the readability -- it's clear that we use `_app_opts` by default, when handling `--help`.

src/core/app-template.cc
```diff
@@ -109,12 +100,33 @@ app_template::app_template(app_template::config cfg)
...
     return _opts;
 }
 
+void
+app_template::set_app_visible_options_for_help(boost::program_options::options_description* opts) {
+    _app_visible_opts_for_help = opts;
+    if (_app_visible_opts_for_help) {
+        add_extra_options(*_app_visible_opts_for_help);
+    }
+}
```
tchaikov, line 128: could be more explicit, like:
```c++
if (opts) {
    add_extra_options(*opts);
    _app_visible_opts_for_help = opts;
}
```

this way, it's clear that we mutate the pointee of `opts`, and it's a noop, if `opts` is  `nullptr`. but i'd suggest accept a reference instead. like
```c++
set_app_visible_options_for_help(boost::program_options::options_description& opts)
```
as, why would the caller want to pass a nullptr for a noop?

include/seastar/core/app-template.hh
```diff
@@ -153,6 +155,7 @@ public:
     boost::program_options::options_description& get_options_description();
     boost::program_options::options_description& get_conf_file_options_description();
     boost::program_options::options_description_easy_init add_options();
+    void set_app_visible_options_for_help(boost::program_options::options_description* opts);
```
tchaikov, line 158: if we really want to accept `nullptr` as a valid input, probably should add a doxygen comment to explain what `nullptr` is for. and, mutating the value passed by caller and expect that caller to keep the referenced object alive might also warrant a comment.

still i think a reference is a better fit in this case.

src/core/app-template.cc
```diff
@@ -211,7 +223,8 @@ app_template::run_deprecated(int ac, char ** av, std::function<void ()>&& func)
         if (!_opts.description.empty()) {
             std::cout << _opts.description << "\n";
         }
-        std::cout << _app_opts << "\n";
+        auto& which_opts = _app_visible_opts_for_help ? *_app_visible_opts_for_help : _app_opts;
```
tchaikov, line 226: if we use `&_app_opts` as the default value of `_app_visible_opts_for_help`. we could simplify this part like:
```c++
        assert(_app_visible_opts_for_help);
        std::cout << *_app_visible_opts_for_help << "\n";
```

tchaikov (commented):
left some comments.

src/core/app-template.cc
```diff
@@ -109,12 +100,33 @@ app_template::app_template(app_template::config cfg)
 {
 }
 
+void
+app_template::add_extra_options(boost::program_options::options_description& opts) {
```
tchaikov, line 104: instead of mutating the parameter, i'd suggest returning a `boost::program_options::options_description`. more readable this way, IMHO. BTW, could make this function `static`, if we don't want to mark it with the `const` specifier.

PR #2422, opened 2024-09-09
rpc: fix compilation error caused by fmt::runtime()

pwrobelse:
Depending on the 'SEASTAR_LOGGER_COMPILE_TIME_FMT' the declaration of 'logger::log()'
function may use either 'fmt::format_string' or 'const char*' format parameter.

In the case of the usage of 'const char*', there is a need to call 'fmt::runtime()' function to
obtain format_string.

When the project was built on Ubuntu 22.04 using the steps described in the tutorial,
there was a compilation error related to passing argument with type 'fmt::format_string<>'
to 'fmt::runtime()' that handles characters strings.

This change introduces contexpr if statement to remove the compilation error and use
'fmt::runtime()' only when the function takes 'const char*'.

### DIFF
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include <type_traits>
 #include <unordered_map>
 #include <unordered_set>
 #include <list>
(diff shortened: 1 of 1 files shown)


pwrobelse:
Perhaps, the problem started to be seen after merging: https://github.com/scylladb/seastar/commit/e3249a8ede9a16127cd2e6995b02602649ea305f

tchaikov:
probably a better fix is to conditionalize this part with `SEASTAR_LOGGER_COMPILE_TIME_FMT` as well.

tchaikov:
> probably a better fix is to conditionalize this part with `SEASTAR_LOGGER_COMPILE_TIME_FMT` as well.

created #2425

PR #2423, opened 2024-09-10
http: clarify the ownership in the router's doxygen comment

kreuzerkrieg:
Update the `routes`'s  `put`, `add`, `add_default_handler`, `add_cookie` and `del_cookie` doxygen documentation

### DIFF
--- a/include/seastar/http/routes.hh
+++ b/include/seastar/http/routes.hh
@@ -74,7 +74,7 @@ struct path_description;
 /**
  * routes object do the request dispatching according to the url.
  * It uses two decision mechanism exact match, if a url matches exactly
- * (an optional leading slash is permitted) it is choosen
+ * (an optional leading slash is permitted) it is chosen
  * If not, the matching rules are used.
  * matching rules are evaluated by their insertion order
  */
@@ -89,7 +89,11 @@ public:
      * adding a handler as an exact match
      * @param url the url to match (note that url should start with /)
      * @param handler the desire handler
-     * @return it self
+     * @return itself
+     * @attention If successful, this method takes ownership of the handler
+        pointer. It will be automatically deleted when the routes instance is
+        destroyed. However, if the method throws, the caller is responsible
+        for deleting the handler pointer.
      */
     routes& put(operation_type type, const sstring& url, handler_base* handler);
 
@@ -107,7 +111,9 @@ public:
      * First in higher priority
      * @param rule a rule to add
      * @param type the operation type
-     * @return it self
+     * @return itself
+     * @attention This method takes ownership of the match_rule pointer. It will be automatically deleted when the
+     * routes instance is destroyed.
      */
     routes& add(match_rule* rule, operation_type type = GET) {
         _rules[type][_rover++] = rule;
@@ -121,6 +127,8 @@ public:
      * @param url
      * @param handler
      * @return
+     * @attention This method takes ownership of the handler's pointer.It will be automatically deleted when the routes
+     * instance is destroyed.
      */
     routes& add(operation_type type, const url& url, handler_base* handler);
 
@@ -129,6 +137,7 @@ public:
      * Example  routes.add_default_handler(handler);
      * @param handler
      * @return
+     * @attention The caller must keep the @c handler active as long as the route instance is active.
      */
     routes& add_default_handler(handler_base* handler);
 
@@ -137,7 +146,7 @@ public:
      * the general handler calls this method with the request
      * the method takes the headers from the request and find the
      * right handler.
-     * It then call the handler with the parameters (if they exists) found in the url
+     * It then calls the handler with the parameters (if they exist) found in the url
      * @param path the url path found
      * @param req the http request
      * @param rep the http reply
@@ -169,7 +178,6 @@ private:
      * Normalize the url to remove the last / if exists
      * and get the parameter part
      * @param url the full url path
-     * @param param_part will hold the string with the parameters
      * @return the url from the request without the last /
      */
     sstring normalize_url(const sstring& url);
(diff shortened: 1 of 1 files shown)


src/http/routes.cc
```diff
@@ -53,7 +53,7 @@ routes::~routes() {
             delete r.second;
         }
     }
-
+    delete _default_handler;
```
tchaikov, line 56: but we don't claim that the ownership of the handler is transformed from the caller of `add_default_handler()` to the `routes` instance. instead of deleting this pointer, i'd suggest revise the doxygen comment of `add_default_handler()` to explain that the callee does not take the ownership of this pointer, the caller is responsible to keep it alive.

this change would break the existing seastar applications. for instance, see https://github.com/redpanda-data/redpanda/blob/ecf2e99c97f5fb60d8971ce4e4760724c2b3cf02/src/v/cloud_storage/tests/s3_imposter.cc#L509

tchaikov (changes requested):
i'd suggest improve the document of this member function instead.

src/http/routes.cc
```diff
@@ -53,7 +53,7 @@ routes::~routes() {
             delete r.second;
         }
```
avikivity, line 54: > Prevent leaking _default_handler

This repository requires patches that have an explanation, and requires the subject line to point to the subsystem that is modified.

src/http/routes.cc
kreuzerkrieg: reverted

src/http/routes.cc
kreuzerkrieg: Done, hope now it is good to go

include/seastar/http/routes.hh
```diff
@@ -107,7 +110,9 @@ public:
      * First in higher priority
      * @param rule a rule to add
      * @param type the operation type
-     * @return it self
+     * @return itself
```
tchaikov, line 113: if we want to s/it self/itself/, probably should be consistent. and i agree that "itself" is better.

include/seastar/http/routes.hh
```diff
@@ -90,6 +90,9 @@ public:
      * @param url the url to match (note that url should start with /)
      * @param handler the desire handler
      * @return it self
+     * @attention If successful, this method takes ownership of the handler's
+     * pointer. It will be deleted upon `routes` instance destruction. If the
+     * method throws, it is up to caller to take care of the handler pointer.
```
tchaikov, line 95:
```suggestion
     * @attention If successful, this method takes ownership of the handler
        pointer. It will be automatically deleted when the routes instance is
        destroyed. However, if the method throws, the caller is responsible
        for deleting the handler pointer.
```

more specific this way, and easier to digest.

include/seastar/http/routes.hh
```diff
@@ -237,6 +248,7 @@ public:
      * @param cookie a cookie returned previously by add_cookie
      * @param type the operation type
      * @return the pointer to the rule
+     * @attention The ownership of the pointer returns to the caller
```
tchaikov, line 251:
```suggestion
     * @attention Ownership of the pointer is returned to the caller.
```

tchaikov (commented):
left a couple comments.

tchaikov:
the title of the commit message reads:

> Update the routes's put, add, add_default_handler, `add_cookie` and `del_cookie` doxygen documentation.

update ... documentation, is not accurate enough, i'd suggest put:

> http: clarify the ownership in the router's doxygen comment

- add the prefix of the area we are changing
- be more specific on the change

include/seastar/http/routes.hh
kreuzerkrieg: done

include/seastar/http/routes.hh
```diff
@@ -129,6 +137,8 @@ public:
      * Example  routes.add_default_handler(handler);
      * @param handler
      * @return
+     * @attention This method does not take ownership of the handler's pointer. It is
+     * the callee responsibility to delete it
```
tchaikov, line 141: > It is the callee responsibility to delete it

should replace "callee" with "caller". but i think this line is redundant. we are not obliged to encourage to programmer to write leak-free code. but we should be very explicit on the ownership and the expectation of the life cycle, if the ownership is not transferred to us.

so i'd suggest replacing this line with

> The caller must keep the @c handler active as long as the route instance is active.

kreuzerkrieg: done

tchaikov (commented):
sorry, Ernest. i forgot to send the pending review last night. this should be the last batch of them for the latest revision.

include/seastar/http/routes.hh
```diff
@@ -225,6 +234,8 @@ public:
      * @param rule a rule to add
      * @param type the operation type
      * @return a cookie using which the rule can be removed
+     * @attention This method takes ownership of the match_rule pointer. It will
```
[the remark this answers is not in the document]
kreuzerkrieg: Done

include/seastar/http/routes.hh
```diff
@@ -121,6 +127,8 @@ public:
      * @param url
      * @param handler
      * @return
+     * @attention This method takes ownership of the handler's pointer. It will
+     * be deleted upon `routes` instance destruction
```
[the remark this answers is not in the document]
kreuzerkrieg: Done

include/seastar/http/routes.hh
```diff
@@ -107,7 +111,9 @@ public:
      * First in higher priority
      * @param rule a rule to add
      * @param type the operation type
-     * @return it self
+     * @return itself
+     * @attention This method takes ownership of the match_rule pointer. It will
```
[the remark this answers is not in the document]
kreuzerkrieg: Done

tchaikov (approved):
lgtm

tchaikov:

PR #2425, opened 2024-09-11
rpc: conditionally use fmt::runtime() based on SEASTAR_LOGGER_COMPILE_TIME_FMT

tchaikov:
in 8c58a723, we fixed the build failure when passing `fmt::format_string<>` to `fmt::runtime()`. but we were checking the fmt string's type to see if it is a `const char*`. this works fine. as there are only two cases,

1. `SEASTAR_LOGGER_COMPILE_TIME_FMT` is defined, and a templated parameter of `fmt::format_string<...>` is passed in.
2. `SEASTAR_LOGGER_COMPILE_TIME_FMT` is not defined, and a `const char*` pointer is passed in.

it would be better if we can check the same macro and dispatch based on it to pass `fmt` or `fmt::runtime(fmt)` to `fmt::format_to()`, for

- better readability. as we are more consistent this way.
- probably faster compilation. the preprocessor pass could be faster than the compiler which actually evaluate the `std::is_same_v` type trait.

so, in this change, let's switch to the plain macro based dispatch.

### DIFF
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -21,7 +21,6 @@
 
 #pragma once
 
-#include <type_traits>
 #include <unordered_map>
 #include <unordered_set>
 #include <list>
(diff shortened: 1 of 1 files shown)


tchaikov:
@pwrobelse hi Patryk, could you help review this change?

pwrobelse (approved):
Looks good to me :+1:

tchaikov:

PR #2427, opened 2024-09-12
build: enable Seastar to build shared and static libs in a single build

tchaikov:
before this change, we respect the CMake variable named `BUILD_SHARED_LIBS`, and build shared libraries if it is set, build static libraries otherwise. but this model cannot fulfill the needs of a parent project which needs to build both static and shared seastar libraries with different configuration in a multi-config generator settings. as `BUILD_SHARED_LIBS` is a CMake variable which cannot be changed at build time, and hence cannot be assigned with
different values using generator expression. 

so, in this change, we add two options

- Seastar_BUILD_STATIC_LIBS
- Seastar_BUILD_SHARED_LIBS

to configure if the build should generated static libraries and shared libraries respectively. but `BUILD_SHARED_LIBS` is still supported, and the behavior is backward compatible. the only user-visible differences are, the libraries of

- seastar
- seastar_testing
- seastar_perf_testing

are now aliases of the targets which builds the static or shared libraries. so one cannot build "seastar" as a target anymore, but should specify which library to build:

```
$ cmake --build build --target seastar # does not work anymore
$ cmake --build build --target seastar-static
$ cmake --build build --target seastar-shared
```

the reason is the limit of an ALIAS library, which cannot be used as a target. but we still need to use a non-interface library to generate .pc file, where, for instance, we use `$<TARGET_FILE_NAME:seastar_testing>` for the library list of a
certain Seastar library.

but a CMake-based project including seastar can still link against it using "seastar" as a library.

please note, in d8a70b39, we pass `--ftls-mode=initial-exec` when building the tree, to address the problem of recursive call of malloc when using seastar allocator. but we cannot use different compiling options when building both shared libraries and static libraries, so we always pass this option as long as `BUILD_SHARED_LIBS` OR `Seastar_BUILD_SHARED_LIBS` is set *and* when we are building dev/release modes. as these two modes enables the seastar allocator. this should not hurt the overall performance, as we are not using global-dynamic mode, but initial-exec. this is verified using
```console
$ cmake \
  -DCMAKE_CONFIGURATION_TYPES="Debug;RelWithDebInfo" \
  -DCMAKE_CROSS_CONFIGS="Debug;RelWithDebInfo" \
  -DCMAKE_DEFAULT_CONFIGS="Debug;RelWithDebInfo" \
  -DSeastar_BUILD_SHARED_LIBS=ON \
  -DSeastar_BUILD_STATIC_LIBS=ON  \
  -G "Ninja Multi-Config" -B build -S .
$ cmake --build build --config RelWithDebInfo --target seastar-static
$ readelf -r
build/CMakeFiles/seastar-objects.dir/RelWithDebInfo/src/core/reactor.cc.o| \
  grep -i tls | tr -s ' ' | cut -d' ' -f 3,5 | uniq
R_X86_64_TPOFF32 __tls_guard
R_X86_64_PLT32 _ZStlsRSoRKNSt15_[...]
```

and

```console
$ cmake -DCMAKE_BUILD_TYPE="RelWithDebInfo" -G "Ninja" -B build -S .
$ cmake --build build --target seastar-static
$ readelf -r build/CMakeFiles/seastar-objects.dir/src/core/reactor.cc.o| \
  grep -i tls | tr -s ' ' | cut -d' ' -f 3,5 | uniq
R_X86_64_TPOFF32 __tls_guard
R_X86_64_PLT32 _ZStlsRSoRKNSt15_[...]
```

so the TLS variables are using "TPOFF32" relocation type, in general, it implies the initial-exec TLS model. the PLT32 is not specific to TLS, it's more related to external symbols linkages.

Refs scylladb/scylladb#2717

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,6 +320,61 @@ set (Seastar_TEST_TIMEOUT
 option (BUILD_SHARED_LIBS
   "Build seastar library as shared libraries instead of static"
   OFF)
+cmake_dependent_option (Seastar_BUILD_SHARED_LIBS
+  "Build Seastar as shared libraries" OFF
+  "NOT DEFINED BUILD_SHARED_LIB" OFF)
+cmake_dependent_option (Seastar_BUILD_STATIC_LIBS
+  "Build Seastar as static libraries" ON
+  "NOT DEFINED BUILD_SHARED_LIB" OFF)
+if (NOT (Seastar_BUILD_SHARED_LIBS OR Seastar_BUILD_STATIC_LIBS))
+  message (FATAL_ERROR "Seastar_BUILD_STATIC_LIBS and Seastar_BUILD_STATIC_LIBS cannot both be disabled")
+endif ()
+
+function (seastar_add_library name objects)
+  if (Seastar_BUILD_STATIC_LIBS)
+    list (APPEND types static)
+  endif ()
+  if (Seastar_BUILD_SHARED_LIBS)
+    list (APPEND types shared)
+  endif ()
+
+  foreach (type ${types})
+    string (TOUPPER ${type} TYPE)
+    set (fullname ${name}-${type})
+    add_library (${fullname} ${TYPE} $<TARGET_OBJECTS:${objects}>)
+    list (APPEND seastar-libraries ${fullname})
+
+    set_target_properties (${fullname} PROPERTIES
+      OUTPUT_NAME "${name}")
+    # add_library() does not populate the property of the included sources
+    # to the library target being added, so we need to do this manually.
+    foreach (property
+        INTERFACE_COMPILE_DEFINITIONS
+        INTERFACE_COMPILE_FEATURES
+        INTERFACE_COMPILE_OPTIONS
+        INTERFACE_INCLUDE_DIRECTORIES
+        INTERFACE_LINK_LIBRARIES
+        LINK_LIBRARIES)
+      get_target_property (value ${objects} ${property})
+      if (value)
+        set_target_properties (${fullname}
+          PROPERTIES ${property} "${value}")
+      endif ()
+    endforeach ()
+
+    add_library (Seastar::${fullname} ALIAS ${fullname})
+
+    # make sure ${name} and Seastar:${name} always exist, and we prefer
+    # the libraries if they are available.
+    if (NOT TARGET ${name})
+      add_library (${name} ALIAS ${fullname})
+      add_library (Seastar::${name} ALIAS ${fullname})
+    endif ()
+  endforeach (type ${types})
+
+  set (seastar-libraries ${seastar-libraries} PARENT_SCOPE)
+endfunction ()
+
 # We set the following environment variables
 # * ASAN_OPTIONS=disable_coredump=0:abort_on_error=1:detect_stack_use_after_return=1:verify_asan_link_order=0
 #   By default ASan disables core dumps because they used to be
@@ -505,7 +560,7 @@ seastar_generate_protobuf (
   IN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/src/proto/metrics2.proto
   OUT_DIR ${Seastar_GEN_BINARY_DIR}/src/proto)
 
-add_library (seastar
+add_library (seastar-objects OBJECT
   ${http_chunk_parsers_file}
   ${http_request_parser_file}
   ${proto_metrics2_files}
@@ -784,15 +839,13 @@ add_library (seastar
 set_source_files_properties(src/core/thread.cc
   PROPERTIES COMPILE_FLAGS -U_FORTIFY_SOURCE)
 
-add_library (Seastar::seastar ALIAS seastar)
-
-add_dependencies (seastar
+add_dependencies (seastar-objects
   seastar_http_chunk_parsers
   seastar_http_request_parser
   seastar_http_response_parser
   seastar_proto_metrics2)
 
-target_include_directories (seastar
+target_include_directories (seastar-objects
   PUBLIC
     $<INSTALL_INTERFACE:include>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
@@ -824,12 +877,12 @@ endif ()
 if (CMAKE_CXX_STANDARD GREATER_EQUAL 23)
   include (CheckP2582R1)
   if (Cxx_Compiler_IMPLEMENTS_P2581R1)
-    target_compile_definitions (seastar
+    target_compile_definitions (seastar-objects
       PUBLIC SEASTAR_P2581R1)
   endif ()
 endif ()
 
-if (BUILD_SHARED_LIBS)
+if (BUILD_SHARED_LIBS OR Seastar_BUILD_SHARED_LIBS)
   # use initial-exec TLS, as it puts the TLS variables in the static TLS space
   # instead of allocating them using malloc. otherwise intercepting mallocs and
   # friends could lead to recursive call of malloc functions when a dlopen'ed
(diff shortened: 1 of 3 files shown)


tchaikov:
@denesb could you please help review this change? it implements the proposal at https://github.com/scylladb/scylladb/issues/2717#issuecomment-2323930591

denesb (approved):
Looks good, I don't understand the Cmake changes in detail, but the concept as explained in the cover letter as well as the code in high level looks good.

tchaikov:
thank your Botond, for your generous review.

tchaikov:
v2:

- rebased against master HEAD.

tchaikov:

tchaikov:

avikivity:
If I understand correctly, we build the object files just once.

Doesn't this degrade the performance of thread-local variables? It moves tls-model from initial-exec (fast) to global-dynamic. Or maybe global-dynamic. Need to examine the generated code to be sure.

tchaikov:
will do and report my findings here later on.

tchaikov:
v2:

- add the verification steps and report on the TLS mode in the commit message

avikivity:
Ok, this was due to my confusion. In d8a70b3906f5d70cf1367feba7099e9bbace18af we switched shared libraries to initial-exec. I remembered we downgraded it from global-dynamic, but thought it wasn't to initial-exec.

avikivity:
But looking at that commit, I see we have another different: -fvisibility=hidden. What happens now? Do shared libraries gain -fvisibility=hidden? Or do static libraries lose it?

tchaikov:
> But looking at that commit, I see we have another different: -fvisibility=hidden. What happens now?

the commit of f34335e5fac747a69800ecc67aa50897ce6f0f36 includes `-fvisibility=hidden` in compile flags only when building static libraries. i am quoting part of the commit message of that commit here:

```
    * do not pass `-fvisibility=hidden` to as a CXXFLAGS, if
      "BUILD_SHARED_LIBS" is TRUE. so all symbols are exported to
      application linked against libseastar.so. we don't have a set
      of mechinary or policy to expose/version symbols exposed via a
      shared library, so all of them are exposed. in the light of
      C++20 modular support, probably we don't need to introduce this
      mechinary in future.
```
so, 

- when building shared libraries we do not include `-fvisibility=hidden`
- when building static libraries we do include `-fvisibility=hidden`

> Do shared libraries gain -fvisibility=hidden? Or do static libraries lose it?

ahh. i see your point now. the latter, the static libraries lose it after this pull request, as long as we build the shared libraries *and* the static libraries in a single pass. do you think it's important? IMHO, it's not important at this moment. because:

- the symbol visibility matters only if we link to a DSO, unless we hide symbols as local ones using `static` or put them into an anonymous namespace
- in the build modes where we care about performance, like "release" mode, we build seastar libraries as static library

but i agree that a smaller list of elf global symbols is desirable. for instance, it helps to reduce the size of DSO, shorten the load time of the DSO, etc. but because the commit message puts:

```
      we don't have a set
      of mechinary or policy to expose/version symbols exposed via a
      shared library, so all of them are exposed.
```

we cannot differentiate the public symbols from the internal ones, without adding the directives like `__attribute__((__visibility__("default")))` and `__attribute__((__visibility__("hidden")))`. so, adding this compiler option do not really help.

IMHO, the recommended approach, although potentially tedious, is to:

- explicitly mark public interfaces with `__attribute__((__visibility__("default")))`.
- and then, apply the `-fvisibility=hidden` compiler flag when building both shared and static libraries.

but this is not in the scope of this change. and probably has lower priority as it's more about the DSO and is not trivial. what do you think?

tchaikov:
BTW, i created #2457 to track the enhancement for adding `[[gnu::visibility("default")]]` to public symbols.

tchaikov:
`Seastar::seastar` and friends are not exposed by the generated CMake config files. after this change, what we have is `Seastar::seastar-static`,  `Seastar::seastar_testing-static` and `Seastar::seastar_perf_testing-static`. 

need to manually create a `SeastarConfig.cmake`, and install it.

tchaikov:
v3:

-  add alias libraries in `cmake/SeastarConfig.cmake.in`.

avikivity:
I'm not worried about shared object size. I don't want to lose optimization opportunities.

Maybe we can replace it with -fvisibility-inlines-hidden -fno-semantic-interposition? (is -fno-sematic-interposition the default?)

tchaikov:
> I'm not worried about shared object size. I don't want to lose optimization opportunities.

i see.

> 
> Maybe we can replace it with -fvisibility-inlines-hidden -fno-semantic-interposition? (is -fno-sematic-interposition the default?)

GCC
-----

it is enabled by default. guess GCC intends to do the right thing as the ELF spec suggests. and it is off only when `-Ofast` or explicitly disabled. please note, `-O3` does not disable it . see also https://github.com/gcc-mirror/gcc/commit/458d2c689963d8461d84670a3d8988cd6ecbfd81 .

Clang
------

it is off by default for better performance. see https://github.com/llvm/llvm-project/commit/d1fd72343c6ff58a3b66bc0df56fed9ac21e4056 . to be more specific, i am quoting the commit message here

> 
> * -fsemantic-interposition: make some linkages interposable and make default visibility external linkage definitions dso_preemptable.
> * (default): selected if the target supports .Lfoo$local: make default visibility external linkage definitions dso_local
> * -fhalf-no-semantic-interposition: if neither option is set or the target does not support .Lfoo$local: like -fno-semantic-interposition but local aliases are not used. So references can be interposed if not optimized out.

so we do want to optimize `-fpic`, i will send a change to include `-fno-semantic-interposition` in compile options if we build shared libraries.

tchaikov:
> so we do want to optimize `-fpic`, i will send a change to include `-fno-semantic-interposition` in compile options ~if we build shared libraries~.
> 
> EDIT, if we don't care about semantic interposition, i think we can disable it for all builds.

created #2465

tchaikov:
@avikivity could you please review my responses? i believe i've addressed all of your questions.

tchaikov:
@avikivity hi Avi, could you help re-review this change ?

tchaikov:
we need to pass `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` to build shared and static libraries in a single pass, this does not change the way how the shared libraries are built. but this would build shared library with `-fPIC`. this introduces overhead due to

- the spatial overhead due to the global offset table (GOT)
- runtime overhead due to the address resolution.

  * if a seastar application addresses a global variable in seastar library, the application needs to look up its address in GOT
  * if a seastar application calls a function in seastar library, the application needs to lookup its addresses in PLT, and resolve the name using  `ld.so`, cache the resolved address in GOT. so that only a single extra indirect memory reference is necessary in the next call. but still, it's not as performant as non-pic code.
- the address resolution might also incur the register pressure -- registers are used for the extra accesses to the GOT

so i don't think we should build the objects for both static and shared libraries in a single pass. it speeds up the build time performance at the expense of shared libraries' performance at runtime .

lersek:
Also, even for a dynamically linked Seastar application, I guess that at least some of the runtime overhead should be possible to offset, by launching the application with `LD_BIND_NOW` set in the environment. The startup could take longer, but later behavior might be more predictable (no PLT / GOT manipulation after launch).

tchaikov:
thank you for your comment Laszlo. i agree with you `LD_BIND_NOW` helps to address the "cold start" issue. but the address translation with GOT cannot be eliminated even the addresses are cached eagerly now lazily. anyway, i wanted to assure that posterity won't be worried about the performance regression introduced by the `-fPIC` when reviewing this change.

lersek:
Right; AIUI, *reading* the GOT all the time cannot be eliminated for dynamically linked applications.

PR #2429, opened 2024-09-12
Coroutinize http client code

xemul:


include/seastar/http/client.hh
```diff
@@ -191,7 +191,7 @@ private:
     future<> shrink_connections();
 
     template <std::invocable<connection&> Fn>
-    auto with_connection(Fn&& fn, abort_source*);
+    futurize_t<std::invoke_result_t<Fn, connection&>> with_connection(Fn fn, abort_source*);
```
tchaikov, line 194: heh, i believe Avi would like this change.

src/http/client.cc
```diff
@@ -336,26 +336,26 @@ future<> client::do_make_request(request req, reply_handler handle, abort_source
 
 future<> client::do_make_request(connection& con, request& req, reply_handler& handle, abort_source* as, std::optional<reply::status_type> expected) {
     auto sub = as ? as->subscribe([&con] () noexcept { con.shutdown(); }) : std::nullopt;
-    return con.do_make_request(req).then([&con, &handle, expected] (connection::reply_ptr reply) mutable {
+    std::exception_ptr ex;
```
tchaikov, line 344: not used.

src/http/client.cc
```diff
@@ -341,13 +341,11 @@ future<> client::do_make_request(connection& con, request& req, reply_handler& h
         connection::reply_ptr reply = co_await con.do_make_request(req);
         auto& rep = *reply;
         if (expected.has_value() && rep._status != expected.value()) {
-            if (!http_log.is_enabled(log_level::debug)) {
-                throw httpd::unexpected_status_error(reply->_status);
+            if (http_log.is_enabled(log_level::debug)) {
```
tchaikov, line 349: i'd prefer early return for less indent. but don't feel strong either way.

src/http/client.cc
```diff
@@ -307,31 +307,36 @@ future<> client::make_request(request req, reply_handler handle, abort_source& a
...
 future<> client::do_make_request(request req, reply_handler handle, abort_source* as, std::optional<reply::status_type> expected) {
-    return do_with(std::move(req), std::move(handle), [this, as, expected] (request& req, reply_handler& handle) mutable {
-        return with_connection([this, &req, &handle, as, expected] (connection& con) {
-            return do_make_request(con, req, handle, as, expected);
-        }, as).handle_exception_type([this, &req, &handle, as, expected] (const std::system_error& ex) {
+    bool retry_with_new = false;
+    try {
+        co_await with_connection(coroutine::lambda([this, &req, &handle, as, expected] (connection& con) -> future<> {
+            co_await do_make_request(con, req, handle, as, expected);
+        }), as);
```
tchaikov, line 314: probably we could add a `co_return` here? just for better readability. this helps reader understand that "if every goes fine, the princess and the ~~frog~~ prince lived happily ever after. FIN.".

tchaikov (commented):
left some comments.

src/http/client.cc
xemul: This is to de-duplicate two throw-s of the same exception

src/http/client.cc
xemul: Yikes

src/http/client.cc
xemul: Hm...

xemul:
upd:
- removed unused `ex` variable
- simplified retry coroutinization in `client::do_make_request (con-less overload)` patch

include/seastar/http/client.hh
xemul: It has less than 4 closing corner braces in a row, so not that bad

Meanwhile in scylla (scroll right):

```c++
test/boost/user_function_test.cc:769:        e.execute_cql("CREATE FUNCTION my_func(val int) CALLED ON NULL INPUT RETURNS map<int, frozen<set<frozen<list<frozen<tuple<text, bigint>>>>>>> LANGUAGE Lua AS  \
```

:scream:

tchaikov (approved):
lgtm

include/seastar/http/client.hh
tchaikov: ah, i meant, he always prefers an explicit return type. this one make it more explicit. so..

gleb-cloudius:
As a general note when I tried to co-routinize cql server in Scylla I did not find a way to do so without introducing more allocations when requests are pipelined.  With continuations if we read more than one request from a socket all of them can be processed without allocating continuations since futures will be ready. With co-routins we always allocate.

avikivity:
With [[clang::coro_elide_safe]] nested coroutines can share the coroutine frame. However it only arrives in clang 20 and isn't standardized.

src/http/client.cc
```diff
@@ -92,19 +92,14 @@ future<> connection::write_body(const request& req) {
 }
 
 future<connection::reply_ptr> connection::maybe_wait_for_continue(const request& req) {
-    if (req.get_header("Expect") == "") {
-        return make_ready_future<reply_ptr>(nullptr);
+    if (req.get_header("Expect") != "") {
+        co_await _write_buf.flush();
```
avikivity: The common case is, that Expect will be empty, no? so we add an allocation here.

Please measure the performance of the series.

xemul: > The common case is, that Expect will be empty, no? so we add an allocation here.

The allocation is unavoidable here anyway, at the end this method does `co_await recv_reply()` and it will wait for it

> Please measure the performance of the series.

ok

src/http/client.cc
```diff
@@ -77,21 +78,16 @@ connection::connection(connected_socket&& fd, internal::client_ref cr)
...
+            co_await _write_buf.write("0\r\n\r\n");
         }
-        return req.body_writer(internal::make_http_chunked_output_stream(_write_buf)).then([this] {
-            return _write_buf.write("0\r\n\r\n");
-        });
     } else if (!req.content.empty()) {
-        return _write_buf.write(req.content);
-    } else {
-        return make_ready_future<>();
+        co_await _write_buf.write(req.content);
```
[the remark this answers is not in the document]
xemul: Yes, but all users we have use lambda body writer and don't step on this branch. But lambda body writer also adds allocation here

nyh:
Hi @xemul,

When you created a 28-patch series which must have taken you many many hours to create, please spend another 10 minutes trying to explain your motivation for doing this. Was it just the generic "coroutines are nicer" sentiment that we all shared, or something specific which bothered you? 

Also, I see a lot of discussion regarding the lower performance of the coroutine code, so I wonder if you made an attempt to benchmark the http server somehow to see if there's anything we need to worry about.

xemul:
> Hi @xemul,
> 
> When you created a 28-patch series which must have taken you many many hours to create, please spend another 10 minutes trying to explain your motivation for doing this. Was it just the generic "coroutines are nicer" sentiment that we all shared, or something specific which bothered you?

No, nothing specific. It just looks much nicer and simpler this way.

> Also, I see a lot of discussion regarding the lower performance of the coroutine code, so I wonder if you made an attempt to benchmark the http server somehow to see if there's anything we need to worry about.

I made a silly test with existing http client demo, but it's not very representative. I'll update the discussion thread with better measurement

src/http/client.cc
xemul: > Please measure the performance of the series.

With #2442 , the test makes 1M requests in
* master: 3.40 +/- 0,02 seconds (min 3,37 / max 3,50)
* this PR: 3,48 +/- 0,02 seconds (min 3,45 / max 3,62)

avikivity: Please report allocs/op

xemul: * master: 45 allocs/req
* this PR: 46 allocs/req

PR #2430, opened 2024-09-12
build: drop the workaround for a GCC bug

tchaikov:
this change partially reverts 87a6540906, which workarounded a bug in GCC-10 and LLVM-3.8, see
https://gcc.gnu.org/git/?p=gcc.git&a=commit;h=1842a04ce06a470b731bc9a29bfec17e6e147c3f and
https://github.com/llvm/llvm-project/commit/894d5821e7af9128398ec8153f3cb4025ae9ad50 respectively.

since scylla builds with GCC-13 and GCC-14, LLVM-17 and LLVM-18. so we don't need to keep this workaround anymore.


tchaikov:

tchaikov:

PR #2434, opened 2024-09-13
addr2line: type checking, debugging

travisdowns:
This series has two changes:

Add additional type hints and a few other type related changes to seastar-addr2line and addr2line.py. After this these scripts type check cleanly with pyright strict.

Add a --debug arg which actives additional explicit debug logging, quite useful for diagnosing what is going on.

`./seastar-addr2line --test` passes after these two changes.

### DIFF
--- a/scripts/addr2line.py
+++ b/scripts/addr2line.py
@@ -57,7 +57,14 @@ class Addr2Line:
         r"(.*0x0: \?\? at .*\n)"  # llvm-addr2line pattern
     )
 
-    def __init__(self, binary: str, concise: bool = False, cmd_path: str = "addr2line"):
+    def __init__(
+        self,
+        parent: 'BacktraceResolver',
+        binary: str,
+        concise: bool = False,
+        cmd_path: str = "addr2line",
+    ):
+        self._parent = parent
         self._binary = binary
 
         # Print warning if binary has no debug info according to `file`.
@@ -68,9 +75,10 @@ class Addr2Line:
         if s.find('ELF') >= 0 and s.find('debug_info', len(self._binary)) < 0:
             print('{}'.format(s))
 
-        options = f"-{'C' if not concise else ''}fpia"
+        args = [cmd_path, f"-{'C' if not concise else ''}fpia", "-e", self._binary]
+        self._parent.debug(f"Addr2line invoking: {' '.join(args)}")
         self._input_proc = subprocess.Popen(
-            [cmd_path, options, "-e", self._binary],
+            args,
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
             universal_newlines=True,
@@ -105,7 +113,8 @@ class Addr2Line:
         return notNone(self._output_proc.stdout)
 
     def _read_resolved_address(self):
-        res = self._output.readline()
+        first = self._output.readline()
+        self._parent.debug('Addr2Line read output (first): ', first)
         # remove the address
         res = first.split(': ', 1)[1]
         while True:
@@ -122,7 +131,9 @@ class Addr2Line:
             return " ".join([self._binary, address, '\n'])
         # We print a dummy 0x0 address after the address we are interested in
         # which we can look for in _read_address
-        self._input.write(address + '\n0x0\n')
+        inputline = address + '\n0x0\n'
+        self._parent.debug('Add2Line sending input to stdin:', inputline)
+        self._input.write(inputline)
         self._input.flush()
         return self._read_resolved_address()
 
@@ -132,7 +143,7 @@ class KernelResolver:
 
     LAST_SYMBOL_MAX_SIZE = 1024
 
-    def __init__(self, kallsyms: str = '/proc/kallsyms'):
+    def __init__(self, parent: 'BacktraceResolver', kallsyms: str = '/proc/kallsyms'):
         syms: list[tuple[int, str]] = []
         ksym_re = re.compile(r'(?P<addr>[0-9a-f]+) (?P<type>.+) (?P<name>\S+)')
         warnings_left = 10
@@ -329,7 +340,9 @@ class BacktraceResolver:
         verbose: bool = False,
         concise: bool = False,
         cmd_path: str = 'addr2line',
+        debug: bool = False,
     ):
+        self._debug = debug
         self._executable = executable
         self._kallsyms = kallsyms
         self._current_backtrace: list[tuple[str, str]] = []
@@ -351,12 +364,17 @@ class BacktraceResolver:
         )  # fail fast if there is something wrong with the exe resolver
         self.parser = self.BacktraceParser()
 
+    def debug(self, *args: Any):
+        if self._debug:
+            print('DEBUG >>', *args, file=sys.stderr)
+
     def _get_resolver_for_module(self, module: str):
         if not module in self._known_modules:
             if module == KERNEL_MODULE:
-                resolver = KernelResolver(kallsyms=self._kallsyms)
+                resolver = KernelResolver(self, kallsyms=self._kallsyms)
             else:
-                resolver = Addr2Line(module, self._concise, self._cmd_path)
+                resolver = Addr2Line(self, module, self._concise, self._cmd_path)
+            self.debug(f'Adding resolver {resolver} for module: {module}')
             self._known_modules[module] = resolver
         return self._known_modules[module]
 
@@ -422,6 +440,10 @@ class BacktraceResolver:
 
         self._known_backtraces[backtrace] = self._i
 
+        self.debug(
+            f'Resolving and printing parsed backtrace with {len(self._current_backtrace)} frames'
+        )
+
         print("[Backtrace #{}]".format(self._i))
 
         for module, addr in self._current_backtrace:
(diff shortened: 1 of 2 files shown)


scripts/seastar-addr2line
```diff
@@ -342,7 +345,6 @@ There are three operational modes:
                            context_re=args.match,
                            verbose=args.verbose,
                            cmd_path=args.addr2line) as resolve:
-        p = re.compile(r'\W+')
```
travisdowns, line 345: this was dead code

scripts/addr2line.py
```diff
@@ -263,27 +287,27 @@ def get_prefix(s):
             #print(f">>> '{line}': None")
             return None
 
-    def __init__(self, executable, kallsyms='/proc/kallsyms', before_lines=1, context_re='', verbose=False, concise=False, cmd_path='addr2line'):
+    def __init__(self, executable: str, kallsyms: str = '/proc/kallsyms', before_lines: int = 1, context_re: Optional[str] = '', verbose: bool = False, concise: bool = False, cmd_path: str = 'addr2line'):
         self._executable = executable
         self._kallsyms = kallsyms
-        self._current_backtrace = []
+        self._current_backtrace : list[tuple[str, str]] = []
```
tchaikov, line 302: nit, might want to put `:` next to `self._current_backtrace` , like:
```py
self._current_backtrace: list[tuple[str, str]] = []
```

scripts/seastar-addr2line
```diff
@@ -51,13 +51,16 @@ def read_backtrace(stdin):
         if char == '' or linefeeds > 1:
             break
 
+TestResult = dict[str, Any]
+MaybeTestResult = Union[TestResult, None]
```
tchaikov, line 55: could use `Optional[TestResult]`.

tchaikov (commented):
left some comment.

scripts/addr2line.py
```diff
@@ -25,11 +25,18 @@
...
 from enum import Enum
 from functools import cache
-from typing import Any
+from typing import Any, Optional, TypeVar, Union, cast
 
 # special binary path/module indicating that the address is from the kernel
 KERNEL_MODULE = '<kernel>'
 
+T = TypeVar('T')
+
```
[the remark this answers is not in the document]
travisdowns: Agreed, and thanks for all of these comments. In fact, I think it would make sense to sprinkle some auto-format magic & enforcement on this project, WDYT? I find auto-format works well in Python, and it seems like `black` aligns quite closely with the existing format (and most of the deltas are just inconsistencies in the existing formatting).

In that spirit I submitted https://github.com/scylladb/seastar/pull/2441 which uses black as the formatting for these 2 addr2line files and enforces it in CI. If that goes through I'll rebase these changes on top of that and reformat which should resolve your formatting related concerns.

scripts/addr2line.py
```diff
@@ -25,11 +25,18 @@
...
+def notNone(o: Union[T, None]) -> T:
```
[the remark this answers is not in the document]
travisdowns: Good call, done in next push!

scripts/addr2line.py
```diff
@@ -54,46 +61,55 @@ def __init__(self, binary, concise=False, cmd_path="addr2line"):
...
         # guarantee on what timeout is sufficient.
-        self._input.stdin.write('\n')
-        self._input.stdin.flush()
-        res = self._output.stdout.readline()
+        self._input().write('\n')
+        self._input().flush()
+        res = self._output().readline()
         self._missing = res == ''
 
+    def _input(self):
```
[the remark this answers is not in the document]
travisdowns: Good call, made it a property.

About the return type, this is already inferred by the type checker: mostly you only need to annotation method arguments, and occasionally local variables, etc, but then the vast majority of types are inferred including return values. So it's a lot of work to add return value for little benefit, I think?

scripts/addr2line.py
```diff
@@ -54,46 +61,55 @@ def __init__(self, binary, concise=False, cmd_path="addr2line"):
...
+        """Returns the input stream for the process/pipe."""
+        return notNone(self._input_proc.stdin)
+
+
```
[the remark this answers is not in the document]
travisdowns: Agreed, this is fixed by black too.

scripts/addr2line.py
```diff
@@ -201,12 +219,18 @@ def split_addresses(self, addrstring: str, default_path=None):
                 addresses.append({'path': m.group(1) or default_path, 'addr': m.group(2)})
             return addresses
 
-        def __call__(self, line):
-            def get_prefix(s):
+        def __call__(self, line: str):
+
+            # would be better represented by a TypedDict but that requires newish Python
+            # LineResult = dict[str, Union[None, self.Type, str, list[dict[str, Any]]]]
```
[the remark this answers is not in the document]
travisdowns: Oops, this line shoudl just be deleted, it's vestigial.

scripts/addr2line.py
travisdowns: good eye, fixed! (even more reason for black :))

scripts/seastar-addr2line
travisdowns: Good call, done!

scripts/addr2line.py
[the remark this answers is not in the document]
tchaikov: i was thinking about human readers actually. but yeah, the type checker can deduce this.

scripts/addr2line.py
```diff
@@ -54,46 +62,55 @@ def __init__(self, binary, concise=False, cmd_path="addr2line"):
...
 
-    def __call__(self, address):
+    def __call__(self, address: str):
         if self._missing:
             return " ".join([self._binary, address, '\n'])
         # We print a dummy 0x0 address after the address we are interested in
         # which we can look for in _read_address
-        self._input.stdin.write(address + '\n0x0\n')
-        self._input.stdin.flush()
+        self._input().write(address + '\n0x0\n')
```
tchaikov: if `_input` is a property, we don't need *call* it for retrieving the value returned by it, am i right?

tchaikov:
and you might want to rebase this changeset against master HEAD.

scripts/addr2line.py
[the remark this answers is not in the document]
travisdowns: (it has been deleted)

scripts/addr2line.py
travisdowns: Sorry, this is correctly without `()` after the 2nd commit in this series, but [absorb](https://github.com/tummychow/git-absorb) assigned it to the wrong commit in this case, let me fix that.

travisdowns: Fixed in https://github.com/scylladb/seastar/commit/db9e2ff7afa96c0e533b55c5b2abbdcf755e8f46.

travisdowns:
> and you might want to rebase this changeset against master HEAD.

Done in push https://github.com/scylladb/seastar/commit/e7b28669b1547fbfdf1a3656bb6f178179faae99.

avikivity:
@bhalevy please review too

scripts/addr2line.py
```diff
@@ -347,12 +363,17 @@ def __init__(
         )  # fail fast if there is something wrong with the exe resolver
         self.parser = self.BacktraceParser()
 
+    def debug(self, *args: Any):
+        if self._debug:
+            print('DEBUG >>', *args, file=sys.stderr)
```
bhalevy, line 368: Why not using a logger?

bhalevy: is it because we want to print to stderr?

bhalevy: otherwise, LGTM

travisdowns: I do want to print it to stderr, yes, to separate it from the primary output which goes to stdout (and which would often be redirected elsewhere). I don't think that excludes using a logger though, does it?

I didn't use a logger because this was easier I think.

scripts/addr2line.py
```diff
@@ -60,48 +69,57 @@ def __init__(self, binary, concise=False, cmd_path="addr2line"):
...
         # guarantee on what timeout is sufficient.
-        self._input.stdin.write('\n')
-        self._input.stdin.flush()
-        res = self._output.stdout.readline()
+        self._input.write('\n')
+        self._input.flush()
+        res = self._output().readline()
         self._missing = res == ''
 
+    @property
```
tchaikov, line 105: the last nit: can we make similar changes to `self._output`, so they are symmetric ?

scripts/addr2line.py
bhalevy: @fruch WDYT?  Should we use a logger in seastar-addr2line or printing to stderr good enough?

fruch: Logger is needed if you want to have context information, like timestamps, filename, line numbers and such.

Or when you have multiple destinations, like stdout and log file.

For one file script like this one its not a must.

travisdowns: Great, I elect not to use a logger then (though ofc still happy to change this if the wind should shift).

tchaikov:

scripts/addr2line.py
travisdowns: Thanks for the suggestion, done in most recent push.

tchaikov:

PR #2436, opened 2024-09-15
iostream/http: Fix output_stream::write(temporary_buffer) overload

### LINKED ISSUE #1701: In `seastar::http::reply::write_body`, using `output_stream::write(temporary_buffer)` causes crash
When using `seastar::http::http_server` and when replying to a request in streaming mode using `reply::write_body`, the following causes a crash:

```c++
void fn(seastar::http::reply& rep) {
    rep.write_body("bin", [](seastar::output_stream<char>&& stream_) -> future<> {
        auto stream = std::move(stream_);
        co_await stream.write(seastar::temporary_buffer<char>{3}));
        co_await stream.close();
    }
}
```
The equivalent code using `write(buff.begin, buff.size())` works.

This seems to be because `write(temporary_buffer)` is implemented in terms of `net::packet`. In this example, `close` triggers a flush, which calls `seastar::http::internal::http_chunked_data_sink_impl::put(seastar::net::packet)`, which is not implemented. 

There may be a good reason for this, but at least it would be nice if this were documented.

niekbouman:
Previously, calling the `write(temporary_buffer<char>)` overload on an `output_stream<char>` obtained from `http::reply::write_body` would cause an abort.

This commit fixes this by providing a helper function (in the `data_sink_impl` base class) named `fallback_put`, which classes that inherit from this base class can use to implement the `put(net::packet)` overload. Also, this commit lets `http_chunked_data_sink_impl` and `http_content_length_data_sink_impl` make use of this helper.

`fallback_put(net::packet)` works simply by extracting the temporary_buffers contained in the packet, and calling the `put(temporary_buffer<char>)` overload for each `temporary_buffer`.

Note that we create this `fallback_put` function instead of defining the virtual member `put(net::packet)` function in the `data_sink_impl` class, to avoid `put(net::packet)` and `put(temporary_buffer<char>)` calling each other leading to infinite recursion.

This commit also adds unit tests to `tests/unit/httpd_test.cc` for testing the `write` overload with `http_chunked_data_sink_impl` and `http_content_length_data_sink_impl`.

Thanks to Pavel (xemul) for his help.

Fixes #1701

### DIFF
--- a/include/seastar/core/iostream.hh
+++ b/include/seastar/core/iostream.hh
@@ -36,6 +36,7 @@
 #pragma once
 
 #include <seastar/core/future.hh>
+#include <seastar/core/coroutine.hh>
 #include <seastar/core/temporary_buffer.hh>
 #include <seastar/core/scattered_message.hh>
 #include <seastar/util/std-compat.hh>
@@ -142,6 +143,20 @@ public:
 
     virtual void on_batch_flush_error() noexcept {
         assert(false && "Data sink must implement on_batch_flush_error() method");
+    }
+
+protected:
+    // This is a helper function that classes that inherit from data_sink_impl
+    // can use to implement the put overload for net::packet.
+    // Unfortunately, we currently cannot define this function as
+    // 'virtual future<> put(net::packet)', because we would get infinite
+    // recursion between this function and
+    // 'virtual future<> put(temporary_buffer<char>)'.
+    future<> fallback_put(net::packet data) {
+        auto buffers = data.release();
+        for (temporary_buffer<char>& buf : buffers) {
+            co_await this->put(std::move(buf));
+        }
     }
 };
 
--- a/src/http/common.cc
+++ b/src/http/common.cc
@@ -119,7 +119,9 @@ class http_chunked_data_sink_impl : public data_sink_impl {
 public:
     http_chunked_data_sink_impl(output_stream<char>& out) : _out(out) {
     }
-    virtual future<> put(net::packet data)  override { abort(); }
+    virtual future<> put(net::packet data) override {
+        return data_sink_impl::fallback_put(std::move(data));
+    }
     using data_sink_impl::put;
     virtual future<> put(temporary_buffer<char> buf) override {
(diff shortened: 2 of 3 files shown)


include/seastar/core/iostream.hh
```diff
@@ -143,6 +144,21 @@ public:
...
+    // This is a helper function that classes that inherit from data_sink_impl
+    // can use to implement the put overload for net::packet.
+    // Unfortunately, we currently cannot define this function as
+    // 'virtual future<> put(net::packet)', because we would get infinite
+    // recursion between this function and
+    // 'virtual future<> put(temporary_buffer<char>)'.
+    future<> fallback_put(net::packet data) {
+        auto buffers = data.release();
+        for (temporary_buffer<char>& buf : buffers)
+        {
```
avikivity, line 158: Please follow our coding style of placing the opening brace on the previous line.

src/http/common.cc
```diff
@@ -119,7 +119,9 @@ class http_chunked_data_sink_impl : public data_sink_impl {
 public:
     http_chunked_data_sink_impl(output_stream<char>& out) : _out(out) {
     }
-    virtual future<> put(net::packet data)  override { abort(); }
+    virtual future<> put(net::packet data) override {
```
xemul, line 122: The http_content_length_data_sink_impl deserves the same update

include/seastar/core/iostream.hh
niekbouman: done.

src/http/common.cc
niekbouman: Done.

I am unfamiliar with this data sink; would you know a useful way to include a unit test for this one?

xemul: Currently only client instantiates one. Check test_basic_content() helper for the example of creating and using a client

niekbouman: Sorry for the delay, and thank you for your suggestion, I've now added a unit test for it via `test_basic_content()`.

niekbouman:
@xemul could you please review?

PR #2437, opened 2024-09-16
build: Add options to select sanitizers in configure.py

devDDen:
We are going to add ThreadSanitizer that can't be used together with
AddressSanitizer. We want to choose which sanitizers to use.

Sanitizers can be specified as semicolon separated list:
```
./configure.py --sanitizers address;undefined_behavior
```

Specified sanitizers are passed to Seastar_SANITIZERS list.

Enabling sanitizers is consistently with Seastar_SANITIZE option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -375,7 +375,13 @@ set (Seastar_SANITIZE
   "DEFAULT"
   CACHE
   STRING
-  "Enable ASAN and UBSAN. Can be ON, OFF or DEFAULT (which enables it for Debug and Sanitize)")
+  "Enable sanitizers. Can be ON, OFF or DEFAULT (which enables it for Debug and Sanitize)")
+
+set (Seastar_SANITIZERS
+  "address;undefined_behavior"
+  CACHE
+  STRING
+  "Sanitizers enabled when building Seastar")
 
 set (Seastar_DEBUG_SHARED_PTR
   "DEFAULT"
@@ -391,11 +397,27 @@ set (Seastar_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set (Seastar_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
 set (Seastar_GEN_BINARY_DIR ${Seastar_BINARY_DIR}/gen)
 
+include (TriStateOption)
+tri_state_option (${Seastar_SANITIZE}
+  DEFAULT_BUILD_TYPES "Debug" "Sanitize"
+  CONDITION sanitizers_enabled)
+
+if (NOT sanitizers_enabled)
+  set (Seastar_SANITIZERS "")
+endif ()
+
 #
 # Dependencies.
 #
 
-include (SeastarDependencies)
+set (Seastar_SANITIZERS_DEPENDENCY_STRING ${Seastar_SANITIZERS})
+
+configure_file (
+  ${CMAKE_CURRENT_LIST_DIR}/cmake/SeastarDependencies.cmake.in
+  ${CMAKE_CURRENT_BINARY_DIR}/SeastarDependencies-build.cmake
+  @ONLY)
+
+include (SeastarDependencies-build)
 seastar_find_dependencies ()
 
 # Private build dependencies not visible to consumers
@@ -880,19 +902,16 @@ if (Seastar_DPDK)
       DPDK::dpdk)
 endif ()
 
-include (TriStateOption)
-tri_state_option (${Seastar_SANITIZE}
-  DEFAULT_BUILD_TYPES "Debug" "Sanitize"
-  CONDITION condition)
-if (condition)
+if (sanitizers_enabled)
   if (NOT Sanitizers_FOUND)
     message (FATAL_ERROR "Sanitizers not found!")
   endif ()
-  set (Seastar_Sanitizers_OPTIONS $<${condition}:${Sanitizers_COMPILE_OPTIONS}>)
-  target_link_libraries (seastar
-    PUBLIC
-      $<${condition}:Sanitizers::address>
-      $<${condition}:Sanitizers::undefined_behavior>)
+  set (Seastar_Sanitizers_OPTIONS $<${sanitizers_enabled}:${Sanitizers_COMPILE_OPTIONS}>)
+  foreach (component ${Seastar_SANITIZERS})
+    target_link_libraries (seastar
+      PUBLIC
+        $<${sanitizers_enabled}:Sanitizers::${component}>)
+  endforeach ()
 endif ()
 
 # We only need valgrind to find uninitialized memory uses, so disable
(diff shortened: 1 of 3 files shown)


devDDen:
@tchaikov please review

CMakeLists.txt
```diff
@@ -391,6 +393,19 @@ set (Seastar_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
...
 set (Seastar_GEN_BINARY_DIR ${Seastar_BINARY_DIR}/gen)
 
+include (TriStateOption)
+tri_state_option (${Seastar_SANITIZE}
+  DEFAULT_BUILD_TYPES "Debug" "Sanitize"
+  CONDITION sanitizers_enabled)
+
+if (sanitizers_enabled)
+  if ((NOT Seastar_SANITIZERS) OR (Seastar_SANITIZERS STREQUAL ""))
+    set (Seastar_SANITIZERS ${Seastar_DEFAULT_SANITIZERS})
```
tchaikov, line 403: i'd suggest making `Seastar_SANITIZERS` a CMake option explicitly, like:
```cmake
set (Seastar_SANITIZERS
  "address;undefined_behavior"
  CACHE
  STRING
  "Sanitizers enabled when building Seastar")
```
and if we build with multi-config generator, this option applies to `Debug` and `Sanitize` modes, otherwise, this option applies the the current build.

what do you think?

configure.py
```diff
@@ -171,15 +173,23 @@ def identify_best_standard(cpp_standards, compiler):
 
 MODE_TO_CMAKE_BUILD_TYPE = {'release': 'RelWithDebInfo', 'debug': 'Debug', 'dev': 'Dev', 'sanitize': 'Sanitize' }
 
+SANITIZER_TO_CMAKE = {'address': 'address', 'undefined': 'undefined_behavior'}
```
tchaikov, line 176: why not just use `undefined_behavior`? what's the merit of introducing another layer?

CMakeLists.txt
devDDen: I agree about multi-config generator.

It doesn't make sense to use sanitizers in other modes because they (at least address and thread) won't work with seastar allocator. But you still can do it by passing `Seastar_SANITIZE=yes`.

I can force to use sanitizers in `./configure.py` if they are specified manually and `args.mode != 'all'`. What do you think?

P.S. `./configure.py` does not use multi-config generator

CMakeLists.txt
```diff
@@ -391,6 +393,19 @@ set (Seastar_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set (Seastar_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
 set (Seastar_GEN_BINARY_DIR ${Seastar_BINARY_DIR}/gen)
 
+include (TriStateOption)
+tri_state_option (${Seastar_SANITIZE}
+  DEFAULT_BUILD_TYPES "Debug" "Sanitize"
+  CONDITION sanitizers_enabled)
+
+if (sanitizers_enabled)
```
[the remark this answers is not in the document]
devDDen: fixed

configure.py
devDDen: Agree, the same thing happens in `FindSanitizers.cmake`. I deleted it

CMakeLists.txt
devDDen: Now I left the logic of sanitizers and build modes unchanged. Enabling sanitizers depends on `Seastar_SANITIZE` variable. By default it enables them only for Debug and Sanitize

tchaikov:
sorry for the latency. i will try to review this change in this week.

cmake/SeastarDependencies.cmake
```diff
@@ -141,6 +141,9 @@ macro (seastar_find_dependencies)
   seastar_set_dep_args (rt REQUIRED)
   seastar_set_dep_args (numactl
     OPTION ${Seastar_NUMA})
+  seastar_set_dep_args (Sanitizers
+    COMPONENTS
+      ${Seastar_SANITIZERS})
```
tchaikov, line 146: please note, this file is installed and redistributed by seastar. it is evaluated when a Seastar application consumes the Seastar library via CMake's `find_package(Seastar)`.

there is a potential issue with the current configuration:

1. At the time this file is processed, `Seastar_SANITIZERS` is empty.
1. This may result in an incomplete Seastar CMake configuration for the parent project.
1. If the user doesn't have sanitizer libraries installed, `find_package(Seastar)` still succeeds.
1. However, Seastar libraries compiled with sanitizers enabled may fail to configure or link due to the missing sanitizer libraries, as what we actually have in this file is

   ```cmake
   find_package(Sanitizers COMPONENTS)
   ```

configure.py
```diff
@@ -142,6 +142,8 @@ def standard_supported(standard, compiler='g++'):
 arg_parser.add_argument('--dpdk-machine', default='native', help='Specify the target architecture')
 add_tristate(arg_parser, name='deferred-action-require-noexcept', dest='deferred_action_require_noexcept', help='noexcept requirement for deferred actions', default=True)
 arg_parser.add_argument('--prefix', dest='install_prefix', default='/usr/local', help='Root installation path of Seastar files')
+arg_parser.add_argument('--sanitizer', dest='sanitizers', action='append', default=[], help='Use specified sanitizer')
+arg_parser.add_argument('--no-sanitizers', dest='no_sanitizers', action='store_true', default=False, help='Do not use sanitizers')
```
tchaikov, line 146: after a second thought, i think probably it's simpler to accept a colon-separated or a comma-separated list as the argument as `--sanitizers`? like

```py
g_parser.add_argument('--sanitizers', dest='sanitizers',  default='address;undefined_behavior', help='Use specified sanitizers')
```
what do you think?

cmake/SeastarDependencies.cmake
tchaikov: i am not sure what is the best way to address this. but there are at least two approaches:

- create a separate `cmake/SeastarDependencies.cmake.in`, and generate the cmake file using `configure_file()`, so we can have a different `SeastarDependencies.cmake` for each mode, and then `install()` it
- check all the valid combinations:

  - `address` + `undefined_behavior` 
  - `thread`

CMakeLists.txt
```diff
@@ -883,19 +898,17 @@ if (Seastar_DPDK)
...
     message (FATAL_ERROR "Sanitizers not found!")
   endif ()
-  set (Seastar_Sanitizers_OPTIONS ${Sanitizers_COMPILE_OPTIONS})
-  target_link_libraries (seastar
-    PUBLIC
-      $<${condition}:Sanitizers::address>
-      $<${condition}:Sanitizers::undefined_behavior>)
+  set (Seastar_Sanitizers_OPTIONS $<${sanitizers_enabled}:${Sanitizers_COMPILE_OPTIONS}>)
+  foreach (component ${Seastar_SANITIZERS})
+    string (TOUPPER ${component} COMPONENT)
```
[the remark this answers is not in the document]
devDDen: fixed

CMakeLists.txt
```diff
@@ -391,6 +397,15 @@ set (Seastar_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set (Seastar_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
 set (Seastar_GEN_BINARY_DIR ${Seastar_BINARY_DIR}/gen)
 
+include (TriStateOption)
+tri_state_option (${Seastar_SANITIZE}
+  DEFAULT_BUILD_TYPES "Debug" "Sanitize"
+  CONDITION sanitizers_enabled)
```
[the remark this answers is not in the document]
devDDen: if sanitizers are disabled for this build type we don't need to find them

configure.py
devDDen: done

CMakeLists.txt
```diff
@@ -883,19 +898,17 @@ if (Seastar_DPDK)
...
+if (sanitizers_enabled)
   if (NOT Sanitizers_FOUND)
     message (FATAL_ERROR "Sanitizers not found!")
   endif ()
-  set (Seastar_Sanitizers_OPTIONS ${Sanitizers_COMPILE_OPTIONS})
-  target_link_libraries (seastar
-    PUBLIC
-      $<${condition}:Sanitizers::address>
-      $<${condition}:Sanitizers::undefined_behavior>)
+  set (Seastar_Sanitizers_OPTIONS $<${sanitizers_enabled}:${Sanitizers_COMPILE_OPTIONS}>)
```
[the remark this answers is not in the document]
devDDen: done in this PR

tchaikov:
@devDDen thank you for addressing the comments. i will try to review your change this week.

devDDen:
@tchaikov ping

devDDen:
@tchaikov is there any progress in reviewing?

cmake/SeastarDependencies.cmake
tchaikov: @devDDen hi Denis, sorry for the latency. i just tested your change after rebasing it onto the latest master HEAD. it seems the generator expression cannot be expanded in the files generated with `configure_file()` -- which only substitute *variables*. anyway, i created #2746 for our further discussion. it supports non-multi-config generators, but it does not support multi-config generators.

tchaikov:
@devDDen replied at https://github.com/scylladb/seastar/pull/2437#discussion_r2041811278

PR #2439, opened 2024-09-17
README.md: use the github homepage of Ceph for Crimson

tchaikov:
Reef is a major version of Ceph, and the URL of "Ceph - Crimson" is a page targeting Crimson developers. We should instead reference the project's homepage or its github project address like other projects.

so, in this change, let's reference Ceph's project page on GitHub.


mykaul:
Since Crimson is not yet GA, I think the current link makes more sense.

tchaikov:
yeah, it's technical preview. but i don't think linking Crimson to Reef and a page for developers makes sense.

PR #2442, opened 2024-09-18
test: Add perf test for http client

xemul:
The test opens a server socket that reads socket up until double CRLF and then responds back with "HTTP/1.1 200 OK host: test" line in a loop. Then runs http client against it measuring time it takes to perform the given amount of requests. It uses in-memory loopback sockets (seastar wrapper around seastar::queue).

### DIFF
--- a/tests/perf/CMakeLists.txt
+++ b/tests/perf/CMakeLists.txt
@@ -88,3 +88,7 @@ seastar_add_test (allocator
 
 seastar_add_test (container
   SOURCES container_perf.cc)
+
+seastar_add_test (http_client
+  SOURCES http_client_perf.cc linux_perf_event.cc
+  NO_SEASTAR_PERF_TESTING_LIBRARY)
--- a/tests/perf/http_client_perf.cc
+++ b/tests/perf/http_client_perf.cc
@@ -1,0 +1,182 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2024 ScyllaDB
+ */
+
+/*
+ * The test runs http::experimental::client against minimalistic (see below) server on
+ * one shard using single "in-memory" connection.
+ *
+ * The client sents one request at-a-time, waiting for the server response before sending
+ * the next one.
+ *
+ * The server is a fiber that runs on top of the raw connection, reads it up until double
+ * CRLF and then responds back with the "HTTP/1.1 200 OK host: test" line. So it's not
+ * http::server instance, but a lightweight mock.
+ *
+ * The connection is net::connected_socket wrapper over seastar::queue, not Linux socket.
+ */
+
+#include <seastar/core/seastar.hh>
+#include <seastar/core/reactor.hh>
+#include <seastar/core/app-template.hh>
+#include <seastar/core/sstring.hh>
+#include <seastar/core/when_all.hh>
+#include <seastar/core/thread.hh>
+#include <seastar/http/client.hh>
+#include <seastar/http/request.hh>
+#include <seastar/testing/linux_perf_event.hh>
+#include <../../tests/unit/loopback_socket.hh>
+#include <fmt/printf.h>
+#include <string>
+
+using namespace seastar;
+using namespace std::chrono_literals;
+
+class server {
+    seastar::server_socket _ss;
+    seastar::connected_socket _cs;
+    seastar::input_stream<char> _in;
+    seastar::output_stream<char> _out;
+    sstring _req;
+
+    future<> run_serve_loop() {
+        while (true) {
+            temporary_buffer<char> buf = co_await _in.read();
+            if (buf.empty()) {
+                co_return;
+            }
+
+            _req += sstring(buf.get(), buf.size());
+            if (_req.ends_with("\r\n\r\n")) {
+                sstring r200("HTTP/1.1 200 OK\r\nHost: test\r\n\r\n");
+                co_await _out.write(r200);
+                co_await _out.flush();
+                _req = "";
+            }
+        }
+    }
+
+public:
+    server(loopback_connection_factory& lcf) : _ss(lcf.get_server_socket()) {}
+    future<> serve() {
+        return _ss.accept().then([this] (seastar::accept_result ar) {
+            _cs = std::move(ar.connection);
+            _in = _cs.input();
+            _out = _cs.output();
+            return run_serve_loop().finally([this] {
+                return when_all(_in.close(), _out.close()).discard_result();
+            });
+        });
+    }
+};
+
+class loopback_http_factory : public http::experimental::connection_factory {
+    loopback_socket_impl lsi;
+public:
+    explicit loopback_http_factory(loopback_connection_factory& f) : lsi(f) {}
+    virtual future<connected_socket> make(abort_source* as) override {
+        return lsi.connect(socket_address(ipv4_addr()), socket_address(ipv4_addr()));
+    }
+};
+
+class client {
+    seastar::http::experimental::client _cln;
+    const unsigned _warmup_limit;
+    const unsigned _limit;
+    linux_perf_event _instructions;
+    linux_perf_event _cpu_cycles;
+
+    struct stats {
+        std::chrono::steady_clock::time_point ts;
+        uint64_t mallocs;
+        uint64_t tasks;
+        uint64_t instructions;
+        uint64_t cpu_cycles;
+    };
+
+    stats stats_snapshot() {
+        return stats {
+            .ts = std::chrono::steady_clock::now(),
+            .mallocs = memory::stats().mallocs(),
+            .tasks = engine().get_sched_stats().tasks_processed,
+            .instructions = _instructions.read(),
+            .cpu_cycles = _cpu_cycles.read(),
+        };
+    }
+
+    future<> make_requests(unsigned nr) {
+        for (unsigned i = 0; i < nr; i++) {
+            auto req = http::request::make("GET", "test", "/test");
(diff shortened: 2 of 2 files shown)


tests/perf/CMakeLists.txt
```diff
@@ -88,3 +88,7 @@ seastar_add_test (allocator
 
 seastar_add_test (container
   SOURCES container_perf.cc)
+
+seastar_add_test (http_client
+ 0 SOURCES http_client_perf.cc
```
tchaikov, line 93: what is this `0` for? might want to drop it.

xemul: Weird indeed, will drop

xemul:
upd:
- fixed trash in CMakefile 
- removed unused capture on lambda
- added total mallocs stats

xemul:
@tchaikov , CI fails compiling a file that's not touched by this PR:

```
[208/360] Building CXX object tests/unit/CMakeFiles/test_unit_rpc.dir/rpc_test.cc.o
FAILED: tests/unit/CMakeFiles/test_unit_rpc.dir/rpc_test.cc.o 
ccache /usr/bin/g++ -DBOOST_ALL_DYN_LINK -DFMT_SHARED -DSEASTAR_API_LEVEL=7 -DSEASTAR_BUILD_SHARED_LIBS -DSEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT -DSEASTAR_DEPRECATED_OSTREAM_FORMATTERS -DSEASTAR_ENABLE_ALLOC_FAILURE_INJECTION -DSEASTAR_HAS_MEMBARRIER -DSEASTAR_HAVE_ASAN_FIBER_SUPPORT -DSEASTAR_HAVE_HWLOC -DSEASTAR_HAVE_NUMA -DSEASTAR_HAVE_SYSTEMTAP_SDT -DSEASTAR_HAVE_URING -DSEASTAR_LOGGER_COMPILE_TIME_FMT -DSEASTAR_LOGGER_TYPE_STDOUT -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_SSTRING -DSEASTAR_TESTING_MAIN -DSEASTAR_TESTING_WITH_NETWORKING=1 -DSEASTAR_TYPE_ERASE_MORE -I/__w/seastar/seastar/tests/unit -I/__w/seastar/seastar/src -I/__w/seastar/seastar/include -I/__w/seastar/seastar/build/dev/gen/include -I/__w/seastar/seastar/build/dev/gen/src -O1 -std=gnu++23 -Wno-maybe-uninitialized -Wno-error=unused-result -fstack-clash-protection -UNDEBUG -Wall -Werror -Wimplicit-fallthrough -Wdeprecated -Wno-error=deprecated -Wno-error=stringop-overflow -Wno-error=array-bounds -Wdeprecated-declarations -Wno-error=deprecated-declarations -ftls-model=initial-exec -gz -MD -MT tests/unit/CMakeFiles/test_unit_rpc.dir/rpc_test.cc.o -MF tests/unit/CMakeFiles/test_unit_rpc.dir/rpc_test.cc.o.d -o tests/unit/CMakeFiles/test_unit_rpc.dir/rpc_test.cc.o -c /__w/seastar/seastar/tests/unit/rpc_test.cc
/__w/seastar/seastar/tests/unit/rpc_test.cc: In lambda function:
/__w/seastar/seastar/tests/unit/rpc_test.cc:1366:89: error: class template argument deduction failed:
 1366 |             return make_ready_future<rpc::tuple<int, long>>(rpc::tuple(1, 0x7'0000'0000L));
      |                                                                                         ^
/__w/seastar/seastar/tests/unit/rpc_test.cc:1366:89: error: call of overloaded 'tuple(int, long int)' is ambiguous
In file included from /usr/include/c++/14/bits/uses_allocator_args.h:39,
                 from /usr/include/c++/14/bits/memory_resource.h:41,
                 from /usr/include/c++/14/string:68,
                 from /usr/include/c++/14/stdexcept:39,
                 from /usr/include/c++/14/system_error:43,
                 from /__w/seastar/seastar/tests/unit/loopback_socket.hh:24,
                 from /__w/seastar/seastar/tests/unit/rpc_test.cc:22:
/usr/include/c++/14/tuple:1970:5: note: candidate: 'std::tuple(_Ts ...)-> seastar::rpc::tuple<T ...> [with T = {int, long int}]'
...
```

tchaikov:
lemme check.

tchaikov:
> lemme check.

i can reproduce this with GCC-14 on fedora 40. seems it's related to ADL.

tchaikov:
filed #2445

xemul:
Thanks! Rebased the branch

tests/perf/http_client_perf.cc
```diff
@@ -0,0 +1,143 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2024 ScyllaDB
+ */
+
+#include <seastar/core/seastar.hh>
```
nyh, line 22: Would have been really nice to have a comment at the top of the file saying what this file tests. Maybe even cut-and-paste from your commit message.

Please also explain somewhere in the code (at the top of the file, or closer to where this happens in the code), exactly which sort of workload is being measured. If I understand the code correctly, you create a single connection and then repeatedly send one request and wait for one response - concurrency isn't being exercised, and it is the latency - not maximum throughput - which is actually being measured. I'm not saying it's necessarily bad, but you (and other developers who might want to use this benchmark) should know what is really being measured.

tests/perf/http_client_perf.cc
```diff
@@ -0,0 +1,143 @@
...
+                return make_ready_future<>();
+            }
+
+            _req += sstring(buf.get(), buf.size());
+            if (_req.ends_with("\r\n\r\n")) {
+                sstring r200("HTTP/1.1 200 OK\r\nHost: test\r\n\r\n");
+                return _out.write(r200).then([this] {
+                    return _out.flush().then([this] {
+                        _req = "";
+                        return do_echo();
```
nyh, line 54: There is the theoretical and extremely unlikely possibility that if in.read() is ready many times in sequence and _out.write() is also ready many times, this code will have very deep recursion and crash. But since this is just a test and since it's so unlikely that this can even happen, I think it's fine.

Maybe everything will be nicer if just written using coroutines and a normal endless loop?

tests/perf/http_client_perf.cc
```diff
@@ -0,0 +1,143 @@
...
+using namespace std::chrono_literals;
+
+class server {
+    seastar::server_socket _ss;
+    seastar::connected_socket _cs;
+    seastar::input_stream<char> _in;
+    seastar::output_stream<char> _out;
+    sstring _req;
+
+    future<> do_echo() {
```
nyh, line 42: The name "echo" seems a misnomer - this server doesn't seem to echo the request (i.e., write the request in the response) but just to always return the same response r200.

Also the verb "do" doesn't really explain what this function is - I think it's basically a loop handling a single connection.

xemul: OK, renamed

tests/perf/http_client_perf.cc
xemul: Added

tests/perf/http_client_perf.cc
xemul: Not only out.write() should be ready, but also out.flush(). The way this test runs (ping-pong mode) it _must_ switch to client fiber after flush(), as client doesn't pipeline requests and generates the new one after it sees response from the server

> Maybe everything will be nicer if just written using coroutines and a normal endless loop?

Nicer, yes, definitely. Also client-part looks nicer with coroutines

xemul:
upd:
- added top-level comment
- renamed server::do_echo -> server::run_serve_loop
- coroutinized server and client core loops

tests/perf/CMakeLists.txt
```diff
@@ -88,3 +88,7 @@ seastar_add_test (allocator
 
 seastar_add_test (container
```
avikivity, line 89: >  It uses in-memory loopback sockets (seastar wrapper around seastar::queue).

But I already prepared mentally to write a comment about it

tests/perf/http_client_perf.cc
```diff
@@ -0,0 +1,150 @@
...
+    sstring _req;
+
+    future<> run_serve_loop() {
+        while (true) {
+            temporary_buffer<char> buf = co_await _in.read();
+            if (buf.empty()) {
+                co_return;
+            }
+
+            _req += sstring(buf.get(), buf.size());
```
avikivity, line 63: This is quadratic, but I expect n=1, so n²=1 too

tests/perf/http_client_perf.cc
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+public:
+    server(loopback_connection_factory& lcf) : _ss(lcf.get_server_socket()) {}
+    future<> serve() {
+        return _ss.accept().then([this] (seastar::accept_result ar) {
+            _cs = std::move(ar.connection);
+            _in = _cs.input();
+            _out = _cs.output();
+            return run_serve_loop();
```
avikivity, line 80: no close()?

tests/perf/http_client_perf.cc
```diff
@@ -0,0 +1,150 @@
...
+        return do_with(loopback_connection_factory(1), [total_ops, warmup_ops] (auto& lcf) {
+            auto sf = do_with(server(lcf), [] (auto& srv) {
+                return srv.serve();
+            });
+            auto cf = do_with(client(lcf, total_ops, warmup_ops), [] (auto& cln) {
+                return cln.work();
+            });
+            return when_all(std::move(sf), std::move(cf)).discard_result();
+        });
+    });
```
avikivity, line 149: coroutine or thread?

Would be nicer to report insn/op and allocs/op.

tests/perf/http_client_perf.cc
xemul: Missed one :( Added in next version

tests/perf/http_client_perf.cc
xemul: Thread looks nicer indeed

Report usec/op and allocs/op

xemul:
upd:
- close server-side streams
- report per-op results
- use async() in main app lambda

xemul:
@avikivity , @nyh , reminder ping

tests/perf/http_client_perf.cc
avikivity: Why not insn/op (in addition)? It's more stable with todays frequency agile processors.

tchaikov: if we don't plan to use `PERF_TEST` because it's complicated to pass command line argument to the individual tests. probably we can still use `linux_perf_event::user_instructions_retired()` directly ?

avikivity: perf-simple-query in scylla

xemul:
upd:
- added insns/op, cpu_cycles/op and tasks/op stats

PR #2443, opened 2024-09-19
scripts/perftune.py: improve a dry-run printout

vladzcloudius:
Let's be optimistic when printing our intention during --dry-run execution: let's print the IRQ affinity configuration intent as if the Rx channels setting was successful.

Without this patch we are going to print a different configuration intent than what we are actually going to configure.

Note: unfortunatelly the same trick is not possible for RPS/XPS configuration intent because different NIC drivers are going to behave differently in regards to rps_cpus and xps_cpus files when one changes the numeber of Rx channels.

### DIFF
--- a/scripts/perftune.py
+++ b/scripts/perftune.py
@@ -1092,6 +1092,10 @@ class NetPerfTuner(PerfTunerBase):
         # For such NICs we've sorted IRQs list so that IRQs that handle Rx are all at the head of the list.
         if rx_channels_set or max_num_rx_queues < len(all_irqs):
             num_rx_queues = self.__get_rx_queue_count(iface)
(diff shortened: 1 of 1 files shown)


vladzcloudius:
Tests failures above don't seem to be related to the patch itself.

cc @xemul

tchaikov:
> CI failures don't seem to be related to the patch itself.
> 
> cc @xemul

tracked by #2445, and has been fixed.

vladzcloudius:
> > CI failures don't seem to be related to the patch itself.
> > cc @xemul
> 
> tracked by #2445, and has been fixed.

Thanks, @tchaikov.
@xemul @avikivity please, consider merging this PR then.

PR #2447, opened 2024-09-20
build: do not print the default value of --c++-standard in help output

tchaikov:
before this change, the output of `./configure.py --help` looks like:
```
  --c++-standard CPP_STANDARD
                        C++ standard to build with [default: ]
```

because the default value of `--c++-standard` option is an empty string. printing out its value is not helpful, and could be confusing.

as the actual default value of this option varies depending on the used compiler, so even if we print out the deduced C++ standard, it is still confusing. not to mention that argparse does not differentiate the defaulted value in the returned namespace from `parse_args()` from the user specified one. so, a simpler approach is just do not specify the default value.

in this change, we

* do not specify the default value for `--c++-standard`, so its value would be `None` if not specified
* do not print its default value in the `--help` output


tchaikov:

PR #2448, opened 2024-09-20
Misc compile include/warn fixes

travisdowns:
Fix wrong includes related to SEASTAR_MODULE
Fix "unused result" warnings in gcc

Tests all pass, see changes for details.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2906,7 +2906,8 @@ reactor::wakeup() {
     _sleeping.store(false, std::memory_order_relaxed);
 
     uint64_t one = 1;
-    ::write(_notify_eventfd.get(), &one, sizeof(one));
+    auto res = ::write(_notify_eventfd.get(), &one, sizeof(one));
+    assert(res == sizeof(one) && "write(2) failed on _reactor._notify_eventfd");
 }
 
 void reactor::start_aio_eventfd_loop() {
@@ -2916,7 +2917,7 @@ void reactor::start_aio_eventfd_loop() {
     future<> loop_done = repeat([this] {
         return _aio_eventfd->readable().then([this] {
             char garbage[8];
-            ::read(_aio_eventfd->get_fd(), garbage, 8); // totally uninteresting
+            std::ignore = ::read(_aio_eventfd->get_fd(), garbage, 8); // totally uninteresting
             return _stopping ? stop_iteration::yes : stop_iteration::no;
         });
     });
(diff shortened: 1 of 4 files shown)


tests/perf/linux_perf_event.cc
```diff
@@ -61,7 +63,8 @@ linux_perf_event::read() {
         return 0;
     }
     uint64_t ret;
-    ::read(_fd, &ret, sizeof(ret));
+    auto res =::read(_fd, &ret, sizeof(ret));
```
tchaikov, line 66: nit, add a space after `=`.

src/core/reactor.cc
```diff
@@ -2916,7 +2917,8 @@ void reactor::start_aio_eventfd_loop() {
     future<> loop_done = repeat([this] {
         return _aio_eventfd->readable().then([this] {
             char garbage[8];
-            ::read(_aio_eventfd->get_fd(), garbage, 8); // totally uninteresting
+            auto res = ::read(_aio_eventfd->get_fd(), garbage, 8); // totally uninteresting
+            (void)res;
```
tchaikov, line 2921: nit, how about using `std::ignore`? i imagine if we can replace all of them with `_` in C++26. as `std::ignore` is easier to grep than `void` , and visually more obvious than the later. so i prefer `std::ignore`.

tchaikov (approved):
thanks. just two nits.

for posterity, we pass `-UNDEBUG` when building seastar, so cannot use `assert()` on the critical pathes.

travisdowns:
> for posterity, we pass `-UNDEBUG` when building seastar, so cannot use `assert()` on the critical pathes.

Yes, thanks for the note. I don't think any of the changes here involve a location that would be a performance concern for `assert`.

src/core/reactor.cc
travisdowns: Thanks of the note, changed to `std::ignore`.

tests/perf/linux_perf_event.cc
travisdowns: Fixed, thanks!

tchaikov:
> > for posterity, we pass `-UNDEBUG` when building seastar, so cannot use `assert()` on the critical pathes.
> 
> Yes, thanks for the note. I don't think any of the changes here involve a location that would be a performance concern for `assert`.

ahh, thought that the `start_aio_eventfd_loop()` one was the loop for reaping the events, i was wrong.

tchaikov (approved):
lgtm

tchaikov:
@scylladb/seastar-maint hello maintainers, could you help review/merge this change?

tchaikov:
@scylladb/seastar-maint hello maintainers, could you help review/merge this change?

PR #2449, opened 2024-09-21
sharded.hh: add invoke_on variant for a shard range

tomershafir:
Add a convenient method to invoke a function on a range of shards, to have better support for the cold control path on more complex compute models like shard groups for different tasks. A workload can benefit from such a model if inter-task cooperation is better when grouping from perf and QoS perspectives, for example. Or a computation requires a only a subset of shards due to internal concurrency limits.

### DIFF
--- a/include/seastar/core/sharded.hh
+++ b/include/seastar/core/sharded.hh
@@ -35,6 +35,8 @@
 #include <boost/iterator/counting_iterator.hpp>
 #include <concepts>
 #include <functional>
+#include <ranges>
+#include <type_traits>
 #endif
 
 /// \defgroup smp-module Multicore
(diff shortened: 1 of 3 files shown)


include/seastar/core/sharded.hh
```diff
@@ -338,6 +340,42 @@ public:
...
+    ///        `future<Value> (Service&, Args...)` (for some `Value` type), or a pointer
+    ///        to a member function of Service
+    /// \param args parameters to the callable; will be copied or moved. To pass by reference,
+    ///              use std::ref().
+    /// \return Future that becomes ready once all calls have completed
+    template <typename Func, typename... Args>
+    requires std::invocable<Func, Service&, Args...>
+        && std::is_same_v<futurize_t<std::invoke_result_t<Func, Service&, internal::sharded_unwrap_t<Args>...>>, future<>>
+    future<>
+    invoke_on(unsigned begin, unsigned end, smp_submit_to_options options, Func func, Args... args) noexcept;
```
avikivity, line 359: Better: pass a std::ranges::range of shards. Then any subset of the shards can be sent. If you want a contiguous range, use std::views::iota.

tomershafir:
@avikivity changed to `std::ranges::range`

PR #2452, opened 2024-09-23
Json2code: full $ref support

bashtanov:
Swagger only allows to refer to custom composite types via "$ref", not via "type", whether it is about array elements or individual items.

Disallowing "type" would break existing use, so allow "$ref" to be used interchangeably.


tchaikov:
@amnonh hi Amnon, could you take a look at this change as well? i plan to review it in this week.

scripts/seastar-json2code.py
```diff
@@ -370,7 +370,13 @@ def is_model_valid(name, model):
...
     properties = getitem(model[name], "properties", name)
     for var in properties:
-        type = getitem(properties[var], "type", name + ":" + var)
+        try:
+            type = getitem(properties[var], "type", name + ":" + var)
+        except Exception as e:
+            try:
+                type = getitem(properties[var], "$ref", name + ":" + var)
+            except:
+                raise e
```
tchaikov, line 379: if `$ref` is the notation supported by swagger spec, probably we should check if first. and i'd suggest just use `dict.get()` . simpler this way, like:

```py
# to be backward compatible, support "type" as well
type = (properties[var].get("$ref") or
        properties[var].get("type"))
if type is None:
  raise KeyError(f"'$ref' not found in {name}:{var}")
```

tchaikov:
also, could be better to use a imperative mood in the commit message, in this case, could use something like:

> seastar-json2code: allow referencing custom types using "$ref"

Issue #2453, opened 2024-09-26
core/reactor: AIO block number is checked regardless of which reactor backend is selected


denesb:
Resulting in failures related to insufficient AIO block number, even when seastar was started with `epoll` reactor backend, which doesn't need them at all.
This means that currently there is no workaround for insufficient AIO blocks, other than tuning `/proc/sys/fs/aio-max-nr`, which requires `sudo`.

PR #2454, opened 2024-09-26
core/reactor: don't check AIO block count when they are not needed

denesb:
Only check for sufficient number of AIO blocks when the linux-aio reactor backend was selected. Otherwise these blocks are not used anyway, but the check might result in failure because it think there aren't sufficient number of AIO blocks available.

Fixes: https://github.com/scylladb/seastar/issues/2453


nyh (approved):
I approve this patch because I think it's good enough, but I think that an even better solution would move the code messing with the max aio checking INTO the "linux-aio" reactor backends, so it simply doesn't run with other backends, and you won't need some string comparison to check it in some main configure() function.

fruch:
@denesb can we get that into scylla ?

PR #2455, opened 2024-09-26
file, scheduling: remove non-unified I/O and CPU scheduling

avikivity:
In f94b1bb9cb8c04b5a ("Merge 'Unify CPU scheduling groups and IO priority classes' from Pavel Emelyanov"), we folded I/O scheduling control into scheduling groups. This was more than a year ago. As it was a breaking change, we introduced API_LEVEL 7 to avoid breaking users. We now make API_LEVEL 7 mandatory and remove non-unified scheduling support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,11 +73,11 @@ set (Seastar_API_LEVEL
   "7"
   CACHE
   STRING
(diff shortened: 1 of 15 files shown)


avikivity:
@xemul review ping

avikivity:
@xemul review ping

PR #2456, opened 2024-09-26
shared_ptr: remove deprecated lw_shared_ptr assignment operator

avikivity:
The assignment operator was deprecated in 8038c131b7c ("shared_ptr: deprecate lw_shared_ptr operator=(T&&)"), more than a year ago. It is safe to remove it now.


nyh (approved):
Looks good. The only risk of removing the deprecated constructor and not leaving a "= delete" behind is that a few years from now, someone might be tempted to add such a constructor again and not know the history ;-)

PR #2458, opened 2024-09-29
allow setting buffer sizes on server_socket

travisdowns:
We add two options to set the recv and send (SO_RCVBUF, ...) buffer sizes on a listening socket (server_socket). This is mostly useful to propagate said sizes to all sockets returned by accept().

It is already possible to set the socket option directly on the connected socket after it returned by accept() but experimentally this results in a socket with the specified buffer size, but whose receive window will not be advertised to the client beyond the default (64K for current typical kernel defaults). So you get only some of the benefit of the larger buffer.

Setting the buffer size on the listening socket however, is mentioned as the correct approach in tcp(7) and does not suffer from the same limitation.

A test is included which checks that the mechanism, including the inheritance, works.

### DIFF
--- a/include/seastar/net/api.hh
+++ b/include/seastar/net/api.hh
@@ -389,6 +389,11 @@ public:
 
 /// @}
 
+/// Options for creating a listening socket.
+///
+/// WARNING: these options currently only have an effect when using
+/// the POSIX stack: all options are ignored on the native stack as they
+/// are not implemented there.
 struct listen_options {
     bool reuse_address = false;
     server_socket::load_balancing_algorithm lba = server_socket::load_balancing_algorithm::default_;
@@ -396,6 +401,19 @@ struct listen_options {
     int listen_backlog = 100;
     unsigned fixed_cpu = 0u;
     std::optional<file_permissions> unix_domain_socket_permissions;
+
+    /// If set, the SO_SNDBUF size will be set to the given value on the listening socket
+    /// via setsockopt. This buffer size is inherited by the sockets returned by
+    /// accept and is the preferred way to set the buffer size for these sockets since
+    /// setting it directly on the already-accepted socket is ineffective (see TCP(7)).
+    std::optional<int> so_sndbuf;
+
+    /// If set, the SO_RCVBUF size will be set to the given value on the listening socket
+    /// via setsockopt. This buffer size is inherited by the sockets returned by
+    /// accept and is the preferred way to set the buffer size for these sockets since
+    /// setting it directly on the already-accepted socket is ineffective (see TCP(7)).
+    std::optional<int> so_rcvbuf;
+
     void set_fixed_cpu(unsigned cpu) {
         lba = server_socket::load_balancing_algorithm::fixed;
         fixed_cpu = cpu;
@@ -450,8 +468,8 @@ public:
         return false;
     }
 
-    /** 
-     * Returns available network interfaces. This represents a 
+    /**
+     * Returns available network interfaces. This represents a
      * snapshot of interfaces available at call time, hence the
      * return by value.
      */
--- a/tests/unit/socket_test.cc
+++ b/tests/unit/socket_test.cc
@@ -27,12 +27,16 @@
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/later.hh>
 #include <seastar/testing/test_case.hh>
+#include <seastar/testing/thread_test_case.hh>
 #include <seastar/core/abort_source.hh>
 #include <seastar/core/sleep.hh>
 #include <seastar/core/thread.hh>
 #include <seastar/core/when_all.hh>
-
+#include <seastar/net/api.hh>
 #include <seastar/net/posix-stack.hh>
+
+#include <optional>
+#include <tuple>
 
 using namespace seastar;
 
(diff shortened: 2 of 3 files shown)


travisdowns:
This was discovered due to very poor throughput between a remote client with ~250 rtt and Redpanda: this transfer is receive window limited and benefits from buffers > 1 MB, but using them had no effect despite the change taking effect per `ss`. The problem was that setting it "too late" prevents the receiving side from advertising the larger size in its receive window. Arguably a kernel flaw? The window scale was set high enough (128x) even when setting it in this way, so that wasn't preventing the window from scaling though this is often given as the reason why "late setting" the recv buffer size does not work.

tests/unit/socket_test.cc
```diff
@@ -258,3 +261,57 @@ SEASTAR_TEST_CASE(socket_connect_abort_test) {
...
+
+SEASTAR_THREAD_TEST_CASE(socket_bufsize) {
+
+    // Test that setting the send and recv buffer sizes on the listening
+    // socket is propagated to the socket returned by accept().
+
+    auto buf_size = [](std::optional<int> buf_size) {
+        listen_options lo{
+            .reuse_address = true,
+            .lba = server_socket::load_balancing_algorithm::fixed,
```
travisdowns, line 273: As we create multiple connections, we need to ensure we LB them all to shard 0, or else the accept() will just hang: this definite caused some confusion for a while.

xemul: Why does it hang?

travisdowns: This test listens connects to the server socket and does an accept() on that socket (lines 280 + 281) expecting to get the connection we just made, all on shard 0, but if load balancing is on the second time through the connection will be send to shard 1, so the accept() will simply hang.

All the other tests just happen to work I think since they only ever accept() a single connection on any server_socket and this first connection always balances to shard 0.

include/seastar/net/api.hh
```diff
@@ -389,13 +389,28 @@ public:
...
     bool reuse_address = false;
     server_socket::load_balancing_algorithm lba = server_socket::load_balancing_algorithm::default_;
     transport proto = transport::TCP;
     int listen_backlog = 100;
     unsigned fixed_cpu = 0u;
     std::optional<file_permissions> unix_domain_socket_permissions;
+
+    /// If set, the SO_SNDBUF size will be set to the given value on the listening socket
+    /// via setsockopt. This buffer size is inherited by the sockets returned by
+    /// accept and is the preferred way to set the buffer size for this sockets since
```
nyh, line 404: typo: should be "these" sockets.

nyh: Same typo in the copy-pasted comment below.

include/seastar/net/api.hh
```diff
@@ -450,8 +465,8 @@ public:
         return false;
     }
 
-    /** 
-     * Returns available network interfaces. This represents a 
+    /**
+     * Returns available network interfaces. This represents a
```
nyh, line 469: I will accept this change because you didn't do many of them, but in general, please avoid "improving" the style of code which isn't related to your patch together with your actual patch.

tests/unit/socket_test.cc
nyh: Yes, we had a discussion about this problem on the Seastar mailing list last year - see https://groups.google.com/g/seastar-dev/c/lMEXD_M04gE/m/hF41lhtDCAAJ, and the conclusions were that:
1. This issue should be described in the woefully incomplete tutorial.md. Everybody who tried to write trivial servers in Seastar will encounter this problem and waste hours understanding what he did wrong - I did too.
2. We have a document doc/network-connection-load-balancing.md
written by @gleb-cloudius which describes load_balancing_algorithm::fixed but is missing the newer set_fixed_cpu() function which is (supposedly?) easier to use.

tests/unit/socket_test.cc
```diff
@@ -258,3 +261,57 @@ SEASTAR_TEST_CASE(socket_connect_abort_test) {
...
+            .so_sndbuf = buf_size,
+            .so_rcvbuf = buf_size
```
nyh, line 275: Nitpick: Why did you decide to check this function with just one parameter (buf_size) used for both sndbuf and rcvbuf? If you had a copy-pasto in the implementation where SNDBUF was taken from the rcvbuf parameter, your test wouldn't catch it.
I think it will be better to pass two options, sndbuf and rcvbuf, to this test function, and pass different sizes to it - not just two identical small and two identical large.

tests/unit/socket_test.cc
```diff
@@ -258,3 +261,57 @@ SEASTAR_TEST_CASE(socket_connect_abort_test) {
...
+    auto [send_small, recv_small] = buf_size(small_size);
+    auto [send_big, recv_big] = buf_size(big_size);
+
+    // Setting socket buffer sizes isn't an exact science: the kernel does
+    // some rounding, and also (currently) doubles the requested size and
+    // also applies so limits. So as a basic check, assert simply that the
+    // explicit small buffer ends up smaller than the explicit big buffer,
+    // and that both results are at least as large as the requested amount.
+    // The latter condition could plausibly fail if the OS clamped the size
+    // at a small amount, but this is unlikely for the chosen buffer sizes.
```
nyh, line 307: Wow, so strange. This bizarre doubling is even documented in socket(7).

nyh (changes requested):
Looks good, I just have a comment about a typo, and one nitpick about the test: I think it's better to test with **different** send and receive buffer sizes, to make sure you don't have copy-pasto bugs in the implementation (e.g., setting send buffer size from the receive buffer parameter).

include/seastar/net/api.hh
```diff
@@ -389,13 +389,28 @@ public:
 
 /// @}
 
+/// Options for creating a listening socket. These options only have an effect
+/// when using the POSIX stack: all options are ignored on the native stack.
```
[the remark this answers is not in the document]
gleb-cloudius: They are waiting to be implemented. Implementing some of them are as simple as implementing UDP stack. Some are more complicated. On a serious node at least proto should be used for sanity check (error out on unsupported one).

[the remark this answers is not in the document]
nyh: In this case, the comment should contain words like "currently", "TODO", "warning" and so on - the way it is phrased today make it sound like this is intentional.

avikivity (approved):

tests/unit/socket_test.cc
travisdowns: @nyh I think part/all of it is because people have certain expectations about what the socket buffer is, but Linux has a more expansive definition. E.g., Linux uses part of the buffer for things other than the data itself, like metadata. So a user who makes a 100KB socket may expect to write 100KB w/o blocking on an empty socket, but Linux only makes 50KB available for that by default (say). By doubling the size, this cancels out with the half taken away for internal purposes, so the expectations are met (though it is perhaps weird that it returns the doubled size in getsockopt calls).

include/seastar/net/api.hh
travisdowns: Fair enough: I don't make manual "improvements" to the style in unrelated changes, but his trimming is done automatically by my editor for edited files. I can turn it off for this project if you've grown fond of your trailing whitespace :).

include/seastar/net/api.hh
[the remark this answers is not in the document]
travisdowns: I wrote this comment as while implementing the option for the POSIX stack I wanted to understand the implication on the native side, and found that we simply drop the options on that side. 

I will make the comment less accepting of the status quo.

tests/unit/socket_test.cc
travisdowns: Good idea, I've updated this to use separate sizes in https://github.com/scylladb/seastar/commit/5b33a115045eeb037ff41db9c22c213e45808508.

include/seastar/net/api.hh
travisdowns: Eagle eye :eagle: ! I've fixed this in push https://github.com/scylladb/seastar/commit/5b33a115045eeb037ff41db9c22c213e45808508.

include/seastar/net/api.hh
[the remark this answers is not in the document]
travisdowns: I've updated the comment to indicate this is not the desired end state in https://github.com/scylladb/seastar/commit/5b33a115045eeb037ff41db9c22c213e45808508.

travisdowns:
@nyh wrote:

> Looks good, I just have a comment about a typo, and one nitpick about the test: I think it's better to test with **different** send and receive buffer sizes, to make sure you don't have copy-pasto bugs in the implementation (e.g., setting send buffer size from the receive buffer parameter).

Thanks for the feedback: I believe I've addressed all of it in force:

https://github.com/scylladb/seastar/commit/5b33a115045eeb037ff41db9c22c213e45808508

I also added a final test case which tests the "nullopt" sizes for both buffers: hard to validate that it DTRT but I check at least that the discovered size falls in the expected range.

tests/unit/socket_test.cc
avikivity: I think there are newer APIs (for CRIU) which round-trip the values.

travisdowns:
@nyh - all feedback addressed, please have a look when you have a chance, thanks!

nyh (approved):

PR #2462, opened 2024-10-01
httpd/file_handler: Always close stream

BenPope:
Always close the stream in `file_interaction_handler::read`, even if there was a failure such as a short read.


avikivity:
Please copy the changelog text into the commit changelog.

PR #2463, opened 2024-10-01
build: add compiler and C++ standard compatibility checks

### LINKED ISSUE #2181: undefined reference to `seastar::logger::failed_to_log
When building a very simple test application

```
# Make sure we use a relatively recent version - 3.27 is Nov 2023
cmake_minimum_required(VERSION 3.27)

project ("Kataklysmos")

find_package (Seastar REQUIRED)

add_executable (my_app
  src/main.cpp)

target_link_libraries (my_app PUBLIC Seastar::seastar)
```

```
#include <seastar/core/seastar.hh>

using namespace seastar;
logger applog("app");

    seastar::app_template app;
    app.run(argc, argv, [] () -> future<> {
        applog.info("Hello world!");
        return make_ready_future<>();
    });
}
```

I get the following linker error:
```
[3/3] Linking CXX executable my_app
FAILED: my_app 
: && /usr/bin/c++ -O3 -DNDEBUG  CMakeFiles/my_app.dir/src/main.cpp.o -o my_app  /data/dev/Kataklysmos/external/seastar/build/release/libseastar.a  /usr/lib/x86_64-linux-gnu/libboost_program_options.so  /usr/lib/x86_64-linux-gnu/libboost_thread.so  /usr/lib/x86_64-linux-gnu/libboost_chrono.so  /usr/lib/x86_64-linux-gnu/libboost_date_time.so  /usr/lib/x86_64-linux-gnu/libboost_atomic.so  /usr/lib/x86_64-linux-gnu/libcares.so  /usr/lib/x86_64-linux-gnu/libfmt.so.9.1.0  /usr/lib/x86_64-linux-gnu/liblz4.so  -ldl  /usr/lib/x86_64-linux-gnu/libgnutls.so  -latomic  /usr/lib/x86_64-linux-gnu/libsctp.so  /usr/lib/x86_64-linux-gnu/libprotobuf.so  /usr/lib/x86_64-linux-gnu/libyaml-cpp.so  /usr/lib/x86_64-linux-gnu/libhwloc.so  /usr/lib/x86_64-linux-gnu/liburing.so  /usr/lib/x86_64-linux-gnu/libnuma.so && :
/usr/bin/ld: CMakeFiles/my_app.dir/src/main.cpp.o: in function `std::_Function_handler<seastar::future<void> (), main::{lambda()#1}>::_M_invoke(std::_Any_data const&) [clone .cold]':
main.cpp:(.text.unlikely+0x40): undefined reference to `seastar::logger::failed_to_log(std::__exception_ptr::exception_ptr, fmt::v9::basic_string_view<char>, seastar::internal::source_location)'
collect2: error: ld returned 1 exit status
```

Seastar was configured with
```
sudo ./install-dependencies.sh
./configure.py --mode=release --enable-dpdk
ninja -C build/release
```
with the Valgrind fix from https://github.com/scylladb/seastar/issues/2180
Tested https://github.com/scylladb/seastar/commit/b74a027c819ee5b9bfcee4f32edf8f37017bba12 and https://github.com/scylladb/seastar/commit/cd8a9133d2c02f63dbd578d882cf7333a427e194

With Clang (Seastar itself build with default GCC) I get the error:
```

/data/dev/Kataklysmos/src/main.cpp:32:21: error: call to consteval function 'seastar::logger::format_info<>::format_info<char[13], 0>' is not a constant expression
        applog.info("Hello world!");
                    ^
/data/dev/Kataklysmos/external/seastar/include/seastar/util/log.hh:141:58: note: non-constexpr function 'current' cannot be used in a constant expression
                           compat::source_location loc = compat::source_location::current()) noexcept
                                                         ^
/data/dev/Kataklysmos/external/seastar/include/seastar/util/source_location-compat.hh:50:21: note: declared here
    source_location current(const char* file = __builtin_FILE(), const char* func = __builtin_FUNCTION(), int line = __builtin_LINE(), int col = 0) noexcept {
                    ^
1 warning and 1 error generated.
```

tchaikov:
we install `SeastarConfig.cmake` so that a Seastar project can consume the Seastar libraries using `find_package(Seastar)`. this CMake config file checks for the dependencies used by CMake, and defines the Seastar libraries to be consumed by the parent project. one of the dependence is detected using FindSourceLocation.cmake, which is also shipped along with Seastar, this find module is provided so that we can detect the C++ compiler support of `std::source_location` related features. But the result of the detection depends on the C++ standard used, which is default to the one configured by the used C++ compiler if not specified. if the parent project consumes precompiled Seastar libraries, and use a differnt C++ standard, the parent project could fail to build, because Seastar and parent project are using different compiling options.

mismatches in C++ standards or compilers between Seastar and the parent project can lead to build failures or runtime issues. while not always problematic, these mismatches can cause unexpected behavior.

in this change, we detect

- the C++ standards used by the parent project and Seastar
- the C++ compiler used by the parent project and Seastar

warnings are non-fatal as compatibility is possible in some cases, but developers should be aware of potential issues and test thoroughly.

Fixes #2181

### DIFF
--- a/cmake/SeastarConfig.cmake.in
+++ b/cmake/SeastarConfig.cmake.in
@@ -25,6 +25,32 @@
 
 list (APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
 
+if (CMAKE_CXX_STANDARD)
+  set (current_cxx_standard ${CMAKE_CXX_STANDARD})
+else ()
+  set (current_cxx_standard ${CMAKE_CXX_STANDARD_DEFAULT})
+endif ()
+
+if (NOT (current_cxx_standard STREQUAL @CMAKE_CXX_STANDARD@))
+  message(WARNING
+    "C++ Standard mismatch detected:
+- Seastar was compiled with: C++@CMAKE_CXX_STANDARD@
+- This project is configured to use: C++${current_cxx_standard}
+This mismatch may lead to build failures due to differences in the supported \
+features of these two standards. Please adjust your project's C++ standard to \
+match Seastar.")
+endif ()
+
+if (NOT ((CMAKE_CXX_COMPILER_ID STREQUAL @CMAKE_CXX_COMPILER_ID@) AND
+         (CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL @CMAKE_CXX_COMPILER_VERSION@)))
+  message (WARNING
+    "Compiler mismatch detected:
+- Seastar was compiled with: @CMAKE_CXX_COMPILER_ID@ @CMAKE_CXX_COMPILER_VERSION@ \
+- This project is configured to use: ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION}
+This mismatch may lead to build failures due to differences in supported features and ABI compatibility. \
+Please configure your project to use the same C++ compiler to match Seastar.")
+endif ()
+
 #
 # Dependencies.
 #


tchaikov:
@avikivity Hi Avi, what this changes is the CMake config file. It is
shipped/installed by a pre-compiled Seastar library. We cannot predict what
compiler or C++ standard used by the parent project when compiling Seastar.

Regards
Kefu Chai


Avi Kivity ***@***.***>于2024年10月1日 周二19:47写道：

PR #2464, opened 2024-10-01
Add new http response codes

kreuzerkrieg:
Add new http response codes to satisfy scylladb needs for S3 error handling

fixes: #2466

### DIFF
--- a/include/seastar/http/reply.hh
+++ b/include/seastar/http/reply.hh
@@ -94,16 +94,21 @@ struct reply {
         uri_too_long = 414, //!< uri_too_long
         unsupported_media_type = 415, //!< unsupported_media_type
         expectation_failed = 417, //!< expectation_failed
+        page_expired = 419, //!< page_expired
         unprocessable_entity = 422, //!< unprocessable_entity
         upgrade_required = 426, //!< upgrade_required
         too_many_requests = 429, //!< too_many_requests
+        login_timeout = 440, //!< login_timeout
         internal_server_error = 500, //!< internal_server_error
         not_implemented = 501, //!< not_implemented
         bad_gateway = 502, //!< bad_gateway
         service_unavailable = 503,  //!< service_unavailable
         gateway_timeout = 504, //!< gateway_timeout
         http_version_not_supported = 505, //!< http_version_not_supported 
-        insufficient_storage = 507 //!< insufficient_storage
+        insufficient_storage = 507, //!< insufficient_storage
+        bandwidth_limit_exceeded = 509, //!< bandwidth_limit_exceeded
+        network_read_timeout = 598, //!< network_read_timeout
+        network_connect_timeout = 599, //!< network_connect_timeout
     } _status;
 
     /**
--- a/include/seastar/http/reply.hh
+++ b/include/seastar/http/reply.hh
@@ -187,7 +187,7 @@ struct reply {
         _response_line = response_line();
         return *this;
     }
-    sstring response_line();
+    sstring response_line() const;
 
     /*!
      * \brief use an output stream to write the message body
--- a/src/http/reply.cc
+++ b/src/http/reply.cc
@@ -186,7 +186,7 @@ std::ostream& operator<<(std::ostream& os, reply::status_type st) {
 }
 
 sstring reply::response_line() {
-    return "HTTP/" + _version + " " + status_strings::to_string(_status) + "\r\n";
+    return seastar::format("HTTP/{} {}\r\n", _version, status_strings::to_string(_status));
 }
 
 void reply::write_body(const sstring& content_type, noncopyable_function<future<>(output_stream<char>&&)>&& body_writer) {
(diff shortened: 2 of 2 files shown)


include/seastar/http/reply.hh
```diff
@@ -187,7 +192,7 @@ struct reply {
         _response_line = response_line();
         return *this;
     }
-    sstring response_line();
+    sstring response_line() const;
```
xemul, line 195: Separate patch

src/http/reply.cc
```diff
@@ -49,144 +50,68 @@ namespace http {
 
 namespace status_strings {
 
-const sstring continue_ = "100 Continue";
```
xemul, line 52: This may break existing 3rd party project compilation

src/http/reply.cc
```diff
@@ -49,144 +50,68 @@ namespace http {
...
+  if (auto found = status_strings.find(status); found != status_strings.end()) [[likely]]
+    return found->second;
+  return status_strings.at(reply::status_type::internal_server_error);
 }
+
 } // namespace status_strings
 
 std::ostream& operator<<(std::ostream& os, reply::status_type st) {
-    return os << status_strings::to_string(st);
+    return os << status_strings::to_string_view(st);
```
xemul, line 110: Separate patch

src/http/client.cc
```diff
@@ -346,31 +346,57 @@ future<> client::make_request(request req, reply_handler handle, abort_source& a
...
+                           return do_make_request(con, req, handle, as, expected);
+                       },as).then([&retries]() -> future<std::optional<uint32_t>> {
+                               return make_ready_future<std::optional<uint32_t>>(retries);
+                       }).handle_exception_type([&retries, &e, as](const std::system_error& ex) {
+                           e = ex;
+                           // TODO: add some logic for issuing a new request, randomization? exponential backoff? a mix of these two?
+                           return handle_request_exception(retries, as, ex);
+                       });
+                   }
+                   if (retries <= _max_retries) {
```
xemul, line 376: This change needs justification. The point of "retry request" PR was that if request fails due to transport error, retrying it more than once over newly established connection makes little sense

src/http/client.cc
```diff
@@ -346,31 +346,57 @@ future<> client::make_request(request req, reply_handler handle, abort_source& a
...
+        return make_exception_future<std::optional<uint32_t>>(ex);
+    }
+    return make_ready_future<std::optional<uint32_t>>(std::nullopt);
+}
 
-            auto code = ex.code().value();
-            if ((code != EPIPE) && (code != ECONNABORTED)) {
-                return make_exception_future<>(ex);
-            }
+future<> client::make_raw_request(request& req, reply_handler& handle, abort_source* as, std::optional<reply::status_type> expected) {
```
xemul, line 362: If we need two make_request-s -- one that takes ownership and the other one that doesn't -- I'd suggest two overloads:

```c++
future<> make_request(request&& req, reply_handler&& handle, ...); // takes ownership
future<> make_request(request& req, reply_handler& handle, ...); // doesn't take ownership
```

src/http/reply.cc
```diff
@@ -49,144 +50,68 @@ namespace http {
...
+    {reply::status_type::bad_gateway, "502 Bad Gateway"},
+    {reply::status_type::service_unavailable, "503 Service Unavailable"},
+    {reply::status_type::gateway_timeout, "504 Gateway Timeout"},
+    {reply::status_type::http_version_not_supported, "505 HTTP Version Not Supported"},
+    {reply::status_type::insufficient_storage, "507 Insufficient Storage"},
+    {reply::status_type::bandwidth_limit_exceeded, "509 Bandwidth Limit Exceeded"},
+    {reply::status_type::network_read_timeout, "598 Network Read Timeout"},
+    {reply::status_type::network_connect_timeout, "599 Network Connect Timeout"}};
+
+static const std::string_view& to_string_view(reply::status_type status) {
```
tchaikov, line 101: // since Pavel is reviewing.. 

how about defining a `fmt::formatter` for `status_type`?

src/http/reply.cc
kreuzerkrieg: How come? These constants are local to this compilation unit

src/http/reply.cc
kreuzerkrieg: Dunno... Who gonna use it? All this string manipulation is local to this file only, no one never used it outside.

src/http/reply.cc
xemul: Ah, it's .cc file, indeed. Disregard this

include/seastar/http/reply.hh
kreuzerkrieg: Moved the retry to another PR

src/http/client.cc
kreuzerkrieg: Ok, there was a misunderstanding of what have to be done there, reverting

src/http/client.cc
kreuzerkrieg: Will do in a separate PR

Issue #2467, opened 2024-10-02
Add non-owning `make_request` to http client


kreuzerkrieg:
To satisfy scylla's s3 client with future retry implementation add non-owning `make_request` to http client which is similar to the existing `make_request` but it doesn't take ownership of `request` and `reply_handler`

PR #2468, opened 2024-10-02
http: Add non-owning `make_request` to http client

kreuzerkrieg:
To satisfy scylla's s3 client with future retry implementation add non-owning `make_request` to http client which is similar to the existing `make_request` but it doesn't take ownership of `request` and `reply_handler`

fixes: #2467

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -199,7 +199,7 @@ private:
     requires std::invocable<Fn, connection&>
     auto with_new_connection(Fn&& fn, abort_source*);
 
-    future<> do_make_request(request req, reply_handler handle, abort_source*, std::optional<reply::status_type> expected);
+    future<> do_make_request(request& req, reply_handler& handle, abort_source*, std::optional<reply::status_type> expected);
     future<> do_make_request(connection& con, request& req, reply_handler& handle, abort_source*, std::optional<reply::status_type> expected);
 
 public:
@@ -283,7 +283,7 @@ public:
      * client may restart the whole request processing in case server closes the connection
      * in the middle of operation
      */
-    future<> make_request(request req, reply_handler handle, std::optional<reply::status_type> expected = std::nullopt);
+    future<> make_request(request&& req, reply_handler&& handle, std::optional<reply::status_type>&& expected = std::nullopt);
 
     /**
      * \brief Send the request and handle the response (abortable)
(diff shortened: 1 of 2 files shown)


src/http/client.cc
```diff
@@ -338,39 +338,45 @@ auto client::with_new_connection(Fn&& fn, abort_source* as) {
...
 
-            auto code = ex.code().value();
-            if ((code != EPIPE) && (code != ECONNABORTED)) {
-                return make_exception_future<>(ex);
-            }
+        if (!_retry) {
+            return make_exception_future<>(ex);
+        }
 
-            // The 'con' connection may not yet be freed, so the total connection
```
xemul, line 367: There's a lot of indentation fixes here. When this happens, we ask to split the patch into two -- first, that does what it needs, but doesn't re-indent unaffected lines. The second patch restores the indentation. Please, do

include/seastar/http/client.hh
```diff
@@ -295,7 +295,16 @@ public:
...
+    future<> make_request(request&& req, reply_handler&& handle, abort_source& as, std::optional<reply::status_type>&& expected = std::nullopt);
+
+    /**
+     * \brief Send the request and handle the response, same as \ref make_request()
+     *
+     *  @attention Note that the method does not take the ownership of the
+     * `request and the `handle`, it caller's responsibility the make sure they
+     * are referencing valid instances
+     */
+    future<> make_request(request& req, reply_handler& handle, abort_source& as, std::optional<reply::status_type> expected = std::nullopt);
```
xemul, line 307: How about non-owning make_request that doesn't take abort-source argument?

src/http/client.cc
kreuzerkrieg: Sure. Done

include/seastar/http/client.hh
kreuzerkrieg: Added

include/seastar/http/client.hh
```diff
@@ -295,7 +295,18 @@ public:
...
+    /**
+     * \brief Send the request and handle the response, same as \ref make_request()
+     *
+     *  @attention Note that the method does not take the ownership of the
+     * `request and the `handle`, it caller's responsibility the make sure they
+     * are referencing valid instances
+     */
+    future<> make_request(request& req, reply_handler& handle, std::optional<reply::status_type> expected = std::nullopt);
+
+    future<> make_request(request& req, reply_handler& handle, abort_source& as, std::optional<reply::status_type> expected = std::nullopt);
```
xemul, line 309: This one should be documented as well

src/http/client.cc
kreuzerkrieg: agggr, squashed, lemme fix that

kreuzerkrieg: done

include/seastar/http/client.hh
kreuzerkrieg: Done

PR #2470, opened 2024-10-04
sharded.hh: seperate invoke_on decls from defs

tomershafir:
Seperate all invoke_on decls from defs on sharded class template for improved readability and loose coupling, maybe even as preparation for `extern template`. Also, cluster `invoke_on(unsigned id, ...)` together with the other variants.

### DIFF
--- a/include/seastar/core/sharded.hh
+++ b/include/seastar/core/sharded.hh
@@ -267,13 +267,7 @@ public:
     /// the message.
     /// Passes the default \ref smp_submit_to_options to the
     /// \ref smp::submit_to() called behind the scenes.
-    future<> invoke_on_all(std::function<future<> (Service&)> func) noexcept {
-      try {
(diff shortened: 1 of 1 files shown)


include/seastar/core/sharded.hh
```diff
@@ -784,6 +759,7 @@ sharded<Service>::stop() noexcept {
 }
 
 template <typename Service>
+inline
```
avikivity, line 762: inlines aren't necessary for templates, and are a net negative for this compile time oriented function (though in practice, the compiler ignores the inline directive for inlining).

tomershafir: right, fixing

tomershafir: will remove inline from templates in sharded.hh on a different PR

tomershafir:
@avikivity ping

PR #2475, opened 2024-10-05
loop: add Sentinel iterator support to parallel_for_each()

tchaikov:
before this change, we only support containers, whose `begin()` and `end()` are of the same type, but ranges do not have this requirement.

in order to support range views, let's relax this requirement. so that, we can pass two iterators denoting a range to `parallel_for_each()`. see also b1b0bff1b8b3a94792f7a37b203ae749d54a1d83.

### DIFF
--- a/include/seastar/core/loop.hh
+++ b/include/seastar/core/loop.hh
@@ -381,15 +381,15 @@ future<> keep_doing(AsyncAction action) noexcept {
 }
 
 namespace internal {
-template <typename Iterator, typename AsyncAction>
+template <typename Iterator, class Sentinel, typename AsyncAction>
 class do_for_each_state final : public continuation_base<> {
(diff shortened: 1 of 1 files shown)


dawmd (approved):
LGTM

tchaikov:
i understand we are in favor of C++20 coroutine variants but when working on d94de9a72fa759eb33832e0552c5f7b48e805450, i found that the missing support to range of `parallel_for_each()` prevented me from migrating one `boost::irange()` to `std::views::iota()`. so i think, for some reason, we (scylladb) and seastar still have some existing callers of the continuation based `parallel_for_each()` . so before we migrate them to C++20 coroutine, probably we can switch them to C++20 range first.

tchaikov:

PR #2480, opened 2024-10-09
build: improve c-ares version extraction method

tchaikov:
this change enhances the version extraction process to handle both literal and macro-based version definitions in c-ares.

in 5d53fb6d, we extract the c-ares version by looking for `ARES_VERSION_STR` and matching the version numbers in it. as, in c-ares 1.32, the line looks like:
```c
  #define ARES_VERSION_STR "1.33.0"
```
but this method fails for c-ares 1.33+ due to non-literal `ARES_VERSION_STR` definitionin. as in c-ares 1.33+, it looks like
```c
  #define ARES_VERSION_STR             \
    ARES_STRINGIFY(ARES_VERSION_MAJOR) \
    "." ARES_STRINGIFY(ARES_VERSION_MINOR) "." ARES_STRINGIFY(ARES_VERSION_PATCH)
```

this fails the parser. and the version number is always empty.

in this change, we match the MAJOR, MINOR and PATCH version components separately. this approach increases resilience. after this change, Seastar is able to detect the versions of both c-ares 1.32 and 1.33.

Refs 5d53fb6d

### DIFF
--- a/cmake/Findc-ares.cmake
+++ b/cmake/Findc-ares.cmake
@@ -37,15 +37,15 @@ find_path (c-ares_INCLUDE_DIR
     ${PC_c-ares_INCLUDE_DIRS})
 
 if (c-ares_INCLUDE_DIR)
-  file(STRINGS "${c-ares_INCLUDE_DIR}/ares_version.h" ares_VERSION_LINE
-    REGEX "^#define[ \t]+ARES_VERSION_STR \"[^\"]*\"$")
-  if (ares_VERSION_LINE MATCHES "ARES_VERSION_STR \"(([0-9]+)\\.([0-9]+)\\.([0-9]+))")
-    set (c-ares_VERSION "${CMAKE_MATCH_1}")
-    set (c-ares_VERSION_MAJOR "${CMAKE_MATCH_2}")
-    set (c-ares_VERSION_MINOR "${CMAKE_MATCH_3}")
-    set (c-ares_VERSION_PATCH "${CMAKE_MATCH_4}")
-  endif ()
-  unset (ares_VERSION_LINE)
+  foreach (v MAJOR MINOR PATCH)
+    file(STRINGS "${c-ares_INCLUDE_DIR}/ares_version.h" ares_VERSION_LINE
+      REGEX "^#define[ \t]+ARES_VERSION_${v}[ \t]+[0-9]+$")
+    if (ares_VERSION_LINE MATCHES "ARES_VERSION_${v} ([0-9]+)")
+      set (c-ares_VERSION_${v} "${CMAKE_MATCH_1}")
+    endif ()
+    unset (ares_VERSION_LINE)
+  endforeach ()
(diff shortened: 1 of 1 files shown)


bradh352:
Out of curiosity, what happens if you just remove your Findc-ares.cmake completely?

c-ares distributes both package config and cmake files when it is installed.  And the default cmake FindPackage(c-ares) should automatically set  c_ares_VERSION (as well as c_ares_VERSION_MAJOR, c_ares_VERSION_MINOR, and c_ares_VERSION_PATCH).

tchaikov:
> Out of curiosity, what happens if you just remove your Findc-ares.cmake completely?

@bradh352 hi Brad, it depends. on my fedora 41 box, if i remove this file, the bug goes away, as the maintainer packages the CMake config files in the prebuilt package:
```console
$ rpm -ql c-ares-devel | grep cmake
/usr/lib64/cmake/c-ares
/usr/lib64/cmake/c-ares/c-ares-config-version.cmake
/usr/lib64/cmake/c-ares/c-ares-config.cmake
/usr/lib64/cmake/c-ares/c-ares-targets-noconfig.cmake
/usr/lib64/cmake/c-ares/c-ares-targets.cmake
```

but unfortunately, the latest debian stable release (bookworm) still does not package them, see https://debian.pkgs.org/12/debian-main-amd64/libc-ares-dev_1.18.1-3_amd64.deb.html .

> 
> c-ares distributes both package config and cmake files when it is installed. And the default cmake FindPackage(c-ares) should automatically set c_ares_VERSION (as well as c_ares_VERSION_MAJOR, c_ares_VERSION_MINOR, and c_ares_VERSION_PATCH).

thank you for the excellent CMake integration. but before the popular distros include these files, i think we should at least fall back to searching c-ares in the module mode. even if we can use these two modes when searching for c-ares, i still want to avoid the workaround like https://github.com/scylladb/seastar/blob/54e25a9d4a285c9489d1a7aeb4c6b8c10d4af4f3/cmake/SeastarDependencies.cmake#L154-L168 . i think the preferred way to find a package is just to use `find_package()` without worrying about which mode it uses.

so, IMHO, a simpler approach before debian stable's libc-ares-dev includes the CMake config files is still to use the module mode.

thank you for pointing this out. i created #2485 to track this cleanup.

bradh352:
@gjasny since you maintain the debian packaging of c-ares can you look into distributing the cmake files for c-ares on debian?  I guess they're missing.

tchaikov:
> The version detection should also happen at runtime, not compile time.

hi Gregor , thank you for your comment. my concern is to enable users to build seastar from source on popular distributions, instead of enabling them to use precompiled seastar package like `libseastar-dev`. so probably we don't need to worry about the runtime behaviour. if, by runtime, you meant "building a seastar application" after `libseastar-dev` or seastar is installed.

but in the case above, since `Findc-ares.cmake` is installed along with seastar's cmake config files. i think we should be able to detect new or old c-ares installations.

> 
> At least in Debian your scylladb package won’t get recompiled for a new c-ares package and also older c-ares Debian packages might satisfy the deb package dependency.
> 
> Edit: if it’s only the missing header that blocks support, compile time looks right.

tchaikov:

PR #2484, opened 2024-10-09
scripts: add pull_github_pr.sh

avikivity:
Add a script that, given a pull request number, will merge it into the current branch, and add the pull request cover letter as a merge commit. This avoids githubs non-informative merge commits.

If the series contains a single patch, the patch is cherry-picked and the cover letter is discarded.

Adapted from scylladb.

### DIFF
--- a/scripts/pull_github_pr.sh
+++ b/scripts/pull_github_pr.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+# Script for pulling a github pull request
+# along with generating a merge commit message.
+# Example usage for pull request #6007:
+#
+# git checkout master
+# git pull
+# ./scripts/pull_github_pr.sh 6007
+
+set -e
+
+gh_hosts=~/.config/gh/hosts.yml
+FORCE="$2"
+
+if [[ ( -z "$GITHUB_LOGIN" || -z "$GITHUB_TOKEN" ) && -f "$gh_hosts" ]]; then
+    GITHUB_LOGIN=$(awk '/user:/ { print $2 }' "$gh_hosts")
+    GITHUB_TOKEN=$(awk '/oauth_token:/ { print $2 }' "$gh_hosts")
+fi
+
+if [[ -z "$GITHUB_LOGIN" || -z "$GITHUB_TOKEN" ]]; then
+    echo 'Please set $GITHUB_LOGIN and $GITHUB_TOKEN'
+    exit 1
+fi
+
+if [[ -z "$1" ]]; then
+    echo Please provide a github pull request number
+    exit 1
+fi
+
+for required in jq curl; do
+    if ! type $required >& /dev/null; then
+        echo Please install $required first
+        exit 1
+    fi
+done
+
+do_curl() {
+    curl --user "${GITHUB_LOGIN}:${GITHUB_TOKEN}" "$@"
+}
+
+
+NL=$'\n'
+
+PR_NUM="$1"
+# convert full repo URL to its project/repo part, in case of failure default to origin/master:
+REMOTE_SLASH_BRANCH="$(git rev-parse --abbrev-ref --symbolic-full-name @{upstream} \
+     || git rev-parse --abbrev-ref --symbolic-full-name master@{upstream} \
+     || echo 'origin/master')"
+REMOTE="${REMOTE_SLASH_BRANCH%/*}"
+REMOTE_URL="$(git config --get "remote.$REMOTE.url")"
+PROJECT=`sed 's/git@github.com://;s#https://github.com/##;s/\.git$//;' <<<"${REMOTE_URL}"`
+PR_PREFIX=https://api.github.com/repos/$PROJECT/pulls
+
+echo "Fetching info on PR #$PR_NUM..."
+PR_DATA="$(do_curl -s "$PR_PREFIX/$PR_NUM")"
+MESSAGE="$(jq -r .message <<< "$PR_DATA")"
+if [[ "$MESSAGE" != null ]]; then
+    # Error message, probably "Not Found".
+    echo "$MESSAGE"
+    exit 1
+fi
+PR_TITLE="$(jq -r .title <<< "$PR_DATA")"
+echo "    $PR_TITLE"
+PR_DESCR="$(jq -r .body <<< "$PR_DATA")"
+PR_LOGIN="$(jq -r .head.user.login <<< "$PR_DATA")"
+echo -n "Fetching full name of author $PR_LOGIN... "
+USER_NAME="$(do_curl -s "https://api.github.com/users/$PR_LOGIN" | jq -r .name)"
+echo "$USER_NAME"
+
+git fetch "$REMOTE" "pull/$PR_NUM/head"
+
+nr_commits=$(git log --pretty=oneline HEAD..FETCH_HEAD | wc -l)
+
+closes="${NL}${NL}Closes ${PROJECT}#${PR_NUM}${NL}"
+
+if [[ $nr_commits == 1 ]]; then
+    commit=$(git log --pretty=oneline HEAD..FETCH_HEAD | awk '{print $1}')
+    message="$(git log -1 "$commit" --format="format:%s%n%n%b")"
+    if ! git cherry-pick "$commit"; then
+        echo "Cherry-pick failed. You are now in a subshell. Either resolve with git cherry-pick --continue or git cherry-pick --abort, then exit the subshell"
+        head_before="$(git rev-parse HEAD)"
+        bash
+        head_after="$(git rev-parse HEAD)"
+        if [[ "$head_before" = "$head_after" ]]; then
+            exit 1
+        fi
+    fi
+    git commit --amend -m "${message}${closes}"
+else
+    git merge --no-ff --log=1000 FETCH_HEAD -m "Merge '$PR_TITLE' from $USER_NAME" -m "${PR_DESCR}${closes}"
+fi
+git commit --amend # for a manual double-check


scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+for required in jq curl; do
+	if ! type $required >& /dev/null; then
+		echo Please install $required first
+		exit 1
+	fi
+done
+
+curl() {
+    local opts=()
+    if [[ -n "$GITHUB_LOGIN" && -n "$GITHUB_TOKEN" ]]; then
```
tchaikov, line 43: `GITHUB_LOGIN` and `GITHUB_TOKEN` are always set, so we don't need to check them here. or, we can do something like:
```
curl() {
    if [[ ( -z "$GITHUB_LOGIN" || -z "$GITHUB_TOKEN" ) && -f "$gh_hosts" ]]; then
        GITHUB_LOGIN=$(awk '/user:/ { print $2 }' "$gh_hosts")
        GITHUB_TOKEN=$(awk '/oauth_token:/ { print $2 }' "$gh_hosts")
     fi

     if [[ -z "$GITHUB_LOGIN" || -z "$GITHUB_TOKEN" ]]; then
        echo 'Please set $GITHUB_LOGIN and $GITHUB_TOKEN'
        exit 1
    fi

    command curl --user "${GITHUB_LOGIN}:${GITHUB_TOKEN}" "$@"
}
```

as these two variables are "closer" to this function, and we only set them once they are set by this function.

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+	GITHUB_TOKEN=$(awk '/oauth_token:/ { print $2 }' "$gh_hosts")
+fi
+
+if [[ -z "$GITHUB_LOGIN" || -z "$GITHUB_TOKEN" ]]; then
+    echo 'Please set $GITHUB_LOGIN and $GITHUB_TOKEN'
+    exit 1
+fi
+
+if [ -z "$1" ]; then
+	echo Please provide a github pull request number
```
tchaikov, line 30: can we only use spaces instead of using both tabs and spaces for indent?

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+    fi
+    command curl "${opts[@]}" "$@"
+}
+
+
+NL=$'\n'
+
+PR_NUM=$1
+# convert full repo URL to its project/repo part, in case of failure default to origin/master:
+REMOTE_SLASH_BRANCH="$(git rev-parse --abbrev-ref --symbolic-full-name  @{upstream} \
```
tchaikov, line 54: extraneous space after `name`.

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+     || git rev-parse --abbrev-ref --symbolic-full-name master@{upstream} \
+     || echo 'origin/master')"
+REMOTE="${REMOTE_SLASH_BRANCH%/*}"
+REMOTE_URL="$(git config --get "remote.$REMOTE.url")"
+PROJECT=`sed 's/git@github.com://;s#https://github.com/##;s/\.git$//;' <<<"${REMOTE_URL}"`
+PR_PREFIX=https://api.github.com/repos/$PROJECT/pulls
+
+echo "Fetching info on PR #$PR_NUM... "
```
tchaikov, line 62: extraneous space after `...`.

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+PR_DATA=$(curl -s $PR_PREFIX/$PR_NUM)
+MESSAGE=$(jq -r .message <<< $PR_DATA)
+if [ "$MESSAGE" != null ]
```
tchaikov, line 65: to be more consistent with the rest of this script, how about moving "next" to the previous line (and add a `;`)?

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+
+nr_commits=$(git log --pretty=oneline HEAD..FETCH_HEAD | wc -l)
+
+closes="${NL}${NL}Closes ${PROJECT}#${PR_NUM}${NL}"
+
+if [[ $nr_commits == 1 ]]; then
+	commit=$(git log --pretty=oneline HEAD..FETCH_HEAD | awk '{print $1}')
+	message="$(git log -1 "$commit" --format="format:%s%n%n%b")"
+	if ! git cherry-pick $commit
+	then
```
tchaikov, line 89: ditto.

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+# Example usage for pull request #6007:
+#
+# git checkout master
+# git pull
+# ./scripts/pull_github_pr.sh 6007
+
+set -e
+
+gh_hosts=~/.config/gh/hosts.yml
+jenkins_url="https://jenkins.scylladb.com"
```
tchaikov, line 14: this variable is not used.

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+FORCE=$2
```
tchaikov, line 15: this variable is not used.

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+ORANGE='\033[0;33m'
```
tchaikov, line 16: this variable is not used.

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+NC='\033[0m'
```
tchaikov, line 17: this variable is not used.

nyh:
> Add a script that, given a pull request number, will merge it into the current branch, and add the pull request cover letter as a merge commit. This avoids githubs non-informative merge commits.
> 
> If the series contains a single patch, the patch is cherry-picked and the cover letter is discarded.
> 
> Adapted from scylladb.

Good idea. I actually copied the version of the script from Scylla before we started to add strange Scylla-CI-specific code to it, and have been using it in Seastar (and other projects) ever since, and it works very nicely :-)

scripts/pull_github_pr.sh
nyh: I agree that there's no need to check these variables again (we checked above that they must be set, if they aren't, the script exited). I also agree that it's nicer to just use crystal-clear shell syntax like you suggested instead of the bizarre arrays used by Avi's code.
But I wouldn't move the check for these variables into the function - I think it's better to check them up-front in the beginning of the script, like Avi did.

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+fi
+
+for required in jq curl; do
+	if ! type $required >& /dev/null; then
+		echo Please install $required first
+		exit 1
+	fi
+done
+
+curl() {
```
nyh, line 41: Please consider renaming this function. I find it confusing that you take a command that has an existing meaning, and redefine it. The fact that you had to use the obscure builtin "command" is a tell-tale sign it's confusing.

scripts/pull_github_pr.sh
nyh: Actually, I would do the exact opposite. Having the then on a new line, and no ";", is the more traditional and better looking, syntax.

scripts/pull_github_pr.sh
nyh: Or maybe he wanted "echo -n"?

nyh (changes requested):
Thanks. Looks good. Joining @tchaikov I only had a bunch of nitpicks.

scripts/pull_github_pr.sh
```diff
@@ -0,0 +1,102 @@
...
+# along with generating a merge commit message.
+# Example usage for pull request #6007:
+#
+# git checkout master
+# git pull
+# ./scripts/pull_github_pr.sh 6007
+
+set -e
+
+gh_hosts=~/.config/gh/hosts.yml
```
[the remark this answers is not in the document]
avikivity: It comes from the `gh` application, that I only used in order to establish login credentials once.

travisdowns:
@avikivity if you enable this setting (first circle) in the "General" section in [settings](https://github.com/scylladb/seastar/settings):

image

Then the merge message will include the cover letter.

Furthermore if you allow the second setting, then there will be a rebase button allowing you to rebase with 1 click (if it rebases cleanly).

The combination would allow merging in your preferred style right from github, and also on any device (phone, so you can merge while taking care of other business, etc). It also lets you confirm that CI was green after the rebase.

Finally, if you allow the "auto-merge" it can make this workflow a bit smoother since you can do the rebase, enable auto-merge for the change which would then merge when the CI is green (need to double check the rules on that one, since I don't think green CI is currently a requirement).

travisdowns:
Actually, it seems like that may already be enabled? From a recent PR, when I clicked "merge pull request":

image

So it does populate the merge commit with the cover letter info.

travisdowns:
Merging it like (with the botton, which I did) that doesn't include "merge ... by Avi ...." in the commit message, which makes it less clear it's a merge commit and makes attribution less clear:

https://github.com/scylladb/seastar/commit/65f7fea0971698ba7074b8819c6a8aeba2131890

avikivity:
The attribution is still there for the individual patches.

I really dislike the committer attribution for that patch:

```
commit 65f7fea0971698ba7074b8819c6a8aeba2131890 (HEAD -> master, origin/master, origin/HEAD)
Merge: 0386af3c54 04657f03d6
Author:     Travis Downs <travis.downs@redpanda.com>
AuthorDate: Sat Jan 24 11:47:11 2026 -0300
Commit:     GitHub <noreply@github.com>
CommitDate: Sat Jan 24 11:47:11 2026 -0300

    build: support sccache as a compiler cache (#3205)
    
    Support sccache [1] as an alternative to ccache. Since it
    is better than ccache, and unlikely to be randomly present
    on some machine, prefer it to ccache.
    
    To avoid caching by both sccache and ccache, skip the
    masquerade directory provided by some ccache installations.
    
    [1] https://github.com/mozilla/sccache
```

A person committed it, not github.

travisdowns:
> A person committed it, not github.

Yes, and addition authorship is attributed to me, I guess as the person who clicks the button is afforded the option to edit the commit message, but I accepted the default so it entirely your text.

travisdowns:
> A person committed it, not github.

The reasoning I found is that the commit is created by GH on their backend (and also signed by GH, which is why it shows up as "verified".

In any case, I did look into it more and this behavior cannot be changed, so if this is not acceptable, using the script is the only way for multi-patch series.

travisdowns:
@avikivity are you going to address the feedback from nyh and tchaikov? Would be nice to get this merged.

I can take it over if you don't have time.

travisdowns:
This script sets both author and committer of the merge commit to the person running the script, I'm not sure if that is desirable. IMO the "author" should be the person who authored the cover letter, which is the PR author.

avikivity:
> @avikivity are you going to address the feedback from nyh and tchaikov? Would be nice to get this merged.
> 
> I can take it over if you don't have time.

I'll revamp it. I got scared by the mass of comments.

avikivity:
v2:
 - remove unused variables
 - improved shell quoting
 - use [[ ]] consistently in conditionals
 - renamed curl() function to do_curl() 
 - dropped unnecessary double checks for github auth
 - whitespace fixes
 - more consistent if/then formatting
 - rrrrrrrrrrrrebased

avikivity:
> This script sets both author and committer of the merge commit to the person running the script, I'm not sure if that is desirable. IMO the "author" should be the person who authored the cover letter, which is the PR author.

Well, the committer authored the merge, including fixing up any merge conflicts.

That's the git behavior without the script as well.

travisdowns:
@avikivity wrote:

> Well, the committer authored the merge, 

Well what do you mean by "authored"? A merge commit itself has no content except for the metadata about the parent changes and the commit message. So the commit message is the only content which has been "authored" in my view and with this script this is extracted from the PR cover letter, i.e., the original author wrote it.

> including fixing up any merge conflicts.

That's reflected in updated committer for those commits. Though personally I would not fix conflicts except in unusual cases: I would ask the original author to rebase, or is that not the policy here?

> That's the git behavior without the script as well.

Yes, but the difference is in git if I merge a commit, the editor opens up for me (the merger) to write the merge commit, so I am the author there: maybe I just except the default git 1-liner or I write something else.

This script extracts content authored by someone else and inserts it as the commit message (which is good), but arguably the author should be set correctly: to the author of the extract content.

That said there's no absolute hard rule or 100% objective answer here, so I'm OK as is: just wanted to point this out.

travisdowns:
@nyh I think your requested changes have been addresses by Avi's v2, could you re-review or at least remove the changes requested flag so others can review & merge?

avikivity:
Sorry, lost the thread.

> @avikivity wrote:
> 
> > Well, the committer authored the merge,
> 
> Well what do you mean by "authored"? A merge commit itself has no content except for the metadata about the parent changes and the commit message. So the commit message is the only content which has been "authored" in my view and with this script this is extracted from the PR cover letter, i.e., the original author wrote it.

The committer had an input here. But perhaps that's best represented by the committer field, and that's what we do for single-patch pull requests.

So I'm open to reconsidering.

Issues:
 - I'm no so comfortable with a script impersonating someone else (but can talk myself into accepting)
 - What to do if github does not provide a reasonable username/email? (not uncommon unfortunately)
 - What if there are multiple authors? Have seen some cases (which is good). I guess use the submitter's account.

> 
> > including fixing up any merge conflicts.
> 
> That's reflected in updated committer for those commits. 

Yes.

> Though personally I would not fix conflicts except in unusual cases: I would ask the original author to rebase, or is that not the policy here?

It's left for personal preference. I don't mind fixing up #include clashes. Or simple things like changes in lines close by that don't actually clash. But I understand people not wanting to deal with it.

> 
> > That's the git behavior without the script as well.
> 
> Yes, but the difference is in git if I merge a commit, the editor opens up for me (the merger) to write the merge commit, so I am the author there: maybe I just except the default git 1-liner or I write something else.
> 
> This script extracts content authored by someone else and inserts it as the commit message (which is good), but arguably the author should be set correctly: to the author of the extract content.
> 
> That said there's no absolute hard rule or 100% objective answer here, so I'm OK as is: just wanted to point this out.

I think there is merit to it. It would also eliminate the awkward "from $author".

PR #2488, opened 2024-10-11
build: set CMAKE_POSITION_INDEPENDENT_CODE when building shared libs

tchaikov:
in 54e25a9d, we enabled Seastar to build shared and static libraries in a single pass. but we failed to set `POSITION_INDEPENDENT_CODE` property for the targets if `BUILD_SHARED_LIBS` is not set. this breaks the build if user does't pass `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` to `cmake` when generating the building system of seastar.

in this change:

- we set the global `CMAKE_POSITION_INDEPENDENT_CODE` as long as `Seastar_BUILD_STATIC_LIBS` is set. please note, this does not incur the runtime performance degration of the static libraries. because, even the object files included by a static library are compiled with `-fPIC`, the linker treats these files as if they were directly compiled *into* the program when it links the seastar application.
- set the correct `Seastar_ASAN_OPTION` if `Seastar_BUILD_SHARED_LIBS` is set. this should fix the test failure when `Seastar_BUILD_SHARED_LIBS` is set, while `BUILD_SHARED_LIBS` is not.
- fix the typo of `BUILD_SHARED_LIB`. it should have been `BUILD_SHARED_LIBS`. this breaks the support of `BUILD_SHARED_LIBS`.
- deduce the values of `Seastar_BUILD_STATIC_LIBS` and `Seastar_BUILD_SHARED_LIBS` from the value of `BUILD_SHARED_LIBS` if it is set. if user wants to use `BUILD_SHARED_LIBS`, we should respect the decision, and in the meanwhile, we don't want to check all of these three variables everywhere. instead we just check `Seastar_BUILD_SHARED_LIBS` and `Seastar_BUILD_STATIC_LIBS`.
- remove `BUILD_SHARED_LIBS` as an option. because, if we have it as an option, we would have to provide a default value for it. either it is TRUE of FALSE, we would have to respect it. this defeats the purpose of 54e25a9d, and prevents us from having both shared *and* static libraries in a single pass. but we still respect this CMake variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,17 +317,26 @@ set (Seastar_TEST_TIMEOUT
   STRING
   "Maximum allowed time for a test to run, in seconds.")
 
-option (BUILD_SHARED_LIBS
-  "Build seastar library as shared libraries instead of static"
-  OFF)
 cmake_dependent_option (Seastar_BUILD_SHARED_LIBS
   "Build Seastar as shared libraries" OFF
-  "NOT DEFINED BUILD_SHARED_LIB" OFF)
+  "NOT DEFINED BUILD_SHARED_LIBS" OFF)
 cmake_dependent_option (Seastar_BUILD_STATIC_LIBS
   "Build Seastar as static libraries" ON
-  "NOT DEFINED BUILD_SHARED_LIB" OFF)
+  "NOT DEFINED BUILD_SHARED_LIBS" OFF)
+if (DEFINED BUILD_SHARED_LIBS)
+  if (BUILD_SHARED_LIBS)
+    set (Seastar_BUILD_SHARED_LIBS ON)
+    set (Seastar_BUILD_STATIC_LIBS OFF)
+  else ()
+    set (Seastar_BUILD_SHARED_LIBS OFF)
+    set (Seastar_BUILD_STATIC_LIBS ON)
+  endif ()
+endif ()
 if (NOT (Seastar_BUILD_SHARED_LIBS OR Seastar_BUILD_STATIC_LIBS))
   message (FATAL_ERROR "Seastar_BUILD_STATIC_LIBS and Seastar_BUILD_STATIC_LIBS cannot both be disabled")
+endif ()
+if (Seastar_BUILD_SHARED_LIBS)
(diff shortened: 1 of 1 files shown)


tchaikov:
@denesb hi Botond, could you please help review this change as well?

54e25a9d4a285c9489d1a7aeb4c6b8c10d4af4f3 was able to pass the seastar's CI, but it is causing build failures with scylla. i apologize for not conducting more thorough testing beforehand.

denesb (approved):

tchaikov:
@scylladb/seastar-maint hello maintainers, could you help review/merge this change?

travisdowns:
> * we set the global `CMAKE_POSITION_INDEPENDENT_CODE` as long as `Seastar_BUILD_STATIC_LIBS` is set. please note, this does not incur the runtime performance degration of the static libraries. because, even the object files included by a static library are compiled with `-fPIC`, the linker treats these files as if they were directly compiled _into_ the program when it links the seastar application.

I am pretty sure that is not true. Using -fPIC causes the .o file to have pessimistic code generation for certain things, like TLS and indirectly through the plt, and the linker can't undo all the damage, certainly not for TLS. This is obvious in Redpanda where we compile with -fPIC, but then link these into a static binary, and worse code is generated for some hotspot I was looking at (the memory allocator). Removing the -fPIC is the only way to get the best code generation.

So if you want to have a mix of static and dynamically lined binaries/libraries: you either use -fPIC everywhere and accept the slower code, or you compile object files twice (that's what Bazel does by default in this scenario).

travisdowns:
Here's a godbolt for a simple function that returns a TLS `int` variable: https://godbolt.org/z/seGP8qr9f.

First panel is without -fPIC, note how it is just a single instruction. Second panel is with -fPIC, it has a function call in there and a bunch of weird padding (the data16 stuff):

image


Third panel is after linking it into a binary. The linker _was_ able to improve the code (this is the only time outside of LTO that the linker does "codegen", AFAIK!) and get rid of the function but it's still much worse than the non -fPIC code.

tchaikov:
@travisdowns thank you very much. i failed to take the impact to tls into account. i will 54e25a9d4a285c9489d1a7aeb4c6b8c10d4af4f3 and compile the object files twice then.

PR #2490, opened 2024-10-12
scheduling: allow to check number of free scheduling group slots

Jadw1:
In ScyllaDB, we'd like to check how many scheduling groups were already created or how many free slots are there.

`scheduling_group_count()` was already present but it was in an internal namespace, making it impossible to use it in any app. This patch makes it exported to the user.
`scheduling_group_free_slots()` returns number of free scheduling group slots.

Refs https://github.com/scylladb/scylladb/issues/21051

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -124,8 +124,6 @@ class buffer_allocator;
 class priority_class;
 class poller;
 
-size_t scheduling_group_count();
-
 void increase_thrown_exceptions_counter() noexcept;
 
 }
--- a/include/seastar/core/scheduling.hh
+++ b/include/seastar/core/scheduling.hh
@@ -126,6 +126,8 @@ future<> rename_scheduling_group(scheduling_group sg, sstring new_name) noexcept
 /// \return a future that is ready when the scheduling group has been renamed
 future<> rename_scheduling_group(scheduling_group sg, sstring new_name, sstring new_shortname) noexcept;
 
+/// Returns number of active scheduling groups.
+size_t scheduling_group_count();
 
 /**
  * Represents a configuration for a specific scheduling group value,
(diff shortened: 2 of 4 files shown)


avikivity:
This is susceptible to races. If two callers call at the same time, each thinks it has a free slot, but only one will take it.

Better to harden the create function so it returns a documented exception if it fails to create a new scheduling group.

Jadw1:
> Better to harden the create function so it returns a documented exception if it fails to create a new scheduling group.

Maybe it's not the best place to discuss this, but I did this patch to do a regression test for number of scheduling groups in open source Scylla https://github.com/scylladb/scylladb/issues/21051#issuecomment-2405533306.
I wanted to assert/test that after all internal scheduling groups are created, there is enough place for enterprises service levels (7 scheduling groups), so regression like https://github.com/scylladb/scylladb/issues/21051 won't happen again. 

Do you have other idea how to test it in open source?

nyh:
I don't think it's necessary to test this in open source Scylla if the limitation doesn't exist there - you can check it where this limitation does exist.

piodul:
> I don't think it's necessary to test this in open source Scylla if the limitation doesn't exist there - you can check it where this limitation does exist.

The problem is that introducing a new scheduling group in open source introduces a problem which is only noticed later when merging OSS -> Enterprise. I assume (maybe incorrectly) that we want to keep `SEASTAR_SCHEDULING_GROUPS_COUNT` in sync between two repos, so the fix in this case is to go back to open source and bump the value of the variable. Therefore we wanted to have a test which forces one to bump the variable in OSS when introducing a new scheduling group - to avoid a painful situation.

Having this API would make the test that we wanted to introduce slightly less cumbersome - we intended to have a test-only API that returns the number of available scheduling groups and run it on a real scylla node. The alternative would be to just try to see how many scheduling groups we can create and report that.

Issue #2494, opened 2024-10-14
Current master as of 14 October 2024 fails to build with DPDK on Debian stable (bookworm) if libarchive-dev is installed


p12tic:
As of DPDK 23.07, if libarchive-dev is present, it will make DPDK depend on the library. Unfortunately DPDK also has a bug in its .pc file generation and will not include libarchive dependency under any circumstance. Therefore building seastar will fail with link errors.

Reproduction steps:

```
sudo docker run -it debian:bookworm /bin/bash
```

Inside:
```
apt update
apt install git g++ cmake ninja-build
git clone https://github.com/scylladb/seastar
cd seastar
git submodule update --init
./install-dependencies.sh
apt install libarchive-dev
./configure.py --mode=release --enable-dpdk
ninja -C build/release
```

Errors such as the following are observed:

```
[102/229] Linking CXX executable demos/block_discard_demo
FAILED: demos/block_discard_demo 
: && /usr/bin/g++ -O2 -g -DNDEBUG  demos/CMakeFiles/demo_block_discard.dir/block_discard_demo.cc.o -o demos/block_discard_demo  /usr/lib/x86_64-linux-gnu/libboost_program_options.so  libseastar.a  /usr/lib/x86_64-linux-gnu/libboost_program_options.so  /usr/lib/x86_64-linux-gnu/libboost_thread.so  /usr/lib/x86_64-linux-gnu/libboost_chrono.so  /usr/lib/x86_64-linux-gnu/libboost_date_time.so  /usr/lib/x86_64-linux-gnu/libboost_atomic.so  /usr/lib/x86_64-linux-gnu/libcares.so  /usr/lib/x86_64-linux-gnu/libfmt.so.9.1.0  /usr/lib/x86_64-linux-gnu/liblz4.so  -ldl  /usr/lib/x86_64-linux-gnu/libgnutls.so  -latomic  /usr/lib/x86_64-linux-gnu/libsctp.so  /usr/lib/x86_64-linux-gnu/libprotobuf.so  /usr/lib/x86_64-linux-gnu/libyaml-cpp.so  /usr/lib/x86_64-linux-gnu/libhwloc.so  /usr/lib/x86_64-linux-gnu/liburing.so  /usr/lib/x86_64-linux-gnu/libnuma.so && :
/usr/bin/ld: libseastar.a(dpdk.o): in function `firmware_read':
eal_firmware.c:(.text+0x2a7d2): undefined reference to `archive_read_new'
[...]

PR #2495, opened 2024-10-14
cmake: Fix DPDK libarchive dep

p12tic:
If libarchive-dev is installed during build time, DPDK will depend on it without a way to turn this off.

Accordingly the only way to address this issue is to ensure that libarchive is added to the dependencies using similar logic as dpdk uses.

Fixes https://github.com/scylladb/seastar/issues/2494.

### DIFF
--- a/cmake/Finddpdk.cmake
+++ b/cmake/Finddpdk.cmake
@@ -136,6 +136,13 @@ if ("bsd" IN_LIST dpdk_PC_LIBRARIES)
   list (APPEND dpdk_dependencies "bsd")
 endif ()
 
+# As of DPDK 23.07, if libarchive-dev is present, it will make DPDK depend on the library.
+# Unfortunately DPDK also has a bug in its .pc file generation and will not include libarchive
+# dependency under any circumstance. Accordingly, the dependency is added explicitly if libarchive
+# exists.
+pkg_check_modules(libarchive_PC libarchive)
+list(APPEND dpdk_dependencies ${libarchive_PC_LIBRARIES})
+
 if (dpdk_FOUND AND NOT (TARGET dpdk))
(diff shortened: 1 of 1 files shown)


cmake/Finddpdk.cmake
```diff
@@ -136,6 +136,13 @@ if ("bsd" IN_LIST dpdk_PC_LIBRARIES)
   list (APPEND dpdk_dependencies "bsd")
 endif ()
 
+# As of DPDK 23.07, if libarchive-dev is present, it will make DPDK depend on the library.
+# Unfortunately DPDK also has a bug in its .pc file generation and will not include libarchive
+# dependency under any circumstance. Accordingly, the dependency is added explicitly if libarchive
+# exists.
+pkg_check_modules(libarchive_PC libarchive)
```
tchaikov, line 143: could you pass `QUIET` parameter to this call. as, whether `libarchive` is detected, is not that important here. so would be better to silence the output.

also, could you add a space before `(`. to be more consistent with the rest of this source file?

tchaikov (commented):
thank you. i also recall this issue when reviewing your change, i ran into it couple years ago. 

apart a couple nits, lgtm.

cmake/Finddpdk.cmake
p12tic: > also, could you add a space before (. to be more consistent with the rest of this source file?

Sorry about this, I copied from a couple of lines below.

tchaikov: my bad. i wasn't quite consistent when creating this cmake module.

p12tic:
@avikivity I'm sorry for being late, I didn't address all comments raised by @tchaikov. This PR addresses them: https://github.com/scylladb/seastar/pull/2500

avikivity:
Oh, I misunderstood the conversation.

PR #2497, opened 2024-10-14
c-ares: update cooking version to 1.32.3

travisdowns:
Seastar no longer builds with the old (1.13) version of c-ares used when cooking that component, and in any case the URL is 404 as it seems that c-ares has moved to hosting their downloads on github.

Update to 1.32.3 which is the newest version less than 1.33 as seastar does not support versions >= 1.33.


tchaikov (approved):
lgtm. thanks!

tchaikov:

PR #2499, opened 2024-10-15
http status classification

kreuzerkrieg:
Add `status_class` enum for possible http code groups (classes) and a function `classify_status` to get code's classification. Intended for general usage to prevent code like `if (http_code >= XXX && http_code < YYY)`

In general all http codes are divided into 5 main categories

1xx: Informational - Request received, continuing process
2xx: Success - The action was successfully received, understood, and accepted
3xx: Redirection - Further action must be taken in order to complete the request
4xx: Client Error - The request contains bad syntax or cannot be fulfilled
5xx: Server Error - The server failed to fulfill an apparently valid request

This implementation does not take into account unassigned ranges, for more details https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

### DIFF
--- a/include/seastar/http/reply.hh
+++ b/include/seastar/http/reply.hh
@@ -112,6 +112,38 @@ struct reply {
     } _status;
 
     /**
+     * HTTP status classes
+     * See https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
+     *
+     * 1xx: Informational - Request received, continuing process
+     * 2xx: Success - The action was successfully received, understood, and accepted
+     * 3xx: Redirection - Further action must be taken in order to complete the request
+     * 4xx: Client Error - The request contains bad syntax or cannot be fulfilled
+     * 5xx: Server Error - The server failed to fulfill an apparently valid request
+     */
+    enum class status_class : uint8_t {
+        informational = 1,
+        success = 2,
+        redirection = 3,
+        client_error = 4,
+        server_error = 5,
+        unclassified
+    };
+
+    /**
+     * Classify http status
+     * @param http_status the http status \ref status_type
+     * @return one of the \ref status_class values
+     */
+    static constexpr status_class classify_status(status_type http_status) {
+        auto sc = static_cast<std::underlying_type_t<status_type>>(http_status) / 100;
+        if (sc < 1 || sc > 5) [[unlikely]] {
+            return status_class::unclassified;
+        }
+        return static_cast<status_class>(sc);
+    }
+
+    /**
      * The headers to be included in the reply.
      */
     std::unordered_map<sstring, sstring, seastar::internal::case_insensitive_hash, seastar::internal::case_insensitive_cmp> _headers;
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -1831,3 +1831,58 @@ BOOST_AUTO_TEST_CASE(test_path_decode_changed) {
     auto expected_chars = seastar::sstring{" "};
     BOOST_REQUIRE_EQUAL(result, expected_chars);
 }
+
+namespace seastar::http {
+std::ostream& boost_test_print_type(std::ostream& os, reply::status_class sc) {
+    constexpr std::string_view status_strings[] {
+        "1xx: Informational",
+        "2xx: Success",
+        "3xx: Redirection",
+        "4xx: Client Error",
+        "5xx: Server Error",
+        "Unclassified"
+    };
+    auto status = static_cast<std::underlying_type_t<reply::status_class>>(sc) - 1u;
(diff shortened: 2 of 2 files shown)


include/seastar/http/reply.hh
```diff
@@ -111,6 +111,38 @@ struct reply {
...
+        unclassified
+    };
+
+    /**
+     * Classify http status
+     * @param http_status the http status \ref status_type
+     * @return one of the \ref status_class values
+     */
+    static constexpr status_class classify_status(status_type http_status) {
+        auto sc = static_cast<uint16_t>(http_status) / 100;
```
tchaikov, line 139: nit, the underlying type of `status_type` is `int`. as by default if the underlying type is not specified, `enum class status_type` uses `int` as its underlying type. so IMHO, `uint16_t` is a, well, "magic type". if we want to perform integer math on it, i think, a better way is probably to cast it to `static_cast<std::underlying_type_t<status_type>>(http_status)`. BTW, in C++23, we would be able to use `std::to_underlying(http_status)`. but we are not there yet..

include/seastar/http/reply.hh
```diff
@@ -111,6 +111,38 @@ struct reply {
...
+     * See https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
+     *
+     * 1xx: Informational - Request received, continuing process
+     * 2xx: Success - The action was successfully received, understood, and accepted
+     * 3xx: Redirection - Further action must be taken in order to complete the request
+     * 4xx: Client Error - The request contains bad syntax or cannot be fulfilled
+     * 5xx: Server Error - The server failed to fulfill an apparently valid request
+     */
+    enum class status_class : uint8_t {
+        informational=1,
```
tchaikov, line 125: could you add space around `=` for better readability?

kreuzerkrieg: sure

include/seastar/http/reply.hh
kreuzerkrieg: yup, looks cleaner this way

src/http/reply.cc
```diff
@@ -110,6 +110,16 @@ std::ostream& operator<<(std::ostream& os, reply::status_type st) {
     return os << status_strings::to_string_view(st);
 }
 
+std::ostream& operator<<(std::ostream& os, reply::status_class sc) {
+    constexpr std::array<std::string_view, 6> status_strings{
```
tchaikov, line 114: we could use `std::to_array`. but why not a plain array like 
```c++
std::string_view status_strings[] = {
  "1xx: Informational",
  /// ..
};
```

src/http/reply.cc
```diff
@@ -110,6 +110,16 @@ std::ostream& operator<<(std::ostream& os, reply::status_type st) {
...
+        "1xx: Informational", "2xx: Success", "3xx: Redirection", "4xx: Client Error", "5xx: Server Error", "Unclassified"};
```
tchaikov, line 115: might want to wrap this line for better readability.

src/http/reply.cc
```diff
@@ -110,6 +110,16 @@ std::ostream& operator<<(std::ostream& os, reply::status_type st) {
...
+    auto status = static_cast<uint8_t>(sc) - 1;
+    if (status < 6) {
```
tchaikov, line 117: might want to replace this magic number with `std::size(status_strings)`.

tchaikov (commented):
thanks for the revision. just some nits.

include/seastar/http/reply.hh
```diff
@@ -247,6 +279,7 @@ private:
 };
 
 std::ostream& operator<<(std::ostream& os, reply::status_type st);
+std::ostream& operator<<(std::ostream& os, reply::status_class sc);
```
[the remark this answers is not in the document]
kreuzerkrieg: I think converting it to formatter will break boost tests, right?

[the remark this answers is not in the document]
tchaikov: IIUC, this is a new formatter. how come it will break existing boost tests?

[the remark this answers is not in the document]
tchaikov: i see you are referencing the newly introduced tests. please see https://gist.github.com/tchaikov/9f4cb52487189f51aa87ca8fd56dfd4f#boosttest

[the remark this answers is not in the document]
kreuzerkrieg: ah, then just move the implementation to the test itself under `boost_test_print_type`? will do it right away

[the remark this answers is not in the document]
kreuzerkrieg: Done

tchaikov (approved):
lgtm

kreuzerkrieg:

PR #2500, opened 2024-10-16
cmake: Don't emit message when searching for libarchive

p12tic:
Existence of libarchive does not change behavior because it's only used to work around a bug in dpdk.

Fixes d6dd7633d53afbfbe3d29143e092f94c459baf88


tchaikov (approved):
thank you for the follow-up change!

tchaikov:

PR #2504, opened 2024-10-17
build: do not check for CMAKE_CXX_STANDARD < 20

tchaikov:
in 5d3ee98073, we dropped the support of C++17. so no need to check if the C++ standard is less than 20 -- it should be always greater or equal to 20.


dawmd (approved):
LGTM

tchaikov:

PR #2506, opened 2024-10-17
stall-analyser: pass args.tmin instead of tmin

tchaikov:
8e583411 introduced a regression, which passes an undefined variable of `tmin` to `print_stats()`.

in this change, we pass `args.tmin` instead of `tmin` to this function.


tchaikov:

Issue #2375, opened 2024-07-29
"smp" metrics are disabled


xemul:
0634989343fa1dd2585ba911383784569f1da5c3 contains such hunk around "smp" group
```patch
@@ -2867,49 +2749,22 @@ size_t smp_message_queue::process_incoming() {
 
 void smp_message_queue::start(unsigned cpuid) {
     _tx.init();
+    namespace sm = seastar::metrics;
     char instance[10];
     std::snprintf(instance, sizeof(instance), "%u-%u", engine().cpu_id(), cpuid);
-    _collectd_regs = scollectd::registrations({
+    _metrics.add_group("smp", {
             // queue_length     value:GAUGE:0:U
             // Absolute value of num packets in last tx batch.
-            scollectd::add_disabled_polled_metric(scollectd::type_instance_id("smp"
-                    , instance
-                    , "queue_length", "send-batch")
-            , scollectd::make_typed(scollectd::data_type::GAUGE, _last_snt_batch)
-            ),
-            scollectd::add_disabled_polled_metric(scollectd::type_instance_id("smp"
-                    , instance
-                    , "queue_length", "receive-batch")
-            , scollectd::make_typed(scollectd::data_type::GAUGE, _last_rcv_batch)
-            ),
-            scollectd::add_disabled_polled_metric(scollectd::type_instance_id("smp"
-                    , instance
-                    , "queue_length", "complete-batch")
-            , scollectd::make_typed(scollectd::data_type::GAUGE, _last_cmpl_batch)
-            ),
-            scollectd::add_disabled_polled_metric(scollectd::type_instance_id("smp"
-                    , instance
-                    , "queue_length", "send-queue-length")
-            , scollectd::make_typed(scollectd::data_type::GAUGE, _current_queue_length)
-            ),
+            sm::make_queue_length("send_batch_queue_length", _last_snt_batch, sm::description("Current send batch queue length"), sm::metric_disabled, instance),
+            sm::make_queue_length("receive_batch_queue_length", _last_rcv_batch, sm::description("Current receive batch queue length"), sm::metric_disabled, instance),
+            sm::make_queue_length("complete_batch_queue_length", _last_cmpl_batch, sm::description("Current complete batch queue length"), sm::metric_disabled, instance),
+            sm::make_queue_length("send_queue_length", _current_queue_length, sm::description("Current send queue length"), sm::metric_disabled, instance),
...
```

so all the "smp" metrics were marked as "disabled" and they are indeed not reported AFAICT

[...]

PR #2507, opened 2024-10-18
Enable `smp_message_queue metrics` by default

ballard26:
These metrics were originally disabled for overloading the collectd servers many years back. However, these days they seem like a drop in the bucket relative to number of metrics seastar generates. Beyond this they would be useful to have when debugging issues related to cross-shard traffic.

### DIFF
--- a/include/seastar/core/metrics.hh
+++ b/include/seastar/core/metrics.hh
@@ -423,8 +423,6 @@ metric_function make_function(T& val, data_type dt) {
 }
 }
 
-extern const bool metric_disabled;
-
 extern label shard_label;
 
 /*
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -544,8 +544,6 @@ void impl::set_metric_family_configs(const std::vector<metric_family_config>& fa
 }
 }
 
-const bool metric_disabled = false;
-
 relabel_config::relabel_action relabel_config_action(const std::string& action) {
     if (action == "skip_when_empty") {
         return relabel_config::relabel_action::skip_when_empty;
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3646,16 +3646,16 @@ void smp_message_queue::start(unsigned cpuid) {
     _metrics.add_group("smp", {
             // queue_length     value:GAUGE:0:U
             // Absolute value of num packets in last tx batch.
-            sm::make_queue_length("send_batch_queue_length", _last_snt_batch, sm::description("Current send batch queue length"), {sm::shard_label(instance)})(sm::metric_disabled),
-            sm::make_queue_length("receive_batch_queue_length", _last_rcv_batch, sm::description("Current receive batch queue length"), {sm::shard_label(instance)})(sm::metric_disabled),
-            sm::make_queue_length("complete_batch_queue_length", _last_cmpl_batch, sm::description("Current complete batch queue length"), {sm::shard_label(instance)})(sm::metric_disabled),
-            sm::make_queue_length("send_queue_length", _current_queue_length, sm::description("Current send queue length"), {sm::shard_label(instance)})(sm::metric_disabled),
+            sm::make_queue_length("send_batch_queue_length", _last_snt_batch, sm::description("Current send batch queue length"), {sm::shard_label(instance)}),
+            sm::make_queue_length("receive_batch_queue_length", _last_rcv_batch, sm::description("Current receive batch queue length"), {sm::shard_label(instance)}),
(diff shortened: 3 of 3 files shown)


tchaikov:
@ballard26 hi Brandon, i think your PR actually fixes #2375 which is still controversial. because, @amnonh think that 

> Those metrics are quadratic in their nature and were disabled

IIUC, by "quadratic", it means we have n x n metrics if the seastar is configured with n cores. guess that's why they were disabled. but i think, despite that they are always disabled, these metrics are still added to the metric subsystem, even they are not added to the metrics map and are not exposed.

does it make sense to add a member to `smp_options` to at least allow user to enable them using `smp::configure()` ? so that if one suspect that the system is suffering from too many cross shard traffic/calls, they can enable this and restart the seastar application, to verify the hypothesis .

amnonh:
I vote against such a configuration. We already have enough options to control metrics from API and files

tchaikov:
> I vote against such a configuration. We already have enough options to control metrics from API and files

@amnonh hi Amnon, do you mean that we can utilize these options to enable the given disabled metric ?

amnonh:
@tchaikov yes, you can use the file or the API to to enable them, for example, I've run two nodes (docker) cluster, 
I've set all metrics using the following config:
```
$cat relabel/conf.yaml 
relabel_configs:
  - source_labels: [__name__]
    action: keep
    regex: (.*)

```
Using the browser: ``http://172.17.0.3:9180/metrics?__name__=smp*``
```
# HELP scylla_smp_complete_batch_queue_length Current complete batch queue length
# TYPE scylla_smp_complete_batch_queue_length gauge
scylla_smp_complete_batch_queue_length{shard="0-1"} 0.000000
scylla_smp_complete_batch_queue_length{shard="1-0"} 0.000000
# HELP scylla_smp_receive_batch_queue_length Current receive batch queue length
# TYPE scylla_smp_receive_batch_queue_length gauge
scylla_smp_receive_batch_queue_length{shard="0-1"} 0.000000
scylla_smp_receive_batch_queue_length{shard="1-0"} 0.000000
# HELP scylla_smp_send_batch_queue_length Current send batch queue length
# TYPE scylla_smp_send_batch_queue_length gauge
scylla_smp_send_batch_queue_length{shard="0-1"} 1.000000
scylla_smp_send_batch_queue_length{shard="1-0"} 1.000000
# HELP scylla_smp_send_queue_length Current send queue length
# TYPE scylla_smp_send_queue_length gauge
scylla_smp_send_queue_length{shard="0-1"} 0.000000
scylla_smp_send_queue_length{shard="1-0"} 0.000000
# HELP scylla_smp_total_completed_messages Total number of messages completed
# TYPE scylla_smp_total_completed_messages counter
scylla_smp_total_completed_messages{shard="0-1"} 3883
scylla_smp_total_completed_messages{shard="1-0"} 6889
# HELP scylla_smp_total_received_messages Total number of received messages
# TYPE scylla_smp_total_received_messages counter
scylla_smp_total_received_messages{shard="0-1"} 3883
scylla_smp_total_received_messages{shard="1-0"} 6889
# HELP scylla_smp_total_sent_messages Total number of sent messages
# TYPE scylla_smp_total_sent_messages counter
scylla_smp_total_sent_messages{shard="0-1"} 3883
scylla_smp_total_sent_messages{shard="1-0"} 6889
```

If you have access to the API, you can do the same in runtime using the ``http://localhost:10000/v2/metrics-config/`` GET and POST APIs

ballard26:
> IIUC, by "quadratic", it means we have n x n metrics if the seastar is configured with n cores. guess that's why they were disabled. but i think, despite that they are always disabled, these metrics are still added to the metric subsystem, even they are not added to the metrics map and are not exposed.

Yeah, knowing this now its understandable why these metrics are not enabled by default. 


Most of the time we're only ever interested in knowing the total ingress and egress of cross shard calls by shard rather than by queue. This would be a linear rather than quadratic metric relative to the number of shards and therefore something that may be more palatable to have enabled by default.

Is there any interest from you folks for having a metric like that enabled by default in seastar? If not, I believe we could achieve a similar result in our application by using `set_relabel_configs(...)` to split the `shard="a-b"` label in the existing metric to `shard="a", dst_shard="b"` then `set_metric_family_configs(...)` to aggregate the `dst_shard` label.

tchaikov:
> @tchaikov yes, you can use the file or the API to to enable them, for example, I've run two nodes (docker) cluster, I've set all metrics using the following config:
>  ...
> 
> If you have access to the API, you can do the same in runtime using the `http://localhost:10000/v2/metrics-config/` GET and POST APIs

thank you very much Amnon, didn't know that we could enable metrics at runtime.

PR #2508, opened 2024-10-18
Add missing `assert.h` header to `chunked_fifo`

ballard26:
`assert(...)` is being used in the `pop_front_n` method and none of the other headers in the file include assert transitively.

### DIFF
--- a/include/seastar/core/chunked_fifo.hh
+++ b/include/seastar/core/chunked_fifo.hh
@@ -23,6 +23,7 @@
 
 #ifndef SEASTAR_MODULE
 #include <algorithm>
+#include <cassert>
 #include <type_traits>
 #include <seastar/util/modules.hh>
 #endif


include/seastar/core/chunked_fifo.hh
```diff
@@ -23,6 +23,7 @@
 
 #ifndef SEASTAR_MODULE
 #include <algorithm>
+#include <assert.h>
```
tchaikov, line 26: how about using `cassert` in place of `assert.h`? i understand they are practically equivalent. and because `assert()` is always implemented as a macro, there is no such as thing like `std::assert()`. but, IMHO, would be nice if we can always use the headers provided by the C++ standard library when writing a C++ application/library. what do you think?

ballard26: Good point, will switch it over to `cassert`.

tchaikov (approved):
lgtm

PR #2512, opened 2024-10-20
rpc: switch rpc::type from boost to std

avikivity:
rpc has a marker type rpc::type which it uses to communicate with the user-provided deserializer. It is currently based on boost::type.

In an ideal universe, switching to the similar std::type_identity would involve only changing the typedef. However, some users pass boost::type directly, so we can't do that without breaking them.

To preserve compatiblity, have all read() calls pass through a helper function that checks which variant is available and calls it. The older one is marked deprecated. With that done, we switch the rpc::type alias to std::type_identity.


gleb-cloudius (approved):

avikivity:
@xemul please review

PR #2516, opened 2024-10-24
doc: compatibility: update the notes on supported GCC versions

tchaikov:
- replace gcc and clang with GCC and Clang. as GCC and Clang are their names. while "gcc" and "clang" are command line names we are using.
- replace gcc 11 and gcc 12 with GCC 13 and GCC 14. as the latest GCC version is GCC 14.

### DIFF
--- a/doc/compatibility.md
+++ b/doc/compatibility.md
@@ -38,11 +38,11 @@ Patches for new platforms (e.g, Windows) are welcome.
 Compilers
 =========
 
-Seastar supports gcc and clang. Ports to other compilers are
+Seastar supports GCC and Clang. Ports to other compilers are
 welcome.
 
 The last two major releases of a compiler are supported (e.g.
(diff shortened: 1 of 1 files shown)


doc/compatibility.md
```diff
@@ -38,11 +38,11 @@ Patches for new platforms (e.g, Windows) are welcome.
 Compilers
 =========
 
-Seastar supports gcc and clang. Ports to other compilers are
+Seastar supports GCC and Clang. Ports to other compilers are
 welcome.
 
 The last two major releases of a compiler are supported (e.g.
-gcc 11 and gcc 12). Patches to support older versions are welcome,
+GCC 13 and GCC 14). Patches to support older versions are welcome,
```
xemul, line 45: "Official" Dockerfile installs gcc-12 and gcc-13

tchaikov: created #2517.

PR #2517, opened 2024-10-24
docker: bump up to clang {18,19} and gcc {13,14}

tchaikov:
since we only support the latest two major releases of compilers. at the moment of writing, Clang just released v19, and the latest major release of GCC is v14.

so we should install them respectively.

because ubuntu mantis is already EOF. see
https://wiki.ubuntu.com/Releases. and it does not include the latest GCC and Clang packages. we need to bump up the base image from ubuntu:mantis to ubuntu:orcular despite that orcular is not an LTS release. we should use new ubuntu LTS release once it's out.

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -1,34 +1,34 @@
 # syntax=docker/dockerfile:1
 
-FROM ubuntu:mantic
+FROM ubuntu:oracular
 
 RUN --mount=type=bind,source=./install-dependencies.sh,target=/install-dependencies.sh \
     apt-get update && apt-get install -y \
     curl \
     gnupg \
-    && echo "deb http://apt.llvm.org/mantic/ llvm-toolchain-mantic-17 main" \
+    && echo "deb http://apt.llvm.org/mantic/ llvm-toolchain-oracular-18 main" \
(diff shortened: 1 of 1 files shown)


docker/dev/Dockerfile
```diff
@@ -1,34 +1,34 @@
...
 
 RUN --mount=type=bind,source=./install-dependencies.sh,target=/install-dependencies.sh \
     apt-get update && apt-get install -y \
     curl \
     gnupg \
-    && echo "deb http://apt.llvm.org/mantic/ llvm-toolchain-mantic-17 main" \
+    && echo "deb http://apt.llvm.org/mantic/ llvm-toolchain-oracular-18 main" \
     >> /etc/apt/sources.list.d/llvm.list \
-    && echo "deb http://apt.llvm.org/mantic/ llvm-toolchain-mantic-18 main" \
+    && echo "deb http://apt.llvm.org/mantic/ llvm-toolchain-oracular-19 main" \
```
xemul, line 11: Commit message says "Clang just released v17", should be v19 maybe?

tchaikov:
v2:

- correct the version numbers in commit message and cover letter: s/v17/v19/, s/v13/v14/

Issue #2520, opened 2024-10-26
Add soname to shared library


bkmgit:
This would help for Fedora packaging and to enable easy upgrades.

PR #2521, opened 2024-10-27
build: add soname to shared libraries

bkmgit:
Closes https://github.com/scylladb/seastar/issues/2520


bkmgit:
At present the version is defined in the CMakeLists.txt file.  It could be defined elsewhere
so that when a new release is made, the tag is also automatically defined.  Currently the
tag seems to be defined manually.

CMakeLists.txt
```diff
@@ -43,8 +43,13 @@ set (Seastar_DPDK_MACHINE
   STRING
   "Configure DPDK for this processor architecture (if `Seastar_DPDK` is enabled). It configures -march or -mcpu")
 
+# Release versio used in shared library soname
+set (Seastar_VERSION "22.11.0")
```
tchaikov, line 47: The version "22.11.0" was derived from its release date, as shown in commit 10b7d604d1f5037a733879d8d171d4405faebbe9. However, I recommend using `SEASTAR_API_LEVEL` instead of this date-based version as the SO (shared object) version number.

Here's why:

1. Applications select and link against shared libraries based on their SO versions
2. If we redistribute Seastar, we should use SO versions that meaningfully reflect API compatibility
3. The SO version should clearly indicate when breaking API changes occur - incrementing it signals that the new API is incompatible with previous versions

This approach provides better clarity around API compatibility compared to date-based versioning.

tchaikov (changes requested):
i'd use `SEASTAR_API_LEVEL` for the SO version.

avikivity:
Please explain in the patch changelog why we're making this change.

bkmgit:
> Please explain in the patch changelog why we're making this change.

Done.

CMakeLists.txt
bkmgit: - Used Seastar_API_LEVEL as soname.
- Version is still defined as the release date, can either:
    - revert to version being 1
    - define the version manually in CMakeLists.txt based on the last tag
    - create a script that will generate a VERSION file in the repository from the last tag made

bkmgit: Have left VERSION in CMakeLists.txt as 1.  If needed, can update in a separate issue.

CMakeLists.txt
```diff
@@ -45,6 +45,8 @@ set (Seastar_DPDK_MACHINE
 
 project (Seastar
   VERSION 1.0
+  DESCRIPTION "Event driven asynchronous programming framework"
+  HOMEPAGE_URL "http://seastar.io/"
```
tchaikov, line 49: nit, strictly speaking, this is not part of this change.

tchaikov (approved):
modulo a nit, lgtm

tchaikov:
@bkmgit hi Benson, i am curious, do you really plan to package seastar for a distro like fedora?

CMakeLists.txt
bkmgit: Fixed.

CMakeLists.txt
```diff
@@ -780,6 +780,11 @@ add_library (seastar
   src/websocket/server.cc
   )
```
avikivity, line 782: > Use API level as soname to enable applications to select
and link agains shared libraries based on their SO versions.

Actually a neat idea. But note, API levels are only about source-level compatibility. It's possible for a function signature to change without change the API level.

Example: if we add a trailing default parameters to a function, it's still source compatible, but links against an older .so will fail.

Given this, is this still helpful for you?

bkmgit:
> @bkmgit hi Benson, just out of curiosity, do you really plan to package seastar for a distro like fedora?

https://bugzilla.redhat.com/show_bug.cgi?id=2321986

DPDK is already there, but need to see if changes allow it to be used.
Sylladb would be nice as well.

CMakeLists.txt
bkmgit: Am happy to follow up with another change.  It would be nice not to carry patches, but yes ideally soname would reflect ABI compatibility, not API compatibility. Still learning about Seastar, so not sure how often ABI is broken in a backwards incompatible manner. If the release year and month is used, then this would always increment the soname, and it would be assumed that typically ABI is broken on each update.

tchaikov:
> > @bkmgit hi Benson, just out of curiosity, do you really plan to package seastar for a distro like fedora?
> 
> https://bugzilla.redhat.com/show_bug.cgi?id=2321986
> 
> DPDK is already there, but need to see if changes allow it to be used. Scylladb would be nice as well.

thanks . i see. regarding DPDK, one thing i can think about is `RTE_MBUF_REFCNT_ATOMIC`. because seastar is inherently shared by CPU cores, so we don't need to enable  `RTE_MBUF_REFCNT_ATOMIC`. this could hurt the performance because the overhead of atomic operations. but this is not the end of world.

Avi's comment about SO versioning makes a lot of sense. I hadn't considered that perspective before. we even sometimes unintentionally change APIs over time, as seen in https://github.com/scylladb/seastar/pull/2410#discussion_r1756255666 . and AFAICT, we don't have a release schedule (yet). so probably it's best to encode date in the version number to be on the safe side, or just starts with "1" and increment it every time a new release is created, and the ABI is changed across the builds. sorry for the bike shedding. BTW, if we want to use date as the version number, probably should include the month in it. and we probably can set the version in the command line, like `cmake -DPROJECT_VERSION=23.3.14` or `cmake -DPROJECT_VERSION=202410.3.14`, and use `PROJECT_VERSION_MAJOR` for so version. or, just pass the `Seastar_VERSION_MAJOR` to `cmake`.

bkmgit:
As the version is currently 1, it seems ok to use that as the soname, but use the date as the software version number since that is the tag that is used in GitHub.  Perhaps we can automate this to be updated whenever a release is made rather than manually setting a CMake variable.  It will be less prone to error.  Will update the pull request with this.

bkmgit:
Have used a versioning scheme of
PROJECT_VERSION.API_LEVEL.RELEASE_DATE

If this is ok, can add tooling that would update this automatically whenever a release is made by checking for ABI changes.  If someone is using a commit that is different from a release, it would be upto them to check ABI compatibility.  Tooling can be done in a follow up pull request.

bkmgit:
PROJECT_VERSION is currently 1, so can keep that.

avikivity:
> > @bkmgit hi Benson, just out of curiosity, do you really plan to package seastar for a distro like fedora?
> 
> https://bugzilla.redhat.com/show_bug.cgi?id=2321986
> 
> DPDK is already there, but need to see if changes allow it to be used. Scylladb would be nice as well.

Note, ScyllaDB releases use a fork of Seastar in https://github.com/scylladb/scylla-seastar.git. The fork is created on-demand (if we need to backport a Seastar commit), so not all releases reference it.

See scylladb/scylladb@6d779f58a9641fe85037d5e9d226126787302d3c (branch-5.4) and following submodule updates for an example. It's really geared towards bundling, not independently moving versions.

bkmgit:
Thanks. Will bundle when packaging Scylladb. Having other consumers of seastar would improve testing. Bundling multiple times is not great and it is already used in a [number of projects](https://github.com/scylladb/seastar#projects-using-seastar) Packages make deployments easier, though usually with a performance tradeoff. However, understand if it is too much extra effort unrelated to the main project goals to support a soname.

CMakeLists.txt
```diff
@@ -69,6 +69,16 @@ option (Seastar_DEPRECATED_OSTREAM_FORMATTERS
   "Enable operator<< for formatting standard library containers, which will be deprecated in future"
   ON)
 
+set (Seastar_RELEASE_DATE
+  20221127
+  CACHE
+  STRING
+  "Last release date")
+
+set_property (CACHE Seastar_RELEASE_DATE
```
tchaikov, line 78: why do we need a property for the release date?

CMakeLists.txt
```diff
@@ -780,6 +790,15 @@ add_library (seastar
   src/websocket/server.cc
   )
 
+# Shared library soname  
+set_target_properties(seastar PROPERTIES
+  VERSION ${PROJECT_VERSION}.${Seastar_API_LEVEL}.${Seastar_RELEASE_DATE}
```
tchaikov, line 795: In semantic versioning (semver), a version number consists of three components: MAJOR.MINOR.PATCH. The major version number changes least frequently and indicates incompatible API changes. The minor version is incremented more often for backward-compatible feature additions, while the patch version is used for backward-compatible bug fixes.

The `Seastar_API_LEVEL` changes less frequently than the ABI version, as it is only incremented for significant API changes rather than ABI-breaking modifications.

`PROJECT_VERSION` is also composed of multiple component. see https://cmake.org/cmake/help/latest/variable/PROJECT_VERSION.html. 

Based on the discussion above, I suggest:

- `PROJECT_VERSION` should be treated as a versioning string with multiple components (major.minor.patch) rather than a single number.
- Although `Seastar_API_LEVEL` is crucial for API versioning, it may not belong in the libraries' version numbers since we make ABI changes within the same `Seastar_API_LEVEL`.
- Including the release date in the version number makes sense given our flexible release schedule.

CMakeLists.txt
```diff
@@ -780,6 +790,15 @@ add_library (seastar
...
+  SOVERSION ${PROJECT_VERSION})
+
+# We disable _FORTIFY_SOURCE because it generates false positives with longjmp() (src/core/thread.cc)
+set_source_files_properties(src/core/thread.cc
+  PROPERTIES COMPILE_FLAGS -U_FORTIFY_SOURCE)
```
tchaikov, line 800: you are partially reverting c96444fbe43acac0d31418b65b04395c11fdda14. probably want to drop this change?

bkmgit:
Thanks for the suggestions.

CMakeLists.txt
bkmgit: Have used release date as the major soname number.  As ABI can change unpredictably, this seems reasonable.  Releases are not made very often, so for packaging this should be sufficient.

CMakeLists.txt
bkmgit: Done.

bkmgit:
> > > @bkmgit hi Benson, just out of curiosity, do you really plan to package seastar for a distro like fedora?
> > 
> > 
> > https://bugzilla.redhat.com/show_bug.cgi?id=2321986
> > DPDK is already there, but need to see if changes allow it to be used. Scylladb would be nice as well.
> 
> Note, ScyllaDB releases use a fork of Seastar in https://github.com/scylladb/scylla-seastar.git. The fork is created on-demand (if we need to backport a Seastar commit), so not all releases reference it.
> 
> See [scylladb/scylladb@6d779f5](https://github.com/scylladb/scylladb/commit/6d779f58a9641fe85037d5e9d226126787302d3c) (branch-5.4) and following submodule updates for an example. It's really geared towards bundling, not independently moving versions.

Disabled DPDK for the packaged build for now.

bkmgit:
@tchaikov @avikivity Package has been approved for Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=2321986 Is the versioning scheme reasonable?

avikivity:
Please rebase and squash the whole thing, we don't like fixes to problems the pull request introduces to be in the pull request itself.

The release date is ambiguous, we should include the branch as well.

PR #2522, opened 2024-10-28
thread: do not use fortify source

rockwotj:
Redpanda is in the process to switching to Bazel instead of CMake as it's build system.

One of the patches we apply is to `#undef _FORTIFY_SOURCE` in thread.cc, because that's
a lot simpler to do than enable a compilation flag for only a single file (via --per_file_copt)
like what the CMake build does. While this is a noop in the existing CMake build, it would
simplify our lives a bit. I'm also happy to remove the CMake per file specific compilation 
setting for the undef too if that's desired.

Reference: https://github.com/scylladb/seastar/pull/2344


src/core/thread.cc
```diff
@@ -1,3 +1,8 @@
+// If _FORTIFY_SOURCE is defined than longjmp ends up using longjmp_chk
```
tchaikov, line 1: might want to s/than/then/ ?

yeah, better off removing the `set_source_files_properties(..)` command. since we undefine `_FORTIFY_SOURCE` in the source file.

tchaikov (approved):
lgtm

src/core/thread.cc
rockwotj: Yes good call. Done

tchaikov:

src/core/thread.cc
```diff
@@ -1,3 +1,8 @@
+// If _FORTIFY_SOURCE is defined then longjmp ends up using longjmp_chk
+// which asserts that you're jumping to the same stack. However, here we
+// are intentionally switching stacks when longjmp'ing, so undefine this
+// option to always use normal longjmp.
+#undef _FORTIFY_SOURCE
```
nyh, line 5: Excellent. Not only does this do the undef in the right place - in the code and not in some unintelligible cmake file, it also reminds the reader that we disabled FORTIFY_SOURCE on this entire source file and if this file grows and this is considered too broad, or maybe should move just jmp_buf_link with its longjmp calls to a separate file.

nyh (approved):

nyh:
By the way, when you have some experience on bazel vs cmake, I'll be happy if you share it on the Seastar developer mailing list (does anyone still use that? :-( ). I never managed to become a cmake fan, and people keep telling me I'm wrong and cmake is great and the best thing since sliced bread - but I still failed to be come a fan... So I wonder if maybe I wasn't wrong all along and there's indeed something (bazel or anything else) that people like even better than cmake.

rockwotj:
> By the way, when you have some experience on bazel vs cmake, I'll be happy if you share it on the Seastar developer mailing list (does anyone still use that? :-( ). I never managed to become a cmake fan, and people keep telling me I'm wrong and cmake is great and the best thing since sliced bread - but I still failed to be come a fan... So I wonder if maybe I wasn't wrong all along and there's indeed something (bazel or anything else) that people like even better than cmake.

I'll see if I can convinced some of the folks in Redpanda to chime in on a mailing list thread. We are hard switching over in about a month, so I am sure there will be plenty of fresh opinions. Most of the team is familiar with CMake, but not so much Bazel.

I'm probably not a great one to ask - I am a former Googler and heavily drank the Bazel/Blaze is great Kool-Aid, I really like it and think it's great (happy to enumerate reasons still!). But I have also contributed to Bazel and am pretty familiar with the build system, where I am only a causal CMake/ninja/ctest user.

PR #2528, opened 2024-11-06
Revert "core/reactor: don't check AIO block count when they are not needed

denesb:
This reverts commit 3c9c2696a45ec0eea524760a75938900d75f79b8.

Turns out all current reactor backends *do* use AIO blocks and so this check is justified. Currently, with this check removed, if there are not enough AIO blocks, the startup of seastar will just fail at a different place.


avikivity:
I think that's wrong too. Different backends need different iocb counts.

Maybe the right thing is to let reactors construct themselves, have then throw some resource_error when they fail to create iocbs, and print that.

denesb:
Indeed, the `max_networking_aio_io_control_blocks` is only used by the AIO reactor backend. So the original patch was correct to make it AIO only.

PR #2530, opened 2024-11-07
file: add file_system_stat

bhalevy:
Return an abstract file_system_stat_data representing
the system statvfs for the filesystem identified
by the given file name.

The motivation is implementing a disk-space monitor
in scylladb.

### DIFF
--- a/include/seastar/core/file-types.hh
+++ b/include/seastar/core/file-types.hh
@@ -180,6 +180,22 @@ struct stat_data {
     std::chrono::system_clock::time_point time_changed;   // Time of last status change (either content or attributes)
 };
 
+/// Filesystem-wide stat information
+/// See statvfs(3)
+struct file_system_stat_data {
+    size_t block_size;          // Filesystem block size
+    size_t fragment_size;       // Fragment size
+    uint64_t size_in_bytes;     // Size of filesystem in bytes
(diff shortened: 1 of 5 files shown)


bhalevy:
- added `#include <coroutine>`

bhalevy:
@xemul can you please review and consider merging?

xemul:
This PR adds
- conversion from statvfs to newly introduced file_system_stat_data
- wrapper around already public reactor::statvfs()

So basically it adds 1:1 wrapper over existing APIs. What's the point?

bhalevy:
> This PR adds
> 
> * conversion from statvfs to newly introduced file_system_stat_data
> * wrapper around already public reactor::statvfs()
> 
> So basically it adds 1:1 wrapper over existing APIs. What's the point?

The wrapper is similar to other "global" entry points we have so that callers won't need to access the reactor api.

As for the abstract file_system_stat_data, again, it's similar to the file stat_data struct that is not linux specific and provides an isolation layer.

bhalevy:
ping @xemul

bhalevy:
re-ping @xemul can we please merge this PR?

bhalevy:
Replaced by https://github.com/scylladb/seastar/pull/2597

PR #2531, opened 2024-11-08
reactor: Always retry waitpid

michael-redpanda:
A possible issue with how Docker Desktop v4.34+ (particulary on macOS) handles `pidfd` functionality causes an assertion failure in Seastar's `reactor::waitpid` method.

The issue:

- Seastar creates a `pidfd` using `pidfd_open`
- It polls this file descriptor until it becomes readable
- When readable, call `waitpid` with `WNOHANG`.  According to [pidfd_open](https://man7.org/linux/man-pages/man2/pidfd_open.2.html), the file descriptor becomes readable when the process terminates
- Seastar expects `waitpid` to either return a positive integer representing the child process that has ended or a negative value to indicate an error.  However if `waitpid` returns `0`, then Seastar asserts, crashing the using application.

This change adds `reactor::do_waitpid` which loops on calling `waitpid` until it returns a non-zero value.  This method is called both when a `pidfd` is in use and when it isn't.  The assertion is also removed.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2075,6 +2075,32 @@ static auto next_waitpid_timeout(std::chrono::milliseconds this_timeout) {
 
 #endif
 
+future<int> reactor::do_waitpid(pid_t pid) {
+    return do_with(int{}, std::chrono::milliseconds(0), [pid, this](int& wstatus,
+                                                                    std::chrono::milliseconds& wait_timeout) {
+        return repeat_until_value([this,
+                                   pid,
+                                   &wstatus,
+                                   &wait_timeout] {
+            return _thread_pool->submit<syscall_result<pid_t>>([pid, &wstatus] {
+                return wrap_syscall<pid_t>(::waitpid(pid, &wstatus, WNOHANG));
+            }).then([&wstatus, &wait_timeout](syscall_result<pid_t> ret) mutable {
+                if (ret.result == 0) {
+                    wait_timeout = next_waitpid_timeout(wait_timeout);
+                    return ::seastar::sleep(wait_timeout).then([] {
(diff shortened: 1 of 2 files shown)


src/core/reactor.cc
```diff
@@ -2075,6 +2075,32 @@ static auto next_waitpid_timeout(std::chrono::milliseconds this_timeout) {
 
 #endif
 
+future<int> reactor::do_waitpid(pid_t pid) {
+    return do_with(int{}, std::chrono::milliseconds(0), [pid, this](int& wstatus,
```
avikivity, line 2079: Please use coroutines in new code.

avikivity: Ah, you're just moving old code around. ok.

src/core/reactor.cc
```diff
@@ -2083,39 +2109,11 @@ future<int> reactor::waitpid(pid_t pid) {
...
-                        }
-                    });
-                });
-            });
+            return do_waitpid(pid);
         } else {
-            return do_with(pollable_fd(file_desc::from_fd(pidfd.result)), int{}, [pid, this](auto& pidfd, int& wstatus) {
-                return pidfd.readable().then([pid, &wstatus, this] {
-                    return _thread_pool->submit<syscall_result<pid_t>>([pid, &wstatus] {
-                        return wrap_syscall<pid_t>(::waitpid(pid, &wstatus, WNOHANG));
```
avikivity, line 2113: Presumably, using _thread_pool here (and above) is wrong, since WNOHANG means the kernel won't block.

src/core/reactor.cc
bhalevy: https://github.com/scylladb/seastar/pull/2532

PR #2532, opened 2024-11-10
Coroutinize waitpid

bhalevy:
Commit fb6c969ccf2f7735fdb43e3f9beac8b68711e29a simplified
waitpid by always retrying, even if pidfd_open failed.

Coroutinizing this path (and folding do_waitpid back into
its only caller: waitpid) simplifies the code even further
and consolidate the different do_with allocations into a single
coroutine frame.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2075,49 +2075,44 @@ static auto next_waitpid_timeout(std::chrono::milliseconds this_timeout) {
 
 #endif
 
-future<int> reactor::do_waitpid(pid_t pid) {
-    return do_with(int{}, std::chrono::milliseconds(0), [pid, this](int& wstatus,
-                                                                    std::chrono::milliseconds& wait_timeout) {
-        return repeat_until_value([this,
-                                   pid,
-                                   &wstatus,
-                                   &wait_timeout] {
-            return _thread_pool->submit<syscall_result<pid_t>>([pid, &wstatus] {
-                return wrap_syscall<pid_t>(::waitpid(pid, &wstatus, WNOHANG));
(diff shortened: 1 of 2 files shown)


src/core/reactor.cc
```diff
@@ -2075,49 +2075,34 @@ static auto next_waitpid_timeout(std::chrono::milliseconds this_timeout) {
...
+    // older kernels. But it could be other error like EMFILE or ENFILE. anyway, we
+    // should always waitpid().
+    std::optional<pollable_fd> pfd;
+    if (pidfd.result != -1) {
+        pfd.emplace(file_desc::from_fd(pidfd.result));
+        co_await pfd->readable();
+    }
+
+    std::chrono::milliseconds wait_timeout(0);
+    for (;;) {
```
avikivity, line 2092: Please don't code things as infinite loops, it means the reader has to reverse-engineer the termination condition.

bhalevy:
In [ee292c4](https://github.com/scylladb/seastar/commit/ee292c4d00de19a9104049f23d3d66ae3e82bbae):
- defined do_waitpid as coroutine lambda to make loop explicit about the retryable case when the result is not ready yet
- overall, the logic is clearer now

bhalevy:
- fixed spelling

src/core/reactor.cc
bhalevy: Fixed in [ee292c4](https://github.com/scylladb/seastar/commit/ee292c4d00de19a9104049f23d3d66ae3e82bbae)

bhalevy:
@scylladb/seastar-maint please review / consider merging

dawmd (approved):

PR #2533, opened 2024-11-11
core: reactor_config: add reserve_io_control_blocks

denesb:
This new option allows reserving the configured amount of IOCBs, so they are available to a side application running in parallell to the seastar application.
The use case is ScyllaDB and its tool applications, both ScyllaDB and its tool applications need IOCBs to run, so ScyllaDB has to be able to tell seastar to reserve some IOCBs for the tools as well.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3800,6 +3800,9 @@ reactor_options::reactor_options(program_options::option_group* parent_group)
     , max_networking_io_control_blocks(*this, "max-networking-io-control-blocks", 10000,
                 "Maximum number of I/O control blocks (IOCBs) to allocate per shard. This translates to the number of sockets supported per shard."
                 " Requires tuning /proc/sys/fs/aio-max-nr. Only valid for the linux-aio reactor backend (see --reactor-backend).")
+    , reserve_io_control_blocks(*this, "reserve-io-control-blocks", 0,
+                "Reserve this many IOCBs, so it is available to any side application that runs parallel to the seastar appliation."
+                " Takes precedence over --max-networking-io-control-blocks. Only valid for the linux-aio reactor backend (see --reactor-backend).")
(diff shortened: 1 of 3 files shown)


yaronkaikov:
Who should review this? @denesb

denesb:
> Who should review this? @denesb

@avikivity would be the best candidate, since he was part of the discussion from which this patch was born (https://github.com/scylladb/scylladb/issues/19185).
But strictly speaking any @scylladb/seastar-maint member can review it.

src/core/reactor.cc
```diff
@@ -4167,14 +4170,14 @@ void smp::log_aiocbs(log_level level, unsigned storage, unsigned preempt, unsign
...
-unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs)
+unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs, unsigned reserve_iocbs)
 {
     static unsigned constexpr storage_iocbs = reactor::max_aio;
     static unsigned constexpr preempt_iocbs = 2;
 
     auto aio_max_nr = read_first_line_as<unsigned>("/proc/sys/fs/aio-max-nr");
     auto aio_nr = read_first_line_as<unsigned>("/proc/sys/fs/aio-nr");
-    auto available_aio = aio_max_nr - aio_nr;
+    auto available_aio = aio_max_nr - aio_nr - reserve_iocbs;
```
avikivity, line 4180: Check for underflow here, a default aio_max_nr + high reserve_iocbs can confuse the logic.

denesb:
New in v2: ensure `available_aio` doesn't underflow.

src/core/reactor.cc
denesb: Done in v2.

PR #2535, opened 2024-11-13
Websocket server fixes

p12tic:
This PR fixes several issues in websocket frame parser. The most important one is that websocket parser didn't properly support partial packets. That is, the parser would fail if it received part of the websocket frame in one chunk of data and then received the rest of the frame separately.

The rest of fixes are general cleanups and fixes for bugs that can only be reproduced when using websocket parser in client mode (there will be more PRs in this area).

Unit tests have been added to better cover this area of the websocket parser.

The PR is best reviewed commit by commit.

### DIFF
--- a/src/websocket/server.cc
+++ b/src/websocket/server.cc
@@ -207,23 +207,27 @@ future<websocket_parser::consumption_result_t> websocket_parser::operator()(
     }
     if (_state == parsing_state::flags_and_payload_data) {
         if (_buffer.length() + data.size() >= 2) {
-            if (_buffer.length() < 2) {
-                size_t hlen = _buffer.length();
-                _buffer.append(data.get(), 2 - hlen);
-                data.trim_front(2 - hlen);
-                _header = std::make_unique<frame_header>(_buffer.data());
-                _buffer = {};
+            // _buffer.length() is less than 2 when entering this if body due to how
+            // the rest of code is structured. The else branch will never increase
+            // _buffer.length() to >=2 and other paths to this condition will always
+            // have buffer cleared.
+            assert(_buffer.length() < 2);
 
-                // https://datatracker.ietf.org/doc/html/rfc6455#section-5.1
-                // We must close the connection if data isn't masked.
-                if ((!_header->masked) ||
-                        // RSVX must be 0
-                        (_header->rsv1 | _header->rsv2 | _header->rsv3) ||
-                        // Opcode must be known.
-                        (!_header->is_opcode_known())) {
-                    _cstate = connection_state::error;
-                    return websocket_parser::stop(std::move(data));
-                }
+            size_t hlen = _buffer.length();
+            _buffer.append(data.get(), 2 - hlen);
+            data.trim_front(2 - hlen);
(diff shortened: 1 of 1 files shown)


src/websocket/server.cc
```diff
@@ -257,16 +257,28 @@ future<websocket_parser::consumption_result_t> websocket_parser::operator()(
         }
     }
     if (_state == parsing_state::payload) {
-        if (_payload_length > data.size()) {
-            _payload_length -= data.size();
-            remove_mask(data, data.size());
-            _result = std::move(data);
-            return websocket_parser::stop(buff_t(0));
+        if (data.size() < _remaining_payload_length) {
```
xemul, line 257: A not for ourselves:
the
```patch
-        if (_payload_length > data.size()) {
+       if (data.size() < _remaining_payload_length) {
```
change is idempotent, it changes nothing, but picks up with _payload_length rename to _remaining_payload_length

Correct?

src/websocket/server.cc
```diff
@@ -257,16 +257,28 @@ future<websocket_parser::consumption_result_t> websocket_parser::operator()(
...
+            _parsed_payload_length += data.size();
+            _remaining_payload_length -= data.size();
+            return websocket_parser::dont_stop();
         } else {
-            _result = data.clone();
-            remove_mask(_result, _payload_length);
-            data.trim_front(_payload_length);
-            _payload_length = 0;
+            if (_result.empty()) {
+                _result = data.clone();
```
xemul: Next patch changes this to be `data.share()`. Can this be share()d right here?

src/websocket/server.cc
p12tic: Yes.

src/websocket/server.cc
p12tic: I didn't want to mix too many changes in a single commit. If there is a regression, then it can be bisected with more granularity.

src/websocket/server.cc
```diff
@@ -207,23 +207,21 @@ future<websocket_parser::consumption_result_t> websocket_parser::operator()(
     }
     if (_state == parsing_state::flags_and_payload_data) {
         if (_buffer.length() + data.size() >= 2) {
-            if (_buffer.length() < 2) {
```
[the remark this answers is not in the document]
p12tic: Agreed.

include/seastar/websocket/server.hh
```diff
@@ -134,7 +134,8 @@ class websocket_parser {
     connection_state _cstate;
     sstring _buffer;
     std::unique_ptr<frame_header> _header;
-    uint64_t _payload_length;
+    uint64_t _remaining_payload_length = 0;
+    uint64_t _parsed_payload_length = 0;
```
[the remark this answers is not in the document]
p12tic: Agreed.

src/websocket/server.cc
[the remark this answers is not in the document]
p12tic: Assert and explaining comment has been added.

PR #2538, opened 2024-11-15
reactor-aio: improve iocbs allocation logs

tomershafir:
When the system does have minimal AIO capacity, but doesnt have enough for optimal network performance, log a note about the reserve and a CPU count decrease suggestion, like in the branch where the system doesnt satisfy the minimal requirements.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4192,7 +4192,16 @@ unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs
             seastar_logger.warn("");
             log_aiocbs(log_level::warn, storage_iocbs, preempt_iocbs, network_iocbs);
             seastar_logger.warn("");
-            seastar_logger.warn("For optimal network performance, set /proc/sys/fs/aio-max-nr to at least {}.", aio_nr + requested_aio);
+            seastar_logger.warn("For optimal network performance, set /proc/sys/fs/aio-max-nr to at least {}", aio_nr + requested_aio);
+            
+            if (reserve_iocbs) {
+                seastar_logger.warn(", with an added reserve of {} (requested via reserve_io_control_blocks config)", reserve_iocbs);
+            }
+
+            unsigned smp_count_recommended = available_aio / (storage_iocbs + preempt_iocbs + network_iocbs);
+            if (smp_count_recommended > 0) {
+                seastar_logger.warn(", or decrease the logical CPU count of the application to {}", smp_count_recommended);
+            }
         } else {
             std::string err = format("Your system does not satisfy minimum AIO requirements. "
                                      "Set /proc/sys/fs/aio-max-nr to at least {} (minimum) or {} (recommended for networking performance)",


src/core/reactor.cc
```diff
@@ -4192,7 +4192,16 @@ unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs
             seastar_logger.warn("");
             log_aiocbs(log_level::warn, storage_iocbs, preempt_iocbs, network_iocbs);
             seastar_logger.warn("");
-            seastar_logger.warn("For optimal network performance, set /proc/sys/fs/aio-max-nr to at least {}.", aio_nr + requested_aio);
+            seastar_logger.warn("For optimal network performance, set /proc/sys/fs/aio-max-nr to at least {}", aio_nr + requested_aio);
```
tchaikov, line 4195: @lersek could you please help review this change? i think ec5da7a606ddf9f561996f3e8b7f64a72dbc883b is good enough. but you might have different insights.

tomershafir: If you mean the specific line, then it just removes the dot from the end to allow append.

lersek: @tchaikov This is a difficult review. Here's my opinion:
1. Precursor commit c5f9e3a234499 didn't extend `smp::adjust_max_networking_aio_io_control_blocks` correctly, in my opinion. That commit silently decreased `available_aio`. Consequences:
   - The debug message `Available AIO control blocks = aio-max-nr - aio-nr` is now stale.
   - `reserve_iocbs` itself is not logged in any way. 
   - The message `with an added reserve of ...` undermines the original goal: namely, to give the user *exact numbers* for storing into `aio-max-nr`.

   I recommend reworking c5f9e3a234499 from the ground up (in `smp::adjust_max_networking_aio_io_control_blocks`) to satisfy that goal again:
   - `available_aio` should continue meaning `aio_max_nr - aio_nr`
   - `reserve_iocbs` should be included in `requested_aio_other` (that is, in "everything requested except network")
   - `log_aiocbs` should be extended to log `reserve_iocbs` (separately, because it is not a per-cpu reservation)
   - The `added reserve` warning should be reverted.
   - The `smp_count_max` and `smp_count_recommended` calculations need to account for `reserve_iocbs` explicitly.

   The point is that `reserve_iocbs` must be a part of `requested_aio_other`. Instead of decreasing `available_aio`, we need to increase `requested_aio_other`.
   ~~~
   diff --git a/src/core/reactor.cc b/src/core/reactor.cc
   index a110602ac788..5a906d66415e 100644
   --- a/src/core/reactor.cc
   +++ b/src/core/reactor.cc
   @@ -4168,53 +4168,50 @@ void smp::log_aiocbs(log_level level, unsigned storage, unsigned preempt, unsign
    unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs, unsigned reserve_iocbs)
    {
        static unsigned constexpr storage_iocbs = reactor::max_aio;
        static unsigned constexpr preempt_iocbs = 2;
    
        auto aio_max_nr = read_first_line_as<unsigned>("/proc/sys/fs/aio-max-nr");
        auto aio_nr = read_first_line_as<unsigned>("/proc/sys/fs/aio-nr");
        auto available_aio = aio_max_nr - aio_nr;
   -    available_aio -= std::min(available_aio, reserve_iocbs);
        auto requested_aio_network = network_iocbs * smp::count;
   -    auto requested_aio_other = (storage_iocbs + preempt_iocbs) * smp::count;
   +    auto requested_aio_other = (storage_iocbs + preempt_iocbs) * smp::count + reserve_iocbs;
        auto requested_aio = requested_aio_network + requested_aio_other;
    
        seastar_logger.debug("Intended AIO control block usage:");
        seastar_logger.debug("");
   -    log_aiocbs(log_level::debug, storage_iocbs, preempt_iocbs, network_iocbs);
   +    log_aiocbs(log_level::debug, storage_iocbs, preempt_iocbs, network_iocbs, reserve_iocbs);
        seastar_logger.debug("");
        seastar_logger.debug("Available AIO control blocks = aio-max-nr - aio-nr = {} - {} = {}", aio_max_nr, aio_nr, available_aio);
    
        if (available_aio < requested_aio) {
            if (available_aio >= requested_aio_other + smp::count) { // at least one queue for each shard
                network_iocbs = (available_aio - requested_aio_other) / smp::count;
                seastar_logger.warn("Your system does not have enough AIO capacity for optimal network performance; reducing `max-networking-io-control-blocks'.");
                seastar_logger.warn("Resultant AIO control block usage:");
                seastar_logger.warn("");
   -            log_aiocbs(log_level::warn, storage_iocbs, preempt_iocbs, network_iocbs);
   +            log_aiocbs(log_level::warn, storage_iocbs, preempt_iocbs, network_iocbs, reserve_iocbs);
                seastar_logger.warn("");
                seastar_logger.warn("For optimal network performance, set /proc/sys/fs/aio-max-nr to at least {}.", aio_nr + requested_aio);
            } else {
                std::string err = format("Your system does not satisfy minimum AIO requirements. "
                                         "Set /proc/sys/fs/aio-max-nr to at least {} (minimum) or {} (recommended for networking performance)",
                                         aio_nr + (requested_aio_other + smp::count), aio_nr + requested_aio);
    
   -            if (reserve_iocbs) {
   -                err.append(format(", with an added reserve of {} (requested via reserve_io_control_blocks config)", reserve_iocbs));
   -            }
   -            unsigned smp_count_max = available_aio / (storage_iocbs + preempt_iocbs + 1);
   +            unsigned available_percpu_total = available_aio >= reserve_iocbs ? available_aio - reserve_iocbs : 0;
   +            unsigned smp_count_max = available_percpu_total / (storage_iocbs + preempt_iocbs + 1);
                if (smp_count_max > 0) {
                    err.append(format(", or decrease the logical CPU count of the application to {} (maximum)", smp_count_max));
    
   -                unsigned smp_count_recommended = available_aio / (storage_iocbs + preempt_iocbs + network_iocbs);
   +                unsigned smp_count_recommended = available_percpu_total / (storage_iocbs + preempt_iocbs + network_iocbs);
                    if (smp_count_recommended > 0) {
                        err.append(format(" or {} (recommended for networking performance)", smp_count_recommended));
                    }
                }
    
                err.append(".");
                throw std::runtime_error(err);
            }
        }
    
        return network_iocbs;
    }
   ~~~
   @denesb please comment! If you wouldn't like to implement the above, I'd be happy to. Thanks.
2. The patch that is the subject of this PR is incorrect for multiple reasons, IMO:
   - It adds a line which contains nothing but whitespace.
   - It duplicates the message `added reserve` which I've just claimed was wrong to introduce in the first place.
   - It does not send a single warning message -- concatenated from parts -- to the logger; instead, it logs the message broken down to multiple logger calls.
   - As far as I can tell, the goal of the PR is to advise on the CPU count reduction, as a means for increasing the network IOCBs, even in case the available AIO count can be made work by decreasing the network IOCBs. This idea looks sensible to me. However, the implementation has two problems IMO, in the `smp_count_recommended` calculation:
     - The `network_iocbs` used in the denominator is already the reduced value, not the originally requested value!
     - The calculation should not be duplicated between both branches, but should be extracted to a helper function (or at least lambda).

denesb: @lersek I think it is better if you implement this. You have a clear vision in mind for how this diagnostics should work.

tomershafir: Im gonna close this PR, and let @lersek re-implement this in the new PR.

PR #2540, opened 2024-11-15
sharded.hh: remove inline from templates

tomershafir:
Templates are distinctly defined in each translation unit, thus they dont cause multiple definitions linkage error, so inline is redundant code to compile.


tchaikov (approved):
lgtm

PR #2541, opened 2024-11-21
json_formatter: Add support for standard range containers

tchaikov:
This commit extends the JSON formatting capabilities to support all standard range containers through the C++20 ranges concept.

Key improvements:
- Add generic range support using std::ranges::input_range concept
- Enable JSON formatting for any range type without materialization
- Maintain existing support for vector, map, and unordered_map
- Provide a flexible, type-agnostic serialization approach

Example use cases:
- Formatting std::span
- Serializing range views
- Converting filter and transform views directly to JSON

### DIFF
--- a/include/seastar/json/formatter.hh
+++ b/include/seastar/json/formatter.hh
@@ -22,7 +22,9 @@
 #pragma once
 
 #ifndef SEASTAR_MODULE
+#include <ranges>
 #include <string>
+#include <tuple>
 #include <vector>
 #include <unordered_map>
 #include <map>
@@ -36,6 +38,33 @@
 #include <seastar/util/modules.hh>
 
 namespace seastar {
+
+namespace internal {
+
+template<typename T>
+concept is_map = requires {
+    typename T::mapped_type;
+};
+
+template<typename T>
+concept is_pair_like = requires {
+    typename std::tuple_size<T>::type;
+    requires std::tuple_size_v<T> == 2;
+};
+
+template<typename T>
+concept is_string_like =
+    std::convertible_to<const T&, std::string_view> &&
+    requires (T s) {
+        { s.data() } -> std::same_as<char*>;
+        // sstring::length() and sstring::find() return size_t instead of
+        // size_type (i.e., uint32_t), so we cannot check their return type
+        // with T::size_type
+        s.find('a');
+        s.length();
+    };
+
+}
 
 namespace json {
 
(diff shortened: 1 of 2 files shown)


tchaikov:
the httpd test is broken with this change.

include/seastar/json/formatter.hh
```diff
@@ -191,23 +220,19 @@ public:
...
-    static sstring to_json(const std::map<Args...>& map) {
-        return to_json(state::map, map.begin(), map.end());
-    }
-
-    template<typename... Args>
-    static sstring to_json(const std::unordered_map<Args...>& map) {
-        return to_json(state::map, map.begin(), map.end());
+    template<std::ranges::input_range Range>
+    requires (!internal::is_string_like<Range>)
+    static sstring to_json(const Range& range) {
```
StephanDollberg, line 230: You'll want to do the same for `formatter::write` such that the zero copy methods get the same support.

tchaikov:
v2:

- add `is_pair_like` concept to consolidate the formatter specialized for `std::tuple<K, V>` and the one specialized for `std::pair<K, V>`.
- tighten the constraints of `is_string_like`, and add comment to explain why we don't add constraints to the return types of `length()` and `find()`.
- add support to `formatter::write()` as well.

include/seastar/json/formatter.hh
tchaikov: @StephanDollberg hi Stephan, thank you for your review. added the support to `formatter::write()` in the latest revision.

include/seastar/json/formatter.hh
```diff
@@ -85,31 +115,58 @@ class formatter {
         return to_json(t);
     }
 
-    template<typename K, typename V>
-    static future<> write(output_stream<char>& stream, state s, const std::pair<K, V>& p) {
+    template<typename T>
+    requires (internal::is_pair_like<std::decay_t<T>> &&
+              std::is_rvalue_reference_v<T>)
+    static future<> write(output_stream<char>& stream, state s, T p) {
```
xemul, line 121: Why can't this look like

```c++
template<typename T>
requires (internal::is_pair_like T>
static future<> write(output_stream<char>& stream, state s, T&& p) {
    ...
```
?

tchaikov: no reasons. i was trying to consolidate these two variants of `write(...)` overloads for `is_pair_like` types, but ended up giving up on this idea, because it hurts the readability, and went with partially repeating them.

and yeah,  `T&& p` is way better. thank you! i switched to the proposed function signature in the latest revision.

tchaikov:
@xemul hi Pavel, could you please help re-review this change?

PR #2542, opened 2024-11-21
Fix dpdk support when using custom prefix

p12tic:
Currently Seastar manages building of DPDK internally and includes the built DPDK library to within Seastar library. Unfortunately when Seastar is installed to a custom prefix, it starts to require an installation of DPDK to that prefix. This does not make sense, as Seastar shouldn't require anything related to DPDK after installation.

This PR addresses the problem. The root cause is that the fact that Seastar uses DPDK leaks to installed CMake targets. First, Seastar is fixed to not search for installed DPDK once Seastar has been installed itself. Second, the dependencies have been adjusted so that the dependency on DPDK library does not transfer to users of Seastar library.

This allows Seastar with DPDK enabled being installed in custom prefix.


p12tic:
The first commit exposes a problem that may be interesting. Some variables have same names across different cmake files. This is problematic because the Seastar project recommends to not set initial value of cmake variables and rely on the default being empty. At least I got this feedback in some of my older PRs. This assumption does not hold if some other file sets the variable to non-empty value beforehand.

Perhaps it makes sense to invert the policy. Setting variables to empty results in additional code, which is small, stable inefficiency. On the other hand, accidentally being hit by cmake variables being overridden in non-obvious ways may waste hours or days of time or even more, if the failure manifests in non-obvious way in different part of codebase.

cmake/Finddpdk.cmake
```diff
@@ -153,7 +153,7 @@ if (dpdk_FOUND AND NOT (TARGET dpdk))
   if(_warning_supported_volatile)
     # include/generic/rte_spinlock.h increments volatiled-qualified type with
     # "++". but this is deprecated by GCC, so silence it.
-    set(compile_options
+    set(dpdk_compile_options
```
tchaikov, line 156: we should not tolerate this name clashing. so i created #2548 to address this.

CMakeLists.txt
```diff
@@ -875,9 +875,15 @@ else ()
 endif ()
 
 if (Seastar_DPDK)
+  # dpdk is included to within seastar, accordingly dpdk should not appear in
```
tchaikov, line 878: > dpdk is included to within seastar

might want reword this like: 

```cmake
# Although Seastar uses DPDK internally, DPDK symbols should not be exposed to
# Seastar users. However, DPDK's dependencies (like numa, pthread) must still
# be included in Seastar's INTERFACE_LINK_LIBRARIES to ensure proper linking
# of applications using Seastar.
```

cmake/SeastarDependencies.cmake
```diff
@@ -103,6 +104,12 @@ macro (seastar_find_dependencies)
     ucontext
     yaml-cpp)
 
+  # Only expose dpdk dependency during build. Since dpdk is included to within seastar library,
+  # there is no need to force it to be discoverable via find_package().
+  if (args_BUILD)
+    list (APPEND _seastar_all_dependencies dpdk)
+  endif()
```
tchaikov, line 111: i'd like to understand your use case better. do you enable `Seastar_DPDK` when building Seastar?

p12tic: I build and use Seastar using the following:

```
./3rdparty/seastar/configure.py --mode=release --enable-dpdk --prefix=$(pwd)/build-install
ninja -C ./3rdparty/seastar/build/release install

...

mkdir -p build
cmake -S . -B build -GNinja \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_PREFIX_PATH=$(pwd)/build-install \
    -DCMAKE_INSTALL_PREFIX=$(pwd)/build-install
ninja -C build
```

I would say this is pretty unconventional setup except for the install prefix and that DPDK is enabled. Without this PR build will fail because there's no dpdk installed in build-install (which is expected)

CMakeLists.txt
p12tic: Applied verbatim, thanks.

cmake/Finddpdk.cmake
p12tic: Dropped this commit after https://github.com/scylladb/seastar/pull/2548 landed.

p12tic:
@tchaikov Thanks for review. I think I've addressed your comments.

CMakeLists.txt
```diff
@@ -875,9 +875,16 @@ else ()
...
+  # Seastar users. However, DPDK's dependencies (like numa, pthread) must still
+  # be included in Seastar's INTERFACE_LINK_LIBRARIES to ensure proper linking
+  # of applications using Seastar.
   target_link_libraries (seastar
     PRIVATE
-      DPDK::dpdk)
+      "$<BUILD_INTERFACE:DPDK::dpdk>")
+  get_target_property(DPDK_INTERFACE_LINK_LIBRARIES DPDK::dpdk INTERFACE_LINK_LIBRARIES)
+  target_link_libraries (seastar
+    PRIVATE "${DPDK_INTERFACE_LINK_LIBRARIES}")
```
tchaikov, line 887: could you use something like https://github.com/scylladb/seastar/blob/665fed057d6ce08de4a78659403203bcce79b6cb/CMakeLists.txt#L867-L875 ?

so we don't forget dropping the workaround for older versions of CMake when we bump up the minimal required version of CMake ?

cmake/SeastarDependencies.cmake
tchaikov: @p12tic how do you build DPDK? do you build a static build of DPDK using something like `cooking.sh` ? i cannot find `--cook dpdk` in your command line arguments passed to `configure.py` .

but in a more unconventional setup, user might want to build Seastar with a dynamic build of DPDK libraries. in that case, we would have to find dpdk when building seastar *and* when building the seastar application. in this case, we cannot skip the `find_package(dpdk)` call when building seastar application, i think.

PR #2546, opened 2024-11-22
net/tls: Add a way to disable certificate validation

p12tic:
This feature is useful when testing TLS client with a server that does not have proper certificates setup yet.

The API to control this has been added to the certificate_credentials class because an instance of that class is already accepted by functions that may want to configure TLS. This way the requirement of additional set of overloads is avoided.

Controlling certificate validation is not completely unrelated to what certificate_credentials already does. It effectively controls which server certificetes the client is going to accept. Calling set_enable_certificate_verification(true) effectively adds all possible certificates to the set of accepted certificates. set_priority_string() already offers a similar control on algorithm quality.


elcallio (approved):
Let's hope no one uses this.

xemul:
Conflicts, please rebase (@p12tic )

p12tic:
@xemul Rebased, thanks!

Issue #799, opened 2020-09-22
tls::server does not wait for the future returned by data_sink::close() and data_source::close() to resolve


tgrabiec:
Not a problem with the current posix stack, because implementations return ready futures, but it's a violation of the API and can break in the future resulting in use-after-free.

PR #2547, opened 2024-11-22
net/tls: Wait for data_{source,sink}::close()

p12tic:
Fixes https://github.com/scylladb/seastar/issues/799

data_{source,sink}::close() return a future. If it is not ready on close() return, then the current tls session close() may result in use after free.

Converting close_after_shutdown() to a coroutine and sequentially co_awaiting on close() addresses this issue. The waiting is done sequentially, as this is shutdown path anyway.


p12tic:
Rebased and removed accidentally added whitespace changes

avikivity:
Can you add a test that fails before the patch and succeeds after it?

PR #2548, opened 2024-11-25
build: Clean up `compile_option` after FindSanitizers module

tchaikov:
Unset the `compile_option` variable after it's used in the FindSanitizers module to prevent potential conflicts with other CMake build configurations. Previously, this variable persisted after `find_package(Sanitizers)`, which could lead to unexpected build behavior if the same variable name was used elsewhere.

This change follows CMake best practices where Find modules clean up temporary variables after use.


p12tic:
Not part of the team, but LGTM.

avikivity:
> Not part of the team, but LGTM.

All reviews are welcome.

PR #2549, opened 2024-11-25
json2code: convert boost integer_range to std iota_view

avikivity:
Reduce dependencies on boost.

json2code converts a Swagger json API definition to C++ code. Remove a use of boost integer_range by converting it to the equivalent iota_view. Some notes:

 - we must return `auto`, otherwise concept evaluation on the class being defined fail
 - we must define an operator-(), also to satisfy concept evaluation for iota_view.


tchaikov (approved):
lgtm

PR #2550, opened 2024-11-25
utils/program-options: selection_value: make get_candidate_names() public

denesb:
To allow aplication code query available candidates, before making a choice.


include/seastar/util/program-options.hh
```diff
@@ -534,7 +534,7 @@ private:
     size_t _selected_candidate = no_selected_candidate;
     bool _defaulted = true;
 
-private:
+public:
```
tchaikov, line 537: just a nit in the commit message, s/aplication/application/

PR #2551, opened 2024-11-25
treewide: Trim trailing spaces

p12tic:
Trimming trailing spaces is an accepted best practice.

The practical benefit is that contributors with editors configured to always trim trailing spaces will not need to add an exception for seastar.

If maintainers want I could look into adding a CI action to prevent future inclusion of trailing spaces.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -147,7 +147,7 @@ find_package (Seastar ${VERSION} REQUIRED)
 
 add_executable (my_program
   my_program.cc)
-  
+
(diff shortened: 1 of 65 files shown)


p12tic:
@avikivity What do you think about adding trailing whitespace check to the CI?

travisdowns:
You are my hero for the day.

avikivity:
> @avikivity What do you think about adding trailing whitespace check to the CI?

I think it's a great idea

PR #2552, opened 2024-11-25
Implement experimental websocket client

p12tic:
This PR consists of 2 parts:
 - general refactoring to make it easier for Websocket client to reuse existing code
 - implementation of the Websocket client itself

It probably makes sense to land the general refactorings in smaller PRs once Websocket client is approved in principle.

One way that the websocket client differs from what is prescribed by RFC 6455 is that the client masking key is not random. The practical reason for that is to avoid needing to pull randomization infrastructure into the Websocket client. This would need additional APIs to control, as it makes sense to turn off randomization in unit tests.

Below is an excerpt from commit description on why this is reasonable solution

```
https://datatracker.ietf.org/doc/html/rfc6455#section-5.3 requires that
the masking key must be unpredictable and derived from a strong source
of entropy. This requirement arose due to usage of WebSocket protocol in
the browsers, where both server and client payload generation code may
be malicious and the only trusted piece is the browser itself.
Consequently there was a need to make the bytes on the wire
unpredictable for the client code, so that it cannot run attacks against
intermediate proxies that do not understand WebSocket.

In the case of Seastar there is no security boundary between payload
generator and payload serializer in this class, accordingly in terms of
security impact it is sufficient to simply use predictable masking key.
Zero is chosen because it does not change the payload. 
```


include/seastar/websocket/common.hh
```diff
@@ -160,5 +162,122 @@ public:
...
+    queue <temporary_buffer<char>> _output_buffer;
+    output_stream<char> _output;
+
+    sstring _subprotocol;
+    handler_t _handler;
+public:
+    /*!
+     * \param fd established socket used for communication
+     */
+    connection(connected_socket&& fd)
```
xemul: Shouldn't it be protected as well?

include/seastar/websocket/common.hh
```diff
@@ -243,16 +243,19 @@ protected:
 
     sstring _subprotocol;
     handler_t _handler;
+    bool _is_client;
```
xemul, line 250: It seems that `std::optional<uint32_t> _masking_key` would fit better here. Constructor and send_key() would need to be updated accordingly

src/websocket/common.cc
```diff
@@ -168,6 +172,26 @@ future<> connection::send_data(opcodes opcode, temporary_buffer<char>&& buff) {
         header[1] = uint8_t(buff.size());
     }
 
+    temporary_buffer<char> write_buf;
```
xemul, line 179: Unused in this patch. Maybe it's used later, but then it should as well be added "later"

include/seastar/websocket/common.hh
```diff
@@ -87,7 +87,7 @@ struct frame_header {
     }
```
xemul, line 87: Patch is missing descriptive comment

include/seastar/http/request.hh
```diff
@@ -313,10 +313,10 @@ struct request {
      */
```
xemul, line 313: What for? Should be in the patch comment

CMakeLists.txt
```diff
@@ -777,6 +777,7 @@ add_library (seastar
   src/util/read_first_line.cc
```
xemul, line 777: Patch comment is missing

CMakeLists.txt
```diff
@@ -777,6 +777,7 @@ add_library (seastar
...
   src/util/tmp_file.cc
   src/util/short_streams.cc
+  src/websocket/client.cc
```
xemul, line 780: Test is missing as well

include/seastar/websocket/client.hh
```diff
@@ -0,0 +1,140 @@
...
+     * \brief serve WebSocket protocol on a client_connection
+     */
+    future<> process();
+
+    /**
+     * @brief Send a websocket message to the server
+     */
+    future<> send_message(temporary_buffer<char> buf, bool flush);
+
+protected:
```
xemul, line 62: Why no private? Are there plans to implement more specific connections inherited from this class?

include/seastar/http/request.hh
p12tic: Thanks, I added comment in https://github.com/scylladb/seastar/pull/2579.

include/seastar/websocket/server.hh
```diff
@@ -25,149 +25,17 @@
 #include <functional>
 
 #include <seastar/http/request_parser.hh>
-#include <seastar/core/seastar.hh>
 #include <seastar/core/sstring.hh>
 #include <seastar/net/api.hh>
 #include <seastar/core/gate.hh>
-#include <seastar/core/queue.hh>
 #include <seastar/core/when_all.hh>
+#include <seastar/websocket/common.hh>
```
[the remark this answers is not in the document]
p12tic: Fixed in https://github.com/scylladb/seastar/pull/2582.

JaydenFish:
PixPin_2025-05-09_21-56-28

PR #2396, opened 2024-08-21
reactor, linux-aio: advise users in more detail on setting `aio-max-nr`

lersek:
Print a table of the AIO counts we intend to use (at "info" level).

If the intended counts cannot be allocated, reduce the network AIO count like before, but now also print an updated table (as a warning). Tell the user the specific number they should write to `/proc/sys/fs/aio-max-nr` for avoiding the reduction (assuming their current "aio-nr" does not change).

If the per CPU network AIO count would have to be reduced under 1, format a clear error message with (up to) four specific recommendations:

- `aio-max-nr` value for reaching 1 network AIOCB per CPU (always printed),
- `aio-max-nr` value for reaching the originally requested network AIOCB per CPU (always printed),
- the highest value to reduce the application's CPU count to, in order to reach 1 network AIOCB per CPU, within the current `(aio-max-nr - aio-nr)` budget (printed only if this SMP count is at least 1),
- the value to reduce the application's CPU count to, in order to reach the originally requested network AIOCB per CPU, within the current `(aio-max-nr - aio-nr)` budget (printed only if this SMP count is at least 1).

Motivated by <https://github.com/scylladb/scylladb/issues/5981>.

PR #2553, opened 2024-11-26
reactor, linux-aio: clean up `smp::adjust_max_networking_aio_io_control_blocks`

lersek:
`smp::adjust_max_networking_aio_io_control_blocks` has become a bit incohesive in the process of adding IOCB reservation; clean up the function.

Triggered by #2538.

See also #2396 (commit ec5da7a606dd) and #2533 (commit c5f9e3a23449).

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4143,7 +4143,7 @@ public:
     }
 };
 
-void smp::log_aiocbs(log_level level, unsigned storage, unsigned preempt, unsigned network) {
+void smp::log_aiocbs(log_level level, unsigned storage, unsigned preempt, unsigned network, unsigned reserve) {
     // Each cell in the table should be
     // - as wide as the grand total,
     // - as wide as its containing column's header,
@@ -4151,13 +4151,14 @@ void smp::log_aiocbs(log_level level, unsigned storage, unsigned preempt, unsign
     std::string percpu_hdr = format("per cpu");
     std::string allcpus_hdr = format("all {} cpus", smp::count);
     unsigned percpu_total = storage + preempt + network;
-    unsigned allcpus_total = percpu_total * smp::count;
+    unsigned allcpus_total = reserve + percpu_total * smp::count;
     size_t num_width = format("{}", allcpus_total).length();
     size_t percpu_width = std::max(num_width, percpu_hdr.length());
     size_t allcpus_width = std::max(num_width, allcpus_hdr.length());
 
     seastar_logger.log(level, "purpose  {:{}}  {:{}}",     percpu_hdr,   percpu_width, allcpus_hdr,          allcpus_width);
     seastar_logger.log(level, "-------  {:-<{}}  {:-<{}}", "",           percpu_width, "",                   allcpus_width);
+    seastar_logger.log(level, "reserve  {:{}}  {:{}}",     "",           percpu_width, reserve,              allcpus_width);
     seastar_logger.log(level, "storage  {:{}}  {:{}}",     storage,      percpu_width, storage * smp::count, allcpus_width);
     seastar_logger.log(level, "preempt  {:{}}  {:{}}",     preempt,      percpu_width, preempt * smp::count, allcpus_width);
     seastar_logger.log(level, "network  {:{}}  {:{}}",     network,      percpu_width, network * smp::count, allcpus_width);
@@ -4173,14 +4174,13 @@ unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs
     auto aio_max_nr = read_first_line_as<unsigned>("/proc/sys/fs/aio-max-nr");
     auto aio_nr = read_first_line_as<unsigned>("/proc/sys/fs/aio-nr");
     auto available_aio = aio_max_nr - aio_nr;
-    available_aio -= std::min(available_aio, reserve_iocbs);
     auto requested_aio_network = network_iocbs * smp::count;
-    auto requested_aio_other = (storage_iocbs + preempt_iocbs) * smp::count;
+    auto requested_aio_other = reserve_iocbs + (storage_iocbs + preempt_iocbs) * smp::count;
     auto requested_aio = requested_aio_network + requested_aio_other;
 
     seastar_logger.debug("Intended AIO control block usage:");
     seastar_logger.debug("");
-    log_aiocbs(log_level::debug, storage_iocbs, preempt_iocbs, network_iocbs);
+    log_aiocbs(log_level::debug, storage_iocbs, preempt_iocbs, network_iocbs, reserve_iocbs);
     seastar_logger.debug("");
     seastar_logger.debug("Available AIO control blocks = aio-max-nr - aio-nr = {} - {} = {}", aio_max_nr, aio_nr, available_aio);
 
@@ -4190,7 +4190,7 @@ unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs
             seastar_logger.warn("Your system does not have enough AIO capacity for optimal network performance; reducing `max-networking-io-control-blocks'.");
             seastar_logger.warn("Resultant AIO control block usage:");
             seastar_logger.warn("");
-            log_aiocbs(log_level::warn, storage_iocbs, preempt_iocbs, network_iocbs);
+            log_aiocbs(log_level::warn, storage_iocbs, preempt_iocbs, network_iocbs, reserve_iocbs);
             seastar_logger.warn("");
             seastar_logger.warn("For optimal network performance, set /proc/sys/fs/aio-max-nr to at least {}.", aio_nr + requested_aio);
         } else {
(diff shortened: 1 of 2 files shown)


lersek:
@tomershafir please check this out. Thanks!

denesb (approved):

src/core/reactor.cc
```diff
@@ -4185,6 +4185,10 @@ unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs
     seastar_logger.debug("Available AIO control blocks = aio-max-nr - aio-nr = {} - {} = {}", aio_max_nr, aio_nr, available_aio);
 
     if (available_aio < requested_aio) {
+        auto get_smp_count = [ linear_over_cpus = available_aio - std::min(reserve_iocbs, available_aio) ] (unsigned network) -> unsigned {
```
tchaikov, line 4188: nit, might want to remove the spaces around the capture list. to find the coding style when it comes to lambda:
```console
$ git grep -E '=\ +\[' | head -n10
apps/iotune/iotune.cc:                auto accuracy_msg = [accuracy, &rates] {
apps/memcached/ascii.rl:key = [^ ]+ >mark %{ _key = memcache::item_key(str()); };
apps/memcached/ascii.rl:        auto str = [this, &g, &p] { g.mark_end(p); return get_str(); };
apps/memcached/tests/test.py:        cmdline = [DIR_PATH + '/test_memcached.py'] + cmd
apps/memcached/tests/test_ascii_parser.cc:auto for_each_fragment_size = [] (auto&& func) {
apps/rpc_tester/rpc_tester.cc:            _call = [this] (unsigned x) { return call_echo(x); };
apps/rpc_tester/rpc_tester.cc:            _call = [this, payload = std::move(payload)] (unsigned x) { return call_write(x, payload); };
apps/rpc_tester/rpc_tester.cc:            _call = [this] (unsigned x) {
```

src/core/reactor.cc
```diff
@@ -4190,13 +4190,19 @@ unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs
...
             seastar_logger.warn("");
             log_aiocbs(log_level::warn, storage_iocbs, preempt_iocbs, network_iocbs, reserve_iocbs);
             seastar_logger.warn("");
-            seastar_logger.warn("For optimal network performance, set /proc/sys/fs/aio-max-nr to at least {}.", aio_nr + requested_aio);
+            std::string advice = format("For optimal network performance, set /proc/sys/fs/aio-max-nr to at least {}", aio_nr + requested_aio);
+            if (smp_count_recommended > 0) {
+                advice.append(format(", or decrease the logical CPU count of the application to {}", smp_count_recommended));
+            }
+            advice.append(".");
+            seastar_logger.warn("{}", advice);
```
tchaikov, line 4205: nit, despite that warning messages are always enabled, we could avoid constructing a `std::string` using something like:

```c++
            seastar_logger.warn("For optimal network performance, set /proc/sys/fs/aio-max-nr to at least {}"
                    "{}"
                    ".",
                    aio_nr + requested_aio,
                    smp_count_recommended > 0
                        ? ", or decrease the logical CPU count of the application to " + std::to_string(smp_count_recommended)
                        : "");
```

src/core/reactor.cc
```diff
@@ -4173,39 +4174,45 @@ unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs
...
     seastar_logger.debug("");
     seastar_logger.debug("Available AIO control blocks = aio-max-nr - aio-nr = {} - {} = {}", aio_max_nr, aio_nr, available_aio);
 
     if (available_aio < requested_aio) {
+        auto get_smp_count = [ linear_over_cpus = available_aio - std::min(reserve_iocbs, available_aio) ] (unsigned network) -> unsigned {
+            return linear_over_cpus / (storage_iocbs + preempt_iocbs + network);
+        };
+
         if (available_aio >= requested_aio_other + smp::count) { // at least one queue for each shard
+            unsigned smp_count_recommended = get_smp_count(network_iocbs);
```
tomershafir, line 4193: Maybe we can move this line before the conditional for both branches.

avikivity:
Why is this good advice? It reduces performance.

There's an easy way to fix the problem without reducing performance.

src/core/reactor.cc
lersek: Thanks! Will do so in v2.

src/core/reactor.cc
lersek: Good observation; I'd thought about this question myself. I have two counter-arguments for this specific proposal:
- If we do something like this, then we need to modify both branches.
- I find the proposed update, as-is, harder to read than what's in the patch (and pre-patch, on the other branch). String concatenation with `+` and `std::to_string` is awkard; I find it harder to read than format strings. (Of course, with this argument, I'm slightly questioning my own code as well, which uses `std::string::append()` for concatenation, not a grand format string...)

What if we replace
~~~
", or decrease the logical CPU count of the application to " + std::to_string(smp_count_recommended)
~~~
with
~~~
format(", or decrease the logical CPU count of the application to {}", smp_count_recommended)
~~~
?

(If that's what we end up doing, then I should probably rework the preexistent "append logic" on the exception-throwing branch, too.)

src/core/reactor.cc
lersek: @tomershafir Yes, I had considered it. The reason I didn't do it is that, on the exception-throwing branch, if `smp_count_max==0`, then `smp_count_recommended` is never defined / computed. Hoisting the definition and the `get_smp_count(network_iocbs)` call above the conditional slightly pessimizes the exception-throwing branch (we'd be computing something that might never be needed).

tomershafir:
Well, the idea was introduced in https://github.com/scylladb/seastar/pull/2396, I only thought to align the the other branch as the same conditions hold. 

But yes, I also see it as a possible tradeoff between CPU capacity and network I/O capacity. A user may not want to increase `aio-max-nr` because its a global settings which may have other effects on the environment. A simplified example can be:

When `available_aio == 12 && **smp::count == 3** && storage_iocbs == 1 && preemt_iocbs == 2 && network_iocbs == 2` then there is 1 aio_cb allocated for network on each core.
But when `available_aio == 12 && **smp::count == 2** && storage_iocbs == 1 && preemt_iocbs == 2 && network_iocbs == 2` then there are 2 aio_cbs allocated for network on each core.

However, it seems less relevant for dedicated HPC seastar apps.

lersek:
@avikivity what's your verdict about patch#3? thanks.

avikivity:
> @avikivity what's your verdict about patch#3? thanks.

I don't have a strong opinion on it. IMO the time is better spend on reducing the limitations than writing mountains of explanations about them.

tomershafir:
Sounds reasonable to me (even though I think reducing limitations can be hard or not needed).

In that case, I suggest to also remove this logic from the original branch.

lersek:
@tomershafir

> In that case, I suggest to also remove this logic from the original branch.

FWIW, recommending a CPU count reduction was not introcuced in #2396 / commit ec5da7a606dd (contrary to your [claim](https://github.com/scylladb/seastar/pull/2553#issuecomment-2501358590)). The goal of that patch was to give precise, directly actionable numbers to the user. The patch didn't invent new ideas, only quantified previous ideas. The code that predated ec5da7a606dd would already emit the message *Try ... reducing the amount of logical CPUs available for your application*, and it would do so only on the exception-throwing branch. Up to and including this PR, there has been no change to that approach.

The CPU count reduction language *in this particular context* goes back to commit 2a3b4ac62865. However, if we're to consider [other error contexts,](https://github.com/scylladb/scylladb/issues/9096), then the same language dates back to commit 8c6cf573e51f even. And, in fact, the same language *persists* in said other context, to this day, most recently touched up by commit d4468c1c1fb5.

Thus, I disagree with the removal of the CPU count reduction language, especially if we remove it _only_ from `smp::adjust_max_networking_aio_io_control_blocks`. That would still leave the other instance in place (in `setup_aio_context()`).

My proposal (alternatives):
* Either merge this series in full,
* or:
  * let us first merge patches 1-2 from this series,
  * and then, Tomer, please submit a separate PR for removing the CPU count reduction recommendation from *both* contexts.

tomershafir:
Sounds good

src/core/reactor.cc
lersek: Done in v2.

src/core/reactor.cc
lersek: Resolved in v2 -- this code was part of v1 patch 3, but in v2, I've dropped patch 3.

src/core/reactor.cc
lersek: Not changed in v2 (I think my argument was solid, and anyway, you'll probably rework or eliminate that code in your upcoming PR).

lersek:
v2:
- drop patch 3
- patch 2:
  - remove leading and trailing space within the capture block [Kefu @tchaikov]
  - update commit message (no third patch, no additional use for lambda)

tomershafir:
lgtm

tomershafir:
@lersek I suggest to edit the PR description

lersek:
@tomershafir 

> @lersek I suggest to edit the PR description

Thanks a lot for pointing that out! Done now.

(It's another facet of github that I'm struggling to get used to. When posting a patch set to a mailing list, it's nearly impossible to forget about the cover letter, and the subject of the cover letter. The force-pushing on github makes it difficult to write up the inter-version changelog (which should be easy), and makes it easy to forget about the cover letter (which should be hard). Go github! Anyway, I'll try to remember it next time.)

tchaikov (approved):
lgtm

lersek:
@scylladb/scylla-maint can you please merge this? Thank you.

tchaikov:
> @scylladb/scylla-maint can you please merge this? Thank you.

@lersek scylla and seastar do not share the same group of maintainers. so we need to `@scylladb/seastar-maint` for seastar related requests.

lersek:
@tchaikov 

> @lersek scylla and seastar do not share the same group of maintainers. so we need to `@scylladb/seastar-maint` for seastar related requests.

Thanks for the correction!

lersek:
@scylladb/seastar-maint can you please merge this? thanks.

lersek:
@scylladb/seastar-maint @tgrabiec @nyh @avikivity @xemul can you please merge this? thanks.

PR #2554, opened 2024-11-26
Minor metrics memory optimizations

StephanDollberg:
The metrics subsystem is quite a heavy memory user at high metric counts. 

This PR features two smaller improvements reducing memory usage by removing some redundant storage. I have a higher impact change following that internalizes labels.

Two commits (details in commit):
 - Don't store full metric_id in the metric_groups_impl
 - Store impl SP in the metric_groups_impl and not each registered_metric

### DIFF
--- a/include/seastar/core/metrics_api.hh
+++ b/include/seastar/core/metrics_api.hh
@@ -202,7 +202,6 @@ class impl;
 class registered_metric final {
     metric_info _info;
     metric_function _f;
-    shared_ptr<impl> _impl;
 public:
     registered_metric(metric_id id, metric_function f, bool enabled=true, skip_when_empty skip=skip_when_empty::no);
     metric_value operator()() const {
@@ -240,8 +239,9 @@ using metrics_registration = std::vector<register_ref>;
 
 class metric_groups_impl : public metric_groups_def {
     metrics_registration _registration;
+    shared_ptr<impl> _impl; // keep impl alive while metrics are registered
 public:
-    metric_groups_impl() = default;
+    metric_groups_impl();
     ~metric_groups_impl();
     metric_groups_impl(const metric_groups_impl&) = delete;
     metric_groups_impl(metric_groups_impl&&) = default;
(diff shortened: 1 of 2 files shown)


StephanDollberg:
@amnonh

include/seastar/core/metrics_api.hh
```diff
@@ -240,8 +239,9 @@ using metrics_registration = std::vector<register_ref>;
 
 class metric_groups_impl : public metric_groups_def {
     metrics_registration _registration;
+    shared_ptr<impl> _impl; // keep impl alive while metrics are registered
 public:
-    metric_groups_impl() = default;
+    metric_groups_impl();
```
amnonh, line 244: This looks like a leftover from some mid-code changes; the contractor still uses the default implementation.

src/core/metrics.cc
```diff
@@ -294,13 +294,22 @@ std::unique_ptr<metric_groups_def> create_metric_groups() {
...
+
 metric_groups_impl::~metric_groups_impl() {
     for (const auto& i : _registration) {
         unregister_metric(i->info().id);
     }
 }
 
 metric_groups_impl& metric_groups_impl::add_metric(group_name_type name, const metric_definition& md)  {
+    // We could just do this in the constructor but some metric groups (like the
+    // smp queue ones) are actually constructed on a different shard originally
```
avikivity, line 307: That's a bug. Of course, you're not required to fix it as part of the series.

StephanDollberg:
> I think a more significant impact will be using some kind of String Interning (like a dictionary or shard ptr) for labels and their values, but that is a much bigger change (especially as it requires handling the lifecycle of the labels)

Yes, I do actually have a related patch for that as well which makes a big difference for us. I will put it up for review shortly.

include/seastar/core/metrics_api.hh
StephanDollberg: Yes you are right. I had this at first when I tried initing the impl SP in the constructor directly but then had to move it because of the bug mentioned in the comment.

PR #2560, opened 2024-11-27
scripts/perftune.py: Implement AWS IMDSv2 call

### LINKED ISSUE #1051: perftune.py should switch to AWS IMDSv2 calls
perftune's [__check_host_type](https://github.com/scylladb/seastar/blob/41be7cc6d3bdf99df21adb6508e62354134b331a/scripts/perftune.py#L549) method uses IMDSv1 for retrieving instance metadata. It turns out that - as described in https://github.com/scylladb/scylla/issues/10490 - Ec2 instances may have only [IMDSv2](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/) HTTP calls allowed.

When requests to IMDSv1 fail, calling `is_aws_i3_non_metal_instance()` will evaluate to `False`, which means that i3.nonmetal instance types will get tuned as if they weren't i3 instances.

syuu1228:
perftune's __check_host_type method uses IMDSv1 for retrieving instance metadata.
It turns out that - as described in scylladb/scylladb#10490 - Ec2 instances may have only IMDSv2 HTTP calls allowed.

When requests to IMDSv1 fail, calling is_aws_i3_non_metal_instance() will evaluate to False, which means that i3.nonmetal instance types will get tuned as if they weren't i3 instances.

To fix the problem, import IMDSv2 implementation from scylla-machine-image.

Fixes #1051
related scylladb/scylla-machine-image#498
related scylladb/scylladb#10490


scripts/perftune.py
```diff
@@ -332,9 +333,20 @@ def auto_detect_irq_mask(cpu_mask, cores_per_irq_core):
...
         return run_hwloc_calc(['--restrict', cpu_mask] + hwloc_args)
 
+# @param headers dict of k:v
+def curl(url, headers=None, method=None, byte=False, timeout=0):
+    req = urllib.request.Request(url, headers=headers or {}, method=method)
+    with urllib.request.urlopen(req, timeout=timeout) as res:
+        if byte:
+            return res.read()
+        else:
+            return res.read().decode('utf-8')
```
avikivity, line 343: Seems like "curl" is adding very little. You can just call urrlib.request.urlopen() in the caller, no? Just add .decode() when byte=False.

scripts/perftune.py
```diff
@@ -549,13 +563,27 @@ def _get_irqs(self):
         pass
 
 #### Private methods ############################
+    def __refresh_metadata_token(self):
```
vladzcloudius, line 566: Another confusing function name: you are reading a token here in order to use it later - not "refreshing" it.

`__generate_metadata_token` is a more appropriate name IMO.

scripts/perftune.py
```diff
@@ -549,13 +563,27 @@ def _get_irqs(self):
...
+        self._metadata_token_time = datetime.datetime.now()
+        self._metadata_token = curl(self.EC2_METADATA_BASE_URL + "api/token", headers={"X-aws-ec2-metadata-token-ttl-seconds": self.EC2_METADATA_TOKEN_TTL}, method="PUT", timeout=0.1)
```
vladzcloudius, line 568: Why these two are "protected" while the only place where you use them are in the `__instance_metadata(...)` below.

Unless I'm missing something they should be "private" together with the rest of the code.

vladzcloudius (changes requested):

scripts/perftune.py
```diff
@@ -332,9 +333,20 @@ def auto_detect_irq_mask(cpu_mask, cores_per_irq_core):
...
+        if byte:
+            return res.read()
+        else:
+            return res.read().decode('utf-8')
 
 ################################################################################
 class PerfTunerBase(metaclass=abc.ABCMeta):
+    EC2_METADATA_BASE_URL = "http://169.254.169.254/latest/"
+    EC2_METADATA_TOKEN_TTL = 21600
+
```
vladzcloudius, line 349: Please, move these into a method that uses it or make these a "property".

scripts/perftune.py
```diff
@@ -549,13 +563,27 @@ def _get_irqs(self):
...
+        return curl(self.EC2_METADATA_BASE_URL + path, headers={"X-aws-ec2-metadata-token": self._metadata_token}, timeout=0.1)
+
     def __check_host_type(self):
         """
         Check if we are running on the AWS i3 nonmetal instance.
         If yes, set self.__is_aws_i3_nonmetal_instance to True, and to False otherwise.
         """
         try:
-            aws_instance_type = urllib.request.urlopen("http://169.254.169.254/latest/meta-data/instance-type", timeout=0.1).read().decode()
+            aws_instance_type = self.__instance_metadata('meta-data/instance-type')
```
vladzcloudius, line 586: Can it be anything other than `meta-data/...` for a metadata?

If you put this part of the URL inside the `def __instance_metadata(...)` as in my previous post the call will become:

```
aws_instance_type = self.__instance_metadata('instance-type')
```

Which is cleaner IMO.

scripts/perftune.py
```diff
@@ -549,13 +563,27 @@ def _get_irqs(self):
...
+
+    def __instance_metadata(self, path):
+        if not self._metadata_token:
+            self.__refresh_metadata_token()
+        else:
+            time_diff = datetime.datetime.now() - self._metadata_token_time
+            time_diff_sec = int(time_diff.total_seconds())
+            if time_diff_sec >= self.EC2_METADATA_TOKEN_TTL - 120:
+                self.__refresh_metadata_token()
+        return curl(self.EC2_METADATA_BASE_URL + path, headers={"X-aws-ec2-metadata-token": self._metadata_token}, timeout=0.1)
```
vladzcloudius, line 578: Why not to hide all this "token refresh" logic (plus the memoization) inside the what is now `__refresh_metadata_token(...)`:
* Rename the method to be a property called `__metadata_token(self)`
* Put the memoization and refresh code inside the property code.

In such a case reading the metadata method is going to look at follows (I also patch it following my next comment below):

```
def __instance_metadata(self, path):
    return curl(self.EC2_METADATA_BASE_URL + "meta-data/" + path, headers={"X-aws-ec2-metadata-token": self.__metadata_token}, timeout=0.1)
```

vladzcloudius (changes requested):

scripts/perftune.py
syuu1228: Fixed on latest push

scripts/perftune.py
```diff
@@ -332,9 +333,20 @@ def auto_detect_irq_mask(cpu_mask, cores_per_irq_core):
 
         return run_hwloc_calc(['--restrict', cpu_mask] + hwloc_args)
 
+# @param headers dict of k:v
```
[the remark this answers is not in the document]
syuu1228: Fixed on latest push

scripts/perftune.py
syuu1228: It was because of just copied the code from scylla-machine-image, fixed on the latest push

scripts/perftune.py
syuu1228: The other thing is `userdata/`, so it cannot be put `meta-data/` part into `__instance_metadata()` in scylla-machine-image (original code).
But in perftune.py we can just do it since we probably never reference userdata.

Fixed on the latest push.

scripts/perftune.py
syuu1228: Fixed on the latest push.

scripts/perftune.py
syuu1228: Renamed this method to `__metadata_token(self)` to make it property:
https://github.com/scylladb/seastar/pull/2560#discussion_r1861306581

scripts/perftune.py
syuu1228: Moved ttl to `__metadata_token(self)`, and make base_url to a property.

scripts/perftune.py
```diff
@@ -549,13 +552,48 @@ def _get_irqs(self):
...
+        return "http://169.254.169.254/latest/"
+
+    @property
+    def __metadata_token(self):
+        """
+        Refresh IMDSv2 session token if it necessary, and return current token
+        :return: current session token
+        """
+        token_ttl = 21600
+        update_token = False
```
vladzcloudius, line 566: `update_token` doesn't seem to be used for anything but assignment.
Unless I missed the place where you read this value, you should be able to safely drop this variable.

scripts/perftune.py
```diff
@@ -549,13 +552,48 @@ def _get_irqs(self):
...
+            update_token = True
+        else:
+            time_diff = datetime.datetime.now() - self.__metadata_token_time
+            time_diff_sec = int(time_diff.total_seconds())
+            if time_diff_sec >= token_ttl - 120:
+                update_token = True
+        self.__metadata_token_time = datetime.datetime.now()
+        req = urllib.request.Request(self.__ec2_metadata_base_url + "api/token", headers={"X-aws-ec2-metadata-token-ttl-seconds": token_ttl}, method="PUT")
+        with urllib.request.urlopen(req, timeout=0.1) as res:
+            self.__metadata_token_value = res.read().decode()
```
vladzcloudius, line 577: You seem to be unconditionally reloading token value because you don't use the value of `update_token` as I believe you intended.

vladzcloudius (changes requested):
Something is very strange about the `__metadata_token` code.

scripts/perftune.py
syuu1228: Sorry I broke the code when I move metadata token related code to __metadata_token() method.
Pushed fixed version.

scripts/perftune.py
vladzcloudius: If we needed `userdata` you could have implemented a corresponding ,method: `__instance_userdata`.

vladzcloudius (approved):
Looks good.
Thanks.

PR #2561, opened 2024-11-28
metrics: Internalize metric label sets

StephanDollberg:
The major memory user in the metric subsystem is the per series label
sets.

This is for two reasons:
 - The labels are stored multiple times in different places: Twice as
   part of `registered_metric` (one current labels and one original one
   for relabeling - these are often just the same), one copy as the
   lookup key in the values map and another final copy in the metadata
   copy for scraping
 - Lots of the labels are actually duplicated across metric series. For
   example for one io-queue class/mountpoint/iogroup combination you get
   15 series all with the same labels. Similar is true for other metrics
   like reactor or scheduler metrics. In Redpanda we have an even more
   extreme pattern where per partition metrics create more than 60
   series with the same label set.

One idea to improve this would be to intern the individual label
strings. In theory this would give improvements on top of the
duplication described above. Most label keys and values are duplicated
even if not all labels are the same (e.g.: shard).

On a more detailed look this however wouldn't be very effective. Most
labels are fairly short, e.g. looking at some seastar and redpanda
metrics:

```
io_queue...{class="raft",iogroup="0",mountpoint="none",shard="0"}
redpanda...{namespace="kafka",partition="9985",shard="2",topic="foobar"}
```

Each of the label key and values are shorter than 16 bytes and hence fit
into the seastar short string optimization. Hence, each string only
takes 16 bytes. In the above example maybe one of the eight strings
might leave the SSO length (e.g.: mountpoint and topic) but those are in
the minority.

Interning the string in less than 16 bytes is not trivial so the gain
would actually not be that large (deletion needs to be thread safe as
the metadata might be copied to different shards at scrape time).

There is further hidden overhead. Labels are stored in a `std::map` for
fast lookup. Each node in the map has a ~32 byte overhead for the node
struct. So even at a zero overhead internalisation we could only save
about 2x in the most likely case.

Hence this patch takes the strategy of internalizing the whole labels
set (`std::map`). This means that we also internalize map overhead and
still get an overall reduction as described above. For the actual
internalization we store the labels `std::map` in a `std::shared_ptr`
(for cross shared safety). It's stored in a `std::set` for lookup when
adding new metrics.

In Redpanda this patch reduces the metrics memory usage by a
factor of 3-4 and hence saving gigabytes of memory at high partition
counts (yes we likey have too many metrics).

This also makes `update_metrics_if_needed` faster as less copying is
needed and also increases cache locality at scrape time making that cheaper
as well (a yet to be committed microbench by Travis Downs shows a ~10%
improvement).

Once all other shared_ptr references have gone out of scope we need to
delete the entry out of the internalization set. We do this scan in
`update_metrics_if_needed`. The overhead is minimal as the set doesn't
actually get super big and extra overhead is compensated by
`update_metrics_if_needed` being cheaper as mentioned above. Ideally we
could do some deleter tricks but this is being complicated by thread
safety issues in case the last reference is not on the origin shard.

### DIFF
--- a/include/seastar/core/metrics_api.hh
+++ b/include/seastar/core/metrics_api.hh
@@ -363,7 +363,50 @@ public:
 
 using value_map = std::map<sstring, metric_family>;
 
-using metric_metadata_fifo = std::deque<metric_info>;
+/*!
+ * \brief Subset of the per series metadata that is shared via get_values to other shards.
+ *
+ * Allows omitting metadata that is already stored elsewhere or not needed by
+ * the metrics scrap handlers.
+ *
+ * Not copyable to allow for safely sharing internalized data.
+ */
+class metric_series_metadata {
+    // prom backend only needs the label from here but scollectd needs group and
+    // metric name separately. metric_family_info only stores the merged and
+    // filtered name so we have to duplicate it here.
+    metric_id _id;
+    skip_when_empty _should_skip_when_empty;
+public:
+    metric_series_metadata() = default;
+    metric_series_metadata(metric_id id, skip_when_empty should_skip_when_empty)
+        : _id(std::move(id)), _should_skip_when_empty(should_skip_when_empty) {
+    }
+
+    metric_series_metadata(const metric_series_metadata&) = delete;
+    metric_series_metadata& operator=(const metric_series_metadata&) = delete;
+
+    metric_series_metadata(metric_series_metadata&&) noexcept = default;
+    metric_series_metadata& operator=(metric_series_metadata&&) noexcept = default;
+
+    const labels_type& labels() const {
+      return _id.labels();
+    }
+
+    skip_when_empty should_skip_when_empty() const {
+        return _should_skip_when_empty;
+    }
+
+    group_name_type group_name() const {
+        return _id.group_name();
+    }
+
+    group_name_type name() const {
+        return _id.name();
+    }
+};
+
+using metric_metadata_fifo = std::deque<metric_series_metadata>;
 
 /*!
  * \brief holds a metric family metadata
@@ -380,7 +423,15 @@ using metric_metadata_fifo = std::deque<metric_info>;
 struct metric_family_metadata {
     metric_family_info mf;
     metric_metadata_fifo metrics;
+
+    metric_family_metadata() = default;
+    metric_family_metadata(metric_family_metadata &&) = default;
+    metric_family_metadata &operator=(metric_family_metadata &&) = default;
+    metric_family_metadata(metric_family_info mf, metric_metadata_fifo metrics)
+        : mf(std::move(mf)), metrics(std::move(metrics)) {}
 };
+
+static_assert(std::is_nothrow_move_assignable_v<metric_family_metadata>);
 
 using value_vector = std::deque<metric_value>;
 using metric_metadata = std::vector<metric_family_metadata>;
--- a/include/seastar/core/metrics_api.hh
+++ b/include/seastar/core/metrics_api.hh
@@ -23,6 +23,7 @@
 
 #include <seastar/core/metrics.hh>
 #include <seastar/util/modules.hh>
+#include <seastar/core/shared_ptr.hh>
 #include <seastar/core/sharded.hh>
 #ifndef SEASTAR_MODULE
 #include <boost/functional/hash.hpp>
@@ -42,6 +43,8 @@ namespace metrics {
 namespace impl {
 
 using labels_type = std::map<sstring, sstring>;
+using internalized_labels_ref = lw_shared_ptr<const labels_type>;
+
 }
 }
 }
@@ -106,7 +109,7 @@ class metric_id {
 public:
     metric_id() = default;
     metric_id(group_name_type group, metric_name_type name,
-                    labels_type labels = {})
+                    internalized_labels_ref labels)
                     : _group(std::move(group)), _name(
                                     std::move(name)), _labels(labels) {
     }
@@ -123,16 +126,19 @@ public:
         _group = name;
     }
     const instance_id_type & instance_id() const {
-        return _labels.at(shard_label.name());
+        return _labels->at(shard_label.name());
     }
     const metric_name_type & name() const {
         return _name;
     }
     const labels_type& labels() const {
+        return *_labels;
+    }
+    internalized_labels_ref internalized_labels() const {
         return _labels;
     }
-    labels_type& labels() {
-        return _labels;
+    void update_labels(internalized_labels_ref labels) {
+        _labels = labels;
     }
     sstring full_name() const;
 
@@ -144,7 +150,7 @@ private:
     }
     group_name_type _group;
     metric_name_type _name;
-    labels_type _labels;
+    internalized_labels_ref _labels;
 };
 }
 }
@@ -191,10 +197,43 @@ struct metric_family_info {
  */
 struct metric_info {
     metric_id id;
-    labels_type original_labels;
+    internalized_labels_ref original_labels;
     bool enabled;
     skip_when_empty should_skip_when_empty;
 };
+
+class internalized_holder {
+    internalized_labels_ref _labels;
+public:
+    explicit internalized_holder(labels_type labels) : _labels(make_lw_shared<labels_type>(std::move(labels))) {
+    }
+
+    explicit internalized_holder(internalized_labels_ref labels) : _labels(std::move(labels)) {
+    }
+
+    internalized_labels_ref labels_ref() const {
+        return _labels;
+    }
+
+    const labels_type& labels() const {
+        return *_labels;
+    }
+
+    size_t has_users() const {
+        // Getting the count wrong isn't a correctness issue but will just make internalization worse
+        return _labels.use_count() > 1;
+    }
+};
+
+inline bool operator<(const internalized_holder& lhs, const labels_type& rhs) {
+    return lhs.labels() < rhs;
+}
+inline bool operator<(const labels_type& lhs, const internalized_holder& rhs) {
+    return lhs < rhs.labels();
+}
+inline bool operator<(const internalized_holder& lhs, const internalized_holder& rhs) {
+    return lhs.labels() < rhs.labels();
+}
 
 
 class impl;
@@ -234,7 +273,7 @@ public:
 };
 
 using register_ref = shared_ptr<registered_metric>;
-using metric_instances = std::map<labels_type, register_ref>;
+using metric_instances = std::map<internalized_holder, register_ref, std::less<>>;
 using metrics_registration = std::vector<register_ref>;
 
 class metric_groups_impl : public metric_groups_def {
@@ -268,12 +307,12 @@ public:
     metric_family(metric_instances&& instances) : _instances(std::move(instances)) {
     }
 
-    register_ref& operator[](const labels_type& l) {
-        return _instances[l];
+    register_ref& operator[](const internalized_labels_ref& l) {
+        return _instances[internalized_holder(l)];
     }
 
-    const register_ref& at(const labels_type& l) const {
-        return _instances.at(l);
+    const register_ref& at(const internalized_labels_ref& l) const {
+        return _instances.at(internalized_holder(l));
     }
 
     metric_family_info& info() {
@@ -356,6 +395,8 @@ struct config {
     sstring hostname;
 };
 
+using internalized_set = std::set<internalized_holder, std::less<>>;
+
 class impl {
     value_map _value_map;
     config _config;
@@ -365,6 +406,7 @@ class impl {
     std::vector<std::deque<metric_function>> _current_metrics;
     std::vector<relabel_config> _relabel_configs;
     std::vector<metric_family_config> _metric_family_configs;
+    internalized_set _internalized_labels;
 public:
     value_map& get_value_map() {
         return _value_map;
@@ -375,6 +417,7 @@ public:
     }
 
     register_ref add_registration(const metric_id& id, const metric_type& type, metric_function f, const description& d, bool enabled, skip_when_empty skip, const std::vector<std::string>& aggregate_labels);
+    internalized_labels_ref internalize_labels(labels_type labels);
     void remove_registration(const metric_id& id);
     future<> stop() {
         return make_ready_future<>();
@@ -412,6 +455,10 @@ public:
     void set_metric_family_configs(const std::vector<metric_family_config>& metrics_config);
 
     void update_aggregate(metric_family_info& mf) const noexcept;
+
+private:
+    void gc_internalized_labels();
+    bool apply_relabeling(const relabel_config& rc, metric_info& info);
 };
 
 const value_map& get_value_map();
--- a/src/core/scollectd.cc
+++ b/src/core/scollectd.cc
@@ -177,12 +177,12 @@ struct cpwriter {
         return *this;
     }
 
-    sstring get_type_instance(const seastar::metrics::impl::metric_id & id) {
-        if (id.labels().empty()) {
-            return id.name();
+    sstring get_type_instance(const metrics::metric_name_type& name, const metrics::impl::labels_type& labels) {
+        if (labels.empty()) {
+            return name;
         }
-        sstring res = id.name();
-        for (auto i : id.labels()) {
+        sstring res = name;
+        for (auto i : labels) {
             if (i.first != seastar::metrics::shard_label.name()) {
                 res += "-" + i.second;
             }
@@ -288,7 +288,8 @@ struct cpwriter {
         }
         return *this;
     }
-    cpwriter & put(const sstring & host, const seastar::metrics::impl::metric_id & id, const type_id& type) {
+    cpwriter & put(const sstring & host, const metrics::group_name_type& group_name, const metrics::metric_name_type& name,
+            const metrics::impl::labels_type& labels, const type_id& type) {
         const auto ts = std::chrono::system_clock::now().time_since_epoch();
         const auto lrts =
                 std::chrono::duration_cast<std::chrono::seconds>(ts).count();
@@ -298,14 +299,15 @@ struct cpwriter {
         // Seems hi-res timestamp does not work very well with
         // at the very least my default collectd in fedora (or I did it wrong?)
         // Use lo-res ts for now, it is probably quite sufficient.
-        put_cached(part_type::Plugin, id.group_name());
+        put_cached(part_type::Plugin, group_name);
         // Optional
+        auto instance_id = labels.at(metrics::shard_label.name());
         put_cached(part_type::PluginInst,
-                id.instance_id() == per_cpu_plugin_instance ?
-                        to_sstring(this_shard_id()) : id.instance_id());
+                instance_id == per_cpu_plugin_instance ?
+                        to_sstring(this_shard_id()) : instance_id);
         put_cached(part_type::Type, type);
         // Optional
-        put_cached(part_type::TypeInst, get_type_instance(id));
+        put_cached(part_type::TypeInst, get_type_instance(name, labels));
         return *this;
     }
     cpwriter & put(const sstring & host,
@@ -313,7 +315,8 @@ struct cpwriter {
             const type_instance_id & id, const value_list & v) {
         const auto ps = std::chrono::duration_cast<collectd_hres_duration>(
                         period).count();
-            put(host, to_metrics_id(id), id.type());
+            auto mid = to_metrics_id(id);
+            put(host, mid.group_name(), mid.name(), mid.labels(), id.type());
             put(part_type::Values, v);
             if (ps != 0) {
                 put(part_type::IntervalHr, ps);
@@ -324,10 +327,11 @@ struct cpwriter {
     cpwriter & put(const sstring & host,
             const duration & period,
             const type_id& type,
-            const seastar::metrics::impl::metric_id & id, const seastar::metrics::impl::metric_value & v) {
+            const metrics::group_name_type& group_name, const metrics::metric_name_type& name,
+            const metrics::impl::labels_type& labels, const seastar::metrics::impl::metric_value & v) {
         const auto ps = std::chrono::duration_cast<collectd_hres_duration>(
                 period).count();
-        put(host, id, type);
+        put(host, group_name, name, labels, type);
         put(part_type::Values, v);
         if (ps != 0) {
             put(part_type::IntervalHr, ps);
@@ -361,7 +365,8 @@ future<> impl::send_metric(const type_instance_id & id,
 future<> impl::send_notification(const type_instance_id & id,
         const sstring & msg) {
     cpwriter out;
-    out.put(_host, to_metrics_id(id), id.type());
+    auto mid = to_metrics_id(id);
+    out.put(_host, mid.group_name(), mid.name(), mid.labels(), id.type());
     out.put(part_type::Message, msg);
     return _chan.send(_addr, net::packet(out.data(), out.size()));
 }
@@ -456,7 +461,8 @@ void impl::run() {
                     continue;
                 }
                 auto m = out.mark();
-                out.put(_host, _period, std::get<type_id>(*ctxt), md_iterator->id, *i);
+                out.put(_host, _period, std::get<type_id>(*ctxt),
+                    md_iterator->group_name(), md_iterator->name(), md_iterator->labels(), *i);
                 if (!out) {
                     out.reset(m);
                     out_of_space = true;
--- a/src/core/scollectd.cc
+++ b/src/core/scollectd.cc
@@ -44,6 +44,7 @@ module seastar;
 #include <seastar/core/metrics_api.hh>
 #include <seastar/core/byteorder.hh>
 #include <seastar/core/print.hh>
+#include <seastar/core/shared_ptr.hh>
 
 #include "core/scollectd-impl.hh"
 #endif
@@ -97,8 +98,9 @@ registration::registration(type_instance_id&& id)
 }
 
 seastar::metrics::impl::metric_id to_metrics_id(const type_instance_id & id) {
-    return seastar::metrics::impl::metric_id(id.plugin(), id.type_instance(),
-            {{seastar::metrics::shard_label.name(), seastar::metrics::impl::shard()}});
+    seastar::metrics::impl::labels_type labels {{seastar::metrics::shard_label.name(), seastar::metrics::impl::shard()}};
+    auto internalized_labels = make_lw_shared<seastar::metrics::impl::labels_type>(std::move(labels));
+    return seastar::metrics::impl::metric_id(id.plugin(), id.type_instance(), std::move(internalized_labels));
 }
 
 
@@ -575,7 +577,7 @@ options::options(program_options::option_group* parent_group)
 
 static seastar::metrics::impl::register_ref get_register(const scollectd::type_instance_id& i) {
     seastar::metrics::impl::metric_id id = to_metrics_id(i);
-    return seastar::metrics::impl::get_value_map().at(id.full_name()).at(id.labels());
+    return seastar::metrics::impl::get_value_map().at(id.full_name()).at(id.internalized_labels());
 }
 
 std::vector<collectd_value> get_collectd_value(
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -642,8 +642,8 @@ private:
         const auto& mf = values.find(m_name);
         assert(mf != values.end());
         for (auto&& mi : mf->second) {
-            auto&& cname = mi.first.find("class");
-            if (cname != mi.first.end() && cname->second == name()) {
+            auto&& cname = mi.first.labels().find("class");
+            if (cname != mi.first.labels().end() && cname->second == name()) {
                 out << YAML::Key << m_name << YAML::Value << mi.second->get_function()().d();
             }
         }
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -145,7 +145,7 @@ const std::vector<metric_family_config>& get_metric_family_configs() {
     return impl::get_local_impl()->get_metric_family_configs();
 }
 
-static bool apply_relabeling(const relabel_config& rc, impl::metric_info& info) {
+bool impl::impl::apply_relabeling(const relabel_config& rc, metric_info& info) {
     std::stringstream s;
     bool first = true;
     for (auto&& l: rc.source_labels) {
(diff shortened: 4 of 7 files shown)


amnonh:
I'm reviewing the code; so far, I just wanted some clarification. My concern is about what happens during metrics reporting when you can get preemptions. Maybe that's not an issue

StephanDollberg:
> I'm reviewing the code; so far, I just wanted some clarification. My concern is about what happens during metrics reporting when you can get preemptions. Maybe that's not an issue

Yeah that shouldn't be an issue (obviously let me know if you see one). The shared_ptr should avoid any thread safety issues and the SP type is const explicitly to avoid accidental modification once internalised.

amnonh (changes requested):
Overall, it looks good and takes a good approach. My comments are around code documentation; that would help to make it clearer.

I agree that, theoretically, we could have dropped the GC and based the deletion on the pointer self-counter, but it's not worth the extra complexity. We don't expect that metrics changing and deletion will happen a lot.

include/seastar/core/metrics_api.hh
```diff
@@ -234,7 +272,7 @@ public:
 };
 
 using register_ref = shared_ptr<registered_metric>;
-using metric_instances = std::map<labels_type, register_ref>;
+using metric_instances = std::map<internalized_holder, register_ref, std::less<>>;
```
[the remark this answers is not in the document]
StephanDollberg: It's needed yes. It enables heterogenuous lookupand hence we can use the extra operator< overloads defined for `internalized_holder`.

include/seastar/core/metrics_api.hh
```diff
@@ -268,12 +306,12 @@ public:
...
 
-    register_ref& operator[](const labels_type& l) {
-        return _instances[l];
+    register_ref& operator[](const internalized_labels_ref& l) {
+        return _instances[internalized_holder(l)];
     }
 
-    const register_ref& at(const labels_type& l) const {
-        return _instances.at(l);
+    const register_ref& at(const internalized_labels_ref& l) const {
```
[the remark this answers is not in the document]
StephanDollberg: Yes just using `labels_type` would be better here. The problem is that `std::map::at` only supports heterogenuous lookup as part of C++26 [1][2].

As a workaround we could just do it manually and do a `find` first and if not found throw a `std::out_of_range` exception ourselves?

[1] https://stackoverflow.com/a/75329629/893693
[2] https://en.cppreference.com/w/cpp/container/map/at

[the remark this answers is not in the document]
amnonh: What makes it more confusing is that you create an object implicitly by passing a const reference to a method that gets an object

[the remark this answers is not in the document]
StephanDollberg: Yes but this isn't really that different to what happens when you do:

```
std::map<std::string, int> m;
m.at("hello are you there");
```
That will also create and allocate a temporary `std::string` that is only used during lookup.

As suggested above we can work around it though by simply implementing `at` ourselves.

avikivity:
@amnonh can you measure the improvement for ScyllaDB?

include/seastar/core/metrics_api.hh
```diff
@@ -42,6 +42,8 @@ namespace metrics {
 namespace impl {
 
 using labels_type = std::map<sstring, sstring>;
+using internalized_labels_ref = std::shared_ptr<const labels_type>;
+
```
avikivity, line 46: BEGIN FROWN

include/seastar/core/metrics_api.hh
```diff
@@ -268,12 +313,12 @@ public:
     metric_family(metric_instances&& instances) : _instances(std::move(instances)) {
     }
 
-    register_ref& operator[](const labels_type& l) {
-        return _instances[l];
+    register_ref& operator[](const internalized_labels_ref& l) {
+        return _instances[internalized_holder(l)];
     }
```
avikivity, line 318: Does this change the public API?

avikivity:
Some comments:
 - I wasn't able to tell if interning would favor one shard over the others; if so we'd want to distribute interning across all shards to avoid skewing memory use.
 - How is the shard label ignored? Maybe it's added later so there's no need to do anything
 - I'd much prefer a foreign_ptr approach (especially if we hash-distribute interning)
 - A vector of pairs would offer even more compression, as long as the API is still std::map

StephanDollberg:
> I wasn't able to tell if interning would favor one shard over the others; if so we'd want to distribute interning across all shards to avoid skewing memory use.

> How is the shard label ignored? Maybe it's added later so there's no need to do anything

> I'd much prefer a foreign_ptr approach (especially if we hash-distribute interning)

Let me answer those three together. I probably spoke too much about threadsafety in the commit message and hence caused some confusion.

So in the current patch we don't actually do any cross-shard interning at this point. Hence the shard label gets interned as any other label (it gets added to the label set pretty early). 

The thread safety bit I was talking about comes from the metadata handling which gets copied across shards at scrape time. This is how an internalized label can live on two different shards temporarily. This is where we are already using SPs.

Using a foreign_ptr would be feasible but would make the API/types quite a bit more complicated.

Interning across shards is certainly another step that could be taken. Note that after this change the memory usage taken by the label set (per shard) as part of the overall metrics memory usage goes down 20-30% so it's not like further improving that bit has a massive effect. It would also require a change to the public API I think as that is currently not async. 

Further while it would probably help for reactor etc. metrics for user metrics it's likely that high cardinality series are probably different across shards in other labels than the shard. E.g.: in redpanda for the high cardinality series the partition would be different across shards so you wouldn't get any advantage there.

So it's not immediately clear the added complexity would be worth it.

> A vector of pairs would offer even more compression, as long as the API is still std::map

Yes this would be another memory optimization that could applied. Would need to see how it affects query time though. As mentioned above the improvements aren't massive at that point so this would need to be considered as scrapping is also quite expensive.

include/seastar/core/metrics_api.hh
avikivity: Ok.

avikivity:
> > I wasn't able to tell if interning would favor one shard over the others; if so we'd want to distribute interning across all shards to avoid skewing memory use.
> 
> > How is the shard label ignored? Maybe it's added later so there's no need to do anything
> 
> > I'd much prefer a foreign_ptr approach (especially if we hash-distribute interning)
> 
> Let me answer those three together. I probably spoke too much about threadsafety in the commit message and hence caused some confusion.
> 
> So in the current patch we don't actually do any cross-shard interning at this point. Hence the shard label gets interned as any other label (it gets added to the label set pretty early).
> 
> The thread safety bit I was talking about comes from the metadata handling which gets copied across shards at scrape time. This is how an internalized label can live on two different shards temporarily. This is where we are already using SPs.

I don't see std::shared_ptr in the metrics code now.

> 
> Using a foreign_ptr would be feasible but would make the API/types quite a bit more complicated.
> 


Could we, perhaps, capture the local owner of all the metrics and pass it around as a foreign_ptr? It would have to be immutable for that to work.

> Interning across shards is certainly another step that could be taken. Note that after this change the memory usage taken by the label set (per shard) as part of the overall metrics memory usage goes down 20-30% so it's not like further improving that bit has a massive effect. It would also require a change to the public API I think as that is currently not async.

We could (with non-negligible effort) have a fiber to intern after the fact.


> 
> Further while it would probably help for reactor etc. metrics for user metrics it's likely that high cardinality series are probably different across shards in other labels than the shard. E.g.: in redpanda for the high cardinality series the partition would be different across shards so you wouldn't get any advantage there.
> 
> So it's not immediately clear the added complexity would be worth it.
> 
> > A vector of pairs would offer even more compression, as long as the API is still std::map
> 
> Yes this would be another memory optimization that could applied. Would need to see how it affects query time though. As mentioned above the improvements aren't massive at that point so this would need to be considered as scrapping is also quite expensive.

I think it would improve scrape time too. We could add a private interface between metrics and prometheus that exposes the labels as a span. Operating on cold node-based data structures is slow.

Meanwhile, please see if you can drop the dependency on std::shared_ptr. I don't have a suggestion but I do want to prevent its encroachment.

StephanDollberg:
> I don't see std::shared_ptr in the metrics code now.

> Meanwhile, please see if you can drop the dependency on std::shared_ptr. I don't have a suggestion but I do want to prevent its encroachment.

Yes you are totally right of course. I always forget about the seastar version when I see an unqualified `shared_ptr`. I do see now that we already use a `foreign_ptr` in the metadata cross-shard flow. 

So in that case we can actually just replace the `std::shared_ptr` usage with `ss::lw_shared_ptr` (which is also a bit slimmer). All the uses are encompased by the object that is already being sent back via the foreign_ptr so all their destructors will run on the correct shard.

There is the risk of someone just copying out one of the `lw_shared_ptr`s but IMO this is well covered by the shared_ptr non-owner debug checks. E.g.: the `Seastar.unit.prometheus_http` catches this when intentionally introducing a bug here.

In any case I have added two further commits. The first just tightens up the API which makes these accidental copies a bit harder and more explicit (in fact this found an unneeded copy in the protobuf prom impl).

The second commit goes the full way and makes the cross-shard metadata code use individual foreign_ptrs. As suggested above this complicates the API quite a bit and is also more expensive in case the foreign references are the last copies. 

I'd be happy to drop the both of these commits or just the latter. Let me know what you think.

include/seastar/core/metrics_api.hh
```diff
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include "seastar/core/shared_ptr.hh"
```
avikivity, line 24: Please use <> includes, not "" includes.

include/seastar/core/metrics_api.hh
```diff
@@ -191,11 +197,44 @@ struct metric_family_info {
...
+};
+
+inline bool operator<(const internalized_holder& lhs, const labels_type& rhs) {
+    return lhs.labels() < rhs;
+}
+inline bool operator<(const labels_type& lhs, const internalized_holder& rhs) {
+    return lhs < rhs.labels();
+}
+inline bool operator<(const internalized_holder& lhs, const internalized_holder& rhs) {
+    return lhs.labels() < rhs.labels();
```
avikivity, line 282: These can be replaces with a transparent comparator (https://en.cppreference.com/w/cpp/container/map/find variants 3, 4) for the key.

But why use an ordered map in the first place?

include/seastar/core/metrics_api.hh
```diff
@@ -356,6 +395,8 @@ struct config {
     sstring hostname;
 };
 
+using internalized_set = std::set<internalized_holder, std::less<>>;
```
avikivity, line 451: This would be a candidate for unordered_set. But since you need ordered compares for the other container, we might as well keep this as is.

src/core/scollectd.cc
```diff
@@ -19,6 +19,7 @@
  * Copyright (C) 2014 Cloudius Systems, Ltd.
  */
 
+#include "seastar/core/shared_ptr.hh"
```
avikivity, line 22: \<> includes

include/seastar/core/metrics_api.hh
```diff
@@ -132,31 +134,48 @@ public:
...
         return std::tie(group_name(), instance_id(), name(), labels());
     }
     group_name_type _group;
     metric_name_type _name;
-    internalized_labels_ref _labels;
+    label_type _labels;
 };
+
+using metric_id = metric_id_t<internalized_labels_ref>;
+using foreign_metric_id = metric_id_t<foreign_ptr<internalized_labels_ref>>;
```
avikivity, line 173: I don't understand why we'd want a foreign metric_id. Isn't it too low-level? On remote shards we work with large collections of metrics, not individual metrics ids.

avikivity:
Regarding cross-shard safety, I think the best approach is to make shard data immutable and have metrics registration clone the data structure, install the metric, and replace the root pointer.

Caveat 1: I don't have a clear mental picture of how it currently works so it may be unnecessary/wrong
Caveat 2: cloning doesn't work during unregistration since it happens in a destructor, perhaps we can set an "invalid" flag and let the next non-noexcept call clean it up

StephanDollberg:
> Regarding cross-shard safety, I think the best approach is to make shard data immutable and have metrics registration clone the data structure, install the metric, and replace the root pointer.

Sorry I don't fully understand this. So internalized labels are immutable. There is no risk of anybody modifying them.

The "risk" I described in my previous message is someone (accidentally) copying the `lw_shared_ptr` out of the shared metadata structure when processing the metrics for scrape. This doesn't happen today. They are all part of a bigger structure which is already using foreign_ptr (`values_reference`) so effectively everything gets destroyed on the right shard.

I see three ways to address this potential risk:
 - Do nothing: This is debug-checked pretty well via the shared-ptr non-owner checks. 
 - Narrow down the API so that it's harder for someone to do that. From your other comment it sounds like you'd rather not do that. Though if we were to use different datastructures in the shared metadata then this might become a bit cleaner. We could remove the `copy` methods from `metric_info` and `metric_id` and have a completely separate uncopyable `foreign_metric` datastructure.
 - Put each individual labelset lw_shared_ptr into its own foreign_ptr. This would make copying of the shared ptr impossible (other than calling foreign_ptr::copy explicitly)

The current three commits reflect each of those three steps.

include/seastar/core/metrics_api.hh
avikivity: Not even per internalized label set, one foreign_ptr per the entire registered metric set.

avikivity:
> > Regarding cross-shard safety, I think the best approach is to make shard data immutable and have metrics registration clone the data structure, install the metric, and replace the root pointer.
> 
> Sorry I don't fully understand this. 


First a disclaimer that I'm not well versed in the metrics code so anything I say could be nonsense.

> So internalized labels are immutable. There is no risk of anybody modifying them.

I'm talking about immutablizing the entire metadata structure for all metrics. The idea is that metadata changes very rarely, so we can spend more time when modifying it to fully clone + add, then install the new structure. A foreign_ptr to that immutable structure is safe to traverse across shards.

> 
> The "risk" I described in my previous message is someone (accidentally) copying the `lw_shared_ptr` out of the shared metadata structure when processing the metrics for scrape. This doesn't happen today. They are all part of a bigger structure which is already using foreign_ptr (`values_reference`) so effectively everything gets destroyed on the right shard.

Doesn't values_reference only refer to values, not the metadata?

Maybe I'm wrong and it's already as I'm describing.

> 
> I see three ways to address this potential risk:
> 
> * Do nothing: This is debug-checked pretty well via the shared-ptr non-owner checks.
> * Narrow down the API so that it's harder for someone to do that. From your other comment it sounds like you'd rather not do that. Though if we were to use different datastructures in the shared metadata then this might become a bit cleaner. We could remove the `copy` methods from `metric_info` and `metric_id` and have a completely separate uncopyable `foreign_metric` datastructure.

If we were using Rust then it would make sense to do it. It also makes sense in C++, but in my experience removing copy-ability makes things worse elsewhere.

Since this is a dedicated type (not generic) and its use is heavily tied to cross-shard work, perhaps it does merit this special copy behavior.

We could rationalize this by saying the type has the semantics of unique_ptr (not copyable) and the sharing is an implementation detail and not exposed to its users.


> * Put each individual labelset lw_shared_ptr into its own foreign_ptr. This would make copying of the shared ptr impossible (other than calling foreign_ptr::copy explicitly)

I think this is overkill. foreign_ptr:s aren't free.


> 
> The current three commits reflect each of those three steps.

Thanks for the detailed explanation.

StephanDollberg:
> I'm talking about immutablizing the entire metadata structure for all metrics. The idea is that metadata changes very rarely, so we can spend more time when modifying it to fully clone + add, then install the new structure. A foreign_ptr to that immutable structure is safe to traverse across shards.

Gotcha, yes that is effectively what happens today (minus it being actually const but that's something we can add). 

> Doesn't values_reference only refer to values, not the metadata?

Yeah it does contain the metadata. The name is a bit confusing I guess :)

Relevant types:

```
using values_reference = shared_ptr<values_copy>;

struct values_copy {
    shared_ptr<metric_metadata> metadata;
    metric_values values;
};
```

> If we were using Rust then it would make sense to do it. It also makes sense in C++, but in my experience removing copy-ability makes things worse elsewhere.

> Since this is a dedicated type (not generic) and its use is heavily tied to cross-shard work, perhaps it does merit this special copy behavior.

> We could rationalize this by saying the type has the semantics of unique_ptr (not copyable) and the sharing is an implementation detail and not exposed to its users.

I didn't really like the whole `copy` stuff neither. Move only seems fine to me from C++ POV in this case. Let me give it a try.

avikivity:
> > I'm talking about immutablizing the entire metadata structure for all metrics. The idea is that metadata changes very rarely, so we can spend more time when modifying it to fully clone + add, then install the new structure. A foreign_ptr to that immutable structure is safe to traverse across shards.
> 
> Gotcha, yes that is effectively what happens today (minus it being actually const but that's something we can add).
> 
> > Doesn't values_reference only refer to values, not the metadata?
> 
> Yeah it does contain the metadata. The name is a bit confusing I guess :)
> 
> Relevant types:
> 
> ```
> using values_reference = shared_ptr<values_copy>;
> 
> struct values_copy {
>     shared_ptr<metric_metadata> metadata;
>     metric_values values;
> };
> ```
> 

Ah, and we generate it on-demand in update_metrics_if_needed(), rather than during construction.

travisdowns:
@avikivity - while we have your attention on metrics, what do you estimate the use fraction to be for scollectd vs prometheus for scylladb users? E.g., to help evaluate the effectiveness of changes that improve only the prometheus side.

avikivity:
> @avikivity - while we have your attention on metrics, what do you estimate the use fraction to be for scollectd vs prometheus for scylladb users? E.g., to help evaluate the effectiveness of changes that improve only the prometheus side.

0:1

StephanDollberg:
I have pushed a change now that uses a separate datastructure. I think this is a lot cleaner like this.

Unfortunately, we can't strip that much of it for now because of scollectd.

mykaul:
> I have pushed a change now that uses a separate datastructure. I think this is a lot cleaner like this.
> 
> Unfortunately, we can't strip that much of it for now because of scollectd.

Is anyone still using scollectd?

travisdowns:
> Is anyone still using scollectd?

I heard from Avi that no, zero people are using scollectd integration in seastar. Seems like a PR to remove support would probably be accepted easily then?

StephanDollberg:
@mykaul how is that coming along? I have created a PR here to remove it from seastar https://github.com/scylladb/seastar/pull/2586

mykaul:
> @mykaul how is that coming along? I have created a PR here to remove it from seastar #2586

https://github.com/scylladb/scylladb/pull/17356 - in the process of passing CI.

avikivity:
I saw build failures compiling scylladb with this series, but all from unit test that reach into the impl namespace.

bhalevy:
@avikivity / @scylladb/seastar-maint: this change breaks scylla.
I suggest to revert it and probably resubmit while bumping the api level.

Also, again, it's a good example why we should run scylla ci as part of seastar's ci
since scylla is a comprehensive representative of the the library users.

For example:
```
bhalevy@home:~/dev/scylla$ ninja dev-objects
[1/2] ./SCYLLA-VERSION-GEN --output-dir build
[2/5] NINJA build/dev/seastar/libseastar.so
ninja: Entering directory `build/dev/seastar'
ninja: no work to do.
[3/5] CXX build/dev/test/boost/query_processor_test.o
FAILED: build/dev/test/boost/query_processor_test.o 
clang++ -MD -MT build/dev/test/boost/query_processor_test.o -MF build/dev/test/boost/query_processor_test.o.d -std=gnu++23 -I/home/bhalevy/dev/scylla/seastar/include -I/home/bhalevy/dev/scylla/build/dev/seastar/gen/include -Werror=unused-result -fstack-clash-protection -DSEASTAR_API_LEVEL=7 -DSEASTAR_BUILD_SHARED_LIBS -DSEASTAR_SSTRING -DSEASTAR_ENABLE_ALLOC_FAILURE_INJECTION -DSEASTAR_LOGGER_COMPILE_TIME_FMT -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_LOGGER_TYPE_STDOUT -DSEASTAR_TYPE_ERASE_MORE -DFMT_SHARED -I/usr/include/p11-kit-1 -DWITH_GZFILEOP -g -ffile-prefix-map=/home/bhalevy/dev/scylla=. -DSEASTAR_NO_EXCEPTION_HACK -march=westmere -DDEVEL -DSEASTAR_ENABLE_ALLOC_FAILURE_INJECTION -DSCYLLA_ENABLE_ERROR_INJECTION -DSCYLLA_ENABLE_PREEMPTION_SOURCE -O2 -DSCYLLA_BUILD_MODE=dev -Xclang -fexperimental-assignment-tracking=disabled -iquote. -iquote build/dev/gen -std=gnu++23 -g -ffile-prefix-map=/home/bhalevy/dev/scylla=. -DSEASTAR_NO_EXCEPTION_HACK -march=westmere -DBOOST_ALL_DYN_LINK    -fvisibility=hidden -isystem abseil -Wall -Werror -Wextra -Wimplicit-fallthrough -Wno-mismatched-tags -Wno-c++11-narrowing -Wno-overloaded-virtual -Wno-unused-parameter -Wno-unsupported-friend -Wno-missing-field-initializers -Wno-deprecated-copy -Wno-enum-constexpr-conversion -Wno-error=deprecated-declarations -DXXH_PRIVATE_API -DSEASTAR_TESTING_MAIN  -c -o build/dev/test/boost/query_processor_test.o test/boost/query_processor_test.cc
test/boost/query_processor_test.cc:187:35: error: no member named 'id' in 'seastar::metrics::impl::metric_series_metadata'
  187 |         const auto found = metric.id.labels().find("consistency_level");
      |                            ~~~~~~ ^
test/boost/query_processor_test.cc:188:39: error: no member named 'id' in 'seastar::metrics::impl::metric_series_metadata'
  188 |         BOOST_REQUIRE(found != metric.id.labels().cend());
      |                                ~~~~~~ ^
/usr/include/boost/test/tools/old/interface.hpp:85:6: note: expanded from macro 'BOOST_REQUIRE'
   85 |     (P), BOOST_TEST_STRINGIZE( P ), REQUIRE, CHECK_PRED, _ )
      |      ^
/usr/include/boost/test/tools/old/interface.hpp:68:61: note: expanded from macro 'BOOST_TEST_TOOL_IMPL'
   68 |         BOOST_JOIN( BOOST_TEST_TOOL_PASS_PRED, frwd_type )( P, ARGS ),          \
      |                                                             ^
/usr/include/boost/test/tools/old/interface.hpp:51:47: note: expanded from macro 'BOOST_TEST_TOOL_PASS_PRED2'
   51 | #define BOOST_TEST_TOOL_PASS_PRED2( P, ARGS ) P
      |                                               ^
2 errors generated.
[4/5] CXX build/dev/test/boost/group0_test.o
FAILED: build/dev/test/boost/group0_test.o 
clang++ -MD -MT build/dev/test/boost/group0_test.o -MF build/dev/test/boost/group0_test.o.d -std=gnu++23 -I/home/bhalevy/dev/scylla/seastar/include -I/home/bhalevy/dev/scylla/build/dev/seastar/gen/include -Werror=unused-result -fstack-clash-protection -DSEASTAR_API_LEVEL=7 -DSEASTAR_BUILD_SHARED_LIBS -DSEASTAR_SSTRING -DSEASTAR_ENABLE_ALLOC_FAILURE_INJECTION -DSEASTAR_LOGGER_COMPILE_TIME_FMT -DSEASTAR_SCHEDULING_GROUPS_COUNT=16 -DSEASTAR_LOGGER_TYPE_STDOUT -DSEASTAR_TYPE_ERASE_MORE -DFMT_SHARED -I/usr/include/p11-kit-1 -DWITH_GZFILEOP -g -ffile-prefix-map=/home/bhalevy/dev/scylla=. -DSEASTAR_NO_EXCEPTION_HACK -march=westmere -DDEVEL -DSEASTAR_ENABLE_ALLOC_FAILURE_INJECTION -DSCYLLA_ENABLE_ERROR_INJECTION -DSCYLLA_ENABLE_PREEMPTION_SOURCE -O2 -DSCYLLA_BUILD_MODE=dev -Xclang -fexperimental-assignment-tracking=disabled -iquote. -iquote build/dev/gen -std=gnu++23 -g -ffile-prefix-map=/home/bhalevy/dev/scylla=. -DSEASTAR_NO_EXCEPTION_HACK -march=westmere -DBOOST_ALL_DYN_LINK    -fvisibility=hidden -isystem abseil -Wall -Werror -Wextra -Wimplicit-fallthrough -Wno-mismatched-tags -Wno-c++11-narrowing -Wno-overloaded-virtual -Wno-unused-parameter -Wno-unsupported-friend -Wno-missing-field-initializers -Wno-deprecated-copy -Wno-enum-constexpr-conversion -Wno-error=deprecated-declarations -DXXH_PRIVATE_API -DSEASTAR_TESTING_MAIN  -c -o build/dev/test/boost/group0_test.o test/boost/group0_test.cc
test/boost/group0_test.cc:49:62: error: no matching constructor for initialization of 'const internalized_labels_ref &' (aka 'const lw_shared_ptr<const map<basic_sstring<char, unsigned int, 15>, basic_sstring<char, unsigned int, 15>>> &')
   49 |             const auto& registered_metric = metric_family.at({{"shard", "0"}});
      |                                                              ^~~~~~~~~~~~~~~~
/home/bhalevy/dev/scylla/seastar/include/seastar/core/shared_ptr.hh:274:5: note: candidate constructor not viable: cannot convert initializer list argument to 'lw_shared_ptr_counter_base *'
  274 |     lw_shared_ptr(lw_shared_ptr_counter_base* p) noexcept : _p(p) {
      |     ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/bhalevy/dev/scylla/seastar/include/seastar/core/shared_ptr.hh:303:5: note: candidate constructor not viable: cannot convert initializer list argument to 'std::nullptr_t'
  303 |     lw_shared_ptr(std::nullptr_t) noexcept : lw_shared_ptr() {}
      |     ^             ~~~~~~~~~~~~~~
/home/bhalevy/dev/scylla/seastar/include/seastar/core/shared_ptr.hh:304:5: note: candidate constructor not viable: cannot convert initializer list argument to 'const lw_shared_ptr<const map<basic_sstring<char, unsigned int, 15>, basic_sstring<char, unsigned int, 15>>>'
  304 |     lw_shared_ptr(const lw_shared_ptr& x) noexcept : _p(x._p) {
      |     ^             ~~~~~~~~~~~~~~~~~~~~~~
/home/bhalevy/dev/scylla/seastar/include/seastar/core/shared_ptr.hh:314:5: note: candidate constructor not viable: cannot convert initializer list argument to 'lw_shared_ptr<const map<basic_sstring<char, unsigned int, 15>, basic_sstring<char, unsigned int, 15>>>'
  314 |     lw_shared_ptr(lw_shared_ptr&& x) noexcept  : _p(x._p) {
      |     ^             ~~~~~~~~~~~~~~~~~
/home/bhalevy/dev/scylla/seastar/include/seastar/core/shared_ptr.hh:302:5: note: candidate constructor not viable: requires 0 arguments, but 1 was provided
  302 |     lw_shared_ptr() noexcept = default;
      |     ^
/home/bhalevy/dev/scylla/seastar/include/seastar/core/metrics_api.hh:314:59: note: passing argument to parameter 'l' here
  314 |     const register_ref& at(const internalized_labels_ref& l) const {
      |                                                           ^
1 error generated.
```

bhalevy:
```
-            const auto& registered_metric = metric_family.at({{"shard", "0"}});
+            const auto& registered_metric = metric_family.at(make_lw_shared<seastar::metrics::impl::labels_type>({{"shard", "0"}}));
```

This may compile but does it do the right thing?
Wouldn't that match the lw_shared_ptr **pointer** rather than the labels value?

bhalevy:
Whatever we do, we cannot update the seastar submodule in scylla at the moment.
So I still vote for reverting ASAP and looking for an agreeable solution out of tree, and then resubmit the code while checking in advance that it doesn't break the user code.

nyh:
Just a small and **too late** bike-shedding comment:

The appropriate verb is **intern**, not "internalize".
Check out https://en.wikipedia.org/wiki/Interning_(computer_science). This term was first (as far as I know) introduced in Lisp, it's called intern in Java (https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#intern%28%29), in the X11 protocol (https://tronche.com/gui/x/xlib/window-information/XInternAtom.html), and probably many other systems I forgot.

StephanDollberg:
> Wouldn't that match the lw_shared_ptr pointer rather than the labels value?

No because metrics_family::at internally does the right thing (uses the holder type which has overloaded operator<).


> it doesn't break the user code

I don't think that's possible without changing the API level but that seems overkill for an internal change.

Whenever we update our seastar pointer in redpanda something breaks usually and we have to make some small changes. I don't think you want to hinder seastar changes just because it would require some changes in user code. That's why you pin the submodule pointer after all. And for such small changes the API level way is also lots of effort.

> The appropriate verb is intern, not "internalize".

Sure, don't think it's too late. Can just rename.

bhalevy:
> > The appropriate verb is intern, not "internalize".
> 
> Sure, don't think it's too late. Can just rename.

Well, this PR is already merged.

travisdowns:
The status of the `impl` namespaces with respect to API compatibility is unclear to me. It is clear that changes in the `internal` namespaces [are allowed without incrementing API version](https://github.com/scylladb/seastar/blob/master/doc/compatibility.md#internal-namespace) but the status of `impl` is unclear. My guess is that it would fall under the same rule as internal and it is simply an oversight that these few methods are in `seastar::impl` rather than `seastar::internal` or `seastar::internal::impl`. 

The `impl` namespace is only used in the metrics code and and one use is `when_all.h`.

> and then resubmit the code while checking in advance that it doesn't break the user code.

Assuming this ends up being judged to be in the internal interface, would this still be your suggestion?

bhalevy:
> The status of the `impl` namespaces with respect to API compatibility is unclear to me. It is clear that changes in the `internal` namespaces [are allowed without incrementing API version](https://github.com/scylladb/seastar/blob/master/doc/compatibility.md#internal-namespace) but the status of `impl` is unclear. My guess is that it would fall under the same rule as internal and it is simply an oversight that these few methods are in `seastar::impl` rather than `seastar::internal` or `seastar::internal::impl`.
> 
> The `impl` namespace is only used in the metrics code and and one use is `when_all.h`.
> 
> > and then resubmit the code while checking in advance that it doesn't break the user code.
> 
> Assuming this ends up being judged to be in the internal interface, would this still be your suggestion?

No, it wouldn't.

PR #2563, opened 2024-12-02
scheduler: Report long queue stall

xemul:
There can be at least two kinds of CPU stalls in seastar: task stall and queue stalls. The former is when a single task non-preemptively occupies a CPU for a too long period of time (configured by the option named blocked-reactor-notify-ms). The latter happens when the list of tasks to run happens to go above a certain threshold (configured by the option named max-task-backlog which is 1000 by default). In that case scheduler bites the buller and doesn't stop running tasks until the queue shrinks ignoring the need_preempt() value.

The stall of the second kind can be detected by checking the "stalls" historgram in the "reactor" group and is also implicitly reported via time-spent-in-task-quota-voilation counter. However, it's good to have more information about this stall other than knowing that "bad things" had just happenned.

The suggestion is to print which tasks are currently there in the queues by iterating the queue, collecting the typeif(*task).name()-s into a histogram and dumping it into logs.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -620,6 +620,7 @@ private:
     friend future<> seastar::destroy_scheduling_group(scheduling_group) noexcept;
     friend future<> seastar::rename_scheduling_group(scheduling_group sg, sstring new_name, sstring new_shortname) noexcept;
     friend future<scheduling_group_key> scheduling_group_key_create(scheduling_group_key_config cfg) noexcept;
+    friend seastar::internal::log_buf::inserter_iterator do_dump_task_queue(seastar::internal::log_buf::inserter_iterator it, const task_queue& tq);
 
     future<struct statfs> fstatfs(int fd) noexcept;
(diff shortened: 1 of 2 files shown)


xemul:
9d1db3f2b89da59aff6492192510bc4cc8016098
```
Author: Avi Kivity <avi@scylladb.com>
Date:   Wed Jul 20 19:15:55 2016 +0300

    reactor: limit task backlog
    
    I've seen a crash where there were 700,000 tasks in reactor::_pending_tasks,
    and all of memory was tied up in those tasks.  Presumably the during the
    2ms quota fewer tasks were processed than were added up by the pollers
    afterwards; the smp poller is most suspect, because in an unbalanced system,
    other shards can easily load us up.
    
    Fix by limiting the backlog; above (by default) 1000 tasks, we will ignore
    I/O and process tasks until we're below the limit.  This will relieve
    the memory load on this shard.
 ```

xemul:
@avikivity , please review

PR #2567, opened 2024-12-05
Fix ipv6 socket_address comparision

### LINKED ISSUE #2566: socket_address compares ipv6 addresses incorrectly
Comparision of addresses like
socket_address addr1(ipv6_addr("::1"));
socket_address addr2(ipv6_addr("::2"));
is false positive.

Problem is in last line of operator== of socket_address.
IN6_ARE_ADDR_EQUAL macro should compare in6_addr structures, while passed types are of sockaddr_in6.

SlowTurtlesEscape:
Fix #2566 

Pass correct types to IN6_ARE_ADDR_EQUAL macro
Add unit test

### DIFF
--- a/tests/unit/ipv6_test.cc
+++ b/tests/unit/ipv6_test.cc
@@ -103,3 +103,26 @@ SEASTAR_TEST_CASE(tcp_packet_test) {
     });
 }
 
+SEASTAR_TEST_CASE(ipv6_equal_test) {
+    const uint16_t port{8080};
+    const uint16_t port2{8088};
+
+    const std::string str_addr1{"abcd:fedc:ba98:7654:3210:0123:4567:89ab"};
+    const std::string str_addr2{"0123:4567:89ab:cdef:3210:0123:4567:89ab"};
+    const std::string str_addr3{"abcd:fedc:ba98:7654:3210:0123:4567:8900"};
+
+    socket_address sock_addr1(ipv6_addr(str_addr1, port));
+    socket_address sock_addr2(ipv6_addr(str_addr2, port));
+    socket_address sock_addr3(ipv6_addr(str_addr1, port));
+
+    socket_address sock_addr4(ipv6_addr(str_addr3, port));
+    socket_address sock_addr5(ipv6_addr(str_addr1, port2));
+
+    BOOST_CHECK_NE(sock_addr1, sock_addr2);
+    BOOST_CHECK_EQUAL(sock_addr1, sock_addr3);
+    BOOST_CHECK_NE(sock_addr1, sock_addr4);
+    BOOST_CHECK_NE(sock_addr1, sock_addr5);
+
+    return make_ready_future();
+}
+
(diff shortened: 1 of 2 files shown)


tests/unit/ipv6_test.cc
```diff
@@ -103,3 +103,18 @@ SEASTAR_TEST_CASE(tcp_packet_test) {
...
+	ipv6_addr addr1{"abcd:fedc:ba98:7654:3210:0123:4567:89ab"};
+	ipv6_addr addr2{"0123:4567:89ab:cdef:3210:0123:4567:89ab"};
+	ipv6_addr addr3{"abcd:fedc:ba98:7654:3210:0123:4567:89ab"};
+
+	socket_address sock_addr1(addr1);
+	socket_address sock_addr2(addr2);
+	socket_address sock_addr3(addr3);
+
+	BOOST_REQUIRE(sock_addr1 != sock_addr2);
+	BOOST_REQUIRE(sock_addr1 == sock_addr3);
```
tchaikov, line 116: the fix in `socket_address::operator==()` is correct. but this test is not able to reproduce the issue you are fixing in this change. if we take a look at the layout of `sockaddr_in6`, see https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/inet/netinet/in.h#L252-L259 . the `sin6_addr` sits in the middle of the struct. 

without your fix, both of the comparison statements return the expected result. because instead comparing the IPv6 address (16 bytes, 4 uint32_t numbers), they are comparing the following fields

- `sa_family_t` : unsigned short (uint16_t)
- `sin6_port`: uint16_t 
- `sin6_flowinfo`: uint32_t
- the lower 64 bits in `sin6_addr`

so, in order to verify your change, we have to create a test case where the lower bits of the compared `sockaddr_6` are identical, while its `sin6_addr`'s higher 32 bits are different.

so i'd suggest

1. specify the port number explicitly. could define a constant instead of a magic number for better maintainability.
2. define yet another `ipv6_addr`, like `ipv6_addr addr5{"abcd:fedc:ba98:7654:3210:0123:4567:89ae", port)`, so that its higher bits in address field is different from `addr1`.
3. use `BOOST_CHECK_NE` and `BOOST_CHECK_EQUAL` instead of `BOOST_REQUIRE`. because:

   - if the test fails, Boot.Test would be able to print out the string representation of the lhs and rhs of the compare operation
   - we don't need to fail the whole test case if a single compare fails, if the succeeding tests do not depend on the completion of previous test
4. use spaces instead tabs for indent. see https://github.com/scylladb/seastar/blob/master/coding-style.md#whitespace

tchaikov: also, please add a "net:" prefix in the title of the commit message. the the output of `git log src/net/socket_address.cc` for more examples.

tests/unit/ipv6_test.cc
```diff
@@ -103,3 +103,22 @@ SEASTAR_TEST_CASE(tcp_packet_test) {
     });
 }
 
+SEASTAR_TEST_CASE(ipv6_equal_test) {
+    uint16_t port{8080};
+    std::string const str_addr1{"abcd:fedc:ba98:7654:3210:0123:4567:89ab"};
```
tchaikov, line 108: could you please put `const` specifier before `std::string`. or just drop it. just to be more consistent with the seastar code base.

and could you take this opportunity to add a test for comparing two addresses with the same IPv6 address but with different ports?

tchaikov (approved):
lgtm

tchaikov:

PR #2569, opened 2024-12-09
tls: Introduce OpenSSL

michael-redpanda:
Introduces OpenSSL as an alternative TLS implementation to GnuTLS.  This is a build-time configuration controlled by the CMake variable `Seastar_USE_OPENSSL`.  The `configure.py` script has been updated to now have a `--crypto-provider` option.  Valid arguments to that are `OpenSSL` and `GnuTLS`.

This implementation was released in [Redpanda v24.2](https://github.com/redpanda-data/redpanda/releases/tag/v24.2.1) on July 31st, and has been running on production clusters since.

Redpanda implemented these changes in order to provide a FIPS-compliant build to customers that require it (such as those wishing to undergo FedRAMP evaluation).  OpenSSL was selected as it allows implementors to maintain the validation of the cryptographic module even when it's built from source.

* [OpenSSL's security policy](https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp4282.pdf)
  * See appendix A for instructions on how to build OpenSSL for FIPS
* [FIPS 140-2 Implementation Guidance](https://csrc.nist.rip/csrc/media/projects/cryptographic-module-validation-program/documents/fips140-2/fips1402ig.pdf)
  * See section G.5 Maintaining validation compliance of software or firmware cryptographic
modules

No changes have been introduced to enable the FIPS provider for Seastar.  It is up to the implementor to enable and use the FIPS cryptographic module if desired.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -11,7 +11,7 @@ concurrency:
 
 jobs:
   regular_test:
-    name: "Test (${{ matrix.compiler }}, C++${{ matrix.standard}}, ${{ matrix.mode }})"
+    name: "Test (${{ matrix.compiler }}, C++${{ matrix.standard}}, ${{ matrix.mode }}, ${{ matrix.crypto_provider }})"
     uses: ./.github/workflows/test.yaml
     strategy:
       fail-fast: false
@@ -19,12 +19,14 @@ jobs:
         compiler: [clang++, g++]
         standard: [20, 23]
         mode: [dev, debug, release]
+        crypto_provider: [GnuTLS, OpenSSL]
     with:
       compiler: ${{ matrix.compiler }}
       standard: ${{ matrix.standard }}
       mode: ${{ matrix.mode }}
       enables: ${{ matrix.enables }}
       options: ${{ matrix.options }}
+      crypto_provider: ${{ matrix.crypto_provider }}
   build_with_dpdk:
     name: "Test with DPDK enabled"
     uses: ./.github/workflows/test.yaml
@@ -36,8 +38,8 @@ jobs:
       mode: release
       enables: --enable-dpdk
       options: --cook dpdk
-  build_with_cxx_modules:
-    name: "Test with C++20 modules enabled"
+  build_with_cxx_modules_gnutls:
+    name: "Test with C++20 modules enabled (GnuTLS)"
     uses: ./.github/workflows/test.yaml
     strategy:
       fail-fast: false
@@ -47,3 +49,16 @@ jobs:
       mode: debug
       enables: --enable-cxx-modules
       enable-ccache: false
+      crypto_provider: GnuTLS
+  build_with_cxx_modules_openssl:
+    name: "Test with C++20 modules enabled (OpenSSL)"
+    uses: ./.github/workflows/test.yaml
+    strategy:
+      fail-fast: false
+    with:
+      compiler: clang++
+      standard: 23
+      mode: debug
+      enables: --enable-cxx-modules
+      enable-ccache: false
+      crypto_provider: OpenSSL
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -760,6 +760,8 @@ add_library (seastar
   src/net/stack.cc
   src/net/tcp.cc
   src/net/tls.cc
+  src/net/tls-impl.cc
+  src/net/tls-impl.hh
   src/net/udp.cc
   src/net/unix_address.cc
   src/net/virtio.cc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,16 @@ if (NOT Seastar_SCHEDULING_GROUPS_COUNT MATCHES "^[1-9][0-9]*")
   message(FATAL_ERROR "Seastar_SCHEDULING_GROUPS_COUNT must be a positive number (${Seastar_SCHEDULING_GROUPS_COUNT})")
 endif ()
 
+option (Seastar_USE_OPENSSL
+  "Use OpenSSL rather than GnuTLS for cryptographic operations, including TLS"
+  OFF)
+
+if (Seastar_USE_OPENSSL)
+  set(Seastar_USE_GNUTLS OFF)
+else()
+  set(Seastar_USE_GNUTLS ON)
+endif ()
+
 #
 # Add a dev build type.
 #
@@ -753,13 +763,14 @@ add_library (seastar
   src/net/native-stack-impl.hh
   src/net/native-stack.cc
   src/net/net.cc
+  $<$<BOOL:${Seastar_USE_OPENSSL}>:src/net/ossl.cc>
   src/net/packet.cc
   src/net/posix-stack.cc
   src/net/proxy.cc
   src/net/socket_address.cc
   src/net/stack.cc
   src/net/tcp.cc
-  src/net/tls.cc
+  $<$<BOOL:${Seastar_USE_GNUTLS}>:src/net/tls.cc>
   src/net/tls-impl.cc
   src/net/tls-impl.hh
   src/net/udp.cc
@@ -858,7 +869,9 @@ target_link_libraries (seastar
     SourceLocation::source_location
   PRIVATE
     ${CMAKE_DL_LIBS}
-    GnuTLS::gnutls
+    $<$<BOOL:${Seastar_USE_GNUTLS}>:GnuTLS::gnutls>
+    $<$<BOOL:${Seastar_USE_OPENSSL}>:OpenSSL::SSL>
+    $<$<BOOL:${Seastar_USE_OPENSSL}>:OpenSSL::Crypto>
     StdAtomic::atomic
     lksctp-tools::lksctp-tools
     protobuf::libprotobuf
@@ -915,6 +928,7 @@ include (CTest)
 target_compile_definitions(seastar
   PUBLIC
   SEASTAR_API_LEVEL=${Seastar_API_LEVEL}
+  $<$<BOOL:${Seastar_USE_OPENSSL}>:SEASTAR_USE_OPENSSL>
   $<$<BOOL:${BUILD_SHARED_LIBS}>:SEASTAR_BUILD_SHARED_LIBS>)
 
 target_compile_features(seastar
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -157,6 +157,18 @@ namespace tls {
      */
     using dn_callback = noncopyable_function<void(session_type type, sstring subject, sstring issuer)>;
 
+    enum class client_auth {
+        NONE, REQUEST, REQUIRE
+    };
+
+    /**
+     * Session resumption support.
+     * We only support TLS1.3 session tickets.
+     */
+    enum class session_resume_mode {
+        NONE, TLS13_SESSION_TICKET
+    };
+
     /**
      * Holds certificates and keys.
      *
@@ -235,24 +247,18 @@ namespace tls {
         template<typename Base>
         friend class reloadable_credentials;
         shared_ptr<impl> _impl;
+
+        // The following methods are provided so classes that inherity from
+        // certificate_credentials can access the underly implementation
+        void enable_load_system_trust();
+        void set_client_auth(client_auth);
+        void set_session_resume_mode(session_resume_mode);
     };
 
     /** Exception thrown on certificate validation error */
     class verification_error : public std::runtime_error {
     public:
         using runtime_error::runtime_error;
-    };
-
-    enum class client_auth {
-        NONE, REQUEST, REQUIRE
-    };
-
-    /**
-     * Session resumption support.
-     * We only support TLS1.3 session tickets.
-    */
-    enum class session_resume_mode {
-        NONE, TLS13_SESSION_TICKET
     };
 
     /**
@@ -312,8 +318,8 @@ namespace tls {
 
         future<> set_system_trust();
         void set_client_auth(client_auth);
-        void set_priority_string(const sstring&);
         void set_session_resume_mode(session_resume_mode);
+        void set_priority_string(const sstring&);
 
         void apply_to(certificate_credentials&) const;
 
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -115,6 +115,16 @@ namespace tls {
         shared_ptr<impl> _impl;
     };
 
+    enum class tls_version {
+        tlsv1_0,
+        tlsv1_1,
+        tlsv1_2,
+        tlsv1_3
+    };
+
+    std::string_view format_as(tls_version);
+    std::ostream& operator<<(std::ostream&, const tls_version&);
+
     class abstract_credentials {
     protected:
         abstract_credentials() = default;
@@ -202,6 +212,7 @@ namespace tls {
 
         // TODO add methods for certificate verification
 
+#ifndef SEASTAR_USE_OPENSSL
         /**
          * TLS handshake priority string. See gnutls docs and syntax at
          * https://gnutls.org/manual/html_node/Priority-Strings.html
@@ -209,6 +220,45 @@ namespace tls {
          * Allows specifying order and allowance for handshake alg.
          */
         void set_priority_string(const sstring&);
+#endif
+
+#ifdef SEASTAR_USE_OPENSSL
+        /**
+         * Used to set the cipher string for TLS versions 1.2 and below
+         *
+         * See https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_cipher_list.html
+         * for documentation on the format of the cipher list string
+         */
+        void set_cipher_string(const sstring&);
+
+        /**
+         * Used to set the cipher suites to use for TLSv1.3
+         *
+         * See https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_ciphersuites.html
+         * for documentation on the format of the ciphersuites string.
+         */
+        void set_ciphersuites(const sstring&);
+
+        /**
+         * Call this when you want to enable server precedence when
+         * negotitating the TLS handshake.  Client precedence is on
+         * by default.
+         */
+        void enable_server_precedence();
+        /**
+         * @brief Set the minimum tls version for this connection
+         *
+         * If unset, will default to the minimum of the underlying
+         * implementation
+         */
+        void set_minimum_tls_version(tls_version);
+        /**
+         * @brief Set the maximum tls version for this connection
+         *
+         * If unset, will default to the maximum of the underly implementation
+         */
+        void set_maximum_tls_version(tls_version);
+#endif
 
         /**
          * Register a callback for receiving Distinguished Name (DN) information
@@ -319,7 +369,17 @@ namespace tls {
         future<> set_system_trust();
         void set_client_auth(client_auth);
         void set_session_resume_mode(session_resume_mode);
+#ifndef SEASTAR_USE_OPENSSL
         void set_priority_string(const sstring&);
+#endif
+
+#ifdef SEASTAR_USE_OPENSSL
+        void set_cipher_string(const sstring&);
+        void set_ciphersuites(const sstring&);
+        void enable_server_precedence();
+        void set_minimum_tls_version(tls_version);
+        void set_maximum_tls_version(tls_version);
+#endif
 
         void apply_to(certificate_credentials&) const;
 
@@ -337,6 +397,11 @@ namespace tls {
         client_auth _client_auth = client_auth::NONE;
         session_resume_mode _session_resume_mode = session_resume_mode::NONE;
         sstring _priority;
+        sstring _cipher_string;
+        sstring _ciphersuites;
+        bool _enable_server_precedence = false;
+        std::optional<tls_version> _min_tls_version;
+        std::optional<tls_version> _max_tls_version;
     };
 
     using session_data = std::vector<uint8_t>;
@@ -586,5 +651,12 @@ template <> struct fmt::formatter<seastar::tls::subject_alt_name> : fmt::formatt
     template <typename FormatContext>
     auto format(const seastar::tls::subject_alt_name& name, FormatContext& ctx) const {
         return fmt::format_to(ctx.out(), "{}={}", name.type, name.value);
+    }
+};
+
+template <> struct fmt::formatter<seastar::tls::tls_version> : fmt::formatter<string_view> {
+    template<typename FormatContext>
+    auto format(seastar::tls::tls_version version, FormatContext& ctx) const {
+        return fmt::format_to(ctx.out(), "{}", format_as(version));
     }
 };
--- a/src/net/ossl.cc
+++ b/src/net/ossl.cc
@@ -1,0 +1,2023 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright 2015 Cloudius Systems
+ * Copyright 2024 Redpanda Data
+ */
+
+// The data flow in the openssl-based seastar tls implementation looks like the following. The session's put/get()
+// methods delegate to openssl functions, which perform the openssl handshake, encryption and decryption, and then they
+// write the raw data into a custom BIO object. This custom BIO object is implemented in this file, it is specific to
+// seastar and it delegates the write/read methods to the seastar::tls::session's data_sink/data_source (which are the
+// interfaces around the underlying seastar connected socket).
+//
+//        +----------------------------+      +-------------------------------------+     +-------------+    +----------------------------+   +------+
+//        |                            |      |                                     |     |             |    |                            |   |      |
+// ------->    seastar::tls::session   +------> SSL_{do_handshake|write_ex|read_ex} +-----> custom BIO  +---->    data_sink/data_source   +--->  OS  |
+//        |                            |      |                                     |     |             |    |      (seastar socket)      |   |      |
+//        +----------------------------+      +-------------------------------------+     +-------------+    +----------------------------+   +------+
+
+#ifdef SEASTAR_MODULE
+module;
+#endif
+
+#include <fmt/ranges.h>
+#include <openssl/bio.h>
+#include <openssl/core_names.h>
+#include <openssl/err.h>
+#include <openssl/evp.h>
+#include <openssl/pem.h>
+#include <openssl/pkcs12.h>
+#include <openssl/provider.h>
+#include <openssl/rand.h>
+#include <openssl/safestack.h>
+#include <openssl/ssl.h>
+#include <openssl/sslerr.h>
+#include <openssl/x509.h>
+#include <openssl/x509_vfy.h>
+#include <openssl/x509v3.h>
+
+#include <cassert>
+#include <chrono>
+#include <functional>
+#include <span>
+#include <system_error>
+#include <unordered_set>
+#include <utility>
+
+#include <netinet/in.h>
+
+#ifdef SEASTAR_MODULE
+module seastar;
+#else
+#include <seastar/core/gate.hh>
+#include <seastar/core/sstring.hh>
+#include <seastar/core/with_timeout.hh>
+#include <seastar/net/stack.hh>
+#include <seastar/net/tls.hh>
+#include <seastar/util/defer.hh>
+#include <seastar/util/later.hh>
+#include <seastar/util/log.hh>
+
+#include "net/tls-impl.hh"
+#endif
+
+namespace seastar {
+
+enum class ossl_errc : int{};
+
+}
+
+namespace std {
+
+template<>
+struct is_error_code_enum<seastar::ossl_errc> : true_type {};
+
+}
+
+template<>
+struct fmt::formatter<seastar::ossl_errc> : public fmt::formatter<std::string_view> {
+    auto format(seastar::ossl_errc error, fmt::format_context& ctx) const -> decltype(ctx.out()) {
+        constexpr size_t error_buf_size = 256;
+        // Buffer passed to ERR_error_string must be at least 256 bytes large
+        // https://www.openssl.org/docs/man3.0/man3/ERR_error_string_n.html
+        std::array<char, error_buf_size> buf{};
+        ERR_error_string_n(
+          static_cast<unsigned long>(error), buf.data(), buf.size());
+        // ERR_error_string_n does include the terminating null character
+        return fmt::format_to(ctx.out(), "{}", buf.data());
+    }
+};
+
+namespace seastar {
+
+int tls_version_to_ossl(tls::tls_version version) {
+    switch(version) {
+    case tls::tls_version::tlsv1_0:
+        return TLS1_VERSION;
+    case tls::tls_version::tlsv1_1:
+        return TLS1_1_VERSION;
+    case tls::tls_version::tlsv1_2:
+        return TLS1_2_VERSION;
+    case tls::tls_version::tlsv1_3:
+        return TLS1_3_VERSION;
+    }
+
+    __builtin_unreachable();
+}
+
+class ossl_error_category : public std::error_category {
+public:
+    constexpr ossl_error_category() noexcept : std::error_category{} {}
+    const char* name() const noexcept override {
+        return "OpenSSL";
+    }
+    std::string message(int error) const override {
+        return fmt::format("{}", static_cast<ossl_errc>(error));
+    }
+};
+
+const std::error_category& tls::error_category() {
+    static const ossl_error_category ec;
+    return ec;
+}
+
+std::error_code make_error_code(ossl_errc e) {
+    return std::error_code(static_cast<int>(e), tls::error_category());
+}
+
+std::system_error make_ossl_error(const std::string & msg) {
+    std::vector<ossl_errc> error_codes;
+    for (auto code = ERR_get_error(); code != 0; code = ERR_get_error()) {
+        error_codes.push_back(static_cast<ossl_errc>(code));
+    }
+    if (error_codes.empty()) {
+        return std::system_error{
+            static_cast<int>(ERR_PACK(ERR_LIB_USER, 0, ERR_R_OPERATION_FAIL)),
+            tls::error_category(),
+            msg};
+    } else {
+        auto err_code = static_cast<unsigned long>(error_codes.front());
+        if (ERR_LIB_SYS == ERR_GET_LIB(err_code)) {
+            // If the error code belongs to ERR_LIB_SYS, then the error is a system error
+            // Extract the errno using ERR_GET_REASON and throw a std::generic_category
+            return std::system_error(
+                ERR_GET_REASON(err_code),
+                std::generic_category(),
+                fmt::format("{}: {}", msg, error_codes));
+        }
+        return std::system_error(
+            static_cast<int>(err_code),
+            tls::error_category(),
+            fmt::format("{}: {}", msg, error_codes));
+    }
+}
+
+template<typename T>
+sstring asn1_str_to_str(T* asn1) {
+    const auto len = ASN1_STRING_length(asn1);
+    return sstring(reinterpret_cast<const char*>(ASN1_STRING_get0_data(asn1)), len);
+};
+
+template<typename T, auto fn>
+struct ssl_deleter {
+    void operator()(T* ptr) { fn(ptr); }
+};
+
+// Must define this method as sk_X509_pop_free is a macro
+void X509_pop_free(STACK_OF(X509)* ca) {
+    sk_X509_pop_free(ca, X509_free);
+}
+
+void X509_INFO_pop_free(STACK_OF(X509_INFO)* infos) {
+    sk_X509_INFO_pop_free(infos, X509_INFO_free);
+}
+
+void GENERAL_NAME_pop_free(GENERAL_NAMES* gns) {
+    sk_GENERAL_NAME_pop_free(gns, GENERAL_NAME_free);
+}
+
+template<typename T, auto fn>
+using ssl_handle = std::unique_ptr<T, ssl_deleter<T, fn>>;
+
+using bio_method_ptr = ssl_handle<BIO_METHOD, BIO_meth_free>;
+using bio_ptr = ssl_handle<BIO, BIO_free>;
+using evp_pkey_ptr = ssl_handle<EVP_PKEY, EVP_PKEY_free>;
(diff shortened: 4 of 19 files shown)


avikivity:
@elcallio please review

Are there any functional differences?

Is hot reload of certificates supported?

Should we support gnutls and openssl in parallel?

michael-redpanda:
> Are there any functional differences?

The only major difference between the OpenSSL vs GnuTLS implementation is how the implementation is configured.  With OpenSSL, there are 5 new methods that can be used to control it (compared to the single `set_priority_string` with GnuTLS):

* `set_cipher_string` - used for controlling which ciphers are available in TLS1.2 and below ([OpenSSL SSL_CTX_set_cihper_list](https://docs.openssl.org/master/man3/SSL_CTX_set_cipher_list/))
* `set_ciphersuites` - used for controlling which cipher suites are available in TLS1.3 (same doc page as above)
* `enable_server_precedence`
* `set_minimum_tls_version` - defaults to 1.0
* `set_maximum_tls_version` - defaults to 1.3

> Is hot reload of certificates supported?

Yes

src/net/tls-impl.hh
```diff
@@ -0,0 +1,236 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright 2015 Cloudius Systems
+ */
+#pragma once
```
elcallio, line 21: Why is this file in public header dir? Seems to me it should only be required by internal compilation units, so should probably be under src.

src/net/ossl.cc
```diff
@@ -815,11 +818,17 @@ class session : public enable_shared_from_this<session>, public session_impl {
...
     }
 
+    const char * get_type_string() const {
+        return _type == session_type::CLIENT ? "Client": "Server";
+    }
+
     // This function waits for the _output_pending future to resolve
     // If an error occurs, it is saved off into _error and returned
     future<> wait_for_output() {
+        _logger.trace("wait_for_output");
```
elcallio, line 828: Why invent a new logger type when you could just add a formatter for the session object that pretty-prints the local/remote etc, and then just use `tls_log::trace("{} wait_for_output")`?

tests/unit/tls_test.cc
```diff
@@ -1039,8 +1039,12 @@ SEASTAR_THREAD_TEST_CASE(test_reload_certificates) {
...
 
             try {
-                f2.get();
-                BOOST_FAIL("should not reach");
+                auto res = f2.get();
+                // If the server completes sending data to the client
+                // during the handshake before the client has fully
+                // closed its connection, then the get() call will
+                // succeed by return an empty buffer indicating EOF
+                BOOST_REQUIRE(res.size() == 0);
```
elcallio, line 1047: Seems like a good change, but not related to the rest of the PR

avikivity:
The different configuration options mean one of two:
 - side-by-side support, at least for a transitional period
 - emulate priority strings for openssl (far from the focus of Seastar)

So I think we should choose one implementation, and deprecate the other after a transition period.

gnutls was chosen due to funky licensing and an unstable API from openssl, but I think that's behind us now. Given that, it's better to use the leader in the field rather than a follower.

tests/unit/tls_test.cc
michael-redpanda: This test was being flaky when using OpenSSL.  It would pass reliably in release but not in debug.

mykaul:
GnuTLS is not FIPS enabled, if compiled with it? (I see https://www.gnutls.org/manual/html_node/FIPS140_002d2-mode.html ) - what's missing?

src/net/tls-impl.hh
elcallio: Doh, sorry, github confused me.

michael-redpanda:
> GnuTLS is not FIPS enabled, if compiled with it? (I see https://www.gnutls.org/manual/html_node/FIPS140_002d2-mode.html ) - what's missing?

The FIPS flag for GnuTLS means that GnuTLS will work in a FIPS _compliant_ way (e.g. rejecting any non FIPS approved crypto like DES or GOST), however that doesn't mean that its implementation was _validated_.  GnuTLS doesn't provide a path to build it and maintain validation, OpenSSL does (see above links to OpenSSL's security policy).

src/net/ossl.cc
michael-redpanda: Thanks for the suggestion, great idea!

elcallio:
> gnutls was chosen due to funky licensing and an unstable API from openssl, but I think that's behind us now. Given that, it's better to use the leader in the field rather than a follower.

While I can somewhat sympathize with the sentiment, this would effectively mean dropping the way we handle TLS config for clients/applications (i.e. usage of the prio string). Since at least one database application I know of exposes this in its customer config, you'd effectively be asking to require changing all customer TLS configs where such a prio is applied. Not sure how many they are and how complicated the configs are. You said a transition period, but not sure how to handle this, nor enforce a config migration with clients?

avikivity:
> > gnutls was chosen due to funky licensing and an unstable API from openssl, but I think that's behind us now. Given that, it's better to use the leader in the field rather than a follower.
> 
> While I can somewhat sympathize with the sentiment, this would effectively mean dropping the way we handle TLS config for clients/applications (i.e. usage of the prio string). Since at least one database application I know of exposes this in its customer config, you'd effectively be asking to require changing all customer TLS configs where such a prio is applied. Not sure how many they are and how complicated the configs are. You said a transition period, but not sure how to handle this, nor enforce a config migration with clients?

I would not want to force a config migration.

Claude says this:

> The gnutls-cli tool includes a "--priority-debug" option that can show detailed information about a priority string's settings, but there isn't a direct tool to convert between GnuTLS priority strings and OpenSSL configuration formats.
> 
> You can analyze the GnuTLS priority string components with `gnutls-cli --priority-debug "PRIORITY STRING"` and then manually map those to equivalent OpenSSL configurations based on the cipher suites and protocols shown.
> 
> For more control, you could write a script that parses the gnutls-cli output and maps the components to OpenSSL's configuration syntax, though you'd need to account for the differences in how each library names and groups their ciphers and protocols.

elcallio:
Yes, but I am honestly very nervous about writing/maintaining a prio string translator. The mapping is not just cipher to cipher etc, it is a state machine in itself, disabling and adding ciphers, exchange modes etc. 
As for using cli tool and manually map - that is what a customer would have to do if we changed the config approach.

avikivity:
> Yes, but I am honestly very nervous about writing/maintaining a prio string translator. The mapping is not just cipher to cipher etc, it is a state machine in itself, disabling and adding ciphers, exchange modes etc. As for using cli tool and manually map - that is what a customer would have to do if we changed the config approach.

I agree with that.

@tzach do you have any insight about priority string configuration across our fleet? Do we ever diverge from the default?

mykaul:
> > Yes, but I am honestly very nervous about writing/maintaining a prio string translator. The mapping is not just cipher to cipher etc, it is a state machine in itself, disabling and adding ciphers, exchange modes etc. As for using cli tool and manually map - that is what a customer would have to do if we changed the config approach.
> 
> I agree with that.
> 
> @tzach do you have any insight about priority string configuration across our fleet? Do we ever diverge from the default?

Since we did not disable TLSv1.1 by default (not sure why), there's a good chance users do it - https://enterprise.docs.scylladb.com/stable/operating-scylla/security/client-node-encryption.html#priority-string-and-tlsv1-2-1-3-support

src/seastar.cc
```diff
@@ -335,6 +335,7 @@ module : private;
 #include <seastar/net/virtio.hh>
 
 #include "net/native-stack-impl.hh"
+#include "net/tls-impl.hh"
```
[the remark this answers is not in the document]
michael-redpanda: This is necessary for when compiling with modules enabled

src/net/tls-impl.hh
```diff
@@ -0,0 +1,236 @@
...
+#include <seastar/net/tls.hh>
+#include <seastar/net/stack.hh>
+#include <seastar/core/reactor.hh>
+namespace seastar {
+
+class net::get_impl {
+public:
+    static std::unique_ptr<connected_socket_impl> get(connected_socket s) {
+        return std::move(s._csi);
+    }
```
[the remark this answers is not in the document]
michael-redpanda: This was just moved from [`tls.cc`](https://github.com/scylladb/seastar/blob/2bba1204dbb018ff1551d08fc80a951af918a37d/src/net/tls.cc#L68-L80) to this file.

src/net/tls-impl.hh
```diff
@@ -0,0 +1,236 @@
...
+    session_ref(session_ref&&) = default;
+    session_ref(const session_ref&) = default;
+    ~session_ref() {
+        // This is not super pretty. But we take some care to only own sessions
+        // through session_ref, and we need to initiate shutdown on "last owner",
+        // since we cannot revive the session in destructor.
+        if (_session && _session.use_count() == 1) {
+            _session->close();
+        }
+    }
```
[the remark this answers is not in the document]
michael-redpanda: same [here](https://github.com/scylladb/seastar/blob/2bba1204dbb018ff1551d08fc80a951af918a37d/src/net/tls.cc#L1833-L1840)

src/net/tls-impl.cc
```diff
@@ -229,9 +231,33 @@ void tls::credentials_builder::set_session_resume_mode(session_resume_mode m) {
...
 }
 
+#ifndef SEASTAR_USE_OPENSSL
 void tls::credentials_builder::set_priority_string(const sstring& prio) {
     _priority = prio;
 }
+#endif
+
+#ifdef SEASTAR_USE_OPENSSL
+void tls::credentials_builder::set_cipher_string(const sstring& cipher_string) {
```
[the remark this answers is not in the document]
michael-redpanda: That's a good idea, I'll move the definitions to the correct source file

src/net/ossl.cc
```diff
@@ -0,0 +1,2023 @@
...
+                tls_version_to_ossl(*max_tls_version))) {
+                    throw make_ossl_error(
+                        fmt::format("Failed to set maximum TLS version to {}",
+                            *max_tls_version));
+            }
+        }
+
+        // This call is required to lower SSL's security level to permit TLSv1.0 and TLSv1.1
+        // See https://www.openssl.org/docs/man3.0/man3/SSL_CTX_set_security_level.html
+        SSL_CTX_set_security_level(ssl_ctx.get(), 0);
```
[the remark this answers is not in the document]
michael-redpanda: Did a little bit of research on this function and what the level controls changes depending on the version of OpenSSL.

For OpenSSL 3.0, setting security level to 3 disables TLS1.0 and below, and setting it to 4 disables TLS1.1 and below ([ref](https://docs.openssl.org/3.0/man3/SSL_CTX_set_security_level/#default-callback-behaviour)).

Starting with OpenSSL 3.1, level 1 disables TLS1.1 and below ([ref](https://docs.openssl.org/3.1/man3/SSL_CTX_set_security_level/#default-callback-behaviour)).

michael-redpanda:
Force push [`fc616ee`](https://github.com/scylladb/seastar/compare/5c709c28d63f9ab9dab5b253e45c352d9d6b0398..fc616ee38bc09af4c56fabd953e4fde29e879145):

* Added compile definition `SEASTAR_USE_GNUTLS` for when GnuTLS is the cryptographic provder
* Updated `#ifndef SEASTAR_USE_OPENSSL` to instead be `#ifdef SEASTAR_USE_GNUTLS`
* Move implementation specific credential builder methods to the implementation specific compilation units
* Removed `tls_session_logger` and added a pretty-print function for the OpenSSL implementation of `seastar::tls::session`
* Dynamically setting minimum security level passed to `SSL_CTX_set_security_level`

michael-redpanda:
Force push [`1a53361`](https://github.com/scylladb/seastar/compare/fc616ee38bc09af4c56fabd953e4fde29e879145..1a533615610fac2ae8fa02b2b5265f8926427188):

* Added missing compile definition for `SEASTAR_USE_GNUTLS` when compiled to use modules
* Added missing header include for `fmt/ostream.hh` when compiled to use modules

src/net/tls-impl.hh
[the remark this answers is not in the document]
elcallio: Yes. But that _was_ a cc. This is a header.

src/net/tls-impl.hh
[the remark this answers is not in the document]
elcallio: Same as above. Code should not be in headers if it can be avoided (even internal, not very included ones)

include/seastar/net/tls.hh
```diff
@@ -312,8 +318,8 @@ namespace tls {
 
         future<> set_system_trust();
         void set_client_auth(client_auth);
-        void set_priority_string(const sstring&);
         void set_session_resume_mode(session_resume_mode);
+        void set_priority_string(const sstring&);
```
elcallio, line 373: Seems like a gratuitous change?

include/seastar/net/tls.hh
```diff
@@ -235,6 +247,12 @@ namespace tls {
         template<typename Base>
         friend class reloadable_credentials;
         shared_ptr<impl> _impl;
+
+        // The following methods are provided so classes that inherity from
```
elcallio, line 301: Nit: inherit

src/net/ossl.cc
```diff
@@ -0,0 +1,2073 @@
...
+        }
+        // A success return code (0) does not signify if a cert was presented or not, that
+        // must be explicitly queried via SSL_get_peer_certificate
+        auto res = SSL_get_verify_result(_ssl.get());
+        if (res != X509_V_OK) {
+            auto stat_str(X509_verify_cert_error_string(res));
+            auto dn = extract_dn_information();
+            if (dn) {
+                std::string_view stat_str_view{stat_str};
+                if (stat_str_view.ends_with(" ")) {
```
elcallio, line 1332: Nit: I assume this is an openssl quirk of some sort? Maybe a while instead (instead of assuming just one space), or maybe even a proper back and front whitespace strip?

src/net/ossl.cc
```diff
@@ -0,0 +1,2073 @@
...
+                    break;
+            }
+        } else {
+            if (_creds->is_server_precedence_enabled()) {
+                SSL_CTX_set_options(ssl_ctx.get(), SSL_OP_CIPHER_SERVER_PREFERENCE);
+            }
+        }
+
+        auto & min_tls_version = _creds->minimum_tls_version();
+        auto & max_tls_version = _creds->maximum_tls_version();
```
elcallio, line 1724: Nit: space between type and ref.

src/net/ossl.cc
```diff
@@ -0,0 +1,2073 @@
...
+        if (max_tls_version.has_value()) {
+            if (!SSL_CTX_set_max_proto_version(ssl_ctx.get(),
+                tls_version_to_ossl(*max_tls_version))) {
+                    throw make_ossl_error(
+                        fmt::format("Failed to set maximum TLS version to {}",
+                            *max_tls_version));
+            }
+        }
+
+        auto get_min_level = [&ssl_ctx]() {
```
elcallio, line 1744: Nit: Should be "max level" really, as we should consider us setting the max level we can use given the tls version bounds.

src/net/ossl.cc
```diff
@@ -0,0 +1,2073 @@
...
+
+    return meth;
+}
+
+} // namespace tls
+
+BIO_METHOD* get_method() {
+    static thread_local bio_method_ptr method_ptr = [] {
+        auto ptr = tls::create_bio_method();
+        if (!ptr) {
```
elcallio, line 2092: Seems redundant - all paths in create throws if failed...

src/websocket/server.cc
```diff
@@ -126,6 +130,30 @@ future<> connection::process() {
 
 static std::string sha1_base64(std::string_view source) {
     unsigned char hash[20];
+
+#ifdef SEASTAR_USE_OPENSSL
```
elcallio, line 134: Could we instead make the function an external/internal-header-decl, and move the full impl into gnutls/openssl compilation units? This just adds to an already bad separation of concern/api exposure.

elcallio (commented):
Code looks ok, but again, I doubt I can really verify it just by looking. :-( 
One huge downside with the openssl code is that imho it seems to make the interface code even cludgier - perhaps partially an aspect of the code it needs to emulate is from a gnutls universe, but I would still argue that a lot of the gnutls interfaces are a bit nicer. 
Thus I would worry a little about maintenance here.

michael-redpanda:
Force push [`2f935ca`](https://github.com/scylladb/seastar/compare/1a533615610fac2ae8fa02b2b5265f8926427188..2f935cac22094152c1ed51daf6d3edcbea087fad):

* Addressed issues from comments
* Created GnuTLS & OpenSSL implementations for SHA1-Base64 used by websocket
* Fixed nits

CMakeLists.txt
```diff
@@ -89,6 +89,16 @@ if (NOT Seastar_SCHEDULING_GROUPS_COUNT MATCHES "^[1-9][0-9]*")
...
 endif ()
 
+option (Seastar_USE_OPENSSL
+  "Use OpenSSL rather than GnuTLS for cryptographic operations, including TLS"
+  OFF)
+
+if (Seastar_USE_OPENSSL)
+  set(Seastar_USE_GNUTLS OFF)
+else()
+  set(Seastar_USE_GNUTLS ON)
```
tchaikov, line 99: could you please add space after `set` and `else`? just to be consistent with the rest of this CMake script.

src/net/tls-impl.cc
```diff
@@ -0,0 +1,740 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright 2015 Cloudius Systems
```
tchaikov, line 19: This header looks like a copy-and-paste. But in general, the copyright notice should reflect the original employer under whom this code was created, as work produced during employment typically belongs to the employer. If, and only *if*, there was a subsequent transfer of copyright to ScyllaDB (formerly known as Cloudius Systems), please update the header to use the current company name 'ScyllaDB' rather than 'Cloudius Systems'.

src/net/ossl.cc
```diff
@@ -79,6 +79,8 @@ module seastar;
 #include "net/tls-impl.hh"
 #endif
 
+template <> struct fmt::formatter<seastar::tls::session> : fmt::ostream_formatter {};
```
tchaikov, line 82: is it possible to implement the specialization of `fmt::fornmatter<seastar::tls::session>` without the `operator<<`? it would be better if we can avoid adding more `operator<<`, for two reasons:

1. avoid adding the unused `operator<<`
2. avoid the overhead of using a temporary ostream for formatting `seastar::tls::session`.

src/net/ossl.cc
```diff
@@ -0,0 +1,2068 @@
...
+
+namespace std {
+
+template<>
+struct is_error_code_enum<seastar::ossl_errc> : true_type {};
+
+}
+
+template<>
+struct fmt::formatter<seastar::ossl_errc> : public fmt::formatter<std::string_view> {
```
tchaikov, line 98: instead of inheriting from `fmt::formatter<std::string_view>` , i'd suggest inheriting from `fmt::formatter<string_view>`. please see https://github.com/scylladb/scylladb/commit/168ade72f8137b84b1f07b11d7d1bafd25adf5ba for more details. we switched from this approach to the proposed one in the referenced commit in scylladb a while ago.

src/net/ossl.cc
```diff
@@ -770,6 +772,8 @@ class session : public enable_shared_from_this<session>, public session_impl {
     session(session_type t, shared_ptr<tls::certificate_credentials> creds,
             std::unique_ptr<net::connected_socket_impl> sock, tls_options options = {})
       : _sock(std::move(sock))
+      , _local_address(fmt::format("{}", _sock->local_address()))
```
tchaikov, line 775: nit, could use `fmt::to_string(_sock->local_address())` instead. simpler this way. and slightly more performant.

src/net/ossl.cc
```diff
@@ -770,6 +772,8 @@ class session : public enable_shared_from_this<session>, public session_impl {
...
+      , _remote_address(fmt::format("{}", _sock->remote_address()))
```
tchaikov, line 776: ditto.

src/net/ossl.cc
```diff
@@ -837,11 +841,25 @@ class session : public enable_shared_from_this<session>, public session_impl {
         assert(_output_pending.available());
     }
 
+    friend std::ostream & operator<<(std::ostream &os, const session & session) {
+        fmt::print(os, "{}:{}:{} -",
+                   session.get_type_string(),
+                   session.local_address(),
+                   session.remote_address());
+        return os;
```
tchaikov, line 849: as explained above, i'd prefer avoid adding more `operator<<` overloads unless they are actually necessary and used.

src/net/ossl.cc
```diff
@@ -1084,17 +1118,21 @@ class session : public enable_shared_from_this<session>, public session_impl {
...
+        tls_log.trace("{} wait_for_input", *this);
         // If we already have data, then it needs to be processed
         if (!_input.empty()) {
+            tls_log.trace("{} wait_for_input: input not empty", *this);
             return make_ready_future();
         }
         return _in.get()
           .then([this](buf_type buf) {
               // Set EOF if it's empty
+              tls_log.debug("{} wait_for_input: buffer {}empty", *this, buf.empty() ? "is ": "");
```
tchaikov, line 1130: i don't really understand this logging message:

- if `buf` is empty, we print "wait_for_input: buffer is empty" -- looks good
- if `buf` is not empty, we print "wait_for_input: buffer empty" -- what does this mean? shall we print "wait_for_input: buffer is not empty" or "wait_for_input: buffer not empty" instead?

.github/workflows/tests.yaml
```diff
@@ -47,3 +49,16 @@ jobs:
...
+    uses: ./.github/workflows/test.yaml
+    strategy:
+      fail-fast: false
+    with:
+      compiler: clang++
+      standard: 23
+      mode: debug
+      enables: --enable-cxx-modules
+      enable-ccache: false
+      crypto_provider: OpenSSL
```
tchaikov, line 64: piggyback the tests with OpenSSL / GnuTLS might not be a great idea. because, we only enable the test step in https://github.com/scylladb/seastar/blob/613d8b39e0d6e8af4b4aff5cffb639a7016c8f7c/.github/workflows/test.yaml#L110-L112 if "cxx-modules" is *not* enabled. in other words, these two jobs do not run any of the unit tests. so i'd suggest adding a dedicated job in `tests.yaml` for building with the OpenSSL backend, and keep the existing job of `build_with_cxx_modules` intact. 

please allow me to provide more context here: because C++20 modules is currently an experimental feature, not all Seastar facilities are exposed as in the "seastar" C++20 module at this moment, and we only have a single "hello_cxx_module" test for testing the build with C++20 module. none of the unit tests is built with the "seastar" C++20 module at the time of writing.

src/net/ossl.cc
```diff
@@ -0,0 +1,2068 @@
...
+            // Extract the errno using ERR_GET_REASON and throw a std::generic_category
+            return std::system_error(
+                ERR_GET_REASON(err_code),
+                std::generic_category(),
+                fmt::format("{}: {}", msg, error_codes));
+        }
+        return std::system_error(
+            static_cast<int>(err_code),
+            tls::error_category(),
+            fmt::format("{}: {}", msg, error_codes));
```
tchaikov, line 171: nit, could reduce the indent level by removing the "else" clause, because of the early return above.

src/net/ossl.cc
```diff
@@ -0,0 +1,2068 @@
...
+        // Store a pointer to the current connection certificate within the impl instance
+        auto cert = X509_STORE_CTX_get_current_cert(store_ctx);
+        X509_up_ref(cert);
+        credential_impl->_last_cert = x509_ptr(cert);
+        return preverify_ok;
+    }
+
+    impl() : _creds([] {
+        auto store = X509_STORE_new();
+        if(store == nullptr) {
```
tchaikov, line 352: could you please add a space after `if`?

tests/unit/tls_test.cc
tchaikov: nit, might want to use `BOOST_REQUIRE_EQUAL()` for better postmortem debugging experience. as Boost.Test prints out the lhs and rhs of the comparison f the check fails.

src/net/ossl.cc
michael-redpanda: Thanks, that was my mess up.  Fixing it

michael-redpanda:
Force push [db318da](https://github.com/scylladb/seastar/commit/db318daadfd31fadb14c45c8f128c1820a0a1a02):

* Addressed PR comments

michael-redpanda:
Force push [245d15c](https://github.com/scylladb/seastar/commit/245d15c8239a7db7f48c23e47dfe88171c716611):

* Rebased off master to fix merge conflicts introduced by https://github.com/scylladb/seastar/commit/6f39b89dbda812c7985918d3d8d6d7f7792b0619#diff-2f6778574276aef7b957daacf9904fc252fe1b44d45819bc89daafce9dac3972R1042

michael-redpanda:
Force push [cab469f](https://github.com/scylladb/seastar/commit/cab469f91be311fec34400502796a3eefe4c282c):

* Fixed typo

michael-redpanda:
Force push [47dfb11](https://github.com/scylladb/seastar/commit/47dfb11081910d5e70c02498d031633f66b1d1ec):

* Fixed logic of `get_security_level`
  * Returning anything >1 when using OpenSSL 3.0 will introduce key restrictions that may cause Seastar to be unable to connect with other endpoints
* Addressed error in `test_reload_certificates_with_only_shard0_notify`

avikivity:
Note test failures.

I'm conflicted here. On the one hand, the tls implementation has a user footprint in priority string or equivalent openssl config. On the other hand, that's a tiny difference and they're otherwise equivalent. OpenSSL would be my choice as the only supported library if there wasn't the user facing stuff.

Can we translate the priority string to openssl config? I asked Claude and it spewed a long python script, maybe that's good enough.

michael-redpanda:
> Note test failures

It appears all failures are from `Seastar.unit.rpc`, which I don't believe interacts with the TLS.  Were they flaky before?

> Can we translate the priority string to openssl config? I asked Claude and it spewed a long python script, maybe that's good enough.

It definitely could be done by setting all of these within an [OpenSSL config file](https://docs.openssl.org/master/man5/config/#openssl-library-configuration).  You'd just have to be careful to tell OpenSSL where to find it so it doesn't attempt to locate and use the system default one (unless that's desired).

michael-redpanda:
> Can we translate the priority string to openssl config? I asked Claude and it spewed a long python script, maybe that's good enough.

I guess your question is "can I take the input to`set_priority_string` and turn that into the appropriate OpenSSL API calls"?  It could be doable, but would take a while and I'm not sure if that's something anyone would really want to maintain.

avikivity:
> > Can we translate the priority string to openssl config? I asked Claude and it spewed a long python script, maybe that's good enough.
> 
> I guess your question is "can I take the input to`set_priority_string` and turn that into the appropriate OpenSSL API calls"? It could be doable, but would take a while and I'm not sure if that's something anyone would really want to maintain.

Is there no equivalent textual configuration?

I wouldn't want to maintain it, but neither do I want to maintain two different tls implementations.

michael-redpanda:
> Is there no equivalent textual configuration?

AFAIK only through the OpenSSL config file which can be loaded either automatically by OpenSSL when it is initialized or can be controlled by the user by calling [`OSSL_LIB_CTX_load_config`](https://docs.openssl.org/master/man3/OSSL_LIB_CTX/). And within there would be number of key-value pairs that would need to be set correctly.

michael-redpanda:
Hi again @avikivity and @elcallio .  Apologies for such a long delay in updating this PR.  I've made some updates and rebased off of master.

michael-redpanda:
Hi again @avikivity and @elcallio .  Was wondering if there was anything I could do to assist in regards to this PR?

avikivity:
I prefer replacing gnutls with openssl rather than adding yet another option. But perhaps we can't due to those minor configuration issues.

Maybe we can start a long deprecation cycle during which users will be encouraged to update their configuration.

avikivity:
If we indeed start a deprecation cycle, it will have to be run-time selectable (so users can update their configuration, change the selector, restart). I don't know how difficult that is.

elcallio:
From an API standpoint, I guess replacing the prio strings with some sort of abstracted algorithm-selection is the biggest hurdle (std::set<tls_algo> allow, forbid?). Once users migrate this, the rest should be fairly transparent/equivalent.

src/net/tls-impl.hh
```diff
@@ -0,0 +1,232 @@
...
+    virtual future<temporary_buffer<char>> get() = 0;
+    virtual void close() = 0;
+    virtual future<std::optional<session_dn>> get_distinguished_name() = 0;
+    virtual seastar::net::connected_socket_impl & socket() const = 0;
+    virtual future<std::vector<subject_alt_name>> get_alt_name_information(std::unordered_set<subject_alt_name_type>) = 0;
+    virtual future<bool> is_resumed() = 0;
+    virtual future<session_data> get_session_resume_data() = 0;
+    virtual future<std::vector<certificate_data>> get_peer_certificate_chain() = 0;
+    virtual future<std::optional<sstring>> get_selected_alpn_protocol() = 0;
+};
```
avikivity, line 56: Looks like this is easy to make run-time (start-time) selectable.

avikivity: Lack of virtual destructor is justified by shared_ptr below, but please add a comment.

But: is shared_ptr called for? why not unique_ptr or lw_shared_ptr? Who's sharing?

src/net/tls-impl.hh
```diff
@@ -0,0 +1,232 @@
...
+    }
+    session_ref(session_ref&&) = default;
+    session_ref(const session_ref&) = default;
+    ~session_ref();
+
+    session_ref& operator=(session_ref&&) = default;
+    session_ref& operator=(const session_ref&) = default;
+
+    shared_ptr<session_impl> _session;
+};
```
avikivity, line 71: `using session_ref = ` would be sufficient, no?

src/net/tls-impl.hh
```diff
@@ -0,0 +1,232 @@
...
+
+class tls_connected_socket_impl : public net::connected_socket_impl, public session_ref {
+public:
```
avikivity, line 74: Why is-a relationship between tls_connected_socket_impl and session_ref?

src/net/tls-impl.hh
```diff
@@ -0,0 +1,232 @@
...
+    future<std::optional<sstring>> get_selected_alpn_protocol() {
+        return _session->get_selected_alpn_protocol();
+    }
+};
+
+
+class tls_connected_socket_impl::source_impl: public data_source_impl, public session_ref {
+public:
+    using session_ref::session_ref;
+private:
```
avikivity, line 149: I see where the sharing comes from. But lw_shared_ptr is sufficient. Or keep a reference to the connected_socket_impl.

src/net/tls-impl.hh
elcallio: This is in current TLS impl. and no. The session ref has a special property that if it is the last one to release the session (i.e. it will cause destruct), it "resurrects" the session and creates a background BYE handshake cycle.

src/net/tls-impl.hh
elcallio: Input and output.

src/net/tls-impl.hh
elcallio: socket_impl is stored with a unique ptr -> cannot be shared. Our IO design allows the input/output streams to exists independently of the creating socket. Thus the need to share the underlying session.

src/net/tls-impl.hh
elcallio: Because it needs to create input/output sinks when asked.

include/seastar/net/tls.hh
```diff
@@ -212,13 +212,53 @@ namespace tls {
...
          * https://gnutls.org/manual/html_node/Priority-Strings.html
          *
          * Allows specifying order and allowance for handshake alg.
          */
         void set_priority_string(const sstring&);
+#endif
+
+#ifdef SEASTAR_USE_OPENSSL
+        /**
+         * Used to set the cipher string for TLS versions 1.2 and below
```
avikivity, line 227: If we are to co-support the two implementations, then we should always compile these, but throw something instead. Maybe std::bad_cast (we're "casting" gnutls to openssl or something).

include/seastar/net/tls.hh
```diff
@@ -212,13 +212,53 @@ namespace tls {
...
+        void set_cipher_string(const sstring&);
+
+        /**
+         * Used to set the cipher suites to use for TLSv1.3
+         *
+         * See https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_ciphersuites.html
+         * for documentation on the format of the ciphersuites string.
+         */
+        void set_ciphersuites(const sstring&);
+
```
avikivity, line 241: std::string_view is more flexible.

include/seastar/net/tls.hh
```diff
@@ -629,6 +685,12 @@ namespace tls {
     extern const int ERROR_NO_CIPHER_SUITES;
     extern const int ERROR_DECRYPTION_FAILED;
     extern const int ERROR_MAC_VERIFY_FAILED;
```
avikivity, line 693: Are those error numbers compatible? And specified somewhere?

include/seastar/net/tls.hh
```diff
@@ -377,6 +428,11 @@ namespace tls {
         sstring _priority;
         std::vector<uint8_t> _session_resume_key;
         std::vector<sstring> _alpn_protocols;
+        sstring _cipher_string;
+        sstring _ciphersuites;
+        bool _enable_server_precedence = false;
+        std::optional<tls_version> _min_tls_version;
+        std::optional<tls_version> _max_tls_version;
```
avikivity, line 441: Let's package those in openssl/gnutls sub-structs.

src/net/ossl.cc
```diff
@@ -0,0 +1,2288 @@
+/*
```
avikivity, line 1: Please use "openssl.cc" to make it immediately understood.

src/net/ossl.cc
```diff
@@ -0,0 +1,2288 @@
...
+        if (RAND_priv_bytes(key.data(), N) <= 0) {
+            throw make_ossl_error("Failed to generate key");
+        }
+    }
+};
+
+// sufficiently large enough to avoid collision with OpenSSL BIO controls
+#define BIO_C_SET_POINTER 1000
+// Index into ex data for SSL structure to fetch a pointer to session
+#define SSL_EX_DATA_SESSION 0
```
avikivity, line 335: Please use C++ constexpr variables instead.

tests/unit/tls_test.cc
avikivity: Was the test enshrining incorrect behavior?

michael-redpanda:
> If we indeed start a deprecation cycle, it will have to be run-time selectable (so users can update their configuration, change the selector, restart). I don't know how difficult that is.

I believe this would require a significant change to the PR to support this.  If this is what is desired, more than happy to look into it, but I would have to close this PR and open a new one.

michael-redpanda:
@avikivity  I haven't looked too deep into it yet, but I think making it runtime selectable would be a change to the API

avikivity:
> @avikivity I haven't looked too deep into it yet, but I think making it runtime selectable would be a change to the API

Why? If the API suits both openssl and gnutl statically, it should suit them dynamically, no?

michael-redpanda:
> > @avikivity I haven't looked too deep into it yet, but I think making it runtime selectable would be a change to the API
> 
> Why? If the API suits both openssl and gnutl statically, it should suit them dynamically, no?

The current implementation does not support building both at the same time, it has to be one or the other.  Both GnuTLS and OpenSSL implement many of the same named methods, e.g. `tls::wrap_client`, `tls::wrap_server`, etc.  Both also define the error codes like `tls::ERROR_UNKNOWN_COMPRESSION_ALGORITHM`, etc.

avikivity:
> > > @avikivity I haven't looked too deep into it yet, but I think making it runtime selectable would be a change to the API
> > 
> > 
> > Why? If the API suits both openssl and gnutl statically, it should suit them dynamically, no?
> 
> The current implementation does not support building both at the same time, it has to be one or the other. Both GnuTLS and OpenSSL implement many of the same named methods, e.g. `tls::wrap_client`, `tls::wrap_server`, etc.

Well, the entry points need to be disambiguated:

```c++
future<connected_socket> tls::wrap_client(shared_ptr<certificate_credentials> cred, connected_socket&& s, tls_options options) {
    session::session_ref sess(make_lw_shared<session>(session::type::CLIENT, std::move(cred), std::move(s),  options));
    connected_socket sock(std::make_unique<tls_connected_socket_impl>(std::move(sess)));
    return make_ready_future<connected_socket>(std::move(sock));
}
```

Here call a factory to create sess instead of hardcoding the implementation.


> Both also define the error codes like `tls::ERROR_UNKNOWN_COMPRESSION_ALGORITHM`, etc.


We'd change them from const globals to non-const globals intialized at library load time.

I don't want to cause unnecessary churn, but I also don't want to break running configurations.

avikivity:
> > > > @avikivity I haven't looked too deep into it yet, but I think making it runtime selectable would be a change to the API
> > > 
> > > 
> > > Why? If the API suits both openssl and gnutl statically, it should suit them dynamically, no?
> > 
> > 
> > The current implementation does not support building both at the same time, it has to be one or the other. Both GnuTLS and OpenSSL implement many of the same named methods, e.g. `tls::wrap_client`, `tls::wrap_server`, etc.
> 
> Well, the entry points need to be disambiguated:
> 
> ```c++
> future<connected_socket> tls::wrap_client(shared_ptr<certificate_credentials> cred, connected_socket&& s, tls_options options) {
>     session::session_ref sess(make_lw_shared<session>(session::type::CLIENT, std::move(cred), std::move(s),  options));
>     connected_socket sock(std::make_unique<tls_connected_socket_impl>(std::move(sess)));
>     return make_ready_future<connected_socket>(std::move(sock));
> }
> ```
> 
> Here call a factory to create sess instead of hardcoding the implementation.
> 
> > Both also define the error codes like `tls::ERROR_UNKNOWN_COMPRESSION_ALGORITHM`, etc.
> 
> We'd change them from const globals to non-const globals intialized at library load time.
> 
> I don't want to cause unnecessary churn, but I also don't want to break running configurations.

Another question is how to roll out a change across the cluster. I guess if the configuration is local to a node, you'd update it as part of each individual node upgrade.

michael-redpanda:
@avikivity If this is the direction your team wants to go, I may have to take this PR down and start from scratch.  I'll probably break this up into two separate PRs.  A 'staging' PR that will refactor the GnuTLS stuff out of `tls.hh` and then a second one to introduce OpenSSL.  Probably should have done that from the start.

avikivity:
> @avikivity If this is the direction your team wants to go, I may have to take this PR down and start from scratch. I'll probably break this up into two separate PRs. A 'staging' PR that will refactor the GnuTLS stuff out of `tls.hh` and then a second one to introduce OpenSSL. Probably should have done that from the start.

What were your plans wrt the small incompatibilities between openssl and gnutls? Do they not impact you?

PR #2575, opened 2024-12-11
smp: disable mlock on overcommit

tomershafir:
Memory locking keeps pages in physical memory, and decreases the amount of pages available for swapping on an overcommit setup. Thus, it can decrease fairness, increase memory pressure, and cause OOMs. So, it is less suitable for a cooperative overcommit setup.

This change disables memory locking on overcommit (overprovisioned - I hope I can rename that config). Also, it reuses the mlock variable. I assume that this change shouldn't break reasonable configs.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3778,7 +3778,7 @@ reactor_options::reactor_options(program_options::option_group* parent_group)
     , kernel_page_cache(*this, "kernel-page-cache", false,
                 "Use the kernel page cache. This disables DMA (O_DIRECT)."
                 " Useful for short-lived functional tests with a small data set.")
-    , overprovisioned(*this, "overprovisioned", "run in an overprovisioned environment (such as docker or a laptop); equivalent to --idle-poll-time-us 0 --thread-affinity 0 --poll-aio 0")
+    , overprovisioned(*this, "overprovisioned", "run in an overprovisioned environment (such as docker or a laptop); equivalent to --idle-poll-time-us 0 --thread-affinity 0 --poll-aio 0 --lock-memory 0")
     , abort_on_seastar_bad_alloc(*this, "abort-on-seastar-bad-alloc", "abort when seastar allocator cannot allocate memory")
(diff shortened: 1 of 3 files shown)


src/core/reactor.cc
```diff
@@ -3814,7 +3814,7 @@ smp_options::smp_options(program_options::option_group* parent_group)
     , memory(*this, "memory", std::nullopt, "memory to use, in bytes (ex: 4G) (default: all)")
     , reserve_memory(*this, "reserve-memory", {}, "memory reserved to OS (if --memory not specified)")
     , hugepages(*this, "hugepages", {}, "path to accessible hugetlbfs mount (typically /dev/hugepages/something)")
-    , lock_memory(*this, "lock-memory", {}, "lock all memory (prevents swapping)")
+    , lock_memory(*this, "lock-memory", {}, "lock all memory (prevents swapping). Disable for overprovisioned environments")
```
xemul, line 3817: The help text for `--overprovisioned` should be updated too

avikivity:
Wasn't mlock default off for all environments?

src/core/reactor.cc
```diff
@@ -4332,7 +4332,7 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
         hugepages_path = smp_opts.hugepages.get_value();
     }
     auto mlock = false;
-    if (smp_opts.lock_memory) {
+    if (smp_opts.lock_memory && !reactor_opts.overprovisioned) {
         mlock = smp_opts.lock_memory.get_value();
```
avikivity, line 4336: It's default off, so why this extra work?

If someone specified both --overprovisioned (which really should have been named --overcommitted) and --lock-memory 1, we can assume they know what they're doing. Maybe only the CPU is overcommitted.

tomershafir: ok, I agree. Then I think we can tell the same for `thread_afinity` option. A user who knows what he is doing can overcommit memory and pin vCPUs. So I think it can make sense to only log a warning in both cases. wdyt?

Issue #2231, opened 2024-05-09
create_scheduling_group / scheduling_group_key_create are not safe to run in parallel


piodul:
I observed that running those two functions in parallel may lead to data of a scheduling group key initialized twice.

Those functions work like this:

- create_scheduling_group
	- Allocate a new SG ID
	- _Preemption may happen here_
	- On all shards:
		- Initialize storage for the SG
		- Read the current SG key count and store in `num_keys` variable
		- _Preemption may happen here_
		- Initialize SG keys for the new SG up to `num_keys`

- scheduling_group_key_create
	- Allocate a new SG key ID
	- _Preemption may happen here_
	- On all shards:
		- Initialize the SG key config
		- _Preemption may happen here_
		- For all current SGs, initialize the new SG key

While I am not exactly sure what the sequence of events for my case was, looking at the implementation I think the following can happen:

- [create_scheduling_group] allocates a new SG ID
- [scheduling_group_key_create] allocates a new SG key ID
- [create_scheduling_group] reads the SG key count and stores to `num_keys` var
- [scheduling_group_key_create] initializes storage for the new SG key
- [create_scheduling_group] initializes storage for the new SG
- [scheduling_group_key_create] initializes the new SG key for all existing SGs
- [create_scheduling_group] initializes the new SG and all existing SG keys for it, including the new one.

Perhaps there might be other kinds of races. I think that operations like `create_scheduling_group` and `scheduling_group_key_create` should be serialized.

Relevant code:

https://github.com/scylladb/seastar/blob/d3657ec70e934a1e747ed9e3e970bd8102648bf6/src/core/reactor.cc#L5016-L5030
https://github.com/scylladb/seastar/blob/d3657ec70e934a1e747ed9e3e970bd8102648bf6/src/core/reactor.cc#L4859-L4926

avikivity:
Makes sense. These are not data plane operations so can safely be serialized.

PR #2585, opened 2024-12-16
scheduling_group: fix race between scheduling group and key creation

mlitvk:
The functions create_scheduling_group and scheduling_group_key_create
may have unexpected results when run concurrently.

when creating a SG key, we:
1. allocate a key
2. insert the key config
3. for all SGs, initialize data for the SG and key

with preemptions possible in between each step.

Another process could see an intermediate state where a key was created
but not fully initialized yet. This may lead to the data being
initialized twice for the new key, or functions operating on
uninitialized key data.

To solve this we introduce a shared_mutex which is locked exclusively
when creating a new key, and is held with shared access when creating a
scheduling group and when a consistent read access of all keys is required.

Fixes https://github.com/scylladb/seastar/issues/2231

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -47,6 +47,7 @@
 #include <seastar/core/scheduling_specific.hh>
 #include <seastar/core/seastar.hh>
 #include <seastar/core/semaphore.hh>
+#include <seastar/core/shared_mutex.hh>
 #include <seastar/core/sstring.hh>
 #include <seastar/core/temporary_buffer.hh>
 #include <seastar/core/thread_cputime_clock.hh>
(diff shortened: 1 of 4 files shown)


include/seastar/core/reactor.hh
```diff
@@ -282,6 +283,7 @@ private:
 
     boost::container::static_vector<std::unique_ptr<task_queue>, max_scheduling_groups()> _task_queues;
     internal::scheduling_group_specific_thread_local_data _scheduling_group_specific_data;
+    shared_mutex _scheduling_group_keys_mutex;
```
avikivity, line 286: This presumes that all calls to the scheduling group management functions happen on the same shard.

avikivity: No, the callers are shard-local already.

PR #2586, opened 2024-12-17
metrics: Remove scollectd backend

StephanDollberg:
It's no longer used. See
https://github.com/scylladb/seastar/pull/2561#issuecomment-2538057831 and following.

Remove it which will allow simplifying some parts of the shared metrics backend.


tchaikov:
see also scylladb/scylladb#17356 which removes collectd support from ScyllaDB.

travisdowns:
@avikivity directionally, WDYT on this one?

StephanDollberg:
Rebased and added a small commit to make metdata info smaller.

travisdowns:
@scylladb/seastar-maint sylladb has dropped the scollectd backend, can we drop it drop seastar? See also [previous comment thread](https://github.com/scylladb/seastar/pull/2561#issuecomment-2538057831) indicating (to me) that this is desirable.

avikivity:
Wasn't completely dropped:

```
test/manual/hint_test.cc:#include <seastar/core/scollectd_api.hh>
```

StephanDollberg:
I can't see this path in the repo? Is it a local file of yours?

StephanDollberg:
Oh wait this is a scylladb path

travisdowns:
> Wasn't completely dropped:
> 
> ```
> test/manual/hint_test.cc:#include <seastar/core/scollectd_api.hh>
> ```

This looks dead, that file doesn't do any scollectd stuff, probably an accidental copy/paste inclusion from when that file was first created.

avikivity:
Yes it's our tree. I'll remove it.

I think we should accept the --collectd option and ignore it, or installations that have it on their command line will fail to start.

avikivity:
https://github.com/scylladb/scylladb/pull/31688

PR #2589, opened 2024-12-18
json: add moving push to json_list

travisdowns:
When building up json lists in memory we have only the `push` method which takes const T&, but one taking an rvalue reference and moves into the list would be useful for heavy elements.


tchaikov (approved):
lgtm

PR #2590, opened 2024-12-19
perf_tests: make column generic

travisdowns:
This introduces a concept of a "column" in the perf test output, which given a result is a rest metric which can be output as a result of the test (as an actual column in stdout and md views, or as a property in the json writer).

This means that all 3 writers will keep a consistent view of the available metrics (cycles was missing from md output before this change), and it also enables things like adjusting the columns shown in the future.

### DIFF
--- a/tests/perf/perf_tests.cc
+++ b/tests/perf/perf_tests.cc
@@ -24,6 +24,7 @@
 #include <cstdio>
 #include <fstream>
 #include <regex>
+#include <type_traits>
 
 #include <boost/range.hpp>
 #include <boost/range/adaptors.hpp>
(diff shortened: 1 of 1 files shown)


travisdowns:
This solves missing cycles and somehow reduces duplication "in principle", though the duplication was mostly limited to horizontal duplication on a single line (but still error prone).

tests/perf/perf_tests.cc
```diff
@@ -41,6 +41,7 @@
 #include <seastar/core/reactor.hh>
 
 #include <signal.h>
+#include <type_traits>
```
tchaikov, line 44: could you please move this `#include` up, so it is closer to the other standard library headers. would be better to group similar headers together.

tests/perf/perf_tests.cc
```diff
@@ -203,6 +204,119 @@ static inline std::ostream& operator<<(std::ostream& os, duration d)
...
+        constexpr auto is_double = std::is_same_v<result_t, double>;
+        constexpr auto is_duration = std::is_same_v<result_t, duration>;
+        if constexpr (is_integral || is_double || is_duration) {
+            static constexpr std::string_view fmt_str = is_double ? "{:>{}.{}f}": "{:>{}}";
+            text_printer = [=](const column& c, std::FILE *file, const result& r) {
+                fmt::print(file, fmt_str, fn(r), c.width, c.prec);
+            };
+        } else {
+            static_assert(false, "unsupported return type");
+        }
```
tchaikov, line 232: instead of having this `static_assert()` in the `else` branch, IMHO, would be simper to put 
```c++
static_assert(is_integral || is_double || is_duration, "unsupported return type");
# initiailize text_printer and double_fn
```

what do you think?

tests/perf/perf_tests.cc
```diff
@@ -203,6 +204,119 @@ static inline std::ostream& operator<<(std::ostream& os, duration d)
...
+        double_fn = [fn](const result& r) {
+            if constexpr(is_duration) {
```
tchaikov, line 234: could you please add a space after `constexpr` to be more consistent with the tree?

tests/perf/perf_tests.cc
```diff
@@ -203,6 +204,119 @@ static inline std::ostream& operator<<(std::ostream& os, duration d)
...
+ * This allows all printers to share a common view of the available
+ * columns.
+ */
+struct column {
+    static constexpr int default_width = 11;
+
+    using print_fn = std::function<void(const column&, std::FILE *file, const result& r)>;
+
+    template <typename F>
+    column(sstring header, int prec, F fn) : header{header}, prec{prec} {
```
tchaikov, line 220: nit, i think `std::string_view` is good enough here. because what we want is just a type which is convertible to `std::string`. and the callers of this constructor always pass a string literal for `header`. if you agree with my reasoning, we might need to 

- update the member variable of `header` to be `std::string_view` as well.
- s/c.header/std::string{c.header}/ in `json_printer::print_result()`

tests/perf/perf_tests.cc
```diff
@@ -203,6 +204,119 @@ static inline std::ostream& operator<<(std::ostream& os, duration d)
...
+    for (auto& c : cols) {
+        fmt::print(out, "{}", middle_delim);
+        c.text_printer(c, out, r);
+    }
+    fmt::print(out, "{}\n", end_delim);
+}
+
+const std::vector<column> text_only_columns{
+    {"iterations" , 0, [](const result& r) { return r.total_iterations / r.runs; }}
+};
```
tchaikov, line 300: i am inclined to reduce the usage of global variables for better maintainability. probably we can move this constant into the lambda which initializes `text_columns` ?  like
```c++
const  std::vector<column> text_columns = [] {
    std::vector<column> ret {
        "iterations" , 0, [](const result& r) { return r.total_iterations / r.runs; }}
    };
    ret.insert(...);
    return ret;
}();
```

tests/perf/perf_tests.cc
```diff
@@ -203,6 +204,119 @@ static inline std::ostream& operator<<(std::ostream& os, duration d)
...
+
+// columns for json ouput
+const std::vector<column> json_columns{
+    {"median"     , 0, [](const result& r) { return duration { r.median };       }},
```
tchaikov, line 304: nit, could remove some of the spaces while keep the variables in the initializer lists aligned like:
```c++
    {"median", 0, [](const result& r) { return duration { r.median }; }},
    {"mad"   , 0, [](const result& r) { return duration { r.mad };    }},
    {"min"   , 0, [](const result& r) { return duration { r.min };    }},
    {"max"   , 0, [](const result& r) { return duration { r.max };    }},
    {"allocs", 3, [](const result& r) { return r.allocs;              }},
    {"tasks" , 3, [](const result& r) { return r.tasks;               }},
    {"inst"  , 1, [](const result& r) { return r.inst;                }},
    {"cycles", 1, [](const result& r) { return r.cycles;              }},
```

or you want to be more future-proof and preallocate more spaces for columns with wider headers?

tests/perf/perf_tests.cc
```diff
@@ -203,6 +204,119 @@ static inline std::ostream& operator<<(std::ostream& os, duration d)
...
+    sstring header;
+
+    // width for the column in text output
+    int width = 11;
+
+    // precision in case of double
+    int prec = 3;
+
+    // used by stdout and md formats to print as text
+    print_fn text_printer;
```
tchaikov, line 256: nit, i'd suggest s/text_printer/print_text/ as we use this member variable as a function, so would be more readable to name it like a verb instead of a noun.

tests/perf/perf_tests.cc
```diff
@@ -203,6 +204,119 @@ static inline std::ostream& operator<<(std::ostream& os, duration d)
...
+
+    // used by json format to extract double result
+    std::function<double(const result&)> double_fn;
```
tchaikov, line 259: nit, for the similar reason, i'd suggest s/double_fn/to_double/ .

tests/perf/perf_tests.cc
```diff
@@ -203,6 +204,119 @@ static inline std::ostream& operator<<(std::ostream& os, duration d)
...
+    }
+    fmt::print(out, "{}\n", end_delim);
+}
+
+const std::vector<column> text_only_columns{
+    {"iterations" , 0, [](const result& r) { return r.total_iterations / r.runs; }}
+};
+
+// columns for json ouput
+const std::vector<column> json_columns{
```
tchaikov, line 303: this variable is not used outside of this translation unit, so probably we can mark it with `static` ? like:
```c++
static const std::vector<column> json_columns{
  // ...
};
```

tchaikov (commented):
thank you for the nice fix and generalization. left a couple nit picks.

tests/perf/perf_tests.cc
travisdowns: Good eye, moved (IDE auto-adds this in the wrong place).

tests/perf/perf_tests.cc
travisdowns: In general members of type `std::string_view` are not safe since it is a view type and non-owning. It may be safe in special cases, e.g., when it always points into another (owning) member of the same object, or in this case where you've audited by hand that all callers pass an object of static (infinite) lifetime, but that's something that's easy to change and a big potential footgun.

It is for ~no benefit here since there are < 10 of these objects statically constructed for the entire lifetime of the process, so I lean heavily towards "no footguns".

tests/perf/perf_tests.cc
travisdowns: Yeah it makes sense: this structure was vestigial from one where was more than one if/else and constexpr if was needed for it compile. Changed.

tests/perf/perf_tests.cc
travisdowns: good eye, fixed

tests/perf/perf_tests.cc
travisdowns: I agree, renamed to verb.

tests/perf/perf_tests.cc
travisdowns: Renamed.

tests/perf/perf_tests.cc
travisdowns: done

tests/perf/perf_tests.cc
travisdowns: Marked static.

tests/perf/perf_tests.cc
travisdowns: Fixed. It was just like that because `"iterations"` was in the list at one point which has a long name and long result function so it was aligned to that width.

travisdowns (commented):
Thanks @tchaikov for the review (somehow I had missed it until now).

All feedback addressed, except one I disagreed with, see what you think.

tchaikov (approved):
lgtm

tchaikov:

travisdowns:
Thanks all for the reviews!

PR #2592, opened 2024-12-22
build,github: add a step to check include style 

tchaikov:
Previously, we do not enforce the policy of including Seastar headers using angle brackets instead of double quotes in our github workflow. In this change, a new step is added to check include style when building the Dev mode. This allows our CI to fail if a change violates this policy.

### DIFF
--- a/cmake/CheckIncludeStyle.cmake
+++ b/cmake/CheckIncludeStyle.cmake
@@ -1,0 +1,17 @@
+# seastar_check_include_style() enforces that all source and header files under
+# specified directories include the headers with predefined list of prefixes
+# with angle brackets instead of quotes.
+
+find_package (Python3 COMPONENTS Interpreter)
+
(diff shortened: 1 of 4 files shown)


cmake/CheckIncludeStyle.cmake
```diff
@@ -0,0 +1,17 @@
+# seastar_check_include_style() enforces that all source and header files under
+# specified directories include the headers with predefined list of prefixes
+# with angle brackets instead of quotes.
+
+find_package (Python3 COMPONENTS Interpreter)
+
+function (seastar_check_include_style target library)
+  get_target_property (sources ${library} SOURCES)
+  set (check-target "${target}-${library}")
+  message("${sources}")
```
tchaikov, line 10: please remove this message, which was added for debugging.

tchaikov: created #2593

PR #2595, opened 2024-12-23
rwlock: add try_hold_{read,write}_lock methods

bhalevy:
Add synchronous try_hold_* methods so that users
can optimize contention-free paths by avoiding
extra-allocations that might be needed for waiting on the lock, if it's taken.


bhalevy:
For planned usage, see https://github.com/bhalevy/scylla/commits/storage_proxy-await-pending-writes-2024-12-23/

PR #2597, opened 2024-12-23
file: add file_system_space

bhalevy:
Return space_info for the filesystem identified
by the given file name.

space_info provides simpler and standard space information about the filesystem, in contrast to the posix statvfs which requires knowledge about the how to convert
f_block to bytes by multiplying by f_frsize.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2263,6 +2263,17 @@ reactor::fstatfs(int fd) noexcept {
     });
 }
 
+future<std::filesystem::space_info>
+reactor::file_system_space(std::string_view pathname) noexcept {
+    auto sr = co_await _thread_pool->submit<syscall_result_extra<std::filesystem::space_info>>([path = std::filesystem::path(pathname)] {
+        std::error_code ec;
+        auto si = std::filesystem::space(path, ec);
+        return wrap_syscall(ec.value(), si);
+    });
+    sr.throw_fs_exception_if_error("std::filesystem::space failed", sstring(pathname));
+    co_return sr.extra;
+}
+
 future<struct statvfs>
 reactor::statvfs(std::string_view pathname) noexcept {
     // Allocating memory for a sstring can throw, hence the futurize_invoke
--- a/tests/unit/file_io_test.cc
+++ b/tests/unit/file_io_test.cc
@@ -19,11 +19,14 @@
  * Copyright (C) 2014-2015 Cloudius Systems, Ltd.
  */
 
+#include <filesystem>
+
 #include <seastar/testing/random.hh>
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
 #include <seastar/testing/test_runner.hh>
 
+#include <seastar/core/reactor.hh>
 #include <seastar/core/seastar.hh>
 #include <seastar/core/semaphore.hh>
 #include <seastar/core/condition-variable.hh>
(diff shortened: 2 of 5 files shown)


include/seastar/core/file-types.hh
```diff
@@ -158,6 +159,18 @@ inline constexpr file_permissions operator&(file_permissions a, file_permissions
     return file_permissions(std::underlying_type_t<file_permissions>(a) & std::underlying_type_t<file_permissions>(b));
 }
 
+/// seastar::space_info is equivalent to std::filesystem::space_info
+/// with renamed members, to prevent a conflict with future::available()
+struct space_info {
```
avikivity, line 164: Why not use std::filesystem::space_info?

include/seastar/core/file-types.hh
```diff
@@ -158,6 +159,18 @@ inline constexpr file_permissions operator&(file_permissions a, file_permissions
...
+    std::uintmax_t capacity;
+    std::uintmax_t free_space;
+    std::uintmax_t available_space;
+
+    bool operator==(const space_info& o) const noexcept {
+        return capacity == o.capacity && free_space == o.free_space && available_space == o.available_space;
+    }
```
avikivity, line 171: = default

src/core/reactor.cc
```diff
@@ -2263,6 +2263,26 @@ reactor::fstatfs(int fd) noexcept {
...
+            sr.throw_fs_exception_if_error("file_system_space failed", pathname);
+            auto& fs_si = sr.extra;
+            return space_info{
+                .capacity = fs_si.capacity,
+                .free_space = fs_si.free,
+                .available_space = fs_si.available
+            };
+        });
+    });
+}
```
avikivity, line 2284: Please use coroutines in new code.

tests/unit/file_io_test.cc
```diff
@@ -947,3 +950,20 @@ SEASTAR_TEST_CASE(test_oversized_io_works) {
...
+        space_info si, si1;
+        do {
+            si = file_system_space(name).get();
+            st = engine().statvfs(name).get();
+            si1 = file_system_space(name).get();
+        } while (si1 != si);
+
+        BOOST_REQUIRE_EQUAL(st.f_blocks * st.f_frsize, si.capacity);
+        BOOST_REQUIRE_EQUAL(st.f_bfree * st.f_frsize, si.free_space);
+        BOOST_REQUIRE_EQUAL(st.f_bavail * st.f_frsize, si.available_space);
```
avikivity, line 967: This isn't guaranteed to pass. If a byte is added to the filesystem after assignment to si and removed after assignment to st, si == si1 && si != st.

There isn't a simple way to test this.

bhalevy: Without syncing the filesystem, it is auto-synced rarely (every 30 seconds?), so we shouldn't see these fluctuations in practice.

include/seastar/core/file-types.hh
bhalevy: I mentioned it in a comment.
The compiler complains about it due to the `available` member conflicting with `future::available`.  I can reproduce and quote the error

tests/unit/file_io_test.cc
avikivity: Who says space is updated on sync?

include/seastar/core/file-types.hh
avikivity: How can it conflict with future::available? It's in a different class.

avikivity: I see the problem.

avikivity: https://github.com/scylladb/seastar/pull/2598

tests/unit/file_io_test.cc
bhalevy: Good point, it's probably file system specific.
I saw it with xfs, but tmpfs indeed updates it synchronously.

bhalevy:
Draft until https://github.com/scylladb/seastar/pull/2598

src/core/reactor.cc
bhalevy: Done in v2

bhalevy:
In v2 ([d9b772b](https://github.com/scylladb/seastar/commit/d9b772bb77ef96433764605447ba277106e362dd)):
- rebased on top of #2598 for now
- use std::filesystem::space_info
- use coroutine in reactor::file_system_space
- simplify unit test

include/seastar/core/file-types.hh
bhalevy: gone in v2

include/seastar/core/file-types.hh
bhalevy: v2 is rebased onto #2598 for now

include/seastar/core/file-types.hh
```diff
@@ -158,6 +159,18 @@ inline constexpr file_permissions operator&(file_permissions a, file_permissions
     return file_permissions(std::underlying_type_t<file_permissions>(a) & std::underlying_type_t<file_permissions>(b));
 }
 
+/// seastar::space_info is equivalent to std::filesystem::space_info
+/// with renamed members, to prevent a conflict with future::available()
+struct space_info {
+    std::uintmax_t capacity;
+    std::uintmax_t free_space;
+    std::uintmax_t available_space;
```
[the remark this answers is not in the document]
bhalevy: gone in v2

bhalevy:
- rebased

PR #2598, opened 2024-12-23
future: avoid inheriting from future payload type

avikivity:
The uninitialized_wrapper_base will, when possible inherit from the future payload type T in order to apply the empty base optimization (if sizeof(T) == 0). However, this injects methods from T into the scope of uninitialized_wrapper and classes that inherit from it, such as future_state. If one of the names of the methods in future_state clashes with a member of T, we get a compilation failure.

Fix by avoiding the inheritance branch of uninitialized_wrapper_base and using the composition branch, and use [[no_unique_address]] to avoid bloating future<some_empty_class>.

It turns out that we delete the specialization that was used with internal::monostate (corresponding to future<>), and so we have to specialize uninitialized_wrapper for it.

I verified that sizeof(future<>) isn't changed by the patch.


bhalevy (approved):
LGTM. Thanks for fixing

PR #2588, opened 2024-12-17
Fix for missing cycles column and more

travisdowns:
In this series:

 - perf_tests_perf: minimal self-benches for PERF_TEST stuff
 - De-duplicate perf_tests main loop, which also fixes #2587 

Example results for `perf_test_perf` after the fix:

```
test                             iterations      median         mad         min         max      allocs       tasks        inst      cycles
perf_tests.test_simple_1          774019546     1.313ns     0.019ns     1.259ns     1.333ns       0.000       0.000         7.0         5.0
perf_tests.test_simple_n         3188188100     0.330ns     0.001ns     0.308ns     0.387ns       0.000       0.000         3.1         1.3
perf_tests.test_ready_async_1    3638567766     0.280ns     0.005ns     0.271ns     0.290ns       0.000       0.000         4.0         1.1
perf_tests.test_ready_async_n    3179866200     0.312ns     0.006ns     0.305ns     0.435ns       0.000       0.000         3.1         1.3
perf_tests.test_unready_async_1    28622990    34.253ns     0.303ns    33.950ns    36.198ns       1.000       2.000       360.1       125.3
perf_tests.test_unready_async_n  1082973200     0.942ns     0.011ns     0.897ns     1.031ns       0.020       0.030         9.3         3.5
fixture.test_fixture_1            781121567     1.303ns     0.006ns     1.282ns     1.309ns       0.000       0.000         7.0         5.0
fixture.test_fixture_n           3096749700     0.330ns     0.005ns     0.318ns     0.363ns       0.000       0.000         3.1         1.3
fixture.test_coro_1                71657240    13.640ns     0.179ns    13.461ns    14.588ns       1.000       0.000       150.1        49.2
fixture.test_coro_n              2214644700     0.475ns     0.019ns     0.453ns     0.722ns       0.010       0.000         4.6         1.9
```

Issue #2587, opened 2024-12-17
async PERF_TEST always returns 0 cycles


travisdowns:
If you write a `PERF_TEST` (or any other variation like `PERF_TEST_F/C` etc) whose testing method is async, the "cycles" column is always zero.

PR #2599, opened 2024-12-24
test: Add perf test to measure the "cost" of chain wakeup

xemul:
test: Add perf test to measure the "cost" of chain wakeup

The test creates a chain of future-promise-s and then wakes up the last one thus causing the cascade of resolutions. There are 2x2 tests -- resolve with value or exception vs co_await-ed or .then()-d chains.

The result is (depth of 32)
    
    test                   iterations      median         mad         min         max      allocs       tasks        inst      cycles
    chain.then_value         29944640    33.158ns     0.026ns    33.059ns    33.740ns       1.063       1.094       314.9         0.0
    chain.await_value        18963648    52.668ns     0.242ns    52.426ns    54.334ns       1.094       1.125       414.3         0.0
    chain.then_exception     29095808    33.633ns     0.093ns    33.541ns    35.308ns       1.094       1.094       316.3         0.0
    chain.await_exception      325344     3.112us    17.301ns     3.083us     3.192us       2.156       1.125     23489.0         0.0
    
Waking up co-await-ed chain with exceptions is extremely expensive.

The result with depth of 8 are the same, which means that exception propagation via co-awaits is expensive on _every_ co_await, as it "scales" linearly with the chain depth.

### DIFF
--- a/tests/perf/future_util_perf.cc
+++ b/tests/perf/future_util_perf.cc
@@ -55,6 +55,73 @@ future<> immediate(int v, int& vs)
     return make_ready_future<>();
 }
 
+struct chain {
+    static constexpr int scale = 32;
+    int value = 0;
+
+    [[gnu::noinline]]
+    future<> do_then(future<> w, int depth = scale) {
+        if (depth > 0) {
+            return do_then(std::move(w), depth - 1).then([this] {
+                value += 1;
+                perf_tests::do_not_optimize(value);
+                return make_ready_future<>();
+            });
+        } else {
+            return w;
+        }
+    }
+
+    [[gnu::noinline]]
+    future<> do_await(future<> w, int depth = scale) {
+        if (depth > 0) {
+            co_await do_await(std::move(w), depth - 1);
+            value += 1;
+            perf_tests::do_not_optimize(value);
+        } else {
+            co_await std::move(w);
+        }
+    }
+};
+
+PERF_TEST_F(chain, then_value)
+{
+    promise<> p;
+    auto f = do_then(p.get_future());
+    p.set_value();
+    return f.then([] { return scale; });
+}
+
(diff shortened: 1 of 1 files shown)


bhalevy:
@xemul thanks.  This is very important.
BTW, why is the cycles/op stat missing?

travisdowns:
> BTW, why is the cycles/op stat missing?

There are two related bugs: cycles may always be zero for some benchmarks (https://github.com/scylladb/seastar/issues/2587) and cycles column is missing in .md output (fix: https://github.com/scylladb/seastar/pull/2588).

The first one would be the cause here.

travisdowns:
> it's clear that exception propagation via co-awaits is expensive on _every_ co_await, as it "scales" linearly with the chain depth.

Makes sense I guess as every co_await in the chain involves a throw-catch and those are the expensive part (high fixed cost, and also a cost per frame unwound though I guess there are very few frames unwound for each throw/catch here).

xemul:
upd:
- fixed (I hope) CI compilation
- normalized results by scale to see the cost of a single continuation

xemul:
> > it's clear that exception propagation via co-awaits is expensive on _every_ co_await, as it "scales" linearly with the chain depth.
> 
> Makes sense I guess as every co_await in the chain involves a throw-catch and those are the expensive part (high fixed cost, and also a cost per frame unwound though I guess there are very few frames unwound for each throw/catch here).

That's exactly my guess too (Cc @regevran @tchaikov )

xemul:
@avikivity , please consider merging

regevran:
> That's exactly my guess too 

It is a possible explanation but a closer look at the implementation is required.
I will try to take a look when hopefully I have the leisure.

xemul:
@avikivity , please consider merging

PR #2601, opened 2024-12-26
Tunnel requested abort exception to semaphore wait caller

bhalevy:
Today, when calling `wait(abort_source&)` the caller always gets `semaphore_aborted` error when abort is requested.
However, with `abort_on_expiry`, the actual reason could be timeout expiry, which should be distinguishable from the abort error.

This series modifies `abort_on_expiry` first to request abort with a `timed_out_error` exception, and then `internal::abortable_fifo` and `semaphore` are changed to tunnel that exception to the caller.

Respective unit tests were added to test those cases.

The motivation for this change is https://github.com/scylladb/scylladb/pull/22017
which will use scylladb utils::composite_abort_source to compose an `abort_on_expiry` abort_source, used for point request timeout and a global abort_source used on shutdown.

### DIFF
--- a/tests/unit/semaphore_test.cc
+++ b/tests/unit/semaphore_test.cc
@@ -33,6 +33,8 @@
 #include <seastar/core/shared_mutex.hh>
 #include <ranges>
 
+#include "expected_exception.hh"
+
 using namespace seastar;
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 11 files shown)


bhalevy:
@scylladb/seastar-maint please review and consider merging

tests/unit/semaphore_test.cc
```diff
@@ -426,6 +427,19 @@ SEASTAR_THREAD_TEST_CASE(test_semaphore_abort_after_wait) {
...
+SEASTAR_THREAD_TEST_CASE(test_semaphore_abort_with_exception_after_wait) {
+    auto sem = semaphore(0);
+    abort_source as;
+    int x = 0;
+    auto fut1 = sem.wait(as).then([&x] {
+        x++;
+    });
+    as.request_abort_ex(std::runtime_error("expected"));
+    sem.signal();
+    BOOST_CHECK_THROW(fut1.get(), std::runtime_error);
```
xemul, line 439: How confident are we, that this thrown-and-caught `runtime_exception` is the one from above `as.request_abort_ex()`, but not some "generic" runtime error that accidentally crept in along the way?

bhalevy: I have no reason to suspect that, but I can change this to a unique exception type to increase the confidence.

bhalevy:
In v2 ([f8c36db](https://github.com/scylladb/seastar/commit/f8c36dbad7741e84262a09fb640c743e719787fd)):
- rebased
- tests: unit: refactor expected_exception
- used expected_exception in new semaphore unit test

tests/unit/semaphore_test.cc
xemul: Yes, please

bhalevy: done in v2

PR #2602, opened 2024-12-28
gitignore: ignore directories that match "build*"

tomershafir:
To ease local build versioning with git, like build_old and build_new.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 .cproject
 .project
 .settings
-build
+build*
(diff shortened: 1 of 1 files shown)


p12tic:
Shouldn't this be part of local configuration using `core.excludesFile` configuration key?

tomershafir:
I would prefer that seastar specifies that, so it "just works" without target dependency. Also, a global state can be trickier to manage if other projects use different conventions

PR #2604, opened 2025-01-02
Add missing include in dpdk_rte.hh

p12tic:
std::optional is used in the file. The build does not break at the moment because all source files that include this header also include <optional> some other way.


tchaikov (approved):
lgtm

tchaikov:

PR #2605, opened 2025-01-03
build: use path to libraries in .pc

tchaikov:
After removing the "MODULE" parameter from find_package(Boost) in af9b804253, the build started using Boost's native package config instead of cmake-boost. This caused seastar.pc to contain target references (e.g. Boost::program_options) rather than actual library paths, breaking builds of projects consuming seastar via pkg-config:

```
/usr/bin/clang++ ... \
  Boost::program_options Boost::thread /usr/lib64/libcares.so /usr/lib64/libfmt.so.11.0.2 -L/usr/lib64 -llz4 -Xlinker --dependency-file -Xlinker CMakeFiles/scylla.dir/Debug/link.d CMakeFiles/scylla_version.dir/Debug/release.cc.o CMakeFiles/scylla.dir/Debug/main.cc.o \
  ...
clang++: error: no such file or directory: 'Boost::program_options'
clang++: error: no such file or directory: 'Boost::thread'
```

In this change, use generator expressions to obtain the actual library paths for pkg-config files, ensuring correct linking for downstream projects.

This should address the regression introduced by af9b804253.

### DIFF
--- a/pkgconfig/seastar-testing.pc.in
+++ b/pkgconfig/seastar-testing.pc.in
@@ -9,7 +9,7 @@ Version: @PROJECT_VERSION@
 
 # Dependencies.
 boost_cflags=-I$<JOIN:@Boost_INCLUDE_DIRS@, -I>
-boost_unit_test_libs=@Boost_UNIT_TEST_FRAMEWORK_LIBRARY@
+boost_unit_test_libs=$<TARGET_PROPERTY:Boost::unit_test_framework,LOCATION>
 
 # Us.
 seastar_testing_cflags=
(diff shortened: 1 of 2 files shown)


tchaikov:
tested by compiling scylladb main HEAD with seastar after applying this change, without this change, the build failed.

tchaikov:

tchaikov:
tested by compiling scylladb and combined_tests from main HEAD with seastar after applying this change, without this change, the build failed.

tchaikov:
@scylladb/seastar-maint hello maintainers, could you help review this change? it addresses a regression, so it would be ideal if we can land it sooner.

PR #2608, opened 2025-01-06
smp: add a function that barriers memory prefault work

tomershafir:
Currently, memory prefault logic is internal and seastar doesnt provide much control to users. In order to improve the situation, I suggest to provide a barrier for the prefault threads. This allows to:

* Prefer predictable low latency and high throughput from the start of request serving, at the cost of a startup delay, depending on machine characteristics and application specific requirements. For example, a fixed capacity on prem db setup, where slower startup can be tolerated. From users perspective, they generally cannot tolerate inconsistency (like spikes in latency).
* Similarly, improve user scheduling decisions, like running less critical tasks while prefault works.
* Reliably test the prefault logic, improving reliability and users trust in seastar.

I tested locally. If you approve, next I will try to submit a prefault test.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4637,6 +4637,10 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     }
 }
 
+void barrier_memory_prefault() {
+    engine()._smp->join_memory_prefault();
+}
+
 bool smp::poll_queues() {
     size_t got = 0;
     for (unsigned i = 0; i < count; i++) {
--- a/src/core/smp.cc
+++ b/src/core/smp.cc
@@ -179,6 +179,13 @@ smp::setup_prefaulter(const seastar::resource::resources& res, seastar::memory::
 #endif
 }
 
+void
+smp::join_memory_prefault() {
+    if (_prefaulter) {
+        _prefaulter->join_all();
+    }
+}
+
 static
 std::optional<size_t>
 get_huge_page_size() {
@@ -224,11 +231,20 @@ internal::memory_prefaulter::memory_prefaulter(const resource::resources& res, m
     }
 }
 
+void
+internal::memory_prefaulter::join_all() {
+    const std::lock_guard<std::mutex> lock(join_mtx);
+    if (!joined) {
+        for (auto& t : _worker_threads) {
+            t.join();
+        }
+        joined = true;
+    }
+}
+
 internal::memory_prefaulter::~memory_prefaulter() {
     _stop_request.store(true, std::memory_order_relaxed);
-    for (auto& t : _worker_threads) {
-        t.join();
-    }
+    join_all();
 }
 
 void
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -374,14 +374,14 @@ private:
         std::atomic<uint64_t> _pending_signals;
         std::unordered_map<int, signal_handler> _signal_handlers;
     };
-
     signals _signals;
     std::unique_ptr<thread_pool> _thread_pool;
+
     friend class thread_pool;
     friend class thread_context;
     friend class internal::cpu_stall_detector;
-
     friend void handle_signal(int signo, noncopyable_function<void ()>&& handler, bool once);
+    friend void barrier_memory_prefault();
 
     uint64_t pending_task_count() const;
     void run_tasks(task_queue& tq);
--- a/include/seastar/core/smp.hh
+++ b/include/seastar/core/smp.hh
@@ -335,6 +335,7 @@ public:
     void cleanup_cpu();
     void arrive_at_event_loop_end();
     void join_all();
+    void join_memory_prefault();
     static bool main_thread() { return std::this_thread::get_id() == _tmain; }
 
     /// Runs a function on a remote core.
@@ -483,6 +484,13 @@ private:
 public:
     static unsigned count;
 };
+
+/// Barriers memory prefault background work. 
+/// This function returns after all memory prefault work is finished. 
+/// It is safe to call from multiple threads.
+/// 
+/// Note: This is a blocking operation, so it should be used with caution.
+void barrier_memory_prefault();
 
 SEASTAR_MODULE_EXPORT_END
 
--- a/src/core/prefault.hh
+++ b/src/core/prefault.hh
@@ -36,11 +36,16 @@ namespace seastar::internal {
 class memory_prefaulter {
     std::atomic<bool> _stop_request = false;
     std::vector<posix_thread> _worker_threads;
+
+    std::mutex join_mtx;
+    bool joined = false;
+    
     // Keep this in object scope to avoid allocating in worker thread
     std::unordered_map<unsigned, std::vector<memory::internal::memory_range>> _layout_by_node_id;
 public:
     explicit memory_prefaulter(const resource::resources& res, memory::internal::numa_layout layout);
     ~memory_prefaulter();
+    void join_all();
 private:
     void work(std::vector<memory::internal::memory_range>& ranges, size_t page_size, std::optional<size_t> huge_page_size_opt);
 };


avikivity:
Did you observe latency impact from the prefault threads? It was written carefully not to have latency impact, but it's of course possible that some workloads suffer.

tomershafir:
As you described in https://github.com/scylladb/seastar/pull/1702, page faults can cause deviation, and following up the example, there can be 25sec where latency is variably higher.

avikivity:
> As you described in #1702, page faults can cause deviation, and following up the example, there can be 25sec where latency is variably higher.

I said nothing about latency being higher there.

We typically run large machines with a few vcpus not assigned to any shards, and the prefault threads run with low priority.

tomershafir:
There are 2 aspects:

1. Page faults

In the previous comment, I meant page fault latency. The page faults can cause high latency unpredictably until the prefaulter finishes.

Regarding page faults measurement, it seems I cannot reliably measure on my env.

2. Prefault threads competition

I tried to non scientifically isolate wall time overhead of prefault threads:

I have a test app that performs file I/O and process memory buffers repeatedly. I used Ubuntu Orbstack VM with 1 NUMA node, 10 cores, --memory=14G, and a small input to let the overhead be most visible.

- With --lock-memory=1 without waiting, I see that the chrono time of the actual work is significantly higher than with --lock-memory=0. (~1800ms > ~600ms)
- When waiting before doing actual work, I see that the overhead is removed. 
- When building seastar without prefault code and --lock-memory=1 I dont see the overhead.

avikivity:
> By default seastar uses all vcpus, which makes sense for resource efficiency.
> 
> Also, do you free specific vcpus? Like one per numa node, the granularity of prefault threads.


1 in 8, with NUMA awareness. They're allocated for kernel network processing. See perftune.py.

tomershafir:
Nice. Let me know if this change makes sense to you

tomershafir:
@avikivity ping

tomershafir:
I also tried to simulate perftune with 1 free vcpu: `--cpuset=0-8` given the above setup, and I still observe the overhead, even though its less (~1600ms).

avikivity:
I don't understand what this 1600ms overhead is.

tomershafir:
> I tried to non scientifically isolate wall time overhead of prefault threads:
> 
> I have a test app that performs file I/O and process memory buffers repeatedly. I used Ubuntu Orbstack VM with 1 NUMA node, 10 cores, --memory=14G - effectively a small NUMA node, and a small input to let the overhead be most visible.
> 
> * With --lock-memory=1 without waiting, I see that the chrono time of the actual work is significantly higher than with --lock-memory=0. (~1800ms > ~600ms)
> * When waiting before doing actual work, I see that the overhead is removed.
> * When building seastar without prefault code and --lock-memory=1 I dont see the overhead.
I mean its the wall time of the work that I observe, with 1 free vcpu: `--cpuset=0-8` given the above setup.

tomershafir:
The problem is that it is slower and not consistent/predictable. After memory initialization it is faster and consistent.

Regarding the implementation, the problem is that `pthread_join` blocks, so in the current implementation can't using a future be misleading?

avikivity:
How is pthread_join relevant?

tomershafir:
Currently, the logical barrier waits for pthread_join on all the threads that perform the prefault work.

avikivity:
Ah, you're referring to the patch while I was referring to the current state. Don't use join then, instead figure out something else that can satisfy a seastar::promise. Maybe it's as simple as `seastar::alien::submit_to(0, [&] { _reactor._prefault_complete.set_value(); })`.

avikivity:
I don't completely see that it's useful but can't deny that it might be.

I'd be happier with an example of a real application requiring it.

avikivity:
> I have only a test application. How about scylladb?

I'm not aware of reports of problem during the prefault stage. It takes some time for a node to join the cluster, and by that time enough memory was prefaulted for it to work well.

tomershafir:
sorry for the delay. I see you implicitly fixed that too in https://github.com/scylladb/seastar/pull/2679. Closing

tomershafir:
sorry for the delay and the confusion. I saw you added a join method: https://github.com/scylladb/seastar/pull/2679 Ill try to rebase my changes

tomershafir:
v2:

* Rebase master and rewrite a non-blocking seastar::join_memory_prefault

tomershafir:
v2:

Rebase master and rewrite a non-blocking seastar::join_memory_prefault

tomershafir:
@avikivity hey, pls review

src/core/reactor.cc
```diff
@@ -4288,6 +4288,20 @@ unsigned smp::adjust_max_networking_aio_io_control_blocks(unsigned network_iocbs
     return network_iocbs;
 }
 
+future<> join_memory_prefault() {
+    auto& r = engine();
+    if (!r._smp->memory_prefault_initialized()) {
+        seastar_logger.warn("Memory prefaulter is not initialized but joined");
```
avikivity, line 4294: This warning isn't helpful to users; what can they do?

src/core/smp.cc
```diff
@@ -176,10 +177,15 @@ void
...
 #ifndef SEASTAR_THREAD_STACK_GUARDS
-    _prefaulter = std::make_unique<internal::memory_prefaulter>(_alien, res, std::move(layout));
+    _prefaulter = std::make_unique<internal::memory_prefaulter>(*this, res, std::move(layout));
 #endif
 }
 
+bool
+smp::memory_prefault_initialized() {
+    return _prefaulter != nullptr;
+}
```
avikivity, line 187: Instead of this, you can promise::set_value() on the promise if you don't initialize the prefaulter.

src/core/smp.cc
```diff
@@ -238,6 +244,14 @@ internal::memory_prefaulter::join_threads() noexcept {
     _layout_by_node_id.clear();
 }
 
+void 
+internal::memory_prefaulter::alien_on_complete(smp& smp_context) {
+    run_on(smp_context._alien, 0, [this, &smp_context] () noexcept {
+        join_threads();
+        run_in_background(smp_context.broadcast_memory_prefault_completion());
```
avikivity, line 251: Alternatively, we can document that that the join() should only be run on shard 0. I expect that most applications run initialization code in shard 0 and don't need it to be available anywhere else.

(I want to deprecate and remove run_in_background, I think it's dangerous)

src/core/smp.cc
tomershafir: I agree, this way the promise is not left unresolved. Will call it on smp::configure.

src/core/reactor.cc
avikivity: The application doesn't know if the user wants to prefault memory or not. It's common in local testing not to prefault, and in production to prefault.

src/core/smp.cc
tomershafir: Why not let it work on any shard? We anyway need the new promise member on the reactor class.
 
Also, maybe make run_in_background internal instead? its needed for such use cases.

tomershafir: wdyt?

tomershafir: @avikivity ping

tomershafir:
@avikivity v3:

- Allow `seastar::join_memory_prefault` only on shard 0. Keep `run_in_background()` which is needed.
- Remove redundant warning.
- If memory prefault is not initialized, instead of checking inside `seastar::join_memory_prefault`, submit dummy completion immediately in `smp::configure`.

tomershafir:
@avikivity v3:

- Allow `seastar::join_memory_prefault` only on shard 0, enforce via assertion and document. Keep `run_in_background()` which is needed.
- Remove redundant warning.
- If memory prefault is not initialized, instead of checking inside `seastar::join_memory_prefault`, submit dummy completion immediately in `smp::configure`.

PR #2610, opened 2025-01-07
net: replace deprecated ip::address_v4::from_string()

cbodley:
`boost::asio::ip::address_v4::from_string()` was removed in boost 1.87, leading to compilation failures:
```
src/seastar/src/net/ip.cc:44:46: error: no member named 'from_string' in 'boost::asio::ip::address_v4'
    auto ipv4 = boost::asio::ip::address_v4::from_string(addr, ec);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
```
prior to 1.87, the function had been [documented as deprecated](https://www.boost.org/doc/libs/1_86_0/doc/html/boost_asio/reference/ip__address_v4/from_string.html):

> (Deprecated: Use make_address_v4().)

### DIFF
--- a/src/net/ip.cc
+++ b/src/net/ip.cc
@@ -41,7 +41,7 @@ namespace net {
 
 ipv4_address::ipv4_address(const std::string& addr) {
     boost::system::error_code ec;
-    auto ipv4 = boost::asio::ip::address_v4::from_string(addr, ec);
+    auto ipv4 = boost::asio::ip::make_address_v4(addr, ec);
     if (ec) {
         throw std::runtime_error(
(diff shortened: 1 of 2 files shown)


tchaikov (approved):
the minimal required boost version in Seastar is 1.73, which contains `make_address_v4()`, so this change looks good to me.

see also https://www.boost.org/doc/libs/1_73_0/doc/html/boost_asio/reference/ip__address_v4/from_string.html

tchaikov:
@scylladb/seastar-maint hello maintainers, could you help review this change? without which, seastar is not compatible with boost 1.87.

Issue #2609, opened 2025-01-06
scripts/seastar-addr2line hangs on I/O


vladzcloudius:
HEAD: ff24926fe4374995abcfb606ef880e146eb40647
OS: Ubuntu 24.04
binutils version: `2.42-4ubuntu2.3`
LLVM version: `18.0-59~exp2`
Python version: `3.12.3-0ubuntu2`

## Description
The most trivial invocation of `seastar-addr2line` hangs on I/O (using `addr2line` from `binutils` gives the same result):

```
$ echo "0x0000000002eebcb2" | strace /opt/scylladb/seastar/scripts/seastar-addr2line -b 3 -e /var/lib/binaries/2021.1.20-0.20230604.0f9b75eaf1.amd64/deb-relocatable/scylla.debug --addr2line llvm-addr2line
write(4, "0x0000000002eebcb2\n0x0\n", 23) = 23
read(5, "0x2eebcb2: void seastar::backtra"..., 8192) = 257
read(5, "0x0: auto seastar::internal::do_"..., 8192) = 281
read(5, 
<here the program hangs forever>
```

Interrupting it prints the following:

```
main()
  File "/opt/scylladb/seastar/scripts/seastar-addr2line", line 503, in main
    with BacktraceResolver(
  File "/opt/scylladb/seastar/scripts/addr2line.py", line 387, in __exit__
    self._print_current_backtrace()
  File "/opt/scylladb/seastar/scripts/addr2line.py", line 452, in _print_current_backtrace
    self._print_resolved_address(module, addr)
  File "/opt/scylladb/seastar/scripts/addr2line.py", line 403, in _print_resolved_address
    sys.stdout.write(self.resolve_address(address, module))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/scylladb/seastar/scripts/addr2line.py", line 397, in resolve_address
    resolved_address = self._get_resolver_for_module(module)(address)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/scylladb/seastar/scripts/addr2line.py", line 140, in __call__
    return self._read_resolved_address()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/scylladb/seastar/scripts/addr2line.py", line 123, in _read_resolved_address
    line = self._output.readline()
[...]

PR #2611, opened 2025-01-08
scripts/addr2line.py: fix hanging with the new llvm-addr2line version

vladzcloudius:
addr2line.py invokes "addr2line" tool in a "server" mode where addresses are fed via stdin. It intentionally makes a corresponding "addr2line" tool generate a known "invalid address" pattern at the end of each address decoding by feeding it a "0x0" address in order to denote the end of it because a single address can be decoded into multiple lines due to inlining.

The above strategy assumes that 0x0 address is always going to be interpreted as an invalid one by both add2line and by llvm-addr2line.

However, at least llvm-addr2line 18.1.3 doesn't always interpret it this way. On the other hand an empty line does always generate an expected (invalid address) output.

On top of that it looks like a new llvm-addr2line changed the "invalid address" pattern format.

This patch adds a new "invalid input" pattern to a dummy_pattern and changes the way we generate such an output by pushing an empty line instead of a 0x0 address.

Ref #2609

### DIFF
--- a/scripts/addr2line.py
+++ b/scripts/addr2line.py
@@ -54,7 +54,9 @@ class Addr2Line:
     dummy_pattern = re.compile(
         r"(.*0x0000000000000000: \?\? \?\?:0\n)"  # addr2line pattern
         r"|"
-        r"(.*0x0: \?\? at .*\n)"  # llvm-addr2line pattern
+        r"(\?\? at \?\?:0\n)"  # llvm-addr2line pattern for LLVM 18 and newer
+        r"|"
+        r"(,\n)"  # llvm-addr2line pattern for LLVM 17 and older
     )
 
     def __init__(
@@ -97,7 +99,7 @@ class Addr2Line:
         # will just exit.  We need to be robust against that.  We
         # can't just wait on self._addr2line since there is no
         # guarantee on what timeout is sufficient.
-        self._input.write('\n')
+        self._input.write(',\n')
         self._input.flush()
         res = self._output.readline()
         self._missing = res == ''
@@ -129,9 +131,9 @@ class Addr2Line:
     def __call__(self, address: str):
         if self._missing:
             return " ".join([self._binary, address, '\n'])
-        # We print a dummy 0x0 address after the address we are interested in
+        # We trigger a dummy "invalid" address printout after the address we are interested in
         # which we can look for in _read_address
-        inputline = address + '\n0x0\n'
+        inputline = address + '\n,\n'
         self._parent.debug('Add2Line sending input to stdin:', inputline)
         self._input.write(inputline)
         self._input.flush()


scripts/addr2line.py
```diff
@@ -131,7 +133,7 @@ def __call__(self, address: str):
             return " ".join([self._binary, address, '\n'])
         # We print a dummy 0x0 address after the address we are interested in
```
michoecho, line 134: This comment is outdated after the patch.

scripts/addr2line.py
```diff
@@ -131,7 +133,7 @@ def __call__(self, address: str):
...
         # which we can look for in _read_address
-        inputline = address + '\n0x0\n'
+        inputline = address + '\n\n'
```
michoecho, line 136: Sending an empty line to llvm-addr2line results in an ugly `llvm-addr2line: error: no module offset has been specified` being printed to stderr.

How about this?
```suggestion
        inputline = address + '\n,\n'
```

(By the way, that's what `perf` does. It uses a comma as the sentinel. See https://github.com/torvalds/linux/blob/eea6e4b4dfb8859446177c32961c96726d0117be/tools/perf/util/srcline.c#L729-L738.)

scripts/addr2line.py
```diff
@@ -54,7 +54,9 @@ class Addr2Line:
     dummy_pattern = re.compile(
         r"(.*0x0000000000000000: \?\? \?\?:0\n)"  # addr2line pattern
         r"|"
-        r"(.*0x0: \?\? at .*\n)"  # llvm-addr2line pattern
+        r"(.*0x0: \?\? at .*\n)"  # llvm-addr2line patterns
+        r"|"
+        r"(\?\? at \?\?:0\n)"
```
michoecho, line 59: If we are no longer using 0x0 as the sentinel, there's no point leaving this part of the pattern in the code, is there?

```suggestion
        r"(\?\? at \?\?:0\n)"
```

scripts/addr2line.py
vladzcloudius: fixed

scripts/addr2line.py
vladzcloudius: > Sending an empty line to llvm-addr2line results in an ugly llvm-addr2line: error: no module offset has been specified being printed to stderr. (At least with llvm-18)

True. But since it's printed to `stderr` it's easily "filtarable".

> How about this?
> ...
This trick works and I can adjust the PR to use it (there is one more place where I'd need to add a coma: line 102). 
It's just this is another "black magic" undocumented thing that may break in some next release of llvm/binutils.

vladzcloudius: Trick is applied

scripts/addr2line.py
vladzcloudius: I didn't check what's the output of older llvm-addr2line versions.
In the binutils'es `addr2line` the output for 0x0, "\n" and ",\n" is the same.

So, I assumed that the same probably holds for older versions of `llvm-add2line` and for them the pattern is as at the line 57 above.

I didn't want to break things. ;)

vladzcloudius: Hmmm...
For `llvm-addr2line-10` the "coma trick" doesn't work.
Here is the output of `llvm-addr2line-10`:

```
$ echo 0x0 | llvm-addr2line-10 -Cfpia -e /usr/bin/bash
0x0: ?? at ??:0
$ echo "," | llvm-addr2line-10 -Cfpia -e /usr/bin/bash
,
$ echo "" | llvm-addr2line-10 -Cfpia -e /usr/bin/bash

$
```

vladzcloudius: llvm-10 comes with U20.04

vladzcloudius: U22.04 comes with llvm-14:

```
$ echo 0x0 | llvm-addr2line-14 -e /usr/bin/bash
??:0
$ echo "," | llvm-addr2line-14 -e /usr/bin/bash
,
$ echo "" | llvm-addr2line-14 -e /usr/bin/bash

$
```

michoecho: Yes, LLVM <= 17 just echoes invalid lines. If you want to support older versions of LLVM, I guess you can just an extra `r"(,\n)"` pattern to `dummy_pattern`.

It seems that LLVM >= 18 adjusted its behaviour for invalid lines to match binutils's format.

michoecho: Validation:
```
~ $ for ver in (seq 12 19);
      nix shell nixpkgs#llvmPackages_$ver.llvm --command bash -c 'llvm-addr2line --version | grep version;
      llvm-addr2line -Cipfe /bin/sh <<< ,';
      end
  LLVM version 12.0.1
,
  LLVM version 13.0.1
,
  LLVM version 14.0.6
,
  LLVM version 15.0.7
,
  LLVM version 16.0.6
,
  LLVM version 17.0.6
,
  LLVM version 18.1.8
?? at ??:0
  LLVM version 19.1.6
?? at ??:0
```

vladzcloudius: Conditioning on `llvm-addr2line` version is going to uglify the code:
* The most platform-independent way to do it is to go via `llvm-addr2line --version` and then parse the output.
* But its output seems to be not uniform:

```
$ llvm-addr2line-14 --version
llvm-addr2line
Ubuntu LLVM version 14.0.0
  
  Optimized build.
  Default target: x86_64-pc-linux-gnu
  Host CPU: znver3
$

$ llvm-addr2line-18 --version
llvm-addr2line
Ubuntu LLVM version 18.1.3
  Optimized build.
$

$ llvm-addr2line-10 --version
LLVM (http://llvm.org/):
  LLVM version 10.0.0
  
  Optimized build.
  Default target: x86_64-pc-linux-gnu
  Host CPU: cascadelake
$
```

It looks like the version always comes in a second line after a word "version".
But assuming this feels too hackish.
But if that's what you want me to do - I'll do it ;)

michoecho: Wouldn't this work?:

```suggestion
        r"(,\n)"  # llvm-addr2line patterns
        r"|"
        r"(\?\? at \?\?:0\n)"
```

I think this pattern should work as we want with both the "old LLVM style" and the "new LLVM style".

And I share your pain that it's all so hacky. But it seems that the only non-hacky alternative is to entirely replace addr2line with something else. (And that might end up more painful than the hacks).

travisdowns: I have had to modify/fix this many times, and have added the `-v` argument to help debug these cases, so you can see when the parsing stalls out due to the failure to match the dummy line. I have a similar fix to this one in a branch somewhere because I ran into the same issue (can't recall exactly why I did not upstream it).

I think we should just make the regex more accepting rather than try to match the dummy line in a addr2line version-dependent way. The chance of a false positive is low.

I'm also wondering if we should just pass in a random 64-bit address on the kernel size of the address space (MSB set). If we see this in the output we know it is sentinel: such an address will never appear in a valid userspace stack trace (and, practically never on the kernel side either given the size of 64-bit space).

vladzcloudius: > I think this pattern should work as we want with both the "old LLVM style" and the "new LLVM style".

That's actually a (brilliantly simple) good idea!

vladzcloudius: done

vladzcloudius: Wait! You wanted to also remove the no-longer-needed patters.

vladzcloudius: Now it's proper-done.

vladzcloudius: @travisdowns your idea seems to be overly complicated compared to ","-line sentinel one.

travisdowns: > @travisdowns your idea seems to be overly complicated compared to ","-line sentinel one.

Sure, whatever is simplest: I just really want to avoid using tool-version specific regex (e.g., parsing `--version`). If you've found a trick all current binutils and llvm addr2line versions that's awesome!

michoecho (approved):

michoecho:
@scylladb/seastar-maint Pinging for merge.

PR #2613, opened 2025-01-09
coroutine: ADL

Comrade88:
every user of the coroutine::exception helpers outside of seastar
namespace has to spell explicitly 'using namespace seastar' statement
to compile their code.
let's move the co_await operator into the exception type namespace to
take advantage of the ADL (argument-dependent lookup) language feature
to give the compiler a hint about the scan scope of the co_await operator.

### DIFF
--- a/tests/unit/coroutines_test.cc
+++ b/tests/unit/coroutines_test.cc
@@ -39,7 +39,25 @@
 #include <seastar/testing/test_case.hh>
 #include <seastar/util/later.hh>
 
-using namespace seastar;
+using seastar::broken_promise;
+using seastar::circular_buffer;
+using seastar::create_scheduling_group;
+using seastar::current_scheduling_group;
+using seastar::default_scheduling_group;
+using seastar::future;
+using seastar::make_exception_future;
+using seastar::make_ready_future;
+using seastar::need_preempt;
+using seastar::promise;
+using seastar::scheduling_group;
+using seastar::semaphore;
+using seastar::semaphore_timed_out;
+using seastar::sleep;
+using seastar::yield;
+
+namespace coroutine = seastar::coroutine;
+namespace testing = seastar::testing;
+
 using namespace std::chrono_literals;
 
 namespace {
--- a/include/seastar/coroutine/exception.hh
+++ b/include/seastar/coroutine/exception.hh
@@ -152,10 +152,9 @@ exception return_exception(T&& t) noexcept {
     return exception(std::make_exception_ptr(std::forward<T>(t)));
 }
 
-} // coroutine
-
-inline auto operator co_await(coroutine::exception ex) noexcept {
+inline auto operator co_await(exception ex) noexcept {
     return internal::exception_awaiter(std::move(ex.eptr));
 }
 
+} // coroutine
 } // seastar


tchaikov:
@Comrade88 thank you for the fix. but would be better if we can have

1. a better reproducer to explain this issue fixed by this change. i am wondering if you could apply following patch:
   ```diff
   diff --git a/tests/unit/coroutines_test.cc b/tests/unit/coroutines_test.cc
   index bf989f46..41605278 100644
   --- a/tests/unit/coroutines_test.cc
   +++ b/tests/unit/coroutines_test.cc
   @@ -39,9 +39,27 @@
    #include <seastar/testing/test_case.hh>
    #include <seastar/util/later.hh>
    
   -using namespace seastar;
    using namespace std::chrono_literals;
    
   +using seastar::broken_promise;
   +using seastar::circular_buffer;
   +using seastar::create_scheduling_group;
   +using seastar::current_scheduling_group;
   +using seastar::default_scheduling_group;
   +using seastar::future;
   +using seastar::make_exception_future;
   +using seastar::make_ready_future;
   +using seastar::need_preempt;
   +using seastar::promise;
   +using seastar::scheduling_group;
   +using seastar::semaphore;
   +using seastar::semaphore_timed_out;
   +using seastar::sleep;
   +using seastar::yield;
   +
   +namespace coroutine = seastar::coroutine;
   +namespace testing = seastar::testing;
   +
    namespace {
    
    future<int> old_fashioned_continuations() {
   ```
2. amend the commit message to

   1. explain the symptom of the issue
   2. its impact
   3. the change we have in the fix
   4. and how it addresses the problem
   5. the potential impact of this change.

what do you think?

Comrade88:
@tchaikov , i've updated the patch description and applied the unit test reproducer

tchaikov:
> @tchaikov , i've updated the patch description and applied the unit test reproducer

hi @Comrade88 , Thank you for adding the reproducer and updating the commit message. The changes have significantly improved the patch. However, I have one remaining suggestion regarding the commit title.
The current title "coroutine: ADL" is too terse and doesn't convey the purpose or scope of the change. I recommend making it more descriptive:

> coroutine: move operator co_await(exception) into seastar::coroutine namespace

IMHO, this revised title clearly indicates both the component being modified and the specific change being made, making it easier for others to understand the commit's purpose at a glance.

what do you think?

Comrade88:
hi @tchaikov. i think u version doesn't explain the intent of the change. better
> use ADL to improve user compilation experience

tchaikov:
IMHO, in general, the title should only puts the exact change. but the intention and the rationales behind the change should be explained in the body of commit message.

tchaikov (approved):
lgtm

tchaikov:

tchaikov:

tests/unit/coroutines_test.cc
```diff
@@ -39,7 +39,25 @@
...
+using seastar::make_exception_future;
+using seastar::make_ready_future;
+using seastar::need_preempt;
+using seastar::promise;
+using seastar::scheduling_group;
+using seastar::semaphore;
+using seastar::semaphore_timed_out;
+using seastar::sleep;
+using seastar::yield;
+
```
avikivity, line 57: I don't understand the changes here.

avikivity:
Looks sensible, but I don't understand the changes to the test.

avikivity:
It's a little messy, but okay. I guess it would be better to write the tests without any `using` directives at all.

Issue #2615, opened 2025-01-10
Cross-shard sharing of IO capacity is effectively IOPS-based, not tokens-based


xemul:
When disk is saturated, shards end up queuing one request at a time to shared token bucket via `pending` facility. As a result shards start to share the tokens in perquest-by-request manner

refs: #2596

xemul:
Fixed by #2616

PR #2616, opened 2025-01-14
fair_queue: make the fair_group token grabbing discipline more fair

michoecho:
The current design of `fair_group` isn't fair enough to shards.
During contention, the group will be -- aproximately -- taking requests
from shards one-by-one, in round robin.

This guarantees that each contender will dispatch an equal *number* of
requests. This is some kind of fairness, but it's not the kind we want,
probably ever.

A better kind of fairness is that under contention each shard should
be guaranteed `1/nr_shards` of the disk's IOPS and/or `1/nr_shards` of
byte-bandwidth, whichever dimension it pressures more.
This is needed so that each shard can be relied on to sustain a certain
rate of requests -- the lower bound of the slowest shard's throughput
usually dictates the throughput of the entire cluster.

But those two kinds of fairness are only the same if all IO requests have
the same size and direction. Otherwise they can be drastically different.

With the current design it's easy to create a situation where a shard receives
an arbitrarily small fraction of both IOPS and bandwidth, despite being
IO-bound. (Example: a node with X shards, where one shard spams only very small
requests and other shards spams only big requests).

This is a problem in practice. In ScyllaDB, we observed IO starvation of
some shards during realistic workloads. While they require some workload
asymmetry to occur, even small asymmetries can cause serious unfairness to occur.
(For example, a shard which receives 6% more of database queries than other
shards can be starved to less than 50% of its fair share of IOPS and/or
bandwidth -- because each of those surplus 1 kiB queries is "fairly" matched with
16x costlier 128 kiB low-priority batch IO requests on other shards).

To improve this, `fair_group` needs a different queueing discipline.
There are many possible ways, but this patch chooses the one which is relatively
the most similar to the current one. The main idea is that we still rely on the
"approximate round robin" of the token bucket as the basis for fairness, but we reserve
a fixed-size batch of tokens at a time, rather than a fixed-size (i.e. 1) batch
of _requests_ at a time. This turns the discipline from request-fair to
token-fair, which is what we want.

The implementation details are non-trivial, though, and should be carefully reviewed.

This change doesn't result in full fairness. The degree of fairness depends on how often
a given shard polls. A continuously-polling shard should get its fair share of disk bandwidth,
but a rarely-polling shard might only get a fraction of its fair share. But with current default values
for IO scheduler parameters, a shard should be guaranteed at least 60% of its fair share
in the worst case, which is decent, and better than the current "arbitrarily-low%".

Also note that by "fair share" we mean 1 / `nr_shards_in_group`, not 1/`nr_actively_contending_shards_in_group`. If there are 30 shards in a group but only 2 are contending for IO — one polling rarely and one polling continuously — the rarely-polling contender will only get ~1/30 of the disk's bandwidth, not ~1/2. This is not a problem for applications where the interactive workloads are mostly symmetric (because then it's the "normal" case that all shards are contending and none can expect more than 1/30), but it's unsatisfying.

Different scheduling algorithms could improve on both of the above. We have some ideas for that. But they would be a bigger departure from status quo, and they should be harder to implement well.

### DIFF
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -350,6 +350,8 @@ private:
     };
 
     std::optional<pending> _pending;
+    // Total capacity of all requests waiting in the queue.
+    capacity_t _queued_capacity = 0;
 
     void push_priority_class(priority_class_data& pc) noexcept;
     void push_priority_class_from_idle(priority_class_data& pc) noexcept;
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -254,6 +254,7 @@ public:
     capacity_t per_tick_grab_threshold() const noexcept { return _per_tick_threshold; }
     capacity_t grab_capacity(capacity_t cap) noexcept;
     clock_type::time_point replenished_ts() const noexcept { return _token_bucket.replenished_ts(); }
+    void refund_tokens(capacity_t) noexcept;
     void replenish_capacity(clock_type::time_point now) noexcept;
     void maybe_replenish_capacity(clock_type::time_point& local_ts) noexcept;
 
@@ -331,25 +332,25 @@ private:
     size_t _nr_classes = 0;
     capacity_t _last_accumulated = 0;
 
-    /*
-     * When the shared capacity os over the local queue delays
-     * further dispatching untill better times
-     *
-     * \head  -- the value group head rover is expected to cross
-     * \cap   -- the capacity that's accounted on the group
-     *
-     * The last field is needed to "rearm" the wait in case
-     * queue decides that it wants to dispatch another capacity
-     * in the middle of the waiting
-     */
+    // _pending represents a reservation of tokens from the bucket.
+    //
+    // In the "dispatch timeline" defined by the growing bucket head of the group,
+    // tokens in the range [_pending.head - cap, _pending.head) belong
+    // to this queue.
+    //
+    // For example, if:
+    //    _group._token_bucket.head == 300
+    //    _pending.head == 700
+    //    _pending.cap == 500
+    // then the reservation is [200, 700), 100 tokens are ready to be dispatched by this queue,
+    // and another 400 tokens are going to be appear soon. (And after that, this queue
+    // will be able to make its next reservation).
     struct pending {
-        capacity_t head;
-        capacity_t cap;
-
-        pending(capacity_t t, capacity_t c) noexcept : head(t), cap(c) {}
+        capacity_t head = 0;
+        capacity_t cap = 0;
     };
+    pending _pending;
 
-    std::optional<pending> _pending;
     // Total capacity of all requests waiting in the queue.
     capacity_t _queued_capacity = 0;
 
@@ -359,9 +360,20 @@ private:
     void plug_priority_class(priority_class_data& pc) noexcept;
     void unplug_priority_class(priority_class_data& pc) noexcept;
 
-    enum class grab_result { grabbed, cant_preempt, pending };
-    grab_result grab_capacity(const fair_queue_entry& ent) noexcept;
-    grab_result grab_pending_capacity(const fair_queue_entry& ent) noexcept;
+    // Replaces _pending with a new reservation starting at the current
+    // group bucket tail.
+    void grab_capacity(capacity_t cap) noexcept;
+    // Shaves off the fulfilled frontal part from `_pending` (if any),
+    // and returns the fulfilled tokens in `ready_tokens`.
+    // Sets `our_turn_has_come` to the truth value of "`_pending` is empty or
+    // there are no unfulfilled reservations (from other shards) earlier than `_pending`".
+    //
+    // Assumes that `_group.maybe_replenish_capacity()` was called recently.
+    struct reap_result {
+        capacity_t ready_tokens;
+        bool our_turn_has_come;
+    };
+    reap_result reap_pending_capacity() noexcept;
 public:
     /// Constructs a fair queue with configuration parameters \c cfg.
     ///
--- a/src/core/fair_queue.cc
+++ b/src/core/fair_queue.cc
@@ -297,12 +297,14 @@ void fair_queue::queue(class_id id, fair_queue_entry& ent) noexcept {
         push_priority_class_from_idle(pc);
     }
     pc._queue.push_back(ent);
+    _queued_capacity += ent.capacity();
 }
 
 void fair_queue::notify_request_finished(fair_queue_entry::capacity_t cap) noexcept {
 }
 
 void fair_queue::notify_request_cancelled(fair_queue_entry& ent) noexcept {
+    _queued_capacity -= ent._capacity;
     ent._capacity = 0;
 }
 
@@ -375,6 +377,7 @@ void fair_queue::dispatch_requests(std::function<void(fair_queue_entry&)> cb) {
         h._accumulated += req_cost;
         h._pure_accumulated += req_cap;
         dispatched += req_cap;
+        _queued_capacity -= req_cap;
 
         cb(req);
 
--- a/src/core/fair_queue.cc
+++ b/src/core/fair_queue.cc
@@ -117,6 +117,10 @@ void fair_group::replenish_capacity(clock_type::time_point now) noexcept {
     _token_bucket.replenish(now);
 }
 
+void fair_group::refund_tokens(capacity_t cap) noexcept {
+    _token_bucket.refund(cap);
+}
+
 void fair_group::maybe_replenish_capacity(clock_type::time_point& local_ts) noexcept {
     auto now = clock_type::now();
     auto extra = _token_bucket.accumulated_in(now - local_ts);
@@ -223,35 +227,22 @@ void fair_queue::unplug_class(class_id cid) noexcept {
     unplug_priority_class(*_priority_classes[cid]);
 }
 
-auto fair_queue::grab_pending_capacity(const fair_queue_entry& ent) noexcept -> grab_result {
-    _group.maybe_replenish_capacity(_group_replenish);
-
-    if (_group.capacity_deficiency(_pending->head)) {
-        return grab_result::pending;
+auto fair_queue::reap_pending_capacity() noexcept -> reap_result {
+    auto result = reap_result{.ready_tokens = 0, .our_turn_has_come = true};
+    if (_pending.cap) {
+        capacity_t deficiency = _group.capacity_deficiency(_pending.head);
+        result.our_turn_has_come = deficiency <= _pending.cap;
+        if (result.our_turn_has_come) {
+            result.ready_tokens = _pending.cap - deficiency;
+            _pending.cap = deficiency;
+        }
     }
-
-    capacity_t cap = ent._capacity;
-    if (cap > _pending->cap) {
-        return grab_result::cant_preempt;
-    }
-
-    _pending.reset();
-    return grab_result::grabbed;
+    return result;
 }
 
-auto fair_queue::grab_capacity(const fair_queue_entry& ent) noexcept -> grab_result {
-    if (_pending) {
-        return grab_pending_capacity(ent);
-    }
-
-    capacity_t cap = ent._capacity;
+auto fair_queue::grab_capacity(capacity_t cap) noexcept -> void {
     capacity_t want_head = _group.grab_capacity(cap);
-    if (_group.capacity_deficiency(want_head)) {
-        _pending.emplace(want_head, cap);
-        return grab_result::pending;
-    }
-
-    return grab_result::grabbed;
+    _pending = pending{want_head, cap};
 }
 
 void fair_queue::register_priority_class(class_id id, uint32_t shares) {
@@ -309,7 +300,7 @@ void fair_queue::notify_request_cancelled(fair_queue_entry& ent) noexcept {
 }
 
 fair_queue::clock_type::time_point fair_queue::next_pending_aio() const noexcept {
-    if (_pending) {
+    if (_pending.cap) {
         /*
          * We expect the disk to release the ticket within some time,
          * but it's ... OK if it doesn't -- the pending wait still
@@ -320,7 +311,7 @@ fair_queue::clock_type::time_point fair_queue::next_pending_aio() const noexcept
          * which's sub-optimal. The expectation is that we think disk
          * works faster, than it really does.
          */
-        auto over = _group.capacity_deficiency(_pending->head);
+        auto over = _group.capacity_deficiency(_pending.head);
         auto ticks = _group.capacity_duration(over);
         return std::chrono::steady_clock::now() + std::chrono::duration_cast<std::chrono::microseconds>(ticks);
     }
@@ -328,11 +319,31 @@ fair_queue::clock_type::time_point fair_queue::next_pending_aio() const noexcept
     return std::chrono::steady_clock::time_point::max();
 }
 
+// This function is called by the shard on every poll.
+// It picks up tokens granted by the group, spends available tokens on IO dispatches,
+// and makes a reservation for more tokens, if needed.
+//
+// Reservations are done in batches of size `_group.per_tick_grab_threshold()`.
+// During contention, in an average moment in time each contending shard can be expected to
+// be holding a reservation of such size after the current head of the token bucket.
+//
+// A shard which is currently calling `dispatch_requests()` can expect a latency
+// of at most `nr_contenders * (_group.per_tick_grab_threshold() + max_request_cap)` before its next reservation is fulfilled.
+// If a shard calls `dispatch_requests()` at least once per X total tokens, it should receive bandwidth
+// of at least `_group.per_tick_grab_threshold() / (X + nr_contenders * (_group.per_tick_grab_threshold() + max_request_cap))`.
+//
+// A shard which is polling continuously should be able to grab its fair share of the disk for itself.
+//
+// Given a task quota of 500us and IO latency goal of 750 us,
+// a CPU-starved shard should still be able to grab at least ~30% of its fair share in the worst case.
+// This is far from ideal, but it's something.
 void fair_queue::dispatch_requests(std::function<void(fair_queue_entry&)> cb) {
-    capacity_t dispatched = 0;
-    boost::container::small_vector<priority_class_ptr, 2> preempt;
+    _group.maybe_replenish_capacity(_group_replenish);
 
-    while (!_handles.empty() && (dispatched < _group.per_tick_grab_threshold())) {
+    const uint64_t max_unamortized_reservation = _group.per_tick_grab_threshold();
+    auto available = reap_pending_capacity();
+
+    while (!_handles.empty()) {
         priority_class_data& h = *_handles.top();
         if (h._queue.empty() || !h._plugged) {
             pop_priority_class(h);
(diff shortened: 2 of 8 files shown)


michoecho:
Refs (to some degree: fixes) https://github.com/scylladb/seastar/issues/2615

src/core/fair_queue.cc
```diff
@@ -318,37 +319,100 @@ fair_queue::clock_type::time_point fair_queue::next_pending_aio() const noexcept
...
+// If shards have low CPU utilization (and so X is negligible), this should result in near-perfect fairness.
+// If shards have high CPU utilization (and so they poll rarely and X is a sizeable fraction of the bucket),
+// it still gives reasonable fairness: with task quota of 500us and IO latency goal of 750 us,
+// a CPU-starved shard should still get at least 60% of fair throughput.
 void fair_queue::dispatch_requests(std::function<void(fair_queue_entry&)> cb) {
-    capacity_t dispatched = 0;
-    boost::container::small_vector<priority_class_ptr, 2> preempt;
+    _group.maybe_replenish_capacity(_group_replenish);
+
+    const uint64_t max_unarmortized_reservation = _group.per_tick_grab_threshold();
```
xemul, line 349: nit: una~~R~~mortized

include/seastar/core/fair_queue.hh
```diff
@@ -331,35 +351,49 @@ private:
...
     };
+    pending _pending;
 
-    std::optional<pending> _pending;
+    // Total capacity of all requests waiting in the queue.
+    // We limit token reservations to this value, to avoid reserving tokens that
+    // we don't need. This isn't strictly necessary, since grabbing more than needed and
+    // returning the surplus should work too. But it feels like the natural thing to do,
+    // and should probably result in slightly better latency if the node is both CPU and IO
+    // are heavily utilized.
```
michoecho, line 378:
```suggestion
    // and should probably result in slightly better latency if both CPU and IO on the node
    // are heavily utilized.
```

include/seastar/core/fair_queue.hh
```diff
@@ -125,16 +125,34 @@ public:
...
+///
+/// Groups provide cross-shard coordination, which is needed to keep the lower-layer queues just long enough
+/// to saturate the disk regardless of how many shards are contending for the queue.
+///
+/// If there was no coordination, each shard would need to guess between:
+/// 1. Optimistically assuming that it's the only contender,
+///    and dispatching IO with concurrency high enough to fully utilize the disk by itself,
+///    which would result in long lower-level queues if there are other contenders.
+/// 2. Pesimistically assuming that every shard is a contender, and dispatching IO with concurrency just low enough
+///    to take its fair share, which would waste bandwidth in case there are actually less contenders.
```
michoecho, line 153:
```suggestion
///    to take its fair share, which would waste bandwidth in case there are actually fewer contenders.
```

include/seastar/core/fair_queue.hh
```diff
@@ -331,35 +351,49 @@ private:
...
-    enum class grab_result { grabbed, cant_preempt, pending };
-    grab_result grab_capacity(const fair_queue_entry& ent) noexcept;
-    grab_result grab_pending_capacity(const fair_queue_entry& ent) noexcept;
+    // Replaces _pending with a new reservation starting at the current
+    // group bucket tail.
+    void grab_capacity(const fair_queue_entry& ent) noexcept;
+    // Shaves off the fulfilled frontal part from `_pending` (if any),
+    // and returns the fulfilled tokens.
+    // Sets `our_turn_has_come` to the truth value of "There are no unfulfilled reservations
+    // (from other shards) earlier than `_pending`".
```
michoecho, line 393:
```suggestion
    // Sets `our_turn_has_come` to the truth value of "`_pending` is empty or there are no unfulfilled reservations
    // (from other shards) earlier than `_pending`".
```

src/core/fair_queue.cc
```diff
@@ -318,37 +319,100 @@ fair_queue::clock_type::time_point fair_queue::next_pending_aio() const noexcept
...
+// If a shard calls `dispatch_requests()` at least once per X total tokens, it should receive bandwidth
+// of at least `_group.per_tick_grab_threshold() / (X + nr_contenders * _group.per_tick_grab_threshold())`.
+// If shards have low CPU utilization (and so X is negligible), this should result in near-perfect fairness.
+// If shards have high CPU utilization (and so they poll rarely and X is a sizeable fraction of the bucket),
+// it still gives reasonable fairness: with task quota of 500us and IO latency goal of 750 us,
+// a CPU-starved shard should still get at least 60% of fair throughput.
 void fair_queue::dispatch_requests(std::function<void(fair_queue_entry&)> cb) {
-    capacity_t dispatched = 0;
-    boost::container::small_vector<priority_class_ptr, 2> preempt;
+    _group.maybe_replenish_capacity(_group_replenish);
```
xemul, line 347: Is it needed here? The next call to reap_pending_capacity() calls it when it needs

src/core/fair_queue.cc
```diff
@@ -318,37 +319,100 @@ fair_queue::clock_type::time_point fair_queue::next_pending_aio() const noexcept
...
+            // above the head.
+            // Proof sketch: it can be observed that at any moment each shard contributes
+            // to `tail` at most `max_unarmortized_reservation` tokens which weren't already paid by some previous addition to `head`.
+            capacity_t grab_amount = std::min<capacity_t>(recycled + max_unarmortized_reservation, _queued_capacity);
+            // There's technically nothing wrong with grabbing more than `_group.maximum_capacity()`,
+            // but the token bucket has an assert for that, and its a reasonable expectation, so let's respect that limit.
+            // It shouldn't matter in practice.
+            grab_amount = std::min<capacity_t>(grab_amount, _group.maximum_capacity());
+            _group.refund_tokens(recycled);
+            grab_capacity(grab_amount);
```
xemul, line 402: There's no `fair_queue::grab_amount(capacity_t)` method here, only the `fair_queue::grab_capacity(const fair_queue_entry&)` one, how does it compile?

src/core/fair_queue.cc
michoecho: Oh, that's an oops. Originally I put the `maybe_replenish_capacity` in `reap_pending_capacity`, then decided it's cleaner to call it just once per `dispatch_requests`, so I moved it outside of `reap_pending_capacity`. There is even a comment about that:
```
    // Assumes that `_group.maybe_replenish_capacity()` was called recently.
    capacity_t reap_pending_capacity(bool& our_turn_has_come) noexcept;
```

But apparently I copy-pasted the old variant of `reap_pending_capacity` into the PR at some point, so the `maybe_replenish_capacity` call ended up being in both places.

src/core/fair_queue.cc
michoecho: `grab_amount` is a local variable, though? Not a function.

michoecho: `grab_amount` is a local variable, though? Not a function.

src/core/fair_queue.cc
michoecho: Oh, that's an oops. Originally I put the `maybe_replenish_capacity` in `reap_pending_capacity`, then decided it's cleaner to call it just once per `dispatch_requests`, so I moved it outside of `reap_pending_capacity`. There is even a comment about that:
```
    // Assumes that `_group.maybe_replenish_capacity()` was called recently.
    capacity_t reap_pending_capacity(bool& our_turn_has_come) noexcept;
```

But apparently I copy-pasted the old variant of `reap_pending_capacity` into the PR at some point, so the `maybe_replenish_capacity` call ended up being in both places.

src/core/fair_queue.cc
xemul: My misprint :( updated the comment

src/core/fair_queue.cc
```diff
@@ -223,35 +231,26 @@ void fair_queue::unplug_class(class_id cid) noexcept {
...
-    capacity_t cap = ent._capacity;
-    if (cap > _pending->cap) {
-        return grab_result::cant_preempt;
+auto fair_queue::reap_pending_capacity(bool& our_turn_has_come) noexcept -> capacity_t {
+    capacity_t ready = 0;
+    if (_pending.cap) {
+        _group.maybe_replenish_capacity(_group_replenish);
+        capacity_t deficiency = _group.capacity_deficiency(_pending.head);
+        our_turn_has_come = deficiency <= _pending.cap;
+        if (deficiency < _pending.cap) {
```
xemul, line 240: nit: Took me a while to understand that this if is equivalent to `if (our_turn_has_come)`

src/core/fair_queue.cc
```diff
@@ -318,37 +319,100 @@ fair_queue::clock_type::time_point fair_queue::next_pending_aio() const noexcept
...
+            // to `tail` at most `max_unarmortized_reservation` tokens which weren't already paid by some previous addition to `head`.
+            capacity_t grab_amount = std::min<capacity_t>(recycled + max_unarmortized_reservation, _queued_capacity);
+            // There's technically nothing wrong with grabbing more than `_group.maximum_capacity()`,
+            // but the token bucket has an assert for that, and its a reasonable expectation, so let's respect that limit.
+            // It shouldn't matter in practice.
+            grab_amount = std::min<capacity_t>(grab_amount, _group.maximum_capacity());
+            _group.refund_tokens(recycled);
+            grab_capacity(grab_amount);
+            ready = reap_pending_capacity(our_turn_has_come);
             continue;
```
xemul, line 404: Let's assume we grabbed some tokens with `grab_capacity(grab_amount)` (presumably this means that pending.cap == N in this case) and see some non-zero deficiency in `reap_pending_capacity()`. In that case it can happen that `pending.cap + ready == deficiency_after_grab`. Respectively, next loop iteration, if we get to the same `else if (our_turn_has_come)` branch, we'll have `recycled = pending.cap + ready = deficiency_after_grab (of previous loop iteration)`, so we just return the grabbed capacity and are thus back to the original state with head and tail fast-forwarded. So race progress seems not to be provided.

src/core/fair_queue.cc
michoecho: Huh, you're right. Apparently it works because there's an implicit conversion from `capacity_t` to `fair_queue_entry`:
https://github.com/scylladb/seastar/blob/faf6a5df9597d59b1e9d9dd60a6c6b9cc9a3608a/include/seastar/core/fair_queue.hh#L118-L119

But it does the expected thing, so... whatever, I guess?

src/core/fair_queue.cc
michoecho: I don't understand the sequence of steps you are point out. Could you say the same thing, but with more explicit steps, and maybe some example numbers? 

But about progress, I can say this:
every time we go through the `else if (our_turn_has_come)` branch without dispatching something, our total reservation (i.e. the value of `ready + _pending.cap` at the entrance to the body of `while`) grows by `max_unarmortized_reservation`. (Well, up to `min(_group.maximum_capacity(), _queued_capacity)`, but that's not important, as both are guaranteed to be big enough to dispatch at least one request).

After several loops, the reservation must grow big enough so that `req._capacity <= ready + _pending.cap` becomes true, and then we won't enter the `else if (our_turn_has_come)` branch again, because we will be entering the previous `else if` instead.

Or at least that's how it's supposed to be. Did I mess up the implementation?

michoecho: The doubt was answered during a conversation on a different channel.

michoecho:
v2:
Functional changes:
- Removed the `maybe_replenish_capacity()` call from `reap_pending_capacity()`. Now `maybe_replenish_capacity()` is only called from `dispatch_requests()`.
- Fixed an incompatibility between the patch and `fair_queue_test`.
Cosmetic changes:
- Addressed review comments:
  - `reap_pending_capacity` now returns `our_turn_has_come` via a return value rather than via an output parameter.
  - The addition of `_queued_capacity` is now in a separate commit.
  - Several typos were fixed.
- Removed the rewrite of the doc comment of `fair_group`, it didn't really belong in this PR. 
- Some misleading comments were rewritten.

@xemul Please re-review.

But I was thinking: is it really necessary to have a "timeline" of dispatches, where each shard only performs the dispatch after the bucket head reaches its `_pending.head`?

If this isn't necessary, then I see an alternative which is both simpler and has better guarantees. See https://github.com/michoecho/seastar/commit/d5a559aa9118b54513ae20f9bd2316a3ed587019. The idea is: each shard has a `_balance` variable. While `_balance` is non-negative, and there any queued requests, we dispatch the request and subtract its cost from `_balance`. While `_balance` is negative, we grab (reserve) a fixed batch of tokens, and add it to `_balance` after the reservation is fulfilled by bucket head.

This means that we dispatch first, then we grab tokens to pay for that, and only then we dispatch again. There is no attempt to use the `_pending.head` as an ideal dispatch timepoint; grabs are only used to limit the throughput. Is there a problem with that?

xemul:
> But I was thinking: is it really necessary to have a "timeline" of dispatches, where each shard only performs the dispatch after the bucket head reaches its _pending.head?

If we want to make shards dispatch independently, then timeline can be avoided, but presumably I mis-understanding what you call "a timeline" (see below)

>  While _balance is negative, we grab (reserve) a fixed batch of tokens, and add it to _balance after the reservation is fulfilled by bucket head.

This "after the reservation is fulfilled by bucket head", how does it differ from "each shard only performs the dispatch after the bucket head reaches its _pending.head"?

apps/io_tester/test_cases/one_cpu_starved_shard_can_still_saturate_io.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
xemul, line 1: Please, move all new tests into `tests/manual/`

michoecho:
> > But I was thinking: is it really necessary to have a "timeline" of dispatches, where each shard only performs the dispatch after the bucket head reaches its _pending.head?
> 
> If we want to make shards dispatch independently, then timeline can be avoided, but presumably I mis-understanding what you call "a timeline" (see below)
> 
> > While _balance is negative, we grab (reserve) a fixed batch of tokens, and add it to _balance after the reservation is fulfilled by bucket head.
> 
> This "after the reservation is fulfilled by bucket head", how does it differ from "each shard only performs the dispatch after the bucket head reaches its _pending.head"?

@xemul Imagine that `per_tick_grab_threshold` is 10 tokens, while all requests cost 100 tokens, and there are 7 shards.
When we are sticking to the "timeline" idea, we want one shard to dispatch at timepoint 100, another at 200, ..., last at 700. And then another round at 800, 900, etc.
If we ignore the "timeline", then we are fine with all 7 shards dispatching together at timepoint 0, then eating tokens (in round robin) to pay for that, and then all dispatching again at ~700.

The main benefit of that is that the total amount of pending reservations has a hard limit at `nr_contenders * per_tick_grab_threshold`. With the design currently used in this PR, where we put big requests on the "timeline" in their entirety, the total amount of pending reservations can reach `nr_contenders * (per_tick_grab_threshold + max_request_size)`.

The "fairness" of the entire batching idea is that each shard can grab a batch once per "round robin", so the shorter the round robin, the better the fairness. In this case, for our default reactor parameters, the improvement from `nr_contenders * (per_tick_grab_threshold + max_request_size)` to `nr_contenders * per_tick_grab_threshold` should result in an improvement in worst-case fairness guarantees from "each shard gets at least ~30% of its fair share" to "each shard gets at least ~60% of its fair share".

src/core/fair_queue.cc
xemul: Please fix it [with the assumption](https://github.com/scylladb/seastar/pull/2625) that fair_queue_entry constructor is `explicit`

xemul:
> If we ignore the "timeline", then we are fine with all 7 shards dispatching together at timepoint 0, then eating tokens (in 5 rotations of round robin) to pay for their requests, and then all dispatching again at ~700.

OK, but isn't it another "timeline"?

> Imagine that per_tick_grab_threshold is 20 tokens, while all requests cost 100 tokens, and there are 7 shards.
When we are sticking to the "timeline" idea, we want one shard to dispatch at timepoint 100, another at 200, ..., last at 700. And then another round at 800, 900, etc.
If we ignore the "timeline", then we are fine with all 7 shards dispatching together at timepoint 0, then eating tokens (in 5 rotations of round robin) to pay for their requests, and then all dispatching again at ~700.

It can work both ways. We overload the disk in a hope that it will handle the load within the upcoming idling time (corresponding to 700 tokens). On one hand, we increased the in-disk concurrency and it can do better job than serving requests coming one-by-one, but on the other, in-disk load->latency doesn't scale linearly, so disk may want time equivalent to, say, 1400 tokens, so once we continue dispatching after 700 disk will still be busy.

michoecho:
> > If we ignore the "timeline", then we are fine with all 7 shards dispatching together at timepoint 0, then eating tokens (in 5 rotations of round robin) to pay for their requests, and then all dispatching again at ~700.
> 
> OK, but isn't it another "timeline"?

It is. Nomenclature is hard. What I meant is that in the "timeline" version, each request has its own designated "slot" in the timeline. If the disk had max concurrency of 1, that slot would match the time span when the disk is handling that request.
In the "non-timeline" version, we aren't trying to order requests so strictly, we are just trying to dispatch them slowly enough, without caring about "local" overlaps.

> in-disk load->latency doesn't scale linearly, so disk may want time equivalent to, say, 1400 tokens

@xemul Is that so? And did we ever observe that? Do we have some past experiments?

In the model of reality in my head, sending several requests immediately should never result in worse disk utilization/speed than spacing them out manually. If that's not true, what's the mechanism which makes the "spaced out" version better? If we send too many requests at once, does it result in some extra delay in the kernel or something?

If the "spaced-out timeline" is important, then I have yet another idea, which combines the "old" and the "new" version — have two separate `shared_token_bucket`s, one for governing fair throughput (this bucket would be replenished by the passage of time, and each shard would grab fixed batches of tokens from it), and one for governing the "spacing out". When a shard was ready to dispatch the request based on the amount of tokens it grabbed from the first bucket, it would make a grab (for the full capacity of the current request) in the second bucket, and it would wait with the actual dispatch to disk until the second bucket head fulfills this grab. (The second bucket would always be replenished with exactly the same amount of tokens as the first bucket). But this is extra complex.

michoecho:
> > in-disk load->latency doesn't scale linearly, so disk may want time equivalent to, say, 1400 tokens
> 
> @xemul Is that so? And did we ever observe that? Do we have some past experiments?
> 
> In the model of reality in my head, sending several requests immediately should never result in worse disk utilization/speed than spacing them out manually. If that's not true, what's the mechanism which makes the "spaced out" version better? If we send too many requests at once, does it result in some extra delay in the kernel or something?
> 
> If the "spaced-out timeline" is important, then I have yet another idea, which combines the "old" and the "new" version — have two separate `shared_token_bucket`s, one for governing fair throughput (this bucket would be replenished by the passage of time, and each shard would grab fixed batches of tokens from it), and one for governing the "spacing out". When a shard was ready to dispatch the request based on the amount of tokens it grabbed from the first bucket, it would make a grab (for the full capacity of the current request) in the second bucket, and it would wait with the actual dispatch to disk until the second bucket head fulfills this grab. (The second bucket would always be replenished with exactly the same amount of tokens as the first bucket). But this is extra complex.

@xemul Ping.

tests/manual/iosched_reproducers/one_cpu_starved_shard_can_still_saturate_io.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+# Test scenario:
+# A single CPU-starved shard has a batch IO job.
```
avikivity, line 4: It's not really starved (IIUC). It's saturated but the I/O fiber will receive CPU every task quota, which is as much as it can expect.

tests/manual/iosched_reproducers/one_cpu_starved_shard_can_still_saturate_io.sh
```diff
@@ -0,0 +1,39 @@
...
+  shard_info:
+    parallelism: 50
+    reqsize: 128kB
+    shares: 200
+- name: cpuhog
+  type: cpu
+  shards: [0]
+  shard_info:
+    parallelism: 1
+    execution_time: 550us
```
avikivity, line 27: Well, almost, since this is above the task quota, but not by much.

avikivity:
While I couldn't follow the code, I think I can follow the reasoning in the detailed patch changelogs and they make sense to me.

tests/manual/iosched_reproducers/one_cpu_starved_shard_can_still_saturate_io.sh
michoecho: "CPU-starved" here means that there is always more work for the CPU to do. The important part is that the shard only polls once per ~500us, not ~50us.

michoecho: I know that "starved" usually has another meaning, but I don't have a better word here. Saturated?

avikivity: The I/O fiber would be starved if it needed CPU but got it less than once per task quota.

The reactor (not the I/O fiber) is saturated: it doesn't have any cycles to spare, so it isn't able to provide better-than-expected service (CPU every 1usec). But it's not starving the I/O fiber, merely sticking to the expectations.

I'd read "reactor starved" to mean that something outside the reactor is eating the CPU.

mykaul:
@xemul - can you give this another round of review, and let's see how far off it is from inclusion?

bhalevy:
> @xemul - can you give this another round of review, and let's see how far off it is from inclusion?

ping @xemul

xemul:
> @xemul Is that so? And did we ever observe that? Do we have some past experiments?

Sorry. Yes, we did. The whole scheduler math is built around assumption that disk cannot be re-loaded. We have scylladb/diskplorer plots showing that "too much requests" render larger latency for each. Also I have metrics run against two distinct io-properties.yaml files, one "correct" and another one "bursty" (i.e. -- with write iops set 2x times larger than they should) and the "bursty" one shows higher in-disk latency.

xemul:
> In the model of reality in my head, sending several requests immediately should never result in worse disk utilization/speed than spacing them out manually. If that's not true, what's the mechanism which makes the "spaced out" version better? If we send too many requests at once, does it result in some extra delay in the kernel or something?

If that was true we'd better not queue requests in seastar and send all of them into disk. You're right, that, say, 2 requests in disk is better than 1, but at the same time something like, say, 200 is worse than 100. The goal is to find a balance point where concurrency increase doesn't result in worse latency that we need.

bhalevy:
@xemul so what are the plans for backporting this to 2025.1 and 2024.[1-2]

xemul:
@regevran decided that we don't need to backport, @bhalevy

avikivity:
This is not the place to discuss ScyllaDB branches.

Issue #2222, opened 2024-05-06
create_scheduling_group / scheduling_group_key_create not exception safe when SG key data constructor throws


piodul:
There are two cases where data for a scheduling group key can be constructed:

- When the SG key already exists and a new SG is created (`create_scheduling_group`)
- When a new SG key is being created (`scheduling_group_key_create`)

In both cases, the SG key data constructor can fail. This is an example test for the second case:

```c++
SEASTAR_THREAD_TEST_CASE(sg_key_constructor_exception_when_creating_new_key) {
    struct thrower {
        thrower() {
            throw std::runtime_error("oopsie daisy");
        }
        ~thrower() {
            // Shouldn't get here because the constructor shouldn't succeed
            BOOST_ASSERT(false);
        }
    };
    scheduling_group_key_config thrower_conf = make_scheduling_group_key_config<thrower>();
    BOOST_REQUIRE_THROW(scheduling_group_key_create(thrower_conf).get(), std::runtime_error);
}
```

Even though the constructor of `thrower` always throws, its destructor is called anyway:

```
scheduling_group_test: /home/piodul/code/seastar/tests/unit/scheduling_group_test.cc:392: sg_key_constructor_exception_when_creating_new_key::do_run_test_case()::thrower::~thrower(): Assertion `false' failed.
scheduling_group_test: /home/piodul/code/seastar/tests/unit/scheduling_group_test.cc:392: sg_key_constructor_exception_when_creating_new_key::do_run_test_case()::thrower::~thrower(): Assertion `false' failed.
scheduling_group_test: /home/piodul/code/seastar/tests/unit/scheduling_group_test.cc:392: sg_key_constructor_exception_when_creating_new_key::do_run_test_case()::thrower::~thrower(): Assertion `false' failed.
Aborted (core dumped)
```

It looks like both functions are not exception safe and cause issues later if the SG key data constructor fails with an exception. I believe that both operations should be able to roll back in case of an exception.

piodul:
@mlitvk Your PR https://github.com/scylladb/seastar/pull/2585 only fixed https://github.com/scylladb/seastar/issues/2231 and not this one (https://github.com/scylladb/seastar/issues/2222), correct?

mlitvk:
[...]

PR #2617, opened 2025-01-15
scheduling_group: improve scheduling group creation exception safety

mlitvk:
Improve handling of exceptions during scheduling group and scheduling group key creation, where a user-provided constructor for the keys may fail, for example.

We use a new struct `specific_val` and smart pointers to manage memory allocation, construction and destruction of scheduling group data in a safe manner.

We also reorder the initialization order to make it safer. For example, when creating a scheduling group, first allocate all data and then swap it into the scheduling group's data structure.

Fixes scylladb/seastar#2222

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -393,7 +393,6 @@ private:
     task_queue* pop_active_task_queue(sched_clock::time_point now);
     void insert_activating_task_queues();
     void account_runtime(task_queue& tq, sched_clock::duration runtime);
-    void allocate_scheduling_group_specific_data(scheduling_group sg, unsigned long key_id);
     future<> rename_scheduling_group_specific_data(scheduling_group sg);
     future<> init_scheduling_group(scheduling_group sg, sstring name, sstring shortname, float shares);
     future<> init_new_scheduling_group_key(scheduling_group_key key, scheduling_group_key_config cfg);
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -322,6 +322,14 @@ private:
     std::atomic<bool> _dying{false};
     gate _background_gate;
 
+    inline auto& get_sg_data(const scheduling_group& sg) {
+        return _scheduling_group_specific_data.per_scheduling_group_data[sg._id];
+    }
+
+    inline auto& get_sg_data(unsigned sg_id) {
+        return _scheduling_group_specific_data.per_scheduling_group_data[sg_id];
+    }
+
 private:
     static std::chrono::nanoseconds calculate_poll_time();
     static void block_notifier(int);
--- a/include/seastar/core/scheduling_specific.hh
+++ b/include/seastar/core/scheduling_specific.hh
@@ -37,6 +37,49 @@ namespace seastar {
 namespace internal {
 
 struct scheduling_group_specific_thread_local_data {
+    using val_ptr = std::unique_ptr<void, void (*)(void*) noexcept>;
+    using cfg_ptr = lw_shared_ptr<scheduling_group_key_config>;
+
+    struct specific_val {
+        val_ptr valp;
+        cfg_ptr cfg;
+
+        specific_val() : valp(nullptr, &free), cfg(nullptr) {}
+
+        specific_val(val_ptr&& valp_, const cfg_ptr& cfg_) : valp(std::move(valp_)), cfg(cfg_) {
+            if (valp && cfg->constructor) {
+                cfg->constructor(valp.get());
+            }
+        }
+
+        ~specific_val() {
+            if (valp && cfg->destructor) {
+                cfg->destructor(valp.get());
+            }
+        }
+
+        specific_val(const specific_val& other) = delete;
+        specific_val& operator=(const specific_val& other) = delete;
+
+        specific_val(specific_val&& other) : valp(std::move(other.valp)), cfg(std::move(other.cfg)) {}
+
+        specific_val& operator=(specific_val&& other) {
+            if (this != &other) {
+                valp = std::move(other.valp);
+                cfg = std::move(other.cfg);
+            }
+            return *this;
+        }
+
+        void* get() { return valp.get(); }
+
+        void rename() {
+            if (valp && cfg->rename) {
+                cfg->rename(valp.get());
+            }
+        }
+    };
+
     struct per_scheduling_group {
         bool queue_is_initialized = false;
         /**
@@ -44,10 +87,16 @@ struct scheduling_group_specific_thread_local_data {
          * data. The pointer is not use as is but is cast to a reference
          * to the appropriate type that is actually pointed to.
          */
-        std::vector<void*> specific_vals;
+        std::vector<specific_val> specific_vals;
+
+        void rename() {
+            for (auto& v : specific_vals) {
+                v.rename();
+            }
+        }
     };
     std::array<per_scheduling_group, max_scheduling_groups()> per_scheduling_group_data;
-    std::map<unsigned long, scheduling_group_key_config> scheduling_group_key_configs;
+    std::map<unsigned long, cfg_ptr> scheduling_group_key_configs;
 };
 
 #ifdef SEASTAR_BUILD_SHARED_LIBS
@@ -78,12 +127,12 @@ template<typename T>
 T* scheduling_group_get_specific_ptr(scheduling_group sg, scheduling_group_key key) noexcept {
     auto& data = internal::get_scheduling_group_specific_thread_local_data();
 #ifdef SEASTAR_DEBUG
-    assert(std::type_index(typeid(T)) == data.scheduling_group_key_configs[key.id()].type_index);
+    assert(std::type_index(typeid(T)) == data.scheduling_group_key_configs[key.id()]->type_index);
 #endif
     auto sg_id = internal::scheduling_group_index(sg);
     if (__builtin_expect(sg_id < data.per_scheduling_group_data.size() &&
             data.per_scheduling_group_data[sg_id].queue_is_initialized, true)) {
-        return reinterpret_cast<T*>(data.per_scheduling_group_data[sg_id].specific_vals[key.id()]);
+        return reinterpret_cast<T*>(data.per_scheduling_group_data[sg_id].specific_vals[key.id()].get());
     }
     return nullptr;
 }
@@ -123,9 +172,9 @@ T& scheduling_group_get_specific(scheduling_group_key key) noexcept {
     // return a reference to an element whose queue_is_initialized is
     // false.
     auto& data = internal::get_scheduling_group_specific_thread_local_data();
-    assert(std::type_index(typeid(T)) == data.scheduling_group_key_configs[key.id()].type_index);
+    assert(std::type_index(typeid(T)) == data.scheduling_group_key_configs[key.id()]->type_index);
     auto sg_id = internal::scheduling_group_index(current_scheduling_group());
-    return *reinterpret_cast<T*>(data.per_scheduling_group_data[sg_id].specific_vals[key.id()]);
+    return *reinterpret_cast<T*>(data.per_scheduling_group_data[sg_id].specific_vals[key.id()].get());
 }
 
 /**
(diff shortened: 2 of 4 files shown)


mlitvk:
The CI fails in Seastar.unit.rpc with timeout
This is a known issue: https://github.com/scylladb/seastar/issues/2620

include/seastar/core/scheduling_specific.hh
```diff
@@ -37,17 +37,66 @@ namespace seastar {
 namespace internal {
 
 struct scheduling_group_specific_thread_local_data {
+    using val_ptr = std::unique_ptr<void, void (*)(void*) noexcept>;
+    using cfg_ptr = std::shared_ptr<scheduling_group_key_config>;
```
xemul, line 41: Can it be `seastar::lw_shared_ptr<scheduling_group_key_config>`?

include/seastar/core/scheduling_specific.hh
```diff
@@ -37,17 +37,66 @@ namespace seastar {
 namespace internal {
 
 struct scheduling_group_specific_thread_local_data {
+    using val_ptr = std::unique_ptr<void, void (*)(void*) noexcept>;
```
xemul, line 40: What's the point in making it smart-pointer if you track construction/destruction by hand anyway?

include/seastar/core/reactor.hh
```diff
@@ -322,6 +322,14 @@ private:
...
     gate _background_gate;
 
+    inline auto& get_sg_data(const scheduling_group& sg) {
+        return _scheduling_group_specific_data.per_scheduling_group_data[sg._id];
+    }
+
+    inline auto& get_sg_data(unsigned sg_id) {
+        return _scheduling_group_specific_data.per_scheduling_group_data[sg_id];
+    }
+
```
xemul, line 332: There's no need in those helpers, AFAICS, the existing `get_scheduling_group_specific_thread_local_data()` (and Co) already provide access to the array of per_scheduling_group_data-s

include/seastar/core/reactor.hh
```diff
@@ -393,7 +401,6 @@ private:
     task_queue* pop_active_task_queue(sched_clock::time_point now);
     void insert_activating_task_queues();
     void account_runtime(task_queue& tq, sched_clock::duration runtime);
-    void allocate_scheduling_group_specific_data(scheduling_group sg, unsigned long key_id);
```
xemul, line 396: I'd appreciate if the change "Move allocate_scheduling_group_specific_data() from reactor class to internal namespace" was made as separate preparational patch (in the same PR)

include/seastar/core/scheduling_specific.hh
```diff
@@ -37,17 +37,66 @@ namespace seastar {
...
+    using cfg_ptr = std::shared_ptr<scheduling_group_key_config>;
+
+    struct specific_val {
+        val_ptr valp;
+        cfg_ptr cfg;
+
+        specific_val() : valp(nullptr, &free), cfg(nullptr) {}
+
+        specific_val(val_ptr&& valp_, const cfg_ptr& cfg_) : valp(std::move(valp_)), cfg(cfg_) {
```
xemul, line 49: I'm probably misreading the intent, but  specific_val::cfg  and scheduling_group_specific_thread_local_data::scheduling_group_key_configs are intended to point to the _same_ shared object. But the way this constructor is defined makes it impossible to _share_ the config, the specific_val gets its _copy_ of it. Please, correct me if I'm wrong.

include/seastar/core/scheduling_specific.hh
mlitvk: The point is to manage the dynamic memory allocation and free it automatically

include/seastar/core/reactor.hh
mlitvk: I added it because in several places we access the sg data and we do
```
    auto& sg_data = _scheduling_group_specific_data;
    auto& this_sg = sg_data.per_scheduling_group_data[sg._id];
```
which I thought is a little cumbersome and I didn't find another method for this

mlitvk:
* changed to `lw_shared_ptr` instead of `std::shared_ptr`
* split the commit and added a preliminary commit to change `allocate_scheduling_group_specific_data` into an internal function
* rebase

include/seastar/core/reactor.hh
mlitvk: done

include/seastar/core/scheduling_specific.hh
mlitvk: done

include/seastar/core/scheduling_specific.hh
xemul: Ah, yes, my bad. It's cfg_ptr

include/seastar/core/scheduling_specific.hh
```diff
@@ -37,17 +37,66 @@ namespace seastar {
...
+        ~specific_val() {
+            if (valp && cfg->destructor) {
+                cfg->destructor(valp.get());
+            }
+        }
+
+        specific_val(const specific_val& other) = delete;
+        specific_val& operator=(const specific_val& other) = delete;
+
+        specific_val(specific_val&& other) : valp(std::move(other.valp)), cfg(other.cfg) {}
```
xemul, line 64: Why not `cfg(std::move(other.cfg))`? In general, the lifetime of `specific_val::cfg` is unclear.

Here `other.cfg` is kept non-nullptr, what for? Presumably (but maybe I'm wrong) this is to make `~specific_val()` avoid checking if `cfg` exists or not? But the default constructor `specific_val()` sets cfg to nullptr, so it looks like `cfg` _can_ be null, so why not std::move() it here as well?

include/seastar/core/scheduling_specific.hh
```diff
@@ -37,17 +37,66 @@ namespace seastar {
...
+
+        void rename() {
+            for (auto& v : specific_vals) {
+                v.rename();
+            }
+        }
     };
     std::array<per_scheduling_group, max_scheduling_groups()> per_scheduling_group_data;
-    std::map<unsigned long, scheduling_group_key_config> scheduling_group_key_configs;
+    std::map<unsigned long, cfg_ptr> scheduling_group_key_configs;
```
xemul, line 99: It looks like this map is no longer needed? The very came config can be obtained via `per_scheduling_group_data[context.sg_id].specific_vals[key_id].cfg`? I'm not proposing to change anything right now, just checking if my understanding is correct

include/seastar/core/scheduling_specific.hh
```diff
@@ -37,17 +37,66 @@ namespace seastar {
...
 
 struct scheduling_group_specific_thread_local_data {
+    using val_ptr = std::unique_ptr<void, void (*)(void*) noexcept>;
+    using cfg_ptr = lw_shared_ptr<scheduling_group_key_config>;
+
+    struct specific_val {
+        val_ptr valp;
+        cfg_ptr cfg;
+
+        specific_val() : valp(nullptr, &free), cfg(nullptr) {}
```
xemul, line 47: Isn't `free` defaulted by compiler for unique-ptr?

src/core/reactor.cc
```diff
@@ -4896,81 +4885,81 @@ deallocate_scheduling_group_id(unsigned id) noexcept {
...
-        std::abort();
-    }
-    if (cfg.constructor) {
-        cfg.constructor(val);
+internal::scheduling_group_specific_thread_local_data::specific_val
+allocate_scheduling_group_specific_data(scheduling_group sg, unsigned long key_id, const lw_shared_ptr<scheduling_group_key_config>& cfg) {
+    using val_ptr = internal::scheduling_group_specific_thread_local_data::val_ptr;
+    using specific_val = internal::scheduling_group_specific_thread_local_data::specific_val;
+
+    val_ptr valp(aligned_alloc(cfg->alignment, cfg->allocation_size), &free);
```
xemul, line 4893: Isn't `free` defaulted by compiler for unique-ptr?

src/core/reactor.cc
```diff
@@ -4896,81 +4885,81 @@ deallocate_scheduling_group_id(unsigned id) noexcept {
...
+    if (!valp) {
+        throw std::runtime_error("memory allocation failed");
```
xemul, line 4895: It was `std::abort()` before this patch. Why did you change that?

src/core/reactor.cc
```diff
@@ -4896,81 +4885,81 @@ deallocate_scheduling_group_id(unsigned id) noexcept {
...
+                        auto& this_sg = get_sg_data(sg);
                         this_sg.specific_vals.resize(std::max<size_t>(this_sg.specific_vals.size(), key_id+1));
-                        this_sg.specific_vals[key_id] = allocate_scheduling_group_specific_data(sg, key_id, sg_data.scheduling_group_key_configs[key_id]);
+                        this_sg.specific_vals[key_id] = allocate_scheduling_group_specific_data(sg, key_id, std::move(cfgp));
                     });
                 }
             }
             return make_ready_future();
+        }).then([this, key_id, cfgp] () {
+            _scheduling_group_specific_data.scheduling_group_key_configs[key_id] = std::move(cfgp);
```
xemul, line 4962: You seem to mention this change in patch comment:

> We also reorder the initialization order to make it safer. For
example, when creating a scheduling group, first allocate all data and
then swap it into the scheduling group's data structure.

but don't explain the motivation. Why can't scheduling_group_key_config[key_id] be assigned where it was before this change?

include/seastar/core/scheduling_specific.hh
mlitvk: It's still used in `init_scheduling_group` when we init a new scheduling group and go over all the keys to allocate them for the new sg

src/core/reactor.cc
mlitvk: because the intention of the patch is to improve error handling and allow the node to continue to function in face of unexpected errors
so if we can handle exceptions in this path I don't think there's reason anymore to abort in this specific case

src/core/reactor.cc
mlitvk: it is to avoid the case of a partially initialized scheduling group.
if we allocate some of the values, transfer the ownership to the sg, and then fail, then we remain with a partially initialized scheduling which could cause problems, like if someone assumes it's all be initialized, or if these allocations "leak" because the operation is considered to have failed but we still store the data.
so instead we either allocate all keys or none

include/seastar/core/scheduling_specific.hh
mlitvk: the default uses `delete` and it doesn't work because `can't delete pointer to incomplete type`

src/core/reactor.cc
mlitvk: answered above

include/seastar/core/scheduling_specific.hh
mlitvk: done, I changed to `std::move`
there was no special reason to copy instead of move, I thought it doesn't matter much, but I suppose it's better to move

mlitvk:
* in move constructor of `specific_val`, changed it to move the cfg shared ptr instead of copy
* rebase

src/core/reactor.cc
xemul: The description you gave applies to `init_scheduling_group` -- it iterates over `for (const auto& [key_id, cfg] : sg_data.scheduling_group_key_configs)`, creates temporary `vals` object, then assigns it to sched group data with `std::swap(this_sg.specific_vals, vals);`.

My concern is about `init_new_scheduling_group_key` which looks like this (simplified):
```
    return parallel_for_each (task_queues, [] (tq) {
        sched_groups[tq.id].specific_vals[key] = allocate_sg_specific_data();
    }).then([] {
        sched_group_configs[key] = config;
    });
```

So it may happen that some sched_groups[tq.id] have its specific_vals[key] set and some not, so this

> ... we either allocate all keys or none

seems not to be true. Neither I see the point in assigning config _after_ all values are allocated. Please, elaborate.

mlitvk: ok, I changed it to be like before.

mlitvk:
* changed the initialization order of sg and sg key back to be like before

Issue #47, opened 2015-07-23
ubsan makes debug-build tests executables are very large


nyh:
As explained in commit 3dd4bd250515f3af2ccbb651689ce78de04762bd we strip the test executables, because the debug information makes each one unacceptably large. We strip both the "release" and "debug" tests.

Stripping made the release executables acceptably small, but the debug-mode tests are still very large. For example, for the tcp_client test, we have

```
1.4M    build/release/tests/tcp_client
38M build/debug/tests/tcp_client
```

It turns out the reason for most of the debug executable's size is ubsan (the undefined behavior sanitizer) enabled in the "debug" build. I don't know why, but ubsan adds a very large data section to the executable, which explains much of the executable's increased size. For example:

```
   text    data     bss     dec     hex filename
12360358    26861464      11409 39233231    256a6cf build/debug/tests/tcp_client
1424669    4512   10584 1439765  15f815 build/release/tests/tcp_client
```

As you can see above, the text section of the executable is much larger in the debug build, which was to be expected because the debug build removes all optimizations which especially in C++ keeps tons of code that should have been optimized away. But this increase is dwarfed by the dramaticly increase data section - from 4 KB to 26 MB - storing something related to ubsan. I don't know what it stores, or if it can somehow be reduced or eliminated (without eliminated ubsan, of course).

The very large data section can be seen in the individual object files too. For example:

```
   text    data     bss     dec     hex filename
1803825 3153480     368 4957673  4ba5e9 build/debug/tests/tcp_client.o
 216723       8     101  216832   34f00 build/release/tests/tcp_client.o
```

I checked the different sanitizer options, and the "-fsanitize=undefined" is responsible for all of the huge data segment (without only the other sanitizers enabled, no such increase is seen).

nyh:
I'm closing this issue. We've been using UBSAN in debug builds for years now, we're using it as designed, and the executable size is not something we can improve.

PR #2618, opened 2025-01-15
Revert "rpc stream: do not abort stream queue if stream connection was closed without error"

p12tic:
Note: I'm not sure about seastar policy about reverting changes that break tests. I created the PR with hope that it will save someone time to be able to just click merge. If the policy is to fix, then of course the PR can be closed.

This reverts commit 27f834e8f04f6b7ff9521424e88f0f32e5d53dc1.

This revert fixes test failures in Seastar.unit.rpc test suite, such as:

      Start 47: Seastar.unit.rpc
47/79 Test #47: Seastar.unit.rpc ..............................***Timeout 300.04 sec

The tests fail randomly and more frequently in debug builds.

### DIFF
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -1021,9 +1021,9 @@ namespace rpc {
                       log_exception(*this, log_level::debug, "fail to connect", ep);
                   }
               }
-            _stream_queue.abort(ep);
           }
           _error = true;
+          _stream_queue.abort(std::make_exception_ptr(stream_closed()));
           return stop_send_loop(ep).then_wrapped([this] (future<> f) {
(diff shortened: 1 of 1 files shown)


mykaul:
@gleb-cloudius - can you please take a look?

gleb-cloudius:
> @gleb-cloudius - can you please take a look?

Looking. Not easy to reproduce, but it looks like this is test loopback implementation problem. Aborting the connection not always result in an error on the other side for some reason.

tchaikov:
FWIW, #2620 was created to track the regression in test.

gleb-cloudius (approved):
Lets revert for now. It looks like there are more than one problem here.

PR #2628, opened 2025-01-22
scheduling: auto-detect scheduling group key rename() method

avikivity:
make_scheduling_group_key_config() is able to create a fully functional scheduling_group_key_config, apart from the rename() method. Provide for that by detecting if the type has a rename() method and if so, create a thunk for it.

This is kept optional for backward compatibility.


piodul:
Technically speaking, this change is not forwards compatible because somebody might have a scheduling group key with a `rename()` method but not create a "rename" thunk for it. However, I would assume the chance of this happening to be rather low.

Other than that, LGTM.

PR #1494, opened 2023-02-11
smp: do not inline function when BUILD_SHARED_LIBS

tchaikov:
`this_shard_id_ptr()` references a static thread_local variable defined in the function. if Seastar library is built as a static library, the linker will guarantee that there will be only a single definition of `g_this_shard_id`, and all translation unit will use this very instance.

but if Seastar is compiled and linked as a shared library. the shared library would have its own copy of this variable, while the executable would have another one, because this function is defined in the header. a symptom of this problem is that `this_shard_id()` is always 0 if any of the translation unit in the executable calls this function directly.

in this change, `this_shard_id_ptr()` is only declared in the header, and defined in .cc, if Seastar is built as shared library. so we have a single definition in the shared library.

a caveat: this further slows down Seastar applications linked against the shared library. but since the shared library build is in general only used for testing, this does not matter.

Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>

PR #2629, opened 2025-01-23
net: Add bytes sent/received metrics

StephanDollberg:
Adds two basic metrics that track bytes sent and received from the
networking stack.

This can give a better per application insight than what host wide
metrics (for example via node exporter) can do.

Further it can serve as a comparison point with host based metrics for
possible write amplification (though this is less likely on the network
side).

The patch follows a similar pattern as to how memory and disk based
metrics work.

Note one might be inclined to introduce some grouping into the metrics
in relation to either the interface or source IP similar to how we have
mountpoints and/or groups on the disk side.

While this sounds useful at first in practice it would be less useful.
Often for the major cloud providers and similar in self hosted
environments there is only a single interface/source-IP and routing
happens at a later point (switches, routers etc.). Further, adding this
separation would make the implementation more expensive in either
compute or memory space.

A possible improvement would be add tagging to the API such that users
could be explicit about attribution. However this is left as a future
improvement and this patch keeps it simple for now.

### DIFF
--- a/include/seastar/net/api.hh
+++ b/include/seastar/net/api.hh
@@ -444,6 +444,19 @@ public:
     bool supports_ipv6() const;
 };
 
+struct statistics {
+    uint64_t bytes_sent = 0;
+    uint64_t bytes_received = 0;
+};
+
+namespace metrics {
+class metric_groups;
+class label_instance;
+}
+
+void register_net_metrics_for_scheduling_group(
+    metrics::metric_groups& m, unsigned sg_id, const metrics::label_instance& name);
+
 class network_stack {
 public:
     virtual ~network_stack() {}
@@ -467,6 +480,11 @@ public:
     virtual bool supports_ipv6() const {
         return false;
     }
+
+    // Return network stats (bytes sent/received etc.) for this stack and scheduling group
+    virtual statistics stats(unsigned scheduling_group_id) = 0;
+    // Clears the stats for this stack and scheduling group
+    virtual void clear_stats(unsigned scheduling_group_id) = 0;
 
     /**
      * Returns available network interfaces. This represents a
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -960,6 +960,9 @@ reactor::task_queue::register_stats() {
         }, sm::description("Total amount in milliseconds we were in violation of the task quota"),
            {group_label}),
     });
+
+    register_net_metrics_for_scheduling_group(new_metrics, _id, group_label);
+
     _metrics = std::exchange(new_metrics, {});
 }
 
@@ -2560,7 +2563,6 @@ void reactor::register_metrics() {
             sm::make_counter("abandoned_failed_futures", _abandoned_failed_futures, sm::description("Total number of abandoned failed futures, futures destroyed while still containing an exception")),
     });
 
-    namespace sm = seastar::metrics;
     _metric_groups.add_group("reactor", {
         sm::make_counter("fstream_reads", _io_stats.fstream_reads,
                 sm::description(
--- a/src/net/native-stack-impl.hh
+++ b/src/net/native-stack-impl.hh
@@ -29,6 +29,17 @@ namespace seastar {
 
 extern logger seastar_logger;
 
+namespace internal {
+
+namespace native_stack_net_stats {
+
+inline thread_local std::array<uint64_t, max_scheduling_groups()> bytes_sent = {};
+inline thread_local std::array<uint64_t, max_scheduling_groups()> bytes_received = {};
+
+};
+
+}
+
 namespace net {
 
 using namespace seastar;
@@ -172,6 +183,8 @@ public:
         }
         return _conn->wait_for_data().then([this] {
             _buf = _conn->read();
+            auto sg_id = internal::scheduling_group_index(current_scheduling_group());
+            internal::native_stack_net_stats::bytes_received[sg_id] += _buf.len();
             _cur_frag = 0;
             _eof = !_buf.len();
             return get();
@@ -193,6 +206,8 @@ public:
         : _conn(std::move(conn)) {}
     using data_sink_impl::put;
     virtual future<> put(packet p) override {
+        auto sg_id = internal::scheduling_group_index(current_scheduling_group());
+        internal::native_stack_net_stats::bytes_sent[sg_id] += p.len();
         return _conn->send(std::move(p));
     }
     virtual future<> close() override {
--- a/src/net/native-stack.cc
+++ b/src/net/native-stack.cc
@@ -194,6 +194,18 @@ public:
     friend class native_network_interface;
 
     std::vector<network_interface> network_interfaces() override;
+
+    virtual statistics stats(unsigned scheduling_group_id) override {
+        return statistics{
+            internal::native_stack_net_stats::bytes_sent[scheduling_group_id],
+            internal::native_stack_net_stats::bytes_received[scheduling_group_id],
+        };
+    }
+
+    virtual void clear_stats(unsigned scheduling_group_id) override {
+        internal::native_stack_net_stats::bytes_sent[scheduling_group_id] = 0;
+        internal::native_stack_net_stats::bytes_received[scheduling_group_id] = 0;
+    }
 };
 
 thread_local promise<std::unique_ptr<network_stack>> native_network_stack::ready_promise;
@@ -427,6 +439,6 @@ std::vector<network_interface> native_network_stack::network_interfaces() {
     return res;
 }
 
-}
+} // namespace net
 
 }
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -215,6 +215,8 @@ public:
     virtual bool has_per_core_namespace() override { return _reuseport; };
     bool supports_ipv6() const override;
     std::vector<network_interface> network_interfaces() override;
+    virtual statistics stats(unsigned scheduling_group_id) override;
+    virtual void clear_stats(unsigned scheduling_group_id) override;
 };
 
 class posix_ap_network_stack : public posix_network_stack {
--- a/src/net/posix-stack.cc
+++ b/src/net/posix-stack.cc
@@ -81,6 +81,9 @@ copy_reinterpret_cast(const void* ptr) {
     return tmp;
 }
 
+thread_local std::array<uint64_t, seastar::max_scheduling_groups()> bytes_sent = {};
+thread_local std::array<uint64_t, seastar::max_scheduling_groups()> bytes_received = {};
+
 }
 
 namespace seastar {
@@ -637,6 +640,8 @@ posix_data_source_impl::get() {
             _config.buffer_size /= 2;
             _config.buffer_size = std::max(_config.buffer_size, _config.min_buffer_size);
         }
+        auto sg_id = internal::scheduling_group_index(current_scheduling_group());
+        bytes_received[sg_id] += b.size();
         return b;
     });
 }
@@ -671,12 +676,16 @@ std::vector<iovec> to_iovec(std::vector<temporary_buffer<char>>& buf_vec) {
 
 future<>
 posix_data_sink_impl::put(temporary_buffer<char> buf) {
+    auto sg_id = internal::scheduling_group_index(current_scheduling_group());
+    bytes_sent[sg_id] += buf.size();
     return _fd.write_all(buf.get(), buf.size()).then([d = buf.release()] {});
 }
 
 future<>
 posix_data_sink_impl::put(packet p) {
     _p = std::move(p);
+    auto sg_id = internal::scheduling_group_index(current_scheduling_group());
+    bytes_sent[sg_id] += _p.len();
     return _fd.write_all(_p).then([this] { _p.reset(); });
 }
 
@@ -876,6 +885,8 @@ future<> posix_datagram_channel::send(const socket_address& dst, const char *mes
     auto len = strlen(message);
     auto a = dst;
     resolve_outgoing_address(a);
+    auto sg_id = internal::scheduling_group_index(current_scheduling_group());
+    bytes_sent[sg_id] += len;
     return _fd.sendto(a, message, len)
             .then([len] (size_t size) { assert(size == len); });
 }
@@ -883,6 +894,8 @@ future<> posix_datagram_channel::send(const socket_address& dst, const char *mes
 future<> posix_datagram_channel::send(const socket_address& dst, packet p) {
     auto len = p.len();
     _send.prepare(dst, std::move(p));
+    auto sg_id = internal::scheduling_group_index(current_scheduling_group());
+    bytes_sent[sg_id] += len;
     return _fd.sendmsg(&_send._hdr)
             .then([len] (size_t size) { assert(size == len); });
 }
@@ -954,6 +967,8 @@ posix_datagram_channel::receive() {
                 break;
             }
         }
+        auto sg_id = internal::scheduling_group_index(current_scheduling_group());
+        bytes_received[sg_id] += size;
         return make_ready_future<datagram>(datagram(std::make_unique<posix_datagram>(
             _recv._src_addr, dst ? *dst : _address, packet(fragment{_recv._buffer, size}, make_deleter([buf = _recv._buffer] { delete[] buf; })))));
     }).handle_exception([p = _recv._buffer](auto ep) {
@@ -1197,6 +1212,18 @@ std::vector<network_interface> posix_network_stack::network_interfaces() {
     }();
 
     return std::vector<network_interface>(thread_local_interfaces.begin(), thread_local_interfaces.end());
+}
+
+statistics posix_network_stack::stats(unsigned scheduling_group_id) {
+    return statistics{
+        bytes_sent[scheduling_group_id],
+        bytes_received[scheduling_group_id],
+    };
+}
+
+void posix_network_stack::clear_stats(unsigned scheduling_group_id) {
+    bytes_sent[scheduling_group_id] = 0;
+    bytes_received[scheduling_group_id] = 0;
 }
 
 }
--- a/src/net/stack.cc
+++ b/src/net/stack.cc
@@ -31,6 +31,8 @@ module;
 #ifdef SEASTAR_MODULE
 module seastar;
 #else
+#include <seastar/core/metrics_api.hh>
+#include <seastar/core/reactor.hh>
 #include <seastar/net/stack.hh>
 #include <seastar/net/inet_address.hh>
 #endif
@@ -284,6 +286,23 @@ network_stack::connect(socket_address sa, socket_address local, transport proto)
 
 std::vector<network_interface> network_stack::network_interfaces() {
     return {};
+}
+
+void register_net_metrics_for_scheduling_group(
+    metrics::metric_groups &metrics, unsigned sg_id, const metrics::label_instance& name) {
+    namespace sm = seastar::metrics;
+    metrics.add_group("network", {
+        sm::make_counter("bytes_sent", [sg_id] { return engine().net().stats(sg_id).bytes_sent; },
+                sm::description("Counts the number of bytes written to network sockets."), {name}),
+        sm::make_counter("bytes_received", [sg_id] { return engine().net().stats(sg_id).bytes_received; },
+                sm::description("Counts the number of bytes received from network sockets."), {name}),
+    });
+
+    // need to clear stats in case we recreated a SG with the same id
+    // but avoid during reactor startup
+    if (engine_is_ready()) {
+        engine().net().clear_stats(sg_id);
+    }
 }
 
 }


src/net/native-stack.cc
```diff
@@ -427,6 +439,17 @@ std::vector<network_interface> native_network_stack::network_interfaces() {
...
+} // namespace net
+
+namespace internal {
+
+void increment_bytes_sent(uint64_t bytes) {
+    net::bytes_sent += bytes;
 }
+void increment_bytes_received(uint64_t bytes) {
+    net::bytes_received += bytes;
+}
```
avikivity, line 451: This must be in the same translation unit as its used in, statistics must compile to a single ADD instruction.

src/net/native-stack-impl.hh
```diff
@@ -193,6 +201,7 @@ public:
         : _conn(std::move(conn)) {}
     using data_sink_impl::put;
     virtual future<> put(packet p) override {
+        internal::increment_bytes_sent(p.len());
```
avikivity, line 204: Better place is the network interface class.

avikivity:
For write amplification, you can compare TCP bytes sent vs interface bytes sent (not sure such a counter exists), or just count packets and multiple by typical header size.

It would be more interesting to make these counters per scheduling group. As is the difference between this and the host counters is too trivial.

src/net/native-stack-impl.hh
StephanDollberg: You mean to measure the bytes or to store the counter. I have put the counter into a separate class now as most of the classes in the native stack are templates so it's not immediately easy to put static members there.

StephanDollberg:
> It would be more interesting to make these counters per scheduling group. As is the difference between this and the host counters is too trivial.

This sounds interesting yes but I think it goes a bit against the "single ADD instruction" as the streams are not directly linked to any scheduling group so we'd always have to get the current group (maybe via the scheduling_group_specific stuff)?

travisdowns:
Just last night I was having a very sweet dream  :sleeping:  that seastar had network in/out metrics, which were in my top 5 metrics along with aio write/write-bytes, reactor utilization and tasks processed, helping me crush performance problems almost as soon as they materialize.

I no longer had to beg end users to separately ship me host-level metrics, I never had to argue with them that _something else_ on the host was slamming the network interface, I never had to guess what shard was doing all the network IO, I just had net metrics there along with all my other seasar metrics, the same labels, aggregation rules, `?foo=bar` query capabilities on my metrics endpoint, I never had to struggle with weird k8s setups with interface names apparently generated like strong passwords, not with per-cgroup statistics which aren't rolled up into the host-level metrics [1].

Then I woke up :disappointed: : .

---

[1] Admittedly, this doesn't happen for bytes as far as I'm aware, but does occur for other metrics like "connections": node_exporter doesn't (and won't) include all connections in their metrics, since the /proc files they query only includes connections for the root cgroup, and not for the cgroups where the action is happening, so your "host level" metrics report very few connections, even if you have 1 million hidden in a pod. So this would be a great case for a seastar-level connection metric.

StephanDollberg:
> but I think it goes a bit against the "single ADD instruction"

Looking at it it seems to take about 2-3 instructions.

src/net/native-stack-impl.hh
avikivity: > You mean to measure the bytes or to store the counter?


Both.

>  I have put the counter into a separate class now as most of the classes in the native stack are templates so it's not immediately easy to put static members there.

It's okay to make the counters namespace-scope.

avikivity:
> > but I think it goes a bit against the "single ADD instruction"
> 
> Looking at it it seems to take about 2-3 instructions.

That's reasonable. The idea is that those instructions don't block the rest of the instruction flow since nothing depends on them.

avikivity:
> Just last night I was having a very sweet dream 😴 that seastar had network in/out metrics, which were in my top 5 metrics along with aio write/write-bytes, reactor utilization and tasks processed, helping me crush performance problems almost as soon as they materialize.
> 
> I no longer had to beg end users to separately ship me host-level metrics, I never had to argue with them that _something else_ on the host was slamming the network interface, I never had to guess what shard was doing all the network IO, I just had net metrics there along with all my other seasar metrics, the same labels, aggregation rules, `?foo=bar` query capabilities on my metrics endpoint, I never had to struggle with weird k8s setups with interface names apparently generated like strong passwords, not with per-cgroup statistics which aren't rolled up into the host-level metrics [1].
> 
> Then I woke up 😞 .


It is good to have dreams to inspire you and as an escape from dreadful reality.

> 
> [1] Admittedly, this doesn't happen for bytes as far as I'm aware, but does occur for other metrics like "connections": node_exporter doesn't (and won't) include all connections in their metrics, since the /proc files they query only includes connections for the root cgroup, and not for the cgroups where the action is happening, so your "host level" metrics report very few connections, even if you have 1 million hidden in a pod. So this would be a great case for a seastar-level connection metric.

We found that connection counts are better when associated with a scope. For example, client-server connection counts vs node-node connection counts.

travisdowns:
> We found that connection counts are better when associated with a scope. For example, client-server connection counts vs node-node connection counts.

Indeed, we have application level connection metrics as well. Still, the host metrics have proven useful there as a second "totally inclusive check" and because there are some states where the connection is still known by the kernel but not by the application or vice-versa (e.g., TCP keepalive as detected that the connection has dropped but the application hasn't noticed because it hasn't sent anything). Annoyingly, connection related metrics, including stuff we can't do at the application layer like some connection memory use, isn't available in host metrics because it's per-cgroup and not rolled up into parent cgroups unlike most other metrics.

travisdowns:
> Looking at AWS i7ie instances that have 0.5 Gbps (50 MB/s) per vcpu and 6 Gbps (600 MB/s) per disk, the bottleneck is no longer in either the CPU or disk, but the network.

Yeah, I guess this is the spiritual successor to i3en, but they've dropped the "n", so bandwidth is halved, and for a mix RW load I think disk throughput is up even more than advertised because these 2nd and 3rd generation nitro SSDs look like they are almost totally full-duplex compared to the old ones which weren't.

OTOH even with older instance types we could be network limited (usually network egress) in cases where workloads hit in cache so bytes out over the network didn't necessarily come from disk at all.

avikivity:
> > Looking at AWS i7ie instances that have 0.5 Gbps (50 MB/s) per vcpu and 6 Gbps (600 MB/s) per disk, the bottleneck is no longer in either the CPU or disk, but the network.
> 
> Yeah, I guess this is the spiritual successor to i3en, but they've dropped the "n", 

Ah, I thought they had to drop the "n" because the second "i" ate its budget.

> so bandwidth is halved, and for a mix RW load I think disk throughput is up even more than advertised because these 2nd and 3rd generation nitro SSDs look like they are almost totally full-duplex compared to the old ones which weren't.
> 
> OTOH even with older instance types we could be network limited (usually network egress) in cases where workloads hit in cache so bytes out over the network didn't necessarily come from disk at all.

Sure, you're moving bytes around while we're managing (usually small) cells, so you're more likely to hit the network limit first.

This strengthens the case for network scheduling.

avikivity:
> > We found that connection counts are better when associated with a scope. For example, client-server connection counts vs node-node connection counts.
> 
> Indeed, we have application level connection metrics as well. Still, the host metrics have proven useful there as a second "totally inclusive check" and because there are some states where the connection is still known by the kernel but not by the application or vice-versa (e.g., TCP keepalive as detected that the connection has dropped but the application hasn't noticed because it hasn't sent anything). Annoyingly, connection related metrics, including stuff we can't do at the application layer like some connection memory use, isn't available in host metrics because it's per-cgroup and not rolled up into parent cgroups unlike most other metrics.

Makes sense.

Did you report the Linux accounting problem upstream?

StephanDollberg:
I have pushed an update now that tracks the metrics per scheduling group.

travisdowns:
> Did you report the Linux accounting problem upstream?

I feel like it is "by design" for the way the networking namespaces works in cgroups so very unlikely to be fixed (I guess networking namespaces are not even really in a hierarchy). I did go to report it for node_exporter (which I think needs to do the hard work of aggregating these values for all namespace), but it was already reported and closed as won't fix. However in that latter case there were sort of two issues mixed together (whether node exporter should report "per pod/cgroup" values (definitely no), or whether NE should aggregate across all cgroups for values that are'nt rolled up to the top level.

So maybe it's worth another shot on the NE side.

include/seastar/net/api.hh
```diff
@@ -444,6 +444,28 @@ public:
...
+public:
+    statistics(uint64_t bytes_sent, uint64_t bytes_received)
+        : _bytes_sent(bytes_sent), _bytes_received(bytes_received) {}
+    uint64_t bytes_sent() const {
+        return _bytes_sent;
+    };
+    uint64_t bytes_received() const {
+        return _bytes_received;
+    };
+};
```
avikivity, line 459: Minor, but such pure data classes are better as structs with exposed data members and inline initializers for all data members. Less noisy.

src/net/native-stack-impl.hh
```diff
@@ -29,6 +29,20 @@ namespace seastar {
...
+namespace internal {
+
+struct native_stack_net_stats {
+#ifdef SEASTAR_BUILD_SHARED_LIBS
+    static thread_local std::array<unsigned, max_scheduling_groups()> bytes_sent;
+    static thread_local std::array<unsigned, max_scheduling_groups()> bytes_received;
+#else
+    static inline thread_local std::array<unsigned, max_scheduling_groups()> bytes_sent = {};
+    static inline thread_local std::array<unsigned, max_scheduling_groups()> bytes_received = {};
+#endif
```
avikivity, line 41: Doesn't the inline option work for both modes?

src/core/reactor.cc
```diff
@@ -960,6 +960,9 @@ reactor::task_queue::register_stats() {
         }, sm::description("Total amount in milliseconds we were in violation of the task quota"),
            {group_label}),
     });
+
+    register_net_metrics_for_scheduling_group(new_metrics, _id, group_label);
+
```
avikivity, line 965: This is the only call, so new groups won't have metrics.

avikivity: Wait that's completely wrong.

src/net/native-stack-impl.hh
travisdowns: > Doesn't the inline option work for both modes?

Well this is a questin I also had, but "no" according to this (where inline was removed from thread_locals in shared lib builds):

https://github.com/scylladb/seastar/pull/1494

However there it was just presented as fact that this doesn't work, but is that a bug in the compilers or an implementation limitation of `thread_local`? Even across shared libs non-thread_local globals will correctly resolve down to a single symbol across the proces at runtime, via weak symbol magic (same thing that makes static members inside templates work), but I guess that didn't extend to `thread_local` even though I think the C++ standard requires it.

avikivity: If static inline doesn't work, why not use static non-inline instead of splitting the implementation?

include/seastar/net/api.hh
travisdowns: We did investigate the struct option on our end, but it added 1 instruction to the hot path and we knew Avi-instruction-counter would be reviewing this we went with the separateed approach.

The extra instruction is probably only on x86 because at 16 bytes the struct can't be indexed with indexed addressing which maxes out at * 8, so it needs a separate `shl` do to the indexing.

src/net/native-stack-impl.hh
travisdowns: Same reason as #1494 does it (presumably): the `static inline` is more efficient since it allows direct access to the thread local from other TUs. If you use non-inline other TU's will need to indirect though an expensive general purpose `get_tls_blah` method.

travisdowns:
> This strengthens the case for network scheduling.

+1

src/net/native-stack-impl.hh
travisdowns: FTW we already went though the same confusion internally and concluded that the split approach in #1494 is the right thing per above: I can dig up the godbolt links from our internal discussion if you want.

src/core/reactor.cc
StephanDollberg: Sorry I don't follow. This gets called whenever a new group is added (or renamed). It works effectively the same way as scheduling groups get their metrics right now.

include/seastar/net/api.hh
avikivity: How can struct/class possibly matter here? The layout is the same.

src/net/native-stack-impl.hh
avikivity: I believe I understand the problem. Since we want to increment the variable in the header we need a declaration. If the declaration is not a definition the compiler will emit a TLS guard since it can't tell how it's initialized.

There is an alternative however: constinit. This way you can have a declaration-not-definition that tells the compiler it need not emit a TLS guard.

https://godbolt.org/z/vG54MG73f

avikivity: Not ideal though, don't know why: https://godbolt.org/z/9ndWPrjTd

avikivity: With local-exec it can be made ideal: https://godbolt.org/z/bW51qK6z9

avikivity: Perfection is achievable: https://godbolt.org/z/TPe7zjWfr

travisdowns: Good call on `constinit`.

> Not ideal though, don't know why: https://godbolt.org/z/9ndWPrjTd

Do you mean because of the extra indirection, i.e., the `movq good@gottpoff(%rip), %rax`? instruction?

> Perfection is achievable

I'm not following: how is inc_the_bestest better than `the_best`? `inc_the_bestest` is just fast because it's in the same TU so the compiler sees through to the same underlying definition, so it can do "direct" TLS access. In another TU it's just going to have `call the_bestest`. Or are you saying this would go in the header? Then it could have the same "bug" as #1494 as far as I can tell.

avikivity: > Good call on `constinit`.
> 
> > Not ideal though, don't know why: https://godbolt.org/z/9ndWPrjTd
> 
> Do you mean because of the extra indirection, i.e., the `movq good@gottpoff(%rip), %rax`? instruction?


Yes.

> 
> > Perfection is achievable
> 
> I'm not following: how is `inc_the_bestest` better than `the_best`? `inc_the_bestest` is just fast because it's in the same TU so the compiler sees through to the same underlying definition, so it can do "direct" TLS access. In another TU it's just going to have `call the_bestest`. Or are you saying this would go in the header? Then it could have the same "bug" as #1494 as far as I can tell.

You're right (I hadn't considered it).


However, the bug doesn't apply here. This native-stack-impl.hh is not a public header and would not be included from both the shared library and the main executable.

src/core/reactor.cc
avikivity: "that's completely wrong" referred to my previous comment

src/net/native-stack-impl.hh
travisdowns: > > Do you mean because of the extra indirection, i.e., the `movq good@gottpoff(%rip), %rax`? instruction?
> 
> Yes.

It the two indirections seem kind of fundamental here: the first is needed to load the TLS offset for the variable, which is a runtime value (cannot be resolved by the linker) as it depends on the libraries loaded, and the second is needed to access that offset from the TLS block pointer in `fs`.

So it is similar to a (non-TLS) global variable access with `-fPIC` which similarly does too indirections, the first through the GOT. The difference being in non-PIC mode this can be folded into on `rip`-relative access since the offset is just a relocation that can be computed by the linker (but in the TLS case it is not).

So the only "single instruction" case for TLS is where the compiler can see the definition of the TLS object _and_ we are in no-PIC or PIE mode, since then the TLS offset is statically computable.

travisdowns: > However, the bug doesn't apply here. This native-stack-impl.hh is not a public header and would not be included from both the shared library and the main executable.

My knowledge of how the native stack works borders on zero (my main expertise is disabling DPDK entirely so we can build faster), but in many demo programs this header is indeed included, e.g., https://github.com/scylladb/seastar/blob/master/demos/echo_demo.cc#L27 in echo_demo.

avikivity: That's a bug. native-stack-impl.hh is under src/ and is not visible to user applications. The last time echo_demo was run is likely the day it was written.

avikivity: > > > Do you mean because of the extra indirection, i.e., the `movq good@gottpoff(%rip), %rax`? instruction?
> > 
> > 
> > Yes.
> 
> ISTM the two indirections seem kind of fundamental here: the first is needed to load the TLS offset for the variable, which is a runtime value (cannot be resolved by the linker) as it depends on the libraries loaded, and the second is needed to access that offset from the TLS block pointer in `fs`.
> 
> So it is similar to a (non-TLS) global variable access with `-fPIC` which similarly does the double indirection, the first through the GOT. The difference being in non-PIC mode this can be folded into on `rip`-relative access since the offset is just a relocation that can be computed by the linker (but in the TLS case it is not).
> 
> So the only "single instruction" case for TLS is where the compiler can see the definition of the TLS object _and_ we are in no-PIC or PIE mode, since then the TLS offset is statically computable.

This happens to be the case I'm most interested in.

travisdowns: > That's a bug. native-stack-impl.hh is under src/ and is not visible to user applications. The last time echo_demo was run is likely the day it was written.

This was my mistake, it includes `native-stack.hh` *not* `native-stack-impl.hh`.

> This happens to be the case I'm most interested in.

Right but in this case you don't need any tricks with extern/tls models/static locals, the compiler already produces good code as it can see the declaration.

include/seastar/net/api.hh
travisdowns: Sorry @avikivity I didn't look carefully enough, I though you were referring to the separate read/write TLS variables, which are not bundled into a struct, not this one which is used to expose the metrics.

src/net/native-stack-impl.hh
travisdowns: There is a twist there though. Try as I might, I cannot reproduce the original problem described in #1494 .

Here is my attempt: https://godbolt.org/z/PT1h7r5fr

I tried many compiler options and togging on and off flags which might be relevant. The final result always DTRT, i.e., both shared libs saw the same underlying `g_this_shard_id` object. @tchaikov do you remember specifics of the problem and how to reproduce it? Did it happen only with C++ modules enabled?

include/seastar/net/api.hh
StephanDollberg: changed

src/net/native-stack-impl.hh
StephanDollberg: So confirm you think we should do:

```
inline uint64_t& bytes_sent(unsigned sg_id) {
    static thread_local std::array<uint64_t, max_scheduling_groups()> bytes_sent;
    return bytes_sent[sg_id];
}
```

travisdowns: Just to clarify, I think we agreed above that static-thread-local-in-a-function isn't any better than a global in terms of (a) removing ifdefs or (b) performance. It behaves identically to an "inline global" in those respects.

So you want that for scoping reasons?

avikivity: I just want the #ifdef gone and not to compromise on static library performance.

I think that the problem with global inline only happens when it is visible to both the shared library and the main executable, which isn't the case. So I think we can go back to global inline.

travisdowns: > I just want the #ifdef gone and not to compromise on static library performance.

As discussed above, the function-local static provide no benefit in that case. It also suffers from the "duplicate copies in .so" bug as evidenced by #1494 which removed exactly that pattern, if shared across TUs and if there is no need to share across TUs we can just put the global in the .cc file no problem.

> So I think we can go back to global inline.

Yes, agree for this case (not in public includes).

StephanDollberg: I have pushed "global inline" now. Let me know if I have misunderstood what you wanted.

tchaikov: > There is a twist there though. Try as I might, I cannot reproduce the original problem described in #1494 .
> 
> Here is my attempt: https://godbolt.org/z/PT1h7r5fr
> 
> I tried many compiler options and togging on and off flags which might be relevant. The final result always DTRT, i.e., both shared libs saw the same underlying `g_this_shard_id` object. @tchaikov do you remember specifics of the problem and how to reproduce it? Did it happen only with C++ modules enabled?

Your reproducer looks valid. As far as I recall, I made that change after encountering a failure similar to the one mentioned here: [#12801](https://github.com/scylladb/scylladb/pull/12801#issuecomment-1425614559). However, I’m currently on vacation and will look into reproducing the issue once I’m back next week.

travisdowns: @tchaikov one thing I didn't try was having the variable used by 1 shard lib + main executable. The repro instead accesses it from two different shared libs. Since main executable is "special" for TLS maybe that matters.

StephanDollberg: ping @avikivity

PR #2633, opened 2025-02-03
core: function_traits noexcept specializations

Comrade88:
starting from 17th standard noexcept-specification is a part of the function type. function_traits doesn't have specializations
for noexcept function types thus cannot resolve the traits.
adding noexcept specializations allows function_traits for a larger class of function types fi noexcept future exception continuations. seastar supports 20th standard and newer thus there is no need to worry about compatibility.

### DIFF
--- a/include/seastar/core/function_traits.hh
+++ b/include/seastar/core/function_traits.hh
@@ -57,6 +57,18 @@ template <typename T, typename Ret, typename... Args>
 struct function_traits<Ret(T::*)(Args...) const> : public function_traits<Ret(Args...)>
 {};
 
+template<typename Ret, typename... Args>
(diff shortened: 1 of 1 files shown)


include/seastar/core/function_traits.hh
```diff
@@ -57,6 +57,21 @@ template <typename T, typename Ret, typename... Args>
 struct function_traits<Ret(T::*)(Args...) const> : public function_traits<Ret(Args...)>
 {};
 
+#if __cplusplus >= 201703L
```
avikivity, line 60: You can drop this #ifdef, Seastar support C++20 and C++23 only.

Comrade88: done

Comrade88: @avikivity , is anything else required here?

Issue #2640, opened 2025-02-12
Recent version of c-ares can cause use-after free in dns.cc


elcallio:
Recent c-ares (fedora 41 version), will do checks, including closing broken connections etc during FD processing, 
_not only for the fd(s) involved in the processing_, unless explicitly disabled. 

This can break our dns.cc state machine, as we might process a fd, but get a recursive callback which will modify the
collection we are iterating over (_sockets) such that our next iterator entry becomes invalid/points to free:d data.

Fix is relatively simple, use the now encouraged (newish) API, and turn off this background processing until we are out of the iteration loop.

PR #2641, opened 2025-02-12
dns: Fix event processing to work safely with recent c-ares

elcallio:
Recent c-ares (fedora 41 version), will do checks, including closing broken connections etc during FD processing, not only for the fd(s) involved in the processing, unless explicitly disabled.

This can break our dns.cc state machine, as we might process a fd, but get a recursive callback which will modify the collection we are iterating over (_sockets) such that our next iterator entry becomes invalid/points to free:d data.

Fix is relatively simple, use the now encouraged (newish) API, and turn off this background processing until we are out of the iteration loop.

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -597,11 +597,14 @@ dns_resolver::impl::end_call() {
     }
 }
 
+#define USE_CARES_EVENTFD (ARES_VERSION >= 0x012200)
+
 void
 dns_resolver::impl::poll_sockets() {
     dns_log.trace("Poll sockets");
 
     bool processed = false;
+
     for (;;) {
         // Retrieve the set of file descriptors that the library wants us to monitor.
         fd_set readers, writers;
@@ -614,9 +617,17 @@ dns_resolver::impl::poll_sockets() {
             break;
         }
 
+#if USE_CARES_EVENTFD
+        // avoid allocations on every poll. the ares_process_fds will not reenter this,
+        // as any read/write/close callbacks will either complete or create a pending
+        // future which cannot execute until the call returns.
+        // I.e. we cannot get here while this call is running, so safe to make this
+        // thread static
+        static thread_local ares_fd_events_t events[FD_SETSIZE];
+#endif
         int processed_fds = 0;
-        for (auto it = _sockets.begin(); it != _sockets.end();) {
-            auto& [fd, e] = *it++;
+
+        for (auto& [fd, e] : _sockets) {
             bool read_monitor = FD_ISSET(fd, &readers);
             bool write_monitor = FD_ISSET(fd, &writers);
             bool read_avail = e.avail & POLLIN;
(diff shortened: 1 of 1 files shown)


avikivity:
So much red.

How new is the new API? If it's three years old or older, I don't mind switching to it without backward compatibility.

tchaikov:
the new API was introduced by https://github.com/c-ares/c-ares/commit/3b80588384060014acabac057a969aa954eca2af, which is included by [v1.34.0](https://github.com/c-ares/c-ares/releases/tag/v1.34.0) and up. and v1.34 was released on Oct 9 2024 -- less than one year ago. the reason why the tests failed was that we are using ubuntu 24.04 for testing. and ubuntu 24.04 comes with libc-ares 1.27.0, see https://packages.ubuntu.com/noble/libc-ares-dev .

so we might need to conditionalize this fix with something like
```c++
#if ARES_VERSION >= 0x012200
/// fix here
#endif
```
in which
```
In [3]: hex(1 << 16 | 34 << 8 | 0)
Out[3]: '0x12200'
```

src/net/dns.cc
```diff
@@ -628,21 +627,39 @@ dns_resolver::impl::poll_sockets() {
...
+            if (read_monitor && read_avail) {
+                events[nevents].fd = fd;
+                events[nevents].events |= ARES_FD_EVENT_READ;
+                ++nevents;
+            }
+            if (write_monitor && write_avail) {
+                events[nevents].fd = fd;
+                events[nevents].events |= ARES_FD_EVENT_WRITE;
+                ++nevents;
+            }
```
tchaikov, line 651: probably we can use a single `ares_fd_events_t` here, like
```c++
            ares_fd_events_t event = {fd, ARES_FD_EVENT_NONE};
            if (read_monitor && read_avail) {
                event.events |= ARES_FD_EVENT_READ;
            }
            if (write_monitor && write_avail) {
                event.events |= ARES_FD_EVENT_WRITE;
            }
            if (events.events) {
                ares_process_fds(_channel, &event, 1, ARES_PROCESS_FLAG_SKIP_NON_FD);
                ++processed_fds;
            }
```

elcallio: Yes, but I prefer to copy the code in ares_process_fd, just to keep it consistent.

tchaikov: fair enough..

tchaikov:
> 
> also, IIUC, this change fixes #2640, so probably you can add `Fixes #2640` at the end of the commit message ? so that the commit and this pull request can be associated with #2640 .

could you please reconsider this suggestion?

elcallio:
Of course. I just got distracted before I could send it. :-/

tchaikov (approved):
lgtm

avikivity:
For my understanding, what is the background processing you skip? The cleanup/timeout stuff with SKIP_NON_FDS?

elcallio:
`ares_check_cleanup_conns` - going through all fd:s and checking error state, closing them. Called from `ares_process_fds` unless the `ARES_PROCESS_FLAG_SKIP_NON_FD` is set.

Honestly, reading through the source code for current and old version though, I am not sure that the problem does not exist for older versions as well, since the corresponding check is done in the old code as well (albeit a bit different). We might just be very lucky in the case of the old code, and should maybe patch for that as well.

elcallio:
I rewrote the fix to use batch processing of events instead (both versions). This is encouraged in the new API anyway, so it seems reasonable. Also makes sure the problem is not hiding in the old API world.

tchaikov:
i will take a look early tomorrow.

src/net/dns.cc
```diff
@@ -628,20 +637,67 @@ dns_resolver::impl::poll_sockets() {
...
+#if USE_CARES_EVENTFD
+            if (events[processed_fds].events) {
+                ++processed_fds;
+            }
+#else
+            if (FD_ISSET(fd, &readers) || FD_ISSET(fd, &writers)) {
                 ++processed_fds;
             }
+#endif
```
tchaikov, line 683: The current implementation uses heavily segmented conditional logic to handle
different c-ares versions, which makes the code harder to follow. How about using a more straightforward approach by duplicating some shared logic in favor of clearer, more maintainable code? like

```c++
#if USE_CARES_EVENTFD
            events[processed_fds] = {0,};
            // Update read/write state based on our poll info
            if (read_monitor && read_avail) {
                events[processed_fds].fd = fd;
                events[processed_fds].events |= ARES_FD_EVENT_READ;
            }
            if (write_monitor && write_avail) {
                events[processed_fds].fd = fd;
                events[processed_fds].events |= ARES_FD_EVENT_WRITE;
            }
            if (events[processed_fds].events) {
                ++processed_fds;
            }
#else
            FD_CLR(fd, &readers);
            FD_CLR(fd, &writers);
            // Update read/write state based on our poll info
            if (read_monitor && read_avail) {
                FD_SET(fd, &readers);
            }
            if (write_monitor && write_avail) {
                FD_SET(fd, &writers);
            }
            if (FD_ISSET(fd, &readers) || FD_ISSET(fd, &writers)) {
                ++processed_fds;
            }
#endif
```

tchaikov (approved):
lgtm

elcallio:
ping?

PR #2642, opened 2025-02-12
reactor: deprecate at_destroy(), at_exit()

avikivity:
at_destroy() tasks are unordered with respect to each other, so are not very useful. They also consume a scheduling group. at_exit() functions are similar.

Since they are public, we can't remove them outright, but prepare by deprecating the public interface. Internal users are converted to a private interface.

We add a section to the tutorial to explain how modern initialization and cleanup are done.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -129,6 +129,9 @@ size_t scheduling_group_count();
 
 void increase_thrown_exceptions_counter() noexcept;
 
+template <typename Func>
+void at_destroy(Func&& func);
+
 }
 
 class io_queue;
@@ -162,6 +165,8 @@ private:
     class io_queue_submission_pollfn;
     class syscall_pollfn;
     class execution_stage_pollfn;
+    template <typename Func>
+    friend void internal::at_destroy(Func&&);
     friend class manual_clock;
     friend class file_data_source_impl; // for fstream statistics
     friend class internal::reactor_stall_sampler;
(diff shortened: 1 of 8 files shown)


include/seastar/core/reactor.hh
```diff
@@ -526,10 +531,16 @@ public:
     void at_exit(noncopyable_function<future<> ()> func);
 
     template <typename Func>
+    [[deprecated("Use app_template::run() for orderly shutdown")]]
```
xemul, line 534: Why `app_template::run()`, shouldn't it be `engine().at_axit()` instead?

avikivity: This patch is deprecating reactor::at_exit().

app_template::run offers a way to destroy stuff in reverse order of construction, unlike at_exit.

xemul: > This patch is deprecating reactor::at_exit().

This patch is deprecating reactor::at_destroy(), reactor::at_exit() remains intact

> app_template::run offers a way to destroy stuff in reverse order of construction, unlike at_exit.

Can you point to corresponding demo/doc/test that shows how to do it? (here, not in deprecation message)

avikivity: > > This patch is deprecating reactor::at_exit().
> 
> This patch is deprecating reactor::at_destroy(), reactor::at_exit() remains intact

Right. I don't even know what the differences are. Should we deprecate them both?

> 
> > app_template::run offers a way to destroy stuff in reverse order of construction, unlike at_exit.
> 
> Can you point to corresponding demo/doc/test that shows how to do it? (here, not in deprecation message)

```c++
    sharded<foo> thing;
    thing.start().get();
    auto undo_thing = defer([] { thing.stop().get(); });
```

xemul: > Right. I don't even know what the differences are. 

The at_exit() tasks are run while reactor is being stopped, while at_destroy() ones -- after it is stopped.

Quoting 6ad9114c0b60df2071b211669697269472036445
>     reactor: add at_destroy() function to the reactor and use it
>        
>     Unfortunately at_exit() cannot be used to delete objects since when
>     it runs the reactor is still active and deleted object may still been used.
>     We need another API that runs its task after reactor is already stopped.
>     at_destroy() will be such api.

> Should we deprecate them both?

Less API the better, I second that

>     sharded<foo> thing;
>     thing.start().get();
>     auto undo_thing = defer([] { thing.stop().get(); });

Ah. Then I double my seconding of at_exit() deprecation. There are traces of at_exit() usage in Scylla main, but it anyway switched to this approach.

However the "Use app_template::run() for orderly shutdown" message is confusing :( How about "Use defer-s in app_template::run() async lambda for orderly shutdown" ? And the example above is definitely worth being added to the patch comment

avikivity: Well, if the message confused you, anyone would be confused, I'll improve it.

avikivity:
v2:
 - rebased
 - also deprecate at_exit (new patch)
 - add tutorial section
 - improved deprecation message
 - add app_template alternative to function documentation

avikivity:
v2:
  - actually add deprecated attribute to at_exit(), don't just comment about it
  - reimplement so at_exit implementation can be called before engine() is ready

PR #2644, opened 2025-02-17
testing: prevent test scheduling after reactor exit

### LINKED ISSUE #2635: When --help flag is passed to a test the program hangs/never exits
This PR breaks `--help` when it is passed to a test https://github.com/scylladb/seastar/pull/2213

I.e.

```
 → tests/unit/rpc_test -- --help
Running 35 test cases...
App options:
  -h [ --help ]                         show help message
  --help-seastar                        show help message about seastar options
  --help-loggers                        print a list of logger names and exit
  --random-seed arg                     Random number generator seed
  --fail-on-abandoned-failed-futures arg (=1)
                                        Fail the test if there are any
                                        abandoned failed futures

^C^C^C^C^C^C
```

The app just hangs instead of exiting.

My guess is that the problem is somewhere around this code block https://github.com/scylladb/seastar/blob/b63b02a1f674a96365ec688f257294cebac3b865/src/testing/test_runner.cc#L77

tchaikov:
Previously, when handling "--help" (introduced in b8a13bebe4), we returned status code 0 but continued scheduling test tasks to the Seastar reactor. This caused test applications to hang since the tasks expected the exchanger 'e' to be available after reactor exit.

Fix this by using the "_done" flag to track reactor state:
- Set "_done" when Seastar application exits
- Skip task scheduling and exchanger wait if "_done" is set
- Change test_runner::start_thread() to return bool indicating successful engine startup instead of exit code (exit code now stored in _exit_code member)

This fixes the regression where "--help" would cause test applications to hang indefinitely.

Fixes #2635

### DIFF
--- a/src/testing/test_runner.cc
+++ b/src/testing/test_runner.cc
@@ -63,8 +63,8 @@ test_runner::start(int ac, char** av) {
     return true;
 }
 
-int test_runner::start_thread(int ac, char** av) {
-    auto init_outcome = std::make_shared<exchanger<int>>();
+bool test_runner::start_thread(int ac, char** av) {
+    auto init_outcome = std::make_shared<exchanger<bool>>();
 
     namespace bpo = boost::program_options;
     _thread = std::make_unique<posix_thread>([this, ac, av, init_outcome]() mutable {
@@ -75,7 +75,7 @@ int test_runner::start_thread(int ac, char** av) {
         // We guarantee that only one thread is running.
         // We only read this after that one thread is joined, so this is safe.
         _exit_code = app.run(ac, av, [this, &app, init_outcome = init_outcome.get()] {
-            init_outcome->give(0);
+            init_outcome->give(true);
             auto init = [&app] {
                 auto conf_seed = app.configuration()["random-seed"];
                 auto seed = conf_seed.empty() ? std::random_device()():  conf_seed.as<unsigned>();
@@ -108,7 +108,7 @@ int test_runner::start_thread(int ac, char** av) {
                 return 0;
             });
         });
-        init_outcome->give(_exit_code);
+        init_outcome->give(false);
     });
 
     return init_outcome->take();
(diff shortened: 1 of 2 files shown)


src/testing/test_runner.cc
```diff
@@ -130,6 +131,11 @@ test_runner::run_sync(std::function<future<>()> task) {
         // it.
         return;
     }
+    if (_done) {
```
xemul, line 134: Can it happen that _done is false, but _exit_code is non-zero?
In other words -- is the above `if (_exit_code != 0)` still required?

tchaikov: ahh, you're right. with my change, if `_done` is true, it implies `app.run()` has returned -- it's an early return, and the return code is either `0` or a non-zero status code. let me update the change to further simplify this implementation.

tchaikov:
added a cleanup, marking this change a draft. will test it tomorrow.

tchaikov:
v2:

- Instead of setting `_done` in `test_runner::start_thread()` member function, change test_runner::start_thread() to return bool indicating successful engine startup instead of exit code (exit code now stored in _exit_code member), and set and `_done` in `test_runner::run_sync()`. simpler this way. and it is more obvious that we early return when `_done` is set.

tchaikov:
tested with
```console
kefu@mega:~/dev/seastar$ build/debug/tests/unit/queue_test -- --help
Running 3 test cases...
WARNING: debug mode. Not for benchmarking or production
App options:
  -h [ --help ]                         show help message
  --help-seastar                        show help message about seastar options
  --help-loggers                        print a list of logger names and exit
  --random-seed arg                     Random number generator seed
  --fail-on-abandoned-failed-futures arg (=1)
                                        Fail the test if there are any 
                                        abandoned failed futures


*** No errors detected
kefu@mega:~/dev/seastar$ 
```

PR #2645, opened 2025-02-17
memory: remove libnuma dependency

rockwotj:
Seastar is currently pulling in all of libnuma to just invoke a single
syscall. Additionally since libnuma being LGPL licensed, it's simpler
to not include it as a dependency. Replace the call in numaif.h with
a direct syscall.

I'm assuming that once libnuma was depended on because it was a
dependency of hwloc, however, hwloc version 2 removed the dependency
on libnuma, so Seastar can too:
https://github.com/open-mpi/hwloc/blob/263908a2c1f21c0e221a8d1f6472daf3a1fc07b9/NEWS#L291

### DIFF
--- a/src/core/memory.cc
+++ b/src/core/memory.cc
@@ -101,10 +101,9 @@ module;
 #include <utility>
 #include <boost/intrusive/list.hpp>
 #include <sys/mman.h>
+#include <sys/syscall.h>
+#include <linux/mempolicy.h>
 
-#ifdef SEASTAR_HAVE_NUMA
-#include <numaif.h>
-#endif
 #endif // !defined(SEASTAR_DEFAULT_ALLOCATOR)
 
 #ifdef SEASTAR_MODULE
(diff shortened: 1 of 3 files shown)


rockwotj:
ugh I missed that dpdk required this, will re-add it to the build for dpdk only

src/core/memory.cc
```diff
@@ -1832,6 +1830,23 @@ void configure_minimal() {
...
+  return syscall(
+    SYS_mbind,
+    reinterpret_cast<long>(addr),
+    len,
+    mode,
+    reinterpret_cast<long>(nodemask),
+    maxnode,
+    flags
+  );
+}
```
avikivity, line 1848: Our coding style uses 4-space indents, not two.

I don't think the reinterpret_casts are required.

src/core/memory.cc
```diff
@@ -1865,16 +1880,18 @@ configure(std::vector<resource::memory> m, bool mbind,
...
     for (auto&& x : m) {
         unsigned long nodemask = 1UL << x.nodeid;
         if (mbind) {
             auto start = get_cpu_mem().mem() + pos;
-            auto r = ::mbind(start, x.bytes,
-                            MPOL_PREFERRED,
-                            &nodemask, std::numeric_limits<unsigned long>::digits,
-                            MPOL_MF_MOVE);
+            static constexpr int mpol_preferred = 1;
+            static constexpr unsigned int mpol_mf_move = 2;
```
avikivity, line 1889: #include \<linux/mempolicy>?

rockwotj: Ah yes thanks! Fixed.

src/core/memory.cc
rockwotj: Corrected. Thank you!

PR #2646, opened 2025-02-17
treewide: replace assert with SEASTAR_ASSERT

rockwotj:
This prevents the need to undefined NDEBUG in seastar source and
prevents headers from conditionally excluding these checks.

See discussion here: https://github.com/scylladb/seastar/discussions/2539

### DIFF
--- a/include/seastar/util/assert.hh
+++ b/include/seastar/util/assert.hh
@@ -16,35 +16,21 @@
  * under the License.
  */
 /*
- * Copyright (C) 2018 ScyllaDB
+ * Copyright (C) 2025 Redpanda Data.
  */
 
 #pragma once
 
-#ifndef SEASTAR_MODULE
-#include <chrono>
-#include <time.h>
-#endif
-#include <seastar/util/assert.hh>
-
-namespace seastar {
-
-class thread_cputime_clock {
-public:
-    using rep = int64_t;
-    using period = std::chrono::nanoseconds::period;
-    using duration = std::chrono::duration<rep, period>;
-    using time_point = std::chrono::time_point<thread_cputime_clock, duration>;
-public:
-    static time_point now() {
-        using namespace std::chrono_literals;
-
-        struct timespec tp;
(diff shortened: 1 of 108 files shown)


rockwotj:
Force pushes: fix C++ modules with SEASTAR_ASSERT

include/seastar/util/assert.hh
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2025 ScyllaDB.
```
avikivity, line 19: You should put your own (or your employer's) copyright here.

include/seastar/util/assert.hh
```diff
@@ -0,0 +1,34 @@
...
+
+/// Like assert(), but independent of NDEBUG. Active in all build modes.
+#define SEASTAR_ASSERT(x)                                                      \
+  do {                                                                         \
+    if (!(x)) [[unlikely]] {                                                   \
+      printf("%s:%u: %s: Assertion `%s` failed.\n", __FILE__, __LINE__,        \
+             __PRETTY_FUNCTION__, #x);                                         \
+      __builtin_trap();                                                        \
+    }                                                                          \
+  } while (0)
```
avikivity, line 34: 4-space indents.

Please call a [[noreturn]] function (you can add it in reactor.cc, no need for a new source file) so that we don't have to bring in \<cstdio> for everyone.

include/seastar/util/assert.hh
rockwotj: Fixed.

include/seastar/util/assert.hh
rockwotj: Done. I added in `log.cc` to keep it in the `util` directory.

PR #2647, opened 2025-02-17
Support chunked_vector and other json/httpd changes

travisdowns:
This series came out of an effort to squash large allocations in a json httpd endpoint in redpanda: the basic idea is to optionally use chunked_fifo instead of std::vector as the underlying type in the json2code generation.

Some other bugs I ran into along the the way are also fixed, and even code that doesn't use chunked_fifo should get some benefits from the addition of move support on the json response hot path.

Test cases added where I thought it made sense.

### DIFF
--- a/include/seastar/core/chunked_fifo.hh
+++ b/include/seastar/core/chunked_fifo.hh
@@ -24,6 +24,7 @@
 #ifndef SEASTAR_MODULE
 #include <algorithm>
 #include <cassert>
+#include <iterator>
 #include <type_traits>
 #include <seastar/util/modules.hh>
 #endif
@@ -176,10 +177,11 @@ public:
 public:
     chunked_fifo() noexcept = default;
     chunked_fifo(chunked_fifo&& x) noexcept;
-    chunked_fifo(const chunked_fifo& X) = delete;
+    chunked_fifo(const chunked_fifo&);
     ~chunked_fifo();
-    chunked_fifo& operator=(const chunked_fifo&) = delete;
+    chunked_fifo& operator=(const chunked_fifo&);
     chunked_fifo& operator=(chunked_fifo&&) noexcept;
+    inline bool operator==(const chunked_fifo& rhs) const;
     inline void push_back(const T& data);
     inline void push_back(T&& data);
     T& back() noexcept;
@@ -294,6 +296,25 @@ chunked_fifo<T, items_per_chunk>::chunked_fifo(chunked_fifo&& x) noexcept
     x._nchunks = 0;
     x._free_chunks = nullptr;
     x._nfree_chunks = 0;
+}
+
+template <typename T, size_t items_per_chunk>
+inline
+chunked_fifo<T, items_per_chunk>::chunked_fifo(const chunked_fifo& rhs)
+        : chunked_fifo() {
+    std::copy_n(rhs.begin(), rhs.size(), std::back_inserter(*this));
+}
+
+template <typename T, size_t items_per_chunk>
+inline
+chunked_fifo<T, items_per_chunk>&
+chunked_fifo<T, items_per_chunk>::operator=(const chunked_fifo& rhs) {
+    if (&rhs != this) {
+        clear();
+        std::copy_n(rhs.begin(), rhs.size(), std::back_inserter(*this));
+        shrink_to_fit();
+    }
+    return *this;
 }
 
 template <typename T, size_t items_per_chunk>
@@ -453,6 +474,11 @@ chunked_fifo<T, items_per_chunk>::emplace_back(Args&&... args) {
         throw;
     }
     ++_back_chunk->end;
+}
+
+template <typename T, size_t items_per_chunk>
+inline bool chunked_fifo<T, items_per_chunk>::operator==(const chunked_fifo& rhs) const {
+    return size() == rhs.size() && std::equal(begin(), end(), rhs.begin());
 }
 
 template <typename T, size_t items_per_chunk>
--- a/include/seastar/json/json_elements.hh
+++ b/include/seastar/json/json_elements.hh
@@ -368,8 +368,8 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
 template<class T>
 std::function<future<>(output_stream<char>&&)> stream_object(T val) {
     return [val = std::move(val)](output_stream<char>&& s) mutable {
-        return do_with(output_stream<char>(std::move(s)), T(std::move(val)), [](output_stream<char>& s, const T& val){
-            return formatter::write(s, val).finally([&s] {
+        return do_with(output_stream<char>(std::move(s)), T(std::move(val)), [](output_stream<char>& s, T& val){
+            return formatter::write(s, std::move(val)).finally([&s] {
                 return s.close();
             });
         });
--- a/include/seastar/json/json_elements.hh
+++ b/include/seastar/json/json_elements.hh
@@ -24,8 +24,6 @@
 #ifndef SEASTAR_MODULE
 #include <string>
 #include <vector>
-#include <time.h>
-#include <sstream>
 #endif
 
 #include <seastar/core/do_with.hh>
--- a/include/seastar/json/json_elements.hh
+++ b/include/seastar/json/json_elements.hh
@@ -26,11 +26,12 @@
 #include <vector>
 #endif
 
+#include <seastar/core/chunked_fifo.hh>
 #include <seastar/core/do_with.hh>
-#include <seastar/core/loop.hh>
-#include <seastar/json/formatter.hh>
-#include <seastar/core/sstring.hh>
 #include <seastar/core/iostream.hh>
+#include <seastar/core/loop.hh>
+#include <seastar/core/sstring.hh>
+#include <seastar/json/formatter.hh>
 #include <seastar/util/modules.hh>
 
 namespace seastar {
@@ -149,13 +150,15 @@ private:
 };
 
 /**
- * json_list is based on std vector implementation.
+ * json_list_template is an array type based on a
+ * container type passed as a template parameter, as we want to
+ * have flavors based on both vector and chunked_fifo.
  *
  * When values are added with push it is set the "set" flag to true
  * hence will be included in the parsed object
  */
-template<class T>
-class json_list : public json_base_element {
+template <class T, class Container>
+class json_list_template : public json_base_element {
 public:
 
     /**
@@ -186,7 +189,7 @@ public:
      * iteration and that it's elements can be assigned to the list elements
      */
     template<class C>
-    json_list& operator=(const C& list) {
+    json_list_template& operator=(const C& list) {
         _elements.clear();
         for  (auto i : list) {
             push(i);
@@ -196,8 +199,15 @@ public:
     virtual future<> write(output_stream<char>& s) const override {
         return formatter::write(s, _elements);
     }
-    std::vector<T> _elements;
+
+    Container _elements;
 };
+
+template <typename T>
+using json_list = json_list_template<T, std::vector<T>>;
+
+template <typename T>
+using json_chunked_list = json_list_template<T, seastar::chunked_fifo<T>>;
 
 class jsonable {
 public:
--- a/tests/unit/api.json
+++ b/tests/unit/api.json
@@ -82,6 +82,18 @@
             "VAL2",
             "VAL3"
           ]
+        },
+        "array_var": {
+          "type": "array",
+          "items": {
+            "type": "int"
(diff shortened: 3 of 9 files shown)


tchaikov (approved):
lgtm.

tests/unit/api.json
```diff
@@ -47,6 +47,18 @@
                 "VAL2",
                 "VAL3"
               ]
+            },
+            {
+              "name": "stream_enum",
+              "description": "Whether to return the response as a stream_object",
+              "required": true,
+              "allowMultiple": false,
+              "type": "string",
```
tchaikov, line 56: i was about to suggest to use "boolean". but ironically, it turns out that the "string" type with enum contraints is indeed simpler.

tchaikov:
@amnonh hi Amnon, could you please take a look as well?

tests/unit/api.json
travisdowns: Happy to use boolean if you prefer. I actually used enum since I did a quick check and dind't see any existing uses of boolean but looking at seastar-json2code.py a `boolean` type is indeed supported.

include/seastar/core/chunked_fifo.hh
```diff
@@ -296,6 +301,18 @@ chunked_fifo<T, items_per_chunk>::chunked_fifo(chunked_fifo&& x) noexcept
...
 }
 
+template <typename T, size_t items_per_chunk>
+inline
+chunked_fifo<T, items_per_chunk>&
+chunked_fifo<T, items_per_chunk>::operator=(const chunked_fifo& rhs) {
+    if (&rhs != this) {
+        clear();
+        std::copy_n(rhs.begin(), rhs.size(), std::back_inserter(*this));
+        shrink_to_fit();
```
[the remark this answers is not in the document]
nyh: I see. No, I don't think that super-optimizing this code is very important. By definition, making a copy of a whole vector is already non-optimal...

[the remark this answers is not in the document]
travisdowns: > By definition, making a copy of a whole vector is already non-optimal...

Indeed, and here we are adding the copy assignment operator only because that's a requirement of the json elements in seastar: we auto-generate a copy constructor (which internally uses assignment) so we require all types used as json elements to have at least an assignment operator.

include/seastar/core/chunked_fifo.hh
```diff
@@ -178,8 +179,9 @@ public:
     chunked_fifo(chunked_fifo&& x) noexcept;
     chunked_fifo(const chunked_fifo& X) = delete;
     ~chunked_fifo();
-    chunked_fifo& operator=(const chunked_fifo&) = delete;
+    chunked_fifo& operator=(const chunked_fifo&);
```
avikivity, line 182: Would be more symmetrical to require x = y.copy(), no?

include/seastar/core/chunked_fifo.hh
```diff
@@ -190,6 +192,9 @@ public:
     inline void pop_front() noexcept;
     inline bool empty() const noexcept;
     inline size_t size() const noexcept;
+    // Return a new chunked_fifo which is a copy of this one, which
+    // is useful as this class does not allow copy creation or assignment.
```
avikivity, line 196: Previous change allows assignment.

tests/unit/api.json
avikivity: I think it's less surprising to have a boolean here.

avikivity:
I'm now regretting my decision not to allow random containers like chunked_vector into Seastar (or maybe, the decision to allow the json stuff in). It's an important building block for intermediate layers like this json stuff.

include/seastar/core/chunked_fifo.hh
travisdowns: > Would be more symmetrical to require x = y.copy(), no?

Yeah, but ... the main problem I'm trying to solve is that json2code generates a "copy constructor" which expects all the elements of the object to be copyable, here's an example from api.json in this repo:

```
    my_object(const my_object& e) {
        register_params();
        var1 = e.var1;
        var2 = e.var2;
        enum_var = e.enum_var;
    }
```

As shown, this copy constructor does element-wise assignment (not sure why it doesn't use member init in order to use copy-ctor instead), so that's why I'm adding the assignment operator: to make the above compile with chunked_fifo as an element.

Actually `copy()` is not needed here at all, it is vestigial from a different approach I tried first, though we do use this `copy()` pattern in Redpanda, exactly as you wanted: the container class declares move-assignment, but not move-copy, so you'd use `x = y.copy()` if you wanted to "force" a copy-assignment.

I'm happy to whatever here:

1) This change as is
2) This change but remove any traces of copy()
3) Add full copy support to `chunked_fifo` and update the code generation to use the copy-ctor
4) Remove copy-assignment, but leave `copy()` and update the code generation to call `copy()` explicitly (I had a change along these lines originally)

Please advise.

I agree that containers without copy are kind of anti-C++ though I know having them as saved us many unecessary copies in Redpanda as it forces you to get move working everywhere (with `copy()` as an escape hatch).

avikivity: There is also 5. change json2code to call a helper template function, which defaults to a regular copy, but which we can override to do something else for specific types.

But I think supporting the copy constructor is the path of least friction (though it opens the door to bad surprises).

We handle unexpected copies by having our small-scale performance tests monitor allocation count (and task count, and instruction count) per op and watching for changes.

include/seastar/json/json_elements.hh
```diff
@@ -186,19 +189,27 @@ public:
...
+    json_list_template& operator=(const C& list) {
         _elements.clear();
         for  (auto i : list) {
             push(i);
         }
         return *this;
     }
     virtual future<> write(output_stream<char>& s) const override {
-        return formatter::write(s, _elements);
+        auto t = const_cast<json_list_template<T, Container> *>(this);
```
nvartolomei, line 200: I'm puzzled by this cast and the intention behind it.

1. Why are we moving data from a const object?
2. If an optimization is possible/desirable why not add a `virtual future<> write(...) &&` overload?

nvartolomei: Also https://en.cppreference.com/w/cpp/language/const_cast notes: "Modifying a const object through a non-const access path and referring to a volatile object through a non-volatile [glvalue](https://en.cppreference.com/w/cpp/language/value_category#glvalue) results in undefined behavior."

travisdowns: Good eye, this is definitely wrong and left over from my earlier experimentation: will fix.

include/seastar/core/chunked_fifo.hh
travisdowns: `copy()` is removed now

include/seastar/core/chunked_fifo.hh
travisdowns: > But I think supporting the copy constructor is the path of least friction (though it opens the door to bad surprises).

Sounds good to me.

Push https://github.com/scylladb/seastar/commit/44e087a88b88002519672938e40535b6241524df adds full copy support to chunked_fifo, and removes `copy()`. This actually simplifies the fix to the dangling code that @nvartolomei pointed out: currently it's hard to fully support move-only types all the way down the serialization hierarchy as we have only a virtual `write(ostream) const` method on json elements: this is not suitable for move-only types, which need a && overload but then can't even compile the `const` method, so it all gets very messy (e.g., the code generator would need to track whether the current object was "tainted" anywhere by a move-only type and then stub out the const write method to throw an exception, or implement a different interface or something.

> We handle unexpected copies by having our small-scale performance tests monitor allocation count (and task count, and instruction count) per op and watching for changes.

I love this idea. Is this source-available so I can peek at it?

tests/unit/rest_api_httpd.cc
```diff
@@ -39,16 +39,25 @@ namespace bpo = boost::program_options;
...
+    api_json::hello_world.set(r, [] (const_req req) -> json::json_return_type {
         api_json::my_object obj;
         obj.var1 = req.param.at("var1");
         obj.var2 = req.param.at("var2");
-        api_json::ns_hello_world::query_enum v = api_json::ns_hello_world::str2query_enum(req.query_parameters.at("query_enum"));
+        query_enum v = str2query_enum(req.query_parameters.at("query_enum"));
         // This demonstrate enum conversion
         obj.enum_var = v;
-        return obj;
+        stream_enum is_streaming =str2stream_enum(req.query_parameters.at("stream_enum"));
```
tchaikov, line 57: might want to add spaces around `=`.

include/seastar/json/json_elements.hh
travisdowns: This is fixed in https://github.com/scylladb/seastar/commit/44e087a88b88002519672938e40535b6241524df

tests/unit/rest_api_httpd.cc
travisdowns: Eagle eyes :eagle: !

Fixed in https://github.com/scylladb/seastar/commit/9f222060c1e51bc2c96cc8b2a49023b140d00eae.

travisdowns:
@tchaikov anything I can help to do to move this one along? I don't see any outstanding concerns.

tests/unit/api.json
travisdowns: I changed it to a boolean in c9f848b1d diff: https://github.com/scylladb/seastar/compare/9f222060c1e51bc2c96cc8b2a49023b140d00eae..c9f848b1d81f616162787b67b2a55b921f20c0ba

travisdowns:
Push [c9f848b](https://github.com/scylladb/seastar/commit/c9f848b1d81f616162787b67b2a55b921f20c0ba) to change enum to boolean in the json api test per Avi's suggestion.

xemul (approved):
Since Avi had some concerns, I'll wait for his review for few more days.

I've a general question though (that doesn't block merging) -- one of the first patches says that making chunked_fifo copyable is the prerequisite to its usage by json2code. I'm not very familiar with the latter script (yet), so why is copyable container a requirement? Is it possible (or worth at all) to make json2code generate zero-copy paths?

travisdowns:
Thanks @xemul for your question.

> I'm not very familiar with the latter script (yet)

Basically it takes API definitions in json, including type definitions, and generates C++ wrapper types for the API types, which are json serializable. These types can be nested arbitrarily.

> so why is copyable container a requirement? Is it possible (or worth at all) to make json2code generate zero-copy paths?

A few reasons: some of the generated code relies on the copy constructor and or copy-assignment operator, in order to implement the copy constructor of the associated types. I don't think we can stop generating the copy constructor as that would be an API break. Arguably, at code generation time (in the script) we could try to detect if a "move only" type was used somewhere in the hierarchy of a type, and disable the copy construction/assignment in that case (and add a move constructor/assignment in all cases). It was discussed a bit above but @avikivity had suggested the path of least-resistance was to make the involved types copy-able.

Another reason not strictly related to the generator is that the C++ json serialization infrastructure does not fully support move-only objects. Before this change it didn't support them at all, but even now the rvalue-ness of the top-level object isn't pushed down arbitrarily far into the serialization process, so types with nested move-only objects can't always be serialized currently.

travisdowns:
@avikivity wrote:

> Could we make the copy functions conditional? That is, if the underlying type supports copy, so does the generated type?

Do you mean the types generated by json2code.py? 

If so, then yes we could and the approach was discussed a couple of times above. However, we don't need it now b/c json2code.py generates a restricted set of types and all of them support copy (including chunked_fifo after this change).

Also, move-only objects are not supported in some other parts of the json/http code so even if this was useful, it would not it itself make move only types seamless in the json endpoints.

include/seastar/core/chunked_fifo.hh
avikivity: Yes, see https://github.com/scylladb/scylladb/blob/master/test/perf/perf_simple_query.cc. It uses Seastar for the actual work.

include/seastar/core/chunked_fifo.hh
```diff
@@ -296,6 +298,25 @@ chunked_fifo<T, items_per_chunk>::chunked_fifo(chunked_fifo&& x) noexcept
     x._nfree_chunks = 0;
 }
 
+template <typename T, size_t items_per_chunk>
+inline
+chunked_fifo<T, items_per_chunk>::chunked_fifo(const chunked_fifo& rhs)
+        : chunked_fifo() {
+    std::copy_n(rhs.begin(), rhs.size(), std::back_inserter(*this));
```
avikivity, line 305: This is suboptimal (better to use uninitialized_copy_n so it can translate to memcpy), but not worth optimizing now.

PR #2648, opened 2025-02-18
build: finalize numactl dependency removal

tchaikov:
Following 8ab0267117, which dropped the build dependency on numactl, this commit removes all remaining references from the build system. This includes cleaning up template files (.pc.in and .cmake.in) that still contained unnecessary numactl references.

Changes:
- Remove Findnumactl.cmake since it's no longer needed at configuration time
- Remove numactl::numactl library linking as we no longer need its headers or libraries
- Remove unused Seastar_NUMA CMake option
- Remove numactl from SeastarDependencies.cmake

This completes the work started in 8ab0267117 and ensures Seastar applications no longer pull in numactl's cflags or libraries during the build process.


tchaikov:
cc @rockwotj

rockwotj:
Oh nice thanks @tchaikov! I thought we couldn't completely drop the dependency due to dpdk, but it seems that was wrong.. Thanks for fixing!

rockwotj:
d01796d4caffb23624e50ec818f3c14e6e38a957 was the bit I was missing. Anyways, LGTM

PR #2649, opened 2025-02-18
treewide: misc fixes for libc++ 19

rockwotj:
- **sstring: fixes for lib++ 19**
- **log: fixes for libc++ 19**
- **net: switch rss_key_type to std::span instead of std::string_view**


tchaikov (approved):
lgtm

PR #2650, opened 2025-02-20
http/client: Pass abort source by pointer

xemul:
Currently the client API suggest passing it by reference and for those, that don't want to abort requests there's 2nd set of calls that don't have abort source at all. Internally the reference is converted into a pointer, and the 2nd set of calls passes nullptr.

At the same time users of this API (scylla S3 client) make the reverse "choice" -- it passes abort source by pointer all the way down to http client and then decides which API to call, with abort source reference or without it.

Fix this mess by making client accept abort source by pointer.

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -180,7 +180,6 @@ private:
     requires std::invocable<Fn, connection&>
     auto with_new_connection(Fn&& fn, abort_source*);
 
-    future<> do_make_request(request& req, reply_handler& handle, abort_source*, std::optional<reply::status_type> expected);
     future<> do_make_request(connection& con, request& req, reply_handler& handle, abort_source*, std::optional<reply::status_type> expected);
 
 public:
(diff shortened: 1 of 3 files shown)


kreuzerkrieg (approved):
lgtm

include/seastar/http/client.hh
```diff
@@ -259,33 +258,13 @@ public:
...
-
-    /**
-     * \brief Send the request and handle the response, same as \ref make_request()
-     *
-     *  @attention Note that the method does not take the ownership of the
-     * `request and the `handle`, it caller's responsibility the make sure they
-     * are referencing valid instances
-     */
-    future<> make_request(request& req, reply_handler& handle, std::optional<reply::status_type> expected = std::nullopt);
+    future<> make_request(request&& req, reply_handler&& handle, std::optional<reply::status_type>&& expected = std::nullopt, abort_source* as = nullptr);
```
avikivity, line 267: Shouldn't you keep the old API, perhaps deprecated?

The change from value parameter to rvalue reference can break callers.

avikivity: Maybe you can pass a request_control struct that contains expected and abort_source, so we can grow more control parameters later.

xemul: > Shouldn't you keep the old API, perhaps deprecated?

It's still `namespace experimental`, so I'd take advantage of it

xemul: > Maybe you can pass a request_control struct that contains expected and abort_source, so we can grow more control parameters later.

But we can as well grow the arguments list. So far there are two "controls" which is probably not enough for the dedicated controller struct

xemul: @avikivity , reminder ping

avikivity: Sorry. Wouldn't we also expect an optional timeout in the future?

xemul: With the `abort_source` at hand it's [arguable](https://github.com/scylladb/seastar/pull/1889#issuecomment-1777915851) that dedicated timeout parameter will be needed

PR #2651, opened 2025-02-20
net: functions describing ssl connection

ptrsmrn:
Needed in https://github.com/scylladb/scylladb/pull/22961

2 methods have been added in order to expose more
information on the characteristics of a TLS socket connection.

### DIFF
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -563,6 +563,23 @@ namespace tls {
     */
     future<std::optional<sstring>> get_selected_alpn_protocol(connected_socket&);
 
+    /**
+     * Returns the cipher suite used in the connection. This string depends on the internal implementation:
+     * it's e.g. "TLS_AES_256_GCM_SHA384" for gnutls and may be different for OpenSSL.
+     *
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
+    */
+    future<sstring> get_cipher_suite(connected_socket& socket);
+
+    /**
+     * Returns the protocol version used in the connection, e.g. "TLS1.3"
+     *
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
+    */
+    future<sstring> get_protocol_version(connected_socket& socket);
+
     std::ostream& operator<<(std::ostream&, const subject_alt_name::value_type&);
     std::ostream& operator<<(std::ostream&, const subject_alt_name&);
 
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -1870,6 +1870,19 @@ public:
     }
 
     struct session_ref;
+
+    future<sstring> get_cipher_suite() {
+        return state_checked_access([this] {
+            return sstring(gnutls_ciphersuite_get(*this));
+        });
+    }
+
+    future<sstring> get_protocol_version() {
+        return state_checked_access([this]() {
+            return sstring(gnutls_protocol_get_name(gnutls_protocol_get_version(*this)));
+        });
+    }
+
 private:
 
     using x509_ctr_ptr = std::unique_ptr<gnutls_x509_crt_int, void (*)(gnutls_x509_crt_t)>;
@@ -2015,6 +2028,12 @@ public:
     }
     future<std::optional<sstring>> get_selected_alpn_protocol() {
         return _session->get_selected_alpn_protocol();
+    }
+    future<sstring> get_cipher_suite() const {
+        return _session->get_cipher_suite();
+    }
+    future<sstring> get_protocol_version() const {
+        return _session->get_protocol_version();
     }
 };
 
@@ -2212,6 +2231,14 @@ future<tls::session_data> tls::get_session_resume_data(connected_socket& socket)
 
 future<std::optional<sstring>> tls::get_selected_alpn_protocol(connected_socket& socket) {
     return get_tls_socket(socket)->get_selected_alpn_protocol();
+}
+
+future<sstring> tls::get_cipher_suite(connected_socket& socket) {
+    return get_tls_socket(socket)->get_cipher_suite();
+}
+
+future<sstring> tls::get_protocol_version(connected_socket& socket) {
+    return get_tls_socket(socket)->get_protocol_version();
 }
 
 std::string_view tls::format_as(subject_alt_name_type type) {
--- a/tests/unit/tls_test.cc
+++ b/tests/unit/tls_test.cc
@@ -1357,7 +1357,7 @@ SEASTAR_THREAD_TEST_CASE(test_dn_name_handling) {
     // - mtls_client2.crt - second client certificate
     //
     // The test runs server that uses mtls_server.crt.
-    // The server accepts two incomming connections, first one uses mtls_client1.crt
+    // The server accepts two incoming connections, first one uses mtls_client1.crt
     // and the second one uses mtls_client2.crt. Every client sends a short string
     // that server receives and tries to find it in the DN string.
 
@@ -1735,7 +1735,7 @@ SEASTAR_THREAD_TEST_CASE(test_tls13_session_tickets_invalidated_by_reload) {
     namespace fs = std::filesystem;
 
     // copy the wrong certs. We don't trust these
-    // blocking calls, but this is a test and seastar does not have a copy 
+    // blocking calls, but this is a test and seastar does not have a copy
     // util and I am lazy...
     fs::copy_file(certfile("test.crt"), tmp.path() / "test.crt");
     fs::copy_file(certfile("test.key"), tmp.path() / "test.key");
@@ -1858,7 +1858,7 @@ SEASTAR_THREAD_TEST_CASE(test_reload_certificates_with_only_shard0_notify) {
     namespace fs = std::filesystem;
 
     // copy the wrong certs. We don't trust these
-    // blocking calls, but this is a test and seastar does not have a copy 
+    // blocking calls, but this is a test and seastar does not have a copy
     // util and I am lazy...
     fs::copy_file(certfile("other.crt"), tmp.path() / "test.crt");
     fs::copy_file(certfile("other.key"), tmp.path() / "test.key");
@@ -1973,4 +1973,19 @@ SEASTAR_THREAD_TEST_CASE(test_reload_certificates_with_only_shard0_notify) {
 
         BOOST_CHECK_EQUAL(sstring(buf.begin(), buf.end()), "apa");
     }
+}
+
+SEASTAR_TEST_CASE(test_tls_cipher_suite_and_protocol_version, *enable_if_with_networking()) {
+    auto certs = ::make_shared<tls::certificate_credentials>();
+    co_await certs->set_system_trust();
+
+    auto c = co_await tls::connect(certs, co_await google_address(), tls::tls_options{ .server_name = google_name });
+    BOOST_CHECK_EQUAL(co_await tls::get_cipher_suite(c), "TLS_AES_256_GCM_SHA384");
+    BOOST_CHECK_EQUAL(co_await tls::get_protocol_version(c), "TLS1.3");
+}
+
+SEASTAR_TEST_CASE(test_cipher_suite_and_protocol_version_for_non_tls_connection, *enable_if_with_networking()) {
+    auto c = co_await seastar::connect(co_await google_address());
+    BOOST_CHECK_THROW(co_await tls::get_cipher_suite(c), std::invalid_argument);
+    BOOST_CHECK_THROW(co_await tls::get_protocol_version(c), std::invalid_argument);
 }


nuivall:
I think you need to open PR against seastar, unless some process changed recently

src/net/tls.cc
```diff
@@ -2147,6 +2169,18 @@ future<tls::session_data> tls::get_session_resume_data(connected_socket& socket)
     return get_tls_socket(socket)->get_session_resume_data();
 }
 
+bool tls::is_operational(connected_socket& socket) {
+    return get_tls_socket(socket)->is_operational();
```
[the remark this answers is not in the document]
ptrsmrn: I don't see a need for that. As I said, the functions I introduced have signatures aligned to already existing ones. Also, adding `mutable` on the scylla-server side https://github.com/scylladb/scylladb/pull/22961#discussion_r1965485786 is enough to work around this minor problem, and find these const cast ugly. I also see the benefit in keeping the interface non-const, at least it doesn't lie about its intent https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es50-dont-cast-away-const
For now I'll close this comment, but if anyone wants to push it, we can reconsider.

include/seastar/net/tls.hh
```diff
@@ -563,6 +563,30 @@ namespace tls {
     */
     future<std::optional<sstring>> get_selected_alpn_protocol(connected_socket&);
 
+    /**
+     * Returns true if the handshake has been completed.
+     *
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
+    */
+    bool is_operational(connected_socket& socket);
```
avikivity: How can this be used? poll continuously until it returns true?

src/net/tls.cc
```diff
@@ -2016,6 +2029,15 @@ class tls_connected_socket_impl : public net::connected_socket_impl, public sess
...
     }
+    bool is_operational() {
+        return _session->is_operational();
+    }
+    sstring get_cipher_suite() {
+        return _session->get_cipher_suite();
+    }
+    sstring get_protocol_version() {
+        return _session->get_protocol_version();
+    }
```
avikivity, line 2037: const?

avikivity:
Missing tests.

Is the cypher suite portable across openssl/gnutls?

include/seastar/net/tls.hh
ptrsmrn: Call only once, an example usage is in the PR that requires this change: https://github.com/scylladb/scylladb/pull/22961

ptrsmrn:
> Missing tests.

@avikivity a test in the PR that requires this change: scylladb/scylladb#22961 wouldn't be enough?

ptrsmrn:
@avikivity 

> Missing tests.

Do we really need tests for functions, which simply forward calls to the library functions?

> Is the cypher suite portable across openssl/gnutls?

No, AFAIK (also see this https://github.com/scylladb/scylladb/pull/22961/files#diff-eaa341e073bc04305e212cf7bd686567c2c94ba47a38f4773c119fbd13ca6fe5R18-R29, which is my attempt at making C lib names translatable to python lib names). Are you just asking out of curiosity or thinking about introducing some translation layer, like this https://github.com/scylladb/seastar/pull/2651/files#diff-435f870642b88c782b2ebbffb892e2b49da775263c775af86a4cbca56e4d2fb3L569-L583 ?

CC @elcallio

include/seastar/net/tls.hh
avikivity: I don't understand. You call it once and then?

avikivity:
> > Missing tests.
> 
> @avikivity a test in the PR that requires this change: [scylladb/scylladb#22961](https://github.com/scylladb/scylladb/pull/22961) wouldn't be enough?

That's a different project.

include/seastar/net/tls.hh
ptrsmrn: This function is used e.g. here https://github.com/scylladb/scylladb/pull/22961/files#diff-5ddc61dd27612d0e2d48deba0325d00a24bf8d14d0b05de0f3b4126f4632236fR657 as:
```
cd.ssl_enabled = tls::is_operational(_fd);
```
so I'm just calling it once to set system.clients.ssl_enabled column.

ptrsmrn:
> > > Missing tests.
> > 
> > 
> > @avikivity a test in the PR that requires this change: [scylladb/scylladb#22961](https://github.com/scylladb/scylladb/pull/22961) wouldn't be enough?
> 
> That's a different project.

UPDATE: Added tests. Actually only added extra asserts to already existing testcases, whole separate testcases for functions that just forward lib calls seemed like to much.
One thing I'm wondering about is whether to assert for specific strings or just if strings are not empty. Both have their disadvantages.

ptrsmrn:
> > > > Missing tests.
> > > 
> > > 
> > > @avikivity a test in the PR that requires this change: [scylladb/scylladb#22961](https://github.com/scylladb/scylladb/pull/22961) wouldn't be enough?
> > 
> > 
> > That's a different project.
> 
> UPDATE: Added tests. Actually only added extra asserts to already existing testcases, whole separate testcases for functions that just forward lib calls seemed like to much. One thing I'm wondering about is whether to assert for specific strings or just if strings are not empty. Both have their disadvantages.

@avikivity WDYT?

include/seastar/net/tls.hh
avikivity: What if it becomes operational later?

avikivity:
> @avikivity
> 
> > Missing tests.
> 
> Do we really need tests for functions, which simply forward calls to the library functions?


Yes

> 
> > Is the cypher suite portable across openssl/gnutls?
> 
> No, AFAIK (also see this https://github.com/scylladb/scylladb/pull/22961/files#diff-eaa341e073bc04305e212cf7bd686567c2c94ba47a38f4773c119fbd13ca6fe5R18-R29, which is my attempt at making C lib names translatable to python lib names). Are you just asking out of curiosity or thinking about introducing some translation layer, like this https://github.com/scylladb/seastar/pull/2651/files#diff-435f870642b88c782b2ebbffb892e2b49da775263c775af86a4cbca56e4d2fb3L569-L583 ?


I want the freedom to migrate to openssl and APIs that aren't compatible make it more difficult.

If it's just informational, okay, but document it so.

include/seastar/net/tls.hh
```diff
@@ -563,6 +563,30 @@ namespace tls {
     */
     future<std::optional<sstring>> get_selected_alpn_protocol(connected_socket&);
 
+    /**
+     * Returns true if the handshake has been completed.
+     *
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
+    */
```
nyh, line 572: Looking at the code, the second type of exception is an std::invalid_argument.

I'm not sure how does a user of this API avoid these exceptions. Is thee some is_tls() function you can call to check if a specific socket is TLS at all, before you call this function and it croaks?

include/seastar/net/tls.hh
```diff
@@ -563,6 +563,30 @@ namespace tls {
...
+    bool is_operational(connected_socket& socket);
+
+    /**
+     * Returns the cipher suite used in the connection.
+     *
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
```
nyh, line 579: What if the handshake hasn't finished, what is returned? Are you supposed to check is_operational() before calling this function?
I'm not sure I understand how it's used.

include/seastar/net/tls.hh
```diff
@@ -563,6 +563,30 @@ namespace tls {
...
+    */
+    sstring get_cipher_suite(connected_socket& socket);
+
+    /**
+     * Returns the protocol version used in the connection.
+     *
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
+    */
```
nyh: Same question I had above also applies here.

src/net/tls.cc
```diff
@@ -1870,6 +1870,19 @@ class session : public enable_lw_shared_from_this<session> {
     }
 
     struct session_ref;
+
+    bool is_operational() const {
+        return gnutls_record_get_direction(*this) >= 0; // handshake succeeded
```
nyh: I plead my ignorance here, but:
1. Why does get_direction >= 0 say anything? The documentation says it can return 0 or 1, saying nothing about a negative response...
2. If, as the comment says, you check if "handshake succeeded", why not call this function handshake_succeeded() instead of "is_operational()"? I think It will be a lot clearer what exactly this function is supposed to do.

tests/unit/socket_test.cc
```diff
@@ -94,6 +95,9 @@ SEASTAR_TEST_CASE(socket_skip_test) {
         abort_source as;
         auto client = async([&as] {
             connected_socket socket = connect(ipv4_addr("127.0.0.1", 1234)).get();
+            BOOST_CHECK_THROW(tls::is_operational(socket), std::invalid_argument);
+            BOOST_CHECK_THROW(tls::get_cipher_suite(socket), std::invalid_argument);
+            BOOST_CHECK_THROW(tls::get_protocol_version(socket), std::invalid_argument);
```
nyh: I don't think it's a good idea to take an existing test socket_skip_test(), which as its name suggest checks skip(), and stick in the middle of it some completely unrelated check of TLS functions...

I think this test belongs as a new test function in test/unit/tls_test.cc.

tests/unit/tls_test.cc
```diff
@@ -229,6 +229,9 @@ SEASTAR_TEST_CASE(test_alpn_client_negotiate_h2_with_google,
     if (*selected_alpn != "h2") {
         BOOST_TEST_MESSAGE("Warning: Google did not select 'h2'. Selected: " + *selected_alpn);
     }
+    BOOST_CHECK(tls::is_operational(c));
+    BOOST_CHECK_EQUAL(tls::get_cipher_suite(c), "TLS_AES_256_GCM_SHA384");
+    BOOST_CHECK_EQUAL(tls::get_protocol_version(c), "TLS1.3");
```
nyh: I'm not thrilled about this test, which just enshrines some sort of arbitrary connection parameters in our connection to some Google server we don't control. I'm actually shocked we even have connections to Google inside our unit tests, which makes no sense (in our unit tests, we should implement both server and client!).

Given that this Google stuff is already in the unit test, I tend to allow this to go in as well, but I'm pretty sure we'll regret this in the future when Google decided to change its default cipher or something and this test will suddenly begin to fail.

nyh:
> > @avikivity
> > > Missing tests.
> > 
> > 
> > Do we really need tests for functions, which simply forward calls to the library functions?
> 
> Yes
> 
> > > Is the cypher suite portable across openssl/gnutls?
> > 
> > 
> > No, AFAIK (also see this https://github.com/scylladb/scylladb/pull/22961/files#diff-eaa341e073bc04305e212cf7bd686567c2c94ba47a38f4773c119fbd13ca6fe5R18-R29, which is my attempt at making C lib names translatable to python lib names). Are you just asking out of curiosity or thinking about introducing some translation layer, like this https://github.com/scylladb/seastar/pull/2651/files#diff-435f870642b88c782b2ebbffb892e2b49da775263c775af86a4cbca56e4d2fb3L569-L583 ?
> 
> I want the freedom to migrate to openssl and APIs that aren't compatible make it more difficult.
> 
> If it's just informational, okay, but document it so.


I agree. I think the API comment should at least be candid that the response from get_cipher_suite() is a string that depends on the internal implementation, and then give an example that the current implementation uses gnutls and the string looks like <give example>.

It also means that the unit test shouldn't check that the cipher suite is something specific. It can test a string is returned (no exception), and maybe even that the string is not empty, but I don't see why we need to check that it returns a specific string, if we don't make any guarantee what this string means.

include/seastar/net/tls.hh
ptrsmrn: I have not found such a function. To the best of my understanding, these functions are ought to be called once you're sure you're interactive with a TLS socket, e.g. from within other TLS functions and generally in a TLS context.

ptrsmrn: Speaking about this specific `is_operational` function, I'll most likely remove it.

include/seastar/net/tls.hh
ptrsmrn: To the best of my understanding, this and other already existing functions should be called on a socket you're sure is a TLS socket.

include/seastar/net/tls.hh
ptrsmrn: To the best of my understanding, this and other already existing functions should be called on a socket you're sure is a TLS socket.

tests/unit/socket_test.cc
ptrsmrn: I'll provide separate testcases.

tests/unit/tls_test.cc
ptrsmrn: I can write a separate testcase.
BTW. assuming that we don't have any other tests for seastar than these UTs, connecting to Google make practical sense, because such tests prove our networking stack really works and if someone works offline, these UTs would fail for him, but he probably doesn't need them in the first place (unless working in air-gapped environments). I could make my testcase less dependent on what TLS strings Google returns, but then I'd be just asserting that these strings are not empty, which has drawbacks of its own, e.g. if Google was to return an error or a warning msg, the testcase would still pass, whereas it should fail.

include/seastar/net/tls.hh
ptrsmrn: This function has been removed altogether. It's not needed anymore, I get this information from another place.

src/net/tls.cc
ptrsmrn: This function has been removed altogether.

tests/unit/tls_test.cc
ptrsmrn: I added 2 separate testcases.

tests/unit/socket_test.cc
ptrsmrn: Done.

src/net/tls.cc
ptrsmrn: Done.

ptrsmrn:
UPDATE 2:
- removed `is_operational()` function
- testing is done in separate testcases, and not within the existing ones

ptrsmrn:
@nyh @avikivity could you please take another look?

PR #2658, opened 2025-02-23
docker: update LLVM debian repo for Ubuntu Orcular migration

tchaikov:
When migrating from Ubuntu Matis to Orcular in afea41b1, the LLVM debian package repository configuration wasn't updated accordingly. This caused Docker builds to fail.

This commit updates the repository configuration to match the new Ubuntu Orcular base image.

Fixes #2656

PR #2660, opened 2025-02-24
doc: Update documentation after removing build.sh

tchaikov:
In 5634985c, scripts/build.sh was removed, but related documentation wasn’t updated. This change revises the docs to reflect that removal. Additionally, to address c-ares incompatibility in the container image, the updated command in doc/tutorial.md now builds a compatible c-ares version from source.


tchaikov:
@xemul hi Pavel, could you please review this change? I noticed the issue while tracking down documentation for our container image in the source tree during testing for #2658.

PR #2662, opened 2025-02-25
httpclient: Add exception handling to connection::close

### LINKED ISSUE #2661: HTTP client can end up with abandoned exception futures iff conenction is broken
In `connection::close()`, the input and output streams are closed in a when_all. However, at least output stream _can_ cause exception throwing, as possible flush calls (on an already broken socket) may throw here. The result of the when_all is not really processed, so the future result is ignored -> warning logs and test failures in caller code.

Assuming we want to swallow the exceptions in close (because what use are they?), fix is simply to add `handle_exception` tails to the `close` calls.

elcallio:
Fixes #2661

In connection::close(), the input and output streams are closed in a when_all. However, at least output stream can cause exception throwing, as possible flush calls (on an already broken socket) may throw here. The result of the when_all is not really processed, so the future result is ignored -> warning logs and test failures in caller code.

Assuming we want to swallow the exceptions in close (because what use are they?), fix is simply to add handle_exception tails to the close calls.


src/http/client.cc
```diff
@@ -195,7 +195,10 @@ void connection::shutdown() noexcept {
 }
 
 future<> connection::close() {
-    return when_all(_read_buf.close(), _write_buf.close()).discard_result().then([this] {
+    // #2661. At least output stream close can fail with exception, because
+    // the stream will do a flush. If connection never managed to send data, we
+    // will throw again here. Need to suppress these exceptions.
+    return when_all(_read_buf.close().handle_exception([](auto&&){}), _write_buf.close().handle_exception([](auto&&){})).discard_result().then([this] {
```
tchaikov, line 201: IIUC, the race condition occurs when `close()` is called while make_request() is still in progress.


Specifically:
1. make_request() isn't fully resolved until _write_buf is completely flushed
2. If close() is called before this flush completes, we'll get exceptions from the streams

I think we might need to make it the caller's responsibility to await the future from `make_request()` before calling `close()` instead of silently discarding exceptions. this could hide potential bugs in the Seastar applications which either discard the future or close the connection without waiting for it.

xemul:
I wonder why CI didn't pick this PR. @tchaikov , any ideas?

xemul:
> ... However, at least output stream [close() method] can cause exception throwing ...

Cc @bhalevy

elcallio:
Up until now we had no tests causing wire errors. EAR has tests that manually creates them -> boom. Was not discovered until now, because previously, the provider used its own http client. Now, using the seastar one, the problem manifests.

xemul:
No, no, I didn't mean Scylla CI. I mean seastar CI. E.g. check #2659 , it had passed through several checks, e.g. [this](https://github.com/scylladb/seastar/actions/runs/13490231942/job/37687213234?pr=2659) one. You PR somehow was skipped by GH actions

elcallio:
Ah, well, that is strange indeed.

elcallio:
Added a small unit test.

tchaikov:
> I wonder why CI didn't pick this PR. @tchaikov , any ideas?

in deed. i checked the runs of https://github.com/scylladb/seastar/actions/workflows/python-lint.yaml, buy the latest two runs from Calle was 

- https://github.com/scylladb/seastar/actions/runs/13548164965
- https://github.com/scylladb/seastar/actions/runs/13497553196

there should have been another one triggered from his fork 3 days ago. i am not able to see the status of the runs triggered by the previous revision of his pull request, as it has been overridden by the the latest revision. but if it happens again, please let me know so that I can check it.

PR #2671, opened 2025-03-03
exception_hacks: Make dl_iterate_phdr resolution manual

### LINKED ISSUE #2670: Resolving "dl_iterate_phdr_org" in exception hacks can fail when called via ASAN _dl_init
On some builds/machines (mine!), running in debug can/will cause dl_iterate_phdr_org to be called during dl_init via ASAN init, before _dl_init reached the actual seastar module. At this point, automatic resolve of (even function local) statics with automatic init does not work yet. 

The dl_iterate_phdr_org resolution uses a "= [] (...)()" lambda invocation to do its resolution. Unfortunately, the lambda will be non-inited at this time (and since debug, not optimised away either).

Simple solution is to make the check and lookup manual.

elcallio:
Fixes #2670

Make check and resolution of "actual" dl_iterate_phdr manual, to avoid chicken-and-egg iff called via dl_init, before module is initialized.

### DIFF
--- a/src/core/exception_hacks.cc
+++ b/src/core/exception_hacks.cc
@@ -75,11 +75,32 @@ using dl_iterate_fn = int (*) (int (*callback) (struct dl_phdr_info *info, size_
 
 [[gnu::no_sanitize_address]]
 static dl_iterate_fn dl_iterate_phdr_org() {
-    static dl_iterate_fn org = [] {
-        auto org = (dl_iterate_fn)dlsym (RTLD_NEXT, "dl_iterate_phdr");
-        SEASTAR_ASSERT(org);
-        return org;
-    }();
+    // #2670 - Do the check and resolve manually, to avoid
+    // init order fiasco iff we're called before dl_init of this
+    // module.
+    // During initialization, ASan verifies it's the first DSO in the library list
+    // by calling dl_iterate_phdr(3). This is necessary for ASan to properly
+    // intercept memory operations.
+    //
+    // However, Seastar also provides the dl_iterate_phdr symbol, and its
+    // implementation depends on a static local variable `dl_iterate_fn`. This
+    // creates a circular dependency:
+    //
+    // 1. ASan loads first (as required)
+    // 2. ASan calls dl_iterate_phdr
+    // 3. The linker resolves this to Seastar's implementation
+    // 4. But Seastar's shared library hasn't been fully initialized via dl_init()
+    // 5. The function local static variable `dl_iterate_fn` that Seastar's implementation requires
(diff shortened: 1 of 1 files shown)


src/core/exception_hacks.cc
```diff
@@ -75,11 +75,14 @@ using dl_iterate_fn = int (*) (int (*callback) (struct dl_phdr_info *info, size_
...
-        SEASTAR_ASSERT(org);
+    // #2670 - do the check and resolve manually, to avoid
+    // init order fiasco iff we're called before dl_init of this
+    // module.
+    static dl_iterate_fn org = nullptr;
+    if (org == nullptr) {
+        org = (dl_iterate_fn)dlsym (RTLD_NEXT, "dl_iterate_phdr");
         return org;
-    }();
+    }
```
tchaikov, line 85: would simplify the change a little bit by dropping the `return org` statement in the `if` block:

```c++
static dl_iterate_fn org = nullptr;
if (org == nullptr) {
    org = (dl_iterate_fn)dlsym (RTLD_NEXT, "dl_iterate_phdr");
}
return org;
```

tchaikov (commented):
thank you for the change. previously, we were using a workaround introduced by f34335e5fac747a69800ecc67aa50897ce6f0f36 to prevent ASan from enabling the "verify_asan_link_order" check. see https://github.com/scylladb/seastar/blob/3ee9204cb792885c4a0bdb19a61619572d88c963/CMakeLists.txt#L350

with your change, we don't need to to use it anymore.

src/core/exception_hacks.cc
```diff
@@ -75,11 +75,14 @@ using dl_iterate_fn = int (*) (int (*callback) (struct dl_phdr_info *info, size_
 
 [[gnu::no_sanitize_address]]
 static dl_iterate_fn dl_iterate_phdr_org() {
-    static dl_iterate_fn org = [] {
-        auto org = (dl_iterate_fn)dlsym (RTLD_NEXT, "dl_iterate_phdr");
-        SEASTAR_ASSERT(org);
+    // #2670 - do the check and resolve manually, to avoid
+    // init order fiasco iff we're called before dl_init of this
+    // module.
```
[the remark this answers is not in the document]
elcallio: Added to both patch blurb and comment in code.

src/core/exception_hacks.cc
elcallio: True.

tchaikov (approved):
lgtm.

would be better to drop the workaround in CMakeLists.txt. see https://github.com/scylladb/seastar/pull/2671#pullrequestreview-2655808319 . but this change alone is an improvement.

PR #2673, opened 2025-03-03
future: remove deprecated functions/concepts

avikivity:
These were deprecated 4 years ago (except one, 1 year ago) and are mostly related to variadic futures.

### DIFF
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -906,9 +906,7 @@ private:
 
 /// \brief promise - allows a future value to be made available at a later time.
 ///
-/// \tparam T A list of types to be carried as the result of the associated future.
-///           A list with two or more types is deprecated; use
-///           \c promise<std::tuple<T...>> instead.
+/// \tparam T A type to be carried as the result of the associated future. Use void (default) for no result.
 SEASTAR_MODULE_EXPORT
 template <typename T>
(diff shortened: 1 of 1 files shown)


include/seastar/core/future.hh
```diff
@@ -1329,31 +1306,7 @@ public:
...
-    /// \c T, while \ref get() returns \c T&& (an rvalue reference to
-    /// \c T). This distinction typically won't cause issues when switching
-    /// from \c get0() to \ref get(). However, in specific metaprogramming
-    /// scenarios, especially when the code expects type \c T, you'll need
-    /// to use \c std::remove_reference_t<decltype(fut.get())> to extract
-    /// the underlying type \c T.
-    /// For new code that utilizes \c future<tuple<...>>, employ
-    /// \c std::get<0>(fut.get()) to access the first element of the tuple,
-    /// rather than the deprecated \ref get0().
     using get0_return_type = typename future_state::get0_return_type;
```
nyh, line 1309: I don't remember - why did we only deprecate one of these get0() functions and leave the other ones undeprecated, and also get0_return_type wasn't deprecated?

nyh (approved):

include/seastar/core/future.hh
avikivity: I guess overlooked.

PR #1419, opened 2023-01-11
file: Generalize in/out streams making

xemul:
In order to make input/output stream for a file there are two helpers -- make_file_(input|output)_stream out there. Both make file-based sink or source, but these sink and source implementations silently assume that the file is a file-on-disk. Strictly speaking, these are compatible with posix_file_impl, but not with generic file. E.g. pipes (if they exist) might break on read-ahead and write-behind.

This patch adds a customization point to file_impl so that custom file implementations could generate specific sink/source for the stream.

One of the next users of it would be S3 file, which supports posix-ish data_source, but will require very specific data_sink with append-only and extra buffering of 5Mb scattered chunks.

Signed-off-by: Pavel Emelyanov <xemul@scylladb.com>

PR #2678, opened 2025-03-05
fstream: Export functions for making file_data_source

xemul:
This is to facilitate making different sources depending on the actual file "type" and letting the caller wrap it into the input_stream().

Symmetrical make_file_data_sink was introduced by 52e0a763e87, not it's file_data_source turn.

Previous attempt to help file->source/sink conversion was #1419

### DIFF
--- a/src/core/fstream.cc
+++ b/src/core/fstream.cc
@@ -342,17 +342,17 @@ private:
     }
 };
 
-class file_data_source : public data_source {
-public:
-    file_data_source(file f, uint64_t offset, uint64_t len, file_input_stream_options options)
-        : data_source(std::make_unique<file_data_source_impl>(
-                std::move(f), offset, len, options)) {}
-};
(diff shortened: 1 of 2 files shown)


src/core/fstream.cc
```diff
@@ -342,17 +342,17 @@ class file_data_source_impl : public data_source_impl {
...
 };
 
-class file_data_source : public data_source {
-public:
-    file_data_source(file f, uint64_t offset, uint64_t len, file_input_stream_options options)
-        : data_source(std::make_unique<file_data_source_impl>(
-                std::move(f), offset, len, options)) {}
-};
+data_source make_file_data_source(file f, uint64_t offset, uint64_t len, file_input_stream_options opt) noexcept {
+    return data_source(std::make_unique<file_data_source_impl>(std::move(f), offset, len, std::move(opt)));
```
avikivity, line 346: It's  not noexcept due to make_unique.

PR #2679, opened 2025-03-06
smp: prefaulter: don't leave zombie worker threads

### LINKED ISSUE #2623: memory_prefaulter leaves a zombie pthread, confuses gdb during coredump debugging
After the `memory_prefaulter` threads do their job and exit, their handles aren't join()ed until `smp::~smp` runs.
This means that they leave a zombie entry in pthread's thread lists. 

As it turns out, this confuses gdb. When looking for threads and their TLS segments, gdb uses `td_ta_thr_iter` (from libthread-db/nplt-db) to iterate over all threads:
https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/linux-thread-db.c;h=9d84187a9ad0897ced25c7639e92ebb2e1e96746;hb=refs/heads/master#l1540

Then, it uses the `ti_lid` field (which is supposed to contain the PID of the thread) returned by libthread-db as an identifier for the thread:
https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/linux-thread-db.c;h=9d84187a9ad0897ced25c7639e92ebb2e1e96746;hb=refs/heads/master#l1513

The issue is: for those zombie entries, libthread-db reports the PID of the process (which in Seastar is equal to the PID of reactor-0), not the PID of the thread:
https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl_db/td_thr_get_info.c;h=7a64ef4c63614cf86714df9fc56f235b04f52253;hb=refs/heads/master#l108

So what happens is that gdb sees the handle of the zombie `memory_prefaulter` thread, thinks it's the entry of reactor-0, and uses *this* entry to serve queries about thread-local variables of reactor-0. The real entry of reactor-0 is ignored, since an entry for this PID has already been recorded by the time the real entry is seen.

```
(gdb) maintenance check libthread-db 

Running libthread_db integrity checks:
  Got thread 0x7f3920bff6c0 => 5938; can't map_lwp2thr ... OK <------------ memory_prefaulter
  Got thread 0x7f39215ff6c0 => 5952; can't map_lwp2thr ... OK
  Got thread 0x7f39223ff6c0 => 5951; can't map_lwp2thr ... OK
  Got thread 0x7f3922dff6c0 => 5950; can't map_lwp2thr ... OK
  Got thread 0x7f39233ff6c0 => 5949; can't map_lwp2thr ... OK
  Got thread 0x7f39239ff6c0 => 5948; can't map_lwp2thr ... OK
  Got thread 0x7f39285ff6c0 => 5947; can't map_lwp2thr ... OK
  Got thread 0x7f3928fff6c0 => 5946; can't map_lwp2thr ... OK
  Got thread 0x7f39295ff6c0 => 5945; can't map_lwp2thr ... OK
  Got thread 0x7f3929bff6c0 => 5944; can't map_lwp2thr ... OK
  Got thread 0x7f392a1ff6c0 => 5943; can't map_lwp2thr ... OK
  Got thread 0x7f392a7ff6c0 => 5942; can't map_lwp2thr ... OK
  Got thread 0x7f392adff6c0 => 5941; can't map_lwp2thr ... OK
  Got thread 0x7f392b3ff6c0 => 5940; can't map_lwp2thr ... OK
  Got thread 0x7f392be64140 => 5938; can't map_lwp2thr ... OK <------------ reactor-0
  Got thread 0x7f392be216c0 => 5939; can't map_lwp2thr ... OK

(gdb) thread
[Current thread is 1 (Thread 0x7f3920bff6c0 (LWP 5938))]
(gdb) p/x $fs_base
$1 = 0x7f392be64140
(gdb) p &seastar::local_engine
$2 = (seastar::reactor **) 0x7f3920be51a8
(gdb) p seastar::local_engine
$3 = (seastar::reactor *) 0x0
```

Refs https://github.com/scylladb/scylladb/issues/15665#issuecomment-1752256111
Refs https://github.com/scylladb/scylladb/issues/19110#issuecomment-2152729026
Refs https://github.com/scylladb/scylladb/issues/19110#issuecomment-2150456968
Refs https://github.com/scylladb/scylladb/issues/22245#issuecomment-2585796150

avikivity:
As explained in #2623 in detail, the prefaulter worker threads that have completed but not joined are left in a zombie state, which confuses gdb thread_local processing. As seastar relies on thread locals heavily, it becomes impossible to debug core dumps.

Fix this by joining the threads after they complete. Use seastar::alien to ask the main reactor threads to join the completed threads when they are done, so it won't stall.

Fixes #2623.

### DIFF
--- a/src/core/smp.cc
+++ b/src/core/smp.cc
@@ -34,6 +34,7 @@ module;
 module seastar;
 #else
 #include <seastar/core/smp.hh>
+#include <seastar/core/alien.hh>
 #include <seastar/core/resource.hh>
 #include <seastar/core/loop.hh>
 #include <seastar/core/semaphore.hh>
@@ -175,7 +176,7 @@ void
 smp::setup_prefaulter(const seastar::resource::resources& res, seastar::memory::internal::numa_layout layout) {
     // Stack guards mprotect() random pages, so the prefaulter will hard-fault.
 #ifndef SEASTAR_THREAD_STACK_GUARDS
-    _prefaulter = std::make_unique<internal::memory_prefaulter>(res, std::move(layout));
+    _prefaulter = std::make_unique<internal::memory_prefaulter>(_alien, res, std::move(layout));
 #endif
 }
 
@@ -199,7 +200,7 @@ get_huge_page_size() {
     return std::nullopt;
 }
 
-internal::memory_prefaulter::memory_prefaulter(const resource::resources& res, memory::internal::numa_layout layout) {
+internal::memory_prefaulter::memory_prefaulter(alien::instance& alien, const resource::resources& res, memory::internal::numa_layout layout) {
     for (auto& range : layout.ranges) {
         _layout_by_node_id[range.numa_node_id].push_back(std::move(range));
     }
@@ -218,17 +219,27 @@ internal::memory_prefaulter::memory_prefaulter(const resource::resources& res, m
             }
             a.set(cpuset);
         }
-        _worker_threads.emplace_back(a, [this, &ranges, page_size, huge_page_size_opt] {
+        _worker_threads.emplace_back(a, [this, &alien, &ranges, page_size, huge_page_size_opt] {
+            ++_active_threads;
             work(ranges, page_size, huge_page_size_opt);
+            if (!--_active_threads) {
+                run_on(alien, 0, [this] () noexcept { join_threads(); });
+            }
         });
     }
 }
 
-internal::memory_prefaulter::~memory_prefaulter() {
+void
+internal::memory_prefaulter::join_threads() noexcept {
     _stop_request.store(true, std::memory_order_relaxed);
     for (auto& t : _worker_threads) {
         t.join();
     }
+    _worker_threads.clear();
+}
+
+internal::memory_prefaulter::~memory_prefaulter() {
+    join_threads();
 }
 
 void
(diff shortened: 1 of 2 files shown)


src/core/smp.cc
```diff
@@ -218,17 +219,27 @@ internal::memory_prefaulter::memory_prefaulter(const resource::resources& res, m
             }
             a.set(cpuset);
         }
-        _worker_threads.emplace_back(a, [this, &ranges, page_size, huge_page_size_opt] {
+        _worker_threads.emplace_back(a, [this, &alien, &ranges, page_size, huge_page_size_opt] {
+            ++_active_threads;
             work(ranges, page_size, huge_page_size_opt);
+            if (!--_active_threads) {
+                run_on(alien, 0, [this] () noexcept { join_threads(); });
```
michoecho, line 226: Is there no risk that this `join_threads()` call will happen after the `memory_prefaulter` is already destroyed?

michoecho: Yes, but's it's not nice to have implicit lifetime assumptions like that.
Not that this case is worth caring about, though.

michoecho (approved):
LGTM.

src/core/smp.cc
avikivity: In such cases it is unavoidable without strong compiler support. The setup is too hairy.

avikivity:
@xemul please review

PR #2682, opened 2025-03-07
tests: adapt the test.py script to build project tests in offline mode and accept arbitrary arguments for ctest

Comrade88:
this pr suggests adding to test.py the ability to build "offline" mode tests. i.e. exclude tests that require access to global resources. such tests always fail in a restricted environment and create a false impression of the project's performance.
also added the ability to feed ctest arbitrary parameters to gain more control over running tests.

### DIFF
--- a/test.py
+++ b/test.py
@@ -35,6 +35,8 @@ if __name__ == "__main__":
     parser.add_argument('--smp', '-c', action="store",default='2',type=int,help="Number of threads for multi-core tests")
     parser.add_argument('--verbose', '-v', action = 'store_true', default = False,
                         help = 'Verbose reporting')
+    parser.add_argument('--offline', action="store_true", default = False,
+                        help="Disable tests accessing internet")
     args = parser.parse_args()
 
     MODES = [args.mode] if args.mode else seastar_cmake.SUPPORTED_MODES
@@ -49,6 +51,7 @@ if __name__ == "__main__":
             tr(args.timeout, 'TEST_TIMEOUT'),
             tr(args.fast, 'EXECUTE_ONLY_FAST_TESTS'),
             tr(args.smp, 'UNIT_TEST_SMP'),
+            tr(not args.offline, 'ENABLE_TESTS_ACCESSING_INTERNET'),
             tr(args.jenkins, 'JENKINS', value_when_none=''),
         ]
 
--- a/test.py
+++ b/test.py
@@ -37,6 +37,7 @@ if __name__ == "__main__":
                         help = 'Verbose reporting')
     parser.add_argument('--offline', action="store_true", default = False,
                         help="Disable tests accessing internet")
+    parser.add_argument('ctest_forward', nargs='*', help="These parameters will be passed directly to ctest")
     args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


tchaikov:
Hi @Comrade88 , Thank you for this fix! You've correctly identified that several tests need proper handling with the `SEASTAR_TESTING_WITH_NETWORKING` macro.

While your approach works, I've taken the initiative to implement an alternative solution in PR #2683 that centralizes all the network-dependency flags in a single location. This should improve code readability and maintenance by avoiding scattered conditional checks throughout the codebase.

Could you please take a look at the alternative implementation when you have a chance. I believe it addresses the same issue you've identified while providing a more consolidated approach to managing network-dependent tests.

test.py
```diff
@@ -37,6 +37,8 @@
     parser.add_argument('--smp', '-c', action="store",default='2',type=int,help="Number of threads for multi-core tests")
     parser.add_argument('--verbose', '-v', action = 'store_true', default = False,
                         help = 'Verbose reporting')
+    parser.add_argument('--offline', action="store_true", default = False,
```
tchaikov, line 40: this change still makes sense even with #2683 .

tests/unit/dns_test.cc
```diff
@@ -36,6 +36,8 @@ using namespace seastar::net;
 
 static const sstring seastar_name = "seastar.io";
 
+#if SEASTAR_TESTING_WITH_NETWORKING
```
tchaikov, line 39: @Comrade88 ,  hi Galas, thank you for this fix! You've correctly identified that several tests need proper handling with the `SEASTAR_TESTING_WITH_NETWORKING` macro.

While your approach works, I've taken the initiative to implement an alternative solution in PR #2683 that centralizes all the network-dependency flags in a single location. This should improve code readability and maintenance by avoiding scattered conditional checks throughout the codebase.

Could you please take a look at the alternative implementation when you have a chance. I believe it addresses the same issue you've identified while providing a more consolidated approach to managing network-dependent tests.

test.py
```diff
@@ -26,6 +27,7 @@
     parser = argparse.ArgumentParser(description="Seastar test runner")
     parser.add_argument('--fast',  action="store_true", help="Run only fast tests")
     parser.add_argument('--name',  action="store", help="Run only test whose name contains given string")
+    parser.add_argument('--skip',  action="append", help="Skip tests whose name contains given string")
```
tchaikov, line 30: how about just exposing a flag allowing user to pass random parameters for `ctest`? like
```
./test.py -- --test-timeout 3600
```

Comrade88: good point. done.
also i've dropped the dns test patch in favor of PR #2683

Comrade88:
is there anything else that needs to be changed here?

xemul:
@Comrade88 , the provided PR description is too shallow, please write more profound one

tchaikov:
@Comrade88 currently the description of this PR puts:

> conditional and in offline mode

i guess Pavel suggested to make it more specific.

Comrade88:
@tchaikov , sorry, I'm a bit absent-minded and didn't realize @xemul suggested changing the description, not the title

PR #2683, opened 2025-03-08
tests: enable/disable internet-dependent tests at runtime

tchaikov:
Previously, some tests in dns_test.cc and tls_test.cc that required internet connectivity were conditionally compiled using `#if SEASTAR_TESTING_WITH_NETWORKING`. However, several other internet-dependent tests were not properly disabled,
causing test failures in environments without internet access.

This change:
- Replaces compile-time guards with runtime conditional execution using `boost::unit_test::enable_if<SEASTAR_TESTING_WITH_NETWORKING>` for tests requiring internet
- Uses `boost::unit_test::enable_if<!SEASTAR_TESTING_WITH_NETWORKING>` for alternative tests that provide equivalent coverage without internet access

This approach improves maintainability, enables running the test suite in offline environments, and makes internet dependencies more explicit in the test code.

### DIFF
--- a/tests/unit/dns_test.cc
+++ b/tests/unit/dns_test.cc
@@ -63,15 +63,20 @@ static future<> test_bad_name(dns_resolver::options opts) {
     });
 }
 
-SEASTAR_TEST_CASE(test_resolve_udp) {
+using enable_if_with_networking = boost::unit_test::enable_if<SEASTAR_TESTING_WITH_NETWORKING>;
+
+SEASTAR_TEST_CASE(test_resolve_udp,
+                  *enable_if_with_networking()) {
     return test_resolve(dns_resolver::options());
 }
 
-SEASTAR_TEST_CASE(test_bad_name_udp) {
+SEASTAR_TEST_CASE(test_bad_name_udp,
+                  *enable_if_with_networking()) {
     return test_bad_name(dns_resolver::options());
 }
 
-SEASTAR_TEST_CASE(test_timeout_udp) {
+SEASTAR_TEST_CASE(test_timeout_udp,
+                  *enable_if_with_networking()) {
     dns_resolver::options opts;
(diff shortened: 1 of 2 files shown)


tchaikov:
cc @Comrade88

Comrade88:
now you reapeat these typedef's in every test file. may be it is better to put them somewhere central or even introduce special SEASTAR_TEST_CASE macro wrappers?

tchaikov:
> now you reapeat these typedef's in every test file. may be it is better to put them somewhere central or even introduce special SEASTAR_TEST_CASE macro wrappers?

so far, we have two test files including this decorator. once it repeats itself in more places, probably we should indeed have a central place for it. what do you think?

Comrade88:
at the moment before my pr there was only one place in the project with a conditional disabling of online tests. I encountered a problem, found this place, understood what needed to be changed, and decided that this is how such issues are solved here.
later someone else will come, see the proposed typedefs and copy them to the next file, because this is how it is done here. in my opinion, it is better to immediately provide prospective users with a good example.

tests/unit/dns_test.cc
```diff
@@ -63,15 +63,20 @@ static future<> test_bad_name(dns_resolver::options opts) {
     });
 }
 
-SEASTAR_TEST_CASE(test_resolve_udp) {
+using enable_if_with_networking = boost::unit_test::enable_if<SEASTAR_TESTING_WITH_NETWORKING>;
+
+SEASTAR_TEST_CASE(test_resolve_udp,
```
xemul, line 68: Before the change this test (in fact, all the tests in dns_test.cc) ran regardless of SEASTAR_TESTING_WITH_NETWORKING, not they are conditionally skipped. These are tests that are described by

> several other internet-dependent tests were not properly disabled, causing test failures in environments without internet access.

, right?

tchaikov: @xemul correct.

PR #2684, opened 2025-03-08
map_reduce: prevent mapper exception from poisoning state

avikivity:
In map_reduce, if the mapper throws, we can end up with poisoned state. This is because the compiler can choose to move s->result before evaluating f.get(), so s->result gets broken (if the mapped type becomes invalid after move).

Fix by triggering the exception before the moving s->result, with a sequence point to force the ordering.

A unit test is added (which does fail with clang before the patch, though if the arguments are evaluated in reverse order, it can pass with a different compiler).

### DIFF
--- a/include/seastar/core/map_reduce.hh
+++ b/include/seastar/core/map_reduce.hh
@@ -191,7 +191,8 @@ map_reduce(Iterator begin, Iterator end, Mapper&& mapper, Initial initial, Reduc
     while (begin != end) {
         ret = futurize_invoke(s->mapper, *begin++).then_wrapped([s = s.get(), ret = std::move(ret)] (auto f) mutable {
             try {
-                s->result = s->reduce(std::move(s->result), f.get());
+                auto mapped_value = f.get(); // Extract exception before touching s->result
+                s->result = s->reduce(std::move(s->result), std::move(mapped_value));
                 return std::move(ret);
             } catch (...) {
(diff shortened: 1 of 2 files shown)


include/seastar/core/map_reduce.hh
```diff
@@ -191,7 +191,8 @@ map_reduce(Iterator begin, Iterator end, Mapper&& mapper, Initial initial, Reduc
     while (begin != end) {
         ret = futurize_invoke(s->mapper, *begin++).then_wrapped([s = s.get(), ret = std::move(ret)] (auto f) mutable {
             try {
-                s->result = s->reduce(std::move(s->result), f.get());
+                auto mapped_value = f.get(); // Extract exception before touching s->result
+                s->result = s->reduce(std::move(s->result), std::move(mapped_value));
```
tgrabiec, line 195: Don't we have the same problem if reduce() throws?

avikivity: Yes, good catch

avikivity:
v2:
 - rewrote to capture the exception in state
 - this enables handling exceptions from the reducer too
 - added unit test for throwing reducer (also fails before patch)
 - tightened test to ensure map_reduce actually throws

avikivity:
v3:
 -  ignore_ready_future if we already saw an exception
 - maintain task chain (`return std::move(ret)`) unconditionally

PR #2688, opened 2025-03-17
gate: add named_gate

bhalevy:
Allow specifying a name for the gate when created using a named exception factory,
to help debugging gate_closed_exception.

### DIFF
--- a/include/seastar/core/gate.hh
+++ b/include/seastar/core/gate.hh
@@ -53,6 +53,27 @@ public:
     }
 };
 
+/// Exception thrown when a \ref named_gate object has been closed
+/// by the \ref gate::close() method.
+SEASTAR_MODULE_EXPORT
+class named_gate_closed_exception : public gate_closed_exception {
+    static constexpr const char* _default_what = "named gate closed";
+    sstring _what;
+public:
+    named_gate_closed_exception(const sstring& name) noexcept : gate_closed_exception() {
+        if (!name.empty()) {
+            try {
+                _what = fmt::format("{} gate closed", name);
+            } catch (...) {
+                // Ignore
+            }
+        }
+    }
+    virtual const char* what() const noexcept override {
+        return _what.empty() ? _default_what : _what.c_str();
+    }
+};
+
 /// Facility to stop new requests, and to tell when existing requests are done.
 ///
 /// When stopping a service that serves asynchronous requests, we are faced with
@@ -109,7 +130,7 @@ public:
     /// a \ref gate_closed_exception is thrown.
     void enter() {
         if (!try_enter()) {
-            throw gate_closed_exception();
+            throw_closed_exception();
         }
     }
     /// Unregisters an in-progress request.
@@ -133,7 +154,7 @@ public:
     /// bail out of the long-running code if the gate is closed.
     void check() const {
         if (_stopped) {
-            throw gate_closed_exception();
+            throw_closed_exception();
         }
     }
     /// Closes the gate.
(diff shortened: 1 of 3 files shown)


bhalevy:
Intended use case: https://github.com/scylladb/scylladb/pull/23329

xemul:
Same problem for semaphores is solved with the help of "named" exception factory (59a6a47e8c4d57bf1610ba376bddad327525dcbf). It's worth making gates and semaphores symmetrical in this sense

bhalevy:
In v2 ([380d057](https://github.com/scylladb/seastar/commit/380d05768365657d3a7b0dab7122dce5393e9a93)):
- rebased
- used exception factory to distinguish between `gate` and `named_gate`.
  - added a templated `basic_gate(T value)` constructor as a helper to construct named gate as `named_gate("foo")`

xemul (changes requested):
LGTM, but
1. "gate: add named_gate" patch comment should be updated (patch title should probably be updated too)
2. CI shows compilation failures

bhalevy:
In v3 ([7c31e6f](https://github.com/scylladb/seastar/commit/7c31e6f53cf2994c7433571e999f7abb2d75920c)):
- rebased
- extracted `gate: templatize class basic_gate`
- renamed patch and edited git commit message
 - also this PR's title and description
- fixed build issues in debug mode

bhalevy:
FWIW, the strong named_gate type causes a lot of churn since it is not derived from `gate` :-/

bhalevy:
@xemul please see alternative design here: https://github.com/bhalevy/seastar/commit/4907d7df1f42f1812c0a2133b3b2eab43e5ce070
This one causes much less churn as named_gate is derived from gate in this model and other than using named_gate for defining members/variables so they can constructed with a name, they are indistinguishable from regular, unnamed, gates.

xemul:
> FWIW, the strong named_gate type causes a lot of churn since it is not derived from `gate` :-/

If this whole churn is isolated in gate.hh, I'd still go this way, it looks more solid and self-contained to me

bhalevy:
> > FWIW, the strong named_gate type causes a lot of churn since it is not derived from `gate` :-/
> 
> If this whole churn is isolated in gate.hh, I'd still go this way, it looks more solid and self-contained to me

No, it requires to change current user code that use gate::holder to be modified to named_gate::holder, since named_gate is strongly distinguishable from gate.
This it not the case with semaphores since there, sempahore_units is an independent template, allowing get_units to operate on any specialized semaphore with no change to user code.
But in contrast, gate::holder is defined as a nested type.

xemul:
I see :( Semaphores+units and gates+holders are so much alike, so it's worth having them _look_ alike too.

Can we take your previous approach with basic_gate, move `seastar::gate::holder` to be `seastar::gate_holder` and please the Backward Compatibility god with `using holder = gate_holder;` statement in `class gate`? Users of `gate` that will be switched to `named_gate` will need to patch their `gate::holder` variables to become `gate_holder` (or `named_gate::holder`) anyway, but AFAICS lots of holders are declared with `auto` so it shouldn't be too disastrous. WDYT?

bhalevy:
> I see :( Semaphores+units and gates+holders are so much alike, so it's worth having them _look_ alike too.
> 
> Can we take your previous approach with basic_gate, move `seastar::gate::holder` to be `seastar::gate_holder` and please the Backward Compatibility god with `using holder = gate_holder;` statement in `class gate`? Users of `gate` that will be switched to `named_gate` will need to patch their `gate::holder` variables to become `gate_holder` (or `named_gate::holder`) anyway, but AFAICS lots of holders are declared with `auto` so it shouldn't be too disastrous. WDYT?

Hmm, that might work.
I'll take a stab at it.

bhalevy:
In v4 ([70f8876](https://github.com/scylladb/seastar/commit/70f88761166faf6959a1dc4664e72c8308e2ce20)):
- Further simplified version using inheritance (named_gate derived from gate, named_gate_closed_exception derived from gate_closed_exception)

The main advantage is that named_gate is a gate (in disguise :)
and named_gate_closed_exception is a gate_closed_exception
so the existing code needs only to change gate to named_gate for constructing the named_gate object while the rest of the logic, holding the gate, intercepting exceptions, etc. needs no change
and the exception generated does produce the gate name for named_gate using its virtual what() method.

I think this is a cleaner object oriented design.

PR #2691, opened 2025-03-20
reactor: more info, robustness on segfault

travisdowns:
On segfault we execute a handler that provides information including a backtrace. This currently emits all information in a single write call after collecting it in a buffer. If anything goes wrong, e.g., the backtrace() call itself crashes, then no information will be emitted. The backtrace() call is not signal safe in theory, and in practice the situation seems mixed as to its safety. So it not unlikely that situations may arise where no output can be emitted on SIGSEGV.

Because we catch the signal and then re-raise it using pthread_kill, the specific information about the IP is lost in re-raise: this prevents the line in syslog which usually captures information about segfaults from appearing at all. So we may be left without useful information after a crash.

In this change, we emit additional information before the backtrace() which is not likely to have any problem, and we emit each as separate write(2) calls so if there is a failure at any point we at least have the information emitted up to that point.

After this, the start of the output on segfault looks like so:

```
Segmentation fault, si_pid: 0, si_addr: 0000000000000000, ip: 0000579ba959751f
Segmentation fault resolved ip: 0000000005e2751f in [0000579ba3770000+000000000e3f98d8] 
Segmentation fault on shard 0, in scheduling group admin.
Backtrace:
 ....
```


src/core/reactor.cc
```diff
@@ -3995,12 +3995,58 @@ static void reraise_signal(int signo) {
...
+    print_zero_padded_hex_safe(f.addr);
+    print_safe(" in ");
+    print_safe(f.so->name.c_str());
+    print_safe("[");
+    print_zero_padded_hex_safe(f.so->begin);
+    print_safe("+");
+    print_zero_padded_hex_safe(f.so->end - f.so->begin);
+    print_safe("]");
+    
+    print_with_backtrace("\nSegmentation fault");
```
xemul, line 4045: Why not append this `\n` to the previous line? Like `print_safe("]\n");`

travisdowns:
Thanks for the  merge @xemul. This was related to https://github.com/scylladb/seastar/issues/2697, and I have a follow on PR to add more robustness to the fault handler.

PR #2692, opened 2025-03-21
smp: prefault: decouple _stop_request from join_threads

tomershafir:
Requesting the prefaulter to stop is logically separate from joining the threads. Thus, join_threads shouldn't impose a stop request. In the current implementation, it has no effect, but it can cause problems in the future if join _threads is used differently (like stop before finish where not intended).


tomershafir:
As I wrote, it has no effect now, but I dont think that `join_threads` should stop the threads. There may be a use case where you would want to just join the threads early, before they finish, and stopping them early would leave some ranges unscanned. I think its better logical separation.

PR #2693, opened 2025-03-24
Improve log output in reactor-stalled situations

radoslawcybulski:
- fix `backtrace_buffer` class to print log on destruction, if the log wasn't printed before (this is what comment suggests should happen)
- `maybe_report_kernel_trace` function will be now called with `backtrace_buffer` object allocated by caller. This will "glue" kernel trace to the rest of the log, making sure the output stays together in the log file.
- `maybe_report_kernel_trace` function will no longer print empty `kernel callstack: ` messages.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -848,7 +848,6 @@ static void print_with_backtrace(backtrace_buffer& buf, bool oneline) noexcept {
     buf.append_backtrace_oneline();
     buf.append("\n");
   }
-    buf.flush();
 }
 
 static void print_with_backtrace(const char* cause, bool oneline = false) noexcept {
@@ -1223,7 +1222,6 @@ void cpu_stall_detector::report_suppressions(sched_clock::time_point now) {
             buf.append(" on shard ");
             buf.append_decimal(_shard_id);
             buf.append("\n");
-            buf.flush();
         }
         reset_suppression_state(now);
     }
@@ -1360,7 +1358,6 @@ cpu_stall_detector_linux_perf_event::maybe_report_kernel_trace(backtrace_buffer&
             }
             buf.append("\n");
         }
-        buf.flush();
     };
 }
 
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -767,9 +767,16 @@ class backtrace_buffer {
     unsigned _pos = 0;
     char _buf[_max_size];
 public:
+    backtrace_buffer() = default;
+    ~backtrace_buffer() {
+        flush();
+    }
+
     void flush() noexcept {
-        print_safe(_buf, _pos);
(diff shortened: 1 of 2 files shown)


xemul:
This PR does several things (and PR comment explicitly states that). Please, split this large patch into smaller patches each doing one item from the PR comment list

radoslawcybulski:
Updated, split into multiple commits.

src/core/reactor.cc
```diff
@@ -1353,14 +1351,13 @@ cpu_stall_detector_linux_perf_event::maybe_report_kernel_trace(backtrace_buffer&
 
         auto nr = reader.read_u64();
         if (nr > 0) {
-        buf.append("kernel callstack:");
```
xemul, line 1349: This indentation fix shouldn't be here, but rather in the patch that broke it

src/core/reactor.cc
```diff
@@ -1345,13 +1345,14 @@ cpu_stall_detector_linux_perf_event::maybe_report_kernel_trace() {
         }
 
         auto nr = reader.read_u64();
-        backtrace_buffer buf;
```
xemul, line 1348: How does this patch compile after removal of a variable that's used below?

radoslawcybulski: Updated. Build all commits.

src/core/reactor.cc
radoslawcybulski: Fixed.

src/core/reactor.cc
```diff
@@ -1442,7 +1441,7 @@ void cpu_stall_detector::generate_trace() {
     buf.append_decimal(uint64_t(delta / 1ms));
     buf.append(" ms");
     print_with_backtrace(buf, _config.oneline);
-    maybe_report_kernel_trace();
+    maybe_report_kernel_trace(buf);
```
xemul, line 1454: The patch description says:
> This will glue kernel trace to whatever raport a caller is generating and print all of it in a single blob to the log sink.

In fact it won't, because `print_with_backtrace()` would flush the buffer before it's passed down to maybe_report_kernel_trace, so printing will be anyway split into two parts. Next patch that moves flush to destructor really makes it happen

src/core/reactor.cc
```diff
@@ -1360,7 +1358,6 @@ cpu_stall_detector_linux_perf_event::maybe_report_kernel_trace(backtrace_buffer&
             }
             buf.append("\n");
         }
-        buf.flush();
```
xemul, line 1355: Make flush() method private, since the buffer is now fully automatic in this sense

radoslawcybulski: The user might still have a need to flush the output manually.

PR #2695, opened 2025-03-25
iotune: Add cli flag to force io depth

StephanDollberg:
This is useful for testing lower io depths that might be more common in real life scenarios and for easy comparing with fio.

### DIFF
--- a/apps/iotune/iotune.cc
+++ b/apps/iotune/iotune.cc
@@ -23,6 +23,7 @@
  */
 #include <iostream>
 #include <chrono>
+#include <optional>
 #include <random>
 #include <memory>
 #include <ranges>
@@ -100,6 +101,7 @@ struct evaluation_directory {
     sstring _name;
     // We know that if we issue more than this, they will be blocked on linux anyway.
     unsigned _max_iodepth = 0;
+    unsigned _force_io_depth;
     uint64_t _available_space;
     uint64_t _min_data_transfer_size = 512;
     unsigned _disks_per_array = 0;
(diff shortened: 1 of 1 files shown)


apps/iotune/iotune.cc
```diff
@@ -143,11 +145,15 @@ struct evaluation_directory {
             iotune_logger.error("Error while parsing sysfs. Will continue with guessed values: {}", se.what());
             _max_iodepth = 128;
         }
+        if (_force_io_depth) {
+            _max_iodepth = *_force_io_depth;
```
xemul, line 149: I guess with _force_io_depth == 0 the whole thing will get stuck ... (see my other comment below for continuation of this guess)

apps/iotune/iotune.cc
```diff
@@ -736,6 +743,8 @@ int main(int ac, char** av) {
             auto accuracy = configuration["accuracy"].as<unsigned>();
             auto saturation = configuration["saturation"].as<sstring>();
             auto random_io_buffer_size = configuration["random-io-buffer-size"].as<unsigned>();
+            auto& io_depth_conf = configuration["force-io-depth"];
+            std::optional<unsigned> force_io_depth = io_depth_conf.empty() ? std::nullopt : std::make_optional(io_depth_conf.as<unsigned>());
```
xemul, line 747: Probably plain `unsigned` value will work if dedicating the value of `0` to "auto-detect" meaning. With that the declaration of "force-io-depth" parameter may have `->default_value(0)` and the whole thing gets much simpler and automatically handles the invalid `--force-io-depth 0` case

StephanDollberg: Sure will change.

StephanDollberg: Have pushed a change with the suggested change. Let me know what you think.

PR #2705, opened 2025-03-29
Rework scheduling group API

xemul:
The core part is class scheduling_group which is really a "handle" with sched group ID that can be easily coped around. To create, destroy and rename one there are three standalone functions in seastar namespace. To do their work, the functions are declared as friends for scheduling_group class itself.

This friendship is not really required. In order to take control over scheduling_group class lifetime, those functions should rather be static methods of that class and that's it. This is what this PR does. Old API methods are deprecated. As a nice side effect, two creation methods are squashed into one with default "shortname" argument.

Other than being friends to scheduling_group, the API methods are also friends of reactor class. Also reactor has scheduling_group itself as a friend. This all is to manipulate reactor's task queues. Once API methods are turned in scheduling_group static methods, that friendship becomes not needed and can be removed.

### DIFF
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -1152,7 +1152,7 @@ int main(int ac, char** av) {
                     return make_ready_future<>();
                 }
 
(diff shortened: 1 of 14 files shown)


avikivity:
I don't understand the deprecations. The API is simple and easy to use.

PR #2707, opened 2025-03-31
Add source location to task and tasktrace object

### LINKED ISSUE #2381: Looking for stuck/deadlocked fibers could be easier
In ScyllaDB, it happens from time to time that some startup/shutdown/maintenance fiber gets stuck. The first step to debugging a problem like this is finding out what fiber got stuck and where, and this usually involves repeatedly adding debug logs and hoping that the bug reproduces again. But that's time-consuming, and not guaranteed to succeed. It would be better to be able to find the stuck fiber in the coredump.

It's already possible to find chains of awaiting `seastar::task`s in the coredump, but AFAIK it's not possible to guess what points in the source code they correspond to. For coroutines, we can find the relevant coroutine, but not point out the relevant `co_await` inside it. For `then` continuations, we can try to guess the relevant function based on the *type* of the continuation, but it's hard (if possible at all) to do that reliably.

So maybe we could, for example, capture the program counter (`rip`) at the call sites of `then` (and its variants) and `co_await`, and put it into the awaiting task (continuation/coroutine)? Combined with debug info, this should allow us to see the source locations of the awaiting continuation points.

And it should be cheap enough. (Probably.)

radoslawcybulski:
Adds source location of next resume point to `task` objects. When constructing `tasktrace` object, source location of resume points will be added to it. When printing, the source location of next resume point will be printed as well, hopefully improving debug experience.

Example:
```
0x2b3899 /home/y/work/scylladb-test/build/Debug/seastar/libseastar.so+0x454c364 ...
   --------
   test_app/main.cpp:15:15
   seastar::continuation<seastar::internal::promise_base_with_type<void>, ...
```

Partially fixes https://github.com/scylladb/seastar/issues/2381

### DIFF
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -43,7 +43,6 @@ void free_sized(void* ptr, size_t size);
 
 #endif
 
-
 namespace seastar {
 
 namespace internal {
@@ -194,8 +193,8 @@ public:
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
-        hndl.promise().update_resume_point(sl);
+    void await_suspend(std::coroutine_handle<U> hndl SEASTAR_COROUTINE_LOC_PARAM) noexcept {
+        SEASTAR_COROUTINE_LOC_STORE(hndl.promise());
         if (!CheckPreempt || !_future.available()) {
             _future.set_coroutine(hndl.promise());
         } else {
@@ -220,8 +219,8 @@ public:
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
-        hndl.promise().update_resume_point(sl);
+    void await_suspend(std::coroutine_handle<U> hndl SEASTAR_COROUTINE_LOC_PARAM) noexcept {
+        SEASTAR_COROUTINE_LOC_STORE(hndl.promise());
         if (!CheckPreempt || !_future.available()) {
             _future.set_coroutine(hndl.promise());
         } else {
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -31,6 +31,7 @@
 #include <coroutine>
 #include <new>
 #include <cstdlib>
+#include <source_location>
 
 #if !(__GLIBC__ == 2 && __GLIBC_MINOR__ >= 43) && !(__GLIBC__ > 2)
 
@@ -193,7 +194,8 @@ public:
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl) noexcept {
+    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
+        hndl.promise().update_resume_point(sl);
         if (!CheckPreempt || !_future.available()) {
             _future.set_coroutine(hndl.promise());
         } else {
@@ -218,7 +220,8 @@ public:
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl) noexcept {
+    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
+        hndl.promise().update_resume_point(sl);
         if (!CheckPreempt || !_future.available()) {
             _future.set_coroutine(hndl.promise());
         } else {
--- a/include/seastar/core/task.hh
+++ b/include/seastar/core/task.hh
@@ -25,12 +25,16 @@
 #include <seastar/util/backtrace.hh>
 
 #include <utility>
+#include <source_location>
 
 namespace seastar {
 
 class task {
+    std::source_location _resume_point = {};
+
 protected:
     scheduling_group _sg;
+
 private:
 #ifdef SEASTAR_TASK_BACKTRACE
     shared_backtrace _bt;
@@ -50,6 +54,8 @@ public:
     virtual void run_and_dispose() noexcept = 0;
     /// Returns the next task which is waiting for this task to complete execution, or nullptr.
     virtual task* waiting_task() noexcept = 0;
+    void update_resume_point(std::source_location sl) { _resume_point = sl; }
+    auto get_resume_point() const { return _resume_point; }
     scheduling_group group() const { return _sg; }
 #ifdef SEASTAR_TASK_BACKTRACE
     void make_backtrace() noexcept;
--- a/include/seastar/util/backtrace.hh
+++ b/include/seastar/util/backtrace.hh
@@ -151,7 +151,17 @@ public:
 // and information about the chain of tasks waiting for the current operation to complete.
 class tasktrace {
 public:
-    using entry = std::variant<shared_backtrace, task_entry>;
+    struct entry {
+        std::variant<shared_backtrace, task_entry> backtrace_or_task_entry;
+        std::source_location resume_point;
+
+        bool operator == (const entry &other) const {
+            return backtrace_or_task_entry == other.backtrace_or_task_entry
+                    && resume_point.line() == other.resume_point.line()
+                    && resume_point.column() == other.resume_point.column()
+                    && std::string_view{ resume_point.file_name() } == std::string_view{ other.resume_point.file_name() };
+        }
+    };
     using vector_type = boost::container::static_vector<entry, 16>;
 private:
     simple_backtrace _main;
--- a/include/seastar/util/std-compat.hh
+++ b/include/seastar/util/std-compat.hh
@@ -56,3 +56,18 @@ using source_location
     = std::source_location;
 
 }
+
+#if defined(__GNUC__) && !defined(__clang__)
+// GCC Workaround: Strip source_location to prevent ICE during RTL expansion.
+// See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114675
+#define SEASTAR_COROUTINE_LOC_PARAM
+#define SEASTAR_COROUTINE_LOC_STORE(promise) (void)0
+#else
+// Standard/Clang: Capture source location naturally.
+// Includes the leading comma to mix cleanly into argument lists.
+#define SEASTAR_COROUTINE_LOC_PARAM \
+    , std::source_location sl = std::source_location::current()
+
+#define SEASTAR_COROUTINE_LOC_STORE(promise) \
+    (promise).update_resume_point(sl)
+#endif
--- a/src/util/backtrace.cc
+++ b/src/util/backtrace.cc
@@ -30,6 +30,7 @@ module;
 #include <cerrno>
 #include <cstring>
 #include <iostream>
+#include <source_location>
 #include <variant>
 #include <vector>
 #include <boost/container/static_vector.hpp>
@@ -152,10 +153,10 @@ tasktrace current_tasktrace() noexcept {
             hash *= 31;
             if (bt) {
                 hash ^= bt->hash();
-                prev.push_back(bt);
+                prev.push_back({ bt, tsk->get_resume_point() });
             } else {
                 const std::type_info& ti = typeid(*tsk);
-                prev.push_back(task_entry(ti));
+                prev.push_back({ task_entry(ti), tsk->get_resume_point() });
                 hash ^= ti.hash_code();
             }
             tsk = tsk->waiting_task();
@@ -204,8 +205,12 @@ auto formatter<seastar::tasktrace>::format(const seastar::tasktrace& b, format_c
     -> decltype(ctx.out()) {
     auto out = ctx.out();
     out = fmt::format_to(out, "{}", b._main);
-    for (auto&& e : b._prev) {
+    for(const auto & [ e, resume_loc ] : b._prev) {
         out = fmt::format_to(out,  "\n   --------");
+
+        if (resume_loc.file_name()[0] != 0) {
+            out = fmt::format_to(out, "\n   {}:{}:{}", resume_loc.file_name(), resume_loc.line(), resume_loc.column());
+        }
         out = std::visit(seastar::make_visitor(
             [&] (const seastar::shared_backtrace& sb) {
                 return fmt::format_to(out,  "\n{}", sb);
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -30,6 +30,7 @@
 #include <stdexcept>
 #include <type_traits>
 #include <utility>
+#include <source_location>
 
 #include <seastar/core/task.hh>
 #include <seastar/core/thread_impl.hh>
@@ -1253,7 +1254,7 @@ private:
         future_base::schedule(tws, &tws->_state);
     }
     template <typename Pr, typename Func, typename Wrapper>
-    void schedule(Pr&& pr, Func&& func, Wrapper&& wrapper) noexcept {
+    void schedule(Pr&& pr, Func&& func, Wrapper&& wrapper, std::source_location sl) noexcept {
         static_assert(internal::expect_only_rvalue_refs<Func>);
         // If this new throws a std::bad_alloc there is nothing that
         // can be done about it. The corresponding future is not ready
@@ -1261,6 +1262,7 @@ private:
         // noexcept, it will call std::terminate if new throws.
         memory::scoped_critical_alloc_section _;
         auto tws = new continuation<Pr, Func, Wrapper, T>(std::move(pr), std::move(func), std::move(wrapper));
+        tws->update_resume_point(sl);
         // In a debug build we schedule ready futures, but not in
         // other build modes.
 #ifdef SEASTAR_DEBUG
@@ -1383,14 +1385,14 @@ public:
     requires std::invocable<Func, T>
                  || (std::same_as<void, T> && std::invocable<Func>)
     Result
-    then(Func&& func) noexcept {
+    then(Func&& func, std::source_location sl = std::source_location::current()) noexcept {
       // Avoid having to special-case lvalue-references downstream by converting
       // them to an rvalue reference here.
       if constexpr (std::is_lvalue_reference_v<Func>) {
-        return then(func_to_rvalue(func));
+        return then(func_to_rvalue(func), sl);
       } else {
 #ifndef SEASTAR_TYPE_ERASE_MORE
-        return then_impl(std::move(func));
+        return then_impl(std::move(func), sl);
 #else
         using func_type = typename internal::future_result<Func, T>::func_type;
         noncopyable_function<func_type> ncf;
@@ -1400,7 +1402,7 @@ public:
                 return futurize_invoke(std::move(func), std::forward<decltype(args)>(args)...);
             });
         }
-        return then_impl(std::move(ncf));
+        return then_impl(std::move(ncf), sl);
 #endif
       }
     }
@@ -1427,14 +1429,14 @@ public:
     template <typename Func, typename Result = futurize_t<internal::result_of_apply_t<Func, T>>>
     requires ::seastar::CanApplyTuple<Func, T>
     Result
-    then_unpack(Func&& func) noexcept {
+    then_unpack(Func&& func, std::source_location sl = std::source_location::current()) noexcept {
       if constexpr (std::is_lvalue_reference_v<Func>) {
         return then_unpack(func_to_rvalue(func));
       } else {
         return then([func = std::forward<Func>(func)] (T&& tuple) mutable {
             // sizeof...(tuple) is required to be 1
             return std::apply(func, std::move(tuple));
-        });
+        }, sl);
       }
     }
 
@@ -1442,7 +1444,7 @@ private:
 
     // Keep this simple so that Named Return Value Optimization is used.
     template <typename Func, typename Result>
-    Result then_impl_nrvo(Func&& func) noexcept {
+    Result then_impl_nrvo(Func&& func, std::source_location sl) noexcept {
         using futurator = futurize<internal::future_result_t<Func, T>>;
         typename futurator::type fut(future_for_get_promise_marker{});
         using pr_type = decltype(fut.get_promise());
@@ -1457,13 +1459,13 @@ private:
                     return internal::future_invoke(std::move(func), std::move(state).get_value());
                 });
             }
-        });
+        }, sl);
         return fut;
     }
 
     template <typename Func, typename Result = futurize_t<internal::future_result_t<Func, T>>>
     Result
-    then_impl(Func&& func) noexcept {
+    then_impl(Func&& func, std::source_location sl) noexcept {
         static_assert(internal::expect_only_rvalue_refs<Func>);
 #ifndef SEASTAR_DEBUG
         using futurator = futurize<internal::future_result_t<Func, T>>;
@@ -1473,7 +1475,7 @@ private:
             return futurator::invoke(std::forward<Func>(func), get_available_state_ref().take_value());
         }
 #endif
-        return then_impl_nrvo<Func, Result>(std::forward<Func>(func));
+        return then_impl_nrvo<Func, Result>(std::forward<Func>(func), sl);
     }
 
 public:
@@ -1494,25 +1496,25 @@ public:
     ///         to the eventual value of this future.
     template <std::invocable<future> Func, typename FuncResult = std::invoke_result_t<Func, future>>
     futurize_t<FuncResult>
-    then_wrapped(Func&& func) & noexcept {
+    then_wrapped(Func&& func, std::source_location sl = std::source_location::current()) & noexcept {
       // Avoid having to special-case lvalue-references downstream by converting
       // them to an rvalue reference here.
       if constexpr (std::is_lvalue_reference_v<Func>) {
-        return then_wrapped(func_to_rvalue(func));
+        return then_wrapped(func_to_rvalue(func), sl);
       } else {
-        return then_wrapped_maybe_erase<false, FuncResult>(std::forward<Func>(func));
+        return then_wrapped_maybe_erase<false, FuncResult>(std::forward<Func>(func), sl);
       }
     }
 
     template <std::invocable<future&&> Func, typename FuncResult = std::invoke_result_t<Func, future&&>>
     futurize_t<FuncResult>
-    then_wrapped(Func&& func) && noexcept {
+    then_wrapped(Func&& func, std::source_location sl = std::source_location::current()) && noexcept {
       // Avoid having to special-case lvalue-references downstream by converting
       // them to an rvalue reference here.
       if constexpr (std::is_lvalue_reference_v<Func>) {
-        return std::move(*this).then_wrapped(func_to_rvalue(func));
+        return std::move(*this).then_wrapped(func_to_rvalue(func), sl);
       } else {
-        return then_wrapped_maybe_erase<true, FuncResult>(std::forward<Func>(func));
+        return then_wrapped_maybe_erase<true, FuncResult>(std::forward<Func>(func), sl);
       }
     }
 
@@ -1520,10 +1522,10 @@ private:
 
     template <bool AsSelf, typename FuncResult, typename Func>
     futurize_t<FuncResult>
-    then_wrapped_maybe_erase(Func&& func) noexcept {
+    then_wrapped_maybe_erase(Func&& func, std::source_location sl) noexcept {
         static_assert(internal::expect_only_rvalue_refs<Func>);
 #ifndef SEASTAR_TYPE_ERASE_MORE
-        return then_wrapped_common<AsSelf, FuncResult>(std::forward<Func>(func));
+        return then_wrapped_common<AsSelf, FuncResult>(std::forward<Func>(func), sl);
 #else
         using futurator = futurize<FuncResult>;
         using WrapFuncResult = typename futurator::type;
@@ -1534,14 +1536,14 @@ private:
                 return futurator::invoke(std::move(func), std::move(f));
             });
         }
-        return then_wrapped_common<AsSelf, WrapFuncResult>(std::move(ncf));
+        return then_wrapped_common<AsSelf, WrapFuncResult>(std::move(ncf), sl);
 #endif
     }
 
     // Keep this simple so that Named Return Value Optimization is used.
     template <typename FuncResult, typename Func>
     futurize_t<FuncResult>
-    then_wrapped_nrvo(Func&& func) noexcept {
+    then_wrapped_nrvo(Func&& func, std::source_location sl) noexcept {
         using futurator = futurize<FuncResult>;
         typename futurator::type fut(future_for_get_promise_marker{});
         using pr_type = decltype(fut.get_promise());
@@ -1549,14 +1551,14 @@ private:
             futurator::satisfy_with_result_of(std::move(pr), [&func, &state] {
                 return std::move(func)(future(std::move(state)));
             });
-        });
+        }, sl);
         return fut;
     }
 
 
     template <bool AsSelf, typename FuncResult, typename Func>
     futurize_t<FuncResult>
-    then_wrapped_common(Func&& func) noexcept {
+    then_wrapped_common(Func&& func, std::source_location sl) noexcept {
 #ifndef SEASTAR_DEBUG
         using futurator = futurize<FuncResult>;
         if (available()) {
@@ -1570,7 +1572,7 @@ private:
             }
         }
 #endif
-        return then_wrapped_nrvo<FuncResult, Func>(std::forward<Func>(func));
+        return then_wrapped_nrvo<FuncResult, Func>(std::forward<Func>(func), sl);
     }
 
     void forward_to(internal::promise_base_with_type<T>&& pr) noexcept {
@@ -1624,13 +1626,13 @@ public:
      * See then() for lifetime and call semantics.
      */
     template <std::invocable Func>
-    future<T> finally(Func&& func) noexcept {
+    future<T> finally(Func&& func, std::source_location sl = std::source_location::current()) noexcept {
       // Avoid having to special-case lvalue-references downstream by converting
       // them to an rvalue reference here.
       if constexpr (std::is_lvalue_reference_v<Func>) {
-        return finally(func_to_rvalue(func));
+        return finally(func_to_rvalue(func), sl);
       } else {
-        return then_wrapped(finally_body<Func, is_future<std::invoke_result_t<Func>>::value>(std::forward<Func>(func)));
+        return then_wrapped(finally_body<Func, is_future<std::invoke_result_t<Func>>::value>(std::forward<Func>(func)), sl);
       }
     }
 
@@ -1679,24 +1681,24 @@ public:
     ///
     /// Terminates the entire program is this future resolves
     /// to an exception.  Use with caution.
-    future<> or_terminate() noexcept {
+    future<> or_terminate(std::source_location sl = std::source_location::current()) noexcept {
         return then_wrapped([] (auto&& f) {
             try {
                 f.get();
             } catch (...) {
                 engine_exit(std::current_exception());
             }
-        });
+        }, sl);
     }
 
     /// \brief Discards the value carried by this future.
     ///
     /// Converts the future into a no-value \c future<>, by
     /// ignoring any result.  Exceptions are propagated unchanged.
-    future<> discard_result() noexcept {
+    future<> discard_result(std::source_location sl = std::source_location::current()) noexcept {
         // We need the generic variadic lambda, below, because then() behaves differently
         // when value_type is when_all_succeed_tuple
-        return then([] (auto&&...) {});
+        return then([] (auto&&...) {}, sl);
     }
 
     /// \brief Handle the exception carried by this future.
@@ -1717,11 +1719,11 @@ public:
                     || (std::tuple_size_v<tuple_type> == 0 && std::is_invocable_r_v<void, Func, std::exception_ptr>)
                     || (std::tuple_size_v<tuple_type> == 1 && std::is_invocable_r_v<T, Func, std::exception_ptr>)
                     || (std::tuple_size_v<tuple_type> > 1 && std::is_invocable_r_v<tuple_type ,Func, std::exception_ptr>)
-    future<T> handle_exception(Func&& func) noexcept {
+    future<T> handle_exception(Func&& func, std::source_location sl = std::source_location::current()) noexcept {
       // Avoid having to special-case lvalue-references downstream by converting
       // them to an rvalue reference here.
       if constexpr (std::is_lvalue_reference_v<Func>) {
-        return handle_exception(func_to_rvalue(func));
+        return handle_exception(func_to_rvalue(func), sl);
       } else {
         return then_wrapped([func = std::forward<Func>(func)]
                              (auto&& fut) mutable -> future<T> {
@@ -1730,7 +1732,7 @@ public:
             } else {
                 return futurize_invoke(func, fut.get_exception());
             }
-        });
+        }, sl);
       }
     }
 
@@ -1745,11 +1747,11 @@ public:
     /// If exception, that future holds, does not match func parameter type
     /// it is propagated as is.
     template <typename Func>
-    future<T> handle_exception_type(Func&& func) noexcept {
+    future<T> handle_exception_type(Func&& func, std::source_location sl = std::source_location::current()) noexcept {
       // Avoid having to special-case lvalue-references downstream by converting
       // them to an rvalue reference here.
       if constexpr (std::is_lvalue_reference_v<Func>) {
-        return handle_exception_type(func_to_rvalue(func));
+        return handle_exception_type(func_to_rvalue(func), sl);
       } else {
         using trait = function_traits<Func>;
         static_assert(trait::arity == 1, "func can take only one parameter");
@@ -1761,7 +1763,7 @@ public:
             } catch(ex_type& ex) {
                 return futurize_invoke(func, ex);
             }
-        });
+        }, sl);
       }
     }
 
--- a/include/seastar/coroutine/all.hh
+++ b/include/seastar/coroutine/all.hh
@@ -161,7 +161,9 @@ private:
                 return (... && futures.available());
             }, state._futures);
         }
-        void await_suspend(coroutine_handle_t h) {
+        template <typename U>
+        void await_suspend(std::coroutine_handle<U> h, std::source_location sl = std::source_location::current()) {
+            h.promise().update_resume_point(sl);
             when_ready = h;
             process<0>();
         }
--- a/include/seastar/coroutine/all.hh
+++ b/include/seastar/coroutine/all.hh
@@ -162,8 +162,8 @@ private:
             }, state._futures);
         }
         template <typename U>
-        void await_suspend(std::coroutine_handle<U> h, std::source_location sl = std::source_location::current()) {
-            h.promise().update_resume_point(sl);
+        void await_suspend(std::coroutine_handle<U> h SEASTAR_COROUTINE_LOC_PARAM) {
+            SEASTAR_COROUTINE_LOC_STORE(h.promise());
             when_ready = h;
             process<0>();
         }
--- a/include/seastar/coroutine/as_future.hh
+++ b/include/seastar/coroutine/as_future.hh
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <seastar/core/coroutine.hh>
+#include <source_location>
 
 namespace seastar {
 
@@ -42,7 +43,8 @@ public:
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl) noexcept {
+    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
+        hndl.promise().update_resume_point(sl);
         if (!CheckPreempt || !_future.available()) {
             _future.set_coroutine(hndl.promise());
         } else {
--- a/include/seastar/coroutine/as_future.hh
+++ b/include/seastar/coroutine/as_future.hh
@@ -43,8 +43,8 @@ public:
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
-        hndl.promise().update_resume_point(sl);
+    void await_suspend(std::coroutine_handle<U> hndl SEASTAR_COROUTINE_LOC_PARAM) noexcept {
+        SEASTAR_COROUTINE_LOC_STORE(hndl.promise());
         if (!CheckPreempt || !_future.available()) {
             _future.set_coroutine(hndl.promise());
         } else {
--- a/include/seastar/coroutine/exception.hh
+++ b/include/seastar/coroutine/exception.hh
@@ -24,6 +24,7 @@
 #include <seastar/core/future.hh>
 #include <coroutine>
 #include <exception>
+#include <source_location>
 
 namespace seastar {
 
@@ -42,8 +43,9 @@ struct exception_awaiter {
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl) noexcept {
-      execute_involving_handle_destruction_in_await_suspend([hndl, eptr = std::move(eptr)] () mutable {
+    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
+      execute_involving_handle_destruction_in_await_suspend([sl, hndl, eptr = std::move(eptr)] () mutable {
+        hndl.promise().update_resume_point(sl);
         hndl.promise().set_exception(std::move(eptr));
         hndl.destroy();
       });
--- a/include/seastar/coroutine/exception.hh
+++ b/include/seastar/coroutine/exception.hh
@@ -43,9 +43,9 @@ struct exception_awaiter {
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
-      execute_involving_handle_destruction_in_await_suspend([sl, hndl, eptr = std::move(eptr)] () mutable {
-        hndl.promise().update_resume_point(sl);
+    void await_suspend(std::coroutine_handle<U> hndl SEASTAR_COROUTINE_LOC_PARAM) noexcept {
+      SEASTAR_COROUTINE_LOC_STORE(hndl.promise());
+      execute_involving_handle_destruction_in_await_suspend([hndl, eptr = std::move(eptr)] () mutable {
         hndl.promise().set_exception(std::move(eptr));
         hndl.destroy();
       });
--- a/include/seastar/coroutine/generator.hh
+++ b/include/seastar/coroutine/generator.hh
@@ -186,8 +186,8 @@ struct generator_promise_base<Yielded>::yield_awaiter final {
         return false;
     }
     template <typename Promise>
-    std::coroutine_handle<> await_suspend(std::coroutine_handle<Promise> producer, std::source_location sl = std::source_location::current()) noexcept {
-        producer.promise().update_resume_point(sl);
+    std::coroutine_handle<> await_suspend(std::coroutine_handle<Promise> producer SEASTAR_COROUTINE_LOC_PARAM) noexcept {
+        SEASTAR_COROUTINE_LOC_STORE(producer.promise());
         _promise->_waiting_task = &producer.promise();
         if (seastar::need_preempt()) {
             auto consumer = std::coroutine_handle<seastar::task>::from_address(
@@ -211,11 +211,10 @@ struct generator_promise_base<Yielded>::copy_awaiter final {
         return false;
     }
     template <typename Promise>
-    std::coroutine_handle<> await_suspend(std::coroutine_handle<Promise> producer, std::source_location sl = std::source_location::current()) noexcept {
-        producer.promise().update_resume_point(sl);
+    std::coroutine_handle<> await_suspend(std::coroutine_handle<Promise> producer SEASTAR_COROUTINE_LOC_PARAM) noexcept {
+        SEASTAR_COROUTINE_LOC_STORE(producer.promise());
         _value_ptr = std::addressof(_value);
         auto& current = producer.promise();
-        current.update_resume_point(sl);
         _promise->_waiting_task = &current;
         if (seastar::need_preempt()) {
             auto consumer = std::coroutine_handle<seastar::task>::from_address(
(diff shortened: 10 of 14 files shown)


include/seastar/core/task.hh
```diff
@@ -35,6 +36,8 @@ class task {
 protected:
     scheduling_group _sg;
 private:
+    std::source_location resume_point = {};
+
```
avikivity, line 40: This is making a very heavily used object larger.

avikivity:
Aren't those addresses derivable from the stack trace? Or are those points lost?

How large is std::source_location?

radoslawcybulski:
> Aren't those addresses derivable from the stack trace? Or are those points lost?
Probably with a little bit of magic. You'd still have to store a pointer somewhere, because stacktrace is valid for an actually running task, with my patch adds "next instruction" to tasks, which are suspended.
 
> How large is std::source_location?
8 bytes.

include/seastar/core/task.hh
radoslawcybulski: After adding `std::source_location`

`coroutine_traits_base<>::promise_type` (which inherits from task) is 64 bytes,
`continuation<Promise, Func, Wrapper, T>` is around 104-120, depending on exact version.

No difference except for that single unlucky value, who is "promoted" to the next 64 byte cache line in case of `continuation`.

radoslawcybulski:
my patch
```
instructions_per_op: mean=48465.87 standard-deviation=33.03 median=48459.42 median-absolute-deviation=22.63 maximum=48520.81 minimum=48437.81
instructions_per_op: mean=48482.90 standard-deviation=21.16 median=48487.16 median-absolute-deviation=6.97 maximum=48504.13 minimum=48447.34
instructions_per_op: mean=48469.42 standard-deviation=18.34 median=48472.45 median-absolute-deviation=17.26 maximum=48488.53 minimum=48448.81
```

master
```
instructions_per_op: mean=48337.41 standard-deviation=21.29 median=48334.08 median-absolute-deviation=19.67 maximum=48364.48 minimum=48317.02
instructions_per_op: mean=48352.36 standard-deviation=27.87 median=48347.35 median-absolute-deviation=16.28 maximum=48395.26 minimum=48322.29
instructions_per_op: mean=48339.83 standard-deviation=26.48 median=48323.12 median-absolute-deviation=20.80 maximum=48372.64 minimum=48319.03
```

My patch will run more instructions, as there are additional stores to the task object (std::source_location needs to be written).

Why do we measure instructions per operation? We are not billed by instruction count, rather by time, we should measure total running time instead? Does it work as a sanity check or am i missing something?

avikivity:
> 
> Why do we measure instructions per operation? We are not billed by instruction count, rather by time, we should measure total running time instead? Does it work as a sanity check or am i missing something?


Time is very unstable (just try it) since it depends on temperature and cooling.

Instructions is a more stable proxy for time (but inaccurate since instructions-per-clock can change).

avikivity:
So we add 0.25% overhead. Hard to judge if it's worthwhile.

avikivity:
> > Time is very unstable (just try it) since it depends on temperature and cooling.
> 
> Agree.
> 
> > Instructions is a more stable proxy for time (but inaccurate since instructions-per-clock can change).
> 
> It's absolutely stable, but we're counting steps, which - in my opinion - sort-of blinds us. It's like two different walks, the same step count, but step sizes differ, so is distance. Here for example if we're running at max IPC, then we've 0.25% overhead. If otherwise we're waiting for memory, then the write is pretty much free (we've already waited for this cache line, because other stuff is read / written there) and proc is waiting for other memory pieces, so it can do our write.

Writes are more-or-less free, except for the increase in instruction footprint (which are extra reads).

We aren't close to running at max IPC. Typical for mini-benchmarks like perf-simple-query is 2, and for production ~1. The main killer is waiting for instruction fetch (although maybe in production mispredicts would also contribute since the the data is less regular). I'm sure an M4 would give much higher IPC.

include/seastar/core/coroutine.hh
```diff
@@ -30,6 +30,7 @@
 
 #ifndef SEASTAR_MODULE
 #include <coroutine>
+#include <source_location>
```
tchaikov, line 33: it took me a while to check if it's safe to use `<source_location>` instead of `util/std-compat.hh` . the answer is yes. as we support the latest two major versions of the GCC and Clang compilers. in general, we use libstdc++, which added the support to `source_location` in https://github.com/gcc-mirror/gcc/commit/57d76ee9cf6265e012fad6286adfaeaba9414c11, which is included by gcc 12 and up. and the latest stable release of GCC is GCC 14.2.

tchaikov (commented):
if the performance impact could be a concern, can we make it an optional feature which is enabled only in the debug builds?

radoslawcybulski:
> if the performance impact could be a concern, can we make it an optional feature which is enabled only in the debug builds?

This has been discussed here (https://github.com/scylladb/seastar/issues/2381) and consensus is that it's the release is where it's useful the most.

tgrabiec (approved):

tgrabiec:
> > > Aren't those addresses derivable from the stack trace? Or are those points lost?
> > 
> > 
> > Probably with a little bit of magic. You'd still have to store a pointer somewhere, because stacktrace is valid for an actually running task, while my patch adds "next instruction" to tasks, which are suspended.
> > > How large is std::source_location?
> > 
> > 
> > 8 bytes.
> 
> Maybe we can tolerate it. @tgrabiec what's your opinion?

I think it's probably ok.

tchaikov (approved):
lgtm

include/seastar/core/task.hh
```diff
@@ -35,6 +36,8 @@ class task {
 protected:
     scheduling_group _sg;
 private:
+    std::source_location resume_point = {};
```
bitpathfinder, line 39: I'd put the `resume_point` first since it's larger than `scheduling_group`. that way the `task` object size is 4 bytes less.

radoslawcybulski: Done.

tgrabiec: Not really - if the class is not packed then the size includes the alignment, so order doesn't matter.

bitpathfinder: Yes, that's true. The object will be aligned to 8 bytes boundary. However in case we add something after these variables, less than 5 bytes, then there will be a difference, so having it ordered by size might be beneficial in the future.

include/seastar/core/task.hh
```diff
@@ -26,14 +26,18 @@
...
 #ifndef SEASTAR_MODULE
 #include <utility>
+#include <source_location>
 #endif
 
 namespace seastar {
 
 SEASTAR_MODULE_EXPORT
 class task {
+    std::source_location resume_point = {};
```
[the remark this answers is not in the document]
radoslawcybulski: Done.

bitpathfinder (approved):

radoslawcybulski:
@scylladb/seastar-maint please merge.

denesb:
> So we add 0.25% overhead. Hard to judge if it's worthwhile.

@avikivity I think the gain is much higher than 0.25% here. We need to reduce the bar of entry into debugging seastar applications.

radoslawcybulski:
@denesb can you tell me, how did you figure out performance differences for my patch (or any patch)? Sub 1% precision looks amazing to me, how did you do that?

denesb:
> @denesb can you tell me, how did you figure out performance differences for my patch (or any patch)? Sub 1% precision looks amazing to me, how did you do that?

I didn't, I just reacted to Avi's comment. I didn't claim that the perf regression is something else, just that 0.25% that Avi claimed is worth it for the gain of seeing line numbers.

michoecho:
> @denesb can you tell me, how did you figure out performance differences for my patch (or any patch)? Sub 1% precision looks amazing to me, how did you do that?

We're talking about the `perf_simple_query` results posted earlier in this thread.

radoslawcybulski:
@scylladb/seastar-maint  shameless ping - please merge

radoslawcybulski:
@scylladb/seastar-maint shameless ping - please merge.

include/seastar/core/coroutine.hh
nyh: Opened https://github.com/scylladb/seastar/issues/2961 on this.

nyh:
> @scylladb/seastar-maint shameless ping - please merge.

@avikivity and @tgrabiec are both maintainers and were involved in this review, I think they should get the honor of deciding to merge it.

Since this is the first time I'm seeing this issue, I have a beginner's question - where are all those fairly big "source location" strings saved in the executable (I'm not talking about the 8-byte pointers, which are also a problem - I'm talking about the actual strings). Can you estimate how much the stripped Scylla executable will grow due to these strings? I don't have any ideas if there are hundreds such strings, or hundreds of thousands.

radoslawcybulski:
> Since this is the first time I'm seeing this issue, I have a beginner's question - where are all those fairly big "source location" strings saved in the executable (I'm not talking about the 8-byte pointers, which are also a problem - I'm talking about the actual strings). Can you estimate how much the stripped Scylla executable will grow due to these strings? I don't have any ideas if there are hundreds such strings, or hundreds of thousands.

We will get one `std::source_location` data per suspention point. Assuming we have 1k suspention points in the code base and each in different file (we don't have that many files) and each path is 1k long (my paths are around 100 - 150 bytes) we would get max of 1mb of pure text. This text will be mapped to memory, but not paged in until someone uses, so it should be no issue for anyone but storage guys.
In reality i'd expect below 100kb (100 different files with 200 bytes per path give - 20 kb of text for example).

avikivity:
There's a gratuitous merge commit at the end.

@travisdowns please opine on whether this would be terrible for performance for you, or great for debugging.

michoecho:
> There's a gratuitous merge commit at the end.
> 
> @travisdowns please opine on whether this would be terrible for performance for you, or great for debugging.

@avikivity If we don't want to pessimize existing applications, we can always hide the stuff added in this patch behind the preprocessor. But please don't just let this patch rot. Let me have it at least in dev mode for starters. I've wasted so much time over the years on tests where the main debugging difficulty was finding out where a continuation came from.

nyh:
> We will get one `std::source_location` data per suspention point. Assuming we have 1k suspention points in the code base and each in different file (we don't have that many files) and each path is 1k long (my paths are around 100 - 150 bytes) we would get max of 1mb of pure text.

In addition to file names, there are function names, and those (with namespaces, argument type names, etc.) can be even longer and more numerous.

Could you do me a favor and compile Scylla (for example) with and without your patch and see how much bigger the stripped Scylla executable became? Maybe the difference is indeed tiny, I am not suggesting that it isn't - I just don't know.

radoslawcybulski:
@nyh sure any particular version? Otherwise i will build dev.

nyh:
> @nyh sure any particular version? Otherwise i will build dev.

No, whatever you want. The idea is to take some "big" Seastar project and see the impact on this change on it, it doesn't matter exactly which version of which build mode (but I wouldn't take "debug" build mode because that already inflates the executable in strange ways).

avikivity:
> > There's a gratuitous merge commit at the end.
> > @travisdowns please opine on whether this would be terrible for performance for you, or great for debugging.
> 
> @avikivity If we don't want to pessimize existing applications, we can always hide the stuff added in this patch behind the preprocessor.

That's the easy way out, but I don't want to take it.

> But please don't just let this patch rot. Let me have it at least in dev mode for starters. I've wasted so much time over the years on tests where the main debugging difficulty was finding out where a continuation came from.

That's reasonable. Let's see if we can push it to release too.

travisdowns:
@nyh wrote:

> where are all those fairly big "source location" strings saved in the executable (I'm not talking about the 8-byte pointers, which are also a problem - I'm talking about the actual strings).

They are stored in `.rodata`, and the `std::source_location` is effectively just a single pointer to this read-only data. So it is pretty efficient in the sense that if you don't access the string/line data itself, you will not even bring it into cache. In some way it's more efficient than old-school `__FILE__` and `__LINE__` passing around, which usually needs an `int` _and_ a `const char *` (though you could concat them in the preprocessor).

radoslawcybulski:
I stand corrected. `scylla` binary in dev mode has grown by 1.7 MB (from 125.41 to 127.13), approximately 1.72%.

EDIT: `std::source_location` contains also function names. Which, given their templated nature, can grown way beyond 200 bytes i assumed.

nyh:
Your patch is very impressive, and I don't have an opinion whether the 0.3% slowdown and the 1.7% executable size increase is a legitimate price to pay or not, I think other reviewers have had stronger opinions on this so I"ll let them decide.

I just want to note that in the past, the tradition wasn't to include the function names, file names, and line numbers in the executable itself - instead, backtraces used only an instruction pointer and nothing more, and an external tool (addr2line) was used to convert such instruction addresses to source code's filename/function-name/line-number - using "debug information" stored in an unstripped version of the executable. I think your patch isn't the first one that undoes this old tradition - we already know how to print backtraces with symbolic names and not just addresses - it just completes the process. Maybe it's a welcome change, because using addr2line was never very convenient.

radoslawcybulski:
Weren't the tradition dictated by a need to be more obscure due to a partially closed source nature of past scyllas?

nyh:
> Weren't the tradition dictated by a need to be more obscure due to a partially closed source nature of past scyllas?

No, I wasn't talking about Scylla's tradition, but in general - addr2line isn't specific to Scylla or closed source.
In any case, addr2line does not need the source code - just an executable with extra (or separate) "debugging information" (https://en.wikipedia.org/wiki/DWARF) which helps convert instruction-pointer (a.k.a. PC) addresses to file name, line number and function name. As I said, I agree that storing all this stuff in the executable and allowing the executable to introspect it is certainly more convenient than addr2line, but in the old days it wasn't done because it has a cost, as you noticed. But we can definitely decide that the convenience is worth the small cost.

avikivity:
The binary size growth isn't very interesting. This is a system that often manages dozens of terabytes per node. Nobody will notice a megabyte of strings, there are seven orders of magnitude separating the two.

However, it does generate an idea - can we use __builtin_return_address() here? It gives a tiny bit more information, and won't grow the executable.

avikivity:
In fact it can give a lot more information. We'll see all callers up to the first uninlined function, and when the direct caller is some glue function that appears often, it can help a lot.

radoslawcybulski:
Yes, i think so.

Note, that source code information is `compile-time` stuff (it will survive optimizations / inlining), while `__builtin_return_address()` is runtime (it will tell us, where return address is after optimizations, which is roughly what we have now).

travisdowns:
> We'll see all callers up to the first uninlined function

Yeah this is handy. In fact, I think probably this information is already encoded in the current task structure in most/all cases (e.g., the IP pointed to by the vtable of the task object), though not necessarily accessible in a standard way. Also very relevant: https://clang.llvm.org/docs/DebuggingCoroutines.html#async-stack-traces.

@avikivity wrote:

> @travisdowns please opine on whether this would be terrible for performance for you, or great for debugging.

I like this idea. We've wanted something similar to "name" tasks before and also considered using the source location as this name (with the option to override it). Using the __builtin_return address does seem to offer even more total information, but at the cost of requiring an external look.

One example of where a task name is useful is if one wanted to create a deterministic version of the reactor: the "trace" of tasks that are scheduled to run can be output in order to catch divergence.

If the performance impact is truly 0.25% I think it's worth it. If there a concern about it, it could be configurable? Using a macro to swap between source_location and an empty class (or a class with static stub methods) means you could limit the "macro damage" to a small area.

radoslawcybulski:
@travisdowns I will prepare updated version with updateable source_location alike class.

travisdowns:
> @travisdowns I will prepare updated version with updateable source_location alike class.

OK, but just to clarify I'm not a maintainer of seastar nor can I accept your PR, so you should probably guide yourself more based on what the maintainers say (I was just throwing in my 2c as Avi had asked).

avikivity:
@radoslawcybulski please check if __builtin_return_address(0) works, as it provides more information.

avikivity:
(and limits executable growth)

travisdowns:
wrt `__builtin_return_address(0)` it seems like the problematic case will be when the thing that's calling this (in this change) was inlined into the caller, in which case your stack won't include the creation of the task at all (only some high level frames).

If you captured that _plus_ the current IP you'd have enough info to reconstruct always at least the caller.

radoslawcybulski:
@avikivity godbolt says it works. Given it's pretty much undocumented builtin - probably.

As i mentioned before - it does something else. The point of `std::source_location` is to give everyone compile time constant that survives inlining optimization. Runtime return address is pretty much useless - you will get return pointer into template inside god knows which header.

I will also argue `__builtin_return_address(0)` provides LESS information, not more - otherwise we would all be using it in our -O3 mass inlined builds everywhere. Heck, even full stacktrace will probably contain less information, due to optimization mess.

I can fork our own source_location implementation using `__builtin_FILE` and `__builtin_LINE`, those should limit memory usage.

EDIT: those are clang / gcc specific.

avikivity:
> @avikivity godbolt says it works. Given it's pretty much undocumented builtin - probably.

It's not undocumented.

> 
> As i mentioned before - it does something else. The point of `std::source_location` is to give everyone compile time constant that survives inlining optimization. Runtime return address is pretty much useless - you will get return pointer into template inside god knows which header.

> 
> I will also argue `__builtin_return_address(0)` provides LESS information, not more - otherwise we would all be using it in our -O3 mass inlined builds everywhere.

I don't understand the logic. If X does not use Y, does it follow that Y is not useful? Perhaps X hasn't thought of using Y yet.

If it applied, it would be impossible to do anything new.


> Heck, even full stacktrace will probably contain less information, due to optimization mess.

Stack traces (when explored with addr2line -i) show much more information. Each return address on the stack encodes all the inlined call-sites that led to it up to the next uninlined caller. The more inlining happens, the more information it carries.

See https://godbolt.org/z/G678Eo7rh. Although there is just one location that calls g(), it knows exactly which path called it through a.

> 
> I can fork our own source_location implementation using `__builtin_FILE` and `__builtin_LINE`, those should limit memory usage.
> 


I suggest you try out __builtin_return_address(0) and compare the results.

radoslawcybulski:
I stand corrected, stack traces are actually sort of useful now, debug info has improved massively.

> Stack traces (when explored with addr2line -i) show much more information. Each return address on the stack encodes all the inlined call-sites that led to it up to the next uninlined caller. The more inlining happens, the more information it carries.

This assumes every instruction has unambiguous inlined sequence of calls leading to it. Or am i missing something?

> See https://godbolt.org/z/G678Eo7rh. Although there is just one location that calls g(), it knows exactly which path called it through a.

Good example. I took the liberty of removing `[[gnu::noinline]]` and `[[gnu::always_inline]]` and adding `h` implementation, that prints the `ptr` value. The code repeats 4 different values at `-O2` and `-O3` (which makes sense as clang just inlined to some depth). Adding flags used for `release` build makes the code print single value.
I also added a small `if` test, that calls `foo` from both branches - prints the same value in both cases.

Am i missing something? It doesn't take much to generate code, that drops ambiguous return pointers after optimization. Is there a way to resolve those i don't know (which would make me a X)? Otherwise that's pretty much the point i'm so poorly making. You can probably make it sort-of work with clever `noinline` flags (adding `noinline` to functions that also call `__builtin_return_address` would probably do it). Why bother? This will also hurt performance.

avikivity:
> I stand corrected, stack traces are actually sort of useful now, debug info has improved massively.
> 
> > Stack traces (when explored with addr2line -i) show much more information. Each return address on the stack encodes all the inlined call-sites that led to it up to the next uninlined caller. The more inlining happens, the more information it carries.
> 
> This assumes every instruction has unambiguous inlined sequence of calls leading to it. Or am i missing something?


It doesn't assume anything.

If inlining happens to duplicate a call to (let's say) future::then, then you will get more information.

If it doesn't, then you don't.

In the worst case __builtin_return_address will return the same amount of information as std::source_location.

In the common case, it will return more, since you will see the direct caller, and its callers up to the first uninlined function.

> 
> > See https://godbolt.org/z/G678Eo7rh. Although there is just one location that calls g(), it knows exactly which path called it through a.
> 
> Good example. I took the liberty of removing `[[gnu::noinline]]` and `[[gnu::always_inline]]` and adding `h` implementation, that prints the `ptr` value. The code repeats 4 different values at `-O2` and `-O3` (which makes sense as clang just inlined to some depth). Adding flags used for `release` build makes the code print single value. I also added a small `if` test, that calls `foo` from both branches - prints the same value in both cases.
> 
> Am i missing something? It doesn't take much to generate code, that drops ambiguous return pointers after optimization. Is there a way to resolve those i don't know (which would make me a X)? Otherwise that's pretty much the point i'm so poorly making. You can probably make it sort-of work with clever `noinline` flags (adding `noinline` to functions that also call `__builtin_return_address` would probably do it). Why bother? This will also hurt performance.

If you compare to std::source_location, then in all combinations of inlining options it returns one call site. __builtin_return_address will return from 1 to 32 call sites, depending on the options.

Here's the example updated to report both std::source_location and __builtin_return_address. If you change h() to print both, you will get 32 different values for __builtin_return_address but only one value for std::source_location.

https://godbolt.org/z/MqaocP1x5

radoslawcybulski:
> In the worst case __builtin_return_address will return the same amount of information as std::source_location.
> In the common case, it will return more, since you will see the direct caller, and its callers up to the first uninlined function.

Consider this stack trace (top is current)
```
__builtin_return_address(0) -> returns next instruction in `caz`
foo [inlined]
bar [inlined]
baz [not inlined]
gez [inlined]
caz [not inlined]
```
`__builtin_return_address` will return `return address` of `baz`, which means you will get `gez` -> `caz`, which is what you said - you get inlined stacks up to first non-inlined. You lose top level inlined stack tho (`foo`, `bar`). Your example works, because in your case you've artificially made `foo` not inlined (which means it has no inline calls between itself and a call to `__builtin_return_address(0)`, so nothing to loose). 
This is what i also said - we can sort-of make it work at the cost of performance (more instructions executed, as those stack frames don't generate themselves).

This might be improvable (although i doubt i'm the first to think about it), if - instead of calling `__builtin_return_address` (which gets return address of current function, which is not that useful or we need to stop inlining) we could get `next execution address pointer`. This sort-of reverves the issue - now we have to force all `then` implementations. Then next instruction pointer will be as if return pointer of `then` call if it were forced to noinline, we should get missing top level inlined frames and it should fly.
My instinct tells me i'm not the first one to think about it and they are major issues with this approach. While `std::source_location` will work 100% of times on all systems.

EDIT: i will give it a try, but it will take some time.

michoecho:
> In the worst case __builtin_return_address will return the same amount of information as std::source_location.

@avikivity No, why?

Imagine I want to debug a deadlocked fiber. It's stuck in this coroutine:
```
future<> my_function() {
    co_await a();
    co_await deadlock();
    co_await b();
}
```

`std::source_location` will point to the `co_await deadlock();` line. That's what I want. `__builtin_return_address(0)` will point to `reactor::do_run()`. (Or some other internal function that wakes the coroutine handle). That's useless.

The replacement for `std::source_location` is not the return address, but `rip`.

michoecho:
> This might be improvable (although i doubt i'm the first to think about it), if - instead of calling __builtin_return_address (which gets return address of current function, which is not that useful or we need to stop inlining) we could get next execution address pointer. This sort-of reverves the issue - now we have to force all then implementations. Then next instruction pointer will be as if return pointer of then call if it were forced to noinline, we should get missing top level inlined frames and it should fly.
My instinct tells me i'm not the first one to think about it and they are major issues with this approach. While std::source_location will work 100% of times on all systems.

@radoslawcybulski You mean, use `rip` instead of `std::source_location` like I did in https://github.com/scylladb/seastar/issues/2381#issuecomment-2263125627?

I wanted to use the instruction pointer instead of source location from the start, because it theoretically gives more information, but I didn't try to push that because in practice (when I was privately using that patch for debugging) the compiler was very often generating bad debug info for the addresses I obtained this way. (E.g. pointing to the right file, but to line 0 instead of the actual line). `std::source_location` prevents problems like that. So it's a tradeoff — less info, but more reliable due to explicit compiler support, or more info with a chance that the info will be less useful (without manual investigation to find the actual line) because the compiler didn't care enough to generate good debug info.

radoslawcybulski:
> @radoslawcybulski You mean, use `rip` instead of `std::source_location` like I did in [#2381 (comment)](https://github.com/scylladb/seastar/issues/2381#issuecomment-2263125627)?

Yep, the same. Also your patch highlights my other point - it's rare to figure out something new, most of the time if people are not using something there's a reason for it.

avikivity:
> > In the worst case __builtin_return_address will return the same amount of information as std::source_location.
> > In the common case, it will return more, since you will see the direct caller, and its callers up to the first uninlined function.
> 
> Consider this stack trace (top is current)
> 
> ```
> __builtin_return_address(0) -> returns next instruction in `caz`
> foo [inlined]
> bar [inlined]
> baz [not inlined]
> gez [inlined]
> caz [not inlined]
> ```
> 
> `__builtin_return_address` will return `return address` of `baz`, which means you will get `gez` -> `caz`, which is what you said - you get inlined stacks up to first non-inlined. You lose top level inlined stack tho (`foo`, `bar`). Your example works, because in your case you've artificially made `foo` not inlined (which means it has no inline calls between itself and a call to `__builtin_return_address(0)`, so nothing to loose). This is what i also said - we can sort-of make it work at the cost of performance (more instructions executed, as those stack frames don't generate themselves).
> 
> This might be improvable (although i doubt i'm the first to think about it), if - instead of calling `__builtin_return_address` (which gets return address of current function, which is not that useful or we need to stop inlining) we could get `next execution address pointer`. This sort-of reverves the issue - now we have to force all `then` implementations. Then next instruction pointer will be as if return pointer of `then` call if it were forced to noinline, we should get missing top level inlined frames and it should fly. My instinct tells me i'm not the first one to think about it and they are major issues with this approach. While `std::source_location` will work 100% of times on all systems.
> 

Right, my idea the point where you capture __builtin_return_address(0) is itself not inlined, but that's hardly a given.

What we want is the current instruction pointer (if inlined) or __builtin_return_address(0) if not, but that's not expressible.

std::source_location works, but is not always useful. It can point to, say, some function in loop.hh and give you not much idea about what's going on.

avikivity:
> > In the worst case __builtin_return_address will return the same amount of information as std::source_location.
> 
> @avikivity No, why?
> 
> Imagine I want to debug a deadlocked fiber. It's stuck in this coroutine:
> 
> ```
> future<> my_function() {
>     co_await a();
>     co_await deadlock();
>     co_await b();
> }
> ```
> 
> `std::source_location` will point to the `co_await deadlock();` line. That's what I want. `__builtin_return_address(0)` will point to `reactor::do_run()`. (Or some other internal function that wakes the coroutine handle). That's useless.


Well, it depends (which is bad) on whether the function capturing %rip was inlined.

Perhaps we could always_inline a wrapper, and let the compiler choose whether to inline the wrappee or or not.

> 
> The replacement for `std::source_location` is not the return address, but `rip`.

yes+no

radoslawcybulski:
I've run three tests (without patch, with `std::source_location` and with `slim_source_location` (RIP + `const char *` filename + `uint64_t` line). Results (size of scylla + output of simple-perf-query run) follows:

```
without patch:
-rwxr-xr-x. 1 y y 117237760 Oct 15 14:04 build/dev/scylla
109785.54 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.2 tasks/op,   56731 insns/op,   34064 cycles/op,        0 errors)
129664.30 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   56783 insns/op,   34771 cycles/op,        0 errors)
132627.36 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   56763 insns/op,   35072 cycles/op,        0 errors)
133472.86 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   56768 insns/op,   34881 cycles/op,        0 errors)
137312.72 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   56748 insns/op,   33911 cycles/op,        0 errors)
throughput:
        mean=   128572.56 standard-deviation=10851.14
        median= 132627.36 median-absolute-deviation=4900.30
        maximum=137312.72 minimum=109785.54
instructions_per_op:
        mean=   56758.37 standard-deviation=19.65
        median= 56762.56 median-absolute-deviation=10.69
        maximum=56782.65 minimum=56731.24
cpu_cycles_per_op:
        mean=   34539.85 standard-deviation=518.31
        median= 34771.03 median-absolute-deviation=475.91
        maximum=35071.78 minimum=33911.26
```

```
with std::source_location:
-rwxr-xr-x. 1 y y 119546200 Oct 15 14:40 build/dev/scylla
132906.63 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   56899 insns/op,   34873 cycles/op,        0 errors)
118051.07 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.2 tasks/op,   56861 insns/op,   39176 cycles/op,        0 errors)
128228.93 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   56843 insns/op,   36077 cycles/op,        0 errors)
135714.37 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   56858 insns/op,   33995 cycles/op,        0 errors)
134982.33 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   56878 insns/op,   34199 cycles/op,        0 errors)
throughput:
        mean=   129976.67 standard-deviation=7277.31
        median= 132906.63 median-absolute-deviation=5005.67
        maximum=135714.37 minimum=118051.07
instructions_per_op:
        mean=   56867.64 standard-deviation=21.30
        median= 56860.67 median-absolute-deviation=10.37
        maximum=56898.66 minimum=56843.22
cpu_cycles_per_op:
        mean=   35663.98 standard-deviation=2125.23
        median= 34872.53 median-absolute-deviation=1464.79
        maximum=39176.33 minimum=33994.54
diff size is 2308440 (~2.2 MB)
```

```
with `slim_source_location`:
-rwxr-xr-x. 1 y y 119431736 Oct 15 13:25 build/dev/scylla
124916.79 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.2 tasks/op,   57187 insns/op,   37190 cycles/op,        0 errors)
118595.81 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.2 tasks/op,   57126 insns/op,   39020 cycles/op,        0 errors)
129827.40 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   57165 insns/op,   35711 cycles/op,        0 errors)
133294.11 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   57172 insns/op,   34779 cycles/op,        0 errors)
133896.32 tps ( 71.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   57152 insns/op,   34626 cycles/op,        0 errors)
throughput:
        mean=   128106.09 standard-deviation=6403.56
        median= 129827.40 median-absolute-deviation=5188.02
        maximum=133896.32 minimum=118595.81
instructions_per_op:
        mean=   57160.30 standard-deviation=23.14
        median= 57164.96 median-absolute-deviation=11.48
        maximum=57187.23 minimum=57125.81
cpu_cycles_per_op:
        mean=   36265.20 standard-deviation=1847.07
        median= 35710.56 median-absolute-deviation=1486.16
        maximum=39019.93 minimum=34626.24
diff size is 2193976 (~2 MB)
```

radoslawcybulski:
We get additional instructions (well, 3 writes instead of one), roughly 200 instructions per op, (0.3%).
`scylla` shrunk by approximately 200kb.

Note sure if it's worth. Most of our tools probably deals with `std::source_location`, but that's adaptable.
@avikivity please make a call or tell me, whom should i pull for more info here.

avikivity:
> We get additional instructions (well, 3 writes instead of one), roughly 200 instructions per op, (0.3%). `scylla` shrunk by approximately 200kb.
> 
> Note sure if it's worth. Most of our tools probably deals with `std::source_location`, but that's adaptable. @avikivity please make a call or tell me, whom should i pull for more info here.

Our tools deal well with addresses too.

Looks like the slim source_location isn't so slim. I can understand it - source_location is fat out-of-line but slim inline.

I need to think more about it. My feeling is that source_location won't work well as more infrastructure is converted to coroutines, we'll just see some internal coroutine there.

avikivity:
> compiler was very often generating bad debug info for the addresses I obtained this way. (E.g. pointing to the right file, but to line 0 instead of the actual line). 

Was this with addr2line or llvm-addr2line? We recently saw that llvm-addr2line is better.

include/seastar/core/slim_source_location.hh
```diff
@@ -0,0 +1,73 @@
...
+        constexpr std::int32_t line() const noexcept { return _line; }
+        constexpr std::int32_t column() const noexcept { return _column; }
+        constexpr const char* file_name() const noexcept { return _file; }
+#ifdef SSL_HAS_RIP
+        constexpr std::uintptr_t next_address() const noexcept { return _next_address; }
+#endif
+
+        [[gnu::always_inline]] slim_source_location(const char* file = __builtin_FILE(), std::int32_t line = __builtin_LINE(), std::int32_t column = __builtin_COLUMN())
+                : _file(file), _line(line), _column(column) {
+#ifdef SSL_HAS_RIP_X86
```
avikivity: SSL?

radoslawcybulski: slim source line.
My brain keeps telling me i've seen this abbreviation, but i've not found anything better. I can just expand into `SLIM_SOURCE_LINE_***`, those macros are contained in this file anyway.

include/seastar/core/slim_source_location.hh
```diff
@@ -0,0 +1,73 @@
...
+            std::uintptr_t rip;
+            asm("leaq 0(%%rip), %0":"=r"(rip));
```
avikivity: We could slim it down with something like

```c++
struct slim_data* sd;
asm ("1: lea 2f, %0 \n"
          ".pushsection .rodata.something \n"
          "2: \n"
          ".quad 1b \n"
          ".quad %1 \n"
          ".long %2, %3 \n"
          ".popsection"
          : "=r"(sd) : "i"(__builtin_FILE()) : "i"(__builtin_LINE()), "i"(__builtin_COLUMN()));
```

This pushes the data to the data section and we end up with one instruction.

include/seastar/core/slim_source_location.hh
avikivity: Prefix macros with SEASTAR_, and don't abbreviate, it just confuses people.

include/seastar/core/slim_source_location.hh
radoslawcybulski: Well, i do understand few identifiers from the example, but that's all. :)

What is `1:` and `2:` (i assume labels? They don't seem used? Are those global identifiers or local to function / module?). Why we do `.quad 1b` (`quad` is 64bit probably, so `1b` is just `1`)? `long` is 32bit i assume?  What is `=r...` syntax?

avikivity: Yes, 1: and 2: are labels, 2f = 2 forward, 1b = 1 back. Using numbers makes them local.

"=r" -> output register

avikivity: It's preparing a static struct pointing back at the lea instruction.

radoslawcybulski: What is `2 forward`?
Will fix it later on, i'm on sick leave until tomorrow.

avikivity: It's black magic and if you ever manage to read it fluently, you're cursed for life.

michoecho: @avikivity I don't think this is going to work. In your example, the `__builtin_FILE()`  and `__builtin_LINE()` will just point to `include/seastar/core/slim_source_location.hh`. And it doesn't seem fixable.

If you want to create a constinit static struct, then the FILE and LINE must be constexpr.
So the FILE and LINE can't be passed from above, because we need the function to be eventually called from a non-constexpr context.
So the only FILEs and LINEs you can put inside the struct are ones obtained pointing to the function that contains the struct definition. They can't come from the caller. So they are useless for our purpose.

Without making `then` (and friends) and `co_await` into macros, I don't see a way to a way to achieve what you are trying to achieve.
It's either RIP or `source_location`, I don't see a non-macro way to combine them into a static struct.

avikivity: Right. I don't think it will work.

I guess we can fall back to the standard source_location.

michoecho:
> > compiler was very often generating bad debug info for the addresses I obtained this way. (E.g. pointing to the right file, but to line 0 instead of the actual line).
> 
> Was this with addr2line or llvm-addr2line? We recently saw that llvm-addr2line is better.

@avikivity I'm 80% sure that the problem was at compile time, not decode time. But I don't remember my problem well, I only have the vague memory that I was getting the RIPs I wanted but the file and line info for those RIPs in the DWARF was wrong(/useless) sometimes.

include/seastar/core/slim_source_location.hh
michoecho: By the way, in cases where macros are on the table (in this case they aren't), you can get this static struct without stooping to inline asm.

Eg. I have this tracepoint macro (which I'm actually using in practice right now, when writing a debug util for https://github.com/scylladb/scylladb/issues/25679). (`entry`) is the static struct in this example. And I use its address as the metadata header for each event in the trace.
```c++
struct tracepoint_entry {
    const char* name;
    const char* file;
    int line;
    const char* function;
    const char* signature;
    void *rip;
};

#define TRACEPOINT(eventlevel, tp_name, ...) { \
__label__ rip; \
    using namespace seastar; \
    static constexpr auto sig __attribute__((section("tracepoint_signatures"), used)) = COMPUTE_SIGNATURE(SIG(__VA_ARGS__)); \
    static constexpr char namearr[] __attribute__((section("tracepoint_names"), used)) = tp_name; \
    static constexpr char filearr[] __attribute__((section("tracepoint_files"), used)) = __FILE__; \
    static constexpr tracepoint_entry entry __attribute__((section("tracepoints"), used)) = { \
        .name = namearr, \
        .file = filearr, \
        .line = __LINE__, \
        .function = __PRETTY_FUNCTION__, \
        .signature = sig.data(), \
        .rip = &&rip, \
    }; \
rip: \
    size_t sz = compute_size(EXTRACT_ARGS(__VA_ARGS__)); \
    auto out = local_tracer->write(eventlevel, sz + 16); \
    seastar::write_le<uintptr_t>(reinterpret_cast<char*>(out), reinterpret_cast<uintptr_t>(&entry)); \
    out += sizeof(uintptr_t); \
    seastar::write_le<uint64_t>(reinterpret_cast<char*>(out), __rdtsc()); \
    out += sizeof(uint64_t); \
    serialize_tracepoint(out __VA_OPT__(,) EXTRACT_ARGS(__VA_ARGS__)); \
}
```

radoslawcybulski: @michoecho why do we need this black magic incantation with `__attribute__((section("..."), ised))`? Woudln't simple `static` suffice?

radoslawcybulski: @michoecho why this woulnd't work?

```
struct slim_data* sd;

void *foo(const char *file, int line, int column) {
asm ("1: lea 2f, %0 \n"
          ".pushsection .rodata.something \n"
          "2: \n"
          ".quad 1b \n"
          ".quad %1 \n"
          ".long %2, %3 \n"
          ".popsection"
          : "=r"(sd) : "i"(file) : "i"(line), "i"(column));
...
}
```
?

michoecho: > Why this woulnd't work?

Because Avi's wish would be to combine the instruction pointer and the source location (and maybe `__PRETTY_FUNCTION__` too, why not) into a static, constant-initialized struct, so that all statically-known info can be represented inside a task by a single pointer at runtime. And that's what the piece of asm is trying to accomplish.

But if you want that struct to be constant-initialized, you must initialize it with constants. And in your example, you are trying to initialize it with not-constants. (If the function is always inlined, then the values of `line` and `column` and `file` _are_ known at compile time, but something like this doesn't fit into the type system. A function must stand on its own, there's no "post-inlining constexpr"). It will be rejected by the compiler. (The `"i"` in `"i"(line)` stands for "immediate". This in x86 terms means a constant woven directly into the encoding of the instruction. And you are trying to pass a variable there).

> why do we need this black magic incantation with ` __attribute__((section("..."), used))`?

I want all `tracepoint_entry` structs to be collected by the linker into their own section (`tracepoints` here) so that my trace decoder can later, while preparing for decoding, look at the Scylla binary and iterate over all `tracepoint_entry` structs scattered inside it, to get the signatures for all event types in the trace.

If they aren't packed into a section, they will be in random places in `.rodata`, and they can't be found without knowing their names. If they are packed into a section, my decoder can open the Scylla executable, find its `tracepoints` section, cast it to a `tracepoint_entry[]` array, iterate over it, and generate a corresponding decoder for each `TRACEPOINT` in the program. 

For example, I can put 
```c++
TRACEPOINT(event_level::debug, "io_begin", "class", io_request.class, "id", io_request.id, "size", io_request.size);
```
anywhere in the program, and it will produce a
```
tracepoint_entry{.name = "io_begin", .signature = "class:u8,id:u64,size:u64", ...}
```
struct instance to the `tracepoints` section at some address X, and the decoder will iterate over X (among others), see the signature and use it generate a decoder function which reads a
```
struct io_begin {
    uint8_t class;
    uint64_t id;
    uint64_t size;
};
```
from the trace whenever the header is X.
And then the functions generated from signatures can be compiled into the actual decoder program.

Other sections are not needed for anything, they are just there to keep the metadata strings neatly ordered in the ELF.

The `used` is probably not needed either, I just wanted to make sure the linker (and/or LTO) won't play garbage collection tricks on me.

---

Anyway, all that's not very applicable in this thread, because macros are off the table.

radoslawcybulski: @michoecho amazing, thank you for this explanation!

radoslawcybulski:
Reverted "slim" source_location commit, we're back to `std::source_location` only. Are we good to go, @avikivity ?

radoslawcybulski:
Patch rebased, "upgraded" two `await_suspend(std::coroutine_handle<> consumer)` functions to `await_suspend(std::coroutine_handle<Promise> consumer)` with template parameter `Promise`.

include/seastar/util/backtrace.hh
```diff
@@ -159,14 +159,17 @@ class tasktrace {
 public:
     using entry = std::variant<shared_backtrace, task_entry>;
     using vector_type = boost::container::static_vector<entry, 16>;
+    using vector_resume_points_type = boost::container::static_vector<std::source_location, 16>;
```
gusev-p: why not include resume point into entry (make it struct)

include/seastar/util/backtrace.hh
```diff
@@ -159,14 +159,17 @@ class tasktrace {
...
 private:
     simple_backtrace _main;
     vector_type _prev;
+    vector_resume_points_type _prev_resume_points;
     scheduling_group _sg;
     size_t _hash;
 public:
     tasktrace() = default;
     tasktrace(simple_backtrace main, vector_type prev, size_t prev_hash, scheduling_group sg) noexcept;
+    tasktrace(simple_backtrace main, vector_type prev, vector_resume_points_type prev_resume_points, size_t prev_hash, scheduling_group sg) noexcept;
```
gusev-p: we can leave only one constructor

include/seastar/util/backtrace.hh
```diff
@@ -159,14 +159,17 @@ class tasktrace {
 public:
     using entry = std::variant<shared_backtrace, task_entry>;
     using vector_type = boost::container::static_vector<entry, 16>;
+    using vector_resume_points_type = boost::container::static_vector<std::source_location, 16>;
 private:
     simple_backtrace _main;
     vector_type _prev;
+    vector_resume_points_type _prev_resume_points;
```
gusev-p: empty commit should be removed

gusev-p (commented):
@radoslawcybulski @scylladb/scylla-maint I see that all maintainers are ok with this approach, lets push this forward, this promises to be very helpful for debugging.

I didn't know about this work, so I [invented my own](https://github.com/scylladb/seastar/pull/3255). I wanted to improve reactor reporting in `do_dump_task_queue` because `typeid(task)->name()` is useless in most cases, especially for coroutines. I wanted to avoid inflating the task by another 8 bytes, so I tried to extract a meaningful symbol from existing data structures instead. We can use resume address stored in the coroutine frame, although it won't tell us the exact co_await within the coroutine. This would probably have been enough to diagnose the reactor queue problems, but if we're ok to store the source_location in the task instance it would be much better.

> std::source_location works, but is not always useful. It can point to, say, some function in `loop.hh` and give you not much idea about what's going on.

We can gradually improve this in the future by capturing `source_location` in the public functions of `loop.hh` and forwarding them to the appropriate task instance.

@radoslawcybulski please update the `do_dump_task_queue` to `use task->get_resume_point()`.

include/seastar/util/backtrace.hh
radoslawcybulski: Updated.

include/seastar/util/backtrace.hh
radoslawcybulski: Done.

include/seastar/util/backtrace.hh
radoslawcybulski: Done.

radoslawcybulski:
Updated patch - modified `tasktrace::entry` to contain both source_location and previous value. Improve printing in `do_dump_task_queue` - it now uses task's resume point.

radoslawcybulski:
```
/__w/seastar/seastar/apps/rpc_tester/rpc_tester.cc:574:42: internal compiler error: in make_decl_rtl, at varasm.cc:1459
```

yay...

EDIT: it's this one https://gcc.gnu.org/bugzilla/show_bug.cgi?format=multiple&id=121643 . We need a different version of gcc or clang.

include/seastar/core/coroutine.hh
```diff
@@ -218,7 +220,8 @@ public:
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl) noexcept {
+    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
+        hndl.promise().update_resume_point(sl);
```
gusev-p: shouldn't we call `update_resume_point` only in the suspend case of the if below? Same of other call sites

gusev-p:
> ```
> /__w/seastar/seastar/apps/rpc_tester/rpc_tester.cc:574:42: internal compiler error: in make_decl_rtl, at varasm.cc:1459
> ```
> 
> yay...
> 
> EDIT: it's this one https://gcc.gnu.org/bugzilla/show_bug.cgi?format=multiple&id=121643 . We need a different version of gcc or clang.

Maybe the compiler would like the `await_transform` more?

https://stackoverflow.com/questions/76110225/why-is-promise-typeawait-transform-greedy
> If you want to have certain specific co_awaits be instrumented by the promise, you can do that by doing co_await my_instrumented_awaitable(<expr>) manually as needed. But if you need all of them to be instrumented (which can be used for tracing suspend/resume points), if it's not something you pick-and-choose, then you use await_transform.

src/util/backtrace.cc
gusev-p, line 207: nit: can be made written more compactly:
```
for (const auto & [ e, resume_loc ] = b._prev) {
}
```

gusev-p:
Cool:
```
WARN  2026-02-13 10:20:52,265 [shard 0:main] seastar - Too long queue accumulated for main (14 tasks)
 1: seastar/src/core/future-util.cc:133
 1: replica/table.cc:1277
 6: compaction/task_manager_module.cc:802
 6: tasks/task_manager.cc:322
```
One problem is that now all seastar api methods should have a `source_location` parameter. For example, here `future-util.cc:133` refers to `sleep_abortable` -- we don't know who called it.

include/seastar/core/coroutine.hh
radoslawcybulski: We don't care - either we're going to be suspended later on and resume point gets updated or we won't, in which case resume point never comes into play at all.

In all cases even out-of-date suspension point gives some info.

radoslawcybulski:
> One problem is that now all seastar api methods should have a `source_location` parameter. For example, here `future-util.cc:133` refers to `sleep_abortable` -- we don't know who called it.

That's a tricky question. For example:

```
template <typename Clock>
future<> sleep_abortable(typename Clock::duration dur, abort_source& as) {
    struct sleeper {
        promise<> done;
        timer<Clock> tmr;
        abort_source::subscription sc;

        sleeper(typename Clock::duration dur, abort_source& as)
                : tmr([this] { done.set_value(); }) {
            auto sc_opt = as.subscribe([this] (const std::optional<std::exception_ptr>& opt_ex) noexcept {
                if (tmr.cancel()) {
                    done.set_exception(opt_ex.value_or(std::make_exception_ptr(sleep_aborted())));
                }
            });
            if (sc_opt) {
                sc = std::move(*sc_opt);
                tmr.arm(dur);
            } else {
                done.set_exception(sleep_aborted());
            }
        }
    };
    //FIXME: Use do_with() after #373
    auto s = std::make_unique<sleeper>(dur, as);
    auto fut = s->done.get_future();
    return fut.finally([s = std::move(s)] { });
}
```

what is more important to know - who called `sleep_abortable` or where within `sleep_abortable` the future will be resumed? We probably don't want to keep two source locations for both cases.

EDIT: personally i think who called us is more important, but that depends on usecase. If you give me functions to annotate i will add them.

src/util/backtrace.cc
radoslawcybulski: Updated.

include/seastar/core/coroutine.hh
gusev-p: I mean if we have a lot of `co_await`-s for already resolved futures we will have many redundant `MOV` instructions -- these resume points won't be needed.

gusev-p:
> what is more important to know - who called `sleep_abortable` or where within `sleep_abortable` the future will be resumed? We probably don't want to keep two source locations for both cases.

Probably we can assume that the target audience of the `source_location` feature are programmers who use Seastar, not Seastar developers -> we should capture source_location at the Seastar api boundaries.

gusev-p:
> If you give me functions to annotate i will add them.

If we used `__builtin_return_address(level)` we could probe several previous frames and store RIP of the first non-Seastar one. We can precompute the Seastar code memory ranges. This way we would't need to pass `source_location` everywhere, probably it would be enough to capture it in `await_suspend` and in `void schedule(Pr&& pr, Func&& func, Wrapper&& wrapper)`.

radoslawcybulski:
> If we used `__builtin_return_address(level)` we could probe several previous frames and store RIP of the first non-Seastar one.

instead we would have to force inline a lot of things. This had been tried somewhere up in the discussion.

gusev-p:
> what is more important to know - who called `sleep_abortable` or where within `sleep_abortable` the future will be resumed? We probably don't want to keep two source locations for both cases.

I don't think we need to forward `source_location` to helpers like `sleep_abortable` -- user code which waits for these sleeps should be reported via its own `source_location`-s. For example, the coroutine `co_await sleep()` will be reported with the proper source location, we just need to walk the `waiting_task()` task chain in `do_dump_task_queue` similar as we do in `current_tasktrace()`.

The `source_location`-s should probably be forwarded in `smp.hh` and `sharded.hh` functions.

gusev-p:
> The `source_location`-s should probably be forwarded in `smp.hh` and `sharded.hh` functions.

Actually we don't need that. We need some way to correlate tasks not directly linked by the `waiting_task()` chain. For example, `smp.hh/async_work_item::run_and_dispose` calls the func on the target shard, but the resulting future is not linked to the future on the original shard -- the completion signal is conveyed by side effects (`smp_message_queue`).

gusev-p:
@radoslawcybulski I've [tuned reactor/do_dump_task_queue](https://github.com/gusev-p/seastar/commit/9721bfcaeda6ab5268f6385492ddb5d8bf29ed14) a bit, please merge into your PR.

I like the result:
```
WARN  2026-02-13 23:08:51,953 [shard 0:main] seastar - Too long queue accumulated for main (21 tasks)
 1: tasks/task_manager.cc:253:9
        tasks/task_manager.cc:220:8
 1: tasks/task_manager.cc:220:8
 6: seastar/src/core/future-util.cc:133:16
        seastar/include/seastar/core/future.hh:1653:54
        tasks/task_manager.cc:327:12
 7: N7seastar13shared_futureIJEE12shared_stateE
 6: replica/distributed_loader.cc:419:5
        replica/distributed_loader.cc:331:9
        replica/distributed_loader.cc:453:13
        N7seastar9coroutine17parallel_for_eachIZN7replica18distributed_loader17populate_keyspaceERNS_7shardedINS2_8databaseEEERNS4_IN2db15system_keyspaceEEERNS2_8keyspaceENS_13basic_sstringIcjLj15ELb1EEEE3$_0EE
        replica/distributed_loader.cc:441:14
        N7seastar12_GLOBAL__N_116thread_wake_taskE
        seastar/include/seastar/core/thread.hh:261:28
        seastar/include/seastar/core/thread.hh:263:12
        N7seastar12_GLOBAL__N_116thread_wake_taskE
        seastar/include/seastar/core/thread.hh:261:28
        seastar/include/seastar/core/thread.hh:263:12
        seastar/src/core/app-template.cc:165:37
        seastar/src/core/app-template.cc:167:12
        seastar/src/core/app-template.cc:169:12
```

Let's review/merge this!

gusev-p (commented):
LGTM

gusev-p (approved):

gusev-p:
@radoslawcybulski Let's just disable `source_location`-s for gcc -- I did that in [this PR](https://github.com/scylladb/seastar/pull/3258) and the CI has passed.

radoslawcybulski:
Updated patch: included @gusev-p improvements.

radoslawcybulski:
@avikivity @scylladb/seastar-maint please merge.

gusev-p:
@avikivity ping, please review

include/seastar/core/coroutine.hh
```diff
@@ -194,8 +193,8 @@ public:
     }
 
     template<typename U>
-    void await_suspend(std::coroutine_handle<U> hndl, std::source_location sl = std::source_location::current()) noexcept {
-        hndl.promise().update_resume_point(sl);
+    void await_suspend(std::coroutine_handle<U> hndl SEASTAR_COROUTINE_LOC_PARAM) noexcept {
+        SEASTAR_COROUTINE_LOC_STORE(hndl.promise());
```
avikivity, line 197: Please link the gcc bug in the changelog entry.

Does this have to be a macro?

gusev-p: > Please link the gcc bug in the changelog entry.

It's linked in the definitions of these macros. Do you want it to be mentioned in the commit message?

> Does this have to be a macro?

We need to conditionally add an optional parameter and pass it to `update_resume_point`. I'm not sure how to do that without the macro. One option is to declare two versions of `await_suspend` -- one with the source_location and one without it. But then we'll need to do that for all `await_suspend` declarations, there are quite a few of them.

avikivity: > > Please link the gcc bug in the changelog entry.
> 
> It's linked in the definitions of these macros. Do you want it to be mentioned in the commit message?


I guess it's enough.

> 
> > Does this have to be a macro?
> 
> We need to conditionally add an optional parameter and pass it to `update_resume_point`. I'm not sure how to do that without the macro. One option is to declare two versions of `await_suspend` -- one with the source_location and one without it. But then we'll need to do that for all `await_suspend` declarations, there are quite a few of them.

For a bug workaround we can relax the rules against macros.

include/seastar/util/std-compat.hh
```diff
@@ -56,3 +56,18 @@ using source_location
     = std::source_location;
 
 }
+
+#if defined(__GNUC__) && !defined(__clang__)
+// GCC Workaround: Strip source_location to prevent ICE during RTL expansion.
+// See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114675
```
nvartolomei, line 62: Wrong link?

> 114675 - warning for "reference to not fully constructed object" - P3 enhancement

PR #2709, opened 2025-04-01
tls: Make session resume key shared across credentials builders creds

### LINKED ISSUE #2708: TLS session resume should share session key in certificate builder
ATM, the session key used for TLS1.3 resume mode is generated per certificate bundle. 
This does not match the common use cases in certain big consumers of seastar API:s (scylla),
where a builder typically creates a certificate set per shard.
As is, sessions will only be reusable per certificate set object.

The obvious fix is to create the key already in builder, and propagate to all generated certificate sets.
This _does_ open up some concern if someone were to use the same cert builder to generate disparate
certificate sets. But it is probably a pattern that can be handled by a simple documented "don't!".

elcallio:
Fixes #2708

Session resume key (if enabled) was only generated/kept on credentials level. This means that access patterns where we create a new such object per shard, using same builder, would have different session keys, and thus not be able to resume each others sessions.

This moves key generation to builder level and re-uses this for all generated server credentials. Note that a reload of any file (in reloadable) will re-create the key (though it should become invalid anyway, but...).

Add tests for this, and a hefty warning that session keys are shared, should anyone try to re-use a builder for disparate credentials.

### DIFF
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -339,6 +339,13 @@ future<tls::x509_cert> tls::x509_cert::from_file(
 
 // wrapper for gnutls_datum, with raii free
 struct gnutls_datum : public gnutls_datum_t {
+    gnutls_datum(size_t s) {
+        data = reinterpret_cast<unsigned char*>(gnutls_malloc(s));
+        if (data == nullptr) {
+           throw std::bad_alloc();
+        }
+        size = s;
+    }
     gnutls_datum() {
         data = nullptr;
         size = 0;
@@ -436,12 +443,17 @@ public:
     client_auth get_client_auth() const {
         return _client_auth;
     }
-    void set_session_resume_mode(session_resume_mode m) {
+    void set_session_resume_mode(session_resume_mode m, std::span<const uint8_t> key = {}) {
         _session_resume_mode = m;
         // (re-)generate session key
         if (m != session_resume_mode::NONE) {
             _session_resume_key = {};
-            gnutls_session_ticket_key_generate(&_session_resume_key);
+            if (key.empty()) {
+                gtls_chk(gnutls_session_ticket_key_generate(&_session_resume_key));
+            } else {
+                _session_resume_key = gnutls_datum(key.size());
+                std::copy(key.begin(), key.end(), _session_resume_key.data);
+            }
         }
     }
     session_resume_mode get_session_resume_mode() const {
@@ -710,6 +722,11 @@ void tls::credentials_builder::set_priority_string(const sstring& prio) {
 
 void tls::credentials_builder::set_session_resume_mode(session_resume_mode m) {
     _session_resume_mode = m;
+    if (m != session_resume_mode::NONE) {
+        gnutls_datum key;
+        gtls_chk(gnutls_session_ticket_key_generate(&key));
+        _session_resume_key.assign(key.data, key.data + key.size);
+    }
 }
 
 template<typename Blobs, typename Visitor>
@@ -760,7 +777,7 @@ void tls::credentials_builder::apply_to(certificate_credentials& creds) const {
 
     creds._impl->set_client_auth(_client_auth);
     // Note: this causes server session key rotation on cert reload
-    creds._impl->set_session_resume_mode(_session_resume_mode);
+    creds._impl->set_session_resume_mode(_session_resume_mode, std::span{_session_resume_key.begin(), _session_resume_key.end()});
 }
 
 shared_ptr<tls::certificate_credentials> tls::credentials_builder::build_certificate_credentials() const {
@@ -925,6 +942,10 @@ public:
                 return;
             }
             try {
+                // force rebuilding session resume mode key if
+                // enabled. should not reuse sessions across certificate
+                // change (should not work anyway)
+                set_session_resume_mode(_session_resume_mode);
                 if (_creds) {
                     _creds->rebuild(*this);
                 }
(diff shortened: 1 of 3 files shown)


nyh (approved):
Thanks, looks good to me although I have not been able to test it yet myself. I understand that @dkropachev was able to test that it fixes the right problem (sessions were not being reused on a multi-shard applications, but were reused on a single shard), but @dkropachev - were you able to test that this specific patch fixes the problem on multi-shard runs?

@elcallio I couldn't understand where exactly in the code do we have a cross-shard cache of these session keys, if this is done in a safe multi-threaded way, and when do we ever drop old session keys from this cache, but I'm sure you thought about it already and it works well :-)

nyh:
I think I can merge this patch already, but since you specifically asked @xemul and @tchaikov to review, I'll give them the courtesy of being able to review, if they want.

elcallio:
Not a cross shard cache, but a pre-computed key in the credentials builder. This is applied to all credentials created with this builder (or a copy of it). 
The usage pattern is 
* create builder (shard 0)
* call each shard with builder, create credentials used for TLS sessions. They now all use the same key

Session keys are re-generated every time you set the session resume mode, or every time a reloadable credentials do a reload.

src/net/tls.cc
```diff
@@ -710,6 +718,11 @@ void tls::credentials_builder::set_priority_string(const sstring& prio) {
 
 void tls::credentials_builder::set_session_resume_mode(session_resume_mode m) {
     _session_resume_mode = m;
+    if (m != session_resume_mode::NONE) {
+        gnutls_datum key;
+        gnutls_session_ticket_key_generate(&key);
```
tchaikov, line 723: might want to check the return value of `gnutls_session_ticket_key_generate()`? as, in theory, it could fail by returning a non-zero value. see https://man7.org/linux/man-pages/man3/gnutls_session_ticket_key_generate.3.html and https://github.com/gnutls/gnutls/blob/94e01a704f307a83ccb7799d1c2626f9635f2b5c/lib/ext/session_ticket.c#L516 .

tchaikov (approved):
lgtm apart from a nit.

src/net/tls.cc
elcallio: True. Can OOM. Added check to both calls.

tchaikov (approved):
lgtm

src/net/tls.cc
```diff
@@ -710,6 +718,11 @@ void tls::credentials_builder::set_priority_string(const sstring& prio) {
 
 void tls::credentials_builder::set_session_resume_mode(session_resume_mode m) {
     _session_resume_mode = m;
+    if (m != session_resume_mode::NONE) {
+        gnutls_datum key;
+        gtls_chk(gnutls_session_ticket_key_generate(&key));
+        _session_resume_key.assign(key.data, key.data + key.size);
+    }
 }
```
dkropachev, line 726: Just an idea, If we want to tls renegotiation to work even when node is being restarted we need `_session_resume_key` to be persisted either in the config or as a file or any other way.

Which means we will need another API that could feed pregenerated `_session_resume_key` to the credentials builder.

elcallio: I think that is unnecessary. If you have node restarts so frequent it impacts TLS session management, you have a bigger problem. 
Should the need arise, get/set API:s for this can be added, but for now I'd say it is overreach.

dkropachev: > I think that is unnecessary. If you have node restarts so frequent it impacts TLS session management, you have a bigger problem. Should the need arise, get/set API:s for this can be added, but for now I'd say it is overreach.

It is not about frequency, it is about of amount of clients that are trying to reconnect at the same time.
When client is reconnecting it establishes TLS connection first and the go thru CQL initialization process, CPU pressure of having thousands of clients reconnecting at the same time could be so high that it impedes CQL initialization process, as result client breaks connection and reconnects again.
Though, on second reconnect, it should have TLS ticket learned, so it should go smoother.

elcallio: The point is that a _restart_ should be infrequent. Connections can use session resume across the nodes uptime or the ticket expiry, whichever comes first (lets hope the latter). But also remember that storing the session key across restarts would more or less require them to be clean restarts (because we do want to refresh the key at least on cert reload etc), which, in the case of crashes, makes it even less useful. 
Also remember that using the same key for to long is a security concern, and already the cross-session usage here is somewhat dubious (but since we can treat cross-shard instances as equivalent to a "normal" server single session ok).

src/net/tls.cc
```diff
@@ -436,12 +436,20 @@ class tls::certificate_credentials::impl: public gnutlsobj {
     client_auth get_client_auth() const {
         return _client_auth;
     }
-    void set_session_resume_mode(session_resume_mode m) {
+    void set_session_resume_mode(session_resume_mode m, const std::vector<uint8_t>& key = {}) {
```
avikivity, line 439: And here a span is more flexible than a vector.

src/net/tls.cc
```diff
@@ -339,9 +339,9 @@ future<tls::x509_cert> tls::x509_cert::from_file(
 
 // wrapper for gnutls_datum, with raii free
 struct gnutls_datum : public gnutls_datum_t {
-    gnutls_datum() {
-        data = nullptr;
-        size = 0;
+    gnutls_datum(void* d = nullptr, size_t s = 0) {
+        data = reinterpret_cast<unsigned char*>(d);
+        size = s;
```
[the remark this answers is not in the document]
elcallio: Better solution is to make the constructor only take a size and do the alloc internally, as the memory must be allocated by gnutls_malloc. Better to enforce.

src/net/tls.cc
elcallio: Seems like a huge overkill for an internal call. But sure.

Issue #2697, opened 2025-03-25
seastar could crash if a stall report is emitted at an inopportune time


travisdowns:
The stall reporter uses `::backtrace` from a signal, and this method is not _officially_ signal safe. There are mixed reports of its "practical" signal safety, including other issues here in this repo, but ultimately it seems that it is not, in fact, safe. In particular, `::backtrace` shares machinery with libgcc exception unwind machinery, and if a stall report is emitted at a critical time when the interrupted fiber is unwinding the stack due to an exception, the stall report back crash in backtrace. This is especially hard to diagnose because the SIGSEGV handler is also not able to emit a backtrace in this case, as it crashes during backtrace in the same way (but I have a partial fix for this which I will push soon).

Here is a reproducer:

```
constexpr int max_depth = 2;

struct state {
    int caught = 0, top = 0, thrown = 0;
};

[[gnu::noinline]]
void impl(state& s, int x) {
    ++s.top;
    if (x <= 0) {
        throw std::runtime_error("foo");
    } else {
        try {
            impl(s, x - 1);
        } catch (...) {
            ++s.caught;
            ++s.thrown;
            throw;
        }
    }
}

SEASTAR_THREAD_TEST_CASE(stall_detector_crash) {
  auto total_iters = 100000000;
  auto now = [] { return std::chrono::high_resolution_clock::now(); };

  auto next_yield = now();
  state s;
  for (int a = 0; a < total_iters; a++) {
    if (now() > next_yield) {
        thread::yield();
        next_yield = now() + 10ms;
    }

    try {
        impl(s, a % max_depth);
    } catch (...) {
    }

    if (a % 100000 == 0) {
        fmt::print("Making progress: {:6.3f}%, top: {}, caught: {}\n", 100. * a / total_iters, s.top, s.caught);
    }
  }
}
```

Compile with clang-18 and run with `-- --blocked-reactor-reports-per-minute=99999 --blocked-reactor-notify-ms=1` and it should crash almost immediately. Note that it can crash also with default parameters for the stall notifier, these ones just help it happen quickly.

rockwotj:
Is this fixable by not emitting stack traces if `std::uncaught_exceptions() > 0`? Or is something more sophisticated possible?

travisdowns:
Yeah that's the plan.

PR #2712, opened 2025-04-01
Print incrementally in sigsegv handler

### LINKED ISSUE #2710: segault logging could be more robust
If a crash of the type described in #2697 occurs, we don't log the backtrace, because `::backtrace` crashes in the handler (this the same kind of crash that triggered the original segfault in this case).

Some of this was addressed in #2691 but we can we can make the segfault logger more robust by logging the segfault frames incrementally, which will help pinpoint the crash.

travisdowns:
If we crash due to segfault, we want to print the backtrace incrementally, frame-by-frame, rather than all at once. This is because the backtrace call may crash at some problematic frame, in which case we get no output at all: these are very hard to diagnose.

Instead, introduce an "immediate" mode to backtrace buffer which prints all appended strings immediately, and add an incremental mode to the backtrace function which calls ::backtrace repeatedly, increasing the number of requested frames by 1 each time. If we crash at some frame N we will have printed all frames 1..N-1 already.

Before this change the backtrace looks like this using the reproducer in scylladb/seastar#2697:

```
Backtrace:
```

After it looks like:

```
Backtrace:
  /home/tdowns/dev/seastar/build-clang-18/dev/libseastar.so+0x36ce6a
  /home/tdowns/dev/seastar/build-clang-18/dev/libseastar.so+0x32fbba
  /home/tdowns/dev/seastar/build-clang-18/dev/libseastar.so+0x342d0b
  /home/tdowns/dev/seastar/build-clang-18/dev/libseastar.so+0x35cf6c
  /lib/x86_64-linux-gnu/libc.so.6+0x4251f
  /lib/x86_64-linux-gnu/libgcc_s.so.1+0x17585
  /lib/x86_64-linux-gnu/libc.so.6+0x133bb2
  /home/tdowns/dev/seastar/build-clang-18/dev/libseastar.so+0x36d0d7
  /home/tdowns/dev/seastar/build-clang-18/dev/libseastar.so+0x32fb84
  /home/tdowns/dev/seastar/build-clang-18/dev/libseastar.so+0x32f0e7
  /home/tdowns/dev/seastar/build-clang-18/dev/libseastar.so+0x330281
  /lib/x86_64-linux-gnu/libc.so.6+0x4251f
  /lib/x86_64-linux-gnu/libgcc_s.so.1+0x14789
  /lib/x86_64-linux-gnu/libgcc_s.so.1+0x16f54
  /home/tdowns/dev/seastar/build-clang-18/dev/libseastar.so+0x2b98a8
  /lib/x86_64-linux-gnu/libgcc_s.so.1+0x173b8
```

Fixes #2710.

### DIFF
--- a/include/seastar/util/backtrace.hh
+++ b/include/seastar/util/backtrace.hh
@@ -61,16 +61,36 @@ bool operator==(const frame& a, const frame& b) noexcept;
 frame decorate(uintptr_t addr) noexcept;
 
 // Invokes func for each frame passing it as argument.
+// incremental=false is the default mode and simply calls ::backtrace once
+// and then calls func for each frame. If the ::backtrace call crashes,
+// func will not be called.
+// incremental=true will call ::backtrace in a loop, increasing the number of
+// frames to capture by one each time. This is useful for cases where
+// ::backtrace itself may crash, e.g., in a crash handler, in this case the
+// func will be called for each frame which didn't crash. See #2710. This
+// is much slower, however.
 SEASTAR_MODULE_EXPORT
 template<typename Func>
-void backtrace(Func&& func) noexcept(noexcept(func(frame()))) {
+void backtrace(Func&& func, bool incremental = false) noexcept(noexcept(func(frame()))) {
 #ifdef HAVE_EXECINFO
     constexpr size_t max_backtrace = 100;
     void* buffer[max_backtrace];
-    int n = ::backtrace(buffer, max_backtrace);
-    for (int i = 0; i < n; ++i) {
-        auto ip = reinterpret_cast<uintptr_t>(buffer[i]);
-        func(decorate(ip - 1));
+
+    if (incremental) {
+        for (size_t last_frame = 1; last_frame <= max_backtrace; ++last_frame) {
+            int n = ::backtrace(buffer, last_frame);
+            if (n < static_cast<int>(last_frame)) {
+                return;
+            }
+            auto ip = reinterpret_cast<uintptr_t>(buffer[last_frame-1]);
+            func(decorate(ip - 1));
+        }
+    } else {
+        int n = ::backtrace(buffer, max_backtrace);
+        for (int i = 0; i < n; ++i) {
+            auto ip = reinterpret_cast<uintptr_t>(buffer[i]);
+            func(decorate(ip - 1));
+        }
     }
 #else
 // Not implemented yet
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -766,8 +766,11 @@ class backtrace_buffer {
     static constexpr unsigned _max_size = 8 << 10;
     unsigned _pos = 0;
     char _buf[_max_size];
+    bool _immediate;
 public:
-    backtrace_buffer() = default;
+    backtrace_buffer(bool immediate = false)
+        : _immediate{immediate} {}
+
     ~backtrace_buffer() {
         flush();
     }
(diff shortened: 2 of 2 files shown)


src/core/reactor.cc
```diff
@@ -4042,18 +4061,20 @@ static void sigsegv_action(siginfo_t *info, ucontext_t* uc) noexcept {
     // with addr2line and other tools which expect an
     // address without any added offset (from ASLR or
     // because it's a relocated shared object).
-    print_safe("Segmentation fault: resolved ip: ");
+    print_safe("Segmentation fault: resolved ip: 0x");
```
travisdowns, line 4064: Make it clear these are hex.

include/seastar/util/backtrace.hh
```diff
@@ -61,16 +61,36 @@ bool operator==(const frame& a, const frame& b) noexcept;
...
 #ifdef HAVE_EXECINFO
     constexpr size_t max_backtrace = 100;
     void* buffer[max_backtrace];
-    int n = ::backtrace(buffer, max_backtrace);
-    for (int i = 0; i < n; ++i) {
-        auto ip = reinterpret_cast<uintptr_t>(buffer[i]);
-        func(decorate(ip - 1));
+
+    if (incremental) {
+        for (size_t last_frame = 1; last_frame <= max_backtrace; ++last_frame) {
```
travisdowns, line 80: In practice, in the cases we've seen this the crash happens fairly far down the stack, and the non-crashing frames are enough to diagnose or go a long way to understanding the problem (e.g., in the stall detector case the full stall detector backtrace appears).

src/core/reactor.cc
xemul: Nice, but it doesn't fit the "Print incrementally in sigsegv handler" patch title

src/core/reactor.cc
```diff
@@ -825,15 +835,15 @@ class backtrace_buffer {
...
+        }, _immediate);
     }
 
     void append_backtrace_oneline() noexcept {
         backtrace([this] (frame f) noexcept {
             reserve(3 + sizeof(f.addr) * 2);
             append(" 0x");
             append_hex(f.addr);
-        });
+        }, _immediate);
```
xemul, line 846: Thinking out loud. This transparently converts "immediate" to "incremental". Oh, well

xemul: > Instead, introduce an "immediate" mode to backtrace buffer which prints all appended strings immediately

Hopefully it won't intermix these lines with more logs (from other shards?) thus making the whole thing unreadable. On the other hand, SIGSEGV shouldn't be easy to debug, and having intermixed lines is still better than having nothing at all

travisdowns: > Thinking out loud. This transparently converts "immediate" to "incremental". Oh, well

Yeah. I felt that these were the right names for the two concepts, as they are not the same and could in principle be set independently, but here in the only existing usage "immediate" directly implies "incremetal". I could imagine a future where incremental would be a separate option inside the backtrace buffer in case you wanted say "non-immediate, incremental" or whatever.

travisdowns: > Hopefully it won't intermix these lines with more logs (from other shards?) thus making the whole thing unreadable.

Yeah. On the one hand it's possible, on the other hand it will hopefully be rare as we do make all these calls back-to-back with no delay in between so the window to race with concurrent output on another shard is small so I expect tearing to be rare in practice, and in on the third hand:

> On the other hand, SIGSEGV shouldn't be easy to debug, and having intermixed lines is still better than having nothing at all

Yup, exactly.

travisdowns: BTW, this is one reason to do this only on the segfault path, not on the other paths (abort handler, reactor stall notifier, etc).

src/core/reactor.cc
travisdowns: Busted. Would you like me to split it out into a new commit or extend the title/commit description?

travisdowns: @xemul I have split this out into a separate PR as requested: https://github.com/scylladb/seastar/pull/2727

Issue #2696, opened 2025-03-25
Horrible write IO performance on older kernels with the default of --poll-aio=1


StephanDollberg:
Hi team,

we are seeing abysmal write IO performance on older kernels. Specifically this at least affects versions up to and including 4.18 which is what RHEL-8 uses and hence is still quite widespread unfortunately.

The issue can be shown easily using iotune. All the below tests are run on an AWS i3en.xlarge instance. Note I am using a patched iotune version that allows manually setting io depth ([PR](https://github.com/scylladb/seastar/pull/2695)).

Default config:

```
[ec2-user@rhel8 ~]$ sudo /opt/redpanda/bin/iotune-redpanda --evaluation-dir /mnt/vectorized/redpanda/ --force-io-depth 1 -c 1
INFO  2025-03-25 12:08:40,782 seastar - Reactor backend: epoll
WARN  2025-03-25 12:08:40,789 seastar - Unable to set SCHED_FIFO scheduling policy for timer thread; latency impact possible. Try adding CAP_SYS_NICE
INFO  2025-03-25 12:08:40,838 [shard 0:main] iotune - /mnt/vectorized/redpanda/ passed sanity checks
INFO  2025-03-25 12:08:40,839 [shard 0:main] iotune - Disk parameters: max_iodepth=1 disks_per_array=1 minimum_io_size=512
INFO  2025-03-25 12:08:40,840 [shard 0:main] iotune - Filesystem parameters: read alignment 512, write alignment 4096
Starting Evaluation. This may take a while...
Measuring sequential write bandwidth: 314 MB/s (deviation 4%)
Measuring sequential read bandwidth: 710 MB/s (deviation 42%)
Measuring random write IOPS: 91 IOPS
Measuring random read IOPS: 11190 IOPS
```

Out of the box we are getting a measly 91 IOPS!

Comparing this to "newer" kernels such as 5.14 on RHEL9 we see no such issue:

```
[ec2-user@rhel9 ~]$ sudo /opt/redpanda/bin/iotune-redpanda --evaluation-dir /mnt/vectorized/redpanda/ --force-io-depth 1 -c 1
INFO  2025-03-25 12:09:32,538 seastar - Reactor backend: linux-aio
INFO  2025-03-25 12:09:32,596 [shard 0:main] iotune - /mnt/vectorized/redpanda/ passed sanity checks
[...]

PR #2715, opened 2025-04-03
reactor: Disable hot polling if wakeup granularity is too high

StephanDollberg:
We are seeing an issue where rand write performance under performs
massively (100x).

This is caused by the linux dio kernel thread/workqueue being starved
and hence aio write completitions aren't being served in a timely
manner.

This doesn't happen using "default" linux settings but only if
`/proc/sys/kernel/sched_wakeup_granularity_ns` or
(`/sys/kernel/debug/sched/wakeup_granularity_ns` on newer kernels) is
raised.

Specifically this effect can be observed on RHEL-8 as the `tuned`
version that ships with it sets this value to 15000000 but can
reproduced on any other system by just bumping that value.

This patch tries to detect this being the case and if so it will warn
and disable hot polling (both `--poll-aio` and `--idle-poll-time-us`)
which gives back the majority of the performance.

Note because this setting has moved to debug fs on newer kernels which
requires root rights to read it's actually not very likely that we will
be able to detect it on these. However, RHEL8 uses an older kernel and
is likely the major offender to run into this bug (we have had multiple
customers run into this).

Ref https://github.com/scylladb/seastar/issues/2696

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3759,6 +3759,30 @@ static bool kernel_supports_aio_fsync() {
     return internal::kernel_uname().whitelisted({"4.18"});
 }
 
+static std::tuple<std::filesystem::path, uint64_t> wakeup_granularity() {
+    auto try_read = [] (auto path) -> uint64_t {
+        try {
+            return read_first_line_as<uint64_t>(path);
+        } catch (...) {
+            return 0;
+        }
+    };
+
+    auto legacy_path = "/proc/sys/kernel/sched_wakeup_granularity_ns";
+    if (auto val = try_read(legacy_path); val) {
+        return {legacy_path, val};
+    }
+
+    // This will in practice almost always fail because debug fs requires root
+    // perms to read so we are out of luck
+    auto debug_fs_path = "/sys/kernel/debug/sched/wakeup_granularity_ns";
+    if (auto val = try_read(legacy_path); val) {
+        return {debug_fs_path, val};
+    }
+
+    return {"", 0};
+}
+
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -4492,6 +4516,22 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
         .no_poll_aio = !reactor_opts.poll_aio.get_value() || (reactor_opts.poll_aio.defaulted() && reactor_opts.overprovisioned),
     };
 
+    // Disable hot polling if sched wakeup granularity is too high
+    // dio thread will be starved otherwise
+    // see https://github.com/scylladb/seastar/issues/2696
+    if (!reactor_cfg.no_poll_aio || reactor_cfg.max_poll_time != 0us) {
+        auto [wakeup_file, granularity] = wakeup_granularity();
```
avikivity, line 4523: I don't think max_poll_time is relevant. If no_poll_aio is false (default) we enter the if and don't care about max_poll_time. If it's true, then we may still want to spin for a few microseconds before entering the kernel to avoid the expensive sleep path.

avikivity: I guess the spin still takes some time and I think the default is 100usec on a virtualized environment. Not sure what's best. Since it's a bad case, anyway, maybe setting it to zero is the safer step.

Note with tuned (not tuned(8)) setting of 0.5ms scheduling granularity and 100us spins,  we can spend a large fraction of time spinning.

StephanDollberg: Yeah as per my tests in the issue thread changing `--idle-poll-time-us` definitely does have an additional impact on top of `--poll-aio` but yes the latter is the more important one.

Issue #964, opened 2021-10-31
HTTP reply's API for setting content type is a mess


nyh:
While trying to make advanced usage of Seastar's HTTP server's mechanism for specifying the content-type header in the response I realized it is a mess:

1. `reply::set_content_type()`, despite its rather self-explanatory name, doesn't actually let you set the content-type as you wish! Rather, you need to pass it one of few fixed file extensions, and the full content_type header gets chosen for you. If you pass something which is *not* a known file extension, you get the `text/plain` content type - so something like `set_content_type("application/xyz") won't do what you think at all! Why can't we fall back to sending the given string instead of text/plain? Especially if the given string begins in known prefixes like "application/" or "text/"?
2. So for a content-type which doesn't have a known file extension, you need to use `reply::set_mime_type()` which actually sets Content-Type ;-)
3. A bunch of other functions **force** or just **mislead** you to use `set_content_type()` instead of the `set_mime_type()` that you may want to use. When  `done()` takes a string, it calls `set_content_type()` with it. Luckily, there is also an argument-less variant of `done()` which avoids this call so the user can call `set_mime_type` and then the argument-less `done()`. We're less lucky in the `reply::write_body()` function - which always calls set_content_type() (there is no variant without it), so the caller is forced to first set the some random content-type and then override it with something else, which is weird and ugly.
 
I think I managed to work around all these problems in my code, so it's not that any feature is really _missing_. It's just that users shouldn't need to pass random parameters to the  API and battle with misleadingly-named methods to get their code working.

nyh:
As you can see in the MIME  [RFC 2045, section 5.1](https://datatracker.ietf.org/doc/html/rfc2045#section-5.1), a full MIME type always needs to look like `type/subtype`, e.g., `application/json`. A string without a slash, like "json", cannot be a full MIME type so it makes sense to look it up in some sort of extension table (a la Unix's /etc/mime.types).

So I propose doing the following:
1. Remove one of the two functions set_content_type() or set_mime_type(). Both are reasonable names, I personally prefer leaving `set_content_type()`. Change all the calls in Seastar to the removed function to call the remaining one, and leave the removed function as a deprecated alias to the remaining one.
2. In the remaining function, if there is a slash in the given content-type, use it directly. Otherwise, use the extension lookup table - as in today's `set_content_type()`.

nyh:
[...]

PR #2725, opened 2025-04-08
`http`: add `xml` support to `http::mime_types::mappings`

WillemKauf:
It seems strange that `xml` isn't supported as a `mime` type here. 

In particular, we expect all [S3 REST error responses](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#RESTErrorResponses) to be returned with a `Content-Type` header of `application/xml`, so being able to map the extension makes life a lot easier for testing with imposter `http` S3 servers/services.

### DIFF
--- a/src/http/mime_types.cc
+++ b/src/http/mime_types.cc
@@ -25,6 +25,7 @@ struct mapping {
     const char* mime_type;
 } mappings[] = {
         { "json", "application/json"},
+	{ "xml", "application/xml"},
         { "gif", "image/gif" },
         { "htm", "text/html" },
         { "css", "text/css" },


avikivity (approved):

xemul:
There's `set_mime_type()` that can be used to set whatever "raw" type you need. What's the exact issue you're facing? @WillemKauf

WillemKauf:
> What's the exact issue you're facing?

Similar to @nyh's problem above, using `function_handler` to mock HTTP responses doesn't allow for an `application/xml` `Content-Type` header, since `rep->done(_type)` is called. 

https://github.com/scylladb/seastar/blob/f05eb1e118ed369d5579f39a9790ea3c67fc66da/include/seastar/http/function_handlers.hh#L111

I'm getting around this currently via a new impl for `handler_base`, but I think the mapping is worth adding to avoid having to do this.

nyh:
I have an idea: What if when the mime type contains a slash, we use it **without** looking it up in any tables?
I don't see why Seastar needs to contain a long list of popular (and unpopular) mime types if the application can supply the full mime type.

xemul:
> I have an idea: What if when the mime type contains a slash, we use it without looking it up in any tables?

You mean the `write_body()` methods? Because if you use `set_mime_type` directly it doesn't look up any tables already. @nyh

xemul:
> > I have an idea: What if when the mime type contains a slash, we use it without looking it up in any tables?
> 
> You mean the `write_body()` methods? Because if you use `set_mime_type` directly it doesn't look up any tables already. @nyh

Ah, it's the `function_handler::done()`. OK

nyh:
> > I have an idea: What if when the mime type contains a slash, we use it without looking it up in any tables?
> 
> You mean the `write_body()` methods?

Yes, `write_body()` calls `set_content_type()` instead of `set_mime_type()` - and `set_content_type()`  is the one that does the lookup and calls `set_content_type()`.
Also `done()` calls `set_content_type()` instead of `set_mime_type()`.

I claim that this distinction, between "content type" and "mime type" is simply unnecessary and even misleading (as the set_content_type() today specifies the **extension** not a MIME or HTTP Content-Type). As you can see in the MIME RFC 2045 section 5.1, https://datatracker.ietf.org/doc/html/rfc2045#section-5.1, a full MIME type always needs to look like type/subtype. E.g., `application/json`. A string without a slash, like "json", cannot be a full MIME type so it makes sense to look it up in some sort of shortcut table (a la Unix's /etc/mime.types). 

I propose the following fix **instead** of this patch:
1. Remove one of the two functions `set_content_type()` or `set_mime_type()` (both are reasonable names), call only the second one.
2. Make the one function call the array lookup function only if there is no slash in the given type.

I think this fix will be very easy to do and will avoid a deluge of patches like this one each adding their favorite MIME type. I'll also use the new patch in ScyllaDB instead of an ugly workaround (as I explained above).

I'll open an issue asking for this.

nyh:
Don't need to open an issue - I already opened one: https://github.com/scylladb/seastar/issues/964
So if you follow my suggestion, your patch can also "Fixes #964".

nyh:
I see that despite my recommendation, Avi merged this match. 
I still claim that the list of extensions, whether or not it contains xml, is the wrong way to go, and we should do
https://github.com/scylladb/seastar/issues/964

I'll send a patch to do it.

PR #2729, opened 2025-04-08
http: generalize reply's Content-Type setting

### LINKED ISSUE #964: HTTP reply's API for setting content type is a mess

nyh:
For setting a content type on a reply in the HTTP server, we had two different functions - one set_mime_type() which accepted only full MIME types (e.g., "text/html"), and one set_content_type() which took a file extension (e.g, "html") and looked it up in a short map we had, like a tiny version of /etc/mime.types.

Even worse, some reply methods like write_body() and done() only called set_content_type() meaning they could only set content types from that short list of known types - and not any general MIME type. This forced Seastar users to jump through hoops to use additional mime types - for example call write_body() on a wrong type and then call set_mime_type() to override it, or patch the extension map to add more and more types to it.

This patch fixes this problem for good: With this patch, we no longer have two different reply methods - there is just one, set_content_type(). The name set_mime_type() remains a deprecated alias. The merged set_content_type() can recognize *either* a full-fledged MIME type like "text/html", or if there is no slash in the name - it is looked up in the list of extensions (e.g., "html").

While doing this, I also changed the set_content_type() from taking a "const sstring&" like in the old days, to the more modern std::string_view. I believe this will not cause problems for applications, as a const sstring& can be implicitly converted into an std::string_view so old code should continue to compile.

Fixes #964


nyh:
@WillemKauf I wrote this patch to make your "xml" patch (https://github.com/scylladb/seastar/pull/2725) and the likes of it unnecessary - with this patch every Seastar application will be able to set its own arbitrary MIME type without needing to ask to add it to some global list. This patch should also help ScyllaDB which right now had (alternator/server.cc) some ugly code doing write_body() with a wrong content type (one from the short list of known extensions) and then overriding it with set_content_type() to set the correct content type. With this patch, this nonsense will no longer be necessary.

For some unknown reason I wasn't able to build Seastar locally (?), so I'll use this CI to build it for me and fix whatever build errors arise.

nyh:
Pushed a new version fixing C++20 build. I forgot that C++20 didn't have `std::string_view::contains()`. Use `std::string_view::find()` instead.

amnonh (approved):
It's a good change.
We should open an issue in Seastar to remove more of the sstring.

I also saw that the mine-type lookup should have been a map and probably with an option to add more type to it, but that's out of the scope of this PR that looks good.

WillemKauf:
Awesome! Thank you @nyh.

nyh:
Pushed a new version that will hopefully fix the module build.

nyh:
> It's a good change. We should open an issue in Seastar to remove more of the sstring.

Yes, we should probably do this one day. Many years ago I opened an issue (https://github.com/scylladb/seastar/issues/634) hoping we could get rid of our "sstring" and use `std::string`, but using std::string_view everywhere instead of `const sstring&` is a good first step for accepting applications that prefer to use `std::string` in the application and want to reduce the amount of friction with Seastar's API.

> I also saw that the mine-type lookup should have been a map and probably with an option to add more type to it, but that's out of the scope of this PR that looks good.

I think the only reason to leave this extension-to-mime-type mapping list for serving of **static files** from the HTTP server. I don't think that it should ever be expanded to a huge list, or that a dynamic server should ever use this mapping. A dynamic server will know what MIME type it is replying, and should state it explicitly, not in shorthand.

nyh:
The module build failed again, pushed another attempt at fixing it (I am completely ignorant about modules, and don't know how to build the module build locally, unfortunately).

tchaikov:
@nyh hi Nadav, could you please take a look at #2743 ? i wanted to address the [build failure](https://github.com/scylladb/seastar/actions/runs/14353893577/job/40238795958?pr=2729) you uncounted in the workflow building the tree with C++20 modules support. i planned to write a doc on the modules support, but this might take a while.

nyh:
Rebased and pushed a new version.

@kreuzerkrieg noticed that this pull request has become even more important since a recent change to Seastar made it mandatory to cell `write_body()` to set content-length - and as I explained above `write_body()` before this patch forces you to choose from the very short (and franky, unusable) list of file extensions - which you usually don't want to do.

nyh:
Sadly, I can't get the modules to work. As you can see in the CI output, there are "conflicts" like
```
In file included from /usr/bin/../lib/gcc/x86_64-redhat-linux/15/../../../../include/c++/15/cwchar:49:
/usr/include/wchar.h:174:15: error: conflicting types for 'wcsxfrm_l'
  174 | extern size_t wcsxfrm_l (wchar_t *__s1, const wchar_t *__s2,
      |               ^
/usr/include/wchar.h:174:15: note: previous declaration is here
  174 | extern size_t wcsxfrm_l (wchar_t *__s1, const wchar_t *__s2,
      |               ^
```

This wchar stuff is included from `<string_view>`.

These error look nonsensical because the two definitions point to the same thing - so how can it be a conflict. I suspect this means that the module and the file were built with slightly different ifdefs while including wchar.h. But have no idea what might be causing this, and every guess I had didn't help.

nyh:
https://github.com/scylladb/seastar/pull/2729#issuecomment-2800321296 pointed me to a solution, which makes absolutely no sense to me (basically, reordering the include files) but worked. I'll push a new version with it.

nyh:
The build finally passed.
@kreuzerkrieg please review this PR, and hopefully a maintainer can finally merge it after four months :-)

include/seastar/http/reply.hh
```diff
@@ -200,22 +201,29 @@ struct reply {
...
+     * Set the content type. The content_type string can be one of:
+     * 1. A MIME (RFC 2045) Content-Type - looking like "type/subtype", e.g.,
+     *   "text/html"
+     * 2. If a "/" is missing in the given string, we look it up in a list of
+     *    common file extensions listed in the http::mime_types map. For
+     *    example "html" will be mapped to "text/html".
      */
-    reply& set_mime_type(const sstring& mime) {
-        _headers["Content-Type"] = mime;
+    reply& set_content_type(std::string_view content_type) {
```
kreuzerkrieg, line 211: what about `seastar::http::request::set_content_type`?

nyh: Good question, I didn't even know it existed, but I should have guessed! :-(
My personal interest was in an HTTP **server**, and this is the only thing we originally had in Seastar, but I guess we grew to clients as well. I'll take a look :-(

nyh: Note that the patch's title, the commit message, and even the 4 year old issue, were all about HTTP server, not http client....

nyh: I'll add a second patch to this pull request, with a http::request change...

kreuzerkrieg: right, but the problem is symetric for both. I guess we can do it in separate PR
In any case, it is up to you, will approve now

nyh:
@kreuzerkrieg I added a second patch for the HTTP client (the seastar::http::request class).

src/http/mime_types.cc
```diff
@@ -40,7 +46,7 @@ struct mapping {
         { "proto", "application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited"},
 };
 
-const char* extension_to_type(const sstring& extension)
+const char* extension_to_type(std::string_view extension)
 {
     for (mapping m : mappings) {
         if (extension == m.extension) {
```
kreuzerkrieg, line 52: If we are already changing the function signature (BTW, it is not under `experimental` namespace) maybe then we will change also the return type and move the 
```
if (content_type.find('/') == std::string_view::npos) {
            content_type = http::mime_types::extension_to_type(content_type);
        }
        _headers["Content-Type"] = sstring(content_type);
```
part here? something like
```
sstring extension_to_type(std::string_view extension)
{
    if (extension.find('/') == std::string_view::npos) {
        for (mapping m : mappings) {
            if (extension == m.extension) {
                return m.mime_type;
            }
        }
        return "text/plain";
    }
    return sstring(extension);
}
```

kreuzerkrieg (approved):

src/http/mime_types.cc
nyh: But what's the benefit? I didn't really want to change extension_to_type, I wanted to change its caller - to get std::string_view instead of sstring&.
In any case, the caller converts the returned char* to sstring, so what's the benefit of doing this inside extension_to_type()?

kreuzerkrieg: > But what's the benefit? I didn't really want to change extension_to_type, I wanted to change its caller - to get std::string_view instead of sstring&. In any case, the caller converts the returned char* to sstring, so what's the benefit of doing this inside extension_to_type()?

just to save on code/comment duplicate in xxx::set_content_type. but I dont insist

nyh:
Three of the tests were "cancelled", I don't know why or how, and I only managed to restart one of them. Looks like I'll need to just push the branch again to get the CI to run automatically because I don't understand how to restart the other failing runs.

src/http/mime_types.cc
nyh: I didn't understand what you mean... We need the long comment in front of set_content_type() because it's public API and I want to explain to users what it does. I can't move that comment to some inner function that nobody uses directly.

The 4 lines of code common to both set_content_type9() I agree is duplicated, but given that the entire request and reply classes are near duplicates (but not exact...), so I don't think this is a problem.

kreuzerkrieg:
@scylladb/scylla-maint could we push it forward please?

PR #2730, opened 2025-04-09
streams: Introduce `closeable_file_data_source_impl` class

kreuzerkrieg:
Add a `datasource` creator function for generating `closeable_file_data_source_impl` instances. This class mirrors `file_data_source_impl` in functionality but overrides the `close()` method to ensure that the associated file is properly closed when the `data_source` is closed.


src/core/fstream.cc
```diff
@@ -365,6 +368,20 @@ input_stream<char> make_file_input_stream(
...
+    closeable_file_data_source_impl(file f, uint64_t offset, uint64_t len, file_input_stream_options options):file_data_source_impl(std::move(f),  offset,  len,  std::move(options)) {
+
+    }
+    future<> close() override {
+        co_await file_data_source_impl::close();
+        co_return co_await _file.close();
+    }
+};
+
+data_source make_closeable_file_data_source(file f, uint64_t offset, uint64_t len, file_input_stream_options opt) {
```
xemul, line 382: It's better to put `bool_class<something> close_file>` bit into `file_input_stream_options`

xemul (changes requested):

PR #2731, opened 2025-04-09
http_client: Improve handling of server response parsing errors

kreuzerkrieg:
Handle cases where the server response might be broken due to a server-side bug or transient error. Retrying in such cases may still be meaningful. Replace `runtime_error`, which prevents retries, with `system_error`, allowing the client to retry the request once using a new connection.

Enhance the response parser to generate more verbose error messages, providing additional context that may help identify the root cause of the issue.

This change addresses an observed issue where the MinIO server returns garbage responses (potentially a server bug), but the same request succeeds when retried.

### DIFF
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -142,7 +142,7 @@ future<connection::reply_ptr> connection::recv_reply() {
             }
             if (parser.failed()) {
                 http_log.trace("Parsing response failed");
-                throw std::runtime_error("Invalid http server response");
+                throw httpd::response_parsing_exception(format("Invalid http server response. Reason: {}", parser.error_message()));
             }
 
             auto resp = parser.get_parsed_response();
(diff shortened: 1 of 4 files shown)


src/http/client.cc
```diff
@@ -141,8 +141,17 @@ future<connection::reply_ptr> connection::recv_reply() {
...
             }
             if (parser.failed()) {
-                http_log.trace("Parsing response failed");
-                throw std::runtime_error("Invalid http server response");
+                http_log.warn("Parsing response failed.{}", http_log.level() == log_level::debug ? format(" Reason: {}", parser.error_message()) : "");
+                // Parsing the response failed, potentially due to a server-side bug or
+                // a transient error. Retrying might resolve the issue. To trigger a retry,
+                // return EPIPE to the caller. The original exception is retained as a
+                // nested exception so that, in case retries keep failing, the caller can
+                // identify the actual cause unrelated to the EPIPE error.
```
xemul, line 149: How would caller identify the actual error if it's going to be system_error anyway? Also, the error message is not in the exception, so the caller is still blind.

src/http/client.cc
```diff
@@ -141,8 +141,17 @@ future<connection::reply_ptr> connection::recv_reply() {
...
+                try {
+                    throw std::system_error(EPIPE, std::system_category());
```
xemul, line 151: Let's better throw a dedicated error-parsing-http-server-response-line-exception thing. It will carry the error message on board, and it will be explicitly caught by the restart checking code.

src/http/client.cc
```diff
@@ -141,8 +141,17 @@ future<connection::reply_ptr> connection::recv_reply() {
                 throw std::system_error(ECONNABORTED, std::system_category());
             }
             if (parser.failed()) {
-                http_log.trace("Parsing response failed");
-                throw std::runtime_error("Invalid http server response");
+                http_log.warn("Parsing response failed.{}", http_log.level() == log_level::debug ? format(" Reason: {}", parser.error_message()) : "");
```
xemul, line 144: Not sure if it's nice to spam logs with all server-related errors that are going to be "workarounded" anyway. If applying [this](https://github.com/scylladb/seastar/pull/2731/commits/8eea143389d341ac5ddbd0e3cb1cd5bf2c8736ba#r2035044584) suggestion, the exact parsing error can be printed by the caller in case retry is not performed.

tests/unit/httpd_test.cc
```diff
@@ -1690,6 +1690,18 @@ SEASTAR_TEST_CASE(case_insensitive_header) {
     return make_ready_future<>();
 }
 
+SEASTAR_TEST_CASE(broken_reply) {
+    http_response_parser parser;
+    parser.init();
+    char r101[] = "Lorem ipsum dolor sit amet, consectetur adipiscing elit.";
```
regevran, line 1696: The most strange usage of `Lorem ipsum` I have ever saw!

tests/unit/httpd_test.cc
```diff
@@ -1690,6 +1690,18 @@ SEASTAR_TEST_CASE(case_insensitive_header) {
...
+
+    parser.parse(r101, r101 + sizeof(r101), r101 + sizeof(r101));
+    BOOST_REQUIRE_EQUAL(parser.failed(), true);
+    BOOST_REQUIRE_EQUAL(parser.error_message().starts_with("Parsing error at offset 0: encountered \"Lorem ipsum dolor sit amet, cons\""), true);
```
regevran, line 1700: I would add a test to check the very last byte as well.

regevran (approved):
Please address @xemul's valid concerns.

src/http/client.cc
kreuzerkrieg: and then we just want to add another condition to the `if` [here](https://github.com/scylladb/seastar/blob/master/src/http/client.cc#L339) instead of reusing existing nested exception thingy?

src/http/client.cc
kreuzerkrieg: then just 
```
http_log.trace("Parsing response failed");
try {
        throw std::system_error(EPIPE, std::system_category());
} catch (...) {
        std::throw_with_nested(std::runtime_error(format("Invalid http server response. Reason: {}", parser.error_message())));
}
```
will suffice I guess, lemme check

tests/unit/httpd_test.cc
kreuzerkrieg: very last byte of the message is a string literal

tests/unit/httpd_test.cc
kreuzerkrieg: Do you prefer "Quick brown fox..."? :)

src/http/client.cc
kreuzerkrieg: how about now?

src/http/client.cc
xemul: Yes. It's not great to masquerade parsing error with EPIPE exception, because it's not such

xemul:
Patch comment needs to be updated (it says that the parser now throws system_error, but it really doesn't)

tests/unit/httpd_test.cc
```diff
@@ -1188,7 +1188,7 @@ SEASTAR_TEST_CASE(test_client_retry_request) {
                 input_stream<char> in = sk.input();
                 read_simple_http_request(in);
                 output_stream<char> out = sk.output();
-                out.write("HTT").get(); // write incomplete response
+                out.write("HTTTT").get(); // write incomplete response
```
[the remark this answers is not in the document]
kreuzerkrieg: Coz it partially matches "HTTP" and then you will get EOF instead of invalid sequence

kreuzerkrieg:
> Patch comment needs to be updated (it says that the parser now throws system_error, but it really doesn't)

done

PR #2732, opened 2025-04-09
smp: expose some NUMA information to the application

michoecho:
For performance reasons, the application might want to do some data structure sharing between nodes, but only if it doesn't cause cross-NUMA-node memory accesses.

To allow for that, expose NUMA information to the application.

Concretely: expose a map from shard ID to an ID of the NUMA node containing this shard's memory.
(If there's no dedicated memory allocation, then the NUMA node ID is 0).


michoecho:
What kind of unholy spaghetti did I just step into?

If the `_shard_to_numa_node_mapping.resize(smp::count);` call happens before `auto inited = std::make_shared<boost::barrier>(smp::count);`, and `smp::count` is smaller than 3, then `httpd_test` times out.

avikivity:
> What kind of unholy spaghetti did I just step into?
> 
> If the `_shard_to_numa_node_mapping.resize(smp::count);` call happens before `auto inited = std::make_shared<boost::barrier>(smp::count);`, and `smp::count` is smaller than 3, then `httpd_test` times out.

The setup code is perhaps not the cleanest there ever was.

include/seastar/core/smp.hh
```diff
@@ -324,6 +324,7 @@ class smp : public std::enable_shared_from_this<smp> {
     static thread_local smp_message_queue**_qs;
     static thread_local std::thread::id _tmain;
     bool _using_dpdk = false;
+    std::vector<unsigned> _shard_to_numa_node_mapping;
```
avikivity, line 327: It would make more sense to return this shard's mapping rather than the entire smp. You can add an accessor to return smp instead.

avikivity: This comment belongs in reactor.hh, not here

michoecho:
> > What kind of unholy spaghetti did I just step into?
> > If the `_shard_to_numa_node_mapping.resize(smp::count);` call happens before `auto inited = std::make_shared<boost::barrier>(smp::count);`, and `smp::count` is smaller than 3, then `httpd_test` times out.
> 
> The setup code is perhaps not the cleanest there ever was.

It turned out to be a bug in `httpd_test`. One of the tests sends an uninitialized `temporary_buffer<char>`, and the custom parser in the test breaks on `\0` bytes. It happened to work by coincidence, but my change disturbed the allocation pattern slightly, modified the contents of the buffer, and broke the test.

And it took a few hours to debug. I hate programming sometimes.

And I hate that Seastar can't telling me at which point the test hanged.

Fix at https://github.com/scylladb/seastar/pull/2736.

include/seastar/core/smp.hh
michoecho: Done in v3.

avikivity:
> > > What kind of unholy spaghetti did I just step into?
> > > If the `_shard_to_numa_node_mapping.resize(smp::count);` call happens before `auto inited = std::make_shared<boost::barrier>(smp::count);`, and `smp::count` is smaller than 3, then `httpd_test` times out.
> > 
> > 
> > The setup code is perhaps not the cleanest there ever was.
> 
> It turned out to be a bug in `httpd_test`. One of the tests sends an uninitialized `temporary_buffer<char>`, and the custom parser in the test breaks on `\0` bytes. It happened to work by coincidence, but my change disturbed the allocation pattern slightly, modified the contents of the buffer, and broke the test.
> 
> And it took a few hours to debug. I hate that Seastar doesn't have an ability to say which task has hanged, and where. And why don't sanitizers complain that we are reading from uninitialized allocations?
> 


Good question. Is that temporary_buffer::temporary_buffer(size_t)? It does nothing to unpoison the allocation.

michoecho:
> > > > What kind of unholy spaghetti did I just step into?
> > > > If the `_shard_to_numa_node_mapping.resize(smp::count);` call happens before `auto inited = std::make_shared<boost::barrier>(smp::count);`, and `smp::count` is smaller than 3, then `httpd_test` times out.
> > > 
> > > 
> > > The setup code is perhaps not the cleanest there ever was.
> > 
> > 
> > It turned out to be a bug in `httpd_test`. One of the tests sends an uninitialized `temporary_buffer<char>`, and the custom parser in the test breaks on `\0` bytes. It happened to work by coincidence, but my change disturbed the allocation pattern slightly, modified the contents of the buffer, and broke the test.
> > And it took a few hours to debug. I hate that Seastar doesn't have an ability to say which task has hanged, and where. And why don't sanitizers complain that we are reading from uninitialized allocations?
> 
> Good question. Is that temporary_buffer::temporary_buffer(size_t)? It does nothing to unpoison the allocation.

Yes, it's the buffer at https://github.com/scylladb/seastar/blob/b3f051c291aed495b59c8e29176e825ac1faf2a3/tests/unit/httpd_test.cc#L792

avikivity:
Theory: the sanitizers propagate the uninitialized status via memcpy, then it ends up in sendmsg(), where they forget about it.

include/seastar/core/reactor.hh
```diff
@@ -603,6 +603,9 @@ public:
     [[deprecated("Use this_shard_id")]]
     shard_id cpu_id() const;
 
+    /// \returns A integer span of size smp::count, with nth integer being the ID of nth shard's NUMA node.
+    const smp& get_smp() const noexcept;
```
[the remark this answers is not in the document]
michoecho: > Comment is now wrong.

Fixed.

> Also, get_ is unnecessary here, smp() is enough.

Done. I used `get_smp()` so that the name `smp` (when used inside `reactor` methods) remains unambiguous. But if you like `smp()` more, then so be it.

michoecho:
v4: fixed the comment, changed `reactor::get_smp()` to `reactor::smp()`.

michoecho:
> > Also, get_ is unnecessary here, smp() is enough.
> 
> Done. I used get_smp() so that the name smp (when used inside reactor methods) remains unambiguous. But if you like smp() more, then so be it.

Seems I left some unfixed ambiguities, will fix in a minute.

PR #2737, opened 2025-04-10
rpc: client: support timeout and cancellation

bhalevy:
Current we support either sending an rpc message
with timeout, or sending it with a `cancellable`
parameter that allows the caller to cancel the rpc.

The caller can implement a client-side timeout
using the cancellable object using an external timer, however, in this case, the timeout value is not
automatically communicated to the message handler
on the receiver side.

To allow users to cancel a verb that already
uses a timeout, add variants of the call operator
that accept both a timeout and a cancellable.

Also, add a respective unit test.

### DIFF
--- a/tests/unit/rpc_test.cc
+++ b/tests/unit/rpc_test.cc
@@ -42,6 +42,7 @@
 #include <seastar/util/closeable.hh>
 #include <seastar/util/noncopyable_function.hh>
 #include <seastar/util/later.hh>
+#include <seastar/core/coroutine.hh>
 
 #include <boost/range/numeric.hpp>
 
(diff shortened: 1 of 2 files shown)


bhalevy:
- unregister_handler in unit test

bhalevy:
See intended use case: https://github.com/bhalevy/scylla/commits/c151f2f1ac7ba84799ddce9ba8e0adc8ba5362bc/

xemul:
@gleb-cloudius , please review

bhalevy:
> @gleb-cloudius , please review

He's on vacation, returning next week.

bhalevy:
ping @gleb-cloudius please review

tests/unit/rpc_test.cc
```diff
@@ -1827,3 +1828,41 @@ SEASTAR_THREAD_TEST_CASE(test_compressor_empty_frames) {
...
+            cancel_rpc.cancel();
+            BOOST_REQUIRE_THROW(f.get(), rpc::canceled_error);
+        }
+        {
+            rpc::cancellable cancel_rpc;
+            auto f = echo(cl, seastar::lowres_clock::now() + std::chrono::milliseconds(10), cancel_rpc, 42);
+            BOOST_REQUIRE(!f.available());
+            cancel_rpc.cancel();
+            BOOST_REQUIRE_THROW(f.get(), rpc::canceled_error);
+        }
```
gleb-cloudius, line 1864: I would add a test that yields after sending an rpc to make sure to cover the case where we cancel rpc that is already in flight.

bhalevy: Added in [1838f5e](https://github.com/scylladb/seastar/commit/1838f5edf64f32b0f2e373043dae1c0cbc5b717a)

gleb-cloudius (approved):

PR #2738, opened 2025-04-11
Share IO queues between mountpoints

### LINKED ISSUE #2733: Teach io-scheduler to merge mountpoints into one io-queue
Sometimes a node is configured to have several virtual volumes (of whatever kind) sitting on the same underlying physical device.
As a result, seastar learns that it needs to use several distinct io-queue-s with its own properties, while in reality the underlying disk is still the same and seastar should have used one shared io-queue for them all

xemul:
Currently, IO-queues assignment is done per-mountpoint. The mountpoint paths are read from io-properties.yaml file, for each the underlying device number is stat()-ed, then devive -> io_queue mapping is established. When opening a file, seastar stat()-s the filepath and finds the io_queue for the found device number.

It's nonetheless possible to have one physical disk be "split" into several virtual block devices, and to mount each into its own mountpoint. In order to configure io-scheduler for this setup the only option is to have two mountpoints in io_properties.yaml. With that, seastar will create two io_queues for both mountpoints, but it's not correct, since underlying physical disk is still the same.

This PR extends the io-properties.yaml configuration so that one disk entry in it can specify a list of mountpoints this disk configuration applies to. When met such a record, seastar would create a single io_queue object, but will map all mountpoints' device numbers to it.

The change has two main parts. First, the smaller one, reactor's map<device, io_queue> now may have several devices point to one shared io_queue object. So than when two files are opened on different mountpoints with, respectively, different device numbers, they both will get reference to the same io_queue. Second, the larger change, is that io_queue internal "identification" changes from the underlying device number to abstract integer value. That's not a breaking change, however, device numbers were not present in any APIs or metric labels. Only in logs.

fixes #2733

### DIFF
--- a/doc/io-properties-file.md
+++ b/doc/io-properties-file.md
@@ -39,3 +39,23 @@ disks:
     write_bandwidth: 510M
     write_saturation_length: 64k
 ```
+
+Optionally, instead of the "mountpoint" there can be the "mountpoints" (plural)
+entry in the list element being a list itself and listing more than one path. As
+a result all the listed mountpoints will share the corresponding internal IO queue.
+
+Example:
+
+```
+disks:
+  - mountpoints:
+      - /var/lib/some_seastar_app/sub_one
+      - /var/lib/some_seastar_app/sub_two
+    read_iops: 95000
+    ...
+```
+
+An example when this configuration is applicable can be an LLVM set of volumes
+from one disk each being mounted at its own path. In that case, different mount
+points would have different (virtual) block devices, yet, they will share the
+same physical disk and thus need to run over one shared IO queue.
--- a/apps/io_tester/ioinfo.cc
+++ b/apps/io_tester/ioinfo.cc
@@ -58,9 +58,11 @@ int main(int ac, char** av) {
                             auto& ioq = engine().get_io_queue(st.st_dev);
                             auto& cfg = ioq.get_config();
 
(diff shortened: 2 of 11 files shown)


ruthea:
Thank you @xemul !

doc/io-properties-file.md
```diff
@@ -39,3 +39,23 @@ disks:
...
+```
+disks:
+  - mountpoints:
+      - /var/lib/some_seastar_app/sub_one
+      - /var/lib/some_seastar_app/sub_two
+    read_iops: 95000
+    ...
+```
+
+An example when this configuration is applicable can be an LLVM set of volumes
```
mykaul, line 58: LVM, not LLVM.

tchaikov: created #2748

regevran:
@xemul - please [backport](https://github.com/scylladb/scylla-enterprise/issues/5353#issuecomment-2817042269) to 2024.1

bhalevy:
> @xemul - please [backport](https://github.com/scylladb/scylla-enterprise/issues/5353#issuecomment-2817042269) to 2024.1

@xemul ping

regevran:
@avikivity  [said](https://github.com/scylladb/scylla-enterprise/issues/5353#issuecomment-2817102992) he will backport - please coordinate with him.

PR #2747, opened 2025-04-15
resource: Rework condition for sysconf fallback

xemul:
The resource allocation fallback code assumes that the number of numa nodes detected by hwloc is 1 (see b0a9f89a7). However, here's my node:

```
Machine (126GB total) + Package L#0
  Die L#0
    NUMANode L#0 (P#0 63GB)
    L3 L#0 (8192KB)
      L2 L#0 (512KB) + L1d L#0 (32KB) + L1i L#0 (64KB) + Core L#0
        PU L#0 (P#0)
        PU L#1 (P#24)
      L2 L#1 (512KB) + L1d L#1 (32KB) + L1i L#1 (64KB) + Core L#1
        PU L#2 (P#1)
        PU L#3 (P#25)
      L2 L#2 (512KB) + L1d L#2 (32KB) + L1i L#2 (64KB) + Core L#2
        PU L#4 (P#2)
        PU L#5 (P#26)
    L3 L#1 (8192KB)
      L2 L#3 (512KB) + L1d L#3 (32KB) + L1i L#3 (64KB) + Core L#3
        PU L#6 (P#3)
        PU L#7 (P#27)
      L2 L#4 (512KB) + L1d L#4 (32KB) + L1i L#4 (64KB) + Core L#4
        PU L#8 (P#4)
        PU L#9 (P#28)
      L2 L#5 (512KB) + L1d L#5 (32KB) + L1i L#5 (64KB) + Core L#5
        PU L#10 (P#5)
        PU L#11 (P#29)
    HostBridge
      PCIBridge
        PCI 01:00.1 (SATA)
        PCIBridge
          PCIBridge
            PCI 03:00.0 (Network)
              Net "wlp3s0"
          PCIBridge
            PCI 04:00.0 (Network)
              Net "wlp4s0"
          PCIBridge
            PCI 05:00.0 (Ethernet)
              Net "enp5s0"
          PCIBridge
            PCI 07:00.0 (Ethernet)
              Net "enp7s0"
      PCIBridge
        PCI 0a:00.2 (SATA)
  Die L#1
    NUMANode L#1 (P#2 63GB)
    L3 L#2 (8192KB)
      L2 L#6 (512KB) + L1d L#6 (32KB) + L1i L#6 (64KB) + Core L#6
        PU L#12 (P#6)
        PU L#13 (P#30)
      L2 L#7 (512KB) + L1d L#7 (32KB) + L1i L#7 (64KB) + Core L#7
        PU L#14 (P#7)
        PU L#15 (P#31)
      L2 L#8 (512KB) + L1d L#8 (32KB) + L1i L#8 (64KB) + Core L#8
        PU L#16 (P#8)
        PU L#17 (P#32)
    L3 L#3 (8192KB)
      L2 L#9 (512KB) + L1d L#9 (32KB) + L1i L#9 (64KB) + Core L#9
        PU L#18 (P#9)
        PU L#19 (P#33)
      L2 L#10 (512KB) + L1d L#10 (32KB) + L1i L#10 (64KB) + Core L#10
        PU L#20 (P#10)
        PU L#21 (P#34)
      L2 L#11 (512KB) + L1d L#11 (32KB) + L1i L#11 (64KB) + Core L#11
        PU L#22 (P#11)
        PU L#23 (P#35)
    HostBridge
      PCIBridge
        PCI 41:00.0 (NVMExp)
          Block(Disk) "nvme0n1"
      PCIBridge
        PCI 42:00.0 (VGA)
      PCIBridge
        PCI 44:00.2 (SATA)
  Die L#2
    NUMANode L#2 (P#1)
    L3 L#4 (8192KB)
      L2 L#12 (512KB) + L1d L#12 (32KB) + L1i L#12 (64KB) + Core L#12
        PU L#24 (P#12)
        PU L#25 (P#36)
      L2 L#13 (512KB) + L1d L#13 (32KB) + L1i L#13 (64KB) + Core L#13
        PU L#26 (P#13)
        PU L#27 (P#37)
      L2 L#14 (512KB) + L1d L#14 (32KB) + L1i L#14 (64KB) + Core L#14
        PU L#28 (P#14)
        PU L#29 (P#38)
    L3 L#5 (8192KB)
      L2 L#15 (512KB) + L1d L#15 (32KB) + L1i L#15 (64KB) + Core L#15
        PU L#30 (P#15)
        PU L#31 (P#39)
      L2 L#16 (512KB) + L1d L#16 (32KB) + L1i L#16 (64KB) + Core L#16
        PU L#32 (P#16)
        PU L#33 (P#40)
      L2 L#17 (512KB) + L1d L#17 (32KB) + L1i L#17 (64KB) + Core L#17
        PU L#34 (P#17)
        PU L#35 (P#41)
  Die L#3
    NUMANode L#3 (P#3)
    L3 L#6 (8192KB)
      L2 L#18 (512KB) + L1d L#18 (32KB) + L1i L#18 (64KB) + Core L#18
        PU L#36 (P#18)
        PU L#37 (P#42)
      L2 L#19 (512KB) + L1d L#19 (32KB) + L1i L#19 (64KB) + Core L#19
        PU L#38 (P#19)
        PU L#39 (P#43)
      L2 L#20 (512KB) + L1d L#20 (32KB) + L1i L#20 (64KB) + Core L#20
        PU L#40 (P#20)
        PU L#41 (P#44)
    L3 L#7 (8192KB)
      L2 L#21 (512KB) + L1d L#21 (32KB) + L1i L#21 (64KB) + Core L#21
        PU L#42 (P#21)
        PU L#43 (P#45)
      L2 L#22 (512KB) + L1d L#22 (32KB) + L1i L#22 (64KB) + Core L#22
        PU L#44 (P#22)
        PU L#45 (P#46)
      L2 L#23 (512KB) + L1d L#23 (32KB) + L1i L#23 (64KB) + Core L#23
        PU L#46 (P#23)
        PU L#47 (P#47)
```

There are 4 nodes, and some of them are empty. So resource allocation code finds the zero-mem node, tries to fallback and asserts that the number of nodes is not 1.

The fix is to fallback to sysconf only if all found nodes report zero memory.

### DIFF
--- a/src/core/resource.cc
+++ b/src/core/resource.cc
@@ -588,6 +588,17 @@ resources allocate(configuration& c) {
     if (procs == 0) {
         throw std::runtime_error("number of processing units must be positive");
     }
+
+    // Get the list of NUMA nodes available
+    std::vector<hwloc_obj_t> nodes;
+
+    hwloc_obj_t tmp = NULL;
+    auto num_nodes = hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_NUMANODE);
+    auto nodes_depth = hwloc_get_type_or_above_depth(topology, HWLOC_OBJ_NUMANODE);
+    while ((tmp = hwloc_get_next_obj_by_depth(topology, nodes_depth, tmp)) != NULL) {
+        nodes.push_back(tmp);
+    }
+
     auto machine_depth = hwloc_get_type_depth(topology, HWLOC_OBJ_MACHINE);
     SEASTAR_ASSERT(hwloc_get_nbobjs_by_depth(topology, machine_depth) == 1);
     auto machine = hwloc_get_obj_by_depth(topology, machine_depth, 0);
@@ -611,7 +622,6 @@ resources allocate(configuration& c) {
     std::vector<std::pair<cpu, size_t>> remains;
 
     auto cpu_sets = distribute_objects(topology, procs);
-    auto num_nodes = hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_NUMANODE);
 
     for (auto&& cs : cpu_sets()) {
         auto cpu_id = hwloc_bitmap_first(cs);
@@ -644,15 +654,6 @@ resources allocate(configuration& c) {
         }
 
         seastar_logger.warn("Assigning some CPUs to remote NUMA nodes");
-
-        // Get the list of NUMA nodes available
-        std::vector<hwloc_obj_t> nodes;
-
-        hwloc_obj_t tmp = NULL;
-        auto depth = hwloc_get_type_or_above_depth(topology, HWLOC_OBJ_NUMANODE);
-        while ((tmp = hwloc_get_next_obj_by_depth(topology, depth, tmp)) != NULL) {
-            nodes.push_back(tmp);
-        }
 
         // Group orphan CPUs by ... some sane enough feature
         std::unordered_map<hwloc_obj_t, std::vector<unsigned>> grouped;
@@ -697,7 +698,6 @@ resources allocate(configuration& c) {
     }
 
     // Divide the rest of the memory
-    auto depth = hwloc_get_type_or_above_depth(topology, HWLOC_OBJ_NUMANODE);
     for (auto&& [this_cpu, remain] : remains) {
         auto node = cpu_to_node.at(this_cpu.cpu_id);
         auto obj = node;
@@ -705,7 +705,7 @@ resources allocate(configuration& c) {
         while (remain) {
             remain -= alloc_from_node(this_cpu, obj, topo_used_mem, remain);
             do {
-                obj = hwloc_get_next_obj_by_depth(topology, depth, obj);
+                obj = hwloc_get_next_obj_by_depth(topology, nodes_depth, obj);
             } while (!obj);
             if (obj == node)
                 break;
--- a/src/core/resource.cc
+++ b/src/core/resource.cc
@@ -335,6 +335,7 @@ static size_t alloc_from_node(cpu& this_cpu, hwloc_obj_t node, std::unordered_ma
         auto node_id = hwloc_bitmap_first(node->nodeset);
         SEASTAR_ASSERT(node_id != -1);
         this_cpu.mem.push_back({taken, unsigned(node_id)});
+        seastar_logger.debug("CPU{} allocated {} bytes from NODE{}", this_cpu.cpu_id, taken, node_id);
     }
     return taken;
 }
--- a/src/core/resource.cc
+++ b/src/core/resource.cc
@@ -589,6 +589,8 @@ resources allocate(configuration& c) {
         throw std::runtime_error("number of processing units must be positive");
     }
 
+    size_t available_memory = 0;
+
     // Get the list of NUMA nodes available
     std::vector<hwloc_obj_t> nodes;
 
@@ -596,17 +598,22 @@ resources allocate(configuration& c) {
     auto num_nodes = hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_NUMANODE);
     auto nodes_depth = hwloc_get_type_or_above_depth(topology, HWLOC_OBJ_NUMANODE);
     while ((tmp = hwloc_get_next_obj_by_depth(topology, nodes_depth, tmp)) != NULL) {
+        available_memory += get_memory_from_hwloc_obj(tmp);
         nodes.push_back(tmp);
     }
+
+    if (!available_memory) {
 
     auto machine_depth = hwloc_get_type_depth(topology, HWLOC_OBJ_MACHINE);
     SEASTAR_ASSERT(hwloc_get_nbobjs_by_depth(topology, machine_depth) == 1);
     auto machine = hwloc_get_obj_by_depth(topology, machine_depth, 0);
-    auto available_memory = get_memory_from_hwloc_obj(machine);
+    available_memory = get_memory_from_hwloc_obj(machine);
     if (!available_memory) {
         available_memory = get_machine_memory_from_sysconf();
         set_memory_to_hwloc_obj(machine, available_memory);
         seastar_logger.warn("hwloc failed to detect machine-wide memory size, using memory size fetched from sysconf");
+    }
+
     }
 
     size_t mem = calculate_memory(c, std::min(available_memory,
--- a/src/core/resource.cc
+++ b/src/core/resource.cc
@@ -603,17 +603,15 @@ resources allocate(configuration& c) {
     }
 
     if (!available_memory) {
-
-    auto machine_depth = hwloc_get_type_depth(topology, HWLOC_OBJ_MACHINE);
-    SEASTAR_ASSERT(hwloc_get_nbobjs_by_depth(topology, machine_depth) == 1);
-    auto machine = hwloc_get_obj_by_depth(topology, machine_depth, 0);
-    available_memory = get_memory_from_hwloc_obj(machine);
-    if (!available_memory) {
-        available_memory = get_machine_memory_from_sysconf();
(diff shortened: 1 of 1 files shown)


xemul:
Ugh, on the other hand, other CPUs are assigned to nudes with mem, so just giving the orphan CPUs sysconf men is not that great idea

xemul:
upd:
- changed fallback condition

avikivity:
Why do we fallback? Shouldn't we allocate non-local memory? Or did we deliberately not try?

Maybe it's time to revisit.

xemul:
> Why do we fallback? 

We fallback when hwloc fails to tell node size (b0a9f89a70e20abea1305e17c083a6aad55b3189). AFAIU it happens one some nodes @syuu1228 worked with. And it just erroneously happens, that it happens on my box -- there are 4 nodes and 2 of them have 0 memory.

> Shouldn't we allocate non-local memory? 

On fallback, or on my node? For fallback we just do it, but assume that there's 1 node. On my node we should allocate non-local memory, but fallback asserts and prevents this. This PR fixes that and makes fallback not _that_ aggressive.

> Or did we deliberately not try?

Yes, it was a bug -- fallback code just aborted the app without trying.

Maybe it's time to revisit.

avikivity:
> > Why do we fallback?
> 
> We fallback when hwloc fails to tell node size ([b0a9f89](https://github.com/scylladb/seastar/commit/b0a9f89a70e20abea1305e17c083a6aad55b3189)). AFAIU it happens one some nodes @syuu1228 worked with. And it just erroneously happens, that it happens on my box -- there are 4 nodes and 2 of them have 0 memory.


Ahh. So it's a combination of an hwloc bug, and a bug in our handling.

> 
> > Shouldn't we allocate non-local memory?
> 
> On fallback, or on my node? For fallback we just do it, but assume that there's 1 node. On my node we should allocate non-local memory, but fallback asserts and prevents this. This PR fixes that and makes fallback not _that_ aggressive.

I see now.

> 
> > Or did we deliberately not try?
> 
> Yes, it was a bug -- fallback code just aborted the app without trying.
> 
> Maybe it's time to revisit.

Is that bug reported/fixed on hwloc?

xemul:
> I don't completely see. What hwloc reported was a valid configuration. It reports your machine has 126GB. So why fallback?

It reports 4 numa nodes, but 2 nodes have 64GB of RAM and 2 have zero. And fallback condtion is -- node->total_memory == 0. Just my case.

> Is that bug reported/fixed on hwloc?

I'm not sure it's a bug in hwloc. It's just the way ,my hardware reveals itself. This problem with memory-less numa node had been there since the day I got my workstation. I fixed it back in 2020 (f525f3e546bbc04902da55a57e33a021581ccc92). Now it strikes back again.

xemul:
@avikivity , here's the hwloc issue reported at 2020 about that hardware: https://github.com/open-mpi/hwloc/issues/430 .

avikivity:
> > I don't completely see. What hwloc reported was a valid configuration. It reports your machine has 126GB. So why fallback?
> 
> It reports 4 numa nodes, but 2 nodes have 64GB of RAM and 2 have zero. And fallback condtion is -- node->total_memory == 0. Just my case.

In that case, why fallback? hwloc is functioning, we should be able to distribute NUMA memory even if the distribution is bad.

> 
> > Is that bug reported/fixed on hwloc?
> 
> I'm not sure it's a bug in hwloc. It's just the way ,my hardware reveals itself. This problem with memory-less numa node had been there since the day I got my workstation. I fixed it back in 2020 ([f525f3e](https://github.com/scylladb/seastar/commit/f525f3e546bbc04902da55a57e33a021581ccc92)). Now it strikes back again.

I see.

xemul:
> In that case, why fallback? hwloc is functioning, we should be able to distribute NUMA memory even if the distribution is bad.

That's the essence of the bug this patch is fixing -- hwloc reports something to work with, but this place

https://github.com/scylladb/seastar/blob/0c7c0193917d924dd4cc1867b7e7430c4871463a/src/core/resource.cc#L624-L633

considers this as enough reason to fallback.

The fix is -- even if a NUMA node doesn't have memory, don't fallback, but put this CPU into "orphan" [1] list. Fallback ony after scanning all CPUs and making sure that _none_ of them could be attributed to any NUMA node.

[1] "orphan CPU" -- a cpu that wasn't assigned to any numa node on the first attempt, and it's then assigned to some "random" numa node later

xemul:
@avikivity , reminder ping. This issue doesn't affect a lot of people, but is still annoying

avikivity:
What if the node doesn't have zero memory, but just some small amount? I don't see zero memory as special.

I remember we did treat unequal distribution, maybe I'm misremembering.

avikivity:
Shouldn't the code under "// Divide the rest of the memory" be sufficient? Just drop the fallback.

avikivity:
> > What if the node doesn't have zero memory, but just some small amount? I don't see zero memory as special.
> > I remember we did treat unequal distribution, maybe I'm misremembering.
> 
> Presumably, your talking about [cbc2f40](https://github.com/scylladb/seastar/commit/cbc2f4068088f134bcf8be97ea98656726a0ba29). 

Yes.

> The algo tries to grab memory from local nodes, if it's available there, then borrows the remainder from other ~groups~ nodes.


Is this not exactly the case we're trying to solve here? 

> 
> > Shouldn't the code under "// Divide the rest of the memory" be sufficient? Just drop the fallback.
> 
> It doesn't seem sufficient. The memory divider needs some memory to distribute. However, if _all_ nodes report to have zero memory, there's nothing to get memory from and something should be done about it. That's what fallback is about. But it has a bug -- once it finds a _single_ node with zero memory, it turns on. This PR fixes it -- one it sees that _all_ nodes have no memory, it turns on.

But you do have nodes with memory. 

I think I see the problem. You're using some --cpuset or --smp that selects a certain subset of PUs, and none of those PUs belong to a node with memory.

If that's the case, then I think we should select all NUMA nodes regardless of whether they host selected PUs. Then run the two-phase algorithm, first allocate from the local NUMA node, then allocate from remote nodes.

src/core/resource.cc
```diff
@@ -619,28 +619,44 @@ resources allocate(configuration& c) {
...
 
     auto cpu_sets = distribute_objects(topology, procs);
-    auto num_nodes = hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_NUMANODE);
 
     for (auto&& cs : cpu_sets()) {
         auto [cpu_id, node] = get_cpu_id_and_numa_node_for_set(topology, cs);
-        if (node == nullptr) {
+        if (node == nullptr || !get_memory_from_hwloc_obj(node)) {
             orphan_pus.push_back(cpu_id);
         } else {
```
avikivity, line 627: I don't understand this special case. What's the difference between a node that has a small amount of memory and no memory? Should work exactly the same.

Maybe you're fast-promoting it to only participate in the second round of distribution.

avikivity: So what if you run with --smp 1 and that happens to be on a memory-less NUMA node?

It will trigger the fallback even though hwloc works normally.

I think the algorithm would work if we referenced all numa nodes, not just those in cpu_to_node.

```
collection\<numa_node> numas;
map\<cpu, numa_node> assignment;

for (all cpus) {
    try to assign from assignment[cpu]
}

for (all cpus) {
     delta = whatever we failed to assign in phase 1
     assign delta from whatever node as free memory
}
```

xemul:
upd:
Keep nodes with zero reported memory as is, the "borrowing" code handles it. Instead, fallback to sysconf only when all nodes report zero memory.

This change also implied one more fix -- when calculating the available memory for the node, get it from numa-node hwloc objects, not from machine one. Although it's natural to expect self-consistent values from hwloc, memory dividing code operates on numa-node objects to distribute memory between cpus, so it's "safer" to get the total available amount of memory from the same place.

PR #2748, opened 2025-04-18
doc/io-properties-file.md: correct a typo

tchaikov:
Replace incorrect "LLVM" with "LVM" (Logical Volume Manager). This avoids confusion with the unrelated LLVM compiler project.


xemul:
Thanks for fixing it!

PR #2752, opened 2025-04-23
reactor: Keep task-queues in std::array instead of static_vector

xemul:
There's no code in reactor that ever needs to know how many elements there are in this vector, so there's no point in keeping it as such. Plain array would work the same.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -62,7 +62,6 @@
 #include "internal/pollable_fd.hh"
 
 #ifndef SEASTAR_MODULE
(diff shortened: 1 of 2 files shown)


tgrabiec:
What's the motivation, getting rid of boost?

xemul:
> What's the motivation, getting rid of boost?

Yes. From the popular header.

And also "the least astonishment" principle -- every time I come across this field I wonder why it is variable-size container, while in reality it's not

avikivity:
reactor.hh should not be a popular header!

PR #2753, opened 2025-04-24
memory: Report oversized alloc count as metric

StephanDollberg:
Oversized allocs are already tracked in the stats object.

Export the count in the memory metrics as well.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2569,7 +2569,8 @@ void reactor::register_metrics() {
             sm::make_current_bytes("total_memory", [] { return memory::stats().total_memory(); }, sm::description("Total memory size in bytes")),
             sm::make_current_bytes("allocated_memory", [] { return memory::stats().allocated_memory(); }, sm::description("Allocated memory size in bytes")),
             sm::make_counter("reclaims_operations", [] { return memory::stats().reclaims(); }, sm::description("Total reclaims operations")),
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -2569,7 +2569,8 @@ void reactor::register_metrics() {
             sm::make_current_bytes("total_memory", [] { return memory::stats().total_memory(); }, sm::description("Total memory size in bytes")),
             sm::make_current_bytes("allocated_memory", [] { return memory::stats().allocated_memory(); }, sm::description("Allocated memory size in bytes")),
             sm::make_counter("reclaims_operations", [] { return memory::stats().reclaims(); }, sm::description("Total reclaims operations")),
-            sm::make_counter("malloc_failed", [] { return memory::stats().failed_allocations(); }, sm::description("Total count of failed memory allocations"))
+            sm::make_counter("malloc_failed", [] { return memory::stats().failed_allocations(); }, sm::description("Total count of failed memory allocations")),
+            sm::make_counter("oversized_allocs", [] { return memory::stats().large_allocations(); }, sm::description("Total count of oversized memory allocations"))
```
xemul, line 2573: A side note: The memory::stats() API is quite terrible in a sense that it copies around ~100 bytes just get 8

mykaul: @amnonh - is that worth looking into?

avikivity: It's not that bad, copying 128 bytes is 8 instructions with AVX.

Moreover, the ABI is to pass the address of the return type as a hidden parameter, so typically it won't be copied.

PR #2757, opened 2025-04-25
reactor: Simplify network stack creation and initialization

xemul:
When reactor starts it spawns two fibers on each shard.

First waits for the local semaphore to acquire smp::count units, then calls network_stack->initialize(), then wakes up start_promise.

Second waits for network_stack_ready promise to resolve, moves the promised stack unique_ptr onto reactor, then gives one unit to the aforementioned semaphore on all cpus.

All those fibers run in the background.

The same goal can be achieved without semaphore and with less cross-shards invocations.

Spawn an invoke_on_all fiber that waits for network_stack_ready promises on all shards. It will resolve after all shards' promise is resolved.

Then() invoke_on_all again, call initialize() on each network_stack, then resolve the start_promise.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1042,7 +1042,6 @@ reactor::reactor(std::shared_ptr<seastar::smp> smp, alien::instance& alien, unsi
     , _notify_eventfd(file_desc::eventfd(0, EFD_CLOEXEC))
     , _task_quota_timer(file_desc::timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC))
     , _id(id)
-    , _cpu_started(0)
     , _cpu_stall_detector(internal::make_cpu_stall_detector())
     , _reuseport(posix_reuseport_detect())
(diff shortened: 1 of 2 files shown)


src/core/reactor.cc
```diff
@@ -3202,20 +3201,22 @@ int reactor::do_run() {
...
+        // Wait for network stack to appear on all cpus.
+        // Communicate when done using _start_promise
+        (void)smp::invoke_on_all([] {
+            return engine()._network_stack_ready->then([] (std::unique_ptr<network_stack> stack) {
+                engine()._network_stack = std::move(stack);
+            });
+        }).then([] {
+            return smp::invoke_on_all([] {
+                return engine()._network_stack->initialize().then([] {
+                    engine()._start_promise.set_value();
```
avikivity, line 3215: This gives me a headache. We should move all the async stuff into app_template where it can be managed with conventional code. But no need to do it now.

Issue #2759, opened 2025-04-28
http::content_length_source_impl may cause problems


xemul:
In case user receives content-length response from server, but doesn't read the whole body, client then puts the connection back into pool and next request will erroneously read the previous' one remaining body.

xemul:
I agree about chunked-encoding bodies, but not about keep-alive connections. If server closed the conneciton, http::client will notice it and will withdraw it from pool after seeing EOF

PR #2762, opened 2025-04-29
http: Track bytes read from a response

kreuzerkrieg:
If the server response includes a "Content-Length" header, keep track of the bytes read from the stream. When reading stops, the connection may be returned to the pool with leftover bytes, which could be read by the next connection user. To prevent this, reset the connection before returning it to the pool.

fixes: https://github.com/scylladb/seastar/issues/2759

### DIFF
--- a/include/seastar/http/internal/content_source.hh
+++ b/include/seastar/http/internal/content_source.hh
@@ -41,9 +41,14 @@ namespace internal {
 class content_length_source_impl : public data_source_impl {
     input_stream<char>& _inp;
     size_t _remaining_bytes = 0;
+    size_t* _consumed_bytes;
 public:
+    content_length_source_impl(input_stream<char>& inp, size_t length, size_t* consumed_bytes)
+        : _inp(inp), _remaining_bytes(length), _consumed_bytes(consumed_bytes) {
+    }
+
     content_length_source_impl(input_stream<char>& inp, size_t length)
-        : _inp(inp), _remaining_bytes(length) {
+        : content_length_source_impl(inp, length, nullptr) {
     }
 
     virtual future<temporary_buffer<char>> get() override {
@@ -52,6 +57,9 @@ public:
         }
         return _inp.read_up_to(_remaining_bytes).then([this] (temporary_buffer<char> tmp_buf) {
             _remaining_bytes -= tmp_buf.size();
+            if (_consumed_bytes) {
+                *_consumed_bytes += tmp_buf.size();
+            }
             return tmp_buf;
         });
     }
@@ -59,6 +67,9 @@ public:
     virtual future<temporary_buffer<char>> skip(uint64_t n) override {
         uint64_t skip_bytes = std::min(n, _remaining_bytes);
         _remaining_bytes -= skip_bytes;
+        if (_consumed_bytes) {
+            *_consumed_bytes += skip_bytes;
+        }
         return _inp.skip(skip_bytes).then([] {
             return temporary_buffer<char>();
         });
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -186,7 +186,7 @@ input_stream<char> connection::in(reply& rep) {
         return input_stream<char>(data_source(std::make_unique<httpd::internal::chunked_source_impl>(_read_buf, rep.chunk_extensions, rep.trailing_headers)));
     }
 
-    return input_stream<char>(data_source(std::make_unique<httpd::internal::content_length_source_impl>(_read_buf, rep.content_length)));
+    return input_stream<char>(data_source(std::make_unique<httpd::internal::content_length_source_impl>(_read_buf, rep.content_length, &rep.consumed_content)));
 }
 
 void connection::shutdown() noexcept {
@@ -216,9 +216,10 @@ client::client(socket_address addr, shared_ptr<tls::certificate_credentials> cre
 {
 }
 
-client::client(std::unique_ptr<connection_factory> f, unsigned max_connections, retry_requests retry)
+client::client(std::unique_ptr<connection_factory> f, unsigned max_connections, retry_requests retry, size_t max_bytes_to_drain)
         : _new_connections(std::move(f))
         , _max_connections(max_connections)
+        , _max_bytes_to_drain(max_bytes_to_drain)
         , _retry(retry)
 {
 }
@@ -375,7 +376,7 @@ future<> client::make_request(request& req, reply_handler& handle, std::optional
 
 future<> client::do_make_request(connection& con, request& req, reply_handler& handle, abort_source* as, std::optional<reply::status_type> expected) {
     auto sub = as ? as->subscribe([&con] () noexcept { con.shutdown(); }) : std::nullopt;
-    return con.do_make_request(req).then([&con, &handle, expected] (connection::reply_ptr reply) mutable {
+    return con.do_make_request(req).then([this, &con, &handle, expected] (connection::reply_ptr reply) mutable {
         auto& rep = *reply;
         if (expected.has_value() && rep._status != expected.value()) {
             if (!http_log.is_enabled(log_level::debug)) {
@@ -390,7 +391,20 @@ future<> client::do_make_request(connection& con, request& req, reply_handler& h
             });
         }
 
-        return handle(rep, con.in(rep)).finally([reply = std::move(reply)] {});
+        return handle(rep, con.in(rep)).then([this, reply = std::move(reply), &con] {
+            if (reply->content_length > reply->consumed_content) {
+                auto bytes_left = reply->content_length - reply->consumed_content;
+                if (bytes_left <= _max_bytes_to_drain) {
+                    http_log.trace("content was not fully consumed, {} bytes were left behind, skipping and returning the connection to the pool", bytes_left);
+                    return con._read_buf.skip(bytes_left);
+                }
+                http_log.trace("content was not fully consumed, content length is {} but consumed only {}, will close the connection",
+                               reply->content_length,
+                               reply->consumed_content);
+                con._persistent = false;
+            }
+            return make_ready_future<>();
+        });
     }).handle_exception([&con] (auto ex) mutable {
         con._persistent = false;
         return make_exception_future<>(std::move(ex));
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -13,6 +13,7 @@
 #include <seastar/core/do_with.hh>
 #include <seastar/core/loop.hh>
 #include <seastar/core/when_all.hh>
+#include <seastar/core/units.hh>
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
 #include "loopback_socket.hh"
@@ -2001,10 +2002,10 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
 SEASTAR_TEST_CASE(test_client_close_connection) {
     return async([] {
         loopback_connection_factory lcf(1);
-        auto cln = http::experimental::client(std::make_unique<loopback_http_factory>(lcf), 1, http::experimental::client::retry_requests::no);
-        auto make_test_request = [&cln]() {
+        auto make_test_request = [&lcf]() {
+            auto cln = http::experimental::client(std::make_unique<loopback_http_factory>(lcf), 1, http::experimental::client::retry_requests::no);
             size_t content_length = 0;
-            for (auto cycle : {1, 2}) {
+            for (auto _ [[maybe_unused]] : {1, 2}) {
                 auto req = http::request::make("GET", "test", "/test");
                 auto make_request = cln.make_request(
                     std::move(req),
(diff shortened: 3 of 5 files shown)


include/seastar/http/internal/content_source.hh
```diff
@@ -40,10 +40,15 @@ namespace internal {
...
+    http::reply* _rep;
     size_t _remaining_bytes = 0;
 public:
+    content_length_source_impl(input_stream<char>& inp, http::reply* rep)
+        : _inp(inp), _rep(rep), _remaining_bytes(_rep->content_length) {
+    }
+
     content_length_source_impl(input_stream<char>& inp, size_t length)
-        : _inp(inp), _remaining_bytes(length) {
+        : _inp(inp), _rep(nullptr), _remaining_bytes(length) {
```
xemul, line 51: Is this constructor still in use?

include/seastar/http/internal/content_source.hh
```diff
@@ -40,10 +40,15 @@ namespace internal {
  * */
 class content_length_source_impl : public data_source_impl {
     input_stream<char>& _inp;
+    http::reply* _rep;
```
xemul, line 43: Let's make it symmetrical to `http_content_length_data_sink_impl` which doesn't reference the request itself, but directly the bytes-written counter on it

src/http/client.cc
```diff
@@ -390,7 +390,14 @@ future<> client::do_make_request(connection& con, request& req, reply_handler& h
             });
         }
 
-        return handle(rep, con.in(rep)).finally([reply = std::move(reply)] {});
+        return handle(rep, con.in(rep)).finally([reply = std::move(reply), &con] {
```
xemul, line 393: Need a test that fails before this patch and passes after it

src/http/client.cc
```diff
@@ -390,7 +390,14 @@ future<> client::do_make_request(connection& con, request& req, reply_handler& h
...
+            if (reply->content_length != reply->consumed_content) {
+                http_log.trace("content was not fully consumed, content length is {} but consumed only {}, will close the connection",
+                               reply->content_length,
+                               reply->consumed_content);
+                con._persistent = false;
```
xemul, line 398: nitpicking: This marking and log message is not really needed in case we get here by exception

src/http/client.cc
```diff
@@ -390,7 +390,14 @@ future<> client::do_make_request(connection& con, request& req, reply_handler& h
             });
         }
 
-        return handle(rep, con.in(rep)).finally([reply = std::move(reply)] {});
+        return handle(rep, con.in(rep)).finally([reply = std::move(reply), &con] {
+            if (reply->content_length != reply->consumed_content) {
+                http_log.trace("content was not fully consumed, content length is {} but consumed only {}, will close the connection",
```
xemul, line 395: As a followup.

Dropping a connection and establishing a new one can be more costly than reading the remaining bytes. Need some (AIish) logic here that prefers consuming the remaining bytes and dropping them on the floor instead of dropping the whole connection

include/seastar/http/internal/content_source.hh
kreuzerkrieg: yes, in `make_content_stream`, it has its own way to take care of unconsumed bytes in `connection::read_one`

src/http/client.cc
kreuzerkrieg: AI-ish, sure :) 128KiB it is

include/seastar/http/internal/content_source.hh
```diff
@@ -52,13 +57,19 @@ public:
         }
         return _inp.read_up_to(_remaining_bytes).then([this] (temporary_buffer<char> tmp_buf) {
             _remaining_bytes -= tmp_buf.size();
+            if (_rep) {
+                _rep->consumed_content += tmp_buf.size();
+            }
```
avikivity, line 62: Is it possible for the reply to get the data_source, static_cast it into content_length_source_impl, and check the result? Would be a little more efficient since it's done one time.

include/seastar/http/internal/content_source.hh
```diff
@@ -40,10 +40,15 @@ namespace internal {
  * */
 class content_length_source_impl : public data_source_impl {
     input_stream<char>& _inp;
+    http::reply* _rep;
     size_t _remaining_bytes = 0;
```
avikivity, line 44: Note: size_t is wrong here in case of a hypothetical 32-bit port reading >4GB. Not the fault of this patch of course so needn't be fixed here.

src/http/client.cc
kreuzerkrieg: how about now? moved the code from `finally` to `then` and then the exception, I guess, will propagate to outer `handle_exception`

src/http/client.cc
kreuzerkrieg: working on it

include/seastar/http/internal/content_source.hh
avikivity: I would prefer to avoid the `if`. Can this data_source exist after the request is gone? If not we can make it unconditional.

If we can't, then ok.

kreuzerkrieg: ah, ok, I see. Let me see what can be done there

tests/unit/httpd_test.cc
```diff
@@ -1997,3 +1997,64 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
...
+                output_stream<char> out = sk.output();
+                sstring r200(format("HTTP/1.1 200 OK\r\nHost: localhost\r\nContent-Length: {}\r\n\r\n", response_size));
+                out.write(r200).get();
+                out.flush().get();
+                out.write(sstring(response_size / 2, 'a')).get();
+                out.flush().get();
+                cv.wait().get();
+                if (response_size / 2 > 128 * 1024) {
+                    ++throw_count;
+                    BOOST_REQUIRE_EXCEPTION(
```
xemul, line 2037: Why will it throw?

include/seastar/http/internal/content_source.hh
xemul: @avikivity , this if shouldn't be there. The httpd (server) request handler also uses the same content-lengh-data-source and it can track the number of unread bytes by the handler as well. Nowadays the server tries to consume the remaining body after the handler is done instead, which is not nice.

src/http/client.cc
```diff
@@ -390,7 +390,20 @@ future<> client::do_make_request(connection& con, request& req, reply_handler& h
             });
         }
 
-        return handle(rep, con.in(rep)).finally([reply = std::move(reply)] {});
+        return handle(rep, con.in(rep)).then([reply = std::move(reply), &con] {
+            if (reply->content_length > reply->consumed_content) {
+                auto bytes_left = reply->content_length - reply->consumed_content;
+                if (bytes_left <= 128 * 1024) {
```
xemul, line 396: This should be at least configurable client variable. not a magic constant in the middle of the code

tests/unit/httpd_test.cc
kreuzerkrieg: coz the socket is already closed when we decide on the client side not to drain the socket and just close the connection

tests/unit/httpd_test.cc
```diff
@@ -1997,3 +1997,73 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
...
+                } else {
+                    ++nothrow_count;
+                    BOOST_REQUIRE_NO_THROW(out.write(sstring(response_size / 2, 'b')).get());
+                }
+                out.flush().get();
+                out.close().get();
+            });
+        };
+
+        for (auto size : {_128KiB, 260 * 1024ul}) {
```
xemul, line 2057: Why not `_260KiB`?

tests/unit/httpd_test.cc
```diff
@@ -1997,3 +1997,73 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
...
+            auto req = http::request::make("GET", "test", "/test");
+            cln.make_request(
+                   std::move(req),
+                   [&cv, &content_length](const http::reply& resp, input_stream<char>&& in) {
+                       content_length = resp.content_length;
+                       return seastar::async([&cv, &content_length, in = std::move(in)]() mutable {
+                           auto buff = in.read().get();
+                           if (content_length <= _128KiB) {
+                               cv.signal();
+                           }
```
xemul, line 2018: `in.close()` is missing?

tests/unit/httpd_test.cc
```diff
@@ -1997,3 +1997,73 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
...
+        auto cln = http::experimental::client(std::make_unique<loopback_http_factory>(lcf), 1, http::experimental::client::retry_requests::yes);
+        auto make_test_request = [&cv, &cln]() {
+            size_t content_length = 0;
+            auto req = http::request::make("GET", "test", "/test");
+            cln.make_request(
+                   std::move(req),
+                   [&cv, &content_length](const http::reply& resp, input_stream<char>&& in) {
+                       content_length = resp.content_length;
+                       return seastar::async([&cv, &content_length, in = std::move(in)]() mutable {
+                           auto buff = in.read().get();
```
xemul, line 2015: Worth having a check that the amt of bytes read is less than the claimed content length

tests/unit/httpd_test.cc
```diff
@@ -1997,3 +1997,73 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
...
+                           if (content_length <= _128KiB) {
+                               cv.signal();
+                           }
+                       });
+                   },
+                   http::reply::status_type::ok)
+                .then([&content_length, &cv] {
+                    if (content_length > _128KiB) {
```
xemul, line 2023: Why cv.signal() can't go unconditionally after cln.make_request() resolves?

kreuzerkrieg: because of the draining, it has to keep writing to the output so, the client can `skip()`

tests/unit/httpd_test.cc
```diff
@@ -1997,3 +1997,73 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
...
+            return async([&cv, response_size, &throw_count, &nothrow_count, sk = std::move(ar.connection)]() mutable {
+                input_stream<char> in = sk.input();
+                read_simple_http_request(in);
+                output_stream<char> out = sk.output();
+                sstring r200(format("HTTP/1.1 200 OK\r\nHost: localhost\r\nContent-Length: {}\r\n\r\n", response_size));
+                out.write(r200).get();
+                out.flush().get();
+                out.write(sstring(response_size / 2, 'a')).get();
+                out.flush().get();
+                cv.wait().get();
```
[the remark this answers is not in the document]
kreuzerkrieg: As I understand you can never miss `signal` since it just sets a flag that stays "on" until someone consumes it

tests/unit/httpd_test.cc
kreuzerkrieg: why to bother declaring `_260KiB` when it used only once? what we would achieve by it? nor readability neither maintainability

include/seastar/http/internal/content_source.hh
nyh: "Nowadays the server tries to consume the remaining body after the handler is done instead" - this isn't true. We discussed this heavily (for the server case), but have not done this.

nyh:
I have not reviewed the code, but just want to make two general comments:
1. As I mentioned in https://github.com/scylladb/seastar/issues/2759 I think it is a mistake to think that the problem is specific to Content-Length, or that it can be solved with some tracking specific to Content-Length.
2. In the server code, to know if we read the entire request, we "just" read from it - if the read is successful, it means we didn't read the entire request and the connection needs to be dropped. I quoted "just" because the read doesn't happen directly to the socket but rather to a wrapper which understands content-length or chunked-encoding, and this wrapper stops reading when it reaches the end of the content-length or the last chunk, so if this wrapper can read anything, it means we haven't read everything.

include/seastar/http/internal/content_source.hh
xemul: I meant this:
https://github.com/scylladb/seastar/blob/3699c0ce7ea98dc170e6d05148d900b5a649d882/src/http/httpd.cc#L277-L287
server gets the next portion of data from the budy and if succeeds, then marks the connection for closing.

I guess I used too strong "consume" verb to describe this :(

nyh: Yes, right. We do not "consume" the rest of the request (which we could have done but didn't) just try to "read" anything more. Note that as I commented above, content_stream() is not the raw socket, it is actually a wrapper which limits itself to the Content-Length or chunked encoding (if used), so this read will return empty() if we're at the real end of the request - we don't hang reading from the socket after the end of the request.

tests/unit/httpd_test.cc
```diff
@@ -2033,31 +2030,38 @@ SEASTAR_TEST_CASE(test_client_close_connection) {
...
-
-                    out.write(r200).get();
-                    out.flush().get();
-                    out.write(sstring(response_size / 2, 'a')).get();
-                    out.flush().get();
+                size_t responses = 0;
+                // In the case the leftover data on the socket is smaller than 128KiB we are going to drain it and leave the connection alive, so here we
+                // have to loop two times to fulfill two request from the client. On the other hand if the leftover data is larger than 128KiB we are going
+                // to close the connection, so we have to loop only once to fulfill one request from the client and make another `accept`
+                while ((response_size <= 128_KiB && responses < 2) || (response_size > 128_KiB && responses < 1)) {
```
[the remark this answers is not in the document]
kreuzerkrieg: Coz in the `response_size <= 128_KiB` I need _**two**_ full responses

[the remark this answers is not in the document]
kreuzerkrieg: dont forget the response is always "full" since the data is being drained

PR #2767, opened 2025-05-14
http: Add test for improper client handling of aborted requests

kreuzerkrieg:
Introduce a test to demonstrate the client's incorrect behavior when a socket is drained or closed before all request data is read.

This test serves as a showcase of the issue and does not include a fix, which will be addressed in a subsequent PR.

### DIFF
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -1997,3 +1997,67 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
 
     c.close().get();
 }
+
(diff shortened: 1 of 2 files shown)


xemul:
I like the test, but the description should include:

1. why the parser fails parsing the 2nd response (link to the issue can work)
2. why we increase the buffer size for loopback socket

kreuzerkrieg:
> I like the test, but the description should include:
> 
> 1. why the parser fails parsing the 2nd response (link to the issue can work)
> 2. why we increase the buffer size for loopback socket

How about now?

PR #2769, opened 2025-05-16
http: in the client, fix malforming of requests with zero-sized bodies

michoecho:
`connection::write_body` assumes that `req.content_length == 0` implies that chunked encoding has been selected.

Indeed, one variant of `request::write_body` sets
`_headers["Transfer-Encoding"] = "chunked"` and doesn't set `.content_length`, while the other variant
doesn't set `Transfer-Encoding`, but sets `.content_length`.

So the following is true: if `.content_length` wasn't set, then chunked encoding was selected.
That's what `connection::write_body` is trying to depend on.

The problem is that `req.content_length == 0` does not mean that `req.content_length` wasn't set. It could have just been set to 0.

If that happens, the client sends a request with no `Content-Length`, no `Transfer-Encoding`, and a body equal to "0\r\n\r\n", and that's a malformed request.

We fix this by making `request::write_body` keep the `request::body_writer` unset if the declared content length is 0.

### DIFF
--- a/src/http/request.cc
+++ b/src/http/request.cc
@@ -113,7 +113,11 @@ void request::write_body(const sstring& content_type, noncopyable_function<futur
 void request::write_body(const sstring& content_type, size_t len, noncopyable_function<future<>(output_stream<char>&&)>&& body_writer) {
     set_content_type(content_type);
     content_length = len;
-    this->body_writer = std::move(body_writer);
+    if (len > 0) {
+        // At the time of this writing, connection::write_body()
+        // assumes that `body_writer` is unset if `content_length` is 0.
+        this->body_writer = std::move(body_writer);
+    }
 }
 
 void request::set_expects_continue() {


michoecho:
Didn't test it, just checked that it fixes my problem.

This change is backward incompatible (if a user of Seastar accesses `request::content_length`), but I don't see a way to avoid that.

Context: I found this when I made a change to Scylla which adds a possibly empty SSTable component, and I got failures in a test which uploads a SSTable backup via HTTP. Turns out that uploading an empty file was breaking the HTTP client.

xemul:
But `client::write_body()` is more complicated than what you have described. It first checks if `req.body_writer` is set or not, and only then checks `req.content_length` deciding on whether content-length-body or chunked-body to use. So if one doesn't have a body to write, then not calling request::write_body() is the way to go.

michoecho:
> > are you saying that you like request::write_body() being broken in the edge case?
> 
> No, if I were I would tell that.
> 
> > Would you rather fix request::write_body(), or document a new precondition that len == 0 and then go and patch all request::write_body() callsites in Scylla to do if (len != 0)?
> 
> The former. The `req.content` is public member of request, turning it into optional without [[depracated]]-ing can happen to be breaking change. I should've been explicit from the very beginning, my bad.

Oh, okay. You said, "if one doesn't have a body to write, then not calling request::write_body()", so I was incredulous because `request::write_body()` is a user-facing interface, so it would be very error-prone if we relied on the caller, since it's not obvious that there's something special about size 0. But after the clarification I understand that you mean "if `write_body` is called with body length 0, then it should just be a no-op"?)

Should I change the patch to do that instead?

michoecho:
> Should I change the patch to do that instead?

v2: updated the patch. @xemul Please re-review.

PR #2779, opened 2025-05-27
httpd: Don't let body into the HEAD replies output

xemul:
RFC 7231 Section 4.3.2

>  The HEAD method is identical to GET except that the server MUST NOT
>  send a message body in the response (i.e., the response terminates at
>  the end of the header section).

Seastar HTTPD doesn't respect that `MUST NOT` thing and allows body content in case handler requested for it.

Fix by explicitly skipping the body regardless of what handler put on it. Since replies are written back to connection in a separate fiber where the initial request is no longer available, add the "skip body" marking on the reply itself.

### DIFF
--- a/src/http/reply.cc
+++ b/src/http/reply.cc
@@ -131,7 +131,9 @@ future<> reply::write_reply_to_connection(httpd::connection& con) {
     }).then([&con] () mutable {
         return con.out().write("\r\n", 2);
     }).then([this, &con] () mutable {
-        return _body_writer(http::internal::make_http_chunked_output_stream(con.out()));
+        return _body_writer(http::internal::make_http_chunked_output_stream(con.out())).then([&con] {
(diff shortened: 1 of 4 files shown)


src/http/reply.cc
```diff
@@ -131,7 +131,9 @@ future<> reply::write_reply_to_connection(httpd::connection& con) {
     }).then([&con] () mutable {
         return con.out().write("\r\n", 2);
     }).then([this, &con] () mutable {
-        return _body_writer(http::internal::make_http_chunked_output_stream(con.out()));
+        return _body_writer(http::internal::make_http_chunked_output_stream(con.out())).then([&con] {
+            return con.out().write("0\r\n\r\n", 5);
+        });
```
avikivity, line 139: This can add an allocation if we're filling the buffer faster than tcp can drain it. But in such case, the extra allocation is amortized over the huge data size.

Issue #2768, opened 2025-05-15
http::chunked_source_impl may cause problems


xemul:
Similarly to #2759, partially read chunked-encoded bodies cause subsequent request read the body continuation instead of new response.

nyh:
As I commented in https://github.com/scylladb/seastar/issues/2759, I believe the same problem can also occur for responses that use neither chunked encoding nor content-length - i.e.,  responses where keep-alive is disabled, and the server plans to write the response and then immediately close the connection:

Normally, when the client reads this response until the very end, it notices the EOF and closes the connection - and for the next request, will use a new request. But if the client does not read the response until the EOF, the client may incorrectly reuse this connection (with a partially unread response and a server who will never read more requests from this connection). If the client doesn't read until the end the library code may be under the delusion that the connection is still kept alive and may be reused.

The scenario I'm describing was very common in HTTP 1.0 (which, I admit, isn't exactly common in modern Web servers), but also legal in HTTP 1.1. In HTTP 1.1, the server will usually also add a header `Connection: close` to the response to _warn_ the client it will get an EOF, but this "advance warning" is not really necessary and clients need to work also without it (such an advance warning did not even exist in the original HTTP 0.9).

PR #2780, opened 2025-05-27
Extend the http test for partially read reply body

xemul:
There's a test that highlights an issue in the client's behavior when a socket is closed before all request data is fully read. When this happens, the subsequent request shouldn't receive leftover data from the previous response and fail. For content-length replies this case is handled by client and the test passes. For chunked-encoded bodies, the problem is yet to be fixed.

This test extension is intended solely to demonstrate the problem and does not include a fix. The resolution will be addressed in a follow-up PR.

ref #2767
ref #2768

### DIFF
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -2000,17 +2000,17 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
 }
 
 future<> test_client_close_connection(bool chunked) {
-    return async([] {
+    return async([chunked] {
         loopback_connection_factory lcf(1);
-        auto make_test_request = [&lcf]() {
+        auto make_test_request = [&lcf, chunked]() {
             auto cln = http::experimental::client(std::make_unique<loopback_http_factory>(lcf), 1, http::experimental::client::retry_requests::no);
             size_t content_length = 0;
             for (auto _ [[maybe_unused]] : {1, 2}) {
                 auto req = http::request::make("GET", "test", "/test");
                 auto make_request = cln.make_request(
                     std::move(req),
-                    [&content_length](const http::reply& resp, input_stream<char>&& in) {
-                        content_length = resp.content_length;
+                    [&content_length, chunked](const http::reply& resp, input_stream<char>&& in) {
+                        content_length = chunked ? 128_KiB : resp.content_length;
                         return async([&content_length, in = std::move(in)]() mutable {
                             // just read some bytes and abandon
                             auto buff = in.read().get();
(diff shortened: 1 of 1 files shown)


xemul:
upd:
- ignore server-side exception on socket closing

tests/unit/httpd_test.cc
```diff
@@ -2051,7 +2051,7 @@ future<> test_client_close_connection(bool chunked) {
                         break;
                     }
                 }
-                out.close().get();
+                out.close().handle_exception([](auto ex){}).get();
```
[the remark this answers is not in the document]
xemul: It may happen that server "survives" up until the very last flush. Then, since this last flush schedules a batch flush, the exception is set and delivered only on close()

kreuzerkrieg (approved):

tests/unit/httpd_test.cc
```diff
@@ -2019,14 +2019,20 @@ future<> test_client_close_connection(bool chunked) {
                         });
                     },
                     http::reply::status_type::ok);
+              if (!chunked || (_ == 1)) {
                 BOOST_REQUIRE_NO_THROW(make_request.get());
+              } else {
+                // FIXME chunked encoded bodies are not yet checked by client to be
+                // under-read, so 2nd request will receive the tail of the 1st response
+                BOOST_REQUIRE_THROW(make_request.get(), httpd::response_parsing_exception);
```
nyh, line 2027: I don't like to have tests which enshrine **wrong** behavior. I'd much rather have a test that should pass, and for now we know fails (i.e., "xfail" in pytest lingo). I don't know how to do this in unit tests.
But it still doesn't mean we can have a test which insists that a known bug can't be fixed...

tests/unit/httpd_test.cc
```diff
@@ -2000,17 +2000,17 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
...
             auto cln = http::experimental::client(std::make_unique<loopback_http_factory>(lcf), 1, http::experimental::client::retry_requests::no);
             size_t content_length = 0;
             for (auto _ [[maybe_unused]] : {1, 2}) {
                 auto req = http::request::make("GET", "test", "/test");
                 auto make_request = cln.make_request(
                     std::move(req),
-                    [&content_length](const http::reply& resp, input_stream<char>&& in) {
-                        content_length = resp.content_length;
+                    [&content_length, chunked](const http::reply& resp, input_stream<char>&& in) {
+                        content_length = chunked ? 128_KiB : resp.content_length;
```
nyh, line 2013: Where does this 128KB come from? Response chunks have a length written by the server - the client shouldn't guess them. I see below that your server indeed returns the entire response as one chunk - and writes the chunk size in the beginning of the chunk.

nyh (changes requested):
It's great that you are writing a test to rerproduce a bug (https://github.com/scylladb/seastar/issues/2768), but I don't think it's a good idea to have tests that enshrine **incorrect** behavior. I think the test should check the correct behavior, and fail. Ideally we should be able able to mark this test "xfail" ("expected failure") as we can do in pytest - but I don't know how. Another ugly option is just to `#if 0` the test out :-(

xemul:
Closing in favor of #3063

PR #2787, opened 2025-05-30
tls: add ALPN support

crackcomm:
This PR introduces client-side ALPN (Application-Layer Protocol Negotiation) to Seastar's TLS.

Clients can now specify a list of preferred application protocols (e.g., "h2", "http/1.1") in `tls::tls_options::alpn_protocols`. After the TLS handshake, the negotiated protocol can be retrieved using `tls::get_selected_alpn_protocol()`.

This is primarily motivated by the need to negotiate HTTP/2.

### DIFF
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -365,6 +365,10 @@ namespace tls {
         /// \brief Optional session resume data. Must be retrieved via
         /// get_session_resume_data below.
         session_data session_resume_data;
+
+        /// \brief Optional list of ALPN protocols to offer to the server,
+        /// in order of preference.
+        std::vector<sstring> alpn_protocols;
     };
 
     /**
@@ -536,6 +540,15 @@ namespace tls {
      * delay this call to sometime before shutting down/closing the socket.
     */
     future<session_data> get_session_resume_data(connected_socket&);
+
+    /**
+     * Gets the Application-Layer Protocol Name (ALPN) selected during the TLS handshake.
+     * Will force handshake if not already done.
+     *
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
+    */
+    future<std::optional<sstring>> get_selected_alpn_protocol(connected_socket&);
 
     std::ostream& operator<<(std::ostream&, const subject_alt_name::value_type&);
     std::ostream& operator<<(std::ostream&, const subject_alt_name&);
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -280,6 +280,12 @@ namespace tls {
          * a new window of TLS session keys.
         */
         void set_session_resume_mode(session_resume_mode);
+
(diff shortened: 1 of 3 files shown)


avikivity:
Please add a unit test.

crackcomm:
It turns out the `gnutls_free` was wrong, it got caught by asan in the test, it was gnutls internal session data.

I added the test with google but we cannot guarantee that it will negotiate h2, it can depend on the IP etc.

I added server side support and client-server test for better coverage, I will submit it as a separate PR so it's easier to review.

EDIT: after a second thought, I will just submit it here with two commits.

elcallio:
Now you have alpn:s in server_credentials and tls_options. Why not just place them in base credentials, and allow overriding per (client) connection basis with a std::optional in tls_options. That way client code can set this once and be done with it for many use cases.

crackcomm:
I opted for `alpn_protocols` in `tls_options` for the client because I think it's generally preferred, e.g. you load system certs once and can use them for many different connections, not all share the same protocol (akin to `server_name`). I would actually prefer to do the same for the server but `tls::listen` does not support `tls_options`, that's the only reason it exists in `server_credentials` atm.

include/seastar/net/tls.hh
```diff
@@ -537,6 +541,16 @@ namespace tls {
...
 
+    /**
+     * Gets the Application-Layer Protocol Name (ALPN) selected during the TLS handshake.
+     * Will force handshake if not already done.
+     *
+     * If the socket is not connected a system_error exception will be thrown.
+     * If the socket is not a TLS socket an exception will be thrown.
+     * If no protocol was negotiated an exception will be thrown.
+    */
+    future<sstring> get_selected_alpn_protocol(connected_socket&);
```
avikivity: Does this need to be a future? Once we have a connected_socket, isn't it already negotiated?

avikivity: I see it does.

avikivity: @elcallio perhaps we should have pushed the handshake into connect/accept.

src/net/tls.cc
```diff
@@ -1830,6 +1841,17 @@ class session : public enable_lw_shared_from_this<session> {
         });
     }
 
+    future<sstring> get_selected_alpn_protocol() {
```
avikivity: Better to return std::optional\<sstring> to make it clear when something wasn't negotiated.

include/seastar/net/tls.hh
crackcomm: If my understanding is correct that `tls::connect` doesn't perform a handshake and only wraps the socket. It's negotiated during the handshake.

elcallio (approved):

include/seastar/net/tls.hh
avikivity: Yes, I'm saying we should have pushed the handshake into connect() when we introduced tls. It returns a future and would report handshake errors early.

Now it's too late.

Issue #2783, opened 2025-05-27
HTTP client body data sinks copy data


xemul:
Recently, the sink impls were fixed not to crash in case caller tries to use zero-copy writing (52eb05829b271984dd4c79dae940057e70c13c2c), but the chunked and content-length sinks both use fallback implementation of `put(net::packet)` methods. The fallback, in turn, calls the `put(temporary_buffer<>)` one which, in case of http body sinks, both call `_out.write(buf.get(), buf.size())` and copy the buffer data into socket stream

regevran:
From the PR:
> Refs https://github.com/scylladb/seastar/issues/2783 (this patch fixes content-length bodies not to copy data, the
chunked-endoded bodies continue doing it)

@xemul - can you please open a relevant issue?

xemul:
Content-length sink is fixed by #2791 + #2785

PR #2791, opened 2025-06-03
http/client: Zero-copy forward content-length body into the underlying stream

xemul:
When making an http request, the caller may put the body-writer lambda
on the request. Depending whether the content-legnth value is specified,
the client will choose either content_length data sink or chunked one.

The data sink implements two put()-s -- one for net::packet and the
other one for temporary_buffer. The former nowadays falls-back to its
"default" implementation which just translates the packet into temporary
buffer and calls the latter (see https://github.com/scylladb/seastar/commit/52eb05829b271984dd4c79dae940057e70c13c2c commit).

When put() into content-length data sink, the temporary buffer is copied
into the underlying (socket) output_stream by calling its bufferred
.write() method:

   return _out.write(buf.get(), size) ...

This patch re-implements the packet overload of put() to zero-copy put
the packet into the underlying (socket) output_stream. Since by that
time the underlying stream already contains the http header in its
buffer, the stream will proceed in semi-mixed mode (see https://github.com/scylladb/seastar/pull/2785) and will
zero-copy append the packet, thus avoiding the unwanted copy.

Refs https://github.com/scylladb/seastar/issues/2783 (this patch fixes content-length bodies not to copy data, the
chunked-endoded bodies continue doing it)

### DIFF
--- a/src/http/common.cc
+++ b/src/http/common.cc
@@ -169,7 +169,16 @@ public:
         _bytes_written = 0;
     }
     virtual future<> put(net::packet data) override {
-        return data_sink_impl::fallback_put(std::move(data));
+        auto size = data.len();
+        if (size == 0) {
+            return make_ready_future<>();
+        }
+        if (_bytes_written + size > _limit) {
+            return make_exception_future<>(std::runtime_error(format("body content length overflow: want {} limit {}", _bytes_written + size, _limit)));
+        }
+        return _out.write(std::move(data)).then([this, size] {
+            _bytes_written += size;
+        });
     }
     using data_sink_impl::put;
     virtual future<> put(temporary_buffer<char> buf) override {


xemul:
upd:
- rebased

src/http/common.cc
```diff
@@ -169,7 +169,16 @@ class http_content_length_data_sink_impl : public data_sink_impl {
         _bytes_written = 0;
     }
     virtual future<> put(net::packet data) override {
-        return data_sink_impl::fallback_put(std::move(data));
+        auto size = data.len();
+        if (size == 0 || _bytes_written == _limit) {
+            return make_ready_future<>();
+        }
+        if (_bytes_written + size > _limit) {
+            return make_exception_future<>(std::runtime_error(format("body conent length overflow: want {} limit {}", _bytes_written + size, _limit)));
```
avikivity: spelling

src/http/common.cc
```diff
@@ -169,7 +169,16 @@ class http_content_length_data_sink_impl : public data_sink_impl {
...
+        }
+        return _out.write(std::move(data)).then([this, size] {
+            _bytes_written += size;
+        });
     }
```
avikivity, line 182: The handling of overflow is inconsistent. If a packet straddles the limit, you throw. If a packet ends on the limit and another one starts at the limit, you ignore it.

src/http/common.cc
```diff
@@ -169,7 +169,16 @@ class http_content_length_data_sink_impl : public data_sink_impl {
...
-        return data_sink_impl::fallback_put(std::move(data));
+        auto size = data.len();
+        if (size == 0 || _bytes_written == _limit) {
+            return make_ready_future<>();
+        }
+        if (_bytes_written + size > _limit) {
+            return make_exception_future<>(std::runtime_error(format("body conent length overflow: want {} limit {}", _bytes_written + size, _limit)));
+        }
+        return _out.write(std::move(data)).then([this, size] {
+            _bytes_written += size;
```
avikivity, line 180: Better to adjust _bytes_written before the write, saves a continuation. If write() throws, we won't look at _bytes_written anyway.

Ah, you need to add the exact amount written in case the remote closed the connection during the transfer. And we can'd return a future carrying a size anyway. Oh well.

src/http/common.cc
xemul: Agree, ~~but~~ and it's pre-existing problem -- there's `put(temporary_buffer)` overload from which this overload is copied from. I'll fix it

xemul:
upd:
- spelling
- made body size overflow checks consistent

PR #2792, opened 2025-06-03
Upstream a couple of minor reactor optimizations

travisdowns:
These opts turned up from profiling have been in our branch for a couple of years but but somehow I never upstreamed them, but here they are now!

Details in the commits.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3245,7 +3245,7 @@ int reactor::do_run() {
 
     bool idle = false;
 
-    std::function<bool()> check_for_work = [this] () {
+    auto check_for_work = [this] () {
         return poll_once() || have_more_tasks();
     };
     std::function<bool()> pure_check_for_work = [this] () {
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -3245,7 +3245,7 @@ int reactor::do_run() {
 
     bool idle = false;
 
-    std::function<bool()> check_for_work = [this] () {
+    auto check_for_work = [this] () {
         return poll_once() || have_more_tasks();
     };
     std::function<bool()> pure_check_for_work = [this] () {
```
xemul: Why not convert it to auto as well? Just for symmetry

xemul: Ah, it's in the next patch :sweat_smile:

travisdowns: Yes, and actually they needed to be different types due to how they are used below (one a lambda, the other one a seastar function).

PR #2794, opened 2025-06-03
Handle ENOENT in getgrnam

### LINKED ISSUE #2793: getgrnam_r inconsistent return code
When the group is not found, `getgrnam_r` may return 0, or it may return ENOENT, depending on the configuration/packages installed on the underlying system. See https://sourceware.org/bugzilla/show_bug.cgi?id=33052 for details.

The reactor treats them differently: 0 return is empty optional, while ENOENT throws a std::system_exception. This breaks test `libc_wrapper.getgrnam_group_name_does_not_exist_test` and other use as well.

travisdowns:
When calling getgrnam_r, handle the case where the group does not exist (ENOENT) gracefully instead of treating it as a fatal error (which throws). The man page isn't completely clear on this, but it is clear that in practice with some packages installed like sssd this libc call stops returning 0 and starts returning ENOENT for getgrnam_r.

Fixes #2793.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2139,7 +2139,9 @@ future<std::optional<struct group_details>> reactor::getgrnam(std::string_view n
             return wrap_syscall(ret, std::optional<struct group_details>(gd));
         });
 
-    if (sr.result != 0) {
+    // Non-existent groups may return 0, or they may return ENOENT, depending on the
+    // configuration of the system, see seastar#2793. Treat both as "not found".
+    if (sr.result != 0 && sr.result != ENOENT) {
         throw std::system_error(sr.ec());
     }
 


nyh:
> The man page isn't completely clear on this,

By the way the man page on my distribution at least is very clear on this:

```
If no matching group record was found, these functions return 0  and store NULL in *result. 
```

So returning -1 and ENOENT is a bug, which they (sssd) need to fix, but it indeed can't hurt to support even this buggy situation, since it's not hard.

travisdowns:
> By the way the man page on my distribution at least is very clear on this:

We probably have the same man page, but yes that part _in isolation_ is clear, but some other parts muddy the water a bit. In any case, I did make it clear in the glibc bug, but the actual problem may be the nss plugin impl, let's see:

https://sourceware.org/bugzilla/show_bug.cgi?id=33052

PR #2795, opened 2025-06-04
input_stream: refactors the control flow in the read_exactly

kreuzerkrieg:
Remove unnecessary branching and condition from the `input_stream<CharType>::read_exactly`. Also stop calling `read_exactly` recursively.

### DIFF
--- a/include/seastar/core/iostream-impl.hh
+++ b/include/seastar/core/iostream-impl.hh
@@ -186,25 +186,15 @@ input_stream<CharType>::read_exactly(size_t n) noexcept {
     if (_buf.size() == n) {
         // easy case: steal buffer, return to caller
         return make_ready_future<tmp_buf>(std::move(_buf));
-    } else if (_buf.size() > n) {
+    }
+    if (_buf.size() > n) {
         // buffer large enough, share it with caller
         auto front = _buf.share(0, n);
         _buf.trim_front(n);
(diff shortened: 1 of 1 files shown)


include/seastar/core/iostream-impl.hh
```diff
@@ -186,25 +186,15 @@ input_stream<CharType>::read_exactly(size_t n) noexcept {
...
         return make_ready_future<tmp_buf>(std::move(front));
-    } else if (_buf.size() == 0) {
-        // buffer is empty: grab one and retry
-        return _fd.get().then([this, n] (auto buf) mutable {
-            if (buf.size() == 0) {
-                _eof = true;
-                return make_ready_future<tmp_buf>(std::move(buf));
-            }
-            _buf = std::move(buf);
-            return this->read_exactly(n);
```
xemul, line 202: The `_buf.size() == 0` case handling is not equivalent to read_exactly_part(). E.g., if _fd.get() returns large enough buffer, the old code would recurse into sharing the trimmer buffer with the caller, but new code will fall-back to copying the buffer content

kreuzerkrieg:
> The `_buf.size() == 0` case handling is not equivalent to read_exactly_part(). E.g., if _fd.get() returns large enough buffer, the old code would recurse into sharing the trimmer buffer with the caller, but new code will fall-back to copying the buffer content

Right, missed this part

PR #2796, opened 2025-06-05
Fix for non-json http function_handlers

travisdowns:
In https://github.com/scylladb/seastar/commit/94c4c7a9e98be31ff11b07b8020ccbee7153e672 I introduced a bug where a particular overload of `function_handler` which had an arbitrary (passed-in) content type and which should simply set the body to the string-like content without encoding, would be instead treated as json. This resulted in a spurious pair of quotes added in the encoding process. 

This PR has a fix for that, courtesy of @mmaslankaprv as well as a test to avoid this in the future.

### DIFF
--- a/include/seastar/http/function_handlers.hh
+++ b/include/seastar/http/function_handlers.hh
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include <concepts>
 #include <seastar/http/handlers.hh>
 #include <functional>
 #include <seastar/json/json_elements.hh>
@@ -90,7 +91,7 @@ public:
     function_handler(const json_request_function& _handle)
             : _f_handle(
                     [_handle](std::unique_ptr<http::request> req, std::unique_ptr<http::reply> rep) {
-                        return append_result(std::move(rep), _handle(*req.get()));
+                        return write_json_reply(std::move(rep), _handle(*req.get()));
                     }), _type("json") {
     }
 
@@ -98,7 +99,7 @@ public:
             : _f_handle(
                     [_handle](std::unique_ptr<http::request> req, std::unique_ptr<http::reply> rep) {
                         return _handle(std::move(req)).then([rep = std::move(rep)](json::json_return_type&& res) mutable {
-                            return append_result(std::move(rep), std::move(res));
+                            return write_json_reply(std::move(rep), std::move(res));
                         });
                     }), _type("json") {
     }
@@ -116,9 +117,9 @@ public:
 
 private:
     // send the json payload of result to reply, return the reply pointer
-    static future<std::unique_ptr<http::reply>> append_result(
+    static future<std::unique_ptr<http::reply>> write_json_reply(
         std::unique_ptr<http::reply>&& reply,
-        json::json_return_type&& result) {
+        std::same_as<json::json_return_type> auto&& result) {
         if (result._body_writer) {
             reply->write_body("json", std::move(result._body_writer));
         } else {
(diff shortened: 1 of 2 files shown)


tests/unit/httpd_test.cc
```diff
@@ -288,6 +290,20 @@ SEASTAR_TEST_CASE(test_routes) {
...
+    route.add(operation_type::GET, url("/hello"), new function_handler([](const_req req) {
+        return "hello, you";
+    }, "txt"));
+
+    auto reply = route.handle("/hello", std::make_unique<http::request>(),
+            std::make_unique<http::reply>()).get();
+    
+    BOOST_CHECK_EQUAL((int )reply->_status, (int )http::reply::status_type::ok);
+    BOOST_CHECK_EQUAL(reply->_headers["Content-Type"], "text/plain");
+    BOOST_CHECK_EQUAL(reply->_content, "hello, you");
```
travisdowns, line 304: Before the fix, this fails with `"hello, you" != hello, you`

nyh (approved):
Looks good, but I wonder, shouldn't `append_result()` be renamed? As you noticed now, this helper function is only for JSON responses, not for anything else, and it also doesn't really "append" anything, it writes. So why not call it something like `write_json()`? 
I think if you called it write_json(), you would have been less tempted to accidentally use it in a function that takes a string, not json.

By the way, it's probably not good that we have an implicit conversion from text to a json version of this text (i.e., add double quotes). This constructor should probably be made "explicit".

travisdowns:
> Looks good, but I wonder, shouldn't `append_result()` be renamed? As you noticed now, this helper function is only for JSON responses, not for anything else, and it also doesn't really "append" anything, it writes. So why not call it something like `write_json()`?

Great call, let me make this change.

> I think if you called it write_json(), you would have been less tempted to accidentally use it in a function that takes a string, not json.

+1

> By the way, it's probably not good that we have an implicit conversion from text to a json version of this text (i.e., add double quotes). This constructor should probably be made "explicit".

~I can do that, if you think this is not part of the public API.~

It seems a bit like this ship has sailed, the doc for this type lists "autoboxing" as a feature for writing succinct handler functions, even giving the string conversion as an example.

Instead, I prevented implicit conversions for this particular method using a concept (2nd change in this series).

include/seastar/http/function_handlers.hh
```diff
@@ -116,9 +117,9 @@ public:
 
 private:
     // send the json payload of result to reply, return the reply pointer
-    static future<std::unique_ptr<http::reply>> append_result(
+    static future<std::unique_ptr<http::reply>> write_json_reply(
         std::unique_ptr<http::reply>&& reply,
-        json::json_return_type&& result) {
+        std::same_as<json::json_return_type> auto&& result) {
```
nyh, line 122: Wow, this is the first time I ever saw anyone using this idiom. Nice trick, but continues the trend making C++ code even more opaque as the years go by :-(

I had to ask ChatGPT if this is a well-known idiom and it told me "Yes, it's a real and valid technique—an idiom of constrained forwarding references with std::same_as to enforce exact type matches. While it doesn't have a snappy name like some classic C++ idioms, it's becoming well-known in modern C++ circles post-C++20 and is considered good practice when you want to prevent implicit conversions in generic code."

nyh (approved):

include/seastar/http/function_handlers.hh
travisdowns: Well for full disclosure, while writing it I asked AI (a different one) what the current best idiom for this was in C++20 and that's what I got, so it's all very circular: LLM will back each other up I guess :).

If I had to come up with it myself I probably would have just used a `static_assert` inside the function asserting on the type.

dotnwat: >LLM will back each other up

🤣

PR #2797, opened 2025-06-06
json2code: Genrate move constructor and move-assignment operator

xemul:
Currently the script generates only default constructor and copy constructor (and a helper copy-from-mapper-type constructor). In Scylla there are several places that try to std::move the generated-types objects around, but objects are silently copied anyway.

Add move constructor and copy-assignment operator so that moving takes effect where possible.

Compile-tested with Scylla.

### DIFF
--- a/scripts/seastar-json2code.py
+++ b/scripts/seastar-json2code.py
@@ -563,6 +563,7 @@ def create_h_file(data, hfile_name, api_name, init_method, base_api):
             fprintln(hfile, "struct ", model_name, " : public json::json_base {")
             member_init = ''
             member_assignment = ''
+            member_move_assignment = ''
             member_copy = ''
             for member_name in model["properties"]:
                 member = model["properties"][member_name]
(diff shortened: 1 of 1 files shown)


xemul:
@nyh , please review

xemul:
@nyh , please review

scripts/seastar-json2code.py
```diff
@@ -589,6 +591,10 @@ def indent(s):
         register_params();
         $member_assignment
     }
+    $model_name($model_name&& e) {
```
avikivity, line 594: Is noexcept possible here? possibly conditional.

scripts/seastar-json2code.py
```diff
@@ -598,13 +604,18 @@ def indent(s):
         $member_assignment
         return *this;
     }
+    $model_name& operator=($model_name&& e) {
```
avikivity, line 607: Again noexcept if possible.

scripts/seastar-json2code.py
xemul: register_params() is not noexcept, so even if assignment is, the whole thing is not

xemul:
>  I don't know about the noexcept that Avi asked about. I don't understand which types are generated by this script and how we can know if their own move constructors are noexcept.

It generates the types described in [Scylla api-doc](https://github.com/scylladb/scylladb/blob/27f66fb1101d45f382b6ca0cbc9745cadaf9d6bd/api/api-doc/storage_service.json#L3166). Mostly it std::move-s sstrings and integers (and other api-doc types maybe), but as I [mentioned](https://github.com/scylladb/seastar/pull/2797#discussion_r2153846466) the move construtor and assign-operator both call `register_params()` which is _not_ noexcept. The latter probably can be marked conditionally noexcept, but [its default implementation](https://github.com/scylladb/seastar/blob/9f905b4c5c0794b363437562c7a800a2316c5aca/src/json/json_elements.cc#L165) calls `std::vector<>::push_back()` and AFAICS we don't have non-default one

PR #2799, opened 2025-06-09
Allow http exceptions to have content type

travisdowns:
This PR allows httpd exceptions to optionally have a content type: if present the usual json encoding of the error reply is disabled and the body is used as is, and the response is set to the specified content type. Nothing changes for clients which don't set the content type.
 
This allows httpd to support non-json content when an error occurs, as it is already does for non-error replies. It also allows it to support json exceptions where the except message is already fully-formed json, to be emitted into the body as-is, rather than a simple string to be used in the `to_json_exception` format.

We have been using this patch for a few years in production w/o issue.

Copilot says:

Key updates include modifications to the `base_exception` class, updates to the `generate_error_reply_and_close` method, and adjustments to error handling logic in various components.

#### Updates to `base_exception`:
* Added a new constructor to `base_exception` that accepts a `content_type` parameter, enabling the specification of a full response body format.
* Introduced a `content_type()` method in `base_exception` to retrieve the `content_type` value.

#### Modifications to error reply generation:
* Updated the `generate_error_reply_and_close` method in `httpd.hh` and `httpd.cc` to include a `content_type` parameter, which is used to set the content type of error replies. [[1]](diffhunk://#diff-fcaeaad69059b06ac52a9c2232f991adbc20e1b4ad142e43c9be240de9c492c1L122-R122) [[2]](diffhunk://#diff-7ea19841c89a6d665298b3e426bb2745445ca10abcbfbe7af47d9b05f085fcdaL206-R213)
* Enhanced the `read_one()` method in `httpd.cc` to pass the `content_type` from exceptions to the `generate_error_reply_and_close` method.

#### Adjustments to exception handling in routes:
* Modified the `exception_reply` method in `routes.cc` to check for the presence of a `content_type` in `base_exception` and set the error reply accordingly. If no `content_type` is provided, the error message is wrapped in a JSON response.

### DIFF
--- a/include/seastar/http/exception.hh
+++ b/include/seastar/http/exception.hh
@@ -43,6 +43,16 @@ public:
             : _msg(msg), _status(status) {
     }
 
+    /**
+     * A base_exception with a content_type is specifying a full response body, whereas
+     * a base_exception with only a _status is specifying a string that may be wrapped
+     * in e.g. a json_exception.
+     */
+    base_exception(const std::string& msg, http::reply::status_type status, const std::string &content_type)
+            : _msg(msg), _status(status), _content_type(content_type) {
+    }
+
+
     virtual const char* what() const noexcept {
         return _msg.c_str();
     }
@@ -54,9 +64,14 @@ public:
     virtual const std::string& str() const {
         return _msg;
     }
+
+    const std::string& content_type() const {
+        return _content_type;
+    }
 private:
     std::string _msg;
     http::reply::status_type _status;
+    std::string _content_type;
 
 };
 
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -203,11 +203,14 @@ set_request_content(std::unique_ptr<http::request> req, input_stream<char>* cont
     }
 }
 
-void connection::generate_error_reply_and_close(std::unique_ptr<http::request> req, http::reply::status_type status, const sstring& msg) {
+void connection::generate_error_reply_and_close(std::unique_ptr<http::request> req, http::reply::status_type status, const sstring& msg, const sstring &content_type) {
     auto resp = std::make_unique<http::reply>();
     // TODO: Handle HTTP/2.0 when it releases
     resp->set_version(req->_version);
     resp->set_status(status, msg);
+    if (!content_type.empty()) {
+        resp->set_content_type(content_type);
+    }
     resp->done();
     _done = true;
     _replies.push(std::move(resp));
@@ -293,7 +296,7 @@ future<> connection::read_one() {
                     // before passing the request to handler - when we were parsing chunks
                     auto err_req = std::make_unique<http::request>();
                     err_req->_version = version;
-                    generate_error_reply_and_close(std::move(err_req), e.status(), e.str());
+                    generate_error_reply_and_close(std::move(err_req), e.status(), e.str(), e.content_type());
                 });
             });
         });
--- a/src/http/routes.cc
+++ b/src/http/routes.cc
@@ -58,6 +58,7 @@ routes::~routes() {
 
 std::unique_ptr<http::reply> routes::exception_reply(std::exception_ptr eptr) {
     auto rep = std::make_unique<http::reply>();
+    sstring content_type = "json";
     try {
         // go over the register exception handler
         // if one of them handle the exception, return.
@@ -76,13 +77,18 @@ std::unique_ptr<http::reply> routes::exception_reply(std::exception_ptr eptr) {
        rep.reset(new http::reply());
        rep->add_header("Location", _e.url).set_status(_e.status());
     } catch (const base_exception& e) {
-        rep->set_status(e.status(), json_exception(e).to_json());
+        if (e.content_type().size()) {
+            rep->set_status(e.status(), e.str());
+            content_type = e.content_type();
+        } else {
+            rep->set_status(e.status(), json_exception(e).to_json());
+        }
     } catch (...) {
         rep->set_status(http::reply::status_type::internal_server_error,
                 json_exception(std::current_exception()).to_json());
     }
 
-    rep->done("json");
+    rep->done(content_type);
     return rep;
 }
 
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -298,7 +298,7 @@ SEASTAR_THREAD_TEST_CASE(test_text_route) {
 
     auto reply = route.handle("/hello", std::make_unique<http::request>(),
             std::make_unique<http::reply>()).get();
-    
+
     BOOST_CHECK_EQUAL((int )reply->_status, (int )http::reply::status_type::ok);
     BOOST_CHECK_EQUAL(reply->_headers["Content-Type"], "text/plain");
     BOOST_CHECK_EQUAL(reply->_content, "hello, you");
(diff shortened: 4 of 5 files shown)


include/seastar/http/exception.hh
```diff
@@ -54,9 +64,14 @@ public:
     virtual const std::string& str() const {
         return _msg;
     }
+
+    virtual const std::string& content_type() const {
```
xemul: Does this really need to be virtual?

src/http/routes.cc
```diff
@@ -76,7 +76,12 @@ std::unique_ptr<http::reply> routes::exception_reply(std::exception_ptr eptr) {
        rep.reset(new http::reply());
        rep->add_header("Location", _e.url).set_status(_e.status());
     } catch (const base_exception& e) {
-        rep->set_status(e.status(), json_exception(e).to_json());
+        if (e.content_type().size()) {
```
xemul, line 80: The patch comment says

> This enables throwing a base_exception from a json request handler with a json payload inside it.

Isn't it a misprint? The "... with a json payload inside it" is what errors currently have, this patch doesn't _add_ this ability

xemul:
Please, also add a test to unit/httpd_test to demonstrate how the new API is to be used (and to test it works, of course)

include/seastar/http/exception.hh
travisdowns: No, removed.

src/http/routes.cc
travisdowns: It's a bit confusing: John is saying that you can pass content type "json" in order to disable the jsonizing of the exception string, because you are already passing fully-formed json.

I.e., if you do:

```
base_exception("foobar", 400) // non content-type ctor
```
you get as the body `{"message": "foobar", "code": 400}`. however, if you want pass a custom error json, you can't use that ctor since the json will get double encoded. So you use:

```
base_exception("{my_object: {}}", 400, "json") // content-type ctor
```
Fixed up the commit message and cover letter to clarify.

travisdowns:
@xemul wrote:

> Please, also add a test to unit/httpd_test to demonstrate how the new API is to be used (and to test it works, of course)

Added, and in fact it caught a bug where the content type was being overwritten to json again. Fixed in push.

src/http/httpd.cc
```diff
@@ -293,7 +296,7 @@ future<> connection::read_one() {
                     // before passing the request to handler - when we were parsing chunks
                     auto err_req = std::make_unique<http::request>();
                     err_req->_version = version;
-                    generate_error_reply_and_close(std::move(err_req), e.status(), e.str());
+                    generate_error_reply_and_close(std::move(err_req), e.status(), e.str(), e.content_type());
```
xemul, line 299: Need a clarification (here or in patch comment) -- how can an exception with non-empty content_type() happen to be here? The only way seems to be from handlers, but handlers exceptions should have been handled by `return generate_reply(std::move(req));` call so handler exception should not reach here. Where am I wrong?

tests/unit/httpd_test.cc
```diff
@@ -1380,6 +1380,61 @@ SEASTAR_TEST_CASE(test_unparsable_request) {
...
 
+struct throwing_handler : public httpd::handler_base {
+    bool is_json;
+
+    throwing_handler(bool is_json) : is_json{is_json} {}
+
+    future<std::unique_ptr<http::reply>> handle(const sstring&,
+            std::unique_ptr<http::request>, std::unique_ptr<http::reply>) override {
+        if (is_json) {
+            throw base_exception("json error", http::reply::status_type::forbidden);
```
xemul, line 1391: This comment is thinking out loud, I don't propose to do anything about it right now.

It looks like (and [this](https://github.com/scylladb/seastar/pull/2799/commits/d73ad06ae928c362e3793a0110bc97646e5f394e) patch supports that) the only way to emit non-json reply is by throwing base_exception. In that case, maybe the whole exception hierarchy should be revisited?

Currently any exception thrown from handler inherits from base_exception. This implicitly mean, that it will be jsonized by wrapping base_exception into json_exception and formatting it into json. This PR effectively adds a switch into base_exception -- whether or not to jsonize it (via wrapping into json_exception, etc.)

It looks more natural to make existring exceptions inherit from json_exception and json_exception inherit from base_exception. Then the code that catches base_exception from handler would just "format" it into the reply, and if the real exception behind it is json_exception, it will be transparently jsonized by the class itself.

Thoughts?

travisdowns: I agree with your reasoning, a separate exception type would be better, and this would be most evident also in that it could have different methods, or at least named differently than the base. E.g. "message" makes sense for the way we encode exceptions into json objects, but less so when it is the body of the response. 

That said, such a think would be a breaking change for existing users I think?

src/http/httpd.cc
travisdowns: I cannot see any way and indeed the tests I added only exercise the other place content type is used.

tests/unit/httpd_test.cc
xemul: Yes, it can be breaking :(

include/seastar/http/exception.hh
```diff
@@ -54,9 +64,14 @@ public:
...
         return _msg;
     }
+
+    const std::string& content_type() const {
+        return _content_type;
+    }
 private:
     std::string _msg;
     http::reply::status_type _status;
+    std::string _content_type;
```
xemul, line 74: Please, initialize it to "json" by default, as this is what really happens when this exception is caught

src/http/routes.cc
```diff
@@ -76,13 +77,18 @@ std::unique_ptr<http::reply> routes::exception_reply(std::exception_ptr eptr) {
...
+        } else {
+            rep->set_status(e.status(), json_exception(e).to_json());
+        }
     } catch (...) {
         rep->set_status(http::reply::status_type::internal_server_error,
                 json_exception(std::current_exception()).to_json());
     }
 
-    rep->done("json");
+    rep->done(content_type);
```
xemul, line 91: If base_excpetion has default value for its content type, this can be replaced with `rep->done(e.content_type())`

xemul: Yikes, we don't have `e` here.

src/http/httpd.cc
xemul: Then let's drop this change

tests/unit/httpd_test.cc
xemul: Meanwhile deprecated json_exception #2813

include/seastar/http/exception.hh
travisdowns: Empty string means something different than json (in the current patch). Here are the two cases:

1) Empty content type: the pre-exisitng behavior of taking the `_msg` (which is _not_ json) and encoding it into the fixed-format json message using ` json_exception(e).to_json()` like `{"code": N, "message": "text"}`.
2) Non-empty content type: we take `_msg` as the body (without transforming it), and the content type as the content type.

So setting this to json would change the default from (1) to (2) and break everything. Specifically we have case 2a) which is when the content type is explicitly "json", the `_msg` should itself be valid json and it returned as the body without further encoding.

xemul: Yes, you're right, I missed that :( In that case I really insist that we have a separate class for "raw body" exceptions, overloading base_exception with more functionality doesn't seem worthwhile.

travisdowns: That's a good suggestion.

I'm going to put this PR on ice for now as I'll be out and not have time to re-work it in the near future, but I may come back to it some day.

travisdowns:
Closing for now.

src/http/httpd.cc
travisdowns: Not sure about that: it seems like any place we handle base_exception we should handle the content type correctly, even if a careful (?) analysis shows that it can only have certain values at this moment: it would be easy for that to change in the future and nobody would know to undo the assumption in the handler.

However, I think this problem goes away anyway in your suggestion to split the exception types: in that case we could simply not catch the new type and it would propagate through.

PR #2800, opened 2025-06-09
httpd: allow associating a tag with each listener

travisdowns:
Seastar's http server supports listening on multiple addresses, but a request handler cannot tell which listener a request arrived on. Redpanda needs this to apply a different authentication method per endpoint (we have carried a variant of this in our fork for several years).

Looking up the server address of the request does not work in general: a listener bound to a wildcard address reports the resolved interface address on accepted connections, so there is no way to map a request back to the address that was configured.

So `http_server::listen()` now takes an optional tag:

```cpp
enum class listener_tag : uintptr_t {};
constexpr listener_tag no_listener_tag{};

future<> listen(socket_address addr, listen_options lo, listener_tag tag = no_listener_tag);
```

Connections accepted by that listener carry the tag, and every request received on such a connection reports it via `http::request::get_listener_tag()`. A request that arrived on a listener created without a tag, or that was not received by an `http_server` at all, reports `no_listener_tag`.

The representation is `uintptr_t` so the caller can either use the tag as an index into its own per-endpoint state or hold a pointer to that state in it. Seastar attaches no meaning to the value; the strong enum just keeps it from being confused with an unrelated integer at the call site.

`http_server_control::listen()` is intentionally left alone: a sharded server that wants tags can go through `server()` and set up the per-shard `http_server` directly.

Revisions, for anyone reading the discussion below out of order:

1. exposed the listener index.
2. took a `std::any`, which accepted a value of any type but pulled `<any>` into the public headers and made the accessor hand back a reference to `any_cast`.
3. (this one) a strong type over `uintptr_t`, as suggested by @avikivity. No type erasure, and no ownership or lifetime question, since the tag is a trivially copyable value held by value.

Tested by a new `test_listener_tag` unit test: two tagged listeners plus one untagged one, checking that each request reports the tag of the listener that accepted it.

### DIFF
--- a/include/seastar/http/common.hh
+++ b/include/seastar/http/common.hh
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <unordered_map>
 
 #include <seastar/core/sstring.hh>
@@ -41,6 +42,22 @@ output_stream<char> make_http_content_length_output_stream(output_stream<char>& 
 
 namespace httpd {
 
+/// An opaque value the caller may associate with an http_server listener.
+///
+/// The value is attached to a listener by passing it to \ref
+/// http_server::listen(); every request received on a connection accepted by
+/// that listener reports it via \ref http::request::get_listener_tag(). Seastar
+/// attaches no meaning to the value: use it as an index into the caller's own
+/// array of per-endpoint state, or hold a pointer to such state in it (hence
+/// the uintptr_t representation).
+///
+/// Requests arriving on a listener created without a tag report \ref
+/// no_listener_tag.
+enum class listener_tag : uintptr_t {};
+
+/// The tag reported for a request that arrived on a listener created without
+/// one, or that was not received by an http_server at all.
+constexpr listener_tag no_listener_tag{};
 
 class parameters {
     // Note: the path matcher adds parameters with the '/' prefix into the params map (eg. "/param1"), and some getters
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -77,26 +77,30 @@ class connection : public boost::intrusive::list_base_hook<> {
     queue<std::unique_ptr<http::reply>> _replies { 10 };
     bool _done = false;
     const bool _tls;
+    const listener_tag _listener_tag;
 public:
-    connection(http_server& server, connected_socket&& fd, bool tls)
+    connection(http_server& server, connected_socket&& fd, bool tls, listener_tag tag = no_listener_tag)
             : _server(server)
             , _fd(std::move(fd))
             , _read_buf(_fd.input())
             , _write_buf(_fd.output())
             , _client_addr(_fd.remote_address())
             , _server_addr(_fd.local_address())
-            , _tls(tls) {
+            , _tls(tls)
+            , _listener_tag(tag) {
         on_new_connection();
     }
     connection(http_server& server, connected_socket&& fd,
-            socket_address client_addr, socket_address server_addr, bool tls)
+            socket_address client_addr, socket_address server_addr, bool tls,
+            listener_tag tag = no_listener_tag)
             : _server(server)
             , _fd(std::move(fd))
             , _read_buf(_fd.input())
             , _write_buf(_fd.output())
             , _client_addr(std::move(client_addr))
             , _server_addr(std::move(server_addr))
-            , _tls(tls) {
+            , _tls(tls)
+            , _listener_tag(tag) {
         on_new_connection();
     }
     ~connection();
@@ -124,6 +128,10 @@ class http_server_tester;
 
 class http_server {
     std::vector<server_socket> _listeners;
+    // Tags associated with the listeners, in the same order. May be shorter
+    // than _listeners if listeners were added directly (e.g. via
+    // http_server_tester); missing entries mean no_listener_tag.
+    std::vector<listener_tag> _listener_tags;
     http_stats _stats;
     uint64_t _total_connections = 0;
     uint64_t _current_connections = 0;
@@ -185,10 +193,17 @@ public:
     /// scheduling group.
     void set_request_scheduling_group(scheduling_group sg);
 
-    future<> listen(socket_address addr, server_credentials_ptr credentials);
-    future<> listen(socket_address addr, listen_options lo, server_credentials_ptr credentials);
-    future<> listen(socket_address addr, listen_options lo);
-    future<> listen(socket_address addr);
+    /// Start listening on the given address.
+    ///
+    /// \param tag an opaque value associated with this listener. Every request
+    ///     received on a connection accepted by this listener exposes it via
+    ///     \ref http::request::get_listener_tag(), letting a handler tell
+    ///     listeners apart (e.g. to apply a different authentication policy per
+    ///     endpoint). See \ref listener_tag.
+    future<> listen(socket_address addr, server_credentials_ptr credentials, listener_tag tag = no_listener_tag);
+    future<> listen(socket_address addr, listen_options lo, server_credentials_ptr credentials, listener_tag tag = no_listener_tag);
+    future<> listen(socket_address addr, listen_options lo, listener_tag tag = no_listener_tag);
+    future<> listen(socket_address addr, listener_tag tag = no_listener_tag);
     future<> stop();
 
     future<> do_accepts(int which);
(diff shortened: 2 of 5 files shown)


avikivity:
Index seems wrong, we should pass some interface that one can call methods on.

travisdowns:
So instead of:

```
connection(http_server& server, connected_socket&& fd, bool tls, int listener_idx)
```

something like:

```
struct listener_opaque {
  virtual ~listener_opaque() = default;
};

using opaque_ptr = std::unique_ptr<listener_opaque>;

connection(http_server& server, connected_socket&& fd, bool tls, opaque_ptr opaque);
```

and then application creates subclasses of `listener_opaque` and application gets them from the request and dynamic casts them to the expected type?

xemul:
How do you map listener index to the data you need? Presumably when you call `http_server::listen()` you "map" the invocation sequence number (the index itself) to the context it was made from? If so, can the `addr` be used as the mapping key?

travisdowns:
> How do you map listener index to the data you need? Presumably when you call `http_server::listen()` you "map" the invocation sequence number (the index itself) to the context it was made from? 

N endpoints are listened on and we rely on the indexes being 0..N-1 in that order:

https://github.com/redpanda-data/redpanda/blob/43a18933f071b7f2db8b774b047ff054b0ef2f31/src/v/pandaproxy/server.cc#L297

> If so, can the `addr` be used as the mapping key?

It looks possible, I'll ask.

travisdowns:
@xemul - BTW, I suppose we are guaranteed that `addr` are unique keys at least for a single http_server object, since it would not be possible to listen on the same address twice on the same server?

avikivity:
> So instead of:
> 
> ```
> connection(http_server& server, connected_socket&& fd, bool tls, int listener_idx)
> ```
> 
> something like:
> 
> ```
> struct listener_opaque {
>   virtual ~listener_opaque() = default;
> };
> 
> using opaque_ptr = std::unique_ptr<listener_opaque>;
> 
> connection(http_server& server, connected_socket&& fd, bool tls, opaque_ptr opaque);
> ```
> 
> and then application creates subclasses of `listener_opaque` and application gets them from the request and dynamic casts them to the expected type?

I feel a little bad about having an abstract class that has no virtual methods, it basically says static_cast me because there is no elegant way to reach the virtual method.

The routes thing is completely orthogonal, yes? You want each listener to respond to exactly the same routes, but still each request wants to know what port/IP it listened on.

Anyway, your proposed solution looks okay. I don't think unique_ptr makes sense though, the listener should be owned by the user. And no need for _opaque to remind me forever that I don't like it.

avikivity:
> @xemul - BTW, I suppose we are guaranteed that `addr` are unique keys at least for a single http_server object, since it would not be possible to listen on the same address twice on the same server?

You could indeed use (addr, port) as a lookup key in a hash table, if it's not performance sensitive. That doesn't require adding anything except some accessors to the connection.

travisdowns:
> I feel a little bad about having an abstract class that has no virtual methods, it basically says static_cast me because there is no elegant way to reach the virtual method.

Yes, that's the price we pay for type-erasure I guess?
 
> The routes thing is completely orthogonal, yes? You want each listener to respond to exactly the same routes, but still each request wants to know what port/IP it listened on.

That's right.

avikivity:
> > I feel a little bad about having an abstract class that has no virtual methods, it basically says static_cast me because there is no elegant way to reach the virtual method.
> 
> Yes, that's the price we pay for type-erasure I guess?
> 
> > The routes thing is completely orthogonal, yes? You want each listener to respond to exactly the same routes, but still each request wants to know what port/IP it listened on.
> 
> That's right.

Then why not look up the address/port?

I don't mind a type-erased thing (I don't think the unique_ptr is necessary), but maybe the whole thing is unneeded.

travisdowns:
> Then why not look up the address/port?

It's a good question and as above I'm not sure: I wasn't the original user, just trying to upstream things we are carrying in our fork that might make sense upstream. 

I'll move this into draft while I I look into it on our end.

StephanDollberg:
> Then why not look up the address/port?

I think this doesn't work because of wildcard addresses. You bind to 0.0.0.0 but then the server socket address that you get out of the request is the actual resolved interface so it's not immediately trivial to map back to what was originally configured with 0.0.0.0?

So using the index is actually the best deterministic identifier.

travisdowns:
@avikivity I reworked this per the discussion: the index is gone. `listen()` now takes an optional `std::any` tag, and handlers read it back via `request::get_listener_tag()` plus `std::any_cast`. That gives the type-erased value without the empty abstract base class, and without the `unique_ptr`: the tag is held by value (the server owns it, each connection carries a copy).

As @StephanDollberg noted, looking up the address/port does not work for wildcard binds, but a tag attached at `listen()` time does not have that problem.

Also rebased on current master and added a unit test covering tagged and untagged listeners.

avikivity:
I don't think std::any is required. Create an empty class http_listener and pass a reference to it, which gets passed back to the connection. The user static_casts it to the real class type and accesses anything they need.

travisdowns:
@avikivity wrote:

> I don't think std::any is required. Create an empty class http_listener and pass a reference to it, which gets passed back to the connection. The user static_casts it to the real class type and accesses anything they need.

I don't like `std::any` any more than the next guy, but I'm still not getting it.

The user does like:

```
struct my_index : public http_listener {
    int index;
 }
 ```
 
 and passes that in? How is the listener going to store it? If the member type is `http_listener` it will be sliced, the user will get an http_listener back, and the static_cast is UB (and wont' work regardless).
 
 As far as I can tell you type erase here, if you want to move/copy the thing in then `std::any` or equivalent, or if you want to allocate it on the heap, then std::unique_ptr + virtual dtor or something like that.
 
 _or_ we can just go back to making it an int :)

avikivity:
> @avikivity wrote:
> 
> > I don't think std::any is required. Create an empty class http_listener and pass a reference to it, which gets passed back to the connection. The user static_casts it to the real class type and accesses anything they need.
> 
> I don't like `std::any` any more than the next guy, but I'm still not getting it.
> 
> The user does like:
> 
> ```
> struct my_index : public http_listener {
>     int index;
>  }
> ```

> 
> and passes that in? How is the listener going to store it? If the member type is `http_listener` it will be sliced, the user will get an http_listener back, and the static_cast is UB (and wont' work regardless).

Store an http_listener&.

> 
> As far as I can tell you type erase here, if you want to move/copy the thing in then `std::any` or equivalent, or if you want to allocate it on the heap, then std::unique_ptr + virtual dtor or something like that.
> 
> _or_ we can just go back to making it an int :)

Or a uintptr_t, so the caller can make it a pointer.

I just dislike untyped stuff.

What's useful for the user? Surely the int will be used to index something? Then just store a reference to that something, no?

travisdowns:
> Store an http_listener&.

We can't because the user-side thing is destroyed later. The listener owns the object.

> What's useful for the user? Surely the int will be used to index something? 

I can't speak to all users but we use the index to look up in a map or something like that.

> Then just store a reference to that something, no?

This is possible in some cases (might as well take a `void *` then), but it seems very fragile and unsafe by design: it can't survive a move, you can never point into a container unless you can guarantee iterator stability, etc. There are a lot of footguns you just avoid if you use a by value object stored in the listender.

Plus, in effect, even if your suggestion you still need erasure: `uintptr_t` or `void *` is just poor-man's erasure where you hope the object will fit in 64 bits.

That said, `uintptr_t` works for us, so I'm happy to go that way if you find it acceptable (because I will store a value in it, not  a reference).

avikivity:
> > Store an http_listener&.
> 
> We can't because the user-side thing is destroyed later. The listener owns the object.
> 
> > What's useful for the user? Surely the int will be used to index something?
> 
> I can't speak to all users but we use the index to look up in a map or something like that.
> 
> > Then just store a reference to that something, no?
> 
> This is possible in some cases (might as well take a `void *` then), but it seems very fragile and unsafe by design: it can't survive a move, you can never point into a container unless you can guarantee iterator stability, etc. There are a lot of footguns you just avoid if you use a by value object stored in the listender.
> 
> Plus, in effect, even if your suggestion you still need erasure: `uintptr_t` or `void *` is just poor-man's erasure where you hope the object will fit in 64 bits.
> 
> That said, `uintptr_t` works for us, so I'm happy to go that way if you find it acceptable (because I will store a value in it, not a reference).

Ok, let's go with uintptr_t.

Perhaps wrap it in an enum class : uintptr_t so it's a strong type.

travisdowns:
Done, pushed. `listen()` now takes `enum class listener_tag : uintptr_t {}` (with `no_listener_tag` for the untagged case), and `request::get_listener_tag()` returns it by value. `<any>` is gone from the public headers, and since the tag is a trivially copyable value there is no ownership or lifetime question left. Description updated.

I left `http_server_control::listen()` alone rather than plumbing the tag through to every shard; a sharded server that wants tags can go through `server()`. Happy to add it if you would prefer the wrapper to cover it.

The unit test is updated too: two tagged listeners plus an untagged one, with the tag used as an index into a caller-side array, which is the intended idiom.

PR #2805, opened 2025-06-11
json: json_elements: add generate_array

bhalevy:
Add a helper function to generate a json array gently by consuming its items from a coroutine generator.

The motivation for this change is to avoid large allocations that are needed for generating a long json array (See https://github.com/scylladb/scylladb/pull/24261).

With the coroutine generator, we can consume and process the array elements one at a time (optionally using a small buffer to reduce context switches between the generator and the consumer), and stream them as json, say to an output file, or to a socket, as part of a higher level protocol.

### DIFF
--- a/include/seastar/json/json_elements.hh
+++ b/include/seastar/json/json_elements.hh
@@ -389,8 +389,8 @@ json_return_type::body_writer_type stream_range_as_array(Container val, Func fun
  * \brief consume jsonable values from a coroutine generator \c gen and write them
  * onto the \c out output_stream as a json array.
  */
-template<Jsonable T>
-future<> generate_array_element(output_stream<char>& out, coroutine::experimental::generator<T>& gen) {
+template<Jsonable T, template<typename> class Container>
+future<> generate_array_element(output_stream<char>& out, coroutine::experimental::generator<T, Container>& gen) {
     bool first = true;
     co_await out.write("[");
     while (auto val = co_await gen()) {
@@ -418,8 +418,8 @@ future<> generate_array_element(output_stream<char>& out, coroutine::experimenta
  * Note that \c gen is passed by reference since we need to return a copyable function but generators cannot be copied.
  * So the caller is responsible for ensuring that the generator remains valid for the lifetime of the returned function.
  */
-template<Jsonable T>
-json_return_type::body_writer_type generate_array(coroutine::experimental::generator<T> gen) {
+template<Jsonable T, template<typename> class Container>
+json_return_type::body_writer_type generate_array(coroutine::experimental::generator<T, Container> gen) {
     return [gen_ = std::move(gen)] (output_stream<char>&& s) mutable -> future<> {
         auto gen = std::move(gen_);
         auto out = std::move(s);
--- a/tests/unit/json_formatter_test.cc
+++ b/tests/unit/json_formatter_test.cc
@@ -25,6 +25,7 @@
 #include <seastar/testing/test_case.hh>
 #include <seastar/core/sstring.hh>
 #include <seastar/core/vector-data-sink.hh>
+#include <seastar/core/circular_buffer.hh>
 #include <seastar/json/formatter.hh>
 #include <seastar/json/json_elements.hh>
 #include <seastar/testing/thread_test_case.hh>
@@ -157,6 +158,35 @@ SEASTAR_THREAD_TEST_CASE(test_generate_array) {
     // Copy the streamer function on purpose to ensure it doesn't cause
     // use-after-stack-return.
     auto streamer = generate_array(generate_json_values());
+
+    formatter_check_expected(expected, [streamer = std::move(streamer)] (auto& out) {
+        streamer(std::move(out)).get();
+    }, false);
+}
+
+namespace {
+template<template<typename> class Container>
+coroutine::experimental::generator<object_json, Container> generate_values_buffered(coroutine::experimental::buffer_size_t size, int count) {
+    for (int i = 1; i <= count; ++i) {
+        object_json obj;
+        obj.subject = std::to_string(i);
+        obj.values.push(i);
+        co_yield obj;
+    }
+}
+
+template<typename T>
+using buffered_container = circular_buffer<T>;
+}
+
+SEASTAR_THREAD_TEST_CASE(test_generate_array_buffered) {
+    sstring expected = R"([{"subject":"1","values":[1]}, {"subject":"2","values":[2]}, {"subject":"3","values":[3]}])";
+    // Test buffer size larger than the number of elements
+    auto gen = generate_values_buffered<buffered_container>(coroutine::experimental::buffer_size_t(16), 3);
+
+    // Copy the streamer function on purpose to ensure it doesn't cause
+    // use-after-stack-return.
+    auto streamer = generate_array(std::move(gen));
 
     formatter_check_expected(expected, [streamer = std::move(streamer)] (auto& out) {
         streamer(std::move(out)).get();


include/seastar/json/json_elements.hh
```diff
@@ -367,6 +370,45 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
...
+        std::exception_ptr ex;
+        try {
+            auto fun = std::move(fun_);
+            auto first = true;
+            co_await out.write("[");
+            while (auto val = co_await gen()) {
+                if (first) {
+                    first = false;
+                } else {
+                    co_await out.write(", ");
```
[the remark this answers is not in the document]
bhalevy: eh?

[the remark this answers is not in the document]
amnonh: @bhalevy this PR does not close any issue, I'm not sure that streaming an array is such an interesting case.
you can do the general case instead.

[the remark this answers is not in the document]
bhalevy: If it's a long array we would like to prevent both large allocations and reactor stalls, hence the generator.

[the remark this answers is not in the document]
bhalevy: > @bhalevy this PR does not close any issue, I'm not sure that streaming an array is such an interesting case. you can do the general case instead.

It wasn't intended to close a seastar issue but rather as a solution for https://github.com/scylladb/scylladb/issues/24216

include/seastar/json/json_elements.hh
```diff
@@ -367,6 +370,45 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
...
+ * To use it, pass a reference to coroutine::generator and a mapping function.
+ * For example, if gen is a coroutine::generator<std::pair<K, V>> and you want to serialize the keys:
+ *
+ * return make_ready_future<json::json_return_type>(generate_array(gen, [](const auto&i) {return i.first}));
+ *
+ * Note that \c gen is passed by reference since we need to return a copyable function but generators cannot be copied.
+ */
+template<typename T, typename Func>
+requires requires (T val, Func aa, output_stream<char> s) { { formatter::write(s, aa(val)) } -> std::same_as<future<>>; }
+std::function<future<>(output_stream<char>&&)> generate_array(coroutine::experimental::generator<T>& gen, Func fun) {
```
robertbindar: Can we instead make the return function get a reference to the output stream and not make it responsible for closing the stream?
I'll use this for streaming an object like: `{ "key": [el1, el2, el3]}` That means I'll need to open a stream, write "{", use the output function of `generate_array` to get `[...]` and then open stream once again to write "}" and close.

robertbindar (approved):
Nice, I know understand what you wanted in https://github.com/scylladb/scylladb/pull/24261/
Please see the comment if it's possible to accommodate that.

include/seastar/json/json_elements.hh
bhalevy: > Can we instead make the return function get a reference to the output stream and not make it responsible for closing the stream? I'll use this for streaming an object like: `{ "key": [el1, el2, el3]}` That means I'll need to open a stream, write "{", use the output function of `generate_array` to get `[...]` and then open stream once again to write "}" and close.

I'm not sure what you mean.
The return function does get a (rvalue) reference to the output stream:
`std::function<future<>(output_stream<char>&&)>`

bhalevy: Do you mean pass an lvalue reference and let the caller close the stream when done?
It does make sense but it changes the api from `stream_range_as_array`

bhalevy: > I'll use this for streaming an object like: { "key": [el1, el2, el3]} That means I'll need to open a stream, write "{", use the output function of generate_array to get [...] and then open stream once again to write "}" and close.

The intended usage is to call generate_array to retrieve the json "encoding" function that encloses the array data in square brackets and then closes the stream.

bhalevy: Hmm, but we want to also enclose the json data in `"{ \t files: [" ... "] }"`

bhalevy: The problem is with the expectation of returning a `json_return_type` with the returned function a its `_body_writer`

bhalevy: All that said, I see that `write_to_stream_and_close` closes the output stream too after invoking the writer.
@xemul can you please explain?

robertbindar: > Do you mean pass an lvalue reference and let the caller close the stream when done? It does make sense but it changes the api from `stream_range_as_array`

that's what I mean yes, the api is ok as it is, but when you need something like `{ files: [....] }`, you have to reopen the stream after calling this function.

bhalevy: I'll send a refactored version that uses a utility function like generate_array_element that will encode the array onto the output stream without consuming and closing the syream itself so it can be used as a building block.

robertbindar: > The problem is with the expectation of returning a `json_return_type` with the returned function a its `_body_writer`

we can template out `json_return_type` with default on `output_stream<char>&&` so no outside code needs changing. It depends how far you want to go to stay consistent with the other functions. These functions seems more like utility functions rather than an API, and for `generate_array` the only user is in https://github.com/scylladb/scylladb/pull/24261/, it's probably alright if we diverge from the pattern of the other functions.

robertbindar: > I'll send a refactored version that uses a utility function like generate_array_element that will encode the array onto the output stream without consuming and closing the syream itself so it can be used as a building block.

looking forward to it

bhalevy: See v2 ([1e79993](https://github.com/scylladb/seastar/commit/1e79993415a7a9203fb0c80f7634ad4ac142060c))

include/seastar/json/json_elements.hh
```diff
@@ -367,6 +370,47 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
     };
 }
 
+/*!
+ * \brief consume values from a coroutine generator \c gen and encode the items using the \c fun mapping function
+ * onto the \c out output_stream as a json array.
+ */
+template<typename T, typename Func>
+requires requires (T val, Func aa, output_stream<char> s) { { formatter::write(s, aa(val)) } -> std::same_as<future<>>; }
+future<> generate_array_element(output_stream<char>& out, coroutine::experimental::generator<T>& gen, Func fun) {
```
avikivity: Looking at std::generator, it implements std::ranges::view_interface, so you normally wouldn't consume a generator, you consume a view. The user can pass anything that's a view (like std::views::iota), or a generator if they like.

We should have the concept of a streaming view: something that has std::optional\<T> operator()().

avikivity: Why an extra mapping? Seems like it can be applied outside.

avikivity: Why generate_array_element? It's writing the whole array, no?

include/seastar/json/json_elements.hh
```diff
@@ -367,6 +370,47 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
...
+            co_await out.write(", ");
+        }
+        co_await formatter::write(out, fun(*val));
+    }
+    co_await out.write("]");
+ }
+
+/*!
+ * \brief consume values from a coroutine generator \c gen and return a serialize function for it as a json array.
+ *
```
avikivity, line 412: What's a serialize function? Is it an existing concept?

avikivity:
Adding @travisdowns for cross-project fertilization.

In scylladb.git, we have the concept of streams - classes with

```c++
    future<std::optional<T>> operator()(); // get next element of nullopt on end-of-stream
    future<> close(); // must be called to release resources, even on exceptions
```

This is also implemented by seastar::experimental::generator, but it is experimental.

Do you have something similar? It seems high time that Seastar had its own concept of an asynchronous stream. I'd like to merge ideas from other repos.

include/seastar/json/json_elements.hh
robertbindar: > Why generate_array_element? It's writing the whole array, no?

I think `element` here means `json element`, which is still confusing, but I don't have a better idea (maybe generate_json_array_type?), i.e. `{ "some_array": generage_array_element() }`

bhalevy: > Why generate_array_element? It's writing the whole array, no?

The name isn't great, it's not generating an element in an array, but rather the whold array as a json element.

bhalevy: > Why an extra mapping? Seems like it can be applied outside.

Hmm, the interface evolved from stream_range_as_array hence the mapping function, but I agree that using a generator indeed allows the user to compose the mapper externally as part of the generator.

bhalevy: > Looking at std::generator, it implements std::ranges::view_interface, so you normally wouldn't consume a generator, you consume a view. The user can pass anything that's a view (like std::views::iota), or a generator if they like.
> 
> We should have the concept of a streaming view: something that has std::optional<T> operator()().

But it's an async generator so the view should have `future<std::optional<T>> operator()()`

include/seastar/json/json_elements.hh
bhalevy: yes, the concept is documented as such in `stream_range_as_array`:
```
 * \brief capture a range and return a serialize function for it as a json array.
```

travisdowns:
@avikivity wrote:

> Do you have something similar? It seems high time that Seastar had its own concept of an asynchronous stream. I'd like to merge ideas from other repos.

No, we don't, though the need has been felt from time to time.

One "problem" with such interfaces is that the coroutine machinery is still quite heavy: every coroutine invocation costs about 2x as much as a memory alloc+free pair, so if you generate small elements and only suspend occasionally (e.g., you are generating integers or small strings from an input_stream, and _usually_ generate many for every time the stream suspends), you pay a high cost for this abstraction.

Coming at this from the other end, we do have a small set of "async algorithms", things like `for_each` but which take async functors, and internally they take care to be efficient, e.g., breaking the input range up into chunks which can are processed totally sync w/o any future/coro machinery, then doing a single suspension check etc. It doesn't really obviate the need for this type of generator, really, since many cases can't be transformed to use, but it does help with the "reactor stalls" case when processing large containers.

bhalevy:
@avikivity I took a stab at the "view" approach you mentioned, but it became quite ugly when considering the need to move the generated objects, so I used `std::unique_ptr<json_base>` as the generated value, like this:
```c++
diff --git a/include/seastar/json/json_elements.hh b/include/seastar/json/json_elements.hh
index ed4b6104..9af38c8e 100644
--- a/include/seastar/json/json_elements.hh
+++ b/include/seastar/json/json_elements.hh
@@ -252,6 +252,9 @@ struct json_base : public jsonable {
 
     json_base operator=(const json_base&) = delete;
 
+    json_base(json_base&&) noexcept = default;
+    json_base& operator=(json_base&&) noexcept = default;
+
     /**
      * create a formatted string of the object.
      * @return the object formatted.
@@ -367,6 +370,30 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
     };
 }
 
+template <class Generator>
+concept AsyncJsonGenerator = requires (Generator gen) {
+    { gen.get_json() } -> std::same_as<seastar::future<std::unique_ptr<json_base>>>;
+};
+
+/*!
+ * \brief consume values from a coroutine generator \c gen and encode the items using the \c fun mapping function
+ * onto the \c out output_stream as a json array.
+ */
+template <AsyncJsonGenerator Generator>
+future<> generate_json_array(output_stream<char>& out, Generator gen) {
+    bool first = true;
+    co_await out.write("[");
+    while (auto val = co_await gen.get_json()) {
+        if (first) {
+            first = false;
+        } else {
+            co_await out.write(", ");
+        }
+        co_await formatter::write(out, std::move(*val));
+    }
+    co_await out.write("]");
+}
+
 /*!
  * \brief capture an object and return a serialize function for it.
  *
diff --git a/tests/unit/json_formatter_test.cc b/tests/unit/json_formatter_test.cc
index 69c820e1..56c8ba3e 100644
--- a/tests/unit/json_formatter_test.cc
+++ b/tests/unit/json_formatter_test.cc
@@ -25,6 +25,8 @@
 #include <seastar/testing/test_case.hh>
 #include <seastar/core/vector-data-sink.hh>
+#include <seastar/coroutine/generator.hh>
 #include <seastar/json/formatter.hh>
 #include <seastar/json/json_elements.hh>
 #include <seastar/testing/thread_test_case.hh>
@@ -135,6 +137,32 @@ SEASTAR_THREAD_TEST_CASE(test_stream_range_as_array) {
     }, false);
 }
 
+SEASTAR_THREAD_TEST_CASE(test_generate_array) {
+    sstring expected = R"([{"subject":"1","values":[1]}, {"subject":"2","values":[2]}, {"subject":"3","values":[3]}])";
+    static auto generate_values = [] () -> coroutine::experimental::generator<int> {
+        for (int i : {1,2,3}) {
+            co_yield i;
+        }
+    };
+    struct generate_json {
+        coroutine::experimental::generator<int> gen = generate_values();
+
+        future<std::unique_ptr<json_base>> get_json() {
+            if (auto val = co_await gen()) {
+                auto obj = std::make_unique<object_json>();
+                obj->subject = std::to_string(*val);
+                obj->values.push(*val);
+                co_return std::move(obj);
+            } else {
+                co_return nullptr;
+            }
+        }
+    };
+    formatter_check_expected(expected, [&] (auto& out) {
+        generate_json_array(out, generate_json()).get();
+    }, true);
+}
+
 SEASTAR_THREAD_TEST_CASE(formatter_write) {
 
     formatter_check_expected("3", [] (auto &out) {
```

bhalevy:
A simpler approach could be:
```c++
diff --git a/include/seastar/json/json_elements.hh b/include/seastar/json/json_elements.hh
index ed4b6104..1dcf4719 100644
--- a/include/seastar/json/json_elements.hh
+++ b/include/seastar/json/json_elements.hh
@@ -367,6 +367,31 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
     };
 }
 
+template <class Generator>
+concept AsyncJsonGenerator = requires (Generator gen, output_stream<char> out) {
+    { gen.has_value() } -> std::same_as<future<bool>>;
+    { gen.write(out) } -> std::same_as<future<>>;
+};
+
+/*!
+ * \brief consume values from a coroutine generator \c gen and encode the items using the \c fun mapping function
+ * onto the \c out output_stream as a json array.
+ */
+template <AsyncJsonGenerator Gen>
+future<> generate_json_array(output_stream<char>& out, Gen& gen) {
+    bool first = true;
+    co_await out.write("[");
+    while (co_await gen.has_value()) {
+        if (first) {
+            first = false;
+        } else {
+            co_await out.write(", ");
+        }
+        co_await gen.write(out);
+    }
+    co_await out.write("]");
+}
+
 /*!
  * \brief capture an object and return a serialize function for it.
  *
diff --git a/tests/unit/json_formatter_test.cc b/tests/unit/json_formatter_test.cc
index 69c820e1..11e1d501 100644
--- a/tests/unit/json_formatter_test.cc
+++ b/tests/unit/json_formatter_test.cc
@@ -25,6 +25,8 @@
 #include <seastar/testing/test_case.hh>
 #include <seastar/core/vector-data-sink.hh>
+#include <seastar/coroutine/generator.hh>
 #include <seastar/json/formatter.hh>
 #include <seastar/json/json_elements.hh>
 #include <seastar/testing/thread_test_case.hh>
@@ -135,6 +137,35 @@ SEASTAR_THREAD_TEST_CASE(test_stream_range_as_array) {
     }, false);
 }
 
+SEASTAR_THREAD_TEST_CASE(test_generate_array) {
+    sstring expected = R"([{"subject":"1","values":[1]}, {"subject":"2","values":[2]}, {"subject":"3","values":[3]}])";
+    auto generate_values = [] () -> coroutine::experimental::generator<int> {
+        for (int i : {1,2,3}) {
+            co_yield i;
+        }
+    };
+    class generate_json_values {
+        coroutine::experimental::generator<int> _gen;
+        std::optional<int> _next;
+    public:
+        generate_json_values(coroutine::experimental::generator<int> gen) : _gen(std::move(gen)) {}
+        future<bool> has_value() {
+            _next = co_await _gen();
+            co_return _next.has_value();
+        }
+        future<> write(output_stream<char>& out) {
+            object_json obj;
+            obj.subject = std::to_string(*_next);
+            obj.values.push(*_next);
+            co_await json::formatter::write(out, obj);
+        }
+    };
+    formatter_check_expected(expected, [&] (auto& out) {
+        auto gen = generate_json_values(generate_values());
+        generate_json_array(out, gen).get();
+    }, true);
+}
+
 SEASTAR_THREAD_TEST_CASE(formatter_write) {
 
     formatter_check_expected("3", [] (auto &out) {
```

bhalevy:
> @avikivity wrote:
> 
> > Do you have something similar? It seems high time that Seastar had its own concept of an asynchronous stream. I'd like to merge ideas from other repos.
> 
> No, we don't, though the need has been felt from time to time.
> 
> One "problem" with such interfaces is that the coroutine machinery is still quite heavy: every coroutine invocation costs about 2x as much as a memory alloc+free pair, so if you generate small elements and only suspend occasionally (e.g., you are generating integers or small strings from an input_stream, and _usually_ generate many for every time the stream suspends), you pay a high cost for this abstraction.
> 
> Coming at this from the other end, we do have a small set of "async algorithms", things like `for_each` but which take async functors, and internally they take care to be efficient, e.g., breaking the input range up into chunks which can are processed totally sync w/o any future/coro machinery, then doing a single suspension check etc. It doesn't really obviate the need for this type of generator, really, since many cases can't be transformed to use, but it does help with the "reactor stalls" case when processing large containers.

Note that the coroutine generator can be used with a buffer of size greater than 1.
I believe it should save "context switching" between the producer and the consumer such that one can generate a number of values without yielding and on the other side of the wire co_await'ing the yielded values would yield based on maybe_yield().  This should reduce the context switch overhead between the consumer and producer coroutines.

I'll add a unit test for that in the next version as an example and to make sure the implementation is solid with that respect.

include/seastar/json/json_elements.hh
avikivity: Yes, I forgot the future there.

include/seastar/json/json_elements.hh
avikivity: But what's a serialize function?

avikivity:
Right, for small objects it shouldn't switch at all, and for large object one switch per task-quota.

travisdowns:
> Right, for small objects it shouldn't switch at all, and for large object one switch per task-quota.

Indeed, but even having the co_await for every item is an order or magnitude cost more than plain old sync stuff, so for small objects you pay a high price for the convenience. If you actually do task switch, that's another order of magnitude on top (but easy to avoid)

include/seastar/json/json_elements.hh
bhalevy: I agree the name is bad.
This is actually a view transformation function from the native type T to a jsonable object.

bhalevy: This function is used to "serialize" values as json onto an output stream.
It is typically used as the _body_writer of json::json_return_type
I can simply document it as such.

bhalevy:
In v2 ([b5119db](https://github.com/scylladb/seastar/commit/b5119db598d7202491fda779c30a4b511420f6b7)):
- Improved doc comments
- Added support for buffered generator

bhalevy:
In v3 ([1ea1719](https://github.com/scylladb/seastar/commit/1ea1719ae14a78b19b1315ee05ca4bb18ca02975)):
- call `generate_values_buffered<circular_buffer>` with explicit template parameter

bhalevy:
[f6971c0](https://github.com/scylladb/seastar/commit/f6971c0b9a4286f648f172d6ea88a43a90467ea0):
- rebased
- dropped default template parameter for the `generate_values_buffered` template (clang++ didn't like it)

bhalevy:
@avikivity, @scylladb/seastar-maint please consider merging

include/seastar/json/json_elements.hh
avikivity: The mapping is still there.

bhalevy: > The mapping is still there.

@avikivity as I mentioned [here](https://github.com/scylladb/seastar/pull/2805#issuecomment-2968032640), folding the mapping into the generated was problematic due to the design of the json base class.

avikivity:
> @avikivity, @scylladb/seastar-maint please consider merging

How? CI red

include/seastar/json/json_elements.hh
avikivity: I don't understand. Change your generator to co_yield fun(whatever) instead of co_yield whatever.

Easy to wrap generators too.

```c++
generator<whatever> wrapped(generator<another> g) {
    while (auto v = g()) {
          co_yield fun(std::move(*g));
    }
}
```

bhalevy: It's simple on paper but the optional result needs to be moved too so it needs to hold a std::unique_ptr<json_base>

bhalevy:
> > @avikivity, @scylladb/seastar-maint please consider merging
> 
> How? CI red

That's odd.  It was green when I asked above.

include/seastar/json/json_elements.hh
avikivity: How can fun return json_base if the generator can't?

bhalevy: > How can fun return json_base if the generator can't?

It doesn't.  It returns an object derived from json_base that isn't moved further, but rather "serialized" onto the json stream.

I think I have a solution by allowing the implementing a noexcept move constructor for the test's object_json
and defining the generator as returning object_json with the right contraints.

I'll post a new version

bhalevy:
In [1ec981a](https://github.com/scylladb/seastar/commit/1ec981a2ac274ce5478eed374c0ea40767d1b7c4):
- Let the json generator co_yield Jsonable objects

include/seastar/json/json_elements.hh
bhalevy: Done in [1ec981a](https://github.com/scylladb/seastar/commit/1ec981a2ac274ce5478eed374c0ea40767d1b7c4)

include/seastar/json/json_elements.hh
```diff
@@ -232,6 +236,12 @@ public:
     }
 };
 
+template <typename T>
+concept Jsonable = requires (T j) {
+    { j.to_json() } -> std::convertible_to<std::string>;
```
avikivity: Isn't sstring okay? Perhaps convertible_to string_view.

include/seastar/json/json_elements.hh
```diff
@@ -232,6 +236,12 @@ public:
...
+    { j.write(std::declval<output_stream<char>&>()) } -> std::same_as<future<>>;
```
avikivity: Nicer to declare the stream in the requires () parameter list.

include/seastar/json/json_elements.hh
```diff
@@ -367,6 +377,47 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
     };
 }
 
+/*!
+ * \brief consume values from a coroutine generator \c gen and encode the items using the \c fun mapping function
+ * onto the \c out output_stream as a json array.
+ */
+template<Jsonable T>
+future<> generate_array_element(output_stream<char>& out, coroutine::experimental::generator<T>& gen) {
```
avikivity: I think we should accept it by value since we consume it.

include/seastar/json/json_elements.hh
```diff
@@ -367,6 +377,47 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
...
+ * To use it, pass a reference to coroutine::generator \c gen and a mapping function \c fun.
+ * For example, if ``gen`` is a coroutine::generator<std::pair<K, V>> and you want to return the keys as json array
+ *
+ * return make_ready_future<json::json_return_type>(generate_array(gen, [](const auto&i) {return i.first}));
+ *
+ * Note that \c gen is passed by reference since we need to return a copyable function but generators cannot be copied.
+ * So the caller is responsible for ensuring that the generator remains valid for the lifetime of the returned function.
+ */
+template<Jsonable T>
+std::function<future<>(output_stream<char>&&)> generate_array(coroutine::experimental::generator<T>& gen) {
```
avikivity: Again, by value.

bhalevy:
It looks like the build issue with clang happens with clang version 18.1.3
but is fixed in later version (19.1.7) as in scylla dbuild environment or 20.1.7 on Fedora 42.

include/seastar/json/json_elements.hh
bhalevy: It would be better indeed, but as I explained in the commit message:
```
 * Note that \c gen is passed by reference since we need to return a copyable function but generators cannot be copied.
```

bhalevy: Also, the generator not consumed by generate_array_element.
The latter is called in an async loop by the function returned by `generate_array`:
```c++
    return [&gen] (output_stream<char>&& s) mutable -> future<> {
        auto out = std::move(s);
        auto f = co_await coroutine::as_future(generate_array_element(out, gen));
        co_await out.close();
        f.get();
    };
```

bhalevy:
In [17b430](https://github.com/scylladb/seastar/commit/17b4307b733e5be0bdbabb364e320005327914ae):
```diff
+template<typename T>
+using buffered_container = circular_buffer<T>;
```
To fix the build issue with older clang version

In [c8448d8](https://github.com/scylladb/seastar/commit/c8448d845ac12e522a9f51710fb09f7dca5127d2):
```diff
 template <typename T>
-concept Jsonable = requires (T j) {
-    { j.to_json() } -> std::convertible_to<std::string>;
-    { j.write(std::declval<output_stream<char>&>()) } -> std::same_as<future<>>;
+concept Jsonable = requires (T j, output_stream<char>& s) {
+    { j.to_json() } -> std::convertible_to<std::string_view>;
+    { j.write(s) } -> std::same_as<future<>>;
};
```

include/seastar/json/json_elements.hh
bhalevy: done in [c8448d8](https://github.com/scylladb/seastar/commit/c8448d845ac12e522a9f51710fb09f7dca5127d2)

include/seastar/json/json_elements.hh
bhalevy: done in [c8448d8](https://github.com/scylladb/seastar/commit/c8448d845ac12e522a9f51710fb09f7dca5127d2)

include/seastar/json/json_elements.hh
bhalevy: See https://github.com/scylladb/seastar/pull/2805#discussion_r2183137775

avikivity:
> It looks like the build issue with clang happens with clang version 18.1.3 but is fixed in later version (19.1.7) as in scylla dbuild environment or 20.1.7 on Fedora 42.

I'll regenerate the toolchain to include clangs 19 and 20.

include/seastar/json/json_elements.hh
avikivity: Can we return noncopyable_function?

Maybe precluded by another requirement, too bad.

bhalevy:
@avikivity please re-review and consider merging

include/seastar/json/json_elements.hh
nyh: This is a very important point. In https://github.com/scylladb/scylladb/pull/24480 I had ScyllaDB Alternator code which used Seastar's "json_elements" and I had to get rid of it because, among other things, its use of copyable std::function.

avikivity:
```c++
struct json_return_type {
    sstring _res;
    std::function<future<>(output_stream<char>&&)> _body_writer;
    json_return_type(std::function<future<>(output_stream<char>&&)>&& body_writer) : _body_writer(std::move(body_writer)) {
    }
```

Let's do this:
1. Add a type alias for `std::function<future<> (output_stream<char>&&)>`
2. Add a new API_LEVEL where this type alias is a noncopyable_function

avikivity: To reduce future churn, place this in the internal subnamespace.

include/seastar/json/json_elements.hh
```diff
@@ -367,6 +377,47 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
     };
 }
 
+/*!
+ * \brief consume values from a coroutine generator \c gen and encode the items using the \c fun mapping function
```
avikivity: There's no fun any more.

include/seastar/json/json_elements.hh
```diff
@@ -367,6 +377,47 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
...
+            co_await out.write(", ");
+        }
+        co_await formatter::write(out, *val);
+    }
+    co_await out.write("]");
+}
+
+/*!
+ * \brief return a json body_writer function that consumes values from a coroutine generator \c gen
+ * and transforms them into a json array using a transformation function \c fun
```
avikivity: There's no fun anymore.

include/seastar/json/json_elements.hh
```diff
@@ -367,6 +377,47 @@ std::function<future<>(output_stream<char>&&)> stream_range_as_array(Container v
...
+ * For example, if ``gen`` is a coroutine::generator<std::pair<K, V>> and you want to return the keys as json array
+ *
+ * return make_ready_future<json::json_return_type>(generate_array(gen, [](const auto&i) {return i.first}));
+ *
+ * Note that \c gen is passed by reference since we need to return a copyable function but generators cannot be copied.
+ * So the caller is responsible for ensuring that the generator remains valid for the lifetime of the returned function.
+ */
+template<Jsonable T>
+std::function<future<>(output_stream<char>&&)> generate_array(coroutine::experimental::generator<T>& gen) {
+    return [&gen] (output_stream<char>&& s) mutable -> future<> {
```
avikivity: Don't we need coroutine::lambda here?

bhalevy: I don't think so because we return a std::function that keeps the lambda alive

avikivity: What if it's copied?

bhalevy: Good point

bhalevy: Indeed there is a general issue with copying std::function holding a lambda coroutine that I reproduced, but having the std::function hold a coroutine::lambda doesn't help either.
The coroutine::lambda is required when generating the continuation task that would invoke the std::function eventually.

At any rate, it seems like it doesn't make a difference here since gen is captured by reference in the lambda function.

For completeness, here is a bad example with stack-use-after-return:
```
    double n = dist(rand_eng);
    auto make_fun = [n] () -> std::function<future<double>()> {
        return [n] () -> future<double> {
            // Expect coroutine capture to be freed after co_await without coroutine::lambda
            co_await yield();
            co_return n;
        };
    };
    auto fun = make_fun();
    auto n2 = co_await yield().then([fun] {
        return fun();
    });
```

Bad example 2:
```
    double n = dist(rand_eng);
    auto make_fun = [n] () -> std::function<future<double>()> {
        return coroutine::lambda([n] () -> future<double> {
            // Expect coroutine capture to be freed after co_await without coroutine::lambda
            co_await yield();
            co_return n;
        });
    };
    auto fun = make_fun();
    auto n2 = co_await yield().then([fun] {
        return fun();
    });
```

Good example 1:
```
    auto make_fun = [n] () -> std::function<future<double>()> {
        return [n] () -> future<double> {
            // Expect coroutine capture to be freed after co_await without coroutine::lambda
            co_await yield();
            co_return n;
        };
    };
    auto fun = make_fun();
    auto n2 = co_await yield().then([&fun] {
        return fun();
    });
```

Good example 2:
```
    double n = dist(rand_eng);
    auto make_fun = [n] () -> std::function<future<double>()> {
        return [n] () -> future<double> {
            // Expect coroutine capture to be freed after co_await without coroutine::lambda
            co_await yield();
            co_return n;
        };
    };
    auto fun = make_fun();
    auto n2 = co_await yield().then(coroutine::lambda([fun] {
        return fun();
    }));
```

include/seastar/json/json_elements.hh
bhalevy: @avikivity do you think the api change should be part of this series or as a follow up?

Here, the returned function captures the generator by reference and the generator is kept alive by the caller so it's ok to return the coroutine lambda as a copyable function.

bhalevy:
In [8dcc2c3](https://github.com/scylladb/seastar/commit/8dcc2c3cf0008fc3e082faff4bd2e02fe2385f65):
- Rebased
- Fixed documentation comments
- Modified added unit tests to test copying of the function returned by `generate_array`

bhalevy:
@avikivity ping, please re-review and consider merging

avikivity:
I don't understand why the lambda coroutine doesn't explode.

bhalevy:
> I don't understand why the lambda coroutine doesn't explode.

I think that since `gen` is captured by reference, it is not lost when the lambda returns on `co_await`, since the compiler refers to the calling function's frame to get to `gen` rather than keeping a variable with `&gen`'s value in the lambda function captures.

include/seastar/json/json_elements.hh
avikivity: It could be a follow up.

avikivity: However, do place internal functions in an internal subnamespace.

bhalevy:
In [63dc17f](https://github.com/scylladb/seastar/commit/63dc17f5d4bc2bbffd7cc79aa9ee6ef04945630b):
- Refactored patches for better clarity and bisectibility
- util: noncopyable_function: enable_lw_shared_from_this
  - to help make `http::shared_body_writer_type = lw_shared_ptr<noncopyable_function<future<>(output_stream<char>&&)>>`
- pass the the coroutine generator by value to `generate_array` and capture it by value now that the returned coroutine lambda function is kept as a shared noncopyable function

include/seastar/util/noncopyable_function.hh
```diff
@@ -104,7 +105,7 @@ struct is_nothrow_if_object<> {
 /// of the contained function.
 SEASTAR_MODULE_EXPORT
 template <typename Ret, typename... Args, bool Noexcept>
-class noncopyable_function<Ret (Args...) noexcept(Noexcept)> : private internal::noncopyable_function_base {
+class noncopyable_function<Ret (Args...) noexcept(Noexcept)> : private internal::noncopyable_function_base, public enable_lw_shared_from_this<noncopyable_function<Ret (Args...) noexcept(Noexcept)>> {
```
avikivity: That's uncalled for. Most noncopyable_function users don't want to pay for the extra storage.

It also prevents replacing it by std::move_only_function, eventually.

bhalevy: I was under the assumption that the footprint of noncopyable_function isn't critical, but I understand if we want to keep it as economical as possible.
But we'll need to allocate in the `write_body(body_writer_type&&)` entry point to move the function into a `lw_shared_ptr<body_writer_type>`.

avikivity: A common low-level utility should not assume anything about its callers.

bhalevy:
[fa7f581](https://github.com/scylladb/seastar/commit/fa7f5815d8a31e808a3129b15312717071bdf551):
- rebased
- dropped "util: noncopyable_function: enable_lw_shared_from_this"

include/seastar/util/noncopyable_function.hh
bhalevy: Dropped in [fa7f581](https://github.com/scylladb/seastar/commit/fa7f5815d8a31e808a3129b15312717071bdf551)

nyh:
> Add a helper function to generate a json array gently by consuming its items from a coroutine generator.

Please rewrite the cover letter. Especially, explain why you modified the API (and introduced "level 8") and what it means for backward-compatibility: Will existing code that returned std::function need to be rewritten - or will temporarily (?) both old and new code will work?

As I noted before, in ScyllaDB Alternator I was using these json_return_type thing until I was annoyed that it used the std::function - which forced us to artificially write ugly code that supported copying, so in https://github.com/scylladb/scylladb/pull/24480/commits/d8fab2a01a459b01e88b16c8d096268074707166 I just stopped using Seastar's json_return_type entirely. I happily discovered that in fact json_return_type doesn't need to be used at all to write an HTTP server. In fact I started to feel that putting all that "json" stuff in Seastar was a mistake - it's not really a full solution for applications that really need to dabble seriously in JSON, and when anybody starts to use it more seriously the flaws become obvious and you need to either start rewriting pieces (as you did here) or stop using it (as I did).  More fundamentally, there is nothing in an HTTP server that inherently scream out JSON. HTTP came out years before JSON did, and nothing in the standard recommends that requests or replies be in JSON format - any more than requests or replies should be in XML or Microsoft Office formats.

Perhaps your "rewriting" approach is indeed better in the sense if Seastar already has these "json" things, for one reason or another, they'll at least be more usable in the future, out of the box. But I'm not sure that you didn't just plug one hole that bothered you, leaving a dozen other holes. I guess time will tell.

bhalevy:
> > Add a helper function to generate a json array gently by consuming its items from a coroutine generator.
> 
> Please rewrite the cover letter. Especially, explain why you modified the API (and introduced "level 8") and what it means for backward-compatibility: Will existing code that returned std::function need to be rewritten - or will temporarily (?) both old and new code will work?

done.

> 
> As I noted before, in ScyllaDB Alternator I was using these json_return_type thing until I was annoyed that it used the std::function - which forced us to artificially write ugly code that supported copying, so in [scylladb/scylladb@d8fab2a](https://github.com/scylladb/scylladb/commit/d8fab2a01a459b01e88b16c8d096268074707166) I just stopped using Seastar's json_return_type entirely. I happily discovered that in fact json_return_type doesn't need to be used at all to write an HTTP server. In fact I started to feel that putting all that "json" stuff in Seastar was a mistake - it's not really a full solution for applications that really need to dabble seriously in JSON, and when anybody starts to use it more seriously the flaws become obvious and you need to either start rewriting pieces (as you did here) or stop using it (as I did). More fundamentally, there is nothing in an HTTP server that inherently scream out JSON. HTTP came out years before JSON did, and nothing in the standard recommends that requests or replies be in JSON format - any more than requests or replies should be in XML or Microsoft Office formats.

I tend to agree.  It seems out of scope for seastar, that should be focused on the OS-interaction level, (cpu, disk, net) scheduling , memory management, async synchronization primitives, and such.

> 
> Perhaps your "rewriting" approach is indeed better in the sense if Seastar already has these "json" things, for one reason or another, they'll at least be more usable in the future, out of the box. But I'm not sure that you didn't just plug one hole that bothered you, leaving a dozen other holes. I guess time will tell.

bhalevy:
@nyh please re-review

include/seastar/http/reply.hh
```diff
@@ -245,7 +245,11 @@ struct reply {
      *
      */
 
-    void write_body(const sstring& content_type, noncopyable_function<future<>(output_stream<char>&&)>&& body_writer);
+    void write_body(const sstring& content_type, http::shared_body_writer_type body_writer);
+
+    void write_body(const sstring& content_type, http::body_writer_type&& body_writer) {
```
avikivity: Suppose I call here with a lambda. How does it choose which signature? Maybe is chooses the shared_body_writer_type signature because it's not a reference.

Why have two signatures at all? There's no reason to support two. Support one based on API level.

include/seastar/http/reply.hh
```diff
@@ -282,7 +286,7 @@ private:
     future<> write_reply_to_connection(httpd::connection& con);
     future<> write_reply_headers(httpd::connection& connection);
 
-    noncopyable_function<future<>(output_stream<char>&&)> _body_writer;
+    http::shared_body_writer_type _body_writer;
```
avikivity: Should be the nonshared type (or the API-selected type).

CMakeLists.txt
```diff
@@ -70,10 +70,10 @@ option (Seastar_DEPRECATED_OSTREAM_FORMATTERS
   ON)
 
 set (Seastar_API_LEVEL
-  "7"
+  "8"
```
avikivity, line 73: > json: API level 8: use noncopyable_function in json_return_type
> Hold a shared_body_writer_type so we can copy json_return_type
> without copying the body_writer function.
> 
> This will allow returning a co_routine lambda as body_writer
> that can be safely held as a lw_shard_ptr<noncopyabe_function>
> in the following patch.
> 

Why does json_return_type need to be copyable?

include/seastar/json/json_elements.hh
```diff
@@ -315,10 +316,20 @@ struct json_void : public jsonable{
...
+    http::shared_body_writer_type _body_writer;
+    json_return_type(body_writer_type&& body_writer) : _body_writer(http::make_shared_body_writer_type(std::move(body_writer))) {
+    }
+#else
     using body_writer_type = std::function<future<>(output_stream<char>&&)>;
     body_writer_type _body_writer;
     json_return_type(body_writer_type&& body_writer) : _body_writer(std::move(body_writer)) {
     }
+#endif
+
```
avikivity: I don't understand why you have one patch supporting both signatures, and then splitting again. It's just confusing.

And we shouldn't need to make he noncopyable function copyable by wrapping it. Why is json_return_type copied?

include/seastar/http/reply.hh
bhalevy: One is a shared ptr the other is passed as rvalue reference meaning the callee assumes ownership of the object, while in the shared ptr case the writer is held by the caller.

include/seastar/json/json_elements.hh
bhalevy: It's just that it has a copy constructor and assignment operator, and there's a unit test that takes advantage of it in httpd_test.test_shared_future.

But if the json_return_type would hold a noncopyable_function there is still a problem with the request and reply write_body functions that take a noncopyable_function rvalue reference where the write is consumed by the write_body function. Today those are "forked" off a corresponding std::function.

This is not easy to change, even with a new API level because now the problem is that the reply object is returned e.g. from the httpd handler_base::handle method, so if the handler would need to make sure the body writer noncopyable function is kept alive until all reply objects that reference it are gone, but currently there isn't a good way to track them (hence I preffered to pass a shared_ptr<noncopyable_function> in the current version, so the reply or request object can own the shared pointer and free the reference when destroyed.

To do that properly with a new API level I'd consider keeping a reference to the handler in the request and reply objects and enter a gate to facilitate a handler::stop() method that can make sure all references to the noncopyable body_writer are gone before the handler itself and its body_writer function are destroyed.

include/seastar/http/reply.hh
avikivity: It's not good practice to have such APIs where the ownership is unclear.

Why not pass the ownership to the callee? The caller can always fake it with std::ref().

include/seastar/json/json_elements.hh
avikivity: > It's just that it has a copy constructor and assignment operator, and there's a unit test that takes advantage of it in httpd_test.test_shared_future.

We could make it part of API level 8 that it's noncopyable.


> 
> But if the json_return_type would hold a noncopyable_function there is still a problem with the request and reply write_body functions that take a noncopyable_function rvalue reference where the write is consumed by the write_body function. Today those are "forked" off a corresponding std::function.

I don't understand, can you give an example?

> 
> This is not easy to change, even with a new API level because now the problem is that the reply object is returned e.g. from the httpd handler_base::handle method, so if the handler would need to make sure the body writer noncopyable function is kept alive until all reply objects that reference it are gone, but currently there isn't a good way to track them (hence I preffered to pass a shared_ptr<noncopyable_function> in the current version, so the reply or request object can own the shared pointer and free the reference when destroyed.
> 
> To do that properly with a new API level I'd consider keeping a reference to the handler in the request and reply objects and enter a gate to facilitate a handler::stop() method that can make sure all references to the noncopyable body_writer are gone before the handler itself and its body_writer function are destroyed.

No, this is descending into a spiral of ever-increasing complexity.

json_return_type encapsulate an object that can be as small as an int, and here we are with gates and shared pointers to noncopyable functions and confusion.

Better to start from scratch. This is an API, the users should be able to use it without having to decide so many things.

bhalevy: We did open a can of worms...
I'll take a stab at it and push a simplified api with the new api level that will allow the callee to hold a noncopyable body writer function and let request and reply use it by reference.
My main worry there is lifecycle management to ensure the function is kept alive while requests and replies using it may use it.

avikivity: The function's lifetime should be fully enclosed in the json_return_type (while at it we can use a better name).

nyh: > json_return_type encapsulate an object that can be as small as an int, and here we are with gates and shared pointers to noncopyable functions and confusion.
> 
> Better to start from scratch. This is an API, the users should be able to use it without having to decide so many things.

By the way, this is what I ended up doing in some Scylla Alternator code that was using json_return_type, and I realized the existing json_return_type is simply not good enough - I wanted it to use a non-copyable function, and couldn't. So I discovered that although Seastar supports json_return_type in some things, I didn't actually need to use it for anything, everything works without it and in fact everything became simpler by not using it :-) See https://github.com/scylladb/scylladb/pull/24480/commits/d8fab2a01a459b01e88b16c8d096268074707166.

avikivity: That's also a route we should consider.

bhalevy: Simplified version pushed in [fa00799](https://github.com/scylladb/seastar/commit/fa00799b8ddabe6ebb28bbffa200154ae6c3a980)

bhalevy:
In [fa00799](https://github.com/scylladb/seastar/commit/fa00799b8ddabe6ebb28bbffa200154ae6c3a980):
- rebased
- simplified model where only `body_writer_type = noncopyable_function<future<>(output_stream<char>&&)>` is used
  - With that json_return_type became noncopyable
  - Disabling `httpd_test.test_shared_future`
  - Adjusted `httpd_test.json_test_handler` to call `rep->write_body` with a lambda calling the noncopyable `http::body_writer_type _write_func` member, instead of implicitly generating a noncopyable_function from a std::function member

avikivity:
> In [fa00799](https://github.com/scylladb/seastar/commit/fa00799b8ddabe6ebb28bbffa200154ae6c3a980):
> 
> * rebased
> * simplified model where only `body_writer_type = noncopyable_function<future<>(output_stream<char>&&)>` is used
>   
>   * With that json_return_type became noncopyable
>   * Disabling `httpd_test.test_shared_future`
>   * Adjusted `httpd_test.json_test_handler` to call `rep->write_body` with a lambda calling the noncopyable `http::body_writer_type _write_func` member, instead of implicitly generating a noncopyable_function from a std::function member

@bhalevy let's split off the json_return_type changes, they're a lot more involved than generate_array.

bhalevy:
> > In [fa00799](https://github.com/scylladb/seastar/commit/fa00799b8ddabe6ebb28bbffa200154ae6c3a980):
> > 
> > * rebased
> > * simplified model where only `body_writer_type = noncopyable_function<future<>(output_stream<char>&&)>` is used
> >   
> >   * With that json_return_type became noncopyable
> >   * Disabling `httpd_test.test_shared_future`
> >   * Adjusted `httpd_test.json_test_handler` to call `rep->write_body` with a lambda calling the noncopyable `http::body_writer_type _write_func` member, instead of implicitly generating a noncopyable_function from a std::function member
> 
> @bhalevy let's split off the json_return_type changes, they're a lot more involved than generate_array.

@avikivity made this PR a draft until https://github.com/scylladb/seastar/pull/2921 that focuses on the json_return_type changes.

bhalevy:
[7d04a2c](https://github.com/scylladb/seastar/commit/7d04a2cbebd896840a1ab1f8b563bc4d21b6c25e): rebased on top of dac7c6de

avikivity:
I'm sure the cover letter wants updating after the large simplifications.

bhalevy:
> I'm sure the cover letter wants updating after the large simplifications.

True.  Updated cover letter.

include/seastar/json/json_elements.hh
avikivity: Still needs to be in an internal namespace.

tests/unit/json_formatter_test.cc
```diff
@@ -135,6 +141,30 @@ SEASTAR_THREAD_TEST_CASE(test_stream_range_as_array) {
...
+    };
+
+    // Copy the streamer function on purpose to ensure it doesn't cause
+    // use-after-stack-return.
+    auto streamer = generate_array(generate_json_values());
+
+    formatter_check_expected(expected, [streamer = std::move(streamer)] (auto& out) {
+        streamer(std::move(out)).get();
+    }, false);
+}
```
avikivity, line 165: Can you add an example showing how such an array is embedded in a containing object?

bhalevy: There is the `expected` string, do you want an example of a higher-level use case?

bhalevy: When returning a `body_writer_type` (for constructing a `json_return_type`), typically the array is the only member of the containing object.

When embedding an array in a higher level json object, I'd use the internal function: `generate_array_element`, that generates only the json array element onto the output stream (without closing it), and then it could be used e.g. to generate one or more arrays in a larger json container.

avikivity: If it's an internal function, it shouldn't be used.

I'm looking for an example that generates something like this:

```json
{
   "a": 4,
   "b": 7,
   "c": [some long array represented by a generator],
   "d": 3
}
```

bhalevy: > If it's an internal function, it shouldn't be used.

So maybe it shouldn't be internal then.

> I'm looking for an example that generates something like this:
> 
> ```json
> {
>    "a": 4,
>    "b": 7,
>    "c": [some long array represented by a generator],
>    "d": 3
> }
> ```

I'm not sure how to do it using the existing api since the returned `body_writer_type` consumes the output stream and closes it.  It is meant to serialize a singular json result, not an element within a larger json object.

`generate_array_element` that you asked to be in the `seastar::internal` namespace does that since it outputs a json array onto the output stream and it doesn't close the stream, so it can be called to compose one json array element as part of a higher level json object.

avikivity: Isn't it very limited? Sure there will be some cases where we want to return a large array as the outer object, but aren't most jsons top-level objects (not arrays)?

bhalevy: For scylla API this is sufficient.

One of the elaborate use cases that I'm aware of is describe_ring:
```c++
static future<json::json_return_type> describe_ring_as_json(sharded<service::storage_service>& ss, sstring keyspace) {
    co_return json::json_return_type(stream_range_as_array(co_await ss.local().describe_ring(keyspace), token_range_endpoints_to_json));
}
```
where `describe_ring` returns `utils::chunked_vector<storage_service::token_range_endpoints>`
where:
```c++
struct token_range_endpoints {
    sstring _start_token;
    sstring _end_token;
    std::vector<sstring> _endpoints;
    std::vector<sstring> _rpc_endpoints;
    std::vector<endpoint_details> _endpoint_details;
};
```
and `token_range_endpoints_to_json` returns the interesting json object:
```c++
struct token_range : public json::json_base {
    json::json_element< sstring > start_token;
    json::json_element< sstring > end_token;
    json::json_list< sstring > endpoints;
    json::json_list< sstring > rpc_endpoints;
    json::json_list< endpoint_detail > endpoint_details;
```
using:
```c++
static ss::token_range token_range_endpoints_to_json(const dht::token_range_endpoints& d) {
    ss::token_range r;
    r.start_token = d._start_token;
    r.end_token = d._end_token;
    r.endpoints = d._endpoints;
    r.rpc_endpoints = d._rpc_endpoints;
    for (auto det : d._endpoint_details) {
        ss::endpoint_detail ed;
        ed.host = fmt::to_string(det._host);
        ed.datacenter = det._datacenter;
        if (det._rack != "") {
            ed.rack = det._rack;
        }
        r.endpoint_details.push(ed);
    }
    return r;
}
```

Note that `r.endpoint_details` is constructed dynamically, but it is fully materialized.
Usually it has a short enough list so it doesn't matter much, but if it to generate a list of undetermined, and potentially long size, it could use a way for nested streaming of data.

Not sure about Alternator use case.
@nyh can you please elaborate?

avikivity: Can't we have a json object that contains the generated array as one of its elements?

Now that I think of it, why does it have to be a generator? It can be a range, and since std::generator is a range, it just works.

bhalevy: > Can't we have a json object that contains the generated array as one of its elements?
> 
> Now that I think of it, why does it have to be a generator? It can be a range, and since std::generator is a range, it just works.

This change is about an async generator which isn't a range unfortunately (it was nice to have something like that)

bhalevy: @avikivity can we proceed with this PR as it is?

avikivity: It seems like a very partial solution.

It feels like we're piling on bandaids instead of making a working whole.

bhalevy: I'm not sure if this interface even belongs in seastar, but we can make the change more radical and deprecate or change the json_return_type body_writer interface to not consume the output stream, making it more composable and so then encoding different members of a structured json object easier.

avikivity: Good question. On one hand json is ubiquitous and we shouldn't require every application to roll out its own. On the other hand a lazy json type is very complicated and therefore hard to nail down in a library that promises source compatibility.

nyh: Indeed, as I commented in https://github.com/scylladb/seastar/pull/2805#issuecomment-3101274892, in Scylla Alternator we discovered that the json_return_type is simply not good enough for real use, and more importantly - it's not actually needed, so we stopped using it completely and just used our own thing. I wouldn't shed tears if it's deprecated (but note that I only looked in Alternator - I don't know if other Scylla source code uses it, and certainly don't know how Seastar users use it or not use it).

bhalevy: We have some simple use cases at the scylla rest api layer which could use a local "library".
And another planned use case for generating a manifest.json file for snapshots, which is a simple array of file names, and another possible file holding a list of `{ sstable_id, min_token, max_token }`

bhalevy: Hmm, and another use case in utils/config_file config_type::to_json

Issue #2803, opened 2025-06-10
Move (partially) http client retry strategy from scylla to seastar


xemul:
Http client may restart requests if the failure exception is because of connection closure. However, eventually it turned out that restarting requests has much more reasons, e.g. scylla S3 client has complicated retryable_client logic around http client.

It's good to have abstract retry_strategy in seastar client, so that retry logic is in seastar and retry reasoning is up to the user

Cc @kreuzerkrieg

kreuzerkrieg:
Something like this?

seastar RFC: https://github.com/scylladb/seastar/pull/2806
scylladb RFC: https://github.com/scylladb/scylladb/pull/24498

xemul:
Yes, but currently http client retries only once, while with retry-strategy it should retry in a loop (presumably I mis-read the wip code though)

kreuzerkrieg:
> Yes, but currently http client retries only once, while with retry-strategy it should retry in a loop (presumably I mis-read the wip code though)

it is done on purpose to preserve the existing behaviour, do we want to retry in loop with max retries and exponential backoff? the same way we do for aws errors?

xemul:
Yes, new retry strategy approach should allow for infinite amount of retries. It's up to strategy to decide whether to retry or not, the caller (client make_request code) should just do what it's told to

PR #2806, opened 2025-06-12
http: Introduce retry strategy machinery for http client

kreuzerkrieg:
Introduce extensible retry strategy machinery for the HTTP client to simplify adding new retry conditions. Rather than modifying seastar code to support additional error types, users can now provide their own custom implementations of the retry strategy—offering flexibility without repeated code changes.

Fixes: https://github.com/scylladb/seastar/issues/2803

### DIFF
--- a/include/seastar/http/retry_strategy.hh
+++ b/include/seastar/http/retry_strategy.hh
@@ -1,0 +1,60 @@
+/*
+ * Copyright (C) 2025-present ScyllaDB
+ */
+
+/*
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
+ */
+
+#pragma once
+#include <seastar/core/future.hh>
+#include <seastar/http/reply.hh>
+#include <seastar/util/modules.hh>
+
+namespace seastar {
+SEASTAR_MODULE_EXPORT_BEGIN
+
+namespace http::experimental {
+
+class retry_strategy {
+public:
+    virtual ~retry_strategy() = default;
+    // Returns true if the error can be retried given the error and the number of times already tried.
+    virtual future<bool> should_retry(std::exception_ptr error, unsigned attempted_retries) const = 0;
+    virtual future<input_stream<char>> analyze_reply(std::optional<reply::status_type> expected, const http::reply& rep, input_stream<char>&& in) const = 0;
+
+    // Calculates the time in milliseconds the client should wait before attempting another request based on the error and attempted_retries count.
+    [[nodiscard]] virtual std::chrono::milliseconds delay_before_retry(std::exception_ptr error, unsigned attempted_retries) const = 0;
+
+    [[nodiscard]] virtual unsigned get_max_retries() const noexcept = 0;
+};
+
+class default_retry_strategy : public retry_strategy {
+private:
+    unsigned _max_retries;
+    unsigned _scale_factor;
+
+public:
+    default_retry_strategy();
+    default_retry_strategy(unsigned max_retries, unsigned scale_factor);
+
+    future<bool> should_retry(std::exception_ptr error, unsigned attempted_retries) const override;
+    future<input_stream<char>> analyze_reply(std::optional<reply::status_type> expected, const http::reply& rep, input_stream<char>&& in) const override;
+
+    [[nodiscard]] std::chrono::milliseconds delay_before_retry(std::exception_ptr error, unsigned attempted_retries) const override;
+
+    [[nodiscard]] unsigned get_max_retries() const noexcept override;
+};
+
+class no_retry_strategy : public default_retry_strategy {
+
+public:
+    future<bool> should_retry(std::exception_ptr error, unsigned attempted_retries) const override;
+    [[nodiscard]] std::chrono::milliseconds delay_before_retry(std::exception_ptr error, unsigned attempted_retries) const override;
+
+    [[nodiscard]] unsigned get_max_retries() const noexcept override;
+};
+}
+
+SEASTAR_MODULE_EXPORT_END
+}
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -393,21 +393,11 @@ future<> client::do_make_request(connection& con, request& req, reply_handler& h
     auto sub = as ? as->subscribe([&con] () noexcept { con.shutdown(); }) : std::nullopt;
     return con.do_make_request(req).then([this, &con, &req, &handle, expected] (connection::reply_ptr reply) mutable {
         auto& rep = *reply;
-        if (expected.has_value() && rep._status != expected.value()) {
-            if (!http_log.is_enabled(log_level::debug)) {
-                return make_exception_future<>(httpd::unexpected_status_error(rep._status));
-            }
-
-            return do_with(con.in(rep), [reply = std::move(reply)] (auto& in) mutable {
-                return util::read_entire_stream_contiguous(in).then([reply = std::move(reply)] (auto message) {
-                    http_log.debug("request finished with {}: {}", reply->_status, message);
-                    return make_exception_future<>(httpd::unexpected_status_error(reply->_status));
-                });
-            });
-        }
-
         auto in = req._method != "HEAD" ? con.in(rep) : input_stream<char>(data_source(std::make_unique<skip_body_source>(rep)));
-        return handle(rep, std::move(in)).then([this, reply = std::move(reply), &con] {
+
+        return _retry_strategy->analyze_reply(expected, rep, std::move(in)).then([this, &handle, &con, reply = std::move(reply)](auto _in) mutable {
+            auto in = std::move(_in);
+        return handle(*reply, std::move(in)).then([this, reply = std::move(reply), &con] {
             if (reply->content_length > reply->consumed_content) {
                 auto bytes_left = reply->content_length - reply->consumed_content;
                 if (bytes_left <= _max_bytes_to_drain) {
@@ -420,6 +410,7 @@ future<> client::do_make_request(connection& con, request& req, reply_handler& h
                 con._persistent = false;
             }
             return make_ready_future<>();
+            });
         });
     }).handle_exception([&con] (auto ex) mutable {
         con._persistent = false;
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -330,24 +330,33 @@ future<> client::make_request(request&& req, reply_handler&& handle, std::option
     });
 }
 
+future<> client::make_request(request&& req, reply_handler&& handle, const retry_strategy& strategy, std::optional<reply::status_type>&& expected, abort_source* as) {
+    return do_with(std::move(req), std::move(handle), [this, &strategy, expected, as](request& req, reply_handler& handle) mutable {
+        return make_request(req, handle, strategy, std::move(expected), as);
+    });
+}
+
 future<> client::make_request(request& req, reply_handler& handle, std::optional<reply::status_type> expected, abort_source* as) {
-    return with_connection([this, &req, &handle, as, expected] (connection& con) {
-        return do_make_request(con, req, handle, as, expected);
-    }, as).handle_exception([this, &req, &handle, as, expected] (std::exception_ptr ex) {
+    return make_request(req, handle, *_retry_strategy, expected, as);
+}
+
+future<> client::make_request(request& req, reply_handler& handle, const retry_strategy& strategy, std::optional<reply::status_type> expected, abort_source* as) {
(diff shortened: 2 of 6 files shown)


avikivity:
Don't post pull requests with such titles.

kreuzerkrieg:
> Don't post pull requests with such titles.

What is the acceptable way to create a PR which I dont want to be reviewed yet but I want the CI to run on it?

avikivity:
> > Don't post pull requests with such titles.
> 
> What is the acceptable way to create a PR which I dont want to be reviewed yet but I want the CI to run on it?

Run the tests locally.

regevran:
> Run the tests locally.

It is not always enough.

But to the point itself - we don't have a mechanism to open a PR that is NOT being reviewed.
We tried to open as a draft - got reviews.
There is no label such as "ready for review" or "work in progress" or "do not review".

Github PR is a good tool to organize one's work - with history, with self comments and with _selected_ participants when asked.

This `DONOTREVIEW` technique worked well for us to do all the mentioned above actions, and I see no replacement to this at the moment.

> What is the acceptable way to create a PR which I dont want to be reviewed yet but I want the CI to run on it?

This is a real need.

kreuzerkrieg:
> > Run the tests locally.
> 
> It is not always enough.
> 
> But to the point itself - we don't have a mechanism to open a PR that is NOT being reviewed. We tried to open as a draft - got reviews. There is no label such as "ready for review" or "work in progress" or "do not review".
> 
> Github PR is a good tool to organize one's work - with history, with self comments and with _selected_ participants when asked.
> 
> This `DONOTREVIEW` technique worked well for us to do all the mentioned above actions, and I see no replacement to this at the moment.
> 
> > What is the acceptable way to create a PR which I dont want to be reviewed yet but I want the CI to run on it?
> 
> This is a real need.

As I suggested more than once, we can have a label what will force the draft PR being built by the CI
Just saying

avikivity:
> 
> > What is the acceptable way to create a PR which I dont want to be reviewed yet but I want the CI to run on it?
> 
> This is a real need.

CI verifies that the tests you ran locally were sufficient and expands them to cover more toolchains.

It's the last step in the process, not the first step. By the time you are ready to pass CI, you are also ready for review.

regevran:
> It's the last step in the process

But what we do with this we tie CI to PRs - and they both do not necessarily depends on each other.
You might want to have a PR for other reasons that are NOT CI.
And you might want to have CI without having a PR.

`DONOTREVIEW` technique is "not having a PR" but run CI.
`DONOTREVIEW` technique is also used for other requirements from a PR witch are not CI  (like collaborating with a small group only).

xemul:
@kreuzerkrieg , before I review the final version, can you show would Scylla S3 client look like on top of this change?

src/http/retry_strategy.cc
```diff
@@ -0,0 +1,82 @@
...
+logger rs_logger("default_http_retry_strategy");
+
+extern bool is_retryable_exception(std::exception_ptr ex);
+
+default_retry_strategy::default_retry_strategy(unsigned max_retries, unsigned scale_factor)
+    : _max_retries(max_retries), _scale_factor(scale_factor) {
+}
+
+default_retry_strategy::default_retry_strategy()
+    : default_retry_strategy(2, 25) /* Default overall retry time to fail is 50ms (0ms)+((1<<1)*25)ms) */ {
```
xemul, line 72: Why 2? Or does _max_retries include the initial attempt as well?

kreuzerkrieg:
> @kreuzerkrieg , before I review the final version, can you show would Scylla S3 client look like on top of this change?

https://github.com/scylladb/scylladb/pull/24498

kreuzerkrieg:
> > @kreuzerkrieg , before I review the final version, can you show would Scylla S3 client look like on top of this change?
> 
> [scylladb/scylladb#24498](https://github.com/scylladb/scylladb/pull/24498)

and a proof that current master still works with these changes :)
https://github.com/scylladb/scylladb/pull/25245

kreuzerkrieg:
v2:
- rebased on master
- added another commit moving [this piece of code](https://github.com/scylladb/scylladb/blob/master/utils/s3/retryable_http_client.cc#L32) to the seastar

src/http/client.cc
```diff
@@ -341,6 +341,9 @@ future<> client::make_request(request& req, reply_handler& handle, std::optional
 }
 
 future<> client::make_request(request& req, reply_handler& handle, const retry_strategy& strategy, std::optional<reply::status_type> expected, abort_source* as) {
+    if (as && as->abort_requested()) {
```
xemul, line 344: But both, with_connection() and do_make_request() would bail out early in case abort was requested, won't they?
Have you seen them got stuck somewhere that required the early abortion?

include/seastar/http/retry_strategy.hh
```diff
@@ -0,0 +1,60 @@
...
+public:
+    virtual ~retry_strategy() = default;
+    // Returns true if the error can be retried given the error and the number of times already tried.
+    virtual future<bool> should_retry(std::exception_ptr error, unsigned attempted_retries) const = 0;
+    virtual future<input_stream<char>> analyze_reply(std::optional<reply::status_type> expected, const http::reply& rep, input_stream<char>&& in) const = 0;
+
+    // Calculates the time in milliseconds the client should wait before attempting another request based on the error and attempted_retries count.
+    [[nodiscard]] virtual std::chrono::milliseconds delay_before_retry(std::exception_ptr error, unsigned attempted_retries) const = 0;
+
+    [[nodiscard]] virtual unsigned get_max_retries() const noexcept = 0;
```
xemul, line 29: This method seems to be "private" to retry strategy in a sense that http client doesn't call it.
Also, why should it be virtual?

src/http/client.cc
```diff
@@ -361,16 +338,37 @@ future<> client::make_request(request& req, reply_handler& handle, std::optional
...
-        // The 'con' connection may not yet be freed, so the total connection
-        // count still account for it and with_new_connection() may temporarily
-        // break the limit. That's OK, the 'con' will be closed really soon
-        return with_new_connection([this, &req, &handle, as, expected] (connection& con) {
-            return do_make_request(con, req, handle, as, expected);
-        }, as);
+        return do_with(std::move(ex), [this, &req, &handle, as, expected](std::exception_ptr& err) {
+            return repeat_until_value([this, &err, count = 0ull, &req, &handle, as, expected]() mutable -> future<std::optional<bool>> {
+                       ++count;
+                       return _retry_strategy->should_retry(err, count)
```
xemul: Thinking out loud: the `should_retry()` should probably do the sleep before retry on its own

src/http/client.cc
```diff
@@ -393,21 +393,11 @@ future<> client::do_make_request(connection& con, request& req, reply_handler& h
...
-                });
-            });
-        }
-
         auto in = req._method != "HEAD" ? con.in(rep) : input_stream<char>(data_source(std::make_unique<skip_body_source>(rep)));
-        return handle(rep, std::move(in)).then([this, reply = std::move(reply), &con] {
+
+        return _retry_strategy->analyze_reply(expected, rep, std::move(in)).then([this, &handle, &con, reply = std::move(reply)](auto _in) mutable {
+            auto in = std::move(_in);
+        return handle(*reply, std::move(in)).then([this, reply = std::move(reply), &con] {
```
xemul: I wonder what benefits a user get from the analyze_reply method? Why can't it do the analysis from inside the handler itself?

src/http/retry_strategy.cc
```diff
@@ -0,0 +1,82 @@
...
+    if (attempted_retries >= get_max_retries()) {
+        rs_logger.warn("Retries exhausted. Retry# {}", attempted_retries);
+        co_return false;
+    }
+
+    co_return is_retryable_exception(error);
+}
+
+future<input_stream<char>> default_retry_strategy::analyze_reply(std::optional<reply::status_type> expected, const reply& rep, input_stream<char>&& in) const {
+    co_return std::move(in);
```
xemul: Looking at how it's implemented in Scylla (https://github.com/scylladb/scylladb/pull/24498) it looks like the implementation of the analyze_reply should resolve into exceptional future if it reads from `in`. Is that correct?

src/http/retry_strategy.cc
```diff
@@ -31,7 +31,38 @@ extern logger http_log;
 namespace http::experimental {
 logger rs_logger("default_http_retry_strategy");
 
-extern bool is_retryable_exception(std::exception_ptr ex);
+static bool is_retryable_exception(std::exception_ptr ex) {
```
xemul, line 34: The [aws retry strategy](https://github.com/scylladb/scylladb/pull/24498/commits/21f6920dcd07614b26ad7246e8b68b8b56ddda2a) should call default_retry_strategy::should_retry() to also inherit _this_ logic, but it it looks like it doesn't Why?

src/http/retry_strategy.cc
kreuzerkrieg: Includes, I never have solid opinion on how to count retries; is first attempt should be counted or not

src/http/client.cc
kreuzerkrieg: right, but not early enough
https://github.com/scylladb/scylladb/pull/24498/commits/0b3c1f20655b006d9a556abcc77cd2063386e506#diff-21766d4add17dd0a3368ae220a4ce8aadbefde5742b37927afb1ad6596b9a8f0L32

we count on this check in our tests

include/seastar/http/retry_strategy.hh
kreuzerkrieg: it is virtual in case you want to override it when (and if) it would provide somewhat more complicated than just returning a member
it is public since "why not?" in case if I want to add some trace printing at higher level, say s3_client, why to prevent from one to get access to this value?

src/http/client.cc
kreuzerkrieg: hm... maybe... it definitely could be done, but do you see some scenario when it will be needed? coz the only reason to do that is to allow not to wait at all. but for that you may tweak the `delay_before_retry` returning 0 for example

src/http/client.cc
kreuzerkrieg: why to bother in reply handler and check everytime you write the handler what to do with the reply? once you have `analyze_reply` which is kinda one-size-fits-all it removes from the handler implementor the responsibility to handle errors, the handler should focus on doing something useful with the reply, not to handle erroneous states

src/http/retry_strategy.cc
kreuzerkrieg: right

src/http/retry_strategy.cc
kreuzerkrieg: Since I want completely to take over the logic of the `should_reply` thus the `aws_retry_strategy` do not call `default_retry_strategy::should_retry()` since `aws_retry_strategy` does the same and much more than that

src/http/retry_strategy.cc
xemul: The way I see it -- try, retry, retry, ... so the first attempt is a try, not retry

include/seastar/http/retry_strategy.hh
xemul: To keep the retry strategy public API minimal

src/http/client.cc
xemul: [Again](https://github.com/scylladb/seastar/pull/2806#discussion_r2278944123), to keep the retry strategy API as little as possible

src/http/client.cc
xemul: It looks artificial that client calls two user-provided callbacks one after another with almost identical set of arguments. The fact that S3 reply handler has pretty universal analyze reply that also needs to read request body is the feature of S3 client.

src/http/retry_strategy.cc
xemul: OK, makes sense. The goal is to provide enough customization for S3 client, and plug it with some sane out-of-the-box r.s.

src/http/client.cc
xemul: The link points to some commit that just removes a bunch of files, so I don't get the point :(
Anyway, is it the essential part of the r.s. rework? Or does it fix its own bug? If the latter, then let's push it as separate PR

kreuzerkrieg: Line 32
https://github.com/scylladb/scylladb/blob/master/utils/s3/retryable_http_client.cc#L32

> then let's push it as separate PR
Please take a look at the code and comment, are you sure we want it as a new PR?

xemul:
Closing in favor of #2960

PR #2809, opened 2025-06-17
reactor: add counter for internal errors

denesb:
on_internal_error() and friends are used where a regular exception is not harsh, mostly replacing calls to assert(). Therefore, calls to this method almost always indicate some malfunction in the code, like violation of pre/post conditions or corruption.
Operators of seastar applications want to know if such events happen but currently calls to on_internal_error() can only be observed from the logs. These logs are not unique and there is no non-ambiguous way to filter on_internal_error() from the logs. All this means that operators have poor visibility into their application when it comes to internal errors.
This patch adds a counter which provides the following benefits:
* unambiguous way to determine whether any internal errors happened
* set up alerts so support staff can intervene and investigate the source of the errors


avikivity:
> on_internal_error() and friends are used where a regular exception is not harsh, mostly replacing calls to assert(). Therefore, calls to this method almost always indicate some malfunction in the code, like violation of pre/post conditions or corruption. Operators of seastar applications want to know if such events happen but currently calls to on_internal_error() can only be observed from the logs. These logs are not unique and there is no non-ambiguous way to filter on_internal_error() from the logs. 

This bit could easily be fixed.

> All this means that operators have poor visibility into their application when it comes to internal errors. This patch adds a counter which provides the following benefits:
> 
> * unambiguous way to determine whether any internal errors happened
> * set up alerts so support staff can intervene and investigate the source of the errors



> on_internal_error() and friends are used where a regular exception is not harsh, mostly replacing calls to assert(). Therefore, calls to this method almost always indicate some malfunction in the code, like violation of pre/post conditions or corruption. Operators of seastar applications want to know if such events happen but currently calls to on_internal_error() can only be observed from the logs. These logs are not unique and there is no non-ambiguous way to filter on_internal_error() from the logs. All this means that operators have poor visibility into their application when it comes to internal errors. This patch adds a counter which provides the following benefits:
> 
> * unambiguous way to determine whether any internal errors happened
> * set up alerts so support staff can intervene and investigate the source of the errors

Issue #2802, opened 2025-06-10
iotune: measurement results does not scale on larger nodes of EC2


syuu1228:
While measuring with iotune to add IO parameters for a new EC instance type to scylla-machine-image, we found that iotune measurements did not scale with larger instance sizes compared to the IOPS officially published by AWS.

To confirm whether this was a problem specific to iotune, we ran fio directly on the block device.

We measured the benchmark performance by running it on a single drive and multiplying the number of disks to estimate the theoretical value for RAID 0, and by running the benchmark on RAID 0 to obtain the actual measured value for RAID 0.

Here are the results on i7i instance type.

The meaning of each column is as follows:
- iotune_theorical: Theoretical value obtained by running iotune on a single drive and multiplying by the number of disks
- iotune_raid0: Actual value obtained by running iotune on RAID0
- fio_theorical: Theoretical value obtained by running fio on a single drive and multiplying by the number of disks
- fio_raid0: Actual value obtained by running fio on RAID0
- aws_spec: The IOPS officially published by AWS
- delta: delta between each value and aws_spec

### read_iops comparison  
instance_type | iotune_theorical | delta | iotune_raid0 | delta | fio_theorical | delta | fio_raid0 | delta | aws_spec
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
i7i.large | 82 K | 9.69% | 82 K | 9.70% | 79 K | 4.89% | 79 K | 5.12% | 75 K
i7i.xlarge | 164 K | 9.64% | 164 K | 9.65% | 158 K | 5.11% | 157 K | 4.83% | 150 K
i7i.2xlarge | 328 K | 9.42% | 328 K | 9.47% | 314 K | 4.76% | 313 K | 4.30% | 300 K
i7i.4xlarge | 562 K | -6.37% | 584 K | -2.69% | 623 K | 3.80% | 622 K | 3.69% | 600 K
i7i.8xlarge | 1,071 K | -10.77% | 1,007 K | -16.10% | 1,244 K | 3.68% | 1,234 K | 2.80% | 1,200 K
i7i.12xlarge | 1,559 K | -13.39% | 1,446 K | -19.69% | 1,866 K | 3.66% | 1,788 K | -0.65% | 1,800 K
i7i.16xlarge | 2,056 K | -14.34% | 1,859 K | -22.55% | 2,485 K | 3.54% | 2,268 K | -5.49% | 2,400 K
[...]

PR #2811, opened 2025-06-19
io_tester: Collect and report disk queue length samples

xemul:
Yet another useful metrics to have. So far it wasn't reported, because queue length (both disk and io-scheduler) are not counters or some static values. Queue lengths is constantly changing, so there's not one value for it to be reported. Capturing maximum is also not nice, because it will capture a (random) spike giving no idea about how disk was in fact saturated.

Having said all of that -- this patch collects disk queue lengths samples and reports median, p90 and maximum values of the set. Samples are collected 5 times a second not to disturb the jobs too much yet to collect enough samples.

### DIFF
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -263,6 +263,7 @@ struct job_config {
 };
 
 std::array<double, 4> quantiles = { 0.5, 0.95, 0.99, 0.999};
+std::array<double, 2> quantiles_short = { 0.5, 0.9 };
 static bool keep_files = false;
 
 future<> maybe_remove_file(sstring fname) {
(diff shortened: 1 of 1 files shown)


apps/io_tester/io_tester.cc
```diff
@@ -568,9 +571,25 @@ class io_class_data : public class_data {
     }
 
 public:
-    io_class_data(job_config cfg) : class_data(std::move(cfg)) {}
+    io_class_data(job_config cfg)
+            : class_data(std::move(cfg))
+            , _queue_length_timer([this] { update_queue_length(); })
+            , _disk_queue_lengths(extended_p_square_probabilities = quantiles_short)
```
avikivity, line 577: odd syntax... boost

avikivity (approved):

robertbindar (approved):
Tested it during i7i tests for https://github.com/scylladb/seastar/issues/2802 and it works alright.

PR #2812, opened 2025-06-20
http/client: Fix content length body overflow check (and a bit more)

xemul:
When writing a data over specified content length, the helper output_stream throws. However, there's a corner case when it doesn't -- when the body is byte-to-byte full and extra bytes are put, they are silently ignored.

This PR fixes it and adds a unit test for that corner case. Also the test covers zero-copy behavior.

### DIFF
--- a/src/http/common.cc
+++ b/src/http/common.cc
@@ -188,7 +188,7 @@ public:
 
         auto size = buf.size();
         if (_bytes_written + size > _limit) {
-            return make_exception_future<>(std::runtime_error(format("body conent length overflow: want {} limit {}", _bytes_written + buf.size(), _limit)));
+            return make_exception_future<>(std::runtime_error(format("body content length overflow: want {} limit {}", _bytes_written + buf.size(), _limit)));
         }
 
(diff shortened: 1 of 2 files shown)


src/http/common.cc
```diff
@@ -173,11 +173,11 @@ class http_content_length_data_sink_impl : public data_sink_impl {
...
     virtual future<> put(temporary_buffer<char> buf) override {
-        if (buf.size() == 0 || _bytes_written == _limit) {
+        auto size = buf.size();
+        if (size == 0) {
             return make_ready_future<>();
         }
 
-        auto size = buf.size();
         if (_bytes_written + size > _limit) {
             return make_exception_future<>(std::runtime_error(format("body conent length overflow: want {} limit {}", _bytes_written + buf.size(), _limit)));
```
avikivity: test?

xemul:
upd:
- fixed a misprint in exception message
- added unit test

xemul:
@avikivity , please re-review

PR #2813, opened 2025-06-20
http: Deprecate json_exception

xemul:
That's not in fact an exception, it's a helper class that's used by routes code to convert base_exception into seastar-specific body error message.

Deprecate the class and leave temporary alias for routes.

### DIFF
--- a/include/seastar/http/exception.hh
+++ b/include/seastar/http/exception.hh
@@ -127,7 +127,7 @@ public:
     explicit response_parsing_exception(const std::string& msg) : server_error_exception(msg) {}
 };
 
-class json_exception : public json::json_base {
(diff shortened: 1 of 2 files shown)


xemul:
@nyh , please review

xemul:
Currently all exceptions from this file render themselves into body as json string like `{"code": N, "message": "text"}`. There's a PR #2799 to support plain-text erroneous replies and [I thought](https://github.com/scylladb/seastar/pull/2799#discussion_r2151348803) that it would be nicer to have slightly different exceptions hierarchy to reflect which exceptions render to which kind of body. The existing `json_exception` class stands on the way. And since it's really an internal thing, I decided to start with making it non-public.

PR #2823, opened 2025-06-26
Nested scheduling groups (CPU only)

xemul:
Introduce the API to create supergroups and tune up CPU scheduler to schedule the hierarchy of super-groups and groups.

First things first. Current seastar scheduling model has the concept of task, task_queue and scheduling_group.

The task_queue is an object that holds a FIFO list of task-s and, when being run, processes them one by one as long as preemption is not requested or the list of tasks is drained. In that sense the  task_queue corresponds to Linux CPU task_struct, and the task itself doesn't have relations in there. The scheduling_group is just the "API name" for task_queue -- when a scheduling_group is created, each reactor gets a task_queue of the group's ID and that's it.

Having said that, in order to introduce nested scheduling, there should appear something that corresponds to "group of task_struct-s" in Linux (i.e. -- a cgroup thing). Since the "group" word is already busy at the API level, the scheduling_supergroup is introduced there which corresponds to a "group of scheduling_group-s or other supergroup-s". in the core the supergroup maps to task_queue_group. This newly introduced struct wraps the existing reactor-level lists of task_queue-s and vruntime management, but allows dynamically creating task_queue_groups and nest them into each other. Thus, the task_queue_group corresponds to Linux cgroup. On start, reactor statically allocates root task_queue_group.

For demonstration purposes API allows creating of supergroups inside root supergroup (i.e. -- next to legacy scheduling_groups), but it can be easily extended. The scheduling_groups creation API gains a new overload that accepts supergroup parent parameter and links the newly created group into the given supergroup. Other API remains backward compatible, i.e. -- calling existing methods would link task_queue into reactor's task_queue_group, i.e. -- the group will happen at _exactly_ the same place as it does now.


xemul:
upd:
- rename sched_entity::activate -> wakeup()
- don't wakeup task_queue_group twice
- lost temporary cast of sched_entity to task_queue when running tasks in a group

src/core/reactor.cc
```diff
@@ -3021,43 +3040,61 @@ void reactor::stop_aio_eventfd_loop() {
 inline
 bool
 reactor::have_more_tasks() const {
-    return _active_task_queues.size() + _activating_task_queues.size();
+    return _cpu_sched.active();
+}
+
+inline bool reactor::task_queue_group::active() const noexcept {
+    return _active.size() + _activating.size();
```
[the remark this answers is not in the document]
xemul: It's pre-existing code
Nonetheless, it was done on purpose to avoid extra conditional instructions

src/core/reactor.cc
```diff
@@ -3140,28 +3190,22 @@ reactor::run_some_tasks() {
...
-    //
-    // FIXME: different scheduling groups have different sensitivity to jitter, take advantage
-    tq._vruntime = std::max(_last_vruntime, tq._vruntime);
     auto now = reactor::now();
-    tq._waittime += now - tq._ts;
-    tq._ts = now;
-    _activating_task_queues.push_back(&tq);
+    _waittime += now - _ts;
+    _ts = now;
+    _parent->activate(this);
```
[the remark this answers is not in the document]
xemul: Parent cannot be null here, because of the _active check above, and root node is always active
Assertion might be good, but debug-mode tests would capture that and in release-mode we try to avoid checks as much as possible

xemul:
upd:
- rebased
- track and limit the number of subgroups in a supergroup
- documented supergroup creation API
- added set_shares() for supergroups

xemul:
upd:
- fixed supergroup registration on non-0 shard
- added tests, perf and unit

xemul:
upd:
- test compilation fix (std::abs usage)

avikivity:
> task_struct-s" in Linux (i.e. -- a cgroup thing). Since the "group" word is already busy at the API level, the scheduling_supergroup is introduced there which corresponds to a "group of scheduling_group-s or other supergroup-s"

Why can't we keep using scheduling_group? A scheduling group can run tasks or other scheduling groups (or both, but there won't be reasonable prioritization between them).

avikivity:
>  Thus, the task_queue_group corresponds to Linux cgroup

The way I saw it (but didn't implement, so very easy) is that a nested scheduling group implements task (or has a proxy task), and when the parent scheduling group runs it, it runs the nested task_queue.

```
scheduling_group
 -> task_queue
        task
        task
        task
           -> nested scheduling_group
                -> task_queue
                       task
                       task
                       task
        task
           -> nested scheduling_group
                -> task_queue
                       task
                       task
                       task
        task
```

src/core/reactor.cc
```diff
@@ -2645,25 +2645,23 @@ seastar::internal::log_buf::inserter_iterator do_dump_task_queue(seastar::intern
 
 void reactor::task_queue::run_tasks() {
     reactor& r = engine();
-    task_queue& tq = *this;
```
avikivity: Likely this is a remnant of extracting run_tasks() from reactor to implement scheduling in the first place.

xemul:
> The way I saw it (but didn't implement, so very easy) is that a nested scheduling group implements task (or has a proxy task), and when the parent scheduling group runs it, it runs the nested task_queue.

In your example nested scheduling groups shares won't be taken into account at all, because the list of tasks in parent task_queue is literally a priority-less list, it doesn't count anyone's runtimes and vruntimes.

xemul:
> Why can't we keep using scheduling_group? A scheduling group can run tasks or other scheduling groups (or both, but there won't be reasonable prioritization between them).

We can use scheduling_group name, but in that case we'll need to tell scheduling_groups with tasks from scheduling_groups with other scheduling_groups. Intermixing tasks and scheduling_groups in one list will only affect the way tasks spawn each other.

xemul:
upd:
- put unit test into its own .cc file, otherwise it clashes with existing sched group tests that don't clean throwing sched group keys after themselves

src/core/reactor.cc
xemul: Yes. It's "continuation of the previous patch"

xemul:
upd:
- Put on-reactor task_queue_group _cpu_sched in front of _task_queues. Otherwise asan complains on reactor destruction:
```
src/core/reactor.cc:966:18: runtime error: member access within address 0x524000023b10 which does not point to an object of type 'task_queue_group'
0x524000023b10: note: object is of type 'seastar::reactor::sched_entity'
00 00 00 00  90 4b 8d a9 d4 7f 00 00  00 00 00 00 00 00 00 00  00 00 80 3f 00 00 00 00  00 00 00 00
```

src/core/reactor.cc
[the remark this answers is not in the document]
avikivity: Can't we eject the co-pilot?

include/seastar/core/reactor.hh
```diff
@@ -269,6 +269,7 @@ private:
 
     struct sched_entity {
         explicit sched_entity(task_queue_group* p, float shares);
+        virtual ~sched_entity();
```
avikivity: Why is it needed to be virtual? Don't we always destroy sched_entities via their real classes?

avikivity: Not that I have an objection - but I'd like to keep things tight.

include/seastar/core/reactor.hh
```diff
@@ -314,6 +314,7 @@ private:
         scheduler_list _activating;;
         unsigned _nr_children = 0;
```
avikivity, line 318: > When running task_queues from the root task_queue_group, reactor pings
stall detector, kicks lowres clock and resets preemption monitor. At the
end, it cleans up the current_scheduling_group_ptr pointer.
>
> All this will become unneeded for nested task_queue_group-s, only the
running loop will be needed.

I don't see why it will be unneeded. Perhaps the parent group will always run and take care of it?

xemul:
upd:
- Provide non-empty sched group fairness test for debug case

include/seastar/core/reactor.hh
xemul: Yes. The point is that all those kicks will be done by the root group only, nested groups can skip it and leave only running loop in their virtual run_tasks method

include/seastar/core/reactor.hh
xemul: Hm... You're right, destruction always happens via final classes

src/core/reactor.cc
```diff
@@ -3063,7 +3066,11 @@ void reactor::task_queue_group::activate(sched_entity* tq) {
     //
     // FIXME: different scheduling groups have different sensitivity to jitter, take advantage
     tq->_vruntime = std::max(_last_vruntime, tq->_vruntime);
+    bool was_active = active();
     _activating.push_back(tq);
+    if (!was_active) {
+        this->sched_entity::wakeup();
```
avikivity: I think this-> is unneeded.

include/seastar/core/scheduling.hh
```diff
@@ -64,6 +65,16 @@ T* scheduling_group_get_specific_ptr(scheduling_group sg, scheduling_group_key k
...
 
+/// Creates a scheduling supergroup with a specified number of shares.
+///
+/// The operation is global and affects all shards. The returned scheduling
+/// supergroup can then be used in any shard.
+///
+/// \param shares number of shares of the CPU time allotted to the group;
+///              Use numbers in the 1-1000 range (but can go above).
+/// \return a scheduling supergroup that can be used on any shard
+future<scheduling_supergroup> create_scheduling_supergroup(float shares) noexcept;
```
avikivity, line 78: I'd like to see if we can share the APIs.

avikivity:
Very good. The only question is whether to reuse scheduling_group or have the new one. Having a separate one is more type-safe, but causes duplication in the API.

avikivity:
> > The way I saw it (but didn't implement, so very easy) is that a nested scheduling group implements task (or has a proxy task), and when the parent scheduling group runs it, it runs the nested task_queue.
> 
> In your example nested scheduling groups shares won't be taken into account at all, because the list of tasks in parent task_queue is literally a priority-less list, it doesn't count anyone's runtimes and vruntimes.

Right.

include/seastar/core/reactor.hh
avikivity: Can ensure by making the base-class destructor protected.

avikivity:
We need more unit tests for wakeup between ssg1->ssg2->sg1->t1 --> ssg4->sg2->t2 and similar weird stuff. Perhaps randomized.

avikivity:
I think we can keep scheduling_supergroup

| Linux | Seastar |
|--------|--------|
| thread | task |
| process (many threads) | scheduling_group (many tasks)|
| pid namespace (many processes, can nest)| supergroup (many scheduling_groups, can nest)|

xemul:
> Linux 	Seastar
> thread 	task

That's not correct, Linux threads and Linux tasks are indistinguishable from the CPU scheduler perspective. Seastar task doesn't have any corresponding entity in the Linux kernel

xemul:
> We need more unit tests for wakeup between ssg1->ssg2->sg1->t1 --> ssg4->sg2->t2 and similar weird stuff. Perhaps randomized.

OK, I'll try to come up with something

xemul:
upd:
- removed this-> from wakeup calling line
- made ~sched_entity protected and non-virtual

avikivity:
> > Linux 	Seastar
> > thread 	task
> 
> That's not correct, Linux threads and Linux tasks are indistinguishable from the CPU scheduler perspective. Seastar task doesn't have any corresponding entity in the Linux kernel

It's from a container/owner perspective. You're right that it's not an analog from the scheduler's point of view (it doesn't schedule processes).

src/core/reactor.cc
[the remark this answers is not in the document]
xemul: We could "not inject" it. I was just curious if it would find anything relevant in large PR

xemul:
upd:
- added wakeup randomized test (last patch)

xemul:
upd:
- since there's no (yet) supergroup destruction API, run all test cases in "shared" groups layout

src/core/reactor.cc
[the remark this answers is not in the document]
avikivity: Yes it's tempting but it uses too little context in its reviews. Here the context includes the history of Linux scheduling.

xemul:
upd:
- changed the order of on-reactor _task_queues and _supergroups

Interesting. When we'll have truly nested supergroups (in this PR only one-level nesting is supported) we'll need to solve recursion problem. Like this:

```c++
    std::array<std::unique_ptr<task_queue_group>, size> _supergroups;
    _supergroups[0] = std::make_unique<task_queue_group>(parent = &engine()::_cpu_sched, ...);
    _supergroups[1] = std::make_uniqye<task_queue_group>(parent = _supergroups[0].get(), ...);
```

now, upon destruction of `_supergroups` the 0th element will be freed before the 1st and when destructing the 1st ASAN will complain about use-after-free or badly-typed _parent pointer (like [here](https://github.com/scylladb/seastar/pull/2823#issuecomment-3024500353))

avikivity:
> upd:
> 
> * changed the order of on-reactor _task_queues and _supergroups
> 
> Interesting. When we'll have truly nested supergroups (in this PR only one-level nesting is supported) we'll need to solve recursion problem. Like this:
> 
> ```c++
>     std::array<std::unique_ptr<task_queue_group>, size> _supergroups;
>     _supergroups[0] = std::make_unique<task_queue_group>(parent = &engine()::_cpu_sched, ...);
>     _supergroups[1] = std::make_uniqye<task_queue_group>(parent = _supergroups[0].get(), ...);
> ```
> 
> now, upon destruction of `_supergroups` the 0th element will be freed before the 1st and when destructing the 1st ASAN will complain about use-after-free or badly-typed _parent pointer (like [here](https://github.com/scylladb/seastar/pull/2823#issuecomment-3024500353))


We can require that the application acquiesce all scheduling groups and destroy them before destroying supergroups, etc.

src/core/reactor.cc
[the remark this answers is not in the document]
xemul: Agree. So far I haven't even noticed that it used anything outside of the pr itself.  Maybe even worse -- each commit individually. In particular, here, the fact that null parent pointer and active bit are connected could be observed from other patches, but copilot failed to do it.

xemul:
upd:
- just rebased

nyh:
> The task_queue is an object that holds a FIFO list of task-s and, when being run, processes them one by one as long as preemption is not requested or the list of tasks is drained. In that sense the task_queue corresponds to Linux CPU task_struct, and the task itself doesn't have relations in there. The scheduling_group is just the "API name" for task_queue -- when a scheduling_group is created, each reactor gets a task_queue of the group's ID and that's it.
> 
> Having said that, in order to introduce nested scheduling, there should appear something that corresponds to "group of task_struct-s" in Linux (i.e. -- a cgroup thing).

Did you mean "groups of task_queue-s" in this sentence?

avikivity:
btw, we can have nested scheduling groups take over the role of execution_stage. Instead of queueing a function to an execution stage, queue a continuation chain to a subgroup. It's not exactly the same, but it can be better if we do it carefully, and it eliminates some duplication in queue management between task_queue and execution_stage.

xemul:
> > Having said that, in order to introduce nested scheduling, there should appear something that corresponds to "group of task_struct-s" in Linux (i.e. -- a cgroup thing).
> 
> Did you mean "groups of task_queue-s" in this sentence?

No, I meant "group of task_struct-s", because I was speaking about "corresponds to ... in Linux". And since seastar::task_queue corresponds to linux::task_struct, this sentence mean that "there should appear group of task_queue-s" in Seastar.

xemul:
upd:
- made some supergroup methods public
- added internal method to get supergroup for a group
- added supergroup::get_shares()

The above is needed for IO classes nesting (the draft PR will come soon)

xemul:
@avikivity , are there any other concerns you have about this PR?

avikivity:
I missed the update, need to review it.

tests/perf/future_util_perf.cc
```diff
@@ -457,6 +457,8 @@ PERF_TEST_CN(parallel_for_each, cor_pfe_suspend_100)
 
 struct sched {
     bool once = false;
+    seastar::scheduling_supergroup s1;
+    seastar::scheduling_supergroup s2;
     seastar::scheduling_group g1;
```
avikivity, line 462: > test: Add perf test for context switch latency
> Next to root-level switching, add a test that measures context switch
> time between two groups sitting in its own supergroups.
> 
> Results (dev mode):
> 
> test                     iterations      median         mad         min         max      allocs       tasks        inst      cycles
> sched.context_switch        6886000   139.930ns     0.087ns   139.844ns   141.176ns       0.000       2.000       815.1       559.7
> sched.context_switch_x1_5   4495000   208.424ns     0.185ns   207.872ns   208.609ns       0.000       2.000      1238.8       832.4
> sched.context_switch_x2     3378000   285.141ns     1.653ns   283.488ns   288.512ns       0.000       2.000      1663.1      1146.1
> 
> The x1_5 test runs /group vs /supergroup/group
> The x2 test runs two distinct /supergroup/group
> 
> Signed-off-by: Pavel Emelyanov <xemul@scylladb.com>

The results are quite poor. Even the base case is slow, but why 1.6k insn for the x2 case?

avikivity:
What about destroy_scheduling_supergroup()?

How does the penalty for waking a task in a sleeping group in a sleeping supergroup work?

tests/perf/future_util_perf.cc
xemul: It was dev-mode, but release-mode is not much better. Here's comparison of master branch with this PR

```
test                         iterations      median         mad         min         max      allocs       tasks        inst      cycles
sched.context_switch(master)    7673000   123.597ns     0.302ns   122.496ns   128.079ns       0.000       2.000       709.2       487.9
sched.context_switch            7419000   130.374ns     0.451ns   129.408ns   137.770ns       0.000       2.000       749.4       521.8
sched.context_switch_x1_5       4791000   202.268ns     0.688ns   201.361ns   206.075ns       0.000       2.000      1132.0       809.1
sched.context_switch_x2         3538000   274.455ns     1.454ns   272.960ns   279.240ns       0.000       2.000      1515.0      1095.5
```

This PR adds some overhead for flat case, yes

> ... but why 1.6k insn for the x2 case?

x2 case wakes up two entities -- one group and the enclosing supergroup, thus 2x more instructions

xemul:
> What about destroy_scheduling_supergroup()?

Scylla doesn't destroy scheduling _groups_, so I judged that destroying supergroups will be dead code as well :) I'll cherry-pick one from my "nested-sched-groups-next" branch

> How does the penalty for waking a task in a sleeping group in a sleeping supergroup work?

That's measured by the context-switch test from above comment -- two groups wake-up each other in a loop and have no other tasks running in it, so they wake each other up from idle. The "regular" test wakes up two root-level groups, x2 case wakes up two groups sitting in two distinct supergroups and x1_5 switches between root-level group and group-in-a-supergroup.

Or did you mean something else?

tests/perf/future_util_perf.cc
avikivity: I guess it's wakeup which should be relatively rare.

avikivity:
We do have some destroy calls (I thought we destroyed all).

```
service/qos/service_level_controller.cc:    // if any destroy_scheduling_group call fails, return one of the exceptions
service/qos/service_level_controller.cc:         auto f = co_await coroutine::as_future(destroy_scheduling_group(deleted_scheduling_groups.front()));
test/boost/database_test.cc:        destroy_scheduling_group(unknown_scheduling_group).get();
test/boost/database_test.cc:        destroy_scheduling_group(unknown_scheduling_group).get();
test/boost/logalloc_test.cc:        destroy_scheduling_group(background_reclaim_scheduling_group).get();
```

avikivity:
> > How does the penalty for waking a task in a sleeping group in a sleeping supergroup work?
> 
> That's measured by the context-switch test from above comment -- two groups wake-up each other in a loop and have no other tasks running in it, so they wake each other up from idle. The "regular" test wakes up two root-level groups, x2 case wakes up two groups sitting in two distinct supergroups and x1_5 switches between root-level group and group-in-a-supergroup.
> 
> Or did you mean something else?

I meant this code (which I don't completely understand even without nested scheduling):

```c++
void
reactor::activate(task_queue& tq) {
    if (tq._active) {
        return;
    }
    // If activate() was called, the task queue is likely network-bound or I/O bound, not CPU-bound. As
    // such its vruntime will be low, and it will have a large advantage over other task queues. Limit
    // the advantage so it doesn't dominate scheduling for a long time, in case it _does_ become CPU
    // bound later.
    //
    // FIXME: different scheduling groups have different sensitivity to jitter, take advantage
    tq._vruntime = std::max(_last_vruntime, tq._vruntime);
    auto now = reactor::now();
    tq._waittime += now - tq._ts;
    tq._ts = now;
    _activating_task_queues.push_back(&tq);
}
```

How do we calculate the new _vruntime/_waittime at the different levels?

This was very tricky code to get right.

avikivity:
Specifically, this line:

```
21badbda93 core/reactor.cc     (Glauber Costa             2018-02-02 14:19:45 -0500 3164)     tq._vruntime = std::max(_last_vruntime, tq._vruntime);
```

xemul:
> I meant this code (which I don't completely understand even without nested scheduling):

This has its peer in IO fair_queue, but in IO it looks sligtly different. It has the "forgiving factor" (the notorious "tau" thing) -- the time by which the waking-up-from-idle class can fall back from its peers and thus monopolize the resource. There's a test in fair_queue_test.cc called test_fair_queue_forgiving_queue. In CPU scheduler this forgiving time is hard-coded to be zero.

> ```c++
> void
> reactor::activate(task_queue& tq) {
>     if (tq._active) {
>         return;
>     }
>     // If activate() was called, the task queue is likely network-bound or I/O bound, not CPU-bound. As
>     // such its vruntime will be low, and it will have a large advantage over other task queues. Limit
>     // the advantage so it doesn't dominate scheduling for a long time, in case it _does_ become CPU
>     // bound later.
>     //
>     // FIXME: different scheduling groups have different sensitivity to jitter, take advantage
>     tq._vruntime = std::max(_last_vruntime, tq._vruntime);
>     auto now = reactor::now();
>     tq._waittime += now - tq._ts;
>     tq._ts = now;
>     _activating_task_queues.push_back(&tq);
> }
> ```
> 
> How do we calculate the new _vruntime/_waittime at the different levels?

Short answer: each level acts independently.
Long answer: Each supergroup has its own `_last_vruntime` value calculated out of all this supergroup's children. When a group/supergroup wakes up, its parent supergroup bumps up the activating _vruntime with that formula.

> This was very tricky code to get right.

AFAIR, Linux kernel has non-zero and non-constant forgiving factor, so it's questionable what is "right" here.

xemul:
upd:
- added supergroup destruction API
- updated context-switch patch comment with release-mode runs for both, master branch and this PR

PR #2830, opened 2025-06-30
http::reply: Add 308 (permanent redirect) and make pretty-print handle unknown values

### LINKED ISSUE #2829: http::reply::status_type missing error 308 + lies in printer
The HTTP status enum in `http::reply::status_type` is  missing 308 "permanent redirect". 
This is a useful error code. It should be there.

What is worse however is that IFF a HTTP reponse comes back, containing say a 308, the pretty-printers 
in reply.cc (ostream operator and `response_line()` will _lie_ about the error code, and print it as 500 internal error.

This is very misleading and not great. If the pretty-printer does not find a code in its number-to-string mapping, it should at least just print the number as is, and not confuse hard-working programmers trying to read logs.

elcallio:
Fixes #2829

The HTTP status enum in http::reply::status_type is missing 308 "permanent redirect". This is a useful error code. It should be there.

What is worse however is that IFF a HTTP reponse comes back, containing say a 308, the pretty-printers in reply.cc (ostream operator and response_line() will lie about the error code, and print it as 500 internal error.

This is very misleading and not great. If the pretty-printer does not find a code in its number-to-string mapping, it should at least just print the number as is, and not confuse hard-working programmers trying to read logs.

This adds the 308 entry to status_type, and makes the printer-code print the actual value in case of an unknown error code (jumping through hoops to avoid creating string objects).

### DIFF
--- a/src/http/reply.cc
+++ b/src/http/reply.cc
@@ -67,6 +67,7 @@ static const std::unordered_map<reply::status_type, std::string_view> status_str
     {reply::status_type::not_modified, "304 Not Modified"},
     {reply::status_type::use_proxy, "305 Use Proxy"},
     {reply::status_type::temporary_redirect, "307 Temporary Redirect"},
+    {reply::status_type::permanent_redirect, "308 Permanent Redirect"},
     {reply::status_type::bad_request, "400 Bad Request"},
     {reply::status_type::unauthorized, "401 Unauthorized"},
     {reply::status_type::payment_required, "402 Payment Required"},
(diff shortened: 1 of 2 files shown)


src/http/reply.cc
```diff
@@ -98,20 +99,27 @@ static const std::unordered_map<reply::status_type, std::string_view> status_str
...
-static const std::string_view& to_string_view(reply::status_type status) {
-  if (auto found = status_strings.find(status); found != status_strings.end()) [[likely]]
-    return found->second;
-  return status_strings.at(reply::status_type::internal_server_error);
+template<typename Func>
+static auto with_string_view(reply::status_type status, Func&& func) -> std::invoke_result_t<Func, std::string_view> {
+  if (auto found = status_strings.find(status); found != status_strings.end()) [[likely]] {
+     return func(found->second);
+  }
+  auto dummy_buf = std::to_string(int(status));
```
xemul, line 107: Patch comment says
> jumping through hoops to avoid creating string objects

but std::to_string does create a string object, so what's the point?

elcallio: Invoking func() on map data does not, thus common case still only touches static data, instead of creating a temporary.

elcallio:
ping?

nyh:
Please mention in the commit message https://datatracker.ietf.org/doc/html/rfc7238 which first defined this code as experimental (to be honest, I never noticed that - I heard of 301 Moved permanently), and https://datatracker.ietf.org/doc/html/rfc9110#name-308-permanent-redirect as the final official standardization of this code.

elcallio:
Done.

PR #2831, opened 2025-07-01
http: complete the list of HTTP status codes

### LINKED ISSUE #884: Complete list of HTTP status codes
In include/seastar/http/reply.hh and src/http/reply.cc we have a partial list of HTTP status codes, to which once in a while we need to add a missing status code when a new application notices it's missing.
Let's make this list complete once and for all. The complete list for classic HTTP can be found in https://tools.ietf.org/html/rfc7231#section-8.2.3 and an even more complete list including various HTTP extensions is https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

nyh:
In recent years, Seastar developers added more and more HTTP status codes to the list of those supported by Seastar, on a "as needed" haphazard fashion.

In this patch, I aim make this list *complete* according to RFC 9110:

1. Added a few standard (RFC 9110) error codes which were missing: 407 Proxy Authentication Required 412 Precondition Failed 416 Range Not Satisfiable 421 Misdirected Request

2. Fixed the user-visible *string* attached to some error codes: 302 was "Moved Temporarily", now "Found" 413 was "Payload Too Large", now "Content Too Large" 422 was "Unprocessable Entity", now "Unprocessable Content"

   I want Seastar to print the modern names, not some archaic names. Note that to avoid API breakage, I did not rename the Seastar constants: Code 302 is still reply::status_type::moved_permanently, not renamed to reply::status_type::found.

3. Deleted error codes that are listed in neither RFC 9110 or IANA's HTTP Status Code Registry www.iana.org/assignments/http-status-codes/ Removed the following: 419 Page Expired 440 Login Timeout 509 Bandwidth Limit Exceeded 598 Network Read Timeout 599 Network Connect Timeout

   I don't know why these error codes were ever added - all of them were added in commit db28fcda90521447d30f3a6d13487e8d2c7d9b42 but none of them  are standard in any way. IANA lists all of them as "unassigned". Seastar should make it possible for users to use non-standard error code like "512 Nice Power Of Two Detected" but we should never have non-standard error codes a permanent part of Seastar's API.

4. I left two error codes that are NOT listed in RFC 9110, but are still listed in IANA and codified by older RFCs: 429 Too Many Requests (RFC 6585) and 507 Insufficient Storage (RFC 4918).

Fixes #884.


nyh:
@kreuzerkrieg I asked you to review this patch, because among other things it removes some of the error code you added a year ago in commit db28fcda90521447d30f3a6d13487e8d2c7d9b42 that are not listed by either RFC 9110 or IANA. Maybe you remember why you added them and want to "defend" some of them.

elcallio:
I really don't think we should remove any of the added codes. Most (all?) of the ones you removed here are codes used by AWS and we use in scylladb for mapping error reponses when talking to AWS services. Similar things does/will appear for other cloud/service providers. 

This should not really be an enum at all, but a 

struct status_code {
 int value;
...

static constexpr inline status_code ok = status_code(0); 
... 
etc
};

And maybe some clever way to map value to name string, Right now, we end up adding to this in seastar and/or doing a bunch of int-casts back and forth...

mykaul:
Shame that the very useful status code 418 was deprecated.

elcallio:
Deprecated does not mean unused...

mykaul:
> Deprecated does not mean unused...

It actually says unused:
[15.5.19. ](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.19)[418 (Unused)](https://www.rfc-editor.org/rfc/rfc9110.html#name-418-unused)
[[RFC2324](https://www.rfc-editor.org/rfc/rfc9110.html#RFC2324)] was an April 1 RFC that lampooned the various ways HTTP was abused; one such abuse was the definition of an application-specific 418 status code, which has been deployed as a joke often enough for the code to be unusable for any future use.

Therefore, the 418 status code is reserved in the IANA HTTP Status Code Registry. This indicates that the status code cannot be assigned to other applications currently. If future circumstances require its use (e.g., exhaustion of 4NN status codes), it can be re-assigned to another use.

kreuzerkrieg:
> Shame that the very useful status code 418 was deprecated.

damn, you were 19 minutes faster than I am :)

elcallio:
Which almost certainly means that somewhere, something that any second now will become our number one priority, is using 418 in a mission critical fashion. :-P

kreuzerkrieg:
> @kreuzerkrieg I asked you to review this patch, because among other things it removes some of the error code you added a year ago in commit [db28fcd](https://github.com/scylladb/seastar/commit/db28fcda90521447d30f3a6d13487e8d2c7d9b42) that are not listed by either RFC 9110 or IANA. Maybe you remember why you added them and want to "defend" some of them.

Sure
Because of this
https://github.com/aws/aws-sdk-cpp/blob/f41f66c054c498ce9064692151e5dc25dafb1c10/src/aws-cpp-sdk-core/include/aws/core/http/HttpResponse.h#L73
Their client is processing such a code, meaning their server may return this code, I guess one you remove it, scylladb build will fail

The same goes for 440, 509, 598, 599

nyh:
@kreuzerkrieg while I may be forced to keep those four non-standard error codes for "backward compatibility" (if I do, I still want to deprecate them!), I think it was a mistake adding them in the first place. I can't find any evidence that Amazon actually uses these codes. - it's not mentioned in https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html or any other document that I could find (I even asked Copilot :-)). Did you ever find such documentation? Did you see such errors in tests?

And no, I don't think that a list that someone wrote in a random Amazon driver is any proof that these error codes exist. The comment in front of this list has "The integer values of the response codes correspond to the values in the RFC." - but clearly this is a mistake - these error codes are not listed in any RFC.

Moreover, looking now at Scylla, you don't actually "use" any of these codes... All you do is:
```
aws_error aws_error::from_http_code(seastar::http::reply::status_type http_code)
 {
  ....
    case seastar::http::reply::status_type::page_expired:
        ret_val = all_errors.at("HTTP_PAGE_EXPIRED");
        break;
```

and elsewhere
```
        {"HTTP_PAGE_EXPIRED", aws_error(aws_error_type::HTTP_PAGE_EXPIRED, retryable::yes)},
```

All of this is irrelevant and can be removed if the "page expired" error doesn't exist. And by the way, how did you even decide that 419 Page Expired is a "retryable" request? Did you see it documented somewhere?

But even more importantly if Seastar (not Scylla) is the focus of this discussion:
Even if the "419 page expired" **is** used by Amazon (and again, I still didn't see evidence that it does), it doesn't mean that Seastar needs to support it officially. You could have used in your code the non-standard number "419" without insisting that Seastar must supply a named constant `seastar::http::reply::status_type::page_expired` equal 419. All we need is for Seastar to correctly return 419, or 418 or 512, as error code if that's what the server used - without trying to "convert" it to 500 (which is the bug that Calle fixed). Error 419 should work even if it doesn't have an official name in Seastar.

nyh:
And @kreuzerkrieg, note that the AWS SDK header file you pointed me to also has IM_A_TEAPOT = 418. How come you didn't expect Seastar to have a constant for that too? Clearly, we need to exercises judgement and not just add any incorrect error code we find in some other random library.

I vote for **deleting** the wrong error codes, but if you want Scylla to compile using this wrong (non-standard and unused) error codes page_expired=419 et al,  I can make them "deprecated" instead of outright delete them. But I think leaving them as they are now is a mistake.

nyh:
> I really don't think we should remove any of the added codes. Most (all?) of the ones you removed here are codes used by AWS

Actually, do you have any evidence that one of these four error codes that I deleted are used by AWS? I can't find any such evidence.

> and we use in scylladb for mapping error reponses when talking to AWS services. Similar things does/will appear for other cloud/service providers.

If Scylla wants to support some non-standard error code 419 or 418 or whatever, if should be able to do so using this number. It doesn't mean that Seastar needs to be patched every time Scylla wants to use an incorrect/unused/non-standard/non-sensical error code. This is what we've been doing so far, but I want to end this now. Seastar should list only real, standard, HTTP codes - and all of them. Not less than the standard, and also not more.

> This should not really be an enum at all, but a

The "enum" is just a fancy new (added in 1989!) replacement for a macro - so we get seastar::http::reply::page_expired as a name for 419. So this "enum" is fine. But it doesn't mean that every error code you can imagine (or, in the AI world, "hallucinate"), needs to be listed. Seastar users should be able to read and write error code 418 or 666 if they so wish even if there is no standard meaning for these - and it doesn't mean it needs to be added as a name for everyone!

> 
> struct status_code { int value; ...
> 
> static constexpr inline status_code ok = status_code(0); ... etc };
> 
> And maybe some clever way to map value to name string, Right now, we end up adding to this in seastar and/or doing a bunch of int-casts back and forth...

"adding it to seastar" is a mistake when it comes to random numbers you found on the Internet and not actually standard or not in use. I don't mind the "int cast" - if you want you can create a patch that does this in a more convenient way.

Of course the most important thing is that if the actual response carries "666 Request Of The Beast" status code, Seastar should be able to return 666 as the error code and not convert it to 500 or anything like that, and without us needing to officially add request_of_the_beast=666 to the enum.

elcallio:
> "adding it to seastar" is a mistake when it comes to random numbers you found on the Internet and not actually standard or not in use. I don't mind the "int cast" - if you want you can create a patch that does this in a more convenient way.
> 
> Of course the most important thing is that if the actual response carries "666 Request Of The Beast" status code, Seastar should be able to return 666 as the error code and not convert it to 500 or anything like that, and without us needing to officially add request_of_the_beast=666 to the enum.

No, I am saying it should be so that you could get typed status codes, pretty-print, int-sized code payload, but without having to change seastar each time.

Thus an wrapper class with constants, and optionally the ability to add to _runtime_ value to string mappings for pretty-print would be a better choice here. Because I like the concept of "status_type" being distinct from int, but not it being closed in possible values like an enum.

nyh:
> No, I am saying it should be so that you could get typed status codes, pretty-print, int-sized code payload, but without having to change seastar each time.

What do you mean "pretty print"? A response already has both an error code and a text, like "666 Request Of The Beast". You can print that directly.

If you're talking about **generating** a response with error code "666 Request Of The Beast", I agree you should be able to specify both, but this doesn't seem to be what Scylla needed so I'm in no hurry to provide such a mechanism. Scylla doesn't seem to actually generate error code 419. It seems it only wanted to support it if AWS ever sent it this error code, but it doesn't need to generate such an error.

> Thus an wrapper class with constants, and optionally the ability to add to _runtime_ value to string mappings for pretty-print would be a better choice here. Because I like the concept of "status_type" being distinct from int, but not it being closed in possible values like an enum.

I agree, but I claim this is orthogonal to my patch. Before my patch, and after my patch, Seastar didn't know how to generate "666 Request Of The Beast" errors. If you want to add such a feature, be my guest. But I want this to be the end to the parade of patches that each adds one missing - and sometimes **not missing, just wrong** error codes - to Seastar.

nyh:
The test failed on
```
/home/runner/work/seastar/seastar/tests/unit/httpd_test.cc(859): fatal error: in "content_length_limit": critical check std::string(resp.get(), resp.size()).find("413 Payload Too Large") != std::string::npos has failed [18446744073709551615 == 18446744073709551615]
```
I'll prepare a new version that fixes this test.

Note that changing the error code's **text** was the least important part of my patch, but I thought it was harmless no users would be doing what this test is doing, checking the error message. Applications only use the error code, not the message. (and if they do check the message, shouldn't we use the message in the RFC that the application is likely to check?)

kreuzerkrieg:
> And @kreuzerkrieg, note that the AWS SDK header file you pointed me to also has IM_A_TEAPOT = 418. How come you didn't expect Seastar to have a constant for that too? Clearly, we need to exercises judgement and not just add any incorrect error code we find in some other random library.
> 
> I vote for **deleting** the wrong error codes, but if you want Scylla to compile using this wrong (non-standard and unused) error codes page_expired=419 et al, I can make them "deprecated" instead of outright delete them. But I think leaving them as they are now is a mistake.

https://github.com/aws/aws-sdk-cpp/blob/f41f66c054c498ce9064692151e5dc25dafb1c10/src/aws-cpp-sdk-core/include/aws/core/http/HttpResponse.h#L113

kreuzerkrieg:
Bottom line, it means the only thing we will have to do in s3 client is to cast the `status` to int (the reverse of what we do in the http response parser `static_cast<http::reply::status_type>(std::atoi(str().c_str()))`) and work with `int`s, right?

kreuzerkrieg:
> Actually, do you have any evidence that one of these four error codes that I deleted are used by AWS? I can't find any such evidence.

If the Amazon AWS client is not considered sufficient evidence, then I suggest we proceed by opening a support ticket with AWS.

nyh:
> > And @kreuzerkrieg, note that the AWS SDK header file you pointed me to also has IM_A_TEAPOT = 418. How come you didn't expect Seastar to have a constant for that too? Clearly, we need to exercises judgement and not just add any incorrect error code we find in some other random library.
> > I vote for **deleting** the wrong error codes, but if you want Scylla to compile using this wrong (non-standard and unused) error codes page_expired=419 et al, I can make them "deprecated" instead of outright delete them. But I think leaving them as they are now is a mistake.
> 
> https://github.com/aws/aws-sdk-cpp/blob/f41f66c054c498ce9064692151e5dc25dafb1c10/src/aws-cpp-sdk-core/include/aws/core/http/HttpResponse.h#L113

The more I look at it, the more I'm inclined to insist to ignore all that AWS garbage and look **only at RFCs**.
Check it out - the source code you pointed me now calls 419 "Authentication Timeout". That is not the same as "Page Expired", not even close. This looks to me like old garbage that people just copied around not understanding why. Google searches yield that 419 was used (as "Page Expired", not "Authentication Timeout") by some obscure "Laravel" project, but nothing on any use on AWS.

nyh:
> amazon aws client is not evidence solid enough? then I guess we can open support ticket to AWS

I claim it isn't. it's just C code that people copied around and may contain dead code. It's full of wrong statements (like one claiming this list is from "The RFC") and contraditions (is 419 "Page Expired" or "Authentication Timeout" - I don't see how those possibly refer to the same thing)? Where are these errors documented?

Even worse, that C code has a lot of other garbage error codes that we didn't copy. IM_A_TEAPOT = 418, METHOD_FAILURE = 420, PRECONDITION_REQUIRED = 427, NO_RESPONSE = 444, NO_RESPONSE = 444, RETRY_WITH = 449,  BLOCKED = 450,             REDIRECT = 451,  REQUEST_HEADER_TOO_LARGE = 494, CERT_ERROR = 495,  NO_CERT = 496,  HTTP_TO_HTTPS = 497,         CLIENT_CLOSED_TO_REQUEST = 499 - **all of these** appear in that source code but not in IANA or any RFC I can find. 

I'm asking seriously: How did you decide to add the 4 specific non-existent error codes that I want to delete now, but not these dozen others?

kreuzerkrieg:
> I'm asking seriously: How did you decide to add the 4 specific non-existent error codes that I want to delete now, but not these dozen others?

As was stated in all error handling PRs opened for our s3 client it 100% mimics the official AWS client.
And I'm dead serious. But sounds like this reasoning is not solid enough, I'm ok with that

nyh:
> As was stated in all error handling PRs opened for our s3 client it 100% mimics the official AWS client. And I'm dead serious. But sounds like this reasoning is not solid enough, I'm ok with that

Note that this is the Seastar repository, not the Scylla repository where that S3 code you mentioned lives. 

But more importantly, I showed you a list of a dozen error codes listed in that AWS client lists, including that IM_A_TEAPOT one, that you didn't add to Seastar. You didn't add them - only those four which I deleted. Why?

My guess (please correct me if I'm wrong) is that for these specific four you wanted them to be listed in your code as "retryable" because the AWS driver which you reviewed said you should. But still doesn't mean you need them in Seastar - you can just use the number "419" in the Scylla code that wants to handle error 419 (which I claim will never happen because this error doesn't exist and isn't documented anywhere...) - it doesn't need to a constant in Seastar's header file.

nyh:
Pushed a new version with:
1. A comment in reply.hh saying where this list comes from and urging developers not to invent their own.
2. For the several errors whose official message has changed (like 413 changed from Payload Too Large to Content Too Large in the official RFC), I created both names as aliases - content_too_large and payload_too_large are now both allowed, and evaluate to 413. I didn't want to remove the old names because this will break valid old code (unlike 419 et al., code 413 is an actual standard status code, and existing applications may be putting it to good use).

nyh:
By the way, @elcallio @kreuzerkrieg if it will ease your objections that instead of outright deleting the 5 "bogus" (as I claim) error codes, I'll make them deprecated?

Note that if I mark them deprecated, your code that uses them will still need to be modified, unless you want to see those deprecation messages forever.

nyh:
I don't know what I broke, the http test now fails with illegal instruction (!?)
Really strange.

nyh:
> I don't know what I broke, the http test now fails with illegal instruction (!?) Really strange.

Turns out that commit 7a91062375 by @rockwotj changed the BOOST_ASSERT in tests by SEASTAR_ASSERT, and at the same time, made assertion failure call `__builtin_trap()`, causing SIGILL instead of the SIGQUIT. I didn't expect that...

nyh:
Pushed a new version, reverting my attempt to give some renamed constants both old and new news. For some reason, the http unit test failed after these renames - complaining about an unclosed gate, and I couldn't figure out why the names of the constants make any difference... So I just gave up on this specific changed.

xemul:
I like the change (sorry for taking too long to express that), but the title "complete list" confuses me. In the comment you reference the iana code registry, but there are codes in there that were not added by this patch, e.g. 423, 431, maybe others. Is there a reason for it, or did you just miss them?

As far as using the "non standard" codes in Scylla. We anyway switch through reply._status with default: clause and can statically cast it back to int and check some extras. Not extremely elegant, but we'll need to handle non-standard codes anyway.

nyh:
> I like the change (sorry for taking too long to express that), but the title "complete list" confuses me. In the comment you reference the iana code registry, but there are codes in there that were not added by this patch, e.g. 423, 431, maybe others. Is there a reason for it, or did you just miss them?

As I tried to explain in the commit message, I consider the **complete list** to be what is in RFC 9110 - the HTTP standard -, and nothing else. IANA's list contains additional stuff, including codes that were used in ancient RFCs and for various extensions of HTTP which are not part of the regular HTTP. As I mentioned in the commit message, in two cases where we listed codes that are listed by IANA but not RFC 9110 I decided to keep them (429 and 507), but I did **not** want to copy everything that IANA lists. I certainly don't want error code 418 just because it's listed in IANA :-)

Specifically 423, for example, it is defined in https://www.rfc-editor.org/rfc/rfc4918.html, it is WebDAV, an **extension** to the HTTP protocol. WebDAV introduces not just new error codes but also new HTTP methods (PROPFIND and others), which Seastar does not support. It is quite pointless to have half-support for this protocol.

> As far as using the "non standard" codes in Scylla. We anyway switch through reply._status with default: clause and can statically cast it back to int and check some extras. Not extremely elegant, but we'll need to handle non-standard codes anyway.

I agree. Seastar's naming of official, standard, HTTP error codes, should not prevent an application developer from adding their own non-standard error codes.

PR #2833, opened 2025-07-02
util: make SEASTAR_ASSERT() failure generate SIGABRT

### LINKED ISSUE #2832: Seastar assertion failure fails with SIGILL instead of SIGABRT
Traditionally in C and C++, assertion failures and their ilk generate the SIGABRT signal. All of `abort()`, `__assert_fail()`, and C++ `std::terminate()` all generate a SIGABRT.

Recently in commit 7a91062375c9b13456713708f10365a118a0d999 (@rockwotj), all calls to `assert()` in Seastar and `BOOST_ASSERT()` in tests were replaced by a new `SEASTAR_ASSERT` inplemented in `include/seastar/util/assert.hh`. This macro uses a new `assert_fail()` function:

```c++
[[noreturn]] void assert_fail(const char* msg, const char* file, int line, const char* func) {
    printf("%s:%u: %s: Assertion `%s` failed.\n", file, line, func, msg);
    __builtin_trap();
}
```

The function used here, `__builtin_trap()`, generates a SIGILL signal instead of SIGABRT.

While there's nothing particularly "wrong" with using SIGILL, it's really confusing because it's not what people expect to see caused by an assertion failure - it typically happens when some memory gets corrupted or code jumps into a data segment.

@rockwotj any reason why you used `__builtin_trap()` here instead of the traditional `abort()` or the C++ `std::terminate()`?
Any reason why we shouldn't change now it to `abort()` or `std::terminate()`?

I noticed this when looking at a test failure. I was really surprised (and to be honest, scared) to see a test crashing with a SIGILL. I thought it was a serious memory corruption or something - and didn't realize it was just a normal (even if unexpected) failure of a check that used to be a BOOST_ASSERT and was changed to the new SEASTAR_ASSERT.

nyh:
Recently commit 7a91062 introduced SEASTAR_ASSERT to replace assert() and BOOST_ASSERT() - but unlike the latters it ignores NDEBUG and unconditionally checks the condition.

When the new SEASTAR_ASSERT() fails, it called __builtin_trap(), which generates a SIGILL (Illegal Instruction) signal. This is surprising and counter-intutive, because C / C++ developers are used to assertion failures always generating SIGABRT, not SIGILL.

Fix this by calling std::terminate() instead of __builtin_trap().

Fixes #2832

### DIFF
--- a/src/util/log.cc
+++ b/src/util/log.cc
@@ -125,7 +125,7 @@ namespace internal {
 
 [[noreturn]] void assert_fail(const char* msg, const char* file, int line, const char* func) {
     printf("%s:%u: %s: Assertion `%s` failed.\n", file, line, func, msg);
-    __builtin_trap();
+    std::terminate();
 }
 
 void log_buf::free_buffer() noexcept {


rockwotj:
This is fine by me. Our internal assert in Redpanda uses SIGILL, but there is no specific reason that was chosen AFAIK

PR #2837, opened 2025-07-04
pollable_fd: Untangle shutdown()

xemul:
The call chain nowadays is

 pollable_fd::shutdown()
 engine -> reactor_backend::shutdown()
 pollable_fd_state -> file_desc -> file_desc::shutdown()

This patch removes reactor and its backend from this sequence.

Other pollable_fd methods are nowadays implemented likewise -- they forward the invocation to pollable_fd_state and the latter, if needed, call reactor backend. Shutdown doesn't need backend, but if it will, it's _state that will need to call backend.

### DIFF
--- a/include/seastar/core/internal/pollable_fd.hh
+++ b/include/seastar/core/internal/pollable_fd.hh
@@ -117,6 +117,7 @@ public:
     future<size_t> recvmsg(struct msghdr *msg);
     future<size_t> sendto(socket_address addr, const void* buf, size_t len);
     future<> poll_rdhup();
+    void shutdown(int how);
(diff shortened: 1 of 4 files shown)


avikivity:
io_uring supports async shutdown.

While it's a rare operation, I can see why we'd want to prefer everything to go via io_uring.

xemul:
This PR doesn't contradict that plan. If async shutdown will appear in iouring, the newly introduced `pollable_fd_state::shutdown` will just call `engine()._backend->shutdown(*this)`, just like it does for pretty much any other method it has

PR #2841, opened 2025-07-07
gate: make destructor virtual

bhalevy:
With clang++ / c++23 we see the following error:
```
include/c++/15/bits/unique_ptr.h:93:2: error: delete called on non-final 'seastar::named_gate' that has virtual functions but non-virtual destructor [-Werror,-Wdelete-non-abstract-non-virtual-dtor]
```

### DIFF
--- a/include/seastar/core/gate.hh
+++ b/include/seastar/core/gate.hh
@@ -109,7 +109,7 @@ public:
         }
         return *this;
     }
-    ~gate() {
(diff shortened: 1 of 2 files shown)


bhalevy:
ping @scylladb/seastar-maint please review and consider merging

avikivity:
This makes me dislike gates even more.

PR #2843, opened 2025-07-08
fair_queue: Ditch queue-wide accumulator reset on overflow

xemul:
The resetting code checks that priority class acumulator value stays below int64_t::max value and resets all accumulators on the io-queue classes if it breaks the threshold.

This check is never true and the guarded branch is dead code. Here's why. The accumulator value grows by request.capacity / class.shares values, so let's try to estimate the growth rate for the request with largest practical capacity and for class with 1 share.

The capacity of request is larger for slow disks and smaller for fast disks, because it's formula is 1/disk.iops + size/disk.bandwidth. Let's pick the disk with 100 IOPS and 100MB/s speed. The ioinfo tool would show the larges request capacity to be 377487360. The 64-bit signed integer would fit 24433591728 such requests which, at 100 requsts/sec rate, would overflow in 7 years.

More "practical" disk, having 300kIOPS and 1GB/s bandwidth. The fattest request capacity is 3038283, the overflow would happen in 11 years.

The "very fast" disk of 1MIOPS and 1TB/s bandwidth has a chance to overflow the counter in 1 year, provided a class with 1 sare runs on a single shard and loads the disk at full bandwidth all that time.

This code is here since the times when the accumulator was added with exponential-growing costs (see 9154b45403) and overflows were real.

### DIFF
--- a/src/core/fair_queue.cc
+++ b/src/core/fair_queue.cc
@@ -273,22 +273,12 @@ void fair_queue::pop_front() {
     // class show no progress and monopolize the queue.
     auto req_cap = req._capacity;
     auto req_cost  = std::max(req_cap / h._shares, (capacity_t)1);
-    // signed overflow check to make push_priority_class_from_idle math work
-    if (h._accumulated >= std::numeric_limits<signed_capacity_t>::max() - req_cost) {
-        for (auto& pc : _priority_classes) {
-            if (pc) {
-                if (pc->_queued) {
-                    pc->_accumulated -= h._accumulated;
-                } else { // this includes h
-                    pc->_accumulated = 0;
-                }
-            }
-        }
-        _last_accumulated = 0;
-    }
     h._accumulated += req_cost;
     h._pure_accumulated += req_cap;
     _queued_capacity -= req_cap;
+
+    // signed overflow check to make push_priority_class_from_idle math work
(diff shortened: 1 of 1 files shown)


robertbindar (commented):
simpler is better, nice.
Please add a comment saying something like `This cannot practically overflow in a reasonable amount of time, see the calculations in the commit message that prove the assert can't happen`.  

> The "very fast" disk of 1MIOPS and 1TB/s bandwidth has a chance to overflow the counter in 1 year, provided a class with 1 sare runs on a single shard and loads the disk at full bandwidth all that time.
what's the usual pattern here, can a class run on multiple shards so that the calculations here end up being a few months?

avikivity:
I guess if you scaled IOPS and bandwidth by exactly the same factor, the overflow time wouldn't change?

In a  sense, the overflow time is defined by the resolution we choose to represent the ticket.

xemul:
> I guess if you scaled IOPS and bandwidth by exactly the same factor, the overflow time wouldn't change?

Yes

> In a sense, the overflow time is defined by the resolution we choose to represent the ticket.

Yes. By the "fixed point factor" constant to be precise

xemul:
upd:
- rebased (no conflicts)
- extended the commit message text

PR #2845, opened 2025-07-08
Split random-io-buffer-size in 2 options

### LINKED ISSUE #2844: Iotune - split random-io-buffer-size option for each io direction
During experiments in #2802, we identified that `iotune` fails to detect the block size for i7i instances and the 512bytes default renders bad write IOPS numbers. Changing that to 4K improves write IOPS quite a lot, but tanks the read IOPS numbers. We need to be able to specify the buffer size for each workload type independently.

robertbindar:
This patch splits the current `--random-io-buffer-size` option in two:
- `--random-write-io-buffer-size`
- `--random-read-io-buffer-size`
so that each workload type can be configured independently.

During experiments in https://github.com/scylladb/seastar/issues/2802, we identified that iotune fails to detect the block size for i7i instances and the 512bytes default renders bad write IOPS numbers. Changing that to 4K improves write IOPS quite a lot, but tanks the read IOPS numbers. We need to be able to specify the buffer size for each workload type independently.

Fixes #2844

### DIFF
--- a/apps/iotune/iotune.cc
+++ b/apps/iotune/iotune.cc
@@ -419,11 +419,13 @@ public:
 class test_file {
 public:
     enum class pattern { sequential, random };
+    enum class access_type { read, write };
 private:
     fs::path _dirpath;
     uint64_t _file_size;
     file _file;
-    uint64_t _forced_random_io_buffer_size;
+    uint64_t _forced_random_write_io_buffer_size;
+    uint64_t _forced_random_read_io_buffer_size;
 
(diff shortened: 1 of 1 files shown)


robertbindar:
@xemul please note I renamed the original option too, if we care about maintaining that, I can leave it and only add a new one for reads.

bhalevy:
> This patch splits the current `--random-io-buffer-size` option in two:
> 
> * `--random-write-io-buffer-size`
> * `--random-read-io-buffer-size`
>   so that each workload type can be configured independently.
> 
> See the ref'd issue below for motivation.

@robertbindar we need to document the motivation in the patch itself (echoed in the PR description)

PR #2846, opened 2025-07-08
scripts/seastar-addr2line: change the default addr2line utility to llvm-addr2line

### LINKED ISSUE #2842: Gnu addr2line doesn't decode dwarf generated by clang 20 properly
Spun off of https://github.com/scylladb/scylladb/issues/24217#issuecomment-3044781111

@michoecho wrote:

> The backtrace doesn't look genuine. It's missing many inlined frames. `create_write_response_handler` doesn't directly call `seastar::memory::allocate_slowpath`.
>
> The file:line information is actually fine. The problem is that the decode is missing inlined frames.
> That is, if you have
> ```
>             service::storage_proxy::create_write_response_handler(...) at ././seastar/src/core/memory.cc:1706
> ```
> then this return address really is at line `././seastar/src/core/memory.cc:1706`, which is `seastar::memory::allocate`, which is the topmost "inlined frame" in the physical `service::storage_proxy::create_write_response_handler` frame.
> 
> `llvm-addr2line` and `lldb` decode the inlined frames correctly. `gdb` and `addr2line` don't.
> 
> I guess the upgrade to clang 20 changed the DWARF in a way such that GNU utils can't decode it properly.
> 
> When decoding the backtraces, you can use llvm-addr2line as a workaround.
> (And llvm-addr2line is better anyway because it's faster).

bhalevy:
The upgrade to clang 20 changed the DWARF in a way such that binutils
addr2line can't decode it properly.
    
Michał Chojnowski <michal.chojnowski@scylladb.com> explains:
> clang encodes the address ranges (DW_AT_ranges) for the missing inlined
> frames with a dwarf 5 construct (DW_FORM_rnglistx) which just hasn't been
> implemented in libbfd (which is what addr2line uses for decoding).
> ...
> So addr2line basically skips over those DW_TAG_inlined_subroutine for
> the purposes of address decoding. They are not inserted into the address maps.
> That's why they don't show up.
>
> So it's more a missing feature in bfd than a bug.

Fixes #2842

### DIFF
--- a/scripts/seastar-addr2line
+++ b/scripts/seastar-addr2line
@@ -450,9 +450,9 @@ There are three operational modes:
         '--addr2line',
         type=str,
         metavar='CMD_PATH',
-        default='addr2line',
+        default='llvm-addr2line',
         help='The path or name of the addr2line command, which should behave as and '
-        'accept the same options as binutils addr2line or llvm-addr2line.',
+        'accept the same options as binutils addr2line or llvm-addr2line (the default).',
     )
 
     cmdline_parser.add_argument(


avikivity:
Did you report the issue to binutils?

bhalevy:
> Did you report the issue to binutils?

Not yet.  i asked @michoecho to follow up on this since he's more familiar with the details.

michoecho:
Do we want to just dump a nightly Scylla binary on the addr2line devs, or do we want to find a minimized reproducer? (Or do we want to debug addr2line ourselves and report a detailed bug explanation?)

bhalevy:
> Do we want to just dump a nightly Scylla binary on the addr2line devs, or do we want to find a minimized reproducer? (Or do we want to debug addr2line ourselves and report a detailed bug explanation?)

A minimized reproducer would be better.
From past experience with compiler issues involving seastar, they would ask for a distilled reproducer.

Let's refrain from debugging addr2line ourselves.

michoecho:
> Let's refrain from debugging addr2line ourselves.

It have a feeling it would be easier than minimizing the reproducer, though. Mostly because it doesn't involve recompiling Scylla code.

avikivity:
> > Let's refrain from debugging addr2line ourselves.
> 
> It have a feeling it would be easier than minimizing the reproducer, though. Mostly because it doesn't involve recompiling Scylla code.

It's okay to spend a few hours on it, but eject if it turns into a time waster.

michoecho:
> > > Let's refrain from debugging addr2line ourselves.
> > 
> > 
> > It have a feeling it would be easier than minimizing the reproducer, though. Mostly because it doesn't involve recompiling Scylla code.
> 
> It's okay to spend a few hours on it, but eject if it turns into a time waster.

I spent 3 hours on it and determined the cause: clang encodes the address ranges (`DW_AT_ranges`) for the missing inlined frames with a dwarf 5 construct (`DW_FORM_rnglistx`) which just hasn't been implemented in libbfd (which is what addr2line uses for decoding).

See [this](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=bfd/dwarf2.c;h=a62c952374378a39c8d7896e0728ff9581ada349;hb=HEAD#l1658) (bfd can't even parse this form) and [this](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=bfd/dwarf2.c;h=a62c952374378a39c8d7896e0728ff9581ada349;hb=HEAD#l4172) (bfd only accepts integer forms for the DW_AT_ranges field).

So addr2line basically skips over those `DW_TAG_inlined_subroutine` for the purposes of address decoding. They are not inserted into the address maps. That's why they don't show up.

So it's more a missing feature in bfd than a bug.

But I support switching to llvm-addr2line anyway, even if this eventually gets implemented in libbfd, because llvm-addr2line is 20x faster.

michoecho:
> gdb and addr2line don't.

After knowing what's the problem, I think it's only a problem in addr2line, not in gdb. I misunderstood gdb's output.

bhalevy:
> > > > Let's refrain from debugging addr2line ourselves.
> > > 
> > > 
> > > It have a feeling it would be easier than minimizing the reproducer, though. Mostly because it doesn't involve recompiling Scylla code.
> > 
> > 
> > It's okay to spend a few hours on it, but eject if it turns into a time waster.
> 
> I spent 3 hours on it and determined the cause: clang encodes the address ranges (`DW_AT_ranges`) for the missing inlined frames with a dwarf 5 construct (`DW_FORM_rnglistx`) which just hasn't been implemented in libbfd (which is what addr2line uses for decoding).
> 
> See [this](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=bfd/dwarf2.c;h=a62c952374378a39c8d7896e0728ff9581ada349;hb=HEAD#l1658) (bfd can't even parse this form) and [this](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=bfd/dwarf2.c;h=a62c952374378a39c8d7896e0728ff9581ada349;hb=HEAD#l4172) (bfd only accepts integer forms for the DW_AT_ranges field).
> 
> So addr2line basically skips over those `DW_TAG_inlined_subroutine` for the purposes of address decoding. They are not inserted into the address maps. That's why they don't show up.
> 
> So it's more a missing feature in bfd than a bug.
> 
> But I support switching to llvm-addr2line anyway, even if this eventually gets implemented in libbfd, because llvm-addr2line is 20x faster.

Thanks for the quick analysis!
Is there an open issue with binutils to implement dwarf 5?

michoecho:
> Thanks for the quick analysis!
Is there an open issue with binutils to implement dwarf 5?

1. Note that this is about bfd, not about binutils in general. Different parts of binutils have their own dwarf-handling code.
2. I don't think they do "epics" on the bugzilla. Parts of dwarf 5 are implemented case by case as needed. There will probably be no interest in an umbrella "dwarf 5" ticket.
3. I don't see any ticket about adding support for DW_FORM_rnglistx to addr2line.

bhalevy:
> > Thanks for the quick analysis!
> > Is there an open issue with binutils to implement dwarf 5?
> 
> 1. Note that this is about bfd, not about binutils in general. Different parts of binutils have their own dwarf-handling code.
> 2. I don't think they do "epics" on the bugzilla. Parts of dwarf 5 are implemented case by case as needed. There will probably be no interest in an umbrella "dwarf 5" ticket.
> 3. I don't see any ticket about adding support for DW_FORM_rnglistx to addr2line.

@michoecho let's be good citizens and open one please

travisdowns:
As a datapoint we are using llvm-addr2line-20 in our CI in concert with seastar-addr2line and are happy with the speed and DWARF support.

bhalevy:

bhalevy:
[9892b00](https://github.com/scylladb/seastar/commit/9892b0071676d8fe7625c5e6f965c84aff4dc12d): updated commit log message based on @michoecho's review [comment](https://github.com/scylladb/seastar/pull/2846#issuecomment-3053564054).

avikivity:
I want a bug filed for bfd.

michoecho:
bfd ticket: https://sourceware.org/bugzilla/show_bug.cgi?id=33154

PR #2848, opened 2025-07-09
core: Don't keep internal errors counter on reactor

xemul:
Sometimes the on_internal_error set of helpers is called without starting seastar app, thus the engine is not initialized. To allow seastar library users still use the internal error machinery (and preserve potential bugwards compatibility) keep the counter as global thread local variable.

### DIFF
--- a/src/core/on_internal_error.cc
+++ b/src/core/on_internal_error.cc
@@ -33,7 +33,6 @@ module;
 module seastar;
 #else
 #include <seastar/core/on_internal_error.hh>
-#include <seastar/core/reactor.hh>
 #include <seastar/util/backtrace.hh>
 #include <seastar/util/log.hh>
 #endif
@@ -46,8 +45,11 @@ bool seastar::set_abort_on_internal_error(bool do_abort) noexcept {
     return abort_on_internal_error.exchange(do_abort);
 }
 
-void internal::increase_internal_errors_counter() noexcept{
-    seastar::engine()._internal_errors++;
+namespace seastar::internal {
+thread_local uint64_t internal_errors = 0;
+void increase_internal_errors_counter() noexcept {
+    internal_errors++;
+}
 }
 
 template <typename Message>
--- a/include/seastar/core/on_internal_error.hh
+++ b/include/seastar/core/on_internal_error.hh
@@ -72,5 +72,9 @@ void on_internal_error_noexcept(logger& logger, std::string_view reason) noexcep
 /// This overload can be used to replace SEASTAR_ASSERT().
 [[noreturn]] void on_fatal_internal_error(logger& logger, std::string_view reason) noexcept;
 
+namespace internal {
+extern thread_local uint64_t internal_errors;
+}
+
 SEASTAR_MODULE_EXPORT_END
 }
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -263,7 +263,6 @@ private:
     io_stats _io_stats;
     uint64_t _fsyncs = 0;
     uint64_t _cxx_exceptions = 0;
-    uint64_t _internal_errors = 0;
     uint64_t _abandoned_failed_futures = 0;
     struct task_queue {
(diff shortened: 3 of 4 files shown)


xemul:
@denesb , this problem is stepped on here: https://github.com/scylladb/scylladb/issues/24849

src/core/on_internal_error.cc
```diff
@@ -46,8 +45,11 @@ bool seastar::set_abort_on_internal_error(bool do_abort) noexcept {
     return abort_on_internal_error.exchange(do_abort);
 }
 
-void internal::increase_internal_errors_counter() noexcept{
-    seastar::engine()._internal_errors++;
+namespace seastar::internal {
+thread_local uint64_t internal_errors = 0;
+void increase_internal_errors_counter() noexcept {
+    internal_errors++;
```
avikivity, line 51: This can produce really bad code if the compiler thinks the variable can be declared elsewhere.

I don't think this is the case here, but please check the increment translates to one instruction and not a guard variable check followed by initialization.

The antidote is to declare the variable as constinit in .hh; see eaa00e761fc5f6694fde58e72e68ffd62c6a3982 for an example.

(not that it matters since internal errors should be rare, but...)

xemul: gcc 14.2.1

```
000000000000000e <_ZN7seastar8internal32increase_internal_errors_counterEv>:
   e:   48 8b 05 00 00 00 00    mov    0x0(%rip),%rax        # 15 <_ZN7seastar8internal32increase_internal_errors_counterEv+0x7>
                        11: R_X86_64_GOTTPOFF   _ZN7seastar8internal15internal_errorsE-0x4
  15:   64 48 83 00 01          addq   $0x1,%fs:(%rax)
  1a:   c3                      ret
```

clang 19.1.7

```
0000000000000010 <_ZN7seastar8internal32increase_internal_errors_counterEv>:
  10:   48 8b 05 00 00 00 00    mov    0x0(%rip),%rax        # 17 <_ZN7seastar8internal32increase_internal_errors_counterEv+0x7>
                        13: R_X86_64_GOTTPOFF   _ZN7seastar8internal15internal_errorsE-0x4
  17:   64 48 ff 00             incq   %fs:(%rax)
  1b:   c3                      ret
  1c:   0f 1f 40 00             nopl   0x0(%rax)
```

xemul: clang 20.1.7 (scylla dbuild)

```
0000000000000010 <_ZN7seastar8internal32increase_internal_errors_counterEv>:
  10:   48 8b 05 00 00 00 00    mov    0x0(%rip),%rax        # 17 <_ZN7seastar8internal32increase_internal_errors_counterEv+0x7>
                        13: R_X86_64_GOTTPOFF   _ZN7seastar12local_engineE-0x4
  17:   64 48 8b 00             mov    %fs:(%rax),%rax
  1b:   48 ff 80 00 17 00 00    incq   0x1700(%rax)
  22:   c3                      ret
  23:   66 2e 0f 1f 84 00 00    cs nopw 0x0(%rax,%rax,1)
  2a:   00 00 00 
  2d:   0f 1f 00                nopl   (%rax)

```

regevran: Can `increase_internal_errors_counter` be inlined resulting in a single instruction?

xemul: With compiler trying hard enough -- why not. But I don't think we care optimizations here _that_ much. Internal errors are not supposed to happen often

regevran: So why do you so care for:

> please check the increment translates to one instruction and not a guard variable check followed by initialization.

xemul: I was curious enough to check

regevran: AND :question:

avikivity: Still one wasted instruction.

https://godbolt.org/z/96K975bhz

PR #2849, opened 2025-07-09
Nested scheduling groups (IO classes)

xemul:
Add one-level nesting support for IO priority classes. Since all shares-based scheduling of IO happens inside fair_queue code, this PR mainly updates it. The io_queue code is touched to propagate scheduling_supergroup hierarchy and changes down to the fair_queue. This change follows those made for CPU task_queues.

Currently, the scheduling_groups is backed by reactor::task_queue object for CPU scheduler and by fair_queue::priority_class_data for IO. For nesting, the scheduling_supergroup was added (see #2823) at the API level. It was then backed by the reactor::task_queue_group in the CPU scheduler, where both task_queue and task_queue_group inherit from sched_entity abstraction. On the IO side supergroup maps to fair_queue::priority_class_group_data and both, priority_class_data and _group_data inherit from fair_queue::priority_entry abstraction.

At the IO-queue level itself, as was said, the only change made is to construct the priority_class_data and _group_data hierarchy upon class creation. On this level the priority_class_data also exists and corresponds to scheduling_group. This object is used to reference fair_queue class data, track IO statistics, and -- the IO bandwidth limiting. Unlike fair_queue level, IO queue level does _not_ gain its _group_data, which in turn means, that IO statistics for supergroups is not available, IO bandwidth limiting for supergroups doesn't appear either (a (in)visible sign of it is that update_io_bandwidth() is the method of scheduling_group, not scheduling_supergroup).

### DIFF
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -39,6 +39,10 @@ namespace bi = boost::intrusive;
 
 namespace seastar {
 
+namespace testing {
+class fair_queue_test;
+}
+
 /// \brief describes a request that passes through the \ref fair_queue.
 ///
 /// A ticket is specified by a \c weight and a \c size. For example, one can specify a request of \c weight
@@ -164,6 +168,7 @@ private:
 
     class priority_entry {
         friend class fair_queue;
+        friend testing::fair_queue_test;
     protected:
         uint32_t _shares = 0;
         capacity_t _accumulated = 0;
@@ -200,6 +205,7 @@ private:
 
     class priority_class_group_data final : public priority_entry {
         friend class fair_queue;
+        friend testing::fair_queue_test;
         priority_queue _children;
         capacity_t _last_accumulated = 0;
         size_t _nr_children = 0;
@@ -229,6 +235,7 @@ private:
     config _config;
     priority_class_group_data _root;
     std::vector<std::unique_ptr<priority_class_data>> _priority_classes;
+    friend testing::fair_queue_test;
 
     // Total capacity of all requests waiting in the queue.
     capacity_t _queued_capacity = 0;
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -185,6 +185,7 @@ private:
     class priority_class_group_data final : public priority_entry {
         friend class fair_queue;
         priority_queue _children;
+        capacity_t _last_accumulated = 0;
     };
 
     class priority_class_data;
@@ -193,7 +194,6 @@ private:
     priority_class_group_data _root;
     std::vector<std::unique_ptr<priority_class_data>> _priority_classes;
     size_t _nr_classes = 0;
-    capacity_t _last_accumulated = 0;
 
     // Total capacity of all requests waiting in the queue.
     capacity_t _queued_capacity = 0;
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -164,6 +164,9 @@ private:
         friend class fair_queue;
     protected:
         uint32_t _shares = 0;
+        capacity_t _accumulated = 0;
+        bool _queued = false;
+        uint32_t _activations = 0;
         priority_entry(uint32_t shares) noexcept
                 : _shares(std::max(shares, 1u))
         {}
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -162,6 +162,11 @@ public:
 private:
     class priority_entry {
         friend class fair_queue;
+    protected:
+        uint32_t _shares = 0;
+        priority_entry(uint32_t shares) noexcept
+                : _shares(std::max(shares, 1u))
+        {}
     };
 
     using clock_type = std::chrono::steady_clock;
@@ -186,6 +191,13 @@ private:
         friend class fair_queue;
         priority_queue _children;
         capacity_t _last_accumulated = 0;
+    public:
+        priority_class_group_data(uint32_t shares) noexcept
+                : priority_entry(shares)
+        {
+        }
+        priority_class_group_data(const priority_class_group_data&) = delete;
+        priority_class_group_data(priority_class_group_data&&) = delete;
     };
 
     class priority_class_data;
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -201,6 +201,10 @@ private:
         }
         priority_class_group_data(const priority_class_group_data&) = delete;
         priority_class_group_data(priority_class_group_data&&) = delete;
+
+        void reserve(size_t len) {
+            _children.reserve(len);
+        }
     };
 
     class priority_class_data;
--- a/include/seastar/core/fair_queue.hh
+++ b/include/seastar/core/fair_queue.hh
@@ -234,6 +234,7 @@ private:
 
     config _config;
     priority_class_group_data _root;
+    std::vector<std::unique_ptr<priority_class_group_data>> _priority_groups;
     std::vector<std::unique_ptr<priority_class_data>> _priority_classes;
     friend testing::fair_queue_test;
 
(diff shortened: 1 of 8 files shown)


xemul:
upd:
- fix wakeup not to go above root group

xemul:
upd:
- rebased

avikivity:
Need description of large-scale testing done to validate.

xemul:
upd:
- rebased
- added linkage test
- fixed class-group to root linkage counters
- renamed _nr_classes to _nr_children
- spawned some fair-queue test fixes to build nesting testing on top (#2904, #2905, #2907, #2908)

xemul:
upd:
- cleanup sched groups in existing io_queue tests

xemul:
upd:
- added more fair_queue tests (similarly to CPU nesting ones -- random fairness and wakeups)

xemul:
> Need description of large-scale testing done to validate.

* Existing fair-queue tests (slightly extended with #2905). They test single level of groups, but since the very same logic exists on all levels, that seem to be enough
* Modified by #2907 random-run test checks the outcome proportions of 3 classes with random shares randomly nested that process requests or random weights
* Exhaustive wake-up test that validates activation of 1...5 requests through 2-level hierarchy of classes
* Small IO-queue-level test that validates linkage of fair_queue classes based on scheduling_groups/_supergroups hierarchy

xemul:
upd:
- rebased to pick up needed unit test changes
- udpated PR comment

xemul:
@avikivity , please review

avikivity:
Due to low confidence in my ability to provide a good review, I'd like a system test to be performed and reported in the cover letter. I think a mix of an io_tester run on a hefty instance (one that at least owns a full disk) to show correctness and a scylladb run to show no regressions (without nested scheduling engaged).

I'd like the io_tester run to have a group activate and deactivate (say once per minute) and some analysis of the results.

avikivity:
(I will of course also review it)

src/core/fair_queue.cc
```diff
@@ -246,7 +246,7 @@ void fair_queue::notify_request_cancelled(fair_queue_entry& ent) noexcept {
 
 fair_queue_entry* fair_queue::top() {
     while (!_handles.empty()) {
-        priority_class_data& h = *_handles.top();
+        priority_class_data& h = *reinterpret_cast<priority_class_data*>(_handles.top());
```
avikivity: static_cast is safer (won't allow casting to an unrelated type, and adjusts the pointer if needed).

include/seastar/core/fair_queue.hh
```diff
@@ -186,6 +191,13 @@ private:
         friend class fair_queue;
         priority_queue _children;
         capacity_t _last_accumulated = 0;
+    public:
+        priority_class_group_data(uint32_t shares) noexcept
```
avikivity: explicit

src/core/fair_queue.cc
```diff
@@ -118,7 +115,7 @@ class fair_queue::priority_class_data final : public priority_entry {
 };
 
 bool fair_queue::class_compare::operator() (const priority_entry_ptr& lhs, const priority_entry_ptr& rhs) const noexcept {
-    return reinterpret_cast<priority_class_data*>(lhs)->_accumulated > reinterpret_cast<priority_class_data*>(rhs)->_accumulated;
+    return lhs->_accumulated > rhs->_accumulated;
```
avikivity, line 128: hmm

src/core/fair_queue.cc
```diff
@@ -202,7 +202,7 @@ void fair_queue::register_priority_class(class_id id, uint32_t shares) {
         SEASTAR_ASSERT(!_priority_classes[id]);
     }
 
-    _root._children.reserve(_nr_classes + 1);
+    _root.reserve(_nr_classes + 1);
```
avikivity: Don't see how reservations help in this control path, or when it's increased by exactly 1 (which would be allocated anyway). But no point in changing it.

src/core/fair_queue.cc
```diff
@@ -269,30 +277,35 @@ fair_queue_entry* fair_queue::priority_class_group_data::top() {
...
     // signed overflow check to make push_priority_class_from_idle math work
     SEASTAR_ASSERT(h._accumulated < std::numeric_limits<signed_capacity_t>::max() - req_cost);
 
-    if (!h._queue.empty()) {
-        push_priority_class(h);
+    if (empty) {
+        h._queued = false;
+    } else {
+        _children.push(&h);
     }
```
avikivity, line 309: This pop-then-push is equivalent to a heap increase-key operation, yes? Some heaps have O(1) operations for that. Don't know it is matters, our heaps are usually very small.

avikivity:
Looks good and relatively straightforward.

Idea for test:

ssg1 (shares1) { sg1A (shares2) sg1B (shares3) }
ssg2 (shares4) { sg2A (shares5) /* shares5 doesn't affect anything */ }

run a square-wave workload on sg1A, sg1B, sg2A with frequencies f1A, f1B, f2A. Frequency is on the order of 0.01 Hz.

We will see different combinations of on-time for the different workloads (2^3 combinations) and can check that the I/O distribution is as expected and that the transitions are smooth.

src/core/fair_queue.cc
xemul: It's temporary and all casts are gone in this PR. But I'll fix, thanks

src/core/fair_queue.cc
xemul: It's needed to have _handles large enough for class pointers to be assigned into its elements with operator[] and without extra re-allocations. In CPU scheduler it's missing, because the active queue is array with fixed capacity. I don't know why io-queue uses vector for that, maybe "just because"

src/core/fair_queue.cc
xemul: Yes, push-pop is to balance the heap's element with adjusted accumulator value (sorting key). The heap is nr-scheduling-groups-long. And the used std::priority_queue doesn't have any ways to rebalance elements.

xemul:
> I'd like the io_tester run to have a group activate and deactivate (say once per minute) and some analysis of the results.

As far as ... pointer-safety (?) is concerned, there's [this](https://github.com/scylladb/seastar/pull/2849/commits/863922d8b0ddbc2c1c524f410c0d6058f8ec2371) test here. As for fairness-correctness -- I'll try to come up with something

src/core/fair_queue.cc
avikivity: Yes, I saw later.

src/core/fair_queue.cc
avikivity: Ok.

xemul:
upd:
- rebased
- added a unit test to verify the wake-from-idle forgiving logic in the nested case

xemul:
upd:
- rebased
- extended the "forgiving group" test case to activate classes in different orders

xemul:
Ran ran backup test against this PR and base master branches. Summary: on "this PR" job nodes got better balanced that resulted in lower bandwidth values, but in the end of the day test measured no differences in write latencies and backup throughputs.

* Base
image

image


image

image

image

image


* This PR
image

image


image

image

image

image

avikivity:
> Looks good and relatively straightforward.
> 
> Idea for test:
> 
> ssg1 (shares1) { sg1A (shares2) sg1B (shares3) } ssg2 (shares4) { sg2A (shares5) /* shares5 doesn't affect anything */ }
> 
> run a square-wave workload on sg1A, sg1B, sg2A with frequencies f1A, f1B, f2A. Frequency is on the order of 0.01 Hz.
> 
> We will see different combinations of on-time for the different workloads (2^3 combinations) and can check that the I/O distribution is as expected and that the transitions are smooth.

Did you run a similar test?

xemul:
I did, but it turned out to be a combination of [random run](https://github.com/scylladb/seastar/pull/2849/commits/eff8a4796665d7ec786cccffd325614b48b068b9) with [wakeups](https://github.com/scylladb/seastar/pull/2849/commits/87cb4400855964510a3b564e40b94956de3b3802). And to check its correctness there were two options:
1. run the test for long enough time and see that resource consumption is consistent with shares. In that sense the test was duplication of the former from the above
2. run the test for shorter time and try to check that resource consumption is consistent with shares _and_ forgiveness factor. Here the "expected consumed resources" were manually calculated like in the latter test from the above

PR #2853, opened 2025-07-11
addr2line timing and optimizations

travisdowns:
We sometimes parse large logfiles thorugh addr2line and noted that it can be quite slow and consume a large amount of memory. E.g., 3-15 minutes (the high end on memory constrained host) and 2.1 GB RSS in order to parse a 1.6 GB file.

The changes here improve that to 100 MB (probably almost all in the child addr2line process) and ~7s parsing time (split 6/1 in the Python code and addr2line itself).

Key perf changes are to not load the entire file into a `list` up front and to do a quick filter on each line to see if we should even bother applying the regexes. Details in individual commits.

### Performance and Timing Enhancements:
* [`scripts/addr2line.py`](diffhunk://#diff-cd6a2af470efe68703f2ffd367c7b7e260f28ced194949fb1571a53d7d5f4ab8R356-R360): Added timing functionality to measure and log the duration of address resolution and subprocess execution. Introduced methods like `timing_now`, `timing_print`, and `print_resolve_time` for timing-related operations. Updated `resolve_address` to track and accumulate resolution time. [[1]](diffhunk://#diff-cd6a2af470efe68703f2ffd367c7b7e260f28ced194949fb1571a53d7d5f4ab8R356-R360) [[2]](diffhunk://#diff-cd6a2af470efe68703f2ffd367c7b7e260f28ced194949fb1571a53d7d5f4ab8R386-R398) [[3]](diffhunk://#diff-cd6a2af470efe68703f2ffd367c7b7e260f28ced194949fb1571a53d7d5f4ab8L397-R426)

### Regex Optimization:
* [`scripts/addr2line.py`](diffhunk://#diff-cd6a2af470efe68703f2ffd367c7b7e260f28ced194949fb1571a53d7d5f4ab8L271-R281): Added a quick upfront check in the `__call__` method to filter lines without address patterns, significantly improving regex matching performance.

### Type Annotations and Test Enhancements:
* [`scripts/seastar-addr2line`](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caL75-R74): Added type annotations to test data structures (`TestList`) in the `TestStringMethods` class for clarity and improved code reliability. [[1]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caL75-R74) [[2]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caL126-R125) [[3]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caL147-R146) [[4]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caL192-R191) [[5]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caL209-R208) [[6]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caL245-R244) [[7]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caL288-R287) [[8]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caL325-R324)

### Operational Mode Enhancements:
* [`scripts/seastar-addr2line`](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caR470-R473): Added a new `--timing` command-line argument to enable timing information output. Updated the main loop to log the total resolution time and the duration of the full resolve loop. [[1]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caR470-R473) [[2]](diffhunk://#diff-daf58e1efe76abbb947170e8009efedaf9fd1e27443351a6636c062a1a98c2caR514-R520)

### DIFF
--- a/scripts/addr2line.py
+++ b/scripts/addr2line.py
@@ -268,7 +268,7 @@ class BacktraceResolver:
                 addresses.append({'path': m.group(1) or default_path, 'addr': m.group(2)})
             return addresses
 
-        def __call__(self, line: str):
+        def __call__(self, line: str) -> dict[str, Any] | None:
 
             def get_prefix(s: Optional[str]):
                 if s is not None:
--- a/scripts/addr2line.py
+++ b/scripts/addr2line.py
@@ -25,6 +25,7 @@ import sys
 import subprocess
 from enum import Enum
 from functools import cache
+import time
 from typing import Any, Optional, TypeVar, Union, cast
 
 # special binary path/module indicating that the address is from the kernel
@@ -353,8 +354,11 @@ class BacktraceResolver:
         concise: bool = False,
         cmd_path: str = 'addr2line',
         debug: bool = False,
+        timing: bool = False,
     ):
         self._debug = debug
+        self._timing = timing
+        self._total_resolve_time = 0.0
         self._executable = executable
         self._kallsyms = kallsyms
         self._current_backtrace: list[tuple[str, str]] = []
@@ -380,6 +384,19 @@ class BacktraceResolver:
         if self._debug:
             print('DEBUG >>', *args, file=sys.stderr)
 
+    def timing_now(self):
+        return time.perf_counter() if self._timing else 0.0
+
+    def timing_print_from_start(self, start: float, *args: Any):
+        self.timing_print(self.timing_now() - start, *args)
+
+    def timing_print(self, duration: float, *args: Any):
+        if self._timing:
+            print(f"TIMING >> {duration:.4f} seconds : ", *args, file=sys.stderr)
+
+    def print_resolve_time(self):
+        self.timing_print(self._total_resolve_time, 'resolve time (addr2line subprocess time)')
+
     def _get_resolver_for_module(self, module: str):
         if not module in self._known_modules:
             if module == KERNEL_MODULE:
@@ -404,7 +421,10 @@ class BacktraceResolver:
             module = self._executable
         if verbose is None:
             verbose = self._verbose
-        resolved_address = self._get_resolver_for_module(module)(address)
+        resolver = self._get_resolver_for_module(module)
+        resolve_start = self.timing_now()
+        resolved_address = resolver(address)
+        self._total_resolve_time += self.timing_now() - resolve_start
         if verbose:
             resolved_address = '{{{}}} {}: {}'.format(module, address, resolved_address)
         return resolved_address
--- a/scripts/addr2line.py
+++ b/scripts/addr2line.py
@@ -241,6 +241,12 @@ class BacktraceResolver:
             token = fr"(?:{path}\+)?{addr}"
             full_addr_match = fr"(?:(?P<path>{path})\s*\+\s*)?(?P<addr>{addr})"
             ignore_addr_match = fr"(?:(?P<path>{path})\s*\+\s*)?(?:{addr})"
+
+            self.asan_ignore_re = re.compile(f"^=.*$", flags=re.IGNORECASE)
+            self.separator_re = re.compile(r'^\W*-+\W*$')
+
+            # All of the below except for separator_re must only match lines containing 0x in them
+            # as we use that as a quick first filter
             self.oneline_re = re.compile(
                 fr"^((?:.*(?:(?:at|backtrace):?|:))?(?:\s+))?({token}(?:\s+{token})*)(?:\).*|\s*)$",
                 flags=re.IGNORECASE,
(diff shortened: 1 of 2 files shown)


scripts/addr2line.py
```diff
@@ -268,8 +272,14 @@ def split_addresses(self, addrstring: str, default_path: Optional[str] = None):
                 addresses.append({'path': m.group(1) or default_path, 'addr': m.group(2)})
             return addresses
 
-        def __call__(self, line: str):
+        def __call__(self, line: str) -> dict[str, Any] | None:
 
+            # quick up front check to eliminate a line from contention, which is at least
+            # 30x faster than just diving into the regex matching (for one test file)
+            if not ("0x" in line or "0X" in line or self.separator_re.match(line)):
```
travisdowns, line 280: This should be a behavior preserving transformation based on my audit of the regexes. I don't think Seastar ever emits backtraces with `0X` in them (as opposed to `0x`) but I included it as some regexes use `IGNORECASE`, probably for other parts of the regex, but omitting it would technically be a change in behavior for those ones.

travisdowns:
Example `--timing` output:

```
TIMING >> 5.7180 seconds :  full resolve loop
TIMING >> 1.2310 seconds :  resolve time (addr2line subprocess time)
```

travisdowns:
@tchaikov do you think you could take an initial look at this or suggest someone who could?

travisdowns:
@bhalevy I think you've been in this file recently, do you think you could take a peek?

scripts/addr2line.py
bhalevy: Not sure how much it may speed things up but since `asan_ignore_re` is also used to ignore lines, we can easily check here also `if line.startswith('=') or not ("0x" in line or "0X" in line or self.separator_re.match(line)):`

scripts/addr2line.py
```diff
@@ -241,6 +241,9 @@ def __init__(self):
             token = fr"(?:{path}\+)?{addr}"
             full_addr_match = fr"(?:(?P<path>{path})\s*\+\s*)?(?P<addr>{addr})"
             ignore_addr_match = fr"(?:(?P<path>{path})\s*\+\s*)?(?:{addr})"
+
+            # all of these regex except for separator_re must only match lines containing 0x in them
```
bhalevy: `asan_ignore_re` does not match an address as well

bhalevy:
looks good overall

scripts/addr2line.py
```diff
@@ -25,6 +25,7 @@
 import subprocess
 from enum import Enum
 from functools import cache
+import time
```
xemul, line 28: This is comment to the patch as a whole: please, include sample output in the patch comment text

scripts/addr2line.py
```diff
@@ -270,6 +273,12 @@ def split_addresses(self, addrstring: str, default_path: Optional[str] = None):
 
         def __call__(self, line: str) -> dict[str, Any] | None:
 
+            # quick up front check to eliminate a line from contention, which is at least
+            # 30x faster than just diving into the regex matching (for one test file)
+            if not ("0x" in line or "0X" in line or self.separator_re.match(line)):
+                # no addresses in this line, so it is not a backtrace line
+                return None
```
avikivity, line 282: Very discouraging that we have to bypass re, regular expressions are supposed to be fast.

avikivity (approved):

scripts/addr2line.py
travisdowns: > Very discouraging that we have to bypass re, regular expressions are supposed to be fast.

Most sources seem to claim that python RE are generally considered pretty fast out of the universe of RE engines, but overall these kind of wildcard/unanchored matches are still slow compared to a plain strstr type search: but another factor is we have like 7 REs we are trying to match serially, whereas here we do we fewer searches which can rule them all out.

scripts/addr2line.py
travisdowns: I had included it here:

https://github.com/scylladb/seastar/pull/2853#issuecomment-3060050955

By "patch comment text" comment do you mean the PR cover letter, or commit message or something else?

scripts/addr2line.py
travisdowns: Unless your file was very unusual anything added to the check basically slows it down, since this executes on 99.9% of lines, wheras the regex only on the remaining 0.01%. So unless you file had a very large number of lines starting with `=` this would be worse.

Because of this effect I only included the things that were needed for correctness (a shame about the separator re).

scripts/addr2line.py
travisdowns: In any case in push https://github.com/scylladb/seastar/commit/c09bf2894517aa0d04ff746cd2be0f3f1a96bed3 I added it to the associated commit message.

scripts/addr2line.py
travisdowns: Good point, I moved the two non address regexs above this command clarified it applies only to the below in https://github.com/scylladb/seastar/commit/c09bf2894517aa0d04ff746cd2be0f3f1a96bed3.

PR #2825, opened 2025-06-27
Add io_queue unit test for checking request rates

robertbindar:
This PR adds two support patches that enable tests to configure an io_queue without duplicating code and calculations from within the reactor code and a unit test that calculates iops and bandwidth out of token bucket rate and costs and asserts that they are within 95% of the configured values.

Fixes #2822

PR #2854, opened 2025-07-11
Fix hang in io_queue for big write ioproperties numbers

### LINKED ISSUE #2817: ioinfo hangs if the write requests are configured with big numbers in io-properties-file
Running ioinfo with
```yaml
disks: 
  - mountpoint: /var/lib/scylla/robitza
    read_iops: 1000
    read_bandwidth: 1000M
    write_iops: 25600000
    write_bandwidth: 256G
```
makes the tool hang.
My gdb seems to wander around this [function](https://github.com/scylladb/seastar/blob/master/src/core/io_queue.cc#L645) 

robertbindar:
This patch fixes an infinite loop in `io_queue` reproducible when io-properties are configured with some big numbers for write requests.

`disk_req_write_to_read_multiplier` ends up being 0 when assigned the result of `io_queue::read_request_base_count * p.read_req_rate) / p.write_req_rate` in reactor code. Same for
`disk_blocks_write_to_read_multiplier`.
Both of them equal to 0 means `internal::request_tokens` always renders 0 tokens, thus capacity 0 and the consequence is an infinite loop in `io_group` constructor.

This patch currently lacks a test because one will be much easier to add once https://github.com/scylladb/seastar/pull/2825 lands. We can either wait for that to land and add it after or do it as a followup.

Fixes #2817

### DIFF
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -183,8 +183,8 @@ public:
         unsigned id;
         unsigned long req_count_rate = std::numeric_limits<int>::max();
         unsigned long blocks_count_rate = std::numeric_limits<int>::max();
-        unsigned disk_req_write_to_read_multiplier = read_request_base_count;
-        unsigned disk_blocks_write_to_read_multiplier = read_request_base_count;
+        double disk_req_write_to_read_multiplier = read_request_base_count;
+        double disk_blocks_write_to_read_multiplier = read_request_base_count;
         size_t disk_read_saturation_length = std::numeric_limits<size_t>::max();
         size_t disk_write_saturation_length = std::numeric_limits<size_t>::max();
         sstring mountpoint = "undefined";
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -642,11 +642,11 @@ io_queue::io_queue(io_group_ptr group, internal::io_sink& sink)
 io_throttler::config io_group::configure_throttler(const io_queue::config& qcfg) noexcept {
     io_throttler::config cfg;
     cfg.label = fmt::format("io-queue-{}", qcfg.id);
-    double min_weight = std::min(io_queue::read_request_base_count, qcfg.disk_req_write_to_read_multiplier);
-    double min_size = std::min(io_queue::read_request_base_count, qcfg.disk_blocks_write_to_read_multiplier);
+    double min_weight = std::min(static_cast<double>(io_queue::read_request_base_count), qcfg.disk_req_write_to_read_multiplier);
+    double min_size = std::min(static_cast<double>(io_queue::read_request_base_count), qcfg.disk_blocks_write_to_read_multiplier);
     cfg.min_tokens = min_weight / qcfg.req_count_rate + min_size / qcfg.blocks_count_rate;
-    double limit_min_weight = std::max(io_queue::read_request_base_count, qcfg.disk_req_write_to_read_multiplier);
-    double limit_min_size = std::max(io_queue::read_request_base_count, qcfg.disk_blocks_write_to_read_multiplier) * qcfg.block_count_limit_min;
+    double limit_min_weight = std::max(static_cast<double>(io_queue::read_request_base_count), qcfg.disk_req_write_to_read_multiplier);
+    double limit_min_size = std::max(static_cast<double>(io_queue::read_request_base_count), qcfg.disk_blocks_write_to_read_multiplier) * qcfg.block_count_limit_min;
     cfg.limit_min_tokens = limit_min_weight / qcfg.req_count_rate + limit_min_size / qcfg.blocks_count_rate;
     cfg.rate_limit_duration = qcfg.rate_limit_duration;
     return cfg;
@@ -685,7 +685,9 @@ io_group::io_group(io_queue::config io_cfg, unsigned nr_queues)
         for (unsigned shift = 0; ; shift++) {
             auto tokens = internal::request_tokens(io_direction_and_length(idx, 1 << (shift + io_queue::block_size_shift)), _config);
             auto cap = fg.tokens_capacity(tokens);
+
             if (cap > max_cap) {
+
                 if (shift == 0) {
                     throw std::runtime_error("IO-group limits are too low");
                 }
@@ -858,8 +860,8 @@ stream_id io_queue::request_stream(io_direction_and_length dnl) const noexcept {
 
 double internal::request_tokens(io_direction_and_length dnl, const io_queue::config& cfg) noexcept {
     struct {
-        unsigned weight;
-        unsigned size;
+        double weight;
+        double size;
     } mult[2];
 
     mult[io_direction_write] = {
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4275,11 +4275,11 @@ public:
 
         if (p.read_bytes_rate != std::numeric_limits<uint64_t>::max()) {
             cfg.blocks_count_rate = (io_queue::read_request_base_count * (unsigned long)per_io_group(p.read_bytes_rate, nr_groups)) >> io_queue::block_size_shift;
-            cfg.disk_blocks_write_to_read_multiplier = (io_queue::read_request_base_count * p.read_bytes_rate) / p.write_bytes_rate;
+            cfg.disk_blocks_write_to_read_multiplier = (static_cast<double>(io_queue::read_request_base_count) * p.read_bytes_rate) / p.write_bytes_rate;
         }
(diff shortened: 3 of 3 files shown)


bhalevy:
@avikivity can you please review this fix while @xemul is on PTO?

src/core/io_queue.cc
```diff
@@ -685,7 +685,9 @@ io_group::io_group(io_queue::config io_cfg, unsigned nr_queues)
         for (unsigned shift = 0; ; shift++) {
             auto tokens = internal::request_tokens(io_direction_and_length(idx, 1 << (shift + io_queue::block_size_shift)), _config);
             auto cap = fg.tokens_capacity(tokens);
+
             if (cap > max_cap) {
+
```
bhalevy: nit: please refrain from mixing cosmetic changes with functional changes

include/seastar/core/io_queue.hh
```diff
@@ -183,8 +183,8 @@ public:
         unsigned id;
         unsigned long req_count_rate = std::numeric_limits<int>::max();
         unsigned long blocks_count_rate = std::numeric_limits<int>::max();
-        unsigned disk_req_write_to_read_multiplier = read_request_base_count;
-        unsigned disk_blocks_write_to_read_multiplier = read_request_base_count;
+        double disk_req_write_to_read_multiplier = read_request_base_count;
```
regevran: To avoid casting all over the code later:
* Either make `read_request_base_count` a double by itself
* Or 
```cpp
double read_request_base_count_as_double = read_request_base_count;
```
and use `read_request_base_count_as_double` later in the code.

src/core/io_queue.cc
```diff
@@ -858,8 +858,8 @@ stream_id io_queue::request_stream(io_direction_and_length dnl) const noexcept {
 
 double internal::request_tokens(io_direction_and_length dnl, const io_queue::config& cfg) noexcept {
     struct {
-        unsigned weight;
-        unsigned size;
+        double weight;
+        double size;
```
xemul: Floating point math is not welcome here

xemul:
> Both of them equal to 0 means `internal::request_tokens` always renders
> 0 tokens, thus capacity 0 and the consequence is an infinite loop in
> `io_group` constructor.

Presumably you're talking about this place

```c++
    auto update_max_size = [this] (unsigned idx) {
        auto g_idx = _config.duplex ? idx : 0;
        const auto& fg = _fgs[g_idx];
        auto max_cap = fg.maximum_capacity();
        for (unsigned shift = 0; ; shift++) {
            auto tokens = internal::request_tokens(io_direction_and_length(idx, 1 << (shift + io_queue::block_size_shift)), _config);
            auto cap = fg.tokens_capacity(tokens);
            if (cap > max_cap) {
                if (shift == 0) {
                    throw std::runtime_error("IO-group limits are too low");
                } 
                _max_request_length[idx] = 1 << ((shift - 1) + io_queue::block_size_shift);
                break;
            }
        };
    };
```

with cap being zero this loop never ends.

If so, then it's this loop that must be fixed that that it bails out on, say, 16MB requests setting them as max length.

src/core/io_queue.cc
robertbindar: well, I know, but we should welcome it here, I don't see it breaking anything here.
In this particular function, you cast to double anyway in the return stmt.
What this change addresses is the fact that for bigger configured iops and bandwidth, `disk_req_write_to_read_multiplier` and `disk_blocks_write_to_read_multiplier` end up < 1, so casting them to unsigned would make them 0, so in the end costs end up 0.
They should end up 0 anyway the closer you get to uint64_t max, but not that soon.

I think this should be split into a different PR as it addresses a different issue now that it's settled the hang needs to be fixed by fixing the loop, what do you think?

include/seastar/core/io_queue.hh
```diff
@@ -176,7 +176,7 @@ public:
     // (read_request_base_count * 130) / 100.
     // It is also technically possible for reads to be the expensive ones, in which case
     // writes will have an integer value lower than read_request_base_count.
-    static constexpr unsigned read_request_base_count = 128;
+    static constexpr unsigned read_request_base_count = 128.0;
```
xemul: `.0` part is excessive

include/seastar/core/io_queue.hh
```diff
@@ -397,8 +397,13 @@ private:
     friend struct ::io_queue_for_tests;
     friend const io_throttler& internal::get_throttler(const io_queue& ioq, unsigned stream);
 
+    static constexpr unsigned request_length_limit = 16 << 20; // 16 MiB
```
xemul, line 405: Need a comment explaining what it is

xemul: Code comment, I mean

src/core/reactor.cc
```diff
@@ -4278,7 +4278,7 @@ class disk_config_params {
         cfg.id = q;
 
         if (p.read_bytes_rate != std::numeric_limits<uint64_t>::max()) {
-            cfg.blocks_count_rate = (io_queue::read_request_base_count * (unsigned long)per_io_group(p.read_bytes_rate, nr_groups)) >> io_queue::block_size_shift;
+            cfg.blocks_count_rate = (io_queue::read_request_base_count * per_io_group(p.read_bytes_rate, nr_groups)) >> io_queue::block_size_shift;
```
xemul: Please explain why this change belongs here (or drop one)

src/core/io_queue.cc
```diff
@@ -683,7 +683,16 @@ io_group::io_group(io_queue::config io_cfg, unsigned nr_queues)
...
         auto max_cap = fg.maximum_capacity();
         for (unsigned shift = 0; ; shift++) {
-            auto tokens = internal::request_tokens(io_direction_and_length(idx, 1 << (shift + io_queue::block_size_shift)), _config);
+            // Break this loop if the calculated request length gets larger than the
+            // request length limit. This avoids an infinite loop if the
+            // configured io_properties are huge and capacity ends up being always 0.
+            // _max_request_length will hold default values of io_group::request_length_limit
+            unsigned long req_length = 1ul << (shift + io_queue::block_size_shift);
+            if (req_length > io_group::request_length_limit) {
+                break;
```
xemul: Need an info-level message here saying that maximum request length limit was cut off early at some "default" value

src/core/io_queue.cc
```diff
@@ -683,7 +683,16 @@ io_group::io_group(io_queue::config io_cfg, unsigned nr_queues)
         const auto& fg = _fgs[g_idx];
         auto max_cap = fg.maximum_capacity();
         for (unsigned shift = 0; ; shift++) {
-            auto tokens = internal::request_tokens(io_direction_and_length(idx, 1 << (shift + io_queue::block_size_shift)), _config);
+            // Break this loop if the calculated request length gets larger than the
+            // request length limit. This avoids an infinite loop if the
+            // configured io_properties are huge and capacity ends up being always 0.
+            // _max_request_length will hold default values of io_group::request_length_limit
+            unsigned long req_length = 1ul << (shift + io_queue::block_size_shift);
+            if (req_length > io_group::request_length_limit) {
```
xemul: Thinking out loud.

Let's assume we hit that if, so req_length > request_length_limit. Then we set max request size to be request_length_limit value, but we don't know what's `request_tokens(request_length_limit) <=> max_cap` -- it can happen (in theory) that `request_tokens(request_length_limit) > max_cap` which is not correct.

src/core/reactor.cc
robertbindar: it doesn't, sorry, it's a leftover from the previous version, I'll remove

src/core/io_queue.cc
robertbindar: hmm, fair point, in theory it's possible. That should be fixable by moving this cutoff point below the request_tokens() logic, correct?

xemul: Then it may happen that length below request_lenght_limit is set as the maximum request size, while tokens(request_length_limit) may still be acceptable

robertbindar: @xemul you mean we end up with tokens(max(2^i)) < tokens(request_length_limit) < max_cap and we choose max(2^i) ?

xemul: Yes

robertbindar: I agree, but I'm ok with this, the current algorithm is "choose the biggest power of two which results in capacity smaller than max_cap", that behavior is preserved if we move the cut off point below. 
let me see if I can cook something to fine tune this.

robertbindar: I added a short snippet to account for that. The path shouldn't be too hot IIUC to worry about the extra calculations.

Issue #2816, opened 2025-06-24
IO Scheduler report costs 0 if read_iops,read_bandwidth get big values


robertbindar:
Running `ioinfo` with the io-properties-file below will have IO scheduler report a cost of 0 for reads. The default numbers for when the scheduler runs unconfigured are UINT64_MAX.

```yaml
disks: 
  - mountpoint: /var/lib/scylla/somedir
    read_iops: 3355443200000
    read_bandwidth: 33554432G
    write_iops: 1000
    write_bandwidth: 1000M
```

Refs: https://github.com/scylladb/seastar/issues/2802

xemul:
Similar (zero or very low costs) should be in #2819 
BTW, maybe zero costs is not logically incorrect, it should be just the rest of the code handles it. E.g. fair_queue dispatching is sort of prepared for that:

```c++
        // Usually the cost of request is tens to hundreeds of thousands. However, for
        // unrestricted queue it can be as low as 2k. With large enough shares this
        // has chances to be translated into zero cost which, in turn, will make the
        // class show no progress and monopolize the queue.
        auto req_cap = req._capacity;
        auto req_cost  = std::max(req_cap / h._shares, (capacity_t)1);
```

robertbindar:
Came across the code whilst investigating #2819.
```c++
double internal::request_tokens(io_direction_and_length dnl, const io_queue::config& cfg) noexcept {
    ...
    return double(m.weight) / cfg.req_count_rate + double(m.size) * (dnl.length() >> io_queue::block_size_shift) / cfg.blocks_count_rate;
}
```
This returns a very small floating point number if you set a big enough value in the io properties file.

Then on this number we call:
```c++

    static capacity_t tokens_capacity(double tokens) noexcept {
        return tokens * token_bucket_t::rate_cast(std::chrono::seconds(1)).count() * fixed_point_factor;
    }
```
`capacity_t` is an alias for `uint64_t`, so we cast the calculated request cost to 0.

I checked the code and I don't think either that cost of 0 is logically incorrect. I'll get back to verify is code is prepared to handle cost of 0, or if it turns out logically incorrect, add a test to cover the case.

xemul:
@robertbindar , are costs still zeroes for large bandwidth/IOPS? Presumably they are and #2857 didn't fix it.

robertbindar:
[...]

PR #2857, opened 2025-07-14
Fix incorrect defaults for io queue iops/bandwidth

### LINKED ISSUE #2819: Unconfigured io-properties render some too large costs
The IO scheduler outputs large costs for read/write requests when running without io-properties, this is what an `ioinfo` run
shows on a i7i.12xlarge machine:
```
disk_read_max_length: 4194304
disk_write_max_length: 4194304
device: 2304
io_latency_goal_ms: 0.75
io_queue:
  id: 0
  req_count_rate: 2147483647
  blocks_count_rate: 2147483647
  disk_req_write_to_read_multiplier: 128
  disk_blocks_write_to_read_multiplier: 128
fair_queue:
  capacities:
    512:
      read: 2000
      write: 2000
    1024:
      read: 3000
      write: 3000
    2048:
      read: 5000
      write: 5000
    4096:
      read: 9000
      write: 9000
    8192:
      read: 17000
      write: 17000
    16384:
      read: 33000
      write: 33000
    32768:
      read: 65000
      write: 65000
    65536:
      read: 129000
      write: 129000
    131072:
      read: 257000
      write: 257000
  per_tick_grab_threshold: 262144
  token_bucket:
    limit: 12582912
    rate: 16777216
    threshold: 2000
```
If you take the 128k request, the computed bandwidth is ~8.1GB/s which is very similar to the throttling point we hit in #2802 

robertbindar:
This patch switches de default values of `req_count_rate` and `blocks_count_rate` from io_queue to `unsigned long` max value.

This fixes an issue we've been chasing for quite a while, iotune would report maximum bandwidth values of aprox. 8.1GB/s which is much less than what should be reported for i7i instances. This issue was also reproducible with the `ioinfo` tool, the io scheduler would render too high request costs when running unconfigured, now the resulting request cost for unconfigured io scheduler is 0, which makes the token bucket not grab any tokens, giving us unbounded output for iotune to enjoy.

I doubled checked usages of these iops/bandwidth uint64_max variables and the costs which now might become 0 and I didn't spot any problematic cases of overflows or division by 0.

Fixes #2819

### DIFF
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -181,8 +181,8 @@ public:
 
     struct config {
         unsigned id;
-        unsigned long req_count_rate = std::numeric_limits<int>::max();
-        unsigned long blocks_count_rate = std::numeric_limits<int>::max();
+        unsigned long req_count_rate = std::numeric_limits<unsigned long>::max();
+        unsigned long blocks_count_rate = std::numeric_limits<unsigned long>::max();
         unsigned disk_req_write_to_read_multiplier = read_request_base_count;
         unsigned disk_blocks_write_to_read_multiplier = read_request_base_count;
         size_t disk_read_saturation_length = std::numeric_limits<size_t>::max();
--- a/tests/unit/io_queue_test.cc
+++ b/tests/unit/io_queue_test.cc
@@ -510,3 +510,15 @@ SEASTAR_TEST_CASE(test_request_iovec_split) {
 
     return make_ready_future<>();
 }
+
+SEASTAR_THREAD_TEST_CASE(test_unconfigured_io_queue) {
+    io_queue_for_tests tio;
+
+    for (uint64_t reqsize = 512; reqsize < 128 << 10; reqsize <<= 1) {
+        auto cost_read = tio.queue.request_capacity(internal::io_direction_and_length(internal::io_direction_and_length::read_idx, reqsize));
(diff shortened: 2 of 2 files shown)


xemul:
This PR says it fixes #2816, but it looks like it doesn't -- generated costs should be still zeroes.

xemul:
@robertbindar , according to https://github.com/scylladb/seastar/issues/2817, large IOPS and bandwidth should cause seastar app to hang in io_group constructor. Please, check why it doesn't happen with this patch applied

xemul:
I had to revers this PR, because it does steps on #2817, but survives because of another overflow.
https://github.com/scylladb/seastar/commit/60b2e7da82ae49c1ac6597e61a9a1dad1be2991e

robertbindar:
The reason why #2857 works is here: https://github.com/scylladb/seastar/blob/master/src/core/reactor.cc#L4280
At this section, `disk_blocks_write_to_read_multiplier` and `disk_req_write_to_read_multiplier` end up 0 if the configured write iops and bandwidth are big.
With this PR in, the execution doesn't enter those if statements, so the `disk_*` variables stay to their default values of read_request_base_count (128).

@xemul is correct though, the capacity keeps being 0 even if `disk_*` variables have values bigger than 0, but another overflow in the shift at `internal::request_tokens(io_direction_and_length(idx, 1 << (shift + io_queue::block_size_shift)), _config);` saves the day and turns capacity positive and the loop is broken.
I'll fix the loop and then I'll submit a new PR to re-merge this.

PR #2858, opened 2025-07-14
addr2line, scripts/stall-analyser: change the default tool to llvm-addr2line

bhalevy:
As a follow-up to 4bd3d123, change the default addr2line tool to llvm-addr2line also for scripts/stall-analyser and the BacktraceResolver class.


michoecho (approved):
Not much to review. If it works, it works.

bhalevy:
> Not much to review. If it works, it works.

It works, and much faster.

bhalevy:

PR #2860, opened 2025-07-15
TLS: Force buffer splitting into gnutls record block sized chunks

### LINKED ISSUE #2859: TLS socket in re-handshake state can cause large allocation in send
If a TLS session enters into a re-handshake (such as session key refresh) at the same time as a large buffer is being sent, gnutls will bypass the normal breaking up of sent data into `block-size` (16k default) buffers, and cause an allocation of a intermediate buffer of <data size> + <session key size> (See gnutls_record_send2, case `RECORD_SEND_KEY_UPDATE_1`.

To prevent this, we need to explicitly break up sent buffers into block-size parts, ensuring the intermediate alloc remains small-is (taking `large_allocation_warning_threshold` into account).

elcallio:
Fixes #2859 

If a TLS session enters into a re-handshake (such as session key refresh) at the same time as a large buffer is being sent, gnutls will bypass the normal breaking up of sent data into block-size (16k default) buffers, and cause an allocation of a intermediate buffer of <data size> + <session key size>.
    
(See gnutls_record_send2, case RECORD_SEND_KEY_UPDATE_1).
    
To prevent this, we need to explicitly break up sent buffers into block-size parts, ensuring the intermediate alloc remains small-ish.

Includes adding a white-box function for forcing re-handshake.    
Includes test using said white-box forcing of re-handshaking.

### DIFF
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -488,6 +488,13 @@ namespace tls {
     future<std::optional<session_dn>> get_dn_information(connected_socket& socket);
 
     /**
+     * Force a re-handshake (session key renegotiotion on TLS1.3).
+     * Can only be called on a server side socket.
+     * Mainly for testing purposes.
+     */
+    future<> force_rehandshake(connected_socket& socket);
+
+    /**
      * Subject alt name types.
     */
     enum class subject_alt_name_type {
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -1516,15 +1516,26 @@ public:
 
     future<> do_put(frag_iter i, frag_iter e) {
         SEASTAR_ASSERT(_output_pending.available());
-        return do_for_each(i, e, [this](net::fragment& f) {
+        // #2859
+        // Normally, gnutls_record_send will break up data into gnutls_record_get_max_size() 
+        // sized chunks for us (only processing the first 16k or so of provided buffer).
+        // However, if the session is in a re-handshake state, gnutls will
+        // make an intermediate alloc to prepend the requested session key
+        // to the sent data. This can cause large alloc warnings.
+        // To avoid this, we explicitly break the message into 
+        // block sized parts (same as normal case in gnutls)
+        auto max_record_len = gnutls_record_get_max_size(*this);
+
+        return do_for_each(i, e, [this, max_record_len](net::fragment& f) {
             auto ptr = f.base;
             auto size = f.size;
             size_t off = 0; // here to appease eclipse cdt
-            return repeat([this, ptr, size, off]() mutable {
+            return repeat([this, ptr, size, off, max_record_len]() mutable {
                 if (off == size) {
                     return make_ready_future<stop_iteration>(stop_iteration::yes);
                 }
-                auto res = gnutls_record_send(*this, ptr + off, size - off);
+                auto n = std::min(max_record_len, size - off);
+                auto res = gnutls_record_send(*this, ptr + off, n);
                 if (res > 0) { // don't really need to check, but...
                     off += res;
                 }
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -1245,15 +1245,12 @@ public:
         });
     }
 
-    future<> do_handshake() {
-        if (_connected) {
-            return make_ready_future<>();
-        }
+    future<> do_handshake_invoke(int (*func)(gnutls_session_t)) {
         if (_type == type::CLIENT && !_options.server_name.empty()) {
             gnutls_server_name_set(*this, GNUTLS_NAME_DNS, _options.server_name.data(), _options.server_name.size());
         }
         try {
-            auto res = gnutls_handshake(*this);
+            auto res = func(*this);
             if (res < 0) {
                 switch (res) {
                 case GNUTLS_E_AGAIN:
@@ -1261,14 +1258,14 @@ public:
                     // If none is pending, it should be a no-op
                 {
                     int dir = gnutls_record_get_direction(*this);
-                    return wait_for_output().then([this, dir] {
+                    return wait_for_output().then([this, dir, func] {
                         // we actually E_AGAIN:ed in a write. Don't
                         // wait for input.
                         if (dir == 1) {
-                            return do_handshake();
+                            return do_handshake_invoke(func);
                         }
-                        return wait_for_input().then([this] {
-                            return do_handshake();
+                        return wait_for_input().then([this, func] {
+                            return do_handshake_invoke(func);
                         });
                     });
                 }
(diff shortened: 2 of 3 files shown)


src/net/tls.cc
```diff
@@ -1217,30 +1217,27 @@ class session : public enable_lw_shared_from_this<session> {
         });
     }
 
-    future<> do_handshake() {
-        if (_connected) {
-            return make_ready_future<>();
-        }
+    future<> do_handshake2(int (*func)(gnutls_session_t)) {
```
regevran:
```suggestion
    future<> do_handshake(int (*func)(gnutls_session_t)) {
```
Why there is a need to introduce `2`?

include/seastar/net/tls.hh
```diff
@@ -470,6 +470,13 @@ namespace tls {
      */
     future<std::optional<session_dn>> get_dn_information(connected_socket& socket);
 
+    /**
+     * Force a re-handshake (session key renegotiotion on TLS1.3).
+     * Can only be called on a server side socket.
+     * Mainly for testing purposes.
+     */
+    future<> force_rehandshake(connected_socket& socket);
```
avikivity, line 495: Testing is great, but please add a test for this too.

tests/unit/tls_test.cc
```diff
@@ -1920,3 +1921,110 @@ SEASTAR_THREAD_TEST_CASE(test_reload_certificates_with_only_shard0_notify) {
...
+        s.connection.shutdown_output();
+
+        c.shutdown_input();
+        c.shutdown_output();
+
+        auto stats_after = memory::stats();
+
+        BOOST_CHECK_EQUAL(stats_after.large_allocations(), stats_before.large_allocations());
+    }
+}
```
avikivity: I'm concerned that this will increase the number of sendmsg calls. I do expect an increase, but I prefer an increase by a factor of 2 to a factor of 8. Please measure that.

src/net/tls.cc
```diff
@@ -1482,15 +1482,27 @@ class session : public enable_lw_shared_from_this<session> {
...
+        // #2859
+        // Normally, gnutls_record_send will break up data into gnutls_record_get_max_size() 
+        // sized chunks (only processing the first 16k or so of provided buffer).
+        // However, if the session is in a re-handshake state, gnutls will
+        // make an intermeddiate alloc to prepend the requested session key
+        // to the sent data. This can cause large alloc warnings.
+        // To avoid this, we explicitly break the message into 
+        // block sized parts (same as normal case in gnutls), also
+        // taking the large alloc threshold into account.
+        auto max_record_len = std::min(gnutls_record_get_max_size(*this), memory::get_large_allocation_warning_threshold()/2);
```
avikivity: large_allocation_warning_threshold isn't the same as the preferred allocation granule. Older ScyllaDB had 128k/1M (newer have 128k/128k).

If we can't get it from elsewhere, add it as a config item.

include/seastar/net/tls.hh
elcallio: Not sure how to verify it. However, session key should change I think, so maybe...

tests/unit/tls_test.cc
elcallio: Manual "measurement" shows that not splitting manually, will instead split in the call to gnutls_record_send. I.e. if sending 1M, only the first 16k of data will be accepted; the loop will increment offset, etc. However, might be worth checking if any internal buffering differs, i.e. how many calls are done to actual vec_push here.

src/net/tls.cc
elcallio: How does allocation granularity fit into max/desired buffer sizes? Typically, granularity would mean align _up_ to fit. Obviously not here. 
As it stands, placing a record limit in tls_options is somewhat subpar here, since these are not (atm) part of the server socket flow (and most parameters are indeed client side ones). 
Feel free to suggest a formula/source for the limit here... Another option is of course to just limit at record size, and maybe add yet another accessor to set it, post creation, if need be.

include/seastar/net/tls.hh
elcallio: Nope. No way for me to extract the active session key. Thus no way to "verify" that this function worked. Other than _not_ having the record size fix, and _do_ provoke the oversized alloc warning...

tests/unit/tls_test.cc
elcallio: So, quick manual test shows there is no amplification of actual wire calls. Regardless of whether we break records apart, gnutls itself does the same thing, even with the oversized buffer alloc. 
Number of calls to actual send(...) remains the exact same.

src/net/tls.cc
avikivity: Granularity was a wrong word choice. I meant the preferred maximum contiguous allocation.

Perhaps add tls_options to listen()? Or wrap listen_options in tls_listen_options?

I don't mind limiting at record size, provided it does not result in an 8-fold increase in sendmsg calls.

tests/unit/tls_test.cc
avikivity: Is it one call per 16k or one call per 128k? (or ...?)

elcallio: One per 16k. Regardless of whether I forcefully limit it to this or not.

avikivity: Ok. That's sad but outside the scope here.

src/net/tls.cc
elcallio: Then all calling places would have to enforce using these options (further making tls/non-tls listen different) to prevent the (albeit not very often occurring) large alloc. Seems like adding a lot of burdens on calling code (and forcing a lot of changes in a certain database) for this issue. 
I'd really vote for manually breaking the buffers in to max_record_size chunks (which is what the code does now, effectively - but we can drop the reference to large alloc threshold in tls.cc), and add the external config of record size to either tls options, or yet another tls::-func.

tests/unit/tls_test.cc
elcallio: As state below, if we allow changing max record size, we can technically reduce this, by for example bumping block size to 64k on connections that benefit. Another argument for a tls::set_max_block_size(socket).

src/net/tls.cc
avikivity: We could add code (just an output_stream) to collect those calls, but that may interfere with gnutls wanting to really push a packet. Does it have a separate signal for that (which would go into gnutls::flush)?

Meanwhile, for the problem at hand, splitting into the max record size is okay, as it doesn't regress.

elcallio: No such luck. Also, sounds overly complicated compared to just upping the record size for high-throughput sockets. 
Will remove the threshold reference, and add an accessor for record size to the fix.

elcallio: Sorry, I need to take that back. Actually record size is limited to between 512 and 16k blocks. 
We cannot up it. 
However, in earlier versions (tls1.2) it was maxed at 4096, so there is that...

avikivity: Well we'll have to live with that.

src/net/tls.cc
elcallio: There isn't, but since it is internal I see no reason for overloading, but instead follow the time honoured tradition (very prevalent in gnutls itself) of adding a "2" to secondary helper.

regevran: The point is that with overloading it is easier to reason about the code - there are no questions like "do I need version 2 or 1?" and "what is the difference between them?".

elcallio: Here the point is they are privates; Perhaps "do_handshake_invoke" would be better?

elcallio:
rebased.

elcallio:
ping?

avikivity:
> ping?

Need to replace the red x of failure with the green circle of success

elcallio:
Solid advise. Totally missed the CI had failed this run. Also did so on a real issue. Fixed and updated.

avikivity:
What were the problem and the fix?

elcallio:
Two things, both in the added "force_rehandshake", i.e. method added only for test repro of base issue: 

1.) A missing semaphore sync. Need to block out any pending read/write. In real code this would not matter (unless weird code), but in the test code it matters due to single async "thread" of operations.
2.) Inserting the handshake in the flow can, in some cases (which debug task scheduling helps provoke), cause a send to generate a EAGAIN from gnutls call. This is harmless and we just need to retry. Added this to do_put.

elcallio:
Now, green ping.

PR #2861, opened 2025-07-16
foreign_ptr: add get_wrapped_ptr method

bhalevy:
The motivation for this new method as to allow
the user to check the wrapped pointer use_count
on the owner shard before e.g. clearing it,
if the use_count equals 1.

Refs scylladb/scylladb#25026

### DIFF
--- a/include/seastar/core/sharded.hh
+++ b/include/seastar/core/sharded.hh
@@ -1038,6 +1038,18 @@ public:
         _cpu = other._cpu;
         return *this;
     }
+    /// Return a reference to the wrapped pointer.
+    ///
+    /// Warning: This method must be called on the
+    /// owner shard to avoid accidents.
+    const PtrType& get_wrapped_ptr() const noexcept {
+        check_shard();
+        return _value;
+    }
+    PtrType& get_wrapped_ptr() noexcept {
+        check_shard();
+        return _value;
+    }
     /// Releases the owned pointer
     ///
     /// Warning: the caller is now responsible for destroying the
--- a/tests/unit/foreign_ptr_test.cc
+++ b/tests/unit/foreign_ptr_test.cc
@@ -173,3 +173,28 @@ SEASTAR_THREAD_TEST_CASE(foreign_ptr_destroy_test) {
     BOOST_REQUIRE_EQUAL(done[1].get_future().get(), true);
     BOOST_REQUIRE_EQUAL(done[0].get_future().get(), false);
 }
+
+SEASTAR_THREAD_TEST_CASE(test_foreign_ptr_use_count) {
+    shard_id shard = (this_shard_id() + 1) % smp::count;
+    auto p0 = smp::submit_to(shard, [] {
+        return make_foreign(make_lw_shared<sstring>("foo"));
+    }).get();
+    smp::submit_to(shard, [&] {
+        BOOST_REQUIRE_EQUAL(p0.get_wrapped_ptr().use_count(), 1);
+    }).get();
+    auto p1 = p0.copy().get();
+    smp::submit_to(shard, [&] {
+        BOOST_REQUIRE_EQUAL(p0.get_wrapped_ptr().use_count(), 2);
+    }).get();
+    smp::submit_to(shard, [&] {
+        auto ptr = p0.release();
+        BOOST_REQUIRE_EQUAL(p0.get_wrapped_ptr().use_count(), 0);
+        BOOST_REQUIRE_EQUAL(p1.get_wrapped_ptr().use_count(), 2);
+        ptr = {};
+        BOOST_REQUIRE_EQUAL(p1.get_wrapped_ptr().use_count(), 1);
+    }).get();
+    p1.reset();
+    smp::submit_to(shard, [&] {
+        BOOST_REQUIRE_EQUAL(p0.get_wrapped_ptr().use_count(), 0);
+    }).get();
+}


bhalevy:
@avikivity please review
this is needed for fixing https://github.com/scylladb/scylladb/issues/25026

(we can also add a method exposing `const PtrType&` that will not `release()` it from the foreign_ptr, but I'm not sure we'd want to do this)

include/seastar/core/sharded.hh
```diff
@@ -1029,6 +1035,12 @@ public:
         _cpu = other._cpu;
         return *this;
     }
+    /// Return the use_count of the wrapped shared objects
+    template <typename U = PtrType>
+    std::enable_if_t<HasUseCount<U>, long>
+    use_count() const noexcept {
+        return _value.use_count();
```
avikivity: How can this work? The use count can be incremented remotely on the owning shard.

nyh: My thoughts exactly. The foreign_ptr<shard_ptr<>> itself cannot be copied directly - there is a `copy()` method but it submits the request to the owning shard, and only it is allowed to handle the use count and make additional copies of the original shared ptr. 

So if you really need the use count, you need to submit this request to the owning shard. But worse - even if you do that, you'll know the result you get back is worthless, because as soon as you get it back it may have changed in the owning shard!

I don't know what you intend to do with use_count(), I'm guessing you want to drop your own copy if you know nobody else is holding a copy. I think you'll have to do this atomically. Something ike (psuedo-code, not real syntax, inspired by the existing destroy_on())
```
    static future<> maybe_reset(foreign_ptr<shared_ptr<T> p))  {
       if (!p) {
            return;
       }
       if (p->cpu != this_shard_id()) {
                return smp::submit_to(p->cpu, [v = std::move(p)] () mutable {
                    // Destroy the contained pointer. We do this explicitly
                    // in the current shard, because the lambda is destroyed
                    // in the shard that submitted the task.
                    v = {};
                });
            } else {
                p = {};
            }
        }
        return make_ready_future<>();
    }
```

bhalevy: yes, similar that, just with clear_gently, and calling clear_gently on the contained shared ptr only when its use_count reaches 1, otherwise it's a no-op.  In any case, the check is done on the shard owner inside submit_to of course.

> How can this work? The use count can be incremented remotely on the owning shard.

We can throw if it's not called on the owner shard and document that requirement.

Alternatively, we can expose a method to return a (const) reference to the contained smart/shared pointer that needs to be called also only on the owner shard, and let the caller call use_count and the shared pointer.

avikivity: If we're on the owning shard, can't we remove the wrapper?

bhalevy: > If we're on the owning shard, can't we remove the wrapper?

The only way to this today is with `release()`, but that's too aggressive (especially if use_count() > 1)

bhalevy: we could have something like:
```c++
const PtrType& foreign_ptr::get_ptr() const {
	return _value;
}
```

bhalevy:
In [fc1988f](https://github.com/scylladb/seastar/commit/fc1988f872e40c3a2730b8357567f067e5cb4ff5):
- instead of providing a `use_count()` method, provide `get_wrapped_ptr()`, that should be called on the owner shard.

include/seastar/core/sharded.hh
bhalevy: In [fc1988f](https://github.com/scylladb/seastar/commit/fc1988f872e40c3a2730b8357567f067e5cb4ff5): add `get_wrapped_ptr()`

include/seastar/core/sharded.hh
```diff
@@ -920,6 +920,11 @@ void sharded<Service>::check_local() const {
     }
 }
 
+template<typename T>
+concept HasUseCount = requires (T ptr) {
+    { ptr.use_count() } -> std::convertible_to<long>;
+};
```
nyh: This is no longer needed in this patch?

include/seastar/core/sharded.hh
```diff
@@ -952,6 +957,7 @@ private:
     PtrType _value;
     unsigned _cpu;
 private:
+    constexpr static bool has_use_count = HasUseCount<PtrType>;
```
nyh: This is no longer needed in this patch?

include/seastar/core/sharded.hh
```diff
@@ -1029,6 +1035,13 @@ public:
         _cpu = other._cpu;
         return *this;
     }
+    /// Return a reference to the wrapped pointer.
+    ///
+    /// Warning: This method is best called on the
+    /// owner shard to avoid accidents.
```
nyh, line 1044: Best replace "best" by "must", right?
And I'm not sure "accidents" is the best description of the problem you'd want to avoid.

Moreover, I don't think it's obvious why anybody might want to use this method. Perhaps it will be useful to have an example (a full correct one, including the redirection to the other cpu) of how to use this correctly. I suspect that if you do sit down to write such an example, you'll realize that this function you added here - get_wrapped_ptr() isn't quite good enough for what you need to do. You also need to redirect execution to the correct CPU (so need access to _cpu), and also need to do both use_count() and destroy the pointer on the same callback on the owner CPU. I suspect you'll end up with a function like I tried to write in my previous review. I suspect that this get_wrapped_ptr() won't entirely, by itself, be able to do the job.

Finally, I think we should add an assert that this is called on the correct CPU, to ensure that if it's used incorrectly, it will be caught quickly.

nyh (changes requested):

include/seastar/core/sharded.hh
bhalevy: See https://github.com/scylladb/scylladb/pull/24606/commits/a651619f022aba2ce212f225f3d466677c0e9b66 utils/stall_free.hh for intended use.
I can add a documented example as well.

As for the assert, why add it here and not on release()?  Maybe we can add them in debug mode, similar to shared ptr cross-shard access checks?

include/seastar/core/sharded.hh
```diff
@@ -1038,6 +1038,18 @@ public:
         _cpu = other._cpu;
         return *this;
     }
+    /// Return a reference to the wrapped pointer.
+    ///
+    /// Warning: This method must be called on the
+    /// owner shard to avoid accidents.
+    const PtrType& get_wrapped_ptr() const noexcept {
```
avikivity, line 1045: Does the documentation show for this overload?

avikivity: Perhaps call it unwrap_on_owner_shard and call by rvalue reference?

bhalevy: > Perhaps call it unwrap_on_owner_shard and call by rvalue reference?

If the intention is to detach the wrapped ptr then the user can call release and in case it ptr is still valid (like when use_count > 1), there is the reset function to wrap the smart pointer back.  It's a bit wasteful if this is the common case, but it should work.

bhalevy: > Does the documentation show for this overload?

Both variants are listed under "Public Member Functions":

image

But only the first one (`const`) has a hyperlink to the per-method documentation:

image

Is that OK @avikivity @annastuchlik

include/seastar/core/sharded.hh
bhalevy: dropped in [affaf8b](https://github.com/scylladb/seastar/commit/affaf8b6129d4784036aae4878cb949c3ab29a14)

include/seastar/core/sharded.hh
bhalevy: rephrased in [affaf8b](https://github.com/scylladb/seastar/commit/affaf8b6129d4784036aae4878cb949c3ab29a14)

bhalevy:
@avikivity / @nyh I accidentally deleted my branch and it auto-closed this PR.
Can you please reopen it with https://github.com/bhalevy/seastar/tree/foreign_ptr-use_count (https://github.com/bhalevy/seastar/commit/6c64ccb38f7317535cd99949d4320aed85c2f024)

bhalevy:
I accidentally deleted my branch and it auto-closed this PR.
Reopening with bhalevy/seastar@foreign_ptr-use_count (bhalevy@6c64ccb)

bhalevy:
In https://github.com/bhalevy/seastar/commit/6c64ccb38f7317535cd99949d4320aed85c2f024:
- rebased
- renamed get_wrapped_ptr to unwrap_on_owner_shard

avikivity:
> @avikivity / @nyh I accidentally deleted my branch and it auto-closed this PR. Can you please reopen it with https://github.com/bhalevy/seastar/tree/foreign_ptr-use_count ([bhalevy@6c64ccb](https://github.com/bhalevy/seastar/commit/6c64ccb38f7317535cd99949d4320aed85c2f024))

Github doesn't allow me to reopen it.

avikivity:
Now it does.

avikivity:
Wait it didn't work. Suggest to open a new one.

bhalevy:
> Wait it didn't work. Suggest to open a new one.

Replaced this PR with https://github.com/scylladb/seastar/pull/2898

PR #2862, opened 2025-07-17
Add label values to the duplicate metrics exception

amnonh:
A metric must be unique. Trying to register two metrics with the same name and the same labels will throw an exception and, in many cases, terminate the process.

This patch adds the label values to the exception message to make it easier to understand which metric caused the exception.

An example
```
registering metrics twice for metrics: scylladb_current_version {"__level": "basic", "shard": "", "version": "2025.4.0~dev-0.20250709.69b5289a0e54"}
```

### DIFF
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -31,6 +31,7 @@ module;
 #include <boost/algorithm/string.hpp>
 #include <boost/algorithm/string/replace.hpp>
 #include <boost/range/algorithm_ext/erase.hpp>
+#include <fmt/ranges.h>
 
 #ifdef SEASTAR_MODULE
 module seastar;
@@ -485,7 +486,7 @@ register_ref impl::add_registration(const metric_id& id, const metric_type& type
     if (_value_map.find(name) != _value_map.end()) {
         auto& metric = _value_map[name];
         if (metric.find(rm->info().id.labels()) != metric.end()) {
-            throw double_registration("registering metrics twice for metrics: " + name);
+            throw double_registration(fmt::format("registering metrics twice for metrics: {} with labels {}", name, rm->info().id.labels()));
         }
         if (metric.info().type != type.base_type) {
             throw std::runtime_error("registering metrics " + name + " registered with different type.");


src/core/metrics.cc
```diff
@@ -475,6 +475,14 @@ std::vector<std::deque<metric_function>>& impl::functions() {
...
 }
 
+static sstring get_label_as_string(const std::map<sstring, sstring>& labels) {
+    sstring res;
+    for (const auto& [key, value] : labels) {
+        res += ",";
+        res += key + "=" + value;
+    }
+    return res;
+}
```
avikivity: You can just format() the map, it will handle it if you include fmt/ranges.h.

src/core/metrics.cc
```diff
@@ -485,7 +493,7 @@ register_ref impl::add_registration(const metric_id& id, const metric_type& type
     if (_value_map.find(name) != _value_map.end()) {
         auto& metric = _value_map[name];
         if (metric.find(rm->info().id.labels()) != metric.end()) {
-            throw double_registration("registering metrics twice for metrics: " + name);
+            throw double_registration("registering metrics twice for metrics: " + name + get_label_as_string(rm->info().id.labels()));
```
avikivity: Surely a space is missing here.

amnonh:
> @amnonh `unknown location(0): fatal error: in "test_tb_params": seastar::metrics::double_registration: registering metrics twice for metrics: io_queue_flow_ratio,iogroup=0,mountpoint=,shard=0`

@robertbindar I think that's the issue, the value is missing, so I'm happy it helps

src/core/metrics.cc
amnonh: yes, I'll send v2 with your two comments shortly.
The original  implementation started with a comma so there was no need for a space

amnonh:
@avikivity I've sent V2, this patch is one of those annoying things that you never need until you really do.

amnonh:
@avikivity ping again. It's not super urgent until someone does something funky, gets an exception, and doesn't know why.

src/core/metrics.cc
```diff
@@ -485,7 +490,7 @@ register_ref impl::add_registration(const metric_id& id, const metric_type& type
     if (_value_map.find(name) != _value_map.end()) {
         auto& metric = _value_map[name];
         if (metric.find(rm->info().id.labels()) != metric.end()) {
-            throw double_registration("registering metrics twice for metrics: " + name);
+            throw double_registration("registering metrics twice for metrics: " + name + " " + get_label_as_string(rm->info().id.labels()));
         }
```
avikivity, line 490: You can just fmt::format() here. No need to mix formatting with string concatentation.

```c++
fmt::format("registering metrics twice for metric {} with labels {}", ...)
```

amnonh: Sure, sending a new version shortly

PR #2865, opened 2025-07-22
io_queue costs shouldn't be zero for large bandwidth/iops

### LINKED ISSUE #2816: IO Scheduler report costs 0 if read_iops,read_bandwidth get big values

robertbindar:
This patch fixes a case in which given large enough iops/bandwitdh configured in io-properties, the requests costs would show up 0.

The bug is in internal::request_tokens, for bigger configured iops and bandwidth, disk_req_write_to_read_multiplier and
disk_blocks_write_to_read_multiplier end up < 1, so casting them to unsigned would make them 0, so in the end costs end up 0. They should end up 0 anyway the closer you get to uint64_t max, but not that soon.

This PR depends on https://github.com/scylladb/seastar/pull/2825 getting some test glue code in.

Fixes #2816

### DIFF
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -176,15 +176,15 @@ public:
     // (read_request_base_count * 130) / 100.
     // It is also technically possible for reads to be the expensive ones, in which case
     // writes will have an integer value lower than read_request_base_count.
-    static constexpr unsigned read_request_base_count = 128;
+    static constexpr double read_request_base_count = 128.0;
     static constexpr unsigned block_size_shift = 9;
 
     struct config {
         unsigned id;
         unsigned long req_count_rate = std::numeric_limits<int>::max();
         unsigned long blocks_count_rate = std::numeric_limits<int>::max();
-        unsigned disk_req_write_to_read_multiplier = read_request_base_count;
-        unsigned disk_blocks_write_to_read_multiplier = read_request_base_count;
+        double disk_req_write_to_read_multiplier = read_request_base_count;
+        double disk_blocks_write_to_read_multiplier = read_request_base_count;
         size_t disk_read_saturation_length = std::numeric_limits<size_t>::max();
         size_t disk_write_saturation_length = std::numeric_limits<size_t>::max();
         sstring mountpoint = "undefined";
--- a/src/core/disk_params.cc
+++ b/src/core/disk_params.cc
@@ -178,7 +178,7 @@ struct io_queue::config disk_config_params::generate_config(const disk_params& p
     cfg.id = q;
 
     if (p.read_bytes_rate != std::numeric_limits<uint64_t>::max()) {
-        cfg.blocks_count_rate = (io_queue::read_request_base_count * (unsigned long)per_io_group(p.read_bytes_rate, nr_groups)) >> io_queue::block_size_shift;
+        cfg.blocks_count_rate = static_cast<unsigned long>(io_queue::read_request_base_count * per_io_group(p.read_bytes_rate, nr_groups)) >> io_queue::block_size_shift;
         cfg.disk_blocks_write_to_read_multiplier = (io_queue::read_request_base_count * p.read_bytes_rate) / p.write_bytes_rate;
     }
     if (p.read_req_rate != std::numeric_limits<uint64_t>::max()) {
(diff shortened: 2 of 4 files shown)


xemul:
> for bigger configured iops and bandwidth, disk_req_write_to_read_multiplier and disk_blocks_write_to_read_multiplier end up < 1

I don't get it. Those multipliers are `128 * read_{iops|bandwidth} / write_{iops|bandiwdth}` so they can only turn 0 in case of heavy read-vs-write imbalance in config, but _not_ when they are both huge.

robertbindar:
> > for bigger configured iops and bandwidth, disk_req_write_to_read_multiplier and disk_blocks_write_to_read_multiplier end up < 1
> 
> I don't get it. Those multipliers are `128 * read_{iops|bandwidth} / write_{iops|bandiwdth}` so they can only turn 0 in case of heavy read-vs-write imbalance in config, but _not_ when they are both huge.

Yes, indeed, they turn 0 if write iops/bandwidth is much higher than read iops/bandwidth. But in integer space they turn out 0 too soon, because the code truncates them to 0 too early and doesn't give the fixed_point_factor the chance to raise the costs above 0 in the final calculation.
IMO the correct place to make the truncation to 0 is in:
```c++
    static capacity_t tokens_capacity(double tokens) noexcept {
        return tokens * token_bucket_t::rate_cast(std::chrono::seconds(1)).count() * fixed_point_factor;
    }
```
let the cast to `capacity_t` do its job if  tokens shifted by fixed_point_factor < 1.

The description in https://github.com/scylladb/seastar/issues/2816 is misleading because it lists an example with read iops/bandwidth much higher than write iops_bandwidth. This is how I initially discovered these 0 costs, but they are correct because in that example I gave, the read numbers are so high that even with this patch in, the calculations in floating point space in this final `tokens_capacity` function end up < 1, so the truncation to `capacity_t` renders a cost of 0.

robertbindar:
the added test will fail till we merge #2854

xemul:
BTW, first things (mush have been) first

> io_queue costs shouldn't be zero for large bandwidth/iops

Why are zero costs wrong? :D

xemul:
> Yes, indeed, they turn 0 if write iops/bandwidth is much higher than read iops/bandwidth.

Then it's not something that we want to fix. Set ups with writes speed being 128+ times faster than reads are not supported (either recommended). We do have set ups where _reads_ are order of magnitude faster than writes, but for those write multiplier will be well above 128.

The best that can be done here is -- check that multipler turned zero, place a warning in logs and force it to 1

xemul:
> But in integer space they turn out 0 too soon, because the code truncates them to 0 too early and doesn't give the fixed_point_factor the chance to raise the costs above 0 in the final calculation.

Actually the token calculations renders double (float) tokens before multiplying them by the 1<<24 factor

robertbindar:
> > But in integer space they turn out 0 too soon, because the code truncates them to 0 too early and doesn't give the fixed_point_factor the chance to raise the costs above 0 in the final calculation.
> 
> Actually the token calculations renders double (float) tokens before multiplying them by the 1<<24 factor

`internal::request_tokens(dnl, cfg);` returns a double tokens number indeed,
but the calculation inside the function: `double(m.weight) / cfg.req_count_rate + double(m.size) * (dnl.length() >> io_queue::block_size_shift) / cfg.blocks_count_rate;`
turns 0 if the multipliers are 0 (weight and size are in the upper part of the fractions).
In unbalanced cases where writes >> reads, these multipliers can end up e.g. 0.721, but they actually end up 0 because we have them as integers.

robertbindar:
> > Yes, indeed, they turn 0 if write iops/bandwidth is much higher than read iops/bandwidth.
> 
> Then it's not something that we want to fix. Set ups with writes speed being 128+ times faster than reads are not supported (either recommended). We do have set ups where _reads_ are order of magnitude faster than writes, but for those write multiplier will be well above 128.
> 
> The best that can be done here is -- check that multipler turned zero, place a warning in logs and force it to 1

I can agree with not supporting this, but we should have a test that checks this guarantee, agree? A warning would also be nice. This way the behavior is documented and will help us save time in the future when we'll hit it (most probably in some unrealistic crazy tests).

xemul:
> I can agree with not supporting this, but we should have a test that checks this guarantee, agree? 

Yes

travisdowns:
This conversation is strong evidence in favor of doing the whole calculation as a float, rather than trying to use fixed-point with all its footguns. Floating point math is fast everywhere (and that's been true for years).

PR #2870, opened 2025-07-23
http/reply: introduce set_cookie()

denesb:
Cookies are problematic to use with existing set_header() and friends, as multiple cookies are expected to be set with multiple Set-Cookie header lines, which is a problem for the current header related infrastructure which assumes one value for one header key. Introduce separate mechanism for setting cookies, allowing for multiple cookies to be set. Set-Cookie headers are written separately after the normal headers are written.

### DIFF
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -98,7 +98,6 @@ future<> connection::do_response_loop() {
 }
 
 future<> connection::start_response() {
-    if (_resp->_body_writer) {
         return _resp->write_reply(out()).then_wrapped([this] (auto f) {
             if (f.failed()) {
                 // In case of an error during the write close the connection
@@ -133,25 +132,6 @@ future<> connection::start_response() {
             _resp.reset();
             return make_ready_future<>();
         });
-    }
-    set_headers(*_resp);
-    _resp->_headers["Content-Length"] = to_sstring(
-            _resp->_content.size());
-    return _write_buf.write(_resp->_response_line.data(),
-            _resp->_response_line.size()).then([this] {
-        return _resp->write_reply_headers(out());
-    }).then([this] {
-        return _write_buf.write("\r\n", 2);
-    }).then([this] {
-        if (_resp->_skip_body) {
-            return make_ready_future<>();
-        }
-        return write_body();
-    }).then([this] {
-        return _write_buf.flush();
-    }).then([this] {
-        _resp.reset();
-    });
 }
 
 connection::~connection() {
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -99,7 +99,7 @@ future<> connection::do_response_loop() {
 
 future<> connection::start_response() {
     if (_resp->_body_writer) {
-        return _resp->write_reply_to_connection(out()).then_wrapped([this] (auto f) {
+        return _resp->write_reply(out()).then_wrapped([this] (auto f) {
             if (f.failed()) {
                 // In case of an error during the write close the connection
                 _server._respond_errors++;
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -98,40 +98,40 @@ future<> connection::do_response_loop() {
 }
 
 future<> connection::start_response() {
-        return _resp->write_reply(out()).then_wrapped([this] (auto f) {
-            if (f.failed()) {
-                // In case of an error during the write close the connection
-                _server._respond_errors++;
-                _done = true;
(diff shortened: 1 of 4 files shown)


xemul:
Please, add a test for the new functionality

denesb:
New in v2:
* Added a test

denesb:
> Please, add a test for the new functionality

I added a test and you got a refactoring as a bonus.

src/http/httpd.cc
```diff
@@ -98,7 +98,8 @@ future<> connection::do_response_loop() {
 }
 
 future<> connection::start_response() {
-    if (_resp->_body_writer) {
+    set_headers(*_resp);
```
xemul: This wasn't called on `if (_resp->_body_writer)` branch, not it is. Is that OK?
Maybe it's not needed here at all, as start_response() is called out of do_response_loop() and resp-s get there via generate_reply which calls set_headers() on its own? (the code is indeed hard to follow)

src/http/reply.cc
```diff
@@ -133,20 +133,28 @@ void reply::write_body(const sstring& content_type, sstring content) {
...
 future<> reply::write_reply(output_stream<char>& out) {
-    add_header("Transfer-Encoding", "chunked");
-    return out.write(response_line()).then([this, &out] () mutable {
-        return write_reply_headers(out);
-    }).then([&out] () mutable {
-        return out.write("\r\n", 2);
-    }).then([this, &out] () mutable {
-        if (_skip_body) {
-            return make_ready_future<>();
+    return out.write(response_line().data(), response_line().size()).then([this, &out] {
```
xemul, line 136: In the `! if (_resp->body_writer)` branch the `resp->_response_line` was written, not `response_line()`. I assume they are the same at this point, but now the `_response_line` becomes obsoleted and can be removed? (can be done incrementally)

src/http/httpd.cc
```diff
@@ -98,7 +98,8 @@ future<> connection::do_response_loop() {
...
+
         return _resp->write_reply(out()).then_wrapped([this] (auto f) {
             if (f.failed()) {
```
xemul: Everything that sits in this then_wrapped lambda is now called if replies without `_body_writer` fail. It wasn't so before the patch. Is that OK?

tests/unit/httpd_test.cc
```diff
@@ -2232,3 +2232,63 @@ SEASTAR_THREAD_TEST_CASE(test_content_length_data_sink) {
     do_check(2, "12", true);
     do_check(2, "123", true);
 }
+
+class ostream_adaptor final : public data_sink_impl {
```
xemul: Would existing `memory_data_sink_impl` work just as well?

src/http/httpd.cc
denesb: Yes, this is possibly not needed.

src/http/reply.cc
denesb: Yes.

src/http/httpd.cc
denesb: Yes, the two code paths were very inconsistent with each other, see the other comments above.
The simple code path had just two continuations with a flush() and then reset. This does some additional things, like adding the response to a list of failed ones. Importantly, it seems to do the cleanup `resp.reset()` even if the response fails.
It looks to me like this is more complete error handling, that was written later and just not applied to the earlier path, which remained with the simpler (incomplete?) one. Possibly we never noticed because the simple path fails rarely.

denesb:
@amnonh as the author of this code, please review, especially my refactoring commits.

tests/unit/httpd_test.cc
denesb: Yes, dropped this in v3.

src/http/httpd.cc
denesb: Dropped in v3.

src/http/reply.cc
denesb: I think also `reply::done()` is possibly obsolete and can be removed (you don't need to call it, seems unnecessary). I will follow-up with another cleanup.

denesb:
For completeness.
New in v3:
* Drop `set_headers()` from `connection::start_response()`
* Use `memory_data_sink_impl` in the test (instead of its reimplementation)
* Fix indentation of `connection::start_response()`

amnonh:
Looks good, I like the changes, I was thinking with myself, should we generalize repeated headers, but I think going with a specific cookies handling is the right way.

I know @bhalevy is changing right now some of the same code, so may the odds be ever in your favor

PR #2871, opened 2025-07-25
Fix incorrect defaults for io queue iops/bandwidth

### LINKED ISSUE #2819: Unconfigured io-properties render some too large costs

robertbindar:
This patch switches de default values of `req_count_rate` and `blocks_count_rate` from io_queue to `unsigned long` max value.

This fixes an issue we've been chasing for quite a while, iotune would report maximum bandwidth values of aprox. 8.1GB/s which is much less than what should be reported for i7i instances. This issue was also reproducible with the `ioinfo` tool, the io scheduler would render too high request costs when running unconfigured, now the resulting request cost for unconfigured io scheduler is 0, which makes the token bucket not grab any tokens, giving us unbounded output for iotune to enjoy.

Fixes #2819

Notes:
- Identical with the initial PR which was merged and reverted: https://github.com/scylladb/seastar/pull/2857 (#2854 is needed first so that seastar stays stable, this PR fixes the issue, but only due to another lucky overflow, #2854 addresses that).
- This needs to be merged only after https://github.com/scylladb/seastar/pull/2854

### DIFF
--- a/tests/unit/io_queue_test.cc
+++ b/tests/unit/io_queue_test.cc
@@ -97,6 +97,14 @@ struct io_queue_for_tests {
     future<size_t> queue_request(internal::priority_class pc, internal::io_direction_and_length dnl, internal::io_request req, io_intent* intent, iovec_keeper iovs) noexcept {
         return queue.queue_request(pc, dnl, std::move(req), intent, std::move(iovs));
     }
+
+    size_t max_request_length(int dnl_idx) const noexcept {
+        return group->max_request_length(dnl_idx);
+    }
+
+    constexpr size_t request_length_limit() const noexcept {
+        return io_group::request_length_limit;
+    }
 };
 
 internal::priority_class get_default_pc() {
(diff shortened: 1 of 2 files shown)


tests/unit/io_queue_test.cc
```diff
@@ -563,3 +563,15 @@ SEASTAR_THREAD_TEST_CASE(test_tb_params) {
...
 }
+
+SEASTAR_THREAD_TEST_CASE(test_unconfigured_io_queue) {
+    io_queue_for_tests tio;
+
+    for (uint64_t reqsize = 512; reqsize < 128 << 10; reqsize <<= 1) {
+        auto cost_read = tio.queue.request_capacity(internal::io_direction_and_length(internal::io_direction_and_length::read_idx, reqsize));
+        auto cost_write = tio.queue.request_capacity(internal::io_direction_and_length(internal::io_direction_and_length::write_idx, reqsize));
+
+        SEASTAR_ASSERT(cost_read == 0);
```
xemul: BOOST_CHECK_EQUAL

tests/unit/io_queue_test.cc
```diff
@@ -563,3 +563,15 @@ SEASTAR_THREAD_TEST_CASE(test_tb_params) {
...
+        SEASTAR_ASSERT(cost_write == 0);
```
xemul: Also need BOOST_CHECK validation for `_max_request_length` values calculated by io_group constructor for that case

PR #2872, opened 2025-07-25
http: Add support of multiple key repetitions for the request

### LINKED ISSUE #693: http requests should support multiple key repetitions
query parameters can technically be used more than once.
The current implementation does not support this, but it can be useful if it does

tarasveretilnyk:
This change improves the handling of HTTP request query parameters
by properly supporting multiple occurrences of the same parameter key.

To maintain backward compatibility, the existing get_query_param() method
and query_parameters field are preserved.
The query_parameters field, which stores the last value for each
query parameter key, is deprecated and will be removed in the future.

New functionality has been introduced that require
handling multiple values per parameter key:
- get_query_param_array()
- get_query_params()
- set_query_params()
- set_query_param()

Fixes #693

### DIFF
--- a/apps/httpd/main.cc
+++ b/apps/httpd/main.cc
@@ -66,7 +66,7 @@ void set_routes(routes& r) {
         demo_json::my_object obj;
         obj.var1 = req.param.at("var1");
         obj.var2 = req.param.at("var2");
-        demo_json::ns_hello_world::query_enum v = demo_json::ns_hello_world::str2query_enum(req.query_parameters.at("query_enum"));
+        demo_json::ns_hello_world::query_enum v = demo_json::ns_hello_world::str2query_enum(req.get_query_param("query_enum"));
         // This demonstrate enum conversion
         obj.enum_var = v;
         return obj;
--- a/include/seastar/http/request.hh
+++ b/include/seastar/http/request.hh
@@ -32,6 +32,7 @@
 
 #include <seastar/core/iostream.hh>
 #include <seastar/core/sstring.hh>
+#include <string_view>
 #include <strings.h>
 #include <seastar/http/common.hh>
 #include <seastar/http/mime_types.hh>
@@ -65,7 +66,8 @@ struct request {
     size_t content_length = 0;
     mutable size_t _bytes_written = 0;
     std::unordered_map<sstring, sstring, seastar::internal::case_insensitive_hash, seastar::internal::case_insensitive_cmp> _headers;
-    std::unordered_map<sstring, sstring> query_parameters;
+    // deprecated: it is used to store last value of query parameters, but will be removed in the future
+    [[deprecated("Use helper methods instead")]] std::unordered_map<sstring, sstring> query_parameters;
     httpd::parameters param;
     sstring content; // server-side deprecated: use content_stream instead
     /*
@@ -78,6 +80,20 @@ struct request {
     std::unordered_map<sstring, sstring> chunk_extensions;
     sstring protocol_name = "http";
     http::body_writer_type body_writer; // for client
+
+    using query_parameters_type = std::unordered_map<sstring, std::vector<sstring>, seastar::internal::string_view_hash, std::equal_to<>>;
+private:
+    query_parameters_type _query_params;
+public:
+
+// NOTE: Remove this once `query_parameters` is removed
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    request() = default;
+    request(request&&) = default;
+    request& operator=(request&&) = default;
+    ~request() = default;
+#pragma GCC diagnostic pop
 
     /**
      * Get the address of the client that generated the request
@@ -110,15 +126,79 @@ struct request {
 
     /**
      * Search for the last query parameter of a given key
-     * @param key the query paramerter key
+     * @param key the query parameter key
      * @return the query parameter value, if it exists or empty string
      */
-    sstring get_query_param(const sstring& key) const {
-        auto res = query_parameters.find(key);
-        if (res == query_parameters.end()) {
-            return "";
+    sstring get_query_param(std::string_view key) const {
+        auto res = _query_params.find(key);
+        if (res != _query_params.end()) {
+            return res->second.back();
         }
-        return res->second;
+
+        return "";
+    }
+
+    /**
+     * Search for all query parameters of a given key
+     * @param key the query parameter key
+     * @return a vector of all query parameter values, if it exists or an empty vector
+     */
+    const std::vector<sstring>& get_query_param_array(std::string_view key) const {
+        if (auto res = _query_params.find(key); res != _query_params.end()) {
+            return res->second;
+        }
+        static const std::vector<sstring> empty_vector;
+        return empty_vector;
+    }
+
+    /**
+     * Get all query parameters
+     * @return a map of all query parameters
+     */
+    const query_parameters_type& get_query_params() const {
+        return _query_params;
+    }
+
+    /**
+     * Set a query parameter value
+     * @param key the query parameter key
+     * @param value the query parameter value
+     * @return a reference to this request object
+     */
+    request& set_query_param(std::string_view key, std::string_view value) {
+        return set_query_param(key, {value});
+    }
+
+    /**
+     * Set a query parameter value
+     * @param key the query parameter key
+     * @param values the query parameter values
+     * @return a reference to this request object
+    */
+    request& set_query_param(std::string_view key, std::initializer_list<std::string_view> values) {
+        _query_params[sstring(key)] = std::vector<sstring>(values.begin(), values.end());
+        return *this;
+    }
+
+    /**
+     * Set a query parameter value
+     * @param key the query parameter key
+     * @param values the query parameter values
+     * @return a reference to this request object
+    */
+    request& set_query_param(std::string_view key, std::vector<sstring> values) {
+        _query_params[sstring(key)] = std::move(values);
+        return *this;
+    }
+
+    /**
+     * Set the query parameters in the request objects.
+     * @param params a map of query parameters
+     * @return a reference to this request object
+     */
+    request& set_query_params(query_parameters_type params) {
+        _query_params = std::move(params);
+        return *this;
     }
 
     /**
--- a/include/seastar/util/string_utils.hh
+++ b/include/seastar/util/string_utils.hh
@@ -47,6 +47,19 @@ SEASTAR_MODULE_EXPORT_BEGIN
 // Collection of utilities for working with strings .
 //
 
+struct string_view_hash {
+    using is_transparent = void;
+    size_t operator()(const char *txt) const {
+        return std::hash<std::string_view>{}(txt);
+    }
+    size_t operator()(std::string_view txt) const {
+        return std::hash<std::string_view>{}(txt);
+    }
+    size_t operator()(const sstring &txt) const {
+        return std::hash<sstring>()(txt);
+    }
+};
+
 struct case_insensitive_cmp {
     bool operator()(const sstring& s1, const sstring& s2) const {
         return std::equal(s1.begin(), s1.end(), s2.begin(), s2.end(),
--- a/src/http/request.cc
+++ b/src/http/request.cc
@@ -24,6 +24,7 @@ module;
 #endif
 
 #include <string_view>
+#include <unordered_map>
 #include <utility>
 
 #ifdef SEASTAR_MODULE
@@ -38,16 +39,47 @@ module seastar;
 namespace seastar {
 namespace http {
 
+namespace internal {
+// NOTE: Remove this once `query_parameters` is removed
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    const std::unordered_map<sstring, sstring>& deprecated_query_parameters(const request& r) noexcept {
+        return r.query_parameters;
+    }
+    std::unordered_map<sstring, sstring>& deprecated_query_parameters(request& r) noexcept {
+        return r.query_parameters;
+    }
+#pragma GCC diagnostic pop
+}
+
 sstring request::format_url() const {
     sstring query = "";
-    sstring delim = "?";
-    for (const auto& p : query_parameters) {
-        query += delim + internal::url_encode(p.first);
-        if (!p.second.empty()) {
-            query += "=" + internal::url_encode(p.second);
+    if (!_query_params.empty()) {
+        sstring delim = "&";
+        for (const auto& [key, values] : _query_params) {
+            auto key_component = delim + internal::url_encode(key) + "=";
+            if (values.empty()) {
+                query += key_component;
+            } else {
+                for (const auto& val : values) {
+                    query += key_component + internal::url_encode(val);
+                }
+            }
         }
-        delim = "&";
+        if (!query.empty()) {
+            query[0] = '?';
+        }
+    } else {
+        sstring delim = "?";
+        for (const auto& p : internal::deprecated_query_parameters(*this)) {
+            query += delim + internal::url_encode(p.first);
+            if (!p.second.empty()) {
+                query += "=" + internal::url_encode(p.second);
+            }
+            delim = "&";
+        }
     }
+
     return _url + query;
 }
 
@@ -65,24 +97,28 @@ future<> request::write_request_headers(output_stream<char>& out) const {
 
 void request::add_query_param(std::string_view param) {
     size_t split = param.find('=');
-
+    auto& deprecated_query_parameters = http::internal::deprecated_query_parameters(*this);
     if (split >= param.length() - 1) {
         sstring key;
         if (http::internal::url_decode(param.substr(0,split) , key)) {
-            query_parameters[key] = "";
+            _query_params[key].push_back("");
+            deprecated_query_parameters[key] = "";
         }
     } else {
         sstring key;
         sstring value;
         if (http::internal::url_decode(param.substr(0,split), key)
                 && http::internal::url_decode(param.substr(split + 1), value)) {
-            query_parameters[key] = std::move(value);
+            deprecated_query_parameters[key] = value;
+            _query_params[key].push_back(std::move(value));
         }
     }
 
 }
 
 sstring request::parse_query_param() {
+    http::internal::deprecated_query_parameters(*this).clear();
+    _query_params.clear();
     size_t pos = _url.find('?');
     if (pos == sstring::npos) {
         return _url;
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -56,6 +56,14 @@ public:
     }
 };
 
+// NOTE: Remove this once `query_parameters` is removed
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+std::unordered_map<sstring, sstring>& deprecated_query_parameters(request& r) noexcept {
+    return r.query_parameters;
+}
+#pragma GCC diagnostic pop
+
 SEASTAR_TEST_CASE(test_reply)
 {
     http::reply r;
@@ -218,12 +226,24 @@ SEASTAR_TEST_CASE(test_decode_url) {
     http::request req;
     req._url = "/a?q=%23%24%23";
     sstring url = req.parse_query_param();
+    const auto& query_parameters = deprecated_query_parameters(req);
     BOOST_REQUIRE_EQUAL(url, "/a");
     BOOST_REQUIRE_EQUAL(req.get_query_param("q"), "#$#");
+    BOOST_REQUIRE_EQUAL(query_parameters.at("q"), "#$#");
     req._url = "/a?a=%23%24%23&b=%22%26%22";
     req.parse_query_param();
     BOOST_REQUIRE_EQUAL(req.get_query_param("a"), "#$#");
+    BOOST_REQUIRE_EQUAL(query_parameters.at("a"), "#$#");
     BOOST_REQUIRE_EQUAL(req.get_query_param("b"), "\"&\"");
+    BOOST_REQUIRE_EQUAL(query_parameters.at("b"), "\"&\"");
+    req._url = "/a?b=%22%26%22&a=%21&b=%23%24%23&a=%23%24%23";
+    req.parse_query_param();
+    const auto& b = req.get_query_param_array("b");
+    auto expected_b = std::vector<sstring>{"\"&\"", "#$#"};
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(b.begin(), b.end(), expected_b.begin(), expected_b.end());
+    const auto& a = req.get_query_param_array("a");
+    auto expected_a = std::vector<sstring>{"!", "#$#"};
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(a.begin(), a.end(), expected_a.begin(), expected_a.end());
     return make_ready_future<>();
 }
 
@@ -1926,21 +1946,106 @@ SEASTAR_TEST_CASE(test_url_encode_decode) {
 
 SEASTAR_TEST_CASE(test_url_param_encode_decode) {
     http::request to_send;
+    auto& to_send_query_parameters = deprecated_query_parameters(to_send);
     to_send._url = "/foo/bar";
-    to_send.query_parameters["a"] = "a+a*a";
-    to_send.query_parameters["b"] = "b/b\%b";
+    to_send_query_parameters["a"] = "a+a*a";
+    to_send_query_parameters["b"] = "b/b\%b";
+    to_send_query_parameters["c"] = "";
 
     http::request to_recv;
     to_recv._url = to_send.format_url();
     sstring url = to_recv.parse_query_param();
 
     BOOST_REQUIRE_EQUAL(url, to_send._url);
-    BOOST_REQUIRE_EQUAL(to_recv.query_parameters.size(), to_send.query_parameters.size());
-    for (const auto& p : to_send.query_parameters) {
-        auto it = to_recv.query_parameters.find(p.first);
-        BOOST_REQUIRE(it != to_recv.query_parameters.end());
+    const auto& to_recv_query_parameters = deprecated_query_parameters(to_recv);
+    BOOST_REQUIRE_EQUAL(to_recv_query_parameters.size(), to_send_query_parameters.size());
+    for (const auto& p : to_send_query_parameters) {
+        auto it = to_recv_query_parameters.find(p.first);
+        BOOST_REQUIRE(it != to_recv_query_parameters.end());
         BOOST_REQUIRE_EQUAL(it->second, p.second);
     }
+
+    return make_ready_future<>();
+}
+
+SEASTAR_TEST_CASE(test_url_param_encode_decode_multiple) {
+    http::request to_send;
+    to_send._url = "/foo/bar";
+
+    to_send.set_query_param("a", {"a+a*a", "a/a\%a", "", "", "a a", "lasta"})
+        .set_query_param("b", {"b/b\%b", "", "lastb"});
+
+    http::request to_recv;
+    to_recv._url = to_send.format_url();
+    sstring url = to_recv.parse_query_param();
+    BOOST_REQUIRE_EQUAL(url, to_send._url);
+    const auto& to_send_a = to_send.get_query_param_array("a");
+    const auto& to_recv_a = to_recv.get_query_param_array("a");
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(to_send_a.begin(), to_send_a.end(), to_recv_a.begin(), to_recv_a.end());
+    const auto& to_send_b = to_send.get_query_param_array("b");
+    const auto& to_recv_b = to_recv.get_query_param_array("b");
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(to_send_b.begin(), to_send_b.end(), to_recv_b.begin(), to_recv_b.end());
+
+    BOOST_REQUIRE_EQUAL(to_recv.get_query_param("a"), to_send.get_query_param("a"));
+    BOOST_REQUIRE_EQUAL(to_recv.get_query_param("b"), to_send.get_query_param("b"));
+    return make_ready_future<>();
+}
+
+SEASTAR_TEST_CASE(test_url_param_empty) {
+    sstring test_url = "/foo/bar?key=v&key=&key&key2&key";
+    http::request req;
+    req._url = test_url;
+    req.parse_query_param();
+    const auto& query_params = deprecated_query_parameters(req);
+    BOOST_REQUIRE_EQUAL(query_params.size(), 2);
+    BOOST_REQUIRE(query_params.at("key").empty());
+    BOOST_REQUIRE(query_params.at("key2").empty());
+
+    std::vector<sstring> expected_key = {"v", "", "", ""};
+    const auto& actual_key = req.get_query_param_array("key");
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(actual_key.begin(), actual_key.end(), expected_key.begin(), expected_key.end());
+
+    std::vector<sstring> expected_key2 = {""};
+    const auto& actual_key2 = req.get_query_param_array("key2");
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(actual_key2.begin(), actual_key2.end(), expected_key2.begin(), expected_key2.end());
+
+    BOOST_REQUIRE_EQUAL(req.get_query_param("key"), "");
+    BOOST_REQUIRE_EQUAL(req.get_query_param("key2"), "");
+
+    return make_ready_future<>();
+}
+
+SEASTAR_TEST_CASE(test_url_params_get_set) {
+    http::request req;
+    req._url = "/foo/bar";
+    http::request::query_parameters_type params = {
+        {"a", {"a+a*a", "a/a%a", "a a", "lasta"}},
+        {"b", {"b/b%b", "lastb"}}
+    };
+
+    req.set_query_params(params);
+
+    const auto& req_params = req.get_query_params();
+    for (const auto&[key, values] : params) {
+        auto it = req_params.find(key);
+        BOOST_REQUIRE(it != req_params.end());
+        BOOST_REQUIRE_EQUAL_COLLECTIONS(it->second.begin(), it->second.end(), values.begin(), values.end());
+    }
+
+    BOOST_REQUIRE_EQUAL(req.get_query_param("a"), "lasta");
+    BOOST_REQUIRE_EQUAL(req.get_query_param("b"), "lastb");
+
+    req.set_query_param("a", "new_a");
+    BOOST_REQUIRE_EQUAL(req.get_query_param("a"), "new_a");
+
+    req.set_query_param("c", "c/c\%c");
+    BOOST_REQUIRE_EQUAL(req.get_query_param("c"), "c/c%c");
+
+    std::vector<sstring> d_params = {"d/d%d", "lastd"};
+    req.set_query_param("d", d_params);
+    BOOST_REQUIRE_EQUAL(req.get_query_param("d"), "lastd");
+    const auto& d_values = req.get_query_param_array("d");
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(d_values.begin(), d_values.end(), d_params.begin(), d_params.end());
 
     return make_ready_future<>();
 }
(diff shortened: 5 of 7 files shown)


src/http/request.cc
```diff
@@ -39,18 +40,27 @@ namespace seastar {
...
+            for (const auto& value : values) {
+                if (!value.empty()) {
+                    components.push_back(encoded_key + "=" + internal::url_encode(value));
+                }
+            }
         }
-        delim = "&";
     }
-    return _url + query;
+    return _url + "?" + fmt::to_string(fmt::join(components, "&"));
```
amnonh: Nitpick, you don't need the concatenation; you can put it all inside the fmt string

tarasveretilnyk: Done

include/seastar/http/request.hh
```diff
@@ -64,7 +64,7 @@ struct request {
     size_t content_length = 0;
     mutable size_t _bytes_written = 0;
     std::unordered_map<sstring, sstring, seastar::internal::case_insensitive_hash, seastar::internal::case_insensitive_cmp> _headers;
-    std::unordered_map<sstring, sstring> query_parameters;
+    std::unordered_map<sstring, std::vector<sstring>> query_parameters;
```
xemul: This is public member of request, so the change is potentially breaking one

xemul: Yeah, it _is_ breaking change. The struct request is used by http client users and they populate query parameters by directly accessing this field like `req.query_parameters["foo"] = "bar"`

tarasveretilnyk: @xemul I don’t see a good way to add the new functionality without changing the API. In any case, we would need to keep query_parameters the same type as it is now to avoid breaking backward compatibility. One possible solution could be to add a new field, like multiple_query_parameters, to store repeated values. In this case, query_parameters would still hold only the last value for each key, while multiple_query_parameters would keep all values. We would also need to define clear rules on how both fields should be used(and synchronized, because members can be changed directly) when building the URL and in other methods.
However, I think this approach is not good. It adds extra complexity and could make the struct harder to understand and use. Maybe you may have a better idea on how to solve this?

xemul: I would research several possible ways.

First, try to replace std::unordered_map with std::unordered_multimap. I don't know whether the latter is superior to the former from the API perspective, but if it is, it would be a nice solution.

Second, add std::unordered_map<sstring, std::vector<sstring>> next to existing map and deprecate the existing map member. Internally, squash both together. In a year we'll be able to drop the query_parameters thing. Good name for new member is a challenge here :)

Third, converts the map into std::unordered_map<sstring, query_parameters_list> where query_parameters_list is a vector-like class that provides compilation-time compatibility with existing query_parameters usage.


But I'd also like to outline the problem itself better. 

The struct request is used in two ways. First, http server receives a request, it parses it, fills in the struct request object and pushes it into a handler. Second, http client prepares a struct request object, populates it with headers, parameters, etc. and then passes it down to client::make_request.

Client seems simpler, let's start with the server. What currently happens if a request with `foo=bar&foo=buz` line arrives, how would parser "parse" it, exactly?

tarasveretilnyk: I think I misunderstood your first comment.
I thought you were referring to not breaking backward compatibility for all users of the library. In that case, any change to the type of query_parameters would be a breaking change and not an option. But if you meant it's fine to change the type as long as it doesn’t break our own code, then that simplifies the problem.

Regarding repetition of the same query parameter multiple times, there will be saved the last one. Currently, parse_query_param parse url sequentialy and insert each param into the map one by one.

xemul: > But if you meant it's fine to change the type as long as it doesn’t break our own code, then that simplifies the problem.

Not only our own. Any change should do its best not to break _compilation_ of 3rd party code

tarasveretilnyk: Okay, so based on your points, the first and third options you previously mentioned would inevitably break backward compatibility:

- Replacing std::unordered_map with std::unordered_multimap:
This approach introduces too many differences. For example, unordered_multimap doesn't even support the [] operator, which would immediately break a lot of existing code, including ours. It’s also a different type altogether, so any third-party code relying on decltype, templates, or type assumptions around query_parameters would likely break as well.
- Converting the map to std::unordered_map<sstring, query_parameters_list>
This is a bit more compatible, but still problematic. Any direct usage of sstring methods (e.g. req.query_parameters["key"].as_str()) would no longer compile. Also, the same issues with decltype and template-based code apply here too.

That leaves the second option, which is also what I initially proposed. While I don’t particularly like it due to the added complexity and potential for confusion, I think it’s the only viable way to introduce this functionality without breaking the existing API.

Here’s what I suggest:

- Kept the query_parameters field as-is, to preserve backward compatibility. It continues to store the last value for each parameter key.
- Introduced a new field query_params to store all values for each key.
- Made query_params private for now and added access methods (get_query_params, set_query_param, etc.) to interact with it.

My thinking is that this approach minimizes disruption. If users can achieve what they need through clean method calls, they’re less likely to be confused by the presence of both fields. Later, once query_parameters is deprecated and removed, we can make query_params public if needed.

Please let me know what do you think.

include/seastar/http/request.hh
```diff
@@ -65,7 +65,7 @@ struct request {
     mutable size_t _bytes_written = 0;
     std::unordered_map<sstring, sstring, seastar::internal::case_insensitive_hash, seastar::internal::case_insensitive_cmp> _headers;
     std::unordered_map<sstring, sstring> query_parameters;
-    httpd::parameters param;
+    httpd::parameters param; // deprecated: it is used to store last value of query parameters, but will be removed in the future
```
xemul: Huh? Why param is deprecated, but not the query_parameters?
Also -- comment is not enough, mind using the `[[deprecated("reason")]]` decoration

include/seastar/http/request.hh
```diff
@@ -77,7 +77,9 @@ struct request {
     std::unordered_map<sstring, sstring> chunk_extensions;
     sstring protocol_name = "http";
     noncopyable_function<future<>(output_stream<char>&&)> body_writer; // for client
-
+private:
+    std::unordered_map<sstring, std::vector<sstring>> query_params;
```
xemul: Private members should start with `_`

include/seastar/http/request.hh
```diff
@@ -109,15 +111,90 @@ struct request {
...
+
+        return "";
+    }
+
+    /**
+     * Search for all query parameters of a given key
+     * @param key the query parameter key
+     * @return a vector of all query parameter values, if it exists or an nullopt
+     */
+    std::optional<std::vector<sstring>> get_query_param_array(const sstring& key) const {
```
xemul: Why std::optional? Returning empty vector in case the key is not found seems to be fine

src/http/request.cc
```diff
@@ -39,15 +39,30 @@ namespace seastar {
...
+        } else {
+            for (const auto& val : values) {
+                query += key_component + internal::url_encode(val);
+            }
         }
         delim = "&";
     }
+
+    for (const auto& [key, value] : query_parameters) {
+        query += delim + internal::url_encode(key) + "=" + internal::url_encode(value);
```
xemul: Can the "value" be empty? It is handled by the existing code, but is lost after patching

src/http/request.cc
```diff
@@ -69,20 +84,28 @@ void request::add_query_param(std::string_view param) {
...
+            }
             query_parameters[key] = std::move(value);
         }
     }
 
 }
 
 sstring request::parse_query_param() {
+    query_parameters.clear();
+    query_params.clear();
```
xemul: Please explain why it's needed

src/http/request.cc
```diff
@@ -39,15 +39,30 @@ namespace seastar {
...
+            query += key_component;
+        } else {
+            for (const auto& val : values) {
+                query += key_component + internal::url_encode(val);
+            }
         }
         delim = "&";
     }
+
+    for (const auto& [key, value] : query_parameters) {
```
xemul: From the `add_query_param` patching I see that legacy query_parameters and new query_params duplicate each other (in a certain sense). With that, formatting a parsed URL would duplicate the query parameters in the output string, which is not nice

include/seastar/http/request.hh
```diff
@@ -109,15 +111,90 @@ struct request {
...
+        if (auto it = query_params.find(key); it != query_params.end()) {
+            query_params.erase(it);
+        }
+    }
+
+    /**
+     * Set query parameters in the request objects.
+     * @param params a map of query parameters, where the key is the parameter name and the value is a vector of values.
+     */
+    void set_query_params(const std::unordered_map<sstring, std::vector<sstring>>& params) {
```
xemul: I don't quite like this method. You walk the params anyway, so it can be obsoleted by the

```c++
    request& set_query_params(std::string_view key, std::initializer_list<std::string_view>> values) {
        ...; // assign query_params[key] = values and query_parameters[key] = tail(values)
        return *this;
    }
    request& set_query_params(std::string_view key, std::string_view val) {
        return set_query_params(key, {val});
    }
```

and then use it like

```c++
    req.set_query_params("foo", "bar").set_query_params("foo", {"buz", "fux"});
```

include/seastar/http/request.hh
```diff
@@ -109,15 +111,90 @@ struct request {
...
+
+        return res;
+    }
+
+    /**
+     * Set a query parameter value
+     * @param key the query parameter key
+     * @param value the query parameter value
+     */
+    void set_query_param(const sstring& key, const sstring& value) {
```
xemul: s/const sstring&/std::string_view/

include/seastar/http/request.hh
```diff
@@ -109,15 +111,90 @@ struct request {
...
+                }
+            }
+        }
+    }
+
+    /**
+     * Get all query parameters as a map, where the key is the parameter name and the value is a vector of values.
+     * @return a map of query parameters
+     */
+    std::unordered_map<sstring, std::vector<sstring>> get_query_params() const {
```
xemul: This method is not tested. Let's drop it, I don't think anyone will benefit from it, the get_query_param(_array)? should be enough, at least for now

include/seastar/http/request.hh
```diff
@@ -109,15 +111,90 @@ struct request {
...
+     * @return a vector of all query parameter values, if it exists or an nullopt
+     */
+    std::optional<std::vector<sstring>> get_query_param_array(const sstring& key) const {
+        std::vector<sstring> res;
+        if (auto it = query_params.find(key); it != query_params.end()) {
+            res = it->second;
+        }
+
+        if (auto old_it = query_parameters.find(key); old_it != query_parameters.end()) {
+            res.push_back(old_it->second);
```
xemul: Copying is not nice. Can it return `const vector<sstring>&` that'd point directly into query_params value?

src/http/request.cc
tarasveretilnyk: Legacy query_parameters and new query_params doesn't duplicate each other. My idea was to make legacy query_parameters responsible for the last values, as it was previously, and new query_params for the rest of the values. So in add_query_param I first push the value from legacy query_parameters into new query_params and than overwrite legacy query_parameters[key] = new_value.

include/seastar/http/request.hh
tarasveretilnyk: As I mentioned in another comment, my intention was to avoid duplicating the last value between the new query_params and the legacy query_parameters.

Returning a const vector<sstring>& would require changing the approach to store duplicates in both structures, which I want to avoid. That would introduce ambiguity: what should happen if query_parameters[key] is not equal to query_params[key].back()?

In that case, it becomes unclear what get_query_param(_array) should return, and how the URL should be generated from the conflicting values.

src/http/request.cc
tarasveretilnyk: That is because parse_query_param can be called multiple times. It processes query params sequentially and adds values to the containers each time it's called.
Without this change, if it's invoked twice, each parameter value will be added twice, resulting in duplicated entries in the final structure. We even have such usage in tests, where we reuse the same request entity just re-assigning the url.

src/http/request.cc
tarasveretilnyk: Yes, in that case it will produce &key=. In the previous patch, it would have produced just &key. Both forms are equivalent and represent an empty value.

avikivity:
Is parameter repetition specified somewhere (in some RFC)? If so please add a link to the patch changelog.

include/seastar/http/request.hh
```diff
@@ -109,15 +111,90 @@ struct request {
...
-     * @param key the query paramerter key
+     * @param key the query parameter key
      * @return the query parameter value, if it exists or empty string
      */
     sstring get_query_param(const sstring& key) const {
-        auto res = query_parameters.find(key);
-        if (res == query_parameters.end()) {
-            return "";
+        if(auto res = query_parameters.find(key); res != query_parameters.end()) {
+            return res->second;
```
nyh: Why do we need to look up both query_parameters and query_params?
You could always fill both query_parameters (for backward compatibility) and query_params (for new code), and the new code will always lookup just in query_params.

include/seastar/http/request.hh
nyh: I think it's better to duplicate the last item - it will also make it easier to eventually get rid of the deprecated API, with the current approach it will be hard work to finally deprecate it.

It will also improve the new API (make it avoid a copy).

But I agree for most intents and purposes, the extra copying is probably negligible for performance.

include/seastar/http/request.hh
tarasveretilnyk: I'm not sure what the expected behavior should be when query_parameters[key] differs from query_params[key].back().

When forming the URL and both maps are non-empty, how should we handle this? Should we combine both maps, and if so, how do we resolve conflicts (e.g., when query_parameters[key] != query_params[key].back())?

Alternatively, would it be acceptable to prioritize the new map - meaning that if query_params contains at least one key, we ignore query_parameters entirely?

Please let me know what do you think

include/seastar/http/request.hh
tarasveretilnyk: Done

include/seastar/http/request.hh
tarasveretilnyk: Done

src/http/request.cc
tarasveretilnyk: Done

include/seastar/http/request.hh
tarasveretilnyk: Done

include/seastar/http/request.hh
tarasveretilnyk: Done

tarasveretilnyk:
v5: Simplified handling of query parameters:

- Duplicated the last query parameter values into both legacy query_parameters and the new query_params.
- Fixed minor issues.
- Note: This introduces partial backward incompatibility for users who reuse the request instance after it has been parsed (via parse_query_param).
Specifically, if users modify legacy query_parameters after parsing and then generate a new URL, the updated form_url method will ignore legacy query_parameters if query_params contains data.

include/seastar/http/request.hh
tarasveretilnyk: I think that it can be useful for iterating through query params or using const methods from unordered_map. I simplified it and improved to return const&.

src/http/request.cc
tarasveretilnyk: Not relevant anymore.

include/seastar/http/request.hh
tarasveretilnyk: I initially avoided adding the attribute because it generates a lot of warnings, which makes it harder to debug actual compilation errors. However, I’ve included it in the current patch. If you prefer it removed, please let me know and I’ll take it out.

xemul: Please, let me rephrase. There are two concerns here

1. why did you mark `param` as deprecated, but not `query_parameters`?
2. please _do_ use `[[deprecated]]` marking, not just code comment

src/http/request.cc
xemul: Isn't it a pre-existing problem? Like -- without this change, what if `parse_query_params` is called multiple times, won't it overpopulate the existing map?

include/seastar/http/request.hh
```diff
@@ -109,15 +113,84 @@ struct request {
...
      * Search for the last query parameter of a given key
-     * @param key the query paramerter key
+     * @param key the query parameter key
      * @return the query parameter value, if it exists or empty string
      */
     sstring get_query_param(const sstring& key) const {
-        auto res = query_parameters.find(key);
-        if (res == query_parameters.end()) {
-            return "";
+        if(_query_params.empty()) {
```
xemul: nit: Why is this check needed?

include/seastar/http/request.hh
```diff
@@ -64,7 +65,7 @@ struct request {
     size_t content_length = 0;
     mutable size_t _bytes_written = 0;
     std::unordered_map<sstring, sstring, seastar::internal::case_insensitive_hash, seastar::internal::case_insensitive_cmp> _headers;
-    std::unordered_map<sstring, sstring> query_parameters;
+    [[deprecated("Use helper methods instead")]] std::unordered_map<sstring, sstring> query_parameters; // deprecated: it is used to store last value of query parameters, but will be removed in the future
```
xemul: nit: Multi-line would look nicer

include/seastar/http/request.hh
```diff
@@ -109,15 +113,84 @@ struct request {
...
+    }
+
+    /**
+     * Set a query parameter value
+     * @param key the query parameter key
+     * @param values the query parameter values
+     * @return a reference to this request object
+    */
+    request& set_query_param(std::string_view key, const std::vector<sstring>& values) {
+        _query_params[sstring(key)] = values;
```
xemul: nic: This copies thev ector. If the 2nd argument is a value and the assignment uses std::move() that'd be slightly better API

src/http/request.cc
```diff
@@ -69,20 +87,24 @@ void request::add_query_param(std::string_view param) {
     if (split >= param.length() - 1) {
         sstring key;
         if (http::internal::url_decode(param.substr(0,split) , key)) {
+            _query_params[key].push_back("");
```
xemul, line 104: Can this result in a bunch of empty strings pushed back to `_query_params[key]` with the same key?

include/seastar/http/request.hh
```diff
@@ -109,15 +113,84 @@ struct request {
...
+    }
+
+    /**
+     * Set a query parameter value
+     * @param key the query parameter key
+     * @param value the query parameter value
+     * @return a reference to this request object
+     */
+    request& set_query_param(std::string_view key, std::string_view value) {
+        return set_query_param(key, {value});
```
xemul, line 169: What if one does

```
    req.set_query_param("foo", "one").set_query_param("foo", "two");
```

?

In that case only the last one will be added, won't it?

include/seastar/http/request.hh
tarasveretilnyk: I got you right.
1. It is just my mistake, fixed it current patch
2. Also added in current patch, I just wanted to  warn you that [[deprecated] attribute generates a lot of warnings, so it complicates analyzing compilation issues.

src/http/request.cc
tarasveretilnyk: Yes, this is intentional. A URL like ?key=&key=&key&key... should be interpreted as an array of empty strings.

src/http/request.cc
tarasveretilnyk: Yes, however, this wasn’t as problematic before, because we were only storing the last value for each query parameter.
That meant calling this method multiple times with the same URL didn’t cause issues — each call would simply overwrite the previous value.

However, parsing a single URL like "/foo?key=v1&key2=v2&key=last" would result in the following map:
`query_parameters = { "key": "last", "key2": "v2" }`
Then, if we updated the URL like this:
`req.url = "/foo?key3=v3"`
and call parse_query_params again, query_parameters will look like this:
`query_parameters = { "key": "last", "key2": "v2", "key3": "v3" }`

include/seastar/http/request.hh
tarasveretilnyk: Done

include/seastar/http/request.hh
tarasveretilnyk: I don't think it's correct to look up the key in both maps. If the key isn't found in one of them, it shouldn't be in the other either. We could probably simplify that method and  just look it up in the new _query_params, ignoring the legacy query_parameters. In any case, after parse_query_params runs, query_parameters[key] will be equal to _query_params[key].last(). The only issue would arise if someone modifies _query_params and then, for some reason, calls get_query_param. But I believe that's an uncommon scenario. Will fix in the new patch.

include/seastar/http/request.hh
tarasveretilnyk: Yes, that’s the intended behavior - it will overwrite the entire value.

apps/httpd/main.cc
```diff
@@ -66,7 +66,7 @@ void set_routes(routes& r) {
         demo_json::my_object obj;
         obj.var1 = req.param.at("var1");
```
avikivity, line 67: s/repititions/repetitions/

include/seastar/http/request.hh
```diff
@@ -77,7 +78,10 @@ struct request {
     std::unordered_map<sstring, sstring> chunk_extensions;
     sstring protocol_name = "http";
     noncopyable_function<future<>(output_stream<char>&&)> body_writer; // for client
-
+    using query_param_type = std::unordered_map<sstring, std::vector<sstring>, seastar::internal::string_view_hash, std::equal_to<>>;
```
avikivity: Prefer full words.

include/seastar/http/request.hh
avikivity: `if`  is followed by a space in our coding style.

include/seastar/util/string_utils.hh
```diff
@@ -47,6 +47,19 @@ SEASTAR_MODULE_EXPORT_BEGIN
 // Collection of utilities for working with strings .
 //
 
+struct string_view_hash {
+  using is_transparent = void;
```
avikivity: Why half indents?

src/http/request.cc
```diff
@@ -40,14 +40,32 @@ namespace http {
 
 sstring request::format_url() const {
     sstring query = "";
-    sstring delim = "?";
-    for (const auto& p : query_parameters) {
-        query += delim + internal::url_encode(p.first);
-        if (!p.second.empty()) {
-            query += "=" + internal::url_encode(p.second);
+    if(!_query_params.empty()) {
```
avikivity: Coding style

src/http/request.cc
```diff
@@ -40,14 +40,32 @@ namespace http {
...
+         sstring delim = "&";
+        for (const auto& [key, values] : _query_params) {
+            auto key_component = delim + internal::url_encode(key) + "=";
+            if( values.empty()) {
```
avikivity: coding style

src/http/request.cc
```diff
@@ -40,14 +40,32 @@ namespace http {
...
+            query[0] = '?';
+        }
+    } else {
+        sstring delim = "?";
+        for (const auto& p : query_parameters) {
+            query += delim + internal::url_encode(p.first);
+            if (!p.second.empty()) {
+                query += "=" + internal::url_encode(p.second);
+            }
+            delim = "&";
```
avikivity, line 79: This is quadratic because sstring doesn't reserve space.

Need a better mechanism.

avikivity: I see it's a preexisting problem.

tests/unit/httpd_test.cc
```diff
@@ -1935,6 +1943,58 @@ SEASTAR_TEST_CASE(test_url_param_encode_decode) {
...
+    req._url = "/foo/bar";
+    http::request::query_param_type params = {
+        {"a", {"a+a*a", "a/a%a", "a a", "lasta"}},
+        {"b", {"b/b%b", "lastb"}}
+    };
+
+    req.set_query_params(params);
+
+    const auto& req_params = req.get_query_params();
+    for(const auto&[key, values] : params) {
```
avikivity: Coding style

apps/httpd/main.cc
tarasveretilnyk: Done

include/seastar/http/request.hh
tarasveretilnyk: Done

src/http/request.cc
tarasveretilnyk: Done

src/http/request.cc
tarasveretilnyk: Done

tests/unit/httpd_test.cc
tarasveretilnyk: Done

include/seastar/http/request.hh
tarasveretilnyk: Done

tarasveretilnyk:
v7: Minor improvements
v8: Rebased

include/seastar/http/request.hh
tarasveretilnyk: Done

include/seastar/http/request.hh
tarasveretilnyk: Done

include/seastar/util/string_utils.hh
tarasveretilnyk: Fixed

src/http/request.cc
xemul: Then I second what [was asked](https://github.com/scylladb/seastar/pull/2872#issuecomment-3163339113) -- please, reference something that justifies this behavior

include/seastar/http/request.hh
```diff
@@ -79,6 +81,10 @@ struct request {
     sstring protocol_name = "http";
     http::body_writer_type body_writer; // for client
 
+    using query_parameters_type = std::unordered_map<sstring, std::vector<sstring>, seastar::internal::string_view_hash, std::equal_to<>>;
```
xemul, line 84: Why doesn't it work with default hash template parameter? I see that you assign params with `_query_params[sstring(string_view_arg)]` anyway, so why string_view_hash is needed here at all?

tests/unit/httpd_test.cc
```diff
@@ -224,6 +224,14 @@ SEASTAR_TEST_CASE(test_decode_url) {
...
     BOOST_REQUIRE_EQUAL(req.get_query_param("a"), "#$#");
     BOOST_REQUIRE_EQUAL(req.get_query_param("b"), "\"&\"");
+    req._url = "/a?b=%22%26%22&a=%21&b=%23%24%23&a=%23%24%23";
+    req.parse_query_param();
+    const auto& b = req.get_query_param_array("b");
+    auto expected_b = std::vector<sstring>{"\"&\"", "#$#"};
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(b.begin(), b.end(), expected_b.begin(), expected_b.end());
+    const auto& a = req.get_query_param_array("a");
+    auto expected_a = std::vector<sstring>{"!", "#$#"};
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(a.begin(), a.end(), expected_a.begin(), expected_a.end());
```
xemul, line 246: Also need a check that legacy (now deprecated) query params still contains the last value

include/seastar/http/request.hh
```diff
@@ -110,15 +116,77 @@ struct request {
...
+    request& set_query_param(std::string_view key, std::vector<sstring> values) {
+        _query_params[sstring(key)] = std::move(values);
+        return *this;
+    }
+
+    /**
+     * Set the query parameters in the request objects.
+     * @param params a map of query parameters
+     */
+    void set_query_params(const query_parameters_type& params) {
```
xemul: Similarly to [this](https://github.com/scylladb/seastar/pull/2872#discussion_r2265778890), copying the map can be avoided

include/seastar/http/request.hh
```diff
@@ -110,15 +116,77 @@ struct request {
...
+        return *this;
+    }
+
+    /**
+     * Set a query parameter value
+     * @param key the query parameter key
+     * @param values the query parameter values
+     * @return a reference to this request object
+    */
+    request& set_query_param(std::string_view key, std::vector<sstring> values) {
```
xemul, line 189: There's no unit test that tests this overload, isn't it?

include/seastar/http/request.hh
tarasveretilnyk: It's needed for heterogeneous lookup (e.g., find). While insertions with heterogeneous types aren't supported yet, they will be added in C++26.
https://en.cppreference.com/w/cpp/container/unordered_map/operator_at

include/seastar/http/request.hh
tarasveretilnyk: Done

include/seastar/http/request.hh
tarasveretilnyk: Done

tests/unit/httpd_test.cc
tarasveretilnyk: Done

include/seastar/http/request.hh
xemul: Is it needed for potential users of get_query_params() that returns the map reference, or is it needed by this PR itself?

tarasveretilnyk: I used it in get_query_param and get_query_param_array, allowing those methods to perform lookups using string_view (_query_params.find(key)). This also benefits users who want to use string_view with find and contains when accessing get_query_params. 
`std::string_view v = "value";
req.get_query_params().contains(v);`

avikivity:
Please make sure the change doesn't add new deprecation warnings inside Seastar. If Seastar code touches the deprecated variable, it should temporarily suspend the deprecated warning.

tarasveretilnyk:
> Is parameter repetition specified somewhere (in some RFC)? If so please add a link to the patch changelog.

I didn't manage to find anything about parameter repetition in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4).

nyh:
> > Is parameter repetition specified somewhere (in some RFC)? If so please add a link to the patch changelog.
> 
> I didn't manage to find anything about parameter repetition in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4).

This is not the relevant RFC - in a URI, the "query" is an opaque string, and not necessarily parameters delimited by ampersands and so on. The convention of using `name=value` pairs separated by `&` is not defined by RFC 3986, it actually originates in HTML forms (`application/x-www-form-urlencoded`), see https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4:

> Control names and values are escaped. Space characters are replaced by `+', and then reserved characters are escaped as described in [RFC1738], section 2.2: Non-alphanumeric characters are replaced by `%HH', a percent sign and two hexadecimal digits representing the ASCII code of the character. Line breaks are represented as "CR LF" pairs (i.e., `%0D%0A').
> The control names/values are listed in the order they appear in the document. The name is separated from the value by `=' and name/value pairs are separated from each other by `&'.

Nowhere does it say what happens if the same name appears twice. It usually just means the original HTML form had two fields with the same name, I think this is perfectly fine. And if the value is empty, yes, it will look like "name=&...".

We need to be careful not to confuse HTML and URL encoding, we had a bunch of bugs related to that in the past, they are not the same thing.

avikivity:
> > Is parameter repetition specified somewhere (in some RFC)? If so please add a link to the patch changelog.
> 
> I didn't manage to find anything about parameter repetition in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4).

I asked Claude and it was super confused [1]. Apparently its customary but not specified.

[1] https://claude.ai/share/510a2876-ea2f-490e-9974-6e996891b5a3

nyh:
> > > Is parameter repetition specified somewhere (in some RFC)? If so please add a link to the patch changelog.
> > 
> > 
> > I didn't manage to find anything about parameter repetition in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4).
> 
> I asked Claude and it was super confused [1]. Apparently its customary but not specified.
> 
> [1] https://claude.ai/share/510a2876-ea2f-490e-9974-6e996891b5a3

I don't think it's confused at all. The answer is simply - the "name1=value1&name2=value2" syntax has nothing to do with the URI or HTTP standard, and everything to do with HTML and browsers. As Claude told you, "When HTML forms contain multiple elements with the same name attribute (like multiple checkboxes), browsers automatically generate URLs using the repeated parameter name format.". This is very much set in stone, for many years. 

Unfortunately, the people writing the HTML standards aren't as meticulous as the IETF with its RFCs, and the result is that their specifications (in this case, the HTML standard) aren't always ironclad, but I think the situation in this case is pretty clear, and has been clear for several decades, it's not that confusing.

nyh:
An interesting quote from https://url.spec.whatwg.org/#application/x-www-form-urlencoded

> The application/x-www-form-urlencoded format is in many ways an aberrant monstrosity, the result of many years of implementation accidents and compromises leading to a set of requirements necessary for interoperability, but in no way representing good design practices. In particular, readers are cautioned to pay close attention to the twisted details involving repeated (and in some cases nested) conversions between character encodings and byte sequences. Unfortunately the format is in widespread use due to the prevalence of HTML forms.

It then goes on to describe the way that the application/x-www-form-urlencoded GET "query" is parsed to, basically a list of name-value pairs.

include/seastar/http/request.hh
```diff
@@ -110,15 +119,77 @@ struct request {
...
+        _query_params[sstring(key)] = std::move(values);
+        return *this;
+    }
+
+    /**
+     * Set the query parameters in the request objects.
+     * @param params a map of query parameters
+     */
+    void set_query_params(query_parameters_type params) {
+        _query_params = std::move(params);
```
xemul, line 200: `return *this` to be consistent with other set_query_param-s

src/http/request.cc
```diff
@@ -63,26 +85,33 @@ future<> request::write_request_headers(output_stream<char>& out) const {
     });
 }
 
+// NOTE: Remove this once `query_parameters` is removed
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
xemul: That's too coarse. Please introduce the `std::unordered_map<sstring, sstring>& internal::deprecated_query_parameters(request& r) noexcept { return r.query_parameters };` method enclosed in those pragmas, and patch the rest of the code to use it.

include/seastar/http/request.hh
xemul: OK

src/http/request.cc
xemul: Given [this](https://github.com/scylladb/seastar/pull/2872#issuecomment-3180256713) observation, the intent is valid. In that case, there must be a unit test, that checks one.

And it looks like there's no test for `name=` parameter with empty value. I guess we should have one, but as it's a pre-existing problem, can be done separately from this work

avikivity:
> > > > Is parameter repetition specified somewhere (in some RFC)? If so please add a link to the patch changelog.
> > > 
> > > 
> > > I didn't manage to find anything about parameter repetition in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4).
> > 
> > 
> > I asked Claude and it was super confused [1]. Apparently its customary but not specified.
> > [1] https://claude.ai/share/510a2876-ea2f-490e-9974-6e996891b5a3
> 
> I don't think it's confused at all.


It wrote:

> When encoding parameters in a URL, you **do not** repeat the parameter name for multiple values of the same parameter. There are two standard approaches:
> 1. Multiple separate parameters (most common):
> https://example.com/search?tag=red&tag=blue&tag=green
> 
> 

So it contradicted itself immediately.

src/http/request.cc
tarasveretilnyk: Added empty values to existing tests

src/http/request.cc
tarasveretilnyk: Done

include/seastar/http/request.hh
tarasveretilnyk: Done

src/http/request.cc
xemul: I can't find the test that parses "/a?key=", "/a?key", "/a?key=&key=" and "/a?key&key", please point me to one

tarasveretilnyk: test_url_param_encode_decode and test_url_param_encode_decode_multiple, both first form url with empty value and than parse

xemul: Both check that multiple empty params are preserved, indeed, but not that both "?key=" and "?key" formats are supported

tarasveretilnyk: Added new test testcase for this

src/http/request.cc
tarasveretilnyk: Yes, since this performance issue is preexisting, I believe it would be more effective to address it as a separate optimization task. I also think it’s worth benchmarking different approaches to make sure that any changes are based on measurable improvements rather than assumptions

tarasveretilnyk:
@xemul please re-review

tarasveretilnyk:
@avikivity , @nyh , @xemul please take a look and consider merging if everything looks good to you

PR #2873, opened 2025-07-26
websocket: Remove unused scattered_message.hh inclusion

xemul:
That helper is not used by server.cc file that includes it


xemul:

PR #2874, opened 2025-07-26
websocket: Avoid sending scattered_message to output_stream

xemul:
There's a helper that sends a simple header and a temporary_buffer payload to the output_stream. Modern output_stream-s allow mixing buffered prefixes (the header) and zero-copy payloads (temp. buffer).

This change will allow to deprecate and remove net::packet from output_stream API eventually.

### DIFF
--- a/src/websocket/common.cc
+++ b/src/websocket/common.cc
@@ -42,7 +42,7 @@ future<> connection::handle_pong() {
     return make_ready_future<>();
 }
 
-future<> connection::send_data(opcodes opcode, temporary_buffer<char>&& buff) {
+future<> connection::send_data(opcodes opcode, temporary_buffer<char> buff) {
     char header[10] = {'\x80', 0};
     size_t header_size = 2;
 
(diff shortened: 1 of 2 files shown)


xemul:
upd:
- keep message buffer as value, not as rvalue ref

xemul:

src/websocket/common.cc
```diff
@@ -60,12 +60,9 @@ future<> connection::send_data(opcodes opcode, temporary_buffer<char>&& buff) {
...
 
-    scattered_message<char> msg;
-    msg.append(sstring(header, header_size));
-    msg.append(std::move(buff));
-    return _write_buf.write(std::move(msg)).then([this] {
-        return _write_buf.flush();
-    });
+    co_await _write_buf.write(header, header_size);
+    co_await _write_buf.write(std::move(buff));
+    co_await _write_buf.flush();
```
avikivity, line 65: The coroutine adds an allocation, but removal of scattered_message removes an allocation, so net zero. In general it's better not to introduce coroutines for functions that are expected not to yield.

avikivity: Although, perhaps, output_stream::flush() should yield with a good io_uring backend.

PR #1582, opened 2023-03-28
file_data_sink_impl: provide a put(seastar::net::packet) implementation.

nlescoua:
When writing a temporary buffer to an output_stream backed by a file_data_sink_impl the implementation of put(seastar::net::packet) is called and abort is called. Provide an implementation that forward calls to put(seastar::temporary_buffer).

Close #1581

PR #2875, opened 2025-07-26
tls: Push iovec with the help of put(vector<temporary_buffer>)

xemul:
This change aims at removing net::packet from output_stream API eventually. One of the places that still needs it is tls session vec_push() helper that copies the passed iovec onto scattered_message (which is a wrapper over net::packet) and then put()-s it into the lower data_dink.

This patch replaces the scattered_message with vector of temporary_buffer-s. The data_sink::put() will then transparently convert them to data_sink_impl's net::packet anyway.

"While at it" -- reserve the vector to avoid re-allocations upon appending buffers. This could've been done with scattered message too.

### DIFF
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -1569,12 +1569,14 @@ public:
             ssize_t n; // Set on the good path and unused on the bad path
 
             if (!_output_pending.failed()) {
-                scattered_message<char> msg;
+                std::vector<temporary_buffer<char>> msg;
+                msg.reserve(iovcnt);
+                n = 0;
                 for (int i = 0; i < iovcnt; ++i) {
-                    msg.append(std::string_view(reinterpret_cast<const char *>(iov[i].iov_base), iov[i].iov_len));
+                    msg.emplace_back(reinterpret_cast<const char*>(iov[i].iov_base), iov[i].iov_len);
+                    n += iov[i].iov_len;
                 }
-                n = msg.size();
-                _output_pending = _out.put(std::move(msg).release());
+                _output_pending = _out.put(std::move(msg));
             }
             if (_output_pending.failed()) {
                 // exception is copied back into _output_pending


src/net/tls.cc
```diff
@@ -1569,12 +1569,14 @@ class session : public enable_lw_shared_from_this<session> {
             ssize_t n; // Set on the good path and unused on the bad path
 
             if (!_output_pending.failed()) {
-                scattered_message<char> msg;
+                std::vector<temporary_buffer<char>> msg;
+                msg.reserve(iovcnt);
```
avikivity, line 1573: It's likely we arrive here even with iovcnt = 1, so can optimize by calling data_sink::put(temporary_buffer) in that case.

src/net/tls.cc
```diff
@@ -1569,12 +1569,14 @@ class session : public enable_lw_shared_from_this<session> {
...
+                n = 0;
                 for (int i = 0; i < iovcnt; ++i) {
-                    msg.append(std::string_view(reinterpret_cast<const char *>(iov[i].iov_base), iov[i].iov_len));
+                    msg.emplace_back(reinterpret_cast<const char*>(iov[i].iov_base), iov[i].iov_len);
```
avikivity, line 1576: We should figure out how to avoid the copy.

src/net/tls.cc
xemul: Yes. I have a draft branch that eliminates net::packet from iostreams and uses `boost::small_vector<temporary_buffer, 1>`. This small vector also handles this place (seemingly) nicely

nyh:
Would have been nice to have an issue that summarizes why you want to get rid of net::packet and what is the overall plan to doing this. We discussed this many times over the years - see for example this Seastar-dev thread from two years ago https://groups.google.com/g/seastar-dev/c/_sfnFJVOTic/m/kZTNlndKAwAJ this PR which was frozen because it added net:packet https://github.com/scylladb/seastar/pull/1582 

In the first link I mentioned a few replacements to net::packet, one of the is indeed vector<temporary_buffer>. It appears in this patch that this is indeed your choice. But it would have been nice to have an issue which explicitly says that this is the choice going forward, so people can explicitly agree (or disagree) with your choice, instead of getting it out-of-the-blue in separate patches.

PR #2879, opened 2025-07-28
io_queue: Narrow down friendship with reactor

xemul:
The io_queue is reactor friend to update io_stats member. For such cases there's reactor::io_stats::local() helper, so this friendship can be more fine-grained.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -179,7 +179,6 @@ private:
     friend class reactor_backend_aio;
     friend class reactor_backend_uring;
     friend class reactor_backend_selector;
-    friend class io_queue; // for aio statistics
     friend struct reactor_options;
     friend class aio_storage_context;
 public:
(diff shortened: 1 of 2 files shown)


include/seastar/core/reactor.hh
```diff
@@ -205,6 +204,7 @@ public:
     private:
         friend class file_data_source_impl;
         friend void io_completion::complete_with(ssize_t);
+        friend class io_queue;
         static io_stats& local() noexcept;
```
avikivity, line 208: This should be inlined to restore fast access.

xemul: It's inlined when defined

https://github.com/scylladb/seastar/blob/66ecd34f2aff7607814a26432a02a441e8d25403/include/seastar/core/reactor.hh#L777-L779

avikivity: Ah, I wrongly assumed it's out-of-line.

PR #2884, opened 2025-07-29
http::reply::status_type: Promote to namespace and make extensible struct

elcallio:
Refs #884

Moves HTTP status codes to top-level type "status_type", and changes it from enum to struct.
The former to be able to forward declare the type.
The latter to be able to type-safely extend the type (i.e. define a typed constant in a different compilation unit, without having to modify a seastar enum).

"Extending" is done by simply constexpr-declaring your custom status in a code unit, then adding a dllink-time constant for static init registration, i.e:

```
.hh:

constexpr seastar::http::status_type MY_STATUS = <number>;

.cc

static const auto my_status_label = seastar::http::bind_status_name(MY_STATUS, "Some textual name");
```

Only valid in static init time.

Motivation:
1.) Using custom values for an enum type is oxymoronic. While it is workable to just cast int:s back and forth into the enum-values fields for status_type, it smells of bad practice. While replacing this with an admittedly weakly aliasing wrapper type is not much more type-safe, at least it is consistent.
2.) Some aspects of seastar level code relies (for some definition thereof) on translating status codes to strings, mainly pretty-printing a received reply (less important) and sending such a reply (see write_reply_to_connection). While a stringized value for the status is not mandatory, it is nice to be consistent, and get all info at all time without reinventing wheels.

Note: contains some haxxor to make the compile-time API as compatible as possible, i.e. include constexpr constants for all existing codes, as well as ensuring old code compiles (doing underlying_type_t stuff etc).

### DIFF
--- a/include/seastar/http/reply.hh
+++ b/include/seastar/http/reply.hh
@@ -94,6 +94,12 @@ struct status_type {
         }
     };
 
+    // The following list of status codes is part of the HTTP standard,
+    // and are defined in RFC 9110, and in a few case in older RFCs as
+    // listed in IANA's "HTTP Status Code Registry". Please do not add
+    // to this list non-standard error codes. Seastar applications should
+    // be able to use non-standard error codes, but shouldn't expect
+    // Seastar to give them official names.
     static constexpr status_init continue_{100}; //!< continue
     static constexpr status_init switching_protocols{101}; //!< switching_protocols
     static constexpr status_init ok{200}; //!< ok
@@ -106,6 +112,7 @@ struct status_type {
     static constexpr status_init multiple_choices{300}; //!< multiple_choices
     static constexpr status_init moved_permanently{301}; //!< moved_permanently
     static constexpr status_init moved_temporarily{302}; //!< moved_temporarily
+    static constexpr status_init found{moved_temporarily}; //!< found is modern name for moved_temporarily
     static constexpr status_init see_other{303}; //!< see_other
     static constexpr status_init not_modified{304}; //!< not_modified
     static constexpr status_init use_proxy{305}; //!< use_proxy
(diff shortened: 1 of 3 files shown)


src/http/reply.cc
```diff
@@ -54,69 +55,88 @@ namespace http {
...
-    {reply::status_type::bad_gateway, "502 Bad Gateway"},
-    {reply::status_type::service_unavailable, "503 Service Unavailable"},
-    {reply::status_type::gateway_timeout, "504 Gateway Timeout"},
-    {reply::status_type::http_version_not_supported, "505 HTTP Version Not Supported"},
-    {reply::status_type::insufficient_storage, "507 Insufficient Storage"},
-    {reply::status_type::bandwidth_limit_exceeded, "509 Bandwidth Limit Exceeded"},
-    {reply::status_type::network_read_timeout, "598 Network Read Timeout"},
-    {reply::status_type::network_connect_timeout, "599 Network Connect Timeout"}};
+static auto& status_strings() {
+    static std::unordered_map<status_type, std::string_view> status_strings = {
```
xemul, line 59: Why can't this map remain a ~~global~~ static file-local variable?

include/seastar/http/reply.hh
```diff
@@ -118,15 +125,19 @@ struct status_type {
...
     static constexpr status_init conflict{409}; //!< conflict
     static constexpr status_init gone{410}; //!< gone
     static constexpr status_init length_required{411}; //!< length_required
+    static constexpr status_init precondition_failed{412}; //!< precondition_failed
     static constexpr status_init payload_too_large{413}; //!< payload_too_large
     static constexpr status_init uri_too_long{414}; //!< uri_too_long
     static constexpr status_init unsupported_media_type{415}; //!< unsupported_media_type
+    static constexpr status_init range_not_satisfiable{416}; //!< range_not_satisfiable
     static constexpr status_init expectation_failed{417}; //!< expectation_failed
     static constexpr status_init page_expired{419}; //!< page_expired
```
xemul, line 139: Can those non-standard codes be marked as `[[deprecated(...)]]` early, to be removed eventually, but also early?

tests/unit/httpd_test.cc
```diff
@@ -2178,3 +2178,45 @@ SEASTAR_TEST_CASE(test_client_close_connection) {
         }
     });
 }
+
+static constexpr http::status_type my_http_status(555);
+static const auto my_http_status_name = http::bind_status_name(my_http_status, "My Super Error");
```
xemul, line 2183: Maybe define

```c++
status_type::status_type(int v, std::string_view name)
    : status_type(v)
{
    bind_status_name(name);
}
```

for one-line registration?

src/http/reply.cc
elcallio: Because we want to modify it at dllink time. If it was global, it would be subject to static init order fiasco. By placing it in a function we can ensure we init it before modifying.

include/seastar/http/reply.hh
elcallio: Probably a good idea.

tests/unit/httpd_test.cc
elcallio: That would make the constructor non-constexpr, and thus make the status_type instance non-constexpr. 
I also opted against that to make the pattern:
```
static const auto name_constant = bind_status_name(status_const, "some string");
```
work.

src/http/reply.cc
xemul: OK. Please add a code comment about it, because it's not immediately obvious

tests/unit/httpd_test.cc
xemul: That'd make only this `(int v, std::string_view name)` constructor non-constexpr, and why is _that_ a problem? Those custom status codes will behave like the "my_http_status_name" from this patch does

elcallio: It is (perhaps) a problem because it disables switches on the status code instance. While another downside of this patch is that said switches degenerates to int:s, you can still mix and match all codes in a single switch if need be.

xemul: Indeed :(

```
/home/xemul/src/seastar/tests/unit/httpd_test.cc: In member function ‘virtual seastar::future<> test_formatting_custom_error::run_test_case() const’:
/home/xemul/src/seastar/tests/unit/httpd_test.cc:2260:10: error: the value of ‘my_http_status’ is not usable in a constant expression
 2260 |     case my_http_status: break; // ok
      |          ^~~~~~~~~~~~~~
```

PR #2886, opened 2025-07-30
Add http2 support. Start with http2 frame parsing logic

muthu90tech:
I would like to work on the http2 support in seastar, mentioned [here](https://github.com/scylladb/seastar/issues/734)
 Since this is a large task, I want to start small get feedback and iterate. Eventually making it compliant with [http2 spec](https://datatracker.ietf.org/doc/html/rfc7540)

### DIFF
--- a/include/seastar/http/http2_connection.hh
+++ b/include/seastar/http/http2_connection.hh
@@ -1,0 +1,93 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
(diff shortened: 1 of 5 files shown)


tests/unit/httpd_test.cc
```diff
@@ -1,7 +1,8 @@
 /*
  * Copyright 2015 Cloudius Systems
  */
-
+    
+#define SEASTAR_HTTP2_TEST
```
muthu90tech, line 5: Would like a better way to expose private functions for test purposes.

include/seastar/http/http2_connection.hh
```diff
@@ -0,0 +1,93 @@
...
+#include <seastar/http/request.hh>
+#include <seastar/http/reply.hh>
+#include <seastar/http/routes.hh>
+#include <seastar/http/http2_frame_parser.hh>
+
+namespace seastar {
+
+namespace http {
+
+class http2_connection {
```
xemul, line 37: This du[licates (at the first glance) the client.hh conneciton class

include/seastar/http/http2_connection.hh
```diff
@@ -0,0 +1,93 @@
...
+        : _routes(routes)
+        , _fd(std::move(fd))
+        , _read_buf(_fd.input())
+        , _write_buf(_fd.output())
+    {}
+
+    // Main entry point to start processing the HTTP/2 connection.
+    future<> process();
+
+    #ifdef SEASTAR_HTTP2_TEST
```
xemul, line 73: We try to avoid ifdefs for no special readons. If the intent is just to get a private field out of a class, declare a method in internal or testing namespace that's friend to that class

xemul:
We have both -- http server and client implementation in seastar. Does support for http2 come for both as well?

PR #2888, opened 2025-07-30
reactor: coroutinize some file-related functions

avikivity:
File-related functions have a mandatory yield when executing
the system call, so converting them to a coroutine is not a loss
in terms of the number of allocations and is a win in terms of
readability.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1722,138 +1722,118 @@ size_t sanitize_iovecs(std::vector<iovec>& iov, size_t disk_alignment) noexcept 
 future<file>
 reactor::open_file_dma(std::string_view nameref, open_flags flags, file_open_options options) noexcept {
     auto open_flags = static_cast<int>(flags);
-    {
-        sstring name(nameref);
-        syscall_result_extra<struct stat> sr = co_await _thread_pool->submit<syscall_result_extra<struct stat>>(
-                internal::thread_pool_submit_reason::file_operation, [this, name, &open_flags, &options, strict_o_direct = _cfg.strict_o_direct, bypass_fsync = _cfg.bypass_fsync] () mutable {
-            // We want O_DIRECT, except in three cases:
-            //   - tmpfs (which doesn't support it, but works fine anyway)
-            //   - strict_o_direct == false (where we forgive it being not supported)
-            //   - kernel_page_cache == true (where we disable it for short-lived test processes)
-            // Because open() with O_DIRECT will fail, we open it without O_DIRECT, try
-            // to update it to O_DIRECT with fcntl(), and if that fails, see if we
-            // can forgive it.
-            auto is_tmpfs = [] (int fd) {
-                struct ::statfs buf;
-                auto r = ::fstatfs(fd, &buf);
-                if (r == -1) {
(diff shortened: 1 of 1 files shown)


avikivity:
v2:
 - eliminate unnecessary conversion in chmod()

src/core/reactor.cc
```diff
@@ -1796,31 +1796,32 @@ reactor::open_file_dma(std::string_view nameref, open_flags flags, file_open_opt
...
-reactor::remove_file(std::string_view pathname) noexcept {
-    // Allocating memory for a sstring can throw, hence the futurize_invoke
-    return futurize_invoke([this, pathname] {
-        return _thread_pool->submit<syscall_result<int>>(
-                internal::thread_pool_submit_reason::file_operation, [pathname = sstring(pathname)] {
+reactor::remove_file(std::string_view pathname_view) noexcept {
+    auto pathname = sstring(pathname_view);
+    {
+        syscall_result<int> sr = co_await _thread_pool->submit<syscall_result<int>>(
+                internal::thread_pool_submit_reason::file_operation, [pathname] {
```
xemul: Can be a reference to avoid filename copying

src/core/reactor.cc
```diff
@@ -1796,31 +1796,32 @@ reactor::open_file_dma(std::string_view nameref, open_flags flags, file_open_opt
...
-    // Allocating memory for a sstring can throw, hence the futurize_invoke
-    return futurize_invoke([this, old_pathname, new_pathname] {
-        return _thread_pool->submit<syscall_result<int>>(
-                internal::thread_pool_submit_reason::file_operation, [old_pathname = sstring(old_pathname), new_pathname = sstring(new_pathname)] {
+reactor::rename_file(std::string_view old_pathname_view, std::string_view new_pathname_view) noexcept {
+    auto old_pathname = sstring(old_pathname_view);
+    auto new_pathname = sstring(new_pathname_view);
+    {
+        syscall_result<int> sr = co_await _thread_pool->submit<syscall_result<int>>(
+                internal::thread_pool_submit_reason::file_operation, [old_pathname, new_pathname] {
```
xemul: Both, old_ and new_ pathname-s can be references not to copy strings

src/core/reactor.cc
```diff
@@ -1796,31 +1796,32 @@ reactor::open_file_dma(std::string_view nameref, open_flags flags, file_open_opt
 }
 
 future<>
-reactor::remove_file(std::string_view pathname) noexcept {
-    // Allocating memory for a sstring can throw, hence the futurize_invoke
-    return futurize_invoke([this, pathname] {
-        return _thread_pool->submit<syscall_result<int>>(
-                internal::thread_pool_submit_reason::file_operation, [pathname = sstring(pathname)] {
+reactor::remove_file(std::string_view pathname_view) noexcept {
```
xemul, line 1795: nit: Patch title says "... rename_file" (which also in this patch though)

src/core/reactor.cc
xemul: It's pretty much everywhere in the set, but since it's a preexisting problem, let's leave it for a (optional) followup

PR #2890, opened 2025-07-31
reactor: coroutinize spawn()

avikivity:
The do_with() makes it an immediate win, though the real win is readability.


avikivity:
https://github.com/scylladb/seastar/pull/2891

avikivity:
v2:
 - rebased past a58fbd1e938313eb9876300b434cd6c1146d6d82

PR #2895, opened 2025-08-01
reactor: Use writeable_eventfd for reactor::_notify_eventfd

xemul:
The member in question is initialized with eventfd descriptor and is then used by reactor to wakeup backend from sleep. The writeable_eventfd class provides "compatible" signalling facility that can be re-used.

The reading part of writeable_eventfd is readable_eventfd, which is too much for the backend usage, so the writeable_eventfd::consume() helper is added. Next, the read-side of the eventfd nowdays works via fd_kernel_completion-s inheritants, that are patched to call the newly introduced method.

In the end of the day, the patch generalizes the "protocol" that notify eventfd peers use to kick each other -- the 8 bytes dummy tranfer.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1026,12 +1026,11 @@ reactor::reactor(std::shared_ptr<seastar::smp> smp, alien::instance& alien, unsi
     : _smp(std::move(smp))
     , _alien(alien)
     , _cfg(std::move(cfg))
-    , _notify_eventfd(file_desc::eventfd(0, EFD_CLOEXEC))
     , _task_quota_timer(file_desc::timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC))
     , _id(id)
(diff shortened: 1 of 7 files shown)


src/core/reactor.cc
```diff
@@ -3748,6 +3744,11 @@ void writeable_eventfd::signal(size_t count) {
     SEASTAR_ASSERT(r == sizeof(c));
 }
 
+std::optional<size_t> writeable_eventfd::consume() {
+    uint64_t c;
+    return _fd.read(&c, sizeof(c));
+}
```
avikivity, line 3750: Shouldn't this be part of the read side?

xemul: The read-side uses await-able wrapper around pollable_fd, which is not needed by the users of this new method. Actually the writeable_eventfd and readable_eventfd are supposed to play with each other, while reactor only needs the writeable end. I'll think on it a bit more

PR #2896, opened 2025-08-01
rpc: client: convert main loop to a coroutine

avikivity:
This both improves readability, and saves an allocation for a
continuation for every packet we have to wait for.

### DIFF
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -5,6 +5,7 @@
 #include <seastar/core/print.hh>
 #include <seastar/core/future-util.hh>
 #include <seastar/core/metrics.hh>
+#include <seastar/coroutine/as_future.hh>
 #include <seastar/util/assert.hh>
 #include <boost/range/adaptor/map.hpp>
 #include <boost/range/adaptor/transformed.hpp>
@@ -944,7 +945,9 @@ namespace rpc {
   }
 
   future<> client::loop(client_options ops, const socket_address& addr, const socket_address& local) {
-      return _socket.connect(addr, local).then([this, ops = std::move(ops)] (connected_socket fd) {
+      std::exception_ptr ep;
+      try {
+          connected_socket fd = co_await _socket.connect(addr, local);
           fd.set_nodelay(ops.tcp_nodelay);
           if (ops.keepalive) {
               fd.set_keepalive(true);
@@ -969,14 +972,17 @@ namespace rpc {
               features[protocol_features::ISOLATION] = _options.isolation_cookie;
           }
 
-          return negotiate_protocol(std::move(features)).then([this] {
+          co_await negotiate_protocol(std::move(features));
+          {
               _propagate_timeout = !is_stream();
               set_negotiated();
-              return do_until([this] { return _read_buf.eof() || _error; }, [this] () mutable {
+              while (!_read_buf.eof() && !_error) {
                   if (is_stream()) {
-                      return handle_stream_frame();
+                      co_await handle_stream_frame();
+                      continue;
                   }
-                  return read_response_frame_compressed(_read_buf).then([this] (response_frame::return_type msg_id_and_data) {
+                  response_frame::return_type msg_id_and_data = co_await read_response_frame_compressed(_read_buf);
+                  {
                       auto& msg_id = std::get<0>(msg_id_and_data);
                       auto& data = std::get<2>(msg_id_and_data);
                       auto it = _outstanding.find(std::abs(msg_id));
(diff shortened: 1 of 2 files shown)


src/rpc/rpc.cc
```diff
@@ -5,6 +5,7 @@
 #include <seastar/core/print.hh>
```
gleb-cloudius, line 5: Why do you say that `read_response_frame_compressed` will often yield? This is a common read from a socket loop which was previously shown to have less allocations with continuation style. What is different here? (I would rewrite all of them with co-routines anyway, lets be honest if they were written today that's how it would have been done anyway).

avikivity: In saturating loads (throughput bound) it can be the case that every read will find a packet already waiting. But most loads aren't saturating (latency oriented) and we'll often wait for a packet. This becomes even more common as the cluster size grows, since the same number of requests are sent to more servers.

Note even in a saturating load we don't lose here, since the allocation is amortized over all packets. If we do wait we win one allocation per packet that we wait for.

wrt coroutinizing everything, I think it's reasonable once we apply [[clang::coro_await_elidable]], it will be better than the continuation style in every way.

gleb-cloudius: > In saturating loads (throughput bound) it can be the case that every read will find a packet already waiting. But most loads aren't saturating (latency oriented) and we'll often wait for a packet. This becomes even more common as the cluster size grows, since the same number of requests are sent to more servers.
> 
> Note even in a saturating load we don't lose here, since the allocation is amortized over all packets. If we do wait we win one allocation per packet that we wait for.

I do no see why the same argument does not apply for cql server co-coutization. 

> 
> wrt coroutinizing everything, I think it's reasonable once we apply [[clang::coro_await_elidable]], it will be better than the continuation style in every way.

What do we need to apply it?

src/rpc/rpc.cc
```diff
@@ -1028,24 +1033,31 @@ namespace rpc {
...
               }
           }
+      }
+      {
           if (is_stream() && (ep || _error)) {
               _stream_queue.abort(std::make_exception_ptr(stream_closed()));
           }
           _error = true;
-          return stop_send_loop(ep).then_wrapped([this] (future<> f) {
+          future<> f = co_await coroutine::as_future(stop_send_loop(ep));
```
gleb-cloudius: Why do it this way instead of:

```
try{
  (void)co_await stop_send_loop();
} catch (...) {}
```

src/rpc/rpc.cc
avikivity: > > In saturating loads (throughput bound) it can be the case that every read will find a packet already waiting. But most loads aren't saturating (latency oriented) and we'll often wait for a packet. This becomes even more common as the cluster size grows, since the same number of requests are sent to more servers.
> > Note even in a saturating load we don't lose here, since the allocation is amortized over all packets. If we do wait we win one allocation per packet that we wait for.
> 
> I do no see why the same argument does not apply for cql server co-coutization.

It does apply (each individual change needs to be inspected for its own merit).

> 
> > wrt coroutinizing everything, I think it's reasonable once we apply [[clang::coro_await_elidable]], it will be better than the continuation style in every way.
> 
> What do we need to apply it?

Fix tests that fail when it's applied.

src/rpc/rpc.cc
```diff
@@ -969,14 +972,17 @@ namespace rpc {
...
               set_negotiated();
-              return do_until([this] { return _read_buf.eof() || _error; }, [this] () mutable {
+              while (!_read_buf.eof() && !_error) {
                   if (is_stream()) {
-                      return handle_stream_frame();
+                      co_await handle_stream_frame();
+                      continue;
                   }
-                  return read_response_frame_compressed(_read_buf).then([this] (response_frame::return_type msg_id_and_data) {
+                  response_frame::return_type msg_id_and_data = co_await read_response_frame_compressed(_read_buf);
```
gleb-cloudius: `auto& [msg_id, ht,  data] = co_await read_response_frame_compressed(_read_buf);`?

src/rpc/rpc.cc
avikivity: It's how I translate then_wrapped. try/catch would work too.

I think it's slightly better - below we call f.ignore_ready_future(), so we avoid a try/catch that doesn't need the exception.

gleb-cloudius: OK. I personally like mixing features and co-routines less if not needed.

src/rpc/rpc.cc
gleb-cloudius: > > > In saturating loads (throughput bound) it can be the case that every read will find a packet already waiting. But most loads aren't saturating (latency oriented) and we'll often wait for a packet. This becomes even more common as the cluster size grows, since the same number of requests are sent to more servers.
> > > Note even in a saturating load we don't lose here, since the allocation is amortized over all packets. If we do wait we win one allocation per packet that we wait for.
> > 
> > 
> > I do no see why the same argument does not apply for cql server co-coutization.
> 
> It does apply (each individual change needs to be inspected for its own merit).

Well, when I sent cql server co-routization patch set you asked me to run saturation benchmark and compare allocations before and after...

> 
> > > wrt coroutinizing everything, I think it's reasonable once we apply [[clang::coro_await_elidable]], it will be better than the continuation style in every way.
> > 
> > 
> > What do we need to apply it?
> 
> Fix tests that fail when it's applied.

Where do you apply it. I thought it supposed to be applied in a specific places where we know that co-routine is "simple", not in a generic code.

src/rpc/rpc.cc
avikivity: It makes sense, but better as a follow up that as part of coroutinization.

gleb-cloudius (approved):

PR #2898, opened 2025-08-03
foreign_ptr: add unwrap_on_owner_shard method

bhalevy:
The motivation for this new method as to allow
the user to check the wrapped pointer use_count
on the owner shard before e.g. clearing it,
if the use_count equals 1.

Refs scylladb/scylladb#25026


bhalevy:
@avikivity / @nyh please consider approving the PR and merge if okay. (it replaced #2861 due to my human error)

PR #2899, opened 2025-08-03
reactor: coroutinize more file related functions

avikivity:
Mostly improving readability.


avikivity:
The problem is that coroutine allocation doesn't disable memory failure injection while allocating the coroutine frame (as we do when allocating a continuation).

avikivity:
v2:
 - rebase past fd2710e0b80e0f53d7b7a493dd4655a351e9e0af

PR #2900, opened 2025-08-03
coroutine: allocate coroutine frame in a critical section

avikivity:
Like continuations, coroutines are glue that cannot be allowed to fail. For example, if cleanup coroutine fails to allocate and returns an exception future, cleanup will not be done and the system will enter an undefined state. Better to crash.

This conflicts with test code that tries to inject memory failures and see how they are handled (file_io_test.cc handle_bad_alloc_test). The coroutine will throw std::bad_alloc (since we don't define get_return_object_on_allocation_failure()), and since it's declared noexcept, will call std::terminate.

To avoid all this, follow future::schedule() and prevent memory allocation failure injection from interfering with coroutine frame allocation. In this regard a coroutine frame is just like a continuation.


avikivity:
Helps https://github.com/scylladb/seastar/pull/2899

avikivity:
v2: don't over-allocate (bad copilot)

include/seastar/core/coroutine.hh
```diff
@@ -85,6 +85,11 @@ public:
         scheduling_group set_scheduling_group(scheduling_group sg) noexcept {
             return std::exchange(this->_sg, sg);
         }
+
+        static void* operator new(size_t size) {
+            memory::scoped_critical_alloc_section _;
+            return ::operator new(size + sizeof(promise_type));
```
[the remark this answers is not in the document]
avikivity: You wrote that code.

avikivity:
v3:
 - extract into helper class
 - supply matching operator delete (gcc seemed to want it) and matching sized delete.

kelbon:
This breaks code that previously handled future allocation errors correctly.

avikivity:
Futures don't allocate.

kelbon:
> Futures don't allocate.

So why you added operator new/ operator delete in this pull request?

avikivity:
> > Futures don't allocate.
> 
> So why you added operator new/ operator delete in this pull request?

Because coroutine frames are allocated.

avikivity:
v3: rebase, mainly to see the failing tests again

avikivity:
v4: switch to malloc/free to avoid gcc being too clever for its own good (explanation in patches)

include/seastar/core/coroutine.hh
```diff
@@ -51,10 +53,26 @@ execute_involving_handle_destruction_in_await_suspend(std::invocable<> auto&& fu
...
+class coroutine_allocators {
+public:
+    static void* operator new(size_t size) {
+        memory::scoped_critical_alloc_section _;
+        return ::malloc(size);
+    }
+    static void operator delete(void* ptr) noexcept {
+        ::free(ptr);
+    }
+#ifdef __cpp_sized_deallocation
```
kelbon, line 73: Why did you write ifdef here?

include/seastar/core/coroutine.hh
```diff
@@ -51,10 +53,26 @@ execute_involving_handle_destruction_in_await_suspend(std::invocable<> auto&& fu
...
 
 
+class coroutine_allocators {
+public:
+    static void* operator new(size_t size) {
+        memory::scoped_critical_alloc_section _;
+        return ::malloc(size);
+    }
+    static void operator delete(void* ptr) noexcept {
+        ::free(ptr);
```
kelbon, line 71: why in one branch ::free and memory::free in other?

include/seastar/core/coroutine.hh
avikivity: So it's hidden if the compiler doesn't support sized deallocations. I don't want to confuse the compiler with something it doesn't understand.

include/seastar/core/coroutine.hh
avikivity: In this branch, we don't have the size, so we can't call sized deallocation.

kelbon: but there are no sized deallocation on system level, and both branches must do same semantic, i dont see how they will stay same semantically if sized free will do smth different from free(ptr)

avikivity: They both free an object. Sized free uses the size to improve performance.

include/seastar/core/coroutine.hh
kelbon: There are no mentions in C++ standard, that this operator delete supported for coroutines

https://eel.is/c++draft/task.promise

standard do not makes ifdef, it just uses normal operator delete

And you just commited what copilot writes without understanding (copilot just bad)

Standard says ptr + size has priority over just delete. I dont think compiler which supports coroutines will not support sized delete

https://eel.is/c++draft/dcl.fct.def.coroutine#13

avikivity: If you cannot be civil, don't review.

kelbon: i just know where copilot stole this code

xemul:
> Helps #2899

How? The tests you mentioned did fail (in CI) in my memory

include/seastar/core/sized-free.hh
```diff
@@ -0,0 +1,33 @@
...
+
+#include <cstddef>
+
+namespace seastar::memory {
+
+/// Deallocate memory with known size.
+///
+/// Similar to C++ sized deallocation, but for the C ::malloc() family instead
+/// of `operator new`. The size must be the same as the size used when allocating.
+void free(void* ptr, size_t size);
```
xemul: Why does it need to become part of official API? Next patch can just call memory::do_free()

avikivity: In fact there is free_sized and free_aligned_sized in C23, I'll rename.

avikivity:
v5:
 - drop first patch
 - rename seastar::memory::free to C23 ::free_sized
 - add C23 free_aligned_sized
 - add free_sized, free_aligned_sized  unit test (mainly a compile test)
 - rebase

avikivity:
> > Helps #2899
> 
> How? The tests you mentioned did fail (in CI) in my memory

One of the tests induces an allocation failure, but we can't tolerate coroutine allocation failures, just like we can't tolerate continuation allocation failures (8040cabaee62cf024ef8b9923d3f90661c0884ef).

avikivity:
v6:
 - adjust build guards to support musl libc

avikivity:
v7: refine macro guard not to use function-like macros, which give the preprocessor a headache when not defined

avikivity:
@xemul please re-review

tests/unit/alloc_test.cc
```diff
@@ -682,3 +683,24 @@ SEASTAR_TEST_CASE(test_large_allocation_warning_off_by_one) {
...
+extern "C" {
+
+void free_sized(void* ptr, size_t size);
+void free_aligned_sized(void* ptr, size_t alignment, size_t size);
+
+}
+
+#endif
+
+SEASTAR_TEST_CASE(c23_free_sized) {
```
xemul, line 698: There's `test_sized_delete()` already in this test. Is the new one still needed?

src/core/memory.cc
```diff
@@ -2289,6 +2289,20 @@ extern "C"
 #endif
 void __libc_free(void* obj) noexcept;
 
+extern "C"
```
xemul, line 2292: Are these implementations needed if `!__cpp_sized_deallocations`?

tests/unit/alloc_test.cc
avikivity: That's for C++ sized delete, this is C sized delete.

src/core/memory.cc
avikivity: They're independent, although we practically always have __cpp_sized_deallocations now (C++14 feature).

PR #2901, opened 2025-08-04
Add iotune --get-best-iops-with-buffer-sizes option

robertbindar:
This patch adds the ` --get-best-iops-with-buffer-sizes` option which
instructs iotune to run the read and write iops experiments
for each buffer size passed as argument and choose the best numbers
obtained.

This option cannot be combined with `--random-[read|write]-io-buffer-size`
options.

Usage example:
```bash
iotune --get-best-iops-with-buffer-sizes={512,1024,4096}
```

This is needed in #25211 where `scylla_io_setup`  needs to find
the combination of buffer sizes from `[512, 1k, 2k, 4k]`
that renders the best iops numbers.
This patch will help this effort by bringing the script
execution time from 5x to 1.6x, where x is the typical 120s iotune
run time (throughput experiments takes 0.8x, iops experiments take 0.8x
given `scylla_io_setup` needs to test 4 buffer sizes)

This is related to enabling support for i7i family.

### DIFF
--- a/apps/iotune/iotune.cc
+++ b/apps/iotune/iotune.cc
@@ -37,6 +37,7 @@
 #include <wordexp.h>
 #include <yaml-cpp/yaml.h>
 #include <fmt/printf.h>
+#include <fmt/ranges.h>
 #include <seastar/core/seastar.hh>
 #include <seastar/core/file.hh>
 #include <seastar/core/thread.hh>
@@ -743,6 +744,7 @@ int main(int ac, char** av) {
         ("random-write-io-buffer-size", bpo::value<unsigned>()->default_value(0), "force buffer size for random write")
         ("random-read-io-buffer-size", bpo::value<unsigned>()->default_value(0), "force buffer size for random read")
         ("force-io-depth", bpo::value<unsigned>()->default_value(0), "force io depth to a certain size (overriding auto detection logic)")
+        ("get-best-iops-with-buffer-sizes", bpo::value<std::vector<uint64_t>>()->default_value(std::vector<uint64_t>{}, ""), "run IOPS tests with buffer sizes passed as arg and choose the best combination")
     ;
 
     return app.run(ac, av, [&] {
@@ -756,6 +758,13 @@ int main(int ac, char** av) {
             auto random_write_io_buffer_size = configuration["random-write-io-buffer-size"].as<unsigned>();
             auto random_read_io_buffer_size = configuration["random-read-io-buffer-size"].as<unsigned>();
             auto force_io_depth = configuration["force-io-depth"].as<unsigned>();
+            auto buffer_sizes_for_best_iops = configuration["get-best-iops-with-buffer-sizes"].as<std::vector<uint64_t>>();
+            if (!buffer_sizes_for_best_iops.empty() && (random_read_io_buffer_size || random_write_io_buffer_size)) {
+                iotune_logger.error("--get-best-iops-with-buffer-sizes is set, but random-read-io-buffer-size or random-write-io-buffer-size are also set. "
+                                    "The options are mutually exclusive, please unset one of them.");
+                return 1;
+            }
+
 
             bool read_saturation, write_saturation;
             if (saturation == "") {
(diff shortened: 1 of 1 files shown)


xemul:
It's very hard to review this patch, because it mixes functional change with re-indentation. Can you please split it into two -- first, add the option and necessary if-s, but _without_ re-indenting the relevant code; second, restore indentation back.

Also, while at it, I've a question:

> ... instructs iotune to only run the read and write iops experiments on a dataset generate with the sequential write workload  ...

Does it mean that with the new option the sequential write workload is still run? Or does it re-use the files left from previous iotune invocation? In "re-uses", how to instruct "previous invocation" not to remove it?

robertbindar:
> It's very hard to review this patch, because it mixes functional change with re-indentation. Can you please split it into two -- first, add the option and necessary if-s, but _without_ re-indenting the relevant code; second, restore indentation back.
> 
sometimes git fails me, I'll split it in 2 patches.

> Also, while at it, I've a question:
> 
> > ... instructs iotune to only run the read and write iops experiments on a dataset generate with the sequential write workload  ...
> 
> Does it mean that with the new option the sequential write workload is still run? Or does it re-use the files left from previous iotune invocation? In "re-uses", how to instruct "previous invocation" not to remove it?
The patch forces iotune to run:

if run_iops_only is true:
```
seqwrite_bandwidth (0.1*duration compared to 0.7*duration in the normal operation) 
write_iops
read_iops
```
else (normal operation)
```
seqwrite_bandwidth
seqread_bandwidth
write_iops
read_iops
```
there is no state shared in-between runs.

robertbindar:
> > It's very hard to review this patch, because it mixes functional change with re-indentation. Can you please split it into two -- first, add the option and necessary if-s, but _without_ re-indenting the relevant code; second, restore indentation back.
> 
> sometimes git fails me, I'll split it in 2 patches.
> 
> > Also, while at it, I've a question:
> > > ... instructs iotune to only run the read and write iops experiments on a dataset generate with the sequential write workload  ...
> > 
> > 
> > Does it mean that with the new option the sequential write workload is still run? Or does it re-use the files left from previous iotune invocation? In "re-uses", how to instruct "previous invocation" not to remove it?
> > The patch forces iotune to run:
> 
> if run_iops_only is true:
> 
> ```
> seqwrite_bandwidth (0.1*duration compared to 0.7*duration in the normal operation) 
> write_iops
> read_iops
> ```
> 
> else (normal operation)
> 
> ```
> seqwrite_bandwidth
> seqread_bandwidth
> write_iops
> read_iops
> ```
> 
> there is no state shared in-between runs.

@xemul patch to be reviewed is https://github.com/scylladb/seastar/pull/2901/commits/2da877dd0d0fab150f1b4e108c265b3a62245d48
The other one just indents the content of the `if (!run_iops_only)` block and git weirdly thinks I'm doing lots of code shuffling

apps/iotune/iotune.cc
```diff
@@ -903,14 +923,8 @@ int main(int ac, char** av) {
...
 
-                struct disk_descriptor desc;
-                desc.mountpoint = mountpoint;
                 desc.read_iops = read_iops.iops;
-                desc.read_bw = read_bw.bytes_per_sec;
-                desc.read_sat_len = read_sat;
                 desc.write_iops = write_iops.iops;
-                desc.write_bw = write_bw.bytes_per_sec;
-                desc.write_sat_len = write_sat;
                 disk_descriptors.push_back(std::move(desc));
```
xemul, line 936: In case run_iops_only is true, the `desc` thing is empty, isn't it? In that case, what will be there in the io-properties.yaml file after this push-back?

apps/iotune/iotune.cc
```diff
@@ -864,44 +864,44 @@ int main(int ac, char** av) {
                 size_t sequential_buffer_size = 1 << 20;
 
                 if (!run_iops_only) {
-                fmt::print("Measuring sequential write bandwidth: ");
```
xemul: Please, _title_ the patch as "Restore indentation" of "Fix indentation after previous patch"

apps/iotune/iotune.cc
robertbindar: `desc` is filled with 0s, so the numbers except iops will be 0 in case `run_iops_only` is true. We can maybe warn that bw numbers are not to be used or something like that, but other than that, not sure what we can do better. The `io-properties.yaml` file is not meant to have any meaning when this option is passed. Another option would be to not dump the file at all if the option is true?

apps/iotune/iotune.cc
```diff
@@ -743,6 +743,7 @@ int main(int ac, char** av) {
         ("random-write-io-buffer-size", bpo::value<unsigned>()->default_value(0), "force buffer size for random write")
         ("random-read-io-buffer-size", bpo::value<unsigned>()->default_value(0), "force buffer size for random read")
         ("force-io-depth", bpo::value<unsigned>()->default_value(0), "force io depth to a certain size (overriding auto detection logic)")
+        ("get-best-iops-with-buffer-sizes", bpo::value<std::vector<uint64_t>>()->default_value(std::vector<uint64_t>{}, ""), "run IOPS tests with buffer sizes passed as arg and choose the best combination")
```
xemul, line 747: It looks like patch comment needs to be updated, it still says that `--run-iops-only` is added.
When changing, add usage example. I cannot immediately remember/guess what's the correct syntax for vector option value :(

apps/iotune/iotune.cc
```diff
@@ -891,26 +899,43 @@ int main(int ac, char** av) {
...
                 desc.read_sat_len = read_sat;
-                desc.write_iops = write_iops.iops;
                 desc.write_bw = write_bw.bytes_per_sec;
                 desc.write_sat_len = write_sat;
+
+                if (buffer_sizes_for_best_iops.empty()) {
+                    buffer_sizes_for_best_iops = {test_directory.minimum_io_size()};
+                } else {
+                    fmt::print("Evaluating various buffer sizes for choosing the best IOPS: [ ");
+                    for (uint64_t buffer_size : buffer_sizes_for_best_iops) {
```
xemul: nit: If including <fmt/ranges.h>, will one-line fmt::print of buffer_sizes_for_bease_iops vector just work?

xemul:
LGTM, but patch comment needs to be fixed (also a small nit about fmt::print usage)

apps/iotune/iotune.cc
robertbindar: took me a while to figure it out, I added it in the patch message.

apps/iotune/iotune.cc
robertbindar: nice! thanks

xemul (approved):

PR #2902, opened 2025-08-04
.github: update test workflow to Fedora 42

avikivity:
Ubuntu 24.04 has too old compilers (clang 18 and gcc 13); the current ones are clang 20 and gcc 15.

Update the workflow to run on Fedora 42 via a container so we access the most recent compilers.


avikivity:
Dependency: https://github.com/scylladb/seastar/pull/2903

avikivity:
Rebased past 4ce3e66f0b42bd4fc7a54db367ca79918da69385

avikivity:
v3: rebased past fixes (719d7880eca70d7be3e25c10f41a01656ea5278c, b1a7ed56183a6f414bd8f654394b31f89245bda3)

avikivity:
v4: rebased past fix (65cbae63f4c4426bf56f5c8ec2d87445fc25889e)

.github/workflows/test.yaml
```diff
@@ -38,32 +38,34 @@ jobs:
   test:
     timeout-minutes: 40
     runs-on: ubuntu-24.04
+    container: fedora:43
```
avikivity, line 41: Hmm, we should forward ~/.ccache to the container

avikivity: Or maybe not, if the cache action runs inside the container.

avikivity: ccache integration appears to be working:

```
  Cacheable calls:   409 / 409 (100.0%)
    Hits:            408 / 409 (99.76%)
      Direct:        408 / 408 (100.0%)
      Preprocessed:    0 / 408 ( 0.00%)
    Misses:            1 / 409 ( 0.24%)
  Local storage:
    Cache size (GB): 0.0 / 0.5 ( 6.58%)
    Hits:            408 / 409 (99.76%)
    Misses:            1 / 409 ( 0.24%)
```

One miss from 65cbae63f4c4426bf56f5c8ec2d87445fc25889e, the rest hits.

PR #2903, opened 2025-08-04
Update dpdk submodule

avikivity:
* dpdk cafaa3cf45...6912c864c5 (2):
  > common/ionic: create common code library
  > net/gve/base: fix build with Fedora Rawhide

Cherry-pick fixes for building on more recent toolchains. Without these, we see clashes in typedef for u16 etc.


avikivity:
@xemul please review, this is part of a long dependency chain

PR #2905, opened 2025-08-05
Remove capacity configuration for fair_queue tests

xemul:
The test helper class accepts `int capacity` parameter, but it's now unused (#2820 obsoleted it). After this parameter is removed, it can be noticed that some test cases become equivalent -- they test how fair-queue balances the number of dispatched requests for classes with different parameters, but do it over short and long runtime. This PR also simplifies those test cases.


xemul:

PR #2906, opened 2025-08-05
rpc.cc: reindent

avikivity:
Whitespace changes only.


avikivity:
v2:
  - fixed problems pointed out by copilot
  - resolved never to trust machine reindent

PR #2907, opened 2025-08-05
test: Rework test_fair_queue_random_run

xemul:
This test in its current implementation is pretty weird. All it does is runs two classes with equal reqests and checks that the result is in 1:1 proportion. The only "randomness" here is the test duration, but it's also pretty pointless, as there are two other test cases that validate equal classes in short- and long-runs.

This patch reworks the test to check something, that's not tested by other test cases in this file. Namely

- three classes
- random shares 1 ... 5
- random request weights 1 ... 5

To check the resuts the existing validation function is not suitable, so a similar, yet more powerful one is added.


xemul:

PR #2908, opened 2025-08-05
test: Make fair_queue_test validation code use BOOST_CHECK_...-s

xemul:
The validation code uses BOOST_REQUIRE(condition) which is OK, but in case of failures prints too little info about the failure.


xemul:

PR #2913, opened 2025-08-06
rpc: server::process: coroutinize

avikivity:
In non-saturating rpc workloads, read_request_frame_compressed() will yield for at least a fraction of packets. This will cause the trailing .then() to allocate memory for the continuation.

Optimize this allocation away by converting server::process() to a coroutine. The coroutine frame is reused for the entire loop and thus its allocation is amortized.

The code for calling the actual rpc handler is duplicated for the cases where we have per-connection scheduling groups in order not to suffer the overhead of an additional coroutine allocation when we have a per-connection scheduling group. I intend to deprecate and remove per-handler scheduling group.

In the cleanup phase, we expect that deregister_this_stream() and abort_all_streams() don't throw (they appear not to) so they don't need extra protection.

### DIFF
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -6,6 +6,7 @@
 #include <seastar/core/future-util.hh>
 #include <seastar/core/metrics.hh>
 #include <seastar/coroutine/as_future.hh>
+#include <seastar/coroutine/switch_to.hh>
 #include <seastar/util/assert.hh>
 #include <boost/range/adaptor/map.hpp>
 #include <boost/range/adaptor/transformed.hpp>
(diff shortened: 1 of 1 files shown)


src/rpc/rpc.cc
```diff
@@ -1205,70 +1206,86 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
...
+        {
             f.ignore_ready_future();
             get_server()._conns.erase(get_connection_id());
             if (is_stream()) {
-                return deregister_this_stream();
+                co_await deregister_this_stream();
             } else {
-                return abort_all_streams();
+                co_await abort_all_streams();
+            }
```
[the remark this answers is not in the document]
gleb-cloudius: Can we teach copilot our coding style?

src/rpc/rpc.cc
```diff
@@ -1205,70 +1206,86 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
...
 future<> server::connection::process() {
-    return negotiate_protocol().then([this] () mutable {
+    // hold onto connection pointer until the while loop exits
+    auto conn_ptr = shared_from_this();
+    std::exception_ptr ep;
+    try {
+        co_await negotiate_protocol();
         auto sg = _isolation_config ? _isolation_config->sched_group : current_scheduling_group();
-        return with_scheduling_group(sg, [this] {
+        co_await coroutine::switch_to(sg);
```
gleb-cloudius, line 1215: Currently the cleanup is done in original sg.

src/rpc/rpc.cc
```diff
@@ -1205,70 +1206,86 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
...
+        {
```
gleb-cloudius, line 1216: Why the scope?

src/rpc/rpc.cc
```diff
@@ -1205,70 +1206,86 @@ future<> server::connection::send_unknown_verb_reply(std::optional<rpc_clock_typ
...
+            }
+        }
+    } catch (...) {
+        {
+            ep = std::current_exception();
             log_exception(*this, log_level::error,
                 format("server{} connection dropped", is_stream() ? " stream" : "").c_str(), ep);
         }
+    }
+    {
```
gleb-cloudius, line 1264: Why the scope?

src/rpc/rpc.cc
avikivity: Right. I don't think it's material.

src/rpc/rpc.cc
avikivity: To avoid broken indentation.

src/rpc/rpc.cc
avikivity: To avoid broken indentation. I'll remove and reindent in a follow-up.

src/rpc/rpc.cc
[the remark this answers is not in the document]
avikivity: Yes, there's .github/copilot-instructions.md which can give it instructions.

src/rpc/rpc.cc
gleb-cloudius: OK. I expected the follow-up in the same PR like it was done for previous co-routisation.

avikivity:
> Look fine except of missing re-indent.

I'll post a follow up.

PR #2919, opened 2025-08-07
dns: Use TCP connection data_sink directly

xemul:
In order to send requests, the dns fiber calls
connected_socket::output() to get the output_stream for the socket. However, the stream itself is not really needed -- the buffer size is set to zero and the only sent message is the zero-copy packet.

For that use-case the detached data_sink fits much better.

This will allow to deprecate and eventually remove the net::packet from the output_stream API.

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -208,7 +208,7 @@ private:
         ;
         connected_socket socket;
         std::optional<input_stream<char>> in;
-        std::optional<output_stream<char>> out;
+        std::optional<data_sink> out;
         temporary_buffer<char> indata;
     };
     struct udp_entry {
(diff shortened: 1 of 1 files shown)


src/net/dns.cc
```diff
@@ -1111,9 +1111,9 @@ dns_resolver::impl::do_sendv(ares_socket_t fd, const iovec * vec, int len) {
             switch (e.typ) {
             case type::tcp:
                 if (!e.tcp.out) {
-                    e.tcp.out = e.tcp.socket.output(0);
+                    e.tcp.out = e.tcp.socket.output(0).detach();
                 }
-                f = e.tcp.out->write(std::move(p));
+                f = e.tcp.out->put(std::move(p));
```
avikivity, line 1116: I see there is no flushing, either before or after. Does output_stream::write(packet) imply a flush? What about data_sink::put(packet)?

Seems wrong to have different overloads of the same function imply or not imply a flush.

xemul: Since the stream is created with buffer size of 0, calling write(net::packet) into it flushes immediately. And this flush is output_stream::flush(), not data_sink::flush(), it purely does _fd.put(), so this change should be idempotent.

Also, sockets don't implement their data_sink_impl::flush(), I guess the data_sink::flush() was intended to work for file sinks only.

avikivity: Ok, maybe this works as long as we don't have a file-based DNS server.

PR #2920, opened 2025-08-07
packet: Mark constructor from temporary_buffer explicit

xemul:
Otherwise attempt to put temporary_buffer into output_stream::write() implicitly converts the buffer into packet and calls the write(packet) overload. Effectively, the write(temporary_buffer) is unused.

Marking the aforementioned constuctor explicit makes complier pick the write(temporary_buffer) overload. Althoug it does the very same buffer to packet conversion, pointing the compiler at it will allow to deprecate the write(packet) one in the future.

### DIFF
--- a/include/seastar/net/packet.hh
+++ b/include/seastar/net/packet.hh
@@ -225,7 +225,7 @@ public:
     // append temporary_buffer (zero-copy)
     packet(packet&& x, temporary_buffer<char> buf);
     // create from temporary_buffer (zero-copy)
-    packet(temporary_buffer<char> buf);
+    explicit packet(temporary_buffer<char> buf);
     // append deleter
     packet(packet&& x, deleter d);
 


include/seastar/net/packet.hh
```diff
@@ -225,7 +225,7 @@ public:
     // append temporary_buffer (zero-copy)
     packet(packet&& x, temporary_buffer<char> buf);
     // create from temporary_buffer (zero-copy)
-    packet(temporary_buffer<char> buf);
+    explicit packet(temporary_buffer<char> buf);
     // append deleter
```
tgrabiec, line 229: > 
> Otherwise attempt to put temporary_buffer into output_stream::write()
> implicitly converts the buffer into packet and calls the write(packet)
> overload. Effectively, the write(temporary_buffer) is unused.

I don't understand why. The compiler is supposed to prefer the exact match overload over implicit ones.

xemul: Yes, you're right. I somehow mis-read the compiler error, it complained about conversion in the write(temporary_buffer) _implementation_, which explicitly converts buffer to packet and calls the deprecated write(packet).

PR #2921, opened 2025-08-07
json: API level 8: use noncopyable_function in json_return_type

bhalevy:
Standardize the http request/reply body writer types
and the conversion from json_return_type::body_writer_type
to the http request/reply equivalent.

Hold a http::body_writer_type in json_return_type, making it non-copyable by itself.

The motivation for doing so is supporting async coroutine generators for
generating a (buffered) stream of json elements for writing a json array body,
onto the output stream, without having to buffer the whole array, that can be arbitrarily large.


include/seastar/http/types.hh
```diff
@@ -0,0 +1,39 @@
...
+
+#pragma once
+
+#include <seastar/core/future.hh>
+#include <seastar/core/iostream.hh>
+#include <seastar/util/noncopyable_function.hh>
+#include <seastar/util/modules.hh>
+
+namespace seastar {
+namespace http {
```
avikivity: You can write: `namespace seastar::http {`.

tests/unit/httpd_test.cc
```diff
@@ -818,13 +818,15 @@ SEASTAR_TEST_CASE(dont_abort) {
...
 public:
-    json_test_handler(std::function<future<>(output_stream<char> &&)>&& write_func) : _write_func(write_func) {
+    json_test_handler(http::body_writer_type&& write_func) : _write_func(std::move(write_func)) {
     }
     future<std::unique_ptr<http::reply>> handle(const sstring& path,
             std::unique_ptr<http::request> req, std::unique_ptr<http::reply> rep) override {
-        rep->write_body("json", _write_func);
+        rep->write_body("json", [this] (output_stream<char>&& out) {
+            return _write_func(std::move(out));
+        });
```
avikivity: You can pass `std::ref(_write_func)`.

doc/compatibility.md
```diff
@@ -74,6 +74,7 @@ versions of the API. For example.
    - Seastar_API_LEVEL=6 makes futures non-variadic
    - Seastar_API_LEVEL=7 unifies CPU scheduling groups and IO priority classes
      "while at it" file_impl API is forced to accept io_intent argument
+   - Seastar_API_LEVEL=8 changes json_return_type to hold a noncopyable function
```
avikivity, line 77: (and be a move-only type)

avikivity:
Seems much simpler than the previous version.

tests/unit/httpd_test.cc
bhalevy:
```
tests/unit/httpd_test.cc:827:14: error: no matching member function for call to 'write_body'
  827 |         rep->write_body("json", std::ref(_write_func));
      |         ~~~~~^~~~~~~~~~
/home/bhalevy/dev/seastar/include/seastar/http/reply.hh:248:10: note: candidate function not viable: no known conversion from 'reference_wrapper<noncopyable_function<future<void> (output_stream<char> &&)>>' to 'http::body_writer_type &&' (aka 'noncopyable_function<future<> (output_stream<char> &&)> &&') for 2nd argument
  248 |     void write_body(const sstring& content_type, http::body_writer_type&& body_writer);
      |          ^                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

include/seastar/http/types.hh
bhalevy: Fixed in [51178c5](https://github.com/scylladb/seastar/commit/51178c5f6ca460841d9df65665f32c4c83f315f3)

doc/compatibility.md
bhalevy: Fixed in [51178c5](https://github.com/scylladb/seastar/commit/51178c5f6ca460841d9df65665f32c4c83f315f3)

bhalevy:
In [51178c5](https://github.com/scylladb/seastar/commit/51178c5f6ca460841d9df65665f32c4c83f315f3):
- use namespace seastar::http in types.hh insead of nested namespaces
- Improved AP Level 8 documentation comment as per https://github.com/scylladb/seastar/pull/2921#discussion_r2261007999

bhalevy:
@avikivity please re-review and consider merging

tests/unit/httpd_test.cc
avikivity: Maybe it's a bug in noncopyable function. This example works: https://godbolt.org/z/8aWjM779z.

avikivity: It looks like a signature mismatch.

```
/home/avi/seastar/include/seastar/http/reply.hh:257:19: note: the expression ‘is_invocable_r_v<seastar::future<void>, W, seastar::output_stream<char>&> [with W = std::reference_wrapper<seastar::noncopyable_function<seastar::future<void>(seastar::output_stream<char>&&)> >]’ evaluated to ‘false’
```

The destination function accepts the output_stream by lvalue, while the input function accepts it by rvalue.

Do we have two signatures? Maybe we should take the opportunity to unify them.

bhalevy: > Maybe it's a bug in noncopyable function. This example works: [godbolt.org/z/8aWjM779z](https://godbolt.org/z/8aWjM779z).

Possibly.  I suggest we do this as a follow up as it requires more research

bhalevy: > It looks like a signature mismatch.
> 
> ```
> /home/avi/seastar/include/seastar/http/reply.hh:257:19: note: the expression ‘is_invocable_r_v<seastar::future<void>, W, seastar::output_stream<char>&> [with W = std::reference_wrapper<seastar::noncopyable_function<seastar::future<void>(seastar::output_stream<char>&&)> >]’ evaluated to ‘false’
> ```
> 
> The destination function accepts the output_stream by lvalue, while the input function accepts it by rvalue.
> 
> Do we have two signatures? Maybe we should take the opportunity to unify them.

We do have 2 signatures for historical reasons, one consumes the output_stream and is responsible for closing it, and the other get the output_stream by reference and it calls `util::write_to_stream_and_close` which conforms to the `body_writer_type`.

That said, it's unrelated to supporting std::ref(noncopyable_function).

@avikivity, the following change adds support for that, is it acceptable?
```c++
diff --git a/include/seastar/util/noncopyable_function.hh b/include/seastar/util/noncopyable_function.hh
index 30efc14d..7ccfb131 100644
--- a/include/seastar/util/noncopyable_function.hh
+++ b/include/seastar/util/noncopyable_function.hh
@@ -57,6 +57,10 @@ class noncopyable_function_base {
         new (&to->_storage.indirect) void_ptr(from->_storage.indirect);
     }
 
+    static void ref_move(noncopyable_function_base* from, noncopyable_function_base* to) noexcept {
+        to->_storage.indirect = std::exchange(from->_storage.indirect, nullptr);
+    }
+
     template <size_t N>
     static void trivial_direct_move(noncopyable_function_base* from, noncopyable_function_base* to) {
         // We use bytewise copy here since we lost the type. This means that
@@ -166,6 +170,20 @@ class noncopyable_function<Ret (Args...) noexcept(Noexcept)> : private internal:
         static constexpr vtable make_vtable() { return { call, indirect_move, destroy }; }
         static const vtable s_vtable;
     };
+    struct ref_vtable_for {
+        static noncopyable_function* access(noncopyable_function* func) { return reinterpret_cast<noncopyable_function*>(func->_storage.indirect); }
+        static const noncopyable_function* access(const noncopyable_function* func) { return reinterpret_cast<const noncopyable_function*>(func->_storage.indirect); }
+        static noncopyable_function* access(noncopyable_function_base* func) { return access(static_cast<noncopyable_function*>(func)); }
+        static Ret call(const noncopyable_function* func, Args... args) noexcept(Noexcept) {
+            auto* other = access(const_cast<noncopyable_function*>(func));
+            return other->_vtable->call(other, std::forward<Args>(args)...);
+        }
+        static void initialize(noncopyable_function& from, noncopyable_function* to) {
+            to->_storage.indirect = &from;
+        }
+        static constexpr vtable make_vtable() { return { call, ref_move, empty_destroy }; }
+        static const vtable s_vtable;
+    };
     template <typename Func, bool Direct = true>
     struct select_vtable_for : direct_vtable_for<Func> {};
     template <typename Func>
@@ -186,6 +204,10 @@ class noncopyable_function<Ret (Args...) noexcept(Noexcept)> : private internal:
         vtable_for<Func>::initialize(std::move(func), this);
         _vtable = &vtable_for<Func>::s_vtable;
     }
+    noncopyable_function(std::reference_wrapper<noncopyable_function> func) {
+        ref_vtable_for::initialize(func.get(), this);
+        _vtable = &ref_vtable_for::s_vtable;
+    }
     template <typename Object, typename... AllButFirstArg>
     noncopyable_function(Ret (Object::*member)(AllButFirstArg...) noexcept(Noexcept)) : noncopyable_function(std::mem_fn(member)) {}
     template <typename Object, typename... AllButFirstArg>
@@ -232,5 +254,8 @@ template <typename Func>
 const typename noncopyable_function<Ret (Args...) noexcept(Noexcept)>::vtable noncopyable_function<Ret (Args...) noexcept(Noexcept)>::indirect_vtable_for<Func>::s_vtable
         = noncopyable_function<Ret (Args...) noexcept(Noexcept)>::indirect_vtable_for<Func>::make_vtable();
 
-}
+template <typename Ret, typename... Args, bool Noexcept>
+const typename noncopyable_function<Ret (Args...) noexcept(Noexcept)>::vtable noncopyable_function<Ret (Args...) noexcept(Noexcept)>::ref_vtable_for::s_vtable
+        = noncopyable_function<Ret (Args...) noexcept(Noexcept)>::ref_vtable_for::make_vtable();
 
+}
```

bhalevy:
In [a06a3d4](https://github.com/scylladb/seastar/commit/a06a3d4c2aafad9672ea1479cd7bbf99cdeaa14e):
- util: noncopyable_function: add reference_wrapper constructor
  - allowing `rep->write_body("json", std::ref(_write_func));` in httpd_test `json_test_handler`

tests/unit/httpd_test.cc
avikivity: Why is any change needed? Shouldn't a reference_wrapper be recognized as a compatible function?

The implementation is nice, but I don't see why it's needed. In the original case, it was because the signature did not match, but here you're matching against the same signature.

I want to be careful not to add any capability missing in std::move_only_function, so we can eventually migrate.

bhalevy: > Why is any change needed? Shouldn't a reference_wrapper be recognized as a compatible function?
> 
> The implementation is nice, but I don't see why it's needed. In the original case, it was because the signature did not match, but here you're matching against the same signature.
> 
> I want to be careful not to add any capability missing in std::move_only_function, so we can eventually migrate.

I pushed a new version with the additional constructors of noncopyable_function to support std::ref and std::cref

include/seastar/util/noncopyable_function.hh
```diff
@@ -57,6 +57,10 @@ private:
         new (&to->_storage.indirect) void_ptr(from->_storage.indirect);
     }
 
+    static void ref_move(noncopyable_function_base* from, noncopyable_function_base* to) noexcept {
```
avikivity: > util: noncopyable_function: add reference_wrapper constructor
> So a noncopyable_function can be used by reference
rather than being copied.

Will it really be copied? the reference_wrapper object will be copied, but that's just a pointer.

tests/unit/noncopyable_function_test.cc
```diff
@@ -85,3 +85,15 @@ BOOST_AUTO_TEST_CASE(large_move_tests) {
...
+    noncopyable_function<int (int)> f1 = [] (int x) { return x + 1; };
+    noncopyable_function<int (int)> f2(std::ref(f1));
+    BOOST_REQUIRE_EQUAL(f2(0), 1);
+
+    noncopyable_function<int (int)> f3 = std::ref(f2);
+    BOOST_REQUIRE_EQUAL(f3(1), 2);
+
+    noncopyable_function<int (int)> f4 = std::cref(f3);
+    BOOST_REQUIRE_EQUAL(f4(2), 3);
+}
```
avikivity: Does this test fail without the patch?

bhalevy: It fails to compile without the change in noncopyable_function:
```
/home/bhalevy/dev/seastar/tests/unit/noncopyable_function_test.cc:93:37: error: copying variable of type 'seastar::noncopyable_function<int (int)>' invokes deleted constructor
   93 |     noncopyable_function<int (int)> f3 = std::ref(f2);
      |                                     ^    ~~~~~~~~~~~~
/home/bhalevy/dev/seastar/include/seastar/util/noncopyable_function.hh:198:5: note: 'noncopyable_function' has been explicitly marked deleted here
  198 |     noncopyable_function(const noncopyable_function&) = delete;
      |     ^
/home/bhalevy/dev/seastar/tests/unit/noncopyable_function_test.cc:96:37: error: copying variable of type 'const seastar::noncopyable_function<int (int)>' invokes deleted constructor
   96 |     noncopyable_function<int (int)> f4 = std::cref(f3);
      |                                     ^    ~~~~~~~~~~~~~
/home/bhalevy/dev/seastar/include/seastar/util/noncopyable_function.hh:198:5: note: 'noncopyable_function' has been explicitly marked deleted here
  198 |     noncopyable_function(const noncopyable_function&) = delete;
      |     ^
2 errors generated.
```

include/seastar/util/noncopyable_function.hh
bhalevy: We need a way to construct a noncopyable_function from a reference_wrapper

avikivity: Why isn't the existing constructor sufficient?

avikivity: This compiles: https://godbolt.org/z/hjv6n5zE6

tests/unit/noncopyable_function_test.cc
avikivity: This is due to lack of an assignment operator from an invocable. Without such an assignment operator, std::cref(f3) converts to a reference to f3, which is then attempted to be copied into f4, which fails since noncopyable_function is noncopyable.

bhalevy: > This is due to lack of an assignment operator from an invocable. Without such an assignment operator, std::cref(f3) converts to a reference to f3, which is then attempted to be copied into f4, which fails since noncopyable_function is noncopyable.

Hmm, I tried adding such an assignment operator but still got the same error.
Even with the following, inspired by your example:
```
    using func_t = noncopyable_function<int (int)>;

    auto f1 = [] (long x) { return x + 1; };

    auto f1f = func_t(f1);
    auto f2f = func_t(std::ref(f1f));   // This works
    func_t f3f = std::ref(f1f);         // This doesn't
```

With this:
```diff
diff --git a/include/seastar/util/noncopyable_function.hh b/include/seastar/util/noncopyable_function.hh
index 30efc14d..6a0f8c5c 100644
--- a/include/seastar/util/noncopyable_function.hh
+++ b/include/seastar/util/noncopyable_function.hh
@@ -186,6 +186,13 @@ class noncopyable_function<Ret (Args...) noexcept(Noexcept)> : private internal:
         vtable_for<Func>::initialize(std::move(func), this);
         _vtable = &vtable_for<Func>::s_vtable;
     }
+    template <typename Func>
+    requires std::is_invocable_r_v<Ret, Func, Args...>
+    noncopyable_function& operator=(Func x) {
+        this->~noncopyable_function();
+        new (this) noncopyable_function(std::move(x));
+        return *this;
+    }
     template <typename Object, typename... AllButFirstArg>
     noncopyable_function(Ret (Object::*member)(AllButFirstArg...) noexcept(Noexcept)) : noncopyable_function(std::mem_fn(member)) {}
     template <typename Object, typename... AllButFirstArg>
```

```
/home/bhalevy/dev/seastar/tests/unit/noncopyable_function_test.cc:95:12: error: copying variable of type 'seastar::noncopyable_function<int (int)>' invokes deleted constructor
   95 |     func_t f3f = std::ref(f1f);         // This doesn't
      |            ^     ~~~~~~~~~~~~~
/home/bhalevy/dev/seastar/include/seastar/util/noncopyable_function.hh:205:5: note: 'noncopyable_function' has been explicitly marked deleted here
  205 |     noncopyable_function(const noncopyable_function&) = delete;
      |     ^
```

In any case, `rep->write_body("json", http::body_writer_type(std::ref(_write_func)));` works fine.

bhalevy:
In [8e956c1](https://github.com/scylladb/seastar/commit/8e956c1d32969da7018d74dd021bd0ff32c85be9):
- dropped changes to noncopyable_function
- added explicit construction of `http::body_writer_type(std::ref(_write_func))` in httpd_test.json_test_handler

tests/unit/noncopyable_function_test.cc
avikivity: noncopyable_function deserves an assignment operator, but let's leave it out of this already-nested series.

avikivity (approved):

PR #2922, opened 2025-08-07
reactor: replace boost::barrier with std::barrier<>

avikivity:
This allows dropping the boost.thread dependency, so do that too.

Some dependencies on std::map were lost as a result, so restore them.


avikivity:
v2: adjust seastar.cc to get the module build working

PR #2924, opened 2025-08-09
build: don't recover from sanitizer failures

avikivity:
The sanitizer is able to recover from some errors, but we want to fail early so we can fix those bugs.

Set -fno-sanitize-recover=all to accomplish that.


avikivity:
This doesn't work well with suppressions.

PR #2926, opened 2025-08-10
noncopyable_function: use memcpy instead of bytewise copy loop

avikivity:
I observed [1] that an 8-byte copy loop is compiled as 8 individual byte copies. This is likely due to the compiler not figuring out the source and destination cannot partially overlap (due to [[gnu::may_alias]]), and therefore following the loop to the letter.

Help it out by calling memcpy. It translates to a single word move [2].

[1] https://godbolt.org/z/7sq6cn837 (trivial_direct_move)
[2] https://godbolt.org/z/dEzvhsTKK (trivial_direct_move)


include/seastar/util/noncopyable_function.hh
```diff
@@ -66,10 +67,11 @@ private:
         // that, and complains. Silence it.
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wuninitialized"
-        // Avoid including <algorithm> just for this
-        for (unsigned i = 0; i != N; ++i) {
-            to->_storage.direct[i] = from->_storage.direct[i];
-        }
+        // A bytewise loop doesn't optimize to a wordwise loop since
```
mykaul, line 70: comment on line 63 isn't accurate anymore?

avikivity: memcpy is still a bytewise copy

include/seastar/util/noncopyable_function.hh
```diff
@@ -66,10 +67,11 @@ private:
...
+        // the compiler can't guess if there is overlap or not. Help
+        // it out by using a memcpy. This translates to wordwise
+        // copies.
+        std::memcpy(to->_storage.direct, from->_storage.direct, N);
```
nyh: Looks good to me, but I thought over the years that you tried to avoid std::memcpy and use std::copy_n instead? Or you like both?

nyh (approved):

include/seastar/util/noncopyable_function.hh
avikivity: I prefer std::copy_n(), but here the optimizer is too careful, as I explained in the commit message. std::copy_n() cannot be optimized if the optimizer thinks overlap is possible.

nyh: According to the documentation of copy_n, "Overlap of ranges is formally permitted, but leads to unpredictable ordering of the results." - so the compiler can do anything if it thinks overlaps are possible (and in particular, use the optimized code). But maybe it indeed doesn't optimize - I didn't check.

In any case, I have nothing against std::memcpy - just that I'm still surprised about the "hutzpah" of compiler writers to convert explicit calls to library functions by their own builtin functions.

avikivity: Why? It's a formally specified interface.

There is -mmemcpy if you want memcpy to be called for some reason.

avikivity: copy_n did work: https://godbolt.org/z/6cjj7ch8q

nyh: So, do you want me to merge the patch with memcpy(), or wait for another version that uses copy_n()?

avikivity: I think memcpy is less fragile here, copy_n depends too much on the library internals (which eventually forward the call to memcpy).

xemul: Does `noncopyable_function(noncopyable_function&& x)` need to be patched either not to move in case move-to-self happens? Otherwise  "Overlap of ranges is formally permitted, but leads to unpredictable ordering of the results." seem to happen

avikivity: I think so.

Perhaps better is to copy the source to a temporary, then copy the temporary to the destination. This avoids problems with overlaps, disambiguates accesses for the optimizer, and compiles to the same thing.

avikivity:
v2: avoid the ambiguity in a more direct way, by using a temporary.

include/seastar/util/noncopyable_function.hh
```diff
@@ -66,9 +67,17 @@ private:
...
+        // overlap.
+        std::array<char, N> tmp;
         // Avoid including <algorithm> just for this
-        for (unsigned i = 0; i != N; ++i) {
-            to->_storage.direct[i] = from->_storage.direct[i];
+        for (size_t i = 0; i != N; ++i) {
+            tmp[i] = from->_storage.direct[i];
+        }
+        for (size_t i = 0; i != N; ++i) {
+            to->_storage.direct[i] = tmp[i];
```
nyh, line 80: And then we wonder why Seastar-based code takes hours to compile ;-)

avikivity: I don't see the connection.

nyh: You are forcing the compiler to instantiate a template `array<N>`, compile two loops that copy data byte by byte and increment the counter i from 0 to N twice and copy each byte twice - and then run its fancy optimizer until absolutely nothing is left from all of this except one assembler instruction.

This results in excellent performance, but also terrible compilation speed.

avikivity: There are only a few values of N that we instantiate for, and the code is in non-template noncopyable_function_base. So the actual number of instantiations and optimizations we ask the compiler for is tiny compared to a translation unit's size.

Compile time bloat comes from large continuation chains, where future::then (or similar) compile the lambda call twice, once for the immediate case and once for the deferred case (well it's compiled once but optimized twice). This doubling happens for each nesting level, leading to exponential compile times. It's all user code so no reuse happens.

nyh: You're right, but I'm not sure that continuations is the only problem (as we co-routinize more and more code, are we seeing compilation time improvements? I didn't see it).
My comment wasn't about this specific instance (noncopyable_function) as being the root of all our problems. It was about the general philosophy that modern C++ forces us to use - writing code which we know will need the compiler to go over a thousand instructions and then eliminate most of them in the optimizer. I think this causes a lot of our build slowness - but obviously not all of it and maybe you're right that not even a majority of it.

avikivity: Either the compiler sees and optimizes a lot of code during compile time, or the CPU sees a lot of instructions during runtime. It's not realistic to have the programmer manually inline everything.

PR #2931, opened 2025-08-13
include: Move array_map.hh to util/internal/

xemul:
And to internal namespace either. And adjust the callers.

Callers could use the std::array<> directly, but they need C99-style designated initializers for convenience.


xemul:

PR #2932, opened 2025-08-13
reactor: Equip make_task() and lambda_task with concepts

xemul:
The latter is in fact a wrapper over fn(void) functions, and the former just inherits that. Adding a concept makes this requirement explicit.

### DIFF
--- a/include/seastar/core/make_task.hh
+++ b/include/seastar/core/make_task.hh
@@ -32,6 +32,7 @@ namespace seastar {
 SEASTAR_MODULE_EXPORT_BEGIN
 
 template <typename Func>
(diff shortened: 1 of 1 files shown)


include/seastar/core/make_task.hh
```diff
@@ -32,6 +32,7 @@ namespace seastar {
 SEASTAR_MODULE_EXPORT_BEGIN
 
 template <typename Func>
+requires std::invocable<Func>
 class lambda_task final : public task {
```
avikivity, line 36: Note: can write `template <std::invocable Func> class lambda_task`.

PR #2934, opened 2025-08-15
Fix and tune http::request setup by client

xemul:
Currently, if a request passed into client::make_request() fails validation (missing version and inconsistent body content vs headers) the method throws, while it's a furure<>-returning method and should return exceptional future instead. This PR turns exception into a future.

Other than that, request validation is moved, so that it happens once per make_request(). Nowadays it happens on every retry.

And finally, placing defaults on request is moved from connection::make_request() into request::make(), so that client request making API accepts const request reference (finalizes the #1848 effort)

### DIFF
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -111,7 +111,7 @@ future<connection::reply_ptr> connection::maybe_wait_for_continue(const request&
     });
 }
 
-static void setup_request(request& req) {
+static void validate_request(const request& req) {
     if (req._version.empty()) {
         throw std::runtime_error("HTTP version not set");
     }
@@ -170,7 +170,11 @@ future<connection::reply_ptr> connection::do_make_request(const request& req) {
 }
 
 future<reply> connection::make_request(request req) {
-    setup_request(req);
+    try {
+        validate_request(req);
+    } catch (...) {
+        return current_exception_as_future<reply>();
+    }
     return do_with(std::move(req), [this] (auto& req) {
         return do_make_request(req).then([] (reply_ptr rep) {
             return make_ready_future<reply>(std::move(*rep));
@@ -351,7 +355,11 @@ static bool is_retryable_exception(std::exception_ptr ex) {
 }
 
 future<> client::make_request(request& req, reply_handler& handle, std::optional<reply::status_type> expected, abort_source* as) {
-    setup_request(req);
+    try {
+        validate_request(req);
+    } catch (...) {
+        return current_exception_as_future();
+    }
     return with_connection([this, &req, &handle, as, expected] (connection& con) {
         return do_make_request(con, req, handle, as, expected);
     }, as).handle_exception([this, &req, &handle, as, expected] (std::exception_ptr ex) {
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -111,7 +111,7 @@ future<connection::reply_ptr> connection::maybe_wait_for_continue(const request&
     });
 }
 
-void connection::setup_request(request& req) {
+static void setup_request(request& req) {
     if (req._version.empty()) {
         req._version = "1.1";
     }
@@ -156,8 +156,7 @@ future<connection::reply_ptr> connection::recv_reply() {
     });
 }
 
-future<connection::reply_ptr> connection::do_make_request(request& req) {
-    setup_request(req);
+future<connection::reply_ptr> connection::do_make_request(const request& req) {
     return send_request_head(req).then([this, &req] {
         return maybe_wait_for_continue(req).then([this, &req] (reply_ptr cont) {
             if (cont) {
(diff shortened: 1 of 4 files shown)


kreuzerkrieg (approved):
LGTM

xemul:
@nyh , please review

xemul:
@nyh , please review

xemul:
@nyh , please review

xemul:
@nyh , please review

src/http/client.cc
```diff
@@ -113,7 +113,7 @@ future<connection::reply_ptr> connection::maybe_wait_for_continue(const request&
 
 static void setup_request(request& req) {
     if (req._version.empty()) {
-        req._version = "1.1";
+        throw std::runtime_error("HTTP version not set");
```
nyh, line 116: If I understand correctly, this changes the API - you now absolutely must specify the request's version and can't have it default to 1.1 like it did. I guess it's fine to change the API because it's in the experiemental namespace?

But why do you want version to be explicitly given, since there is a very good default - 1.1 has been the default for almost 20 years now?

src/http/client.cc
```diff
@@ -170,7 +170,11 @@ future<connection::reply_ptr> connection::do_make_request(const request& req) {
 }
 
 future<reply> connection::make_request(request req) {
-    setup_request(req);
+    try {
+        validate_request(req);
+    } catch (...) {
+        return current_exception_as_future<reply>();
```
nyh, line 176: Wow, I completely missed that we switched from the idiom
```
return seastar::make_exception_future(std::current_exception());
```
to
```
return current_exception_as_future<reply>();
```
I'm curious why. We need to update doc/tutorial.md which refers to the old idiom. Sad there is no grant any more forcing us to write such documentation :-(

src/http/client.cc
xemul: > If I understand correctly, this changes the API - you now absolutely must specify the request's version and can't have it default to 1.1 like it did.

It still defaults to 1.1 but in another place -- in the `request::make()` helper that's used by all our code anyway.

>  I guess it's fine to change the API because it's in the experiemental namespace?

Yes, exactly. 

> But why do you want version to be explicitly given, since there is a very good default - 1.1 has been the default for almost 20 years now?

Do I get it correctly that you suggest to always put "1.1" by the client regardless of what's there on the `request::_version`? No strong reason. However, someone [is working on](https://github.com/scylladb/seastar/pull/2886) http/2, presumably (I didn't check that) sending a request with `req._version = "2.0"` and still seeing "1.1" string  appearing on the socket would be quite surprising

xemul:
> Or maybe I misunderstood and it's actually still the default?

Yes, it's still the default, just moved to another place

src/http/client.cc
xemul: e2b81a92bcd07f4a3fc1ef3e3bf492d99e753348:

>     This is the future version of std::current_exception. This moves the
>     std::current_exception call out of line, which also move the
>     std::exception_ptr destructor call offline.

src/http/client.cc
nyh: I may be asking a stupid question because I don't understand the API. I got the impression that the API looks something like:
```
request req;
request.content = "hello";
client.make_request(request)
```
In that case, I thought it made sense that before your patch this request will default to 1.1.
Clearly, if someone creates a request explicitly asking for a different version - that version should be used.

src/http/client.cc
nyh: Thanks.

To be honest, I don't understand the justification in that commit (it's the first time I'm seeing it, maybe I was on vacation in December 2019 ;-)). By "offline", did he mean "out of line"? But in any case, since both functions would be called in an **exception handler***, wouldn't they be out-of-line anyway?

It's sad that that patch did not:
1. Update tutorial.md. @avikivity one day we should decide to spend real resources (i.e., time) on making tutorial.md into a real book and not the 10%-written old-style-continuation thing we have today. For new employees, and perhaps we can even see Seastar really take off externally.
2. Contain any sort of benchmarks or something proving why it's better to use the new idiom. Commit 0bde519638068bc313296e5037b56077632b02b9 did say that half a dozen patches together reguced the code size by 1.4%, but how much of this is due to using the new current_exception_as_future()? I'll open an issue about this.

nyh: I see the doxygen in include/seastar/core/future.hh actually says about current_exception_as_future that it is:

> This is equivalent to make_exception_future(std::current_exception()), but expands to less code.

I still have no idea why and would have been nice to read about somewhere. We don't pay for doxygen comments by the byte ;-)

src/http/client.cc
xemul: The API nowadays is
```
auto req = request::make("GET", "google.com", "/foo/bar");
co_await client.make_request(req);
```

so assigning "_version = 1.1" now happens in `request::make()`. If someone needs another version, then it should use

```patch
 auto req = request::make("GET", "google.com", "/foo/bar");
+req._version = "2.0";
 co_await client.make_request(req);
```

nyh: Ok, though I don't understand why we have a "factory function" `request::make(...)` instead of an ordinary constructor, `request(...)`. But in any case, that is not new to this patch, so I am only asking for my own education, it's not really related to this review.

PR #2935, opened 2025-08-18
include: Keep linux-aio.hh deeper in internal

xemul:
All this file contents is now in internal namespace. On top of that, this patch

- removes the pointless namespace breakage in the middle
- adds comments to terminating curly braces
- moves the file into internal/ sub-directory
- ... and updates the includes from .cc (quite a few)


xemul:

PR #2937, opened 2025-08-21
Expell net::packet from output_stream API stack

xemul:
The existing data_sink_impl API has three put() virtual overloads

* put(net::packet) -- pure virtual one
* put(temporary_buffer) -- implementation creates net::packet out of the buffer
* put(vector<temporary_buffer>) -- similarly, the implementation converts the vector to packet

Also there's fallback_put(net::packet) method for those implementations that don't want to mess with packet and prefer to convert the packet back to buffers.

This API presumably was driven by the output_stream zero-copy buffers extension, that maintains net::packet on the output stream itself that accumulates written buffers and then put()-s them into sink. And the fallback_put() appeared later to facilitate sink implementations.

Maintaining net::packet as zero-copy buffers on output_stream is, in turn, makes the stream work in two (and a half) modes -- the users can either buffers data, then flush, or append zero-copy buffers, then, again, flush. There's also a semi-mixed mode, where zero-copy buffers may come after a bunch of buffered writes. Using mixed mode should still happen with care -- after zero-copy writes and flush(), starting buffered writes can step on assertion if the stream is batch-flushed.

Also the need to implement put(net::packet) overload is pretty harsh requirement, sinks that are not network sockets plug this implementation with abort() and require callers not to perform zero-copy writes into such streams.

This PR eliminates the net::packet from the output_stream+data_sink layer and leaves it on socket sink implementations only. For that both, output_stream and data_sink are changed.

First, the data_sink_impl. The new API (backward incompatible and thus under new API level) has just one put() that accepts std::span<temporary_buffer>. It's pure virtual method, implementations must grab the buffers before returning (even if the returned future is unresolved).

The data_sink() has the put(span) overload as well as put(temporary_buffer) and put(vector<temporary_buffer>) ones, for convenience.

File sink benefit from that change by just dropping the plugged put(packet) overload and that's mostly it. To submit buffers from span it picks them one-by-one, but later it should be tuned to submit iovec request.

Network sinks immediatly convert span of buffers into net::packet, no other changes are made.

HTTP chunked encoding sink converts each buffer from span into a "chunk", no other changes from current implementation is made. When we have mixed-mode (not this PR) it will be able to relax it to zero-copy buffers. The content-length sink zero-copy forwards the span to lower output_stream with the help of ... (see next paragraph)

The output_stream change is -- the _zc_bufs member is changed from net::packet to be std::vector<temporary_buffer> and zero-copy write()-s accepting packet and scattered_message are removed in the new API level (could be deprecated, but it would delay full expelling of packet from streams API). The write of single temporary buffer is preserved. A new write(span<temporary_buffer>) overload is added.

After the PR the existing _buf for buffered writes and _zc_bufs with zero-copied buffers still co-exist on output_stream and the old non-mixed behavior is preserved. However, this opens a way to implement the fully-mixed mode itself  :crossed_fingers:  eventually.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,10 +70,10 @@ option (Seastar_DEPRECATED_OSTREAM_FORMATTERS
   ON)
 
 set (Seastar_API_LEVEL
-  "8"
+  "9"
   CACHE
   STRING
-  "Seastar compatibility API level (7=unified CPU/IO scheduling groups, 8=noncopyable function in json_return_type")
+  "Seastar compatibility API level (7=unified CPU/IO scheduling groups, 8=noncopyable function in json_return_type, 9=new sink API")
 
 set_property (CACHE Seastar_API_LEVEL
   PROPERTY
--- a/configure.py
+++ b/configure.py
@@ -74,7 +74,7 @@ arg_parser.add_argument('--ldflags', action='store', dest='user_ldflags', defaul
                         help='Extra flags for the linker')
 arg_parser.add_argument('--optflags', action='store', dest='user_optflags', default='',
                         help='Extra optimization flags for the release mode')
-arg_parser.add_argument('--api-level', action='store', dest='api_level', default='8',
+arg_parser.add_argument('--api-level', action='store', dest='api_level', default='9',
                         help='Compatibility API level (8=latest)')
 arg_parser.add_argument('--compiler', action='store', dest='cxx', default='g++',
                         help='C++ compiler path')
--- a/doc/compatibility.md
+++ b/doc/compatibility.md
@@ -76,6 +76,8 @@ versions of the API. For example.
      "while at it" file_impl API is forced to accept io_intent argument
    - Seastar_API_LEVEL=8 changes json_return_type to hold a noncopyable function
      and become a move-only type
+   - Seastar_API_LEVEL=9 defines the data_sink_impl::put(span<temporary_buffer>)
+     as the new and only method to be implemented
 
 Applications can use an old API_LEVEL during a transition
 period, fix their code, and move to the new API_LEVEL.
@@ -117,6 +119,7 @@ API Level History
 | 6   |  2020-09  | 2023-03 | future<T> instead of future<T...>            |
 | 7   |  2023-05  | 2024-09 | unified CPU/IO scheduling groups             |
 | 8   |  2025-08  |         | noncopyable function in json_return_type     |
+| 9   |  2025-08  |         | data_sink_impl new API                       |
 
 
 Note: The "mandatory" column indicates when backwards compatibility
--- a/include/seastar/core/internal/api-level.hh
+++ b/include/seastar/core/internal/api-level.hh
@@ -23,7 +23,7 @@
 
 // For IDEs that don't see SEASTAR_API_LEVEL, generate a nice default
 #ifndef SEASTAR_API_LEVEL
-#define SEASTAR_API_LEVEL 8
+#define SEASTAR_API_LEVEL 9
 #endif
 
 #if SEASTAR_API_LEVEL == 8


xemul:
@avikivity , please, share your thoughts about the general idea

avikivity:
> The existing data_sink_impl API has three put() virtual overloads
> 
> * put(net::packet) -- pure virtual one
> * put(temporary_buffer) -- implementation creates net::packet out of the buffer
> * put(vector<temporary_buffer>) -- similarly, the implementation converts the vector to packet
> 
> Also there's fallback_put(net::packet) method for those implementations that don't want to mess with packet and prefer to convert the packet back to buffers.
> 
> This API presumably was driven by the output_stream zero-copy buffers extension, that maintains net::packet on the output stream itself that accumulates written buffers and then put()-s them into sink. And the fallback_put() appeared later to facilitate sink implementations.
> 
> Maintaining net::packet as zero-copy buffers on output_stream is, in turn, makes the stream work in two (and a half) modes -- the users can either buffers data, then flush, or append zero-copy buffers, then, again, flush. There's also a semi-mixed mode, where zero-copy buffers may come after a bunch of buffered writes. Using mixed mode should still happen with care -- after zero-copy writes and flush(), starting buffered writes can step on assertion if the stream is batch-flushed.
> 
> Also the need to implement put(net::packet) overload is pretty harsh requirement, sinks that are not network sockets plug this implementation with abort() and require callers not to perform zero-copy writes into such streams.
> 
> This PR eliminates the net::packet from the output_stream+data_sink layer and leaves it on socket sink implementations only. For that both, output_stream and data_sink are changed.
> 
> First, the data_sink. The new API (backward incompatible and thus under new API level) is two put()-s
> 
> * put(temporary_buffer) -- pure virtual one
> * put(vector<temporary_buffer>) -- the implementation falls-back to putting vector elements with the former overload one-by-one

Should we accept a span here?

I think the implementation will not be able to reuse the vector. If it has a queue of fragments, it won't be a queue of vectors of fragments, but a direct queue of fragments.

An advantages is that put(temporary_buffer) can call into put(span) and we only have one virtual entry point.


> 
> If the sink implementation thinks it can do vector put better, than sequentially it can override one and this is what socket sinks do -- they convert the passed vector of buffers into net::packet and proceed. In fact, this is what already happens today, partially at the output_stream level itself, so this change just moves the net::packet instantiation to src/net/ code.
> 
> File sink benefit from that change by just dropping the plugged put(packet) overload and that's it.
> 
> The loser here is http content length body sink that stops being zero-copy, but that'll be changed in v2 (if it follows).

Why does it stop being zero-copy?

> 
> The output_stream change is -- the _zc_bufs member is changed from net::packet to be std::vector<temporary_buffer> 

Excellent.


> and zero-copy write()-s accepting packet and scattered_message are deprecated in favor of (exiting) write of temporary_buffer itself.
> 
> The deprecated write()-s are implemented to convert packet/scattered_message into a vector of buffers and write the buffers themselves. That's going to generate some CPU overhead converting things back-and-forth plus memory allocation, but these methods are deprecated and callers are support to switch to write(temporary_buffer) which doesn't do it.

span will eliminate the memory allocation for the small cases.

> 
> Right now the existing _buf for buffered writes and _zc_bufs with zero-copied buffers co-exist on output_stream and the old non-mixed behavior is preserved. However, next step is the fully-mixed mode itself 🤞

avikivity:
> @avikivity , please, share your thoughts about the general idea

From reading the cover letter I'm in favor. For @nyh (also should be in the cover letter), net::packet is a low-level API representing packets for device drivers, having things like headroom and offload information, that are irrelevant for the higher levels of the stack. A net::packet sent from a high level will need to be split and merged in a tcp stream to suit the hardware and window limits.

include/seastar/core/iostream.hh
```diff
@@ -112,6 +114,7 @@ public:
     virtual temporary_buffer<char> allocate_buffer(size_t size) {
         return temporary_buffer<char>(size);
     }
+#if SEASTAR_API_LEVEL < 9
```
avikivity: Should prefer the new API in reading order.

include/seastar/core/iostream.hh
```diff
@@ -124,6 +127,19 @@ public:
...
+    virtual future<> put(temporary_buffer<char> buf) = 0;
+    virtual future<> put(std::vector<temporary_buffer<char>> bufs) {
+        if (bufs.size() == 1) {
+            return put(std::move(bufs.front()));
+        }
+        return do_with(std::move(bufs), [this] (std::vector<temporary_buffer<char>>& bufs) {
+            return do_for_each(bufs, [this] (temporary_buffer<char>& buf) {
+                return put(std::move(buf));
+            });
+        });
```
avikivity: If we use a span here, we need to decide if it is the responsibility of the caller to maintain the span, or if the sink implementation must stabilize it.

I believe sinks can stabilize it. Native tcp has an output queue. Posix tcp converts the entire thing into a msghdr. We cannot really use the vector. It only helps if we have to yield before (if we yield we must stabilize it temporarily).

include/seastar/core/iostream.hh
```diff
@@ -189,13 +207,22 @@ public:
...
       try {
         return _dsi->put(std::move(p));
       } catch (...) {
         return current_exception_as_future();
       }
     }
+    future<> deprecated_put(net::packet p) noexcept {
+        return put(std::move(p));
+    }
+#else
```
avikivity: Why not [[deprecated]]? Motivate users to move to level 9.

avikivity: (but internal callers should suppress it)

include/seastar/core/iostream.hh
```diff
@@ -189,13 +207,22 @@ public:
...
+    future<> deprecated_put(net::packet p) noexcept {
+        return put(p.release());
+    }
+#endif
```
avikivity, line 243: Why have deprecated_put in the new API?

include/seastar/websocket/common.hh
```diff
@@ -92,10 +92,16 @@ protected:
...
 
+#if SEASTAR_API_LEVEL < 9
         virtual future<> put(net::packet d) override {
             net::fragment f = d.frag(0);
             return data->push_eventually(temporary_buffer<char>{std::move(f.base), f.size});
         }
+#else
+        virtual future<> put(temporary_buffer<char> buf) override {
+            return data->push_eventually(std::move(buf));
+        }
```
avikivity: Something like push_eventually will need to stabilize a span if it gets one.

avikivity: Perhaps a better approach is to push immediately, but resolve the future only when control flow allows. So at least we know a fast implementation is possible.

src/net/native-stack-impl.hh
```diff
@@ -206,11 +206,28 @@ public:
...
+        return _conn->send(std::move(p));
+    }
+public:
+    virtual future<> put(temporary_buffer<char> buf) override {
+        return put(net::packet(std::move(buf)));
+    }
+
+    virtual future<> put(std::vector<temporary_buffer<char>> bufs) override {
+        return put(net::packet::make(std::move(bufs)));
+    }
```
avikivity: The conversion to packets is premature. tcp should keep a queue of temporary_buffer:s and assemble them into packets at the last moment. I'm not suggesting to change anything now.

avikivity: (the queue now is a vector of vectors because packet is fragmented)

src/net/packet.cc
```diff
@@ -137,6 +137,14 @@ std::ostream& operator<<(std::ostream& os, const packet& p) {
     return os;
 }
 
+packet packet::make(std::vector<temporary_buffer<char>> bufs) {
+    packet p(bufs.size());
+    for (auto&& buf : bufs) {
+        p = packet(std::move(p), std::move(buf));
+    }
```
avikivity: Is there no nicer way?

src/net/posix-stack.cc
```diff
@@ -681,13 +681,23 @@ posix_data_sink_impl::put(temporary_buffer<char> buf) {
...
     bytes_sent[sg_id] += _p.len();
     return _fd.write_all(_p).then([this] { _p.reset(); });
 }
+#else
+future<>
+posix_data_sink_impl::put(std::vector<temporary_buffer<char>> bufs) {
+    _p = net::packet::make(std::move(bufs));
+    auto sg_id = internal::scheduling_group_index(current_scheduling_group());
+    bytes_sent[sg_id] += _p.len();
+    return _fd.write_all(_p).then([this] { _p.reset(); });
```
avikivity: Okay as a transition, but eventually should drop reliance on net::packet.

include/seastar/core/temporary_buffer.hh
```diff
@@ -32,6 +32,7 @@
 #include <seastar/util/eclipse.hh>
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/modules.hh>
+#include <seastar/util/assert.hh>
```
avikivity, line 36: > A (small?) issue with this method -- the vector being split implies
> erasure from front, which copies elements. Just to keep in mind when
> reviewing future patches.

Maybe spannization helps?

include/seastar/core/iostream.hh
```diff
@@ -447,9 +447,10 @@ class output_stream final {
     static_assert(sizeof(CharType) == 1, "must buffer stream of bytes");
     data_sink _fd;
     temporary_buffer<CharType> _buf;
-    net::packet _zc_bufs = net::packet::make_null_packet(); //zero copy buffers
+    std::vector<temporary_buffer<CharType>> _zc_bufs; // zero copy buffers
```
avikivity, line 468: Maybe circular_buffer? If it has hooks into processing spans.

I believe a vector is fine too for reasonable pending data sizes.

btw, do we have flow control over _zc_bufs, or can it grow indefinitely?

include/seastar/core/iostream.hh
```diff
@@ -495,8 +495,10 @@ public:
     future<> write(const std::basic_string<char_type>& s) noexcept;
 
     /// Appends the packet as zero-copy buffer
+    [[deprecated("Use write(temporary_buffer) overload")]]
     future<> write(net::packet p) noexcept;
```
avikivity, line 517: Isn't the vector/span variant more similar?

include/seastar/core/iostream.hh
```diff
@@ -495,8 +495,10 @@ public:
...
     /// Appends the scattered message as zero-copy buffer
+    [[deprecated("Use write(temporary_buffer) overload")]]
     future<> write(scattered_message<char_type> msg) noexcept;
```
avikivity, line 519: Even more.

xemul:
> > First, the data_sink. The new API (backward incompatible and thus under new API level) is two put()-s
> > 
> > * put(temporary_buffer) -- pure virtual one
> > * put(vector<temporary_buffer>) -- the implementation falls-back to putting vector elements with the former overload one-by-one
> 
> Should we accept a span here?

What do you call "a span"? Presumably the std::span<>. do you?

> I think the implementation will not be able to reuse the vector. If it has a queue of fragments, it won't be a queue of vectors of fragments, but a direct queue of fragments.
> 
> An advantages is that put(temporary_buffer) can call into put(span) and we only have one virtual entry point.

But how? Right now calling put(temporary_buffer) carries the buffer from output_stream into the data_sink. Calling put(span) requires that the buffer itself is stored somewhere, and this "somewhere" shouldn't be the output_stream any longer, it should be std::move()-d from there.

> > If the sink implementation thinks it can do vector put better, than sequentially it can override one and this is what socket sinks do -- they convert the passed vector of buffers into net::packet and proceed. In fact, this is what already happens today, partially at the output_stream level itself, so this change just moves the net::packet instantiation to src/net/ code.
> > File sink benefit from that change by just dropping the plugged put(packet) overload and that's it.
> > The loser here is http content length body sink that stops being zero-copy, but that'll be changed in v2 (if it follows).
> 
> Why does it stop being zero-copy?

It's mostly a note for myself not to forget and fix that in vN+1

Before the patch it had:

* put(packet)  -> zero-copy move the packet into underlying output_stream::write(packet)
* put(temporary_buffer) -> buffered copy the buffer into underlying output_stream::write(buffer.get(), buffer.size())

And this PR just drops the former put() and leaves the latter. One more patch and it's zero-copy again.

> > The output_stream change is -- the _zc_bufs member is changed from net::packet to be std::vector<temporary_buffer>
> 
> Excellent.
> 
> > and zero-copy write()-s accepting packet and scattered_message are deprecated in favor of (exiting) write of temporary_buffer itself.
> > The deprecated write()-s are implemented to convert packet/scattered_message into a vector of buffers and write the buffers themselves. That's going to generate some CPU overhead converting things back-and-forth plus memory allocation, but these methods are deprecated and callers are support to switch to write(temporary_buffer) which doesn't do it.
> 
> span will eliminate the memory allocation for the small cases.

Smallest case (vector with one buffer) doesn't allocate more than the vector storage itself in this PR.
But again, spam is not a container, so whether or not it will eliminate allocations depends on the underlying container. What am I missing?

> > Right now the existing _buf for buffered writes and _zc_bufs with zero-copied buffers co-exist on output_stream and the old non-mixed behavior is preserved. However, next step is the fully-mixed mode itself 🤞

include/seastar/core/iostream.hh
xemul: It should be removed by the end of the PR. In this version there are few places in seaster itself that still call it. I plan to convert them in one of next versions of that PR and drop this deprecated_put().

src/net/native-stack-impl.hh
xemul: OK, but why? Not to re-allocate packet fragments array along with the window change?

src/net/packet.cc
xemul: Not today :D There's a bunch of "appending" constructors for packet:

```c++
    // append fragment (copying new fragment)
    packet(packet&& x, fragment frag);
    // append fragment (zero-copy)
    packet(packet&& x, fragment frag, deleter d);
    // append temporary_buffer (zero-copy)
    packet(packet&& x, temporary_buffer<char> buf);
```

The `packet(std::vector<temporary_buffer>)` one (or better `packet(std::span<temporary_buffer>)`) in the next version TODO list

src/net/posix-stack.cc
xemul: It's pollable_fd::write_all(packet) that instantly maps into sendmsg, what's wrong with using packet at that level?

include/seastar/core/temporary_buffer.hh
xemul: Probably yes, but the way I see it now -- this vector exists only within the output_stream::zero_copy_split_and_put()'s repeat lambda. So it could as well maintain the start iterator and then drop the vector altogether. Looks pretty much the same to what std::span<> would do.

include/seastar/core/iostream.hh
xemul: > Maybe circular_buffer?

I thought of it. It only makes life better in zero_copy_split_and_put(). But [there are](https://github.com/scylladb/seastar/pull/2937#discussion_r2291138302) other ways to handle it.

> btw, do we have flow control over _zc_bufs, or can it grow indefinitely?

It can grow infinitely, but once its total length hits the stream buffer size it's being put() into sink.

It's the same as constructing the net::packet outside of output_stream before write()-ing it today.
E.g.  [`cql_server::response::make_message`](https://github.com/scylladb/scylladb/blob/47750496d21a8260f896f3d203bd50f4ce4135b7/transport/server.cc#L1705-L1716). If it's patched to use this PR's recommended output_stream API, it would grow _zc_bufs as large as it nowadays grows the net::packet.

src/net/native-stack-impl.hh
avikivity: Let's say the workload pushes 300-byte messages (and flushes the output stream after every message). We'll have a series of 300 byte packets (each with its own metadata header + headroom). Then tcp is awakened by a poller, gathers 1500 bytes - tcp overhead into a new packet, throws away the original packet headers. Wasteful.

With a queue of temporary_buffer:s, the tcp code can do

```c++
    if (!q.empty() && adjacent(q.back(), incoming_buffer)) {
         // merge incoming_buffer into q.back()
    } else {
         q.push_back(incoming_buffer);
    }
```

This reduces fragmentation (we observed that drivers have a hard time sending packets with small fragments and @gleb-cloudius even wrote code to linearize them). Even without  the merging, there's just less overhead.

src/net/packet.cc
avikivity: ok

src/net/posix-stack.cc
avikivity: net::packet = device driver friendly structure (it's really struct skbuff). But there's no device driver here and no reason to keep headroom and offload info.

include/seastar/core/temporary_buffer.hh
avikivity: Hmm, now I don't understand _trim_to_size. How can breaking a stream into random chunks work with datagrams?

Probably it should have been done on the data_sink level.

For now, let's just keep the functionality and not care about its performance. Everyone is using tcp anyway.

include/seastar/core/iostream.hh
avikivity: I think a vector is okay here.

Note: erasing from the front will be faster in C++26 due to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2786r11.html.

avikivity:
> > > First, the data_sink. The new API (backward incompatible and thus under new API level) is two put()-s
> > > 
> > > * put(temporary_buffer) -- pure virtual one
> > > * put(vector<temporary_buffer>) -- the implementation falls-back to putting vector elements with the former overload one-by-one
> > 
> > 
> > Should we accept a span here?
> 
> What do you call "a span"? Presumably the std::span<>. do you?

Yes.

> 
> > I think the implementation will not be able to reuse the vector. If it has a queue of fragments, it won't be a queue of vectors of fragments, but a direct queue of fragments.
> > An advantages is that put(temporary_buffer) can call into put(span) and we only have one virtual entry point.
> 
> But how? Right now calling put(temporary_buffer) carries the buffer from output_stream into the data_sink. Calling put(span) requires that the buffer itself is stored somewhere, and this "somewhere" shouldn't be the output_stream any longer, it should be std::move()-d from there.

Yes, data_sink_derived_class::put(std::span\<temporary_buffer>) should move all its input buffers. So the caller can construct a temporary_buffer on the stack, construct a span on the stack containing it, and call put(). By the time put returns (before the future resolved) the arguments are consumed.

> 
> > > If the sink implementation thinks it can do vector put better, than sequentially it can override one and this is what socket sinks do -- they convert the passed vector of buffers into net::packet and proceed. In fact, this is what already happens today, partially at the output_stream level itself, so this change just moves the net::packet instantiation to src/net/ code.
> > > File sink benefit from that change by just dropping the plugged put(packet) overload and that's it.
> > > The loser here is http content length body sink that stops being zero-copy, but that'll be changed in v2 (if it follows).
> > 
> > 
> > Why does it stop being zero-copy?
> 
> It's mostly a note for myself not to forget and fix that in vN+1
> 
> Before the patch it had:
> 
> * put(packet)  -> zero-copy move the packet into underlying output_stream::write(packet)
> * put(temporary_buffer) -> buffered copy the buffer into underlying output_stream::write(buffer.get(), buffer.size())
> 
> And this PR just drops the former put() and leaves the latter. One more patch and it's zero-copy again.

Ok.

> 
> > > The output_stream change is -- the _zc_bufs member is changed from net::packet to be std::vector<temporary_buffer>
> > 
> > 
> > Excellent.
> > > and zero-copy write()-s accepting packet and scattered_message are deprecated in favor of (exiting) write of temporary_buffer itself.
> > > The deprecated write()-s are implemented to convert packet/scattered_message into a vector of buffers and write the buffers themselves. That's going to generate some CPU overhead converting things back-and-forth plus memory allocation, but these methods are deprecated and callers are support to switch to write(temporary_buffer) which doesn't do it.
> > 
> > 
> > span will eliminate the memory allocation for the small cases.
> 
> Smallest case (vector with one buffer) doesn't allocate more than the vector storage itself in this PR. But again, spam is not a container, so whether or not it will eliminate allocations depends on the underlying container. What am I missing?

The allocations of the vector itself? Not comparing to the patches, but to existing state.

> 
> > > Right now the existing _buf for buffered writes and _zc_bufs with zero-copied buffers co-exist on output_stream and the old non-mixed behavior is preserved. However, next step is the fully-mixed mode itself 🤞

include/seastar/core/temporary_buffer.hh
xemul: Not random. Stream is configured with the desired buffer size. Once it accumulates greater-or-equal bytes it flushes. The trim-to-size tell whether sink can be fed with greater-than bytes, or equal-to only is allowed. The remainder (if trim_to_size=true) is kept in stream until next buffer-size hit or explicit flush

xemul:
> Yes, data_sink_derived_class::put(std::span<temporary_buffer>) should move all its input buffers.

Well, I had a version that only had data_sink::put(small_vector<temporary_buffer, 1>), so single buffer vs multiple buffers is all the same, and allocation optimization for single-buffer case. It turned out that all non-network sink implementations are duplicating their code like this

```
    private future<> put(temporary_buffer) {
        // single buffer processing
    }
    public future<> put(small_vector<temporary_buffer> buffers) override {
        for (auto& buf : buffers) {
            co_await put(std::move(buf));
        }
    }
```

Adding virtual single-buffer put() is surprisingly less code. I think it will be the same for spans.

>>> span will eliminate the memory allocation for the small cases.
>>>
>> Smallest case (vector with one buffer) doesn't allocate more than the vector storage itself in this PR. But again, spam is not a container, so whether or not it will eliminate allocations depends on the underlying container. What am I missing?
>>
> The allocations of the vector itself? Not comparing to the patches, but to existing state.
>

Yes. The deprecated write(net::packet) doesn't allocate a vector of temporary_buffers is the packet consists of a single fragment. If comparing to existing state it becomes worse -- before the patches, a write(net::packet) into an empty stream just moves the packet onboard, after the patches a write(net::packet) into an empty stream allocates an inner vector storage to hold temporary_vector constructed out of the packet's buffers

include/seastar/core/temporary_buffer.hh
avikivity: But what does it mean? It's used with UDP (bcea3a67cadf89b74d26fd3b281fbc896b6ffaa5). What does it mean to split on some boundary? Unless the user cooperates and sends headers on exactly that boundary.

avikivity:
> > Yes, data_sink_derived_class::put(std::span<temporary_buffer>) should move all its input buffers.
> 
> Well, I had a version that only had data_sink::put(small_vector<temporary_buffer, 1>), so single buffer vs multiple buffers is all the same, and allocation optimization for single-buffer case. It turned out that all non-network sink implementations are duplicating their code like this
> 
> ```
>     private future<> put(temporary_buffer) {
>         // single buffer processing
>     }
>     public future<> put(small_vector<temporary_buffer> buffers) override {
>         for (auto& buf : buffers) {
>             co_await put(std::move(buf));
>         }
>     }
> ```
> 
> Adding virtual single-buffer put() is surprisingly less code. I think it will be the same for spans.
> 

Ok. But the naive loop is wrong, it's not letting the lower layer coalesce. For example, posix tcp can use a single sendmsg instead of multiple ones.

I think that's some motivation for removing the non-span version - to discourage lazy+bad implementations.


> > > > span will eliminate the memory allocation for the small cases.
> > > 
> > > 
> > > Smallest case (vector with one buffer) doesn't allocate more than the vector storage itself in this PR. But again, spam is not a container, so whether or not it will eliminate allocations depends on the underlying container. What am I missing?
> > 
> > 
> > The allocations of the vector itself? Not comparing to the patches, but to existing state.
> 
> Yes. The deprecated write(net::packet) doesn't allocate a vector of temporary_buffers is the packet consists of a single fragment. If comparing to existing state it becomes worse -- before the patches, a write(net::packet) into an empty stream just moves the packet onboard, after the patches a write(net::packet) into an empty stream allocates an inner vector storage to hold temporary_vector constructed out of the packet's buffers

We shouldn't compare to write(net::packet) since we're deprecating it.

Perhaps we should add write(span\<temporary_buffer>).

include/seastar/core/temporary_buffer.hh
xemul: I don't see how it's used with UDP:

```
$ git grep trim_to_size
apps/memcached/memcache.cc:1256:            opts.trim_to_size = true;
include/seastar/core/iostream-impl.hh:133:            if (_trim_to_size) {
include/seastar/core/iostream-impl.hh:375:                if (_trim_to_size) {
include/seastar/core/iostream-impl.hh:384:            if (_trim_to_size) {
include/seastar/core/iostream.hh:135:    // buffer size and doesn't put larger buffers (see trim_to_size).
include/seastar/core/iostream.hh:382:    bool trim_to_size = false; ///< Make sure that buffers put into sink haven't
include/seastar/core/iostream.hh:389:/// When trim_to_size is true it's guaranteed that data sink will not receive
include/seastar/core/iostream.hh:426:    bool _trim_to_size = false;
include/seastar/core/iostream.hh:447:        : _fd(std::move(fd)), _size(size), _trim_to_size(opts.trim_to_size), _batch_flushes(opts.batch_flushes && _fd.can_batch_flushes()) {}
include/seastar/core/iostream.hh:449:        : _fd(std::move(fd)), _size(_fd.buffer_size()), _trim_to_size(true) {}
src/http/common.cc:153:    opts.trim_to_size = true;
src/http/common.cc:213:    opts.trim_to_size = true;
src/http/transformers.cc:219:    opts.trim_to_size = true;
tests/unit/httpd_test.cc:415:    opts.trim_to_size = true;
tests/unit/httpd_test.cc:429:        opts.trim_to_size = true;
tests/unit/output_stream_test.cc:55:        opts.trim_to_size = do_trim;
```

and memcached uses the stream with vector_data_sink that accumulates data inside in-memory vector<packet>

xemul:
>> Adding virtual single-buffer put() is surprisingly less code. I think it will be the same for spans.
>>
> Ok. But the naive loop is wrong, it's not letting the lower layer coalesce. For example, posix tcp can use a single sendmsg instead of multiple ones.

This is what happens in this PR -- posix and tcp sinks move the whole vector into a packet and pass it lower. The loop I demonstrated occurs in file sink, but since nobody uses it with zero-copy buffers, it doesn't notice that.

include/seastar/core/temporary_buffer.hh
avikivity:
```c++
        future<> respond(udp_channel& chan) {
            int i = 0;
            return do_for_each(_out_bufs.begin(), _out_bufs.end(), [this, i, &chan] (packet& p) mutable {
                header* out_hdr = p.prepend_header<header>(0);
                out_hdr->_request_id = _request_id;
                out_hdr->_sequence_number = i++;
                out_hdr->_n = _out_bufs.size();
                *out_hdr = hton(*out_hdr);
                return chan.send(_src, std::move(p));
            });
        }
```

We can keep trim_to_size, I see it helps http.

include/seastar/core/iostream.hh
xemul: Gone in new version

src/net/packet.cc
xemul: `net::packet(std::span<temporary_buffer>)` constructor in new version

include/seastar/core/iostream.hh
xemul: Removed this overload for API_LEVEL >= 9 in new version

include/seastar/core/iostream.hh
xemul: Removed this overload for API_LEVEL >= 9 in new version

xemul:
upd:

The `data_sink_impl::put(std::span<temporary_buffer<char>>)` is the only method to implement by sinks. The implementation is to grab (move to stable memory) all buffers before returning even unresolved futures. Socket sinks immediately construct net::packet out of spans, other implementations use fallback_put(span) helper that does one-by-one putting, but grabs the buffers from span as it should.

Some cosmetic changes, like nicer span-to-packet conversion, http sinks preserve zero-copy feature, output_stream::write() zero-copy overloads cleanup

xemul:
upd:
Keep chunked-encoded http sink use buffered writes. Filly-mixed mode for output_stream is required

xemul:
upd:
- do not append empty buffers to zc-bufs vector

include/seastar/core/temporary_buffer.hh
```diff
@@ -249,4 +250,34 @@ public:
 
 /// @}
 
+namespace internal {
+
+template <typename CharType>
+inline std::vector<temporary_buffer<CharType>> split_buffers(std::vector<temporary_buffer<CharType>>& bufs, size_t offset) {
```
avikivity: Even though it's internal, the API should be documented, since it's not at all clear what the return values are.

Maybe nicer:

```c++
struct something {
    std::vector<temporary_buffer<char>> before_split_point;
    std::vector<temporary_buffer<char>> after_split_point;
};

something split_buffers(std::vector<temporary_buffer<CharType>> bufs, size_t split_point);
```

If we don't make use of the output vectors, perhaps accept a function to append a before buf and a function to accept and after buf, and call those functions from here.

include/seastar/net/packet.hh
```diff
@@ -228,6 +229,8 @@ public:
     packet(temporary_buffer<char> buf);
     // append deleter
     packet(packet&& x, deleter d);
+    // create from span of buffers (zero-copy, ownership is transfered)
+    packet(std::span<temporary_buffer<char>>);
```
avikivity: explicit

include/seastar/core/iostream.hh
```diff
@@ -151,6 +159,23 @@ public:
...
+    template <typename Fn>
+    requires std::is_invocable_r_v<future<>, Fn, temporary_buffer<char>&&>
+    static future<> fallback_put(std::span<temporary_buffer<char>> bufs, Fn fn) {
+        if (bufs.size() == 1) [[likely]] {
+            return fn(std::move(bufs.front()));
+        }
+
+        auto f = fn(std::move(bufs.front()));
+        for (auto&& buf : bufs.subspan(1)) {
+            f = std::move(f).then([fn, buf = std::move(buf)] () mutable {
```
avikivity, line 175: Doesn't this violate the contract? Nothing holds bufs safe.

I'd like the name of the function to better help the caller understand when to call it.

avikivity: Ah, the name already exists.

avikivity: Well, it should inherit the comment.

avikivity: > Doesn't this violate the contract? Nothing holds bufs safe.

It does not violate the contract, it creates a large future chain.

avikivity: Can be written as: std::ranges::fold_left(bufs | std::views::as_rvalue, make_ready_future<>(), []  (future<>, temporary_buffer) { ... });

include/seastar/core/iostream.hh
```diff
@@ -472,6 +518,8 @@ public:
     future<> write(scattered_message<char_type> msg) noexcept;
     /// Appends the temporary buffer as zero-copy buffer
     future<> write(temporary_buffer<char_type>) noexcept;
+    /// Appends a bunch of buffers as zero-copy
+    future<> write(std::span<temporary_buffer<char_type>>) noexcept;
```
avikivity, line 525: Strict patch splitting would defer it to a later patch and concentrate on data_sink here.

include/seastar/core/vector-data-sink.hh
```diff
@@ -37,10 +37,18 @@ private:
 public:
     vector_data_sink(vector_type& v) : _v(v) {}
 
+#if SEASTAR_API_LEVEL >= 9
+    future<> put(std::span<temporary_buffer<char>> bufs) override {
+        net::packet p = net::packet(bufs);
+        _v.push_back(std::move(p));
+        return make_ready_future<>();
+    }
```
avikivity, line 45: vector_data_sink should be of temporary_buffer too.

include/seastar/websocket/common.hh
```diff
@@ -92,10 +92,18 @@ protected:
     public:
         connection_sink_impl(queue<buff_t>* data) : data(data) {}
 
+#if SEASTAR_API_LEVEL >= 9
+        future<> put(std::span<temporary_buffer<char>> d) override {
+            return data_sink_impl::fallback_put(d, [this] (temporary_buffer<char>&& buf) {
+                return data->push_eventually(std::move(buf));
```
avikivity, line 98: Note: now I think it's bad design to use an extra queue here. There should be one queue, tcp. But not the fault of this series.

src/core/fstream.cc
```diff
@@ -379,11 +379,19 @@ class file_data_sink_impl : public data_sink_impl {
     virtual temporary_buffer<char> allocate_buffer(size_t size) override {
         return temporary_buffer<char>::aligned(_file.memory_dma_alignment(), size);
     }
+#if SEASTAR_API_LEVEL >= 9
+    future<> put(std::span<temporary_buffer<char>> bufs) override {
+        return data_sink_impl::fallback_put(bufs, [this] (temporary_buffer<char>&& buf) {
+            return do_put(std::move(buf));
```
avikivity, line 385: This should be cleared so that files support vectored I/O natively.

include/seastar/core/iostream-impl.hh
```diff
@@ -72,62 +73,59 @@ future<> output_stream<CharType>::write(const std::basic_string<CharType>& s) no
...
 future<>
-output_stream<CharType>::zero_copy_split_and_put(net::packet p) noexcept {
-    return repeat([this, p = std::move(p)] () mutable {
-        if (p.len() < _size) {
-            if (p.len()) {
-                _zc_bufs = std::move(p);
-            } else {
-                _zc_bufs = net::packet::make_null_packet();
-            }
+output_stream<CharType>::zero_copy_split_and_put(std::vector<temporary_buffer<CharType>> b, size_t len) noexcept {
```
avikivity, line 92: Hmm. Maybe we should move split_and_put to the data sink that backs http chunked content? It doesn't seem like output_stream should have this very specific functionality.

But shouldn't be done in this series, it's complicated already.

avikivity:
Looks good. But title needs to indicate the transition from net::packet to span\<t_b>, and cover letter needs updating.

include/seastar/core/iostream.hh
xemul: Yes, but only after we stop supporting C++20 :man_shrugging:

include/seastar/core/iostream.hh
xemul: This write overload is needed for http content-length data sink which in turn needs to forward span of buffers from its put() method down to another _output_stream_, not to other sink. I'll probably re-split this PR a bit differently to address that

src/core/fstream.cc
xemul: Agree, but so far no code uses it. And when we have support for it, how is it going to look like? I mean -- file IO must be aligned, caller cannot just put a bunch of buffered and zero-copy chunks in a hope that pwritev will flush it into a file, all buffers must be aligned.

include/seastar/core/iostream-impl.hh
xemul: It's output_stream that maintains trim-to-size option.
And even if data sink has such a method, what should it do with the "remainder" after split? Emit several subsequent impl->put()-s?

include/seastar/core/temporary_buffer.hh
xemul: Renamed it to `detach_front()` but it still updates the input vector "in place". The something-returning option doesn't look better, but we can change it later.

And added documentation

xemul:
upd:
- renamed split_buffers -> detach_front and documented it
- introduced output_stream::write(span<temporary_buffer>) with separate patch
- changed the way write(packet) converts packet to span of buffers

include/seastar/core/iostream.hh
avikivity: Right

src/core/fstream.cc
avikivity: We can make a choice. Either require that the user align properly, or fix any misalignment.

I think requiring alignment is fine, as long as it's possible to detect what alignment is required.

Possible uses can be writing to a spinning disk, where a reasonable buffer size is 1-2MB, so must be fragmented to 128k fragments.

include/seastar/core/iostream-impl.hh
avikivity: I think output_stream should lose the ability.

Standard data_sinks should not have it. If the user wants to split, they can implement their own interposing data_sink and do anything they want with it.

avikivity:
Did you test the old API_LEVEL still works?

xemul:
Without last patch, that activates the new level, Seastar compiles and its unit tests pass
With the whole PR and level=8 Scylla compiles and cluster.object_store_test_basic passed (it should use both, network and file streams)

xemul:
Unmerged as it conflicted with #3018 moving on parallel track

xemul:
Created #3039

PR #2938, opened 2025-08-22
exceptions: log thrown and propagated exception with distinct log levels

denesb:
Log thrown exceptions with debug and propagated (via future) exceptions with trace levels. Allows seprately logging just thrown exceptions, which are the actual bad ones, hurting performance.

### DIFF
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -43,6 +43,7 @@
 #include <seastar/util/backtrace.hh>
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/modules.hh>
+#include <seastar/util/log-level.hh>
 
 namespace seastar {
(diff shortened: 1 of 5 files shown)


include/seastar/core/future.hh
```diff
@@ -1890,7 +1890,7 @@ future<T> current_exception_as_future() noexcept {
     return future<T>(future_state_base::current_exception_future_marker());
 }
 
-void log_exception_trace() noexcept;
+void log_exception_trace(bool thrown) noexcept;
```
xemul: Maybe pass the log-level itself as an explicit argument?

denesb:
New in v2:
* Rebased on master
* Pass log-level to `log_exception_trace()` instead of `bool`.

PR #2941, opened 2025-08-22
Add --sequential-buffer-size option to IOTune

robertbindar:
Specifying the sequential buffer size becomes helpful as we've recently discovered that i7 NVMEs have saturation request lengths of 1M, IOTune uses 1M buffer size to measure throughput, but Scylla uses buffers of at most 128k which ends up overloading these disks.

We need to be able to tune down IOTune throughput measurements to what Scylla can momentarily achieve by specifying the sequential buffer size of 128k for i7 instances.

Refs #25532


apps/iotune/iotune.cc
```diff
@@ -745,6 +745,7 @@ int main(int ac, char** av) {
         ("random-read-io-buffer-size", bpo::value<unsigned>()->default_value(0), "force buffer size for random read")
         ("force-io-depth", bpo::value<unsigned>()->default_value(0), "force io depth to a certain size (overriding auto detection logic)")
         ("get-best-iops-with-buffer-sizes", bpo::value<std::vector<uint64_t>>()->default_value(std::vector<uint64_t>{}, ""), "run IOPS tests with buffer sizes passed as arg and choose the best combination")
+        ("sequential_buffer_size", bpo::value<uint64_t>()->default_value(1 << 20), "Set the sequential buffer size used for read/write bandwidth calculation. Default is 1MiB")
```
xemul: Random IO buffer sizes have separate options, so why should sequential not have separate too?

PR #2942, opened 2025-08-22
test: Convert directory_test into seastar test

xemul:
It currently has three test-cases, it's easier to maintain them using standard seastar test framework.


xemul:

xemul:

Issue #2190, opened 2024-04-15
Abandoned generator leaks memory


xemul:
Consider the following test case (can be applied to tests/unit/coroutines_test.cc)

```c++
#include <seastar/core/queue.hh>

coroutine::experimental::generator<int> my_generator() {
    auto q = make_shared<queue<std::optional<int>>>(16);
    auto f = q->push_eventually(1).then([] {
        return sleep(1s);
    }).then([q] {
        return q->push_eventually({});
    });

    while (auto i = co_await q->pop_eventually()) {
        co_yield *i;
    }

    co_await std::move(f);
}

SEASTAR_TEST_CASE(test_early_abort_generator)
{
    auto gen = my_generator();
    std::optional<int> i = co_await gen();
    fmt::print("{}\n", i.value_or(42));
}
```

What it does is creates a my_generator, then co-awaits a single result out of it, then exists and (implicitly) destroys the generator. Generator itself, in turn, wants to generate two values, not one. But being destroyed early it leaks memory, reported in debug mode

```
=================================================================
==6430==ERROR: LeakSanitizer: detected memory leaks

Indirect leak of 168 byte(s) in 1 object(s) allocated from:
    #0 0x582641 in operator new(unsigned long) (/home/xemul/src/seastar/build/debug/tests/unit/coroutines_test+0x582641) (BuildId: 7560442d38b7f9121245a32bf098080d3a5f56a1)
    #1 0x830a8c in seastar::shared_ptr<seastar::queue<std::optional<int>>> seastar::shared_ptr_make_helper<seastar::queue<std::optional<int>>, false>::make<int>(int&&) /home/xemul/src/seastar/include/seastar/core/shared_ptr.hh:665:30
    #2 0x7a5616 in seastar::shared_ptr<seastar::queue<std::optional<int>>> seastar::make_shared<seastar::queue<std::optional<int>>, int>(int&&) /home/xemul/src/seastar/include/seastar/core/shared_ptr.hh:684:12
    #3 0x70497e in my_generator() (.resume) /home/xemul/src/seastar/tests/unit/coroutines_test.cc:556:14
```

line 556 here is the `auto q = make_shared<queue<std::optional<int>>>(16);` one

```
Indirect leak of 136 byte(s) in 1 object(s) allocated from:
[...]

PR #2943, opened 2025-08-22
Correctly wrap up abandoned yielding directory lister

xemul:
There are two experimental yielding listers in file.cc -- the posix lister [1] that reads directory and co_yield-s entries, and the fallback yielding lister [2] that implements yielding lister on top of existing seastar::stream-based one. Added by respectively 95a7ff9989 (file: Add co_yield-ing list_directory() method) and 087dddf9ce (file: Add fallback generator implementation).

In case caller doesn't co_await for the lister until it EOFs, the generator is freed but fallback's background fibers are left behind causing use-after-free of the on-stack stuff. Despite it was suspected (#2190) that the memory is leaked by the generator code, it's not -- it _is_ destructed freed correctly, but the fiber is not kicked and continues pushing more entries into the queue.

The fix is in arming a guard that, when called before EOF, aborts the queue and waits the fiber in the background.

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -1342,16 +1342,16 @@ file_impl::dup() {
 }
 
 static coroutine::experimental::generator<directory_entry> make_list_directory_fallback_generator(file_impl& me) {
-    queue<std::optional<directory_entry>> ents(16);
-    auto lister = me.list_directory([&ents] (directory_entry de) {
-        return ents.push_eventually(std::move(de));
+    auto ents = make_lw_shared<queue<std::optional<directory_entry>>>(16);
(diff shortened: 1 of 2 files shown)


xemul:
upd: rebased

xemul:

avikivity:
This will be important when we want to move coroutine generators to be more formally supported. We have to decide on a model to stop consumption. We had this problem with seastar::stream (not well resolved) and with scylladb's mutation_reader (caller is responsible for calling close() and waiting for it, supplying the waiting fiber).

The solution here isn't nice, adding an extra allocation to any coroutine generator that needs to clean up after async work, but at least is convenient.

PR #2945, opened 2025-08-25
reactor: Move stopping activity out of main loop

xemul:
If the reactor is stopping it raises the _stopping flag. Main loop, when sees that, wrap things up and breaks. The "wrap things up" code looks better outside of the loop.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3276,19 +3276,6 @@ int reactor::do_run() {
     while (true) {
         _cpu_sched.run_some_tasks();
         if (_stopped) {
-            load_timer.cancel();
-            // Final tasks may include sending the last response to cpu 0, so run them
-            while (have_more_tasks()) {
-                _cpu_sched.run_some_tasks();
-            }
-            while (_at_destroy_tasks->run_tasks()) {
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -3342,6 +3329,20 @@ int reactor::do_run() {
             }
         }
     }
+
+    load_timer.cancel();
+    // Final tasks may include sending the last response to cpu 0, so run them
```
nyh, line 3334: @avikivity wrote this code (a decade ago!) so I'll let him be the final judge on what "looks better".

On one hand I agree with you that there is additional cleanup code below so it's makes sense to do all the cleanup in one place, below. But on the other hand,  much of the "cleanup code" that you are moving in this patch actually **continues** to run some tasks for a little while, so it also makes sense to keep it where it is in the run loop.

avikivity: I think it's nicer this way.

avikivity: What would be even nicer is to move the various fork/joins into a separate coroutine and launch it from app_template. This way the reactor runs until its initial task is complete.

PR #2950, opened 2025-08-27
test: Add test for input_stream::read_exactly()

xemul:
I came across commit 0e5a5798e9 (iostream: document the read_exactly() function) and noticed that the statement this commit adds was not justified with facts. Here's the test that does it.


xemul:
@nyh , please review

nyh (approved):
Looks good, although the statement in the commit message "not justified with facts" is a bit harsh - it was definitely justified by the actual code. But what is true is that we didn't have a regression test that verifies that it is really true and always remains true, and this is what you added now.

PR #2951, opened 2025-08-28
SEASTAR_ASSERT: assert to stderr and flush stream

travisdowns:
We notice that assertion messages are not being printed as expected.

In these cases you simply see a message like (libc++):

```
libc++abi: terminating
Aborting on shard 0, in scheduling group main.
Backtrace:
```

libstd++ is similar:

```
terminate called without an active exception
Aborting on shard 0, in scheduling group main.
Backtrace:
```

Underlying cause is that we print to stdout, and stdout is not explicitly flushed. In cases where stdout is not line buffered, we will close the message (so e.g., when the output is piped or redirected).

We should (a) assert to stderr, not stdout, and (b) flush the stream.

This is in line with libc and the C++ standard (though it specifies that abort() is called, not std::terminate()).

### DIFF
--- a/src/util/log.cc
+++ b/src/util/log.cc
@@ -124,7 +124,8 @@ namespace seastar {
 namespace internal {
 
 [[noreturn]] void assert_fail(const char* msg, const char* file, int line, const char* func) {
-    printf("%s:%u: %s: Assertion `%s` failed.\n", file, line, func, msg);
+    fprintf(stderr, "%s:%u: %s: Assertion `%s` failed.\n", file, line, func, msg);
+    std::fflush(stderr);
     std::terminate();
 }
 


travisdowns:
FYI @avikivity you may be seeing this on your side, probably a good one to pull.

src/util/log.cc
```diff
@@ -124,7 +124,8 @@ namespace seastar {
 namespace internal {
 
 [[noreturn]] void assert_fail(const char* msg, const char* file, int line, const char* func) {
-    printf("%s:%u: %s: Assertion `%s` failed.\n", file, line, func, msg);
+    fprintf(stderr, "%s:%u: %s: Assertion `%s` failed.\n", file, line, func, msg);
+    std::fflush(stderr);
```
nyh, line 128: I don't understand how we never noticed the missing flush before - stdout is never lined buffer when the output is redirected to a file, and we always redirect it to a file. It's very odd that we never noticed this bug.

That being said, once you changed the output to use stderr, what's the point of the extra fflush(stderr)? unlike stdout, stderr is always unbuffered so the fflush will do nothing. Or maybe you want to do this just in case the user messed around with completely non-standard things for stderr (because technically, if you really want, a user can make stderr buffered if they really really insist)?

nyh (approved):

src/util/log.cc
travisdowns: > I don't understand how we never noticed the missing flush before - stdout is never lined buffer when the output is redirected to a file, and we always redirect it to a file. It's very odd that we never noticed this bug.

Yes, it is surprising. It is easy to reproduce, however: just trigger an assert and pipe stdout like `| cat`. Note that `SEASTAR_ASSERT` was added relatively recently, perhaps scylladb hasn't been using this new impl long enough to notice? Maybe there is also some dependence on libc/lib[std]c++ version, e.g., maybe terminate itself sometimes flushes (but the two variations that I tried both exhibited the issue).

Perhaps you just don't notice the missing assert message given the backtrace sort of gives strictly more info? Looking back I actually saw this a few times previously but was assuming it was a non-assert failure (e.g., throw from `noexcept` function).

> That being said, once you changed the output to use stderr, what's the point of the extra fflush(stderr)? unlike stdout, stderr is always unbuffered so the fflush will do nothing.

My thoughts in list form:

1) "Better safe than sorry"
2) Asymmetric cost/benefit: i.e., if you're right, it's at most a pointless fflush right before the process dies, if you're wrong the assert goes missing
3) I looked up what glibc `assert` does, and it does it like that (this assert replaced `assert`)
4) A brief search for "is stderr always unbuffered" seems to return more or less unequivocal "no, it may be buffered"

avikivity: It's not surprising, ScyllaDB uses syslog for everything, not stdout/stderr.

avikivity: Or maybe, SEASTAR_ASSERTs are just very rare.

travisdowns: > Or maybe, SEASTAR_ASSERTs are just very rare.

Indeed, the asserts are not going through syslog calls, they go to stdout and in the syslog context I suppose this will be buffered. So I suspect it is their rarity. I noticed it in testing, not production.

PR #2957, opened 2025-09-02
Improve io_tester's seqwrite and append workloads

xemul:
The PR does several things with the aforementioned workload types.

First, it fixes the way APPEND interacts with file-size parameter. Nowadays the append fills a file and then starts appending data to it, which is weird -- pre-filling data is thus pretty pointless. After this PR append will start from position of zero all the time and not fill the file beforehand. The configured file size will only be used to truncate the file to that size. Note, that append workload is _not_ O_APPEND and it will write into file from start to end regardless of its size.

Also the PR renames "seqwrite" workload into "overwrite", because strictly speaking "append" is also sequential write. So this new "overwrite" name better explains what the workload type is about.

And, while at it, coroutinizes some code and tosses fields and methods between class_data and io_class_data.

### DIFF
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -103,19 +103,19 @@ future<std::pair<file, uint64_t>> create_and_fill_file(sstring name, uint64_t fs
         co_return std::make_pair(std::move(f), 0ul);
     }
 
-                    const uint64_t buffer_size{256ul << 10};
-                    const uint64_t additional_iteration = (fsize % buffer_size == 0) ? 0 : 1;
-                    const uint64_t buffers_count{static_cast<uint64_t>(fsize / buffer_size) + additional_iteration};
-                    const uint64_t last_buffer_id = (buffers_count - 1u);
-                    const uint64_t last_write_position = buffer_size * last_buffer_id;
-                    auto buffers_range = std::views::iota(UINT64_C(0), buffers_count);
-                    co_await max_concurrent_for_each(buffers_range.begin(), buffers_range.end(), 64, [f, buffer_size] (auto buffer_id) mutable {
-                            auto source_buffer = allocate_and_fill_buffer(buffer_size);
-                            auto write_position = buffer_id * buffer_size;
-                            return do_with(std::move(source_buffer), [f, write_position, buffer_size] (const auto& buffer) mutable {
-                                return f.dma_write(write_position, buffer.get(), buffer_size).discard_result();
-                            });
-                    });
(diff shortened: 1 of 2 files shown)


robertbindar (approved):
Code looks alright, I don't like the choice to add `overwrite` instead of `seqwrite` and make `append` act like what seqwrite from `fio` from instance does, it's confusing to me, but I'm not against it either, I'll create an issue to add an `.md` file in `io_tester` to explain these, it was long needed anyway.

Also, any reason why you don't `fallocate` as well the file given we've seen pretty big throughput improvements in the last round of tests?

robertbindar:
https://github.com/scylladb/seastar/issues/2958

xemul:
> Also, any reason why you don't fallocate as well the file given we've seen pretty big throughput improvements in the last round of tests?

No reasons, but it should come with its own config option in conf.yaml. Maybe incrementally

xemul:
upd:
- added one more patch to correctly configure append_is_unlikely for append workloads

xemul:
> I don't like the choice to add overwrite instead of seqwrite and make append act like what seqwrite from fio from instance does, it's confusing to me

The only thing in "append" this patch does is -- it doesn't pre-fill the file with data, because it's pointless, the workflow would append data past prefilled one anyway. Implicit consequence of it is that after the PR altering `data_size` parameter for "append" workflow finally makes sense -- it controls after which point kernel will shift the file-size on each write.

apps/io_tester/io_tester.cc
```diff
@@ -414,7 +414,7 @@ class class_data {
     // unlink them. So every job (a class in a shard) will have its own file(s) and will operate differently depending on the type:
     //
     // sequential reads  : will read the file from pos = 0 onwards, back to 0 on EOF
-    // sequential writes : will write the file from pos = 0 onwards, back to 0 on EOF
+    // overwrites        : will write the file from pos = 0 onwards, back to 0 on EOF
```
avikivity, line 413: It's also misleading if part of the job is appends, and he rest is overwrites. Maybe the warm-up covers the append part. We should extend the warm-up until we cycle through at least once.

xemul: Yikes. It never overwrites, but appends and appends without end :( I'll fix the comment incrementally

xemul: Wait, no, it always overwrites. All workloads but the append one first write data into a file, then start the measurements, so overwrite will first overwrite the pre-written data, then (given enough time) will overwrite it for the 2nd, 3rd, etc. time

PR #2960, opened 2025-09-02
http: Introduce retry strategy machinery for http client (take two)

kreuzerkrieg:
Introduce extensible retry strategy machinery for the HTTP client to simplify adding new retry conditions. Rather than modifying seastar code to support additional error types, users can now provide their own custom implementations of the retry strategy—offering flexibility without repeated code changes.

Fixes: https://github.com/scylladb/seastar/issues/2803

Scylladb based on these changes: https://github.com/scylladb/scylladb/pull/25801

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -278,6 +278,24 @@ public:
     future<> make_request(request&& req, reply_handler&& handle, std::optional<reply::status_type>&& expected = std::nullopt, abort_source* as = nullptr);
 
     /**
+     * \brief Send the request and handle the response with retry strategy
+     *
+     * Same as \ref make_request()
+     * The retry strategy defines how the client should behave in case of transient failures.
+     *
+     * \param req -- request to be sent
+     * \param handle -- the response handler
+     * \param strategy -- retry strategy to apply on transient failures
+     * \param expected -- the optional expected reply status code, default is std::nullopt
+     * \param as -- abort source that aborts the request
+     *
+     * Note that the handle callback should be prepared to be called more than once, because
+     * client may restart the whole request processing in case server closes the connection
+     * in the middle of operation
+     */
+    future<> make_request(request&& req, reply_handler&& handle, const retry_strategy& strategy, std::optional<reply::status_type>&& expected = std::nullopt, abort_source* as = nullptr);
+
+    /**
      * \brief Send the request and handle the response (abortable), same as \ref make_request()
      *
      *  @attention Note that the method does not take the ownership of the
@@ -285,6 +303,24 @@ public:
      * are referencing valid instances
      */
     future<> make_request(const request& req, reply_handler& handle, std::optional<reply::status_type> expected = std::nullopt, abort_source* as = nullptr);
+
+    /**
+     * \brief Send the request and handle the response with retry strategy (abortable), same as \ref make_request()
+     *
+     * Same as \ref make_request()
+     * The retry strategy defines how the client should behave in case of transient failures.
+     *
+     * \param req -- request to be sent (non-owning reference)
+     * \param handle -- the response handler (non-owning reference)
+     * \param strategy -- retry strategy to apply on transient failures
+     * \param expected -- the optional expected reply status code, default is std::nullopt
+     * \param as -- abort source that aborts the request
+     *
+     * @attention Note that the method does not take the ownership of the
+     * `request` and the `handle`, it is the caller's responsibility to ensure they
+     * are referencing valid instances
+     */
+    future<> make_request(const request& req, reply_handler& handle, const retry_strategy& strategy, std::optional<reply::status_type> expected = std::nullopt, abort_source* as = nullptr);
 
     /**
      * \brief Updates the maximum number of connections a client may have
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -27,6 +27,7 @@
 #include <seastar/net/api.hh>
 #include <seastar/http/connection_factory.hh>
 #include <seastar/http/reply.hh>
+#include <seastar/http/retry_strategy.hh>
 #include <seastar/core/condition-variable.hh>
 #include <seastar/core/iostream.hh>
 #include <seastar/util/modules.hh>
@@ -163,7 +164,7 @@ private:
     unsigned _max_connections;
     size_t _max_bytes_to_drain;
     unsigned long _total_new_connections = 0;
-    const retry_requests _retry;
+    std::unique_ptr<retry_strategy> _retry_strategy;
     condition_variable _wait_con;
     connections_list_t _pool;
 
@@ -180,6 +181,14 @@ private:
     template <typename Fn>
     requires std::invocable<Fn, connection&>
     auto with_new_connection(Fn&& fn, abort_source*);
+
+    future<> maybe_retry_request(std::exception_ptr ex,
+                                 unsigned retry_count,
+                                 const request& req,
+                                 reply_handler& handle,
+                                 const retry_strategy& strategy,
+                                 std::optional<reply::status_type> expected,
+                                 abort_source* as);
 
     future<> do_make_request(connection& con, const request& req, reply_handler& handle, abort_source*, std::optional<reply::status_type> expected);
 
@@ -218,7 +227,7 @@ public:
      * \param f -- the factory pointer
      * \param max_connections -- maximum number of connection a client is allowed to maintain
      * (both active and cached in pool)
-     * \param retry -- whether or not to retry requests on connection IO errors
+     * \param retry_strategy -- optional custom logic for retrying failed requests
      *
      * The client uses connections provided by factory to send requests over and receive responses
      * back. Once request-response cycle is over the connection used for that is kept by a client
@@ -247,6 +256,7 @@ public:
      * second attempt fails, this error is reported back to user.
      */
     explicit client(std::unique_ptr<connection_factory> f, unsigned max_connections = default_max_connections, retry_requests retry = retry_requests::no, size_t max_bytes_to_drain = default_max_bytes_to_drain);
+    client(std::unique_ptr<connection_factory> f, unsigned max_connections, size_t max_bytes_to_drain, std::unique_ptr<retry_strategy>&& retry_strategy);
 
     /**
      * \brief Send the request and handle the response
--- a/include/seastar/http/retry_strategy.hh
+++ b/include/seastar/http/retry_strategy.hh
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2025-present ScyllaDB
+ */
+
+/*
+ * SPDX-License-Identifier: LicenseRef-ScyllaDB-Source-Available-1.0
+ */
+
+#pragma once
+#include <seastar/core/future.hh>
+#include <seastar/http/reply.hh>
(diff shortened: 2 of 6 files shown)


src/http/client.cc
```diff
@@ -366,17 +343,37 @@ future<> client::make_request(const request& req, reply_handler& handle, std::op
...
+                               if (retry) {
+                                    ++retry_count;
+
+                                   // The 'con' connection may not yet be freed, so the total connection
+                                   // count still account for it and with_new_connection() may temporarily
+                                   // break the limit. That's OK, the 'con' will be closed really soon
+
+                                   return with_new_connection(
+                                              [this, &req, &handle, as, expected](connection& con) {
+                                                  return do_make_request(con, req, handle, as, expected).then([]() {
```
xemul: This will hold the `do_with`'s context allocated for every retry until the last one succeeds.
T `retry_strategy::should_retry()` accepts exception_ptr by value, why is it needed to hold one in a stable storage?

include/seastar/http/retry_strategy.hh
```diff
@@ -0,0 +1,55 @@
...
+class retry_strategy {
+    [[nodiscard]] virtual unsigned get_max_retries() const noexcept = 0;
+
+public:
+    virtual ~retry_strategy() = default;
+    // Returns true if the error can be retried given the error and the number of times already tried.
+    virtual future<bool> should_retry(std::exception_ptr error, unsigned attempted_retries) const = 0;
+
+    // Calculates the time in milliseconds the client should wait before attempting another request based on the error and attempted_retries count.
+    [[nodiscard]] virtual std::chrono::milliseconds delay_before_retry(std::exception_ptr error, unsigned attempted_retries) const = 0;
```
xemul: It looks like this method is not called by client::make_request() changes from the next patch

kreuzerkrieg: removed

src/http/client.cc
kreuzerkrieg: got rid of  `do_with`

kreuzerkrieg:
@scylladbbot trigger-ci

src/http/client.cc
kreuzerkrieg: my bad, there is no way around `do_with` here since I need the error being passed by reference since it may be updated in the loop to keep the last error which is passed to the `should_retry` on each iteration

include/seastar/http/retry_strategy.hh
```diff
@@ -0,0 +1,49 @@
...
+#include <seastar/http/reply.hh>
+#include <seastar/util/modules.hh>
+
+namespace seastar {
+SEASTAR_MODULE_EXPORT_BEGIN
+
+namespace http::experimental {
+
+class retry_strategy {
+    [[nodiscard]] virtual unsigned get_max_retries() const noexcept = 0;
```
xemul: Why is this method virtual?
Or even better -- why does it exist at all? The default retry_strategy could just use `this->_max_retries`

include/seastar/http/retry_strategy.hh
```diff
@@ -0,0 +1,49 @@
...
+
+public:
+    virtual ~retry_strategy() = default;
+    // Returns true if the error can be retried given the error and the number of times already tried.
+    virtual future<bool> should_retry(std::exception_ptr error, unsigned attempted_retries) const = 0;
+};
+
+class default_retry_strategy : public retry_strategy {
```
xemul: Either the class is better be marked `final` or all its internals as `protected`?

src/http/retry_strategy.cc
```diff
@@ -0,0 +1,65 @@
...
+    : _max_retries(max_retries) {
+}
+
+default_retry_strategy::default_retry_strategy()
+    : default_retry_strategy(1) {
+}
+
+future<bool> default_retry_strategy::should_retry(std::exception_ptr error, unsigned attempted_retries) const {
+    if (attempted_retries >= get_max_retries()) {
+        rs_logger.warn("Retries exhausted. Retry# {}", attempted_retries);
```
xemul: Too noisy this place is going to be

src/http/retry_strategy.cc
```diff
@@ -31,7 +31,38 @@ extern logger http_log;
 namespace http::experimental {
 logger rs_logger("default_http_retry_strategy");
 
-extern bool is_retryable_exception(std::exception_ptr ex);
+static bool is_retryable_exception(std::exception_ptr ex) {
```
xemul, line 34: nit: Moving this thing deserves its own patch

src/http/client.cc
```diff
@@ -366,17 +343,31 @@ future<> client::make_request(const request& req, reply_handler& handle, std::op
...
+                        }, as).handle_exception([&err](std::exception_ptr error) mutable {
+                            err = std::move(error);
+                            return make_ready_future<std::optional<bool>>();
+                        });
+                    }
+                    return make_exception_future<std::optional<bool>>(std::move(err));
+                });
+            });
+        }).then([](auto) {
+            return make_ready_future<>();
```
xemul: nit: Presumably you want to plug the repeat_unil_value's future<bool> into future<>. If I'm correct, it can do it like

```
    return do_with(..., [] {
        return repeat_until_value([] {
             ...
        }).discard_result();
    });
```

src/http/client.cc
```diff
@@ -366,17 +343,31 @@ future<> client::make_request(const request& req, reply_handler& handle, std::op
...
-        }
-
-        // The 'con' connection may not yet be freed, so the total connection
-        // count still account for it and with_new_connection() may temporarily
-        // break the limit. That's OK, the 'con' will be closed really soon
-        return with_new_connection([this, &req, &handle, as, expected] (connection& con) {
-            return do_make_request(con, req, handle, as, expected);
-        }, as);
+        size_t retry_count = 0;
+        return do_with(std::move (ex), retry_count, [this, &req, &handle, as, expected](std::exception_ptr& err, size_t& retry_count){
```
xemul: I was keeping in mind something like this

```patch
@@ -341,6 +341,17 @@ future<> client::make_request(const request& req, reply_handler& handle, std::op
     return make_request(req, handle, *_retry_strategy, expected, as);
 }
 
+future<> client::maybe_retry_request(std::exception_ptr ex, unsigned retry_count,
+        const request& req, reply_handler& handle, const retry_strategy& strategy, std::optional<reply::status_type> expected, abort_source* as) {
+    return strategy.should_retry(std::move(ex), retry_count).then([this, &req, &handle, &strategy, expected, as] {
+        return with_new_connection([this, &req, &handle, as, expected] (connection& con) {
+            return do_make_request(con, req, handle, as, expected);
+        }, as).handle_exception([this, &req, &handle, &strategy, as, expected] (std::exception_ptr ex) {
+            return maybe_retry_request(std::move(ex), retry_count + 1, req, handle, strategy, expected, as);
+        });
+    });
+}
+
 future<> client::make_request(const request& req, reply_handler& handle, const retry_strategy& strategy, std::optional<reply::status_type> expected, abort_source* as) {
     if (as && as->abort_requested()) {
         return make_exception_future(as->abort_requested_exception_ptr());
@@ -356,31 +367,7 @@ future<> client::make_request(const request& req, reply_handler& handle, const r
         if (as && as->abort_requested()) {
             return make_exception_future<>(as->abort_requested_exception_ptr());
         }
-        size_t retry_count = 0;
-        return do_with(std::move (ex), retry_count, [this, &req, &handle, &strategy, as, expected](std::exception_ptr& err, size_t& retry_count){
-            return repeat_until_value([this, &err, &retry_count, &req, &handle, &strategy, as, expected]() mutable -> future<std::optional<bool>> {
-                return strategy.should_retry(err, retry_count).then([this, &err, &retry_count, &req, &handle, as, expected](bool retry) mutable -> future<std::optional<bool>> {
-                    if (retry) {
-                        ++retry_count;
-                        // The 'con' connection may not yet be freed, so the total connection
-                        // count still account for it and with_new_connection() may temporarily
-                        // break the limit. That's OK, the 'con' will be closed really soon
-
-                        return with_new_connection([this, &req, &handle, as, expected](connection& con) {
-                            return do_make_request(con, req, handle, as, expected).then([]() {
-                                return make_ready_future<std::optional<bool>>(true);
-                            });
-                        }, as).handle_exception([&err](std::exception_ptr error) mutable {
-                            err = std::move(error);
-                            return make_ready_future<std::optional<bool>>();
-                        });
-                    }
-                    return make_exception_future<std::optional<bool>>(std::move(err));
-                });
-            });
-        }).then([](auto) {
-        });
+        return maybe_retry_request(std::move(ex), 0, req, handle, strategy, expected, as);
     });
 }
 
```

and if `strategy::should_retry()` decides that it shouldn't it just resolves into exceptional future with the exception passed to it by value to break the retry loop

xemul:
LGTM, left few comments. The main concern is about do_with + repeat_until_value combination

xemul:
> @scylladbbot trigger-ci

I guess it doesn't work in seastar repo

src/http/retry_strategy.cc
kreuzerkrieg: It is somewhat problematic, compiler starts to complain about unused function and marking it `maybe_unused` still makes the linker to complain it doesnt see the definition of the symbol

include/seastar/http/client.hh
```diff
@@ -276,6 +278,13 @@ public:
      */
     future<> make_request(const request& req, reply_handler& handle, std::optional<reply::status_type> expected = std::nullopt, abort_source* as = nullptr);
 
+    future<> maybe_retry_request(std::exception_ptr ex,
```
xemul: nit: private

src/http/client.cc
```diff
@@ -321,37 +338,24 @@ future<> client::make_request(request&& req, reply_handler&& handle, std::option
...
-        } catch (...) {
-            return false;
+future<> client::maybe_retry_request(std::exception_ptr ex,
+                                     unsigned retry_count,
+                                     const request& req,
+                                     reply_handler& handle,
+                                     const retry_strategy& strategy,
+                                     std::optional<reply::status_type> expected,
+                                     abort_source* as) {
+    return strategy.should_retry(ex, retry_count).then([this, ex = std::move(ex), retry_count, &req, &handle, &strategy, as, expected](bool retry) mutable {
```
xemul, line 358: This copies the exception.
It can be made the other way:

```
future<> stratey::should_retry(std::exception_ptr ex, ...) {
    if (yes) {
        return make_ready_future<>();
    } else {
        return make_exception_future<>(std::move(ex));
    }
}
```

and maybe_retry_request doesn't need to keep ex copy in should_retry and and continuation lambda (and doesn't need to check bool return value)

Thoughts?

kreuzerkrieg: > This copies the exception.

Nope, it is like `shared_ptr` it just counts references for the owned exception

> It can be made the other way:
> 
> ```
> future<> stratey::should_retry(std::exception_ptr ex, ...) {
>     if (yes) {
>         return make_ready_future<>();
>     } else {
>         return make_exception_future<>(std::move(ex));
>     }
> }
> ```
> 
> and maybe_retry_request doesn't need to keep ex copy in should_retry and and continuation lambda (and doesn't need to check bool return value)
> 
> Thoughts?

This is how I did it initially and didnt like it, it looked really odd

PR #2962, opened 2025-09-04
tls: Add helper to skip waiting for EOF on shutdown

xemul:
When TLS socket is closed or shutdown, it spawns a background fiber that waits for the EOF on a socket (and kills it after 10 seconds). That behavior is not always welcome, sometimes client want to close the socket instantly.

In S3 client, there's a function that downloads a file. For that, it sends a GET request with large Range header asking for big portion of data in advance. Sometimes, however, the on-going GET request should be closed by the client before all the arriving data is consumed. In that case client calls connected_socket::close() which immediately launches the aforementioned background EOF wait. As a result, the unwanted data arrives on a node, given NIC speed of 100MB/s, up to 1GB can arrive.

At the same time, this early termination is exceptional situlation, normally the socket data is read and if it gets closed, it gets closed cleanly with no data to be waited, so turning this option off by default is probably bad idea to.

This patch proposes the helper that turns the waiting OFF on a live socket. The intented usage are those emerency cases when the socket has to be closed instantly.

### DIFF
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -495,6 +495,14 @@ namespace tls {
     future<> force_rehandshake(connected_socket& socket);
 
     /**
+     * Clear the "wait for EOF on shutdown" bit.
+     * Useful if the caller is willing to close the connection immediately
+     * Must be called before socket is shutdown or closed, otherwise it
+     * has no effect
+     */
+    void skip_wait_for_eof_on_shutdown(connected_socket& socket);
+
+    /**
      * Subject alt name types.
     */
     enum class subject_alt_name_type {
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -1334,6 +1334,9 @@ public:
         }
         return do_handshake_sync(&session::do_force_rehandshake);
     }
+    void skip_wait_for_eof_on_shutdown() {
+        _options.wait_for_eof_on_shutdown = false;
+    }
 
     future<> handshake() {
         // maybe load system certificates before handshake, in case we
@@ -2078,6 +2081,9 @@ public:
     future<> force_rehandshake() {
         return _session->force_rehandshake();
     }
+    void skip_wait_for_eof_on_shutdown() {
+        _session->skip_wait_for_eof_on_shutdown();
+    }
 };
 
 
@@ -2291,6 +2297,9 @@ future<> tls::force_rehandshake(connected_socket& socket) {
     return s->force_rehandshake();
 }
 
+void tls::skip_wait_for_eof_on_shutdown(connected_socket& socket) {
+    get_tls_socket(socket)->skip_wait_for_eof_on_shutdown();
+}
 
 std::string_view tls::format_as(subject_alt_name_type type) {
     switch (type) {


xemul:
This is for https://github.com/scylladb/scylladb/issues/25714

include/seastar/net/tls.hh
```diff
@@ -494,6 +494,14 @@ namespace tls {
...
     future<> force_rehandshake(connected_socket& socket);
 
+    /**
+     * Clear the "wait for EOF on shutdown" bit.
+     * Useful if the caller is willing to close the connection immediately
+     * Must be called before socket is shutdown or closed, otherwise it
+     * has no effect
+     */
+    void skip_wait_for_eof_on_shutdown(connected_socket& socket);
+
```
avikivity, line 504: Let's assume for a second that this was pure TCP. How would you deal with it?

You'd reset the connection. I think we need an API that mirrors that rather than referring to a bit in an internal implementation.

A little digging shows that it's setting SO_LINGER to 0. So we need an equivalent on connected_socket, which would then be implemented here too.

xemul: > Let's assume for a second that this was pure TCP. How would you deal with it?
> You'd reset the connection. I think we need an API that mirrors that rather than referring to a bit in an internal implementation.

If it comes to unification, then I'd rather go the route that TLS sockets don't wait for EOF _by default_ and it's up to the caller to turn this ON with the option. But nowadays it works the other way around

> A little digging shows that it's setting SO_LINGER to 0

Seastar doesn't seem to do it. Is it gnutls lib that does? And in kernel by default sockets have linger flag cleared.

In fact, this is RFC-2525 Sec 2.17:

> When an application closes a connection in such a way that it can no longer read any received data, the TCP SHOULD ... send a RST if there is any unread received data, or if any new data is received.

Seastar TLS sockets, apparently, behave differently: when a session is closed with data in it, the data is picked up

xemul: |close() condition|TCP|Seastar TLS|"Why not" seastar TLS
|-|-|-|-|
|unread data|send RST, then CLOSE| wait for EOF | just close |
|default|send FIN, then ..._WAIT-s|wait for EOF | wait for EOF |
|forced (a.k.a. SO_LINGER(timeout=0))|send RST, then CLOSE| wait for EOF | just close |

avikivity: > > Let's assume for a second that this was pure TCP. How would you deal with it?
> > You'd reset the connection. I think we need an API that mirrors that rather than referring to a bit in an internal implementation.
> 
> If it comes to unification, then I'd rather go the route that TLS sockets don't wait for EOF _by default_ and it's up to the caller to turn this ON with the option. But nowadays it works the other way around

I'm okay with unifying behavior.

But I think I misread, this is for reading data, yes? It's equivalent to shutdown(SHUT_RD)?


According to Claude, shutdown(SHUT_RD) keeps reading from TCP and only stops the kernel->user interface.

So it seems like we really need to set linger to 0 (for pure TCP) or emulate it (for TLS). In any case, I don't want a special API for TLS, since it's not a TLS problem, it's a general "I don't want to read any more data and I want the connection aborted".

> 
> > A little digging shows that it's setting SO_LINGER to 0
> 
> Seastar doesn't seem to do it. Is it gnutls lib that does? And in kernel by default sockets have linger flag cleared.
> 
> In fact, this is RFC-2525 Sec 2.17:
> 
> > When an application closes a connection in such a way that it can no longer read any received data, the TCP SHOULD ... send a RST if there is any unread received data, or if any new data is received.
> 
> Seastar TLS sockets, apparently, behave differently: when a session is closed with data in it, the data is picked up

But that's not how shutdown(SHUT_RD) or close() behave, no?

xemul: > I'm okay with unifying behavior.
> 
> But I think I misread, this is for reading data, yes? It's equivalent to shutdown(SHUT_RD)?

The issue is about closing the connection. HTTP client wants to terminate a connection before reading all data from it, it sort of succeeds, but TLS leaves a ... let's call it "timewait socket" in the background that continues reading what's being sent. If that were a TCP socket, the connection would be RST-ed (see below)

> So it seems like we really need to set linger to 0 (for pure TCP) or emulate it (for TLS). In any case, I don't want a special API for TLS, since it's not a TLS problem, it's a general "I don't want to read any more data and I want the connection aborted".

OK

> But that's not how shutdown(SHUT_RD) or close() behave, no?

Exactly. Seastar TLS implementation [now differs](https://github.com/scylladb/seastar/pull/2962#discussion_r2330236230) from how TCP sockets behave

shutdown(SHUT_RD) just closes the kernel-user communication, yes
close() -- no, it's logic for ESTABLISHED sockets [is](https://github.com/scylladb/seastar/pull/2962#discussion_r2330236230):
```
    if (data_was_unread || (so_linger && timeout == 0)) {
        send(RST)
        set_state(CLOSE)
    } else {
        send(FIN)
        set_state(FIN_WAIT1)
    }

avikivity: I understood that after close(), without so_linger && timeout === 0 TCP will continue reading in the kernel side. Unread data and newly received data will be discarded.

avikivity: The FIN sent only shuts down the writing half of the connection. The remote may continue writing until it receives an RST or runs out of data.

xemul: > Unread data and newly received data will be discarded.

No, unread data will cause connection to RST upon close()

> The FIN sent only shuts down the writing half of the connection. The remote may continue writing until it receives an RST or runs out of data.

Yes, FIN is like EOF mark that gets queued into the stream after all the previous data

PR #2963, opened 2025-09-04
resource.cc: Remove some dead code

xemul:


xemul:

PR #2964, opened 2025-09-04
memory: Remove deprecated enable_abort_on_allocation_failure()

xemul:
The method was deprecated in favor of set_abort_on_allocation_failure more than 2 years ago (19285ca35f) and can be safely removed now.


xemul:

PR #2965, opened 2025-09-05
More flexibility in configuring io_tester append workload

xemul:
This is the continuation of #2957 

First, when "data_size" parameter is not specified, io_tester will set 1GB by default, which is not what append workload naturally assumes.

Second, appending a pre-truncate-d file and pre-fallocate-d behaves differently and io-tester should be able to explore both scenarios.


xemul:

Issue #669, opened 2019-08-06
Decay large allocation warning threshold with time


tgrabiec:
Currently, on each violation of the threshold, we exponentially increase it. When server has been running for a while, we're guaranteed to not see any warning. When analyzing issues we may not have access to logs since the beginning of server's start, but only a few days. In such cases, we may see no warnings, but not know if that means that there is no problem, or that the threshold is high.

We could exponentially lower the threshold periodically (and log this fact) to solve this.

PR #2967, opened 2025-09-05
memory: Decay large allocation warning threshold

xemul:
After being bumped, the threshold remains high and doesn't capture more
potential problems. The patch splits the threshold into two values --
check and warn, and once the allocation size hits the check part it
either warns it and increases the warn part, or doesn't warn and checks
if the warn threshold can be decreased. The latter happens based on the
lowres_clock timestamps, the next decrease is allowed only after 10
seconds from the prevuous one, the decrease step is page-size.

fixes: #669

### DIFF
--- a/src/core/memory.cc
+++ b/src/core/memory.cc
@@ -548,7 +548,35 @@ struct cpu_pages {
     uint32_t nr_pages;
     uint32_t nr_free_pages;
     uint32_t current_min_free_pages = 0;
-    size_t large_allocation_warning_threshold = std::numeric_limits<size_t>::max();
+    struct {
+        size_t warn = std::numeric_limits<size_t>::max();
+        size_t check = std::numeric_limits<size_t>::max();
+        unsigned descend_attempt = 0;
+
+        void set(size_t value) noexcept {
+            warn = value;
+            check = value;
+            descend_attempt = 0;
+        }
+
+        void ascend() noexcept {
+            warn *= 1.618;
+            descend_attempt = 0;
+        }
+
+        void maybe_descend() noexcept {
+            // Large allocation rate per second is much smaller than one
+            // So in the worst case the warn threshold should be descended after 20+ seconds
+            if (++descend_attempt >= 20) {
+                if (warn > check + page_size) {
+                    warn -= page_size;
+                } else {
+                    warn = check;
+                }
+                descend_attempt = 0;
+            }
+        }
+    } large_allocation_warning_threshold = {};
     unsigned cpu_id = -1U;
     std::function<void (std::function<void ()>)> reclaim_hook;
     std::vector<reclaimer*> reclaimers;
@@ -838,7 +866,6 @@ void
 cpu_pages::warn_large_allocation(size_t size) {
     alloc_stats::increment_local(alloc_stats::types::large_allocs);
     seastar_memory_logger.warn("oversized allocation: {} bytes. This is non-fatal, but could lead to latency and/or fragmentation issues. Please report: at {}", size, current_backtrace());
-    large_allocation_warning_threshold *= 1.618; // prevent spam
 }
 
 allocation_site_ptr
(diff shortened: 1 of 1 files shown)


travisdowns:
This is a good idea, we've discussed same internally.

src/core/memory.cc
```diff
@@ -1234,6 +1243,7 @@ bool cpu_pages::initialize() {
     pages[nr_pages].free = false;
     free_span_unaligned(reserved, nr_pages - reserved);
     live_cpus[cpu_id].store(true, std::memory_order_relaxed);
+    large_allocation_warning_decay.set_callback(default_scheduling_group(), [] { internal::decrease_large_allocation_warning_threshold(); });
```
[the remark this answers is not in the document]
xemul: Timers don't migrate between CPUs in seastar

src/core/memory.cc
```diff
@@ -1813,14 +1823,30 @@ reclaimer::~reclaimer() {
...
+}
+
+namespace internal {
+void decrease_large_allocation_warning_threshold() {
+    size_t& threshold = get_cpu_mem().large_allocation_warning_threshold;
+    size_t base = get_cpu_mem().large_allocation_warning_threshold_base;
+
+    if (threshold > base + cpu_pages::large_allocation_warning_decay_step) {
+        threshold -= cpu_pages::large_allocation_warning_decay_step;
+        get_cpu_mem().large_allocation_warning_decay.arm(cpu_pages::large_allocation_warning_decay_period);
```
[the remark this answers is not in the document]
xemul: If timer execution delays, "now()" will also be delayed
Also, seastar timers' relative arms are overloads on top of absolute that do exactly the same anyway
And finally -- we don't care about potential delays, this decay is not about precision timing

src/core/memory.cc
```diff
@@ -553,6 +553,10 @@ struct cpu_pages {
     uint32_t nr_free_pages;
     uint32_t current_min_free_pages = 0;
     size_t large_allocation_warning_threshold = std::numeric_limits<size_t>::max();
+    size_t large_allocation_warning_threshold_base = std::numeric_limits<size_t>::max();
+    static constexpr size_t large_allocation_warning_decay_step = page_size;
+    static constexpr lowres_clock::duration large_allocation_warning_decay_period = std::chrono::seconds(10);
+    timer<lowres_clock> large_allocation_warning_decay;
```
avikivity: memory.cc is much lower level than the reactor etc.

I'd like to avoid entangling it with the rest of the reactor.

avikivity: I'd even prefer to avoid having it talk with the clocks, but I don't see a way.

How about this:

```
if larger than original threshold)
   if larger than current threshold
       report
       consider increasing threshold based on history
   else
       consider decreasing threshold based on history
```

xemul: If considering without clocks, then what's the ... average (?) expected (?) typical (?) allocations-per-second rate to rely on?

xemul:
upd:
- do threshold decreasing based on lowres_clock timestamps

xemul:
@avikivity , please re-review

src/core/memory.cc
avikivity: large allocation rate per second is much smaller than one.

avikivity:
Sorry about the delay. Maybe I should use a timer.

Even lowres_clock is too entangled. memory.cc depends on nothing now, let's keep it that way.

Decay can be every N large allocations that aren't greater than the current threshold. If N = 20, and large allocations are rare, that throttles large allocations significantly but still allows them to surface.

src/core/memory.cc
xemul: > large allocation rate per second is much smaller than one.

OK, let's take this as an axiom

xemul:
upd:
- descend the warn threshold after check threshold is hit several times

PR #2968, opened 2025-09-08
rpc: batch destroy rpc buffers on owner shard

bhalevy:
Prevent long cross-shard task queues for destroying foreign rpc buffers.  Instead, queue up buffers to destroy on the local shard while the previous batch is being destroyed on the owner shard.  The worker fiber picks up the list of buffers to destroy in a loop, and then destroying the whole batch, one buffer at a time, on the owner shard.

Refs scylladb/scylladb#24818

### DIFF
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -42,10 +42,13 @@
 #include <seastar/core/scheduling.hh>
 #include <seastar/util/backtrace.hh>
 #include <seastar/util/log.hh>
+#include <seastar/rpc/rpc_types.hh>
 
 namespace bi = boost::intrusive;
 
 namespace seastar {
+
+extern logger seastar_logger;
 
 namespace rpc {
 
@@ -323,7 +326,7 @@ public:
     bool error() const noexcept { return _error; }
     void abort();
     future<> stop() noexcept;
-    future<> stream_receive(circular_buffer<foreign_ptr<std::unique_ptr<rcv_buf>>>& bufs);
+    future<> stream_receive(circular_buffer<foreign_rpc_buf<rcv_buf>>& bufs);
     future<> close_sink() {
         _sink_closed = true;
         if (stream_check_twoway_closed()) {
@@ -392,10 +395,28 @@ class sink_impl : public sink<Out...>::impl {
     alignas (cache_line_size) uint64_t _next_seq_num = 1;
 
     // Used on the shard the _conn lives on.
-    struct alignas (cache_line_size) {
+    struct alignas (cache_line_size) remote_state {
         uint64_t last_seq_num = 0;
         std::map<uint64_t, deferred_snd_buf> out_of_order_bufs;
-    } _remote_state;
+        xshard_destroy_queue<snd_buf> destroy_queue;
+    };
+
+    class snd_buf_deleter_impl final : public deleter::impl {
+        remote_state& _state;
+        foreign_rpc_buf<snd_buf> _obj;
+    public:
+        snd_buf_deleter_impl(remote_state& state, foreign_rpc_buf<snd_buf> obj)
+            : impl(deleter())
+            , _state(state), _obj(std::move(obj))
+        {}
+        snd_buf_deleter_impl(const snd_buf_deleter_impl&) = delete;
+        snd_buf_deleter_impl(snd_buf_deleter_impl&&) = delete;
+        virtual ~snd_buf_deleter_impl() override {
+            _state.destroy_queue.delete_buf(std::move(_obj));
+        }
+    };
+
+    remote_state _remote_state;
 public:
     sink_impl(xshard_connection_ptr con) : sink<Out...>::impl(std::move(con)) { this->_con->get()->_sink_closed = false; }
     future<> operator()(const Out&... args) override;
--- a/include/seastar/rpc/rpc_impl.hh
+++ b/include/seastar/rpc/rpc_impl.hh
@@ -30,6 +30,7 @@
 #include <seastar/core/simple-stream.hh>
 #include <seastar/net/packet-data-source.hh>
 #include <seastar/core/deleter.hh>
+#include <seastar/rpc/rpc_types.hh>
 
 #include <boost/type.hpp> // for compatibility
 
@@ -823,10 +824,7 @@ std::optional<protocol_base::handler_with_holder> protocol<Serializer, MsgType>:
     return std::nullopt;
 }
 
-template<typename T> T make_shard_local_buffer_copy(foreign_ptr<std::unique_ptr<T>> org,
-        std::function<deleter(foreign_ptr<std::unique_ptr<T>> org)> make_deleter = [] (foreign_ptr<std::unique_ptr<T>> org) {
-            return make_object_deleter(std::move(org));
-        });
+template<typename T> T make_shard_local_buffer_copy(foreign_rpc_buf<T> org, std::function<deleter(foreign_rpc_buf<T> org)> make_deleter);
 
 template<typename Serializer, typename... Out>
 future<> sink_impl<Serializer, Out...>::operator()(const Out&... args) {
@@ -840,7 +838,7 @@ future<> sink_impl<Serializer, Out...>::operator()(const Out&... args) {
     // but only one at a time
     auto size = std::min(size_t(data.size), max_stream_buffers_memory);
     const auto seq_num = _next_seq_num++;
-    return get_units(this->_sem, size).then([this, data = make_foreign(std::make_unique<snd_buf>(std::move(data))), seq_num] (semaphore_units<> su) mutable {
+    return get_units(this->_sem, size).then([this, data = foreign_rpc_buf(std::make_unique<snd_buf>(std::move(data))), seq_num] (semaphore_units<> su) mutable {
         if (this->_ex) {
             return make_exception_future(this->_ex);
         }
@@ -858,7 +856,10 @@ future<> sink_impl<Serializer, Out...>::operator()(const Out&... args) {
             auto& last_seq_num = _remote_state.last_seq_num;
             auto& out_of_order_bufs = _remote_state.out_of_order_bufs;
 
-            auto local_data = make_shard_local_buffer_copy(std::move(data));
+            std::function<deleter(foreign_rpc_buf<snd_buf> org)> make_deleter = [this] (foreign_rpc_buf<snd_buf> org) {
+                return deleter(new snd_buf_deleter_impl(_remote_state, std::move(org)));
+            };
+            auto local_data = make_shard_local_buffer_copy(std::move(data), make_deleter);
             const auto seq_num_diff = seq_num - last_seq_num;
             if (seq_num_diff > 1) {
                 auto [it, _] = out_of_order_bufs.emplace(seq_num, deferred_snd_buf{promise<>{}, std::move(local_data)});
@@ -915,7 +916,9 @@ future<> sink_impl<Serializer, Out...>::close() {
             } else {
                 f = this->_ex ? make_exception_future(this->_ex) : make_exception_future(closed_error());
             }
-            return f.finally([con] { return con->close_sink(); });
+            return f.finally([con] { return con->close_sink(); }).finally([this] {
+                return _remote_state.destroy_queue.stop();
+            });
         });
     });
 }
@@ -930,12 +933,15 @@ sink_impl<Serializer, Out...>::~sink_impl() {
 template<typename Serializer, typename... In>
 future<std::optional<std::tuple<In...>>> source_impl<Serializer, In...>::operator()() {
     auto process_one_buffer = [this] {
-        foreign_ptr<std::unique_ptr<rcv_buf>> buf = std::move(this->_bufs.front());
+        foreign_rpc_buf<rcv_buf> buf = std::move(this->_bufs.front());
         this->_bufs.pop_front();
+        std::function<deleter(foreign_rpc_buf<rcv_buf> org)> make_deleter = [this] (foreign_rpc_buf<rcv_buf> org) {
+            return deleter{this->make_rcv_buf_deleter(std::move(org))};
+        };
         return std::apply([] (In&&... args) {
             auto ret = std::make_optional(std::make_tuple(std::move(args)...));
             return make_ready_future<std::optional<std::tuple<In...>>>(std::move(ret));
-        }, unmarshall<Serializer, In...>(*this->_con->get(), make_shard_local_buffer_copy(std::move(buf))));
+        }, unmarshall<Serializer, In...>(*this->_con->get(), make_shard_local_buffer_copy(std::move(buf), make_deleter)));
     };
 
     if (!this->_bufs.empty()) {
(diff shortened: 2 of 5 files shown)


bhalevy:
- fixed typos in git commit message

include/seastar/rpc/rpc_impl.hh
```diff
@@ -842,7 +842,7 @@ future<> sink_impl<Serializer, Out...>::operator()(const Out&... args) {
         }
         // It is OK to discard this future. The user is required to
         // wait for it when closing.
-        (void)smp::submit_to(this->_con->get_owner_shard(), [this, data = std::move(data), seq_num] () mutable {
+        (void)smp::submit_to(this->_con->get_owner_shard(), [this, local_data = make_shard_local_buffer_copy(data), seq_num] () mutable {
```
gleb-cloudius: That does not do what you want it to do. It runs on a local shard, not remote one.

avikivity: Perhaps we need a circular_buffer similar to _bufs but in the other direction.

bhalevy: > That does not do what you want it to do. It runs on a local shard, not remote one.

It doesn't matter where `local_data` is prepared since it just contains pointers to the buffers and they are not deleted, so it can be prepared on the calling shard as well as on the callee shard.

gleb-cloudius: It may also contains a vector which will be be freed on the wrong cpu through the slow shared queue.

bhalevy: Hmm, I tried to avoid an extra allocation to keep the foreign ptr alive across the submit_to call.  Moving it to the .then continuation is too early since the local_data is prepared on the remote shard. Can we keep it as a member in the sink_impl object?

gleb-cloudius: Keep what as a member?

I understand what you are trying to do. You want to keep original buffer alive on the original shard until the end of processing. Note that you need to audit the entire code path that the buffer returned by `make_shard_local_buffer_copy` cannot outlive the fiber completion because the way it is now it is safe to do so. But may be this is overkill and the buffer's lifetime is contained in the fiber's lifetime. In this case you do not even need to wrap it into `foreign_ptr` even.

You need to solve the problem I pointed out though somehow and I do not understand what exactly you propose here.

bhalevy: Although we move local_data snd_buf to connection::send it is indeed processed in the background.

> Perhaps we need a circular_buffer similar to _bufs but in the other direction.

@avikivity that might not be a bad idea for queuing buffers to destroy and signaling the owner shard, that can destroy the queued buffers in a batch when waking up.  Instead of queuing cross-shard tasks to do that one by one.

avikivity: Well, the logic was that the cross-shard task queue is also an efficient queue, we shouldn't need to optimize it.

But maybe we should batch the smp calls. Each call can carry all the buffers that were freed since the last call.

bhalevy: There's also an extra allocation in smp_message_queue https://github.com/scylladb/seastar/blob/d2a5a8a9c73c5ac16765eae6849d9fb170e198d5/include/seastar/core/smp.hh#L286 that I don't like.

Regarding batching, don't we already have it?
https://github.com/scylladb/seastar/blob/d2a5a8a9c73c5ac16765eae6849d9fb170e198d5/src/core/reactor.cc#L3604-L3606

and https://github.com/scylladb/seastar/blob/d2a5a8a9c73c5ac16765eae6849d9fb170e198d5/src/core/reactor.cc#L3567-L3580

bhalevy: > But maybe we should batch the smp calls. Each call can carry all the buffers that were freed since the last call.

And do that using a timer to let more buffers get freed on the sending shard before submitting the queue to the owner shard,  or use atomic instructions similar to the smp message queue?

avikivity: No need for timers.

I'm talking about batching the return path.

bhalevy: > No need for timers.
> 
> I'm talking about batching the return path.

Done in [3972199](https://github.com/scylladb/seastar/commit/39721995c75f0134601845ac81a92cf3c43f905d)

bhalevy:
In [3972199](https://github.com/scylladb/seastar/commit/39721995c75f0134601845ac81a92cf3c43f905d):
- implemented per-shard queues for batch-destroying snd_buf:s across shards.

include/seastar/core/sharded.hh
```diff
@@ -978,6 +983,13 @@ private:
         }
 #endif
     }
+
+    PtrType unsafe_release() noexcept(std::is_nothrow_default_constructible_v<PtrType>) {
+        return std::exchange(_value, {});
+    }
+
+    template<typename Serializer, typename... Out>
+    friend class rpc::sink_impl;
```
gleb-cloudius: Seriously?

include/seastar/core/sharded.hh
```diff
@@ -1049,7 +1061,7 @@ public:
     /// owner shard to avoid accidents.
     PtrType release() noexcept(std::is_nothrow_default_constructible_v<PtrType>) {
         check_shard();
-        return std::exchange(_value, {});
+        return unsafe_release();
```
gleb-cloudius: Why use `foreign_ptr` in the first place?

avikivity:
I don't see a test

include/seastar/rpc/rpc.hh
```diff
@@ -392,12 +392,37 @@ class sink_impl : public sink<Out...>::impl {
...
-    struct alignas (cache_line_size) {
+    struct alignas (cache_line_size) remote_state {
+        using snd_buf_list_t = boost::intrusive::list<snd_buf, boost::intrusive::constant_time_size<false>>;
+
+        shard_id shard;
         uint64_t last_seq_num = 0;
         std::map<uint64_t, deferred_snd_buf> out_of_order_bufs;
-    } _remote_state;
+        std::vector<snd_buf_list_t> per_shard_snd_bufs_to_destroy;
+        std::vector<future<>> per_shard_snd_bufs_destroy_futures;
```
gleb-cloudius: I do not understand what those arrays are doing here. One sink can  only forward to one shard.

include/seastar/rpc/rpc_impl.hh
```diff
@@ -823,10 +823,43 @@ std::optional<protocol_base::handler_with_holder> protocol<Serializer, MsgType>:
...
+
+template<typename Serializer, typename... Out>
+sink_impl<Serializer, Out...>::snd_buf_deleter_impl::~snd_buf_deleter_impl() {
+    // We cannot delete the buffer directly, since it may be
+    // accessed by the remote shard. Instead, we enqueue it
+    // for deletion on the remote shard.
+    auto owner_shard = _obj.get_owner_shard();
+    auto& lst = _state.per_shard_snd_bufs_to_destroy[owner_shard];
+    auto& fut = _state.per_shard_snd_bufs_destroy_futures[owner_shard];
+    lst.push_back(*_obj.unsafe_release().release());
```
gleb-cloudius: Why is it safe to push to a list that you may be destroying on another shard?

include/seastar/rpc/rpc_impl.hh
```diff
@@ -823,10 +823,43 @@ std::optional<protocol_base::handler_with_holder> protocol<Serializer, MsgType>:
...
+            return smp::submit_to(owner_shard, [lst = std::move(lst)] () mutable {
+                return do_until([&lst] { return lst.empty(); }, [&lst] {
+                    auto* buf_ptr = &lst.front();
+                    // Destroy the snd_buf (this will call its destructor).
+                    lst.pop_front();
+                    // And free its memory on the owner shard (that was allocated by std::unique_ptr<snd_buf>)
+                    std::default_delete<snd_buf>()(buf_ptr);
+                    return make_ready_future();
+                });
+            });
```
gleb-cloudius: I do not know what to say. Looks like a badly implemented foreign_ptr. I do not underrated what is wrong with smp queue batching.

bhalevy:
> I don't see a test

Avi, the seastar rpc unit test covers this path in the sense that it easily caught issues during development of this branch like use-after-free (stack and heap) and memory leaks.

include/seastar/core/sharded.hh
bhalevy: I think we can work around it, if we make sure that the original buffer is properly destroyed on all successful and error paths.

include/seastar/rpc/rpc.hh
bhalevy: ah, I was under the impression that it could forward to all shards.

include/seastar/rpc/rpc_impl.hh
bhalevy: It is an intrusive list so it is not destroyed per say.
But it is a good point if we can touch the boost intrusive list hook across shards, like one pushing to the back and one popping from the front.

Edit: please disregard the last sentence. In fact, the batched queue is moved-from on the **calling** shard, into a lambda capture that is drained on the submitted-to shard while new buffers can be safely queued onto a new list while the previous one is being destroyed, so there is no cross-shard race.

gleb-cloudius: By destroying I mean removing items from it and no you cannot.

avikivity:
> > I don't see a test
> 
> Avi, the seastar rpc unit test covers this path in the sense that it easily caught issues during development of this branch like use-after-free (stack and heap) and memory leaks.

Do they check that this patch accomplishes what it sets out to do?

include/seastar/core/sharded.hh
bhalevy: Either this or remove the `check_shard()` call from `release()` or expose an `unsafe_release()` in the api, r add an `bool_class unsafe` parameter to `release()` defaulted to `unsafe::no` showing that the caller is aware of the risks of releasing the wrapped pointer on a non-owner shard.

gleb-cloudius: IMO the need for such a function shows that `foreign_ptr` is used where it should not be.

include/seastar/rpc/rpc_impl.hh
bhalevy: The problem was state here: https://github.com/scylladb/scylladb/issues/24818#issuecomment-3223534070:

> Long queue = stall.
> 
> When Seastar detects a long queue, it stops polling for new I/O completions, because an I/O completion can add tasks to the queue causing unbounded growth.
> 
> Instead it goes into emergency mode and processes tasks until the queue is short again. This emergency processing is one long (or very short) stall.

I'd add that the smp queue also adds yet another allocation, but this is just room for optimization, not the gist of the problem.

gleb-cloudius: I do not see how implementing ad hoc queue (incorrectly) will solve a long queue. This new queue will be long. OK, so you would not stall because of it, but them it will grow indefinitely. We need to fix back pressure mechanism in streaming to include not yet released foreign ptrs. Currently we release semaphore too early.

bhalevy:
In [9b8b02e](https://github.com/scylladb/seastar/commit/9b8b02edd1b8c2e8bfa44003e1b3d2c31a6dc191):
- replace foreign_ptr with a tailored foreign_rpc_buffer structure
- derive snd_bug and rcv_buf from boost::intrusive::list_base_hook for queuing them up in a xshard_destroy_queue to be destroyed in batches

I'm working on reproducing scylladb/scylladb#24818 in a unit test to verify that this patch indeed fixes the long smp task queue issue.

FWIW, I tested this patch manually by making the first destroy sleep for some while and observed the other buffer deletes queue up and be destroyed in a batch as planned.

include/seastar/rpc/rpc_impl.hh
bhalevy: > I do not see how implementing ad hoc queue (incorrectly) will solve a long queue.

Where is the incorrectness?
It's not aimed at solving the long queue, but improving its execution across cpus.

> This new queue will be long. OK, so you would not stall because of it, but them it will grow indefinitely. We need to fix back pressure mechanism in streaming to include not yet released foreign ptrs. Currently we release semaphore too early.

I'd leave that to you, as I'm not familiar with this part of the code at all.

gleb-cloudius: > > I do not see how implementing ad hoc queue (incorrectly) will solve a long queue.
> 
> Where is the incorrectness? It's not aimed at solving the long queue, but improving its execution across cpus.
> 

Latest code solved access to the same queue from two shards as far as I see, so may be it is not incorrect any longer but it is ad hoc. We have mechanism to send work across shard and batch it.

> > This new queue will be long. OK, so you would not stall because of it, but them it will grow indefinitely. We need to fix back pressure mechanism in streaming to include not yet released foreign ptrs. Currently we release semaphore too early.
> 
> I'd leave that to you, as I'm not familiar with this part of the code at all.

You are trying to do something much more complicated in the code you claim you are not familiar with. The problem (as far as I can tell since we do not have a reproducer) is that backpresure mechanism in `sink_impl<Serializer, Out...>::operator()` does not work as expected. It uses semaphore `_sem` to limit the amount of work on the other shard but units are released before `foreign_ptr` is freed, so another shard may accumulate a lot of them. The solution as I see it is to make `semaphore_units` lifetime to be the same as `foreign_ptr` (by holding it in the `snd_buf` for instance).

bhalevy: you mean something like this?
```diff
diff --git a/include/seastar/rpc/rpc_impl.hh b/include/seastar/rpc/rpc_impl.hh
index bcb0d8f38..68b74a42e 100644
--- a/include/seastar/rpc/rpc_impl.hh
+++ b/include/seastar/rpc/rpc_impl.hh
@@ -840,6 +840,7 @@ future<> sink_impl<Serializer, Out...>::operator()(const Out&... args) {
         if (this->_ex) {
             return make_exception_future(this->_ex);
         }
+        data->su = std::move(su);
         // It is OK to discard this future. The user is required to
         // wait for it when closing.
         (void)smp::submit_to(this->_con->get_owner_shard(), [this, data = std::move(data), seq_num] () mutable {
@@ -871,7 +872,7 @@ future<> sink_impl<Serializer, Out...>::operator()(const Out&... args) {
                 out_of_order_bufs.erase(it);
             }
             return ret_fut;
-        }).then_wrapped([su = std::move(su), this] (future<> f) {
+        }).then_wrapped([this] (future<> f) {
             if (f.failed() && !this->_ex) { // first error is the interesting one
                 this->_ex = f.get_exception();
             } else {
diff --git a/include/seastar/rpc/rpc_types.hh b/include/seastar/rpc/rpc_types.hh
index d69a38b33..8bb464f59 100644
--- a/include/seastar/rpc/rpc_types.hh
+++ b/include/seastar/rpc/rpc_types.hh
@@ -41,6 +41,7 @@
 #include <seastar/core/lowres_clock.hh>
 #include <boost/functional/hash.hpp>
 #include <seastar/core/sharded.hh>
+#include <seastar/core/semaphore.hh>
 
 namespace seastar {
 
@@ -254,6 +255,7 @@ struct snd_buf {
     static constexpr size_t chunk_size = 128*1024;
     uint32_t size = 0;
     std::variant<std::vector<temporary_buffer<char>>, temporary_buffer<char>> bufs;
+    semaphore_units<> su;
     using iterator = std::vector<temporary_buffer<char>>::iterator;
     snd_buf() {}
     snd_buf(snd_buf&&) noexcept;
```

gleb-cloudius: Yes.

bhalevy:
Replaced by https://github.com/scylladb/seastar/pull/2980

PR #2971, opened 2025-09-09
core/coroutine: add a specialization for std::expected<void, Err>

rockwotj:
As recorded in https://www.scylladb.com/2024/05/14/better-goodput-performance-through-c-exception-handling/ under `Approach 1: Avoid The Exceptions` there are places where we would like to avoid exceptions. One approach we do is use `boost::result`, but now that we're on C++23 have instead favored using `std::expected` as it's a bit simpler and "modern" (for what that is worth).

However, for side effecting functions we've started to represent that with `std::expected<void, Err>`, which is fine, but very cumbersome to fully type out. We'd like to specialize coroutine_traits to make this case less verbose, but can't do it in our application due to it being in an `internal` namespace and compatibility not being ensured.

So we propose a specialization for `std::expected` that allows the syntax `co_return {};`. This is really a small quality of life improvement for us we'd like to understand if there is an appetite for having upstream.


include/seastar/core/coroutine.hh
```diff
@@ -130,6 +133,59 @@ public:
...
+        promise_type(const promise_type&) = delete;
+
+        template<typename... U>
+        void return_value(U&&... value) {
+            _promise.set_value(std::forward<U>(value)...);
+        }
+
+        void return_value(std::expected<void, Err>&& value) {
+            _promise.set_value(std::forward<std::expected<void, Err>>(value));
+        }
```
avikivity, line 154: I like it (and we want to move from boost::outcome to std::expected too).

But there's too much copying here. Perhaps add this to the main template with a constraint?

rockwotj: @avikivity are you infavor of https://github.com/scylladb/seastar/pull/2977? If so, I'll close this patch as it also handles the case.

avikivity: It's better to do it deliberately in its own pull request rather than sneak it in as part of another change.

rockwotj: Well the fundamental issue is that the compiler can't infer the return type because of the template in `return_value` so if we want to move to `return_value` not being a template then this case doesn't need to be specialized anymore.

avikivity: ok

PR #2973, opened 2025-09-10
coroutine: introduce try_future

denesb:
Specialized awaiter for future<T> which returns control to the coroutine only if the future didn't fail. If the future failed, the returned exception is returned to the waiter directly, without resuming the coroutine. This allows for elegant control-flow, without the if spamming that is required whe using coroutine::as_future().

### DIFF
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -40,8 +40,12 @@ void
 execute_involving_handle_destruction_in_await_suspend(std::invocable<> auto&& func) noexcept {
 #if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ > 15 || (__GNUC__ == 15 && __GNUC_MINOR__ >= 2))
     // See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=121961
-    memory::scoped_critical_alloc_section _;
-    schedule(new lambda_task(current_scheduling_group(), std::forward<decltype(func)>(func)));
+    if constexpr (std::is_base_of_v<seastar::task, std::remove_cvref_t<decltype(func)>>) {
+        schedule(&func);
+    } else {
+        memory::scoped_critical_alloc_section _;
+        schedule(new lambda_task(current_scheduling_group(), std::forward<decltype(func)>(func)));
+    }
 #else
     std::invoke(std::forward<decltype(func)>(func));
 #endif
--- a/include/seastar/coroutine/try_future.hh
+++ b/include/seastar/coroutine/try_future.hh
@@ -1,0 +1,149 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2025-present ScyllaDB
+ */
+
+#pragma once
+
+#include <seastar/core/coroutine.hh>
+
+namespace seastar::internal {
+
+template <typename T, typename U>
+void try_future_resume_or_destroy_coroutine(seastar::future<T>& fut, seastar::task& coroutine_task) {
+    auto promise_ptr = static_cast<U*>(&coroutine_task);
+    auto hndl = std::coroutine_handle<U>::from_promise(*promise_ptr);
+
+    if (fut.failed()) {
+        hndl.promise().set_exception(std::move(fut).get_exception());
+        hndl.destroy();
+    } else {
+        hndl.resume();
+    }
+}
+
+template <bool CheckPreempt, typename T>
+class [[nodiscard]] try_future_awaiter : public seastar::task {
+    seastar::future<T> _future;
+    void (*_resume_or_destroy)(seastar::future<T>&, seastar::task&){};
+    seastar::task* _coroutine_task{};
+    seastar::task* _waiting_task{};
+
+public:
+    explicit try_future_awaiter(seastar::future<T>&& f) noexcept : _future(std::move(f)) {}
+
+    try_future_awaiter(const try_future_awaiter&) = delete;
+    try_future_awaiter(try_future_awaiter&&) = delete;
+
+    bool await_ready() const noexcept {
+        // Will suspend+schedule for ready failed futures too.
+        return _future.available() && !_future.failed() && (!CheckPreempt || !need_preempt());
+    }
+
+    template<typename U>
+    void await_suspend(std::coroutine_handle<U> hndl) noexcept {
+        _resume_or_destroy = try_future_resume_or_destroy_coroutine<T, U>;
+        _coroutine_task = &hndl.promise();
+        _waiting_task = hndl.promise().waiting_task();
+
+        if (_future.available()) {
+            execute_involving_handle_destruction_in_await_suspend(*this);
+        } else {
+            _future.set_coroutine(*this);
+        }
+    }
+
+    T await_resume() {
+        if constexpr (std::is_void_v<T>) {
+            _future.get();
+        } else {
+            return std::move(_future).get();
+        }
+    }
+
+    void operator()() noexcept {
+        run_and_dispose();
+    }
+
+    virtual void run_and_dispose() noexcept override {
+        _resume_or_destroy(_future, *_coroutine_task);
+    }
+
+    virtual task* waiting_task() noexcept override {
+        return _waiting_task;
+    }
+};
+
+} // namespace seastar::internal
+
+namespace seastar::coroutine {
+
+/// \brief co_await:s a \ref future and returns the wrapped result if successful,
+/// terminates the coroutine otherwise, propagating the exception directly to the
+/// waiter.
+///
+/// If the future was successful, this is identical to co_await-ing the future
+/// directly. If the future failed, the coroutine is not resumed and instead the
+/// exception from the future is forwarded to the waiter directly and the
+/// coroutine is destroyed.
+///
+/// For example:
+/// ```
+/// // Function careful to not throw exceptions, instead returning failed futures.
+/// future<int> bar() {
+///    if (something_bad_happened) {
+///        return make_exception_future<>(std::runtime_error("error"));
+///    }
+///    return result;
+/// }
+///
+/// future<> foo() {
+///     auto result = co_await coroutine::try_future(bar());
+///     // This code is only executed if bar() returned a successful future.
+///     // Otherwise the exception is forwarded to the waiter future directly
+///     // and the coroutine is destroyed.
+///     check_result(result);
+/// }
+/// ```
+///
+/// Note that by default, `try_future` checks for if the task quota is depleted,
+/// which means that it will yield if the future is ready and \ref seastar::need_preempt()
+/// returns true.  Use \ref coroutine::try_future_without_preemption_check
+/// to disable preemption checking.
+template<typename T>
+class [[nodiscard]] try_future : public seastar::internal::try_future_awaiter<true, T> {
+public:
+    explicit try_future(seastar::future<T>&& f) noexcept
+        : seastar::internal::try_future_awaiter<true, T>(std::move(f))
+    {}
+};
+
+/// \brief co_await:s a \ref future, returns the wrapped result if successful,
+/// terminates the coroutine otherwise, propagating the exception to the waiter.
+///
+/// Same as \ref coroutine::try_future, but does not check for preemption.
+template<typename T>
+class [[nodiscard]] try_future_without_preemption_check : public seastar::internal::try_future_awaiter<false, T> {
+public:
+    explicit try_future_without_preemption_check(seastar::future<T>&& f) noexcept
+        : seastar::internal::try_future_awaiter<false, T>(std::move(f))
+    {}
+};
+
+} // namespace seastar::coroutine
--- a/tests/unit/coroutines_test.cc
+++ b/tests/unit/coroutines_test.cc
@@ -22,6 +22,7 @@
 #include <exception>
 #include <numeric>
 #include <ranges>
+#include <any>
 
 #include <seastar/core/circular_buffer.hh>
 #include <seastar/core/coroutine.hh>
@@ -34,8 +35,10 @@
 #include <seastar/coroutine/parallel_for_each.hh>
 #include <seastar/coroutine/as_future.hh>
 #include <seastar/coroutine/exception.hh>
+#include <seastar/coroutine/try_future.hh>
 #include <seastar/testing/random.hh>
 #include <seastar/testing/test_case.hh>
+#include <seastar/util/defer.hh>
 #include <seastar/util/later.hh>
 
 using seastar::broken_promise;
@@ -811,3 +814,134 @@ SEASTAR_TEST_CASE(test_lambda_value_capture_continuation) {
 
 
 #endif
+
+class test_exception : std::exception { };
+
+future<> return_ex_void() {
+    fmt::print("return_ex_void()\n");
+    return make_exception_future<>(test_exception{});
+}
+
+future<> return_void() {
+    fmt::print("return_void()\n");
+    co_await sleep(1ms);
+}
+
+future<int> return_ex_int() {
+    fmt::print("return_ex_int()\n");
+    return make_exception_future<int>(test_exception{});
+}
+
+future<int> return_int() {
+    fmt::print("return_int()\n");
+    co_await sleep(1ms);
+    co_return 128;
+}
+
+template <typename T>
+struct result_wrapper {
+    T value;
+    explicit result_wrapper(T v) : value(v) {}
+};
+
+template <>
+struct result_wrapper<seastar::internal::monostate> {
+};
+
+template <bool CheckPreempt, std::invocable<> F>
+// Use result_wrapper to create a mismatch between the return type of
+// the coroutine and that of the underlying function, to ensure that
+// try_future handles this case correctly.
+future<result_wrapper<typename std::invoke_result_t<F>::value_type>>
+do_run_try_future_test(F underlying_func, int& ctor_dtor_counter, bool& run_past) {
+    auto func = [&] () -> future<result_wrapper<typename std::invoke_result_t<F>::value_type>> {
+        counter_ref c1{ctor_dtor_counter};
+        counter_ref c2{ctor_dtor_counter};
+
+        BOOST_REQUIRE_GT(ctor_dtor_counter, 0);
+
+        using return_future_type = std::invoke_result_t<F>;
+        using return_type = typename return_future_type::value_type;
+        constexpr bool is_void = std::is_same_v<return_future_type, future<>>;
+
+        std::any ret;
+
+        try {
+            if constexpr (is_void) {
+                if constexpr (CheckPreempt) {
+                    co_await seastar::coroutine::try_future(underlying_func());
+                } else {
+                    co_await seastar::coroutine::try_future_without_preemption_check(underlying_func());
+                }
+            } else {
+                if constexpr (CheckPreempt) {
+                    ret = co_await seastar::coroutine::try_future(underlying_func());
+                } else {
+                    ret = co_await seastar::coroutine::try_future_without_preemption_check(underlying_func());
+                }
+            }
+            run_past = true;
+        } catch (...) {
+            BOOST_FAIL(fmt::format("Exception should be handled in try_future, bug caught: {}", std::current_exception()));
+        }
+
+        if constexpr (is_void) {
+            co_return result_wrapper<seastar::internal::monostate>{};
+        } else {
+            co_return result_wrapper(std::any_cast<return_type>(ret));
+        }
+    };
+
+    return seastar::do_with(func, [] (auto& func) {
+        const auto cxx_exception_before = seastar::engine().cxx_exceptions();
+        return func().then_wrapped([cxx_exception_before] (auto&& fut) {
+            const auto cxx_exception_after = seastar::engine().cxx_exceptions();
+            BOOST_REQUIRE_EQUAL(cxx_exception_before, cxx_exception_after);
+            return std::move(fut);
+        });
+    });
+}
+
+template <bool CheckPreempt, std::invocable<> F>
+future<> run_try_future_test(F underlying_func, std::optional<std::any> expected_value, std::source_location sl = std::source_location::current()) {
+    fmt::print("running test case at {}:{}\n", sl.file_name(), sl.line());
+
+    int ctor_dtor_counter{0};
+    bool run_past{false};
+
+    const bool throws = !expected_value.has_value();
+
+    using return_future_type = std::invoke_result_t<F>;
+    using return_type = typename return_future_type::value_type;
+    constexpr bool is_void = std::is_same_v<return_future_type, future<>>;
+
+    try {
+        if constexpr (is_void) {
+            co_await do_run_try_future_test<CheckPreempt>(std::move(underlying_func), ctor_dtor_counter, run_past);
+            BOOST_REQUIRE(expected_value);
+        } else {
+            auto res = co_await do_run_try_future_test<CheckPreempt>(std::move(underlying_func), ctor_dtor_counter, run_past);
+            BOOST_REQUIRE(expected_value);
+            BOOST_REQUIRE_EQUAL(res.value, std::any_cast<return_type>(*expected_value));
+        }
+    } catch (test_exception&) {
+        BOOST_REQUIRE(throws);
+    } catch (...) {
+        BOOST_FAIL(fmt::format("Unexpected exception {}", std::current_exception()));
+    }
+
+    BOOST_REQUIRE_EQUAL(run_past, !throws);
+    BOOST_REQUIRE_EQUAL(ctor_dtor_counter, 0);
+}
+
+SEASTAR_TEST_CASE(test_try_future) {
+    co_await run_try_future_test<true>(return_void, std::any{});
+    co_await run_try_future_test<false>(return_void, std::any{});
+    co_await run_try_future_test<true>(return_ex_void, std::nullopt);
+    co_await run_try_future_test<false>(return_ex_void, std::nullopt);
+
+    co_await run_try_future_test<true>(return_int, 128);
+    co_await run_try_future_test<false>(return_int, 128);
+    co_await run_try_future_test<true>(return_ex_int, std::nullopt);
+    co_await run_try_future_test<false>(return_ex_int, std::nullopt);
+}
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -550,24 +550,35 @@ seastar::future<> exception_handling() {
 Both `throw` and `std::rethrow_exception()` involve managing stack unwinding and exception objects, potentially
 impacting performance. Additionally, the C++ standard permits `std::rethrow_exception()` to create a copy of the
 exception object, introducing further overhead. Fortunately, in certain cases, exceptions can also be propagated
-directly, without throwing or rethrowing them. It can be achieved by returning a `coroutine::exception` wrapper.
-This approach can be advantageous when aiming to minimize overhead associated with exception handling. But it only
-works for coroutines which return `future<T>`, not `future<>`, due to the limitations in compilers. In particular,
-the example above won't compile if the return type is changed to `future<>`.
+directly, without throwing or rethrowing them. There are multiple facilities for this.
+`coroutine::try_future` can be used to propagate exceptions automatically from a called asynchronous function.
+It is analagous to rust's [try operator](https://google.github.io/comprehensive-rust/error-handling/try.html).
+If the awaited function resolves with an exception, it is automatically propagated to the coroutine's waiter
+without throwing. If manual error handling is needed, one can use `coroutine::as_future` and `coroutine::exception`.
+`coroutine::as_future` is analagous to `future<>::then_wrapped()`. It returns a ready future, allowing the code to
+probe for exception without expensive throw and catch. The exception can then be propagated directly using
+the `coroutine::exception` wrapper. `coroutine::exception` only works for coroutines which return `future<T>`, not
+`future<>`, due to the limitations in compilers. In particular, the example below won't compile if the return type is
+changed to `future<>`.
 
 Example:
 
 ```cpp
 seastar::future<int> exception_propagating() {
-    std::exception_ptr eptr;
-    try {
-        co_await function_returning_an_exceptional_future();
-    } catch (...) {
-        eptr = std::current_exception();
+    // Will automatically propagate the exception, if there is one.
+    auto result = co_await seastar::coroutine::try_future(
+        function_returning_an_exceptional_future()
+    );
+
+    // Will extract the ready future. Will not result in throwing, even if the future is exceptional.
+    auto fut = co_await seastar::coroutine::as_future(
+        function_returning_an_exceptional_future()
+    );
+    if (fut.failed()) {
+        // Saved exception pointer can be propagated without rethrowing
+        co_return seastar::coroutine::exception(fut.get_exception());
     }
-    if (eptr) {
-        co_return seastar::coroutine::exception(eptr); // Saved exception pointer can be propagated without rethrowing
-    }
+
     co_return seastar::coroutine::make_exception(3); // Custom exceptions can be propagated without throwing
 }
 ```


nyh:
This sounds like it may be a good idea, but I'm having a hard time imagining in what situations I will use this new function. In the commit message, you said that: "This allows for elegant control-flow, without the if spamming that is required whe using coroutine::as_future().". Can you please give a concrete example to back this statement up? Perhaps some real code from Seastar or ScyllaDB or another Seastar project, which looks "spammy" because of its use of as_future() but will look better with the new try_future() feature? Thanks.

avikivity:
@denesb the best way to answer @nyh is with a tutorial update

include/seastar/core/coroutine.hh
```diff
@@ -41,6 +41,7 @@ class coroutine_traits_base {
 public:
     class promise_type final : public seastar::task {
         seastar::promise<T> _promise;
+        noncopyable_function<std::exception_ptr()> _get_exception;
```
avikivity: Not happy about this. Let's see if it can be avoided.

tests/unit/coroutines_test.cc
```diff
@@ -1018,3 +1020,121 @@ SEASTAR_TEST_CASE(test_lambda_coroutine_in_continuation) {
...
+public:
+    explicit dummy(int& c) : _c(c) { ++_c; }
+    dummy(const dummy& o) : _c(o._c) { ++_c; }
+    dummy(dummy&&) = delete;
+    ~dummy() { --_c; }
+};
+
+using try_future_exception_handler_fn = seastar::noncopyable_function<std::exception_ptr(std::exception_ptr)>;
+
+auto do_run_try_future_test(auto f, try_future_exception_handler_fn ex_handler, int& ctor_dtor_counter, bool& run_past)
```
avikivity: Constraint? Guessing std::invocable<>?

Also give it a better name, like return_failed_future.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,135 @@
...
+    try_future_exception_handler_fn _ex_handler;
+
+public:
+    explicit try_future_awaiter(seastar::future<T>&& f, try_future_exception_handler_fn ex_handler) noexcept : _future(std::move(f)), _ex_handler(std::move(ex_handler)) {}
+
+    try_future_awaiter(const try_future_awaiter&) = delete;
+    try_future_awaiter(try_future_awaiter&&) = delete;
+
+    bool await_ready() const noexcept {
+        return _future.available() && (!CheckPreempt || !need_preempt());
```
avikivity: Don't you need to check for failure here? Otherwise `co_await make_exception_future() ` will never invoke the machinery.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,135 @@
...
+/// from the future is forwarded to the waiter directly.
+/// The caller can provide an exception handler function: in this case the
+/// exception is passed to the handler and the return value of this handler is
+/// used as the exception.
+/// This is similar to future<>::handle_exception()
+///
+/// For example:
+/// ```
+/// static future<> did_future_fail(future<int> fut) {
+///     auto value = co_await coroutine::try_future(std::move(fut));
```
avikivity: More realistic example is calling a function that can perhaps fail.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,135 @@
...
+///
+/// Note that by default, `try_future` checks for if the task quota is depleted,
+/// which means that it will yield if the future is ready and \ref seastar::need_preempt()
+/// returns true.  Use \ref coroutine::as_future_without_preemption_check
+/// to disable preemption checking.
+template<typename T = void>
+class [[nodiscard]] try_future : public seastar::internal::try_future_awaiter<true, T> {
+public:
+    explicit try_future(seastar::future<T>&& f, seastar::internal::try_future_exception_handler_fn ex_handler = {}) noexcept
+        : seastar::internal::try_future_awaiter<true, T>(std::move(f), std::move(ex_handler))
```
avikivity: What is this ex_handler?

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,135 @@
...
+            }
+            try {
+                if (_ex_handler) {
+                    return _ex_handler(_future.get_exception());
+                }
+                return _future.get_exception();
+            } catch (...) {
+                return std::current_exception();
+            }
+        });
```
avikivity: Perhaps a way to avoid the footprint increase of `promise` is to make the awaiter itself a seastar::task, schedule it self instead of the original coroutine (or call _future.set_coroutine on itself), inspect the received future, and decide whether to resume the coroutine or destroy it based on the result. Not at all sure it can work, this is really difficult.

tests/unit/coroutines_test.cc
```diff
@@ -1018,3 +1020,121 @@ SEASTAR_TEST_CASE(test_lambda_coroutine_in_continuation) {
...
+
+future<int> throw_int() {
+    co_await sleep(1ms);
+    throw test_exception{};
+}
+
+future<int> return_int() {
+    co_await sleep(1ms);
+    co_return 128;
+}
```
avikivity, line 839: Need to check immediately-throwing functions.

denesb:
> This sounds like it may be a good idea, but I'm having a hard time imagining in what situations I will use this new function. In the commit message, you said that: "This allows for elegant control-flow, without the if spamming that is required whe using coroutine::as_future().". Can you please give a concrete example to back this statement up? Perhaps some real code from Seastar or ScyllaDB or another Seastar project, which looks "spammy" because of its use of as_future() but will look better with the new try_future() feature? Thanks.

This `try_future` was suggested on the review for my ScyllaDB PR https://github.com/scylladb/scylladb/pull/25068, where I am trying to eliminate exception throwing on the read path. This resulted in having to sprinkle code like this in many functions:
```
auto fut = coroutine::as_future(foo());
if (fut.failed()) {
    co_return coroutine::return_exception_ptr(f.get_exception());
}
```
Besides being ugly and sometimes requiring restructuring of the control flow, to make the early return possible, `coroutine::return_exception_ptr()` also doesn't work when the current coroutine returns `future<>`, so in some cases I even had to add a dummy return type just to make this ugly code works.
With try future, this would be much more elegant, I could just replace `co_await foo()` with `co_await try_future(foo())` and I would get seamless throw-less exception propagation.

tests/unit/coroutines_test.cc
denesb: Will add template and constraint.

include/seastar/coroutine/try_future.hh
denesb: Yes, the flip side is that we will need to force a suspend here, even though the future is ready. But this is the error path, so we can afford it, it is still cheaper than throwing the exception.

include/seastar/coroutine/try_future.hh
denesb: I wanted the caller to pass an exception handler, similar to `future<>::handle_exception()`.
The handler can return the same exception pointer or create a new one (e.g. with additional information added).

tests/unit/coroutines_test.cc
denesb: Those won't work, they will bypass the awaiter completely.
But anyway the goal of `try_future<>` is to help propagate exceptions where the called function already is careful to not throw them. We cannot prevent somebody from throwing in the stack below.

denesb:
New in v2:
* Instead of installing hooks to the coroutine's promise, schedule the awaiter as a task and take care of resume/destroy of the coroutine in the awaiter.
* test: added template args with constraints as well as better names instead of `auto f`
* test: add test cases with functions returning `make_exception_future<>()`.

If this new approach looks good, I will tend to the documentation too.

include/seastar/coroutine/try_future.hh
denesb: Implemented this approach in v2. It seems to work fine.

include/seastar/coroutine/try_future.hh
denesb: Done in v2.

tests/unit/coroutines_test.cc
denesb: Done in v2.

include/seastar/core/coroutine.hh
denesb: Reverted in v2.

include/seastar/coroutine/try_future.hh
avikivity: It's best done via another API.

This one should have a single way of operating, await the future, return the value if successful, return from the coroutine with the error if not.

include/seastar/coroutine/try_future.hh
avikivity: I'm impressed, I was worried I was sending you on a wild goose chase here.

tests/unit/coroutines_test.cc
avikivity: I meant immediately-returning-exception-future functions, sorry for being imprecise.

Like try_with_gate (this is not an endorsement of seastar::gate).

avikivity:
So much red

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,166 @@
...
+namespace seastar {
+
+namespace internal {
+
+using try_future_exception_handler_fn = noncopyable_function<std::exception_ptr(std::exception_ptr)>;
+
+struct coroutine_handle {
+    virtual void set_value() noexcept = 0;
+    virtual void destroy() noexcept = 0;
+};
```
avikivity: I need some words (in a comment) about what the role of this is. It's hard to understand especially as there already exists std::coroutine_handle.

avikivity: A virtual destructor seems to be missing. It may be okay, but need a commend and make the non-virtual destructor protected.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,166 @@
...
+
+struct polymorphic_coroutine_handle : public coroutine_handle {
+};
```
avikivity: What does this do?

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,166 @@
...
+
+template <bool CheckPreempt, typename T>
+class [[nodiscard]] try_future_awaiter : public seastar::task {
+    seastar::future<T> _future;
+    try_future_exception_handler_fn _ex_handler;
+    noncopyable_function<void(seastar::future<T>&)> _resume_or_destroy;
```
avikivity: I'm guessing an owning function isn't needed here, since the calling coroutine will embed try_future_awaiter into its coroutine frame for the duration.

A function pointer, or a pointer into some interface should be sufficient.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,166 @@
...
+    void resume_or_destroy(std::coroutine_handle<U> hndl) {
+        if (!_future.failed()) {
+            hndl.resume();
+            return;
+        }
+
+        std::exception_ptr ex;
+        try {
+            if (_ex_handler) {
+                ex = _ex_handler(_future.get_exception());
```
avikivity: Again let's drop the exception handler for now (unless you see an immediate use for it).

avikivity: I see it's not really a handler but a transformer. I don't think it's allowed to return an empty exception_ptr, is it?

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,166 @@
...
+    try_future_awaiter(try_future_awaiter&&) = delete;
+
+    bool await_ready() const noexcept {
+        // Will suspend+schedule for ready failed futures too.
+        return _future.available() && !_future.failed() && (!CheckPreempt || !need_preempt());
+    }
+
+    template<typename U>
+    void await_suspend(std::coroutine_handle<U> hndl) noexcept {
+        _resume_or_destroy = std::bind(&try_future_awaiter<CheckPreempt, T>::resume_or_destroy<U>, this, hndl);
```
avikivity: You can type-erase resume_or_destroy<U> by adding a static function wrapper (also a template) that accepts a void*, casts the void* into a try_future_awaiter*, and calls resume_or_destroy<U>.

include/seastar/coroutine/try_future.hh
denesb: I ended up not using it, but forgot it here, sorry, will drop.

include/seastar/coroutine/try_future.hh
denesb: Nothing, see above. Will drop.

include/seastar/coroutine/try_future.hh
denesb: Yes, we only need to keep the coroutine handle around somehow. Wrapping it into the noncopyable_function was convenient.

denesb:
New in v3:
* Replace noncopyable_function with type-erased function pointer.
* Dropped exception handler.
* Drop unused garbage
* Fix broken CI (hopefully)
* Improve test:
    - test both CheckPreempt true and false
    - test futures that are available and failed (e.g. function returns make_exception_future<>())

Docs are still TODO

include/seastar/coroutine/try_future.hh
denesb: Dropped in v3.

include/seastar/coroutine/try_future.hh
denesb: Dropped in v3.

include/seastar/coroutine/try_future.hh
denesb: Replaced with function pointer in v3.

include/seastar/coroutine/try_future.hh
denesb: Dropped in v3.

include/seastar/coroutine/try_future.hh
denesb: Done in v3 (using free function).

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,149 @@
...
+ * under the License.
+ */
+/*
+ * Copyright (C) 2025-present ScyllaDB
+ */
+
+#pragma once
+
+#include <seastar/core/coroutine.hh>
+#include <seastar/util/noncopyable_function.hh>
```
avikivity: Unnecessary now.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,149 @@
...
+
+namespace seastar {
+
+namespace internal {
```
avikivity: You can write namespace seastar::internal {.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,149 @@
...
+
+template <typename T, typename U>
+void resume_or_destroy(seastar::future<T>& fut, task& coroutine_task) {
+    auto promise_ptr = reinterpret_cast<seastar::internal::coroutine_traits_base<T>::promise_type*>(&coroutine_task);
```
avikivity: static_cast should work too.

The function name is too generic. It's tied to the try_future thing and the name should reflect that.

tests/unit/coroutines_test.cc
denesb: I am already testing these, see `return_ex_void()` and `return_ex_int()`.

tests/unit/coroutines_test.cc
```diff
@@ -1018,3 +1020,138 @@ SEASTAR_TEST_CASE(test_lambda_coroutine_in_continuation) {
...
+                co_await seastar::coroutine::try_future_without_preemption_check(underlying_func());
+            }
+        } else {
+            if constexpr (CheckPreempt) {
+                ret = co_await seastar::coroutine::try_future(underlying_func());
+            } else {
+                ret = co_await seastar::coroutine::try_future_without_preemption_check(underlying_func());
+            }
+        }
+        run_past = true;
```
avikivity: Let's also verify that objects constructed in the coroutine were destroyed when co_await decided to return.

avikivity: Oh you have that!

denesb: I have `ctor_dtor_counter` for this, I bump the counter in the `dummy` constructors and decrement in the destructors. I have a check that it is > 0 after constructing the objects and that it is 0 after the coroutine returns (caller checks it).

tests/unit/coroutines_test.cc
```diff
@@ -1018,3 +1020,138 @@ SEASTAR_TEST_CASE(test_lambda_coroutine_in_continuation) {
...
+        if constexpr (is_void) {
+            co_await do_run_try_future_test<CheckPreempt>(std::move(underlying_func), ctor_dtor_counter, run_past);
+            BOOST_REQUIRE(expected_value);
+        } else {
+            auto value = co_await do_run_try_future_test<CheckPreempt>(std::move(underlying_func), ctor_dtor_counter, run_past);
+            BOOST_REQUIRE(expected_value);
+            BOOST_REQUIRE_EQUAL(value, std::any_cast<return_type>(*expected_value));
+        }
+    } catch (test_exception&) {
+        BOOST_REQUIRE(throws);
```
avikivity, line 928: Let's check the exception throwing counter was not incremented.

avikivity:
Looks good.

tests/unit/coroutines_test.cc
denesb: Yes, good idea. Will do.

include/seastar/coroutine/try_future.hh
denesb: Dropped in v3.

denesb:
New in v4:
* Fix do-comment, provide better example and better explanation
* Use seastar::internal and seastar::coroutine namespaces instead of embedding them
* Add exception counter check to the test
* Remove unused include of noncopyable_function
* Use static cast insead of reinterpret cast
* Renamed resume_or_destroy() to something more specific

tests/unit/coroutines_test.cc
denesb: Done in v4.

include/seastar/coroutine/try_future.hh
denesb: Done in v4.

include/seastar/coroutine/try_future.hh
denesb: Done in v4.

include/seastar/coroutine/try_future.hh
denesb: Removed in v4.

include/seastar/coroutine/try_future.hh
denesb: Changed in v4.

denesb:
> @denesb the best way to answer @nyh is with a tutorial update

You mean *the* seastar tutorial?

denesb:
I will now plug this into https://github.com/scylladb/scylladb/pull/25068 to see how it works in a large project.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,145 @@
...
+///    return result;
+/// }
+///
+/// future<> foo() {
+///     auto result = co_await coroutine::try_future(bar());
+///     // This code is only executed if bar() returned a successful future.
+///     // Otherwise the exception is forwarded to the waiter future directly
+///     // and the coroutine is destroyed.
+///     check_result(result);
+/// }
```
xemul, line 122: How would `foo()` look like if there were no `try_future()`? If I understand it correctly -- like this

```
future<> foo() {
    auto result = co_await bar();
    check_result(result);
}
```

and the result is the same -- if bar() resolves into exceptional future, the foo() is interrupted and results into exceptional future either, isn't it?

denesb: Yes, but in that case an exception will be thrown. Currently, if you want to avoid that, you have to do this:
```
auto fut = coroutine::as_future(bar());
if (fut.failed()) {
    co_await return_exception_ptr(fut.get_exception()); // doesn't even work with future<> return type, in that case one has to either throw or change future<> to future<T>
}
check_result(fut.get());
```

xemul: OK, thanks for clarification.
 Would you please add this "comparison" example somewhere? This doc-comment or tutorial.md maybe

denesb:
New in v5:
* Correctly handle the case where `T != U`: the type of the awaited future is not the same as that of the coroutine. Modify unit test to handle this case.

include/seastar/coroutine/try_future.hh
avikivity: Would be good to explain both bad alternatives.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,145 @@
...
+/// ```
+///
+/// Note that by default, `try_future` checks for if the task quota is depleted,
+/// which means that it will yield if the future is ready and \ref seastar::need_preempt()
+/// returns true.  Use \ref coroutine::try_future_without_preemption_check
+/// to disable preemption checking.
+template<typename T = void>
```
avikivity: I s  `= void`  necessary? I think not.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,145 @@
...
+    explicit try_future(seastar::future<T>&& f) noexcept
+        : seastar::internal::try_future_awaiter<true, T>(std::move(f))
+    {}
+};
+
+/// \brief co_await:s a \ref future, returns the wrapped result if successful,
+/// terminates the coroutine otherwise, propagating the exception to the waiter.
+///
+/// Same as \ref coroutine::try_future, but does not check for preemption.
+template<typename T = void>
```
avikivity: Here too

denesb:
Weird, a pre-existing test, testing an un-modified component started segfaulting, very consistently. The test is probably broken and somehow adding a new test to the end of the file makes it break all the time.

denesb:
We have a double-free in `exception_awaiter`: https://github.com/scylladb/seastar/blob/acd5720f5461283234a64298bb2f2196d364b6ac/include/seastar/coroutine/exception.hh#L44-L48
`hndl.destroy()` can only be called on a suspended coroutine, but here it is called on a not yet suspended one. As a result, the coroutine frame is freed twice. How was this not noticed before?

denesb:
This test is also crashing when run on master, so my branch has nothing to do with it.

avikivity:
How come? All pull requests and master regularly run the tests.

avikivity:
Running coroutines_test locally does not reproduce.

avikivity:
Randomly adding tests at the end did not make it reproduce.

denesb:
I don't understand either. FYI I'm using ScyllaDB's most recent toolchain and debug build. Doesn't reproduce in dev or release (although in CI it does reproduce with dev/release).

denesb:
I can reproduce with this command line on a freshly cleaned master (`git clean -xdff` and ` ccache -c`):
```
./configure.py --cflags='-Wfatal-errors -g' && ninja -C build/debug tests/unit/coroutines_test && ./build/debug/tests/unit/coroutines_test -- -c 2
```

denesb:
BTW the use-after-free looks like this:
```
==5777==ERROR: AddressSanitizer: heap-use-after-free on address 0x7c3011de7a42 at pc 0x00000048d583 bp 0x7b500f60ce40 sp 0x7b500f60ce38
READ of size 2 at 0x7c3011de7a42 thread T1
    #0 0x00000048d582 in operator() /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:473
    #1 0x000000511305 in check_coroutine_throws /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:466
    #2 0x00000050febd in check_coroutine_throws<std::runtime_error, test_coroutine_exception::run_test_case(_ZNK24test_coroutine_exception13run_test_caseEv.Frame*)::<lambda(int&)> > /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:454
    #3 0x000000495f3d in run_test_case /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:488
    #4 0x0000005f0e41 in std::__n4861::coroutine_handle<seastar::internal::coroutine_traits_base<void>::promise_type>::resume() const /usr/include/c++/15/coroutine:247
    #5 0x0000005d4c09 in seastar::internal::coroutine_traits_base<void>::promise_type::run_and_dispose() /home/bdenes/ScyllaDB/seastar/include/seastar/core/coroutine.hh:122
    #6 0x7f5014d19b51 in seastar::reactor::task_queue::run_tasks() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:2617
    #7 0x7f5014d22a3e in seastar::reactor::task_queue_group::run_tasks() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:3120
    #8 0x7f5014d21f1b in seastar::reactor::task_queue_group::run_some_tasks() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:3104
    #9 0x7f5014d27fce in seastar::reactor::do_run() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:3277
    #10 0x7f5014d240a9 in seastar::reactor::run() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:3165
    #11 0x7f501475e100 in seastar::app_template::run_deprecated(int, char**, std::function<void ()>&&) /home/bdenes/ScyllaDB/seastar/src/core/app-template.cc:273
    #12 0x7f501475b38d in seastar::app_template::run(int, char**, std::function<seastar::future<int> ()>&&) /home/bdenes/ScyllaDB/seastar/src/core/app-template.cc:167
    #13 0x7f5023e1ed23 in operator() /home/bdenes/ScyllaDB/seastar/src/testing/test_runner.cc:77
    #14 0x7f5023e2a355 in __invoke_impl<void, seastar::testing::test_runner::start_thread(int, char**)::<lambda()>&> /usr/include/c++/15/bits/invoke.h:63
    #15 0x7f5023e28681 in __invoke_r<void, seastar::testing::test_runner::start_thread(int, char**)::<lambda()>&> /usr/include/c++/15/bits/invoke.h:113
    #16 0x7f5023e2666f in _M_invoke /usr/include/c++/15/bits/std_function.h:292
    #17 0x7f50147a7582 in std::function<void ()>::operator()() const /usr/include/c++/15/bits/std_function.h:593
    #18 0x7f5014a7d4bd in seastar::posix_thread::start_routine(void*) /home/bdenes/ScyllaDB/seastar/src/core/posix.cc:90
    #19 0x7f502432fee5 in asan_thread_start(void*) (/lib64/libasan.so.8+0x28ee5) (BuildId: 10b8ccd49f75c21babf1d7abe51bb63589d8471f)
    #20 0x7f5013903f53 in start_thread (/lib64/libc.so.6+0x71f53) (BuildId: 48c4b9b1efb1df15da8e787f489128bf31893317)
    #21 0x7f501398732b in __clone3 (/lib64/libc.so.6+0xf532b) (BuildId: 48c4b9b1efb1df15da8e787f489128bf31893317)

0x7c3011de7a42 is located 98 bytes inside of 160-byte region [0x7c3011de79e0,0x7c3011de7a80)
freed by thread T1 here:
    #0 0x7f50243ef99b in operator delete(void*, unsigned long) (/lib64/libasan.so.8+0xe899b) (BuildId: 10b8ccd49f75c21babf1d7abe51bb63589d8471f)
    #1 0x00000048f7a7 in operator() /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:477
    #2 0x00000048fcd0 in operator() /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:473
    #3 0x000000645b0c in std::__n4861::coroutine_handle<seastar::internal::coroutine_traits_base<int>::promise_type>::destroy() const /usr/include/c++/15/coroutine:249
    #4 0x000000611945 in void seastar::internal::exception_awaiter::await_suspend<seastar::internal::coroutine_traits_base<int>::promise_type>(std::__n4861::coroutine_handle<seastar::internal::coroutine_traits_base<int>::promise_type>) /home/bdenes/ScyllaDB/seastar/include/seastar/coroutine/exception.hh:47
    #5 0x00000048e97d in operator() /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:475
    #6 0x00000048d44c in operator() /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:473
    #7 0x000000511305 in check_coroutine_throws /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:466
    #8 0x00000050febd in check_coroutine_throws<std::runtime_error, test_coroutine_exception::run_test_case(_ZNK24test_coroutine_exception13run_test_caseEv.Frame*)::<lambda(int&)> > /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:454
    #9 0x000000495f3d in run_test_case /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:488
    #10 0x0000005f0e41 in std::__n4861::coroutine_handle<seastar::internal::coroutine_traits_base<void>::promise_type>::resume() const /usr/include/c++/15/coroutine:247
    #11 0x0000005d4c09 in seastar::internal::coroutine_traits_base<void>::promise_type::run_and_dispose() /home/bdenes/ScyllaDB/seastar/include/seastar/core/coroutine.hh:122
    #12 0x7f5014d19b51 in seastar::reactor::task_queue::run_tasks() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:2617
    #13 0x7f5014d22a3e in seastar::reactor::task_queue_group::run_tasks() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:3120
    #14 0x7f5014d21f1b in seastar::reactor::task_queue_group::run_some_tasks() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:3104
    #15 0x7f5014d27fce in seastar::reactor::do_run() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:3277
    #16 0x7f5014d240a9 in seastar::reactor::run() /home/bdenes/ScyllaDB/seastar/src/core/reactor.cc:3165
    #17 0x7f501475e100 in seastar::app_template::run_deprecated(int, char**, std::function<void ()>&&) /home/bdenes/ScyllaDB/seastar/src/core/app-template.cc:273
    #18 0x7f501475b38d in seastar::app_template::run(int, char**, std::function<seastar::future<int> ()>&&) /home/bdenes/ScyllaDB/seastar/src/core/app-template.cc:167
    #19 0x7f5023e1ed23 in operator() /home/bdenes/ScyllaDB/seastar/src/testing/test_runner.cc:77
    #20 0x7f5023e2a355 in __invoke_impl<void, seastar::testing::test_runner::start_thread(int, char**)::<lambda()>&> /usr/include/c++/15/bits/invoke.h:63
    #21 0x7f5023e28681 in __invoke_r<void, seastar::testing::test_runner::start_thread(int, char**)::<lambda()>&> /usr/include/c++/15/bits/invoke.h:113
    #22 0x7f5023e2666f in _M_invoke /usr/include/c++/15/bits/std_function.h:292
    #23 0x7f50147a7582 in std::function<void ()>::operator()() const /usr/include/c++/15/bits/std_function.h:593
    #24 0x7f5014a7d4bd in seastar::posix_thread::start_routine(void*) /home/bdenes/ScyllaDB/seastar/src/core/posix.cc:90
    #25 0x7f502432fee5 in asan_thread_start(void*) (/lib64/libasan.so.8+0x28ee5) (BuildId: 10b8ccd49f75c21babf1d7abe51bb63589d8471f)
```

Notice how the use/free backtraces are identical up to `    #1 0x000000511305 in check_coroutine_throws /home/bdenes/ScyllaDB/seastar/tests/unit/coroutines_test.cc:466`. It seems to me that the coroutine frame is used immediately after we call destroy. I'm cloning `gcc.git` to dig into how `await_suspend()` is called.

avikivity:
Seems like co_await coroutine::exception is illegal, but can't see why.

avikivity:
Perhaps we need the same trick you used in try_future: make exception_awaiter a seastar::task, and have the task call hndl.destroy() instead of await_suspend().

Not based on real understanding, just guesses.

avikivity:
btw, it would be sad, since it's converting immediate execution to scheduled execution.

avikivity:
It's also possible this is just a gdb bug.

denesb:
According to my read of the gcc source code, a `auto y = co_await x` expression is translated roughly to:
```c++
    await_suspend(hndl);
    goto return_with_no_cleanup;
resume:
    auto y = await_resume();
    [...]
return_with_no_cleanup:
    // return without calling destructors
```
According to this, calling `hndl. destroy()` inside `await_suspend()` should not be a problem, the code is careful to not use the handle or the frame after `await_suspend()` resumes. This is when `await_suspend` returns `void`, the other cases (`bool` and `std::coroutine_handle<Z>` return type) are more involved.

denesb:
> Perhaps we need the same trick you used in try_future: make exception_awaiter a seastar::task, and have the task call hndl.destroy() instead of await_suspend().
> 
> Not based on real understanding, just guesses.

Actually, this crash is the reason I chose this route for `try_future`, my first version did call `hndl.destroy()` from `await_suspend()`, just like `exception_awaiter` does. After observing this crash, I concluded that this is not legal and pivoted to scheduling a task and destroying the coroutine from `run_and_dispose()`.
Now I'm confused, apparently this worked just fine up to now, and with certain compilers it still does.

denesb:
One more observation: the crash only happens if there was a previous `co_await`, prior to the `co_await exception_awaiter()` one. It doesn't matter what the prior `co_await` was, it reproduces even with a simple `co_await sleep()`.
Strangely, the extra (prior) `co_await` is needed in the *parent* coroutine, i.e. the one which calls the coroutine which actually uses `exception_awaiter`. I will try to produce a minimal reproducer.

avikivity:
Here's a reproducer:

https://godbolt.org/z/xxEv5n9hc

avikivity:
In the reproducer, there is no previous co_await.

avikivity:
It looks like a gcc regression. In 15.1, it only detects an (expected) memory leak. Or it could be that gcc 15.2 tightened the implementation. I'll file a gcc bug.

denesb:
Minimal reproducer:

[exp.cpp.txt](https://github.com/user-attachments/files/22360641/exp.cpp.txt)

```
g++ -Wl,-rpath=/home/bdenes/ScyllaDB/seastar/build/debug/ $(pkg-config --libs --cflags /home/bdenes/ScyllaDB/seastar/build/debug/seastar.pc) -g exp.cpp -o exp
```

[out.txt](https://github.com/user-attachments/files/22360664/out.txt)

avikivity:
If this is acknowledged as a gcc bug, we can #ifdef this facility away from bad compilers.

If it's really illegal, we'll have to think.

avikivity:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=121961

avikivity:
Guessing gcc-mirror/gcc@b4da8ee3e613502b50501773f5ecb494d3a51c8b, but will bisect.

avikivity:
> Guessing [gcc-mirror/gcc@b4da8ee](https://github.com/gcc-mirror/gcc/commit/b4da8ee3e613502b50501773f5ecb494d3a51c8b), but will bisect.

Confirmed

denesb:
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=121961

This is still unassigned after more than 2 months. How do we unblock this PR?
The new `try_future<>()` takes the conservative approach, which doesn't hit the new gcc bug, it is the pre-existing `exception_awaiter` which triggers it and fails CI.

avikivity:
> > https://gcc.gnu.org/bugzilla/show_bug.cgi?id=121961
> 
> This is still unassigned after more than 2 months. How do we unblock this PR? The new `try_future<>()` takes the conservative approach, which doesn't hit the new gcc bug, it is the pre-existing `exception_awaiter` which triggers it and fails CI.

https://github.com/scylladb/seastar/pull/3106

(filed before seeing your comment!)

denesb:
Thanks, I will rebase and send a new version soon.

denesb:
Rebased

denesb:
CI is failing in DNS test. Related to recent DNS changes? @xemul

denesb:
I think https://github.com/scylladb/seastar/pull/3113 should fix the CI.

xemul:
Reverse lookup of seastar.io's address stopped working few days ago. And yes, #3113 fixes it

denesb:
Rebased again.

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,145 @@
...
+    T await_resume() {
+        if constexpr (std::is_void_v<T>) {
+            _future.get();
+        } else {
+            return std::move(_future).get();
+        }
+    }
+
+    virtual void run_and_dispose() noexcept override {
+        _resume_or_destroy(_future, *_coroutine_task);
```
avikivity, line 85: Why do you need _resume_or_destroy to be a function pointer? You set it in await_suspend, and I'm guessing it doesn't change later. Why not call it directly?

include/seastar/coroutine/try_future.hh
avikivity: Still think so.

tests/unit/coroutines_test.cc
```diff
@@ -811,3 +814,130 @@ SEASTAR_TEST_CASE(test_lambda_value_capture_continuation) {
...
+// the coroutine and that of the underlying function, to ensure that
+// try_future handles this case correctly.
+future<result_wrapper<typename std::invoke_result_t<F>::value_type>>
+do_run_try_future_test(F underlying_func, int& ctor_dtor_counter, bool& run_past) {
+    const auto check_cxx_exceptions_on_exit = seastar::defer([cxx_exception_before = seastar::engine().cxx_exceptions()] () noexcept {
+        if (seastar::engine().cxx_exceptions() != cxx_exception_before) {
+            // We are in a destructor, cannot throw
+            std::abort();
+        }
+    });
```
avikivity: You could convert it to a BOOST_REQUIRE_EQUAL with a `with_cxx_exceptions_count_check(std::function<std::future<> ()>)` thing.

avikivity:
A tutorial entry is still needed. 

Can also mention prior art: https://google.github.io/comprehensive-rust/error-handling/try.html

avikivity:
Also need to use the facility from b1a7ed56183a6f414bd8f654394b31f89245bda3 for gcc 15.2+.

include/seastar/coroutine/try_future.hh
denesb: `await_suspend()` is a `template<typename U>`, but `run_and_dispose()` is not. I need to remember `U` and I do it via the function pointer: although the arguments of the function are up-casted to common ancestor, in the body I down-cast back to `U`.

denesb:
> Also need to use the facility from [b1a7ed5](https://github.com/scylladb/seastar/commit/b1a7ed56183a6f414bd8f654394b31f89245bda3) for gcc 15.2+.

I don't need it. Unlike `exception`, which is always ready and uses `await_suspend()` just to destroy the coroutine, `try_future` has to wait for a non-ready future, so it has to defer anyway.

tests/unit/coroutines_test.cc
denesb: I'm not sure I follow.
I want this check to be done before leaving scope, so I do have to use `defer()` here.

denesb:
New in v6:
* Added docs/tutorial.md update

include/seastar/coroutine/try_future.hh
avikivity: Yes. Thanks.

I don't suggest doing this now, but I think there's a workaround: define `coroutine_traits_base::promise_type::await_transform<try_future_awaiter<T>>()`, this allows mixing in the corutine's concrete type into the awaiter.

tests/unit/coroutines_test.cc
avikivity:
```c++
do_run_try_future_test(F underlying_func...) {
    co_await  with_cxx_exceptions_count_check([...] {
           ...
    });
}
```

with_cxx_exceptions_count_check samples the exception counter, runs the given functions, awaits it, samples the exception counter again, and BOOST_CHECK_EQUALs.

denesb:
New in v7:
* Use BOOST_REQUIRE_EQUALS to check cxx exceptions before/after

tests/unit/coroutines_test.cc
denesb: Done in v7.

avikivity:
> > Also need to use the facility from [b1a7ed5](https://github.com/scylladb/seastar/commit/b1a7ed56183a6f414bd8f654394b31f89245bda3) for gcc 15.2+.
> 
> I don't need it. Unlike `exception`, which is always ready and uses `await_suspend()` just to destroy the coroutine, `try_future` has to wait for a non-ready future, so it has to defer anyway.

In

```c++
    template<typename U>
    void await_suspend(std::coroutine_handle<U> hndl) noexcept {
        _resume_or_destroy = try_future_resume_or_destroy_coroutine<T, U>;
        _coroutine_task = &hndl.promise();
        _waiting_task = hndl.promise().waiting_task();

        if (!_future.available()) {
            _future.set_coroutine(*this);
        } else {
            schedule(this);
        }
    }
```


When you call schedule(), the future is known to be available. So you could destroy the coroutine then and there.

Calling schedule is a pessimization even in the case where you return a non-failed future. I think we should have three cases:

1. non-ready future: schedule ourselves to handle it
2. ready non-failed future: return `false` to resume immediately (instead of scheduling)
3. ready failed future: use the execute_thing to either destroy the handle immediately or schedule its destruction depending on the compiler

Cases 1, 3 return true so we don't auto-resume.

denesb:
New in v7:
* Use `execute_involving_handle_destruction_in_await_suspend()` in the ready-future case. Massage it to not allocate when `func` is a task already.

denesb:
New in v7.1:
* Remove `= void` from `template <typename T = void>`.

include/seastar/coroutine/try_future.hh
denesb: Dropped in v7.1

include/seastar/coroutine/try_future.hh
denesb: Dropped in v7.1

include/seastar/coroutine/try_future.hh
```diff
@@ -0,0 +1,149 @@
...
+        if constexpr (std::is_void_v<T>) {
+            _future.get();
+        } else {
+            return std::move(_future).get();
+        }
+    }
+
+    void operator()() noexcept {
+        run_and_dispose();
+    }
```
avikivity, line 82: This is a little icky. Better to have a separate overload of execute_involving_handle_destruction_in_await_suspend() for tasks. But can clear up afterwards.

denesb: I was trying very hard to not have two or more copies of the `#ifdef` with the version sniff, fearing that they would get out-of-sync.
This type is not accessible to anyone but the coroutine mechanism, so I figured the risk of someone calling `operator()()` by accident is close to 0.

PR #2975, opened 2025-09-11
io: Count and export number of AIO retries

xemul:
The AIO requests may come via two places -- the reactor itself polling
for IO and thread-pool that retries requests that had been rejected by
the kernel on the first try.

This patch counts the latter events and exports them as metrics. The
former can be calculated using existing aio-reads and aio-writes.

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -306,6 +306,7 @@ bool aio_storage_context::reap_completions(bool allow_retry)
         auto iocb = get_iocb(_ev_buffer[i]);
         if (_ev_buffer[i].res == -EAGAIN && allow_retry) {
             set_nowait(*iocb, false);
+            _r._io_stats.aio_retries++;
             _pending_aio_retry.push_back(iocb);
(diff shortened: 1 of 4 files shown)


src/core/reactor_backend.cc
```diff
@@ -232,6 +232,7 @@ aio_storage_context::submit_work() {
     }
 
     size_t nr_consumed = 0;
+    _r._io_stats.io_submits_in_reactor += to_submit;
```
avikivity: Isn't this already counted in aio_reads+aio_writes?

xemul: ...+fdatasyncs, but we don't count it

xemul:
upd:
- count only retries and export them under "aio_retries" name

xemul:
@avikivity , please re-review

xemul:
@avikivity , please re-review

travisdowns:
This patch would be very nice!

xemul:
@avikivity , please re-review

PR #2977, opened 2025-09-12
co_return to accept same expressions and types as return

bashtanov:
Current implementation of `co_return` behaves differently from plain
`return` when interpreting the expression returned. This includes both
- `co_return` accepting expressions `return` would reject;
- `co_return` rejecting expressions `return` would accept.

Unexpectedly accepted expressions allow bugs from accidental misuse:
- a value of a strongly-defined type may be implicitly instantiated from
its underlying type thus defeating the purpose, e.g. a coroutine
declared to return `future<std::chrono::seconds>`, may `co_return 42;`
- instead of a value of an enum type used in the future, a value of an
unrelated enum type may be used if it uses the same underlying type;
- obscure convertions, such as a number to `std::vector`, may be applied
implicitly.

Unexpectedly rejected expressions include:
- those rejected due to false ambiguity, where compiler cannot decide
between a reasonable interpretation and one of the situtaions described
above;
- brace-enclosed initializer list without type specified.

This happens because we save the `co_return`'ed value in a state by:
- calling a chain of unnecessarily templated functions;
- implicitly calling the type ctor even if it is marked explicit.

The fix is to avoid the above.

One of the test cases was shamelessly stolen from https://github.com/scylladb/seastar/pull/2971 as my change fixes their use case.

### DIFF
--- a/doc/compatibility.md
+++ b/doc/compatibility.md
@@ -78,6 +78,10 @@ versions of the API. For example.
      and become a move-only type
    - Seastar_API_LEVEL=9 defines the data_sink_impl::put(span<temporary_buffer>)
      as the new and only method to be implemented
+   - Seastar_API_LEVEL=10 makes co_return and promise.set_value() semantics
+     closer to those of plain C++ return regarding how they convert
+     expressions used as their arguments to the declared type.
+
 
 Applications can use an old API_LEVEL during a transition
 period, fix their code, and move to the new API_LEVEL.
@@ -112,15 +116,15 @@ API Level History
 
 |Level|Introduced |Mandatory|Description                                   |
 |:---:|:---------:|:-------:| -------------------------------------------- |
-| 2   |  2019-07  | 2020-04 | Non-variadic futures in socket::accept()     |
-| 3   |  2020-05  | 2023-03 | make_file_data_sink() closes file and returns a future<>  |
-| 4   |  2020-06  | 2023-03 | Non-variadic futures in when_all_succeed()   |
-| 5   |  2020-08  | 2023-03 | future::get() returns std::monostate() instead of void |
-| 6   |  2020-09  | 2023-03 | future<T> instead of future<T...>            |
-| 7   |  2023-05  | 2024-09 | unified CPU/IO scheduling groups             |
-| 8   |  2025-08  |         | noncopyable function in json_return_type     |
-| 9   |  2025-08  |         | data_sink_impl new API                       |
-
+|  2  |  2019-07  | 2020-04 | Non-variadic futures in socket::accept()     |
+|  3  |  2020-05  | 2023-03 | make_file_data_sink() closes file and returns a future<>  |
+|  4  |  2020-06  | 2023-03 | Non-variadic futures in when_all_succeed()   |
+|  5  |  2020-08  | 2023-03 | future::get() returns std::monostate() instead of void |
+|  6  |  2020-09  | 2023-03 | future<T> instead of future<T...>            |
+|  7  |  2023-05  | 2024-09 | unified CPU/IO scheduling groups             |
+|  8  |  2025-08  |         | noncopyable function in json_return_type     |
+|  9  |  2025-08  |         | data_sink_impl new API                       |
+| 10  |  2026-04  |         | co_return and set_value strict type semantics |
 
 Note: The "mandatory" column indicates when backwards compatibility
 support for the API preceding the new level was removed.
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -95,9 +95,14 @@ public:
         promise_type(promise_type&&) = delete;
         promise_type(const promise_type&) = delete;
 
-        template<typename... U>
-        void return_value(U&&... value) {
-            _promise.set_value(std::forward<U>(value)...);
+        template<typename U>
+        void return_value(U&& value) {
+            _promise.set_value(std::forward<U>(value));
+        }
+
+        [[deprecated("Forwarding coroutine returns are deprecated as too dangerous. Use 'co_return co_await ...' until explicit syntax is available.")]]
+        void return_value(future<T>&& fut) noexcept {
+            fut.forward_to(std::move(_promise));
         }
 
         void return_value(coroutine::exception ce) noexcept {
@@ -106,11 +111,6 @@ public:
 
         void set_exception(std::exception_ptr&& eptr) noexcept {
             _promise.set_exception(std::move(eptr));
-        }
-
-        [[deprecated("Forwarding coroutine returns are deprecated as too dangerous. Use 'co_return co_await ...' until explicit syntax is available.")]]
-        void return_value(future<T>&& fut) noexcept {
-            fut.forward_to(std::move(_promise));
         }
 
         void unhandled_exception() noexcept {
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -95,6 +95,7 @@ public:
         promise_type(promise_type&&) = delete;
         promise_type(const promise_type&) = delete;
 
+#if SEASTAR_API_LEVEL < 10
         template<typename U>
         void return_value(U&& value) {
             _promise.set_value(std::forward<U>(value));
@@ -104,6 +105,18 @@ public:
         void return_value(future<T>&& fut) noexcept {
             fut.forward_to(std::move(_promise));
         }
+#else
+        // this non-templated version only exists to support co_returning a braced-init-list
+        void return_value(T&& value) noexcept {
+            _promise.set_value(std::forward<T>(value));
+        }
+
+        template<typename U>
+        requires std::is_convertible_v<U&&, T>
+        void return_value(U&& value) noexcept {
+            _promise.set_value(std::forward<U>(value));
+        }
+#endif
 
         void return_value(coroutine::exception ce) noexcept {
             _promise.set_exception(std::move(ce.eptr));
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -336,7 +336,7 @@ public:
     void uninitialized_set_loose_type_conversion(tuple_type&& v) {
         uninitialized_set(std::move(std::get<0>(v)));
     }
-    void uninitialized_set_legacy(const tuple_type& v) {
+    void uninitialized_set_loose_type_conversion(const tuple_type& v) {
         uninitialized_set(std::get<0>(v));
     }
     maybe_wrap_ref<T>& uninitialized_get() {
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -273,6 +273,13 @@ struct get0_return_type {
 template<typename T>
 using maybe_wrap_ref = std::conditional_t<std::is_reference_v<T>, std::reference_wrapper<std::remove_reference_t<T>>, T>;
 
+/// Convert U to T using copy-initialization (implicit conversions only).
+/// C++17 mandatory copy elision ensures no extra moves or copies.
+template<typename T, typename U>
+T implicit_convert(U&& u) {
+    return std::forward<U>(u);
+}
+
 /// \brief Wrapper for keeping uninitialized values of non default constructible types.
 ///
 /// This is similar to a std::optional<T>, but it doesn't know if it is holding a value or not, so the user is
@@ -290,16 +297,33 @@ struct uninitialized_wrapper {
 
 public:
     uninitialized_wrapper() noexcept = default;
+    template<typename U>
+    requires std::is_convertible_v<U&&, T>
+    void uninitialized_set(U&& u) {
+        // Despite the `requires` condition above, the line below may still fail compilation if T is a reference type.
+        // E.g. when T is a const lvalue reference and U is an rvalue reference to a non-const type.
+        // These kind of conversions make little sense both here and with plain `return`.
+        new(&_v.value) auto(implicit_convert<maybe_wrap_ref<T>>(std::forward<U>(u)));
+    }
     template<typename... U>
-    requires (!std::same_as<std::tuple<std::remove_cv_t<U>...>, std::tuple<tuple_type>>)
-    void
-    uninitialized_set(U&&... vs) {
+    void uninitialized_emplace(U&&... vs) {
         new (&_v.value) maybe_wrap_ref<T>(T(std::forward<U>(vs)...));
     }
-    void uninitialized_set(tuple_type&& v) {
+    void uninitialized_set_from_tuple(tuple_type&& v) {
+        uninitialized_set_loose_type_conversion(std::move(std::get<0>(v)));
+    }
+    void uninitialized_set_from_tuple(const tuple_type& v) {
+        uninitialized_set_loose_type_conversion(std::get<0>(v));
+    }
+    template<typename... U>
+    requires (!std::same_as<std::tuple<std::remove_cv_t<U>...>, std::tuple<tuple_type>>)
+    void uninitialized_set_loose_type_conversion(U&&... vs) {
+        new (&_v.value) maybe_wrap_ref<T>(T(std::forward<U>(vs)...));
+    }
+    void uninitialized_set_loose_type_conversion(tuple_type&& v) {
         uninitialized_set(std::move(std::get<0>(v)));
     }
-    void uninitialized_set(const tuple_type& v) {
+    void uninitialized_set_legacy(const tuple_type& v) {
         uninitialized_set(std::get<0>(v));
     }
     maybe_wrap_ref<T>& uninitialized_get() {
@@ -315,13 +339,23 @@ struct uninitialized_wrapper<internal::monostate> {
     [[no_unique_address]] internal::monostate _v;
 public:
     uninitialized_wrapper() noexcept = default;
-    void uninitialized_set() {
-    }
     void uninitialized_set(internal::monostate) {
     }
-    void uninitialized_set(std::tuple<>&& v) {
+    void uninitialized_emplace(internal::monostate) {
     }
-    void uninitialized_set(const std::tuple<>& v) {
+    void uninitialized_emplace() {
+    }
+    void uninitialized_set_from_tuple(std::tuple<>&& v) {
+    }
+    void uninitialized_set_from_tuple(const std::tuple<>& v) {
+    }
+    void uninitialized_set_loose_type_conversion() {
+    }
+    void uninitialized_set_loose_type_conversion(internal::monostate) {
+    }
+    void uninitialized_set_loose_type_conversion(std::tuple<>&& v) {
+    }
+    void uninitialized_set_loose_type_conversion(const std::tuple<>& v) {
     }
     internal::monostate& uninitialized_get() {
         return _v;
@@ -556,7 +590,11 @@ inline void future_state_base::any::check_failure() noexcept {
     }
 }
 
-struct ready_future_marker {};
+struct loose_type_conversion_ready_future_marker {};
+struct set_ready_future_marker {};
+struct set_from_tuple_ready_future_marker {};
+struct emplace_ready_future_marker {};
+
 struct exception_future_marker {};
 struct future_for_get_promise_marker {};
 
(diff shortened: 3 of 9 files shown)


include/seastar/core/future.hh
```diff
@@ -609,6 +620,10 @@ struct future_state :  public future_state_base, private internal::uninitialized
         new (this) future_state(current_exception_future_marker());
       }
     }
+    void set_strict(T v) {
+        assert(_u.st == state::future);
+        new (this) future_state(strict_type_ready_future_marker(), std::forward<T>(v));
+    }
```
avikivity, line 709: Please don't change future.hh as a side effect. This is the main header and changes here have to be deliberate.

include/seastar/core/future.hh
```diff
@@ -954,6 +978,16 @@ public:
...
 
+    /// \brief Sets the promises value using exactly declared type
+    ///
+    /// Forwards the argument and makes them available to the associated
+    /// future. May be called either before or after \c get_future().
+    ///
+    /// pr.set_value_strict(42);
+    void set_value_strict(value_type v) noexcept {
+        internal::promise_base_with_type<T>::set_value_strict(std::forward<value_type>(v));
+    }
```
avikivity, line 1084: Is set_value_strict needed? What's the difference compared to set_value(T(...))?

If we want set_value_strict, I think an API change is in order:
1. rename set_value() to emplace_value(Args&&...) to construct T in place
2. add set_value(T).

Unfortunately this is a breaking change and must be done by introducing a new API_LEVEL.

bashtanov: The difference is the same as with `co_return`: it does not perform any unexpected type conversions.
I'm not sure if we need it TBH, if `set_value` was called `emplace_value` I wouldn't touch it.

Do we need a new API_LEVEL only for functions like this that user call explicitly, or for co_return semantics change as well?

bashtanov: We do need `promise::set_value_strict`, whether renamed or not, for `coroutine_traits_base::return_value` to call it though. I can try to hide it from user if we decide to go without it.

include/seastar/core/future.hh
```diff
@@ -297,6 +297,9 @@ struct uninitialized_wrapper {
 
 public:
     uninitialized_wrapper() noexcept = default;
+    void uninitialized_set_strict(T v) {
+        new (&_v.value) auto(std::forward<T>(v));
```
bashtanov: I'll need to change this to support reference types

include/seastar/core/future.hh
avikivity: > Do we need a new API_LEVEL only for functions like this that user call explicitly, or for co_return semantics change as well?

I think we'll need it for co_return too as people likely rely on it.

include/seastar/core/future.hh
bashtanov: Do you mean specifically this private function, or any changes to the file?
Other than `promise::set_value_strict` that is being discussed below, my changes only affect `future_state` which is marked as internal, and classes in `seastar::internal` namespace.
In any case, if you can think of a way to move this logic out of this header I'd be really grateful for suggestions.

avikivity: I'm not saying to make the changes elsewhere, but to move them to their own commit with clear justification.

That holds triple if we rename set_value() to emplace_value() and introduce a new set_value().

avikivity:
If you update the series, kindly inform the reviewer about what has changed rather than having them reverse-engineer it.

bashtanov:
@avikivity sure, there we go:
- To avoid excessive moves, implementation changed:
   - the argument is passed via the function chain perfectly forwarded, but every time with a static check that it is implicitly convertible to the target type
   - the actual conversion happens when the value is stored by calling a return-its-argument function `implicit_convert`, copy/move elided thanks to C++17 rules
- As requested, split commits into preparations and user-facing header changes
- Added more tests, mostly to make sure no excessive moves or copies happen
- Rebased
- Bumped supported C++ versions to 23 and 26 in docs. I used `new (placement) auto (...)` syntax I thought was C++23-specific, but now I double checked and it seems that, unlike function-style cast `auto(foo)`, it's been supported since C++11. I'll remove it in the next review iteration if needed.

Hope it helps. If any code or commit will benefit from a more detailed comment I would appreciate if you could let me know.

include/seastar/core/coroutine.hh
```diff
@@ -95,9 +95,14 @@ public:
...
-        void return_value(U&&... value) {
-            _promise.set_value(std::forward<U>(value)...);
+        template<typename U>
+        void return_value(U&& value) {
+            _promise.set_value(std::forward<U>(value));
+        }
+
+        [[deprecated("Forwarding coroutine returns are deprecated as too dangerous. Use 'co_return co_await ...' until explicit syntax is available.")]]
+        void return_value(future<T>&& fut) noexcept {
+            fut.forward_to(std::move(_promise));
```
avikivity, line 105: Why this code movement?

include/seastar/core/future.hh
```diff
@@ -290,16 +297,33 @@ struct uninitialized_wrapper {
 
 public:
     uninitialized_wrapper() noexcept = default;
+    template<typename U>
+    requires std::is_convertible_v<U&&, T>
+    void uninitialized_set(U&& u) {
```
avikivity, line 302: Slightly better to use std::convertible_to from the concept library (an even say `uninialized_set(auto std::convertible_to<T>&& u)`

include/seastar/core/future.hh
```diff
@@ -290,16 +297,33 @@ struct uninitialized_wrapper {
...
+        uninitialized_set_legacy(std::move(std::get<0>(v)));
+    }
+    void uninitialized_set_from_tuple(const tuple_type& v) {
+        uninitialized_set_legacy(std::get<0>(v));
+    }
     template<typename... U>
     requires (!std::same_as<std::tuple<std::remove_cv_t<U>...>, std::tuple<tuple_type>>)
-    void
-    uninitialized_set(U&&... vs) {
+    void uninitialized_set_legacy(U&&... vs) {
```
avikivity, line 320: "legacy" doesn't mean much outside the point in time it is introduced.

Try to find a descriptive name (unless it's gone by the end of the series).

include/seastar/core/future.hh
```diff
@@ -290,16 +297,33 @@ struct uninitialized_wrapper {
 
 public:
     uninitialized_wrapper() noexcept = default;
+    template<typename U>
+    requires std::is_convertible_v<U&&, T>
+    void uninitialized_set(U&& u) {
+        // Despite the `requires` condition above, the line below may still fail compilation if T is a reference type.
+        // E.g. when T is a const lvalue reference and U is an rvalue reference to a non-const type.
```
avikivity, line 304: Do we even support future\<T&>?

include/seastar/core/future.hh
```diff
@@ -612,18 +650,59 @@ struct future_state :  public future_state_base, private internal::uninitialized
         move_it(std::move(x));
         return *this;
     }
+    template<typename U>
+    requires std::is_convertible_v<U&&, T>
+    future_state(set_ready_future_marker, U&& v) noexcept : future_state_base(state::result) {
+      try {
```
avikivity, line 656: We use 4-space indents.

include/seastar/core/future.hh
```diff
@@ -612,18 +650,59 @@ struct future_state :  public future_state_base, private internal::uninitialized
...
+        this->uninitialized_set(std::forward<U>(v));
+      } catch (...) {
+        new (this) future_state(current_exception_future_marker());
```
avikivity, line 659: I don't know if it's legal to double-construct *this.

include/seastar/core/future.hh
```diff
@@ -612,18 +650,59 @@ struct future_state :  public future_state_base, private internal::uninitialized
...
+    template <typename... A>
+    void emplace(A&&... a) noexcept {
         assert(_u.st == state::future);
-        new (this) future_state(ready_future_marker(), std::forward<A>(a)...);
+        new (this) future_state(emplace_ready_future_marker(), std::forward<A>(a)...);
+    }
+    template <typename... A>
+    void set_legacy(A&&... a) noexcept {
+        assert(_u.st == state::future);
+        new (this) future_state(legacy_ready_future_marker(), std::forward<A>(a)...);
```
avikivity, line 705: Again I don't like double construction.  Since this is from existing code I'll allow it here.

include/seastar/core/coroutine.hh
```diff
@@ -104,6 +105,18 @@ public:
...
             fut.forward_to(std::move(_promise));
         }
+#else
+        // this non-templated version only exists to support co_returning a braced-init-list
+        void return_value(T&& value) noexcept {
+            _promise.set_value(std::forward<T>(value));
+        }
+
+        template<typename U>
+        requires std::is_convertible_v<U&&, T>
```
avikivity, line 115: Again prefer std::convertible_to

tests/unit/coroutines_test.cc
```diff
@@ -946,3 +949,186 @@ SEASTAR_TEST_CASE(test_try_future) {
...
+}
+
+future<int> co_return_tup_int_clv() {
+    const std::tuple t(42);
+    co_return t;
+}
+
+SEASTAR_TEST_CASE(test_std_expected_void_specialization) {
+  BOOST_REQUIRE_EQUAL(co_await co_return_tup_int_rv(), 42);
+  BOOST_REQUIRE_EQUAL(co_await co_return_tup_int_clv(), 42);
```
avikivity, line 965: Again 4-space indents.

README.md
```diff
@@ -146,11 +146,11 @@ There are also instructions for building on any host that supports [Docker](doc/
 
 Use of the [DPDK](http://dpdk.org) is [optional](doc/building-dpdk.md).
 
-#### Seastar's C++ standard: C++20 or C++23
+#### Seastar's C++ standard: C++23 or C++26
```
avikivity, line 149: This doesn't belong in this series.

include/seastar/core/future.hh
bashtanov: 1) `std::convertible_to<From, To>` is semantically different: it checks that `From` can be converted to `To` _both_ implicitly and explicitly. Is there a pair of types convertible implicitly but not explicitly? [Yes!](https://stackoverflow.com/a/62644127) If a type is not return'able it shouldn't be co_return'able.
2) The concept syntactic sugar is still achievable if we create a concept based on `std::is_convertible_v`. However, the desired condition here is `std::is_convertible_v<U&&, T>`, not `std::is_convertible_v<U, T>`. Assuming `U&&` is well-formed, these conditions are actually equivalent, see proof below. And `U&&` is present as a function argument. Would you like me to implement such concept and use it here and in other places instead of `requires std::is_convertible_v<U&&, T>`?
----
Proof:
`std::is_convertible_v<From, To>` tests whether `To f() { return std::declval<From>(); }` is well-formed. And `std::declval<T>()` returns `std::add_rvalue_reference_t<T>`. So:
`is_convertible_v<U, T>` effectively tests `add_rvalue_reference_t<U> → T`
`is_convertible_v<U&&, T>` effectively tests `add_rvalue_reference_t<U&&> → T`
By reference collapsing, these produce identical types in every well-formed case.

include/seastar/core/future.hh
bashtanov: Any preference between `_loose`, `_lax`, `_relaxed`, `_compat`? My fav is `_lax`.

include/seastar/core/future.hh
bashtanov: [We do](https://github.com/scylladb/seastar/blob/master/include/seastar/core/future.hh#L287-L288)

include/seastar/core/future.hh
bashtanov: This is legal due to [[basic.life]/6](https://eel.is/c++draft/basic.life#6)

include/seastar/core/future.hh
avikivity: The main advantage of the library concept is that it's a library concept. Not a random concept that people have to study (and get lost in).

Let's keep it as is.

include/seastar/core/future.hh
avikivity: Let's be descriptive, loose_type_conversion.

btw, do we not have the same problem in future::set_value? and in make_ready_future? If we do this surgery, it's better to capture all the problematic areas at once.

include/seastar/core/future.hh
bashtanov: This comes from existing code too. I don't like it either, it does something nasty about object lifetimes. The standard is vague about whether it's legal. But fixing it is probably out of scope.

avikivity: Ok. Existing code gets a pass.

include/seastar/core/future.hh
bashtanov: The naming problem? I'll change it in all places I introduced the `_legacy` suffix.

Or the general type conversion problem?
For `promise::set_value` I changed the semantics of `set_value` and introduced `emplace_value`.
As for `make_ready_future`, I kept it loose, but made sure the set semantics is honoured -- for omitting types accepting braced lists.

I'll add tests for these to illustrate.

include/seastar/core/coroutine.hh
bashtanov: to group functions that only make sense for API_LEVEL <= 9

bashtanov:
changes:
- renamed *_legacy as discussed
- removed C++ version change from readme
- fixed formatting issues
- added tests for promise.set_value and make_ready_future

include/seastar/core/future.hh
avikivity: We should consider having emplace_ready_future vs make_ready_future.

Not sure what the prior art is around here (in the Standard).

bashtanov: > We should consider having emplace_ready_future vs make_ready_future.

We have them in a sense:
1) `make_ready_future<specified_type>(whatever, arguments)` has emplace semantics.
2) `make_ready_future(single_argument)` has copy/move semantics.
That's similar to the logic of `make_lw_shared` and `std::make_optional`.

Unlike these, `co_return` and `promise::set_value` have prior type info from the object, so they are typically called without explicit template parameters.

The only thing I need change for make_ready_future is to make it 100% compatible with the pre-change version for API_VERSION <=9.

> Not sure what the prior art is around here (in the Standard).

Could you rephrase? I'm not sure I understand it right.

bashtanov:
changes:
* made `make_ready_future` 100% compatible with the pre-change version for API_VERSION <=9
* added a test for it
* removed an incorrect code comment

include/seastar/core/future.hh
avikivity: We have vector::push_back(T) that allows the implicit conversion during argument construction, and vector::emplace_back(A&&...) that explicitly constructs a T from the tuple of A&&s. That's prior are. It matches well for the proposed promise::set_value() and promise::emplace_value().

I'm asking if we can draw a similar parallel for make_ready_future: one name for the implicit-only conversion, and one for explicit construct-from-tuple. I don't think the name emplace_ready_future() works, because we're not emplacing it anywhere. Practically all of the use is returning the same type as the single input, I think it makes sense to have a special name for the constructing variant.

include/seastar/core/coroutine.hh
```diff
@@ -95,9 +95,14 @@ public:
         promise_type(promise_type&&) = delete;
         promise_type(const promise_type&) = delete;
 
-        template<typename... U>
-        void return_value(U&&... value) {
-            _promise.set_value(std::forward<U>(value)...);
+        template<typename U>
+        void return_value(U&& value) {
+            _promise.set_value(std::forward<U>(value));
```
avikivity, line 101: I suppose this is a safe change because there is no way to get this called with sizeof...(U) != 1?

include/seastar/core/future.hh
```diff
@@ -290,16 +297,33 @@ struct uninitialized_wrapper {
 
 public:
     uninitialized_wrapper() noexcept = default;
+    template<typename U>
+    requires std::is_convertible_v<U&&, T>
+    void uninitialized_set(U&& u) {
+        // Despite the `requires` condition above, the line below may still fail compilation if T is a reference type.
+        // E.g. when T is a const lvalue reference and U is an rvalue reference to a non-const type.
+        // These kind of conversions make little sense both here and with plain `return`.
+        new(&_v.value) auto(implicit_convert<maybe_wrap_ref<T>>(std::forward<U>(u)));
```
avikivity, line 319: IIUC auto(x) is C++23 only while we are still C++20/23.

Maybe we can push the C++23/26 patch before this series (but outside it) so we don't have to change this.

include/seastar/core/future.hh
```diff
@@ -290,16 +297,33 @@ struct uninitialized_wrapper {
...
-    uninitialized_set(U&&... vs) {
+    void uninitialized_set_loose_type_conversion(U&&... vs) {
         new (&_v.value) maybe_wrap_ref<T>(T(std::forward<U>(vs)...));
     }
-    void uninitialized_set(tuple_type&& v) {
+    void uninitialized_set_loose_type_conversion(tuple_type&& v) {
         uninitialized_set(std::move(std::get<0>(v)));
     }
-    void uninitialized_set(const tuple_type& v) {
+    void uninitialized_set_legacy(const tuple_type& v) {
```
avikivity: Here is legacy again

include/seastar/core/future.hh
```diff
@@ -557,6 +591,10 @@ inline void future_state_base::any::check_failure() noexcept {
 }
 
 struct ready_future_marker {};
+struct loose_type_conversion_ready_future_marker: ready_future_marker {};
+struct set_ready_future_marker: ready_future_marker {};
+struct set_from_tuple_ready_future_marker: ready_future_marker {};
+struct emplace_ready_future_marker: ready_future_marker {};
```
avikivity: Why the inheritance?

Style: we have space before :.

include/seastar/core/future.hh
```diff
@@ -1232,8 +1355,8 @@ private:
     future(future_for_get_promise_marker) noexcept { }
 
     future(promise<T>* pr) noexcept : future_base(pr, &_state), _state(std::move(pr->_local_state)) { }
-    template <typename... A>
-    future(ready_future_marker m, A&&... a) noexcept : _state(m, std::forward<A>(a)...) { }
+    template <std::derived_from<ready_future_marker> Marker, typename... A>
+    future(Marker m, A&&... a) noexcept : _state(m, std::forward<A>(a)...) { }
```
avikivity: Okay, I see the reason for the inheritance. I think it's better to be explicit and just repeat the constructor for every marker, than risk decaying into the plain marker and being surprised.

doc/compatibility.md
```diff
@@ -78,6 +78,10 @@ versions of the API. For example.
      and become a move-only type
    - Seastar_API_LEVEL=9 defines the data_sink_impl::put(span<temporary_buffer>)
      as the new and only method to be implemented
+   - Seastar_API_LEVEL=10 makes co_return and promise.set_value() semantics
+     closer to those of plain C++ return regarding how they convert
+     expressions used as their arguments to the declared type.
+
```
avikivity, line 84: Please extend the history table below.

tests/unit/CMakeLists.txt
```diff
@@ -346,7 +346,9 @@ seastar_add_test (content_source
   SOURCES content_source_test.cc)
 
 seastar_add_test (coroutines
-  SOURCES coroutines_test.cc)
+  SOURCES
+    coroutines_test.cc
+    conversion_test_types.hh)
```
avikivity: Why do you need to add .hh? Won't the build system deduce it?

avikivity:
Looks good, I added low-level comments. Main remaining point is whether we want to split make_ready_future into two names. If we do, I think it makes sense to do it in this series as it's the same problem.

include/seastar/core/future.hh
bashtanov: auto(x) is C++23, but new auto(x) is C++11

avikivity: Ah, thanks.

include/seastar/core/coroutine.hh
bashtanov: yes, `return_value` is only called by co_return with a single argument https://eel.is/c++draft/stmt.return.coroutine#2.1

include/seastar/core/future.hh
bashtanov: `vector<T>` is similar to `promise<T>` and `future<T>`: the object carries the type info, so neither `push_back` nor `emplace_back` need to specify `T` again. Let me illustrate why it matters.

---
We have 4 ways to set a value for a promise `pr` of type `promise<T>`:
1) `pr.set_value(a value of type T)` -- put as is
2) `pr.set_value(something implicitly convertible to T)` -- implicitly convert and put
3) `pr.emplace_value(any parameters)` -- invoke a ctor, possibly an explicit one; in case of a single argument explicit ctors are called
4) `pr.set_value(T(something))` -- invoke an explicit conversion (whether a ctor or not)

The main reason to have different function names (set vs emplace) is to distinguish between situations (2) and (3): set_value, like vector's push_back, is not expected to invoke an explicit ctor, not least because T is not mentioned in the call (2).

---
In case of `make_ready_future` (or, similarly, `make_lw_shared_ptr` or `make_optional`) we don't yet have an object that holds type info. We have the following ways to call it:
1) `make_ready_future(a value of type T)` -- put as is; this includes supplying a copyable lvalue or a reference -- we still put the value as we consider `future<Foo&>` marginal
2) `make_ready_future<T>(something else implicitly convertible to T)` -- invoke an implicit convertion; note that we need to specity T in angle brackets to define the result type
3) `make_ready_future<T>(any parameters)` -- invoke a ctor, possibly an explicit one; in case of a single argument explicit ctors are called but not other explicit conversions
4) `make_ready_future(T(something))` -- invoke an explicit conversion; note that here it's not longer to type than (2)

Here we don't mind the syntax clash between implicit (2) and potentially-calling-explicit-ctor (3), as they both mention T. There's no risk of accidentally invoking an explicit conversion without ever mentioning the target type -- the very situation this PR fights against.

---
A direct analogy would be `std::optional`. It uses separate `emplace` and `operator=(T&& value )` for changing the value of an existing object, but a single `make_optional` for creating a new one. They cover all the 4 situations above.

bashtanov:
changes:
- fixed build errors
- removed `ready_future_marker` base class
- added v10 to API version table
- removed conversion_test_types.hh from CMakeLists.txt

bashtanov:
The failure seems to be an infrastructure issue

PR #2980, opened 2025-09-15
rpc: sink_impl: extend backpressure until snd_buf destroy

### LINKED ISSUE #2979: rpc: sink_impl: operator(): backpressure units are released too soon
Extracted from https://github.com/scylladb/seastar/pull/2968#discussion_r2348718357:

@gleb-cloudius wrote:
backpresure mechanism in sink_impl<Serializer, Out...>::operator() does not work as expected. It uses semaphore _sem to limit the amount of work on the other shard but units are released before foreign_ptr is freed, so another shard may accumulate a lot of them.
The solution as I see it is to make semaphore_units lifetime to be the same as foreign_ptr (by holding it in the snd_buf for instance).

bhalevy:
Gleb wrote:
> backpresure mechanism in sink_impl<Serializer, Out...>::operator()
> does not work as expected. It uses semaphore _sem to limit the
> amount of work on the other shard but units are released before
> foreign_ptr is freed, so another shard may accumulate a lot of them.
> The solution as I see it is to make semaphore_units lifetime to be
> the same as foreign_ptr (by holding it in the snd_buf for instance).

Fixes #2979
Refs scylladb/scylladb#24818


include/seastar/rpc/rpc_types.hh
```diff
@@ -254,6 +255,7 @@ struct snd_buf {
     static constexpr size_t chunk_size = 128*1024;
     uint32_t size = 0;
     std::variant<std::vector<temporary_buffer<char>>, temporary_buffer<char>> bufs;
```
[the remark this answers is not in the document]
bhalevy: In c8f1c1cb

bhalevy:
c8f1c1cb: added comment documenting new snd_buf semaphore_units member

avikivity:
What about a test?

bhalevy:
> What about a test?

I am working on a test that reproduces an issue without this backpressure extension
but I wasn't able to produce any issue yet in release mode.

bhalevy:
upd: I was able to reproduce `Too long queue` reports, as the one below:
```
WARN  2025-09-22 14:45:10,551 [shard  6:main] seastar - Too long queue accumulated for main (1180 tasks)
 1: N7seastar8internal17do_for_each_stateIN9__gnu_cxx17__normal_iteratorIPNS_16temporary_bufferIcEESt6vectorIS5_SaIS5_EEEESA_ZZNS_23loopback_data_sink_impl3putENS_3net6packetEENKUlRS9_E_clESE_EUlRS5_E_EE
 1: N7seastar12continuationINS_8internal22promise_base_with_typeIvEEZNS_3rpc10connection4sendENS4_7snd_bufESt8optionalINSt6chrono10time_pointINS_12lowres_clockENS8_8durationIlSt5ratioILl1ELl1000000000EEEEEEEPNS4_11cancellableEE3$_1ZNS_6futureIvE14then_impl_nrvoISJ_SL_EET0_OT_EUlOS3_RSJ_ONS_12future_stateINS1_9monostateEEEE_vEE
 484: N7seastar12continuationINS_8internal22promise_base_with_typeIvEEZZNS_3rpc9sink_implI10serializerJNS_13basic_sstringIcjLj15ELb1EEEEEclERKS8_ENUlNS_15semaphore_unitsINS_35semaphore_default_exception_factoryENSt6chrono3_V212steady_clockEEEE_clESH_EUlNS_6futureIvEEE_ZNSK_17then_wrapped_nrvoIvSL_EENS_8futurizeIT_E4typeEOT0_EUlOS3_RSL_ONS_12future_stateINS1_9monostateEEEE_vEE
 484: N7seastar17smp_message_queue15async_work_itemIZNS_11foreign_ptrISt10unique_ptrINS_3rpc7snd_bufESt14default_deleteIS5_EEE10destroy_onES8_jEUlvE_EE
 1: N7seastar17smp_message_queue15async_work_itemIZZZNS_23loopback_data_sink_impl3putENS_3net6packetEENKUlRSt6vectorINS_16temporary_bufferIcEESaIS7_EEE_clESA_ENKUlRS7_E_clESC_EUlvE_EE
 114: N7seastar12continuationINS_8internal22promise_base_with_typeIvEEZNS_6futureIvE16handle_exceptionIZZNS_7reactor17run_in_backgroundES5_EN3$_0clEvEUlNSt15__exception_ptr13exception_ptrEE_Qoooooosr3stdE16is_invocable_r_vINS4_IT_EETL0__SA_Eaaeqsr3stdE12tuple_size_vINSt11conditionalIXsr3stdE9is_same_vINS1_18future_stored_typeIJSC_EE4typeENS1_9monostateEEESt5tupleIJEESL_IJSJ_EEE4typeEELi0Esr3stdE16is_invocable_r_vIvSF_SA_Eaaeqsr3stdE12tuple_size_vISP_ELi1Esr3stdE16is_invocable_r_vISC_SF_SA_Eaagtsr3stdE12tuple_size_vISP_ELi1Esr3stdE16is_invocable_r_vISP_SF_SA_EEES5_OSC_EUlSQ_E_ZNS5_17then_wrapped_nrvoIS5_SR_EENS_8futurizeISC_E4typeEOT0_EUlOS3_RSR_ONS_12future_stateISK_EEE_vEE
 95: N7seastar17smp_message_queue15async_work_itemIZZNS_3rpc9sink_implI10serializerJNS_13basic_sstringIcjLj15ELb1EEEEEclERKS6_ENUlNS_15semaphore_unitsINS_35semaphore_default_exception_factoryENSt6chrono3_V212steady_clockEEEE_clESF_EUlvE_EE
b
```

Note also long queue for waiting for semaphore units.
That said, it was also reproduced with the change (although apparently less frequently), and I want to understand why exactly.

bhalevy:
In [48452c0](https://github.com/scylladb/seastar/commit/48452c01c1c9000f2239416c7c13e91004185730):
- rpc: sink_impl: extend backpressure until snd_buf destroy
  - Reduce concurrency when sending across shards to avoid too-long queues
- reactor: add abort_on_too_long_task_queue option
  - for testing
- rpc: make sink::close noexcept
  - for deferred_close
- test: rpc_test: add test_rpc_stream_backpressure_across_shards
  - tidied-up to focus on cross shard rpc_sink by streaming back to originator from rpc source handler (since source.make_sink() generates a cross-shard connection)

tests/unit/rpc_test.cc
```diff
@@ -1900,3 +1912,94 @@ SEASTAR_TEST_CASE(test_timeout_cancel) {
...
+                                }
+                                count++;
+                                continue;
+                            } else {
+                                log.debug("cl_rep_loop: got end-of-stream");
+                                continue;
+                            }
+                        } else {
+                            log.debug("cl_rep_loop: got empty data");
+                            continue;
```
avikivity, line 1986: Doesn't empty data mark the end-of-stream?

tests/unit/rpc_test.cc
```diff
@@ -1900,3 +1912,94 @@ SEASTAR_TEST_CASE(test_timeout_cancel) {
...
+                        }
+                    } catch (const rpc::stream_closed&) {
+                        log.debug("cl_rep_loop: stream closed");
+                    } catch (...) {
+                        auto msg = format("cl_rep_loop: unexpected exception: {}", std::current_exception());
+                        log.error("{}", msg);
+                        throw std::runtime_error(msg);
+                    }
+                    break;
+                }
```
avikivity, line 1995: This infinite loop is not easy to understand.

I'd expect something like

```c++
    try {
          while (auto msg = source().get()) {
             ...
          }
    } catch ...
```

src/core/reactor.cc
```diff
@@ -1485,6 +1485,29 @@ bool reactor::test::linux_aio_nowait() {
     return engine()._cfg.aio_nowait_works;
 }
 
+reactor::test::long_task_queue_state::~long_task_queue_state() {
+    reactor::test::set_long_task_queue_state(*this);
+}
+
+reactor::test::long_task_queue_state
+reactor::test::save_long_task_queue_state() {
```
avikivity: Why save_...? Seems like get_ is natural here.

include/seastar/core/reactor.hh
```diff
@@ -756,6 +756,18 @@ public:
         static void set_stall_detector_report_function(std::function<void ()> report);
         static std::function<void ()> get_stall_detector_report_function();
         static bool linux_aio_nowait();
+
+        struct long_task_queue_state {
+            std::optional<bool> abort_on_too_long_task_queue;
+            std::optional<unsigned> max_task_backlog;
+
+            // restores this state on destruction
+            ~long_task_queue_state();
```
avikivity: This is easy to misuse. It's both data and RAII guard. Easy to get two objects mixed up.

Better to have a struct for state and a RAII object for a guard (or just use defer()).

avikivity: And better avoid the optionals, it's too much flexibility for a use-once thing.

include/seastar/rpc/rpc_types.hh
```diff
@@ -342,8 +342,8 @@ public:
     future<> operator()(const Out&... args) {
         return _impl->operator()(args...);
     }
-    future<> close() {
-        return _impl->close();
+    future<> close() noexcept {
+        return futurize_invoke([this] { return _impl->close(); });
```
avikivity: IMO, better to make impl::close() noexcept and audit implementations.

include/seastar/rpc/rpc_impl.hh
```diff
@@ -835,11 +836,19 @@ future<> sink_impl<Serializer, Out...>::operator()(const Out&... args) {
     // we do not want to dead lock on huge packets, so let them in
     // but only one at a time
     auto size = std::min(size_t(data.size), max_stream_buffers_memory);
+    if (this->_con->get_owner_shard() != this_shard_id()) {
+        const auto& cfg = engine().get_reactor_config();
+        // Reduce concurrency when sending to a remote shard
+        // to avoid too long task queues.
+        // We take into account 3 outstanding tasks per message.
+        size = std::max(size, max_stream_buffers_memory / (size_t(cfg.max_task_backlog) / 3));
```
avikivity: This calculation works if there is only one stream operating.

avikivity:
I have a feeling this is the wrong approach. For every message we send, we create a cross-shard set of tasks. We have to account for out-of-order messages due to task reordering (which @gleb-cloudius hates). But messages can be tiny (tombstone-only mutation fragments in ScyllaDB).

The producer (caller of sink_impl::operator()) will have no problem producing messages back-to-back.

Suggest this:
1. Add a buffer: vector of messages 
2. Allow at most one cross-smp call at a time
3. When we are called with a new message, and the buffer is empty, and there are no active cross-smp calls, send it to the remote
4. Otherwise, append the new message to the buffer. Return a ready future if the buffer is small enough, or a non-ready future if it's too large.
5. When a cross-smp call returns: send the pending buffer over, and signal the non-ready future we returned in step 4 so the producer can produce again.

The goal here is to have the local producer producing into the local buffer, and send the local buffer batch over with a concurrency of 1. There is no reordering.

tests/unit/rpc_test.cc
bhalevy: Yes, but the connection needs to be drained until we get the closed error to be sure that the other side is done.
Otherwise we'd need a sideband mechanism to await for it.

src/core/reactor.cc
bhalevy: Since it's self-restoring, when destroyed.
I'm ok with get_ too, but I think that save_ conveyes the functionality better.

include/seastar/rpc/rpc_types.hh
bhalevy: okay

bhalevy: just that it would be an an api change, though minor so maybe doesn't require an api level.

include/seastar/rpc/rpc_impl.hh
bhalevy: We can have a semaphore per shard to consume from, implemented in the reactor.
I think we'd need to consume from both shards' semaphores to accomodate for the task submitted to the connection shard and for the destruction of the foreign ptr back on the sender shard.

tests/unit/rpc_test.cc
avikivity: That seems wrong. Doesn't source() return a disengaged optional when done?

Or are you interested in knowing when the implementation is done? If so it should be explained, not hidden.

src/core/reactor.cc
avikivity: It mixes containing the state and restoring it. The idiom is to have two separate objects for that.

include/seastar/rpc/rpc_types.hh
avikivity: I don't think this type is exposed.

gleb-cloudius:
> 
> Suggest this:
> 
>     1. Add a buffer: vector of messages
> 
>     2. Allow at most one cross-smp call at a time
> 
>     3. When we are called with a new message, and the buffer is empty, and there are no active cross-smp calls, send it to the remote
> 
>     4. Otherwise, append the new message to the buffer. Return a ready future if the buffer is small enough, or a non-ready future if it's too large.
> 
>     5. When a cross-smp call returns: send the pending buffer over, and signal the non-ready future we returned in step 4 so the producer can produce again.
> 

It may work, but it will not fix the problem @bhalevy tries to fix here. The problem is that the work submitted to another shard is not fully completed when cross-smp call returns since snd_buf's are freed asynchronously.

bhalevy:
> > Suggest this:
> > ```
> > 1. Add a buffer: vector of messages
> > 
> > 2. Allow at most one cross-smp call at a time
> > 
> > 3. When we are called with a new message, and the buffer is empty, and there are no active cross-smp calls, send it to the remote
> > 
> > 4. Otherwise, append the new message to the buffer. Return a ready future if the buffer is small enough, or a non-ready future if it's too large.
> > 
> > 5. When a cross-smp call returns: send the pending buffer over, and signal the non-ready future we returned in step 4 so the producer can produce again.
> > ```
> 
> It may work, but it will not fix the problem @bhalevy tries to fix here. The problem is that the work submitted to another shard is not fully completed when cross-smp call returns since snd_buf's are freed asynchronously.

The high level problem here is the long queue of cross-shard tasks.
So even if throttling is extended until the snd_buf:s are destroyed, we can still generate a long queue of tasks since each snd_buf is sent via submit_to to the connection shard (and then we need to deal with the aftermath of losing ordering using the hand-crafted sequencing and `out_of_order_bufs`.

What Avi suggests is managing the send queue using a single cross-shard task that would preserve the messages order.
It would still need to extend throttling until the snd_buf's are freed too though.  I agree with that.

gleb-cloudius:
> > > Suggest this:
> > > ```
> > > 1. Add a buffer: vector of messages
> > > 
> > > 2. Allow at most one cross-smp call at a time
> > > 
> > > 3. When we are called with a new message, and the buffer is empty, and there are no active cross-smp calls, send it to the remote
> > > 
> > > 4. Otherwise, append the new message to the buffer. Return a ready future if the buffer is small enough, or a non-ready future if it's too large.
> > > 
> > > 5. When a cross-smp call returns: send the pending buffer over, and signal the non-ready future we returned in step 4 so the producer can produce again.
> > > ```
> > 
> > 
> > It may work, but it will not fix the problem @bhalevy tries to fix here. The problem is that the work submitted to another shard is not fully completed when cross-smp call returns since snd_buf's are freed asynchronously.
> 
> The high level problem here is the long queue of cross-shard tasks. So even if throttling is extended until the snd_buf:s are destroyed, we can still generate a long queue of tasks since each snd_buf is sent via submit_to to the connection shard (and then we need to deal with the aftermath of losing ordering using the hand-crafted sequencing and `out_of_order_bufs`.
> 
> What Avi suggests is managing the send queue using a single cross-shard task that would preserve the messages order. It would still need to extend throttling until the snd_buf's are freed too though. I agree with that.

May be I am wrong, but the report we saw were all about foreign_ptr destruction, not about tasks submitted from sink::op(). The later is limited by the semaphore while the former is not. But I see how a lot of very small message can create a lot of submit_to calls.

avikivity:
> > Suggest this:
> > ```
> > 1. Add a buffer: vector of messages
> > 
> > 2. Allow at most one cross-smp call at a time
> > 
> > 3. When we are called with a new message, and the buffer is empty, and there are no active cross-smp calls, send it to the remote
> > 
> > 4. Otherwise, append the new message to the buffer. Return a ready future if the buffer is small enough, or a non-ready future if it's too large.
> > 
> > 5. When a cross-smp call returns: send the pending buffer over, and signal the non-ready future we returned in step 4 so the producer can produce again.
> > ```
> 
> It may work, but it will not fix the problem @bhalevy tries to fix here. The problem is that the work submitted to another shard is not fully completed when cross-smp call returns since snd_buf's are freed asynchronously.

The work would be reduced by a large amount since cross-smp calls would happen for entire batches (in both directions).

gleb-cloudius:
> 
> The work would be reduced by a large amount since cross-smp calls would happen for entire batches (in both directions).

I do not see what will batch `foregin_prt` destructors.

avikivity:
We could wrap the vector with a foreign_ptr rather than individual snd_bufs. Though it may be hard to keep them in the vector.

Alternatively, detach them from the vector, then collect them again after use.

Note the smp call that sends the vector blocks until it's processed, in order to let the tcp listener accumulate a new batch.

gleb-cloudius:
`snd_buf` is moved around. Hard to collect. We can create a fancy deleter that tries to collect them, but the point is that the sender should wait for them to be deleted, not just sent otherwise they may accumulate.

Why do we suddenly care so much about cross shard streaming which should not be happening in normal circumstances? Is this because of tablet brokenness that does not  preserve shard locality? I will kill our performance in many other places as well.

avikivity:
It's not only tablets, it's also mixed node that sometimes happens.

Eventually we need to full to full mesh (shard-to-shard) but we can't do that with TCP. I'd like to see RPC over QUIC.

avikivity:
```c++
            auto ret_fut = con->send(std::move(local_data), {}, nullptr);
```

This becomes a loop over the vector, no? We can make ret_fut return local_data.

gleb-cloudius:
> ```c++
>             auto ret_fut = con->send(std::move(local_data), {}, nullptr);
> ```
> 
> This becomes a loop over the vector, no? We can make ret_fut return local_data.

I do not understand the context of the question. If we will change the code to pass vectors, then yes it will become a loop, but again the issue solved by this PR is different.

avikivity:
> > ```c++
> >             auto ret_fut = con->send(std::move(local_data), {}, nullptr);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This becomes a loop over the vector, no? We can make ret_fut return local_data.
> 
> I do not understand the context of the question. If we will change the code to pass vectors, then yes it will become a loop, but again the issue solved by this PR is different.

If ret_fut returns local_data (or rather a share of it since we pass the buffer to tcp), then we can collect all local_data:s back in a vector, and return it to the caller, which can then free them all locally.

avikivity:
btw I see only the snd_buf/rcv_buf are managed with foreign_ptr, the actual temporary_buffers are managed with cross-shard frees.

temporary_buffers are hard to manage cross-shard since they're consumed by tcp if we use zero-copy to send them. Not sure what to do here.

gleb-cloudius:
> > > ```c++
> > >             auto ret_fut = con->send(std::move(local_data), {}, nullptr);
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This becomes a loop over the vector, no? We can make ret_fut return local_data.
> > 
> > 
> > I do not understand the context of the question. If we will change the code to pass vectors, then yes it will become a loop, but again the issue solved by this PR is different.
> 
> If ret_fut returns local_data (or rather a share of it since we pass the buffer to tcp), then we can collect all local_data:s back in a vector, and return it to the caller, which can then free them all locally.

If `local_data` does not outlive `send()` then no `foreign_ptr` is needed at all and it is enough to make sure that the buffer is alive on original shard until `submit_to` returns. This is what @bhalevy did initially and I pointed out that that it only works if above condition is true and asked to review the code with this in mind. As far as I understand he arrived to the conclusion that it is not the case. May be I need to review it as well. May be you should do it too. The consequences of a mistake are too severe.

gleb-cloudius:
> btw I see only the snd_buf/rcv_buf are managed with foreign_ptr, the actual temporary_buffers are managed with cross-shard frees.
> 

Where do you see that? The code creates `temporary_buffers` from the pointers on another shard using `temporary_buffer(CharType* buf, size_t size, deleter d)` constructor and provided deleter uses foreign_ptr to release those pointers on original shard. See `make_shard_local_buffer_copy` for details.

avikivity:
> > btw I see only the snd_buf/rcv_buf are managed with foreign_ptr, the actual temporary_buffers are managed with cross-shard frees.
> 
> Where do you see that? The code creates `temporary_buffers` from the pointers on another shard using `temporary_buffer(CharType* buf, size_t size, deleter d)` constructor and provided deleter uses foreign_ptr to release those pointers on original shard. See `make_shard_local_buffer_copy` for details.

Right.

avikivity:
I'm thinking now it may be better to copy the data. All these tiny allocations are expensive, if the messages are small.

If the messages are large, we can zero-copy.

Copying 100 bytes is 14 instructions + a small if-tree.

gleb-cloudius:
> I'm thinking now it may be better to copy the data. All these tiny allocations are expensive, if the messages are small.
> 
> If the messages are large, we can zero-copy.
> 
> Copying 100 bytes is 14 instructions + a small if-tree.

May be. We can hide it in `make_shard_local_buffer_copy` probably. But why would streaming generates very small packets. I would expect it to send large chunks.

bhalevy:
> > > > ```c++
> > > >             auto ret_fut = con->send(std::move(local_data), {}, nullptr);
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > This becomes a loop over the vector, no? We can make ret_fut return local_data.
> > > 
> > > 
> > > I do not understand the context of the question. If we will change the code to pass vectors, then yes it will become a loop, but again the issue solved by this PR is different.
> > 
> > 
> > If ret_fut returns local_data (or rather a share of it since we pass the buffer to tcp), then we can collect all local_data:s back in a vector, and return it to the caller, which can then free them all locally.
> 
> If `local_data` does not outlive `send()` then no `foreign_ptr` is needed at all and it is enough to make sure that the buffer is alive on original shard until `submit_to` returns. This is what @bhalevy did initially and I pointed out that that it only works if above condition is true and asked to review the code with this in mind. As far as I understand he arrived to the conclusion that it is not the case. May be I need to review it as well. May be you should do it too. The consequences of a mistake are too severe.

My conclusion was that in the version I submitted local_data, moved to the output stream might outlive the original foreign snd_buf.

gleb-cloudius:
> 
> My conclusion was that in the version I submitted local_data, moved to the output stream might outlive the original foreign snd_buf.

Makes sense. Otherwise why would I over engineer is so much :)

gleb-cloudius:
> > I'm thinking now it may be better to copy the data. All these tiny allocations are expensive, if the messages are small.
> > If the messages are large, we can zero-copy.
> > Copying 100 bytes is 14 instructions + a small if-tree.
> 
> May be. We can hide it in `make_shard_local_buffer_copy` probably. But why would streaming generates very small packets. I would expect it to send large chunks.

But I do not see how it will solve small allocations problem since allocations are done during marshaling, not during `submit_to` which zero copies.

avikivity:
> > I'm thinking now it may be better to copy the data. All these tiny allocations are expensive, if the messages are small.
> > If the messages are large, we can zero-copy.
> > Copying 100 bytes is 14 instructions + a small if-tree.
> 
> May be. We can hide it in `make_shard_local_buffer_copy` probably. But why would streaming generates very small packets. I would expect it to send large chunks.

One mutation fragment = one message.

A mutation fragment can be tiny (partition_start, partition_end)

avikivity:
> > > I'm thinking now it may be better to copy the data. All these tiny allocations are expensive, if the messages are small.
> > > If the messages are large, we can zero-copy.
> > > Copying 100 bytes is 14 instructions + a small if-tree.
> > 
> > 
> > May be. We can hide it in `make_shard_local_buffer_copy` probably. But why would streaming generates very small packets. I would expect it to send large chunks.
> 
> But I do not see how it will solve small allocations problem since allocations are done during marshaling, not during `submit_to` which zero copies.

Every deleter is an allocation (except free_deleter).

deleter::share() converts a free_deleter to an allocation.

So every fragment acquires an allocation in make_shard_local_buffer_copy(). Might as well make it a true copy for small fragments.

bhalevy:
New in [688fcff](https://github.com/scylladb/seastar/commit/688fcff78adc7f79a3b43cd3256c7d294f9f04bb):
- rpc: sink_impl: extend backpressure until snd_buf destroy
  - simplified.  no longer limits the queue in relation to the queue length warning level.
- rpc: move sink_impl and source_impl into internal namespace
  - just a code cleanup, no functional changes in this patch
- rpc: sink_impl: batch sending and deletion of snd_buf:s
  - this is the core change
- rpc: make sink flush and close noexcept
  - made the impl flush() and close() noexcept, as per https://github.com/scylladb/seastar/pull/2980#discussion_r2384122307

test left and related infrastructure changes left unchanged.

I'll address @avikivity's comments about the test infrastructure in the next version

bhalevy:
In [8f505d1](https://github.com/scylladb/seastar/commit/8f505d1c21d318b25f6ff73ae6a5c781f64ede41):
- simplified test infrastructure for getting, setting, and restoring the reactor `abort_on_too_long_task_queue` and `max_task_backlog` config options.

include/seastar/rpc/rpc_impl.hh
```diff
@@ -822,68 +825,83 @@ std::optional<protocol_base::handler_with_holder> protocol<Serializer, MsgType>:
...
-                this->_ex = f.get_exception();
-            } else {
-                f.ignore_ready_future();
-            }
+        // No exceptions allowed beyond this point since we release the unique_ptr deleter
+        data->su = std::move(su);
+        return std::invoke([this, data = data.release()] () noexcept {
+            _queue.push_back(*data);
+            return send_loop();
         });
```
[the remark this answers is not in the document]
bhalevy: Just following the local coding convention

tests/unit/rpc_test.cc
```diff
@@ -1899,3 +1911,93 @@ SEASTAR_TEST_CASE(test_timeout_cancel) {
...
+                            auto val = std::get<0>(*data);
+                            if (val.size()) {
+                                if (count && !(count % 100000)) {
+                                    log.debug("cl_rep_loop: received {} messages...", count);
+                                }
+                                count++;
+                                continue;
+                            } else {
+                                log.debug("cl_rep_loop: got end-of-stream");
+                                continue;
```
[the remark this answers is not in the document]
bhalevy: Not really.  Added a comment explaining why we continue rather than break in the next version

bhalevy:
[2f9fbc9](https://github.com/scylladb/seastar/commit/2f9fbc93bacbb41a8729086e1a399cd5e220c83d): addressed copilot's review comments.

include/seastar/core/reactor_config.hh
```diff
@@ -135,6 +136,8 @@ struct reactor_options : public program_options::option_group {
     program_options::value<> overprovisioned;
     /// \brief Abort when seastar allocator cannot allocate memory.
     program_options::value<> abort_on_seastar_bad_alloc;
+    /// \brief Abort when seastar allocator cannot allocate memory.
```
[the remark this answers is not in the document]
bhalevy: Fixed in [2f9fbc9](https://github.com/scylladb/seastar/commit/2f9fbc93bacbb41a8729086e1a399cd5e220c83d)

include/seastar/rpc/rpc_impl.hh
```diff
@@ -822,68 +825,83 @@ std::optional<protocol_base::handler_with_holder> protocol<Serializer, MsgType>:
...
+namespace internal {
+
+snd_buf make_shard_local_buffer_copy(snd_buf* org, std::function<deleter(snd_buf*)> make_deleter);
+
+template<typename Serializer, typename... Out>
+future<> sink_impl<Serializer, Out...>::send_loop() {
+    if (!_send_loop.available()) {
+        return make_ready_future();
+    }
+    return do_until([this] { return _queue.empty(); }, [this] {
```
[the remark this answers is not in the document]
bhalevy: Copilot, you're hallucinating :)
Getting all available semaphore units in flush() and close() makes sure that all outstanding snd_buf:s are processed and destroyed, so there is no need for further synchronization with `_send_loop` (or `delete_loop`)

src/rpc/rpc.cc
```diff
@@ -74,14 +74,61 @@ temporary_buffer<char>& snd_buf::front() {
...
+        return;
+    }
+    _state.delete_loop = do_until([&state = _state] { return state.delete_queue.empty(); }, [&state = _state] {
+        auto batch = std::move(state.delete_queue);
+        return smp::submit_to(state.local_shard, [batch = std::move(batch)] () mutable {
+            return do_until([&batch] { return batch.empty(); }, [&batch] () mutable {
+                auto* buf_ptr = &batch.front();
+                batch.pop_front();
+                destroy_and_delete(buf_ptr);
+                return make_ready_future();
```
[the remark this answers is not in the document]
bhalevy: there is no race and there are no multiple threads.
The sink_impl works across 2 shards at the most - one (referred to as local_shard) generating the messages and one (referred to as remote_shard) sends the message on the rpc connection, possibly on another shard.

bhalevy:
[030f127](https://github.com/scylladb/seastar/commit/030f1274a860bc7a9860ab6f6d7fbf07e8746da3): fixed header dependency issue

bhalevy:
Alpine Linux / build-and-test fails with rpc_test timeout.
See https://github.com/scylladb/seastar/actions/runs/18129947086/job/51593922425?pr=2980

But I'm not sure how to reproduce it and get to the bottom of it.
It doesn't reproduce for me locally.

bhalevy:
```
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
```
what are these?

```
WARN  2025-09-30 12:47:32,783 [shard 0:main] seastar - Exceptional future ignored: St9bad_alloc (std::bad_alloc), backtrace: 
   --------
   seastar::smp_message_queue::async_work_item<seastar::rpc::server::connection::deregister_this_stream()::{lambda()#1}>
ERROR 2025-09-30 12:47:34,662 [shard 0:main] seastar - Timer callback failed: St9bad_alloc (std::bad_alloc)
```
eh?

bhalevy:
[874520d](https://github.com/scylladb/seastar/commit/874520d4ff222578c28074476bb6b9a34ad13e20):
- rebased
- used boost::intrusive singly-linked list instead of doubly-linked list to reduce the per-buffer footprint to a single pointer
  - this is possible since the lists are pure FIFOs and entries are always queued at the back and consumed from the front.

bhalevy:
> ```
> ninja: corrupt build log: missing field
> ninja: corrupt build log: missing field
> ninja: corrupt build log: missing field
> ninja: corrupt build log: missing field
> ninja: corrupt build log: missing field
> ninja: corrupt build log: missing field
> ninja: corrupt build log: missing field
> ninja: corrupt build log: missing field
> ```
> 
> what are these?
> 
> ```
> WARN  2025-09-30 12:47:32,783 [shard 0:main] seastar - Exceptional future ignored: St9bad_alloc (std::bad_alloc), backtrace: 
>    --------
>    seastar::smp_message_queue::async_work_item<seastar::rpc::server::connection::deregister_this_stream()::{lambda()#1}>
> ERROR 2025-09-30 12:47:34,662 [shard 0:main] seastar - Timer callback failed: St9bad_alloc (std::bad_alloc)
> ```
> 
> eh?

@avikivity @xemul I don't know how to reproduce this failure, which seems persistent.
Who knows more about this Alpine Linux / build_and_test test?

avikivity:
There's some persistent ccache state. I'll try to delete it.

avikivity:
> There's some persistent ccache state. I'll try to delete it.

Apparently the build recovered from it. It failed on the rpc unit test.

bhalevy:
[1d9dc73](https://github.com/scylladb/seastar/commit/1d9dc731ea6ca8254b9b9f59246d9e29bbfc0a6c):
- no need to send empty string to indicate end-of-stream.
  - rely on disengaged option from source().get()
  - verify singular EOS

bhalevy:
@avikivity @xemul [now](https://github.com/scylladb/seastar/actions/runs/18256558350/job/51978698307?pr=2980) the Alpine Linux build_and_test job failed in scheduling_group_nesting_test.
I don't know what's special about this test environment that yields those failures.

include/seastar/rpc/rpc.hh
```diff
@@ -390,9 +391,34 @@ public:
 
 namespace internal {
 
-struct deferred_snd_buf {
-    promise<> pr;
-    snd_buf data;
+// Used on the shard the _conn lives on.
+struct alignas (cache_line_size) sink_impl_remote_state {
+    shard_id local_shard;
+    shard_id remote_shard;
```
avikivity: local and remote are subjective terms. They can't be understood (and I still don't) without knowing where you stand.

Let's use objecting name like tcp_connection_shard and sink_shard.

I can't see why you'd need both in the same struct. If you're in one, you know what your shard is.

include/seastar/rpc/rpc.hh
```diff
@@ -390,9 +391,34 @@ public:
...
+    future<> delete_loop = make_ready_future();
+};
+
+// Safely delete the original allocation buffer on the local shard
+// When deleted after it was sent on the remote shard, we queue
+// up the buffer pointers to be destroyed and deleted as a batch
+// back on the local shard.
+class snd_buf_deleter_impl final : public deleter::impl {
+    snd_buf* _obj_ptr;
+    sink_impl_remote_state& _state;
```
avikivity: Where does this deleter live?

Let's use objective terms so it can be understood directly.

include/seastar/rpc/rpc.hh
```diff
@@ -390,9 +391,34 @@ public:
...
+
+    static void destroy_and_delete(snd_buf* obj_ptr) {
+        obj_ptr->~snd_buf();
+        std::default_delete<snd_buf>()(obj_ptr);
```
avikivity: Isn't this `delete obj_ptr`?

avikivity: And you just called the destructor! confused.

include/seastar/rpc/rpc.hh
```diff
@@ -402,16 +428,24 @@ class sink_impl : public sink<Out...>::impl {
     alignas (cache_line_size) uint64_t _next_seq_num = 1;
 
     // Used on the shard the _conn lives on.
-    struct alignas (cache_line_size) {
-        uint64_t last_seq_num = 0;
-        std::map<uint64_t, deferred_snd_buf> out_of_order_bufs;
-    } _remote_state;
+    sink_impl_remote_state _remote_state;
+
+    bi::slist<snd_buf, bi::constant_time_size<false>, bi::cache_last<true>> _queue;
```
avikivity: It may be better to use a vector.

include/seastar/rpc/rpc_impl.hh
```diff
@@ -824,64 +825,76 @@ std::optional<protocol_base::handler_with_holder> protocol<Serializer, MsgType>:
...
+        // Grab the current queue and send it on the remote shard
+        // It can be moved safely across shards as it uses an intrusive list
+        auto batch = std::move(_queue);
```
avikivity: std::exchange() is better here since the clearing of the current queue is part of the job, not a side effect.

src/rpc/rpc.cc
```diff
@@ -76,14 +76,59 @@ temporary_buffer<char>& snd_buf::front() {
 
 namespace internal {
 
+snd_buf_deleter_impl::~snd_buf_deleter_impl() {
+    if (this_shard_id() == _state.local_shard) {
+        destroy_and_delete(_obj_ptr);
+        return;
+    }
+    SEASTAR_ASSERT(this_shard_id() == _state.remote_shard);
+    _state.delete_queue.push_back(*_obj_ptr);
```
avikivity: I see, there's no ordering around destruction of the buffers on the application side (though in practice they'll be destroyed in the same order), so you reconstruct a new queue from destroyed temporary buffers.

avikivity:
While complicated, looks good.

Worth to test with scylladb mutation streaming and tiny mutations, with cross-shard streams, and observe the reduction in cross-shard calls.

include/seastar/rpc/rpc.hh
bhalevy: > local and remote are subjective terms. They can't be understood (and I still don't) without knowing where you stand.
> 
> Let's use objecting name like tcp_connection_shard and sink_shard.

okay

> I can't see why you'd need both in the same struct. If you're in one, you know what your shard is.

The sink shard submits snd_buf:s for transmission on the connection shard (in batches).
The connection shard submits the exhausted snd_buf:s back to the sink shard to be destroyed (also in batches)

include/seastar/rpc/rpc.hh
bhalevy: The deleter lives on the connection shard (and therefore it can modify `sink_impl_remote_state`)

include/seastar/rpc/rpc.hh
bhalevy: I did that since it was allocated using std::unique_ptr.
If we'd use `new` to allocate it, then `delete` would be appropriate here.
I thought of this too after submitting this version.
I think using new and delete makes more sense in this context.
We just need to be careful to delete the snd_buf on exception in `sink_impl::operator()`, which is achieved today since were using std::unique_ptr

include/seastar/rpc/rpc.hh
bhalevy: The problem is transferring the vector across shards so it would probably need to be a `foreign_ptr<std::unique_ptr<std::vector>>` which isn't too bad.  Otherwise, maybe we could reuse the vector allocation for queuing the exhausted snd_buf:s for destruction, but it is really ugly since the production rate isn't the same as the consumption rate - i.e. when we process a batch of snd_buf:s they are consumed by the socket and eventually the respective deleter is destroyed and it queues the snd_buf back for destruction, but I don't want to wait for the whole batch to complete and refill the same vector, so the foreign vector used by the sink shard to queue the outgoing messages can be destroyed when full consumed, and a foreign vector can be allocation by the connection shard to queue up exhausted snd_buf:s for destruction, and then be destroyed on the sink shard when the destroy batch completes.

So all in all, I think that an intrusive slist could be simpler and more efficient than a vector.

include/seastar/rpc/rpc_impl.hh
bhalevy: Can do, though std::moving a vector practically clears it.
std::exchange makes it more explicit.

src/rpc/rpc.cc
bhalevy: exactly

include/seastar/rpc/rpc.hh
bhalevy: ok, no need for foreign_ptr.  See 7bd9b0b814c40e6166957ed4281cd0486feb8b9e

bhalevy:
[7bd9b0b](https://github.com/scylladb/seastar/commit/7bd9b0b814c40e6166957ed4281cd0486feb8b9e):
- define  snd_buf_batched_queue that is based small_vector based 
- used both for batch processing on both the send and delete paths.

include/seastar/rpc/rpc_impl.hh
bhalevy: Used std::swap in [7bd9b0b](https://github.com/scylladb/seastar/commit/7bd9b0b814c40e6166957ed4281cd0486feb8b9e)
But for a slightly different purpose :)
There it may help reduce allocations for tx bursts

include/seastar/rpc/rpc.hh
bhalevy: Used `delete` in [7bd9b0b](https://github.com/scylladb/seastar/commit/7bd9b0b814c40e6166957ed4281cd0486feb8b9e)

avikivity:
@gleb-cloudius please review again

include/seastar/rpc/rpc_impl.hh
```diff
@@ -837,50 +872,12 @@ future<> sink_impl<Serializer, Out...>::operator()(const Out&... args) {
     // we do not want to dead lock on huge packets, so let them in
     // but only one at a time
     auto size = std::min(size_t(data.size), max_stream_buffers_memory);
-    const auto seq_num = _next_seq_num++;
-    return get_units(this->_sem, size).then([this, data = make_foreign(std::make_unique<snd_buf>(std::move(data))), seq_num] (semaphore_units<> su) mutable {
+    return get_units(this->_sem, size).then([this, data = std::make_unique<snd_buf>(std::move(data))] (semaphore_units<> su) mutable {
```
gleb-cloudius: Why do you need to move `data` into a `unique_ptr` now?

src/rpc/rpc.cc
```diff
@@ -76,14 +76,57 @@ temporary_buffer<char>& snd_buf::front() {
...
+future<> snd_buf_batched_queue::process_loop() {
+    return seastar::do_until([this] { return _queue.empty(); }, [this] {
+        _cur_batch.clear();
+        std::swap(_cur_batch, _queue);
+        _cur_batch_pos = _cur_batch.begin();
+        return smp::submit_to(_processing_shard, [this] {
+            return seastar::do_until([this] { return _cur_batch_pos == _cur_batch.end(); }, [this] {
+                auto* buf = *_cur_batch_pos;
+                ++_cur_batch_pos;
+                return _process_func(buf);
```
gleb-cloudius: Why not co-routine and then you can just loop with `for (auto i : _cur_batch)`?

src/rpc/rpc.cc
```diff
@@ -76,14 +76,57 @@ temporary_buffer<char>& snd_buf::front() {
...
+    }
+
+    return buf;
+}
+
 // Make a copy of a remote buffer. No data is actually copied, only pointers and
 // a deleter of a new buffer takes care of deleting the original buffer
-template<typename T> // T is either snd_buf or rcv_buf
-T make_shard_local_buffer_copy(foreign_ptr<std::unique_ptr<T>> org) {
+rcv_buf make_shard_local_buffer_copy(foreign_ptr<std::unique_ptr<rcv_buf>> org) {
```
gleb-cloudius, line 103: Don't we have the same problem during receive?

include/seastar/rpc/rpc.hh
```diff
@@ -390,28 +392,63 @@ public:
...
+
+public:
+    snd_buf_deleter_impl(snd_buf* obj_ptr, snd_buf_batched_queue& delete_queue)
+        : impl(deleter())
+        , _obj_ptr(obj_ptr)
+        , _delete_queue(delete_queue)
+    {}
+
+    virtual ~snd_buf_deleter_impl() override {
+        _delete_queue.enqueue(_obj_ptr);
```
gleb-cloudius, line 468: This may throw, no?

bhalevy: Yes, it might.
We can reserve the space ahead of time when processing the send queue to the maximum of outstanding snd_buf:s, but it may be an overkill.

That's one of the reasons I prefer the intrusive list approach.

include/seastar/rpc/rpc_impl.hh
bhalevy: It was to ensure that it will get destroyed on failure (including say semaphore broken).
That said, below should change to the following to achieve that in the continuation:
```
    _send_queue.enqueue(data.get());
    data.release();
```

src/rpc/rpc.cc
bhalevy: To optimize for the common case of batch of size 1 in which case no allocation would be needed.

include/seastar/rpc/rpc_impl.hh
bhalevy: using an intrusive list would also remove the need to potentially allocate here...

src/rpc/rpc.cc
bhalevy: I wasn't able to reproduce it.
In any case, I'd separate the changes

tests/unit/rpc_test.cc
bhalevy: Can do.
the special handling is for `rpc::stream_closed`, and since it's a final state we can catch it outside the infinite loop.

include/seastar/rpc/rpc.hh
bhalevy: @avikivity what do you think regarding using intrusive lists for the queues vs. a vector?
Intrusive lists save allocations and make the buffers more easily transferable across shards.

include/seastar/rpc/rpc_impl.hh
gleb-cloudius: But data is not a pointer, so it will be destroyed if the lambda is destroyed.

src/rpc/rpc.cc
gleb-cloudius: `do_until` also allocates a task, no?

include/seastar/rpc/rpc.hh
avikivity: I think it's okay.

intrusive lists are expensive because each item will take a cache miss and stall the machine. With a vector, each miss brings in multiple items and can initiate a prefetch for even more, so the stall is better amortized.

Still, we can allow the list here.

src/rpc/rpc.cc
avikivity: do_until allocates a task only if its subordinate func does not return a ready future.

include/seastar/rpc/rpc.hh
avikivity: An alternative is boost::static_vector (or std::inplace_vector). It does not allocate. The downside is that if the capacity is exhausted, the connection shard has to stop processing.

include/seastar/rpc/rpc_impl.hh
bhalevy: With intrusive list I created data initially as unique_ptr and released the pointer when pushing to the queue in the .then continuation, after getting the semaphore units.

include/seastar/rpc/rpc.hh
bhalevy: > An alternative is boost::static_vector (or std::inplace_vector). It does not allocate. The downside is that if the capacity is exhausted, the connection shard has to stop processing.

We can use a null-terminated std::array as well for that and then the queue size would be limited either by the queue length or by the semaphore and the queue can be statically sized based on a reasonable small message size.

bhalevy: > I think it's okay.
> 
> intrusive lists are expensive because each item will take a cache miss and stall the machine. With a vector, each miss brings in multiple items and can initiate a prefetch for even more, so the stall is better amortized.
> 
> Still, we can allow the list here.

Here the vector contains pointers to the snd_buf:s so the random access to memory happens anyway.
I'm not sure how much prefetching the next pointers in the queue would help.  We can prefetch manually using the boost intrusive slist as well, but again, I'm not sure how much it will buy us.

include/seastar/rpc/rpc_impl.hh
gleb-cloudius: From that reply I cannot figure out if it is still relevant to the current code.

include/seastar/rpc/rpc.hh
bhalevy: @avikivity after playing around a bit with the idea of a static vector / array, it will require another semaphore and managing the units throughout the snd_buf lifecycle which adds another complication.
All in all, I'll go back to the boost intrusive list which is the simplest overall.

bhalevy:
In v5 ([1fb710f](https://github.com/scylladb/seastar/commit/1fb710fb8d9eeffb0378b5d7b6861d98826fc774)):
- rpc: sink_impl: batch sending and deletion of snd_buf:s
  - Use boost::intrusive::slist
  - Templatize `batched_queue<snd_buf>`
  - rpc_test: move try/catch around infinite for loop, rather than the other was around

include/seastar/rpc/rpc_impl.hh
bhalevy: Moved back to using boost::intrusive::slist in [1fb710f](https://github.com/scylladb/seastar/commit/1fb710fb8d9eeffb0378b5d7b6861d98826fc774)

include/seastar/rpc/rpc.hh
bhalevy: Used boost::intrusive::slist in [1fb710f](https://github.com/scylladb/seastar/commit/1fb710fb8d9eeffb0378b5d7b6861d98826fc774)
Its push_back method doesn't throw

tests/unit/rpc_test.cc
bhalevy: Move try/catch around infinite for loop in [1fb710f](https://github.com/scylladb/seastar/commit/1fb710fb8d9eeffb0378b5d7b6861d98826fc774)

include/seastar/rpc/rpc.hh
avikivity: The hardware will prefetch pointers from the array automatically.

You cannot prefetch across a linked list because you don't know what address to prefetch. You have to wait until you have data in the next node.

avikivity: > @avikivity after playing around a bit with the idea of a static vector / array, it will require another semaphore and managing the units throughout the snd_buf lifecycle which adds another complication. All in all, I'll go back to the boost intrusive list which is the simplest overall.

ok

avikivity:
> Templatize batched_queue<snd_buf>

Why?

bhalevy:
> > Templatize batched_queue<snd_buf>
> 
> Why?

In case we'd want to use it for rcv_buf as a followup.
Though the templatization can be done later if needed.

avikivity:
> > > Templatize batched_queue<snd_buf>
> > 
> > 
> > Why?
> 
> In case we'd want to use it for rcv_buf as a followup. Though the templatization can be done later if needed.

Would it be needed for rcv_buf?

bhalevy:
> > > > Templatize batched_queue<snd_buf>
> > > 
> > > 
> > > Why?
> > 
> > 
> > In case we'd want to use it for rcv_buf as a followup. Though the templatization can be done later if needed.
> 
> Would it be needed for rcv_buf?

Given the code in `connection::stream_receive` and `source_impl<Serializer, In...>::operator()` I suspect it would be needed, as the foreign receive buffers are also deleted one at a time (using make_shard_local_buffer_copy and a deferred deleter) , so a long burst of received buffers could cause a long queue of smp tasks to destroy them when the stream source connection shard is different than the source shard, but I haven't reproduced long tasks queues on this path yet.

include/seastar/rpc/rpc.hh
```diff
@@ -478,8 +478,8 @@ class sink_impl : public sink<Out...>::impl {
 public:
```
gleb-cloudius, line 478: > Make sure any errors are returned as an exceptional

The patch asserts that, not "makes sure" :)

include/seastar/rpc/rpc.hh
```diff
@@ -401,28 +403,88 @@ public:
...
+
+    future<> stop() noexcept {
+        return std::exchange(_process_fut, make_ready_future());
+    }
+
+    void enqueue(T* buf) noexcept {
+        _queue.push_back(*buf);
+        if (_process_fut.available()) {
+            _process_fut = process_loop();
+        }
```
avikivity, line 435: Hmm, this depends on do_until (in process_loop) atomically checking the termination condition and making the future ready if needed. Which does hold.

src/rpc/rpc.cc
```diff
@@ -78,12 +78,33 @@ namespace internal {
...
+        std::vector<temporary_buffer<char>> newbufs;
+        newbufs.reserve(orgbufs.size());
+        auto d = make_deleter(org);
+        for (auto&& b : orgbufs) {
+            newbufs.emplace_back(b.get_write(), b.size(), d.share());
+        }
+        buf.bufs = std::move(newbufs);
+    }
+
+    return buf;
```
avikivity, line 98: We could share the deleter for an entire batch rather than making one for each snd_buf. That would remove a lot of allocations. Since it's a batch we don't gain anything by having fine-grained deleters.

We can do that in a follow up.

bhalevy:
Cc @xemul

PR #2981, opened 2025-09-15
io_tester: Don't skip 0 position in sequential workflows

xemul:
When performing IO sequentially, io_tester's get_pos() helper adds request size to initial value of the "current position" and then returns it, thus skipping the 0th position in a file.

Fix it, and rename the _last_pos to be _next_seq_pos to better reflect what the field is about.

### DIFF
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -532,7 +532,7 @@ public:
 };
 
 class io_class_data : public class_data {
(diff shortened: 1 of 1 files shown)


xemul:
Came from https://github.com/scylladb/scylladb/issues/25532#issuecomment-3292301862, but this change doesn't affect the results of those measurements

xemul:

xemul:

Issue #2983, opened 2025-09-16
Closing socket that still receives data from the peer should abort the connection early


xemul:
At least, this is how TCP sockets behave by default (see SO_LINGER). When closing a socket with non-empty read buffer, the connection is instantly reset (RST packet is sent).

Test from https://github.com/xemul/seastar/tree/br-socket-shutdown-test passes on TCP sockets, but fail on TLS

PR #2986, opened 2025-09-17
tls: Rework session termination

xemul:
First things first, when a TLS socket is shutdown or closed, it leaves a background fiber that sends GNUTLS_BYE message, wires a timer for 10 seconds and waits until EOF appears on the socket. If the timer fires earlier, the underlying socket is closed (thus implicitly causing the EOF to appear).

Whether or not to wait for the EOF is controlled by the tls_options bit called wait_for_eof_on_shutdown, and it's "true" by default.

This patch includes several interconnected enhancements to that process.

First, the aforementioned bool value is replaced with with the timeout value in seconds (default is 10). The related change: if the configured timeout is zero, the BYE message is not sent at all, the socket is just closed. This is not compatible with existing wait_for_eof_on_shutdown which does sends BYE message, but since the socket is closed instantly without letting TLS wrap the session correctly, the incompatibility seems to be OK.

Next, here's the second option called wait_for_data_on_shutdown. After BYE message is sent, if the option is false (which it is by default), the session waits for any value resolved from the do_get() method's future. The value can be a buffer with data or an empty buffer indicating an EOF.  In any case, after an event the transport socket is closed.

That option plays well with clients that read all data from server before closing the socket. In that case, the only "event" that would appear from do_get() is the EOF one -- despite BYE message implies data transfer between client and server, session data doesn't resolve into buffer from do_get(), so only after successful session termination the EOF would appear. In case client shuts down the connection without reading all the data from the server, some unread data would be read, the underlying socket will be closed thus aborting the proper TLS termintation. If that's not called for, the wait_for_data_on_shutdown can be set to true.

Finally, the existing wait_for_eof_on_shutdown is deprecated in favor of setting the bye_timeout to zero.

Fixes: #2983

### DIFF
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -366,8 +366,21 @@ namespace tls {
 
     /// TLS configuration options
     struct tls_options {
+    private:
+        struct deprecated_wait_for_eof_on_shutdown {
+            bool _value = true;
+            deprecated_wait_for_eof_on_shutdown() noexcept : _value(true) {}
+            [[deprecated("Use tls::options::bye_timeout instead")]]
+            deprecated_wait_for_eof_on_shutdown(bool value) noexcept : _value(value) {}
+            [[deprecated("Use tls::options::bye_timeout instead")]]
+            void operator=(bool value) noexcept { _value = value; }
+            [[deprecated("Use tls::options::bye_timeout instead")]]
+            operator bool() const noexcept { return _value; }
+        };
+
+    public:
         /// \brief whether to wait for EOF from server on session termination
-        bool wait_for_eof_on_shutdown = true;
+        deprecated_wait_for_eof_on_shutdown wait_for_eof_on_shutdown;
         /// \brief server name to be used for the SNI TLS extension
         sstring server_name = {};
 
@@ -382,6 +395,16 @@ namespace tls {
         /// \brief Optional list of ALPN protocols to offer to the server,
         /// in order of preference.
         std::vector<sstring> alpn_protocols;
+
+        // \brief Time to wait for correct session wrap-up
+        // If set to zero, the TLS-level closing is not performed, the
+        // connection is just aborted
+        std::chrono::seconds bye_timeout = std::chrono::seconds(10);
+
+        // \brief Whether or not to pick up any unread data during shutdown
+        // When false and any data arrives at the socket, the connection is
+        // immediately aborted without waiting for graceful wrap-up
+        bool wait_for_data_on_shutdown = false;
     };
 
     /**
--- a/tests/unit/tls_test.cc
+++ b/tests/unit/tls_test.cc
@@ -1575,7 +1575,7 @@ SEASTAR_THREAD_TEST_CASE(test_skip_wait_for_eof) {
         auto sa = server.accept();
         auto c = engine().connect(addr).get();
         auto c_tls = tls::wrap_client(creds, std::move(c),
-                                      tls::tls_options{.wait_for_eof_on_shutdown = false}).get();
+                                      tls::tls_options{.bye_timeout = std::chrono::seconds(0)}).get();
         auto s = sa.get();
 
         auto in = s.connection.input();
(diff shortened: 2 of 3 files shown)


tests/unit/tls_test.cc
```diff
@@ -2097,3 +2097,56 @@ SEASTAR_THREAD_TEST_CASE(test_send_recv_alloc_limits) {
...
+            try {
+                out.write(buf.get(), buf.size()).get();
+                out.flush().get();
+                bytes_sent += buf.size();
+            } catch (...) {
+                break;
+            }
+        }
+        auto delay = std::chrono::duration_cast<std::chrono::duration<double>>(std::chrono::steady_clock::now() - start);
+        fmt::print("Wrote {} bytes in {:.3f} seconds\n", bytes_sent, delay.count());
```
elcallio: `BOOST_TEST_MESSAGE`?

tests/unit/tls_test.cc
```diff
@@ -2097,3 +2097,56 @@ SEASTAR_THREAD_TEST_CASE(test_send_recv_alloc_limits) {
...
+        out.close().handle_exception([] (auto x) {}).get();
+        s.shutdown_input();
+        BOOST_CHECK_LT(delay.count(), 1.0);
```
elcallio, line 2148: This seems like a great source of flakey failures...

elcallio: Maybe record a timestamp on the sender side and compare the time elapsed between closer and closee? Though still subject to any weird delay in the network... (though 1s delay even on an overloaded CI machine is unlikely...)

elcallio (approved):
A few comment, but looks ok.

mykaul:
> Next, here's the second option called wait_for_data_on_shutdown. After BYE message is sent, if the option is false (which it is by default), the session waits for any value resolved from the do_get() method's future. The value can be a buffer with data or an empty buffer indicating an EOF. In any case, after an event the transport socket is closed.

It waits if the option is false?

tests/unit/tls_test.cc
xemul: Agree about potential flakiness :(
I was inspired by test_skip_wait_for_eof's timeout of 1s to measure similar thing

xemul:
> It waits if the option is false?

It waits in any case, but if the option is false, it stops waiting once the first buffer arrives, which should mean EOF. Then is closes the underlying socket. If the buffer is not empty, then it means that the data wasn't read beforehand and TLS wrap up will be terminated once the underlying socket is closed

tests/unit/tls_test.cc
elcallio: Yeah, that one is not perfect either...

tests/unit/tls_test.cc
xemul: Good point (though I always forget how to make them visible on successful runs)

xemul:
upd:
- use BOOST_TEST_MESSAGE
- fix compilation for C++20

tests/unit/tls_test.cc
xemul: > Maybe record a timestamp on the sender side and compare the time elapsed between closer and closee?

Closing happens in the background, I cannot capture a time when it finished without interfering with src/net/tls.cc somehow

elcallio: No, but you can record when the other side calls "close", and count from there. At least keeps the potential weird delays to the actual wire time from then.

include/seastar/net/tls.hh
```diff
@@ -366,8 +366,21 @@ namespace tls {
...
+            bool _value = true;
+            deprecated_wait_for_eof_on_shutdown() noexcept : _value(true) {}
+            [[deprecated("Use tls::options::bye_timeout instead")]]
+            deprecated_wait_for_eof_on_shutdown(bool value) noexcept : _value(value) {}
+            [[deprecated("Use tls::options::bye_timeout instead")]]
+            void operator=(bool value) noexcept { _value = value; }
+            [[deprecated("Use tls::options::bye_timeout instead")]]
+            operator bool() const noexcept { return _value; }
+        };
+
```
avikivity, line 380: Can it be used if it's private? I guess it can't be named but the fields can be used.

include/seastar/net/tls.hh
```diff
@@ -382,6 +395,16 @@ namespace tls {
         /// \brief Optional list of ALPN protocols to offer to the server,
         /// in order of preference.
         std::vector<sstring> alpn_protocols;
+
+        // \brief Time to wait for correct session wrap-up
+        // If set to zero, the TLS-level closing is not performed, the
+        // connection is just aborted
+        std::chrono::seconds bye_timeout = std::chrono::seconds(10);
+
```
avikivity, line 403: Should we allow adjusting it?

If people just set it to zero, it will evoke errors on the other end. I don't know if that can cause problems or not.

avikivity: I meant, adjusting after the connection is established, so it's set to 0 only if we want to abandon the connection. Not sure.

include/seastar/net/tls.hh
xemul: It seems it can be used, at least g++ and clang++ from Fedora-42 didn't object

include/seastar/net/tls.hh
xemul: > Should we allow adjusting it?

Maybe as a followup `connected_socket::set_linger()` method or alike. It would change that value for tls socket and tune SO_LINGER option for tcp one.

> If people just set it to zero, it will evoke errors on the other end. I don't know if that can cause problems or not.
> I meant, adjusting after the connection is established, so it's set to 0 only if we want to abandon the connection. 

Setting it to zero is the replacement for `wait_for_eof_on_shutdown=false` which is not the default behavior. It was (7164ae143e0d0493a1c8bd91b45e9c0c87ab5deb) a workaround against badly implemented servers, and this zero timeout works similarly.

For our needs, the intent is to avoid _reading data_ that might have been left unread by the time socket is closed. For that purpose keeping this value of 10s (and keeping default wait_for_data_on_shutdown=false as well) just works.

xemul: Like this.

Good http connection:
- client sends request
- server sends response
- client reads response fully and places connected socket in a pool
- ...
- client recycles the socket and closes it
- tls code sends BYE, waits for EOF, receives it and wraps up gracefully

Bad http connection:
- client sends request
- server sends response with a lot of data
- client reads response, _part_ of the data and decides to close the connection early
- tls code sends BYE, waits for EOF, but receives more data and just closes the underlying socket (graceful shutdown is broken, but there's no other way)

Both scenarios work with default TLS options and no extra kicks from the http client side

Issue #2974, opened 2025-09-10
RWF_NOWAIT breakage comes unnoticed


xemul:
The flag in question was disabled (e67572ef68a6cd1f0) and then enabled back again (487d04ee7d48f26881fa6896a8bcb7e22564eaa1). The "modern" kernels are believed to have it working, and the flag is enabled on fresh enough kernels. However, sometimes it breaks and it comes unnoticed, reactor just re-submits the requests again and again in the thread-pool.

Need to at least export the re-submission counters to notice that or (better) qualify FS on start and disable the flag dynamically

StephanDollberg:
This is getting worse on newer kernels in general as per our discoveries.

There is some more background on this here: https://github.com/axboe/liburing/issues/1184

travisdowns:
Define "broken": as far as I know, RWF_NOWAIT is behaving within its contract if it returns EAGAIN even in some cases where it may not block. This often happens because the place the decision can be made to return this value is a bit too high up in the stack to be _certain_ that blocking will occur, so it just returns this when it will call down into functions that may block (e.g., which take a lock).

xemul:
Yes, you're right, I'm not clear enough. In my experiments I see 100% of iocbs submitted via reactor thread, immediately got EAGAIN and get resubmitted via thread-pool with RWF_NOWAIT cleared. 

If I disable RWF_NOWAIT, no EAGAINs happen at all and 100% submissions happen from the reactor thread (the latter is sort of expected, of course)

avikivity:
> Define "broken": as far as I know, RWF_NOWAIT is behaving within its contract if it returns EAGAIN even in some cases where it may not block. This often happens because the place the decision can be made to return this value is a bit too high up in the stack to be _certain_ that blocking will occur, so it just returns this when it will call down into functions that may block (e.g., which take a lock).

If it happens once a day, okay. If it happens all the time, it's functionally broken. aio exists to do I/O synchronously, not to tell us that it can't and to go use a thread.

But we also have a responsibility that is to submit aios that are very likely to work.

avikivity:
If the kernel is regressing, it won't magically fix itself. The number of people who care about this and actually measure it is quite small. io_uring actually makes it worse, because io_uring covers up  lower layer regressions by automatically moving requests to a worker, so everything still seems to work.

xemul:
> If it happens all the time, it's functionally broken. 

It had been happening for 2 days in a raw in the "all day long" mode :)

[...]

PR #2987, opened 2025-09-17
Handle nowait support for reads and writes independently

xemul:
Whether or not to set RWF_NOWAIT flag for iocb is controlled by the nowait_works bit that travels a long journey from reactor option via fs_info, then file, then struct request (or more exactly -- one of its united sub-structures). Nowadays this bit is boolean, but we've seen that it might be ultimately broken for writes and work for reads. This PR turns the nowait_works bit on file and fs_info to be a tri-state enum. The request struct still carries it as bit (because they are independent for reads and writes) as well as reactor option remains boolean.

"By default" (i.e. -- on modern kernels with no --linux-aio-nowait option) the mode activates to read_only for XFS. As a result, reads are submitted with the flag and writes are submitted without it. With that we trade guaranteed write-retry for potential delay in the kernel submitting it. Hopefully, chances for the latter are low enough.

refs: #2974

### DIFF
--- a/src/core/file-impl.hh
+++ b/src/core/file-impl.hh
@@ -40,7 +40,8 @@ class io_sink;
 
 }
 
-enum class nowait_mode { yes, no };
+enum class nowait_mode { yes, no, read_only };
+
 template <typename FileImpl>
 class posix_file_handle_impl : public seastar::file_handle_impl {
     int _fd;
--- a/src/core/file-impl.hh
+++ b/src/core/file-impl.hh
@@ -40,6 +40,7 @@ class io_sink;
 
 }
 
+enum class nowait_mode { yes, no };
 template <typename FileImpl>
 class posix_file_handle_impl : public seastar::file_handle_impl {
     int _fd;
@@ -50,7 +51,7 @@ class posix_file_handle_impl : public seastar::file_handle_impl {
     uint32_t _disk_read_dma_alignment;
     uint32_t _disk_write_dma_alignment;
     uint32_t _disk_overwrite_dma_alignment;
-    bool _nowait_works;
+    const nowait_mode _nowait_works;
     bool _durable;
     bool _aio_fdatasync;
 public:
@@ -59,7 +60,7 @@ public:
             uint32_t disk_read_dma_alignment,
             uint32_t disk_write_dma_alignment,
             uint32_t disk_overwrite_dma_alignment,
-            bool nowait_works, bool durable, bool aio_fdatasync)
+            nowait_mode nowait_works, bool durable, bool aio_fdatasync)
             : _fd(fd), _refcount(refcount), _device_id(device_id), _open_flags(f)
             , _memory_dma_alignment(memory_dma_alignment)
             , _disk_read_dma_alignment(disk_read_dma_alignment)
@@ -79,7 +80,7 @@ public:
 
 class posix_file_impl : public file_impl {
     std::atomic<unsigned>* _refcount = nullptr;
-    const bool _nowait_works;
+    const nowait_mode _nowait_works;
     const bool _durable;
     const bool _aio_fdatasync;
     const dev_t _device_id;
@@ -94,7 +95,7 @@ protected:
             uint32_t disk_read_dma_alignment,
             uint32_t disk_write_dma_alignment,
             uint32_t disk_overwrite_dma_alignment,
-            bool nowait_works, bool durable, bool aio_fdatasync);
+            nowait_mode nowait_works, bool durable, bool aio_fdatasync);
 public:
     virtual ~posix_file_impl() override;
     future<> flush() noexcept override;
@@ -173,7 +174,7 @@ public:
     posix_file_real_impl(int fd, open_flags of, file_open_options options, const internal::fs_info& fsi, dev_t device_id)
         : posix_file_impl(fd, of, std::move(options), device_id, fsi) {}
     posix_file_real_impl(int fd, open_flags of, std::atomic<unsigned>* refcount, dev_t device_id,
-            uint32_t memory_dma_alignment, uint32_t disk_read_dma_alignment, uint32_t disk_write_dma_alignment, uint32_t disk_overwrite_dma_alignment, bool nowait_works, bool durable, bool aio_fdatasync)
+            uint32_t memory_dma_alignment, uint32_t disk_read_dma_alignment, uint32_t disk_write_dma_alignment, uint32_t disk_overwrite_dma_alignment, nowait_mode nowait_works, bool durable, bool aio_fdatasync)
         : posix_file_impl(fd, of, refcount, device_id, memory_dma_alignment, disk_read_dma_alignment, disk_write_dma_alignment, disk_overwrite_dma_alignment, nowait_works, durable, aio_fdatasync) {}
     virtual future<size_t> read_dma(uint64_t pos, void* buffer, size_t len, io_intent* intent) noexcept override;
     virtual future<size_t> read_dma(uint64_t pos, std::vector<iovec> iov, io_intent* intent) noexcept override;
@@ -294,7 +295,7 @@ public:
     blockdev_file_impl(int fd, open_flags f, file_open_options options, const internal::fs_info& fsi, dev_t device_id)
         : posix_file_impl(fd, f, options, device_id, fsi) {}
     blockdev_file_impl(int fd, open_flags of, std::atomic<unsigned>* refcount, dev_t device_id,
-            uint32_t memory_dma_alignment, uint32_t disk_read_dma_alignment, uint32_t disk_write_dma_alignment, uint32_t disk_overwrite_dma_alignment, bool nowait_works, bool durable, bool aio_fdatasync)
+            uint32_t memory_dma_alignment, uint32_t disk_read_dma_alignment, uint32_t disk_write_dma_alignment, uint32_t disk_overwrite_dma_alignment, nowait_mode nowait_works, bool durable, bool aio_fdatasync)
         : posix_file_impl(fd, of, refcount, device_id, memory_dma_alignment, disk_read_dma_alignment, disk_write_dma_alignment, disk_overwrite_dma_alignment, nowait_works, durable, aio_fdatasync) {}
 
     future<> truncate(uint64_t length) noexcept override;
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -133,7 +133,7 @@ file_handle::to_file() && {
 }
 
 posix_file_impl::posix_file_impl(int fd, open_flags f, file_open_options options, dev_t device_id, const internal::fs_info& fsi)
-        : _nowait_works(fsi.nowait_works)
+        : _nowait_works(fsi.nowait_works ? nowait_mode::yes : nowait_mode::no)
         , _durable(options.durable)
         , _aio_fdatasync(engine().have_aio_fdatasync())
         , _device_id(device_id)
@@ -189,7 +189,7 @@ posix_file_impl::posix_file_impl(int fd, open_flags f, std::atomic<unsigned>* re
         uint32_t disk_read_dma_alignment,
         uint32_t disk_write_dma_alignment,
         uint32_t disk_overwrite_dma_alignment,
-        bool nowait_works, bool durable, bool aio_fdatasync)
+        nowait_mode nowait_works, bool durable, bool aio_fdatasync)
         : _refcount(refcount)
         , _nowait_works(nowait_works)
         , _durable(durable)
@@ -566,27 +566,27 @@ posix_file_impl::list_directory(std::function<future<> (directory_entry de)> nex
 
 future<size_t>
 posix_file_impl::do_write_dma(uint64_t pos, const void* buffer, size_t len, io_intent* intent) noexcept {
-    auto req = internal::io_request::make_write(_fd, pos, buffer, len, _nowait_works);
+    auto req = internal::io_request::make_write(_fd, pos, buffer, len, _nowait_works == nowait_mode::yes);
     return _io_queue.submit_io_write(len, std::move(req), intent);
 }
 
 future<size_t>
 posix_file_impl::do_write_dma(uint64_t pos, std::vector<iovec> iov, io_intent* intent) noexcept {
     auto len = internal::sanitize_iovecs(iov, _disk_write_dma_alignment);
-    auto req = internal::io_request::make_writev(_fd, pos, iov, _nowait_works);
+    auto req = internal::io_request::make_writev(_fd, pos, iov, _nowait_works == nowait_mode::yes);
     return _io_queue.submit_io_write(len, std::move(req), intent, std::move(iov));
 }
 
 future<size_t>
 posix_file_impl::do_read_dma(uint64_t pos, void* buffer, size_t len, io_intent* intent) noexcept {
-    auto req = internal::io_request::make_read(_fd, pos, buffer, len, _nowait_works);
+    auto req = internal::io_request::make_read(_fd, pos, buffer, len, _nowait_works == nowait_mode::yes);
     return _io_queue.submit_io_read(len, std::move(req), intent);
 }
 
 future<size_t>
 posix_file_impl::do_read_dma(uint64_t pos, std::vector<iovec> iov, io_intent* intent) noexcept {
     auto len = internal::sanitize_iovecs(iov, _disk_read_dma_alignment);
-    auto req = internal::io_request::make_readv(_fd, pos, iov, _nowait_works);
+    auto req = internal::io_request::make_readv(_fd, pos, iov, _nowait_works == nowait_mode::yes);
     return _io_queue.submit_io_read(len, std::move(req), intent, std::move(iov));
 }
 
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -579,14 +579,14 @@ posix_file_impl::do_write_dma(uint64_t pos, std::vector<iovec> iov, io_intent* i
 
 future<size_t>
 posix_file_impl::do_read_dma(uint64_t pos, void* buffer, size_t len, io_intent* intent) noexcept {
-    auto req = internal::io_request::make_read(_fd, pos, buffer, len, _nowait_works == nowait_mode::yes);
+    auto req = internal::io_request::make_read(_fd, pos, buffer, len, _nowait_works == nowait_mode::yes || _nowait_works == nowait_mode::read_only);
     return _io_queue.submit_io_read(len, std::move(req), intent);
 }
 
 future<size_t>
 posix_file_impl::do_read_dma(uint64_t pos, std::vector<iovec> iov, io_intent* intent) noexcept {
     auto len = internal::sanitize_iovecs(iov, _disk_read_dma_alignment);
-    auto req = internal::io_request::make_readv(_fd, pos, iov, _nowait_works == nowait_mode::yes);
+    auto req = internal::io_request::make_readv(_fd, pos, iov, _nowait_works == nowait_mode::yes || _nowait_works == nowait_mode::read_only);
     return _io_queue.submit_io_read(len, std::move(req), intent, std::move(iov));
 }
 
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -711,6 +711,37 @@ static bool blockdev_nowait_works(dev_t device_id) {
     return blockdev_gen_nowait_works;
 }
 
+static nowait_mode filesystem_nowait_mode(bool fs_capable, std::optional<bool> cfg_override) {
+    if (!fs_capable) {
+        return nowait_mode::no;
+    }
+    if (cfg_override.has_value()) {
+        return *cfg_override ? nowait_mode::yes : nowait_mode::no;
+    }
+
+    // First, the nowait became useable in 4.13, see
+    // https://lore.kernel.org/linux-xfs/20210117213401.GB78941@dread.disaster.area/
+    // and seastar commit 487d04ee (file, reactor: reinstate RWF_NOWAIT support)
+    //
+    // Then it was (un)intentionally broken by Linux-6.0 commit 66fa3ced (fs: Add
+    // async write file modification handling) so that lots of writes hit the need
+    // to update cmtimes for an inode and returned EAGAIN seeing the nowait flag.
+    // The change effectively allowed only read-only nowait AIO
+    //
+    // In Linux-7.0 lazytime mode cmtime update was patched to work nicely with the
+    // nowait flag, see 77ef2c3f (re-enable IOCB_NOWAIT writes to files v6)
+
+    if (internal::kernel_uname().whitelisted({"7.0"})) {
+        return nowait_mode::yes;
+    } else if (internal::kernel_uname().whitelisted({"6.0"})) {
+        return nowait_mode::read_only; // seastar issue #2974
+    } else if (internal::kernel_uname().whitelisted({"4.13"})) {
+        return nowait_mode::yes;
+    } else {
+        return nowait_mode::no;
+    }
+}
+
 future<>
 blockdev_file_impl::truncate(uint64_t length) noexcept {
     return make_ready_future<>();
@@ -1351,11 +1382,8 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
                 fsi.append_concurrency = 0;
                 fsi.fsync_is_exclusive = true;
             }
-            if (fs_nowait_works && engine()._cfg.aio_nowait_works) {
-                fsi.nowait_works = nowait_mode::yes;
-            } else {
-                fsi.nowait_works = nowait_mode::no;
-            }
+
+            fsi.nowait_works = filesystem_nowait_mode(fs_nowait_works, engine()._cfg.aio_nowait_works);
             fsi.align = filesystem_alignments(fd, st.st_dev, fsi.block_size, sfs.f_type);
             s_fstype.insert(std::make_pair(st.st_dev, std::move(fsi)));
             return make_file_impl(fd, std::move(options), flags, std::move(st));
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -100,7 +100,7 @@ struct fs_info {
     bool append_challenged;
     unsigned append_concurrency;
     bool fsync_is_exclusive;
-    bool nowait_works;
+    nowait_mode nowait_works;
     std::optional<alignments> align;
 };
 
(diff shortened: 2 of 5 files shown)


xemul:
For the record:
Even after the kernel gets fixed (though we [were asked](https://marc.info/?l=linux-fsdevel&m=175812231630932&w=2) to move longer route), the fix will probably only work for lazytime-mounted partitions, thus the decision how to submit writes -- via thread-pool or without the nowait flag -- will partially (see scylladb/scylladb#26002) remain open

xemul:
@avikivity , please review

xemul:
@avikivity , please review

avikivity:
What's the goal? Why would one enable nowait for one and not the other?

Is it because we think nowait is misreported for writes but not reads?

xemul:
It's not that someone enables or disables it, this "switch" is not controlled by any new option. The old option to enable/disable RWF_NOWAIT altogether doesn't change its behavior. The change here is only about automatic selection of whether or not seastar will use this bit on iocbs. And the suggestion is -- reads can be submitted with nowait and writes not. Or is your question -- why should we submit writes without nowait at all?

avikivity:
Yes, why the different behavior for reads and writes?

xemul:
Because nowait writes have many chances to instantly EAGAIN and it's greatly out of our control. Major issue [is cmtime update](https://github.com/scylladb/seastar/issues/2974#issuecomment-3281849236), someone (us?) have to fix the kernel. Two options
1. lazytime vs nowait handling ([patch](https://marc.info/?l=linux-fsdevel&m=175768745515859&w=2))
2. O_NOCMTIME flag ([patch](https://marc.info/?l=linux-fsdevel&m=175948382030502&w=2))

Both are turning to be rather long journey, then wait for the fixes to reach cloud instances. And seastar needs to live with it somehow

Other than cmtime update, XFS files [are still](https://github.com/scylladb/seastar/issues/2974#issuecomment-3361281849) append-challenged, truncating it ahead of writes doesn't help at all, extent-allocation-size-hint helps partially. But that's, probably, minor

xemul:
Running a "restore from backup" test with #2975 and one more metrics.
Here are AIO statistics that shows 100% retry rate:

image

I also added a metrics to show the number of context switches [1]. For now it looks like this

image

but no comments about it yet. My plan is to re-run the very same test with explicit ban of RWF_NOWAIT for writes and check if context switches grow up.

[1]
```patch
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2543,6 +2543,11 @@ void reactor::register_metrics() {
             io_fallback_counter("file_operation", internal::thread_pool_submit_reason::file_operation),
             // total_operations value:DERIVE:0:U
             io_fallback_counter("process_operation", internal::thread_pool_submit_reason::process_operation),
+            sm::make_counter("ru_nvcsw", sm::description("Number of voluntary context switches"), [] {
+                struct ::rusage ru;
+                ::getrusage(RUSAGE_THREAD, &ru);
+                return ru.ru_nvcsw;
+            })
     });
 
     _metric_groups.add_group("memory", {
```

avikivity:
Yes, but how does it help to disable RWF_NOWAIT? Instead of a retry, you'll stall the reactor.

I can understand this logic: the write EAGAINs are bogus and the request without RWF_NOWAIT will succeed without stalling. But is this the case?

xemul:
> Yes, but how does it help to disable RWF_NOWAIT? Instead of a retry, you'll stall the reactor.

I'm collecting the data to see if we really do. I agree that we trade guaranteed retry for potential stall though. 

> I can understand this logic: the write EAGAINs are bogus and the request without RWF_NOWAIT will succeed without stalling. But is this the case?

Almost. Request without RWF_NOWAIT will _likely_ succeed without stalling, that's my point, and once I have some stats, I'll post it here.

And also even when cmtime update is fixed, we still need to live somehow until the fix propagates to cloud instances. Maybe the way to go in such cases is to route all writes into thread-pool retry right at once, without waiting for EAGAIN from the kernel?

xemul:
With RWF_NOWAIT being explicitly cleared for write AIOs:

image

image

avikivity:
I think it's not necessary. RWF_NOWAIT is mostly for writes (I guess a file that doesn't have its mapping in memory yet would need it for reads). If it doesn't work for writes, we should just disable it rather than introduce yet another switch.

We should blacklist known-bad kernels so users don't have to worry about it.

xemul:
> If it doesn't work for writes, we should just disable it rather than introduce yet another switch.
> We should blacklist known-bad kernels so users don't have to worry about it.

That's what this PR [does](https://github.com/scylladb/seastar/pull/2987/commits/09e7cb7b950d0560296dd9e05f8588db17861a92):

```patch
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -1122,10 +1122,22 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
                 fsi.fsync_is_exclusive = true;
             }
 
-            if (fs_nowait_works && engine()._cfg.aio_nowait_works) {
-                fsi.nowait_works = nowait_mode::yes;
-            } else {
+            if (!fs_nowait_works) {
                 fsi.nowait_works = nowait_mode::no;
+            } else if (engine()._cfg.aio_nowait_works.has_value()) {
+                if (*engine()._cfg.aio_nowait_works) {
+                    fsi.nowait_works = nowait_mode::yes;
+                } else {
+                    fsi.nowait_works = nowait_mode::no;
+                }
+            } else {
+                if (internal::kernel_uname().whitelisted({"6.0"})) {
+                    fsi.nowait_works = nowait_mode::read_only; // seastar issue #2974
+                } else if (internal::kernel_uname().whitelisted({"4.13"})) {
+                    fsi.nowait_works = nowait_mode::yes;
+                } else {
+                    fsi.nowait_works = nowait_mode::no;
+                }
             }
             s_fstype.insert(std::make_pair(st.st_dev, std::move(fsi)));
             return make_file_impl(fd, std::move(options), flags, std::move(st));
```

no new user-visible switches, it just detects how the kernel can do RWF_NOWAIT for reads or writes of both and acts respectively

xemul:
upd:
- rebased to fix conflicts with other PRs merged so far (#2980 and #3045)

travisdowns:
> With that we trade guaranteed write-retry for potential delay in the kernel submitting it.

This isn't correct, unless I'm reading it wrong: only "some" writes retry, some still go through with this flag. The exact numbers depend on the workload but in practice we see retry rates of 5%-20% in our workloads.

avikivity:
> > With that we trade guaranteed write-retry for potential delay in the kernel submitting it.
> 
> This isn't correct, unless I'm reading it wrong: only "some" writes retry, some still go through with this flag. The exact numbers depend on the workload but in practice we see retry rates of 5%-20% in our workloads.

io_uring should handle those retries in-kernel, so more efficiently.

20% is quite high. Are you setting the extent size hint for allocating writes?

travisdowns:
> > Yes, but how does it help to disable RWF_NOWAIT? Instead of a retry, you'll stall the reactor.
> 
> I'm collecting the data to see if we really do. I agree that we trade guaranteed retry for potential stall though.

In Redpanda, we definitely do stall if you remove this flag. I.e., EAGAIN is returned for both legitimate "will stall" situations and well as "actually we wouldn't stall, but the kernel couldn't figure that out but returns EAGAIN conservatively". The cases of the latter have increased substantially in recent kernels.

So disabling this will introduce real stalls (e.g., during XFS block allocation), but maybe it's worth it. I was talking to @ballard26 today and he had done some testing on this, and did see true stalls introduced when the flag is dropped, but in this test it did still work out overall better (i.e., the pain of false positives was worse than the pain of missing true stalls, for that workload).

travisdowns:
> io_uring should handle those retries in-kernel, 

Are you using io_uring yet? We are using aio, and this thread is only relevant to aio.

> so more efficiently.

To some extent. It also means the problem is hidden. I think real problems still exist even with the kernel thread: e.g., seastar wants to hot poll, when it has no work, but that's the wrong thing to do when writes have been shunted off to the "syscall" thread, since it is being starved (at least until the end of the quantum/waves hands), and this is true also of kernel threads.

> 
> 20% is quite high. Are you setting the extent size hint for allocating writes?

We don't use allocating writes: we use explicit fallocate, so in practice we are allocating large extents (32 MB). So that's not the problem. As discussed previously the problem occurs even if you do that right because:

1) mtime needs to be udpated (sometimes), which causes EAGAIN on newer kernels
2) unwritten extent conversion requires a bit of extra work in XFS which I think also causes EAGAIN sometimes

There are a few other sources too, but (1) is the big one and the one that changed. So you will get EAGAIN 100 to 1000 times per second on a given file if you are writing it frequently as mtime needs to be udpated that often. It's not 100% only because mtime is only updated at "HZ" granularity.

xemul:
> 20% is quite high. Are you setting the extent size hint for allocating writes?

Yes. But without recent hch's VFS+XFS patches it still hits the need to update inodes write timestamp

xemul:
>> With that we trade guaranteed write-retry for potential delay in the kernel submitting it.
>>
> This isn't correct, unless I'm reading it wrong: only "some" writes retry, some still go through with this flag

Yes, it was too restrictive to say "guaranteed". Some writes still come through

xemul:
> So disabling this will introduce real stalls (e.g., during XFS block allocation), but maybe it's worth it.

It's workload-dependent and is indeed "maybe". With the RWF_NOWAIT flag we have X% of writes EAGAIN-ed and retried via thread-pool and (1-X)% work. Without it we have Y% of writes stall in the kernel and (1-Y)% work nicely. And measuring (whatever it means) how X-vs-Y affects the final latency distributions is how to make the decision whether or not to use it or not

avikivity:
> > io_uring should handle those retries in-kernel,
> 
> Are you using io_uring yet? We are using aio, and this thread is only relevant to aio.

We are also using linux-aio.

The underlying issues are the same. If you have many aio fallbacks with linux-aio, you will have many with io_uring, but they will be handled silently and efficiently. The silently part is actually a problem because you won't be aware of the issue.

> 
> > so more efficiently.
> 
> To some extent. It also means the problem is hidden.

Exactly, why are you repeating me?!

>  I think real problems still exist even with the kernel thread: e.g., seastar wants to hot poll, when it has no work, but that's the wrong thing to do when writes have been shunted off to the "syscall" thread, since it is being starved (at least until the end of the quantum/waves hands), and this is true also of kernel threads.

Right. Maybe it will improve with the asymmetric backend (so far we have bad results with it).

> 
> > 20% is quite high. Are you setting the extent size hint for allocating writes?
> 
> We don't use allocating writes: we use explicit fallocate, so in practice we are allocating large extents (32 MB). So that's not the problem. As discussed previously the problem occurs even if you do that right because:
> 
> 1. mtime needs to be udpated (sometimes), which causes EAGAIN on newer kernels

This was fixed recently (we sponsored the fix).


> 2. unwritten extent conversion requires a bit of extra work in XFS which I think also causes EAGAIN sometimes


I thought this happens prinicipally on fdatasync().

> 
> There are a few other sources too, but (1) is the big one and the one that changed. So you will get EAGAIN 100 to 1000 times per second on a given file if you are writing it frequently as mtime needs to be udpated that often. It's not 100% only because mtime is only updated at "HZ" granularity.

Perhaps this then:
 - detect the range of bad kernels
 - for those bad kernels, set mtime in the future
 - move it back to current time just before fdatasync

Though there may be negative side effects to moving mtime around (with backup tools etc, not the I/O machinery).

avikivity:
> > So disabling this will introduce real stalls (e.g., during XFS block allocation), but maybe it's worth it.
> 
> It's workload-dependent and is indeed "maybe". With the RWF_NOWAIT flag we have X% of writes EAGAIN-ed and retried via thread-pool and (1-X)% work. Without it we have Y% of writes stall in the kernel and (1-Y)% work nicely. And measuring (whatever it means) how X-vs-Y affects the final latency distributions is how to make the decision whether or not to use it or not

And RWF_NOWAIT amplifies X vs Y?

avikivity:
The fix for the mtime updates causing EAGAIN is torvalds/linux@08489c4f41333913c9be27a031b070f4452e9374, in 7.0. I thought it was committed earlier.

xemul:
> And RWF_NOWAIT amplifies X vs Y?

Yes, on "buggy" kernels with intense workload amplifies significantly

avikivity:
> > And RWF_NOWAIT amplifies X vs Y?
> 
> Yes, on "buggy" kernels with intense workload amplifies significantly

Ok. So we can either disable RWF_NOWAIT for writes on bad kernels, or do the future mtime trick (if it works and doesn't cause other problems).

avikivity:
The mtime trick will need extra care for O_DSYNC files, since they're never fdatasync()ed explicitly.

xemul:
> ... we can either disable RWF_NOWAIT for writes on bad kernels ...

Yes, and this is what this PR does (it now needs rebase)

> ... do the future mtime trick (if it works and doesn't cause other problems).

How about ctime? IIRC it caused EAGAIN-s too, but it's not changeable

avikivity:
> > ... we can either disable RWF_NOWAIT for writes on bad kernels ...
> 
> Yes, and this is what this PR does (it now needs rebase)
> 

Yes, now I understand it.

> > ... do the future mtime trick (if it works and doesn't cause other problems).
> 
> How about ctime? IIRC it caused EAGAIN-s too, but it's not changeable

Please check with 7.0.

xemul:
> The fix for the mtime updates causing EAGAIN is [torvalds/linux@08489c4](https://github.com/torvalds/linux/commit/08489c4f41333913c9be27a031b070f4452e9374), in 7.0. I thought it was committed earlier.

Not only this, it oftent stepped on generic inode update code, but the whole PR matters https://github.com/torvalds/linux/commit/77ef2c3ff5916d358c436911ca6a961060709f04 . It's still 7.0 though

xemul:
>> How about ctime? IIRC it caused EAGAIN-s too, but it's not changeable
>>
> Please check with 7.0.

7.0 have the lazytime fix that works for both, C and M times. My comment was about the "do mtime tricks" suggestion -- we can do it, but ctime will still trigger EAGAIN-s on buggy kernels:

```c
static int file_update_time_flags(struct file *file, unsigned int flags)
{
    struct inode *inode = file_inode(file);
    struct timespec64 now, ts;
    int sync_mode = 0;
    int ret = 0;

    /* First try to exhaust all avenues to not sync */
    if (IS_NOCMTIME(inode))
        return 0;
    if (unlikely(file->f_mode & FMODE_NOCMTIME))
        return 0;

    now = current_time(inode);

    ts = inode_get_mtime(inode);
    if (!timespec64_equal(&ts, &now))
        sync_mode |= S_MTIME;
    ts = inode_get_ctime(inode);
    if (!timespec64_equal(&ts, &now))
        sync_mode |= S_CTIME;
    if (IS_I_VERSION(inode) && inode_iversion_need_inc(inode))
        sync_mode |= S_VERSION;

    if (!sync_mode)
        return 0;

    if (flags & IOCB_NOWAIT)
        return -EAGAIN; <------------------- this is our case, since up above both, ctime and mtime, are checked against "now" 
    ...
```

(the above is pre-77ef2c3ff5916d358c436911ca6a961060709f04 commit

avikivity:
> > > How about ctime? IIRC it caused EAGAIN-s too, but it's not changeable
> > 
> > 
> > Please check with 7.0.
> 
> 7.0 have the lazytime fix that works for both, C and M times. My comment was about the "do mtime tricks" suggestion -- we can do it, but ctime will still trigger EAGAIN-s on buggy kernels:
> 
> ```c
> static int file_update_time_flags(struct file *file, unsigned int flags)
> {
>     struct inode *inode = file_inode(file);
>     struct timespec64 now, ts;
>     int sync_mode = 0;
>     int ret = 0;
> 
>     /* First try to exhaust all avenues to not sync */
>     if (IS_NOCMTIME(inode))
>         return 0;
>     if (unlikely(file->f_mode & FMODE_NOCMTIME))
>         return 0;
> 
>     now = current_time(inode);
> 
>     ts = inode_get_mtime(inode);
>     if (!timespec64_equal(&ts, &now))
>         sync_mode |= S_MTIME;
>     ts = inode_get_ctime(inode);
>     if (!timespec64_equal(&ts, &now))
>         sync_mode |= S_CTIME;
>     if (IS_I_VERSION(inode) && inode_iversion_need_inc(inode))
>         sync_mode |= S_VERSION;
> 
>     if (!sync_mode)
>         return 0;
> 
>     if (flags & IOCB_NOWAIT)
>         return -EAGAIN; <------------------- this is our case, since up above both, ctime and mtime, are checked against "now" 
>     ...
> ```
> 
> (the above is pre-77ef2c3ff5916d358c436911ca6a961060709f04 commit

Ok. So it leaves disabling RWF_NOWAIT for writes on bad kernels.

src/core/file.cc
```diff
@@ -1107,10 +1107,22 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
...
+            if (!fs_nowait_works) {
                 fsi.nowait_works = nowait_mode::no;
+            } else if (engine()._cfg.aio_nowait_works.has_value()) {
+                if (*engine()._cfg.aio_nowait_works) {
+                    fsi.nowait_works = nowait_mode::yes;
+                } else {
+                    fsi.nowait_works = nowait_mode::no;
+                }
+            } else {
+                if (internal::kernel_uname().whitelisted({"6.0"})) {
```
avikivity: Need to adjust for 7.0 here.

xemul: Sure

xemul:
upd:
- rebased (had conflicts)
- whitlisted 7.0 for full-nowait mode
- extracted filesystem nowait mode decicion making into a helper function

travisdowns:
> This was fixed recently (we sponsored the fix).

This is huge! This is the [merge](https://github.com/torvalds/linux/commit/77ef2c3ff5916d358c436911ca6a961060709f04), right? So we can expect this to be greatly improved in 7.0 then.

> I thought this happens prinicipally on fdatasync().

UWEC happens during the O_DIRECT write, IIRC in the dio completion callback. One relationship to fdatasync is that UWEC means that a subsequent fdatasync actually has to do an additional write to flush that updated XFS metadata. If no UWEC happened, fdatasync may only need a device cache flush, which is often a no-op (e.g., on AWS local disks).

src/core/file-impl.hh
```diff
@@ -50,7 +52,7 @@ class posix_file_handle_impl : public seastar::file_handle_impl {
     uint32_t _disk_read_dma_alignment;
     uint32_t _disk_write_dma_alignment;
     uint32_t _disk_overwrite_dma_alignment;
-    bool _nowait_works;
+    const nowait_mode _nowait_works;
```
travisdowns, line 55: should this be renamed _nowait_mode?

xemul:
> > This was fixed recently (we sponsored the fix).
> 
> This is huge! This is the [merge](https://github.com/torvalds/linux/commit/77ef2c3ff5916d358c436911ca6a961060709f04), right? 

Yes

> So we can expect this to be greatly improved in 7.0 then.

src/core/file-impl.hh
xemul: Good point

src/core/file.cc
```diff
@@ -711,6 +711,24 @@ static bool blockdev_nowait_works(dev_t device_id) {
...
 }
 
+static nowait_mode filesystem_nowait_mode(bool fs_capable, std::optional<bool> cfg_override) {
+    if (!fs_capable) {
+        return nowait_mode::no;
+    }
+    if (cfg_override.has_value()) {
+        return *cfg_override ? nowait_mode::yes : nowait_mode::no;
+    }
+    if (internal::kernel_uname().whitelisted({"7.0"})) {
```
travisdowns, line 734: Could use some "why" comments here on these specific kernel version ranges listing relevant fixes, etc.

src/core/file-impl.hh
xemul: On the other hand `_nowait_mode == yes` sounds weird compared to `_nowait_works == yes`
Yet `_nowait_works == read_only` sounds ... roughly the same as `_nowait_mode == read_only`
To my non-native-english-speaker ear

src/core/file.cc
```diff
@@ -711,6 +711,24 @@ static bool blockdev_nowait_works(dev_t device_id) {
     return blockdev_gen_nowait_works;
 }
 
+static nowait_mode filesystem_nowait_mode(bool fs_capable, std::optional<bool> cfg_override) {
```
travisdowns, line 714: Could use a basic doc above or below.

src/core/file.cc
xemul: Agree, worth it.
For the record -- it was broken in 6.0 with https://github.com/torvalds/linux/commit/66fa3cedf16abc82d19b943e3289c82e685419d5 and was fixed in 7.0 with https://github.com/torvalds/linux/commit/77ef2c3ff5916d358c436911ca6a961060709f04

xemul: And it was pre-fixed in 4.13 with ... :thinking: https://lore.kernel.org/linux-xfs/20210117213401.GB78941@dread.disaster.area/ is the best link I have found so far

avikivity:
> > This was fixed recently (we sponsored the fix).
> 
> This is huge! This is the [merge](https://github.com/torvalds/linux/commit/77ef2c3ff5916d358c436911ca6a961060709f04), right? So we can expect this to be greatly improved in 7.0 then.
> 
> > I thought this happens prinicipally on fdatasync().
> 
> UWEC happens during the O_DIRECT write, IIRC in the dio completion callback.


But that's too late for EAGAIN, no?

Anyway, torvalds/linux@5ec58e6acdecb0c2b1707002123883fe1da29a98 (also from us)

> One relationship to fdatasync is that UWEC means that a subsequent fdatasync actually has to do an additional write to flush that updated XFS metadata. If no UWEC happened, fdatasync may only need a device cache flush, which is often a no-op (e.g., on AWS local disks).

For us, the fdatasync:write ratio is very low so we don't care.

Should aim for O_DSYNC (+ only overwrites), or in-band allocation (unwritten extent conversion is the continuation of allocation by other means) and late fdatasync().

xemul:
upd:
- rebased (no conflicst)
- added code comment to explain the way nowait mode is selected

travisdowns:
@avikivity wrote:

> But that's too late for EAGAIN, no?
> 
> Anyway, [torvalds/linux@5ec58e6](https://github.com/torvalds/linux/commit/5ec58e6acdecb0c2b1707002123883fe1da29a98) (also from us)

Yes, EAGAIN can't come from something that happens in the callback, but the up-front sync code that runs in io_submit can EAGAIN for the same types of cases where it knows a callback will be needed. That I double checked our tickets and using bpf probes @ballard26 had tracked down that 75% of the EAGAINs come from the mtime/inode update path `file_modified_flags` (most/all of which could plausibly be fixed by hch change), and almost all the remaining ones from `iomap_dio_rw` though we didn't drill into the specific root cause of that one, but it was present in similar amounts prior to kernel 6.0. Everything else was < 3% total.

So my comment about UWEC being related to fallbacks may be nonsense.

> Should aim for O_DSYNC (+ only overwrites)


We discusses this before but it is not an obvious win due to our write patterns. 

> or in-band allocation (unwritten extent conversion is the continuation of allocation by other means) and late fdatasync().

This I didn't understand.

We have also thought about pre zero filling but for many workload this effectively doubles the IOPS required, though the IOPS involved are now friendlier (zeroing IOPS are large, and the true payload writes don't have to do UWEC).

avikivity:
> @avikivity wrote:
> 
> > But that's too late for EAGAIN, no?
> > Anyway, [torvalds/linux@5ec58e6](https://github.com/torvalds/linux/commit/5ec58e6acdecb0c2b1707002123883fe1da29a98) (also from us)
> 
> Yes, EAGAIN can't come from something that happens in the callback, but the up-front sync code that runs in io_submit can EAGAIN for the same types of cases where it knows a callback will be needed. That I double checked our tickets and using bpf probes @ballard26 had tracked down that 75% of the EAGAINs come from the mtime/inode update path `file_modified_flags` (most/all of which could plausibly be fixed by hch change), and almost all the remaining ones from `iomap_dio_rw` though we didn't drill into the specific root cause of that one, but it was present in similar amounts prior to kernel 6.0. Everything else was < 3% total.

Internal distribution of EAGAINs doesn't matter so much, what matters is ratio of EAGAINs to successful async issue. If that drops to reasonable levels, (<3%), all is well. If it's high, take the highest contributor and beat it int submission.

> 
> So my comment about UWEC being related to fallbacks may be nonsense.
> 
> > Should aim for O_DSYNC (+ only overwrites)
> 
> We discusses this before but it is not an obvious win due to our write patterns.

Change your write patterns: pre-allocate and reuse files (works well for commitlog patterns).

> 
> > or in-band allocation (unwritten extent conversion is the continuation of allocation by other means) and late fdatasync().

> 
> This I didn't understand.

It works for the cases where you don't care about time-to-durability of an individual write, but about the time-to-durability of the entire file (e.g. compaction). Use large extent hints, O_DIRECT but not O_DSYNC writes, fdatasync() at the end.


> 
> We have also thought about pre zero filling but for many workload this effectively doubles the IOPS required, though the IOPS involved are now friendlier (zeroing IOPS are large, and the true payload writes don't have to do UWEC).

Another alternative is to write directly to the disk, but it's not for the faint of heart.

travisdowns:
> Internal distribution of EAGAINs doesn't matter so much, what matters is ratio of EAGAINs to successful async issue. If that drops to reasonable levels, (<3%), all is well. If it's high, take the highest contributor and beat it int submission.

Sure, but this analysis was with the "bad" levels of EAGAIN, i.e. 10% to 70% (!!) EAGAIN. So I put it there to give an idea of how much this change can help. E.g., if we are getting 50% fallbacks and 75% of those are caused by the mtime issue, then we might expect that after this fix we have only 12.5% fallbacks left (but this ignores the case where a write could have multiple fallback reasons and if you remove the first one it hits it now hits the second). We are still collecting numbers but preliminary results seem to show an substantial improvement but much less than what is suggested by the above. It is far above 3% still.

> Change your write patterns: pre-allocate

I don't think pre-allocating necessarily helps. You probably mean one of:

1) Do fallocation/truncate the file out its full size

Effectively, we do this (32 MB at a time). Doesn't avoid UWEC.

2)  and reuse files (works well for commitlog patterns).

This would work.

> It works for the cases where you don't care about time-to-durability of an individual write, but about the time-to-durability of the entire file (e.g. compaction). Use large extent hints, O_DIRECT but not O_DSYNC writes, fdatasync() at the end.

Usually we care about "time-to-durability of an individual write" but in any case none of " large extent hints, O_DIRECT but not O_DSYNC writes, fdatasync()" avoids UWEC. Only writing to already converted blocks does that.

avikivity:
> 
> > It works for the cases where you don't care about time-to-durability of an individual write, but about the time-to-durability of the entire file (e.g. compaction). Use large extent hints, O_DIRECT but not O_DSYNC writes, fdatasync() at the end.
> 
> Usually we care about "time-to-durability of an individual write" but in any case none of " large extent hints, O_DIRECT but not O_DSYNC writes, fdatasync()" avoids UWEC. Only writing to already converted blocks does that.

But that's the typical commitlog pattern, where pre-formatted files can be reused again and again.

travisdowns:
@avikivity 

> But that's the typical commitlog pattern, where pre-formatted files can be reused again and again.

Big update on this, continued the discuss here: https://github.com/scylladb/seastar/issues/2974#issuecomment-4593449676 instead of having it on this merged PR.

PR #2988, opened 2025-09-19
memory: Call finish_allocation() at the end of allocate_aligned()

xemul:
The helper in questionb does the very same sequence of steps and is already used in other allocating functions.


xemul:

PR #2990, opened 2025-09-22
resource: Tune up some allocate_io_queues() arguments

xemul:
The vector of CPUs can be made const reference instead of value to avoid its copying (although the method is called for each IO queue and there are typically two of them)

The cpu-to-node map is a reference already, but keeping it const ref is safer.


xemul:

PR #2991, opened 2025-09-22
io_queue: Use own logger for messages

xemul:
Some messages from io_queue.cc are printed using seastar_logger, but there's io-queue own logger for that.


xemul:

PR #2993, opened 2025-09-23
io_queue: Destroy priority class data with scheduling group

xemul:
When a scheduling group is destroyed, the IO queue's priority_class_data
that corresponds to it stays intact. It's not actually a leak, as when
the reactor is destroyed all class data objects are collected.

However, if another scheduling group under the same ID is created, the
IO queue will re-use the dangling priority class for the newly created
group, but its name and shares won't match the new group until rename
or shares update. And it's not nice that IO class remains alive anyway.

Since scheduling group destruction must happen after all tasks from it
are completed and picked up, the class destruction code may also assume
that no queued or in-flight requests exist.

Test included.

### DIFF
--- a/include/seastar/core/io_queue.hh
+++ b/include/seastar/core/io_queue.hh
@@ -222,6 +222,7 @@ public:
     void update_shares_for_class_group(unsigned index, size_t new_shares);
     future<> update_bandwidth_for_class(internal::priority_class pc, uint64_t new_bandwidth);
     void rename_priority_class(internal::priority_class pc, sstring new_name);
+    void destroy_priority_class(internal::priority_class pc) noexcept;
     void throttle_priority_class(const priority_class_data& pc) noexcept;
     void unthrottle_priority_class(const priority_class_data& pc) noexcept;
 
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -192,6 +192,11 @@ public:
     priority_class_data(const priority_class_data&) = delete;
     priority_class_data(priority_class_data&&) = delete;
 
+    ~priority_class_data() {
+        SEASTAR_ASSERT(_nr_queued == 0);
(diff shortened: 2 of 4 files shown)


include/seastar/core/io_queue.hh
```diff
@@ -222,6 +222,7 @@ public:
     void update_shares_for_class_group(unsigned index, size_t new_shares);
     future<> update_bandwidth_for_class(internal::priority_class pc, uint64_t new_bandwidth);
```
avikivity, line 223: > Next, priority class data may have executing requests. That's worse, as
they cannot be canceled. The destroying code could wait for them to

How can they have executing requests? The user destroying the scheduling group is supposed to have drained it of any tasks. If there is pending I/O, it is linked to tasks that it will schedule.

xemul: User can (in theory) ignore the future from IO. Not necessarily intentionally, e.g. like this

```c++
    auto f = file.dma_read();
    /* do something else */
    auto result = co_await f;
```

in case "do something else" throws the submitted request is left in the background

avikivity: That's a bug. Mr User should wrap something else in try/catch.

xemul:
upd:
- assume that when priority class is destroyed, no queued or executing requests exist

xemul:
@avikivity , please review

avikivity:
I unmerged it, it needs rebasing after the data_sink API changes.

xemul:
I don't see how it conflicts with data_sink API changes. Maybe it's API changes that need to be unmerged, because they clashed with #3018 's on counting_data_sink? OTOH, API level 8 in "library mode", i.e. without tests, demos and apps, still compiles fine

xemul:
Created #3040

Issue #2992, opened 2025-09-23
Scheduling group renaming leads to `Execution stage ... already exists` error


andrzej-jackowski-scylladb:
Seastar's execution_stage_manager doesn't allow creating two execution stages with the same name at the same time:

From `seastar/src/core/execution_stage.cc`:
```
void execution_stage_manager::register_execution_stage(execution_stage& stage) {
    auto ret = _stages_by_name.emplace(stage.name(), &stage);
    if (!ret.second) {
        throw std::invalid_argument(format("Execution stage {} already exists.", stage.name()));
    }
   ...
}
```

However, the implementation of the inheriting_execution_stage verifies the existence of the execution_stage based on scheduling group index rather than its name:

From `seastar/include/seastar/core/execution_stage.hh`:
```
    return_type operator()(typename internal::wrap_for_es<Args>::type... args) {
        auto sg = current_scheduling_group();
        auto sg_id = internal::scheduling_group_index(sg);
        auto& slot = _stage_for_group[sg_id];
        if (!slot) {
            slot.emplace(make_stage_for_group(sg));
        }
        return (*slot)(std::move(args)...);
    }
```

The following test case implemented by @piodul in https://github.com/scylladb/scylladb/issues/26190#issuecomment-3319933186 reproduces the issue (i.e. `std::invalid_argument: Execution stage my_stage.foo already exists.` is thrown):
```
SEASTAR_THREAD_TEST_CASE(scheduling_group_rename_recreate_with_the_same_name) {
    auto do_nothing = [] {};
    inheriting_concrete_execution_stage<void> stage{"my_stage", do_nothing};

    auto sg1 = create_scheduling_group("foo", 100).get();
    with_scheduling_group(sg1, [&] { return stage(); }).get();

    rename_scheduling_group(sg1, "bar").get();

    auto sg2 = create_scheduling_group("foo", 100).get();
    with_scheduling_group(sg2, [&] { return stage(); }).get();
}
```

andrzej-jackowski-scylladb:
There are two problems here that require fixing:

1.  `execution_stage_manager::register_execution_stage` throws `throw std::invalid_argument(format("Execution stage {} already exists.", stage.name()));`
2. If we workaround `register_execution_stage` to prevent from throwing, `metrics::metric_group` is created in `execution_stage::execution_stage`. If the same label is given twice for the metrics, `double_registration` is thrown with `registering metrics twice for metrics: ...` error from the metrics implementation.

[...]

PR #2994, opened 2025-09-23
execution_stage: update stage name on scheduling_group rename

andrzej-jackowski-scylladb:
Previously, an execution_stage's name was not updated when its scheduling_group was renamed. This caused:
  1. Metric names not being updated, making them inconsistent and confusing.
  2. Since two execution stages with the same name cannot coexist in execution_stage_manager, creating a scheduling group `s1` after it had earlier been renamed to `s2` led to an exception.

This change updates the execution stage's name when its scheduling_group is renamed.

`test_renaming_execution_stage` and `sg_rename_recreate_with_the_same_name` tests are added to reproduce both aforementioned scenarios.

Fixes: https://github.com/scylladb/seastar/issues/2992

### DIFF
--- a/tests/unit/scheduling_group_test.cc
+++ b/tests/unit/scheduling_group_test.cc
@@ -372,6 +372,26 @@ SEASTAR_THREAD_TEST_CASE(sg_create_and_key_create_in_parallel) {
     }
 }
 
+SEASTAR_THREAD_TEST_CASE(sg_rename_recreate_with_the_same_name) {
+    auto do_nothing = [] {};
+    inheriting_concrete_execution_stage<void> stage{"my_stage", do_nothing};
+
+    std::vector<scheduling_group> sgs;
+    sgs.push_back(create_scheduling_group("foo", 100).get());
+    with_scheduling_group(sgs[0], [&] { return stage(); }).get();
+
+    rename_scheduling_group(sgs[0], "bar").get();
+
+    BOOST_CHECK_NO_THROW({
+        sgs.push_back(create_scheduling_group("foo", 100).get());
+        with_scheduling_group(sgs[1], [&] { return stage(); }).get();
+    });
+
+    for (scheduling_group sg : sgs) {
+        destroy_scheduling_group(sg).get();
+    }
+}
+
 SEASTAR_THREAD_TEST_CASE(sg_key_constructor_exception_when_creating_new_key) {
     scheduling_group_key_config key_conf = make_scheduling_group_key_config<int>();
     scheduling_group_key_create(key_conf).get();


andrzej-jackowski-scylladb:
@xemul, @piodul, I really hate the `rename_func_t _rename_func` construction introduced in the commit, but I had no better idea how to do it simply, because `inheriting_concrete_execution_stage` is not a subclass of `execution_stage`. I really look forward to your advice how to implement it better.

andrzej-jackowski-scylladb:
v2:
 - added missing call of `_metric_group.clear()`
 - added `test_execution_stage_rename` that verifies rename() properly (and detects that the clear() was missing)

src/core/execution_stage.cc
```diff
@@ -111,6 +111,13 @@ execution_stage::execution_stage(const sstring& name, scheduling_group sg)
 {
     internal::execution_stage_manager::get().register_execution_stage(*this);
     auto undo = defer([&] () noexcept { internal::execution_stage_manager::get().unregister_execution_stage(*this); });
+    update_metric_group();
+    undo.cancel();
+}
+
+void execution_stage::update_metric_group() {
+    sstring name = _name;
```
nuivall: maybe worth using _name directly below

nuivall: either capture this in lambdas or something like [name = _name]

src/core/execution_stage.cc
```diff
@@ -67,6 +67,20 @@ execution_stage* execution_stage_manager::get_stage(const sstring& name) {
     return _stages_by_name[name];
 }
 
+void execution_stage_manager::update_scheduling_group_name(const sstring& new_sg_name) {
+    for (auto* execution_stage_ptr : _execution_stages) {
```
nuivall: I see in other places we iterate like this: 
`
for (auto&& stage : _execution_stages)
`

src/core/execution_stage.cc
```diff
@@ -67,6 +67,20 @@ execution_stage* execution_stage_manager::get_stage(const sstring& name) {
...
+        if (execution_stage_ptr != nullptr && execution_stage_ptr->get_sg().name() == new_sg_name) {
```
nuivall: ::flush and :pool don't have null checks, maybe it is not possible to contain null stage?

src/core/execution_stage.cc
```diff
@@ -67,6 +67,20 @@ execution_stage* execution_stage_manager::get_stage(const sstring& name) {
...
+            auto old_name = execution_stage_ptr->name();
+            execution_stage_ptr->rename();
+            _stages_by_name.erase(old_name);
+            auto ret = _stages_by_name.emplace(execution_stage_ptr->name(), execution_stage_ptr);
+            if (!ret.second) {
+                throw std::invalid_argument(format("Execution stage {} already exists.", execution_stage_ptr->name()));
+            }
+        }
```
nuivall, line 77: should we stop after finding stage with new_sg_name ? is it many to many relation or one to one?

tests/unit/metrics_test.cc
```diff
@@ -121,6 +123,24 @@ SEASTAR_THREAD_TEST_CASE(test_renaming_scheuling_groups) {
...
+    using namespace seastar;
+    sstring sg_name1 = "sg1";
+    sstring sg_name2 = "sg2";
+    inheriting_concrete_execution_stage<void> stage{"my_stage", []{}};
+    scheduling_group sg = create_scheduling_group(sg_name1, 111).get();
+    with_scheduling_group(sg, [&] { return stage(); }).get();
+
+    std::set<sstring> label_vals = get_label_values(sstring("execution_stages_tasks_scheduled"), sstring("execution_stage"));
+    BOOST_REQUIRE(label_vals.find(sstring("my_stage.") + sstring(sg_name1)) != label_vals.end());
+    rename_scheduling_group(sg, sg_name2).get();
```
ptrsmrn: maybe separate asserts with executing some logic that changes the state by a newline character

tests/unit/metrics_test.cc
```diff
@@ -121,6 +123,24 @@ SEASTAR_THREAD_TEST_CASE(test_renaming_scheuling_groups) {
...
+    with_scheduling_group(sg, [&] { return stage(); }).get();
```
ptrsmrn: is executing `with_scheduling_group` again necessary?

tests/unit/scheduling_group_test.cc
```diff
@@ -372,6 +372,19 @@ SEASTAR_THREAD_TEST_CASE(sg_create_and_key_create_in_parallel) {
...
+SEASTAR_THREAD_TEST_CASE(sg_rename_recreate_with_the_same_name) {
+    auto do_nothing = [] {};
+    inheriting_concrete_execution_stage<void> stage{"my_stage", do_nothing};
+
+    auto sg1 = create_scheduling_group("foo", 100).get();
+    with_scheduling_group(sg1, [&] { return stage(); }).get();
+
+    rename_scheduling_group(sg1, "bar").get();
+
+    auto sg2 = create_scheduling_group("foo", 100).get();
```
ptrsmrn: Consider adding ASSERT_NO_THROW or something like, which would make the error more obvious if the test fails.

tests/unit/scheduling_group_test.cc
```diff
@@ -372,6 +372,19 @@ SEASTAR_THREAD_TEST_CASE(sg_create_and_key_create_in_parallel) {
...
+    with_scheduling_group(sg2, [&] { return stage(); }).get();
```
ptrsmrn: Consider adding explicit asserts here as well.

ptrsmrn (commented):
Honestly I don't understand the whole logic of naming metrics, so I just left some minor remarks.

include/seastar/core/execution_stage.hh
```diff
@@ -337,8 +359,14 @@ private:
         auto wrapped_function = [&_function = _function] (Args... args) {
             return _function(std::forward<Args>(args)...);
         };
-        auto name = fmt::format("{}.{}", _name, sg.name());
-        return per_group_stage_type(name, sg, wrapped_function);
+
+        auto base_name = _name;
+        auto current_name = fmt::format("{}.{}", base_name, sg.name());
+        auto execution_stage = per_group_stage_type(current_name, sg, wrapped_function);
+        execution_stage.set_rename_function([base_name](const class execution_stage& stage) {
```
nuivall: I am not sure if we really need this function injection, it complicates already complicated relationship across those classes.

src/core/execution_stage.cc
andrzej-jackowski-scylladb: There will be more than one - e.g. in ScyllaDB `transport`, and `cql3_select` are two separate execution stages that can have the same scheduling group.

andrzej-jackowski-scylladb:
v3:
 - Addressed the reviewers comments

src/core/execution_stage.cc
andrzej-jackowski-scylladb: Ok, changed to name = _name

src/core/execution_stage.cc
andrzej-jackowski-scylladb: Ok, better make it consistent

src/core/execution_stage.cc
andrzej-jackowski-scylladb: Yes, I changed the loop to be the same as in flush and pool

tests/unit/metrics_test.cc
andrzej-jackowski-scylladb: Actually not, removed it from the test

tests/unit/scheduling_group_test.cc
andrzej-jackowski-scylladb: added

tests/unit/scheduling_group_test.cc
andrzej-jackowski-scylladb: Added BOOST_CHECK_NO_THROW

include/seastar/core/execution_stage.hh
piodul: I agree that this looks a bit awkward. I think that putting the rename callback inside `execution_stage` is the wrong direction. The `execution_stage`/`concrete_execution_stage` IMO should be dumb and simple, it should not have the knowledge about how to rename itself.

A different proposal:

- Add a `rename(sstring)` function to the `execution_stage` which takes care of renaming it to the given name and also re-registering it in the `execution_stage_manager`, updating metrics, etc. This would make it possible to rename it explicitly from the outside.
- Introduce a per-shard registry of inheriting execution stages. When a scheduling group gets renamed, we go over all inheriting execution stages and tell them to rename their execution stages. I thought that maybe we could use a `scheduling_group_key` to manage this (it has a callback for a rename operation) as it would reduce coupling, but creating such a key requires an asynchronous operation, so the seastar framework would have to be responsible for creating it anyways, so maybe it's not worh it...

It's not super pretty as we have to introduce yet another manager class, but it makes the interface of `execution_stage` a bit simpler. One could argue that the execution stages created by the inheriting execution stage objects are a bit magical in the sense that they change their name on scheduling group rename, but it is perfectly possible to implement something like that by the user via the aforementioned `scheduling_group_key` mechanism, so I think that it's not too magical in the end.

@xemul WDYT?

src/core/execution_stage.cc
```diff
@@ -67,6 +67,20 @@ execution_stage* execution_stage_manager::get_stage(const sstring& name) {
     return _stages_by_name[name];
 }
 
+void execution_stage_manager::update_scheduling_group_name(const sstring& new_sg_name) {
+    for (auto&& stage : _execution_stages) {
+        if (stage->get_sg().name() == new_sg_name) {
```
xemul: So you relay on the fact that stage->get_sg().name() would return the _new_ name of a scheduling group that had been renamed by `engine()._task_queues[sg._id]->rename(new_name, new_shortname);` performed by the caller. Is my understanding correct?

If so, why can't it be explicit, namely `if (stage->name() == old_name)`? Or at least `if (stage->get_sg() == current_sg)`?

src/core/execution_stage.cc
```diff
@@ -67,6 +67,20 @@ execution_stage* execution_stage_manager::get_stage(const sstring& name) {
...
+            auto old_name = stage->name();
+            stage->rename();
+            _stages_by_name.erase(old_name);
+            auto ret = _stages_by_name.emplace(stage->name(), stage);
+            if (!ret.second) {
+                throw std::invalid_argument(format("Execution stage {} already exists when updating.", stage->name()));
```
xemul: That's horrible. Once thrown, the scheduling group happens to be half-renamed, i.e. -- tasks queues are renamed, but io-queues and sched group specific data remain with old names. This method should be noexcept

include/seastar/core/execution_stage.hh
```diff
@@ -170,6 +173,22 @@ public:
     bool poll() const noexcept {
         return !_empty;
     }
+
+    /// Returns const reference to scheduling_group
+    const scheduling_group& get_sg() const noexcept { return _sg; }
```
xemul: nit: scheduling_group is lightweight handler, it can be copied around by value

include/seastar/core/execution_stage.hh
xemul: Maybe we can
1. make `virtual execution_stage::rename()` that re-registers itself in the manager (provided the new name differs from the old name)
2. `inheriting_concrete_execution_stage::rename() override` walks the _stage_for_group and renames them as it needs?

tests/unit/scheduling_group_test.cc
```diff
@@ -372,6 +372,21 @@ SEASTAR_THREAD_TEST_CASE(sg_create_and_key_create_in_parallel) {
...
+    auto sg1 = create_scheduling_group("foo", 100).get();
+    with_scheduling_group(sg1, [&] { return stage(); }).get();
+
+    rename_scheduling_group(sg1, "bar").get();
+
+    BOOST_CHECK_NO_THROW({
+        auto sg2 = create_scheduling_group("foo", 100).get();
+        with_scheduling_group(sg2, [&] { return stage(); }).get();
+    });
+}
```
xemul, line 393: There should also be a test that destroys scheduling group and creates another one under the different name (it should check that sg index is re-used with internal::scheduling_group_index())

include/seastar/core/execution_stage.hh
xemul: Ouch, `inheriting_concrete_execution_stage` is not `execution_stage` :man_facepalming:

xemul: It looks like `inheriting_concrete_execution_stage::per_group_stage_type` should be not `concrete_execution_stage`, but rather an explicit sub-class that inherits from `concrete_execution_stage`, wraps the `auto name = fmt::format("{}.{}", _name, sg.name());` logic and overrides its virtual rename method to do proper update of that name formatting.

The `concrete_execution_stage` is final, but I don't see why it can't stop being such. Probably I'm missing something, so maybe the change should be more hairy, like change

```
class concrete_execution_stage<func> final
     : public execution_stage { implementation }
```
into
```
class concrete_execution_stage_base<func>
     : public execution_stage { implementation }
class concrete_execution_stage<func> final
     : public concrete_execution_stage_base<func> { no new code here, just inherit }
```
and then add
```
class inheriting_concrete_execution_stage::per_group_stage_type 
    : public concrete_execution_stage_base<func> { override rename() method }
```

src/core/execution_stage.cc
andrzej-jackowski-scylladb: I understand that throwing here is horrible, however, realistically we will throw a couple of lines above, in `stage->rename()`. That's because `update_metric_group()` will throw `double_registration` if the metrics already exists.

I see the following possible improvements:
 1. I can make `update_scheduling_group_name` noexcept, by removing the `invalid_argument` and handling metrics' exceptions. However, the system state will be in kind-of corrupted state anyway, as I think that's impossible to fix the metrics at this point.
 2. Keep `update_scheduling_group_name` throwing and implement kind of cleanup in `rename_scheduling_group` - revert `task_queues `renaming to roll-back to initial scheduling group state before throwing from `rename_scheduling_group`.
 3. Change the throwing for something stronger, e.g. SEASTAR_ASSERT.

@xemul please let me know which of the options seems the best or let me know if you have a better idea.

src/core/execution_stage.cc
```diff
@@ -111,33 +111,38 @@ execution_stage::execution_stage(const sstring& name, scheduling_group sg)
...
+
+void execution_stage::update_metric_group() {
+    _metric_group.clear();
     _metric_group = metrics::metric_group("execution_stages", {
              metrics::make_counter("tasks_scheduled",
                                   metrics::description("Counts tasks scheduled by execution stages"),
-                                  { metrics::label_instance("execution_stage", name), },
-                                  [name, &esm = internal::execution_stage_manager::get()] {
+                                  { metrics::label_instance("execution_stage", _name), },
+                                  [name=_name, &esm = internal::execution_stage_manager::get()] {
```
avikivity: Please give the equals sign some space.

include/seastar/core/execution_stage.hh
avikivity: Moreover, acronyms are frowned upon in public names.

avikivity:
Perhaps we can take the scheduling group name out of the equation. We already have an automatic shard label attached to metrics. Perhaps we can designate metrics as having automatic scheduling group labels.

Notice the class `label_instance`. It's now a pair sstring, sstring. What if it was a pair sstring, variant<sstring, scheduling_group>, with the rule that if the value is a scheduling group, its name at the point of sampling is returned? Think of it as a replacable "%sg%" template. In fact we could implement it like that if we want to.

This isn't to be taken lightly, it puts scheduling in a layer underneath metrics, but it's already quite fundamental so we can consider it. @xemul @amnonh what do you think?

xemul:
>  Perhaps we can designate metrics as having automatic scheduling group labels.

Not all metrics are labeled by scheduling group.

> This isn't to be taken lightly, it puts scheduling in a layer underneath metrics ...

There can be three layers -- scheduling groups, CPU scheduling and IO scheduling. Nowadays the first layer is pretty thin and relies heavily on the second, but it's better be reversed. And $specific scheduling will stay above metrics while groups management can move to below it.

src/core/execution_stage.cc
xemul: In a "fixed" version (after this patch) name conflict shouldn't exist, so no exceptions of that kind.

amnonh:
> Perhaps we can take the scheduling group name out of the equation. We already have an automatic shard label attached to metrics. Perhaps we can designate metrics as having automatic scheduling group labels.
> 
> Notice the class `label_instance`. It's now a pair sstring, sstring. What if it was a pair sstring, variant<sstring, scheduling_group>, with the rule that if the value is a scheduling group, its name at the point of sampling is returned? Think of it as a replacable "%sg%" template. In fact we could implement it like that if we want to.
> 
> This isn't to be taken lightly, it puts scheduling in a layer underneath metrics, but it's already quite fundamental so we can consider it. @xemul @amnonh what do you think?

Multiply the total number of metrics by the number of scheduling groups by default? I would prefer not to.

About adding a mechanism to make it simpler to report a metric by scheduling group, in the per-shard case, we get it for free by simply storing data on a shard level. We can create such a mechanism for SG, The reactor knows which SG is running at each time; you would need to shard a memory by SG and make it updatable using an api that calls the right metric.

It doesn't seem to be too complicated for the basic stuff, like an object wrapping counter and gauge.
It's less trivial for things like a function call (think about reporting the number of sstables or cache size).
As long as it's for a specific good reason, I'm in favour.
See also https://github.com/scylladb/scylladb/issues/25274

Also, as part of my current log-vendetta, it would be good if we expand our log mechanism, on top of other things, adding the SG automatically would be nice.

avikivity:
> > Perhaps we can designate metrics as having automatic scheduling group labels.
> 
> Not all metrics are labeled by scheduling group.

I meant, one would opt-in to having the label, but the value would be automatically provided.

> 
> > This isn't to be taken lightly, it puts scheduling in a layer underneath metrics ...
> 
> There can be three layers -- scheduling groups, CPU scheduling and IO scheduling. Nowadays the first layer is pretty thin and relies heavily on the second, but it's better be reversed. And $specific scheduling will stay above metrics while groups management can move to below it.

It's still not clear whether scheduling should be above or below the metrics API, even the veneer.

tests/unit/metrics_test.cc
andrzej-jackowski-scylladb: fixed

include/seastar/core/execution_stage.hh
andrzej-jackowski-scylladb: Updated the `concrete_execution_stage` class hierarchy

src/core/execution_stage.cc
andrzej-jackowski-scylladb: Ok, changed the method to `noexcept`

src/core/execution_stage.cc
andrzej-jackowski-scylladb: fixed

tests/unit/scheduling_group_test.cc
andrzej-jackowski-scylladb: Reported https://github.com/scylladb/seastar/issues/3010 for the issue with destroy + create and included draft of the tests in the issue description

include/seastar/core/execution_stage.hh
andrzej-jackowski-scylladb: fixed both

src/core/execution_stage.cc
andrzej-jackowski-scylladb: Changed the function argument to `sg` and changed the if to `if (stage->get_scheduling_group() == sg)`

andrzej-jackowski-scylladb:
v3:
- Replaced the `_rename_func` callback with a new subclass in `concrete_execution_stage` type hierarchy
- Added `noexcept` to `update_scheduling_group_name`
- Addressed other reviewers' remarks

andrzej-jackowski-scylladb:
> Perhaps we can take the scheduling group name out of the equation. We already have an automatic shard label attached to metrics. Perhaps we can designate metrics as having automatic scheduling group labels.

Having the invalid metrics values is one problem but the bottom line is that (without this fix) the names of execution_stages are incorrect. Unless we want to entirely remove the scheduling group from `inheriting_concrete_execution_stage`, I think it should be fixed (e.g. the invalid name can be misleading during core-dump analysis).

PR #2998, opened 2025-09-27
file: coroutinize some functions

avikivity:
Improve readability (and sometimes performance) by coroutinizing
functions that are sure to block.


avikivity:
v2: fix include order for modules build

xemul:
It looks like `internal::file_read_state` becomes unneeded. It was there to carry the state across then-chain, but now with coroutine it can all be open-coded directly into do_dma_read_bulk

avikivity:
> It looks like `internal::file_read_state` becomes unneeded. It was there to carry the state across then-chain, but now with coroutine it can all be open-coded directly into do_dma_read_bulk

Yes, I'll follow-up with its removal.

avikivity:
It's not trivial, I'll see if it makes sense.

xemul:
Thanks

PR #2999, opened 2025-09-30
Add perf test hooks

travisdowns:
This is mostly about pre-run hooks for perf-tests. On Redpanda side we use this to reset global RNG objects to make test/perf runs more deterministic. Other miscellaneous improvements (maybe?) to perf tests and documentation are included.

I put the new functions in the `internal` namespace so as not to officially expose them in the public API, hopefully making for a less fraught discussion about whether this is the exact API we'd want, etc.

AI summary follows.

This pull request introduces support for pre-run hooks in the Seastar performance testing framework, allowing users to register functions that are executed before each test run. It also improves documentation and adds new tests to verify the hook functionality and measure timer overhead. The most important changes are grouped below.

### Performance Test Framework Enhancements

* Added a `pre_hook` type and the `register_pre_run_hook()` function to allow registration of functions that are called before each test run; multiple hooks can be registered and are executed in order. Hooks cannot be deregistered. (`include/seastar/testing/perf_tests.hh` [include/seastar/testing/perf_tests.hhR297-R307](diffhunk://#diff-dabffe0cb7dcf52e9324ce05054be3244a5844578c8723cbb2838a0d0da05ffeR297-R307))
* Implemented storage and invocation of pre-run hooks: hooks are stored in a vector and executed via the new `run_hooks()` method, which is called at the start of each test run. (`tests/perf/perf_tests.cc` [[1]](diffhunk://#diff-d476a7e19fc219a468a4cf5b27a5025797cc207ca5f73b5b922c8210121ef635R396-R407) `include/seastar/testing/perf_tests.hh` [[2]](diffhunk://#diff-dabffe0cb7dcf52e9324ce05054be3244a5844578c8723cbb2838a0d0da05ffeR146-R148) [[3]](diffhunk://#diff-dabffe0cb7dcf52e9324ce05054be3244a5844578c8723cbb2838a0d0da05ffeR253)

### Documentation Improvements

* Clarified fixture lifecycle and usage in performance tests, explaining that a unique fixture is created for each test and how to share setup between test cases. (`tests/perf/perf-tests.md` [tests/perf/perf-tests.mdL56-R56](diffhunk://#diff-25d610d230519b68b4212fd49f537ef76365e099056d2abdb8da455ae0de48fbL56-R56))
* Documented the substantial overhead of manual timer measurement and recommended best practices for minimizing error. (`tests/perf/perf-tests.md` [tests/perf/perf-tests.mdR107-R112](diffhunk://#diff-25d610d230519b68b4212fd49f537ef76365e099056d2abdb8da455ae0de48fbR107-R112))
* Fixed minor spelling error in the documentation ("initialisation" → "initialization"). (`tests/perf/perf-tests.md` [tests/perf/perf-tests.mdL87-R87](diffhunk://#diff-25d610d230519b68b4212fd49f537ef76365e099056d2abdb8da455ae0de48fbL87-R87))

### New and Improved Tests

* Added new performance tests to verify timer overhead and the correct execution of pre-run hooks, including tests that assert whether hooks have run or not. (`tests/perf/perf_tests_perf.cc` [tests/perf/perf_tests_perf.ccR78-R109](diffhunk://#diff-14bcf85b3d588684bcdd2e10d2943b8b227e41615e6c63f48885ae614afa4696R78-R109))
* Added a test with a large inner loop to help observe measurement variability. (`tests/perf/perf_tests_perf.cc` [tests/perf/perf_tests_perf.ccL32-R52](diffhunk://#diff-14bcf85b3d588684bcdd2e10d2943b8b227e41615e6c63f48885ae614afa4696L32-R52))


tests/perf/perf_tests_perf.cc
```diff
@@ -85,3 +87,23 @@ PERF_TEST(perf_tests, test_timer_overhead) {
     return TIMER_LOOPS;
 }
 
+// The following tests check that pre-run hooks are executed properly.
+static bool hook_ran = false;
+
+auto hook_registrar = [] {
+    perf_tests::internal::register_pre_run_hook([](const std::string& g, const std::string& c) {
+        if (g == "perf_tests" && c == "check_that_hook_ran") {
```
xemul: Why such a weird way to register a hook?
Can the test group (perf_tests in the example below doesn't fit here, but e.g. `PERF_TEST_F(fixture, test_fixture_1)` does) do it in its constructor?

travisdowns: @xemul do you mean how I'm actually registering the here hook is weird, or the whole API is weird?

I registered it like this since I plan to use a similar global constructor on the Redpanda side, as our hook is "global", so we want to simply link the hook registrar into all our benchmarks and have it work. So that explains the API anyway.

Let me adjust this "test" to register the hook as you suggest.

If the weirdness is about std::monostate ... I'm not really aware of a cleaner way to run a function at global construction time :).

travisdowns: Fixed to use a fixture constructor as you suggest in https://github.com/scylladb/seastar/compare/c6cbcd803d2a41c011adf14262f4544410dd993f..079f9729e8c572621f59df5b63d12d05d6c87cf8

xemul: >  do you mean how I'm actually registering the here hook is weird, or the whole API is weird?

The former, but once you moved hooks registration to group constructor, the `const std::string& g, const std::string& c` arguments to the hook itself are no longer required, aren't they? So the API itself is also affected

travisdowns: > The former, but once you moved hooks registration to group constructor, the const std::string& g, const std::string& c arguments to the hook itself are no longer required, aren't they? So the API itself is also affected

To be clear, only this test uses the group constructor and it happens to install a hook which doesn't care about the test that is running - but an arbitrary hook would care, e.g., to log the name of the test or other logic. Similar hooks in other perf or testing frameworks all pass the name and other identifying information.

If you want I can update the test to ensure the name is correct (i.e., in test "foo" the hook has seen the name "foo").

xemul: I see. I was confused by the required scope of the hook, now it's clear that hooks are global, like tests are.

In that case, please, make "all tests" and "all hooks" symmetrical in a way hooks are registered -- basically, let's get back with the global `auto hook_registrar = [] { ... }()` thing from the previous iteration, but wrap it with PERF_HOOK or PERF_TEST_HOOK. It's not nice the generic test .cc file directly calls methods from internal namespace (I mean the `perf_tests::internal::register_pre_run_hook()`)

travisdowns: OK, I'm on it.

travisdowns: Done.

PR #3000, opened 2025-09-30
.clangd: tell the LSP about seastar's header style

travisdowns:
.clangd: tell the LSP about seastar's header style

Seastar includes all public using angle brackets.

Tell clangd about it so that it gets it right (available since clangd 20).

Also removes .clangd from .gitignore. This was added in:

https://github.com/scylladb/seastar/pull/1640

with the assertion that clangd stores its index there (and .cache), but
that is not correct: it uses only .cache and possibly some other
directories outside the project, but .clangd itself has always been
a configuration file (that should be checked in).

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,4 @@
+Style:
+  AngledHeaders:
+    # public headers (under include/seastar) use angle brackets
+    - "seastar/.*"
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,4 @@ tags
 .idea/
 .vscode/
 compile_commands.json
-.clangd
 .cache


.clangd
```diff
@@ -0,0 +1,3 @@
+Style:
+  AngledHeaders:
+    - ".*" # Headers in third-party directories
```
avikivity: We include private headers with double quotes (e.g. headers in src/core/)

travisdowns: @avikivity thanks for the note - it's rare enough I've forgotten!

I've adjusted the regex to apply only to seastar public headers and tested it out on a few use cases and it looks good.

.clangd
```diff
@@ -0,0 +1,4 @@
+Style:
+  AngledHeaders:
+    # public headers (under include/seastar) use angle brackets
+    - "seastar/.*"
```
avikivity, line 4: How will this handle `#include "zlib.h"` or similar?

travisdowns: Headers not matching this regex will use the default rules which uses `<>` for system headers (headers found through -isystem includes). Based on my quick tests of boost, stdlib headers it DTRT (uses <>).

travisdowns: This rule is effectively just an override for the seastar case where the headers are _not_ included via a system include path, yet use <>.

baiwfg2:
@avikivity  Hi, avi. I've found a little problem about clangd compatibility and this is a suitable place to talk about it.

I notice that `rpc_impl.hh` is included in `rpc.hh` at last line and cmake function `seastar_check_self_contained` has excluded the `_impl` header for detecting self-contained issue. So as long as `rpc.hh` is included by `.cc`  instead of `rpc_impl.hh`, the build will be OK even if some `_impl.hh` are not self-contained.

However it does disturb clangd. In `rpc_impl.hh` there'll be multiple red squiggles and we can't do `go to definition` against some symbols like `optional, signature`. How do you resolve this in your team ?

BTW, I switched into clangd because c/c++ by Microsoft is often stalled in large projects. so  using C/C++ may not a good alternative for me.

avikivity:
I don't have red squiggle, not sure why.


We can just inline rpc_impl.hh into rpc.hh.

PR #3005, opened 2025-10-01
condition_variable: support aborting condition variable waits

rockwotj:
Sometimes you have an abort source, and the trick to signal the
condition variable to wake up your waiter is not ideal as there might be
many waiters and you don't want to wastefully invoke all of their
predicates. Instead push the abort source into the condition variable so
we can remove the waiter directly from the intrusive list.


xemul:
It continues the https://github.com/scylladb/seastar/pull/1889 work, which is good

I think @avikivity 's concerns still stand

rockwotj:
> It continues the https://github.com/scylladb/seastar/pull/1889 work, which is good
> 
> 
> 
> I think @avikivity 's concerns still stand

Ah I should have looked for other PRs. Thanks for the pointer. If we are worried about too many methods, then what if there was a way to specifically remove a waiter from the list? Maybe some kind of subscription like API that can be cancelled

rockwotj:
To be honest, really the abort source is the smallest primitive here. You can't use timeouts to simulate abort sources, but you can use abort sources to simulate timeouts

include/seastar/core/condition-variable.hh
```diff
@@ -256,6 +256,49 @@ public:
...
+    ///
+    /// \param timeout time point at which wait will exit with a timeout
+    /// \param as the \ref abort_source that eventually notifies that the wait
+    ///            should be aborted.
+    ///
+    /// \return a future that becomes ready when \ref signal() is called
+    ///         If the condition variable was \ref broken() will return \ref broken_condition_variable
+    ///         exception. If abort source is aborted will return an exception.
+    template<typename Clock = typename timer<>::clock, typename Duration = typename Clock::duration>
+    future<> wait(std::chrono::time_point<Clock, Duration> timeout, abort_source& as) noexcept {
```
xemul, line 279: Waiting for _both_ is definitely excessive

PR #3008, opened 2025-10-01
Add `touch` command to `rl` file processing

radoslawcybulski:
Files processed by `ragel` and `sed` produce files with time modification set to "now". As a result produced precompiled header is different every time ragel files are rebuild (most often after removing `build` directory), which will cause full rebuild of the whole project, when precompiled header is used (because it's different binary wise). After this patch rebuild precompiled header will have the same binary content, which will allow `ccache` alike tools to detect it and reuse previously built content.
This patch adds a `touch` command, which sets output file time modification to the same as input file time modification (since output is just mechanically processed from input). This will make time modification the same after full rebuild, precompiled header will be binary identical and we will avoid full rebuild in precompiled header enabled build mode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -426,7 +426,8 @@ function (seastar_generate_ragel)
     OUTPUT ${args_OUT_FILE}
     COMMAND ${CMAKE_COMMAND} -E make_directory ${out_dir}
     COMMAND ${ragel_RAGEL_EXECUTABLE} -G2 -o ${args_OUT_FILE} ${args_IN_FILE}
-    COMMAND sed -i -e "'1h;2,$$H;$$!d;g'" -re "'s/static const char _nfa[^;]*;//g'" ${args_OUT_FILE})
+    COMMAND sed -i -e "'1h;2,$$H;$$!d;g'" -re "'s/static const char _nfa[^;]*;//g'" ${args_OUT_FILE}
+    COMMAND touch ${args_OUT_FILE} -r ${args_IN_FILE})
 
   add_custom_target (${args_TARGET}
     DEPENDS ${args_OUT_FILE})


CMakeLists.txt
```diff
@@ -426,7 +426,8 @@ function (seastar_generate_ragel)
     OUTPUT ${args_OUT_FILE}
     COMMAND ${CMAKE_COMMAND} -E make_directory ${out_dir}
     COMMAND ${ragel_RAGEL_EXECUTABLE} -G2 -o ${args_OUT_FILE} ${args_IN_FILE}
-    COMMAND sed -i -e "'1h;2,$$H;$$!d;g'" -re "'s/static const char _nfa[^;]*;//g'" ${args_OUT_FILE})
+    COMMAND sed -i -e "'1h;2,$$H;$$!d;g'" -re "'s/static const char _nfa[^;]*;//g'" ${args_OUT_FILE}
+    COMMAND touch ${args_OUT_FILE} -r ${args_IN_FILE})
```
xemul, line 430: I'm not cmake expert , but shouldn't this problem be solved by making OUT_FILE depend on the IN_FILE, so that if IN_FILE doesn't change, the OUT_FILE is not regenerated either?

radoslawcybulski: The issue is slightly different - if we remove the `OUT` file and tell it to rebuild, it will, but the new file will have identical content (we didn't change original file) and different timestamp. It's fine for others, but it makes precompiled header (my another patch) different binary wise, which causes rebuilds (and other issues...) further down. The `touch` makes the timestamp "stable", so if the timestamp of original file don't change, the timestamp of the output file doesn't change either, even if `build` directory or `OUT` file is removed.

xemul: Yes, that's true, but how does it work for, let's say, reactor.hh? The way I assume it works, please correct me if I'm wrong -- the reactor.hh is pre-compiled and leaves reactor.something in build/ dir. If that .something file removed and re-precompiled, its resulting timestamp is renewed and the rest of the tree is re-compiled either, even though its content didn't change. No?

radoslawcybulski: I explained myself poorly - i skipped important thing, `ccache`.

It's as you said, the `pch` will be identical binary wise, but with different timestamp. This will (probably) cause a full rebuild in non-`ccache` build anyway, but `ccache` will compare input files by it's content (hash), will detect that `pch` hasn't changed and will pull output files from it's cache. Previously, because `pch` has changed (binary wise), `ccache` was pretty much ignored first time after build directory was removed.

radoslawcybulski: Updated commit's message.

xemul: You tell

>  As a result produced precompiled header is different every time ragel files are rebuild (most often after removing build directory), which will cause full rebuild of the whole project, when precompiled header is used (because it's different binary wise)

When sed-ing and/or ragel-ing a file with the same content several times, would the produced .hh files be different? If no, why would precompiled headers be different?

xemul: Let me ask the same question from different angle.

Your change is effectively:

```patch
     cat source.rl | do_something > target.hh
+    touch target.hh -r source.rl
```

would the problem you're trying to solve (and me failing to understand) be solved like this

```patch
-     cat source.rl | do_something > target.hh
+     if !exists(target.hh) or mctime(source.rl) > mctime(target.hh):
+         cat source.rl | do_something > target.hh
```

?

radoslawcybulski: > When sed-ing and/or ragel-ing a file with the same content several times, would the produced .hh files be different? If no, why would precompiled headers be different?

The file is the same, timestamp is different. I'm speculating here as i didn't dive into clang source code - the precompiled header build by clang contains timestamps of header files (recursively) used to build that pch. So if one of headers is produced by a tool, that creates the same content, but different timestamp file every time, we get a binary different pch every time as well.

I assume it's done for safety reasons.

xemul: Ok, now this explains the intent well, please, include this observation into the patch comment.

However it still doesn't answer my initial question: shouldn't we instead instruct cmake not to update target.hh if source.rl timestamp is "old"? Like I suggested [here](https://github.com/scylladb/seastar/pull/3008#discussion_r2397067275)

xemul: Ah, the "remove build dir" case. The target.hh is gone and is re-created anyway, so it should "inherit" the timestamp from source.rl

radoslawcybulski: @xemul updated, please check if it makes sense.

PR #3011, opened 2025-10-01
iotune: Fix warmup calculation bug and botched rebase

StephanDollberg:
Correctly calculate 5% (brainfart pointed out by Pavel).

Also fixes a botched rebase that removed some of my earlier changes.


StephanDollberg:
@xemul

PR #3015, opened 2025-10-02
DPDK looks like it's flaking in CI?

travisdowns:
just checking if CI is broken


travisdowns:
DPDK and Alpine look broken or flakey?

xemul:
The "test with dpdk enabled" definitely is

travisdowns:
Per Pavel seems to have resolved itself.

PR #3018, opened 2025-10-02
Add prometheus bench, coroutinze prom

travisdowns:
Three primary changes related to prometheus endpoint:

 - Introduce bench for prometheus
 - Add test for global label functionality, which was missing
 - Convert prometheus endpoint to use (more) coroutines

These are precursors to a perf improvement I want to make.

Here is the benchmark output before/after the coroutine change:

***before:***

| test                                     |  iterations |      median |         mad |         min |         max |      allocs |       tasks |        inst |      cycles |
| -                                        |          -: |          -: |          -: |          -: |          -: |          -: |          -: |          -: |          -: |
| metrics_perf_fixture.test_few_metrics    |      533000 |     1.847us |     0.000ns |     1.847us |     1.847us |      25.129 |       0.026 |     17925.7 |      6783.4 |
| metrics_perf_fixture.test_large_families |     2000000 |   548.379ns |     0.000ns |   548.379ns |   548.379ns |       4.164 |       0.002 |      8921.8 |      2174.8 |
| metrics_perf_fixture.test_many_families  |     2000000 |   745.484ns |     0.000ns |   745.484ns |   745.484ns |      10.272 |       0.002 |     11362.2 |      2955.7 |
| metrics_perf_fixture.test_middle_ground  |     2000000 |   542.494ns |     0.000ns |   542.494ns |   542.494ns |       4.745 |       0.002 |      8727.8 |      2152.5 |

***after***

| test                                     |  iterations |      median |         mad |         min |         max |      allocs |       tasks |        inst |      cycles |
| -                                        |          -: |          -: |          -: |          -: |          -: |          -: |          -: |          -: |          -: |
| metrics_perf_fixture.test_few_metrics    |      569100 |     1.724us |     0.000ns |     1.724us |     1.724us |      25.128 |       0.021 |     17575.3 |      6297.5 |
| metrics_perf_fixture.test_large_families |     2000000 |   532.980ns |     0.000ns |   532.980ns |   532.980ns |       4.164 |       0.002 |      8922.6 |      2115.2 |
| metrics_perf_fixture.test_many_families  |     2000000 |   722.662ns |     0.000ns |   722.662ns |   722.662ns |      10.272 |       0.002 |     11356.0 |      2867.4 |
| metrics_perf_fixture.test_middle_ground  |     2000000 |   530.925ns |     0.000ns |   530.925ns |   530.925ns |       4.745 |       0.002 |      8728.0 |      2107.9 |

As expected the differences are ~nil, below the noise floor, as the coroutines are only introduced around a relatively long-running function.

### DIFF
--- a/tests/perf/metrics_perf.cc
+++ b/tests/perf/metrics_perf.cc
@@ -133,7 +133,7 @@ struct metrics_perf_fixture {
         perf_tests::start_measuring_time();
         for ([[maybe_unused]] auto _: irange(iterations)) {
             output_stream<char> out{counting_data_sink{}};
-            co_await access{}.write_body(config, "", false, true, false, always_true, std::move(out));
+            co_await access{}.write_body(config, false, "", false, true, false, always_true, std::move(out));
         }
         perf_tests::stop_measuring_time();
 
--- a/include/seastar/core/prometheus.hh
+++ b/include/seastar/core/prometheus.hh
@@ -60,7 +60,7 @@ namespace details {
 using filter_t = std::function<bool(const metrics::impl::labels_type&)>;
 
 class test_access {
-    future<> write_body(config cfg, sstring metric_family_name, bool prefix, bool show_help, bool enable_aggregation, filter_t filter, output_stream<char>&& s);
+    future<> write_body(config cfg, bool use_protobuf_format, sstring metric_family_name, bool prefix, bool show_help, bool enable_aggregation, filter_t filter, output_stream<char>&& s);
 
     friend struct metrics_perf_fixture;
 };
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -28,6 +28,8 @@
 #include <sstream>
 
 #include <seastar/core/metrics_api.hh>
+#include <seastar/core/future.hh>
+#include <seastar/core/iostream.hh>
 #include <seastar/core/scollectd.hh>
 #include <seastar/http/function_handlers.hh>
 #include <boost/algorithm/string/replace.hpp>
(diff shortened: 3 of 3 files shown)


avikivity:
Please fix spelling, especially "prom", I'm allergic.

tests/perf/metrics_perf.cc
```diff
@@ -0,0 +1,163 @@
...
+
+#include <seastar/core/iostream.hh>
+#include <seastar/core/metrics.hh>
+#include <seastar/core/metrics_api.hh>
+#include <seastar/core/metrics_registration.hh>
+#include <seastar/core/prometheus.hh>
+#include <seastar/testing/perf_tests.hh>
+#include <seastar/util/tmp_file.hh>
+
+#include <boost/range/irange.hpp>
```
avikivity: For new code, let's avoid boost::ranges, std::views::subrange is the replacement.

tests/perf/metrics_perf.cc
```diff
@@ -0,0 +1,163 @@
...
+
+struct metrics_perf_fixture {
+
+    const filter_t always_true = [](auto& mi){ return true; };
+
+    seastar::future<size_t> run_metrics_bench(size_t group_count, size_t families_per_group, size_t series_per_family) {
+        using namespace seastar;
+        using namespace seastar::metrics;
+
+        remove_existing_metrics();
```
avikivity, line 96: Not an expert on fixtures, but wouldn't that poison global state? So if a later test dependent on standard metrics, it would fail.

tests/perf/metrics_perf.cc
avikivity: Or maybe iota

avikivity:
Looks good. The results are normalized per-metric?

travisdowns:
> Looks good. The results are normalized per-metric?

Yes, exactly.

tests/perf/metrics_perf.cc
travisdowns: Thanks for the suggestion, fixed to use `iota` plus a helper method.

tests/perf/metrics_perf.cc
travisdowns: Yes it definitely does not "leave no trace".

It only applies to this particular process which has only these all tests that all want to start from a clean slate. So it causes a problem for some hypothetical future benchmark added to this same binary which wants to use the standard metrics.

I went this way since the metrics infra is itself sort of a global singleton. If it's a bit problem I can look at further refactoring that to make it independently test/benchable.

travisdowns:
@avikivity wrote:

> Please fix spelling, especially "prom", I'm allergic.

I couldn't find any use of `prom` in the commits or commit messages, only in the PR cover page & title. I have fixed those.

travisdowns:
@avikivity anything else you're looking for here? I have some income changes stacked on top of this one but won't add them here to keep the reviews some.

PR #3019, opened 2025-10-03
futurize::invoke: use std::invoke and forwarding

travisdowns:
- Use std::invoke to call the passed function, rather than a direct call. This allows member function pointers and other callable types to be used. This is already the case in other calls like future_invoke.
 - Forward the function itself, not just the arguments. This allows an rvalue-qualified operator() to be called if present.

This is good for a 0.06% reduction in the .text section of the release redpanda binary, so I guess we do have some && call operators?

The real reason for this though is the std::invoke part. Later I build on this.

### DIFF
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -1994,14 +1994,14 @@ template<typename T>
 template<typename Func, typename... FuncArgs>
 typename futurize<T>::type futurize<T>::invoke(Func&& func, FuncArgs&&... args) noexcept {
     try {
-        using ret_t = decltype(std::forward<Func>(func)(std::forward<FuncArgs>(args)...));
+        using ret_t = std::invoke_result_t<Func, FuncArgs&&...>;
         if constexpr (std::is_void_v<ret_t>) {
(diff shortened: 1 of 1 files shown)


include/seastar/core/future.hh
```diff
@@ -1952,14 +1952,14 @@ template<typename T>
 template<typename Func, typename... FuncArgs>
 typename futurize<T>::type futurize<T>::invoke(Func&& func, FuncArgs&&... args) noexcept {
```
avikivity, line 1995: >  - Forward the function itself, not just the arguments. This allows
   an rvalue-qualified operator() to be called if present.

950b54ef3d3fd8205d5424dc4bdae94f8573bb25 beat you to it, so now you get to rebase and I get to re-run CI

travisdowns: Yeah haha, weird how it was like that for 8 years or whatever then suddenly two very similar changes collide.

travisdowns: Rebased and commit message updated.

PR #3026, opened 2025-10-03
io_tester: add aio fallbacks count

travisdowns:
Add aio fallbacks metrics (the io_threaded_fallback metrics with "reason" aio_fallback), as this is an important aspect of IO performance which is getting more investigation at the moment.

Note that this is the count of fallback "batches" submitted to the syscall thread, not the number of individual IO ops that fell back (as one batch may have more than one IO in it).

### DIFF
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -38,6 +38,7 @@
 #include <chrono>
 #include <optional>
 #include <ranges>
+#include <stdexcept>
 #include <utility>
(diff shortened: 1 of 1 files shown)


xemul:
@travisdowns , #2975 should work too, but suggests simpler and a bit more precise approach

travisdowns:
Ah I hadn't seen that, in fact I was about to do something similar to get the exact IOCB count. Closing this in favor of that one.

PR #3027, opened 2025-10-04
operator co_await: use rvalue reference

travisdowns:
Use an rvalue reference for the future argument to operator co_await, to avoid an unnecessary move when co_awaiting an lvalue future.

This saves one move in these cases, which is 10 instructions for a trivially ready future<void> like this:

    future<> f = make_ready_future<>();
    co_await f;

or 13 instructions if the `make_ready_future<>` function is not inlined.

The savings goes up if the future has a non-trivial type, or carries an exception.

In Redpanda, this saves slightly more than 0.5% of .text and .debug_info, and I expect a similar savings in CPU.

I think this is a source compatible change, at least all of Redpanda and seastar compile without complaint.

### DIFF
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -185,7 +185,7 @@ public:
 SEASTAR_MODULE_EXPORT_BEGIN
 
 template<typename T>
-auto operator co_await(future<T> f) noexcept {
+auto operator co_await(future<T>&& f) noexcept {
     return internal::awaiter<true, T>(std::move(f));
 }
 


include/seastar/core/coroutine.hh
```diff
@@ -185,7 +185,7 @@ public:
 SEASTAR_MODULE_EXPORT_BEGIN
 
 template<typename T>
-auto operator co_await(future<T> f) noexcept {
+auto operator co_await(future<T>&& f) noexcept {
```
dotnwat, line 188: Oh, how does the lvalue bind to this? Do you think it's binding to `auto operator co_await(coroutine::without_preemption_check<T> f)` with one conversion?

travisdowns: It doesn't of course, sorry for the confusion.  A std::move is needed which was missing in the example in the cover letter (updated).

The key observation is that this change only seems to help for this moved lcalue case, not for the case like `co_await make_ready_future()` where in either case there is 1 future objected created (not sure if this requires RVO of not for the pre-change code).

dotnwat: thanks!

travisdowns: To add, yes for these to be equivalent for the most common rvalue case it requires the obligatory RVO introduced in C++17.

PR #3031, opened 2025-10-06
Internalize reactor::posix_... API methods

xemul:
There are two of those -- posix_connect() and posix_listen(). Both have nothing to do with reactor and are only called by posix_stack.cc. Despite both just deserve being in posix_stack.cc, they are public methods, and need to be deprecated first, for safety.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -136,6 +136,7 @@ void at_exit(noncopyable_function<future<> ()> func);
 void set_current_task(task* t);
 
 pollable_fd posix_listen(socket_address sa, listen_options opts = {});
+future<> posix_connect(pollable_fd pfd, socket_address sa, socket_address local);
 
 }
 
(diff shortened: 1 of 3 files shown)


include/seastar/core/reactor.hh
```diff
@@ -506,7 +508,10 @@ public:
     future<connected_socket> connect(socket_address sa);
     future<connected_socket> connect(socket_address, socket_address, transport proto = transport::TCP);
 
-    pollable_fd posix_listen(socket_address sa, listen_options opts = {});
+    [[deprecated("Internal reactor function, consider using seastar::listen()")]]
+    pollable_fd posix_listen(socket_address sa, listen_options opts = {}) {
+        return internal::posix_listen(sa, opts);
+    }
```
avikivity, line 515: I think there isn't much risk that someone is using it.  But I don't mind being careful.

PR #3032, opened 2025-10-07
Add lib info to one line backtrace

radoslawcybulski:
Previously oneline backtrace would consist only of raw offsets, interpreted as offsets into scylla binary. This prevents backtrace decoding on cases, where portion of it ends in external libraries.

This patch adds library path to the backtrace output, in format `<libpath>+0xAABB....`, where `libpath` is absolute path to the library. This format is handled by both `stall-analyser.py` and `addr2line`, although the latter one requires a minor fix for invalid addresses case.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -793,8 +793,13 @@ public:
 
     void append_backtrace_oneline() noexcept {
         backtrace([this] (frame f) noexcept {
-            reserve(3 + sizeof(f.addr) * 2);
-            append(" 0x");
+            reserve(3 + sizeof(f.addr) * 2 + (f.so->name.empty() ? 0 : f.so->name.size() + 1));
+            append(" ");
+            if (!f.so->name.empty()) {
+                append(f.so->name.c_str(), f.so->name.size());
+                append("+");
+            }
+            append("0x");
             append_hex(f.addr);
         }, _immediate);
     }
--- a/scripts/stall-analyser.py
+++ b/scripts/stall-analyser.py
@@ -62,6 +62,13 @@ def get_command_line_parser():
 class Node:
     def __init__(self, addr: str):
         self.addr = addr
+        x = addr.find('+')
+        if x >= 0:
+            self.addr_only = addr[x + 1:]
+            self.module = addr[:x]
+        else:
+            self.addr_only = addr
+            self.module = None
         self.callers = {}
         self.callees = {}
         self.printed = False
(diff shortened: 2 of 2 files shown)


radoslawcybulski:
@xemul `test_unit_scheduling_group_nesting_run` failed, looks like some statistical check.

scripts/addr2line.py
```diff
@@ -118,6 +118,9 @@ def _output(self):
     def _read_resolved_address(self):
         first = self._output.readline()
         self._parent.debug('Addr2Line read output (first): ', first)
+        # check for invalid address at first line, which might ends in `:0` not in `: 0`
+        if Addr2Line.dummy_pattern.fullmatch(first):
+            return "?? ??:0"
```
michoecho: I have no idea what this section is about.
Could you add some example output in the comment, to illustrate what case this is guarding against?

michoecho: Also, typo: `s/might ends/might end/`

src/core/reactor.cc
```diff
@@ -793,8 +793,13 @@ class backtrace_buffer {
 
     void append_backtrace_oneline() noexcept {
         backtrace([this] (frame f) noexcept {
-            reserve(3 + sizeof(f.addr) * 2);
-            append(" 0x");
+            reserve(3 + sizeof(f.addr) * 2 + (f.so->name.empty() ? 0 : f.so->name.size() + 1));
```
xemul, line 796: Is `f.so` guaranteed to be non-null?

michoecho: It is. Even in the "worst-case", it points to the fake `uknown_shared_object` (sic), not to null.

That said, it might be wise to check here for null anyway.

scripts/addr2line.py
radoslawcybulski: Updated comment.

src/core/reactor.cc
radoslawcybulski: The code was borrowed from `append_backtrace`. No code checks for `.so` being null, i can add it, but it would make sense to add it everywhere, which i'd rather add in different patch, if you guys want.

scripts/addr2line.py
radoslawcybulski: Removed - my original idea was wrong (i had to fix stall-analyser.py instead). Thanks @michoecho !

radoslawcybulski: The reason my idea was wrong is this:
- when we put an invalid address, but valid hex number to underying `llvm-addr2line`, it will reply with `<hex number>: ?? at ??:0`
- if we put not a valid hex number - it will just emit `?? at ??:0`
- address `/lib64/libc.so.6+0xAABBCCDD` is considered not valid hex (`llvm-addr2line` doesn't do module resolving), hence wrong output and an exception
- `?? at ??:0` is used as "end of transmition" marker, because after we send an address to `llvm-addr2line`, we also send a `,`, the first will produce our output, the second will produce single `?? at ??:0` line, which we consider as eof.
- 
Correct solution is to teach `stall-analyser.py` to do module resolving itself and pass `module` as additional parameter to `resolve_address` call.
It might be also useful to actually drop `?? at ??:0` shenanigans and use `llvm-addr2line --output-style=JSON` and parse output as json.

radoslawcybulski: Issue created https://github.com/scylladb/seastar/issues/3035

PR #3034, opened 2025-10-08
prometheus: remove hostname and metric_help config

travisdowns:
Neither are used (they are never read). metric_help is set in one place and hostname is never set. They are vestigial.

Officially deprecated them and remove them at API_LEVEL 9 and above.

Use some macros to suppress warnings about their use in implicitly generated constructors.

### DIFF
--- a/include/seastar/core/internal/api-level.hh
+++ b/include/seastar/core/internal/api-level.hh
@@ -39,3 +39,12 @@ namespace seastar {
         }
     }
 }
+
+// Helpers for ignoring deprecation warnings in code that has to deal with
+// deprecated APIs, e.g., constructors/etc for structs with deprecated fields.
+#define SEASTAR_INTERNAL_BEGIN_IGNORE_DEPRECATIONS \
+    _Pragma("GCC diagnostic push") \
+    _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
+
+#define SEASTAR_INTERNAL_END_IGNORE_DEPRECATIONS \
+    _Pragma("GCC diagnostic pop")
--- a/apps/httpd/main.cc
+++ b/apps/httpd/main.cc
@@ -100,7 +100,6 @@ int main(int ac, char** av) {
                 prometheus::config pctx;
                 net::inet_address prom_addr(config["prometheus_address"].as<sstring>());
 
-                pctx.metric_help = "seastar::httpd server statistics";
                 pctx.prefix = config["prometheus_prefix"].as<sstring>();
 
                 std::cout << "starting prometheus API server" << std::endl;
--- a/include/seastar/core/prometheus.hh
+++ b/include/seastar/core/prometheus.hh
@@ -40,8 +40,19 @@ SEASTAR_MODULE_EXPORT_BEGIN
  * Holds prometheus related configuration
  */
 struct config {
-    sstring metric_help; //!< Default help message for the returned metrics
-    sstring hostname; //!< hostname is deprecated, use label instead
+
+    [[deprecated("metric_help is deprecated and no longer used, to be removed in 2027")]]
+    sstring metric_help;
+
+    [[deprecated("hostname is deprecated and unused, use label instead, to be removed in 2027")]]
+    sstring hostname;
+
+    SEASTAR_INTERNAL_BEGIN_IGNORE_DEPRECATIONS // prevent warnings about deprecated fields in implicitly-defined special member functions
+    config() = default;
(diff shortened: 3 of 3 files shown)


avikivity:
/cc @amnonh 

To create an API_LEVEL, you need to update configure.py, compatibility.md, CMakeStupidName.txt

travisdowns:
> To create an API_LEVEL, you need to update configure.py, compatibility.md, CMakeStupidName.txt

My goal wasn't to create an API level, but mark this as for deletion in the next API level. After re-reading compat.md I'm to sure if that's the right approach: perhaps deprecation (which is addressed there separately from "breaking changes")  does not need to be tied to an API level, just a long-enough notice period.

If it does need an API level, I imagine we don't want to create a new API level changes for each minor change like this, rather we would schedule/batch several small changes into the one API level bump: that was my attempt here: schedule removal for API_LEVEL=9, and when some bigger changes comes along that pushes us to actually introduce that level all the associated changes are noted.

avikivity:
Yes, if you want to [[deprecate]] something, just do it. It can be garbage collected later.

API_LEVELs are reserved for truly breaking changes.

travisdowns:
Thanks @avikivity, I've removed the API_LEVEL stuff and just deprecated these two fields with a 2027 suggested removal date.

include/seastar/core/internal/api-level.hh
```diff
@@ -39,3 +39,12 @@ namespace seastar {
         }
     }
 }
+
+// Helpers for ignoring deprecation warnings in code that has to deal with
+// deprecated APIs, e.g., constructors/etc for structs with deprecated fields.
+#define BEGIN_IGNORE_DEPRECATIONS \
```
avikivity: Please qualify with SEASTAR_INTERNAL so we don't intrude on the global macro namespace in a visible header.

travisdowns: Done.

avikivity:
Still need review from @amnonh

travisdowns:
> where collectd was supported and not all metrics had help lines.

Is collectd no longer supported? I would love to that backend as would @StephanDollberg in https://github.com/scylladb/seastar/pull/2586.

For now can I just unconditionally delete these?

travisdowns:
@avikivity these changes have been approved.

xemul:
It now spoils logs with 

```
[203/383] Building CXX object apps/httpd/CMakeFiles/app_httpd.dir/main.cc.o
In file included from /home/xemul/src/seastar/apps/httpd/main.cc:33:
/home/xemul/src/seastar/include/seastar/core/prometheus.hh: In destructor ‘seastar::prometheus::config::~config()’:
/home/xemul/src/seastar/include/seastar/core/prometheus.hh:42:8: warning: ‘seastar::prometheus::config::metric_help’ is deprecated: metric_help is deprecated and no longer used, to be removed in 2027 [-Wdeprecated-declarations]
   42 | struct config {
      |        ^~~~~~
/home/xemul/src/seastar/include/seastar/core/prometheus.hh:45:13: note: declared here
   45 |     sstring metric_help;
      |             ^~~~~~~~~~~
/home/xemul/src/seastar/include/seastar/core/prometheus.hh:42:8: warning: ‘seastar::prometheus::config::metric_help’ is deprecated: metric_help is deprecated and no longer used, to be removed in 2027 [-Wdeprecated-declarations]
   42 | struct config {
      |        ^~~~~~
/home/xemul/src/seastar/include/seastar/core/prometheus.hh:45:13: note: declared here
   45 |     sstring metric_help;
      |             ^~~~~~~~~~~
/home/xemul/src/seastar/include/seastar/core/prometheus.hh:42:8: warning: ‘seastar::prometheus::config::hostname’ is deprecated: hostname is deprecated and unused, use label instead, to be removed in 2027 [-Wdeprecated-declarations]
   42 | struct config {
      |        ^~~~~~
/home/xemul/src/seastar/include/seastar/core/prometheus.hh:48:13: note: declared here
   48 |     sstring hostname;
      |             ^~~~~~~~
/home/xemul/src/seastar/apps/httpd/main.cc: In lambda function:
/home/xemul/src/seastar/apps/httpd/main.cc:100:36: note: synthesized method ‘seastar::prometheus::config::~config()’ first required here
  100 |                 prometheus::config pctx;
      |                                    ^~~~
```

```
$ clang --version
clang version 20.1.8 (Fedora 20.1.8-4.fc42)
```

avikivity:
https://github.com/llvm/llvm-project/issues/165230

PR #3040, opened 2025-10-10
io_queue: Destroy priority class data with scheduling group

xemul:
When a scheduling group is destroyed, the IO queue's priority_class_data that corresponds to it stays intact. It's not actually a leak, as when the reactor is destroyed all class data objects are collected.

However, if another scheduling group under the same ID is created, the IO queue will re-use the dangling priority class for the newly created group, but its name and shares won't match the new group until rename or shares update. And it's not nice that IO class remains alive anyway.

Since scheduling group destruction must happen after all tasks from it are completed and picked up, the class destruction code may also assume that no queued or in-flight requests exist.

Test included.


xemul:
This is re-submitted #2993 which was unmerged by mistake

PR #3046, opened 2025-10-11
file: Query physical block size and minimum I/O size

### LINKED ISSUE #3006: iotune uses minimal_io_size for IOPS measurements
This value reflects the size that application should use to make request work at all. However, there's physical_block_size value that shows something different -- the IO size that the drive uses internal.

One consequence of not using the physical IO size is that random writes test may stop on RMW behavior of the disk thus measuring lower IOPS

tchaikov:
Enhance block device initialization to query additional device characteristics beyond logical block size:

- use physical_block_size for write alignment
- use logical_block_size for read alignment

Fixes #3006

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -497,6 +497,11 @@ public:
         }
     }
 
+    io_queue* try_get_io_queue(dev_t devid) noexcept {
+        auto queue = _io_queues.find(devid);
+        return queue != _io_queues.end() ? queue->second.get() : nullptr;
+    }
+
 private:
     future<> update_bandwidth_for_queues(internal::priority_class pc, uint64_t bandwidth);
     void rename_queues(internal::priority_class pc, sstring new_name);
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -32,6 +32,14 @@
 
 #define __user /* empty */  // for xfs includes, below
 
+// Define STATX_DIOALIGN and related fields if not available from system headers (Linux < 6.1)
+#ifndef STATX_DIOALIGN
+#define STATX_DIOALIGN 0x00002000U
+// Check if struct statx has the stx_dio_mem_align field (added in Linux 6.1)
+// We need to extend the structure for older kernels/headers
+#define SEASTAR_STATX_NEEDS_DIO_FIELDS 1
+#endif
+
 #include <sys/syscall.h>
 #include <dirent.h>
 #include <linux/types.h> // for xfs, below
@@ -1059,32 +1067,73 @@ query_statx_mem_align(int fd) {
             return stx.stx_dio_mem_align;
         }
     }
+#else
+    // Headers don't have the stx_dio_mem_align field, so we can't use it
+    // even if the kernel supports it. We'll rely on other fallbacks.
 #endif
     return std::nullopt;
 }
 
-static internal::alignments xfs_alignments(int fd, const dioattr& da, unsigned block_size) {
-    static bool xfs_with_relaxed_overwrite_alignment = internal::kernel_uname().whitelisted({"5.12"});
+// Query device-level alignment properties (common to all filesystems)
+struct device_alignment_info {
+    std::optional<unsigned> memory_alignment;      // from statx
+    std::optional<unsigned> physical_block_size;   // from io_queue override
+};
 
-    unsigned memory = da.d_mem;
+static device_alignment_info query_device_alignment_info(int fd, dev_t device_id) {
+    device_alignment_info info;
 
-    // Try to get better memory alignment from statx first
-    if (auto mem_align = query_statx_mem_align(fd)) {
-        memory = std::max(*mem_align, static_cast<size_t>(internal::min_memory_alignment));
-    } else {
-        memory = std::max(da.d_mem, internal::min_memory_alignment);
+    // Query statx for DIO memory alignment (kernel 6.1+)
+    info.memory_alignment = query_statx_mem_align(fd);
+
+    // Check for physical_block_size override from io_properties.yaml
+    // Note: I/O queues may not be registered for all devices (e.g., tmpfs, test environments)
+    if (auto* io_queue = engine().try_get_io_queue(device_id)) {
+        info.physical_block_size = io_queue->physical_block_size();
     }
 
-    // xfs wants at least the block size for writes
-    // FIXME: really read the block size
-    auto disk_write = std::max<unsigned>(da.d_miniosz, block_size);
+    return info;
+}
 
-    return {
-        .memory = memory,
-        .disk_read = da.d_miniosz,
-        .disk_write = disk_write,
-        .disk_overwrite = xfs_with_relaxed_overwrite_alignment ? da.d_miniosz : disk_write,
+// Unified function for all filesystem alignment calculations
+static internal::alignments filesystem_alignments(
+    int fd,
+    dev_t device_id,
+    unsigned block_size,
+    unsigned long fs_type  // from statfs.f_type
+) {
+    auto device_info = query_device_alignment_info(fd, device_id);
+
+    // Initialize with generic defaults for all filesystems
+    internal::alignments align = {
+        .memory = std::max(device_info.memory_alignment.value_or(4096), internal::min_memory_alignment),
+        .disk_read = 512,  // Linux O_DIRECT minimum
+        .disk_write = block_size,
+        .disk_overwrite = block_size,
     };
+
+    // Override with filesystem-specific alignments if available
+    if (fs_type == internal::fs_magic::xfs) {
+        dioattr da;
+        if (::ioctl(fd, XFS_IOC_DIOINFO, &da) == 0) {
+            static bool xfs_with_relaxed_overwrite_alignment = internal::kernel_uname().whitelisted({"5.12"});
+
+            // Override with XFS-specific values
+            align.memory = std::max(device_info.memory_alignment.value_or(da.d_mem), internal::min_memory_alignment);
+            align.disk_read = da.d_miniosz;
+            align.disk_write = std::max<unsigned>(da.d_miniosz, block_size);
+            align.disk_overwrite = xfs_with_relaxed_overwrite_alignment ? da.d_miniosz : align.disk_write;
+        }
+    }
+
+    // Common: apply physical_block_size override for all filesystems
+    // This ensures we avoid hardware read-modify-write regardless of filesystem
+    if (device_info.physical_block_size) {
+        align.disk_write = std::max<unsigned>(align.disk_write, *device_info.physical_block_size);
+        align.disk_overwrite = std::max<unsigned>(align.disk_overwrite, *device_info.physical_block_size);
+    }
+
+    return align;
 }
 
 // Query block device alignment properties using ioctl and statx
@@ -1165,11 +1214,6 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
             fsi.block_size = sfs.f_bsize;
             switch (sfs.f_type) {
             case internal::fs_magic::xfs:
-                dioattr da;
-                if (::ioctl(fd, XFS_IOC_DIOINFO, &da) == 0) {
-                    fsi.align = xfs_alignments(fd, da, fsi.block_size);
-                }
-
                 fsi.append_challenged = true;
                 static auto xc = xfs_concurrency_from_kernel_version();
                 fsi.append_concurrency = xc;
@@ -1208,6 +1252,7 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
                 fsi.nowait_works = false;
             }
             fsi.nowait_works &= engine()._cfg.aio_nowait_works;
+            fsi.align = filesystem_alignments(fd, st.st_dev, fsi.block_size, sfs.f_type);
             s_fstype.insert(std::make_pair(st.st_dev, std::move(fsi)));
             return make_file_impl(fd, std::move(options), flags, std::move(st));
         });


tchaikov:
This table shows the parameters resulting from various device types being registered with the Linux kernel.


  Table 9: Common device types and their resulting parameters

  | Device                          | logical | physical | min_io | opt_io | align_off |
  |---------------------------------|---------|----------|--------|--------|-----------|
  | Disk 512/512                    | 512     | 512      | 512    | 0      | 0         |
  | Disk 512/4KiB                   | 512     | 4096     | 4096   | 0      | 0         |
  | Disk 512/4KiB, 1-aligned        | 512     | 4096     | 4096   | 0      | 3585      |
  | Disk 4 KiB/4 KiB                | 4096    | 4096     | 4096   | 0      | 0         |
  | RAID0, 64 KiB × 4 drives        | 512     | 512      | 65536  | 262144 | 0         |
  | RAID1, 16 KiB                   | 512     | 512      | 16384  | 0      | 0         |
  | RAID5, 8 KiB × 3 drives, 1-aln. | 512     | 512      | 8192   | 16384  | 3584      |

quoted from https://people.redhat.com/msnitzer/docs/linux-advanced-storage-6.1.pdf, section 1.5

src/core/file.cc
```diff
@@ -1033,12 +1033,26 @@ xfs_concurrency_from_kernel_version() {
...
+        size_t minimum_io_size;
+        if (auto ret = ::ioctl(fd, BLKIOMIN, &minimum_io_size); ret == -1) {
+            return make_exception_future<shared_ptr<file_impl>>(
+                    std::system_error(errno, std::system_category(), "ioctl(BLKIOMIN) failed"));
+        }
+        // use the maximum size of:
+        // - minimum_io_size: preferred minimum I/O size the device can perform without performing read-modify-write
+        // - physical block size: smallest unit a physical storage device can write atomically
+        // - logical block size: smallest unit the the storage device can address (typically 512 bytes)
+        size_t block_size = std::ranges::max({logical_block_size, physical_block_size, minimum_io_size});
```
avikivity: physical_block_size should be the write alignment, but not the read alignment. There's no downside to reading a 512 byte logical sector from a 4096 byte physical sector disk.

Wrt write alignment, even there it's iffy. Writing 4096 avoids RMW but can generate space amplification. With the current exposed parameters, physical_block_size for writes is the best match. We may want to expose another write alignment (choosing a name will be hard) to indicate a non-optimal block write block size that is smaller than the other one.

avikivity:
> This table shows the parameters resulting from various device types being registered with the Linux kernel.
> 
> Table 9: Common device types and their resulting parameters
> 
> Device	logical	physical	min_io	opt_io	align_off
> Disk 512/512	512	512	512	0	0
> Disk 512/4KiB	512	4096	4096	0	0
> Disk 512/4KiB, 1-aligned	512	4096	4096	0	3585
> Disk 4 KiB/4 KiB	4096	4096	4096	0	0
> RAID0, 64 KiB × 4 drives	512	512	65536	262144	0

I don't understand min_io and opt_io for RAID0. The disks could just as easily read and write 512 byte blocks here.


> RAID1, 16 KiB	512	512	16384	0	0

Nor this. Why is 16k more optimal than anything else?

> RAID5, 8 KiB × 3 drives, 1-aln.	512	512	8192	16384	3584

16k makes sense here because it avoids a RMW. But I don't understand 8k.

> quoted from https://people.redhat.com/msnitzer/docs/linux-advanced-storage-6.1.pdf, section 1.5

tchaikov:
> > This table shows the parameters resulting from various device types being registered with the Linux kernel.
> > Table 9: Common device types and their resulting parameters
> > Device	logical	physical	min_io	opt_io	align_off
> > Disk 512/512	512	512	512	0	0
> > Disk 512/4KiB	512	4096	4096	0	0
> > Disk 512/4KiB, 1-aligned	512	4096	4096	0	3585
> > Disk 4 KiB/4 KiB	4096	4096	4096	0	0
> > RAID0, 64 KiB × 4 drives	512	512	65536	262144	0
> 
> I don't understand min_io and opt_io for RAID0. The disks could just as easily read and write 512 byte blocks here.
> 
> > RAID1, 16 KiB	512	512	16384	0	0
> 
> Nor this. Why is 16k more optimal than anything else?
> 
> > RAID5, 8 KiB × 3 drives, 1-aln.	512	512	8192	16384	3584
> 
> 16k makes sense here because it avoids a RMW. But I don't understand 8k.
> 
> > quoted from https://people.redhat.com/msnitzer/docs/linux-advanced-storage-6.1.pdf, section 1.5

I looked into the Linux kernel source code to understand how these values are set. Here's what I found:

TL;DR: For RAID devices, `min_io` is set to the chunk/stripe unit size (the amount written to each individual disk), not the minimum physically possible I/O size.

In the Linux kernel, all RAID types set io_min to the chunk size:

 - RAID0 (https://github.com/torvalds/linux/blob/master/drivers/md/raid0.c#L386):
    lim.io_min = mddev->chunk_sectors << 9;
    lim.io_opt = lim.io_min * mddev->raid_disks;

- RAID5 (https://github.com/torvalds/linux/blob/master/drivers/md/raid5.c#L7748):
    lim.io_min = mddev->chunk_sectors << 9;
    lim.io_opt = lim.io_min * (conf->raid_disks - conf->max_degraded);

- RAID10 (https://github.com/torvalds/linux/blob/master/drivers/md/raid10.c#L4003):
    lim.io_min = mddev->chunk_sectors << 9;
    lim.io_opt = lim.io_min * raid10_nr_stripes(conf);

Answering Your Questions

> RAID0 (64 KiB × 4 drives): "The disks could just as easily read and write 512 byte blocks here."

  Yes, physically they can, but:
  - min_io = 65536 (64 KiB) is the chunk size - aligning I/O to chunk boundaries avoids fragmenting operations across multiple drives unnecessarily
  - opt_io = 262144 (256 KiB) is the full stripe width (64 KiB × 4) - this maximizes parallelism by utilizing all drives

  You're right that smaller I/Os work, but they defeat the purpose of RAID0's striping.

> RAID1 (16 KiB): "Why is 16k more optimal than anything else?"

  I couldn't find this in the kernel code. RAID1 doesn't set io_min to a chunk size (it has no striping concept). This 16k value is likely either:
  - Inherited from the underlying disk's characteristics
  - Controller/configuration specific
  - Or possibly an artifact of how that specific example was set up

  RAID1 should typically just mirror the underlying device topology.

> RAID5 (8 KiB × 3 drives): "I don't understand 8k"

  From the kernel code:
  - min_io = 8192 (8 KiB) is the chunk size - smallest alignment unit
  - opt_io = 16384 (16 KiB) is the data stripe width (8 KiB × 2 data drives, excluding parity)

  The 8k ensures I/O aligns with chunk boundaries. The 16k (which you understood) is optimal because it writes a full data stripe.

src/core/file.cc
tchaikov: Thanks for pointing out the read/write differentiation! I've updated the implementation:

- Read alignment: Now uses logical_block_size only (as you suggested - no downside to reading 512-byte sectors)
- Write alignment: Now uses physical_block_size (not max(logical, physical, min_io))

You're right about the space amplification issue. I verified this in the kernel source - the Linux kernel only enforces `logical_block_size` alignment for O_DIRECT (see `block/fops.c`:`blkdev_dio_invalid()`):
```c
  return (iocb->ki_pos | iov_iter_count(iter)) &
          (bdev_logical_block_size(bdev) - 1);
```

This confirms that `physical_block_size` and `min_io` are optimization hints, not requirements. Using `physical_block_size` provides the best balance:
  - Avoids hardware-level RMW (4K physical sectors)
  - Prevents space amplification from RAID stripe alignment (`min_io` can be 64 KiB+)

For context, I found that RAID devices set min_io to the chunk/stripe size (see `drivers/md/raid0.c`:386, `raid5.c`:7748, `raid10.c`:4003), which would cause massive space waste if used for write alignment.

Regarding exposing another write alignment: This is an interesting idea. We could expose something like:

  - disk_write_dma_alignment = physical_block_size (current, safe default)
  - disk_write_dma_alignment_optimal = max(physical_block_size, min_io) (for apps willing to trade space for throughput)

However, I'm inclined to defer this until we have a concrete use case. Most Seastar applications probably want "do the right thing" rather than having to choose between alignment strategies.

tchaikov:
v2:

- use `physical_block_size` for write alignment
- use `logical_block_size` for read alignment

mykaul:
> quoted from https://people.redhat.com/msnitzer/docs/linux-advanced-storage-6.1.pdf, section 1.5

This is a great document, but it's also ~15 years old. It's less relevant for SSD/NVMe, for example.

tchaikov:
> This is a great document, but it's also ~15 years old. It's less relevant for SSD/NVMe, for example.

Good point. While the document is dated, the kernel topology API it describes is still current and works for modern devices.

For modern SSDs/NVMe (verified from Linux kernel source drivers/nvme/host/core.c):

```c
/* NOWS = Namespace Optimal Write Size */
if (id->nows)
    io_opt = bs * (1 + le16_to_cpu(id->nows));
```

- logical_block_size and physical_block_size: Usually 512 or 4096 bytes
- minimum_io_size: Set to physical_block_size (typically 512 or 4096)
- optimal_io_size: Derived from NOWS (Namespace Optimal Write Size) field in NVMe spec
  - If NOWS = 0 (not specified): optimal_io_size = 0
  - If NOWS is set (e.g., 7): optimal_io_size = block_size * (1 + NOWS) = 512 * 8 = 4096
  - Many consumer NVMe drives don't set NOWS, so optimal_io_size = 0
  - Enterprise/datacenter NVMe drives often set NOWS to indicate write granularity

The key difference from the RAID examples is that RAID arrays have complex stripe/chunk geometry (64 KiB chunks × multiple drives) leading to much larger min_io/opt_io values. NVMe drives, when they set NOWS, typically indicate page-size granularity (4-16 KiB).

Important: Our implementation in this PR deliberately uses only physical_block_size for write alignment, not optimal_io_size or minimum_io_size. This avoids write amplification that would occur with large optimal_io_size values (e.g., forcing all writes to 256 KiB alignment for RAID0 would be wasteful). We only need to avoid hardware-level RMW, which physical_block_size handles.

So while the document's RAID examples show legacy use cases with their stripe-based optimal sizes, we use the simpler physical_block_size approach that works well for both modern SSDs and RAID without write amplification issues.

avikivity:
> v2:
> 
> * use `physical_block_size` for write alignment
> * use `logical_block_size` for read alignment

We recently discovered that some disks lie about physical_block_size.

@robertbindar (or @tchaikov if you want), suggest adjusting iotune to detect the physical block size and write it in io_properties.yaml. Then the reactor can pick it up and use it to override what it detects from the disk.

src/core/file.cc
```diff
@@ -619,13 +619,15 @@ static bool blockdev_nowait_works(dev_t device_id) {
...
         : posix_file_impl(fd, f, options, device_id, blockdev_nowait_works(device_id)) {
-    // Configure DMA alignment requirements based on block device block size
-    _memory_dma_alignment = block_size;
-    _disk_read_dma_alignment = block_size;
-    _disk_write_dma_alignment = block_size;
-    _disk_overwrite_dma_alignment = block_size;
+    // Configure DMA alignment requirements based on block device characteristics
+    // - Read alignment: logical_block_size (no performance penalty for reading 512-byte sectors)
+    // - Write alignment: physical_block_size (avoids hardware-level RMW)
+    _memory_dma_alignment = write_block_size;
```
avikivity: memory_dma_alignment is fixed 512 IIRC regardless of logical/physical sector sizes.

avikivity: I see it can be obtained via statx stx_dio_mem_align.

avikivity: Interestingly it returns 4 for my nvme (and 512 for xfs files)

tchaikov: > Interestingly it returns 4 for my nvme (and 512 for xfs files)

`stx_dio_mem_align` is always 4 for NVMe devices. 

1. NVMe driver rets `dma_alignment`

  file: `/home/kefu/dev/linux/drivers/nvme/host/core.c`
  ```c
  static void nvme_set_ctrl_limits(struct nvme_ctrl *ctrl,
                struct queue_limits *lim)
  {
        lim->max_hw_sectors = ctrl->max_hw_sectors;
        lim->max_segments = min_t(u32, USHRT_MAX,
                min_not_zero(nvme_max_drv_segments(ctrl), ctrl->max_segments));
        lim->max_integrity_segments = ctrl->max_integrity_segments;
        lim->virt_boundary_mask = NVME_CTRL_PAGE_SIZE - 1;
        lim->max_segment_size = UINT_MAX;
        lim->dma_alignment = 3;  // <-- LINE 2080
  }
  ```
  2. block layer reads it and sets `stx_dio_mem_align`

  File: `/home/kefu/dev/linux/block/bdev.c`
  ```c
  void bdev_statx(const struct path *path, struct kstat *stat, u32 request_mask)
  {
        ...
        if (request_mask & STATX_DIOALIGN) {
                stat->dio_mem_align = bdev_dma_alignment(bdev) + 1;  // 3 + 1 = 4
                stat->dio_offset_align = bdev_logical_block_size(bdev);
                stat->result_mask |= STATX_DIOALIGN;
        }
        ...
  }
  ```

so, it's indeed 4 bytes. will prepare a new revision of this pull request to use `statx()` to query memory buffer alignment for block devices, and fall back to physical_block_size if statx unavailable or unsupported

tchaikov:
> > v2:
> > 
> > * use `physical_block_size` for write alignment
> > * use `logical_block_size` for read alignment
> 
> We recently discovered that some disks lie about physical_block_size.
> 
> @robertbindar (or @tchaikov if you want), suggest adjusting iotune to detect the physical block size and write it in io_properties.yaml. Then the reactor can pick it up and use it to override what it detects from the disk.

will see what i can do. the io performance is critical to crimson project. so this is important to us.

tchaikov:
@scylla/seastar-maint ping?

@xemul hi Pavel, could you help review this change as well?

src/core/file.cc
```diff
@@ -1051,16 +1059,50 @@ static internal::alignments xfs_alignments(const dioattr& da, unsigned block_siz
...
+        // The Linux kernel only enforces logical_block_size alignment for O_DIRECT (see block/fops.c:blkdev_dio_invalid).
+        // Using physical_block_size avoids RMW at the hardware level.
+        size_t write_block_size = physical_block_size;
+
         internal::fs_info fsi;
-        fsi.block_size = block_size;
+        fsi.block_size = write_block_size;
         fsi.nowait_works = blockdev_nowait_works(st.st_dev);
-        fsi.align = internal::alignments { fsi.block_size, fsi.block_size, fsi.block_size, fsi.block_size };
+        fsi.align = internal::alignments { memory_dma_alignment, read_block_size, write_block_size, write_block_size };
```
xemul: Please, wrap the block sizes detection logic into `blkdev_alignments()` method (similarly to how it's done with `xfs_alignments()` helper)

src/core/file.cc
```diff
@@ -1051,16 +1059,50 @@ static internal::alignments xfs_alignments(const dioattr& da, unsigned block_siz
...
+        if (syscall(__NR_statx, fd, "", AT_EMPTY_PATH, STATX_DIOALIGN, &stx) == 0) {
+            if (stx.stx_mask & STATX_DIOALIGN) {
+#ifndef SEASTAR_STATX_NEEDS_DIO_FIELDS
+                // Use kernel-reported memory alignment (stx_dio_mem_align)
+                // Field is available in Linux 6.1+ headers
+                memory_dma_alignment = stx.stx_dio_mem_align;
+#else
+                // On older headers, the field doesn't exist in the struct
+                // The kernel may support it, but we can't access it safely
+                // Fall back to physical_block_size
```
xemul: If SEASTAR_STATX_NEEDS_DIO_FIELDS is not defined, calling statx makes no sense at all

include/seastar/core/reactor.hh
```diff
@@ -236,6 +236,8 @@ private:
     // ... when dispatched all requests get into this single sink
     internal::io_sink _io_sink;
     unsigned _num_io_groups = 0;
+    // Override physical_block_size for devices that lie about it (from io_properties.yaml)
+    std::unordered_map<dev_t, uint32_t> _physical_block_size_overrides;
```
xemul: Please, put physical_block_size value on io_queue and export it from there

src/core/disk_params.cc
```diff
@@ -65,6 +65,9 @@ struct convert<seastar::internal::disk_params> {
         if (node["write_saturation_length"]) {
             mp.write_saturation_length = parse_memory_size(node["write_saturation_length"].as<std::string>());
         }
+        if (node["physical_block_size"]) {
```
xemul, line 68: doc/io-properties-file.md needs to be updated

src/core/file.cc
```diff
@@ -1109,7 +1109,12 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
         internal::fs_info fsi;
         fsi.block_size = write_block_size;
         fsi.nowait_works = blockdev_nowait_works(st.st_dev);
-        fsi.align = internal::alignments { memory_dma_alignment, read_block_size, write_block_size, write_block_size };
+        fsi.align = internal::alignments {
+            static_cast<unsigned>(memory_dma_alignment),
+            static_cast<unsigned>(read_block_size),
+            static_cast<unsigned>(write_block_size),
+            static_cast<unsigned>(write_block_size)
+        };
```
xemul: Does this change belong to the first patch in this PR?

apps/iotune/iotune.cc
```diff
@@ -971,6 +1074,14 @@ int main(int ac, char** av) {
 
                 desc.read_iops = best_read_iops.iops;
                 desc.write_iops = best_write_iops.iops;
+
+                // Detect physical block size
```
xemul: Please, add a cli option (off by default) not to run this test all the time

apps/iotune/iotune.cc
```diff
@@ -140,6 +142,15 @@ struct evaluation_directory {
 
                 _min_data_transfer_size = std::max(_min_data_transfer_size, disk_min_io_size);
                 _max_iodepth += read_first_line_as<uint64_t>(queue_dir / "nr_requests");
+
+                // Read physical_block_size from sysfs for comparison with detected value
+                if (!_reported_physical_block_size) {
+                    auto pbs_file = queue_dir / "physical_block_size";
```
xemul, line 148: We know (scylladb/scylladb#25315) that sometimes this value is mis-reported, so iotune shouldn't configure it all the time by default

src/core/file.cc
```diff
@@ -1070,6 +1070,13 @@ make_file_impl(int fd, file_open_options options, int flags, struct stat st) noe
                     std::system_error(errno, std::system_category(), "ioctl(BLKPBSZGET) failed"));
         }
 
+        // Check for physical_block_size override from io_properties.yaml
```
xemul: Shouldn't filesystem files (posix_file_readl_impl and append_challenged_posix_file_impl) also take this value into consideration? Currently they just grab whatever xfs filesystem tells them to or just stick to defaults (4k)

include/seastar/core/disk_params.hh
```diff
@@ -92,6 +93,19 @@ public:
     const std::vector<dev_t>& queue_devices(unsigned q) const {
         return _disks.at(q).devices;
     }
+
+    // Look up physical_block_size override for a device
+    // Returns std::nullopt if no override is configured
+    std::optional<uint32_t> get_physical_block_size(dev_t device) const {
```
xemul: Presumably this helper is not needed any longer

apps/iotune/iotune.cc
xemul: Another concern. Previous patch from this PR does this (simplified):

```
    auto size = ioctl(BLKPBSZGET)
    if (io_queue.physical_block_size)
        size = *io_queue.physical_block_size
```

in case io_queue gets its physical_block_size value from this code, does it make any sense? AFAICS from Linux kernel sources

```c
static int blkdev_common_ioctl(struct block_device *bdev,...
{
    ...
    case BLKPBSZGET: /* get block device physical block size */
        return put_uint(argp, bdev_physical_block_size(bdev));
}

QUEUE_LIM_RO_ENTRY(queue_physical_block_size, "physical_block_size");

static inline unsigned int queue_physical_block_size(const struct request_queue *q)
{
    return q->limits.physical_block_size;
}

static inline unsigned int bdev_physical_block_size(struct block_device *bdev)
{
    return queue_physical_block_size(bdev_get_queue(bdev));
}
```

these are indeed the same values obtained via different APIs

src/core/file.cc
```diff
@@ -1048,20 +1048,74 @@ static internal::alignments xfs_alignments(const dioattr& da, unsigned block_siz
     return ret;
 }
 
+// Query block device alignment properties using ioctl and statx
+static
+future<internal::alignments>
```
xemul: This function is not really a "coroutine", it only returns ready futures all the time

tchaikov: ahh, yes. changed to a regular function.

include/seastar/core/disk_params.hh
tchaikov: yup, thanks for pointing this out. dropped.

apps/iotune/iotune.cc
tchaikov: added an cli option.

src/core/file.cc
tchaikov: yeah, added a dedicated commit to apply physical_block_size override to filesystem files

src/core/file.cc
tchaikov: This has been addressed in the latest revision. The block device alignment detection logic has been extracted into a `blkdev_alignments()` helper function following the same pattern as `xfs_alignments()`.

The refactored code:
- Introduces `blkdev_alignments(int fd, dev_t device_id)` at src/core/file.cc:1078
- Calls it from `make_file_impl()` at line 1130: `auto align = blkdev_alignments(fd, st.st_rdev);`
- Follows the same structure as `xfs_alignments()` with proper error handling

The helper encapsulates all ioctl/statx queries and physical_block_size override logic in one place.

tchaikov:
Addressed the concern about circular override logic in iotune.cc (commit 2df5cbd0).

**The Issue:**
As you correctly pointed out, sysfs `/sys/block/*/queue/physical_block_size` and `ioctl(BLKPBSZGET)` both read from the same kernel value (`q->limits.physical_block_size`). Writing the sysfs value to io_properties.yaml just to override BLKPBSZGET with the same value is pointless.

**The Fix:**
Now iotune only writes `physical_block_size` to io_properties.yaml when empirical detection is performed via `--detect-physical-block-size`. The override mechanism is only useful when the disk lies and we detect the real value through performance testing.

Behavior:
- `iotune` (default): No physical_block_size in io_properties.yaml
- `iotune --detect-physical-block-size`: Writes empirically detected value

This way, the override only contains meaningful data (the real physical block size when it differs from what the kernel reports), not circular references.

apps/iotune/iotune.cc
tchaikov: Addressed the concern about circular override logic in `iotune.cc`

**The Issue:**
As you correctly pointed out, sysfs `/sys/block/*/queue/physical_block_size` and `ioctl(BLKPBSZGET)` both read from the same kernel value (`q->limits.physical_block_size`). Writing the sysfs value to io_properties.yaml just to override BLKPBSZGET with the same value is pointless.

**The Fix:**
Now iotune only writes `physical_block_size` to `io_properties.yaml` when empirical detection is performed via `--detect-physical-block-size`. The override mechanism is only useful when the disk lies and we detect the real value through performance testing.

Behavior:
- `iotune` (default): No physical_block_size in io_properties.yaml
- `iotune --detect-physical-block-size`: Writes empirically detected value

This way, the override only contains meaningful data (the real physical block size when it differs from what the kernel reports), not circular references.

src/core/disk_params.cc
tchaikov: thanks. updated.

include/seastar/core/reactor.hh
tchaikov: sure. done.

src/core/file.cc
tchaikov: yes, should have been fixed.

src/core/file.cc
tchaikov: should have been fixed.

tchaikov:
v3:

- dropped unused helper function of `get_physical_block_size`
- use plain function instead of coroutine when possible
- use statx and fall back to XFS_IOC_DIOINFO
- added a dedicated commit to apply physical_block_size override to filesystem files
- do not configure `physical_block_size` with iotune by default

avikivity:
Note: we found that on some AWS NVMe disks, the physical block size is reported as 512 bytes whereas it really is 4096, so we may want to add a quirk layer to detect those disks and adjust.

src/core/file.cc
```diff
@@ -1048,20 +1082,60 @@ static internal::alignments xfs_alignments(const dioattr& da, unsigned block_siz
     return ret;
 }
 
+// Query block device alignment properties using ioctl and statx
+static
+internal::alignments
+blkdev_alignments(int fd, dev_t device_id) {
+    // Query logical block size (smallest addressable unit)
+    size_t logical_block_size;
+    if (auto ret = ::ioctl(fd, BLKBSZGET, &logical_block_size); ret == -1) {
```
avikivity: Is BLKBSZGET not intended for the filesystem block size?

See comments to this response: https://unix.stackexchange.com/a/52225/343739

avikivity:
> Note: we found that on some AWS NVMe disks, the physical block size is reported as 512 bytes whereas it really is 4096, so we may want to add a quirk layer to detect those disks and adjust.

Ah, I see it's part of the series.

src/core/file.cc
tchaikov: Thanks for pointing this out, Avi. I am quoting the comments here:

> Well, I realized that [block/ioctl.c](http://lxr.free-electrons.com/source/block/ioctl.c) in Linux kernel source contains misleading comments. BLKPBSZGET (notice the P) gets physical sector size, BLKSSZGET gets logical sector size and BLKBSZGET (or BLKBSZGET_32 in block/compat_ioctl.c) gets file system allocation unit (cluster) size. [Logical and physical sector size differ when Advanced Format 512e is used.](http://www.oeey.com/2014/10/invalid-argument-directio-4kn.html) ([512e @ WP](http://en.wikipedia.org/wiki/Advanced_Format#512e)) – 
> [Palec](https://unix.stackexchange.com/users/50602/palec)

so, i am switching to  `BLKSSZGET` for the `logical_block_size` in `blkdev_alignments()`.

tchaikov:
v4:

- use `BLKSSZGET` instead of `BLKBSZGET` for the `logical_block_size` in `blkdev_alignments()`.
- update the commit messages to explain the rationales behind these decisions.

xemul:
@tchaikov , as a followup -- please, make `blkdev_alignments` use `query_device_alignment_info` to get statx memory alignment and overriden physical block size from io-queue.

tchaikov:
> @tchaikov , as a followup -- please, make `blkdev_alignments` use `query_device_alignment_info` to get statx memory alignment and overriden physical block size from io-queue.

created #3186

travisdowns:
```
Filesystem Write Alignment Strategy
====================================

We use conservative alignment that prioritizes performance over space:

- Generic default: block_size (typically 4096 bytes)
- XFS-specific: XFS_IOC_DIOINFO values when available
- Override: physical_block_size from io_properties.yaml takes precedence
```

@tchaikov I am confused, does this mean mean that if I set 512 bytes in physical_block_size in io_properties.yaml, it will override the calculated _filesystem_ alignment value?

I always assumed file alignment needs to be: `max(optimal/allowed block device block size, required filesystem alignment size)`, so overrides for the first term in the max wouldn't eliminate the effect of the second term (if larger).

tchaikov:
@travisdowns hi Travis, your assumption is correct: the final alignment should be max(block device block size, filesystem alignment). The implementation does exactly this:

  1. Filesystem alignment is calculated first:
    - Generic default: block_size (typically 4096)
    - XFS-specific: values from `XFS_IOC_DIOINFO`
  2. Then the override is applied using `std::max()`:
    - If override is 512 and filesystem requires 4096 → uses 4096
    - If override is 8192 and filesystem requires 4096 → uses 8192

The Confusion in the Commit Message

The commit message states:

> Override: physical_block_size from io_properties.yaml takes precedence

I think the wording is misleading. it should say:

> Override: physical_block_size from io_properties.yaml is applied as a minimum requirement via max()

Sorry for the confusion =(

PR #3050, opened 2025-10-12
http: add "Connection: close" header to final server response.

### LINKED ISSUE #3041: HTTP server should send "close" header when closing the connection.
In reference to this issue: https://github.com/scylladb/scylladb/issues/26298

with HTTP1.1 when the server closes connection, the "close" header is not included in the response.
As far as I understand client at HTTP layer may not notice that TCP connection was actually closed, until it tries to reuse it.

https://github.com/scylladb/seastar/blob/85492711a2b01e8f0fbf05baba105bdf018c335c/src/http/httpd.cc#L338-L342
This seems to solve the original issue: 
```cpp
 else if (!keep_alive && req->_version == "1.1") {
        resp->_headers["Connection"] = "close";
    }
```

The question is, if it should be also added in other situations, like:
https://github.com/scylladb/seastar/blob/85492711a2b01e8f0fbf05baba105bdf018c335c/src/http/httpd.cc#L186-L194

m-szymon:
According to HTTP/1.1 standard (RFC9112 chapter 9.6 Tear-down):
> **A sender SHOULD send a Connection header field containing the "close" connection option when it intends to close a connection.**
> [...] in a response, the same field indicates that the server is going to close this connection after the response message is complete. [...]
> A server that receives a "close" connection option MUST initiate closure of the connection (see below) after it sends the final response to the request that contained the "close" connection option. **The server SHOULD send a "close" connection option in its final response on that connection.**

Fixes #3041
Refs scylladb/scylladb#26298

### DIFF
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -183,11 +183,24 @@ set_request_content(std::unique_ptr<http::request> req, input_stream<char>* cont
     }
 }
 
+static void set_header_connection(http::reply& resp, bool keep_alive) {
+    if (keep_alive) {
+        if (resp._version == "1.0") {
+            resp.add_header("Connection", "Keep-Alive");
+        }
+    } else {
+        if (resp._version == "1.1") {
+            resp.add_header("Connection", "close");
+        }
+    }
+}
+
 void connection::generate_error_reply_and_close(std::unique_ptr<http::request> req, http::reply::status_type status, const sstring& msg) {
     auto resp = std::make_unique<http::reply>();
     // TODO: Handle HTTP/2.0 when it releases
     resp->set_version(req->_version);
     resp->set_status(status, msg);
+    set_header_connection(*resp, false);
     resp->done();
     _done = true;
     _replies.push(std::move(resp));
@@ -336,9 +349,7 @@ future<bool> connection::generate_reply(std::unique_ptr<http::request> req) {
     resp->set_version(req->_version);
     set_headers(*resp);
     bool keep_alive = req->should_keep_alive();
-    if (keep_alive && req->_version == "1.0") {
-        resp->_headers["Connection"] = "Keep-Alive";
-    }
+    set_header_connection(*resp, keep_alive);
 
     sstring url = req->parse_query_param();
     sstring version = req->_version;
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -1824,6 +1824,16 @@ SEASTAR_TEST_CASE(test_bad_chunk_length) {
     }, {"400 Bad Request", "Can't parse chunk size and extensions"}, true, new echo_stream_handler());
 }
 
+SEASTAR_TEST_CASE(test_close_response) {
+    return check_http_reply({
+        "GET /test HTTP/1.1\r\nHost: test\r\nConnection: close\r\n\r\n"
+    }, {"200 OK", "Connection: close"}, true, new echo_stream_handler()).then([] {
+        return check_http_reply({
+            "/test\r\n\r\n"
+        }, {"400 Bad Request", "Connection: close", "Can't parse the request"}, false, new echo_string_handler());
+    });
+}
+
 SEASTAR_TEST_CASE(case_insensitive_header) {
     std::unique_ptr<seastar::http::request> req = std::make_unique<seastar::http::request>();
     req->_headers["conTEnt-LengtH"] = "17";


src/http/httpd.cc
```diff
@@ -183,11 +183,24 @@ set_request_content(std::unique_ptr<http::request> req, input_stream<char>* cont
     }
 }
 
+static void set_keep_alive_header(http::reply& resp, bool keep_alive) {
```
ScyllaPiotr: `set_connection_header` would be a better name IMO

src/http/httpd.cc
```diff
@@ -336,9 +349,7 @@ future<bool> connection::generate_reply(std::unique_ptr<http::request> req) {
     resp->set_version(req->_version);
     set_headers(*resp);
     bool keep_alive = req->should_keep_alive();
-    if (keep_alive && req->_version == "1.0") {
```
ScyllaPiotr, line 339: In your patch you test `resp._version` instead, is it guaranteed to work the same way? (Is `resp._version == req->_version` here every time?)

src/http/httpd.cc
m-szymon: I would agree, but these are not the only "connection" options.
And it is another problem - we don't support repeating headers or multiple values listed in the header.
Kind of problem is that user can overwrite these values in "handler". On the other hand it would be nice to let user "close" connection from the handler - he can set the header to be sent, but it will not close the socket immediately on the server side.
But these are bigger changes - maybe they should be tackled with HTTP2.0 which adds it own header features. Here I wanted to solve the simplest issue.

src/http/httpd.cc
m-szymon: It must be used after `resp->set_version(`.

ScyllaPiotr: Agreed, the question is if it is actually used after `resp->set_version()`. I assume it is.

ScyllaPiotr:
> > Also, are you sure your patch actually `fixes` (why `closes`?) the issue? Have you tested it?
> 
> It closes the seastar issue.

What I meant is that you used the word `closes` instead of `fixes`.

> I tested that it fixes the scylladb issue

Could you describe how did you test it? The problem manifests from time to time, how have you verified that the problem is actually fixed and not just randomly not triggered?

> should it be referenced in the commit?

If it (actually) fixes 2 issues, it should have 2 `fixes`.

src/http/httpd.cc
ScyllaPiotr: Agreed.
Let's file separate issues for the observations you've made.

src/http/httpd.cc
m-szymon: yes, it is.

ScyllaPiotr:
> > > > Also, are you sure your patch actually `fixes` (why `closes`?) the issue? Have you tested it?
> > > 
> > > 
> > > It closes the seastar issue.
> > 
> > 
> > What I meant is that you used the word `closes` instead of `fixes`.
> 
> Well, for seastar it is not a bug - we SHOULD send this header, not MUST.

Agreed.

> > > I tested that it fixes the scylladb issue
> > 
> > 
> > Could you describe how did you test it? The problem manifests from time to time, how have you verified that the problem is actually fixed and not just randomly not triggered?
> 
> It reproduces regularly if you repeat the test 3000 times. With the patch I tried 10x that.

Perfect!

> > > should it be referenced in the commit?
> > 
> > 
> > If it (actually) fixes 2 issues, it should have 2 `fixes`.
> 
> It fixes it only when we update scylladb repo. I am not sure what is 'correct' procedure here.

Let the Seastar maintainers decide how to correctly mark this, if at all.

xemul:
> Let the Seastar maintainers decide how to correctly mark this, if at all.

It fixes the issue, then it's Fixes tag

> If it (actually) fixes 2 issues, it should have 2 fixes

If the 2nd issue is in Scylla, then corresponding seastar update SHOULD(*) include the corresponding Fixes tag

(*) but must likely we will forget about it, in this case someone will pick up the issue later and close it by hand

> It reproduces regularly if you repeat the test 3000 times. With the patch I tried 10x that.

Please, include a seastar unit test for the header being in "expected" state

nyh:
Thanks for finding where the RFC dictates this behavior (even if just a SHOULD and not a MUST). For people who are curious exactly which RFC this is (as I was), it is https://www.rfc-editor.org/rfc/rfc9112.html#name-tear-down

Indeed,

> A server that receives a "close" connection option MUST initiate closure of the connection (see below) after it sends the final response to the request that contained the "close" connection option. **The server SHOULD send a "close" connection option in its final response on that connection.** The server MUST NOT process any further requests received on that connection.

src/http/httpd.cc
nyh: I agree with @ScyllaPiotr. In areas related to standards, it's very important to use exactly the same terminology used by the standard and not invent our own terminology.  There is something else called a keep-alive header (https://datatracker.ietf.org/doc/html/rfc2068#section-19.7.1.1) and what you're writing here is the connection header, not a "keep-alive" header.

If you think the name "set_connection_header" is too ambiguous - does it set the header called "connection", or any header on a connection - maybe set_header_connection will be clearer?

src/http/httpd.cc
```diff
@@ -183,11 +183,24 @@ set_request_content(std::unique_ptr<http::request> req, input_stream<char>* cont
     }
 }
 
+static void set_keep_alive_header(http::reply& resp, bool keep_alive) {
+    if (keep_alive) {
+        if (resp._version == "1.0") {
+            resp.add_header("Connection", "Keep-Alive");
```
nyh, line 189: Looks good, but worthy of a comment: This case is NOT standard HTTP 1.0, but rather a non-standard extension over HTTP 1.0 which is documented here https://datatracker.ietf.org/doc/html/rfc2068#section-19.7.1. 

Anyway, since this is correct and also appears to be pre-existing code that you moved, I can't complain.

nyh:
> > Let the Seastar maintainers decide how to correctly mark this, if at all.
> 
> It fixes the issue, then it's Fixes tag

Right. The Seastar issue is about a behavior that the HTTP 1.1 standard says we SHOULD do, but we didn't do, so this patch fixes this bug, and it's a proper "Fixes".

> If the 2nd issue is in Scylla, then corresponding seastar update SHOULD(*) include the corresponding Fixes tag

Exactly. The Seastar patch should "Fixes" the Seastar issue. Later, in the ScyllaDB repository when you update the Seastar submodule  you should "Fixes" the ScyllaDB bug. It would be great if **you** do this when the Seastar patch is merged, but if you don't or forget, as @denesb said someone will notice this later and close the ScyllaDB issue manually saying it was already fixed in Seastar.

avikivity:
> > On the downside, it will add volume to messages for anyone who uses ephemeral connections (although using them is not optimal) and we've seen some users using them. Should we care?
> 
> If that's the problem, then striping "closing" headers is not the solution.
> 
> > Also, are you sure your patch actually `fixes` (why `closes`?) the issue? Have you tested it?
> 
> It closes the seastar issue. I tested that it fixes the scylladb issue - should it be referenced in the commit?

It technically doesn't - a commit in Seastar doesn't affect ScyllaDB, until the submodule is updated. It's the submodule update that will fix the issue in ScyllaDB.

It's okay (even good) to reference it.

m-szymon:
I added unit test and changed the function name as requested.

> > If the 2nd issue is in Scylla, then corresponding seastar update SHOULD(*) include the corresponding Fixes tag
> 
> Exactly. The Seastar patch should "Fixes" the Seastar issue. Later, in the ScyllaDB repository when you update the Seastar submodule you should "Fixes" the ScyllaDB bug.

But that doesn't exclude referring to ScyllaDb issue here, which was the main question. I guess comment in pull request is not too much: https://github.com/scylladb/scylladb/issues/26298

avikivity:
> I added unit test and changed the function name as requested.
> 
> > > If the 2nd issue is in Scylla, then corresponding seastar update SHOULD(*) include the corresponding Fixes tag
> > 
> > 
> > Exactly. The Seastar patch should "Fixes" the Seastar issue. Later, in the ScyllaDB repository when you update the Seastar submodule you should "Fixes" the ScyllaDB bug.
> 
> But that doesn't exclude referring to ScyllaDb issue here, which was the main question. I guess comment in pull request is not too much: [scylladb/scylladb#26298](https://github.com/scylladb/scylladb/issues/26298)

As I said, it's good practice to refer to the ScyllaDB issue, even in the patch changelog. Just don't mark it as "Fixes".

m-szymon:
> As I said, it's good practice to refer to the ScyllaDB issue, even in the patch changelog. Just don't mark it as "Fixes".

Thank you. That's clear now.

nyh:
Thanks, I'll merge this patch, and then submit a PR for ScyllaDB to update Seastar.

PR #3051, opened 2025-10-13
Deprecate in/out streams move-assignment operator

xemul:
The current default implementation of this operator is just wrong. The target of assignment should be "aborted" (or "closed") before being re-written, but it's currently not, so all callers try to be nice and stick to a simplified pattern -- first the stream is default-initialized and not used, later the stream is move-assigned.

However, no callers really need this ability -- in all cases the stream can be constructed properly at that "latter" time. This PR does so.

### DIFF
--- a/include/seastar/core/iostream.hh
+++ b/include/seastar/core/iostream.hh
@@ -358,9 +358,11 @@ public:
     // unconsumed_remainder is mapped for compatibility only; new code should use consumption_result_type
     using unconsumed_remainder = std::optional<tmp_buf>;
     using char_type = CharType;
+    [[deprecated("Uninitialized input_stream is useless")]]
     input_stream() noexcept = default;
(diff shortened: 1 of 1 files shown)


xemul:

tests/perf/http_client_perf.cc
```diff
@@ -52,22 +52,22 @@ using namespace std::chrono_literals;
 class server {
     seastar::server_socket _ss;
     seastar::connected_socket _cs;
-    seastar::input_stream<char> _in;
-    seastar::output_stream<char> _out;
+    std::optional<seastar::input_stream<char>> _in;
+    std::optional<seastar::output_stream<char>> _out;
```
avikivity, line 56: This style is sad, it's better to have initialization structure so those are already ready by the time we initialize.

But I understand it's a lot of effort for little gain.

PR #3052, opened 2025-10-13
Rework output_stream::slow_write()

xemul:
The main change here is unification of some code paths (patch number 1). And few more "while at it" changes.
No functional changes, just improve the readability of that method and facilitate "mixed mode" ongoing effort.

### DIFF
--- a/include/seastar/core/iostream-impl.hh
+++ b/include/seastar/core/iostream-impl.hh
@@ -369,9 +369,9 @@ output_stream<CharType>::write(const char_type* buf, size_t n) noexcept {
 template <typename CharType>
 future<>
 output_stream<CharType>::slow_write(const char_type* buf, size_t n) noexcept {
-  try {
-    SEASTAR_ASSERT(_zc_bufs.empty() && "Mixing buffered writes and zero-copy writes not supported yet");
-    if (!_end && (n >= _size)) {
+    try {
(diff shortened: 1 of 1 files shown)


include/seastar/core/iostream-impl.hh
```diff
@@ -371,32 +371,14 @@ future<>
 output_stream<CharType>::slow_write(const char_type* buf, size_t n) noexcept {
   try {
     SEASTAR_ASSERT(_zc_bufs.empty() && "Mixing buffered writes and zero-copy writes not supported yet");
-    auto bulk_threshold = _end ? (2 * _size - _end) : _size;
```
avikivity, line 374: Does the behavior change? 

bulk_threshold wasn't documented in b8344e31e0eedde64bb8a37a946d3e39ee175600.

xemul: This checks tells "first" from "third" cases I described in the ~~PR~~ patch description:

_end == 0 means buffer is empty, and bulk_threshold of _size means -- new buffer doesn't fit and should be put() on its own

_end != 0 means something is buffered already, and if new buffer exceeds _size - _ens + _size, it means that even if we append the head of new buffer to inner buffer (_size - _end bytes), the remainder will _not_ fit and will have to be put() right after the inner buffer is

So the behavior is not changed :crossed_fingers:

xemul: @avikivity , reminder ping

PR #3053, opened 2025-10-13
memory, smp: support more than 256 shards

avikivity:
The memory manager hard-codes a limit for 256 shards in its memory map. Of the
47 bits available to userspace, 3 are taken by the mmap prefix (usually 6), 8 are taken
by cpu_id (limiting the number of shards to 256), and the remaining 36 are given to
the shard address space (allowing 64 GB/shard). There also a few static arrays sized
for 256 shards.

Here we increase the limit by making the cpu_id bit reservation dynamic, taking more
for cpu_id if needed, at the expense of reducing maximum memory per shard.

Two related changes are made:
 - the --overprovisioned switch now allows --smp n with n greater than the number of vcpus
 - the smp queue allocations are now distributed across shards

These changes allow testing very large configurations on one's workstation.

The memory map for a 2000-shard application now looks like

```
600000000000-600002000000 rw-p 00000000 00:00 0 
600002000000-600002800000 rw-p 00000000 00:00 0 
600002800000-600200000000 ---p 00000000 00:00 0 
600200000000-600202000000 rw-p 00000000 00:00 0 
```
...
```
6f9c00000000-6f9c02000000 rw-p 00000000 00:00 0 
6f9c02000000-6f9c02800000 rw-p 00000000 00:00 0 
6f9c02800000-6f9e00000000 ---p 00000000 00:00 0 
6f9e00000000-6f9e02000000 rw-p 00000000 00:00 0 
6f9e02000000-6f9e02800000 rw-p 00000000 00:00 0 
6f9e02800000-700000000000 ---p 00000000 00:00 0 
```

cpu_id occupies 11 bits (bits 33:43) rather than the usual 8 (bits 36:43). 11 bits is enough to
represent 2000 shards.

Lightly tested with httpd and ScyllaDB.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4526,11 +4526,30 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     auto backend_selector = reactor_opts.reactor_backend.get_selected_candidate();
     seastar_logger.info("Reactor backend: {}", backend_selector);
 
+    _qs_owner = decltype(smp::_qs_owner){new smp_message_queue* [smp::count], qs_deleter{}};
+
+    auto allocate_qs_owner = [this] (unsigned i) {
+        // smp_message_queue has members with hefty alignment requirements.
+        // if we are reactor thread, or not running with dpdk, doing this
+        // new default aligned seemingly works, as does reordering
+        // dlinit dependencies (ugh). But we should enforce calling out to
+        // aligned_alloc, instead of pure malloc, if possible.
+        smp::_qs_owner[i] = reinterpret_cast<smp_message_queue*>(operator new[] (sizeof(smp_message_queue) * smp::count
+            , std::align_val_t(alignof(smp_message_queue))
+        ));
+    };
+
+    auto allocate_smp_queues = [this, &reactors] (unsigned i) {
+        for (unsigned j = 0; j < smp::count; ++j) {
+            new (&smp::_qs_owner[i][j]) smp_message_queue(reactors[j], reactors[i]);
+        }
+    };
+
     unsigned i;
     auto smp_tmain = smp::_tmain;
     for (i = 1; i < smp::count; i++) {
         auto allocation = allocations[i];
-        create_thread([this, smp_tmain, inited, &reactors_registered, &smp_queues_constructed, &smp_opts, &reactor_opts, &reactors, hugepages_path, i, allocation, assign_io_queues, alloc_io_queues, thread_affinity, heapprof_sampling_rate, mbind, backend_selector, reactor_cfg, &mtx, &layout, use_transparent_hugepages] {
+        create_thread([this, smp_tmain, inited, &reactors_registered, &smp_queues_constructed, &smp_opts, &reactor_opts, &reactors, hugepages_path, i, allocation, assign_io_queues, alloc_io_queues, thread_affinity, heapprof_sampling_rate, mbind, backend_selector, reactor_cfg, &mtx, &layout, use_transparent_hugepages, allocate_qs_owner, allocate_smp_queues] {
           try {
             // initialize thread_locals that are equal across all reacto threads of this smp instance
             smp::_tmain = smp_tmain;
@@ -4562,10 +4581,11 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
             allocate_reactor(i, backend_selector, reactor_cfg);
             reactors[i] = &engine();
             alloc_io_queues(i);
-            reactors_registered.arrive_and_wait();
-            smp_queues_constructed.arrive_and_wait();
-            // _qs_owner is only initialized here
+            allocate_qs_owner(i);
             _qs = _qs_owner.get();
+            reactors_registered.arrive_and_wait();
+            allocate_smp_queues(i);
+            smp_queues_constructed.arrive_and_wait();
             start_all_queues();
             assign_io_queues(i);
             inited->arrive_and_wait();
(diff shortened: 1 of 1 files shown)


avikivity:
v2: fix debug build by adding stubs to new internal memory.hh functions

avikivity:
v3: further fix debug mode by pushing the global_setup stub into the internal namespace

avikivity:
v4: fix debug mode even more by returning a valid number of per_shard_memory().

avikivity:
v5: add this must truly be the last fix for debug mode with #include align.hh

Somehow it compiles on my machine.

src/core/memory.cc
```diff
@@ -1832,14 +1838,28 @@ static long mbind(void *addr,
...
+    auto w = std::bit_width(nr_shards - 1);
+    // Preserve at least 8 bits for CPU ID to have a traditional memory map
+    return std::max(w, 8);
+}
+
 size_t
 internal::per_shard_memory(size_t mem, unsigned procs) {
     // limit memory address to fit in 36-bit, see Memory map
-    constexpr size_t max_mem_per_proc = 1UL << 36;
+    size_t max_mem_per_proc = 1UL << (bits_for_cpu_id_and_memory - cpu_id_bits(procs));
```
xemul: Why not `= 1UL << cpu_id_shift` ? 
The `global_setup()` is called before `per_shard_memory()` so cpu_id_shift should be correct

avikivity: Probably a result of evolution, I'll change it and see what breaks.

avikivity:
v6: reuse cpu_id_shift, calculated in global_setup(), in per_shard_memory().

xemul (approved):
LGTM
@avikivity , do you want to do something else else with, or we're good to go?

avikivity:
No further changes planned.

travisdowns:
@avikivity it's worth noting that this adds 5 instructions to the hot path for `free/delete`:

before:
```
test                                                                    iters            runtime     allocs      tasks       inst     cycles
alloc_bench.malloc_only                                             208960000     2.37ns ± 1.57%      1.000      0.000      36.34        7.9
alloc_bench.free_only                                               209315000     2.45ns ± 0.07%      0.000      0.000      30.34        8.6
```

after:
```
test                                                                    iters            runtime     allocs      tasks       inst     cycles
alloc_bench.malloc_only                                             204391000     2.34ns ± 0.26%      1.000      0.000      36.34        7.9
alloc_bench.free_only                                               207044000     2.46ns ± 0.13%      0.000      0.000      35.34        8.6
```

Perhaps this is what you expected, but this probably adds up to a 0.1%-1% regression (in instruction count) for a typical seastar application. The runtime itself in these tight benchmarks ~did not change as I guess these extra loads of constants are off the critical path but the effect in a more realistic setting is likely to be different.

avikivity:
I should have benched and noted this in the cover letter, but I forgot about the existence of alloc_bench.

That said, I don't see what we can do. We can't ignore > 256 vcpu machines.

btw, we could try exposing the small pool allocator where we constexpr-calculate the small pool index for things like continuations and coroutines, so we can skip those for both the allocation and deallocation paths.

travisdowns:
> That said, I don't see what we can do. We can't ignore > 256 vcpu machines.

Yup.

One option is to make some part of the tradeoff a configurable compile-time option. E.g., maybe I want max 16 GB per shard, which would give my access to up to 1024 cores while still having a static bit allocation. Or maybe I want fully dynamic as you've done.

Another option is to do a quick check of "are we in a standard (i.e., 8 bits of CPU) configuration as part of of the fast-path eligibility check we already do, it can probably even be zero cost by rolling it into another variable, and then the fast path can use the hard-coded shifts, and the > 256 vCPU case uses a (still fast) slow path.

> btw, we could try exposing the small pool allocator where we constexpr-calculate the small pool index for things like continuations and coroutines, so we can skip those for both the allocation and deallocation paths.

We are using LTO, and I've had luck with this w/o the need to expose it: and it works for all allocations, not just ones tried to a specific allocator. You can use __builtin_constexpr to only inline in the case the size is known (which it is, most of the time), so all the calculation goes away: there is ~no size overhead in the inlined part: it just calls a different entry point with the calculation already done. That's the "strictly better" option. Arguably inlining a bit more of the fast path is interesting, e.g., the fast path as this is only a handful of  instructions once the size calculation disappears.

I will submit a PR at some point.

Is Scylladb using LTO at all?

mykaul:
> Is Scylladb using LTO at all?

Even PGO (https://github.com/scylladb/scylladb/tree/master/pgo )

avikivity:
> > That said, I don't see what we can do. We can't ignore > 256 vcpu machines.
> 
> Yup.
> 
> One option is to make some part of the tradeoff a configurable compile-time option. E.g., maybe I want max 16 GB per shard, which would give my access to up to 1024 cores while still having a static bit allocation. Or maybe I want fully dynamic as you've done.


I don't think it's worth a compile-time option. It adds some instructions yes, but we have to weigh the extra complexity too. 5 instructions, even in this hottest of paths, aren't worth yet another configuration.

And it's not like some Seastar applications will be for embedded. Large machines are everywhere, and Seastar likes to use whole machines.

> 
> Another option is to do a quick check of "are we in a standard (i.e., 8 bits of CPU) configuration as part of of the fast-path eligibility check we already do, it can probably even be zero cost by rolling it into another variable, and then the fast path can use the hard-coded shifts, and the > 256 vCPU case uses a (still fast) slow path.

You mean `template <bool cpu_id_bits_equals_8> memory.cc`? You'll lose a lot more on the runtime dispatch.

btw, where are those instructions spent? I don't think the allocation path is changed, and in the free path, instead of a 64-bit constant we have a 64-bit thread-local. I would have guessed zero extra instructions.

> 
> > btw, we could try exposing the small pool allocator where we constexpr-calculate the small pool index for things like continuations and coroutines, so we can skip those for both the allocation and deallocation paths.
> 
> We are using LTO, and I've had luck with this w/o the need to expose it: and it works for all allocations, not just ones tried to a specific allocator. You can use __builtin_constexpr to only inline in the case the size is known (which it is, most of the time), so all the calculation goes away: there is ~no size overhead in the inlined part: it just calls a different entry point with the calculation already done. 

Wow, though expected, now that I think of it.

That's the "strictly better" option. Arguably inlining a bit more of the fast path is interesting, e.g., the fast path as this is only a handful of instructions once the size calculation disappears.


Yes. For continuations etc we can also avoid checking if we're on a reactor thread (alloc+free).

We may even drop trim (and instead rely on the reactor loop do do that). Or maybe avoiding the memory clobber in free isn't so important.

> 
> I will submit a PR at some point.
> 
> Is Scylladb using LTO at all?

yes

travisdowns:
> You mean `template <bool cpu_id_bits_equals_8> memory.cc`? You'll lose a lot more on the runtime dispatch.

Yeah, more or less. I don't think you would lose anything if you are in the `true` case, you would ensure that some check we are already doing on the fast path returns false if we aren't in that regime, so when you are, it would be identical code up front. It's not like an actual dispatch.

> btw, where are those instructions spent? I don't think the allocation path is changed, and in the free path, instead of a 64-bit constant we have a 64-bit thread-local. I would have guessed zero extra instructions.

I didn't check but it makes sense to me: the original `cpu_id_shift` and `cpu_id_and_mem_base_mask` were not 64 bits but small constants, they can be inlined right into the instruction and also enable cheaper forms of some instructions. 64-bit constants always need a separate load.

PR #3054, opened 2025-10-14
iotune: fix very long warm up duration on systems with high cpu count

robertbindar:
In an i7i.16xlarge instance iotune with default duration of 120s would run for 120s + 18s + 48 * 6s = 426s which blows most of our timeouts in Scylla setup scripts.

This patches changes the warmup period calculation so that for the sequential write step, it stays a total of the original max(3s,5% * duration) for all shards.

### DIFF
--- a/apps/iotune/iotune.cc
+++ b/apps/iotune/iotune.cc
@@ -337,11 +337,12 @@ class io_worker {
         unsigned _prev_requests = 0;
         std::chrono::duration<double> _duration;
         timer<> _tick;
+        bool _with_warmup;
 
         static constexpr auto period = 1s;
 
     public:
-        requests_rate_meter(std::chrono::duration<double> duration, std::vector<unsigned>& rates, const unsigned& requests)
+        requests_rate_meter(std::chrono::duration<double> duration, std::vector<unsigned>& rates, const unsigned& requests, bool with_warmup)
             : _rates()
             , _parent_rates(rates)
             , _requests(requests)
@@ -350,6 +351,7 @@ class io_worker {
                 _rates.push_back(_requests - _prev_requests);
                 _prev_requests = _requests;
             })
+            , _with_warmup(with_warmup)
         {
             _rates.reserve(256); // ~2 minutes
             if (duration > 4 * period) {
@@ -369,7 +371,7 @@ class io_worker {
             // logic is really independent of when we start counting requests in
             // `issue_request` itself so the first bucket might not be a full
             // measurement and cause skew otherwise.
-            size_t samples_to_drop = warmup_period(_duration) / period + 1;
+            size_t samples_to_drop = _with_warmup ? warmup_period(_duration) / period + 1 : 0;
             _rates.erase(_rates.begin(), _rates.begin() + std::min(samples_to_drop, _rates.size()));
             _parent_rates.insert(_parent_rates.end(), _rates.begin(), _rates.end());
         }
@@ -397,13 +399,13 @@ public:
         return iotune_clock::now() >= _end_load;
     }
 
-    io_worker(size_t buffer_size, std::chrono::duration<double> duration, std::unique_ptr<request_issuer> reqs, std::unique_ptr<position_generator> pos, std::vector<unsigned>& rates)
+    io_worker(size_t buffer_size, std::chrono::duration<double> duration, std::unique_ptr<request_issuer> reqs, std::unique_ptr<position_generator> pos, std::vector<unsigned>& rates, bool with_warmup = true)
         : _buffer_size(buffer_size)
-        , _start_measuring(iotune_clock::now() + std::chrono::duration<double>(warmup_period(duration)))
+        , _start_measuring(iotune_clock::now() + (with_warmup ? std::chrono::duration<double>(warmup_period(duration)) : 0s))
         , _end_measuring(_start_measuring + duration)
         , _end_load(_end_measuring + 10ms)
         , _last_time_seen(_start_measuring)
-        , _rr_meter(duration, rates, _requests)
+        , _rr_meter(duration, rates, _requests, with_warmup)
         , _pos_impl(std::move(pos))
         , _req_impl(std::move(reqs))
     {}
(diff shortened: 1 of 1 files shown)


apps/iotune/iotune.cc
```diff
@@ -321,9 +321,13 @@ class read_request_issuer : public request_issuer {
 };
 
 std::chrono::duration<double>
-warmup_period(std::chrono::duration<double> duration) {
+warmup_period(std::chrono::duration<double> duration, bool is_sequential) {
     auto min_warmup = 3s;
     auto five_percent = duration * 0.05;
+    if (is_sequential) {
```
xemul: How come it depends on the smp::count?
The seq-write workload runs like this:

```
                for (unsigned shard = 0; shard < smp::count; ++shard) {
                    write_bw += iotune_tests.write_sequential_data(shard, sequential_write_buffer_size, duration * 0.70 / smp::count).get();
                }
```

so it already calls write_sequential_data with `duration = 120s*70%/smp::count`. Next

```c++
future<io_rates> write_workload(...) {
    ...
    auto worker = std::make_unique<io_worker>(..., duration, ...);
```

and it's worker that adjusts itself with warmup_period, so the warmup should already account for smp::count

robertbindar: @xemul you're correct, duration is already divided by smp::count for each worker. I don't think we have other options though than
a) either give each worker a very small warmup period (in this case I think `duration=120s*0,7/(smp::count * smp::count)` is correct, the divide operation here needs to be dependent on smp::count again, not percentage because with percentage I might still exceed timeouts on instances which have crazy shards like i7i.48xlarge with 192).
b) run shards for seqwrite in parallel. I already have a patch for that since running experiments on i7i.

I like option b) more. What's your thoughts?

xemul: Running seqwrite in parallel doesn't make sense, it won't (well -- shouldn't) change the resulting measurement

robertbindar:
Pushed a version that does:
a) Runs a separate warmup workload before seqwrite job
b) disable warmup (warmup 0s) for seqwrite jobs

apps/iotune/iotune.cc
```diff
@@ -889,6 +920,9 @@ int main(int ac, char** av) {
 
                 iotune_tests.create_data_file().get();
 
+                fmt::print("Warming up the disk for sequential write job...\n");
+                iotune_tests.warm_up_sequential_data(sequential_write_buffer_size, std::max(3s, std::chrono::duration_cast<std::chrono::seconds>(duration * 0.05))).get();
```
xemul: It should better use `warmup_period()` helper rather than calculate the same math by hand

apps/iotune/iotune.cc
```diff
@@ -315,15 +325,23 @@ class read_request_issuer : public request_issuer {
...
+    }
+
     future<size_t> issue_request(uint64_t pos, char* buf, uint64_t size) override {
         return _file.dma_read(pos, buf, size);
     }
 };
 
 std::chrono::duration<double>
-warmup_period(std::chrono::duration<double> duration) {
+warmup_period(std::chrono::duration<double> duration, bool is_sequential, issuer_type issuer_type) {
```
xemul: Why is this complexity needed? Why can't `io_worker` constructor accept `bool with_warmup` flag and only the write_sequential_data() passes false to it (via write_workload())?

xemul:
Usually a patch title describe what the patch does, but not what a current state is (which is implied to be problematic, so it then implicitly means that the patch fixes it)

apps/iotune/iotune.cc
robertbindar: that's a better solution indeed, pushed a new version that addresses.

apps/iotune/iotune.cc
xemul: This concern still stands

apps/iotune/iotune.cc
```diff
@@ -641,7 +651,7 @@ class iotune_multi_shard_context {
     template <typename Fn>
     future<uint64_t> saturate(float rate_threshold, size_t buffer_size, std::chrono::duration<double> duration, Fn&& workload) {
         return _iotune_test_file.invoke_on(0, [this, rate_threshold, buffer_size, duration, workload] (test_file& tf) {
-            return (tf.*workload)(buffer_size, test_file::pattern::sequential, 1, duration, serial_rates).then([this, rate_threshold, buffer_size, duration, workload] (io_rates rates) {
+            return (tf.*workload)(buffer_size, test_file::pattern::sequential, 1, duration, serial_rates, true).then([this, rate_threshold, buffer_size, duration, workload] (io_rates rates) {
```
xemul, line 655: This `true` is likely not required, all workload methods would append `true` to it by default

apps/iotune/iotune.cc
```diff
@@ -529,15 +530,15 @@ class test_file {
         });
     }
 
-    future<io_rates> read_workload(size_t buffer_size, pattern access_pattern, unsigned max_os_concurrency, std::chrono::duration<double> duration, std::vector<unsigned>& rates) {
+    future<io_rates> read_workload(size_t buffer_size, pattern access_pattern, unsigned max_os_concurrency, std::chrono::duration<double> duration, std::vector<unsigned>& rates, bool _=true) {
```
xemul: This is to facilitate `(tf.*workload)(...)` below, but if `true` is removed from there, this dummy `_` can be removed too.

apps/iotune/iotune.cc
```diff
@@ -384,7 +386,6 @@ class io_worker {
     std::chrono::time_point<iotune_clock, std::chrono::duration<double>> _end_load;
     // track separately because in the sequential case we may exhaust the file before _duration
     std::chrono::time_point<iotune_clock, std::chrono::duration<double>> _last_time_seen;
-
```
xemul, line 387: nit: Not needed

apps/iotune/iotune.cc
```diff
@@ -397,13 +398,13 @@ class io_worker {
         return iotune_clock::now() >= _end_load;
     }
 
-    io_worker(size_t buffer_size, std::chrono::duration<double> duration, std::unique_ptr<request_issuer> reqs, std::unique_ptr<position_generator> pos, std::vector<unsigned>& rates)
+    io_worker(size_t buffer_size, std::chrono::duration<double> duration, std::unique_ptr<request_issuer> reqs, std::unique_ptr<position_generator> pos, std::vector<unsigned>& rates, bool with_warmup=true)
```
xemul: nit: Please, give `=` some space (around)

apps/iotune/iotune.cc
robertbindar: this one is definitely needed so signatures match when compiler expands tf.*workload.

apps/iotune/iotune.cc
robertbindar: this one shouldnt be needed, it's ugly, but compiler complains.

PR #3058, opened 2025-10-15
reactor: Remove unused try_sleep() method

xemul:
It was partially open-coded by faa171e30c and became unused since then.


xemul:

PR #3059, opened 2025-10-15
reactor: Move read_directory() to posix_file_impl

xemul:
This method is a thread-pool wrapper over raw readdir64 syscall. It's used only by posix_file_impl's list_directory() API, but this file impl can talk to thread-pool on its own (all the more so it already does so in many other ways).

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -549,6 +549,7 @@ public:
     future<> link_file(std::string_view oldpath, std::string_view newpath) noexcept;
     future<> chmod(std::string_view name, file_permissions permissions) noexcept;
 
+    [[deprecated("Use file::list_directory API instead")]]
     future<size_t> read_directory(int fd, char* buffer, size_t buffer_size);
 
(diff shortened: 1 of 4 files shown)


xemul:

nyh (commented):
This raises an interesting question on what sort of APIs should Seastar provide over the standard Linux/POSIX APIs. Should we just provide a thin facade to the Linux system calls or library functions - which uses futures and syscall threads in appropriate places - or should we invent our own new APIs from scratch?

It seems in this patch you are deleting an API of the first kind (read_directory() was a thin wrapper around the system call getdents) and leaving just the second kind. Which is fine, but did we ever decide that's what we want to do?

In any case, the patch itself looks good to me.

xemul:
That's correct question, but as far as this particular read_directory() API method is concerned it has one huge inconsistency. It accepts a file descriptor argument, while seastar itself doesn't (almost) provide a way to obtain. "Almost" because there is class file_desc that can be used, but there should rather be file_desc::readdir() file_desc::getdents() to be consistent.

PR #3062, opened 2025-10-20
perf_tests: Fix build with fmt 12.0.0

tchaikov:
The fmt library removed fmt::detail::compute_width in version 12.0.0. Rather than using the new internal functions (for_each_codepoint and display_width_of), implement a simpler solution that avoids relying on fmt's internal API entirely.

The only multi-byte UTF-8 characters used in perf_tests output are:
- ± (U+00B1, 2 bytes) in MAD percentages: " ±5.23%"
- µ (U+00B5, 2 bytes) in microsecond units: "µs"

Both display as 1 column width despite being 2 bytes. The new apparent_width() function uses std::ranges::search to count occurrences of these specific byte subsequences and calculates the correct display width.

This approach is simpler, more maintainable, works with any fmt version, and doesn't depend on fmt's internal implementation details.

### DIFF
--- a/tests/perf/perf_tests.cc
+++ b/tests/perf/perf_tests.cc
@@ -251,13 +251,36 @@ struct format_options {
     int precision = -1;    // -1 triggers adaptive reduction logic
 };
 
-// compute the apparent width of a utf-8 string, taking into account
-// that some characters (e.g. CJK ideographs) are wider than one column
-// and some (e.g. combining accents) are zero-width. Use a concept so we
-// accept either std::string_view directly or any type exposing data()/size()
-// (e.g. seastar::sstring) without creating multiple overloads.
+// compute the apparent width of a utf-8 string for terminal display.
+// The only multi-byte UTF-8 characters we use are ± (U+00B1) and µ (U+00B5),
+// both of which are 2 bytes but display as 1 column. All other characters are ASCII.
+// Use a concept so we accept either std::string_view directly or any type exposing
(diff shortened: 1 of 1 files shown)


tchaikov:
cc @travisdowns

travisdowns:
Thanks @tchaikov!

All these is just needed due to a the use of the `±` character which is 2 bytes, when the "deviation" column is added. I was a bit reluctant to use the `details` method in the first place, and I see now it may have been a mistake: arguably it might be simpler just to hardcode special handling just for this 1 character. All other characters are highly likely to be ASCII as they are the test name and then numeric columns.

I didn't actually know we supported fmt 12 (tried recently to use this but had some compiler errors). Is this part of fmt 12 support?

tchaikov:
@travisdowns Hi Travis, you're absolutely right! I've updated the PR to use a simpler approach.

Instead of relying on fmt's internal `detail` namespace functions, the new implementation just handles the two specific multi-byte UTF-8 characters we actually use:
  - `±` (U+00B1) in MAD percentages
  - `µ` (U+00B5) in microsecond units

Regarding fmt 12: it's not officially supported yet, but I'm working on compatibility since I develop on Arch Linux and Debian Sid which ship with fmt 12. This change (along with e5b109702df4b20d0dd45f397f0e94b08db85095 for another build issue) is part of that effort.

travisdowns:
> µ (U+00B5) in microsecond units

Oh right, that was actually the original reason I had to add it (in the same series).

Thanks again for these fixes!

PR #3063, opened 2025-10-21
Close http client connection if handler doesn't consume all chunked-encoded body

### LINKED ISSUE #2768: http::chunked_source_impl may cause problems

xemul:
If when handling an http reply the handler doesn't consume all the body, the subsequent request shouldn't receive leftover data from the previous response and fail. For content-length replies this case is handled by client itself by either consuming (and discarding) the remaining body bytes or by just closing the connection (#2762). For chunked-encoded bodies, the problem is fixed here. Since the length of the body in that case is not known, client cannot consider "read the remaining bytes" approach and just closes the connection.

Clearly, this problem is not specific only to the http client, and also exists in the http server. The way it's handled there is that server just tries to continue reading from the _body_ stream. If it reads an empty buffer, then everything is fine. If, however, it's still possible to read from the stream, it means the application didn't read all of it, and server just closes the connection. Client, however, consumes the unread bytes if there are few of them left in the socket in a hope that re-establishing new connection is more expensive that that. Again, for chunked-encoded body (this patch) this is unfortunately not a option.

refs #2767
fixes #2768

### DIFF
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -2284,23 +2284,23 @@ future<> test_client_close_connection(bool chunked) {
                     }
                     ++responses;
                     try {
-                      if (!chunked) {
-                        out.write(format("HTTP/1.1 200 OK\r\nHost: localhost\r\nContent-Length: {}\r\n\r\n", response_size)).get();
-                        out.flush().get();
-                      } else {
-                        out.write(format("HTTP/1.1 200 OK\r\nHost: localhost\r\nTransfer-Encoding: chunked\r\n\r\n{:x}\r\n", response_size)).get();
-                        out.flush().get();
-                      }
+                        if (!chunked) {
+                            out.write(format("HTTP/1.1 200 OK\r\nHost: localhost\r\nContent-Length: {}\r\n\r\n", response_size)).get();
+                            out.flush().get();
+                        } else {
+                            out.write(format("HTTP/1.1 200 OK\r\nHost: localhost\r\nTransfer-Encoding: chunked\r\n\r\n{:x}\r\n", response_size)).get();
+                            out.flush().get();
+                        }
                         out.write(sstring(response_size / 2, 'a')).get();
                         out.flush().get();
 
                         out.write(sstring(response_size / 2, 'a')).get();
                         out.flush().get();
 
-                      if (chunked) {
-                        out.write(format("\r\n0\r\n\r\n")).get();
-                        out.flush().get();
-                      }
+                        if (chunked) {
+                            out.write(format("\r\n0\r\n\r\n")).get();
+                            out.flush().get();
+                        }
                     } catch (...) {
                         break;
                     }


xemul:
upd:
- Fix constructing of sources with builtin remaining bytes. Somehow clang++ in debug mode mist-detects invalid object vptr:

```
/home/runner/work/seastar/seastar/include/seastar/http/internal/content_source.hh:53:51: runtime error: member access within address 0x503000044890 which does not point to an object of type 'seastar::httpd::internal::content_length_source_impl'
0x503000044890: note: object has invalid vptr
 01 00 00 00  be be be be be be be be  be be be be be be be be  be be be be be be be be  be be be be
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
```

include/seastar/http/internal/content_source.hh
```diff
@@ -40,15 +40,17 @@ namespace internal {
  * */
 class content_length_source_impl : public data_source_impl {
     input_stream<char>& _inp;
-    size_t _remaining_bytes = 0;
-    size_t* _consumed_bytes;
+    size_t _builtin_remaining_bytes = 0;
```
nyh, line 43: It's not possible to understand, or even guess, how _builtin_remaining_bytes differ from _remaining_bytes, what "builtin" refers to. Please either reconsider the names, or add a comment.

nyh: Looking at the code below, it seems you don't even use this _builtin_remaining_bytes. Can it just be removed entirely?
If not, I have no idea why not so a comment or rename is definitely needed. :-)

nyh:
I'm reviewing the actual code in your patch, but I want to make a more general comment:

Clearly, the problem you are trying to solve is not specific only to the http **client**, and also affects, or affected in the past, the http **server**, so it's worth looking at how it was solved (or not) in the server.

And we did solve it in the server using the following code in src/http/httpd.cc:

```c++
                        // If the handler did not read the entire request
                        // content, this connection cannot be reused so we
                        // need to close it (via "_done = true"). But we can't
                        // just check content_stream.eof(): It may only become
                        // true after read(). Issue #907.
                        return content_stream.read().then([this] (temporary_buffer<char> buf) {
                            if (!buf.empty()) {
                                _done = true;
                            }
                        });
```

The way this works is that we just try to continue reading from the stream. If we read the entire content-length or chunked-encoding-length or whatever (it supports both!), we get an empty buffer, and everything is fine. If, however, it's still possible to read from the stream, it means the application didn't read all of it, and we need to close the connection.

I don't know if this solution is perfect, but it's certainly much simpler and didn't require any changes to the input stream implementation and so on.

include/seastar/http/internal/content_source.hh
```diff
@@ -99,10 +99,12 @@ class chunked_source_impl : public data_source_impl {
         // references to fields in the request structure
         std::unordered_map<sstring, sstring>& _chunk_extensions;
         std::unordered_map<sstring, sstring>& _trailing_headers;
+        size_t& _remaining_bytes;
```
nyh, line 102: The above commit message suggests this isn't remaining_bytes, it's more of a zero or one. So please add a comment explaining what it is, or even rename it to something like _has_remaining_bytes.

include/seastar/http/internal/content_source.hh
```diff
@@ -211,10 +214,15 @@ class chunked_source_impl : public data_source_impl {
     };
     input_stream<char>& _inp;
     chunk_parser _chunk;
+    size_t _builtin_remaining_bytes = 0;
```
nyh, line 217: Again, I have no idea what is "builtin" remaining bytes.

tests/unit/httpd_test.cc
```diff
@@ -2314,6 +2314,10 @@ SEASTAR_TEST_CASE(test_client_close_connection) {
     });
 }
 
+SEASTAR_TEST_CASE(test_client_close_connection_content_length) {
+    return test_client_close_connection(false);
+}
```
nyh, line 2335: I just noticed that our client tests are in the file `httpd_test.cc` (where the "d" means daemon - i.e., server). That's strange.

include/seastar/http/internal/content_source.hh
xemul: This variable is needed to make `size_t& _remaining_bytes` reference something in case the class is constucted without "external" remaining bytes counter

include/seastar/http/internal/content_source.hh
xemul: Caller (http client) passes `size_t reply::left_content_length` as the target for that reference. The source needs to store here how many bytes are left unread, but since in chunked source it's not possible to find it out, it stores here either 0 (all read) or "infinity" (something is left to be read, but the exact amount is not known). I agree it's not extremely elegant, but how else can it be unified with content-length source?

include/seastar/http/internal/content_source.hh
xemul: When chunked download source is constructed without "external" size_t variable, the `size_t& chunk_parser::remaining_bytes` needs to reference some variable, and this is the one. Reference can't be left uninitialized

tests/unit/httpd_test.cc
xemul: That's strage, yes :( The excuse for that is -- when http client appeared its tests re-used some code form http server test thus they were added to httpD file. Since then new tests just followed it. I will split this test file (or just rename it into http_test) as a followup

xemul:
> I'm reviewing the actual code in your patch, but I want to make a more general comment:
>  Clearly, the problem you are trying to solve is not specific only to the http client, and also affects, or affected in the past, the http server, so it's worth looking at how it was solved (or not) in the server.

Yes, you're right. However, in client this handling works differently. In server the connection is closed even if there's a single byte left unread from the body. In client we wanted to keep connections alive harder, so in case body source reports that there's less than certain amount of bytes left to be read from the reply, the client just does so (reads those remaining bytes) but throws the buffers instantly and then returns the connection back to pool (#2762)

include/seastar/http/internal/content_source.hh
```diff
@@ -40,15 +40,17 @@ namespace internal {
...
     input_stream<char>& _inp;
-    size_t _remaining_bytes = 0;
-    size_t* _consumed_bytes;
+    size_t _builtin_remaining_bytes = 0;
+    size_t& _remaining_bytes;
 public:
-    content_length_source_impl(input_stream<char>& inp, size_t length, size_t* consumed_bytes)
-        : _inp(inp), _remaining_bytes(length), _consumed_bytes(consumed_bytes) {
+    content_length_source_impl(input_stream<char>& inp, size_t length, size_t& remaining)
+        : _inp(inp), _remaining_bytes(remaining)
```
kreuzerkrieg, line 47: I dont follow here, why you are first initialize the reference in member initialization list and the update the value of the `remaining` argument effectively overwriting whatever was stored there?

include/seastar/http/internal/content_source.hh
```diff
@@ -40,15 +40,17 @@ namespace internal {
...
+    {
+        _remaining_bytes = length;
     }
 
     content_length_source_impl(input_stream<char>& inp, size_t length)
-        : content_length_source_impl(inp, length, nullptr) {
+        : _inp(inp), _builtin_remaining_bytes(length), _remaining_bytes(_builtin_remaining_bytes) {
```
kreuzerkrieg, line 53: why not just call the above constructor overload? something like `content_length_source_impl(inp, length, _builtin_remaining_bytes)`

src/http/client.cc
```diff
@@ -423,15 +423,15 @@ future<> client::do_make_request(connection& con, const request& req, reply_hand
 
         auto in = req._method != "HEAD" ? con.in(rep) : input_stream<char>(data_source(std::make_unique<skip_body_source>(rep)));
         return handle(rep, std::move(in)).then([this, reply = std::move(reply), &con] {
-            if (reply->content_length > reply->consumed_content) {
-                auto bytes_left = reply->content_length - reply->consumed_content;
+            if (reply->left_content_length) {
```
kreuzerkrieg: implicit conversion from numeric to 'bool'

include/seastar/http/internal/content_source.hh
kreuzerkrieg: nit: `the client either skips those bytes from the socket` should be `the client either drains those bytes from the socket`

tests/unit/httpd_test.cc
```diff
@@ -2295,7 +2295,7 @@ future<> test_client_close_connection(bool chunked) {
                         break;
                     }
                 }
-                out.close().get();
+                out.close().handle_exception([](auto ex){}).get();
```
kreuzerkrieg: "it may happen that client had already closed its end and there's nothing to flush into" if it happens we will get some concrete error type/message, right? why not to check for it to prevent cases when something else is going wrong and we just ignore it?

tests/unit/httpd_test.cc
```diff
@@ -2244,17 +2244,17 @@ SEASTAR_THREAD_TEST_CASE(test_http_with_broken_wire) {
 }
 
 future<> test_client_close_connection(bool chunked) {
```
kreuzerkrieg, line 2246: nit: "response writer write" should be "response writer writes" I guess

include/seastar/http/internal/content_source.hh
xemul: Before the patch we have `_remaining_bytes(length)`, now we need to effectively do the same

include/seastar/http/internal/content_source.hh
xemul: I tried, but debug clang build [detected](https://github.com/scylladb/seastar/pull/3063#issuecomment-3427190190) this as an error

src/http/client.cc
xemul: Why not?

tests/unit/httpd_test.cc
xemul: Maybe :( I forgot what exactly went wrong here, maybe this patch can be dropped. I'll revisit

src/http/client.cc
kreuzerkrieg: except it is ugly?
https://clang.llvm.org/extra/clang-tidy/checks/readability/implicit-bool-conversion.html

tests/unit/httpd_test.cc
xemul: It's system_error(EPIPE). Added specific handling in next version

tests/unit/httpd_test.cc
xemul: Not quite. I'll rephrase

include/seastar/http/internal/content_source.hh
xemul: > nit: `the client either skips those bytes from the socket` should be `the client either drains those bytes from the socket`

Fixed in next version

src/http/client.cc
xemul: Changed to check `if (left_content_length > 0)`

xemul:
upd:
- rebased (no conflicts)
- rephrased some comments
- check for int > 0 explicitly
- catch specific server socket closing exception in test

kreuzerkrieg (approved):
LGTM

xemul:
Updated PR comment with comparison of how this situation is handled for server

nyh:
> > I'm reviewing the actual code in your patch, but I want to make a more general comment:
> > Clearly, the problem you are trying to solve is not specific only to the http client, and also affects, or affected in the past, the http server, so it's worth looking at how it was solved (or not) in the server.
> 
> Yes, you're right. However, in client this handling works differently. In server the connection is closed even if there's a single byte left unread from the body. In client we wanted to keep connections alive harder, so in case body source reports that there's less than certain amount of bytes left to be read from the reply, the client just does so (reads those remaining bytes) but throws the buffers instantly and then returns the connection back to pool (#2762)

By the way, you could (and I'm not saying you must) use the same technique as we did in the server also in the client to achieve what you wanted:
1. Try to read some fixed amount (say 1MB) from the remaining body and descard it.
2. If you managed to read less than 1MB (0 or anything less than 1MB), congrats, the socket is reusable.
3. If you read the full 1MB it means the response had more than 1MB left and since you don't want to continue reading it all, close the connection.

The benefit is that this is very easy change and doesn't require changing any chunked-encoding or content-length specific code.

But this was just a thought. You don't have to do it this way, and your approach is valid too.

include/seastar/http/internal/content_source.hh
nyh: It doesn't help that you write this answer in a review comment, not a comment. How will anybody who reads this in the future understand?

include/seastar/http/internal/content_source.hh
nyh: My point is that there is a hidden knowledge that this variable will get 0 or infinity - not anything between - in the chunked encoding case. Why should this knowledge be hidden in the commit message that nobody will ever see, and not in a comment in the code?

include/seastar/http/internal/content_source.hh
nyh: Yes, now I understand. I didn't understand this at all before your explanation. This place is even less clear than the other place I asked about, because there you actually see the reference variable and may guess it points to this "builtin" one (although I didn't guess!) but here you don't even see a reference variable. In **this** code, why not just call it _remaining_bytes and that's it?

PR #3065, opened 2025-10-21
tests: Move thread_context_switch_test into perf/

xemul:
It's measuring some timing, no need for it to be in unit/ section


xemul:

xemul:

PR #3066, opened 2025-10-21
headers: Remove smp.hh from app-template.hh

xemul:
It needs it for std::shared_ptr<smp> field, but can just go with forward declaration. The smp_options.hh carried the resource::cpuset indirectly via this inclusion, no it needs to include resource.hh explicitly.

### DIFF
--- a/include/seastar/core/smp_options.hh
+++ b/include/seastar/core/smp_options.hh
@@ -21,8 +21,9 @@
 
 #pragma once
 
-#ifndef SEASTAR_MODULE
+#include <seastar/core/resource.hh>
 #include <seastar/util/program-options.hh>
+#ifndef SEASTAR_MODULE
 #include <seastar/util/modules.hh>
 #include <string>
 #endif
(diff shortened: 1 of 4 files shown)


include/seastar/core/smp_options.hh
```diff
@@ -21,8 +21,9 @@
 
 #pragma once
 
-#ifndef SEASTAR_MODULE
+#include <seastar/core/resource.hh>
 #include <seastar/util/program-options.hh>
+#ifndef SEASTAR_MODULE
```
avikivity, line 26: Why did the #ifndef location change? Typically its scope contains all seastar includes. Not that I understand it.

xemul: When it compiles src/seastar.cc it includes app-template.hh (line 168) which includes smp_options.hh (this file, line 29) which then needs declarations of resource::cpuset and seastar::program_options from those two headers, which came out of SEASTAR_MODULE scope in app-template.hh before

PR #3067, opened 2025-10-22
BUILD FIX: memory max constant in allocator

travisdowns:
After recent changes to memory.cc, it fails to compile as standard
libraries vary in their return value for std::bit_width. Initial C++20
behavior was to return an integer of the same signeness as the input,
but later standards specify an `int` return type.

This patch changes the variable `w` to be an `int` from auto so it
works everywhere.

### DIFF
--- a/src/core/memory.cc
+++ b/src/core/memory.cc
@@ -1873,7 +1873,7 @@ static long mbind(void *addr,
 static
 unsigned
 cpu_id_bits(unsigned nr_shards) {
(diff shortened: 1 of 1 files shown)


travisdowns:
I see, the problem is that https://cplusplus.github.io/LWG/issue3656 changed the return type from same as input to always `int`, so standard libs in the wild vary. So the right fix is to use int instead of auto.

PR #3068, opened 2025-10-23
Prometheus metrics handler optimization

travisdowns:
This series focuses on improving performance in the prometheus text handler.

@StephanDollberg noticed some time ago that a large amount of tail latency in Redpanda actually occurs due to the metric scrape itself, in scenarios where there are many metrics (think 10K per shard or more). Of course, this periodic latency spike during scrape is at exactly the wrong frequency to be visible as a spike in the metrics: it just get averaged out. You could only see it clearly if you have a second source of latency data (e.g., client metrics, or a faster scrape targeting fewer metrics).

Based on profiling, cost of the scrape is largely CPU time spent in in the formatting/serialization code of the metrics handler. Earlier, a metrics_perf bench was created and even in a "microbench" the per-series timings were similar. So we have a high degree of confidence that improvements to this bench will flow through to the real-world scenario.

So in order to improve the situation, we make some optimizations to the metrics handler. The overall result is: based on "middle of the road" benchmark which we think is relatively aligned to a typical metrics shape, we see a 7x to 9x improvement in instruction count and runtime. Other benchmarks do better or worse, but almost all are significantly improved.

See the individual changes for details, and in some cases per-change benchmark results, but there is an overall summary:

Aggregate summary:
	Runtime speedup (arithmetic mean of factors): 6.07x
	Runtime speedup (median of factors): 7.21x
	Allocation reduction (mean percentage): -73.9%
	Instruction reduction (geometric mean factor): 5.54x fewer
	Cycle reduction (geometric mean factor): 5.59x fewer

| Test | Before Allocs | After Allocs | Alloc Δ | Instr Before | Instr After | Instr Δ (x fewer) | Cycles Before | Cycles After | Cycle Δ (x fewer) |
|------|--------------:|-------------:|--------:|-------------:|------------:|------------------:|--------------:|-------------:|------------------:|
| few_metrics | 25.1 | 22.1 | -12.0% | 20300 | 8690 | 2.33x | 6920 | 3520 | 1.97x |
| large_families | 4.17 | 0.165 | -96.0% | 11000 | 1310 | 8.43x | 2630 | 325 | 8.11x |
| large_families_int | 4.17 | 0.165 | -96.0% | 10200 | 839 | 12.1x | 2400 | 235 | 10.2x |
| large_families_int_aggr | 8.16 | 0.158 | -98.1% | 2710 | 1010 | 2.69x | 745 | 274 | 2.72x |
| many_families_int | 10.3 | 7.27 | -29.2% | 12700 | 2770 | 4.60x | 3160 | 694 | 4.55x |
| many_families_int_aggr | 20.3 | 12.3 | -39.4% | 9960 | 5050 | 1.97x | 2660 | 1290 | 2.05x |
| middle_ground | 4.75 | 0.846 | -82.2% | 10800 | 1500 | 7.21x | 2610 | 362 | 7.22x |
| middle_ground_int | 4.75 | 0.845 | -82.2% | 9970 | 1040 | 9.63x | 2390 | 278 | 8.59x |
| histogram | 3.63 | 0.092 | -97.5% | 12700 | 1560 | 8.19x | 3210 | 381 | 8.42x |
| histogram_aggr | 0.545 | 0.113 | -79.3% | 884 | 241 | 3.66x | 240 | 59.4 | 4.04x |

Legend: 🟩 excellent (runtime ≥8x or alloc reduction ≥80%), 🟨 good (runtime ≥4x or alloc reduction ≥50%), 🟧 moderate (runtime ≥2x or alloc reduction ≥25%), 🟥 mild (<2x or alloc reduction <25%). Instruction/cycle factors use same thresholds as runtime.

| Test | Runtime Speedup | Alloc Reduction | Instr Reduction | Cycle Reduction |
|------|----------------:|----------------:|----------------:|----------------:|
| few_metrics | 🟨 1.84x | 🟧 12.0% | 🟨 2.33x | 🟨 1.97x |
| large_families | 🟩 8.10x | 🟩 96.0% | 🟩 8.43x | 🟩 8.11x |
| large_families_int | 🟩 10.21x | 🟩 96.0% | 🟩 12.10x | 🟩 10.23x |
| large_families_int_aggr | 🟨 2.72x | 🟩 98.1% | 🟨 2.69x | 🟨 2.72x |
| many_families_int | 🟨 4.55x | 🟧 29.2% | 🟨 4.60x | 🟨 4.55x |
| many_families_int_aggr | 🟧 2.05x | 🟧 39.4% | 🟧 1.97x | 🟧 2.05x |
| middle_ground | 🟩 7.21x | 🟩 82.2% | 🟩 7.21x | 🟩 7.22x |
| middle_ground_int | 🟩 8.56x | 🟩 82.2% | 🟩 9.63x | 🟩 8.59x |
| histogram | 🟩 8.42x | 🟩 97.5% | 🟩 8.19x | 🟩 8.42x |
| histogram_aggr | 🟨 4.04x | 🟩 79.3% | 🟨 3.66x | 🟨 4.04x |

### DIFF
--- a/src/core/prometheus-impl.hh
+++ b/src/core/prometheus-impl.hh
@@ -19,12 +19,75 @@
 
 
 #include <boost/container_hash/hash_fwd.hpp>
+#include <ranges>
 #include <seastar/core/metrics.hh>
 #include <seastar/core/metrics_api.hh>
+#include <string_view>
 
 #include <unordered_map>
 #include <vector>
 #include <string>
+
+namespace seastar::prometheus {
+
+
+namespace internal {
+// return true if label_name appears in aggr_labels
+inline bool is_aggregated(const std::vector<std::string>& aggr_labels, std::string_view label_name) {
+    return std::find_if(aggr_labels.begin(), aggr_labels.end(), [&](const std::string_view& lhs) {
+        return lhs == label_name;
+    }) != aggr_labels.end();
+}
+}
+
+struct label_key {
+    using label_list = std::vector<std::string>;
+
+    label_key() {
+        // only the scratch key is created with this constructor, we want to reserve
+        // some reasonable amount to avoid a bunch of small allocations for the first
+        // key (which is sometimes the only key)
+        key.reserve(200);
+    }
+
+    label_key(const metrics::impl::labels_type& labels, const label_list& aggr_labels) {
+        construct(labels, aggr_labels);
+    }
+
+    void construct(const metrics::impl::labels_type& labels, const label_list& aggr_labels) {
+        key.clear();
+        for (auto& [lkey, lvalue] : labels) {
+            std::string_view key_view = lkey;
+            if (!internal::is_aggregated(aggr_labels, key_view)) {
+                key += key_view;
+                key += '\n';
+                key += lvalue.value();
+                key += '\n';
+            }
+        }
+
+        hash = std::hash<std::string>{}(key);
+    }
+
+    bool operator==(const label_key& o) const noexcept {
+        return hash == o.hash && key == o.key;
+    }
+
+    std::string key;
+    size_t hash = 0;
+};
+
+} // namespace seastar::prometheus
+
+namespace std {
+template <>
+struct hash<seastar::prometheus::label_key> {
+    size_t operator()(const seastar::prometheus::label_key& lk) const noexcept {
+        return lk.hash;
+    }
+};
+}
+
 
 namespace seastar::prometheus {
 
@@ -42,10 +105,20 @@ namespace seastar::prometheus {
  */
 class metric_aggregate_by_labels {
     using labels_type = metrics::impl::labels_type;
-    std::vector<std::string> _labels_to_aggregate_by;
-    std::unordered_map<labels_type, seastar::metrics::impl::metric_value> _values;
+
+    struct labels_value {
+        labels_type labels;
+        metrics::impl::metric_value m;
+    };
+
 public:
-    metric_aggregate_by_labels(std::vector<std::string> labels) : _labels_to_aggregate_by(std::move(labels)) {
+    using label_list_type = std::vector<std::string>;
+    using map_type = std::unordered_map<label_key, labels_value>;
+    const label_list_type& _labels_to_aggregate_by;
+    label_key scratch_key;
+    map_type _values;
+public:
+    metric_aggregate_by_labels(const label_list_type& labels) : _labels_to_aggregate_by(labels) {
     }
     /*!
      * \brief add a metric
@@ -56,24 +129,27 @@ public:
      * The metric would be added to the aggregated metric with labels {'name':'myhist'}.
      *
      */
-    void add(const seastar::metrics::impl::metric_value& m, labels_type labels) noexcept {
-        for (auto&& l : _labels_to_aggregate_by) {
-            labels.erase(l);
-        }
-        std::unordered_map<labels_type, seastar::metrics::impl::metric_value>::iterator i = _values.find(labels);
-        if ( i == _values.end()) {
-            _values.emplace(std::move(labels), m);
+    void add(const seastar::metrics::impl::metric_value& m, const labels_type& input_labels) noexcept {
+        scratch_key.construct(input_labels, _labels_to_aggregate_by);
+        auto i = _values.find(scratch_key);
+        if (i == _values.end()) {
+            labels_type labels;
+            for (auto& l : input_labels) {
+                if (!internal::is_aggregated(_labels_to_aggregate_by, l.first)) {
+                    labels.insert(l);
+                }
+            }
+            _values.emplace(scratch_key, labels_value{std::move(labels), m});
         } else {
-            i->second += m;
+            i->second.m += m;
         }
     }
-    const std::unordered_map<labels_type, seastar::metrics::impl::metric_value>& get_values() const noexcept {
+    const auto& get_values() const noexcept {
         return _values;
     }
     bool empty() const noexcept {
         return _values.empty();
     }
 };
-
 
 }
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -27,6 +27,7 @@
 
 #include <seastar/core/future.hh>
 #include <seastar/core/iostream.hh>
+#include <seastar/core/metrics.hh>
 #include <seastar/core/metrics_api.hh>
 #include <seastar/core/prometheus.hh>
 #include <seastar/core/scollectd.hh>
@@ -942,11 +943,14 @@ future<> write_context::write_text_representation() {
             if (!aggregated_values.empty()) {
                 for (auto&& h : aggregated_values.get_values()) {
                     s.clear();
-                    if (h.second.type() == mi::data_type::HISTOGRAM) {
-                        write_histogram(s, ctx, name, h.second.get_histogram(), h.first);
+                    // Labels are already filtered (aggregated labels removed)
+                    auto& labels = h.second.labels;
+                    auto& value = h.second.m;
+                    if (value.type() == mi::data_type::HISTOGRAM) {
+                        write_histogram(s, ctx, name, value.get_histogram(), labels);
                     } else {
-                        write_name_and_labels(s, name, "", h.first, ctx);
-                        write_value_as_string(s, h.second);
+                        write_name_and_labels(s, name, "", labels, ctx);
+                        write_value_as_string(s, value);
                     }
                     out.write(s.data(), s.size()).get();
                     thread::maybe_yield();
@@ -978,8 +982,8 @@ future<> write_context::write_protobuf_representation() {
                 empty_metric = false;
             }
         });
-        for (auto& [labels, value] : aggregated_values.get_values()) {
-            fill_metric(mtf, value, labels, ctx);
+        for (auto& [_, value] : aggregated_values.get_values()) {
+            fill_metric(mtf, value.m, value.labels, ctx);
             empty_metric = false;
         }
         if (empty_metric) {
--- a/tests/unit/prometheus_text_test.cc
+++ b/tests/unit/prometheus_text_test.cc
@@ -444,11 +444,11 @@ SEASTAR_TEST_CASE(test_metric_aggregate_by_labels_basic_counter) {
     // Check that the aggregated value has the correct labels (without "shard")
     auto it = values.begin();
     // Labels are already filtered (aggregated labels removed)
-    BOOST_REQUIRE_EQUAL(it->first.size(), 1);
-    BOOST_REQUIRE_EQUAL(it->first.at("name").value(), "counter1");
+    BOOST_REQUIRE_EQUAL(it->second.labels.size(), 1);
+    BOOST_REQUIRE_EQUAL(it->second.labels.at("name").value(), "counter1");
 
     // Check that values were summed: 123 + 456 = 579
-    BOOST_REQUIRE_EQUAL(it->second.d(), 579);
+    BOOST_REQUIRE_EQUAL(it->second.m.d(), 579);
 
     return make_ready_future<>();
 }
@@ -483,10 +483,10 @@ SEASTAR_TEST_CASE(test_metric_aggregate_by_labels_multiple_labels) {
 
     auto it = values.begin();
     // Labels are already filtered (aggregated labels removed)
-    BOOST_REQUIRE_EQUAL(it->first.size(), 2);
-    BOOST_REQUIRE_EQUAL(it->first.at("name").value(), "metric1");
-    BOOST_REQUIRE_EQUAL(it->first.at("type").value(), "typeA");
-    BOOST_REQUIRE_EQUAL(it->second.d(), 300);
+    BOOST_REQUIRE_EQUAL(it->second.labels.size(), 2);
+    BOOST_REQUIRE_EQUAL(it->second.labels.at("name").value(), "metric1");
+    BOOST_REQUIRE_EQUAL(it->second.labels.at("type").value(), "typeA");
+    BOOST_REQUIRE_EQUAL(it->second.m.d(), 300);
 
     return make_ready_future<>();
 }
@@ -519,12 +519,12 @@ SEASTAR_TEST_CASE(test_metric_aggregate_by_labels_different_label_values) {
     int count_metric1 = 0;
     int count_metric2 = 0;
     for (auto it = values.begin(); it != values.end(); ++it) {
-        if (it->first.at("name").value() == "metric1") {
+        if (it->second.labels.at("name").value() == "metric1") {
             count_metric1++;
-            BOOST_REQUIRE_EQUAL(it->second.d(), 100);
-        } else if (it->first.at("name").value() == "metric2") {
+            BOOST_REQUIRE_EQUAL(it->second.m.d(), 100);
+        } else if (it->second.labels.at("name").value() == "metric2") {
             count_metric2++;
-            BOOST_REQUIRE_EQUAL(it->second.d(), 200);
+            BOOST_REQUIRE_EQUAL(it->second.m.d(), 200);
         }
     }
     BOOST_REQUIRE_EQUAL(count_metric1, 1);
@@ -593,7 +593,7 @@ SEASTAR_TEST_CASE(test_metric_aggregate_by_labels_real_counters) {
     BOOST_REQUIRE_EQUAL(values.size(), 1);
 
     auto it = values.begin();
-    BOOST_REQUIRE_CLOSE(it->second.d(), 580.2, 0.001);
+    BOOST_REQUIRE_CLOSE(it->second.m.d(), 580.2, 0.001);
 
     return make_ready_future<>();
 }
@@ -634,7 +634,7 @@ SEASTAR_TEST_CASE(test_metric_aggregate_by_labels_histograms) {
     BOOST_REQUIRE_EQUAL(values.size(), 1);
 
     auto it = values.begin();
-    const auto& aggregated_hist = it->second.get_histogram();
+    const auto& aggregated_hist = it->second.m.get_histogram();
 
     // Check aggregated histogram values
     BOOST_REQUIRE_EQUAL(aggregated_hist.sample_count, 30);
@@ -665,7 +665,7 @@ SEASTAR_TEST_CASE(test_metric_aggregate_by_labels_same_metric_added_twice) {
     BOOST_REQUIRE_EQUAL(values.size(), 1);
 
     auto it = values.begin();
-    BOOST_REQUIRE_EQUAL(it->second.d(), 250);
+    BOOST_REQUIRE_EQUAL(it->second.m.d(), 250);
 
     return make_ready_future<>();
 }


avikivity:
Wow

include/seastar/core/metrics.hh
```diff
@@ -261,6 +261,51 @@ SEASTAR_MODULE_EXPORT_END
...
 
+// Concept that matches std::string or sstring types.
+template<typename T>
+concept string_like = std::same_as<T, std::string> || std::same_as<T, sstring>;
+
+// A escaped label value.
+//
+// This is "just" as a string, with the additional invariant that it has already been
+// escaped, and this type preserves that invariant.
+struct escaped_string {
```
avikivity: Since it encapsulates its members, it's a class, not a struct. Also please move the data members to the top as is common.

include/seastar/core/metrics.hh
```diff
@@ -261,6 +261,51 @@ SEASTAR_MODULE_EXPORT_END
...
+    operator const sstring&() const {
+        return value();
+    }
+
+    bool operator==(const escaped_string& o) const = default;
+    auto operator<=>(const escaped_string& o) const = default;
+
+    bool operator==(const string_like auto& o) const {
+        return _value == o;
+    }
```
avikivity: Is this correct? A let's say o is a sstring and needs escaping, and *this is the escaped form. They're logically the same but will compare different.

What does "==" mean? the rendered form or the preimage?

src/core/prometheus-impl.hh
```diff
@@ -19,13 +19,76 @@
...
+#include <string_view>
 
 #include <unordered_map>
 #include <vector>
 #include <string>
 
+namespace seastar::prometheus {
+
+
+namespace {
```
avikivity: Unnamed namespaces in a header are transformed to different external names for each translation unit, causing bloat. Better use `namespace internal`.

avikivity:
Looks good. @amnonh please review.


I had an idea to optimize this to the hilt by preparing a template of the output (the entire output with spaces left for the numbers) similar to database prepared expressions. The template is invalidated every time metrics are added or removed, but that is rare.

The template is further compressed using gzip, but preventing the space left for the values from being considered. There's  huge amount of redundancy in the text format. The values are "compressed" as literals. The new positions of the values in the compressed templates are recorded.

When the time comes to use the template, it is copied somewhere and the values are formatted into the offsets computed for them, then the entire thing is sent. Rendering the values happens on the original shards, not the coordinator.

It's a large effort and probably not proportional to the gain, but would be a massive improvement since nothing is rendered at runtime except the actual values. Will also reduce bandwidth massively.

include/seastar/core/metrics.hh
travisdowns: Your repo, your rules :).

Fixed in push https://github.com/scylladb/seastar/commit/705c9e5da305947f42643e181ff00522f76e2628.

include/seastar/core/metrics.hh
travisdowns: I think it could go either way, if you consider it a transformed version of the string into a new domain, then it's wrong as you are comparing across domains so you should transform one side (e.g, escape the RHS) to be equal. That's probably the most correct view when swapping this into existing code.

OTOH if you consider them both the same domain but just enforcing a restriction on contained values (think "all ASCII string" as a clear example of this) then it's just a subset of the domain of unescaped strings and then arguably comparison is fine and has clear semantics.

Since this is a replacement of existing code where the string was previously escaped later, I think the first view is more correct, so I've just removed these operators in push https://github.com/scylladb/seastar/commit/705c9e5da305947f42643e181ff00522f76e2628. We used it only in a single place a test, so it was silly to add this in the first place.

src/core/prometheus-impl.hh
travisdowns: I think it's negative bloat since this is included by only one .cc file in seastar (and once more in a test), but I get the point this can change. Fixed in push https://github.com/scylladb/seastar/commit/705c9e5da305947f42643e181ff00522f76e2628.

travisdowns:
> I had an idea to optimize this to the hilt by preparing a template of the output (the entire output with spaces left for the numbers) similar to database prepared expressions. The template is invalidated every time metrics are added or removed, but that is rare.

Yeah I considered something along these lines (edit: it was a much less ambitious thing, I hadn't read the whole post yet).

Doing it for the entire output is not too easy as there are several runtime-dynamic choices in the loop, such as filtering, aggregation on or off, additional label in `ctx` which can affect the output, but even  doing it on a line-by-line basis (i.e., caching the whole line except the value) would speed things up dramatically again. I shied away because of the extra memory use. We already have a lot of metrics and memory is a problem so additional memory per metric I didn't want. _If_ is scollecd was not a think any more, we could think of making the cached form sort of the one thing we store, and we can still extract things like label names/values from there (e.g., list a list of string_view pointing into the seralized form) which is needed to implement e.g., filters.

However, I think the basic serialization is already fairly fast. In Redpanda we are almost always using metrics aggregation, so the relatively poorer performance of aggregation is a problem and makes further serializaiton improvements mostly useless. As @StephanDollberg points out, this could also be cached, e.g., doing the aggregation on when metrics need updating (or the aggregation status changes), which would probably make aggregation blazing fast. I only added the aggregated benchmarks relatively late: had I seen how slow they were from the start I might have tried that: but it required deeper surgery to the metrics stack (again, shared with scollectd) which I was more reluctant to do.

> The template is further compressed using gzip, but preventing the space left for the values from being considered. There's huge amount of redundancy in the text format. The values are "compressed" as literals. The new positions of the values in the compressed templates are recorded.

That's a really interesting idea. I also looked at sending with gzip encoding (very briefly), encoding every time in the usual way, but this only makes the CPU problem worse: but doing it rarely and reusing it sounds good. As you say, you can leave literal placeholders in the compressed output which can be overwritten later with bounded size. Probably have to hack an encoder to do this, h/e, I don't think off-the-shelf ones allow it (you could hack it by doing a full flush before/after every character in the placeholders but this would destroy compression.

> Will also reduce bandwidth massively.

Yup. Our metric scrapes sometimes exceed 100 MB for a single scrape.

src/core/prometheus-impl.hh
travisdowns: One interesting note: just dropping the `fmt_but` class into an anonymous namespace, instead of a named one, improved perf signitifantly, like 5% or more over the whole benchmark. Probably the visibility change affects the inlining heuristic? Since the compiler knows there are no other callers outside the current TU it can make a more accurate assessment of inlining bloat.

mykaul:
@travisdowns - I believe that @amnonh added (experimental?) support for the native protocol - did you try it out? I'm not sure it's (now/already) production / supported in Prometheus though.

travisdowns:
> @travisdowns - I believe that @amnonh added (experimental?) support for the native protocol - did you try it out? I'm not sure it's (now/already) production / supported in Prometheus though.

From what I've read, "native" protocol (i.e., protobuffers) is quite a bit slower to ingest on the prom side, believe it or not. Apparently they've put more effort in the text side. Customers almost exclusively use the text format as well, we've had ~zero requests for the pb format. This may change in the future as it becomes better supported (if it does, it sort of has a weird history: it was supported in the past in prom, then removed, then made a comeback lately when they introduced native summaries or whatever it is called).

I'll do a quick test to see what's faster on the seastar side.

travisdowns:
@mykaul here are some text vs protobuf results (on top of this patch):

```
test                                                      iters            runtime     allocs      tasks       inst     cycles
metrics_perf_fixture.test_middle_ground                 1000000    74.78ns ± 0.00%      0.846      0.001    1470.93      408.5
metrics_perf_fixture.test_middle_ground_protobuf        1000000   280.24ns ± 0.00%     22.943      0.001    5776.56     1530.4

metrics_perf_fixture.test_middle_ground_int             1000000    57.27ns ± 0.00%      0.845      0.001    1017.96      314.0
metrics_perf_fixture.test_middle_ground_int_protobuf    1000000   271.44ns ± 0.00%     22.943      0.001    5776.32     1487.4

metrics_perf_fixture.test_histogram                     2200000    75.59ns ± 0.00%      0.092      0.000    1531.51      414.7
metrics_perf_fixture.test_histogram_protobuf            2200000    18.82ns ± 0.00%      1.367      0.000     416.53      102.9

metrics_perf_fixture.test_histogram_aggr                2200000    10.91ns ± 0.00%      0.113      0.000     235.12       59.7
metrics_perf_fixture.test_histogram_aggr_protobuf       2200000     7.19ns ± 0.00%      0.216      0.000     145.14       39.1
```

So pb looks bad relative to text for vanilla metrics ("middle ground"), but is faster for histograms (probably, the histogram pb format is much more efficient and DRY).

Perhaps the pb side just needs some optimization love: before these changes it was more efficient even in the vanilla case by about 2x (results below are without this series):


```
test                                                      iters            runtime     allocs      tasks       inst     cycles
metrics_perf_fixture.test_middle_ground                 1000000   520.92ns ± 0.00%      4.746      0.002   10578.36     2842.9
metrics_perf_fixture.test_middle_ground_protobuf        1000000   273.82ns ± 0.00%     22.843      0.001    5774.79     1494.6

metrics_perf_fixture.test_middle_ground_int             1000000   468.54ns ± 0.00%      4.745      0.002    9735.21     2557.1
metrics_perf_fixture.test_middle_ground_int_protobuf    1000000   271.65ns ± 0.00%     22.843      0.001    5774.54     1485.0

metrics_perf_fixture.test_histogram                     2200000   618.67ns ± 0.00%      3.633      0.002   12946.17     3371.0
metrics_perf_fixture.test_histogram_protobuf            2200000    18.92ns ± 0.00%      1.362      0.000     416.44      102.0

metrics_perf_fixture.test_histogram_aggr                2200000    42.43ns ± 0.00%      0.545      0.000     905.29      229.4
metrics_perf_fixture.test_histogram_aggr_protobuf       2200000    10.12ns ± 0.00%      0.612      0.000     210.82       55.0
```

mykaul:
Thanks @travisdowns
I hope that it's also more efficient on the wire.

src/core/prometheus-impl.hh
avikivity: Hmm. We can ignore it and assume non-test users will use lto/pgo/-O3.

avikivity: (ignore the 5% loss, not the potential bloat from unnamed namespaces)

avikivity:
protobuf should also benefit from the prepared statement approach. Pre-format a buffer, record the offsets, copy the data into the offsets.

Regarding run-time choices, make these parameters part of the key used to look up the prepared statement. Since they change rarely, the cache size will be 1.

travisdowns:
> Regarding run-time choices, make these parameters part of the key used to look up the prepared statement. Since they change rarely, the cache size will be 1.

Fair, though the name?

> Thanks @travisdowns
> I hope that it's also more efficient on the wire.

For regular metrics a bit smaller (< 2x), for histograms, _much_ smaller (more than 20x).

```
Total bytes written: 2465570
metrics_perf_fixture.test_middle_ground              1000000    73.55ns ± 0.00%      0.846      0.001    1471.95      400.8

--

Total bytes written: 1759890
metrics_perf_fixture.test_middle_ground_protobuf     1000000   282.62ns ± 0.00%     22.943      0.001    5777.14     1556.2

==============

Total bytes written: 5273370
metrics_perf_fixture.test_histogram                  2200000    74.56ns ± 0.00%      0.092      0.000    1532.39      413.2

--

Total bytes written: 205090
metrics_perf_fixture.test_histogram_protobuf         2200000    18.35ns ± 0.00%      1.367      0.000     416.55      100.0
```

avikivity:
> > Regarding run-time choices, make these parameters part of the key used to look up the prepared statement. Since they change rarely, the cache size will be 1.
> 
> Fair, though the name?

name?

travisdowns:
> > > Regarding run-time choices, make these parameters part of the key used to look up the prepared statement. Since they change rarely, the cache size will be 1.
> > 
> > 
> > Fair, though the name?
> 
> name?

Odd. I _think_ I was trying to ask about the `?name=` filter you can specify but regardless this should still obey your "rarely changes" rule and in any case this is usually used to return a very small slice of metrics which will be fast any way as it is small.

travisdowns:
Push https://github.com/scylladb/seastar/commit/c64949472b36dc3cedb9ae10ec03d318dce6d021 adds some prom tests as @mykaul had asked about.

travisdowns:
Push https://github.com/scylladb/seastar/commit/797adb87e9c961b056a04ff3ac730682eb3b7539 [diff](https://github.com/scylladb/seastar/compare/c64949472b36dc3cedb9ae10ec03d318dce6d021..797adb87e9c961b056a04ff3ac730682eb3b7539) simplifies the "starts with `__`" check as suggested by @StephanDollberg. This was simpler and even saves 5 instructions per metric.

src/core/prometheus-impl.hh
travisdowns: > (ignore the 5% loss, not the potential bloat from unnamed namespaces)

Just to clarify, I did move this into a named namespace as you requested. The example I gave of the 5% loss was from a .cc file, so no bloat there.

include/seastar/core/metrics.hh
```diff
@@ -261,6 +261,41 @@ SEASTAR_MODULE_EXPORT_END
  */
 namespace impl {
 
+// A escaped label value.
+//
+// This is "just" as a string, with the additional invariant that it has already been
+// escaped, and this type preserves that invariant.
+class escaped_string {
+    sstring _value;
```
xemul, line 269: Should this be `const sstring _value`?

src/core/scollectd.cc
```diff
@@ -301,7 +303,7 @@ struct cpwriter {
         // Use lo-res ts for now, it is probably quite sufficient.
         put_cached(part_type::Plugin, group_name);
         // Optional
-        auto instance_id = labels.at(metrics::shard_label.name());
+        sstring instance_id = labels.at(metrics::shard_label.name());
```
xemul, line 306: Just curious -- why `auto instance_id` doesn't work and needs explicit type?

src/core/prometheus.cc
```diff
@@ -43,18 +43,178 @@
...
+    }
+
+    fmt_buf& operator<<(std::string_view str) {
+        return append(str);
+    }
+
+    void append_slowpath(std::string_view sv) {
+        for (auto c : sv) {
+            append(c);
+        }
```
xemul, line 185: Why not implement the `realloc_buffer_and_append(std::string_view str)` that would behave similarly to `r.b.and.a(char c)` -- grow the buffers with some reserve and std::copy the string into it?

src/core/prometheus.cc
```diff
@@ -875,7 +1044,7 @@ bool is_accept_protobuf(const std::string& accept) {
 class metrics_handler : public httpd::handler_base  {
     sstring _prefix;
     config _ctx;
-    static std::function<bool(const mi::labels_type&)> _true_function;
+    static const std::function<bool(const mi::labels_type&)> _true_function;
```
xemul: Is this `const` addition an essential part of this patch, or just happened to be here?

src/core/prometheus.cc
```diff
@@ -43,18 +43,178 @@
...
 namespace pm = io::prometheus::client;
 
 namespace mi = metrics::impl;
 
 using mi::labels_type;
 
+namespace {
+
+// A wrapper around a std::vector for efficient formatting and appending
+// of text data. Just using a std::vector is much slower.
```
xemul, line 107: Why is plain vector much slower? The fmt_buf implementation grows the vector "on demand" and keeps pointer to the last "unused" char in it. But isn't native vector's size vs capacity management enough to achieve the same goal?

src/core/prometheus-impl.hh
```diff
@@ -19,13 +19,76 @@
...
+#include <string_view>
 
 #include <unordered_map>
 #include <vector>
 #include <string>
 
+namespace seastar::prometheus {
+
+
+namespace details {
```
xemul: `namespace internal` maybe?

src/core/metrics.cc
```diff
@@ -220,6 +220,47 @@ static std::string get_unique_id() {
...
+/*
+ * Sanitizes the label value as per the line format rules
+ * > label_value can be any sequence of UTF-8 characters, but the backslash (\), double-quote ("), and
+ * > line feed (\n) characters have to be escaped as \\, \", and \n, respectively.
+ */
+inline sstring escape_label_value(sstring label_value) {
+    if (!label_needs_escaping(label_value)) {
+        return label_value;
+    }
+    std::string escaped;
```
xemul, line 243: It's probably worth reserving the escaped string size to be at least label_value size

include/seastar/core/metrics.hh
travisdowns: No, because operations like assignment are valid and useful.

src/core/scollectd.cc
travisdowns: After this change, `auto` will resolve to `escaped_string`, which will fail to compare with `sstring` below: so by being explicit that we want an sstring we avoid this problem.

amnonh:
@travisdowns, as a side note, if you have many histograms, you can switch to native histograms (we have a semi-match implementation with approximate_exponential_histogram).

When enabled on both sides (Seastar and Prometheus), the reporting layer would switch to protobuf and use native_histograms when available. This is not only more efficient on the Seastar and the network side, but on the Prometheus side.

travisdowns:
> @travisdowns, as a side note, if you have many histograms, you can switch to native histograms (we have a semi-match implementation with approximate_exponential_histogram).
> 
> When enabled on both sides (Seastar and Prometheus), the reporting layer would switch to protobuf and use native_histograms when available. This is not only more efficient on the Seastar and the network side, but on the Prometheus side.

I'm interested in this, but the details aren't exactly clear. Some folks aren't going to switch to protobuf on the prom side, but can we offer both endpoints (one for text, one for prom) in parallel? 

On the application side there's nothing I need to do with the metrics themselves if we are already using `HISTOGRAM` type metrics, right? Those are compatible with native histogram once pb is enabled, right?

src/core/prometheus.cc
travisdowns: Because in practice this path is taken either "never" or "almost never" as the buffer is cleared (but the memory re-used) for every metric output, and also starts large enough for a typical metric. So in the case no individual metric exceeds 1024 bytes, it never happens at all, or if the largest metric is say 10K, then once this buffer grows to 10K once during a request it will never grow again.

Since this function accounts for miniscule amount of total runtime, I just did the simplest and most obviously correct thing.

Willing to change it if it's a blocker.

travisdowns:
Push https://github.com/scylladb/seastar/commit/87121d671277789d0764efbe65649bf5d020f558 was a rebase to tip of dev.

src/core/prometheus.cc
travisdowns: It's an interesting question, which I do happen to have a ready-made answer for.

That's because the same question came up during internal review, I'll a summary of my answer from that internal review here, please excuse the formatting etc. Please LMK if anything is unclear.

# Efficient Appending to std::vector

The high level summary is you can't (or it is hard to) do efficient "appending" of an array of elements to a vector, because of the way vector does not expose uninitialised memory and that you have to have split resize/append operations.

You remember the append() function from fmt_buf, right? It's like:

```cpp
fmt_buf& append(std::string_view str) {
    if (str.size() <= remaining()) [[likely]] {
        std::copy(str.begin(), str.end(), _current);
        _current += str.size();
    } else {
        append_slowpath(str);
    }
    return *this;
}
```

How do you write that with vector? Probably:

```cpp
void append(std::string_view str) {
    // insert(end(), str.begin(), str.end());
    resize(size() + str.size());
    std::copy(str.begin(), str.end(), end() - str.size());
    // std::copy_n(str.begin(), str.size(), end() - str.size());
}
```

(Commented out variations result in very similar code)

Then check out the assembly for `buf.append("a")` for both styles of buf.

## Assembly Comparison

### fmt_buf

```asm
160:      	mov	rax, qword ptr [rdi]
163:      	cmp	qword ptr [rdi + 0x10], rax
167:      	je	0x170 <...+0x10>
169:      	mov	byte ptr [rax], 0x61
16c:      	inc	qword ptr [rdi]
16f:      	ret
170:      	lea	rdx, [rip]
	0000000000000173:  R_X86_64_PC32	.L.str-0x4
177:      	mov	esi, 0x1
17c:      	jmp	0x29f0 <...::fmt_buf::append_slowpath(...)>
181:      	nop	word ptr cs:[rax + rax]
```

That's about the best you can hope for (when standalone in a function): load the "current" and "end" pointers, see if they are equal (no room for 1 more char), and if not just append the char, increment the current pointer and get out. The slow path does the expected thing but we never go there.

### vec_buf

```asm
120:      	push	rbx
121:      	mov	rcx, qword ptr [rdi]
124:      	mov	rax, qword ptr [rdi + 0x8]
128:      	mov	rdx, rax
12b:      	sub	rdx, rcx
12e:      	cmp	rdx, -0x1

# is the current vector.size() == 1
132:      	je	0x147 <...+0x27>
134:      	mov	esi, 0x1
139:      	mov	rbx, rdi
13c:      	call	0x141 <...+0x21>
	000000000000013d:  R_X86_64_PLT32	std::vector<char>::_M_default_append(unsigned long)-0x4
141:      	mov	rax, qword ptr [rbx + 0x8]
145:      	jmp	0x153 <...+0x33>

# is the vector empty
147:      	cmp	rax, rcx
14a:      	je	0x153 <...+0x33>
14c:      	mov	qword ptr [rdi + 0x8], rcx
150:      	mov	rax, rcx
153:      	mov	byte ptr [rax - 0x1], 0x61
157:      	pop	rbx
158:      	ret
159:      	nop	dword ptr [rax]
```

The first part of the code is weird, probably a result of inlining, but it basically never executes, only if the vector has `size() == 1`, and the jump at 14a looks dead to me (jumps if `size() == 0` I think but we only got here when we determined `size() == 1` already), I'm probably missing something here.

But the main thing is it always goes down through the call to `_M_default_append` which is an out of line function which resizes the vector by the required amount, it involves a bunch of work and a further call to memcpy. This is the work which we move entirely off into the slow path in the other impl. So it takes like 50 instructions to add 1 character.

## Root Causes

Which is an outcome of:

1. Going down the `_M_default_append` path in the first place, this is partly in consequence of resize being separate from the append itself, so it needs to be unconditional
2. `_M_default_append` not being inlined (possibly because libstdc++ devs force it out of line), which makes it more expensive since now we lost context, and pay around round of function pro/epilog
3. resize()/append being split meaning we need to initalize the memory in the resize, so there is a memset in there, even though this memory is then immediately overwritten by the string during the append

libc++ does somewhat better because the `_M_default_append` doesn't get outlined. So it still does the memset etc but it's inlined as a 1 character 0-write, etc.

A lot of these are "constant additive const" stuff so this is worst when the string is short, like 10x worse: as the string gets very long the badness approaches 2:1. Since only the memset and copy are O(n) in the length and they both take the same number of instructions apparently (and fmt doesn't have the memset).

## Alternative Approaches

*An anonymous reviewer asked about using boost::vector with the uninit flag.*

No, I didn't try it with boost vector. I did try with `std::string::resize_and_overwrite` which sort of seems to solve some of the problems in practice. But it's not just about the init thing either, even if you remove the init most of the problem still exists (it just doesn't call the memset).

Using boost vector with the uninit thing does help:

- fmt_buf: 840 instructions
- vec_buf: 1940 instructions
- boost vec with uninit: 1243

This is `metrics_perf_fixture.test_large_families_int`.

src/core/prometheus.cc
travisdowns: It's not essential, I just noticed this should be const when I was evaluating how much the function eval costs (not enough to bother, so only the const was changed). Happy to remove it. I don't think it makes  a perf difference.

src/core/prometheus-impl.hh
travisdowns: Good call, changed to `internal` in latest push.

src/core/metrics.cc
travisdowns: I went with size + 1, since we call this for strings that need escaping so the minimum size is input size + 1 as there must be at least one character needing escaping, which adds 1 to size. Updated in latest push.

amnonh:
> I'm interested in this, but the details aren't exactly clear. Some folks aren't going to switch to protobuf on the prom side, but can we offer both endpoints (one for text, one for prom) in parallel?

Yes, we support both endpoints in parallel, and when enabled, the result format will be based on the Accept header.

> On the application side there's nothing I need to do with the metrics themselves if we are already using `HISTOGRAM` type metrics, right? Those are compatible with native histogram once pb is enabled, right?

Not exactly, native histograms are exponential histograms and have specific limitations that are stronger than regular histograms. 

By the Prometheus definitions, a native histogram is defined by a schema and the number of buckets. 
Seastar Prometheus implementation only reports positive buckets and only supports integer histograms.

In Seastar implementation, we support the situation that a native histogram does not start from 0, so the native_histogram_info includes the schema and the id of the first bucket (in native histogram representation).

If a ```metrics:histogram``` has the (optional) ```native_histogram``` member, it means it can be represented as a native histogram on the wire; it also means that the histograms bucket follows the given schema, and there are no further validations.

To enable protobuf on the Seastar side, the parameter allow_protobuf  (which is part of the Prometheus config) should be set to true.

travisdowns:
> If a `metrics:histogram` has the (optional) `native_histogram` member, it means it can be represented as a native histogram on the wire; it also means that the histograms bucket follows the given schema, and there are no further validations.

Thanks. This is the part I was missing. So I can see that the existing `metrics_perf` bench is in fact already testing the native histogram path as it does support that member (it is using `approximate_exponential_histogram`), which explains the very much better performance and size of histograms compared to the text output [discussed above](https://github.com/scylladb/seastar/pull/3068#issuecomment-3444712502).

src/core/prometheus.cc
avikivity: Wouldn't std::vector::insert_range() work?

src/core/prometheus.cc
xemul: No, not a blocker at all. But a code comment describing this reasoning in few words would be helpful.

src/core/prometheus.cc
xemul: Please

src/core/prometheus.cc
xemul: The insert_range() is C++23 feature, but seastar still needs to support C++20.
From my side it's OK to have the above justification as code-comment and a TODO item to re-visit this class in potential favor of std::vector (or maybe something else) eventually

travisdowns: > Wouldn't std::vector::insert_range() work?

This is really just syntactic sugar [1] for the existing `vector::insert` available since forever. I also tried that, but it wasn't in the copied snippet above. It's _even worse!_ (despite that in theory it avoids the uninit problems)

Godbolt: https://godbolt.org/z/dE7Kjvcvv

Look at that, 400 lines of junk (how many cache lines is that, 25-ish?) for `insert()`ing `"a"`. Of course it doesn't execute all of that, but it executes something like 50 instructions worth regardless, which should just be:
```
mov rax, [vector end]
mov rbx, [vector capacity end]
cmp rbx, rax
je  slow_path # need to grow
mov [rax], "a"
add rbx, 1
mov [vector_end], rbx
ret
```
That's pretty much exactly what you get from the code in this PR (slow path not shown but we ~never take it).

You can also see in this link that `insert_range` is compiled in exactly the same way as `insert`, so no C++23 doesn't save us.

src/core/prometheus.cc
travisdowns: Removed in latest push.

src/core/prometheus.cc
travisdowns: Comment added in latest push.

avikivity:
@amnonh please indicate your opinion of the series

amnonh:
@avikivity I'm in favour, just too many people commenting on one pr is not too good

travisdowns:
Thanks all for the reviews.

PR #3071, opened 2025-10-27
metrics: Add default ~config()

xemul:
After 7a08db98b0 (prometheus: remove hostname and metric_help config) some clang versions (20.1.8 in particular) started to generate warning in generated destructor about some members being deprecated. This patch puts the default destructor explicitly into "ignore deprecation warning" section.


xemul:
Cc @travisdowns

travisdowns:
Thanks for the follow-up, Pavel!

PR #3074, opened 2025-10-28
http: Document the way "unread body bytes" accounting works

xemul:
The PRs that implemented this tracking were pretty confusing, this patch aims to improve it by documenting how the bytes read from body are accounted and how this accounting is used.

refs: #2762
refs: #3063


xemul:
@nyh , please review

PR #3076, opened 2025-10-30
smp: Fix exception safety of invoke_on_... internal copying

xemul:
Both invoke_on_all and invoke_on_others copy the function passed to it to be then submit_to()-ed on other shards. And only the invoke_on_others requires that the function is nothrow-copy-constructible. Marking the invoke_on_all() to requires nothros-copy-constructible function is hardly possible, there's lots of code out there that already passes functions with captured strings that can throw when being copied. So this change checks the if function is nothrow copyable and just copies that, otherwise it captures the potential exception and returns it as a future. The invoke_on_others is patched to do the same, for symmetry, and its requirement to have nothrow copyable function is relaxed.

### DIFF
--- a/include/seastar/core/smp.hh
+++ b/include/seastar/core/smp.hh
@@ -403,6 +403,23 @@ public:
     static std::ranges::range auto all_cpus() noexcept {
         return std::views::iota(0u, count);
     }
+private:
+    template <typename Func>
+    requires std::is_nothrow_copy_constructible_v<Func>
+    static futurize_t<std::invoke_result_t<Func>> copy_and_submit_to(unsigned t, smp_submit_to_options options, const Func& func) noexcept {
+        return submit_to(t, options, Func(func));
+    }
+
+    template <typename Func>
+    requires (!std::is_nothrow_copy_constructible_v<Func>)
+    static futurize_t<std::invoke_result_t<Func>> copy_and_submit_to(unsigned t, smp_submit_to_options options, const Func& func) noexcept {
+        try {
+            return submit_to(t, options, Func(func));
+        } catch (...) {
+            return current_exception_as_future();
+        }
+    }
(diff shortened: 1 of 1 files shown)


xemul:

avikivity:
@gleb-cloudius please review, I'm afraid of this code

include/seastar/core/smp.hh
```diff
@@ -403,6 +403,23 @@ public:
...
+    template <typename Func>
+    requires std::is_nothrow_copy_constructible_v<Func>
+    static futurize_t<std::invoke_result_t<Func>> copy_and_submit_to(unsigned t, smp_submit_to_options options, const Func& func) noexcept {
+        return submit_to(t, options, Func(func));
+    }
+
+    template <typename Func>
+    requires (!std::is_nothrow_copy_constructible_v<Func>)
+    static futurize_t<std::invoke_result_t<Func>> copy_and_submit_to(unsigned t, smp_submit_to_options options, const Func& func) noexcept {
+        try {
```
gleb-cloudius, line 416: Do you want to avoid always doing try/catch to make executable smaller?

xemul: Yes

gleb-cloudius: Do you think the compiler is not smart enough to drop try/catch code if the code inside cannot throw?

xemul: I think it can, and I checked that in simple cases it really did, but I'd prefer to be explicit here.
If you think it's an overkill, I'm OK to shorten the code

gleb-cloudius: I approved. Just wondered how smart compilers can be.

xemul:
@avikivity , please consider merging

PR #3077, opened 2025-10-30
net: tighten port load balancing port access

avikivity:
We fetch the port from the sockaddr structure as if it's always sockaddr_in, where it could also be sockaddr_in6.

This doesn't cause any problems, since in both sockaddr_in and sockaddr_in6 the port is in the same offset. However, it's not nice to access using the wrong type and to rely on such accidents.

"Fix" by dispatching based on the actual family.

A unit test is added to make sure we don't regress. It passes both before and after the patch.

### DIFF
--- a/tests/unit/socket_test.cc
+++ b/tests/unit/socket_test.cc
@@ -24,9 +24,13 @@
 #include <seastar/core/app-template.hh>
 #include <seastar/core/print.hh>
 #include <seastar/core/memory.hh>
+#include <seastar/core/byteorder.hh>
+#include <seastar/core/alien.hh>
+#include <seastar/core/smp.hh>
 #include <seastar/util/assert.hh>
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/later.hh>
+#include <seastar/util/defer.hh>
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
 #include <seastar/core/abort_source.hh>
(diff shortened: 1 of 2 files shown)


avikivity:
v2: use unsigned consistently for shard numbers so gcc doesn't complain

avikivity:
v2:
  - drop commented out code
  - explain bad_binds <= 20

tests/unit/socket_test.cc
```diff
@@ -337,3 +342,130 @@ SEASTAR_THREAD_TEST_CASE(socket_bufsize) {
...
+        alien::submit_to(alien, 0, [&] { client_done.set_value(); return make_ready_future<>(); });
+        return r;
+    });
+    client_done.get_future().get();
+    server.stop().get();
+    auto results = client.get();
+    BOOST_REQUIRE_EQUAL(results.attempts, 100);
+    // We can have bad binds due to other connection using the ports.
+    // 20 should be plenty of margin.
+    BOOST_REQUIRE_LE(results.bad_socket, 20);
```
xemul: The comment says "we can have bad binds", while the `bad_socket` counter counts the number of failures to create socket itself

avikivity:
v3: correct anti-flakiness bad_bind check to actually check binds

tests/unit/socket_test.cc
avikivity: Right.

avikivity:
@xemul please review

PR #3078, opened 2025-10-31
Generalize memory data sink in tests

xemul:
There's a memory_data_sink thing in http test that accumulates the put buffers into a stringstream. At the end, the test checks that the stream contains expected value. The very same checking pattern is found in few more tests, but instead of using this sink they collect buffers in net::packet with the help of vector_data_sink and then toss the packet to perform the necessary check. Patching those tests to re-use the stringstream sink from http test makes those tests a lot simpler.

Also there's a test case that checks how output_stream splits the incoming buffers. It also uses net::packet. And the stringstream sink doesn't help making it simpler, however, implementing the dedicated checking sink allows to get rig of net::packet usage from this test.

### DIFF
--- a/tests/unit/output_stream_test.cc
+++ b/tests/unit/output_stream_test.cc
@@ -21,11 +21,10 @@
 
 #include <seastar/core/app-template.hh>
 #include <seastar/core/shared_ptr.hh>
-#include <seastar/core/vector-data-sink.hh>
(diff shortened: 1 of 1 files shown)


xemul:
upd:
- keep shared ptr with stream to survive until close() future resulves :facepalm:

PR #3080, opened 2025-11-05
memcache: Implement own data sink for responses

xemul:
Currently memcache connection class collects its response data into a vector of net::packet-s with the help of vector_data_sink. This patch implements its own data sink that collects temporary_buffer-s into a vector of vectors.

This is pre-requisite patch for removing net::packet from UDP posix channels implemntation and to deprecate the vector_data_sink itself.

### DIFF
--- a/apps/memcached/memcache.cc
+++ b/apps/memcached/memcache.cc
@@ -36,7 +36,6 @@
 #include <seastar/core/memory.hh>
 #include <seastar/core/units.hh>
 #include <seastar/core/sharded.hh>
-#include <seastar/core/vector-data-sink.hh>
 #include <seastar/core/bitops.hh>
 #include <seastar/core/slab.hh>
 #include <seastar/core/align.hh>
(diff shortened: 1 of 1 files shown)


xemul:
Disable ccache for DPDK build. CI fails on it constantly when run on PR, but [branch](https://github.com/xemul/seastar/actions/runs/19127883299/job/54663593500) test is OK

xemul:
Enable ccache back. It [should](https://github.com/scylladb/seastar/issues/3083) fail :crossed_fingers: , but there's #3084 to fix it

xemul:

apps/memcached/memcache.cc
```diff
@@ -1250,9 +1249,29 @@ class udp_server {
...
+                stable_bufs.emplace_back(sizeof(header));
+                stable_bufs.insert(stable_bufs.end(), std::make_move_iterator(bufs.begin()), std::make_move_iterator(bufs.end()));
+                _con._out_bufs.emplace_back(std::move(stable_bufs));
+                return make_ready_future<>();
+            }
+
+            virtual future<> close() override {
+                return make_ready_future<>();
+            }
+        };
```
avikivity, line 1273: Will this compile with older API level?

xemul: No, but it's apps/. Every time we change API level, all apps/, tests/ and demos/ are updated accordingly without supporting previous API levels

PR #3081, opened 2025-11-05
demos: Simplify udp_zero_copy_demo's way of preparing the packet

xemul:
There are two modes in it, actually -- zero-copy and not. In each, the response packet is prepared using pre-allocated array of chars. A portion of bytes is appended three times to the response data, either with the help of scattered_message::append_static (zero-copy) or with the help of output_stream + vector_data_sink (copying mode). In either case, it's much simpler to just prepare three temp buffers in a vector and then send them as net::packet into datagram channel.

This is pre-requisite patch for removing net::packet from UDP posix channels implemntation and to deprecate the vector_data_sink itself.

### DIFF
--- a/demos/udp_zero_copy_demo.cc
+++ b/demos/udp_zero_copy_demo.cc
@@ -21,8 +21,6 @@
 
 #include <seastar/core/seastar.hh>
 #include <seastar/core/app-template.hh>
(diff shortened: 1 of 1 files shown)


xemul:

Issue #3083, opened 2025-11-06
CI DPDK job fails with ccache


xemul:
PRs constantly fail in DPDK CI job -- all tests abort early with "Illegal Instruction" message in logs. E.g. see #3082 that changes literally nothing, but [its CI fails](https://github.com/scylladb/seastar/actions/runs/19128838741/job/54664504895?pr=3082)
Disabling ccache for DPDK job helps. E.g. see #3080 and [its CI run](https://github.com/scylladb/seastar/actions/runs/19128763870/job/54664272897?pr=3080)

xemul:
https://github.com/scylladb/seastar/pull/3082/commits/a5e295b7c0cc234c970475031992cae78ce5ba8f is the "failing" empty commit. In #3082 I revert 81f2dc9dd976b0019ff84274b8b7fb7507c3e4e7 because it's roughly after this commit where it started (can be just an unlucky coincidence)

xemul:
Revert helped 😳 

<img width="875" height="511" alt="Image" src="https://github.com/user-attachments/assets/3ec4fe30-c5b1-43ae-9ac9-01a31579993e" />

xemul:
<img width="868" height="167" alt="Image" src="https://github.com/user-attachments/assets/77ef95cd-9b0a-437d-ad3e-4792fb1e71f2" />

xemul:
Cc @tchaikov . Kefu, maybe you have some better ideas what the problem can be in, but meanwhile I suggest to disable ccache for dpdk builds (#3084)

PR #3085, opened 2025-11-06
Stop using packets in tls data_sink and session

xemul:
When span of buffers is put into tls sink, it constructs a packet from the span and passes it into session::put(). The session::put(net::packet) in turn iterates over packet's fragments sending them one-by-one. Also there's optimization for "short" packet -- it gets linearized into a single fragment.

This PR makes session::put() accept span of buffers argument and iterate over them. There's small optimization for cases when the span is single-element or needs to be linearized.

### DIFF
--- a/src/net/tls.cc
+++ b/src/net/tls.cc
@@ -31,6 +31,7 @@ module;
 #include <chrono>
 #include <span>
 #include <unordered_set>
+#include <numeric>
 
 #include <seastar/util/assert.hh>
 
@@ -1513,13 +1514,22 @@ public:
     }
 
 private:
-    typedef net::fragment* frag_iter;
+    typedef std::vector<temporary_buffer<char>>::iterator frag_iter;
 
     future<> do_put(frag_iter i, frag_iter e) {
         SEASTAR_ASSERT(_output_pending.available());
-        return do_for_each(i, e, [this](net::fragment& f) {
-            return do_put_one(f.base, f.size);
+        return do_for_each(i, e, [this](temporary_buffer<char>& b) {
+            return do_put_one(b.get(), b.size());
         });
+    }
+
+    future<> do_put_one(temporary_buffer<char> buf) {
+        auto ptr = buf.get();
+        auto size = buf.size();
+        return with_semaphore(_out_sem, 1, [this, ptr, size] {
+            SEASTAR_ASSERT(_output_pending.available());
+            return do_put_one(ptr, size);
+        }).finally([b = std::move(buf)] {});
     }
 
     future<> do_put_one(const char* ptr, size_t size) {
@@ -1559,7 +1569,7 @@ private:
     }
 
 public:
-    future<> put(net::packet p) {
+    future<> put(std::span<temporary_buffer<char>> bufs) {
         if (_error) {
             return make_exception_future<>(_error);
         }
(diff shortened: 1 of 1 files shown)


xemul:
DPDK failure: #3083

src/net/tls.cc
```diff
@@ -1559,32 +1569,50 @@ class session : public enable_lw_shared_from_this<session> {
...
+        size_t size = std::accumulate(bufs.begin(), bufs.end(), size_t(0), [] (size_t s, const auto& b) { return s + b.size(); });
+        if (size <= gnutls_record_get_max_size(*this)) {
+            temporary_buffer<char> linear(size);
+            char* pos = linear.get_write();
+            for (auto& buf : bufs) {
+                std::copy_n(buf.get(), buf.size(), pos);
+                pos += buf.size();
+            }
+            return do_put_one(std::move(linear));
         }
```
avikivity, line 1609: @elcallio could we use gnutls_record_cork / gnutls_record_uncork instead here?

Not affecting the current series.

src/net/tls.cc
```diff
@@ -1559,32 +1569,50 @@ class session : public enable_lw_shared_from_this<session> {
...
 
-        auto i = p.fragments().begin();
-        auto e = p.fragments().end();
+        std::vector<temporary_buffer<char>> p;
+        p.reserve(bufs.size());
+        p.insert(p.end(), std::make_move_iterator(bufs.begin()), std::make_move_iterator(bufs.end()));
+        auto i = p.begin();
+        auto e = p.end();
```
avikivity, line 1615: Could write: `auto p = std::vector(bufs | std::views::as_rvalue)`, but not worth an iteration.

avikivity: Also, C++23 only.

src/net/tls.cc
elcallio: Probably. It may or may not be better, since gnutls is gonna copy data into internal buffers anyway.

src/net/tls.cc
```diff
@@ -1513,15 +1514,24 @@ class session : public enable_lw_shared_from_this<session> {
...
-            return do_put_one(f.base, f.size);
+        return do_for_each(i, e, [this](temporary_buffer<char>& b) {
+            return do_put_one(b.get(), b.size());
         });
     }
 
+    future<> do_put_one(temporary_buffer<char> buf) {
+        auto ptr = buf.get();
+        auto size = buf.size();
+        return with_semaphore(_out_sem, 1, [this, ptr, size] {
```
elcallio, line 1529: Nit: for symmetry, should not either this grabbing of the semaphore be in caller (put, below), or the semaphore in put below (in the iter case) be moved to the above iter-based `do_put`?

xemul: I disliked it too :( This method is symmetrical with `do_put(frag_iter i, frag_iter e)`, not with `do_put_one(const char *ptr, size_t)`

xemul: I called it `do_put(temporary_buffer)` first, but `std::bind(&session::do_put, this, i, e)` failed to resolve correct do_put() overload

elcallio: Not complaining about naming. I am complaining that this method _contains_ the `with_semaphore(_out_sem` call, whereas the other `do_put_one` has the semaphore grab done in `put` at line 1616. Seems to me the semaphore grab should be also in put on line 1608.

xemul: I got your point. Then it should have semphore grab in line 1589 too, and I decided to generalize it rather than keep symmetrical.

From symmetry considerations, it should look like line 1616 does -- `return with_semaphore(_out_sem, 1, std::bind(&session::do_put, ...`. In that case std::bind fails on do_put overload. If  `return with_semaphore(... std::bind(&session::do_put_one, ...` then fails on do_put_one overload :( So naming again. I'll suggest a follow up later

xemul: #3095

PR #3086, opened 2025-11-07
Generalize packet_data_source into memory_data_source

xemul:
The packet_data_source provides data_source_impl API against net::packet. Most of the code that uses it creates packet out of strings or temporary buffers. This PR generalizes packet_data_source into memory_data_source that wraps container with buffers into data source API and deprecates the packet source.

### DIFF
--- a/apps/memcached/tests/test_ascii_parser.cc
+++ b/apps/memcached/tests/test_ascii_parser.cc
@@ -23,7 +23,7 @@
 #include <limits>
 #include <seastar/testing/test_case.hh>
 #include <seastar/core/shared_ptr.hh>
-#include <seastar/net/packet-data-source.hh>
+#include <seastar/util/memory-data-source.hh>
 #include "ascii.hh"
 #include <seastar/core/loop.hh>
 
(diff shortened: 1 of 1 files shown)


xemul:
upd: rebased (no conflicts)

xemul:
upd: fixed as_input_stream deprecation message

xemul:

include/seastar/util/memory-data-source.hh
```diff
@@ -0,0 +1,67 @@
...
+#pragma once
+
+#include <seastar/core/iostream.hh>
+
+namespace seastar {
+namespace util {
+
+template <typename Container>
+requires std::is_nothrow_move_constructible_v<Container>
+class memory_data_source final : public data_source_impl {
```
avikivity: Suggest to call this basic_memory_data_source and define basic_memory_data_source\<std::vector\<temporary_buffer\<char>>> as memory_data_source.

Need more constraints.

Do we want to export it? I guess it's generally useful.

Option: make the container parameter a template template parameter, so you can write memory_data_source\<std::vector>.

include/seastar/util/memory-data-source.hh
```diff
@@ -0,0 +1,67 @@
...
+    Container _bufs;
+    Container::iterator _cur;
+
+public:
+    explicit memory_data_source(Container&& b) noexcept
```
avikivity: It works with std::span or views, maybe Container is too contraining.

In that case, a template template parameter wouldn't work.

xemul:
upd:
- renamed memory_data_source template into basic_memory_data_source
- detached temporary_buffer_data_source not to be specialization of (basic_)memory_data_source
- made Container template parameter to be templatized container class name
- added alias memory_data_source = basic_memory_data_source<std::vector>
- exported the new data sources and as_input_stream helpers

xemul:
upd:
- dropped accidental inclusion of #3109

xemul:
Clang-18 cannot compile this template template parameter :(

avikivity:
> Clang-18 cannot compile this template template parameter :(

We can delay this a little until we update the toolchain to support clang 20/21.

Only the modules stuff is missing I think.

xemul:
GH latest Linux image is ubuntu:24.04, but it has clang-20. I'm testing it (https://github.com/xemul/seastar/tree/refs/heads/br-gh-use-clang-20), it works, but has problems compiling existing modules (https://github.com/xemul/seastar/actions/runs/19496007131). Maybe #3101 will help (https://github.com/xemul/seastar/actions/runs/19498051361)

xemul:
upd: :tada: #3101 + [clang-20](https://github.com/xemul/seastar/commits/refs/heads/br-gh-use-clang-20/)

xemul:
upd
- rebased to pick up compiler update (#3114) and new modules support (#3101)

PR #3089, opened 2025-11-10
rpc: harden sink_impl::close()

### LINKED ISSUE #3088: Assertion failed in ~batched_queue when destroying rpc::sink
It appears like sink_impl<Serializer, Out...>::close() is insufficient for asserting `_process_fut.available()` in ~batched_queue.

We should also break the connection semaphore to prevent any new messages from being sent (after the semaphore units are returned on the close() path).
And we better await send and deleted queue stop() to ensure that the queues are drained before they are destroyed.

See also https://github.com/scylladb/scylladb/issues/26926

* No backport required

bhalevy:
Following up on 06c671c,
Break the rpc sink semaphore in sink_impl::close to
prevent any new message from being queued after the sink is closed,
and also stop the send and delete queue to ensure they
can be destroyed safely.

Fixes #3088

Signed-off-by: Benny Halevy <bhalevy@scylladb.com>

### DIFF
--- a/include/seastar/rpc/rpc_impl.hh
+++ b/include/seastar/rpc/rpc_impl.hh
@@ -900,25 +900,25 @@ future<> sink_impl<Serializer, Out...>::close() noexcept {
         // break the semaphore to prevent any new messages to be sent
         this->_sem.broken(stream_closed());
         return _send_queue.stop().finally([this] {
-        return smp::submit_to(this->_con->get_owner_shard(), [this] {
-          return _delete_queue.stop().finally([this] {
-            connection* con = this->_con->get();
-            if (con->sink_closed()) { // double close, should not happen!
-                return make_exception_future(stream_closed());
-            }
-            future<> f = make_ready_future<>();
-            if (!con->error() && !this->_ex) {
-                snd_buf data = marshall(con->template serializer<Serializer>(), 4);
-                static_assert(snd_buf::chunk_size >= 4, "send buffer chunk size is too small");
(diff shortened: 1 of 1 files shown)


include/seastar/rpc/rpc_impl.hh
```diff
@@ -897,9 +897,11 @@ future<> sink_impl<Serializer, Out...>::flush() noexcept {
 template<typename Serializer, typename... Out>
 future<> sink_impl<Serializer, Out...>::close() noexcept {
     return with_semaphore(this->_sem, max_stream_buffers_memory, [this] {
-        // the send and delete queues should be drained already
-        // since we acquired all the semaphore units, so no need to stop them.
+        // break the semaphore to prevent any new messages to be sent
+        this->_sem.broken(stream_closed());
+        return _send_queue.stop().finally([this] {
         return smp::submit_to(this->_con->get_owner_shard(), [this] {
+          return _delete_queue.stop().finally([this] {
```
xemul: Shouldn't it happen on the original shard as well? Both, send_queue and delete_queue are created and initialized on the same shard, and the process_fut is resolved on the same shard as well.

bhalevy: Nope, because the buffers are enqueued onto _delete_queue by the connection shard, and it will create the process_loop there (the loop, in turn will submit_to back to the sink shard, but it needs to be awaited on the shard where the process_loop task was created on)

PR #3090, opened 2025-11-10
socket: Return unspecified socket_address for unconnected socket

### LINKED ISSUE #2379: Posix remote_address() may throw, but it is marked noexcept
I ran into the following error:

```
terminate called after throwing an instance of 'std::system_error'
  what():  getpeername: Transport endpoint is not connected
Aborting on shard 1.
```

Partial backtrace:

```
std::terminate() at ??:?
__clang_call_terminate at :?
seastar::net::posix_connected_socket_impl::remote_address() const at ??:?
seastar::tls::tls_connected_socket_impl::remote_address() const at ??:?
seastar::connected_socket::remote_address() const at ??:?
```

I believe the cause is `posix_connected_socket_impl::remote_address()` is marked noexcept, but it calls `posix_connected_socket_operations::remote_address()`, which is not noexcept, and in fact the latter calls `file_desc::get_remote_address()` which explicitly throws.

See:

https://github.com/scylladb/seastar/blob/master/src/net/posix-stack.cc#L288
https://github.com/scylladb/seastar/blob/master/src/net/posix-stack.cc#L110
https://github.com/scylladb/seastar/blob/master/include/seastar/core/posix.hh#L298

xemul:
When calling connected_socket::remote_address() it may happen that the kernel-level socket is really unconnected, e.g. because it was connected before and then was forcibly disconnected.

For posix_connected_socket_impl this leads to app termination, because its remote_address() method is noexcept, it calls file_desc wrapper of get_peer_address() libc function which, in turn, returns ENOTCON error and the file_desc throws system_error with it.

Fix by catching exceptions from file_desc and returning default-initialized socket_address instead.

To test it there must appear foribly disconnected socket. For TCP it can be done by closing a socket with unread data in it. The attached test does it, it crashes before the fix and passes after it.

fixes #2379

### DIFF
--- a/src/net/posix-stack.cc
+++ b/src/net/posix-stack.cc
@@ -111,7 +111,14 @@ public:
         return _fd.get_address();
     }
     virtual socket_address remote_address(file_desc& _fd) const {
-        return _fd.get_remote_address();
+        try {
+            return _fd.get_remote_address();
+        } catch (const std::system_error& e) {
+            if (e.code().value() != ENOTCONN) {
+                throw;
+            }
+            return socket_address();
+        }
     }
 };
 
--- a/tests/unit/socket_test.cc
+++ b/tests/unit/socket_test.cc
@@ -469,3 +469,35 @@ SEASTAR_THREAD_TEST_CASE(load_balancing_algorithm_port_ipv4_test) {
 SEASTAR_THREAD_TEST_CASE(load_balancing_algorithm_port_ipv6_test) {
     test_load_balancing_algorithm_port(ipv6_addr("::1", 11001));
 }
+
+SEASTAR_THREAD_TEST_CASE(inet_local_remote_address_sanity) {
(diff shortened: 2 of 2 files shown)


xemul:

src/net/posix-stack.cc
```diff
@@ -289,7 +289,11 @@ class posix_connected_socket_impl final : public connected_socket_impl {
         return _ops->local_address(_fd.get_file_desc());
     }
     socket_address remote_address() const noexcept override {
-        return _ops->remote_address(_fd.get_file_desc());
+        try {
+            return _ops->remote_address(_fd.get_file_desc());
+        } catch (...) {
```
avikivity: It's better to catch the specific exception and errno, so that future bugs (say that causes remote_address to always throw) are caught.

xemul:
upd:
- catch system_error and check that errno is ENOTCON

PR #3092, opened 2025-11-11
timer: Enable highres timer based on next timeout value

### LINKED ISSUE #3013: No timer callback after periodic timer cancellation
```cpp
#include <seastar/core/timer.hh>
#include <seastar/core/thread.hh>
#include <seastar/core/sleep.hh>
#include <chrono>
#include <iostream>

using namespace seastar;

struct timer_hang_test {
    timer<steady_clock_type> t1;
    timer<steady_clock_type> t2;
    promise<> pr1;
    promise<> pr2;

    future<> run() {
        t1.set_callback([this](){
            std::cerr << "has t1 callback"<< std::endl;
            if (t1.cancel()) {
                std::cerr << "t1 has been cancelled"<< std::endl;
            }
            pr1.set_value();
        });
        t2.set_callback([this](){
            std::cerr << "no t2 callback"<< std::endl;
            pr2.set_value();
        });

        (void)pr1.get_future().then([this]{
            t2.arm(2ms);
            std::cerr << "t2 has been armed"<< std::endl;
        });

        t1.arm_periodic(1ms);

        // some work here
        std::this_thread::sleep_for(1ms);
        std::cerr << "some work is done"<< std::endl;

        // this future never resolves
        return pr2.get_future();
    }
};

int main(int ac, char** av) {
    app_template app;
    timer_hang_test t1;
    return app.run_deprecated(ac, av, [&t1] {
        fmt::print("=== Start Timer hang test\n");
        return t1.run().then([] {
            fmt::print("Done\n");
            engine().exit(0);
        });
    });
}
```

The test hangs untill Ctrl+C is pressed. The text fails 100%.  Full test output:
```txt
WARN  2025-10-02 12:29:11,090 seastar - Your system does not have enough AIO capacity for optimal network performance; reducing `max-networking-io-control-blocks'.
WARN  2025-10-02 12:29:11,090 seastar - Resultant AIO control block usage:
WARN  2025-10-02 12:29:11,090 seastar - 
WARN  2025-10-02 12:29:11,090 seastar - purpose  per cpu  all 8 cpus
WARN  2025-10-02 12:29:11,090 seastar - -------  -------  ----------
WARN  2025-10-02 12:29:11,090 seastar - reserve                    0
WARN  2025-10-02 12:29:11,090 seastar - storage     1024        8192
WARN  2025-10-02 12:29:11,090 seastar - preempt        2          16
WARN  2025-10-02 12:29:11,090 seastar - network     7166       57328
WARN  2025-10-02 12:29:11,090 seastar - -------  -------  ----------
WARN  2025-10-02 12:29:11,090 seastar - total       8192       65536
WARN  2025-10-02 12:29:11,090 seastar - 
WARN  2025-10-02 12:29:11,090 seastar - For optimal network performance, set /proc/sys/fs/aio-max-nr to at least 88208.
INFO  2025-10-02 12:29:11,091 seastar - Reactor backend: linux-aio
INFO  2025-10-02 12:29:11,092 seastar - Perf-based stall detector creation failed (EACCESS), try setting /proc/sys/kernel/perf_event_paranoid to 1 or less to enable kernel backtraces: falling back to posix timer.
INFO  2025-10-02 12:29:11,102 [shard 0:main] seastar - IO queue was unable to find a suitable maximum request length, the search was cut-off early at: 16MB
INFO  2025-10-02 12:29:11,102 [shard 0:main] seastar - IO queue was unable to find a suitable maximum request length, the search was cut-off early at: 16MB
=== Start Timer hang test
some work is done
has t1 callback
t1 has been cancelled
t2 has been armed
```

xemul:
When highres timers are completed, the post-completion callback checks if there are any timers queued to decide whether or not to set up the timerfd notification. The check is wrong, timer list can be empty at that point if a timer is (was) canceled while being run. At the same time, since "next timeout" value is set, timer-set thinks that the timerfd is active and may decide not to re-activate it when another timer appears.

The fix is to engage timerfd based on "next timeout" value.

Test that reproduced the issue is included, it fails before the fix.

fixes #3013

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -3207,8 +3207,8 @@ void reactor::sched_entity::wakeup() {
 
 void reactor::service_highres_timer() noexcept {
     _timers.complete(_expired_timers, [this] () noexcept {
-        if (!_timers.empty()) {
-            enable_timer(_timers.get_next_timeout());
+        if (auto next = _timers.get_next_timeout(); next != decltype(next)::time_point::max()) {
+            enable_timer(next);
         }
     });
 }
--- a/tests/unit/timer_test.cc
+++ b/tests/unit/timer_test.cc
@@ -27,6 +27,7 @@
 #include <seastar/core/print.hh>
 #include <seastar/core/thread.hh>
 #include <seastar/core/sleep.hh>
+#include <seastar/util/later.hh>
 #include <chrono>
 #include <iostream>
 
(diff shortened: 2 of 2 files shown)


tgrabiec:
Isn't `do_expire_lowres_timers` also susceptible to this? It sets `_lowres_next_timeout = lowres_clock::time_point::max();` diregarding `_lowres_timers.get_next_timeout()`,

xemul:
In theory yes, but loweres timers always have some timer armed (e.g. `timer<lowres_clock> load_timer` one). As a result, the situation when do_expire_lowres_timer()'s callback notices that the timer set is empty never happens.

tgrabiec:
I see. It's not good to depend on that though.

xemul:
Agree, but I want to come up with some reliable testing of it though

PR #3093, opened 2025-11-11
github: disable AVX instruction set when building dpdk

### LINKED ISSUE #3083: CI DPDK job fails with ccache

tchaikov:
by default, we build dpdk with "meson setup
-Dcpu_instruction_set=native". in which, "cpu_instruction_set=native" option tells the building system to configures "-march", "-mcpu" and "-mtune" to "native". this might instruct the compiler to emit target code using advanced instruction set like AVX-512, while the the CPU might not be able to execute it, because the hypervisor might not expose all CPU features in a virtualization environment.

in this change, instead of using "generic" which is mapped to "corei7", we use a conservative but performant alternative -- "corei7-avx" where SSE4.2 and AVX is enabled, but AVX2 is disabled.

Fixes #3083

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -35,7 +35,8 @@ jobs:
       standard: 23
       mode: release
       enables: --enable-dpdk
-      options: --cook dpdk
+      options: --cook dpdk --dpdk-machine corei7-avx
+
   build_with_cxx_modules:
     name: "Test with C++20 modules enabled"
     uses: ./.github/workflows/test.yaml


mykaul:
How sad that we need to resort to a 2008 architecture :( 
You'd think by now 2011 architecture is common enough.

.github/workflows/tests.yaml
```diff
@@ -35,7 +35,8 @@ jobs:
       standard: 23
       mode: release
       enables: --enable-dpdk
-      options: --cook dpdk
+      options: --cook dpdk --dpdk-machine corei7-avx
```
xemul, line 38: But patch comments says it should be westmere

tchaikov: sorry, i changed it to corei7-avx after reading @mykaul 's comment. will update the commit message.

tchaikov:
v2:

- corrected the commit message and cover letter to refect the change.

avikivity:
What hypervisor exposes AVX-512 but then doesn't actually implement all of it?

PR #3094, opened 2025-11-11
file: recursive_remove_directory(): use a list instead of a deque

avikivity:
recursive_remove_directory() stores a list of pending files in a dequeue. This can lead to large allocations with large directories.

Switch to std::list. This is less memory and cpu efficient, but doesn't have large allocations.

Performance will be dominated by the OS anyway.

Ref https://github.com/scylladb/scylladb/issues/26925

### DIFF
--- a/src/util/file.cc
+++ b/src/util/file.cc
@@ -25,7 +25,7 @@ module;
 #endif
 
 #include <cstdint>
-#include <deque>
+#include <list>
 #include <optional>
 #include <filesystem>
 #include <iostream>
@@ -177,7 +177,7 @@ static future<> do_recursive_remove_directory(const fs::path path) noexcept {
         }
     };
 
-    return do_with(std::deque<work_entry>(), [path = std::move(path)] (auto& work_queue) mutable {
+    return do_with(std::list<work_entry>(), [path = std::move(path)] (auto& work_queue) mutable {
         work_queue.emplace_back(std::move(path), false);
         return do_until([&work_queue] { return work_queue.empty(); }, [&work_queue] () mutable {
             auto ent = work_queue.back();


nyh:
We have in Seastar **chunked_fifo** for exactly this purpose - avoiding the potentially large allocations of `std::deque()`, while not being as wasteful as `std::list`.

I agree that since the OS is involved, `std::list` is "good enough" and there's nothing wrong with using it here. But since we already have seastar::chunked_fifo, maybe we should just use it?

avikivity:
> We have in Seastar **chunked_fifo** for exactly this purpose - avoiding the potentially large allocations of `std::deque()`, while not being as wasteful as `std::list`.


recursive_reuse_directory uses the deque as a stack, not a queue. I don't know if it's really necessary, but better not to find out the hard way.

xemul:
Why accumulating all file names in memory? After a bunch of them being collected, they can be removed and directory can be rewinded to zero (or re-opened)

avikivity:
> Why accumulating all file names in memory? After a bunch of them being collected, they can be removed and directory can be rewinded to zero (or re-opened)

Why rewind the directory? Normal recursive descent can be used.

avikivity:
0034b3104fe978be2fe48a15d4c8c560aa57cad8 says "make it non recursive", but futures convert recursion to task allocation anyway.

xemul:
Because AFAIR if you remove files in the middle of it being readdir()-ed it may no necessarily "continue" listing

nyh:
> [0034b31](https://github.com/scylladb/seastar/commit/0034b3104fe978be2fe48a15d4c8c560aa57cad8) says "make it non recursive", but futures convert recursion to task allocation anyway.

So now I know why it needed to be a stack, not a queue (and can't be chunked_fifo) - you need to delete the last (deeper) files before you can delete the first (upper level) directories.

But still a better implementation would be to have a stack only of directory levels, not of the entire list of files. And like you said even that might not be need with futures, if there is no real recursion anyway, just a chain of tasks.

avikivity:
> Because AFAIR if you remove files in the middle of it being readdir()-ed it may no necessarily "continue" listing

Aha. I think it's not worth the effort to reimplement.

PR #3102, opened 2025-11-14
reactor: remove blocking of SIGILL

rockwotj:
It's not clear why we do this, while the application shard (smp::_tmain) has a bunch of signals unblocked on it. This is annoying in Redpanda moreso now because we have turned on libc++ hardening mode[1] in our debug build, which crashes with SIGILL, and currently we don't get signals for that.

References: https://github.com/scylladb/scylladb/issues/14163 (but doesn't fix SIGFPE)

[1]: https://libcxx.llvm.org/Hardening.html

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4166,6 +4166,11 @@ static void sigabrt_action(siginfo_t *info, ucontext_t* uc) noexcept {
     reraise_signal(SIGABRT);
 }
 
+static void sigill_action(siginfo_t *info, ucontext_t* uc) noexcept {
+    print_with_backtrace("Invalid instruction");
+    reraise_signal(SIGILL);
+}
+
 // We don't need to handle SIGSEGV when asan is enabled.
 #ifdef SEASTAR_ASAN_ENABLED
 template<>
@@ -4275,6 +4280,7 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
 
     install_oneshot_signal_handler<SIGSEGV, sigsegv_action>();
     install_oneshot_signal_handler<SIGABRT, sigabrt_action>();
+    install_oneshot_signal_handler<SIGILL, sigill_action>();
 
 #ifdef SEASTAR_HAVE_DPDK
     const auto* native_stack = dynamic_cast<const net::native_stack_options*>(reactor_opts.network_stack.get_selected_candidate_opts());
@@ -4604,7 +4610,7 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
             }
             sigset_t mask;
             sigfillset(&mask);
-            for (auto sig : { SIGSEGV }) {
+            for (auto sig : { SIGSEGV, SIGILL }) {
                 sigdelset(&mask, sig);
             }
             auto r = ::pthread_sigmask(SIG_BLOCK, &mask, NULL);


rockwotj:
> The SIGFPE is the same in this sense and thus also needs to be unlocked in ... non-master (?) reactor thread.
> 
> Also neither ILL (nor FPE) have installed handler, while SEGV and ABRT have it and print the backtrace. What's the expected behavior of catching ILL (and FPE)? Just die silently, prints backtrace and die or let 3rd party lib handle it?

Pushed a commit that adds a handler for SIGILL. SIGFPE I'm not planning on handling that here. While that fixes the Wasm bug I pointed too in Scylla, my actual motivation is to handle SIGILL for the libc++ hardening, which crashes with SIGILL when there is an issue. Users of seastar can always unblock and register handlers themselves. We're doing this today for our main applications, but instead of adding this to all our unit tests, it's easier to just do it in seastar directly (TBH everyone should probably turn on stdlib hardening at least in debug mode with sanitizers, it helped us catch a bunch of bugs around missing std::optional, many of which were real races).

avikivity:
This 5c2660badd123be73681cc8feb2c14e9ecd2e287 introduced SIGILL blocking. The commit should explain how it's compatible with the intent, or why the intent is wrong.

rockwotj:
> This [5c2660b](https://github.com/scylladb/seastar/commit/5c2660badd123be73681cc8feb2c14e9ecd2e287) introduced SIGILL blocking. The commit should explain how it's compatible with the intent, or why the intent is wrong.

Hrm, I'm confused that bit of code seems to **not** be blocking SIGILL explicitly? As `sigdelset(&sigs, SIGILL)` is being called.

avikivity:
> > This [5c2660b](https://github.com/scylladb/seastar/commit/5c2660badd123be73681cc8feb2c14e9ecd2e287) introduced SIGILL blocking. The commit should explain how it's compatible with the intent, or why the intent is wrong.
> 
> Hrm, I'm confused that bit of code seems to **not** be blocking SIGILL explicitly? As `sigdelset(&sigs, SIGILL)` is being called.

Was wrong here, sorry.

So what introduced blocking of SIGILL?

nyh:
> Partially Fixes: [scylladb/scylladb#14163](https://github.com/scylladb/scylladb/issues/14163)

When github sees this wording it just sees "Fixes ..." and will close the ScyllaDB issue automatically.
Please change the wording to be something different if you didn't intent to close the ScyllaDB issue.
E.g., something like:

Refs scylladb/scylladb#14163 (but doesn't fix it beause... or maybe - it might fix it but I don't know... or whatever.)

nyh:
Related issue:
https://github.com/scylladb/seastar/issues/261 - which complains that SIGINT and SIGTERM are also trapped automatically for no good reason (if the application didn't ask to handle them)

rockwotj:
> Please change the wording to be something different if you didn't intent to close the ScyllaDB issue.

Done, thanks!

> https://github.com/scylladb/seastar/issues/261 - which complains that SIGINT and SIGTERM are also trapped automatically for no good reason (if the application didn't ask to handle them)

There is this `auto_handle...` flag, should SIGILL be there too? It's not really clear to me what the experience here should be.

rockwotj:
> So what introduced blocking of SIGILL?

It's always been there as far I can tell:

https://github.com/scylladb/seastar/commit/0a93385a4d2794be3ee163765c26082c58fc2ddd

But I don't know of why all signals were blocked, later on SIGSEGV was unblocked:

https://github.com/scylladb/seastar/commit/1ef66deddfaec1b0654aedd62d484e6f8a186e2c

avikivity:
> > So what introduced blocking of SIGILL?
> 
> It's always been there as far I can tell:
> 
> [0a93385](https://github.com/scylladb/seastar/commit/0a93385a4d2794be3ee163765c26082c58fc2ddd)
> 
> But I don't know of why all signals were blocked, later on SIGSEGV was unblocked:
> 
> [1ef66de](https://github.com/scylladb/seastar/commit/1ef66deddfaec1b0654aedd62d484e6f8a186e2c)

Wait I'm confused. The patch I mentioned before as introduced SIGILL blocking was really unblocking them. Or was it just a no-op wrt SIGILL?

rockwotj:
> Wait I'm confused. The patch I mentioned before as introduced SIGILL blocking was really unblocking them. Or was it just a no-op wrt SIGILL?

The commit you mentioned does not touch SIGILL it only blocks signals that are **not** in that list. It does not unblock anything (so a noop for SIGILL). Presumably from the DPDK comment, there are a bunch of "non common" signals DPDK uses that we don't want seastar threads to ever handle. Since reactor threads already have a bunch of signals blocked, I think it's a noop for the blocking there anyways, but possibly other threads than the reactor threads have the codepath called that is changed in that patch.

rockwotj:
Okay sorry I read a little closer. There are two signal masks here:

The first your patch points too, is only applied on shard0, all other reactor shards have a different set of signals blocks (which I don't think was on purpose).

So shard0 has SIGILL not blocked, but other shards block all signals except SIGSEGV.

Maybe we just use the same set of signal masking for all threads here? (which would be deleting the additional signal blocking in the other reactor shards, as they should get the same signal mask as shard0)

avikivity:
> Okay sorry I read a little closer. There are two signal masks here:
> 
> The first your patch points too, is only applied on shard0, all other reactor shards have a different set of signals blocks (which I don't think was on purpose).
> 
> So shard0 has SIGILL not blocked, but other shards block all signals except SIGSEGV.
> 
> Maybe we just use the same set of signal masking for all threads here? (which would be deleting the additional signal blocking in the other reactor shards, as they should get the same signal mask as shard0)

Yes.

We have three kinds of threads: reactor, syscall, alien.

Some signals are thread-local (the synchronous ones, like SIGILL). Some are process-wide. The process-wide signals might want blocking in some threads.

Let's go over all the signals:

       SIGABRT      P1990      Core    Abort signal from abort(3)

local, allow

       SIGALRM      P1990      Term    Timer signal from alarm(2)

Not used, may block

       SIGBUS       P2001      Core    Bus error (bad memory access)

local, allow

       SIGCHLD      P1990      Ign     Child stopped or terminated

process management doesn't use signals IIRC, block

       SIGCLD         -        Ign     A synonym for SIGCHLD
       SIGCONT      P1990      Cont    Continue if stopped

block?

       SIGEMT         -        Term    Emulator trap

N/A

       SIGFPE       P1990      Core    Floating-point exception


local, allow

       SIGHUP       P1990      Term    Hangup detected on controlling terminal
                                       or death of controlling process

block by default, allow application to override (ScyllaDB does)

       SIGILL       P1990      Core    Illegal Instruction

local, allow

       SIGINFO        -                A synonym for SIGPWR

block, alien can unblock if it wants

       SIGINT       P1990      Term    Interrupt from keyboard

block by default, stop_signal facility unblocks

       SIGIO          -        Term    I/O now possible (4.2BSD)

block, app may unblock if it wants

       SIGIOT         -        Core    IOT trap. A synonym for SIGABRT
       SIGKILL      P1990      Term    Kill signal

Block!!!!!eleven111

       SIGLOST        -        Term    File lock lost (unused)
       SIGPIPE      P1990      Term    Broken pipe: write to pipe with no
                                       readers; see pipe(7)

Hmm, block on syscall threads?

       SIGPOLL      P2001      Term    Pollable event (Sys V);
                                       synonym for SIGIO
       SIGPROF      P2001      Term    Profiling timer expired
       SIGPWR         -        Term    Power failure (System V)
       SIGQUIT      P1990      Core    Quit from keyboard

block, app can override (ScyllaDB does)

       SIGSEGV      P1990      Core    Invalid memory reference

local, allow

       SIGSTKFLT      -        Term    Stack fault on coprocessor (unused)
       SIGSTOP      P1990      Stop    Stop process

Don't stop me now

       SIGTSTP      P1990      Stop    Stop typed at terminal
       SIGSYS       P2001      Core    Bad system call (SVr4);
                                       see also seccomp(2)
       SIGTERM      P1990      Term    Termination signal

Block?, app can override

       SIGTRAP      P2001      Core    Trace/breakpoint trap
       SIGTTIN      P1990      Stop    Terminal input for background process
       SIGTTOU      P1990      Stop    Terminal output for background process
       SIGUNUSED      -        Core    Synonymous with SIGSYS
       SIGURG       P2001      Ign     Urgent condition on socket (4.2BSD)
       SIGUSR1      P1990      Term    User-defined signal 1
       SIGUSR2      P1990      Term    User-defined signal 2
       SIGVTALRM    P2001      Term    Virtual alarm clock (4.2BSD)
       SIGXCPU      P2001      Core    CPU time limit exceeded (4.2BSD);
                                       see setrlimit(2)
       SIGXFSZ      P2001      Core    File size limit exceeded (4.2BSD);
                                       see setrlimit(2)
       SIGWINCH       -        Ign     Window resize signal (4.3BSD, Sun)


I guess, for these, block and app may override.

xemul:
Effectively, local signals are only those sent by kernel on CPU exception  -- SEGV, ILL and FPE. Keyboard events send signals from the kernel as well, but these are sent to a process. Same for terminal signals (like HUP). All other signals are safe to be unblocked and handled by shard-0 only. Though user can tkill(2) a specific thread with any signal, what can be the reason to handle it in seastar?

>     SIGCONT      P1990      Cont    Continue if stopped
> 
> block?

CONT is non-blockable

PR #3104, opened 2025-11-14
net: tcp: change unsent queue from packets to temporary_buffer:s

avikivity:
The unsent queue is currently a queue of packets. This is awkward because the packets queued don't correspond to packets to be sent; they have to be cut and pasted to fit the send window or maximum packet size, whichever is greater.

The recent change to data_sink to work in terms of temporary_buffer:s rather than packets provides an additional incentive to fix this.

The fix is simple: change the queue type to hold temporary buffers. The compatibility code to construct a packet out of the data_sink temporary_buffer:s is no longer needed.

The code to reshape the unsent packets to a packet to be sent is simplified: we append buffers to the packet until a complete buffer will overflow the maximum send size, then split the last buffer (if available and needed) to complete the packet.

Tested with httpd on a tap device, seems to work.

### DIFF
--- a/include/seastar/net/tcp.hh
+++ b/include/seastar/net/tcp.hh
@@ -28,6 +28,7 @@
 #include <deque>
 #include <chrono>
 #include <random>
+#include <span>
 #include <stdexcept>
 #include <system_error>
 #include <gnutls/crypto.h>
@@ -339,7 +340,7 @@ private:
             tcp_seq wl2;
             tcp_seq initial;
             std::deque<unacked_segment> data;
-            std::deque<packet> unsent;
+            std::deque<temporary_buffer<char>> unsent;
             uint32_t unsent_len = 0;
             bool closed = false;
             promise<> _window_opened;
@@ -436,7 +437,7 @@ private:
         void abort_reader() noexcept;
         future<> wait_for_all_data_acked();
         future<> wait_send_available();
-        future<> send(packet p);
+        future<> send(std::span<temporary_buffer<char>> data);
         void connect();
         packet read();
         void close() noexcept;
@@ -689,8 +690,8 @@ public:
         future<> connected() {
             return _tcb->connect_done();
         }
-        future<> send(packet p) {
-            return _tcb->send(std::move(p));
+        future<> send(std::span<temporary_buffer<char>> data) {
+            return _tcb->send(data);
         }
         future<> wait_for_data() {
             return _tcb->wait_for_data();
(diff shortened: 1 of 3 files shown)


avikivity:
v2: instead of std::ranges::as_rvalue (C++23 only), use std::ranges::move.

include/seastar/net/tcp.hh
```diff
@@ -1590,27 +1591,9 @@ packet tcp<InetTraits>::tcb::get_transmit_packet() {
...
-    }
-    // hard case: merge some packets, possibly split last
-    auto p = std::move(_snd.unsent.front());
-    _snd.unsent.pop_front();
-    can_send -= p.len();
-    while (!_snd.unsent.empty()
-            && _snd.unsent.front().len() <= can_send) {
-        can_send -= _snd.unsent.front().len();
+    auto p = packet();
+    while (!_snd.unsent.empty() && _snd.unsent.front().size() <= can_send) {
```
xemul, line 1595: This looks like potential deadlock. If `can_send` never grows above the front buffer of `_snd` queue, the queue stops moving forward

xemul:
> Tested with httpd on a tap device, seems to work.

Can you, please, save this work into `tests/manual/` script so that it doesn't get lost and there's a native stack working demo (those in demos/ seem not to work at all) potentially turning into a regular unit test

include/seastar/net/tcp.hh
xemul: Ah, there's continuation if that picks up remaining fraction from front

xemul:
> This is awkward because the packets queued don't correspond to packets to be sent; they have to be cut and pasted to fit the send window or maximum packet size, whichever is greater.

I agree with the reasoning, and the code looks much better (except for the need to accumulate total size of buffers span over and over), but this contradicts [the concept](https://github.com/scylladb/seastar/pull/2937#discussion_r2291262607) of `net::packet` being `struct skbuff`. In-kernel TCP send queue is the list of skbuff-s

avikivity:
> > Tested with httpd on a tap device, seems to work.
> 
> Can you, please, save this work into `tests/manual/` script so that it doesn't get lost and there's a native stack working demo (those in demos/ seem not to work at all) potentially turning into a regular unit test

I just used scripts/tap.sh (actually I had to recreate the bridge, the script assumes libvirt is running and maintaining virbr0).

It's documented in native-stack.md (of course, I didn't read it and had to re-learn all the bits).

avikivity:
> > This is awkward because the packets queued don't correspond to packets to be sent; they have to be cut and pasted to fit the send window or maximum packet size, whichever is greater.
> 
> I agree with the reasoning, and the code looks much better (except for the need to accumulate total size of buffers span over and over), but this contradicts [the concept](https://github.com/scylladb/seastar/pull/2937#discussion_r2291262607) of `net::packet` being `struct skbuff`. In-kernel TCP send queue is the list of skbuff-s

The goal was to be inspired by skbuff, not to clone Linux.

I think the reason the kernel has to use skbuffs while we're better off with temporary_buffer is that we split ~twothree~four responsibilities of skbuff:

1. maintaining reference counters - in temporary_buffer/deleter
2. maintaining headroom/tailroom - in packet
3. grouping memory into a deliberately sized block suitable for the network - in packet
4. maintaining device related bits like checksum offload etc - in packet

The first one is useful for the user-facing layer, but the rest are not (you could say grouping is useful for users, but it would be a different grouping than that useful for the network, so we'd need to ungroup/regroup, the code this patch removes).

Linux skbuff recognizes the problem with headroom/tailroom and supports "headerless" [1] skbs.


Out of curiosity, I looked [2] at the Linux tcp code to see how it splits/merges packets. It splits like we do/did, but merging is delayed to the driver, which presumably doesn't really create a new skbuff, instead it likely just writes the fragments into the ring and knows not to stop in the middle. So the responsibility is split between tcp and the driver.

[1] https://docs.kernel.org/networking/skbuff.html
[2] Claude looked

PR #3105, opened 2025-11-14
Revert posix_impl seastar assert

joe-redpanda:
This assert was added to guard against multiple talkers on the same socket-bound output stream in ossl.

This assert is hitting, but not in ossl, see the below build.

https://buildkite.com/redpanda/redpanda/builds/76335#019a83bf-ac64-48a9-8520-388126d7d7f6

Reverting this assert for the time being and cutting a ticket on above failure


joe-redpanda:
Wrong base

PR #3106, opened 2025-11-15
coroutine: exception: workaround broken destroy coroutine handle in a…

avikivity:
gcc 15.2 and above crash if a coroutine handle is destroyed when await_suspend for that coroutine is called [1]. An internal reference counter is accessed after the coroutine is destroyed.

Work around this by scheduling the work to a task. This is slower, but doesn't crash, and exceptions are supposed to be somewhat rare.

[1] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=121961

### DIFF
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -36,6 +36,20 @@ namespace seastar {
 
 namespace internal {
 
+
+inline
+void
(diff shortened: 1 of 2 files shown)


avikivity:
v2:
 - wrap the magic in a helper function that can be reused elsewhere
 - guard task allocation with critical section like we do for other task allocations

Issue #3103, opened 2025-11-14
seastar-json2code.py reports Unknown type 'integer'


andrzej-jackowski-scylladb:
`scripts/seastar-json2code.py` accepts `int` as the name of numeric type, whereas `integer` seems to be the correct type name according to the Swagger 2.0 specification (https://swagger.io/specification/v2/).

Consider the following `test.json` and `test.def.json` files:

Content of `test.json`:
```
    "/test":{
        "post": {
            "description":"",
            "operationId":"test",
            "tags":[""],
            "parameters":[
                {
                    "name":"body",
                    "in":"body",
                    "required":true,
                    "schema":{
                        "type":"array",
                        "items":{ "$ref":"#/definitions/obj" }
                    }
                }
            ],
            "responses":{
                "200":{ "description": "OK" },
                "default":{
                    "description":"unexpected error"
                }
            }
        }
    }
```
Content of `test.def.json`
 ```
"obj": {
        "id": "obj",
        "properties": {
            "connection_id": {"type": "integer"}
        }
    }
```

`seastar-json2code.py` fails with the following error:
```
./seastar/scripts/seastar-json2code.py --create-cc -f test.json -o test.json.hh
Error while parsing JSON file 'test.json' error Unknown type 'integer' in Model 'obj'
```

If  `"connection_id": {"type": "integer"}` in `test.def.json` is changed to `"connection_id": {"type": "int"}` or `"connection_id": {"type": "string"}`, the script works without a failure.

**Comment by andrzej-jackowski-scylladb** [comment]
@amnonh the fix seems trivial, I think that it's sufficient to extend the following map in `seastar-json2code.py`:
```
valid_vars = {'string': 'sstring', 'int': 'int', 'double': 'double',
             'float': 'float', 'long': 'long', 'boolean': 'bool', 'char': 'char',
             'datetime': 'json::date_time'}
```

Is my understanding of the problem correct, and `'integer': 'int'`should be added to `valid_vars` or there is something I'm missing here?

amnonh:
That's right, just adding a mapping should work. I do think that in older swagger version it was int, so we should keep both mapping

PR #3107, opened 2025-11-17
scripts: add 'integer' type to seastar-json2code

andrzej-jackowski-scylladb:
Swagger 2.0 expects the integer type to be named `integer`, whereas
`seastar-json2code` only supports the `int` type. As a result, running
`seastar-json2code` on a model that contains the `integer` type fails.

This commit extends the list of supported types to allow `integer`
in addition to `int`. The old naming (`int`) is retained for backward
compatibility.

Additionally, tests/unit/api.json is extended to include the 'integer'
type in the model so the json2code script parses it and validates
the fix for issue #3103.

Fixes: scylladb/seastar#3103

### DIFF
--- a/tests/unit/api.json
+++ b/tests/unit/api.json
@@ -94,6 +94,9 @@
           "items": {
             "type": "int"
           }
+        },
+        "integer_var": {
+          "type": "integer"
         }
       }
     }


amnonh (approved):
LGTM

PR #3111, opened 2025-11-19
Fix incorrect union access in dns resolver

xemul:
The resolver maintains a map of sock_entry's each containing union of tcp_entry and udp_entry and its type. There's (at least) one place that accesses the union without checking if it contains the expected object.

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -1350,15 +1350,17 @@ dns_resolver::impl::do_sendv(ares_socket_t fd, const iovec * vec, int len) {
 
         for (;;) {
             // check if we're already writing.
-            if (e.typ == type::tcp && !(e.avail & POLLOUT)) {
-                dns_log.trace("Send already pending {}", fd);
-                errno = EWOULDBLOCK;
-                return -1;
-            }
+            if (e.typ == type::tcp) {
+                if (!(e.avail & POLLOUT)) {
+                    dns_log.trace("Send already pending {}", fd);
+                    errno = EWOULDBLOCK;
+                    return -1;
+                }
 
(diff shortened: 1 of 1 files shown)


src/net/dns.cc
```diff
@@ -1350,15 +1350,17 @@ dns_resolver::impl::do_sendv(ares_socket_t fd, const iovec * vec, int len) {
 
         for (;;) {
             // check if we're already writing.
-            if (e.typ == type::tcp && !(e.avail & POLLOUT)) {
-                dns_log.trace("Send already pending {}", fd);
-                errno = EWOULDBLOCK;
-                return -1;
-            }
+            if (e.typ == type::tcp) {
```
elcallio, line 1353: This is very much a matter of taste. I consider two distinct logical union statements clearer than nested singular ones. It allows you to see the full condition in one line, instead of finding (potentially diff-hidden) other conditionals. 
Won't argue the point much, but this is not really a bugfix, but just a preference change.

xemul: _This_ is not a bugfix indeed, butfix was in the 1st patch of this PR

elcallio: Which is what I meant. No quarrel with that one. Indeed, looks like there is a piece of code missing these, or the member was a primitive at some point, allowing such access.

avikivity: I agree with @xemul here. The two conditionals are for different levels. The first one selects the flow, the second one is conditional behavior within the flow. But I agree it's subjective.

avikivity: And indeed, the fact that two within-the-flow paths are merged strengthens the point.

PR #3113, opened 2025-11-20
Reinforce DNS reverse resolution test 

xemul:
The test does three resolutions in a row -- hostname to address, then address to hostname and finally the reverse-resolved name back to address. The last list of IP-s is checked to contain front IP from the first list. The piggy host for the fun is seastar.io.

Sometimes, reverse resolution fails. E.g. 2b51022073 changed the name from www.google.com to seastar.io with the explanation that google hostname's address cannot be resolved back. Now it happened with seastar.io (and www.google.com healed back). The proposal is to try several hostnames in a row, and tolerate the reverse lookup failure.

### DIFF
--- a/tests/unit/dns_test.cc
+++ b/tests/unit/dns_test.cc
@@ -38,14 +38,24 @@ static const sstring seastar_name = "seastar.io";
 
 static future<> test_resolve(dns_resolver::options opts) {
     auto d = dns_resolver(std::move(opts));
-    {
-        auto hostname = seastar_name;
+
+    for (auto hostname : {"seastar.io", "scylladb.com", "kernel.org", "www.google.com"}) {
         hostent e = co_await d.get_host_by_name(hostname, inet_address::family::INET);
-        hostent a = co_await d.get_host_by_addr(e.addr_list.front());
+        hostent a;
+        try {
+            a = co_await d.get_host_by_addr(e.addr_list.front());
+        } catch (const std::system_error& e) {
+            if (e.code().category() != dns::error_category()) {
+                throw;
+            }
+            continue;
+        }
         hostent e2 = co_await d.get_host_by_name(a.names.front(), inet_address::family::INET);
         BOOST_REQUIRE(std::count(e2.addr_list.begin(), e2.addr_list.end(), e.addr_list.front()));
         co_await d.close();
+        co_return;
     }
+    BOOST_FAIL("No more hosts to try");
 }
 
 static future<> test_bad_name(dns_resolver::options opts) {


xemul:
It's not my localhost problem, github CI also suffers. [The branch](https://github.com/xemul/seastar/tree/br-ci-dns-test-check) that changes nothing, and [the CI job](https://github.com/xemul/seastar/actions/runs/19534143186/) on it that fails

tests/unit/dns_test.cc
```diff
@@ -37,9 +37,22 @@ using namespace seastar::net;
 static const sstring seastar_name = "seastar.io";
 
 static future<> test_resolve(dns_resolver::options opts) {
+    auto hostnames = std::list<std::string>({"seastar.io", "scylladb.com", "kernel.org", "www.google.com"});
     auto d = dns_resolver(std::move(opts));
-    hostent e = co_await d.get_host_by_name(seastar_name, inet_address::family::INET);
-    hostent a = co_await d.get_host_by_addr(e.addr_list.front());
+
+try_again:
```
elcallio: Suggest:
```
while (!hostnames.empty()) {

try {
   auto d = ...;
  break;
} catch(...)
...
  continue;
}

```

xemul: OK, reimplemented with a for-loop

xemul:
upd:
- replaced goto with a for-loop

crackcomm:
It might be a good idea to use these domains:

```bash
$ dig +short one.one.one.one
1.0.0.1
1.1.1.1
$ dig +short dns.google     
8.8.4.4
8.8.8.8
$ dig +short localhost 
127.0.0.1
```

tests/unit/dns_test.cc
```diff
@@ -37,16 +37,15 @@ using namespace seastar::net;
...
-        return d->close();
-    });
+    auto d = dns_resolver(std::move(opts));
+    {
+        auto hostname = seastar_name;
+        hostent e = co_await d.get_host_by_name(hostname, inet_address::family::INET);
+        hostent a = co_await d.get_host_by_addr(e.addr_list.front());
+        hostent e2 = co_await d.get_host_by_name(a.names.front(), inet_address::family::INET);
+        BOOST_REQUIRE(std::count(e2.addr_list.begin(), e2.addr_list.end(), e.addr_list.front()));
+        co_await d.close();
```
avikivity, line 55: This is not an accurate translation, we lose d.close() on exception. And since you do expect failures, we shouldn't leak here.

avikivity:
IMO, a better approach is to set up local dns with dnsmasq as a fixture so we have reliable infrastructure. But I don't known if it's worth the effort.

tests/unit/dns_test.cc
xemul: Failures that are "expected" don't lose close -- the `d` is created once before loop, and in case loop continues it re-uses that single resolver. Exceptions that are not caught make test fail anyway

avikivity: aha

PR #3114, opened 2025-11-21
Update compilers in GH actions scripts

xemul:
Actions run in native gihub ubuntu-24.04 containers, which is the newest available one. Default compliers versions in it are pretty old, but fortunately newer versions are also available from built-in repositories.


xemul:
CI fails on DNS test, fix is #3113. The very same branch with fixed test was run here: https://github.com/xemul/seastar/actions/runs/19571340163

xemul:
upd:
- rebased to pick up dns test fix

PR #3116, opened 2025-11-24
Refine file_impl creation by reactor

xemul:
Creation of file is done in two reactor methods -- open_file_dma() and open_directory(). Both follow similar scheme: first they switch to thread-poll to call open(name) and stat(fd), then call make_file_impl() which returns a future (!) with file_impl pointer, and finally the file is created. The make_file_impl, in turn, is sleeping function because it sometimes (usually once in a lifetime) wants to call thread-poll again to call fstatfs(fd) and populate a collection of fs_info objects.

This PR moves the call to fstats() from make_file_impl() to open_...()'s thread-pool lambda. This change has several interesting consequences.

1. The make_file_impl becomes non-future function, as it really is nowadays
2. The reactor::fstatfs(fd) method is removed
3. One less reactor friend

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2269,19 +2269,6 @@ reactor::file_system_at(std::string_view pathname_view) noexcept {
     co_return ret;
 }
 
-future<struct statfs>
-reactor::fstatfs(int fd) noexcept {
-    syscall_result_extra<struct statfs> sr = co_await _thread_pool->submit<syscall_result_extra<struct statfs>>(
-            internal::thread_pool_submit_reason::file_operation, [fd] {
-        struct statfs st;
-        auto ret = ::fstatfs(fd, &st);
-        return wrap_syscall(ret, st);
-    });
-    sr.throw_if_error();
-    struct statfs st = sr.extra;
-    co_return st;
-}
-
 future<std::filesystem::space_info>
 reactor::file_system_space(std::string_view pathname) noexcept {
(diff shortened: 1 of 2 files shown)


xemul:

xemul:
upd:
- rebased (conflicted with ##2899)

src/core/reactor.cc
```diff
@@ -1805,16 +1805,17 @@ reactor::open_file_dma(std::string_view nameref, open_flags flags, file_open_opt
...
         if (fd == -1) {
-            return wrap_syscall(fd, st);
+            return wrap_syscall(fd, std::make_pair(st, fsi));
         }
         auto close_fd = defer([fd] () noexcept { ::close(fd); });
         int o_direct_flag = _cfg.kernel_page_cache ? 0 : O_DIRECT;
         int r = ::fcntl(fd, F_SETFL, open_flags | o_direct_flag);
         if (r == -1  && strict_o_direct) {
-            auto maybe_ret = wrap_syscall(r, st);  // capture errno (should be EINVAL)
+            auto maybe_ret = wrap_syscall(r, std::make_pair(st, fsi));  // capture errno (should be EINVAL)
```
avikivity, line 1818: I don't think we should be moving fs_info here. First, I'd like to keep the syscall thread minimal since it operates without a full environment (we don't want to allocate here), and second, this code is specific to non-io_uring paths. io_uring is able to execute open() via io_uring and we'd have to duplicate this code.

PR #3117, opened 2025-11-24
Remove net::packet usage from dns.cc

xemul:
The only user of net::packet in dns code is do_sendv() method. It uses packet to copy the buffers provided by the c-ares library. This PR replaces packet with the vector of temporary_buffers.

Also, the newer c-ares librarry calls the do_sendv() with a single C buffer. This case is worth optimizing by not allocating a vector and just keeping a single temporary-buffer around.

There's a mess in do_sendv() inner loop + switch between udp/tcp modes. For the clarity of main patching, the method is brushed up first a little bit.

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -216,7 +216,11 @@ private:
     socket_address sock_addr(const sockaddr * addr, socklen_t len);
     int do_connect(ares_socket_t fd, const sockaddr * addr, socklen_t len);
     ssize_t do_recvfrom(ares_socket_t fd, void * dst, size_t len, int flags, struct sockaddr * from, socklen_t * from_len);
+#if ARES_VERSION >= 0x012200
+    ssize_t do_send(ares_socket_t fd, const void * buf, size_t bytes);
+#else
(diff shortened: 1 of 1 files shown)


src/net/dns.cc
```diff
@@ -216,7 +216,11 @@ class dns_resolver::impl
     socket_address sock_addr(const sockaddr * addr, socklen_t len);
     int do_connect(ares_socket_t fd, const sockaddr * addr, socklen_t len);
     ssize_t do_recvfrom(ares_socket_t fd, void * dst, size_t len, int flags, struct sockaddr * from, socklen_t * from_len);
+#if ARES_VERSION >= 0x012200
+    ssize_t do_sendv(ares_socket_t fd, const void * buf, size_t bytes);
+#else
```
avikivity, line 221: It's wrong to have a function called to_sendv and not receive a vector. With void*, it will even cast an iovec pointer into buf and send it.

xemul: Ugh, I just realised that it is sendV :man_facepalming:

xemul:
upd:
- clarify usage of "v" suffix in send methods

PR #3118, opened 2025-11-24
reactor: Mark some sched-stats getters const

xemul:
The methods that return idle, busy and steal time are naturally const The last one, however, "caches" the result of prevuous evaluation, so the cache members perfectly deserve being marked mutable.


xemul:

PR #3119, opened 2025-11-24
test: Add http rules test of matching order

xemul:
Docs say that routes.add()-ed rules are matched in the order they are added. There can be cases when rules can "conflict" with each other, and these are the cases when this order matters.

Add a test for one of such cases -- url matching with param.

### DIFF
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -395,6 +395,26 @@ SEASTAR_TEST_CASE(test_json_path) {
     });
 }
 
+SEASTAR_TEST_CASE(test_match_rule_order_with_param) {
+    parameters param;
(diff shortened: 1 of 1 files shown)


tests/unit/httpd_test.cc
```diff
@@ -395,6 +395,26 @@ SEASTAR_TEST_CASE(test_json_path) {
     });
 }
 
+SEASTAR_TEST_CASE(test_match_rule_order_with_param) {
```
nyh, line 398: Excuse me for being repetitive like Cato the Elder, but:
It would have been **so useful** to have a comment in front of each test to explain why it exists. The name of the test is never helpful on its own - in this case, what does "with param" even refers to?

PR #3123, opened 2025-11-26
sharded: include cpu info in foreign cpu check

nvartolomei:
I _believe_ this should help debugging.

I didn't test this. Yet.


nvartolomei:
Apologies for noise. Will re-open when ready.

PR #3126, opened 2025-11-28
thread: Remove unused sched_group() helper

xemul:


xemul:

PR #3129, opened 2025-12-02
httpd: Add API to set tcp keepalive params

StephanDollberg:
Allow setting TCP keepalive settings on the httpd connections.

### DIFF
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -135,6 +135,7 @@ class http_server {
     size_t _content_length_limit = std::numeric_limits<size_t>::max();
     bool _content_streaming = false;
     gate _task_gate;
+    std::optional<net::keepalive_params> _keepalive_params;
 public:
     routes _routes;
     using connection = seastar::httpd::connection;
@@ -170,6 +171,10 @@ public:
      */
     [[deprecated("use listen(socket_address addr, server_credentials_ptr credentials)")]]
     void set_tls_credentials(server_credentials_ptr credentials);
+
+    void set_keepalive_parameters(std::optional<net::keepalive_params> params) {
+        _keepalive_params = std::move(params);
+    }
 
     size_t get_content_length_limit() const;
 
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -440,6 +440,10 @@ future<> http_server::do_accepts(int which){
 
 future<> http_server::do_accept_one(int which, bool tls) {
     return _listeners[which].accept().then([this, tls] (accept_result ar) mutable {
+        if (_keepalive_params) {
+            ar.connection.set_keepalive(true);
+            ar.connection.set_keepalive_parameters(_keepalive_params.value());
+        }
         auto local_address = ar.connection.local_address();
         auto conn = std::make_unique<connection>(*this, std::move(ar.connection),
                 std::move(ar.remote_address), std::move(local_address), tls);


include/seastar/http/httpd.hh
```diff
@@ -171,6 +172,10 @@ public:
     [[deprecated("use listen(socket_address addr, server_credentials_ptr credentials)")]]
     void set_tls_credentials(server_credentials_ptr credentials);
 
+    void set_keepalive_parameters(const net::keepalive_params& params) {
```
travisdowns: Perhaps it should take an `optional<>`? Otherwise it's a "one way" API - once you've set the params, you can only modify their value but no longer disable keepalive.

StephanDollberg: Yes good idea. Fixed

avikivity:
Please enhance the patch changelog with an explanation about why it's useful. I would have expected it to be done on the client side.

Isn't the server allowed to close unused connections? I think that's better than keeping them alive artificially.

StephanDollberg:
> Please enhance the patch changelog with an explanation about why it's useful. 

I have amended the commit message.

> I would have expected it to be done on the client side.

Both sides may want to enable them. 

> Isn't the server allowed to close unused connections? I think that's better than keeping them alive artificially.

Keepalives aren't really supposed to keep connections alive artificially but to detect broken connections. 

What we are seeing with AWS load balancers is that they silently drop idle connection state after a while. This is transparent to the server, the connections just stays established forever. 

Given httpd doesn't offer any way for uselevel timeouts using TCP keepalives is a cheap way to detect this state. 

We are already using this for other connection types besides our HTTP endpoint.

src/http/httpd.cc
```diff
@@ -440,6 +440,10 @@ future<> http_server::do_accepts(int which){
 
 future<> http_server::do_accept_one(int which, bool tls) {
     return _listeners[which].accept().then([this, tls] (accept_result ar) mutable {
+        if (this->_keepalive_params) {
```
nyh: style nitpick: as you can see elsewhere in this file, we don't usually use `this->` for class members. The "_" is already a visual indicator that the variable is a member.

include/seastar/http/httpd.hh
```diff
@@ -171,6 +172,10 @@ public:
     [[deprecated("use listen(socket_address addr, server_credentials_ptr credentials)")]]
     void set_tls_credentials(server_credentials_ptr credentials);
 
+    void set_keepalive_parameters(std::optional<net::keepalive_params> params) {
```
nyh, line 175: I wonder if anybody would ever want to call this function with a disengaged optional. Wouldn't you only call it if you want to set the parameters? Maybe the parameter of this function shouldn't be an optional?
(it's not really wrong that it's optional, I just find it strange).

avikivity:
> > Please enhance the patch changelog with an explanation about why it's useful.
> 
> I have amended the commit message.
> 
> > I would have expected it to be done on the client side.
> 
> Both sides may want to enable them.
> 
> > Isn't the server allowed to close unused connections? I think that's better than keeping them alive artificially.
> 
> Keepalives aren't really supposed to keep connections alive artificially but to detect broken connections.


Why would a server care? It's passive.

> 
> What we are seeing with AWS load balancers is that they silently drop idle connection state after a while. This is transparent to the server, the connections just stays established forever.

Do they not close the connection first?

If they don't, this is reasonable justification.

> 
> Given httpd doesn't offer any way for uselevel timeouts using TCP keepalives is a cheap way to detect this state.
> 
> We are already using this for other connection types besides our HTTP endpoint.

httpd should have the ability to garbage-collect idle connections.

This puts the keep-alive onus on the client, where it should be.

Suppose we replaced this with a per-connection timer that closes the connection after N minutes if idleness: wouldn't it be better? avoid the (small amount of) work in sending keepalives.

nyh:
> What we are seeing with AWS load balancers is that they silently drop idle connection state after a while. This is transparent to the server, the connections just stays established forever.
> 
> Given httpd doesn't offer any way for uselevel timeouts using TCP keepalives is a cheap way to detect this state.

These are very good points. I opened https://github.com/scylladb/seastar/issues/3130 to consider whether we need to make some sort of idle-connection-dropping on by default, and not just if the application author explicitly remembers to enable this feature (like you allow in your patch). I also wonder if we really need to bother with checking the client (like keepalive does), or maybe it's enough to simply drop old idle connections, without even checking if the other side is alive.

nyh:
> httpd should have the ability to garbage-collect idle connections.
> 
> This puts the keep-alive onus on the client, where it should be.

I agree. I opened https://github.com/scylladb/seastar/issues/3130 asking that we enable some sort of idle-connect cleanup feature in httpd **by default**. TCP keep-alive is one easy way to implement it (since the kernel does the work for us...) but perhaps not the best, because:

> Suppose we replaced this with a per-connection timer that closes the connection after N minutes if idleness: wouldn't it be better? avoid the (small amount of) work in sending keepalives.

I wholeheartedly agree. This is the second option I suggested in https://github.com/scylladb/seastar/issues/3130. This is also what all other popular http servers do (I gave a few examples in the issue).

travisdowns:
> Do they not close the connection first?

It depends, but big ones like AWS ones don't: when the idle timeout is reached the load balancer just stops tracking the connection, it doesn't send anything to either end. If either end sends something after that, the LB doesn't even recognize the flow and sends back RST to the sending side.

See e.g.:

https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#connection-idle-timeout

This kind of thing is a constant source of connection "leaks", since even if the client does itself has a graceful timeout, if that's longer than the LB timeout, the graceful close will never get through, and the server just keeps the connection open forever thinking there is a connected client on the other side.

include/seastar/http/httpd.hh
StephanDollberg: I have no strong feelings either way, I think most people just set it once and forget about it but as Travis suggested in another comment one could imagine that one might want to reset it.

src/http/httpd.cc
StephanDollberg: Dropped `this`.

StephanDollberg:
> Do they not close the connection first?

No, this specific one doesn't unfortunately. 

Yes user level timeouts are good to have on top but as discussed a bit heavier than just TCP keepalives.

include/seastar/http/httpd.hh
nyh: I see. Makes sense. Ok.

src/http/httpd.cc
```diff
@@ -440,6 +440,10 @@ future<> http_server::do_accepts(int which){
 
 future<> http_server::do_accept_one(int which, bool tls) {
     return _listeners[which].accept().then([this, tls] (accept_result ar) mutable {
+        if (_keepalive_params) {
+            ar.connection.set_keepalive(true);
+            ar.connection.set_keepalive_parameters(this->_keepalive_params.value());
```
nyh: you left this-> here :-)

nyh (approved):
One last "this->". I'm anticipating the future and already marking "approve".

src/http/httpd.cc
StephanDollberg: Indeed, fixed.

nyh:
We had a CI failure in
```
/home/runner/work/seastar/seastar/tests/unit/scheduling_group_nesting_test.cc(132): error: in "test_fairness": check dev < 0.07 has failed
```
I don't see how it can be related to this patch, I'll try to run the CI again.
@xemul could this test be flaky?

xemul:
@nyh , yes it can. I've seen it couple of times before, but it was too rare to attract enough attention :(

PR #3133, opened 2025-12-03
Make datagram API use temporary_buffer<>-s

xemul:
To read data from `datagram_channel` one calls `receive()` on it, co-awaits the returned future into `datagram` object and then grabs `net::packet` reference by calling `datagram::get_data()` on it.

This PR deprecates the `get_data()` and suggest `get_bufs()` datagram method that returns `span<temporary_buffer<char>>`. The implementations of datagram-s are updated to also maintain temporary_buffer-s internally.

### DIFF
--- a/include/seastar/net/api.hh
+++ b/include/seastar/net/api.hh
@@ -110,6 +110,7 @@ public:
     socket_address get_src() { return _impl->get_src(); }
     socket_address get_dst() { return _impl->get_dst(); }
     uint16_t get_dst_port() { return _impl->get_dst_port(); }
+    [[deprecated("Use get_buf() instead")]]
     packet& get_data() {
         if (!_p) {
             _p.emplace(_impl->get_buffers());


include/seastar/net/api.hh
```diff
@@ -92,20 +92,27 @@ public:
     virtual socket_address get_src() = 0;
     virtual socket_address get_dst() = 0;
     virtual uint16_t get_dst_port() = 0;
-    virtual packet& get_data() = 0;
+    virtual temporary_buffer<char>& get_data() = 0;
```
avikivity: This forces a copy. It's sad for QUIC which is UDP based (although: QUIC uses path MTU discovery so perhaps won't see fragmentation).

include/seastar/net/api.hh
```diff
@@ -92,20 +92,27 @@ public:
...
     socket_address get_dst() { return _impl->get_dst(); }
     uint16_t get_dst_port() { return _impl->get_dst_port(); }
-    packet& get_data() { return _impl->get_data(); }
+    packet& get_data() {
+        if (!_p) {
+            _p.emplace(std::move(_impl->get_data()));
+        }
+        return _p.value();
+    }
+    temporary_buffer<char>& get_buf() { return _impl->get_data(); }
```
avikivity: If we change the API, we might try to avoid the references. Though it's not easy.

Perhaps: `temporary_buffer<>& buffer(size_t idx)` (not avoiding the reference, but avoiding linearization)

src/net/udp.cc
```diff
@@ -52,15 +52,19 @@ class native_datagram : public datagram_impl {
...
-    native_datagram(ipv4_address src, ipv4_address dst, packet p)
-            : _p(std::move(p)) {
+    native_datagram(ipv4_address src, ipv4_address dst, packet _p)
+    {
         udp_hdr* hdr = _p.get_header<udp_hdr>();
         auto h = ntoh(*hdr);
         _p.trim_front(sizeof(*hdr));
         _src = to_ipv4_addr(src, h.src_port);
         _dst = to_ipv4_addr(dst, h.dst_port);
+        _p.linearize();
```
avikivity: :(

We'd have fragments with posix stack io_uring IOSQE_BUFFER_SELECT. It allows you to populate a pool of (say) 1k buffers, then a recvmsg can return a vector of length 1 for a short packet or a vector of length 9 for a jumbo frame, or even a vector of length 64 for a maximum UDP packet, without wasting too much memory.

include/seastar/net/api.hh
nyh: UDP protocol designers typically try hard to avoid UDP fragmentation. So I think it's useful to have a function that is **optimized** for the case of no fragmentation and returns a single chunk. 

Still, why not _also_ have a function that can return fragmented content? For example in ScyllaDB we use the type `using chunked_content = std::vector<temporary_buffer<char>>;`. I think we need something like that in Seastar. In fact, isn't net::packet already something similar? I have a vague recollection we have yet another similar object, but can't recall why.

nyh: I just realized that it's not the get_data() function that does the linearization, you do it unconditionally. So you're not even leaving the option of not doing this copy, it's always done when the fragmented packet is received.  I hope that I at least understand correctly that in the common case of no fragmentation, a copy isn't done.

xemul: How does it force a copy? The temporary_buffer is not copy-constructible, in the first place. This (abstract) method returns a reference to internal buffer, and if the caller doesn't want to explicitly copy it, it can `.share()` one.

include/seastar/net/api.hh
xemul: I'll try to get back to the span-returning branch then. It was pretty hairy, that's why I decided that linearization would be OK for UDP

src/net/udp.cc
xemul: Out of curiosity. Is it theoretical interest to have first-class support for up to maxumum UDP packet, or are there practical examples when such packets are in use (with scylla or some other seastar-based project)?

include/seastar/net/api.hh
xemul: > I hope that I at least understand correctly that in the common case of no fragmentation, a copy isn't done.

Ah, I should've looked inside `packet::linearize()`. Even if the packet is single-fragment, it will copy it into another single fragment. But anyway, the support for fragmentation is needed, so this will be reworked.

include/seastar/net/api.hh
avikivity: It's okay when UDP==DNS, but not okay when UDP==QUIC.

src/net/udp.cc
avikivity: I think UDP > MTU is very rare.

However, realistic MTU is 9000 bytes, and since we don't want to pre-allocate 9k per packet, we can expect memory fragmentation in normal operation (not IP fragmentation).

xemul:
v2:
- return `std::span<temporary_buffer<char>>`

include/seastar/net/api.hh
xemul: Are there QUIC implementations over seastar? Maybe in plans?

avikivity: ZPP

include/seastar/net/api.hh
```diff
@@ -92,20 +92,27 @@ public:
     virtual socket_address get_src() = 0;
     virtual socket_address get_dst() = 0;
     virtual uint16_t get_dst_port() = 0;
-    virtual packet& get_data() = 0;
+    virtual std::span<temporary_buffer<char>> get_bufs() = 0;
```
avikivity: Seastar uses complete words, let's use get_buffers().

include/seastar/net/api.hh
```diff
@@ -92,20 +92,27 @@ public:
...
 };
 
 using udp_datagram_impl = datagram_impl;
 
 class datagram final {
 private:
     std::unique_ptr<datagram_impl> _impl;
+    std::optional<net::packet> _p;
```
avikivity, line 107: packet is nullable (has operator bool), no need for std::optional.

src/net/udp.cc
```diff
@@ -52,15 +52,16 @@ class native_datagram : public datagram_impl {
 private:
     ipv4_addr _src;
     ipv4_addr _dst;
-    packet _p;
+    std::vector<temporary_buffer<char>> _bufs;
```
avikivity, line 55: The vector is wasteful, but not more wasteful than packet, so okay.

include/seastar/net/api.hh
xemul: The get_data() should tell "packet wasn't initialized from get_buffers()" from "packet was initialized from get_buffers(), and then later release()-d by user"

src/net/posix-stack.cc
```diff
@@ -1178,7 +1178,7 @@ posix_datagram_channel::receive() {
         auto sg_id = internal::scheduling_group_index(current_scheduling_group());
         bytes_received[sg_id] += size;
         return make_ready_future<datagram>(datagram(std::make_unique<posix_datagram>(
-            _recv._src_addr, dst ? *dst : _address, packet(fragment{_recv._buffer, size}, make_deleter([buf = _recv._buffer] { delete[] buf; })))));
+            _recv._src_addr, dst ? *dst : _address, temporary_buffer<char>(_recv._buffer, size, make_deleter([buf = _recv._buffer] { delete[] buf; })))));
```
avikivity, line 1181: This is extremely wasteful - we allocate 64k for each datagram. But not the fault of this patch.

include/seastar/net/api.hh
avikivity: ok. I expect it will be gone by the time we need really fast UDP.

avikivity:
Looks good.

xemul:
v3:
- rename get_bufs() to get_buffers()
- add comment about usage of optional<> in deprecated get_data()

PR #3134, opened 2025-12-03
test,perf: Measure time it takes to traverse a runqueue

xemul:
Add a test that populates run-queue and measures the time it takes for the scheduler to process one.
Useful to measure the effect the runqueue implementation has.

### DIFF
--- a/include/seastar/testing/perf_tests.hh
+++ b/include/seastar/testing/perf_tests.hh
@@ -253,6 +253,10 @@ inline void stop_measuring_time()
     internal::time_measurement_stop_iteration();
 }
 
+namespace internal { std::string get_parameter(std::string); }
+inline std::string get_parameter(std::string name) {
+    return internal::get_parameter(std::move(name));
+}
 
 template<typename T>
 void do_not_optimize(const T& v)
--- a/tests/perf/future_util_perf.cc
+++ b/tests/perf/future_util_perf.cc
@@ -19,10 +19,12 @@
  * Copyright (C) 2018 ScyllaDB Ltd.
  */
 
+#include <random>
 #include <boost/range.hpp>
 #include <boost/range/irange.hpp>
 
 #include <seastar/testing/perf_tests.hh>
+#include <seastar/testing/random.hh>
 #include <seastar/core/coroutine.hh>
 #include <seastar/coroutine/as_future.hh>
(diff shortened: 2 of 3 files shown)


xemul:
upd:
- shuffle the tasks (once -- upon creation) for [this](https://github.com/scylladb/seastar/pull/3125#issuecomment-3610881364) test

xemul:
upd:
- shortened with the help of parametrization patch from #3138

xemul:
@avikivity , please review

xemul:
@avikivity , what's your thoughts on this test? Maybe it can be useful for [future run-list rework](https://github.com/avikivity/seastar/commits/sort-tasks/), if not, then I'd just close this PR

travisdowns:
@xemul looking at this one...

In what context do we do a full scan of the runqueues? That's the primary case the split list optimizes, right?

xemul:
@travisdowns , the context is reactor::task_queue::run_tasks -- it loops though the collection of tasks and runs each. It's not exactly the plain full scan, as it a) calls virtual run() and b) may stop scanning if task quota exceeds, but still, if the queue is full of short continuations, the approximation is good enough. And yes, the split-list thing was supposed to handle exactly this case and kill two birds with one stone -- keep the full-scan efficiency and avoid memory (re)allocation of add_task()

travisdowns (approved):
LGTM but needs a rebase

xemul:
upd:
- rebased
- added involcation exaple to the 2nd patch (with the test itself)

travisdowns:
@xemul happy to merge this, or are you still waiting for a y/n from Avi per above?

xemul:
Yes, would be good to have Avi's ack/nack, the test is only needed if doing something with the tasks runqueue

travisdowns:
@avikivity ping

xemul:
@avikivity ping

xemul:
@avikivity what's your take on this? If it's valuable to have "runqueue traverse time" perf test in the tree?

xemul:
@avikivity , a light reminder ping

avikivity:
Sorry for the long delay. Maybe we should have a test for the time it takes me to review something.

avikivity:
I'll use the slot to advertise my instruction-cache friendly variant: https://github.com/scylladb/seastar/pull/3547

PR #3136, opened 2025-12-05
coroutine/try_future: call set_current_task() when resuming the coroutine

denesb:
After calling hndl.resume(), the try_future_awaiter is destroyed so the reactor's _current_task becomes dangling pointer. Call set_current_task() with the coroutine's task to communicate the task switch to the reactor.

Bug introduced in https://github.com/scylladb/seastar/commit/8e749a0975a649686a4a650953a813df136a2966.

### DIFF
--- a/include/seastar/coroutine/try_future.hh
+++ b/include/seastar/coroutine/try_future.hh
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <seastar/core/coroutine.hh>
+#include <seastar/core/internal/current_task.hh>
 
 namespace seastar::internal {
 
@@ -34,6 +35,7 @@ void try_future_resume_or_destroy_coroutine(seastar::future<T>& fut, seastar::ta
         hndl.promise().set_exception(std::move(fut).get_exception());
         hndl.destroy();
     } else {
+        set_current_task(promise_ptr);
         hndl.resume();
     }
 }


include/seastar/coroutine/try_future.hh
```diff
@@ -34,6 +35,7 @@ void try_future_resume_or_destroy_coroutine(seastar::future<T>& fut, seastar::ta
         hndl.promise().set_exception(std::move(fut).get_exception());
         hndl.destroy();
     } else {
+        set_current_task(promise_ptr);
```
bhalevy, line 38: Should this be done in hndl.resume()?
Are there other paths that call resume that may be suceptible to the same issue?

Also, should we open a GH issue for tracking?

denesb: Usually `hndl.resume()` is called by the coroutine's promise,so the current task doesn't change.
This is a corner case, because the `try_future_awaiter` resumes the coroutine instead. Apparently, this corner case happened before, because we have this `set_current_task()` with a comment suggesting that this is the intended usage.

Not sure what is the issue policy in seastar. This is very recently added code, so I just sent a PR instead of opening an issue.

bhalevy: Please just document the commit that introduced the issue in the patch log message

include/seastar/coroutine/try_future.hh
```diff
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <seastar/core/coroutine.hh>
+#include <seastar/core/reactor.hh>
```
avikivity: Please avoid reactor.hh includes in header files, it's heavyweight.

denesb:
New in v2:
* Don't include reactor.hh, just forward-declare set_current_task() instead.
* Add mention of commit which introduced the bug in the commit message.

include/seastar/coroutine/try_future.hh
denesb: Added mention of the patch in v2.

include/seastar/coroutine/try_future.hh
denesb: I dropped the include and added forward-declare of `set_current_task()`. It works but I wonder how, `set_current_task()` is an inline function. How is the linker resolving the reference?

avikivity: It's likely still included via a different route:

```console
$ git grep reactor.hh | grep hh:
apps/lib/stop_signal.hh:#include <seastar/core/reactor.hh>
include/seastar/coroutine/parallel_for_each.hh:#include <seastar/core/reactor.hh>
```

So if we remove those, it will break.

avikivity: (or if you use it from a translation unit that doesn't include parallel_for_each.hh)

avikivity: Suggest to move local_engine and set_current_task() to a new internal header, and include that.

denesb:
New in v3:
* Extract `set_current_task()` to separate header and make it out-of-line.

include/seastar/coroutine/try_future.hh
denesb: Fixed in v3.

denesb:
New in v3.1:
* Fixed fallout from removing `#include <seastar/core/reactor.hh>` from `coroutine/parallel_for_each.hh`.

avikivity:
Since you made it out-of-line, a forward declaration would have been sufficient.

Since everyone uses lto, out-of-lineness doesn't matter so much.

PR #3140, opened 2025-12-09
Remove posix file helper file_read_state class

xemul:
The class was used when the bulk reading code used continuation chains. Now it's coroutinized and the "state" is just confusing. Also add a test, that passes before removal (and rework) and after it.

A test actually notices one ... weirdness -- when the bulk reading method crosses the file-size boundary, it actually emits a single unaligned IO, that succeeds with zero size.

refs #2998


xemul:
upd:
- use random-generated buffer (it was handy to debug it with "integers", but it's cleaner to use random engine)

PR #3141, opened 2025-12-09
file: Finally override dma_read_bulk() in posix_file_impl

xemul:
There are three files that inherit from posix_file_impl -- the blockdev file, append-challenged file and posix file per-se. All three override the bulk reading method by calling their base class implementation.

This method was implemented like this for append-challenged file, which runs its IO via internal queue. However, for bulk reading method this queue is avoided, because "real" IO happens inside do_dma_read_bulk() -- it allocates properly aligned buffers and calls read_dma() on it.

Said that, there's no need in re-overriding the dma_read_bulk() for all those sub-classes, the base posix_file_impl final override is enough.


xemul:

xemul:

PR #3142, opened 2025-12-09
add missing headers

travisdowns:
span and optional were missing from two files which used std::span and std::optional, respectively. This fails to compile unless those happen to be included indirectly via other headers (in practice, it fails to compile for me on Ubuntu 22.04 and distro libstc++).


travisdowns:
cc @xemul

PR #3143, opened 2025-12-09
net/dns: fixed socket concurrent access

travisdowns:
The problem with DNS queries is caused by the fact that the UDP socket is used concurrently i.e. write is scheduled while the other one is pending.

DNS queries are based on c-ares library. The library is initialized with a set of functions (callbacks) that are called whenever there is a need to send, receive, create socket, connect or close.

Fixed processing of send requests leading to assertion being triggered in reactor.

The fix involves checking if send is already pending and if so returning an error. The error returned from send callback will instruct c-ares library to retry the query.

(cherry picked from commit 74d5e9cc02ce9e75be01e4985df6b28a9bac750e)

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -1442,17 +1442,15 @@ dns_resolver::impl::do_sendv(ares_socket_t fd, const iovec * vec, int len) {
 
 
             // check if we're already writing.
-            if (e.typ == type::tcp) {
-                if (!(e.avail & POLLOUT)) {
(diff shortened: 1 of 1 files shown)


travisdowns:
@xemul this problem was observed occurring in practice, so we apply the same block of concurrent sends we already have on the TCP side. Note that this effective undoes your refactoring of these branches from a few weeks ago since the latter condition now no longer shares the same conditions as the first.

PR #3145, opened 2025-12-10
API level 8 build fixes

### LINKED ISSUE #3144: The tests do not build at API_LEVEL=8
When using `--api-level=8` (on less than the current level 9), `ninja seastar` is OK, but trying to build the tests failed (largely around the input stream adjustments to use `temporary_buffer` over `packet`).

I'm not sure if this is considered a bug and if patches would be accepted to fix this, or if this is by design. 

Currently we are on API level < 9 so we have an interest here as we run seastar tests in CI on our fork.

travisdowns:
A few things weren't building on API level 8, fix them.

Details in the commits.

Fixes #3144 .


apps/CMakeLists.txt
```diff
@@ -52,5 +52,7 @@ add_subdirectory (httpd)
 add_subdirectory (io_tester)
 add_subdirectory (rpc_tester)
 add_subdirectory (iotune)
-add_subdirectory (memcached)
+if (${Seastar_API_LEVEL} GREATER_EQUAL 9)
+  add_subdirectory (memcached)
+endif ()
```
xemul, line 57: Not only memcached. With non-maximum API level, tests, demos and all apps are not expected to compile either

travisdowns: Currently only memcached doesn't compile, which is why I excluded it only, h/e I could exclude all of them if you wish. Currently other tests do have code such that they compile on API level < 9. 

Is this patch desirable here upstream? If not I will close it.

xemul: > Currently only memcached doesn't compile, ...

And tests too, there's 2nd patch in this PR :)

> Is this patch desirable here upstream? If not I will close it.

Yes, the patch is very welcome. My comment was rather about the 2nd patch from this PR -- the tests should be excluded, not fixed.

travisdowns: > And tests too, there's 2nd patch in this PR :)

Yes sorry, I was saying that in the scope of that the list of applications in that file, the only ones compile.

> the tests should be excluded, not fixed.

I see, thanks for the clarification. 

We don't want to lose these tests so I will probably just keep this patch on our fork then. If you want I can update this one exclude memcached only.

travisdowns:
@xemul the test failure is a flake. Should we increase that threshold?

apps/CMakeLists.txt
xemul: > We don't want to lose these tests so I will probably just keep this patch on our fork then. If you want I can update this one exclude memcached only.

Yes, please

xemul:
> @xemul the test failure is a flake. Should we increase that threshold?

Yes, I guess that's the only thing we can do here. Thank you

travisdowns:
@xemul most recent push drops the test changes and keeps only the memcached side of things.

travisdowns:
> Yes, I guess that's the only thing we can do here. Thank you

Over here https://github.com/scylladb/seastar/pull/3172

PR #3146, opened 2025-12-10
`util/variant_utils`: correct type deduction for `seastar::visit`

WillemKauf:
`auto` decays reference, and therefore `seastar::visit` cannot at present return l-value references.

Through use of `decltype(auto)` instead, we can preserve the exact type, including references.

Previously, code like this:

```cpp
auto& r = seastar::visit(std::variant<int, bool>{true},
                         [&t](const int& i) -> noncopyable_type& { return t; },
                         [&t](const bool& b) -> noncopyable_type& { return t; });
```

would be a compilation error:

```
error: cannot bind non-const lvalue reference of type ‘noncopyable_type&’ to an rvalue of type ‘noncopyable_type’
```

 It is now made valid.

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -771,6 +771,9 @@ seastar_add_test (pipe
 seastar_add_test (spawn
   SOURCES spawn_test.cc)
 
+seastar_add_test (variant_utils
+  SOURCES variant_utils_test.cc)
+
 seastar_generate_swagger (
   TARGET rest_api_httpd_swagger
   VAR rest_api_httpd_swagger_files
--- a/tests/unit/variant_utils_test.cc
+++ b/tests/unit/variant_utils_test.cc
@@ -1,0 +1,65 @@
+#define BOOST_TEST_MODULE core
+
+#include <seastar/util/variant_utils.hh>
+
+#include <boost/test/unit_test.hpp>
+
+#include <variant>
+
+struct noncopyable_type {
+    noncopyable_type() = default;
+    ~noncopyable_type() = default;
+    noncopyable_type(const noncopyable_type&) = delete;
+    noncopyable_type& operator=(const noncopyable_type&) = delete;
+    noncopyable_type(noncopyable_type&&) noexcept = default;
+    noncopyable_type& operator=(noncopyable_type&&) noexcept = default;
+};
+
+using variant_t = std::variant<int, bool>;
+static noncopyable_type t{};
+
+static_assert(
+  std::is_same_v<
+    decltype(seastar::visit(
+      variant_t{true},
+      [](const int& i) -> noncopyable_type { return noncopyable_type{}; },
+      [](const bool& b) -> noncopyable_type { return noncopyable_type{}; })),
+    noncopyable_type>);
+
+static_assert(std::is_same_v<
+              decltype(seastar::visit(
+                variant_t{true},
+                [](const int& i) -> noncopyable_type& { return t; },
+                [](const bool& b) -> noncopyable_type& { return t; })),
+              noncopyable_type&>);
+
+static_assert(
+  std::is_same_v<
+    decltype(std::visit(
+      [](auto&& v) -> noncopyable_type& { return t; }, variant_t{true})),
+    decltype(seastar::visit(
+      variant_t{true},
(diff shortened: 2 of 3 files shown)


tests/unit/variant_utils_test.cc
```diff
@@ -0,0 +1,58 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Copyright (C) 2025 ScyllaDB
```
nyh: Nice of you to contribute the copyright to ScyllaDB, but I don't think you have to do it.

tests/unit/variant_utils_test.cc
```diff
@@ -0,0 +1,58 @@
...
+          })),
+        noncopyable_type>);
+
+    noncopyable_type t;
+    static_assert(std::is_same_v<
+                  decltype(seastar::visit(
+                    std::variant<int, bool>{true},
+                    [&t](const int& i) -> noncopyable_type& { return t; },
+                    [&t](const bool& b) -> noncopyable_type& { return t; })),
+                  noncopyable_type&>);
```
nyh: To be honest I'm not even sure why this case should be allowed why would someone ever do "return t" when the type of t is known to be a noncopyable type?

Maybe you can add to this test also a similar check for std::visit, to "prove" that std::visit works exactly like this as well. This will be convincing that Seastar should allow this to. Looking at documentation I believe you're right, and std::visit allows this too - but the test doesn't show this.

tests/unit/variant_utils_test.cc
```diff
@@ -0,0 +1,58 @@
...
+}
```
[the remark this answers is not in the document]
nyh: That's a good point.

By the way, if you do add static_assert() tests, there is no reason why they need to be inside a function - they can be outside any test function. For example in chunked_fifo_test.cc we have both a bunch of static_assert() checks, and then additionally test functions which actually run (and don't contain static_assert).

tests/unit/variant_utils_test.cc
nvartolomei: > To be honest I'm not even sure why this case should be allowed why would someone ever do "return t" when the type of t is known to be a noncopyable type?

That's a surprising comment to me. What would you expect people to write?

nyh: I didn't notice the `-> noncopyable_type&` return type in the lambda's signature. That indeed makes the "return t" natural. Youre right.

tests/unit/variant_utils_test.cc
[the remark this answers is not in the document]
WillemKauf: Sure, moved the static asserts outside the test.

tests/unit/variant_utils_test.cc
WillemKauf: Removed.

tests/unit/variant_utils_test.cc
WillemKauf: Yes, `std::visit` does allow this, and we had some uses of `std::visit` in the `redpanda` codebase which did return an l-value reference- so when attempting to rewrite those chunks with `ss::visit`, we would get a compilation error for the same code.

WillemKauf: Added some test cases to ensure parity between `std::visit` and `seastar::visit`.

tests/unit/CMakeLists.txt
```diff
@@ -771,6 +771,9 @@ seastar_add_test (pipe
 seastar_add_test (spawn
   SOURCES spawn_test.cc)
 
+seastar_add_test (variant_utils
```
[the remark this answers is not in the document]
nyh: While it seems that copilot is right here, looking at other tests, it's also evident that the CI passed so nothing went wrong with the wrong "KIND". So why is this KIND needed at all?
Who's the expert in our cmake build? I know nothing about it :-(

nyh (approved):

WillemKauf:
Good to merge, @nyh, or should we be requesting more reviews?

PR #3147, opened 2025-12-10
posix_data_sink_impl: Assert to guard put UB

joe-redpanda:
Posix stack's data_sink implementation stores a net packet locally and passes it by reference to the file descriptor.

If put is called concurrently, UB can occur.
With two invocations, one faster and one slower the faster invocation will reset the packet on completion of its write_all invocation, and the slower may dereference a nullptr in using the referenced packet

This commit changes three things:
1. it initializes the net::packet in posix_data_sink_impl to a null packet
2. reset of posix_data_sink_impl's net::packet will now occur in a finally s.t. the contained packet will also be reset on exceptions
3. invocation of posix_data_sink_impl::put will now assert that the contained packet is null before overwriting.

Any concurrent usages should now fail on an assert.

### DIFF
--- a/src/net/posix-stack.cc
+++ b/src/net/posix-stack.cc
@@ -891,10 +891,11 @@ posix_data_sink_impl::put(temporary_buffer<char> buf) {
 
 future<>
 posix_data_sink_impl::put(packet p) {
+    SEASTAR_ASSERT(!_p);
     _p = std::move(p);
     auto sg_id = internal::scheduling_group_index(current_scheduling_group());
     bytes_sent[sg_id] += _p.len();
-    return _fd.write_all(_p).then([this] { _p.reset(); });
+    return _fd.write_all(_p).finally([this] { _p.reset(); });
 }
 #endif
 
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -144,7 +144,7 @@ class posix_data_sink_impl : public data_sink_impl {
 #if SEASTAR_API_LEVEL >= 9
     internal::wrapped_iovecs _vecs;
 #else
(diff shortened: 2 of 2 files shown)


xemul:
data_sink::put() is not supposed to be called concurrently, before previous future resolves

src/net/posix-stack.cc
```diff
@@ -891,10 +891,11 @@ posix_data_sink_impl::put(temporary_buffer<char> buf) {
 
 future<>
 posix_data_sink_impl::put(packet p) {
+    SEASTAR_ASSERT(!_p);
     _p = std::move(p);
     auto sg_id = internal::scheduling_group_index(current_scheduling_group());
     bytes_sent[sg_id] += _p.len();
-    return _fd.write_all(_p).then([this] { _p.reset(); });
+    return _fd.write_all(_p).finally([this] { _p.reset(); });
```
xemul, line 898: This "finally" is a good catch :) In case sending fails, the saved packet doesn't need to stay until it's overwritten by next put. Even if the put() is called as it should be -- sequentially

joe-redpanda: Yep, this was intended to be a cheap-ish check on "what happens when it isn't used properly" because its a race toward UB making mistakes here fairly difficult to catch

It looks like API_VERSION 9+ has a similar check [here](https://github.com/scylladb/seastar/blob/a365d56a3896e191b051886a37063fb597026ea2/src/net/posix-stack.cc#L101)

xemul: I see. Yes, this makes perfect sense. I'd suggest using `on_internal_error` here either, but it's "outdated" API level, so .. good enough

PR #3148, opened 2025-12-10
httpd: add timeout on requests

### LINKED ISSUE #3130: httpd should close connections that have been idle for a long time
https://github.com/scylladb/seastar/pull/3129 reports that: "What we are seeing with AWS load balancers is that they silently drop idle connection state after a while. This is transparent to the server, the connections just stays established forever.".

For this and many other reasons (clients disappearing, rebooting, etc.) our http server might end up with a very large number of idle connection, supposedly kept alive for the client to reuse them but the client will never do and these connections will never get closed. This is bad, and even dangerous.

In #3129, it was suggested to use kernel-level TCP keepalive (with low settings) to solve this problem. But #3129 did not turn this feature on by default - it just provided a mechanism. I'm suggesting here that perhaps we should turn this feature on by default.

Alternatively, instead of using TCP keepalive, we could do something even simpler in user-space - we could automatically close any idle connection which is older than some time limit - not even checking if the client is still alive. For example, Apache HTTPd has KeepAliveTimeout https://httpd.apache.org/docs/2.4/mod/core.html#keepalivetimeout - which defaults to just 5 seconds (!). Nginx has keepalive_timeout https://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_timeout which defaults to 75 seconds. Every other HTTP server I looked at seems to have such a timeout, and its default configuration is usually just a few seconds. I don't know if Seastar's default should be so low, but infinity is too high :-)

nyh:
This patch adds to http_server a set_request_timeout() method which enables a time limit on reading a request from a connection. When this timeout is reached before a request is read, the connection is closed.

The primary goal for this feature is to clean up persistent connections (HTTP keep-alive) from clients that have died, or were "forgotten" by a proxy or load balancer which created them. Such dead connections can accumulate over time and consume server resources, and the new request timeout feature will remove them.

The request timeout can drop not only idle connections, but also connections where a client died in the middle of sending a request, and will never continue. This situation isn't common, but again if we don't clean up these connections, they can accumulate over time.

Finally, the request timeout can also help defend against slowloris DoS attacks - where an attacker deliberately sends a request extremely slowly. The request timeout can drop these connections even though they are not completely idle.

The set_keepalive_parameters() feature added recently can also solve some of these problems, but not all of them, and needs to work harder to verify a connection is really dead - whereas set_request_timeout() just assumes that a connection can be closed immediately if a request wasn't received for a long time.

The request timeout continues to default to 0, meaning it is disabled, so this patch doesn't change the http server's behavior for applications unless they choose to use the new feature.

Fixes #3130

### DIFF
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -37,6 +37,7 @@
 #include <seastar/http/routes.hh>
 #include <seastar/net/tls.hh>
 #include <seastar/core/shared_ptr.hh>
+#include <seastar/core/lowres_clock.hh>
 
 namespace seastar {
 
@@ -72,6 +73,8 @@ class connection : public boost::intrusive::list_base_hook<> {
     queue<std::unique_ptr<http::reply>> _replies { 10 };
     bool _done = false;
     const bool _tls;
+    // Timer to enforce set_request_timeout()
+    timer<lowres_clock> _request_timeout_timer;
 public:
     [[deprecated("use connection(http_server&, connected_socket&&, bool tls)")]]
     connection(http_server& server, connected_socket&& fd, socket_address, bool tls)
@@ -94,7 +97,8 @@ public:
             , _write_buf(_fd.output())
             , _client_addr(std::move(client_addr))
             , _server_addr(std::move(server_addr))
-            , _tls(tls) {
+            , _tls(tls)
+            , _request_timeout_timer([this] { _fd.shutdown_input(); }) {
         on_new_connection();
     }
     ~connection();
@@ -136,6 +140,7 @@ class http_server {
     bool _content_streaming = false;
     gate _task_gate;
     std::optional<net::keepalive_params> _keepalive_params;
+    std::chrono::milliseconds _request_timeout = 0ms;
 public:
     routes _routes;
     using connection = seastar::httpd::connection;
@@ -172,8 +177,43 @@ public:
     [[deprecated("use listen(socket_address addr, server_credentials_ptr credentials)")]]
     void set_tls_credentials(server_credentials_ptr credentials);
 
+    /**
+     * Enable or disable TCP keepalive on accepted connections. The kernel
+     * will periodically send keepalive probes on idle connections, and close
+     * connections that do not respond - with the given keepalive_params
+     * determining how often to probe and when to give up.
+     *
+     * Note that set_keepalive_parameters() and set_request_timeout() are
+     * different mechanisms that may be used to achieve similar goals, and
+     * it's not recommended to use both simultaneously.
+     */
     void set_keepalive_parameters(std::optional<net::keepalive_params> params) {
         _keepalive_params = std::move(params);
+    }
+
+    /**
+     * Enable closing a connection if reading the next request from it takes
+     * more than the given timeout duration. This feature will typically close
+     * idle connections (a.k.a. persistent connections or HTTP "keep alive")
+     * but can also close connections to clients that died in the middle of
+     * sending a request. Expiring such hung connections is important -
+     * otherwise defunct connections belonging to dead clients or forgotten
+     * load-balancer sessions can accumulate indefinitely, consuming growing
+     * server resources. A timeout on reading a request, whether or not it is
+     * making slow progress, is also a protection against "slowloris" attacks.
+     * 
+     * If the timeout is zero (this is the default), request timeout is
+     * disabled.
+     *
+     * A server which expects to handle low-bandwidth clients or very long
+     * requests should be careful not to set the request timeout too low.
+     *
+     * Note that set_keepalive_parameters() and set_request_timeout() are
+     * different mechanisms that may be used to achieve similar goals, and
+     * it's not recommended to use both simultaneously.
+     */
+    void set_request_timeout(std::chrono::milliseconds timeout) {
+        _request_timeout = timeout;
     }
 
     size_t get_content_length_limit() const;
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -98,6 +98,12 @@ future<> connection::do_response_loop() {
 }
 
 future<> connection::start_response() {
+    // While writing the response, disable the read timeout even though
+    // the server might already be reading the next request. If the connection
+    // dies we'll notice this when failing to write the response, and don't
+    // need to notice this via a read timeout. we'll re-arm the read timeout
+    // timer below - when we finish writing the response.
+    _request_timeout_timer.cancel();
     return _resp->write_reply(out()).then_wrapped([this] (auto f) {
         if (f.failed()) {
             // In case of an error during the write close the connection
@@ -130,6 +136,10 @@ future<> connection::start_response() {
             f.ignore_ready_future();
         }
         _resp.reset();
+        if (_server._request_timeout.count() > 0) {
+            _request_timeout_timer.cancel();
+            _request_timeout_timer.arm(_server._request_timeout);
+        }
         return make_ready_future<>();
     });
 }
@@ -208,7 +218,12 @@ void connection::generate_error_reply_and_close(std::unique_ptr<http::request> r
 
 future<> connection::read_one() {
     _parser.init();
+    if (_server._request_timeout.count() > 0) {
+        _request_timeout_timer.cancel();
+        _request_timeout_timer.arm(_server._request_timeout);
+    }
     return _read_buf.consume(_parser).then([this] () mutable {
+        _request_timeout_timer.cancel();
         if (_parser.eof()) {
             _done = true;
             return make_ready_future<>();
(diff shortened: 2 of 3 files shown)


nyh:
Pushed a new version with a fourth test with two more verifications (as usual, explained in comments).

Dear reviewers, if you're curious why I decided to put the timeout on receiving whole requests and not, for example, on a single read making progress, well - it started with an implementation problem: The problem was that the code uses consume() which takes care of the whole read loop for a single request and doesn't let me do things like reset the timer between each read. But once I made it into a timeout on the whole request I realized this is actually "good enough" for most "modern" applications, where you don't expect extremely slow 100 baud connections or extremely long request, so a (for example) 10 second timeout on receiving a request is not a problem. But beyond being good enough, it also has a benefit: It protects against attackers deliberately trickling long requests very slowly to consume a lot of server memory with little client effort ("slowloris" attack).

nyh:
@StephanDollberg @travisdowns this is a feature I propose can be used instead of TCP keepalive to clean up dead httpd connections.

StephanDollberg:
Yeah I think this is another layer which is good to have. 

> "good enough" for most "modern" applications, where you don't expect extremely slow 100 baud connections or extremely long request, so a (for example) 10 second timeout on receiving a request is not a problem. But beyond being good enough, it also has a benefit: It protects against attackers deliberately trickling long requests very slowly to consume a lot of server memory with little client effort ("slowloris" attack).

Getting a complete timeout solution can actually be quite tricky. Having one total timeout can be good for slowloris but if you have an http/file server which serves bigger files you need to account for slower clients as well which might only be downloading with XXX KB/s. With only one total timeout they'll in all likely hood always run into the timeout.

Inter-read timeouts can also be advantagous as they allow to keep idle connections open for longer which is good for keep-alive (which is always at odds with timeouts).

For full slowloris prevention you'd also need send timeouts as people could just receive very slowly on their and hence keep you busy for longer.

But anyway none of that makes the simple approach good to have in combination with TCP keepalives.

avikivity:
I think there's possibly two different timeouts here. One is for idle connections for which not a single byte was received. Another is for a connection which started receiving data, but is receiving it slowly.

Should we have two different timeouts here? Should we reset the timeout any time a byte is received?

nyh:
> Getting a complete timeout solution can actually be quite tricky. Having one total timeout can be good for slowloris but if you have an http/file server which serves bigger files you need to account for slower clients as well which might only be downloading with XXX KB/s. With only one total timeout they'll in all likely hood always run into the timeout.

This is why I clearly called the new feature "request_timeout" and applications where the total request time can be high - like when you have 100-baud clients or when requests can include large uploads - will not be able to use this new feature. I also wrote such a warning in the comment.

If you want a timeout on **individual reads** (not the complete request), it's actually not very trickyl - you can do exactly what I did but just restart the timer before every single read from the socket - not in the beginning of the request like I did. The only tricky part is that our current implementation uses a utility function `consume()` which itself calls `read()` in a loop and doesn't let us add anything to this loop. If we want to add such a feature, and call it, say `set_read_timeout()` (as an alternative to `set_query_timeout()`) we can actually do that. But since I didn't want to change consume() and don't care too much about the very-slow-request workloads , I didn't bother. Maybe I should have bothered.

> Inter-read timeouts can also be advantagous as they allow to keep idle connections open for longer which is good for keep-alive (which is always at odds with timeouts).

What is true that if the timeout is 30 seconds and you just spent 29.9 seconds with the connection being idle (keep-alive), now you only have 0.1 seconds to completely receive the request, and may close the connection in the middle. Of course, the chance that this will happen is unlikely, unless you badly tune the timeout to your workload. For example, imagine that you know that your workload often creates a connection and has a loop which sends a request every exactly 10 seconds - you wouldn't want to use query_timeout for exactly 10 seconds, this would be bad.
 
> For full slowloris prevention you'd also need send timeouts as people could just receive very slowly on their and hence keep you busy for longer.

This is, unfortunately, true.

With my patch, the attacker will not be able to force you to hold a very large (and very slowly growing) request in memory, but it might be able (depending on what the server is actually doing and whether it needs to generate large responses in memory) be able to force you to generate a very large response in memory and then read it very slowly.

> But anyway none of that makes the simple approach good to have in combination with TCP keepalives.

It makes the "simple approach" (I guess you mean my patch?) good to have, or not good to have? I couldn't get to the bottom of this sentence :-)

src/http/httpd.cc
```diff
@@ -130,6 +136,10 @@ future<> connection::start_response() {
             f.ignore_ready_future();
         }
         _resp.reset();
+        if (_server._request_timeout.count() > 0) {
+            _request_timeout_timer.cancel();
```
avikivity, line 140: Should you not cancel unconditionally? In case the timeout was changed.

I guess it doesn't matter - setting the request timeout doesn't have to synchronize with running requests.

StephanDollberg:
> It makes the "simple approach" (I guess you mean my patch?) good to have, or not good to have? I couldn't get to the bottom of this sentence :-)

Sorry! Yes I missed a few words there. I was trying to say that none of these points speak against having your patch :).

nyh:
> I think there's possibly two different timeouts here. One is for idle connections for which not a single byte was received. Another is for a connection which started receiving data, but is receiving it slowly.
> 
> Should we have two different timeouts here? Should we reset the timeout any time a byte is received?

Initially I wanted to be able to do that, but then I discovered we're using a function `consume()` which takes care of the read loop for us and doesn't let me do things between each read. So instead of hacking on consume() I went the "easy" way and added a timeout just for the entire request. Maybe I should indeed put more effort into this.

By the way, with the two different timeouts you suggested you won't be able to implement the timeout on the full query as I implemented now. For example, if you add just a 1-second timeout between bytes, you can still send a large request over a very long time (the "slowloris" attack). Moreover, you can't really use some very low number for this between-bytes timeout, because packet loss and network glitches - you don't really want to drop a request in the middle just because of a half-second network glitch. So I'm not sure what making the two timeouts have different values can buy us.

src/http/httpd.cc
nyh: Yes, I plan for changing the timeout to change already existing timers (the same connection will pick up the new timeout after the next request, but it doesn't matter what happens to the already-running read, I thought).

But perhaps the if() isn't helpful in the sense that cancel() is already very cheap to call if the timer wasn't set - it also just do a if(), so why bother with another if?

avikivity:
What if someone uploads large amounts of data?

nyh:
> What if someone uploads large amounts of data?

As I explained several times, including in the doxygen comment:
```
     * A server which expects to handle low-bandwidth clients or very long
     * requests should be careful not to set the request timeout too low.
```

So the request_timeout feature is not useful when the server application expects to serve good requests that really take an hour to send - e.g., a "upload of large amounts of data" - because you won't be able to set the request_timeout to less than hour. But I expect most server applications - including the ones that everyone in this discussion are writing - simply do not have this problem, and never expect any single request to be send over one second - so that timing out a request after, say, one minute, is good for them.

travisdowns:
> and never expect any single request to be send over one second

Well in Redpanda we do have prometheus endpoint requests which take 10-60 seconds for a single request (but that's no problem for this approach really as we'd probably set this timeout to minutes).

swasik:
@nyh please rebase and consider what should be the next step

include/seastar/http/httpd.hh
```diff
@@ -172,10 +177,45 @@ public:
...
+     * disabled.
+     *
+     * A server which expects to handle low-bandwidth clients or very long
+     * requests should be careful not to set the request timeout too low.
+     *
+     * Note that set_keepalive_parameters() and set_request_timeout() are
+     * different mechanisms that may be used to achieve similar goals, and
+     * it's not recommended to use both simultaneously.
+     */
+    void set_request_timeout(std::chrono::milliseconds timeout) {
```
travisdowns, line 215: Great comment overall, but I think this comment and the cover letter should explain when the timeout clock starts and stops. 

E.g., does the timer start at the end of the prior request? Or does it start at the beginning of the current request? I suppose it is the former.

travisdowns:
As I see it, the advantage of TCP-level timeouts is they actually detect a dead connection rather than simple lack of request from the remote side, so there is less tension in setting the parameters. E.g., if you want to support keep alive for up to 5 minutes, you need to set the http timeout to at least that amount. The TCP timeout has no such restriction since it does not require application traffic so it can be set lower.

The TCP timeout also works for any protocol, not just http (thought that's not relevant to this review).

The advantages of the http timeout as I see it:

You get a clear application level signal for the timeout, you can have a metric for it, etc. For a TCP timeout you just get broken pipe or whatever and have to make an educated guess at what happened.

You can define the behavior more exactly in terms of the protocol, e.g., from last request to current, etc. You can protect against malicious clients, etc, per the discussions above.

Agree with @StephanDollberg that they aren't mutually exclusive, it's fine to offer both, though having both enabled at once on the same server socket doesn't seem to make much sense.

PR #3152, opened 2025-12-11
Relax large allocation error originating from json_list_template

### LINKED ISSUE #3110: Calling `json_list_template& operator=(const C& list)` results in oversized allocation
Calling `json_list_template& operator=(const C& list)` results in oversized allocation if the list you want to assign from is big enough.
Given there's no option of a move assignment operator, we should either implement one or optimize the copy operator to avoid the large allocation.

Ref https://github.com/scylladb/scylladb/issues/26736

robertbindar:
This PR relaxes the cause of a large allocation errors like the one reported in https://github.com/scylladb/scylladb/issues/26736
The PR adds 2 patches, one that reserves the capacity of the container upfront within the copy assignment operator, and another that adds a move assignment operator for `json_list_template`.

Fixes #3110

### DIFF
--- a/include/seastar/json/json_elements.hh
+++ b/include/seastar/json/json_elements.hh
@@ -198,6 +198,15 @@ public:
         }
         return *this;
     }
+
+    template<class C>
+    json_list_template& operator=(json_list_template<T, Container>&& list) noexcept {
+        if (this != &list) {
+            _elements = std::move(list._elements);
+        }
+        return *this;
+    }
+
     virtual future<> write(output_stream<char>& s) const override {
         return formatter::write(s, _elements);
     }


include/seastar/json/json_elements.hh
```diff
@@ -192,6 +192,7 @@ public:
     template<class C>
     json_list_template& operator=(const C& list) {
         _elements.clear();
+        _elements.reserve(std::size(list));
```
xemul, line 195: How does reservation help, other than in weird corner cases when list size is `2^n + k (k is "small enough")` and it's below large allocation threshold, but  the last push causes `2^(n+1)` allocation which suddenly goes above it?

robertbindar: that's indeed the only case when this helps, it's a slim optimization, the actual fix for the large allocation error is in the other commit which adds the move constructor.

robertbindar: @xemul anything else I should do here?

xemul: I don't understand why it _fixes_ large allocation.

This reservation improves the re-allocations problem for adding a large bunch of elements to the list, but only _fixes_ the large allocation withing some very restrictive bounds.
The second patch in this set also improves things by not _re_allocating memory for move-assignment, but since move-from list should've been allocated somehow, it's still large allocation somewhere.

So from my current understanding the PR is "relaxes" large allocation problem, but not fixes it.

bhalevy: I agree with @xemul.
And the added move assignment also helps to mitigate allocations.
The fix at the higher level is to use e.g. a chunked vector as the container type.

bhalevy:
Can we please advance this PR?

xemul:
> Can we please advance this PR?

@robertbindar , please rephrase the PR and patches' comments according to [this](https://github.com/scylladb/seastar/pull/3152#discussion_r2622239466) discussion

robertbindar:
> > Can we please advance this PR?
> 
> @robertbindar , please rephrase the PR and patches' comments according to [this](https://github.com/scylladb/seastar/pull/3152#discussion_r2622239466) discussion

Hey, sorry, I context-switched very far from this area, I'll try to get to it this week.

xemul:
Just a light reminder ping to @robertbindar

robertbindar:
@xemul descriptions updated, thanks and sorry for the wait.

PR #3161, opened 2025-12-15
resource: Hide hwloc from public interface

travisdowns:
Forward-delcare hwloc pointer hwloc_topology_t type instead of including hwloc.h in the public interface. This reduces compile times both by trimming the inclusion tree but also by not requiring that hwloc be built before seastar can build (which happens when hwloc is a black-box "foreign cc" compile in bazel).

The main benefit though is not injecting the hwloc header into seastar's public interface.

### DIFF
--- a/include/seastar/core/resource.hh
+++ b/include/seastar/core/resource.hh
@@ -32,9 +32,9 @@
 #include <set>
 #include <sched.h>
 #include <unordered_map>
-#ifdef SEASTAR_HAVE_HWLOC
-#include <hwloc.h>
-#endif
+
+struct hwloc_topology;
+using hwloc_topology_t = hwloc_topology*;
 
 namespace seastar {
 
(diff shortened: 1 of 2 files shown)


rockwotj (approved):

include/seastar/core/resource.hh
```diff
@@ -32,9 +32,9 @@
 #include <set>
 #include <sched.h>
 #include <unordered_map>
-#ifdef SEASTAR_HAVE_HWLOC
-#include <hwloc.h>
-#endif
+
+struct hwloc_topology;
+using hwloc_topology_t = hwloc_topology*;
```
avikivity, line 37: It's a bit risky to forward-declare something outside our control, but I expect ABI compatibility lowers the risk.

travisdowns: > It's a bit risky to forward-declare something outside our control, but I expect ABI compatibility lowers the risk.

Yes, though if it becomes "wrong" we will get a compile error due to conflicting declarations as we do include hwloc.h in the .cc file where both the forward declaration and true declaration will appear.

PR #3162, opened 2025-12-16
file: experimental_list_directory: use buffered generator

bhalevy:
Now, that the following issues are fixed, we can safely use a buffered generator for listing directories.
scylladb/seastar#1913
scylladb/seastar#1677

### DIFF
--- a/include/seastar/core/circular_buffer_fixed_capacity.hh
+++ b/include/seastar/core/circular_buffer_fixed_capacity.hh
@@ -35,6 +35,8 @@
 #include <utility>
 #include <memory>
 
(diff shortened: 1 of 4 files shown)


bhalevy:
[7494359](https://github.com/scylladb/seastar/commit/7494359328ad233480cc9a65d02a58d3f96b7213): fixed capacity calculation

bhalevy:
[0046c83](https://github.com/scylladb/seastar/commit/0046c83f652945f0060bf0db9383481623889cf8):
- defined list_directory_generator_buffer_size (as 16 entries)
  - used also in make_list_directory_fallback_generator

PR #3163, opened 2025-12-16
[RFC] file: add experimental_list_directory_ext generator

bhalevy:
This series adds file::statat methods to get stat data using relative paths in an open directory.
On top of that,  `experimental_list_directory_ext` generator is added that yields
an extended directory entry holding the entry name as well as its stat data.

It optimizes this process by using `statat` for every listed directory entry using
the open directory file descriptor which saves the lookup of all leading path components
and it reuses the directory entries recently read from the directory.

### DIFF
--- a/include/seastar/core/file.hh
+++ b/include/seastar/core/file.hh
@@ -121,6 +121,19 @@ constexpr size_t list_directory_generator_buffer_size = calc_circular_buffer_cap
 using list_directory_generator_type = coroutine::experimental::generator<directory_entry, directory_entry,
         circular_buffer_fixed_capacity<directory_entry, list_directory_generator_buffer_size>>;
 
+struct directory_entry_ext : public stat_data {
+    /// Name of the file in a directory entry.  Will never be "." or "..".  Only the last component is included.
+    sstring name;
+
+    directory_entry_ext(stat_data sd, sstring n)
+        : stat_data(std::move(sd))
+        , name(std::move(n))
+    {}
+};
+constexpr size_t list_directory_ext_generator_buffer_size = calc_circular_buffer_capacity<directory_entry_ext, 512>();
+using list_directory_ext_generator_type = coroutine::experimental::generator<directory_entry_ext, directory_entry_ext,
+        circular_buffer_fixed_capacity<directory_entry_ext, list_directory_ext_generator_buffer_size>>;
+
 // A handle that can be transported across shards and used to
 // create a dup(2)-like `file` object referring to the same underlying file
 class file_handle_impl {
@@ -164,6 +177,7 @@ public:
     virtual std::unique_ptr<file_handle_impl> dup();
     virtual subscription<directory_entry> list_directory(std::function<future<> (directory_entry de)> next) = 0;
     virtual list_directory_generator_type experimental_list_directory();
+    virtual list_directory_ext_generator_type experimental_list_directory_ext();
 };
 
 future<shared_ptr<file_impl>> make_file_impl(int fd, file_open_options options, int oflags, struct stat st) noexcept;
@@ -525,6 +539,9 @@ public:
 
     /// Returns a directory listing, given that this file object is a directory.
     list_directory_generator_type experimental_list_directory();
+
+    /// Returns a directory listing that includes the stat_data for each entry, given that this file object is a directory.
+    list_directory_ext_generator_type experimental_list_directory_ext();
 
     /**
      * Read a data bulk containing the provided addresses range that starts at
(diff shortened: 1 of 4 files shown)


bhalevy:
Draft until #3162

bhalevy:
[a4dc75c](https://github.com/scylladb/seastar/commit/a4dc75c4d23969aa590641c4b5b0b91d589c6960):
- skip entries that were removed between list_directory callback to fstatat() call.
- defined and use list_directory_ext_generator_buffer_size

include/seastar/core/file.hh
```diff
@@ -86,6 +86,8 @@ struct stat_data {
     std::chrono::system_clock::time_point time_changed;   // Time of last status change (either content or attributes)
 };
 
+stat_data make_stat_data(const struct stat& st);
```
xemul, line 89: I'd keep in out of seastar public API. If it has to be present in file.hh for some reason, then `namespace internal` can be used

src/core/file.cc
```diff
@@ -440,6 +441,37 @@ list_directory_generator_type posix_file_impl::experimental_list_directory() {
...
+
+        for (const char* b = buf.get(); b < buf.get() + size; ) {
+            const auto de = reinterpret_cast<const linux_dirent64*>(b);
+            b += de->d_reclen;
+            sstring name(de->d_name);
+            if (name == "." || name == "..") {
+                continue;
+            }
+            try {
+                directory_entry_ext ret(make_stat_data(co_await statat(name, AT_SYMLINK_NOFOLLOW)), std::move(name));
```
xemul, line 461: If it's calling `file::statat()` anwyay, why can't the user of experimental_list_directory do it on his own? This new API saves one line of user code at the cost of the whole new API method.

bhalevy: Unlike `list_directory`, With the generator, the caller doesn't open the listed directory, this state is private to the generator, so the caller can't call statat on it.

bhalevy: Also, it is inspired by NFS READDIRPLUS :-)

include/seastar/core/file.hh
```diff
@@ -121,6 +121,19 @@ constexpr size_t list_directory_generator_buffer_size = calc_circular_buffer_cap
 using list_directory_generator_type = coroutine::experimental::generator<directory_entry, directory_entry,
         circular_buffer_fixed_capacity<directory_entry, list_directory_generator_buffer_size>>;
```
avikivity, line 122: > file: add experimental_list_directory_ext generator
> Generates struct directory_ent_ext for each entry
> that includes the entry name and stat data.
> 

Why can't the user add a stat call after obtaining a regular entry?

src/core/file.cc
bhalevy: I'm confused, sorry.
experimental_list_directory_ext is a method of `file`, so the caller does have the directory opened already and it can indeed call statat()

include/seastar/core/file.hh
bhalevy: Hmm, I was confused.  The caller should be able to easily call statat(), so we don't really need `experimental_list_directory_ext`.  I'll reduce the scope of this PR only to `statat()` then.

bhalevy:
Replaced by https://github.com/scylladb/seastar/pull/3174 which contains only `statat()`

avikivity:
Exercise: suppose we're running over a slow disk (or network like your READDIRPLUS). Suppose further that stat/statat is implemented over io_uring so it can run concurrently. How does one use the generator + stat to list+stat files quickly?

```c++
generator<name_and_stat>
list_and_stat(dir) {
    auto names = list_directory(dir);
    while (auto name_opt = co_await names()) {
          ... some magic ...
          co_yield name_and_stat(*name_opt, stat);
}
```

where "some magic" is somewhat similar to max_concurrent_for_each - it pushes name_opt into a queue for statting, then co_awaits the other side for the stat, but in between there is concurrency. Can Seastar solve the problem elegantly?

avikivity:
```c++
class magic {
    std::queue<sstring> _not_yet_statted_names;
    std::queue<future<struct stat>> _return_queue;
    void push(sstring name) {
        _not_yet_statted_names.push(name);
        refill();
    }
    void refill() {
        while (!_not_yet_statted.empty() && _return_queue.size() < max) {
            _return_queue.push(stat(_not_yet_statted.front()));
            _not_yet_statted.pop_front();
        }
    }
    future<struct stat> pop() {
        auto ret = co_await  _return_queue.front();
        _return_queue.pop_front();
        refill();
        co_return ret;
    }
}
```

Maybe not elegant, but serviceable.

bhalevy:
@avikivity don't we get the queue and parallel execution with the [buffered generator](https://github.com/scylladb/seastar/pull/3162)?

avikivity:
We get a queue, but not parallel execution.

PR #3166, opened 2025-12-16
seastar::test: Add exception interception to extract nested info

### LINKED ISSUE #3165: Boost unit test runner does not report nested c++ exceptions
This is not a seastar problem as such, but a weakness in the current boost::test. Any C++ exception failure in a test will only report the type and message of the outer most exception, which in some cases, where nesting is used to package the info, means in the end we have no idea why something failed. 

Since seastar to a large extent augment the boost::test framework already, it is not a huge challenge to simply wrap the invocation of the actual test method with a better version of the exception catching normally done in [execution_monitor.ipp](https://github.com/boostorg/test/blob/boost-1.90.0/include/boost/test/impl/execution_monitor.ipp) added at the invocation point. We can then do similar unwinding of nested-ness as is done in logging etc.

elcallio:
Fixes #3165

While not a seastar problem as such, but a weakness in the current boost::test, any C++ exception failure in a test will only report the type and message of the outer most exception, which in some cases, where nesting is used to package the info, means in the end we have no idea why something failed.

This patch adds exception interception at the invocation of the test case, normally done in execution_monitor.ipp, and similar to this, creates exception info to be propagated to the test logging. However, unlike the boost impl, we recurse the nested exceptions and add these types and messages as well.

Small unit test included.

Note: the actual (XML) logging of tests is limited to one type for an exception causing failure, so we are limited to tracking additional info in the "message" portion.

Note: similar to the original code in boost::test, this code avoid buffer allocs to at least some degree (which can be argued if it really helps any).

### DIFF
--- a/src/testing/seastar_test.cc
+++ b/src/testing/seastar_test.cc
@@ -23,6 +23,10 @@
 #include <thread>
 #include <iostream>
 
+#include <boost/test/execution_monitor.hpp>
+#include <boost/exception/diagnostic_information.hpp>
+#include <boost/core/type_name.hpp>
+
 #include <seastar/testing/entry_point.hh>
 #include <seastar/testing/seastar_test.hh>
 #include <seastar/testing/test_fixture.hh>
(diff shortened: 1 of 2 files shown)


src/testing/seastar_test.cc
```diff
@@ -39,6 +43,54 @@ namespace testing {
...
+[[noreturn]]
+static void repackage_exception(const std::exception* ep) {
+    // Note: using a static buffer for formatting, same as boost::test code,
+    // so we make it less prone to fail in failure handling for OOM
+    // situations etc.
+    static const int REPORT_ERROR_BUFFER_SIZE = 4096;
+    static char buf[REPORT_ERROR_BUFFER_SIZE];
+
+    auto loc = add_exception_message(ep, false, buf, buf + sizeof(buf) - 1);
+    boost:: BOOST_TEST_I_THROW(boost::execution_exception(boost::execution_exception::cpp_exception_error, buf, loc));
```
xemul, line 91: why not just re-throw repackaged exception? (I don't propose to do it, just wondering)

src/testing/seastar_test.cc
```diff
@@ -39,6 +43,54 @@ namespace testing {
...
+        } catch (...) {
+            add_exception_message(nullptr, true, out, end);
+        }
+    }
+
+    return loc;
+}
+
+[[noreturn]]
+static void repackage_exception(const std::exception* ep) {
```
xemul: nit: `repackage_exception_and_report()` or `..._and_rethrow()` would be more descriptive. Looking at the caller code makes an impression that the excepiton is just swallowed.

xemul (approved):

src/testing/seastar_test.cc
elcallio: I'm mimicking the boost test framework code as close as possible here. So thus the `BOOST_TEST_I_THROW`.

src/testing/seastar_test.cc
elcallio: Mimicked naming from boost test again. But in this case, verbosity is probably more helpful. :-) Changed.

denesb:
@xemul ping for merge.

PR #3168, opened 2025-12-17
Revert to ~old lifetime semantics for lvalues passed to then()-alikes

### LINKED ISSUE #3157: QUESTION: lifetime of lvalues passed to future::then
We are seeing some UAF/SUAR crashes in Redpanda after upgrading to Seastar. 

I believe the cause is: https://github.com/scylladb/seastar/commit/1f233b4ebe6c943e5b8e60cee4994efeba7b2f6e by @avikivity .

As I understand it, this changed the lifetime of lvalues `func` passed to `f.then(func)` for future `f`. We consider only the case where `f` is not ready, since in the ready case `func` is simply executed synchronously in any case. Previously lvalues passed to unready futures would ultimately be copied into the continuation (see `then_impl_nrvo` and `void schedule(Pr&& pr, Func&& func, Wrapper&& wrapper)`), but now they are wrapped in `std::ref`, so no copy happens.

The exact lifetime semantics of `then()` for all inputs was never clear to me, so I suppose this is possibly within the scope of the spec (the rule could be "the lvalue must live until at least until the returned future resolves"). Note that this places a larger burden on the caller than rvalues, which are guaranteed to be moved in and have their lifetime extended naturally (though this split is common in other places in seastar as well).

My main concern is this is now a lifetime "timebomb", because even if you have test coverage of all these cases, with ASAN on, the issue doesn't arise until the parent future is unready, and even the always-preempt thing in debug mode that is often not the case (then() doesn't preempt check) in tests so it's hard to effectively test this.

travisdowns:
Commit https://github.com/scylladb/seastar/commit/1f233b4ebe6c943e5b8e60cee4994efeba7b2f6e changed the lifetime semantics for continuations passed
as lvalues to future methods (then, then_wrapped, finally, etc).
Previously, the continuation was moved or copied (itself an
inconsistency as the copy vs move behavior depended on the "type erase
more" configuration and moving from an lvalue is very surprising).

This effectively extended the lifetime of the continuation until it was
run. After that commit, the continuation is wrapped in std::ref, which
means no copy or move is performed and the relevant continuation is
the original lvalue passed in, which often dies quickly if it is e.g.,
from the stack of the function where then() was called.

The second patch in this series effectively reverts to the old behavior, this time by
copying lvalues instead of wrapping them in std::ref. This means that
lvalues are converted into rvalues (via copy) which preserves the
intent of the prior change which was to deal only with rvalues in the
downstream implementation, but also extends the lifetime.

This requires the continuation type to be copyable, but that was already
the case before https://github.com/scylladb/seastar/commit/1f233b4ebe6c943e5b8e60cee4994efeba7b2f6e. The introduction of a copy is undesirable
for performance, but lvalue use is quite rare and I don't see any way
to preserve all of the following:

 - Allow lvalues to be passed
 - Do not move out of the passed lvalue (bug, surprising)
 - Extend the lifetime as before

In the future we may simply disable lvalue support entirely, per the
discussion in https://github.com/scylladb/seastar/issues/3157.

This change also adds various test_lifetimes_and_movement tests to
verify continuation handling semantics. This counts the number of copies
and moves to catch cases where the behavior changes (the existing counts
are not somehow sacred, so the tests can be updated after checking if
this was intended). It also checks explicitly this case of lvalue
lifetime extension and fails before the changes in this patch. It checks
a few other attributes too, e.g., that we don't call a moved from
lambda, etc.

It tests more than just then(), also then_wrapped() and finally(), each
in two variations as internally the implementation as variations for
when then_wrapped takes future&& or future by value, and when finally
returns a future or not. This turned up additional issues in those
methods which didn't select the operator() && overload, which are fixed
in the first change in this series.

Test details:

The test_lifetimes_and_movement<Traits>() template function provides a
trait-based framework for testing different future continuation methods
with consistent test cases. It is instantiated with multiple trait types
to cover:
- then() - standard continuation chaining
- then_wrapped() - continuations that receive the future itself
- then_wrapped() with future&& - reference-based variant
- finally() - cleanup handlers returning futures
- finally() returning void - non-future cleanup handlers

Each instantiation runs 6 test cases:
1. Passing move-only rvalue callable - verifies zero-copy rvalue
   forwarding
2. Passing lvalue callable - verifies lvalue is copied (not moved)
3. Passing std::move(lvalue) - verifies explicit move semantics
4. Ready future with rvalue callable - tests fast-path optimization
5. Ready future with lvalue callable - tests fast-path with copies
6. Local lvalue lifetime extension - ensures continuations don't use
   after destruction (addresses
   https://github.com/scylladb/seastar/issues/3157)

Test infrastructure includes:
- canary class with moved_from/alive/destroyed state tracking
- canary_stats to count copies, moves, assignments, and operator() calls
- canary_callable_base template supporting both copyable and move-only
  variants
- BOOST_TEST_CHECKPOINT markers for each test case to aid debugging

The tests verify:
- Exact counts of copy/move operations (with conditional checks for
  SEASTAR_TYPE_ERASE_MORE and SEASTAR_DEBUG configurations)
- Correct operator() overload selection (lvalue vs rvalue-qualified)
- Original lvalues remain valid after passing to continuations
- Moved-from objects are properly marked and not used
- Objects are not accessed after destruction

This test suite serves as a regression test for continuation handling
and documents the expected behavior of future continuation APIs.

Fixes #3157.

### DIFF
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -1497,7 +1497,7 @@ private:
         {
             memory::scoped_critical_alloc_section _;
             ncf = noncopyable_function<WrapFuncResult(future &&)>([func = std::forward<Func>(func)](future&& f) mutable {
-                return futurator::invoke(func, std::move(f));
+                return futurator::invoke(std::move(func), std::move(f));
             });
         }
         return then_wrapped_common<AsSelf, WrapFuncResult>(std::move(ncf));
@@ -1513,7 +1513,7 @@ private:
         using pr_type = decltype(fut.get_promise());
         schedule(fut.get_promise(), std::move(func), [](pr_type&& pr, Func&& func, future_state&& state) {
             futurator::satisfy_with_result_of(std::move(pr), [&func, &state] {
-                return func(future(std::move(state)));
+                return std::move(func)(future(std::move(state)));
             });
         });
         return fut;
@@ -1609,8 +1609,8 @@ public:
         finally_body(Func&& func) noexcept : _func(std::forward<Func>(func))
         { }
 
-        future<T> operator()(future<T>&& result) noexcept {
-            return futurize_invoke(_func).then_wrapped([result = std::move(result)](auto&& f_res) mutable {
+        future<T> operator()(future<T>&& result) && noexcept {
+            return futurize_invoke(std::move(_func)).then_wrapped([result = std::move(result)](auto&& f_res) mutable {
                 if (!f_res.failed()) {
                     return std::move(result);
                 } else {
@@ -1627,9 +1627,9 @@ public:
         finally_body(Func&& func) noexcept : _func(std::forward<Func>(func))
         { }
 
-        future<T> operator()(future<T>&& result) noexcept {
+        future<T> operator()(future<T>&& result) && noexcept {
             try {
-                _func();
+                std::move(_func)();
                 return std::move(result);
             } catch (...) {
                 return result.rethrow_with_nested();
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -358,6 +358,25 @@ static constexpr bool is_tuple_effectively_trivially_move_constructible_and_dest
 
 }
 
+// Helper to convert lvalue reference functions to rvalue.
+// This is used early in then()-alike entry points to convert lvalue references into
+// rvalues (which are used to recursively call back into the same entry point) to
+// avoid having to handle both lvalues and rvalues downstream: we convert them all
+// to rvalues. This helper handles both references to callable objects such as lambdas,
+// copying them, and references to functions, decaying them to function pointers.
+template <typename Func>
+static constexpr auto func_to_rvalue(Func&& func) {
+    static_assert(std::is_lvalue_reference_v<Func>, "call with lvalue reference");
+    if constexpr (std::is_function_v<std::remove_reference_t<Func>>) {
+        // return decayed function pointer
+        return +func;
+    } else {
+        // return a copy (used as an rvalue by the called)
+        return std::forward<Func>(func);
+    }
+}
+
+
 //
 // A future/promise pair maintain one logical value (a future_state).
 // There are up to three places that can store it, but only one is
@@ -1141,6 +1160,15 @@ struct result_of_apply<Func, std::tuple<T...>> : std::invoke_result<Func, T...> 
 template <typename Func, typename... T>
 using result_of_apply_t = typename result_of_apply<Func, T...>::type;
 
+// To simply implementation, at various top-level (user-callable) entry-points
+// like then(), finally(), handle_exception(), etc, we convert lvalue callables
+// into rvalues, using a recursive call back to that entry point, so that all
+// the downstream implementation functions can deal with rvalues only. Inside
+// those functions we then assert on this trait to ensure we do only have rvalues
+// since we can't match rvalue references directly (because that syntax was
+// co-opted for universal references).
+template <typename Func>
+constexpr bool expect_only_rvalue_refs = !std::is_lvalue_reference_v<Func>;
 }
 
 template <typename Promise, typename T>
@@ -1226,6 +1254,7 @@ private:
     }
     template <typename Pr, typename Func, typename Wrapper>
     void schedule(Pr&& pr, Func&& func, Wrapper&& wrapper) noexcept {
+        static_assert(internal::expect_only_rvalue_refs<Func>);
         // If this new throws a std::bad_alloc there is nothing that
         // can be done about it. The corresponding future is not ready
         // and we cannot break the chain. Since this function is
@@ -1341,6 +1370,11 @@ public:
     /// If the future failed, the function is not called, and the exception
     /// is propagated into the return value of then().
     ///
+    /// The passed function is moved (if an rvalue) or copied (if an lvalue) into
+    /// the continuation, so their lifetime is automatically extended until the
+    /// continuation runs (but not further - if the function itself may suspend
+    /// you will need to ensure its lifetime is sufficiently long).
+    ///
     /// \param func - function to be called when the future becomes available,
     ///               unless it has failed.
     /// \return a \c future representing the return value of \c func, applied
@@ -1353,7 +1387,7 @@ public:
       // Avoid having to special-case lvalue-references downstream by converting
       // them to an rvalue reference here.
       if constexpr (std::is_lvalue_reference_v<Func>) {
-        return then(std::ref(func));
+        return then(func_to_rvalue(func));
       } else {
 #ifndef SEASTAR_TYPE_ERASE_MORE
         return then_impl(std::move(func));
@@ -1394,10 +1428,8 @@ public:
     requires ::seastar::CanApplyTuple<Func, T>
     Result
     then_unpack(Func&& func) noexcept {
-      // Avoid having to special-case lvalue-references downstream by converting
-      // them to an rvalue reference here.
       if constexpr (std::is_lvalue_reference_v<Func>) {
-        return then_unpack(std::ref(func));
+        return then_unpack(func_to_rvalue(func));
       } else {
         return then([func = std::forward<Func>(func)] (T&& tuple) mutable {
             // sizeof...(tuple) is required to be 1
@@ -1432,6 +1464,7 @@ private:
     template <typename Func, typename Result = futurize_t<internal::future_result_t<Func, T>>>
     Result
     then_impl(Func&& func) noexcept {
+        static_assert(internal::expect_only_rvalue_refs<Func>);
 #ifndef SEASTAR_DEBUG
         using futurator = futurize<internal::future_result_t<Func, T>>;
         if (failed()) {
@@ -1465,7 +1498,7 @@ public:
       // Avoid having to special-case lvalue-references downstream by converting
       // them to an rvalue reference here.
       if constexpr (std::is_lvalue_reference_v<Func>) {
-        return then_wrapped(std::ref(func));
+        return then_wrapped(func_to_rvalue(func));
       } else {
         return then_wrapped_maybe_erase<false, FuncResult>(std::forward<Func>(func));
       }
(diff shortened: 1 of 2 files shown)


include/seastar/core/future.hh
```diff
@@ -358,6 +358,25 @@ static constexpr bool is_tuple_effectively_trivially_move_constructible_and_dest
...
+// rvalues (which are used to recursively call back into the same entry point) to
+// avoid having to handle both lvalues and rvalues downstream: we convert them all
+// to rvalues. This helper handles both references to callable objects such as lambdas,
+// copying them, and references to functions, decaying them to function pointers.
+template <typename Func>
+static constexpr auto func_to_rvalue(Func&& func) {
+    static_assert(std::is_lvalue_reference_v<Func>, "call with lvalue reference");
+    if constexpr (std::is_function_v<std::remove_reference_t<Func>>) {
+        // return decayed function pointer
+        return +func;
```
xemul, line 372: What does this unary `+` mean here? :exploding_head:

rockwotj: It's an old trick to convert to a pointer. Basically lambdas get implicitly converted to a function pointer

travisdowns: I could have also use `std::decay_t` here but I left it as `+` as `+` only works on pointers and a few other things so more likely to catch unexpected cases at compile time (decay works on everything).

include/seastar/core/future.hh
```diff
@@ -1497,7 +1497,7 @@ private:
         {
             memory::scoped_critical_alloc_section _;
             ncf = noncopyable_function<WrapFuncResult(future &&)>([func = std::forward<Func>(func)](future&& f) mutable {
-                return futurator::invoke(func, std::move(f));
+                return futurator::invoke(std::move(func), std::move(f));
```
avikivity, line 1534: Is it not more correct to std::forward here?

include/seastar/core/future.hh
```diff
@@ -1513,7 +1513,7 @@ private:
         using pr_type = decltype(fut.get_promise());
         schedule(fut.get_promise(), std::move(func), [](pr_type&& pr, Func&& func, future_state&& state) {
             futurator::satisfy_with_result_of(std::move(pr), [&func, &state] {
-                return func(future(std::move(state)));
+                return std::move(func)(future(std::move(state)));
```
avikivity, line 1550: And here?

include/seastar/core/future.hh
```diff
@@ -358,6 +358,25 @@ static constexpr bool is_tuple_effectively_trivially_move_constructible_and_dest
...
+static constexpr auto func_to_rvalue(Func&& func) {
+    static_assert(std::is_lvalue_reference_v<Func>, "call with lvalue reference");
+    if constexpr (std::is_function_v<std::remove_reference_t<Func>>) {
+        // return decayed function pointer
+        return +func;
+    } else {
+        // return a copy (used as an rvalue by the called)
+        return std::forward<Func>(func);
+    }
+}
```
avikivity, line 377: The semantics are very confusing (before and after the patch).

Proposal for the future: delete the lvalue-accepting overloads. If someone wants an lvalue, they can wrap it in std::ref(), or copy it, both are explicit. Requires an API_LEVEL.

travisdowns: > Proposal for the future: delete the lvalue-accepting overloads. 

Agree.

Though I don't actually think the semantics are confusing after the patch? The lifetime is _always_ extended, in all build modes, and with lvalues and rvalues. lvalues are copied and rvalues are moved but you can also just think of this as "always copy" since move vs copy is largely invisible for rvalues.

The main reason to change it later, along the lines you suggest, is because the copy is inefficient.

include/seastar/core/future.hh
travisdowns: I don't think so: `func` is always an rvalue reference here (this was the goal of your original change), it's just that C++ (very annoyingly) doesn't have a way to express that parameter type so it looks like a universal reference - but it isn't supposed to be here. This is aligned with your changes I think.

I could add a static assert that `func` is not an lvalue maybe?

include/seastar/core/future.hh
travisdowns: See above, let us continue the discussion there.

include/seastar/core/future.hh
avikivity: It's confusing because universal-reference parameters are expected to be passed by reference, not by value. It's less confusing than the intermediate state.

include/seastar/core/future.hh
avikivity: From a big-picture perspective, yes, it's always an rvalue. But it's not nice to make understanding of the function require this knowledge.

A static_assert (or constraint) would make it clear in the function scope that the move is fine.

include/seastar/core/future.hh
travisdowns: > It's confusing because universal-reference parameters are expected to be passed by reference, not by value. It's less confusing than the intermediate state.

Well it _is_ passed by reference at the initial call, though it is ultimately copied at some point inside the implementation. This doesn't strike me as _that_ unusual for reference parameters, it happens frequently e.g., in constructors which take inputs by reference and then copy or move values into the object. 

A "normal" function which doesn't capture any of its values and doesn't have the lifetime issues of a continuation sure, you wouldn't expect a copy because that's inefficient, but these functions do have lifetime concerns. In fact, I'd say that universal references _often_ a copy will happen for lvalues: the main reason you use them is to be able to distinguish between rvalues and lvalues, and you want to do that so you can move the rvalues, and if you need to move in the first place it implies there is some "capturing" going on which often implies a copy in the lvalue case. If you didn't need to move/copy, you could choose choose a `const T&` which would bind to rvalues and lvalues and you can pass it along as a reference everywhere.

Longer term, I see three valid points in the design space:

1) Approach in this patch: always copy/move lvalues/rvalues. Least lifetime issues, least efficient for lvalues. Avi thinks it's confusing semantics, Travis does not. Requires that lvalues be copyable.

2) Approach immediatley pre-patch: std::ref() for lvalues, move rvalues. This also seems valid, it's just that it's a breaking change. Efficient for lvalues, lvalues don't need to be copyable. Lifetime footgun for lvalues though, and it seems risk to introduce this even if hidden behind API_LEVEL because it's not something that will be detected at compile time, it requires a careful audit.

3) Ban lvalues. Lets the user choose between (1) and (2) effectively: make a copy at the call site for behavior (1) or use std::ref() for behaviour (2).

I think you favor (3) and I tend to agree. It is fully flexible, so the downsides of (1) and (2) can be avoided on a per-call basis. The only real downside is a bit more thinking/boilerplate at each call site (and I think the "thinking" part of that is _good_).

travisdowns: ... and it would be great to see the chosen approach applied widely, in all the then()-alike's but also functions like submit_to, etc.

include/seastar/core/future.hh
travisdowns: ack, will add `static_assert`

travisdowns: Added the static assert in last push, along with a comment explaining what is going on:

https://github.com/scylladb/seastar/compare/9d0eed6b9b5c24226a41f1d9c91dbe35db8724d6..efe5d7521d86e2ec924d360f29b1597a1e5917ad

include/seastar/core/future.hh
avikivity: > > It's confusing because universal-reference parameters are expected to be passed by reference, not by value. It's less confusing than the intermediate state.
> 
> Well it _is_ passed by reference at the initial call, though it is ultimately copied at some point inside the implementation. This doesn't strike me as _that_ unusual for reference parameters, it happens frequently e.g., in constructors which take inputs by reference and then copy or move values into the object.
> 
> A "normal" function which doesn't capture any of its values and doesn't have the lifetime issues of a continuation sure, you wouldn't expect a copy because that's inefficient, but these functions do have lifetime concerns. In fact, I'd say that universal references _often_ a copy will happen for lvalues: the main reason you use them is to be able to distinguish between rvalues and lvalues, and you want to do that so you can move the rvalues, and if you need to move in the first place it implies there is some "capturing" going on which often implies a copy in the lvalue case. If you didn't need to move/copy, you could choose choose a `const T&` which would bind to rvalues and lvalues and you can pass it along as a reference everywhere.


When I write something that copies lvalues, I mark the argument as pass-by-value. If the actual argument is an rvalue it will be moved into the parameter, if it's an lvalue it will be copied.

> 
> Longer term, I see three valid points in the design space:
> 
> 1. Approach in this patch: always copy/move lvalues/rvalues. Least lifetime issues, least efficient for lvalues. Avi thinks it's confusing semantics, Travis does not. Requires that lvalues be copyable.



> 2. Approach immediatley pre-patch: std::ref() for lvalues, move rvalues. This also seems valid, it's just that it's a breaking change. Efficient for lvalues, lvalues don't need to be copyable. Lifetime footgun for lvalues though, and it seems risk to introduce this even if hidden behind API_LEVEL because it's not something that will be detected at compile time, it requires a careful audit.



> 3. Ban lvalues. Lets the user choose between (1) and (2) effectively: make a copy at the call site for behavior (1) or use std::ref() for behaviour (2).
> 
> I think you favor (3) and I tend to agree. It is fully flexible, so the downsides of (1) and (2) can be avoided on a per-call basis. The only real downside is a bit more thinking/boilerplate at each call site (and I think the "thinking" part of that is _good_).

Yes, I intended 3. If lvalue and rvalue calls were evenly distributed, then it would make sense to think it through. But the ratio is around 100:0, so it makes sense to force the rare case to think it through and to annotate its decision.

avikivity: > ... and it would be great to see the chosen approach applied widely, in all the then()-alike's but also functions like submit_to, etc.

Yes yes, also invoke_on.

Seastar is very move intensive and after a while that's how you think of things.

travisdowns: > When I write something that copies lvalues, I mark the argument as pass-by-value. If the actual argument is an rvalue it will be moved into the parameter, if it's an lvalue it will be copied.

Well sure that's the other, simpler way with less boilerplate, but it has a lot of downsides: there are some places you just can't use it, but also even when you can it is inefficient since it's eagerly doing a copy/move (to construct the parameter), then another move after: passing by reference/perfect forwarding does exactly 1 final copy/move. That's for 0 intermediate layers: the inefficiency gets worse the more layers you move/copy through using that idiom.

avikivity: That's a good point.

Option 1: create a new marker function/type and recognize it
Option 2: leave things as is, recognize that its yet another footgun in the well stocked arsenal of C++ weaponry

travisdowns: I don't think there's anything to do here at the current API level? I thought this discussion was just an offshoot of the observation that (to you) making copies of lvalues references was a confusing API. This whole conversation is moot once we go with option (3) above in a new API level.

PR #3170, opened 2025-12-18
defer: Replace static_assert() with concept

xemul:


include/seastar/util/defer.hh
```diff
@@ -40,12 +40,11 @@ concept deferrable_action = requires (Func func) {
 namespace seastar {
 
 template <typename Func>
-requires deferrable_action<Func>
+requires deferrable_action<Func> && std::is_nothrow_move_constructible_v<Func>
 class [[nodiscard]] deferred_action {
     Func _func;
     bool _cancelled = false;
 public:
-    static_assert(std::is_nothrow_move_constructible_v<Func>, "Func(Func&&) must be noexcept");
```
avikivity, line 48: We already have this:

```c++

#ifdef SEASTAR_DEFERRED_ACTION_REQUIRE_NOEXCEPT
#define SEASTAR_DEFERRED_ACTION_NOEXCEPT noexcept
#else
#define SEASTAR_DEFERRED_ACTION_NOEXCEPT
#endif

template <typename Func>
concept deferrable_action = requires (Func func) {
    { func() } SEASTAR_DEFERRED_ACTION_NOEXCEPT -> std::same_as<void>;
};
```

avikivity: Ah, it's a different noexcept. But why not add the constraint to deferrable_action?

xemul: Because `deferrable_action` constraint also applies to `defer(Func&&)` helper method, and strictly speaking, it does not even need Func be moveable

avikivity: There are two cases:
1. Func is not a reference - in this case defer(Func()) will construct a deferred_action<Func> which will fail if it's not movable
2. Func is an lvalue reference Func1& - in this case the static_assert is works: https://godbolt.org/z/hWddjb3j7

So I think moving the constraint to deferrable_action works.

xemul: Fair point. Placed constraint in deferrable_action

xemul:
upd:
- rebased (no conflicts)

Issue #578, opened 2019-01-13
move to the aio api for sockets throws different system error. breaks api


emaxerrno:
Before the move to the aio for all io i had error handling code of:

```cpp

  .handle_exception([](std::exception_ptr eptr) {
      try {
        std::rethrow_exception(eptr);
      } catch (const std::system_error &e) {
        // Current and future \ref accept() calls will terminate immediately
        // 103 is expected
        if (e.code().value() == 103) {
          return;
        } else {
          LOG_ERROR("Unknown system error: {}", e);
        }
      } catch (const std::exception &e) {
        LOG_ERROR("Abrupt server stop: {}", e);
      }
    });
```

However, the code is now 

```
Unknown system error: std::system_error (error system:22, Invalid argument)
```

I assume this is a bug. However, please close if it isn't.

xemul:
Added accept/abort_accept test. It passes on me on all three supported backends with std::system_error(ECONNABORTED).
Feel free to re-open if I misundrstood the issue and it still stands

nyh:
In https://github.com/scylladb/seastar/pull/3171 @xemul added a test that verified that on all backends an aborted accept() returns ECONNABORTED (103). So we can't seem to reproduce the problem that you saw that it now returns EINVAL (22). If you can create a reproducer for this, please reopen this issue.

PR #3171, opened 2025-12-19
test: Add accept() vs accept_abort() socket test

xemul:
When server socket sleeps in .accept() method, it should be possible to abort the waiting. The accept() should then resolve into exception of known type and error code.

refs #578

### DIFF
--- a/tests/unit/socket_test.cc
+++ b/tests/unit/socket_test.cc
@@ -269,6 +269,32 @@ SEASTAR_TEST_CASE(socket_connect_abort_test) {
     });
 }
 
+// Check that server_socket::accept() is abortable by server_socket::abort_accept()
+SEASTAR_THREAD_TEST_CASE(socket_accept_abort_test) {
+    ipv4_addr addr("127.0.0.1", 3174);
+    server_socket ss = seastar::listen(addr, listen_options{ .reuse_address = true });
+    bool too_late = false;
+    auto f = ss.accept().then([] (auto ar) {
+        BOOST_FAIL("Accept didn't resolve into exception");
+    }).handle_exception_type([&too_late] (std::system_error e) {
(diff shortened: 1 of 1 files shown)


tests/unit/socket_test.cc
```diff
@@ -269,6 +269,32 @@ SEASTAR_TEST_CASE(socket_connect_abort_test) {
...
 
+// Check that server_socket::accept() is abortable by server_socket::abort_accept()
+SEASTAR_THREAD_TEST_CASE(socket_accept_abort_test) {
+    ipv4_addr addr("127.0.0.1", 3174);
+    server_socket ss = seastar::listen(addr, listen_options{ .reuse_address = true });
+    bool too_late = false;
+    auto f = ss.accept().then([] (auto ar) {
+        BOOST_FAIL("Accept didn't resolve into exception");
+    }).handle_exception_type([&too_late] (std::system_error e) {
+        BOOST_REQUIRE(!too_late);
```
avikivity, line 280: Please change to a coroutine or thread, they're so much more readable.

xemul:
> Looks good, but there's something I don't understand: #578, which you "refs", claimed there is a bug in this area. Are you saying that your test proves there is no such bug, and the issue was invalid (you closed it as "completed" but I guess you meant "wontfix"?)?

Yes, I ran this test with all 3 backends and it all passed.

We cannot yet make this check run all the time, as test.py runs tests with "most suitable" backend once, and individual tests can request for some specific backend (like socket_test explicitly asks for linux-aio backend). But that's yet to be improved

tests/unit/socket_test.cc
xemul: Too late, @nyh already merged it. I created #3177 to address that comment

PR #3172, opened 2025-12-19
scheduling_group_nesting_test: adjust threshold

travisdowns:
To reduce flakes in GHA.

Adjust the allowed deviation from "perfect scheduling" from 0.07 to 0.50. Locally, even 0.50 isn't enough to prevent flakes if there is some load on the system (e.g., `stress -c 128`): tests flake ~100% of the time in that scenario even at 0.50, but hopefully on the quieter GHA runners this will help.

### DIFF
--- a/tests/unit/scheduling_group_nesting_test.cc
+++ b/tests/unit/scheduling_group_nesting_test.cc
@@ -129,7 +129,7 @@ static future<> run_busyloops(std::vector<seastar::scheduling_group> groups, std
     for (unsigned i = 0; i < groups.size(); i++) {
         auto dev = float(std::abs(counts[i] / expected[i] - average_adjusted_count)) / average_adjusted_count;
         fmt::print("{}: count={} expected={:.2f} adjusted={} deviation={:.2f}\n", i, counts[i], expected[i], int(float(counts[i]) / expected[i]), dev);
-        BOOST_CHECK(dev < 0.07);
+        BOOST_CHECK(dev < 0.50);
     }
 }
 


tests/unit/scheduling_group_nesting_test.cc
```diff
@@ -129,7 +129,7 @@ static future<> run_busyloops(std::vector<seastar::scheduling_group> groups, std
     for (unsigned i = 0; i < groups.size(); i++) {
         auto dev = float(std::abs(counts[i] / expected[i] - average_adjusted_count)) / average_adjusted_count;
         fmt::print("{}: count={} expected={:.2f} adjusted={} deviation={:.2f}\n", i, counts[i], expected[i], int(float(counts[i]) / expected[i]), dev);
-        BOOST_CHECK(dev < 0.07);
+        BOOST_CHECK(dev < 0.50);
```
avikivity, line 132: Does this test anything then?

I haven't observed this test to fail. Do you run it in an underprovisioned environment?

xemul: It fails from time to time in github CI runs

The problem is that it uses real clocks, instead of manual to provide reliable outcome. Possible solution is to templatize CPU sched classes with Clock and make the test use manual_clock, but that's lots of work :(

xemul: And example of CI failure: https://github.com/scylladb/seastar/actions/runs/20481085546/job/58854507324?pr=3177

```
=== TEST-4 ===
Running in 5 groups: [0, 5, 4, 2, 1]
Shares: [4, 1, 2, 1, 5]
  supergroup 1 shares = 4
--------8<--------
0: count=8593 expected=0.29 adjusted=30075 deviation=0.01
1: count=2973 expected=0.10 adjusted=31216 deviation=0.05
2: count=5529 expected=0.19 adjusted=29027 deviation=0.03
3: count=1950 expected=0.07 adjusted=27299 deviation=0.08
/__w/seastar/seastar/tests/unit/scheduling_group_nesting_test.cc(132): error: in "test_fairness": check dev < 0.07 has failed
4: count=11204 expected=0.36 adjusted=31371 deviation=0.05
```

just a single check failed

xemul: Another attempt #3179

travisdowns: > Does this test anything then?
> 
> I haven't observed this test to fail. Do you run it in an underprovisioned environment?

I observed it failing in GHA, as Pavel says. Even if a test only has a 1% failure rate in GHA, it fails "often" in practice since there are so many CI permutations which each run the test.

Locally it doesn't fail for me if I run it on a quiet box. If I run it concurrently with `stress -c 128` it fails 100% of the time (so yes if fails if "overprovisioned" as you say).

xemul:
Closing in favor of #3179

PR #3173, opened 2025-12-20
rpc_tester: introduce rpc_streaming job based on streaming API

MrD4rkne:
In this PR we introduce new `rpc_streaming` job - with `STREAM_UNIDIRECTIONAL` and `STREAM_BIDIRECTIONAL` verbs.

## Overview

The streaming job uses Seastar’s RPC streaming API by continuously sending payloads from workers to the server, optionally receiving responses on a return stream. It is intended for measuring end-to-end throughput.

Similar to the `rpc` job's `rpc_verb::write` behavior but replaces per-payload RPC calls with the streaming API and doesn't wait for response from the server between writes.

### Metrics

- *messages* - total number of `payload_t` that was put by workers.
- *throughput* - estimated number of bytes send per second - `_total_messages * _payload_size_bytes / _total_duration.count()`, measured in `kB/s`.
- *messages per second* - `messages / _total_duration.count()`

### Benchmark

Below you can find a simple benchmark that was run on `potwor2` for 30s machine with:
- client: `cpuset=0-7`
- server: `cpuset=8-15`

```
        parallelism: 10
        shares: 100
        payload: 64kB
```

All configs:
[suite.yaml](https://github.com/user-attachments/files/24274096/suite.yaml)

#### rpc_streaming::bidirectional

image

image

#### rpc_streaming::unidirectional

image

image

#### rpc::write

image

*rpc::write doesn't produce throughput metric*

### Client

In `client` mode app:
1. Connects to the server.
2. Spawns *parallelism* workers, each starting with own delay if *sleep_time* was specified.
3. Each worker
   -  Creates `rpc::sink<payload_t>`, send it to the server and receives a corresponding `source<uint64_t>`
   - Performs the streaming phase:
      - Streams *payload* bytes in a loop through the sink until *duration* elapses.
      - For bidirectional streaming, simultaneously reads responses from the received source.
    - Waits for the server to close it's sink so the client can shut down the connection safely.

### Server

- For each `STREAM_UNIDIRECTIONAL` or `STREAM_BIDIRECTIONAL` call:
    1. Creates a  `rpc::sink<uint64_t>`
    2. Schedules background processing for the `rpc::source<payload_t>` received from the client.
    3. Returns the `rpc::source<uint64_t>` associated with the created `rpc::sink<uint64_t>`, used for streaming back the number of bytes received.
    
- For the bidirectional verb, the server sends back, on its stream, the number of bytes received for each chunk of data.

- After client closes their `rpc::sink<payload_t>`, the server writes the final total number of bytes into corresponding `rpc::sink<uint64_t>` and then closes it.

### DIFF
--- a/apps/rpc_tester/rpc_tester.cc
+++ b/apps/rpc_tester/rpc_tester.cc
@@ -355,6 +355,7 @@ enum class rpc_verb : int32_t {
     ECHO = 2,
     WRITE = 3,
     STREAM_BIDIRECTIONAL = 4,
+    STREAM_UNIDIRECTIONAL = 5,
 };
 
 using rpc_protocol = rpc::protocol<serializer, rpc_verb>;
@@ -504,6 +505,10 @@ public:
             _call = [this] (unsigned worker_id, const payload_t& payload) {
                 return run_streaming_worker(worker_id, payload, rpc_verb::STREAM_BIDIRECTIONAL);
             };
+        } else if (_cfg.verb == "unidirectional") {
+            _call = [this] (unsigned worker_id, const payload_t& payload) {
+                return run_streaming_worker(worker_id, payload, rpc_verb::STREAM_UNIDIRECTIONAL);
+            };
         } else {
             throw std::runtime_error("unknown verb, should be 'bidirectional' or 'unidirectional'");
         }
@@ -541,6 +546,7 @@ private:
     // Streaming worker: 
     // - client sends payload_t
     // - in bidirectional case: server echoes back payload_t
+    // - in unidirectional case: server only sends EOS at the end
     future<> run_streaming_worker(unsigned worker_id, const payload_t& payload, enum rpc_verb verb) {
         auto sink = co_await _client->make_stream_sink<serializer, payload_t>();
 
@@ -635,6 +641,43 @@ public:
         }
 
         fmt::print("Server received total {} messages on bidirectional stream, total payload: {} bytes\n", total_messages, total_payload);
+    }
+
+    static future<> process_uni_source(rpc::source<payload_t> source, rpc::sink<uint64_t> sink) {
+        uint64_t total_messages = 0, total_payload = 0;
+        std::exception_ptr error;
+
+        try {
+            while (true) {
+                auto data = co_await source();
+                if (!data) {
+                    // We need to have some kind of synchronization with client, so they don't close the main RPC connection
+                    // until server received EOS (client -> server connection was closed). If we don't do that, server might 
+                    // receive `seastar::rpc::stream_closed` exception on reading from the source, as the stream is terminated 
+                    // on connection close.
+                    // In order to do that, we create a connection to the other side - from server to client, and keep it
+                    // open until client sends EOS - then, by closing this connection, server also sends EOS.
+                    // This gives client the guarantee that server received all data before closing the main RPC connection.
+                    break;
+                }
+                ++total_messages;
+                total_payload += std::get<0>(*data).size() * sizeof(payload_t::value_type);
+            }
+        } catch (...) {
+            error = std::current_exception();
+        }
+
+        auto flush_future = co_await coroutine::as_future(sink.flush());
+        if (flush_future.failed() && !error) [[unlikely]] {
+            error = flush_future.get_exception();
+        }
+        co_await sink.close();
+
+        if (error) {
+            std::rethrow_exception(error);
+        }
+
+        fmt::print("Server received total {} messages on unidirectional stream, total payload: {} bytes\n", total_messages, total_payload);
     }
 };
 
@@ -762,6 +805,13 @@ public:
             auto sink = source.make_sink<serializer, payload_t>();
 
             (void)job_rpc_streaming::process_bi_source(std::move(source), sink);
+
+            return sink;
+        });
+        _rpc->register_handler(rpc_verb::STREAM_UNIDIRECTIONAL, [] (rpc::source<payload_t> source) {
+            auto sink = source.make_sink<serializer, uint64_t>();
+            
+            (void)job_rpc_streaming::process_uni_source(std::move(source), sink);
 
             return sink;
         });
--- a/apps/rpc_tester/sample-conf.yaml
+++ b/apps/rpc_tester/sample-conf.yaml
@@ -13,8 +13,9 @@ jobs:
     timeout: # optional rpc send timeout duration
   - name: # any parseable string
     type: rpc_streaming
-    verb: # string: bidirectional
+    verb: # string, one of: bidirectional, unidirectional
            # - bidirectional: client sends payload_t, server responds asynchronously with received data
+           # - unidirectional: client sends payload_t, server consumes without responding
     parallelism: # number of streaming workers to run simultaneously
     shares: # sched group shares (100 by default)
     payload: # number of bytes in the payload sent from client to server, accepts kB suffix


apps/rpc_tester/rpc_tester.cc
```diff
@@ -474,6 +476,152 @@ class job_rpc : public job {
...
+                    return src().then([] (std::optional<std::tuple<uint64_t>> data) {
+                        if (!data) {
+                            // EOS from server
+                            return stop_iteration::yes;
+                        }
+                        return stop_iteration::no;
+                    });
+                });
+                return when_all(std::move(sender), std::move(receiver)).discard_result();
+            });
```
avikivity, line 565: Coroutines would be much more readable, here and elsewhere.

apps/rpc_tester/rpc_tester.cc
```diff
@@ -474,6 +476,152 @@ class job_rpc : public job {
...
+        });
+    }
+
+    virtual void emit_result(YAML::Emitter& out) const override {
+        out << YAML::Key << "messages" << YAML::Value << _total_messages;
+
+        auto total_bytes = _total_messages * _payload_size_bytes;
+        if (_total_duration.count() > 0) {
+            double throughput_kBps = (total_bytes >> 10) / _total_duration.count();
```
avikivity: Better to keep throughput in bytes/second, natural units are easier to scale later.

apps/rpc_tester/rpc_tester.cc
MrD4rkne: If we got that right, @xemul advised us to used continuations on purpose. @xemul could you please spill some details?

apps/rpc_tester/rpc_tester.cc
MrD4rkne: Sure, we will change that

apps/rpc_tester/rpc_tester.cc
xemul: I told that latency-sensitive parts of the code should better use then-chains. Preparational code that sets things up (e.g. prepares sinks/sources) is not such and can benefit from coroutines readability

apps/rpc_tester/rpc_tester.cc
```diff
@@ -474,6 +476,152 @@ class job_rpc : public job {
...
+        return do_with(std::move(source), std::move(sink), uint64_t{0}, uint64_t{0},
+                [] (rpc::source<payload_t>& src, rpc::sink<uint64_t>& sink, uint64_t& total_messages, uint64_t& total_payload) {
+                    return repeat([&src, &sink, &total_messages, &total_payload] {
+                        return src().then([&sink, &total_messages, &total_payload](std::optional<std::tuple<payload_t>> data) {
+                            if (!data) {
+                                return make_ready_future<stop_iteration>(stop_iteration::yes);
+                            }
+                            ++total_messages;
+                            total_payload += std::get<0>(*data).size() * sizeof(payload_t::value_type);
+                            // Send current total_payload back to client
```
xemul: I think that sending back to client small messages (with received sizes) doesn't meke much sense. The bidirectional streaming should better stream the same amount of data in both directions

apps/rpc_tester/rpc_tester.cc
```diff
@@ -499,10 +500,14 @@ class job_rpc_streaming : public job {
             , _payload_size_bytes(_cfg.payload) {
                 if (_cfg.verb == "bidirectional") {
                     _call = [this] (unsigned worker_id, const payload_t& payload) {
-                        return run_bidirectional_worker(worker_id, payload);
+                        return run_streaming_worker(worker_id, payload, rpc_verb::STREAM_BIDIRECTIONAL);
```
xemul: Please, don't re-name the suff that you introduced in the previous patch of the PR. Give it the final name immediately. Otherwise it complicates the review

apps/rpc_tester/rpc_tester.cc
```diff
@@ -530,10 +535,10 @@ class job_rpc_streaming : public job {
...
-    future<> run_bidirectional_worker(unsigned worker_id, const payload_t& payload) {
-        return _client->make_stream_sink<serializer, payload_t>().then([this, &payload] (rpc::sink<payload_t> sink) {
-            auto rpc_call = _rpc.make_client<rpc::source<uint64_t>(rpc::sink<payload_t>)>(rpc_verb::STREAM_BIDIRECTIONAL);
+    // Streaming worker: client sends payload_t and receives uint64_t totals from server
+    future<> run_streaming_worker(unsigned worker_id, const payload_t& payload, enum rpc_verb verb) {
+        return _client->make_stream_sink<serializer, payload_t>().then([this, &payload, verb] (rpc::sink<payload_t> sink) {
+            auto rpc_call = _rpc.make_client<rpc::source<uint64_t>(rpc::sink<payload_t>)>(verb);
             return rpc_call(*_client, sink).then([this, sink = std::move(sink), &payload] (rpc::source<uint64_t> source) mutable {
                 auto sender = stream_data(std::move(sink), payload);
                 // Receiver: drain partial totals from server until EOS
```
xemul: How is unidirectional worker going to "drain partial totals from server"? Server doesn't send anything for unidirectional job

MrD4rkne: It does receive the total payload. When the server received EOS it sends the total received payload. In bidirectional scenario, the sum is sent after each received `payload_t`. In unidirectional, only after EOS.

Might sounds confusing, but the worker receives `source<uint64_t>` in unidirectional scenario so it can safely close the connection.

Without such application-level synchronization, we've seen packet race of TCP `FIN` packets. Basically, chain:

```cpp
stream.close().then([&](){ return connection.close(); });
```

resulted in `seastar::rpc::stream_closed` when the server was reading from the `source<payload_t>`. Happened sometimes, especially if the underlying TCP connection for the stream was overloaded.

apps/rpc_tester/rpc_tester.cc
witek-formanski: @avikivity we used kB/s to stay concise with throughput in `io_tester` (see [here](https://github.com/scylladb/seastar/blob/master/apps/io_tester/io_tester.cc#L699)). Just to make sure - should we still change it to B/s?

apps/rpc_tester/rpc_tester.cc
witek-formanski: Sure, we will change it

apps/rpc_tester/rpc_tester.cc
avikivity: A do_with() should strongly suggest a coroutine. The main problem with coroutines is forced allocation, but a do_with() also allocates so it becomes even. The coroutine's readability then wins.

Continuation changes can win on very simple code paths that often return ready futures, like semaphore::wait(), or short loops from loop.hh if the body doesn't yield, but anything complicated is better done in coroutines.

apps/rpc_tester/rpc_tester.cc
avikivity: Better to use bytes/sec. kB/sec will still generate large numbers so it isn't more readable.

witek-formanski:
## Changelog

### Requested changes
- coroutines instead of continuations (for readability)
- throughput in bytes instead of kilobytes
- bidirectional server responds with data, not number of bytes received so far

### Additional changes
- payload in constructor
- refactored `run()` method (extracted `run_worker_with_delay()`)
- slightly modified comments and descriptions
- unidirectional server doesn't send anything to client (sending EOS when ending stream is sufficient to avoid bug noticed before)

## Benchmark after changes


**rpc_streaming::bidirectional**
image
image

**rpc_streaming::unidirectional**
image
image

**rpc::write**
image

apps/rpc_tester/rpc_tester.cc
MrD4rkne: We've utilized co-routines for the setup and the "handshake", but kept it for the it  read/write loops

apps/rpc_tester/rpc_tester.cc
MrD4rkne: Done

apps/rpc_tester/rpc_tester.cc
```diff
@@ -474,6 +476,154 @@ class job_rpc : public job {
...
+    payload_t _payload;
+
+public:
+    job_rpc_streaming(job_config cfg, rpc_protocol& rpc, client_config ccfg, socket_address caddr)
+            : _cfg(cfg)
+            , _caddr(std::move(caddr))
+            , _ccfg(ccfg)
+            , _rpc(rpc)
+            , _stop(std::chrono::steady_clock::now() + _cfg.duration)
+            , _payload_size_bytes(_cfg.payload),
```
Deexie: Nit: comma in the next line

apps/rpc_tester/rpc_tester.cc
```diff
@@ -474,6 +476,154 @@ class job_rpc : public job {
...
+            _payload(_cfg.payload / sizeof(payload_t::value_type), 0) {
+                if (_cfg.verb == "bidirectional") {
+                    _call = [this] (unsigned worker_id, const payload_t& payload) {
```
Deexie: Nit: We usually do the following indentation:
```
name()
        : _a()
        , _b()
{
    do_sth();
}
```

Deexie: (Twice before params initialization, once before instructions)

apps/rpc_tester/rpc_tester.cc
```diff
@@ -474,6 +476,154 @@ class job_rpc : public job {
...
+                } else {
+                    throw std::runtime_error("unknown verb, should be 'bidirectional' or 'unidirectional'");
+                }
+            }
+
+    virtual std::string name() const override { return _cfg.name; }
+
+private:
+    future<> stream_data(rpc::sink<payload_t> sink, const payload_t& payload) {
+        return do_with(std::move(sink), [this, &payload] (rpc::sink<payload_t>& sink) {
```
Deexie: Why not a coroutine here?

apps/rpc_tester/rpc_tester.cc
```diff
@@ -474,6 +476,154 @@ class job_rpc : public job {
...
+        } else {
+            out << YAML::Key << "throughput" << YAML::Value << "inf" << YAML::Comment("kB/s");
+            out << YAML::Key << "messages per second" << YAML::Value << "inf";
+        }
+    }
+
+    static future<> process_bi_source(rpc::source<payload_t> source, rpc::sink<payload_t> sink) {
+        uint64_t total_messages = 0, total_payload = 0;
+
+        co_await repeat([&source, &sink, &total_messages, &total_payload] {
```
Deexie: This should be changed to `while` loop and `co_await`s

apps/rpc_tester/rpc_tester.cc
```diff
@@ -622,6 +628,34 @@ class job_rpc_streaming : public job {
 
         fmt::print("Server received total {} messages on bidirectional stream, total payload: {} bytes\n", total_messages, total_payload);
     }
+
+    static future<> process_uni_source(rpc::source<payload_t> source, rpc::sink<uint64_t> sink) {
+        uint64_t total_messages = 0, total_payload = 0;
+
+        co_await repeat([&source, &sink, &total_messages, &total_payload] {
```
Deexie: Same here, if you are using coroutines, use their whole potential

apps/rpc_tester/rpc_tester.cc
MrD4rkne: What about https://github.com/scylladb/seastar/pull/3173#discussion_r2640123556 ?
We tried to keep continuations in the loop for streaming/consuming the data, and co-routines for everything else

apps/rpc_tester/rpc_tester.cc
MrD4rkne: to flush and close the sink even on error. Can it be done easier?

apps/rpc_tester/rpc_tester.cc
Deexie: it may be a continuation chain but then you don't need to co_await repeat as it doesn't give much

apps/rpc_tester/rpc_tester.cc
Deexie: you can use try-catch clause

apps/rpc_tester/rpc_tester.cc
Deexie: @xemul @avikivity I'm not sure about this one. Which approach do you suggest here (as it's in the main path):
1. The current one: couroutine in the outer scope + repeat
2. Only continuation chain in the whole `process_bi_source`
3. Only coroutines?

avikivity: For an rpc tester, there's no reason not to coroutines everywhere, it's the cleanest option and the small performance hit won't matter.

avikivity:
Are those results repeatable? The dependency on shard count looks random.

apps/rpc_tester/rpc_tester.cc
MrD4rkne: what about:

```cpp
    template<typename T>
    static future<> flush_and_close(rpc::sink<T>& sink){
        try {
            co_await sink.flush();
        } catch (...) {
        }
        try {
            co_await sink.close();
        } catch (...) {
        }
    }

    future<> stream_data(rpc::sink<payload_t> sink, const payload_t& payload) {
        auto d = defer([&sink] () noexcept {
            seastar::async([s = std::move(sink)] () mutable {
                flush_and_close(s).get();
            }).handle_exception([](auto) {});
        });
        
        while (std::chrono::steady_clock::now() <= _stop) {
            ++_total_messages;
            co_await sink(payload);
            if (_cfg.sleep_time) {
                co_await seastar::sleep(std::chrono::duration_cast<std::chrono::nanoseconds>(*_cfg.sleep_time));
            }
        }
    }
```

is it viable?

MrD4rkne:
> Are those results repeatable? The dependency on shard count looks random.

they tend to have some deviations, but we suspect it's because of server's load (not only we're using it). We can run benchmarks and plot it with standard deviation. Would you want to see such charts?

apps/rpc_tester/rpc_tester.cc
MrD4rkne: oh, we're ignoring future here (::async)...

avikivity: .close() shouldn't cause exceptions if you flush() before, so the second try is unnecessary.

Creating a thread should be reserved for top-level operations so their numbers can be kept in check (threads are heavyweight). 

You can keep it simple - a try/catch around the outer loop + flush, capturing the exception, then a close, then returning the exception if any.

See utils/closeable.hh, with_closeable.

avikivity:
> > Are those results repeatable? The dependency on shard count looks random.
> 
> they tend to have some deviations, but we suspect it's because of server's load (not only we're using it). We can run benchmarks and plot it with standard deviation. Would you want to see such charts?

Prefer to see known stable results.

apps/rpc_tester/rpc_tester.cc
MrD4rkne: Thanks, we will check it out

witek-formanski:
> The dependency on shard count looks random.

What do you mean by "dependency on shard count"? Do you mean that shards with different ids (in the graphs above from 0 to 7) differ among each other for the same run of `rpc_tester`? Or something else?

witek-formanski:
> > > Are those results repeatable? The dependency on shard count looks random.
> > 
> > 
> > they tend to have some deviations, but we suspect it's because of server's load (not only we're using it). We can run benchmarks and plot it with standard deviation. Would you want to see such charts?
> 
> Prefer to see known stable results.

We run the tests on a server that is shared among other employees, so the most we can do for stability of the results is to run the tests a couple of times and calculate standard deviation (to minimize chance of results being impacted by other users of the server).

witek-formanski:
## Changelog

### Changes in code
- coroutines everywhere instead of continuations
- style

### Different graphs

We changed the way we generate graphs - now we ran each test three times and calculated the standard deviation (plotted in black).

## Benchmark after changes

**rpc_streaming::bidirectional**
image
image

**rpc_streaming::unidirectional**
image
image

**rpc::write**
image

apps/rpc_tester/rpc_tester.cc
```diff
@@ -474,6 +476,173 @@ class job_rpc : public job {
...
+                if (_cfg.sleep_time) {
+                    co_await seastar::sleep(std::chrono::duration_cast<std::chrono::nanoseconds>(*_cfg.sleep_time));
+                }
+            }
+        } catch (...) {
+            error = std::current_exception();
+        }
+
+        try {
+            co_await sink.flush();
```
Deexie: Those try-catches don't look best. Try this instead:

```
auto flush_future = co_await coroutine::as_future(sink.flush());
if (flush_future.failed() && !error) [[unlikely]] {
    error = f.get_exception();
}
```

Deexie: Note: `flush_future.failed()` is first on purpose. It would complain if future failed and it wasn't checked

apps/rpc_tester/rpc_tester.cc
```diff
@@ -474,6 +476,173 @@ class job_rpc : public job {
...
+                total_payload += received_data.size() * sizeof(payload_t::value_type);
+                // Send data back to client
+                co_await sink(received_data);
+            }
+        } catch (...) {
+            error = std::current_exception();
+        }
+
+        try {
+            co_await sink.flush();
```
Deexie: Ditto

apps/rpc_tester/rpc_tester.cc
```diff
@@ -641,6 +647,46 @@ class job_rpc_streaming : public job {
...
+        } catch (...) {
+        }
+        try {
+            co_await sink.close();
```
Deexie: And here

apps/rpc_tester/rpc_tester.cc
Deexie: Check this out: https://github.com/scylladb/seastar/pull/3225

witek-formanski:
## Changelog

- modified try-catch clauses (`coroutine::as_future`)

Deexie:
@avikivity please re-review

avikivity:
> > The dependency on shard count looks random.
> 
> What do you mean by "dependency on shard count"? Do you mean that shards with different ids (in the graphs above from 0 to 7) differ among each other for the same run of `rpc_tester`? Or something else?

I misunderstood the chart in https://github.com/scylladb/seastar/pull/3173#issuecomment-3723954476. I thought the x axis was the number of shards (can't be due to 0).

Still, why so wildly different results for different shards?

Not so important.

avikivity:
The huge standard deviations mean the test can't be relied on.

Still, it can serve as a base for further improvements.

PR #3174, opened 2025-12-21
Add file::statat and a corresponding file_stat overload

bhalevy:
Exposing the fstatat system call.
To be used for efficient directory listing
where, instead of calling file_stat for the full path, the caller can use dir.statat(dirent.name) for each entry which is more efficient as it doesn't require to construct the full path and to look up the directory prefix on every call.

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -210,6 +210,18 @@ posix_file_impl::stat() noexcept {
     co_return ret.extra;
 }
 
+future<struct stat>
+posix_file_impl::statat(std::string_view name, int flags) noexcept {
+    auto ret = co_await engine()._thread_pool->submit<syscall_result_extra<struct stat>>(
+            internal::thread_pool_submit_reason::file_operation, [fd = _fd, name = sstring(name), flags] {
+        struct stat st;
+        auto ret = ::fstatat(fd, name.data(), &st, flags);
+        return wrap_syscall(ret, st);
+    });
+    ret.throw_if_error();
+    co_return ret.extra;
+}
+
 future<>
 posix_file_impl::truncate(uint64_t length) noexcept {
     auto sr = co_await engine()._thread_pool->submit<syscall_result<int>>(
@@ -1239,6 +1251,14 @@ future<struct stat> file::stat() noexcept {
   } catch (...) {
     return current_exception_as_future<struct stat>();
   }
+}
+
+future<struct stat> file::statat(std::string_view name, int flags) noexcept {
+    try {
+        return _file_impl->statat(name, flags);
+    } catch (...) {
+        return current_exception_as_future<struct stat>();
+    }
 }
 
 future<> file::flush() noexcept {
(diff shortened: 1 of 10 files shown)


src/core/file.cc
```diff
@@ -1241,6 +1253,14 @@ future<struct stat> file::stat() noexcept {
   }
 }
 
+future<struct stat> file::statat(std::string_view name, int flags) noexcept {
+  try {
```
avikivity: Why half indents?

tests/unit/directory_test.cc
```diff
@@ -110,6 +110,7 @@ class test_file_impl : public file_impl {
 
     virtual future<> flush() override { return get_file_impl(_lower)->flush(); }
     virtual future<struct stat> stat() override { return get_file_impl(_lower)->stat(); }
+    virtual future<struct stat> statat(std::string_view name, int flags) override { return get_file_impl(_lower)->statat(name, flags); }
```
avikivity, line 113: stat() means operate on this object, statat() means operate on a nested name.

I wonder if we can make it clearer. I wouldn't want to lose the "statat" name, which, while awkward, is immediately understood.

Perhaps: add an additional file_stat overload that accepts a file object representing the base directory to seastar.hh.

avikivity: Should we encapsulate the flags in an enum? Like we have for open_flags.

src/core/file.cc
bhalevy: That's copy-pasta from file::stat(), but I agree there's no good reason to keep it this way.

tests/unit/directory_test.cc
bhalevy: > Perhaps: add an additional file_stat overload that accepts a file object representing the base directory to seastar.hh

I like that

bhalevy: > Should we encapsulate the flags in an enum? Like we have for open_flags.

I contemplated about that too.  Since the function returns `struct stat` which is system/posix defined so I thought it'd be appropriate for it to get system flags, similar to `make_file_impl`. Yet, since in linux the only "standard" (i.e. not linux-specific) flag is AT_SYMLINK_NOFOLLOW, we can just add a `follow_symlink` parameter to the higher-level file_stat overload and translate follow_symlink::no to AT_SYMLINK_NOFOLLOW.

src/core/file.cc
bhalevy: Fixed in [0b01023](https://github.com/scylladb/seastar/commit/0b010234e0fba4b5e2de2785837ff150bc3dd23d)

tests/unit/directory_test.cc
bhalevy: Add file_stat overload (accepting `follow_symlink` parameter) in [0b01023](https://github.com/scylladb/seastar/commit/0b010234e0fba4b5e2de2785837ff150bc3dd23d)

PR #3178, opened 2025-12-24
demos:improve the output of demo_with_io_intent() in file_demo

baiwfg2:
During my test, the assertion when cancelled=false is never checked, so add a simple sleep call betwen f2 request and cancel action, so that f2 has chance of being ready before request with io_intent is cancelled. In this case cancelled is still false

Interestingly even we give wait_ms 0, f2 is still fulfilled. I suppose that's because even sleep with 0 also triggers one reactor scheduling, letting io request get the chance to run

If anyone cannot produce `cancelled` = false, then wait_ms could be increased a little bit.

### DIFF
--- a/demos/file_demo.cc
+++ b/demos/file_demo.cc
@@ -35,12 +35,15 @@
 #include <seastar/core/sstring.hh>
 #include <seastar/core/temporary_buffer.hh>
 #include <seastar/core/loop.hh>
+#include <seastar/core/sleep.hh>
 #include <seastar/core/io_intent.hh>
 #include <seastar/util/assert.hh>
 #include <seastar/util/log.hh>
 #include <seastar/util/tmp_file.hh>
 
 using namespace seastar;
+namespace bpo = boost::program_options;
+using namespace std::chrono_literals;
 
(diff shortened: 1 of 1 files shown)


demos/file_demo.cc
```diff
@@ -169,9 +171,9 @@ future<> demo_with_file_close_on_failure() {
 
 static constexpr size_t half_aligned_size = aligned_size / 2;
 
-future<> demo_with_io_intent() {
+future<> demo_with_io_intent(int64_t wait_ms) {
```
avikivity: Please use std::chrono::duration

baiwfg2: done. changed to  std::chrono:duration<double>, just like what other codes did (to allow for diverse conversion, I think)

xemul:
Alpine job fails on #3180 (needs #3179)

baiwfg2:
@xemul  You mean I should rebase after https://github.com/scylladb/seastar/pull/3179 gets merged ?  No need to close ?

baiwfg2:
Now I know what `baiwfg2 authored and xemul committed` means. Different git collaboration style.

PR #3182, opened 2025-12-25
reactor_backend: Fix iocb double-free during AIO teardown

### LINKED ISSUE #3169: bad_alloc on AIO teardown
Sometimes (once a year or so) reactor crashes re-allocating a vector of iocbs for retry

```
[2024-02-16T09:17:41.425Z] ERROR 2024-02-16 10:19:34,274 [shard 0:main] seastar - aio retry failed: boost::container::bad_alloc (boost::container::bad_alloc thrown). Aborting.
```

The code is trying to push an iocb pointer to `_free_iocbs` even it already had it at the top. and at that moment, the underlying static_vector was already full. BTW, the iocb pointer points to the last slot in the _iocb_pool.

```
(gdb) p seastar::local_engine                                                                                          
$1 = (seastar::reactor *) 0x6010000be000  
(gdb) p seastar::local_engine->_backend                                                                                
$2 = std::unique_ptr<seastar::reactor_backend> = {                                                                                                                                                                                            
  get() = 0x601000140000                                                                                               
}   
(gdb) p iocbs[0]                                                                                                                                                                                                                              
$8 = (seastar::internal::linux_abi::iocb *) 0x601000152040      
(gdb) p/x &(((seastar::reactor_backend_aio*)(0x601000140000)))->_storage_context._iocb_pool._iocb_pool._M_elems[0]
$16 = 0x601000142080               
(gdb) p/x &(((seastar::reactor_backend_aio*)(0x601000140000)))->_storage_context._iocb_pool._iocb_pool._M_elems[1024]
$17 = 0x601000152080 
(gdb) p/x sizeof (((seastar::reactor_backend_aio*)(0x601000140000)))->_storage_context._iocb_pool._iocb_pool._M_elems[0]
$19 = 0x40 
(gdb) python-interactive                                                                                                                                                                                                                      
>>> free_cbs = static_vector(gdb.parse_and_eval("(((seastar::reactor_backend_aio*)(0x601000140000)))->_storage_context._iocb_pool._free_iocbs.c.m_holder"))
>>> print(len(free_cbs))
1024
>>> for i, e in enumerate(free_cbs):
...     if str(e) == '0x601000152040':  
...         print(i)
1023
>>> iocbs_0 = gdb.parse_and_eval('*iocbs[0]')
>>> print(iocbs_0)
{
  aio_data = 105621836876800,
  aio_key = 0,
  aio_rw_flags = 0,
  aio_lio_opcode = seastar::internal::linux_abi::iocb_cmd::PREAD,
  aio_reqprio = 0,
  aio_fildes = 14,
  aio_buf = 105621837724672,
  aio_nbytes = 512,
  aio_offset = 0,
  aio_reserved2 = 0,
  aio_flags = 1,
  aio_resfd = 12
}
```


No reliable reproducer

tchaikov:
Fix the double-free bug in iocb pool management that causes bad_alloc during AIO teardown (issue #3169, https://github.com/scylladb/scylladb/issues/17369, https://github.com/scylladb/scylladb/issues/26481).

Root Cause:
The bug occurs when stop() is called while AIO retry operations are in progress. The race condition:

1. schedule_retry() A is processing iocbs in _aio_retries
2. stop() calls std::exchange(_pending_aio_retry_fut, ready_future)
3. New I/Os complete with -EAGAIN, added to _pending_aio_retry
4. submit_work() sees need_to_retry()=TRUE && !retry_in_progress() (because we exchanged the future) and starts a NEW schedule_retry() B
5. schedule_retry() A completes, .finally() block runs
6. .finally() calls reap_completions(false) which returns -EAGAIN iocbs to the pool (since allow_retry=false)
7. But schedule_retry() B is still running with those same iocbs in _aio_retries!
8. When B tries to process them and encounters errors, handle_aio_error() calls put_one() again
9. Double-free: attempting to push to a full 1024-capacity free list, causing boost::container::bad_alloc

The Fix:
Add a _stopping flag that is set at the start of stop(). Check this flag in need_to_retry() to prevent new schedule_retry() loops from starting during shutdown. This eliminates the race window where a second retry loop could start after the future is exchanged.

The existing retry loop completes normally, and stop()'s .finally() block simply waits for all in-flight iocbs to be reaped and returned. No iocbs are leaked, and no double-free can occur.

Additionally, add debug-mode tracking (_iocb_allocated[]) to detect future double-free bugs and provide diagnostic information.

Fixes: #3169

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -115,6 +115,9 @@ aio_storage_context::iocb_pool::iocb_pool() {
     for (unsigned i = 0; i != max_aio; ++i) {
         _free_iocbs.push(&_all_iocbs[i]);
     }
+#ifdef SEASTAR_IOCB_POOL_DEBUG
+    _iocb_allocated.reset();
+#endif
 }
 
 aio_storage_context::aio_storage_context(reactor& r)
@@ -130,10 +133,43 @@ aio_storage_context::~aio_storage_context() {
     internal::io_destroy(_io_context);
 }
 
+void aio_storage_context::reap_pending_retries() {
+    // Drain pending retries and complete them with -ECANCELED.
+    for (auto iocb : _pending_aio_retry) {
+        _iocb_pool.put_one(iocb);
+        auto desc = get_user_data<kernel_completion>(*iocb);
+        desc->complete_with(-ECANCELED);
+    }
+    _pending_aio_retry.clear();
+
+    // _aio_retries is empty in the normal call path: this function only runs
+    // after the retry loop's future has resolved, and that loop's predicate
+    // only returns true when both vectors are empty. The drain below is
+    // defensive — if a future change to schedule_retry() leaves entries
+    // here, we still avoid leaking iocbs.
+    for (auto iocb : _aio_retries) {
+        _iocb_pool.put_one(iocb);
+        auto desc = get_user_data<kernel_completion>(*iocb);
+        desc->complete_with(-ECANCELED);
+    }
+    _aio_retries.clear();
+}
+
 future<> aio_storage_context::stop() noexcept {
+    // Set _stopping first so any reap_completions() and schedule_retry()
+    // calls that race with the drain below see the right state.
+    _stopping = true;
+
+    // Drain items in io_sink (operations without allocated iocbs yet).
+    _r._io_sink.drain([] (const internal::io_request& req, io_completion* desc) -> bool {
+        desc->complete_with(-ECANCELED);
+        return true;
+    });
+
     return std::exchange(_pending_aio_retry_fut, make_ready_future<>()).finally([this] {
         return do_until([this] { return !_iocb_pool.outstanding(); }, [this] {
-            reap_completions(false);
+            reap_completions();
+            reap_pending_retries();
             return make_ready_future<>();
         });
     });
@@ -144,12 +180,29 @@ internal::linux_abi::iocb&
 aio_storage_context::iocb_pool::get_one() {
     auto io = _free_iocbs.top();
     _free_iocbs.pop();
+#ifdef SEASTAR_IOCB_POOL_DEBUG
+    auto index = io - _all_iocbs.data();
+    SEASTAR_ASSERT(index < max_aio);
+    SEASTAR_ASSERT(!_iocb_allocated[index] && "Double allocation of iocb");
+    _iocb_allocated[index] = true;
+#endif
     return *io;
 }
 
 inline
 void
 aio_storage_context::iocb_pool::put_one(internal::linux_abi::iocb* io) {
+#ifdef SEASTAR_IOCB_POOL_DEBUG
+    auto index = io - _all_iocbs.data();
+    SEASTAR_ASSERT(index < max_aio && "iocb pointer out of range");
+    if (!_iocb_allocated[index]) {
+        seastar_logger.error("Double-free detected: iocb at index {} (ptr={}) is being returned but was not allocated",
+                           index, fmt::ptr(io));
+        std::abort();
+    }
+    _iocb_allocated[index] = false;
+    SEASTAR_ASSERT(_free_iocbs.size() < max_aio && "Double-free: iocb pool already full");
+#endif
     _free_iocbs.push(io);
 }
 
@@ -243,7 +296,7 @@ aio_storage_context::submit_work() {
         did_work = true;
     }
 
-    if (need_to_retry() && !retry_in_progress()) {
+    if (need_to_retry()) {
         schedule_retry();
     }
 
@@ -251,6 +304,15 @@ aio_storage_context::submit_work() {
 }
 
 void aio_storage_context::schedule_retry() {
+    // Don't start a new retry loop if either:
+    //   (a) one is already in flight  (!_pending_aio_retry_fut.available()),
+    //   (b) we're shutting down       (_stopping) — stop() will drain the
+    //       queues itself; starting a second loop here would race with
+    //       stop()'s drain on the same iocb pool (the original double-free).
+    if (!_pending_aio_retry_fut.available() || _stopping) {
+        return;
+    }
+
     // loop until both _pending_aio_retry and _aio_retries are empty.
     // While retrying _aio_retries, new retries may be queued onto _pending_aio_retry.
     _pending_aio_retry_fut = do_until([this] {
@@ -296,6 +358,11 @@ void aio_storage_context::schedule_retry() {
 
 bool aio_storage_context::reap_completions(bool allow_retry)
 {
+    // During shutdown, complete EAGAIN iocbs immediately instead of
+    // queuing them for retry — schedule_retry() is gated on _stopping.
+    if (_stopping) {
+        allow_retry = false;
+    }
     struct timespec timeout = {0, 0};
     auto n = io_getevents(_io_context, 1, max_aio, _ev_buffer, &timeout, _r._cfg.force_io_getevents_syscall);
     if (n == -1 && errno == EINTR) {
--- a/src/core/reactor_backend.hh
+++ b/src/core/reactor_backend.hh
@@ -32,12 +32,17 @@
 
 #include <fmt/ostream.h>
 #include <sys/time.h>
+#include <bitset>
 #include <thread>
 #include <stack>
 #include <boost/any.hpp>
 #include <boost/program_options.hpp>
 #include <boost/container/static_vector.hpp>
 
+
+#ifdef SEASTAR_DEBUG
+#define SEASTAR_IOCB_POOL_DEBUG
+#endif
 
 namespace seastar {
 
@@ -63,6 +68,11 @@ class aio_storage_context {
     class iocb_pool {
         alignas(cache_line_size) std::array<internal::linux_abi::iocb, max_aio> _all_iocbs;
         std::stack<internal::linux_abi::iocb*, boost::container::static_vector<internal::linux_abi::iocb*, max_aio>> _free_iocbs;
+#ifdef SEASTAR_IOCB_POOL_DEBUG
+        // Track which iocbs are currently allocated (not in free list)
+        // This helps detect double-free bugs
+        std::bitset<max_aio> _iocb_allocated;
+#endif
     public:
         iocb_pool();
         internal::linux_abi::iocb& get_one();
@@ -80,15 +90,14 @@ class aio_storage_context {
     pending_aio_retry_t _pending_aio_retry; // Pending retries iocbs
     pending_aio_retry_t _aio_retries;       // Currently retried iocbs
     future<> _pending_aio_retry_fut = make_ready_future<>();
+    bool _stopping = false;
     internal::linux_abi::io_event _ev_buffer[max_aio];
 
     bool need_to_retry() const noexcept {
         return !_pending_aio_retry.empty() || !_aio_retries.empty();
     }
 
-    bool retry_in_progress() const noexcept {
-        return !_pending_aio_retry_fut.available();
-    }
+    void reap_pending_retries();
 
 public:
     explicit aio_storage_context(reactor& r);


tchaikov:
### The Problem

After `stop()` calls `std::exchange(_pending_aio_retry_fut, make_ready_future<>())`, the function `retry_in_progress()` returns `FALSE` (since it checks `!_pending_aio_retry_fut.available()`). This allows a **second** `schedule_retry()` loop to start while the first one is completing, creating a race condition.

```mermaid
sequenceDiagram
    participant Reactor
    participant ScheduleRetry_A as schedule_retry() A
    participant Stop as stop()
    participant ScheduleRetry_B as schedule_retry() B
    participant Kernel
    participant Pool as iocb_pool

    Note over ScheduleRetry_A: Processing iocbs in _aio_retries
    ScheduleRetry_A->>Kernel: io_submit(iocb_X, ...)
    Note over Kernel: iocb_X in flight

    Reactor->>Stop: stop() called
    Stop->>Stop: exchange(_pending_aio_retry_fut)
    Note over Stop: retry_in_progress() now FALSE

    Kernel-->>Reactor: iocb_Y completes with -EAGAIN
    Reactor->>Reactor: reap_completions(true)
    Note over Reactor: iocb_Y → _pending_aio_retry

    Reactor->>Reactor: submit_work() checks
    Note over Reactor: need_to_retry() = TRUE<br/>retry_in_progress() = FALSE
    Reactor->>ScheduleRetry_B: NEW schedule_retry() B starts!
    ScheduleRetry_B->>ScheduleRetry_B: swap _pending_aio_retry → _aio_retries
    Note over ScheduleRetry_B: iocb_Y now in _aio_retries

    ScheduleRetry_A->>ScheduleRetry_A: Completes (vectors empty)
    ScheduleRetry_A->>Stop: .finally() block runs

    Kernel-->>Stop: iocb_Y completes with -EAGAIN again
    Stop->>Stop: reap_completions(false)
    Note over Stop: allow_retry=FALSE
    Stop->>Pool: put_one(iocb_Y) ✓ FIRST FREE
    Note over Pool: iocb_Y returned to pool

    ScheduleRetry_B->>ScheduleRetry_B: Process iocb_Y
    ScheduleRetry_B->>ScheduleRetry_B: Gets error (EBADF)
    ScheduleRetry_B->>Pool: put_one(iocb_Y) ✗ SECOND FREE
    Note over Pool: CRASH! Double-free detected<br/>boost::container::bad_alloc
```
### The Fix

```mermaid
sequenceDiagram
    participant Reactor
    participant ScheduleRetry_A as schedule_retry() A
    participant Stop as stop()
    participant SubmitWork as submit_work()
    participant Kernel
    participant Pool as iocb_pool

    Note over ScheduleRetry_A: Processing iocbs in _aio_retries
    ScheduleRetry_A->>Kernel: io_submit(iocb_X, ...)

    Reactor->>Stop: stop() called
    Stop->>Stop:  _stopping = true
    Stop->>Stop: exchange(_pending_aio_retry_fut)

    Kernel-->>Reactor: iocb_Y completes with -EAGAIN
    Reactor->>Reactor: reap_completions(true)
    Note over Reactor: iocb_Y → _pending_aio_retry

    Note over ScheduleRetry_A: Existing loop continues<br/>Processes iocb_Y
    ScheduleRetry_A->>ScheduleRetry_A: swap _pending_aio_retry → _aio_retries
    ScheduleRetry_A->>Kernel: io_submit(iocb_Y, ...)

    Reactor->>SubmitWork: submit_work() checks
    SubmitWork->>SubmitWork: need_to_retry() checks
    Note over SubmitWork: _stopping = TRUE<br/>Returns FALSE
    Note over SubmitWork: No new retry starts!

    Kernel-->>ScheduleRetry_A: iocb_Y completes
    ScheduleRetry_A->>Pool: put_one(iocb_Y) Returned
    ScheduleRetry_A->>ScheduleRetry_A: Both vectors empty

    ScheduleRetry_A->>Stop: .finally() block runs
    Stop->>Stop: Waits for outstanding() = 0
    Note over Stop: All iocbs accounted for
    Note over Pool: No double-free!<br/> No leaks!
```

xemul:
> * schedule_retry() A is processing iocbs in _aio_retries

OK, so in the picture, there's iocb_X that gets into the kernel.

> * stop() calls std::exchange(_pending_aio_retry_fut, ready_future)
> 
> * New I/Os complete with -EAGAIN, added to _pending_aio_retry

In the picture, there's iocb_Y that completes with EAGAIN and gets into _pending_aio_retry

> * submit_work() sees need_to_retry()=TRUE && !retry_in_progress() (because we exchanged the future) and starts a NEW schedule_retry() B

In the picture, it says that iocb_Y gets into _aio_retries

> * schedule_retry() A completes, .finally() block runs
> 
> * .finally() calls reap_completions(false) which returns -EAGAIN iocbs to the pool (since allow_retry=false)

In the picture, this place says that iocb_Y completes with EAGAIN. If iocb_Y completed, it means that it was submitted in previous step by schedule_retry() B, is that correct?

> * But schedule_retry() B is still running with those same iocbs in _aio_retries!
> 
> * When B tries to process them and encounters errors, handle_aio_error() calls put_one() again

But why? The schedule_retry() B calls io_submit(iocb_Y) and gets error (picture says it's EBADF). In that case, how could reap_completions(false) see that iocb, if it wasn't submitted by schedule_retry() (it returned EBADF on submission). At the same time, if iocb_Y got to that place, it means that it was submitted by schedule_retry() B without errors.

> * Double-free: attempting to push to a full 1024-capacity free list, causing boost::container::bad_alloc

src/core/reactor_backend.cc
```diff
@@ -115,6 +115,9 @@ aio_storage_context::iocb_pool::iocb_pool() {
     for (unsigned i = 0; i != max_aio; ++i) {
         _free_iocbs.push(&_all_iocbs[i]);
     }
+#ifndef NDEBUG
```
xemul: Please, define SEASTAR_IOCB_POOL_DEBUG, like e.g. SEASTAR_GATE_HOLDER_DEBUG

src/core/reactor_backend.hh
```diff
@@ -64,6 +64,11 @@ class aio_storage_context {
     class iocb_pool {
         alignas(cache_line_size) std::array<internal::linux_abi::iocb, max_aio> _all_iocbs;
         std::stack<internal::linux_abi::iocb*, boost::container::static_vector<internal::linux_abi::iocb*, max_aio>> _free_iocbs;
+#ifndef NDEBUG
+        // Track which iocbs are currently allocated (not in free list)
+        // This helps detect double-free bugs
+        std::array<bool, max_aio> _iocb_allocated;
```
xemul: `std::bitset<max_aio>` maybe?

xemul:
I think I got it. The problem is that two schedule_retry()'s are both submitting each iocb from _aio_retry . So they complete both and are handled both.

If that's correct, the problem is that submit_work() has work to do after aio_storage_context::stop() is called and replaced _aio_retry_fut with ready future from under running retry fiber.  With the current fix it may happen that schedule_retry() A finishes, then submit_work() puts more stuff into _pending_aio_retry, then reap_completions(false) is called, but pending retry is never cleaned.

Maybe the proper fix should be that aio_storage_context::stop() first "makes sure" submit_work() is not called again (or at best -- instantly completes all it has with ESOMETHING), then replaces retry fut with ready-future and starts it waiting?

src/core/reactor_backend.cc
tchaikov: sure. defined.

src/core/reactor_backend.hh
tchaikov: indeed. switched to `std::bitset`.

tchaikov:
> > * schedule_retry() A is processing iocbs in _aio_retries
> 
> OK, so in the picture, there's iocb_X that gets into the kernel.
> 
> > * stop() calls std::exchange(_pending_aio_retry_fut, ready_future)
> > * New I/Os complete with -EAGAIN, added to _pending_aio_retry
> 
> In the picture, there's iocb_Y that completes with EAGAIN and gets into _pending_aio_retry
> 
> > * submit_work() sees need_to_retry()=TRUE && !retry_in_progress() (because we exchanged the future) and starts a NEW schedule_retry() B
> 
> In the picture, it says that iocb_Y gets into _aio_retries
> 
> > * schedule_retry() A completes, .finally() block runs
> > * .finally() calls reap_completions(false) which returns -EAGAIN iocbs to the pool (since allow_retry=false)
> 
> In the picture, this place says that iocb_Y completes with EAGAIN. If iocb_Y completed, it means that it was submitted in previous step by schedule_retry() B, is that correct?
> 
> > * But schedule_retry() B is still running with those same iocbs in _aio_retries!
> > * When B tries to process them and encounters errors, handle_aio_error() calls put_one() again
> 
> But why? The schedule_retry() B calls io_submit(iocb_Y) and gets error (picture says it's EBADF). In that case, how could reap_completions(false) see that iocb, if it wasn't submitted by schedule_retry() (it returned EBADF on submission). At the same time, if iocb_Y got to that place, it means that it was submitted by schedule_retry() B without errors.


  @xemul You're absolutely right to question this  - the specific sequence I described may not be entirely accurate. Let me clarify just to make sure we are now on the same page:

  ## The Actual Race Condition

  The core issue is simpler than my original description: **two concurrent `schedule_retry()` loops can run simultaneously during shutdown**, which leads to shared state corruption in `_aio_retries` and `_pending_aio_retry` vectors.

  Here's how it happens:

  1. **`schedule_retry() A`** is running: it has swapped `_aio_retries` with `_pending_aio_retry` and is in the middle of submitting iocbs via the thread pool
  2. **`stop()` is called**: executes `std::exchange(_pending_aio_retry_fut, make_ready_future<>())`
  3. **`submit_work()` gets called** by the poller (which is still running)
  4. **Line 277 check**: `retry_in_progress()` now returns **false** (because we just exchanged the future), but there are still items in the retry queues
  5. **`schedule_retry() B` starts**: now we have TWO loops both manipulating the same `_aio_retries` and `_pending_aio_retry` vectors

  ## The Double-Free

  The double-free doesn't necessarily happen through the completion path - it happens because:

  - Both loops perform `_aio_retries.erase(...)` operations (line 323)
  - Both loops perform `std::swap(_aio_retries, _pending_aio_retry)` (line 294)
  - These operations on shared state can cause the same iocb to be erased/processed twice
  - When iocbs are "consumed" and erased, they get returned to the pool via `handle_aio_error()` or via the completion handler
  - If both loops think they consumed the same iocb, it gets returned to the pool twice = double-free

> 
> > * Double-free: attempting to push to a full 1024-capacity free list, causing boost::container::bad_alloc

tchaikov:
let me update the time sequence diagram to explain the problem:

```mermaid
  sequenceDiagram
      participant Poller as kernel_submit_work_pollfn
      participant SW as submit_work()                                                                    
      participant SR_A as schedule_retry() A
      participant SR_B as schedule_retry() B     
      participant Pool as iocb_pool
      participant Vectors as _aio_retries / _pending_aio_retry

      Note over SR_A: Already running in thread pool
      SR_A->>Vectors: swap(_aio_retries, _pending_aio_retry)
      SR_A->>SR_A: Submitting iocbs from _aio_retries
                                                    
      Note over Poller: stop() is called on reactor
      Poller->>Poller: _stopping = true       
      Poller->>Poller: exchange(_pending_aio_retry_fut, ready)
                                                    
      Note over Poller: Poller still active!
      Poller->>SW: kernel_submit_work()                                                                  
      SW->>SW: Check: need_to_retry() && !retry_in_progress()
      Note over SW: retry_in_progress() = FALSE<br/>(future was exchanged!)
      Note over SW: need_to_retry() = TRUE<br/>(vectors still have items)
                                                                                                         
      SW->>SR_B: schedule_retry()         
                                                                                                         
      rect rgb(100, 100, 100)
          Note over SR_A,SR_B: RACE: Both loops operating on same vectors!

          par schedule_retry() A                                                                                                                                                                                   
              SR_A->>Vectors: _aio_retries.erase(...)                                          
              SR_A->>Pool: handle_aio_error() -> put_one(iocb_X)                                                                                                                                                   
              Note over Pool: iocb_X returned to pool                                  
          and schedule_retry() B
              SR_B->>Vectors: swap(_aio_retries, _pending_aio_retry)                                                                                                                                               
              SR_B->>Vectors: Process same iocbs
              SR_B->>Vectors: _aio_retries.erase(...)                                                                                                                                                              
              SR_B->>Pool: handle_aio_error() -> put_one(iocb_X)
              Note over Pool: DOUBLE FREE!                                                                                                                                                                         
          end                                  
      end

      Pool->>Pool: ❌ Assertion/abort on double-free
```

src/core/reactor_backend.hh
```diff
@@ -82,9 +92,10 @@ class aio_storage_context {
     pending_aio_retry_t _aio_retries;       // Currently retried iocbs
     future<> _pending_aio_retry_fut = make_ready_future<>();
     internal::linux_abi::io_event _ev_buffer[max_aio];
+    bool _stopping = false;
 
     bool need_to_retry() const noexcept {
-        return !_pending_aio_retry.empty() || !_aio_retries.empty();
+        return !_stopping && (!_pending_aio_retry.empty() || !_aio_retries.empty());
```
xemul: Can _stopping be true here? The submit_work doesn't do anything in that case

src/core/reactor_backend.cc
```diff
@@ -199,6 +230,10 @@ aio_storage_context::handle_aio_error(linux_abi::iocb* iocb, int ec) {
 
 bool
 aio_storage_context::submit_work() {
+    if (_stopping) [[unlikely]] {
+        return false;  // Don't accept new work during shutdown
```
xemul: What if _stopping == true, but there's something in io_sink? Those IOs won't be completed
Maybe something like that:

```c++
void submit_work() {
    ...
    if (need_to_retry()) /* as it is before this patch */ {
        if (_stopping) {
            /* walk _pending_aio_retry and complete all with ESOMETHING */
        } else if (!retry_in_progress()) {
            schedule_retry();
        }
    }
}
```

?

tchaikov:
v3:

- do not check for `_stopping` in `need_to_retry()`
- incorporated Pavel's suggestion in `aio_storage_context::submit_work()`

src/core/reactor_backend.cc
```diff
@@ -131,7 +134,17 @@ aio_storage_context::~aio_storage_context() {
...
 future<> aio_storage_context::stop() noexcept {
+    // Prevent new retries from starting during shutdown
+    _stopping = true;
+
+    // Drain io_sink and complete all pending IOs with error
+    _r._io_sink.drain([] (const internal::io_request& req, io_completion* desc) -> bool {
+        desc->complete_with(-ECANCELED);
+        return true;
+    });
+
```
avikivity, line 168: Q: why are we even here? If we're shutting down the reactor, nothing should be in progress.

tchaikov (commented):
v4:

- dropped the defensive draining logic in `aio_storage_context::stop()`
- guarded the assert of `_free_iocbs.size() < max_aio` in `aio_storage_context::iocb_pool::put_one()` with `SEASTAR_IOCB_POOL_DEBUG`, as it's in the hot path.

src/core/reactor_backend.cc
tchaikov: it was a defensive draining approach, which intended to address the misbehaved applications which fail to drain all in-progress operations. but yes, it seems we don't tolerate this. so i am removing it.

xemul: The problem this PR fixes happens when there's something in submit_work() after aio_storage_context::stop() has been called.

Still I think it's not great to put such ifs in submit_work() main path. All dangling iocbs should be collected by aio_storage_context::stop() and submit_work() shouldn't put more load into the stopped/ing retry fiber (https://github.com/scylladb/seastar/pull/3182#discussion_r2647744192)

src/core/reactor_backend.cc
```diff
@@ -199,6 +238,10 @@ aio_storage_context::handle_aio_error(linux_abi::iocb* iocb, int ec) {
 
 bool
 aio_storage_context::submit_work() {
+    if (_stopping) [[unlikely]] {
```
xemul: Let's allow `submit_work()` do whatever it prefers and make `aio_storage_context::stop()` collect whatever appears from it. At most, add extra activity under existing `need_retry()` check.

Maybe something like that would work? (I'm not sure that putting default-initialized future into `_pending_aio_retry_fut` is good)

```patch
diff --git a/src/core/reactor_backend.cc b/src/core/reactor_backend.cc
index 725f0daf6..4e6dc20bf 100644
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -131,9 +131,10 @@ aio_storage_context::~aio_storage_context() {
 }
 
 future<> aio_storage_context::stop() noexcept {
-    return std::exchange(_pending_aio_retry_fut, make_ready_future<>()).finally([this] {
+    return std::exchange(_pending_aio_retry_fut, future<>() /* not future::available() future */).finally([this] {
         return do_until([this] { return !_iocb_pool.outstanding(); }, [this] {
             reap_completions(false);
+            reap_pending_retries();
             return make_ready_future<>();
         });
     });
@@ -243,7 +244,7 @@ aio_storage_context::submit_work() {
         did_work = true;
     }
 
-    if (need_to_retry() && !retry_in_progress()) {
+    if (need_to_retry()) {
         schedule_retry();
     }
 
@@ -251,6 +252,10 @@ aio_storage_context::submit_work() {
 }
 
 void aio_storage_context::schedule_retry() {
+    if (!_pending_aio_retry_fut.available()) {
+        return;
+    }
+
     // loop until both _pending_aio_retry and _aio_retries are empty.
     // While retrying _aio_retries, new retries may be queued onto _pending_aio_retry.
     _pending_aio_retry_fut = do_until([this] {
@@ -317,6 +322,16 @@ bool aio_storage_context::reap_completions(bool allow_retry)
     return n;
 }
 
+void aio_storage_context::reap_pending_retries()
+{
+    for (auto* iocb : _pending_aio_retry) {
+        auto desc = get_user_data<kernel_completion>(*iocb);
+        desc->complete_with(-ECANCELED);
+        _iocb_pool.put_one(iocb);
+    }
+    _pending_aio_retry.clear();
+}
+
 bool aio_storage_context::can_sleep() const {
     // Because aio depends on polling, it cannot generate events to wake us up, Therefore, sleep
     // is only possible if there are no in-flight aios. If there are, we need to keep polling.
diff --git a/src/core/reactor_backend.hh b/src/core/reactor_backend.hh
index aa240b8c6..3d4efb9c0 100644
--- a/src/core/reactor_backend.hh
+++ b/src/core/reactor_backend.hh
@@ -87,9 +87,7 @@ class aio_storage_context {
         return !_pending_aio_retry.empty() || !_aio_retries.empty();
     }
 
-    bool retry_in_progress() const noexcept {
-        return !_pending_aio_retry_fut.available();
-    }
+    void reap_pending_retries();
 
 public:
     explicit aio_storage_context(reactor& r);

```

tchaikov: Thanks Pavel for the suggested patch! This implementation is largely based on your approach, with a few differences:

1. `future<>()` vs `promise<>.get_future()`: because `future<>()` doesn't have a public default constructor. . The only ways to create futures are:
    - `make_ready_future<>()` - available immediately
    - `promise<>.get_future()` - available when promise is fulfilled
    - `make_exception_future<>()` - available with exception

2. single `reap_pending_retries()` vs repeated draining
      The reasoning: after exchange, no new items can enter _pending_aio_retry because:                                                                                                                                                                                                                    
   1. io_sink already drained → no new iocbs allocated
   2. schedule_retry() returns early → no new retry loops                                                                                           
   3. reap_completions(false) doesn't add to retry queue            
                                                                         
   So calling it once is sufficient by design.
   
3. .finally() vs .then(): The difference: .then() waits for the in-flight retry loop to complete successfully before draining. With .finally(), if the old future fails, we'd still proceed with cleanup - but the old retry loop should always complete normally (it's a do_until that exits when vectors are empty).

avikivity:
> The existing retry loop completes normally, and stop()'s .finally() block simply waits for all in-flight iocbs to be reaped and returned. No iocbs are leaked, and no double-free can occur.

This means the application leaked futures, no?

In general, if the application leaks work-in-progress, we can't recover.

xemul:
Maybe the race is that both, app_template::run() and aio_storage_context::aio_storage_context() both schedule at_exit callbacks? The run's exit happens after aio context's one, so by the time user func is trying to collect all its futures, the aio context's cleanup had been executed already and _pending_aio_retry_fut had been exchanged with dummy ready one thus opening the window for two concurrent retry fibers

emdroid:
Hi, what is the status of this fix? This seems to be fixing a recurring CI-stability issue:
* [SCYLLADB-302](https://scylladb.atlassian.net/browse/SCYLLADB-302)
* [SCYLLADB-1530](https://scylladb.atlassian.net/browse/SCYLLADB-1530) , [SCYLLADB-1669](https://scylladb.atlassian.net/browse/SCYLLADB-1669), [SCYLLADB-1866](https://scylladb.atlassian.net/browse/SCYLLADB-1866) (new duplicates)

[SCYLLADB-302]: https://scylladb.atlassian.net/browse/SCYLLADB-302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SCYLLADB-1530]: https://scylladb.atlassian.net/browse/SCYLLADB-1530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SCYLLADB-1669]: https://scylladb.atlassian.net/browse/SCYLLADB-1669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SCYLLADB-1866]: https://scylladb.atlassian.net/browse/SCYLLADB-1866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

tchaikov:
@emdroid  Thanks for the ping. i just reworked the change and rewrote the commit message in hope to address the questions from the reviewers.

v5:

- while re-tracing the code paths, i also found that the earlier version left a follow-on shutdown hang under `--kernel-page-cache` (every iocb is routed through `_pending_aio_retry` there, and once `schedule_retry()` is gated during shutdown nothing drains it), so this revision closes that too.

xemul:
@tchaikov , [there's](https://github.com/scylladb/seastar/pull/3182#issuecomment-4262551874) still a concern

> This means the application leaked futures, no?

I partially [second](https://github.com/scylladb/seastar/pull/3182#issuecomment-4278000930) one: maybe the problem is that aio_storage_context::stop() should be called later?

src/core/reactor_backend.cc
```diff
@@ -130,8 +133,45 @@ aio_storage_context::~aio_storage_context() {
...
+    });
+
+    // Wait for any in-flight retry loop to finish, then drain remaining items.
+    // Use then_wrapped so the drain runs even if the retry loop future holds
+    // an exception.
+    return std::exchange(_pending_aio_retry_fut, make_ready_future<>()).then_wrapped([this] (future<> f) noexcept {
+        f.ignore_ready_future();
+        reap_pending_retries();
         return do_until([this] { return !_iocb_pool.outstanding(); }, [this] {
             reap_completions(false);
```
xemul: This `false` seem not to be needed any longer. The `_stopping` is set, so reap_completions() would disable retries on its own

src/core/reactor_backend.cc
```diff
@@ -201,6 +258,22 @@ bool
 aio_storage_context::submit_work() {
     bool did_work = false;
 
+    // During shutdown, cancel any new I/O instead of allocating iocbs.
```
xemul: Error handling at the beginning of the happy path is weird. What if we process normally and do submit these IOs while _stopping == true? Isn't there already a code that suggests to pick it up? Or this path is not supposed to happen on stop when _stopping == true?

src/core/reactor_backend.cc
```diff
@@ -130,8 +133,45 @@ aio_storage_context::~aio_storage_context() {
...
+    _r._io_sink.drain([] (const internal::io_request& req, io_completion* desc) -> bool {
+        desc->complete_with(-ECANCELED);
+        return true;
+    });
+
+    // Wait for any in-flight retry loop to finish, then drain remaining items.
+    // Use then_wrapped so the drain runs even if the retry loop future holds
+    // an exception.
+    return std::exchange(_pending_aio_retry_fut, make_ready_future<>()).then_wrapped([this] (future<> f) noexcept {
+        f.ignore_ready_future();
```
xemul: Why is it necessary to replace .finally() with .then_wrapped + ignore_ready_future ?

mykaul:
@tchaikov - hi - are you still working on this?
@emdroid - if the answer to the above is negative, please own it.

src/core/reactor_backend.cc
tchaikov: ahh, yes. it's a leftover of an experiment. it's not necessary. reverted.

src/core/reactor_backend.cc
tchaikov: indeed. dropped.

src/core/reactor_backend.cc
tchaikov: yup, dropped.

tchaikov:
> @tchaikov , [there's](https://github.com/scylladb/seastar/pull/3182#issuecomment-4262551874) still a concern
> 
> > This means the application leaked futures, no?
> 
> I partially [second](https://github.com/scylladb/seastar/pull/3182#issuecomment-4278000930) one: maybe the problem is that aio_storage_context::stop() should be called later?

i am not sure if i am following you, but `aio_storage_context::stop()` is actually already called last, because the AIO context registers it with `_r.do_at_exit()`. since `run_exit_tasks()` iterates that list in reverse, `stop()` fires after all application cleanup.

tchaikov:
v6:

- addressed @xemul's comments.

tchaikov:
> @tchaikov - hi - are you still working on this? @emdroid - if the answer to the above is negative, please own it.

still on it. just pushed v6 addressing Pavel's comments. happy for anyone to pile on.

src/core/reactor_backend.cc
```diff
@@ -130,10 +133,43 @@ aio_storage_context::~aio_storage_context() {
...
+
+    // _aio_retries is empty in the normal call path: this function only runs
+    // after the retry loop's future has resolved, and that loop's predicate
+    // only returns true when both vectors are empty. The drain below is
+    // defensive — if a future change to schedule_retry() leaves entries
+    // here, we still avoid leaking iocbs.
+    for (auto iocb : _aio_retries) {
+        _iocb_pool.put_one(iocb);
+        auto desc = get_user_data<kernel_completion>(*iocb);
+        desc->complete_with(-ECANCELED);
```
avikivity, line 153: This will probably unleash a bunch of latent bugs, because most code will be unprepared for these errors.

It might be better to let them complete naturally.

We can try with cancellation first, it won't be worse than the crash.

src/core/reactor_backend.cc
```diff
@@ -243,14 +296,23 @@ aio_storage_context::submit_work() {
...
 }
 
 void aio_storage_context::schedule_retry() {
+    // Don't start a new retry loop if either:
+    //   (a) one is already in flight  (!_pending_aio_retry_fut.available()),
+    //   (b) we're shutting down       (_stopping) — stop() will drain the
+    //       queues itself; starting a second loop here would race with
+    //       stop()'s drain on the same iocb pool (the original double-free).
+    if (!_pending_aio_retry_fut.available() || _stopping) {
+        return;
```
avikivity, line 313: What if it's not available now, but is in the process of becoming available?

avikivity: For example, do_until was preempted and is waiting to be scheduled. Might not actually be possible, since it will re-evaluate the condition. Still, it's not nice to depend on such details.

avikivity: But, I see it's a preexisting problem.

More modern is to start a retry fiber in the constructor, wake it each time something needs to be done, and collect it in stop().

avikivity (approved):
Looks good but will let @xemul make the final call.

src/core/reactor_backend.cc
```diff
@@ -130,10 +133,43 @@ aio_storage_context::~aio_storage_context() {
...
+    // after the retry loop's future has resolved, and that loop's predicate
+    // only returns true when both vectors are empty. The drain below is
+    // defensive — if a future change to schedule_retry() leaves entries
+    // here, we still avoid leaking iocbs.
+    for (auto iocb : _aio_retries) {
+        _iocb_pool.put_one(iocb);
+        auto desc = get_user_data<kernel_completion>(*iocb);
+        desc->complete_with(-ECANCELED);
+    }
+    _aio_retries.clear();
```
bhalevy, line 155: nit: since the logic for reaping `_pending_aio_retry` and `_aio_retries` is identical, can we implement it in a shared function, called for each queue?

src/core/reactor_backend.cc
```diff
@@ -296,6 +358,11 @@ void aio_storage_context::schedule_retry() {
 
 bool aio_storage_context::reap_completions(bool allow_retry)
 {
+    // During shutdown, complete EAGAIN iocbs immediately instead of
+    // queuing them for retry — schedule_retry() is gated on _stopping.
+    if (_stopping) {
+        allow_retry = false;
```
bhalevy, line 364: nit: `allow_retry &= !_stopping` rather than adding a conditional

src/core/reactor_backend.cc
bhalevy: Agree with the above.
With a fiber design, the cancellation of the retried iocb's can be done by the retry fiber when it sees `_stopping` rather than special handled in the stop() path.
Maybe it's a good opportunity to do that in the scope of this change.

src/core/reactor_backend.cc
tchaikov: created #3406

src/core/reactor_backend.cc
tchaikov: it's not in the hot paths, so i'd prefer trading a condition statement for better readability.

bhalevy: I'm working on a cleanup series that addresses this in even a simpler way.
I'll post a link once pushed

src/core/reactor_backend.cc
bhalevy: See https://github.com/scylladb/seastar/pull/3408

src/core/reactor_backend.cc
bhalevy: See https://github.com/scylladb/seastar/pull/3408

src/core/reactor_backend.cc
bhalevy: See https://github.com/scylladb/seastar/pull/3408 - the allow_retry parameter is removed there altogether and when _stopping we cancel the iocb in the same way it is handled by the retry_fiber so we don't end up with 2 different states of canceled retries on shutdown

Issue #3180, opened 2025-12-25
scheduling_group_nesting fails


tchaikov:
```
63/87 Test #64: Seastar.unit.scheduling_group_nesting .........***Failed    3.13 sec
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: invalid mtime
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
ninja: corrupt build log: missing field
...
random-seed=775790339
Creating supergroups
Creating groups
Creating supergroups
Creating groups
=== TEST-0 ===
Running in 3 groups: [8, 4, 6]
Shares: [4, 5, 4]
  supergroup 1 shares = 4
  supergroup 2 shares = 5
--------8<--------
0: count=7053 expected=0.28 adjusted=25390 deviation=0.00
1: count=11293 expected=0.44 adjusted=25409 deviation=0.00
2: count=7029 expected=0.28 adjusted=25304 deviation=0.00
=== TEST-1 ===
Running in 3 groups: [2, 4, 0]
Shares: [1, 3, 1]
  supergroup 1 shares = 4
--------8<--------
0: count=4263 expected=0.17 adjusted=25578 deviation=0.01
1: count=16943 expected=0.67 adjusted=25414 deviation=0.00
2: count=4194 expected=0.17 adjusted=25164 deviation=0.01
=== TEST-2 ===
Running in 4 groups: [1, 3, 2, 5]
Shares: [5, 4, 1, 5]
  supergroup 1 shares = 3
--------8<--------
0: count=19497 expected=0.56 adjusted=35094 deviation=0.00
1: count=5225 expected=0.15 adjusted=35268 deviation=0.00
2: count=3919 expected=0.11 adjusted=35271 deviation=0.00
3: count=6559 expected=0.19 adjusted=35418 deviation=0.00
=== TEST-3 ===
Running in 4 groups: [6, 8, 0, 4]
Shares: [3, 2, 2, 2]
  supergroup 1 shares = 5
  supergroup 2 shares = 1
--------8<--------
0: count=2158 expected=0.08 adjusted=28773 deviation=0.08
/__w/seastar/seastar/tests/unit/scheduling_group_nesting_test.cc(132): error: in "test_fairness": check dev < 0.07 has failed
1: count=1358 expected=0.05 adjusted=27160 deviation=0.02
[...]

PR #3184, opened 2025-12-28
json: formatter: fix formatting of std:string_view

### LINKED ISSUE #3183: std::string_view is wrongly formatter by to_json as array of integers rather than a json string
For example: std::string_view("hello, world") is wrongly formatted as [104,101,108,108,111,44,32,119,111,114,108,100]

bhalevy:
Currently, to_json of a std::string_view resolved to
the Range overload, and so it is printed as array of integers
rather than as a string.

For example: std::string_view("hello, world") is wrongly formatted
as [104,101,108,108,111,44,32,119,111,114,108,100]

Note that we add formatting both for std::string_view and for std::string.
The latter is required to resolve ambiguities with formatting of
seastar::sstring (when Seastar_SSTRING is ON).

Fixes #3183

### DIFF
--- a/include/seastar/json/formatter.hh
+++ b/include/seastar/json/formatter.hh
@@ -53,7 +53,7 @@ template<typename T>
 concept is_string_like =
     std::convertible_to<const T&, std::string_view> &&
     requires (T s) {
-        { s.data() } -> std::same_as<char*>;
+        { s.data() } -> std::convertible_to<const char*>;
         // sstring::length() and sstring::find() return size_t instead of
         // size_type (i.e., uint32_t), so we cannot check their return type
         // with T::size_type
--- a/include/seastar/json/formatter.hh
+++ b/include/seastar/json/formatter.hh
@@ -177,10 +177,10 @@ public:
 
     /**
      * return a json formatted string
-     * @param str the string to format
+     * @param str the string_view to format
      * @return the given string in a json format
      */
-    static sstring to_json(const sstring& str);
+    static sstring to_json(std::string_view str);
 
     /**
      * return a json formatted int
@@ -276,14 +276,12 @@ public:
      */
     static sstring to_json(unsigned long l);
 
-
-
     /**
      * return a json formatted string
-     * @param str the string to format
+     * @param str the string_view to format
      * @return the given string in a json format
      */
-    static future<> write(output_stream<char>& s, const sstring& str) {
+    static future<> write(output_stream<char>& s, std::string_view str) {
         return s.write(to_json(str));
     }
 
--- a/src/json/formatter.cc
+++ b/src/json/formatter.cc
@@ -113,7 +113,7 @@ static sstring string_view_to_json(const string_view& str) {
     return oss.str();
 }
 
-sstring formatter::to_json(const sstring& str) {
+sstring formatter::to_json(std::string_view str) {
     return string_view_to_json(str);
 }
 
--- a/tests/unit/json_formatter_test.cc
+++ b/tests/unit/json_formatter_test.cc
@@ -74,6 +74,17 @@ SEASTAR_TEST_CASE(test_ranges) {
     return make_ready_future();
 }
 
+SEASTAR_TEST_CASE(test_strings) {
+    sstring s = "hello, world";
+    const char* expected = "\"hello, world\"";
+    BOOST_CHECK_EQUAL(expected, formatter::to_json(s));
+    BOOST_CHECK_EQUAL(expected, formatter::to_json(std::string(s)));
+    BOOST_CHECK_EQUAL(expected, formatter::to_json(std::string_view(s)));
+    BOOST_CHECK_EQUAL(expected, formatter::to_json(s.c_str()));
+
+    return make_ready_future();
+}
+
 struct object_json : public json_base {
     json_element<sstring> subject;
     json_list<long> values;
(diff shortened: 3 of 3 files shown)


bhalevy:
[d344d5c](https://github.com/scylladb/seastar/commit/d344d5cb2706a573c0a748b97ed959787578086d): simplified, left only formatting for std::string_view since both std::string and seastar::sstring are implicitly convertible to std::string_view.
Doing so also saves consideration for Seastar_STRING build config option.

bhalevy:
Hit https://github.com/scylladb/seastar/issues/3180

include/seastar/json/formatter.hh
```diff
@@ -53,7 +53,7 @@ template<typename T>
 concept is_string_like =
     std::convertible_to<const T&, std::string_view> &&
     requires (T s) {
-        { s.data() } -> std::same_as<char*>;
+        { s.data() } -> std::convertible_to<const char*>;
         // sstring::length() and sstring::find() return size_t instead of
```
avikivity, line 57: Perhaps a better approach is to replace is_string_like<T> with std::convertible_to<std::string_view, T>. Both sstring and std::string are convertible to std::string_view. is_string_like likely predates std::string_view.

But let's merge this first as the simplest fix.

travisdowns:
Heads up: this is a breaking API change for downstream users. Types that have a user-defined conversion to `sstring` (but not to `std::string_view`) can no longer be passed to `formatter::to_json` / `formatter::write`.

Previously, `to_json(x)` worked when `x` had `operator sstring()` — one user-defined conversion bound it to `const sstring&`. After this change, the chain `x → sstring → string_view` requires two user-defined conversions, which C++ doesn't allow in implicit conversion sequences. Such types now fail to compile (and they don't satisfy the range overload either, unless they happen to expose `begin()/end()`).

Hit this in Redpanda where our `uuid_t` has `operator sstring() const` and was relied upon for JSON serialization. Might be worth keeping a `const sstring&` overload alongside the new `string_view` one to preserve the previous behavior.

bhalevy:
> Heads up: this is a breaking API change for downstream users. Types that have a user-defined conversion to `sstring` (but not to `std::string_view`) can no longer be passed to `formatter::to_json` / `formatter::write`.
> 
> Previously, `to_json(x)` worked when `x` had `operator sstring()` — one user-defined conversion bound it to `const sstring&`. After this change, the chain `x → sstring → string_view` requires two user-defined conversions, which C++ doesn't allow in implicit conversion sequences. Such types now fail to compile (and they don't satisfy the range overload either, unless they happen to expose `begin()/end()`).
> 
> Hit this in Redpanda where our `uuid_t` has `operator sstring() const` and was relied upon for JSON serialization. Might be worth keeping a `const sstring&` overload alongside the new `string_view` one to preserve the previous behavior.

Can you please open a new issue about this with a concrete reproducer test?

bhalevy:
> > Heads up: this is a breaking API change for downstream users. Types that have a user-defined conversion to `sstring` (but not to `std::string_view`) can no longer be passed to `formatter::to_json` / `formatter::write`.
> > Previously, `to_json(x)` worked when `x` had `operator sstring()` — one user-defined conversion bound it to `const sstring&`. After this change, the chain `x → sstring → string_view` requires two user-defined conversions, which C++ doesn't allow in implicit conversion sequences. Such types now fail to compile (and they don't satisfy the range overload either, unless they happen to expose `begin()/end()`).
> > Hit this in Redpanda where our `uuid_t` has `operator sstring() const` and was relied upon for JSON serialization. Might be worth keeping a `const sstring&` overload alongside the new `string_view` one to preserve the previous behavior.
> 
> Can you please open a new issue about this with a concrete reproducer test?

See https://github.com/scylladb/seastar/pull/3389

avikivity:
> Heads up: this is a breaking API change for downstream users. Types that have a user-defined conversion to `sstring` (but not to `std::string_view`) can no longer be passed to `formatter::to_json` / `formatter::write`.
> 
> Previously, `to_json(x)` worked when `x` had `operator sstring()` — one user-defined conversion bound it to `const sstring&`. After this change, the chain `x → sstring → string_view` requires two user-defined conversions, which C++ doesn't allow in implicit conversion sequences. Such types now fail to compile (and they don't satisfy the range overload either, unless they happen to expose `begin()/end()`).
> 
> Hit this in Redpanda where our `uuid_t` has `operator sstring() const` and was relied upon for JSON serialization. Might be worth keeping a `const sstring&` overload alongside the new `string_view` one to preserve the previous behavior.

I don't think a C++ project can guarantee this level of source compatibility.

I try very hard to keep compatibility, but the maze of conversions can defeat any plan.

travisdowns:
> Can you please open a new issue about this with a concrete reproducer test?

Based on your PR, I think you got it: it was an object with a conversion to `sstring` but not to `string_view` and double conversion is not allowed.

travisdowns:
> I don't think a C++ project can guarantee this level of source compatibility.
> 
> I try very hard to keep compatibility, but the maze of conversions can defeat any plan.

yeah I this wasn't meant to be a major complaint or even a request to fix (already fixed on RP side), just a note and also a bit of assessment of how deep the source compatibility should go

avikivity:
> > I don't think a C++ project can guarantee this level of source compatibility.
> > I try very hard to keep compatibility, but the maze of conversions can defeat any plan.
> 
> yeah I this wasn't meant to be a major complaint or even a request to fix (already fixed on RP side), just a note and also a bit of assessment of how deep the source compatibility should go

As deep as we can, but no deeper.

PR #1601, opened 2023-04-06
treewide: drop the support of fmtlib < 8.0.0

tchaikov:
in the light of requirement of newer compiler versions, and distros providing these compiler versions also package newer fmtlib, we can now able to assume that developers should have the access to the fmtlib > 8.1.1 if they are using the compiler(s) shipped by distros. so, to have a cleaner code base and to improve the maintainability, in this series, we bump up the minimal required version of fmtlib from 5.0.0 to 8.1.1, also the code supporting fmtlib < 8.0 is removed.

PR #3185, opened 2025-12-29
treewide: drop the support of fmtlib < 9.0.0

tchaikov:
Since we've bumped up the minimum supported fmtlib version to 9.0.0, we now have the luxury of dropping the support of fmtlib < 9.0.0 for a cleaner code base.

In this change, the support of fmtlib < 9.0.0 is dropped by removing all #if FMT_VERSION >= 90000 guards.

Notable features available in fmt 9.0.0:
- Improved floating-point formatting with consistent cross-platform rounding
- std::variant and std::filesystem::path formatting support
- fmt::ostream_formatter for custom formatters
- ADL-based format_as extension API for enums
- Nested format specifiers for ranges (e.g., {:#x})

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -387,7 +387,7 @@ find_package (Valgrind REQUIRED)
 
 cmake_dependent_option (Seastar_LOGGER_COMPILE_TIME_FMT
   "Enable the compile-time {fmt} check when formatting logging messages" ON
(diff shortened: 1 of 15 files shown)


tchaikov:
please note, fmt 9.0.0 was released three years ago, see https://github.com/fmtlib/fmt/releases/tag/9.0.0

see also #1601, where we dropped the support of fmtlib < 8.0.0

PR #3186, opened 2025-12-29
file: Use query_device_alignment_info in blkdev_alignments 

tchaikov:
Refactor `blkdev_alignments` to use `query_device_alignment_info()` helper
function instead of duplicating the logic for querying statx memory
alignment and `physical_block_size` override from io-queue.

This change:
- Eliminates duplicate calls to query_statx_mem_align()
- Removes manual `engine().get_io_queue()` and `io_queue.physical_block_size()` check
- Makes `blkdev_alignments` consistent with `filesystem_alignments()`

Signed-off-by: Kefu Chai <k.chai@proxmox.com>

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -1051,10 +1051,11 @@ xfs_concurrency_from_kernel_version() {
     return 0;
 }
 
+namespace {
+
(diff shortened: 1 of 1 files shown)


tchaikov:
@xemul could you please take a look?

PR #3188, opened 2025-12-30
test: Extend statat() test to check ENOENT exception reporting

xemul:
When calling file.statat("non-existing-file") the method should resolve into exceptional future with ENOENT on board.


xemul:
upd:
- use BOOST_REQUIRE_EXCEPTION

PR #3191, opened 2026-01-05
net: expose DNS TTL via net::hostent

kreuzerkrieg:
Add vector of integers field to net::hostent to carry the DNS TTL, and populate it when resolving addresses through `ares`.

### DIFF
--- a/include/seastar/net/dns.hh
+++ b/include/seastar/net/dns.hh
@@ -55,6 +55,7 @@ struct hostent {
     std::vector<sstring> names;
     // Primary address is also always first.
     std::vector<inet_address> addr_list;
+    std::vector<unsigned int> addr_ttls;
 };
 
 typedef std::optional<inet_address::family> opt_family;
--- a/tests/unit/dns_test.cc
+++ b/tests/unit/dns_test.cc
@@ -41,6 +41,10 @@ static future<> test_resolve(dns_resolver::options opts) {
 
     for (auto hostname : {"seastar.io", "scylladb.com", "kernel.org", "www.google.com"}) {
         hostent e = co_await d.get_host_by_name(hostname, inet_address::family::INET);
+        for (auto ttl: e.addr_ttls) {
+            BOOST_REQUIRE(ttl > 0);
+        }
+
         hostent a;
         try {
             a = co_await d.get_host_by_addr(e.addr_list.front());
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -1002,12 +1002,17 @@ dns_resolver::impl::make_hostent(const ares_addrinfo* ai) {
         e.names.emplace_back(cname->alias);
     }
     for (auto node = ai->nodes; node != nullptr; node = node->ai_next) {
+        // The TTL can be zero (dont cache) or greater up to 2^31 - 1 (in seconds)
+        // https://datatracker.ietf.org/doc/html/rfc2181#section-8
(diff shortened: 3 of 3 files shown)


tests/unit/dns_test.cc
```diff
@@ -41,6 +41,10 @@ static future<> test_resolve(dns_resolver::options opts) {
 
     for (auto hostname : {"seastar.io", "scylladb.com", "kernel.org", "www.google.com"}) {
         hostent e = co_await d.get_host_by_name(hostname, inet_address::family::INET);
+        for (auto ttl: e.addr_ttls) {
+            BOOST_REQUIRE(ttl > 0);
+        }
```
avikivity, line 46: I suppose it could be 0, no?

include/seastar/net/dns.hh
```diff
@@ -55,6 +55,7 @@ struct hostent {
     std::vector<sstring> names;
     // Primary address is also always first.
     std::vector<inet_address> addr_list;
+    std::vector<unsigned int> addr_ttls;
```
xemul, line 58: This makes iterating over the returned hostent data being a terrible experience. One cannot do

```
    for (const auto& info : hostent.addresses) {
        do_something_with(info.address, info.ttl);
    }
```

instead, it should look something like 

```
    for (unsigned i = 0; i < hostent.addr_list.size(); i++) {
        do_something_with(hostent.addr_list[i], hostent.addr_ttls[i]);
    }
```

or implement it in a coarse manner like in https://github.com/scylladb/scylladb/pull/27891/commits/c19bc4cfbfee4a4e752d0dc64348804b6eed43b2 -- by getting `std::chrono::seconds(*std::ranges::min_element(hent.addr_ttls));` for all addresses in the list.

But that's not the only problem here. In `get_host_by_name()` there's such a code:

```
    if (!family) {
        auto res = inet_address::parse_numerical(name);
        if (res) {
            return make_ready_future<hostent>(hostent{ {name}, {*res}});
        }
    }
```

that generates a hostent with one address and zero ttls, which makes iterations horrible. None of the above code works. The former for-loop just doesn't, the latter for-loop can no longer assume that it can do `hostent.addr_ttls[i]` for any given `i`, and the `*std::ranges::min_element()` just dereferences an iterator pointing to nop.

xemul: Also it's seconds, but nobody is informed about it by neither a code-comment, nor by using std::chrono::second type, nor by the member name (addr_ttls_sec?)

xemul:
I reverted this change as per above concerns and suggest to come up with some better implementation.
Ideally it should be a vector of struct-s containing addres and ttls (with the ability to extend in the future), and the older vector is to be deprecated in favor of the new one. And also giving ttl members a std::chrono::second (or at least std::chrono::duration) type is very welcome.

kreuzerkrieg:
> Ideally it should be a vector of struct

this is what I did initially but later reverted the change due to backward compatibility concerns

tests/unit/dns_test.cc
kreuzerkrieg: As I understand from the RFC, the TTL value 0 is very specific case when the remote machine wants to signal you that you must resolve the address for each request. Usually this is not the case.
>Zero are interpreted to mean that the RR can only be for the transaction in progress, and should not be.  For example, SOA records are always distributed a zero TTL to prohibit caching.  Zero values can be used for extremely volatile data.

PR #3199, opened 2026-01-13
net: Add DNS TTL to the net::hostent

kreuzerkrieg:
- Introduce new struct to hold the resolved address and its TTL
 - Add new data member to hold a list of above structures
 - Deprecate the address vector

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -1618,20 +1618,22 @@ future<inet_address> inet_address::find(
     return dns::resolve_name(name, f);
 }
 
-future<std::vector<inet_address>> inet_address::find_all(
-                const sstring& name) {
-    return dns::get_host_by_name(name).then([](hostent e) {
+future<std::vector<inet_address>> inet_address::find_all_impl(
+                const sstring& name, opt_family f) {
+    return dns::get_host_by_name(name, f).then([](hostent e) {
         auto rng = e.addr_entries | std::views::transform([](auto& entry) { return entry.addr; });
         return make_ready_future<std::vector<inet_address>>(rng.begin(), rng.end());
     });
 }
 
 future<std::vector<inet_address>> inet_address::find_all(
+                const sstring& name) {
+    return inet_address::find_all_impl(name, std::nullopt);
+}
+
+future<std::vector<inet_address>> inet_address::find_all(
                 const sstring& name, family f) {
-    return dns::get_host_by_name(name, f).then([](hostent e) {
-        auto rng = e.addr_entries | std::views::transform([](auto& entry) { return entry.addr; });
-        return make_ready_future<std::vector<inet_address>>(rng.begin(), rng.end());
-    });
+    return inet_address::find_all_impl(name, std::make_optional(f));
 }
 
 const std::error_category& dns::error_category() {
--- a/include/seastar/net/inet_address.hh
+++ b/include/seastar/net/inet_address.hh
@@ -57,6 +57,8 @@ private:
     };
 
     uint32_t _scope = invalid_scope;
+
+    static future<std::vector<inet_address>> find_all_impl(const sstring& name, std::optional<family> family);
 public:
     static constexpr uint32_t invalid_scope = std::numeric_limits<uint32_t>::max();
 


include/seastar/net/dns.hh
```diff
@@ -55,6 +55,15 @@ struct hostent {
     std::vector<sstring> names;
     // Primary address is also always first.
     std::vector<inet_address> addr_list;
+    struct addr_entry
+    {
+        inet_address addr;
+        // https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.4
+        // https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.3
+        // https://datatracker.ietf.org/doc/html/rfc2181#section-8
+        unsigned int ttl_seconds{std::numeric_limits<signed int>::max()};
```
xemul: Please, use std::chrono::seconds or std::chrono::duration

src/net/dns.cc
```diff
@@ -1031,18 +1036,20 @@ dns_resolver::impl::make_hostent(const ::hostent& host) {
         case AF_INET:
             SEASTAR_ASSERT(size_t(host.h_length) >= sizeof(in_addr));
             e.addr_list.emplace_back(*reinterpret_cast<const in_addr*>(*p));
+            e.addr_entries.emplace_back(hostent::addr_entry{*reinterpret_cast<const in_addr*>(*p), std::numeric_limits<signed int>::max()});
```
xemul: Whoa, [last time](https://github.com/scylladb/seastar/pull/3191/commits/fdd2f75e3a4c347cb731ea4dbc4f21dcb6326334)  only make_hostent(const ares_addrinfo* ai) got patches, so one more "bug" spotted?

include/seastar/net/dns.hh
```diff
@@ -64,6 +64,24 @@ struct hostent {
...
+
+    // Remove the whole section below once we drop `addr_list`
+    // from the struct.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    hostent() = default;
+    hostent(std::vector<sstring>&& _names, std::vector<inet_address>&& _addr_list, std::vector<addr_entry>&& _addr_entries):
+        names(std::move(_names)),
+        addr_list(std::move(_addr_list)),
+        addr_entries(std::move(_addr_entries)){
```
xemul: Please, format it like

```
hostent(_a, _b, _c)
    : a(_a)
    , b(_b)
    , c(_c)
{}
```

src/net/dns.cc
```diff
@@ -532,7 +532,7 @@ dns_resolver::impl::get_host_by_name(sstring name, opt_family family)  {
     if (!family) {
         auto res = inet_address::parse_numerical(name);
         if (res) {
-            return make_ready_future<hostent>(hostent{ {name}, {*res}, {{*res}}});
+            return make_ready_future<hostent>(hostent{{name}, {{*res}}, {{*res}}});
```
xemul: To reduce the curlybracyness of this line, let's use `hostent(name, {*res}, {*res})` construction, I noticed that you added the relevant constructor for that to work.

Also std::move()-ing arguments seem to worth it

src/net/dns.cc
```diff
@@ -1605,14 +1611,16 @@ future<inet_address> inet_address::find(
 future<std::vector<inet_address>> inet_address::find_all(
                 const sstring& name) {
     return dns::get_host_by_name(name).then([](hostent e) {
-        return make_ready_future<std::vector<inet_address>>(std::move(e.addr_list));
+        auto rng = e.addr_entries | std::views::transform([](auto& entry) { return entry.addr; });
+        return make_ready_future<std::vector<inet_address>>(rng.begin(), rng.end());
     });
 }
 
 future<std::vector<inet_address>> inet_address::find_all(
```
xemul, line 1634: Probably worth squashing two find_all()-s into a single one with defaulted-argument (as a followup is OK)

tests/unit/dns_test.cc
```diff
@@ -41,23 +41,23 @@ static future<> test_resolve(dns_resolver::options opts) {
 
     for (auto hostname : {"seastar.io", "scylladb.com", "kernel.org", "www.google.com"}) {
         hostent e = co_await d.get_host_by_name(hostname, inet_address::family::INET);
-        BOOST_REQUIRE_EQUAL(e.addr_list.size(), e.addr_entries.size());
```
xemul: I'd keep testing that deprecated addr_list is in sync with new addr_entries until it's removed

include/seastar/net/dns.hh
```diff
@@ -55,6 +55,15 @@ struct hostent {
     std::vector<sstring> names;
     // Primary address is also always first.
     std::vector<inet_address> addr_list;
+    struct addr_entry
+    {
```
avikivity: Please use full names (addr->address) and put the brace on the right place (same line as struct).

include/seastar/net/dns.hh
avikivity: Most accurate is `std::chrono::duration<int, std::ratio<1, 1>>`

avikivity: But maybe std::chrono::seconds is better as it's immediately understood.

src/net/dns.cc
kreuzerkrieg: kind of...

avikivity:
Please make sure Seastar builds without new deprecation warnings.

avikivity:
> net: expose DNS TTL via net::hostent

TTL is already exposed, so subject is misleading.

kreuzerkrieg:
> Please make sure Seastar builds without new deprecation warnings.

Dont we fail CI on warnings? kinda, "-Wall -Wextra -Werror"?

avikivity:
> > Please make sure Seastar builds without new deprecation warnings.
> 
> Dont we fail CI on warnings? kinda, "-Wall -Wextra -Werror"?

Not on deprecation warnings, since that would fail the build if a dependency caused deprecation.

See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=122442

kreuzerkrieg:
v1:
- rename the struct to address_entry
- keep tests of the deprecated data member
- extract common code from inet_address::find_all
- fix format issues
- used chrono seconds instead of integer

PR #3200, opened 2026-01-13
bool_class: add operator<=>

### LINKED ISSUE #3198: Add <=> operator to bool_class
Otherwise including a bool_class<> member into another class prevents the latter from having default-generated <=>

Like this

```
./db/object_storage_endpoint_param.hh:34:30: error: explicitly defaulted three-way comparison operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
   34 |         std::strong_ordering operator<=>(const s3_storage&) const = default;
      |                              ^
./db/object_storage_endpoint_param.hh:32:23: note: defaulted 'operator<=>' is implicitly deleted because there is no viable three-way comparison function for member 'is_legacy'
   32 |         legacy_format is_legacy;
      |                       ^
```

bhalevy:
and a respective unit test.

Fixes #3198

### DIFF
--- a/include/seastar/util/bool_class.hh
+++ b/include/seastar/util/bool_class.hh
@@ -88,6 +88,10 @@ public:
     /// Equal-to and not-equal-to operators.
     friend bool operator==(bool_class x, bool_class y) noexcept = default;
 
+#if __cpp_lib_three_way_comparison
+    auto operator<=>(const bool_class& other) const noexcept = default;
+#endif
+
     /// Prints bool_class value to an output stream.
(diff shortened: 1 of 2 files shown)


bhalevy:
[9721e36](https://github.com/scylladb/seastar/commit/9721e36466b75698275694e46c590710a21ef000):
- use default 3-way comparison operator
- yes the ::yes and ::no constants in unit test
  - and test equality to test_bool_class(true) and test_bool_class(false), respectively.

PR #3205, opened 2026-01-18
build: support sccache as a compiler cache

avikivity:
Support sccache [1] as an alternative to ccache. Since it
is better than ccache, and unlikely to be randomly present
on some machine, prefer it to ccache.

To avoid caching by both sccache and ccache, skip the
masquerade directory provided by some ccache installations.

[1] https://github.com/mozilla/sccache

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -100,6 +100,51 @@ def find_compiler_cache(preference):
     return ''
 
 
+def find_compiler(name):
+    """
+    Find a compiler by name, skipping ccache wrapper directories.
+
+    This is useful when using sccache to avoid double-caching through ccache.
+
+    Args:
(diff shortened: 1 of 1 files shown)


avikivity:
@travisdowns as you added the original --ccache option, please review.

avikivity:
v2: adjust github test workflow for changed option name

travisdowns:
Took a look and LGTM.

avikivity:
@travisdowns then you might as well merge it

travisdowns:
> @travisdowns then you might as well merge it

I am still working out how I will "properly" merge multi-commit changes. I added a comment here which could help: 

https://github.com/scylladb/seastar/pull/2484#issuecomment-3794762795

PR #3207, opened 2026-01-19
core/loop: repeater,repeat_until_value: don't rethrow exceptions

denesb:
Poll f.failed() before calling f.get(), so if f contains an exception it is propagated up without rethrowing. If _action() is careful enough to avoid throwing itself (instead uses make_exception_future()), the repeater should not regress to throwing.
The other loop primitives in loop.hh already use this approach.

### DIFF
--- a/include/seastar/core/loop.hh
+++ b/include/seastar/core/loop.hh
@@ -78,6 +78,11 @@ public:
                     internal::set_callback(std::move(f), this);
                     return;
                 }
+                if (f.failed()) {
+                    _promise.set_exception(f.get_exception());
+                    delete this;
+                    return;
+                }
                 if (f.get() == stop_iteration::yes) {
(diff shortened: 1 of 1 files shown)


include/seastar/core/loop.hh
```diff
@@ -78,6 +78,11 @@ public:
                     internal::set_callback(std::move(f), this);
                     return;
                 }
+                if (f.failed()) {
+                    _promise.set_exception(f.get_exception());
+                    delete this;
+                    return;
+                }
```
avikivity, line 85: The outer try/catch can now be removed.

denesb: Other loop primites also have it. Maybe they can drop it too. I tend to agree but left it in due to paranoia, we are running in a noexcept context after all.

Issue #66, opened 2015-10-19
Support other forms of AIO


nyh:
Currently, Seastar uses Linux AIO and O_DIRECT for all its I/O, which means that:
- Seastar doesn't work at all on filesystems which don't support O_DIRECT, such as ZFS or even tmpfs.
- Seastar doesn't work efficiently on filesystems where AIO can block, including the popular ext4.

It would have been better if Seastar could emulate asynchronous I/O, perhaps using additional threads. qemu, for example, has this feature (aio=threads, and it is used by default). It would be interesting to see how close we can get performance to the Linux AIO, but more importantly, it will at least give us an option to run Seastar on filesystems which currently can't use it at all.

As a second alternative, we can consider implementing our own simple filesystem (SeastarFS) over block-level AIO which is more universally implemented. I.e., Seastar will be given a block device, and implement the filesystem inside it on its own. This is what the leading database companies are doing.

nyh:
I'm closing this issue. While I guess there's still some theoretical benefit in supporting archaic systems via threads, and it will certainly be interesting to implement SeastarFS, Seastar today already works well fine on tmpfs and ext4, so the details in this issue are no longer relevant.

nyh:
More than four years after I closed this issue as not worth it, @avikivity fixed it in https://github.com/scylladb/seastar/pull/3208.

PR #3208, opened 2026-01-19
reactor: add pure-epoll backend

avikivity:
This series adds a reactor backend that does not use linux-aio, even for
disk I/O. Instead, it uses the syscall thread for disk I/O.

The new backend is meant for deterministic replay environments such as `rr`[1]
that don't support aio APIs like linux-aio or io_uring. They emulate threading
by scheduling one thread at a time, so the limitations of the syscall thread aren't
material.

The series also improves tolerance to ENOSYS on such platforms.

[1] https://rr-project.org/

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -172,6 +172,7 @@ private:
     friend class internal::reactor_stall_sampler;
     friend class preempt_io_context;
     friend struct hrtimer_aio_completion;
+    friend class reactor_backend_epoll_base;
     friend class reactor_backend_epoll;
     friend class reactor_backend_aio;
     friend class reactor_backend_uring;
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -459,6 +459,11 @@ private:
     future<temporary_buffer<char>>
     do_read_some(pollable_fd_state& fd, internal::buffer_allocator* ba);
 
+    future<size_t> do_write(pollable_fd_state& fd, const void* buffer, size_t size);
+    future<size_t> do_writev(pollable_fd_state& fd, std::span<iovec> iov);
+    future<> write_all_part(pollable_fd_state& fd, const void* buffer, size_t len, size_t completed);
+    future<> write_all(pollable_fd_state& fd, const void* buffer, size_t len);
+
 #if SEASTAR_API_LEVEL < 9
     future<size_t>
     do_send(pollable_fd_state& fd, const void* buffer, size_t size);
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -174,6 +174,7 @@ private:
     friend struct hrtimer_aio_completion;
     friend class reactor_backend_epoll_base;
     friend class reactor_backend_epoll;
+    friend class reactor_backend_epoll_pure;
     friend class reactor_backend_aio;
     friend class reactor_backend_uring;
     friend class reactor_backend_selector;
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -462,6 +462,10 @@ private:
     future<temporary_buffer<char>>
     do_read_some(pollable_fd_state& fd, internal::buffer_allocator* ba);
 
+    future<size_t> do_write(pollable_fd_state& fd, const void* buffer, size_t size);
+    future<> write_all_part(pollable_fd_state& fd, const void* buffer, size_t len, size_t completed);
+    future<> write_all(pollable_fd_state& fd, const void* buffer, size_t len);
+
 #if SEASTAR_API_LEVEL < 9
     future<size_t>
     do_send(pollable_fd_state& fd, const void* buffer, size_t size);
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -179,6 +179,7 @@ private:
     friend struct hrtimer_aio_completion;
     friend class reactor_backend_epoll_base;
     friend class reactor_backend_epoll;
+    friend class reactor_backend_epoll_pure;
     friend class reactor_backend_aio;
     friend class reactor_backend_uring;
     friend class reactor_backend_uring_base;
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -177,6 +177,7 @@ private:
     friend class internal::reactor_stall_sampler;
     friend class preempt_io_context;
     friend struct hrtimer_aio_completion;
+    friend class reactor_backend_epoll_base;
     friend class reactor_backend_epoll;
     friend class reactor_backend_aio;
     friend class reactor_backend_uring;
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1123,12 +1123,15 @@ reactor::reactor(std::shared_ptr<seastar::smp> smp, alien::instance& alien, unsi
     , _cpu_stall_detector(internal::make_cpu_stall_detector())
     , _cpu_sched(nullptr, 0)
     , _thread_pool(std::make_unique<thread_pool>(seastar::format("syscall-{}", id), _notify_eventfd)) {
+    auto backend_config = reactor_backend_config{
+        .use_aio_fdatasync = cfg.have_aio_fsync,
+    };
     /*
      * The _backend assignment is here, not on the initialization list as
      * the chosen backend constructor may want to handle signals and thus
      * needs the _signals._signal_handlers map to be initialized.
      */
-    _backend = rbs.create(*this);
+    _backend = rbs.create(*this, backend_config);
     *internal::get_scheduling_group_specific_thread_local_data_ptr() = &_scheduling_group_specific_data;
     _task_queues[0] = std::make_unique<task_queue>(&_cpu_sched, 0, "main", "main", 1000);
     _task_queues[1] = std::make_unique<task_queue>(&_cpu_sched, 1, "atexit", "exit", 1000);
@@ -2434,10 +2437,7 @@ reactor::touch_directory(std::string_view name_view, file_permissions permission
 future<>
 reactor::fdatasync(int fd) noexcept {
     ++_fsyncs;
-    if (_cfg.bypass_fsync) {
-        co_return;
-    }
-    if (_cfg.have_aio_fsync) {
+    if (!_cfg.bypass_fsync) {
         // Does not go through the I/O queue, but has to be deleted
         struct fsync_io_desc final : public io_completion {
             promise<> _pr;
@@ -2462,13 +2462,7 @@ reactor::fdatasync(int fd) noexcept {
         auto req = internal::io_request::make_fdatasync(fd);
         _io_sink.submit(desc, std::move(req));
         co_await std::move(fut);
-        co_return;
     }
-    syscall_result<int> sr = co_await _thread_pool->submit<syscall_result<int>>(
-            internal::thread_pool_submit_reason::file_operation, [fd] {
-        return wrap_syscall<int>(::fdatasync(fd));
-    });
-    sr.throw_if_error();
 }
 
 // Note: terminate if arm_highres_timer throws
@@ -3880,10 +3874,6 @@ bool operator==(const ::sockaddr_in a, const ::sockaddr_in b) {
 
 namespace seastar {
 
-static bool kernel_supports_aio_fsync() {
-    return internal::kernel_uname().whitelisted({"4.18"});
-}
-
 static std::tuple<std::filesystem::path, uint64_t> wakeup_granularity() {
     auto try_read = [] (auto path) -> uint64_t {
         try {
@@ -3978,8 +3968,8 @@ reactor_options::reactor_options(program_options::option_group* parent_group)
                  " Note that if the seastar_memory logger is set to debug or trace level, the diagnostics will be logged irrespective of this setting.")
     , reactor_backend(*this, "reactor-backend", backend_selector_candidates(), reactor_backend_selector::default_backend().name(),
                 fmt::format("Internal reactor implementation ({})", reactor_backend_selector::available()))
-    , aio_fsync(*this, "aio-fsync", kernel_supports_aio_fsync(),
-                "Use Linux aio for fsync() calls. This reduces latency; requires Linux 4.18 or later.")
+    , aio_fsync(*this, "aio-fsync", true,
+                "Use Linux aio for fsync() calls, if available, for backends that use linux-aio. This reduces latency; requires Linux 4.18 or later.")
     , max_networking_io_control_blocks(*this, "max-networking-io-control-blocks", 10000,
                 "Maximum number of I/O control blocks (IOCBs) to allocate per shard. This translates to the number of sockets supported per shard."
                 " Requires tuning /proc/sys/fs/aio-max-nr. Only valid for the linux-aio reactor backend (see --reactor-backend).")
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1372,7 +1372,7 @@ cpu_stall_detector_linux_perf_event::cpu_stall_detector_linux_perf_event(file_de
         : cpu_stall_detector(cfg), _fd(std::move(fd)) {
     void* ret = ::mmap(nullptr, 2*getpagesize(), PROT_READ|PROT_WRITE, MAP_SHARED, _fd.get(), 0);
     if (ret == MAP_FAILED) {
-        abort();
+        throw std::system_error(errno, std::system_category(), "mmap() failed for perf_event_open");
     }
     _mmap = static_cast<struct ::perf_event_mmap_page*>(ret);
     _data_area = reinterpret_cast<char*>(_mmap) + getpagesize();
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -392,11 +392,11 @@ future<temporary_buffer<char>> pollable_fd_state::read_some(internal::buffer_all
 }
 
 #if SEASTAR_API_LEVEL >= 9
-future<size_t> pollable_fd_state::write_some(std::span<iovec> iovs) {
+future<size_t> pollable_fd_state::send_some(std::span<iovec> iovs) {
     return engine()._backend->sendmsg(*this, iovs, internal::iovec_len(iovs));
 }
 #else
-future<size_t> pollable_fd_state::write_some(net::packet& p) {
+future<size_t> pollable_fd_state::send_some(net::packet& p) {
     static_assert(offsetof(iovec, iov_base) == offsetof(net::fragment, base) &&
         sizeof(iovec::iov_base) == sizeof(net::fragment::base) &&
         offsetof(iovec, iov_len) == offsetof(net::fragment, size) &&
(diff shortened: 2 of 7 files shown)


avikivity:
v2: fixed typo in comment pointed out by @copilot

nyh:
Ten years ago (!) I opened an issue asking for this - https://github.com/scylladb/seastar/issues/66.
After six years, I gave up on the idea and closed the issue.
I guess as they say, better late than never!

src/core/reactor.cc
```diff
@@ -337,6 +337,52 @@ reactor::do_sendmsg(pollable_fd_state& fd, std::span<iovec> iovs, size_t len) {
     });
 }
 
+future<size_t>
+reactor::do_write(pollable_fd_state& fd, const void* buffer, size_t len) {
+    return writeable(fd).then([this, &fd, buffer, len] () mutable {
+        auto r = fd.fd.write(buffer, len);
+        if (!r) {
```
nyh, line 385: When do you expect write() to return 0, and not an error? Why is calling the function again immediately the right thing to do?

src/core/reactor.cc
```diff
@@ -337,6 +337,52 @@ reactor::do_sendmsg(pollable_fd_state& fd, std::span<iovec> iovs, size_t len) {
...
+        return do_write(fd, static_cast<const char*>(buffer) + completed, len - completed).then(
+                [&fd, buffer, len, completed, this] (size_t part) mutable {
+            return write_all_part(fd, buffer, len, completed + part);
+        });
+    }
+}
+
+future<>
+reactor::write_all(pollable_fd_state& fd, const void* buffer, size_t len) {
+    SEASTAR_ASSERT(len);
```
nyh: Why is this assert necessary? Why is writing 0 bytes not a perfectly valid thing to do (it will write nothing)?
I see below that for len==0, you won't even call this function. So this assert will never trigger. But, still, why even bother with it - it's still a useless if(). It seems to me that write_all_part will work fine even if given len=0. Or am I missing something?

src/util/process.cc
```diff
@@ -67,29 +67,19 @@ class pipe_data_source_impl final : public data_source_impl {
...
-        auto req = internal::io_request::make_write(_fd.get(), 0, buf.get(), buf_size, false);
-        return _io_queue.submit_io_write(buf_size, std::move(req), nullptr).then(
-            [this, buf = std::move(buf), buf_size] (size_t written) mutable {
-                if (written < buf_size) {
-                    buf.trim_front(written);
-                    return do_put(std::move(buf));
-                }
-                return make_ready_future();
-            });
+        return _fd.write_all(buf.get(), buf_size).then([buf = std::move(buf)] {});
```
nyh, line 71: I'm not experienced enough with how this works, so I want to ask:
When aio is used, does the new code continue to do the same thing it did in the past (using aio)?

src/core/reactor_backend.hh
```diff
@@ -225,7 +225,7 @@ public:
 // Linux. Can wait on multiple file descriptors, and converts other events
 // (such as timers, signals, inter-thread notifications) into file descriptors
 // using mechanisms like timerfd, signalfd and eventfd respectively.
-class reactor_backend_epoll : public reactor_backend {
+class reactor_backend_epoll_base : public reactor_backend {
```
nyh, line 275: I think it will be useful to explain in the comment why this is the epoll "base", what doesn't it do, and what does the one without "base" in its name add to it.

include/seastar/core/reactor.hh
```diff
@@ -174,6 +174,7 @@ private:
     friend struct hrtimer_aio_completion;
     friend class reactor_backend_epoll_base;
     friend class reactor_backend_epoll;
+    friend class reactor_backend_epoll_pure;
```
nyh, line 182: I think the word "pure" doesn't quite explain what this backend does. It isn't just a more "pure" use of epoll, it's actually entirely dependent on worker threads (#66). I wonder if it should be called something like epoll_thread or something. Or maybe just a comment is needed to explain what it is - whatever name we use for it.

nyh (approved):
Looks good. I've waited for this for ten years (https://github.com/scylladb/seastar/issues/66) :-)

How did you test this patch, and how do we test all the different reactor backends going forward? Can we run some set of tests for all the existing reactor backends, for example? Do we do anything of this sort?

nyh:
Oh, and I think the name "pure-epoll" is a bit misleading, since a big part of how it works is those worker threads. Also "pure epoll" sounds somehow better than unpure epoll, while the new reactor backend will probably be the least efficient we have and shouldn't be used unless you must, right?
As I noted in https://github.com/scylladb/seastar/issues/66, qemu for example calls this technique `aio=threads`, focusing on the thread aspect. I wonder if we shouldn't use the word "threads" in the name of our reactor as well. Maybe epoll-threads?
But it's your call. I don't feel I understand this well enough to definitely recommend any name.

src/core/reactor.cc
avikivity: > When do you expect write() to return 0, and not an error?

The kernel buffer full. writeable() return a ready future due to bad speculation.

> Why is calling the function again immediately the right thing to do?

writeable() will return a future that is ready when the buffer is not full.

src/core/reactor.cc
avikivity: It's not necessary.

src/util/process.cc
avikivity: No, it will use write(2). Similar to how socket writes use send*(2).

src/core/reactor_backend.hh
avikivity: Sure

mykaul:
> The new backend is meant for deterministic replay environments such as `rr`[1]

[1] ?

avikivity:
> > The new backend is meant for deterministic replay environments such as `rr`[1]
> 
> [1] ?

Fixed.

avikivity:
v3:
 - removed bad assert in write_all()
 - added comments to the three epoll-related reactor backend classes

src/util/process.cc
nyh: I don't understand. If we use the old "epoll" backend which uses (right?) classic AIO, did something change? Before it didn't use write()? What did it use?

avikivity:
> Oh, and I think the name "pure-epoll" is a bit misleading, since a big part of how it works is those worker threads. Also "pure epoll" sounds somehow better than unpure epoll, while the new reactor backend will probably be the least efficient we have and shouldn't be used unless you must, right? As I noted in #66, qemu for example calls this technique `aio=threads`, focusing on the thread aspect. I wonder if we shouldn't use the word "threads" in the name of our reactor as well. Maybe epoll-threads? But it's your call. I don't feel I understand this well enough to definitely recommend any name.

epoll isn't about how to issue I/O (we don't write to sockets with epoll, we use sendmsg). It's about the readiness/completion mechanism.

reactor_backend_epoll_pure uses only epoll for readiness/completion
reactor_backend_epoll uses epoll for non-disk readiness and linux-aio for disk issue and completions
reactor_backend_aio uses linux-aio for readiness/completion and disk issue, while issuing non-disk I/O via syscalls
reactor_backend_io_uring uses io_uring for readiness/completion and disk issue, while issuing non-disk I/O via syscalls
reactor_backend_tbd will use io_uring for everything, bypassing syscalls for almost everything

src/core/reactor.cc
nyh: Maybe I'm misunderstanding what I'm seeing here. This is a blocking write to the filesystem, not a socket write, right? So why would the kernel return 0 bytes instead of waiting until it can write at least some bytes? Did you see this behavior in practice, or documented somewhere, or it's just defensive programming just in case it ever happens?

avikivity: It is a non-blocking write to a non-disk, non-socket file descriptor (pollable_fd -> non-disk).

The non-socket pollable fds we support are eventfd (for inter-thread communication) and pipes (for communicating with subprocesses).

src/util/process.cc
avikivity: This patch doesn't change anything. It adds some methods.

The following patch changes pipe writes from using our disk write mechanism (which happened to work with linux-aio but won't with pwrite()) to these new methods.

The trigger is that epoll_pure will issue disk writes with pwrite(). These happen not to work with pipes.

nyh:
> epoll isn't about how to issue I/O (we don't write to sockets with epoll, we use sendmsg). It's about the readiness/completion mechanism.
> 
> reactor_backend_epoll_pure uses only epoll for readiness/completion reactor_backend_epoll uses epoll for non-disk readiness and linux-aio for disk issue and completions reactor_backend_aio uses linux-aio for readiness/completion and disk issue, while issuing non-disk I/O via syscalls reactor_backend_io_uring uses io_uring for readiness/completion and disk issue, while issuing non-disk I/O via syscalls reactor_backend_tbd will use io_uring for everything, bypassing syscalls for almost everything

That's true. I guess I find it confusing that the backend is named after **only** the readiness part (and as you saw in the old "epoll" one, sometimes just half of the readiness part), while both readiness and issue is important to understand how a backend works. As you noted, in the future you'll want another one "tbd", which will use io_uring for readiness but also for issue - how will you call this new backend without mentioning the issue?

I thought it makes sense to include both the names of the readiness mechanism and the issue mechanism in the backend's name, especially in this new backend's case where the issue mechanism is such a critical part of the implementation (and what makes it so inefficient and unrecommended except in specific scenarios where you need it).

But you're right, none of the existing backends mention their issue mechanism in their names, so we don't have to start now. But I do think we need clear documentation and/or comments on what these different backends are. I don't think that a year from now, it will be easy to guess what "epoll" is vs "epoll_pure", and which one to use when.

avikivity:
> > epoll isn't about how to issue I/O (we don't write to sockets with epoll, we use sendmsg). It's about the readiness/completion mechanism.
> > reactor_backend_epoll_pure uses only epoll for readiness/completion reactor_backend_epoll uses epoll for non-disk readiness and linux-aio for disk issue and completions reactor_backend_aio uses linux-aio for readiness/completion and disk issue, while issuing non-disk I/O via syscalls reactor_backend_io_uring uses io_uring for readiness/completion and disk issue, while issuing non-disk I/O via syscalls reactor_backend_tbd will use io_uring for everything, bypassing syscalls for almost everything
> 
> That's true. I guess I find it confusing that the backend is named after **only** the readiness part (and as you saw in the old "epoll" one, sometimes just half of the readiness part), while both readiness and issue is important to understand how a backend works. As you noted, in the future you'll want another one "tbd", which will use io_uring for readiness but also for issue - how will you call this new backend without mentioning the issue?

It's really about what's special about the backend. pure epoll is special in that it doesn't use linux-aio. epoll (which is really mixed epoll/aio) is where things started, so it was named so to distinguish it from aio. aio introduced pure aio for socket readiness and disk issue. io_uring doesn't use io_uring for non-disk issue, but since there's only one io_uring backend so far, it wasn't necessary to distinguish it.


> 
> I thought it makes sense to include both the names of the readiness mechanism and the issue mechanism in the backend's name, especially in this new backend's case where the issue mechanism is such a critical part of the implementation (and what makes it so inefficient and unrecommended except in specific scenarios where you need it).

The names are all historical and made sense at a point in time when they were introduced. They are not systematic.

> 
> But you're right, none of the existing backends mention their issue mechanism in their names, so we don't have to start now. But I do think we need clear documentation and/or comments on what these different backends are. I don't think that a year from now, it will be easy to guess what "epoll" is vs "epoll_pure", and which one to use when.

The default should be fine.

avikivity:
> Looks good. I've waited for this for ten years (#66) :-)
> 
> How did you test this patch, and how do we test all the different reactor backends going forward?

I set the backend in ~/.config/seastar and ran the test suite.

> Can we run some set of tests for all the existing reactor backends, for example?

Seems a waste to do it for every run, most don't touch this area.

> Do we do anything of this sort?

No.

src/core/reactor_backend.cc
```diff
@@ -959,6 +959,43 @@ bool reactor_backend_epoll::kernel_submit_work() {
...
+            }
+            case o::write: {
+                auto& req_write = req.as<o::write>();
+                return ::pwrite(req_write.fd, req_write.addr, req_write.size, req_write.pos);
+            }
+            case o::writev: {
+                auto& req_writev = req.as<o::writev>();
+                return ::pwritev(req_writev.fd, req_writev.iovec, req_writev.iov_len, req_writev.pos);
+            }
+            default:
```
xemul, line 1113: This should step on `fsync_io_desc`'s request submitted from `reactor::fdatasync()`

xemul:
> reactor_backend_io_uring uses io_uring for readiness/completion and disk issue, while issuing non-disk I/O via syscalls

But that's not quite so. Both sendmsg/recvmsg of uring backend only do syscalls for socket if they see the speculation marks on pollable fd, otherwise the issue uring CBs

avikivity:
> > reactor_backend_io_uring uses io_uring for readiness/completion and disk issue, while issuing non-disk I/O via syscalls
> 
> But that's not quite so. Both sendmsg/recvmsg of uring backend only do syscalls for socket if they see the speculation marks on pollable fd, otherwise the issue uring CBs

Ah, forgot. So it's more nuanced.

src/core/reactor_backend.cc
avikivity: Ah. I guess I didn't see this because I have unsafe-bypass-fsync set in my config. I'll reproduce and fix.

avikivity:
v4:
 - new patch: reactor: rationalize --aio-fsync switch
 - implement fdatasync in pure-epoll backend

travisdowns:
Is there work being done to have a deterministic mode for seastar? Does it already work in `rr` with this patch? We have done a bit of brainstorming on this our end, perhaps we can swap notes.

avikivity:
> Is there work being done to have a deterministic mode for seastar?

No, but this gets us closer.

> Does it already work in `rr` with this patch?

I managed to crash `rr` and immediately stopped looking in order not to get discouraged.

> We have done a bit of brainstorming on this our end, perhaps we can swap notes.

If we swap notes, you will not have any notes.


But yes, there is potential here, if we manage to tame smp::submit_to().

src/core/reactor.cc
```diff
@@ -1123,12 +1123,15 @@ reactor::reactor(std::shared_ptr<seastar::smp> smp, alien::instance& alien, unsi
...
+    auto backend_config = reactor_backend_config{
+        .use_aio_fdatasync = cfg.have_aio_fsync,
+    };
     /*
      * The _backend assignment is here, not on the initialization list as
      * the chosen backend constructor may want to handle signals and thus
      * needs the _signals._signal_handlers map to be initialized.
      */
-    _backend = rbs.create(*this);
+    _backend = rbs.create(*this, backend_config);
```
xemul: Backends (including aio_storage_context) already have access to reactor and its config, no need to add more configs

src/core/reactor_backend.cc
```diff
@@ -197,12 +202,30 @@ aio_storage_context::handle_aio_error(linux_abi::iocb* iocb, int ec) {
     }
 }
 
+void
+aio_storage_context::fdatasync_via_syscall_thread(int fd, kernel_completion* desc) {
+    // complete_with below satisfies the original promise, so it is safe to
+    // ignore the returned future.
+    (void)std::invoke([] (reactor& r, int fd, kernel_completion* desc) -> future<> {
```
xemul: This ties aio_storage_context back to reactor even tighter 

Can, instead, this be resolved on reactor->backend level with the help of `virtual future<> reactor_backend::fdatasync(int fd) = 0` overridden by backends. AIO/epoll backends would call the nowadays aio/thread-pool fdatasync, uring would just submit the CB, and the epoll-pure would unconditionally call the thread-pool version?

src/core/reactor.cc
avikivity: I think this reduces interdependencies. The reactor and its backend are quite interdependent, but still, it's easier to look up a local thing.

src/core/reactor_backend.cc
avikivity: I'll look into that.

travisdowns:
@nyh 

> How did you test this patch, and how do we test all the different reactor backends going forward? Can we run some set of tests for all the existing reactor backends, for example? Do we do anything of this sort?

On the latter questions we have a good solution for this now: we generate the test matrix by using an "all pairs" strategy: we have a list of interesting dimensions we wnat to test (compilers, build modes, etc), and the full Cartesian product is too big, but we create about 15 combinations which cover all pairs of choices, at least, giving good coverage. We can extend this to backends with little or no increase in CI run counts. As long as we stay under 20 runs, the e2e time stays fast as all 20 runs launch in parallel.

travisdowns:
@avikivity, I've taken the liberty of rebasing this series on master, with the idea of getting this in, as I think all of the outstanding issues are essentially resolved. If you didn't want that, of course feel free to just force push back over it.

A bunch of the commits dropped out as you pushed them piecemeal in smaller changes. A summary is below. There were also a few changes needed to the remaining commits, summarized below too. Finally I added a new commit to fix a bug.

### Dropped or reduced

- `reactor: rename pollable_fd write methods`: reduced. eeb8a9476 already renamed the `span<iovec>` variants to `send_some`/`send_all`. What is left is the `sendmsg`/`recvmsg` -> `send`/`recv` rename plus the `SEASTAR_API_LEVEL < 9` variants.

- `reactor: add write() methods to pollable_fd and friends`: reduced. 705c1568c already added writev-backed `write_some`/`write_all(span<iovec>)`, routed through `reactor_backend::writev` rather than calling `reactor::do_writev` directly, which is the better shape. I dropped the duplicates and kept the scalar `write_some`/`write_all(const void*, size_t)` plus `reactor::do_write`/`write_all_part`. Master's iovec definitions moved out of the `SEASTAR_API_LEVEL >= 9` guard so they are unconditional, matching the header here.

- `reactor: rationalize --aio-fsync switch`: dropped entirely. Master solved this the way @xemul suggested in review: `reactor_backend` carries a `supports_aio_fdatasync` bool_class fixed at construction, `reactor::have_aio_fdatasync()` combines it with the user flag, and `reactor::fdatasync()` moved into `posix_file_impl::fdatasync(with_aio, ...)`. The `reactor_backend_config` plumbing is not needed at all now; the pure-epoll backend simply declares `supports_aio_fdatasync::no`. That also makes the fdatasync/`fsync_io_desc` collision @xemul reported structurally impossible, since fdatasync never reaches the io_sink for this backend. The `o::fdatasync` case in the drain is now dead defensive code.

  One user-visible piece of that commit did not survive: it also changed the `--aio-fsync` default from `kernel_supports_aio_fsync()` to unconditional `true`. Master still uses the kernel check. If that cleanup is still wanted, it is now independent of this series.

### Changes to existing commits

- `reactor_backend: split epoll backend into a base class and complete implementation`: the base ctor now takes `supports_aio_fdatasync` and forwards to `reactor_backend(uses_blocking_io, supports_aio_fdatasync)`. `_r` is `protected`. `get_backend_name()` stays pure virtual in the base and is implemented per concrete backend. `_dying` is a backend member since efb71db02. `writev`, added after this series was written, moved to the base along with the rest.

- `reactor_backend: add pure-epoll backend`: explicit ctor passing `supports_aio_fdatasync::no` in place of the inherited ctor, and `get_backend_name()` returning `"pure-epoll"`.

The `process:`, ENOSYS and perf_event commits applied unchanged.

### New commit

`reactor_backend_epoll_pure::kernel_submit_work()` passed the raw syscall return value to `io_completion::complete_with()`, which expects a negative errno. A failing `pread`/`pwrite`/etc returns -1, which decodes as `-EPERM`, and errno itself was never read on the syscall thread, so the real error was lost. Every disk I/O failure under this backend reported "Operation not permitted". Forcing `EINVAL` out of an unaligned O_DIRECT read:

```
linux-aio:   ERRNO=22 Invalid argument
epoll:       ERRNO=22 Invalid argument
pure-epoll:  ERRNO=1  Operation not permitted
```

The fix captures errno on the syscall thread with `wrap_syscall()` and negates it, matching what the aio fdatasync helper in this series already did. It matters beyond the message, since ENOSPC and EIO handling above the backend keys off the actual code.

### Testing

Full `test.py --mode dev` suite, clang-22, 95 tests, run twice: once with the default backend and once with `reactor-backend=pure-epoll` forced through `~/.config/seastar/seastar.conf`.

Of the 95, 66 actually brought up a pure-epoll reactor. `unit.epoll` and `unit.socket` pin their own backend on the command line, and the remaining 27 never start a reactor.

I will also ensure we test this in CI, though groundwork for that is coming in a separate PR.

PR #3209, opened 2026-01-20
Estimate timing overhead, allow failing if it is too high

travisdowns:
A recent "audit" of Redpanda microbenchmarks showed that many were returning essentially useless results as the overhead of reading the performance counters was much higher than the signal they are trying to measure: in several cases more than 99% of the test was spent just reading the counters.

This is a problem that can't just be fought with documentation (though it helps) - so instead estimate the overhead and warn if it exceeds 10% (default, really you should aim for < 1%). 

First commits refactor a bit the odd way we were passing around the perf counters and including them in the header for no good reason.

Updates the docs.

Example output after this change, showing 99.2% overhead in this artificial test:

```
inja perf_tests_perf && tests/perf/perf_tests_perf -t perf_tests.test_timer_overhead
[3/3] Linking CXX executable tests/perf/perf_tests_perf
WARN  2026-01-20 00:20:46,432 seastar - Could not parse cgroups v1 file (memory/memory.limit_in_bytes).
INFO  2026-01-20 00:20:46,435 seastar - Reactor backend: io_uring
INFO  2026-01-20 00:20:46,474 [shard 20:main] io - IO queue was unable to find a suitable maximum request length, the search was cut-off early at: 16MB
INFO  2026-01-20 00:20:46,474 [shard 20:main] io - IO queue was unable to find a suitable maximum request length, the search was cut-off early at: 16MB
single run iterations:    0
single run duration:      1.000s
number of runs:           5
number of cores:          32
random seed:              2037298104
start/stop overhead:      1.023µs (1.602µs)

test                                iters            runtime     allocs      tasks       inst     cycles   overhead
perf_tests.test_timer_overhead     486000     1.03µs ± 0.07%      0.000      0.000     317.01      351.3       99.2
```

### DIFF
--- a/tests/perf/perf-tests.md
+++ b/tests/perf/perf-tests.md
@@ -25,6 +25,8 @@ combined.single_active                7871    85.271us    76.185ns    85.145us  
 * `-r <n>` or `--runs <n>` – the number of runs of each test to execute
 * `-t <regexs>` or `--tests <regexs>` – executes only tests which names match any regular expression in a comma-separated list `regexs`
 * `--list` – lists all available tests
+* `--overhead-threshold <percent>` – warn if measurement overhead exceeds this percentage (default: 10)
+* `--fail-on-high-overhead` – fail the test run if any test exceeds the overhead threshold
 
 ## Example usage
 
@@ -107,6 +109,32 @@ PERF_TEST(example, custom_time_measurement2)
 
 #### Measurement overhead
 
-The cost of starting and stopping the timers is substantial, about 0.5μs for a start/stop pair, due to the overhead of reading the performance counters in the kernel. In the case you use manual time measurement with start/stop_measuring_time you should ensure that the timed region is substantially longer than time in order to reduce the error imposed by this overhead: for example by using a loop inside the timed region and returning the number of iterations from the test method.
+The cost of starting and stopping the timers is substantial, about 1μs for a start/stop pair, due to the overhead of reading the performance counters in the kernel. When using manual time measurement with `start_measuring_time()`/`stop_measuring_time()`, you should ensure that the timed region is substantially longer than this overhead to reduce measurement error. A common approach is to use a loop inside the timed region and return the number of iterations from the test method.
 
 If you do _not_ use these manual methods, the overhead is very low (a few instructions) as the test method is already called in such a loop by the framework, with the time manipulation methods outside that.
+
+##### Overhead column
+
+The framework tracks and reports the estimated measurement overhead as a percentage in the `overhead` column. This is calculated by:
+
+1. Calibrating the cost of a single `start_measuring_time()`/`stop_measuring_time()` pair at startup
+2. Counting how many times these functions are called during each test run
+3. Computing `overhead% = (call_count × cost_per_call) / measured_runtime × 100`
+
+A high overhead percentage (e.g., >10%) indicates that the timing instrumentation is consuming a significant portion of the measured time, which reduces the accuracy of the results. This typically happens when:
+- The timed region is very short (comparable to the ~1μs overhead)
+- `start_measuring_time()`/`stop_measuring_time()` are called many times with little work between them
+
+To reduce overhead, either:
+- Increase the amount of work in each timed region
+- Use an inner loop and return the iteration count, rather than calling start/stop for each iteration
+
+##### Overhead warnings
+
+By default, a warning is printed if any test has median overhead exceeding 10%:
+
+```
+WARNING: test 'example.my_test' has high measurement overhead: 15.2% (threshold: 10.0%)
+```
+
+You can adjust the threshold with `--overhead-threshold <percent>`, or fail the test run entirely when overhead is too high with `--fail-on-high-overhead`.


tests/perf/perf_tests.cc
```diff
@@ -690,18 +826,29 @@ void performance_test::do_run(const config& conf)
     for (auto& rp : conf.printers) {
         rp->print_result(r);
     }
+
+    // Check if overhead exceeds threshold and warn
+    double median_overhead = r.overhead.stats().med;
+    if (median_overhead > conf.overhead_threshold) {
+        fmt::print("WARNING: test '{}' has high measurement overhead: {:.1f}% (threshold: {:.1f}%)\n",
```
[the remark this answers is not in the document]
travisdowns: The rest of the test output goes to stdout, we want this to go to the same place. stderr is already heavily used by seastar loggers so it is common to run benchmarks with `2>/dev/null` or similar redirection.

tests/perf/perf-tests.md
```diff
@@ -107,6 +109,32 @@ PERF_TEST(example, custom_time_measurement2)
 
 #### Measurement overhead
 
-The cost of starting and stopping the timers is substantial, about 0.5μs for a start/stop pair, due to the overhead of reading the performance counters in the kernel. In the case you use manual time measurement with start/stop_measuring_time you should ensure that the timed region is substantially longer than time in order to reduce the error imposed by this overhead: for example by using a loop inside the timed region and returning the number of iterations from the test method.
+The cost of starting and stopping the timers is substantial, about 1μs for a start/stop pair, due to the overhead of reading the performance counters in the kernel. When using manual time measurement with `start_measuring_time()`/`stop_measuring_time()`, you should ensure that the timed region is substantially longer than this overhead to reduce measurement error. A common approach is to use a loop inside the timed region and return the number of iterations from the test method.
```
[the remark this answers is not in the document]
travisdowns: is "about" not good enough?

travisdowns:
@scylladb/seastar-maint please have a look.

avikivity:
@bhalevy please review

tests/perf/perf_tests.cc
```diff
@@ -146,6 +146,12 @@ class time_measurement {
     linux_perf_event _instructions_retired_counter = linux_perf_event::user_instructions_retired();
     linux_perf_event _cpu_cycles_retired_counter = linux_perf_event::user_cpu_cycles_retired();
 
+    uint64_t _start_stop_count = 0;
+
+    // Calibrated cost of a single start_measuring_time()/stop_measuring_time() pair in nanoseconds
+    double _start_stop_overhead_ns = 0.0;         // internal timing (used for overhead calculations)
+    double _start_stop_overhead_external_ns = 0.0; // external clock measurement (for comparison)
```
avikivity: Why in nanoseconds? Since you're using floating point, might as well keep it in natural units.

avikivity:
Looks good, but will await @bhalevy review.

tests/perf/perf_tests.cc
travisdowns: Good call, this should actually just be using `clock_time::duration` like other durations. Implemented that in last push.

tests/perf/perf_tests.cc
```diff
@@ -242,6 +288,8 @@ struct config {
     unsigned number_of_runs;
     std::vector<std::unique_ptr<result_printer>> printers;
     unsigned random_seed = 0;
+    double overhead_threshold = 10.0;  // warn if overhead exceeds this percentage
```
bhalevy: nit: it would clearer IMO to be expressed as a pure number like `0.1` rather than as percentage.

tests/perf/perf_tests.cc
```diff
@@ -555,10 +611,11 @@ struct column {
...
 static const std::vector<column> common_columns{
-    {"allocs", 3, [](const result& r) { return r.allocs;              }},
-    {"tasks" , 3, [](const result& r) { return r.tasks;               }},
-    {"inst"  , 2, [](const result& r) { return r.inst;                }},
-    {"cycles", 1, [](const result& r) { return r.cycles;              }},
+    {"allocs"  ,  3, [](const result& r) { return r.allocs;              }},
+    {"tasks"   ,  3, [](const result& r) { return r.tasks;               }},
+    {"inst"    ,  2, [](const result& r) { return r.inst;                }},
+    {"cycles"  ,  1, [](const result& r) { return r.cycles;              }},
+    {"overhead%", 1, [](const result& r) { return r.overhead;           }},
```
bhalevy: nit: misaligned closing brace.

tests/perf/perf_tests.cc
```diff
@@ -757,6 +816,11 @@ void performance_test::do_run(const config& conf)
                 add(r.tasks, rr.stats.tasks_executed);
                 add(r.inst, rr.stats.instructions_retired);
                 add(r.cycles, rr.stats.cpu_cycles_retired);
+
+                // Calculate overhead percentage from start/stop timing calls
+                auto overhead = rr.start_stop_count * measure_time.start_stop_overhead();
+                double overhead_pct = dt.count() ? (100.0 * overhead / dt) : 0.;
```
bhalevy: here too, it would be clearer if the overhead will be calculated as a pure ratio rather than converted to percentage.

tests/perf/perf_tests.cc
```diff
@@ -768,18 +832,29 @@ void performance_test::do_run(const config& conf)
     for (auto& rp : conf.printers) {
         rp->print_result(r);
     }
+
+    // Check if overhead exceeds threshold and warn
+    double median_overhead = r.overhead.stats().med;
+    if (median_overhead > conf.overhead_threshold) {
+        fmt::print("WARNING: test '{}' has high measurement overhead: {:.1f}% (threshold: {:.1f}%)\n",
+                   name(), median_overhead, conf.overhead_threshold);
+        return true;  // overhead was too high
```
bhalevy: why not throw an exception if `fail_on_high_overhead` is set and return void if not without changing the signature?
I think that the handling of the overhead should be in one place, either here or by the caller, but not split between the two.

travisdowns: Good call, changed to that approach.

tests/perf/perf_tests.cc
travisdowns: Eagle eyes :), fixed.

tests/perf/perf_tests.cc
travisdowns: Switched to ratio everywhere except the exception message.

tests/perf/perf_tests.cc
travisdowns: switched to ratio

bhalevy (approved):

travisdowns:
@scylladb/seastar-maint this has been approved, please consider merging.

tests/perf/perf_tests.cc
```diff
@@ -228,10 +355,17 @@ struct result {
     float_stats<> tasks;
     float_stats<> inst;
     float_stats<> cycles;
+    float_stats<> overhead;  // overhead percentage from start/stop timing calls
 };
 
-struct duration {
-    double value;
+struct duration {
```
bhalevy, line 361: Looks like pre-commit is complaining about a trailing whitespace introduced in this patch.

Cc @xemul @avikivity

travisdowns: let me fix it, this was a race with the introduction of pre-commit

travisdowns: PR https://github.com/scylladb/seastar/pull/3245

PR #3210, opened 2026-01-20
Introduce "integrated queue length" metrics and use it for IO classes

xemul:
Sometimes a performance metrics is inherently volatile and fluctuates rapidly. The great example of such a metrics is length of task queues or IO queues. Nowadays those values are exported as gauges. When monitoring system periodically "scrapes" (polls) those values, the collected data point is merely a snapshot of the queue state at that precise moment that doesn’t really show much information about what really happens in that area. Scraping happens once every several minutes, while queue length changes almost every millisecond. The extracted information is less than one thousand's part of a percent. Negligibly little information it is.

In the old days the very same problem existed for IO request latencies. Back then the latency of requests in a class was tracked by saving the latency of the last completed request and exporting it as gauge. Similarly -- almost all the information about request latencies was lost. The 803e790598fdc9ff28e75d1222749d62015fbf90 fixed that and introduced a monotonically growing total_queue_time value exported as counter and monitored using promql `rate()` function.

This PR extends this approach onto IO queue length by introducing the integrated_length<T> wrapper.

It makes little sense just to add-up the queue-lengths whenever the queue gets modified. Instead, the integrator maintains the sum of `length * duration` products taken at specific checkpoints. In it the `length` is the length of the queue when being checkpoint-ed, and the `duration` is the time passed since last checkpoint. If drawing the XY plot where X is time and Y is queue length, the accumulated value is then the square of the area below the plot. If, as promql rate does, getting two integrated values, subtracting the latest one from previous and dividing it by the duration (difference of X points) -- the result would apparently be the average Y value between two scrape points. Which is _much_ more information compared to a randomly scraped point.

It's not perfect, of course. Short spikes are averaged-out and are thus swallowed by this metrics. The shorter the spike is the less noticeable it comes. Nonetheless, this gives much more information about how queue length changes over time, as compared to randomly taken gauge snapshots.

### DIFF
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -31,6 +31,7 @@
 #include <boost/container/small_vector.hpp>
 #include <sys/uio.h>
 #include <seastar/util/assert.hh>
+#include <seastar/util/integrated-length.hh>
 
 #include <seastar/core/file.hh>
 #include <seastar/core/io_queue.hh>
@@ -136,8 +137,8 @@ class io_queue::priority_class_data {
             bytes += len;
         }
     } _rwstat[2] = {}, _splits = {};
-    uint32_t _nr_queued;
-    uint32_t _nr_executing;
+    util::integrated_length<unsigned short, lowres_clock> _nr_queued;
+    util::integrated_length<unsigned short, lowres_clock> _nr_executing;
     std::chrono::duration<double> _queue_time;
     std::chrono::duration<double> _total_queue_time;
     std::chrono::duration<double> _total_execution_time;
@@ -238,6 +239,14 @@ public:
 
     void on_split(io_direction_and_length dnl) noexcept {
         _splits.add(dnl.length());
+    }
+
+    void on_before_dispatch() noexcept {
+        _nr_queued.checkpoint();
+    }
+
+    void on_after_dispatch() noexcept {
+        _nr_executing.checkpoint();
     }
 
     fair_queue::class_id fq_class() const noexcept { return _pc.id(); }
@@ -781,12 +790,15 @@ std::vector<seastar::metrics::impl::metric_definition_impl> io_queue::priority_c
             // In other words: the new counter tells you how busy a class is, and the
             // old counter tells you how busy the system is.
 
-            sm::make_queue_length("queue_length", _nr_queued, sm::description("Number of requests in the queue")),
-            sm::make_queue_length("disk_queue_length", _nr_executing, sm::description("Number of requests in the disk")),
+            sm::make_queue_length("queue_length", [this] { return _nr_queued.value(); }, sm::description("Number of requests in the queue")),
+            sm::make_queue_length("disk_queue_length", [this] { return _nr_executing.value(); }, sm::description("Number of requests in the disk")),
             sm::make_gauge("delay", [this] {
                 return _queue_time.count();
             }, sm::description("random delay time in the queue")),
-            sm::make_gauge("shares", _shares, sm::description("current amount of shares"))
+            sm::make_gauge("shares", _shares, sm::description("current amount of shares")),
+
+            sm::make_counter("integrated_queue_length", [this] { return _nr_queued.integral(); }, sm::description("Integrated queue length")),
+            sm::make_counter("integrated_disk_queue_length", [this] { return _nr_executing.integral(); }, sm::description("Integrated disk queue length")),
     });
 }
 
@@ -1040,6 +1052,12 @@ future<size_t> io_queue::submit_io_write(size_t len, internal::io_request req, i
 // This is far from ideal, but it's something.
 
 void io_queue::poll_io_queue() {
+    for (auto& pc : _priority_classes) {
+        if (pc) {
+            pc->on_before_dispatch();
+        }
+    }
+
     for (auto&& st : _streams) {
         st.out.maybe_replenish_capacity(st.replenish);
         auto available = st.reap_pending_capacity();
@@ -1071,6 +1089,12 @@ void io_queue::poll_io_queue() {
         // unpleasant (it can bloat the bucket beyond its size limit, and its hard to write a correct
         // countermeasure for that), so we just discard the tokens. There's no harm in it, IO cancellation
         // can't have resource-saving guarantees anyway.
+    }
+
+    for (auto& pc : _priority_classes) {
+        if (pc) {
+            pc->on_after_dispatch();
+        }
     }
 }
 


xemul:
image

An example of how two metrics look like (old, gauge, left-sife and new, rate-d counter, right side) for compaction class during the backup test

xemul:
image

One more example -- the commitlog queue lengths for a node lightly loaded by latte writer

xemul:
image

And the proof of correctness -- a static integrator with the constant value of 1, collected, exported and rate-d similarly to the queue lengths from above examples (not included in this PR)

travisdowns:
Very nice, this is the right way to do those metrics IMO, i.e., the old "sample at random moment" ones could just be deleted.

If you do want a counterpart to these, which can capture the sort of "max length" micro-behavior within a sample period which is something lost in the averaging, what I have found useful is a complementary metric which is "rolling max" of the value (sometimes rolling min is useful, and rolling average is essentially this integrated one). The metric exposes the max over some configurable lookback period. Ideally you align the lookback period to the scrape frequency but it still works alright if they are mismatched.

include/seastar/util/integrated-length.hh
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    bool operator!=(const T& v) const noexcept {
+        return _value != v;
+    }
+
+    // Updates the integral value by adding yet another
+    // value * duration component. The "now" time-point can be
+    // provided (e.g. for tests or optimizations)
+    void checkpoint(Clock::time_point now = Clock::now()) noexcept {
```
travisdowns, line 98: It feels like the need to manual checkpointing is unecessary? We have this same kind of integrated metric in Redpanda (also for queue lengths), but it always keep the integrated value up to date, no need to call checkpoint. Critically, it is useful that the `integral` method updates the integral (effectively, that it checkspoints). Otherwise it is very easy to get bogus data: e.g., the queue has 5 elements in it, but isn't checkpointed for a metric interval, then the rate drops to 0 for that period, then in the next period it jumps to 10, not reflecting reality.

`integral` can be non-const and do a true update, or it stay cost and just calculate the additional term to fill out the integral since the last checkpoint (probably better?).

avikivity:
Looks reasonable, but maybe there is an accepted name for such a metric? We should adopt it if so. AI tools can help find the name given a description.

travisdowns:
> maybe there is an accepted name for such a metric? We should adopt it if so. AI tools can help find the name given a description.

When we added this 60% of the time spent was discussing the name IIRC.

AI suggests "Integral Metrics or Cumulative Time-Weighted Metrics"

I think _Integral_ is pretty bad as it has about 2 other, arguably more common, meanings.

avikivity:
Cumulative is a good word here.

avikivity:
Gemini suggests "request_duration_seconds_total".

avikivity:
No, that's the sum of durations, not the sum of length*duration.

travisdowns:
`iostat` has similar naming to that for its cumulative metric and yes it's confusing/wrong

avikivity:
So now:

 - total waiting time
 - queue request-seconds
 - integrated queue length

I think I like total waiting time best.

xemul:
> If you do want a counterpart to these, which can capture the sort of "max length" micro-behavior within a sample period which is something lost in the averaging, what I have found useful is a complementary metric which is "rolling max" of the value (sometimes rolling min is useful, and rolling average is essentially this integrated one). The metric exposes the max over some configurable lookback period. Ideally you align the lookback period to the scrape frequency but it still works alright if they are mismatched.

I've been looking for "max queue length between scrapes' too, but don't like re-setting the maximum on timer. I admit it works, just doesn't look nice.

include/seastar/util/integrated-length.hh
xemul: I added explicit checkpoints for several reasons
1. de-queuing happens in batches (when dispatching), so kicking the `result += value * (now - ts)` once every -- is overkill
2. time since previous udpate resolution is either too expensive for frequent updates (highres clock) or will be zero most of the time (lowres clock)
3. (the main reason) I want to have "queue length the dispatch loop deals with" metrics, explicitly. Checkpointing is right before doing the dispatch makes it happen

But by and large yes, it can be automatically checkpointed on every modification

xemul:
> Looks reasonable, but maybe there is an accepted name for such a metric? We should adopt it if so. AI tools can help find the name given a description.

The closest analogy "man-hours" analogue in pro(duct|ject|cess) management.
AI suggested "average queue load", but the newly added class is not about queues, it's about integrating any fluctuating integer, so not perfect match

In physics there are "something * time" units:
https://en.wikipedia.org/wiki/Action_(physics)
https://en.wikipedia.org/wiki/Impulse_(physics)

My favorite is from kinematics :) [absement](https://en.wikipedia.org/wiki/Absement) = absence + displacement

>  total waiting time

Confuses with "total latency"

>  queue request-seconds

reqond (REQest*secOND) 

>  integrated queue length

Current name, yes

xemul:
> When we added this 60% of the time spent was discussing the name IIRC.

There are two hard problems in computer science – cache invalidation, naming and off-by-one errors. 
So yes, we're going to re-approve this theory here :D

xemul:
> I think Integral is pretty bad as it has about 2 other, arguably more common, meanings.

But the new helper class literally calculates the approximation of the integral of the XY plot

xemul:
> I've been looking for "max queue length between scrapes' too, but don't like re-setting the maximum on timer. I admit it works, just doesn't look nice.

I have a very draft version that collects a histogram of durations where buckets are selected based on the queue lengths. Like "queue was sized [0,2) for 1.2 seconds, sized [2,4) for 3.2 seconds, sized [4,8) for 0.1 seconds and sized [8,infty) for 2.6 seconds". Out of it I can derive the p99-ish metrics saying that "the queue was sized above N 99% of the time so far". But it has more problems than benefits for now :( Maybe something can come out of it later

travisdowns:
> I have a very draft version that collects a histogram of durations where buckets are selected based on the queue lengths. Like "queue was sized [0,2) for 1.2 seconds, sized [2,4) for 3.2 seconds, sized [4,8) for 0.1 seconds and sized [8,infty) for 2.6 seconds". Out of it I can derive the p99-ish metrics saying that "the queue was sized above N 99% of the time so far". But it has more problems than benefits for now :( Maybe something can come out of it later

Right, that's one way to keep it "counter based" without the ugliness of the rolling window (note that the rolling window doesn't reset on scrape: it is a rolling window, separately configured, scape is a "read only" option, but perhaps you still consider it ugly), but it comes at a high cost of many metrics and imprecise identification of the max (of course, you get something for this cost: a histogram).

xemul:
@travisdowns , what's your preference on the new metrics name? Provided the "integral" [still](https://github.com/scylladb/seastar/pull/3210#issuecomment-3776509194)  not good

travisdowns:
> @travisdowns , what's your preference on the new metrics name? Provided the "integral" [still](https://github.com/scylladb/seastar/pull/3210#issuecomment-3776509194) not good

Sorry, I missed your earlier comment. Yes, it is the integral, but in english words always need 3+ meanings :). 

So I think the right name would be `integrated_...` not `integral_...` as you need an adjective there. So "integral" as an adjective (meaning "whole") seems wrong. It should be `integrated` which as an adjective means "resulting from mathematical integration" or something.

travisdowns:
And sorry, I had provided that offhand comment without noticing that `integral` was actually the chosen approach in the PR, otherwise I would have been more constructive.

xemul:
upd:
- renamed class and file integrated_length -> integrated_value
- renamed metrics integral_queue_length -> integrated_queue_length
- moved unit test into its own .cc and split into explicit static/disperse cases
- added licensing comments

include/seastar/util/integrated-length.hh
travisdowns: OK, makes sense.

What integral() not calculating the "tail" since the last checkpoint? Is that also desirable?

travisdowns: FWIW the only expensive part of the update/checkpoint is now(), the rest is trivial, so if that's a concern it's good to capture now() at the start and pass the same value down (more for other cases where you do want implicit checkpointing).

xemul:
upd:
- fixed spelcheck mistakes spotted by copilot

xemul:
:facepalm: indeed. Renamed metrics names

travisdowns (approved):

xemul:
@avikivity , @travisdowns , a light reminder ping

travisdowns:
> @avikivity , @travisdowns , a light reminder ping

 I haven't yet merged any multi-commit changes since I need to set myself up to do that, probably using Avi's script.

travisdowns:
I merged this via github since the script didn't seem much better after playing with it.

PR #3213, opened 2026-01-21
Remove global scheduling group ID bitmap

xemul:
The bitmap is used to allocate ID for newly created scheduling group. This PR replaces it with the logic applied in supergroup ID allocation -- the ID is "allocated" by looking up free slot in sched group pointers on shard-0, then the existing group initialization proceeds.

The main intent is to keep the sched group counting control in one place and keep groups / supergroups ID allocation symmetrical


xemul:

PR #3215, opened 2026-01-23
reactor: Unfriend pollable_fd via pollable_fd_state::make()

xemul:
The only reason why pollable_fd is friend of reactor is for the former to call engine()._backend->... method. The pollable_fd is assumed to be a lightweight pimpl-style wrapper over pollable_fd_state which, in turn, has access to reactor/reactor_backend methods.

This patch applies this design to pollable_fd constructor as well.


xemul:

travisdowns (approved):

Issue #3216, opened 2026-01-23
Allow minimal/zero HTTP headers in reply


m-szymon:
Applications should be able to control exactly which headers appear in responses, including the option to send zero optional headers.
Seastar currently:
- Always adds Server and Date headers to every response
- Always sets Content-Type when writing a response body via `write_body()`

Of course application could remove or overwrite these header, however when desired the should not be added at all.

PR #3217, opened 2026-01-23
http: allow minimal/zero HTTP headers in reply

m-szymon:
Seastar's HTTP server adds Server and Date headers to every response
by default, and the write_body() helpers implicitly set Content-Type.
Applications embedding the server (e.g. ScyllaDB) may need to
customize or suppress these headers.  While one could remove them
after the fact, this is wasteful and not discoverable.  Per RFC 9110:
the Server header is optional (MAY); the Date header is required
(MUST) for origin servers with a clock, though the RFC does allow
origin servers without a clock to omit it; and Content-Type is
recommended (SHOULD).  This change gives applications explicit
control over all three, letting each decide which RFC trade-offs are
acceptable for its use case.

- Add set_server_header(sstring) to http_server: sets the Server
  header value; an empty string suppresses the header entirely
  (default: "Seastar httpd").
- Add set_generate_date_header(bool) to http_server: when false,
  omits the Date header and stops the periodic date-update timer
  (default: true).
- In write_body(), treat an empty content_type as "don't set
  Content-Type" rather than adding an empty header.  An empty
  media-type is invalid per RFC 9110, so no legitimate value is
  lost.  Note: set_content_type("") previously fell through to
  "text/plain" via the extension lookup; it is now a no-op.

Fixes https://github.com/scylladb/seastar/issues/3216

### DIFF
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -132,6 +132,8 @@ class http_server {
     timer<> _date_format_timer { [this] {_date = http_date();} };
     size_t _content_length_limit = std::numeric_limits<size_t>::max();
     bool _content_streaming = false;
+    std::optional<sstring> _server_header = sstring("Seastar httpd");
+    bool _generate_date_header = true;
     gate _task_gate;
     std::optional<net::keepalive_params> _keepalive_params;
 public:
@@ -181,6 +183,21 @@ public:
     bool get_content_streaming() const;
 
     void set_content_streaming(bool b);
+
+    /// Returns the value of the "Server" header that will be added to each response.
+    /// std::nullopt means the header will not be added.
+    const std::optional<sstring>& get_server_header() const;
+
+    /// Sets the value of the "Server" header added to each response.
+    /// Pass std::nullopt to suppress the header entirely.
+    void set_server_header(std::optional<sstring> value);
+
+    /// Returns whether the server adds a "Date" header to each response.
+    bool get_generate_date_header() const;
+
+    /// Controls whether the server adds a "Date" header to each response.
+    /// When set to false the periodic date-update timer is also stopped.
+    void set_generate_date_header(bool b);
 
     future<> listen(socket_address addr, server_credentials_ptr credentials);
     future<> listen(socket_address addr, listen_options lo, server_credentials_ptr credentials);
--- a/include/seastar/http/reply.hh
+++ b/include/seastar/http/reply.hh
@@ -31,6 +31,7 @@
 //
 #pragma once
 
+#include <optional>
 #include <unordered_map>
 #include <string_view>
 #include <seastar/core/sstring.hh>
@@ -202,12 +203,21 @@ struct reply {
      * 2. If a "/" is missing in the given string, we look it up in a list of
      *    common file extensions listed in the http::mime_types map. For
      *    example "html" will be mapped to "text/html".
+     *
+     * std::nullopt and empty string are both no-ops (an empty media-type is
+     * invalid per RFC 9110).  To remove a previously set Content-Type use
+     * _headers.erase("Content-Type").  To force an empty value use
+     * add_header("Content-Type", "").
      */
-    reply& set_content_type(std::string_view content_type) {
-        if (content_type.find('/') == std::string_view::npos) {
-            content_type = http::mime_types::extension_to_type(content_type);
+    reply& set_content_type(std::optional<std::string_view> content_type) {
+        if (!content_type) {
+            return *this;
         }
-        _headers["Content-Type"] = sstring(content_type);
+        std::string_view ct = *content_type;
+        if (ct.find('/') == std::string_view::npos) {
+            ct = http::mime_types::extension_to_type(ct);
+        }
+        _headers["Content-Type"] = sstring(ct);
         return *this;
     }
 
@@ -249,7 +259,8 @@ struct reply {
      * with a function.
      *
      * \param content_type - is used to choose the content type of the body. Use the file extension
-     *  you would have used for such a content, (i.e. "txt", "html", "json", etc')
+     *  you would have used for such a content, (i.e. "txt", "html", "json", etc').
+     *  std::nullopt or an empty string skips setting Content-Type (see set_content_type()).
      * \param body_writer - a function that accept an output stream and use that stream to write the body.
      *   The function should take ownership of the stream while using it and must close the stream when it
      *   is done.
@@ -258,7 +269,7 @@ struct reply {
      *
      */
 
-    void write_body(const sstring& content_type, http::body_writer_type&& body_writer);
+    void write_body(std::optional<std::string_view> content_type, http::body_writer_type&& body_writer);
 
     /*!
      * \brief use and output stream to write the message body
@@ -268,7 +279,7 @@ struct reply {
      */
     template <typename W>
     requires std::is_invocable_r_v<future<>, W, output_stream<char>&>
-    void write_body(const sstring& content_type, W&& body_writer) {
+    void write_body(std::optional<std::string_view> content_type, W&& body_writer) {
         write_body(content_type, [body_writer = std::move(body_writer)] (output_stream<char>&& out) mutable -> future<> {
             return util::write_to_stream_and_close(std::move(out), std::move(body_writer));
         });
@@ -278,12 +289,13 @@ struct reply {
      * \brief Write a string as the reply
      *
      * \param content_type - is used to choose the content type of the body. Use the file extension
-     *  you would have used for such a content, (i.e. "txt", "html", "json", etc')
+     *  you would have used for such a content, (i.e. "txt", "html", "json", etc').
+     *  std::nullopt or an empty string skips setting Content-Type (see set_content_type()).
      * \param content - the message content.
      * This would set the the content and content type of the message along
      * with any additional information that is needed to send the message.
      */
-    void write_body(const sstring& content_type, sstring content);
+    void write_body(std::optional<std::string_view> content_type, sstring content);
 
     // RFC7231 Sec. 4.3.2
     // For HEAD replies collect everything from the handler, but don't write the body itself
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -327,8 +327,12 @@ future<> connection::respond() {
 }
 
 void connection::set_headers(http::reply& resp) {
-    resp._headers["Server"] = "Seastar httpd";
-    resp._headers["Date"] = _server._date;
+    if (_server._server_header.has_value()) {
+        resp._headers["Server"] = *_server._server_header;
+    }
+    if (_server._generate_date_header) {
+        resp._headers["Date"] = _server._date;
+    }
 }
 
 future<bool> connection::generate_reply(std::unique_ptr<http::request> req) {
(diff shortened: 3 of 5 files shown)


xemul:
Instead of intruducing yet another API knob for httpd, can we fix the issue by suggesting that handler code calls `resp._headers.clear()` when being called? AFAICS handler is called _after_ default headers are configured:

```c++
future<bool> connection::generate_reply(std::unique_ptr<http::request> req) {
    auto resp = std::make_unique<http::reply>();
    resp->set_version(req->_version);
    set_headers(*resp);
    ...
    return _server._routes.handle(url, std::move(req), std::move(resp)) ...
```

nyh:
@xemul even if this is possible, I don't think it's a good idea. Besides being wasteful (why bother creating this header just to immediately delete it), it's also not really discoverable. 

@m-szymon I think we should make it even more configurable than what you did, because "Server" and "Date" aren't the same:

The "Server" header is especially problematic: According to the HTTP standard (RFC 9110), the "Server" header is optional. The value that Seastar puts there by default, "Seastar httpd", is - to put it bluntly - nothing more than a Seastar advertisement :-( There is no reason why Seastar needs to advertise itself in the HTTP server of bigger projects that use it (like ScyllaDB), any more than you'd expect every executable compiled by clang to print "Compiled by clang" when it boots. So even if Seastar does default to printing this string, there should be a configuration option to **change this string**, to some other string (e.g., "ScyllaDB"), or to nothing (in which case a "Server" header is not printed).

The "Date" header is different - here the RFC says that a server (at least one which has a clock) "MUST" generate the Date field. This header is important for use-cases where caching plays a role. So not generating the "Date" header should be an explicit choice made by the application - an application which knows its responses are never cached might choose to drop it but other applications will not - and shouldn't be tied to the decision on the Server header.

By the way, I would call this configuration "generate date header" (defaulting to true), not the reverse "skip date header". It's not like the Date field just appears miraculously and we need to skip it - the question is whether we generate it or not. Moreover, the phrase "skipping" a header makes it sound like maybe the _request_ header is being skipped, which is not the case - this option is about whether or not to generate this header in the response.

include/seastar/http/httpd.hh
```diff
@@ -184,6 +185,13 @@ public:
 
     void set_content_streaming(bool b);
 
+    /// When set to true, the server will not add default "Server" and "Date" headers to responses.
+    /// This allows the application to fully control these headers.
+    bool get_skip_default_headers() const;
+
+    /// Set whether to skip adding default "Server" and "Date" headers to responses.
+    void set_skip_default_headers(bool b);
```
nyh: As I explained in legth in a separate comment, I think the option for Server and Date should be separate. Moreover, for "Server" the option shouldn't be just to generate it or not, but the actual string that will be printed (defaulting to "Seastar httpd") - if the string is empty, the header will not be generated.

I also think this option should be called `generate_date_header()`, not its inverse `skip_date_header()`, to emphasize that what the server is doing (or not) is to **generate** this **reponse** header - it's not about skipping headers (request headers?)

include/seastar/http/reply.hh
```diff
@@ -275,6 +275,27 @@ struct reply {
         });
     }
 
+    /*!
+     * \brief use an output stream to write the message body without setting Content-Type
+     *
+     * Same as write_body(content_type, body_writer) but does not set the Content-Type header.
+     * Useful when the caller wants to control Content-Type separately or omit it entirely.
+     */
+    void write_body(http::body_writer_type&& body_writer);
```
nyh: The RFC says that the server "SHOULD" sent a Content-Type in the response. This is not as strong as "MUST" but still pretty strong. I think that dropping this header in a response to shave a few bytes is a mistake. That being said, I guess it does indeed make sense to give the user at least the option to not do it so I support this change.

The only thing that worries me a bit is that because this is an overload of the existing write_body(), and it has fewer parameters, new users will start using it "by mistake" instead of the original overload with the content-type.

src/http/httpd.cc
```diff
@@ -340,8 +340,10 @@ future<> connection::write_body() {
 }
 
 void connection::set_headers(http::reply& resp) {
-    resp._headers["Server"] = "Seastar httpd";
-    resp._headers["Date"] = _server._date;
+    if (!_server._skip_default_headers) {
+        resp._headers["Server"] = "Seastar httpd";
+        resp._headers["Date"] = _server._date;
```
nyh, line 334: We have special code in class http_server (httpd.hh) that updates this "_date" every one second, using a timer. I think if we disable the Date setting, we should also not start that timer (or cancel it / start it when the option's value is changed later).

src/http/reply.cc
```diff
@@ -124,14 +124,22 @@ sstring reply::response_line() const {
...
+    write_body(std::move(body_writer));
+}
+
+void reply::write_body(body_writer_type&& body_writer) {
     _body_writer  = std::move(body_writer);
 }
 
 void reply::write_body(const sstring& content_type, sstring content) {
-    _content = std::move(content);
+    write_body(std::move(content));
```
nyh: Why make this change and make things more difficult for the optimizer? :-)
But more importantly, I think you can't do this change because write_body(content) also needs to set done() on its own - see below.

src/http/reply.cc
```diff
@@ -124,14 +124,22 @@ sstring reply::response_line() const {
...
     done(content_type);
 }
 
+void reply::write_body(sstring content) {
+    _content = std::move(content);
```
nyh, line 124: I don't remember the exact reason (@amnonh is the expert in this code), but I think you need to call done() in this function, like the original function did. There's a done() overload without a content_type parameter.

If I'm right that this is a bug, then it should have been something you would have seen in a test. Which reminds me - please add a test case which tries to use this content-type-less response and see that it actually works, including setting the proper _response_line.

swasik:
@m-szymon could we follow-up on it?

include/seastar/http/reply.hh
m-szymon: That's a valid concern.
I considered just omitting header generation for empty string, however empty header is a valid value.
It would be the easiest way and if user really wants to set empty string he can workaround by setting header manually and then omitting overwrite.
But this is a breaking change (although very minor and unlikely), so I didn't follow it. But maybe good idea?

include/seastar/http/httpd.hh
m-szymon: Treating Date and Server separately makes sense.
But if the user wants some other values, he can set them manually.
Adding such special helper for Server seemed a little weird as in general we don't have it for other headers. But having a flag only would be even weirder.

src/http/reply.cc
m-szymon: If it is a simple setter, optimizer should handle it.
If it is anything complex, having the same thing it two places, is just asking for errors.

Regarding `done` - it should be cleaned on its own, see below.

src/http/reply.cc
m-szymon: I didn't check the timeline, but it seems there were some changes not really finished - I am not sure in which direction.
`write_body` with `body_writer` uses directly `set_content_type` - not `done(...)`.
And in Scylla we call `done()` again - so it seems it has been a responsibility of a user.
And `done()` just sets `_response_line = response_line();`, which is then nowhere used - `write_reply` generates it again (see just 3 lines below).

So if it read it correctly `done(content_type)` should be replaced with `set_content_type`, `done()` should be no-op, and `_response_line` should disappear. Which would be breaking, if anyone for any reason accesses `_response_line`.
Unless the direction is opposite.

Anyway I will update tests accordingly.

m-szymon: I knew my code is correct, but now I checked around and pushed separate PR: https://github.com/scylladb/seastar/pull/3306

include/seastar/http/httpd.hh
nyh: I agree that if the user wants to set "Server", they can do it manually, but since we already do set it, what do you suggest? That we should have a flag to just **not** set it? That the user should take our "Seastar Server" advertisement header and then remove it or change it to what they want?

For Date, since the standard claims that the server should produce it, I think it's a good idea for the server to have the code to produce it and not require every user to write code to do it. On the other hand, since the standard reluctantly allows a server "which does not have clock" to not generate a Date, I think we should have a flag to disable generating the Date. This can help users who are obsessed with having smaller headers to disable the Date header.

src/http/reply.cc
nyh: Thanks! Merged #3306. So please rebase on top of it (if you need to).

src/http/httpd.cc
```diff
@@ -372,6 +376,28 @@ void http_server::set_content_streaming(bool b) {
...
+}
+
+void http_server::set_generate_date_header(bool b) {
+    _generate_date_header = b;
+    if (b) {
+        _date = http_date();
+        _date_format_timer.arm_periodic(1s);
+    } else {
+        _date_format_timer.cancel();
+    }
```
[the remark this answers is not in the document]
nyh: I think Copilot is right here. I'm also not sure if it's legal to call cancel() on an unarmed timer (maybe it is, I don't know).
Please try to add to the test two consecutive calls to set_generate_date_header(true), as well as two consecutive calls to set_generate_data_header(false), and see if it works ok.

avikivity:
> Add two independently controllable server-header options to http_server:
> 
> * _server_header (sstring, default "Seastar httpd"): emitted as the
>   Server header value; empty string suppresses the header entirely.
> * _generate_date_header (bool, default true): when false, the Date
>   header is omitted and the periodic date-update timer is stopped.
> 
> Add write_body overloads in reply to allow setting body without setting Content-Type.
> 
> Fixes #3216

Cover letter (and patch) should say why, not just what.

src/http/httpd.cc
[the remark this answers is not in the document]
m-szymon: fixed.

src/http/reply.cc
```diff
@@ -116,11 +116,19 @@ sstring reply::response_line() const {
...
 }
 
 void reply::write_body(const sstring& content_type, sstring content) {
     set_content_type(content_type);
+    write_body(std::move(content));
+}
+
+void reply::write_body(sstring content) {
     _content = std::move(content);
 }
```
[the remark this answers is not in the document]
m-szymon: fixed

include/seastar/http/reply.hh
m-szymon: I decided to remove these overloads.
While empty header value in general is valid, for Content-Type it actually has to have some token.
So using empty string is workaround to not set a Content-Type. Maybe not the most elegant on, but it should be a rare case.

include/seastar/http/httpd.hh
```diff
@@ -182,6 +184,21 @@ public:
 
     void set_content_streaming(bool b);
 
+    /// Returns the value of the "Server" header that will be added to each response.
+    /// An empty string means the header will not be added.
+    const sstring& get_server_header() const;
+
+    /// Sets the value of the "Server" header added to each response.
+    /// Pass an empty string to suppress the header entirely.
+    void set_server_header(sstring value);
```
avikivity: I think std::optional\<sstring> is preferable here. Separate data (the server name) from metadata (do we have a server name).

include/seastar/http/reply.hh
```diff
@@ -202,8 +202,16 @@ struct reply {
...
+     *
+     * An empty string is a no-op (an empty media-type is invalid per
+     * RFC 9110).  To remove a previously set Content-Type use
+     * _headers.erase("Content-Type").  To force an empty value use
+     * add_header("Content-Type", "").
      */
     reply& set_content_type(std::string_view content_type) {
+        if (content_type.empty()) {
+            return *this;
+        }
```
avikivity, line 215: I think it's better to change the signature to std::optional\<std::string_view> to separate data from metadata. Since there is a conversion constructor, it's a compatible change.

include/seastar/http/httpd.hh
nyh: In theory, you're right, but in practice, it doesn't make sense to have an "empty server name", and nobody will want to set an empty server, so it's not terrible to "optimize" this to just sstring...

include/seastar/http/reply.hh
nyh: Why would we want to allow `set_content_type()` on an empty optional? Who would ever want to use that?

m-szymon: Changed to std::optional<std::string_view>. The nullopt leaves Content-Type untouched, while "" still falls back to text/plain as before — so existing code is fully compatible, we just now have an explicit way to not set the header at all.

include/seastar/http/httpd.hh
m-szymon: Changed to std::optional<sstring> as suggested. The reason we have set_server_header at all is to keep the default "Seastar httpd" value backward compatible — if someone wants to set "empty" they can always call set_header directly. But since we're adding the API anyway, optional is cleaner than overloading the meaning of an empty string.

PR #3218, opened 2026-01-23
posix-stack: Reuse parent class _reuseport from child

xemul:
The posix_ap_network_stack inherits from posix_network_stack, but introduces its own bool _reuseport bit. It's OK to re-use one from parent class, both are the same.


src/net/posix-stack.cc
```diff
@@ -949,7 +950,7 @@ posix_ap_network_stack::listen(socket_address sa, listen_options opt) {
         return server_socket(std::make_unique<posix_ap_server_socket_impl>(0, sa, _allocator));
     }
     auto protocol = static_cast<int>(opt.proto);
-    return _reuseport ?
+    return posix_network_stack::_reuseport ?
```
travisdowns, line 953: Is this explicit scoping necessary for some reason?

xemul: No, just my personal preference

PR #3220, opened 2026-01-23
posix-stack: Make internal::posix_connect() resolve exceptions into f…

xemul:
…utures

The helper in question returns a future, but may sometimes throw a valid runtime exception. E.g. when binding to a busy port/path or when the call to file_desc::connect() throws itself.

When the connecting socket is not-unix one, the call to posix_connect() is properly wrapped with futurator::invoke()-s and all exceptions a caught and resolved into futures. However, unix socket connecting calls this method directly, so a thrown exception will be throws further.

Fortunately, nowadays upper level wraps posix_socket_impl::connect() into do_with(), which, in turn, uses futurize_invoke(), but it's fragile, it's better to have safer posix_connected_socket_impl::connect().

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1684,6 +1684,7 @@ reactor::make_pollable_fd(socket_address sa, int proto) {
 namespace internal {
 
 future<> posix_connect(pollable_fd pfd, socket_address sa, socket_address local) {
+  try {
 #ifdef IP_BIND_ADDRESS_NO_PORT
     if (!sa.is_af_unix()) {
         try {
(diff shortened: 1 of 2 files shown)


xemul:

src/core/reactor.cc
```diff
@@ -1684,6 +1684,7 @@ reactor::make_pollable_fd(socket_address sa, int proto) {
 namespace internal {
 
 future<> posix_connect(pollable_fd pfd, socket_address sa, socket_address local) {
+  try {
```
travisdowns, line 1687: making it a couroutine would also solve this, right?

PR #3221, opened 2026-01-23
configure: validate --cook ingredient names

travisdowns:
Fail with a clear error if an invalid ingredient name is passed to --cook, listing all valid ingredients from cooking_recipe.cmake.

This has bitten me more than once, often due to capitalization.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -18,8 +18,10 @@
 #
 import argparse
 import os
+import re
 import seastar_cmake
 import subprocess
+import sys
 import tempfile
 
 tempfile.tempdir = "./build/tmp"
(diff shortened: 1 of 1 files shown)


travisdowns:
@scylladb/seastar-maint please have a look.

configure.py
```diff
@@ -172,6 +174,16 @@ def identify_best_standard(cpp_standards, compiler):
...
 
 
+def get_valid_ingredients():
+    """Extract valid ingredient names from cooking_recipe.cmake."""
+    recipe_path = os.path.join(seastar_cmake.ROOT_PATH, 'cooking_recipe.cmake')
+    with open(recipe_path, 'r') as f:
+        content = f.read()
+    # Match cooking_ingredient(name or cooking_ingredient (name
+    matches = re.findall(r'cooking_ingredient\s*\(\s*([\w-]+)', content)
+    return set(matches)
```
avikivity, line 184: Normally I'd say, have the it return the list programmatically, but it's cmake so...

travisdowns:
Exactly. "Slightly dirty" is the best you can hope for cmake.

PR #3222, opened 2026-01-23
prometheus: support multiple __name__ filters and prefixed names

travisdowns:
## Summary
- Enhance the existing `__name__` filter implementation to handle the case __name__ is passed multiple times, returning any of the provided names
- Support using either prefixed (`seastar_foo*`) or unprefixed (`foo*`) metric names in filters
- Add benchmarks for name filter performance
- Update documentation with filtering best practices

## Motivation
When querying a subset of metrics, it's useful to be able to specify multiple metric names in a single request. Previously only a single `__name__` parameter was supported.

Additionally, users see metric names like `seastar_scheduler_tasks_pending` in prometheus output and would naturally use these full names when filtering. The prefix is an internal implementation detail that users shouldn't need to know about. This change allows both prefixed and unprefixed names to work identically.

## Test plan
- [x] Added unit tests for family_filter functionality (exact match, prefix match, multiple filters, combined filters)
- [x] Added unit tests for prefix stripping behavior
- [x] Added HTTP-level test for multiple `__name__` parameters
- [x] Added performance benchmarks for name filtering
- [x] All prometheus tests pass: `./test.py --mode dev --name prometheus -v`

### DIFF
--- a/doc/prometheus.md
+++ b/doc/prometheus.md
@@ -18,6 +18,11 @@ While seastar still supports the binary protocol, it would be deprecated in a fu
 ## Querying subset of the metrics
 Seastar supports querying for a subset of the metrics by their names and labels.
 
+Filtering is recommended when you only need a subset of the available metrics, especially
+in systems with many metrics. Filtering reduces the processing overhead on the Seastar
+application, results in smaller HTTP responses, and decreases the load on your Prometheus
+server when scraping and storing metrics.
+
 ### Filtering by a metric name
 Use the `__name__` query parameter to select according to a metric name or a prefix.
 
@@ -26,6 +31,19 @@ For example, to get all the http metrics, point your browser to:
 Filtering by name only supports prefix matching.
 
 To query for only the http requests served metric, point your browser to `http://localhost:9180/metrics?__name__=httpd_requests_served`
+
+You can use either the full metric name as it appears in the output (e.g., `seastar_httpd_requests_served`)
+or the name without the prefix (e.g., `httpd_requests_served`). Both forms work identically.
+
+#### Multiple name filters
+You can specify multiple `__name__` parameters to query for several specific metrics at once.
+A metric is included if it matches any of the specified names.
+
+For example, to get both the http requests and connections metrics:
+`http://localhost:9180/metrics?__name__=httpd_requests_served&__name__=httpd_connections_total`
+
+This also works with prefix matching:
+`http://localhost:9180/metrics?__name__=httpd_requests*&__name__=httpd_connections*`
 
 ### Filtering by a label value
 The Prometheus protocol uses labels to differentiate the characteristics of the thing that is being measured.
@@ -67,9 +85,9 @@ This can be achieved by adding `__aggregate__=false` to the query string. For ex
 ### Configuring the Prometheus server for picking specific metrics
 The [Prometheus configuration](https://prometheus.io/docs/prometheus/1.8/configuration/configuration/) describes the general Prometheus configuration.
 
-To specify a specific metric or metrics add a `metrics_path` to the scrap config in the prometheue.yml file
+To specify a specific metric or metrics add a `metrics_path` to the scrape config in the prometheus.yml file
 
-For example, the following scrap config, will query for all the http metrics:
+For example, the following scrape config will query for all the http metrics:
 
 ```
   scrape_configs:
@@ -78,5 +96,16 @@ For example, the following scrap config, will query for all the http metrics:
       metrics_path: /metrics
       params:
         __name__: ['http*']
+```
+
+To query for multiple specific metrics, list them in the array:
+
+```
+  scrape_configs:
+    - job_name: selected_metrics
+      honor_labels: true
+      metrics_path: /metrics
+      params:
+        __name__: ['httpd_requests_served', 'httpd_connections_total', 'scheduler*']
 ```
 
--- a/include/seastar/core/prometheus.hh
+++ b/include/seastar/core/prometheus.hh
@@ -69,8 +69,17 @@ future<> add_prometheus_routes(httpd::http_server& server, config ctx);
 namespace details {
 using filter_t = std::function<bool(const metrics::impl::labels_type&)>;
 
+struct write_body_args {
+    filter_t filter;
+    sstring metric_family_name;
+    bool use_protobuf_format;
+    bool use_prefix;
+    bool show_help;
+    bool enable_aggregation;
+};
+
 class test_access {
-    future<> write_body(config cfg, bool use_protobuf_format, sstring metric_family_name, bool prefix, bool show_help, bool enable_aggregation, filter_t filter, output_stream<char>&& s);
+    future<> write_body(config cfg, write_body_args args, output_stream<char>&& s);
 
     friend struct metrics_perf_fixture;
     friend struct ::prometheus_test_fixture;
--- a/include/seastar/core/prometheus.hh
+++ b/include/seastar/core/prometheus.hh
@@ -68,12 +68,22 @@ future<> add_prometheus_routes(httpd::http_server& server, config ctx);
 
 namespace details {
 using filter_t = std::function<bool(const metrics::impl::labels_type&)>;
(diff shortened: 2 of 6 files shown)


xemul:
Who (what other code?) is the user of multiple names filtering? Presumably we're not the first who adds filtering to prometheus exporting

travisdowns:
> Who (what other code?) is the user of multiple names filtering?

I am not sure I understand the question, but from our end we use this (and/or want to use) in cases where we care only about a small subset of metrics. E.g., I have somethign like a "redpanda top" application which shows in ~real time the behavior of sesatar + redpanda, and most of the information it collects to enable this is from metrics: we want targeted filtering because it is much cheaper to target the 10-20 metrics we want than to query all metrics (there are thousands at least, up to ~1 million), or to issue 10-20 individual name queries. This means we can update the metrics fairly frequently without incurring a large overhead.

Similarly we are already heavily using __name__ filter in our integration testing, in order to validate that specific metrics behave in the expected way. Using __name__ speed up these queries a ton and increases the test reliability too (since queries don't time out): however we currently need to issue N queries filtered on __name__.

Filtering is really useful at the endpoint as evidenced by the variety of filtering options which are already there (including rich label-based filtering based on regex, but which somehow doesn't work for name).

>  Presumably we're not the first who adds filtering to prometheus exporting

I did not understand the question/comment.

xemul:
Sorry, let me try to rephrase the question to be more clear. You already partially replied, but still.

In general, I was curious whether "filter metrics by names when scraping" idea is a unique invention from Redpanda, or are there already known examples of exporters/scrapers that do it?

Let's take, for example, prometheus, who's the existing known user of the exporting endpoint. For an applications that export their metrics for prometheus, the exporting format is defined and should be supported by the application. Respectively, if prometheus supports "filtering" (or maybe it doesn't -- that's my question), then the way it specifies this filter should also be defined somewhere. Is it?

Or, maybe, there's some other monitoring/diagnosting application that re-uses the prometheus-compatible metrics exporting APIs, and this application has the filtering ability. Do you know any? The mentioned "redpanda top" is great example, maybe there are others.

And regarding

> Presumably we're not the first who adds filtering to prometheus exporting

I meant -- provided there exists an application that scrapes filtered metrics metrics (or -- prometheus has a way to configure filtering), there must already be some known way to request such a filtering. I'm afraid that we're inventing some uniqe API for metrics filtering, that will not be compatible with existing one(s) (if they exist, of course)

xemul:
Like [here](https://last9.io/docs/how-to-scrape-only-selective-metrics-in-prometheus/)

> Prometheus scrape config with filtering
>```
>    global:
>      scrape_interval: 1m
>    
>    scrape_configs:
>      - job_name: 'node-exporter-01'
>        static_configs:
>          - targets: [ 'localhost:9100' ]
>        metric_relabel_configs:
>        - source_labels: [__name__]
>          action: keep
>          regex: '(node_cpu)'
>```
> This will scrape all metrics, but drop anything that does not match the entries in the regex section.

Yes, it says that it will scrape all, then drop unrelated. But maybe there's some way (ongoing effort in prometheus, or some other monitoring app) to ask application not to export unrelated right when scraping

avikivity:
In other words, is this a standard or nonstandard use of __name__?

travisdowns:
Thanks, I understand now.

> In general, I was curious whether "filter metrics by names when scraping" idea is a unique invention from Redpanda, or are there already known examples of exporters/scrapers that do it?

To clarify, I don't think any of this is from Redpanda? It's a pre-existing feature in Seastar, I am pretty sure it even precedes the existence of Redpanda. See [this commit](https://github.com/scylladb/seastar/commit/887ee9baa496f933191d9193589454225e7f0147) from @amnonh .

I am only making it work when the query parameter is specified multiple times, as you would expect it to (and which uses the newly added http support for handling multiple query parameters, thanks for that!).

Prometheus supports this since it doesn't care about query parameters, it just passes them along (see also the doc linked in the above commit). Note also that prometheus is "also just a format" so other tools/applications will scrape this endpoint. 

To try to answer your underlying question anyway: based on a cursory search, there are precious few "standards" relating to this. Prom does have the philosophy that you should probably expose all the metrics, not try to cut them down, and use prom side features or just avoid querying the ones you don't care about. They also say things like "dozens of metrics, not hundreds", so seastar with 100 shards has already blown way past the recommendations. Filtering at the source as seastar does is an order of magnitude more efficient (in my tests).

Some clients (go, python) do support filtering by name using the query parameter, like so `GET /metrics?name[]=http_requests_total&name[]=process_cpu_seconds_total` (AI tells me).

travisdowns:
> Some clients (go, python) do support filtering by name using the query parameter, like so `GET /metrics?name[]=http_requests_total&name[]=process_cpu_seconds_total` (AI tells me).

If you like, I can optionally support that syntax.

travisdowns:
Here's where the python metrics client does the same thing we are doing in seastar, with the slightly different `name[]` syntax:

https://github.com/prometheus/client_python/blob/a8541354519d04852d24688845f1d2d495eef59c/prometheus_client/exposition.py#L120-L121

avikivity:
@amnonh can you explain this `__name__` thing?

travisdowns:
FWIW I updated the first item in the cover list to clarify the extent of that change which I think is where some of the confusion lies.

travisdowns:
Gentle ping @amnonh .

amnonh:
> @amnonh can you explain this `__name__` thing?

yes, seastar promethues api supports filtering by labels. Prometheus treats the metric name as ```__name__``` label (that's because labels that start with __ are not stored).

Filtering by name was from a very early stage, and supports only prefix match, this way, it does not need to go over all the metrics and has little overhead (metrics are sorted by names, so you can quickly find the right metric).

Regular labels support regex matching.
I never added multi-label support as there were no requests for it.

travisdowns:
> I never added multi-label support as there were no requests for it.

Thanks!

At the time you added the filtering support, seastar http handling didn't properly support repeated query parameters, but now it does so in this change I extended the filtering to work when multiple name filters are passed. For us it is very useful since for some applications we target only a few specific metrics and filtering greatly reduces the cost (to seastar, over the wire and on the receiving end) of polling these metrics.

Does it make sense to you? (I think @avikivity is looking for an answer along these lines).

amnonh:
@travisdowns I can see why it could be useful; this is why I added the filtering in the first place. There is the question of the filtering logic of ```and``` vs ```or```,
When we include multiple labels, we expect that all of them would match (logical AND) and here it will be a logical OR.

I don't think that's a problem.

As a side note, if you have lots of metrics that you don't really need, you can enable/disable them at run time, which will reduce most of the overhead.

travisdowns:
> here it will be a logical OR.

Right, predicates for multiple _different_ labels are ANDed together, but multiple possible values for the name predicate are using OR, as AND would never make sense (as a given series can never have > 1 name labels).

> As a side note, if you have lots of metrics that you don't really need, you can enable/disable them at run time, which will reduce most of the overhead.

Right. It's more like we do want/need all of these metrics in the general case (or at least cannot easily remove them for compatibility), but certain tools want to target a very few specific metrics.

travisdowns:
@scylladb/seastar-maint looks like this one is on track for merging, please have a look?

amnonh:
> > here it will be a logical OR.
> 
> Right, predicates for multiple _different_ labels are ANDed together, but multiple possible values for the name predicate are using OR, as AND would never make sense (as a given series can never have > 1 name labels).

I'm not against the general request as it seems there is a real need for it. I'm also not against this PR, just checking if you considered the alternatives.
Other labels get full regex support, so to match multiple values, you can use the regex ```|```.
You could have continue with the semi-reg matching and used ```|``` as a delimiter for ```___name__```.

> 
> > As a side note, if you have lots of metrics that you don't really need, you can enable/disable them at run time, which will reduce most of the overhead.
> 
> Right. It's more like we do want/need all of these metrics in the general case (or at least cannot easily remove them for compatibility), but certain tools want to target a very few specific metrics.

You can also use hidden labels (i.e., labels that start with ```__``` and are not reported) and query by those labels, again, just so you'll know you have the option.

PR #3225, opened 2026-01-29
coroutine: introduce coroutine::schedule_at_coroutine_exit()

denesb:
Schedules a lambda to be run right after the coroutine finishes (whether with success or failure). The lambda can optionally accept an `std::exception_ptr`, which will contain the coroutine's exception, if there was one. Coroutines opt into this facility by wrapping their return type in `coroutine::with_at_exit<>`, so coroutines that don't use it pay
no overhead (see the implementation notes below). Example:
```c++
future<coroutine::with_at_exit<>> coroutine_foo() {
    co_await coroutine::schedule_at_coroutine_exit([] { // async cleanup });
    // stuff (that might throw)
}
```

This roughly equivalent to:
```c++
future<> bar() {
    co_await coroutine_foo().finally([] { // async cleanup });
}
```

The functor passed to at coroutine exit is executed while the coroutine frame is still alive, so capturing local variables by reference is safe.

Implementation notes
====================

Opting in via the return type
-----------------------------

Supporting coroutine::schedule_at_coroutine_exit() requires extra state and work in the coroutine promise (see below), which would add overhead to *every* coroutine -- around 10 extra instructions for the simplest
coroutines -- even those that never use the facility. To avoid this, the facility is opt-in: a coroutine enables it by returning `future<coroutine::with_at_exit<T>>` instead of the plain `future<T>`.

The wrapper selects a different std::coroutine_traits specialization, coroutine_traits_with_at_exit_base<T>, whose promise carries the at-exit machinery. Plain `future<T>` coroutines keep using the original zero-overhead promise (coroutine_traits_base<T>), whose final_suspend() returns std::suspend_never and which instantiates none of the machinery.

`coroutine::with_at_exit<T>` wraps a value of type T and decays to it transparently, so callers observe the awaited result as if it were a plain T. coroutine::schedule_at_coroutine_exit() static_asserts that the enclosing coroutine opted in, producing a clear diagnostic otherwise.

Running the at-exit functions
-----------------------------

To run the registered functions, the final_suspend() extension point is used. This is invoked by the coroutines runtime after the coroutine body finished executing and the coroutine's future already resolved with
either an exception or a value, but before the coroutine frame is destroyed. final_suspend() can return an awaiter (final_awaiter in this case), which is executed with the usual semantics.

Normally, seastar schedules the coroutine's waiter immediately upon making the coroutine's future ready, even before the coroutine frame is destroyed. If the coroutine has registered coroutine::schedule_at_coroutine_exit() functors, this is undesirable: the waiters should be scheduled *after* all at exit functors finished, both to preserve ordering (follow-up code runs after cleanup) and to be able to propagate errors. To this end, when there are at-exit
function(s) registered, the coroutine's promise is split into proxy and return promises. The resolution of the coroutine's body is stored in the proxy promise and the return promise is used for the final result, which
combines the proxy promise and the result of the at exit functors.

If there is no at_exit function registered, the final_awaiter returned from final_suspend() will return true from final_awaiter::await_ready() and the subsequent call to final_awaiter::await_resume() is a no-op, resuming the coroutine's destruction immediately. This is the fast path for an opt-in coroutine that happened not to register any functor.

If there was one or more calls to coroutine::schedule_at_coroutine_exit(), final_awaiter::await_ready() will return false and the subsequent call to final_awaiter::await_suspend() will loop over the at exit functors in LIFO order, executing them one-by-one. After all finished, the return promise is resolved with the combined result of the proxy promise and the combined result of the at exit functors.

### DIFF
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -45,6 +45,73 @@ void free_sized(void* ptr, size_t size);
 
 namespace seastar {
 
+namespace coroutine {
+
+/// \brief Return-type wrapper that opts a coroutine into asynchronous cleanup.
+///
+/// By default, a coroutine returning \c future<T> has no support for
+/// \ref coroutine::schedule_at_coroutine_exit() and pays no overhead for it.
+/// To opt into asynchronous cleanup, declare the coroutine to return
+/// \c future<coroutine::with_at_exit<T>> instead of \c future<T>. This selects
+/// a different \c std::coroutine_traits specialization whose promise carries the
+/// machinery required by \ref coroutine::schedule_at_coroutine_exit().
+///
+/// \c with_at_exit<T> wraps a value of type \c T and decays to it transparently,
+/// so callers can treat the awaited result as if it were a plain \c T:
+/// ```
+/// future<coroutine::with_at_exit<int>> foo();
+///
+/// future<> bar() {
+///     int x = co_await foo();  // with_at_exit<int> decays to int
+/// }
+/// ```
+///
+/// \tparam T the wrapped value type (may be \c void).
+template <typename T = void>
+class with_at_exit {
+    T _value;
+
+public:
+    using value_type = T;
+
+    with_at_exit() = default;
+
+    /// Construct from any value convertible to \c T. Used by the coroutine
+    /// promise to wrap the \c co_return-ed value.
+    template <typename U>
+    requires std::is_convertible_v<U&&, T>
+    with_at_exit(U&& value) noexcept(std::is_nothrow_constructible_v<T, U&&>)
+        : _value(std::forward<U>(value)) {
+    }
+
+    /// \name Transparent decay to the wrapped value.
+    /// \{
+    operator T&() & noexcept { return _value; }
+    operator const T&() const& noexcept { return _value; }
+    operator T&&() && noexcept { return std::move(_value); }
+    /// \}
+
+    /// \name Explicit access to the wrapped value.
+    /// \{
+    T& value() & noexcept { return _value; }
+    const T& value() const& noexcept { return _value; }
+    T&& value() && noexcept { return std::move(_value); }
+    /// \}
+};
+
+/// \brief \c void specialization of \ref with_at_exit.
+///
+/// Carries no value; used as the return-type wrapper for cleanup-enabled
+/// coroutines that would otherwise return \c future<>.
+template <>
+class with_at_exit<void> {
+public:
+    using value_type = void;
+    with_at_exit() = default;
+};
+
+} // namespace coroutine
+
 namespace internal {
 
 
@@ -85,6 +152,199 @@ public:
 #endif
 };
 
+using at_exit_function = noncopyable_function<future<>(std::exception_ptr)>;
+
+/// Satisfied only by coroutine promise types that opt into asynchronous cleanup,
+/// i.e. those generated for coroutines returning `future<coroutine::with_at_exit<T>>`.
+/// Used to reject `coroutine::schedule_at_coroutine_exit()` at compile time in
+/// plain `future<T>` coroutines.
+template <typename Promise>
+concept coroutine_promise_with_at_exit = requires (Promise& promise, at_exit_function&& func) {
+    promise.push_at_exit_function(std::move(func));
+};
+
+class at_exit_func_stack {
+    std::optional<at_exit_function> _func;
+    std::unique_ptr<at_exit_func_stack> _next;
+
+public:
+    at_exit_func_stack() = default;
+    explicit at_exit_func_stack(at_exit_function&& func) : _func(std::move(func)) { }
+
+    bool empty() const noexcept { return !_func; }
+
+    at_exit_function& top() noexcept { return *_func; }
+
+    void push(at_exit_function&& func) {
+        auto new_node = _func ? std::make_unique<at_exit_func_stack>(std::move(*this)) : nullptr;
+        _func = std::move(func);
+        _next = std::move(new_node);
+    }
+
+    void pop() noexcept {
+        _func.reset();
+        if (auto next = std::exchange(_next, nullptr)) {
+            new (this) at_exit_func_stack(std::move(*next));
+        }
+    }
+};
+
+template <typename U>
+void destroy_coroutine(seastar::task& coroutine_task) {
+    auto promise_ptr = static_cast<U*>(&coroutine_task);
+    auto hndl = std::coroutine_handle<U>::from_promise(*promise_ptr);
+    hndl.destroy();
+}
+
+inline std::exception_ptr combine_exceptions(std::exception_ptr ex1, std::exception_ptr ex2) noexcept {
+    if (ex1 && ex2) {
+        return std::make_exception_ptr(nested_exception(std::move(ex1), std::move(ex2)));
+    } else if (ex1) {
+        return ex1;
+    } else {
+        return ex2;
+    }
+}
+
+class at_coroutine_exit_base : public seastar::task {
+protected:
+    at_exit_func_stack _funcs;
+
+private:
+    seastar::task* _coroutine_task;
+    seastar::task* _waiting_task;
+    void (*_destroy_coroutine)(seastar::task&);
+
+    std::optional<seastar::future<>> _future;
+    std::exception_ptr _ex;
+
+private:
+    virtual std::exception_ptr get_exception() noexcept = 0;
+    virtual void resume_after_at_exit(std::exception_ptr at_exit_ex) noexcept = 0;
+
+    future<> execute_one() noexcept {
+        try {
+            return _funcs.top()(get_exception());
+        } catch (...) {
+            return seastar::make_exception_future<>(std::current_exception());
+        }
+    }
+
+public:
+    at_coroutine_exit_base() = default;
+    at_coroutine_exit_base(at_coroutine_exit_base&&) = delete;
+    at_coroutine_exit_base(const at_coroutine_exit_base&) = delete;
+
+    template <typename U>
+    void run(std::coroutine_handle<U> hndl) noexcept {
+        _coroutine_task = &hndl.promise();
+        _waiting_task = hndl.promise().waiting_task();
+        _destroy_coroutine = destroy_coroutine<U>;
+        _future = execute_one();
+
+        if (_future->available()) {
+            execute_involving_handle_destruction_in_await_suspend(this);
+        } else {
+            _future->set_coroutine(*this);
+        }
+    }
+
+    virtual void run_and_dispose() noexcept final override {
+        while (_future->available()) {
+            if (_future->failed()) {
+                _ex = combine_exceptions(std::move(_ex), _future->get_exception());
+            }
+
+            _funcs.pop();
+
+            if (!_funcs.empty()) {
+                _future = execute_one();
+                continue;
+            }
+
+            resume_after_at_exit(std::move(_ex));
+            _destroy_coroutine(*_coroutine_task);
+
+            return;
+        }
+
+        _future->set_coroutine(*this);
+    }
+
+    virtual task* waiting_task() noexcept final override {
+        return _waiting_task;
+    }
+};
+
+template <typename T = void>
+class at_coroutine_exit final : public at_coroutine_exit_base {
+    seastar::promise<T>& _proxy_promise;
+    seastar::promise<T> _return_promise;
+
+private:
+    virtual std::exception_ptr get_exception() noexcept final override {
+        return _proxy_promise.get_exception();
+    }
+
+    virtual void resume_after_at_exit(std::exception_ptr at_exit_ex) noexcept final override {
+        auto proxy_fut = _proxy_promise.get_future();
+        if (at_exit_ex) {
+            auto final_ex = combine_exceptions(std::move(at_exit_ex), _proxy_promise.get_exception());
+            proxy_fut.ignore_ready_future();
+            proxy_fut = seastar::make_exception_future<T>(std::move(final_ex));
+        }
+        proxy_fut.forward_to(std::move(_return_promise));
+    }
+
+public:
+    explicit at_coroutine_exit(seastar::promise<T>& promise)
+        : _proxy_promise(promise)
+        , _return_promise(std::exchange(_proxy_promise, seastar::promise<T>{}))
+    { }
+
+    seastar::promise<T>& get_return_promise() { return _return_promise; }
+
+    void push_at_exit_function(at_exit_function&& func) {
+        _funcs.push(std::move(func));
+    }
+};
+
+class final_awaiter final {
+    at_coroutine_exit_base* _at_exit;
+
+public:
+    explicit final_awaiter(at_coroutine_exit_base* at_exit) noexcept : _at_exit(at_exit) { }
+
+    bool await_ready() const noexcept { return !_at_exit; }
+
+    void await_resume() noexcept { }
+
+    template <typename U>
+    void await_suspend(std::coroutine_handle<U> hndl) noexcept { _at_exit->run(hndl); }
+};
+
+// Does not implement the full awaiter control semantics described at
+// https://en.cppreference.com/w/cpp/language/coroutines.html.
+// It is enough if this is compatible with our promise_type<>::final_suspend().
+template <typename U>
+void finalize_and_destroy_coroutine(std::coroutine_handle<U> hndl) {
+    auto final_suspend = hndl.promise().final_suspend();
+    if (final_suspend.await_ready()) {
+        final_suspend.await_resume();
+        hndl.destroy();
+    } else {
+        final_suspend.await_suspend(hndl);
+    }
+}
+
+// Fast path: promise type for coroutines returning `future<T>`.
+//
+// This is the default, zero-overhead promise. It has no support for
+// `coroutine::schedule_at_coroutine_exit()`: `final_suspend()` returns
+// `std::suspend_never`, so the coroutine is not suspended at its final suspend
+// point and no cleanup machinery is instantiated. Coroutines that need
+// asynchronous cleanup opt in by returning `future<coroutine::with_at_exit<T>>`,
+// which selects `coroutine_traits_with_at_exit_base<T>` below instead.
 template <typename T = void>
 class coroutine_traits_base {
 public:
@@ -188,6 +448,139 @@ public:
 
         scheduling_group set_scheduling_group(scheduling_group new_sg) noexcept {
             return task::set_scheduling_group(new_sg);
+        }
+    };
+};
+
+// Opt-in path: promise type for coroutines returning
+// `future<coroutine::with_at_exit<T>>`.
+//
+// Selected by the `std::coroutine_traits` specialization at the bottom of this
+// file. Carries the machinery required by `coroutine::schedule_at_coroutine_exit()`:
+// `final_suspend()` returns a `final_awaiter` that runs any registered at-exit
+// functions before the coroutine frame is destroyed. When no at-exit function is
+// registered, the fast path through `final_awaiter` is taken (see `final_awaiter`).
+//
+// The seastar future/promise carries `with_at_exit<T>` as its value type; the
+// caller-visible future is therefore `future<with_at_exit<T>>`, which decays to
+// `future<T>`-like semantics via `with_at_exit<T>`'s conversions.
+template <typename T = void>
+class coroutine_traits_with_at_exit_base {
+public:
+    using value_type = coroutine::with_at_exit<T>;
+
+    class promise_type final : public seastar::task, public coroutine_allocators {
+        seastar::promise<value_type> _promise;
+        seastar::promise<value_type>* _return_promise;
+        std::unique_ptr<at_coroutine_exit<value_type>> _at_exit;
+
+    public:
+        promise_type() : _return_promise(&_promise) { }
+        promise_type(promise_type&&) = delete;
+        promise_type(const promise_type&) = delete;
+
+        // this non-templated version only exists to support co_returning a braced-init-list
+        void return_value(T&& value) noexcept {
+            _promise.set_value(value_type(std::forward<T>(value)));
+        }
+
+        template<typename U>
+        requires std::is_convertible_v<U&&, T>
+        void return_value(U&& value) noexcept {
+            _promise.set_value(value_type(std::forward<U>(value)));
+        }
+
+        void return_value(coroutine::exception ce) noexcept {
+            _promise.set_exception(std::move(ce.eptr));
+        }
+
+        void set_exception(std::exception_ptr&& eptr) noexcept {
+            _promise.set_exception(std::move(eptr));
+        }
+
+        void unhandled_exception() noexcept {
+            _promise.set_exception(std::current_exception());
+        }
+
+        seastar::future<value_type> get_return_object() noexcept {
+            return _return_promise->get_future();
+        }
+
+        std::suspend_never initial_suspend() noexcept { return { }; }
+        final_awaiter final_suspend() noexcept { return final_awaiter(_at_exit.get()); }
+
+        virtual void run_and_dispose() noexcept override {
+            auto handle = std::coroutine_handle<promise_type>::from_promise(*this);
+            handle.resume();
+        }
+
+        task* waiting_task() noexcept override { return _return_promise->waiting_task(); }
+
+        scheduling_group set_scheduling_group(scheduling_group sg) noexcept {
+            return std::exchange(this->_sg, sg);
+        }
+
+        void push_at_exit_function(at_exit_function&& func) {
+            if (!_at_exit) {
+                _at_exit = std::make_unique<at_coroutine_exit<value_type>>(_promise);
+                _return_promise = &_at_exit->get_return_promise();
+            }
+            _at_exit->push_at_exit_function(std::move(func));
+        }
+    };
+};
+
+template <>
+class coroutine_traits_with_at_exit_base<void> {
+public:
+    using value_type = coroutine::with_at_exit<void>;
+
+    class promise_type final : public seastar::task, public coroutine_allocators {
+        seastar::promise<value_type> _promise;
+        seastar::promise<value_type>* _return_promise;
+        std::unique_ptr<at_coroutine_exit<value_type>> _at_exit;
+
+    public:
+        promise_type() : _return_promise(&_promise) { }
+        promise_type(promise_type&&) = delete;
+        promise_type(const promise_type&) = delete;
+
+        void return_void() noexcept {
+            _promise.set_value(value_type{});
+        }
+
+        void set_exception(std::exception_ptr&& eptr) noexcept {
+            _promise.set_exception(std::move(eptr));
+        }
+
+        void unhandled_exception() noexcept {
+            _promise.set_exception(std::current_exception());
+        }
+
+        seastar::future<value_type> get_return_object() noexcept {
+            return _return_promise->get_future();
+        }
+
+        std::suspend_never initial_suspend() noexcept { return { }; }
+        final_awaiter final_suspend() noexcept { return final_awaiter(_at_exit.get()); }
+
+        virtual void run_and_dispose() noexcept override {
+            auto handle = std::coroutine_handle<promise_type>::from_promise(*this);
+            handle.resume();
+        }
+
+        task* waiting_task() noexcept override { return _return_promise->waiting_task(); }
+
+        scheduling_group set_scheduling_group(scheduling_group new_sg) noexcept {
+            return task::set_scheduling_group(new_sg);
+        }
+
+        void push_at_exit_function(at_exit_function&& func) {
+            if (!_at_exit) {
+                _at_exit = std::make_unique<at_coroutine_exit<value_type>>(_promise);
+                _return_promise = &_at_exit->get_return_promise();
+            }
+            _at_exit->push_at_exit_function(std::move(func));
         }
     };
 };
@@ -318,8 +711,16 @@ auto operator co_await(internal::without_preemption_check<T> f) noexcept {
 
 namespace std {
 
+// Default: plain `future<T>` coroutines get the zero-overhead promise.
 template<typename T, typename... Args>
 class coroutine_traits<seastar::future<T>, Args...> : public seastar::internal::coroutine_traits_base<T> {
+};
+
+// Opt-in: `future<coroutine::with_at_exit<T>>` coroutines get the cleanup-capable
+// promise. This partial specialization is more specialized than the one above and
+// so is preferred when the value type is a `with_at_exit<>`.
+template<typename T, typename... Args>
+class coroutine_traits<seastar::future<seastar::coroutine::with_at_exit<T>>, Args...> : public seastar::internal::coroutine_traits_with_at_exit_base<T> {
 };
 
 } // std
--- a/tests/unit/coroutines_test.cc
+++ b/tests/unit/coroutines_test.cc
@@ -39,6 +39,7 @@
 #include <seastar/coroutine/as_future.hh>
 #include <seastar/coroutine/exception.hh>
 #include <seastar/coroutine/try_future.hh>
+#include <seastar/coroutine/at_coroutine_exit.hh>
 #include <seastar/testing/random.hh>
 #include <seastar/testing/test_case.hh>
 #include <seastar/util/defer.hh>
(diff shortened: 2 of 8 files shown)


gleb-cloudius:
May be I am missing something but why regular RAII will not work? It it to avoid code clutter in case async clean up is needed in a `catch`?

denesb:
> May be I am missing something but why regular RAII will not work? It it to avoid code clutter in case async clean up is needed in a `catch`?

RAII doesn't work if the cleanup is async. If you have non-async cleanup, you don't need `coroutine::finally()`. But unfortunately all our `close()`, `flush()`, `stop()` and etc. are async and need a bunch of boilerplate to work.

include/seastar/coroutine/finally.hh
```diff
@@ -0,0 +1,154 @@
...
+class [[nodiscard]] finally_awaiter {
+    Func _func;
+
+public:
+    explicit finally_awaiter(Func&& func) noexcept : _func(std::forward<Func>(func)) {}
+
+    finally_awaiter(const finally_awaiter&) = delete;
+    finally_awaiter(finally_awaiter&&) = delete;
+
+    bool await_ready() const noexcept { return false; }
```
xemul: If it's never ready, how does `co_await finally(...)` ever proceeds?

nyh:
Hi, sounds like a worthy feature, but I have a few questions about this API.

First, why do you need to `co_await` the result of `coroutine::finally`? What does it "wait" for? What does this co_await do?

Second, with normal "finally", it allows you to add a finally to a particular **block** of code. It's clear visually and conceptually where this "finally" is done. Here it's not clear to me where it exactly happens, or it gives you enough control of when it happens. For this reason, I wonder if "finally" is a good name for this function, or users will find it confusing. Another possibility is calling it `coroutine::cleanup(...)`. As it basically defines a cleanup function for this coroutine.

Third, you mention that "the lambda passed to finally cannot capture variables by references, as the coroutine frame is already destroyed by the time this is called.". This means, for example, that you can't ask this "finally" (or cleanup) function to close() an object that is used by the coroutine. It forces you to use something like a reference-counted shared pointer (or an object that includes this feature intrinsicly, like file) and pass a _copy_ of it into the cleanup function. This sounds like it will complicate the use of this feature, and definitely needs to be documented with examples.

Finally, is it allowed to call finally() multiple times, and they will be run in reverse order, like RAII?

tests/unit/coroutines_test.cc
```diff
@@ -946,3 +947,83 @@ SEASTAR_TEST_CASE(test_try_future) {
...
+    ~canary() {
+        if (_flag) {
+            *_flag = true;
+        }
+    }
+};
+
+template <typename R, typename Func>
+future<R> finally_coroutine(canary c, std::exception_ptr ex, Func&& func) {
+    co_await coroutine::finally(std::move(func));
```
xemul: Probably worth adding a test for two (or more) finally-s calls with co-awaits in between. Like

```
future<> foo() {
    co_await finally(cleanup_one);
    co_await bar();
    co_await finally(cleanup_two);
    co_return;
}
```

include/seastar/coroutine/finally.hh
avikivity: await_suspend schedules the handle immediately.

However, I don't like it. There's lots of overhead here, both allocations and scheduling. This should be on the same order as manually capturing the exception with coroutine::as_future, running the finally body, then returning the exception if we indeed had one.

The closest I can see is

```c++
    co_await coroutine::finally(
        body lambda coroutine,
        cleanup lambda coroutine
     );
```

It adds two extra allocations, but they can be elided with [[clang::coro_await_elidable_argument]], if I can make it work, which I can't.

denesb: Yes, there is a lot of overhead, this is equivalent to attaching a `finally()` continuation to the coroutine's future, in fact this is what happens behind the scenes.
I don't see another way, I found no way to inject code to be executed at the end of the coroutine, after it's future is resolved, but while the coroutine frame is still intact.

denesb:
> Hi, sounds like a worthy feature, but I have a few questions about this API.
> 
> First, why do you need to `co_await` the result of `coroutine::finally`? What does it "wait" for? What does this co_await do?

There is nothing to actually await here, the awaiter is used solely to access the coroutine handle, so we can inject a finally() continuation to the coroutine's future, from within the coroutine itself.

> Second, with normal "finally", it allows you to add a finally to a particular **block** of code. It's clear visually and conceptually where this "finally" is done. Here it's not clear to me where it exactly happens, or it gives you enough control of when it happens. For this reason, I wonder if "finally" is a good name for this function, or users will find it confusing. Another possibility is calling it `coroutine::cleanup(...)`. As it basically defines a cleanup function for this coroutine.

Yes, the name should probably be `coroutine::schedule_cleanup()`.

> Third, you mention that "the lambda passed to finally cannot capture variables by references, as the coroutine frame is already destroyed by the time this is called.". This means, for example, that you can't ask this "finally" (or cleanup) function to close() an object that is used by the coroutine. It forces you to use something like a reference-counted shared pointer (or an object that includes this feature intrinsicly, like file) and pass a _copy_ of it into the cleanup function. This sounds like it will complicate the use of this feature, and definitely needs to be documented with examples.

Yes, you either need a shared pointer or s resource you can copy into the cleanup lambda.

> Finally, is it allowed to call finally() multiple times, and they will be run in reverse order, like RAII?

Yes, it will act just like attaching multiple `finally()` continuations to a continuation chain.

denesb:
This solution is not perfect, there is a large overhead and the way it works harks back to the old continuation days. But with all that said, it solves a real problem.

You can see an example of use here: https://github.com/scylladb/scylladb/pull/28375/commits/ce17725a390482112ddbd9726b6844ba84936d29.
Without `coroutine::finally()`, this code required multiple embedded `try-catch` blocks and ifs checking for exception pointers. It was almost unreadable.

include/seastar/coroutine/finally.hh
```diff
@@ -0,0 +1,154 @@
...
+///     // do things with file (which might throw)
+/// }
+/// ```
+///
+/// Note that by default, `finally` checks for if the task quota is depleted,
+/// which means that it will yield if \ref seastar::need_preempt()
+/// returns true. Use \ref coroutine::finally_without_preemption_check
+/// to disable preemption checking.
+template <FinallyFunc Func>
+class [[nodiscard]] finally : public seastar::internal::finally_awaiter<true, Func> {
```
travisdowns: Nice doc!

The semantics of what happens when `func` throws or returns an exceptional future should be mentioned (or at least refer the reader to future::finally() if the semantics are the same).

travisdowns:
It's a shame to need to use a shared pointer, since the decrement order is deterministic, i.e, the pointer held in the function body is always destroyed first, then the finally. So lifetime is statically known which usually means you shouldn't need a shared pointer.

Maybe a variation of this which accepts an object "with_-style" and returns a reference to it, would be useful for the common case you have an async-RAII object that you want to clean up but still use in the function.

```
file f0 = co_await open_file();
// we want to call f.close() which is async, always, RAII style

file& f1 = co_await with_cleanup(std::move(f0), [](file& f) { return f.close(); });
// use f1 after here

// of course, you normally wouldn't even name f0 but just pass it directly into with_cleanup as a temporary
```

`with_cleanup` returns a reference to the passed-in object (which lives for the lifetime of the coroutine, and a bit more) and also passes it to the cleanup function.

denesb:
> It's a shame to need to use a shared pointer, since the decrement order is deterministic, i.e, the pointer held in the function body is always destroyed first, then the finally. So lifetime is statically known which usually means you shouldn't need a shared pointer.
> 
> Maybe a variation of this which accepts an object "with_-style" and returns a reference to it, would be useful for the common case you have an async-RAII object that you want to clean up but still use in the function.
> 
> ```
> file f0 = co_await open_file();
> // we want to call f.close() which is async, always, RAII style
> 
> file& f1 = co_await with_cleanup(std::move(f0), [](file& f) { return f.close(); });
> // use f1 after here
> 
> // of course, you normally wouldn't even name f0 but just pass it directly into with_cleanup as a temporary
> ```
> 
> `with_cleanup` returns a reference to the passed-in object (which lives for the lifetime of the coroutine, and a bit more) and also passes it to the cleanup function.

This is a good idea, I will experiment with it. It solves the problem of lifetimes and the interface is less surprising.

nyh:
> file f0 = co_await open_file();
> // we want to call f.close() which is async, always, RAII style
> 
> file& f1 = co_await with_cleanup(std::move(f0), [](file& f) { return f.close(); });
> // use f1 after here
> // of course, you normally wouldn't even name f0 but just pass it directly into with_cleanup as a temporary

Interesting idea.  I agree it's nicer without f0, and the risk to accidentally use f0 in the code below. So it looks like:
```
file& f = co_await with_cleanup(co_await open_file(), [](file& f) { return f.close(); });
```
In this API it's still not clear why the outer co_await needs to be there and it looks very artificial. I wonder if it makes more sense to have an overload that takes a future, so you forgo the inner co_await:

```
file& f = co_await with_cleanup(open_file(), [](file& f) { return f.close(); });
```

Now it's clearer why we need to outer co_await, it really needs to wait for a future (the one returned by open_file()) before returning.

denesb:
> > file f0 = co_await open_file();
> > // we want to call f.close() which is async, always, RAII style
> > file& f1 = co_await with_cleanup(std::move(f0), [](file& f) { return f.close(); });
> > // use f1 after here
> > // of course, you normally wouldn't even name f0 but just pass it directly into with_cleanup as a temporary
> 
> Interesting idea. I agree it's nicer without f0, and the risk to accidentally use f0 in the code below. So it looks like:
> 
> ```
> file& f = co_await with_cleanup(co_await open_file(), [](file& f) { return f.close(); });
> ```
> 
> In this API it's still not clear why the outer co_await needs to be there and it looks very artificial. I wonder if it makes more sense to have an overload that takes a future, so you forgo the inner co_await:
> 
> ```
> file& f = co_await with_cleanup(open_file(), [](file& f) { return f.close(); });
> ```
> 
> Now it's clearer why we need to outer co_await, it really needs to wait for a future (the one returned by open_file()) before returning.

The resource you need to clean up may be passed down from the caller, or constructing it may be synchronous, with only the cleanup being asynchronous.
In either case, `file& f = co_await with_cleanup(make_ready_future<file>(std::move(file_from_caller)), [] (file& f) { return f.close(); }));` will look equally artificial

avikivity:
Pass two coroutines, not a future and a coroutine.

```c++
    auto f = co_await open(...);
    co_await coroutine::finally(
        [&] -> future<> {
            do stuffs with f
        },
        [&] -> future<> {
            co_await f.close();
        }
    );
```

This also plays better with coro_await_elidable_arg, assuming we ever get it to work.

travisdowns:
> I agree it's nicer without f0, and the risk to accidentally use f0 in the code below.

Yes. BTW, clang-tidy "use-after-move" is a good check that also helps catch these.

travisdowns:
@avikivity wrote:

> Pass two coroutines, not a future and a coroutine.

This is where I started, but it restores a lot of the verbosity you want to avoid in the first place: and it seems like a giant footgun since the outer `f` is left used-but-in-scope. You pass the object in so it can be an unnamed temporary.

avikivity:
> @avikivity wrote:
> 
> > Pass two coroutines, not a future and a coroutine.
> 
> This is where I started, but it restores a lot of the verbosity you want to avoid in the first place: and it seems like a giant footgun since the outer `f` is left used-but-in-scope. You pass the object in so it can be an unnamed temporary.

We can add wrappers around it that pass a particular object and close/stop it (see closeable.hh). But we need to start with a general primitive.

We can teach people to write `co_return co_wait coroutine::finally(...);` to avoid the dead object.

travisdowns:
> We can teach people to write co_return co_wait coroutine::finally(...); to avoid the dead object.

This helper is to be used anywhere in the function, or multiple times, not necessarily in the return statement.

avikivity:
> > We can teach people to write co_return co_wait coroutine::finally(...); to avoid the dead object.
> 
> This helper is to be used anywhere in the function, or multiple times, not necessarily in the return statement.

Well, people will have to be careful.

And I'm not against a close-at-end helper, I just think we need to start with a general facility and then specialize.

travisdowns:
> And I'm not against a close-at-end helper, I just think we need to start with a general facility and then specialize.

OK but recognize that nobody :stuck_out_tongue: wants:

```
    co_await coroutine::finally(
        [&] -> future<> {
            do stuffs with f
        },
        [&] -> future<> {
            co_await f.close();
        }
    );
```

Because it nests "do stuffs with f" inside a lambda, which if you were willing to do you'd just use then->finally or as_future.finally() or whatever. The main point of the helper as I see it is to use RAII in coroutines with the mainline code being plain code w/o any nesting/not inside a lambda/not needing captures/not needing to think about any "fiasco".

avikivity:
> > And I'm not against a close-at-end helper, I just think we need to start with a general facility and then specialize.
> 
> OK but recognize that nobody 😛 wants:
> 
> ```
>     co_await coroutine::finally(
>         [&] -> future<> {
>             do stuffs with f
>         },
>         [&] -> future<> {
>             co_await f.close();
>         }
>     );
> ```
> 
> Because it nests "do stuffs with f" inside a lambda, which if you were willing to do you'd just use then->finally or as_future.finally() or whatever. The main point of the helper as I see it is to use RAII in coroutines with the mainline code being plain code w/o any nesting/not inside a lambda/not needing captures/not needing to think about any "fiasco".

I find the combination of coroutines and continuations to have poor readability.

Scheduling a continuation a-la seastar::defer() is nice, but there is too much hidden overhead in the solution presented.

denesb:
> > > And I'm not against a close-at-end helper, I just think we need to start with a general facility and then specialize.
> > 
> > 
> > OK but recognize that nobody 😛 wants:
> > ```
> >     co_await coroutine::finally(
> >         [&] -> future<> {
> >             do stuffs with f
> >         },
> >         [&] -> future<> {
> >             co_await f.close();
> >         }
> >     );
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Because it nests "do stuffs with f" inside a lambda, which if you were willing to do you'd just use then->finally or as_future.finally() or whatever. The main point of the helper as I see it is to use RAII in coroutines with the mainline code being plain code w/o any nesting/not inside a lambda/not needing captures/not needing to think about any "fiasco".
> 
> I find the combination of coroutines and continuations to have poor readability.
> 
> Scheduling a continuation a-la seastar::defer() is nice, but there is too much hidden overhead in the solution presented.

The two-lambda solution you presented has no less overhead than what I propose: you still have a coroutine embedded in another coroutine and have to execute two embedded coroutines sequentially.
I think my solution is even more lightweight as the "body" of the coroutine is not embedded into another.

avikivity:
> > > > And I'm not against a close-at-end helper, I just think we need to start with a general facility and then specialize.
> > > 
> > > 
> > > OK but recognize that nobody 😛 wants:
> > > ```
> > >     co_await coroutine::finally(
> > >         [&] -> future<> {
> > >             do stuffs with f
> > >         },
> > >         [&] -> future<> {
> > >             co_await f.close();
> > >         }
> > >     );
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Because it nests "do stuffs with f" inside a lambda, which if you were willing to do you'd just use then->finally or as_future.finally() or whatever. The main point of the helper as I see it is to use RAII in coroutines with the mainline code being plain code w/o any nesting/not inside a lambda/not needing captures/not needing to think about any "fiasco".
> > 
> > 
> > I find the combination of coroutines and continuations to have poor readability.
> > Scheduling a continuation a-la seastar::defer() is nice, but there is too much hidden overhead in the solution presented.
> 
> The two-lambda solution you presented has no less overhead than what I propose: you still have a coroutine embedded in another coroutine and have to execute two embedded coroutines sequentially. I think my solution is even more lightweight as the "body" of the coroutine is not embedded into another.

With [[clang::coro_await_elidable_argument]], the two nested coroutines do not incur allocations for their frames.

I have not yet been able to make it work though.

avikivity:
> 
> I have not yet been able to make it work though.

llvm/llvm-project@98e83adc0104025b6127a128124d7d5079c4e4cb ?

denesb:
I found a way to circumvent the overhead and the reference problem, by returning a custom awaiter from `promise_type::final_suspend()`. This awaiter is created after the coroutine finished, but *before* the frame is destroyed. It can be used to inject cleanup code, then resume the flow after.
I think I will also be able to get rid of the confusing `await`, converting `finally()` to a regular synchronous call.

v2 coming soon(ish).

denesb:
New in v2:
* Completely new approach, based on `promise_type::final_suspend()`. Finally functions are now executed in the context of the coroutine, while the coroutine frame is still alive.
* `coroutine::finally()` now returns `void`, no need to `co_await` it.

This is not completely finished yet, but given how controversial v1 was, I want some early feedback before I polish this to merge-quality. The functionality is there, but the tests need some more work.

tests/unit/coroutines_test.cc
```diff
@@ -946,3 +947,97 @@ SEASTAR_TEST_CASE(test_try_future) {
...
+        co_await do_run(true, sl);
+    }
+};
+
+SEASTAR_TEST_CASE(test_finally) {
+    co_await finally_tester<void, false>{}();
+    //co_await finally_tester<int, false>{}();
+    //co_await finally_tester<void, true>{}();
+    //co_await finally_tester<int, true>{}();
+}
```
raphaelsc, line 1342: why commented out?

nyh:
> This is not completely finished yet, but given how controversial v1 was, I want some early feedback before I polish this to merge-quality. 

I like the new API, which I think looks much less contrived than the previous one:
```
future<> my_coroutine() {
    auto file = co_await open_file_dma("myfile");

    coroutine::finally([&file] (std::exception_ptr eptr) {
       return file.close();
    });

    // do things with file (which might throw)
}
```

I'll leave it to bigger experts to review the implementation and its performance.

Just a thought (feel free to reject or even ignore): I wonder if it might be possible to put a finally not at the very end of the continuation, but at the end of any block, as you can do in normal synchronous code. What I mean is something like this:

```
future<> my_coroutine() {
    if (something) {
        auto file = co_await open_file_dma("myfile");
        auto handle = coroutine::finally([&file] (std::exception_ptr eptr) {
            return file.close();
        });
        // do things with file (which might throw)
    }
   // Additional code we can reach if the above code didn't throw. At this
   // point, the file is already closed (and since it's not in this block, it's unreachable.
}
```

In other words, imagine that finally() returns a "handle", and the cleanup gets called not at the very end of the coroutine, but at the moment than "handle" goes out of scope.
I don't know if this can even be implemented, or if it can, if it's even a good idea, so take this idea with skepticism.

raphaelsc:
> > This is not completely finished yet, but given how controversial v1 was, I want some early feedback before I polish this to merge-quality.
> 
> I like the new API, which I think looks much less contrived than the previous one:
> 
> ```
> future<> my_coroutine() {
>     auto file = co_await open_file_dma("myfile");
> 
>     coroutine::finally([&file] (std::exception_ptr eptr) {
>        return file.close();
>     });
> 
>     // do things with file (which might throw)
> }
> ```
> 
> I'll leave it to bigger experts to review the implementation and its performance.
> 
> Just a thought (feel free to reject or even ignore): I wonder if it might be possible to put a finally not at the very end of the continuation, but at the end of any block, as you can do in normal synchronous code. What I mean is something like this:
> 
> ```
> future<> my_coroutine() {
>     if (something) {
>         auto file = co_await open_file_dma("myfile");
>         auto handle = coroutine::finally([&file] (std::exception_ptr eptr) {
>             return file.close();
>         });
>         // do things with file (which might throw)
>     }
>    // Additional code we can reach if the above code didn't throw. At this
>    // point, the file is already closed (and since it's not in this block, it's unreachable.
> }
> ```
> 
> In other words, imagine that finally() returns a "handle", and the cleanup gets called not at the very end of the coroutine, but at the moment than "handle" goes out of scope. I don't know if this can even be implemented, or if it can, if it's even a good idea, so take this idea with skepticism.

Essentially a futurized variant of defer(). Why not call it defer?

travisdowns:
> Just a thought (feel free to reject or even ignore): I wonder if it might be possible to put a finally not at the very end of the continuation, but at the end of any block, as you can do in normal synchronous code. What I mean is something like this:

It feels like it's not possible, because you don't have an "entry point" into the coroutine machinery at the end of a block: the function is not "split" there into before and after parts by the compiler (unlike at co_await). So I don't see how you can run async code there ... but I could be wrong since I didn't even know you can do what's already been done in this PR.

include/seastar/coroutine/finally.hh
```diff
@@ -0,0 +1,79 @@
...
+///
+/// When the function is called, the coroutine frame is still alive, so capturing
+/// local variables *is* allowed.
+///
+/// Multiple calls to `finally()` are allowed, the functions will be executed in
+/// reverse (LIFO) order.
+///
+/// Exceptions from `func()` will be propagated to the coroutine's caller.
+/// If both the coroutine and one or more finally func throw, the final exception
+/// will be a nested one.
```
travisdowns: It could be clearer about how the nesting will work, is the outermost exception the one from the body? and what is the ordering of the nested finally exceptions if multiple throw.

tests/unit/coroutines_test.cc
denesb: The current v2 is an RFC basically, not completely finished yet. I wanted to get some early feedback before I go to polishing phase.

denesb:
> > Just a thought (feel free to reject or even ignore): I wonder if it might be possible to put a finally not at the very end of the continuation, but at the end of any block, as you can do in normal synchronous code. What I mean is something like this:
> 
> It feels like it's not possible, because you don't have an "entry point" into the coroutine machinery at the end of a block: the function is not "split" there into before and after parts by the compiler (unlike at co_await). So I don't see how you can run async code there ... but I could be wrong since I didn't even know you can do what's already been done in this PR.

Yes, we don't have "hooks" for blocks. We have `initial_suspend()` which is a hook for just before the coroutine body is executed and `final_suspend()` (which is what this PR uses) which is executed right after the coroutine body. There are no extension points in the middle. So `coroutine::finally()` is not a perfect RAII replacement, but it will still be tremendously useful for cutting down boilerplate and makeing code easier to follow.

nyh:
> Yes, we don't have "hooks" for blocks. We have `initial_suspend()` which is a hook for just before the coroutine body is executed and `final_suspend()` (which is what this PR uses) which is executed right after the coroutine body. There are no extension points in the middle.

I'm very far from being an expert in this. A quick search turned up https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1662r0.pdf from 6 years ago, where the author discusses these issues, but https://github.com/cplusplus/papers/issues/445 is not optimistic: "The author tells me they won't bring an update to Prague, because async RAII is hard to retrofit into the current coroutines design. This might require a different approach to coroutines.". I can't find a later discussion on this issue, maybe everyone just gave up. 

> So `coroutine::finally()` is not a perfect RAII replacement, but it will still be tremendously useful for cutting down boilerplate and makeing code easier to follow.

I agree, what you ended up with looks quite useful.

denesb:
> I'm very far from being an expert in this. A quick search turned up https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1662r0.pdf from 6 years ago, where the author discusses these issues, but [cplusplus/papers#445](https://github.com/cplusplus/papers/issues/445) is not optimistic: "The author tells me they won't bring an update to Prague, because async RAII is hard to retrofit into the current coroutines design. This might require a different approach to coroutines.". I can't find a later discussion on this issue, maybe everyone just gave up.

To the best of my knowledge, for async RAII to work, one needs stackful coroutines. Seastar has an implementation of this: `seastar::thread`.
The C++ standard implements stackless coroutines, which is equivalent to the seastar continuation-style code. Both of them struggle with RAII, it is just very hard to implement in a stackless coroutine context.

nyh:
> > I'm very far from being an expert in this. A quick search turned up https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1662r0.pdf from 6 years ago, where the author discusses these issues, but [cplusplus/papers#445](https://github.com/cplusplus/papers/issues/445) is not optimistic: "The author tells me they won't bring an update to Prague, because async RAII is hard to retrofit into the current coroutines design. This might require a different approach to coroutines.". I can't find a later discussion on this issue, maybe everyone just gave up.
> 
> To the best of my knowledge, for async RAII to work, one needs stackful coroutines.

As the aforementioned paper explains (have a look), doing "async RAII" using the traditional C++ destructors is hopeless, and indeed needs either something like threads or "stackful coroutines" or massive changes to the language and how destructors are used. 

So the workable ideas need to revolve around new ideas for adding asynchronous cleanup at the end of scopes (in general, or only the outer coroutine scope as you did), not using the C++ destructor but something else (like file::close()).
ine context.

avikivity:
> > > Just a thought (feel free to reject or even ignore): I wonder if it might be possible to put a finally not at the very end of the continuation, but at the end of any block, as you can do in normal synchronous code. What I mean is something like this:
> > 
> > 
> > It feels like it's not possible, because you don't have an "entry point" into the coroutine machinery at the end of a block: the function is not "split" there into before and after parts by the compiler (unlike at co_await). So I don't see how you can run async code there ... but I could be wrong since I didn't even know you can do what's already been done in this PR.
> 
> Yes, we don't have "hooks" for blocks. We have `initial_suspend()` which is a hook for just before the coroutine body is executed and `final_suspend()` (which is what this PR uses) which is executed right after the coroutine body. There are no extension points in the middle. So `coroutine::finally()` is not a perfect RAII replacement, but it will still be tremendously useful for cutting down boilerplate and makeing code easier to follow.

I think the fact that it's deferred to the end of the coroutine will be surprising.

denesb:
> > > > Just a thought (feel free to reject or even ignore): I wonder if it might be possible to put a finally not at the very end of the continuation, but at the end of any block, as you can do in normal synchronous code. What I mean is something like this:
> > > 
> > > 
> > > It feels like it's not possible, because you don't have an "entry point" into the coroutine machinery at the end of a block: the function is not "split" there into before and after parts by the compiler (unlike at co_await). So I don't see how you can run async code there ... but I could be wrong since I didn't even know you can do what's already been done in this PR.
> > 
> > 
> > Yes, we don't have "hooks" for blocks. We have `initial_suspend()` which is a hook for just before the coroutine body is executed and `final_suspend()` (which is what this PR uses) which is executed right after the coroutine body. There are no extension points in the middle. So `coroutine::finally()` is not a perfect RAII replacement, but it will still be tremendously useful for cutting down boilerplate and makeing code easier to follow.
> 
> I think the fact that it's deferred to the end of the coroutine will be surprising.

Should we call it `coroutine::defer()` instead? Or you think there is a more intuitive behavior that we should implement?

denesb:
New in v3:
* Code is reworked to be easier to follow.
* De-templated finally_awaiter<>.
* Test is fixed and extended to cover cases of multiple finally clauses.

avikivity:
Maybe coroutine::on_coroutine_exit() so it's super explicit.

Provided it has zero cost when not used.

denesb:
New in v4:
* Renamed `coroutine::finally()` to `coroutine::at_coroutine_exit()`.
* Split coroutine promises to proxy and return promises to have full control over scheduling.
* Integration with `coroutines::try_future()` and `coroutines::return_exception_ptr()`. These need to execute the `final_suspend()` semantics, at least to the degree that `final_awaiter` expects it.
* More and better tests

denesb:
> Maybe coroutine::on_coroutine_exit() so it's super explicit.
> 
> Provided it has zero cost when not used.

Cost is not zero unfortunately:
* There is extra storage: one more `seastar::promise<>` object and one `at_exit_func_stack` [^1] object.
* There is one extra `_proxy_promise.forward_to(_return_promise)`.

[^1]: `class at_exit_func_stack { noncopyable_function<future<>(std::exception_ptr)> _func; std::unique_ptr<at_exit_func_stack> _next; };`

denesb:
@avikivity ping for review.

avikivity:
With the extra overhead, I'm not so happy about it.

It's possible to customize the promise object by specializing std::coroutine_traits for a magic argument type.

e.g.

```c++
future<> i_am_finally_enabled(int arg1, int arg2, seastar::coroutine::enable_at_exit_t = {}) {
    ... do things including at_coroutine_exit()
}
```

However, it is awkward to put something in the signature that is part of the implementation.

travisdowns:
1 vote for `coroutine::at_exit`, instead of repeating `coroutine` again in the function name. The whole point of these functions are to be used relatively frequently as "vocabulary" to reduce boilerplate so I think this favors a shorter name.

denesb:
New in v5:
* Pushed cost to actual users of `coroutine_at_exit()` as much as possible. Coroutines not using the at exit functionality now only pay with some storage (can be reduced further at the cost of an extra allocation, i.e. use `std::unique_ptr<>` instead of `std::optional<>`) and a few ifs. No promise->future->promise forwarding for non-users.
* Fixed an assert failure when the at-exit function returns ready future.

> 1 vote for coroutine::at_exit, instead of repeating coroutine again in the function name. The whole point of these functions are to be used relatively frequently as "vocabulary" to reduce boilerplate so I think this favors a shorter name.

I agree.

denesb:
@avikivity please check if the overhead is acceptable now. Non-users should be minimally impacted now.

travisdowns:
There are some coro benches in the repo, maybe run those? 

There are some more here which could be copy-pasted into seastar and should just work:

https://github.com/redpanda-data/redpanda/blob/dev/src/v/utils/tests/coro_bench.cc

> can be reduced further at the cost of an extra allocation, i.e. use std::unique_ptr<>

An extra allocation only in the case at_exit is being used? Or for any co_await?

denesb:
Rebased and renamed to `coroutine::at_exit()`.

denesb:
> > can be reduced further at the cost of an extra allocation, i.e. use std::unique_ptr<>
> 
> An extra allocation only in the case at_exit is being used? Or for any co_await?

Only when at_exit() is used.

denesb:
I ran `test/perf/perf_coroutine`:

Before:
```
single run iterations:    0
single run duration:      1.000s
number of runs:           5
number of cores:          32
random seed:              3151880769
start/stop overhead:      328.000ns (461.000ns)

test                                         iters            runtime     allocs      tasks       inst     cycles   overhead
coroutine_test.empty                     166925544     5.91ns ± 0.01%      1.000      0.000     180.00       32.1        0.0
coroutine_test.without_preemption_check  154611267     6.47ns ± 0.01%      1.000      0.000     194.00       35.1        0.0
coroutine_test.ready                     148977992     6.72ns ± 0.02%      1.000      0.000     202.06       36.4        0.0
coroutine_test.maybe_yield               159132258     6.46ns ± 0.01%      1.000      0.000     184.05       35.1        0.0
coroutine_test.unbuffered_generator        1413549   706.55ns ± 0.02%      2.000      0.003    9084.87     3984.6        0.0
coroutine_test.buffered_generator          2187283   457.09ns ± 0.01%      2.000      0.002    7652.80     2562.9        0.0
```
After:
```
single run iterations:    0
single run duration:      1.000s
number of runs:           5
number of cores:          32
random seed:              3151880769
start/stop overhead:      330.000ns (459.000ns)

test                                         iters            runtime     allocs      tasks       inst     cycles   overhead
coroutine_test.empty                     122534882     8.14ns ± 0.02%      1.000      0.000     231.00       44.1        0.0
coroutine_test.without_preemption_check  119878857     8.36ns ± 0.02%      1.000      0.000     242.00       45.1        0.0
coroutine_test.ready                     112398218     8.90ns ± 0.01%      1.000      0.000     253.07       48.2        0.0
coroutine_test.maybe_yield               125190772     7.99ns ± 0.03%      1.000      0.000     236.07       43.1        0.0
coroutine_test.unbuffered_generator        1464106   682.40ns ± 0.03%      2.000      0.003    9134.69     3843.7        0.0
coroutine_test.buffered_generator          2360654   423.67ns ± 0.02%      2.000      0.002    7704.53     2372.0        0.0
```

Instruction count increased and in most tests there is a regression in runtime. For some reason the generator ones performed better.

denesb:
New in v6:
* Another big refactor, with the goal of reducing the performance overhead (see below)
    - Strip out all logic from `final_awaiter`, this object has to be constructed at the end of each coroutine and it was the single biggest contributor to additional instructions.
    - Logic was moved from `final_awaiter` to `coroutine_at_exit`, which now contains all logic for the at-exit functionality in a single class, no more scattering across several classes.
    - `co_await` is back, in order to shave off more instruction, the code to track the current coroutine was dropped, to schedule a function to run at exit, one has to call `co_await coroutine::schedule_at_exit()`.

New perf_coroutine run (same seed):
```
single run iterations:    0
single run duration:      1.000s
number of runs:           5
number of cores:          32
random seed:              3151880769
start/stop overhead:      274.000ns (361.000ns)

test                                         iters            runtime     allocs      tasks       inst     cycles   overhead
coroutine_test.empty                     161865212     6.18ns ± 0.01%      1.000      0.000     191.00       33.1        0.0
coroutine_test.without_preemption_check  145141573     6.90ns ± 0.01%      1.000      0.000     207.00       37.1        0.0
coroutine_test.ready                     133212208     7.51ns ± 0.08%      1.000      0.000     215.06       40.5        0.0
coroutine_test.maybe_yield               156697322     6.39ns ± 0.02%      1.000      0.000     196.05       34.3        0.0
coroutine_test.unbuffered_generator        1415845   698.93ns ± 0.12%      2.000      0.003    9096.81     3925.0        0.0
coroutine_test.buffered_generator          2352242   424.06ns ± 0.11%      2.000      0.002    7665.53     2363.9        0.0
```

include/seastar/coroutine/finally.hh
denesb: The PR now uses a new approach.

tests/unit/coroutines_test.cc
denesb: Done.

avikivity:
> 1 vote for `coroutine::at_exit`, instead of repeating `coroutine` again in the function name. The whole point of these functions are to be used relatively frequently as "vocabulary" to reduce boilerplate so I think this favors a shorter name.

But at the exit of what? The program? the scope? the coroutine? The two "coroutine" instances mean different things: the first means it's part of the coroutine support library, the second qualifies the word "exit".

include/seastar/core/coroutine.hh
```diff
@@ -85,11 +85,193 @@ public:
...
+
 template <typename T = void>
 class coroutine_traits_base {
 public:
     class promise_type final : public seastar::task, public coroutine_allocators {
         seastar::promise<T> _promise;
+        std::optional<coroutine_at_exit<T>> _at_exit;
+
+    private:
+        seastar::promise<T>& return_promise() noexcept { return _at_exit ? _at_exit->get_return_promise() : _promise; }
```
avikivity: Unfortunately this adds overhead. While it's not a lot of overhead, it's multiplied by all the coroutines ever processed, only a small fraction of which will use this facility.

Can we exchange the promise with the _at_exit promise to avoid this?

Also make _at_exit a unique_ptr to reduce footprint.

denesb: > Can we exchange the promise with the _at_exit promise to avoid this?

I already do this, but after I instantiate `_at_exit`, I have to return a different promise for the return path. The set-{value/exception} paths are unchanged, due to this exchange.
Would saving the current return promise to a pointer to avoid the branch help? Note that this method is invoked once -- when the coroutine either returns or suspends the first time (whichever comes first).

denesb:
New in v7:
* Make `_at_exit` `std::unique_ptr<>` instead of `std::optional<>`, to reduce space overhead.
* Remove `return_promise()`, save address of appropriate return promise in `promise<>* _return_promise` to avoid the branch.
* Rename `coroutine::schedule_at_exit()` to `coroutine::schedule_at_coroutine_exit()`
* Renamed `seastar::internal::coroutine_at_exit` to `seastar::internal::at_coroutine_exit`.

Space overhead is now reduced to `promise<>*` + `std::unique_ptr<>`.
No branches on the no-use path.

```
single run iterations:    0
single run duration:      1.000s
number of runs:           5
number of cores:          32
random seed:              3151880769
start/stop overhead:      336.000ns (460.000ns)

test                                         iters            runtime     allocs      tasks       inst     cycles   overhead
coroutine_test.empty                     156964872     6.16ns ± 0.19%      1.000      0.000     188.00       33.0        0.0
coroutine_test.without_preemption_check  148480074     6.75ns ± 0.08%      1.000      0.000     203.00       36.2        0.0
coroutine_test.ready                     134734981     7.43ns ± 0.03%      1.000      0.000     212.06       39.8        0.0
coroutine_test.maybe_yield               157333765     6.40ns ± 0.01%      1.000      0.000     193.05       34.2        0.0
coroutine_test.unbuffered_generator        1428250   699.68ns ± 0.03%      2.000      0.003    9092.82     3906.3        0.0
coroutine_test.buffered_generator          2335812   426.69ns ± 0.13%      2.000      0.002    7661.55     2362.0        0.0
```

denesb:
@avikivity ping.

denesb:
@avikivity ping^2.

avikivity:
Sorry about the delay.

IMO, the overhead is still too high to accept it for all coroutines.

I can think of ~two~three ways to make it opt-in:

1. Return a different type from the coroutine

If the coroutine returns a non-future, it can choose a different promise type that has the extra facilities. The return type can decay to a future. This will require thunks if we need to match a specific signature (e.g. in a virtual function)

2. Adjust the promise type using the function signature

The promise type is selected using the full signature: `template< class R, class... Args > struct coroutine_traits;`. If we add a trailing coroutine::atexit_type, then we can specialize coroutine_traits to select a different promise_type (likely just a specialization of the main promise_type). This also requires thunks, but at least the return type is a plain future.

3. User-supplied struct in the body

Require the user to prepare a local coroutine::atexit_t atexit; and use it

```c++
    coroutine::atexit_t atexit;
    auto f = co_await open_file();
    co_await atexit([&] { return f.close(); });
    ...
```

Maybe we can somehow std::exchange promise_type::_promise with a fresh promise, so that a later set_value() triggers our core, then we call set_value() on the promise we hijacked.

The first one looks too heavyweight (new return type for a minor feature). The third one looks best but I'm worried about lifetime issues.

meilofveeningen-rl:
> It's a shame to need to use a shared pointer, since the decrement order is deterministic, i.e, the pointer held in the function body is always destroyed first, then the finally. So lifetime is statically known which usually means you shouldn't need a shared pointer.
> 
> Maybe a variation of this which accepts an object "with_-style" and returns a reference to it, would be useful for the common case you have an async-RAII object that you want to clean up but still use in the function.
> 
> ```
> file f0 = co_await open_file();
> // we want to call f.close() which is async, always, RAII style
> 
> file& f1 = co_await with_cleanup(std::move(f0), [](file& f) { return f.close(); });
> // use f1 after here
> 
> // of course, you normally wouldn't even name f0 but just pass it directly into with_cleanup as a temporary
> ```
> 
> `with_cleanup` returns a reference to the passed-in object (which lives for the lifetime of the coroutine, and a bit more) and also passes it to the cleanup function.

We have a project that uses seastar, in which we need something like this. We ended up implementing our own class `cleanup_context` that allows to schedule cleanup tasks and that as a very similar API to this one:

[cleanup_context.hpp.txt](https://github.com/user-attachments/files/28093112/cleanup_context.hpp.txt)

Would this be interesting as a contribution to seastar?

denesb:
> We have a project that uses seastar, in which we need something like this. We ended up implementing our own class `cleanup_context` that allows to schedule cleanup tasks and that as a very similar API to this one:
> 
> [cleanup_context.hpp.txt](https://github.com/user-attachments/files/28093112/cleanup_context.hpp.txt)
> 
> Would this be interesting as a contribution to seastar?

This is useful, but we already have similar primitives, see https://github.com/scylladb/seastar/blob/master/include/seastar/util/closeable.hh. These are still continuation based, so they use `do_with()` and thus cause extra allocations.

meilofveeningen-rl:
> > We have a project that uses seastar, in which we need something like this. We ended up implementing our own class `cleanup_context` that allows to schedule cleanup tasks and that as a very similar API to this one:
> > [cleanup_context.hpp.txt](https://github.com/user-attachments/files/28093112/cleanup_context.hpp.txt)
> > Would this be interesting as a contribution to seastar?
> 
> This is useful, but we already have similar primitives, see https://github.com/scylladb/seastar/blob/master/include/seastar/util/closeable.hh. These are still continuation based, so they use `do_with()` and thus cause extra allocations.

Nice, I did know about `deferred_stop` but not about `with_closeable`, that looks useful! By the way, your `do_stop` and `do_close` functions are `noexcept`, so when using `deferred_stop`/`deferred_close` the `.stop()`/`.close()` are not allowed to throw exceptions? But in general it would be considered OK to have `.stop()`/`.close()` throw? 

(By the way: in our project, we sometimes have the pattern that we have a function that needs to return the closeable object. It creates the object, then performs some operations on then, and finally returns it. We want to use something like `with_closeable` to ensure that, if those operations throw, then `.close` is still called. But for that use case, `with_closeable` does not seem easily applicable?)

denesb:
> > > We have a project that uses seastar, in which we need something like this. We ended up implementing our own class `cleanup_context` that allows to schedule cleanup tasks and that as a very similar API to this one:
> > > [cleanup_context.hpp.txt](https://github.com/user-attachments/files/28093112/cleanup_context.hpp.txt)
> > > Would this be interesting as a contribution to seastar?
> > 
> > 
> > This is useful, but we already have similar primitives, see https://github.com/scylladb/seastar/blob/master/include/seastar/util/closeable.hh. These are still continuation based, so they use `do_with()` and thus cause extra allocations.
> 
> Nice, I did know about `deferred_stop` but not about `with_closeable`, that looks useful! By the way, your `do_stop` and `do_close` functions are `noexcept`, so when using `deferred_stop`/`deferred_close` the `.stop()`/`.close()` are not allowed to throw exceptions? But in general it would be considered OK to have `.stop()`/`.close()` throw?

We consider `stop()` and `close()` to be like destructors. A destructor is not allowed to fail.

> (By the way: in our project, we sometimes have the pattern that we have a function that needs to return the closeable object. It creates the object, then performs some operations on then, and finally returns it. We want to use something like `with_closeable` to ensure that, if those operations throw, then `.close` is still called. But for that use case, `with_closeable` does not seem easily applicable?)

If the function throws, then it doesn't return the closable and you cannot clean up from the outside. You have to clean up inside the function.

denesb:
New in v7 (https://github.com/denesb/seastar/releases/tag/coroutine-finally-v7):
* Introduce a new coroutine traits specialization for coroutines which want to opt-in for `coroutines::at_coroutine_exit()`. The new coroutine traits is selected by wrapping the coroutine's return type with `seastar::coroutines::with_at_exit<>`. Note that this wraps the value type, not the future, so the return type changes from `future<int>` -> `future<coroutine::with_at_exit<int>>`. The `with_at_exit<>` transparently decays to value type. Coroutines which do not opt in pay no price at all.
* Expand the perf test suite with variants which use the new coroutine traits.
* Split the single (now too large) commit into 4 smaller ones.

## Performance impact of `coroutine::schedule_at_coroutine_exit()`

Measured in release on x86-64 (`--smp 1`). Instruction counts are deterministic;
per-coroutine figures are isolated by differencing the 40-field and 10-field
per-field benchmarks so fixed loop overhead cancels.

| Case | Return type | Δ instructions / coroutine | Δ allocations / coroutine | vs. before the feature |
| --- | --- | --- | --- | --- |
| No opt-in | `future<T>` | 0 | 0 | identical to master |
| Opt-in, not used | `future<coroutine::with_at_exit<T>>`, never call `schedule_at_coroutine_exit()` | ≈ +18 | 0 | — |
| Opt-in and used | same, register one immediately-ready function | ≈ +440 | +1 | — |

Absolute per-coroutine instruction figures behind the deltas: plain ≈ 214,
opt-in-unused ≈ 232, opt-in-used ≈ 654.

denesb:
@avikivity please review, I have eliminated all overhead from non-users by using a variant of option 1 from https://github.com/scylladb/seastar/pull/3225#issuecomment-4091357487.

denesb:
@avikivity ping.

denesb:
@avikivity please review.

PR #3226, opened 2026-01-29
treewide: remove remnants of SEASTAR_MODULE

avikivity:
In 719d7880eca70 ("modules: make module support standards-compliant") we unifdefed the source tree for SEASTAR_MODULE and removed the definition. However, the unifdef was incomplete since unifdef terminates when it sees a C++ literal with "'" separators.

Here we complete the unifdefing. Since SEASTAR_MODULE is never defined, it's a no-op for current compilers, and allows clang 23 to build the tree.

### DIFF
--- a/src/core/alien.cc
+++ b/src/core/alien.cc
@@ -20,9 +20,6 @@
  * Copyright (C) 2018 Red Hat
  */
 
-#ifdef SEASTAR_MODULE
-module;
-#endif
 
 #include <compare>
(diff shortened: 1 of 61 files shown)


nyh:
Hats off to Copilot for an excellent review as usual (it didn't have comments, but it summarized the patch very well, even clarifying a detail that I didn't understand in Avi's original patch).

PR #3227, opened 2026-01-29
set_iterator: increase compatibility with C++ ranges

avikivity:
Define the iterator_concept type to communicate the iterator type, and define the default constructor to make it a semiregular type. The latter is required to interoperate with libstdc++ 20.

### DIFF
--- a/include/seastar/core/bitset-iter.hh
+++ b/include/seastar/core/bitset-iter.hh
@@ -136,10 +136,16 @@ private:
     }
 public:
     using iterator_category = std::input_iterator_tag;
+    using iterator_concept = std::input_iterator_tag;
     using value_type = int;
     using difference_type = std::ptrdiff_t;
     using pointer = int*;
     using reference = int&;
+
+    set_iterator() noexcept
+        : _bitset()
+        , _index(0) {
+    }
 
     set_iterator(std::bitset<N> bitset, int offset = 0) noexcept
         : _bitset(bitset)


include/seastar/core/bitset-iter.hh
```diff
@@ -136,11 +136,14 @@ private:
     }
 public:
     using iterator_category = std::input_iterator_tag;
+    using iterator_concept = std::input_iterator_tag;
     using value_type = int;
     using difference_type = std::ptrdiff_t;
     using pointer = int*;
     using reference = int&;
 
+    set_iterator() = default;
```
travisdowns: why not just assign `_index = 0` in this case

while it may be strictly within the standard to have undefined values in a default constructed input iterator (but _not_ for other iterator types), this just seems like a big footgun for no gain: uninitailized value lurking around: it is even forbidden to use them on the RHS of an assignment statement, which validly happen (imagine an object itself default initialized which contains an iterator, which is then assigned)

avikivity: I don't mind. In practice, in modern C++ one rarely instantiates such an iterator, and when one does, one's optimizer covers up for one's lazy coding practice.

I expect it's mostly seen in constructs such as

```c++
    iterator i;
    if (something) {
         i = this_;
    } else {
         i = that;
    }
```

or very old code.

avikivity:
v2: have the default constructor build a valid object, even if the Standard says one mustn't use it.

PR #3228, opened 2026-01-30
test: fix race condition in test_connection_statistics

travisdowns:
The test was flaky because it queried connection stats immediately after opening a TCP connection. While socket.connect() returns once the client-side TCP handshake completes, the server may not have finished accepting the connection and updating its stats counter yet.

Fix by issuing a version command on each new connection before querying stats, ensuring the server has fully processed the connection.

It has been flaking quite a bit in GHA CI.

Before this change it fails fairly reliably within ~10 runs for me locally, after I could not get it to fail in ~100 runs.

### DIFF
--- a/apps/memcached/tests/test_memcached.py
+++ b/apps/memcached/tests/test_memcached.py
@@ -210,9 +210,14 @@ class TcpSpecificTests(MemcacheTest):
     @slow
     def test_connection_statistics(self):
         with tcp_connection() as conn:
+            # Issue some command to ensure server has fully processed this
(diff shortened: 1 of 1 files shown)


travisdowns:
@scylladb/seastar-maint please take a peek at this 11 y/o code :)

avikivity:
> 
> It has been flaking quite a bit in GHA CI.


What is GHA?

travisdowns:
> What is GHA?

Sorry GitHub Actions, i.e., the CI we use for this repo.

PR #3231, opened 2026-02-02
io_tester: Revive IO bandwidth configuration

xemul:
The option to apply bandwidth limit on a class was added long ago (ff6f57faa5), but was accidentally temporary removed while unifying CPU and IO classes (ea95dd8d85)


travisdowns:
~WDYT about a unit test for this?~

~Or failing that, were you able to test this manually say with `io_tester` to see that's it's working?~

~We rely on this in Redpanda but somehow didn't notice it was missing.~

I missed that it was only in io_tester, oops.

PR #3233, opened 2026-02-02
Add pre-commit with trailing whitespace hook

travisdowns:
Enough with the trailing whitespace, let us ban it in the repo. Also adds pre-commit to manage hit hooks for the repo, so  folks can opt into that if they don't like their change being rejected by CI.

We use [pre-commit](https://pre-commit.com/) to execute both the check locally, if the user has opted into it, and also to run the check in CI.

See an example of it failing here: https://github.com/scylladb/seastar/actions/runs/21596526786/job/62230437469

In practice this hook runs ~instantly locally, though that's something you have work at as you add more hooks.

## Summary
- Add pre-commit framework with trailing-whitespace hook
- Add GitHub Actions workflow to enforce pre-commit checks on PRs
- Add developer documentation for local pre-commit setup
- Remove existing trailing whitespace from source files

## Test plan
- [x] `pre-commit run --all-files` passes locally
- [x] CI workflow runs successfully

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,5 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
+    hooks:
+      - id: trailing-whitespace
--- a/HACKING.md
+++ b/HACKING.md
@@ -189,3 +189,43 @@ g++ -o foo_bar foo.o bar.o $(pkg-config --libs --static /path/to/seastar.pc)
 ```
 
 The `--static` flag is needed to include transitive (private) dependencies of `libseastar.a`.
+
+## Development Tools
+
+### Pre-commit hooks
+
+This project uses [pre-commit](https://pre-commit.com/) to enforce some basic checks. These
+checks run in CI, but you can also run them locally as a pre-commit hook as follows.
+
+#### Installation
+
+[Install pre-commit](https://pre-commit.com/#install), following those instructions or
+perhaps using `uv` to avoid polluting your global environment:
+
+```
+uv tool install pre-commit
+```
+
+#### Setup
+
+Install the git hooks:
+
+```
+pre-commit install
+```
+
+This will run the configured hooks automatically on every commit.
+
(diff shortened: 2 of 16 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
+    hooks:
+      - id: trailing-whitespace
```
avikivity, line 5: How does this work? I remember pre-commit hooks as running on the client, but who processes this yaml?

HACKING.md
```diff
@@ -189,3 +189,43 @@ g++ -o foo_bar foo.o bar.o $(pkg-config --libs --static /path/to/seastar.pc)
 ```
 
 The `--static` flag is needed to include transitive (private) dependencies of `libseastar.a`.
+
+## Development Tools
+
+### Pre-commit hooks
+
```
avikivity, line 196: Ok, this could have been part of the previous commit.

avikivity (approved):

.pre-commit-config.yaml
travisdowns: This YAML is processed by a program called pre-commit https://pre-commit.com/, which manages your pre-commit hooks. Git only exposes a single hook, so in general you need something like this to manage multiple hooks. It also comes with a bunch of hooks you want out of the box. 

Folks can use it or not locally, but it also runs in CI, so will flag any out of spec changes.

This is in part covered in the HACKING.md change, and I have updated the cover letter as well.

HACKING.md
travisdowns: Because you review commits in-order, right?

So in general doc commits should come first.

I had done it in the opposite order, but I will reverse it in the future or squash them.

travisdowns:
@scylladb/seastar-maint please review and consider merging.

HACKING.md
avikivity: Best in the same commit. X + doc-of-X are logically one unit.

PR #3237, opened 2026-02-03
test: add fuzz testing infrastructure and sstring fuzzer

travisdowns:
Initial fuzzing support, only supports fuzzing stuff which doesn't require the reactor. Adding reactor support now.

The only bug the fuzzer found for sstring was the `npos` one (see #3236), so that's good I guess. I did test it by introducing a few subtle bugs and it found them.

## Summary

- Add fuzz testing infrastructure using libFuzzer
- Add differential fuzzer for sstring using std::string as oracle

### Infrastructure changes
- Add 'fuzz' build mode to configure.py
- Add fuzzer sanitizer components to FindSanitizers.cmake
- Add `seastar_add_fuzz_test()` CMake function
- Document fuzz testing in doc/testing.md

### sstring fuzzer
The fuzzer interprets input bytes as a sequence of operations, applying each to both an sstring and std::string, then verifying equivalence. Coverage includes:
- Construction (copy, move, fill, iterator, raw data)
- Mutation (append, resize, erase, replace, insert, swap)
- Search (find, find_last_of, starts_with, ends_with, contains)
- Comparison (==, !=, <, >, <=, >=, compare)
- Access (at, front, back, operator[], c_str, data)
- Iteration (forward, reverse, distance)

## Test plan

```bash
./configure.py --mode fuzz --compiler clang++
ninja -C build/fuzz fuzz_tests
./build/fuzz/tests/fuzz/sstring_fuzz -runs=10000
```

### DIFF
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -49,3 +49,60 @@ seastar_add_test(woof_test
   SOURCE woof.cc)
 ```
 
+## Fuzz Testing
+
+Seastar supports fuzz testing using [libFuzzer](https://llvm.org/docs/LibFuzzer.html) for testing non-reactor code. Fuzz tests are built only in `fuzz` mode.
+
+### Building Fuzz Tests
+
+Configure with fuzz mode (requires Clang):
+
+```bash
+./configure.py --mode fuzz --compiler clang++ [other options]
+```
+
+Build all fuzz tests:
+
+```bash
+ninja -C build/fuzz fuzz_tests
+```
+
+### Running Fuzz Tests
+
+Fuzz test executables are located in `build/fuzz/tests/fuzz/`. Run a fuzzer:
+
+```bash
+# Basic run (runs indefinitely until stopped or crash found)
+./build/fuzz/tests/fuzz/sstring_fuzz
+
+# With a corpus directory (recommended for persistent fuzzing)
+mkdir -p corpus/sstring
+./build/fuzz/tests/fuzz/sstring_fuzz corpus/sstring
+
+# Limit iterations
+./build/fuzz/tests/fuzz/sstring_fuzz -runs=10000
+
+# Limit time (in seconds)
+./build/fuzz/tests/fuzz/sstring_fuzz -max_total_time=60
+```
+
+For more libFuzzer options, see the [libFuzzer documentation](https://llvm.org/docs/LibFuzzer.html).
+
+### Adding Fuzz Tests
+
+Fuzz tests are added in `tests/fuzz/CMakeLists.txt`:
+
+```cmake
+seastar_add_fuzz_test(mycomponent
+  SOURCES mycomponent_fuzz.cc)
+```
+
+The fuzz test source must define `LLVMFuzzerTestOneInput`:
+
+```cpp
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+    // Test code here using data[0..size-1] as input
+    return 0;
+}
+```
+
--- a/tests/fuzz/CMakeLists.txt
+++ b/tests/fuzz/CMakeLists.txt
@@ -1,0 +1,31 @@
+#
+# This file is open source software, licensed to you under the terms
+# of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+# distributed with this work for additional information regarding copyright
+# ownership.  You may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+#
+# Copyright (C) 2026 Redpanda Data.
+#
+
+#
+# sstring fuzzer (only when using custom sstring, not std::string)
+#
+if (Seastar_SSTRING)
+  seastar_add_test (sstring
+    KIND FUZZ
+    SOURCES sstring_fuzz.cc
+    RUN_ARGS -max_len=100)
+endif ()
--- a/tests/fuzz/sstring_fuzz.cc
+++ b/tests/fuzz/sstring_fuzz.cc
@@ -1,0 +1,822 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2026 Redpanda Data.
+ */
+
+/**
+ * Differential Fuzzer for seastar::sstring
+ *
+ * This fuzzer uses std::string as an oracle to find correctness bugs in sstring.
+ * The fuzzer interprets the random input as a sequence of operations (opcodes),
+ * applying each operation to both an sstring and a std::string in parallel.
+ * After each operation, we verify that both strings remain equivalent.
+ *
+ * Architecture:
+ * - NUM_SLOTS string pairs (sstring + std::string) for multi-string operations
+ * - Each byte from fuzz input is interpreted as an opcode + parameters
+ * - FuzzReader consumes bytes and generates deterministic test data from seeds
+ * - Operations cover construction, mutation, search, comparison, and iteration
+ *
+ * The fuzzer is designed to exercise both the internal (small string optimization)
+ * and external (heap-allocated) storage paths in sstring by using a mix of small
+ * and large strings generated from the fuzz input.
+ */
+
+#include <seastar/core/sstring.hh>
+#include <seastar/util/assert.hh>
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
+#include <optional>
+#include <random>
+#include <string>
+#include <string_view>
+#include <type_traits>
+#include <typeindex>
+
+using seastar::sstring;
+
+// Helpers for C++20 compatibility (std::string::contains added in C++23)
+bool contains(const std::string& str, std::string_view sv) {
+    return str.find(sv) != std::string::npos;
+}
+
+bool contains(const std::string& str, char c) {
+    return str.find(c) != std::string::npos;
+}
+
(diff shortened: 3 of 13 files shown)


travisdowns:
@scylladb/seastar-maint please take a look

nyh:
Hi,
I wasn't familiar with this clang fuzzer, so I went to have a look and see its developer moved to a different fuzzer project, and then moved a second time - to https://github.com/google/fuzztest. Do you think it's worth investing in the older one? Are the clang developers keeping it alive?

doc/testing.md
```diff
@@ -49,3 +49,60 @@ seastar_add_test(woof_test
...
+## Fuzz Testing
+
+Seastar supports fuzz testing using [libFuzzer](https://llvm.org/docs/LibFuzzer.html) for testing non-reactor code. Fuzz tests are built only in `fuzz` mode.
+
+### Building Fuzz Tests
+
+Configure with fuzz mode (requires Clang):
+
+```bash
+./configure.py --mode fuzz --compiler clang++ [other options]
```
xemul, line 61: Is it correct that "fuzz" is build mode? AFAIU it only really applies to some tests, while e.g. apps/iotune won't benefit from being built in that mode

nyh: Yes, this also concerned me, but then I realized (or at least, thought I realized), that the fuzzer works by instrumenting code at compile time, and you need to compile specially all the involved code (in this case, the sstring implementation, maybe it has pieces in some source file), and perhaps it's not easy to find only a specific subset of files that need to be rebuilt. Although you're right that certainly top-level "apps" don't need to be built in this mode.

travisdowns:
> Hi,
> I wasn't familiar with this clang fuzzer, so I went to have a look and see its developer moved to a different fuzzer project, and then moved a second time - to https://github.com/google/fuzztest. Do you think it's worth investing in the older one? Are the clang developers keeping it alive?

libfuzzer seems pretty alive as it is built into llvm and probably the #1 fuzzing approach for clang today. So it has probably stopped moving but is widely used and well supported.

Ultimately my goal is to use FuzzTest also, but they are not mutually exclusive: FuzzTest is a rich API for writing fuzz tests, but it can use various "backends" to do the coverage guided mutations: libfuzzer is one of the backends, originally sort of the default, but now it seems work has moved on to "centipede", which is another engine.

FuzzTest especially changes the way you write your fuzz tests, it offers a lot of functionality for declarative specification of how the input randomness gets mapped into the objects your test cares about.

For now, my preference is to check in this relatively simple version with libfuzzer, and then build on that later if we add FuzzTest. One thing I'm uncertain about with FuzzTest is how to integrate it with seastar: seastar currently expects all of its dependencies to be apt-installable, but FuzzTest is not. So we can always cook it or something like that, but it would be a first I think.

doc/testing.md
travisdowns: Correct. Fuzzing requires the ~whole tree to be built with special flags like `-fsanitize=fuzzer` which effectively means it needs to be a new mode (even if some of the binaries are useless in that mode). Do you think it's worth disabling all "leaf" outputs that aren't interesting like the demo applications and non-fuzz binaries?

avikivity: Does it need anything in install-dependencies.sh or is everything bundled in llvm?

avikivity: Please also add a github actions thing for it.

avikivity: Or maybe it doesn't make sense to run continuously? I don't have experience with fuzzers.

tests/fuzz/sstring_fuzz.cc
```diff
@@ -0,0 +1,816 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2025 ScyllaDB
```
avikivity: Wrong attribution

tests/fuzz/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+#
+# Copyright (C) 2025 Scylladb, Ltd.
+#
```
avikivity, line 21: Wrong attribution

tests/fuzz/sstring_fuzz.cc
```diff
@@ -0,0 +1,816 @@
...
+ *
+ * The fuzzer is designed to exercise both the internal (small string optimization)
+ * and external (heap-allocated) storage paths in sstring by using a mix of small
+ * and large strings generated from the fuzz input.
+ */
+
+// Force the custom sstring implementation for fuzzing
+#ifndef SEASTAR_SSTRING
+#define SEASTAR_SSTRING
+#endif
```
avikivity: I think it will cause problems with modules later. Perhaps just skip the test.

avikivity:
Very interesting. If you introduce some subtle bug, will it find it?

Perhaps revert 242cb57fe7f89bf0458506968ef64de7db137d06 or 30185fd901275d354091312b876b891d65a09491.

travisdowns:
> If you introduce some subtle bug, will it find it?

Yes, I did try this. Almost any random mutation like skipping the last character in some operation it will find immediately, but also so would comprehensive unit tests. I tried some harder ones and it did OK: e.g., I changed operator== to skip the first char on the LHS if the string size is > 10:

```
    bool operator==(const basic_sstring& x) const noexcept {
        size_t offset = (size() > 10) ? 1 : 0;
        return size() == x.size() && std::equal(begin() + offset, end(), x.begin());
    }
```

This it finds in < 1s almost every time.

However, a much harder one is skip the _last_ character if the size is over the threshold:

```
    bool operator==(const basic_sstring& x) const noexcept {
        size_t offset = (size() > 10) ? 1 : 0;
        return size() == x.size() && std::equal(begin(), end() - offset, x.begin());
    }
```

This takes > 1 minute to find and earlier versions of the fuzzer couldn't find it in a reasonable amount of time, because it is hard to set up the situation where only the last character matters: the first 9 characters must be the same, but then the last character must be different. Coverage-based guidance tries to execute every basic block, but that's not sufficient here since the "error basic block" (offset = 1) happens quickly but usually doesn't cause an error: it's data dependent. The comparison itself happens in a loop so there is no dedicated basic block for "the 10th character differs", just for "some character differs". Still fuzzers are clever enough to partly work around this.

In part a fuzzer is there to find issues you didn't even think of including in a unit test, but it's also there to add quick coverage to classes which don't have comprehensive unit tests in the first place, in kind of a "stupid" way without needing to know a ton about the internals of the class.

travisdowns:
@avikivity the fuzzer also found this bug in libstdc++: https://godbolt.org/z/7Pdq1Pqz1

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=123991

avikivity:
> @avikivity the fuzzer also found this bug in libstdc++: https://godbolt.org/z/7Pdq1Pqz1
> 
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=123991

Impressive!

travisdowns:
> Very interesting. If you introduce some subtle bug, will it find it?
> 
> Perhaps revert [242cb57](https://github.com/scylladb/seastar/commit/242cb57fe7f89bf0458506968ef64de7db137d06) or [30185fd](https://github.com/scylladb/seastar/commit/30185fd901275d354091312b876b891d65a09491).

It finds 242cb57 in a fraction of a second (~300 iterations, whlie it does about 60K iters/sec).

travisdowns:
> Perhaps revert [242cb57](https://github.com/scylladb/seastar/commit/242cb57fe7f89bf0458506968ef64de7db137d06) or [30185fd](https://github.com/scylladb/seastar/commit/30185fd901275d354091312b876b891d65a09491).

It doesn't catch 30185fd as it was using `memcmp` with explicit size to check the result of `c_str` which bascially ignores the termination character :facepalm:. Changed it to `strcmp` and it detects it now.

doc/testing.md
travisdowns: > Does it need anything in install-dependencies.sh or is everything bundled in llvm?

No dependencies. In a later change I add support for async fuzzing (i.e. stuff that needs the reactor), and there there is the need to link against an explicit library, but it's still included in llvm.

travisdowns: > Or maybe it doesn't make sense to run continuously? I don't have experience with fuzzers.

There are different ways to do it. One goal is to run the fuzzers for a long time in aggregate, this can be done with a scheduled run. Another is to pass each individual changed through the fuzzer to catch new bugs. This also sort of serves the first requirement, though people get annoyed if the fuzzer finds a bug which is not related to their change, but given what's here so far that doesn't seem very likely.

I'll add a fuzz step to CI.

avikivity:
Respect!

tests/fuzz/sstring_fuzz.cc
travisdowns: Fixed. Note sesatar does not currently compile in std::string mode: need a CI for that if we want to ensure it works.

doc/testing.md
travisdowns: I've done the following:

 - Fuzz tests "run" in all build modes (if you using llvm), though in modes other than fuzzer nothing is instrumneted except the foo_fuzz.cc translation unit itself, so coverage will be poor (though for header-mostly stuff it can still be high, like sstring). They run for 1 second in non-fuzz modes.
 - In fuzz mode everyting is properly instrumented and fuzz tests run for 60 seconds by default
 
In CI, we add a single additional job to the matrix which builds in fuzz mode and runs only fuzz tests. We don't add fuzz mode to the general matrix. Fuzz tests will still run in all modes (1 second) per above.

tests/fuzz/sstring_fuzz.cc
travisdowns: fixed

tests/fuzz/CMakeLists.txt
travisdowns: fixed

travisdowns:
force https://github.com/scylladb/seastar/commit/dc84c33b87aac4d6ff0516795233ed3133ec6125 fixes the C++20 compile error by not using str::string::contains, and also squashes this and previous fixup commits into the main change.

tests/unit/CMakeLists.txt
```diff
@@ -109,33 +142,57 @@ function (seastar_add_test name)
...
-
-    if ((Seastar_STACK_GUARDS STREQUAL "ON") OR
-        ((Seastar_STACK_GUARDS STREQUAL "DEFAULT") AND
-         (CMAKE_BUILD_TYPE IN_LIST Seastar_STACK_GUARD_MODES)))
+    # FUZZ tests use libFuzzer which provides its own main
+    if (parsed_args_KIND STREQUAL "FUZZ")
+      target_compile_options (${executable_target}
+        PRIVATE -fsanitize=fuzzer)
+      target_link_options (${executable_target}
+        PRIVATE -fsanitize=fuzzer)
```
avikivity, line 176: I read that -fsanitize=address is recommended for fuzzing, to avoid memory corruption carrying a hidden bug from one execution to the next.

PR #3238, opened 2026-02-03
Add ability to configure IO bandwidth limit for supergroup

xemul:
Currently only scheduling groups have the update_io_bandwidth() method that translates into asking the io_group::priority_class_data to make real use of its shared token bucket. This PR adds the same facility for a supergroup. The result is -- all groups that populate the given supergroup are not allowed to exceed the supergroup's limit altogether. The internal "balancing" happens as it goes -- either according to shares, or as individual groups do or do not submit their IO.

Wiring the per-supergroup throttler means adding a hierarchy to io_group::priority_class_data objects that maintain the shared IO state for a scheduling group. For scheduling_supergroup there appears io_group::priority_class_group_data with bandwidth limiter, similar to the io_group::priority_class_data limiter (implemented as the util::shared_token_bucket).

The actual "throttling" happens inside the io_queue::priority_class_data methods. The class in question maintains the shard-local IO state for a scheduling group. When a request is dispatched, the queue-level class data udpates the group-level limited with the amount of consumed tokens. If the limiter reports that the limit was reached, the queue-level class data is "unplugged" from dispatch until limiter is replenished. To support supergroup's limiter as well, each queue-level class data maintains several references to limiters (with replenishment machinery), and when a request is dispatched each limiter is updated. If at least one limiter reports the deficiency of tokens, the queue-level class data is unplugged, when all limiters are replenished the queue-level class is plugged back.

Unit test included.

### DIFF
--- a/tests/unit/io_queue_test.cc
+++ b/tests/unit/io_queue_test.cc
@@ -36,6 +36,7 @@
 #include <seastar/core/disk_params.hh>
 #include <seastar/core/internal/io_request.hh>
 #include <seastar/core/internal/io_sink.hh>
+#include <seastar/core/loop.hh>
 #include <seastar/util/assert.hh>
 #include <seastar/util/internal/iovec_utils.hh>
 #include <seastar/util/defer.hh>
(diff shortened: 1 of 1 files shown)


xemul:
upd:
- fixed copilot spelling checks

mitso23:
@xemul  are you planning to ask somebody to review this :)

xemul:
Yes, weekly maintainer ping timer is armed

xemul:

xemul:
upd:
- rebased (simple conflict with #3210)

xemul:

xemul:

avikivity:
Sorry, reviewing.

src/core/io_queue.cc
```diff
@@ -1146,8 +1141,9 @@ void io_queue::update_shares_for_class_group(unsigned index, size_t new_shares)
 future<> io_queue::update_bandwidth_for_class(internal::priority_class pc, uint64_t new_bandwidth) {
     return futurize_invoke([this, pc, new_bandwidth] {
         if (_group->_allocated_on == this_shard_id()) {
-            auto& pclass = find_or_create_class(pc);
+            auto& pclass = _group->find_or_create_class(pc);
```
avikivity, line 1217: Side note: we should create the I/O data structures eagerly when the scheduling group is created.

src/core/io_queue.cc
```diff
@@ -868,15 +872,46 @@ io_queue::priority_class_data& io_queue::find_or_create_class(internal::priority
     return *_priority_classes[id];
 }
 
+io_group::priority_class_group_data& io_group::find_or_create_class_group(unsigned group_index) {
+    std::lock_guard _(_lock);
+    return find_or_create_class_group_locked(group_index);
```
avikivity, line 915: Please file an issue for removing these locks.

I think that creating everything eagerly will be sufficient, since the create functions return a future.

avikivity: And, we won't need the parallel structure we have for I/O.

tests/unit/io_queue_test.cc
```diff
@@ -754,3 +754,179 @@ SEASTAR_THREAD_TEST_CASE(test_gauge_integrator_test) {
     }
     fmt::print("done\n");
 }
```
avikivity, line 757: > Three tests here:

off-by-one.

I'd like to see a test that with a supergroup limit and subgroups unlimited, they subgroups still respect the shares.

avikivity:
Looks good. Sorry about the delayed review.

src/core/io_queue.cc
xemul: > I think that creating everything eagerly will be sufficient, since the create functions return a future.

It's a bit wasteful. There are always at least two io_queues in the system -- one for configured mountpoint from the io-properties.yaml and a fallback one. Also, there are classes in scylla that never do any IO like memory compaction one. Is this lock the only problem that bothers you? It's almost never taken, the io_queue::find_or_create_class() (the caller of io_group::find_or_create_class_group()) doesn't have this lock and most often it find the priority-class allocated and doesn't call io_group

> And, we won't need the parallel structure we have for I/O.

By parallel structure do you mean that tasks_queue-s for CPU and priority_class_data-s for io_queue grow in parallel (and there's also fair_queue::priority_class_data doing the same)?

xemul:
upd:
- rebased (no conflicts)
- added plug/unplug for fair queue groups
- added a test to validate how shares are respected for classes in a throttled group

src/core/io_queue.cc
avikivity: > > I think that creating everything eagerly will be sufficient, since the create functions return a future.
> 
> It's a bit wasteful. There are always at least two io_queues in the system -- one for configured mountpoint from the io-properties.yaml and a fallback one. Also, there are classes in scylla that never do any IO like memory compaction one. Is this lock the only problem that bothers you? It's almost never taken, the io_queue::find_or_create_class() (the caller of io_group::find_or_create_class_group()) doesn't have this lock and most often it find the priority-class allocated and doesn't call io_group

It's mostly about something that can fail during runtime. We can defer the discussion about it.

> 
> > And, we won't need the parallel structure we have for I/O.
> 
> By parallel structure do you mean that tasks_queue-s for CPU and priority_class_data-s for io_queue grow in parallel (and there's also fair_queue::priority_class_data doing the same)?

Yes, nest everything under task_queue (or a structure that contains both).

I now realize that io scheduling as a further level (per mountpoint), so it's not that simple.

PR #3239, opened 2026-02-03
rpc: log handler type

gusev-p:
Helps debugging and troubleshooting.

### DIFF
--- a/include/seastar/rpc/rpc_impl.hh
+++ b/include/seastar/rpc/rpc_impl.hh
@@ -606,7 +606,7 @@ auto send_helper(MsgType xt, signature<Ret (InArgs...)> xsig) {
 static constexpr size_t response_frame_headroom = 16;
 
 template<typename Serializer, typename RetTypes>
-inline future<> reply(wait_type, future<RetTypes>&& ret, int64_t msg_id, shared_ptr<server::connection> client,
+inline future<> reply(wait_type, future<RetTypes>&& ret, uint64_t verb, int64_t msg_id, shared_ptr<server::connection> client,
         std::optional<rpc_clock_type::time_point> timeout, std::optional<rpc_clock_type::duration> handler_duration) {
     if (!client->error()) {
         snd_buf data;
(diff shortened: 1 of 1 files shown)


gusev-p:

include/seastar/rpc/rpc_impl.hh
```diff
@@ -777,7 +777,7 @@ auto protocol<Serializer, MsgType>::register_handler(MsgType t, scheduling_group
     using clean_sig_type = typename sig_type::clean;
     using want_client_info = typename sig_type::want_client_info;
     using want_time_point = typename sig_type::want_time_point;
-    auto recv = recv_helper<Serializer>(clean_sig_type(), std::forward<Func>(func),
+    auto recv = recv_helper<Serializer>(static_cast<uint64_t>(t), clean_sig_type(), std::forward<Func>(func),
```
travisdowns, line 780: How does this cast work? Is it guaranteed to work for any MsgType, e.g., because of a concept?

gusev-p: We already [have such casts on the sender side](https://github.com/scylladb/seastar/blob/4cca6905e48464e03e99fa3265747c861869c910/include/seastar/rpc/rpc_impl.hh#L577). I [don't see](https://github.com/scylladb/seastar/blob/4cca6905e48464e03e99fa3265747c861869c910/include/seastar/rpc/rpc.hh#L884) any explicit concepts on the protocol class declaration. I don't see much value in using `MsgType` anywhere except the public api of the protocol class (make_client/register_handler). Probably we should add a concept that `MsgType` is convertible to `uint64_t` and cast it away when it crosses protocol class public api. This would make the requirements more explicit and we wouldn't need this wording in the comments:

> /// `MsgType` should be hashable and serializable. It is preferable to use enum
> /// for message types, but do not forget to provide hash function for it.

@gleb-cloudius (you're the author of this code, right?) WDYT?

gleb-cloudius: > We already [have such casts on the sender side](https://github.com/scylladb/seastar/blob/4cca6905e48464e03e99fa3265747c861869c910/include/seastar/rpc/rpc_impl.hh#L577). I [don't see](https://github.com/scylladb/seastar/blob/4cca6905e48464e03e99fa3265747c861869c910/include/seastar/rpc/rpc.hh#L884) any explicit concepts on the protocol class declaration. I don't see much value in using `MsgType` anywhere except the public api of the protocol class (make_client/register_handler). Probably we should add a concept that `MsgType` is convertible to `uint64_t` and cast it away when it crosses protocol class public api. This would make the requirements more explicit and we wouldn't need this wording in the comments:
> 
> > /// `MsgType` should be hashable and serializable. It is preferable to use enum
> > /// for message types, but do not forget to provide hash function for it.
> 
> @gleb-cloudius (you're the author of this code, right?) WDYT?

Make sense to me.

avikivity: But why not pass it as MsgType?

gusev-p: > But why not pass it as MsgType?

It does not add much value here. We still need to cast to `uint64_t` before passing it to `fmt::format`. It's simpler to just pass the uint64 -- no need to specify additional template parameter in all the called functions.

avikivity: We could require format it as MsgType.

I don't mind, but we can improve here.

PR #3240, opened 2026-02-04
Deprecate reactor::make_pollable_fd(socket_address, int)

xemul:
The method in question is special-purpose helper for posix stack, that uses it to create file_desc for socket. It's hardly a good fit for the public API, pollable_fd-s are seastar internal entities. The posix stack can create file_desc on its own, without reactor help. However, it needs to guess whether or not to use SOCK_NONBLOCK flag for it, which depends on the backend in use, so some interaction with reactor is still needed.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -526,6 +526,7 @@ public:
     bool posix_reuseport_available() const { return false; }
     bool posix_sock_need_nonblock() const;
 
+    [[deprecated("Internal reactor function, consider using net sockets")]]
     pollable_fd make_pollable_fd(socket_address sa, int proto);
 
     [[deprecated("Internal reactor function, consider using seastar::connect)")]]


dotnwat:
@xemul i was contemplating recently how I might integrate a particular async library into seastar: the library user is expected to wait for a fd to become readable/writable, and then drive operation within the library by calling a non-blocking flush/consume. i had only just started exploring, but had made a mental note that pollable_fd might be the way to do this. since this isn't real yet, deprecating this won't be a problem, but I am curious if I should be looking towards some other solution here?

xemul:
Yes, pollable_fd can be a good fit for it, but without more details it's hard to tell whether it will indeed. For example, we have TLS and DNS libraries integrated with seastar, and both manage to work without messing with pollable_fd, even though sockets in seastar are built around pollable_fd-s.

Similar story with eventfd-s, there are wrapper classes around pollable_fd-s. But it's not as good example, as no external library is involved, and the usage is also internal to reactor.

xemul:
This PR tries to follow a concept that one wraps a file descriptor with the `file_desc` object, which is not internal and more posix-ish than pollable_fd, and then, if reactor assistance with polling is needed, wraps file_desc transparently into pollable_fd, which is the next level of abstractions between kernel/posix and reactor.

dotnwat:
Thanks, makes total sense.

> TLS and DNS libraries integrated with seastar, and both manage to work without messing with pollable_fd, even though sockets in seastar are built around pollable_fd-s.

I guess this works because the TLS libraries allow the data from the socket to be obtained out-of-band (e.g. using futureized reads from an input_stream).

xemul:
Well, sort of. TLS allows configuring own read/write callbacks and seastar uses it, spawning background iostream operations, kicking the library when needed

xemul:

travisdowns:
@dotnwat do you have another solution to your integration if pollable_fd goes away? I.e., does:

> if reactor assistance with polling is needed, wraps file_desc transparently into pollable_fd, which is the next level of abstractions between kernel/posix and reactor.

provide the same support for your use case as the code being deprecated?

xemul:
@travisdowns , pollable_fd doesn't go away, it stays (in this PR)

avikivity:
Is make_pollable_fd in fact used outside Seastar? Maybe just remove it.

Generally libraries that are designed to be driven by a separate event loop supply read/write callbacks, not just poll, so they don't need to be provided with real fds.

xemul:
> Is make_pollable_fd in fact used outside Seastar? Maybe just remove it.

It's public method, so we can't be sure. Scylla doesn't use it. Keeping it in "deprecated" state for a while shouldn't be hard

PR #3242, opened 2026-02-04
test: coroutine_test: adjust illegal tests

avikivity:
test_abandond_coroutine() contains two violations:

1. it captures (by reference) p2 and p3, but ensure the coroutine runs only after they go out of scope (co_await std::move(*f)). If the scheduler decides to delay execution, the coroutine would access destroyed promises. This probably doesn't happen, but is still illegal.
2. The lambda (inner) coroutine's execution is delayed, but the compiler only has to extend its captures' lifetime until the end of the statement. However, it is executed after the statement completes. This does trigger when we apply [[coro_await_elidable]] to future<>.

We just remove the test. It's not clear what it's testing. Destroying the promise to evoke a broken_promise before the coroutine executes would just store a broken_promise in the future, and isn't interesting.

In test_scheduling_group() we have a similar lambda coroutine that is executed after the statement where it was defined completes. Here a simple workaround is to avoid capturing anything. If a capture group that doesn't hold any captures is destroyed too early, no one notices.

### DIFF
--- a/tests/unit/coroutines_test.cc
+++ b/tests/unit/coroutines_test.cc
@@ -125,23 +125,6 @@ SEASTAR_TEST_CASE(test_simple_coroutines) {
     }), 17);
 }
 
-SEASTAR_TEST_CASE(test_abandond_coroutine) {
-    std::optional<future<int>> f;
-    {
-        auto p1 = promise<>();
-        auto p2 = promise<>();
-        auto p3 = promise<>();
-        f = p1.get_future().then([&] () -> future<int> {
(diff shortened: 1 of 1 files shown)


xemul:
> it captures (by reference) p2 and p3, but ensure the coroutine
>   runs only after they go out of scope (co_await std::move(*f))

But before `co_await std::move(*f)` there's ` co_await p2.get_future();` whose promise is set_value-ed inside the coroutine, so it doesn't go out of scope that early. Or you mean some other coroutine in the above description?

Issue #2713, opened 2025-04-01
seastar::make_file_output_stream by-pass fdatasync?


nvartolomei:
Hi, we have some cases where we use `seastar::output_stream` created by calling `seastar::make_file_output_stream` for writing temporary files (quite a few of them) and I think we could benefit from avoiding calling fdatasync on them. Currently, to make sure the (output stream buffered) data made it to the file we have to call `seastar::output_stream::flush()` (and `close()` too) but that flush also calls unconditionally `seastar::file::flush()` which results in a fdatasync call.

Any thoughts on adding an option for avoiding that? I.e. as an option inside `ss::file_output_stream_options`.

dotnwat:
do i understand correctly that you want to be able to (1) flush buffered data from the `output_stream` to the underlying file so that it is visible through the file interface, and (2) do that without incurring a call to `fdatasync`, but can't avoid it because `output_stream::close|flush` both call `flush` on the underlying file?

that's what the behavior looks like to me and the alternative makes sense. from what i've observed, the writes incurred by flushing the `output_stream` buffers to the underlying file (which are always necessary) are expensive in comparison to the fsync which is usually very quite speedy on nvme.

nvartolomei:
Yes, you described it perfectly.

> from what i've observed, the writes incurred by flushing the `output_stream` buffers to the underlying file (which are always necessary) are expensive in comparison to the fsync which is usually very quite speedy on nvme.

I've made the opposite observation when writing small temporary files due to disk volatile caches being involved. Especially on consumer grade hardware.

xemul:
Fixed by b0ab5199bbdc8dee1f72c7b78028047a8b302432. Setting file_open_options::durable = false avoids fdatasync-s

PR #3243, opened 2026-02-05
Move reactor::fdatasync() into posix_file_impl

xemul:
The method in question is alien to reactor and is performs file IO. Also, it works on a file descriptor, but seastar file API is about class file, not descriptors.

While moving the PR solves a feature request to open files with the ability to skip flushes (#2713), and facilitates the AIO-less implementation of fdatasync for pure-epoll backend (#3208)

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -703,6 +703,7 @@ private:
     future<> send_all_part(pollable_fd_state& fd, const void* buffer, size_t size, size_t completed);
 #endif
 
+    [[deprecated("Use file::flush() instead")]]
     future<> fdatasync(int fd) noexcept;
 
     void add_timer(timer<steady_clock_type>*) noexcept;


src/core/file.cc
```diff
@@ -198,12 +202,54 @@ posix_file_impl::posix_file_impl(int fd, open_flags f, std::atomic<unsigned>* re
...
-    if ((_open_flags & open_flags::dsync) != open_flags{}) {
+    if (_skip_flush || ((_open_flags & open_flags::dsync) != open_flags{})) {
         // If the file is opened with open_flags::dsync, all writes are already
         // fdatasync-ed to the disk, so we don't need to fdatasync again.
         return make_ready_future<>();
     }
-    return engine().fdatasync(_fd);
+
+    reactor::io_stats::local().fsyncs++;
+    return fdatasync(_aio_fdatasync, _fd, _io_queue.sink());
```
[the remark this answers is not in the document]
xemul: Yes, that's intentional

xemul:

xemul:
upd:
- rebased (conflicted with #3249 and #3240)

xemul:

xemul:

xemul:

include/seastar/core/file.hh
```diff
@@ -98,6 +98,7 @@ struct file_open_options {
     uint64_t sloppy_size_hint = 1 << 20; ///< Hint as to what the eventual file size will be
     file_permissions create_permissions = file_permissions::default_file_permissions; ///< File permissions to use when creating a file
     bool append_is_unlikely = false; ///< Hint that user promises (or at least tries hard) not to write behind file size
+    bool skip_flush = false; ///< If true, calls to file::flush() are no-ops and no durability flush is issued to the underlying storage.
```
avikivity: Would be better to call it durable = true. skip_flush = false requires some logic to deduce we actually are doing something.

src/core/file.cc
```diff
@@ -203,7 +205,7 @@ posix_file_impl::posix_file_impl(int fd, open_flags f, std::atomic<unsigned>* re
 
 future<>
 posix_file_impl::flush() noexcept {
-    if ((_open_flags & open_flags::dsync) != open_flags{}) {
+    if (_skip_flush || ((_open_flags & open_flags::dsync) != open_flags{})) {
```
avikivity: If the file is not durable, we should remove dsync from the open flags earlier.

avikivity: But that doesn't affect this calculation.

src/core/reactor.cc
```diff
@@ -1813,6 +1813,7 @@ reactor::open_file_dma(std::string_view nameref, open_flags flags, file_open_opt
         open_flags |= O_CLOEXEC;
         if (bypass_fsync) {
             open_flags &= ~O_DSYNC;
+            options.skip_flush = true;
```
avikivity: Should be chained: if bypass fsync, it's not durable. If not durable (perhaps because the user asked so), disable O_DSYNC.

xemul:
upd:
- rebased
- renamed file_open_oprions::skip_flush into durable and updated the open() and flush() paths accordingly

PR #3244, opened 2026-02-05
file: coroutinize the with_file templates

bhalevy:
with_file with_file_close_on_failure already allocate using do_with and are also likely block when awaiting the file_fut so it makes sense to simply and optimize them by using a coroutine.

### DIFF
--- a/include/seastar/core/file.hh
+++ b/include/seastar/core/file.hh
@@ -23,6 +23,7 @@
 
 #include <seastar/util/std-compat.hh>
 #include <seastar/core/coroutine.hh>
+#include <seastar/coroutine/as_future.hh>
 #include <seastar/coroutine/generator.hh>
 #include <seastar/core/do_with.hh>
 #include <seastar/core/stream.hh>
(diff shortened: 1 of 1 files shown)


bhalevy:
Cc @tarasveretilnyk

bhalevy:
rebased to re-trigger ci

include/seastar/core/file.hh
```diff
@@ -592,16 +592,38 @@ private:
...
-        });
-    });
+template <std::invocable<file&> Func, typename T = futurize<std::invoke_result_t<Func, file&>>::value_type>
+requires (!std::is_same_v<T, internal::monostate>)
+future<T> with_file(future<file> file_fut, Func func) noexcept {
+    auto f = co_await std::move(file_fut);
+    std::optional<T> res;
+    std::exception_ptr ex;
+    try {
+        res.emplace(co_await futurize_invoke(func, f));
```
xemul: Maybe `coroutine::as_future` will help to relax the try-catch logic?

include/seastar/core/file.hh
```diff
@@ -592,16 +592,38 @@ private:
...
+    co_await f.close();
+    if (ex) {
+        co_await coroutine::return_exception_ptr(std::move(ex));
+    }
+    co_return std::move(*res);
+}
+
+template <std::invocable<file&> Func, typename T = futurize<std::invoke_result_t<Func, file&>>::value_type>
+requires (std::is_same_v<T, internal::monostate>)
+future<> with_file(future<file> file_fut, Func func) noexcept {
```
xemul: Unless I mis-parsed it, the only difference from the previous overload is the return type. But AFAIK C++ doesn't allow overloads to differ only using the return type. So where's the catch?

xemul: Ah, it's for void-returning functions 🤦

bhalevy:
[55a1418](https://github.com/scylladb/seastar/commit/55a14188feec299929394a5f68ce0e8fb3290619):
- got rid of try/catch block
- put f.close() in `with_file` into a `finally` clause so to handle both `func` and `f.close()` failing (although close is never supposed to fail)

bhalevy:
[682677a](https://github.com/scylladb/seastar/commit/682677a64e0c93fc7bccf323e4e2b2703ec49a5b):
- use coroutine::as_future to simplify with_file_close_on_failure
- both with_file and with_file_close_on_failure now do not need variants for void/future<>-returning `func` parameters

PR #3245, opened 2026-02-05
fix trailing whitespace

travisdowns:
Hopefully for the last time. It snuck in due to a race with the change that introduced the check in CI.


travisdowns:
@scylladb/seastar-maint please consider for a quick merge.

PR #3246, opened 2026-02-06
reactor: handle transient EPERM / EACCES in touch_directory

emdroid:
**Problem Statement:** We have observed intermittent EPERM failures inside `touch_directory()` concurrent calls on specific CI runners. Despite extensive local testing on XFS, Docker, and OverlayFS, the error is highly environmental and difficult to reproduce on demand.

**Technical Justification**: While mkdir should ideally return EEXIST when a directory already exists, the Linux VFS layer and Linux Security Modules (LSMs) like AppArmor can return EPERM or EACCES under heavy dentry contention. This could happens when:
1. **OverlayFS Copy-Up**: A "copy-up" operation is in progress from a lower to an upper layer, and a second thread attempts to access the dentry before the transition is finalized.
2. **LSM Interference**: The security module intercepts a concurrent syscall during a metadata lock and rejects it as a conflict rather than allowing the filesystem driver to return a standard EEXIST.

**The "Idempotent Verification" Strategy**: Rather than blindly ignoring EPERM, this PR implements a "Verify-After-Error" pattern:
* Step 1: Attempt mkdir.
* Step 2: If EPERM or EACCES is returned, perform a stat().
* Step 3: If stat() confirms the path exists and is a directory, we consider the goal achieved.
* Step 4: If the directory is missing, we re-throw the original error.

**Why this is safe even without a local repro**: This change does not mask genuine failures. It only converts a "Permission Denied" into a "Success" if we can prove (via stat) that the operation's intended outcome was actually reached.

Fixes: scylladb/scylladb#28259

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2375,9 +2375,33 @@ reactor::touch_directory(std::string_view name_view, file_permissions permission
         auto mode = static_cast<mode_t>(permissions);
         return wrap_syscall<int>(::mkdir(name.c_str(), mode));
     });
-    if (sr.result == -1 && sr.error != EEXIST) {
-        sr.throw_fs_exception("mkdir failed", fs::path(name));
+
+    if (sr.result != -1) {
+        // Directory created successfully
+        co_return;
     }
+
+    if (sr.error == EEXIST) {
+        // Directory already exists, that's fine
+        co_return;
+    }
+
+    if (sr.error == EPERM || sr.error == EACCES) {
+        // Check if the directory actually exists and has the right permissions
+        try {
+            auto sd = co_await file_stat(name, follow_symlink::no);
+            if (sd.type == directory_entry_type::directory &&
+                    (sd.mode & static_cast<mode_t>(file_permissions::all_permissions)) == static_cast<mode_t>(permissions)) {
+                co_return;
+            }
+            // Directory exists but has wrong type or permissions - fall through to throw EPERM
+        } catch (...) {
+            // file_stat failed (e.g., ENOENT, EACCES), fall through to throw original EPERM error
+        }
+    }
+
+    // Some other error occurred, report it
+    sr.throw_fs_exception("mkdir failed", fs::path(name));
 }
 
 future<>


bhalevy (approved):
Looks good to me.
Why is the PR in draft mode?

bhalevy:
This is supposed to fix a ci instability issue (and a bunch of duplicates), let's move it forward.

src/core/reactor.cc
```diff
@@ -2359,9 +2359,33 @@ reactor::touch_directory(std::string_view name_view, file_permissions permission
...
+
+    if (sr.error == EEXIST) {
+        // Directory already exists, that's fine
+        co_return;
+    }
+
+    if (sr.error == EPERM || sr.error == EACCES) {
+        // Check if the directory actually exists and has the right permissions
+        try {
+            auto sd = co_await file_stat(name, follow_symlink::yes);
```
xemul: Shouldn't it be `follow_symlink::no`? You want to check that the `name` resolves into a directory, not to a symlink, that points to some other directory

emdroid:
> Looks good to me. Why is the PR in draft mode?

> This is supposed to fix a ci instability issue (and a bunch of duplicates), let's move it forward.

This change will definitely not hurt (it should be idempotent), but not yet completely sure it will really fix the issue, as we didn't reproduce this consistently yet (also it seems to only have happened on the spider17 machine, but can't reliably reproduce even there) - so we don't know that the mechanism of the issue is really what is being done in this PR.

Besides, we have seen also the EPERM happening unexpectedly for file open (i.e. not only mkdir), also on the spider17.

Therefore still investigating this issue.

avikivity:
Was this reported to linux-kernel or fuse?

emdroid:
> Was this reported to linux-kernel or fuse?

I could find this one opened in 2021 and still in NEW state: https://bugzilla.kernel.org/show_bug.cgi?id=211171

But seems to be only remotely similar (is on Windows host in VirtualBox).

emdroid:
As mentioned in the ticket, we couldn't actually reproducing it any more even after trying multiple things and high number of repeats, so it seems that might have been a temporary issue on spider17.

So it seems like we don't need this right now, but we could still decide to merge this to "harden" the code - it shouldn't make things worse, as the check is idempotent (doesn't skip real EPERM errors where the directory truly can't be created and doesn't already exist).

emdroid:
Rebased, change to `follow_symlinks::no`.

avikivity:
> Problem Statement: We have observed intermittent EPERM failures inside touch_directory() concurrent calls on specific CI runners. Despite extensive local testing on XFS, Docker, and OverlayFS, the error is highly environmental and difficult to reproduce on demand.

Really hard to make a decision with such a vague report. How can I tell we're not papering over a problem in some other place?

PR #3249, opened 2026-02-09
Refine posix file .dup() implementation

xemul:
There are two issues with posix_file_impl::dup()

First, despite being documented that file-handle can only be obtainer for read-only files, posix files generate handles for any file.

Second, when a file-handle is converted back to a file, the result is always posix_file_real_impl, even though the handle could be obtained from e.g. blockdev_file_impl. 

Unfortunately, there's no easy way to test how blockdev file handles work, because block devices are not open()able for non-root. So this PR only comes with some basic tests for regular files dup()-ing

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -165,6 +165,7 @@ void posix_file_impl::configure_io_lengths() noexcept {
     _write_max_length = std::min<size_t>(_write_max_length, limits.max_write);
 }
 
+template <typename FileImpl>
 std::unique_ptr<seastar::file_handle_impl>
 posix_file_impl::do_dup() {
     if ((_open_flags & open_flags::ro) != open_flags{}) {
@@ -174,7 +175,7 @@ posix_file_impl::do_dup() {
     if (!_refcount) {
         _refcount = new std::atomic<unsigned>(1u);
     }
-    auto ret = std::make_unique<posix_file_handle_impl>(_fd, _open_flags, _refcount, _device_id,
+    auto ret = std::make_unique<posix_file_handle_impl<FileImpl>>(_fd, _open_flags, _refcount, _device_id,
             _memory_dma_alignment, _disk_read_dma_alignment, _disk_write_dma_alignment, _disk_overwrite_dma_alignment,
             _nowait_works);
     _refcount->fetch_add(1, std::memory_order_relaxed);
@@ -550,7 +551,7 @@ posix_file_real_impl::read_dma(uint64_t pos, std::vector<iovec> iov, io_intent* 
 
 std::unique_ptr<seastar::file_handle_impl>
 posix_file_real_impl::dup() {
-    return posix_file_impl::do_dup();
+    return posix_file_impl::do_dup<posix_file_real_impl>();
 }
 
 future<temporary_buffer<uint8_t>>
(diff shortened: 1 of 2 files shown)


avikivity:
> 
> Unfortunately, there's no easy way to test how blockdev file handles work, because block devices are not open()able for non-root. So this PR only comes with some basic tests for regular files dup()-ing

unshare() could work, though it would be a pain to integrate.

src/core/file.cc
```diff
@@ -166,15 +166,15 @@ void posix_file_impl::configure_io_lengths() noexcept {
...
+posix_file_impl::do_dup(const std::type_info& file_impl) {
     if ((_open_flags & open_flags::ro) != open_flags{}) {
         throw std::runtime_error("File is not read-only");
     }
 
     if (!_refcount) {
         _refcount = new std::atomic<unsigned>(1u);
     }
-    auto ret = std::make_unique<posix_file_handle_impl>(_fd, _open_flags, _refcount, _device_id,
+    auto ret = std::make_unique<posix_file_handle_impl>(file_impl, _fd, _open_flags, _refcount, _device_id,
```
avikivity: I think it's more accurate to pass it a "maker function" that returns a different handle implementation for each device. Will know more in the next patch.

src/core/file.cc
```diff
@@ -1034,8 +1034,18 @@ posix_file_handle_impl::clone() const {
...
+    auto ret = [this] () -> shared_ptr<file_impl> {
+        if (_impl_type == typeid(posix_file_real_impl)) {
+            return ::seastar::make_shared<posix_file_real_impl>(_fd, _open_flags, _refcount, _device_id,
+                    _memory_dma_alignment, _disk_read_dma_alignment, _disk_write_dma_alignment, _disk_overwrite_dma_alignment, _nowait_works);
+        }
+        if (_impl_type == typeid(blockdev_file_impl)) {
+            return ::seastar::make_shared<blockdev_file_impl>(_fd, _open_flags, _refcount, _device_id,
+                    _memory_dma_alignment, _disk_read_dma_alignment, _disk_write_dma_alignment, _disk_overwrite_dma_alignment, _nowait_works);
+        }
+        throw std::runtime_error("Unknown file impl type stored on handle");
```
avikivity: Either pass the make_shared\<...> as a parameter to the constructor, or just have different file_handle_impl overloads. It will be cleaner.

If you write `+[] (... the parameters ...) { return ::seastar::make_shared<... the type you want ...>(... the arguments ...)` you get a function pointer, with the same signature for the two options. You can store this in the handle instead of the type_index.

xemul: How about making posix_file_handle a template? posix_file_handle_impl<T> takes T as the file_impl class it needs to make_shares inside its to_file() method.

xemul:
upd:
- templatize posix_file_handle_impl with the file_impl specific class name to create in to_file() method. This makes two different types for handles of posix file real impl and blockdev file impl -s

xemul:
> > Unfortunately, there's no easy way to test how blockdev file handles work, because block devices are not open()able for non-root. So this PR only comes with some basic tests for regular files dup()-ing
> 
> unshare() could work, though it would be a pain to integrate.

Or device cgroup. Looks simpler, but stlil painful

PR #3254, opened 2026-02-11
reactor: Make do_dump_task_queue a task_queue method

xemul:
This helper function dump the contents of a task_queue when queue stall event occurs. For that it's declared as reactor friend, but it's an overkill -- making it (private) method of task_queue itself solves the class members access and makes this code itself a bit shorter to write.


xemul:
upd:
- use explicit lambda to call new method, std::bind_front wasn't accepted by clang++ & C23

xemul:

xemul:
upd: rebased (had conflicts)

xemul:

Issue #12, opened 2015-01-15
seastar tcpserver - goclient txtx test on DPDK  is underperfoming x2 


slivne:
http://jenkins.cloudius-systems.com:8080/job/seastar-private-tcpserver-perf/8/PerfPublisher/

increasing the size of user_queue_space x2 does not help (user_queue_space = {212992 * 2}) - the associated results are from a build using asias/txtx branch in seastar-dev that has this change.

On Thor-Loki:

Goserver | tcp-server stack-native-dpdk171 | tcpserver stack-native-dpdk18 | tcpserver stack-posix
------------ | ------------- | ------------ | -------------
8.915581406 | 17.087416845  | 17.867953145  | 8.914382154 

values are seconds to complete the tests - lower is better

To reproduce you can tun the tests in the following manner:
  * install + configure dpdk
  * start tcpserver
  * clone tests.git
  * cd tests
  * scripts/tester.py  run  --config_param sut.ip:<ip> --config_param tester.ip:x apps/tcp/tests/test_1_goclient/tester
  * check the output at apps/tcp/tests/test_1_goclient/tester/out

or 
  * install + configure dpdk
  * start tcpserver 
  * clone tests.git
  * cd tests/apps/tcp/tests/test_1_goclient/tester
  * go run client.go -h <ip>

Please note you can use client.go -conn 1 option to limit the number of connections

slivne:
@asias the latest results from the nightly perf tests indicate that we have closed the gap 
can you have a look

http://jenkins.cloudius-systems.com:8080/job/seastar-tcpserver-perf/38/

if all is ok can we close this issue ?

asias:
@slivne let's close this issue. TSO support (193416054902915a1a99916e2243682d0c4326ae) in dpdk is supposed to close the x2 gap.

PR #3255, opened 2026-02-11
task: add symbol() method for improved task identification

gusev-p:
Currently, we identify tasks in logs and error messages using RTTI (typeid). However, this is insufficient for generic task implementations. Most notably, all C++ coroutines share the same underlying `coroutine_traits_base::promise_type` class. As a result, logs cannot distinguish between different coroutines (e.g., `sstable_write` vs `gossip_recv`) because they all report the same type name.

Introduce a virtual `task::symbol()` method that returns a `task_symbol` (a variant of `const std::type_info*` and `void*`). This allows tasks to expose a more precise identifier:
 * `const std::type_info*`: Used for strongly-typed functor tasks where RTTI provides enough information.
 * `void*`: Used for raw addresses, such as function pointers or coroutine resume addresses.

The default implementation returns `typeid(*this)`. For coroutines, we override this to return the resume function address. We rely on the Itanium C++ ABI, which guarantees that the first pointer in the coroutine frame is the resume function pointer. This address can later be resolved to a human-readable symbol (e.g., via `addr2line`).

Note that this symbol refers to the whole coroutine function. If the coroutine contains multiple `co_await` statements, we still won't be able to distinguish between them. One way to solve this would be to declare an `await_transform` function in
`coroutine_traits_base::promise_type` and capture the optional `std::source_location` parameter there. However, this would introduce significant overhead:
 * Each coroutine frame would need to store an additional line number field.
 * Each `co_await` would have to assign it.
 * Each `co_await` point would add an item to the binary's string table (for the filename), causing significant bloat.

The current solution introduces minimal overhead and should be sufficient for most purposes.

### DIFF
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -85,6 +85,13 @@ public:
 #endif
 };
 
+template <typename Promise>
+void* get_resume_address(std::coroutine_handle<Promise> coro) {
+    void* frame_ptr = coro.address();
+    // ABI: The first pointer in the frame is the resume function address
+    return *reinterpret_cast<void**>(frame_ptr);
+}
+
 template <typename T = void>
 class coroutine_traits_base {
 public:
@@ -134,6 +141,12 @@ public:
         scheduling_group set_scheduling_group(scheduling_group sg) noexcept {
             return std::exchange(this->_sg, sg);
         }
+
+        void* symbol() const override {
+            auto& self = const_cast<promise_type&>(*this);
+            auto handle = std::coroutine_handle<promise_type>::from_promise(self);
+            return get_resume_address(handle);
+        }
     };
 };
 
@@ -175,6 +188,12 @@ public:
 
         scheduling_group set_scheduling_group(scheduling_group new_sg) noexcept {
             return task::set_scheduling_group(new_sg);
+        }
+
+        void* symbol() const override {
+            auto& self = const_cast<promise_type&>(*this);
+            auto handle = std::coroutine_handle<promise_type>::from_promise(self);
+            return get_resume_address(handle);
         }
     };
 };
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -720,7 +720,7 @@ protected:
     Promise _pr;
 };
 
-template <typename Promise, typename Func, typename Wrapper, typename T = void>
+template <typename Promise, typename Func, typename Wrapper, void* Symbol, typename T = void>
 struct continuation final : continuation_base_with_promise<Promise, T> {
     // Func is the original function passed to then/then_wrapped. The
     // Wrapper is a helper function that implements the specific logic
@@ -743,6 +743,9 @@ struct continuation final : continuation_base_with_promise<Promise, T> {
             this->_pr.set_to_current_exception();
         }
         delete this;
+    }
+    virtual void* symbol() const override {
+        return Symbol;
     }
     Func _func;
     [[no_unique_address]] Wrapper _wrapper;
@@ -1461,7 +1464,7 @@ private:
         return fut;
     }
 
-    template <typename Func, typename Result = futurize_t<internal::future_result_t<Func, T>>>
+    template <void* Symbol, typename Func, typename Result = futurize_t<internal::future_result_t<Func, T>>>
     Result
     then_impl(Func&& func) noexcept {
         static_assert(internal::expect_only_rvalue_refs<Func>);
--- a/include/seastar/core/loop.hh
+++ b/include/seastar/core/loop.hh
@@ -534,6 +534,7 @@ class parallel_for_each_state final : private continuation_base<> {
     std::vector<future<>> _incomplete;
     promise<> _result;
     std::exception_ptr _ex;
+    void* _symbol;
 private:
     // Wait for one of the futures in _incomplete to complete, and then
     // decide what to do: wait for another one, or deliver _result if all
@@ -541,8 +542,9 @@ private:
     void wait_for_one() noexcept;
     virtual void run_and_dispose() noexcept override;
     task* waiting_task() noexcept override { return _result.waiting_task(); }
+    void* symbol() const override { return _symbol; }
 public:
-    parallel_for_each_state(size_t n);
+    parallel_for_each_state(size_t n, void* symbol);
     void add_future(future<>&& f);
     future<> get_future();
 };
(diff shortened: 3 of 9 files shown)


nyh:
I'm lacking some context to understand the impact of this change. What kind of error messages try to print tasks? Can you give an example of an error message that today prints something ugly like "coroutine_traits_base::promise_type" and how it will improve by using this new function? Does this improvement somehow affects also backtrace printouts - probably not, right?

gusev-p:
> I'm lacking some context to understand the impact of this change. What kind of error messages try to print tasks? Can you give an example of an error message that today prints something ugly like "coroutine_traits_base::promise_type" and how it will improve by using this new function? 

The primary motivation for this work was to improve `reactor.cc/do_dump_task_queue` -- this function is called when there are too many active tasks to execute in the reactor queue.

Before the patch:
`Jan 30 05:30:22 PLM2CA10 scylla[3360986]: [shard  0: gms] seastar - Too long queue accumulated for gossip (1076 tasks)#012 1: N7seastar12continuationINS_8internal22promise_base_with_typeINS_16temporary_bufferIcEEEEZNS_7reactor12do_recv_someERNS_17pollable_fd_stateEPNS1_16buffer_allocatorEE3$_0ZNS_6futureIvE14then_impl_nrvoISB_NSC_IS4_EEEET0_OT_EUlOS5_RSB_ONS_12future_stateINS1_9monostateEEEE_vEE#012 1075: N7seastar8internal21coroutine_traits_baseIvE12promise_typeE`
which can be decoded to
```
seastar::continuation<
    seastar::internal::promise_base_with_type<seastar::temporary_buffer<char>>, 
    seastar::reactor::do_recv_some(seastar::pollable_fd_state&, seastar::buffer_allocator*)::$_0, 
    // ... (internal future chaining logic) ...
>

seastar::internal::coroutine_traits_base<void>::promise_type
```

After the patch:
```
gusev-p@potwor2:~/src/scylladb/seastar/scripts$ ./seastar-addr2line -e ~/src/scylladb/build/dev/scylla
WARN  2026-02-11 11:22:28,335 [shard 0:main] seastar - Too long queue accumulated for main (19 tasks)
 1: 0x18c0560
 1: 0x19845a0
 4: seastar::shared_future<std::optional<compaction::compaction_stats> >::shared_state
 3: seastar::continuation<seastar::internal::promise_base_with_type<void>, seastar::noncopyable_function<seastar::future<void> (seastar::future<unsigned long>&&)>, seastar::future<unsigned long>::then_wrapped_nrvo<seastar::future<void>, seastar::noncopyable_function<seastar::future<void> (seastar::future<unsigned long>&&)> >(seastar::noncopyable_function<seastar::future<void> (seastar::future<unsigned long>&&)>&&)::{lambda(seastar::internal::promise_base_with_type<void>&&, seastar::noncopyable_function<seastar::future<void> (seastar::future<unsigned long>&&)>&&, seastar::future_state<unsigned long>&&)#1}, unsigned long>
 1: 0x34fd550
 6: seastar::continuation<seastar::internal::promise_base_with_type<void>, seastar::noncopyable_function<seastar::future<void> (seastar::future<void>&&)>, seastar::future<void>::then_wrapped_nrvo<seastar::future<void>, seastar::noncopyable_function<seastar::future<void> (seastar::future<void>&&)> >(seastar::noncopyable_function<seastar::future<void> (seastar::future<void>&&)>&&)::{lambda(seastar::internal::promise_base_with_type<void>&&, seastar::noncopyable_function<seastar::future<void> (seastar::future<void>&&)>&&, seastar::future_state<seastar::internal::monostate>&&)#1}, void>
 3: std::function<seastar::future<void> (unsigned int)>
WARN  2026-02-11 11:22:28,335 [shard 0:main] seastar - Too long queue accumulated for main (19 tasks)
1:
[Backtrace #0]
replica::table::perform_offstrategy_compaction(tasks::task_info)::$_0::operator()(compaction::compaction_group_view&) const (.resume) at ./replica/table.cc:2365
replica::table_populator::populate_subdir(seastar::sharded<sstables::sstable_directory>&) (.resume) at ./replica/distributed_loader.cc:395

3: seastar::continuation<seastar::internal::promise_base_with_type<void>, seastar::noncopyable_function<seastar::future<void> (seastar::future<unsigned long>&&)>, seastar::future<unsigned long>::then_wrapped_nrvo<seastar::future<void>, seastar::noncopyable_function<seastar::future<void> (seastar::future<unsigned long>&&)> >(seastar::noncopyable_function<seastar::future<void> (seastar::future<unsigned long>&&)>&&)::{lambda(seastar::internal::promise_base_with_type<void>&&, seastar::noncopyable_function<seastar::future<void> (seastar::future<unsigned long>&&)>&&, seastar::future_state<unsigned long>&&)#1}, unsigned long>
1:
[Backtrace #1]
tasks::task_manager::task::children::mark_as_finished(utils::tagged_uuid<tasks::task_id_tag>, tasks::task_manager::task::task_essentials) const (.resume) at ./tasks/task_manager.cc:59

```

> Does this improvement somehow affects also backtrace printouts - probably not, right?

It does -- `current_tasktrace()` prints the chain of blocked tasks based on the `waiting_task()` method. Previously, it used RTTI, now we rely on `task->symbol()` there as well.

gusev-p:
Those `seastar::continuation<seastar::internal::promise_base_with_type` are not particularly useful too, maybe we should pass type_info of the actual user-provided lambda from the caller..

gusev-p:
> Those `seastar::continuation<seastar::internal::promise_base_with_type` are not particularly useful too, maybe we should pass type_info of the actual user-provided lambda from the caller..

It seems this is because of the type-erasure in `then_wrapped_maybe_erase`

gusev-p:
> It seems this is because of the type-erasure in `then_wrapped_maybe_erase`

Gemini proposes
```
virtual task_symbol symbol() const override {
        // 1. Check if Func is a wrapper (has target_type) via C++20 Concept/SFINAE
        if constexpr (requires { _func.target_type(); }) {
            const std::type_info& target = _func.target_type();
            // target_type() returns void type_info if empty
            if (target != typeid(void)) {
                return &target;
            }
        }
        
        // 2. Fallback: Return the type of the function object itself.
        // If not type-erased, this is the UserLambda. 
        // If type-erased but empty, this is noncopyable_function.
        return &typeid(_func);
    }
}
```
but we don't have target_type on `noncopyable_function`.. We could probably add this in a follow-up.

gusev-p:
It seems it's simpler to just always use the target entry point address and don't bother with `type_info`, will try

gusev-p:
close in favor of https://github.com/scylladb/seastar/pull/2707

PR #3260, opened 2026-02-17
build: support tuning -ffile-prefix-map

avikivity:
When an executable is built on one machine and distributed, it encodes the full path (esp. with cmake) to the source files. This rarely matches the paths to the source files for whoever ends up debugging core dumps from that executable. Allow using -ffile-prefix-map to relativize the source paths, improving the debugging experience a tiny bit.

The feature is exposed using a new cmake variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,6 +228,12 @@ set (Seastar_CXX_FLAGS
   STRING
   "Semicolon-separated list of extra compilation flags for Seastar itself.")
 
+set (Seastar_SOURCE_SUBDIR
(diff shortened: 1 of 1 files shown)


avikivity:
@travisdowns for cross-project review

avikivity:
@travisdowns ping for review

travisdowns:
LGTM, though we use bazel now, including for the seastar build which already inherently uses machine-independent relative paths. So we don't have much opinion on pure cmake changes.

travisdowns (approved):

PR #3261, opened 2026-02-18
Add metric description override functionality

amnonh:
Currently, metric descriptions are hardcoded at registration time and cannot
be modified without changing source code. This makes it difficult to maintain
documentation separately, update descriptions across versions, or load
descriptions from configuration files.

This series adds a mechanism to override metric descriptions dynamically at
runtime. The implementation adds the set_description_overrides() API that accepts
a map from metric names to description strings. Overrides apply both to
existing metrics (updated immediately) and future metrics (applied during
registration).

Example usage:

    std::unordered_map<std::string, std::string> overrides;
    overrides["reactor_utilization"] = "CPU utilization per reactor (0-1)";
    overrides["memory_allocated_bytes"] = "Total allocated memory in bytes";
    
    seastar::metrics::set_description_overrides(overrides);
    
    // Metrics registered after this will use overridden descriptions
    metric_groups app_metrics;
    app_metrics.add_group("reactor", {
        sm::make_gauge("utilization", sm::description("util"), 
                       [] { return get_utilization(); })
    });

This enables loading descriptions from configuration files or maintaining
centralized documentation databases without requiring recompilation.

Additional tests were added for the new functionality

### DIFF
--- a/include/seastar/core/metrics.hh
+++ b/include/seastar/core/metrics.hh
@@ -134,6 +134,10 @@ public:
     const sstring& str() const {
         return _s;
     }
+    description& operator=(const sstring& s) {
+        _s = s;
+        return *this;
+    }
 private:
     sstring _s;
 };
--- a/include/seastar/core/metrics_api.hh
+++ b/include/seastar/core/metrics_api.hh
@@ -451,6 +451,7 @@ class impl {
     std::vector<relabel_config> _relabel_configs;
     std::vector<metric_family_config> _metric_family_configs;
     internalized_set _internalized_labels;
+    std::unordered_map<std::string, std::string> _description_overrides;
 public:
     value_map& get_value_map() {
         return _value_map;
@@ -472,7 +473,7 @@ public:
     void set_config(const config& c) {
         _config = c;
     }
-
+    void set_description_overrides(const std::unordered_map<std::string, std::string>& description_overrides);
     shared_ptr<metric_metadata> metadata();
 
     std::vector<std::deque<metric_function>>& functions();
@@ -609,6 +610,19 @@ const std::vector<relabel_config>& get_relabel_configs();
  * sm::set_metric_family_configs(fc);
  */
 void set_metric_family_configs(const std::vector<metric_family_config>& metrics_config);
+
+/*
+ * \brief Set an override description map for metrics
+ *
+ * Maps metric names to replacement description strings. When a metric name appears in the map,
+ * the provided description replaces the original description specified during metric registration.
+ *
+ * This affects both existing metrics (descriptions are updated immediately) and metrics that
+ * will be added later (the override is applied when they are registered).
+ *
+ * \param description_overrides Map from metric name to description string
+ */
+void set_description_overrides(const std::unordered_map<std::string, std::string>& description_overrides);
 
 /*
  * \brief return the current metric_family_config
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -136,6 +136,10 @@ void set_metric_family_configs(const std::vector<metric_family_config>& family_c
     impl::get_local_impl()->set_metric_family_configs(family_config);
 }
 
+void set_description_overrides(const std::unordered_map<std::string, std::string>& description_overrides) {
+    impl::get_local_impl()->set_description_overrides(description_overrides);
+}
+
 const std::vector<metric_family_config>& get_metric_family_configs() {
     return impl::get_local_impl()->get_metric_family_configs();
 }
@@ -531,7 +535,7 @@ register_ref impl::add_registration(const metric_id& id, const metric_type& type
         }
     } else {
         _value_map[name].info().type = type.base_type;
-        _value_map[name].info().d = d;
+        _value_map[name].info().d = (_description_overrides.find(name) != _description_overrides.end()) ? description(_description_overrides[name]) : d;
         _value_map[name].info().inherit_type = type.type_name;
         _value_map[name].info().name = id.full_name();
         _value_map[name].info().aggregate_labels = aggregate_labels;
@@ -613,6 +617,31 @@ void impl::set_metric_family_configs(const std::vector<metric_family_config>& fa
             if (fc.name == mf_metadata.mf.name || fc.regex_name.match(mf_metadata.mf.name)) {
                 mf_metadata.mf.aggregate_labels = fc.aggregate_labels;
             }
+        }
+    }
+}
+void impl::set_description_overrides(const std::unordered_map<std::string, std::string>& description_overrides) {
+    _description_overrides = description_overrides;
+    for (auto& [name, description] : _description_overrides) {
+        if (_value_map.find(name) != _value_map.end()) {
+            _value_map[name].info().d = description;
+        }
+
+    }
+
+    if (!_metadata) {
+        // The metadata structure may not have been built yet,
+        // or it may rebuild right now, in this case just return
+        // and it will be updated the next time, setting the dirty to true
+        // is just in case we are somehow in the middle of rebuilding the
+        // metadata.
+        dirty();
+        return;
+    }
+    for (auto& mf_metadata: *_metadata) {
+        auto it = _description_overrides.find(mf_metadata.mf.name);
+        if (it != _description_overrides.end()) {
+            mf_metadata.mf.d = it->second;
         }
     }
 }
--- a/tests/unit/metrics_test.cc
+++ b/tests/unit/metrics_test.cc
@@ -413,3 +413,70 @@ SEASTAR_THREAD_TEST_CASE(test_estimated_histogram) {
         BOOST_CHECK_EQUAL(mh.buckets[i].count, 33 + i);
     }
 }
+
+static seastar::sstring get_metric_description(const seastar::sstring& metric_name) {
+    seastar::foreign_ptr<seastar::metrics::impl::values_reference> values = seastar::metrics::impl::get_values();
+    for (auto&& md : (*values->metadata)) {
+        if (md.mf.name == metric_name) {
+            return md.mf.d.str();
+        }
+    }
+    return "";
+}
+
+SEASTAR_THREAD_TEST_CASE(test_description_overrides_before_metric_added) {
+    using namespace seastar::metrics;
+    namespace sm = seastar::metrics;
+
+    // Set override descriptions before metrics are added
+    std::unordered_map<std::string, std::string> override_map;
+    override_map["test_override_gauge"] = "Overridden gauge description";
+    override_map["test_override_counter"] = "Overridden counter description";
+    sm::set_description_overrides(override_map);
+
+    // Now add the metrics
+    sm::metric_groups app_metrics;
+    app_metrics.add_group("test_override", {
+        sm::make_gauge("gauge", sm::description("Original gauge description"), [] { return 100; }),
+        sm::make_counter("counter", sm::description("Original counter description"), [] { return 200; })
+    });
+
+    // Verify that the descriptions were overridden
+    seastar::sstring gauge_desc = get_metric_description("test_override_gauge");
+    seastar::sstring counter_desc = get_metric_description("test_override_counter");
+
+    BOOST_CHECK_EQUAL(gauge_desc, "Overridden gauge description");
+    BOOST_CHECK_EQUAL(counter_desc, "Overridden counter description");
+}
+
+SEASTAR_THREAD_TEST_CASE(test_description_overrides_after_metric_added) {
+    using namespace seastar::metrics;
+    namespace sm = seastar::metrics;
+
+    // Add metrics first with original descriptions
+    sm::metric_groups app_metrics;
+    app_metrics.add_group("test_override_after", {
+        sm::make_gauge("gauge", sm::description("Original gauge description"), [] { return 100; }),
+        sm::make_counter("counter", sm::description("Original counter description"), [] { return 200; })
+    });
+
+    // Verify original descriptions are in place
+    seastar::sstring gauge_desc = get_metric_description("test_override_after_gauge");
+    seastar::sstring counter_desc = get_metric_description("test_override_after_counter");
+
+    BOOST_CHECK_EQUAL(gauge_desc, "Original gauge description");
+    BOOST_CHECK_EQUAL(counter_desc, "Original counter description");
+
+    // Now set override descriptions after metrics are already added
+    std::unordered_map<std::string, std::string> override_map;
+    override_map["test_override_after_gauge"] = "Overridden gauge description";
+    override_map["test_override_after_counter"] = "Overridden counter description";
+    sm::set_description_overrides(override_map);
+
+    // Verify that the descriptions were overridden
+    gauge_desc = get_metric_description("test_override_after_gauge");
+    counter_desc = get_metric_description("test_override_after_counter");
+
+    BOOST_CHECK_EQUAL(gauge_desc, "Overridden gauge description");
+    BOOST_CHECK_EQUAL(counter_desc, "Overridden counter description");
+}


amnonh:
@avikivity can you take a look?
The idea is that once the functionality is in Seastar, I'll move the descriptions to a file, making it obvious for future developers to do the same.

amnonh:
@avikivity ping

swasik:
@avikivity friendly ping

avikivity:
For an application metric, it could just change the description directly, no?

So this is only for Seastar-defined metrics.

Which ones need overrides?

amnonh:
@avikivity The motivation is to place all the descriptions in a file, it will replace the description in the code. The seastar functionality by itself enables this.

avikivity:
> Currently, metric descriptions are hardcoded at registration time and cannot be modified without changing source code. This makes it difficult to maintain documentation separately, update descriptions across versions, or load descriptions from configuration files.
> 
> This series adds a mechanism to override metric descriptions dynamically at runtime. The implementation adds the set_override_descriptions() API that accepts a map from metric names to description strings. Overrides apply both to existing metrics (updated immediately) and future metrics (applied during registration).
> 
> Example usage:
> 
> ```
> std::map<std::string, std::string> overrides;

Should be unordered_map

> overrides["reactor_utilization"] = "CPU utilization per reactor (0-1)";
> overrides["memory_allocated_bytes"] = "Total allocated memory in bytes";

Adding the underscore breaks the structure.

We can use a small struct for group/name. I expect it will come from some yaml file.

> 
> seastar::metrics::set_override_descriptions(overrides);

set_description_overrides()

> 
> // Metrics registered after this will use overridden descriptions
> metric_groups app_metrics;
> app_metrics.add_group("reactor", {
>     sm::make_gauge("utilization", sm::description("util"), 
>                    [] { return get_utilization(); })
> });
> ```

We need to make the description optional so those who switch don't need to add garbage descriptions (but throw if there is no description override). Something like make_gauge("utilization, sm::from_external_source(), ...);

> 
> This enables loading descriptions from configuration files or maintaining centralized documentation databases without requiring recompilation.
> 
> Additional tests were added for the new functionality

amnonh:
> Should be unordered_map

Will do

> 
> > overrides["reactor_utilization"] = "CPU utilization per reactor (0-1)";
> > overrides["memory_allocated_bytes"] = "Total allocated memory in bytes";
> 
> Adding the underscore breaks the structure.
> 
> We can use a small struct for group/name. I expect it will come from some yaml file.

That is my plan, but it's agnostic to the seastar implementation; whoever uses the API can choose their own format.

> We need to make the description optional so those who switch don't need to add garbage descriptions (but throw if there is no description override). Something like make_gauge("utilization, sm::from_external_source(), ...);
> 

The issue with throwing is the order of execution; some metrics are registered early.  We can't guarantee the description file will be loaded at a specific time.

Once we'll get to move the description to an external file, I was thinking of placing something in the code like:
```
using description_from_file = sestar::metrics::description;

```
To remind users, they should update the description file.

But we can always add that, either inside Seastar or in the application using Seastar. 

I do want to deprecate much of the metrics definition, but that's a different PR and not urgent.

I'll send a V2 shortly

avikivity:
> > Should be unordered_map
> 
> Will do
> 
> > > overrides["reactor_utilization"] = "CPU utilization per reactor (0-1)";
> > > overrides["memory_allocated_bytes"] = "Total allocated memory in bytes";
> > 
> > 
> > Adding the underscore breaks the structure.
> > We can use a small struct for group/name. I expect it will come from some yaml file.
> 
> That is my plan, but it's agnostic to the seastar implementation; whoever uses the API can choose their own format.

How? The set_overrides accepts a map with a particular key/value type choices, made by Seastar.


> 
> > We need to make the description optional so those who switch don't need to add garbage descriptions (but throw if there is no description override). Something like make_gauge("utilization, sm::from_external_source(), ...);
> 
> The issue with throwing is the order of execution; some metrics are registered early. We can't guarantee the description file will be loaded at a specific time.

We can provide them very early as an argument to app_template.

It ties metrics to app_template though.

I don't think it's good for the description to change mid-run.

> 
> Once we'll get to move the description to an external file, I was thinking of placing something in the code like:
> 
> ```
> using description_from_file = sestar::metrics::description;
> ```
> 
> To remind users, they should update the description file.
> 
> But we can always add that, either inside Seastar or in the application using Seastar.
> 
> I do want to deprecate much of the metrics definition, but that's a different PR and not urgent.
> 
> I'll send a V2 shortly

amnonh:
> > > Adding the underscore breaks the structure.
> > > We can use a small struct for group/name. I expect it will come from some yaml file.
> > 
> > 
> > That is my plan, but it's agnostic to the seastar implementation; whoever uses the API can choose their own format.
> 
> How? The set_overrides accepts a map with a particular key/value type choices, made by Seastar.

Right, but the calling app is not limited to any specific file or format. We can indeed add a structure so it would look more like the definition in the code, but then it looks less like how the metrics look for the users.

Using the full metric name pushes the decision of how to create the file higher up the stack.  

> > > We need to make the description optional so those who switch don't need to add garbage descriptions (but throw if there is no description override). Something like make_gauge("utilization, sm::from_external_source(), ...);
> > 
> > 
> > The issue with throwing is the order of execution; some metrics are registered early. We can't guarantee the description file will be loaded at a specific time.
> 
> We can provide them very early as an argument to app_template.
> 
> It ties metrics to app_template though.

Even then, you have seastar-related metrics that start early.

> 
> I don't think it's good for the description to change mid-run.

Not mid-run, but make it safe for any kind of weird load order.

avikivity:
> > > > Adding the underscore breaks the structure.
> > > > We can use a small struct for group/name. I expect it will come from some yaml file.
> > > 
> > > 
> > > That is my plan, but it's agnostic to the seastar implementation; whoever uses the API can choose their own format.
> > 
> > 
> > How? The set_overrides accepts a map with a particular key/value type choices, made by Seastar.
> 
> Right, but the calling app is not limited to any specific file or format. We can indeed add a structure so it would look more like the definition in the code, but then it looks less like how the metrics look for the users.


I meant the key to the map should be

```c++
struct {
     sstring group;
     sstring metric;
};
```

Instead of a single string.

> 
> Using the full metric name pushes the decision of how to create the file higher up the stack.
> 
> > > > We need to make the description optional so those who switch don't need to add garbage descriptions (but throw if there is no description override). Something like make_gauge("utilization, sm::from_external_source(), ...);
> > > 
> > > 
> > > The issue with throwing is the order of execution; some metrics are registered early. We can't guarantee the description file will be loaded at a specific time.
> > 
> > 
> > We can provide them very early as an argument to app_template.
> > It ties metrics to app_template though.
> 
> Even then, you have seastar-related metrics that start early.

I don't think anything starts before app_template.

> 
> > I don't think it's good for the description to change mid-run.
> 
> Not mid-run, but make it safe for any kind of weird load order.

amnonh:
> > > How? The set_overrides accepts a map with a particular key/value type choices, made by Seastar.
> > 
> > 
> > Right, but the calling app is not limited to any specific file or format. We can indeed add a structure so it would look more like the definition in the code, but then it looks less like how the metrics look for the users.
> 
> I meant the key to the map should be
> 
> ```c++
> struct {
>      sstring group;
>      sstring metric;
> };
> ```
> 
> Instead of a single string.
> 

I figured it out afterwards, missed it the first time. The structure is more like how we define the metrics in the code.
It does not look like how users are looking at metrics.

A full name is how metrics are stored in the code and how users see them.
From an implementation, adding a structure will be just as easy.
 
I'm in favour of using the full metric name, but I accept your decision.

> > Using the full metric name pushes the decision of how to create the file higher up the stack.
> > > > > We need to make the description optional so those who switch don't need to add garbage descriptions (but throw if there is no description override). Something like make_gauge("utilization, sm::from_external_source(), ...);
> > > > 
> > > >

travisdowns:
> @avikivity The motivation is to place all the descriptions in a file, it will replace the description in the code. The seastar functionality by itself enables this.

Do you mean also all the seastar metrics descriptions? So any sesatar application would not be standalone, it would also require this file to be packaged alongside it?

amnonh:
> > @avikivity The motivation is to place all the descriptions in a file, it will replace the description in the code. The seastar functionality by itself enables this.
> 
> Do you mean also all the seastar metrics descriptions? So any sesatar application would not be standalone, it would also require this file to be packaged alongside it?

Note that there's no file here, just an API, any application that would use this functionality can create its own file and just use the file to override the description. If it wishes, it can override the seastar descriptions, but we can keep the default ones in the code.

travisdowns:
@amnonh right sorry I quoted the wrong part 🤦 I meant:

> The idea is that once the functionality is in Seastar, I'll move the descriptions to a file, making it obvious for future developers to do the same.

I wasn't sure if by "the descriptions" you refer to seastar metrics or scylladb ones. 

If the former or both, then does it means seastar apps need two files to get the existing fleshed out descriptions?

amnonh:
> > The idea is that once the functionality is in Seastar, I'll move the descriptions to a file, making it obvious for future developers to do the same.
> 
> I wasn't sure if by "the descriptions" you refer to seastar metrics or scylladb ones.
> 
> If the former or both, then does it means seastar apps need two files to get the existing fleshed out descriptions?

I'll keep the default description inside Seastar. I'll add a general file for ScyllaDB, but that's a ScyllaDB thing; other users of Seastar can do the same and override Seastar's default descriptions if they wish (for example, if someone wants to localize the description, or add specific information for it)

include/seastar/core/metrics.hh
```diff
@@ -134,6 +134,10 @@ public:
     const sstring& str() const {
         return _s;
     }
+    description& operator=(const sstring& s) {
+        _s = s;
+        return *this;
+    }
```
avikivity, line 140: Generally tightening a type implicitly is frowned upon. The constructor is not explicit, so this is okay.

include/seastar/core/metrics_api.hh
```diff
@@ -451,6 +451,7 @@ class impl {
     std::vector<relabel_config> _relabel_configs;
     std::vector<metric_family_config> _metric_family_configs;
     internalized_set _internalized_labels;
+    std::unordered_map<std::string, std::string> _description_overrides;
```
avikivity, line 454: I asked that the key represent the group, name structure.

avikivity: and maybe the value type should be a description, not a string.

src/core/metrics.cc
```diff
@@ -531,7 +535,7 @@ register_ref impl::add_registration(const metric_id& id, const metric_type& type
         }
     } else {
         _value_map[name].info().type = type.base_type;
-        _value_map[name].info().d = d;
+        _value_map[name].info().d = (_description_overrides.find(name) != _description_overrides.end()) ? description(_description_overrides[name]) : d;
```
avikivity, line 538: It's really minor, but let's avoid the double lookup.

travisdowns:
> I'll keep the default description inside Seastar. I'll add a general file for ScyllaDB, but that's a ScyllaDB thing; other users of Seastar can do the same and override Seastar's default descriptions if they wish (for example, if someone wants to localize the description, or add specific information for it)

Thanks SGTM then, my concern was if we needed now to package a new file.

> /cc @travisdowns I expect it will be useful for you as well.

I think we wouldn't use it for now? Generally we fix the metric descriptions in the source. Though I understand the motivations here.

include/seastar/core/metrics_api.hh
amnonh: I expressed why I think a full name was better, and as I wrote, I let you make the call, so I'll update it in the next version.
I'll change to description, it's a good idea.

src/core/metrics.cc
amnonh: sure

src/core/metrics.cc
```diff
@@ -616,6 +620,31 @@ void impl::set_metric_family_configs(const std::vector<metric_family_config>& fa
...
+        // The metadata structure may not have been built yet,
+        // or it may rebuild right now, in this case just return
+        // and it will be updated the next time, setting the dirty to true
+        // is just in case we are somehow in the middle of rebuilding the
+        // metadata.
+        dirty();
+        return;
+    }
+    for (auto& mf_metadata: *_metadata) {
+        auto it = _description_overrides.find(mf_metadata.mf.name);
```
amnonh, line 642: @avikivity going from group + metric -> name is easy.
The other way is hard; you can't look at a name and know what the original group and metric were.

There are ways around that, for once, I  can get a map of compose keys (like you wanted), but store a map based on the full name.
I prefer not to change all the metric implementation to reflect that separation if possible.

Please let me know what you prefer and I'll follow

avikivity: But the key is {group, name}, no?

include/seastar/core/metrics_api.hh
avikivity: Both the group and the name. Just keep it structured.

Group = a name = b_c map to a_b_c
Group = a_b name = c map to a_b_c

The implementation right now will get confused, but it can be fixed one day. The API should not bind us to the confusion.

src/core/metrics.cc
amnonh: No, the key is full name, group name is a leftover from the collectd era, there's no concept of groups in Prometheus.
It does make sense to group metrics together, but as Prometheus does not have it, it's not reflected to the users. The only place it exists is in the code, specifically, in the registration.

avikivity: There are groups in the metrics API, no? metrics_group class.

amnonh: @avikivity It's a group of metrics with the same name, but different labels.

avikivity:
```c++
    metric_id(group_name_type group, metric_name_type name,
```

A metric has a compound identifier, not a single string.

amnonh: @avikivity Yes, a metric is a combination of name and labels; 

The description belongs to the metric family:
https://github.com/scylladb/seastar/blob/59225b1c6d2225b67dd2f2cebd3aa22be84b55d3/include/seastar/core/metrics_api.hh#L179

```
struct metric_family_info {
    data_type type;
    metric_type_def inherit_type;
    description d;
    sstring name;
    std::vector<std::string> aggregate_labels;
};
```

And the name here is the full metric name.

avikivity: I don't understand. When I define a metric, I give it a group name ("database") and a metric name ("requests_blocked_memory_current"). Why can't we preserve this?

amnonh: @avikivity I'm not saying we can't; I'm saying we don't. This concept comes from the collectd era; Prometheus doesn't support it, and it exists only in the code.

Users don't know if it's database + requests_blocked_memory_current or database_requests + blocked_memory_current.

So yes, if you want to change how metrics are stored, we can do that. The question is why? Who would use that and how?
What functionality will the users get?

avikivity: It exists in the metrics API. The metrics API isn't Prometheus.

Developers certainly know they defined the metric in terms of a group and a metric in the group. If one part of the API defined metrics in terms of groups and members of a group, and another part of the API defined metrics with a combined name, it's confusing.

amnonh: @avikivity, the question: Do you want me to change the metrics layer implementation?

What about other configurations that we have, and are currently based on metric names?
Changing the description is just one option.

https://github.com/scylladb/seastar/blob/59225b1c6d2225b67dd2f2cebd3aa22be84b55d3/include/seastar/core/metrics_api.hh#L587

https://github.com/scylladb/seastar/blob/59225b1c6d2225b67dd2f2cebd3aa22be84b55d3/include/seastar/core/metrics_api.hh#L611

amnonh: @avikivity See my previous comment: the metric layer store, and later allow configuring metrics based on the full name.

Assuming we won't break existing APIs, if we keep the current implementation of set_description_overrides based on full name, if someone would like to define their metrics description with {group, name} they can still do that, and would pass the full name to set_description_overrides.

I can also add a wrapper for set_description_overrides that accepts unordered_map<{group, name}, description> and call the set_description_overrides with unordered_map<full-name, description>.

amnonh: @avikivity ping, please see my comment

avikivity: I don't understand the relevance of the APIs you highlighted.

To create a metric, you create a group with a name, and create a metric in that group with its own name.

To reference a metric, it seems to be most natural to refer to the group name and to the metric name separately. Not join the group name and the metric name with an underscore. That's a Prometheus thing.

amnonh: @avikivity I'm talking about the current implementation of the metric layer with its current API, before any changes.
Currently, the implementation does not store the group and name independently. There are multiple ways of changing it; those are my comments. Keep in mind, we already have an API for the metric layer; we do not want to break.

avikivity: >the implementation does not store the group and name independently

The implementation does not dictate the API. The API has a distinction between the group name and the metric name. If new APIs are added, they have to preserve the distinction.

amnonh: @avikivity, but there's already an existing API for the metric layer, and there's already an implementation for how metrics are stored, so to support what you're asking, things will need to be changed, I pointed out in the code and the existing API (before any changes) what's not with an agreement to your comment.

avikivity: The API already uses groups, no?

What's not using groups in the API?

amnonh: @avikivity When registering metrics, we use the groups, for anything else, not.

 like: the relable config:
https://github.com/scylladb/seastar/pull/3261/changes#diff-d18eb21b1cc86afdf3709f705743de28bb3bd15e925a145536b8259ae5518058L490

Or the metric family config

avikivity: > @avikivity When registering metrics, we use the groups, for anything else, not.
> 
> like: the relable config: https://github.com/scylladb/seastar/pull/3261/changes#diff-d18eb21b1cc86afdf3709f705743de28bb3bd15e925a145536b8259ae5518058L490

This points at the diff for description overrides.

> 
> Or the metric family config

This points at nothing, so I can't inspect it.

amnonh: @avikivity 
future<metric_relabeling_result> set_relabel_configs(const std::vector<relabel_config>& relabel_configs);

https://github.com/scylladb/seastar/blob/ccec5f27e6d0890348565e0293ae5be2b98edf5a/include/seastar/core/metrics_api.hh#L587

void set_metric_family_configs(const std::vector<metric_family_config>& metrics_config);
https://github.com/scylladb/seastar/blob/ccec5f27e6d0890348565e0293ae5be2b98edf5a/include/seastar/core/metrics_api.hh#L611-L612

PR #3264, opened 2026-02-19
http: Add formatters for http::request and http::reply

xemul:
It's useful to print both while debugging http(s) communication. In Scylla we have a custom formatter for http::request, and an attempt to add one more :) in another module. It's worth having a "standard" one.

This would solve the need to access a deprecated request::content member to print the request body.

### DIFF
--- a/include/seastar/http/request.hh
+++ b/include/seastar/http/request.hh
@@ -434,3 +434,23 @@ using request [[deprecated("Use http::request instead")]] = http::request;
 }
 
 }
+
+template <>
+struct fmt::formatter<seastar::http::request> {
+    template <typename ParseContext>
+    constexpr auto parse(ParseContext& ctx) { return ctx.begin(); }
+    template <typename FormatContext>
+    auto format(const seastar::http::request& rq, FormatContext& ctx) const {
+        auto out = fmt::format_to(ctx.out(), "{} {}", rq._method, rq._url);
+        for (const auto& h : rq._headers) {
+            out = fmt::format_to(out, " {}:{}", h.first, h.second);
+        }
+        if (!rq.body_writer) {
+            const auto& body = seastar::http::internal::deprecated_content(rq);
+            if (!body.empty()) {
+                out = fmt::format_to(out, " {}", body);
+            }
+        }
+        return out;
+    }
+};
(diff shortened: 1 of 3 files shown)


include/seastar/http/request.hh
```diff
@@ -434,3 +434,21 @@ using request [[deprecated("Use http::request instead")]] = http::request;
 }
 
 }
+
+template <>
+struct fmt::formatter<seastar::http::request> {
+    constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+    auto format(const seastar::http::request& rq, fmt::format_context& ctx) const {
+        auto out = fmt::format_to(ctx.out(), "{} {}", rq._method, rq._url);
```
kreuzerkrieg, line 444: how about just calling `request_line()`?

include/seastar/http/request.hh
```diff
@@ -434,3 +434,21 @@ using request [[deprecated("Use http::request instead")]] = http::request;
...
+        for (const auto& h : rq._headers) {
```
kreuzerkrieg, line 445: what about all other members of this class, like _trailing_headers?

xemul: Possible, but none of the printers in Scylla show them, I just copied it from there

include/seastar/http/request.hh
xemul: It allocates a new string, and runs though `format_url()`. Not sure if it's good or bad for printer

xemul:
upd:
- use boost::split instead of std::views::split to comfort CI compilation

xemul:
upd:
- templatize formatter methods for CI's version of libfmt

xemul:
@nyh , please review

xemul:
@nyh , please review

xemul:
@nyh , please review

xemul:
@nyh , please review

xemul:
@nyh , please review. 
We have deprecation warning in Scylla compilation (formatting http requests/responses) and this PR is supposed to fix it

xemul:
@nyh , please review

xemul:
@nyh , please review

PR #3266, opened 2026-02-22
src/core/reactor: introduce reactor::get_backend_name()

Matan-B:
```
reactor_backend_selector determines which reactor backends are
available on the system. The selected backend is stored in reactor::_backend,
but there is currently no way to query the selected backend type at runtime.

Add reactor::get_backend_name(), which exposes the backend name
via engine().get_backend_name().

Note: While smp::configure() does log the selected backend using reactor_opts,
a runtime getter is useful for user visibility.
```

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -499,6 +499,8 @@ public:
         return queue != _io_queues.end() ? queue->second.get() : nullptr;
     }
 
+    std::string_view get_backend_name() const;
+
 private:
     future<> update_bandwidth_for_queues(internal::priority_class pc, uint64_t bandwidth);
     void rename_queues(internal::priority_class pc, sstring new_name);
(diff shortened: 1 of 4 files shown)


include/seastar/core/reactor.hh
```diff
@@ -499,6 +499,8 @@ public:
         return queue != _io_queues.end() ? queue->second.get() : nullptr;
     }
 
+    std::string_view get_backend_type() const;
```
avikivity: Please call it backend_name(), since it returns a name, not a type.

src/core/reactor_backend.cc
```diff
@@ -1445,6 +1453,9 @@ class reactor_backend_uring final : public reactor_backend {
     ~reactor_backend_uring() {
         ::io_uring_queue_exit(&_uring);
     }
+    virtual std::string_view get_backend_name() const override {
+      return "io_uring";
```
avikivity: We use 4-space indents, not 2 (elsewhere too).

PR #3270, opened 2026-03-05
Replace duplicate SEGV_MAPERR check in sigsegv_action() with SEGV_ACC…

xemul:
…ERR.

Looks like a misprint.
Permission-fault segfaults also worth printing their si_addr.


xemul:

travisdowns:
Good eye, looks like a typo on my part.

PR #3271, opened 2026-03-05
Fix _shard_to_numa_node_mapping double population

xemul:
The _shard_to_numa_node_mapping is first resize()-d to smp::count elements and is thus filled with zeros. Then the subsequent loop push_back()-s the same number more entries, resulting in a two times larger vector than needed, having wrong first half.

Presumably, the intent was to _reserve_ the vector, then populate.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4435,7 +4435,7 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
         memory::configure_minimal();
     }
 
-    _shard_to_numa_node_mapping.resize(smp::count);
+    _shard_to_numa_node_mapping.reserve(smp::count);
     for (unsigned i = 0; i < smp::count; i++) {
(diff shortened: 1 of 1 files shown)


michoecho:
Right. Thanks for the fix.

Did you find this just by reading the code or did it lead to some observable bug?

raphaelsc:
for my education, what's the side effect? the description explains what's wrong with the code, but not what problems it caused.

michoecho:
> for my education, what's the side effect? the description explains what's wrong with the code, but not what problems it caused.

This piece of code is something I added so that application shards can learn what NUMA nodes they are on.
In Scylla, this is used for replicating compression dictionaries across NUMA nodes so that every shard can use a dictionary on its own NUMA node. (Compressing with dictionaries can naturally involve lots of random memory accesses, which would be particularly expensive if the memory was on a different NUMA node. We never tested if it's a real performance problem in practice).
So AFAICS the effect is that this optimization doesn't work, and there's a single copy of each dictionary, used by all shards.

xemul:
> Did you find this just by reading the code or did it lead to some observable bug?

I was reading start-up code and this place caught my eye

Issue #3201, opened 2026-01-13
unfortunate io_scheduler behavior with low iodepth and multiple shards


travisdowns:
When using io-properties with multiple shards, a Seastar application may experience unexpectedly poor performance degradation, far below the capabilities of the hardware and fall less than the io-properties limit. With a single shard, the system achieves ~24K IOPS, but with two shards actively running I/O, each shard only achieves ~1.4K IOPS, resulting in a **~17x per-shard performance degradation**.

## Environment

- **Seastar version**: Current master (commit bbd0001a)
- **Test workload**: Random 4KB reads, iodepth=1
- **Storage**: NVMe SSD on XFS
- **Disk capabilities** (measured by iotune):
  - Random read IOPS: 268,337 IOPS
  - Random write IOPS: 134,175 IOPS
  - Sequential read: 1200 MB/s
  - Sequential write: 576 MB/s

## io-properties Configuration

The exact values don't matter too much, but they can't be arbitrarily high. I used:

```yaml
disks:
  - mountpoint: /mnt/xfs
    read_iops: 1000000
    read_bandwidth: 1000000000
    write_iops: 134608
    write_bandwidth: 604995968
```

## Test Configuration Files

Doesn't matter if it's randread or randwrite, both show the same (and probably other modes too).

### io2.yaml (single shard)
```yaml
- name: highprio
  shards: [0]
  type: randread
  shard_info:
    parallelism: 1
    reqsize: 4kB
    shares: 1000
    think_time: 0
```

### io2-multishard.yaml (two shards)
```yaml
- name: highprio
  shards: [0, 1]
  type: randread
  shard_info:
    parallelism: 1
    reqsize: 4kB
    shares: 1000
    think_time: 0
```

## Test Results

Without any `--io-properties-file` IOPS are about 35K, latency bound at about ~30us
per IO.

### Single Shard Test

```bash
$ build/release/apps/io_tester/io_tester --io-properties-file ~/io-props.yaml \
    --conf ~/io2.yaml --storage /mnt/xfs/io_tester --duration=5 -c1
Job highprio -> sched class highprio
    IOPS: 24319.1328
```

Results are similar with -c2 also.

**Result: 24,319 IOPS on single shard**

### Two Shard Test
```bash
$ build/release/apps/io_tester/io_tester --io-properties-file ~/io-props.yaml \
    --conf ~/io2-multishard.yaml --storage /mnt/xfs/io_tester --duration=5 -c2
[...]

PR #3275, opened 2026-03-06
fair_queue: Subtract from _queued_capacity on pop_front()

xemul:
When a request is withdrawn from fair queue, the total queued capacity counter should be decreased, not increased.

Bug introduced by ab61e44b.

refs: #3201


avikivity:
How come no test caught this? Please follow up with one.

PR #3276, opened 2026-03-06
memory: expose cumulative allocated bytes statistic

nuivall:
Add total_bytes_allocated() to memory::statistics and update it on each successful allocation.

The counter is useful, for example, for estimating the memory usage of non-yielding code blocks.


travisdowns:
This added ~3 instructions on the hot path!

PR #3278, opened 2026-03-07
websocket: add template parameter for text/binary frame mode and implement client-side WebSocket

wineway:
Add a template parameter `text_frame` to `connection`, `client_connection`, and `client` classes to support both text and binary frame modes. Implement the client-side WebSocket functionality including:
- WebSocket client connection handling with HTTP upgrade handshake
- Frame exchange protocol implementation
- Client connection tests
- Demo application for WebSocket client
- Introducing _half_close fixes the issue where, when actively sending a CLOSE frame, the remote returning a CLOSE frame before the input is closed would cause a second CLOSE frames to be sent.

### DIFF
--- a/include/seastar/websocket/client.hh
+++ b/include/seastar/websocket/client.hh
@@ -37,7 +37,8 @@ namespace seastar::experimental::websocket {
  * Represents a single WebSocket connection initiated by a client.
  * Handles the HTTP upgrade handshake and WebSocket frame exchange.
  */
-class client_connection : public basic_connection<true> {
+template <bool text_frame = false>
+class client_connection : public basic_connection<true, text_frame> {
     http_response_parser _http_parser;
     sstring _resource;
     sstring _host;
@@ -82,8 +83,9 @@ public:
  * A client capable of establishing a WebSocket connection to a server.
  * Manages the connection lifecycle.
  */
+template<bool text_frame = false>
 class client {
-    std::unique_ptr<client_connection> _conn;
+    std::unique_ptr<client_connection<text_frame>> _conn;
     gate _task_gate;
 
 public:
--- a/include/seastar/websocket/common.hh
+++ b/include/seastar/websocket/common.hh
@@ -53,7 +53,7 @@ public:
 /*!
  * \brief a server WebSocket connection
  */
-template <bool is_client = false>
+template <bool is_client = false, bool text_frame = false>
 class basic_connection : public boost::intrusive::list_base_hook<> {
 protected:
     using buff_t = temporary_buffer<char>;
@@ -174,7 +174,7 @@ protected:
     future<> send_data(opcodes opcode, temporary_buffer<char> buff);
 };
 
-using connection = basic_connection<false>;
+using connection = basic_connection<false, false>;
 
 std::string sha1_base64(std::string_view source);
 std::string encode_base64(std::string_view source);
--- a/src/websocket/client.cc
+++ b/src/websocket/client.cc
@@ -43,18 +43,20 @@ static sstring generate_websocket_key() {
     return sstring(encode_base64(std::string_view(raw, 16)));
 }
 
-client_connection::client_connection(connected_socket&& fd, sstring resource,
+template <bool text_frame>
+client_connection<text_frame>::client_connection(connected_socket&& fd, sstring resource,
                                      sstring host, sstring subprotocol,
                                      handler_t handler)
-    : basic_connection(std::move(fd))
+    : basic_connection<true, text_frame>(std::move(fd))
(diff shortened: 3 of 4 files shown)


demos/websocket_client_demo.cc
```diff
@@ -0,0 +1,112 @@
...
+                ws_client.close().get();
+            });
+
+            if (use_tls) {
+                auto creds = make_shared<tls::certificate_credentials>();
+                creds->set_system_trust().get();
+                ws_client.connect(addr, std::move(creds),
+                    sstring(path), sstring(host), "",
+                    [] (input_stream<char>& in, output_stream<char>& out) -> future<> {
+                        return repeat([&in]() {
```
avikivity: Please use coroutines in demos, the code is much more readable.

(also in the core code, when performance allows, which is almost always the case)

tests/unit/websocket_test.cc
```diff
@@ -251,3 +252,68 @@ SEASTAR_TEST_CASE(test_websocket_parser_split) {
...
+
+SEASTAR_TEST_CASE(test_websocket_client_server) {
+    return seastar::async([] {
+        loopback_connection_factory factory;
+        loopback_socket_impl lsi(factory);
+
+        // Setup server side
+        websocket::server ws;
+        ws.register_handler("echo", [] (input_stream<char>& in,
+                        output_stream<char>& out) {
```
avikivity: Also prefer coroutines in tests.

src/websocket/client.cc
```diff
@@ -0,0 +1,167 @@
...
+#include <seastar/websocket/client.hh>
+#include <seastar/core/when_all.hh>
+#include <seastar/core/loop.hh>
+#include <seastar/http/reply.hh>
+
+namespace seastar::experimental::websocket {
+
+using namespace std::string_view_literals;
+
+constexpr auto magic_key_suffix_client = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"sv;
```
avikivity, line 33: Where does this magic come from?

demos/CMakeLists.txt
```diff
@@ -65,6 +65,9 @@ seastar_add_demo (hello-world
 seastar_add_demo (websocket_server
   SOURCES websocket_server_demo.cc)
```
avikivity, line 66: > websocket: impl client side websocket function

Please update the changelog for this patch with references to the relevant RFCs.

include/seastar/websocket/common.hh
```diff
@@ -53,6 +53,7 @@ public:
 /*!
  * \brief a server WebSocket connection
  */
+template <bool is_client = false>
 class connection : public boost::intrusive::list_base_hook<> {
```
avikivity: Is this connection class exposed to user code?

I would guess not, but let's be safe: name the template basic_connection\<is_client>, and alias basic_connection\<false> to connection.

src/websocket/client.cc
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    websocket_logger.debug("WebSocket client handshake completed");
+}
+
+future<> client_connection::process() {
+    co_await send_http_upgrade_request();
+    co_await read_http_upgrade_response();
+
+    co_await when_all_succeed(
```
avikivity: Better: `co_await coroutine::all(...)`

src/websocket/client.cc
```diff
@@ -0,0 +1,167 @@
...
+        _handler(_input, _output).handle_exception([this] (std::exception_ptr e) {
+            return _read_buf.close().then([e = std::move(e)] () mutable {
+                return make_exception_future<>(std::move(e));
+            });
+        }),
+        do_until([this] { return _done; }, [this] { return read_one(); }),
```
avikivity: And better to use nested coroutines.

include/seastar/websocket/client.hh
```diff
@@ -36,7 +36,8 @@ namespace seastar::experimental::websocket {
  * Represents a single WebSocket connection initiated by a client.
  * Handles the HTTP upgrade handshake and WebSocket frame exchange.
  */
```
avikivity, line 39: > websocket: add template parameter for text/binary frame mode


The changelog should explain what this is.

wineway:
hi @avikivity I've addressed your review comments, please take another look, thanks!
 
- Added detailed changelog in commit message
- Replaced `future` base code with `coroutines` in demo, tests, and long-running tasks
- Added a comment explaining the origin of `258EAFA5-E914-47DA-95CA-C5AB0DC85B11`
- Added `connection` as an alias for `basic_connection<false>`
- Fixed flaky test caused by `connect.close()` reentrancy

PR #3281, opened 2026-03-11
reactor_backend: replace virtual bool methods with const bool_class m…

xemul:
…embers

Replace the virtual methods do_blocking_io() and have_aio_fdatasync() with const bool_class member variables on the reactor_backend base class. These values are fixed per backend and never change at runtime, so virtual dispatch is unnecessary.

Introduce uses_blocking_io and supports_aio_fdatasync bool_class types for type safety. Each derived backend now explicitly passes both values to the base constructor with no defaults.

### DIFF
--- a/src/core/reactor_backend.hh
+++ b/src/core/reactor_backend.hh
@@ -28,6 +28,7 @@
 #include <seastar/core/internal/poll.hh>
 #include <seastar/core/internal/linux-aio.hh>
 #include <seastar/core/cacheline.hh>
+#include <seastar/util/bool_class.hh>
 
 #include <fmt/ostream.h>
 #include <sys/time.h>
@@ -169,6 +170,14 @@ public:
 // file-descriptors (reactor_backend_epoll), one implementation based on
 // linux aio, and one implementation based on io_uring.
 class reactor_backend {
+protected:
+    using uses_blocking_io = bool_class<struct uses_blocking_io_tag>;
+    using supports_aio_fdatasync = bool_class<struct supports_aio_fdatasync_tag>;
+
+private:
+    const uses_blocking_io _blocking_io;
+    const supports_aio_fdatasync _aio_fdatasync;
+
 public:
     virtual ~reactor_backend() {};
     virtual std::string_view get_backend_name() const = 0;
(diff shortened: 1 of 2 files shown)


src/core/reactor_backend.hh
```diff
@@ -170,6 +171,12 @@ public:
 // linux aio, and one implementation based on io_uring.
 class reactor_backend {
 public:
+    using uses_blocking_io = bool_class<struct uses_blocking_io_tag>;
+    using supports_aio_fdatasync = bool_class<struct supports_aio_fdatasync_tag>;
+
+    const uses_blocking_io blocking_io;
+    const supports_aio_fdatasync aio_fdatasync;
```
avikivity: Would be nicer to keep them as methods.

src/core/reactor.cc
```diff
@@ -1692,17 +1692,17 @@ void pollable_fd::shutdown(int how, shutdown_kernel_only kernel_only) {
...
 reactor::make_pollable_fd(socket_address sa, int proto) {
-    int maybe_nonblock = _backend->do_blocking_io() ? 0 : SOCK_NONBLOCK;
+    int maybe_nonblock = bool(_backend->blocking_io) ? 0 : SOCK_NONBLOCK;
     file_desc fd = file_desc::socket(sa.u.sa.sa_family, SOCK_STREAM | maybe_nonblock | SOCK_CLOEXEC, proto);
     return pollable_fd(std::move(fd));
 }
 
 bool reactor::posix_sock_need_nonblock() const {
-    return !_backend->do_blocking_io();
+    return !bool(_backend->blocking_io);
```
avikivity: I think the cast to bool is unneeded (in the other places too).

xemul: It is needed, operator bool() is marked as explicit on bool_class

src/core/reactor_backend.hh
xemul: OK, preserved them as methods in next version

xemul:
upd:
- keep have_aio_fdatasync() and do_blocking_io() as methods

src/core/reactor.cc
avikivity: https://godbolt.org/z/dEMGfr754

PR #3285, opened 2026-03-12
Implement fully mixed mode for output_stream-s

xemul:
Nowadays output_stream supports two write() options -- buffered and zero-copy. The former copies provided data into an internal buffer and the latter just grabs the buffers and stores them in internal vector. 

Combining both methods together is limitedly possible -- one may do zero-copy writes after buffered writes, but must flush and (!) wait for the flush to actually happen to make another buffered write again. The "waiting" can be complicated in case batch-flushing option is ON, so this mode is very delicate and must be used with care.

This PR makes the output_stream support intermixed buffered and zero-copy writes in any order without restrictions :crossed_fingers: . The PR also includes an exhaustive unit test that validates all combinations of buffered vs zero-copy write-s for 1 through 4 buffers of various sizes (below, equal and up to 3x times above the stream buffer size) for trim-to-size option on and off.

All patches are AI-assisted (Claude Sonnet 4.6)

### DIFF
--- a/tests/unit/output_stream_test.cc
+++ b/tests/unit/output_stream_test.cc
@@ -24,6 +24,7 @@
 #include <seastar/core/sstring.hh>
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
+#include <seastar/testing/random.hh>
 #include <vector>
 #include <list>
 #include <deque>
@@ -169,6 +170,7 @@ static void for_each_chunk_combination(Fn fn) {
 
 // Calls fn(write_types) for every assignment of buffered/zero_copy to
 // n write calls (2^n patterns total; n <= MAX_CHUNKS so at most 16).
+#ifndef SEASTAR_DEBUG
 template <typename Fn>
 static void for_each_type_pattern(size_t n, Fn fn) {
     std::vector<write_type> pattern(n);
@@ -179,6 +181,31 @@ static void for_each_type_pattern(size_t n, Fn fn) {
         fn(pattern);
     }
 }
+#else
+// In SEASTAR_DEBUG builds, sample ~10% of patterns to keep sanitizer
+// run times acceptable. The all-buffered (0...0) and all-zero-copy
+// (1...1) patterns are always included.
+template <typename Fn>
+static void for_each_type_pattern(size_t n, Fn fn) {
+    const size_t total = size_t(1) << n;
+    size_t sample_size = std::max(size_t(2), size_t(std::round(total * 0.1)));
+
+    std::vector<size_t> masks(total);
+    std::iota(masks.begin(), masks.end(), 0);
+    std::shuffle(masks.begin(), masks.end(), seastar::testing::local_random_engine);
+    // Ensure all-buffered (0) and all-zero-copy (total-1) are always first.
+    std::swap(*std::find(masks.begin(), masks.end(), size_t(0)), masks[0]);
+    std::swap(*std::find(masks.begin(), masks.end(), total - 1), masks[1]);
+
+    std::vector<write_type> pattern(n);
+    for (size_t i = 0; i < sample_size; i++) {
+        for (size_t j = 0; j < n; j++) {
+            pattern[j] = (masks[i] >> j) & 1 ? write_type::zero_copy : write_type::buffered;
+        }
+        fn(pattern);
+    }
+}
+#endif
 
 // Builds a string of `len` bytes filled with a cycling pattern,
 // so that data integrity failures produce readable diffs.


avikivity:
Not reviewed yet, but wow

include/seastar/core/iostream-impl.hh
```diff
@@ -111,12 +111,16 @@ future<> output_stream<CharType>::write(std::span<temporary_buffer<CharType>> bu
     size_t size = std::accumulate(bufs.begin(), bufs.end(), size_t(0), [] (size_t s, const auto& b) { return s + b.size(); });
     if (size != 0) {
```
avikivity, line 112: > allocation), share the filled prefix as _buf.share(0, _end) and
  advance _buf past it with trim_front(_end). The same underlying

This is good independently of the changes, but it does cause a new allocation, for the reference count.

I think we should enhance the allocator to be able to return a (ptr, refcount_ptr) pair, with the refcount stored adjacent to the memory (for small allocations) or in struct page (for large allocations). IIRC the freebsd kernel has a similar facility.

Not for now.

include/seastar/core/iostream-impl.hh
```diff
@@ -370,29 +383,69 @@ template <typename CharType>
...
     try {
-        SEASTAR_ASSERT(_zc_bufs.empty() && "Mixing buffered writes and zero-copy writes not supported yet");
         if (!_end && (n >= _size)) {
             temporary_buffer<char> tmp = _fd.allocate_buffer(n);
             std::copy(buf, buf + n, tmp.get_write());
+            if (!_zc_bufs.empty()) {
+            // No buffered data yet, but zero-copy data is pending.
+            // Append to _zc_bufs so ordering is preserved.
+            _zc_len += n;
+            _zc_bufs.emplace_back(std::move(tmp));
```
avikivity: No reason for broken indents here.

avikivity: I'm not sure it matters, but better to switch the order of these two lines for exception safety. One isn't supposed to call flush() after an exception, so the inconsistency likely doesn't matter, but better to be careful.

include/seastar/core/iostream-impl.hh
```diff
@@ -370,29 +383,69 @@ template <typename CharType>
...
+            if (_zc_len >= _size) {
```
avikivity: Please document the invariants near _buf and _zc_bufs.

If I understand correctly, the logical state of the stream is now _zc_bufs followed by span(_buf.data(), _end). If a copying write comes next it is appended to _bug, if a zero copy write comes the used prefix of _buf is appended to _zc_bufs, then the new writes, then the reminder of _buf.

It may make sense to rename _zc_bufs to _prefix_bufs.

include/seastar/core/iostream-impl.hh
```diff
@@ -419,12 +472,21 @@ void add_to_flush_poller(output_stream<char>& x) noexcept;
...
+        if (_zc_bufs.empty()) {
         _buf.trim(_end);
         _end = 0;
         return _fd.put(std::move(_buf)).then([this] {
             return _fd.flush();
         });
-    } else if (!_zc_bufs.empty()) {
+        } else {
+            // Fold buffered tail into the zero-copy vector and flush together.
+            _zc_bufs.emplace_back(_buf.share(0, _end));
```
avikivity, line 482: I wonder if using a small_vector\<temporary_buffer> where we append to the last vector would simplify things.

avikivity:
Very good.

include/seastar/core/iostream-impl.hh
xemul: Presumably, you're reviewing only [this patch](https://github.com/scylladb/seastar/pull/3285/changes/9e3b578a9af3f69a9d243586023bd552a8044a8d). The indentation is left broken deliberately to simplify the review (well -- my review from what AI generated), [next patch](https://github.com/scylladb/seastar/pull/3285/changes/3e57be4ec3f6a0f107c4ada7d383459cd5cba3a4) restores it

include/seastar/core/iostream-impl.hh
xemul: Yes, that's how it works now. I'll add the code comment

include/seastar/core/iostream-impl.hh
avikivity: Yes. But here you introduce an if and some code it controls, and the new code it controls is not indented.

No need to change.

include/seastar/core/iostream-impl.hh
avikivity: It think it also makes sense to rename _size, it's too generic. Maybe call it _buffer_size.

xemul:
upd:
- renamed _size -> _buffer_size
- added comment about new _buf/_zc_bufs lifecycle

include/seastar/core/iostream-impl.hh
xemul: Fixed both in next version

include/seastar/core/iostream-impl.hh
xemul: Not really. When switching from buffered write to zero-copy write there's a need to maintain two buffers -- the currently accumulated one (_buf) and its remnant if it wasn't filled. Currently this is done with the help of _buf itself and switching everything onto unified small_vector<temporary_buffer, 1> complicates this state management

avikivity: I think you're right.

include/seastar/core/iostream-impl.hh
avikivity: This?

xemul:
upd:
- made slow-write zero-copy path thread safer

include/seastar/core/iostream-impl.hh
xemul: Fixed in next version. I believe there are more (preexisting) places like that, but I didn't check them carefully

PR #3287, opened 2026-03-12
file: fix `dup()` read-only check in `posix_file_impl::do_dup()`

ballard26:
The current check `(_open_flags & open_flags::ro) != open_flags{}` is a no-op because `open_flags::ro` is `O_RDONLY`(0). `_open_flags & 0` will always be 0 and the check will never be true.

The test, `file_rw_dup_test`, still passes because the created file is always using the `append_challenged_posix_file_impl` impl. This impl unconditionally throws on `dup()` which is what the test expects.

This PR fixes this by making the mask `(open_flags::rw | open_flags::wo)` instead and adds a new test that should always use the `posix_file_impl` impl.

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -170,7 +170,7 @@ void posix_file_impl::configure_io_lengths() noexcept {
 template <typename FileImpl>
 std::unique_ptr<seastar::file_handle_impl>
 posix_file_impl::do_dup() {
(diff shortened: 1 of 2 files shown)


ballard26:
this is a dupe of https://github.com/scylladb/seastar/pull/3274

PR #3288, opened 2026-03-13
chunked_fifo: make cached chunk retention configurable

ballard26:
In Redpanda we can maintain thousands of mostly idle `chunked_fifo` objects per shard. Each briefly sees a spike in use then largely stays empty for the remainder of its lifetime. This results in empty `chunked_fifo`s consuming a significant amount of memory. And in some rare cases we've seen OOMs due to that amount of memory reaching hundreds of megabytes. 

Therefore while caching a free chunk is a good tradeoff for queues that repeatedly push and pop, it'd be desirable to disable it for our usecase. Specifically in `expiring_fifo` which has been the cause of our observed OOMs. 

This PR therefore:
  - Makes `save_free_chunks` a template parameter (defaulting to 1 to preserve existing behavior).
  - Sets `save_free_chunks = 0` in `expiring_fifo`, since it caches the front element separately there is far less reasons to cache an entire chunk.
  - Adds the `nfree_chunks()` method for testing.

### DIFF
--- a/tests/perf/container_perf.cc
+++ b/tests/perf/container_perf.cc
@@ -233,3 +233,76 @@ PERF_TEST_F(container_perf, iter_small_boost_deque) {
     return iteration_bench<boost_deque_traits>(small_size);
 }
 
+template <size_t N>
+struct sized_elem {
+    char data[N];
+    sized_elem() { std::memset(data, 0, N); }
+    sized_elem(int i) { std::memset(data, static_cast<char>(i), N); }
+};
+
+template <typename Container>
+size_t push_pop_single_bench() {
+    using T = typename Container::value_type;
+    Container c;
+
+    perf_tests::start_measuring_time();
+    for (size_t i = 0; i < total_iters; i++) {
+        c.push_back(T(i));
+        perf_tests::do_not_optimize(c.front());
+        c.pop_front();
+    }
+    perf_tests::stop_measuring_time();
+
+    return total_iters;
+}
+
+// Push/pop benchmarks varying item size with chunk=1
+PERF_TEST_F(container_perf, push_pop_1b_chunk1_save1) {
+    return push_pop_single_bench<chunked_fifo<sized_elem<1>, 1, 1>>();
+}
+
+PERF_TEST_F(container_perf, push_pop_1b_chunk1_save0) {
+    return push_pop_single_bench<chunked_fifo<sized_elem<1>, 1, 0>>();
+}
+
+PERF_TEST_F(container_perf, push_pop_100b_chunk1_save1) {
+    return push_pop_single_bench<chunked_fifo<sized_elem<100>, 1, 1>>();
+}
+
+PERF_TEST_F(container_perf, push_pop_100b_chunk1_save0) {
+    return push_pop_single_bench<chunked_fifo<sized_elem<100>, 1, 0>>();
+}
+
+PERF_TEST_F(container_perf, push_pop_1000b_chunk1_save1) {
+    return push_pop_single_bench<chunked_fifo<sized_elem<1000>, 1, 1>>();
+}
+
+PERF_TEST_F(container_perf, push_pop_1000b_chunk1_save0) {
+    return push_pop_single_bench<chunked_fifo<sized_elem<1000>, 1, 0>>();
+}
+
+// Push/pop benchmarks varying item size with chunk=128
+PERF_TEST_F(container_perf, push_pop_1b_chunk128_save1) {
+    return push_pop_single_bench<chunked_fifo<sized_elem<1>, 128, 1>>();
+}
+
+PERF_TEST_F(container_perf, push_pop_1b_chunk128_save0) {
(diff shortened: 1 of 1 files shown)


xemul:
> Since expiring_fifo caches the front element separately, the underlying chunked_fifo rarely benefits from retaining a free chunk.

Frankly, I fail to see how caching the first element _implies_ that caching remaining chunk is pointless

travisdowns:
> Frankly, I fail to see how caching the first element _implies_ that caching remaining chunk is pointless

It's not pointless but I guess the case where the queue flips between 0 and 1 (but rarely more) elements is very common (and IIRC was among the motivations behind caching a chunk in chunked_fifo) and this benefit is already captured by the size 1 cache in expiring fifo. So the cache in chunked_fifo is less useful in that case.

In Redpanda we had a chunked_fifo-like structure, which didn't cache an empty chunk, but did still use fixed size chunks so as soon as the container held one element, it would use `sizeof(T) * elems_per_chunk` memory (close to 32K or 128K bytes depending on the container config). This is unexpected for the casual container user and turned out to be a large footgun: I think about half of production OOM incidents were related to many copies of this container having 1 or a few elements, with memory use 100x higher than you'd expect. We replaced it with a variation that uses small allocations initially, growing (like vector, but not contiguous).

ballard26:
> Frankly, I fail to see how caching the first element implies that caching remaining chunk is pointless

Fair, I likely worded it too strongly. Definitely not pointless as you and @travisdowns mentioned. I meant to say it removes one of the reasons for needing the cached free chunk. There are still other reasons like frequent additions/removals of elements near a chunk boundary. However, the case where the queue flips between 0 and 1 is the most common for us as Travis said.

avikivity:
I appreciate you took care not to regress existing workloads by making it configurable and off by default. You correctly predicted that if you just changed it I'd ask about regressions.

But now I must ask the opposite question: do we know of a use-case that will in fact regress?

I suppose we can't really tell.

In ScyllaDB we don't have a performance sensitive user of chunked_fifo. There is only one non-test user (which is not performance sensitive) that would be unaffected by this change (it uses reserve()).

There are a lot of internal Seastar users, for example shared_future uses abortable_fifo which, similar to expiring_fifo, would benefit from dropping the reservation.

It might make sense to default to 0.

avikivity:
> We replaced it with a variation that uses small allocations initially, growing (like vector, but not contiguous).

An exponentially growing chunked_fifo bounded by the blessed 128k sounds like a good idea.

avikivity:
btw I remember chunked_fifo related OOMs in ScyllaDB as well.

travisdowns:
> An exponentially growing chunked_fifo bounded by the blessed 128k sounds like a good idea.

Yes, this is what we do in [chunked_vector](https://github.com/redpanda-data/redpanda/blob/dev/src/v/container/chunked_vector.h#L30) - and we grow in a way that's sensitive to the known sesatar allocator size classes to avoid waste there too (we use 1.5 grow factor, if you use 2 it happens automatically).

No problems with that since introduction.

ballard26:
Out of curiosity I went ahead and added some microbenchmarks to see the performance differences between the options in the case where the queue flips between 0 and 1 element. The following table is a run in from a release build of those microbenchmarks:

   | item_size | items_per_chunk | saved_free_chunks=1 | saved_free_chunks=0 |
  |---|---|---|---|
  | 1 B | 1 | 5.53 ns | 10.44 ns |
  | 1 B | 128 | 5.96 ns | 10.56 ns |
  | 100 B | 1 | 7.22 ns | 11.03 ns |
  | 100 B | 128 | 7.76 ns | 11.76 ns |
  | 1 KB | 1 | 45.98 ns | 50.34 ns |
  | 1 KB | 128 | 44.45 ns | 64.15 ns |
  
I also added a case where `items_per_chunk=1` to each permutation of the test to represent what the perf would be if we implemented an exponentially growing `chunked_fifo`. Overall though, with the release build cfg, not caching a free chunk increases the runtime of a push & pop by at most ~90%. 

For our use cases `chunk_fifo` is plenty fast enough without caching a chunk. And free/wasted memory is by far the larger concern. If you folks also lack any perf-sensitive users would it be fine if I changed the PR to disable caching free chunks by default?

travisdowns:
@avikivity I suggest we set this to 0 (no chunk caching by default) based on the balace of tradeoffs here. Agree?

Also @scylladb/seastar-maint for any other thoughts.

avikivity:
> @avikivity I suggest we set this to 0 (no chunk caching by default) based on the balace of tradeoffs here. Agree?


Yes

ballard26:
This PR has been updated to set the number of retained free chunks to 0 by default. Also I've dropped the commit that changed the expiring fifo as it was no longer needed. The microbenchmark commit isn't entirely needed either so I'm happy to drop it if desired.

travisdowns:
This looks good to me, @scylladb/seastar-maint any objections?

PR #3289, opened 2026-03-13
httpd: support scheduling group for TLS handshakes

ScyllaPiotr:
Add `set_post_connection_setup_sched_group()` on http_server, allowing callers to choose the scheduling group used after connection setup completes.

Connection setup, including TLS handshakes, runs in the scheduling group that started the listener. This lets callers run HTTPS listen/accept/setup in a dedicated scheduling group while switching accepted connections back to another scheduling group before request processing.

TLS listeners now perform the handshake eagerly through `connection::prepare()`, before request processing starts.

A test verifies that a TLS connection handler runs in the configured post-setup scheduling group.

### DIFF
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include <functional>
 #include <limits>
 #include <cctype>
 #include <vector>
@@ -37,6 +38,7 @@
 #include <seastar/http/routes.hh>
 #include <seastar/net/tls.hh>
 #include <seastar/core/shared_ptr.hh>
+#include <seastar/core/scheduling.hh>
 
 namespace seastar {
 
@@ -100,6 +102,7 @@ public:
     ~connection();
     void on_new_connection();
 
+    future<> prepare();
     future<> process();
     void shutdown();
     future<> read();
@@ -136,6 +139,7 @@ class http_server {
     bool _generate_date_header = true;
     gate _task_gate;
     std::optional<net::keepalive_params> _keepalive_params;
+    std::optional<scheduling_group> _request_scheduling_group;
 public:
     routes _routes;
     using connection = seastar::httpd::connection;
@@ -199,6 +203,15 @@ public:
     /// When set to false the periodic date-update timer is also stopped.
     void set_generate_date_header(bool b);
 
+    /// Sets the scheduling group used for request processing.
+    ///
+    /// Connection setup (including TLS handshake when enabled) runs in the
+    /// scheduling group of the accept loop. After setup completes, the
+    /// connection switches to the configured group before processing requests.
+    /// Without this setting, request processing continues in the accept loop
+    /// scheduling group.
+    void set_request_scheduling_group(scheduling_group sg);
+
     future<> listen(socket_address addr, server_credentials_ptr credentials);
     future<> listen(socket_address addr, listen_options lo, server_credentials_ptr credentials);
     future<> listen(socket_address addr, listen_options lo);
@@ -219,6 +232,7 @@ public:
     static sstring http_date();
 private:
     future<> do_accept_one(int which, bool with_tls);
+    future<> do_process_connection(connected_socket conn_fd, socket_address remote_address, bool tls);
     boost::intrusive::list<connection> _connections;
     friend class seastar::httpd::connection;
     friend class http_server_tester;
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -47,6 +47,7 @@
 #include <seastar/util/short_streams.hh>
 #include <seastar/util/log.hh>
 #include <seastar/util/string_utils.hh>
+#include <seastar/coroutine/switch_to.hh>
 
 
 using namespace std::chrono_literals;
@@ -306,6 +307,13 @@ future<> connection::process() {
         });
     });
 }
+
+future<> connection::prepare() {
+    if (_tls) {
+        co_await tls::get_protocol_version(_fd);
+    }
+}
+
 void connection::shutdown() {
     _fd.shutdown_input();
     _fd.shutdown_output();
@@ -401,6 +409,10 @@ void http_server::set_generate_date_header(bool b) {
     }
 }
 
+void http_server::set_request_scheduling_group(scheduling_group sg) {
+    _request_scheduling_group = sg;
+}
+
 future<> http_server::listen(socket_address addr, listen_options lo,
             server_credentials_ptr listener_credentials) {
     if (listener_credentials) {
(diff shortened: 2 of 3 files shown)


include/seastar/http/httpd.hh
```diff
@@ -134,6 +138,11 @@ class http_server {
     bool _content_streaming = false;
     gate _task_gate;
     std::optional<net::keepalive_params> _keepalive_params;
+    // Callback returning the scheduling group for new HTTPS connections.
+    // Used by Alternator to run CPU-intensive TLS handshakes in a
```
avikivity: Seastar knows nothing about alternator.

avikivity:
The comments and changelog have to be phrased in a neutral way. Any change to Seastar has to be reasonable for all its users.

src/http/httpd.cc
```diff
@@ -410,15 +420,38 @@ future<> http_server::stop() {
     return tasks_done;
 }
 
-// FIXME: This could return void
+// Accept loop for a single listener. For TLS listeners with a
+// new-connection scheduling group callback set, the loop switches to
+// the callback's scheduling group so that TLS handshakes (triggered
+// lazily on first I/O inside conn->process()) run in that group.
+future<> http_server::accept_loop(int which, bool tls) {
+    if (tls && _new_connection_scheduling_group) {
```
avikivity: Please separate the coroutine rewrite from the new functionality.

src/http/httpd.cc
```diff
@@ -410,15 +420,38 @@ future<> http_server::stop() {
...
+            // We expect a ECONNABORTED when http_server::stop is called,
+            // no point in warning about that.
+            if (e.code().value() != ECONNABORTED) {
+                hlogger.error("accept failed: {}", e);
+            }
+        } catch (...) {
+            hlogger.error("accept failed: {}", std::current_exception());
+        }
+        if (tls && _new_connection_scheduling_group) {
+            co_await coroutine::switch_to(_new_connection_scheduling_group());
```
avikivity: Why this?

include/seastar/http/httpd.hh
```diff
@@ -182,6 +191,20 @@ public:
...
+    /// scheduling group returned by the callback before accepting
+    /// connections. This ensures that TLS handshakes (which are
+    /// CPU-intensive) run in the designated scheduling group rather than
+    /// the caller's scheduling group.
+    ///
+    /// The callback is invoked each time the accept loop resumes after
+    /// accepting a connection, allowing dynamic scheduling group changes.
+    ///
+    /// Has no effect on plain HTTP listeners.
+    void set_new_connection_scheduling_group(std::function<scheduling_group()> fn);
```
avikivity: Why a function and not std::optional\<scheduling_group>?

nyh:
> The comments and changelog have to be phrased in a neutral way. Any change to Seastar has to be reasonable for all its users.

I agree 100%. Moreover, any reference to JIRA repositories which are not readable to the public should be removed, as they are offensive to non-Scylla developers who are not allowed to access them.

include/seastar/http/httpd.hh
ScyllaPiotr: Done. All Alternator/ScyllaDB/JIRA references removed from comments and commit messages.

src/http/httpd.cc
ScyllaPiotr: Done. Split into two commits:

1. `httpd: rewrite do_accepts/do_accept_one as coroutines` — pure mechanical coroutine rewrite, no behavioral changes.
2. `httpd: add scheduling group support for TLS accept loop` — the new functionality only.

src/http/httpd.cc
ScyllaPiotr: The scheduling group *is* preserved across `co_await` points — Seastar stores it in the coroutine's `promise_type::_sg` field and uses it to reschedule the task into the correct queue on resumption (proven by `coroutines_test.cc:239-264`). So the second `switch_to` is not needed for restoration.

It is there to **re-evaluate** the callback, because the scheduling group can change dynamically at runtime. This mirrors the pattern in `generic_server::do_accepts()` (`transport/generic_server.cc:365-388`), which calls `get_scheduling_group_for_new_connection()` both before the loop and after each accept. In the CQL server, that virtual method performs a dynamic QoS lookup — it checks whether a service level exists at call time and returns its scheduling group, falling back to `default_scheduling_group()` (`transport/server.hh:378-383`).

Added a clarifying comment in the code.

include/seastar/http/httpd.hh
ScyllaPiotr: The equivalent CQL mechanism uses a virtual method (`generic_server::get_scheduling_group_for_new_connection()`) that performs a dynamic QoS lookup at call time — checking whether a service level exists and returning its scheduling group, or falling back to `default_scheduling_group()` (`transport/server.hh:378-383`). Since `http_server` isn't designed for inheritance like `generic_server`, `std::function<scheduling_group()>` is the equivalent of the virtual method pattern.

ScyllaPiotr:
Agreed, thanks for the feedback. Updated: all comments, commit messages, and references are now neutral — no mentions of Alternator, ScyllaDB, or private JIRA tickets. The SCYLLADB-884 ref has been removed.

ScyllaPiotr:
FYI it's all vibe-coded.

mykaul:
> FYI it's all vibe-coded.

It's lacking tests, shouldn't we have some?

ScyllaPiotr:
> > FYI it's all vibe-coded.
> 
> It's lacking tests, shouldn't we have some?

Added a test as a separate commit. Thanks Claude!

include/seastar/http/httpd.hh
avikivity: Note, it's okay to give it as an example. But it's for all users.

include/seastar/http/httpd.hh
avikivity: I don't see the relationship to TLS. For TLS, you want to move out to some designated group for the handshake, then switch back, no?

Unless the philosophy is different: you spawn in the TLS scheduling group, then switch?

xemul:
Silly question, but why only for TLS loop? Caller knows whether or not it configures TLS or non-TLS server, so it can set or not set the sched-group switching callback

ScyllaPiotr:
> Silly question, but why only for TLS loop? Caller knows whether or not it configures TLS or non-TLS server, so it can set or not set the sched-group switching callback

In the current (updated) version, the scheduling group is changed specifically for TLS negotiation.

include/seastar/http/httpd.hh
ScyllaPiotr: Your remark was correct. I have fixed the issue.

avikivity:
Please rebase and split into two pull requests - I'm happy to commit the coroutinization immediately but will need to look at the second patch again.

ScyllaPiotr:
> Please rebase and split into two pull requests - I'm happy to commit the coroutinization immediately but will need to look at the second patch again.

Please see https://github.com/scylladb/seastar/pull/3359/

mykaul:
@ScyllaPiotr please resolve conflicts.

include/seastar/http/httpd.hh
avikivity: I still don't see why we need a function.

btw, perhaps it would be simpler to give a scheduling group to run the entire accept sequence in.

ScyllaPiotr: I've addressed both points:

1. **Replaced `std::function<scheduling_group()>` with a plain `scheduling_group` value** — the API is now `set_new_tls_connection_scheduling_group(scheduling_group sg)`.

2. **The accept loop for TLS listeners now runs entirely in the designated scheduling group** via `with_scheduling_group()`. This eliminates one `switch_to` per connection (previously we switched `default→tls→default`, now just `tls→default`).

The TLS handshake is now always performed eagerly (via `tls::get_protocol_version()`) for all TLS connections, regardless of whether a custom scheduling group is configured. This is needed so we know when the handshake is done and can switch back to `default_scheduling_group()` for request processing.

include/seastar/http/httpd.hh
```diff
@@ -229,6 +250,9 @@ public:
     static std::vector<server_socket>& listeners(http_server& server) {
         return server._listeners;
     }
+    static void set_on_tls_handshake_sg_switch(http_server& server, std::function<void()> fn) {
```
xemul: Patch comment doesn't describe this new method. Neither it has the documentation. Is it purely for testing?

ScyllaPiotr: Yes, it's purely a test hook — a setter for `_on_tls_handshake_sg_switch`, which is commented right next to the member variable. The test uses it to verify that `current_scheduling_group()` matches the configured scheduling group right before the handshake.

ScyllaPiotr:
@xemul please re-review.

xemul:
I'd like to reconsider the approach, as the newly introduced method looks like a random patch against current HTTPD processing.

With the PR the contract is pretty convoluted. One starts httpd in some scheduling groups, then the server temporarily switches the group into a configured one to do "something" (from the PR title this something == TLS handshake, but it really also includes accept()-s and a bit more), and then switch back to process the connection. And somehow this only applies to TLS.

I suggest to change the contract. Add the `set_connection_sched_group()` that does one thing -- makes httpd switch the scheduling group into the given one to do the connectionn processing. Period. Internally, the http_server::do_accept_one() would do

```c++
    auto conn = std::make_unique<connection>(...);
    (void)with_gate(_task_gate, [conn] {
        co_await conn->prepare(); // does TLS handshake if conn->_tls is set
        co_await coroutine::switch_to(_process_sg);
        co_await conn->process();
    });
```

include/seastar/http/httpd.hh
xemul: It's a pity to have it in generic code with public method and test-only always-false if in the happy path. Not that I have some better suggestion how to implement it

ScyllaPiotr:
@xemul thanks, that was good idea. I followed it. See if you like it more now.

include/seastar/http/httpd.hh
```diff
@@ -199,6 +203,14 @@ public:
     /// When set to false the periodic date-update timer is also stopped.
     void set_generate_date_header(bool b);
 
+    /// Sets scheduling group to switch to after connection setup.
+    ///
+    /// Connection preparation, including TLS handshake when enabled, runs in
+    /// the scheduling group used by the accept loop. After preparation, the
+    /// connection switches to this scheduling group. Without this setting,
+    /// processing continues in the accept loop group.
+    void set_post_connection_setup_sched_group(scheduling_group sg);
```
xemul: The name is quite clumsy. The rest (well, most of) of seastar API tries not to abbreviate scheduling_group to sched_group. Also the "post_connection_setup" part assumes that there's some setup, while it's purely internal thing.

How about set_request_scheduling_group() ?

ScyllaPiotr: I didn't like it and opted for sth IMO more explicit, but I'll follow your suggestion.

xemul (approved):
I'll wait for Avi/Nadav review for few days and will merge if no objections

PR #3290, opened 2026-03-16
perf_tests: add --no-perf-counters option

travisdowns:
Hardware perf counters (inst/cycles) require perf_event_open which may
be unavailable in containers or restricted environments. Enable them by
default but allow disabling via --no-perf-counters.

Reasons to disable perf counters:

- They add measurable overhead to start/stop timing calls, and for some
  benchmarks it isn't possible to structure the test to amortize this
  cost, so time-only measurement is preferred.
- Some environments don't support perf_event_open, but you may want
  apples-to-apples comparisons with environments that do, so disabling
  in both ensures consistent measurement methodology.
- Active perf events interfere with collecting perf data from outside
  the process, e.g., using perf record to profile a benchmark.

Add `linux_perf_event::always_zero()` factory method that returns a no-op
event which reads as zero, allowing disabled counters to be handled
without conditional logic throughout the measurement code.

Defer construction of the `time_measurement` object until after command
line parsing so the perf counter state is known at construction time.

### DIFF
--- a/include/seastar/testing/linux_perf_event.hh
+++ b/include/seastar/testing/linux_perf_event.hh
@@ -35,6 +35,7 @@ struct perf_event_attr; // from <linux/perf_event.h>
 
 class linux_perf_event {
     int _fd = -1;
+    linux_perf_event() = default;
 public:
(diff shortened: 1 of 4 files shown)


travisdowns:
@scylladb/seastar-maint please take a look.

PR #3293, opened 2026-03-17
when_all: reduce temporary object creation-deletion on synchronous execution path in when_all_state::wait_all

Comrade88:
currently we wrap incoming parameter pack of futures into tuple, then move each future::get result several times before placing into result tuple.
propose to pass original parameter pack by reference, extract xvalue values from future by reference and create final tuple via single move from each original ready future.

### DIFF
--- a/include/seastar/core/when_all.hh
+++ b/include/seastar/core/when_all.hh
@@ -50,8 +50,8 @@ struct identity_futures_tuple {
         p.set_value(std::move(futures));
     }
 
-    static future_type make_ready_future(std::tuple<Futures...> futures) noexcept {
-        return seastar::make_ready_future<std::tuple<Futures...>>(std::move(futures));
+    static future_type make_ready_future(Futures&&... futures) noexcept {
+        return seastar::make_ready_future<std::tuple<Futures...>>(std::move(futures)...);
     }
 
     static future_type current_exception_as_future() noexcept {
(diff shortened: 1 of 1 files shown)


Comrade88:
could you please take a look at this PR

include/seastar/core/when_all.hh
```diff
@@ -330,72 +330,45 @@ when_all(FutureIterator begin, FutureIterator end) noexcept {
...
-        auto prepare_result = [] (auto futures) noexcept {
-            auto fs = tuple_filter_by_type<internal::future_has_value>(std::move(futures));
-            return tuple_map(std::move(fs), [] (auto&& e) {
-                return e.get();
-            });
-        };
 
-        using tuple_futurizer = internal::tuple_to_future<decltype(prepare_result(std::move(futures)))>;
+    template<typename Future>
+    static constexpr bool is_blank_v = std::same_as<std::decay_t<Future>, future<>>;
```
avikivity, line 337: is_blank? odd name

include/seastar/core/when_all.hh
```diff
@@ -330,72 +330,45 @@ when_all(FutureIterator begin, FutureIterator end) noexcept {
...
 public:
-    using future_type = decltype(transform(std::declval<std::tuple<Futures...>>()));
+    using future_type = futurize_t<decltype(std::tuple_cat(
+        std::declval<std::conditional_t<is_blank_v<Futures>,
+            std::tuple<>, std::tuple<typename Futures::value_type>>>()...))>;
     using promise_type = typename future_type::promise_type;
 
     static void set_promise(promise_type& p, std::tuple<Futures...> tuple) {
-        transform(std::move(tuple)).forward_to(std::move(p));
+        std::apply(&make_ready_future, std::move(tuple)).forward_to(std::move(p));
```
avikivity, line 346: Hmm. Could change make_ready_future to return a tuple (not future\<tuple>) and name it make_value_tuple, then call promise::set_value with that tuple. Error handling will by yuckier. Perhaps better to leave as is.

avikivity:
Looks reasonable. The double loop isn't nice but I can't think of an alternative.

Can you show the difference in generated code?

PR #3294, opened 2026-03-17
perf_tests: exit non-zero when -t pattern matches no tests

travisdowns:
## Summary
- When `-t` is used but no tests match the given pattern(s), exit non-zero instead of silently succeeding
- Prints a warning to stderr listing the unmatched patterns, e.g.:

```
WARNING: no tests matched the given pattern(s): 'nonexistent_test'
```

### DIFF
--- a/tests/perf/perf_tests.cc
+++ b/tests/perf/perf_tests.cc
@@ -895,10 +895,20 @@ void run_all(const std::vector<std::string>& test_patterns, config& conf) {
         return false;
     };
     size_t max_name_column_length = 0;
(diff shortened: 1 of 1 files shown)


travisdowns:
@scylladb/seastar-maint check it out, I think it's the right behavior. Avoids a lot of silent mistakes. We run all benchmarks in a fast most as a test, and if someone uses a wrong pattern to restrict it, we want a failure (most common case is the pattern is correct but later becomes wrong as benchmark name changes).

PR #3295, opened 2026-03-18
Fair(er) allocation of outgoing tokens for IO classes

### LINKED ISSUE #1430: Cross-shard preemption is needed

xemul:
This PR re-implements the IO requests dispatch model (again), solving the cross-shard preemption problem.

Prior art a.k.a. current design and its problem(s).

The IO scheduler maintains two distinct mechanisms: class fairness and outgoing throttling. For fairness, classes are organized into a hierarchical priority queue (implemented in class fair_queue) sorted by normalized cost weighted by class shares. For throttling, a node-global shared_token_bucket rate-limits the total outgoing request flow to keep the disk from building up internal queues beyond io_properties.yaml configuration.

The problem happens when those two mechanisms interact. The token bucket operate as a simple FIFO: each shard grabbs tokens from a single shared tail pointer in the first-come-first-served manner, with no awareness of class shares or priorities. When classes on shards are competing, a high-priority class can easily preempt same-shard competitors and try to get dispatched, but when checking the global queue, the owning shard can come to the bucket after the preceding N−1 shards had consumed their full burst allowance (up to latency_goal worth of tokens) and yielded. In the worst case a high-priority request on that shard has to wait for all other shards to exhaust their bursts, resulting in dispatch latency proportional to latency_goal × nr_shards, regardless of the request's priority class and without the ability to preempt other shards' classes.

In this PR there's a perf test that reproduces this problem at its extreme, and suggest a change in the dispatch model to fix it.

The new approach makes all globally accumulated tokens be fairly distributed among active classes across all shards without the help of per-class priority queues. Once a class accumulates enough tokens for at least one request, it is dispatched immediately within the same poll cycle.
   
To compute each class's proportional share of generated tokens, the bucket maintains a global total_shares counter which is the sum of shares of all active (only active) classes across all shards. Since updating a shared atomic on every activation/deactivation would be expensive, each shard batches its share deltas in a local_publisher and flushes them once per poll, amortizing the cross-shard cost.
   
Token distribution is tree-based. The shard_aware_token_bucket converts elapsed time into tokens, similarly to how it was done before. Next each shard's consumer claims its share of tokens: (generated - last_seen) * shares / 
  total_shares. These tokens land in the dispatch_root and are then redistributed recursively: first among the root's direct children (top-level classes and groups) proportionally to their shares, then each group redistributes its portion among its child classes the same way.

```mermaid
   flowchart TD
       subgraph "Global (shared across shards)"
           SATB["shard_aware_token_bucket"]
       end
   
       subgraph "Per-shard poll_io_queue()"
           direction TB
   
           R["replenish_capacity(now)"] -->|"converts elapsed time<br/>to tokens (rate × Δt)"| SATB
   
           subgraph "Phase 1: Token distribution"
               REFILL["dispatch_root.refill():<br/>root.tokens += tokens × shard_shares / total_shares<br/>(root accumulates proportionally to active shard shares / node-wide active shares)"]
   
               subgraph "Proportional split"
                   ROOT_SPLIT["for each active child of root:<br/>child.tokens += root.tokens × child.shares / root.total_shares"]
                   GROUP_SPLIT["for each active child of group:<br/>child.tokens += group.tokens × child.shares / group.total_shares"]
               end
           end
   
           subgraph "Phase 2: Dispatch (for each class)"
               LOOP["for req in class.queue:"]
               TRY{"class.try_consume(req.cost)?"}
               DISP["req.dispatch()"]
               BREAK["break"]
           end
       end
   
       SATB --> REFILL
       REFILL --> ROOT_SPLIT
       ROOT_SPLIT -->|"groups recurse"| GROUP_SPLIT
       ROOT_SPLIT --> LOOP
       GROUP_SPLIT --> LOOP
       LOOP --> TRY
       TRY -->|yes| DISP
       DISP --> LOOP
       TRY -->|no| BREAK

 ```

Respectively, the whole fair-queue layer goes away (not yet removed in this PR). The change is very intrusive, but unfortunately there's no way to leap this chasm in two jumps.

The amount of accumulated tokens on a class is strictly limited. However, if a fatter request arrives, it wlil be split into smaller requests earlies based on maximum length-s values, and thus will not get stuck forver. A unit test is added.

Also, due to the priority-less new nature of dispatch cycle, if a group gets IO-throttled (per-group bandwidth limit) in the middle of a poll, nested classes can get into unfair starvation. To prevent this from happening, throttling a group only takes effect on next poll. This enlrges the potential single-poll IO burst, but on the long run these bursts are averaged.

Nonetheless, the approach shows some nice improvement. Below are two tables for the aforementioned perf-test that measures worst-case latency of high-priority request that has to compete with several low-prio but fat-requests shards. Rows (fg) -- concurrency of the high-priority class, columns (nq) -- total number of shards, cells -- average high-prio requests latency in ms. The test uses 100MB/s 10kIOPS io-properties.

Before:
| fg \ nq | 1 | 2 | 4 | 8 | 16 |
| --- | --- | --- | --- | --- | --- |
| 1 | 0.060 | 0.607 | 0.718 | 1.167 | 2.375 |
| 2 | 0.099 | 0.807 | 1.380 | 1.629 | 1.304 |
| 4 | 0.322 | 1.665 | 2.958 | 2.105 | 4.123 |
| 8 | 0.802 | 2.927 | 4.854 | 7.694 | 13.136 |

After:
| fg \ nq | 1 | 2 | 4 | 8 | 16 |
| --- | --- | --- | --- | --- | --- |
| 1 | 0.022 | 0.025 | 0.027 | 0.032 | 0.035 |
| 2 | 0.060 | 0.061 | 0.064 | 0.070 | 0.070 |
| 4 | 0.062 | 0.062 | 0.061 | 0.071 | 0.073 |
| 8 | 0.287 | 0.288 | 0.306 | 0.472 | 0.626 |

fixes #1430

### DIFF
--- a/include/seastar/util/shared_token_bucket.hh
+++ b/include/seastar/util/shared_token_bucket.hh
@@ -369,11 +369,17 @@ public:
         explicit tokens(T burst_limit) noexcept : _burst_limit(burst_limit) {}
 
         /*
-         * Add amount tokens to the pouch, saturating at burst_limit. The cap
-         * prevents unbounded accumulation while the owner is idle or stalled.
+         * Add amount tokens to the pouch, saturating at burst_limit.
+         * Returns any overflow that could not be absorbed.
          */
-        void refill(T amount) noexcept {
-            _available = std::min(_available + amount, _burst_limit);
+        T refill(T amount) noexcept {
+            auto sum = _available + amount;
+            if (sum > _burst_limit) {
+                _available = _burst_limit;
+                return sum - _burst_limit;
+            }
+            _available = sum;
+            return T(0);
         }
 
         /*
@@ -393,6 +399,7 @@ public:
         }
 
         T available() const noexcept { return _available; }
+        bool full() const noexcept { return _available >= _burst_limit; }
     };
 
     /*
--- a/include/seastar/util/shared_token_bucket.hh
+++ b/include/seastar/util/shared_token_bucket.hh
@@ -396,6 +396,46 @@ public:
     };
 
     /*
+     * Per-shard accumulator for pending contributions to _total_shares.
+     *
+     * A consumer's activate()/deactivate()/update_shares() no longer RMW the
+     * shared _total_shares atomic directly; instead they add a signed delta
+     * to a local_publisher's shard-local _pending counter (plain integer
+     * arithmetic, no cacheline contention).  Once per poll the owner calls
+     * flush(), which turns the whole pending delta into a single atomic RMW
+     * on the shared cacheline regardless of how many consumers flipped or
+     * how many times they flipped.
+     */
+    class local_publisher {
+        shard_aware_token_bucket& _bucket;
+        std::make_signed_t<T> _pending{0};
+
+    public:
+        explicit local_publisher(shard_aware_token_bucket& bucket) noexcept
+            : _bucket(bucket) {}
+
+        local_publisher(const local_publisher&) = delete;
+        local_publisher(local_publisher&&) = delete;
+
+        shard_aware_token_bucket& bucket() const noexcept { return _bucket; }
+
+        void apply(std::make_signed_t<T> delta) noexcept { _pending += delta; }
+
+        /* Commit accumulated delta to the shared _total_shares atomic. */
+        void flush() noexcept {
+            if (_pending == 0) {
+                return;
+            }
+            if (_pending > 0) {
+                _bucket._total_shares.fetch_add(T(_pending), std::memory_order_relaxed);
+            } else {
+                _bucket._total_shares.fetch_sub(T(-_pending), std::memory_order_relaxed);
+            }
+            _pending = 0;
+        }
+    };
+
+    /*
      * Per-class, shard-local share/generation bookkeeping. Not thread-safe on
      * its own; intended to be owned and accessed by a single shard.
      *
@@ -407,7 +447,7 @@ public:
      * compute them and feed the result into an externally-owned tokens pouch.
      */
     class consumer {
-        shard_aware_token_bucket& _bucket;
+        local_publisher& _publisher;
         T _shares;
         // Per-consumer watermark into _tokens_generated (see above).
         // accrued() computes gain as (generated - _last_seen) * _shares
@@ -417,11 +457,13 @@ public:
         T _last_seen;
         bool _active{false};
 
+        shard_aware_token_bucket& bucket() const noexcept { return _publisher.bucket(); }
+
     public:
-        consumer(shard_aware_token_bucket& bucket, T shares) noexcept
-            : _bucket(bucket)
+        consumer(local_publisher& publisher, T shares) noexcept
+            : _publisher(publisher)
             , _shares(shares)
-            , _last_seen(bucket._tokens_generated.load(std::memory_order_relaxed))
+            , _last_seen(publisher.bucket()._tokens_generated.load(std::memory_order_relaxed))
         {}
 
         consumer(const consumer&) = delete;
@@ -434,8 +476,9 @@ public:
         }
 
         /*
-         * Mark this class as actively competing for tokens. Contributes _shares
-         * to total_shares until deactivate() is called.
+         * Mark this class as actively competing for tokens.  The contribution
+         * to total_shares is accumulated into the publisher's pending delta
+         * and flushed to the shared atomic by the owner once per poll.
          *
          * On re-activation after idle, the class receives credit for at most
          * tau_tokens worth of global generation (the forgiveness window). This
@@ -445,21 +488,22 @@ public:
         void activate() noexcept {
             SEASTAR_ASSERT(!_active);
             _active = true;
-            auto gen = _bucket._tokens_generated.load(std::memory_order_relaxed);
-            auto tau = _bucket._tau_tokens;
+            auto& b = bucket();
+            auto gen = b._tokens_generated.load(std::memory_order_relaxed);
+            auto tau = b._tau_tokens;
             if (tau > 0 && gen >= tau) {
                 _last_seen = std::max(_last_seen, gen - tau);
             } else if (tau == 0) {
                 _last_seen = gen;
             }
-            _bucket._total_shares.fetch_add(_shares, std::memory_order_relaxed);
+            _publisher.apply(std::make_signed_t<T>(_shares));
         }
 
         /* Remove this class from active competition. */
         void deactivate() noexcept {
             SEASTAR_ASSERT(_active);
             _active = false;
-            _bucket._total_shares.fetch_sub(_shares, std::memory_order_relaxed);
+            _publisher.apply(-std::make_signed_t<T>(_shares));
         }
 
         bool is_active() const noexcept { return _active; }
@@ -470,31 +514,32 @@ public:
          * advance the internal watermark. The caller is expected to feed the
          * result into a tokens pouch via tokens::refill().
          *
+         * The caller must have flushed the shard's local_publisher at least
+         * once in this poll before calling accrued(), so that _total_shares
+         * reflects the full shard-local active set.
+         *
          * Fairness is preserved by the proportional gain formula; no capping
          * is applied here -- the pouch's burst_limit caps accumulation.
          */
         T accrued() noexcept {
-            auto gen = _bucket._tokens_generated.load(std::memory_order_relaxed);
+            auto& b = bucket();
+            auto gen = b._tokens_generated.load(std::memory_order_relaxed);
             auto delta = gen - _last_seen;
             if (delta == 0) {
                 return T(0);
             }
             _last_seen = gen;
-            auto total = _bucket._total_shares.load(std::memory_order_relaxed);
+            auto total = b._total_shares.load(std::memory_order_relaxed);
             if (total == 0) {
                 return T(0);
             }
             return T(double(delta) * _shares / total);
         }
 
-        /* Update shares; adjusts total_shares immediately if currently active. */
+        /* Update shares.  The difference is accumulated into the publisher. */
         void update_shares(T new_shares) noexcept {
             if (_active) {
-                if (new_shares >= _shares) {
-                    _bucket._total_shares.fetch_add(new_shares - _shares, std::memory_order_relaxed);
-                } else {
-                    _bucket._total_shares.fetch_sub(_shares - new_shares, std::memory_order_relaxed);
-                }
+                _publisher.apply(std::make_signed_t<T>(new_shares) - std::make_signed_t<T>(_shares));
             }
             _shares = new_shares;
         }
--- a/include/seastar/util/shared_token_bucket.hh
+++ b/include/seastar/util/shared_token_bucket.hh
@@ -214,5 +214,294 @@ public:
     }
 };
 
+/*
+ * A token bucket where each competing class accumulates its proportional share
+ * of tokens independently, based on its shares relative to the total.
+ *
+ * The global state holds a monotonically increasing `tokens_generated` counter
+ * that any shard can advance by calling replenish(). Each class has shard-local
+ * state and computes its token allocation lazily as:
+ *
+ *   gain = (tokens_generated - last_seen) * class_shares / total_shares
+ *
+ * This means a "stuck" shard that doesn't poll for a while simply catches up
+ * to its burst cap when it wakes up, without being starved by other shards.
+ * Conversely, a fast-polling shard cannot grab tokens that belong to other
+ * classes — each class's share is computed independently from the shared clock.
+ *
+ * Architecture:
+ *
+ *                     +-------------------------------+
+ *                     |   shard_aware_token_bucket    |
+ *                     |                               |
+ *                     |  _rate                        | <-- tokens per second
+ *                     |  _tokens_generated (atomic)   | <-- replenish() advances this
+ *                     |  _total_shares     (atomic)   | <-- sum of all active consumers
+ *                     |  _tau_tokens                  | <-- forgiveness window
+ *                     +---------------+---------------+
+ *                                     |
+ *              +----------------------+----------------------------+
+ *              |                                                   |
+ *    +---------v-----------------+                       +---------v--------+
+ *    | local_publisher (shard 0) |                       | local_publisher  |
+ *    |                           |                       |    (shard 1)     |
+ *    |  _pending                 | <-- accumulated       +--+----------+---+
+ *    |                           |     delta-shares         |          |
+ *    |  flush()                  | <-- commits           consumer  consumer
+ *    |                           |     _pending to
+ *    +--+-------------------+----+     _total_shares
+ *       |                   |          in one atomic RMW
+ *       |                   |
+ *  +----v-----------+  +----v-----------+
+ *  | consumer       |  | consumer       |
+ *  |                |  +----------------+
+ *  |                |
+ *  |  _shares       | <-- weight in the proportional split
+ *  |  _last_seen    | <-- snapshot of _tokens_generated at previous accrued()
+ *  |  accrued()     | <-- returns (generated - _last_seen) * _shares / total_shares
+ *  |  activate()    | <-- adds _shares to _total_shares via publisher
+ *  |  deactivate()  | <-- removes _shares from _total_shares via publisher
+ *  +----------------+
+ *
+ * Per-poll data flow (one shard):
+ *
+ *  1. replenish(now) -- any shard may advance _tokens_generated based on
+ *     elapsed time; concurrent calls are serialized via CAS.
+ *
+ *  2. local_publisher::flush() -- commits locally batched share deltas to
+ *     the global _total_shares in a single atomic fetch-and-add.
+ *
+ *  3. consumer::accrued() -- each consumer computes its token gain since the
+ *     last call as (generated - _last_seen) * _shares / total_shares, then
+ *     updates _last_seen. The result is a plain token value returned to the
+ *     caller (io_queue's dispatch_root) for redistribution.
+ *
+ *  4. activate/deactivate -- when a class becomes active or idle, its shares
+ *     are added to or removed from _pending in the local_publisher, to be
+ *     committed on the next flush().
+ */
+template <typename T, typename Period, typename Clock = std::chrono::steady_clock>
+requires std::is_nothrow_copy_constructible_v<T> && supports_wrapping_arithmetics<T>
+class shard_aware_token_bucket {
+    using rate_resolution = std::chrono::duration<double, Period>;
+
+    T _rate;
+    const T _burst_limit;
+    const T _threshold;
+    const T _tau_tokens;
+    // Monotonically increasing counter of tokens produced globally.
+    // replenish() converts elapsed wall-clock time into tokens and
+    // advances this via atomic fetch-add. Consumers read it in
+    // accrued() and compute their per-class share of the delta
+    // relative to their own _last_seen snapshot (see consumer below).
+    std::atomic<T> _tokens_generated{0};
+    std::atomic<typename Clock::time_point> _last_replenished;
+    std::atomic<T> _total_shares{0};
+
+    static constexpr rate_resolution max_delta = std::chrono::duration_cast<rate_resolution>(std::chrono::hours(1));
+
+    T accumulated_in(typename Clock::duration delta) const noexcept {
+        auto d = std::min(std::chrono::duration_cast<rate_resolution>(delta), max_delta);
+        return T(std::round(_rate * d.count()));
+    }
+
+public:
+    static constexpr T max_rate = std::numeric_limits<T>::max() / 2 / max_delta.count();
+
+    template <typename Rep, typename Per>
+    static auto rate_cast(const std::chrono::duration<Rep, Per> d) noexcept {
+        return std::chrono::duration_cast<rate_resolution>(d);
+    }
+
+    rate_resolution duration_for(T tokens) const noexcept {
+        return rate_resolution(double(tokens) / _rate);
+    }
+
+    T rate() const noexcept { return _rate; }
+    T limit() const noexcept { return _burst_limit; }
+    T threshold() const noexcept { return _threshold; }
+    T total_shares() const noexcept { return _total_shares.load(std::memory_order_relaxed); }
+    T tokens_generated() const noexcept { return _tokens_generated.load(std::memory_order_relaxed); }
+
+    T tau_tokens() const noexcept { return _tau_tokens; }
+
+    shard_aware_token_bucket(T rate, T burst_limit, T threshold,
+                              bool add_replenish_offset = true,
+                              typename Clock::duration tau = {}) noexcept
+        : _rate(std::min(rate, max_rate))
+        , _burst_limit(burst_limit)
+        , _threshold(std::clamp(threshold, T(1), burst_limit))
+        , _tau_tokens(accumulated_in(tau))
+        , _last_replenished(Clock::now() - std::chrono::hours(add_replenish_offset ? 24 : 0))
+    {}
+
+    /*
+     * Advance the global token counter based on elapsed time.
+     * Only one shard succeeds per threshold window via CAS on _last_replenished.
+     */
+    void replenish(typename Clock::time_point now) noexcept {
+        auto ts = _last_replenished.load(std::memory_order_relaxed);
+        if (now <= ts) {
+            return;
+        }
+        auto extra = accumulated_in(now - ts);
+        if (extra >= _threshold) {
+            if (!_last_replenished.compare_exchange_strong(ts, now)) {
+                return;
+            }
+            _tokens_generated.fetch_add(extra, std::memory_order_relaxed);
+        }
+    }
+
+    /*
+     * A shard-local token pouch. Carries a number of currently-available
+     * tokens and enforces a burst cap on top-ups. Has no ties to the bucket
+     * or to any particular token source: it is fed via refill(amount) from
+     * whatever source the owner chooses (typically either consumer::accrued()
+     * for a direct draw from the global bucket, or a raw amount passed down
+     * from a parent entity redistributing its own drained tokens).
+     */
+    class tokens {
+        T _available{0};
+        T _burst_limit;
+
+    public:
+        explicit tokens(T burst_limit) noexcept : _burst_limit(burst_limit) {}
+
+        /*
+         * Add amount tokens to the pouch, saturating at burst_limit. The cap
+         * prevents unbounded accumulation while the owner is idle or stalled.
+         */
+        void refill(T amount) noexcept {
+            _available = std::min(_available + amount, _burst_limit);
+        }
+
+        /*
+         * Consume cost tokens. Returns true and deducts if enough are available.
+         */
+        bool try_consume(T cost) noexcept {
+            if (_available >= cost) {
+                _available -= cost;
+                return true;
+            }
+            return false;
+        }
+
+        /* Empty the pouch and return what was in it (for redistribution). */
+        T drain() noexcept {
+            return std::exchange(_available, T(0));
+        }
+
+        T available() const noexcept { return _available; }
+    };
+
+    /*
+     * Per-class, shard-local share/generation bookkeeping. Not thread-safe on
+     * its own; intended to be owned and accessed by a single shard.
+     *
+     * A class starts inactive and does not contribute to total_shares until
+     * activate() is called. Call activate() when the class first has pending
+     * requests, and deactivate() when it becomes idle again.
+     *
+     * Tokens generated for this consumer are not kept here; use accrued() to
+     * compute them and feed the result into an externally-owned tokens pouch.
+     */
+    class consumer {
+        shard_aware_token_bucket& _bucket;
+        T _shares;
+        // Per-consumer watermark into _tokens_generated (see above).
+        // accrued() computes gain as (generated - _last_seen) * _shares
+        // / total_shares, then advances _last_seen to generated. On
+        // re-activation after idle, _last_seen is clamped so the class
+        // receives at most tau_tokens worth of catch-up credit.
+        T _last_seen;
+        bool _active{false};
+
+    public:
+        consumer(shard_aware_token_bucket& bucket, T shares) noexcept
+            : _bucket(bucket)
+            , _shares(shares)
+            , _last_seen(bucket._tokens_generated.load(std::memory_order_relaxed))
+        {}
+
+        consumer(const consumer&) = delete;
+        consumer(consumer&&) = delete;
+
+        ~consumer() {
+            if (_active) {
+                deactivate();
+            }
+        }
+
+        /*
+         * Mark this class as actively competing for tokens. Contributes _shares
+         * to total_shares until deactivate() is called.
+         *
+         * On re-activation after idle, the class receives credit for at most
+         * tau_tokens worth of global generation (the forgiveness window). This
+         * allows briefly-idle classes to burst on return without granting
+         * unbounded catch-up to long-idle ones.
+         */
+        void activate() noexcept {
+            SEASTAR_ASSERT(!_active);
+            _active = true;
+            auto gen = _bucket._tokens_generated.load(std::memory_order_relaxed);
+            auto tau = _bucket._tau_tokens;
+            if (tau > 0 && gen >= tau) {
+                _last_seen = std::max(_last_seen, gen - tau);
+            } else if (tau == 0) {
+                _last_seen = gen;
+            }
+            _bucket._total_shares.fetch_add(_shares, std::memory_order_relaxed);
+        }
+
+        /* Remove this class from active competition. */
+        void deactivate() noexcept {
+            SEASTAR_ASSERT(_active);
+            _active = false;
+            _bucket._total_shares.fetch_sub(_shares, std::memory_order_relaxed);
+        }
+
+        bool is_active() const noexcept { return _active; }
+
+        /*
+         * Return the number of tokens proportionally produced for this consumer
+         * since the previous call (or since construction / reactivation), and
+         * advance the internal watermark. The caller is expected to feed the
+         * result into a tokens pouch via tokens::refill().
+         *
+         * Fairness is preserved by the proportional gain formula; no capping
+         * is applied here -- the pouch's burst_limit caps accumulation.
+         */
+        T accrued() noexcept {
+            auto gen = _bucket._tokens_generated.load(std::memory_order_relaxed);
+            auto delta = gen - _last_seen;
+            if (delta == 0) {
+                return T(0);
+            }
+            _last_seen = gen;
+            auto total = _bucket._total_shares.load(std::memory_order_relaxed);
+            if (total == 0) {
+                return T(0);
+            }
+            return T(double(delta) * _shares / total);
+        }
+
+        /* Update shares; adjusts total_shares immediately if currently active. */
+        void update_shares(T new_shares) noexcept {
+            if (_active) {
+                if (new_shares >= _shares) {
+                    _bucket._total_shares.fetch_add(new_shares - _shares, std::memory_order_relaxed);
+                } else {
+                    _bucket._total_shares.fetch_sub(_shares - new_shares, std::memory_order_relaxed);
+                }
+            }
+            _shares = new_shares;
+        }
+
+        T shares() const noexcept { return _shares; }
+    };
+};
+
 } // internal namespace
 } // seastar namespace
--- a/tests/perf/io_queue_latency.cc
+++ b/tests/perf/io_queue_latency.cc
@@ -1,0 +1,155 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2022 ScyllaDB
+ */
+
+#include <seastar/core/app-template.hh>
+#include <seastar/core/thread.hh>
+#include <seastar/core/reactor.hh>
+#include <seastar/core/sleep.hh>
+#include <seastar/core/when_all.hh>
+#include <seastar/core/loop.hh>
+#include <seastar/core/io_queue.hh>
+#include <seastar/core/internal/io_request.hh>
+#include <seastar/core/internal/io_sink.hh>
+
+using namespace seastar;
+using namespace std::chrono;
+
+static seastar::logger testlog("testlog");
+
+struct io_queue_for_tests {
+    struct io_group_for_tests {
+        io_group_ptr group;
+
+        io_group_for_tests(const io_queue::config& cfg, unsigned nr_queues)
+            : group(std::make_shared<io_group>(cfg, nr_queues))
+        {
+        }
+    };
+
+    internal::io_sink sink;
+    io_queue queue;
+
+    io_queue_for_tests(io_group_for_tests& group)
+        : queue(group.group, sink)
+    {
+    }
+
+    void poll_once() {
+        queue.poll_io_queue();
+        sink.drain([] (const internal::io_request& rq, io_completion* desc) -> bool {
+            const auto& op = rq.as<internal::io_request::operation::write>();
+            desc->complete_with(op.size);
+            return true;
+        });
+    }
+
+    future<> enqueue(scheduling_group sg, size_t req_size) {
+        return queue.queue_request(internal::priority_class(sg),
+            internal::io_direction_and_length(internal::io_direction_and_length::read_idx, req_size),
+            internal::io_request::make_read(0, 0, nullptr, req_size, false),
+            nullptr, {}).discard_result();
+    }
+
+    future<> enqueue(scheduling_group sg, unsigned nr, size_t req_size) {
+        return parallel_for_each(std::views::iota(0u, nr), [this, sg, req_size] (auto i) -> future<> {
+            return enqueue(sg, req_size);
+        });
+    }
+};
+
+int main(int ac, char** av) {
+    app_template at;
+    namespace bpo = boost::program_options;
+    at.add_options()
+            ("iterations", bpo::value<unsigned>()->default_value(7), "Number of iterations")
+            ("nr-queues", bpo::value<unsigned>()->default_value(4), "Number of queues to simulate")
+            ("iops", bpo::value<unsigned>()->default_value(10000), "IOPS")
+            ("bandwidth-mb", bpo::value<unsigned>()->default_value(100), "Bandwidth in MB/s")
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by xemul** [comment]
Upd:
- fixed a bug in io-group creation (it took smp::count as number-of-queues, not nr_queues)
- added fg-concurrency to exagerrate the problem

The results now look like below

| nr_queues | fg_concurrency | iter-1 delay | steady-state delay |
   |-----------|----------------|--------------|-------------------|
   | 1         | 1              | 0.31 ms      | ~0.003 ms         |
   | 1         | 2              | 0.55 ms      | ~0.003 ms         |
   | 1         | 4              | 0.55 ms      | ~0.27 ms          |
   | 1         | 16             | 1.92 ms      | ~1.92 ms          |
   | 4         | 1              | 1.66 ms      | ~0.54 ms          |
   | 4         | 2              | 3.15 ms      | ~1.30 ms          |
   | 4         | 4              | 4.22 ms      | ~3.0 ms           |
   | 4         | 16             | 10.27 ms     | ~9.1 ms           |

**Comment by xemul** [comment]
upd:
- moved the shard-aware-token-bucket hunk from io-queue patch to shard-aware-token-bucket patch
- fixed s.a.t.b. test to conform "forgiveness" changes made in this hunk and that were lost

**Comment by xemul** [comment]
upd:
- added unit test to validate that oversized request is properly split and doesn't get stuck in s.a.t.b. maximum tokens accumulation
- added unit test for end-to-end math validation -- an io-queue is configured with io-properties yaml, that unbound workflow starts, the test validates that bw/bw_max + iops/iop_max <= 1

**Comment by xemul** [comment]
upd:
- revsed the poll_io_queue() methanics wrt throttling. Briefly: now one poll cycle dispatches all requests that accumulated tokens allow, ignoring if the class/group got throttled in the middle of the loop. The throttling only takes effect on _next_ polling

**Comment by xemul** [comment]
upd:
- keep track of sum of sub-classes' shares in class group for simpler redistribution of tokens
- revisited and unified plug/unplug vs activate/deactivate logic

**Comment by xemul** [comment]
upd:
- restored the "consumption" metrics

**Comment by xemul** [comment]
upd:
- batch shares update on shared token bucket by shards

**Comment by xemul** [comment]
upd:
- revisited (simplified) latch-plug for individual classes

**Comment by xemul** [comment]
Some nice results from @fee-mendes testing

Here's what he does:

> I can get quite decent P99's under ideal conditions of temperature and pressure, they only last so long compaction or tablet load balancing streaming don't occur. ... I am stressing ScyllaDB using the worst case scenario of a cache unfriendly workload, thus every read/write touches the disk.

And here's the graphana plots.

## Disk consumption and flow-ratio (before-after)

image
image

Disk is ~60% loaded, reactor is polling IO in timely manner. So far so good

## IO stressing workloads (compaction+memtable, before-after)

### Compaction
image
image

### Memtable
image
image

Both are active in both runs, no notable difference here.

## Latency-sensitive workloads (commitlog+sl:default, before-after

### Commitlog
image
image

### User queries (sl:default)
image
image


Here we see that throughput/iops is the same before and after, the latency is notably better, a well as starvation times

**Comment by xemul** [comment]
Two major outcomes from the above plots
* It works :relaxed:
* Latency sensitive classes seem not to contend on each other, aligned with the perf-test demonstration (table in the PR description)

**Comment by xemul** [comment]
For the record -- right now this PR is in DRAFT mode for two reasons
* the code flow in new poll_io_queues() is quite messy
* the overall split into patches is very hard to review :( hard to expect much better split, taking into account the amount of changes, but worth trying anyway

**Comment by xemul** [comment]
Some experiments with the io-tester

### Demo 1 — Cross-shard throughput allocation

**Setup:** 2 shards, 2 classes with different shares running sequential reads (128 kB requests, parallelism 16). Both classes saturate the disk (`read_bandwidth: 500 MB/s, read_iops: 100K`).

| Class | Shares | Master throughput | Branch throughput |
|-------|--------|-------------------|-------------------|
| `compaction` (shard 0) | 1000 | 237 MB/s | **401 MB/s** |
| `streaming` (shard 1) | 200 | 237 MB/s | **80 MB/s** |
| **Ratio** | **5 : 1** | **1 : 1** ❌ | **5 : 1** ✅ |

On master, each shard's `io_queue` allocates tokens independently — since each class is alone on its shard, both get equal disk bandwidth regardless of shares. On branch, the `shard_aware_token_bucket` coordinates cross-shard allocation so the 5:1 shares ratio is actually enforced.

<details>
<summary>Config and CLI</summary>

`workload.yaml`:
```yaml
- name: compaction
  type: seqread
  shards: [0]
  shard_info:
    parallelism: 16
    reqsize: 128kB
    shares: 1000

- name: streaming
  type: seqread
  shards: [1]
  shard_info:
    parallelism: 16
    reqsize: 128kB
    shares: 200
```

`io-properties.yaml`:
```yaml
disks:
  - mountpoint: /dev/null
    read_bandwidth: 500M
    read_iops: 100000
    write_bandwidth: 400M
    write_iops: 80000
```

```
io_tester --smp 2 --storage /dev/null --conf workload.yaml --duration 30 --io-properties-file io-properties.yaml
```
</details>

---

### Demo 2 — Foreground latency isolation under background saturation

**Setup:** 4 shards. Shards 0–2 run background sequential reads (128 kB, parallelism 16, shares 200) that fully saturate the disk (`read_bandwidth: 100 MB/s, read_iops: 10K`). Shard 3 runs a high-priority foreground random-read workload (4 kB, shares 1000, fixed rate of 100 RPS).

**Foreground (shard 3) latencies:**

| Metric | Master | Branch | Improvement |
|--------|--------|--------|-------------|
| Average | 1,346 µs | **11 µs** | **122×** |
| p50 | 1,392 µs | 10 µs | 139× |
| p95 | 2,393 µs | 17 µs | 141× |
| p99 | 2,790 µs | 29 µs | 96× |
| p99.9 | 3,140 µs | 41 µs | 77× |

The foreground request rate is identical (~398 IOPS) in both cases — only the latency changes. On master, each background shard independently depletes the shared token pool, so foreground requests must wait for the pool to replenish. On branch, background throughput is rate-limited by shares cross-shard, leaving tokens consistently available for foreground → near-zero queuing delay.

Background throughput is unaffected: all 3 background shards sustain ~29 MB/s each on both master and branch.

<details>
<summary>Config and CLI</summary>

`workload.yaml`:
```yaml
- name: background
  type: seqread
  shards: [0, 1, 2]
  shard_info:
    parallelism: 16
    reqsize: 128kB
    shares: 200

- name: foreground
  type: randread
  shards: [3]
  shard_info:
    parallelism: 4
    reqsize: 4kB
    shares: 1000
    rps: 100
```

`io-properties.yaml`:
```yaml
disks:
  - mountpoint: /dev/null
    read_bandwidth: 100M
    read_iops: 10000
    write_bandwidth: 80M
    write_iops: 8000
```

```
io_tester --smp 4 --storage /dev/null --conf workload.yaml --duration 30 --io-properties-file io-properties.yaml
```
</details>

xemul:
upd:
- rebased (had conflicts)
- fixed io-tester and ran [some measurements](https://github.com/scylladb/seastar/pull/3295#issuecomment-4406126520) with it
- revived "activations" metrics

xemul:
upd:
- simiplified throttle/unthrottle logic in bandwith_limiter helper class. with new io_queue::priority_entry it becomes possible to make the plug/unplug code flow much saner
- tuned the core patch a bit to provide easier to review diff

xemul:
upd:
- detached more changes from core patch into the shard-local class-vs-group hierarchy changes one
- added comment about plug/unplug idempotency
- fixed a bug when a class being unregistered leaves itself in the parent children list

xemul:
upd:
- reorder vectors of class datas and class group datas on io-queue to keep proper destruction order

xemul:
OK, I think the PR is ready. @avikivity , please review. The change is huge, unfortunately.

Summary:
* Original bug is #1430 
* [Test that demonstrates the problem](https://github.com/scylladb/seastar/pull/3295/changes/8fe7286b4108b7c423389335d739e70d4ec96f5d)
* [Core patch and description](https://github.com/scylladb/seastar/pull/3295/changes/83f6ef3b174aa71182479fec77923e9af46fa1d9)
* [Real-life-ish metrics from @fee-mendes](https://github.com/scylladb/seastar/pull/3295#issuecomment-4341739225)
* [IO-tester workloads validation](https://github.com/scylladb/seastar/pull/3295#issuecomment-4406126520)

xemul:
upd:
- refined the token re-distribution mechanics. Old logic -- all root-level classes and groups grabbed shared tokens themselves, then groups re-distributed tokens to sub-classes. New logic -- the appears root consumer that grabs shard tokens, then re-distributes them amon its children, then children that are groups re-distribut the tokens below. Semantically it's the same (plus avoids a theoretical race), but the code-flow in poll_io_queue is now much more easy to follow

xemul:
@avikivity , a reminder ping. The change is huge, if there's anything else apart from existing tests and measurements that can help?

avikivity:
Let me review it for starters

avikivity:
One think I can already tell missing is a detailed description of the problem and the algorithm in doc/.

In today's AI world, at least you can be sure documentation isn't write-only, because AI will read all of it.

avikivity:
Also cc: @travisdowns for redpanda distribtution

tests/perf/io_queue_latency.cc
```diff
@@ -0,0 +1,155 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
```
avikivity, line 4: > perf: Add perftest that measures worst-case IO preemption latency

This is really excellent. It allows reproducing the problem reliably without depending on disks that are always different.

Please document the disk it models (zero-latency infinite-bandwidth, so completely trusting the I/O scheduler here).

In the future we can introduce more realistic models.

tests/perf/io_queue_latency.cc
```diff
@@ -0,0 +1,155 @@
...
+    app_template at;
+    namespace bpo = boost::program_options;
+    at.add_options()
+            ("iterations", bpo::value<unsigned>()->default_value(7), "Number of iterations")
+            ("nr_queues", bpo::value<unsigned>()->default_value(4), "Number of queues to simulate")
+            ("iops", bpo::value<unsigned>()->default_value(10000), "IOPS")
+            ("bandwidth_mb", bpo::value<unsigned>()->default_value(100), "Bandwidth in MB/s")
+            ("concurrency", bpo::value<unsigned>()->default_value(4), "Concurrency of competing low-prio group")
+            ("fg_concurrency", bpo::value<unsigned>()->default_value(1), "Concurrency of foreground group (burst size)")
+            ("poll_interval_us", bpo::value<unsigned>()->default_value(250), "Interval between queue polls")
```
avikivity: Please use traditional dash-separated words for the options.

include/seastar/util/shared_token_bucket.hh
```diff
@@ -214,5 +214,232 @@ public:
     }
 };
 
+/*
```
avikivity, line 217: > util: Add shard_aware_token_bucket

I could really use a graphic to get an understanding here.

avikivity:
I gave a shallow review, but didn't really attempt to go deep, as I'm sure if I build a mental model from the description I'll get it wrong. A diagram explaining what's what will go a long way to simplify here.

tests/perf/io_queue_latency.cc
xemul: This test doesn't model a disk, it models the concurrency between shards for tokens in shared token bucket. More specifically -- it deliberately alings shards into worst situation they can get

xemul:
upd:
- rephrased cover letter to describe the problem and proposed design (hopefully) better
- added a diagram to cover letter with the new deispatch mechanics
- updated doc/io-scheduler.md
- added more descriptive (with diagram) code comment to new shard_aware_token_bucket
- explained generated/last-seen tokens mechanics with code comment
- use dashes for perf test CLI options

include/seastar/util/shared_token_bucket.hh
xemul: Added one, please check

tests/perf/io_queue_latency.cc
xemul: Done

xemul:
```
=== Tick 0: Initial state ===

                     satb [generated=0, rate=1000/tick]
                      |                      |
              consumer(sh=1000)       consumer(sh=1200)
                      |                      |
                shard-0 root           shard-1 root
                [tokens=0]             [tokens=0]
                    |                    /       \
               G(sh=1000)         C(sh=200)   D(sh=1000)
               [tok=0]           [tok=0]      [tok=0]
                /     \
          G/A(sh=500) G/B(sh=100)
          [tok=0]     [tok=0]
                                          total_shares = 1000 + 1200 = 2200

    Two shards with different class layouts. Shard-0 has a group with two
    classes, shard-1 has two ungrouped classes. All competing for the same
    token bucket.


=== Tick 1: replenish + redistribute ===

                     satb [generated=1000, rate=1000/tick]
                      |                      |
                     454                    545        accrued = 1000 * sh / 2200
                      v                      v
                shard-0 root           shard-1 root
                [tokens=454]           [tokens=545]
                    |                    /       \
                   454                 90        454    root * child_sh / sum(children_sh)
                    v                  v          v
               G(sh=1000)         C(sh=200)   D(sh=1000)
               [tok=454]          [tok=90]    [tok=454]
                /     \
              378      75                              group * child_sh / sum(children_sh)
               v       v
          G/A(sh=500) G/B(sh=100)
          [tok=378]   [tok=75]

    --- dispatch (cost=128) ---
    G/A [tok=378] -> dispatch [378/128=2] -> [tok=122]
    G/B [tok=75]  -> no dispatch
    C   [tok=90]  -> no dispatch
    D   [tok=454] -> dispatch [454/128=3] -> [tok=70]
    total dispatched: 5

    High-share classes (G/A, D) dispatch immediately. Low-share G/B and C
    don't have enough tokens yet -- they start accumulating leftovers.


=== Tick 2: replenish + redistribute ===

                     satb [generated=2000, rate=1000/tick]
                      |                      |
                     454                    545        accrued = 1000 * sh / 2200
                      v                      v
                shard-0 root           shard-1 root
                [tokens=454]           [tokens=545]
                    |                    /       \
                   454                 90        454    root * child_sh / sum(children_sh)
                    v                  v          v
               G(sh=1000)         C(sh=200)   D(sh=1000)
               [tok=454]          [tok=90+90] [tok=454+70]  <-- new + leftover
                /     \            =180        =524
              378      75                              group * child_sh / sum(children_sh)
               v       v
          G/A(sh=500) G/B(sh=100)
          [tok=378+122][tok=75+75]                     <-- new + leftover
           =500        =150

    --- dispatch (cost=128) ---
    G/A [tok=500] -> dispatch [500/128=3] -> [tok=116]
    G/B [tok=150] -> dispatch [150/128=1] -> [tok=22]
    C   [tok=180] -> dispatch [180/128=1] -> [tok=52]
    D   [tok=524] -> dispatch [524/128=4] -> [tok=12]
    total dispatched: 5 + 9 = 14

    Leftovers from tick 1 push G/B and C over the threshold -- every class
    dispatches at least once. G/A and D ramp up with more requests.


=== Tick 3: replenish + redistribute ===

                     satb [generated=3000, rate=1000/tick]
                      |                      |
                     454                    545        accrued = 1000 * sh / 2200
                      v                      v
                shard-0 root           shard-1 root
                [tokens=454]           [tokens=545]
                    |                    /       \
                   454                 90        454    root * child_sh / sum(children_sh)
                    v                  v          v
               G(sh=1000)         C(sh=200)   D(sh=1000)
               [tok=454]          [tok=90+52] [tok=454+12]  <-- new + leftover
                /     \            =142        =466
              378      75                              group * child_sh / sum(children_sh)
               v       v
          G/A(sh=500) G/B(sh=100)
          [tok=378+116][tok=75+22]                     <-- new + leftover
           =494        =97

    --- dispatch (cost=128) ---
    G/A [tok=494] -> dispatch [494/128=3] -> [tok=110]
    G/B [tok=97]  -> no dispatch
    C   [tok=142] -> dispatch [142/128=1] -> [tok=14]
    D   [tok=466] -> dispatch [466/128=3] -> [tok=82]
    total dispatched: 14 + 7 = 21

    G/B drops back below threshold -- its 75 tokens/tick can't sustain
    every-tick dispatch. It'll fire again once leftovers cross 128.


=== Tick 4: replenish + redistribute ===

                     satb [generated=4000, rate=1000/tick]
                      |                      |
                     454                    545        accrued = 1000 * sh / 2200
                      v                      v
                shard-0 root           shard-1 root
                [tokens=454]           [tokens=545]
                    |                    /       \
                   454                 90        454    root * child_sh / sum(children_sh)
                    v                  v          v
               G(sh=1000)         C(sh=200)   D(sh=1000)
               [tok=454]          [tok=90+14] [tok=454+82]  <-- new + leftover
                /     \            =104        =536
              378      75                              group * child_sh / sum(children_sh)
               v       v
          G/A(sh=500) G/B(sh=100)
          [tok=378+110][tok=75+97]                     <-- new + leftover
           =488        =172

    --- dispatch (cost=128) ---
    G/A [tok=488] -> dispatch [488/128=3] -> [tok=104]
    G/B [tok=172] -> dispatch [172/128=1] -> [tok=44]   <-- back in action
    C   [tok=104] -> no dispatch
    D   [tok=536] -> dispatch [536/128=4] -> [tok=24]
    total dispatched: 21 + 8 = 29

    G/B fires again after accumulating two ticks of leftovers. C misses
    this tick -- its 90 tokens/tick means it dispatches roughly every
    other tick, alternating with G/B in a natural stutter pattern.
```

bhalevy:
> I gave a shallow review, but didn't really attempt to go deep, as I'm sure if I build a mental model from the description I'll get it wrong. A diagram explaining what's what will go a long way to simplify here.

@avikivity please re-review

avikivity:
If I understand correctly,

 - each shard gathers demand from active queues
 - the gathered demand is consumed from the shared token bucket
 - the consumed tokens are then distributed among the active queues

?

include/seastar/util/shared_token_bucket.hh
```diff
@@ -214,5 +214,293 @@ public:
...
+    const T _threshold;
+    const T _tau_tokens;
+    // Monotonically increasing counter of tokens produced globally.
+    // replenish() converts elapsed wall-clock time into tokens and
+    // advances this via atomic fetch-add. Consumers read it in
+    // accrued() and compute their per-class share of the delta
+    // relative to their own _last_seen snapshot (see consumer below).
+    std::atomic<T> _tokens_generated{0};
+    std::atomic<typename Clock::time_point> _last_replenished;
+    std::atomic<T> _total_shares{0};
```
avikivity, line 299: So, each shard will compare its own shares to _total_shares to see what it's fair share (and it promises not to steal)?

include/seastar/util/shared_token_bucket.hh
```diff
@@ -214,5 +214,293 @@ public:
...
+     * Only one shard succeeds per threshold window via CAS on _last_replenished.
+     */
+    void replenish(typename Clock::time_point now) noexcept {
+        auto ts = _last_replenished.load(std::memory_order_relaxed);
+        if (now <= ts) {
+            return;
+        }
+        auto extra = accumulated_in(now - ts);
+        if (extra >= _threshold) {
+            if (!_last_replenished.compare_exchange_weak(ts, now)) {
```
avikivity: compare_exchange_weak is recommended when it's in a loop. It is in fact in a loop, but the next iteration (from this shard's side) can be in 0.5ms. I think it's safer to call compare_exchange_strong() here.

include/seastar/util/shared_token_bucket.hh
```diff
@@ -214,5 +214,293 @@ public:
...
+                _available -= cost;
+                return true;
+            }
+            return false;
+        }
+
+        /* Empty the pouch and return what was in it (for redistribution). */
+        T drain() noexcept {
+            return std::exchange(_available, T(0));
+        }
```
avikivity, line 399: What about expiry? Since we generate tokens at a fixed rate, if we consume tokens and then re-distribute them at a later time, we increase token generation at that time. But the disk in the future isn't faster!

include/seastar/util/shared_token_bucket.hh
```diff
@@ -214,5 +214,293 @@ public:
...
+     * A class starts inactive and does not contribute to total_shares until
+     * activate() is called. Call activate() when the class first has pending
+     * requests, and deactivate() when it becomes idle again.
+     *
+     * Tokens generated for this consumer are not kept here; use accrued() to
+     * compute them and feed the result into an externally-owned tokens pouch.
+     */
+    class consumer {
+        shard_aware_token_bucket& _bucket;
+        T _shares;
```
avikivity, line 458: Q about low-share, high demand (e.g. compaction) on shard X competing with high-share, low-demand (e.g. reads) on shard Y. X's pouch will contain few tokens since the share count is low, Y's pouch will contain many tokens since the share count it high. But X can consume many more tokens than Y! how is that reconciled?

include/seastar/util/shared_token_bucket.hh
```diff
@@ -214,5 +214,293 @@ public:
...
+
+        /*
+         * Mark this class as actively competing for tokens. Contributes _shares
+         * to total_shares until deactivate() is called.
+         *
+         * On re-activation after idle, the class receives credit for at most
+         * tau_tokens worth of global generation (the forgiveness window). This
+         * allows briefly-idle classes to burst on return without granting
+         * unbounded catch-up to long-idle ones.
+         */
```
avikivity, line 494: What happens to a class that activates and deactivates continuously? Does it allow it to bump its consumption to burst rate? Or is it averages out (e.g. bursts penalize later consumption)?

include/seastar/util/shared_token_bucket.hh
```diff
@@ -214,5 +214,293 @@ public:
...
+
+        /* Update shares; adjusts total_shares immediately if currently active. */
+        void update_shares(T new_shares) noexcept {
+            if (_active) {
+                _bucket._total_shares.fetch_sub(_shares, std::memory_order_relaxed);
+            }
+            _shares = new_shares;
+            if (_active) {
+                _bucket._total_shares.fetch_add(new_shares, std::memory_order_relaxed);
+            }
```
avikivity: Surely this can be merged into a single atomic.

avikivity:
If I understand correctly (not a small if), it is now much more difficult for a shard to take bandwidth that is reserved for another shard. So long as a class is active, it has a bandwidth reservation of shares/Σshares that cannot be dipped into.

But we typically have lots of high-shares low-demand classes. Would these not reserve too much bandwidth?

Perhaps instead of tracking active/inactive we should track actual consumption over the last few milliseconds.

I would not want low-share high-demand classes (e,g. compaction) to be throttled against high-share demand that doesn't exist.

xemul:
> If I understand correctly,
> 
>     * each shard gathers demand from active queues

Yes. Demand is the sum of active classes' shares

>     * the gathered demand is consumed from the shared token bucket

Yes. Proportionally to the node-wide demand. However, it's easy to change. The shard-level consumption can be altered to work as it is now, i.e. each shard gets 1/Nth fraction of generated tokens where N is just the number os active shards.

>     * the consumed tokens are then distributed among the active queues

Yes. Again -- proportionally to shard's shared. Hierarchy is taken into account

include/seastar/util/shared_token_bucket.hh
xemul: Yes. But it's pretty easy to change to behave like it currently does. With this PR shard_X gets `tokens_generated * shard_X.shares / sum (shard_i.shares)`, but it can get `tokens_generated / sum(shard_i.is_active)` tokens.

include/seastar/util/shared_token_bucket.hh
xemul: It's common to any token bucket implementation. Tokens are generated in one place/time at fixed rate but and consumer elsewhere. Here it's the same -- the generated tokens counter grows monotonically over time, the consumed tokens counter grows somewhere else staying below it "on average" in the long run

include/seastar/util/shared_token_bucket.hh
xemul: X's pouch not necessarily contain few tokens. At times where Y is inactive X would grab all tokens that its shard got from global pool. In the long run the X's consumption may be comparable to Y's

include/seastar/util/shared_token_bucket.hh
xemul: When class deactivates is withdraws itself from shard's "total shares" counter and from global "total shares" counter too, so when shard starts to grab global tokens it would get less (provided other shards don't change their activity), and when the shard would start re-distributing tokens among classes, the inactive class would get nothing, only preserving what left there from previous distribute/dispatch activity.

When class activates back it adds its shares to shard's "total shares" and to global "total shares". Additionally it trims its pouch with the forgiveness factor (the tau parameter), if it was inactive for not too long it would preserve its accumulation.

xemul:
> If I understand correctly (not a small if), it is now much more difficult for a shard to take bandwidth that is reserved for another shard. So long as a class is active, it has a bandwidth reservation of shares/Σshares that cannot be dipped into.

Well, it depends on what you mean by "bandwidth reservation". I remember that concern that if we have shard-A with one 1000-shares class only and shard-B with 100-shares class only whether or not to divide disk capacity (not bandwidth!) in 10:1 ratio is highly questionable. In this PR it will be 10:1, but it's easy to change.

However (!) this PR serves two purposes and making cross-shard fairness is minor effect of it. The major effect is cross-shard preemption. As the perf test in this PR shows, it's now impossible that several shards line up on shared token bucket and high-priority class that pops up cannot preempt them.

> But we typically have lots of high-shares low-demand classes. Would these not reserve too much bandwidth?

There's no reservation here, I think. Whenever a shard needs to dispatch (i.e. -- when reactor starts polling phase) it tries to grab its share of tokens accumulated so far, distributes them among active classes and dispatches as many requests as fit into grabbed amount of token. Shards no longer create a global queue using two atomic counters chasing each other.

> Perhaps instead of tracking active/inactive we should track actual consumption over the last few milliseconds.
> 
> I would not want low-share high-demand classes (e,g. compaction) to be throttled against high-share demand that doesn't exist.

They don't. Once high-share class deactivates (i.e. zero requests queued) it withdraws itself from pretty much everywhere and doesn't participate in re-distribution of generated token until next wake-up

avikivity:
> > If I understand correctly (not a small if), it is now much more difficult for a shard to take bandwidth that is reserved for another shard. So long as a class is active, it has a bandwidth reservation of shares/Σshares that cannot be dipped into.
> 
> Well, it depends on what you mean by "bandwidth reservation". 

Reservation means that some bandwidth will be left unused even if there is a low-share high-demand class that is willing to consume it.

That is, consumption is a two-step process:
1. A shard "reserves" bandwidth according to the sum of the shares of the its active classes, divided by the global sum of the shares of the active classes
2. The individual classes then consume from the reservation as much as they want

If the consumption in step 2 is less than the reservation in step 1, we leave some bandwidth unused.

If we change "active" from a bool to be an indication of expected consumption, then it's better, because we reserve how much we expect to consume. Right now, "inactive" maps to "we expect to consume nothing" and "active" maps to "we expect to consume everything our shares will let us". But in practice many high-share classes consume a lot less.

We have to be very careful with reserving according to past consumption, because consumption can be spiky. We may end up reserving less than we need. Perhaps we should measure according to max(consumption in some window) rather than average(consumption in some window). And we should actually reserve more than max(), otherwise if demand increases we haven't reserved enough. So somewhere between max(history) and (our fair share according to the share calculation).

> I remember that concern that if we have shard-A with one 1000-shares class only and shard-B with 100-shares class only whether or not to divide disk capacity (not bandwidth!) in 10:1 ratio is highly questionable. In this PR it will be 10:1, but it's easy to change.

We also have not to go overboard with cross-shard sharing. The CPU is hard-partitioned across shards (ignoring SMT), perhaps we should hard-partition the disk as well?

It will leave unused bandwidth contrary to what I said above (I can argue for both sides here).

> 
> However (!) this PR serves two purposes and making cross-shard fairness is minor effect of it. The major effect is cross-shard preemption. As the perf test in this PR shows, it's now impossible that several shards line up on shared token bucket and high-priority class that pops up cannot preempt them.

Aha.

> 
> > But we typically have lots of high-shares low-demand classes. Would these not reserve too much bandwidth?
> 
> There's no reservation here, I think. Whenever a shard needs to dispatch (i.e. -- when reactor starts polling phase) it tries to grab its share of tokens accumulated so far, distributes them among active classes and dispatches as many requests as fit into grabbed amount of token. Shards no longer create a global queue using two atomic counters chasing each other.
> 
> > Perhaps instead of tracking active/inactive we should track actual consumption over the last few milliseconds.
> > I would not want low-share high-demand classes (e,g. compaction) to be throttled against high-share demand that doesn't exist.
> 
> They don't. Once high-share class deactivates (i.e. zero requests queued) it withdraws itself from pretty much everywhere and doesn't participate in re-distribution of generated token until next wake-up

But the pouch is filled on half-millisecond intervals, no? So if a queue is always active (on half-millisecond scale), but only has a few requests per task quota interval, will it ask for tokens on the scale of its shares or on the scale of its requests?

xemul:
> But the pouch is filled on half-millisecond intervals, no?

It's filled every time a class has something to dispatch. Under ideal circumstances this happens every task_quota, yes

> So if a queue is always active (on half-millisecond scale), but only has a few requests per task quota interval, will it ask for tokens on the scale of its shares or on the scale of its requests?

Hm, yes, that's possible. Unused tokens are kept in the class pouch until next redistribution and are limited in their amount. So a class that always has some _fraction_ of its per-tick redistribution volume will always have its configured maximum in its pouch, and the overflow tokens will be lost.

xemul:
> We also have not to go overboard with cross-shard sharing. The CPU is hard-partitioned across shards (ignoring SMT), perhaps we should hard-partition the disk as well?
> 
> It will leave unused bandwidth contrary to what I said above (I can argue for both sides here).

Hard-partition the configured capacity? AFAIK seastar has this model in the early days, each shard got its 1/smp::count share of the configured maximum. At certain instance types it happened that within default task_quota a "large enough" write request could not be served (`disk_capacity / smp::count * task_quota < request_capacity`), so the idea of IO coordinators appeared, then the "shared capacity" thing :)

include/seastar/util/shared_token_bucket.hh
xemul: Yes. Fixed in next update
This place is removed by one of the next patches anyway

xemul:
upd:
- rebased (no conflicts)
- fixed two simple review comments about atomics manipulation in shard-aware-token-bucket

todo:
- over-accumulated tokens shouldn't be lost ([here](https://github.com/scylladb/seastar/pull/3295#issuecomment-4722216092))

avikivity:
> > We also have not to go overboard with cross-shard sharing. The CPU is hard-partitioned across shards (ignoring SMT), perhaps we should hard-partition the disk as well?
> > It will leave unused bandwidth contrary to what I said above (I can argue for both sides here).
> 
> Hard-partition the configured capacity? AFAIK seastar has this model in the early days, each shard got its 1/smp::count share of the configured maximum. At certain instance types it happened that within default task_quota a "large enough" write request could not be served (`disk_capacity / smp::count * task_quota < request_capacity`), so the idea of IO coordinators appeared, then the "shared capacity" thing :)

I think modern machines grew disk bandwidth faster than shard count. i8g.24xlarge has 20 GB/s across ~100 shards, so 200 MB/s/shard. In a task quota that's 100kB. So even on these modern machines hard partitioning won't work (but we should really raise the io latency goal above the task quota).

In any case we need to support older/slower disks too.

avikivity:
> > But the pouch is filled on half-millisecond intervals, no?
> 
> It's filled every time a class has something to dispatch. Under ideal circumstances this happens every task_quota, yes
> 
> > So if a queue is always active (on half-millisecond scale), but only has a few requests per task quota interval, will it ask for tokens on the scale of its shares or on the scale of its requests?
> 
> Hm, yes, that's possible. Unused tokens are kept in the class pouch until next redistribution and are limited in their amount. So a class that always has some _fraction_ of its per-tick redistribution volume will always have its configured maximum in its pouch, and the overflow tokens will be lost.

If we have high-share, low-demand classes, the behavior can split according to request latency:
 - low latency: the queue will deactivate and we won't fill the pouch (and perhaps won't have enough tokens for the demand)
 - high latency: the queue will not deactivate and we will fill the pouch for all of our shares
 
It also depends on when activity is sampled. If it's sampled before the code to deactivate runs, it will follow the high-latency path.

Maybe we need to convert activity to a float, based on very recent history, so we don't reduce effective shares to 0 just because we deactivated.

May work with the CPU scheduler too (I never liked the penalty we give to activating tasks), though I don't remember ever seeing a problem with it.

xemul:
How about straightforward approach -- collect all overflows from nearly-full pouches and re-distribute them among other classes. It only affects one shard distribution though, if high-prio+low-demand vs low-prio+high-demand pair exists on two different shards the issue stands.

avikivity:
> How about straightforward approach -- collect all overflows from nearly-full pouches and re-distribute them among other classes.

It means no cross-shard sharing. "sum(shares) group by shard" is more or less a constant, since there will be a few high-share active classes for all shards.

I'm not sure it's so bad, but it's wasteful.

i8ge.24xlarge has 12GB/s write bandwidth across ~100 vcpus. That's just 120 MB/s. I think it's survivable but it would be better to have more, if some shards need less.

cc @travisdowns for redpanda opinions.

The alternative (I'm repeating for Travis and for clarity) it to change "active" from a boolean to a recent-history consumption rate as float, and keep the formula this_shard_shares = sum(active*shares) across all scheduling groups.

>  It only affects one shard distribution though, if high-prio+low-demand vs low-prio+high-demand pair exists on two different shards the issue stands.

Examples are compaction and tablet migration. If compaction starts on some random shard, why not let it consume more than 120 MB/s write bandwidth, if it can? The tokens not spent are gone forever.

xemul:
Ah, I see. Yes, for asymmetric shards IO this PR "resolves" only half of the problem, and leaves something for future improvement

bhalevy:
@xemul please create a custom scylla build so that @pehala can verify that this doesn't cause any regressions for the default performance workloads.

I understand that @fee-mendes had a problematic workload that was greatly improved by this change, so please close the loop so we can make it an official SCT test.

xemul:
upd:
- just rebased

xemul:
for the record: Created https://github.com/xemul/seastar/tree/br-io-queue-tokens-new-redistribution to make custom Scylla build with

xemul:
upd:
- returned trace events lost by previous rebase

xemul:
Thanks to @pehala , now we have comparison of scylla-enterprise-perf-regression-predefined-throughput-steps-i8g-tablets for this branch on top of scylladb/scylladb@4fc12a81b3c53164d5c8290afa5a2de146f50631 vs 2026.4.0.dev-20260801.4307126ef6b2 . It's not exactly apple-to-apple, but good enough (in between there are many patches, but none is performance game-changer).

There are 4 runs in this test, read, write, mixed and read_disk_only. I'll post them one-by-one eventually. Hang tight.

xemul:
# read_disk_only
Summary: same numbers for throttled periods, unthrottled throughput is slightly better with this PR, starvation time is also a bit better (smaller). Other classes are not here, because it's read-only test

### This PR
|  | P90 read [ms] | P99 read [ms] | Throughput read [op/s] |
| --- | --- | --- | --- |
| 110000 | $${\color{green}0.61}$$ | $${\color{green}0.85}$$ | 109,943 |
| 220000 | $${\color{green}0.67}$$ | $${\color{green}1.21}$$ | 219,907 |
| 330000 | $${\color{green}0.97}$$ | $${\color{green}1.95}$$ | 329,811 |
| 400000 | $${\color{green}1.51}$$ | $${\color{green}2.61}$$ | 399,753 |
| unthrottled | $${\color{green}3.68}$$ | $${\color{green}62.78}$$ | $${\color{green}549,596}$$ |

image


### Baseline
|  | P90 read [ms] | P99 read [ms] | Throughput read [op/s] |
| --- | --- | --- | --- |
| 110000 | $${\color{green}0.62}$$ | $${\color{green}0.86}$$ | 109,927 |
| 220000 | $${\color{green}0.68}$$ | $${\color{green}1.25}$$ | 219,950 |
| 330000 | $${\color{green}1.01}$$ | $${\color{green}2.08}$$ | 329,812 |
| 400000 | $${\color{green}1.73}$$ | $${\color{green}3.03}$$ | 399,810 |
| unthrottled | $${\color{green}3.14}$$ | $${\color{green}84.54}$$ | $${\color{green}507,995}$$ |

image

xemul:
# WRITE

Summary: a bit worse, but looks well within statistical error. All green. IO metrics are more interesting -- this PR renders better latencies and smaller starvation times for all writing classes (commitlog, compaction, memtable and schema_commitlog)

### This PR

|  | P90 write [ms] | P99 write [ms] | Throughput write [op/s] |
| --- | --- | --- | --- |
| 350000 | $${\color{green}1.87}$$ | $${\color{green}2.49}$$ | 349,788 |
| 600000 | $${\color{green}2.65}$$ | $${\color{green}4.28}$$ | 599,584 |
| unthrottled | $${\color{green}3.05}$$ | $${\color{green}4.22}$$ | $${\color{green}708,041}$$ |

image

image

image



### Baseline

|  | P90 write [ms] | P99 write [ms] | Throughput write [op/s] |
| --- | --- | --- | --- |
| 350000 | $${\color{green}1.68}$$ | $${\color{green}2.24}$$ | 349,808 |
| 600000 | $${\color{green}2.17}$$ | $${\color{green}3.43}$$ | 599,916 |
| unthrottled | $${\color{green}3.05}$$ | $${\color{green}14.03}$$ | $${\color{green}714,343}$$ |

image

image

image


An interesting artifact -- this PR somehow rendered disc consumption above 100%. It means that it dispatches "faster" than IO properties allow, but in the end of the day it doesn't make latency worse. Still worth checking

image

xemul:
# Mixed

Summary. Throughput is slightly off, though read latency for unthrottled case is much better. All green. IO latencies are more-or-less equal (which is expected, disk consumption is below 80%), but starvation time with this PR for sl:default and commitlog classes is lower. Also with this PR the population phase renders much better latencies (1ms -> 100us) for commitlog class

### This PR

|  | P90 write [ms] | P90 read [ms] | P99 write [ms] | P99 read [ms] | Throughput write [op/s] | Throughput read [op/s] |
| --- | --- | --- | --- | --- | --- | --- |
| 250000 | $${\color{green}0.57}$$ | $${\color{green}1.31}$$ | $${\color{green}2.14}$$ | $${\color{green}2.37}$$ | 124,933 | 124,922 |
| 480000 | $${\color{green}1.86}$$ | $${\color{green}2.06}$$ | $${\color{green}2.29}$$ | $${\color{green}3.15}$$ | 239,903 | 239,853 |
| 60000 | $${\color{green}2.01}$$ | $${\color{green}2.31}$$ | $${\color{green}2.69}$$ | $${\color{green}3.74}$$ | 299,890 | 299,915 |
| 750000 | $${\color{green}2.22}$$ | $${\color{green}2.89}$$ | $${\color{green}3.42}$$ | $${\color{green}4.29}$$ | 374,756 | 374,744 |
| unthrottled | $${\color{green}4.04}$$ | $${\color{green}25.79}$$ | $${\color{green}9.90}$$ | $${\color{green}205.39}$$ | $${\color{green}465,929}$$ | 465,965 |

image

image



### Baseline

|  | P90 write [ms] | P90 read [ms] | P99 write [ms] | P99 read [ms] | Throughput write [op/s] | Throughput read [op/s] |
| --- | --- | --- | --- | --- | --- | --- |
| 250000 | $${\color{green}0.61}$$ | $${\color{green}1.30}$$ | $${\color{green}2.13}$$ | $${\color{green}2.37}$$ | 124,889 | 124,913 |
| 480000 | $${\color{green}1.87}$$ | $${\color{green}2.08}$$ | $${\color{green}2.33}$$ | $${\color{green}3.29}$$ | 239,795 | 239,794 |
| 600000 | $${\color{green}1.99}$$ | $${\color{green}2.30}$$ | $${\color{green}2.67}$$ | $${\color{green}3.74}$$ | 299,854 | 299,812 |
| 750000 | $${\color{green}2.17}$$ | $${\color{green}2.82}$$ | $${\color{green}3.34}$$ | $${\color{green}4.18}$$ | 374,896 | 374,832 |
| unthrottled | $${\color{green}4.07}$$ | $${\color{green}44.17}$$ | $${\color{green}7.96}$$ | $${\color{green}150.99}$$ | $${\color{green}483,436}$$ | 483,482 |

image

image

xemul:
# Read (cached read)

Summary: It got _worse_ for unthrottled case, plus a 500ms p99 for 1.5M case, but I cannot attribute it to this PR -- IO dashboard shows zero IO activity throughout the test (left part on the screenshot below is population phase)

### This PR

|  | P90 read [ms] | P99 read [ms] | Throughput read [op/s] |
| --- | --- | --- | --- |
| 500000 | $${\color{green}0.55}$$ | $${\color{green}0.71}$$ | 499,602 |
| 900000 | $${\color{green}0.73}$$ | $${\color{green}0.96}$$ | 899,491 |
| 1200000 | $${\color{green}0.99}$$ | $${\color{green}1.29}$$ | 1,198,899 |
| 1500000 | $${\color{green}1.91}$$ | $${\color{red}508.04}$$ | 1,499,669 |
| unthrottled | $${\color{green}1.82}$$ | $${\color{green}15.83}$$ | $${\color{green}1,515,358}$$ |

image


### Baseline

|  | P90 read [ms] | P99 read [ms] | Throughput read [op/s] |
| --- | --- | --- | --- |
| 500000 | $${\color{green}0.55}$$ | $${\color{green}0.70}$$ | 499,690 |
| 900000 | $${\color{green}0.73}$$ | $${\color{green}0.96}$$ | 899,374 |
| 1200000 | $${\color{green}0.96}$$ | $${\color{green}1.23}$$ | 1,198,665 |
| 1500000 | $${\color{green}1.41}$$ | $${\color{green}1.86}$$ | 1,500,047 |
| unthrottled | $${\color{green}1.97}$$ | $${\color{green}4.65}$$ | $${\color{green}1,666,815}$$ |

image

xemul:
upd:
- rebased (conflicts wih #3570)

xemul:
upd:
- fixed potential tokens overflow on redistribution that could cause [consumption overgrow](https://github.com/scylladb/seastar/pull/3295#issuecomment-5202393205)

xemul:
upd:
- rebased to resolve conflicts

xemul:
@avikivity , let's continue with this. Since your last pass
   
  - Make the tokens redistribution "work-conserving" to address lost overflow tokens concern. Overflow from saturated pouches is now re-distributed among siblings that still have room (unit test included). As discussed, this covers the single-shard case; cross-shard asymmetry is left as future work
  - Perf-regression runs by @pehala are posted above. No regressions, some small latency/starvation improvements

PR #3297, opened 2026-03-18
Introduce asymmetric io_uring backend

mikolajbilski:
This PR introduces a new reactor backend - `asymmetric_io_uring`.

The new backend is mostly simillar to the existing `io_uring` backend. However, the threads performing the actual I/O for `io_uring` instances, live on separate vCPUs, called worker cores, whose set should be disjoint from the "normal" app cpuset. The purpose of this change is to improve efficiency and isolation of the I/O. Most importantly, these workers poll the `io_uring`'s submission queue, so that frequenly shards don't have to perform syscalls to submit the I/O requests.

The single worker core might (and usually is) shared between multiple shards.
This is done by using the `IORING_SETUP_ATTACH_WQ` flag when setting up `io_uring` instances.

Additionally, the backend doesn't use speculation. In difference to
other backends it always uses the "backend mechanism" - submit the operation
to the `io_uring` without speculating whether to or not perform a syscall
instead.

## Usage

To select this backend, use `--reactor-backend=asymmetric_io_uring` command line option. Note, that similarly to the `io_uring` backend, it needs to be compiled in by passing `--enable-io_uring` to `configure.py`.

The new option `--async-workers-cpuset` is used to specify which cores are the worker cores. It must be used if the new backend is selected. The worker cores are reponsible for processing the networking I/O requests submitted by shards. `io_uring` worker/poller threads are pinned to those worker cores, and those worker cores shouldn't be included in cpuset - see the table below for details of how this interacts with various options. Those worker/poller threads are shared between shards that were assigned the same networking core. The algorithm to assign worker cores to shards is as follows:

1. Let: `n` be the number of worker cores, `x` be the ID of the shard for worker core assignment.
2. Calculate `k = x % n`.
3. Select the `k`-th core from the `async-workers-cpuset` - shard `x` will be assgned this worker core.

For example, if `cpuset` is `0-13` and `async-workers-cpuset` is `14-15`, then cores `0, 2, 4, 6, 8, 10, 12` will have their worker pinned to cpu `14`, and the remaining cores `1, 3, 5, 7, 9, 11, 13` will have their worker pinned to cpu `15`.

It is recommended to specify `cpuset` explicitly when using this backend, instead of using `--smp` option.

The table below specifies the behaviour for specific combination of options used for cores configuration:

| async-workers-cpuset | smp | cpuset | taskset | overprovisioned | Resulting Behavior |
| --- | --- | --- | --- | --- | --- |
| None | Any | Any | Any | Any | **Fail** - Mandatory parameter missing |
| X | None | None | None | True | **Fail** - Overprovisioned without explicit CPU config |
| X | None | None | X | False | **Fail** - All available CPUs assigned to workers, none for shards |
| X | None | Y | N/A | False | **Success** - Shards on Y, workers on X. Warning if X ∩ Y ≠ ∅ |
| X ⊆ taskset | n | None | taskset | False | **Success** - n shards on taskset, workers on X, async-workers-cpuset don't affect cpuset for shards |
| X ⊂ taskset | None | None | taskset | False | **Success** - Shards on (taskset - X), workers on X |
| X | n | None | taskset | True | **Success** - n shards created despite contention |
| X | None | Y ⊆ taskset | taskset | True | **Success** - \|Y\| shards created regardless of X |

- `taskset` is the CPU affinity of the process, see [taskset](https://man7.org/linux/man-pages/man1/taskset.1.html). We've used it to simulate
various scenarios if number of CPUS on the system.

## Performance

The main advantage of the new backend comes from properly utilizing the *networking cores* - pinning io_uring workers/pollers to *networking cores* makes it possible to move the networking I/O away from the app cores without reducing the number of cores available to the application.

- *networking cores* - cores that are handling NIC IRQs, set by the
`perftune.py` script. Usually no shards are pinned to those cores, and
that's why we want to take advantage of them by pinning `io_uring` workers/pollers there.

The backend in some cases can yield better performance than the
existing backends, but in other cases it can be worse. It's
not a silver bullet, but rather a tool that can be used
in specific scenarios. For the biggest limitation, please see [this part](#payload-1kb-parallelism-2000)

### Networking I/O

- rpc_tester with rpc::write job, config:
    - \*P\* is replaced with the payload size, e.g. `1kB`:

    ```yaml
    jobs:
    - name: rpc_write_*P*B_parallelism_2000
      parallelism: 2000
      payload: *P*B
      type: rpc
      verb: write
      nodelay: true
    ```

- **metrics are outputted by the client**, so that’s why you always see 7/8 shards;
    - when running with 4 server shards, we run 8 client shards, so that the
    load can be evenly distributed between server shards, as `4 divides 8`
    - in other cases, we've run 7 client shards, as it's closer
    to the desired 7+1 ratio.
    - you may distinct the number of client shards just by looking at the bottom axis of the charts;

- Instance type: [c8gn.4xlarge](https://aws.amazon.com/ec2/instance-types/c8g/).
- perftune was configured with:
    > sudo ./seastar/scripts/perftune.py --tune net --nic ens34 --irq-core-auto-detection-ratio 8
- more info can be found in the [linked discussion](https://github.com/zpp-2025-io-uring/seastar/discussions/10)

#### 1 server shard

- payload size: `1kB`

image

[full results](https://github.com/zpp-2025-io-uring/seastar/discussions/10#discussioncomment-16718734)

##### Payload 450B

image

##### Payload 100B

image

After investigating the reason with `perf` we’ve identified that the networking vCPU seems to be spending around 50% time on copying memory (kernel’s `__copy_to_user`). For more details, please see the [discussion thread](https://github.com/zpp-2025-io-uring/seastar/discussions/10#discussioncomment-16733544).

This stems from moving the I/O from each shard to a single vCPU, which becomes a bottleneck. This is especially visible with the increasing ratio of shards per networking vCPU.

##### adding compute-time

If we switch the scenario and simulate some artificial processing on the server side, then `asymmetric_io_uring`, in some cases, starts to be better than the `linux-aio`. In the `rpc_tester` we’ve added an artificial busy loop, spinning for X microseconds before responding to the client’s request:
[full results](https://github.com/zpp-2025-io-uring/seastar/discussions/10#discussioncomment-16881903).

- X=2 microseconds
- payload size: `1kB`

image

#### 7 server shards

This is a desired configuration as usually there is one networking core
per 8 cores.

- X=1 microseconds
- payload size: `1kB`

image

Although `asymmetric_io_uring` yields lower raw performance than `linux-aio`, it offloads the shard vCPUs by dedicating a single vCPU to handle I/O for every 7 shards. This, consequently, gives more headroom for performing computation.

### File I/O

Benchmarks were run with io_tester. Tests were performed on a `i4i.4xlarge` instance, on the instance store nvme disk with the xfs filesystem. All benchmarks were run with the io-properties file generated by iotune.

Perftune was configured with
> sudo ./perftune.py --tune net --nic ens5 --irq-core-auto-detection-ratio 8

In our tests, the `asymmetric_io_uring` backend performed similarly to `linux-aio`. This applies both when testing only writes, as well as large writes mixed with small reads with high priority, but a limited number of requests per second.
io_throughput_writes_throughput

More results can be found in the [discussion](https://github.com/zpp-2025-io-uring/seastar/discussions/13).

## Next steps:

- #3315
- #3316

Co-authored by:
@mikolajbilski
@MrD4rkne
@witek-formanski
@pixelkubek

### DIFF
--- a/include/seastar/core/internal/pollable_fd.hh
+++ b/include/seastar/core/internal/pollable_fd.hh
@@ -93,6 +93,7 @@ public:
     friend class reactor;
     friend class pollable_fd;
     friend class reactor_backend_uring;
+    friend class reactor_backend_uring_base;
 
     future<size_t> read_some(char* buffer, size_t size);
     future<size_t> read_some(uint8_t* buffer, size_t size);
--- a/include/seastar/core/internal/pollable_fd.hh
+++ b/include/seastar/core/internal/pollable_fd.hh
@@ -94,6 +94,7 @@ public:
     friend class pollable_fd;
     friend class reactor_backend_uring;
     friend class reactor_backend_uring_base;
+    friend class reactor_backend_asymmetric_uring;
 
     future<size_t> read_some(char* buffer, size_t size);
     future<size_t> read_some(uint8_t* buffer, size_t size);
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -163,6 +163,7 @@ struct reactor_options : public program_options::option_group {
     /// * \p linux-aio
     /// * \p epoll
     /// * \p io_uring
+    /// * \p asymmetric_io_uring
     ///
     /// Default: \p linux-aio (if available).
     program_options::selection_value<reactor_backend_selector> reactor_backend;
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -25,6 +25,10 @@
 #include <seastar/util/memory_diagnostics.hh>
 #include <seastar/core/scheduling.hh>
 
+#ifdef SEASTAR_HAVE_URING
+#include <liburing.h>
+#endif
+
 namespace seastar {
 
 /// \cond internal
@@ -43,6 +47,9 @@ struct reactor_config {
     bool no_poll_aio = false;
     std::optional<bool> aio_nowait_works = false;
     bool abort_on_too_long_task_queue = false;
+#ifdef SEASTAR_HAVE_URING
+    std::variant<std::monostate, int, ::io_uring> asymmetric_uring;
+#endif
 };
 /// \endcond
 
--- a/include/seastar/core/reactor_config.hh
+++ b/include/seastar/core/reactor_config.hh
@@ -24,6 +24,7 @@
 #include <seastar/util/program-options.hh>
 #include <seastar/util/memory_diagnostics.hh>
 #include <seastar/core/scheduling.hh>
+#include <seastar/core/resource.hh>
 
 #ifdef SEASTAR_HAVE_URING
 #include <liburing.h>
@@ -174,6 +175,10 @@ struct reactor_options : public program_options::option_group {
     ///
     /// Default: \p linux-aio (if available).
     program_options::selection_value<reactor_backend_selector> reactor_backend;
+    /// \brief CPUs to use (in cpuset(7) format) for backend's async workers. Used for asymmetric_io_uring.
+    ///
+    /// \note This option is only valid when the \p reactor_backend is set to \p asymmetric_io_uring.
+    program_options::value<resource::cpuset> async_workers_cpuset;
     /// \brief Use Linux aio for fsync() calls.
     ///
     /// This reduces latency. Requires Linux 4.18 or later.
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -19,6 +19,10 @@
  * Copyright 2014 Cloudius Systems
  */
 
+#ifdef SEASTAR_HAVE_URING
+#include <liburing.h>
+#endif
+
 #include <algorithm>
 #include <atomic>
 #include <chrono>
@@ -32,6 +36,7 @@
 #include <thread>
 #include <unordered_set>
 #include <barrier>
+#include <memory>
 
 #include <grp.h>
 #include <spawn.h>
@@ -167,6 +172,7 @@
 #include <seastar/util/internal/iovec_utils.hh>
 #include <seastar/util/internal/magic.hh>
 #include <seastar/util/internal/build_id.hh>
+#include <seastar/core/reactor_config.hh>
 #include "core/crypto.hh"
 #include "core/reactor_backend.hh"
 #include "core/syscall_result.hh"
@@ -4054,6 +4060,10 @@ reactor_options::reactor_options(program_options::option_group* parent_group)
                  " Note that if the seastar_memory logger is set to debug or trace level, the diagnostics will be logged irrespective of this setting.")
     , reactor_backend(*this, "reactor-backend", backend_selector_candidates(), reactor_backend_selector::default_backend().name(),
                 fmt::format("Internal reactor implementation ({})", reactor_backend_selector::available()))
+    , async_workers_cpuset(*this, "async-workers-cpuset", resource::cpuset{},
+                "CPUs to use (in cpuset(7) format) for backend's async workers."
+                " Only applicable to, and required by, the asymmetric_io_uring reactor backend (see --reactor-backend)."
+                " Note that if the --cpuset is not set, using --async-workers-cpuset will restrict the CPUs available to the SMP shards.")
     , aio_fsync(*this, "aio-fsync", kernel_supports_aio_fsync(),
                 "Use Linux aio for fsync() calls. This reduces latency; requires Linux 4.18 or later.")
     , max_networking_io_control_blocks(*this, "max-networking-io-control-blocks", 10000,
@@ -4460,6 +4470,17 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
         cpu_set = opts_cpuset;
     }
 
+    // Let the backend selector allocate async worker cores if needed
+    auto backend_selector = reactor_opts.reactor_backend.get_selected_candidate();
+    std::shared_ptr<reactor_backend_configurator> backend_configurator;
+    try {
+        backend_configurator = backend_selector.configurator(cpu_set, reactor_opts, smp_opts);
+        cpu_set = backend_configurator->configured_cpuset();
+    } catch (const std::exception& e) {
+        seastar_logger.error("{}", e.what());
+        exit(1);
+    }
+
     if (smp_opts.smp) {
         _shard_count = smp_opts.smp.get_value();
     } else {
@@ -4560,6 +4581,8 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
         memory::configure_minimal();
     }
 
+    backend_configurator->verify_allocations(allocations);
+
     _shard_to_numa_node_mapping.reserve(_shard_count);
     for (unsigned i = 0; i < _shard_count; i++) {
         _shard_to_numa_node_mapping.push_back(allocations[i].mem.size() > 0 ? allocations[i].mem[0].nodeid : 0);
@@ -4642,6 +4665,7 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
     }
 #endif
 
+    std::barrier backend_configuration_initialized(_shard_count);
     // Better to put it into the smp class, but at smp construction time
     // correct _shard_count is not known.
     std::barrier reactors_registered(_shard_count);
@@ -4702,7 +4726,6 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
 
     _all_event_loops_done.emplace(_shard_count);
 
-    auto backend_selector = reactor_opts.reactor_backend.get_selected_candidate();
     seastar_logger.info("Reactor backend: {}", backend_selector);
 
     _qs_owner = decltype(smp::_qs_owner){new smp_message_queue* [_shard_count], qs_deleter{}};
@@ -4724,12 +4747,17 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
         }
     };
 
+    auto master_uring_fds = std::make_shared<std::vector<int>>(_shard_count, -1);
+
+    const shard_id first_shard_id = 0;
+    backend_configurator->initialize_shard_configuration(first_shard_id);
+
     unsigned i;
     auto smp_tmain = smp::_tmain;
     smp::_this_smp = this;
     for (i = 1; i < _shard_count; i++) {
         auto allocation = allocations[i];
-        create_thread([this, smp_tmain, inited, &reactors_registered, &smp_queues_constructed, &smp_opts, &reactor_opts, &reactors, hugepages_path, i, allocation, assign_io_queues, alloc_io_queues, thread_affinity, heapprof_sampling_rate, mbind, backend_selector, reactor_cfg, &mtx, &layout, use_transparent_hugepages, allocate_qs_owner, allocate_smp_queues] {
+        create_thread([this, smp_tmain, inited, &reactors_registered, &smp_queues_constructed, &smp_opts, &reactor_opts, &reactors, hugepages_path, i, allocation, assign_io_queues, alloc_io_queues, thread_affinity, heapprof_sampling_rate, mbind, backend_selector, reactor_cfg, &mtx, &layout, use_transparent_hugepages, allocate_qs_owner, allocate_smp_queues, backend_configurator, &backend_configuration_initialized] {
           try {
             // initialize thread_locals that are equal across all reacto threads of this smp instance
             smp::_tmain = smp_tmain;
@@ -4759,7 +4787,12 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
             throw_pthread_error(r);
             init_default_smp_service_group(i);
             lowres_clock::update();
-            allocate_reactor(i, backend_selector, reactor_cfg);
+
+            backend_configurator->initialize_shard_configuration(i);
+            backend_configuration_initialized.arrive_and_wait();
+            auto reactor_config = backend_configurator->finalize_apply_shard_configuration(i, reactor_cfg);
+
+            allocate_reactor(i, backend_selector, reactor_config);
             reactors[i] = &engine();
             alloc_io_queues(i);
             allocate_qs_owner(i);
@@ -4781,8 +4814,12 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
 
     init_default_smp_service_group(0);
     lowres_clock::update();
+
+    backend_configuration_initialized.arrive_and_wait();
+    auto reactor_config = backend_configurator->finalize_apply_shard_configuration(first_shard_id, reactor_cfg);
+
     try {
-        allocate_reactor(0, backend_selector, reactor_cfg);
+        allocate_reactor(0, backend_selector, reactor_config);
     } catch (const std::exception& e) {
         seastar_logger.error("{}", e.what());
         _exit(1);
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -1307,9 +1307,12 @@ detect_io_uring() {
     return bool(ring_opt);
 }
 
-class reactor_backend_uring final : public reactor_backend {
+// Base class for uring backends not to duplicate common logic.
+class reactor_backend_uring_base : public reactor_backend {
+protected:
     reactor& _r;
     ::io_uring _uring;
+private:
     bool _did_work_while_getting_sqe = false;
     bool _has_pending_submissions = false;
     file_desc _hrtimer_timerfd;
@@ -1351,7 +1354,7 @@ class reactor_backend_uring final : public reactor_backend {
             SEASTAR_ASSERT(!ret || *ret == 8);
             _armed = false;
         }
-        void maybe_rearm(reactor_backend_uring& be) {
+        void maybe_rearm(reactor_backend_uring_base& be) {
             if (_armed) {
                 return;
             }
@@ -1390,7 +1393,7 @@ private:
     ::io_uring_sqe* try_get_sqe() {
         return ::io_uring_get_sqe(&_uring);
     }
-
+protected:
     bool do_flush_submission_ring() {
         if (_has_pending_submissions) {
             _has_pending_submissions = false;
@@ -1540,10 +1543,10 @@ private:
         return fut;
     }
 public:
-    explicit reactor_backend_uring(reactor& r)
+    explicit reactor_backend_uring_base(reactor& r, ::io_uring uring)
             : reactor_backend(uses_blocking_io::yes, supports_aio_fdatasync::yes)
             , _r(r)
-            , _uring(try_create_uring(uring::QUEUE_LEN, true).value())
+            , _uring(uring)
             , _hrtimer_timerfd(make_timerfd())
             , _preempt_io_context(_r, _r._task_quota_timer, _hrtimer_timerfd)
             , _hrtimer_completion(_r, _hrtimer_timerfd)
@@ -1553,11 +1556,8 @@ public:
         auto tfd = _r._task_quota_timer.get();
         ::fcntl(tfd, F_SETFL, ::fcntl(tfd, F_GETFL) | O_NONBLOCK);
     }
-    ~reactor_backend_uring() {
+    ~reactor_backend_uring_base() override {
         ::io_uring_queue_exit(&_uring);
-    }
-    virtual std::string_view get_backend_name() const override {
-        return "io_uring";
     }
     virtual bool reap_kernel_completions() override {
         return do_process_kernel_completions();
@@ -1615,6 +1615,43 @@ public:
         auto* pfd = static_cast<uring_pollable_fd_state*>(&fd);
         delete pfd;
     }
+
+    virtual void signal_received(int signo, siginfo_t* siginfo, void* ignore) override {
+        _r._signals.action(signo, siginfo, ignore);
+    }
+    virtual void start_tick() override {
+        _preempt_io_context.start_tick();
+    }
+    virtual void stop_tick() override {
+        _preempt_io_context.stop_tick();
+    }
+    virtual void arm_highres_timer(const ::itimerspec& its) override {
+        _hrtimer_timerfd.timerfd_settime(TFD_TIMER_ABSTIME, its);
+    }
+    virtual void reset_preemption_monitor() override {
+        _preempt_io_context.reset_preemption_monitor();
+    }
+    virtual void request_preemption() override {
+        _preempt_io_context.request_preemption();
+    }
+    virtual void start_handling_signal() override {
+        // We don't have anything special wrt. signals
+    }
+    virtual pollable_fd_state_ptr make_pollable_fd_state(file_desc fd, pollable_fd::speculation speculate) override {
+        return pollable_fd_state_ptr(new uring_pollable_fd_state(std::move(fd), std::move(speculate)));
+    }
+};
+
+class reactor_backend_uring final : public reactor_backend_uring_base {
+public:
+    explicit reactor_backend_uring(reactor& r)
+        : reactor_backend_uring_base(r, try_create_uring(uring::QUEUE_LEN, true).value()) {
+    }
+
+    virtual std::string_view get_backend_name() const override {
+        return "io_uring";
+    }
+
     virtual future<std::tuple<pollable_fd, socket_address>> accept(pollable_fd_state& listenfd) override {
         if (listenfd.take_speculation(POLLIN)) {
             try {
@@ -1956,31 +1993,6 @@ public:
         auto desc = std::make_unique<recv_completion>(fd, ba->allocate_buffer());
         auto req = internal::io_request::make_recv(fd.fd.get(), desc->get_write(), desc->get_size(), 0);
         return submit_request(std::move(desc), std::move(req));
-    }
-
-    virtual void signal_received(int signo, siginfo_t* siginfo, void* ignore) override {
-        _r._signals.action(signo, siginfo, ignore);
-    }
-    virtual void start_tick() override {
-        _preempt_io_context.start_tick();
-    }
-    virtual void stop_tick() override {
-        _preempt_io_context.stop_tick();
-    }
-    virtual void arm_highres_timer(const ::itimerspec& its) override {
-        _hrtimer_timerfd.timerfd_settime(TFD_TIMER_ABSTIME, its);
-    }
-    virtual void reset_preemption_monitor() override {
-        _preempt_io_context.reset_preemption_monitor();
-    }
-    virtual void request_preemption() override {
-        _preempt_io_context.request_preemption();
-    }
-    virtual void start_handling_signal() override {
-        // We don't have anything special wrt. signals
-    }
-    virtual pollable_fd_state_ptr make_pollable_fd_state(file_desc fd, pollable_fd::speculation speculate) override {
-        return pollable_fd_state_ptr(new uring_pollable_fd_state(std::move(fd), std::move(speculate)));
     }
 };
 
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -1308,10 +1308,6 @@ detect_io_uring() {
 }
 
 class reactor_backend_uring final : public reactor_backend {
-    // s_queue_len is more or less arbitrary. Too low and we'll be
-    // issuing too small batches, too high and we require too much locked
-    // memory, but otherwise it doesn't matter.
-    static constexpr unsigned s_queue_len = 200;
     reactor& _r;
     ::io_uring _uring;
     bool _did_work_while_getting_sqe = false;
@@ -1521,8 +1517,8 @@ private:
 
     // Returns true if completions were processed
     bool do_process_kernel_completions_step() {
-        struct ::io_uring_cqe* buf[s_queue_len];
-        auto n = ::io_uring_peek_batch_cqe(&_uring, buf, s_queue_len);
+        struct ::io_uring_cqe* buf[uring::QUEUE_LEN];
+        auto n = ::io_uring_peek_batch_cqe(&_uring, buf, uring::QUEUE_LEN);
         do_process_ready_kernel_completions(buf, n);
         ::io_uring_cq_advance(&_uring, n);
         return n != 0;
@@ -1547,7 +1543,7 @@ public:
     explicit reactor_backend_uring(reactor& r)
             : reactor_backend(uses_blocking_io::yes, supports_aio_fdatasync::yes)
             , _r(r)
-            , _uring(try_create_uring(s_queue_len, true).value())
+            , _uring(try_create_uring(uring::QUEUE_LEN, true).value())
             , _hrtimer_timerfd(make_timerfd())
             , _preempt_io_context(_r, _r._task_quota_timer, _hrtimer_timerfd)
             , _hrtimer_completion(_r, _hrtimer_timerfd)
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -2019,6 +2019,87 @@ public:
     }
 };
 
+class reactor_backend_asymmetric_uring final : public reactor_backend_uring_base {
+public:
+    explicit reactor_backend_asymmetric_uring(reactor& r)
+        : reactor_backend_uring_base(r, try_create_uring(uring::QUEUE_LEN, true).value()) {
+    }
+
+    virtual std::string_view get_backend_name() const override {
+        return "asymmetric_io_uring";
+    }
+
+    virtual future<std::tuple<pollable_fd, socket_address>> accept(pollable_fd_state& listenfd) override {
+        class accept_completion final : public accept_completion_base {
+        public:
+            accept_completion(pollable_fd_state& listenfd)
+                : accept_completion_base(listenfd) {}
+            void complete(size_t fd) noexcept final {
+                pollable_fd pfd(file_desc::from_fd(fd));
+                _result.emplace_value(std::move(pfd), std::move(_sa));
+                delete this;
+            }
+        };
+        auto desc = std::make_unique<accept_completion>(listenfd);
+        auto req = internal::io_request::make_accept(listenfd.fd.get(), desc->posix_sockaddr(), desc->socklen_ptr(), SOCK_NONBLOCK | SOCK_CLOEXEC);
+        return submit_request(std::move(desc), std::move(req));
+    }
+
+    virtual future<> connect(pollable_fd_state& fd, socket_address& sa) override {
+        auto desc = std::make_unique<connect_completion_base>(sa);
+        auto req = internal::io_request::make_connect(fd.fd.get(), desc->posix_sockaddr(), desc->socklen());
+        return submit_request(std::move(desc), std::move(req));
+    }
+
+    virtual future<size_t> read(pollable_fd_state& fd, void* buffer, size_t len) override {
+        auto desc = std::make_unique<sized_promise_completion_base>();
+        const uint64_t position_file_offset = -1;
+        auto req = internal::io_request::make_read(fd.fd.get(), position_file_offset, buffer, len, false);
+        return submit_request(std::move(desc), std::move(req));
+    }
+
+    virtual future<size_t> recvmsg(pollable_fd_state& fd, const std::vector<iovec>& iov) override {
+        auto desc = std::make_unique<recvmsg_completion_base>(iov);
+        auto req = internal::io_request::make_recvmsg(fd.fd.get(), desc->msghdr(), 0);
+        return submit_request(std::move(desc), std::move(req));
+    }
+
+    virtual future<temporary_buffer<char>> read_some(pollable_fd_state& fd, internal::buffer_allocator* ba) override {
+        auto desc = std::make_unique<read_completion_base>(ba->allocate_buffer());
+        const uint64_t position_file_offset = -1;
+        auto req = internal::io_request::make_read(fd.fd.get(), position_file_offset, desc->get_write(), desc->get_size(), false);
+        return submit_request(std::move(desc), std::move(req));
+    }
+
+    virtual future<size_t> sendmsg(pollable_fd_state& fd, std::span<iovec> iovs, size_t len) final {
+        auto desc = std::make_unique<sendmsg_completion_base>(iovs, len);
+        auto req = internal::io_request::make_sendmsg(fd.fd.get(), desc->msghdr(), MSG_NOSIGNAL);
+        return submit_request(std::move(desc), std::move(req));
+    }
+
+#if SEASTAR_API_LEVEL < 9
+    virtual future<size_t> send(pollable_fd_state& fd, const void* buffer, size_t len) override {
+        auto desc = std::make_unique<send_completion_base>(len);
+        auto req = internal::io_request::make_send(fd.fd.get(), buffer, len, MSG_NOSIGNAL);
+        return submit_request(std::move(desc), std::move(req));
+    }
+#endif
+
+    virtual future<temporary_buffer<char>> recv_some(pollable_fd_state& fd, internal::buffer_allocator* ba) override {
+        auto desc = std::make_unique<read_completion_base>(ba->allocate_buffer());
+        auto req = internal::io_request::make_recv(fd.fd.get(), desc->get_write(), desc->get_size(), 0);
+        return submit_request(std::move(desc), std::move(req));
+    }
+
+    virtual future<size_t> writev(pollable_fd_state& fd, std::span<iovec> iovs) override {
+        auto desc = std::make_unique<sized_promise_completion_base>();
+        const uint64_t position_file_offset = -1;
+        std::vector<iovec> iov(iovs.begin(), iovs.end());
+        auto req = internal::io_request::make_writev(fd.fd.get(), position_file_offset, iov, false);
+        return submit_request(std::move(desc), std::move(req));
+    }
+};
+
 #endif
 
 static bool detect_aio_poll() {
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -22,8 +22,11 @@
 #include <atomic>
 #include <chrono>
 #include <filesystem>
+#include <memory>
+#include <optional>
 #include <thread>
 #include <utility>
+#include <vector>
 #include <fcntl.h>
 #include <signal.h>
 #include <sys/epoll.h>
@@ -49,8 +52,10 @@
 #include <seastar/core/coroutine.hh>
 #include <seastar/core/reactor_config.hh>
 #include <seastar/core/smp.hh>
+#include <seastar/core/shard_id.hh>
 #include <seastar/util/defer.hh>
 #include <seastar/util/read_first_line.hh>
+#include <seastar/core/resource.hh>
 
 namespace seastar {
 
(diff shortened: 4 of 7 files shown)


src/core/reactor_backend.cc
```diff
@@ -1820,63 +1939,443 @@ class reactor_backend_uring final : public reactor_backend {
...
+
+        buffer get_buf(size_t id) {
+            SEASTAR_ASSERT(id < _buffers.size());
+            return _buffers[id];
+        }
+
+        void return_buf(size_t id) {
+            SEASTAR_ASSERT(id < _buffers.size());
+            const auto& buf = _buffers[id];
+            ::io_uring_buf_ring_add(_buffer_ring, buf.ptr, buf.len, id, _mask, 0);
```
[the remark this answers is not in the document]
MrD4rkne: We're advancing the buffer ring right away.

> If just one buffer is provided before the
       ring tail is committed with [io_uring_buf_ring_advance(3)](https://man7.org/linux/man-pages/man3/io_uring_buf_ring_advance.3.html) or
       [io_uring_buf_ring_cq_advance(3)](https://man7.org/linux/man-pages/man3/io_uring_buf_ring_cq_advance.3.html), then buf_offset should be 0.
       
https://man7.org/linux/man-pages/man3/io_uring_buf_ring_add.3.html

src/core/reactor.cc
```diff
@@ -19,6 +19,14 @@
...
  */
 
+#ifdef SEASTAR_HAVE_URING
+#include <liburing.h>
+#endif
+
+#ifdef SEASTAR_MODULE
+module;
+#endif
+
```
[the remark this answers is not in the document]
MrD4rkne: Somehow missed during review, i think it's an artifact of a rebase. See https://github.com/scylladb/seastar/commit/a94927fbb3d064d7866f3862b6a025372bbcfa04

avikivity:
> This PR introduces a new reactor backend - `asymmetric_io_uring`.

We should look for a name that will be more self-explanatory to users. But that discussion can take months, let's table it for now.

> 
> The new backend is mostly simillar to the existing `io_uring` backend. However, the threads performing the actual I/O for io_uring instances live on separate CPUs, called worker cores, whose set should be disjoint from the "normal" app cpuset. The purpose of this change is to improve efficiency and isolation of the networking I/O.

Another distinction is moving from the notification model for networking (IORING_OP_ADD) to the completion based model (IORING_OP_SENDMSG).

So we're really doing two things
 - completion model (can work without SQPOLL)
 - asymmetric (completion + SQLPOLL)

avikivity:
> The algorithm to assign worker cores to shards is as follows:

We'll probably have to revise it for NUMA. Nothing says we have to have NUMA nodes of equal size. But it's enough to start with.

avikivity:
> So we're really doing two things
> 
> * completion model (can work without SQPOLL)


That's wrong, this was done in 5d8d5ae625fc443f6d5ed467a1bdc512a98a7e62.

include/seastar/core/internal/pollable_fd.hh
```diff
@@ -94,6 +94,7 @@ public:
     friend class pollable_fd;
     friend class reactor_backend_uring;
```
avikivity, line 95: > The new backend is mostly simillar to the existing `io_uring` backend. However, the threads performing the actual I/O for io_uring instances live on separate CPUs, called worker cores, whose set should be disjoint from the "normal" app cpuset. The purpose of this change is to improve efficiency and isolation of the networking I/O.

Since the existing backend is also completion based, why do we need a new backend? Isn't the change mostly about configuring the ring? I'll read on.

avikivity: Ok, the existing backend isn't pure completion based, it also speculatively tries to read directly.

Would the existing backend work if we just disabled the speculation part?

src/core/reactor.cc
```diff
@@ -19,6 +19,14 @@
  * Copyright 2014 Cloudius Systems
  */
 
+#ifdef SEASTAR_HAVE_URING
+#include <liburing.h>
+#endif
+
+#ifdef SEASTAR_MODULE
+module;
+#endif
```
avikivity: This isn't needed.

avikivity: (and breaks with newer compilers)

include/seastar/core/internal/io_request.hh
```diff
@@ -165,6 +193,18 @@ public:
         return req;
     }
 
+    static io_request make_uring_buf_group_recv(int fd, size_t size, int flags, uint16_t buf_group) {
+        io_request req;
+        req._uring_buf_group_recv = {
+          .op = operation::uring_buf_group_recv,
```
avikivity: We use 4-space indents, not 2.

src/core/reactor_backend.cc
```diff
@@ -2073,12 +2090,215 @@ unsigned get_uring_group_id(seastar::shard_id shard_id, const resource::cpuset&
...
+            _buffer_ring = ::io_uring_setup_buf_ring(_uring, ring_entries, s_buffer_group_id, 0, &err);
+            if (err) {
+                throw std::system_error(-err, std::generic_category(), "io_uring_setup_buf_ring");
+            }
+
+            _buffers.reserve(ring_entries);
+            for (unsigned i = 0; i < ring_entries; ++i) {
+                void* ptr;
+                if (int err = posix_memalign(&ptr, memory::page_size, ring_buffer_size); err) {
+                    throw std::system_error(err, std::generic_category(), "posix_memalign");
```
avikivity: The buffer ring has to be replenished in runtime via new allocations. That's because the lifetime of a buffer returned by recvmsg is controlled by the application; if it holds onto the temporary_buffer it can starve the ring of buffers.

src/core/reactor_backend.cc
```diff
@@ -2073,12 +2090,215 @@ unsigned get_uring_group_id(seastar::shard_id shard_id, const resource::cpuset&
...
+    };
+
+    class ring_buffer_provider {
+        lw_shared_ptr<ring_buffer_provider_impl> _impl;
+
+    public:
+        explicit ring_buffer_provider(::io_uring* ring, std::optional<uring_buffer_ring_config> config_opt) {
+            if (config_opt.has_value()) {
+                try {
+                    _impl = seastar::make_lw_shared<ring_buffer_provider_impl>(ring, config_opt.value());
```
avikivity: Why lw_shared_ptr and indirection? This is all internal.

src/core/reactor_backend.cc
```diff
@@ -2121,6 +2341,12 @@ class reactor_backend_asymmetric_uring final : public reactor_backend_uring_base
     }
 
     virtual future<temporary_buffer<char>> read_some(pollable_fd_state& fd, internal::buffer_allocator* ba) override {
+        if (_uring_buffer_ring.reserve()) {
+            auto desc = std::make_unique<uring::buf_group_io_completion>(fd);
+            const uint64_t position_file_offset = -1;
+            auto req = internal::io_request::make_uring_buf_group_read(fd.fd.get(), position_file_offset, _uring_buffer_ring.buffer_size_in_bytes(), _uring_buffer_ring.buf_group());
+            return submit_request(std::move(desc), std::move(req));
+        }
```
avikivity: You're just ignoring the ba parameter, no?

avikivity: There's a big problem here. The latency of this backend is high, because any completions can be delayed as much as 0.5ms. This limits the throughput of any single stream to buffer_size/period. For buffer_size=8k and period=0.5ms, that's 16 MB/sec, too low.

Larger buffer sizes end up wasting a lot of memory for small request/response workloads.

A solution is to post recvmsg with 16 8k buffers. A request/respond workload will consume just one of those buffers, limiting waste. Throughput workloads will consume all 16 buffers, so we have 128k/0.5ms = 256 MB/s, sufficient for a single stream in a shard.

src/core/reactor_backend.cc
```diff
@@ -1205,6 +1205,9 @@ detect_io_uring() {
     return bool(ring_opt);
 }
 
+class reactor_backend_uring_base : public reactor_backend {
+};
+
 class reactor_backend_uring final : public reactor_backend {
```
xemul: This should be patched to inherit from reactor_backend_uring_base right in this patch

src/core/reactor_backend.cc
```diff
@@ -1541,14 +1541,18 @@ class reactor_backend_uring_base : public reactor_backend {
         });
     }
 
+    virtual void handle_ready_kernel_completion(kernel_completion* completion, ::io_uring_cqe* cqe) {
```
xemul: Is it really required? kernel_completion->complete_with is already pure virtual method, why doing extra virtual dispatch? (I'll probably find it out when I get to new backend's implementation, so asking it here not to forget)

include/seastar/core/internal/pollable_fd.hh
```diff
@@ -94,6 +94,7 @@ public:
...
     friend class reactor_backend_uring_base;
+    friend class reactor_backend_asymmetric_uring;
```
xemul, line 97: A comment about the whole patch, not this particular line:

The patch comment describes what asymetric backend is, in general. This should come as PR description. All the more so "threads living on separate CPUs" are not introduced in this patch.

Next, the phrase

> The fast track has been removed, as it does not make sense to use it with this backend.

doesn't explain what "fast track" is and why it's removed. "Does not make sense" can be clear now, but after some time (or to newcoments) this "sense" will need explanations.

Please, tune the patch comment to say only what this patch does. In particular -- explain what's the difference in implementation of the methods, that come with this patch. One difference I see -- no speculation on FDs (presumably, that's the fast-track in question). Anything else?

src/core/reactor_backend.cc
```diff
@@ -1914,10 +1915,123 @@ class reactor_backend_uring final : public reactor_backend_uring_base {
...
+    }
+}
+
+} // namespace uring
+
 class reactor_backend_asymmetric_uring final : public reactor_backend_uring_base {
 public:
     explicit reactor_backend_asymmetric_uring(reactor& r)
-        : reactor_backend_uring_base(r, try_create_uring(uring::QUEUE_LEN, true).value()) {
+        : reactor_backend_uring_base(r, uring::try_create_asymmetric_uring(r._cfg.asymmetric_uring, true).value()) {
```
xemul, line 2303: The patch comment got "cut":

> and `try_create_attached_asymmetric_uring` which will have its workers attached t>

Please, update.

There are three options to get uring here:

- try_create_attached_asymmetric_uring
- try_create_base_asymmetric_uring (not called in this patch)
- std::get<compile_safe_io_uring>(variant)

So two clarifications are required:
- which shards get the ..._attached_.. uring and which get "std::get<>(variant)" one
- who's going to get the ..._base_... one

src/core/reactor_backend.cc
```diff
@@ -1923,7 +1923,8 @@ std::optional<::io_uring>
 try_create_asymmetric_uring_impl(::io_uring_params params, bool throw_on_error) {
     auto required_features =
             IORING_FEAT_SUBMIT_STABLE
-            | IORING_FEAT_NODROP;
+            | IORING_FEAT_NODROP
+            | IORING_FEAT_SQPOLL_NONFIXED;
```
xemul: If it comes in its own patch, and you say it's questionable wheter it helps or not, probably it deserves its own on/off switch on reactor_config and CLI option to control one

src/core/reactor.cc
```diff
@@ -4552,7 +4602,6 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
 
     _all_event_loops_done.emplace(smp::count);
 
-    auto backend_selector = reactor_opts.reactor_backend.get_selected_candidate();
     seastar_logger.info("Reactor backend: {}", backend_selector);
```
xemul, line 4729: How does this info() work with backend_selector removed?

xemul: Ah, it's move upstairs

src/core/reactor.cc
```diff
@@ -4410,6 +4457,8 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
         memory::configure_minimal();
     }
 
+    warn_if_shards_and_async_workers_share_cpu(allocations, async_worker_cpus);
```
xemul: Maybe it makes sense to re-consider the new option to be CPUs that are withdrawn from --cpuset mask, not added to it

src/core/reactor_backend.cc
```diff
@@ -2189,4 +2205,17 @@ std::vector<reactor_backend_selector> reactor_backend_selector::available() {
     return ret;
 }
 
+async_worker_allocation reactor_backend_selector::allocate_async_workers(const resource::cpuset& async_workers_cpu_set, const resource::cpuset& cpu_set) const {
```
xemul: Why is this selector method, but not a static/lambda in reactor.cc itself?

src/core/reactor_backend.cc
```diff
@@ -2073,12 +2090,215 @@ unsigned get_uring_group_id(seastar::shard_id shard_id, const resource::cpuset&
...
+            return ring_buffer_size;
+        }
+
+        uint16_t buf_group() const noexcept {
+            return s_buffer_group_id;
+        }
+    };
+
+    class ring_buffer_provider {
+        lw_shared_ptr<ring_buffer_provider_impl> _impl;
```
xemul: Why is it needed? I don't see where _impl pointer is shared with any other holder

src/core/reactor_backend.cc
```diff
@@ -2073,12 +2090,215 @@ unsigned get_uring_group_id(seastar::shard_id shard_id, const resource::cpuset&
     return shard_id % worker_cpus.size();
 }
 
+class buf_group_io_completion final: public io_completion {
+    std::optional<temporary_buffer<char>> _buffer_opt;
```
xemul: It looks like instead of this field, the one can have a shared pointer on buffer_provider and override its complete so that it gets the buffer from it and resolves _result with the buffer, instead of handle_ready_kernel_completion doing this via set_buffer (and the handle_ready_kernel_completion itself goes away, as I suggested before)

src/core/reactor.cc
MrD4rkne: We didn't even think of it, it sounds better than the current one.

src/core/reactor_backend.cc
pixelkubek: Getting the correct buffer from the buffer ring requires reading the buffer id from the CQE's `flags` field, so wouldn't `handle_ready_kernel_completion` still be needed to pass the buffer id to the `buf_group_io_completion`?

src/core/reactor_backend.cc
pixelkubek: Currently we fallback to the buffer allocator if there are no buffers in the buffer ring. Do I understand correctly that we should never use the buffer allocator and always use the buffer ring?

In our benchmarking we were always using large buffer sizes (e.g. 128kB) for the buffer ring. Is it OK to allocate such large buffers during the runtime, or should we change our approach with regards to how the buffer size is decided (currently given by the user as a commandline option)?

src/core/reactor_backend.cc
witek-formanski: There’s still ongoing discussion about this here: https://github.com/zpp-2025-io-uring/seastar/discussions/6. Once we understand the issue better, we can either fix it properly or remove the mechanism entirely. But I think that adding a switch at this stage would likely increase complexity from user's perspective without providing reliable value.

src/core/reactor.cc
avikivity: It would be better if starting from scratch, but current setup already has networking vcpus and reactor vcpus. It's easier to migrate if those roles are kept.

MrD4rkne: > It would be better if starting from scratch, but current setup already has networking vcpus and reactor vcpus. It's easier to migrate if those roles are kept.

Current setup - is it the general idea of perftune, or you mean what we introduce in this PR?

Both approaches have their ups and downs

Having vCPUs being removed from the cpuset reduce many "edge cases", like, we don't wanna share the vCPU between shard and the worker, as both - reactor thread and sq_thread try to have as much of the cpu time as possible.

However, it requires more tuning for the user, who has already the setup done. If somebody has all the configs done and wanna use networking cores for the async workers, then it needs to add them back to the cpuset and specify them in the `--async-workers-cpuset`.

@xemul @avikivity would you have some other arguments for one of these options?

avikivity: > > It would be better if starting from scratch, but current setup already has networking vcpus and reactor vcpus. It's easier to migrate if those roles are kept.
> 
> Current setup - is it the general idea of perftune, or you mean what we introduce in this PR?

The general idea of perftune.

> 
> Both approaches have their ups and downs
> 
> Having vCPUs being removed from the cpuset reduce many "edge cases", like, we don't wanna share the vCPU between shard and the worker, as both - reactor thread and sq_thread try to have as much of the cpu time as possible.
> 
> However, it requires more tuning for the user, who has already the setup done. If somebody has all the configs done and wanna use networking cores for the async workers, then it needs to add them back to the cpuset and specify them in the `--async-workers-cpuset`.


If we'd started from scratch, it would make more sense to have a global cpuset and specialized cpusets for roles (like io_uring workers).

But we aren't starting from scratch, and --cpuset now means "reactor shards" while workers (called "networking vcpus") are defined elsewhere.

Usually, a script sets the options so it's not a problem.

Later, if we want, we can add new option set, for example --dedicated-cpuset (or whatenere) for (reactors+workers) and --worker-cpuset for workers. But let's not define --cpuset now, consider a system migrating from the current setup to io_uring.

> 
> @xemul @avikivity would you have some other arguments for one of these options?

Deexie: I second staying with the current version. `--cpuset` has its own meaning and it should remain consistent across different backends. Especially since networking cores aren't something new either

MrD4rkne: Thanks for the input!

src/core/reactor_backend.cc
witek-formanski: I understand that relying on a single 8k provided buffer pessimistically caps us at 16 MB/s due to the 0.5ms reactor tick, and that recvmsg with a 16x8k `iovec` fixes this. However, the `IOSQE_BUFFER_SELECT` option currently only gives us one buffer from the ring per operation. Are you suggesting we should bypass the io_uring buffer ring for `recv_some` and allocate the `iovec` array directly from seastar's internal `buffer_allocator`? Or is there a specific way you recommend for combining `recvmsg` with the provided buffer ring?

travisdowns:
In all of the performance results presented, this new backend is the worst performing option, or am I misreading that?

Maybe that's expected if this is mostly for isolation, but I think it makes sense to try to also bench and provide results for the type of scenario for when you expect this to help?

src/core/reactor_backend.cc
avikivity: I wasn't aware of this limitation.

>  Are you suggesting we should bypass the io_uring buffer ring for recv_some and allocate the iovec array directly from seastar's internal buffer_allocator

No, this would require 128k * nr_connections to be allocated. If nr_connections = 10,000, that's 1.2 GB.

Possible solutions:

1. Use IOSQE_IO_LINK to chain several recvmsg calls each with its own IOSQE_BUFFER_SELECT. The first in the chain should have MSG_DONTWAIT unset (so it waits), the others should have MSG_DONTWAIT set, so they only trigger if there is data to consume. Somewhat heavyweight.

2. Do something like posix_data_source_impl::get() to scale the buffer sizes:

```c++
future<temporary_buffer<char>>
posix_data_source_impl::get() {
    return _fd.recv_some(static_cast<internal::buffer_allocator*>(this)).then([this] (temporary_buffer<char> b) {
        if (b.size() >= _config.buffer_size) {
            _config.buffer_size *= 2;
            _config.buffer_size = std::min(_config.buffer_size, _config.max_buffer_size);
        } else if (b.size() <= _config.buffer_size / 4) {
            _config.buffer_size /= 2;
            _config.buffer_size = std::max(_config.buffer_size, _config.min_buffer_size);
        }
        bytes_received[sg_id] += b.size();
        return b;
    });
}
```

In fact this should have been the first option, since it's already implemented. We need buffer pools for sizes 8k->128k, doubling in size.

There is an impedance mismatch between the Seastar stack an io_uring: Seastar works in the traditional way where the buffer is allocated before recvmsg(), while io_uring in SELECT_BUFFER mode allocates late. In fact how does it work now? Do we just ignore the allocated buffer?

Suggest this:
a. Drop PROVIDE_BUFFERS/SELECT_BUFFER for now since it requires some surgery
b. Verify that the buffer growing works as expected and allows high throughput for a single connection even when the reactor polls once per 0.5ms
c. come back to it later and think of a way to make it work

I don't understand how the allocation in posix_data_source_impl::allocate_buffer() works with PROVIDE_BUFFERS.

include/seastar/core/internal/io_request.hh
mikolajbilski: We used 2-space indents here for consistency with the way other `io_request`'s `operation`s are initialised in this file. Should we change it to 4-spaces just in the code we added/everywhere in the file, or just leave it as-is?

src/core/reactor.cc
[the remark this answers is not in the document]
mikolajbilski: The `module` shouldn't be here at all

src/core/reactor_backend.cc
witek-formanski: You are right. This is moved in the new revision

src/core/reactor_backend.cc
witek-formanski: Could you explain a bit more what is your idea (especially the second proposed approach)? We are all somewhat lost. We are not sure, where exactly we should implement the buffer scaling like in the mentioned `posix_data_source_impl::get()`. And if it is desirable to "duplicate" this logic, if we could just use this method (by effectively reverting part of our current implementation)?

src/core/reactor_backend.cc
mikolajbilski: iirc, we have decided on commit strategy of "introduce an empty base uring class in one commit, then move common functions to this class in the next commit". Applying this suggestion would mean that we have to either:

- move the constructor from `reactor_backend_uring` to `reactor_backend_uring_base` immediatly, along with data members and `make_timerfd()`, which I believe creates more confusion as there is no such clean commit structure as before
- add a "dummy" constructor to `reactor_backend_uring_base`, that just passes its arguments up to `reactor_backend` - such constructor would be immediately deleted in the next patch, and I believe also only add to the confusion of the reader

Which of these options should we implement? Or should we do something else entirely? As this inheritance structure is patched right in the next commit, I believe the best option is to leave it as-is, but we can do whatever option you prefer

mikolajbilski:
This force-push changes the following:

- move `allocate_async_workers` from the `reactor_backend.*` to the `reactor.cc`
- remove indirection in `ring_buffer_provider` implementation
- allocate new buffer ring buffers at runtime
- remove unnecessary `module` from `reactor.cc`
- rewrite the description of `Introduce asymmetric io_uring backend`
- fix accidentally cut commit description of `Implement asymmetric uring creation functions`

src/core/reactor_backend.cc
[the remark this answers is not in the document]
mikolajbilski: Resolving, as this suggestion is incorrect

src/core/reactor.cc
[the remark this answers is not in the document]
mikolajbilski: Resolving, as the `module` declaration has been completely removed

src/core/reactor.cc
mikolajbilski: This `module` has been removed in the recent force-push

include/seastar/core/internal/pollable_fd.hh
mikolajbilski: The description for this comment has been updated in the recent force-push

src/core/reactor_backend.cc
mikolajbilski: This has been fixed in the recent force-push

src/core/reactor.cc
mikolajbilski: Resolving as this is a non-issue

src/core/reactor.cc
mikolajbilski: Resolving, as the conclusion is to leave the system as-is

src/core/reactor_backend.cc
pixelkubek: The `_impl` pointer was copied to the temporary_buffer's deleter.
We used it to ensure that the `ring_buffer_provider_impl` isn't destroyed while there exist temporary buffers with a reference to it.
This, however, was not needed, so we removed the indirection in the latest force-push.

mikolajbilski:
> In all of the performance results presented, this new backend is the worst performing option, or am I misreading that?
> 
> Maybe that's expected if this is mostly for isolation, but I think it makes sense to try to also bench and provide results for the type of scenario for when you expect this to help?

@travisdowns there are currently some performance issues regarding this backend, e.g. related to the use of `SQPOLL` (see https://github.com/zpp-2025-io-uring/seastar/discussions/6 for more details). The current results also aren't reliable, as we are testing it on a "standard" PC instead of an AWS machine - those results will come in soon and hopefully will be much more favourable for this backend. We managed to get some test runs wirh `rpc write` where the new backend is the best, however we are unable to reproduce them, most likely due to the limitations of the machine we are using for testing. Here are the results:

image

src/core/reactor_backend.cc
pixelkubek: It was unnecessary, we removed it in the force-push.

We used it to guarantee that `io_uring_buf_ring` isn't destroyed while a temporary buffer still has a reference to it.
However, this was not needed, as that would require the reactor backend to be destroyed before the temporary buffer.

src/core/reactor_backend.cc
pixelkubek: This is now implemented in the new revision.

include/seastar/core/internal/io_request.hh
Deexie: Change it to 4 spaces in your code.

You can fix it in other places in a separate commit

src/core/reactor_backend.cc
Deexie: The latter. 

We can't go with the former as it would mess up the commit structure.

We should not leave it as-is, because now we are adding an empty "leaf" class that does nothing and it is confusing why we even have it here. We introduce the class - not only the name of the class, but also it's meaning and placement in the inheritance structure.

> such constructor would be immediately deleted in the next patch

It will be modified not deleted, right? Anyway, this is justified

travisdowns:
> @travisdowns there are currently some performance issues regarding this backend, e.g. related to the use of `SQPOLL` (see [zpp-2025-io-uring#6](https://github.com/zpp-2025-io-uring/seastar/discussions/6) for more details). The current results also aren't reliable, as we are testing it on a "standard" PC instead of an AWS machine - those results will come in soon and hopefully will be much more favourable for this backend. We managed to get some test runs wirh `rpc write` where the new backend is the best, however we are unable to reproduce them, most likely due to the limitations of the machine we are using for testing. Here are the results:

Thanks for the clarification!

When we say that the async workers should be put on the "networking cores" we refer to the cores that are deidcated to network IRQs in a "sq" type of deployment, or something else?

avikivity:
> > @travisdowns there are currently some performance issues regarding this backend, e.g. related to the use of `SQPOLL` (see [zpp-2025-io-uring#6](https://github.com/zpp-2025-io-uring/seastar/discussions/6) for more details). The current results also aren't reliable, as we are testing it on a "standard" PC instead of an AWS machine - those results will come in soon and hopefully will be much more favourable for this backend. We managed to get some test runs wirh `rpc write` where the new backend is the best, however we are unable to reproduce them, most likely due to the limitations of the machine we are using for testing. Here are the results:
> 
> Thanks for the clarification!
> 
> When we say that the async workers should be put on the "networking cores" we refer to the cores that are deidcated to network IRQs in a "sq" type of deployment, or something else?

Yes, we're replacing the "network IRQ cores" (which are also called just "networking cores" here with the kernel-side SQ polling. This hopfully results in the entire TCP stack running on a single core (per group for reactors) rather than on both the reactor and the networking core.

src/core/reactor_backend.cc
pixelkubek: We have the following propositions of implementations of the suggested solutions. Do we correctly understand the possible solutions and address them with the following propositions? Which one of them should we pursue?

# IOSQE_IO_LINK
We allocate one large buffer ring with many `8kB` buffers.
For each operation we will submit `16` SQEs with `IOSQE_BUFFER_SELECT`.
They will be linked with `IOSQE_IO_LINK` - first with unset `MSG_DONTWAIT`, next with set `MSG_DONTWAIT` so we receive up to N successful completions. If there is B data, we will receive ceil(B/N) completions, so that we don't waste memory.
This benefits from the buffer rings, as we can schedule more SQE than we have buffers in the ring. 
With this approach, however, we might receive `-ENOBUFS`. What should happen then? 
Is it acceptable for SQEs with `MSG_DONTWAIT` to fail because of that?
Can the SQE without `MSG_DONTWAIT` fail because of that? We considered, in that case, e.g. reissuing the request, but with the first SQE having a normally allocated buffer.

Because `read_some` and `recv_some` return a `future<temporary_buffer<char>>`, which requires contiguous memory,
we only use the first (for a given fd) SQE to resolve the future and store the buffers from the other successful CQEs in 
`pollable_fd_state`. Future calls to `read_some` and `recv_some` will first check the list of buffers inside `pollable_fd_state` and,
if there are any, we pop the first and return as a ready future. In this way `128kB` of data can be handled in a single reactor loop, but it will be spread into 16 calls to `recv_some`/`read_some`.

# IORING_RECVSEND_BUNDLE
This approach is similar to the previous one, but instead of `16` SQEs linked together, we post
one with `IORING_RECVSEND_BUNDLE`. This might be better, as it prevents unnecessary calls to the kernel networking stack,
but we couldn't find a way to limit the amount of data received for a given request to `128kB`.

# Scaling like `posix_data_source_impl::get()`
We have some questions regarding this approach:
- Should we implement separate scaling logic, or utilize the existing implementation in `posix_data_source_impl::get()`
- If we have buffer pools of sizes `8kB` to `128kB`, what should the number of buffers in each pool be? Should it be constant or dynamic? What should we do if we run out of buffers is one of the pools?
- With this approach, are buffer rings beneficial?

MrD4rkne: for (hopefully) easier comparison:

| **Basic Idea** | Submit 16 linked SQEs using `IOSQE_IO_LINK` and `IOSQE_BUFFER_SELECT` | Submit a single SQE using `IORING_RECVSEND_BUNDLE` | Have different buffer rings for sizes 8kB ~~ 128 kB and rotate between them |
|--------|---------------|------------------------|--------------------------------------------|
| **Submission Model** | Multiple SQEs (linked chain per request) | Single SQE per request | Single SQE per request, picking the buffer from the appropriate pool |
| **SQEs** | Multiple kernel interactions (one per SQE) | Potentially fewer kernel interactions, less roundtrips to kernel networking stack | 1 SQE |
| **Buffer Strategy** | One large buffer ring with many 8kB buffers | One large buffer ring with many 8kB buffers | Multiple buffer pools (e.g., 8kB–128kB) |
| **Returned Memory Layout** | Multiple buffers | Multiple buffers | Contiguous buffer |
| **Future Resolution** | Only first CQE resolves `future<temporary_buffer<char>>`; others cached | Only first CQE resolves `future<temporary_buffer<char>>`; others cached | Directly returns appropriately sized buffer |
| **Extra Data Handling** | Remaining CQEs stored in `pollable_fd_state` | Remaining CQEs stored in `pollable_fd_state` | As it is |
| **Efficiency** | ± Overhead from multiple SQEs vs parallelism benefits | + Fewer kernel calls, better batching | + Efficient if scaling tuned well<br>− Risk of mismatch between workload and pool sizes |
| **Control Over Read Size** | + Explicit (e.g., 16 × 8kB = 128kB) | − Limited (no clear way to cap at 128kB) | + Full control via buffer sizing logic |
| **Complexity** | − High (linking, CQE management, caching) | + Medium (CQE management, caching) | − High (requires dynamic scaling logic and tuning) |
| **Flexibility to workload** | + High (can scale number of SQEs dynamically) | + High (kernel handles batching) | − Low (risk of running out of buffers for certain sizes) |
| **Buffers** | In buffer ring, need the logic of releasing / allocating back the buffers for the pool not to be exhausted | In buffer ring, need the logic of releasing / allocating back the buffers for the pool not to be exhausted | Explicit pool management strategy, buffer rings don't bring value here |
| **Open Questions** | What's the impact for having 16x SQEs? | Limiting receive size? Do we need to cap it? How if so? | Pool sizing, dynamic vs static, buffer exhaustion handling |
| **Overall Pros (+)** | + Fine-grained control<br>+ Predictable max read size<br>+Might be scalable even more by adjusting buffer size / number of SQEs | + Simpler model<br>+ Fewer trips to kernel networking stack<br>+ Potentially better kernel efficiency | + Clean API (contiguous buffers)<br>+ Full control over sizing |
| **Overall Cons (−)** | − Complex implementation<br>− A lot of SQEs, we don't know how it affects performance, especial SQ_POLL<br>− Cached buffers might never be read<br>− Logic of adding back buffers to the pool, handling buffer exhaustion | − Limited control over read size<br>− Less predictable behavior<br>− Cached buffers might never be read<br>− Logic of adding back buffers to the pool, handling buffer exhaustion | − Complex tuning<br>− Fragmentation across pools<br>− Risk of buffer exhaustion |

pixelkubek: In the current implementation, it is only used as a fallback, if the buffer ring cannot be used

src/core/reactor_backend.cc
pixelkubek: For operations utilizing the buffer ring, we need to check the CQE's `flags` field to get the buffer id

src/core/reactor_backend.cc
MrD4rkne: @avikivity

avikivity: IORING_RECVSEND_BUNDLE looks like how I thought it was working in the first place. They are ideal except that they're only available in 6.10, which is very recent. It's acceptable to start with.

Some more options:
Linked SQEs, but scale the buffers - first 8k, then 16k, then 32k, up to 128k. Reduces the number of buffer crossings.

Multishot SQEs - very dangerous, will result in consuming data from the socket even if the application side isn't making progress. This kills flow control (application slows down -> stops/slows consuming from buffer -> TCP window closes -> remote stops sending)

re the table - very good. Some comments:

Control Over Read Size - Limited (no clear way to cap at 128kB) - isn't the `len` parameter respected?

Extra Data Handling - might be better to store extra buffers in custom data_source_impl (but need careful consideration)

But I think we should optimize for simplicity and try to get something merged.

How about this:
1. Defer IOSQE_BUFFER_SELECT and IORING_RECVSEND_BUNDLE until post merge. They have too many interactions with the rest of the system.
2. Consider

```c++
future<temporary_buffer<char>>
posix_data_source_impl::get() {
    return _fd.recv_some(static_cast<internal::buffer_allocator*>(this)).then([this] (temporary_buffer<char> b) {
        if (b.size() >= _config.buffer_size) {
            _config.buffer_size *= 2;
            _config.buffer_size = std::min(_config.buffer_size, _config.max_buffer_size);
        } else if (b.size() <= _config.buffer_size / 4) {
            _config.buffer_size /= 2;
            _config.buffer_size = std::max(_config.buffer_size, _config.min_buffer_size);
        }
        bytes_received[sg_id] += b.size();
        return b;
    });
}
```

It already has the buffer scaling logic.

include/seastar/core/internal/pollable_fd.hh
mikolajbilski: Introducing a new class was a design choice on our part, we have decided that a new class would be cleaner than changing all the functions that use speculations to have the ability to disable the speculations.

mikolajbilski:
This force-push changes the following:
- rebase onto the newest `master`
- make `reactor_backend_uring` inherit from `reactor_backend_uring_base` right from the commit where the latter is introduced
- remove commits implementing the buffer rings support, as they have been deprioritized until a suitable implementation is found
- add implementation of `writev()`, which has since appeared as a `virtual` method in `reactor_backend`

src/core/reactor_backend.cc
```diff
@@ -1820,63 +1939,443 @@ class reactor_backend_uring final : public reactor_backend {
...
+
+unsigned get_uring_group_id(seastar::shard_id shard_id, const resource::cpuset& worker_cpus) noexcept {
+    return shard_id % worker_cpus.size();
+}
+
+class buf_group_io_completion final: public io_completion {
+    std::optional<temporary_buffer<char>> _buffer_opt;
+    promise<temporary_buffer<char>> _result;
+public:
+    buf_group_io_completion(pollable_fd_state& fd) {}
```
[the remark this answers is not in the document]
mikolajbilski: Buffer rings have been disabled in the recent force-push, this is now irrelevant

include/seastar/core/internal/io_request.hh
mikolajbilski: Buffer rings have been disabled in the recent force-push, and therefore this PR doesn't touch `io_request.hh` now. The internal WIP version of buffer rings has that fixed

src/core/reactor_backend.cc
mikolajbilski: Buffer rings have been removed in the recent force-push, keeping this open as this discussion will still be relevant when we try to find a suitable implementation

src/core/reactor_backend.cc
mikolajbilski: This has been addressed in the recent force-push

src/core/reactor_backend.cc
mikolajbilski: This is now irrelevant as buffer rings have been disabled in the recent force-push

witek-formanski:
This `force-push` changes the following:
- rebase onto `master`
- change usage of now-deprecated `smp::count` to `_shard_count`
- squash `Add Submission Queue polling` into `Implement asymmetric uring creation functions`
- add dynamic cpuset allocation to support machines with more than `CPU_SETSIZE` (1024) cores

src/core/reactor_backend.cc
```diff
@@ -1300,7 +1300,7 @@ class reactor_backend_uring final : public reactor_backend {
 
     hrtimer_completion _hrtimer_completion;
     smp_wakeup_completion _smp_wakeup_completion;
-private:
+
```
avikivity: We often use these as a separator between data and functions.

src/core/reactor_backend.cc
```diff
@@ -1222,7 +1222,14 @@ detect_io_uring() {
     return bool(ring_opt);
 }
 
-class reactor_backend_uring final : public reactor_backend {
+class reactor_backend_uring_base : public reactor_backend {
+    public:
```
avikivity: Bad indentation.

src/core/reactor_backend.cc
```diff
@@ -1222,7 +1222,14 @@ detect_io_uring() {
...
+    reactor_backend_uring_base(uses_blocking_io blocking_io, supports_aio_fdatasync aio_fdatasync)
+        : reactor_backend(blocking_io, aio_fdatasync)
+    {}
+};
+
+class reactor_backend_uring final : public reactor_backend_uring_base {
```
avikivity: I don't see what this base class does.

src/core/reactor_backend.cc
```diff
@@ -1223,15 +1223,10 @@ detect_io_uring() {
 }
 
 class reactor_backend_uring_base : public reactor_backend {
```
avikivity, line 1413: > Move common methods for io_uring based backends into the base class

This should be part of the commit that introduces the base class, otherwise it does nothing.

src/core/reactor_backend.cc
```diff
@@ -1937,6 +1937,88 @@ class reactor_backend_uring final : public reactor_backend_uring_base {
...
+        auto desc = std::make_unique<read_completion_base>(ba->allocate_buffer());
+        auto req = internal::io_request::make_recv(fd.fd.get(), desc->get_write(), desc->get_size(), 0);
+        return submit_request(std::move(desc), std::move(req));
+    }
+
+    virtual future<size_t> writev(pollable_fd_state& fd, std::span<iovec> iovs) override {
+        auto desc = std::make_unique<sized_promise_completion_base>();
+        const uint64_t position_file_offset = -1;
+        std::vector<iovec> iov;
+        iov.assign(iovs.begin(), iovs.end());
```
avikivity: Using the constructor seems simpler.

include/seastar/core/reactor_config.hh
```diff
@@ -21,12 +21,17 @@
...
+#include <any>
 #include <seastar/util/program-options.hh>
 #include <seastar/util/memory_diagnostics.hh>
 #include <seastar/core/scheduling.hh>
 
 namespace seastar {
 
+// Valid type when compiling without SEASTAR_HAVE_URING.
+// This avoids changing the size of a struct which contains it depending on conditional compilation.
+using compile_safe_io_uring = std::any;
```
avikivity: I don't understand the comment about changing the struct size. Why do we care?

Seastar is source-compatible only, not binary compatible.

src/core/reactor_backend.cc
```diff
@@ -1937,10 +1938,135 @@ class reactor_backend_uring final : public reactor_backend_uring_base {
...
+/// Helper functions that manage the lifecycle and configuration of asymmetric io_uring backend
+/// Handles CPU allocation, worker thread management, and backend creation
+namespace uring {
+
+std::optional<::io_uring>
+try_create_asymmetric_uring_impl(::io_uring_params params, bool throw_on_error) {
+    auto required_features =
+            IORING_FEAT_SUBMIT_STABLE
+            | IORING_FEAT_NODROP
+            | IORING_FEAT_SQPOLL_NONFIXED;
```
avikivity: Need IORING_FEAT_FAST_POLL, or it will be slower than the symmetric variant.

src/core/reactor_backend.hh
```diff
@@ -380,12 +384,21 @@ public:
...
+try_create_attached_asymmetric_uring(int uring_fd, bool throw_on_error);
+
+std::optional<::io_uring>
+try_create_base_asymmetric_uring(unsigned worker_cpu, bool throw_on_error);
+
 // QUEUE_LEN is more or less arbitrary. Too low and we'll be
 // issuing too small batches, too high and we require too much locked
 // memory, but otherwise it doesn't matter.
 inline constexpr unsigned QUEUE_LEN = 200;
+inline constexpr std::chrono::milliseconds POLLER_SLEEP_TIMEOUT(2000);
```
avikivity: That's way too much, let's try 5.

src/core/reactor.cc
```diff
@@ -4688,12 +4752,28 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
         }
     };
 
+    auto master_uring_fds = std::make_shared<std::vector<int>>(_shard_count, -1);
+
+    auto reactor_config = reactor_cfg;
+
+#ifdef SEASTAR_HAVE_URING
+    if (reactor_opts.reactor_backend.get_selected_candidate().name() == "asymmetric_io_uring") {
```
avikivity: Ugh, add a virtual function to reactor_backend_selector().

src/core/reactor.cc
```diff
@@ -4688,12 +4752,28 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
...
+        using namespace uring;
+        const bool is_master = is_master_shard(0, async_worker_cpus);
+        const unsigned uring_group_id = get_uring_group_id(0, async_worker_cpus);
+        if (is_master) {
+            reactor_config.asymmetric_uring.emplace<compile_safe_io_uring>(try_create_base_asymmetric_uring(select_worker_cpu(0, async_worker_cpus), true).value());
+            (*master_uring_fds)[uring_group_id] = std::any_cast<::io_uring>(std::get<compile_safe_io_uring>(reactor_config.asymmetric_uring)).ring_fd;
```
avikivity: If all you need is the ring fd, then make it an int and store the fd there, no need for std::any.

src/core/reactor.cc
```diff
@@ -4723,7 +4803,28 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
...
+#ifdef SEASTAR_HAVE_URING
+            if (reactor_opts.reactor_backend.get_selected_candidate().name() == "asymmetric_io_uring") {
+                using namespace uring;
+                const bool is_master = is_master_shard(i, async_worker_cpus);
+                const unsigned uring_group_id = get_uring_group_id(i, async_worker_cpus);
+                if (is_master) {
+                    reactor_config.asymmetric_uring.emplace<compile_safe_io_uring>(try_create_base_asymmetric_uring(select_worker_cpu(i, async_worker_cpus), true).value());
+                    (*master_uring_fds)[uring_group_id] = std::any_cast<::io_uring>(std::get<compile_safe_io_uring>(reactor_config.asymmetric_uring)).ring_fd;
+                }
+
```
avikivity: Extract this into a function and avoid duplication.

avikivity:
Looks good, I posted some minor comments.

Patch changelogs need to specify their scope, use `git log` to see how other patches specify the area they work on.

include/seastar/core/reactor_config.hh
pixelkubek: Originally, we had
```cpp
#ifdef SEASTAR_HAVE_URING
std::optional<std::variant<int, ::io_uring>> asymmetric_uring;
#endif
```
inside `struct reactor_config`.

Because `reactor_config _cfg;` is stored inside the reactor class, each file that ultimately includes `reactor.hh` would see a different (with respect to struct size) definition of `class reactor` depending on whether it was compiled with `-DSEASTAR_HAVE_URING`.

This caused some tests to fail when testing Scylla compiled with our Seastar branch, as Seastar was compiled with `-DSEASTAR_HAVE_URING`, but Scylla files were not.

This was not a problem before adding the asymmetric backend, because there were no uses of `#ifdef SEASTAR_HAVE_URING` in header files.

avikivity: Everything has to be built with the same SEASTAR_HAVE_URING. It's defined by cmake and propagated to all consumers. No need to contort the code.

pixelkubek:
This force-push introduces the following changes:
- Add comment to the `reactor_backend_uring_base` class.
- Re-add removed `private` specifier.
- Change `POLLER_SLEEP_TIMEOUT` to 5ms.
- Add `IORING_FEAT_FAST_POLL` check when creating an asymmetric io_uring.
- Return `std::nullopt` after an unsuccessful `CPU_ALLOC` allocation.
- Whitespace changes.
- Reword "for the SMP" into "available to the SMP shards".
- Correct `we shutting down` to `we are shutting down`.
- Squash commits that introduce the asymmetric backend and extract common functions
- Remove the `compile_safe_uring` workaround.
- Modify CMake to export `SEASTAR_HAVE_URING`.
- Move uring assignment logic from `smp::configure` to a `reactor_backend_configurator` class
- Rebase onto a newer version of `master`

src/core/reactor_backend.cc
pixelkubek: The commits are squashed in the latest force-push.

include/seastar/core/reactor_config.hh
pixelkubek: This workaround is removed in the latest force push.
At the same time, CMake is modified to propagate `SEASTAR_HAVE_URING` to consumers,
instead of adding it to `Seastar_PRIVATE_COMPILE_DEFINITIONS`.

src/core/reactor_backend.cc
pixelkubek: Added in the latest force-push.

src/core/reactor.cc
pixelkubek: The newest force-push introduces a new `reactor_backend_configurator` class with virtual methods.
The logic for uring assignment and creation is now removed from `smp::configure` and encapsulated in its subclass `asymmetric_uring_reactor_backend_configurator`.

The reason why virtual functions were not added directly to `reactor_backend_selector` is because it would require changes everywhere where
the backend selector is stored or returned by value to use some kind of pointer.
This causes for example the output of `reactor_backend_selector::available()` to no longer be printable, which the `reactor_options` constructor relies on.
I believe that the implementation in the recent force-push reduces unneeded code changes.

src/core/reactor.cc
pixelkubek: For shards which create attached urings (with `IORING_SETUP_ATTACH_WQ`) we store the `ring_fd` to attach to.
For other shards, we store the whole `struct io_uring`.
`std::any` was needed for the `compile_safe_uring` workaround and has been removed in the latest force-push.

src/core/reactor.cc
witek-formanski: Done

src/core/reactor_backend.hh
witek-formanski: Done

src/core/reactor_backend.cc
witek-formanski: Done

src/core/reactor_backend.cc
witek-formanski: Added a comment

src/core/reactor_backend.cc
witek-formanski: Already changed

src/core/reactor_backend.cc
witek-formanski: Re-added after prior removal

witek-formanski:
This `force-push` introduces the following changes:
- specify the scope in commit messages
- remove an unused declaration

Deexie:
@avikivity @xemul please re-review

avikivity:
There's a conflict, please resolve it. Make sure to look at recent changes to the original io_uring backend so we don't open a gap.

pixelkubek:
This `force-push` indroduces the following changes:
- rebase onto a newer version of `master`
- add scope to the `Extract io_uring queue length` commit

After looking at the recent changes to the `io_uring` backend, we did not find that any of them required additional changes to our code - we don't use `smp::count` and the changes from `fix pollable_fd_state_completion reuse after co_await in io_uring` are present in `reactor_backend_uring_base`

MrD4rkne:
@avikivity @xemul  should you have time, please re-review

PR #3299, opened 2026-03-19
input_stream: fix post-EOF re-entry into data source

kreuzerkrieg:
input_stream: fix post-EOF re-entry into data source

input_stream::read_exactly(), read_exactly_part(), and skip()
could call into the underlying data source after it had
already returned an empty buffer (EOS signal).

In read_exactly(), when _buf was empty and _eof was already
set from a previous call, the code fell through to _fd.get()
instead of returning an empty buffer immediately.

The same issue existed in read_exactly_part(): it lacked an
_eof guard at the entry point, so callers reaching it with
_eof already set would enter the loop and call _fd.get() on
an exhausted source.

In skip(), when _buf was empty and _eof was set, the code
would call _fd.skip() on an exhausted source.

Fix all three by checking _eof before calling into the
source.

Add a strict_source_impl test helper that asserts if get()
is called after EOS, and a randomized test that builds a
shuffled sequence of read/read_up_to/read_exactly/skip
operations, drains a 5 MB stream while verifying data
integrity, then replays the same sequence post-EOF to
confirm nothing gets stuck.

### DIFF
--- a/include/seastar/core/iostream-impl.hh
+++ b/include/seastar/core/iostream-impl.hh
@@ -169,6 +169,9 @@ future<> output_stream<CharType>::write(scattered_message<CharType> msg) noexcep
 template <typename CharType>
 future<temporary_buffer<CharType>>
 input_stream<CharType>::read_exactly_part(size_t n) noexcept {
+    if (_eof) {
+        co_return temporary_buffer<CharType>();
+    }
     temporary_buffer<CharType> out(n);
     size_t completed{0U};
     while (completed < n) {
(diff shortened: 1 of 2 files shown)


xemul:
The fix looks OK, but the test is not very nice. It should be exhaustive combinatorial test, that iterates over all possible combinations of input and checks some invariants of the output. E.g. check the test_splitting_invariants one.

For input stream, I'd prompt AI like this:
- there's an input source that generates N buffers of the same size S
- there's a user that has `read + skip + { read_up_to, read_exactly } * {s1, s2, ..., sN}` operations that are tested against the input stream in all possible combinations. Sizes s1, ..., sN are probably enough to be from [1, S*3) range

The test must ensure that

- read_up_to and read_exactly contract is respected
- No bytes are implicitly skipped by stream
- No get-s past EOF
- ... maybe something else

xemul:
The data_source.get() contract is -- it's not called after EOF
I don't know if we want to support "read from input_stream past EOF is, again, EOF" contract though

avikivity:
> The data_source.get() contract is -- it's not called after EOF I don't know if we want to support "read from input_stream past EOF is, again, EOF" contract though

I don't think we need to.

The user should recognize EOF and stop reading.

kreuzerkrieg:
> > The data_source.get() contract is -- it's not called after EOF I don't know if we want to support "read from input_stream past EOF is, again, EOF" contract though
> 
> I don't think we need to.
> 
> The user should recognize EOF and stop reading.

ok then

Issue #935, opened 2021-08-25
Report the io properties (io_properties.yaml) as metrics


amnonh:
It could be useful if the reactor would expose the io properties as metrics.
It can be done only on shard 0, no need to report it for each of the shards.

Relates to: https://github.com/scylladb/scylla-monitoring/issues/1382

amnonh:
I came back to the original https://github.com/scylladb/scylla-monitoring/issues/1382 issue, and I have double-guessed myself.
True, we want to get the information from the node, but does it make sense to report it as a metric?
It never changes during a run, we just need an easy way to report it

PR #3300, opened 2026-03-19
reactor: add get_all_io_queues() method

travisdowns:
## Summary
- Move `get_io_queue` and `try_get_io_queue` out of the header into reactor.cc — no reason for these to be inline
- Add `get_all_io_queues()` returning `std::vector<io_queue*>` for all io_queues on the current shard

This method doesn't expose any additional info beyond what is already exposed, but prevents the need of trying arbitrary devid to extract it.

### DIFF
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -487,19 +487,8 @@ public:
         return now() - _start_time;
     }
 
-    io_queue& get_io_queue(dev_t devid = 0) {
-        auto queue = _io_queues.find(devid);
-        if (queue == _io_queues.end()) {
-            return *_io_queues.at(0);
-        } else {
-            return *(queue->second);
-        }
-    }
-
-    io_queue* try_get_io_queue(dev_t devid) noexcept {
-        auto queue = _io_queues.find(devid);
-        return queue != _io_queues.end() ? queue->second.get() : nullptr;
-    }
+    io_queue& get_io_queue(dev_t devid = 0);
+    io_queue* try_get_io_queue(dev_t devid) noexcept;
 
     std::string_view get_backend_name() const;
 
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -490,6 +490,9 @@ public:
     io_queue& get_io_queue(dev_t devid = 0);
     io_queue* try_get_io_queue(dev_t devid) noexcept;
 
+    /// Returns pointers to all io_queues on this shard.
+    std::vector<io_queue*> get_all_io_queues();
+
     std::string_view get_backend_name() const;
 
 private:
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -19,6 +19,7 @@
  * Copyright 2014 Cloudius Systems
  */
 
+#include <algorithm>
 #include <atomic>
 #include <chrono>
 #include <cmath>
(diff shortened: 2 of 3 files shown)


src/core/reactor.cc
```diff
@@ -247,6 +247,15 @@ io_queue* reactor::try_get_io_queue(dev_t devid) noexcept {
     return queue != _io_queues.end() ? queue->second.get() : nullptr;
 }
 
+std::vector<io_queue*> reactor::get_all_io_queues() {
+    std::vector<io_queue*> queues;
+    queues.reserve(_io_queues.size());
+    for (auto& [_, q] : _io_queues) {
+        queues.push_back(q.get());
```
xemul: io_queue-s can be shared between two distinct dev-s (#2738), so the resulting vector can have duplicates

xemul:
What's the intented usage of the exported vector?

travisdowns:
> What's the intented usage of the exported vector?

I can imagine various, but in Redpanda specifically we get the io_queues to export additional metrics such as the io queue configuration settings.

src/core/reactor.cc
travisdowns: Thanks, I'm fixing this.

From that change:

> It's nonetheless possible to have one physical disk be "split" into several virtual block devices, and to mount each into its own mountpoint. In order to configure io-scheduler for this setup the only option is to have two mountpoints in io_properties.yaml. With that, seastar will create two io_queues for both mountpoints, but it's not correct, since underlying physical disk is still the same.

By "virtual block devices" are you including things like vanilla partitions?

travisdowns: I de-duplicate the list during creation now. Just do a linear search in the vector as the number of queues is modest.

travisdowns: Added a basic test also.

xemul: > By "virtual block devices" are you including things like vanilla partitions?

It all started with LLVM support when one disk was split into two then each was mounted separately. But classical partitions would work just the same

xemul:
> > What's the intented usage of the exported vector?
> 
> I can imagine various, but in Redpanda specifically we get the io_queues to export additional metrics such as the io queue configuration settings.

Like [io-properties](https://github.com/scylladb/seastar/issues/935) values (and derived)?

I'm afraid that exporting io-queue-s as public API will bring us another problem. Nowadays, the `class io_queue` (as well as `class fair_queue`) is not considered to be the part of public API. Thus, any changes to it can be straghtforward and not care about being backward compatibe. With thie new method, every future change io_queue class will need to be thought of twise :(

Thoughts?

include/seastar/core/reactor.hh
```diff
@@ -487,19 +487,11 @@ public:
...
-    }
+    io_queue& get_io_queue(dev_t devid = 0);
+    io_queue* try_get_io_queue(dev_t devid) noexcept;
 
-    io_queue* try_get_io_queue(dev_t devid) noexcept {
-        auto queue = _io_queues.find(devid);
-        return queue != _io_queues.end() ? queue->second.get() : nullptr;
-    }
+    /// Returns pointers to all io_queues on this shard.
+    std::vector<io_queue*> get_all_io_queues();
```
xemul, line 494: Why not references? Looks more "idiomatic", as no nullptrs can be there

travisdowns: > Why not references? Looks more "idiomatic", as no nullptrs can be there

Because containers can't hold references. Or do you mean reference_wrapper? Happy to make that change.

xemul: > Because containers can't hold references.

:flushed: 

> Or do you mean reference_wrapper? Happy to make that change.

No I mistakenly meant references, but pointer is better than reference wrapper

travisdowns:
> Like [io-properties](https://github.com/scylladb/seastar/issues/935) values (and derived)?
> 
> I'm afraid that exporting io-queue-s as public API will bring us another problem. Nowadays, the `class io_queue` (as well as `class fair_queue`) is not considered to be the part of public API. Thus, any changes to it can be straghtforward and not care about being backward compatibe. With thie new method, every future change io_queue class will need to be thought of twise :(
> 
> Thoughts?

There were already exposed via get_io_queue and related methods, so I they are already public AFACT. This only adds a convenience method to get them all at once instead of passing in device IDs to get them one at a time (the use case is we want to expose the config properties for all io queues, without needing to figure out the dev id to pass). So I assumed we are not increasing the public surface area here.

PR #3302, opened 2026-03-21
http: Fix hex_to_byte() to only accept valid hex digits (a-f, A-F)

### LINKED ISSUE #3303: hex_to_byte with invalid hex.
check used 'a'-'z' and 'A'-'Z', which incorrectly accepted invalid hex characters like 'g', 'z', etc. and produced wrong results.

https://github.com/scylladb/seastar/blob/master/src/http/url.cc#L33-L40

cuiweixie:
The previous range check used 'a'-'z' and 'A'-'Z', which incorrectly accepted invalid hex characters like 'g', 'z', etc. and produced wrong results. Restrict to valid hex digit range.

Closes #3303

### DIFF
--- a/src/http/url.cc
+++ b/src/http/url.cc
@@ -30,21 +30,23 @@ namespace internal {
 
 namespace {
 
-short hex_to_byte(char c) {
-    if (c >='a' && c <= 'z') {
-        return c - 'a' + 10;
-    } else if (c >='A' && c <= 'Z') {
-        return c - 'A' + 10;
-    }
-    return c - '0';
+// https://url.spec.whatwg.org/#percent-encoded-bytes — a percent-encoded byte is
+// '%' followed by two ASCII hex digits; otherwise the '%' is left in the output.
+
+bool is_ascii_hex_digit(char c) noexcept {
+    return (c >= '0' && c <= '9') ||
+           (c >= 'A' && c <= 'F') ||
+           (c >= 'a' && c <= 'f');
 }
 
-/**
- * Convert a hex encoded 2 bytes substring to char
- */
-char hexstr_to_char(std::string_view in, size_t from) {
-
-    return static_cast<char>(hex_to_byte(in[from]) * 16 + hex_to_byte(in[from + 1]));
+unsigned hex_digit_value(char c) noexcept {
+    if (c <= '9') {
+        return static_cast<unsigned>(c - '0');
+    }
+    if (c <= 'F') {
+        return static_cast<unsigned>(c - 'A' + 10);
+    }
+    return static_cast<unsigned>(c - 'a' + 10);
 }
 
 bool should_encode(char c) {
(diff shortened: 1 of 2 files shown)


nyh:
> The previous range check used 'a'-'z' and 'A'-'Z', which incorrectly accepted invalid hex characters like 'g', 'z', etc. and produced wrong results. Restrict to valid hex digit range.

Please open an issue and/or write a regression test on what is the impact of this problem. What do you mean "wrong results"?

Moreover, what you patch does is that for the silly non-hex character "z" it returns
```
return c - '0';
```
instead of before your patch
```
return c - 'a' + 10;
```
Aren't both equally wrong? Why is one less wrong than the other?

I absolutely agree with you that the 'z' in the old code was silly and wrong, but I think the new code is equally (though differently) wrong.

According to https://url.spec.whatwg.org/#percent-encoded-bytes maybe the really right thing to do here is to leave "%xy" - where "xy" are not valid hex characters - as just "%xy", unchanged.

Again, I think you should create an issue explaining what is the **wrong** behavior you see, and what is the **right** behavior you expect - and only then fix the issue.

cuiweixie:
> > The previous range check used 'a'-'z' and 'A'-'Z', which incorrectly accepted invalid hex characters like 'g', 'z', etc. and produced wrong results. Restrict to valid hex digit range.
> 
> Please open an issue and/or write a regression test on what is the impact of this problem. What do you mean "wrong results"?
> 
> Moreover, what you patch does is that for the silly non-hex character "z" it returns
> 
> ```
> return c - '0';
> ```
> 
> instead of before your patch
> 
> ```
> return c - 'a' + 10;
> ```
> 
> Aren't both equally wrong? Why is one less wrong than the other?
> 
> I absolutely agree with you that the 'z' in the old code was silly and wrong, but I think the new code is equally (though differently) wrong.
> 
> According to https://url.spec.whatwg.org/#percent-encoded-bytes maybe the really right thing to do here is to leave "%xy" - where "xy" are not valid hex characters - as just "%xy", unchanged.
> 
> Again, I think you should create an issue explaining what is the **wrong** behavior you see, and what is the **right** behavior you expect - and only then fix the issue.

send a new version.

nyh:
> Looks correct to me, but @xemul you wrote the original code being fixed here, can you please also take a look? Thanks.

Sorry, misattributed the original code to @xemul, who just moved it. Apparently the old code is very old code, from 2015 (commit 6512d7f7fb08a4b8b083be1f0600556bb7c40064) from @amnonh. So @amnonh maybe you want to take a look.

src/http/url.cc
amnonh, line 85: in practice this means that we no longer validate that something that starts with % is a valid encoded.
this could be fine, but if the function does no validation, we should probably remove the return type as it does nothing.

src/http/url.cc
```diff
@@ -30,21 +30,23 @@ namespace internal {
...
- */
-char hexstr_to_char(std::string_view in, size_t from) {
+bool is_ascii_hex_digit(char c) noexcept {
+    return (c >= '0' && c <= '9') ||
+           (c >= 'A' && c <= 'F') ||
+           (c >= 'a' && c <= 'f');
+}
 
-    return static_cast<char>(hex_to_byte(in[from]) * 16 + hex_to_byte(in[from + 1]));
+unsigned hex_digit_value(char c) noexcept {
```
amnonh, line 42: you changed the type from char, but you didn't save much, in the code you still need a char

src/http/url.cc
```diff
@@ -65,11 +67,12 @@ bool decode(bool replace_plus, std::string_view in, sstring& out) {
     sstring buff(in.length(), 0);
     for (size_t i = 0; i < in.length(); ++i) {
         if (in[i] == '%') {
```
amnonh, line 69: now you can combine those two ifs and drop buff[pos++] = '%';  for the default behaviour

src/http/url.cc
nyh: I agree, we need to decide whether - if we see a "bad" `%` - do we continue decoding after it, or don't do any decoding on the entire parameter. I propose that we silently continue to decode as much as we can, and don't avoid decoding and certainly not throwing an error or anything. So I agree, url_decode() should stop returning a bool - it should be changed to return void.

Note that nominally, until now url_decode() was bool, the **only case** it could ever return false is if the '%' character was the last or before-last character in the string. In nother case (including `%xy`) would it return false. So the fact that url_decode() was bool was never useful anyway.

So I also think we should change in this patch url_decode() to be void.

PR #3305, opened 2026-03-23
rpc: include exception type for remote errors

gusev-p:
Sometimes custom exceptions inherit from std::exception but don't override what(). In this case, only the generic "std::exception" message is sent to the RPC client, making it difficult to diagnose what happened.


gusev-p:

gusev-p:
@avikivity ping

gusev-p:
@avikivity another gentle ping

PR #3306, opened 2026-03-23
http: deprecate reply::done(), remove _response_line dead field

m-szymon:
reply::done() was originally intended as a required finalisation step that cached the response line into _response_line. However _response_line was never read back — write_reply() always called response_line() directly. The done(content_type) overload was likewise redundant, doing nothing beyond set_content_type() internally.

- Remove the _response_line field entirely
- Make done() a no-op and mark it [[deprecated]]
- Make done(content_type) delegate to set_content_type() and mark it [[deprecated]]
- Fix write_body(content_type, sstring) in reply.cc which was calling done(content_type) instead of set_content_type() directly
- Replace all in-tree callers of done() / done(content_type) with set_content_type() or plain removal (routes.cc, httpd.cc, file_handler.cc, api_docs.hh, function_handlers.hh, httpd_test.cc)
- Add tests test_write_reply_content and test_write_reply_body_writer covering the full write_reply() output: response line, Content-Type, Content-Length / Transfer-Encoding, and body framing

### DIFF
--- a/include/seastar/http/reply.hh
+++ b/include/seastar/http/reply.hh
@@ -157,7 +157,6 @@ struct reply {
     size_t left_content_length = std::numeric_limits<size_t>::max();
     bool _skip_body = false;
 
-    sstring _response_line;
     std::unordered_map<sstring, sstring> trailing_headers;
     std::unordered_map<sstring, sstring> chunk_extensions;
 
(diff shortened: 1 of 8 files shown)


include/seastar/http/reply.hh
```diff
@@ -230,15 +229,15 @@ struct reply {
...
 
+    [[deprecated("Use set_content_type() or write_body() instead")]]
     reply& done(const sstring& content_type) {
-        return set_content_type(content_type).done();
+        return set_content_type(content_type);
     }
     /**
-     * Done should be called before using the reply.
-     * It would set the response line
+     * \deprecated done() is no longer required; remove the call.
```
nyh, line 237: nitpick: no need for a comment that just repeats the "deprecated" line below.

PR #3312, opened 2026-03-25
net: add set_listen_backlog() to update backlog on a live server socket

ScyllaPiotr:
- Add `set_listen_backlog(int)` to the `server_socket` class hierarchy, allowing the listen backlog to be changed on an already-listening socket
- On Linux, this re-invokes the `listen()` syscall on the underlying file descriptor, which is supported for bound sockets
- Also add `set_listen_backlog()` to `http_server`, which iterates all active listeners and updates their backlog

### DIFF
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -201,6 +201,7 @@ class posix_server_socket_impl : public server_socket_impl {
     socket_address _sa;
     int _protocol;
     pollable_fd _lfd;
+    int _listen_backlog;
     conntrack _conntrack;
     server_socket::load_balancing_algorithm _lba;
     shard_id _fixed_cpu;
@@ -210,23 +211,30 @@ public:
     explicit posix_server_socket_impl(int protocol, socket_address sa, pollable_fd lfd,
         server_socket::load_balancing_algorithm lba, shard_id fixed_cpu,
         bool proxy_protocol,
-        std::pmr::polymorphic_allocator<char>* allocator=memory::malloc_allocator) : _sa(sa), _protocol(protocol), _lfd(std::move(lfd)), _lba(lba), _fixed_cpu(fixed_cpu), _proxy_protocol(proxy_protocol), _allocator(allocator) {}
+        int listen_backlog,
+        std::pmr::polymorphic_allocator<char>* allocator=memory::malloc_allocator) : _sa(sa), _protocol(protocol), _lfd(std::move(lfd)), _listen_backlog(listen_backlog), _lba(lba), _fixed_cpu(fixed_cpu), _proxy_protocol(proxy_protocol), _allocator(allocator) {}
     virtual future<accept_result> accept() override;
     virtual void abort_accept() override;
     virtual socket_address local_address() const override;
+    virtual void set_listen_backlog(int backlog) override;
+    virtual int get_listen_backlog() const override;
 };
 
 class posix_reuseport_server_socket_impl : public server_socket_impl {
     socket_address _sa;
     int _protocol;
     pollable_fd _lfd;
+    int _listen_backlog;
     std::pmr::polymorphic_allocator<char>* _allocator;
 public:
     explicit posix_reuseport_server_socket_impl(int protocol, socket_address sa, pollable_fd lfd,
-        std::pmr::polymorphic_allocator<char>* allocator=memory::malloc_allocator) : _sa(sa), _protocol(protocol), _lfd(std::move(lfd)), _allocator(allocator) {}
+        int listen_backlog,
+        std::pmr::polymorphic_allocator<char>* allocator=memory::malloc_allocator) : _sa(sa), _protocol(protocol), _lfd(std::move(lfd)), _listen_backlog(listen_backlog), _allocator(allocator) {}
     virtual future<accept_result> accept() override;
     virtual void abort_accept() override;
     virtual socket_address local_address() const override;
+    virtual void set_listen_backlog(int backlog) override;
+    virtual int get_listen_backlog() const override;
 };
 
 class posix_network_stack : public network_stack {
--- a/include/seastar/net/stack.hh
+++ b/include/seastar/net/stack.hh
@@ -70,6 +70,13 @@ public:
     virtual future<accept_result> accept() = 0;
     virtual void abort_accept() = 0;
     virtual socket_address local_address() const = 0;
+    /// Update the listen backlog on an already-listening socket.
+    /// Throws std::system_error (EOPNOTSUPP) if the implementation
+    /// does not support runtime backlog changes.
+    virtual void set_listen_backlog(int backlog);
+    /// Get the currently configured listen backlog.
+    /// Returns -1 if the implementation does not track the backlog value.
+    virtual int get_listen_backlog() const;
 };
 
 class datagram_channel_impl {
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -427,6 +427,13 @@ future<> http_server::listen(socket_address addr) {
     lo.reuse_address = true;
     return listen(addr, lo);
 }
+
+void http_server::set_listen_backlog(int backlog) {
+    for (auto& listener : _listeners) {
+        listener.set_listen_backlog(backlog);
+    }
+}
+
 future<> http_server::stop() {
     future<> tasks_done = _task_gate.close();
     for (auto&& l : _listeners) {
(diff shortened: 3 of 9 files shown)


include/seastar/net/posix-stack.hh
```diff
@@ -214,6 +214,7 @@ public:
     virtual future<accept_result> accept() override;
     virtual void abort_accept() override;
     virtual socket_address local_address() const override;
+    virtual void set_listen_backlog(int backlog) override { _lfd.get_file_desc().listen(backlog); }
```
xemul: Shouldn't TLS socket also get its set_listen_backlog() override?

ScyllaPiotr: Good catch, added the forwarding override in `server_session` so TLS sockets delegate to the underlying socket.

avikivity: Please move the implementation to .cc to reduce include load.

include/seastar/net/stack.hh
```diff
@@ -70,6 +70,11 @@ public:
     virtual future<accept_result> accept() = 0;
     virtual void abort_accept() = 0;
     virtual socket_address local_address() const = 0;
+    /// Update the listen backlog on an already-listening socket.
+    /// The default implementation is a no-op; POSIX implementations
+    /// re-invoke the listen() syscall which Linux supports on a
+    /// bound socket.
+    virtual void set_listen_backlog(int backlog) {}
```
avikivity: Just like that?

include/seastar/net/posix-stack.hh
ScyllaPiotr: Done. Moved both `posix_server_socket_impl::set_listen_backlog` and `posix_reuseport_server_socket_impl::set_listen_backlog` implementations to `posix-stack.cc`, consistent with the other methods in those classes.

include/seastar/net/stack.hh
ScyllaPiotr: Changed the default to log a warning instead of silently no-op'ing, following the pattern in `native-stack-impl.hh` where unsupported operations (e.g. `Reuseaddr is not supported by native stack`) are logged. The implementation is now in `stack.cc`, keeping the header clean.

nyh:
@ScyllaPiotr there are merge conflicts.

nyh:
By the way, this is a single-patch commit so its cover letter gets ignored and isn't saved in the git log. If there is any important information in the cover letter you wish to save, please copy to the commit message of the single patch. Thanks.

include/seastar/net/stack.hh
```diff
@@ -70,6 +70,11 @@ public:
     virtual future<accept_result> accept() = 0;
     virtual void abort_accept() = 0;
     virtual socket_address local_address() const = 0;
+    /// Update the listen backlog on an already-listening socket.
+    /// The default implementation logs a warning; POSIX implementations
```
nyh: Nitpick: I don't think the superclass should document how the subclasses handle this API. The word "default implementation" is even misleading - the default socket is the Posix one, the superclass is not really "default", it's just a superclass that nobody should be using directly.

nyh (commented):
Looks good to me, but you are missing a test.

If you tested this externally (e.g., in ScyllaDB) it will be good to say so, but even better would be to add a unit test. Should be fairly easy - create a server with opts.listen_backlog = 42, confirm the listen_backlog is 42, then set_listen_backlog(67) and check that listen_backlog changed to 67.

By the way, do we actually have a function get_server_backlog() to use in that test? If we don't, this test will help you realize that we're missing this ;-)

ScyllaPiotr:
Addressed in 09606fecba9203d11918128d0836ccb953c2fc6b:

- Rebased and resolved conflicts
- Updated superclass API comment wording in server_socket_impl
- Added get_listen_backlog() plumbing (including TLS forwarding)
- Added unit test socket_listen_backlog_runtime_update_test that verifies runtime update (1111 -> 1888)
- Expanded single-commit message with important implementation details

Local verification:
./build/dev/seastar/tests/unit/socket_test --run_test=socket_listen_backlog_runtime_update_test -- -c 2 --reactor-backend linux-aio

@nyh please take another look when convenient.

ScyllaPiotr:
@avikivity please re-review

ScyllaPiotr:
> Addressed in [09606fe](https://github.com/scylladb/seastar/commit/09606fecba9203d11918128d0836ccb953c2fc6b):
> 
>     * Rebased and resolved conflicts
> 
>     * Updated superclass API comment wording in server_socket_impl
> 
>     * Added get_listen_backlog() plumbing (including TLS forwarding)
> 
>     * Added unit test socket_listen_backlog_runtime_update_test that verifies runtime update (1111 -> 1888)
> 
>     * Expanded single-commit message with important implementation details
> 
> 
> Local verification: ./build/dev/seastar/tests/unit/socket_test --run_test=socket_listen_backlog_runtime_update_test -- -c 2 --reactor-backend linux-aio
> 
> @nyh please take another look when convenient.

@nyh please re-review.

ScyllaPiotr:
Let's try to broadcast. @scylladb/seastar-maint please review.

src/net/stack.cc
```diff
@@ -227,6 +227,23 @@ socket_address server_socket::local_address() const noexcept {
...
+void server_socket::set_listen_backlog(int backlog) {
+    _ssi->set_listen_backlog(backlog);
+}
+
+int server_socket::get_listen_backlog() const {
+    return _ssi->get_listen_backlog();
+}
+
+void net::server_socket_impl::set_listen_backlog(int backlog) {
+    seastar_logger.warn("set_listen_backlog() not supported by this socket implementation");
```
xemul: Unsupported functionality usually throws, not ... well :) silently spoils logs

src/net/stack.cc
```diff
@@ -227,6 +227,23 @@ socket_address server_socket::local_address() const noexcept {
...
+}
+
+int net::server_socket_impl::get_listen_backlog() const {
+    seastar_logger.warn("get_listen_backlog() not supported by this socket implementation");
```
xemul: Similarly here -- spoiling logs is uncalled for. Returning -1 should be enough (if properly documented)

src/net/stack.cc
avikivity: The application doesn't know what stack it is using, so wouldn't expect the exception.

I think warning once is enough.

avikivity:
Force-push is not the correct way to interact with a maintainer.

ScyllaPiotr:
> Force-push is not the correct way to interact with a maintainer.

I'm marking stuff which I agree with and follow through on with :+1:, but I'll be more explicit in the future.

src/net/stack.cc
ScyllaPiotr: @xemul WDYT about warning once, like @avikivity suggested? Because I don't like it TBH.
1. Setting the backlog is unlikely to be done frequently, so why filter the warnings?
2. This would require careful wording, not to make someone think that only the 1st call failed.
3. It would be unexpected for an _nth_ call to [silently fail](https://en.wikipedia.org/wiki/Talk%3AUnix_philosophy#No_news_is_good_news).
4. One searching why a call failed to enact the expected backlog would need to see the source code to learn that logging is sparsed.

xemul: I just found out that in seastar network stack unsupported functionality actually doesn't throw :( to be more specific -- not always throws. E.g. the native_socket_impl::set_reuseaddr() just "silently spoils logs" :\  oh, well... Let's log the warning, maybe rate-limited-ly

src/http/httpd.cc
```diff
@@ -427,6 +427,13 @@ future<> http_server::listen(socket_address addr) {
     lo.reuse_address = true;
     return listen(addr, lo);
 }
+
+void http_server::set_listen_backlog(int backlog) {
+    for (auto& listener : _listeners) {
+        listener.set_listen_backlog(backlog);
```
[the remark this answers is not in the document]
nyh: @ScyllaPiotr I think copilot is onto something here :-( We are indeed using by default (and this is, for example, which ScyllaDB Alternator also uses in its HTTP server) this non-so-reuse implementation where listen() happens on just one socket, but if this class is NOT the one you modify, it will fail.

In your test, please make sure that an EOPNOTSUPP will cause the test to fail, and also run the test with more than one shard to make sure that posix_ap_server_socket_impl is used. Also please check that when the thread asked to change the listen limit sends this requests to all its sibling threads, we will report any EOPNOTSUPP from any of those other threads and don't just silently swallow it.

PR #3318, opened 2026-03-26
iovec: fix iovec_trim_front infinite loop on zero-length iovecs

travisdowns:
## Summary

- Fix `iovec_trim_front` to skip zero-length iovecs instead of looping infinitely when all remaining iovecs have zero length
- Add regression tests for zero-length iovec handling

Without this, some zero length writes could crash due to infinite recursion in `write_all`.

## Test plan

- [x] Unit tests added in `temporary_buffer_test.cc` covering single, multiple, and mixed zero-length iovecs
- [x] Tests pass on both api8 and api9 builds

### DIFF
--- a/include/seastar/util/internal/iovec_utils.hh
+++ b/include/seastar/util/internal/iovec_utils.hh
@@ -51,7 +51,7 @@ inline size_t iovec_len(std::span<const iovec> iov) {
 // Can update some iovecs in-place
 inline std::span<iovec> iovec_trim_front(std::span<iovec> iov, size_t size) {
     unsigned pos = 0;
-    while (pos < iov.size() && size > 0) {
+    while (pos < iov.size()) {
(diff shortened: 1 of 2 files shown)


travisdowns:
draft mode while I look into failures

xemul:
iostreams don't produce vector<iovec> with zero-length elements in, do they?

avikivity:
> iostreams don't produce vector with zero-length elements in, do they?

I think it's still reasonable to have the function defend against such legal-but-strange iovecs.

travisdowns:
@xemul wrote:

> iostreams don't produce vector with zero-length elements in, do they?

They don't but there are call paths that don't go through iostreams, like the TLS layer (which is where this came up), where they could (at least, OpenSSL documentation says zero length writes are possible and must be handled).

PR #3341, opened 2026-04-02
reactor: Deprecate add_task/add_urgent_task/add_high_priority_task

xemul:
These methods may have out-of-tree users and cannot be simply removed, but they should not be part of the public reactor API going forward. Deprecate them in favor of the seastar::schedule() and seastar::schedule_urgent() free functions.

Internal call sites in reactor.cc are wrapped with diagnostic pragmas to suppress the deprecation warnings. add_high_priority_task() is additionally exposed via reactor::test::add_high_priority_task() to prepare for the eventual privatization of the method, with the pragma applied there so test code remains clean.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -1115,9 +1115,12 @@ reactor::reactor(std::shared_ptr<seastar::smp> smp, alien::instance& alien, unsi
     auto r = ::pthread_sigmask(SIG_UNBLOCK, &mask, NULL);
     SEASTAR_ASSERT(r == 0);
     memory::set_reclaim_hook([this] (std::function<void ()> reclaim_fn) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
         add_high_priority_task(make_task(default_scheduling_group(), [fn = std::move(reclaim_fn)] {
             fn();
         }));
+#pragma GCC diagnostic pop
     });
 
     _loads.reserve(loads_size);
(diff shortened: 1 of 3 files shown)


xemul:

src/core/reactor.cc
```diff
@@ -1115,9 +1115,12 @@ reactor::reactor(std::shared_ptr<seastar::smp> smp, alien::instance& alien, unsi
     auto r = ::pthread_sigmask(SIG_UNBLOCK, &mask, NULL);
     SEASTAR_ASSERT(r == 0);
     memory::set_reclaim_hook([this] (std::function<void ()> reclaim_fn) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
         add_high_priority_task(make_task(default_scheduling_group(), [fn = std::move(reclaim_fn)] {
             fn();
         }));
```
avikivity, line 1122: Seastar code shouldn't use deprecated declarations. 

Once we deprecate something, from Seastar's persepective, it's ready to be removed at any time.

xemul: The goal is not to _remove_ those methods, but move them to private eventually and "undeprecated" them back. 

To follow the suggestion not to use deprecated symbols, there must appear add_high_priority_task_private() method, the add_high_priority_task() (now deprecated) should call it, and after deprecation quiscent timeout we'll need to rename add_..._private() back into add_...() one. Does it worth it?

bhalevy: @xemul maybe add corresponding seastar::internal entry points that would be called from the public entry points that are deprecated in this path, and change the library calls to use the internal entry points also in this patch.

xemul: And mark those entry points as friends? I'll try to re-structure it somehow else

travisdowns: Seems fine to me to delay that restructure/decision to the removal point?

PR #3343, opened 2026-04-03
tests: Replace ad-hoc http_consumer with production HTTP parser

xemul:
The http_consumer struct was a hand-rolled HTTP response parser introduced in 2015, predating the production chunked_source_impl by 6 years. It had a known bug: Content-Length was parsed as hex instead of decimal.

Replace it with http_response_parser (for headers) and chunked_source_impl / content_length_source_impl (for the body), which are the same classes used by experimental::client. Update run_test() to pass body_size (size_t) to the reader callback instead of http_consumer&.

### DIFF
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -33,6 +33,8 @@
 #include <seastar/util/assert.hh>
 #include <seastar/util/later.hh>
 #include <seastar/util/short_streams.hh>
+#include <seastar/util/string_utils.hh>
+#include <seastar/http/internal/content_source.hh>
 #include <seastar/util/closeable.hh>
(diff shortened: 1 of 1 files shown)


xemul:
@nyh , please review

xemul:
@nyh , please review

xemul:
@nyh , please review

xemul:
@nyh , please review

nyh (approved):
Looks good, but there's something I don't understand: If the test had a bug in parsing content length, does it mean it didn't actually test content length? Maybe we should add more tests for it?

xemul:
Yes, it didn't actually test content length, server responded with chunked encoding in those tests

PR #3344, opened 2026-04-03
Move some chunked_source_impl test from httpd set to content_source one

xemul:
There are several tests in httpd set that feed server with hard-coded request containing chunked-encoded body, make it respond with the echo handler and validate the received string with silly string.find() matchers. These tests effectively validate chunked_source_impl parsing, not the server itself. For such tests there's another file. In it the validation can be made explicit, but checking that conent_source_impl correctly identified and parsed all the relevant parts of the input.

### DIFF
--- a/tests/unit/content_source_test.cc
+++ b/tests/unit/content_source_test.cc
@@ -226,3 +226,22 @@ SEASTAR_TEST_CASE(test_chunk_extension_parser_fail) {
             });
     });
 }
+
+SEASTAR_TEST_CASE(test_trailer_part_parser_fail) {
+    return seastar::async([] {
+        // Valid data chunk followed by a trailer with invalid syntax (= instead of :)
+        auto inp = input_stream<char>(data_source(std::make_unique<buf_source_impl>(sstring(
+            "8\r\nparsable\r\n0\r\ngood:header\r\nbad=header\r\n\r\n"))));
+        std::unordered_map<sstring, sstring> chunk_extensions;
+        std::unordered_map<sstring, sstring> trailing_headers;
+        auto content_stream = input_stream<char>(data_source(std::make_unique<httpd::internal::chunked_source_impl>(inp, chunk_extensions, trailing_headers)));
+        // Read the data chunk successfully first
+        auto buf = content_stream.read().get();
+        BOOST_REQUIRE_EQUAL(sstring(buf.get(), buf.size()), sstring("parsable"));
+        // Next read triggers trailer parsing and must throw
+        BOOST_REQUIRE_EXCEPTION(content_stream.read().get(), httpd::bad_request_exception,
+            [] (const httpd::bad_request_exception& e) {
+                return sstring(e.what()).find("Can't parse chunked request trailer") != sstring::npos;
+            });
+    });
+}
--- a/tests/unit/content_source_test.cc
+++ b/tests/unit/content_source_test.cc
@@ -184,3 +184,30 @@ SEASTAR_TEST_CASE(test_fragmented_chunks) {
         BOOST_REQUIRE(trailing_headers[sstring("trailer")] == sstring("part"));
     });
 }
+
+SEASTAR_TEST_CASE(test_full_chunk_format) {
+    return seastar::async([] {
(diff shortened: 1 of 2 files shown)


xemul:
@nyh , please review

tests/unit/content_source_test.cc
```diff
@@ -184,3 +184,30 @@ SEASTAR_TEST_CASE(test_fragmented_chunks) {
         BOOST_REQUIRE(trailing_headers[sstring("trailer")] == sstring("part"));
     });
 }
+
+SEASTAR_TEST_CASE(test_full_chunk_format) {
```
nyh, line 189: It would have been nice to hear a bit about the motivation of this change - why did you want to do it?

I can see both pros and cons to this change: On one hand you can argue that if this feature can be tested in a lower level (content_source instead of http server) it should. On the other hand you can argue the exact opposite: Since the http server is the more important of the two and content source is just an implementation details that may change in the future, it's better to test the higher-level feature, which is less likely to change in the future!

Sometimes it's good to test things in a lower level because it gives you access to checking more things. But this is not the case here - the higher-level test already checked everything that the lower level test now checks.

tests/unit/content_source_test.cc
```diff
@@ -211,3 +212,17 @@ SEASTAR_TEST_CASE(test_full_chunk_format) {
...
+SEASTAR_TEST_CASE(test_chunk_extension_parser_fail) {
+    return seastar::async([] {
+        // Space after semicolon with no extension name is invalid
+        auto inp = input_stream<char>(data_source(std::make_unique<buf_source_impl>(sstring("7; \r\nnoparse\r\n0\r\n\r\n"))));
+        std::unordered_map<sstring, sstring> chunk_extensions;
+        std::unordered_map<sstring, sstring> trailing_headers;
+        auto content_stream = input_stream<char>(data_source(std::make_unique<httpd::internal::chunked_source_impl>(inp, chunk_extensions, trailing_headers)));
+        BOOST_REQUIRE_EXCEPTION(content_stream.read().get(), httpd::bad_request_exception,
+            [] (const httpd::bad_request_exception& e) {
+                return sstring(e.what()).find("Can't parse chunk size and extensions") != sstring::npos;
```
nyh, line 225: Continuing my previous comment on the motivation of this change, here you can see two more downsides of this change:

1. The test becomes longer and more complicated, not simpler...
2. The test stops "in the middle" - you get the bad_request_exception, but will that actually result in a 400 error from the server? You don't check, and the old test did.

tests/unit/content_source_test.cc
```diff
@@ -245,3 +245,22 @@ SEASTAR_TEST_CASE(test_trailer_part_parser_fail) {
...
+SEASTAR_TEST_CASE(test_too_long_chunk) {
+    return seastar::async([] {
+        // Second chunk declares 10 bytes but has an extra 'X' after them
+        auto inp = input_stream<char>(data_source(std::make_unique<buf_source_impl>(sstring(
+            "a\r\n1234567890\r\na\r\n1234521345X\r\n0\r\n\r\n"))));
+        std::unordered_map<sstring, sstring> chunk_extensions;
+        std::unordered_map<sstring, sstring> trailing_headers;
+        auto content_stream = input_stream<char>(data_source(std::make_unique<httpd::internal::chunked_source_impl>(inp, chunk_extensions, trailing_headers)));
+        // Read both chunk bodies in one call
+        auto buf = content_stream.read_exactly(20).get();
```
nyh, line 258: This test is very specific, about read_exactly(20). Why is this interesting? Will the HTTP server driving this code actually use read_exactly(20)?

nyh (commented):
It's hard for me to "approve" this patch, I don't understand the motivation. It looks to me like you're making these tests weaker (checking some internal implementation details instead of public APIs that the Seastar user is actually using), longer, and sometimes strangely specific (e.g., checking read_exactly(20) - I'm not sure why). 

I'm not against this change - I don't think there's anything wrong with it and if you like it I can merge it - but I also don't understand its motivation.

xemul:
Yes, the motivation is not really strong enough. I'll postpone this for now

PR #3345, opened 2026-04-06
rpc: Fix uint64 wraparound of expired timeout in send_entry()

### LINKED ISSUE #672: rpc: potential race condition can replace already-passed timeout with infinite timeout
Because clocks on different nodes aren't perfectly synchronized, the sending node converts the timeout from an absolute time to a relative time (how much time left), and then the receiving node converts it back to an absolute time in its own clock.

The code doing the first conversion, from absolute time to relative time on the sender, is in `connection::send_loop()`:

```c++
                      auto expire = d.t.get_timeout();
                      uint64_t left = 0;
                      if (expire != typename timer<rpc_clock_type>::time_point()) {
                          left = std::chrono::duration_cast<std::chrono::milliseconds>(expire - timer<rpc_clock_type>::clock::now()).count();
                      }
```

`left` is an unsigned integer so if `expire < now`, instead of the result being negative (or truncated to zero), it will wrap to a huge integer, making the requested timeout that had actually already occurred, postponed to the distant future.

The chance that this issue will occur in practice is rather low: First, it requires that the timeout be so short (or load is so high) that by the time we get to sending a request, it had already timed out (`expire < now`). Second, we already have code which culls expired requests from the send queue, so this issue is a _race condition_ - it will only happen if the send loop reaches this request before we had a chance to cull it.

I think the solution is to simply check if `expire < now`, and if it is, drop this request.

@gleb-cloudius

xemul:
When a request entry sits at the front of the outgoing queue and its
timeout deadline expires before send_entry() is called, the timer
callback fires but does nothing -- withdrawal is guarded by
'it != _outgoing_queue.begin()'.  The entry stays in the queue with
a stale _expiry.  send_entry() then computes
    
        expire - timer<rpc_clock_type>::clock::now()
    
which is a negative duration.  Assigning the negative int64_t result
to uint64_t left wraps around to a near-infinite value (~2^63), which
the server receives and interprets as a timeout far in the future,
causing the server to sleep for an effectively infinite period waiting
for a deadline that already passed on the client.
    
Fix: in send_entry(), if the timer was set and the deadline has already
passed, drop the entry without sending -- equivalent to the timer
callback having arrived just a little later than it did.  The server
never receives the corrupted timeout, and the client-side reply handler
times out normally and delivers timeout_error to the caller.
    
The expiration check is placed outside the 'buf.size && _propagate_timeout'
block intentionally: whether a timed-out entry should be dropped is a
client-side liveness decision, independent of whether the timeout value
is encoded into the wire frame.  Even when timeout propagation is
disabled, sending an already-expired request only makes the server do
work for a caller that will receive timeout_error regardless.
    
For the bug to surface two rare events must coincide:
      1. The future returned by the previous entry's send completes in a
         way that schedules the next entry's continuation as a new task
         rather than calling it immediately (i.e. it is not a ready future
         at the .then() site).
      2. That scheduling happens at a preemption point, so the reactor
         runs a poll cycle -- advancing the lowres clock and firing pending
         timer callbacks -- before the continuation calls uncancellable()
         and send_entry().

The bug cannot be deterministically reproduced without modifying the
production code-flow, because the .then() continuation on a ready
future is called inline in the current implementation (condition 1 is
never met in practice without an artificial yield).  To surface it
in a test one would need to inject a force_poll()+yield() step into
connection::send() between the two continuations, and run a
CPU-hogging scheduling group that busy-spins long enough to advance
the lowres clock past the request deadline.  That approach requires
modifying production code for test purposes, so no automated test is
added here; the fix is a straightforward defensive check.
    
Fixes #672

### DIFF
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -163,15 +163,21 @@ future<> connection::send_buffer(snd_buf buf) {
 
 future<> connection::send_entry(outgoing_entry& d) noexcept {
     return futurize_invoke([this, &d] {
+        auto expire = d.t.get_timeout();
+        // left_ms is 0 when no timer is set (server treats 0 as "no timeout").
+        // When a timer is set, drop the entry if already expired; otherwise send
+        // the remaining time so the server can honour the deadline too.
+        uint64_t left_ms = 0;
+        if (expire != typename timer<rpc_clock_type>::time_point()) {
+            left_ms = std::chrono::duration_cast<std::chrono::milliseconds>(expire - timer<rpc_clock_type>::clock::now()).count();
+            if (int64_t(left_ms) <= 0) {
+                return make_ready_future<>();
+            }
+        }
         if (d.buf.size && _propagate_timeout) {
             static_assert(snd_buf::chunk_size >= sizeof(uint64_t), "send buffer chunk size is too small");
             if (_timeout_negotiated) {
-                auto expire = d.t.get_timeout();
-                uint64_t left = 0;
-                if (expire != typename timer<rpc_clock_type>::time_point()) {
-                    left = std::chrono::duration_cast<std::chrono::milliseconds>(expire - timer<rpc_clock_type>::clock::now()).count();
-                }
-                write_le<uint64_t>(d.buf.front().get_write(), left);
+                write_le<uint64_t>(d.buf.front().get_write(), left_ms);
             } else {
                 d.buf.front().trim_front(sizeof(uint64_t));
                 d.buf.size -= sizeof(uint64_t);


nyh:
I have a question about the validity of your fix:

If we have an already expired request - even if it's the first request in the queue - why shouldn't we just _expire_ it immediately?
Your solution seems to be that in this case, instead of sending a timeout a million years in the future (the unsigned version of a negative number), we'll send a timeout of 0. But this 0 means (if I understand correctly) - no timeout at. I.e., an infinite amount of time in the future. How is an infinite timeout better, or even different, from the current behavior of a million-year timeout?

xemul:
> If we have an already expired request - even if it's the first request in the queue - why shouldn't we just expire it immediately?

That's _excellent_ question, and the answer is -- it's larger patch, but I wanted to discuss the testing effort first (thus RFC):

> ... do we want to try hard and implement the test without these two cludges, or do we trust the analyzis and just go on with the fix?

xemul:
> The PR is not for merging, because of this "error injection" in the connection::send(). The goal was to demonstrate the problem is real, but triggering this coincidence is not extremely easy. The question is -- do we want to try hard and implement the test without these two cludges, or do we trus the analyzis and just go on with the fix?

@nyh , @avikivity , what's your take on this? Does it make sense to try hard and implement decent test for it, or just acknowledging that the issue exists and fixing it with code comment would be enough?

bhalevy:
> > The PR is not for merging, because of this "error injection" in the connection::send(). The goal was to demonstrate the problem is real, but triggering this coincidence is not extremely easy. The question is -- do we want to try hard and implement the test without these two cludges, or do we trus the analyzis and just go on with the fix?
> 
> @nyh , @avikivity , what's your take on this? Does it make sense to try hard and implement decent test for it, or just acknowledging that the issue exists and fixing it with code comment would be enough?

ping @nyh @avikivity

nyh:
> > > The PR is not for merging, because of this "error injection" in the connection::send(). The goal was to demonstrate the problem is real, but triggering this coincidence is not extremely easy. The question is -- do we want to try hard and implement the test without these two cludges, or do we trus the analyzis and just go on with the fix?
> > 
> > 
> > @nyh , @avikivity , what's your take on this? Does it make sense to try hard and implement decent test for it, or just acknowledging that the issue exists and fixing it with code comment would be enough?
> 
> ping @nyh @avikivity

I'm fine with having a kludgy test that you used as a developer but isn't a good test to permanently run in CI. I think it's fine that the final code will not include a test, but you still describe in the commit message how you tested it.

But I had a different question above about the validity of your test, unrelated to the test, and I still don't understand the answer:

_Your solution seems to be that in this case, instead of sending a timeout a million years in the future (the unsigned version of a negative number), we'll send a timeout of 0. But this 0 means (if I understand correctly) - no timeout at. I.e., an infinite amount of time in the future. How is an infinite timeout better, or even different, from the current behavior of a million-year timeout?_

src/rpc/rpc.cc
```diff
@@ -169,7 +171,8 @@ future<> connection::send_entry(outgoing_entry& d) noexcept {
                 auto expire = d.t.get_timeout();
                 uint64_t left = 0;
                 if (expire != typename timer<rpc_clock_type>::time_point()) {
-                    left = std::chrono::duration_cast<std::chrono::milliseconds>(expire - timer<rpc_clock_type>::clock::now()).count();
+                    auto left_ms = std::chrono::duration_cast<std::chrono::milliseconds>(expire - timer<rpc_clock_type>::clock::now()).count();
+                    left = left_ms > 0 ? left_ms : 0;
```
avikivity: Wouldn't it be enough to change `left` to be a signed type?

xemul:
> I'm fine with having a kludgy test that you used as a developer but isn't a good test to permanently run in CI. I think it's fine that the final code will not include a test, but you still describe in the commit message how you tested it.

OK

> Your solution seems to be that in this case, instead of sending a timeout a million years in the future (the unsigned version of a negative number), we'll send a timeout of 0. But this 0 means (if I understand correctly) - no timeout at. I.e., an infinite amount of time in the future. How is an infinite timeout better, or even different, from the current behavior of a million-year timeout?

It's not better, it's worse. The correct fix is to either resolve the send-side with relevant error, or send some minmal timeout to the sender. I think that the former is more correct behavior.

src/rpc/rpc.cc
xemul: A rare case of expire == clock::now() (left = 0) still stays. In that case, receiver reads it as "no timeout" (since 592e4337aa5a2c2c431ca3945fb2fdee8fbc8f3e). Probably a bug too

xemul: No, not a bug :( If timer is not armed, the server side should be notified that the calls is not about to timeout, and the zero value was picked for that. So it's already the part of the protocol

avikivity: Aha. So better keep it unsigned.

I don't like using 0 as a special case, but it's already baked into the protocol.

xemul:
upd:
- rebased (no conflicts)
- resolve the outgoing entry if the timeout is expired without sending it
- kept only the fix, as suggested, and updated the patch comment to explain why no test included and how to cook one if really needed

src/rpc/rpc.cc
```diff
@@ -163,15 +163,21 @@ future<> connection::send_buffer(snd_buf buf) {
...
+        // left_ms is 0 when no timer is set (server treats 0 as "no timeout").
+        // When a timer is set, drop the entry if already expired; otherwise send
+        // the remaining time so the server can honour the deadline too.
+        uint64_t left_ms = 0;
+        if (expire != typename timer<rpc_clock_type>::time_point()) {
+            left_ms = std::chrono::duration_cast<std::chrono::milliseconds>(expire - timer<rpc_clock_type>::clock::now()).count();
+            if (int64_t(left_ms) <= 0) {
+                return make_ready_future<>();
+            }
+        }
```
avikivity, line 176: If the timer expires, it can end up calling it->done.set_exception(), but here we always return a ready future. Are we missing it here?

xemul: If the timer expires, it does not set exception:

```c++
        auto deleter = [this, it = _outgoing_queue.iterator_to(d)] {
            // Front entry is most likely (unless _negotiated is unresolved, check enqueue_zero_frame()) sitting
            // inside send_entry() continuations and thus it cannot be cancelled.
            if (it != _outgoing_queue.begin()) {
                withdraw(it); // <--- mind only one argument here
            }
        };


void connection::withdraw(outgoing_entry::container_t::iterator it, std::exception_ptr ex) {
    ...
    if (ex == nullptr) {
        it->done.set_value(); // <-- we're here on timeout
    } else {
        it->done.set_exception(ex);
    }
}
```

The caller-visible timeout comes from reply-wait timer, not from send-entry one. Exceptions from send-entry are about transport errors

PR #3347, opened 2026-04-08
Use util::memory_data_source in tests

xemul:
Tests actively use in-memory data sources to simulate data that comes from e.g. sockets. Some of those were replaced with "standard" one in util::, here are few more.


xemul:
@scylladb/seastar-maint , pleasw review

travisdowns (approved):

PR #3350, opened 2026-04-08
io_queue: Fix starvation time measurement to start on outgoing throttle

### LINKED ISSUE #1790: IO starvation time times larger than IO queue delay
image

It's either a bug or misunderstanding. Need fix or explanation

Cc @vladzcloudius

xemul:
The previous implementation measured starvation time starting when the first request was queued. This conflated normal queueing delay with actual starvation caused by fairness issues or token exhaustion, resulting in inflated metrics.

Change the starvation time measurement to start when a dispatch attempt fails due to lack of available outgoing tokens (grab_result::stop), which represents true starvation - when a class has work queued but cannot progress.

When grab_capacity returns stop, all priority classes with pending requests are marked as starving via on_dispatch_throttled(). The starvation timer ends when a request from that class is successfully dispatched.

The _activated timestamp now uses its default value (zero time_point) as a sentinel to indicate whether starvation timing is active.

fixes #1790

### DIFF
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -146,7 +146,7 @@ class io_queue::priority_class_data {
     std::chrono::duration<double> _total_queue_time;
     std::chrono::duration<double> _total_execution_time;
     std::chrono::duration<double> _starvation_time;
-    io_queue::clock_type::time_point _activated;
+    io_queue::clock_type::time_point _activated = io_queue::clock_type::time_point();
 
     class bandwidth_throttler {
         io_group::priority_class_data::token_bucket_t& _tb;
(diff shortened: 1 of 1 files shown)


src/core/io_queue.cc
```diff
@@ -146,7 +146,7 @@ class io_queue::priority_class_data {
     std::chrono::duration<double> _total_queue_time;
     std::chrono::duration<double> _total_execution_time;
     std::chrono::duration<double> _starvation_time;
-    io_queue::clock_type::time_point _activated;
+    io_queue::clock_type::time_point _activated = io_queue::clock_type::time_point();
```
bhalevy, line 149: nit: using an std::optional would be more elegant, but using a special "null" value is okay too

bhalevy: Can you please add a comment describing what `_activated` holds and how it is managed?

bhalevy: It looks like `_activated`, when not null, keeps the last time the queue was preempted, and so it's name is quite confusing.

xemul: Ah, the meaning of this field had changed with this PR, so I'd better rename it too

PR #3353, opened 2026-04-09
reactor: support rename2

rockwotj:
This is useful if you want to rename something but only if the
destination file does not exist.

### DIFF
--- a/include/seastar/core/file-types.hh
+++ b/include/seastar/core/file-types.hh
@@ -21,9 +21,11 @@
 
 #pragma once
 
+#include <cstdio>
 #include <fcntl.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <linux/fs.h>
 #include <type_traits>
 
 #include <seastar/util/bool_class.hh>
@@ -194,6 +196,31 @@ inline void operator&=(mmap_prot& a, mmap_prot b) {
 ///
 /// \see file::mmap()
 using mmap_private = bool_class<struct mmap_private_tag>;
+
+/// Flags for the rename_file() operation.
+///
+/// \see rename_file()
+enum class rename_flags {
+    none = 0,
+    /// Don't overwrite the target if it already exists; instead, fail with EEXIST.
+    noreplace = RENAME_NOREPLACE,
+    /// Atomically exchange the source and destination paths.
+    ///
+    /// See RENAME_EXCHANGE
+    exchange = RENAME_EXCHANGE,
+    /// Create a whiteout at the source (for overlayfs/union mounts).
+    ///
+    /// See RENAME_WHITEOUT
+    whiteout = RENAME_WHITEOUT,
+};
+
+inline constexpr rename_flags operator|(rename_flags a, rename_flags b) {
+    return rename_flags(std::underlying_type_t<rename_flags>(a) | std::underlying_type_t<rename_flags>(b));
+}
+
+inline constexpr rename_flags operator&(rename_flags a, rename_flags b) {
+    return rename_flags(std::underlying_type_t<rename_flags>(a) & std::underlying_type_t<rename_flags>(b));
+}
 
 /// @}
 
--- a/include/seastar/core/reactor.hh
+++ b/include/seastar/core/reactor.hh
@@ -552,7 +552,7 @@ public:
     future<std::filesystem::space_info> file_system_space(std::string_view pathname) noexcept;
     future<struct statvfs> statvfs(std::string_view pathname) noexcept;
     future<> remove_file(std::string_view pathname) noexcept;
-    future<> rename_file(std::string_view old_pathname, std::string_view new_pathname) noexcept;
+    future<> rename_file(std::string_view old_pathname, std::string_view new_pathname, rename_flags flags = rename_flags::none) noexcept;
     future<> link_file(std::string_view oldpath, std::string_view newpath) noexcept;
     future<> chmod(std::string_view name, file_permissions permissions) noexcept;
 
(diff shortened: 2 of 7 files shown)


src/core/reactor.cc
```diff
@@ -1971,6 +1971,18 @@ reactor::rename_file(std::string_view old_pathname_view, std::string_view new_pa
...
 }
 
+future<>
+reactor::rename_file2(std::string_view old_pathname_view, std::string_view new_pathname_view, unsigned int flags) noexcept {
+    auto old_pathname = sstring(old_pathname_view);
+    auto new_pathname = sstring(new_pathname_view);
+    syscall_result<int> sr = co_await _thread_pool->submit<syscall_result<int>>(
+            internal::thread_pool_submit_reason::file_operation, [old_pathname, new_pathname, flags] {
+        return wrap_syscall<int>(static_cast<int>(
+                ::syscall(SYS_renameat2, AT_FDCWD, old_pathname.c_str(), AT_FDCWD, new_pathname.c_str(), flags)));
```
rockwotj: had to switch to the raw syscall because it looks like musl on alpine doesn't have `::renameat2` yet

include/seastar/core/reactor.hh
```diff
@@ -553,6 +553,7 @@ public:
     future<struct statvfs> statvfs(std::string_view pathname) noexcept;
     future<> remove_file(std::string_view pathname) noexcept;
     future<> rename_file(std::string_view old_pathname, std::string_view new_pathname) noexcept;
+    future<> rename_file2(std::string_view old_pathname, std::string_view new_pathname, unsigned int flags) noexcept;
```
xemul: Just adding flags with default value to existing method is backward-compatible enough

rockwotj: Done

include/seastar/core/reactor.hh
```diff
@@ -552,7 +552,7 @@ public:
     future<std::filesystem::space_info> file_system_space(std::string_view pathname) noexcept;
     future<struct statvfs> statvfs(std::string_view pathname) noexcept;
     future<> remove_file(std::string_view pathname) noexcept;
-    future<> rename_file(std::string_view old_pathname, std::string_view new_pathname) noexcept;
+    future<> rename_file(std::string_view old_pathname, std::string_view new_pathname, unsigned int flags = 0) noexcept;
```
xemul: We try to avoid passing raw system flags into seastar API methods. E.g. see open_flags or access_flags

rockwotj: using the libc constants to define flags here is not really possible because alpine doesn't have them yet

avikivity: You can do something like

```c++
enum class rename_flags = {
    foo = 
#ifdef THE_FLAG_YOU_WANT
        THE_FLAG_YOU_WANT
#else
        the_actual_number
\#endif
};
```

It's a but silly (if you know the number, why #define) but does the job.

rockwotj: Okay added.

rockwotj: I just used the actual numbers (taken from my system header file) :shrug:

avikivity: It's better to use the linux header. To protect against different archs using different numbers (long-discouraged practice and unlikely to ever happen).

avikivity:
Please squash the two patches, this repository frowns upon introducing a mistake and fixing it in the same series.

include/seastar/core/reactor.hh
rockwotj: I'm very certain that musl on alpine won't have the linux header, but if CI fails I'll update with the fallback you mentioned above

rockwotj: Seems I am wrong. yay! Resolving

include/seastar/core/file-types.hh
```diff
@@ -160,6 +162,31 @@ inline constexpr file_permissions operator&(file_permissions a, file_permissions
...
 }
 
+/// Flags for the rename_file() operation.
+///
+/// \see rename_file()
+enum class rename_flags {
+    none = 0,
+    /// Don't overwrite the target if it already exists; instead, fail with EEXIST.
+    noreplace = RENAME_NOREPLACE,
+    /// Atomically exchange the source and destination paths.
```
avikivity, line 172: How was the problem with Alpine not exposing the flags solved?

include/seastar/core/seastar.hh
```diff
@@ -334,11 +334,14 @@ future<> remove_file(std::string_view name) noexcept;
 ///
 /// \param old_name existing file name
 /// \param new_name new file name
+/// \param flags \c renameat2 flags (e.g., \c RENAME_NOREPLACE, \c RENAME_EXCHANGE,
+///              \c RENAME_WHITEOUT; see \c linux/fs.h). Defaults to \c 0 for
+///              standard rename semantics.
```
avikivity: Now outdated; also refer to the manual page, not a random header.

avikivity:
Please add a small test.

include/seastar/core/file-types.hh
rockwotj: Turns out it wasn't a problem with the flags. alpine has the flags but doesn't have renameat2 surprisingly (at least to me)

avikivity: I guess musl is slower to update than glibc.

include/seastar/core/seastar.hh
rockwotj: Updated

rockwotj (commented):
Added a test

xemul (approved):

rockwotj:
gentle ping

rockwotj:
ping

avikivity:
Sorry. Please add rename_flags to seastar.cppm (and rebase).

rockwotj:
> Sorry. Please add rename_flags to seastar.cppm (and rebase).

done

PR #3354, opened 2026-04-12
backtrace: print the build_id along with the backtrace

### LINKED ISSUE #3352: Print the build_id along with backtrace
To decode a backtrace using the binary that produced it one typically uses the binary's unique build_id to locate the binary.

However, this build_id may not be readily available along with the backtrace, e.g. when it is printed to the system log when the binary starts (as done in Scylla) and the log is rotated.

That has lead to a number of common backtrace-decoding issues like:
- Using the wrong build_id for issues that happened during rolling upgrade. In this case the old build_id was collected from the node log, while is restarted with a new build_id after the upgrade.
- Using the wrong build_id due to different architecture or build mode.  Official release version taken from backoffice, confusion about the instance it ran on: x86_64 vs. aarch64.

To make backtraces self-descriptive, print the build_id along with the backtrace similar to the AddressSanitzer format as a `(BuildId: <sha>)` suffix.

bhalevy:
To decode a backtrace using the binary that produced it one typically
uses the binary's unique build_id to locate the binary.

However, this build_id may not be readily available along with the
backtrace, e.g. when it is printed to the system log when the binary
starts and the log is rotated.

To make backtraces self-descriptive, print the build_id along with the
backtrace similar to the AddressSanitzer format as a
`(BuildId: <sha>)` suffix.

The build_id is cached in a statically allocated buffer, avoiding
dynamic allocation. It is statically initialized at program start
via init_build_id(), ensuring it is available in all contexts without
requiring an explicit initialization call from the reactor.

Use `#if __has_include(<elf.h>)` to detect ELF support at compile
time. On platforms without ELF headers, get_build_id() returns an
empty string and the `(BuildId: ...)` suffix is omitted from
backtraces, maintaining backward compatibility.

Relax the dl_phdr_info size check in the dl_iterate_phdr callback:
older library implementations may pass a smaller struct since some
members were added later. Verify the size covers at least up to
dlpi_phnum rather than requiring the full struct size.

Remove the assertion that dlpi_name is empty since the callback may
be invoked for shared libraries, not just the main program.

Fixes #3352

Signed-off-by: Benny Halevy <bhalevy@scylladb.com>

### DIFF
--- a/src/testing/test_runner.cc
+++ b/src/testing/test_runner.cc
@@ -27,6 +27,7 @@
 #include <seastar/testing/random.hh>
 #include <seastar/testing/test_runner.hh>
 #include <seastar/util/assert.hh>
+#include <seastar/util/internal/build_id.hh>
 
 namespace seastar {
 
@@ -86,6 +87,7 @@ bool test_runner::start_thread(int ac, char** av) {
                 });
             };
 
+            std::cout << fmt::format("build_id={}", internal::get_build_id()) << std::endl;
             return init().then([this] {
               return do_until([this] { return _done; }, [this] {
                 // this will block the reactor briefly, but we don't care
--- a/src/util/build_id.cc
+++ b/src/util/build_id.cc
@@ -1,0 +1,138 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2019-present ScyllaDB
+ */
+
+#include <cstring>
+#include <ostream>
+
+#include <seastar/util/internal/build_id.hh>
+
+#if __has_include(<elf.h>)
+#include <elf.h>
+#include <link.h>
+#define HAS_ELF
+
+#include <stddef.h>
+#include <fmt/format.h>
+
+#include <seastar/core/align.hh>
+#endif
+
+namespace seastar::internal {
+
+#ifdef HAS_ELF
+
+constexpr size_t max_size = 32; // // SHA-256 (32 bytes) is the largest hash algorithm supported by ld --build-id.
+
+static const Elf64_Nhdr* get_nt_build_id(dl_phdr_info* info) {
+    auto base = info->dlpi_addr;
+    const auto* h = info->dlpi_phdr;
+    auto num_headers = info->dlpi_phnum;
+    for (int i = 0; i != num_headers; ++i, ++h) {
+        if (h->p_type != PT_NOTE) {
+            continue;
+        }
+
+        auto* p = reinterpret_cast<const char*>(base + h->p_vaddr);
+        auto* e = p + h->p_memsz;
+        while (p + sizeof(Elf64_Nhdr) <= e) {
+            const auto* n = reinterpret_cast<const Elf64_Nhdr*>(p);
+            if (n->n_type == NT_GNU_BUILD_ID) {
+                return n;
+            }
+
+            p += sizeof(Elf64_Nhdr);
+
+            p += n->n_namesz;
+            p = align_up(p, 4);
+
+            p += n->n_descsz;
+            p = align_up(p, 4);
+        }
+    }
+
+    return nullptr;
+}
+
+static int callback(dl_phdr_info* info, size_t size, void* data) {
+    // size is the size of the dl_phdr_info struct passed by dl_iterate_phdr.
+    if (size < offsetof(dl_phdr_info, dlpi_phnum) + sizeof(info->dlpi_phnum)) {
+        fmt::print(stderr, "warning: build_id: dl_iterate_phdr callback size {} is too small\n", size);
+        return -1;
+    }
+
+    char* buf = static_cast<char*>(data);
+
+    auto* n = get_nt_build_id(info);
+    if (!n) {
+        fmt::print(stderr, "warning: build_id: no NT_GNU_BUILD_ID note\n");
+        return -1;
+    }
+
+    if (n->n_descsz > max_size) {
+        fmt::print(stderr, "warning: build_id: n_descsz {} exceeds max {}\n",
+                   n->n_descsz, max_size);
+        return -1;
+    }
+
+    auto* p = reinterpret_cast<const unsigned char*>(n);
+
+    p += sizeof(Elf64_Nhdr);
+
+    p += n->n_namesz;
+    p = align_up(p, 4);
+
+    for (size_t i = 0; i < n->n_descsz; ++i) {
+        fmt::format_to(buf + 2 * i, "{:02x}", p[i]);
+    }
+    buf[2 * n->n_descsz] = '\0';
+    return n->n_descsz;
+}
+
+struct build_id {
+    char buf[2 * max_size + 1];
+
+    build_id() {
+        if (dl_iterate_phdr(callback, &buf) < 0) {
+            std::strcpy(buf, "unknown");
+        }
+    }
+
+    const char* get() const {
+        return buf;
+    }
+};
+
+#else // HAS_ELF
+
+struct build_id {
+    const char* get() const {
+        return "unknown";
+    }
+};
+
+#endif // HAS_ELF
+
+const char* get_build_id() {
+    static internal::build_id cache;
+    return cache.get();
+}
+
+} // namespace seastar::internal
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 .cproject
 .project
 .settings
-build*
+build/
 build.ninja
 cscope.*
 __pycache__/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -807,6 +807,7 @@ add_library (seastar
   src/util/read_first_line.cc
   src/util/tmp_file.cc
   src/util/short_streams.cc
+  src/util/build_id.cc
   src/websocket/parser.cc
   src/websocket/common.cc
   src/websocket/client.cc
--- a/include/seastar/util/internal/build_id.hh
+++ b/include/seastar/util/internal/build_id.hh
@@ -1,0 +1,32 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2019-present ScyllaDB
+ */
+
+#pragma once
+
+namespace seastar::internal {
+
+/// Returns the build ID of the currently running executable, if available.
+/// The build ID is a unique identifier for a specific build of the executable,
+/// often used for debugging and profiling purposes.
+/// If the build ID is not available, this function returns "unknown".
+const char* get_build_id();
+
+} // namespace seastar::internal
--- a/scripts/addr2line.py
+++ b/scripts/addr2line.py
@@ -240,6 +240,7 @@ class BacktraceResolver:
             addr = "0x[0-9a-f]+"
             path = r"\S+"
             token = fr"(?:{path}\+)?{addr}"
+            build_id = r"\(BuildId: [0-9a-fA-F]+\)"
             full_addr_match = fr"(?:(?P<path>{path})\s*\+\s*)?(?P<addr>{addr})"
             ignore_addr_match = fr"(?:(?P<path>{path})\s*\+\s*)?(?:{addr})"
 
@@ -249,7 +250,7 @@ class BacktraceResolver:
             # All of the below except for separator_re must only match lines containing 0x in them
             # as we use that as a quick first filter
             self.oneline_re = re.compile(
-                fr"^((?:.*(?:(?:at|backtrace):?|:))?(?:\s+))?({token}(?:\s+{token})*)(?:\).*|\s*)$",
+                fr"^((?:.*(?:(?:at|backtrace):?|:))?(?:\s+))?({token}(?:\s+{token})*)(\s+{build_id})?(?:\).*|\s*)$",
                 flags=re.IGNORECASE,
             )
             self.address_re = re.compile(full_addr_match, flags=re.IGNORECASE)
@@ -259,7 +260,7 @@ class BacktraceResolver:
             )
             self.kernel_re = re.compile(fr'^.*kernel callstack: (?P<addrs>(?:{addr}\s*)+)$')
             self.asan_re = re.compile(
-                fr"^(?:.*\s+)\({full_addr_match}\)(\s+\(BuildId: [0-9a-fA-F]+\))?$",
+                fr"^(?:.*\s+)\({full_addr_match}\)(\s+{build_id})?$",
                 flags=re.IGNORECASE,
             )
             self.generic_re = re.compile(fr"^(?:.*\s+){full_addr_match}\s*$", flags=re.IGNORECASE)
--- a/scripts/seastar-addr2line
+++ b/scripts/seastar-addr2line
@@ -146,7 +146,7 @@ class TestStringMethods(unittest.TestCase):
     def test_reactor_stall_reports(self):
         data: TestList = [
             (
-                'Reactor stalled on shard 1. Backtrace: 0x12f34',
+                'Reactor stalled on shard 1. Backtrace: 0x12f34 (BuildId: 2d4295e82afee91a64c6dffbb1935537e002df2d)',
                 {
                     'type': BacktraceResolver.BacktraceParser.Type.ADDRESS,
                     'prefix': 'Reactor stalled on shard 1. Backtrace:',
@@ -154,7 +154,7 @@ class TestStringMethods(unittest.TestCase):
                 },
             ),
             (
-                'Reactor stalled on shard 1. Backtrace: 0xa1234 0xb5678',
+                'Reactor stalled on shard 1. Backtrace: 0xa1234 0xb5678 (BuildId: 2d4295e82afee91a64c6dffbb1935537e002df2d)',
                 {
                     'type': BacktraceResolver.BacktraceParser.Type.ADDRESS,
                     'prefix': 'Reactor stalled on shard 1. Backtrace:',
@@ -165,7 +165,7 @@ class TestStringMethods(unittest.TestCase):
                 },
             ),
             (
-                'Reactor stalled on shard 1. Backtrace: /my/path+0xabcd',
+                'Reactor stalled on shard 1. Backtrace: /my/path+0xabcd (BuildId: 2d4295e82afee91a64c6dffbb1935537e002df2d)',
                 {
                     'type': BacktraceResolver.BacktraceParser.Type.ADDRESS,
                     'prefix': 'Reactor stalled on shard 1. Backtrace:',
@@ -191,7 +191,7 @@ class TestStringMethods(unittest.TestCase):
     def test_boost_failure(self):
         data: TestList = [
             (
-                'Expected partition_end(), but got end of stream, at 0xa1234 0xb5678 /my/path+0xabcd',
+                'Expected partition_end(), but got end of stream, at 0xa1234 0xb5678 /my/path+0xabcd (BuildId: 2d4295e82afee91a64c6dffbb1935537e002df2d)',
                 {
                     'type': BacktraceResolver.BacktraceParser.Type.ADDRESS,
                     'prefix': 'Expected partition_end(), but got end of stream, at',
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -165,6 +165,7 @@
 #include <seastar/util/spinlock.hh>
 #include <seastar/util/internal/iovec_utils.hh>
 #include <seastar/util/internal/magic.hh>
+#include <seastar/util/internal/build_id.hh>
 #include "core/crypto.hh"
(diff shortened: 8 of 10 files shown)


bhalevy:
- updated double license (as the build_id sources were copied from scylladb)

bhalevy:
[89915fa](https://github.com/scylladb/seastar/commit/89915facd5d42bbe6c4d52873b5e36fd9b322161):
- rebased
- prevent allocation in get_build_id() (caused alloc_test to fail when reactor stalled during `test_diagnostics_allocation`) by:
  - returning std::string_view (into thread local static std::string)
  - printing the build_id in the beginning of every test (filling the static std::string buffer)

src/util/build_id.cc
```diff
@@ -0,0 +1,102 @@
...
+    std::string ret;
+    int r = dl_iterate_phdr(callback, &ret);
+    SEASTAR_ASSERT(r == 1);
+    return ret;
+}
+
+std::string_view get_build_id() {
+    static thread_local std::string cache;
+    if (cache.empty()) {
+        cache = really_get_build_id();
```
tgrabiec, line 97: This can deadlock if backtrace is printed from a signal handler.

Maybe we should only do it on init and skip printing if not yet initialized.

bhalevy: Can do.
That makes sense.

src/util/build_id.cc
```diff
@@ -0,0 +1,106 @@
...
+
+            p += n->n_namesz;
+            p = align_up(p, 4);
+
+            p += n->n_descsz;
+            p = align_up(p, 4);
+        }
+    }
+
+    SEASTAR_ASSERT(0 && "no NT_GNU_BUILD_ID note");
```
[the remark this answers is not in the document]
bhalevy: good point

src/util/build_id.cc
```diff
@@ -0,0 +1,106 @@
...
+                return n;
+            }
+
+            p += sizeof(Elf64_Nhdr);
+
+            p += n->n_namesz;
+            p = align_up(p, 4);
+
+            p += n->n_descsz;
+            p = align_up(p, 4);
```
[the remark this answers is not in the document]
bhalevy: @Copilot  The suggestion is overcomplucated. Wouldn't be enough to run the existing loop while (p <= e - sizeof(Elf64_Nhdr)) ?

src/util/build_id.cc
```diff
@@ -0,0 +1,106 @@
...
+    SEASTAR_ASSERT(0 && "no NT_GNU_BUILD_ID note");
+}
+
+static int callback(dl_phdr_info* info, size_t size, void* data) {
+    std::string& ret = *(std::string*)data;
+    std::ostringstream os;
+
+    // The first DSO is always the main program, which has an empty name.
+    SEASTAR_ASSERT(strlen(info->dlpi_name) == 0);
+
```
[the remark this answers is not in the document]
bhalevy: @Copilot ok to include missing include.

Regarding size, what does it supposed to contain?
Can the callback verify that or use it in any other way?

src/util/build_id.cc
bhalevy: Used static buffer and static initialization in [c4e17ad](https://github.com/scylladb/seastar/commit/c4e17ad298fd40bb6ca775ff9294c9a86cd768c3)

include/seastar/util/build_id.hh
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma once
+
+#include <string>
+
+namespace seastar {
+
+void init_build_id();
+std::string_view get_build_id();
+
```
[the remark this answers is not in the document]
bhalevy: Done in [c4e17ad](https://github.com/scylladb/seastar/commit/c4e17ad298fd40bb6ca775ff9294c9a86cd768c3)

src/util/build_id.cc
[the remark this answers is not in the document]
bhalevy: Revised implementation in [c4e17ad](https://github.com/scylladb/seastar/commit/c4e17ad298fd40bb6ca775ff9294c9a86cd768c3)

src/util/build_id.cc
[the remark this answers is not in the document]
bhalevy: Verified `size` in [c4e17ad](https://github.com/scylladb/seastar/commit/c4e17ad298fd40bb6ca775ff9294c9a86cd768c3)

bhalevy:
Changes since the previous push (89915fa):

- **Squashed into a single commit**: Merged the two commits (`backtrace: print the build_id` + `reactor: initialize build_id on all shards`) into one
- **Rebased onto current `origin/master`** (7a0b8b28)
- **Replaced `std::string` + `std::ostringstream` with a statically allocated `build_id_buf`**: Avoids dynamic allocation entirely. Uses `fmt::format_to` to write hex digits directly into a fixed-size `char` buffer (sized for SHA-256, the largest hash `ld --build-id` supports)
- **Replaced `thread_local` cache with file-scope static initialization**: The build_id is now initialized once at program start via a file-scope `static const std::string_view build_id = init_build_id()`, eliminating the need for `thread_local` and the per-reactor `init_build_id()` call
- **Removed `init_build_id()` from public API and from `reactor::do_run()`**: No longer needed since initialization happens statically
- **Added `SEASTAR_ASSERT(size == sizeof(dl_phdr_info))`** in the `dl_iterate_phdr` callback as a sanity check
- **Added overflow check**: Verifies `n_descsz` does not exceed `sha256_digest_length` (32 bytes) before writing to the buffer
- **Added `check_regex_match()` helper** in `exception_logging_test.cc`: Prints both the actual message and regex pattern on mismatch for easier debugging
- **Used `static_cast` instead of C-style cast** in the callback

bhalevy:
Wrapped `check_regex_match()` in `#ifndef SEASTAR_BACKTRACE_UNIMPLEMENTED` to fix `-Werror=unused-function` build failure on platforms where backtrace is unimplemented.

bhalevy:
Changes since the previous push (7b81246):

- **ELF portability**: Guard all ELF-specific code with `#if __has_include(<elf.h>)` / `#ifdef HAS_ELF`. On platforms without ELF headers, `get_build_id()` returns an empty string
- **Conditional BuildId output**: Only append `(BuildId: ...)` to backtraces when the build_id is non-empty, both in `backtrace.cc` (formatter) and `reactor.cc` (`backtrace_buffer`). This maintains backward compatibility on non-ELF platforms
- **Relaxed dl_phdr_info size check**: Replaced `SEASTAR_ASSERT(size == sizeof(dl_phdr_info))` with a check that `size >= offsetof(dl_phdr_info, dlpi_phnum)`, since older library implementations may pass a smaller struct (later members were added incrementally per `<link.h>`)
- **Removed dlpi_name assertion**: Dropped `SEASTAR_ASSERT(std::strlen(info->dlpi_name) == 0)` since the callback may be invoked for shared libraries, not just the main program

All tests pass (94/94, except pre-existing `tls_openssl` failure).

bhalevy:
> All tests pass (94/94, except pre-existing tls_openssl failure).

See https://github.com/scylladb/seastar/issues/3364

bhalevy:
https://github.com/scylladb/seastar/actions/runs/24762567324/job/72449388147 timed out and got cancelled after an hour.  It took it more than half an hour to install dependencies. From the log it looks like download speed was just slow.

@xemul  Is there a way to re-run ci, or just this particular job?

bhalevy:
rebased, no conflicts

bhalevy:
Sorry, this Pr wasn't approved yet.
@tgrabiec please re-review

bhalevy:
Force-pushed with the following changes since last push:

- Introduced a `build_id` class (in `include/seastar/util/build_id.hh`) that stores the raw binary build-id bytes (up to 32 for SHA-256) instead of a pre-formatted hex string, avoiding unnecessary static hex formatting.
- Added `fmt::formatter<seastar::build_id>` specialization that formats bytes as hex on demand, used by both `simple_backtrace` formatter and `operator<<`.
- Added out-of-line `operator<<(std::ostream&, const build_id&)` in `build_id.cc` that delegates to the fmt formatter.
- Constructor `build_id(const unsigned char*, size_t)` is `noexcept`, asserts `size <= max_size`, and copies into an internal buffer without dynamic allocation.
- Generalized `backtrace_buffer::append_hex()` with a `Padding` template parameter (matching `convert_hex_safe`) for zero-padded hex output in the signal-safe backtrace path.
- Removed the now-unused `append(std::string_view)` overload and `build_id_buf` struct.
- Rebased on top of `origin/master`.

include/seastar/util/build_id.hh
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2019-present ScyllaDB
```
tgrabiec: 2026

src/util/build_id.cc
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2019-present ScyllaDB
```
tgrabiec, line 19: 2026

src/util/build_id.cc
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    SEASTAR_ASSERT(0 && "no NT_GNU_BUILD_ID note");
+    __builtin_unreachable();
+}
+
+static int callback(dl_phdr_info* info, size_t size, void* data) {
+    // size is the size of the dl_phdr_info struct passed by dl_iterate_phdr.
+    if (size < offsetof(dl_phdr_info, dlpi_phnum)) {
```
tgrabiec: Should be `size < offsetof(dl_phdr_info, dlpi_phnum) + sizeof(info->dlpi_phnum)`

tgrabiec (commented):
Just some nits

bhalevy:
Fixed review comments:

1. Updated copyright year to 2026 in both `build_id.hh` and `build_id.cc`
2. Fixed `callback()` size check to `size < offsetof(dl_phdr_info, dlpi_phnum) + sizeof(info->dlpi_phnum)` to ensure `dlpi_phnum` is fully accessible

include/seastar/util/build_id.hh
bhalevy: Fixed in eb1bfd37

src/util/build_id.cc
bhalevy: Fixed in eb1bfd37

src/util/build_id.cc
bhalevy: Fixed in eb1bfd37

src/util/build_id.cc
bhalevy: So the copyright from 2019 was taken from the original files in scylladb/utils/build_id.{cc,hh}
Should it be reset to 2026 in seastar?
A lot of the implementation changed, but some low level functions were not fully rewritten.

bhalevy:
Unrelated timeout in debug mode:
```
36/93 Test #36: Seastar.unit.websocket ........................***Timeout 300.06 sec
[0/1] cd /home/runner/work/seastar/seastar/build/debug/tests/unit && /usr/local/bin/cmake -E env ASAN_OPTIONS=disable_coredump=0:abort_on_error=1:detect_stack_use_after_return=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1 BOOST_TEST_CATCH_SYSTEM_ERRORS=no /home/runner/work/seastar/seastar/build/debug/tests/unit/websocket_test -- -c 2
Running 6 test cases...
INFO  2026-04-28 06:54:56,889 seastar - Reactor backend: io_uring
INFO  2026-04-28 06:54:56,890 seastar - Perf-based stall detector creation failed (EACCESS), try setting /proc/sys/kernel/perf_event_paranoid to 1 or less to enable kernel backtraces: falling back to posix timer.
INFO  2026-04-28 06:54:56,896 [shard 0:main] io - IO queue was unable to find a suitable maximum request length, the search was cut-off early at: 16MB
INFO  2026-04-28 06:54:56,896 [shard 0:main] io - IO queue was unable to find a suitable maximum request length, the search was cut-off early at: 16MB
build_id=53fb28752d722b166d4fd3338da3085c91ed2d5d
random-seed=2125371533
==7082==WARNING: ASan doesn't fully support makecontext/swapcontext functions and may produce false positives in some cases!
Sec-WebSocket-Protocol:echo
Sec-WebSocket-Accept:s3pPLMBiTxaQ9kYGzzhZRbK+xOo=
Sec-WebSocket-Version:13
Connection:Upgrade
Upgrade:websocket
f.size(): 0
Sec-WebSocket-Accept:s3pPLMBiTxaQ9kYGzzhZRbK+xOo=
Sec-WebSocket-Version:13
Connection:Upgrade
Upgrade:websocket
f.size(): 0
f.size(): 4
f.size(): 0
f.size(): 4
f.size(): 0
```

Doesn't reproduce for me locally.

src/util/build_id.cc
tgrabiec: I wasn't aware it was moved. In that case, the original year should be preserved.

bhalevy: Restored in 4dc92a7c

include/seastar/util/build_id.hh
bhalevy: Restored in 4dc92a7c

tgrabiec (approved):

tgrabiec:
I would like @avikivity to also approve it

bhalevy:
> I would like @avikivity to also approve it

@avikivity ping

bhalevy:
> > I would like @avikivity to also approve it
> 
> @avikivity ping

@avikivity ping, please review and consider merging

bhalevy:
> > > I would like @avikivity to also approve it
> > 
> > 
> > @avikivity ping
> 
> @avikivity ping, please review and consider merging

@avikivity ping^, please review and consider merging

include/seastar/util/build_id.hh
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#pragma once
+
+#include <cstddef>
+#include <iosfwd>
+#include <fmt/format.h>
+
+namespace seastar {
+
```
avikivity: There's no reason for this header or the symbols to be public.

include/seastar/util/build_id.hh
```diff
@@ -0,0 +1,57 @@
...
+    /// Asserts that \c size does not exceed \c max_size.
+    build_id(const unsigned char* data, size_t size) noexcept;
+    const unsigned char* data() const noexcept { return _data; }
+    size_t size() const noexcept { return _size; }
+    bool empty() const noexcept { return _size == 0; }
+};
+
+const build_id& get_build_id();
+
+std::ostream& operator<<(std::ostream& os, const build_id& id);
```
avikivity: Is this needed?

src/testing/test_runner.cc
```diff
@@ -86,6 +87,7 @@ bool test_runner::start_thread(int ac, char** av) {
                 });
             };
 
+            std::cout << "build_id=" << get_build_id() << std::endl;
```
avikivity: Use fmt::print()

src/util/build_id.cc
```diff
@@ -0,0 +1,141 @@
...
+#include <stddef.h>
+
+#include <seastar/core/align.hh>
+#endif
+
+namespace seastar {
+
+build_id::build_id(const unsigned char* data, size_t size) noexcept
+    : _size(size) {
+    SEASTAR_ASSERT(size <= max_size);
```
avikivity: Do we want to crash while printing a backtrace?

bhalevy: It will crash when the program starts.
But that said, we can just leave the build_id empty in this case, and maybe print a warming to the standard error (the logging subsystem is not guaranteed to be ready during static initialization).

include/seastar/util/build_id.hh
bhalevy: Currently, scylla has its own implementation. Should it not use seastar's?

bhalevy:
Here's a summary of the changes since the last push (4dc92a7):

Major design change: The public build_id class has been entirely removed. Instead of a class storing binary data with a fmt::formatter, the implementation is now much simpler:
1. include/seastar/util/build_id.hh — deleted. The get_build_id() declaration moved to include/seastar/util/process.hh as const char* get_build_id() (returns a C string or nullptr).
2. src/util/build_id.cc — Complete redesign:
   - removed all SEASTAR_ASSERT calls
   - No more build_id class with binary storage and formatter.
   - The hex formatting is now done at init time into a char buf[65] inside an internal build_id struct (private to the implementation).
   - get_build_id() returns const char* (null-terminated hex string, or nullptr on non-ELF).
   - get_nt_build_id() returns nullptr instead of asserting on failure.
   - The callback prints "warning:" prefixed messages on errors.
   - dl_iterate_phdr failure is no longer an assertion — silently returns nullptr.
4. src/core/reactor.cc — Reverted the append_hex Padding template parameter. Now uses append(s) directly on the const char* returned by get_build_id(). The if condition is simply if (auto s = get_build_id()).
5. src/util/backtrace.cc — Uses seastar::get_build_id() returning const char*, formats with "(BuildId: {})" directly (no formatter specialization needed).
6. src/testing/test_runner.cc — Uses get_build_id() with s ?: "" for null-safety.
7. CMakeLists.txt — Removed include/seastar/util/build_id.hh from the headers list; added a -Wno-error=unused-command-line-argument workaround for C++ module builds.
8. Other changes in reactor.cc are from unrelated rebased commits (smp refactoring, rename2 support, nowait detection, stall detector optimization, etc.).

In essence, the approach reverted from "store binary + format on demand" back to "store pre-formatted hex string" but simplified significantly — no class, no formatter, just a const char* API.

include/seastar/util/build_id.hh
bhalevy: Removed the header in [31ec32d](https://github.com/scylladb/seastar/commit/31ec32dfe48a8bf80a860b943f4b5c2f47a9de92)

Exposing just a `const char* get_build_id()` functions.

include/seastar/util/build_id.hh
bhalevy: Dropped in [31ec32d](https://github.com/scylladb/seastar/commit/31ec32dfe48a8bf80a860b943f4b5c2f47a9de92)

src/testing/test_runner.cc
bhalevy: Done in [31ec32d](https://github.com/scylladb/seastar/commit/31ec32dfe48a8bf80a860b943f4b5c2f47a9de92)

src/util/build_id.cc
bhalevy: Removed all SEASTAR_ASSERT calls in build_id.cc in [31ec32d](https://github.com/scylladb/seastar/commit/31ec32dfe48a8bf80a860b943f4b5c2f47a9de92)

include/seastar/util/process.hh
```diff
@@ -113,4 +113,6 @@ using seastar::process;
 using seastar::spawn_parameters;
 }
 
+const char* get_build_id();
+
```
avikivity: This should be in a private header.

src/testing/test_runner.cc
```diff
@@ -86,6 +87,8 @@ bool test_runner::start_thread(int ac, char** av) {
                 });
             };
 
+            auto s = get_build_id();
+            std::cout << fmt::format("build_id={}", s ?: "") << std::endl;
```
avikivity: Minor, but why not fmt::print?

src/util/build_id.cc
```diff
@@ -0,0 +1,139 @@
...
+    const char* get() const {
+        return buf;
+    }
+};
+
+#else // HAS_ELF
+
+struct build_id {
+    const char* get() const {
+        return nullptr;
```
avikivity: Are callers prepared for nullptr? Maybe better to return "unknown".

include/seastar/util/process.hh
bhalevy: Okay. I'm not sure why you don't want it to be public in the library, but I don't feel that strongly about it either.  It just seems wasteful that apps like scylla would need to reimplement this on their own.

src/testing/test_runner.cc
bhalevy: There is also a printout of the random seed that uses std::cout so I followed it by convention (and there are also errors printed into std::cerr).
I'm not sure if it would interleave well with fmt::print.

src/util/build_id.cc
bhalevy: okay

bhalevy:
Force-pushed with the following changes since last push (4dc92a7):

- **Header moved to internal**: Deleted `include/seastar/util/build_id.hh`. New minimal header at `include/seastar/util/internal/build_id.hh` declaring `seastar::internal::get_build_id()` returning `const char*`. Removed from `process.hh`.
- **`get_build_id()` never returns null**: Returns `"unknown"` on non-ELF platforms or when `dl_iterate_phdr` fails, instead of `nullptr`. The `SEASTAR_ASSERT` on init failure has been removed.
- **BuildId always printed**: Removed the conditional guard — `(BuildId: ...)` suffix is now unconditionally appended since `get_build_id()` always returns a valid string.
- **`test_runner.cc` simplified**: Uses `fmt::format()` directly, no null check needed.
- **`build_id.cc` namespace**: Uses `namespace seastar::internal`.
- **`backtrace.cc`**: Combined two `fmt::format_to` calls into a single format string.

include/seastar/util/build_id.hh
bhalevy: Done. Moved to `include/seastar/util/internal/build_id.hh` with the symbol in `namespace seastar::internal`.

include/seastar/util/build_id.hh
bhalevy: Removed. The `build_id` class and its `operator<<` are gone. Now just `const char* get_build_id()` — no special formatting needed.

src/testing/test_runner.cc
bhalevy: Done. Using `fmt::format()` now.

src/util/build_id.cc
bhalevy: Done. Removed the assert. `get_build_id()` now returns `"unknown"` if `dl_iterate_phdr` fails or on non-ELF platforms, so it never crashes.

include/seastar/util/process.hh
bhalevy: Moved to include/seastar/util/internal/build_id.hh in [46b9a89](https://github.com/scylladb/seastar/commit/46b9a891c36e2fd3bf28972d7027a7805864a759)

src/util/build_id.cc
bhalevy: Done in [46b9a89](https://github.com/scylladb/seastar/commit/46b9a891c36e2fd3bf28972d7027a7805864a759)

src/testing/test_runner.cc
avikivity: It should work (fmt::print will accept cout as an input).

But you can keep it like this.

PR #3355, opened 2026-04-13
smp: make smp::count non-static

avikivity:
smp::count is now a static variable as a convenience - it can be accessed from
anywhere without requiring any object to access it. But that means we can only
have one smp::count per process. This stands in the way for running multiple
Seastar instances in a single process, which is desirable for testing (cluster-in-a-box).

Remove this limitation by making smp::count a private non-static member named
_shard_count. smp::count is kept for compatibility, but deprecated. Similarly smp::all_cpus()
gains a non-static equivalent smp::all_shards().

To simplify access from within reactor threads, which is the vast majority of accesses,
we add convenience this_smp(), this_smp_shard_count(), and this_smp_all_shards().

Two uses of smp::count before it is initialized are converted to FIXMEs.

The tree is converted to use the new accessors.

### DIFF
--- a/apps/io_tester/io_tester.cc
+++ b/apps/io_tester/io_tester.cc
@@ -202,7 +202,7 @@ class shard_config {
     std::unordered_set<unsigned> _shards;
 public:
     shard_config()
-        : _shards(boost::copy_range<std::unordered_set<unsigned>>(boost::irange(0u, smp::count))) {}
+        : _shards(boost::copy_range<std::unordered_set<unsigned>>(boost::irange(0u, this_smp_shard_count()))) {}
     shard_config(std::unordered_set<unsigned> s) : _shards(std::move(s)) {}
 
     bool is_set(unsigned cpu) const {
@@ -647,7 +647,7 @@ private:
         return open_file_dma(name, flags).then([this] (auto f) {
             _file = std::move(f);
             return _file.size().then([this] (uint64_t size) {
-                auto shard_area_size = align_down<uint64_t>(size / smp::count, 1 << 20);
+                auto shard_area_size = align_down<uint64_t>(size / this_smp_shard_count(), 1 << 20);
                 if (_config.offset_in_bdev + _config.file_size > shard_area_size) {
                     throw std::runtime_error("Data doesn't fit the blockdevice");
                 }
@@ -897,7 +897,7 @@ private:
     uint64_t max_concurrency() const {
         // When we have many files it is easy to exceed the limit of open file descriptors.
         // To avoid that the limit is divided between shards (leaving some room for other jobs).
-        return static_cast<uint64_t>((1024u / smp::count) * 0.8);
+        return static_cast<uint64_t>((1024u / this_smp_shard_count()) * 0.8);
     }
 
     bool all_files_removed() const {
@@ -1159,7 +1159,7 @@ struct convert<job_config> {
         // constant) disk space between workloads. Each shard inside the
         // workload thus uses its portion of the assigned space.
         if (node["data_size"]) {
-            const uint64_t per_shard_bytes = node["data_size"].as<byte_size>().size / smp::count;
+            const uint64_t per_shard_bytes = node["data_size"].as<byte_size>().size / this_smp_shard_count();
             cl.file_size = align_up<uint64_t>(per_shard_bytes, extent_size_hint_alignment);
         } else if (cl.type == request_type::append) {
             cl.file_size = 0;
@@ -1261,7 +1261,7 @@ static void show_results(sharded<context>& ctx) {
     YAML::Emitter out;
     out << YAML::BeginDoc;
     out << YAML::BeginSeq;
-    for (unsigned i = 0; i < smp::count; ++i) {
+    for (unsigned i = 0; i < this_smp_shard_count(); ++i) {
         out << YAML::BeginMap;
         out << YAML::Key << "shard" << YAML::Value << i;
         ctx.invoke_on(i, [&out] (auto& c) {
--- a/include/seastar/core/smp.hh
+++ b/include/seastar/core/smp.hh
@@ -89,7 +89,7 @@ struct smp_service_group_config {
 /// smp::submit_to() and similar calls. While it's easy to limit
 /// the caller's concurrency (for example, by using a semaphore),
 /// the concurrency at the remote end can be multiplied by a factor
-/// of smp::count-1, which can be large.
+/// of smp::shard_count()-1, which can be large.
 ///
 /// The class is called a service _group_ because it can be used
 /// to group similar calls that share resource usage characteristics,
@@ -302,8 +302,18 @@ class smp_message_queue;
 struct reactor_options;
 struct smp_options;
 
+/// A set of cooperating reactor threads.
+///
+/// An smp instance manages a set of reactor threads, sharing memory and
+/// communicating via messages. The number of threads is determined
+/// at construction time and does not change over the lifetime of the smp
+/// instance.
+///
+/// Multiple smp instances may exist in the same process, typically for
+/// testing purposes.
 class smp : public std::enable_shared_from_this<smp> {
     alien::instance& _alien;
+    unsigned _shard_count = 0;
     std::vector<posix_thread> _threads;
     std::vector<std::function<void ()>> _thread_loops; // for dpdk
     std::optional<std::barrier<>> _all_event_loops_done;
@@ -314,6 +324,7 @@ class smp : public std::enable_shared_from_this<smp> {
     std::unique_ptr<smp_message_queue*[], qs_deleter> _qs_owner;
     static thread_local smp_message_queue**_qs;
     static thread_local std::thread::id _tmain;
+    static inline thread_local smp* _this_smp = nullptr;
     bool _using_dpdk = false;
     std::vector<unsigned> _shard_to_numa_node_mapping;
 
@@ -323,13 +334,17 @@ public:
     explicit smp(alien::instance& alien);
     ~smp();
     void configure(const smp_options& smp_opts, const reactor_options& reactor_opts);
+
+    /// The number of shards available in this `smp` instance. Does not change over the lifetime of the instance.
+    unsigned shard_count() const { return _shard_count; }
+
     void cleanup() noexcept;
     void cleanup_cpu();
     void arrive_at_event_loop_end();
     void join_all();
     static bool main_thread() { return std::this_thread::get_id() == _tmain; }
 
-    /// \returns A integer span of size smp::count, with nth integer being the ID of nth shard's NUMA node.
+    /// \returns A integer span of size smp::shard_count(), with nth integer being the ID of nth shard's NUMA node.
     std::span<const unsigned> shard_to_numa_node_mapping() const noexcept;
 
     /// Runs a function on a remote core.
@@ -390,8 +405,22 @@ public:
     }
     static bool poll_queues();
     static bool pure_poll_queues();
+
+    /// Returns a range of all shard IDs.
+    ///
+    /// Returns a range of all shard IDs (a range with a value_type
+    /// of some unspecified unsigned type) in this `smp` instance.
+    /// Order is unspecified. Does not change over the lifetime of the instance.
+    std::ranges::range auto all_shards() const noexcept {
+        return std::views::iota(0u, _shard_count);
+    }
+
+    [[deprecated("use smp::all_shards instead")]]
     static std::ranges::range auto all_cpus() noexcept {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
         return std::views::iota(0u, count);
+#pragma GCC diagnostic pop
     }
 private:
     template <typename Func>
@@ -423,7 +452,7 @@ public:
     static future<> invoke_on_all(smp_submit_to_options options, Func&& func) noexcept {
         static_assert(std::is_same_v<future<>, typename futurize<std::invoke_result_t<Func>>::type>, "bad Func signature");
         static_assert(std::is_nothrow_move_constructible_v<Func>);
-        return parallel_for_each(all_cpus(), [options, &func] (unsigned id) {
+        return parallel_for_each(this_smp().all_shards(), [options, &func] (unsigned id) {
             return smp::copy_and_submit_to(id, options, func);
         });
     }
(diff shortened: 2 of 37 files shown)


avikivity:
v2: update module exports in patch 4

avikivity:

nyh:
> smp::count is kept for compatibility, but deprecated.

I'm worried this deprecation will be annoying. I assume virtually every Seastar application uses `smp::count`, and all of them will need to be changed now to use the new `this_smp_shard_count()` - or face a deluge of annoying deprecation warnings. In the ScyllaDB project, for example, there are almost 400 references to smp::count.

I wonder, wouldn't it be easier to leave `smp::count` working, without deprecating it?

I think it should be fairly easy to implement `smp::count` based on your new `this_smp_shard_count:

`smp::count` can be an object, with a conversion operator, so it can be read like a variable but actually does something else when read:
```c++
struct count_t {
    // magic read
    operator unsigned() const {
        return this_smp_shard_count()
    }
    // write forbidden
    count_t& operator=(unsigned) = delete;
};

inline count_t count;
```

avikivity:
> > smp::count is kept for compatibility, but deprecated.
> 
> I'm worried this deprecation will be annoying. I assume virtually every Seastar application uses `smp::count`, and all of them will need to be changed now to use the new `this_smp_shard_count()` - or face a deluge of annoying deprecation warnings. In the ScyllaDB project, for example, there are almost 400 references to smp::count.

It takes 10 minute to have AI fix them up.

> 
> I wonder, wouldn't it be easier to leave `smp::count` working, without deprecating it?
> 
> I think it should be fairly easy to implement `smp::count` based on your new `this_smp_shard_count:
> 
> `smp::count` can be an object, with a conversion operator, so it can be read like a variable but actually does something else when read:
> 
> ```c++
> struct count_t {
>     // magic read
>     operator unsigned() const {
>         return this_smp_shard_count()
>     }
>     // write forbidden
>     count_t& operator=(unsigned) = delete;
> };
> 
> inline count_t count;
> ```

The semantics changed. smp::count works from alien threads (and is in fact used from alien threads, see first patch). this_smp_shard_count() doesn't work from alien threads, and the name suggest so. A semantic change requires a new name.

nyh:
> > > smp::count is kept for compatibility, but deprecated.
> > 
> > 
> > I'm worried this deprecation will be annoying. I assume virtually every Seastar application uses `smp::count`, and all of them will need to be changed now to use the new `this_smp_shard_count()` - or face a deluge of annoying deprecation warnings. In the ScyllaDB project, for example, there are almost 400 references to smp::count.
> 
> It takes 10 minute to have AI fix them up.

That's not my point. My point was that until you fix them up, your project gets hundreds of deprecation warnings. So you are forced to now do those fixes now. It's not like a deprecation which gives 3 annoying warnings and you can just ignore for a while.

> The semantics changed. smp::count works from alien threads (and is in fact used from alien threads, see first patch). this_smp_shard_count() doesn't work from alien threads, and the name suggest so. A semantic change requires a new name.

So the new `smp::count` class can have another if, to check if it's an alien thread. Or something else. I think this can be solved (maybe by 10 minutes of AI, like you said).

I wonder if users in other projects care about the deprecation warnings that I'm scared of. If they don't, then feel free to ignore my proposal.

avikivity:
> > > > smp::count is kept for compatibility, but deprecated.
> > > 
> > > 
> > > I'm worried this deprecation will be annoying. I assume virtually every Seastar application uses `smp::count`, and all of them will need to be changed now to use the new `this_smp_shard_count()` - or face a deluge of annoying deprecation warnings. In the ScyllaDB project, for example, there are almost 400 references to smp::count.
> > 
> > 
> > It takes 10 minute to have AI fix them up.
> 
> That's not my point. My point was that until you fix them up, your project gets hundreds of deprecation warnings. So you are forced to now do those fixes now. It's not like a deprecation which gives 3 annoying warnings and you can just ignore for a while.

I don't understand how you can do that. How can you make smp::count give a compile-time warning when accessed from an alien thread, but not from a reactor thread?


> 
> > The semantics changed. smp::count works from alien threads (and is in fact used from alien threads, see first patch). this_smp_shard_count() doesn't work from alien threads, and the name suggest so. A semantic change requires a new name.
> 
> So the new `smp::count` class can have another if, to check if it's an alien thread. Or something else. I think this can be solved (maybe by 10 minutes of AI, like you said).


This is a runtime check. It's not guaranteed to trigger, and it won't move users away from the bad API.

> 
> I wonder if users in other projects care about the deprecation warnings that I'm scared of. If they don't, then feel free to ignore my proposal.

If users don't care about deprecation warnings, they don't work. It only works because users care. You're saying that if the deprecation warning doesn't fulfill its intended purpose, then I can ignore  your proposal not to have a deprecation warning.

nyh:
> I don't understand how you can do that. How can you make smp::count give a compile-time warning when accessed from an alien thread, but not from a reactor thread?

I don't understand why you need a compile-time warning. I'm proposing to make it compatible with the old smp::count, and the old smp::count worked in alien threads - it didn't give a compile time warning.

I'll be honest - I didn't carefully researched your implementation. Why can't smp::count read from a per-thread variable which is initialized on every thread to the same default count?

> This is a runtime check. It's not guaranteed to trigger, and it won't move users away from the bad API.

I don't understand why it's a "bad API". It's bad for applications that want to run multiple copies of themselves in one process. I suspect that most just don't want to do it.

But if you really think that the old API is **bad**, then I agree - deprecation warnings is the way to go.

avikivity:
> > I don't understand how you can do that. How can you make smp::count give a compile-time warning when accessed from an alien thread, but not from a reactor thread?
> 
> I don't understand why you need a compile-time warning. I'm proposing to make it compatible with the old smp::count, and the old smp::count worked in alien threads - it didn't give a compile time warning.

The old smp::count was indeed compatible with alien threads, therefore it did not need to warn.

The new system cannot have a global smp::count that works with alien threads. It's just impossible. You can have two Seastar instances in the same process with different smp::counts. How will smp::count know what to return?

It's possible to replace the compile-time warning with a run-time warning. But I think compile-time warnings are superior.



> 
> I'll be honest - I didn't carefully researched your implementation. Why can't smp::count read from a per-thread variable which is initialized on every thread to the same default count?

That's what this_smp_shard_count does. But it doesn't work from alien threads. Which smp::count will this_smp_shard_count return from an alien thread?

> 
> > This is a runtime check. It's not guaranteed to trigger, and it won't move users away from the bad API.
> 
> I don't understand why it's a "bad API". It's bad for applications that want to run multiple copies of themselves in one process. I suspect that most just don't want to do it.

Let's say I have a database written with the expectation that there is one Seastar instance per process, and so has numerous smp::count references, including in alien threads. Let's say I want to convert it to run tests with multiple Seastar instances per process. Without a warning, how do I know which smp::count instances to fix?

> 
> But if you really think that the old API is **bad**, then I agree - deprecation warnings is the way to go.

The old API is not safe to use without auditing every call site.

nyh:
Isn't each alien thread tied to one particular "instance" of Seatar when it is created? If so, can't we copy the originating thread's smp::count into the new alien thread?

avikivity:
> Isn't each alien thread tied to one particular "instance" of Seatar when it is created?

No, alien threads are independent. 

>  If so, can't we copy the originating thread's smp::count into the new alien thread?

An alien thread is created with std::thread or std::jthread, we don't have any control over its TLS variables. Moreover, an alien thread may create more threads.

avikivity:
@nyh let's resume the discussion

nyh:
> > Isn't each alien thread tied to one particular "instance" of Seatar when it is created?
> 
> No, alien threads are independent.

I'm not sure what you mean - how can multiple Seastar smps - each a set of threads which is supposedly not aware of each other - be able to share a single alien thread? Which of them created this alien thread?
If they don't share an alien thread, then my question is still relevant - an alien thread "belongs" to a single SMP, so could return its smp::count.

But I am afraid I still don't understand something even more basic - why this is needed. Why do Alien threads actually need to know smp::count? What do we lose if an alien thread calls smp::count and gets the default value (e.g., 0)?

avikivity:
> > > Isn't each alien thread tied to one particular "instance" of Seatar when it is created?
> > 
> > 
> > No, alien threads are independent.
> 
> I'm not sure what you mean - how can multiple Seastar smps - each a set of threads which is supposedly not aware of each other - be able to share a single alien thread? 

In general, they would not (though it's not impossible).


> Which of them created this alien thread? If they don't share an alien thread, then my question is still relevant - an alien thread "belongs" to a single SMP, so could return its smp::count.

The alien thread is created outside Seastar. Typically main() launches an std::thread or std::jthread. It's not created via a Seastar API.


> 
> But I am afraid I still don't understand something even more basic - why this is needed. Why do Alien threads actually need to know smp::count?

See the first patch in the series for an example.


> What do we lose if an alien thread calls smp::count and gets the default value (e.g., 0)?

There are two use cases for alien:

1. reactor receives a request, wakes an alien thread, waits (asynchronously) for a reply
2. alien thread receives a request, dispatches to some_id % smp::count, waits for a reply (or mabe the reactor sends the reply)

It's the second case that's broken.

avikivity:
@nyh let's resume the discussion

include/seastar/core/smp.hh
```diff
@@ -484,16 +494,45 @@ public:
...
+this_smp() noexcept {
+    return smp::this_smp();
+}
+
+/// Returns the number of shards in the current smp instance.
+/// May only be called from a reactor thread.
+inline
+unsigned
+this_smp_shard_count() noexcept {
+    return this_smp().shard_count();
```
nyh, line 545: Nitpick: I wonder if we need `this_smp_shard_count()` to save two characters from `this_smp().shard_count()`. But I guess it can't hurt.

apps/io_tester/io_tester.cc
```diff
@@ -202,7 +202,7 @@ class shard_config {
     std::unordered_set<unsigned> _shards;
 public:
     shard_config()
-        : _shards(boost::copy_range<std::unordered_set<unsigned>>(boost::irange(0u, smp::count))) {}
+        : _shards(boost::copy_range<std::unordered_set<unsigned>>(boost::irange(0u, this_smp_shard_count()))) {}
```
nyh, line 205: A thought: Do we really want to "advertise" the concept of "this smp" vs. "some other smp"? I believe that normal applications - not special test harnesses - would find this concept of "this smp" a bit bizarre. We could have called this function just `seastar::shard_count()` instead of `seastar::this_smp_shard_count()`, and make the code using it a little bit shorter and less bizarre.

nyh (approved):
I approve. Had some minor thoughts about naming, but it's not critical.

I would have liked to see someone outside ScyllaDB review this, that the massive changes required (in the ScyllaDB source code, we have about 400 uses of smp::count) will be ok with them. But if nobody screams, I think we can merge this.

nyh:
> I approve. Had some minor thoughts about naming, but it's not critical.

@avikivity as soon as you tell me you thought about the naming, and you like it as-is (or want to change it), I can merge this PR.

apps/io_tester/io_tester.cc
avikivity: The convention comes from [this_thread](https://en.cppreference.com/cpp/thread/sleep_for).

If we just call it shard_count(), it will be easy to misuse, precisely because it's typical to have just one smp.

this_smp_shard_count forces users to think about it, and maybe read the documentation, before using it.

avikivity:
> I approve. Had some minor thoughts about naming, but it's not critical.
> 
> I would have liked to see someone outside ScyllaDB review this, that the massive changes required (in the ScyllaDB source code, we have about 400 uses of smp::count) will be ok with them. But if nobody screams, I think we can merge this.

@travisdowns please review

include/seastar/core/smp.hh
```diff
@@ -323,13 +325,14 @@ public:
     explicit smp(alien::instance& alien);
     ~smp();
     void configure(const smp_options& smp_opts, const reactor_options& reactor_opts);
+    unsigned shard_count() const { return _shard_count; }
```
travisdowns, line 339: Missing doc.

include/seastar/core/smp.hh
```diff
@@ -390,8 +393,15 @@ public:
     }
     static bool poll_queues();
     static bool pure_poll_queues();
+    std::ranges::range auto all_shards() noexcept {
```
travisdowns: Missing doc.

include/seastar/core/smp.hh
travisdowns: I also question the need of the free function. Maybe it can just be called shard_count(), as everyone should either not even be aware of the "multiple applications" functionality, or otherwise understand that these free functions always refer to the current application.

include/seastar/core/smp.hh
```diff
@@ -484,16 +494,45 @@ public:
...
+/// Returns the smp object used for cross-shard communications.
+/// May only be called from a reactor thread.
+inline
+smp&
+this_smp() noexcept {
+    return smp::this_smp();
+}
+
+/// Returns the number of shards in the current smp instance.
+/// May only be called from a reactor thread.
```
travisdowns, line 541: Should we additionally document that it is guaranteed not to change for the lifetime of the application/thread?

Otherwise it is harder to use, e.g., passing a fixed value to an alien thread may become stale.

travisdowns (approved):

travisdowns:
`this_smp_shard_count` is long and ugly but in any I case I approve, though I left some minor comments

include/seastar/core/smp.hh
avikivity: I think we should.

Will we want one day to dynamically add/remove shards?

That way lies madness (if indeed we are not there already).

crackcomm: > Will we want one day to dynamically add/remove shards?

I was wondering about that and if that'd warrant another rename and if `current_shard_count` wasn't a better choice because of that.

avikivity: I think it's unrealistic. Nothing in seastar::sharded is prepared for it.

Moreover, all Seastar applications are sharded. Add a new node and remove an old node, or restart a node with a different shard count.

avikivity:
- documented smp::all_shards()
 - made it const
 - documented smp::shard_count()
 - documented smp itself and the fact that the shard count doesn't
   change over its lifetime

travisdowns:
LGTM. @avikivity are there any outstanding conversations from your PoV? If not I can merge.

avikivity:
I don't have further plans in this area (but undoubtedly actually trying to create a cluster-in-a-box will uncover more statics)

travisdowns:
> I don't have further plans in this area (but undoubtedly actually trying to create a cluster-in-a-box will uncover more statics)

Sorry I mean I think all the feedback in this review is addressed, so I am going to merge this, unless you tell me otherwise (e.g., because "No, I need to resolve XYZ convo with ABC first").

avikivity:
I believe everything was addressed.

PR #3357, opened 2026-04-14
ci: update GitHub Actions to Node.js 24 compatible versions

avikivity:
Node.js 20 actions are deprecated and will be forced to Node.js 24 starting June 2nd, 2026, with Node.js 20 removed from runners on September 16th, 2026. Update all actions to their latest major versions that support Node.js 24.

Actions updated across all workflow files:
- actions/checkout: v4 -> v6
- actions/setup-python: v5 -> v6
- hendrikmuhs/ccache-action: v1 -> v1.2
- docker/setup-buildx-action: v3 -> v4
- docker/build-push-action: v6 -> v7
- psf/black: 24.8.0 -> 26.3.1

Note: pre-commit/action@v3.0.1 internally uses actions/cache@v4 which still runs on Node.js 20. This is an upstream issue -- v3.0.1 is the latest release of pre-commit/action.

### DIFF
--- a/.github/workflows/alpinelinux.yaml
+++ b/.github/workflows/alpinelinux.yaml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
 
     - name: Install build dependencies
       run: |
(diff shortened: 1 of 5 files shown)


mykaul:
@avikivity - please see https://github.com/scylladb/scylladb/pull/29421 - which is similar, but moving to hashes, instead of versions.

avikivity:
v2: rebased

avikivity:
Obsoleted by abefa0b1a3d

PR #3359, opened 2026-04-16
httpd: rewrite do_accepts/do_accept_one as coroutines

ScyllaPiotr:
## Summary
- Coroutine rewrite of `do_accepts()` and `do_accept_one()` in `httpd.cc`
- No behavioral changes — prepares the ground for future scheduling group support
- Split from #3289 per reviewer request

### DIFF
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -438,15 +438,31 @@ future<> http_server::stop() {
     return tasks_done;
 }
 
-// FIXME: This could return void
+// This is a named class member coroutine, so that 'this', 'which' and 'tls'
+// live safely in the coroutine frame, therefore `accept_loop()` can safely suspend
+// at `co_await do_accept_one()`.
+future<> http_server::accept_loop(int which, bool tls) {
+    while (!_task_gate.is_closed()) {
+        try {
+            co_await do_accept_one(which, tls);
+        } catch (const gate_closed_exception&) {
+            co_return;
+        } catch (const std::system_error& e) {
+            // We expect a ECONNABORTED when http_server::stop is called,
+            // no point in warning about that.
+            if (e.code().value() != ECONNABORTED) {
+                hlogger.error("accept failed: {}", e);
+            }
+        } catch (...) {
+            hlogger.error("accept failed: {}", std::current_exception());
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/http/httpd.cc
```diff
@@ -438,15 +438,31 @@ future<> http_server::stop() {
...
 future<> http_server::do_accepts(int which, bool tls) {
     (void)try_with_gate(_task_gate, [this, which, tls] {
-        return keep_doing([this, which, tls] {
-            return try_with_gate(_task_gate, [this, which, tls] {
-                return do_accept_one(which, tls);
-            });
-        }).handle_exception_type([](const gate_closed_exception& e) {});
-    }).handle_exception_type([](const gate_closed_exception& e) {});
+        return accept_loop(which, tls);
+    }).handle_exception_type([](const gate_closed_exception&) {});
```
avikivity, line 465: But accept_loop won't throw gate_closed_exception now, will it?

avikivity, line 465: And why have both accept_loop and do_accepts?

ScyllaPiotr: Inlining `accept_loop()` as a coroutine lambda turned out to cause a use-after-free: the lambda object (and its captures) can be destroyed once `do_accepts()` returns `make_ready_future<>()`, while the coroutine frame still references them through a dangling pointer. Reinstated `accept_loop()` as a named member coroutine where `this`, `which` and `tls` are function parameters safely copied into the coroutine frame.

ScyllaPiotr: Correct, but it catches what `try_with_gate` itself throws if the gate is already closed at entry.

avikivity: Would be nicer to just check the gate on entry.

avikivity: Could it even happen? That the gate is closed as we start?

avikivity: Could have instructed the lambda to copy the captures into the frame: https://github.com/scylladb/seastar/blob/master/doc/lambda-coroutine-fiasco.md#solution-c23-and-up

But that's C++23 only (will make it soon mandatory)

ScyllaPiotr: No, only on some future misuse. For regression safety I propose to keep the guard and to additionally log: #3372

Issue #3361, opened 2026-04-19
QUESTION: Non-blocking `ftruncate` on `file_desc`


crackcomm:
I have a `file_desc` that is opened only once (on tmpfs), but `ftruncate` is called multiple times during its lifetime. I don't want the call to block the reactor thread (it is fast most of the time, but not truly non-blocking). What are my choices?

1. Is it possible to submit a syscall to the reactor's `_thread_pool`?

I couldn't find a way to do that. It would be ideal.

2. Should I use `file` instead?

I need access to the file descriptor for `mmap`, but `file` does not expose it.

3. Should I use `make_file_impl`? This way I can access the `fd`.

This would solve the problem of non-blocking truncate, but not the blocking `open` and `fstat` calls.

---

I would be happy to submit a PR if there is a clear, desirable solution.

---

Related: #346 (Note: `ftruncate` is no longer called in the constructor of `append_challenged_posix_file_impl`, so it could probably be closed.)

crackcomm:
Maybe even better would be to just get `file_desc` from `file` (for `mmap`).

avikivity:
Use file, file_desc is semi-internal.

avikivity:
If you want mmap (why??????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????) then add mmap to file.

crackcomm:
Long story short, for shared memory.

Should I add [`map`](https://github.com/scylladb/seastar/blob/master/include/seastar/core/posix.hh#L332-L337) to `file` and add implementation to `posix_file_impl`, and submit a PR?

If yes, should it be an async function (through reactor's pool)? AFAIK `mmap` can be a blocking call if `MAP_POPULATE` is set (I'm not using it, not sure).

avikivity:
Only you can answer that.

For sure it should be async, since mmap will block on the mm semaphore even without MAP_POPULATE.

crackcomm:
[...]

PR #3363, opened 2026-04-19
file: add mmap support to file

crackcomm:
Introduce file::mmap() and a file_mapping class to support memory-mapped files. This is primarily intended for shared memory use cases (e.g., tmpfs) where data resides in RAM.

Both mmap and unmap are performed asynchronously via the reactor's  thread pool to avoid blocking the reactor thread on kernel memory  management locks. file_mapping requires an explicit call to unmap()  to ensure this asynchronous teardown.

A flush() method is included to call msync(), ensuring consistency  between the kernel page cache and Seastar's O_DIRECT DMA operations,  which is particularly useful for testing.

Ref #3361

### DIFF
--- a/include/seastar/core/file-types.hh
+++ b/include/seastar/core/file-types.hh
@@ -22,8 +22,11 @@
 #pragma once
 
 #include <fcntl.h>
+#include <sys/mman.h>
 #include <sys/stat.h>
 #include <type_traits>
+
+#include <seastar/util/bool_class.hh>
 
 namespace seastar {
 
@@ -160,6 +163,37 @@ inline constexpr file_permissions operator|(file_permissions a, file_permissions
 inline constexpr file_permissions operator&(file_permissions a, file_permissions b) {
     return file_permissions(std::underlying_type_t<file_permissions>(a) & std::underlying_type_t<file_permissions>(b));
 }
+
+/// \brief Memory protection flags for mmap.
+///
+/// \see file::mmap()
+enum class mmap_prot {
+    none = PROT_NONE,
+    read = PROT_READ,
+    write = PROT_WRITE,
+    execute = PROT_EXEC,
+};
+
+inline constexpr mmap_prot operator|(mmap_prot a, mmap_prot b) {
+    return mmap_prot(std::underlying_type_t<mmap_prot>(a) | std::underlying_type_t<mmap_prot>(b));
+}
+
+inline void operator|=(mmap_prot& a, mmap_prot b) {
+    a = (a | b);
+}
+
+inline constexpr mmap_prot operator&(mmap_prot a, mmap_prot b) {
+    return mmap_prot(std::underlying_type_t<mmap_prot>(a) & std::underlying_type_t<mmap_prot>(b));
+}
+
+inline void operator&=(mmap_prot& a, mmap_prot b) {
+    a = (a & b);
+}
+
+/// \brief Visibility flag for mmap (`MAP_PRIVATE` when true, `MAP_SHARED` when false).
+///
+/// \see file::mmap()
+using mmap_private = bool_class<struct mmap_private_tag>;
 
 /// @}
 
--- a/include/seastar/core/file.hh
+++ b/include/seastar/core/file.hh
@@ -115,6 +115,87 @@ class file_handle;
 class file_data_sink_impl;
 class file_data_source_impl;
 
+/// \brief A memory mapped region of a file.
+///
+/// Represents a memory region created by \ref seastar::file::mmap().
+///
+/// \warning This API is strongly discouraged for general-purpose file I/O.
+/// Accessing file-backed memory mappings relies on the kernel page cache and
+/// can trigger major page faults. These faults will silently block the
+/// Seastar reactor thread, bypassing Seastar's I/O scheduler and severely
+/// degrading overall system performance. Seastar applications should prefer
+/// \c O_DIRECT DMA interfaces (e.g., \ref dma_read) for persistent storage.
+///
+/// This facility is provided specifically for shared memory use cases (e.g.,
+/// mapping files on \c tmpfs or \c shm), where data resides entirely in RAM
+/// and blocking disk I/O is not a concern.
+///
+/// Because \c munmap requires acquiring kernel \c mmap_sem locks, it cannot
+/// be executed synchronously in a destructor without risking blocking the reactor.
+/// Therefore, the user must explicitly await the asynchronous \ref unmap()
+/// method before this object is destroyed.
+class file_mapping {
+    void* _addr = nullptr;
+    size_t _length = 0;
+
+    file_mapping(void* addr, size_t length) noexcept
+        : _addr(addr), _length(length) {}
+
+    friend class posix_file_impl;
+
+public:
+    /// Constructs an uninitialized file mapping.
+    file_mapping() noexcept = default;
+    file_mapping(const file_mapping&) = delete;
+    file_mapping& operator=(const file_mapping&) = delete;
+
+    file_mapping(file_mapping&& other) noexcept
+        : _addr(std::exchange(other._addr, nullptr))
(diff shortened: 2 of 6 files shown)


include/seastar/core/file.hh
```diff
@@ -115,6 +115,89 @@ class file_handle;
...
+        if (this != &other) {
+            SEASTAR_ASSERT(!_addr && !_length && "file_mapping assigned to without unmapping");
+            _addr = std::exchange(other._addr, nullptr);
+            _length = std::exchange(other._length, 0);
+        }
+        return *this;
+    }
+
+    /// \cond internal
+    file_mapping(void* addr, size_t length) noexcept
```
avikivity: private?

include/seastar/core/file.hh
```diff
@@ -115,6 +115,89 @@ class file_handle;
...
+        : _addr(addr), _length(length) {}
+    /// \endcond
+
+    ~file_mapping() noexcept {
+        SEASTAR_ASSERT(!_addr && !_length && "file_mapping destroyed without unmapping");
```
avikivity: Maybe too harsh, log a warning (in .cc), not worth a crash over this.

src/core/file-impl.hh
```diff
@@ -106,7 +106,8 @@ public:
     future<int> ioctl_short(uint64_t cmd, void* argp) noexcept override;
     future<int> fcntl(int op, uintptr_t arg) noexcept override;
     future<int> fcntl_short(int op, uintptr_t arg) noexcept override;
-    virtual future<> allocate(uint64_t position, uint64_t length) noexcept override;
+    future<> allocate(uint64_t position, uint64_t length) noexcept override;
```
avikivity: Avoid unrelated changes.

src/core/file.cc
```diff
@@ -1423,6 +1456,14 @@ future<> file::allocate(uint64_t position, uint64_t length) noexcept {
   }
 }
 
+future<file_mapping> file::mmap(size_t length, int prot, int flags, size_t offset) noexcept {
+  try {
+    return _file_impl->mmap(length, prot, flags, offset);
+  } catch (...) {
+    return current_exception_as_future<file_mapping>();
+  }
```
avikivity: 4-space indents.

crackcomm: It's a copy-paste, see function above and below - I left it as it is for consistency, but it's your call.

include/seastar/core/file.hh
crackcomm: It certainly could be but it might provide async `unmap` for `mmap_area`. It's your call but both are fine with me.

crackcomm: On the second thought, if it was mmap-ed synchronously, it'll probably be unmapped synchronously. I'll make it private.

include/seastar/core/file.hh
```diff
@@ -159,6 +240,7 @@ public:
     virtual future<int> fcntl(int op, uintptr_t arg) noexcept;
     virtual future<int> fcntl_short(int op, uintptr_t arg) noexcept;
     virtual future<> allocate(uint64_t position, uint64_t length) = 0;
+    virtual future<file_mapping> mmap(size_t length, int prot, int flags, size_t offset) noexcept;
```
xemul: Passing raw libc/kernel flags into seastar API is not very welcome. See e.g. open_flags, access_flags, or [rename flags](https://github.com/scylladb/seastar/pull/3353#discussion_r3084980193)

crackcomm: I added `mmap_flags` and `mmap_prot` to `file-types.hh`. Let me know if you want them moved/renamed.

include/seastar/core/file-types.hh
```diff
@@ -160,6 +161,60 @@ inline constexpr file_permissions operator&(file_permissions a, file_permissions
...
+inline void operator&=(mmap_prot& a, mmap_prot b) {
+    a = (a & b);
+}
+
+// Mapping flags for mmap
+enum class mmap_flags {
+    shared = MAP_SHARED,
+    private_ = MAP_PRIVATE,
+    fixed = MAP_FIXED,
+    anonymous = MAP_ANONYMOUS,
```
xemul: Since the new API is to map files, the `anonymous` bit looks completely alien.
Also, since the `file::mmap()` method doesn't have the `address` argument, `fixed` bit is also not applicable.
So we end up having only public/private pair, and since they are mutually exclusive the `using map_private = bool_class<map_flags_tag>` argument seems to fit much better than bitfield-style flags

include/seastar/core/file-types.hh
```diff
@@ -160,6 +161,60 @@ inline constexpr file_permissions operator&(file_permissions a, file_permissions
     return file_permissions(std::underlying_type_t<file_permissions>(a) & std::underlying_type_t<file_permissions>(b));
 }
 
+// Memory protection flags for mmap
+enum class mmap_prot {
+    none = PROT_NONE,
+    read = PROT_READ,
+    write = PROT_WRITE,
+    execute = PROT_EXEC,
+    read_write = read | write,
```
xemul: Since there's operator|, what's this read_write alias for?

crackcomm: strictly for convenience but if you think it doesn't fit, I'll remove it

include/seastar/core/file-types.hh
crackcomm: That makes sense, I'll apply patches later today.

xemul (approved):

src/core/file.cc
```diff
@@ -1095,6 +1106,35 @@ posix_file_handle_impl<FileImpl>::to_file() && {
     return ret;
 }
 
+file_mapping::~file_mapping() noexcept {
+    if (_addr || _length) {
+        seastar_logger.warn("file_mapping was not unmapped before destruction, contact support");
```
avikivity: We should munmap it however.

src/core/file.cc
```diff
@@ -1423,6 +1463,14 @@ future<> file::allocate(uint64_t position, uint64_t length) noexcept {
   }
 }
 
+future<file_mapping> file::mmap(size_t length, mmap_prot prot, mmap_private priv, size_t offset) noexcept {
+    try {
+        return _file_impl->mmap(length, prot, priv, offset);
+    } catch (...) {
+        return current_exception_as_future<file_mapping>();
```
avikivity: Since file_impl::unmap() is noexcept, this is unnecessary.

src/core/file.cc
crackcomm: I assume we don't want to `handle_exception()` (discard) there, it should never happen in reality (AFAIK, kernel failure?)

crackcomm: Should we keep the warning though, if it is completely safe to `run_in_background`?

crackcomm:
Rebased on top of master, only requested changes in `file.cc`.

src/core/file.cc
avikivity: No, better to run it in foreground and take the latency hit.

xemul: Suggestion to un it in foreground is not implemented. @crackcomm , do you disagree?

crackcomm: Not disagree, maybe misunderstand. I understood it as "keep the warning, user should use `unmap`". @avikivity did you mean something different?

crackcomm: Unless by "foreground" you meant a direct `::munmap` call in the destructor, block the thread, and let the user suffer the latency penalty for writing bad code.

crackcomm: This would make the destructor more deterministic, guaranteeing the resources to be released by the time the destructor returns. I agree with that, I'll adjust the warning accordingly.

avikivity: Yes, that's what I meant: don't leak the mapping, warn, and the user suffers the latency penalty (likely a stall even if no contention, if the mapping is very large).

PR #3367, opened 2026-04-25
task_profiler: add sampling profiler for suspended coroutine chains

gusev-p:
While investigating bottlenecks in ScyllaDB's strong consistency path,
we needed a way to see which suspended coroutine chains accumulate the
most samples over a profiling session — i.e., which fibers tend to stay
suspended.

The task_profiler is a per-shard sampling profiler for dormant tasks.
A periodic timer snapshots all currently-suspended task chains, walks
their waiting_task() links to collect source_location frames, and
accumulates per-stack sample counts in an unordered_map.  On stop(),
the map is streamed in folded-stack format consumable by flamegraph.pl.

It hooks into three points in the future/promise machinery:

 - on_task_suspend(): called from promise_base::set_task() when a task
   becomes dormant (suspended on a non-ready future).  Appends a
   task_entry to a dense vector and records the index in task::_prof_idx.

 - on_task_resume(): called from promise_base::make_ready() and clear()
   when a dormant task is about to be scheduled.  Records the child's
   resume_point in the parent entry (so recently-completed children
   appear as [completed] frames), then removes the entry via swap-and-pop.

 - mark_completion(): called from coroutine awaiter preemption paths
   where the awaited future is already available but need_preempt() is
   true.  The child was never tracked (it completed immediately), but
   this records its source_location in the parent entry.

sample_once() works in two phases:

 Phase 1 — identify leaves: walk each entry's waiting_task() chain and
 flag ancestors.  Early-break on already-flagged entries gives O(N)
 amortised cost.  Only leaves are sampled — ancestors' suspension is
 already accounted for by their leaves.

 Phase 2 — build stacks: for each leaf, walk waiting_task() from leaf to
 root collecting source_location frames, and increment the corresponding
 entry in the map.

The task::_prof_idx field (4 bytes) is placed after _sg (also 4 bytes),
filling an existing padding gap — zero per-task memory overhead.  All
three hook functions are [[gnu::always_inline]] with a single branch
(TLS pointer check or _prof_idx != UINT_MAX) predicted not-taken, so the
cost when the profiler is off is a single untaken branch per
suspend/resume.

An example of the flamegraph from a [test run](https://github.com/gusev-p/scylla/commits/async_profiler/):

out-333

### DIFF
--- a/include/seastar/core/task.hh
+++ b/include/seastar/core/task.hh
@@ -24,21 +24,29 @@
 #include <seastar/core/scheduling.hh>
 #include <seastar/util/backtrace.hh>
 
+#include <cstdint>
 #include <utility>
 #include <source_location>
 
 namespace seastar {
+
+class task_profiler;
 
 class task {
     std::source_location _resume_point = {};
 
 protected:
     scheduling_group _sg;
+    // Index into task_profiler's entries vector. UINT32_MAX means the
+    // task is not tracked. Placed right after _sg (both 4 bytes) to
+    // fill the padding gap — zero per-task memory overhead.
+    uint32_t _prof_idx = UINT32_MAX;
 
 private:
 #ifdef SEASTAR_TASK_BACKTRACE
     shared_backtrace _bt;
 #endif
+
 protected:
     // Task destruction is performed by run_and_dispose() via a concrete type,
     // so no need for a virtual destructor here. Derived classes that implement
@@ -64,6 +72,8 @@ public:
     void make_backtrace() noexcept {}
     shared_backtrace get_backtrace() const { return {}; }
 #endif
+
+    friend class task_profiler;
 };
 
 
--- a/include/seastar/core/task_profiler.hh
+++ b/include/seastar/core/task_profiler.hh
@@ -1,0 +1,120 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2026 ScyllaDB
+ */
+
+#pragma once
+
+#include <seastar/core/task.hh>
+
+#include <chrono>
+#include <cstdint>
+
+namespace seastar {
+
+template <typename CharType> class output_stream;
+
+/// \brief Per-shard async coroutine profiler.
+///
+/// Tracks suspended tasks via a vector of task entries indexed by
+/// task::_prof_idx. A periodic timer samples the entries, walks
+/// waiting_task() chains, and accumulates per-stack counts in a
+/// pmr unordered_map backed by a monotonic arena. On stop(), the
+/// accumulated data is formatted and written to the given streams.
+///
+/// Two complementary views of the same data are emitted by stop():
+///
+///  - raw stream: each stack's count is the total number of leaf
+///    observations across all ticks. A stack with many concurrent
+///    fibers blocked at the same place accumulates a proportionally
+///    larger count, even if those fibers were only blocked briefly.
+///
+///  - normalized stream: each stack's count is the number of distinct
+///    ticks during which the stack appeared on at least one fiber.
+///    This count is comparable across stacks regardless of fiber
+///    multiplicity, and is a better proxy for "fraction of wall-clock
+///    time this code path was blocked anywhere on the shard".
+///
+/// All methods operate on the current shard only. Use smp::invoke_on_all()
+/// to orchestrate across shards.
+class task_profiler {
+public:
+    struct session;
+
+    /// Start profiling session on this shard.
+    /// \param sampling_interval  Time between samples. Typical: 10-100ms.
+    static void start(std::chrono::steady_clock::duration sampling_interval);
+
+    /// Stop the profiling session, dump the accumulated samples, and
+    /// reset internal state. Writes folded-stack output to two streams:
+    /// \param raw_out         per-leaf sample counts (raw view)
+    /// \param normalized_out  per-tick presence counts (normalized view)
+    /// Both streams receive identical frame strings; only the trailing
+    /// count value differs. Each stream is independently consumable by
+    /// flamegraph.pl.
+    static future<> stop(output_stream<char>& raw_out,
+                         output_stream<char>& normalized_out);
+
+    /// Called from set_task() when a task becomes dormant (suspended on
+    /// a non-ready future). Fast path: single TLS pointer check,
+    /// predicted not-taken.
+    [[gnu::always_inline]]
+    static void on_task_suspend(task& t) noexcept {
+        if (auto* const session = _session; session != nullptr) [[unlikely]] {
+            do_suspend_task(*session, t);
+        }
+    }
+
+    /// Called from make_ready()/clear() when a suspended task is about
+    /// to be scheduled for execution. Fast path: single field check
+    /// on the task itself — no TLS access when profiler is off.
+    [[gnu::always_inline]]
+    static void on_task_resume(task& t) noexcept {
+        if (t._prof_idx != UINT32_MAX) [[unlikely]] {
+            do_resume_task(t);
+        }
+    }
+
+    /// Called when a task's awaited future is already available (e.g.
+    /// preemption path in co_await) — the task was never tracked, but
+    /// its parent should record the completion location so it doesn't
+    /// appear as a bare leaf. Fast path: single TLS pointer check.
+    [[gnu::always_inline]]
+    static void mark_completion(task& t) noexcept {
+        if (auto* const session = _session; session != nullptr) [[unlikely]] {
+            do_mark_completion(*session, t);
+        }
+    }
+
+private:
+    static void do_suspend_task(session& session, task& t) noexcept;
+    static void do_resume_task(task& t) noexcept;
+    static void do_mark_completion(session& session, task& t) noexcept;
+    static void sample_once(session& session) noexcept;
+
+    // Raw pointer (POD) rather than unique_ptr so that the TLS access on the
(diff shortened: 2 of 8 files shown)


gusev-p:
[Measured](https://github.com/gusev-p/scylla/commits/async_profiler/) `sample_once()` cost on `test_task_profiler` (3 nodes × 2 shards, 10 ms sampling interval):

| Node | Shard | Samples | Total entries | Total time | Avg entries/sample | Avg time/sample |
|------|-------|--------:|--------------:|-----------:|-------------------:|----------------:|
| 1    | 0     | 1499    | 66,446        | 5,811 µs   | 44.3               | **3,877 ns**    |
| 1    | 1     | 1498    | 50,187        | 4,247 µs   | 33.5               | **2,835 ns**    |
| 2    | 0     | 1506    | 120,807       | 9,372 µs   | 80.2               | **6,224 ns**    |
| 2    | 1     | 1506    | 120,905       | 8,583 µs   | 80.3               | **5,700 ns**    |
| 3    | 0     | 1506    | 134,713       | 9,818 µs   | 89.5               | **6,520 ns**    |
| 3    | 1     | 1506    | 140,067       | 10,158 µs  | 93.0               | **6,746 ns**    |

~3–7 µs per sample at the 10 ms interval = 0.03–0.07% reactor overhead. ~70 ns per processed entry on average across both phases.

gusev-p:
## perf-simple-query: profiler-off overhead

**Setup**: `./build/release/scylla perf-simple-query --smp 1 -m 1G`, release build, LTO disabled, 5 runs each, run 1 discarded as warmup, medians of runs 2-5 reported.

**Builds compared**:
- Baseline: scylla `183c6d120e` + seastar `4d268e0ef` (parents of task_profiler commits)
- With profiler: scylla `4eec97a178` + seastar `66bbbabb5` (HEAD of `async_profiler`)

| Metric | Baseline | Profiler | Δ% |
|---|---:|---:|---:|
| median tps | 247,396 | 250,801 | **+1.38%** |
| max tps | 248,388 | 252,330 | +1.59% |
| min tps | 245,292 | 248,730 | +1.40% |
| instructions_per_op | 37,549.5 | 37,799.3 | **+0.67%** |
| cpu_cycles_per_op | 18,864.4 | 18,584.0 | −1.49% |
| tasks_per_op | 14.1223 | 14.1243 | +0.014% |
| allocs_per_op | 58.0644 | 58.0652 | +0.001% |

## Analysis

- **Throughput went up +1.4%** with the profiler-instrumented binary; cpu_cycles_per_op went down −1.5%. This is counter-intuitive given the added hooks; almost certainly noise / different code-layout effects on a single, small benchmark on one machine. The 4 runs are tightly clustered (mad tps ≈ 0.2% of median in both), but inter-build variance can dwarf this on micro-benchmarks of this size.

- **The one signal that does look real: instructions_per_op +0.67%** (~250 extra instructions per operation). Consistent across all 4 runs (baseline 37537–37555 vs profiler 37781–37813, no overlap). This is the cost of the always-inlined check-and-skip code added at every \`set_task\` / \`make_ready\` / \`clear\` / \`mark_completion\` site (the \`if (_session) ...\` / \`if (_prof_idx != UINT_MAX) ...\` early-out branches plus the field write).

- **No change in tasks_per_op or allocs_per_op** — confirms the profiler doesn't change application behavior.

- **No throughput regression despite +0.67% instructions**, plausibly because:
  - The hot-path checks (\`_session == nullptr\`, \`_prof_idx == UINT_MAX\`) are perfectly predictable when profiler is off
  - Profiler-off path adds only branch+memory-read on already-hot cache lines (\`task\` object is already being touched)
  - cpu_cycles_per_op actually decreased, suggesting CPU frontend is not the bottleneck for this query mix; extra always-taken-not-branches don't hurt

**Conclusion**: profiler-off overhead on this micro-benchmark is **+0.67% instructions, no measurable throughput regression** (in fact +1.4%, but that's noise). Acceptable.

Caveats: single machine, single run-of-runs, micro-benchmark, no statistical significance test.

include/seastar/core/task.hh
```diff
@@ -24,21 +24,29 @@
...
+
 class task {
     std::source_location _resume_point = {};
 
 protected:
     scheduling_group _sg;
+    // Index into task_profiler's entries vector.  UINT_MAX means the
+    // task is not tracked.  Placed right after _sg (both 4 bytes) to
+    // fill the padding gap — zero per-task memory overhead.
+    unsigned _prof_idx = UINT_MAX;
```
radoslawcybulski: Please use `std::uint32_t` or something similar with fixed, hardware / compiler independent size.

include/seastar/core/task_profiler.hh
```diff
@@ -0,0 +1,98 @@
...
+    /// Stop profiling session, dump accumulated samples in folded-stack format to
+    /// the given stream, and reset internal state.
+    static future<> stop(output_stream<char>& out);
+
+    /// Called from set_task() when a task becomes dormant (suspended on
+    /// a non-ready future).  Fast path: single TLS pointer check,
+    /// predicted not-taken.
+    [[gnu::always_inline]]
+    static void on_task_suspend(task& t) noexcept {
+        if (auto* session = _session; __builtin_expect(session != nullptr, false)) {
```
radoslawcybulski: Why not `[[unlikely]]` (official c++20 standard)?

src/core/task_profiler.cc
```diff
@@ -0,0 +1,295 @@
...
+// indexed by task::_prof_idx.  Freed slots are removed by swapping
+// with the last entry and popping, keeping the vector dense — sample_once()
+// iterates only live entries with no holes to skip.
+//
+// Two concerns are separated:
+//
+// 1. Task entries (vector<task_entry>): tracks currently-suspended tasks.
+//    Managed by do_suspend_task (appends entry, sets _prof_idx)
+//    and do_resume_task (saves child_last_completion_loc in parent, swaps
+//    out entry).  Ephemeral — entries come and go as tasks suspend/resume.
```
radoslawcybulski: Nit: two whitespaces.

src/core/task_profiler.cc
```diff
@@ -0,0 +1,295 @@
...
+
+        if (key.depth > 0) {
+            session.folded_stacks[key]++;
+        }
+    }
+}
+
+void task_profiler::do_suspend_task(session& session, task& t) noexcept {
+    // Append a new entry at the end of the dense vector.
+    unsigned idx = session.suspended_tasks.size();
```
radoslawcybulski: We should get here a warning, right? `size_t` (64 bit) implicitly converted to `unsigned int` (hopefully 32 bit)?
Add a cast, please.

src/core/task_profiler.cc
```diff
@@ -0,0 +1,295 @@
...
+    if (!session) {
+        return;
+    }
+
+    unsigned idx = t._prof_idx;
+    t._prof_idx = UINT_MAX;
+
+    // When this child task resumes, its entry is about to be removed
+    // from the vector.  Record the child's resume point in the parent's
+    // entry so that if the parent ends up as a leaf at sampling time
```
radoslawcybulski, line 269: Just a question - do i understand correctly, that if a task had child and grandchild and both completed between the same ticks, only child will be emited as `completed` and grandchild will be skipped?

src/core/task_profiler.cc
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    // Take ownership so the session is destroyed even if we throw.
+    auto owner = std::unique_ptr<task_profiler::session>(session);
+
+    session->sampling_timer.cancel();
+
+    // Detach instance so on_task_suspend becomes a no-op and
+    // do_resume_task handles stale indexes gracefully.
+    _session = nullptr;
```
radoslawcybulski: Please move it just after `std::unique_ptr` ownership change, otherwise if `sampling_timer.cancel()` throws and someone calls `stop` second time we will get double free.

src/core/task_profiler.cc
```diff
@@ -0,0 +1,295 @@
...
+        sample_once(*s);
+    });
+    using namespace std::chrono;
+    auto period = duration_cast<timer<>::duration>(milliseconds(sampling_interval_ms));
+    session->sampling_timer.arm(timer<>::clock::now() + period, {period});
+    _session = session.release();
+}
+
+future<> task_profiler::stop(output_stream<char>& out) {
+    auto* session = _session;
```
radoslawcybulski: Do we need `session` here at all? You could compare against `_session` (thread local after all, so compiler should be able to figure out it will not change) and reuse `session` name for `unique_ptr` object, dropping `owner`. What do you think?

src/core/task_profiler.cc
```diff
@@ -0,0 +1,295 @@
...
+    auto& suspended = session->suspended_tasks;
+    unsigned last = suspended.size() - 1;
+    if (idx != last) {
+        suspended[idx] = suspended[last];
+        suspended[idx].t->_prof_idx = idx;
+    }
+    suspended.pop_back();
+}
+
+void task_profiler::start(unsigned sampling_interval_ms) {
```
radoslawcybulski: Why not `std::chrono::milliseconds` instead of `unsigned int`?

radoslawcybulski (commented):
Amazing job!

src/core/task_profiler.cc
gusev-p: When do_resume_task is called for a grandchild, child is still blocked on this grandchild, to unblock it we need to actually run_and_dispose() on the grandchild, which happens after do_resume_task() returns. If both do_resume_task and run_and_dispose for the grandchild are run between the ticks, on the second tick we'll observe child as `[completed]` and no grandchild. If only do_resume_task() happens between the ticks -- then we'll see grandchild as `[completed]` for the child.

src/core/task_profiler.cc
gusev-p: fixed

include/seastar/core/task_profiler.hh
gusev-p: updated

include/seastar/core/task.hh
gusev-p: updated

src/core/task_profiler.cc
gusev-p: done

src/core/task_profiler.cc
gusev-p: updated

src/core/task_profiler.cc
gusev-p: done

src/core/task_profiler.cc
gusev-p: updated

gusev-p:

radoslawcybulski (approved):
Looks good to me.

gusev-p:
new version:
  * write folded stacks in two versions -- with normalized frame counts (how many sample ticks have observed the given frame) and with all of them
  * use `std::pmr::unordered_map` for the folded stacks hash map

PR #3368, opened 2026-04-26
github: update compiler toolchains

avikivity:
Switch to Fedora 44 (in a container, as github doesn't support it
directly yet), Clang 22, and gcc 16. This lets us test with modern
toolchains and sets the stage for bumping the standard to C++26.

Since we're running in a container, drop sudo, which isn't provided
by the container and also isn't needed.

Disable the g++ debug mode combination due to
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=124584.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -43,37 +43,43 @@ jobs:
   test:
     timeout-minutes: 60
     runs-on: ubuntu-24.04
+    container:
+      image: docker.io/fedora:44
     steps:
+      - run: |
+          dnf -y upgrade
+
+      - name: Install git
+        run: |
+          dnf install -y git
+
       - uses: actions/checkout@v4
         with:
           submodules: "${{ contains(inputs.enables, 'dpdk') }}"
 
-      - run: |
-          sudo apt-get update
-
       - name: Install build dependencies
         run: |
-          sudo ./install-dependencies.sh
+          ./install-dependencies.sh
 
       - name: Install clang++
         if: ${{ inputs.compiler == 'clang++' }}
         run: |
-          sudo apt-get -y install clang-20
+          dnf install -y clang
 
       - name: Install clang-scan-deps
         if: ${{ contains(inputs.enables, 'cxx-modules') }}
         run: |
-          sudo apt-get -y install clang-tools-20
+          dnf install -y clang-tools-extra
 
       - name: Install g++
         if: ${{ inputs.compiler == 'g++' }}
         run: |
-          sudo apt-get -y install gcc-14 g++-14
+          dnf install -y gcc gcc-c++
 
       - name: Install ccache
         if: ${{ inputs.enable-ccache }}
         run: |
-          sudo apt-get -y install ccache
+          dnf install -y ccache
 
       - name: Setup ccache
         if: ${{ inputs.enable-ccache }}
@@ -84,11 +90,11 @@ jobs:
       - name: Configure
         run: |
           if [ ${{ inputs.compiler }} = "clang++" ]; then
-            CC=clang-20
-            CPP=clang++-20
+            CC=clang
+            CPP=clang++
           else
-            CC=gcc-14
-            CPP=g++-14
+            CC=gcc
+            CPP=g++
           fi
           if ${{ inputs.enable-ccache }}; then
             MAYBE_CCACHE_OPT="--compiler-cache=ccache"
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -22,6 +22,10 @@ jobs:
         compiler: [clang++, g++]
         standard: [20, 23]
         mode: [dev, debug, release]
+        exclude:
+          # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=124584
+          - compiler: g++
+            mode: debug
     with:
       compiler: ${{ matrix.compiler }}
       standard: ${{ matrix.standard }}


nyh (approved):
Does this mean our tests will run only on clang-22? Shouldn't we also build on clang-21, which Seastar users are far more likely to use and we would care more if it generates bad code?

avikivity:
> Does this mean our tests will run only on clang-22? 

Yes

> Shouldn't we also build on clang-21, which Seastar users are far more likely 

Far more likely?!

> to use and we would care more if it generates bad code?

The tests aren't to verify the compiler. They are to verify Seastar itself builds on the supported compiler/library combinations.

In principle we should build on the last two compiler versions, since that's our compatibility promise. Usually newer compilers are more strict (they catch violations that previous compilers did not) and older libraries are more strict (they implement fewer features). But the test environment allows testing only one, for now, and this changes from clang-20 (which we no longer support) to clang-22 (which we do).

avikivity:
v2: drop sudo; isn't supplied by the container, and now unneeded

avikivity:
v3:
 - install git
 - explicitly name libstdc++ version we target

avikivity:
v4:
 - make sure apt-get update runs before any package installation

nyh:
> > Shouldn't we also build on clang-21, which Seastar users are far more likely
> 
> Far more likely?!

I use the latest Fedora officially released (Fedora 43)  and it has clang 21.1.8, not clang 22.
ScyllaDB's official build system ("dbuild") uses clang 21.1.8 (not surprising, it also uses Fedora 43).

I didn't do a survey on how many developers using what, so I can't prove my hunch, but my hunch is that most of the Seastar users have not yet switched to clang 22.

avikivity:
> > > Shouldn't we also build on clang-21, which Seastar users are far more likely
> > 
> > 
> > Far more likely?!
> 
> I use the latest Fedora officially released (Fedora 43) and it has clang 21.1.8, not clang 22. ScyllaDB's official build system ("dbuild") uses clang 21.1.8 (not surprising, it also uses Fedora 43).

For a few more days.

> 
> I didn't do a survey on how many developers using what, so I can't prove my hunch, but my hunch is that most of the Seastar users have not yet switched to clang 22.

Seastar should be leading, not trailing, its users, so we encounter and fix the problems early.

avikivity:
https://github.com/scylladb/seastar/pull/3376

avikivity:
v5: switch to Fedora 44 instead of Ubuntu 26.04. Ubuntu 26.04 had too-old libfmt which contained an unfixed bug wrt newer clang. We have more expertise in Fedora, and Fedora tracks upstream releases better.

avikivity:
https://github.com/scylladb/seastar/pull/3377

avikivity:
v6: rebased past 2891a4bba58cd868b0f2c32fe0fa56b2418d09dc

avikivity:
https://github.com/scylladb/seastar/pull/3378

avikivity:
v7: rebased past 6be1e64f54370988094311168312733f17741817

nyh (approved):
Looks good but one silly nitpick: The last pagraph of the commit message doesn't look correct. Maybe it was relevant for some earlier version of this patch?

nyh:
> Looks good but one silly nitpick: The last pagraph of the commit message doesn't look correct. Maybe it was relevant for some earlier version of this patch?

Also, there seem to be some tests failing with the new compiler/library combination, please take a look.

avikivity:
https://github.com/scylladb/seastar/pull/3385

avikivity:
> Looks good but one silly nitpick: The last pagraph of the commit message doesn't look correct. Maybe it was relevant for some earlier version of this patch?

Yes, forgot to trim it.

avikivity:
v8:
 - rebased past 87cdb03cb60ddea1313049ce7d110fbc6f37f5fa
 - trimmed no-longer-relevant part of changelog

avikivity:
https://github.com/scylladb/seastar/pull/3387

avikivity:
v9:
 - rebased
 - disable leak sanitizer for g++ until https://gcc.gnu.org/bugzilla/show_bug.cgi?id=124584 is fixed

avikivity:
v10:
 - completely disable g++/debug until the gcc bug is fixed (causes problems in rpc_test)

avikivity:

travisdowns:
> Switch to Fedora 44 (in a container, as github doesn't support it
> directly yet), Clang 22, and gcc 16.

Is there a good to reason to switch away from Ubuntu? My impression is this is the most commonly used distro and probably the most common distro for seastar use and has long-lived LTS versions.

avikivity:
> > Switch to Fedora 44 (in a container, as github doesn't support it
> > directly yet), Clang 22, and gcc 16.
> 
> Is there a good to reason to switch away from Ubuntu? My impression is this is the most commonly used distro and probably the most common distro for seastar use and has long-lived LTS versions.

Ubuntu 26.04 has clang 21 and gcc 15, and I want 22/16.

If we want recent compiler versions (and I do), LTS doesn't help.

travisdowns:
> Ubuntu 26.04 has clang 21 and gcc 15, and I want 22/16.
> 
> If we want recent compiler versions (and I do), LTS doesn't help.

Dunno about gcc side but you can trivially install any clang version with llvm.sh.

avikivity:
> > Ubuntu 26.04 has clang 21 and gcc 15, and I want 22/16.
> > If we want recent compiler versions (and I do), LTS doesn't help.
> 
> Dunno about gcc side but you can trivially install any clang version with llvm.sh.

What is llvm.sh?

But we do want gcc.

travisdowns:
> What is llvm.sh?

The installer script maintained by the llvm project, it installs any version of llvm/clang/etc you'd like "side by side".

avikivity:
> > What is llvm.sh?
> 
> The installer script maintained by the llvm project, it installs any version of llvm/clang/etc you'd like "side by side".

Nice, we can use it later to test multiple clang versions.

gcc is trickier though.

travisdowns:
> gcc is trickier though.

ubuntu 26.04 offers gcc-16 as an installable option, so that can be used via contianer if Ubuntu is preferred (26.04 not offered as "runner" yet)

on the runner itself gcc seems pretty easy with the toolchain ppa which is how I've done this many times

That said is there any interest in staying on ubuntu, and/or directly using the runner instead of container mode? If so I could do that, but trying to read the room here.

avikivity:
> > gcc is trickier though.
> 
> ubuntu 26.04 offers gcc-16 as an installable option, so that can be used via contianer if Ubuntu is preferred (26.04 not offered as "runner" yet)
> 
> on the runner itself gcc seems pretty easy with the toolchain ppa which is how I've done this many times
> 
> That said is there any interest in staying on ubuntu, and/or directly using the runner instead of container mode? If so I could do that, but trying to read the room here.

I don't have any love for Ubuntu, but also no motivation for the change other than it was the path of least resistance for me.

Ubuntu 26.04 isn't yet available as a runner, so we have to use a container.

The goal is to update on a half-year cadence like clang. Fedora offers this and also has mid-life updates. I don't know enough about Ubuntu.

If it's possible to do this on Ubuntu, and it is also advantageous (for example, supporting the last two compilers instead of just one), then by all means, post a patch.

travisdowns:
> post a patch.

I will

travisdowns:
https://github.com/scylladb/seastar/pull/3412

avikivity:
Superseded by  https://github.com/scylladb/seastar/commit/4ccea322aca5b79668f6bd53082a0ef575616682.

PR #3369, opened 2026-04-28
net/tls: improve OpenSSL error queue hygiene

pgellert:
OpenSSL uses a per-thread error queue to communicate failure details. In
seastar's cooperative scheduling model multiple TLS sessions share the
same thread, so if stale errors are left on the queue across a
scheduling point, `SSL_get_error` on an unrelated session can
misclassify the result (e.g. reporting `SSL_ERROR_SYSCALL` instead of
`SSL_ERROR_SSL`) because it peeks at the oldest error on the queue to
decide the classification.

We have hit this in practice at redpanda and it resulted in
hard-to-debug customer issues. This PR hardens the OpenSSL error queue
handling. See the individual commit messages for the details.

**Drain stale errors on error paths**
- `net/tls: use make_openssl_error for BIO creation failure`. Aligns
  the `BIO_new` failure with every other error path in the session
  constructor.
- `net/tls: drain error queue on unexpected error codes`. Ensures the
  default switch cases in error handlers drain the queue rather than
  silently discarding errors.

**Drain stale errors after successful operations**
- `net/tls: clear error queue after successful SSL_CTX_new`.
  `SSL_CTX_new` can return success while leaving errors from
  partially-failed system config parsing (e.g. an invalid
  `Ciphersuites` value in `openssl.cnf`).
- `net/tls: clear error queue after successful SSL operations`.
  `SSL_shutdown`, `SSL_do_handshake`, `SSL_write_ex`, and `SSL_read_ex`
  can return success while leaving errors from internal BIO write
  failures that OpenSSL absorbed. Root cause is a return-value
  translation gap in `bio_write_intern` where our `bwrite` callback's
  0-means-failure is misclassified as success by
  `tls_retry_write_records`. See the commit message for detailed call
  chains.

**Assert the queue is clean**
- `net/tls: assert clean error queue before SSL operations`. Adds
  `verify_clean_error_queue` checks before every `SSL_do_handshake`,
  `SSL_write_ex`, `SSL_read_ex`, and `SSL_shutdown` call as a safety
  net to surface regressions in tests and to make issues easier to debug in production.

### DIFF
--- a/src/net/tls_openssl.cc
+++ b/src/net/tls_openssl.cc
@@ -961,6 +961,7 @@ public:
         // This do_until runs until either a renegotiation occurs or the packet is empty
         while (!eof() && size > 0) {
             size_t bytes_written = 0;
+            verify_clean_error_queue("SSL_write_ex");
             auto write_rc = SSL_write_ex(_ssl.get(), ptr, size, &bytes_written);
             tls_log.trace("{} do_put: SSL_write_ex: {}", *this, write_rc);
             if (write_rc != 1) {
@@ -1053,6 +1054,7 @@ public:
             [this] { return connected() || eof(); },
             [this] {
                 try {
+                    verify_clean_error_queue("SSL_do_handshake");
                     auto n = SSL_do_handshake(_ssl.get());
                     tls_log.trace("{} do_handshake: SSL_do_handshake: {}", *this, n);
                     if (n <= 0) {
@@ -1175,6 +1177,7 @@ public:
             tls_log.trace("{} do_get: available: {}", *this, avail);
             buf_type buf(avail);
             size_t bytes_read = 0;
+            verify_clean_error_queue("SSL_read_ex");
             auto read_result = SSL_read_ex(
               _ssl.get(), buf.get_write(), avail, &bytes_read);
             tls_log.trace("{} do_get: SSL_read_ex: {}", *this, read_result);
@@ -1271,6 +1274,7 @@ public:
             return make_ready_future();
         }
 
+        verify_clean_error_queue("SSL_shutdown");
         auto res = SSL_shutdown(_ssl.get());
         tls_log.trace("{} do_shutdown: SSL_shutdown: {}", *this, res);
         if (res == 1) {
@@ -1632,6 +1636,21 @@ private:
         }
         auto errors = get_all_openssl_errors();
         tls_log.debug("{} {}: ignoring stale errors on queue: {}", *this, operation, errors);
+    }
+
+    // Checks that the OpenSSL per-thread error queue is clean before
+    // calling an SSL function.  A dirty queue can cause SSL_get_error
+    // to misclassify results (e.g. reporting SSL_ERROR_SYSCALL instead
+    // of SSL_ERROR_SSL), which can affect unrelated sessions that share
+    // the same thread.
+    void verify_clean_error_queue(const char* operation) {
+        auto err = ERR_peek_error();
+        if (err == 0) [[likely]] {
+            return;
+        }
+        char buf[256];
+        ERR_error_string_n(err, buf, sizeof(buf));
+        tls_log.warn("{} stale error on queue before {}: {}", *this, operation, buf);
     }
 
     std::vector<subject_alt_name> do_get_alt_name_information(const x509_ptr &peer_cert,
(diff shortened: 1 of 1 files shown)


src/net/tls_openssl.cc
```diff
@@ -1614,6 +1627,33 @@ class openssl_session : public enable_shared_from_this<openssl_session>, public
...
+    // the same thread.
+    void verify_clean_error_queue(const char* operation) {
+        auto err = ERR_peek_error();
+        if (err == 0) [[likely]] {
+            return;
+        }
+        char buf[256];
+        ERR_error_string_n(err, buf, sizeof(buf));
+        tls_log.warn("{} stale error on queue before {}: {}", *this, operation, buf);
+        assert(0 && "stale errors on OpenSSL error queue");
```
dotnwat: in production there is no assert so we just get the warning. but does this mean the queue is now dirty/inconsistent? should it be cleared? or is this a hard invariant and should actually be a hard assert in production?

pgellert: > in production there is no assert so we just get the warning. but does this mean the queue is now dirty/inconsistent?

Yes. It is a sign that there is a bug somewhere in OpenSSL or our `tls_openssl.cc` implementation, which leads to an error propagating across connections/fibres.

> should it be cleared?

Possibly. It felt too heavy-handed for me to clear errors here when we don't know where they are coming from, which is why I opted to maintain the existing behaviour.

If you think it would be more appropriate to clear the queue here, I'm happy with that too, but it wasn't obvious to me if that's the right call, which is why I opted to leave the current behaviour.

> or is this a hard invariant and ...

It's an undocumented expectation of the `SSL_* -then- SSL_get_error` call pairs that the error stack is clear before the first openssl call. Not maintaining this invariant leads to a misclassified error returned from `SSL_get_error` in all the cases that I've seen.

> ... should actually be a hard assert in production?

The reason why I don't think we should hard assert in production on this is that it significantly increases the blast radius of the issue. It can turn a benign bug (e.g. leaving an error after a successful operation on the error queue and then causing a misreported error on another connection) into a hard crash of the whole process. The bug might only affect a low-priority connection (e.g. telemetry data), but the crash can cause unavailability.

src/net/tls_openssl.cc
```diff
@@ -971,6 +971,7 @@ class openssl_session : public enable_shared_from_this<openssl_session>, public
                     co_return;
                 }
             } else {
+                clear_stale_ssl_errors("SSL_write_ex");
```
avikivity, line 975: The distance between SSL_write_ex, the error code check, and this point distributes the error handling across a large span of lines and risks that a refactoring will introduce new bugs.

Consider a followup to add some mechanism to wrap it up. For example

```c++
auto write_rc = with_clearing_stale_errors(SL_write_ex(_ssl.get(), ptr, size, &bytes_written));
...
```

src/net/tls_openssl.cc
```diff
@@ -1634,6 +1638,22 @@ class openssl_session : public enable_shared_from_this<openssl_session>, public
...
+    // the same thread.
+    void verify_clean_error_queue(const char* operation) {
+        auto err = ERR_peek_error();
+        if (err == 0) [[likely]] {
+            return;
+        }
+        char buf[256];
+        ERR_error_string_n(err, buf, sizeof(buf));
+        tls_log.warn("{} stale error on queue before {}: {}", *this, operation, buf);
+        assert(0 && "stale errors on OpenSSL error queue");
```
avikivity: The assertion is too much. It converts a tolerable bug into a crash.

avikivity:
Looks very good, but the assertion is too much. Not all Seastar users have NDEBUG in production.

src/net/tls_openssl.cc
pgellert: What makes this trickier is that it's not just the ssl call itself (e.g. `SSL_write_ex`) that needs to be wrapped, but instead the pair call of `SSL_write_ex`-then-`SSL_get_error`, since `SSL_get_error` may read the state of the error queue, so we can only clear the errors after that.

It should be doable, and I'll try to find the time to do that as a follow-up. I've considered doing it even as part of this PR, but it would have grown too large.

pgellert:
[force-push](https://github.com/scylladb/seastar/compare/eb74904815e07d200cfeba94da30399221c38f0b..4bc660ccb559c1636cd8239b4cdc60f6f69d6c42): address review feedback, specifically to remove the `assert` from `verify_clean_error_queue`

src/net/tls_openssl.cc
avikivity: Yes, that's why I didn't ask for it here.

PR #3371, opened 2026-04-28
net: Introduce round-robin connection balancing policy

tgrabiec:
Useful for some protocols which assume this policy, like shard-aware CQL protocol, which keeps opening connections until it hits the desired shard. With the current default load balancing strategy, which picks the least-loaded shard, that is not guaranteed to converge and can live-lock if there is connection imbalance.

### DIFF
--- a/include/seastar/net/api.hh
+++ b/include/seastar/net/api.hh
@@ -360,6 +360,9 @@ public:
         port,
         // This algorithm distributes all new connections to listen_options::fixed_cpu shard only.
         fixed,
+        // This algorithm distributes new connections to shards in a round-robin fashion,
+        // cycling through shards 0, 1, ..., N-1, 0, 1, ... regardless of current load.
+        round_robin,
         default_ = connection_distribution
     };
     /// Constructs a \c server_socket without being bound to any address
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -60,6 +60,7 @@ using namespace seastar;
 class conntrack {
     class load_balancer {
         std::vector<unsigned> _cpu_load;
+        unsigned _rr_counter = 0;
     public:
         load_balancer() : _cpu_load(size_t(smp::count), 0) {}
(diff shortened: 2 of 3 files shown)


tgrabiec:
Needed by ScyllaDB: https://github.com/scylladb/scylladb/pull/29687

include/seastar/net/api.hh
```diff
@@ -360,6 +360,9 @@ public:
         port,
         // This algorithm distributes all new connections to listen_options::fixed_cpu shard only.
         fixed,
+        // This algorithm distributes new connections to shards in a round-robin fashion,
```
nyh, line 363: nitpick: typo in the commit message title ("balaning").

include/seastar/net/posix-stack.hh
```diff
@@ -60,6 +60,7 @@ using namespace seastar;
 class conntrack {
     class load_balancer {
         std::vector<unsigned> _cpu_load;
+        unsigned _rr_counter = 0;
```
nyh, line 63: nitpick: If you're familiar with this code, you can guess that the acronym "rr" refers to round-robin, but it's not obvious. I would add a comment that this variable is used by load_balancing_algorithm::round_robin to choose the next shard.

include/seastar/net/posix-stack.hh
```diff
@@ -74,6 +75,12 @@ class conntrack {
             _cpu_load[cpu]++;
             return cpu;
         }
+        shard_id next_cpu_rr() {
+            auto cpu = shard_id(_rr_counter % smp::count);
```
nyh, line 79: Note that in https://github.com/scylladb/seastar/pull/3355 @avikivity  deprecates smp::count, so this call will need to be fixed (in the same way Avi fixed the lines above).

nyh (approved):
Looks good to me, had just some trivial nitpicks (and also there's some interaction with https://github.com/scylladb/seastar/pull/3355 that somebody will need to fix, sooner or later).

PR #3372, opened 2026-04-29
httpd: log do_accepts gate-closed instead of swallowing it

ScyllaPiotr:
Follow-up to #3359 review.
The handle_exception_type(gate_closed_exception) in do_accepts() is
unreachable under correct use (gate is only closed by stop()). Log
the failure instead of swallowing it silently.

### DIFF
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -462,7 +462,9 @@ future<> http_server::accept_loop(int which, bool tls) {
 future<> http_server::do_accepts(int which, bool tls) {
     (void)try_with_gate(_task_gate, [this, which, tls] {
         return accept_loop(which, tls);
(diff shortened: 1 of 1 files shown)


nyh (approved):

nyh:
I approved, but I hope you are 100% convinced that this can never ever happen during a normal shutdown. Our dtests have a frustrating feature where a dtest fails if ever it sees a "error"-level log message that it didn't expect. If this log message ever appears during normal shutdowns, tests will start failing.

xemul:
Why not make it warn-level, the error is non-fatal anyway

xemul:
Changes made by the 2nd version of the patch now exceed the description of it

ScyllaPiotr:
Patch scope now limited.

ScyllaPiotr:
@xemul please review

PR #3375, opened 2026-04-30
Zero-copy http chunked data sink

### LINKED ISSUE #2916: HTTP client chunked body data sinks copy data
The sink in question has no other choice, because potentially zero-copied body chunks will be interleaved with tiny chunk delimiters (size octets and `\n\r` trailers). This calls for fully mixed mode in the underlying stream

detached from #2783

xemul:
Make http_chunked_data_sink_impl (API ≥ 9) splice user buffers into the underlying output_stream's zero-copy queue instead of memcpy'ing them through its small buffer. Chunk-size header and trailing CRLF still go through the buffered write, piggy-backing on output_stream's mixed-mode facility.

First it rewrites httpd/content_length_limit and prometheus tests on top of http::client to fix a latent fragility (single input.read() per response) that the new sink would otherwise expose. The second commit is the sink change itself, plus a unit test.

Fixes #2916

### DIFF
--- a/tests/unit/httpd_test.cc
+++ b/tests/unit/httpd_test.cc
@@ -762,32 +762,32 @@ SEASTAR_TEST_CASE(content_length_limit) {
         loopback_connection_factory lcf(1);
         http_server server("test");
         server.set_content_length_limit(11);
-        loopback_socket_impl lsi(lcf);
         httpd::http_server_tester::listeners(server).emplace_back(lcf.get_server_socket());
 
-        future<> client = seastar::async([&lsi] {
-            connected_socket c_socket = lsi.connect(socket_address(ipv4_addr()), socket_address(ipv4_addr())).get();
-            input_stream<char> input(c_socket.input());
-            output_stream<char> output(c_socket.output());
+        future<> client = seastar::async([&lcf] {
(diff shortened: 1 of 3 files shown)


tests/unit/httpd_test.cc
```diff
@@ -762,32 +762,32 @@ SEASTAR_TEST_CASE(content_length_limit) {
...
-            output.flush().get();
-            auto resp = input.read().get();
-            BOOST_REQUIRE_NE(std::string(resp.get(), resp.size()).find("200 OK"), std::string::npos);
-
-            output.write(sstring("GET /test HTTP/1.1\r\nHost: test\r\nContent-Length: 11\r\n\r\nxxxxxxxxxxx")).get();
-            output.flush().get();
-            resp = input.read().get();
-            BOOST_REQUIRE_NE(std::string(resp.get(), resp.size()).find("200 OK"), std::string::npos);
+        future<> client = seastar::async([&lcf] {
+            auto cln = http::client(std::make_unique<loopback_http_factory>(lcf));
```
avikivity, line 768: There's some circularity here, if http and httpd share the same bug, we wouldn't detect it.

I suppose the odds are low.

PR #3376, opened 2026-05-02
build: fix cmake missing ninja on Ubuntu 26.04

avikivity:
Apparently, the ninja-build package on Ubuntu 26.04 provides a ninja command but not ninja-build. On the other hand cmake apparently looks for ninja-build. Detect this situation and steer cmake in the right direction when it happens.


bhalevy (approved):

avikivity:

nyh (approved):

PR #3378, opened 2026-05-04
vla.hh: add missing includes

avikivity:
vla.hh uses malloc() and uintptr_t, but doesn't include their respective headers. Fix this.


avikivity:

nyh (approved):

PR #3381, opened 2026-05-05
coroutine/generator: fix setup of generator's waiting task

### LINKED ISSUE #3380: Infinite loop when walking the continuation chain of a seastar::coroutine::generator
In seastar, every schedulable unit of work is a `seastar::task` with a `virtual waiting_task()` method. This forms a linked chain used by diagnostics (e.g., `do_dump_task_queue`) to answer "who is ultimately blocked on this task?" The chain goes bottom-up: from a leaf task, through each intermediate coroutine, up to the root initiator, terminating at nullptr.

For regular coroutines that return `seastar::future<T>`, this chain is maintained automatically. Each such coroutine's `promise_type` contains a `seastar::promise<T>` member. When another coroutine does `co_await` on that future, the standard awaiter calls `future::set_coroutine(awaiting_promise)`, which stores the awaiting task in `seastar::promise::_task`. The coroutine's `waiting_task()` simply delegates to `_promise.waiting_task()`, returning whoever is waiting for its result. No manual bookkeeping required — the chain assembles itself through the normal `co_await` machinery.

A generator coroutine (`seastar::coroutine::experimental::generator`) does not return a `seastar::future`. Its `promise_type` inherits directly from `seastar::task` — it IS a task — but it contains no `seastar::promise`. There is no future/promise pair mediating the generator-consumer relationship.

Despite potentially being resumed many times (once per value pulled by the consumer), the generator is a single `seastar::task` throughout its lifetime. The expected continuation chain involving a generator looks like:
inner_task  →  generator  →  consumer  →  ...  →  nullptr

Where:
- inner_task is whatever the generator is currently co_awaiting internally (e.g., a file I/O operation). It points to the generator via the standard set_coroutine mechanism and it may change while the generator proceeds
- generator should point to the consumer (the outside coroutine that called `co_await gen()`). Since there's no `seastar::promise::_task` to track this automatically, the generator has an explicit `_waiting_task` field for this purpose.
- consumer points to its own caller through the normal future-based mechanism.

The generator's consumer calls `co_await gen()`, which goes through a `call_awaiter`. This awaiter resumes the generator via direct coroutine handle manipulation (symmetric transfer) — no `set_coroutine` is ever called, so nothing automatically sets `_waiting_task` to the consumer. `call_awaiter::await_suspend` stores the consumer handle but does not set `_waiting_task`.

When the generator produces a value via `co_yield`, `yield_awaiter::await_suspend` or `copy_awaiter::await_suspend` is called with the generator's own handle as the `producer` parameter (since co_yield suspends the generator). This method executes:
`_promise->_waiting_task = &producer.promise();`

But we also always create `yield_awaiter`s and `copy_awaiter`s setting `_promise` to `this` in the generator's task (`generator_promise_base`).
Since `_promise` is the generator's promise and `producer.promise()` also is the generator's promise (the generator is the coroutine being suspended), this sets `_waiting_task` to itself.

Nothing ever corrects this. When the generator subsequently does `co_await` on an internal future, the standard awaiter only calls `set_coroutine` on that future's promise — it doesn't touch the generator's own `_waiting_task`.

The actual chain becomes:
inner_task  →  generator  →  generator  →  generator  →  ...  (infinite)

Under normal operation this is harmless because `_waiting_task` is never traversed. But when the task queue exceeds `max_task_backlog`, `reactor::task_queue::do_dump` walks the `waiting_task()` chain for tasks in the queue. It follows the self-referential pointer indefinitely, growing a log buffer until memory allocation fails and the process aborts.

To fix this we should set the `_waiting_task` in the `call_awaiter` to the caller coroutine. We should also remove setting the `_promise->_waiting_task` to the same task.

wmitros:
When we dump tasks in `reactor::task_queue::do_dump`, we walk the continuation chains of each tasks using `waiting_task()`. For a generator coroutine task, this `_waiting_task` was incorrectly set to the generator itself, instead of its caller. Additionally, it was only set when a value was `co_yield`ed, so if it spawned some tasks before that, we couldn't have seen the full continuation chains of them.

In this patch we fix this by setting the `_waiting_task` of the generator to its caller in the call_awaiter constructor, and we remove setting of the `_waiting_task` from `yield_awaiter` and `copy_awaiter`. The `_promise` field in these classes is no longer used so it's also removed.

We also add a test that reproduces the infinite continuation chain walk in `tests/unit/generator_test.cc`. Note that we can't use `current_tasktrace()` there because this continuation chain walk limits the number of tasks to 16 (`tasktrace::vector_type::max_size()`), so instead we force the dump by spawning many tasks. The test never finishes without the fix and when the fix is applied, the "Too long queue accumulated for ..." warn message is correctly printed and test finishes quickly.

Fixes #3380

### DIFF
--- a/tests/unit/generator_test.cc
+++ b/tests/unit/generator_test.cc
@@ -21,6 +21,7 @@
 
 #include <seastar/core/circular_buffer_fixed_capacity.hh>
 #include <seastar/core/coroutine.hh>
+#include <seastar/core/loop.hh>
 #include <seastar/coroutine/as_future.hh>
 #include <seastar/coroutine/generator.hh>
 #include <seastar/coroutine/exception.hh>
@@ -545,4 +546,23 @@ SEASTAR_TEST_CASE(test_batch_generator_partial_last_batch) {
     }
     auto exhausted = co_await gen();
     BOOST_REQUIRE(!exhausted.has_value());
+}
+
+// Reproduces issue #3380
+SEASTAR_TEST_CASE(test_generator_waiting_task_no_infinite_loop) {
+    auto gen = [&]() -> coroutine::experimental::generator<int> {
+        co_yield 0;
+        co_await parallel_for_each(std::views::iota(0, 10000), [](int) {
+            return yield();
+        });
+        co_yield 1;
+    }();
+    // First yield would set the generator's waiting task to itself
+    auto val = co_await gen();
+    BOOST_REQUIRE_EQUAL(*val, 0);
+    // To reach the second yield, the generator must finish a large number of concurrent tasks, which should
+    // cause a dump of tasks in the task queue. If the generator's waiting task was set to itself,
+    // the dump would never finish.
+    auto val2 = co_await gen();
+    BOOST_REQUIRE_EQUAL(*val2, 1);
 }
--- a/include/seastar/coroutine/generator.hh
+++ b/include/seastar/coroutine/generator.hh
@@ -162,7 +162,7 @@ public:
 
     yield_awaiter final_suspend() noexcept {
         _value = nullptr;
-        return {this, this->_consumer};
+        return {this->_consumer};
     }
 
     void unhandled_exception() noexcept {
@@ -171,7 +171,7 @@ public:
 
     yield_awaiter yield_value(Yielded&& value) noexcept {
         this->_value = std::addressof(value);
-        return {this, this->_consumer};
+        return {this->_consumer};
     }
 
     copy_awaiter yield_value(const yielded_deref_type& value)
@@ -182,7 +182,7 @@ public:
                   std::constructible_from<
                     yielded_decvref_type,
                     const yielded_deref_type&>) {
-        return {this, this->_consumer, yielded_decvref_type(value), _value};
+        return {this->_consumer, yielded_decvref_type(value), _value};
     }
 
     void return_void() noexcept {}
@@ -213,7 +213,6 @@ private:
 
 template <typename Yielded>
 struct generator_promise_base<Yielded>::yield_awaiter final {
-    generator_promise_base* _promise;
     std::coroutine_handle<> _consumer;
 
     bool await_ready() const noexcept {
@@ -222,7 +221,6 @@ struct generator_promise_base<Yielded>::yield_awaiter final {
     template <typename Promise>
     std::coroutine_handle<> await_suspend(std::coroutine_handle<Promise> producer SEASTAR_COROUTINE_LOC_PARAM) noexcept {
         SEASTAR_COROUTINE_LOC_STORE(producer.promise());
-        _promise->_waiting_task = &producer.promise();
         return schedule_or_resume(_consumer);
     }
     void await_resume() noexcept {}
@@ -230,7 +228,6 @@ struct generator_promise_base<Yielded>::yield_awaiter final {
 
 template <typename Yielded>
 struct generator_promise_base<Yielded>::copy_awaiter final {
-    generator_promise_base* _promise;
     std::coroutine_handle<> _consumer;
     yielded_decvref_type _value;
     value_ptr_type& _value_ptr;
@@ -242,7 +239,6 @@ struct generator_promise_base<Yielded>::copy_awaiter final {
     std::coroutine_handle<> await_suspend(std::coroutine_handle<Promise> producer SEASTAR_COROUTINE_LOC_PARAM) noexcept {
         SEASTAR_COROUTINE_LOC_STORE(producer.promise());
         _value_ptr = std::addressof(_value);
-        _promise->_waiting_task = &producer.promise();
         return schedule_or_resume(_consumer);
     }
     constexpr void await_resume() const noexcept {}
@@ -414,6 +410,7 @@ public:
         std::coroutine_handle<> await_suspend(std::coroutine_handle<Promise> consumer SEASTAR_COROUTINE_LOC_PARAM) noexcept {
             SEASTAR_COROUTINE_LOC_STORE(consumer.promise());
             _gen->_coro.promise()._consumer = consumer;
+            _gen->_coro.promise()._waiting_task = &consumer.promise();
             return schedule_or_resume(_gen->_coro);
         }
 
@@ -562,7 +559,7 @@ public:
 
     yield_awaiter final_suspend() noexcept {
         _finished = true;
-        return yield_awaiter{this, this->_consumer, true};
+        return yield_awaiter{this->_consumer, true};
     }
 
     void unhandled_exception() noexcept {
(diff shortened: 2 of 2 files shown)


tests/unit/generator_test.cc
```diff
@@ -546,3 +547,22 @@ SEASTAR_TEST_CASE(test_batch_generator_partial_last_batch) {
...
+            return yield();
+        });
+        co_yield 1;
+    }();
+    // First yield would set the generator's waiting task to itself
+    auto val = co_await gen();
+    BOOST_REQUIRE_EQUAL(*val, 0);
+    // To reach the second yield, the generator must finish a large number of concurrent tasks, which should
+    // cause a dump of tasks in the task queue. If the generator's waiting task was set to itself,
+    // the dump would never finish.
```
bhalevy, line 565: nit:  How long does the 10000 parallel_for_each loop take?
If it's too long and we want to make it shorter, see `test_rpc_stream_backpressure_across_shards` for a way to temporarily reduce `max_task_backlog` for the test.

tests/unit/generator_test.cc
```diff
@@ -21,6 +21,7 @@
 
 #include <seastar/core/circular_buffer_fixed_capacity.hh>
 #include <seastar/core/coroutine.hh>
+#include <seastar/core/when_all.hh>
```
bhalevy: nit: is it used?

tests/unit/generator_test.cc
```diff
@@ -546,3 +547,22 @@ SEASTAR_TEST_CASE(test_batch_generator_partial_last_batch) {
     auto exhausted = co_await gen();
     BOOST_REQUIRE(!exhausted.has_value());
 }
+
+// Reproduces issue #3380
+SEASTAR_TEST_CASE(test_generator_waiting_task_no_infinite_loop) {
+    auto gen = [&]() -> coroutine::experimental::generator<int> {
+        co_yield 0;
+        co_await parallel_for_each(std::views::iota(0u, 10000u), [](int) {
```
tchaikov: FWIW, i created another test locally to verify the fix:

```c++
struct capture_task_ptr {
    seastar::task** _out;
    bool await_ready() noexcept { return false; }
    template <typename P>
    bool await_suspend(std::coroutine_handle<P> h) noexcept {
        *_out = static_cast<seastar::task*>(&h.promise());
        return false; // resume immediately without truly suspending
    }
    void await_resume() noexcept {}
};

SEASTAR_TEST_CASE(test_generator_waiting_task_chain) {
    using gen_t = coroutine::experimental::generator<int>;
    seastar::task* gen_task = nullptr;

    auto gen = std::invoke([](seastar::task** out) -> gen_t {
        co_await capture_task_ptr{out};
        co_yield 1;
        co_await seastar::yield();
        co_yield 2;
    }, &gen_task);

    auto val = co_await gen();
    BOOST_REQUIRE(val.has_value());
    BOOST_REQUIRE_EQUAL(*val, 1);

    // gen_task was populated by capture_task_ptr inside the generator.
    // After co_yield 1, the generator is suspended and _waiting_task must
    // point to the consumer (this test coroutine), not to the generator itself.
    BOOST_REQUIRE(gen_task != nullptr);
    BOOST_REQUIRE(gen_task->waiting_task() != gen_task);

    val = co_await gen();
    BOOST_REQUIRE(val.has_value());
    BOOST_REQUIRE_EQUAL(*val, 2);

    val = co_await gen();
    BOOST_REQUIRE(!val.has_value());
}
```

tchaikov (approved):
lgtm

tests/unit/generator_test.cc
wmitros: I don't see how to run a single test case but the entire generator_test takes ~0.2s on my machine with and without this test case (it's sometimes faster with the test than without) so the runtime seems negligible

tests/unit/generator_test.cc
wmitros: It's used for `parallel_for_each` but as Copilot pointed out, it's more precise to include `loop.hh` because that's actually where `parallel_for_each` is declared - I changed in the rebase

tests/unit/generator_test.cc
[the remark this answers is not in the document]
wmitros: Fixed in rebase

tests/unit/generator_test.cc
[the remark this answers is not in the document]
wmitros: Fixed in rebase

tests/unit/generator_test.cc
```diff
@@ -546,3 +547,22 @@ SEASTAR_TEST_CASE(test_batch_generator_partial_last_batch) {
...
+        co_yield 1;
+    }();
+    // First yield would set the generator's waiting task to itself
+    auto val = co_await gen();
+    BOOST_REQUIRE_EQUAL(*val, 0);
+    // To reach the second yield, the generator must finish a large number of concurrent tasks, which should
+    // cause a dump of tasks in the task queue. If the generator's waiting task was set to itself,
+    // the dump would never finish.
+    auto val2 = co_await gen();
+    BOOST_REQUIRE_EQUAL(*val2, 1);
```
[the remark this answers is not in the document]
wmitros: When setting the max backlog too low the test is unstable and setting it to something like `max_task_backlog/2` doesn't really change anything in practice. The test finishes quickly regardless so I'll keep it

tests/unit/generator_test.cc
wmitros: I went with a test that verifies that we prevent an issue that may be observable by the user but this test would also confirm that it won't happen by checking the internals. Do you think it's also worth adding to `generator_test.cc`? With the `gen_task->waiting_task() != gen_task` check we only prevent an infinite loop, but if we instead managed to verify that the generator points to the outside coroutine I think it could catch some issues that the test that I added wouldn't

wmitros:
Looks like @avikivity is busy/not getting notifications. @scylladb/seastar-maint please take a look

travisdowns:
Has this existed since the OG generators were introduced, or was it introduced only with the big generator refactor: https://github.com/scylladb/seastar/commit/81f2dc9dd ?

wmitros:
> Has this existed since the OG generators were introduced, or was it introduced only with the big generator refactor: [81f2dc9dd](https://github.com/scylladb/seastar/commit/81f2dc9dd) ?

Looks like it was correct before: `generator::operator()` (called by the outside coroutine) returned `wait_for_next_value()` which returned `_wait_for_next_value.emplace().get_future()` and `_wait_for_next_value->waiting_task()` was the task returned by `waiting_task()`. So the task associated with the generator returned the task associated with the future returned to the outside coroutine

tests/unit/generator_test.cc
avikivity: 10,000 is not a large number for a computer

tests/unit/generator_test.cc
tchaikov: @wmitros I was not sure. was trying to come up with a minimal test to verify the fix. but you are right, my test might be way too narrow in the sense of testing coverage. and testing the generator as a blackbox is probably the better way to go.

PR #3386, opened 2026-05-06
net/tls: introduce ssl_call wrapper for SSL I/O

pgellert:
Consolidate the verify-clean-queue / SSL operation / SSL_get_error / drain dance into a single ssl_call helper. The helper guarantees the OpenSSL error queue is empty on return and surfaces any drained codes as a vector in the result, so callers do not have to reason about queue state.

This is a refactor: the same exceptions are thrown on the same paths with the same error codes, and the success-path stale-error logging is preserved. Applied to SSL_write_ex, SSL_read_ex, SSL_do_handshake, and SSL_shutdown call sites.

Addresses @avikivity's review feedback from https://github.com/scylladb/seastar/pull/3369#discussion_r3180786727

### DIFF
--- a/src/net/tls_openssl.cc
+++ b/src/net/tls_openssl.cc
@@ -171,8 +171,12 @@ std::system_error make_openssl_error(const std::string & msg) {
     return make_openssl_error(msg, get_all_openssl_errors());
 }
 
+std::runtime_error make_unknown_openssl_error(const std::string & msg, std::vector<openssl_errc> error_codes) {
+    return std::runtime_error(fmt::format("{}: {}", msg, error_codes));
+}
+
 std::runtime_error make_unknown_openssl_error(const std::string & msg) {
-    return std::runtime_error(fmt::format("{}: {}", msg, get_all_openssl_errors()));
+    return make_unknown_openssl_error(msg, get_all_openssl_errors());
 }
 
 bool contains_openssl_error(const std::vector<openssl_errc> & error_codes, int lib, int reason) {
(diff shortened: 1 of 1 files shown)


src/net/tls_openssl.cc
```diff
@@ -1653,6 +1651,26 @@ class openssl_session : public enable_shared_from_this<openssl_session>, public
...
     }
 
+    struct ssl_call_result {
+        int rc;
+        int ssl_err;
+        std::vector<openssl_errc> error_codes;
+    };
+
+    template <typename Op, typename SuccessPred>
+    ssl_call_result ssl_call(const char* operation, SuccessPred&& is_success, Op&& op) {
```
avikivity: Please constrain the parameters.

Possibly add Args&&... so that you can just pass the function name and args for op, instead of creating a lambda.

src/net/tls_openssl.cc
```diff
@@ -1653,6 +1651,26 @@ class openssl_session : public enable_shared_from_this<openssl_session>, public
         tls_log.warn("{} stale error on queue before {}: {}", *this, operation, buf);
     }
 
+    struct ssl_call_result {
+        int rc;
+        int ssl_err;
+        std::vector<openssl_errc> error_codes;
```
avikivity, line 1649: Perhaps: std::expected (if it makes the call sites easier to understand)

pgellert:
[force-push](https://github.com/scylladb/seastar/compare/0ca326081942bea3fe57fe597fcdfe9fff126a4e..3711be573125d8d5dad8137af126a33f199fc2c1): further improvements based on review feedback

pgellert:
[force-push](https://github.com/scylladb/seastar/compare/3711be573125d8d5dad8137af126a33f199fc2c1..b403adc0d8a025a3b740205b54244cdc5ea335da): back out of `std::expected` usage to remain C++20 compatible (which at least the CI builds require)

src/net/tls_openssl.cc
pgellert: This would be nice, but it looks like seastar can't yet use C++23 features :( At least based on the fact that CI is running against both C++20 and C++23.

avikivity: Right (soon)

PR #3387, opened 2026-05-06
build: disable unused command line argument error for C++ module

avikivity:
cmake generates a command line with both --precompile and -c for the module build, but clang rejects -c with an unused command line error. This was reported in [1]. Meanwhile, squelch the error (just for that file).

[1] https://gitlab.kitware.com/cmake/cmake/-/work_items/27794


avikivity:
@scylladb/seastar-maint please take a look

PR #3389, opened 2026-05-07
json: formatter: support types with user-defined conversion to sstring

bhalevy:
## Summary

- Fixes a regression from #3184 where types with a user-defined conversion to `sstring` (but not to `std::string_view`) can no longer be passed to `formatter::to_json` / `formatter::write`
- The issue is that `to_json(std::string_view)` requires two implicit user-defined conversions (`T → sstring → string_view`), which C++ doesn't allow
- Adds constrained template overloads that accept types convertible to `sstring` but not directly to `std::string_view`, performing the conversion explicitly
- Adds a test case to verify the fix

Reported in https://github.com/scylladb/seastar/pull/3184#issuecomment-4392847668

### DIFF
--- a/include/seastar/json/formatter.hh
+++ b/include/seastar/json/formatter.hh
@@ -183,6 +183,18 @@ public:
     static sstring to_json(std::string_view str);
 
     /**
+     * return a json formatted string for types with a user-defined
+     * conversion to sstring but not directly to std::string_view.
+     * Without this, such types would require two user-defined conversions
+     * (T -> sstring -> string_view), which C++ doesn't allow implicitly.
+     */
+    template<typename T>
+    requires (std::convertible_to<const T&, sstring> && !std::convertible_to<const T&, std::string_view>)
+    static sstring to_json(const T& v) {
+        return to_json(std::string_view(sstring(v)));
+    }
+
+    /**
      * return a json formatted int
      * @param n the int to format
      * @return the given int in a json format
@@ -283,6 +295,16 @@ public:
      */
     static future<> write(output_stream<char>& s, std::string_view str) {
         return s.write(to_json(str));
+    }
+
+    /**
+     * write a json formatted string for types with a user-defined
+     * conversion to sstring but not directly to std::string_view.
+     */
+    template<typename T>
+    requires (std::convertible_to<const T&, sstring> && !std::convertible_to<const T&, std::string_view>)
+    static future<> write(output_stream<char>& s, const T& v) {
+        return write(s, std::string_view(sstring(v)));
     }
 
     /**
--- a/tests/unit/json_formatter_test.cc
+++ b/tests/unit/json_formatter_test.cc
@@ -202,3 +202,32 @@ SEASTAR_THREAD_TEST_CASE(formatter_write) {
     });
 #endif
 }
+
+SEASTAR_THREAD_TEST_CASE(formatter_write_strings) {
+    sstring s = "hello, world";
+    const char* expected = "\"hello, world\"";
+    formatter_check_expected(expected, [&s] (auto& out) {
+        json::formatter::write(out, s).get();
+    });
+    formatter_check_expected(expected, [&s] (auto& out) {
+        json::formatter::write(out, std::string(s)).get();
+    });
+    formatter_check_expected(expected, [&s] (auto& out) {
+        json::formatter::write(out, std::string_view(s)).get();
+    });
+    formatter_check_expected(expected, [&s] (auto& out) {
+        json::formatter::write(out, s.c_str()).get();
+    });
+
+    // Test that formatter::write works with a type that has a user-defined
+    // conversion to sstring but not to std::string_view.
+    // See scylladb/seastar#3184.
+    struct sstring_convertible {
+        sstring value;
+        operator sstring() const { return value; }
+    };
+    sstring_convertible sc{s};
+    formatter_check_expected(expected, [&sc] (auto& out) {
+        json::formatter::write(out, sc).get();
+    });
+}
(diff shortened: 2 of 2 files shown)


include/seastar/json/formatter.hh
```diff
@@ -255,6 +255,18 @@ public:
...
     }
 
+    /**
+     * return a json formatted string for types with a user-defined
+     * conversion to sstring but not directly to std::string_view.
+     * Without this, such types would require two user-defined conversions
+     * (T -> sstring -> string_view), which C++ doesn't allow implicitly.
+     */
+    template<typename T>
+    requires (std::convertible_to<T, sstring> && !std::convertible_to<T, std::string_view>)
```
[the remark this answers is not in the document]
bhalevy: Addressed in the force-push: changed to `std::convertible_to<const T&, sstring>` / `!std::convertible_to<const T&, std::string_view>` in both overloads.

include/seastar/json/formatter.hh
```diff
@@ -360,6 +372,16 @@ public:
         });
     }
 
+    /**
+     * write a json formatted string for types with a user-defined
+     * conversion to sstring but not directly to std::string_view.
+     */
+    template<typename T>
+    requires (std::convertible_to<T, sstring> && !std::convertible_to<T, std::string_view>)
```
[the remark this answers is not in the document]
bhalevy: Addressed together with the `to_json` overload above.

tests/unit/json_formatter_test.cc
```diff
@@ -82,6 +82,16 @@ SEASTAR_TEST_CASE(test_strings) {
...
+    // (but not to std::string_view) works with to_json via a single
+    // implicit conversion. See scylladb/seastar#3184.
+    struct sstring_convertible {
+        sstring value;
+        operator sstring() const { return value; }
+    };
+    sstring_convertible sc{s};
+    BOOST_CHECK_EQUAL(expected, formatter::to_json(sc));
+
     return make_ready_future();
```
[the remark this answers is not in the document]
bhalevy: Addressed: added a dedicated `formatter_write_strings` test case that covers all string types (sstring, std::string, std::string_view, const char*) as well as the sstring-convertible type.

bhalevy:
Changes in v2 (force-push):

- **`requires` clause fix**: Changed `std::convertible_to<T, sstring>` to `std::convertible_to<const T&, sstring>` (and similarly for the `std::string_view` check) in both `to_json` and `write` template overloads, so the constraint correctly matches the `const T&` parameter signature.
- **Expanded test coverage**: Added a `formatter_write_strings` test case that covers `sstring`, `std::string`, `std::string_view`, `const char*`, and the sstring-convertible type — mirroring `test_strings` for `to_json`.

include/seastar/json/formatter.hh
```diff
@@ -255,6 +255,18 @@ public:
...
+     * return a json formatted string for types with a user-defined
+     * conversion to sstring but not directly to std::string_view.
+     * Without this, such types would require two user-defined conversions
+     * (T -> sstring -> string_view), which C++ doesn't allow implicitly.
+     */
+    template<typename T>
+    requires (std::convertible_to<const T&, sstring> && !std::convertible_to<const T&, std::string_view>)
+    static sstring to_json(const T& v) {
+        return to_json(std::string_view(sstring(v)));
+    }
```
avikivity: Please put this next to the other to_json string-like overload.

include/seastar/json/formatter.hh
```diff
@@ -360,6 +372,16 @@ public:
...
 
+    /**
+     * write a json formatted string for types with a user-defined
+     * conversion to sstring but not directly to std::string_view.
+     */
+    template<typename T>
+    requires (std::convertible_to<const T&, sstring> && !std::convertible_to<const T&, std::string_view>)
+    static future<> write(output_stream<char>& s, const T& v) {
+        return write(s, std::string_view(sstring(v)));
+    }
```
avikivity: Probably needs a guard if sstring is defined as std::string.

bhalevy: Claude claimed to have testing it with `Seastar_SSTRING` set to false as well, but I'll double check.
We can put it inside `#ifdef SEASTAR_SSTRING` just in case.

bhalevy: The same problem exists even without `SEASTAR_SSTRING`. The `#ifdef` guard is wrong — we need the template in both cases.

bhalevy:
Note: hit the preexisting https://github.com/scylladb/seastar/issues/3391 while testing with Seastar_SSTRING off.

bhalevy:
Changes in v3 (force-push):

- **Removed `#ifdef SEASTAR_SSTRING` guards**: The template overloads are needed regardless of the `Seastar_SSTRING` setting. When `sstring` is `std::string`, a type with `operator std::string()` still can't implicitly convert to `std::string_view` (two user-defined conversions: `T → std::string → std::string_view`). The `requires` clause naturally becomes unsatisfiable only for types that are already directly convertible to `std::string_view`, so no guard is needed.
- **Moved template overloads** next to their respective `std::string_view` overloads as requested.
- **Verified** the build and tests pass with both `Seastar_SSTRING=ON` and `Seastar_SSTRING=OFF`.

include/seastar/json/formatter.hh
bhalevy: Done in v3 — moved both template overloads to be right next to their respective `std::string_view` overloads.

include/seastar/json/formatter.hh
bhalevy: Tested with `Seastar_SSTRING=OFF` and it turns out the guard is not needed — and in fact would be wrong. When `sstring` is `std::string`, a type with `operator std::string()` still can't implicitly convert to `std::string_view` (two user-defined conversions: `T → std::string → std::string_view`). The `requires` clause naturally handles both cases correctly, so removed the guards in v3.

include/seastar/json/formatter.hh
```diff
@@ -285,6 +297,16 @@ public:
...
```
avikivity, line 308: I have a feeling it can be done in a more neutral way. But not worth obsessing about.

PR #3394, opened 2026-05-07
net: posix: fix pollable_fd_state leak on cross-shard connection forwarding

avikivity:
Fix a memory leak of pollable_fd_state objects during cross-shard
connection forwarding.

The leak has two contributing factors, all in the cross-shard
connection forwarding path of posix_server_socket_impl::accept():

1. posix_ap_server_socket_impl::move_connected_socket() unconditionally
   queues connections into the static conn_q when no acceptor is
   waiting. If the server_socket has already been destroyed (entry
   removed from ports), these queued connections are never consumed.
   Fix by checking ports.contains() before queuing, and closing the
   fd otherwise.

2. posix_ap_server_socket_impl's destructor only removed from ports
   but did not drain conn_q entries for its address. Connections
   queued between abort_accept() (which drains conn_q) and the
   destructor would be leaked. Fix by also erasing from conn_q in
   the destructor.

### DIFF
--- a/src/net/posix-stack.cc
+++ b/src/net/posix-stack.cc
@@ -759,7 +759,9 @@ posix_ap_server_socket_impl::posix_ap_server_socket_impl(int protocol, socket_ad
 }
 
 posix_ap_server_socket_impl::~posix_ap_server_socket_impl() {
-    ports.erase(std::make_tuple(_protocol, _sa));
+    auto t_sa = std::make_tuple(_protocol, _sa);
+    conn_q.erase(t_sa);
+    ports.erase(t_sa);
 }
 
 future<accept_result> posix_ap_server_socket_impl::accept() {
@@ -830,8 +832,11 @@ posix_ap_server_socket_impl::move_connected_socket(int protocol, socket_address 
             i->second.set_exception(std::current_exception());
         }
         sockets.erase(i);
-    } else {
+    } else if (ports.contains(t_sa)) {
         conn_q.emplace(std::piecewise_construct, std::make_tuple(t_sa), std::make_tuple(std::move(fd), std::move(addr), std::move(cth), std::move(addr_data_opt)));
+    } else {
+        // No one is listening anymore; drop the connection to avoid leaking it.
+        fd.close();
     }
 }
 


travisdowns:
> In practice, the coroutine frame did not destroy the structured binding variable between iterations (observed with the proxy protocol path where I/O is performed on the fd before forwarding)

Wait, isn't this a massive compiler bug? Are you saying `fd` dtor is not called at the closing bracket for the loop? What happens later? Is it ever called? Does the next `co_await _lfd.accept();` re-use the space?

avikivity:
> > In practice, the coroutine frame did not destroy the structured binding variable between iterations (observed with the proxy protocol path where I/O is performed on the fd before forwarding)
> 
> Wait, isn't this a massive compiler bug? Are you saying `fd` dtor is not called at the closing bracket for the loop? What happens later? Is it ever called? Does the next `co_await _lfd.accept();` re-use the space?

It's indeed a massive bug.

I started a fruitless bisect, need to retry it.

travisdowns:
> It's indeed a massive bug.
> 
> I started a fruitless bisect, need to retry it.

Does this occur in clang or gcc?

avikivity:
> > It's indeed a massive bug.
> > I started a fruitless bisect, need to retry it.
> 
> Does this occur in clang or gcc?

gcc, we'd know about a such clang bug much sooner.

travisdowns:
> gcc, we'd know about a such clang bug much sooner.

Right. 

I guess I question the wisdom of doing one-off fixes like this one (actually I was unclear what was going on, I had to check that the dtor had the same effect as close()) given the magnitude of the issue, it just seems unsafe to use gcc in this state.

avikivity:
> > gcc, we'd know about a such clang bug much sooner.
> 
> Right.
> 
> I guess I question the wisdom of doing one-off fixes like this one (actually I was unclear what was going on, I had to check that the dtor had the same effect as close()) given the magnitude of the issue, it just seems unsafe to use gcc in this state.

I'm against this fix too, this was before I realized it is likely a gcc bug.

avikivity:
> > > gcc, we'd know about a such clang bug much sooner.
> > 
> > 
> > Right.
> > I guess I question the wisdom of doing one-off fixes like this one (actually I was unclear what was going on, I had to check that the dtor had the same effect as close()) given the magnitude of the issue, it just seems unsafe to use gcc in this state.
> 
> I'm against this fix too, this was before I realized it is likely a gcc bug.

Or rather, the patch is actually good, just not for its stated purpose. The code now can hang on to an fd for an unlimited amount of time, and it's better to drop the fd early.

avikivity:
Likely cause: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=124584

travisdowns:
@avikivity ISTM gcc is simply unsuitable for seastar applications until this is fixed? just very hard to avoid that problem.

avikivity:
> @avikivity ISTM gcc is simply unsuitable for seastar applications until this is fixed? just very hard to avoid that problem.

Coroutines, gcc, structured bindings - pick any two

travisdowns:
> Coroutines, gcc, structured bindings - pick any two

Coroutines, structured bindings ... every time :)

avikivity:
> > Coroutines, gcc, structured bindings - pick any two
> 
> Coroutines, structured bindings ... every time :)

I'm with you here

avikivity:
v2:
 - rebase
 - drop bad workaround for #3431
 - reworded commit message to reflect this

PR #3395, opened 2026-05-07
reactor: always use oneline backtraces in signal handlers and fatal paths

### LINKED ISSUE #16922: Make scylla backtraces always print backtraces in oneline (Aborting/Segmentation Fault)
https://github.com/scylladb/scylladb/issues/5464 was closed as reactor stalls now report backtrace in one line (and all the places in scylla)
but there are still two places in seastar that doesn't use oneline backtraces by default (and don't have configuration for):

- https://github.com/scylladb/seastar/blob/b73f5ffa9e1eb13e9f2c5afde58f70c0df6efc73/src/core/reactor.cc#L4062
- https://github.com/scylladb/seastar/blob/b73f5ffa9e1eb13e9f2c5afde58f70c0df6efc73/src/core/reactor.cc#L4057

fruch:
## Summary

- All backtrace output from signal handlers (SIGSEGV, SIGABRT, SIGILL) and the reactor main loop exception handler now unconditionally use the oneline format.
- This aligns signal handler output with reactor stall backtraces which already default to oneline, making log parsing more reliable and avoiding interlaced multi-line output from concurrent shards.

## Approach

Rather than making this configurable (which ran into template issues with `install_oneshot_signal_handler` as discussed in #2052), this simply hardcodes `oneline=true` for all fatal/crash backtrace paths. The rationale:

1. Community consensus from scylladb/scylladb#5464 is that oneline is strictly better for automated parsing
2. Signal handlers fire at most once (fatal) so there is no performance concern
3. The existing `blocked_reactor_report_format_oneline` option already controls stall detector output separately
4. addr2line.py and backtrace tools already support the oneline format

Supersedes #2052.
Fixes: scylladb/scylladb#16922

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -913,7 +913,7 @@ static void print_with_backtrace(backtrace_buffer& buf, bool oneline) noexcept {
 // where the backtrace itself may crash at some point down the stack
 // while the latter is more efficient and avoids splitting output
 // in the face of concurrent logging by other shards.
-static void print_with_backtrace(const char* cause, bool oneline = false, bool immediate = false) noexcept {
+static void print_with_backtrace(const char* cause, bool oneline = true, bool immediate = false) noexcept {
     backtrace_buffer buf(immediate);
     buf.append(cause);
     print_with_backtrace(buf, oneline);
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -3284,7 +3284,7 @@ int reactor::run() noexcept {
         return do_run();
     } catch (const std::exception& e) {
         seastar_logger.error("{}", e.what());
-        print_with_backtrace("exception running reactor main loop");
+        print_with_backtrace("exception running reactor main loop", true);
```
avikivity: If all the calls now pass true, we can eliminate the parameter and simplify the function.

fruch: v2: Changed the default of `oneline` parameter from `false` to `true` instead. This simplifies all the callsites (they no longer need to pass `true` explicitly) while the stall detector still passes its configurable `_config.oneline` value explicitly — so it's unaffected.

The diff is now just flipping the default value in the function signature.

PR #3396, opened 2026-05-07
scheduling_group: expose usage statistics

travisdowns:
Extends the public interface of `scheduling_group` to expose its task
queue usage statistics — `runtime`, `waittime`, and `starvetime` — via
a new `get_stats()` accessor returning a `scheduling_group::stats`
struct. The underlying counters are already maintained on the per-shard
task queue; this just makes them readable from outside the reactor.

This patch has been carried in the Redpanda fork since 2022. We have
two production uses motivating upstreaming:

1. **Per-scheduling-group Prometheus metric.** Redpanda exports a
   `scheduler_runtime_seconds_total` counter with one series per
   scheduling group, sampled from `get_stats().runtime` in the metric
   callback — visibility into how reactor time is split between SGs,
   useful for capacity planning and noisy-neighbor diagnosis.

2. **Fetch PID controller.** A PID loop samples the Kafka fetch
   scheduling group's `runtime` to compute its utilization fraction
   (delta runtime / delta wallclock), combines it with overall reactor
   busy time, and outputs a per-fetch debounce delay that keeps the
   fetch SG within a configured share of the reactor.

Original: https://github.com/redpanda-data/seastar/commit/a618fe8da390d34609a8debbf6c42df61754602f

### DIFF
--- a/include/seastar/core/scheduling.hh
+++ b/include/seastar/core/scheduling.hh
@@ -426,6 +426,13 @@ public:
     friend unsigned internal::scheduling_group_index(scheduling_group sg) noexcept;
     friend scheduling_group internal::scheduling_group_from_index(unsigned index) noexcept;
 
+    struct stats {
+        sched_clock::duration runtime;
+        sched_clock::duration waittime;
+        sched_clock::duration starvetime;
+    };
+    stats get_stats() const noexcept;
+
     template<typename SpecificValType, typename Mapper, typename Reducer, typename Initial>
     requires requires(SpecificValType specific_val, Mapper mapper, Reducer reducer, Initial initial) {
         {reducer(initial, mapper(specific_val))} -> std::convertible_to<Initial>;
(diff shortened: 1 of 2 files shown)


include/seastar/core/scheduling.hh
```diff
@@ -426,6 +426,13 @@ public:
     friend unsigned internal::scheduling_group_index(scheduling_group sg) noexcept;
     friend scheduling_group internal::scheduling_group_from_index(unsigned index) noexcept;
 
+    struct stats {
```
xemul, line 429: Doxygen doc for public API is missing

xemul:
> Per-scheduling-group Prometheus metric. A public counter
>     `scheduler_runtime_seconds_total` is exported with one series per
>     scheduling group, sampled from `get_stats().runtime` in the metric
>     callback. This gives operators visibility into how reactor time is
>     being divided between scheduling groups.

But reactor already registers metrics that export runtime, waittime and starvetime

travisdowns:
> But reactor already registers metrics that export runtime, waittime and starvetime

It's true, but we export runtime again on a different metrics endpoint.

PR #3397, opened 2026-05-12
websocket: refactor connection lifecycle and harden frame parser

wineway:
## Summary

  Refactor the WebSocket connection lifecycle to replace the previous
  dual-queue producer/consumer model with an on-demand `consume()` loop and a
  per-connection `_send_chain` for outbound serialization, introduce a proper
  state machine for orderly close, and harden the frame parser against RFC 6455
  edge cases.

  ## Motivation

  The previous implementation used two `queue<temporary_buffer>` mailboxes of size
  512 between the network I/O loop and the user handler. That design:

  - Buffered an unbounded amount of work in the queues without bounded backpressure
    semantics that the handler could observe.
  - Had no proper CLOSE handshake — receiving a peer CLOSE could lead to a second
    CLOSE being sent, and there was no timeout for a peer that never responded.
  - Conflated stream lifetime with the queue's drain state via `_done`/`_half_close`
    booleans, which were hard to reason about across re-entrant calls.
  - Did not bound parser memory: an attacker-supplied 64-bit length field would
    drive an allocation of that size before any validation.
  - Had a long-standing bug in `is_opcode_known()` (`opcode < 0xA`) that rejected
    legitimate PONG frames.

  The dual-queue design made WebSocket state management hard to reason about, so this PR refactors it into an explicit lifecycle state machine.

  ## Key changes

  ### Connection lifecycle

  - New `lifecycle_state { running, closing_local, closing_peer, tearing_down,
    torn_down }` plus `shared_future`-backed `_close_future` and
    `_tear_down_future` for idempotent, concurrent-safe close.
  - `initiate_close_handshake()` sends CLOSE, arms a configurable `close_timeout`
    (default 200 ms, `lowres_clock`) that shuts the socket read side down if the
    peer never echoes CLOSE, then tears the streams down.
  - `respond_to_peer_close()` handles peer-initiated CLOSE on its own path.
  - A re-entrant close from inside `consume()` (the handler closes its output
    stream in response to a peer frame) skips `wait_close()` to avoid
    deadlocking against the in-progress read; gated by `_in_consume`.
  - `tear_down_streams()` relies on `input_stream::close()` issuing `SHUT_RD` so a blocked `in.read()` is unblocked without an explicit
    `shutdown_input()` at the `client::close()` site.

  ### Outbound serialization

  Outbound frames are serialized through a per-connection `_send_chain` future.
  Once any send fails, the chain stays poisoned until `drain_send_chain()` runs
  at teardown, so a write failure is terminal for the send path. This matches
  WebSocket's "any framing error is fatal" model and makes the behavior testable
  (`test_websocket_send_chain_failure_is_terminal_until_drained`).

  ### Parser hardening

  - Added `connection_options::max_payload_length` (default 16 MiB) so an
    attacker-controlled 64-bit length field cannot drive arbitrarily large
    allocations. The check fires before the payload buffer is allocated.
  - Enforced RFC 6455 §5.5 control-frame limits: CLOSE / PING / PONG payloads
    must be ≤ 125 bytes and must not be fragmented.
  - Fixed `is_opcode_known()` to accept PONG (`0xA`); the previous
    `opcode < 0xA` check rejected legitimate PONG frames.

  ### Public API

  - New `connection_options { max_payload_length, output_buffer_size,
    close_timeout }`, threaded through `basic_connection`, `server`, `client`,
    `server_connection`, and `client_connection` constructors.
  - New `server::server(connection_options)` ctor.
  - New `server::listen(server_socket)` overload (accept on a pre-created
    listener).
  - New `websocket::client::shutdown()` — an out-of-band escape hatch for the
    case where the outbound path itself is stalled (peer recv-window full, dead
    peer holding the connection open). Normal teardown still goes through
    `close()`.

  ## Behavior changes

  - `client::close()` now resets `_conn` after teardown, so a `client` is
    single-use across `connect`/`close`. A second `connect()` is guarded by
    `SEASTAR_ASSERT(!_conn)`.
  - `server::stop()` no longer initiates a WebSocket CLOSE handshake on active
    connections; it shuts inputs down and lets the task gate drain. This is the
    abortive-teardown semantics documented in `server.hh` and
    `doc/websocket.md`.
  - Existing protected methods `read_one()`, `response_loop()`, and
    `server_connection::read_loop()` are removed.

  ## Tests

  `tests/unit/websocket_test.cc` grows by ~1062 lines. New coverage:

  - Parser: unknown opcodes, oversized control frames, fragmented control
    frames, oversized data frames declared before allocation, configurable
    per-parser limit.
  - Close lifecycle: close timer firing when the peer never echoes CLOSE,
    close-handshake send-failure teardown, double close (both `(false, false)`
    and `(true, true)`), `respond_to_peer_close` write failure still tearing
    down, close from inside the consume path not deadlocking.
  - Send chain: poisoned send chain is terminal until drained; concurrent
    `send_data` calls are FIFO-serialized on the wire.
  - Input path: masked-frame unmasking on the server side, peer CLOSE while the
    handler is mid-read, invalid frame reporting once and then EOF, `wait_close()`
    stopping on parse error.
  - Server / client: `server::stop()` while handlers are blocked in read,
    `client::shutdown()` before / during / after the connect handshake,
    `process()` propagating handler exceptions, `client::process` preserving
    the handler exception even when output close fails.

  ## Refs

  - RFC 6455 §1.3 (handshake), §5.1 (opcodes), §5.5 (control frames), §5.5.1
    (close handshake), §5.5.2 (ping/pong)

  ## AI assistance disclosure

  Parts of this PR were produced with AI assistance:

  - **Test cases** (`tests/unit/websocket_test.cc`): generated and iterated with
    Codex and Claude.
  - **Documentation** (`doc/websocket.md`, doxygen comments in
    `include/seastar/websocket/*.hh`): drafted and refined with Codex and
    Claude.
  - **Code review**: a structured multi-reviewer pass was run with both Codex
    and Claude (via the compound-engineering review pipeline); findings were
    triaged and the substantive ones (close-handshake resource cleanup, parser
    opcode default-arm safety, doxygen accuracy) addressed before submission.

  The C++ refactor itself — connection lifecycle, send-chain serialization, and
  parser hardening — was authored manually; the AI tools contributed test
  coverage, prose, and review feedback.

### DIFF
--- a/demos/websocket_client_demo.cc
+++ b/demos/websocket_client_demo.cc
@@ -97,6 +97,11 @@ int main(int argc, char** argv) {
             duration);
         co_await seastar::sleep_abortable(std::chrono::seconds(duration));
         ws_demo_logger.info("Done, closing connection.");
+        // close() drives the CLOSE handshake and tears the streams down,
+        // which unblocks the handler's in.read() via SHUT_RD. shutdown() is
+        // an OOB escape hatch for outbound-stall cases; here it is harmless
+        // belt-and-suspenders.
+        ws_client.shutdown();
         co_await ws_client.close();
     });
 }
--- a/doc/websocket.md
+++ b/doc/websocket.md
@@ -7,7 +7,7 @@ https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocke
 
 ## Handlers
 
-A WebSocket server needs a user-defined handler in order to be functiomal. WebSocket specification defines a concept of subprotocols, and Seastar WebSocket server allows registering a single handler per subprotocol.
+A WebSocket server needs a user-defined handler in order to be functional. WebSocket specification defines a concept of subprotocols, and Seastar WebSocket server allows registering a single handler per subprotocol.
 
 Each subprotocol has a unique name and is expected to be sent by the connecting client during handshake,
 by sending a `Sec-Websocket-Protocol` header with the chosen value.
@@ -34,6 +34,68 @@ ws.register_handler("echo", [] (input_stream<char>& in, output_stream<char>& out
 
 Note: the developers should assume that the input stream provides decoded and unmasked data - so the stream should be treated as if it was backed by a TCP socket. Similarly, responses should be sent to the output stream as is, and the WebSocket server implementation will handle its proper serialization, masking and so on.
 
+## Connection lifecycle
+
+The handler owns the normal WebSocket connection lifecycle. To actively close a
+connection, close the output stream passed to the handler:
+
+```cpp
+ws.register_handler("echo", [] (input_stream<char>& in, output_stream<char>& out) -> future<> {
+    while (true) {
+        auto buf = co_await in.read();
+        if (buf.empty()) {
+            break;
+        }
+        co_await out.write(std::move(buf));
+        co_await out.flush();
+    }
+    co_await out.close();
+});
+```
+
+On EOF or after the connection is torn down, `in.read()` returns an empty
+buffer. If WebSocket frame parsing fails, the current `in.read()` fails with
+`websocket::exception`; after the connection teardown completes, later reads
+return an empty buffer.
+
+The frame parser enforces RFC 6455 control-frame limits: CLOSE, PING and PONG
+frames must not be fragmented and their payload is limited to 125 bytes. Data
+frames are limited to 16 MiB by default to avoid allocating memory from
+untrusted frame lengths; the limit is configurable via
+`connection_options::max_payload_length` passed to the server/client
+constructor. A frame that exceeds either limit fails the active `in.read()`
+with `websocket::exception`.
+
+Outgoing frames are serialized internally. If an output write or close fails,
+the connection's send path is considered failed: later writes fail with the same
+exception until the connection is torn down. Applications should treat any
+failed output operation as terminal for that WebSocket connection.
+
+When the application initiates the close handshake (by closing the output
+stream from inside the handler, or `websocket::client::close()` from outside),
+the implementation waits up to `connection_options::close_timeout` (200 ms by
+default) for the peer's CLOSE response before tearing down the streams.
+
+For `websocket::client`, `close()` initiates a WebSocket CLOSE handshake and
+tears down the connection's streams; the stream teardown shuts down the socket
+read side, so a handler blocked in `in.read()` is unblocked and `close()`
+resolves once the processing task has finished.
+
+`close()` may fail to make progress only when the outbound path itself stalls
+(for example, the peer's TCP receive window is full so the CLOSE frame can
+never be sent). In that case `shutdown()` provides an out-of-band escape hatch
+that synchronously shuts the connection's input down so the handler can return
+and `close()` can complete:
+
+```cpp
+ws_client.shutdown();   // only needed when the outbound is stalled
+co_await ws_client.close();
+```
+
+`websocket::server::stop()` stops accepting new connections and tears down
+active connections. It does not initiate a WebSocket CLOSE handshake for those
+active connections.
+
 ## Error handling
 
 Registered WebSocket handlers can throw arbitrary exceptions during their operation. Currently, exceptions that aren't explicitly handled within the handler will cause the established WebSocket connection to be terminated, and a proper error message will be logged.
@@ -42,4 +104,3 @@ Registered WebSocket handlers can throw arbitrary exceptions during their operat
 
 Implementation of Secure WebSocket standard, based on HTTPS is currently work in advanced progress. Once reviewed and merged, this section will contain documentation for it.
 Ref: https://github.com/scylladb/seastar/pull/1044
-
--- a/include/seastar/websocket/common.hh
+++ b/include/seastar/websocket/common.hh
@@ -34,6 +34,23 @@ using handler_t = std::function<future<>(input_stream<char>&, output_stream<char
 
 class server;
 
+/// RFC 6455 section 1.3 accept-key GUID shared by server and client handshakes.
+constexpr std::string_view websocket_magic_guid = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
+
+/*!
+ * \brief Configuration knobs for a WebSocket connection.
+ */
+struct connection_options {
+    /*!
+     * \brief Maximum accepted payload size for incoming data frames.
+     */
+    uint64_t max_payload_length = websocket_parser::default_max_payload_length;
+    /*!
+     * \brief Buffer size of the output_stream wrapping the connection sink.
+     */
+    size_t output_buffer_size = 8192;
+};
+
 /// \defgroup websocket WebSocket
 /// \addtogroup websocket
 /// @{
@@ -147,15 +164,16 @@ public:
      * \param is_client if true, this is a client-side connection (sends masked
      *        frames and expects unmasked frames from server)
      */
-    basic_connection(connected_socket&& fd)
+    basic_connection(connected_socket&& fd,
+            connection_options options = {})
         : _fd(std::move(fd))
         , _read_buf(_fd.input())
         , _write_buf(_fd.output())
-        , _websocket_parser(!is_client)
+        , _websocket_parser(!is_client, options.max_payload_length)
         , _input_buffer{PIPE_SIZE}
         , _input(data_source{std::make_unique<connection_source_impl>(&_input_buffer)})
         , _output_buffer{PIPE_SIZE}
-        , _output(data_sink{std::make_unique<connection_sink_impl>(&_output_buffer)})
+        , _output(data_sink{std::make_unique<connection_sink_impl>(&_output_buffer)}, options.output_buffer_size)
     {
     }
 
--- a/include/seastar/websocket/common.hh
+++ b/include/seastar/websocket/common.hh
@@ -146,6 +146,7 @@ protected:
     connected_socket _fd;
     input_stream<char> _read_buf;
     output_stream<char> _write_buf;
+    future<> _send_chain;
     bool _done = false;
     // TODO: implement https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.2
     bool _half_close = false;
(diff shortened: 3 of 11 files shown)


xemul:
Please, split the change into several smaller patches each doing small self-contained fix. Otherwise it's impossible to review it carefully. The description of the patch already contains the list of things done, likely each item in it deserves its own patch

wineway:
> Please, split the change into several smaller patches each doing small self-contained fix. Otherwise it's impossible to review it carefully. The description of the patch already contains the list of things done, likely each item in it deserves its own patch

@xemul  I split the change into 6 commits now. Could you please take another look?
f9629088162a5ce97ad76053e15deefec165cb1c is still a bit large though :(

wineway:
@scylladb/seastar-maint ping

include/seastar/websocket/parser.hh
```diff
@@ -129,16 +130,35 @@ class websocket_parser {
...
+     * \param max_payload_length maximum accepted payload size for data frames.
+     *        The default is 16 MiB. Control frames are always limited to the
+     *        RFC 6455 maximum of 125 bytes and must not be fragmented. A frame
+     *        violating either limit puts the parser into the error state
+     *        (\ref is_valid returns false) and stops consumption; the caller
+     *        sees this as a websocket::exception on the active read.
      */
-    explicit websocket_parser(bool require_mask = true)
+    explicit websocket_parser(bool require_mask = true,
+            uint64_t max_payload_length = default_max_payload_length)
```
xemul, line 158: After next patch, that default arument is no longer needed. So the patch split could have been better -- first thread the option here and there, then start using it

src/websocket/server.cc
```diff
@@ -138,7 +138,7 @@ future<> server_connection::read_http_upgrade_request() {
     sstring sec_key = req->get_header("Sec-Websocket-Key");
     sstring sec_version = req->get_header("Sec-Websocket-Version");
 
-    auto sha1_input = fmt::format("{}{}", sec_key, magic_key_suffix);
+    auto sha1_input = fmt::format("{}{}", sec_key, websocket_magic_guid);
```
xemul, line 141: You explicitly stated this in the patch comment, that this commit does more than one logical thing. Once more example of where the split could have been better

src/websocket/common.cc
```diff
@@ -60,6 +61,24 @@ static void apply_mask(char* data, size_t len, uint32_t masking_key) {
 
 template <bool is_client, bool text_frame>
 future<> basic_connection<is_client, text_frame>::send_data(opcodes opcode, temporary_buffer<char> buff) {
+    promise<> p;
```
xemul, line 64: This is a very good example of a patch whose commit text doesn't help :) It explain _what_ the patch does, but one can easily read it from the code. Instead, the commit text must concentrate on explaining _why_ it does so, i.e. -- the motivation for the change.

src/websocket/common.cc
```diff
@@ -60,6 +61,24 @@ static void apply_mask(char* data, size_t len, uint32_t masking_key) {
...
+    return std::exchange(_send_chain, p.get_future())
+        .then([this, opcode, buff = std::move(buff)] () mutable -> future<> {
+            return do_send(opcode, std::move(buff));
+        }).then_wrapped([p = std::move(p)] (future<> f) mutable {
+            if (f.failed()) {
+                auto e = f.get_exception();
+                p.set_exception(e);
+                return make_exception_future(std::move(e));
```
xemul, line 72: Very similar chaining exists in RPC's connection::send(), but in there it doesn't propagate the exception around. Is the per-caller exception propagation intentional here? Or is the RPC-style "initiate teardown and let in.read() signal the handler" model sufficient?

include/seastar/websocket/common.hh
```diff
@@ -21,34 +21,73 @@
...
+#include <seastar/core/shared_future.hh>
+#include <seastar/core/temporary_buffer.hh>
 #include <seastar/net/api.hh>
 #include <seastar/util/log.hh>
 #include <seastar/websocket/parser.hh>
 
 namespace seastar::experimental::websocket {
 
+/*!
+ * \brief Application handler for an established WebSocket connection.
```
xemul, line 40: The same comment to the patch as for the previous one -- please, explain _why_, not just _what_

include/seastar/websocket/common.hh
```diff
@@ -68,132 +107,220 @@ public:
...
+            : _conn(conn) {}
 
         virtual future<buff_t> get() override {
-            return data->pop_eventually().then_wrapped([](future<buff_t> f){
-                try {
-                    return make_ready_future<buff_t>(std::move(f.get()));
-                } catch(...) {
-                    return current_exception_as_future<buff_t>();
-                }
+            return do_until([this] () { return _conn._pending_inbound.has_value(); }, [this] () {
```
xemul, line 148: Why do you need this _pending_inbound indirection? Instead of making the consume() method return a future with the buffer

include/seastar/websocket/common.hh
```diff
@@ -68,132 +107,220 @@ public:
...
+    std::optional<timer<lowres_clock>> _close_timer;
+    lowres_clock::duration _close_timeout;
     future<> _send_chain;
-    bool _done = false;
-    // TODO: implement https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.2
-    bool _half_close = false;
+    lifecycle_state _state;
+    bool _in_consume = false;
+    std::optional<shared_future<>> _close_future;
+    std::optional<shared_future<>> _tear_down_future;
```
xemul, line 214: Please, explain why shared_future is needed. I don't see if there are multiple waiter-s on it in this patch, but maybe I'm missing something

src/websocket/common.cc
```diff
@@ -111,89 +116,160 @@ future<> basic_connection<is_client, text_frame>::do_send(opcodes opcode, tempor
...
             case opcodes::TEXT:
             case opcodes::BINARY:
-                return _input_buffer.push_eventually(_websocket_parser.result());
+                _pending_inbound = _websocket_parser.result();
+                return make_ready_future();
             case opcodes::CLOSE:
                 websocket_logger.debug("Received close frame.");
                 // datatracker.ietf.org/doc/html/rfc6455#section-5.5.1
-                return close(true);
+                return respond_to_peer_close();
```
xemul, line 239: So the input data source get() would only return empty buffer to the caller after the teardown is complete. Is that intentional?

src/websocket/common.cc
```diff
@@ -111,89 +116,160 @@ future<> basic_connection<is_client, text_frame>::do_send(opcodes opcode, tempor
...
+template <bool is_client, bool text_frame>
+future<> basic_connection<is_client, text_frame>::wait_close() {
+    return repeat([this] () {
+        return _read_buf.consume(_websocket_parser).then([this] () mutable {
+            if (_websocket_parser.eof()
+                || !_websocket_parser.is_valid()
+                || _websocket_parser.opcode() == opcodes::CLOSE) {
+                return make_ready_future<stop_iteration>(stop_iteration::yes);
+            };
+            return make_ready_future<stop_iteration>(stop_iteration::no);
```
xemul, line 209: It probably still needs to handle PINGs. RFC 6455 §5.5.2 is pretty direct: `Upon receipt of a Ping frame, an endpoint MUST send a Pong frame in response.` and makes no exception to whether we're in close or not

src/websocket/server.cc
```diff
@@ -103,17 +104,21 @@ void server_connection::on_new_connection() {
...
 future<> server_connection::process() {
-    return when_all_succeed(read_loop(), response_loop()).discard_result().handle_exception([] (const std::exception_ptr& e) {
-        websocket_logger.debug("Processing failed: {}", e);
-    });
+    return read_http_upgrade_request()
+        .then([this] () {
+            return _handler(_input, _output);
+        }).handle_exception([this] (std::exception_ptr e) {
+            return close(false).handle_exception([] (auto) {}).then([e = std::move(e)] () {
+                return make_exception_future(e);
```
xemul, line 112: Old process() didn't propagate handler excaptions back. And here "back" is the ignored future of accept_one()'s try_with_gate call. Abandoned exceptional futures spoil logs with warnings

include/seastar/websocket/common.hh
```diff
@@ -68,132 +107,220 @@ public:
...
 template <bool is_client = false, bool text_frame = false>
 class basic_connection : public boost::intrusive::list_base_hook<> {
+    template <bool test_is_client, bool test_text_frame>
+    friend struct basic_connection_test_accessor;
+
 protected:
     using buff_t = temporary_buffer<char>;
-
+    /*!
+     * \brief Connection close lifecycle.
```
xemul, line 123: I would appreciate if the closing state machine is detached into its own patch. Unless there are unavoidable consequences that require all other changes in this patch to exist (e.g. the sink/source rework)

demos/websocket_client_demo.cc
```diff
@@ -97,6 +97,11 @@ int main(int argc, char** argv) {
             duration);
         co_await seastar::sleep_abortable(std::chrono::seconds(duration));
         ws_demo_logger.info("Done, closing connection.");
+        // close() drives the CLOSE handshake and tears the streams down,
+        // which unblocks the handler's in.read() via SHUT_RD. shutdown() is
+        // an OOB escape hatch for outbound-stall cases; here it is harmless
+        // belt-and-suspenders.
+        ws_client.shutdown();
```
xemul, line 104: If shutting down a socket before close in unnecessary, demo should avoid doing it

doc/websocket.md
```diff
@@ -34,6 +34,68 @@ ws.register_handler("echo", [] (input_stream<char>& in, output_stream<char>& out
...
+            break;
+        }
+        co_await out.write(std::move(buf));
+        co_await out.flush();
+    }
+    co_await out.close();
+});
+```
+
+On EOF or after the connection is torn down, `in.read()` returns an empty
```
xemul, line 56: There's third outcome of read(). When teadown happens (on bad frame or concurrent close) it may resolve with exception. That's new behavior introduced by previous patch, AFAICS

src/websocket/common.cc
wineway: Yes, I misread §5.5.1 — it only prohibits data frames, not control frames. I'll update the implementation.

src/websocket/common.cc
wineway: My initial intention was to distinguish a peer-initiated close from a send failure. After thinking about it more, I agree this doesn't need to be reported through each send caller; the RPC-style model is sufficient.

PR #3398, opened 2026-05-12
iostream: fix default skip() implementation when skipping-to-eof

avikivity:
skip() should never consume the final empty buffer indicating eof, but if the stream is empty and n=0, it does. This can cause get() after eof, which is a contract violation.

Fix by checking for n=0 and returning immediately in that case.

### DIFF
--- a/tests/unit/input_stream_test.cc
+++ b/tests/unit/input_stream_test.cc
@@ -22,6 +22,7 @@
 #include <seastar/core/iostream.hh>
 #include <seastar/testing/thread_test_case.hh>
 #include <functional>
+#include <numeric>
 #include <sstream>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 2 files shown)


avikivity:
Credit to Claude for the test case.

travisdowns (approved):

xemul:
There's test_read_invariants that, apparently, skips the sequence of steps that leads to this?

tests/unit/input_stream_test.cc
```diff
@@ -300,6 +301,63 @@ static void run_sequence(const std::string& data,
...
+            std::fill_n(buf.get_write(), sz, 'x');
+            bufs.push_back(std::move(buf));
+        }
+
+        auto* raw = new tracking_data_source_impl(std::move(bufs));
+        auto& tracker = *raw;
+        auto ds = data_source(std::unique_ptr<data_source_impl>(raw));
+
+        // Skip exactly to EOF.
+        auto leftover = ds.skip(total).get();
```
xemul, line 328: The added if (n == 0) check is not triggered by this call, isn't it? The data_source::skip() just calls data_source_impl::skip() and total != 0 here

avikivity:
> There's test_read_invariants that, apparently, skips the sequence of steps that leads to this?

That's for input_stream, and input_stream::skip() might protect data_source::skip() from being called in the wrong cases. Didn't check. Seems reasonable to check it directly, no?

tests/unit/input_stream_test.cc
avikivity: The test fails without the fix. See test({}) below, the stream is of length 0 and so is the call.

xemul: Ah, it's zero for the very first test({}) case. Why having the remaining tests with non-zero total? The skip(0) is not exploited in those, so the thing partially duplicates the invariants test from the same test file

xemul:
> That's for input_stream, and input_stream::skip() might protect data_source::skip() from being called in the wrong cases. Didn't check. 

Yes, input_stream::skip(0) swallows the skip and doesn't call the underlying data source

> Seems reasonable to check it directly, no?

Well, input_stream may protect the underlying data source from some other violations as well. It's more correct to ask Claude to generate invariants test for data_source too then. Or modify the existing one to test both, input_stream and data_source. Or not to re-introduce tests with magic vectors of buffers that don't have anything with skip(0) and only check the desired behavior ([this](https://github.com/scylladb/seastar/pull/3398#discussion_r3228497997) question rephrased)

tests/unit/input_stream_test.cc
avikivity: Yes, Claude generated all the test cases except {}, until I helpfully reminded it that it forgot one case.

avikivity:
> > That's for input_stream, and input_stream::skip() might protect data_source::skip() from being called in the wrong cases. Didn't check.
> 
> Yes, input_stream::skip(0) swallows the skip and doesn't call the underlying data source
> 
> > Seems reasonable to check it directly, no?
> 
> Well, input_stream may protect the underlying data source from some other violations as well. It's more correct to ask Claude to generate invariants test for data_source too then. Or modify the existing one to test both, input_stream and data_source. Or not to re-introduce tests with magic vectors of buffers that don't have anything with skip(0) and only check the desired behavior ([this](https://github.com/scylladb/seastar/pull/3398#discussion_r3228497997) question rephrased)

Makes sense, I'll widen the test.

PR #3399, opened 2026-05-12
file: assert dma alignments are powers of two

travisdowns:
Add a `validated()` method on the internal `alignments` struct that asserts each of `memory`, `disk_read`, `disk_write`, and `disk_overwrite` is a power of two, and call it from the two producers (`filesystem_alignments` and `blkdev_alignments`).

The kernel/FS ABIs that supply these values (`BLKSSZGET`, `XFS_IOC_DIOINFO`, `statx dio_*_align`, `posix_memalign`) all guarantee power-of-two alignments, and seastar's own I/O paths consume them via `x & (align - 1)` masking (e.g. computing the unaligned prefix of a read offset). A non-power-of-two value would silently produce wrong offsets, so we check once at the producer rather than per-I/O.

Also documents the power-of-two guarantee on the four public accessors (`disk_read_dma_alignment` / `disk_write_dma_alignment` / `disk_overwrite_dma_alignment` / `memory_dma_alignment`).

### DIFF
--- a/include/seastar/core/file.hh
+++ b/include/seastar/core/file.hh
@@ -311,12 +311,16 @@ public:
     // we will end up with various pages around, some of them with
     // overlapping ranges. Those would be very challenging to cache.
 
-    /// Alignment requirement for file offsets (for reads)
+    /// Alignment requirement for file offsets (for reads).
+    ///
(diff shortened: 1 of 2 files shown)


travisdowns:
@scylladb/seastar-maint quick one

PR #3400, opened 2026-05-13
tls: improved support for builds with a single backend (GnuTLS or OpenSSL)

travisdowns:
The thrust here is that the dual backend TLS mode has a few possible footguns, in particular that certain error "constants" and helper functions will return wrong values or trigger UB if they are called before the TLS backend is initalized (which necessarily happens after main(), in the smp::configure phase).

It is not easy to just solve these in an API-compatible way. So this series tries to improves this in the following ways:

 - When only one backend is configured, which is probably the most common today, we can simply avoid most of the footguns since the backend is known statically so we can do what we did before and everything can work even before main().
 - We add some asserts in the relevant places to catch misuse, e.g., trying to access the provider when none has been set.

Details:

Make seastar's TLS backend choice an explicit build-time setting and let
single-backend builds use the active backend without any reactor-time setup.

* New `configure.py --tls-mode={gnutls,openssl,both}` (default `gnutls`),
  backed by a new public `SEASTAR_TLS_DUAL_BACKEND` compile def when both
  backends are enabled. OpenSSL is no longer auto-enabled — users opt in
  via `--tls-mode=openssl` or `--tls-mode=both`.
* In single-backend builds, the `seastar::tls::ERROR_*` globals are declared
  `const` and statically initialized to the active backend's values, so
  they're valid from static initializers and from unit tests that never
  start a reactor. Today they read silently as 0 before reactor startup,
  which has bitten downstream unit tests that compare against these
  constants without spinning up a reactor.
* In single-backend builds, `internal::crypto::provider()` returns a
  function-local static singleton constructed lazily on first call — no
  `set_provider()` is compiled or needed.
* In dual-backend builds, the runtime-installed provider is now paired
  with `reset_provider()` called from `smp::cleanup()`, so a subsequent
  `app::run()` in the same process starts from a clean slate. The previous
  silent-overwrite behavior obscured cross-app lifecycle bugs.
* The lifecycle is now asserted: `SEASTAR_ASSERT` on double-install in
  `set_provider()`, `SEASTAR_DEBUG_ASSERT` on access-before-install in
  `provider()`. A new `SEASTAR_DEBUG_ASSERT` macro (no-op outside
  Debug/Sanitize/Fuzz) supports the latter.
* CI adds two dedicated jobs to keep the non-default modes covered:
  `build_with_dual_tls` (`--tls-mode=both`) and `build_with_openssl_tls`
  (`--tls-mode=openssl`). The existing matrix exercises the new
  gnutls-only default.

### DIFF
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -48,10 +48,31 @@ class socket_address;
  * Relatively thin SSL wrapper for socket IO.
  * (Can be expanded to other IO forms).
  *
- * The current underlying mechanism is
- * gnutls, however, all interfaces are kept
- * agnostic, so in theory it could be replaced
- * with OpenSSL or similar.
+ * Two underlying mechanisms are supported, GnuTLS and OpenSSL. Each one can
+ * be enabled or disabled at build time (\c Seastar_GNUTLS, \c Seastar_OPENSSL).
+ * When both are compiled in (\c SEASTAR_TLS_DUAL_BACKEND), the active backend
+ * is chosen at reactor startup via the \c --crypto-provider option; when only
+ * one is compiled in, that backend is fixed for the lifetime of the program.
+ *
+ * The interfaces here are kept agnostic so that either backend can be used
+ * without changes to client code.
+ *
+ * \section backend_lifetime When backend-dependent state is valid
+ *
+ * Several entry points below expose state that comes from the active backend
+ * (\ref error_category, \ref backend_name, the \c ERROR_* globals, and any
+ * function that internally creates a TLS session, credentials, or DH params).
+ *
+ * - In **single-backend** builds the backend is fixed at compile time and
+ *   initialized on first use. These entry points are valid at any time,
+ *   including from static initializers and from unit tests that never start
+ *   a reactor, with one exception: the \c ERROR_* globals are
+ *   zero-initialized and only take their final values once the backend is
+ *   first used (at reactor startup at the latest).
+ * - In **dual-backend** builds the backend is installed by
+ *   \c smp::configure() at reactor startup. Calling any backend-dependent
+ *   entry point before that point is undefined; the \c ERROR_* globals in
+ *   particular are zero-initialized and silently read as 0 until then.
  *
  */
 namespace tls {
@@ -696,20 +717,24 @@ namespace tls {
     std::ostream& operator<<(std::ostream&, subject_alt_name_type);
 
     /**
-     * Error handling.
+     * The error_category instance used by exceptions thrown by TLS.
      *
-     * The error_category instance used by exceptions thrown by TLS
+     * See \ref backend_lifetime for when this is valid to call.
      */
     const std::error_category& error_category();
 
     /**
      * Returns the name of the active TLS backend (e.g. "gnutls", "openssl").
+     *
+     * See \ref backend_lifetime for when this is valid to call.
      */
     const char* backend_name();
 
     /**
-     * The more common error codes encountered in TLS.
-     * Not an exhaustive list. Add exports as needed.
+     * The more common error codes encountered in TLS. Not an exhaustive
+     * list — add exports as needed.
+     *
+     * See \ref backend_lifetime for when these are valid to read.
      */
     extern int ERROR_UNKNOWN_COMPRESSION_ALGORITHM;
     extern int ERROR_UNKNOWN_CIPHER_TYPE;
--- a/.github/workflows/gen-matrix.py
+++ b/.github/workflows/gen-matrix.py
@@ -126,6 +126,22 @@ SPECIAL_ITEMS: list[dict[str, Any]] = [
         "mode": "fuzz",
         "test-args": "-- -R 'Seastar.fuzz.'",
     },
+    {
+        "compiler": "clang++-22",
+        "standard": 23,
+        "arch": "x86",
+        "mode": "debug",
+        "options": "--tls-mode=both",
+        "info": "dual-tls, ",
+    },
+    {
+        "compiler": "clang++-22",
+        "standard": 23,
+        "arch": "x86",
+        "mode": "debug",
+        "options": "--tls-mode=openssl",
+        "info": "openssl-tls, ",
+    },
 ]
 
 def _build_include(items: list[dict[str, Any]]) -> CommentedSeq:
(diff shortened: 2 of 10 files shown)


travisdowns:
@scylladb/seastar-maint please take a look, also @dotnwat (this is almost identical to the RP-side changes you reviewed already, I only added the clear_provider stuff which is necessary to support multiple apps)

travisdowns:
@scylladb/seastar-maint ping

dotnwat (approved):
LGTM

> SEASTAR_TLS_DUAL_BACKEND

If in the future we take the next step of dropping GnuTLS, then the off-ramp for this would be to hard code `SEASTAR_TLS_DUAL_BACKEND` as true since that seems like the more general case?

travisdowns:
> If in the future we take the next step of dropping GnuTLS, then the off-ramp for this would be to hard code `SEASTAR_TLS_DUAL_BACKEND` as true since that seems like the more general case?

Well we would have a choice.

We could effectively assume we only have the openssl backend and use the single-backend compile path only (removing the other path). We'd still have the abstraction of dynamic providers mostly in place but it wouldn't solve the global variable problem.

Or you can leave the "single provider" vs multi-provider split in place, like today, and build both ways in CI (as today), and then a sub-choice is what the default is. If there's only one supported backend it seems weird to default it to the mode with the footgun.

If we wanted to fix this "even better" I think we should expose these error codes behind an API call, not as globals, and then this API simply throws if there are > 1 providers and init hasn't been called. Then we can remove this DUAL_MODE stuff in the cmake. That requires deprecation and API levels etc.

avikivity:
> It is not easy to just solve these in an API-compatible way. 

We can go through a deprecation or API_LEVEL cycle if need to avoid contortions.

src/net/tls_gnutls.cc
```diff
@@ -1405,24 +1405,57 @@ std::unique_ptr<tls::dh_params_impl> tls::gnutls::make_dh_params(const tls::blob
...
+const int tls::ERROR_PREMATURE_TERMINATION           = GNUTLS_E_PREMATURE_TERMINATION;
+const int tls::ERROR_PUSH                            = GNUTLS_E_PUSH_ERROR;
+const int tls::ERROR_PULL                            = GNUTLS_E_PULL_ERROR;
+const int tls::ERROR_UNEXPECTED_PACKET               = GNUTLS_E_UNEXPECTED_PACKET;
+const int tls::ERROR_UNSUPPORTED_VERSION             = GNUTLS_E_UNSUPPORTED_VERSION_PACKET;
+const int tls::ERROR_NO_CIPHER_SUITES                = GNUTLS_E_NO_CIPHER_SUITES;
+const int tls::ERROR_DECRYPTION_FAILED               = GNUTLS_E_DECRYPTION_FAILED;
+const int tls::ERROR_MAC_VERIFY_FAILED               = GNUTLS_E_MAC_VERIFY_FAILED;
+#endif // !SEASTAR_TLS_DUAL_BACKEND
+
```
avikivity: As an alternative to this pain, could we translate gnutls/openssl errors to our own errors?

We'd make ERROR_NO_CIPHER_SUITES some fixed enum value, and translate the error when we read it.

travisdowns: That's a good idea. I can see two overall ways:

1) "in place", i.e., actually change the existing globals. I guess it doesn't even need an API change: this is an ABI break but arguably not an API one (though that's only true if folks are only comparing to these enums, not expecting them to be GnuTLS/OpenSSL errors).

2) Introduce them in addition to the existing thing. Deprecate the existing thing.

I guess I would favor 1.

avikivity: > That's a good idea. I can see two overall ways:
> 
> 1. "in place", i.e., actually change the existing globals. I guess it doesn't even need an API change: this is an ABI break but arguably not an API one (though that's only true if folks are only comparing to these enums, not expecting them to be GnuTLS/OpenSSL errors).
> 2. Introduce them in addition to the existing thing. Deprecate the existing thing.
> 
> I guess I would favor 1.

Me2.

Can actually combine 2 and  1: add an enum class for the errors, and constexpr inline int variables initialized from the enum. But it would be overkill, not recommending it.

travisdowns: @avikivity I decided to split this error code stuff out into its own PR: https://github.com/scylladb/seastar/pull/3564

It implements version 2, the one you preferred. The rest of the stuff which remains in this PR should be less controversial.

avikivity: I believe "Me2" was "Me too", not "me two", sorry for the ambiguity.

travisdowns: > It implements version 2, the one you preferred. 

Lol. It briefly crossed by mind but then I settled on this is definitely a request for (2).

avikivity: Well, that will teach me to be more careful.

include/seastar/net/tls.hh
```diff
@@ -48,10 +48,31 @@ class socket_address;
...
+ * - In **single-backend** builds the backend is fixed at compile time and
+ *   initialized on first use. These entry points are valid at any time,
+ *   including from static initializers and from unit tests that never start
+ *   a reactor, with one exception: the \c ERROR_* globals are
+ *   zero-initialized and only take their final values once the backend is
+ *   first used (at reactor startup at the latest).
+ * - In **dual-backend** builds the backend is installed by
+ *   \c smp::configure() at reactor startup. Calling any backend-dependent
+ *   entry point before that point is undefined; the \c ERROR_* globals in
+ *   particular are zero-initialized and silently read as 0 until then.
```
avikivity, line 75: Is there a reason to have such complex lifetime rules? Why not say you can only call those APIs after app_template::run?

avikivity: Well, this is what the series is trying to accomplish. But why, really? Do you really need to look up errors before the application starts?

travisdowns: For unit tests, etc. We hit this bug in practice in redpanda in unit tests.

Beyond tests though its' easy for someone to introduce a static/global const variable in their code which is set to one of these values, so effectively "locking in" the wrong value even if it's not used until after the reactor starts. I would say it is highly unexpected for exposed global variables to have their values changed as a side effect of some other call: it's a dangerous footgun.

That said, I'm OK with requiring these to be accessed before the application starts, but it should be enforced, e.g., by putting the errors in an `ssl_errors` struct which you need to get via an accessor, which asserts if called too early (e.g. the [last paragraph here](https://github.com/scylladb/seastar/pull/3400#issuecomment-4512878227)).

I'm not married to this approach. I don't really like it either. So better suggestions are welcome.

travisdowns: In the meantime I would pull out all the _definitely_ no controversial parts of this PR like the CI change so we at least have coverage of this stuff in the meantime.

travisdowns: _Yet another_ suggestion is we change these error codes to structs with an implicit conversion to `int`. This is "mostly" a compile-time compatible change (not entirely), but it gives us a hook to put our assert that the variables are initialized.

PR #3401, opened 2026-05-13
Fix `-Wmaybe-uninitialized` false positive

dotnwat:
The Seastar CMake build suppresses `-Wmaybe-uninitialized`, however it can still be triggered in public headers (e.g. `future.hh`) and not all projects will want to turn off this warning.

* First commit is the fix
* Second commit is a demonstration test

To reproduce, comment out `#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"` and then build the test with `-DCMAKE_BUILD_TYPE=Release`.

### DIFF
--- a/tests/unit/future_state_maybe_uninit_test.cc
+++ b/tests/unit/future_state_maybe_uninit_test.cc
@@ -1,0 +1,64 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Regression test for a GCC -Wmaybe-uninitialized false positive raised
+// in seastar::future_state<T>::clear(). future_state<T> stores its
+// payload in a discriminated raw-storage union; the payload is
+// constructed in move_it() and destroyed in clear() under the same
+// _u.has_result() predicate, so the invariant "value is constructed
+// iff _u.has_result()" holds at runtime. GCC's data-flow analysis
+// cannot correlate the two checks across function boundaries and, after
+// -O2 inlining of ~future -> ~future_state -> clear() -> destroy_at(),
+// flags the destroy as reading uninitialized memory whenever ~T() reads
+// a member before destroying (e.g. ~foreign_ptr reads _cpu).
+//
+// The minimal trigger we know of is future<foreign_ptr<unique_ptr<T>>>.
+// std::expected payloads and other shapes can amplify the warning to
+// additional fields but are not required to reproduce it.
+//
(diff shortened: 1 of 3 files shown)


dotnwat:
Force-push: disable check with clang because clang doesn't recognize this warning flag.

dotnwat:
Force-push: also forgot to disable the flag for clang in the reproducer

tests/unit/future_state_maybe_uninit_test.cc
```diff
@@ -0,0 +1,64 @@
...
+// -O2 inlining of ~future -> ~future_state -> clear() -> destroy_at(),
+// flags the destroy as reading uninitialized memory whenever ~T() reads
+// a member before destroying (e.g. ~foreign_ptr reads _cpu).
+//
+// The minimal trigger we know of is future<foreign_ptr<unique_ptr<T>>>.
+// std::expected payloads and other shapes can amplify the warning to
+// additional fields but are not required to reproduce it.
+//
+// The CMakeLists.txt entry for this test re-enables
+// -Wmaybe-uninitialized for just this translation unit (upstream
```
travisdowns: "upstream seastar" comment seems out of place here, this is upstream seastar

tests/unit/future_state_maybe_uninit_test.cc
```diff
@@ -0,0 +1,64 @@
...
+// Seastar disables it globally as a PUBLIC compile option on the
+// seastar target) so any regression of the warning fails the build.
```
travisdowns, line 37: I wonder if maybe this global disablement is what should be fixed.

dotnwat: When you say fixed, do you mean that disabling the warning should be injected as required C flags into users of libseastar? If so, it seems pretty aggressive to force application code to opt-in to disabling GCC's data-flow analysis. If you mean disabling the warning and fixing the root causes, then that'd be great but it seems like that is the nature of false-positives?

tests/unit/future_state_maybe_uninit_test.cc
dotnwat: Fixed, thanks.

tests/unit/future_state_maybe_uninit_test.cc
travisdowns: > When you say fixed, do you mean that disabling the warning should be injected as required C flags into users of libseastar? 

No I mean sesatar CMake should not disable that warning, since it hides the errors users might encounter when using seastar (and who do not disable it), which is why you needed the workaround of adding this test with the warning turned back on.

I.e., is seems like the existing approach in seastar is to disable these with pragmas at the call side, so we should not also need a global disable: but maybe if you remove the global disable there are just a flood of them beyond this location?

dotnwat: If we disable the warning globally then differant instances of this issue occur in internal private compilation units which is likely why this is supressed globally, so the fix for the public usage case in future.hh would still be needed.

Good point though about disabling the warning to help catch other cases. Not sure if the exact case I hit is triggered by seastar build itself, but I'll take a look. Maybe there are only a few root cause spots that can be patched up.

travisdowns: Maybe it should be disabled for the demo "apps" that build in this repo, that could catch a lot of the cases since they are users of the seastar public API, though I guess in cases like this it actually depends on the specific calls they make, and how inlining pans out etc.

travisdowns (approved):

travisdowns:
Looks good to me @scylladb/seastar-maint maybe have a look

dotnwat:
FWIW [this is roughly what it looks like](https://github.com/dotnwat/seastar/pull/6) if we wanted to remove the global suppression of the warning. If there is appetite for that I can clean it up.

PR #3402, opened 2026-05-14
file: fix default DMA alignment

avikivity:
Prior to 15adc3ce590 ("file: Apply physical_block_size override to filesystem files"), the default DMA alignment was 4096, the initializer in file::_disk_read_dma_alignment that was not overridden.

After this commit, the default was changed to 512 (see "Linux O_DIRECT minimum" comment), later overriden if we had additional information. But if we did not, that ended up the alignment.

This alignment doesn't work if the device is mounted on a device with 4096 byte sectors.

Change to the filesystem block size, which should always work. It can still be overriden later by XFS specific code.


avikivity:
@xemul please review

bhalevy (approved):
LGTM

PR #3403, opened 2026-05-14
reactor: add EPERM touch_directory diagnostics without behavior change

emdroid:
## Summary
- add targeted diagnostics in `reactor::touch_directory()` when `mkdir` fails with `EPERM`/`EACCES`
- perform a single `file_stat()` on the target path and log the resulting status (exists/type/mode/uid/gid, or stat failure)
- include shard id and process id in the same log line for correlation with host security/audit logs
- keep existing behavior unchanged: after logging, still throw the original filesystem exception so CI failures remain visible

## Why
Issue scylladb/scylladb#28259 is intermittent and hard to reproduce. We need one more failing occurrence with richer context to determine whether directories are actually created despite the error and whether follow-up mitigation (e.g. stat-and-ignore) would be valid.

Refs: scylladb/scylladb#28259

## Notes
- diagnostic-only PR (no success-path masking)
- scoped only to `touch_directory()` (`mkdir` path)

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2459,12 +2459,39 @@ reactor::make_directory(std::string_view name_view, file_permissions permissions
 future<>
 reactor::touch_directory(std::string_view name_view, file_permissions permissions) noexcept {
     auto name = sstring(name_view);
+    auto mode = static_cast<mode_t>(permissions);
     syscall_result<int> sr = co_await _thread_pool->submit<syscall_result<int>>(
             internal::thread_pool_submit_reason::file_operation, [&] {
-        auto mode = static_cast<mode_t>(permissions);
         return wrap_syscall<int>(::mkdir(name.c_str(), mode));
     });
     if (sr.result == -1 && sr.error != EEXIST) {
+        if (sr.error == EPERM || sr.error == EACCES) {
+            // Refs: scylladb/scylladb#28259
+            // Diagnostic-only logging to capture mkdir/stat state on intermittent CI failures.
+            auto pid = ::getpid();
+            auto shard_id = this_shard_id();
+            sstring stat_state;
+            try {
+                auto sd = co_await file_stat(name, follow_symlink::no);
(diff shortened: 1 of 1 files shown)


src/core/reactor.cc
```diff
@@ -2459,12 +2459,39 @@ reactor::make_directory(std::string_view name_view, file_permissions permissions
...
     auto name = sstring(name_view);
+    auto mode = static_cast<mode_t>(permissions);
     syscall_result<int> sr = co_await _thread_pool->submit<syscall_result<int>>(
             internal::thread_pool_submit_reason::file_operation, [&] {
-        auto mode = static_cast<mode_t>(permissions);
         return wrap_syscall<int>(::mkdir(name.c_str(), mode));
     });
     if (sr.result == -1 && sr.error != EEXIST) {
+        if (sr.error == EPERM || sr.error == EACCES) {
+            // Refs: scylladb/scylladb#28259
```
xemul, line 2469: Why does it have to go as part of seastar. Scylla touches directories in known places, can we catch the exception there and validate the failed path?

emdroid: Yes we could ... the only thing is that it is called from multiple places (in particular, the most recent hit was in a boost test that uses a different call stack) so we'd need to log on multiple places in the scylladb (basically all places that might call the `touch_directory()`)

avikivity: We can't use Seastar to debug our code.

bhalevy: It is odd that we hit EPERM/EACCES since scylla verifies the owner and permissions of the data directories we use when starting.

I agree that the diagnostics should be implemented in scylla rather than seastar

emdroid: Yeah, it still seems like some more basic filesystem / overlay race (when directory is being created concurrently), but still not really sure about the exact reason. The diag is to verify if the dir has actually been created (and thus https://github.com/scylladb/seastar/pull/3246 would work).

So I'll move the diag to the scylladb itself.

emdroid:
Closing as this should be done in scylladb instead.

PR #3404, opened 2026-05-15
`core`: propagate syscall errors via `coroutine::exception`

WillemKauf:
According to newer guidance from `tutorial.MD` [1], we should tend to prefer propagating exceptions in coroutines using tools such as `co_return coroutine::exception()` or `co_await coroutine::return_exception_ptr()` over bare calls to `throw` within hot paths, since 1. coroutines which throw convert their exceptions to exceptional futures anyways (e.g. no observable difference to callers) and 2. we save on the cost of potential copies and stack unwinding when `throw` is called.

We also started logging these thrown exceptions at `DEBUG` level [2], which means these exceptions are inherently noisier than using the `coroutine` utilities.

For these reasons, rework `syscall_result` and callers of it with the `coroutine` exception utilities.

For more personal motivation, we had a user complain [3] about the increased level of logging/noise when setting all loggers in `redpanda` to `DEBUG`, with many of the logs tracing back all the way to these syscall utilities. 

[1]: https://github.com/scylladb/seastar/commit/56d18e2160308921e48ef939235bc1f550a707d9
[2]: https://github.com/scylladb/seastar/pull/2938/changes/2f802f57ee331997ce47c6be2084f2a58f392294
[3]: https://github.com/redpanda-data/redpanda/issues/30422

### DIFF
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -74,6 +74,7 @@
 #include <seastar/core/io_queue.hh>
 #include <seastar/core/queue.hh>
 #include <seastar/coroutine/as_future.hh>
+#include <seastar/coroutine/exception.hh>
 #include "core/file-impl.hh"
 #include "core/syscall_result.hh"
 #include "core/thread_pool.hh"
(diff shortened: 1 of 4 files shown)


avikivity:
> For more personal motivation, we had a user complain https://github.com/redpanda-data/redpanda/issues/30422 about the increased level of logging/noise when setting all loggers in redpanda to DEBUG, with many of the logs tracing back all the way to these syscall utilities.

This part is a self-inflicted wound.

PR #3408, opened 2026-05-18
reactor_backend: refactor aio retry into a long-lasting coroutine

### LINKED ISSUE #3407: aio_storage_context::reap_completions does not call schedule_retry
when the first iocb is appended to `_pending_aio_retry` in `reap_completions` when the `_aio_retries` queue is empty, it will not get retried until the next `submit_work` call, that does:
```c++
    if (need_to_retry()) {
        schedule_retry();
    }
```

bhalevy:
Followup to #3182.

Refactor `aio_storage_context::schedule_retry` into a long-lasting coroutine that waits on a condition variable when idle, and is awaited in `stop()` via `_pending_aio_retry_fut`.

**Patch series:**

1. **turn aio retry into a long-lasting coroutine** — Replace the one-shot `schedule_retry()` future chain with a persistent `retry_loop()` coroutine started in the constructor. Uses `co_await` on a condition variable instead of `seastar::async`, avoiding the dependency on the thread system being initialized at construction time.

2. **rename schedule_retry to signal_retry_loop** — Rename `schedule_retry()` to `signal_retry_loop()` and make it private, since it is only called internally.

3. **drain pending retries in the retry fiber on shutdown** — Incorporate `reap_pending_retries` logic into the retry loop. When `_stopping`, iocbs are completed with `-ECANCELED` instead of resubmitted. Remove the now-unused `reap_pending_retries()` method.

4. **extract cancel_iocb and retry_iocb helpers** — Extract reusable `cancel_iocb()` and `retry_iocb()` helpers. Use them in the retry loop's shutdown drain and `reap_completions()`. Remove the `allow_retry` parameter.

5. **signal retry_loop from retry_iocb** — Move `signal_retry_loop()` into `retry_iocb()` so the CV is signaled each time an iocb is queued for retry. Fixes #3407.

**Tests:** `socket_test`, `epoll_test`, `reactor_backend_test`, `file_io_test` pass.

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -168,7 +168,7 @@ future<> aio_storage_context::stop() noexcept {
     });
 
     // Wake up the retry coroutine so it can observe _stopping and exit.
-    _retry_cv.signal();
+    signal_retry_loop();
 
     return std::exchange(_pending_aio_retry_fut, make_ready_future<>()).finally([this] {
         return do_until([this] { return !_iocb_pool.outstanding(); }, [this] {
@@ -279,7 +279,7 @@ aio_storage_context::submit_work() {
         // so we don't want to call io_submit() from the reactor thread.
         //
         // Pretend that all aio failed with EAGAIN and submit them
-        // via schedule_retry(), below.
+        // via signal_retry_loop(), below.
         did_work = !_submission_queue.empty();
         for (auto& iocbp : _submission_queue) {
             set_nowait(*iocbp, false);
@@ -301,13 +301,13 @@ aio_storage_context::submit_work() {
     }
 
     if (need_to_retry()) {
-        schedule_retry();
+        signal_retry_loop();
     }
 
     return did_work;
 }
 
-void aio_storage_context::schedule_retry() {
+void aio_storage_context::signal_retry_loop() {
     _retry_cv.signal();
 }
 
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -327,6 +327,9 @@ void aio_storage_context::signal_retry_loop() {
 // Long-lasting retry coroutine.  Loops until _stopping, waiting on
 // _retry_cv when idle.
 future<> aio_storage_context::retry_loop() {
+    timer retry_timer([this] {
+        signal_retry_loop();
+    });
     do {
         // Process retries: loop until both vectors are empty or _stopping is set.
         // While retrying _aio_retries, new retries may be queued onto _pending_aio_retry.
@@ -369,8 +372,10 @@ future<> aio_storage_context::retry_loop() {
             });
         } else if (!_stopping) {
             // Some iocbs failed retry and are left in _aio_retries.
-            // Wait for submit_work or stop to wake us up to re-attempt them or return, respectively.
+            // Wait up to 10ms submit_work or stop to wake us up to re-attempt them or return, respectively.
+            retry_timer.arm(10ms);
             co_await _retry_cv.wait();
+            retry_timer.cancel();
         }
     } while (!_stopping);
 }
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -46,6 +46,7 @@
 #include <seastar/core/internal/uname.hh>
 #include <seastar/core/print.hh>
 #include <seastar/core/reactor.hh>
+#include <seastar/core/coroutine.hh>
 #include <seastar/core/smp.hh>
 #include <seastar/util/defer.hh>
 #include <seastar/util/read_first_line.hh>
@@ -127,6 +128,9 @@ aio_storage_context::aio_storage_context(reactor& r)
                   "Mismatch between maximum allowed io and what the IO queues can produce");
     internal::setup_aio_context(max_aio, &_io_context);
     _r.do_at_exit([this] { return stop(); });
+
+    // Start the long-lasting retry coroutine.
+    _pending_aio_retry_fut = retry_loop();
 }
 
 aio_storage_context::~aio_storage_context() {
@@ -142,11 +146,8 @@ void aio_storage_context::reap_pending_retries() {
     }
     _pending_aio_retry.clear();
 
-    // _aio_retries is empty in the normal call path: this function only runs
-    // after the retry loop's future has resolved, and that loop's predicate
-    // only returns true when both vectors are empty. The drain below is
-    // defensive — if a future change to schedule_retry() leaves entries
-    // here, we still avoid leaking iocbs.
+    // _aio_retries is empty in the normal call path.
+    // However, if the retry_loop() fails to complete all retries, it might leave behind iocbs in _aio_retries.
     for (auto iocb : _aio_retries) {
         _iocb_pool.put_one(iocb);
         auto desc = get_user_data<kernel_completion>(*iocb);
@@ -156,8 +157,8 @@ void aio_storage_context::reap_pending_retries() {
 }
 
 future<> aio_storage_context::stop() noexcept {
-    // Set _stopping first so any reap_completions() and schedule_retry()
-    // calls that race with the drain below see the right state.
+    // Set _stopping first so the retry coroutine and reap_completions()
+    // see the right state.
     _stopping = true;
 
     // Drain items in io_sink (operations without allocated iocbs yet).
@@ -165,6 +166,9 @@ future<> aio_storage_context::stop() noexcept {
         desc->complete_with(-ECANCELED);
         return true;
     });
+
+    // Wake up the retry coroutine so it can observe _stopping and exit.
+    _retry_cv.signal();
 
     return std::exchange(_pending_aio_retry_fut, make_ready_future<>()).finally([this] {
         return do_until([this] { return !_iocb_pool.outstanding(); }, [this] {
@@ -304,41 +308,33 @@ aio_storage_context::submit_work() {
 }
 
 void aio_storage_context::schedule_retry() {
-    // Don't start a new retry loop if either:
-    //   (a) one is already in flight  (!_pending_aio_retry_fut.available()),
-    //   (b) we're shutting down       (_stopping) — stop() will drain the
-    //       queues itself; starting a second loop here would race with
-    //       stop()'s drain on the same iocb pool (the original double-free).
-    if (!_pending_aio_retry_fut.available() || _stopping) {
-        return;
-    }
+    _retry_cv.signal();
+}
 
-    // loop until both _pending_aio_retry and _aio_retries are empty.
-    // While retrying _aio_retries, new retries may be queued onto _pending_aio_retry.
-    _pending_aio_retry_fut = do_until([this] {
-        if (_aio_retries.empty()) {
-            if (_pending_aio_retry.empty()) {
-                return true;
+// Long-lasting retry coroutine.  Loops until _stopping, waiting on
+// _retry_cv when idle.
+future<> aio_storage_context::retry_loop() {
+    do {
+        // Process retries: loop until both vectors are empty or _stopping is set.
+        // While retrying _aio_retries, new retries may be queued onto _pending_aio_retry.
+        // If we're stopping, abandoned iocbs will be reaped and completed with -ECANCELED by reap_pending_retries().
+        while (need_to_retry() && !_stopping) {
+            if (_aio_retries.empty()) {
+                std::swap(_aio_retries, _pending_aio_retry);
             }
-            // _pending_aio_retry, holding a batch of new iocbs to retry,
-            // is swapped with the empty _aio_retries.
-            std::swap(_aio_retries, _pending_aio_retry);
-        }
-        return false;
-    }, [this] {
-        return _r._thread_pool->submit<syscall_result<int>>(
-                internal::thread_pool_submit_reason::aio_fallback, [this] () mutable {
-            auto r = io_submit(_io_context, _aio_retries.size(), _aio_retries.data());
-            return wrap_syscall<int>(r);
-        }).then_wrapped([this] (future<syscall_result<int>> f) {
-            // If submit failed, just log the error and exit the loop.
-            // The next call to submit_work will call schedule_retry again.
-            if (f.failed()) {
-                auto ex = f.get_exception();
-                seastar_logger.warn("aio_storage_context::schedule_retry failed: {}", std::move(ex));
-                return;
+            syscall_result<int> result(0, 0);
+            try {
+                result = co_await _r._thread_pool->submit<syscall_result<int>>(
+                        internal::thread_pool_submit_reason::aio_fallback, [this] () mutable {
+                    auto r = io_submit(_io_context, _aio_retries.size(), _aio_retries.data());
+                    return wrap_syscall<int>(r);
+                });
+            } catch (...) {
+                // If submit failed, just log the error and break out of the inner loop.
+                // The coroutine will wait for the next condition variable signal to retry (or return if stopping).
+                seastar_logger.warn("aio_storage_context::retry_loop failed: {}", std::current_exception());
+                break;
             }
-            auto result = f.get();
             auto iocbs = _aio_retries.data();
             size_t nr_consumed = 0;
             if (result.result == -1) {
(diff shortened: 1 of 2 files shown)


bhalevy:
Force-pushed with the following changes:

- Addressed Copilot review comment: `retry_iocb()` now checks `_stopping` and calls `cancel_iocb()` instead of queuing, so iocbs are properly canceled even if `submit_work()` runs after the retry fiber has exited during shutdown.
- Fused the "extract cancel_iocb" and "extract retry_iocb" patches into a single patch since they are closely related.
- Dropped the "only signal retry CV when fiber might be idle" patch — its `_aio_retries.empty()` guard is now introduced in the final patch where `schedule_retry()` moves into `retry_iocb()` (where it naturally belongs since it's called per-iocb).
- Simplified `reap_completions()`: no more if/else on `_stopping` — just calls `retry_iocb()` which handles both cases.

Series is now 4 patches:
1. `d02014d7e` turn aio retry into a long-lasting fiber
2. `78e6805ba` drain pending retries in the retry fiber on shutdown
3. `208faf256` extract cancel_iocb and retry_iocb helpers
4. `f7a829712` call schedule_retry from retry_iocb (Fixes #3407)

src/core/reactor_backend.cc
```diff
@@ -250,6 +234,20 @@ aio_storage_context::handle_aio_error(linux_abi::iocb* iocb, int ec) {
...
 
+void
+aio_storage_context::cancel_iocb(linux_abi::iocb* iocb) {
+    auto desc = get_user_data<kernel_completion>(*iocb);
+    _iocb_pool.put_one(iocb);
+    desc->complete_with(-ECANCELED);
+}
+
+void
+aio_storage_context::retry_iocb(linux_abi::iocb* iocb) {
```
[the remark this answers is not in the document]
bhalevy: Fixed in [f7a8297](https://github.com/scylladb/seastar/commit/f7a829712cbae5829c5055fa7b1ce91f93007aed)

bhalevy:
**Re: CI failures in `Seastar.unit.socket` and `Seastar.unit.epoll`**

These are preexisting flaky failures, not regressions from this series:

1. Both tests crash with a segfault at the same libc offset (`+0x44ed4`), null pointer dereference (`si_addr: 0x0`), happening immediately after reactor startup (right after the "Reactor backend" / "Perf-based stall detector" log lines).

2. The crash reproduces in both the `linux-aio` and `epoll` backends — our changes only touch `aio_storage_context` which is exclusive to the linux-aio path.

3. The same crash pattern appears across both clang++ and g++ builds at the same libc instruction, suggesting an environment/infrastructure issue rather than a code regression.

4. `socket_test` and `epoll_test` don't exercise AIO retry paths at all.

5. Master itself had a recent flaky CI failure (run 25992344608, `Seastar.unit.websocket` timeout in g++/C++20/debug).

bhalevy:
> **Re: CI failures in `Seastar.unit.socket` and `Seastar.unit.epoll`**
> 
> These are preexisting flaky failures, not regressions from this series:
> 
> 1. Both tests crash with a segfault at the same libc offset (`+0x44ed4`), null pointer dereference (`si_addr: 0x0`), happening immediately after reactor startup (right after the "Reactor backend" / "Perf-based stall detector" log lines).
> 2. The crash reproduces in both the `linux-aio` and `epoll` backends — our changes only touch `aio_storage_context` which is exclusive to the linux-aio path.
> 3. The same crash pattern appears across both clang++ and g++ builds at the same libc instruction, suggesting an environment/infrastructure issue rather than a code regression.
> 4. `socket_test` and `epoll_test` don't exercise AIO retry paths at all.
> 5. Master itself had a recent flaky CI failure (run 25992344608, `Seastar.unit.websocket` timeout in g++/C++20/debug).

I wonder why we haven't seen those failures in https://github.com/scylladb/seastar/pull/3182
Maybe ci didn't run there?

tchaikov:
> I wonder why we haven't seen those failures in #3182 Maybe ci didn't run there?

ci did run there.

i think, these two failures could be a regression introduced in this change: starting the retry fiber from the `io_storage_context` constructor introduces an initialization-order bug. one way to fix it is to move `seastar::thread_impl::init()` ahead of `_backend = rbs.create(*this)` in the reactor constructor, so the thread subsystem is ready before any backend constructor starts a `seastar::thread`. see https://github.com/scylladb/seastar/pull/3409

bhalevy:
Force-pushed with the following changes:

- Converted the `retry_loop()` from a `seastar::async` fiber to a coroutine (`co_await`), avoiding the dependency on the seastar thread system being initialized at `aio_storage_context` construction time. This fixes unit test failures (e.g. socket_test, epoll_test) where the thread system wasn't ready early enough.
- Replaced `#include <seastar/core/thread.hh>` with `#include <seastar/core/coroutine.hh>`.
- Updated comments from "fiber" to "coroutine" throughout.
- Updated the commit message for the first patch accordingly.

src/core/reactor_backend.cc
```diff
@@ -304,41 +311,40 @@ aio_storage_context::submit_work() {
...
-            // is swapped with the empty _aio_retries.
-            std::swap(_aio_retries, _pending_aio_retry);
+// Long-lasting retry coroutine.  Loops until _stopping, waiting on
+// _retry_cv when idle.
+future<> aio_storage_context::retry_loop() {
+    while (!_stopping) {
+        co_await _retry_cv.wait([this] {
+            return _stopping || need_to_retry();
+        });
+        if (_stopping) {
```
xemul: This check is not needed (in the first patch at least) -- the subsequence `while (need_to_retry() && !_stopping)` would not execute and the code would get back to the outer `while (!_stopping)` and "break" naturally

src/core/reactor_backend.cc
```diff
@@ -314,52 +291,60 @@ void aio_storage_context::schedule_retry() {
...
+                    }
+                } else {
+                    nr_consumed = result.result;
+                }
+                _aio_retries.erase(_aio_retries.begin(), _aio_retries.begin() + nr_consumed);
             }
-            _aio_retries.erase(_aio_retries.begin(), _aio_retries.begin() + nr_consumed);
         }
-    }
+    } while (!_stopping);
```
xemul, line 380: nit: It looks like the `do {} while (!_stopping)` could have been introduced in the first patch

src/core/reactor_backend.cc
```diff
@@ -314,52 +291,60 @@ void aio_storage_context::schedule_retry() {
...
+            if (_stopping) {
+                // During shutdown, complete pending iocbs with -ECANCELED
+                // instead of resubmitting them.
+                for (auto iocb : _aio_retries) {
+                    _iocb_pool.put_one(iocb);
+                    auto desc = get_user_data<kernel_completion>(*iocb);
+                    desc->complete_with(-ECANCELED);
+                }
+                _aio_retries.clear();
+            } else {
```
xemul: Ah, the "patch with broken indentation first, restore indentation in the next patch" fits perfectly here

src/core/reactor_backend.cc
```diff
@@ -314,52 +291,60 @@ void aio_storage_context::schedule_retry() {
...
+                        return wrap_syscall<int>(r);
+                    });
                 } catch (...) {
-                    seastar_logger.error("aio retry failed: {}. Aborting.", std::current_exception());
-                    abort();
+                    // If submit failed, just log the error and break out
+                    // of the inner loop.  The coroutine will wait for the
+                    // next condition variable signal to retry.
+                    seastar_logger.warn("aio_storage_context::schedule_retry failed: {}", std::current_exception());
+                    break;
```
xemul: It looks like this break will break everything in case reactor is stopping -- the outer loop would not go on another lap and draining will not finish

src/core/reactor_backend.cc
```diff
@@ -296,15 +297,13 @@ aio_storage_context::submit_work() {
...
-    if (need_to_retry()) {
-        schedule_retry();
-    }
-
     return did_work;
 }
 
 void aio_storage_context::schedule_retry() {
-    _retry_cv.signal();
+    if (_aio_retries.empty()) {
```
xemul: This check prevents background fiber from one spurious wake-up. Does it really worth it?

src/core/reactor_backend.cc
```diff
@@ -296,15 +297,13 @@ aio_storage_context::submit_work() {
...
     }
 
-    if (need_to_retry()) {
-        schedule_retry();
-    }
-
     return did_work;
 }
 
 void aio_storage_context::schedule_retry() {
```
xemul: nit: It looks like this method can be marked private in its class, no longer called from the outside

src/core/reactor_backend.cc
xemul: nit: This comment is a bit invalid after the 4th patch -- "schedule_retry() below" doesn't exist anymore

src/core/reactor_backend.cc
xemul: nit-2: And technically it no longer schedules anything, so maybe worth renaming one as well

src/core/reactor_backend.cc
bhalevy: This should be handled by `reap_pending_retries` that is called by stop() after `_pending_aio_retry_fut` resolves (i.e. `retry_loop` is done)

src/core/reactor_backend.cc
bhalevy: > nit: It looks like this method can be marked private in its class, no longer called from the outside
> nit-2: And technically it no longer schedules anything, so maybe worth renaming one as well

It can be marked private even before this series by the way.
I can incorporate it when renaming to e.g. `signal_retry`

bhalevy:
Force-pushed with the following changes since last push:

- **Patch 1** ("turn aio retry into a long-lasting coroutine"):
  - Removed the redundant `if (_stopping) { break; }` check after `co_await` — the outer loop handles this naturally.
  - Changed from `while (!_stopping)` to `do {} while (!_stopping)` so the loop structure is introduced in this patch.
  - Wrapped the retry submit block in a `{ }` with a FIXME comment to prepare indentation for the next patch that adds the `if (_stopping)` cancellation branch, avoiding a noisy indentation-only diff.
  - Updated warn message to reference `retry_loop` instead of `schedule_retry`.
- **New patch 2** ("rename schedule_retry to signal_retry_loop"): Renamed `schedule_retry()` to `signal_retry_loop()` and made it private.
- **Patch 3** ("drain pending retries in the retry fiber on shutdown"): Replaces the FIXME block with the `if (_stopping)` cancellation branch, restoring proper indentation.
- **Patch 5** ("signal retry_loop from retry_iocb"): Updated stale comment "via signal_retry_loop(), below" to "via retry_iocb() which signals the retry coroutine".

All unit tests pass (socket_test, epoll_test, reactor_backend_test, file_io_test).

src/core/reactor_backend.cc
bhalevy: Agreed, removed. The first patch now uses `do {} while (!_stopping)` so the loop exits naturally without the extra check.

src/core/reactor_backend.cc
bhalevy: Done. The first patch now uses `do {} while (!_stopping)` directly.

src/core/reactor_backend.cc
bhalevy: Done. The first patch wraps the retry submit block in a `{ }` with a FIXME comment, and the third patch ("drain pending retries") replaces it with the `if (_stopping)` branch, restoring proper indentation.

src/core/reactor_backend.cc
bhalevy: The third patch ("drain pending retries in the retry fiber on shutdown") adds an `if (_stopping)` branch that cancels iocbs with `-ECANCELED` before the `else` block containing the `break`. So when stopping, the `break` is only reached on submit failure in the non-stopping path. Any remaining iocbs after the coroutine exits are drained by `reap_pending_retries()` called from `stop()`.

src/core/reactor_backend.cc
bhalevy: Agreed, removed. `signal_retry_loop()` now just calls `_retry_cv.signal()` unconditionally.

src/core/reactor_backend.cc
bhalevy: Done. Added a new patch ("rename schedule_retry to signal_retry_loop") that renames and makes it private.

src/core/reactor_backend.cc
bhalevy: Fixed in the last patch. The comment now reads "via retry_iocb() which signals the retry coroutine".

src/core/reactor_backend.cc
xemul: The reap_pending_retries() is removed in this PR, so the concern still stands :)

xemul: >>> It looks like this break will break everything in case reactor is stopping ...
>>>
>> ... Any remaining iocbs after the coroutine exits are drained by reap_pending_retries() called from stop()
>>
> The reap_pending_retries() is removed in this PR, so the concern still stands :)

@bhalevy , a light reminder ping :point_up:

bhalevy:
Force-pushed with the following changes since last push:

- **Dropped the "drain pending retries in the retry fiber on shutdown" patch** — The retry loop no longer handles `_stopping` inline. Instead, the inner loop condition is `while (need_to_retry() && !_stopping)`, so it simply exits when stopping.

- **Restored `reap_pending_retries()`** — Re-added (now using the `cancel_iocb` helper) and called from `stop()`'s `.finally()` block. This addresses the concern that the `break` on submit failure could leave iocbs unhandled during shutdown — `reap_pending_retries()` drains both `_pending_aio_retry` and `_aio_retries` after the retry loop exits.

- **New patch: "retry_loop: add a timer when waiting after a retry failure"** — When `_aio_retries` is non-empty (from a previous failed submit), arms a 10ms timer before waiting on the CV, ensuring retries are re-attempted in a timely manner even if no new retries are signaled.

- **CV wait predicate changed** — Now `_stopping || !_pending_aio_retry.empty()` instead of `_stopping || need_to_retry()`, so leftover `_aio_retries` from a failed submit use the timer path rather than immediately re-triggering the loop.

- **Rebased** onto current upstream.

All unit tests pass (socket_test, epoll_test, reactor_backend_test, file_io_test).

src/core/reactor_backend.cc
bhalevy: In [2b07e61](https://github.com/scylladb/seastar/commit/2b07e61e4af67c4032cfaee0f8f2732e752f59e7) I restored `reap_pending_retries` as it turned out to be simpler to handle the abandoned retries this way.

That said, I noticed a problem with the previous implementation of the condition variable wait call that would exit immediately after failing to retry, and retry the system call effectively in a tight loop, so I changed the condition to wait only for a new `_pending_aio_retry`, and added a short timer to keep retrying in case no new retry was scheduled.

bhalevy:
Force-pushed with the following change since last push:

- **Patch 5** ("signal retry_loop from retry_iocb"): Kept the conditional `signal_retry_loop()` call in `submit_work()` — it is needed for fast re-attempt when `_aio_retries` has leftover iocbs from a previous failed submit. The 10ms timer in patch 6 remains as a fallback for when `submit_work()` isn't called in a timely manner.

All unit tests pass (socket_test, epoll_test, reactor_backend_test, file_io_test).

src/core/reactor_backend.cc
```diff
@@ -327,14 +327,23 @@ void aio_storage_context::signal_retry_loop() {
 // Long-lasting retry coroutine.  Loops until _stopping, waiting on
 // _retry_cv when idle.
 future<> aio_storage_context::retry_loop() {
+    timer retry_timer([this] {
+        signal_retry_loop();
+    });
     do {
+        if (!_aio_retries.empty()) {
```
xemul: So if there are _aio_retries, the timer is armed to kick the _retry_cv. But when kicked, the cv's predicate checking will see if it's _stopping or has something in (!) _pending_aio_retry. So the purpose of this kick is not clear. If _retry_cv.wait() is stuck, then non-empty _aio_retry + kick in 10ms won't wake it up

bhalevy: Oh, you're right.  I didn't take into account that the predicate is not only an entrance condition but also an exit condition from the wait loop.  I'll need to fix this.
The idea is that normally, `_aio_retries` is empty, and new iocbs may be queued on `_pending_aio_retry`.
When we get here with non-empty `_aio_retries`, it means that we failed to retry some iocbs in the previous round. Since we don't want to keep retrying in a tight loop after such a failure, we want to wait for next submission to wake us up, and limit that wait time to 10ms.

bhalevy:
Force-pushed with the following change since last push:

- **Patch 6** ("retry_loop: add a timer when waiting after a retry failure"): Restructured the retry loop to fix the bug identified in https://github.com/scylladb/seastar/pull/3408#discussion_r3323057783.

**The problem:** The timer was armed *before* `co_await _retry_cv.wait(predicate)`, but the predicate only checked `!_pending_aio_retry.empty()`. So even when the timer fired and signaled the CV, the predicate would still be false (only `_aio_retries` had leftover items), and the wait would never wake up. The timer was effectively useless.

**The fix:** Moved the CV wait to *after* the retry processing loop, with two distinct paths:

1. **`_aio_retries` empty** (normal case): Predicate wait — wakes on `_stopping` or new items in `_pending_aio_retry`.
2. **`_aio_retries` non-empty** (failed retry): Uses `co_await _retry_cv.wait()` *without a predicate* + 10ms timer. This wakes up unconditionally on any signal (timer, `submit_work()`, or `stop()`), avoiding the predicate deadlock.

The loop structure also changed from "wait first, process second" to "process first, wait second" — which is more natural since on initial entry there's nothing to wait for.

All unit tests pass (socket_test, epoll_test, reactor_backend_test, file_io_test).

src/core/reactor_backend.cc
bhalevy: Fixed in [9523930](https://github.com/scylladb/seastar/commit/95239305f9dae7ed8fa8e5bb60cf7d97f16ecd8b)

StephanDollberg:
So I think this change introduced an interesting change. 

aio_storage_context is the only place that internally uses `do_at_exit`. Since this change this actually required as otherwise the condition variable might have outstanding waiters which on destruction will cause scheduling on the already destroyed task queue. 

do_at_exit prevents that but is not being called when a seastar app terminates via exit(0) (not sure whether that is supported or not).

PR #3410, opened 2026-05-19
http/client: handle early server response during streaming body write

levenon:
When using body_writer for streaming uploads (e.g., chunked transfer), the server may send an error response and close the connection before the client finishes writing the body. Previously, do_make_request() serialized write_body and recv_reply, meaning the client could not detect the server's early response until write_body completed — which may never happen for long-lived streams, as writes continue to succeed into kernel/userspace buffers even after the remote end has closed.

Fix this by running write_body and recv_reply concurrently via when_all(). If recv_reply completes first (server responded early), shutdown_output() is called to immediately abort the in-progress write. If write_body completes first (normal flow), recv_reply proceeds as before without shutdown.

This resolves a hang where streaming HTTP clients would remain stuck in write_body indefinitely after the server had already rejected the request.

### DIFF
--- a/src/http/client.cc
+++ b/src/http/client.cc
@@ -151,10 +151,24 @@ future<connection::reply_ptr> connection::do_make_request(const request& req) {
             if (cont) {
                 return make_ready_future<reply_ptr>(std::move(cont));
             }
-
-            return write_body(req).then([this] {
-                return _write_buf.flush().then([this] {
-                    return recv_reply();
+        
+            return do_with(std::exception_ptr{}, [this, &req](std::exception_ptr& write_err) {
+                return when_all(
+                    write_body(req).then([this] {
+                        return _write_buf.flush();
+                    }).handle_exception([this, &write_err](std::exception_ptr e) {
+                        // Save the write error, then send EOF to the server so
+                        // recv_reply can complete (server will close after seeing EOF).
+                        write_err = std::move(e);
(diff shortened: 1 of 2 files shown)


src/http/client.cc
```diff
@@ -151,10 +151,26 @@ future<connection::reply_ptr> connection::do_make_request(const request& req) {
             if (cont) {
                 return make_ready_future<reply_ptr>(std::move(cont));
             }
-
-            return write_body(req).then([this] {
-                return _write_buf.flush().then([this] {
-                    return recv_reply();
+            
+            return do_with(false, [this, &req](bool& write_done) {
```
xemul: For short bodies (those without body writer function set) this can be avoided

src/http/client.cc
```diff
@@ -151,10 +151,26 @@ future<connection::reply_ptr> connection::do_make_request(const request& req) {
...
+                            write_done = true;
+                        });
+                    }),
+                    recv_reply().then([this, &write_done](reply_ptr rep) {
+                        if (!write_done) {
+                            _fd.shutdown_output();
+                        }
+                        return make_ready_future<reply_ptr>(std::move(rep));
+                    })
+                ).then([](auto joined) {
```
xemul: Good fit for .then_unpack() sugar

xemul:
Test that reproduces the issue is needed badly

PR #3412, opened 2026-05-19
github: revamp test matrix (allpairs, containers, arm)

travisdowns:
## Thrust

- **Generate the build matrix in an all-pairs way** — the actual matrix is getting too big to enumerate fully; switch to pairwise coverage so every (parameter, value) pair is exercised by at least one job without the cartesian-product blow-up.
- **Test the last two major releases of each compiler** — switch to clang 21/22 and gcc 15/16 on Ubuntu 26.04, matching the support window in `compatibility.md`. `install-build-env.sh` derives the apt package, CC/CPP, and clang-tools version from `COMPILER=<family>-<N>`.
- **Include arm64 in the build matrix** — `ubuntu-24.04-arm` runners now cover the arm path alongside x86.

## How it's wired

- `.github/workflows/gen-matrix.py` — uv-shebang Python script (deps declared via PEP 723 inline metadata). Uses `allpairspy` to compute the regular_test rows; emits the explicit `include:` block in `tests.yaml` between AUTOGENERATED markers; everything outside the markers is left untouched. Run it locally to regenerate.
- `.github/workflows/install-build-env.sh` — single post-checkout step that calls `install-dependencies.sh`, layers the compiler packages, and runs `configure.py`. Driven by env vars (COMPILER, STANDARD, MODE, ARCH, ENABLES, OPTIONS, ENABLE_CCACHE). Locally runnable via `COMPILER=clang++-22 STANDARD=23 ... ./install-build-env.sh`.
- `.github/workflows/check-matrix.sh` + `tests-matrix.gold` — `act -n` diffs the resolved job list against a checked-in goldfile so a refactor that doesn't change the resolved jobs is verifiable without pushing.
- `ubuntu:26.04` is used as the container image so the toolchain version is independent of GHA's stock runner. The arm runner label is `ubuntu-24.04-arm`; the same `ubuntu:26.04` image works for both archs since it's multi-arch.
- `cmake/SeastarDependencies.cmake` adds a configure-time `FATAL_ERROR` if clang >= 20 ends up paired with fmt < 11, so the cryptic `fmt/chrono.h` build failure is replaced by a clear "re-run with --cook fmt" hint.

## Misc

- Pins all third-party actions to commit SHAs (with the resolved version in a trailing comment) so retagging can't silently swap an action's implementation.
- Bumps `actions/checkout` → v5 and pins `hendrikmuhs/ccache-action` → v1.2.23 to silence the Node 20 deprecation notice.
- `Show parameters` step echoes inputs at the head of every job log so triage doesn't require expanding the matrix definition.


travisdowns:
huh, all arm runs are failing on a few tests

travisdowns:
https://github.com/scylladb/seastar/issues/3413

.github/workflows/alpinelinux.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
```
mykaul: It's a bit more work, but can you move to hash based versions?
Example:
```
uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

avikivity:
> Add support for C++26 and fix C++26-related issues — configure.py --c++-standard 26 works, a few real C++26 ambiguities in the test sources are fixed, and the fmt cooking recipe gains a dev pin for the optional/ranges fix.

Please defer this to a later series, the upgrade to C++26 (and the retiring of C++23) deserve a ceremonial series of its own.

.github/workflows/gen-matrix.py
```diff
@@ -19,7 +19,15 @@
 HERE = pathlib.Path(__file__).parent
 TESTS_YAML = HERE / "tests.yaml"
```
avikivity, line 53: > github: test on clang-19, clang-22, gcc-14, gcc-16

Our promise is last two major release (compatibility.md), can we test those?

avikivity:
Reviewed changelogs of everything up to the C++26 support, seems reasonable.

.github/workflows/gen-matrix.py
travisdowns: yes

.github/workflows/alpinelinux.yaml
travisdowns: yes no problem

avikivity:
> #3413

Please extract the fix into a separate pull request so that those who maintain their own branches can backport it easily.

travisdowns:
> Please extract the fix into a separate pull request so that those who maintain their own branches can backport it easily.

Yes, that was the plan, the commits here were only to validate the fix (since arm runs only here).

Extracted at #3416.

This PR needs be rebased after that merges for the arm tests to pass.

travisdowns:
> Please defer this to a later series, the upgrade to C++26 (and the retiring of C++23) deserve a ceremonial series of its own.

Done, the source fixes and build changes for C++26 are gone. I will re-submit the build fixes (which aren't specific to C++26, per se) in a series of their own and leave pulling the trigger on 26 for later.

.github/workflows/gen-matrix.py
travisdowns: Done — switched to clang-21/22 and gcc-15/16 (last two majors each) and amended into the original commit.

travisdowns:
(pure rebase 9ab709d565146c358016814fcaeb798c8090f80d)

travisdowns:
Force-pushed: dropped `tests-matrix.gold` and `.github/workflows/check-matrix.sh` from the branch (and their mentions in commit messages) — I don't think they need to be checked in.

travisdowns:
Feedback addressed. We only test gcc 15, 16, clang 21 and 22 now.

travisdowns:
The failed tests are this compiler bug: https://github.com/scylladb/seastar/pull/3394 https://gcc.gnu.org/bugzilla/show_bug.cgi?id=124584

I will exclude gcc/debug from the configurations for now.

travisdowns:
Excluded the (g++, debug) combos from the regular matrix for now — see tip commit bb9e22072.

travisdowns:
I forgot to add that all this rigamarole about "allpairs" and so on is to keep the nunmber of CI runs under 20, since that's the number that GHA will kick off in parallel. Once you have more than that you effectively double CI time as some run must wait for one in the first round to finish.

.github/workflows/alpinelinux.yaml
travisdowns: (done)

travisdowns:
The sanitize CI failures on gcc 15/16 are a (different than the designated init) gcc bug and will be fixed by: https://github.com/scylladb/seastar/pull/3423

travisdowns:
force push cc9097eb4 was a pure rebase

travisdowns:
Heads-up: the `(clang++-*, sanitize, arm)` jobs are failing on this PR but not because of the matrix changes. They started failing after the rebase onto master picked up 5a8218f58 (cache_line_size bump to 256 on aarch64), which exposes a latent under-alignment bug in `alien::message_queue` array allocation. Tracking it in #3429 with a fix already in hand.

travisdowns:
Cherry-picked the alien message_queue alignment fix (#3430) onto this branch so the arm sanitize jobs go green here. Marking the PR as draft until #3430 lands - I'll drop that commit and undraft once it does.

travisdowns:
remaining failures are https://github.com/scylladb/seastar/issues/3431

travisdowns:
force push 49214cf3a was a rebase onto current master. The draft commit (alien.cc message_queue alignment) was dropped since master now has the equivalent fix. This should be ready for review/merge now (fingers crossed).

travisdowns:
@scylladb/seastar-maint this one is green, so ready for re-review, merge, whatever

.github/workflows/install-build-env.sh
```diff
@@ -0,0 +1,97 @@
...
+./install-dependencies.sh
+
+packages=()
+case "$COMPILER" in
+    clang++-*)
+        version="${COMPILER#clang++-}"
+        # libstdc++-16-dev: libboost-all-dev pulls gcc-16-base but not
+        # libstdc++-16-dev; clang prefers the highest gcc dir, so
+        # without this `ld: cannot find -lstdc++` aborts the link.
+        packages+=("clang-${version}" libstdc++-16-dev)
```
tchaikov, line 61: not sure if it's overkill, but we could derive the package name with something like:
```bash
gcc_major=$(gcc -dumpversion | cut -d. -f1)                                                                                                                                                             
packages+=("clang-${version}" "libstdc++-${gcc_major}-dev")
```

travisdowns: This would cause the problem this is trying to solve. The default GCC is 15, but some GCC components from 16 get installed (due to an arcane path involving fortran or something like that).

This triggers a bug in clang/Ubuntu packaging where it sees some GCC 16 stuff and therefore expects stdlib++ for 16 to be installed as well.

tchaikov: i see. gcc-15 takes precedence over gcc-16 because of `update-alternative` machinery. while clang preferes higher versions of libstdc++.

travisdowns: > i see. gcc-15 takes precedence over gcc-16 because of `update-alternative` machinery. while clang preferes higher versions of libstdc++.

Yes, more or less — one correction though: clang doesn't choose based on libstdc++ at all, it picks the highest-versioned GCC *runtime* dir and only afterwards expects libstdc++ to live there. And I don't even know if update-alternative is involved here: every ubuntu has a "default" gcc which is invoked simply as `gcc`, without going through any update-alternative symlink. Then you can install the gcc-N versions alongside. In fact, you can usually install the _same_ gcc-N version as the base version, e.g. you can install gcc-15 on ubuntu despite `gcc` already being version 15: and these are not the same (they usually differ in minor version).

Note that in this case, gcc-16 C/C++ is not installed at all — it's just that `libgcc-16-dev` (pulled in transitively via the fortran chain, e.g. `libboost-all-dev` → `libgfortran-16-dev` → `libgcc-16-dev`) populates the gcc-16 folder with the GCC runtime objects (`crtbegin.o`, `libgcc.a`). clang's GCC-install detector keys on exactly those, so it picks 16 over 15 and then expects a libstdc++ in there. `gcc-16-base` alone doesn't trigger it (no runtime dir is created). It doesn't even look at the default gcc in this case.

travisdowns: Filed this upstream against Ubuntu's clang packaging: https://bugs.launchpad.net/ubuntu/+source/llvm-toolchain-22/+bug/2154746

The precise trigger is `libgcc-16-dev`: it populates `/usr/lib/gcc/x86_64-linux-gnu/16/` with the GCC runtime objects (`crtbegin.o`, `libgcc.a`) that clang's GCC-install detector keys on, so clang selects 16 over 15 and then can't find libstdc++ there → `ld: cannot find -lstdc++`. It gets pulled in transitively via the fortran chain (e.g. `libboost-all-dev` → `libgfortran-16-dev` → `libgcc-16-dev`); `gcc-16-base` on its own creates no runtime dir and does not trigger it.

Minimal repro is just two packages:
```dockerfile
FROM ubuntu:26.04
RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
    apt-get install -y --no-install-recommends clang-22 libgcc-16-dev
RUN echo 'int main(){}' > /t.cpp && clang++-22 /t.cpp -o /t   # fails
```

So hardcoding `libstdc++-16-dev` stays the right workaround here until the clang-side fix lands (llvm/llvm-project#145056, "Take libstdc++ into account during GCC detection" — merged, with the default-flip planned for a future release).

tchaikov: Thank you Travis for the root cause, for offering the detailed explanation and for filing this bug report!

travisdowns: Well thank Claude for that part, but I'll take credit for now as he's off doing somethign else :)

travisdowns:
@avikivity it would be quite handy to have this button enabled in GH:

image

It's on the [settings main page](https://github.com/scylladb/seastar/settings), just scroll down.

It basically just offers an "update this branch" button in the PR you we can click at any time to rebase. Quite handy to rebase + re-kick CI on changes that have been open for a while (e..g, prior to merge), especially ones like this which tend to catch a lot of stuff.

travisdowns:
Force push d72a727f6d34ffc410d8940a3df223288ec43d41 was a pure rebase onto master tip (no content changes).

avikivity:
> @avikivity it would be quite handy to have this button enabled in GH:
> 
> image
> It's on the [settings main page](https://github.com/scylladb/seastar/settings), just scroll down.
> 
> It basically just offers an "update this branch" button in the PR you we can click at any time to rebase. Quite handy to rebase + re-kick CI on changes that have been open for a while (e..g, prior to merge), especially ones like this which tend to catch a lot of stuff.

Done, though I'm too old-fashioned to use it myself.

.github/workflows/gen-matrix.py
```diff
@@ -0,0 +1,110 @@
...
+        "compiler": "clang++",
+        "standard": 23,
+        "mode": "fuzz",
+        "test-args": "-- -R 'Seastar.fuzz.'",
+        "label": "Fuzz Tests",
+    },
+]
+
+# Characters that require a value to be double-quoted in YAML flow style.
+_NEEDS_QUOTE: frozenset[str] = frozenset(' ,:{}"[]#')
```
avikivity: Wouldn't it be nicer to have a yaml library dump the yaml?

.github/workflows/gen-matrix.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python3
+"""
```
avikivity, line 28: License/copyright blurb please

.github/workflows/install-build-env.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+# Install the seastar build toolchain and run ./configure.py inside the
```
avikivity, line 21: Copyright/license.

avikivity:
Looks good!

.github/workflows/gen-matrix.py
travisdowns: Yes, that's how I originally planned to do it, but this was shorter. Happy to swap it. It's not just a totally simple dump the yaml operation because we want to _patch_ the YAML keeping existing comments/formatting in the rest.

avikivity: You could keep the skeleton in this file. But no need to overengineer every little thing (okay there is a need, but we can resist it with a combined effort)

travisdowns: > But no need to overengineer every little thing

Not sure which of the various approaches that points to, but in any case, I used ruamel to patch & dump the file now. Leads to somewhat simpler code, and you can still edit the rest of the file as usual in your editor.

.github/workflows/gen-matrix.py
travisdowns: Done - added the Apache license + copyright header.

.github/workflows/install-build-env.sh
travisdowns: Done - added the Apache license + copyright header.

travisdowns:
Pushed updates addressing review feedback:

- Added the Apache license + copyright header to `gen-matrix.py` and `install-build-env.sh`.
- `gen-matrix.py` now emits the matrix via `ruamel.yaml` (load the workflow → replace the strategy matrix `include` list by key path → dump) instead of hand-rolled string formatting. `tests.yaml` regenerated: the YAML library normalizes some scalar quoting, otherwise unchanged.

Folded into the relevant existing commits.

travisdowns:
Force push 7c4cf9549 was a pure rebase on master (range-diff confirms all 11 commits unchanged).

travisdowns:
@scylladb/seastar-maint this bad boy feels ready to go in, all feedback addressed

LMK if anything is outstanding

.github/workflows/gen-matrix.py
avikivity: I meant that the hack with escaping the yaml chars is fine. But I'll look at the new thing.

Issue #3411, opened 2026-05-19
C++26 build: fmt/optional vs fmt/ranges ambiguity when libstdc++-16 makes std::optional a range


travisdowns:
## Symptom

Building seastar with `-std=c++26` on clang-20+/gcc-16 against libstdc++-16 (Ubuntu 26.04) fails with the cooked fmt 11.2.0 (and also with the recently released fmt 12.0.0 / 12.1.0). Affected TU in our matrix is `tests/unit/sstring_test.cc`:

```
fmt/base.h:2262:45: error: implicit instantiation of undefined template
  'fmt::detail::type_is_unformattable_for<std::optional<seastar::basic_sstring<char, unsigned int, 15>>, char>'

sstring_test.cc:322:
  std::ignore = fmt::format("{}", std::optional(sstring{"hello"}));
```

A minimal local probe shows the real issue is ambiguous partial specializations:

```
error: ambiguous partial specializations of
  'formatter<std::optional<seastar::basic_sstring<char, unsigned int, 15>>>'
  fmt/ranges.h:491: partial specialization matches [with R = std::optional<...>, Char = char]
  fmt/std.h:217:   partial specialization matches [with T = ..., Char = char]
```

`is_formattable<std::optional<sstring>>` therefore evaluates `false`, which is what surfaces as the cryptic `type_is_unformattable_for` error at the call site.

## Root cause

libstdc++-16 implements [P3168R2 (`std::optional` range support)](https://wg21.link/P3168R2) for C++26, so `std::optional<T>` now models `std::ranges::range`. fmt 11.2.0 / 12.0.0 / 12.1.0 have two unconstrained formatter partial specializations that both match `std::optional<T>` once it's a range:

1. `fmt/std.h` — formatter for `std::optional<T>` gated only on `is_formattable<T>` (correct).
2. `fmt/ranges.h` — generic range formatter that matches any type whose `range_format_kind` is non-disabled, including the new optional-as-range.

This is fmtlib/fmt#4760, fixed by fmtlib/fmt#4761 (merged 2026-05-03, master commit `9cb8c0f92b4c345fb974a75d71370c23047528aa`). No tagged fmt release carries the fix yet (12.1.0 was 2025-10-29).

## Reproduction

Run inside an `ubuntu:26.04` container:

```
apt-get update
apt-get install -y --no-install-recommends g++ libfmt-dev clang-22 libstdc++-16-dev
cat > /tmp/t.cc <<EOF
#include <fmt/format.h>
[...]

PR #3418, opened 2026-05-20
build: add fmt-12-1-dev cooking ingredient

travisdowns:
Adds a second fmt cooking ingredient (`fmt-12-1-dev`) pinned to a post-12.1.0 master snapshot. That fmt version is needed to work around [fmtlib/fmt#4760](https://github.com/fmtlib/fmt/issues/4760) — fmt 11.2.0 / 12.0.0 / 12.1.0 have two unconstrained formatter partial specializations that both match `std::optional<T>` once libstdc++-16 makes `std::optional` model `std::ranges::range` (C++26 P3168R2). The fix landed as [fmtlib/fmt#4761](https://github.com/fmtlib/fmt/pull/4761) but no tagged fmt release ships it yet.

The default `fmt` ingredient stays at 11.2.0. Consumers that hit the bug can opt in with `--cook fmt-12-1-dev`. Drop the new ingredient once a tagged fmt release includes the fix.

Tracked at #3411.

### DIFF
--- a/cooking_recipe.cmake
+++ b/cooking_recipe.cmake
@@ -315,6 +315,20 @@ cooking_ingredient (fmt
     -DFMT_DOC=OFF
     -DFMT_TEST=OFF)
 
+# Post-12.1.0 master snapshot, picked for fmt PR #4761 which fixes the
+# fmt/std.h vs fmt/ranges.h optional ambiguity that surfaces when
+# libstdc++-16 makes std::optional model std::ranges::range (C++26
+# P3168R2). Use via `--cook fmt-12-1-dev` instead of `--cook fmt`.
+# Drop once a tagged fmt release carries the fix.
+# Tracked at https://github.com/scylladb/seastar/issues/3411
+cooking_ingredient (fmt-12-1-dev
+  EXTERNAL_PROJECT_ARGS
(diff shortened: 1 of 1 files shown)


travisdowns:

avikivity:
We should consider migrating to std::format. I don't know if there are substantial differences in expressiveness. In any case a significant transition period is needed.

travisdowns:
> We should consider migrating to std::format. I don't know if there are substantial differences in expressiveness. In any case a significant transition period is needed.

Yes. I'm not sure about it. At least the initial version of std::format is missing a bunch of stuff. Also both stdlibs apparently re-implemented it from scratch and quality varies. OTOH moving to std would avoid stuff like this.

My view would be to stick with fmt for now.

avikivity:
> > We should consider migrating to std::format. I don't know if there are substantial differences in expressiveness. In any case a significant transition period is needed.
> 
> Yes. I'm not sure about it. At least the initial version of std::format is missing a bunch of stuff. Also both stdlibs apparently re-implemented it from scratch and quality varies. OTOH moving to std would avoid stuff like this.
> 
> My view would be to stick with fmt for now.

Agree, let it mature.

Issue #3415, opened 2026-05-20
Cooked Boost 1.81 fails to build with clang on aarch64 Linux ('cstddef' not found)


travisdowns:
# Cooked Boost 1.81 fails to build with clang on aarch64 Linux: `'cstddef' file not found`

## Repro

On aarch64 Linux with clang (tested with `clang++-20`, host glibc, GCC 11 toolchain installed):

```
./configure.py --cook fmt --cook GnuTLS --cook Boost \
    --compiler clang++-20 --c++-standard=20
ninja -C build/dev
```

Every Boost source file fails compilation with:

```
./boost/config/detail/select_stdlib_config.hpp:26:14: fatal error: 'cstddef' file not found
   26 | #    include <cstddef>
```

Sample command line emitted by B2:

```
/usr/bin/clang++-20 -c -x c++ ... --target=arm64-pc-linux ... libs/test/src/xml_report_formatter.cpp
```

## Cause

`tools/build/src/tools/clang.jam` in Boost 1.81 (function `init-flags-cross`) builds an explicit `--target=` flag from a shared architecture/vendor map used for both darwin and linux:

- `clang.jam:78`  `case linux : vendor = pc ; sys = linux ;`
- `clang.jam:88`  `case arm-64 : arch = arm64 ;`
- `clang.jam:96`  `: "--target=$(arch)-$(vendor-sys)"`

For Apple this yields the valid `arm64-apple-darwin`. For Linux on aarch64 it yields `arm64-pc-linux`, which clang doesn't recognise as a known triple. With no recognised triple, clang locates no GCC installation and adds no C++ standard library search paths, so the first `#include <cstddef>` in `select_stdlib_config.hpp` fails.

Clang's own default triple on this host (`aarch64-unknown-linux-gnu`) works fine; the breakage is entirely from the `--target=` B2 passes.

The Seastar recipe doesn't pass `architecture=` or `address-model=` — B2 auto-detects `arm` + `64` and feeds it into this cross-target map.

## Upstream status

Tracked in B2 as [bfgroup/b2#449](https://github.com/bfgroup/b2/issues/449) "Invalid target triple selected with clang 20+" (closed Oct 2025). Symptom in that report is the same `arm64-pc-linux` triple emitted on aarch64 Linux. The visible breakage is recent because clang ≥20 tightened target-triple validation; older clang silently accepted the bogus triple and located headers anyway.

[...]

PR #3419, opened 2026-05-20
build: fail fast when cooking Boost with clang>=20 on aarch64

travisdowns:
## Summary
- Detect the aarch64 + clang>=20 combination when Boost is among the requested cooking ingredients, and abort `cmake` configuration with a clear `FATAL_ERROR` pointing at #3415, rather than letting B2 fail mid-build with the cryptic `'cstddef' file not found`.
- The check only fires when Boost would actually be cooked (it consults `Cooking_INCLUDED_INGREDIENTS` / `Cooking_EXCLUDED_INGREDIENTS`), so users on the same host who skip `--cook Boost` are unaffected.

## Why
Boost 1.81's bundled B2 emits `--target=arm64-pc-linux` for the clang toolset on aarch64 Linux. clang < 20 silently accepted the unknown triple and located the standard library headers via its usual heuristics; clang >= 20 tightened triple validation and rejects it outright, so every Boost compile fails. See #3415 and the upstream B2 ticket [bfgroup/b2#449](https://github.com/bfgroup/b2/issues/449). The follow-up [bfgroup/b2#466](https://github.com/bfgroup/b2/pull/466) only added an opt-in `<triple>` mechanism; the default behavior is still broken on `boostorg/build` master and in every released Boost through 1.91.0.

Failing at configure-time tells the user immediately what's happening and points at the documented workaround (use a distro-provided Boost — Seastar's minimum is 1.79.0 per `cmake/SeastarDependencies.cmake`, and recent Ubuntu/Fedora/Debian ship 1.83+).

Refs #3415.

### DIFF
--- a/cooking_recipe.cmake
+++ b/cooking_recipe.cmake
@@ -154,6 +154,39 @@ file (WRITE "${boost_user_config}"
   " : ${boost_cflags}${boost_cxxflags} <cxxflags>-std=c++${CMAKE_CXX_STANDARD}"
   " ;\n")
 
+# Boost 1.81's bundled B2 emits an invalid `--target=arm64-pc-linux` flag for
+# the clang toolset on aarch64 Linux. clang < 20 silently accepted the unknown
+# triple and the build worked; clang >= 20 rejects it, so every Boost source
+# fails with `'cstddef' file not found`. See
+# https://github.com/scylladb/seastar/issues/3415 and the upstream B2 ticket
+# https://github.com/bfgroup/b2/issues/449. Detect this combination up front
+# rather than letting the cook fail mid-build.
+if (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64)$"
+    AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang"
+    AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 20)
+  set (_seastar_boost_will_be_cooked TRUE)
+  if (Cooking_INCLUDED_INGREDIENTS)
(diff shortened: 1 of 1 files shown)


travisdowns:
@scylladb/seastar-maint

avikivity:
Please add a reference to an open issue on boost, which we (or an interested user) can subscribe to, to update the condition when it is finally fixed.

travisdowns:
> Please add a reference to an open issue on boost, which we (or an interested user) can subscribe to, to update the condition when it is finally fixed.

The way I read https://github.com/bfgroup/b2/issues/449 there aren't going to be more fixes coming in this area from boost so an upstream issue would be fruitless. The "fix" they did apply is https://github.com/bfgroup/b2/pull/466, which requires us to specify the target triple explicitly, which we could do. This only landed in boost 1.91. So we can this check if we update boost cooking boost to 1.91 but I decided against that. Maybe you prefer the 1.91 upgrade?

avikivity:
> > Please add a reference to an open issue on boost, which we (or an interested user) can subscribe to, to update the condition when it is finally fixed.
> 
> The way I read [bfgroup/b2#449](https://github.com/bfgroup/b2/issues/449) there aren't going to be more fixes coming in this area from boost so an upstream issue would be fruitless. The "fix" they did apply is [bfgroup/b2#466](https://github.com/bfgroup/b2/pull/466), which requires us to specify the target triple explicitly, which we could do. This only landed in boost 1.91. So we can this check if we update boost cooking boost to 1.91 but I decided against that. Maybe you prefer the 1.91 upgrade?

I think it's better than failing. I'm not much of a cook though.

travisdowns:
Given the discussion above, putting this on ice for now. I may try cooking the cutting edge version at some point.

Issue #3422, opened 2026-05-22
Build failure with gcc >= 15.2: -Wmismatched-new-delete false positive in generator_test.cc (sanitize mode)


travisdowns:
## Summary

Building `tests/unit/generator_test.cc` in **sanitize** mode (`-Werror` enabled) with **gcc 15.2** fails with a `-Wmismatched-new-delete` diagnostic raised inside libstdc++'s `<generator>` header. The same source compiles cleanly with earlier gcc-15 snapshots (e.g. a pre-15.1 trunk from April 2025) — only the compiler changed.

## Reproducer

Fedora 42 ships gcc `15.2.1-7`, which reliably reproduces:

```
docker run --rm -v /path/to/seastar:/seastar -w /seastar fedora:42 bash -c '
  dnf install -y gcc gcc-c++ cmake ninja-build ragel doxygen \
    boost-devel c-ares-devel fmt-devel gnutls-devel hwloc-devel \
    libpciaccess-devel libtool liburing-devel libxml2-devel \
    lksctp-tools-devel lz4-devel make meson numactl-devel \
    openssl openssl-devel protobuf-compiler protobuf-devel \
    python3 python3-pyelftools python3-pyyaml stow \
    systemtap-sdt-devel valgrind-devel xfsprogs-devel yaml-cpp-devel \
    libasan libubsan libatomic diffutils which &&
  ./configure.py --mode=sanitize &&
  ninja -C build/sanitize tests/unit/CMakeFiles/test_unit_generator.dir/generator_test.cc.o
'
```

This was also seen in CI at https://github.com/scylladb/seastar/actions/runs/26201469110/job/77092176645?pr=3412.

## Error

```
tests/unit/generator_test.cc: In function 'sync_generator<int> sync_fibonacci_sequence(unsigned int)':
tests/unit/generator_test.cc:49:1: error: 'static void std::__gen::_Promise_alloc<void>::operator delete(void*, std::size_t)' called on pointer returned from a mismatched allocation function [-Werror=mismatched-new-delete]
   49 | sync_fibonacci_sequence(unsigned count) {
      | ^~~~~~~~~~~~~~~~~~~~~~~
In static member function 'static void* std::__gen::_Promise_alloc<void>::operator new(std::size_t)',
    inlined from 'sync_generator<int> sync_fibonacci_sequence(unsigned int)' at tests/unit/generator_test.cc:49:1:
[...]

PR #3423, opened 2026-05-22
tests/generator: silence gcc -Wmismatched-new-delete around sync_fibonacci_sequence

travisdowns:
## Summary

- gcc >= 15.2 strengthened `-Wmismatched-new-delete` enough to flag libstdc++'s `std::generator` promise type (`_Promise_alloc`) as a false positive: the class-level `operator new` calls unsized global `::operator new`, the class-level `operator delete` is the sized form, and after inlining gcc reports the pair as mismatched. Pairing unsized new with sized delete is well-formed C++ (sized delete was added in C++14 specifically to coexist with unsized new), so this is a compiler false positive — see [gcc PR middle-end/109224](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=109224).
- In `tests/unit/generator_test.cc` the only `std::generator` coroutine is `sync_fibonacci_sequence`, so suppress `-Wmismatched-new-delete` just around that one function. Guard is `!defined(__clang__)` since clang doesn't have this diagnostic.
- Triggered in sanitize mode under e.g. Fedora 42's gcc 15.2.1-7. Background and reproducer in scylladb/seastar#3422 — this PR is a narrow workaround, not a fix for the underlying gcc bug.

## Test plan

- [x] Reproduce the failure on Fedora 42 (gcc 15.2.1-7) sanitize mode prior to the change
- [x] Verify `ninja -C build/sanitize tests/unit/CMakeFiles/test_unit_generator.dir/generator_test.cc.o` builds clean with the change applied
- [x] Verify `tests/unit/generator_test` links and all test cases pass (`*** No errors detected`)
- [x] CI green on master (though we need to check again on the new matrix)

### DIFF
--- a/tests/unit/generator_test.cc
+++ b/tests/unit/generator_test.cc
@@ -45,6 +45,11 @@ using namespace seastar;
 
 using do_suspend = bool_class<struct do_suspend_tag>;
 
+#if !defined(__clang__)
+// Work around gcc -Wmismatched-new-delete false positive in std::generator; see seastar#3422.
+#  pragma GCC diagnostic push
+#  pragma GCC diagnostic ignored "-Wmismatched-new-delete"
+#endif
 sync_generator<int>
 sync_fibonacci_sequence(unsigned count) {
     auto a = 0, b = 1;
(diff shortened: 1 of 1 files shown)


travisdowns:
@scylladb/seastar-maint please take a look, quick one

blocks #3412

PR #3427, opened 2026-05-24
core, util: add initial RISC-V port

sunyuechi:
- cacheline: 64B
- memory: 2 MiB huge pages
- spinlock: cpu_relax() emits Zihintpause `pause`
- thread: .cfi_undefined ra to stop unwinders at thread bottom
- reactor: read faulting PC from uc_mcontext.__gregs[REG_PC]

Tested on SG2044, GCC 16.1.0 / glibc 2.39:

- release ctest: pass
- sanitize-mode thread_test: pass

### DIFF
--- a/include/seastar/core/cacheline.hh
+++ b/include/seastar/core/cacheline.hh
@@ -36,6 +36,8 @@ constexpr size_t cache_line_size =
     128;
 #elif defined(__aarch64__)
     128; // from Linux, may vary among different microarchitetures?
+#elif defined(__riscv)
+    64;
 #else
 #error "cache_line_size not defined for this architecture"
 #endif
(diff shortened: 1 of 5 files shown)


include/seastar/core/cacheline.hh
```diff
@@ -36,6 +36,8 @@ constexpr size_t cache_line_size =
     128;
 #elif defined(__aarch64__)
     128; // from Linux, may vary among different microarchitetures?
+#elif defined(__riscv)
+    64;
```
avikivity, line 40: In general we should switch to std::hardware_{constructive,destructive}_interference_size, but out of scope for this patch.

tchaikov: created #3428

PR #3430, opened 2026-05-26
core/alien: use aligned new/delete for message_queue array

### LINKED ISSUE #3429: core/alien: message_queue array under-aligned on aarch64 sanitize after cache_line_size bump to 256
On `aarch64` with UBSan, after commit 5a8218f58 ("core: use `std::hardware_destructive_interference_size` for `cache_line_size`") raised `cache_line_size` on aarch64 from 128 to 256, nearly every test on the `(clang++-21, C++23, sanitize, arm)` job now aborts at startup with:

```
src/core/alien.cc:131:26: runtime error: constructor call on misaligned address 0xfd1da73f3680 for type 'alien::message_queue', which requires 256 byte alignment
```

Example failed run (PR #3412 after rebase onto master): https://github.com/scylladb/seastar/actions/runs/26455978811/job/77889952838 - 69 of 94 tests failed, all with the same UBSan diagnostic.

## Root cause

`alien::message_queue` declares `alignas(seastar::cache_line_size)` sub-structs ([include/seastar/core/alien.hh#L60](https://github.com/scylladb/seastar/blob/master/include/seastar/core/alien.hh#L60) and [#L68](https://github.com/scylladb/seastar/blob/master/include/seastar/core/alien.hh#L68)), so `alignof(alien::message_queue) == cache_line_size`.

[src/core/alien.cc#L129](https://github.com/scylladb/seastar/blob/master/src/core/alien.cc#L129) allocates the queue array with the plain non-aligned overload:

```cpp
auto queues = reinterpret_cast<alien::message_queue*>(
    operator new[](sizeof(alien::message_queue) * reactors.size()));
```

That overload only guarantees `__STDCPP_DEFAULT_NEW_ALIGNMENT__` (16 on aarch64), so the result is not guaranteed to be aligned to `alignof(message_queue)`. The matching deleter at line 125 likewise uses the non-aligned `::operator delete[]`.

This was a latent bug: on aarch64 the system allocator happens to return 128-byte-aligned blocks for sufficiently large allocations, which satisfied the prior 128-byte `cache_line_size`. With `cache_line_size = 256` it no longer does (the reported address `0xfd1da73f3680` ends in `0x680`, i.e. 128-aligned, not 256-aligned), and UBSan now catches it.

## Fix

Use the aligned `operator new[]` / `operator delete[]` overloads, matching the existing pattern for `smp_message_queue` in [src/core/reactor.cc#L4720](https://github.com/scylladb/seastar/blob/master/src/core/reactor.cc#L4720) and the corresponding deleter at line 4319:

```cpp
operator new[](sizeof(alien::message_queue) * reactors.size(),
               std::align_val_t(alignof(alien::message_queue)));
...
::operator delete[](qs, std::align_val_t(alignof(alien::message_queue)));
```

Note the comment already in `reactor.cc` ("we should enforce calling out to aligned_alloc, instead of pure malloc") flagged this class of bug. The same audit should sweep any other manual `operator new`/`new[]` sites for over-aligned types.

I'll send a PR with the fix shortly.

travisdowns:
`alien::message_queue` has `alignas(cache_line_size)` members but the
queue array was allocated with the plain `operator new[]` overload,
which only guarantees `__STDCPP_DEFAULT_NEW_ALIGNMENT__`. This was
latent until 5a8218f58 raised `cache_line_size` on aarch64 to 256,
after which UBSan aborts at startup on nearly every sanitize-arm test:

```
src/core/alien.cc:131:26: runtime error: constructor call on misaligned
address ... for type 'alien::message_queue', which requires 256 byte alignment
```

Switch to the `std::align_val_t` overloads, mirroring the existing
`smp_message_queue` pattern in `reactor.cc`.

Fixes #3429.

### DIFF
--- a/src/core/alien.cc
+++ b/src/core/alien.cc
@@ -118,15 +118,19 @@ void message_queue::start() {
 }
 
 
+static constexpr std::align_val_t message_queue_alignment{alignof(alien::message_queue)};
+
 void internal::qs_deleter::operator()(alien::message_queue* qs) const {
     for (unsigned i = 0; i < count; i++) {
         qs[i].~message_queue();
     }
-    ::operator delete[](qs);
+    ::operator delete[](qs, message_queue_alignment);
 }
 
 instance::qs instance::create_qs(const std::vector<reactor*>& reactors) {
-    auto queues = reinterpret_cast<alien::message_queue*>(operator new[] (sizeof(alien::message_queue) * reactors.size()));
+    auto queues = reinterpret_cast<alien::message_queue*>(
+            operator new[](sizeof(alien::message_queue) * reactors.size(),
+                           message_queue_alignment));
     for (unsigned i = 0; i < reactors.size(); i++) {
         new (&queues[i]) alien::message_queue(reactors[i]);
     }


travisdowns:
@tchaikov this broke due to https://github.com/scylladb/seastar/pull/3428

not sure why I can't add you as a reviewer

travisdowns:
@scylladb/seastar-maint please consider, a small one

blocks: https://github.com/scylladb/seastar/pull/3412

avikivity (approved):

tchaikov:
> @tchaikov this broke due to #3428

Sorry for the regression. should have audited all places using `cache_line_size` =(

> 
> not sure why I can't add you as a reviewer

guess i don't have enough repo permissions -- a user needs at least the "Triage" permission to be selected as "Reviewer", see also https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#permissions-for-each-role

travisdowns:
> Sorry for the regression. should have audited all places using `cache_line_size` =(

Well I think we can mostly blame CI for not having Arm runs but hope to fix that soon.

travisdowns:
> guess i don't have enough repo permissions -- a user needs at least the "Triage" permission to be selected as "Reviewer", see also https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#permissions-for-each-role

That's weird, I feel you can add any person on GH as a reviewer,  they show up in a separate category "Other reviewers" if they don't have review perms, but :shrug: .

tchaikov:
> > Sorry for the regression. should have audited all places using `cache_line_size` =(
> 
> Well I think we can mostly blame CI for not having Arm runs but hope to fix that soon.

@travisdowns good point! probably we can take the opportunity of revamping the test matrix to run some combinations on [ubuntu-24.04-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md)? see also https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories 

or, i can help on this. once #3412 lands.

avikivity:
> > @tchaikov this broke due to #3428
> 
> Sorry for the regression. should have audited all places using `cache_line_size` =(
> 
> > not sure why I can't add you as a reviewer
> 
> guess i don't have enough repo permissions -- a user needs at least the "Triage" permission to be selected as "Reviewer", see also https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#permissions-for-each-role

I can give you Triage if you want.

travisdowns:
> or, i can help on this. once https://github.com/scylladb/seastar/pull/3412 lands.

It already includes Arm! That's how it turned up these bugs (and a couple others).

PR #3432, opened 2026-05-26
coroutine: disable g++ coroutine destroy-in-suspend bug workaround for fixed versions

avikivity:
A g++ bug [1] involving destruction during await_suspend() was
fixed for g++ 15.3/16, leaving 15.2 as the only vulenerable version.

Tighten the version check to include only g++ 15.2.
[1] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=121961

### DIFF
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -51,7 +51,7 @@ namespace internal {
 inline
 void
 execute_involving_handle_destruction_in_await_suspend(std::invocable<> auto&& func) noexcept {
-#if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ > 15 || (__GNUC__ == 15 && __GNUC_MINOR__ >= 2))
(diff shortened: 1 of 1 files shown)


avikivity:
v2: s/libstdc/g/g, it was a compiler bug, not library.

include/seastar/core/coroutine.hh
```diff
@@ -51,7 +51,7 @@ namespace internal {
 inline
 void
 execute_involving_handle_destruction_in_await_suspend(std::invocable<> auto&& func) noexcept {
-#if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ > 15 || (__GNUC__ == 15 && __GNUC_MINOR__ >= 2))
+#if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ == 15 && __GNUC_MINOR__ == 2)
```
travisdowns, line 54: Looks like time for a IS_GCC_VERSION macro, but there's always tomorrow...

travisdowns (approved):

PR #3433, opened 2026-05-27
rpc: use fmt::format for remote exception reporting

gusev-p:
The [previous approach](https://github.com/scylladb/seastar/pull/3305) (pretty_type_name + ": " + what()) fails to convey useful information for nested exceptions:
`seastar::nested_exception::what()` just returns the static string "seastar::nested_exception", discarding all inner/outer context.

Switch to `fmt::format("{}", std::current_exception())` which delegates to seastar's `operator<<(ostream&, const exception_ptr&)`. That function knows how to unpack `seastar::nested_exception`, `std::nested_exception`, `std::system_error` and regular `std::exception`, producing a complete human-readable description in every case.

As a side-effect, catch(...) instead of catch(std::exception&) so non-std::exception types also get reported rather than propagating unhandled.


gleb-cloudius (approved):

gusev-p:
@scylladb/seastar-maint please review/merge

travisdowns (approved):

PR #3437, opened 2026-05-30
build: also detect GCC's -Wno-error=cpp for #warning

sunyuechi:
The existing check only tries clang's "-Wno-error=#warnings" spelling, which GCC rejects, so an advisory #warning (e.g. from core/dpdk_rte.hh when DPDK has atomic mbuf refcnt) becomes fatal under -Werror.


avikivity:
Lost in c79e9dec76186f76d1a2e48187e3073449a60bfe

PR #3438, opened 2026-05-31
tests: avoid printing std::vector in BOOST_CHECK_EQUAL

sunyuechi:
With -DSeastar_DEPRECATED_OSTREAM_FORMATTERS=OFF, std::vector has no
  operator<<, so BOOST_CHECK_EQUAL on a vector fails to build. Mark the
  types with BOOST_TEST_DONT_PRINT_LOG_VALUE.

### DIFF
--- a/apps/memcached/tests/test_ascii_parser.cc
+++ b/apps/memcached/tests/test_ascii_parser.cc
@@ -27,6 +27,8 @@
 #include "ascii.hh"
 #include <seastar/core/loop.hh>
 
+BOOST_TEST_DONT_PRINT_LOG_VALUE(std::vector<seastar::sstring>)
+
 using namespace seastar;
 using namespace net;
 using namespace memcache;
--- a/tests/unit/simple_stream_test.cc
+++ b/tests/unit/simple_stream_test.cc
@@ -24,6 +24,8 @@
 #include <boost/test/unit_test.hpp>
 #include <seastar/core/simple-stream.hh>
 
+BOOST_TEST_DONT_PRINT_LOG_VALUE(std::vector<char>)
+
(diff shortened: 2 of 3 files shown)


avikivity:
What's boost's problem with printing a vector? I've never seen it.

We have

```
include/seastar/core/sstring.hh:std::ostream& operator<<(std::ostream& os, const std::vector<T>& v) {
```

(not that it's good practice, and we shouldn't have it for a non-seastar type, but I don't understand the failure).

sunyuechi:
```
boost/test/tools/detail/print_helper.hpp:53:13: error: static assertion failed:
    Type has to implement operator<< to be printable
    BOOST_STATIC_ASSERT_MSG( (boost::has_left_shift<std::ostream,T>::value), ... )
  • 'boost::integral_constant<bool, false>::value' evaluates to false
...:55:18: error: no match for 'operator<<' (operand types are 'std::ostream'
    and 'const std::vector<seastar::basic_sstring<char, unsigned int, 15> >')
```

I'm building with -DSeastar_DEPRECATED_OSTREAM_FORMATTERS=OFF (to drop the deprecated).

```
#ifdef SEASTAR_DEPRECATED_OSTREAM_FORMATTERS
std::ostream& operator<<(std::ostream& os, const std::vector<T>& v) {
#endif
```

avikivity:
Ah. It should really default to OFF.

avikivity:
So please update the changelog. It's not boost and not libstdc++3 16, it's that flag.

I'll post a patch switching it to OFF. We should have really made OFF the default.

sunyuechi:
Ok, updated.

PR #3439, opened 2026-05-31
build: default Seastar_DEPRECATED_OSTREAM_FORMATTERS to OFF

avikivity:
In order for us to be able to eventually drop the deprecated functions, we must alert users that they are deprecated by defaulting to OFF.

Since some tests compare std::vectors which don't have std::ostream formatters, add a wrapper
to funnel printing to a fmt::formatter if available, and fall back to std::ostream if not.

### DIFF
--- a/tests/unit/coroutines_test.cc
+++ b/tests/unit/coroutines_test.cc
@@ -24,6 +24,8 @@
 #include <ranges>
 #include <any>
 #include <expected>
+#include <fmt/ranges.h>
+#include "test_comparisons.hh"
 
 #include <seastar/core/circular_buffer.hh>
 #include <seastar/core/coroutine.hh>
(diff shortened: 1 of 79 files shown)


avikivity:
v2: convert all tests to prefer fmt to std::ostream

tests/unit/coroutines_test.cc
```diff
@@ -1015,13 +1016,13 @@ future<std::vector<int> &> co_return_reference() {
 SEASTAR_TEST_CASE(test_co_return_conversions) {
     std::ignore = co_await i2e();
     std::ignore = co_await e2i();
-    BOOST_REQUIRE_EQUAL(co_await co_return_vector(0), (std::vector<std::string>{}));
+    SEASTAR_BOOST_REQUIRE_EQUAL(co_await co_return_vector(0), (std::vector<std::string>{}));
```
tchaikov, line 1020: we'd need to `#include <fmt/ranges.h>` to format range-alike types like `std::vector<std::string>`. see https://godbolt.org/z/5n3KqTjj8 for a reproducer.

avikivity: Somehow it builds locally, I'll need to reproduce the CI environment.

avikivity:
v3:L sprinkled #include \<fmt/ranges.h> where needed

avikivity:
v4: special-case const char* to compare by string value rather than pointer address (following boost.test)

avikivity:
Obsoleted by me own 8a8019b8bd14e1292631c72ad93f0dd4c29930a7

PR #3440, opened 2026-05-31
quic: add experimental ngtcp2 transport

pkorczlab:
Add an experimental QUIC transport backed by ngtcp2 and GnuTLS.

The series adds a bundled ngtcp2 build, exposes an ngtcp2::ngtcp2
CMake target, and initializes the ngtcp2 submodule in CI. The DPDK
submodule checkout remains limited to DPDK-enabled builds.

The new API lives under seastar::quic::experimental and provides
client and server endpoints, connections, bidirectional and
unidirectional streams, transport configuration, QUIC errors, and
conversion of bidirectional streams to connected_socket.

Each connection owns its ngtcp2 and TLS state and drives it from a
single async loop that handles received packets, timer expiry, stream
commands, and connection shutdown.

Also add QUIC client/server demos and unit coverage for the public API,
stream lifecycle, flow control, blocked sends, blocked stream opens,
packet handling, close paths, error classification, and socket address
conversion.

### DIFF
--- a/.github/workflows/alpinelinux.yaml
+++ b/.github/workflows/alpinelinux.yaml
@@ -15,8 +15,18 @@ jobs:
       image: alpine:latest
 
     steps:
+    - name: Install git for checkout
+      run: |
+        apk update
+        apk add --no-cache git
+
     - name: Checkout repository
       uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+
+    - name: Checkout ngtcp2 submodule
+      run: |
+        git config --global --add safe.directory "$GITHUB_WORKSPACE"
+        git submodule update --init ngtcp2
 
     - name: Install build dependencies
       run: |
--- a/.github/workflows/alpinelinux.yaml
+++ b/.github/workflows/alpinelinux.yaml
@@ -23,11 +23,6 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
 
-    - name: Checkout ngtcp2 submodule
-      run: |
-        git config --global --add safe.directory "$GITHUB_WORKSPACE"
-        git submodule update --init ngtcp2
-
     - name: Install build dependencies
       run: |
         apk update
@@ -49,6 +44,7 @@ jobs:
           lksctp-tools-dev   \
           lz4-dev            \
           numactl-dev        \
+          ngtcp2-dev         \
           openssl            \
           openssl-dev        \
           protobuf-dev       \
@@ -65,7 +61,8 @@ jobs:
         cmake -B build -G Ninja            \
          -DCMAKE_BUILD_TYPE=RelWithDebInfo \
          -DSeastar_LTTNG=OFF               \
-         -DSeastar_DOCS=OFF
+         -DSeastar_DOCS=OFF                \
+         -DSeastar_NGTCP2_PROVIDER=SYSTEM
 
     - name: Build Seastar
       run: |
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,7 @@
 [submodule "dpdk"]
 	path = dpdk
 	url = ../dpdk
+
+[submodule "ngtcp2"]
+	path = ngtcp2
+	url = https://github.com/scylladb-zpp-2025-seastar-quic/ngtcp2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -280,6 +280,12 @@ option (Seastar_DPDK
   "Enable DPDK support."
   OFF)
 
+set (Seastar_NGTCP2_PROVIDER
+  "AUTO"
+  CACHE
+  STRING
+  "Select the ngtcp2 provider: AUTO, SYSTEM, or BUNDLED.")
+
 option (Seastar_EXCLUDE_APPS_FROM_ALL
   "When enabled alongside Seastar_APPS, do not build applications by default."
   OFF)
@@ -426,7 +432,7 @@ find_package (ragel 6.10 REQUIRED)
 find_package (Threads REQUIRED)
 find_package (PthreadSetName REQUIRED)
 find_package (Valgrind REQUIRED)
-find_package (ngtcp2 REQUIRED)
+find_package (ngtcp2 1.14.0 REQUIRED)
 
 cmake_dependent_option (Seastar_LOGGER_COMPILE_TIME_FMT
   "Enable the compile-time {fmt} check when formatting logging messages" ON
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -426,6 +426,7 @@ find_package (ragel 6.10 REQUIRED)
 find_package (Threads REQUIRED)
 find_package (PthreadSetName REQUIRED)
 find_package (Valgrind REQUIRED)
+find_package (ngtcp2 REQUIRED)
 
 cmake_dependent_option (Seastar_LOGGER_COMPILE_TIME_FMT
   "Enable the compile-time {fmt} check when formatting logging messages" ON
@@ -719,6 +720,10 @@ add_library (seastar
   include/seastar/websocket/client.hh
   include/seastar/websocket/common.hh
   include/seastar/websocket/server.hh
+  include/seastar/quic/quic.hh
+  include/seastar/quic/quic_client.hh
+  include/seastar/quic/quic_server.hh
+  include/seastar/quic/quic_error.hh
   src/core/alien.cc
   src/core/file.cc
   src/core/fair_queue.cc
@@ -815,6 +820,12 @@ add_library (seastar
   src/websocket/common.cc
   src/websocket/client.cc
   src/websocket/server.cc
+  src/quic/quic_command_runtime.cc
+  src/quic/quic_common.cc
+  src/quic/quic_client.cc
+  src/quic/quic_connection_state.cc
+  src/quic/quic_server.cc
+  src/quic/quic_error.cc
   )
 
 if (Seastar_GNUTLS)
@@ -922,6 +933,8 @@ target_link_libraries (seastar
     lz4::lz4
   PRIVATE
     ${CMAKE_DL_LIBS}
+    ngtcp2::ngtcp2
+    GnuTLS::gnutls
     StdAtomic::atomic
     lksctp-tools::lksctp-tools
     protobuf::libprotobuf
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -825,6 +825,7 @@ add_library (seastar
   src/quic/quic_client.cc
   src/quic/quic_connection_state.cc
   src/quic/quic_server.cc
+  src/quic/sharded_quic_server.cc
   src/quic/quic_error.cc
   )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -723,6 +723,7 @@ add_library (seastar
   include/seastar/quic/quic.hh
   include/seastar/quic/quic_client.hh
   include/seastar/quic/quic_server.hh
+  include/seastar/quic/sharded_quic_server.hh
   include/seastar/quic/quic_error.hh
   src/core/alien.cc
   src/core/file.cc
--- a/cooking.sh
+++ b/cooking.sh
@@ -265,6 +265,10 @@ cat <<'EOF' > "${build_dir}/Cooking.cmake"
 #
 
 macro (project name)
+  if (Cooking_USE_CMAKE_PROJECT_COMMAND)
+    _project (${name} ${ARGN})
+  endif ()
+
   set (_cooking_dir ${CMAKE_CURRENT_BINARY_DIR}/_cooking)
 
   if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
--- a/demos/quic_server_demo.cc
+++ b/demos/quic_server_demo.cc
@@ -31,7 +31,8 @@
 #include <seastar/core/coroutine.hh>
 #include <seastar/core/gate.hh>
 #include <seastar/core/reactor.hh>
-#include <seastar/quic/quic_server.hh>
+#include <seastar/core/smp.hh>
+#include <seastar/quic/sharded_quic_server.hh>
 
 #include "../apps/lib/stop_signal.hh"
 
@@ -55,13 +56,6 @@ static socket_address parse_ip_address(const std::string& ip, uint16_t port) {
     }
 
     throw std::runtime_error("Invalid IP address: " + ip);
-}
-
-static std::string format_endpoint(const std::string& ip, uint16_t port) {
-    if (ip.find(':') != std::string::npos) {
-        return "[" + ip + "]:" + std::to_string(port);
-    }
-    return ip + ":" + std::to_string(port);
 }
 
 static future<> handle_stream(seastar::quic::experimental::stream quic_stream, bool verbose, uint64_t conn_id, uint64_t stream_no) {
@@ -136,37 +130,6 @@ static future<> handle_session(connection session, bool verbose, uint64_t conn_i
     }
 }
 
-static future<> accept_loop(quic_server& server, gate& sessions, bool verbose) {
-    uint64_t next_conn_id = 1;
-    while (true) {
-        connection session;
-        try {
-            session = co_await server.accept();
-        } catch (const quic_error& e) {
-            if (e.code() == quic_error::closed) {
-                co_return;
-            }
-            throw;
-        }
-        auto conn_id = next_conn_id++;
-        if (verbose) {
-            std::cout << "[server conn#" << conn_id << "] accepted\n";
-            std::cout.flush();
-        }
-
-        (void)with_gate(
-          sessions, [session = std::move(session), verbose, conn_id]() mutable { return handle_session(std::move(session), verbose, conn_id); })
-          .handle_exception([](std::exception_ptr ep) {
-              try {
-                  std::rethrow_exception(ep);
-              } catch (const std::exception& e) {
-                  std::cerr << "[server] connection task failed: " << e.what() << "\n";
-              }
-          })
-          .or_terminate();
-    }
-}
-
 int main(int argc, char** argv) {
     app_template app;
     app.add_options()
@@ -177,9 +140,7 @@ int main(int argc, char** argv) {
       ("verbose,v", bpo::value<bool>()->default_value(false)->implicit_value(true), "Verbose logging");
 
     return app.run(argc, argv, [&app]() -> future<int> {
-        quic_server server;
-        gate sessions;
-        std::optional<future<>> accept_task;
+        sharded_quic_server server;
         std::exception_ptr error;
         bool verbose = false;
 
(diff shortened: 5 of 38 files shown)


.gitmodules
```diff
@@ -1,3 +1,7 @@
 [submodule "dpdk"]
 	path = dpdk
 	url = ../dpdk
+
+[submodule "ngtcp2"]
+	path = ngtcp2
+	url = https://github.com/TusikNusik/ngtcp2
```
ewienik: ngtcp2 uses some generic CMake names (ENABLE_STATIC_LIB, ENABLE_SHARED_LIB, etc...) that can collide with seastar CMake names. For this reason there is open PR ngtcp2/ngtcp2#2008. This url points to the PR branch.

The question for maintainers is how to finally maintain this dependency. We could create a fork of ngtcp2 in scylladb organization and put fork the requested change there.

cooking.sh
```diff
@@ -254,6 +254,10 @@ cat <<'EOF' > "${build_dir}/Cooking.cmake"
 #
 
 macro (project name)
+  if (Cooking_USE_CMAKE_PROJECT_COMMAND)
+    _project (${name} ${ARGN})
+  endif ()
+
```
ewienik, line 271: cmake-cooking overrides CMake's function `project` with its own macro. Ngtcp2's CMakefile doesn't work with this trick so this is a workaround to cmake-cooking. The question is if it is acceptable.

ewienik:
@tchaikov FYI

avikivity:
Please update install-dependencies.sh.

.github/workflows/alpinelinux.yaml
```diff
@@ -15,9 +15,17 @@ jobs:
       image: alpine:latest
```
avikivity, line 16: > Add ngtcp2 as a submodule and expose its static GnuTLS build as the

Why not use packaged ngtcp2?

include/seastar/quic/quic.hh
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2026 ScyllaDB Ltd.
```
avikivity, line 19: I think your own copyright should go there, not sure how the ZPP agreement works.

include/seastar/quic/quic.hh
```diff
@@ -0,0 +1,165 @@
...
+enum class congestion_control_algorithm : uint8_t {
+    reno,
+    cubic,
+    bbr,
+    bbr2,
+};
+
+// Transport-level knobs passed to ngtcp2 when a connection is created.
+struct transport_config {
+    uint64_t max_idle_timeout_ns = 60ULL * 1000 * 1000 * 1000;
```
avikivity: std::chrono::nanoseconds

include/seastar/quic/quic.hh
```diff
@@ -0,0 +1,165 @@
...
+    std::optional<size_t> max_tx_udp_payload_size{};
+    std::optional<uint64_t> max_udp_payload_size{};
+    std::optional<uint64_t> initial_rtt_ns{};
+    std::optional<uint64_t> max_window{};
+    std::optional<uint64_t> max_stream_window{};
+    std::optional<size_t> ack_thresh{};
+    std::optional<congestion_control_algorithm> congestion_control{};
+    bool disable_tx_udp_payload_size_shaping = false;
+    bool disable_pmtud = false;
+};
```
avikivity, line 119: Anything that is part of the public API should have brief documentation. I don't want you to duplicate the ngtcp2 documentation, it's okay to redirect there.

Anything that's not public should be in the internal subnamespace.

avikivity: I see you're already aware of the internal namespace.

include/seastar/quic/quic.hh
```diff
@@ -0,0 +1,165 @@
...
+    class impl;
+    explicit connection(std::unique_ptr<impl> state);
+
+    std::unique_ptr<impl> _impl;
+
+    friend class quic_client;
+    friend class quic_server;
+};
+
+connected_socket to_connected_socket(stream&& s);
```
avikivity, line 249: good!

include/seastar/quic/quic.hh
avikivity: By documentation, I mean something that doxygen recognizes. You can run `ninja doc_api` and see if it gets indexed correctly.

include/seastar/quic/quic.hh
```diff
@@ -0,0 +1,165 @@
...
+    stream& operator=(stream&&) noexcept;
+
+    stream(const stream&) = delete;
+    stream& operator=(const stream&) = delete;
+
+    bool is_open() const noexcept;
+    stream_id id() const noexcept;
+    stream_type type() const noexcept;
+    bool can_read() const noexcept;
+    bool can_write() const noexcept;
```
avikivity, line 173: What does can_read() / can_write() means? Are those readiness notifications or open/close accessors? Need explanations.

include/seastar/quic/quic_error.hh
```diff
@@ -0,0 +1,53 @@
...
+
+#pragma once
+
+#include <cstdint>
+#include <stdexcept>
+#include <string>
+
+namespace seastar::quic::experimental {
+
+class quic_error final : public std::runtime_error {
```
avikivity: maybe create an std::error_category and use std::system_error instead, it fits numbered errors better.

include/seastar/quic/quic_server.hh
```diff
@@ -0,0 +1,63 @@
...
+
+    quic_server(const quic_server&) = delete;
+    quic_server& operator=(const quic_server&) = delete;
+
+    future<> start(quic_server_config config);
+    future<connection> accept();
+    future<> stop();
+
+private:
+    std::shared_ptr<quic_server_impl> _impl;
```
avikivity: Why std::shared_ptr? The implementation should be per shard.

demos/quic_server_demo.cc
```diff
@@ -0,0 +1,246 @@
...
+            auto key = cfg["key"].as<std::string>();
+            verbose = cfg["verbose"].as<bool>();
+
+            quic_server_config server_cfg;
+            server_cfg.listen_address = parse_ip_address(address, port);
+            server_cfg.crt_file = crt;
+            server_cfg.key_file = key;
+
+            co_await server.start(std::move(server_cfg));
+            accept_task.emplace(accept_loop(server, sessions, verbose));
```
avikivity: Where is the distribution across shards?

avikivity:
Looks generally good. But where is the sharding?

I'd like to see something similar to seawreck (or perhaps patch seawreck) to stress the server and report numbers.

CMakeLists.txt
```diff
@@ -919,6 +930,8 @@ target_link_libraries (seastar
     lz4::lz4
   PRIVATE
     ${CMAKE_DL_LIBS}
+    ngtcp2::ngtcp2
+    GnuTLS::gnutls
```
dotnwat, line 945: is this only going to work with GnuTLS?

src/quic/quic_impl.hh
```diff
@@ -0,0 +1,369 @@
...
+    size_t consumed = 0;
+};
+
+struct transport_open_stream_result {
+    int rv = 0;
+    stream_id sid = invalid_stream_id;
+};
+
+// Callback-based view of client/server transport operations shared by common helpers.
+struct connection_transport {
```
tchaikov: have you thought about using a virtual base class here instead of hand-rolling the function-pointer table? the shared functions could be concrete methods, and the owner-specific ones pure virtual. that way there's just one place to add an op instead of the fn-ptr table and the lambda table you have to keep in sync, and the compiler catches a missing one at build time instead of it turning into a null call at runtime. CRTP would work too, but that pushes the shared code into headers, which means slower builds and tighter coupling, so a plain virtual base feels like the nicer fit.

CMakeLists.txt
TusikNusik: Yes, the current integration is GnuTLS-based. ngtcp2 itself supports multiple TLS backends, but this build uses ngtcp2_crypto_gnutls_static, so GnuTLS is required here.

.github/workflows/alpinelinux.yaml
TusikNusik: Ubuntu 24.04 ships libngtcp2-dev version 0.12.1, which uses an older ngtcp2 API that is incompatible with the current QUIC implementation.

For that reason, the ngtcp2 submodule provides a compatible version across supported distributions. Would this be a sufficient justification for keeping it?

avikivity: (checks watch) the current year is 2026

include/seastar/quic/quic.hh
pkorczlab: The timeout and initial RTT fields now use std::chrono::nanoseconds, with chrono-based defaults. This change is in fbcb40d

(https://github.com/scylladb-zpp-2025-seastar-quic/seastar/commit/fbcb40d73475d913a1a313a5f63069e43ab9886f).

include/seastar/quic/quic.hh
pkorczlab: The documentation now states that can_read() and can_write() report whether the local endpoint may use the corresponding stream direction; they are not readiness notifications. This clarification is in fbcb40d

(https://github.com/scylladb-zpp-2025-seastar-quic/seastar/commit/fbcb40d73475d913a1a313a5f63069e43ab9886f).

include/seastar/quic/quic_error.hh
pkorczlab: quic_error now derives from std::system_error, and the implementation provides a dedicated QUIC std::error_category and make_error_code() function. This change is in ae31fa6

(https://github.com/scylladb-zpp-2025-seastar-quic/seastar/commit/ae31fa6fe49ae352d55abca541e722dc97a26116).

include/seastar/quic/quic_server.hh
pkorczlab: Replaced with Seastar lw_shared_ptr in 9910e64

(https://github.com/scylladb-zpp-2025-seastar-quic/seastar/commit/9910e64798d74defab3c79c5de9e9b779a741ff1).

src/quic/quic_impl.hh
pkorczlab: Addressed in 2b8ab77
(https://github.com/scylladb-zpp-2025-seastar-quic/seastar/commit/2b8ab773a0d813eab3ff2acd1d876d73c01dc751).

I replaced the hand-written function-pointer adapter with an abstract connection_transport interface. Client- and server-specific operations are pure virtual, while shared algorithms remain free functions taking connection_transport&.

demos/quic_server_demo.cc
pkorczlab: Unlike TCP, QUIC runs over UDP, so accepting a QUIC connection does not produce a separate connected file descriptor that can be handed off to another shard. The main options are to receive all datagrams on one shard and forward them in userspace, to create one SO_REUSEPORT listener per shard and let the kernel distribute incoming datagrams across the reuseport group, or to use a platform-specific reuseport eBPF classifier that selects a socket from the QUIC connection ID. We implemented the second approach: the shard that receives the Initial packet owns the connection, server-generated connection IDs encode the owner shard, and eligible short-header packets carrying such an ID are forwarded when they arrive on a different shard. This implementation is not included in the current PR - it is available in d10cfea

(https://github.com/scylladb-zpp-2025-seastar-quic/seastar/commit/d10cfea08d2df0ef2e657710fcbafe1f2a2211c7).

avikivity: > to receive all datagrams on one shard and forward them in userspace

no

> to create one SO_REUSEPORT listener per shard and let the kernel distribute incoming datagrams across the reuseport group

yes

> This implementation is not included in the current PR 

Seastar is a multi-core framework and cannot merge a single-shard facility.

ewienik: > > This implementation is not included in the current PR
> 
> Seastar is a multi-core framework and cannot merge a single-shard facility.

IMHO we could provide sharding with SO_REUSEPORT as a follow-up PR as it needs to modify network stack and to not complicate this PR. Is it acceptable as a follow-up?

avikivity: No

src/quic/quic_server.cc
```diff
@@ -0,0 +1,1894 @@
...
+        }
+        rv = gnutls_priority_set_direct(tls, "NORMAL:-VERS-ALL:+VERS-TLS1.3", nullptr);
+        if (rv < 0) {
+            gnutls_deinit(tls);
+            throw_quic_error(classify_gnutls_error(rv), gnutls_error_message(rv));
+        }
+
+        std::vector<gnutls_datum_t> alpns;
+        alpns.reserve(_cfg.alpns.size());
+        for (const auto& alpn : _cfg.alpns) {
```
niekbouman, line 1347: You take `const auto&` here, and two lines below you have to perform a `const_cast`.

Although this seems legal to me, why not take non-const `auto&`, then you do not need the `const_cast`. Or is your current choice deliberate?

src/quic/quic_server.cc
```diff
@@ -0,0 +1,1894 @@
...
+            conn = nullptr;
+        }
+        if (tls) {
+            gnutls_deinit(tls);
+            tls = nullptr;
+        }
+    }
+
+    void fill_path(ngtcp2_path& path) {
+        init_ngtcp2_addr(&path.local, reinterpret_cast<sockaddr*>(&local_ss), local_ss_len);
```
niekbouman, line 284: Is this `reinterpret_cast` allowed in the light of strict aliasing rules and UB? (And the same question for the line below, and for other occurrences of this pattern in this file, or more broadly, in this PR.)

I am aware of this comment from https://man7.org/linux/man-pages/man3/sockaddr.3type.html

>  These structures were invented before modern ISO C strict-aliasing
       rules.  If aliasing rules are applied strictly, these structures
       would be extremely difficult to use without invoking undefined
       behavior.  POSIX.1-2024 fixed this by requiring that
       implementations make sure that these structures can be safely used
       as they were designed.

but I am wondering what C++ says about this (and whether C++ standard respects "POSIX.1-2024", or not at all).

Would a `std::memcpy`, or machinery like C++23's `std::start_lifetime_as` be better here, or would this be overly pedantic?

@avikivity Do you know this?

niekbouman (commented):
Very nice work! Left two questions.

src/quic/quic_server.cc
avikivity: I don't understand start_lifetime_as() well enough to say if it's correct here.

Since init_ngtcp2_addr() is extern, I don't think there is a practical concern.

demos/quic_server_demo.cc
niekbouman: > to create one SO_REUSEPORT listener per shard and let the kernel distribute incoming datagrams across the reuseport group

It would be nice to be able to control on which shard these datagrams land, similar to the `server_socket::load_balancing_algorithm` machinery for Seastar's posix-TCP-stack wrapper, e.g. to be able to also use the `fixed_cpu` strategy, or to have `source_port % N` behavior (where N is the size of the REUSEPORT-group) instead of the Linux kernel's `4-tuple-hash % N` where the 4-tuple is `(src IP, src port, dst IP, dst port)`.

According to Google Gemini, for UDP this can only be achieved via `SO_ATTACH_REUSEPORT_EBPF`, i.e. by attaching a bpf program to the REUSEPORT-group in which the steering logic is implemented. Although attaching such program does not require extra capabilities, loading the bpf program would require CAP_BPF. Gemini suggests a privilege separation approach with a separate BPF loader. (See part of its output in the quote further below)

Other ideas for achieving some form of customisable load balancing: 
- There could be a helper function to choose a source port such that Linux' 4-tuple-hashing method will make the datagrams land on the desired shard. (Disadvantage of controlling shard via source port is that there could be middle boxes that are not transparent w.r.t. source port and destroy the correspondence)
- I'd say one obvious alternative for `load_balancing_algorithm::fixed` would to run an independent QUIC server on each shard, and give each shard its own port. (Useful for example in a setting where you want that two Seastar peers, each having $C$ cores, can interconnect, such that for every shard $i \in \lbrace 0, 1,\ldots,C-1 \rbrace$, shard $i$ connects to the corresponding, remote shard $i$).  
- Another alternative, which is even more complex than the REUSEPORT-eBPF route, would be to let the QUIC implementation use an AF_XDP-based reduced-functionality UDP stack (e.g., without fragmentation, which QUIC doesn't need anyway) in which customizable steering logic is built in, but of course this would also require BPF trickery for AF_XDP.)

Below is Gemini's privilege separation suggestion:

> ## How This Works in Practice (The Delegation Model)
> Because of this separation of concerns, high-performance systems often use a privilege separation model to keep the main application secure.
> 
> Instead of running your Seastar application as root just to route UDP packets, you can design it like this:
> 
> 1. A small, privileged helper daemon (running with CAP_BPF) boots up.
> 
> 1. The helper loads the eBPF program and the map into the kernel, receiving the file descriptors.
> 
> 1. The helper passes those file descriptors to your unprivileged Seastar application via a Unix domain socket (using SCM_RIGHTS ancillary messages).
> 
> 1. The helper daemon exits or drops privileges.
> 
> 1. Your unprivileged Seastar application creates its UDP sockets, populates the eBPF map, and calls setsockopt using the file descriptors it received.

avikivity: Yes, it can come afterwards. Without basic sharding this can't be merged.

More advanced NICs can use ntuple filtering to sort out the packets on the NIC, but they're hard to find.

niekbouman:
Hi @pkorczlab (+ @avikivity )

I've built a benchmark application for the Seastar QUIC implementation using an AI coding tool.
It can be found here: https://github.com/niekbouman/quic_bm

The throughput benchmark shows that QUIC, on my laptop on localhost with the current parameters hard-coded in the benchmark script, attains only one tenth (10%) of the throughput of TCP+TLSv13.
Curious if this agrees with your benchmark tests. Or that some parameters are set incorrectly in my script.

Also, there is a multi-stream benchmark ("Phase 4"), with bidirectional and unidirectonal _use_ of bidirectional QUIC streams. I observe large imbalance between the throughput of the substreams, as if some get starved by others, any idea if I am doing something wrong here?

```
$ ./bench/quic_vs_tls_bench --client-shard 0

[Phase 2] QUIC single-stream results:
  QUIC / A -> B (client -> server)
    sent by peer A (client)                  490.50 MiB in  3.322679 s ->     147.62 MiB/s
    received by peer B (server)              474.47 MiB in  3.320323 s ->     142.90 MiB/s
  QUIC / B -> A (server -> client)
    sent by peer B (server)                  828.75 MiB in  5.001117 s ->     165.71 MiB/s
    received by peer A (client)              812.63 MiB in  5.000452 s ->     162.51 MiB/s
    B->A / A->B                                1.14x

[Phase 3] TLSv1.3-over-TCP single-stream results:
  TLSv1.3 + TCP / A -> B (client -> server)
    sent by peer A (client)                13387.12 MiB in  7.227401 s ->    1852.27 MiB/s
    received by peer B (server)            13383.12 MiB in  7.225869 s ->    1852.11 MiB/s
  TLSv1.3 + TCP / B -> A (server -> client)
    sent by peer B (server)                 9089.12 MiB in  5.000285 s ->    1817.72 MiB/s
    received by peer A (client)             9085.52 MiB in  5.001356 s ->    1816.61 MiB/s
    B->A / A->B                                0.98x

========================================================================
[Phase 4] Multi-stream QUIC saturation (8 streams, 12 data flows)
========================================================================

[Phase 4] Multi-stream QUIC results (single connection, 8 streams):
    wall clock          : 8.685831 s

    stream   role            A->B MiB     A->B MiB/s       B->A MiB     B->A MiB/s
    ---------------------------------------------------------------------------------
    1        A -> B              0.08           0.01           0.00           0.00
    2        A -> B              1.09           0.13           0.00           0.00
    3        B -> A              0.00           0.00         190.07          22.19
    4        B -> A              0.00           0.00         210.08          24.52
    5        A <-> B             4.00           0.47         209.38          24.44
    6        A <-> B             2.00           0.23         203.25          23.72
    7        A <-> B             2.00           0.23         203.25          23.72
    8        A <-> B             2.00           0.23         203.25          23.72
    ---------------------------------------------------------------------------------
    total                       11.17           1.30        1219.28         142.32

    aggregate throughput: 143.62 MiB/s over 1230.45 MiB in 12 flows
```


Could part of the lower throughput for QUIC be caused by lack of use of UDP_SEGMENT / UDP_GRO / sendmmsg techniques?
See also [Willem de Bruijn: LPC 2018 - Optimizing UDP for content delivery: GSO, pacing and zerocopy](https://pdfs.semanticscholar.org/854e/ce39aa38618333c5f77febccc6a0437c7b95.pdf)
and https://blog.cloudflare.com/accelerating-udp-packet-transmission-for-quic/

include/seastar/quic/quic.hh
nikagra: I've clarified with @swasik and Scylla copyright is ok here

PR #3441, opened 2026-06-01
cmake: guard DPDK::dpdk against redefinition in Finddpdk

sunyuechi:
The block guard tests `NOT (TARGET dpdk)`, but the shared-library
branch only creates DPDK::dpdk, never `dpdk`, so a second
find_package(dpdk) redefines DPDK::dpdk and errors out. Guard on
DPDK::dpdk instead.

### DIFF
--- a/cmake/Finddpdk.cmake
+++ b/cmake/Finddpdk.cmake
@@ -151,7 +151,7 @@ endforeach ()
 pkg_check_modules (libarchive_PC QUIET libarchive)
 list(APPEND dpdk_dependencies ${libarchive_PC_LIBRARIES})
 
-if (dpdk_FOUND AND NOT (TARGET dpdk))
+if (dpdk_FOUND AND NOT (TARGET DPDK::dpdk))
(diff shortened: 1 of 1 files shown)


tchaikov (approved):
thanks for the fix. lgtm

tchaikov:
@scylladb/seastar-maint please take a look. This change helps to address the build failure of parent project which happens to define the same CMake target.

avikivity:
> @scylladb/seastar-maint please take a look. This change helps to address the build failure of parent project which happens to define the same CMake target.

cmake is terrible enough even without the parent and child projects competing for the definition of a third-party library.

travisdowns:
> cmake is terrible enough even without the parent and child projects competing for the definition of a third-party library.

Redpanda switched to Bazel (we have a bazel build for seastar, which we patch-in). IMO it's been a big improvement.

avikivity:
> > cmake is terrible enough even without the parent and child projects competing for the definition of a third-party library.
> 
> Redpanda switched to Bazel (we have a bazel build for seastar, which we patch-in). IMO it's been a big improvement.

I understand Bazel has poor support for C++20 modules.

PR #3446, opened 2026-06-01
util: Remove deprecated later() function

xemul:
The function was deprecated in January 2022 in favor of yield() or check_for_io_immediately().


xemul:
upd:
- lost mention in cppm

PR #3449, opened 2026-06-03
Add LTTng-UST tracepoints (and add some IO ones)

xemul:
Introduce LTTng-UST tracepoint infrastructure into Seastar and provide the first set of tracepoints covering the IO flow.

There were attempts before, but they were discarded because all used home-brew tracepoints implementation and custom tools. This PR uses standard LTTng-UST integration which provides:

   - ~3 ns overhead when dormant (single branch, predicted not-taken) *
   - ~100–300 ns per event when recording (userspace ring buffer, no kernel transition) *
   - Attach/detach to running applications without restart
   - SDT probe markers emitted automatically (perf/bpftrace interoperability for free)
   - Structured, typed events with nanosecond timestamps in a standard format (CTF)

(*) Numbers are from CI's general knowledge of LTTng-UST performance characteristics commonly cited in LTTng documentation and conference talks.

This lays the groundwork for instrumenting other Seastar subsystems (scheduling, networking, memory) in the future using the same infrastructure. 

As first inhabitants this PR add tracepoints to the IO request lifecycle: io_queue_queued, io_queue_dispatched, io_queue_completed and io_queue_cancelled. All come with request_id and dev_id, the "queued" one also has direction, offset, size and priority class. Timestamps are provided by LTTng (per-event, nanosecond precision), so queue/execution latencies are derived by correlating req_id across events. Per-shard attribution uses LTTng's procname/vtid context — no shard_id bloat in each event.

LTTng support is enabled by default. To disable it: -DSeastar_LTTNG=OFF (or ./configure.py --disable-lttng). When OFF, tracepoints are compiled out entirely with zero overhead. When ON but not actively tracing, overhead is a single not-taken branch per IO operation.

An integration test (io_trace_test.py) exercises the full pipeline:
   - Configures an lttng session with per-thread context
   - Runs io_tester with distinct IO patterns per shard (reads on shard 0, writes on shard 1)
   - Verifies event types, lifecycle ordering, and per-shard separation via babeltrace2
CI runs all matrix jobs with --enable-lttng to prevent bitrot.

Nearest future work (not in this series):
   - Additional IO hint events (throttle/unthrottle, request splitting)
   - Tracepoints in other subsystems (scheduler, networking)
   - Trace decode/analysis library
   - Replay recorded IO patterns through io_tester


.github/workflows/tests.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       compiler: ${{ matrix.compiler }}
       standard: ${{ matrix.standard }}
       mode: ${{ matrix.mode }}
-      enables: ${{ matrix.enables }}
+      enables: --enable-lttng ${{ matrix.enables }}
```
travisdowns: The matrix is generated in a different way now.

So the goal is to keep it off by default in the build, but on by default for the primary CI matrix, right?

xemul: Yes. I didn't want to disturb the existing projects that are built on top of seastar, but to keep validating it in the CI

avikivity:
> ~3 ns overhead when dormant (single branch, predicted not-taken) *

That is not very impressive (but probably wrong).

avikivity:
cc @michoecho who looked at LTTng and was unimpressed

avikivity:
install-dependencies.sh?

avikivity:
> install-dependencies.sh?

duh, it's there

avikivity:
There are a few dtrace-compatible events (sdt.h, STAP_PROBE). Are they compatible?

We should probably migrate to avoid two tracing systems.

xemul:
> > ~3 ns overhead when dormant (single branch, predicted not-taken) *
> 
> That is not very impressive (but probably wrong).

Added perf test that compares a no-op loop with no-op loop + inactive tracepoint.

      baseline:    0.139 ns/call
      tracepoint:  0.245 ns/call
      overhead:    0.106 ns/call

avikivity:
> > > ~3 ns overhead when dormant (single branch, predicted not-taken) *
> > 
> > 
> > That is not very impressive (but probably wrong).
> 
> Added perf test that compares a no-op loop with no-op loop + inactive tracepoint.
> 
> ```
>   baseline:    0.139 ns/call
>   tracepoint:  0.245 ns/call
>   overhead:    0.106 ns/call
> ```

I'd estimate it at 0 ns/call on modern wide machines for microbenchmarks. The problem is that it adds instruction cache footprint and consumes branch predictor resources. For high-value tracepoints like the I/O scheduler (and the CPU scheduler) it's worth it for sure.

xemul:
> There are a few dtrace-compatible events (sdt.h, STAP_PROBE). Are they compatible?

One can "attach" to SDT probes using lttng tool, but they are different under the hood. SDT probe is a `nop` sequence when inactive, it gets live-patched into a "trap" on activation. Run-time it enters the kernel for processing. LTTng tracepoints are fully user-space, use mov+test+je all the time, but when active the je passes control elsewhere (not into the kernel).

> We should probably migrate to avoid two tracing systems.

I replaced the existing stap probes with lttng ones. The perf tests running shows no difference on my node

avikivity:
> > There are a few dtrace-compatible events (sdt.h, STAP_PROBE). Are they compatible?
> 
> One can "attach" to SDT probes using lttng tool, but they are different under the hood. SDT probe is a `nop` sequence when inactive, 

That's the good part.

> it gets live-patched into a "trap" on activation. Run-time it enters the kernel for processing.

That's the terrible part.

There was also a binary-rewriting mode with --dyninst/stapbin, but I never got it to work.


> LTTng tracepoints are fully user-space, use mov+test+je all the time, but when active the je passes control elsewhere (not into the kernel).
> 
> > We should probably migrate to avoid two tracing systems.
> 
> I replaced the existing stap probes with lttng ones. The perf tests running shows no difference on my node

Yes, the overhead is negligible for microbenchmarks. It's probably negligible for macrobenchmarks too, it's just that the proof that it can be done better exists (the kernel) which annoys me.

michoecho:
> cc @michoecho

Witnessed.

> who looked at LTTng and was unimpressed

I didn't like the overhead. My ideal tracepoint would be a nop when disabled, and when enabled:
```
unconditional jump
rdtsc
load cache line with buffer metadata
conditional branch to rotate_buffers() if no space in current buffer
store tracepoint ID, timestamp and arguments to buffer tail
store new buffer tail position
unconditional jump back to regular logic
```

But this compiles to a conditional branch over a load when disabled, and when enabled, it makes a shared library call which iterates over a list of virtual callbacks which write things into concurrent queues.

A tracepoint per IO op is rare enough, but in my experiments the overhead made it too expensive to have a tracepoint per seastar task, for example, while custom tracepoints were cheap enough.

And then you also have a separate daemon process that has to be taken care of, which also writes potentially a lot of data to the disk or network outside of Seastar's control, etc. And I didn't see anything useful in the "ecosystem" either. It all has more value if you are tracing many processes and/or the kernel together, or have short-lived threads, or emitting tracepoints from signal handlers, etc. But the use cases I had in mind didn't need that; all I cared about was a way to dump some timestamps and integers at given points as cheaply as possible. And this isn't it.

But it's better than nothing. Just don't get married to a specific tracing framework too much.

.github/workflows/tests.yaml
avikivity: I think it's reasonable to enable it by default, if the only overhead when not in use is a not-taken branch.

avikivity:
@xemul let's continue (maybe the ball is on my side)

xemul:
It's partially on my side. 
- my next plan is to add IO record-replay tool and this PR is missing one bit for it
- one more patch to ditch trace-level logs duplicating the tracepoints
- [make it ON by default](https://github.com/scylladb/seastar/pull/3449#discussion_r3473159989)

xemul:
upd:
- keep it ON by default
- replace device with file-descriptor in IO trace events
- remove now redundant io_log.trace() logging

xemul:
upd:
- disable LTTnG for Alpine job

xemul:
@avikivity , now the ball is on your side

bhalevy:
> @avikivity , now the ball is on your side

@avikivity ping, please re-review

src/core/reactor.cc
```diff
@@ -2785,12 +2781,12 @@ bool reactor::task_queue::run_tasks() {
     while (!_q.empty()) {
         auto tsk = _q.front();
         _q.pop_front();
-        STAP_PROBE(seastar, reactor_run_tasks_single_start);
+        SEASTAR_REACTOR_TRACE(task_start, _id, _name.c_str());
```
avikivity, line 2784: Need to remove from install-dependencies.sh

CMakeLists.txt
```diff
@@ -1171,6 +1171,17 @@ if (Seastar_LTTNG)
 
   target_link_libraries (seastar
     PRIVATE LTTng::UST)
+
```
avikivity, line 1167: > This is a bug in lttng-ust itself.

Please find/file a bug report and add a reference here.

avikivity:
>  When OFF (the default), 

Outdated

src/core/reactor.cc
xemul: Removed systemtap dependency in next version

xemul:
upd:
- rebased (no conflicts)
- fixed the cover letter text
- dropped sdt dependency

CMakeLists.txt
xemul: https://bugs.lttng.org/issues/1443

Issue #2133, opened 2024-03-07
chunked_map in seastar


travisdowns:
Seastar cannot in general support large allocations after an application has been running for some time, due to its use of a custom allocator which allocate a fixed virtual memory region which never grows: in general the heap becomes fragmented and allocations some significant multiple of the page size are likely to fail.

Hence the existence of `chunked_fifo` which is a non-contiguous replacement for `std::vector` in some cases.

H/e, the same basic problem also arises with maps. In Redpanda, we will solve this problem using a "fragmented map" whose backing arrays are compose of chunks of some maximum size, not unlike `chunked_fifo` (except that we need random access, so there is a 2-level indexing scheme, not unlike `std::deque`).

Some of the large allocations can occur in seastar itself, e.g., if there are many metrics (link issue).

Given this, we propose that we upstream the fragmented map here instead of keeping in Redpanda. It would almost certainly be an existing 3rd party map implementation, with a compatible license, modified to be chunked. 

WDYT? 

Perhaps this problem already has a solution in Scylla: we'd be interested to hear about it if so.

travisdowns:
cc @tchaikov any thoughts on this one?

avikivity:
We don't have a fragmented unordered map (don't think we ever saw a need - to exceed 128k one would need 16k buckets, no?

In general I don't want Seastar to become a collections library.

travisdowns:
> We don't have a fragmented unordered map (don't think we ever saw a need - to exceed 128k one would need 16k buckets, no?

Something like that: but this is possible for example (and in practice) if you have many metrics: 16K isn't an outlandish number of metrics. Another option would be to use something like `std::map`, which is inherently non-contiguous, for those use cases but performance of that implemetnation is ... not ideal.

nyh:
> In general I don't want Seastar to become a collections library.

Yes, this is the current sentiment. Some historical context: We added chunked_fifo to Seastar almost eight years ago (commit e66ec715eae6e83d3c040429ace8566bdd9ae886) to use it **in** Seastar, namely the semaphore implementation. As you noticed, chunked_fifo is optimized to be just a queue, not a random-addressable vector, which is exactly what we needed for the semaphore.

Here's a core dump of my thoughts on this issue:

[...]

PR #3450, opened 2026-06-05
rpc: replace `_outstanding` unordered_map with log_indexed_container

emdroid:
`std::unordered_map<id_type, unique_ptr<reply_handler_base>>` rehashes with a single large contiguous allocation when the number of in-flight requests crosses a bucket threshold. At high concurrency (e.g. 1000 concurrent ALL-consistency reads) this produces >128 KiB allocations, triggering Seastar's large-alloc warning.

Introduce a new generic utility, `log_indexed_container<T, IndexType>` (`include/seastar/util/log_indexed_container.hh`), which maps monotonically increasing integer indices to values without hashing or large contiguous allocations:

  - Backed by `boost::container::deque`, which releases front blocks on `pop_front` (unlike `std::deque` in libstdc++), keeping memory bounded.
  - Stores values in direct index-offset slots, so lookup is O(1).
  - Trims empty slots from the front after each removal, keeping the deque short relative to the concurrency window.
  - Supports any nullable type (raw pointer, `unique_ptr`) natively, and wraps non-nullable types in `std::optional` internally.
  - API: `emplace(idx, value)`, `find(idx)`, `take(idx)`, `clear_at(idx)`, `clear()`, `size()`, `empty()`, `base_index()`.

Replace `_outstanding` with a `log_indexed_container` specialised on `reply_handler_base` `unique_ptr`. RPC message ids are already monotonically increasing per-connection, making them a natural key. Outstanding reply handlers are removed via `take()`. The timer and cancellation callbacks preserve the original invariant that the slot always exists when they fire: the receive path disarms both atomically (no yield point) before returning to the reactor, so a missing slot indicates a bug.

Also add `tests/unit/log_indexed_container_test.cc` with 13 unit tests covering `emplace`, `find`, `take`, `clear_at`, `trim_front`, `base_index` advancement, and a custom index type.

Fixes: SCYLLADB-1541

### DIFF
--- a/include/seastar/rpc/rpc.hh
+++ b/include/seastar/rpc/rpc.hh
@@ -44,6 +44,7 @@
 #include <seastar/core/semaphore.hh>
 #include <seastar/util/backtrace.hh>
 #include <seastar/util/log.hh>
+#include <seastar/util/log_indexed_container.hh>
 
 namespace bi = boost::intrusive;
 
@@ -555,7 +556,13 @@ public:
         virtual ~reply_handler() {}
     };
 private:
-    std::unordered_map<id_type, std::unique_ptr<reply_handler_base>> _outstanding;
+    // In-flight reply handlers keyed by message id, using log_indexed_container
+    // to avoid large contiguous rehash allocations.
+    // Note: no_wait verbs consume a message id without inserting a handler;
+    // those ids produce an out-of-range lookup on reply, handled gracefully.
+    using reply_table = log_indexed_container<std::unique_ptr<reply_handler_base>, id_type>;
+
+    reply_table _outstanding;
     socket_address _server_addr, _local_addr;
     client_options _options;
     weak_ptr<client> _parent; // for stream clients
--- a/include/seastar/util/log_indexed_container.hh
+++ b/include/seastar/util/log_indexed_container.hh
@@ -1,0 +1,237 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2026-present ScyllaDB
+ */
+
+#pragma once
+
+#include <concepts>
+#include <optional>
+#include <boost/container/deque.hpp>
+
+#include <seastar/util/assert.hh>
+
+namespace seastar {
+
+/// \brief A container that maps monotonically increasing indices to values,
+/// providing O(1) access by index.
+///
+/// Internally uses a deque of slots indexed by `(index - base_index)`.
+/// Completed/cleared slots are nulled in-place; the front is trimmed lazily
+/// (via trim_front()) to reclaim memory.
+///
+/// For value types that already encode an empty/absent state natively
+/// (raw pointer, unique_ptr, shared_ptr, and similar pointer-like types
+/// that are default-constructible to a null state and convertible to bool),
+/// slots store the value directly using its natural empty state.
+/// For all other types, slots are wrapped in std::optional.
+///
+/// Unlike a hash map, the backing deque grows in fixed-size chunks.
+/// boost::container::deque guarantees that memory blocks are released when
+/// items are popped from the ends, so trim_front() actually reclaims memory
+/// (unlike std::deque in libstdc++, which retains blocks).
+///
+/// The container does not require indices to be inserted in strictly
+/// increasing order. Gaps between occupied slots are allowed.
+/// Out-of-range or absent lookups return nullptr safely.
+///
+/// \tparam T         Value type to store.
+/// \tparam IndexType Integer-like type used as the index (default: size_t).
+///                   Must support: operator-(IndexType) → ptrdiff_t-compatible,
+///                   operator<, and prefix operator++.
+template<typename T, typename IndexType = size_t>
+class log_indexed_container {
+    // Detect pointer-like types that already carry a null/absent state:
+    //   - default-constructs to an "empty" value
+    //   - convertible to bool (true = has value, false = empty)
+    //   - is a raw pointer OR has an element_type member (smart pointer)
+    // Plain integers and enums satisfy the first two but not the third,
+    // so they fall through to optional<T> wrapping as intended.
+    template<typename U>
+    static constexpr bool is_nullable =
+        std::is_default_constructible_v<U> &&
+        requires(const U& u) { { static_cast<bool>(u) } -> std::same_as<bool>; } &&
+        (std::is_pointer_v<U> || requires { typename U::element_type; });
+
+    using slot_t = std::conditional_t<is_nullable<T>, T, std::optional<T>>;
+
+    static bool slot_occupied(const slot_t& s) noexcept {
+        if constexpr (is_nullable<T>) {
+            return static_cast<bool>(s);
+        } else {
+            return s.has_value();
+        }
+    }
+
+    // boost::container::deque releases memory blocks on pop_front/pop_back,
+    // unlike std::deque (libstdc++) which may retain them indefinitely.
+    boost::container::deque<slot_t> _slots;
+    IndexType _base_idx{};
+    size_t _count = 0; // number of occupied (non-empty) slots
+
+public:
+    /// Returns a pointer to the value at the given index,
+    /// or nullptr if the slot is empty or out of range.
+    T* find(IndexType idx) noexcept {
+        if (_slots.empty() || idx < _base_idx) {
+            return nullptr;
+        }
+        auto off = static_cast<size_t>(idx - _base_idx);
+        if (off >= _slots.size()) {
+            return nullptr;
+        }
+        auto& slot = _slots[off];
+        if constexpr (is_nullable<T>) {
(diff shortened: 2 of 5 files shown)


include/seastar/rpc/rpc.hh
```diff
@@ -555,7 +556,71 @@ public:
...
+            }
+            return _slots[idx].get();
+        }
+
+        // Remove and return the handler at id. Returns nullptr if not present.
+        std::unique_ptr<reply_handler_base> remove(id_type id) noexcept {
+            auto idx = id - _base_id;
+            if (idx < 0 || static_cast<size_t>(idx) >= _slots.size() || !_slots[idx]) {
+                return nullptr;
+            }
```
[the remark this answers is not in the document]
emdroid: The `id_type` is specifically defined as signed type:

```
using id_type = int64_t;
```

in the same header file, so these checks should work.

include/seastar/rpc/rpc.hh
```diff
@@ -555,7 +556,71 @@ public:
...
+        }
+
+        [[nodiscard]] size_t size() const noexcept {
+            return _count;
+        }
+
+        void clear() noexcept {
+            _slots.clear();
+            _count = 0;
+        }
```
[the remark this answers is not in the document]
emdroid: A real issue, fixed.

include/seastar/rpc/rpc.hh
```diff
@@ -555,7 +556,71 @@ public:
...
+        void trim() noexcept {
+            while (!_slots.empty() && !_slots.front()) {
+                _slots.pop_front();
+                ++_base_id;
+            }
+        }
+
+    private:
+        std::deque<std::unique_ptr<reply_handler_base>> _slots;
+        id_type _base_id = 1; // id corresponding to _slots[0]
```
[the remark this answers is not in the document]
emdroid: fixed

include/seastar/rpc/rpc.hh
```diff
@@ -555,7 +556,71 @@ public:
...
+
+        // Register a handler at id, extending the deque if needed.
+        void insert(id_type id, std::unique_ptr<reply_handler_base> h) {
+            auto idx = id - _base_id;
+            while (static_cast<id_type>(_slots.size()) <= idx) {
+                _slots.push_back(nullptr);
+            }
+            _slots[idx] = std::move(h);
+            ++_count;
+        }
```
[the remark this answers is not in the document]
emdroid: While it's correct that we should address that, it would mean a protocol violation (only expecting the insertions in increasing order, and never reusing an index). Therefore added an assertion on that invariant.

include/seastar/rpc/rpc.hh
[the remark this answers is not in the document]
emdroid: While it is a valid point, it would be a contract violation (of the current id generator). Therefore added assertion to catch any eventual future breakage of this invariant.

include/seastar/rpc/rpc.hh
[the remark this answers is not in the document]
emdroid: While this is a valid point, inserting into an occupied slot will break the contract (expecting always increasing distinct ids) - therefore added assertion to catch a breakage of that contract.

include/seastar/rpc/rpc.hh
```diff
@@ -555,7 +556,78 @@ public:
...
+    // reply_table maps message ids to in-flight reply handlers using a
+    // slot-indexed std::deque rather than std::unordered_map, to avoid large
+    // contiguous rehash allocations. Message ids are monotonically increasing
+    // integers, so slot_index = id - base_id. Completed/cancelled/timed-out
+    // slots are nulled in-place; the front is trimmed lazily to reclaim memory.
+    // std::deque allocates fixed-size chunks (~512B) so no single allocation
+    // ever exceeds a page regardless of in-flight count.
+    // Note: no_wait verbs consume a message id without registering a slot;
+    // those ids produce an out-of-range lookup on reply, handled gracefully.
+    struct reply_table {
```
gusev-p: We used the same idea recently here https://github.com/scylladb/scylladb/pull/29794. Maybe extract a template data structure for reuse? The `log_indexed_container` already has unit tests and also handles the case when insert index is smaller than base_index.

include/seastar/rpc/rpc.hh
```diff
@@ -555,7 +556,78 @@ public:
...
+        void trim() noexcept {
+            while (!_slots.empty() && !_slots.front()) {
+                _slots.pop_front();
+                ++_base_id;
+            }
+        }
+
+    private:
+        std::deque<std::unique_ptr<reply_handler_base>> _slots;
+        id_type _base_id = 1; // id corresponding to _slots[0]
```
gusev-p: `id_type _base_id = 1;` This seems strange, why not zero?

include/seastar/rpc/rpc.hh
```diff
@@ -555,7 +556,78 @@ public:
...
+            return h;
+        }
+
+        // Register a handler at id, extending the deque if needed.
+        void insert(id_type id, std::unique_ptr<reply_handler_base> h) {
+            auto idx = id - _base_id;
+            // ids are monotonically increasing and always >= _base_id;
+            // a negative index means the caller is violating this invariant.
+            SEASTAR_ASSERT(idx >= 0);
+            while (static_cast<id_type>(_slots.size()) <= idx) {
```
gusev-p: batch-insert at the end is likely more efficient `_slots.insert(_slots.end(), count, nullptr)`

include/seastar/rpc/rpc.hh
```diff
@@ -555,7 +556,78 @@ public:
...
+        }
+
+        void clear() noexcept {
+            _slots.clear();
+            _base_id = 1;
+            _count = 0;
+        }
+
+        // Pop null entries from the front, advancing _base_id.
+        void trim() noexcept {
```
gusev-p: This can be made private? There can't be null slots at the front after `remove`-s

src/rpc/rpc.cc
```diff
@@ -805,18 +805,22 @@ void client::wait_for_reply(id_type id, std::unique_ptr<reply_handler_base>&& h,
     }
     if (cancel) {
         cancel->cancel_wait = [this, id] {
-            _outstanding[id]->cancel();
-            _outstanding.erase(id);
+            if (auto* h = _outstanding.get(id)) {
```
gusev-p: why relax the assumptions? previously it was guaranteed that id exists

src/rpc/rpc.cc
```diff
@@ -805,18 +805,22 @@ void client::wait_for_reply(id_type id, std::unique_ptr<reply_handler_base>&& h,
...
         cancel->wait_back_pointer = &h->pcancel;
     }
-    _outstanding.emplace(id, std::move(h));
+    _outstanding.insert(id, std::move(h));
 }
 void client::wait_timed_out(id_type id) {
     _stats.timeout++;
-    _outstanding[id]->timeout();
-    _outstanding.erase(id);
+    if (auto* h = _outstanding.get(id)) {
```
gusev-p: why relax the assumptions? previously it was guaranteed that id exists

include/seastar/rpc/rpc.hh
emdroid: _message_id starts at 1 in client, so the first id inserted is 1, meaning slot 0 would always be empty if _base_id = 0. Starting at 1 avoids that wasted slot. Which is what Claude inferred when working at the fix - but starting at 0 doesn't make a big difference I guess, or at least comment could be added with the explanation. But especially if we'd create a shared implementation, starting at 0 would be better indeed

include/seastar/rpc/rpc.hh
emdroid: Right, that would make sense, only we'd need to move (copy first) the `log_indexed_container` to Seastar. Btw. there is also the idea of "chunked map" https://github.com/scylladb/seastar/issues/2133 - which this might eventually evolve into.

emdroid:
New version:
* copied over the existing `log_indexec_container` from https://github.com/scylladb/scylladb/pull/29794, that is to be used in similar case in the scylladb.

include/seastar/util/log_indexed_container.hh
bitpathfinder, line 1: IMO the changes should be separated. 1st commit should introduce the new container, the second replace the `unordered_map`.

include/seastar/rpc/rpc.hh
patjed41: Is this PR urgent? There are some comments in https://github.com/scylladb/scylladb/pull/29794 which I need to address, and IIUC I must do it before we merge this PR. I'm not sure when I'll do it, though, as I have some higher-priority work (mostly reviews) and I'm going on vacation on Thursday.

emdroid: It's P3 and not marked as CI stability, so not really urgent.

avikivity:
Any new containers should be internal to Seastar and not exported, unless good reason is demonstrated.

emdroid:
> Any new containers should be internal to Seastar and not exported, unless good reason is demonstrated.

Here one reason might be its reusability in the Scylla. But as mentioned in https://scylladb.atlassian.net/browse/SCYLLADB-1541?focusedCommentId=66075, there are multiple ways of addressing this:
1. Reserve + increasing the load factor of the map (used in one similar case)
2. RPC local-only structure (which was the initial version of the PR)
3. The log_indexed_container from [scylladb#29794](https://github.com/scylladb/scylladb/pull/29794) (= this current version)
4. Actually implementing the Seastar chunked map as discussed in [seastar#2133](https://github.com/scylladb/seastar/issues/2133) (highest effort, but general reusable structure then)
5. Implement chunked_deque (in addition to chunked_vector) that might be used in this case
6. anything else?

For the immediate fix the solution 1. (the simplest) or 2. (internal structure, also fairly simple) would suffice, however the current ongoing  [scylladb#29794](https://github.com/scylladb/scylladb/pull/29794) has been pointed out by @gusev-p as possibly reusable structure.

And, implementing the chunked_map would be the most reusable solution for all the similar cases where an unordered map causes oversized allocations.

emdroid:
Closing in favour of #3461 which takes the minimal approach agreed at the Raft sync (2026-06-15): tune `max_load_factor` and `reserve` on the existing `unordered_map` rather than replacing the container. A follow-up ticket will track evaluation of structural alternatives.

PR #3453, opened 2026-06-10
core: sleep_abortable propagates custom abort exception when already aborted

### LINKED ISSUE #3452: sleep_abortable(duration, abort_source&) discards a custom abort exception when the source is already aborted before subscribing
## Summary

`sleep_abortable(typename Clock::duration, abort_source&)` has two paths for handling an abort, and they disagree about which exception the returned future fails with:

- **Abort happens *during* the sleep** (subscription is live): the subscriber callback re-throws the exception passed to `request_abort_ex(...)`, so a custom abort exception is preserved.
- **Source is *already aborted* before `sleep_abortable` is called** (subscription is never linked): it unconditionally fails the future with a freshly-constructed `sleep_aborted`, discarding the exception the `abort_source` already stored.

So whether a custom abort exception survives depends on the race between the abort and the moment `sleep_abortable` subscribes, which is surprising and inconsistent.

## Where

`sleep_abortable<Clock>(duration, abort_source&)` (declared in `include/seastar/core/sleep.hh`):

```cpp
auto sc_opt = as.subscribe([this](const std::optional<std::exception_ptr>& opt_ex) noexcept {
    if (tmr.cancel()) {
        done.set_exception(opt_ex.value_or(std::make_exception_ptr(sleep_aborted())));
    }
});
if (sc_opt) {
    sc = std::move(*sc_opt);
    tmr.arm(dur);
} else {
    done.set_exception(sleep_aborted());   // <-- ignores the abort_source's stored exception
}
```

`abort_source::subscribe()` returns a disengaged `optimized_optional` when the source is already aborted (the `subscription` is constructed but not linked, and `subscription::operator bool()` returns `is_linked()`), so the `else` branch runs. Meanwhile `do_request_abort()` has already stored the canonical exception in `_ex` (the custom one for `request_abort_ex`, or `get_default_exception()` otherwise), and `abort_source::check()` rethrows it.

## Reproduction

```cpp
struct my_error : std::exception {
    const char* what() const noexcept override { return "my_error"; }
};

// (A) abort DURING the sleep -> propagates my_error
abort_source as_a;
auto f = sleep_abortable<steady_clock_type>(10s, as_a);  // subscribes first
as_a.request_abort_ex(my_error{});                       // aborts the live subscription
// co_await std::move(f)  =>  throws my_error

// (B) abort BEFORE subscribing -> throws sleep_aborted, my_error is lost
abort_source as_b;
as_b.request_abort_ex(my_error{});
auto g = sleep_abortable<steady_clock_type>(10ms, as_b); // subscribe() already disengaged
// co_await std::move(g)  =>  throws sleep_aborted  (NOT my_error)
```

## Which side is buggy

I'd argue the **during-sleep behavior (propagating the abort source's exception) is the correct/intended one** -- that's the reason `request_abort_ex` exists -- so the already-aborted fast path is the bug: it should surface the source's stored exception too, not fabricate a `sleep_aborted`. Something like:

```cpp
} else {
    // source already aborted: honor its exception, consistent with the live path
    // (falls back to sleep_aborted only when no exception was supplied)
    done.set_exception(/* as's stored exception, e.g. via as.check() */);
}
```

There's currently no public getter for the stored `exception_ptr` (only `check()` rethrows it), so this likely needs a small helper or a `try { as.check(); } catch (...) { done.set_exception(std::current_exception()); }` in the `else` branch.

Relatedly, the doc comment on `sleep_abortable` ("Returns a future which completes after a specified time has elapsed or throws `sleep_aborted` exception if the sleep is aborted") is now inaccurate for `request_abort_ex` and should say it propagates the abort source's exception, falling back to `sleep_aborted` when none was supplied.

## Impact

Callers that abort with a meaningful custom exception and rely on catching that specific type (rather than the `sleep_aborted` / `abort_requested_exception` base) silently lose it in the already-aborted window. We hit this in redpanda's http client reconnect loop, where an abort can land during a connect attempt so that the backoff `sleep_abortable` is entered already-aborted; we work around it locally by translating `sleep_aborted` back through `abort_source::check()`.

travisdowns:
## Problem

`sleep_abortable(dur, abort_source&)` had two abort paths that disagreed on the
exception the returned future fails with:

- **Abort *during* the sleep**: the subscriber callback propagates the exception
  supplied via `abort_source::request_abort_ex(...)`, so a custom exception is
  preserved.
- **Source *already aborted* before `sleep_abortable` subscribes**: `subscribe()`
  returns a disengaged `optimized_optional` (the subscription is built but never
  linked), so the `else` fast path ran and always fabricated a fresh
  `sleep_aborted`, **discarding** the exception the `abort_source` already
  stored.

So whether a custom abort exception survived depended on a race with the moment
`sleep_abortable` subscribed.

## Fix

Make the already-aborted fast path mirror the during-sleep path: propagate the
caller-supplied exception when one was given, otherwise fall back to
`sleep_aborted`.

To distinguish "caller supplied an exception" from "default abort" on the fast
path, `abort_source::get_default_exception()` now returns a stable (per-shard)
shared `exception_ptr`. A plain `request_abort()` stores that same object, so
`sleep_abortable` can tell a caller-supplied exception apart by comparing the
stored abort exception against it by identity. A plain `request_abort()` still
yields `sleep_aborted` exactly as before.

## Behavior

| Abort | Exception |
|-------|-----------|
| `request_abort_ex(custom)` during sleep | `custom` |
| `request_abort_ex(custom)` already aborted | `custom` (previously `sleep_aborted` — the bug) |
| `request_abort()` during sleep | `sleep_aborted` |
| `request_abort()` already aborted | `sleep_aborted` (unchanged) |

## Tests

Added to `tests/unit/abort_source_test.cc`:

- `test_sleep_abortable_already_aborted_with_exception` — custom exception is
  propagated on the already-aborted path (fails before the fix).
- `test_sleep_abortable_already_aborted_no_exception` — a plain abort still
  yields `sleep_aborted`.

Fixes #3452

### DIFF
--- a/tests/unit/abort_source_test.cc
+++ b/tests/unit/abort_source_test.cc
@@ -197,6 +197,17 @@ SEASTAR_THREAD_TEST_CASE(test_sleep_abortable_already_aborted_with_exception) {
     });
 }
 
+// Companion to the above: when the already-aborted source carries no custom
+// exception (plain request_abort), sleep_abortable still fails with
+// sleep_aborted, matching the during-sleep path. See scylladb/seastar#3452.
+SEASTAR_THREAD_TEST_CASE(test_sleep_abortable_already_aborted_no_exception) {
+    abort_source as;
+    as.request_abort();
+    auto f = sleep_abortable(10s, as);
+
+    BOOST_REQUIRE_THROW(f.get(), sleep_aborted);
+}
+
 SEASTAR_THREAD_TEST_CASE(test_destroy_with_moved_subscriptions) {
     auto as = std::make_unique<abort_source>();
     int aborted = 0;
--- a/tests/unit/abort_source_test.cc
+++ b/tests/unit/abort_source_test.cc
@@ -20,6 +20,7 @@
  */
 
 #include <exception>
+#include <string_view>
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
 
@@ -178,6 +179,22 @@ SEASTAR_THREAD_TEST_CASE(test_sleep_abortable_with_exception) {
         caught_exception = true;
     }
     BOOST_REQUIRE(caught_exception);
+}
+
+// Regression test for scylladb/seastar#3452: when the abort_source is already
+// aborted before sleep_abortable subscribes, the custom abort exception must
+// still be propagated, just as it is when the abort lands during the sleep.
+SEASTAR_THREAD_TEST_CASE(test_sleep_abortable_already_aborted_with_exception) {
+    abort_source as;
+    auto expected_message = "expected";
+    // Abort the source *before* calling sleep_abortable, so subscribe() returns
+    // a disengaged optional and the already-aborted fast path is taken.
+    as.request_abort_ex(std::runtime_error(expected_message));
+    auto f = sleep_abortable(10s, as);
+
+    BOOST_REQUIRE_EXCEPTION(f.get(), std::runtime_error, [&] (const std::runtime_error& e) {
+        return e.what() == std::string_view(expected_message);
+    });
 }
 
 SEASTAR_THREAD_TEST_CASE(test_destroy_with_moved_subscriptions) {
--- a/include/seastar/core/abort_source.hh
+++ b/include/seastar/core/abort_source.hh
@@ -213,8 +213,15 @@ public:
 
     /// Returns the default exception type (\ref abort_requested_exception) for this abort source.
     /// Overridable by derived classes.
+    ///
+    /// The same exception_ptr object is returned on every call, so callers can
+    /// compare an abort exception (\ref abort_requested_exception_ptr) against
+    /// it by identity to tell whether an abort used the default exception or a
+    /// caller-supplied one. Overrides should preserve this property by likewise
+    /// returning a stable object.
     virtual std::exception_ptr get_default_exception() const noexcept {
-        return make_exception_ptr(abort_requested_exception());
+        static const thread_local std::exception_ptr ex = make_exception_ptr(abort_requested_exception());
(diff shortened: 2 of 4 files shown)


xemul:
[this](https://github.com/scylladb/seastar/pull/3453/changes/c8b02de8472218df0483ff3161ffac5d19b9b1f8) commit is alien here

travisdowns:
> [this](https://github.com/scylladb/seastar/pull/3453/changes/c8b02de8472218df0483ff3161ffac5d19b9b1f8) commit is alien here

Removed via rebase.

tests/unit/abort_source_test.cc
```diff
@@ -180,6 +180,27 @@ SEASTAR_THREAD_TEST_CASE(test_sleep_abortable_with_exception) {
...
+    auto f = sleep_abortable(10s, as);
+
+    bool caught_exception = false;
+    try {
+        f.get();
+    } catch (const std::runtime_error& e) {
+        BOOST_REQUIRE_EQUAL(e.what(), expected_message);
+        caught_exception = true;
+    }
+    BOOST_REQUIRE(caught_exception);
```
xemul: Would something like `BOOST_REQUIRE_EXCEPTION(f.get(), std::runtime_error, [] (auto ex) { return ex.what() == expected_message; })` work just as nice?

travisdowns: Yes, done. One tweak versus the snippet as written: `ex.what() == expected_message` would compare `const char*` pointers (and `runtime_error` copies its message), so the predicate compares via `std::string_view` instead. Also collapsed the companion no-exception test to a plain `BOOST_REQUIRE_THROW`.

PR #3460, opened 2026-06-16
http/client: add per-HTTP-method I/O statistics

kreuzerkrieg:
### Problem

Users of the seastar HTTP client have no built-in way to observe per-method I/O activity (request counts, retries, latency). Today every consumer that wants this visibility has to wrap the client and re-implement its own tracking, which is duplicated work and easy to get wrong (e.g. missing the retry path, or only counting one verb).

### Changes

The HTTP `client` now tracks per-HTTP-method statistics directly:

- A new `http::http_method_stats` struct holds `ops`, `retries`, and accumulated `latency` (as `std::chrono::duration<double>`).
- A new `http::client_stats` container holds an `std::array<http_method_stats, NUM_OPERATION>` indexed by `httpd::operation_type`, with `operator[]` overloads for ergonomic access.
- `client::make_request` resolves the verb via `httpd::str2type`, increments `ops`, and accumulates wall-clock latency in a `.finally()` block using `lowres_clock`.
- `client::maybe_retry_request` takes the resolved `operation_type` as an extra argument and increments `retries` whenever the retry strategy decides to retry.
- A `const get_stats()` accessor exposes the container so callers can read counters without owning any tracking code.

This keeps the wiring entirely inside the client — no API break for existing callers of `make_request`, no behavior change on the success/failure paths beyond the counter updates.

### DIFF
--- a/include/seastar/http/client.hh
+++ b/include/seastar/http/client.hh
@@ -21,8 +21,11 @@
 
 #pragma once
 
+#include <array>
+#include <chrono>
 #include <boost/intrusive/list.hpp>
 #include <seastar/net/api.hh>
+#include <seastar/http/common.hh>
 #include <seastar/http/connection_factory.hh>
 #include <seastar/http/reply.hh>
 #include <seastar/http/retry_strategy.hh>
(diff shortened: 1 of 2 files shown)


include/seastar/http/client.hh
```diff
@@ -43,6 +46,26 @@ class client;
 struct request;
 struct reply;
 
+/// Per-HTTP-method I/O statistics
+struct http_method_stats {
```
xemul, line 55: Fields must be documented

include/seastar/http/client.hh
```diff
@@ -43,6 +46,26 @@ class client;
...
+    uint64_t ops = 0;
+    uint64_t retries = 0;
+    std::chrono::duration<double> latency{0};
+};
+
+/// Container holding per-method stats indexed by operation_type
+struct client_stats {
+    static constexpr size_t num_methods = static_cast<size_t>(httpd::operation_type::NUM_OPERATION);
+    std::array<http_method_stats, num_methods> methods{};
```
xemul, line 69: If there's operator[], why keeping the array public?

kreuzerkrieg: just convenience to avoid `slients_stats().methods[POST]` and use just `slients_stats()[POST]` instead.

kreuzerkrieg:
v next:
- fields annotated
- client_stats became a class
- unnecessary static_casts removed

PR #3462, opened 2026-06-17
build: update supported C++ standards to C++23, C++26

avikivity:
Since C++26 was completed [1], we retire C++20 support and introduce C++26. The tests matrix was regenerated.

Cooking fmt version updated to 12.2.0 to include a bug fix for std::optional formatting with C++26 [2].

[1] https://herbsutter.com/2026/03/29/c26-is-done-trip-report-march-2026-iso-c-standards-meeting-london-croydon-uk/

[2] https://github.com/fmtlib/fmt/issues/4760


avikivity:
v2: cook fmt for gcc and C++26 combination, not just clang.

nyh (approved):
Looks good. Now I'll go learn C++26 :-)

avikivity:
> Looks good. Now I'll go learn C++26 :-)

You should have started three years ago!

configure.py
```diff
@@ -244,7 +244,7 @@ def identify_best_standard(cpp_standards, compiler):
 
 
 if not args.cpp_standard:
-    cpp_standards = ['23', '20']
+    cpp_standards = ['26', '23']
```
travisdowns, line 247: I always found this a bit backwards: feels like we should choose the lower version by default as that's going to work with more compilers (and it doesn't affect CI as it chooses the std explicitly) but but it's the way it has always been, I guess :).

travisdowns (approved):
LGTM!

configure.py
avikivity: It's intentional, nudge the user towards a newer Standard (we get higher kickbacks from ISO this way)

PR #3463, opened 2026-06-17
core: add input/output stream copy_n (copies exactly n)

### LINKED ISSUE #962: the stream copy function should have a copy_n variant
It is sometimes helpful to copy only n bytes. While there is a copy(stream, stream) there is no copy_n variant of the function

raphaelroshan:
`copy(input_stream&, output_stream&)` copies an entire input stream to an
output stream, but there is no way to copy a bounded number of bytes.

Add `copy_n(input_stream&, output_stream&, size_t n)` which copies **exactly
n** bytes and throws `std::runtime_error` if the input reaches end-of-stream
before n bytes have been copied. This mirrors `std::copy_n` and seastar's own
`read_exactly`, and keeps the `future<>` return meaningful (success implies
exactly n bytes were copied, rather than a silently truncated count).

It is implemented as a coroutine that reads via `read_up_to()` and writes each
chunk, so larger-than-buffer copies use a single coroutine frame rather than a
chain of `future::then` continuations.

Fixes #962.

### DIFF
--- a/include/seastar/core/iostream-impl.hh
+++ b/include/seastar/core/iostream-impl.hh
@@ -617,6 +617,7 @@ public:
         });
     }
 };
+
 /// \endcond
 
 }
@@ -626,6 +627,22 @@ extern template struct internal::stream_copy_consumer<char>;
 template <typename CharType>
 future<> copy(input_stream<CharType>& in, output_stream<CharType>& out) {
     return in.consume(internal::stream_copy_consumer<CharType>(out));
+}
+
+/// \brief copy exactly \c n bytes from the input stream to the output stream
+///
+/// \throws std::runtime_error if the input stream reaches end-of-stream before
+/// \c n bytes have been copied.
+template <typename CharType>
+future<> copy_n(input_stream<CharType>& in, output_stream<CharType>& out, size_t n) {
+    while (n != 0) {
+        auto buf = co_await in.read_up_to(n);
+        if (buf.empty()) {
+            throw std::runtime_error("copy_n: input stream reached end-of-stream before copying the requested number of bytes");
+        }
+        n -= buf.size();
+        co_await out.write(std::move(buf));
+    }
 }
 
 extern template future<> copy<char>(input_stream<char>&, output_stream<char>&);
--- a/tests/unit/output_stream_test.cc
+++ b/tests/unit/output_stream_test.cc
@@ -25,6 +25,7 @@
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
 #include <seastar/testing/random.hh>
+#include <seastar/util/memory-data-source.hh>
 #include <vector>
 #include <list>
 #include <deque>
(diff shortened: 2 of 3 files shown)


tests/unit/output_stream_test.cc
```diff
@@ -257,6 +257,44 @@ SEASTAR_THREAD_TEST_CASE(test_flush_on_empty_buffer_does_not_push_empty_packet_d
     BOOST_REQUIRE(ss.str().empty());
 }
 
+namespace {
+
+// A data_source that yields a single fixed buffer once, then EOF.
+class string_data_source_impl final : public data_source_impl {
```
xemul: Consider using basic_memory_data_sink instead

tests/unit/output_stream_test.cc
```diff
@@ -257,6 +257,44 @@ SEASTAR_THREAD_TEST_CASE(test_flush_on_empty_buffer_does_not_push_empty_packet_d
...
+        auto in = make_string_input_stream(src);
+        auto out = output_stream<char>(testing::memory_data_sink(ss), 8);
+        copy_n(in, out, n).get();
+        out.close().get();
+        BOOST_REQUIRE_EQUAL(ss.str(), expected);
+    };
+    check("hello world", 5, "hello");
+    check("hello", 100, "hello");
+    check("hello", 0, "");
+    check("hello", 5, "hello");
```
xemul: It's better to write a for-loop that checks all cases, rather than several hand-crafted ones

include/seastar/core/iostream-impl.hh
```diff
@@ -628,5 +658,11 @@ future<> copy(input_stream<CharType>& in, output_stream<CharType>& out) {
     return in.consume(internal::stream_copy_consumer<CharType>(out));
 }
 
+/// \brief copy at most \c n bytes from the input stream to the output stream
```
avikivity: If it copies at most n, it should be named in a way that suggests it.

Why is copying at most n bytes useful? I think it's more useful to copy exactly n bytes, or fail, but maybe you have a use case in mind.

tests/unit/output_stream_test.cc
```diff
@@ -257,6 +257,44 @@ SEASTAR_THREAD_TEST_CASE(test_flush_on_empty_buffer_does_not_push_empty_packet_d
...
+}
```
xemul, line 315: The test must also validate that input still has the expected remaining (unconsumed) bytes available

include/seastar/core/iostream-impl.hh
```diff
@@ -617,6 +617,36 @@ public:
...
+                return make_ready_future<consumption_result_type>(stop_consuming(std::move(leftover)));
+            });
+        }
+        return _os.write(data.get(), sz).then([this] {
+            if (_remaining == 0) {
+                return make_ready_future<consumption_result_type>(stop_consuming(temporary_buffer<CharType>()));
+            }
+            return make_ready_future<consumption_result_type>(continue_consuming());
+        });
+    }
```
avikivity: Consider making this a coroutine.

For small copies, everything will be buffered, and the continuations will work better.

For anything that is larger than the buffer (either input or output), coroutines will be better, since we avoid repeated allocations by future::then and replace them with a single allocation for the coroutine frame.

tests/unit/output_stream_test.cc
```diff
@@ -257,6 +257,44 @@ SEASTAR_THREAD_TEST_CASE(test_flush_on_empty_buffer_does_not_push_empty_packet_d
...
+
+SEASTAR_THREAD_TEST_CASE(test_copy_n) {
+    auto check = [] (sstring src, size_t n, sstring expected) {
+        std::stringstream ss;
+        auto in = make_string_input_stream(src);
+        auto out = output_stream<char>(testing::memory_data_sink(ss), 8);
+        copy_n(in, out, n).get();
+        out.close().get();
+        BOOST_REQUIRE_EQUAL(ss.str(), expected);
```
avikivity: Please check the unconsumed reminder in in as well.

tests/unit/output_stream_test.cc
avikivity: Need to check cases where the consumer loops (copy larger than the buffer but smaller than the stream).

raphaelroshan:
Thank you for the thorough reviews, let me fix these!

include/seastar/core/iostream-impl.hh
raphaelroshan: Yup that makes sense to me, let me update it to copy exactly n bytes!

include/seastar/core/iostream-impl.hh
raphaelroshan: Thanks for explaining, let me do that

xemul:
Conflicts, please rebase

tests/unit/output_stream_test.cc
raphaelroshan: good idea, replaced with a for loop!

raphaelroshan:
Have updated - 
Replaced with a coroutine over read_up_to(), test is a for loop and over sizes that cross the buffer and assert input's remaining bytes. Copy_n copies exactly n and throws if unexpected EOF - any concerns here?

avikivity:
Looks good, will let others review too.

PR #3466, opened 2026-06-18
util: print concise exception messages in logs

### LINKED ISSUE #1029: Exception printout is too verbose
In the Scylla project we often print exceptions using Seastar's logging. Here is an example error message, where the entire text in bold is one exception:

> ERROR 2022-03-21 14:33:13,657 [shard 0] init - Startup failed: **std::_Nested_exception<std::runtime_error> (Failed to set up Alternator TLS credentials): std::_Nested_exception<std::runtime_error> (Could not read certificate file conf/scylla.crt): std::filesystem::__cxx11::filesystem_error (error system:2, filesystem error: open failed: No such file or directory [conf/scylla.crt])**

This is ugly, and I want to suggest that maybe most of the type information is redundant and not helpful to end-users. Users of the application don't know what to make of internal C++ type names like `std::_nested_exception<std::runtime_error>`. In this case simple punctuation like a colon could replace the entire "nested exception" song and dance (see example below). Even std::filesystem::__cxx11::filesystem_error is not helpful, and arguably so is the sentence "error system:2" that we go out of our way to print (in `log.cc`, `std::ostream& operator<<(std::ostream& out, const std::exception_ptr& eptr)`) is also not helpful. Rather, the `what()` of this error would have been enough.

In the above example, I would argue that the following shorter version of the same message would be much easier to understand:

> ERROR 2022-03-21 14:33:13,657 [shard 0] init - Startup failed: **Failed to set up Alternator TLS credentials): Could not read certificate file conf/scylla.crt: open failed: No such file or directory [conf/scylla.crt]**

Perhaps for some rare types of exceptions it can be helpful to print their name, but arguably for types (like runtime_error and filesystem_error) which implement a `what()`, only that should be printed and not the exception's type.

raphaelroshan:
When logging an `exception_ptr`, `operator<<` always prefixes the
demangled C++ type name and wraps the message in parentheses, e.g.

```
std::runtime_error (Failed to do X)
std::_Nested_exception<std::runtime_error> (msg1): std::runtime_error (msg2)
```

The type name is rarely useful and makes nested chains hard to read.

Print only `what()` for `std::exception` subclasses, keeping the
demangled type name only for unknown (non-`std::exception`) objects,
where it is the only information available. `seastar::nested_exception`
is formatted as `{inner} (while cleaning up after {outer})` without the
type prefix, and a null `std::nested_exception` no longer appends a
trailing `: <no exception>`.

The examples above become:

```
Failed to do X
msg1: msg2
```

Fixes #1029.

### DIFF
--- a/src/util/log.cc
+++ b/src/util/log.cc
@@ -640,10 +640,14 @@ std::ostream& operator<<(std::ostream& out, const std::exception_ptr& eptr) {
         std::rethrow_exception(eptr);
     } catch(...) {
         // Print what() for std::exception; fall back to the type name otherwise.
+        // system_error additionally carries an error code that what() does not
+        // include, so keep that without the verbose type-name prefix.
         try {
             throw;
         } catch (const seastar::nested_exception& ne) {
             out << fmt::format("{} (while cleaning up after {})", ne.inner, ne.outer);
+        } catch (const std::system_error& e) {
+            out << e.what() << " (error " << e.code() << ")";
         } catch (const std::exception& e) {
             out << e.what();
         } catch (...) {
@@ -673,7 +677,7 @@ std::ostream& operator<<(std::ostream& out, const std::exception& e) {
 }
 
 std::ostream& operator<<(std::ostream& out, const std::system_error& e) {
-    return out << e.what();
+    return out << e.what() << " (error " << e.code() << ")";
 }
 
 }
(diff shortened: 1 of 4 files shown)


tgrabiec:
> is formatted as {inner} (while cleaning up after {outer}) 

Nesting doesn't always happen due to failure during cleanup, sometimes the chaining happens to translate between abstraction levels. For example:

```
    } catch (const std::system_error &) {
        std::throw_with_nested(exceptions::authentication_exception("Could not verify password"));

```

Java's "caused by" formatting captures this fact better.

PR #3469, opened 2026-06-19
http: fix response line for unknown status codes

### LINKED ISSUE #3468: HTTP replies with unknown status codes omit required status-line separator
### Problem

`http::reply::response_line()` formats the HTTP status line from `status_strings::with_string_view()`. For statuses missing from the known-status map, the fallback text is only the numeric code, for example `451` or `299`.

That makes the generated start line look like:

```http
HTTP/1.1 451\r\n
```

HTTP status lines require a space after the three-digit status code even when the reason phrase is empty. Seastar's response parser also encodes this expectation in `src/http/response_parser.rl` with `http_version space status_code space ... crlf`, so such responses cannot be parsed by Seastar's own HTTP client.

### Expected

Unknown or custom status codes should be serialized with an empty reason phrase separator:

```http
HTTP/1.1 451 \r\n
```

### Suggested fix

Keep the existing known-status strings unchanged, but make `reply::response_line()` add the trailing separator when the status text contains no reason phrase. Add a regression test that serializes a status not present in the known-status map.

dkropachev:
Fixes #3468

## Problem

`http::reply::response_line()` builds the HTTP start line from known status strings that include both the numeric code and reason phrase, for example `200 OK`.

For custom or unmapped status values, the generic status formatter falls back to only the numeric code, for example `299` or `451`. When `response_line()` used that same generic formatting path, unknown statuses serialized as:

```http
HTTP/1.1 299\r\n
```

HTTP status lines still require the separator after the three-digit status code when the reason phrase is empty. Seastar's own response parser expects the same shape in `src/http/response_parser.rl`:

```ragel
http_version space status_code space ... crlf
```

So replies using custom or missing standard status codes could be malformed on the wire and rejected by Seastar's HTTP client parser.

## Solution

- Add a status table lookup helper so `response_line()` can distinguish known status strings from numeric fallback values.
- Keep known statuses unchanged, for example `HTTP/1.1 200 OK\r\n`.
- Format unknown statuses directly in `response_line()` with an empty reason phrase separator, for example `HTTP/1.1 299 \r\n`.
- Keep `operator<<` behavior unchanged for status values by moving its numeric fallback directly into the stream operator and removing the now-unnecessary `with_string_view()` callback helper.
- Add a `write_reply()` regression test using status `299`, which is intentionally not present in the known-status table, so the fallback path remains covered even if another standard code is added later.

## Testing

- `git diff --check HEAD~1..HEAD`
- `ninja -C build/dev test_unit_httpd_run`

### DIFF
--- a/src/http/reply.cc
+++ b/src/http/reply.cc
@@ -91,27 +91,27 @@ static const std::unordered_map<reply::status_type, std::string_view> status_str
     {reply::status_type::network_read_timeout, "598 Network Read Timeout"},
     {reply::status_type::network_connect_timeout, "599 Network Connect Timeout"}};
 
-template<typename Func>
-static auto with_string_view(reply::status_type status, Func&& func) -> std::invoke_result_t<Func, std::string_view> {
-  if (auto found = status_strings.find(status); found != status_strings.end()) [[likely]] {
-     return func(found->second);
-  }
-  auto dummy_buf = std::to_string(int(status));
-  return func(dummy_buf);
+static std::optional<std::string_view> find(reply::status_type status) {
+    if (auto found = status_strings.find(status); found != status_strings.end()) [[likely]] {
+        return found->second;
+    }
+    return std::nullopt;
 }
 
 } // namespace status_strings
 
 std::ostream& operator<<(std::ostream& os, reply::status_type st) {
-    return status_strings::with_string_view(st, [&](std::string_view txt) -> std::ostream& {
-        return os << txt;
-    });
+    if (auto status = status_strings::find(st); status) [[likely]] {
+        return os << *status;
+    }
+    return os << static_cast<int>(st);
 }
 
 sstring reply::response_line() const {
-    return status_strings::with_string_view(_status, [this](std::string_view txt) {
-        return seastar::format("HTTP/{} {}\r\n", _version, txt);
-    });
+    if (auto status = status_strings::find(_status); status) [[likely]] {
+        return seastar::format("HTTP/{} {}\r\n", _version, *status);
+    }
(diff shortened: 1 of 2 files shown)


avikivity:
Please don't add empty commits. And make sure the commit changelogs have the explanations.

dkropachev:
> Please don't add empty commits. And make sure the commit changelogs have the explanations.

Dropped it.

src/http/reply.cc
```diff
@@ -110,7 +110,8 @@ std::ostream& operator<<(std::ostream& os, reply::status_type st) {
 
 sstring reply::response_line() const {
     return status_strings::with_string_view(_status, [this](std::string_view txt) {
-        return seastar::format("HTTP/{} {}\r\n", _version, txt);
+        auto has_reason_phrase = txt.find(' ') != std::string_view::npos;
```
xemul: This is ugly. This txt.find "fixes" the problem introduced few lines above by the static auto with_string_view, so it's better to patch with_string_view to generate extra space if needed

xemul (changes requested):

src/http/reply.cc
dkropachev: Agreed, since behavior should be different for `operator<<` and `response_line` I split logic and since `with_string_view` is not used anywhere else I dropped it, please take a look.

xemul:
For the record. https://httpwg.org/specs/rfc9112.html#status.line says

```
The first line of a response message is the status-line, consisting of the protocol version, a space (SP), the status code, and another space and ending with an OPTIONAL textual phrase describing the status code.

  status-line = HTTP-version SP status-code SP [ reason-phrase ]
```

so yes, extra space is required. I'll add it to commit text

PR #3472, opened 2026-06-20
memory: honor large allocation alignment

### LINKED ISSUE #3471: memory: large aligned allocations can return under-aligned pointers
## Summary

The Seastar allocator's large aligned allocation path computes the requested alignment in pages, but the lower-level allocator ignores that value. As a result, valid aligned allocation requests whose alignment is larger than the requested size can return a pointer that is only aligned to the allocated buddy span size, not to the requested alignment.

For example, on the default 4 KiB allocator page size, a reactor-thread call like:

```c++
void* p = nullptr;
int rc = posix_memalign(&p, 2 << 20, 4096);
```

can succeed with `p` only 4 KiB-aligned instead of 2 MiB-aligned. That violates the `posix_memalign()` alignment contract.

## Code path

Current `origin/master` has this flow:

- `posix_memalign()` calls `allocate_aligned(align, size)`.
- `allocate_aligned()` routes requests with `align > page_size` through `allocate_large_aligned(align, size)`, even when `size <= max_small_allocation`.
- `allocate_large_aligned(size_t align, size_t size, ...)` computes `align_in_pages` and passes it down.
- `cpu_pages::allocate_large_aligned(unsigned align_pages, unsigned n_pages, ...)` ignores `align_pages` and delegates directly to `allocate_large_and_trim(n_pages, ...)` with the comment `// buddy allocation is always aligned`.
- `allocate_large_and_trim()` selects a suitable buddy span, trims it down to the requested `n_pages`, and returns `mem() + span_idx * page_size`.

Buddy allocation only guarantees natural alignment to the span size that is actually allocated. If `n_pages == 1`, the returned pointer is only guaranteed to be one allocator page aligned. It is not guaranteed to satisfy a larger requested alignment such as 512 pages / 2 MiB.

## Why this matters

This is reachable through public aligned allocation entry points on reactor threads, including:

- `posix_memalign()`
- `memalign()`
- `aligned_alloc()`
- aligned C++ `operator new(size_t, std::align_val_t)`

Existing allocator tests cover cases where `size >= align` up to 64 KiB, but they do not cover the valid case where `align > size`, such as `posix_memalign(&p, 2 MiB, 4 KiB)`.

## Expected behavior

Large aligned allocations should either return a pointer aligned to the requested alignment or fail. For requests where `align_pages > n_pages`, the allocator likely needs to allocate/search a span large enough to contain an aligned subspan, return the aligned portion, and free the unused prefix/suffix appropriately.

Related: #3467 also touches `allocate_large_aligned()`, but it is about page-count overflow/truncation. This issue is about the requested alignment being ignored after conversion to pages.

dkropachev:
## Problem

`cpu_pages::allocate_large_aligned()` accepted `align_pages`, but treated any buddy allocation as sufficiently aligned and delegated to `allocate_large_and_trim(n_pages)`. For requests where the requested alignment is larger than the requested size, such as `posix_memalign(&p, 2 MiB, 4 KiB)`, the allocator could return a span that is only allocator-page aligned.

Sized aligned deallocation had a related issue for this class of allocations: when `align > page_size` and `size <= max_small_allocation`, `free_aligned()` could pass the small size into sized free and take the small-pool path even though allocation used the large path.

## Fix

- For large aligned requests, allocate from a buddy span sized for `max(align_pages, n_pages)` so the returned span satisfies the requested alignment.
- Route sized aligned frees with `align > page_size` through the ordinary pointer-inspecting free path.
- Add regression coverage for `posix_memalign()` and aligned C++ new where alignment is larger than size.

Fixes #3471

## Testing

- `ninja -C build/dev test_unit_alloc`
- `ctest --test-dir build/dev -R '^Seastar\.unit\.alloc$' --output-on-failure`
- `build/dev/tests/unit/alloc_test --run_test=test_large_alignment_larger_than_size --logger=HRF,test_suite --color_output=false -- --smp 1`

### DIFF
--- a/src/core/memory.cc
+++ b/src/core/memory.cc
@@ -943,8 +943,7 @@ cpu_pages::allocate_large(unsigned n_pages, bool should_sample) {
 void*
 cpu_pages::allocate_large_aligned(unsigned align_pages, unsigned n_pages, bool should_sample) {
     check_large_allocation(n_pages * page_size);
-    // buddy allocation is always aligned
-    return allocate_large_and_trim(n_pages, should_sample);
+    return allocate_large_and_trim(std::max(align_pages, n_pages), should_sample);
 }
 
 disable_backtrace_temporarily::disable_backtrace_temporarily()
@@ -1782,6 +1781,10 @@ inline void free(void* obj, S size = {}) {
 }
 
 void free_aligned(void* obj, size_t align, size_t size) {
+    if (align > page_size) {
+        free(obj);
+        return;
+    }
     if (size <= sizeof(free_object)) {
         size = sizeof(free_object);
     }
--- a/tests/unit/alloc_test.cc
+++ b/tests/unit/alloc_test.cc
@@ -114,6 +114,34 @@ SEASTAR_TEST_CASE(test_aligned_alloc) {
     return make_ready_future<>();
 }
 
+SEASTAR_TEST_CASE(test_large_alignment_larger_than_size) {
+    static constexpr size_t large_align = memory::huge_page_size < 512 * memory::page_size
+            ? memory::huge_page_size
+            : 512 * memory::page_size;
+    static constexpr size_t size = memory::page_size;
+    static_assert(large_align > size);
+
+    std::vector<void*> allocations;
+    allocations.reserve(4);
+    for (unsigned i = 0; i != 4; ++i) {
+        void* p = nullptr;
+        BOOST_REQUIRE_EQUAL(posix_memalign(&p, large_align, size), 0);
+        BOOST_REQUIRE(p != nullptr);
+        BOOST_REQUIRE_EQUAL(reinterpret_cast<uintptr_t>(p) % large_align, 0);
+        allocations.push_back(p);
+    }
+    for (auto p : allocations) {
+        free(p);
+    }
+
+    auto p = ::operator new(size, std::align_val_t(large_align));
+    BOOST_REQUIRE(p != nullptr);
+    BOOST_REQUIRE_EQUAL(reinterpret_cast<uintptr_t>(p) % large_align, 0);
+    ::operator delete(p, size, std::align_val_t(large_align));
+
+    return make_ready_future<>();
+}
+
 #ifdef __cpp_sized_deallocation
 SEASTAR_TEST_CASE(test_sized_delete) {
     for (size_t size = 0; size <= 65536; size++) {


src/core/memory.cc
```diff
@@ -602,7 +602,7 @@ struct cpu_pages {
         unsigned nr_pages;
     };
```
avikivity, line 603: > memory: honor large allocation alignment

missing changelog

src/core/memory.cc
```diff
@@ -824,27 +824,30 @@ void cpu_pages::maybe_reclaim() {
 }
 
 void*
-cpu_pages::allocate_large_and_trim(unsigned n_pages, bool should_sample) {
+cpu_pages::allocate_large_and_trim(unsigned span_pages, unsigned n_pages, bool should_sample) {
```
avikivity: Why have these two parameters? If someone requests a small allocation with large alignment (should never happen), then just round up the allocation to the alignment. Saves a future reader from having to understand similar-but-not-identical parameters.

src/core/memory.cc
dkropachev: Changelog: Fix allocator alignment for large aligned allocations where the requested alignment is larger than the requested size, including `posix_memalign()` and aligned C++ `new`.

avikivity: It needs to be in the patch changelog, visible with `git log`, not here.

src/core/memory.cc
dkropachev: Makes sense. I see two reasonable ways to make this easier to read:

1. Keep the single-size helper: call `allocate_large_and_trim(max(align_pages, n_pages))` for aligned large allocations, then immediately `shrink(ptr, n_pages * page_size)` when `align_pages > n_pages`. That keeps the “round up for alignment” model while still recording `posix_memalign(&p, 2 MiB, 4 KiB)` as a one-page allocation and returning the suffix pages.

2. Keep the current two-size helper, but rename the parameters and add a comment explaining the split: one size is the buddy span needed for alignment, the other is the allocation span we record for `object_size()`, heap profiling, `realloc()`, and `free_large()`.

I slightly prefer option 2 because it returns the suffix before `maybe_reclaim()` runs, while round-then-shrink briefly holds the larger span and may trigger reclaim under pressure. But either option is fine with me; if you prefer the single-size helper for readability, I’ll switch to that.

src/core/memory.cc
dkropachev: > It needs to be in the patch changelog, visible with `git log`, not here.

I just replied and not pushed, working on second item, did not expect you to reply that fast ))
Please check now.

src/core/memory.cc
avikivity: Why shrink? Nobody cares if you over-allocate.

This never happens in practice so the amount of memory wasted is (2MB - 4kB) * 0 == 0.

dkropachev: agreed, done.

src/core/memory.cc
```diff
@@ -837,11 +837,12 @@ cpu_pages::allocate_large_and_trim(unsigned n_pages, bool should_sample) {
     auto span_size = span->span_size;
     auto span_idx = span - pages;
     nr_free_pages -= span->span_size;
-    while (span_size >= n_pages * 2) {
+    while (span_size / 2 >= n_pages) {
```
avikivity: Why this change?

dkropachev: Good point, this is unrelated now. Reverted.

avikivity:
If you force-push, add a summary of what changed. I'm not going to reverse-engineer it.

dkropachev:
> If you force-push, add a summary of what changed. I'm not going to reverse-engineer it.

Pardon, updated

avikivity:
> > If you force-push, add a summary of what changed. I'm not going to reverse-engineer it.
> 
> Pardon, updated

What was updated?

dkropachev:
> > > If you force-push, add a summary of what changed. I'm not going to reverse-engineer it.
> > 
> > 
> > Pardon, updated
> 
> What was updated?

Oh, i misread your comment, last forcepush fixed exactly issue you pointed out [here](https://github.com/scylladb/seastar/pull/3472#pullrequestreview-4539994921)

One before did the following:
cpu_pages::allocate_large_and_trim():
  - Removed the separate “requested span size” argument.
  - The function now trims the buddy span to the requested size and keeps that whole span as the allocation.
  - Removed the extra logic that allocated a smaller part and freed the remaining tail with free_span_unaligned().

cpu_pages::allocate_large():
  - Updated the call to use the simpler allocate_large_and_trim() signature.

 cpu_pages::allocate_large_aligned():
  - Still requests max(alignment pages, allocation pages), so alignment larger than size is honored.
  - Updated the call to use the simpler allocate_large_and_trim() signature.

PR #3474, opened 2026-06-20
dns: drain c-ares socket monitor readiness

### LINKED ISSUE #3473: dns: socket-state c-ares path can stall split TCP responses
When Seastar is built with c-ares >= 1.34, enabling `ARES_OPT_SOCK_STATE_CB` puts the resolver on the socket-monitor path in `dns_resolver::impl::poll_sockets()`. That path currently calls `ares_process_fds()` at most once per wakeup.

The legacy path drains readiness in a loop. That matters for TCP DNS responses because `do_recvfrom()` may complete a synchronous `read_up_to()` and leave `POLLIN` set while c-ares still wants readability. If the TCP response is split across multiple reads, the one-shot socket-monitor path can process the first partial buffer and then return without arming another read. The remaining response bytes can sit unread until the periodic DNS timeout tick calls back into c-ares.

Expected behavior: the socket-monitor path should preserve the old drain-loop behavior, or otherwise immediately re-arm reads while c-ares still wants readability.

A focused regression test can reproduce this with a loopback TCP DNS server that sends the TCP length and one DNS payload byte, waits briefly, then sends the rest of a valid A response. Without draining monitored readiness, the resolver waits for the periodic timeout instead of completing promptly.

dkropachev:
Fixes #3473.

## Problem

With c-ares >= 1.34, `ARES_OPT_SOCK_STATE_CB` puts the resolver on the socket-monitor path in `dns_resolver::impl::poll_sockets()`. That path processed ready monitored sockets with a single `ares_process_fds()` call per wakeup, unlike the legacy path which drains readiness in a loop.

For TCP DNS responses split across reads, c-ares can consume a partial buffer while still wanting readability. If Seastar still has `POLLIN` set after that synchronous read, returning from `poll_sockets()` leaves the remaining response data unread until the periodic DNS timeout tick.

## Solution

Drain the socket-monitor path the same way as the legacy path: rebuild the ready event list and call `ares_process_fds()` until no requested read or write readiness remains. Preserve the timeout-only `ares_process_fd()` call when no sockets were processed.

The regression test uses a loopback TCP DNS server that sends the TCP length plus one DNS response byte, waits briefly, then sends the rest of a valid A response. The old one-shot path times out waiting for the periodic DNS tick; the fixed path completes promptly.

## Testing

- `ninja -C build/dev test_unit_dns`
- `./build/dev/tests/unit/dns_test --run_test=test_resolve_tcp_split_response`
- `./build/dev/tests/unit/dns_test --run_test=test_connection_refused_tcp`
- `./build/dev/tests/unit/dns_test --run_test=test_resolve_numeric`
- `git diff --check origin/master...HEAD`

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -751,7 +751,8 @@ dns_resolver::impl::poll_sockets() {
 #if ARES_VERSION >= 0x011300 && USE_CARES_EVENTFD
     // When using ARES_OPT_SOCK_STATE_CB with modern c-ares >= 1.34.0
     // we know exactly which sockets c-ares cares about through the callback
-    if (!_socket_monitors.empty()) {
+    bool processed = false;
+    for (;;) {
         // Most DNS queries use few sockets, so optimize for the common case by
         // stack-allocating a small buffer and only allocating on heap if needed
         constexpr int MAX_EVENTS = 16;
@@ -800,12 +801,12 @@ dns_resolver::impl::poll_sockets() {
 
         if (event_count > 0) {
             ares_process_fds(_channel, events, event_count, ARES_PROCESS_FLAG_NONE);
+            processed = true;
         } else {
-            // No sockets ready, just process timeouts
-            ares_process_fd(_channel, ARES_SOCKET_BAD, ARES_SOCKET_BAD);
+            break;
         }
-    } else {
-        // No sockets monitored, just handle timeouts
+    }
+    if (!processed) {
         ares_process_fd(_channel, ARES_SOCKET_BAD, ARES_SOCKET_BAD);
     }
 #elif USE_CARES_EVENTFD
--- a/tests/unit/dns_test.cc
+++ b/tests/unit/dns_test.cc
@@ -22,12 +22,14 @@
 #include <vector>
 #include <algorithm>
 
-#include <seastar/core/do_with.hh>
+#include <seastar/core/seastar.hh>
+#include <seastar/core/sleep.hh>
+#include <seastar/core/with_timeout.hh>
 #include <seastar/testing/test_case.hh>
 #include <seastar/core/sstring.hh>
 #include <seastar/core/reactor.hh>
-#include <seastar/core/do_with.hh>
 #include <seastar/core/when_all.hh>
+#include <seastar/net/api.hh>
 #include <seastar/net/dns.hh>
 #include <seastar/net/inet_address.hh>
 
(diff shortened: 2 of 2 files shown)


tests/unit/dns_test.cc
```diff
@@ -152,6 +232,35 @@ SEASTAR_TEST_CASE(test_connection_refused_tcp) {
...
+            BOOST_REQUIRE_EQUAL(h.addr_entries.size(), 1);
+            BOOST_REQUIRE_EQUAL(h.addr_entries.front().addr, inet_address("127.0.0.42"));
+            BOOST_REQUIRE_EQUAL(h.addr_entries.front().ttl, std::chrono::seconds(60));
+        }).finally([d] {
+            return d->close();
+        }).finally([&listener, server = std::move(server)] () mutable {
+            listener.abort_accept();
+            return std::move(server);
+        });
+    });
```
avikivity: Please rewrite as a coroutine,  it will be clearer.

src/net/dns.cc
```diff
@@ -800,12 +801,12 @@ dns_resolver::impl::poll_sockets() {
 
         if (event_count > 0) {
             ares_process_fds(_channel, events, event_count, ARES_PROCESS_FLAG_NONE);
```
avikivity, line 803: The manual suggests to add ARES_PROCESS_FLAG_SKIP_NON_FD, and let the final ares_process_fd() run unconditionally to process timeouts. Though what happens if a timeout asks for a notification on some socket? I guess it's safer to keep as is.

avikivity (approved):
Looks good, waiting for @elcallio review too.

src/net/dns.cc
dkropachev: I think it is safer to keep this as-is.

`ARES_PROCESS_FLAG_SKIP_NON_FD` would defer timeout/cleanup handling while we drain fd readiness. In this path the important bit is that after each `ares_process_fds()` call we rebuild the event list from Seastar's current readiness and c-ares' current socket-interest state. If the deferred non-fd pass creates, closes, or changes interest for a socket, doing it only after the loop has already broken could leave immediately-ready work unprocessed until a later timer wakeup.

The current code keeps timeout/cleanup processing inside the same drain loop and only falls back to `ares_process_fd(...BAD...)` when no fd event was processed, matching the legacy path's conservative behavior.

tests/unit/dns_test.cc
dkropachev: good idea, done.

src/net/dns.cc
dkropachev: At the same time I see that both options quite close and don't see any serious problem with @avikivity proposal

elcallio (approved):

dkropachev:
@avikivity , @elcallio , can we merge it ?

PR #3480, opened 2026-06-24
`core/reactor`: add `SIGFPE` handler

WillemKauf:
`SIGFPE` is a notable & potentially common enough error that it deserves a handler in `seastar` - divisions by zero (including % 0 operations), integer overflows, etc., are commonplace bugs for developers to run into.

Add one alongside the existing signal handlers to ensure we print a backtrace for aiding developers in debugging before crashing.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4303,6 +4303,11 @@ static void sigill_action(siginfo_t *info, ucontext_t* uc) noexcept {
     reraise_signal(SIGILL);
 }
 
+static void sigfpe_action(siginfo_t *info, ucontext_t* uc) noexcept {
+    print_with_backtrace("Floating point exception");
+    reraise_signal(SIGFPE);
+}
+
 // We don't need to handle SIGSEGV when asan is enabled.
 #ifdef SEASTAR_ASAN_ENABLED
 template<>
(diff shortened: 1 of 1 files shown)


WillemKauf:
@scylladb/seastar-maint PTAL if you get the chance :slightly_smiling_face:

src/core/reactor.cc
```diff
@@ -4320,6 +4303,11 @@ static void sigill_action(siginfo_t *info, ucontext_t* uc) noexcept {
     reraise_signal(SIGILL);
 }
```
avikivity, line 4304: > `core/reactor`: add `SIGFPE` handler


Please avoid the gratuitous quoting.

src/core/reactor.cc
```diff
@@ -4320,6 +4303,11 @@ static void sigill_action(siginfo_t *info, ucontext_t* uc) noexcept {
...
 
+static void sigfpe_action(siginfo_t *info, ucontext_t* uc) noexcept {
+    print_with_backtrace("Floating point exception");
+    reraise_signal(SIGFPE);
+}
+
```
avikivity, line 4310: > Add one alongside the existing signal handlers to ensure we print a backtrace

Wouldn't you recover it from the coredump?

travisdowns (approved):

src/core/reactor.cc
travisdowns: What core dump? We're lucky if the customer even provides us the tail of the log without truncating it somehow. Successfully hand-holding a core dump through k8s etc is roughly in between brain surgery and rocket science in complexity.

src/core/reactor.cc
WillemKauf: Removed

src/core/reactor.cc
WillemKauf: > Wouldn't you recover it from the coredump?

Even putting aside the annoyances Travis mentioned above, isn't it just nice to have a backtrace immediately available in logs for this type of failure? :slightly_smiling_face:

travisdowns:
@scylladb/seastar-maint I'm inclined to merged this unless I get a nack.

src/core/reactor.cc
travisdowns: @avikivity are these answers satisfactory?

PR #3481, opened 2026-06-25
httpd: Remove deprecated set_tls_credentials() method

xemul:
Deprecated in 2023 by 024e8a3b05bc062506fc8b1729f5417a2e2951be (httpd: added ability to configure different credentials per listener)


xemul:

PR #3483, opened 2026-06-25
iostream: Deprecate consumption_result(optional<buffer>) constructor

xemul:
The user is supposed to construct it with explicit stop_consuming or continue_consuming types. This "shortcut" ("sugar"?) had been deprecated with a comment since it was introduced, but it's better to deprecate it explicitly.


xemul:

PR #3484, opened 2026-06-25
test/iostream: Fix test_move_after_batched_write deprecation usage

xemul:
The test uses two deprecated features of output_stream -- default constructor and move-assignment operator like

    output_stream<char> out; // uses deprecated default constructor
    out = std::move(out2);   // uses deprecated operator=(&&)

Since the goal of the test (and the fix that this test tests, see PR #3464) was to validate the moved-from state of out2 the test should use move constructor instead


xemul:
Cc @raphaelroshan

xemul:

PR #3486, opened 2026-06-25
print: Remove deprecated fmt_print()

xemul:
It was deprecated in 2021 by commit a189cdc4 (print: deprecate print() family)


xemul:

PR #3487, opened 2026-06-26
reactor_backend: include <span>

tchaikov:
std::span is used in this header. Since a header should be self-contained and not rely on transitive includes, include <span> explicitly.


tchaikov:
the test failure at https://github.com/scylladb/seastar/actions/runs/28214372297/job/83582166380 should be addressed by #3489

PR #3490, opened 2026-06-26
CMakeLists: link URING::uring as PUBLIC so consumers find liburing.h

tchaikov:
Building Ceph against seastar fails to find liburing.h. Ceph builds liburing out of tree, so the header is off the default search path.

`reactor_config.hh` embeds `::io_uring` in a public struct under `SEASTAR_HAVE_URING`, so any consumer of the seastar headers needs `liburing.h` once that macro is set. `SEASTAR_HAVE_URING` is `PUBLIC`, but `URING::uring` was linked PRIVATE, so its include directory never reached consumers. Link it `PUBLIC` so the include directory propagates with the macro.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1143,7 +1143,7 @@ if (Seastar_IO_URING)
   target_compile_definitions (seastar
     PUBLIC SEASTAR_HAVE_URING)
   target_link_libraries (seastar
-    PRIVATE URING::uring)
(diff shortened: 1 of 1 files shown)


tchaikov:
an alternative is to make `::io_uring` a (smart) pointer in `asymmetric_uring`, or to use pimp, but that'd be a more intrusive change instead of a minimal fix. so i'd defer it to a follow-up change.

PR #3491, opened 2026-06-26
core/reactor: don't read the thread-local shard count in smp::qs_deleter

tchaikov:
`smp::qs_deleter::operator()` sizes its destruction loops with `this_smp_shard_count()`, which dereferences the `thread_local _this_smp`.  The deleter runs from `~smp()`, but `_this_smp` is only set on reactor threads, so it is null when the smp is destroyed on another thread.  That happens when the reactor runs on a dedicated thread and the `app_template` is destroyed elsewhere, for example at exit on the main thread; the deleter then dereferences null and faults under ASan/UBSan.

The queue array is allocated from `_shard_count`, so store that count in the deleter and use it at teardown instead of querying the thread-local.

### DIFF
--- a/include/seastar/core/smp.hh
+++ b/include/seastar/core/smp.hh
@@ -319,6 +319,7 @@ class smp : public std::enable_shared_from_this<smp> {
     std::optional<std::barrier<>> _all_event_loops_done;
     std::unique_ptr<internal::memory_prefaulter> _prefaulter;
     struct qs_deleter {
+      unsigned shard_count;
       void operator()(smp_message_queue** qs) const;
     };
     std::unique_ptr<smp_message_queue*[], qs_deleter> _qs_owner;
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -4329,8 +4329,8 @@ void install_oneshot_signal_handler<SIGSEGV, sigsegv_action>() {
 #endif
 
 void smp::qs_deleter::operator()(smp_message_queue** qs) const {
-    for (unsigned i = 0; i < this_smp_shard_count(); i++) {
-        for (unsigned j = 0; j < this_smp_shard_count(); j++) {
+    for (unsigned i = 0; i < shard_count; i++) {
+        for (unsigned j = 0; j < shard_count; j++) {
             qs[i][j].~smp_message_queue();
         }
         ::operator delete[](qs[i], std::align_val_t(alignof(smp_message_queue))
@@ -4739,7 +4739,7 @@ void smp::configure(const smp_options& smp_opts, const reactor_options& reactor_
 
     seastar_logger.info("Reactor backend: {}", backend_selector);
 
-    _qs_owner = decltype(smp::_qs_owner){new smp_message_queue* [_shard_count], qs_deleter{}};
+    _qs_owner = decltype(smp::_qs_owner){new smp_message_queue* [_shard_count], qs_deleter{_shard_count}};
 
     auto allocate_qs_owner = [this] (unsigned i) {
         // smp_message_queue has members with hefty alignment requirements.


xemul:
@tchaikov , can you shed more light on the 

> That happens when the reactor runs on a dedicated thread and the app_template is destroyed elsewhere

statement? What's the way the app-template is maintained so that it's destroyed "elsewhere"?

tchaikov:
@xemul hi Pavel, this is a minimal reproducer:
```c++
#include <seastar/core/alien.hh>
#include <seastar/core/posix.hh>

#include <sys/eventfd.h>
#include <thread>
#include <iostream>
#include <memory>

// A pared-down SeastarRunner with the same ownership/threading shape.
struct runner {
    seastar::app_template app;
    seastar::file_desc begin_fd = seastar::file_desc::eventfd(0, 0);
    std::unique_ptr<seastar::readable_eventfd> on_end;
    std::thread thread;

    void init(int argc, char** argv) {
        // The reactor -- and the thread_local _this_smp -- lives ONLY on this thread.
        thread = std::thread([this, argc, argv] {
            app.run(argc, argv, [this] {
                on_end = std::make_unique<seastar::readable_eventfd>();
                ::eventfd_write(begin_fd.get(), 1);                 // "reactor is up"
                return on_end->wait().then([](size_t) {})
                    .finally([this] { on_end.reset(); });
            });
        });
        eventfd_t v;
        ::eventfd_read(begin_fd.get(), &v);                         // block until reactor is up
    }

    void stop() {
        // Ask the reactor to return from app.run(), then join its thread.
        seastar::alien::submit_to(app.alien(), 0, [this] {
            on_end->write_side().signal(1);
            return seastar::make_ready_future<>();
        }).get();
        thread.join();
    }
};

    runner r;                 // app_template CONSTRUCTED on the main thread
    r.init(argc, argv);       // app.run() executes on r.thread (the reactor thread)
    r.stop();                 // reactor returns; r.thread joined -> reactor thread gone
    std::cout << "reactor stopped; destroying app_template on the main thread...\n";
    // `r` (and its app_template) is DESTROYED here, on the main thread:
    //   ~app_template -> ~smp -> qs_deleter -> this_smp_shard_count() on null _this_smp.
    return 0;
}
```
in Ceph's crimson testing harness, we run `app_template` on a dedicated thread, where `_this_smp` is set), but `app_template` instance is owned and destroyed on the main thread, that's why `~smp` -> `qs_deleter`  -> reads the null `_this_smp`.

the above is a minimal reproducer, the full code can be found at https://github.com/ceph/ceph/blob/main/src/test/crimson/gtest_seastar.cc and https://github.com/ceph/ceph/blob/main/src/test/crimson/seastar_runner.h

avikivity:
I see, so app_template's constructor and destructor run on one thread, but app_template::run() runs on another?

It's odd, but I don't mind accommodating it.

travisdowns:
ping @tchaikov see Avi's question above.

tchaikov:
@travisdowns hi Travis, thanks for your ping!

@avikivity sorry Avi, i missed your question. 

tl;dr: yes, that's right. 

longer explanation: the `app_template` is a member of a `runner` that lives on the main thread, so it's constructed and destroyed there, but `run()` is called from a separate thread the runner spawns for the reactor. `_this_smp` only gets set on that spawned thread, so when the runner goes away on the main thread, `~smp` (and the `qs_deleter`) runs where `_this_smp` is null. storing `shard_count` in the `deleter` drops that thread-local read, so teardown works no matter which thread it runs on.

avikivity:
I think it's a fragile setup.

However, it's in general better to use locals instead of globals, so the patch is an improvement, and I don't mind supporting non-standard setups as long as there is no pain involved in it.

PR #3492, opened 2026-06-27
rpc: optimize usage of unordered_map

avikivity:
Avoid using operator[] for lookups, as it creates the entry if it is not found. Note that if we did create the entry, it would be null, and we would have crashed, so this is a safe transformation.

Reuse the looked-up iterator for erasing, avoiding a second lookup.

Use try_emplace() instead of a check-and-insert sequence, combining the two lookups.

Note: I don't expect significant performance improvements, but every small reduction in instruction count helps.


gleb-cloudius (approved):

avikivity:

nyh (approved):

PR #3493, opened 2026-06-28
reactor: don't backtrace during seastar thread switches

calebxyz:
The stall detector takes a backtrace from a signal handler. If the signal lands while a seastar::thread is switching stacks via longjmp(), the interrupted state may not be unwindable. On aarch64/X86 this can crash in libgcc's unwinder.

Track the longjmp switch window and omit the stall backtrace while it is active, similar to the existing guard for exception unwinding. The stall is still reported.

Fixes: https://scylladb.atlassian.net/browse/SCYLLADB-2908

### DIFF
--- a/include/seastar/core/thread_impl.hh
+++ b/include/seastar/core/thread_impl.hh
@@ -74,6 +74,7 @@ void yield();
 void switch_in(thread_context* to);
 void switch_out(thread_context* from);
 void init();
+bool is_context_switch_in_progress() noexcept;
 
 }
 }
--- a/src/core/thread.cc
+++ b/src/core/thread.cc
@@ -33,6 +33,7 @@
 #include <valgrind/valgrind.h>
 #include <exception>
 #include <utility>
+#include <atomic>
 #include <boost/intrusive/list.hpp>
 
 #include <seastar/core/thread.hh>
@@ -46,6 +47,10 @@ namespace seastar {
 
 thread_local jmp_buf_link g_unthreaded_context;
 thread_local jmp_buf_link* g_current_context;
+
+namespace {
+thread_local std::atomic_flag g_context_switch_in_progress{};
+}
 
 #ifdef SEASTAR_ASAN_ENABLED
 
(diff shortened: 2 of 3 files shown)


src/core/thread.cc
```diff
@@ -333,7 +356,13 @@ void switch_out(thread_context* from) {
     from->switch_out();
 }
 
+bool is_context_switch_in_progress() noexcept {
+        std::atomic_signal_fence(std::memory_order_seq_cst);
+        return g_context_switch_in_progress.test(std::memory_order_relaxed);
+}
+
```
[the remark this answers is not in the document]
calebxyz: fixed, moved the flag to a position where both are valid

avikivity:
> The stall detector takes a backtrace from a signal handler. If the signal lands while a seastar::thread is switching stacks via longjmp(), the interrupted state may not be unwindable. On aarch64/PAC this can crash in libgcc's unwinder.

Need a link to a libunwinder bug report.

> 
> Track the longjmp switch window and omit the stall backtrace while it is active, similar to the existing guard for exception unwinding. The stall is still reported.
> 
> Fixes: https://scylladb.atlassian.net/browse/SCYLLADB-2908


A seastar commit cannot fix anything in scylladb, it's the submodule update that actually fixes things.

include/seastar/core/thread_impl.hh
```diff
@@ -74,6 +74,7 @@ void yield();
 void switch_in(thread_context* to);
 void switch_out(thread_context* from);
```
avikivity, line 75: > may not be unwindable. On aarch64/PAC this can crash in libgcc's unwinder.

What is PAC?

calebxyz: Pointer Authentication Code -  return addresses and function pointers on ARM

calebxyz:
> > The stall detector takes a backtrace from a signal handler. If the signal lands while a seastar::thread is switching stacks via longjmp(), the interrupted state may not be unwindable. On aarch64/PAC this can crash in libgcc's unwinder.
> 
> Need a link to a libunwinder bug report.
> 
> > Track the longjmp switch window and omit the stall backtrace while it is active, similar to the existing guard for exception unwinding. The stall is still reported.
> > Fixes: https://scylladb.atlassian.net/browse/SCYLLADB-2908
> 
> A seastar commit cannot fix anything in scylladb, it's the submodule update that actually fixes things.

got it, i have open this tickets so i would have something to link to, should i open a github issue on seastar project and change this ?

avikivity:
> > > The stall detector takes a backtrace from a signal handler. If the signal lands while a seastar::thread is switching stacks via longjmp(), the interrupted state may not be unwindable. On aarch64/PAC this can crash in libgcc's unwinder.
> > 
> > 
> > Need a link to a libunwinder bug report.
> > > Track the longjmp switch window and omit the stall backtrace while it is active, similar to the existing guard for exception unwinding. The stall is still reported.
> > > Fixes: https://scylladb.atlassian.net/browse/SCYLLADB-2908
> > 
> > 
> > A seastar commit cannot fix anything in scylladb, it's the submodule update that actually fixes things.
> 
> got it, i have open this tickets so i would have something to link to, should i open a github issue on seastar project and change this ?

It's not strictly required (but is good practice).

calebxyz:
@avikivity i looked at GCC Bugzilla 
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71890 
this is a know issue that they will not fix

calebxyz:
tested against https://github.com/fee-mendes/seastar/commit/a2a421471ff2c2ce8473ddfbe46f05f56e12e5f8
the crush is fixed, in the logs i can see 
`Reactor stalled for 1 ms, backtrace omitted (context switch in progress) `
and the test keeps running

calebxyz:
@avikivity please consider merging

include/seastar/core/thread_impl.hh
avikivity: Are you sure it's related?

calebxyz: well, not directly but because it reproduces in X86 as well, this is not strictly PAC related, its more bad address used so i will remove it from the message

dorlaor:
Cheers Alex and Felipe!

PR #3496, opened 2026-06-29
dns: opt out of c-ares RFC 6724 address sorting

### LINKED ISSUE #3456: dns: buggy address sorting / c-ares integration
While trying to improve some of the tcp connection machinery in my codebase I found some surprising behavior's in Seastar's dns resolver.

c-ares then tries to sort every multi-address DNS result ([RFC 6724: Default Address Selection for Internet Protocol Version 6 (IPv6)](https://datatracker.ietf.org/doc/html/rfc6724)) by probing each address:
1. Open a UDP socket,
2. connect to it,
3. read back the source address.

However, because of how Seastar sets [socket hooks](https://github.com/scylladb/seastar/blob/8a69e1f4cf41dc00a149484c176e270fe52781a6/src/net/dns.cc#L409) makes this impossible to work:
  - do_socket ignores the address family and always creates an AF_INET channel
  - connecting a probe to an IPv6 address throws `std::invalid_argument (No ipv6 yet)`
  - connecting to an IPv4 address is a no-op that "succeeds"
  - either way c-ares then needs agetsockname, which Seastar sets to nullptr — c-ares treats that as fatal and abandons the sort.

The observed behavior is that every multi-address lookup returns addresses in DNS-response order (including getting ipv6 as a first address even if it's not routable). On top of that, a datagram socket is created and destroyed for absolutely no reason.

I don't have enough insight to suggest the right fix. There are some tradeoffs to be made I guess. Worst case, we can explicitly set `ARES_AI_NOSORT` to opt-out of the sorting behavior.

tchaikov:
Multi-address lookups come back in DNS-response order, sometimes with a non-routable IPv6 address first, and each one needlessly creates and destroys a datagram socket.

c-ares sorts getaddrinfo() results per RFC 6724 by probing each address: it opens a UDP socket, connects, and reads back the source with getsockname. Our socket hooks cannot run that probe. do_socket ignores the address family, an IPv6 probe connect throws, and we install no getsockname callback, so c-ares gives up on the sort after creating and closing the socket. Set ARES_AI_NOSORT so it skips the sort rather than failing it.

Implementing the probe does not work in our model. A getsockname callback cannot be limited to sort probes: c-ares also calls it on every connecting query socket to read the local binding. Seastar's connect() is async, so the connected_socket and its local_address() do not exist until the connection completes, and returning an error or a wildcard with port 0 makes c-ares treat the socket as dead and drop the query. Calling the sort directly is not an option either: c-ares does not export ares_sortaddrinfo(), and ares_set_sortlist() applies only to the deprecated gethostbyname() path. So NOSORT is the only knob c-ares gives us.

A caller that needs routable ordering can do that at the connect layer instead: resolve the full address list and dial each address in turn until one connects, as redpanda does in redpanda-data/redpanda#30796. That only partially matches c-ares's sort, but a robust client needs it regardless, since ordering alone never guarantees the first address is reachable.

Closes scylladb/seastar#3456

### DIFF
--- a/include/seastar/net/dns.hh
+++ b/include/seastar/net/dns.hh
@@ -98,6 +98,11 @@ struct srv_record {
  * stack of choice, though for "normal" non-test
  * querying, you are probably better of with the
  * global calls further down.
+ *
+ * \note Addresses come back in DNS-response order, not sorted per RFC 6724.
+ * c-ares's sort probes each address through a connected socket; this stack
+ * can't, so it is disabled. A caller that needs a routable address should dial
+ * the returned addresses in turn until one connects.
  */
 class dns_resolver {
 public:
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -547,8 +547,15 @@ dns_resolver::impl::get_host_by_name(sstring name, opt_family family)  {
 // in Gcc 11 (see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96003).
 #pragma GCC diagnostic ignored "-Wnonnull"
 
+    // c-ares would sort multi-address results per RFC 6724, but that needs a
+    // getsockname callback we cannot provide; opt out. See scylladb/seastar#3456.
+    int ai_flags = ARES_AI_CANONNAME;
+#ifdef ARES_AI_NOSORT
+    ai_flags |= ARES_AI_NOSORT;
+#endif
+
     ares_addrinfo_hints hints = {
-        .ai_flags = ARES_AI_CANONNAME,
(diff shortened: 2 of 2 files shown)


tchaikov:
cc @nvartolomei

nvartolomei:
Seems reasonable. @tchaikov, should we add a note about this behavior/the fact that sorting is left to the caller on module/class?

tchaikov:
> Seems reasonable. @tchaikov, should we add a note about this behavior/the fact that sorting is left to the caller on module/class?

good idea, added.

PR #3498, opened 2026-06-30
net: fix abandoned failed future in TLS session handshake

### LINKED ISSUE #3497: TLS session: ~session() does not drain _output_pending, causing "Exceptional future ignored" on connection teardown
## Summary

The TLS `session` destructor in `src/net/tls_gnutls.cc` can trigger an "Exceptional future ignored" warning when its `_output_pending` member holds a failed future (e.g. `EPIPE` / Broken pipe) at destruction time. Nobody consumes the exception during the close/destruction lifecycle.

This was found while investigating SCYLLADB-2906.

## Root Cause

The `session` class has a `future<> _output_pending` member that tracks the state of the last socket write initiated by GnuTLS's `vec_push()` callback. When a write fails (e.g. peer closes the connection during TLS handshake), `vec_push()` stores the error:

```cpp
// tls_gnutls.cc, vec_push():
    gnutls_transport_set_errno(*this, e.code().value());
    _output_pending = make_exception_future<>(std::current_exception()); // EPIPE stored here
}
```

The destructor only asserts availability but never consumes the exception:

```cpp
~session() {
    SEASTAR_ASSERT(_output_pending.available()); // passes — it IS resolved (with an exception)
    // implicit member destruction: _output_pending's destructor triggers the warning
}
```

Neither the `session::close()` background task nor any other code path drains `_output_pending` before the session is destroyed.

## Destruction Chain (from decoded backtrace)

The full chain that leads to the warning:

```
reactor::run_tasks()
  └─ continuation::run_and_dispose()       — a continuation completes, then `delete this`
       └─ ~continuation()                   — destroys captured state
            └─ put_connection(...)::λ::~λ() — the [con]{} lambda from con->close().finally([con]{})
                 └─ ~shared_ptr<connection>  — refcount → 0
                      └─ ~connection()       — HTTP connection members destroyed
                           └─ ~connected_socket()
                                └─ ~tls_connected_socket_impl()
                                     └─ ~session_ref() → ~shared_ptr<session_impl>
                                          └─ ~session()             ← TLS session destroyed
                                               └─ ~future<>         ← _output_pending holds Broken pipe!
                                                    └─ check_failure() → report_failed_future()
```

## Reproduction Scenario

1. `vector_search::client` makes an HTTPS request to a host with a certificate mismatch
2. During TLS handshake, GnuTLS calls `vec_push()` to send handshake data
3. The write fails with EPIPE (peer closed connection) → stored in `_output_pending`
4. The TLS verification error is surfaced as the request-level failure
5. The HTTP client calls `put_connection(con)` → `con->close().finally([con]{})`
6. After `close()` completes and the continuation is destroyed, the connection is dropped
7. Destruction cascades to `session::~session()` with `_output_pending` still holding EPIPE
8. Seastar reports "Exceptional future ignored"

### Observed symptom

```
WARN seastar - Exceptional future ignored: std::system_error (error system:32, sendmsg: Broken pipe), backtrace: ...
```

Followed by:
```
WARN vector_store_client - Request to vector store wrong.scylla 127.0.0.1:39041 failed: seastar::tls::verification_error (...)
```

## Suggested Fix

Drain `_output_pending` in the session destructor:

```cpp
~session() {
    SEASTAR_ASSERT(_output_pending.available());
    _output_pending.ignore_ready_future();
}
```

The exception is safe to ignore because:
- The session is shutting down
- The actual error (e.g. `tls::verification_error`) was already surfaced to the caller at the application level
- The underlying socket is being closed

Alternatively, the `close()` background task could drain it after acquiring `_out_sem`:

```cpp
return with_semaphore(_in_sem, 1, [this] {
    return with_semaphore(_out_sem, 1, [this] {
        _output_pending.ignore_ready_future();
    });
});
```

## Notes

- Related to #2661 which fixed similar abandoned futures from `_read_buf.close()` / `_write_buf.close()` in the HTTP client
- The `connection::close()` code already suppresses exceptions from stream close via `.handle_exception()`, and from `_closed` — but none of that reaches the TLS session's internal `_output_pending`
- This is a cosmetic/noise issue — no data loss or incorrect behavior results

knowack1:
When gnutls handshake processing throws - for example when a DN verification callback aborts the connection — the catch block in `do_handshake_invoke()` returned the exception immediately and skipped the trailing `wait_for_output()` that drains `_output_pending`.

If a deferred write was still in flight, `_output_pending` later resolved into a ready, failed future that nothing consumed, which surfaced as an "abandoned failed future" warning when the TLS session was destroyed.

This change drains `_output_pending` in the catch block (waiting for the pending output and ignoring its failed result) before propagating the original handshake error, so the failed future is always consumed. A regression test reproduces the leak using deferred writes and a DN callback that aborts the connection mid-handshake.

Fixes: #3497

### DIFF
--- a/src/net/tls_gnutls.cc
+++ b/src/net/tls_gnutls.cc
@@ -650,7 +650,13 @@ public:
             // make sure we reset output_pending
             return wait_for_output();
         } catch (...) {
-            return make_exception_future<>(std::current_exception());
+            auto ep = std::current_exception();
+            return wait_for_output().then_wrapped(
+                [this, ep = std::move(ep)](future<> f) {
+                  f.ignore_ready_future();
+                  _error = ep;
+                  return make_exception_future<>(ep);
+                });
         }
     }
     future<> do_handshake() {
--- a/tests/unit/tls_test.cc
+++ b/tests/unit/tls_test.cc
@@ -35,6 +35,7 @@
 #include <seastar/core/iostream.hh>
 #include <seastar/core/when_all.hh>
 #include <seastar/core/with_timeout.hh>
+#include <seastar/util/later.hh>
 #include <seastar/util/std-compat.hh>
 #include <seastar/util/process.hh>
 #include <seastar/net/tls.hh>
@@ -2433,4 +2434,112 @@ SEASTAR_THREAD_TEST_CASE(test_send_two_large) {
     });
 
     seastar::when_all(std::move(sender), std::move(receiver)).discard_result().get();
+}
+
+// Reproduces an abandoned-failed-future leak in tls::session::~session(). #3497
+//
+// The reproduction path:
+//  1. The server issues a write during the handshake. deferred_sink_impl delays
+//     the write by one reactor tick, so the session holds a not-yet-resolved
+//     future in _output_pending.
+//  2. The handshake layer reports success and immediately invokes the registered
+//     DN verification callback, while _output_pending is still pending.
+//  3. The DN callback aborts the server's TX buffer and throws. verify() lets
+//     the exception escape do_handshake_invoke() through its catch block, so the
+//     trailing wait_for_output() call (the one commented "make sure we reset
+//     output_pending") is skipped and _output_pending is never drained.
+//  4. Asynchronously, the deferred write resolves with EPIPE because the buffer
+//     was aborted. _output_pending is now a ready, failed future — but nothing
+//     drains it because the wait_for_output() step was skipped in step 3.
+//  5. The session is destroyed while still holding that failed future, which
+//     triggers the "abandoned failed future" warning.
+SEASTAR_THREAD_TEST_CASE(test_output_pending_exception_on_destroy) {
+    static const sstring tls12_priority = "NORMAL:-VERS-TLS-ALL:+VERS-TLS1.2";
+
+    tls::credentials_builder cb;
+    cb.set_x509_key_file(certfile("test.crt"), certfile("test.key"), tls::x509_crt_format::PEM).get();
+    cb.set_x509_trust_file(certfile("catest.pem"), tls::x509_crt_format::PEM).get();
+    cb.set_priority_string(tls12_priority);
+    cb.set_dh_level();
+    auto client_creds = cb.build_certificate_credentials();
+
+    tls::credentials_builder sb;
+    sb.set_x509_key_file(certfile("test.crt"), certfile("test.key"), tls::x509_crt_format::PEM).get();
+    sb.set_x509_trust_file(certfile("catest.pem"), tls::x509_crt_format::PEM).get();
+    sb.set_priority_string(tls12_priority);
+    sb.set_client_auth(tls::client_auth::REQUIRE);
+    sb.set_dh_level();
+    auto server_creds = sb.build_server_credentials();
+
+    // Defers every write by seastar::yield(). Without the yield the loopback
+    // buffer's put() may complete synchronously (ready future), making
+    // _output_pending available before gnutls_handshake() returns and before
+    // verify() runs. In that case do_handshake_invoke() sees an already-resolved
+    // _output_pending and drains it correctly, so the bug is not triggered and
+    // the test becomes flaky.
+    class deferred_sink_impl : public data_sink_impl {
+        data_sink _sink;
+
+    public:
+        explicit deferred_sink_impl(data_sink sink)
+            : _sink(std::move(sink)) {
+        }
+
+        future<> put(std::span<temporary_buffer<char>> bufs) override {
+            std::vector<temporary_buffer<char>> owned(std::make_move_iterator(bufs.begin()), std::make_move_iterator(bufs.end()));
+            return seastar::yield().then([this, owned = std::move(owned)]() mutable {
+                return _sink.put(std::move(owned));
+            });
+        }
+
+        future<> flush() override {
+            return _sink.flush();
+        }
+
+        future<> close() override {
+            return _sink.close();
+        }
+
+        size_t buffer_size() const noexcept override {
+            return _sink.buffer_size();
+        }
+    };
+
+    class deferred_socket_impl : public loopback_connected_socket_impl {
+    public:
+        using loopback_connected_socket_impl::loopback_connected_socket_impl;
+        data_sink sink() override {
+            return data_sink(std::make_unique<deferred_sink_impl>(loopback_connected_socket_impl::sink()));
+        }
+    };
+
+    auto b1 = ::make_lw_shared<loopback_buffer>(nullptr, loopback_buffer::type::SERVER_TX);
+    auto b2 = ::make_lw_shared<loopback_buffer>(nullptr, loopback_buffer::type::CLIENT_TX);
+
+    auto ssi = std::make_unique<deferred_socket_impl>(b1, b2);
+    auto csi = std::make_unique<loopback_connected_socket_impl>(b2, b1);
+
+    server_creds->set_dn_verification_callback([b1](tls::session_type, sstring, sstring) {
+        // Abort the server's TX buffer while the verification callback is
+        // running so that the deferred write already in flight resolves with
+        // an error once control returns to the reactor.
+        b1->abort();
+        throw tls::verification_error("aborting handshake from DN callback");
+    });
+
+    auto ss = tls::wrap_server(server_creds, connected_socket(std::move(ssi))).get();
+    auto cs = tls::wrap_client(client_creds, connected_socket(std::move(csi)), tls::tls_options{.server_name = "test.scylladb.org"}).get();
+
+    auto strms = ::make_lw_shared<streams>(std::move(cs));
+
+    // Start writing on the client side.
+    auto client_loop = strms->out.write(temporary_buffer<char>(10))
+                               .then([strms] {
+                                   return strms->out.flush();
+                               })
+                               .handle_exception([strms](std::exception_ptr) {});
+
+    // Reading on the server side drives the handshake, triggering the DN
+    // verification callback, which aborts the server's TX buffer and throws.
+    BOOST_REQUIRE_THROW(ss.input().read().get(), tls::verification_error);
 }
--- a/tests/unit/tls_test.cc
+++ b/tests/unit/tls_test.cc
@@ -2527,19 +2527,23 @@ SEASTAR_THREAD_TEST_CASE(test_output_pending_exception_on_destroy) {
         throw tls::verification_error("aborting handshake from DN callback");
     });
 
-    auto ss = tls::wrap_server(server_creds, connected_socket(std::move(ssi))).get();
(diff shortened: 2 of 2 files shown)


kreuzerkrieg (approved):
lgtm

knowack1:
The CI failure looks unrelated to this change: https://github.com/scylladb/seastar/actions/runs/28437784404/job/84267837035?pr=3498. Let me double-check, though—I couldn't find a reported issue against this test

knowack1:
> The CI failure looks unrelated to this change: https://github.com/scylladb/seastar/actions/runs/28437784404/job/84267837035?pr=3498. Let me double-check, though—I couldn't find a reported issue against this test

It looks unrelated. Opened a dedicated ticket: https://github.com/scylladb/seastar/issues/3499.

xemul:
Can you trace the posix sockets calls and show how exactly the _closed future resolves into exception? @knowack1

knowack1:
> Can you trace the posix sockets calls and show how exactly the _closed future resolves into exception? @knowack1

I tried to find the exact path that could lead to this exception in the `_close` future.
However, based on the reactor's code, it seems it can never be resolved to such an exceptional future.
Hence, my root-cause assumption was wrong, and this fix does not solve it.

For this reason, I took a step back.
Initially, I couldn't decode the backtrace - Jenkins build was already wiped; however, I managed to recover it from the downloaded artifacts.
The decoded backtrace reveals that the problem was caused by the following future: tls::session::_output_pending (see: https://github.com/scylladb/seastar/issues/3497).

I will update this PR accordingly.

tests/unit/tls_test.cc
```diff
@@ -2434,3 +2434,111 @@ SEASTAR_THREAD_TEST_CASE(test_send_two_large) {
...
+        throw tls::verification_error("aborting handshake from DN callback");
+    });
+
+    auto ss = tls::wrap_server(server_creds, connected_socket(std::move(ssi))).get();
+    auto cs = tls::wrap_client(client_creds, connected_socket(std::move(csi)), tls::tls_options{.server_name = "test.scylladb.org"}).get();
+
+    auto strms = ::make_lw_shared<streams>(std::move(cs));
+
+    // Start writing on the client side.
+    auto client_loop = strms->out.write(temporary_buffer<char>(10))
```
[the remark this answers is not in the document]
knowack1: No warning is issued because client_loop is not trivially destructible.

tests/unit/tls_test.cc
```diff
@@ -2434,3 +2434,111 @@ SEASTAR_THREAD_TEST_CASE(test_send_two_large) {
...
+        explicit deferred_sink_impl(data_sink sink)
+            : _sink(std::move(sink)) {
+        }
+
+        future<> put(std::span<temporary_buffer<char>> bufs) override {
+            std::vector<temporary_buffer<char>> owned(std::make_move_iterator(bufs.begin()), std::make_move_iterator(bufs.end()));
+            return seastar::yield().then([this, owned = std::move(owned)]() mutable {
+                return _sink.put(std::move(owned));
+            });
+        }
```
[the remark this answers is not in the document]
knowack1: Code compiles.

knowack1:
@kreuzerkrieg @xemul  Could you please re-review? After a more careful root-cause investigation, the fix has been moved to a different (and correct) place. I also tried to create a reasonable reproduction unit test.

kreuzerkrieg:
> @kreuzerkrieg @xemu Could you please re-review? After a more careful root-cause investigation, the fix has been moved to a different (and correct) place. I also tried to create a reasonable reproduction unit test.

I guess you wanted to page @xemul :)

tests/unit/tls_test.cc
```diff
@@ -2434,3 +2435,111 @@ SEASTAR_THREAD_TEST_CASE(test_send_two_large) {
...
+// The reproduction path:
+//  1. The server issues a write during the handshake. deferred_sink_impl delays
+//     the write by one reactor tick, so the session holds a not-yet-resolved
+//     future in _output_pending.
+//  2. The handshake layer reports success and immediately invokes the registered
+//     DN verification callback, while _output_pending is still pending.
+//  3. The DN callback aborts the server's TX buffer and throws. The session
+//     treats this as a handshake failure and propagates the exception, skipping
+//     the step that would normally drain _output_pending after a successful
+//     handshake.
```
xemul: Which "step" (in the code) is that?

elcallio:
Would it not be a more correct fix to always do a wait_for_output on exception from verify? Or is this the verify called from verify_wrapper?

tests/unit/tls_test.cc
knowack1: I meant this step which drains _output_pending https://github.com/knowack1/seastar/blob/3cf054ef3ba145fb03302220814071e2c28122e4/src/net/tls_gnutls.cc#L652, But is not performed when verify() throws. I can improve description to be more specific.

knowack1:
> Would it not be a more correct fix to always do a wait_for_output on exception from verify? Or is this the verify called from verify_wrapper?

This is verify called from do_handshake_invoke. It could be done this way, as you suggested. However, I am not sure if this is the only pathway that can lead to an exceptional future in the destructor.

I also thought that it's not a bad idea to ignore a ready (including exceptional) future in the destructor, since the object is being destroyed and there is nothing more to be done with it.

However I can change the fix to call wait_for_output in case of exception in verify if you think its better to make this more targeted fix.

elcallio:
I think it is better to handle the error than the symptom, yes. If I understand the issue correctly, we should do a defensive wait_for_output  on all error paths of do_handshake_invoke, before actually throwing the exceptions?

elcallio (approved):

knowack1:
> I think it is better to handle the error than the symptom, yes. If I understand the issue correctly, we should do a defensive wait_for_output on all error paths of do_handshake_invoke, before actually throwing the exceptions?

Changed. Now, the failed future is handled in the `catch` block in `do_handshake_invoke`.

tests/unit/tls_test.cc
knowack1: Description updated to be more precise. Also, the implementation was updated according to @elcallio 's advice.

src/net/tls_gnutls.cc
```diff
@@ -650,7 +650,13 @@ class session : public enable_shared_from_this<session>, public tls::session_imp
             // make sure we reset output_pending
             return wait_for_output();
         } catch (...) {
-            return make_exception_future<>(std::current_exception());
+            auto ep = std::current_exception();
+            return wait_for_output().then_wrapped(
```
xemul, line 654: But vec_push() can (in theory) be called from any gnutls_... library function, thus potentially setting the _output_pending into exception. You spotted one place, but is there a guarantee that no other code needs similar draining?

elcallio: gnutls will afaik (from reading source) only call vec_push/pull on either write or handshake calls. We cover all other paths with wait_for_output (or via handle_output_error). The issue (as I understand it) is the invoked handshake call here leaving an exception in the output future, while also returning an error code that causes us to throw via verify().

knowack1: I'm not entirely sure, but this pathway is most likely what caused the issue in SCYLLADB-2906. I think we can either:
1. Drain here (assuming this pathway was the root cause of the original problem).
2. Ignore it in the destructor.
3. Do both, just in case.

I think we should stick to 1 or 2.
Option 3 doesn't make sense because draining in `do_handshake_invoke` doesn't do any meaningful work since the exception is ignored anyway. Therefore, ignoring it in the destructor is equivalent but has a broader scope.

I suggest starting with the more targeted fix first (1.), and we can revisit this if the problem reoccurs.

avikivity:
I tried the test with the change backed out. It succeeds (it does leak a failed future). It does fail with the test.py harness.

It's not really the fault of the patch, but please tighten the test by verifying that reactor::abandoned_failed_futures() did not increase during the test.

knowack1:
> I tried the test with the change backed out. It succeeds (it does leak a failed future). It does fail with the test.py harness.
> 
> It's not really the fault of the patch, but please tighten the test by verifying that reactor::abandoned_failed_futures() did not increase during the test.

Fixed. Now the reproduction test case calls:
```
BOOST_REQUIRE(!engine().abandoned_failed_futures());
```

tests/unit/tls_test.cc
```diff
@@ -2527,19 +2527,24 @@ SEASTAR_THREAD_TEST_CASE(test_output_pending_exception_on_destroy) {
...
                                .handle_exception([strms](std::exception_ptr) {});
 
-    // Reading on the server side drives the handshake, triggering the DN
-    // verification callback, which aborts the server's TX buffer and throws.
-    BOOST_REQUIRE_THROW(ss.input().read().get(), tls::verification_error);
+        BOOST_REQUIRE_THROW(ss.input().read().get(), tls::verification_error);
+    }
+
+    // Verify that no abandoned failed futures were created during the test
+    BOOST_REQUIRE(!engine().abandoned_failed_futures());
```
elcallio: As @avikivity pointed out, this should be a check against the value before the above test code. Since test cases run sequentially in the same engine(), a dropped future in a previous case could pollute this.

knowack1: I see. Fixed.

elcallio (approved):

avikivity:
> > I tried the test with the change backed out. It succeeds (it does leak a failed future). It does fail with the test.py harness.
> > It's not really the fault of the patch, but please tighten the test by verifying that reactor::abandoned_failed_futures() did not increase during the test.
> 
> Fixed. Now the reproduction test case calls:
> 
> ```
> BOOST_REQUIRE(!engine().abandoned_failed_futures());
> ```

This is too sensitive. Better to capture it before and after, and compare the difference.

avikivity:
> > > I tried the test with the change backed out. It succeeds (it does leak a failed future). It does fail with the test.py harness.
> > > It's not really the fault of the patch, but please tighten the test by verifying that reactor::abandoned_failed_futures() did not increase during the test.
> > 
> > 
> > Fixed. Now the reproduction test case calls:
> > ```
> > BOOST_REQUIRE(!engine().abandoned_failed_futures());
> > ```
> 
> This is too sensitive. Better to capture it before and after, and compare the difference.

Although, test.py has --abort-on-abanadoned-failed-future, so should not happen.

knowack1:
> > > > I tried the test with the change backed out. It succeeds (it does leak a failed future). It does fail with the test.py harness.
> > > > It's not really the fault of the patch, but please tighten the test by verifying that reactor::abandoned_failed_futures() did not increase during the test.
> > > 
> > > 
> > > Fixed. Now the reproduction test case calls:
> > > ```
> > > BOOST_REQUIRE(!engine().abandoned_failed_futures());
> > > ```
> > 
> > 
> > This is too sensitive. Better to capture it before and after, and compare the difference.
> 
> Although, test.py has --abort-on-abanadoned-failed-future, so should not happen.

Eventually I changed this to:
```
    uint64_t before = engine().abandoned_failed_futures();
    {
       // reproduction code here
    }
    uint64_t after = engine().abandoned_failed_futures();
    BOOST_REQUIRE_EQUAL(after, before);
 ```

PR #3500, opened 2026-06-30
net/tls: fix concurrent put() on the socket in the OpenSSL backend

dotnwat:
A data_sink permits only one put() in flight at a time (posix_data_sink_impl asserts !_p). The OpenSSL backend could violate this and crash:

    seastar::net::posix_data_sink_impl::put(): Assertion `!_p' failed.

The read and write paths use disjoint semaphores (_in_sem, _out_sem) and both can write to the underlying socket: a write encrypts application data, and a read emits a TLS key-update/renegotiation response from inside SSL_read_ex. When a read emitted output while a write's put() was in flight, a second, concurrent put() was issued on the same data_sink, tripping the assert.

bio_write_ex(), the only caller of _out.put(), already declines to issue a put() while one is in flight (BIO_set_retry_write), so OpenSSL retries after the in-flight put() drains. The guard was reading falsified state: wait_for_output() obtained an awaitable with

    std::exchange(_output_pending, make_ready_future())

which marked _output_pending available while the put() was still draining, so the guard saw an idle sink and issued the concurrent put(). This is not a missing lock; the reactor is single-threaded and the guard is already atomic.

Make _output_pending a shared_future so wait_for_output() can hand out an awaitable via get_future() without consuming it. available()/failed() then stay truthful while the put() drains, and both the read and write paths can await the same put().

Drop the SEASTAR_ASSERT(_output_pending.available()) preconditions in the two do_put() overloads and do_put_one(). They asserted that no put() is in flight when a write begins, which is false: _in_sem and _out_sem are disjoint, so a read-path put() can be in flight. The decline/retry path handles that correctly. The destructor's assert is kept; it still holds once close() has drained output.

Add test_concurrent_put_with_key_update, which installs an instrumented data_sink that flags overlapping put()s and drives two server key-updates to open the window deterministically. It is skipped for GnuTLS, which re-runs the handshake holding both semaphores and so never issues the concurrent put(). The full tls_test suite passes under both the OpenSSL and GnuTLS backends.

### DIFF
--- a/src/net/tls_openssl.cc
+++ b/src/net/tls_openssl.cc
@@ -52,6 +52,7 @@
 
 #include <seastar/core/gate.hh>
 #include <seastar/core/reactor.hh>
+#include <seastar/core/shared_future.hh>
 #include <seastar/core/sstring.hh>
 #include <seastar/core/with_timeout.hh>
 #include <seastar/net/stack.hh>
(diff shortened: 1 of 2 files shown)


src/net/tls_openssl.cc
```diff
@@ -2044,11 +2052,42 @@ class openssl_session : public enable_shared_from_this<openssl_session>, public
...
+    // a shared_future: get_future() hands out an awaitable without consuming
+    // _output_pending, so available()/failed() keep reflecting the in-flight put()
+    // and several fibers can await the same put(). A plain future would have to be
+    // moved out to be awaited, falsifying available() and letting the guard issue
+    // a second, concurrent put().
+    //
+    // On output failure _output_pending is left failed as a circuit breaker:
+    // bio_write_ex() issues no further puts and re-raises. The failure is terminal
+    // (the socket is gone), so _output_pending is never reset.
+    shared_future<> _output_pending;
```
dotnwat, line 2090: A semaphore/condvar is probably a more natural solution here, but threading it through abstractions was getting rather ugly. Perhaps the cleanest solution we could move to later would be to put a single queue in front of the output socket and a background fiber to drain that queue.

elcallio (approved):
This seems reasonable, though I am not really familiar enough with the OpenSSL back end to judge.

PR #3502, opened 2026-07-01
Trigger scylla next-gating on every seastar PR

yaronkaikov:
Today, integration problems between seastar and scylla only surface when the
seastar submodule is bumped in scylla - well after the seastar change has
merged. This workflow moves that feedback onto the seastar PR itself.

On every pull request it asks the scylla-master/seastar-ci Jenkins job to build
scylla master against this PR's merge ref and run the full next-gating suite
(unit tests and gating dtests). The job posts the result back as a commit status
on the PR, so the outcome is visible before merging. If the job cannot even be
triggered, the workflow pings #releng-team on Slack.

It runs as pull_request_target so that PRs from forks can reach the Jenkins and
Slack secrets; it only reads the PR number and checks out no PR code. It can also
be run manually via workflow_dispatch, passing a PR number, which is handy for
testing.

The Jenkins pipeline itself lives in scylla-pkg (scylladb/scylla-pkg#6405)

### DIFF
--- a/.github/workflows/trigger-scylla-gating.yaml
+++ b/.github/workflows/trigger-scylla-gating.yaml
@@ -1,0 +1,97 @@
+# Place this file in the scylladb/seastar repo at:
+#   .github/workflows/trigger-scylla-gating.yaml
+#
+# Triggers the scylla next-gating pipeline (RELENG-31) on a seastar PR, on demand,
+# so we learn about scylla/seastar integration breakage before the seastar change
+# merges. The Jenkins job (scylla-master/job/seastar-ci) checks out scylla master,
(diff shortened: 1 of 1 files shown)


yaronkaikov:
@scylladbbot trigger-ci

denesb:
@yaronkaikov do we have some validation on who can trigger CI? We need to make sure only people in the scylla project can actually start CI. Otherwise we are opening a backdoor here.

yaronkaikov:
> @yaronkaikov do we have some validation on who can trigger CI? We need to make sure only people in the scylla project can actually start CI. Otherwise we are opening a backdoor here.

```
# Only comments from users with write access or above (OWNER/MEMBER/COLLABORATOR)
# are honored.
```

PR #3508, opened 2026-07-06
lazy: move operator<< to the seastar namespace

avikivity:
Cleaner, and allows us to export it from the seastar namespace too.

The operator is found by argument-dependent lookup (ADL).

### DIFF
--- a/include/seastar/util/lazy.hh
+++ b/include/seastar/util/lazy.hh
@@ -109,9 +109,6 @@ lazy_deref(const T& p) {
     return lazy_deref_wrapper<T>(p);
 }
 
-}
-
-namespace std {
 /// Output operator for a seastar::lazy_eval<Func>
 /// This would allow printing a seastar::lazy_eval<Func> as if it's a regular
 /// value.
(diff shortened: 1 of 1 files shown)


include/seastar/util/lazy.hh
```diff
@@ -128,22 +125,22 @@ namespace std {
 ///
 /// \return os
 template <typename Func>
-ostream& operator<<(ostream& os, const seastar::lazy_eval<Func>& lf) {
+std::ostream& operator<<(std::ostream& os, const seastar::lazy_eval<Func>& lf) {
```
xemul: Since it now sits in ::seastar namespace, the namespace part of the lf argument type can be omitted? (same for other operator<<s)

avikivity:
v2: drop now-unneeded seastar:: namespace qualifiers

avikivity:

include/seastar/util/lazy.hh
```diff
@@ -128,22 +125,22 @@ namespace std {
 ///
 /// \return os
 template <typename Func>
-ostream& operator<<(ostream& os, const seastar::lazy_eval<Func>& lf) {
+std::ostream& operator<<(std::ostream& os, const lazy_eval<Func>& lf) {
```
travisdowns, line 128: This could be a compile-incompatible change in some exotic cases, but they are really pretty exotic, so I think we should just merge this and if you did one of those things you need to update your code.

travisdowns (approved):

PR #3510, opened 2026-07-06
perf: avoid `using namespace seastar` in header

avikivity:
This can bring seastar::file_handle name into the global namespace, where it clashes with the Linux file_handle symbol.


avikivity:
v2: distribute more `using namespace seastar` to .cc that need it

PR #3512, opened 2026-07-06
module: export more symbols

avikivity:
With these, a full build of ScyllaDB using modular Seastar passes.

To reduce work on grouping and ordering, which this write-only file doesn't really need, we first
sort all the symbols alphabetically.

### DIFF
--- a/src/seastar.cppm
+++ b/src/seastar.cppm
@@ -172,6 +172,7 @@ module;
 #include <seastar/util/tmp_file.hh>
 #include <seastar/util/memory-data-source.hh>
 #include <seastar/util/memory-data-sink.hh>
+#include <seastar/util/memory_diagnostics.hh>
 
 #include <seastar/net/arp.hh>
 #include <seastar/net/packet.hh>
@@ -198,6 +199,7 @@ module;
 #include <seastar/http/file_handler.hh>
 #include <seastar/http/function_handlers.hh>
 #include <seastar/http/httpd.hh>
+#include <seastar/http/api_docs.hh>
 #include <seastar/http/json_path.hh>
 #include <seastar/http/reply.hh>
 #include <seastar/http/response_parser.hh>
@@ -214,6 +216,7 @@ module;
 #include <seastar/rpc/rpc_types.hh>
 #include <seastar/rpc/lz4_compressor.hh>
 #include <seastar/rpc/lz4_fragmented_compressor.hh>
+#include <seastar/rpc/multi_algo_compressor_factory.hh>
 
 #include <seastar/coroutine/all.hh>
 #include <seastar/coroutine/as_future.hh>
@@ -222,6 +225,7 @@ module;
 #include <seastar/coroutine/maybe_yield.hh>
 #include <seastar/coroutine/parallel_for_each.hh>
 #include <seastar/coroutine/switch_to.hh>
+#include <seastar/coroutine/try_future.hh>
 
 export module seastar;
 
@@ -232,6 +236,7 @@ using seastar::abort_requested_exception;
 using seastar::abort_source;
 using seastar::accept_result;
 using seastar::access_flags;
+using seastar::adder;
 using seastar::align_down;
 using seastar::align_up;
 using seastar::allocate_aligned_buffer;
@@ -249,7 +254,9 @@ using seastar::broken_named_semaphore;
 using seastar::broken_pipe_exception;
 using seastar::broken_promise;
 using seastar::broken_semaphore;
+using seastar::cache_line_size;
 using seastar::cancelled_error;
+using seastar::CanInvoke;
 using seastar::check_direct_io_support;
 using seastar::checked_ptr;
 using seastar::checked_ptr_is_null_exception;
@@ -284,6 +291,7 @@ using seastar::data_source;
 using seastar::data_source_impl;
 using seastar::default_scheduling_group;
 using seastar::default_smp_service_group;
+using seastar::default_timeout_exception_factory;
 using seastar::defer;
 using seastar::deferred_action;
 using seastar::deferred_close;
@@ -301,9 +309,11 @@ using seastar::dynamic_pointer_cast;
 using seastar::enable_lw_shared_from_this;
 using seastar::enable_shared_from_this;
 using seastar::engine;
+using seastar::engine_is_ready;
 using seastar::enum_hash;
 using seastar::expiring_fifo;
 using seastar::file;
+using seastar::file_accessible;
 using seastar::file_desc;
 using seastar::file_exists;
 using seastar::file_handle;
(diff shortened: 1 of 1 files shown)


avikivity:
Depends on https://github.com/scylladb/seastar/pull/3511

avikivity:
v2: rebased past 5ed18dc922c829a1bfd7baa229212088e5ef48f6

avikivity:

travisdowns (approved):
LGTM but I know little about modules, I would leave it to others to merge this.

avikivity:
@scylladb/seastar-maint maybe someone else can review?

src/seastar.cppm
```diff
@@ -783,6 +902,20 @@ using seastar::rpc::streaming_domain_type;
...
+using seastar::rpc::multi_algo_compressor_factory;
+using seastar::rpc::client_info;
+using seastar::rpc::optional;
+using seastar::rpc::client_options;
+using seastar::rpc::server_options;
+using seastar::rpc::resource_limits;
+using seastar::rpc::canceled_error;
+using seastar::rpc::tuple;
+using seastar::rpc::no_wait_type;
+using seastar::rpc::no_wait;
```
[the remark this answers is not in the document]
avikivity: fixed

src/seastar.cppm
```diff
@@ -804,6 +941,11 @@ using seastar::util::basic_memory_data_sink;
 using seastar::util::read_entire_stream_contiguous;
 using seastar::util::as_input_stream;
 using seastar::util::read_entire_file_contiguous;
+using seastar::util::memory_data_sink;
+using seastar::util::memory_data_source;
+using seastar::util::skip_entire_stream;
+using seastar::util::read_entire_stream;
+using seastar::util::read_entire_stream_contiguous;
```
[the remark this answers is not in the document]
avikivity: fixed

avikivity:
v2:
 - don't introduce duplicate using-declarations

src/seastar.cppm
```diff
@@ -121,8 +122,6 @@ module;
 #include <seastar/core/shared_mutex.hh>
 #include <seastar/core/shared_ptr.hh>
 #include <seastar/core/shared_ptr_debug_helper.hh>
-// NOT: #include <seastar/core/shared_ptr_incomplete.hh>
-// shared_ptr_incomplete.hh provides a split-header pattern for lw_shared_ptr
```
travisdowns, line 125: Incorrectly deleted comment line? line 125 is a fragment now

travisdowns: So either delete the whole chunk or restore it the problem described here still exists.

src/seastar.cppm
```diff
@@ -388,10 +412,18 @@ using seastar::file_size;
...
 using seastar::make_file_output_stream;
+using seastar::make_file_data_sink;
+using seastar::make_file_data_source;
 using seastar::make_pipe_input_stream;
 using seastar::make_pipe_output_stream;
 using seastar::with_file;
 using seastar::with_file_close_on_failure;
+using seastar::link_file;
+using seastar::make_directory;
+using seastar::recursive_touch_directory;
```
travisdowns: duplicate see line 408

src/seastar.cppm
```diff
@@ -388,10 +412,18 @@ using seastar::file_size;
...
+using seastar::touch_directory;
```
travisdowns: duplicate

src/seastar.cppm
```diff
@@ -214,14 +215,15 @@ module;
 #include <seastar/rpc/rpc_types.hh>
 #include <seastar/rpc/lz4_compressor.hh>
 #include <seastar/rpc/lz4_fragmented_compressor.hh>
+#include <seastar/rpc/multi_algo_compressor_factory.hh>
 
 #include <seastar/coroutine/all.hh>
 #include <seastar/coroutine/as_future.hh>
-#include <seastar/coroutine/exception.hh>
```
travisdowns, line 220: Why drop this include? The purview still exports `coroutine::exception` / `coroutine::return_exception_ptr`, and this PR adds `coroutine::return_exception` and `coroutine::operator co_await` (whose only namespace-scope overload is declared in this header). It still compiles only because `core/coroutine.hh` includes it transitively, so the modules build is now one include-cleanup away from breaking.

If it needs to stay out of the GMF for some reason, a `// NOT:` comment like the `shared_ptr_incomplete.hh` one above would help; otherwise I would keep the direct include.

src/seastar.cppm
```diff
@@ -354,8 +374,12 @@ using seastar::open_flags;
 using seastar::access_flags;
 using seastar::file_permissions;
 using seastar::rename_flags;
+using seastar::file_accessible;
 using seastar::follow_symlink;
 using seastar::directory_entry_type;
+using seastar::file_type;
+using seastar::make_unbound_datagram_channel;
+using seastar::make_bound_datagram_channel;
```
travisdowns: These two are networking, but they sit between `directory_entry_type` and `directory_entry` in the file-I/O group — the sockets section below seems like a better home.

src/seastar.cppm
```diff
@@ -475,12 +508,20 @@ using seastar::simple_memory_output_stream;
...
 using seastar::on_internal_error;
 using seastar::on_internal_error_noexcept;
 using seastar::on_fatal_internal_error;
 using seastar::set_abort_on_internal_error;
+
+// Subprocess (also re-exported under seastar::experimental for legacy
+// code that used the deprecated namespace).
+using seastar::process;
+using seastar::spawn_parameters;
+using seastar::spawn_process;
```
travisdowns: This block landed in the middle of the error-handling group, splitting `set_abort_on_internal_error` from `set_abort_on_ebadf` / `current_backtrace`. Since the section groupings are the only map this file has, could you move it after that group?

src/seastar.cppm
```diff
@@ -658,7 +735,26 @@ using seastar::memory::allocation_site;
...
+using seastar::memory::min_free_memory;
+using seastar::memory::memory_layout;
+using seastar::memory::get_memory_layout;
+using seastar::memory::memory_diagnostics_writer;
+using seastar::memory::set_additional_diagnostics_producer;
+using seastar::memory::generate_memory_diagnostics_report;
+
+}
 
+export namespace seastar::scollectd {
```
travisdowns: Nit: the other namespace blocks in this file have a blank line after the opening brace. (Likewise, the new operators block at the bottom could use a `// Operators` section comment to match the rest of the file.)

src/seastar.cppm
avikivity: I will change the file to be sorted. Any grouping we choose will inevitably fail as people will mis-categorize or just append. At least sorting is objective.

src/seastar.cppm
avikivity: I think it should be kept.

src/seastar.cppm
avikivity: Yes, should have been kept.

avikivity:
v2:
 - new first commit to sort the file by namespace and exported symbol name
 - undo some gratuitous removal of comments and includes

PR #3513, opened 2026-07-07
reactor: Replace indirect_compare struct with a lambda

xemul:
The single-use struct only obfuscates a simple vruntime comparison. Replace it with an inline lambda at the point of use.


bhalevy:
What's the motivation?

xemul:
Mainly that the struct with an operator is confusing -- we can go with much shorter and cleaner code for simple vruntime comparisons

bhalevy (approved):

PR #3519, opened 2026-07-08
collectd: avoid static template functions in headers

avikivity:
A static function tells the compiler the function will not be used outside this translation unit. If a static function is in a header, the compiler will complain if a random translation unit uses the header without using the function (which is likely).

Clang 23 also complains about static function templates. Our -Werror promotes this to an error.

Fix by removing `static`. This is a good thing anyway, as it creates multiple copies of the function (if it happens to indeed be used in multiple translation units).

### DIFF
--- a/include/seastar/core/scollectd.hh
+++ b/include/seastar/core/scollectd.hh
@@ -727,7 +727,7 @@ void add_polled(const type_instance_id &, const shared_ptr<value_list> &, bool e
 
 typedef std::function<void()> notify_function;
 template<typename... _Args>
-static auto make_type_instance(description d, _Args && ... args) -> values_impl < decltype(value<_Args>(std::forward<_Args>(args)))... >
+auto make_type_instance(description d, _Args && ... args) -> values_impl < decltype(value<_Args>(std::forward<_Args>(args)))... >
 {
     return values_impl<decltype(value<_Args>(std::forward<_Args>(args)))...>(
                     std::move(d), value<_Args>(std::forward<_Args>(args))...);
@@ -737,7 +737,7 @@ static auto make_type_instance(description d, _Args && ... args) -> values_impl 
  *
  */
 template<typename ... _Args>
-[[deprecated("Use the metrics layer")]] static type_instance_id add_polled_metric(const plugin_id & plugin,
+[[deprecated("Use the metrics layer")]] type_instance_id add_polled_metric(const plugin_id & plugin,
         const plugin_instance_id & plugin_instance, const type_id & type,
         const scollectd::type_instance & type_instance, _Args&& ... args) {
     return add_polled_metric(plugin, plugin_instance, type, type_instance, description(),
@@ -748,7 +748,7 @@ template<typename ... _Args>
  *
  */
 template<typename ... _Args>
-[[deprecated("Use the metrics layer")]] static type_instance_id add_polled_metric(const plugin_id & plugin,
+[[deprecated("Use the metrics layer")]] type_instance_id add_polled_metric(const plugin_id & plugin,
         const plugin_instance_id & plugin_instance, const type_id & type,
         const scollectd::type_instance & type_instance, description d, _Args&& ... args) {
     return add_polled_metric(
@@ -756,7 +756,7 @@ template<typename ... _Args>
             std::forward<_Args>(args)...);
 }
 template<typename ... _Args>
-static future<> send_explicit_metric(const plugin_id & plugin,
+future<> send_explicit_metric(const plugin_id & plugin,
         const plugin_instance_id & plugin_instance, const type_id & type,
         const scollectd::type_instance & type_instance, _Args&& ... args) {
     return send_explicit_metric(
@@ -764,7 +764,7 @@ static future<> send_explicit_metric(const plugin_id & plugin,
             std::forward<_Args>(args)...);
 }
 template<typename ... _Args>
-static notify_function create_explicit_metric(const plugin_id & plugin,
+notify_function create_explicit_metric(const plugin_id & plugin,
         const plugin_instance_id & plugin_instance, const type_id & type,
         const scollectd::type_instance & type_instance, _Args&& ... args) {
     return create_explicit_metric(
@@ -778,7 +778,7 @@ seastar::metrics::impl::metric_id to_metrics_id(const type_instance_id & id);
  *
  */
 template<typename Arg>
-[[deprecated("Use the metrics layer")]] static type_instance_id add_polled_metric(const type_instance_id & id, description d,
+[[deprecated("Use the metrics layer")]] type_instance_id add_polled_metric(const type_instance_id & id, description d,
         Arg&& arg, bool enabled = true) {
     seastar::metrics::impl::get_local_impl()->add_registration(to_metrics_id(id), arg.type, seastar::metrics::impl::make_function(arg.value, arg.type), d, enabled);
     return id;
@@ -788,7 +788,7 @@ template<typename Arg>
  *
  */
 template<typename Arg>
-[[deprecated("Use the metrics layer")]] static type_instance_id add_polled_metric(const type_instance_id & id,
+[[deprecated("Use the metrics layer")]] type_instance_id add_polled_metric(const type_instance_id & id,
         Arg&& arg) {
     return std::move(add_polled_metric(id, description(), std::forward<Arg>(arg)));
 }
@@ -798,19 +798,19 @@ template<typename Arg>
  *
  */
 template<typename Args>
-[[deprecated("Use the metrics layer")]] static type_instance_id add_disabled_polled_metric(const type_instance_id & id, description d,
+[[deprecated("Use the metrics layer")]] type_instance_id add_disabled_polled_metric(const type_instance_id & id, description d,
         Args&& arg) {
     return add_polled_metric(id, d, std::forward<Args>(arg), false);
 }
 
 template<typename Args>
-static type_instance_id add_disabled_polled_metric(const type_instance_id & id,
+type_instance_id add_disabled_polled_metric(const type_instance_id & id,
         Args&& args) {
     return add_disabled_polled_metric(id, description(), std::forward<Args>(args));
 }
 
 template<typename ... Args>
-static type_instance_id add_disabled_polled_metric(const type_instance_id & id,
+type_instance_id add_disabled_polled_metric(const type_instance_id & id,
         Args&& ... args) {
     return add_disabled_polled_metric(id, description(), std::forward<Args>(args)...);
 }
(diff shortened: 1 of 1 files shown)


avikivity:

travisdowns:
> A static function tells the compiler the function will not be used outside this translation unit. If a static function is in a header, the compiler will complain if a random translation unit uses the header without using the function (which is likely).

Yes, but IMO this warning behavior is buggy/undesirable.

The compiler does not complain if inline is added. It is totally valid to want to have only-static (not inline) functions in a header, such as laziness or a project distributed header only.

avikivity:
> > A static function tells the compiler the function will not be used outside this translation unit. If a static function is in a header, the compiler will complain if a random translation unit uses the header without using the function (which is likely).
> 
> Yes, but IMO this warning behavior is buggy/undesirable.
> 
> The compiler does not complain if inline is added. It is totally valid to want to have only-static (not inline) functions in a header, such as laziness or a project distributed header only.

`inline` today means "don't complain if the function is defined but not used". Just like an inline variable means "don't complain if the variable is defined but not used".

The complaints about static are generated (for ages) for non-template non-inline functions, the compiler is only becoming more consistent here (more consistent in its inconsistency with the English language, but still).

travisdowns:
> `inline` today means "don't complain if the function is defined but not used". Just like an inline variable means "don't complain if the variable is defined but not used".

No, that's a common myth: it is _also_ still an optimization hint. `inline static` is a stronger hint to optimize than `static`. (I mean in actual practice in modern gcc and clang, not "in theory").

travisdowns (approved):

avikivity:
Do you have an example?

travisdowns:
> Do you have an example?

Yes — and the mechanism is explicit in the compiler sources.

*(Disclosure: this comment was researched and written by an AI assistant (Claude), at my direction, from the actual clang/LLVM sources; I've reviewed it.)*

### The mechanism in clang/LLVM

1. **Clang turns the keyword into an IR attribute.** In `SetLLVMFunctionAttributesForDefinition`, clang checks whether any redeclaration has `isInlineSpecified()` — i.e. the `inline` keyword was literally written — and if so adds the `inlinehint` attribute to the IR function:

https://github.com/llvm/llvm-project/blob/2b20cf7291c1/clang/lib/CodeGen/CodeGenModule.cpp#L2716-L2733

2. **The LLVM inliner pays a premium for it.** The default cost threshold is 225 at -O2 (250 at -O3, 50 at -Os), but `inlinehint` callees get 325:

https://github.com/llvm/llvm-project/blob/2b20cf7291c1/llvm/lib/Analysis/InlineCost.cpp#L77-L83

https://github.com/llvm/llvm-project/blob/2b20cf7291c1/llvm/lib/Analysis/InlineCost.cpp#L2128-L2132

So at -O2 writing `inline` buys ~44% more inlining budget. At -Os it's dramatic — the hint check runs *after* the optsize clamp, so it's 325 vs 50, a 6.5× larger budget.

Note also that clang checks `isInlineSpecified()` (keyword literally written), not `isInlined()` (has inline semantics). Functions that are *semantically* inline without the keyword — in-class member definitions, `constexpr` — do **not** get the hint. If `inline` were purely "don't complain about unused / ODR permission", that distinction wouldn't exist; clang deliberately treats the written keyword as an optimization request.

### Example

A function whose inline cost (295) lands between the two thresholds, address-taken to defeat the last-call-to-static bonus, called from two callers. The only difference between the two compiles is the `inline` keyword:

```
$ clang-20 -O2 -c inline_demo.c -Rpass=inline -Rpass-missed=inline
remark: 'process' not inlined into 'caller1' because too costly to inline
        (cost=295, threshold=225) [-Rpass-missed=inline]

$ clang-20 -O2 -DUSE_INLINE -c inline_demo.c -Rpass=inline -Rpass-missed=inline
remark: 'process' inlined into 'caller1' with (cost=295, threshold=325) [-Rpass=inline]
```

`objdump` confirms: 2 residual calls to `process` in the plain-`static` object, 0 in the `static inline` one.

**gcc 13.3 flips on the exact same file** at -O2 (2 residual calls without the keyword, 0 with it). gcc's mechanism is analogous: `DECL_DECLARED_INLINE_P` selects the much larger `max-inline-insns-single` budget instead of `max-inline-insns-auto` in `ipa-inline.cc`.

<details>
<summary>inline_demo.c</summary>

```c
// The callee's inline cost must land between the default threshold (225)
// and the inlinehint threshold (325) at -O2.

#ifdef USE_INLINE
#define MAYBE_INLINE inline
#else
#define MAYBE_INLINE
#endif

static MAYBE_INLINE int process(const int *data, int n) {
    int a = 0, b = 1, c = 2, d = 3;
    for (int i = 0; i < n; i++) {
        int v = data[i];
        if (v & 1) a += v * 3 + 1;
        else       b ^= v + a;
        if (v & 2) c += b - (v >> 2);
        else       d -= c ^ (v << 1);
        if (v & 4) a ^= d + b;
        else       b += a - c;
        if (v > 100) { a += 7; c ^= a; }
        if (v < -100) { b -= 9; d += b; }
        if (v == 7) { a = b * c; d ^= a; }
        if (v == 42) { b = c - d; a += b; }
        a = a * 31 + c;
        b = b * 17 - d;
        c = c ^ (a >> 3);
        d = d + (b << 2);
        a ^= (v * 13) >> 1;
        d -= (v * 7) << 1;
    }
    return a + b + c + d;
}

// Keep an address-taken use alive so the "last call to static function"
// bonus (which force-inlines the sole remaining call) can't kick in.
int (*volatile keep)(const int *, int) = process;

int caller1(const int *data, int n) { return process(data, n) + 1; }
int caller2(const int *data, int n) { return process(data, n) + 2; }
```

</details>

To be fair: as a statement about the *standard* ("the intent-to-inline meaning carries no guarantee"), your framing is right. But as a statement about what modern compilers actually do, both gcc and clang still honor the keyword as a costed optimization hint.

travisdowns:
> Do you have an example?

Sure. Summary: on both current clang and gcc there is a wide band of function sizes where the *same* function body is inlined when declared `static inline` but stays an out-of-line call as plain `static`. Quantitatively, the keyword buys a **~1.4× larger inlining budget on clang and ~2.4× on gcc**: a function declared `inline` can be that much larger and still get inlined.

Reproducers (standalone example + size-sweep script): https://gist.github.com/travisdowns/8f448811c49f7388e64ca5525443525b

The test function is N statements of straight-line integer arithmetic with two external call sites:

```c++
QUAL int compute(int x) {   // QUAL = static | static inline
    x += (x ^ 0) * 3;
    x += (x ^ 31153) * 5;
    x += (x ^ 62306) * 7;
    // ... N such statements ...
    return x;
}

int caller1(int a) { return compute(a) + 1; }
int caller2(int a) { return compute(a * 2 + 1) + 2; }
```

Sweeping N and checking the asm for surviving calls:

| compiler | `static` inlined up to | `static inline` inlined up to | ratio |
|---|---|---|---|
| clang 20.1.8 `-O2` | N=25 | N=35 | ~1.4× |
| gcc 13 `-O2` | N=10 | N=24 | ~2.4× |

(Below the window both inline, above it neither does; in between only the `inline`-declared variant does.)

The mechanism isn't incidental — both compilers explicitly key their cost model on the keyword:

* **clang**: `inline` sets the `inlinehint` IR attribute, which selects a higher inline-cost threshold: 325 (`-inlinehint-threshold`) vs 225 (`-inline-threshold`) at `-O2`. You can watch it decide, at N=30 (cost 405) the identical callee is rejected or accepted purely on the threshold (the 1.5× vs the defaults is a single-basic-block bonus applied to both):

  ```
  $ clang++-20 -O2 -c example.cpp -Rpass-missed=inline
  remark: '_ZL7computei' not inlined into '_Z7caller1i' because too costly to inline (cost=405, threshold=337)
  $ clang++-20 -O2 -c example.cpp -DUSE_INLINE -Rpass=inline
  remark: '_ZL7computei' inlined into '_Z7caller1i' with (cost=405, threshold=487)
  ```

* **gcc**: functions declared `inline` are budgeted against `max-inline-insns-single` (default 70 at `-O2`) while non-`inline` functions get `max-inline-insns-auto` (default 15).

Two details for anyone reproducing this: you need **at least two call sites** — a `static` with a single call site gets LLVM's last-call-to-static bonus (inlining lets the function be deleted, so it inlines at almost any size) — and the callers must do something after the call, otherwise it's a tail call and appears as `jmp` rather than `call` in the asm.

So "`inline` is just ODR/don't-warn relief" doesn't match practice: it directly selects a materially larger inlining budget in both compilers.

*(Disclosure: this analysis and the reproducer scripts were produced with the help of Claude, an AI assistant.)*

avikivity:
Thanks. Learned something!

Do you know if this hold for implicit inlines? Is there a difference between struct x { void f(); }; and struct x { inline void f(); };?

Note, this is tangential to the patch, as those functions weren't inline, before or after.

It may make some sense to annotate small helpers this way, or perhaps -O3 and LTO is enough to overcome the heuristics.

travisdowns:
> -O3 and LTO is enough to overcome the heuristics.

-O3 no, I don't think LTO has an effect. Maybe you meant PGO: PGO for sure changes all of this, decisions are based on hotness rather than size or a combination of such. I'm not 100% sure if it means `inline` has zero effect or not though.

Note that the effect isn't _that_ big, about 2x in terms of the threshold, and functions vary in size over many orders of magnitude, so this only affects decisions a bit, but it's not nothing.

travisdowns:
> Do you know if this hold for implicit inlines? Is there a difference between struct x { void f(); }; and struct x { inline void f(); };?

I'll save you the slop, but based on my testing, it depends. clang uses the NOT inline threshold for implicitly inlined memory functions, while gcc uses the inline one. Your example appears slightly different: neither is implictly inline (not delcared in the struct body), both clang and gcc treat the former as NOT inline for optimization, and the latter as inline.

Here's a table:

  | case | clang-20: `inlinehint` attr | clang-20: threshold | clang-20: inlined? | gcc-13: inlined? |
  |---|---|---|---|---|
  | defined out-of-line, no `inline` (`struct x { void f(); };`) | no | 337 | **no** | **no** |
  | `inline` on declaration, defined out-of-line (`struct x { inline void f(); };`) | yes | 487 | yes | yes |
  | `inline` only on the out-of-line definition | yes | 487 | yes | yes |
  | defined in the class body (implicitly inline) | **no** | **337** | **no** | **yes** |
  | defined in class body + redundant explicit `inline` | yes | 487 | yes | yes |
  | `constexpr` (implicitly inline) | no | 337 | **no** | yes |

travisdowns:
(so on clang, adding a "redundant" `inline` specifier on member functions defined in the class body is _not_ strictly redundant...)

avikivity:
Wow. Well we have -mllvm -inline-threshold=2500, so we're a little insulated from this mess.

travisdowns:
> Wow. Well we have -mllvm -inline-threshold=2500, so we're a little insulated from this mess.

Yes. We finally deleted that one on redpanda side :) and saved many hours of compile time.

mykaul:
Won't it be better to remove collectd support from seastar?

avikivity:
Fixing the compile problem is easy, while knowing if there are any collectd users is hard.

PR #3520, opened 2026-07-08
http: refuse requests on connections with an expired client certificate

nyh:
http: refuse requests on connections with an expired client certificate
    
    TLS certificate verification at handshake time already rejects expired
    certificates, but this only protects against expired certs at connection
    establishment. On long-lived connections a client certificate can expire
    while the connection is still open; subsequent requests on that connection
    were previously accepted without any expiry check.
    
    To close this gap, add tls::get_certificate_expiry(), which extracts the
    notAfter field from the peer's leaf certificate after the TLS handshake.
    Implemented for both GnuTLS (gnutls_x509_crt_get_expiration_time) and
    OpenSSL (X509_get0_notAfter + ASN1_TIME_to_tm).
    
    The HTTP server fetches the expiry in connection::prepare() alongside the
    DN, stores it on the connection, and checks it in read_one() before
    dispatching each request. If the current time is past the expiry, the
    request is refused with 401 Unauthorized and the connection is closed.

### DIFF
--- a/include/seastar/http/request.hh
+++ b/include/seastar/http/request.hh
@@ -30,14 +30,15 @@
 //
 #pragma once
 
-#include <seastar/core/iostream.hh>
 #include <seastar/core/sstring.hh>
 #include <string_view>
+#include <optional>
 #include <strings.h>
 #include <seastar/http/common.hh>
 #include <seastar/http/mime_types.hh>
 #include <seastar/http/types.hh>
 #include <seastar/net/socket_defs.hh>
+#include <seastar/net/api.hh>
 #include <seastar/core/iostream.hh>
 #include <seastar/util/string_utils.hh>
 #include <seastar/util/iostream.hh>
@@ -80,6 +81,8 @@ struct request {
     std::unordered_map<sstring, sstring> trailing_headers;
     std::unordered_map<sstring, sstring> chunk_extensions;
     sstring protocol_name = "http";
+    /// Distinguished name from the client's TLS certificate, if mTLS was used.
+    std::optional<session_dn> _tls_dn;
     http::body_writer_type body_writer; // for client
 
     using query_parameters_type = std::unordered_map<sstring, std::vector<sstring>, seastar::internal::string_view_hash, std::equal_to<>>;
--- a/src/net/tls_gnutls.cc
+++ b/src/net/tls_gnutls.cc
@@ -1248,6 +1248,20 @@ public:
             return extract_dn_information();
         });
     }
+    future<std::optional<std::chrono::system_clock::time_point>> get_certificate_expiry() override {
+        return state_checked_access([this] {
+            auto peer = get_peer_certificate();
+            if (!peer) {
+                return std::optional<std::chrono::system_clock::time_point>{};
+            }
+            time_t t = gnutls_x509_crt_get_expiration_time(peer.get());
+            if (t == (time_t)-1) {
+                return std::optional<std::chrono::system_clock::time_point>{};
+            }
+            return std::optional<std::chrono::system_clock::time_point>{
+                std::chrono::system_clock::from_time_t(t)};
+        });
+    }
     future<std::vector<subject_alt_name>> get_alt_name_information(std::unordered_set<subject_alt_name_type> types) override {
         return state_checked_access([this](std::unordered_set<subject_alt_name_type> types) {
             std::vector<subject_alt_name> res;
--- a/src/net/tls_gnutls.cc
+++ b/src/net/tls_gnutls.cc
@@ -575,6 +575,31 @@ public:
         return s;
     }
 
+    // Calls verify() and, if verification fails, sends the appropriate TLS alert
+    // to the peer before propagating the exception. Returns nullopt when
+    // verification succeeded (caller should continue normally), or a future
+    // that sends the alert and resolves to the verification error.
+    std::optional<future<>> verify_and_send_alert() {
+        gnutls_alert_description_t cert_alert = GNUTLS_A_BAD_CERTIFICATE;
+        std::exception_ptr vex;
+        try {
+            verify(&cert_alert);
+        } catch (...) {
+            vex = std::current_exception();
+        }
+        if (!vex) {
+            return std::nullopt;
+        }
+        _error = vex;
+        return wait_for_output().then_wrapped([this, cert_alert, vex = std::move(vex)](future<> f) mutable {
+            f.ignore_ready_future();
+            return send_alert(GNUTLS_AL_FATAL, cert_alert).then_wrapped([vex = std::move(vex)](future<> f) mutable {
+                f.ignore_ready_future();
+                return make_exception_future<>(std::move(vex));
+            });
+        });
+    }
+
     future<> send_alert(gnutls_alert_level_t level, gnutls_alert_description_t desc) {
         return repeat([this, level, desc]() {
             auto res = gnutls_alert_send(*this, level, desc);
@@ -624,7 +649,13 @@ public:
                     return make_exception_future<>(verification_error("No certificate was found"));
 #if GNUTLS_VERSION_NUMBER >= 0x030406
                 case GNUTLS_E_CERTIFICATE_ERROR:
-                    verify(); // should throw. otherwise, fallthrough
+                    // Before propagating the error, send the appropriate TLS
+                    // alert to the peer so it gets a meaningful error rather
+                    // than an abrupt connection close.
+                    if (auto vf = verify_and_send_alert()) {
+                        return std::move(*vf);
+                    }
+                    // verify() didn't throw (cert verification disabled?), fall through
                     [[fallthrough]];
 #endif
                 default:
@@ -644,7 +675,12 @@ public:
                 }
             }
             if (_type == type::CLIENT || _creds->get_client_auth() != client_auth::NONE) {
-                verify();
+                // Certificate verification failed after successful TLS handshake.
+                // Send the appropriate alert to the peer so it gets a meaningful
+                // error rather than an abrupt connection close.
+                if (auto vf = verify_and_send_alert()) {
+                    return std::move(*vf);
+                }
             }
             _connected = true;
             // make sure we reset output_pending
@@ -742,7 +778,22 @@ public:
         return from_transport_ptr(ptr)->pull(dst, len);
     }
 
-    void verify() {
+    // Returns the TLS alert description appropriate for a certificate
+    // that failed verification with the given gnutls_certificate_verify_peers3 status.
+    static gnutls_alert_description_t cert_status_to_alert(unsigned int status) noexcept {
+        if (status & (GNUTLS_CERT_SIGNER_NOT_FOUND | GNUTLS_CERT_SIGNER_NOT_CA)) {
+            return GNUTLS_A_UNKNOWN_CA;
+        } else if (status & GNUTLS_CERT_REVOKED) {
+            return GNUTLS_A_CERTIFICATE_REVOKED;
+        } else if (status & GNUTLS_CERT_EXPIRED) {
+            return GNUTLS_A_CERTIFICATE_EXPIRED;
+        }
+        return GNUTLS_A_BAD_CERTIFICATE;
+    }
+
+    // out_alert, if non-null, is set to the appropriate TLS alert before
+    // verification_error is thrown, so the caller can send the alert to the peer.
+    void verify(gnutls_alert_description_t* out_alert = nullptr) {
         if (!_creds->_enable_certificate_verification) {
             return;
         }
@@ -769,6 +820,9 @@ public:
                 }
                 ss << "(Issuer=[" << dn->issuer << "], Subject=[" << dn->subject << "])";
                 stat_str = ss.str();
+            }
+            if (out_alert) {
+                *out_alert = cert_status_to_alert(status);
             }
             throw verification_error(stat_str);
         }
--- a/src/net/tls_gnutls.cc
+++ b/src/net/tls_gnutls.cc
@@ -1202,6 +1202,20 @@ public:
             return extract_dn_information();
         });
     }
+    future<std::optional<std::chrono::system_clock::time_point>> get_certificate_expiry() override {
+        return state_checked_access([this] {
+            auto peer = get_peer_certificate();
+            if (!peer) {
+                return std::optional<std::chrono::system_clock::time_point>{};
+            }
+            time_t t = gnutls_x509_crt_get_expiration_time(peer.get());
+            if (t == (time_t)-1) {
+                return std::optional<std::chrono::system_clock::time_point>{};
+            }
+            return std::optional<std::chrono::system_clock::time_point>{
+                std::chrono::system_clock::from_time_t(t)};
+        });
+    }
     future<std::vector<subject_alt_name>> get_alt_name_information(std::unordered_set<subject_alt_name_type> types) override {
         return state_checked_access([this](std::unordered_set<subject_alt_name_type> types) {
             std::vector<subject_alt_name> res;
--- a/src/net/tls_openssl.cc
+++ b/src/net/tls_openssl.cc
@@ -40,6 +40,7 @@
 
 #include <cassert>
 #include <chrono>
+#include <ctime>
 #include <functional>
 #include <span>
 #include <system_error>
@@ -1521,6 +1522,39 @@ public:
         }
         result_t dn = extract_dn_information(is_verification_error::no);
         return make_ready_future<result_t>(std::move(dn));
+    }
+
+    future<std::optional<std::chrono::system_clock::time_point>> get_certificate_expiry() override {
+        using result_t = std::optional<std::chrono::system_clock::time_point>;
+        if (_error) {
+            return make_exception_future<result_t>(_error);
+        }
+        if (_shutdown) {
+            return make_exception_future<result_t>(
+              std::system_error(ENOTCONN, std::system_category()));
+        }
+        if (!connected()) {
+            return handshake().then(
+              [this]() mutable { return get_certificate_expiry(); });
+        }
+        const auto peer_cert = get_peer_certificate();
+        if (!peer_cert) {
+            return make_ready_future<result_t>(std::nullopt);
+        }
+        const ASN1_TIME* not_after = X509_get0_notAfter(peer_cert.get());
+        if (!not_after) {
+            return make_ready_future<result_t>(std::nullopt);
+        }
+        struct tm tm = {};
+        if (ASN1_TIME_to_tm(not_after, &tm) != 1) {
+            return make_ready_future<result_t>(std::nullopt);
+        }
+        time_t t = timegm(&tm);
+        if (t == (time_t)-1) {
+            return make_ready_future<result_t>(std::nullopt);
+        }
+        return make_ready_future<result_t>(
+            std::chrono::system_clock::from_time_t(t));
     }
 
     future<std::vector<subject_alt_name>> get_alt_name_information(
--- a/src/net/tls_openssl.cc
+++ b/src/net/tls_openssl.cc
@@ -40,6 +40,7 @@
 
 #include <cassert>
 #include <chrono>
+#include <ctime>
 #include <functional>
 #include <ranges>
 #include <span>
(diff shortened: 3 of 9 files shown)


nyh:
Please ignore the first two patches (they are just a copy of https://github.com/scylladb/seastar/pull/3517). The only relevant patch is the last one.

elcallio:
I think this should be optional, since there are cases (mainly tests) where client code might want to ignore this.

tests/unit/httpd_test.cc
```diff
@@ -2736,3 +2736,127 @@ SEASTAR_TEST_CASE(test_request_scheduling_group) {
...
+// can perform certificate-based authentication.
+SEASTAR_TEST_CASE(test_mtls_dn_propagation) {
+    return seastar::async([] {
+        // Self-signed CA (CN=server.com) that signs both the server and client certificates below.
+        static const char mtls_ca_cert[] =
+            "-----BEGIN CERTIFICATE-----\n"
+            "MIIBkTCCATagAwIBAgIUHAGWAd2vYMIUpcWjOX8J2d4DCX4wCgYIKoZIzj0EAwIw\n"
+            "FTETMBEGA1UEAwwKc2VydmVyLmNvbTAeFw0yNjA3MDgxOTE4MjZaFw0zNjA3MDUx\n"
+            "OTE4MjZaMBUxEzARBgNVBAMMCnNlcnZlci5jb20wWTATBgcqhkjOPQIBBggqhkjO\n"
+            "PQMBBwNCAASbsF3KPrYN03EHUBgmaTx8F3g2WHySxf1WAJMNA9KwPlSTpDCISXFQ\n"
```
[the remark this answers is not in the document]
nyh: Interesting. I don't know why my AI added this notBefore date.
But since this date is of *yesterday*, and nobody will ever run this test with a clock skew of a whole day, I think we're fine.

nyh:
> I think this should be optional, since there are cases (mainly tests) where client code might want to ignore this.

I guess I can make it optional, but can you give me an example why users will ever want to disable it? What sort of tests will want to disable it?

That being said, I think I might have overplayed the security implications of this fix a bit. The main point of having temporary certificates is so that if the certificate is stolen, it will not be usable later. Not allowing new connections with expired certificates already prevents most of that. The only remaining risk is that a hacker, immediately after stealing the certificate (before it's expired) connects to the server, and then maintains this connection open for a very long time (e.g., by periodically sending a silly request), until he wants to run the attack. This is a somewhat uncommon mode of attack, but it is possible.

Another option we could use, and will avoid these specific hacks to the HTTP server, is just to limit the time duration of any connection - close any connection after (say) one hour, no matter of how many requests were sent over it. The performance implications of forcing reconnection after an hour is negligible, but it will ensure that no request can be accepted more than an hour after the certificate's expiration.

elcallio:
> I guess I can make it optional, but can you give me an example why users will ever want to disable it? What sort of tests will want to disable it?

Not unit tests, I mean while testing stuff you are building. The snakeoil scenario. It is not significant obviously, But I still feel things like this should be guarded by a default-on flag in the tls options.

ScyllaPiotr:
> The performance implications of forcing reconnection after an hour is negligible

This is not true. We can have **many** open connections. A 1-hour long connection TTL was recently significantly prolonged by @dkropachev

src/net/tls_openssl.cc
```diff
@@ -1523,6 +1524,39 @@ class openssl_session : public enable_shared_from_this<openssl_session>, public
         return make_ready_future<result_t>(std::move(dn));
     }
 
+    future<std::optional<std::chrono::system_clock::time_point>> get_certificate_expiry() override {
```
ScyllaPiotr, line 1584: Could and adversary use one of the several possibilities for this function to return `std::nullopt` to create a connection which would never expire? After all, we are expiring connections only if this function returns non-null.

src/net/tls_gnutls.cc
```diff
@@ -1248,6 +1248,20 @@ class session : public enable_shared_from_this<session>, public tls::session_imp
             return extract_dn_information();
         });
     }
+    future<std::optional<std::chrono::system_clock::time_point>> get_certificate_expiry() override {
+        return state_checked_access([this] {
+            auto peer = get_peer_certificate();
+            if (!peer) {
+                return std::optional<std::chrono::system_clock::time_point>{};
```
ScyllaPiotr, line 1209: With proper declaration of the return type for the lambda:
` -> std::optional<std::chrono::system_clock::time_point>`
these elaborate returns can be substituted with `return std::nullopt;`.

Also, see my other comment about the risk with how `nullopt` is interpreted.

dkropachev:
> > The performance implications of forcing reconnection after an hour is negligible
> 
> This is not true. We can have **many** open connections. A 1-hour long connection TTL was recently significantly prolonged by @dkropachev

It could be the case that user issues certificates in bulk and all the http connections will outdate at the same time, which will create problems.
Having that said probably it would be a good thing to consider that and outdate not more than `X` (say 1000) number of connections a second per shard.

include/seastar/http/request.hh
```diff
@@ -80,6 +81,8 @@ struct request {
     std::unordered_map<sstring, sstring> trailing_headers;
     std::unordered_map<sstring, sstring> chunk_extensions;
     sstring protocol_name = "http";
+    /// Distinguished name from the client's TLS certificate, if mTLS was used.
+    std::optional<session_dn> _tls_dn;
```
dkropachev: Could we also export `session_dn` from the module facade? `seastar::http::request` is exported from `src/seastar.cppm`, and this branch adds a public `std::optional<session_dn> _tls_dn` member to it. Module consumers that `import seastar` can name `http::request`, but the member type does not appear to be re-exported there.

It may also be worth exporting the related TLS accessors (`tls::get_dn_information` / `tls::get_certificate_expiry`) while touching this list, but the `session_dn` export is the part needed for the new request member.

avikivity:
> On long-lived connections a client certificate can expire
while the connection is still open; subsequent requests on that connection
were previously accepted without any expiry check.

What do other HTTP/TLS servers do?

Seems like an unimportant edge case. Certificate expiration dates are measured in years, no?

ScyllaPiotr:
> Certificate expiration dates are measured in years, no?

We want to enable enable short-lived certificates for https://docs.google.com/document/d/1gF1D8FgSBOec-yhMkhuVXUGVkT_VoWMdrIq1e7zyzi4/

avikivity:
> > Certificate expiration dates are measured in years, no?
> 
> We want to enable enable short-lived certificates for https://docs.google.com/document/d/1gF1D8FgSBOec-yhMkhuVXUGVkT_VoWMdrIq1e7zyzi4/

What's short-lived? A second? A minute? A month?

elcallio:
> Seems like an unimportant edge case. Certificate expiration dates are measured in years, no?

Technically they can be very short-lived. Most server certs are of course at least order of days. Auth, maybe much less.

ScyllaPiotr:
> > > Certificate expiration dates are measured in years, no?
> > 
> > 
> > We want to enable enable short-lived certificates for https://docs.google.com/document/d/1gF1D8FgSBOec-yhMkhuVXUGVkT_VoWMdrIq1e7zyzi4/
> 
> What's short-lived? A second? A minute? A month?

The linked doc mentions _every 10 min_ as an example.

avikivity:
> > > > Certificate expiration dates are measured in years, no?
> > > 
> > > 
> > > We want to enable enable short-lived certificates for https://docs.google.com/document/d/1gF1D8FgSBOec-yhMkhuVXUGVkT_VoWMdrIq1e7zyzi4/
> > 
> > 
> > What's short-lived? A second? A minute? A month?
> 
> The linked doc mentions _every 10 min_ as an example.

Is it realistic? Do we have a mechanism to to distribute certificates so often?

What's accepted practice?

dkropachev:
> > > > > Certificate expiration dates are measured in years, no?
> > > > 
> > > > 
> > > > We want to enable enable short-lived certificates for https://docs.google.com/document/d/1gF1D8FgSBOec-yhMkhuVXUGVkT_VoWMdrIq1e7zyzi4/
> > > 
> > > 
> > > What's short-lived? A second? A minute? A month?
> > 
> > 
> > The linked doc mentions _every 10 min_ as an example.
> 
> Is it realistic? Do we have a mechanism to to distribute certificates so often?
> 
> What's accepted practice?

It all depends on customer case, flow is simple - before getting access to system/API/database app/script/user is authenticated, getting temporary certificate based on that authentication and then go do it's job, after times is out it either renew it's certificate or gets kicked out.
There is no standard regarding temporary certificate lifetime, but usually at the places where scylla is deployed it is 30-60 minutes.
But again, it could be 5-10 minutes easily.

avikivity:
> > > > > > Certificate expiration dates are measured in years, no?
> > > > > 
> > > > > 
> > > > > We want to enable enable short-lived certificates for https://docs.google.com/document/d/1gF1D8FgSBOec-yhMkhuVXUGVkT_VoWMdrIq1e7zyzi4/
> > > > 
> > > > 
> > > > What's short-lived? A second? A minute? A month?
> > > 
> > > 
> > > The linked doc mentions _every 10 min_ as an example.
> > 
> > 
> > Is it realistic? Do we have a mechanism to to distribute certificates so often?
> > What's accepted practice?
> 
> It all depends on customer case, flow is simple - before getting access to system/API/database app/script/user is authenticated, getting temporary certificate based on that authentication and then go do it's job, after times is out it either renew it's certificate or gets kicked out. There is no standard regarding temporary certificate lifetime, but usually at the places where scylla is deployed it is 30-60 minutes. But again, it could be 5-10 minutes easily.

"accepted practice" asks what others do, not on our plans. What's commonly done in other HTTP servers?

If it is common to re-validate expiration before each request, then it makes sense we adopt it too. But if no one does it then it doesn't make sense for us to do it. If many reissue certificates every 10 minutes, we can adopt it too. But if we're the only ones there's no need for us to innovate in this space.

nyh:
> > The linked doc mentions _every 10 min_ as an example.
> 
> Is it realistic? Do we have a mechanism to to distribute certificates so often?
> 
> What's accepted practice?

@scylladb-buff did you guys ever present the mTLS feature request to Avi?

@avikivity here's the short story about why users love mTLS so much, at least to my understanding. It hope it answers your questions above:

The traditional password-based approach to authentication (e.g., in Scylla) means that at any time, the server (let's call it Scylla) needs to know what are the allowed passwords. This is sometimes not great for security (for one example, see https://github.com/scylladb/scylladb/issues/5206). More importantly, it also means the user's administrator can't really be in charge of their own user accounts - if they ever create or expire a user, they need to update the server. With Scylla, it's not difficult to update the server every time an account is created or deleted (it can be done through CQL), but it's still inconvenient for users and not how many of them want to work in their modern environment.

Instead, the idea of mTLS is that the users' administrator doesn't tell us (the server) which accounts they allow. Instead, just give us a public key of a CA (certificate authority). They can then sign as many certificates as they want and hand them to users or applications, and don't need to tell us when they do so. You asked "Do we have a mechanism to to distribute certificates so often?" - but that's the wrong question: **we** (the server) don't have a mechanism to distribute certificates - it is the user who creates these certificates, as often (or not) as they want, for their own application.

You asked what's the point of temporary certificates lasting **minutes**, not years. Well, the answer is that it's needed for exactly the same reason AWS has STS (security token service): With certificate expiration, a customer can create a temporary certificate which expires after an hour (or much less), and hand it to some application started on some VM. This customer then knows that if their application is ever breached, all the attacker will get is a certificate good for one hour, not more.

Maybe now you're asking yourself this question: If we give a temporary certificate for 10 minutes, does it mean the server will need to create and delete "roles" (users) every 10 minutes? The answer is a resounding **no**, and exactly the goal of mTLS: Consider that you have a user called "nyh". The administrator can create a thousand different signed certificates, which each has the name "nyh" in the _subject_ of the certificate (CN or SAN). When Scylla receives any of these certificates, it authenticates the connection as "nyh". Scylla doesn't need to know when any of these certificates get created, nor when they expire.

scylladb-buff:
> > > The linked doc mentions _every 10 min_ as an example.
> > 
> > 
> > Is it realistic? Do we have a mechanism to to distribute certificates so often?
> > What's accepted practice?
> 
> @scylladb-buff did you guys ever present the mTLS feature request to Avi?

No, this was just the requirements and draft solution design. We sped up the technical design and PR to catch this window. 
It happens that this PR is the technical approval, bending the process I am not fan of, apologies @avikivity and everyone else.
Appreciate the review of my requirements and the technical validation. 

> 
> @avikivity here's the short story about why users love mTLS so much, at least to my understanding. It hope it answers your questions above:
> 
> The traditional password-based approach to authentication (e.g., in Scylla) means that at any time, the server (let's call it Scylla) needs to know what are the allowed passwords. This is sometimes not great for security (for one example, see [scylladb/scylladb#5206](https://github.com/scylladb/scylladb/issues/5206)). More importantly, it also means the user's administrator can't really be in charge of their own user accounts - if they ever create or expire a user, they need to update the server. With Scylla, it's not difficult to update the server every time an account is created or deleted (it can be done through CQL), but it's still inconvenient for users and not how many of them want to work in their modern environment.
> 
> Instead, the idea of mTLS is that the users' administrator doesn't tell us (the server) which accounts they allow. Instead, just give us a public key of a CA (certificate authority). They can then sign as many certificates as they want and hand them to users or applications, and don't need to tell us when they do so. You asked "Do we have a mechanism to to distribute certificates so often?" - but that's the wrong question: **we** (the server) don't have a mechanism to distribute certificates - it is the user who creates these certificates, as often (or not) as they want, for their own application.
> 
> You asked what's the point of temporary certificates lasting **minutes**, not years. Well, the answer is that it's needed for exactly the same reason AWS has STS (security token service): With certificate expiration, a customer can create a temporary certificate which expires after an hour (or much less), and hand it to some application started on some VM. This customer then knows that if their application is ever breached, all the attacker will get is a certificate good for one hour, not more.
> 
> Maybe now you're asking yourself this question: If we give a temporary certificate for 10 minutes, does it mean the server will need to create and delete "roles" (users) every 10 minutes? The answer is a resounding **no**, and exactly the goal of mTLS: Consider that you have a user called "nyh". The administrator can create a thousand different signed certificates, which each has the name "nyh" in the _subject_ of the certificate (CN or SAN). When Scylla receives any of these certificates, it authenticates the connection as "nyh". Scylla doesn't need to know when any of these certificates get created, nor when they expire.

Thanks @nyh and @dkropachev, @avikivity 

This is clarified in the requirements but the AI did a nice correct summary. 
We only need to validate the certificates from time to time and drop the connection if the certificate expired. 
This will allow customers to manage authentication using the ScyllaDB Cloud UI and their own PKE facilities. 

Not all customers will rotate client certificates that often (10m), but some industries will. 
I would say 30 min is example for short client lived certificate for most OIDC integrations. Our OKTA is less. 

The trust stores typically expire yearly but customers are free to rotate them as often as they practically - 3-12 months,
more than 12 is considered insecure

avikivity:
> > > The linked doc mentions _every 10 min_ as an example.
> > 
> > 
> > Is it realistic? Do we have a mechanism to to distribute certificates so often?
> > What's accepted practice?
> 
> @scylladb-buff did you guys ever present the mTLS feature request to Avi?
> 
> @avikivity here's the short story about why users love mTLS so much, at least to my understanding. 

I totally understand the desire of mTLS.  My questions about whether we really need to break an authenticated connection.

> It hope it answers your questions above:
> 
> The traditional password-based approach to authentication (e.g., in Scylla) means that at any time, the server (let's call it Scylla) needs to know what are the allowed passwords. This is sometimes not great for security (for one example, see [scylladb/scylladb#5206](https://github.com/scylladb/scylladb/issues/5206)). More importantly, it also means the user's administrator can't really be in charge of their own user accounts - if they ever create or expire a user, they need to update the server. With Scylla, it's not difficult to update the server every time an account is created or deleted (it can be done through CQL), but it's still inconvenient for users and not how many of them want to work in their modern environment.
> 
> Instead, the idea of mTLS is that the users' administrator doesn't tell us (the server) which accounts they allow. Instead, just give us a public key of a CA (certificate authority). They can then sign as many certificates as they want and hand them to users or applications, and don't need to tell us when they do so. You asked "Do we have a mechanism to to distribute certificates so often?" - but that's the wrong question: **we** (the server) don't have a mechanism to distribute certificates - it is the user who creates these certificates, as often (or not) as they want, for their own application.

I meant we as in "the ecosystem", not the server itself.

> 
> You asked what's the point of temporary certificates lasting **minutes**, not years. Well, the answer is that it's needed for exactly the same reason AWS has STS (security token service): With certificate expiration, a customer can create a temporary certificate which expires after an hour (or much less), and hand it to some application started on some VM. This customer then knows that if their application is ever breached, all the attacker will get is a certificate good for one hour, not more.
> 
> Maybe now you're asking yourself this question: If we give a temporary certificate for 10 minutes, does it mean the server will need to create and delete "roles" (users) every 10 minutes? The answer is a resounding **no**, and exactly the goal of mTLS: Consider that you have a user called "nyh". The administrator can create a thousand different signed certificates, which each has the name "nyh" in the _subject_ of the certificate (CN or SAN). When Scylla receives any of these certificates, it authenticates the connection as "nyh". Scylla doesn't need to know when any of these certificates get created, nor when they expire.

Will the client know to abandon those connections? If the driver doesn't know, then it will get broken requests if it reuses a connection that is about to be broken.

Are there standard http clients that are aware of those short-expiration certificates?

scylladb-buff:
> > > > The linked doc mentions _every 10 min_ as an example.
> > > 
> > > 
> > > Is it realistic? Do we have a mechanism to to distribute certificates so often?
> > > What's accepted practice?
> > 
> > 
> > @scylladb-buff did you guys ever present the mTLS feature request to Avi?
> > @avikivity here's the short story about why users love mTLS so much, at least to my understanding.
> 
> I totally understand the desire of mTLS. My questions about whether we really need to break an authenticated connection.
> 
> > It hope it answers your questions above:
> > The traditional password-based approach to authentication (e.g., in Scylla) means that at any time, the server (let's call it Scylla) needs to know what are the allowed passwords. This is sometimes not great for security (for one example, see [scylladb/scylladb#5206](https://github.com/scylladb/scylladb/issues/5206)). More importantly, it also means the user's administrator can't really be in charge of their own user accounts - if they ever create or expire a user, they need to update the server. With Scylla, it's not difficult to update the server every time an account is created or deleted (it can be done through CQL), but it's still inconvenient for users and not how many of them want to work in their modern environment.
> > Instead, the idea of mTLS is that the users' administrator doesn't tell us (the server) which accounts they allow. Instead, just give us a public key of a CA (certificate authority). They can then sign as many certificates as they want and hand them to users or applications, and don't need to tell us when they do so. You asked "Do we have a mechanism to to distribute certificates so often?" - but that's the wrong question: **we** (the server) don't have a mechanism to distribute certificates - it is the user who creates these certificates, as often (or not) as they want, for their own application.
> 
> I meant we as in "the ecosystem", not the server itself.
> 
> > You asked what's the point of temporary certificates lasting **minutes**, not years. Well, the answer is that it's needed for exactly the same reason AWS has STS (security token service): With certificate expiration, a customer can create a temporary certificate which expires after an hour (or much less), and hand it to some application started on some VM. This customer then knows that if their application is ever breached, all the attacker will get is a certificate good for one hour, not more.
> > Maybe now you're asking yourself this question: If we give a temporary certificate for 10 minutes, does it mean the server will need to create and delete "roles" (users) every 10 minutes? The answer is a resounding **no**, and exactly the goal of mTLS: Consider that you have a user called "nyh". The administrator can create a thousand different signed certificates, which each has the name "nyh" in the _subject_ of the certificate (CN or SAN). When Scylla receives any of these certificates, it authenticates the connection as "nyh". Scylla doesn't need to know when any of these certificates get created, nor when they expire.
> 
> Will the client know to abandon those connections? If the driver doesn't know, then it will get broken requests if it reuses a connection that is about to be broken.
> 
> Are there standard http clients that are aware of those short-expiration certificates?

Some products authenticate only when the connections established. 
This is not optimal since it allow some users with expired certificates to continue to use the connection. 
The better way is for the drivers to refresh certificates shortly before they actually expire and if they can't verify them, they should drop the connection.
The access for the user will be denied.

nyh:
Pushed a new version to fix conflicts. No functional changes - in particular it doesn't answer the doubts raised by some reviewers above whether this patch is a good idea or not. I think it is and will try to answer this question later.
Also, I want to add a test for this feature - which this PR is still missing.

nyh:
Added a second patch with tests for client-certificate expiration - both the case of a certificate expiring while a connection is already alive (the check implemented in the first patch), and the simpler case of a connection being attempted when the client-certificate is already expired.

PR #3523, opened 2026-07-09
http: reject requests with oversized URL or headers

### LINKED ISSUE #2698: HTTP server should reject requests with oversized URL or headers
Seastar's HTTP server reads each header as a contiguous string, and saves all of them in memory before calling the application-defined handler. A malicious user can force Seastar to allocate unlimited amount of contiguous memory (for a single header) or non-contiguous memory (with many headers), and potentially crash the application with an OOM.  The same problem also exists for the request line (the URL). It also needs to be limited, for the same reason.

I am marking this issue a "bug" and not an "enhancement" because this issue can be used to OOM any Seastar application with any sort of HTTP server or REST API, and because the bug must be fixed in Seastar and can't be fixed or worked around by the application. https://github.com/scylladb/scylladb/issues/23438 is for example a ScyllaDB issue caused by this issue.

This missing feature must be implemented in Seastar - it can't be implemented in the application using Seastar because the application's handler gets called only after the URL and the headers have been fully read into memory by Seastar.

Other HTTP servers typically have very low length limits for both the URL and the headers. Apache Httpd defaults to a limit of 8 KB for each, and AWS DynamoDB's server has a limit of 16 KB. Larger URLs or headers are not useful, and there is no interesting reason to support them - although if we wish we could make these limits configurable like Apache Httpd does. When the limit is exceeded, servers typically return a 400 Bad Request sometimes accompanied by a message "Request Header or Cookie Too Large".

raphaelroshan:
Fixes #2698.

The HTTP request parser buffered the request line and all headers with no upper bound, so a client could force unbounded allocation — one huge header (contiguous) or many headers (non-contiguous) — and OOM the server before the handler runs. The handler cannot guard against this since it only sees a fully-parsed request.

Track the bytes consumed for the request line + headers in the parser and reject once a configurable limit is exceeded. The limit is a single combined bound on the request line and headers, defaulting to 32 KiB and settable via `http_server::set_request_size_limit()`, mirroring the existing `content_length_limit` plumbing. Adds parser-level (single-buffer + scattered) and end-to-end tests.

Update addressing review: an oversized request now also makes the parser `failed()` return true (callers check one condition; `size_limit_exceeded()` explains why). The response is 414 (URI Too Long) when the URI overflows and 431 (Request Header Fields Too Large) otherwise, distinguished by the recorded request-line length; adds the 431 `status_type`.

### DIFF
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -201,6 +201,7 @@ void connection::generate_error_reply_and_close(std::unique_ptr<http::request> r
 
 future<> connection::read_one() {
     _parser.init();
+    _parser.set_size_limit(_server.get_request_size_limit());
     return _read_buf.consume(_parser).then([this] () mutable {
         if (_parser.eof()) {
             _done = true;
@@ -214,6 +215,13 @@ future<> connection::read_one() {
 
         if (_tls) {
             req->protocol_name = "https";
+        }
+        if (_parser.size_limit_exceeded()) {
+            if (req->_version.empty()) {
+                req->_version = "1.1";
+            }
+            generate_error_reply_and_close(std::move(req), http::reply::status_type::bad_request, "Request header or URL too large");
+            return make_ready_future<>();
         }
         if (_parser.failed()) {
             if (req->_version.empty()) {
@@ -370,6 +378,14 @@ size_t http_server::get_content_length_limit() const {
 
 void http_server::set_content_length_limit(size_t limit) {
     _content_length_limit = limit;
+}
+
+size_t http_server::get_request_size_limit() const {
+    return _request_size_limit;
+}
+
+void http_server::set_request_size_limit(size_t limit) {
+    _request_size_limit = limit;
 }
 
 bool http_server::get_content_streaming() const {
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -216,17 +216,17 @@ future<> connection::read_one() {
         if (_tls) {
             req->protocol_name = "https";
         }
-        if (_parser.size_limit_exceeded()) {
-            if (req->_version.empty()) {
-                req->_version = "1.1";
-            }
-            generate_error_reply_and_close(std::move(req), http::reply::status_type::bad_request, "Request header or URL too large");
-            return make_ready_future<>();
-        }
         if (_parser.failed()) {
             if (req->_version.empty()) {
                 // we might have failed to parse even the version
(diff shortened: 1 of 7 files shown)


src/http/httpd.cc
```diff
@@ -215,6 +216,13 @@ future<> connection::read_one() {
         if (_tls) {
             req->protocol_name = "https";
         }
+        if (_parser.size_limit_exceeded()) {
+            if (req->_version.empty()) {
+                req->_version = "1.1";
+            }
+            generate_error_reply_and_close(std::move(req), http::reply::status_type::bad_request, "Request header or URL too large");
```
xemul: The http-level error doesn't look correct for this case. The 414 or 431 would be more appropriate

src/http/request_parser.rl
```diff
@@ -183,6 +201,9 @@ public:
     bool failed() const {
         return _state == state::error;
```
xemul: So oversized request doesn't "fail" the parser. This doesn't look like correct design decision and forces the user to check _both_ conditions to check if the parsing succeeded or not

raphaelroshan:
Thanks for the review @xemul — addressed both:
  - Parser now fails on oversize (failed() covers the too-large state), so it's a single check; size_limit_exceeded() just explains why.
  - Returns 414 for an oversized URI, 431 for oversized headers (split by the recorded request-line length, so it's correct even when the whole request lands in one buffer). Added the missing 431 status_type.

tests/unit/request_parser_test.cc
```diff
@@ -72,3 +72,59 @@ SEASTAR_TEST_CASE(test_header_parsing) {
...
+    auto feed = [] (http_request_parser& parser, const sstring& s) {
+        return parser(temporary_buffer<char>(s.c_str(), s.size())).get();
+    };
+
+    http_request_parser parser;
+    parser.init();
+    parser.set_size_limit(512);
+    BOOST_REQUIRE(!feed(parser, part1).has_value());
+    BOOST_REQUIRE(!parser.size_limit_exceeded());
+    feed(parser, part2);
```
xemul: This is a good place to check that the parser actually stopped consuming the request, not just flagged that it is (was) oversized. However, to make it work properly the PR needs to be rebased to pick up the 768516c69

xemul (approved):
ACK, but needs a rebase (and a test fixlet)
Also, when rebasing, please, squash all commits into one

PR #3524, opened 2026-07-09
net: fix conntrack close accounting

### LINKED ISSUE #3514: Posix connection load balancer selects shard 0 too much

**The bug**

```c++
// smp.hh
if (t == this_shard_id()) {
} else {
}
```

**Possible fixes**

ScyllaPiotr:
Replace conntrack's shard load counters with relaxed atomic counters. These are used to decrement the selected shard's counter directly from the connection handle destructor instead of posting the decrement back to the accepting shard by `smp::submit_to()`.

Also avoid repeatedly preferring shard 0 on non-empty equal-load ties. The first zero-load connection still stays on the current shard to preserve existing same-shard accept behavior.

Fixes #3514

### DIFF
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -20,6 +20,7 @@
  */
 
 #pragma once
+#include <atomic>
 #include <unordered_set>
 #include <seastar/core/sharded.hh>
 #include <seastar/core/internal/pollable_fd.hh>
@@ -59,41 +60,49 @@ using namespace seastar;
 // Right now this class is used by the posix_server_socket_impl, but it could be used by any other.
 class conntrack {
     class load_balancer {
-        std::vector<unsigned> _cpu_load;
+        // Tracks the number of connections assigned to each shard. The accepting shard
+        // increments the selected shard's counter; the shard that closes the connection
+        // decrements it directly from the connection handle.
+        std::vector<std::atomic<unsigned>> _cpu_load;
     public:
-        load_balancer() : _cpu_load(size_t(this_smp_shard_count()), 0) {}
-        void closed_cpu(shard_id cpu) {
-            _cpu_load[cpu]--;
-        }
+        load_balancer() : _cpu_load(size_t(this_smp_shard_count())) {}
         shard_id next_cpu() {
             // FIXME: The naive algorithm will just round robin the connections around the shards.
             // A more complex version can keep track of the amount of activity in each connection,
             // and use that information.
-            auto min_el = std::min_element(_cpu_load.begin(), _cpu_load.end());
-            auto cpu = shard_id(std::distance(_cpu_load.begin(), min_el));
-            _cpu_load[cpu]++;
-            return cpu;
+            shard_id cpu_number = 0;
+            auto min_load = _cpu_load[0].load(std::memory_order_relaxed);
+            for (shard_id i = 1; i < _cpu_load.size(); ++i) {
+                auto load = _cpu_load[i].load(std::memory_order_relaxed);
+                if (load < min_load) {
+                    min_load = load;
+                    cpu_number = i;
+                }
+            }
+            _cpu_load[cpu_number].fetch_add(1, std::memory_order_relaxed);
+            return cpu_number;
         }
         shard_id force_cpu(shard_id cpu) {
-            _cpu_load[cpu]++;
+            _cpu_load[cpu].fetch_add(1, std::memory_order_relaxed);
             return cpu;
+        }
+        // For each shard to then store its count of connections
+        std::atomic<unsigned>* get_counter_ptr(shard_id cpu) {
+            return &_cpu_load[cpu];
         }
     };
 
     lw_shared_ptr<load_balancer> _lb;
-    void closed_cpu(shard_id cpu) {
-        _lb->closed_cpu(cpu);
-    }
 public:
     class handle {
-        shard_id _host_cpu;
         shard_id _target_cpu;
+        std::atomic<unsigned>* _connection_counter = nullptr;
         foreign_ptr<lw_shared_ptr<load_balancer>> _lb;
     public:
         handle() : _lb(nullptr) {}
         handle(shard_id cpu, lw_shared_ptr<load_balancer> lb)
-            : _host_cpu(this_shard_id())
-            , _target_cpu(cpu)
+            : _target_cpu(cpu)
+            , _connection_counter(lb->get_counter_ptr(cpu))
             , _lb(make_foreign(std::move(lb))) {}
 
         handle(const handle&) = delete;
@@ -104,10 +113,7 @@ public:
             if (!_lb) {
                 return;
             }
-            // FIXME: future is discarded
-            (void)smp::submit_to(_host_cpu, [cpu = _target_cpu, lb = std::move(_lb)] {
-                lb->closed_cpu(cpu);
-            });
+            _connection_counter->fetch_sub(1, std::memory_order_relaxed);
         }
         shard_id cpu() {
             return _target_cpu;
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -72,9 +72,14 @@ class conntrack {
             // and use that information.
             shard_id cpu_number = 0;
             auto min_load = _cpu_load[0].load(std::memory_order_relaxed);
+            // Prefer current shard for the first connection.
+            if (min_load == 0) {
+                _cpu_load[0].fetch_add(1, std::memory_order_relaxed);
+                return 0;
+            }
             for (shard_id i = 1; i < _cpu_load.size(); ++i) {
                 auto load = _cpu_load[i].load(std::memory_order_relaxed);
-                if (load < min_load) {
+                if (load <= min_load) {
                     min_load = load;
                     cpu_number = i;
                 }


ScyllaPiotr:
I tried to validate this by re-running [the reproducer](https://github.com/scylladb/seastar/issues/3514#issuecomment-4912858450), comparing current `origin/master` with this PR commit.

Setup:

```text
before: origin/master, 4a23ff87
 after: PR #3524, 654f8eee332008c84a255551a5478d51ffd850b1
build: release
cmd: build/release/tests/perf/http_shard_balance_perf --smp 4 --cpuset 0-3 --requests-per-shard 100000 --concurrency 10
```

I also temporarily instrumented `conntrack::load_balancer` for both versions to print the selected target shard counts. Those `conntrack assignments` matched the HTTP handler's per-shard request counts exactly, so the numbers below reflect conntrack shard selection, not later HTTP dispatch skew.

Before this PR:

```text
conntrack assignments: 0=143392 1=79535 2=77199 3=99874

Results: requests received per shard
shard          requests     errors
----------------------------------
0                143392          0
1                 79535          0
2                 77199          0
3                 99874          0
----------------------------------
total            400000          0

Expected per shard (ideal):  100000
Shard 0 ratio vs ideal:      1.43x
```

After this PR:

```text
conntrack assignments: 0=134235 1=84287 2=79129 3=102349

Results: requests received per shard
shard          requests     errors
----------------------------------
0                134235          0
1                 84287          0
2                 79129          0
3                102349          0
----------------------------------
total            400000          0

Expected per shard (ideal):  100000
Shard 0 ratio vs ideal:      1.34x
```

I also ran several shorter/full variants. The PR generally reduced shard-0 over-selection, but not enough to call the issue fixed. In my local environment the reproducer still shows significant residual imbalance after the atomic close-accounting change. So this PR addresses one plausible source of stale accounting, but the representative result suggests the main imbalance, or at least another substantial part of it, lies elsewhere...

ScyllaPiotr:
Re-ran Nadav's published HTTP reproducer from #3514 against `origin/master` and this PR **after adding the latest tie-breaker commit**.

Setup:

```text
before: origin/master, 4a23ff87
after: PR #3524 with a91c8ad3
build: release
cmd: build/release/tests/perf/http_shard_nadav_balance_perf --smp 4 --cpuset 0-3 --requests-per-shard 100000 --concurrency 10
```

The proof worktrees also had temporary `conntrack::load_balancer` instrumentation. The `conntrack assignments` matched HTTP handler per-shard request counts, so the numbers below reflect conntrack target selection.

Before this PR, run 1:

```text
Results: requests received per shard
shard          requests     errors
----------------------------------
0                144405          0
1                 86324          0
2                 82017          0
3                 87254          0
----------------------------------
total            400000          0

Expected per shard (ideal): 100000
Shard 0 ratio vs ideal: 1.44x
conntrack assignments: 0=144405 1=86324 2=82017 3=87254
conntrack tie assignments: 0=112095 1=66267 2=48991 3=0 total=227353
```

After this PR, run 1:

```text
Results: requests received per shard
shard          requests     errors
----------------------------------
0                121259          0
1                 76630          0
2                 77341          0
3                124770          0
----------------------------------
total            400000          0

Expected per shard (ideal): 100000
Shard 0 ratio vs ideal: 1.21x
conntrack assignments: 0=121259 1=76630 2=77341 3=124770
conntrack tie assignments: 0=0 1=55851 2=62008 3=92227 total=210086
```

Sequential repeat, before:

```text
conntrack assignments: 0=148083 1=86925 2=83267 3=81725
Shard 0 ratio vs ideal: 1.48x
conntrack tie assignments: 0=113498 1=64517 2=47680 3=0 total=225695
```

Sequential repeat, after:

```text
conntrack assignments: 0=118287 1=75576 2=76669 3=129468
Shard 0 ratio vs ideal: 1.18x
conntrack tie assignments: 0=1 1=54473 2=61331 3=94353 total=210158
```

I also reran the variant Nadav mentioned in the issue discussion:

```text
cmd: build/release/tests/perf/http_shard_nadav_balance_perf --smp 4 --cpuset 0-3 --requests-per-shard 100000 --concurrency 300 --handler-sleep 5
```

Before:

```text
conntrack assignments: 0=120593 1=80677 2=82077 3=116653
Shard 0 ratio vs ideal: 1.21x
conntrack tie assignments: 0=84645 1=58675 2=51492 3=0 total=194812
```

After:

```text
conntrack assignments: 0=112204 1=81193 2=83221 3=123382
Shard 0 ratio vs ideal: 1.12x
conntrack tie assignments: 0=1 1=45284 2=52765 3=79426 total=177476
```

Takeaways:

* Nadav's reproducer still reproduces the shard-0 skew before the PR.
* This PR substantially reduces shard-0 over-selection in the published no-sleep shape: about `1.44x-1.48x` before, about `1.18x-1.21x` after.
* The tie-breaker commit is doing what it is intended to do: tied minimum-load selections no longer go to shard 0. They shift mostly to the highest-numbered shard.
* The result is not perfectly balanced; it intentionally avoids shard 0 on ties rather than implementing randomized/rotating tie breaking.
* The `0=1` is likely a test fluke, as shard 0 is impossible to be selected for a tie at all.

ScyllaPiotr:
Additional validation with longer request lifetime of 10ms. The goal was to check whether shard-0 skew remains when connection lifetime is much longer than shard handoff latency. With `--handler-sleep 10`, the remaining bias mostly shifts to the highest-numbered shard, as expected from the current `<=` tie policy.

Command, concurrency 300:

```text
build/release/tests/perf/http_shard_nadav_balance_perf --smp 4 --cpuset 0-3 --requests-per-shard 100000 --concurrency 300 --handler-sleep 10
```

Result:

```text
conntrack assignments: 0=107360 1=91069 2=91714 3=109857
Shard 0 ratio vs ideal: 1.07x
conntrack tie assignments: 0=1 1=47533 2=56779 3=76445 total=180758
```

Command, concurrency 600:

```text
build/release/tests/perf/http_shard_nadav_balance_perf --smp 4 --cpuset 0-3 --requests-per-shard 100000 --concurrency 600 --handler-sleep 10
```

Result:

```text
conntrack assignments: 0=110458 1=87507 2=89189 3=112846
Shard 0 ratio vs ideal: 1.10x
conntrack tie assignments: 0=1 1=46759 2=55799 3=76586 total=179145
```

So with 10ms handler lifetime there still is some preference of shard 0, albeit small (`1.07x` to `1.10x`).

ScyllaPiotr:
I added proof-only instrumentation to classify selections from the exact load values observed during `conntrack::load_balancer::next_cpu()`, instead of classifying ties by rereading the atomics after target selection. This avoids the race in the previous diagnostic.

With the current PR state, shard 0 is not being selected from true equal-min ties. It is selected when the observed live conntrack load makes it a unique minimum.

Run 1:

```text
build/release/tests/perf/http_shard_nadav_balance_perf \
  --smp 4 --cpuset 0-3 \
  --port 12310 \
  --requests-per-shard 100000 \
  --concurrency 300 \
  --handler-sleep 10
```

```text
conntrack assignments:                  0=106566 1=89495 2=91966 3=111973
conntrack exact tie assignments:        0=0      1=47332 2=57204 3=76718  total=181254
conntrack exact unique-min assignments: 0=106566 1=42163 2=34762 3=35255
conntrack lifetime avg:                 0=13.40ms 1=16.31ms 2=15.87ms 3=12.94ms
```

Run 2:

```text
build/release/tests/perf/http_shard_nadav_balance_perf \
  --smp 4 --cpuset 0-3 \
  --port 12320 \
  --requests-per-shard 100000 \
  --concurrency 600 \
  --handler-sleep 10
```

```text
conntrack assignments:                  0=109273 1=87939 2=87033 3=115755
conntrack exact tie assignments:        0=0      1=46799 2=54528 3=78359  total=179686
conntrack exact unique-min assignments: 0=109273 1=41140 2=32505 3=37396
conntrack lifetime avg:                 0=14.45ms 1=18.42ms 2=18.65ms 3=13.86ms
```

Interpretation:

Connections live significantly longer on shard 1 and shard 2 (16.31ms, 15.87ms, 18.42ms and 18.65ms on shard 1 and shard 2 vs. 13.40ms, 12.94ms, 14.45ms and 13.86ms on shard 0 and shard 3) . This makes the total number of connections processed by them smaller - they are less frequently the ones with the smallest number of connections.

include/seastar/net/posix-stack.hh
```diff
@@ -59,41 +60,49 @@ using namespace seastar;
 // Right now this class is used by the posix_server_socket_impl, but it could be used by any other.
 class conntrack {
     class load_balancer {
-        std::vector<unsigned> _cpu_load;
+        // Tracks the number of connections assigned to each shard. The accepting shard
+        // increments the selected shard's counter; the shard that closes the connection
+        // decrements it directly from the connection handle.
+        std::vector<std::atomic<unsigned>> _cpu_load;
```
avikivity, line 66: This is death on large multicores, every decrement will bounce 32 different counters

avikivity:
> significantly

quantify

ScyllaPiotr:
> > significantly
> 
> quantify

16.31ms, 15.87ms, 18.42ms and 18.65ms on shard 1 and shard 2 vs. 13.40ms, 12.94ms, 14.45ms and 13.86ms on shard 0 and shard 3.
Corrected the original place too.

avikivity:
At a minimum, we need sharded\<std::atomic> here so the counters don't share cache lines.

However, it will still be painful for shard 0 to collect the statistics on a large machine. If every shard processes 100 conn/sec (shouldn't really happen) We can see 50,000 on shard 0. Each reasd 500 counters so it's 2.5M atomic reads/sec on one shard, each taking potentially hundreds of cycles with NUMA.

ScyllaPiotr:
Thanks.
I have created an alternative fix https://github.com/scylladb/seastar/pull/3526 which improves the earlier solution instead of substituting it.

PR #3525, opened 2026-07-09
net/tls: send TLS alert when client certificate verification fails

### LINKED ISSUE #912: TLS handshake failure should return an alert - not just silently close the connection
This problem was first noticed in the Scylla project, https://github.com/scylladb/scylla/issues/8827:

If we configure TLS to accept only TLSv1.2, then a client trying to connect with an earlier version of TLS will, rightly, fail to connect. However, currently the client will just see the connection abruptly disconnected, without getting any error message. For example, Python's SSL functions report that "EOF occurred in violation of protocol" - which tells the client nothing about the nature of the problem.

The problem is that the `gnutls_handshake()` function which we use in our TLS implementation indeed doesn't return a reply to the client! The caller (us) is supposed to use `gnutls_alert_send()` to send an alert to the client. We didn't do that. In fact, it seems we don't use `gnutls_alert_send()` anywhere in Seastar.

There have been many different alert numbers specified over the evolution of SSL/TLS, but the current RFC - https://datatracker.ietf.org/doc/html/rfc8446 - suggests the right alert should be the "protocol_version" alert. Indeed if this alert is returned, Python reports as a connection error: "[SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version". This is perhaps not the clearest error message in the world, but at least contains the words "protocol version" hinting what the problem is, and can be searched in the RFC or Google for the meaning of this message.

nyh:
When a TLS server rejects a client certificate, the peer was seeing an abrupt and unexplained connection close instead of a proper TLS alert.

gnutls_session_set_verify_function() was only installed for CLIENT sessions (where Seastar verifies the server's cert). For SERVER sessions, gnutls_handshake() succeeds even when the client presents a bad certificate. The certificate check was then done post-handshake by calling verify() directly, but if verify() throws, the exception propagated to the outer catch() block with no TLS alert sent.

Additionally, the GNUTLS_E_CERTIFICATE_ERROR path (reached for CLIENT sessions via verify_wrapper) also swallowed the exception without first sending an alert.

Fix:
- Add cert_status_to_alert() to map gnutls_certificate_verify_peers3 status flags to the semantically correct TLS alert description: GNUTLS_A_UNKNOWN_CA for an untrusted signer, GNUTLS_A_CERTIFICATE_REVOKED for revoked certs, GNUTLS_A_CERTIFICATE_EXPIRED for expired certs, and GNUTLS_A_BAD_CERTIFICATE for other failures.
- Extend verify() with an optional out_alert parameter that is set to the appropriate alert before a verification_error is thrown.
- Add verify_and_send_alert() helper that calls verify(), and when it fails, flushes pending output then calls send_alert() before propagating the exception. Returns nullopt on success so call sites can fall through normally.
- Use verify_and_send_alert() in both the GNUTLS_E_CERTIFICATE_ERROR switch case and the post-handshake verify() call.

This is similar to commit 0e48ba883 ("net/tls: on TLS handshake failure, send error to client", fixes #912) which added send_alert() for other handshake failures, but was not applied to certificate verification failures.

### DIFF
--- a/src/net/tls_gnutls.cc
+++ b/src/net/tls_gnutls.cc
@@ -575,6 +575,31 @@ public:
         return s;
     }
 
+    // Calls verify() and, if verification fails, sends the appropriate TLS alert
+    // to the peer before propagating the exception. Returns nullopt when
+    // verification succeeded (caller should continue normally), or a future
+    // that sends the alert and resolves to the verification error.
+    std::optional<future<>> verify_and_send_alert() {
+        gnutls_alert_description_t cert_alert = GNUTLS_A_BAD_CERTIFICATE;
+        std::exception_ptr vex;
+        try {
+            verify(&cert_alert);
+        } catch (...) {
+            vex = std::current_exception();
+        }
+        if (!vex) {
+            return std::nullopt;
+        }
+        _error = vex;
+        return wait_for_output().then_wrapped([this, cert_alert, vex = std::move(vex)](future<> f) mutable {
+            f.ignore_ready_future();
+            return send_alert(GNUTLS_AL_FATAL, cert_alert).then_wrapped([vex = std::move(vex)](future<> f) mutable {
+                f.ignore_ready_future();
+                return make_exception_future<>(std::move(vex));
+            });
+        });
+    }
+
     future<> send_alert(gnutls_alert_level_t level, gnutls_alert_description_t desc) {
         return repeat([this, level, desc]() {
             auto res = gnutls_alert_send(*this, level, desc);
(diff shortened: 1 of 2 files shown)


nyh:
This PR was split from https://github.com/scylladb/seastar/pull/3517, which was already reviewed by @knowack1 and @m-szymon. Their comments where:

1. Consider adding unit test for this problem.
2. We have 'case GNUTLS_E_CERTIFICATE_ERROR:'. shouldn't GNUTLS_E_NO_CERTIFICATE_FOUND case send alert as well?

nyh:
I (and my AI) is having trouble creating a unit test for this fix. I've managed to create a very unnatural situation where this patch makes a difference, but it doesn't seem my patch is needed for the more "natural" failure cases. So I'm not exactly sure why my patch is needed. In any case, it's not criticial for the mTLS feature, I just wanted it to improve error handling in tests. So let's put this PR on hold.

PR #3526, opened 2026-07-09
posix: improve conntrack shard balancing

### LINKED ISSUE #3514: Posix connection load balancer selects shard 0 too much

ScyllaPiotr:
1. Pin single-shard test servers to the invoking shard via `listen_options::set_fixed_cpu()`. Before this PR these tests only pass because the balancer resolves the all-idle tie to shard 0, which the last patch here changes.
2. Defer local connection close accounting so local and remote close reports have comparable, delayed visibility. The observable result was shard 0 receiving the majority of new connections under short-lived connection churn.
3. Cache the least-loaded shard not to recalculate for every new connection, with the TTL at 2% of the shard's currently open connections, additionally capped at 10.
4. Avoid lower-numbered shard preference on connection count ties.

Fixes #3514.
Alternative to #3524.

Note: the `Seastar.unit.prometheus` CI failure is a pre-existing bug in the prometheus handler (scrape label filtering depends on the serving shard) that the changed tie-breaking exposes; it is being fixed separately (#3630 / #3649). This PR depends on either of them landing first.

## Measurements
Two binaries were measured: `master` = master at `50015067`, and `#3526` = this PR's four patches applied on top of that same commit, so the only difference between them is this PR. Both release mode, measured back-to-back on the same machine, 4 shards (`--smp 4 --cpuset 0-3`). **ideal** = total connections / 4, i.e. a perfectly equal per-shard share; the counts column shows the per-shard tally of the median-CoV wave.

**HTTP rows**: connection-per-request HTTP load (each request opens a fresh connection, closed after the response) - 5 waves x 400k requests (3 x 40k for the 10ms rows); *handler* = server-side handler delay, so it is ~ each connection's lifetime; *bg* = long-lived idle connections held open per shard throughout, verified exact before measurement. **TCP rows**: raw TCP connections (no HTTP) arriving at the given rate for 30 s, each closed by the server after 10 ms; counts are accepted connections per shard - same accept-and-distribute path as HTTP, swept by arrival rate.

| workload | tree | connection counts [s0,s1,s2,s3] | s0/ideal | max/ideal | CoV |
|---|---|---|---:|---:|---:|
| HTTP conn-per-request, 0ms handler, no bg conns | master | `[151380, 63193, 58765, 126662]` | 1.514 | 1.514 | 0.400 |
| HTTP conn-per-request, 0ms handler, no bg conns | #3526 | `[134417, 59351, 63090, 143142]` | 1.344 | 1.431 | 0.389 |
| HTTP conn-per-request, 0ms handler, 4 bg conns/shard | master | `[147076, 62127, 61032, 129765]` | 1.471 | 1.471 | 0.389 |
| HTTP conn-per-request, 0ms handler, 4 bg conns/shard | #3526 | `[133074, 60771, 64504, 141651]` | 1.331 | 1.417 | 0.375 |
| HTTP conn-per-request, 0ms handler, 40 bg conns/shard | master | `[145208, 63015, 61397, 130380]` | 1.452 | 1.452 | 0.382 |
| HTTP conn-per-request, 0ms handler, 40 bg conns/shard | #3526 | `[132612, 59590, 64337, 143461]` | 1.326 | 1.435 | 0.383 |
| HTTP conn-per-request, 5ms handler, no bg conns | master | `[101563, 99735, 99421, 99281]` | 1.016 | 1.016 | 0.009 |
| HTTP conn-per-request, 5ms handler, no bg conns | #3526 | `[100139, 99351, 99833, 100677]` | 1.001 | 1.007 | 0.005 |
| HTTP conn-per-request, 5ms handler, 4 bg conns/shard | master | `[101573, 99868, 99478, 99081]` | 1.016 | 1.016 | 0.009 |
| HTTP conn-per-request, 5ms handler, 4 bg conns/shard | #3526 | `[100173, 99326, 99808, 100693]` | 1.002 | 1.007 | 0.005 |
| HTTP conn-per-request, 5ms handler, 40 bg conns/shard | master | `[101620, 99782, 99410, 99188]` | 1.016 | 1.016 | 0.010 |
| HTTP conn-per-request, 5ms handler, 40 bg conns/shard | #3526 | `[97967, 99131, 100670, 102232]` | 0.980 | 1.022 | 0.016 |
| HTTP conn-per-request, 10ms handler, no bg conns | master | `[10160, 9989, 9946, 9905]` | 1.016 | 1.016 | 0.010 |
| HTTP conn-per-request, 10ms handler, no bg conns | #3526 | `[9997, 9958, 9992, 10053]` | 1.000 | 1.005 | 0.003 |
| HTTP conn-per-request, 10ms handler, 4 bg conns/shard | master | `[10148, 9994, 9944, 9914]` | 1.015 | 1.015 | 0.009 |
| HTTP conn-per-request, 10ms handler, 4 bg conns/shard | #3526 | `[9979, 9942, 9994, 10085]` | 0.998 | 1.008 | 0.005 |
| HTTP conn-per-request, 10ms handler, 40 bg conns/shard | master | `[10162, 9983, 9939, 9916]` | 1.016 | 1.016 | 0.010 |
| HTTP conn-per-request, 10ms handler, 40 bg conns/shard | #3526 | `[9776, 9927, 10106, 10191]` | 0.978 | 1.019 | 0.016 |
| raw TCP, 1000 conn/s, 10ms lifetime | master | `[7506, 7500, 7500, 7494]` | 1.001 | 1.001 | 0.001 |
| raw TCP, 1000 conn/s, 10ms lifetime | #3526 | `[7497, 7500, 7503, 7500]` | 1.000 | 1.000 | 0.000 |
| raw TCP, 2000 conn/s, 10ms lifetime | master | `[16553, 15114, 14647, 13686]` | 1.104 | 1.104 | 0.069 |
| raw TCP, 2000 conn/s, 10ms lifetime | #3526 | `[14121, 14143, 15103, 16633]` | 0.941 | 1.109 | 0.068 |
| raw TCP, 4000 conn/s, 10ms lifetime | master | `[31799, 30056, 29167, 28978]` | 1.060 | 1.060 | 0.037 |
| raw TCP, 4000 conn/s, 10ms lifetime | #3526 | `[28952, 29606, 30226, 31216]` | 0.965 | 1.041 | 0.028 |
| raw TCP, 8000 conn/s, 10ms lifetime | master | `[61782, 60280, 59057, 58881]` | 1.030 | 1.030 | 0.019 |
| raw TCP, 8000 conn/s, 10ms lifetime | #3526 | `[58739, 59996, 60195, 61070]` | 0.979 | 1.018 | 0.014 |

Under short-lived-connection churn, #3526 removes the systematic shard-0 concentration: s0/ideal drops from 1.45-1.51 to 1.33-1.34 in extreme churn and from 1.016 to ~1.00 at realistic lifetimes. The residual tie-break surplus lands on the highest-numbered shard instead (max/ideal <= 1.44 in extreme churn, <= 1.02 elsewhere; visible as the mirrored counts in the 2000 conn/s row).

### DIFF
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -25,6 +25,7 @@
 #include <unordered_set>
 #include <utility>
 #include <seastar/core/sharded.hh>
+#include <seastar/core/make_task.hh>
 #include <seastar/core/internal/pollable_fd.hh>
 #include <seastar/net/stack.hh>
 #include <seastar/core/polymorphic_temporary_buffer.hh>
@@ -129,12 +130,14 @@ public:
     class handle {
         shard_id _host_cpu;
         shard_id _target_cpu;
+        bool _defer_local_close = false;
         foreign_ptr<lw_shared_ptr<load_balancer>> _lb;
     public:
         handle() : _lb(nullptr) {}
-        handle(shard_id cpu, lw_shared_ptr<load_balancer> lb)
+        handle(shard_id cpu, lw_shared_ptr<load_balancer> lb, bool defer_local_close)
             : _host_cpu(this_shard_id())
             , _target_cpu(cpu)
+            , _defer_local_close(defer_local_close)
             , _lb(make_foreign(std::move(lb))) {}
 
         handle(const handle&) = delete;
@@ -146,9 +149,18 @@ public:
                 return;
             }
             // FIXME: future is discarded
-            (void)smp::submit_to(_host_cpu, [cpu = _target_cpu, lb = std::move(_lb)] {
-                lb->closed_cpu(cpu);
-            });
+            if (_defer_local_close && _host_cpu == this_shard_id()) {
+                // smp::submit_to() executes same-shard calls inline. For connection
+                // distribution handles, defer local close reports too so local and remote
+                // close accounting have comparable, delayed visibility.
+                schedule(make_task([cpu = _target_cpu, lb = std::move(_lb)] () mutable {
+                    lb->closed_cpu(cpu);
+                }));
+            } else {
+                (void)smp::submit_to(_host_cpu, [cpu = _target_cpu, lb = std::move(_lb)] {
+                    lb->closed_cpu(cpu);
+                });
+            }
         }
         shard_id cpu() {
             return _target_cpu;
@@ -158,10 +170,10 @@ public:
 
     conntrack() : _lb(make_lw_shared<load_balancer>()) {}
     handle get_handle() {
-        return handle(_lb->next_cpu(), _lb);
+        return handle(_lb->next_cpu(), _lb, true);
     }
     handle get_handle(shard_id cpu) {
-        return handle(_lb->force_cpu(cpu), _lb);
+        return handle(_lb->force_cpu(cpu), _lb, false);
     }
 };
 
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -20,15 +20,16 @@
  */
 
 #pragma once
+#include <algorithm>
+#include <cmath>
 #include <unordered_set>
+#include <utility>
 #include <seastar/core/sharded.hh>
 #include <seastar/core/internal/pollable_fd.hh>
 #include <seastar/net/stack.hh>
 #include <seastar/core/polymorphic_temporary_buffer.hh>
 #include <seastar/core/internal/buffer_allocator.hh>
 #include <seastar/util/program-options.hh>
-
-#include <unordered_set>
 
 namespace seastar {
 
@@ -60,6 +61,16 @@ using namespace seastar;
 class conntrack {
     class load_balancer {
         std::vector<unsigned> _cpu_load;
+        // The least-loaded shard is cached, not to calculate it for every new connection.
+        // The value is the { shard_id, TTL } pair. The TTL is the number of future
+        // new connections that may reuse the cached shard before recalculating it.
+        // Destroyed connections also consume the TTL, since they change shard load.
+        std::pair<shard_id, unsigned> _least_loaded_cpu = {0, 0};
+        // The value used to reset the TTL depends on the number of currently open
+        // connections on the selected shard. The _least_loaded_cpu_ttl determines
+        // the fraction of that shard's open connections.
+        static constexpr float _least_loaded_cpu_ttl = 0.02; // 2%
+
         shard_id find_min_cpu() const {
             // Prefer shard 0 for the 1st connection
             if (_cpu_load[0] == 0) {
@@ -81,17 +92,31 @@ class conntrack {
         load_balancer() : _cpu_load(size_t(this_smp_shard_count()), 0) {}
         void closed_cpu(shard_id cpu) {
             _cpu_load[cpu]--;
+            if (_least_loaded_cpu.second > 0) {
+                _least_loaded_cpu.second--;
+            }
         }
         shard_id next_cpu() {
             // FIXME: The naive algorithm will just round robin the connections around the shards.
             // A more complex version can keep track of the amount of activity in each connection,
             // and use that information.
-            auto cpu = find_min_cpu();
-            ++_cpu_load[cpu];
-            return cpu;
+            // Use the cached value if still valid.
+            if (_least_loaded_cpu.second == 0) {
+                _least_loaded_cpu.first = find_min_cpu();
+                auto ttl = static_cast<unsigned>(std::round(
+                        _cpu_load[_least_loaded_cpu.first] * _least_loaded_cpu_ttl));
+                _least_loaded_cpu.second = std::min(ttl, 10u);
+            } else {
+                _least_loaded_cpu.second--;
+            }
+            ++_cpu_load[_least_loaded_cpu.first];
+            return _least_loaded_cpu.first;
         }
         shard_id force_cpu(shard_id cpu) {
             _cpu_load[cpu]++;
+            if (_least_loaded_cpu.second > 0) {
+                _least_loaded_cpu.second--;
+            }
             return cpu;
         }
     };
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -60,6 +60,23 @@ using namespace seastar;
 class conntrack {
     class load_balancer {
         std::vector<unsigned> _cpu_load;
+        shard_id find_min_cpu() const {
+            // Prefer shard 0 for the 1st connection
+            if (_cpu_load[0] == 0) {
+                return 0;
+            }
+            shard_id min_cpu = 0;
+            auto min_load = _cpu_load[0];
+            for (shard_id i = 1; i < _cpu_load.size(); ++i) {
+                auto load = _cpu_load[i];
+                // Choose the highest-numbered shard for ties to avoid shard 0
+                if (load <= min_load) {
+                    min_load = load;
+                    min_cpu = i;
+                }
+            }
+            return min_cpu;
+        }
     public:
         load_balancer() : _cpu_load(size_t(this_smp_shard_count()), 0) {}
         void closed_cpu(shard_id cpu) {
@@ -69,9 +86,8 @@ class conntrack {
             // FIXME: The naive algorithm will just round robin the connections around the shards.
             // A more complex version can keep track of the amount of activity in each connection,
             // and use that information.
-            auto min_el = std::min_element(_cpu_load.begin(), _cpu_load.end());
-            auto cpu = shard_id(std::distance(_cpu_load.begin(), min_el));
-            _cpu_load[cpu]++;
+            auto cpu = find_min_cpu();
+            ++_cpu_load[cpu];
             return cpu;
         }
         shard_id force_cpu(shard_id cpu) {
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -73,8 +73,17 @@ class conntrack {
         static constexpr float _least_loaded_cpu_ttl = 0.02; // 2%
 
         shard_id find_min_cpu() const {
-            auto min_el = std::min_element(_cpu_load.begin(), _cpu_load.end());
-            return shard_id(std::distance(_cpu_load.begin(), min_el));
+            shard_id min_cpu = 0;
+            auto min_load = _cpu_load[0];
+            for (shard_id i = 1; i < _cpu_load.size(); ++i) {
+                auto load = _cpu_load[i];
+                // Choose the highest-numbered shard for ties to avoid shard 0
+                if (load <= min_load) {
+                    min_load = load;
+                    min_cpu = i;
+                }
+            }
+            return min_cpu;
         }
     public:
         load_balancer() : _cpu_load(size_t(this_smp_shard_count()), 0) {}
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -20,7 +20,10 @@
  */
 
 #pragma once
+#include <algorithm>
+#include <cmath>
 #include <unordered_set>
+#include <utility>
 #include <seastar/core/sharded.hh>
 #include <seastar/core/make_task.hh>
 #include <seastar/core/internal/pollable_fd.hh>
@@ -28,8 +31,6 @@
 #include <seastar/core/polymorphic_temporary_buffer.hh>
 #include <seastar/core/internal/buffer_allocator.hh>
 #include <seastar/util/program-options.hh>
-
-#include <unordered_set>
 
 namespace seastar {
 
@@ -61,22 +62,48 @@ using namespace seastar;
 class conntrack {
     class load_balancer {
         std::vector<unsigned> _cpu_load;
+        // The least-loaded shard is cached, not to calculate it for every new connection.
+        // The value is the { shard_id, TTL } pair. The TTL is the number of future
+        // new connections that may reuse the cached shard before recalculating it.
+        // Destroyed connections also consume the TTL, since they change shard load.
+        std::pair<shard_id, unsigned> _least_loaded_cpu = {0, 0};
+        // The value used to reset the TTL depends on the number of currently open
+        // connections on the selected shard. The _least_loaded_cpu_ttl determines
+        // the fraction of that shard's open connections.
+        static constexpr float _least_loaded_cpu_ttl = 0.02; // 2%
+
+        shard_id find_min_cpu() const {
+            auto min_el = std::min_element(_cpu_load.begin(), _cpu_load.end());
+            return shard_id(std::distance(_cpu_load.begin(), min_el));
+        }
     public:
         load_balancer() : _cpu_load(size_t(this_smp_shard_count()), 0) {}
         void closed_cpu(shard_id cpu) {
             _cpu_load[cpu]--;
+            if (_least_loaded_cpu.second > 0) {
+                _least_loaded_cpu.second--;
+            }
         }
         shard_id next_cpu() {
             // FIXME: The naive algorithm will just round robin the connections around the shards.
             // A more complex version can keep track of the amount of activity in each connection,
             // and use that information.
-            auto min_el = std::min_element(_cpu_load.begin(), _cpu_load.end());
-            auto cpu = shard_id(std::distance(_cpu_load.begin(), min_el));
-            _cpu_load[cpu]++;
-            return cpu;
+            // Use the cached value if still valid.
+            if (_least_loaded_cpu.second == 0) {
+                _least_loaded_cpu.first = find_min_cpu();
+                auto ttl = static_cast<unsigned>(std::round(_cpu_load[_least_loaded_cpu.first] * _least_loaded_cpu_ttl));
+                _least_loaded_cpu.second = std::min(ttl, 10u);
+            } else {
+                _least_loaded_cpu.second--;
+            }
+            ++_cpu_load[_least_loaded_cpu.first];
+            return _least_loaded_cpu.first;
         }
         shard_id force_cpu(shard_id cpu) {
             _cpu_load[cpu]++;
+            if (_least_loaded_cpu.second > 0) {
+                _least_loaded_cpu.second--;
+            }
             return cpu;
         }
     };
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -22,6 +22,7 @@
 #pragma once
 #include <unordered_set>
 #include <seastar/core/sharded.hh>
+#include <seastar/core/make_task.hh>
 #include <seastar/core/internal/pollable_fd.hh>
 #include <seastar/net/stack.hh>
 #include <seastar/core/polymorphic_temporary_buffer.hh>
@@ -88,12 +89,14 @@ public:
     class handle {
         shard_id _host_cpu;
         shard_id _target_cpu;
+        bool _defer_local_close = false;
         foreign_ptr<lw_shared_ptr<load_balancer>> _lb;
     public:
         handle() : _lb(nullptr) {}
-        handle(shard_id cpu, lw_shared_ptr<load_balancer> lb)
+        handle(shard_id cpu, lw_shared_ptr<load_balancer> lb, bool defer_local_close)
             : _host_cpu(this_shard_id())
             , _target_cpu(cpu)
+            , _defer_local_close(defer_local_close)
             , _lb(make_foreign(std::move(lb))) {}
 
         handle(const handle&) = delete;
@@ -105,9 +108,18 @@ public:
                 return;
             }
             // FIXME: future is discarded
-            (void)smp::submit_to(_host_cpu, [cpu = _target_cpu, lb = std::move(_lb)] {
-                lb->closed_cpu(cpu);
-            });
+            if (_defer_local_close && _host_cpu == this_shard_id()) {
+                // smp::submit_to() executes same-shard calls inline. For connection
+                // distribution handles, defer local close reports too so local and remote
+                // close accounting have comparable, delayed visibility.
+                schedule(make_task([cpu = _target_cpu, lb = std::move(_lb)] () mutable {
+                    lb->closed_cpu(cpu);
+                }));
+            } else {
+                (void)smp::submit_to(_host_cpu, [cpu = _target_cpu, lb = std::move(_lb)] {
+                    lb->closed_cpu(cpu);
+                });
+            }
         }
         shard_id cpu() {
             return _target_cpu;
@@ -117,10 +129,10 @@ public:
 
     conntrack() : _lb(make_lw_shared<load_balancer>()) {}
     handle get_handle() {
-        return handle(_lb->next_cpu(), _lb);
+        return handle(_lb->next_cpu(), _lb, true);
     }
     handle get_handle(shard_id cpu) {
-        return handle(_lb->force_cpu(cpu), _lb);
+        return handle(_lb->force_cpu(cpu), _lb, false);
     }
 };
 
--- a/tests/unit/connect_test.cc
+++ b/tests/unit/connect_test.cc
@@ -28,7 +28,9 @@ SEASTAR_TEST_CASE(test_unconnected_socket_shutsdown_established_connection) {
     std::default_random_engine& rnd = testing::local_random_engine;
     auto distr = std::uniform_int_distribution<uint16_t>(12000, 65000);
     auto sa = make_ipv4_address({"127.0.0.1", distr(rnd)});
-    return do_with(engine().net().listen(sa, listen_options()), [sa] (auto& listener) {
+    listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
+    return do_with(engine().net().listen(sa, lo), [sa] (auto& listener) {
         auto f = listener.accept();
         auto unconn = make_socket();
         auto connf = unconn.connect(sa);
@@ -54,7 +56,9 @@ SEASTAR_TEST_CASE(test_accept_after_abort) {
     std::default_random_engine& rnd = testing::local_random_engine;
     auto distr = std::uniform_int_distribution<uint16_t>(12000, 65000);
     auto sa = make_ipv4_address({"127.0.0.1", distr(rnd)});
-    return do_with(seastar::server_socket(engine().net().listen(sa, listen_options())), [] (auto& listener) {
+    listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
+    return do_with(seastar::server_socket(engine().net().listen(sa, lo)), [] (auto& listener) {
         using ftype = future<accept_result>;
         promise<ftype> p;
         future<ftype> done = p.get_future();
--- a/tests/unit/epoll_test.cc
+++ b/tests/unit/epoll_test.cc
@@ -42,6 +42,7 @@ SEASTAR_THREAD_TEST_CASE(epoll_busy_spin_on_socket_error_test) {
     auto start_busy_time = engine().total_busy_time();
 
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     server_socket ss = seastar::listen(ipv4_addr(0), lo);
 
--- a/tests/unit/ipv6_test.cc
+++ b/tests/unit/ipv6_test.cc
@@ -75,7 +75,9 @@ SEASTAR_TEST_CASE(tcp_packet_test) {
     }
 
     return async([] {
-        auto sc = server_socket(engine().net().listen(ipv6_addr{"::1"}, {}));
+        listen_options lo;
+        lo.set_fixed_cpu(this_shard_id());
+        auto sc = server_socket(engine().net().listen(ipv6_addr{"::1"}, lo));
         auto la = sc.local_address();
 
         BOOST_REQUIRE(la.addr().is_ipv6());
--- a/tests/unit/proxy_protocol_test.cc
+++ b/tests/unit/proxy_protocol_test.cc
@@ -186,6 +186,7 @@ static void test_proxy_header_negative(
     std::vector<char> header) {
 
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     lo.proxy_protocol = true;
 
@@ -220,6 +221,7 @@ SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_bad_signature) {
 
 SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_small_packet_incomplete_header) {
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     lo.proxy_protocol = true;
 
@@ -251,6 +253,7 @@ SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_small_packet_incomplete_header) {
 
 SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_small_packet_incomplete_addresses) {
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     lo.proxy_protocol = true;
 
@@ -295,6 +298,7 @@ SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_small_packet_incomplete_addresses) {
 
 SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_local_mode) {
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     lo.proxy_protocol = true;
 
@@ -333,6 +337,7 @@ SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_local_mode_wrong_family) {
 
 SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_ipv4_addresses) {
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     lo.proxy_protocol = true;
 
@@ -363,6 +368,7 @@ SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_ipv4_addresses) {
 
 SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_ipv6_addresses) {
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     lo.proxy_protocol = true;
 
@@ -417,6 +423,7 @@ SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_unsupported_protocol) {
 
 SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_length_too_short_ipv4) {
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     lo.proxy_protocol = true;
 
@@ -458,6 +465,7 @@ SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_length_too_short_ipv4) {
 
 SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_length_too_short_ipv6) {
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     lo.proxy_protocol = true;
 
@@ -497,6 +505,7 @@ SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_length_too_short_ipv6) {
 SEASTAR_THREAD_TEST_CASE(proxy_protocol_v2_with_tlvs) {
     // TLVs should be skipped/ignored by the implementation
     listen_options lo;
+    lo.set_fixed_cpu(this_shard_id());
     lo.reuse_address = true;
     lo.proxy_protocol = true;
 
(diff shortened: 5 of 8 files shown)


include/seastar/net/posix-stack.hh
```diff
@@ -60,22 +63,60 @@ using namespace seastar;
...
-            return cpu;
+            // Use the cached value if still valid.
+            if (_least_loaded_cpu.second == 0) {
+                _least_loaded_cpu.first = find_min_cpu();
+                _least_loaded_cpu.second = std::clamp(static_cast<unsigned>(_cpu_load[_least_loaded_cpu.first] * _least_loaded_cpu_ttl), 1u, 10u);
+            } else {
+                _least_loaded_cpu.second--;
+            }
+            ++_cpu_load[_least_loaded_cpu.first];
+            return _least_loaded_cpu.first;
```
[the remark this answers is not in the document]
ScyllaPiotr: Changed the logic to avoid clamping to at least 1.

include/seastar/net/posix-stack.hh
```diff
@@ -60,22 +63,60 @@ using namespace seastar;
...
         std::vector<unsigned> _cpu_load;
+        // The least-loaded shard is cached, not to calculate it for every new connection.
+        // The value is the { shard_id, TTL } pair. The TTL is the number of connection count changes
+        // before the value is recalculated.
+        // Each new or destroyed connection will decrement the TTL.
+        std::pair<shard_id, unsigned> _least_loaded_cpu = {0, 0};
+        // The value which is used to reset the TTL above each time it reaches zero depends on
+        // the number of currently open connections.
+        // The _least_loaded_cpu_ttl determines the fraction of the currently open connections.
+        static constexpr float _least_loaded_cpu_ttl = 0.02; // 2%
```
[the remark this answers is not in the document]
ScyllaPiotr: Fixed the comment to refer to the number of open connections **per shard**.

avikivity:
Please enrich the cover letter with a description of how you tested this and what the test results were.

ScyllaPiotr:
I ran a local comparison of this PR against both the baseline and #3524.

Tested revisions:

| Variant | Revision |
|---|---|
| baseline | `4a23ff87` |
| #3524 | `769cf2cb` |
| #3526 | `144340af` |

All runs used 4 shards:

```bash
--smp 4 --cpuset 0-3
```

I used an extended version of Nadav's reproducer with:
- repeated HTTP waves, so each run reports per-wave deltas
- optional long-lived background connections per server shard, to reduce the empty-shard-0 startup/tie effect
- optional handler sleep, to keep short-lived requests live long enough for conntrack counters to matter
- a TCP-only stream mode for lower-level conntrack churn checks

## Main HTTP test

Command shape:

```bash
tests/perf/http_shard_nadav_balance_perf \
  --smp 4 --cpuset 0-3 \
  --mode http \
  --background-per-shard {0,4,40} \
  --repeats 5 \
  --requests-per-shard 100000 \
  --concurrency 10 \
  --handler-sleep {0,5}
```

Each HTTP row below is the median across 5 repeated waves. Counts are representative per-shard request counts for a 400k-request wave.

| Background hold connections per server shard | Handler sleep per measured request | Variant | Representative per-shard request counts `[shard0, shard1, shard2, shard3]` | Shard 0 load / ideal shard load | Max shard load / ideal shard load | Min shard load / ideal shard load | Coefficient of variation |
|---:|---:|---|---:|---:|---:|---:|---:|
| 0 | 0ms | baseline | `[144318, 66931, 67359, 121392]` | `1.443` | `1.443` | `0.669` | `0.317` |
| 0 | 0ms | #3524 | `[123270, 67466, 69475, 139789]` | `1.265` | `1.398` | `0.675` | `0.321` |
| 0 | 0ms | #3526 | `[133020, 69487, 69406, 128087]` | `1.330` | `1.330` | `0.699` | `0.305` |
| 4 | 0ms | baseline | `[149134, 65442, 64152, 121272]` | `1.491` | `1.491` | `0.649` | `0.361` |
| 4 | 0ms | #3524 | `[118707, 72162, 75372, 133759]` | `1.206` | `1.338` | `0.738` | `0.268` |
| 4 | 0ms | #3526 | `[133437, 69505, 68997, 128061]` | `1.334` | `1.334` | `0.690` | `0.308` |
| 40 | 0ms | baseline | `[152500, 61477, 61408, 124615]` | `1.525` | `1.525` | `0.632` | `0.395` |
| 40 | 0ms | #3524 | `[110499, 78897, 88734, 121870]` | `1.105` | `1.219` | `0.794` | `0.170` |
| 40 | 0ms | #3526 | `[138510, 77449, 64829, 119212]` | `1.385` | `1.385` | `0.665` | `0.284` |
| 0 | 5ms | baseline | `[102160, 99714, 99587, 98539]` | `1.022` | `1.022` | `0.986` | `0.013` |
| 0 | 5ms | #3524 | `[100358, 99557, 100262, 99823]` | `1.004` | `1.004` | `0.994` | `0.004` |
| 0 | 5ms | #3526 | `[100169, 99239, 100009, 100583]` | `1.002` | `1.006` | `0.993` | `0.005` |
| 4 | 5ms | baseline | `[101880, 99533, 99596, 98991]` | `1.019` | `1.019` | `0.990` | `0.011` |
| 4 | 5ms | #3524 | `[100381, 99444, 100223, 99952]` | `1.004` | `1.004` | `0.994` | `0.004` |
| 4 | 5ms | #3526 | `[100527, 98651, 100334, 100488]` | `1.004` | `1.005` | `0.989` | `0.007` |
| 40 | 5ms | baseline | `[102193, 99392, 99640, 98775]` | `1.022` | `1.022` | `0.987` | `0.013` |
| 40 | 5ms | #3524 | `[100604, 99107, 100187, 100102]` | `1.006` | `1.006` | `0.991` | `0.005` |
| 40 | 5ms | #3526 | `[107625, 108285, 91620, 92470]` | `1.076` | `1.083` | `0.916` | `0.080` |

## Shorter 10ms HTTP test

I also ran a shorter `handler-sleep=10ms` matrix:

```bash
--repeats 3
--requests-per-shard 10000
--concurrency 10
--handler-sleep 10
--background-per-shard {0,4,40}
```

Each wave has 40k measured HTTP requests.

| Background hold connections per server shard | Variant | Representative per-shard request counts `[shard0, shard1, shard2, shard3]` | Shard 0 load / ideal shard load | Max shard load / ideal shard load | Min shard load / ideal shard load | Coefficient of variation |
|---:|---|---:|---:|---:|---:|---:|
| 0 | baseline | `[10277, 9962, 9955, 9806]` | `1.028` | `1.028` | `0.982` | `0.017` |
| 0 | #3524 | `[10011, 9966, 10016, 10007]` | `1.001` | `1.002` | `0.997` | `0.002` |
| 0 | #3526 | `[9995, 9950, 10003, 10052]` | `1.000` | `1.005` | `0.995` | `0.004` |
| 4 | baseline | `[10147, 9973, 9984, 9896]` | `1.015` | `1.015` | `0.990` | `0.009` |
| 4 | #3524 | `[10019, 9963, 10028, 9990]` | `1.001` | `1.003` | `0.996` | `0.003` |
| 4 | #3526 | `[10008, 9920, 10018, 10054]` | `1.001` | `1.005` | `0.995` | `0.004` |
| 40 | baseline | `[10217, 9984, 9952, 9847]` | `1.022` | `1.022` | `0.985` | `0.014` |
| 40 | #3524 | `[10014, 9982, 10008, 9996]` | `1.001` | `1.001` | `0.998` | `0.002` |
| 40 | #3526 | `[10670, 10882, 9167, 9281]` | `1.071` | `1.088` | `0.914` | `0.080` |

## TCP-only short-connection stream

I also ran a TCP-only stream test with 10ms connection lifetime:

```bash
--mode tcp-stream
--duration 30
--connection-lifetime-ms 10
--tcp-rate {1000,2000,4000,8000}
```

| Target TCP connection rate | Variant | Representative per-shard accepted counts `[shard0, shard1, shard2, shard3]` | Shard 0 accepted / ideal shard accepted | Max shard accepted / ideal shard accepted | Min shard accepted / ideal shard accepted | Coefficient of variation |
|---:|---|---:|---:|---:|---:|---:|
| 1000/s | baseline | `[7510, 7501, 7498, 7491]` | `1.001` | `1.001` | `0.999` | `0.001` |
| 1000/s | #3524 | `[7482, 7498, 7505, 7515]` | `0.998` | `1.002` | `0.998` | `0.002` |
| 1000/s | #3526 | `[7476, 7493, 7508, 7523]` | `0.997` | `1.003` | `0.997` | `0.002` |
| 2000/s | baseline | `[16728, 15028, 14223, 14021]` | `1.115` | `1.115` | `0.935` | `0.071` |
| 2000/s | #3524 | `[14126, 14186, 15070, 16618]` | `0.942` | `1.092` | `0.942` | `0.059` |
| 2000/s | #3526 | `[14045, 14707, 15264, 15984]` | `0.936` | `1.069` | `0.936` | `0.049` |
| 4000/s | baseline | `[31406, 30147, 29805, 28642]` | `1.049` | `1.049` | `0.955` | `0.035` |
| 4000/s | #3524 | `[29188, 29190, 30080, 31542]` | `0.973` | `1.051` | `0.973` | `0.033` |
| 4000/s | #3526 | `[29124, 29132, 30082, 31662]` | `0.971` | `1.055` | `0.971` | `0.034` |
| 8000/s | baseline | `[61511, 59968, 59691, 58830]` | `1.028` | `1.028` | `0.981` | `0.017` |
| 8000/s | #3524 | `[59459, 59139, 60012, 61390]` | `0.988` | `1.019` | `0.986` | `0.014` |
| 8000/s | #3526 | `[58982, 59021, 60488, 61509]` | `0.983` | `1.023` | `0.983` | `0.016` |

## Summary

The no-sleep Nadav HTTP workload is still the hardest case. With no background load, #3526 improves max/ideal from baseline `1.443` to `1.330`, while #3524 reduces shard0 bias more but still has high spread in that exact no-background case.

When I add long-lived background connections to avoid the empty-shard-0 effect, #3524 is clearly stronger on the zero-sleep Nadav workload:

| Background hold connections per server shard | baseline coefficient of variation | #3524 coefficient of variation | #3526 coefficient of variation |
|---:|---:|---:|---:|
| 4 | `0.361` | `0.268` | `0.308` |
| 40 | `0.395` | `0.170` | `0.284` |

With handler sleep of 5ms or 10ms, baseline is already mostly balanced. #3524 remains consistently closest to ideal. #3526 is close for `background-per-shard=0` and `background-per-shard=4`, but shows a reproducible skew with `background-per-shard=40`:

| Handler sleep per measured request | Background hold connections per server shard | baseline coefficient of variation | #3524 coefficient of variation | #3526 coefficient of variation |
|---:|---:|---:|---:|---:|
| 5ms | 40 | `0.013` | `0.005` | `0.080` |
| 10ms | 40 | `0.014` | `0.002` | `0.080` |

The TCP-only stream test is less severe than Nadav HTTP. #3526 looks best at `2000/s`, and #3524/#3526 are close at higher rates.

My interpretation is:
- #3526 does improve some short-connection cases, especially the TCP-only stream and no-background Nadav max/ideal.
- It does not fully address the Nadav HTTP imbalance once the empty-shard-0 preference is removed with background load.
- #3524 gives the stronger and more consistent balancing result in the HTTP reproducer, especially with `background-per-shard=40`.
- The `background-per-shard=40` + `handler-sleep={5,10}` skew in #3526 looks suspicious and may be worth investigating before relying on this approach.

ScyllaPiotr:
Correction to my previous benchmark comment: the `background-per-shard=40` rows were not apples-to-apples.

The benchmark opened background `/hold` clients in batches and stopped when every shard had at least the requested count. For #3526 this produced `[47,47,49,49]` background clients, while baseline and #3524 had `[40,40,40,40]`. That uneven persistent background load explains the earlier #3526 `handler-sleep=5/10ms` skew.

I fixed the benchmark to:
- keep exactly the requested number of `/hold` clients per server shard
- actively verify every kept background client immediately before measurement by sending another `/hold` request on that same client
- start measurement only after the verified live background distribution is exactly `[40,40,40,40]`

Corrected `background-per-shard=40` HTTP results:

| Variant | Handler sleep per measured request | Verified live background `/hold` clients before measurement `[shard0, shard1, shard2, shard3]` | Representative median per-shard measured requests `[shard0, shard1, shard2, shard3]` | Shard 0 measured requests / ideal shard load | Max measured shard load / ideal shard load | Min measured shard load / ideal shard load | Coefficient of variation |
|---|---:|---:|---:|---:|---:|---:|---:|
| baseline | 0ms | `[40, 40, 40, 40]` | `[146789, 62315, 63404, 127492]` | `1.468` | `1.468` | `0.623` | `0.378` |
| #3524 | 0ms | `[40, 40, 40, 40]` | `[130161, 64019, 67405, 138415]` | `1.302` | `1.384` | `0.640` | `0.344` |
| #3526 | 0ms | `[40, 40, 40, 40]` | `[130034, 64213, 68820, 136933]` | `1.300` | `1.369` | `0.642` | `0.336` |
| baseline | 5ms | `[40, 40, 40, 40]` | `[102221, 99684, 99547, 98548]` | `1.022` | `1.022` | `0.985` | `0.014` |
| #3524 | 5ms | `[40, 40, 40, 40]` | `[100352, 99358, 100135, 100155]` | `1.004` | `1.004` | `0.994` | `0.004` |
| #3526 | 5ms | `[40, 40, 40, 40]` | `[98321, 98640, 101335, 101704]` | `0.983` | `1.017` | `0.983` | `0.015` |
| baseline | 10ms | `[40, 40, 40, 40]` | `[10232, 9976, 9924, 9868]` | `1.023` | `1.023` | `0.987` | `0.014` |
| #3524 | 10ms | `[40, 40, 40, 40]` | `[10020, 9960, 10018, 10002]` | `1.002` | `1.002` | `0.996` | `0.002` |
| #3526 | 10ms | `[40, 40, 40, 40]` | `[9810, 9896, 10104, 10190]` | `0.981` | `1.019` | `0.981` | `0.015` |

So please disregard my earlier claim that #3526 has a suspicious `background-per-shard=40` skew with `handler-sleep=5/10ms`. That was caused by the benchmark setup creating uneven extra background clients for #3526, not by #3526 itself.

Updated interpretation:
- the zero-sleep Nadav HTTP workload remains imbalanced for all variants
- with exact verified background clients, #3526 is slightly better than #3524 in the zero-sleep row in this rerun
- with 5ms/10ms handler sleep, #3524 remains closest to ideal, but #3526 no longer shows the earlier severe skew

ScyllaPiotr:
@nyh @avikivity please review, this definitely needs your eyes.

avikivity:
Please see the idea from https://github.com/scylladb/seastar/pull/3578#issuecomment-5134641935

The advantage is that it reduces shard 0 load by spreading initial processing across all shards.

include/seastar/net/posix-stack.hh
```diff
@@ -60,6 +60,23 @@ using namespace seastar;
 class conntrack {
     class load_balancer {
         std::vector<unsigned> _cpu_load;
+        shard_id find_min_cpu() const {
+            // Prefer shard 0 for the 1st connection
+            if (_cpu_load[0] == 0) {
+                return 0;
+            }
```
avikivity: I think it's an unnecessary distinction. A real system will have many connections so there's no need to special-case the first.

include/seastar/net/posix-stack.hh
```diff
@@ -60,6 +60,23 @@ using namespace seastar;
...
+            shard_id min_cpu = 0;
+            auto min_load = _cpu_load[0];
+            for (shard_id i = 1; i < _cpu_load.size(); ++i) {
+                auto load = _cpu_load[i];
+                // Choose the highest-numbered shard for ties to avoid shard 0
+                if (load <= min_load) {
+                    min_load = load;
+                    min_cpu = i;
+                }
+            }
```
avikivity, line 85: Are ties really expected? Typically you will have many connections.

include/seastar/net/posix-stack.hh
ScyllaPiotr: This was added in response to a test failing. I could fix the test instead...

include/seastar/net/posix-stack.hh
```diff
@@ -60,6 +61,16 @@ using namespace seastar;
...
         std::vector<unsigned> _cpu_load;
+        // The least-loaded shard is cached, not to calculate it for every new connection.
+        // The value is the { shard_id, TTL } pair. The TTL is the number of future
+        // new connections that may reuse the cached shard before recalculating it.
+        // Destroyed connections also consume the TTL, since they change shard load.
+        std::pair<shard_id, unsigned> _least_loaded_cpu = {0, 0};
+        // The value used to reset the TTL depends on the number of currently open
+        // connections on the selected shard. The _least_loaded_cpu_ttl determines
+        // the fraction of that shard's open connections.
+        static constexpr float _least_loaded_cpu_ttl = 0.02; // 2%
```
avikivity, line 73: I think there's a good chance that a auto-vectorized scan will be faster than these tricks, even on very high smp count.

include/seastar/net/posix-stack.hh
```diff
@@ -146,9 +149,18 @@ public:
...
+                // distribution handles, defer local close reports too so local and remote
+                // close accounting have comparable, delayed visibility.
+                schedule(make_task([cpu = _target_cpu, lb = std::move(_lb)] () mutable {
+                    lb->closed_cpu(cpu);
+                }));
+            } else {
+                (void)smp::submit_to(_host_cpu, [cpu = _target_cpu, lb = std::move(_lb)] {
+                    lb->closed_cpu(cpu);
+                });
+            }
```
avikivity, line 158: This is a little bit artificial, but sensible.

Is this patch the source of most of the improvement?

avikivity:
> the zero-sleep Nadav HTTP workload remains imbalanced for all variants

I think with SO_REUSEPORT accept running on all shards we'll get both balancing, and better throughput because shard 0 will not be singled out.

Perhaps make the tie-breaking different for every shard: compute min_element(begin() + this_shard_id, end()) and min_element(begin(), begin() + this_shard_id) and pick the best of those.

include/seastar/net/posix-stack.hh
ScyllaPiotr: I measured this. Microbenchmark of the `conntrack::load_balancer::next_cpu()` shard-selection variants:

**Variants measured**

**min_element** — the pre-PR baseline, a single-pass argmin scan:

```cpp
auto min_el = std::min_element(_cpu_load.begin(), _cpu_load.end());
```

This form cannot auto-vectorize: an argmin loop must track the winning index inside the loop, and neither GCC 15 nor clang 21 vectorizes that pattern at any optimization level.

**two_pass** — the auto-vectorizable rewrite. The scan is split into a value min-reduction followed by an index lookup, both plain streaming loops the vectorizer accepts:

```cpp
auto min_load = std::ranges::min(_cpu_load);
auto min_el = std::ranges::find(_cpu_load, min_load);
```

clang vectorizes the reduction already at `-O2` (confirmed in the emitted assembly, SSE2), GCC at `-O3` or `-O2 -fvect-cost-model=dynamic`; under Seastar's default GCC `-O2` release flags it runs scalar (GCC's `-O2` cost model declines loops with unknown trip counts).

**cached_pair** — the cached least-loaded shard from the earlier revision of this PR: a `std::pair<shard_id, unsigned> _least_loaded_cpu` member holding `{shard, TTL}`. After each full rescan the TTL is reset to `round(0.02 * open connections on the chosen shard)` capped at 10; both new and destroyed connections decrement the TTL, and the rescan happens when it reaches 0.

**Method**

N shards each hold C open connections (C = open connections per shard, kept stationary). The in-flight connections live in a FIFO ring of N*C entries; every cycle closes the oldest connection and opens a new one through `next_cpu()`. Reported numbers are ns per close+open cycle, 1M cycles after warmup, Intel Core Ultra 7 155H, GCC 15.3.1 and clang 21.1.8, `-O2` in both cases. Each value is the average of 5 independent runs of the whole benchmark; run-to-run spread is within ~±3% for the larger cells and 1–2 ns for the sub-10 ns cells.

**Scan variants — cost independent of C (ns per cycle, GCC / clang):**

| shards | min_element | two_pass |
|---:|---:|---:|
| 8 | 13.4 / 6.9 | 4.4 / 5.7 |
| 32 | 53 / 20 | 19 / 16 |
| 128 | 165 / 71 | 103 / 42 |
| 512 | 749 / 296 | 414 / 147 |

**cached_pair (ns per cycle, GCC / clang; columns: C = open connections per shard):**

| shards | C=10 | C=100 | C=1000 | C=10000 |
|---:|---:|---:|---:|---:|
| 8 | 7.6 / 9.2 | 5.0 / 5.4 | 2.7 / 2.7 | 2.7 / 2.6 |
| 32 | 21 / 23 | 10 / 13 | 5.3 / 5.4 | 4.7 / 5.4 |
| 128 | 68 / 86 | 34 / 45 | 12 / 16 | 12 / 16 |
| 512 | 273 / 355 | 136 / 178 | 46 / 59 | 46 / 60 |

For reading the cached_pair tables: the TTL arithmetic makes the cost a staircase in C. At C=10 the TTL rounds to 0, so every accept rescans; at C=100 (TTL=2, with closes also consuming TTL) every 2nd accept rescans; from C≥500 the cap of 10 gives a rescan every ~6th accept, so C=1000 and C=10000 sit on the same plateau.

ScyllaPiotr:
> Please see the idea from [#3578 (comment)](https://github.com/scylladb/seastar/pull/3578#issuecomment-5134641935)
> 
> The advantage is that it reduces shard 0 load by spreading initial processing across all shards.

It took me a while to fully understand what you proposed.

Today only one shard runs the real listener: it does every `accept()`, consults the single global conntrack balancer, and on average hands `~(N−1)/N` of the connections off cross-shard.

The dormant reuseport path is the opposite extreme - every shard listens, keeps whatever the kernel's 4-tuple hash delivers, and nothing corrects the hash's lumpiness (hence the "heavy load imbalance" `FIXME` that disabled it).

Your proposal combines the layers: re-enable `SO_REUSEPORT` so all N shards accept - the kernel hash spreads incoming connections across listeners roughly uniformly, for free (the necessary hash is calculated anyway) - and give each listener its own **private** conntrack balancer, so a listener can still move an accepted connection to another shard based on the counters of the connections it has distributed. Each listener's view is partial (`~1/N` of the total), so each balancer is individually biased; the bet is that N independent biases point in different directions and cancel in aggregate, unlike today's single (global, all-knowing) balancer whose bias always pushes the same way. The proposed per-shard tie-breaking (maybe added with additional tolerance, to further limit shard jumps) keeps those biases decorrelated at the ties, and as a bonus each listener defaults to keeping the connection local, which also eliminates most cross-shard `fd` handoffs. The wins: shard 0 stops being the accept bottleneck, and the systematic shard-0 skew from close-accounting lag turns into symmetric self-preference of every shard, which cancels out.

However, the design trades systemic skew with global accounting (one balancer, biased, but it sees every connection and the fix in this PR removes the bias at its source deterministically) with statistical balance and locally-driven corrections, where nothing in the system ever sees the aggregate.

The problem appears when the local corrections get a tolerance, and there will be tolerance of at least one - some shard is always momentarily one connection below, so accepts keep paying cross-shard handoffs to chase ±1 differences - a tolerance band (keep local unless the accepting shard exceeds the slice minimum by A) is what makes the design cheap. But hysteresis and partial accounting compose badly: a listener stops correcting once its slice is within `±tolerance`, so one shard can sit at up to `N × tolerance` excess connections in aggregate while every one of the N balancers reports its slice as balanced. Even the minimal `tolerance=1` gives N excess; at high shard counts with a realistic band that's hundreds of connections of skew that no component can detect. I agree this problem is improbable, but not impossible.

So my take: the accept-spreading benefit is real, but the pure per-listener form trades a known, monitorable, deterministically-fixable skew for a rare, invisible, uncorrectable one - especially for a long-running node.

avikivity:
> > Please see the idea from [#3578 (comment)](https://github.com/scylladb/seastar/pull/3578#issuecomment-5134641935)
> > The advantage is that it reduces shard 0 load by spreading initial processing across all shards.
> 
> It took me a while to fully understand what you proposed.
> 
> Today only one shard runs the real listener: it does every `accept()`, consults the single global conntrack balancer, and on average hands `~(N−1)/N` of the connections off cross-shard.
> 
> The dormant reuseport path is the opposite extreme - every shard listens, keeps whatever the kernel's 4-tuple hash delivers, and nothing corrects the hash's lumpiness (hence the "heavy load imbalance" `FIXME` that disabled it).
> 
> Your proposal combines the layers: re-enable `SO_REUSEPORT` so all N shards accept - the kernel hash spreads incoming connections across listeners roughly uniformly, for free (the necessary hash is calculated anyway) - and give each listener its own **private** conntrack balancer, so a listener can still move an accepted connection to another shard based on the counters of the connections it has distributed. Each listener's view is partial (`~1/N` of the total), so each balancer is individually biased; the bet is that N independent biases point in different directions and cancel in aggregate, unlike today's single (global, all-knowing) balancer whose bias always pushes the same way. The proposed per-shard tie-breaking (maybe added with additional tolerance, to further limit shard jumps) keeps those biases decorrelated at the ties, and as a bonus each listener defaults to keeping the connection local, which also eliminates most cross-shard `fd` handoffs. The wins: shard 0 stops being the accept bottleneck, and the systematic shard-0 skew from close-accounting lag turns into symmetric self-preference of every shard, which cancels out.
> 
> However, the design trades systemic skew with global accounting (one balancer, biased, but it sees every connection and the fix in this PR removes the bias at its source deterministically) with statistical balance and locally-driven corrections, where nothing in the system ever sees the aggregate.

Re-enabling SO_REUSEPORT removes load from shard 0. For short-lived connections on large nodes, that can be important. It's not just the balancing.

> 
> The problem appears when the local corrections get a tolerance, and there will be tolerance of at least one - some shard is always momentarily one connection below, so accepts keep paying cross-shard handoffs to chase ±1 differences - a tolerance band (keep local unless the accepting shard exceeds the slice minimum by A) is what makes the design cheap. But hysteresis and partial accounting compose badly: a listener stops correcting once its slice is within `±tolerance`, so one shard can sit at up to `N × tolerance` excess connections in aggregate while every one of the N balancers reports its slice as balanced. Even the minimal `tolerance=1` gives N excess; at high shard counts with a realistic band that's hundreds of connections of skew that no component can detect. I agree this problem is improbable, but not impossible.
> 
> So my take: the accept-spreading benefit is real, but the pure per-listener form trades a known, monitorable, deterministically-fixable skew for a rare, invisible, uncorrectable one - especially for a long-running node.

The SO_REUSEPORT approach is weak when there is a small number of long-lived connections. If each listener sees < nr_shards connections, there will definitely be imbalance. This is typical with long-lived connections in benchmark schenarios (where there are few clients).

So perhaps we need higher synchronization penalty until we have c * nr_shards connections per listent (small c), then no synchronization penalty with higher risk of imbalance (but less important because there are many connections).

ScyllaPiotr:
> Re-enabling SO_REUSEPORT removes load from shard 0. For short-lived connections on large nodes, that can be important. It's not just the balancing.

Short-lived connections are aberrations. I agree we could optimise for this case too, only if there was no cost for our primary scenario, which IIUC is long-lived connections, and long-living nodes.

> The SO_REUSEPORT approach is weak when there is a small number of long-lived connections. If each listener sees < nr_shards connections, there will definitely be imbalance. This is typical with long-lived connections in benchmark schenarios (where there are few clients).
> 
> So perhaps we need higher synchronization penalty until we have c * nr_shards connections per listent (small c), then no synchronization penalty with higher risk of imbalance (but less important because there are many connections).

Like with any performance improvements, I propose an iterative approach based on measurements. What you're suggesting is distributed change in many places at once. This will take a bit longer to implement and much longer to fine tune the thresholds. I don't suspect it's the right way now, when there already is an IMO sensible idea for a fix of the current mechanism.

I'd rather proceed with the idea in this PR with answering your other questions here. However, if you like this new idea more anyway, I'll propose a separate PR for it so we can discuss it further.

include/seastar/net/posix-stack.hh
ScyllaPiotr: No, the close-report deferral is not the source of the improvement. Three trees measured below:

1. **baseline** — the conntrack balancer untouched.
2. **deferral+cache** — baseline plus two changes:
   - *deferral*: same-shard close reports go through the task queue instead of inline `smp::submit_to()`; key line: `schedule(make_task([cpu = _target_cpu, lb = std::move(_lb)] () mutable { lb->closed_cpu(cpu); }));`
   - *cache*: the least-loaded shard is reused between full rescans; key line: `std::pair<shard_id, unsigned> _least_loaded_cpu = {0, 0}; // { shard, TTL }`
3. **deferral+cache+tie-break** — the July version, which additionally avoided shard 0 on load ties (*tie-break*); key line: `if (load <= min_load) { min_load = load; min_cpu = i; }` — the `<=` makes equal loads resolve to the highest-numbered shard instead of the first one.

Shard 0 load / ideal (sleep = handler sleep per request ≈ connection lifetime; bg = long-lived background connections held per shard):

| workload | baseline | deferral+cache | deferral+cache+tie-break |
|---|---:|---:|---:|
| HTTP sleep=0ms, bg=0 | 1.437 | 1.450 | 1.374 |
| HTTP sleep=0ms, bg=4 | 1.493 | 1.440 | 1.334 |
| HTTP sleep=0ms, bg=40 | 1.468 | 1.465 | 1.300 |
| HTTP sleep=5ms, bg=0 | 1.022 | 1.017 | 1.002 |
| HTTP sleep=5ms, bg=4 | 1.019 | 1.016 | 1.013 |
| HTTP sleep=5ms, bg=40 | 1.022 | 1.032 | 0.983 |
| HTTP sleep=10ms, bg=0 | 1.028 | 1.013 | 1.000 |
| HTTP sleep=10ms, bg=4 | 1.015 | 1.015 | 1.002 |
| HTTP sleep=10ms, bg=40 | 1.023 | 1.029 | 0.981 |
| TCP-stream 1000/s | 1.001 | 1.001 | 0.997 |
| TCP-stream 2000/s | 1.115 | 1.113 | 0.936 |
| TCP-stream 4000/s | 1.049 | 1.050 | 0.970 |
| TCP-stream 8000/s | 1.028 | 1.029 | 1.005 |

include/seastar/net/posix-stack.hh
ScyllaPiotr: The primary goal of the current balancer is to equalize the connection counts, so I'd say _yes_ to this question.
My proposal with TTL-cached value allows limited imbalance, which should make the equal counts rarer.

avikivity:
> > Re-enabling SO_REUSEPORT removes load from shard 0. For short-lived connections on large nodes, that can be important. It's not just the balancing.
> 
> Short-lived connections are aberrations. I agree we could optimise for this case too, only if there was no cost for our primary scenario, which IIUC is long-lived connections, and long-living nodes.

But isn't that what triggered the patch? Long lived connections don't suffer from the bias in shard 0 lifetime measurement.

> 
> > The SO_REUSEPORT approach is weak when there is a small number of long-lived connections. If each listener sees < nr_shards connections, there will definitely be imbalance. This is typical with long-lived connections in benchmark schenarios (where there are few clients).
> > So perhaps we need higher synchronization penalty until we have c * nr_shards connections per listent (small c), then no synchronization penalty with higher risk of imbalance (but less important because there are many connections).
> 
> Like with any performance improvements, I propose an iterative approach based on measurements. What you're suggesting is distributed change in many places at once. This will take a bit longer to implement and much longer to fine tune the thresholds. I don't suspect it's the right way now, when there already is an IMO sensible idea for a fix of the current mechanism.
> 
> I'd rather proceed with the idea in this PR with answering your other questions here. However, if you like this new idea more anyway, I'll propose a separate PR for it so we can discuss it further.

If you have concrete reasons, I'm happy to discuss.

SO_RESUSEPORT is a larger change, and riskier (since it mixes in kernel behavior, which can vary greatly across machine sizes and NICs, and kernel versions), but it has a larger payoff for high connection rates since it shards the workload.

include/seastar/net/posix-stack.hh
ScyllaPiotr: The fix to the tests was put as a 1st commit of this PR. Lmk if you want it as a separate PR instead.

ScyllaPiotr:
I have finished working on this PR for now. I believe it's ready for review - all remarks to this PR (this PR's direction) have been handled.
The latest changes:
- The PR begins with bugfix for the tests, which relied on the 1st connection being directed to shard 0
- The PR depends on #3630 or #3649 to land first, otherwise the CI breaks. Due to this, I have converted the PR to draft for now.

Now, regarding the direction of the fix.

> > > Re-enabling SO_REUSEPORT removes load from shard 0. For short-lived connections on large nodes, that can be important. It's not just the balancing.
> > 
> > Short-lived connections are aberrations. I agree we could optimise for this case too, only if there was no cost for our primary scenario, which IIUC is long-lived connections, and long-living nodes.
> 
> But isn't that what triggered the patch? Long lived connections don't suffer from the bias in shard 0 lifetime measurement.

That's true, short-lived connections triggered this, and both ideas target fixing this case. The difference is where the risk goes.

`SO_REUSEPORT`'s payoff (sharding the accept path) only matters at high new connection rates, i.e. the short-lived case. The risk that this approach introduces - imbalance from per-listener accounting - lands in the long-lived case (our **primary use case**).

This PR improves the same short-lived case while leaving the long-lived case's accounting untouched: one global balancer, deterministic.

> > > The SO_REUSEPORT approach is weak when there is a small number of long-lived connections. If each listener sees < nr_shards connections, there will definitely be imbalance. This is typical with long-lived connections in benchmark schenarios (where there are few clients).
> > > So perhaps we need higher synchronization penalty until we have c * nr_shards connections per listent (small c), then no synchronization penalty with higher risk of imbalance (but less important because there are many connections).
> > 
> > 
> > Like with any performance improvements, I propose an iterative approach based on measurements. What you're suggesting is distributed change in many places at once. This will take a bit longer to implement and much longer to fine tune the thresholds. I don't suspect it's the right way now, when there already is an IMO sensible idea for a fix of the current mechanism.
> > I'd rather proceed with the idea in this PR with answering your other questions here. However, if you like this new idea more anyway, I'll propose a separate PR for it so we can discuss it further.
> 
> If you have concrete reasons, I'm happy to discuss.
> 
> SO_RESUSEPORT is a larger change, and riskier (since it mixes in kernel behavior, which can vary greatly across machine sizes and NICs, and kernel versions), but it has a larger payoff for high connection rates since it shards the workload.

Do you not agree that `SO_REUSEPORT` introduces risk for our primary scenario, which is long-lived connections?
Do you not agree that long-lived conns is our primary scenario we should optimise for, or at least not worsen in the process?
If you still believe in the `SO_REUSEPORT` case and would like me to pursue it to empirically check how would it perform, I can certainly do that.

ScyllaPiotr:
I published the latest perf. results directly in the cover letter.

avikivity:
> I have finished working on this PR for now. I believe it's ready for review - all remarks to this PR (this PR's direction) have been handled. The latest changes:
> 
> * The PR begins with bugfix for the tests, which relied on the 1st connection being directed to shard 0
> * The PR depends on [prometheus: build label filter from all shards, not just the local one - if needed. #3630](https://github.com/scylladb/seastar/pull/3630) or [prometheus: make scrape output independent of the serving shard #3649](https://github.com/scylladb/seastar/pull/3649) to land first, otherwise the CI breaks. Due to this, I have converted the PR to draft for now.
> 
> Now, regarding the direction of the fix.
> 
> > > > Re-enabling SO_REUSEPORT removes load from shard 0. For short-lived connections on large nodes, that can be important. It's not just the balancing.
> > > 
> > > 
> > > Short-lived connections are aberrations. I agree we could optimise for this case too, only if there was no cost for our primary scenario, which IIUC is long-lived connections, and long-living nodes.
> > 
> > 
> > But isn't that what triggered the patch? Long lived connections don't suffer from the bias in shard 0 lifetime measurement.
> 
> That's true, short-lived connections triggered this, and both ideas target fixing this case. The difference is where the risk goes.
> 
> `SO_REUSEPORT`'s payoff (sharding the accept path) only matters at high new connection rates, i.e. the short-lived case. The risk that this approach introduces - imbalance from per-listener accounting - lands in the long-lived case (our **primary use case**).

Right.

But high connection rates are still important. Not because we want to encourage them, but because we want incorrect usage not to also cause high shard 0 load.

> 
> This PR improves the same short-lived case while leaving the long-lived case's accounting untouched: one global balancer, deterministic.
> 
> > > > The SO_REUSEPORT approach is weak when there is a small number of long-lived connections. If each listener sees < nr_shards connections, there will definitely be imbalance. This is typical with long-lived connections in benchmark schenarios (where there are few clients).
> > > > So perhaps we need higher synchronization penalty until we have c * nr_shards connections per listent (small c), then no synchronization penalty with higher risk of imbalance (but less important because there are many connections).
> > > 
> > > 
> > > Like with any performance improvements, I propose an iterative approach based on measurements. What you're suggesting is distributed change in many places at once. This will take a bit longer to implement and much longer to fine tune the thresholds. I don't suspect it's the right way now, when there already is an IMO sensible idea for a fix of the current mechanism.
> > > I'd rather proceed with the idea in this PR with answering your other questions here. However, if you like this new idea more anyway, I'll propose a separate PR for it so we can discuss it further.
> > 
> > 
> > If you have concrete reasons, I'm happy to discuss.
> > SO_RESUSEPORT is a larger change, and riskier (since it mixes in kernel behavior, which can vary greatly across machine sizes and NICs, and kernel versions), but it has a larger payoff for high connection rates since it shards the workload.
> 
> Do you not agree that `SO_REUSEPORT` introduces risk for our primary scenario, which is long-lived connections? Do you not agree that long-lived conns is our primary scenario we should optimise for, or at least not worsen in the process? If you still believe in the `SO_REUSEPORT` case and would like me to pursue it to empirically check how would it perform, I can certainly do that.


It does, but I believe we can mitigate the risk.

For example, make the accounting more accurate as the number of connections decrease.

Say, below some threshold, we collect information from all shards.

As the number of connections increase, reduce the rate of collecting information.

As a result, the first N connections/shard are quite expensive.

If they are followed by many other connections, then the cost of these first few connections is amortized over many other following connections.

If they are not followed by many other connections, then while the cost per connection is high, the absolute cost is low.

First iteration: prob_collect_stats_other_shards = min(10/avg_connections_per_shard, 1)

When we have <= 10 connections/shard, we improve our statistics with 100% probability.
When we have 100 connections/shard, there is 10% probability we will improve our statistics.
When we have 1000 connections/shard, there is 1% probability we will improve our statistics.

PR #3527, opened 2026-07-10
dns: fix build with c-ares 1.34.7 constified host callback

avikivity:
c-ares 1.34.7 marked [1]  callback parameters as const.
This changed ares_host_callback from

    void (*)(void *, int, int, struct hostent *)

to

    void (*)(void *, int, int, const struct hostent *)

The captureless lambda we pass to ares_gethostbyaddr() in
get_host_by_addr() took a non-const `::hostent *`, so it no longer
converted to the expected function-pointer type, breaking the build:

    error: invalid user-defined conversion from '...(void *, int, int, hostent *)'
    to 'ares_host_callback' {aka 'void (*)(void *, int, int, const hostent *)'}

The change was reverted again in 1.34.8, so keying on the version would
have to special-case exactly 1.34.7. Instead, use a template lambda that
matches both versions.

[1] https://github.com/c-ares/c-ares/pull/1060

### DIFF
--- a/src/net/dns.cc
+++ b/src/net/dns.cc
@@ -44,6 +44,7 @@
 #include <arpa/nameser.h>
 #include <chrono>
 #include <memory>
+#include <concepts>
 
 #include <ares.h>
 #include <boost/lexical_cast.hpp>
@@ -602,7 +603,10 @@ dns_resolver::impl::get_host_by_addr(inet_address addr) {
 
     dns_call call(*this);
 
-    ares_gethostbyaddr(_channel, p->addr.data(), p->addr.size(), int(p->addr.in_family()), [](void* arg, int status, int timeouts, ::hostent* host) {
+    // c-ares 1.34.7 constified the hostent argument to ares_host_callback (upstream
+    // PR #1060), but 1.34.8 reverted it. Rather than key on the version, use a template
+    // that matches both variants.
+    ares_gethostbyaddr(_channel, p->addr.data(), p->addr.size(), int(p->addr.in_family()), [](void* arg, int status, int timeouts, std::convertible_to<const ::hostent*> auto host) {
         // we do potentially allocating operations below, so wrap the pointer in a
         // unique here.
         std::unique_ptr<promise_wrap> p(reinterpret_cast<promise_wrap *>(arg));


avikivity:
v2: adjust so it works for 1.34.8 (which reverted the breaking change). Match the signature instead of the version number.

avikivity:
v3: improved detection to work on older c-ares too

src/net/dns.cc
```diff
@@ -602,7 +615,11 @@ dns_resolver::impl::get_host_by_addr(inet_address addr) {
 
     dns_call call(*this);
 
-    ares_gethostbyaddr(_channel, p->addr.data(), p->addr.size(), int(p->addr.in_family()), [](void* arg, int status, int timeouts, ::hostent* host) {
+    // See ares_host_cb_takes_const above: pick the hostent const-qualification
+    // that matches the installed c-ares so the lambda converts to
+    // ares_host_callback.
+    using ares_host_callback_hostent = std::conditional_t<ares_host_cb_takes_const<>, const ::hostent, ::hostent>;
+    ares_gethostbyaddr(_channel, p->addr.data(), p->addr.size(), int(p->addr.in_family()), [](void* arg, int status, int timeouts, ares_host_callback_hostent* host) {
```
elcallio: Why not just do `[](void* arg, int status, int timeouts, auto* host)`? Works for my compiler and seems simplier...

elcallio (changes requested):

src/net/dns.cc
avikivity: It would work, though I'd add a constraint to make it clear.

The whole thing is moot since the change was backed out, but since I started I'll continue.

elcallio: If you are worried, just add a `static_assert(std::is_convertible_v<decltype(host), const ::hostent*>);`. But function call a few lines below will more or less to this for you also...

avikivity:
v4:
 - use simpler template lambda that matches to the right type instead of type computations
 - pointed reference to c-ares pull request in an unambiguous way

src/net/dns.cc
avikivity: I used std::convertible_to, it fits well.

elcallio (approved):
Nice. Concepts for the win.

PR #3530, opened 2026-07-13
tests/unit: replace hardcoded TLS certificates with build-system certs

nyh:
Several httpd unit tests (test_request_scheduling_group, test_mtls_dn_propagation, test_mtls_san_propagation) embedded self-signed PEM certificates as string literals directly in the source. This is fragile: the certs expire, reviewers cannot easily inspect what they certify, and they diverge from the cert infrastructure already maintained by the build system for tls_test.cc.

Replace all hardcoded certs with file-based loading via a certfile() helper (same pattern used in tls_test.cc).  The scheduling group test reuses the existing test.crt/catest.pem pair.  The two mTLS tests reuse the existing mtls_ca/mtls_server/mtls_client1 certs.  For the SAN propagation test, which requires a client cert carrying both a DNS and a URI (SPIFFE) SAN, a new mtls_client_san cert is added to the build system.

To support per-cert SAN injection when signing with a CA, seastar_sign_cert gains an optional SANS list argument that writes an OpenSSL extension file at configure time and passes it to openssl x509 at build time.

The httpd test target is updated to depend on testcrt and mtls_certs and to link Boost::filesystem (needed by boost::dll::program_location).

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -417,7 +417,10 @@ seastar_add_test (httpd
   SOURCES
     httpd_test.cc
     loopback_socket.hh
-    memory-data-sink.hh)
+    memory-data-sink.hh
+  DEPENDS testcrt mtls_certs
+  LIBRARIES Boost::filesystem
+  WORKING_DIRECTORY ${Seastar_BINARY_DIR})
 
 seastar_add_test (websocket
   SOURCES websocket_test.cc)
@@ -690,7 +693,7 @@ function(seastar_sign_cert name)
   cmake_parse_arguments(CERT
     ""
     "CA;SERIAL_NUM;COMMON_NAME"
-    ""
+    "SANS"
     ${ARGN})
   set(cert ${name}.crt)
   set(privkey ${name}.key)
@@ -718,6 +721,16 @@ function(seastar_sign_cert name)
     # sign the cert with the request
     set(ca_cert ${CERT_CA}.crt)
     set(ca_privkey ${CERT_CA}.key)
+    if(CERT_SANS)
+      string(JOIN ", " san_str ${CERT_SANS})
+      file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/${name}.ext"
+        "[req_ext]\nsubjectAltName = ${san_str}\n")
+      set(extfile_args -extfile ${name}.ext -extensions req_ext)
+      set(extfile_dep "${CMAKE_CURRENT_BINARY_DIR}/${name}.ext")
+    else()
+      set(extfile_args "")
+      set(extfile_dep "")
+    endif()
     add_custom_command(OUTPUT ${cert}
       COMMAND ${OPENSSL} x509
         -req -days 1000 -sha256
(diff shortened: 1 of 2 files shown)


nyh:
Due to popular demand (@elcallio and @knowack1) during the review of https://github.com/scylladb/seastar/pull/3517, I am submitting this followup cleanup of the HTTPS tests.

tests/unit/CMakeLists.txt
```diff
@@ -718,16 +721,32 @@ function(seastar_sign_cert name)
...
-        -in ${req}
-        -CA ${ca_cert}
-        -CAkey ${ca_privkey}
-        -out ${cert}
-      DEPENDS ${req} ${ca_cert} ${ca_privkey}
-      WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+    if(CERT_SANS)
+      string(JOIN ", " san_str ${CERT_SANS})
+      file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/${name}.ext" "subjectAltName = ${san_str}\n")
+      add_custom_command(OUTPUT ${cert}
```
elcallio: Nit: could just write the file and add the `-extfile ...` to an otherwise empty string var, always expanded in the command. To reduce repetition.

elcallio (approved):

tests/unit/CMakeLists.txt
```diff
@@ -718,16 +721,32 @@ function(seastar_sign_cert name)
...
+        COMMAND ${OPENSSL} x509
+          -req -days 1000 -sha256
+          -set_serial ${CERT_SERIAL_NUM}
+          -in ${req}
+          -CA ${ca_cert}
+          -CAkey ${ca_privkey}
+          -out ${cert}
+          -extfile ${name}.ext
+        DEPENDS ${req} ${ca_cert} ${ca_privkey}
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
```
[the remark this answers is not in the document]
nyh: Fixed.

tests/unit/CMakeLists.txt
nyh: Will do.

nyh:
Pushed a new version addressing reviewer comments

elcallio (approved):

tests/unit/CMakeLists.txt
```diff
@@ -718,6 +721,16 @@ function(seastar_sign_cert name)
     # sign the cert with the request
     set(ca_cert ${CERT_CA}.crt)
     set(ca_privkey ${CERT_CA}.key)
+    if(CERT_SANS)
+      string(JOIN ", " san_str ${CERT_SANS})
+      file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/${name}.ext"
+        "[req_ext]\nsubjectAltName = ${san_str}\n")
+      set(extfile_args -extfile ${name}.ext -extensions req_ext)
+      set(extfile_dep "${CMAKE_CURRENT_BINARY_DIR}/${name}.ext")
+    else()
```
avikivity, line 730: Who made else a function

nyh: I was shocked when I saw this. It's even worse, in the past they even expected you to write:
```
if(condition)
   ..
else(condition)           // same condition again, repeated, not the negation. How unclear!
  ...
endif(condition)
```
Apparently you no longer need to write the condition again, but you still need the parentheses?

In any case, I let the AI write the cmake for me. That language, or the whole software, never made any sense to me so I'm glad some atificial intelligence can make sense of it.

PR #3531, opened 2026-07-14
tls_gnutls: Enable verify_wrapper for both client and server mode

### LINKED ISSUE #3522: tls_gnutls.cc has a strange if() that's probably outdated?
src/net/tls_gnutls.cc has the following code:
```c++
         // This would be nice, because we preferably want verification to
         // abort hand shake so peer immediately knows we bailed...
 #if GNUTLS_VERSION_NUMBER >= 0x030406
        if (_type == type::CLIENT) {
            gnutls_session_set_verify_function(*this, &verify_wrapper);
        }
 #endif
```

The gnutls_session_set_verify_function() was introduced unconditionally in commit 61c593398 by @elcallio (December 2015), but then in commit 6fbd792d8 a year later (December 2016) the strange **`if (_type == type::CLIENT) {`** was added.

The commit message explained:
```
    Our "verify" function is meant to provide more context in
    verifying a server certificate. It does not work for a server
    session (verifying client certs).
    Caused test failure on non-obsolete systems.
```

However, it seems the **reason** why this was done for client-side checks (i.e., server certificate checks) should also apply for server-side checks (i.e., client certificates). The reason why this code "caused test failures" was probably bugs that existed in the old code that have been already fixed years ago. An AI I consulted claims that the old code passed a wrong hostname, and didn't handle GNUTLS_E_NO_CERTIFICATE_FOUND correctly, and both these bugs were fixed since.

It's trivial to remove the `if (_type == type::CLIENT) {`, but something that bothers me is that removing it does not solve a bug I've been trying to solve (failure to produce an alert when a client certificate is **expired** and instead the connection is closed abruptly), so I'm not sure that "fixing" this will actually solve anything...

elcallio:
Fixes #3522

Previously, we only set a verification wrapper callback for client
mode sockets. This because for server mode, the callback is never
called, unless certificate authentication is enabled. Nowdays,
we have support for this, so can make the design unilateral.
This also means we can drop explicit `verify()` calls in handshake,
and do one less extra rethrow in the error handling.

The test case test_non_tls was somewhat misnamed as it really
tested that a non-tls related reset will propagate through the
tls socket.

Renamed + added explicit exception check (and error code).
Added another test named test_non_tls that does exactly this, 
since this PR is about verifying the TLS connection.

### DIFF
--- a/src/net/tls_gnutls.cc
+++ b/src/net/tls_gnutls.cc
@@ -530,10 +530,12 @@ public:
         // This would be nice, because we preferably want verification to
         // abort hand shake so peer immediately knows we bailed...
 #if GNUTLS_VERSION_NUMBER >= 0x030406
-        if (_type == type::CLIENT) {
-            gnutls_session_set_verify_function(*this, &verify_wrapper);
-        }
-#endif
+        // #3522 - we can set our verify for both client and server.
+        // It will never be called in the server case, unless the above
+        // call to gnutls_certificate_server_set_request was request or
+        // require. Which is what we want.
+        gnutls_session_set_verify_function(*this, &verify_wrapper);
+ #endif
         // if we are a client, check if we have a session ticket to unpack.
         if (_type == type::CLIENT && !_options.session_resume_data.empty()) {
             gtls_chk(gnutls_session_set_data(*this, _options.session_resume_data.data(), _options.session_resume_data.size()));
@@ -624,7 +626,10 @@ public:
                     return make_exception_future<>(verification_error("No certificate was found"));
 #if GNUTLS_VERSION_NUMBER >= 0x030406
                 case GNUTLS_E_CERTIFICATE_ERROR:
-                    verify(); // should throw. otherwise, fallthrough
+                    // callback should have thrown. otherwise, fallthrough
+                    if (_error) {
+                        return clear_output_and_return_exception(_error);
+                    }
                     [[fallthrough]];
 #endif
                 default:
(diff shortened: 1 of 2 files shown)


avikivity:
What's this pre-commit failure?

elcallio:
> What's this pre-commit failure?

It says "trim trailing whitespace.................................................Failed". 
Does this mean "please remove white space" or does it mean it tried to do it for me, and failed somehow?

elcallio:
Btw, if we are gonna run automatic formatting on patches, I humbly suggest they should be limited to the changed lines. Not apply unrelated changes elsewhere...

elcallio:
I repeat my question: It looks like "pre-commit" fails on pre-existing white space. Should I remove it (unrelated)?

xemul:
I think it failed on newly added whitespaces:

image

The line is

```
        // #3522 - we can set our verify for both client and server. 
```

it has one. There are more, that the CI job reports:

```patch
--- a/src/net/tls_gnutls.cc
+++ b/src/net/tls_gnutls.cc
@@ -530,9 +530,9 @@ class session : public enable_shared_from_this<session>, public tls::session_imp
         // This would be nice, because we preferably want verification to
         // abort hand shake so peer immediately knows we bailed...
 #if GNUTLS_VERSION_NUMBER >= 0x030406
-        // #3522 - we can set our verify for both client and server. 
+        // #3522 - we can set our verify for both client and server.
         // It will never be called in the server case, unless the above
-        // call to gnutls_certificate_server_set_request was request or 
+        // call to gnutls_certificate_server_set_request was request or
         // require. Which is what we want.
         gnutls_session_set_verify_function(*this, &verify_wrapper);
  #endif
@@ -725,7 +725,7 @@ class session : public enable_shared_from_this<session>, public tls::session_imp
            return make_exception_future(ep);
         });
     }
-    // See #3498. Throw helper that ensures we don't leak abandoned exception 
+    // See #3498. Throw helper that ensures we don't leak abandoned exception
     // futures in _output_pending
     future<> clear_output_and_return_exception(std::exception_ptr ep) {
         return std::exchange(_output_pending, make_ready_future()).then_wrapped([this, ep](future<> f) {
```

elcallio:
Fixed.

avikivity:
> Fixed.

Still there

PR #3533, opened 2026-07-14
Fix metrics queue label

### LINKED ISSUE #2075: wrong metrics names in net/dpdk.cc
net/dpdk.cc holds a bunch of queue metrics.
The queue name should be a label, not part of the name.

ankitxvx:
# net: move queue name from metric name to label

Queue-level network metrics currently embed the queue name in the metric
name itself (e.g. `queue0_rx_packets`, `queue1_tx_bytes`). This makes it
impossible to aggregate across queues using standard Prometheus queries
and generates an unbounded set of metric names that grows with the queue
count.

Move the queue name out of the metric name and into a `queue` label, so
metrics are now named like:

```
rx_packets{queue="queue0"}
tx_bytes{queue="queue1"}
```

instead of:

```
queue0_rx_packets
queue1_tx_bytes
```

This is the idiomatic Prometheus pattern for per-instance metrics and
enables straightforward `sum by (queue)` / `sum without (queue)` queries.

### Changes

**`src/net/net.cc`** — Base `qp` class (15 metrics):

- Declare `auto queue_l = sm::label("queue")(_queue_name)` once at the
  top of the constructor.
- Remove `_queue_name + "_"` prefix from all 11 main-group metric names
  and all 4 copy-stats metric names.
- Attach `(queue_l)` to each metric registration.
- Replace `seastar::format(...)` description strings that interpolated
  `_queue_name + "_<metric>"` with plain strings referencing the bare
  metric name (e.g. `tx_packets` instead of `queue0_tx_packets`).
- Update descriptions to use "per queue" phrasing for consistency.

**`src/net/dpdk.cc`** — `dpdk_qp` subclass (3 metrics):

- Same transformation for `rx_csum_errors`, `rx_errors`, and
  `rx_no_memory_errors`.

No header changes — `_queue_name` is still used to produce the label
value.

> [!WARNING]
> **Breaking change**: any existing dashboards or alerting rules that
> reference metric names like `queue0_rx_packets` will need to be updated
> to use `rx_packets{queue="queue0"}`.

Fixes #2075

### DIFF
--- a/src/net/dpdk.cc
+++ b/src/net/dpdk.cc
@@ -348,7 +348,6 @@ class dpdk_device : public device {
     rss_key_type _rss_key;
     port_stats _stats;
     timer<> _stats_collector;
-    const std::string _stats_plugin_name;
     const std::string _stats_plugin_inst;
     seastar::metrics::metric_groups _metrics;
     bool _is_i40e_device = false;
@@ -407,7 +406,6 @@ public:
         , _home_cpu(this_shard_id())
         , _use_lro(use_lro)
         , _enable_fc(enable_fc)
-        , _stats_plugin_name("network")
         , _stats_plugin_inst(std::string("port") + std::to_string(_port_idx))
         , _xstats(port_idx)
     {
@@ -420,7 +418,7 @@ public:
 
         // Register port statistics pollers
         namespace sm = seastar::metrics;
-        _metrics.add_group(_stats_plugin_name, {
+        _metrics.add_group("network", {
             // Rx Good
             sm::make_counter("rx_multicast", _stats.rx.good.mcast,
                             sm::description("Counts a number of received multicast packets."), {sm::shard_label(_stats_plugin_inst)}),
@@ -1244,8 +1242,7 @@ build_mbuf_cluster:
     };
 
 public:
-    explicit dpdk_qp(dpdk_device* dev, uint16_t qid,
-                     const std::string stats_plugin_name);
+    explicit dpdk_qp(dpdk_device* dev, uint16_t qid, std::string port_name);
 
     virtual void rx_start() override;
     virtual future<> send(packet p) override {
@@ -1911,9 +1908,8 @@ void dpdk_device::check_port_link_status()
 #pragma GCC diagnostic ignored "-Winvalid-offsetof"
 
 template <bool HugetlbfsMemBackend>
-dpdk_qp<HugetlbfsMemBackend>::dpdk_qp(dpdk_device* dev, uint16_t qid,
-                                      const std::string stats_plugin_name)
-     : qp(true, stats_plugin_name, qid), _dev(dev), _qid(qid),
+dpdk_qp<HugetlbfsMemBackend>::dpdk_qp(dpdk_device* dev, uint16_t qid, std::string port_name)
+     : qp(true, port_name, qid), _dev(dev), _qid(qid),
        _rx_gc_poller(reactor::poller::simple([&] { return rx_gc(); })),
        _tx_buf_factory(qid),
        _tx_gc_poller(reactor::poller::simple([&] { return _tx_buf_factory.gc(); }))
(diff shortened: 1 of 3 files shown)


amnonh:
@ankitxvx Please split the alignment patch to a different PR, it does not relate to metric names.

amnonh:
@ankitxvx if you are already changing metric names, let's follow the Prometheus naming convention,  https://prometheus.io/docs/practices/naming/

Specifically, the counters should end with _total.
Second, about the _stats_plugin_name, can you see if there's a reason it's a variable?
If not, let's use a pure string; it's better when people are looking for metrics in the code.

Last, it's a breaking change and should be addressed as such.

ankitxvx:
Thanks for the review @amnonh! I've updated the PR and pushed the changes to address all of your feedback:

Split the patch: I've dropped the posix_memalign alignment commit from this branch. I will open a separate PR for that fix.
Prometheus naming conventions: All queue-level metric counters touched by this PR have been renamed to end with _total (e.g., rx_packets_total, tx_bytes_total).
_stats_plugin_name variable: I removed the _stats_plugin_name variable from the qp base class constructors and replaced it with the "network" string literal directly to make it easily searchable.
Breaking change: I've explicitly noted in the commit message that this is a breaking change for existing dashboards/alerts.
Let me know if there's anything else needed!

src/net/net.cc
```diff
@@ -104,88 +104,87 @@ bool qp::poll_tx() {
...
+        sm::make_gauge("tx_packet_queue", [this] { return _tx_packetq.size(); },
+                        sm::description("Holds a number of packets pending to be sent per queue. "
+                                        "This metric will have high values if the network backend doesn't keep up with the upper layers or if upper layers send big bursts of packets."))(queue_l),
 
         //
         // Linearization counter: DERIVE:0:U
         //
-        sm::make_counter(_queue_name + "_xmit_linearized", _stats.tx.linearized,
-                        sm::description("Counts a number of linearized Tx packets. High value indicates that we send too fragmented packets.")),
+        sm::make_counter("xmit_linearized_total", _stats.tx.linearized,
```
amnonh: I would add packets to the name to indicate the type

src/net/net.cc
```diff
@@ -104,88 +104,87 @@ bool qp::poll_tx() {
...
+        sm::make_counter("tx_bytes_total", _stats.tx.good.bytes,
+                        sm::description("Transmit throughput per queue."))(queue_l),
         //
         // Queue length: GAUGE:0:U
         //
         // Tx
-        sm::make_gauge(_queue_name + "_tx_packet_queue", [this] { return _tx_packetq.size(); },
-                        sm::description("Holds a number of packets pending to be sent. "
-                                        "This metric will have high values if the network backend doesn't keep up with the upper layers or if upper layers send big bursts of packets.")),
+        sm::make_gauge("tx_packet_queue", [this] { return _tx_packetq.size(); },
```
amnonh: I would use the plural packets (also for the other names)

amnonh (changes requested):

amnonh:
LGTM

ankitxvx:
Awesome, thank you so much for the review and the guidance

src/net/dpdk.cc
```diff
@@ -2238,11 +2236,9 @@ std::unique_ptr<qp> dpdk_device::init_local_queue(const program_options::option_
 
     std::unique_ptr<qp> qp;
     if (net_opts->_hugepages) {
-        qp = std::make_unique<dpdk_qp<true>>(this, qid,
-                                 _stats_plugin_name + "-" + _stats_plugin_inst);
```
xemul, line 2242: Before the change the port participated in the _stats_plugin_inst and, thus, in the metrics name. After the change the port is gone from metrics naming, so double metrics registration may happen when app brings up two or more DPDK ports

ankitxvx: Thanks for catching this @xemul! You're completely right—by removing _stats_plugin_inst from the metric group name, we lost the port identity which would cause a double registration collision on multi-port setups.

I've just pushed a new commit that reintroduces the port identity by appending a port label to all the queue-level metrics (e.g., rx_packets_total{port="port0", queue="queue0"}). This keeps the metrics cleanly grouped under the "network" plugin for Prometheus while ensuring they remain uniquely identifiable across different ports.

Let me know if this approach looks good to you!

ankitxvx:
Hi @xemul @amnonh,

I've force-pushed a fix to correct the metric label chaining syntax (`(queue_l)(port_l)` instead of `(queue_l, port_l)`). 

All CI builds and unit tests are now green and passing cleanly. Please take a look when you have a moment!

src/net/net.cc
```diff
@@ -104,88 +104,88 @@ bool qp::poll_tx() {
...
-qp::qp(bool register_copy_stats,
-       const std::string stats_plugin_name, uint8_t qid)
+qp::qp(bool register_copy_stats, std::string port_name, uint8_t qid)
         : _tx_poller(std::make_unique<internal::poller>(reactor::poller::simple([this] { return poll_tx(); })))
-        , _stats_plugin_name(stats_plugin_name)
         , _queue_name(std::string("queue") + std::to_string(qid))
 {
     namespace sm = metrics;
+    auto queue_l = sm::label("queue")(_queue_name);
+    auto port_l = sm::label("port")(port_name);
```
xemul: The virtio::qp inherits from net::qp, and when constructed implicitly call the net::qp with default-empty post_name, so here an empty label would appear. Virtio device is singleton, so no double-reg here, but it's anyway not nice to have value-less label

ankitxvx: @xemul! Fixed by conditionally adding the port label only when port_name is non-empty. Now:

Virtio (singleton): net::qp is implicitly default-constructed with port_name="" → only queue="queue0" label, no value-less port="" label.
DPDK (multi-port): _stats_plugin_inst (e.g. "port0") is passed through → both queue="queue0" and port="port0" labels are present.
Also switched from operator() label chaining to passing the labels vector directly as a constructor argument to make_counter/make_gauge, which is cleaner when the number of labels is variable.

xemul (approved):
Looks good to me. @amnonh , please, re-review

xemul:
@amnonh , a re-review ping

amnonh (approved):
It looks good, and I really like the changes.

There are a couple of places where you pass a string as object rather than const reference (I've seen the original code pass const string for whatever reason) I'm not sure if it's on purpose. In anyway, metrics registration is such a unique event that it has no effect on the system performance anyhow so even if it's not on purpose, you don't have to change it

nyh:
> Last, it's a breaking change and should be addressed as such.

How do we normally "address" changes to the metrics? Do we just do these changes and assume the users will adapt to it, or do we have a more sophisticated process for these changes?

amnonh:
> > Last, it's a breaking change and should be addressed as such.
> 
> How do we normally "address" changes to the metrics? Do we just do these changes and assume the users will adapt to it, or do we have a more sophisticated process for these changes?

We have a sad process in which, on every release, I collect metrics changes, and it's part of the release notes

nyh:
> We have a sad process in which, on every release, I collect metrics changes, and it's part of the release notes

I never noticed we even have release notes for Seastar...

amnonh:
@nyh sorry, you're right, I was talking about the ScyllaDB release that includes Seastar metrics changes

PR #3534, opened 2026-07-15
coroutine: introduce try_expected_future

avikivity:
Special awaiter which co_await:s a future<std::expected<V, E>> and returns the wrapped value if the future resolved with one, terminating the coroutine otherwise, propagating the error directly to its waiter.

This is the std::expected counterpart of coroutine::try_future, distinguishing three outcomes of the awaited future:

 - if the future failed (holds an exception), the exception is forwarded to the waiter directly and the coroutine is destroyed, exactly like coroutine::try_future;
 - if the future resolved with an unexpected, it is rebound to the coroutine's return type (a future<std::expected<V2, E>> with a matching error type) and returned to the waiter, without resuming the coroutine;
 - if the future resolved with a value, that value is the result of the co_await.

In all cases the error is propagated up the std::expected chain without being thrown, analogous to rust's try operator.

Like coroutine::try_future, a preemption-checking and a non-preemption-checking variant are provided. Since a ready future offers no non-destructive way to inspect its expected, await_ready() unwraps it and puts it right back via make_ready_future(), so a ready value can still take the fast path without suspending.

The new awaiter and both variants are exported from the seastar module, tested in coroutines_test, and documented in a new section of the coroutine exception handling chapter of doc/tutorial.md.

### DIFF
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -583,6 +583,47 @@ seastar::future<int> exception_propagating() {
 }
 ```
 
+### Propagating `std::expected` errors
+
+When errors are represented as values rather than exceptions, a function returns a
+`future<std::expected<T, E>>`: the future carries either a value, or an error `E`, or (should
+something go wrong at the future level) an exception. `coroutine::try_expected_future` is the
+`std::expected` counterpart of `coroutine::try_future`. It inspects the three possible outcomes
+of the awaited future and:
+
+- if the future resolved with a value, that value becomes the result of the `co_await`;
+- if the future resolved with an unexpected, the error is rebound to the coroutine's return type
+  (which must be a `future<std::expected<U, E>>` with a matching error type `E`) and returned to
+  the waiter directly, without resuming the coroutine;
+- if the future failed with an exception, the exception is forwarded to the waiter directly, just
+  like `coroutine::try_future`.
+
+In all cases the error is propagated without throwing, and the code following the `co_await` only
+runs on the success path. This is analogous to rust's
+[try operator](https://google.github.io/comprehensive-rust/error-handling/try.html) applied to
+`Result`.
+
+Example:
+
+```cpp
+enum class error { not_found, corrupted };
+
+seastar::future<std::expected<int, error>> read_value();
+
+seastar::future<std::expected<seastar::sstring, error>> read_value_as_string() {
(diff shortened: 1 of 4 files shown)


avikivity:
cc @scylladb/seastar-maint 

I'm conflicted about whether to give it a new name, or to fold it into try_future.

Options:
1. new name (so you can try_future an std::expected and get the expected as the value)
2. try_future (short easy expected name), but in principle breaks source compatibility
3. try_future (with a cmake flag to disable it, to give time for users to eliminate their incompatible use)

avikivity:
v2: avoid nodiscard violations on clang

avikivity:
Alternative (better IMO): https://github.com/scylladb/seastar/pull/3541

travisdowns:
@avikivity wrote:

> 2\. try_future (short easy expected name), but in principle breaks source compatibility

Do you mean that `try_future` would effectively have an overload which behaves differently when the called coro returns std::expected? Even apart from the breakage, I don't think that's a great API.

Using the longer name seems OK and if it's really popular folks can alias it to a shorter one locally?

I consider the whole `try_future` situation kind of unfortunate: it's mostly strictly better performance-wise to use, so you'd kind of always want it, but who wants to add this boilerplate everywhere?

avikivity:
> @avikivity wrote:
> 
> > 2. try_future (short easy expected name), but in principle breaks source compatibility
> 
> Do you mean that `try_future` would effectively have an overload which behaves differently when the called coro returns std::expected? 

Yes. https://github.com/scylladb/seastar/pull/3541

> Even apart from the breakage, I don't think that's a great API.

I understand, that's why put up a competing patch instead of replacing it.

On the other hand, it's quite slick. Just write co_await try_future(...), and it works.

btw, with the advent of [[coro_await_elidable]], we might have to make all the coroutine decorators accept lambdas instead of futures.

> 
> Using the longer name seems OK and if it's really popular folks can alias it to a shorter one locally?


The longer name is okay, but aliasing is not an option, the template try_future is there and you can't give it a new meaning (other than by specializing it, but it's a seastar name, not the application's to override).

> 
> I consider the whole `try_future` situation kind of unfortunate: it's mostly strictly better performance-wise to use, so you'd kind of always want it, but who wants to add this boilerplate everywhere?

People who care about performance?

PR #3535, opened 2026-07-15
coroutine: use symmetric transfer when returning to a waiting coroutine

avikivity:
When a coroutine completes and the task waiting on it is itself a
coroutine, resume that waiter directly via symmetric transfer instead of
scheduling it on the reactor's task queue. This eliminates a task-queue
round-trip on the common "one coroutine awaits another" path, and, being
a guaranteed tail call, does not grow the stack across chains of
back-to-back completions.

The transfer is only taken when the waiter is a coroutine in the current
scheduling group and the task quota is not depleted (!need_preempt());
every other case - a non-coroutine continuation, a waiter in a different
scheduling group, a depleted quota, or no waiter at all - falls back to
the existing schedule() behavior, preserving scheduling-group accounting
and preemption bounds.

Mechanics:

 - task gains a virtual to_coroutine_handle_address() returning nullptr
   by default; only the future-coroutine promise overrides it, so
   non-coroutine waiters transparently take the scheduling fallback.

 - promise_base gains take_waiting_task(), which removes the waiter
   without scheduling it. The coroutine promise captures the waiter this
   way on each synchronous completion path before setting its value, so
   the regular ready path does not schedule it; final_suspend() then
   either transfers to it or schedules it. The deprecated asynchronous
   forward_to() return path deliberately keeps the waiter attached so the
   normal machinery resumes it.

 - final_suspend() now returns a suspending awaiter that self-destroys the
   (self-owned) frame and returns the waiter's handle for transfer, or
   noop_coroutine() otherwise. The reactor's current-task pointer is
   repointed to the waiter before the frame is destroyed so stall
   backtraces taken while the waiter runs stay valid.

Capturing the waiter for final_suspend() is only correct on paths that
actually reach final_suspend(). The coroutine::exception and
coroutine::try_future awaiters instead resolve the promise via
promise_type::set_exception() and then destroy the frame directly,
bypassing final_suspend(); set_exception() therefore leaves the waiter
attached so make_ready() schedules it as before. (Capturing it there
orphaned the waiter and hung file_io_test's thread-blocked get().) These
paths consequently do not use symmetric transfer yet; extending them,
particularly try_future, is left for a later change.

test_exception_awaiter_wakes_suspended_waiter covers this: it forces a
waiter (a coroutine, and a thread in future::get()) to be suspended
before each bypass awaiter injects its result, and hangs without the fix.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -69,6 +69,36 @@ execute_involving_handle_destruction_in_await_suspend(task* tsk) noexcept {
 #endif
 }
 
+// Defined in reactor.cc; declared here to avoid pulling in the heavy
+// reactor.hh from this widely-included header.
+void set_current_task(task* t);
+
+// Transfer control from a just-completed coroutine to the task waiting on it.
+//
+// `self` is the handle of the completing coroutine; its frame is destroyed here
+// (the frame is self-owned, so it must dispose of itself before transferring).
+// If the waiter is itself a coroutine in the current scheduling group and the
+// task quota is not depleted, resume it directly via symmetric transfer,
+// bypassing the reactor's task queue; the returned handle is then resumed by the
+// coroutine machinery as a tail call. Otherwise fall back to scheduling the
+// waiter through the reactor and return the no-op handle.
+inline std::coroutine_handle<> coroutine_return_to_waiter(task* waiter, std::coroutine_handle<> self) noexcept {
+    if (waiter) {
+        void* addr = waiter->to_coroutine_handle_address();
+        if (addr && waiter->group() == current_scheduling_group() && !need_preempt()) {
+            // The reactor's current-task pointer still refers to `self`, whose
+            // frame we are about to destroy; repoint it at the waiter so that
+            // e.g. stall backtraces taken while the waiter runs stay valid.
+            set_current_task(waiter);
+            self.destroy();
+            return std::coroutine_handle<>::from_address(addr);
+        }
+        schedule(waiter);
+    }
+    self.destroy();
+    return std::noop_coroutine();
+}
+
 class coroutine_allocators {
 public:
     static void* operator new(size_t size) {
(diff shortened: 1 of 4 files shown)


avikivity:
Credits:
 - @tchaikov used it in the async generator
 - @ChuanqiXu9 for nudging me in this direction

avikivity:
v2:
 - fixed bug when `co_await coroutine::try_future` and `co_await coroutine::exception` lost wakeups; they now don't participate in symmetric transfer (deferred to a later patch)
 - add regression test to ensure wakeups in these paths

avikivity:
ScyllaDB Results:



vanilla

```
259073.93 tps ( 58.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   37247 insns/op,   17795 cycles/op,        0 errors)
263437.64 tps ( 58.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   37229 insns/op,   17569 cycles/op,        0 errors)
265493.15 tps ( 58.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   37267 insns/op,   17428 cycles/op,        0 errors)
264574.32 tps ( 58.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   37265 insns/op,   17504 cycles/op,        0 errors)
261664.64 tps ( 58.1 allocs/op,   0.0 logallocs/op,  14.1 tasks/op,   37213 insns/op,   17669 cycles/op,        0 errors)
throughput:
	mean=   262848.74 standard-deviation=2547.97
	median= 263437.64 median-absolute-deviation=1725.58
	maximum=265493.15 minimum=259073.93
instructions_per_op:
	mean=   37244.17 standard-deviation=23.58
	median= 37247.28 median-absolute-deviation=20.88
	maximum=37267.36 minimum=37212.61
cpu_cycles_per_op:
	mean=   17592.96 standard-deviation=143.40
	median= 17568.64 median-absolute-deviation=88.70
	maximum=17794.85 minimum=17427.93
```

symmetric

```
259755.26 tps ( 58.1 allocs/op,   0.0 logallocs/op,  13.1 tasks/op,   37573 insns/op,   17754 cycles/op,        0 errors)
264027.99 tps ( 58.1 allocs/op,   0.0 logallocs/op,  13.1 tasks/op,   37557 insns/op,   17519 cycles/op,        0 errors)
265664.08 tps ( 58.1 allocs/op,   0.0 logallocs/op,  13.1 tasks/op,   37583 insns/op,   17416 cycles/op,        0 errors)
261380.19 tps ( 58.1 allocs/op,   0.0 logallocs/op,  13.1 tasks/op,   37547 insns/op,   17677 cycles/op,        0 errors)
263515.56 tps ( 58.1 allocs/op,   0.0 logallocs/op,  13.1 tasks/op,   37517 insns/op,   17544 cycles/op,        0 errors)

throughput:
	mean=   262868.62 standard-deviation=2317.63
	median= 263515.56 median-absolute-deviation=1488.43
	maximum=265664.08 minimum=259755.26
instructions_per_op:
	mean=   37555.33 standard-deviation=25.87
	median= 37556.76 median-absolute-deviation=18.12
	maximum=37583.21 minimum=37516.71
cpu_cycles_per_op:
	mean=   17582.11 standard-deviation=133.71
	median= 17544.08 median-absolute-deviation=95.17
	maximum=17753.94 minimum=17416.32
```

The reduction in tasks/op shows there's only one coroutine call into another coroutine. Not surprising since most of the code here predates coroutines.

The instruction count increase (37.2k -> 37.6k) is surprising. Maybe it's wasted cycles for coroutines that return to ordinary continuations, maybe some other pessimization.

avikivity:
The lack of support for coroutine::as_future etc may also hurt.

StephanDollberg:
We have a bunch of coroutine / continuation benches in RP. Patch ported to seastar here: https://gist.githubusercontent.com/StephanDollberg/36a14c6bffe9df8a50dc1d1e1cfcae4e/raw/af240be4343185b64b33e375102fd4bd4f09f54f/gistfile0.txt

Probably worth comparing before and after. 

Note from my previous HALO research I think making `final_suspend` not return `std::suspend_never` leads to an optimization barrier. There might even be a microbench in the above showing exactly that.

avikivity:
I expect you're right, at the very least it's packaging all the state back into the coroutine frame, then unpacking it again to run the destructors.

This is pretty sad as symmetric transfer can improve cache locality when returning from a coroutine.

Perhaps it should be an optional feature, to be enabled when a codebase has enough coroutines for it to matter (and perhaps with coro_await_elidable).

PR #3537, opened 2026-07-16
file: query DIO alignment via statx regardless of build headers

### LINKED ISSUE #3536: file: STATX_DIOALIGN DIO alignment query is never compiled in (feature check runs before the defining header is included)
The statx-based DIO memory alignment query added in b8fadb1ea ("block_device: Query alignment requirements separately for memory and I/O") is dead code in every build environment, no matter how new the kernel headers are, because the feature check runs before any header that could define `STATX_DIOALIGN` has been included.

## Details

`src/core/file.cc` opens with:

```c++
// Define STATX_DIOALIGN and related fields if not available from system headers (Linux < 6.1)
#ifndef STATX_DIOALIGN
#define STATX_DIOALIGN 0x00002000U
// Check if struct statx has the stx_dio_mem_align field (added in Linux 6.1)
// We need to extend the structure for older kernels/headers
#define SEASTAR_STATX_NEEDS_DIO_FIELDS 1
#endif
```

At that point the only prior includes are C++ standard headers and `seastar/util/assert.hh` (which includes nothing), so `STATX_DIOALIGN` is never defined there: `sys/stat.h` only enters the translation unit later, transitively via `<xfs/linux.h>`. The `#ifndef` therefore always fires, `SEASTAR_STATX_NEEDS_DIO_FIELDS` is always defined, and `query_statx_mem_align()` always reduces to `return std::nullopt;`.

Preprocessing `src/core/file.cc` with the project's own compile command on a machine with 6.1+ headers (which do define `STATX_DIOALIGN` in both `linux/stat.h` and glibc's `bits/statx-generic.h`) shows the compiled result:

```
query_statx_mem_align(int fd) {
# 1250 "src/core/file.cc"
    return std::nullopt;
}
```

## History

- b8fadb1ea introduced the query guarded by `#ifndef SEASTAR_STATX_NEEDS_DIO_FIELDS`, but nothing anywhere defined that macro, so the statx branch was unconditionally compiled in. It worked on 6.1+ headers, but would fail to compile against older headers (`STATX_DIOALIGN` used as an undeclared identifier).
- 15adc3ce5 (the next day) added the fallback block above, presumably to fix old-header builds, but placed it before the includes — which fixed those builds while silently disabling the feature everywhere.

Nobody noticed because the failure mode is silent: the code falls back to the conservative alignment defaults.

## Impact

DMA buffer memory alignment stays at the fallbacks (`XFS_IOC_DIOINFO`'s `d_mem`, i.e. the logical sector size, for XFS; 4096 generic) instead of the kernel-reported `stx_dio_mem_align`. Since Linux 6.0 direct I/O buffers only need to be aligned to the device DMA alignment, which e.g. NVMe reports as 4 bytes — so on XFS-on-NVMe the required alignment is 512 when it could be `max(4, sizeof(void*))`.

## Suggested fix

Either include `<sys/stat.h>` before the feature check, or (better, and in the spirit of the runtime detection seastar already does) drop the compile-time gate entirely: issue the raw statx syscall unconditionally against a layout-compatible 256-byte struct and rely on the runtime `stx_mask` check — kernels without support simply don't set the bit. I have a patch for the latter approach and will send a PR.

travisdowns:
The statx-based DIO memory alignment query (`query_statx_mem_align`) is compiled out on every system, regardless of kernel header version, because the `STATX_DIOALIGN` feature check at the top of `file.cc` runs before any header that defines the macro has been included. See #3536 for the full analysis and history.

Instead of fixing the include order, this drops the compile-time gate entirely: issue the raw statx syscall unconditionally against a layout-compatible stand-in for `struct statx` (the kernel UAPI struct is fixed at 256 bytes), and rely on the runtime check — kernels without `STATX_DIOALIGN` support simply don't set the bit in `stx_mask` and we fall back as before. When building against 6.1+ headers, `static_assert`s verify the stand-in layout against the real `struct statx`.

This also means the feature no longer depends on the headers of the build environment at all, which matters for builds that target newer kernels than they build against (e.g. old-sysroot toolchains).

The second commit moves the function into `seastar::internal`, declares it in `file-impl.hh`, and adds a unit test asserting the query returns an engaged optional on kernels ≥ 6.1 for filesystems that report it (xfs, ext4) — guarding against a regression of the silent failure, which went unnoticed precisely because the fallback path works.

Verified on kernel 6.17: on XFS-on-NVMe the query now returns 4 (the device DMA alignment), where the `XFS_IOC_DIOINFO` fallback reports 512, so DMA buffer memory alignment drops to `max(4, sizeof(void*))` = 8 bytes.

Fixes #3536

### DIFF
--- a/src/core/file-impl.hh
+++ b/src/core/file-impl.hh
@@ -27,6 +27,7 @@
 #include <deque>
 #include <functional>
 #include <memory>
+#include <optional>
 #include <vector>
 #include <sys/uio.h>
 
@@ -37,6 +38,11 @@ namespace internal {
 
 struct fs_info;
 class io_sink;
+
+// Query DIO memory alignment using statx (kernel 6.1+). Returns the memory
+// buffer alignment required for direct I/O on the file, or std::nullopt if
+// the kernel or the filesystem doesn't report STATX_DIOALIGN.
+std::optional<size_t> query_statx_mem_align(int fd);
 
 }
 
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -1220,7 +1220,7 @@ xfs_concurrency_from_kernel_version() {
     return 0;
 }
 
-namespace {
+namespace internal {
 
 // Query DIO memory alignment using statx (kernel 6.1+)
 // Returns the optimal memory buffer alignment for this file descriptor,
@@ -1258,6 +1258,10 @@ query_statx_mem_align(int fd) {
     return std::nullopt;
 }
 
+} // namespace internal
+
+namespace {
+
 // Query device-level alignment properties (common to all filesystems)
 struct device_alignment_info {
     std::optional<unsigned> memory_alignment;      // from statx
@@ -1268,7 +1272,7 @@ device_alignment_info query_device_alignment_info(int fd, dev_t device_id) {
     device_alignment_info info;
 
     // Query statx for DIO memory alignment (kernel 6.1+)
-    info.memory_alignment = query_statx_mem_align(fd);
+    info.memory_alignment = internal::query_statx_mem_align(fd);
 
     // Check for physical_block_size override from io_properties.yaml
     // Note: I/O queues may not be registered for all devices (e.g., tmpfs, test environments)
--- a/src/core/file.cc
+++ b/src/core/file.cc
@@ -24,6 +24,7 @@
 #include <atomic>
 #include <bit>
 #include <coroutine>
+#include <cstddef>
 #include <deque>
 #include <functional>
 #include <memory>
@@ -33,15 +34,8 @@
 
 #define __user /* empty */  // for xfs includes, below
 
-// Define STATX_DIOALIGN and related fields if not available from system headers (Linux < 6.1)
-#ifndef STATX_DIOALIGN
-#define STATX_DIOALIGN 0x00002000U
-// Check if struct statx has the stx_dio_mem_align field (added in Linux 6.1)
-// We need to extend the structure for older kernels/headers
-#define SEASTAR_STATX_NEEDS_DIO_FIELDS 1
-#endif
-
 #include <sys/syscall.h>
+#include <sys/stat.h> // statx etc
 #include <dirent.h>
 #include <linux/types.h> // for xfs, below
 #include <linux/fs.h> // BLKBSZGET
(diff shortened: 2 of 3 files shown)


travisdowns:
@scylladb/seastar-maint this raises a larger question. Current sesatar often takes the approach of compiling out certain syscalls if the local headers at compile time don't have the associated defines. At runtime, we still handle the situation gracefully when the kernel doesn't implement the call. 

It doesn't make much sense to me: we already do dynamic handling at runtime, so there is no need to restrict the impl at compile time based on the local headers. I feel like we should always compile in the capability (if possible) and rely on the runtime check, to not couple the binary to the build system's capability.

As a specific example, we compile Redpanda using a "sysroot" from Debian which has 5.15-ish headers, but we run that binary anywhere (older than 5.15, or newer, usually newer). We don't want to be restricted to 5.15 features.

Note this only applies when we are doing the syscall directly (often the case), it doesn't make any sense if we are talking about the existence of a libc function or not.

So that's the approach I took here, WDYT?

avikivity:
> @scylladb/seastar-maint this raises a larger question. Current sesatar often takes the approach of compiling out certain syscalls if the local headers at compile time don't have the associated defines. At runtime, we still handle the situation gracefully when the kernel doesn't implement the call.
> 
> It doesn't make much sense to me: we already do dynamic handling at runtime, so there is no need to restrict the impl at compile time based on the local headers. I feel like we should always compile in the capability (if possible) and rely on the runtime check, to not couple the binary to the build system's capability.
> 
> As a specific example, we compile Redpanda using a "sysroot" from Debian which has 5.15-ish headers, but we run that binary anywhere (older than 5.15, or newer, usually newer). We don't want to be restricted to 5.15 features.
> 
> Note this only applies when we are doing the syscall directly (often the case), it doesn't make any sense if we are talking about the existence of a libc function or not.
> 
> So that's the approach I took here, WDYT?

We should just be practical. If a compile-time dependency on the headers means someone building on SYSV UNIX won't get some obscure syscall, no need to go to the trouble of replicating the ABI in our own headers. If a super-important new syscall is only available in Linux 19.53 release 8 minutes ago, then is makes sense not to assume people have their build machines downloading kernels from kernel.org in a loop and build against them. Everything else is somewhere in between these extremes, and we can choose what's important, and also adjust later if a user surfaces with stricter requirements.

I have a mild preference towards using the original headers, but acknowledge that it doesn't work everywhere.

We have linux-aio in our own headers: e44bc6f97337f5a0d803dded960ff9f555d1af18 (for a different reason)

travisdowns:
> We should just be practical. If a compile-time dependency on the headers means someone building on SYSV UNIX won't get some obscure syscall, no need to go to the trouble of replicating the ABI in our own headers. If a super-important new syscall is only available in Linux 19.53 release 8 minutes ago, then is makes sense not to assume people have their build machines downloading kernels from kernel.org in a loop and build against them. Everything else is somewhere in between these extremes, and we can choose what's important, and also adjust later if a user surfaces with stricter requirements.

OK fair, I take this as carte blanche to start implementing portable calls then, because I think it's really more the second case, and the replication is in general very small.

A practical reason to this is that it's often popular to build applications on an old distro intentionally to get good libc compatibility, but this would then shut off most new kernel features unnecessarily, leading to a pointless tension.

travisdowns:
BTW this change resulted in a large practical difference: the required memory alignment dropped from 512 to 4 for 6.1+ kernels. So I wouldn't be surprised if something breaks.

avikivity:
> > We should just be practical. If a compile-time dependency on the headers means someone building on SYSV UNIX won't get some obscure syscall, no need to go to the trouble of replicating the ABI in our own headers. If a super-important new syscall is only available in Linux 19.53 release 8 minutes ago, then is makes sense not to assume people have their build machines downloading kernels from kernel.org in a loop and build against them. Everything else is somewhere in between these extremes, and we can choose what's important, and also adjust later if a user surfaces with stricter requirements.
> 
> OK fair, I take this as carte blanche to start implementing portable calls then, because I think it's really more the second case, and the replication is in general very small.
> 
> A practical reason to this is that it's often popular to build applications on an old distro intentionally to get good libc compatibility, but this would then shut off most new kernel features unnecessarily, leading to a pointless tension.

We build on late-model Fedora and ship all the libraries. Kind of like containers, but without the actual containers.

travisdowns:
> We build on late-model Fedora and ship all the libraries. Kind of like containers, but without the actual containers.

Right, same, except Ubuntu.

PR #3538, opened 2026-07-16
`thread`: track context switches in the `ucontext` variant too

WillemKauf:
Commit 8ced1f1832c80365a5c6869ea6120d9f3d78995a sets `g_context_switch_in_progress` around the `setjmp`/`longjmp` based switches only. Sanitized builds take the `SEASTAR_ASAN_ENABLED` branch and switch stacks via `swapcontext()`, which has the same problem: glibc's aarch64 swapcontext carries an empty (nop-only) CFI program while its third instruction already clobbers the link register, so a backtrace taken from a signal handler that lands mid-switch walks garbage frames and can crash in libgcc's unwinder.

Move `{begin}/{end}_context_switch()` to shared scope and apply them to the ucontext-based switch functions, mirroring the `setjmp` variant: the flag set before a switch is cleared at the resume point control transfers to, or in `initial_switch_in_completed()` for a thread's first entry.

### DIFF
--- a/src/core/thread.cc
+++ b/src/core/thread.cc
@@ -50,6 +50,16 @@ thread_local jmp_buf_link* g_current_context;
 
 namespace {
 thread_local std::atomic_flag g_context_switch_in_progress{};
+
+inline void begin_context_switch() noexcept {
+    g_context_switch_in_progress.test_and_set(std::memory_order_relaxed);
+    std::atomic_signal_fence(std::memory_order_seq_cst);
+}
+
+inline void end_context_switch() noexcept {
+    std::atomic_signal_fence(std::memory_order_seq_cst);
+    g_context_switch_in_progress.clear(std::memory_order_relaxed);
+}
 }
 
 #ifdef SEASTAR_ASAN_ENABLED
@@ -80,6 +90,7 @@ thread_local jmp_buf_link* g_previous_context;
 
 void jmp_buf_link::initial_switch_in(ucontext_t* initial_context, const void* stack_bottom, size_t stack_size)
 {
+    begin_context_switch();
     auto prev = std::exchange(g_current_context, this);
     link = prev;
     g_previous_context = prev;
@@ -87,10 +98,12 @@ void jmp_buf_link::initial_switch_in(ucontext_t* initial_context, const void* st
     swapcontext(&prev->context, initial_context);
     __sanitizer_finish_switch_fiber(g_current_context->fake_stack, &g_previous_context->stack_bottom,
                                     &g_previous_context->stack_size);
+    end_context_switch();
 }
 
 void jmp_buf_link::switch_in()
 {
+    begin_context_switch();
     auto prev = std::exchange(g_current_context, this);
     link = prev;
     g_previous_context = prev;
@@ -98,10 +111,12 @@ void jmp_buf_link::switch_in()
     swapcontext(&prev->context, &context);
     __sanitizer_finish_switch_fiber(g_current_context->fake_stack, &g_previous_context->stack_bottom,
                                     &g_previous_context->stack_size);
+    end_context_switch();
 }
 
 void jmp_buf_link::switch_out()
 {
+    begin_context_switch();
     g_current_context = link;
     g_previous_context = this;
     __sanitizer_start_switch_fiber(&fake_stack, g_current_context->stack_bottom,
@@ -109,6 +124,7 @@ void jmp_buf_link::switch_out()
     swapcontext(&context, &g_current_context->context);
     __sanitizer_finish_switch_fiber(g_current_context->fake_stack, &g_previous_context->stack_bottom,
                                     &g_previous_context->stack_size);
+    end_context_switch();
 }
 
 void jmp_buf_link::initial_switch_in_completed()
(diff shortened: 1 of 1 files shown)


src/core/thread.cc
```diff
@@ -50,6 +50,16 @@ thread_local jmp_buf_link* g_current_context;
 
 namespace {
 thread_local std::atomic_flag g_context_switch_in_progress{};
```
avikivity, line 53: > `thread`: track context switches in the `ucontext` variant too

No need to quote "thread".

> glibc's aarch64 swapcontext carries an empty (nop-only) CFI program
while its third instruction already clobbers the link register, so a

Please add a reference to the bug tracker entry for this.

WillemKauf: > No need to quote "thread".

Removed.

> Please add a reference to the bug tracker entry for this.

I don't believe there is one from my searching, obviously the setjmp/longjmp crashes on x86 are referenced in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71890, but I don't see an equivalent discussion around swapcontext on aarch64. I derived this PR from some debug analysis of a crash in `redpanda`'s `cpu_profiler` running on an aarch64 ASan build, which hit a `SIGSEGV` in `uw_update_context` while running in an `ss::async` thread. Through some additional LLM input I came to the conclusion in the commit message.

WillemKauf: For context, this is what I had in my `redpanda` broker logs:

```
Recorded crash reason to crash file on shard 1 (SIGSEGV)
AddressSanitizer:DEADLYSIGNAL
=================================================================
==2062==ERROR: AddressSanitizer: SEGV on unknown address 0x00000000080e (pc 0xffffa31ae320 bp 0xfbff9fae0c40 sp 0xfbff9fae0c40 T1)
==2062==The signal is caused by a READ memory access.
==2062==Hint: address points to the zero page.
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer: nested bug in the same thread, aborting.
```

decoding debug symbols,

* `pc 0xffffa31ae320` -> `libgcc_s.so.1+0xe320: ldr x19, [x19]`, which `addr2line` resolves to `uw_update_context()` at [libgcc/unwind-dw2.c:1544](https://github.com/gcc-mirror/gcc/blob/releases/gcc-12.3.0/libgcc/unwind-dw2.c#L1544). The load itself seems to be line 1543's `_Unwind_GetGR()`, whose dereference has a topical comment  (`/* This will segfault if the register hasn't been saved.*/`). [unwind-dw2.c:L249](https://github.com/gcc-mirror/gcc/blob/8fc1a49c9312b05d925b7d21f1d2145d70818151/libgcc/unwind-dw2.c#L249)
* `0x00000000080e` (fault address) -> the garbage slot pointer in `val = context->reg[index]`: [unwind-dw2.c:L236](https://github.com/gcc-mirror/gcc/blob/8fc1a49c9312b05d925b7d21f1d2145d70818151/libgcc/unwind-dw2.c#L236).
* `bp == sp == 0xfbff9fae0c40` -> apparently expected inside `uw_update_context()`, and the `0xfbff9fae0c40` address is the `sigaltstack`, showing that the unwind ran inside the profiler's signal handler

avikivity: > > No need to quote "thread".
> 
> Removed.
> 
> > Please add a reference to the bug tracker entry for this.
> 
> I don't believe there is one from my searching, obviously the setjmp/longjmp crashes on x86 are referenced in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71890, but I don't see an equivalent discussion around swapcontext on aarch64. I derived this PR from some debug analysis of a crash in `redpanda`'s `cpu_profiler` running on an aarch64 ASan build, which hit a `SIGSEGV` in `uw_update_context` while running in an `ss::async` thread. Through some additional LLM input I came to the conclusion in the commit message.

This is Avi-speak for "file a bug report before committing a workaround".

WillemKauf: Ah, fair enough :slightly_smiling_face:

travisdowns: I filed:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=127112

for the libgcc (unwinder) side and

https://sourceware.org/bugzilla/show_bug.cgi?id=34575

for the glibc (setcontext, etc) side.

I wouldn't hold my breath though.

travisdowns:
https://github.com/scylladb/seastar/commit/e8efedbca2456897e1f0487465b2924719b59343 is pure rebase

travisdowns (approved):

travisdowns:
@scylladb/seastar-maint I plan to merge this unless there are objections in the next week or so

src/core/thread.cc
WillemKauf: my bad for letting this slide, I applied for my `gcc` bugzilla account and was promptly sidetracked by other things for the last month. Thank you Travis!

travisdowns: Sure, any time.

Feels like we are probably better off just swapping out the unwinder and this point, but this is still strictly better than what we have.

avikivity: > I wouldn't hold my breath though.

It will come back from the dead in 7-8 years and resolve itself.

PR #3541, opened 2026-07-19
coroutine: propagate std::expected errors from try_future

avikivity:
Extend coroutine::try_future so that awaiting a future<std::expected<V, E>> also propagates unexpected values, not just exceptions. This is the std::expected counterpart of try_future's exception handling, analogous to rust's try operator: the error is propagated up the std::expected chain without being thrown.

For a future<std::expected<V, E>>, try_future distinguishes three outcomes of the awaited future:

 - if the future failed (holds an exception), the exception is forwarded to the waiter directly and the coroutine is destroyed, exactly like for a non-std::expected result;
 - if the future resolved with an unexpected, it is rebound to the coroutine's return type (a future<std::expected<V2, E>> with a matching error type) and returned to the waiter, without resuming the coroutine;
 - if the future resolved with a value, that value is the result of the co_await.

Rather than a separate awaiter, the std::expected-specific checks are folded into the existing try_future_awaiter behind an if constexpr, so both the preemption-checking and non-preemption-checking variants get the behavior for free. Since a ready future offers no non-destructive way to inspect its expected, await_ready() unwraps it and puts it right back via make_ready_future(), so a ready value can still take the fast path without suspending.

The behavior is guarded by the SEASTAR_TRY_FUTURE_EXPECTED macro (the Seastar_TRY_FUTURE_EXPECTED CMake option), on by default. When disabled, awaiting a future<std::expected> through try_future is deprecated and behaves like any other result: only a failed future exits the coroutine, and the whole std::expected is handed out as the co_await result.

try_future is now exported from the seastar module. The behavior is tested in coroutines_test in both configurations, and documented in the coroutine exception handling chapter of doc/tutorial.md.

### DIFF
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -583,6 +583,53 @@ seastar::future<int> exception_propagating() {
 }
 ```
 
+### Propagating `std::expected` errors
+
+When errors are represented as values rather than exceptions, a function returns a
+`future<std::expected<T, E>>`: the future carries either a value, or an error `E`, or (should
+something go wrong at the future level) an exception. When awaited through `coroutine::try_future`,
+a `future<std::expected<T, E>>` extends the error propagation described above to unexpected values.
+It inspects the three possible outcomes of the awaited future and:
+
+- if the future resolved with a value, that value becomes the result of the `co_await`;
+- if the future resolved with an unexpected, the error is rebound to the coroutine's return type
+  (which must be a `future<std::expected<U, E>>` with a matching error type `E`) and returned to
+  the waiter directly, without resuming the coroutine;
+- if the future failed with an exception, the exception is forwarded to the waiter directly, just
+  like for a non-`std::expected` result.
+
+In all cases the error is propagated without throwing, and the code following the `co_await` only
+runs on the success path. This is analogous to rust's
+[try operator](https://google.github.io/comprehensive-rust/error-handling/try.html) applied to
+`Result`.
+
+Example:
+
+```cpp
+enum class error { not_found, corrupted };
+
+seastar::future<std::expected<int, error>> read_value();
+
+seastar::future<std::expected<seastar::sstring, error>> read_value_as_string() {
+    // If read_value() resolves with an error, it is rebound to this coroutine's
+    // return type and returned directly. If it fails with an exception, the
+    // exception is propagated to the waiter. Otherwise, value holds the int.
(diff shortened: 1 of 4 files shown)


doc/tutorial.md
```diff
@@ -583,6 +583,53 @@ seastar::future<int> exception_propagating() {
...
+### Propagating `std::expected` errors
+
+When errors are represented as values rather than exceptions, a function returns a
+`future<std::expected<T, E>>`: the future carries either a value, or an error `E`, or (should
+something go wrong at the future level) an exception. When awaited through `coroutine::try_future`,
+a `future<std::expected<T, E>>` extends the error propagation described above to unexpected values.
+It inspects the three possible outcomes of the awaited future and:
+
+- if the future resolved with a value, that value becomes the result of the `co_await`;
+- if the future resolved with an unexpected, the error is rebound to the coroutine's return type
```
nuivall, line 595: since we have 2 coroutines involved, maybe we can say it's the caller or parent: `error is rebound to the caller's coroutine return type`

doc/tutorial.md
```diff
@@ -583,6 +583,53 @@ seastar::future<int> exception_propagating() {
...
+
+seastar::future<std::expected<int, error>> read_value();
+
+seastar::future<std::expected<seastar::sstring, error>> read_value_as_string() {
+    // If read_value() resolves with an error, it is rebound to this coroutine's
+    // return type and returned directly. If it fails with an exception, the
+    // exception is propagated to the waiter. Otherwise, value holds the int.
+    auto value = co_await seastar::coroutine::try_future(read_value());
+
+    // Only reached if read_value() resolved with a value.
```
nuivall, line 619: to be super clear maybe we can add that it `resolved with expected value (int in this case)`.
unexpected value may be considered a value...

doc/tutorial.md
avikivity: yes

doc/tutorial.md
avikivity: @travisdowns didn't like the excessive magic, we might go with the more explicit try_expected_future.

nuivall: ok, I think from the user perspective it's a bit nicer that try_future knows what to do with expected, and the outcome is analogous to base form ( so works as expected ;) ), it's just a matter of documentation. Maybe from maintaining pov it's more difficult to assure it works.

PR #3542, opened 2026-07-20
CMakeLists: link URING::uring as PRIVATE, keep include dir PUBLIC

knowack1:
PR scylladb/seastar#3490 linked URING::uring as PUBLIC so consumers find liburing.h, which is included in seastar's public headers.

However making the link PUBLIC also propagates -luring to consumers. This is not needed, since consumers do not call into liburing directly, which is only used internally by seastar.

Additionally this makes the inconsistent linkage flags on consumer side when linking with seastar using .pc file vs cmake, since the .pc file does not add -luring.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1147,8 +1147,11 @@ set_option_if_package_is_found (Seastar_IO_URING LibUring)
 if (Seastar_IO_URING)
   target_compile_definitions (seastar
     PUBLIC SEASTAR_HAVE_URING)
-  target_link_libraries (seastar
-    PUBLIC URING::uring)
+  # Uring include directories are public as seastar includes them in its public headers.
+  # The library is private as seastar uses it internally.
+  target_link_libraries (seastar PRIVATE URING::uring)
+  target_include_directories (seastar
+    SYSTEM PUBLIC $<TARGET_PROPERTY:URING::uring,INTERFACE_INCLUDE_DIRECTORIES>)
 endif ()
 
 if (Seastar_LD_FLAGS)


knowack1:
This problem was discovered after a Seastar bump in ScyllaDB: https://github.com/scylladb/scylladb/pull/30762. 

ScyllaDB verifies that the linker flags produced by the pkg-config (.pc)-based  build system and the CMake-based build system are the same. This check failed because the CMake-based build system includes an additional `-luring`: https://github.com/scylladb/scylladb/actions/runs/29724880716/job/88295603902?pr=30762
```
Run git config --global --add safe.directory $GITHUB_WORKSPACE
debug: ⚠ SKIPPED (canary 'dev' failed)
dev: ✗ MISMATCH
  Linker: 48 targets with lib diffs
    lib only in CMake: uring  (48 targets)
release: ⚠ SKIPPED (canary 'dev' failed)
sanitize: ⚠ SKIPPED (canary 'dev' failed)
coverage: ⚠ SKIPPED (canary 'dev' failed)
Error: Process completed with exit code 1.
```

CC @tchaikov

knowack1:
@avikivity Could you review whether this is the right fix? I believe it is - only the liburing include directories need to propagate to consumers (the intention of  #3490) , not `-luring` itself.

avikivity:
I have no idea, my strategy with cmake is to make random changes until whatever didn't work starts working.

CMakeLists.txt
```diff
@@ -1147,8 +1147,11 @@ set_option_if_package_is_found (Seastar_IO_URING LibUring)
 if (Seastar_IO_URING)
   target_compile_definitions (seastar
```
avikivity, line 1148: > However making the link PUBLIC also propagates -luring to consumers.
This is not needed, since consumers do not call into liburing directly
- is only used internally by seastar.

Won't it be needed for static linking?

knowack1: Yes, it will be needed for static linking. However I think this will be handled by CMake correctly as it was before https://github.com/scylladb/seastar/pull/3490. I will double-check.

knowack1: I just double-checked the build system behavior regarding static vs. dynamic linking and compared the linker command-line output using CMake and pkg-config. After this patch, both pkg-config and CMake linker flags look the same: liburing is added for static linking, while it is omitted for dynamic linking (resolved by the linker from libseastar.so ELF):

| Config                                                                                                       | `liburing` present? | Exact link command                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
| ------------------------------------------------------------------------------------------------------------ | ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **1) CMake, shared** (`add_subdirectory` + `target_link_libraries(probe PRIVATE seastar)`, `libseastar.so`) | **No**              | `clang++ -O2 -Xlinker --dependency-file=CMakeFiles/probe.dir/link.d CMakeFiles/probe.dir/main.cc.o -o probe -Wl,-rpath,.../build-shared/seastar-build seastar-build/libseastar.so /usr/lib64/libboost_program_options.so.1.90.0 /usr/lib64/libboost_container.so.1.90.0 /usr/lib64/libcares.so /usr/lib64/libfmt.so.11.2.0 /usr/lib64/liblz4.so`                                                                                                                                |
| **2) CMake, static** (`add_subdirectory` + `target_link_libraries(probe PRIVATE seastar)`, `libseastar.a`)  | **Yes**             | `clang++ -O2 -g -DNDEBUG -Xlinker --dependency-file=CMakeFiles/probe.dir/link.d CMakeFiles/probe.dir/main.cc.o -o probe seastar-build/libseastar.a /usr/lib64/libboost_program_options.so.1.90.0 /usr/lib64/libboost_container.so.1.90.0 /usr/lib64/libcares.so /usr/lib64/libfmt.so.11.2.0 /usr/lib64/liblz4.so -ldl -latomic /usr/lib64/libsctp.so /usr/lib64/libprotobuf.so /usr/lib64/libyaml-cpp.so /usr/lib64/libhwloc.so /usr/lib64/libgnutls.so /usr/lib64/liburing.so` |
| **3) pkg-config `--shared`** (`pkg-config --libs --shared seastar`, `libseastar.so`)                         | **No**              | `clang++ /home/.../seastar-build-shared/libseastar.so /usr/lib64/libboost_program_options.so.1.90.0 /usr/lib64/libcares.so /usr/lib64/libfmt.so.11.2.0 -llz4 -Xlinker --dependency-file=CMakeFiles/probe.dir/link.d CMakeFiles/probe.dir/main.cc.o -o probe`                                                                                                                                                                                                                    |
| **4) pkg-config `--static`** (`pkg-config --libs --static seastar`, `libseastar.a`)                          | **Yes**             | `clang++ -O2 -g -DNDEBUG /home/.../seastar-build-static/libseastar.a /usr/lib64/libboost_program_options.so.1.90.0 /usr/lib64/libcares.so /usr/lib64/libfmt.so.11.2.0 -ldl /usr/lib64/libsctp.so /usr/lib64/liburing.so -latomic -llz4 -lgnutls -lgmp -lunistring -lhogweed -lnettle -ltasn1 -lidn2 -lp11-kit -lprotobuf -lz -lhwloc -lm -ludev -lpthread -luring -lyaml-cpp -Xlinker --dependency-file=CMakeFiles/probe.dir/link.d CMakeFiles/probe.dir/main.cc.o -o probe`    |

PR #3547, opened 2026-07-21
reactor: sort tasks in task queue to maximize instruction cache utilization

avikivity:
Increase the affinity of tasks with instruction locality by placing them close
by in the task queue. This increases the chance that the instruction cache
will hold the instructions for the tasks, and similarly for the branch predictor
training history.

Tasks are deemed similar if they share the virtual table pointer. The matching
is lossy, so unrelated tasks can be executed together with some probability.

Note: I was not able to show any gain from this when it was written.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2673,8 +2673,8 @@ bool reactor::task_queue::run_tasks() {
     // Make sure new tasks will inherit our scheduling group
     *internal::current_scheduling_group_ptr() = scheduling_group(_id);
     auto& tasks = _q;
-    while (!tasks.empty()) {
-      auto frag = &tasks.fragments.front();
+    auto frag = tasks.fragments.begin();
+    while (frag != tasks.fragments.end()) {
       // Note: we have to reload frag->start, because queue_fragment::push_front() can update it.
       while (frag->start < frag->end) {
         auto tsk = frag->tasks[frag->start ++];
@@ -2691,10 +2691,11 @@ bool reactor::task_queue::run_tasks() {
         if (internal::scheduler_need_preempt()) {
             if (tasks.size() <= r._cfg.max_task_backlog) {
                 if (frag->start == frag->end) {
-                    tasks.fragments.erase(tasks.fragments.iterator_to(*frag));
-                    delete frag;
-                    frag = nullptr;
+                    auto tmp = &*frag;
+                    frag = tasks.fragments.erase(frag);
+                    delete tmp;
                 }
+                frag = tasks.fragments.end(); // exit outer loop
                 break;
             } else {
                 // While need_preempt() is set, task execution is inefficient due to
@@ -2713,12 +2714,10 @@ bool reactor::task_queue::run_tasks() {
             }
         }
       }
-      if (frag && frag->start == frag->end) {
-          tasks.fragments.erase(tasks.fragments.iterator_to(*frag));
-          delete frag;
-      }
-      if (!frag) {
-          break;
+      if (frag != tasks.fragments.end() && frag->start == frag->end) {
+          auto tmp = &*frag;
+          frag = tasks.fragments.erase(frag);
+          delete tmp;
       }
     }
 
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -972,12 +972,20 @@ void reactor::task_queue::queue_fragment::push_front(task* tsk) noexcept {
 }
 
 void reactor::task_queue::fragmented_queue::push_back(task* tsk) noexcept {
-    if (fragments.empty() || fragments.back().is_full()) {
+    auto type_hash_index = uint32_t(reinterpret_cast<uintptr_t>(&typeid(*tsk))) % type_hash_size;
+    auto frag = type_hash[type_hash_index];
+    if (!frag) {
         // Note: we can't survive allocation failure here
         // Note: the fragments list takes ownership of the new queue_fragment
-        fragments.push_back(*new queue_fragment);
+        frag = new queue_fragment(type_hash_index);
+        type_hash[type_hash_index] = frag;
+        fragments.push_back(*frag);
     }
-    fragments.back().push_back(tsk);
+    frag->push_back(tsk);
+    if (frag->is_full()) {
+        type_hash[type_hash_index] = nullptr;
+        frag->type_hash_index = type_hash_size;
+    }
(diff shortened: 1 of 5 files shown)


src/core/reactor.cc
```diff
@@ -1024,6 +959,46 @@ reactor::task_queue::task_queue(task_queue_group* parent, unsigned id, sstring n
...
 }
 
+bool reactor::task_queue::queue_fragment::can_push_front() const {
+    return start != 0;
+    //return end - start != max_tasks;
+}
+
+void reactor::task_queue::queue_fragment::push_front(task* tsk) noexcept {
+    if (end == 0) {
+        start = end = 1;
```
xemul, line 969: It means that only one push_front() is allowed, subsequent will allocate new queue_fragment

src/core/reactor.cc
```diff
@@ -2721,68 +2655,56 @@ void reactor::register_metrics() {
...
     // Make sure new tasks will inherit our scheduling group
     *internal::current_scheduling_group_ptr() = scheduling_group(_id);
-    while (!_q.empty()) {
-        auto tsk = _q.front();
-        _q.pop_front();
-        SEASTAR_REACTOR_TRACE(task_start, _id, _name.c_str());
+    auto& tasks = _q;
+    auto frag = tasks.fragments.begin();
+    while (frag != tasks.fragments.end()) {
+      // Note: we have to reload frag->start, because queue_fragment::push_front() can update it.
```
xemul, line 2686: But push_front() may also insert a new grag ahead of the current one, so add_urgent_task() is no longer urgen and will be processed next tick. This is "a regression" against current curculer_buffer which always pops from the front of the global queue

tests/unit/queue_test.cc
```diff
@@ -68,7 +68,7 @@ SEASTAR_THREAD_TEST_CASE(test_queue_pop_eventually) {
         testlog.trace("pusher: full={} empty={} stop={}", q.full(), q.empty(), stop);
         return q.push_eventually(std::move(data)).then([&] {
             pushed++;
-            if (stop && !q.empty()) {
+            if (stop && pushed > 100) {
```
xemul, line 71: Why did this condition change?

src/core/reactor.cc
```diff
@@ -2721,68 +2655,56 @@ void reactor::register_metrics() {
...
+      while (frag->start < frag->end) {
+        auto tsk = frag->tasks[frag->start ++];
+        --tasks.nr_tasks;
```
xemul, line 2689: nit: It's probably better to have a pop_front method, rather than updating two counteres by hand from the caller

src/core/reactor.cc
```diff
@@ -1024,6 +959,46 @@ reactor::task_queue::task_queue(task_queue_group* parent, unsigned id, sstring n
...
+        start = end = 1;
+    }
+    tasks[--start] = tsk;
+}
+
+void reactor::task_queue::fragmented_queue::push_back(task* tsk) noexcept {
+    auto type_hash_index = uint32_t(reinterpret_cast<uintptr_t>(&typeid(*tsk))) % type_hash_size;
+    auto frag = type_hash[type_hash_index];
+    if (!frag) {
+        // Note: we can't survive allocation failure here
```
xemul, line 978: Should the scoped_critical_allocation_section be here in that case?

xemul:
> Note: I was not able to show any gain from this when it was written.

How did you test it? There's no perf test in this PR

src/core/reactor.cc
avikivity: I need to refresh myself about when push_front()s make sense. I remember one is needed when we install or remove a poller, and for "urgent" tasks.

src/core/reactor.cc
avikivity: Well, urgent tasks are an optimization, so if sorting tasks is overall a win, then it's okay to lose it.

However, last time I measured (long ago), sorting tasks was not a win.

tests/unit/queue_test.cc
avikivity: Unfortunately, I don't remmeber.

If this ever matures into something mergable, I'll either revert this change, or come up with an explanation.

src/core/reactor.cc
avikivity: yes

src/core/reactor.cc
avikivity: yes

avikivity:
> > Note: I was not able to show any gain from this when it was written.
> 
> How did you test it? There's no perf test in this PR

I tested it with ScyllaDB mixed workload, where one would expect a large number of tasks types to cause instruction cache and branch predictor conflicts. It showed a small degradation.

Most likely the branch's value was reduced over time. Intel instruction caches are 32kB and are shared over two hyperthreads. Modern aaaaaaarch64 instruction cache size is 64kB and is dedicated to one vcpu. So we have 4X instruction cache per vcpu. The instruction footprint also grew, but not by this much.

src/core/reactor.cc
```diff
@@ -1024,6 +959,46 @@ reactor::task_queue::task_queue(task_queue_group* parent, unsigned id, sstring n
...
+
+void reactor::task_queue::queue_fragment::push_front(task* tsk) noexcept {
+    if (end == 0) {
+        start = end = 1;
+    }
+    tasks[--start] = tsk;
+}
+
+void reactor::task_queue::fragmented_queue::push_back(task* tsk) noexcept {
+    auto type_hash_index = uint32_t(reinterpret_cast<uintptr_t>(&typeid(*tsk))) % type_hash_size;
```
xemul, line 975: The `&typeid(*tsk)` is likely aligned. Presumably, by the pointer size, so 3 lower bits are zero.  So if it were `(&typeid(*tsk)) & type_hash_mask` some of the slots would be skipped.

However, it's `% type_hash_size` and size if 61, so the index is evently distributed. However, non-power-of-two size for a hash is also unusual. Even though in assembly it looks decent.

If this all is intentional, some code comment would be helpful.

avikivity: Yes, it was intentional. It deserves a comment, and the series is not in a reviewable state yet (mostly because I wasn't able to get good results).

It would be a bad idea to look at the low-order bits (even ignoring the first three), because the compiler/linker could choose to lay out the typeinfo objects one after the other, causing aliasing. % 61 mixes the bits very well and is quite cheap.

PR #3549, opened 2026-07-21
seastar-json2code: add fine-grained module flags

avikivity:
Add --module-std and --module-seastar to switch only the standard library or only seastar dependencies to C++ module imports, independently of each other. --module-dependencies now acts as a shorthand for enabling both.

This allows an applicaiton to incremental switch to using modules.


avikivity:

PR #3550, opened 2026-07-21
httpd: add tls_handshake_errors metric

### LINKED ISSUE #3521: Missing metric for HTTPS handshake failures
Seastar's HTTP server has for metrics, `httpd_connections_total`, `httpd_read_errors`, `httpd_reply_errors`, `httpd_requests_served`.
There is no metric counting how many HTTPS (TLS) connections were rejected because of handshake failures (e.g., bad certificate).  I think having such a metric is important.

`connections_total` is incremented for these rejected connections, by the way. `httpd_read_errors` is not incremented - it is only relevant in connection::read(), which isn't called when the handshake failed.

ankitxvx:
## Problem

Seastar's HTTP server exposes several metrics (`httpd_connections_total`, `httpd_read_errors`, `httpd_reply_errors`, `httpd_requests_served`), but none of them track TLS handshake failures.

When a client connects via HTTPS and the TLS handshake fails (e.g., bad certificate, unsupported cipher), `connections_total` is incremented because the TCP connection was accepted, but no error counter reflects the failure. `read_errors` is not incremented because `connection::read()` is never called — the connection is dropped during `connection::prepare()` before reaching the read phase.

This leaves a blind spot for operators monitoring HTTPS servers.

 

## Solution

Add a new `tls_handshake_errors` counter that is incremented in `do_process_connection()` when `conn->prepare()` throws an exception on a TLS-enabled connection.

The counter is guarded by `if (tls)` so it only tracks TLS-specific failures, not any hypothetical future non-TLS preparation errors.

## Changes

- **`include/seastar/http/httpd.hh`** — Add `_tls_handshake_errors` member and `tls_handshake_errors()` accessor to `http_server`
- **`src/http/httpd.cc`** — Register `tls_handshake_errors` metric in `http_stats`, increment in the `catch` block of `do_process_connection()`, add accessor implementation
- **`tests/unit/httpd_test.cc`** — Add `test_tls_handshake_error_metric` test: sets up an HTTPS server, connects a client that does not trust the server's CA to trigger a handshake failure, and verifies the counter is incremented

## Testing

Added `test_tls_handshake_error_metric` which:
1. Starts an HTTPS server with TLS credentials
2. Connects a client without a trusted CA certificate (forcing handshake failure)
3. Asserts `tls_handshake_errors() >= 1`
4. Asserts `total_connections() >= 1` (confirming the existing behavior is preserved)

Fixes #3521

### DIFF
--- a/src/http/httpd.cc
+++ b/src/http/httpd.cc
@@ -68,7 +68,8 @@ http_stats::http_stats(http_server& server, const sstring& name)
             sm::make_gauge("connections_current", [&server] { return server.current_connections(); }, sm::description("The current number of open  connections"), labels),
             sm::make_counter("read_errors", [&server] { return server.read_errors(); }, sm::description("The total number of errors while reading http requests"), labels),
             sm::make_counter("reply_errors", [&server] { return server.reply_errors(); }, sm::description("The total number of errors while replying to http"), labels),
-            sm::make_counter("requests_served", [&server] { return server.requests_served(); }, sm::description("The total number of http requests served"), labels)
+            sm::make_counter("requests_served", [&server] { return server.requests_served(); }, sm::description("The total number of http requests served"), labels),
+            sm::make_counter("tls_handshake_errors", [&server] { return server.tls_handshake_errors(); }, sm::description("The total number of TLS handshake failures"), labels)
     });
 }
 
@@ -520,6 +521,7 @@ future<> http_server::do_process_connection(connected_socket conn_fd, socket_add
     try {
         co_await conn->prepare();
     } catch (...) {
+        ++_tls_handshake_errors;
         hlogger.debug("connection preparation failed: {}", std::current_exception());
         co_return;
     }
@@ -547,6 +549,9 @@ uint64_t http_server::read_errors() const {
 }
 uint64_t http_server::reply_errors() const {
     return _respond_errors;
+}
+uint64_t http_server::tls_handshake_errors() const {
+    return _tls_handshake_errors;
 }
 
 // Write the current date in the specific "preferred format" defined in
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -130,6 +130,7 @@ class http_server {
     uint64_t _requests_served = 0;
     uint64_t _read_errors = 0;
     uint64_t _respond_errors = 0;
+    uint64_t _tls_handshake_errors = 0;
     shared_ptr<seastar::tls::server_credentials> _credentials;
     sstring _date = http_date();
     timer<> _date_format_timer { [this] {_date = http_date();} };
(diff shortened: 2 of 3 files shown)


src/http/httpd.cc
```diff
@@ -520,6 +521,9 @@ future<> http_server::do_process_connection(connected_socket conn_fd, socket_add
     try {
         co_await conn->prepare();
     } catch (...) {
+        if (tls) {
+            ++_tls_handshake_errors;
+        }
```
avikivity: I think it's better do drop the conditional, so that those hypothetical future errors don't go unnoticed. Better a misleading metric than for failures to go unaccounted.

ankitxvx: Dropped the `if (tls)` check as requested so any failure during `conn->prepare()` increments `_tls_handshake_errors`.

src/http/httpd.cc
```diff
@@ -29,26 +27,26 @@
...
-#include <seastar/core/sharded.hh>
-#include <seastar/core/queue.hh>
-#include <seastar/core/when_all.hh>
 #include <seastar/core/metrics.hh>
 #include <seastar/core/print.hh>
+#include <seastar/core/queue.hh>
+#include <seastar/core/sharded.hh>
+#include <seastar/core/sstring.hh>
+#include <seastar/core/when_all.hh>
+#include <seastar/coroutine/switch_to.hh>
```
avikivity: Why all these changes?

src/http/httpd.cc
```diff
@@ -57,554 +55,623 @@ namespace seastar {
...
-    std::vector<sm::label_instance> labels;
-
-    labels.push_back(sm::label_instance("service", name));
-    _metric_groups.add_group("httpd", {
-            sm::make_counter("connections_total", [&server] { return server.total_connections(); }, sm::description("The total number of connections opened"), labels),
-            sm::make_gauge("connections_current", [&server] { return server.current_connections(); }, sm::description("The current number of open  connections"), labels),
-            sm::make_counter("read_errors", [&server] { return server.read_errors(); }, sm::description("The total number of errors while reading http requests"), labels),
-            sm::make_counter("reply_errors", [&server] { return server.reply_errors(); }, sm::description("The total number of errors while replying to http"), labels),
-            sm::make_counter("requests_served", [&server] { return server.requests_served(); }, sm::description("The total number of http requests served"), labels)
-    });
```
avikivity, line 72: And these? It's impossible to review now.

ankitxvx: Apologies for the noise! An auto-formatter ran on `src/http/httpd.cc` by mistake in the previous push. 

I've cleaned up the file and force-pushed a minimal commit (`c3c42468`). The PR diff is now down to just 3 clean additions (+78 / -1 lines total), so it should be easy to review now. Thanks for pointing that out!

PR #3551, opened 2026-07-23
coroutine: add selection, a poll()-style multiplexer over futures

avikivity:
coroutine::selection lets a coroutine wait on several futures concurrently and react to each as it completes, in completion order, rather than waiting for all of them like coroutine::all.

Each operation is registered with add(), tagged with a caller-chosen token, and returns a future<T> for its result. co_await on the selection parks the coroutine until one registered future becomes ready and returns its token (std::optional<Token>, empty once the selection is drained), giving O(1) dispatch in the spirit of epoll. The matching future is then guaranteed ready, so the caller reads its result without suspending (co_await or get()); exceptions, including a synchronous throw from the callable, propagate through it.

Operations may have different value types, and more may be registered between iterations. Slots are packed into a small inline arena (a bitmap chunk allocator) and spill to the heap only when it fills. Because a future cannot be cancelled and a slot cannot be relocated out of the coroutine frame, the selection must be drained before it is destroyed; violating this is reported via on_fatal_internal_error().

selection is a ready-to-use type (std::size_t token, 512-byte arena); basic_selection<Token, InlineStorageBytes> exposes the parameters.

Deadlines are not built in: coroutine::deadline packages a timer and a promise into a cancellable future that is registered with add() like any other. deadline::cancel() resolves it early so the selection can drain without waiting for the timer to elapse.

The headers are registered in the seastar module, tests live in coroutines_test.cc, and a tutorial section documents the feature.

### DIFF
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -631,6 +631,86 @@ seastar::future<bool> all_exist(std::vector<seastar::sstring> filenames) {
 
 Here, the lambda function passed to parallel_for_each is launched concurrently for each element in the filenames vector. The coroutine is paused until all calls complete.
 
+## Selecting among several results
+
+`seastar::coroutine::all` waits for *all* of its sub-computations and hands you the results together, in launch order. Sometimes a coroutine instead wants to react to several concurrent operations *as they complete* — in completion order — for example to process each result the moment it arrives, or to launch additional work in response. `seastar::coroutine::selection` provides this, modelled loosely on the classic `poll()`/`select()` system calls. (`selection` is a ready-made type using a `size_t` token; `seastar::coroutine::basic_selection` is the underlying template if you want a different token type, clock, or inline-storage size.)
+
+A `selection` is declared as a local variable in the coroutine frame. You register operations with `add()`, tagging each with a caller-chosen *token* (an index, an enum, anything copyable). `add()` hands back a `future<T>` for the result. (A deadline is just another future — see [Timeouts](#timeouts) below.) You then repeatedly `co_await` the selection in a `while` loop: each `co_await` suspends the coroutine until one registered event becomes ready and returns its token (as a `std::optional<Token>`, empty once the selection is drained). Use the token to dispatch; the matching future is then guaranteed ready, so `co_await`-ing it (or calling `get()`) yields the result without suspending — a failed operation propagates its exception, just like any future:
+
+```cpp
+#include <seastar/coroutine/selection.hh>
+
+seastar::future<int> read_from(int shard);
+
+// Read a value from each of several shards, processing each result the moment
+// it arrives rather than waiting for the slowest.
+seastar::future<int> sum_as_they_arrive(std::vector<int> shards) {
+    seastar::coroutine::selection sel;   // tokens are size_t; use the shard index
+    std::vector<seastar::future<int>> reads;
+    for (size_t i = 0; i < shards.size(); ++i) {
+        reads.push_back(sel.add(i, [shard = shards[i]] { return read_from(shard); }));
+    }
+
+    int total = 0;
+    while (auto which = co_await sel) {
+        total += co_await std::move(reads[*which]);   // O(1) dispatch, process on arrival
+    }
+    co_return total;
+}
+```
+
+Because the token is returned directly, dispatch is O(1) — no scanning to find which operation completed, in the spirit of `epoll` versus `poll`/`select`.
+
+A few things to note:
+
+* `add()` takes a *callable* returning a future (like `seastar::coroutine::all`), not a ready-made future. This lets the selection invoke it in a controlled context: an exception thrown synchronously by the callable is captured into the returned future, just like a failed future.
+* The operations may have *different* value types; `add()` returns a correctly-typed `future<T>` for each, so a single `selection` can mix, say, a `future<int>` and a `future<sstring>` (dispatch on the token, then `co_await` the appropriately-typed future).
+* You may register more operations between iterations, but not while the coroutine is suspended in `co_await` (see the timeout example below).
+* Storage for the registered operations is normally taken from a small in-place arena inside the `selection` object, falling back to the heap only when that fills up.
+
+### The draining contract
+
+A `selection` **must be drained before it is destroyed**: the `while` loop must run until `co_await sel` yields a false token, which happens only when nothing is ready and nothing is still pending. This is not merely a convention. A Seastar future cannot be cancelled, and a slot registered inside the coroutine frame cannot be relocated out of it, so an operation that is still in flight when the `selection` is destroyed would leave a dangling reference. Leaving the loop early — via `break`, `return`, or an exception unwinding the frame — while operations are still pending is therefore a programming error and aborts the process.
+
+Consequently a `selection` never abandons work: every registered operation runs to completion and is reported. To "stop caring" about further results, keep looping but ignore them (`co_await`-and-discard the future). The loop still exits only once everything has drained.
+
+### Timeouts
+
+A `selection` has no built-in timeout: a deadline is just another future. The `seastar::coroutine::deadline` helper packages a timer and a promise — it arms a timer on construction and its `get_future()` resolves when the timer expires — so you register it with `add()` like anything else. Crucially, `deadline::cancel()` resolves the future early, which (unlike a plain future) lets you drop a deadline once it is moot so the selection can drain:
+
+```cpp
+seastar::future<> await_with_hedge() {
+    enum class ev { primary, hedge, deadline };
+    // a custom token type (here a scoped enum) needs basic_selection
+    seastar::coroutine::basic_selection<ev, 512> sel;
+    auto primary = sel.add(ev::primary, [] { return read_from(0); });
+    std::optional<seastar::future<int>> hedge;
+
+    seastar::coroutine::deadline deadline(100ms);
+    auto deadline_f = sel.add(ev::deadline, [&] { return deadline.get_future(); });
+
+    std::optional<int> answer;
+    while (auto which = co_await sel) {
+        switch (*which) {
+        case ev::deadline:
+            co_await std::move(deadline_f);
+            // primary is taking too long; start a hedged read alongside it
+            hedge = sel.add(ev::hedge, [] { return read_from(1); });
+            break;
+        case ev::primary:
+            if (!answer) { answer = co_await std::move(primary); }
+            deadline.cancel();   // got an answer; drop the deadline so we can drain
+            break;
+        case ev::hedge:
+            if (!answer) { answer = co_await std::move(*hedge); }
+            break;
+        }
+    }
+}
+```
+
+Because the timeout is an ordinary registered future, the loop will not drain while it is still pending — a 100ms deadline keeps the selection alive for up to 100ms. Calling `deadline.cancel()` resolves it immediately (its token is still reported once), so once you no longer need the deadline you can cancel it and let the loop finish promptly.
+
 ## Breaking up long running computations
 
 Seastar is generally used for I/O, and coroutines usually launch I/O operations and consume their results, with little computation in between. But occasionally a long running computation is needed, and this risks preventing the reactor from performing I/O and scheduling other tasks.
--- a/include/seastar/coroutine/selection.hh
+++ b/include/seastar/coroutine/selection.hh
@@ -1,0 +1,463 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2026-present ScyllaDB
+ */
+
+#pragma once
+
+#include <bit>
+#include <cstdint>
+#include <optional>
+#include <type_traits>
+
+#include <fmt/format.h>
+
+#include <boost/intrusive/list.hpp>
+
+#include <seastar/core/coroutine.hh>
+#include <seastar/core/future.hh>
+#include <seastar/core/lowres_clock.hh>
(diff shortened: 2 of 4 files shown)


avikivity:
cc @gleb-cloudius for suitability for abstract_read_executor
cc @travisdowns for panda feedback

travisdowns:
To start with the shallowest feedback first, why did you choose "selection" over the more common (and shorter), select? I guess because it's a class (so noun-like) not a function (verb-like)?

doc/tutorial.md
```diff
@@ -631,6 +631,86 @@ seastar::future<bool> all_exist(std::vector<seastar::sstring> filenames) {
 
 Here, the lambda function passed to parallel_for_each is launched concurrently for each element in the filenames vector. The coroutine is paused until all calls complete.
 
+## Selecting among several results
+
+`seastar::coroutine::all` waits for *all* of its sub-computations and hands you the results together, in launch order. Sometimes a coroutine instead wants to react to several concurrent operations *as they complete* — in completion order — for example to process each result the moment it arrives, or to launch additional work in response. `seastar::coroutine::selection` provides this, modelled loosely on the classic `poll()`/`select()` system calls. (`selection` is a ready-made type using a `size_t` token; `seastar::coroutine::basic_selection` is the underlying template if you want a different token type, clock, or inline-storage size.)
```
travisdowns, line 636: "launch order"? It's also the order you pass the sub-computations which seems like the more useful way to describe it, since launch order is an internal detail.

doc/tutorial.md
```diff
@@ -631,6 +631,86 @@ seastar::future<bool> all_exist(std::vector<seastar::sstring> filenames) {
...
+seastar::future<int> sum_as_they_arrive(std::vector<int> shards) {
+    seastar::coroutine::selection sel;   // tokens are size_t; use the shard index
+    std::vector<seastar::future<int>> reads;
+    for (size_t i = 0; i < shards.size(); ++i) {
+        reads.push_back(sel.add(i, [shard = shards[i]] { return read_from(shard); }));
+    }
+
+    int total = 0;
+    while (auto which = co_await sel) {
+        total += co_await std::move(reads[*which]);   // O(1) dispatch, process on arrival
```
travisdowns, line 656: `reads[*which]` is guaranteed to be available here, right? So why not just `.get()`? It would make the example clearer as it uses the fact the future is now available.

doc/tutorial.md
```diff
@@ -631,6 +631,86 @@ seastar::future<bool> all_exist(std::vector<seastar::sstring> filenames) {
...
+    }
+    co_return total;
+}
```
travisdowns, line 659: This example also raises the question of whether a simplified variant where `add` returns nothing and `co_await sel` just returns the unwrapped value (`int`, here) directly would be useful. It would avoid a lot of boilerplate. Do you expect this usage of the token being an index into an array of futures to be the dominant one?

travisdowns: I guess a key thing the token indirection adds is the mixed type support. Maybe an example of that would be nice too since it seems a bit awkward to use.

doc/tutorial.md
```diff
@@ -631,6 +631,86 @@ seastar::future<bool> all_exist(std::vector<seastar::sstring> filenames) {
...
+```
+
+Because the token is returned directly, dispatch is O(1) — no scanning to find which operation completed, in the spirit of `epoll` versus `poll`/`select`.
+
+A few things to note:
+
+* `add()` takes a *callable* returning a future (like `seastar::coroutine::all`), not a ready-made future. This lets the selection invoke it in a controlled context: an exception thrown synchronously by the callable is captured into the returned future, just like a failed future.
+* The operations may have *different* value types; `add()` returns a correctly-typed `future<T>` for each, so a single `selection` can mix, say, a `future<int>` and a `future<sstring>` (dispatch on the token, then `co_await` the appropriately-typed future).
+* You may register more operations between iterations, but not while the coroutine is suspended in `co_await` (see the timeout example below).
```
travisdowns, line 668: Any co_await, or only the `co_await sel`? I assume the latter.

avikivity:
> To start with the shallowest feedback first, why did you choose "selection" over the more common (and shorter), select? I guess because it's a class (so noun-like) not a function (verb-like)?

You know me too well.

doc/tutorial.md
avikivity: Right, will adjust.

(IIRC one or more pre-coroutine when_all overloads actually launch in reverse submission order, on the theory that there is increased likelihood they complete in-order, which then reduces the number of task wakeups).

doc/tutorial.md
avikivity: Yes, and that's actually a core promise of the API. Will fix.

doc/tutorial.md
avikivity: The motivating example is ScyllaDB's read executor, which launches a data read and some digest reads (typically 1), sets a timeout timer and a speculative-retry timer. When the speculative-retry timer fires, if we didn't receive a response to one of the reads, we launch a replacement (could be a data read or a digest read). When we receive sufficient responses, we need to deal with digest mismatches (more reads and writes), cancel the timers, and return a response, but still wait for any excessive reads we launched.

So we have at least 4 roles and 3 types: 2 timeouts (in different roles), data read future(s), digest read future(s). This is even before dealing with the reads and writes for healing inconsistencies, which are currently handled in a different class.

This is designed for heterogeneous, dynamic logic. I'll come up with a more illustrative example.

doc/tutorial.md
avikivity: Correct, but we might want to forbid any co_await to give the implementation some leeway. I don't see a reason to adjust what a coroutine waits for while it is waiting on some other thing. You could always have it process a queue of requests before the next co_await sel.

doc/tutorial.md
avikivity: btw a token could well be a pointer into some abstract class, and then you'd just call on a function:

```c++
    while (auto callback = co_await sel) {
        callback->process(); // virtual function call
    }
```

avikivity:
Should deadlines return a future\<bool> so we know if it was cancelled or not?

doc/tutorial.md
avikivity: btw2, the hedge example below gives an example of heterogeneous types, as the the reads and the timeouts have different types.

avikivity:
Alternative API: add() accepts two lambdas, one to launch an async operation, and another that processes the result (accepts a future and returns void). This avoids the switch, but I think results in less-clear code in the caller.

doc/tutorial.md
```diff
@@ -631,6 +631,86 @@ seastar::future<bool> all_exist(std::vector<seastar::sstring> filenames) {
...
+
+A `selection` is declared as a local variable in the coroutine frame. You register operations with `add()`, tagging each with a caller-chosen *token* (an index, an enum, anything copyable). `add()` hands back a `future<T>` for the result. (A deadline is just another future — see [Timeouts](#timeouts) below.) You then repeatedly `co_await` the selection in a `while` loop: each `co_await` suspends the coroutine until one registered event becomes ready and returns its token (as a `std::optional<Token>`, empty once the selection is drained). Use the token to dispatch; the matching future is then guaranteed ready, so `co_await`-ing it (or calling `get()`) yields the result without suspending — a failed operation propagates its exception, just like any future:
```
gleb-cloudius, line 638: Does `selection` has to be a local variable in the coroutine frame? Can it be a variable on a heap?

doc/tutorial.md
```diff
@@ -631,6 +631,86 @@ seastar::future<bool> all_exist(std::vector<seastar::sstring> filenames) {
...
+    while (auto which = co_await sel) {
+        switch (*which) {
+        case ev::deadline:
+            co_await std::move(deadline_f);
+            // primary is taking too long; start a hedged read alongside it
+            hedge = sel.add(ev::hedge, [] { return read_from(1); });
+            break;
+        case ev::primary:
+            if (!answer) { answer = co_await std::move(primary); }
+            deadline.cancel();   // got an answer; drop the deadline so we can drain
```
gleb-cloudius, line 702: Does cancel make `deadline_f` exceptional future? If yes won't this loop throw when primary resolves? If no won't hedge be started when primary resolves?

gleb-cloudius:
> cc @gleb-cloudius for suitability for abstract_read_executor 

I guess you can implement it using this API, though if `sel` must be declared as coroutine variable only it poses inconvenient limitation that the whole logic should sit in a single coroutine. But even without the limitation I am not sure it will simplify the code by much if at all TBH.

doc/tutorial.md
avikivity: Even when it is a local in the coroutine frame, it is on the heap.

You can make it explicitly heap-allocated, but there's no advantage in it.

doc/tutorial.md
avikivity: > Does cancel make `deadline_f` exceptional future?

no

> If yes won't this loop throw when primary resolves? If no won't hedge be started when primary resolves?

Don't know. I think the future should return whether the deadline was cancelled or actually timed out.

```c++
enum class deadline_outcome { timed_out, cancelled };
```

avikivity:
> > cc @gleb-cloudius for suitability for abstract_read_executor
> 
> I guess you can implement it using this API, though if `sel` must be declared as coroutine variable only it poses inconvenient limitation that the whole logic should sit in a single coroutine. But even without the limitation I am not sure it will simplify the code by much if at all TBH.

Really? There's lots and lots of continuations there to massage results and store them into the executor.

These could all be folded into the coroutine.

I suppose I should show the improvement.

doc/tutorial.md
gleb-cloudius: > Even when it is a local in the coroutine frame, it is on the heap.

Sure. But this is special king of heap managed by coroutine infra, and since you specifically mentioned that it should be local variable in coroutine I assumed that there is an importance to it. 

> 
> You can make it explicitly heap-allocated, but there's no advantage in it.

Why? I can have a class that manages all its work through the selection and have a lot of functions that add work to it.

avikivity: > > Even when it is a local in the coroutine frame, it is on the heap.
> 
> Sure. But this is special king of heap managed by coroutine infra, and since you specifically mentioned that it should be local variable in coroutine I assumed that there is an importance to it.

Just performance and implicit lifetime. If you s/selection/unique_ptr\<selection>/, it will work.

> 
> > You can make it explicitly heap-allocated, but there's no advantage in it.
> 
> Why? I can have a class that manages all its work through the selection and have a lot of functions that add work to it.

It's the extra allocation.

There's no problem passing it to a function:

```c++
    coroutine::selection s;
    add_things(s);
    while (auto tok = co_await s) {
        process(*tok, s); // can add things
    }
```

gleb-cloudius:
> > > cc @gleb-cloudius for suitability for abstract_read_executor
> > 
> > 
> > I guess you can implement it using this API, though if `sel` must be declared as coroutine variable only it poses inconvenient limitation that the whole logic should sit in a single coroutine. But even without the limitation I am not sure it will simplify the code by much if at all TBH.
> 
> Really? There's lots and lots of continuations there to massage results and store them into the executor.
> 
> These could all be folded into the coroutine.

And I do not think that huge coroutine with a switch like in your example (that actually has bug, starts hedge when non needed, even though it is simple) and ifs everywhere to check for which condition actually happened will be much simpler. The logic will probably still collect the results into the executor and will make a decision based on its state. In your example you have this as well. You check for `!answer`, in real code you will have something like `executor->has_answer()` or similar.


> 
> I suppose I should show the improvement.

The real complexity is not in future management, but the logic itself. Such api will allow us to consolidate logic into a coroutine with a switch instead of callbacks. If you consider it an improvement that such improvement can be achieved.

avikivity:
> > > > cc @gleb-cloudius for suitability for abstract_read_executor
> > > 
> > > 
> > > I guess you can implement it using this API, though if `sel` must be declared as coroutine variable only it poses inconvenient limitation that the whole logic should sit in a single coroutine. But even without the limitation I am not sure it will simplify the code by much if at all TBH.
> > 
> > 
> > Really? There's lots and lots of continuations there to massage results and store them into the executor.
> > These could all be folded into the coroutine.
> 
> And I do not think that huge coroutine with a switch like in your example (that actually has bug, starts hedge when non needed, even though it is simple) and ifs everywhere to check for which condition actually happened will be much simpler. The logic will probably still collect the results into the executor and will make a decision based on its state. In your example you have this as well. You check for `!answer`, in real code you will have something like `executor->has_answer()` or similar.
> 
> > I suppose I should show the improvement.
> 
> The real complexity is not in future management, but the logic itself. Such api will allow us to consolidate logic into a coroutine with a switch instead of callbacks. If you consider it an improvement that such improvement can be achieved.

I consider the reduction in allocations and tasks an improvement. The code layout can be made similar if wanted.

gleb-cloudius:
> 
> I consider the reduction in allocations and tasks an improvement. The code layout can be made similar if wanted.

OK. Than I misunderstood your question. I thought you asked if readability can be improved.

include/seastar/coroutine/selection.hh
```diff
@@ -0,0 +1,463 @@
...
+/// ```
+///
+/// Call \ref cancel to resolve the future early — e.g. once another event has
+/// made the deadline moot — so the selection can drain without waiting for the
+/// timer to elapse.
+///
+/// Like \ref selection, a \c deadline is non-movable and must outlive any pending
+/// future it handed out.
+template <typename Clock = lowres_clock>
+class deadline {
```
gleb-cloudius, line 433: Why not use `abort_source` + `timer` like we do in other places to express the same idea?

avikivity: abort_source is for long-running operations, this is more lightweight.

It's true abort_source is abused for timeouts, but we shouldn't encourage it.

avikivity:
> > I consider the reduction in allocations and tasks an improvement. The code layout can be made similar if wanted.
> 
> OK. Than I misunderstood your question. I thought you asked if readability can be improved.

Well I think readability can be improved too. But that has to be shown.

include/seastar/coroutine/selection.hh
gleb-cloudius: > abort_source is for long-running operations, this is more lightweight.

You keep saying it, but nobody except you see why. How do you define "long-running"? Is "issuing two abortable reads to two different nodes and canceling one after another completes" a valid use case? If yes, how is it long running comparing to your 100ms example here? 

If you introduce one more primitive that support both timeout and cancel operation then make it usable for more things than just for this API. And notice how this API pollutes any of its users with templates while abort_source does tot require any templates.

> It's true abort_source is abused for timeouts, but we shouldn't encourage it.

Why abused? A timeout is just one thing that may cancel an operation. Why using abort_source to abort a request because node shutdowns is not an abuse, but using it to abort it because the result it not needed any longer for some other reason (timeout) is an abuse?

PR #3552, opened 2026-07-23
build: add lttng-ust to seastar.pc Libs.private

avikivity:
Since commit f98cc0cfdd ("io: Add LTTng-UST tracepoints to IO subsystem"), Seastar is built with Seastar_LTTNG=ON by default and links LTTng::UST as a PRIVATE dependency. The io_trace.cc/reactor_trace.cc translation units reference lttng_ust_* symbols, so a static libseastar.a leaves them unresolved unless the final executable also links lttng-ust.

seastar.pc.in was not updated to match, so its Libs.private omits lttng-ust and consumers linking Seastar statically via pkg-config are under-linked. Add lttng-ust to Libs.private, guarded by Seastar_LTTNG and following the existing Seastar_IO_URING pattern, so it collapses to nothing when the feature is disabled.

### DIFF
--- a/pkgconfig/seastar.pc.in
+++ b/pkgconfig/seastar.pc.in
@@ -27,6 +27,7 @@ lksctp_tools_cflags=-I$<JOIN:@lksctp-tools_INCLUDE_DIRS@, -I>
 lksctp_tools_libs=$<JOIN:@lksctp-tools_LIBRARIES@, >
 liburing_cflags=$<$<BOOL:@Seastar_IO_URING@>:-I$<JOIN:$<TARGET_PROPERTY:URING::uring,INTERFACE_INCLUDE_DIRECTORIES>, -I>>
 liburing_libs=$<$<BOOL:@Seastar_IO_URING@>:$<JOIN:@URING_LIBRARIES@, >>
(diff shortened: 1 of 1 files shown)


nyh:
I just updated Scylla and Seastar, and now can't run configure.py in Scylla:

```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find LTTngUST (missing: LTTngUST_INCLUDE_DIR LTTngUST_LIBRARY)
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindLTTngUST.cmake:43 (find_package_handle_standard_args)
  cmake/SeastarDependencies.cmake:59 (find_package)
  cmake/SeastarDependencies.cmake:111 (seastar_find_dep)
  CMakeLists.txt:422 (seastar_find_dependencies)
```

I don't think anything is actually missing on my system (I installed all the dependencies, I think), I'm guessing something is stale in my Seastar and I'll need to figure out how to fix it. Or something. What a mess :-(

nyh:
Oh, sorry about the spam. it was really a missing dependency - lttng-ust-devel. It's in Seastar's install-dependencies.sh, but not in Scylla's :-(

PR #3553, opened 2026-07-24
reactor: rearm high resolution timer

### LINKED ISSUE #1890: io_uring reactor doesn't work with DPDK
Using the basic seastar::sleep() code from the tutorials:

`int main(int argc, char** argv) { `
`    seastar::app_template app; `
`    app.run(argc, argv, [] { `
`        std::cout << "Sleeping... " << std::flush;`
`        using namespace std::chrono_literals;`
`        return seastar::sleep(1s).then([] {`
`            std::cout << "Done.\n";`
`        });`
`    });`
`}`

 doesn't work when running with "--reactor-backend io_uring ". Future never completes and printing of "Done" does not occur. Using io_uring without enabling DPDK works correctly as well as enabling DPDK with epoll and linuix-aio.

dbekinalkar:
Fixes #1890 

Regression example: 

```c++
   seastar::app_template app;
   app.run(argc, argv, [] {
        std::cout << "Sleeping... " << std::flush;
        return seastar::sleep(1s).then([] {
            std::cout << "Done.\n";
        });
    });
}
```

When Seastar uses the io_uring reactor with DPDK, `seastar::sleep()` may never complete. DPDK keeps the reactor active, so it does not enter the blocking path where the high-resolution timer completion is normally armed.

As a result, the timerfd expiration is not observed and the application hangs after printing `Sleeping...`.

By rearming the high-resolution timer completion from `kernel_submit_work()` before submitting pending `io_uring` operations, we ensure that timer expirations are monitored during active processing as well as while the reactor is idle.

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -1743,6 +1743,7 @@ public:
         bool did_work = false;
         did_work |= _preempt_io_context.service_preempting_io();
         did_work |= queue_pending_file_io();
+        _hrtimer_completion.maybe_rearm(*this);
         did_work |= ::io_uring_submit(&_uring);
         return did_work;
     }
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -1753,7 +1753,6 @@ public:
     }
     virtual void wait_and_process_events(const sigset_t* active_sigmask) override {
         _smp_wakeup_completion.maybe_rearm(*this);
-        _hrtimer_completion.maybe_rearm(*this);
         ::io_uring_submit(&_uring);
         bool did_work = false;
         did_work |= _preempt_io_context.service_preempting_io();
--- a/tests/unit/reactor_backend_test.cc
+++ b/tests/unit/reactor_backend_test.cc
@@ -21,17 +21,11 @@
 
 #include <seastar/core/coroutine.hh>
 #include <seastar/core/future.hh>
-#include <seastar/core/reactor.hh>
-#include <seastar/core/sleep.hh>
 #include <seastar/core/posix.hh>
-#include <seastar/core/internal/poll.hh>
 #include <seastar/core/internal/pollable_fd.hh>
 #include <seastar/testing/test_case.hh>
 
-#include <chrono>
-
 using namespace seastar;
-using namespace std::chrono_literals;
 
 // After a co_await on a pollable_fd future resolves, await_resume() calls
 // future::get() which calls get_available_state_ref(), which sees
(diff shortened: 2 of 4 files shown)


xemul:
This bug can presumably be reproduced using --poll-mode option. Am I right? If so, a regression test would be welcome

src/core/reactor_backend.cc
xemul: Presumably, this line is now redundant, isn't it?

dbekinalkar: Yep, removed

dbekinalkar:
> This bug can presumably be reproduced using --poll-mode option. Am I right? If so, a regression test would be welcome

I attempted to add a `--poll-mode` based test, but it does not reproduce the bug locally. Instead, this test uses an explicit active poller to model DPDK registered poller, which prevents the reactor from entering interrupt mode.

xemul:
> I attempted to add a --poll-mode based test, but it does not reproduce the bug locally

Why? It seems that it should, reactor doesn't go to sleep, timer doesn't get re-armed

tests/unit/reactor_backend_test.cc
```diff
@@ -63,3 +69,13 @@ SEASTAR_TEST_CASE(pollable_fd_state_completion_reuse_test) {
     ::close(sv[1]);
     co_return;
 }
+
+// Like DPDK, this poller prevents the reactor from entering interrupt mode.
+// The high-resolution timer must therefore be serviced while polling.
+SEASTAR_TEST_CASE(highres_timer_with_active_poller_test) {
```
xemul: I'm not sure this is enough. Tests are run with "random" (whatever it means) reactor beckend, but the goal is to explicitly use uring one. You might want to create a dedicated .cc file with this test and explicitly mark it with the corresponding --reactor-backend option in cmakefile.txt

dbekinalkar: Created a new file with the `--reactor-backend io_uring` option

dbekinalkar:
> Why? It seems that it should, reactor doesn't go to sleep, timer doesn't get re-armed

Right, the reactor doesn't go to sleep. The error is in the failed interrupt mode setup/teardown path. With an active poller, the reactor still tries to enter interrupt mode through `reactor::pollers_enter_interrupt_mode() -> lowres_timer_pollfn::try_enter_interrupt_mode()`.

`lowres_timer_pollfn::try_enter_interrupt_mode()` arms `_nearest_wakeup`, which reprograms the shared hrtimer fd. Then the active poller rejects interrupt mode via through `simple_pollfn<false>::try_enter_interrupt_mode()`. That makes `pollers_enter_interrupt_mode()` unwind earlier pollers through the teardown `lowres_timer_pollfn::exit_interrupt_mode() -> _nearest_wakeup.cancel()`.

So the temporary timer is removed, but the hrtimer fd is not restored to the original sleep deadline. Poll mode skips `pollers_enter_interrupt_mode()` entirely, so it cannot reproduce it so simply. I can switch this to the poll-mode with a callback rearm if you prefer, but the active poller version matches the DPDK path more directly.

dbekinalkar:
@xemul It looks like the CI is failing due to io_uring backend option not being available on all CI platforms. I removed the forced backend for now. What would be the preferred way to handle this?

xemul:
@dbekinalkar , eeb1a469 should have solved it, so please rebase and restore the backend selection, it should now work

Issue #116, opened 2016-02-22
Cannot build with -DNDEBUG


unknownzerx:
If I compile with the flag `DNDEBUG`, it will complain
```
/root/seastar/core/future.hh:953:13: error: control reaches end of non-void function [-Werror=return-type]
```
near the `assert`
```c++
future<T...> finally(Func&& func) noexcept {
    // ...
    if (!f_res.failed()) {
        return std::move(result);
    } else {
        if (!result.failed()) {
            return make_exception_future<T...>(f_res.get_exception());
        } else {
            // ...
            assert(0 && "we should not be here");
        }
    }
    // ...
}
```
(-Wall and -Werr are defined in SEASTAR_CFLAGS)

The solution could be return or throw some dummy object.

nyh:
You are right: By defining the NDEBUG macro, all assert()s can be compiled away, so we must write all our assertions with this in mind: assertions must never have side-effects (e.g., never do "assert(do_something() == 0)") and as you noted, assert(0) **cannot** be assumed to never return. 

When we really want to abort in case of a bug, I think we should use **abort()**, not assert(0), nor throw or return a dummy object. This is not only place in the code which uses assert(0), and they should probably all be fixed, but you're right that the place you found in the header file is the most important, because application code uses this header file, and this application code might choose to compile with NDEBUG, even if Seastar itself was compiled without this flag.

xemul:
Seems to build fine now (see https://github.com/scylladb/seastar/pull/3554#issuecomment-5103560585)

PR #3554, opened 2026-07-24
tests: Fix NDEBUG build

StephanDollberg:
When building with NDEBUG assert becomes nothing and clang-23 complains about unused globals.

```
/build/seastar/tests/perf/perf_tests_perf.cc:142:12: error: variable 'hook_1_count' set but not used [-Werror,-Wunused-but-set-global]
  142 | static int hook_1_count = 0, hook_2_count = 1;
      |            ^
/build/seastar/tests/perf/perf_tests_perf.cc:142:30: error: variable 'hook_2_count' set but not used [-Werror,-Wunused-but-set-global]
  142 | static int hook_1_count = 0, hook_2_count = 1;
```

Mark them as maybe_unused to resolve this issue.

### DIFF
--- a/tests/perf/perf_tests_perf.cc
+++ b/tests/perf/perf_tests_perf.cc
@@ -139,7 +139,7 @@ PERF_TEST(overhead_check, high_overhead_test) {
 
 // The following tests run in order check that pre-run hooks are executed properly.
 
-static int hook_1_count = 0, hook_2_count = 1;
+[[maybe_unused]] static int hook_1_count = 0, hook_2_count = 1;
 
 PERF_PRE_RUN_HOOK([](const std::string& g, const std::string& c) {
     ++hook_1_count;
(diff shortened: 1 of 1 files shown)


StephanDollberg:
@travisdowns

xemul:
Fixes https://github.com/scylladb/seastar/issues/116 ?

StephanDollberg:
> Fixes https://github.com/scylladb/seastar/issues/116 ?

Hah, I feel like that issue is very stale :). It seems to compile fine in general with NDEBUG already.

xemul:
> Hah, I feel like that issue is very stale :). 

It is, still we couldn't just close it without checking :)

> It seems to compile fine in general with NDEBUG already.

\+ this PR

travisdowns:
FWIW we already build with NDEBUG in release so effectively we are always testing ndebug. What changed I guess here is clang-23.

PR #3556, opened 2026-07-26
core: io_queue: Fix low performance with io_scheduler and --overprovisioned (Fixes #3202)

### LINKED ISSUE #3202: Low performance (high effective IO latency) with io_scheduler and no/short polling or --overprovisioned
## Summary

When using `--overprovisioned` or `--idle-poll-time-us=1` and a using non-default io-properties appropriate for the host, single-shard I/O performance drops to ~2.7K IOPS compared to ~12K IOPS with baseline configuration (no `--overprovisioned`). This represents a roughly 4-5x performance reduction. The issue appears to be caused by the reactor sleeping longer than "necessary" when waiting for token bucket capacity: specifically because the per grab amount is large, and we wait for the full amount.

## Environment

- **Seastar version**: Current master (commit bbd0001a)
- **Test workload**: Random 4KB reads (or writes, behavior is the same), iodepth=1
- **Storage**: NVMe SSD on XFS
- **Disk capabilities** (measured by iotune):
  - Random read IOPS: 268,337 IOPS

## io-properties Configuration

```yaml
disks:
  - mountpoint: /mnt/xfs
    read_iops: 268337
    read_bandwidth: 1259085440
    write_iops: 134175
    write_bandwidth: 604742528
```

### Baseline (no --overprovisioned)
```bash
$ build/release/apps/io_tester/io_tester --io-properties-file ~/io-props.yaml \
    --conf ~/io2.yaml --storage /mnt/xfs/io_tester --duration=5 -c1
Job highprio -> sched class highprio
    IOPS: 12126.1523
```

Result: ~12,126 IOPS

### With --overprovisioned
```bash
$ build/release/apps/io_tester/io_tester --io-properties-file ~/io-props.yaml \
    --conf ~/io2.yaml --storage /mnt/xfs/io_tester --duration=5 -c1 --overprovisioned
Job highprio -> sched class highprio
    IOPS: 2676.50513
```

Result: ~2,677 IOPS (about 4.5x slower)

### With --idle-poll-time-us=1 (same behavior)
```bash
$ build/release/apps/io_tester/io_tester --io-properties-file ~/io-props.yaml \
    --conf ~/io2.yaml --storage /mnt/xfs/io_tester --duration=5 -c1 --idle-poll-time-us=1
Job highprio -> sched class highprio
    IOPS: 2670.78101
```

Result: ~2,671 IOPS (same degradation as --overprovisioned, confirming the issue is caused by reduced polling frequency)

### Without io-properties (control)
```bash
$ build/release/apps/io_tester/io_tester \
    --conf ~/io2.yaml --storage /mnt/xfs/io_tester --duration=5 -c1 --overprovisioned
Job highprio -> sched class highprio
    IOPS: 12289.8516
```

Result: ~12,290 IOPS (similar to baseline, no degradation without io-properties)

## Root Cause Analysis

The issue appears to be caused by two factors:

### Issue 1: Sleep Time Based on Full Deficiency, Not I/O Need

The sleep duration is calculated based on replenishing the full token deficiency (the gap between pending reservation and bucket head), not just the tokens needed for the next I/O.

From `ioinfo -c1 --directory /mnt/xfs`:
```
fair_queue:
  capacities:
    4096:
      read: 117101    # tokens needed for 4KB read
  per_tick_grab_threshold: 12582912
  token_bucket:
    rate: 16777216    # tokens per millisecond
```

For a 4KB read:
- Tokens needed: 117,101 tokens
- Time to replenish 1 I/O: ~7 us

But the token bucket reserves in large chunks:
- per_tick_grab_threshold: 12,582,912 tokens
- Time to replenish full reservation: ~750 us

This means the system may sleep considerably longer than needed to dispatch a single I/O.

### Issue 2: Tokens Only Added During Polling

Token bucket replenishment is not passive — tokens are only added when `maybe_replenish_capacity()` is called, which only happens from `poll_io_queue()`.

1. With `--overprovisioned`, `max_poll_time=0us` causes the reactor to sleep immediately when idle
2. While sleeping, no polling occurs, so no tokens are added to the bucket
3. The reactor sleeps for the full calculated deficiency time

### The Flow

1. I/O completes at time T=0
2. io-queue checks for next dispatch — not enough tokens available
3. `next_pending_aio()` calculates delay based on the full deficiency, which is typically close to `per_tick_grab_threshold` (12.5M tokens, ~750us) since that's the reservation size used in `grab_capacity()`
4. Timer armed for T+delay, reactor goes to sleep
5. Reactor sleeps for full delay (even though only ~7us worth of tokens are needed for the next I/O)
6. Timer fires, reactor wakes
7. `poll_io_queue()` called, which calls `maybe_replenish_capacity()`
8. Tokens are now added (for the full elapsed time)
9. Next I/O dispatches

The result is that I/O dispatch frequency is limited by the sleep duration rather than the actual token replenishment time needed.

### Why Baseline Works

With baseline configuration:
- `max_poll_time=200us` means the reactor actively polls before sleeping
- `poll_io_queue()` is called frequently
- `maybe_replenish_capacity()` runs every few microseconds
- Tokens appear in the bucket almost as soon as time passes

With `--overprovisioned`:
- `max_poll_time=0us` means no active polling
- Tokens only appear when the reactor wakes from sleep
- Sleep duration becomes the limiting factor for throughput

### Evidence from Logging

**Baseline replenishment pattern:**
```
io_throttler::maybe_replenish: elapsed=0us, extra=10335, REPLENISHING
io_throttler::maybe_replenish: elapsed=0us, extra=7650, REPLENISHING
io_throttler::maybe_replenish: elapsed=1us, extra=26055, REPLENISHING
```
Tokens replenished every 0-1 microseconds.

**Overprovisioned replenishment pattern:**
```
io_throttler::maybe_replenish: elapsed=543us, extra=9124541, REPLENISHING
io_throttler::maybe_replenish: elapsed=431us, extra=7236634, REPLENISHING
io_throttler::maybe_replenish: elapsed=589us, extra=9894866, REPLENISHING
```
Tokens only replenished every 400-600 microseconds (when reactor wakes).

## Proposed Solutions

Some of the solutions proposed for the multi-shard io-properties issue (see #3201) may also apply here, particularly those related to passive token replenishment or changes to how sleep duration is calculated.

## Impact

This issue affects workloads that:
- Use `--overprovisioned` mode (common in containerized/virtualized environments)
- Have io-properties configured for disk throttling
- Have low iodepth where token bucket throttling dominates

The performance reduction may make `--overprovisioned` less suitable for use with io-properties in latency-sensitive scenarios.

## Reproducibility

100% reproducible with the test configurations provided.

### Steps to Reproduce

1. Create io-properties file (`~/io-props.yaml`):
```yaml
disks:
  - mountpoint: /mnt/xfs
    read_iops: 268337
    read_bandwidth: 1259085440
    write_iops: 134175
    write_bandwidth: 604742528
```

2. Create test config (`~/io2.yaml`):
```yaml
- name: highprio
  shards: [0]
  type: randread
  shard_info:
    parallelism: 1
    reqsize: 4kB
    shares: 1000
    think_time: 0
```

3. Build Seastar:
```bash
./configure.py --mode=release
ninja -C build/release apps/io_tester/io_tester
```

4. Run baseline test:
```bash
build/release/apps/io_tester/io_tester \
    --io-properties-file ~/io-props.yaml \
    --conf ~/io2.yaml \
    --storage /mnt/xfs/io_tester \
    --duration=5 -c1
```

5. Run overprovisioned test:
```bash
build/release/apps/io_tester/io_tester \
    --io-properties-file ~/io-props.yaml \
    --conf ~/io2.yaml \
    --storage /mnt/xfs/io_tester \
    --duration=5 -c1 --overprovisioned
```

6. Compare results: overprovisioned should show noticeably lower IOPS.

## Related Issues

This issue is related to but distinct from the multi-shard io-properties issue (#3201). Both issues stem from the token bucket design in the sense they are related to the grab granularity but have different root causes:

- **Multi-shard issue**: Token loss when `ready_tokens` are discarded on empty queue
- **Overprovisioned issue**: Tokens not replenished without active polling

NITIN9181:
Fixes #3202

## Description
When using `--overprovisioned` (or `--idle-poll-time-us=1`), the token bucket wait duration in `next_pending_aio` was calculated based on replenishing the entire reservation chunk (`_pending.cap`), which is usually ~12.5M tokens. This caused the reactor to sleep for ~750µs between each I/O dispatch, throttling throughput down to ~2.7K IOPS.

This PR adjusts the `over` (token deficiency) calculation by checking the exact capacity needed for the next request (`fq.top()->capacity()`). If the required capacity is smaller than the pending reservation, we reduce the wait time accordingly. This allows the reactor to sleep for the precise time needed by the next request (e.g. ~7µs for a 4KB read), completely restoring the throughput.

### DIFF
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -1305,6 +1305,19 @@ io_queue::clock_type::time_point io_queue::stream::next_pending_aio() const noex
          * works faster, than it really does.
          */
         auto over = out.capacity_deficiency(_pending.head);
+
+        auto* ent = fq.top();
+        if (ent != nullptr) {
+            auto cap = ent->capacity();
+            if (cap <= _pending.cap) {
+                if (over > _pending.cap - cap) {
+                    over -= (_pending.cap - cap);
+                } else {
+                    over = 0;
+                }
+            }
+        }
+
         auto ticks = out.capacity_duration(over);
         return std::chrono::steady_clock::now() + std::chrono::duration_cast<std::chrono::microseconds>(ticks);
     }
--- a/src/core/io_queue.cc
+++ b/src/core/io_queue.cc
@@ -1306,7 +1306,7 @@ io_queue::clock_type::time_point io_queue::stream::next_pending_aio() const noex
          */
         auto over = out.capacity_deficiency(_pending.head);
 
-        auto* ent = fq.top();
+        auto* ent = const_cast<fair_queue&>(fq).top();
         if (ent != nullptr) {
             auto cap = ent->capacity();
             if (cap <= _pending.cap) {


xemul:
How is it connected with --overprovisioned? The tokens reservation/claiming mechanics doesn't depend on it

src/core/io_queue.cc
```diff
@@ -1305,6 +1305,19 @@ io_queue::clock_type::time_point io_queue::stream::next_pending_aio() const noex
          * works faster, than it really does.
          */
         auto over = out.capacity_deficiency(_pending.head);
+
+        auto* ent = fq.top();
+        if (ent != nullptr) {
```
xemul, line 1310: _pending.cap being non-zero already implies that there's at least one request in the queue

NITIN9181:
You're totally right that the reservation mechanics themselves don't depend on --overprovisioned. The connection is strictly about how --overprovisioned changes the reactor's polling behavior, which dictates when tokens are actually replenished.

Tokens aren't added passively by a timer; they only get replenished when maybe_replenish_capacity() is called, which happens inside poll_io_queue().

Normally (without --overprovisioned), the reactor has a non-zero idle-poll-time-us, meaning it actively spins when idle. During this time, it constantly calls poll_io_queue(), so tokens are replenished in the background as time passes. Because the reactor is already awake and polling, this effectively hides the fact that next_pending_aio() was calculating a sleep timer that was way too long.

But when you use --overprovisioned, idle-poll-time-us drops to 0. The reactor stops spinning and goes straight to sleep via epoll_wait. While it's sleeping, no tokens are added. The reactor relies entirely on the timer from next_pending_aio() to wake up. Since that timer was calculating the wait for the entire reservation chunk (e.g. ~12.5M tokens, taking ~750µs) instead of just what is needed for the next request (~117K tokens, taking ~7µs), the reactor over-sleeps heavily. It wakes up, dispatches a single request, and goes straight back to sleep for another ~750µs, tanking the IOPS.

By adjusting the timer in next_pending_aio(), the reactor now wakes up exactly when the next request is actually ready to go.

src/core/io_queue.cc
NITIN9181: Actually, _pending.cap being non-zero does not guarantee there is a request in the queue.

Because reservations are made in large chunks (per_tick_grab_threshold(), usually ~12.5M tokens), the _pending.cap can remain non-zero if the queue empties out before the full reservation is completely generated.

For example:

We make a 12.5M token reservation.
The bucket generates enough tokens to satisfy the only pending request (say, a 4KB read needing 117K tokens).
poll_io_queue() runs, reaps those tokens, and dispatches the request.
The queue is now empty (fq.top() == nullptr), but _pending.cap still holds the remainder of the 12.5M reservation because deficiency > 0.
Additionally, if the only request in the queue gets cancelled via cancel_request(), it's removed from the fair_queue but the stream's _pending.cap reservation isn't cleared.

In both of these perfectly normal scenarios, _pending.cap is non-zero but fq.top() is nullptr. So the ent != nullptr check is strictly necessary to prevent a segfault.

xemul: > We make a 12.5M token reservation.

It means, the queued requests are at least 12.5M in sum. Reservation is capped by fq.queued_capacity(). How exactly did you achieve that state?

xemul:
I see. But anyway, I'd like to ask you to fix the PR (and patch) title. Even when not --overprovisioned, reactor may go to sleep and rely on this value wake it up in timely manner, and in case [this happens](https://github.com/scylladb/seastar/pull/3556#discussion_r3665350009) it will be delayed for too long.

src/core/io_queue.cc
```diff
@@ -1306,7 +1306,7 @@ io_queue::clock_type::time_point io_queue::stream::next_pending_aio() const noex
          */
         auto over = out.capacity_deficiency(_pending.head);
 
-        auto* ent = fq.top();
+        auto* ent = const_cast<fair_queue&>(fq).top();
```
xemul, line 1309: Huh? The ent is effectively const here, why casting it to non-const ref?

travisdowns:
> How is it connected with --overprovisioned? The tokens reservation/claiming mechanics doesn't depend on it

`--overprovisioned` disables polling, so the reactor goes to sleep, and the sleep time is calculated based on the full reservation amount (12.5M in this case) which causes the issue. Without `--overprovisioned` polling happens frequently and we replenish in a fine-grained way. Details are in #3202.

PR #3555, opened 2026-07-24
future: Enable HALO

StephanDollberg:
Now that https://github.com/llvm/llvm-project/issues/188230 is fixed we can enable the LLVM HALO attributes in seastar.

This patch series consists of the following patches:

 - Extend the coroutine / continuation microbenches in seastar with a set of microbenches we had been using in Redpanda
 - Introduce macros for the llvm HALO related attributes
 - Mark `future` with `clang::coro_await_elidable`
 - Make some `future` related methods better inlinable
 - Mark `when_all` arguments with `clang::coro_await_elidable_argument`

The main benefit of HALO is that it allows eliding nested coroutine frame allocations which are fairly common in coroutine rich code.

Microbenchmark deltas from the HALO commit:

```
Test                                       inst(d)            allocs(d)
---------------------------------------------------------------------------------
coroutine_test.buffered_generator            +0.17  (+0.0%)       +0.00  (+0.0%)
coroutine_test.cont_chain_then5               0.00  (-0.0%)        0.00  (-1.6%)
coroutine_test.cont_chained_after_yield      +0.01  (+0.0%)       +0.00  (+0.0%)
coroutine_test.cont_empty                     0.00  (-0.0%)        0.00 (-17.2%)
coroutine_test.cont_nested_after_yield       -0.01  (-0.0%)        0.00  (-0.0%)
coroutine_test.cont_nested_then5             +0.00  (+0.0%)        0.00  (-3.7%)
coroutine_test.cont_ss_now                    0.00  (-0.0%)        0.00  (-9.0%)
coroutine_test.cont_ss_now_collect            0.00  (-0.0%)       +0.00  (+0.0%)
coroutine_test.coro_after_yield               0.00  (-0.0%)        0.00  (-0.0%)
coroutine_test.coro_empty                     0.00  (-0.0%)        0.00  (-0.0%)
coroutine_test.coro_never_awaits_collect     +0.00  (+0.0%)       +0.00  (+0.0%)
coroutine_test.coro_per_field_1               0.00  (-0.0%)        0.00  (-0.3%)
coroutine_test.coro_per_field_10             +0.21  (+0.0%)        0.00  (-4.9%)
coroutine_test.coro_per_field_40             +0.04  (+0.0%)       +0.00  (+0.1%)
coroutine_test.coro_ready                    +0.00  (+0.0%)        0.00  (-0.0%)
coroutine_test.coro_ready_collect            -0.03  (-0.0%)       +0.00  (+0.0%)
[...]

PR #3557, opened 2026-07-27
coroutine: Mark more methods with clang::coro_await_elidable_argument

StephanDollberg:
In our previous patch series (https://github.com/scylladb/seastar/pull/3555)
which introduced HALO (heap-allocation-elision-optimization) originally we left
some coroutine wrappers untouched that could also benefit from
`clang::coro_await_elidable_argument`. 

This was because:

> Note that try_future would also be good candidate to mark but
> currently the argument doesn't work on constructor arguments so it's not
> applicable there.

It's easy to work around these issues by simply using factory functions instead
of using the type+constructor way.

This is technically an API break in case anybody is having weird usecases like:

```
coroutine::as_future<int> wrapper(foo());
co_await wrapper;
```

but this isn't really how anybody would use this so I don't think it's worth an
API version bump. I can't see any such uses in either Redpanda or ScyllaDb

The normal function style call-site stays in exactly the same
form.

Resulting perf improvements. 

```
Test                                                   inst(d)            allocs(d)        
-------------------------------------------------------------------------------------------
coroutine_test.without_preemption_check_ready_chain    -95.02 (-21.8%)       -1.00 (-50.0%)
coroutine_test.wrapped_ready_chain                     -93.02 (-19.0%)       -1.00 (-50.0%)
coroutine_test.try_future_ready_chain                  -93.02 (-21.0%)       -1.00 (-50.0%)
```

### DIFF
--- a/include/seastar/core/coroutine.hh
+++ b/include/seastar/core/coroutine.hh
@@ -244,6 +244,11 @@ public:
     void await_resume() { _future.get(); }
 };
 
+template<typename T>
+struct [[nodiscard]] SEASTAR_CORO_AWAIT_ELIDABLE without_preemption_check : public seastar::future<T> {
+    explicit without_preemption_check(seastar::future<T>&& f) noexcept : seastar::future<T>(std::move(f)) {}
+};
+
 } // seastar::internal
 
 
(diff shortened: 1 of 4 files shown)


avikivity:
But what about

```c++
future<>
g(future<> f) {
    co_await coroutine::as_future(std::move(f));
}
```

It is exceedingly unlikely someone would write code like that, I support we can just document it.

avikivity:
> But what about
> 
> ```c++
> future<>
> g(future<> f) {
>     co_await coroutine::as_future(std::move(f));
> }
> ```
> 
> It is exceedingly unlikely someone would write code like that, I support we can just document it.

Ah, the elision machinery would just ignore it because there is no opportunity for putting one coroutine frame inside the other.

PR #3558, opened 2026-07-28
rpc: Call seastar::format explicitly

StephanDollberg:
In similar vein to 52b26333c0a4a307a291d4ad22e9135f34455dd2 make explicit format calls to avoid ambiguity between std:: and fmt::.


StephanDollberg:
@travisdowns

PR #3559, opened 2026-07-28
Fix various deprecation warnings

StephanDollberg:
The build is currently quite noisy with deprecation from previous commits that marked various things as deprecated. 

 - Fmt12 deprecated APIs
 - Deprecated exception_ptr formatter
 - Deprecated consumption_result constructor
 - Deprecated input_stream move-assignment.

### DIFF
--- a/include/seastar/util/log.hh
+++ b/include/seastar/util/log.hh
@@ -268,7 +268,11 @@ public:
                 });
                 do_log(level, writer);
             } catch (...) {
+#if defined(SEASTAR_LOGGER_COMPILE_TIME_FMT) && FMT_VERSION >= 100000
+                failed_to_log(std::current_exception(), fmt.format.get(), fmt.loc);
+#else
                 failed_to_log(std::current_exception(), fmt::string_view(fmt.format), fmt.loc);
+#endif
             }
         }
     }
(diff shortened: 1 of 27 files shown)


include/seastar/util/log.hh
```diff
@@ -295,11 +299,19 @@ public:
...
                     return fmt::format_to(it, fmt::runtime(fmt.format), std::forward<Args>(args)...);
+#endif
                 });
                 do_log(level, writer);
             } catch (...) {
+#if defined(SEASTAR_LOGGER_COMPILE_TIME_FMT) && FMT_VERSION >= 100000
+                failed_to_log(std::current_exception(), fmt.format.get(), fmt.loc);
+#else
                 failed_to_log(std::current_exception(), fmt::string_view(fmt.format), fmt.loc);
+#endif
```
travisdowns, line 314: better to define a helper for this, e.g., which takes a format and returns a string_view or whatever

include/seastar/util/log.hh
```diff
@@ -268,7 +268,11 @@ public:
                 });
                 do_log(level, writer);
             } catch (...) {
+#if defined(SEASTAR_LOGGER_COMPILE_TIME_FMT) && FMT_VERSION >= 100000
```
travisdowns, line 271: why is `defined(SEASTAR_LOGGER_COMPILE_TIME_FMT)` here?

StephanDollberg: Because if that is off you still need to use the string_view

include/seastar/util/log.hh
StephanDollberg: I don't fully understand? Helper for what? 

Also I don't think this format stuff needs more helpers.

travisdowns (commented):
not really in favor for [treewide: wrap exception_ptr before formatting](https://github.com/scylladb/seastar/pull/3559/changes/2fb382fdbd1242429851f1b8fafbfa2b9627fc55) specifically, let me poke around a bit

include/seastar/util/log.hh
travisdowns: well yeah but the question was why

include/seastar/util/log.hh
travisdowns: this same pattern is repeated 3x with this mess:

```
#if defined(SEASTAR_LOGGER_COMPILE_TIME_FMT) && FMT_VERSION >= 100000
             foo(..., fmt.format.get(), ...);
#else
             foo(..., fmt::string_view(fmt.format), ...);
#endif
```

so just a helper like:

```
auto defail::format_string(...) {
#if defined(SEASTAR_LOGGER_COMPILE_TIME_FMT) && FMT_VERSION >= 100000
             return ...
#else
             return ...
#endif
}
```

which restores all the rest to 1 line and also this helper should explain what is going on, when it can be removed, etc.

PR #3563, opened 2026-07-28
tls: throw GnuTLS errors with the tls::error_category() instance

### LINKED ISSUE #3562: tls: GnuTLS errors are thrown with a different error_category instance than tls::error_category() returns
Since 46b57ae0f ("net/tls: dispatch tls::error_category() through crypto_provider"), errors thrown by the GnuTLS backend carry a different `std::error_category` instance than the one `tls::error_category()` returns, so identity comparisons against the public category no longer match.

### Details

All throw sites in `src/net/tls_gnutls.cc` use a file-local accessor with its own function-local static:

```cpp
// src/net/tls_gnutls.cc:120
static const gnutls_error_category& local_error_category() {
    static const gnutls_error_category ec;
    return ec;
}
```

while the public `tls::error_category()` dispatches through the crypto provider to `tls::gnutls::error_category()`, which creates a *second* static instance of the same class:

```cpp
// src/net/tls_gnutls.cc:1388
const std::error_category& tls::gnutls::error_category() {
    static const gnutls_error_category ec;
    return ec;
}
```

`std::error_category::operator==` compares object identity, so for every error thrown by the GnuTLS backend:

```cpp
try {
    // any failing TLS operation
    assert(e.code().category() == tls::error_category()); // fails since 46b57ae0f
}
```

Before 46b57ae0f the throw sites and `tls::error_category()` shared a single instance, so this comparison held.

### Impact

Client code that identifies TLS errors by comparing categories silently stops matching. For example ScyllaDB does exactly this in `transport/generic_server.cc` (`is_broken_pipe_or_connection_reset`) to classify `ERROR_PREMATURE_TERMINATION` / `ERROR_PUSH` / `ERROR_PULL` and suppress log noise for ordinary disconnects; the category gate now fails before the value comparison is ever reached.

The OpenSSL backend is not affected: its throw sites and `tls::openssl::error_category()` both use the single `openssl_error_cat()` static.

### Fix

Minimal fix is to make the GnuTLS throw sites use the same instance the public accessor returns, e.g. have `local_error_category()` forward to `tls::gnutls::error_category()`.

Longer term, the backend-neutral error code rework discussed in #3400 (translating backend errors to fixed neutral values in a single category) removes this class of bug by construction.

travisdowns:
Since 46b57ae0f ("net/tls: dispatch tls::error_category() through crypto_provider"), the GnuTLS backend throws its errors with a file-local static `gnutls_error_category` instance, while `tls::error_category()` resolves (via the crypto provider) to `tls::gnutls::error_category()`, which owns a second static instance of the same class. `std::error_category` compares by object identity, so

```cpp
e.code().category() == tls::error_category()
```

is false for every error thrown by the GnuTLS backend, silently breaking client code that identifies TLS errors this way (for example ScyllaDB's `is_broken_pipe_or_connection_reset` in `transport/generic_server.cc`). The OpenSSL backend is unaffected: its throw sites and accessor share a single static.

Fix by forwarding the file-local accessor to `tls::gnutls::error_category()`, and add a regression test that fails against the previous code (verified: 2 failures without the fix, green with it; full tls_test suite passes).

Fixes #3562

### DIFF
--- a/src/net/tls_gnutls.cc
+++ b/src/net/tls_gnutls.cc
@@ -117,9 +117,11 @@ public:
 };
 
 
-static const gnutls_error_category& local_error_category() {
-    static const gnutls_error_category ec;
-    return ec;
+// Must return the same instance that tls::error_category() resolves to
+// (via the crypto provider, tls::gnutls::error_category()), so that errors
+// thrown here compare equal to the public category by identity.
+static const std::error_category& local_error_category() {
+    return tls::gnutls::error_category();
 }
 
 // Checks a gnutls return value.
--- a/tests/unit/tls_test.cc
+++ b/tests/unit/tls_test.cc
@@ -458,6 +458,21 @@ SEASTAR_TEST_CASE(test_non_tls) {
     });
 }
 
+SEASTAR_THREAD_TEST_CASE(test_error_category_instance) {
+    // Errors thrown by the TLS layer must carry the exact category instance
+    // that tls::error_category() returns: std::error_category compares by
+    // object identity, so a second instance of the same category class never
+    // compares equal (issue #3562).
+    tls::credentials_builder b;
+    b.set_x509_key(tls::blob("garbage"), tls::blob("garbage"), tls::x509_crt_format::PEM);
+    try {
+        b.build_certificate_credentials();
+        BOOST_FAIL("expected an exception parsing garbage certificates");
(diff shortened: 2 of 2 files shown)


travisdowns:
FYI @scylladb/seastar-maint , may affect scylladb.

elcallio (approved):

PR #3564, opened 2026-07-28
tls: give the ERROR_* globals fixed backend-neutral values

### LINKED ISSUE #3562: tls: GnuTLS errors are thrown with a different error_category instance than tls::error_category() returns

travisdowns:
The exported `tls::ERROR_*` globals hold raw backend error values (GnuTLS `E_` codes or OpenSSL `ERR_PACK()`ed codes, depending on the crypto provider selected at startup). They are zero-initialized and only filled in by the active backend's `init_error_codes()` at reactor startup, so any read before that (static initializers, unit tests that never start a reactor) silently yields 0, and their values differ between backends.

Instead, give each exported constant a fixed, backend-neutral value and translate backend errors to these values at the point the error is created:

- The values start at 0x10001, chosen to be distinguishable from anything else that can appear in a value comparison against them: above the errno range, positive (raw GnuTLS errors are negative) and below 2^23 (where untranslated OpenSSL `ERR_PACK()` values start).
- `tls::error_category()` becomes a single backend-neutral category ("tls"), statically initialized and valid at any time. Its `message()` describes the exported constants itself and delegates to the active backend for raw values.
- Backend errors with no exported constant keep their backend-specific value in the same category, so nothing loses fidelity.
- The `init_error_codes()` machinery is removed entirely.

This is an ABI break (the constants change value) but arguably not an API break: code comparing TLS error codes against the exported constants keeps working and now behaves identically under either backend, per the discussion with @avikivity in https://github.com/scylladb/seastar/pull/3400#discussion_r3294418704.

This also fixes #3562 by construction (all errors now carry the one neutral category instance); the minimal standalone fix proposed as #3563 is subsumed by this change.

Also makes the http retry_strategy check the error category before comparing against `ERROR_PREMATURE_TERMINATION`; with the chosen value range this is hardening rather than a bug fix.

Tests cover the constants (range, uniqueness, validity from static initializers, category identity and messages), end-to-end translation of mapped errors (premature termination on abrupt transport EOF; decryption/MAC failure on a corrupted record), raw pass-through of unmapped errors and the `message()` delegation for them. Full `tls_test` suite validated locally under the GnuTLS provider.

Independent of #3400 (either can merge first; the other takes a small rebase).

Fixes #3562

### DIFF
--- a/include/seastar/net/tls.hh
+++ b/include/seastar/net/tls.hh
@@ -53,6 +53,14 @@ class socket_address;
  * agnostic, so in theory it could be replaced
  * with OpenSSL or similar.
  *
+ * Errors are reported as \c std::system_error exceptions carrying
+ * \ref error_category. The \c ERROR_* globals below are backend-neutral
+ * error values: backend errors with an exported constant are translated to
+ * it when the error is created, so comparisons against these constants
+ * behave the same under either backend.
+ *
+ * The \c ERROR_* globals and \ref error_category are backend-neutral,
+ * statically initialized, and valid at any time in every build mode.
  */
 namespace tls {
 
@@ -662,7 +670,13 @@ namespace tls {
     /**
      * Error handling.
      *
-     * The error_category instance used by exceptions thrown by TLS
+     * The error_category instance used by exceptions thrown by TLS.
+     *
+     * This category is backend-neutral: it is the same instance no matter
+     * which backend is active, and it is valid at any time in every build
+     * mode. Error values in this category are either one of the \c ERROR_*
+     * constants below, or a backend-specific value for errors that have no
+     * exported constant.
      */
     const std::error_category& error_category();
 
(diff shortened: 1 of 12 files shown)


travisdowns:
9ea327f5f is a rebase onto master with one conflict resolved: the local_error_category() fix from 1fbdf84f0 (#3563) is subsumed here, since GnuTLS errors are now thrown with the backend-neutral tls::error_category() directly. No other changes to the patch.

PR #3565, opened 2026-07-28
github: enable heap profiling on the dev job, fix alloc_test under it

travisdowns:
Nothing in the test matrix sets `Seastar_HEAP_PROFILING`, so the sampled memory profile code and every test behind `SEASTAR_HEAPPROF` are never compiled. Add `--heap-profiling` to the existing one-off `dev` job instead of a new matrix row, so this costs no extra CI time. The pairwise rows are untouched.

That immediately caught a latent breakage, fixed in the second commit: `alloc_test` compared `std::vector<allocation_site>` with `BOOST_CHECK_EQUAL`, which needs an `operator<<` for the vector. The only thing that ever supplied one is the generic printer in `sstring.hh`, behind `SEASTAR_DEPRECATED_OSTREAM_FORMATTERS` (default off since after those assertions were written in 06c1941). `BOOST_CHECK_EQUAL_COLLECTIONS` needs only the per-element `operator<<` the test already defines.

The failure was reproduced by this PR's own CI before the fix; `alloc_test` now builds and passes in a `--heap-profiling` dev build.

### DIFF
--- a/.github/workflows/gen-matrix.py
+++ b/.github/workflows/gen-matrix.py
@@ -95,11 +95,16 @@ MATRIX_PATH = ("jobs", "regular_test", "strategy", "matrix")
 # Items only set enable-ccache when overriding the default (test.yaml treats
 # absent/empty as enabled).
 SPECIAL_ITEMS: list[dict[str, Any]] = [
+    # The dev job doubles as our heap profiling coverage: nothing else in
+    # the matrix defines SEASTAR_HEAPPROF, so the sampled-memory-profile
+    # code and the tests guarded by it are otherwise never compiled.
     {
         "compiler": "clang++-22",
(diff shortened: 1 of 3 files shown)


travisdowns:
@scylladb/seastar-maint

travisdowns:
@scylladb/seastar-maint gentle ping, build fix

travisdowns:
@scylladb/seastar-maint gentle ping, build fix

travisdowns:
@scylladb/seastar-maint gentle ping

travisdowns:
push to https://github.com/scylladb/seastar/commit/fcfebbabc70fd92d7a80d1c1a0ff3ea4566c492e is a pure rebase on master tip

PR #3566, opened 2026-07-28
iostream: avoid deprecated consumption_result conversion

travisdowns:

Avoids the deprecated `consumption_result` constructor conversion, silencing the deprecation warnings it produces in the build.

Commit is `6c0450cfe` from #3559, cherry-picked verbatim onto master.

### DIFF
--- a/include/seastar/core/iostream-impl.hh
+++ b/include/seastar/core/iostream-impl.hh
@@ -243,8 +243,18 @@ input_stream<CharType>::consume(Consumer&& consumer) noexcept(std::is_nothrow_mo
                 return make_ready_future<stop_iteration>(stop_iteration::no);
             });
         }
-        return consumer(std::move(_buf)).then([this] (consumption_result_type result) {
-            return seastar::visit(result.get(), [this] (const continue_consuming&) {
+        return consumer(std::move(_buf)).then([this] (auto result) {
+            auto normalized = [&] {
+                if constexpr (InputStreamConsumer<Consumer, CharType>) {
+                    return std::move(result);
+                } else {
+                    if (result.has_value()) {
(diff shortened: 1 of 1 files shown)


travisdowns:
@scylladb/seastar-maint take a look, deprecation fix

include/seastar/core/iostream-impl.hh
```diff
@@ -243,8 +243,18 @@ input_stream<CharType>::consume(Consumer&& consumer) noexcept(std::is_nothrow_mo
                 return make_ready_future<stop_iteration>(stop_iteration::no);
             });
         }
-        return consumer(std::move(_buf)).then([this] (consumption_result_type result) {
-            return seastar::visit(result.get(), [this] (const continue_consuming&) {
+        return consumer(std::move(_buf)).then([this] (auto result) {
+            auto normalized = [&] {
+                if constexpr (InputStreamConsumer<Consumer, CharType>) {
+                    return std::move(result);
+                } else {
```
xemul, line 250: But it was the goal of deprecation to explicitly warn users of legacy consumers that it's time to switch to new consumer model. This `else` branch hides the deprecation and keeps supporting legacy consumers

What am I missing?

travisdowns: Right, though though this function is an internal user, so it wasn't clear if we wanted to emit a warning when a ObsoleteInputStreamConsumer is used, vs complaining when such an object is constructed or whatever.

Anyway though the goal was to get rid of the in-tree deprecations and this should do a better job of that:

https://github.com/scylladb/seastar/pull/3579

This just removes the remaining obsolete input stream users.

travisdowns:
Closing in favor of https://github.com/scylladb/seastar/pull/3579.

PR #3567, opened 2026-07-28
dns: move-construct optional input stream

travisdowns:

Uses `emplace()` to move-construct the optional `input_stream` instead of the deprecated `input_stream` move-assignment, silencing that deprecation warning.

Commit is `dabe12f3a` from #3559, cherry-picked verbatim onto master.


travisdowns:
@scylladb/seastar-maint please take a look at this 1-liner deprecation fix.

Issue #3569, opened 2026-07-28
tests/io_queue: test_2_class_group_bandwidth_throttler_fair_shares is flaky (1-second measurement window is dominated by the burst)


travisdowns:
`test_2_class_group_bandwidth_throttler_fair_shares` in `tests/unit/io_queue_test.cc` fails intermittently in CI on the fair-share ratio check:

```
/__w/seastar/seastar/tests/unit/io_queue_test.cc(977): fatal error: in "test_2_class_group_bandwidth_throttler_fair_shares": critical check float(bw0) / float(bw1) >= 3.95 has failed [3.93820214 < 3.9500000000000002]
```

Seen in https://github.com/scylladb/seastar/actions/runs/30399475894/job/90410502701?pr=3567

> **Note:** an earlier version of this description attributed the failure to sampling noise on a
> loaded CI machine. That was wrong; see the root-cause analysis below, which is measured rather
> than inferred. Thanks to @xemul for pushing back on it.

## Root cause

The measurement, not the throttler, is at fault, and specifically the length of the measurement
window.

`run_and_check_bandwidth()` polls once a second for up to 60s, but each workload compares **its own**
bandwidth against **its own** goal (`real_bandwidth >= bandwidth_goal`) and returns as soon as that is
met. It never looks at the ratio; the ratio is checked once, afterwards, on whatever two values came
back. The goals here are 76 and 19 MB/s and the first 1-second sample already measures ~87 and
~22 MB/s, so **the loop always exits on the first sample** and the 60s budget is never used.

`real_bandwidth` is a cumulative average with an integer-second denominator, so what the test
actually compares is two integer request counts accumulated over a single second:

- the CI failure value 3.93820214 is exactly **701/178**
- locally the same test yields **703/176** (3.994318) and **704/176** (4.000000)

The entire difference between passing and failing is ±2 requests. At those counts one request of
jitter in the low-share class moves the ratio by 0.0221, i.e. **22% of the whole 0.1-wide band**;
there are only ~4 request-counts of slack. This is also why three separate CI runs printed a
bit-identical 3.93820214 - at these magnitudes the ratio is quantized to a handful of values.

### What actually jitters: the burst

The counts sum to 109.88 MiB, i.e. bandwidth (100 MiB) + the configured burst (10 MiB). Splitting
each measurement into steady-state share plus consumed burst credit:

| | burst consumed (hi : lo) | burst split ratio |
| --- | --- | --- |
[...]

PR #3570, opened 2026-07-28
tests/io_queue: measure fair shares over a longer window

travisdowns:
`test_2_class_group_bandwidth_throttler_fair_shares` compares two bandwidths against each other
within 1.25%, but `run_and_check_bandwidth()` returns on the first sample that meets the goal, and
that is always the sample at t=1s. The value is a cumulative average since the start of the run, so
it folds in the initial burst, which at one second is ~9% of the bytes measured.

Per-second request counts over increasing windows, idle machine:

| window | pc0 | pc1 |
| --- | --- | --- |
| 1s | 704 | 176 |
| 2s | 672 | 168 |
| 4s | 656 | 164 |
| 8s | 648 | 162 |

Steady state is 640 and 160 requests/s (80 and 20 MiB/s); the burst contributes 64 and 16 requests,
i.e. 8 and 2 MiB, an exact 4:1 split. Both the burst term and the quantisation of the measurement
fall off as 1/window, and the observed spread of the ratio follows: 0.0057 at one second, 0.0008 at
eight.

At one second the entire comparison rests on two integer counts of 704 and 176, where one request of
difference in the smaller one moves the ratio by 22% of the interval being checked. The CI failure
value 3.93820214 is exactly `701/178`, against `703/176` and `704/176` locally - two requests either
way is the whole margin.

So: measure for at least 4s in this test, diluting both error terms by 4x. Other callers of
`run_and_check_bandwidth` keep the previous behaviour, so only this test gets slower - full
`io_queue_test` goes from 15.3s to 18.3s. 2s would have been enough for the observed failures (1.6x
margin); 4s is the smallest value with a comfortable one.

Supersedes the earlier version of this PR, which just disabled the test. Refs #3569.

### One thing worth flagging separately

This makes the test robust to the burst term, but not to reactor starvation, and starvation produces
a much larger effect that a longer window will not fix. Pinning the test to cores 0-1 and running
spinners on those same cores:

| load on the reactor's cores | result |
| --- | --- |
| 1 spinner/core | 25/25 pass, pc1 count still 176 |
| 2 spinners/core | 24/24 fail, ratios 2.96-3.05, pc1 count 217-222 |
| 3 spinners/core | 25/25 fail, ratios 2.02-2.58, pc1 count 272-333 |
| 5 spinners/core | 25/25 fail, ratios 2.48-3.16, pc1 count 211-252 |

The signature is consistent: pc0's count stays roughly flat while **pc1's grows**, so under
starvation the low-share class takes progressively more than its share. That is the shape I would
expect from the anti-starvation credit in `fair_queue::priority_class_group_data::push_from_idle()`:

```c++
pc._accumulated = std::max<signed_capacity_t>(_last_accumulated - cfg.forgiving_factor / pc._shares, pc._accumulated);
```

The backdate is `forgiving_factor / shares`, so the 100-share class gets a 4x larger credit than the
400-share class on every idle->queued transition, and a starved reactor means many more such
transitions. I have not chased this further and it may well be intended, but if it is not, it would
be a throttler issue rather than a test one. Happy to open a separate issue if that is useful.

### DIFF
--- a/tests/unit/io_queue_test.cc
+++ b/tests/unit/io_queue_test.cc
@@ -766,7 +766,14 @@ SEASTAR_THREAD_TEST_CASE(test_gauge_integrator_test) {
 // the configured bandwidth+burst limit by up to one threshold unit.
 static constexpr size_t bw_slack = 128*1024;
 
-static future<size_t> run_and_check_bandwidth(io_queue_for_tests& tio, internal::priority_class pc, size_t bandwidth_goal, unsigned parallelizm = 1, size_t req_size = 128*1024) {
+// The value returned is a cumulative average since the start of the run, so the
+// initial burst is folded into it: over a one second window the burst accounts
+// for ~9% of the bytes seen, and it shrinks as 1/window from there. A caller
+// that compares two of these against each other wants a window long enough that
+// the burst no longer dominates the comparison; min_window puts a floor on the
+// measured interval for those. The default preserves the historical behaviour of
+// returning on the first sample that meets the goal.
+static future<size_t> run_and_check_bandwidth(io_queue_for_tests& tio, internal::priority_class pc, size_t bandwidth_goal, unsigned parallelizm = 1, std::chrono::seconds min_window = std::chrono::seconds(1), size_t req_size = 128*1024) {
     fmt::print("Run {} workload\n", pc.id());
     bool keep_going = true;
     uint64_t nr_requests = 0;
@@ -793,7 +800,7 @@ static future<size_t> run_and_check_bandwidth(io_queue_for_tests& tio, internal:
         auto now = std::chrono::steady_clock::now();
         real_bandwidth = (nr_requests * req_size) / std::chrono::duration_cast<std::chrono::seconds>(now - start).count();
         fmt::print("Measured for {} {} MB/s, goal {} MB/s\n", pc.id(), real_bandwidth >> 20, bandwidth_goal >> 20);
-        if (real_bandwidth >= bandwidth_goal || now >= stop) {
+        if ((real_bandwidth >= bandwidth_goal && now >= start + min_window) || now >= stop) {
             break;
         }
     }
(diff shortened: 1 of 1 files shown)


xemul:
It will get broken very fast without a test. The run_and_check_bandwidth() actually tries to compensate for the "loaded CI machine" -- it runs two workloads in parallel and checks for the ratio once every while, if it doesn't converge into the given interval within a minute, it's unlikely the "loaded CI machine", it can be something else

travisdowns:
> It will get broken very fast without a test.

Fair.

Maybe it's possible to "unit test" this stuff with a manual clock that advances more deterministically, or perhaps that is very difficult?

CI is relatively flaky in seastar at the moment, I'd say over 50% of runs fail with _some_ error. Even a few tests with 1% flake rate lead to a high overall rate since we do a dozen runs.

travisdowns:
@xemul is the exact threshold you are checking against required for the test to be useful? Based on the flakes I've seen if you widen it slighlty, like [3.93, ... most flakes would be eliminated.

travisdowns:
You were right that disabling it was the wrong move, so I have replaced this PR with a fix. But the convergence loop does not do what it looks like: each workload compares **its own** bandwidth to **its own** goal and returns as soon as that is met, never the ratio, and both goals are already met at the first 1s sample. So the 60s budget is never used, and the single sample it does take is the one most contaminated by the burst (~9% of the bytes at one second).

Instrumented per-second counts, idle machine:

| window | pc0 | pc1 |
| --- | --- | --- |
| 1s | 704 | 176 |
| 4s | 656 | 164 |
| 8s | 648 | 162 |

Steady state is 640/160, burst is 64/16 requests = 8/2 MiB, an exact 4:1 split. Both the burst term and the measurement quantisation fall off as 1/window, so the ratio spread goes 0.0057 (1s) -> 0.0008 (8s).

To your question about the threshold: at 1s the comparison is two integer counts of 704 and 176, and one request of difference in the smaller moves the ratio by 22% of the interval. The CI value 3.93820214 is exactly `701/178` vs `703/176` locally. So rather than widen the threshold, this measures for 4s, which buys 4x on both error terms and lets the threshold stay as tight as it is (it could be tightened). Only this test gets longer; full `io_queue_test` 15.3s -> 18.3s.

Separately, and not fixed here: under actual reactor starvation (spinners pinned to the reactor cores) the test fails 100% with ratios down to 2.0, and the signature is pc1's count growing while pc0's stays flat. That looks like the `forgiving_factor / pc._shares` backdate in `push_from_idle()` handing the low-share class a 4x larger credit per idle->queued transition. Details in the PR description - possibly intended, but if not it is a throttler question rather than a test one.

xemul:
> Maybe it's possible to "unit test" this stuff with a manual clock that advances more deterministically, or perhaps that is very difficult?

Not difficult but too much patching. All internals of io_/fair_queue use lowres_clock explicitly. Patching the whole thing with template<typename Clock> parameter would do the trick, but is a huge change

xemul:
> @xemul is the exact threshold you are checking against required for the test to be useful? Based on the flakes I've seen if you widen it slighlty, like [3.93, ... most flakes would be eliminated.

Well, I picked 1 minute because I've seen that the test converges in less than a second in my local runs even when I explicitly pinned it to a single code and stressed the CPU with alien load. So I judged that 1 minute should be enough for anybody

travisdowns:
> Well, I picked 1 minute because I've seen that the test converges in less than a second in my local runs even when I explicitly pinned it to a single code and stressed the CPU with alien load. So I judged that 1 minute should be enough for anybody

Right, but per the slop above, the 1 minute threshold doesn't come into play here: that part of the test finishes within 1 second, but then the check fails due to a small amount of noise in the request count. So the fix here expands the minimum window. Other fix could be to put sort of an outer loop around the whole thing to keep checking the ratio (for up to 1 minute) until it stabilizes. 

In any case, looks like the current one merged!

PR #3574, opened 2026-07-29
test: fix reversed buffer bound in add_exception_message()

travisdowns:
`add_exception_message()` in `src/testing/seastar_test.cc` formats a failing test's exception chain into a fixed 4096 byte buffer, passing the space left to `fmt::format_to_n()` as `end - out`. The branch that reports exceptions not derived from `std::exception` passes `out - end` instead:

```c++
    if (ep) {
        out = fmt::format_to_n(out, end - out, "{}{}: {}", pfx, ...).out;
    } else {
        auto tp = abi::__cxa_current_exception_type();
        out = fmt::format_to_n(out, out - end, "{}{}", pfx, ...).out;   // <--
    }
```

`out <= end` always holds, so `out - end` is non-positive and converts to a huge `size_t`. The call is then effectively unbounded: it writes the whole formatted string regardless of the room left, and the iterator it returns can end up past `end`, which makes the following `*out = 0` write out of bounds as well. The comment there ("always valid") holds only while the bound is honoured.

### Reachability

The branch runs when the exception that failed the test does not derive from `std::exception` - either thrown directly (the `catch (...)` in `seastar_test::run()`) or as the innermost link of a nested chain (the `catch (...)` around `std::rethrow_if_nested`). It is terminal: the recursion at the end is guarded by `if (ep)`, so the bad `out` is never passed to another frame.

Overflowing therefore needs the preceding, correctly bounded, frames to have very nearly filled the buffer. With `end = buf + 4095`, if a chain of `std::exception` frames leaves, say, 5 bytes free, the terminal frame writes `"; Caused by "` plus the type name - 12 bytes plus the name - in full, running past the end of the static buffer and then NUL-terminating beyond it. Short chains have room to spare and stay inside the buffer, which is why this has not been seen in practice.

It is confined to `libseastar_testing`, so it cannot affect production builds; the failure mode is memory corruption while reporting a test failure.

### Test

The test added with the original code in 7039ebc17 nests `std::out_of_range` around `std::invalid_argument`, so both frames take the correctly bounded branch and the buggy one had no coverage at all. This adds a case that throws a type which is not a `std::exception` and checks the reported message names it.

To be explicit: the new test covers the branch, not the overflow. Provoking the overflow needs a message sized to leave an exact few bytes free, which would be a brittle thing to commit; I left it out deliberately.

Verified with `tests/unit/test_fixture_test` in a dev (ASan/UBSan) build - full binary passes.

### DIFF
--- a/src/testing/seastar_test.cc
+++ b/src/testing/seastar_test.cc
@@ -47,7 +47,7 @@ add_exception_message(const std::exception* ep, bool rec, char *out, char *end) 
         out = fmt::format_to_n(out, end - out, "{}{}: {}", pfx, seastar::pretty_type_name(typeid(*ep)), ep->what()).out;
     } else {
         auto tp = abi::__cxa_current_exception_type();
-        out = fmt::format_to_n(out, out - end, "{}{}", pfx, tp ? seastar::pretty_type_name(*tp) : "<unknown>").out;
+        out = fmt::format_to_n(out, end - out, "{}{}", pfx, tp ? seastar::pretty_type_name(*tp) : "<unknown>").out;
     }
 
     boost::execution_exception::location loc;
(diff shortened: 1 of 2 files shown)


travisdowns:
@scylladb/seastar-maint please have a look

PR #3575, opened 2026-07-29
treewide: wrap exception_ptr before formatting

travisdowns:
Split out of #3559 (one deprecation fix per PR). Stephan's commit, cherry-picked verbatim onto master.

`fmt::formatter<std::exception_ptr>` was deprecated in c0e72e546 in favour of `seastar::formattable()`; this stops Seastar formatting exception_ptr through it. libseastar instantiates the deprecated formatter in 62 places across 14 translation units before this change, and none after.

Worth noting for reviewers that those 62 are not visible as warnings in every build. The deprecation sits on `formatter::parse()`, which {fmt} only ever calls from inside `fmt/base.h`, so the diagnostic's primary location is a {fmt} header rather than the offending call site. Where {fmt} is included with `-isystem`, as it is when cooked, clang-22 does not report it and the build looks clean. They do show up with `-Wsystem-headers`, for builds that have {fmt} on a plain `-I` include path, and on clang-23, which reports this class of diagnostic where clang-22 stays silent.

Verified by building `libseastar` plus `iotune`, `rpc_tester`, `httpd`, `http_client_demo` and the six changed unit tests: no errors, and zero remaining exception_ptr deprecation diagnostics under `-Wsystem-headers`.

### DIFF
--- a/apps/httpd/main.cc
+++ b/apps/httpd/main.cc
@@ -116,7 +116,7 @@ int main(int ac, char** av) {
                 prometheus_started = true;
 
                 prometheus_server.listen(socket_address{prom_addr, pport}).handle_exception([prom_addr, pport] (auto ep) {
-                    std::cerr << seastar::format("Could not start Prometheus API server on {}:{}: {}\n", prom_addr, pport, ep);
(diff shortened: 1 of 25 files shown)


travisdowns:
@scylladb/seastar-maint more deprecation warning fixes

Issue #3571, opened 2026-07-28
nested_exception::what() discards both nested exceptions, making failures undiagnosable


travisdowns:
`seastar::nested_exception` carries the two exceptions that produced it (`inner` and `outer`), but
`what()` throws that information away and returns a fixed string literal:

https://github.com/scylladb/seastar/blob/master/src/core/future.cc#L161-L163

```c++
const char* nested_exception::what() const noexcept {
    return "seastar::nested_exception";
}
```

Anything that reports a failure via `what()` therefore tells you only that *some* pair of
exceptions occurred, not what either of them was. In practice this makes a `nested_exception`
reaching a log line or a test harness nearly undiagnosable.

### Concrete impact

A CI failure of the TLS tests reports exactly this and nothing more, twice over, because Boost
prints the test name followed by `what()`:

```
unknown location(0): fatal error: in "test_x509_client_with_builder_system_trust_multiple": seastar::nested_exception: seastar::nested_exception
/__w/seastar/seastar/tests/unit/tls_test.cc(105): last checkpoint
```

https://github.com/scylladb/seastar/actions/runs/30400681315/job/90414636407

Both the underlying network/TLS error and the secondary error from the cleanup path are present in
the exception object at that point and are simply not printed. Working out even the general shape
of the failure required reasoning backwards from which `finally()` could have produced the nested
exception at all, rather than reading it off the log.

Since `nested_exception` is what `future::finally()` produces whenever both the body and the
cleanup callback fail, this affects a fairly common failure shape: an operation fails, and the
close/cleanup on the now-broken resource fails too. That is precisely the case where you most want
to see the original error.

### Suggested fix

Have `what()` report both nested causes, e.g. something along the lines of

```
seastar::nested_exception: <outer what()> (while handling <inner what()>)
```

Two constraints worth noting for whoever picks this up:

- `what()` is `noexcept` and must return a pointer that stays valid for the lifetime of the
  exception, so the message has to be materialised into storage owned by the object rather than
  built on demand.
- The constructors are all `noexcept` (including the copy constructor), so formatting the message
  eagerly in the constructor cannot be allowed to propagate a `bad_alloc`. A `try`/`catch` falling
  back to the current fixed literal would keep the existing guarantees intact.

Extracting the nested `what()` strings requires rethrow-and-catch on the stored `exception_ptr`s
(or `std::current_exception` plumbing); `rethrow_nested()` already exists but only rethrows `outer`.

[...]

PR #3577, opened 2026-07-29
test: report exception chains through seastar::formattable()

travisdowns:
### Why this code exists

When a C++ exception escapes a test case, Boost.Test reports its type and `what()`, and nothing more. For exceptions that keep their real detail somewhere other than `what()` - nested exceptions - that drops the actual cause, which is what #3165 was about.

`seastar_test::run()` therefore gets in ahead of boost. It is the function boost knows as the test case (registered with `make_test_case([this] { run(); }, ...)`), and rather than let the exception escape into boost's own translator, it catches it, formats a message, and throws a `boost::execution_exception` - boost's *already-translated* form. `unit_test_monitor_t::execute_and_translate()` catches that directly and logs it as-is, so the message seastar builds is what lands in

```
<location>: fatal error: in "<test name>": <message>
```

`add_exception_message()` builds that message. (The buffer is `static` because `execution_exception` keeps its message as a non-owning `const_string`.)

### The problem

`add_exception_message()` walks the chain itself, recursing with `std::rethrow_if_nested()`. That only knows `std::nested_exception`, and `seastar::nested_exception` does not derive from it, so the walk stops at the first one - and since its `what()` is a fixed string, everything below it is lost too:

```
before:  seastar::nested_exception: seastar::nested_exception
after:   seastar::nested_exception: std::runtime_error (cleanup failed) (while cleaning up
         after std::_Nested_exception<std::out_of_range> (Message 2): std::invalid_argument (Message 1))
```

`future::finally()` produces that shape whenever both the body and the cleanup fail, so it is not exotic; #3571 is a CI failure reporting exactly the first line and nothing more.

### The change

`seastar::formattable()` already renders all of it - both flavours of nested exception, `std::system_error` detail, `backtraced<>` frames, non-`std::exception` objects. This delegates to it rather than teaching a second walker about each type, leaving one renderer for exception chains. The recursion goes too: only the outermost location reaches boost, so a single `catch (const boost::exception&)` replaces the walk. Net +20/-44.

### Behaviour changes

Chains that already worked are restyled, as the message now comes from the same renderer as every seastar log line:

```
before:  std::_Nested_exception<std::out_of_range>: Message 2; Caused by std::invalid_argument: Message 1
after:   std::_Nested_exception<std::out_of_range> (Message 2): std::invalid_argument (Message 1)
```

The per-frame `(file:function:line)` annotation for *nested* `boost::exception` frames is lost - the formatter has no notion of frames. The outermost location, the one boost reports, is kept.

### Testing

New case fails a test with a `seastar::nested_exception` and requires both causes in the message; it fails without this change (`critical check e.what().find("Message 1") != std::string::npos has failed`). Full `tests/unit/test_fixture_test` passes in a dev ASan/UBSan build, pre-existing cases unmodified.

Related: #3574 fixes a reversed buffer bound in `add_exception_message()`, which this deletes. They are independent - either can land first.

### DIFF
--- a/src/testing/seastar_test.cc
+++ b/src/testing/seastar_test.cc
@@ -33,55 +33,37 @@
 #include <seastar/core/on_internal_error.hh>
 #include <seastar/core/app-template.hh>
 #include <seastar/testing/on_internal_error.hh>
+#include <seastar/util/log.hh>
 
 namespace seastar {
 
 namespace testing {
 
 // #3165 - build a message for a possibly nested exception chain.
-static boost::execution_exception::location
-add_exception_message(const std::exception* ep, bool rec, char *out, char *end) {
-    auto pfx = rec ? "; Caused by " : "";
-
-    if (ep) {
-        out = fmt::format_to_n(out, end - out, "{}{}: {}", pfx, seastar::pretty_type_name(typeid(*ep)), ep->what()).out;
-    } else {
-        auto tp = abi::__cxa_current_exception_type();
-        out = fmt::format_to_n(out, end - out, "{}{}", pfx, tp ? seastar::pretty_type_name(*tp) : "<unknown>").out;
-    }
-
-    boost::execution_exception::location loc;
-    if (auto* be = dynamic_cast<const boost::exception*>(ep)) {
-        auto sloc = boost::exception_detail::get_exception_throw_location(*be);
-        if (rec) {
-            out = fmt::format_to_n(out, end - out, "({}:{}:{})", sloc.file_name(), sloc.function_name(), sloc.line()).out;
-        }
-        loc = boost::execution_exception::location(sloc.file_name(), sloc.line(), sloc.function_name());
-    }
-
-    *out = 0; // see initial invoke below. always valid
-
-    if (ep) {
-        try {
-            std::rethrow_if_nested(*ep);
-        } catch (std::exception& e) {
-            add_exception_message(&e, true, out, end);
-        } catch (...) {
-            add_exception_message(nullptr, true, out, end);
-        }
-    }
-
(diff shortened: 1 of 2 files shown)


avikivity:
I have a patch somewhere that makes seastar::nested_exception use std::nested_exception (since we'll eventually move to fmt's exception formatter, which isn't aware of seastar::nested_exception).

travisdowns:
> I have a patch somewhere that makes seastar::nested_exception use std::nested_exception (since we'll eventually move to fmt's exception formatter, which isn't aware of seastar::nested_exception).

That SG. In that light, is this PR DOA or do we want it in the meantime? It seems like both a simplification and diagnostic improvement in the meantime.

travisdowns:
force push 526090dec26a3a5ce22db5eed377b99b30794e9e was a pure rebase

avikivity:
> > I have a patch somewhere that makes seastar::nested_exception use std::nested_exception (since we'll eventually move to fmt's exception formatter, which isn't aware of seastar::nested_exception).
> 
> That SG.

SG?

> In that light, is this PR DOA or do we want it in the meantime? It seems like both a simplification and diagnostic improvement in the meantime.

I think it's an improvement regardless.

avikivity:
> I think it's an improvement regardless.

(I'm waiting for @elcallio's review)

travisdowns:
> SG?

Sorry, "Sounds Good", as in SGTM but it probably not valid to take prefixes like this :)

elcallio (approved):

travisdowns:
Moving to draft as this is probably partly or completely obsoleted by by https://github.com/scylladb/seastar/issues/3571

PR #3578, opened 2026-07-30
net: posix-stack: Fix conntrack load balancer skew (#3514)

### LINKED ISSUE #3514: Posix connection load balancer selects shard 0 too much

NITIN9181:
Fixes #3514.

Hey guys, taking a stab at the load balancer skew issue reported in #3514. 

The root cause here is that `conntrack::load_balancer` uses a plain `std::vector<unsigned>` to track connection counts. When a connection closes on any shard other than Shard 0, the handle's destructor fires off an `smp::submit_to` message back to Shard 0 to decrement the count.

The problem is that Shard 0 processes its own closed connections inline (synchronously), but handles the cross-core decrements asynchronously whenever it gets around to polling the message queue. Under heavy load with lots of short-lived connections, Shard 0's decrements happen instantly while the other shards' decrements get backed up in the queue. This makes the other shards look artificially bloated, so the load balancer just keeps dumping more and more new connections onto Shard 0, creating a vicious cycle.

### The Fix
To fix this, I swapped out the `std::vector` for a `std::unique_ptr<std::atomic<unsigned>[]>`. 

Now, when a connection closes on any core, the handle just does a direct `fetch_sub(1, std::memory_order_relaxed)` on the atomic array. By bypassing the SMP message queue entirely, Shard 0 always has an instantly accurate view of the load across all shards, stopping the feedback loop. 

I kept the `foreign_ptr` around so the load balancer object itself is still safely cleaned up on its home core when it's done.

Tested locally against the unit tests and it builds perfectly. Let me know what you think of this approach!

### DIFF
--- a/include/seastar/net/posix-stack.hh
+++ b/include/seastar/net/posix-stack.hh
@@ -59,24 +59,35 @@ using namespace seastar;
 // Right now this class is used by the posix_server_socket_impl, but it could be used by any other.
 class conntrack {
     class load_balancer {
-        std::vector<unsigned> _cpu_load;
+        std::unique_ptr<std::atomic<unsigned>[]> _cpu_load;
     public:
-        load_balancer() : _cpu_load(size_t(this_smp_shard_count()), 0) {}
+        load_balancer() : _cpu_load(new std::atomic<unsigned>[this_smp_shard_count()]()) {}
         void closed_cpu(shard_id cpu) {
-            _cpu_load[cpu]--;
+            _cpu_load[cpu].fetch_sub(1, std::memory_order_relaxed);
         }
         shard_id next_cpu() {
             // FIXME: The naive algorithm will just round robin the connections around the shards.
             // A more complex version can keep track of the amount of activity in each connection,
             // and use that information.
-            auto min_el = std::min_element(_cpu_load.begin(), _cpu_load.end());
-            auto cpu = shard_id(std::distance(_cpu_load.begin(), min_el));
-            _cpu_load[cpu]++;
-            return cpu;
+            unsigned min_val = std::numeric_limits<unsigned>::max();
+            shard_id min_cpu = 0;
+            unsigned count = this_smp_shard_count();
+            for (shard_id i = 0; i < count; ++i) {
+                unsigned val = _cpu_load[i].load(std::memory_order_relaxed);
+                if (val < min_val) {
+                    min_val = val;
+                    min_cpu = i;
+                }
+            }
+            _cpu_load[min_cpu].fetch_add(1, std::memory_order_relaxed);
+            return min_cpu;
         }
         shard_id force_cpu(shard_id cpu) {
-            _cpu_load[cpu]++;
+            _cpu_load[cpu].fetch_add(1, std::memory_order_relaxed);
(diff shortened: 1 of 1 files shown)


avikivity:
> To fix this, I swapped out the std::vector for a std::unique_ptr<std::atomic<unsigned>[]>.



This will murder any large SMP with cache line contention.

Seastar avoids atomic operations, especially cross-shard, to retain scalability.

NITIN9181:
You are completely right, thank you for catching that. I was too focused on bypassing the message queue lag and overlooked the cache line contention this would introduce on large SMP machines. Breaking the shared-nothing architecture with cross-core atomics definitely violates Seastar's scalability principles.

Given that cross-core atomics are a no-go, what would be your preferred architectural approach to bound the decrement lag described in the issue? 

1. **Batching:** Track closures locally on each shard and flush them back to Shard 0 via `smp::submit_to` periodically (e.g., every N connections or every few milliseconds) to reduce message volume while keeping Shard 0's view reasonably accurate.
2. **Polling:** Ensure Shard 0 yields or polls the SMP message queue more aggressively during heavy `accept()` loops so the lag stays bounded.
3. **SO_REUSEPORT:** Avoid the counting problem entirely by removing the hardcoded `return false` in `posix_reuseport_available()` and relying on the kernel's hash to distribute connections instead of `conntrack`.

I'm happy to close this and pivot to whichever direction aligns best with the project's long-term goals. Let me know what you think!

avikivity:
> You are completely right, thank you for catching that. 

You're the second one to suggest this exact solution.


> I was too focused on bypassing the message queue lag and overlooked the cache line contention this would introduce on large SMP machines. Breaking the shared-nothing architecture with cross-core atomics definitely violates Seastar's scalability principles.
> 
> Given that cross-core atomics are a no-go, what would be your preferred architectural approach to bound the decrement lag described in the issue?
> 
> 1. **Batching:** Track closures locally on each shard and flush them back to Shard 0 via `smp::submit_to` periodically (e.g., every N connections or every few milliseconds) to reduce message volume while keeping Shard 0's view reasonably accurate.
> 2. **Polling:** Ensure Shard 0 yields or polls the SMP message queue more aggressively during heavy `accept()` loops so the lag stays bounded.
> 3. **SO_REUSEPORT:** Avoid the counting problem entirely by removing the hardcoded `return false` in `posix_reuseport_available()` and relying on the kernel's hash to distribute connections instead of `conntrack`.

SO_REUSEPORT is not reliable enough to get good distribution.


But perhaps a combination of SO_REUSEPORT _and_ the current connection tracking. Each SO_REUSEPORT listener tries to distribute its own connections to all shards, and keeps track of its connections' distribution. Each such listener will have bias, but since there are many, the biases will cancel out.

> 
> I'm happy to close this and pivot to whichever direction aligns best with the project's long-term goals. Let me know what you think!

NITIN9181:
That's a clever way to use the biases to cancel each other out! I'll drop the atomics and work on implementing the SO_REUSEPORT + conntrack combo.

PR #3584, opened 2026-08-02
reactor: probe reactor backend availability once, not three times

### LINKED ISSUE #3583: FATAL: Exception during startup, aborting: std::invalid_argument (find_candidate(): failed to find candidate io_uring)
Apparently, under a loaded machine, `detect_io_uring()` may fail with ` -ENOMEM/-EAGAIN` and there's no retry mechanism. 

This leads to the following sequence of events:
- Call 1 (candidates list) → io_uring probe fails transiently → "io_uring" not added to candidates.
- Call 2 (default_backend()) → probes again, this time succeeds → returns "io_uring" as the chosen default name.
- selection_value constructor does find_candidate("io_uring") against the candidates list from call 1, which doesn't have it → throws std::invalid_argument("find_candidate(): failed to find candidate io_uring") → uncaught at startup → FATAL: Exception during startup, aborting → process exits before ever acquiring a host_id.

Seen in https://github.com/scylladb/scylladb/pull/30955#issuecomment-5136374972 - a completely unrelated PR, the 3rd node failed to be added to the ScyllaDB Cluster with the above error.

mykaul:
## Summary

`reactor_options`'s constructor called `reactor_backend_selector::available()` independently 3 times: to build the `reactor-backend` option's candidates list, via `default_backend()` for its default value,  3rd time to format its description. 
Each call performed a real `io_uring_setup(2)` probe (`detect_io_uring()` / `detect_asymmetric_io_uring()`).

If one of the 3 calls disagreed with another (e.g. the "default" call detected `io_uring` while the "candidates" call didn't), `selection_value`'s constructor performed `find_candidate(default_candidate)` against a candidates list that did not contain that name, and then threw `std::invalid_argument` uncaught during startup:

```
FATAL: Exception during startup, aborting: std::invalid_argument (find_candidate(): failed to find candidate io_uring)
```

## Fix

Probe `available()` exactly once ever, cached in a function-local static inside `available()` itself. Every caller (candidates list, `default_backend()`, description) agrees for free, regardless of why any single probe might vary. Also stop probing `asymmetric_io_uring` once plain `io_uring` has already failed: it requires a strict superset of the same required features, so a second, more expensive (SQPOLL, kernel-worker-thread-creating) probe can't succeed where the first already didn't.

**v2**: reworked from the original approach after review. v1 cached the probe result in a new `reactor_backend_setup` member added to `reactor_options` (flagged as an internal detail that shouldn't live in a public struct) and retried the underlying syscall on `-ENOMEM`/`-EAGAIN`. Both are gone now: the cache moved into `available()` itself, so `reactor_options`/`reactor_config.hh` are back to their original, unmodified form; the retry is dropped since neither errno is actually a documented, transient `io_uring_setup(2)` failure worth retrying (`EAGAIN` isn't even a documented return code for it), and probing once removes the possibility of disagreement outright rather than trying to paper over one occurrence of it.

## Verification

- Reproduced the original crash deterministically against the *unmodified* binary using an `LD_PRELOAD` shim that intercepts the real `io_uring_queue_init_params` and forces only the first of the process's three `entries=1` probe calls to fail.
- Applied the fix and confirmed the same shim no longer crashes the process: since there's now only one probe, `io_uring` is consistently excluded from (or included in) both the candidates and the default together.
- `program_options_test` and `reactor_backend_test` pass.
- `hello-world` demo builds and runs correctly with `--help-seastar`, listing consistent candidates/default.

## Root cause note

This fixes a definitively real, reproducible bug (three independent probes that can disagree with each other), confirmed via the LD_PRELOAD repro above. It is *not* confirmed to be the root cause of the specific incident reported in #3583 (a ScyllaDB node failing to join with `host_id <missing>`) -- the link between that incident and this bug is circumstantial, not evidenced by a captured log/stack trace. So: `Refs #3583` rather than `Fixes #3583` -- please keep that issue open until/unless the connection is actually confirmed.

Refs: #3583

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -2643,20 +2643,33 @@ reactor_backend_selector reactor_backend_selector::default_backend() {
 }
 
 std::vector<reactor_backend_selector> reactor_backend_selector::available() {
-    std::vector<reactor_backend_selector> ret;
+    // Each of the checks below does a real, sometimes expensive kernel probe
+    // (e.g. io_uring detection creates and tears down actual io_uring
+    // instances). Probe once and cache the result: callers (the
+    // --reactor-backend candidates list, its default, and its description)
+    // must all agree on what's available, and repeated independent probing
+    // can give different answers under load.
+    static const std::vector<reactor_backend_selector> cached = [] {
+        std::vector<reactor_backend_selector> ret;
 #ifdef SEASTAR_HAVE_URING
-    if (detect_io_uring()) {
-        ret.push_back(reactor_backend_selector("io_uring"));
-    }
-    if (detect_asymmetric_io_uring()) {
-        ret.push_back(reactor_backend_selector("asymmetric_io_uring"));
-    }
+        if (detect_io_uring()) {
+            ret.push_back(reactor_backend_selector("io_uring"));
+            // asymmetric_io_uring requires a strict superset of what plain
+            // io_uring requires, so there's no point probing it (another
+            // real io_uring_setup(2) call, plus SQPOLL kernel worker thread
+            // creation) once the plain probe has already failed.
+            if (detect_asymmetric_io_uring()) {
+                ret.push_back(reactor_backend_selector("asymmetric_io_uring"));
+            }
+        }
 #endif
-    if (has_enough_aio_nr() && detect_aio_poll()) {
-        ret.push_back(reactor_backend_selector("linux-aio"));
-    }
-    ret.push_back(reactor_backend_selector("epoll"));
-    return ret;
+        if (has_enough_aio_nr() && detect_aio_poll()) {
+            ret.push_back(reactor_backend_selector("linux-aio"));
+        }
+        ret.push_back(reactor_backend_selector("epoll"));
+        return ret;
+    }();
+    return cached;
 }
 
 std::shared_ptr<reactor_backend_configurator> reactor_backend_selector::configurator(resource::cpuset cpu_set, const reactor_options& reactor_opts, const smp_options& smp_opts) const {


mykaul:
Pushed a follow-up commit: `build_backend_candidates()` now copies instead of moves out of `available` (each `reactor_backend_selector` is just one string, so this is free), removing the forced ordering constraint that had `make_reactor_backend_setup()` populating `description`/`default_name` before `candidates` — backwards from the struct's declared order. Fields are now populated in their natural order: candidates, then default_name, then description.

mykaul:
Re: the CI failure on the fork (https://github.com/mykaul/seastar/actions/runs/30743066121) — both red jobs are unrelated to this diff:

1. `Test (dpdk, ...)` fails at checkout: `repository 'https://github.com/mykaul/dpdk/' not found`. This is a submodule-mirroring issue specific to running this CI config against a personal fork that doesn't have a mirrored `dpdk` submodule — not something this PR's code touches.
2. `Test (g++-15, C++23, release, x86)` fails on `Seastar.unit.prometheus`'s `test_filtering_by_label_with_aggregation` (expected 0 matches, got 4). This diff never touches metrics/prometheus code — reactor backend selection is unrelated to that subsystem. All other 12 build/test matrix jobs (debug/release/sanitize/dev/modules, x86/arm, clang 21/22, g++ 15/16, C++23/26) passed cleanly.

Also pushed a second commit: retry handling for the underlying `io_uring_setup(2)` probe on transient `-ENOMEM`/`-EAGAIN` (with a warning log), since the original crash fix alone still left a single unlucky probe permanently downgrading the backend with no visibility into why.

avikivity:
Please reduce the verbiage here. The title indicates its a good and simple fix. But the cover letter is a PhD thesis. The effort required to review this is out of proportion. It's easy for you to generate but hard for me to consume.

mykaul:
> Please reduce the verbiage here. The title indicates its a good and simple fix. But the cover letter is a PhD thesis. The effort required to review this is out of proportion. It's easy for you to generate but hard for me to consume.

Done.

src/core/reactor_backend.cc
```diff
@@ -1223,6 +1223,29 @@ void reactor_backend_epoll::reset_preemption_monitor() {
 
 #ifdef SEASTAR_HAVE_URING
 
+// io_uring_setup(2) can fail transiently under load (-ENOMEM/-EAGAIN from ring
+// allocation or, for SQPOLL rings, kernel worker thread creation). Retry a few
+// times before giving up so a momentarily loaded machine doesn't permanently
+// lose this backend for the process's lifetime.
```
avikivity: This sounds extremely dubious. ENOMEM would be very very very rare.

include/seastar/core/reactor_config.hh
```diff
@@ -58,6 +58,17 @@ class reactor_backend_selector;
 class crypto_provider_factory;
 class network_stack_factory;
```
avikivity, line 60: > probe, which can transiently fail under load (-ENOMEM/-EAGAIN). If one call

Is there any evidence for this?

include/seastar/core/reactor_config.hh
```diff
@@ -58,6 +58,17 @@ class reactor_backend_selector;
...
 
+/// \cond internal
+// A single reactor_backend_selector::available() probe result, cached so
+// reactor_backend's candidates/default/description all agree (probing more
+// than once can give different results under load).
+struct reactor_backend_setup {
+    program_options::selection_value<reactor_backend_selector>::candidates candidates;
+    std::string default_name;
+    std::string description;
+};
```
avikivity: This should be in namespace internal.

include/seastar/core/reactor_config.hh
```diff
@@ -165,6 +176,12 @@ struct reactor_options : public program_options::option_group {
     /// \note Even if the \p seastar_memory logger is set to debug or trace
     /// level, the diagnostics will be logged irrespective of this setting.
     program_options::value<memory::alloc_failure_kind> dump_memory_diagnostics_on_alloc_failure_kind;
+
+    /// \cond internal
+    // Must be declared (and thus initialized) before reactor_backend below.
+    reactor_backend_setup _reactor_backend_setup;
+    /// \endcond
```
avikivity: It's wrong to add this internal thing in a public struct.

include/seastar/core/reactor_config.hh
```diff
@@ -58,6 +58,17 @@ class reactor_backend_selector;
...
+/// \cond internal
+// A single reactor_backend_selector::available() probe result, cached so
+// reactor_backend's candidates/default/description all agree (probing more
+// than once can give different results under load).
+struct reactor_backend_setup {
+    program_options::selection_value<reactor_backend_selector>::candidates candidates;
+    std::string default_name;
+    std::string description;
+};
+/// \endcond
```
avikivity: I don't understand this new thing. Why is it needed?

src/core/reactor_backend.cc
mykaul: OK. Do you want me to remove the -ENOMEM handling?

include/seastar/core/reactor_config.hh
mykaul: 1. No - the log is not great. The Scylla node just died.
2. Yes - see https://github.com/scylladb/seastar/issues/3583#issuecomment-5156991994 - there was a kinda reproduction. Copy pasting from there:
Deterministic repro against the unmodified binary (no seastar/scylla source changes): an LD_PRELOAD shim intercepts the real, dynamically-linked io_uring_queue_init_params (exported by liburing.so.2) and makes only the first of the process's three entries=1 probe calls fail, leaving the rest untouched:

```
[shim] forcing detect_io_uring() probe #0 to fail with -ENOMEM
[shim] letting detect_io_uring() probe #1 proceed normally
[shim] letting detect_io_uring() probe #2 proceed normally
FATAL: Exception during startup, aborting: std::invalid_argument (find_candidate(): failed to find candidate io_uring)
```

src/core/reactor_backend.cc
avikivity: I want to understand what the goal of the series is. The series wants to handle ENOMEM, now you ask me if I want to remove ENOMEM handling, I'm left completely confused.

What's the problem you're solving?

include/seastar/core/reactor_config.hh
avikivity: I don't understand what LD_PRELOAD has to do with anything.

src/core/reactor_backend.cc
mykaul: I'm hoping to solve https://github.com/scylladb/seastar/issues/3583

include/seastar/core/reactor_config.hh
mykaul: It's a hack to show that if we do fail in detect_io_uring, you'd get this failure to load.

mykaul:
I think this issue needs a re-visit - I believe it's a regression due to the asymmetric IO uring probing. I'll see what I can do about it.

src/core/reactor_backend.cc
avikivity: The way to do it is to reproduce under controlled settings so you understand what the problem is. Not to guess what the problem is and confuse everyone.

It's probably related to this code:

```c++
    {
        const auto candidates = app_cfg.reactor_opts.reactor_backend.get_candidate_names();

        // We don't want ScyllaDB to run with the io_uring backend.
        // So select the default reactor backend explicitly here.
        if (std::ranges::contains(candidates, "linux-aio")) {
            app_cfg.reactor_opts.reactor_backend.select_default_candidate("linux-aio");
        } else {
            app_cfg.reactor_opts.reactor_backend.select_default_candidate("epoll");
        }

        // Leave some reserve IOCBs for scylla-nodetool and other native tool apps.
        if (std::ranges::contains(candidates, "io_uring")) {
            app_cfg.reactor_opts.reserve_io_control_blocks.set_default_value(10);
        } else {
            startlog.warn("Need to leave extra IOCBs in reserve for tools because the io_uring reactor backend is not available."
                    " Tools will fall-back to the epoll reactor backend, which requires more IOCBs to function.");
            app_cfg.reactor_opts.reserve_io_control_blocks.set_default_value(1024);
        }
    }
```

mykaul: I can't see how. Neither call/use IO_uring ?

mykaul:
v2 pushed, need to update the PR description.

src/core/reactor_backend.cc
avikivity: Maybe it's not. I now think we exhausted memlock limit.

The retry wouldn't help here, please remove it. All the complications that AI is happy to pile on make reading the code more difficult, and doesn't actually help.

include/seastar/core/reactor_config.hh
avikivity: :facepalm: yes if you inject an error, you get an error, and if you inject a transient error, then a retry helps.

src/core/reactor_backend.cc
```diff
@@ -2643,20 +2643,33 @@ reactor_backend_selector reactor_backend_selector::default_backend() {
 }
```
avikivity, line 2644: > description. Each call does a real, sometimes expensive kernel probe (e.g.

It's not expensive.

src/core/reactor_backend.cc
```diff
@@ -2643,20 +2643,33 @@ reactor_backend_selector reactor_backend_selector::default_backend() {
...
 
 std::vector<reactor_backend_selector> reactor_backend_selector::available() {
```
avikivity, line 2645: > This fixes a real, reproducible bug (see PR description), but is not
confirmed to be the root cause of the incident reported in https://github.com/scylladb/seastar/issues/3583.

I don't think it fixes a bug. It's good to merge the calls (if it can be done without adding complication), but it doesn't fix anything. In practice Seastar will create one ring per shard, and if any of those fail, the process will fail. Consolidating a few rings early in creation doesn't matter.

The cause of the incident is likely exhaustion of the memlock limit, which should be fixed by raising the limit. Seastar failing to start on ENOMEM is the correct behavior and won't be really fixed with retries.

src/core/reactor_backend.cc
```diff
@@ -2643,20 +2643,33 @@ reactor_backend_selector reactor_backend_selector::default_backend() {
...
+        if (detect_io_uring()) {
+            ret.push_back(reactor_backend_selector("io_uring"));
+            // asymmetric_io_uring requires a strict superset of what plain
+            // io_uring requires, so there's no point probing it (another
+            // real io_uring_setup(2) call, plus SQPOLL kernel worker thread
+            // creation) once the plain probe has already failed.
+            if (detect_asymmetric_io_uring()) {
+                ret.push_back(reactor_backend_selector("asymmetric_io_uring"));
+            }
+        }
```
avikivity, line 2664: This only helps if you run it on an old kernel that doesn't support the features we need for asymmetric io_uring.

So it's an improvement, saving a few microseconds during startup on old kernels. Is it worth the churn?

src/core/reactor_backend.cc
mykaul: I'm to blame for the retry, as I had assumed it is a transient issue. I'll remove it.

include/seastar/core/reactor_config.hh
mykaul: I'm open to other ideas how to reproduce. Perhaps I need the journal of the system that ran this test - I did not have it, and I don't know if we save it.

src/core/reactor_backend.cc
mykaul: I asked to do it (again not AI) for readability sake. I can remove it.

include/seastar/core/reactor_config.hh
avikivity: If you run a test that creates 3 nodes, run with --repeat 1000, and increase the parallelism to 64, you'll see it for sure (if I'm correct in blaming memlock exhaustion).

Since we consume about 32k per shard, it also means production will fail with with 250+ shards, if/when we move to io_uring, so better increase the limits there too.

avikivity: Production already raises the limit, so it's a test-only thing.

nyh:
This bug is hurting ScyllaDB's CI (latest occurance - https://github.com/scylladb/scylladb/pull/29892) where attempts to run ScyllaDB just randomly fail with an error 
```
FATAL: Exception during startup, aborting: std::invalid_argument (find_candidate(): failed to find candidate io_uring)
```
I don't now if this fix is correct or not (I'll go review it now), but if it's not we need to make a decision that it's incorrect and create a good fix.

src/core/reactor_backend.cc
nyh: The commit message overplays the "expensiveness" of the probe, and mentions but downplays the real failure scenario that causes us to want to fix this bug: If the test machines has a transient memory problem, one of the probes can decide that io_uring is not a viable option, yet a fraction of a second later a second probe can decide that io_uring will be the default option to pick :-)

Please remove the mention "(see PR description)" in the commit message. Readers of the commit message will not know which PR this is, and also don't need to go to the PR to understand. Just add a one paragraph explanation of the bug like I did here (and like you also mentioned in the middle of the long introduction paragraph - but I think the mention of this bug should be more prominent).

src/core/reactor_backend.cc
nyh: @avikivity I think it's a real bug (irrespective to whether this is the right fix). The fact you didn't understand that there is a real bug is means the commit message didn't explain it well enough.

The code currently checks which backends are allowed 3 separate times. So what can happen is that the first call decides that io_uring **cannot be used** (indeed, lack of memory is the root cause), but then, a fraction of a second later, the code wants to pick which backend to use by default and this time determines that it can use io_uring (because the memory problem subsided), so it tries to use `io_uring` and fails because we removed it earlier from the list of viable backends. So we get the error:
```
FATAL: Exception during startup, aborting: std::invalid_argument (find_candidate(): failed to find candidate io_uring)
```

A very reasonable fix for this is to just check if io_uring is allowed or not allowed one time, not three.

I agree that this might not solve the underlying problem completely: It is still possible that we will decide that io_uring is our choice, but then we to allocate things and these allocations will fail. But then, we'll get an error message on what really failed - e.g., memlock or whatever. And not some mysterious "failed to find candidate" that nobody understands how to fix. This is basically what this patch is fixing, it cannot magically cause a machine that is out of memory to work as if it's not out of memory, but it can report the correct error.

src/core/reactor_backend.cc
nyh: I don't mind this specific change (I think it makes sense) - but I believe it has nothing to do with the bug you are fixing. 

The bug you are fixing is that one call to available() returned "io_uring" in the list, and another call returned a list without "io_uring". It is unrelated to whether returning "assymetric_io_uring" or not is inconsistent with "io_uring".

nyh (approved):
I think the static variable thing is a good fix for a real bug, but there is a change (which Avi didn't like) that isn't needed and can be removed.

I agree with Avi that this does not fix the underlying problem like lack of memory - but when this problem happens I believe that now we'll get the real error instead a bizarre "can't find candidate" error.

Alternatively, after this patch it is possible you won't get any error at all because Seastar will just pick a different backend, and continue to run normally (albeit perhaps not with the optimal choice of backend).

Either behaviors is in my opinion and improvement over the current situation, and can be considered a fix to https://github.com/scylladb/seastar/issues/3583.

nyh:
> ## Root cause note
> 
> This fixes a definitively real, reproducible bug (three independent probes that can disagree with each other), confirmed via the LD_PRELOAD repro above. It is _not_ confirmed to be the root cause of the specific incident reported in #3583 (a ScyllaDB node failing to join with `host_id <missing>`) -- the link between that incident and this bug is circumstantial, not evidenced by a captured log/stack trace. So: `Refs #3583` rather than `Fixes #3583` -- please keep that issue open until/unless the connection is actually confirmed.
> 
> Refs: #3583

I think this is a mistake. If you **believe** and can **explain** (even if you can't conclusively prove it) that it fixes #3583, I think you should make it "Fixes". If the bug occurs again, we can always reopen.

PR #3586, opened 2026-08-02
core: make nested_exception transparent to {fmt} and std::rethrow_if_nested

### LINKED ISSUE #3571: nested_exception::what() discards both nested exceptions, making failures undiagnosable

avikivity:
Seastar deprecated its {fmt} formatter for std::exception_ptr in favor
of the formatter provided by {fmt}. But {fmt} (like std::rethrow_if_nested)
can only see through a nested exception via std::nested_exception, and
seastar::nested_exception neither derived from it nor exposed the wrapped
exceptions in what() -- it returned the opaque string
"seastar::nested_exception", so formatting one dropped the actual
cause entirely.

Derive nested_exception from std::nested_exception with its nested pointer
set to `inner` (the std::nested_exception base captures the currently-handled
exception, so we rethrow `inner` and copy-assign a freshly captured base
rather than poking at libstdc++ internals). {fmt} follows only that single
pointer, so render `outer` in what() as well; {fmt} appends `inner` after it,
reproducing the full chain without duplication:

    seastar::nested_exception (while cleaning up after <outer>): <inner>

Deriving from std::nested_exception has a side effect on Seastar's own
(deprecated) exception_ptr formatter in util/log.cc: it already had a
special case that renders a seastar::nested_exception as
"<inner> (while cleaning up after <outer>)", and it also has a generic
std::nested_exception case that appends the nested pointer. A
seastar::nested_exception now matches both, so `inner` was printed twice.
Skip the generic case when the exception is a seastar::nested_exception,
which has already been rendered in full.

futures_test additionally checks formatting the concrete object with {fmt}'s
own std::exception formatter, which names `outer` via what() and -- as of
{fmt} 12.2.1 (https://github.com/fmtlib/fmt/pull/4844, post-12.2.0) -- also follows the single nested
pointer and appends `inner`, reproducing the full chain much like our
formatter does. The expected string is selected on FMT_VERSION, since earlier
{fmt} stops at what().

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

Fixes https://github.com/scylladb/seastar/issues/3571

### DIFF
--- a/src/core/future.cc
+++ b/src/core/future.cc
@@ -151,15 +151,47 @@ void future_state_base::ignore() noexcept {
     }
 }
 
+static std::string describe_exception(const std::exception_ptr& e) noexcept {
+    try {
+        std::rethrow_exception(e);
+    } catch (const std::exception& ex) {
+        return ex.what();
+    } catch (...) {
+        return "unknown exception";
+    }
+}
+
 nested_exception::nested_exception(std::exception_ptr inner, std::exception_ptr outer) noexcept
-    : inner(std::move(inner)), outer(std::move(outer)) {}
+    : inner(std::move(inner)), outer(std::move(outer)) {
+    // The std::nested_exception base captures std::current_exception() in its
+    // constructor, which is not necessarily our `inner`. It exposes no setter,
+    // so make `inner` the current exception (by rethrowing it inside a handler)
+    // and copy-assign a freshly captured base, so that std::rethrow_if_nested()
+    // and {fmt} unwrap us to the original cause rather than stopping at what().
+    if (this->inner) {
+        try {
+            std::rethrow_exception(this->inner);
+        } catch (...) {
+            static_cast<std::nested_exception&>(*this) = std::nested_exception();
+        }
+    }
+    // {fmt} follows only the single std::nested_exception pointer (`inner`),
+    // which it appends after what(); render `outer` here so the full chain shows.
+    try {
+        _what = this->outer
+                ? "seastar::nested_exception (while cleaning up after " + describe_exception(this->outer) + ")"
+                : "seastar::nested_exception";
+    } catch (...) {
+        _what = "seastar::nested_exception";
+    }
+}
 
 nested_exception::nested_exception(nested_exception&&) noexcept = default;
 
 nested_exception::nested_exception(const nested_exception&) noexcept = default;
 
 const char* nested_exception::what() const noexcept {
-    return "seastar::nested_exception";
+    return _what.c_str();
 }
 
 [[noreturn]] void nested_exception::rethrow_nested() const {
--- a/include/seastar/core/future.hh
+++ b/include/seastar/core/future.hh
@@ -28,6 +28,7 @@
 #include <functional>
 #include <memory>
 #include <stdexcept>
+#include <string>
 #include <type_traits>
 #include <utility>
 #include <source_location>
@@ -45,7 +46,15 @@
 
 namespace seastar {
 
-struct nested_exception : public std::exception {
+// Derives from std::nested_exception with its nested pointer set to \ref inner
+// (see the constructor), so that std::rethrow_if_nested() and {fmt}'s exception
+// formatter unwrap to the original cause. {fmt} follows only that single nested
+// pointer, so \ref what() additionally renders \ref outer (the exception that
+// was in flight while \ref inner was being handled); {fmt} appends \ref inner
+// after it, reproducing the full "<outer> ... <inner>" chain without duplication.
+class nested_exception : public std::exception, public std::nested_exception {
+    std::string _what;
+public:
     std::exception_ptr inner;
     std::exception_ptr outer;
     nested_exception(std::exception_ptr inner, std::exception_ptr outer) noexcept;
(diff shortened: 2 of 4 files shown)


avikivity:
v2:
 - prevent both fmt's exception printer and our own trying to unroll seastar::nested_exception
 - adjust tests for minor formatting changes
 - tested against upcoming fmt 12.2.1
 - changed to class-with-private-on-top rather than struct-with-private-on-bottom

src/core/future.cc
```diff
@@ -151,15 +151,47 @@ void future_state_base::ignore() noexcept {
     }
 }
 
+static std::string describe_exception(const std::exception_ptr& e) noexcept {
```
travisdowns, line 154: This can allocate (and so can violate `noexcept`) due to the return type, maybe it's better to return a `char *`?

src/core/future.cc
```diff
@@ -151,15 +151,47 @@ void future_state_base::ignore() noexcept {
...
-    : inner(std::move(inner)), outer(std::move(outer)) {}
+    : inner(std::move(inner)), outer(std::move(outer)) {
+    // The std::nested_exception base captures std::current_exception() in its
+    // constructor, which is not necessarily our `inner`. It exposes no setter,
+    // so make `inner` the current exception (by rethrowing it inside a handler)
+    // and copy-assign a freshly captured base, so that std::rethrow_if_nested()
+    // and {fmt} unwrap us to the original cause rather than stopping at what().
+    if (this->inner) {
+        try {
+            std::rethrow_exception(this->inner);
```
travisdowns, line 173: This seems wrong. 

The seastar inner/outer terminology is inverted wrt the std one.

Let us talk unambiguously about the _first_ and _second_ exceptions to avoid existing terms. The _first_ exception is the one that occurs first, chronologically. The second exception is the one that occurs second, while throwing/handling the first.

In std world, the _first_ exception becomes the _nested_ exception, the second one is the top-level one (i.e., the one whose type can be caught directly. This is, e.g., consistent with Python, etc.

In seastar world, on the other hand, the _first_ exception is _outer_ and _second_ is _inner_.

This code, however, sets sesatar `inner` as nested, which because the terminology is backwards reverses the usual ordering of exceptions. 

The naming mixup seems fairly pervasive, also, e.g., the future doc says:

> the callback exception on top and the original future exception nested

So does "nested" refer to _inner_ or _outer_ exceptions? 99% would think _inner_ I think but actually it was outer. So perhaps this was always intended be inner and `make_nested` is just bugged? Very hard to change it now.

travisdowns (changes requested):
resolve discussion wrt reversal of inner and outer

src/core/future.cc
avikivity: Wow. I'm confused.

Maybe the way out is to deprecate seastar::nested_exception and use std::nested_exception exclusively? It requires an API_LEVEL (because I'm sure we have code that depends on it), but that's better than having to untangle this mixup.

travisdowns: We could add accessors on sesatar::nested_exception with std-aligned names, with lots of comments too I guess.

Dunno. Tough one.

PR #3587, opened 2026-08-03
memory: back shard memory with a memfd and expose it

avikivity:
Add the ability to back each shard's memory allocator with a memfd instead of anonymous memory. This lets the shard's memory be exposed to users (via memory_layout::memfd), so other components can map the same physical memory, for example into another process.

The memfd path mirrors the existing hugetlbfs path: a per-shard file is grown with ftruncate() and mapped MAP_SHARED in place, with MADV_HUGEPAGE re-applied on each mapping since a fresh memfd mapping does not inherit the bootstrap region's transparent-hugepage hint.

memfd backing is enabled by default when memfd_create() is available and either transparent hugepages are enabled for memfd (per /sys/kernel/mm/transparent_hugepage/shmem_enabled) or the reactor is overprovisioned. The new --memfd option overrides the default. It is ignored when --hugepages is given, since hugetlbfs provides its own backing file.

The motivation for this is the ability to recover contiguous memory from a bunch of 128k allocations, by mmaping those allocations using the memfd as a reference. Since a file can be allocated in multiple mappings (unlike anonymous memory), one can have a multi-megabyte contiguous mmap that points at several discontiguous 128k chunks. This contiguous mmap won't use transparent hugepages, but sometimes it's a reasonable tradeoff.

Add a helper to take advantage of this: the helper receives a range of contiguous spans, and remaps it
into a contiguous scans. The helper supports growing.

### DIFF
--- a/include/seastar/core/memory.hh
+++ b/include/seastar/core/memory.hh
@@ -162,10 +162,32 @@ size_t per_shard_memory(size_t total_memory, unsigned nr_shards);
 
 void global_setup(unsigned nr_shards);
 
+/// Returns \p true if memfd_create() is available on the running kernel.
+bool memfd_create_available();
+
+/// Returns \p true if transparent hugepages are enabled for memfd (shmem)
+/// memory on the running system, i.e., if
+/// /sys/kernel/mm/transparent_hugepage/shmem_enabled selects a value
+/// (always, within_size, advise or force) under which a memfd mapping we
+/// madvise() with MADV_HUGEPAGE can be backed by transparent hugepages.
+bool transparent_hugepages_enabled_for_memfd();
+
 }
 
+/// Configures the memory allocator for the current shard.
+///
+/// \param m NUMA-aware description of the memory assigned to this shard.
+/// \param mbind whether to bind the memory to its NUMA node.
+/// \param transparent_hugepages whether to madvise(MADV_HUGEPAGE) the memory.
+/// \param use_memfd whether to back the shard's memory with a memfd (see
+///        \ref get_memory_layout() to obtain the resulting file descriptor).
+///        Ignored when \p hugetlbfs_path is set, since hugetlbfs provides its
+///        own backing file.
+/// \param hugetlbfs_path optional path to a hugetlbfs mount used to back the
+///        shard's memory with explicit huge pages.
 internal::numa_layout configure(std::vector<resource::memory> m, bool mbind,
         bool transparent_hugepages,
+        bool use_memfd = false,
         std::optional<std::string> hugetlbfs_path = {});
 
 void configure_minimal();
@@ -353,6 +375,13 @@ public:
 struct memory_layout {
     uintptr_t start;
     uintptr_t end;
+    /// File descriptor backing the shard's memory, or disengaged if the memory
+    /// is anonymous. This is set when the memory is backed by a memfd (see the
+    /// \p use_memfd parameter of \ref configure()), letting other components
+    /// map the same physical memory (for example, into another process). The
+    /// descriptor is owned by the allocator and remains valid for the lifetime
+    /// of the shard; users that wish to keep it beyond that must dup() it.
+    std::optional<int> memfd;
 };
 
 // Discover virtual address range used by the allocator on current shard.
--- a/include/seastar/core/memory.hh
+++ b/include/seastar/core/memory.hh
@@ -25,11 +25,14 @@
 #include <seastar/core/bitops.hh>
 #include <seastar/util/backtrace.hh>
 #include <seastar/util/sampler.hh>
+#include <algorithm>
 #include <new>
 #include <cstdint>
+#include <span>
 #include <functional>
 #include <optional>
 #include <string>
+#include <utility>
 #include <vector>
 
 namespace seastar {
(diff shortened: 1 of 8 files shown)


xemul:
> The motivation for this is the ability to recover contiguous memory from a bunch of 128k allocations, by mmaping those allocations using the memfd as a reference.

Is this remapper supposed to live outside of seastar? If so and it does `void* contiguous = mmap(NULL, big_size, PROT_READ|PROT_WRITE, MAP_SHARED, layout.memfd.value(), offset);` it now has a second virtual address of the same physical pages that the seastar allocator manages and the allocator itself is completely unaware of this alias. So some coordination is needed and raw descriptor doesn't look as good abstraction to enforce or control one.

src/core/memory.cc
```diff
@@ -1384,7 +1404,8 @@ memory::memory_layout cpu_pages::memory_layout() {
     SEASTAR_ASSERT(is_initialized());
     return {
         reinterpret_cast<uintptr_t>(memory),
-        reinterpret_cast<uintptr_t>(memory) + nr_pages * page_size
+        reinterpret_cast<uintptr_t>(memory) + nr_pages * page_size,
+        memfd ? std::optional<int>(memfd->get()) : std::nullopt
```
xemul: nit: Since we support C++23 and C++26 (and thus not obliged to support C++20), `std::optional::transform` can be benefited from

avikivity: yes

avikivity:
> > The motivation for this is the ability to recover contiguous memory from a bunch of 128k allocations, by mmaping those allocations using the memfd as a reference.
> 
> Is this remapper supposed to live outside of seastar? 

yes

> If so and it does `void* contiguous = mmap(NULL, big_size, PROT_READ|PROT_WRITE, MAP_SHARED, layout.memfd.value(), offset);` it now has a second virtual address of the same physical pages that the seastar allocator manages and the allocator itself is completely unaware of this alias. So some coordination is needed and raw descriptor doesn't look as good abstraction to enforce or control one.

The idea is to steal some 128k allocations and make them contiguous, and pass them to a user that requires contiguous memory.

My example is wasmtime linear memory. See https://github.com/scylladb/scylladb/blob/9a3aba9e452a19c184ec5e8cbc21d719d1e211be/rust/wasmtime_bindings/src/memory_creator.rs#L46.

From the point of view of the seastar allocator, the memory is in use and there is no second mapping.

Other possible uses: utils::reusable_buffer.

avikivity:
v2:
 - use std::optional::transform (much nicer)
 - add a trailing comma for the above

xemul:
> The idea is to steal some 128k allocations and make them contiguous, and pass them to a user that requires contiguous memory. ... From the point of view of the seastar allocator, the memory is in use and there is no second mapping.

A-ha! Why not hide this inside seastar itself and expose via dedicated API?

avikivity:
> > The idea is to steal some 128k allocations and make them contiguous, and pass them to a user that requires contiguous memory. ... From the point of view of the seastar allocator, the memory is in use and there is no second mapping.
> 
> A-ha! Why not hide this inside seastar itself and expose via dedicated API?

Good question, we can add a contiguous_allocation helper.

```c++
class contiguous_allocation {
public:
    explicit contiguous_allocation(size_t nr_bytes);
    std::span<std::byte> view();
    void reallocate(size_t nr_bytes);
};
```

Note, excessive use on large machines can stall on mmap_semaphore. Perhaps we can make it async and push the actual mmaps to the syscall thread.

avikivity:
I added a helper. It doesn't allocate itself, to give the user maximum flexibility. If we want we can provide a higher-level helper that also allocates.

avikivity:
I found a good reason not to expose the fd. So the helper will become the only way to use the facility.

crackcomm:
> I found a good reason not to expose the fd. So the helper will become the only way to use the facility.

I guess it makes sense. Would it have to be non-contiguous for safe fd sharing? Multiple fds (possibly per shard)?
It would increase complexity greatly, but I can see use cases for IPC. I'm interested in what the reason is.

crackcomm:
It would probably be even more complex than I initially thought. The application might as well create it's own allocator when it wants to share the fd.

avikivity:
> > I found a good reason not to expose the fd. So the helper will become the only way to use the facility.
> 
> I guess it makes sense. Would it have to be non-contiguous for safe fd sharing?

Would what have to be non-contiguous?

> Multiple fds (possibly per shard)? It would increase complexity greatly, but I can see use cases for IPC. I'm interested in what the reason is.

It's actually broken. I want to use 1GB + 2MB page combos from hugetlbfs (have patch in testing). This means part of shard memory is mapped with one fd, part with another.  But hugetlbfs sub-pages cannot me mmaped, you have to map the whole thing.

crackcomm:
> Would what have to be non-contiguous?

What I meant is impossible to map contiguously whole seastar memory from one fd. I assumed the reason was related to security concerns, or, integrity of memory - like overriding memory from another process.

> It's actually broken. I want to use 1GB + 2MB page combos from hugetlbfs (have patch in testing).

I'm gonna wait and see. Is the idea to zero-copy share 2MB aligned allocations with wasm? That's what `contiguous_mapping::append` allows IIUC. I love it if that works with wasmtime.

Edit:

But why not expose the fd and allow that for IPC too?

avikivity:
> > Would what have to be non-contiguous?
> 
> What I meant is impossible to map contiguously whole seastar memory from one fd. I assumed the reason was related to security concerns, or, integrity of memory - like overriding memory from another process.

I don't know, I just haven't considered it. Maybe it is possible.

But it's best not to bake it into the API, it's too constraining.

> 
> > It's actually broken. I want to use 1GB + 2MB page combos from hugetlbfs (have patch in testing).
> 
> I'm gonna wait and see. Is the idea to zero-copy share 2MB aligned allocations with wasm? That's what `contiguous_mapping::append` allows IIUC. I love it if that works with wasmtime.
> 
> Edit:
> 
> But why not expose the fd and allow that for IPC too?

It's possible to expose a range of fds per shard, so we don't restrict the implementation (and allow multi-size hugepages), and make it optional so the application knows it cannot rely on it.

include/seastar/core/memory.hh
```diff
@@ -353,6 +375,13 @@ public:
 struct memory_layout {
     uintptr_t start;
     uintptr_t end;
+    /// File descriptor backing the shard's memory, or disengaged if the memory
+    /// is anonymous. This is set when the memory is backed by a memfd (see the
+    /// \p use_memfd parameter of \ref configure()), letting other components
+    /// map the same physical memory (for example, into another process). The
+    /// descriptor is owned by the allocator and remains valid for the lifetime
+    /// of the shard; users that wish to keep it beyond that must dup() it.
+    std::optional<int> memfd;
```
xemul, line 384: The "... and expose it" in the patch title is not relevant any longer

include/seastar/core/memory.hh
```diff
@@ -388,6 +391,104 @@ struct memory_layout {
...
+///        are to appear in the result; each range must start on, and have a
+///        size that is a multiple of, \ref page_size.
+/// \param capacity number of bytes of address space to reserve, for later
+///        growth with \ref contiguous_mapping::append(); must be page aligned.
+///        Clamped up to the total size of \p ranges.
+/// \return a mapping of the concatenation of \p ranges.
+/// \throw see \ref contiguous_mapping::contiguous_mapping() and
+///        \ref contiguous_mapping::append().
+inline
+contiguous_mapping map_contiguous(std::span<const std::span<char>> ranges, size_t capacity = 0) {
```
xemul, line 482: The std::span\<char\> as the "address range" abstraction looks very unusual, why not common iovec instead?

src/core/memory.cc
```diff
@@ -2046,6 +2049,62 @@ memory_layout get_memory_layout() {
...
+}
+
+void contiguous_mapping::append(std::span<const std::span<char>> ranges) {
+    auto layout = get_memory_layout();
+    if (!layout.memfd) {
+        throw std::runtime_error("contiguous_mapping: the shard's memory is not backed by a file"
+                " (see the --memfd option)");
+    }
+    // Validate everything before mapping anything, so a bad request leaves the
+    // mapping unchanged.
```
xemul, line 2081: That's not quite so, the loop below can throw mid-way and keep the mapping changed by ranges that succeeded

tests/unit/CMakeLists.txt
```diff
@@ -311,6 +311,10 @@ seastar_add_test (allocator
 seastar_add_app_test (alien
   SOURCES alien_test.cc)
 
+seastar_add_test (contiguous_mapping
+  SOURCES contiguous_mapping_test.cc
+  RUN_ARGS --memfd 1)
```
xemul, line 316: This obsoletes the skipping logic in tests. If memfd is not available reactor will fail to start and the test will end up with an error rather than skip

tests/unit/contiguous_mapping_test.cc
```diff
@@ -0,0 +1,268 @@
...
+            : _mem(allocate_aligned_buffer<char>(size, memory::page_size))
+            , _size(size) {
+    }
+    std::span<char> range() const noexcept { return {_mem.get(), _size}; }
+    char* data() const noexcept { return _mem.get(); }
+    size_t size() const noexcept { return _size; }
+};
+
+// Tests whether \p p can be read, without risking a fatal SIGSEGV: a syscall
+// reading from an inaccessible address fails with EFAULT instead of faulting.
```
xemul, line 74: Nice trick :) The `process_vm_readv()` could be an option too

src/core/memory.cc
```diff
@@ -2046,6 +2049,62 @@ memory_layout get_memory_layout() {
...
+    size_t total = 0;
+    for (auto range : ranges) {
+        auto start = reinterpret_cast<uintptr_t>(range.data());
+        if (start % page_size != 0 || range.size() % page_size != 0) {
```
xemul, line 2085: `|| range.size() == 0` would be good here too, otherwise ::mmap(size=0) will return EINVAL and the exception will be system one rather than std::invalid_argument

include/seastar/core/memory.hh
avikivity: It's still exposed. We may want to unexpose it.

There are other reasons to provide an fd, to share the memory with other processes. But my unproven 1GB pages patch requires multiple fds per shard, so we have to be careful.

include/seastar/core/memory.hh
avikivity: Funnily I cannot see this comment in context, only in the conversation page. Maybe github are using AI to maintain their web server.

I agree it's odd. I want to spend some time on the API surface. It's fine to let Claude implement it but the design needs more thought.

src/core/memory.cc
avikivity: Will fix.

avikivity:
Set to draft. This logically conflicts with the hugetlbfs patches (we can decide to merge both but make them mutually exclusive) and I want to think a little bit more on the API.

tests/unit/contiguous_mapping_test.cc
avikivity: Credit to Claude

PR #3588, opened 2026-08-03
build: add opt-in precompiled-header support (Seastar_PRECOMPILE_HEAD…

mykaul:
…ERS)

Add an opt-in CMake option that precompiles core/reactor.hh - the heaviest, near-universally included header - for the seastar library target, and on Clang additionally passes -fpch-instantiate-templates so templates are instantiated once in the PCH instead of in every TU.

Measured on a 16-core machine (clang 22.1.7, libstdc++ 16, Dev mode, no compiler cache): a clean build of the seastar library target drops from 26.8s to 17.6s wall (-34%); the sum of per-TU compile durations drops from 428.9s to 198.6s (-54%). As part of a ScyllaDB clean build, this is worth about 8 seconds of end-to-end wall-clock time.

The option is OFF by default: PCH interacts with ccache (users need "sloppiness = pch_defines,time_macros" in ccache.conf), and GCC's PCH support (which CMake drives the same way) is unmeasured here - the -fpch-instantiate-templates flag itself is applied only for Clang.

Verified: with the option ON, the seastar target builds cleanly on current master (Dev mode, clang); with it OFF (the default), the generated build files are identical to before this change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -280,6 +280,10 @@ option (Seastar_DPDK
   "Enable DPDK support."
   OFF)
 
+option (Seastar_PRECOMPILE_HEADERS
+  "Precompile the heaviest, near-universally included header (core/reactor.hh) to speed up building Seastar itself. Note for ccache users: requires 'sloppiness = pch_defines,time_macros' in ccache.conf."
+  OFF)
+
 option (Seastar_EXCLUDE_APPS_FROM_ALL
   "When enabled alongside Seastar_APPS, do not build applications by default."
   OFF)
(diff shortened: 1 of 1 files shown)


mykaul:
@avikivity - this is the PCH patch I discussed. I could not find in Seastar anything like that - what am I missing?

avikivity:
When embedded in an application, Seastar changes rarely and therefore is served from ccache/sccache, rarely compiled directly.

cc @travisdowns if this is useful for you

travisdowns:
> cc @travisdowns if this is useful for you

We use bazel, so only bazel caching, so this build change would not directly affect us. On the topic though, I have tried precompiled headers (within redpanda) and they do give about a 2x speedup, in my basic test but I ultimately did not go ahead because they are difficult to integrate in bazel and require several compromises (e.g., reduction of hermeticity). They also require maintenance, since you have to curate the set of included headers (on a per module basis) for the best speedup. This is easier now with AI.

As you say, typical application dev in Redpanda does not recompile seastar all that often, though for the kind of experiments I am doing, e.g., changing compiler flags, etc, everything does get recompiled a lot including seastar. However, seastar complication is off the critical path in our bazel build: the application .cc depends on the headers which are present at the start of the build, so we only depend on the seastar build at the link step. _Still_ the build is heavily parallel in general so even off the critical path it costs something since most machines can max out all cores most of the time during compilation.

I do have some additional thoughts/changes on compile time that I will share at a future time.

mykaul:
It's an opt-in, is there any downside?

avikivity:
Yes it's more stuff to maintain

PR #3589, opened 2026-08-03
github: treat deprecation warnings as errors in CI

travisdowns:
Uses of deprecated entities within the seastar tree have produced warnings without failing the build since 861da6e37 ("build: Allow deprecated declarations internally"). In practice warnings in an otherwise green build go unread, so in-tree uses of deprecated APIs accumulate silently: the fmt 12 bump surfaced dozens of them (see the fixes split out of #3559: #3568, #3575, #3579 and 37ed95ad8).

With those merged the tree builds deprecation-clean, so keep it that way by promoting the deprecation diagnostics to errors in CI, by passing the `-Werror` flags to `configure.py --cflags` in the CI build script. They land in `Seastar_CXX_FLAGS`, which CMakeLists.txt appends after its `-Wno-error` demotions, so the `-Werror` additions win. `-Werror=deprecated-declarations` is passed separately because GCC's `-Wdeprecated` group does not include it (clang's does).

Only CI is affected: a default build keeps deprecations as warnings, so compilers outside the CI matrix that diagnose additional deprecations (e.g. clang trunk reaches deprecated formatters through fmt's constexpr parse machinery that clang 22 does not diagnose) do not break developer builds. Intentional in-tree uses of deprecated APIs (implementation details and tests of deprecated code paths) already suppress the diagnostic locally with pragmas.

Note that configure.py also forwards `--cflags` to the cooked ingredients, so the fmt cook (and dpdk in the dpdk job) builds with these flags as well; they are clean under them today and CI would catch a regression there the same way.

An earlier revision of this PR, which unconditionally dropped the `-Wno-error` demotions for all builds, passed the full CI matrix green (16/16 checks), confirming the tree is currently deprecation-clean under errors.

### DIFF
--- a/.github/workflows/install-build-env.sh
+++ b/.github/workflows/install-build-env.sh
@@ -100,6 +100,12 @@ if [[ "$COMPILER" == clang++-* && "$clang_ver" -ge 20 || "$STANDARD" -ge 26 ]] ;
     cook_args=(--cook fmt)
 fi
 
+# Deprecation diagnostics are errors in CI only. The flags land in
+# Seastar_CXX_FLAGS, which is appended after the -Wno-error demotions
+# in CMakeLists.txt, so the -Werror here wins. GCC's -Wdeprecated
+# does not include -Wdeprecated-declarations, hence both.
+deprecation_errors='-Werror=deprecated -Werror=deprecated-declarations'
+
 group "configure.py"
 # OPTIONS / ENABLES intentionally unquoted: each carries multiple
 # whitespace-separated args (e.g. "--cook dpdk --dpdk-machine corei7-avx").
@@ -109,6 +115,7 @@ group "configure.py"
     --compiler "$CPP"           \
     --c-compiler "$CC"          \
     --mode "$MODE"              \
+    --cflags "$deprecation_errors" \
     "${cook_args[@]}"           \
     "${ccache_opt[@]}"          \
     $OPTIONS                    \
--- a/.github/workflows/install-build-env.sh
+++ b/.github/workflows/install-build-env.sh
@@ -100,6 +100,12 @@ if [[ "$COMPILER" == clang++-* && "$clang_ver" -ge 20 || "$STANDARD" -ge 26 ]] ;
     cook_args=(--cook fmt)
 fi
 
+# Deprecation diagnostics are errors in CI only. The flags land in
+# Seastar_CXX_FLAGS, which is appended after the -Wno-error demotions
+# in CMakeLists.txt, so the -Werror here wins. GCC's -Wdeprecated
+# does not include -Wdeprecated-declarations, hence both.
+deprecation_errors='-Werror=deprecated -Werror=deprecated-declarations'
+
 group "configure.py"
 # Both TLS backends are requested explicitly rather than left to cmake's
 # auto-detection, so that a TLS development package going missing from the
@@ -116,6 +122,7 @@ group "configure.py"
     --mode "$MODE"              \
     --enable-gnutls             \
     --enable-openssl            \
+    --cflags "$deprecation_errors" \
     "${cook_args[@]}"           \
     "${ccache_opt[@]}"          \
     $OPTIONS                    \


avikivity:
This is not something under the user's control. If a library that Seastar uses deprecates something, then it will break Seastar until it is updated.

See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=122442 for an attempt at fine-grained controls.

I think it's okay to enable -Werror=deprecated for CI so that at least CI passes (but then that slows us down when attempting to migrate CI to newer dependecies).

travisdowns:
> I think it's okay to enable -Werror=deprecated for CI so that at least CI passes (but then that slows us down when attempting to migrate CI to newer dependecies).

Yes, but (a) there are currently no instances of this at the moment, and (b) we should try for finer grained control when one does pop, I think.

I wonder if the system/user header split can be used to fix that also.

travisdowns:
Note also that `-Wno-error=deprecations` was probably doing a lot more than we want in clang: it's a group optioin so this wasn't just turning off deprecated declarations but other language-level deprecated stuff also.

travisdowns:
> I think it's okay to enable -Werror=deprecated for CI so that at least CI passes (but then that slows us down when attempting to migrate CI to newer dependecies).

I'll do it this way.

travisdowns:
On a related topic CI should really build with min/max supported version of dependencies cooked, which would catch a ton of leading/trailing 3rd party library version issues like new deprecated functions being introduced (and also many more).

travisdowns:
@avikivity wrote:

> I think it's okay to enable -Werror=deprecated for CI so that at least CI passes (but then that slows us down when attempting to migrate CI to newer dependecies).

I've done it this way. We error only in CI, where we control the inputs.

travisdowns:

xemul:
AFAIU the `--cflags` flags propagate down to cooked recipes whichi s a time-bomb. If the cxx deprecation flags were kept in Seastar_PRIVATE_CXX_FLAGS, that would mitigate it, but this requires a dedicated flag for configure.py

travisdowns:
> AFAIU the `--cflags` flags propagate down to cooked recipes whichi s a time-bomb. If the cxx deprecation flags were kept in Seastar_PRIVATE_CXX_FLAGS, that would mitigate it, but this requires a dedicated flag for configure.py

Note that this is CI and currently we cook very little in CI.

Not sure what you mean by "time bomb" the desired behavior is that when we make changes to 3rd party deps that affect CI, that same change fixes/suppresses the warning. Do you agree with that behavior?

What we don't want, I think, is is that changes outside of the codebase suddenly start breaking CI, is that what you refer to? 

For cooked versions that latter thing shouldn't happen at all: the versions are pinned in-source, so if they are clean now they should only be broken by an in-source change.

For versions picked up from the distro it can happen if the distro upgrades the version within a given release, though that seems quite uncommon? In any case we don't compile those so the scope for new warnings is less.

xemul:
> > AFAIU the `--cflags` flags propagate down to cooked recipes whichi s a time-bomb. If the cxx deprecation flags were kept in Seastar_PRIVATE_CXX_FLAGS, that would mitigate it, but this requires a dedicated flag for configure.py
> 
> Note that this is CI and currently we cook very little in CI.

True. That's why I called it a time bomb -- it's doesn't explode right now, but it's armed (see below...)

> Not sure what you mean by "time bomb" the desired behavior is that when we make changes to 3rd party deps that affect CI, that same change fixes/suppresses the warning. Do you agree with that behavior?
> 
> What we don't want, I think, is is that changes outside of the codebase suddenly start breaking CI, is that what you refer to?
> 
> For cooked versions that latter thing shouldn't happen at all: the versions are pinned in-source, so if they are clean now they should only be broken by an in-source change.

(... here). When/If we update what we cook for (either by pinning some other version, or by adding more ingredients), and the new one has internal unfixed deprecation, this flag propagation will explode. So we'll either have to pick some other version/ingredient, or patch seastar not to leak this flag (my suggestion)

> For versions picked up from the distro it can happen if the distro upgrades the version within a given release, though that seems quite uncommon? In any case we don't compile those so the scope for new warnings is less.

travisdowns:
Makes sense. I will keep the flag private.

travisdowns:
cab26c982 is a rebase onto master; the patch content is unchanged, the only conflict was positional with the --enable-gnutls/--enable-openssl lines from #3608, resolved by keeping both sets of configure flags.

PR #3590, opened 2026-08-04
net/tls: unify the put() linearization threshold across backends

travisdowns:
Both TLS backends linearize a multi-buffer packet in `put()` before handing it to the backend, but they used different thresholds: GnuTLS used the runtime `gnutls_record_get_max_size()`, while OpenSSL used a local 16 KiB "max record size" constant.

The threshold is not really about the TLS record size. Linearization is useful because it amortizes the per-call overhead of pushing many small fragments through the backend, each of which ultimately becomes a sendmsg syscall (and small fragments also produce small TLS records, which encrypt/decrypt less efficiently). How much copying is worth doing to save that overhead does not depend on the negotiated record size.

This change introduces a single shared `linearize_put_threshold` constant (16 KiB) in `tls-impl.hh`, documented as such, and uses it in both backends. For GnuTLS this is a slight behavior change only when a session negotiates a smaller max record size, in which case it previously linearized less.

This is a precursor to a more unified `put()` implementation across the two backends.

cc @StephanDollberg

### DIFF
--- a/src/net/tls_gnutls.cc
+++ b/src/net/tls_gnutls.cc
@@ -952,13 +952,11 @@ public:
         // We want to make sure that we call gnutls_record_send with as large
         // packets as possible. This is because each call to gnutls_record_send
         // translates to a sendmsg syscall. Further it results in larger TLS
-        // records which makes encryption/decryption faster. Hence to avoid
-        // cases where we would do an extra syscall for something like a 100
-        // bytes header we linearize the packet if it's below the max TLS record
-        // size.
+        // records which makes encryption/decryption faster. Hence we linearize
+        // small packets; see linearize_put_threshold for the rationale.
 
         size_t size = std::accumulate(bufs.begin(), bufs.end(), size_t(0), [] (size_t s, const auto& b) { return s + b.size(); });
-        if (size <= gnutls_record_get_max_size(*this)) {
+        if (size <= linearize_put_threshold) {
             temporary_buffer<char> linear(size);
             char* pos = linear.get_write();
             for (auto& buf : bufs) {
--- a/src/net/tls-impl.hh
+++ b/src/net/tls-impl.hh
@@ -164,6 +164,18 @@ public:
     bool _load_system_trust = false;
 };
 
+/// Threshold below which put() linearizes a multi-buffer packet into a
+/// single contiguous buffer before handing it to the TLS backend.
+///
+/// Each buffer passed to the backend becomes a separate record_send /
+/// BIO write and ultimately a separate sendmsg syscall, so coalescing
+/// small fragments (e.g. a ~100 byte header preceding a payload) avoids
+/// that per-call overhead and also produces larger TLS records, which
+/// encrypt/decrypt more efficiently. Note that this value is not tied to
+/// the TLS record size: it bounds how much we are willing to copy to
+/// save those overheads, a tradeoff independent of the record size.
+constexpr size_t linearize_put_threshold = 16 * 1024;
+
 /// Abstract interface for a TLS session.
 ///
 /// This is the primary abstraction that TLS backends (GnuTLS, OpenSSL)
--- a/src/net/tls_openssl.cc
+++ b/src/net/tls_openssl.cc
@@ -995,7 +995,6 @@ public:
     // encrypt it and then place it into the output bio.
     future<> put(std::span<temporary_buffer<char>> bufs) override {
         tls_log.trace("{} put", *this);
-        constexpr size_t openssl_max_record_size = 16 * 1024;
         if (_error) {
             return make_exception_future(_error);
         }
@@ -1020,11 +1019,10 @@ public:
         // We want to make sure that we write to the underlying bio with as large
         // packets as possible. This is because eventually this translates to a
         // sendmsg syscall. Further it results in larger TLS records which makes
-        // encryption/decryption faster. Hence to avoid cases where we would do
-        // an extra syscall for something like a 100 bytes header we linearize the
-        // packet if it's below the max TLS record size.
+        // encryption/decryption faster. Hence we linearize small packets; see
+        // linearize_put_threshold for the rationale.
         size_t size = std::accumulate(bufs.begin(), bufs.end(), size_t(0), [] (size_t s, const auto& b) { return s + b.size(); });
-        if (size <= openssl_max_record_size) {
+        if (size <= linearize_put_threshold) {
             temporary_buffer<char> linear(size);
             char* pos = linear.get_write();
             for (auto& buf : bufs) {


src/net/tls_gnutls.cc
```diff
@@ -952,13 +952,11 @@ class session : public enable_shared_from_this<session>, public tls::session_imp
...
-        // records which makes encryption/decryption faster. Hence to avoid
-        // cases where we would do an extra syscall for something like a 100
-        // bytes header we linearize the packet if it's below the max TLS record
-        // size.
+        // records which makes encryption/decryption faster. Hence we linearize
+        // small packets; see linearize_put_threshold for the rationale.
 
         size_t size = std::accumulate(bufs.begin(), bufs.end(), size_t(0), [] (size_t s, const auto& b) { return s + b.size(); });
-        if (size <= gnutls_record_get_max_size(*this)) {
+        if (size <= linearize_put_threshold) {
```
elcallio, line 959: I am not 100% on this. We do this call instead of using a constant, because the gnutls API makes this technically possible for this size to vary (thought it won't in any meaningful way - for us at least).
The reason to coalesce to this value is that the buffers used by gnutls will be this size, ergo there will be no
splitting (either implicit or by us, below in do_put_one) at this size.

What is the real benefit of adding a constant unrelated to the API? If anything, it might be the openssl path that maybe should not use a hardcoded constant?

travisdowns: It's explained a bit in the commit message: the idea of this linearization isn't about avoid splitting across the TLS record or reducing TLS records (primarily), but rather that sending many small writes into GnuTLS or OpenSSL is very expensive, since each write turns into a `sendmsg` syscall. So it is much better to linearize them. This doesn't depend on the TLS record size, really. E.g., if the TLS record size was reduced to 1K by the client, and a vector of many 2K buffers is passed in here, it would be terrible to send those 2K writes down individually: you _still_ want to send them down linearized, and the TLS library will split them into 1K TLS records, but send them all in one (or fewer) sendmsg calls.  TLS record splitting is relatively cheap (and in any case sending down larger chunks doesn't result in _more_ records, rather less or tied). sendmsg calls are expensive. So the tradeoff is really performance vs memory use, record size doesn't play into it.

Another way of looking at is this threshold should actually be "infinite": we should almost always be doing linearization in cases where the input shape makes sense, no matter the total size: that change shows very large wins, and it's coming but first we need:

1) This change
2) De-duplicate some GnuTLS and OpenSSL identical which is the reason I'm this PR in the first place since it's a divergence (see https://github.com/scylladb/seastar/pull/3591)
3) Then get rid of this threshold entirely with a different linearization strategy

Also cc @StephanDollberg who made the original findings and wrote the linearization in both GnuTLS and later OpenSSL to see if I'm way off base here or what.

travisdowns: Here's are tables to back up my claims:

Table 1 — varying the negotiated max record size, 50 × 100-byte fragments per write (5000 B). dyn = gnutls_record_get_max_size() threshold (today), cst = shared 16 KiB constant (this PR).

  | negotiated max | dyn rec/write | dyn rec size | cst rec/write | cst rec size | dyn MB/s | cst MB/s | gain |
  |---|---|---|---|---|---|---|---|
  | **16384 (default)** | **0.50** | **10017** | **0.50** | **10017** | **878** | **932** | **1.06x** |
  | 4096 | 50.00 | 117 | 1.50 | 4113 | 22 | 523 | 23.8x |
  | 2048 | 50.00 | 117 | 2.50 | 2065 | 22 | 365 | 16.6x |
  | 1024 | 50.00 | 117 | 5.00 | 1041 | 22 | 198 | 9.0x |
  | 512 | 50.00 | 117 | 10.00 | 529 | 22 | 105 | 4.8x |

Table 2 — negotiated max fixed at 1024, varying payload.

  | layout | payload/write | dyn rec/write | cst rec/write | dyn MB/s | cst MB/s | gain |
  |---|---|---|---|---|---|---|
  | 8 × 100 | 800 | 8.00 | 0.82 | 22 | 195 | 8.9x |
  | 20 × 100 | 2000 | 20.00 | 2.00 | 22 | 196 | 8.9x |
  | 50 × 100 | 5000 | 50.00 | 5.00 | 22 | 197 | 9.0x |
  | 100 × 100 | 10000 | 100.00 | 10.00 | 22 | 193 | 8.8x |
  | 160 × 100 | 16000 | 160.00 | 16.00 | 22 | 198 | 9.0x |
  | 200 × 100 | 20000 | 200.00 | 200.00 | 22 | 22 | 1.00x |

"Gain" is what this PR gives over the baseline. As you can see, in the case where the TLS record size was negotiated down, this change gives a massive speedup. Linearization is just very important for performance.

elcallio:
I don't doubt this, but given that we _also_ to _splitting_ of records (for gnutls at least) in do_put_one, based on the actual gnutls value (splitting here to avoid large allocs, and pick the max buffer size to nicely fit into the normal packaging inside gnutls -> avoid it having to re-do our work), it seems to me that in this and you #3591 it would be better to make this either a virtual call returning size (based on gnutls query), or a member var, updated on creation and handshake. 
The constant value here will of course be fine for all intents and purposes, but why not try to align with the underlying mechanism?

avikivity:
Is unification of put() across the backends really dependent on this? We can easily make the size a parameter passed by the backend, if we decide each backend deserves a different threshold for some reason.

travisdowns:
@elcallio wrote:

> I don't doubt this, but given that we _also_ to _splitting_ of records (for gnutls at least) in do_put_one

I see, I didn't understand we did this additional capping of the size in do_put_one (OpenSSL side doesn't need it), now I'm up to speed. Thanks!

> The constant value here will of course be fine for all intents and purposes, but why not try to align with the underlying mechanism?

Because it there's no need for alignment and it's the wrong thing to do, see the above tables. 

The linearization in `put` is to avoid many small calls into `gnutls_record_send`. It does this by checking the _total_ size of all buffers passed to `put` and comparing it against the linearization threshold. This threshold _should_ be infinite. It shouldn't relate to the record size. That it does is just a historical artifact.

Let me give a concrete example: someone calls put with 50 x 100 byte buffers (i.e., span.size() == 50, each buffer has 100 bytes). TLS record size for the session is 4K. Current logic compares the total size of all buffers (5000 bytes) to 4K, sees that it's bigger, and doesn't linearize, and makes 50 `gnutls_record_send` calls, and 50 `sendmsg` calls and sends <= 50 packets. Splitting doesn't kick in.

If we linearize, we send a 5K buffer to `do_put_one` which sends this as 4K then 1K to `gnutls_record_send` (due to the splitting) which is optimal (given the need to split at all), resulting in 2 sendmsg calls and <= 2 packets.

I'm only doing the change to hard-coded 16K here, instead of "infinite", because it's strictly better. The 16K value does not relate to the record size, that was just a red-herring. In a subsequent change I will remove the size limit entirely.

travisdowns:
@avikivity wrote:

> Is unification of put() across the backends really dependent on this? We can easily make the size a parameter passed by the backend, if we decide each backend deserves a different threshold for some reason.

It's not, but it's just wrong anyway, see my argument, and it should be fixed. It can be a big speed boost for folks using GnuTLS. This OG change was a large boost for RP: like XX% speed boost overall for e2e workloads (because per above the narrowly measured TLS cost can be reduced by up to 20x in realistic cases).

Especially I don't want to slow down OpenSSL backend, which is doing closer to right thing in order to coalesce the impls. 

Also, as above, I'm changing this all completely in a subsequent change. So current GnuTLS linearization is "2 out of 10", this makes it "4 out of 10" and later we get to 9 out of 10 or whatever.

elcallio:
> Let me give a concrete example: someone calls put with 50 x 100 byte buffers (i.e., span.size() == 50, each buffer has 100 bytes). TLS record size for the session is 4K. Current logic compares the total size of all buffers (5000 bytes) to 4K, sees that it's bigger, and doesn't linearize, and makes 50 `gnutls_record_send` calls, and 50 `sendmsg` calls and sends <= 50 packets. Splitting doesn't kick in.

I don't argue against linearizing. I argue against the hardcoded value. Since we later _do_ split, and even if we did not, the underlying gnutls would, it is better to linearize to the actual max buffer size. Which will be 16k in 99.9% of the time, but still should not be hard coded. 
Also, the change here does no change to the logic passing N small buffers through, so I am unclear what your argument is.

avikivity:
> @avikivity wrote:
> 
> > Is unification of put() across the backends really dependent on this? We can easily make the size a parameter passed by the backend, if we decide each backend deserves a different threshold for some reason.
> 
> It's not, but it's just wrong anyway, see my argument, and it should be fixed. 

I'm not arguing either for or against it. Just that it can be done separately (or maybe each backend will prefer a different size).

travisdowns:
> I don't argue against linearizing. I argue against the hardcoded value. Since we later do split, and even if we did not, the underlying gnutls would, it is better to linearize to the actual max buffer size.

If I understand you correctly, you are suggesting that we separate the threshold from the linearization buffer size?

That could make sense, because using the TLS record size as the threshold does not make sense per my worked examples above.

> Which will be 16k in 99.9% of the time, but still should not be hard coded.

To be very clear I'm not arguing we can just set the threshold to 16K because the TLS record size is usually 16K. I'm saying it's totally wrong to use the TLS record size here as the threshold _especially_ in cases where the size has been negotiated down. The API call is totally counter-productive in that case, and the smaller the record size the worse it gets.

travisdowns:
Closing this, as based on feedback from @elcallio this could better align with the negotiated record size. I will just jump straight to the final linearization solution on top of the implementation merging work instead of this intermediate step, as well.

Issue #474, opened 2018-07-17
tls session close does not handle all potential exceptions in worst-case shutdown


elcallio:
Can cause abandoned exception futures in reactor -> annoying log messages. 
Since fail-case `close()` is done in a background task, it should fully guard against any errors - at this point nothing can propagate to user really.

Issue #453, opened 2018-05-28
tls: Handshake should always ensure output is drained. 


elcallio:
Before ec5ca3a0e4d63ff5265c27d7c12dfd16679403d0 "Make put/push mechanism operate "opportunisticly"", all input/output was waited for explicitly before continuing. Now, however, output is considered ok until proven otherwise. 

However, even if an error is detected later in state machine, we can still have a pending output we consider "benign" that causes an exception. This leads to the dreaded "exceptional future ignored" on session delete. 
This is handled in normal IO (knock on wood), but handshake mechanism might actually miss a broken write because of a broken (later) read. 

Should fix by ensuring both waiting and error handling mechanism for this always waits for output first.

PR #3591, opened 2026-08-04
tls: hoist identical session logic into a session_base class

travisdowns:
The GnuTLS and OpenSSL session implementations duplicate the entire async orchestration layer around their respective TLS engines, and this is where the historically tricky concurrency behavior lives (#453, #474, #2859). The copies had already quietly diverged in a few places.

This is step 1 of an incremental unification: introduce `tls::session_base`, holding the shared state (socket, source/sink, read/write semaphores, options, input buffer, error/eof/shutdown flags, session type) and only the methods that were identical or near-identical in both backends. The backends implement five virtual hooks: `connected()`, `handshake()`, `do_get()`, `do_put_one()` and `do_shutdown()`.

## What moved into `session_base`, and how identical it was before

**Verbatim identical** (only formatting/`std::bind`-vs-lambda differences):

- `flush()`, `socket()`, `eof()`, `in_avail()`
- `state_checked_access()`: character-identical member template in both files.
- `shutdown()`: identical including comments; the base uses the lambda form rather than GnuTLS's `std::bind`.

**Identical logic, differing only in tracing** (OpenSSL had `tls_log` trace/debug lines, GnuTLS had none):

- `wait_for_input()`
- `wait_for_eof()`: also GnuTLS used a `repeat` loop where OpenSSL used `do_until`; these are equivalent and the base keeps the `do_until` form. OpenSSL's trailing trace-only `.finally` was dropped, the entry trace kept.

Decision for this class: the tracing is kept in the base, so the GnuTLS backend gains the logging the OpenSSL backend already had. To support that, the session identity previously private to the OpenSSL file also moves to the base: the `_local_address`/`_remote_address` capture, `get_type_string()`, the fmt formatter, and the `tls` logger definition itself.

**Near-identical, with a real decision taken:**

- `put()`: identical except for the linearization threshold: GnuTLS used `gnutls_record_get_max_size()` (dynamic), OpenSSL a hardcoded 16 KiB. Decision: fixed 16 KiB constant for both. GnuTLS's default max record size is exactly 16384, so this only differs when the peer negotiates a smaller record size, and even then the threshold only controls whether small writes are copied into one buffer; `gnutls_record_send` still chunks to the real record size inside `do_put_one()`. Worst case is an occasional extra copy, never incorrect framing.
- `do_put()` (both overloads): identical except GnuTLS asserted `_output_pending.available()` inside the semaphore. Since `_output_pending` stays backend-owned (see below), the base cannot assert it; the assert is relocated to the top of GnuTLS's `do_put_one()`, which is entered at exactly the points the old assert covered (immediately after semaphore grant, and after each `wait_for_output()` round trip), so the check is preserved, not weakened.
- `close()`: same logic, but the error handling had diverged in style. GnuTLS wraps each of `_in.close()`/`_out.close()` in try/catch plus a per-future `handle_exception` (added in 60892e8ce for #474, where the bare discarded close futures caused abandoned-exceptional-future reports and a possible half-close); OpenSSL, written six years later explicitly mirroring the already-fixed GnuTLS code, restyled it as a `.finally` chain. Decision: the base takes the `.finally` chain, which is protection-equivalent by construction: a synchronous throw inside a `finally` body is captured into the resulting future and cannot escape, the subsequent `finally` runs regardless (so the second stream is always closed, no half-close), and the trailing `handle_exception` swallows the outcome (no abandoned futures). GnuTLS's final `then_wrapped`+`ignore_ready_future()` and OpenSSL's `handle_exception().discard_result()` are likewise equivalent; the latter is kept.
- `handshake()` serialization (`do_handshake_sync`): the acquire-both-semaphores-and-record-first-error body was identical, but GnuTLS had it as a separate method taking a member-function pointer (reused by `force_rehandshake()`), while OpenSSL had the same body inlined in `handshake()` with `do_handshake()` hard-coded. Decision: the base exposes `do_handshake_sync(func)` as a template over any invocable; both backends now pass lambdas, and `handshake()` itself remains a backend hook because its system-trust preamble genuinely differs (async semaphore-guarded load vs synchronous `SSL_CTX_set_default_verify_paths`).
- GnuTLS's private `session::type` enum (values `GNUTLS_CLIENT`/`GNUTLS_SERVER`) is replaced by the common `session_type` held in the base, mapped to the GnuTLS constants at the single `gnutls_init()` call. This also deletes the enum-to-`session_type` mapping switch inside GnuTLS's `verify()`.

## What deliberately did not move

- `_output_pending`, `wait_for_output()`, `handle_output_error()`: genuinely divergent. GnuTLS uses a plain `future<>` reset on every wait; OpenSSL uses a `shared_future<>` that acts as a terminal circuit breaker after an output failure (see the comment block on the member). `handle_output_error()` additionally differs in signature (GnuTLS takes an error code, OpenSSL an exception object). Converging these is a semantic change, not a refactor, and is left for a later step. Notably, none of the moved methods touch `_output_pending` directly, so no hook was needed for it.
- `get()`: GnuTLS has an empty-buffer-means-rehandshake retry loop with no OpenSSL counterpart; giving OpenSSL that loop could spin, since its `do_get()` can return an empty non-eof buffer on `SSL_ERROR_NONE`.
- The engine-shaped code: `do_handshake`/`do_get`/`do_put_one`/`do_shutdown` state machines and error decoding, `verify()`, certificate/DN/SAN introspection, the transport callbacks (GnuTLS push/pull, OpenSSL BIO), and the destructors (which assert on the backend-owned `_output_pending`).

Net delta: +6 lines, with the two backends shedding 576 lines in favor of 487 written once (about a quarter of the new-file line count is documentation of the class, hooks and members).

Tested: `tests/unit/tls_test` passes with both `--crypto-provider gnutls` and `--crypto-provider openssl`.

### DIFF
--- a/src/net/tls-session-base.cc
+++ b/src/net/tls-session-base.cc
@@ -1,0 +1,245 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright 2015 Cloudius Systems
+ * Copyright 2024 Redpanda Data
+ */
+#include "tls-session-base.hh"
+
+#include <algorithm>
+#include <numeric>
+
+#include <fmt/ostream.h>
+
+#include <seastar/core/loop.hh>
+#include <seastar/core/reactor.hh>
+#include <seastar/core/timer.hh>
+#include <seastar/core/with_timeout.hh>
+#include <seastar/net/stack.hh>
+
+namespace seastar {
+
+logger tls_log("tls");
+
+namespace tls {
+
+session_base::session_base(session_type t, std::unique_ptr<net::connected_socket_impl> sock,
+        tls_options options)
+    : _sock(std::move(sock))
+    , _local_address([this]() -> sstring {
+        try {
+            return fmt::to_string(_sock->local_address());
+        } catch (const std::system_error&) {
+            return "DISCONNECTED";
+        }
+    }())
+    , _remote_address([this]() -> sstring {
+        try {
+            return fmt::to_string(_sock->remote_address());
+        } catch (const std::system_error&) {
+            return "DISCONNECTED";
+        }
+    }())
+    , _in(_sock->source())
+    , _out(_sock->sink())
+    , _in_sem(1)
+    , _out_sem(1)
+    , _options(std::move(options))
+    , _type(t) {
+}
+
+// Used to push unencrypted data through the TLS engine, which will
+// encrypt it and then send it down the underlying socket.
+future<> session_base::put(std::span<temporary_buffer<char>> bufs) {
+    tls_log.trace("{} put", *this);
+    // The TLS default maximum record size (2^14 bytes). This only
+    // controls whether we linearize small writes below -- the backends
+    // still chunk the actual record sends to the negotiated record size.
+    constexpr size_t max_tls_record_size = 16 * 1024;
+    if (_error) {
+        return make_exception_future<>(_error);
+    }
+    if (_shutdown) {
+        return make_exception_future<>(std::system_error(EPIPE, std::system_category()));
+    }
+    if (!connected()) {
+        tls_log.trace("{} put: not connected, performing handshake", *this);
+        std::vector<temporary_buffer<char>> p;
+        p.reserve(bufs.size());
+        p.insert(p.end(), std::make_move_iterator(bufs.begin()), std::make_move_iterator(bufs.end()));
+        return handshake().then([this, p = std::move(p)]() mutable {
+           return put(std::span(p));
+        });
+    }
+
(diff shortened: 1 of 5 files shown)


src/net/tls-session-base.cc
```diff
@@ -0,0 +1,245 @@
...
+namespace seastar {
+
+logger tls_log("tls");
+
+namespace tls {
+
+session_base::session_base(session_type t, std::unique_ptr<net::connected_socket_impl> sock,
+        tls_options options)
+    : _sock(std::move(sock))
+    , _local_address([this]() -> sstring {
```
elcallio, line 44: I realize this is in the openssl code, but to me it seems like an unwarranted waste of strings. Why not just stringize the addresses when needed? Now you are pushing this questionable practice into shared code.

src/net/tls-session-base.cc
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// Used to push unencrypted data through the TLS engine, which will
+// encrypt it and then send it down the underlying socket.
+future<> session_base::put(std::span<temporary_buffer<char>> bufs) {
+    tls_log.trace("{} put", *this);
+    // The TLS default maximum record size (2^14 bytes). This only
+    // controls whether we linearize small writes below -- the backends
+    // still chunk the actual record sends to the negotiated record size.
+    constexpr size_t max_tls_record_size = 16 * 1024;
```
elcallio, line 73: Really think this should be an impl member/method and derived from underlying provider.

elcallio (commented):
A few comments, but in general I am for this.

src/net/tls-session-base.cc
travisdowns: This maximum comes from the TLS specification and is provider independent. It is not the "current" record size of the current session, which is a negotiated value.

This relates, however, to our other discussion. Maybe I can delete it.

travisdowns:
30d31a729 is a rebase onto master with one conflict resolved: 4c8930e76 added the _alert_read member in the block where this PR removes the members hoisted into session_base; kept both. No other changes to the patch.

PR #3593, opened 2026-08-07
net/tls: name the peer as the source of received TLS alerts

mnajda-redpanda:
When the peer rejects the certificate we presented and aborts, OpenSSL renders the alert it read off the wire as `tlsv1 alert unknown ca` — as though we found a bad CA, when the peer did:

```
- Failed to establish SSL handshake: [error:0A000418:SSL routines::tlsv1 alert unknown ca]
+ Failed to establish SSL handshake: [Received TLS alert from peer: error:0A000418:SSL routines::tlsv1 alert unknown ca]
```

The reason code does not encode which side produced the alert, so an `SSL_CB_READ_ALERT` info callback records what was actually read from the peer, and only matching codes are prefixed. OpenSSL's text is kept verbatim, so log matchers keep working.

### DIFF
--- a/src/net/tls_openssl.cc
+++ b/src/net/tls_openssl.cc
@@ -41,6 +41,7 @@
 #include <cassert>
 #include <chrono>
 #include <functional>
+#include <ranges>
 #include <span>
 #include <system_error>
 #include <unordered_set>
@@ -185,6 +186,21 @@ bool contains_openssl_error(const std::vector<openssl_errc> & error_codes, int l
         return ERR_GET_LIB(static_cast<unsigned long>(code)) == lib &&
                ERR_GET_REASON(static_cast<unsigned long>(code)) == reason;
     });
+}
+
+// OpenSSL encodes a TLS alert as an ERR_LIB_SSL reason of SSL_AD_REASON_OFFSET
+// plus the one-byte alert description. ERR_GET_REASON() keeps the ERR_RFLAG_*
+// bits, so flagged reasons exceed the offset without being alerts.
+static std::optional<uint8_t> tls_alert_of_reason(openssl_errc error) {
+    const auto code = static_cast<unsigned long>(error);
+    if (ERR_GET_LIB(code) != ERR_LIB_SSL) {
+        return std::nullopt;
+    }
+    const auto alert_description = ERR_GET_REASON(code) - SSL_AD_REASON_OFFSET;
+    if (!std::in_range<uint8_t>(alert_description)) {
+        return std::nullopt;
+    }
+    return static_cast<uint8_t>(alert_description);
 }
 
 template<typename T>
@@ -795,6 +811,21 @@ public:
         if (1 != SSL_set_ex_data(_ssl.get(), SSL_EX_DATA_SESSION, this)) {
             throw make_openssl_error("Failed to set EX data for SSL session");
         }
+        // Record alerts read from the peer, so a failure caused by a received
+        // alert can be attributed to the peer when it is reported. The reason
+        // code alone cannot make that call: the state machine also raises
+        // alert-range reasons for alerts the local side is about to send.
+        SSL_set_info_callback(_ssl.get(), [](const SSL* ssl, int where, int ret) {
+            if ((where & SSL_CB_READ_ALERT) != SSL_CB_READ_ALERT) {
+                return;
+            }
+            auto* session = static_cast<openssl_session*>(
+                SSL_get_ex_data(ssl, SSL_EX_DATA_SESSION));
+            SEASTAR_ASSERT(session != nullptr);
+            // The description is ret's low byte, the part
+            // SSL_alert_desc_string_long() reads; the high byte is the level.
+            session->_alert_read = static_cast<uint8_t>(ret);
+        });
         bio_ptr in_bio(BIO_new(get_method()));
         bio_ptr out_bio(BIO_new(get_method()));
         if (!in_bio || !out_bio) {
@@ -1059,6 +1090,10 @@ public:
             tls_log.trace("{} do_handshake: already connected", *this);
             return make_ready_future<>();
         }
+        // Scope the recorded alert to this handshake attempt: a renegotiation
+        // clears SSL_is_init_finished(), so get() can drive a second one.
+        // WANT_READ and WANT_WRITE retries stay inside the loop below.
+        _alert_read.reset();
         return do_until(
             [this] { return connected() || eof(); },
             [this] {
@@ -1112,7 +1147,7 @@ public:
                                 }
                                 // Verify did not throw, fall through and make a generic error
                             }
-                            auto err = make_openssl_error("Failed to establish SSL handshake", std::move(error_codes));
+                            auto err = make_handshake_error(std::move(error_codes));
                             return handle_output_error(std::move(err));
                         }
                         default:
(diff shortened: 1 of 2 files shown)


elcallio (approved):
Not really familiar with this code, but looks good to me.

travisdowns:
@elcallio 
> travisdowns requested review from [elcallio](https://github.com/elcallio) and a balanced review from https://github.com/apps/copilot-pull-request-reviewer [now](https://github.com/scylladb/seastar/pull/3593#event-29282779547)

Ignore, it was a mis-double-click when adding copilot.

elcallio (approved):
Already reviewed. My previous statement stands.

mnajda-redpanda:
[force-push](https://github.com/scylladb/seastar/compare/676c8cd990707bc5d2b0d2cd5310932b09f6cd86..911dc5db8129b98905f8ff4a88ef616cf60ac8bd):
* applied copilot suggestion - the check is now bounded to the one-byte alert-description range in a `is_received_tls_alert()` helper, so flagged reasons such as `ERR_R_INTERNAL_ERROR` no longer match.

src/net/tls_openssl.cc
```diff
@@ -2429,8 +2433,13 @@ auto fmt::formatter<seastar::openssl_errc>::format(seastar::openssl_errc error,
     // Buffer passed to ERR_error_string must be at least 256 bytes large
     // https://www.openssl.org/docs/man3.0/man3/ERR_error_string_n.html
     std::array<char, error_buf_size> buf{};
-    ERR_error_string_n(
-        static_cast<unsigned long>(error), buf.data(), buf.size());
+    const auto code = static_cast<unsigned long>(error);
+    ERR_error_string_n(code, buf.data(), buf.size());
     // ERR_error_string_n does include the terminating null character
+    if (ERR_GET_LIB(code) == ERR_LIB_SSL && ERR_GET_REASON(code) >= SSL_AD_REASON_OFFSET) {
```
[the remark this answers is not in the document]
mnajda-redpanda: Done. The check is now bounded to the one-byte alert-description range in a `is_received_tls_alert()` helper, so flagged reasons such as `ERR_R_INTERNAL_ERROR` no longer match.

travisdowns:
> Already reviewed. My previous statement stands.

Yes, I'm sorry I accidentally retriggered your review while adding copilot due to janky GH UX.

src/net/tls_openssl.cc
```diff
@@ -1096,9 +1109,13 @@ class openssl_session : public enable_shared_from_this<openssl_session>, public
...
 
-                            if (contains_openssl_error(error_codes, ERR_LIB_SSL, SSL_R_PEER_DID_NOT_RETURN_A_CERTIFICATE) ||
-                                contains_openssl_error(error_codes, ERR_LIB_SSL, SSL_R_CERTIFICATE_VERIFY_FAILED) ||
-                                contains_openssl_error(error_codes, ERR_LIB_SSL, SSL_R_NO_CERTIFICATES_RETURNED)) {
+                            // Consult verify() when the peer certificate is at fault,
+                            // keyed on the verification result rather than on
+                            // version-dependent reason codes. The no-peer-cert codes
+                            // leave that result at X509_V_OK, so check them explicitly.
+                            if (SSL_get_verify_result(_ssl.get()) != X509_V_OK
+                                || contains_openssl_error(error_codes, ERR_LIB_SSL, SSL_R_PEER_DID_NOT_RETURN_A_CERTIFICATE)
```
travisdowns: Should this be an &&? I.e., `SSL_get_verify_result(_ssl.get()) != X509_V_OK && (contains(...PEER) || contains(...NO_CERT_RETURNED)`?

I'm wondering because before we did not call `SSL_get_verify_result` before, only checked for those 3 specific error codes: are we including more error conditions in this path now?

This is similar to/overlapping with this edge case claude dug up: https://gist.github.com/travisdowns/c00f8f5d46c98a4663e40909213d2685

src/net/tls_openssl.cc
```diff
@@ -187,6 +187,19 @@ bool contains_openssl_error(const std::vector<openssl_errc> & error_codes, int l
     });
 }
 
+// OpenSSL records an alert read from the peer as SSL_AD_REASON_OFFSET plus the
+// one-byte alert description carried on the wire. ERR_GET_REASON() retains the
+// ERR_RFLAG_* bits, so flagged reasons such as ERR_R_INTERNAL_ERROR also exceed
+// the offset without being alerts.
+static bool is_received_tls_alert(openssl_errc error) {
```
travisdowns: The premise here (an `ERR_LIB_SSL` reason in the alert range means the alert came from the peer) holds on every released OpenSSL I checked: on the 3.0 and 3.5 branches, the record layer's received-alert path is the only producer of these reasons. But it is already broken on OpenSSL master, where two sites raise alert-range reasons locally, at the point where the local side decides to *send* the alert:

- `statem_clnt.c`: `SSL_R_TLSV1_BAD_CERTIFICATE_STATUS_RESPONSE` when the client cannot parse a stapled OCSP response
- `statem_srvr.c`: `SSL_R_TLSV13_ALERT_MISSING_EXTENSION` when ECH processing rejects a ClientHello

Built against such an OpenSSL, the client that failed to parse the OCSP response would report "Received TLS alert from peer: ... bad certificate status response", the exact direction inversion this prefix is meant to prevent. It is the same kind of version-dependence the first commit removed, in a different shape.

A robust alternative: record alerts actually read off the wire with `SSL_set_info_callback` (`SSL_CB_READ_ALERT`, documented API; the session is already in ex-data slot 0) and prefix only the error code whose alert description matches. That also lets the labeling move out of `fmt::formatter<openssl_errc>`, restoring verbatim output for the formatter's other consumers, including `openssl_error_category::message()`, which currently gains the prefix in every TLS `system_error` message, not just handshake ones.

Details, evidence, and a validated patch (both new tests pass unchanged, full `tls_test` suite passes with the openssl provider): https://gist.github.com/travisdowns/6b2269efb23dbd395c253e897d2aab77

mnajda-redpanda:
[force-push](https://github.com/scylladb/seastar/compare/911dc5db8129b98905f8ff4a88ef616cf60ac8bd..e808317a1d2ab455acb52b1955bd5975ce909c87): addressed @travisdowns findings
* dropped the `SSL_get_verify_result()` gate and restored the original three reason codes: it could inherit a verification result imported from a resumed session
* direction now comes from an `SSL_CB_READ_ALERT` info callback instead of the reason-code range, since that range also covers alerts we are about to send
* squashed into a single commit, reworded the PR title and message

src/net/tls_openssl.cc
mnajda-redpanda: You're right. The incident analysis, that this change originated from, had the direction reversed: unknown_ca was sent by the client rejecting the certificate. I restored the original `verify()` gate. The direction fix is entirely in the remaining change.

src/net/tls_openssl.cc
mnajda-redpanda: Thanks for catching that! Fixed using `SSL_CB_READ_ALERT`. The recorded alert is scoped to the logical handshake, and only a matching error is labeled as received from the peer. The global `openssl_errc` formatter remains unchanged. I've used your patch and refactored it a bit :)

travisdowns (approved):

PR #3594, opened 2026-08-07
coroutine: make future await preemption customizable 

StephanDollberg:
One big difference between coroutines and continuation based code is
that `co_await` by default may preempt while an immediately available
`.then` won't.

A common pattern when acquiring a lock is the following:

```
ss::get_units(lock, 1).then([] (units) {
    ...
});
```

vs.

```
auto units = co_await ss::get_units(lock, 1);
...
```

Because of the aforementioned difference in behaviour between coroutines
and continuations the coroutine version can have a the weird side effect
where the lock is acquired and then we immediately suspend due to task
quota preemption.

While seastar offers manual work arounds for this like
`coroutine::without_preemption_check` adding this everywhere is very
cumbersome and easy to forget.

Hence, this patch introduces a trait that allows future value types to
categorically opt out of preemption.

We use a trait to allow for easy user customization and also allow for
opting in types that are not owned by the user.

In an ideal world the semaphore methods would return a custom awaiter
that would use preemption but because of the duality of future and its
use across legacy continuation style code and coroutines getting this
behaviour is difficult.

Apply this trait to `semaphore_units`.

### DIFF
--- a/include/seastar/core/semaphore.hh
+++ b/include/seastar/core/semaphore.hh
@@ -22,6 +22,7 @@
 #pragma once
 
 #include "future.hh"
+#include <seastar/core/coroutine.hh>
 #include <seastar/core/future.hh>
 #include <seastar/core/chunked_fifo.hh>
 #include <seastar/core/timer.hh>
@@ -601,6 +602,9 @@ public:
         return _n != 0;
     }
 };
+
+template<typename ExceptionFactory, typename Clock>
+struct future_await_checks_preemption<semaphore_units<ExceptionFactory, Clock>> : std::false_type {};
 
 /// \brief Take units from semaphore temporarily
 ///
--- a/tests/unit/coroutines_test.cc
+++ b/tests/unit/coroutines_test.cc
@@ -70,6 +70,11 @@ using namespace seastar::tests;
 
 using namespace std::chrono_literals;
 
+struct future_await_without_preemption_check_type {};
+
+template <>
+struct seastar::future_await_checks_preemption<future_await_without_preemption_check_type> : std::false_type {};
+
 namespace {
 
 future<int> old_fashioned_continuations() {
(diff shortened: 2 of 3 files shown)


include/seastar/core/semaphore.hh
```diff
@@ -602,6 +603,9 @@ public:
     }
 };
 
+template<typename ExceptionFactory, typename Clock>
```
travisdowns, line 606: This could use a comment like: preemption check happens _after_ obtaining units which has the unfortunate side effect of sometimes suspending immediate after getting units, which is not ideal as ....

(I think other types, should they be added in the future, should also get a similar comment explaining why)

tests/unit/coroutines_test.cc
```diff
@@ -328,6 +333,22 @@ SEASTAR_TEST_CASE(test_no_preemption) {
     co_return;
 }
 
+SEASTAR_TEST_CASE(test_future_await_without_preemption_check) {
+    bool yielded = false;
+    auto f = yield().then([&yielded] {
```
travisdowns, line 338: This is clever, a "yield detector". `yielded` being true is a sufficient but not necessary condition for a yield to have actually happened, but seems good enough.

I've often thought the engine should just a expose a "poll count" which increments every time though the polling loop which would make checks like this easy.

travisdowns:
Overall this seems reasonable. We guide prememption to places where it will be less impactful. A bit of the same vibes as `PR_RSEQ_SLICE_EXTENSION_SET`.

I guess the only problem I can see if you have a long hot loop, where `co_await ... ss::get_units` is the _only_ suspension point, now this loop may stall, but that seems very unlikely.

tests/unit/coroutines_test.cc
StephanDollberg: (copied from existing tests in there btw)

avikivity:
I don't think it's good to customize this based on the type.

- it's invisible to the casual reader
- functions downstream can return the same type and be infected with the same trait

It needs to stand out more.

avikivity:
One thing we can do is add configuration to switch the default of whether co_await checks for preemption (requiring every loop to co_await maybe_yield). This is a system-wide default, not per-type.

nyh:
One request: The question of whether or not certain things (co_await, then(), etc.) do a preemption check or not, and whether a manual maybe_yield is needed or not, is really hard to remember. And over the last decade, this behavior actually changed a bit. So whatever you end up changing, please make sure that you find all the relevant documentation and update it to be fully up-to-date with the new situation.

StephanDollberg:
> it's invisible to the casual reader

> It needs to stand out more.

I mean that's kinda the point of the whole thing. For explicit opt-in the wrapper already exists. 

Anyway seems like this is not the direction we want to take.

travisdowns:
> (requiring every loop to co_await maybe_yield)

Well not _every_ loop, but any loop that could run long enough to be a problem, yes. Seems like an interesting option to have.

PR #3595, opened 2026-08-10
Fix tcp cwnd bug 1015

### LINKED ISSUE #1015: Incorrect handling of congestion window in native TCP stack
Anh Dũng Phan reported on [this mailing-list thread](https://groups.google.com/g/seastar-dev/c/9jNZ5xbDVdw/m/DY64hjPTAAAJ) a bug in Seastar's native TCP stack's handling of the congestion window. In `include/seastar/net/tcp.hh` in the `can_send()` function, we have:

```c++
    // Can not send more than congestion window allows
    x = std::min(_snd.cwnd, x);
```

But the congestion window is not supposed to be a limit on individual sends - but rather a limit on the total unacknowledged bytes (which in the code is in the `window_used` variable). The code therefore should be something like this:
```c++
    if (window_used > _snd.cwnd) {
        return 0;
    }
    x = std::min(_snd.cwnd - window_used, x)
```

The `if` handles the case where the congestion window is full, and we can't send anything yet, and the `std::min` handles the case where we can send something - but not the entire number of bytes we wanted to send.

Note how `_snd.window` gets exactly the same handling a few lines above.

Anh Dũng Phan explained why fixing this is important:
> This is huge because in that case, all of these things: slow start, TCP congestion control and fast recovery are non-effective, since they rely heavily on CWND value. This makes the whole TCP stack extremely inefficient in case of non-ideal network connection. When congestion happens, it will create even more congestion by keep pumping more TCP segments into the network. Needless to say this is a huge violation of RFC5681.

ankitxvx:
Fixes #1015

### Description
This PR addresses a bug in the native TCP stack's handling of the congestion window (`cwnd`).

Previously, in the `can_send()` method of `include/seastar/net/tcp.hh`, the congestion window was incorrectly used as a limit on the individual sends. The congestion window is supposed to limit the total unacknowledged bytes, not the size of individual segments.

This change ensures that we properly evaluate the unacknowledged bytes (`window_used`) against the `cwnd`, properly handling the scenarios where the window is full, as well as where we can send segments up to the limit of the congestion window.

### Changes
* `include/seastar/net/tcp.hh`: Updated `can_send()` to properly evaluate `_snd.cwnd` minus `window_used` when calculating `x` (the size allowed to send).

### Impact
Resolves a violation of RFC5681 which could cause slow start, TCP congestion control, and fast recovery to be non-effective and generate more congestion in non-ideal network connections.

### DIFF
--- a/include/seastar/http/httpd.hh
+++ b/include/seastar/http/httpd.hh
@@ -130,6 +130,7 @@ class http_server {
     uint64_t _requests_served = 0;
     uint64_t _read_errors = 0;
     uint64_t _respond_errors = 0;
+    uint64_t _tls_handshake_errors = 0;
     shared_ptr<seastar::tls::server_credentials> _credentials;
     sstring _date = http_date();
     timer<> _date_format_timer { [this] {_date = http_date();} };
@@ -199,6 +200,7 @@ public:
     uint64_t requests_served() const;
     uint64_t read_errors() const;
     uint64_t reply_errors() const;
+    uint64_t tls_handshake_errors() const;
     // Write the current date in the specific "preferred format" defined in
     // RFC 7231, Section 7.1.1.1.
     static sstring http_date();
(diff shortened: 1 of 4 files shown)


ankitxvx:
Closed — this PR accidentally included unrelated changes from another branch.
Reopened as a clean PR in #3596.

PR #3599, opened 2026-08-12
io_queue: Make io_sink non-moveable

xemul:
It usually sits on reactor and thus doesn't move naturally. However, tests may accidentally std::move one, and it's better be prevented compil-time


xemul:
Extracted from #3295 not to rebase it over and over

travisdowns (approved):

PR #3600, opened 2026-08-12
CI: run the tests on every reactor backend in Github Actions

travisdowns:
We currently tests on only the aio reactor backend in CI. Previously this used to be "only io uring" (the default arbitrarily flipped when we starting running in containers).

This results in other backends rotting. E.g., currently the async iouring backend has a few bugs which are caught by tests, but not by CI since this doesn't run there.

This series then does 3 things:

 - Allows reactor command line arguments to be set by environment variable. I think this is independently useful and have wanted this several times (to "tunnel" through layers of wrappers and scripts, especially when seastar arguments need to be separated from the application arguments with `--`), but it's included here since we need it to change the backend reliably in tests, `ctest` doesn't provide a way unfortunately. 
 - The plumbing in test.py to allow specifying the backend (uses the above)
 - The adjustment to the CI matrix to actually run across the 3 working backends. This only bumps our matrix size from 13 to 16 due to the magic of "allpairs".


test.py
```diff
@@ -67,9 +75,24 @@ def run_tests(mode):
...
+        # Seastar reads SEASTAR_<OPTION> from the environment, which a test's own
+        # arguments override, so this leaves the tests which pin a backend on the
+        # one they asked for. It also reaches the helper binaries that some tests
+        # spawn themselves, which arguments added here would not.
+        TEST_ENV = dict(os.environ)
+        if args.reactor_backend:
+            TEST_ENV['SEASTAR_REACTOR_BACKEND'] = args.reactor_backend
+        cpuset = args.async_workers_cpuset
+        if not cpuset and args.reactor_backend == 'asymmetric_io_uring':
+            # That backend refuses to start without CPUs for its async workers.
```
MrD4rkne, line 87: I would make sure if there is more than 1 core available if so. Running anything on the same core as `SQ_POLL` thread is a bad idea

travisdowns: Do you want a hard fail here if number of CPUs == 1? 

FWIW it's important for testing purposes that this works with any number of cores, even if it's not ideal or whatever, along the lines of the existing --overprovisioned mode.

MrD4rkne: Hm. Any configuration where shards overlap with sq_thread s might produce really unreliable results.

By default, sq_thread (being a kernel thread) will try to take as much of the cpu as possible, so do shards. 

You may just observe: shards' stalls, really slow I/O.

Maybe it's worth allowing it now. 

If it starts to fail, then I would recommend marking the test as "inconclusive" for single core machines.

travisdowns: > By default, sq_thread (being a kernel thread) will try to take as much of the cpu as possible, so do shards.

I'm not yet fully up on the design, but is sq_thread always doing 100% hot polling? In overprovisioned mode shards definitely do not take up 100% of the CPU, they sleep quickly when there is no work to do.

> Maybe it's worth allowing it now.

OK, thanks: for now I'll just see how it works out. In any case, CI has 4 threads so this is mostly a "theoretical discussion for the future".

avikivity: We can change the backend so that when --overprovisioned (to be renamed --overcommitted) is specified, it configures SQPOLL to go to sleep quickly. It doesn't matter for tests anyway.

src/core/app-template.cc
```diff
@@ -45,6 +48,29 @@ namespace bpo = boost::program_options;
 
 using namespace std::chrono_literals;
 
+// Translates an environment variable name into the option it sets: SEASTAR_ is
+// stripped, the rest is lowercased and underscores become hyphens, so
+// SEASTAR_REACTOR_BACKEND names --reactor-backend. Every option is spelled with
+// hyphens, so the translation is unambiguous.
+//
```
avikivity: There is boost::program_options::parse_environment()

We can also have SEASTAR_OPTIONS="--reactor-backend blah"

Not saying either is better, just that they should be considered.

avikivity: oh, you are using parse_environment.

test.py
```diff
@@ -33,6 +34,13 @@
     parser.add_argument('--timeout', action="store",default="300",type=int, help="timeout value for test execution")
     parser.add_argument('--jenkins', action="store",help="jenkins output file prefix")
     parser.add_argument('--smp', '-c', action="store",default='2',type=int,help="Number of threads for multi-core tests")
+    parser.add_argument('--reactor-backend', action="store", default='',
+                        help="Run the tests with this reactor backend (io_uring, asymmetric_io_uring, "
+                             "linux-aio or epoll) instead of the one the reactor picks by default. Tests "
+                             "which pass --reactor-backend themselves are left on their own choice.")
+    parser.add_argument('--async-workers-cpuset', action="store", default='',
+                        help="CPUs for the asymmetric_io_uring async workers, which that backend requires. "
+                             "Defaults to the highest CPU id this process may run on.")
```
avikivity, line 43: We should auto-detect this (perhaps only with --overprovisioned). Shouldn't block the series.

avikivity:
Looks good.

src/core/app-template.cc
travisdowns: We are already using `bpo::parse_environment` though perhaps not in the way you are thinking. We can't use the generic prefix version like `parse_environment(desc, "SEASTAR_")` which avoids the custom mapper, since we use hyphen delimited arguments but this prefix version can't handle that.


> SEASTAR_OPTIONS="--reactor-backend blah"

Evaluating this.

test.py
travisdowns: You mean inside seastar, right? I guess as you say it only makes sense for `--overprovisioned` as requiring the core to use for SQ_POLL to come from "outside" is presumably a hard requirement for a production deployment where tuning happened.

src/core/app-template.cc
travisdowns: I went with `SEASTAR_OPTIONS`: it has the advantage of allowing flags which need to appear multiple times to work properly and is probably easier overall to use as you can just slam the entire command line in there. It removes the need for the mapper code as well.

test.py
avikivity: > You mean inside seastar, right? 

Yes.

> I guess as you say it only makes sense for `--overprovisioned` as requiring the core to use for SQ_POLL to come from "outside" is presumably a hard requirement for a production deployment where tuning happened.

Yes, also how would Seastar know to pick a core?

With the incorrectly name --overprovisioned, we let the core affinity float.

Not sure we can do that with the asymmetric backend, probably it requires dedicating a core and even with sq_thread_idle=0. 

Or perhaps, just remove SQPOLL with the asymmetric backend when --overprovisioned. It should work the same way, just have different performance characteristics.

travisdowns: > Not sure we can do that with the asymmetric backend, probably it requires dedicating a core and even with sq_thread_idle=0.

In overprovisioned we can just remove the `SQ_AFF` but keep using `SQPOLL`. So then --overprovisioned means "all shard threads and the async polling thread float". The only question is do we create N shards or N-1 shards on an N CPU box?

I can PR this separately.

travisdowns:
@scylladb/seastar-maint please consider

.github/workflows/gen-matrix.py
```diff
@@ -149,9 +172,15 @@ def _build_include(items: list[dict[str, Any]]) -> CommentedSeq:
...
+        # BACKENDS leads because the generator varies the earlier dimensions
+        # more systematically. Leading with it gives 12 rows which are 12
+        # distinct configurations; trailing it gives 12 rows covering only 8,
+        # the other 4 being a config already present under another backend.
+        # Either way pairwise coverage is complete, and 12 is the floor here
+        # (4 compilers x 3 modes, and 4 compilers x 3 backends).
+        for b, c, s, m, a in AllPairs(
+            [BACKENDS, COMPILERS, STANDARDS, MODES, ARCHS],
             filter_func=_keep_row,
         )
```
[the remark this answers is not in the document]
travisdowns: This is true, it's due to https://github.com/thombashi/allpairspy/issues/2.

I'm going to switch away from allpairs, but I will do it in a different PR.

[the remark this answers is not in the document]
travisdowns: https://github.com/scylladb/seastar/pull/3614

test.py
avikivity: > > Not sure we can do that with the asymmetric backend, probably it requires dedicating a core and even with sq_thread_idle=0.
> 
> In overprovisioned we can just remove the `SQ_AFF` but keep using `SQPOLL`. So then --overprovisioned means "all shard threads and the async polling thread float". The only question is do we create N shards or N-1 shards on an N CPU box?

N shards. Typically N will come from --smp. Even if not, --overprovisioned says "the reactors will mostly be sleepy so optimize for CPU cycles and don't poll", meaning thread-per-core is off and adding an additional thread isn't a problem.

> 
> I can PR this separately.

Thanks

src/core/app-template.cc
```diff
@@ -187,11 +187,25 @@ app_template::run_deprecated(int ac, char ** av, std::function<void ()>&& func)
...
+        if (const char* env_options = std::getenv("SEASTAR_OPTIONS")) {
+            // Split the way a shell would, so a value containing spaces can be
+            // quoted. The empty positional description rejects anything which is
+            // not an option: this sets options, it does not stand in for a
+            // command line. Without it a stray word is silently dropped.
+            bpo::store(bpo::command_line_parser(bpo::split_unix(env_options))
+                        .options(all_opts)
+                        .positional(bpo::positional_options_description())
+                        .run()
+                , configuration);
```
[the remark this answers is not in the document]
avikivity: Agree with copilot.

I also have an unpublished patch to expose what source an option came from, it will have to be adjusted to support this too.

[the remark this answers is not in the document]
travisdowns: Testing added in https://github.com/scylladb/seastar/pull/3600/commits/2b7b125d00e7ecc470785b0465a3ca9dcef04e3c.

test.py
travisdowns: Seems like SQPOLL already goes to sleep quickly by default, at 5ms. I guess overprovisioned could be "even quicker", like 0ms.

src/core/app-template.cc
travisdowns: In as SEASTAR_OPTIONS, and now covered by tests in 2b7b125d0.

travisdowns:
@scylladb/seastar-maint feedback addressed, have a look.

doc/tutorial.md
```diff
@@ -1931,6 +1931,33 @@ TODO: list and explain more of these options.
...
+
+The two configuration files use the `name = value` form, one option per line:
+
+```
+reactor-backend = epoll
+```
+
+The files are optional, and are only read when `HOME` is set. They accept Seastar's own options; `SEASTAR_OPTIONS` and the command line additionally accept the application's own options, as described below.
+
+The first place which sets an option wins it, so a command line argument overrides `SEASTAR_OPTIONS`, which overrides the files. An option set in more than one place is not an error and produces no warning: the lower-precedence value is simply ignored. Setting the same option twice in one place *is* an error, unless that option is one which may be repeated.
```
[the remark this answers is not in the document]
travisdowns: Fair, but we have no such options in the source at the moment. Will add a source comment but too obscure to clutter the tutorial.

src/core/app-template.cc
```diff
@@ -187,11 +187,35 @@ app_template::run_deprecated(int ac, char ** av, std::function<void ()>&& func)
 
     bpo::variables_map configuration;
     try {
+        // bpo::store() keeps the first value it is given for an option, so the
+        // order of these three is their precedence: the command line beats
+        // SEASTAR_OPTIONS, which beats the configuration file.
```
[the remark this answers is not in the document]
travisdowns: Yes, true (same nit as above), will update the comment.

avikivity:
GHA!

test.py
avikivity: Yes, 5ms is wasteful if another SQPOLL thread (or an application thread) could take its place.

avikivity:
Looks good, but don't know if you promised copilot some change.

travisdowns:
> GHA!

Pronounced as gahhh

travisdowns:
> Looks good, but don't know if you promised copilot some change.

I did, and I did it, but did not reply, it's in the last push: https://github.com/scylladb/seastar/compare/2b7b125d00e7ecc470785b0465a3ca9dcef04e3c..a8f51cad7d302dd1a8a866f53e004a2854cfd28b

doc/tutorial.md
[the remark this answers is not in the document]
travisdowns: Added: https://github.com/scylladb/seastar/compare/2b7b125d00e7ecc470785b0465a3ca9dcef04e3c..a8f51cad7d302dd1a8a866f53e004a2854cfd28b

travisdowns:
https://github.com/scylladb/seastar/commit/6b01bfa548bd455a1751b3977aafcad1ba7ad86a is a pure rebase

travisdowns:
https://github.com/scylladb/seastar/commit/f4c384d3f740caad57aef108346455d7ca40d62e is a pure rebase

travisdowns:

travisdowns:
@scylladb/seastar-maint please review and/or merge

PR #3602, opened 2026-08-12
core: fix iovec lifetime in asymmetric_io_uring writev

### LINKED ISSUE #3601: asymmetric_io_uring: pipe and PTY reads return corrupted data, pipe_stream test fails
Both cases of `tests/unit/pipe_stream_test` fail under `--reactor-backend asymmetric_io_uring`, and only under that backend.

The reader gets back the correct *number* of bytes, but the content is binary garbage rather than the payload that was written, so this looks like corruption rather than a short read.

## Test output

```
tests/unit/pipe_stream_test.cc(56): fatal error: in "pipe_stream_roundtrip":
  critical check sstring(buf.get(), buf.size()) == payload has failed
  [\x0d  P   } P  \x0d != Hello, pipe streams!]
tests/unit/pipe_stream_test.cc(115): fatal error: in "pipe_pty_stream_roundtrip":
  critical check sstring(buf.get(), buf.size()) == payload has failed
  [\x0d  P   } P  \x0d != Hello, PTY streams!]
*** 4 failures are detected in the test module "Master Test Suite"
```

## Scope

- Fails: `asymmetric_io_uring`. Deterministic, every run, about 0.6s.
- Passes: `io_uring`, `linux-aio`, `epoll`.
- Reproduces on current master (449afe9c9).

## Reproducer

```bash
./configure.py --mode=dev
cmake --build build/dev --target test_unit_pipe_stream

cd build/dev/tests/unit
./pipe_stream_test -- -c 2 --reactor-backend asymmetric_io_uring \
    --async-workers-cpuset $(( $(nproc) - 1 ))
```

`--async-workers-cpuset` is required, otherwise the backend aborts at startup with
"No CPUs specified for asymmetric_io_uring workers".

## Environment

Ubuntu, kernel 6.8.0-137-generic, clang 22.1.8, liburing 2.5, 32-core x86-64, `dev` mode.

## Notes

Surfaced by #3600, which runs the unit test suite against each reactor backend in CI. The
failure itself does not depend on that branch and reproduces on plain master.

travisdowns:
Pipe and PTY writes under `asymmetric_io_uring` delivered the right number of bytes with
garbage content, failing both cases of `pipe_stream_test` on that backend only.

`reactor_backend_asymmetric_uring::writev` built its `io_request` from a function-local
vector. `io_request` refers to the iovec array by pointer and `submit_request` only queues
the request, so the SQE was flushed from the poll loop after that vector had been destroyed,
leaving the kernel to read iovec descriptors out of freed memory.

The fix gives the iovecs the lifetime of the operation by storing them in the completion
descriptor, as `recvmsg_completion_base` already does. Only this backend was affected; the
others poll and then perform the syscall inline, where the caller's iovecs are still alive.

Also adds a large-payload pipe roundtrip test: every existing case writes a single small
iovec in one go, so the partial-write path had no coverage. It fails without the backend
change and passes with it.

Tested on a 32-core x86-64 box, `dev` mode, kernel 6.8:

- `pipe_stream_test` passes on all four backends; 3/3 repeats on `asymmetric_io_uring`.
- No regressions under `asymmetric_io_uring` in `fstream`, `socket`, `connect`,
  `output_stream`, `input_stream`, `websocket`.

Fixes #3601

### DIFF
--- a/include/seastar/core/internal/io_request.hh
+++ b/include/seastar/core/internal/io_request.hh
@@ -140,6 +140,8 @@ public:
         return req;
     }
 
+    // No copy of iov is made: the returned request points at iov's data(), whose lifetime must extend
+    // until the request is complete.
     static io_request make_readv(int fd, uint64_t pos, std::vector<iovec>& iov, bool nowait_works) {
         io_request req;
         req._readv = {
@@ -212,6 +214,8 @@ public:
         return req;
     }
 
+    // No copy of iov is made: the returned request points at iov's data(), whose lifetime must extend
+    // until the request is complete.
     static io_request make_writev(int fd, uint64_t pos, std::vector<iovec>& iov, bool nowait_works) {
         io_request req;
         req._writev = {
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -1615,6 +1615,8 @@ protected:
         ::msghdr _mh = {};
         const size_t _to_write;
     public:
+        // iovs is not copied, the span data must remain valid until the
+        // associated request completes.
         sendmsg_completion_base(std::span<iovec> iovs, size_t to_write)
             : _to_write(to_write) {
             _mh.msg_iov = iovs.data();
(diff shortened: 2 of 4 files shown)


travisdowns:
Superseded by #3604, same commits from an in-repo branch.

PR #3603, opened 2026-08-12
core: cap asymmetric_io_uring writev at IOV_MAX iovecs

travisdowns:
Our existing write paths clamp writev type writes at IOV_MAX, returning a 
short write if you have more... except the new async ioruing backend, which
will return EINVAL.

Fix it: copies at most `IOV_MAX` iovecs into the completion descriptor, matching
`sendmsg_completion_base`, and states the short-write semantic on the
`reactor_backend` interface so it is a property of the interface rather than a
coincidence of each implementation. No data is lost: `write_all` trims the
completed prefix and resubmits.

Two new test cases cover it, both backend-agnostic: an oversized `write_some` is a
short write of exactly the first `IOV_MAX` iovecs, and `write_all` over the same
span still delivers everything. Each iovec holds a distinct byte so the payload
identifies which iovecs were submitted, and the total stays under `PIPE_BUF` so
the pipe write is atomic and the byte counts are exact. Verified on epoll,
linux-aio, io_uring and asymmetric_io_uring; without the clamp both fail with
`EINVAL` under asymmetric_io_uring.

Note the tests only exercise the backend that diverged when run with
`--reactor-backend=asymmetric_io_uring --async-workers-cpuset=<cpus>`; under the
default backend they guard the documented semantic only. A parallel change
adds async backend to the CI.

### DIFF
--- a/include/seastar/core/internal/io_request.hh
+++ b/include/seastar/core/internal/io_request.hh
@@ -34,6 +34,14 @@ extern logger io_log;
 
 namespace internal {
 
+// A trivially-copyable description of one kernel I/O operation. It refers to
+// buffers, iovec arrays and socket addresses by pointer and never owns them:
+// whoever creates a request must keep the pointed-to storage alive, and
+// unmodified, until the operation completes, not merely until the request is
+// built or submitted. Requests are typically queued (see io_sink) and reach
+// the kernel later, from a different stack. See iovec_keeper and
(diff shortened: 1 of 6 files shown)


travisdowns:
cc @MrD4rkne

travisdowns:
Superseded by #3605, which is hosted on a scylladb branch and stacked on #3604 so it carries just the one commit.

PR #3604, opened 2026-08-12
core: fix iovec lifetime in asymmetric_io_uring writev

### LINKED ISSUE #3601: asymmetric_io_uring: pipe and PTY reads return corrupted data, pipe_stream test fails

travisdowns:
Pipe and PTY writes under `asymmetric_io_uring` delivered the right number of bytes with
garbage content, failing both cases of `pipe_stream_test` on that backend only.

`reactor_backend_asymmetric_uring::writev` built its `io_request` from a function-local
vector. `io_request` refers to the iovec array by pointer and `submit_request` only queues
the request, so the SQE was flushed from the poll loop after that vector had been destroyed,
leaving the kernel to read iovec descriptors out of freed memory.

The fix gives the iovecs the lifetime of the operation by storing them in the completion
descriptor, as `recvmsg_completion_base` already does. Only this backend was affected; the
others poll and then perform the syscall inline, where the caller's iovecs are still alive.

Also adds a large-payload pipe roundtrip test: every existing case writes a single small
iovec in one go, so the partial-write path had no coverage. It fails without the backend
change and passes with it.

Tested on a 32-core x86-64 box, `dev` mode, kernel 6.8:

- `pipe_stream_test` passes on all four backends; 3/3 repeats on `asymmetric_io_uring`.
- No regressions under `asymmetric_io_uring` in `fstream`, `socket`, `connect`,
  `output_stream`, `input_stream`, `websocket`.

Fixes #3601

Note that sanitizer doesn't catch this clearly because the UAF has the "U" side in the kernel, invisible to the sanitizers.

### DIFF
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -1615,6 +1615,8 @@ protected:
         ::msghdr _mh = {};
         const size_t _to_write;
     public:
+        // iovs is not copied, the span data must remain valid until the
+        // associated request completes.
         sendmsg_completion_base(std::span<iovec> iovs, size_t to_write)
             : _to_write(to_write) {
             _mh.msg_iov = iovs.data();
--- a/src/core/reactor_backend.cc
+++ b/src/core/reactor_backend.cc
@@ -1628,6 +1628,20 @@ protected:
         }
     };
 
+    // io_request refers to the iovec array by pointer, and requests are submitted
+    // asynchronously, so the array must live as long as the operation rather than
+    // as long as the call that builds it.
+    class writev_completion_base : public sized_promise_completion_base {
+    protected:
+        std::vector<iovec> _iov;
+    public:
+        explicit writev_completion_base(std::span<iovec> iovs)
+            : _iov(iovs.begin(), iovs.end()) {}
+        std::vector<iovec>& iovs() noexcept {
+            return _iov;
+        }
+    };
+
     class send_completion_base : public sized_promise_completion_base {
     protected:
         const size_t _to_write;
@@ -2535,10 +2549,9 @@ public:
     }
 
     virtual future<size_t> writev(pollable_fd_state& fd, std::span<iovec> iovs) override {
-        auto desc = std::make_unique<sized_promise_completion_base>();
+        auto desc = std::make_unique<writev_completion_base>(iovs);
         const uint64_t position_file_offset = -1;
-        std::vector<iovec> iov(iovs.begin(), iovs.end());
-        auto req = internal::io_request::make_writev(fd.fd.get(), position_file_offset, iov, false);
+        auto req = internal::io_request::make_writev(fd.fd.get(), position_file_offset, desc->iovs(), false);
         return submit_request(std::move(desc), std::move(req));
     }
 };
--- a/src/core/reactor_backend.hh
+++ b/src/core/reactor_backend.hh
@@ -228,6 +228,13 @@ public:
     virtual future<> poll_rdhup(pollable_fd_state& fd) = 0;
     virtual void forget(pollable_fd_state& fd) noexcept = 0;
 
+    // For all the data-path methods below, buffer and iovec-array arguments
+    // are borrowed: the caller must keep them alive and unmodified until the
+    // returned future resolves. This includes the iovec array itself, not
+    // just the buffers it points at. Polling backends read the array before
+    // the future resolves anyway, but backends that submit asynchronously
+    // (asymmetric_io_uring) may hand it to the kernel after the calling
+    // function has returned.
     virtual future<std::tuple<pollable_fd, socket_address>>
     accept(pollable_fd_state& listenfd) = 0;
     virtual future<> connect(pollable_fd_state& fd, socket_address& sa) = 0;
--- a/include/seastar/core/internal/io_request.hh
+++ b/include/seastar/core/internal/io_request.hh
@@ -140,6 +140,8 @@ public:
         return req;
     }
 
+    // No copy of iov is made: the returned request points at iov's data(), whose lifetime must extend
+    // until the request is complete.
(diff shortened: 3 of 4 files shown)


travisdowns:
cc @MrD4rkne (can't add you as reviewer due to perms)

travisdowns:
@scylladb/seastar-maint please take look, probably @xemul ?

src/core/reactor_backend.cc
```diff
@@ -2535,10 +2549,9 @@ class reactor_backend_asymmetric_uring final : public reactor_backend_uring_base
     }
 
     virtual future<size_t> writev(pollable_fd_state& fd, std::span<iovec> iovs) override {
-        auto desc = std::make_unique<sized_promise_completion_base>();
+        auto desc = std::make_unique<writev_completion_base>(iovs);
```
xemul, line 2552: That's weird. Other backedns call reactor::do_writev() which doesn't care about keeping what iovs point to across its preemption point:

```c++
future<size_t>
reactor::do_writev(pollable_fd_state& fd, std::span<iovec> iovs) {
    return writeable(fd).then([this, &fd, iovs] () mutable {
        ...
```

if `writeable(fd)` yields, the iovs underlying storage is considered to be alive up until then's lambda is executed. Let me check one more thing ...

xemul: Yup, the only caller (well, the only _checked_ caller) of reactor::writev() is pipe_data_sink_impl::put() which preserves the iovec in its continuation block:

```c++
        std::vector<iovec> iov;
        ... // populate iov here
        auto iovspan = std::span<iovec>(iov);
        return _fd.write_all(iovspan).finally([del = std::move(del), iov = std::move(iov)] {});

```

travisdowns: Yes but the bug is in reactor_backend_asymmetric_uring::writev which makes a local copy of iovs on the stack which isn't preserved.

src/core/reactor_backend.hh
```diff
@@ -228,6 +228,13 @@ public:
     virtual future<> poll_rdhup(pollable_fd_state& fd) = 0;
     virtual void forget(pollable_fd_state& fd) noexcept = 0;
 
+    // For all the data-path methods below, buffer and iovec-array arguments
```
xemul, line 231: This can be too coarse. The span<temporary_buffer> thing appeared relatively recently, it was supposed to serve the needs of data_sink::put() methods, and the ownership requirement was very different there -- it was the put() code who had to copy all the buffers into stable location before returning, not the other way around. So this statement deserves double check

src/core/reactor_backend.cc
travisdowns: To be clear, `do_writev` does avoid this problem:

```
future<size_t>
reactor::do_writev(pollable_fd_state& fd, std::span<iovec> iovs) {
    return writeable(fd).then([this, &fd, iovs] () mutable {
```

`iovs` is copied by value into the .then() lambda so there is no problem. However I have a better fix also which avoids duplicating the iovs array, which I will post in a sec.

xemul: But maybe the asymmetric iouring's writev() should _not_ take the copy of the vector and rely on the caller to provide stable one (which is true)? It needs a local vector to call io_rquest::make_writev() overload, but this one, in turn, only needs the vector base and count, both can be taken from span itself

travisdowns: Yes, I'm doing that in in the other PR.

travisdowns: Superseded by https://github.com/scylladb/seastar/pull/3610.

src/core/reactor_backend.hh
travisdowns: Good call, it didn't make sense as an overall comment: I narrowed the comment to apply to `sendmsg` and `writev` where it definitely applies and made it more precise:

https://github.com/scylladb/seastar/pull/3610/changes#diff-341c3fb85f7e02d3f2b05a45e4830fd038adc1498430b1868754d8264d893b3fR237

travisdowns:
Superseded by https://github.com/scylladb/seastar/pull/3610.

PR #3606, opened 2026-08-12
doc: update the supported C++ standards to C++23 and C++26

travisdowns:
d9cbd5a20 ("build: update supported C++ standards to C++23, C++26") retired
C++20 and introduced C++26, but left `README.md` describing the old pair.

The stale text matters because it is actively misleading: it names
`--c++-standard=20` as the example, and C++20 is no longer a supported
standard. `CMAKE_CXX_STANDARD` now defaults to 26 and `configure.py` selects
the first of C++26, C++23 that the compiler accepts, so nothing auto-selects
C++20 any more.

`doc/compatibility.md` stated the policy generically ("the last two standards
approved by the ISO C++ committee") but illustrated it with a worked C++17 /
C++20 example, which now reads as a claim about what is supported. The example
is dropped and replaced with a pointer to the README section, so the standards
are named in exactly one place and the policy text survives the next bump.

Not touched: `doc/tutorial.md` mentions C++20 only as the standard that
introduced coroutines, which is still accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,11 +146,11 @@ There are also instructions for building on any host that supports [Docker](doc/
 
 Use of the [DPDK](http://dpdk.org) is [optional](doc/building-dpdk.md).
 
-#### Seastar's C++ standard: C++20 or C++23
+#### <a id="cxx-standard"></a>Seastar's C++ standard: C++23 or C++26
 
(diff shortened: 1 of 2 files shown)


travisdowns:
@scylladb/seastar-maint easy drive-by

PR #3609, opened 2026-08-13
ci: scan pull requests for credentials and injection with ThreatCrush

ralyodio:
Adds a pull-request workflow that scans the diff for hardcoded credentials,
injection, SSRF and unsafe deserialisation. Results go to the Security tab as
SARIF and to a comment on the pull request.

**It is report-only.** `failOn` is empty, so it annotates and never fails a build.
A repository with pre-existing findings should get a report on its first install,
not a blocked pull request — a gate that fires on everything gets switched off
within a day. Tighten it to `critical,high` in the workflow once any backlog is
triaged.

- `.github/workflows/threatcrush-scan.yml` — the workflow
- `.github/scripts/threatcrush-to-sarif.py` — a compatibility shim for CLI versions
  older than native SARIF output; unused once the installed CLI can emit it itself

Permissions are least-privilege (`contents: read`, `pull-requests: write`,
`security-events: write`). It runs on `pull_request`, not `pull_request_target`,
so contributor code never executes with your secrets in scope. The SARIF upload
is `continue-on-error` and degrades quietly where code scanning is unavailable.

Disclosure: I maintain [ThreatCrush](https://github.com/profullstack/threatcrush).
It is free and MIT, and the workflow installs it from npm — nothing here phones
home. If this is not something you want, closing it is the right answer, and I
will not send another.


.github/workflows/threatcrush-scan.yml
```diff
@@ -0,0 +1,291 @@
...
+  security-events: write
+
+jobs:
+  scan:
+    name: Scan for credentials and vulnerable patterns
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - uses: actions/checkout@v4
```
mykaul: this should use an explicit hash.

.github/workflows/threatcrush-scan.yml
```diff
@@ -0,0 +1,291 @@
...
+
+      - uses: actions/setup-node@v4
```
mykaul: same - this should use an explicit hash.

.github/workflows/threatcrush-scan.yml
```diff
@@ -0,0 +1,291 @@
...
+          node-version: "20"
+
+      # An unretried `npm i -g` is a network call to a registry that decides
+      # whether a security gate runs at all. Retry before giving up; a
+      # transient registry blip is not a security signal and should not read
+      # like one.
+      - name: Install ThreatCrush
+        run: |
+          for attempt in 1 2 3; do
+            if npm install -g "@profullstack/threatcrush@latest"; then
```
mykaul: '@latest' is not very stable. What if it gets hijacked as well?

.github/scripts/threatcrush-to-sarif.py
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/env python3
+"""Convert ThreatCrush terminal output to SARIF 2.1.0.
+
+Compatibility shim for CLI versions older than native ``--format sarif``.
```
mykaul, line 4: which cli?

ralyodio: threatcrush cli

.github/workflows/threatcrush-scan.yml
ralyodio: i push a lot though and don't want to hvae to manually update it.

.github/workflows/threatcrush-scan.yml
ralyodio: easier to get out dated and insecure is more likely then it getting hijacked, imo.

.github/workflows/threatcrush-scan.yml
mykaul: I don't want to be on the bleeding edge, or be concerned with malicious packages :-/ 
There's a balance here.

.github/scripts/threatcrush-to-sarif.py
mykaul: Why would we need to use an old CLI? If there is one that has native SARIF support?

mykaul:
We could just turn on the security features of Github I believe. What's the advantage of this tool?

PR #3613, opened 2026-08-13
core/iostream: assert data_sink put() lifetime rules in debug builds

travisdowns:
A `data_sink_impl` may hold per-sink state for the duration of a `put()`. At API level 9 and above `posix_data_sink_impl::put()` hands the kernel a pointer into its own `_vecs.v` and chains a deleter that captures `this`; the API < 9 `put(packet)` overload keeps the packet in `_p`.

That gives the sink two rules, neither of which was checked in a way that covers all sinks:

1. **Do not destroy an owner with a put in flight.** Both the pending operation and the continuation that cleans up after it point into the sink, so this dangles twice. Only TLS guarded against it, in `~openssl_session`, even though the memory at risk belongs to the *sink* rather than the session.
2. **Do not issue a put while one is in flight.** `wrapped_iovecs::populate()` reports `"wrapped_iovecs are re-populated live"` and the API < 9 overload asserts `!_p`, but both are posix-sink-only. `output_stream` already serializes deliberately, yet nothing verified the rule at the sink.

## Approach

`data_sink::put()` is a non-virtual forwarder and, at API level 9 and above, the only caller of `data_sink_impl::put()` anywhere in the tree. That makes it a ready-made choke point: set a flag on the impl for as long as a put is unresolved, assert it is clear on entry to `put()`, and assert it is clear in `~data_sink_impl`. This covers all 18 `data_sink_impl` subclasses without editing any of them.

The flag is cleared by an `ss::defer` guard moved into the continuation that observes the put, so it stays accurate on every path: it is set before the sink is called, so a put re-entered from inside `put()`, or a sink destroyed there, is covered too; a `put()` that throws clears it by unwinding the guard; otherwise it is cleared when the continuation resolves.

The two asserts lean on each other: a `bool` is only sufficient for rule 1 because rule 2 is enforced. Without it, overlapping puts would let the inner guard clear the flag while the outer put was still in flight.

## Documentation

Both rules were unstated at the sink layer, even though implementations depend on them. The last commit states them next to the existing note about buffer storage on `data_sink_impl::put()`, where an implementer will read them, and on `data_sink` in the same words `input_stream` and `output_stream` use for their own sequential-access rules.

Note `data_source_impl::get()` has the same unstated contract. Left for a separate change.

## Cost

Everything is confined to `SEASTAR_DEBUG` builds: no member, no `.finally()`, no branch otherwise. Verified rather than assumed: compiling `posix-stack.cc`, `tls_openssl.cc` and `fstream.cc` against pristine master and against this branch, non-debug and `-g0`, gives a byte-identical `tls_openssl.o` and two objects whose only differing bytes are `__LINE__` immediates that pre-existing asserts pass to `assert_fail`. Symbols and all section sizes are identical.

The first commit adds `SEASTAR_DEBUG_ASSERT`, since the `#ifdef SEASTAR_DEBUG` + `SEASTAR_ASSERT` pattern had no name yet.

## Testing

Debug build, 8 suites (`output_stream`, `tls` with both crypto providers, `socket`, `fstream`, `httpd`, `websocket`, `rpc`) under both `io_uring` and `asymmetric_io_uring`, 16 runs, all clean: neither rule is violated anywhere in those suites today. `tls_test`'s `test_output_pending_exception_on_destroy`, the sharpest false-positive probe in the tree, stays quiet, as it must: resolved means the cleanup has already run.

Both asserts were confirmed to actually fire, via throwaway tests not included here, including one where a sink synchronously re-enters `data_sink::put` on itself.

## Limitations

- **Neither assert has automated regression coverage, and the committed tests would pass against unpatched master.** An assert that terminates cannot be observed in-process, and seastar has no death-test facility today, so the two committed cases only check that draining a put and failing to issue one leave the sink destructible. Both aborts were verified by hand instead, as noted above. This is because we don't have "death tests".
- API level < 9 is not covered, since `data_sink::put()` is not a single choke point there.

### DIFF
--- a/include/seastar/core/iostream.hh
+++ b/include/seastar/core/iostream.hh
@@ -133,6 +133,10 @@ public:
     // The caller assumes that the storage that backs this span can be released
     // once this method returns, so implementations should move the buffers into
     // stable storage on their own early, before the returned future resolves.
+    //
+    // The caller will not issue another put() until the returned future resolves,
+    // and will not destroy the sink before then, so implementations may keep
+    // per-put state in the sink for the duration of the operation.
     virtual future<> put(std::span<temporary_buffer<char>> data) = 0;
 #else
     virtual future<> put(net::packet data) = 0;
@@ -210,6 +214,12 @@ protected:
 #endif
 };
 
+/// Pushes bulk data to a provider, such as a TCP connection, a disk file or a
+/// memory buffer.
+///
+/// \note All methods must be called sequentially.  That is, no method may be
+/// invoked before the previous method's returned future is resolved, and the
+/// sink may not be destroyed while a put() is outstanding.
 class data_sink {
     std::unique_ptr<data_sink_impl> _dsi;
 public:
--- a/include/seastar/core/iostream.hh
+++ b/include/seastar/core/iostream.hh
@@ -40,12 +40,19 @@
 #include <seastar/core/temporary_buffer.hh>
 #include <seastar/core/scattered_message.hh>
 #include <seastar/util/assert.hh>
+#include <seastar/util/defer.hh>
 #include <boost/intrusive/slist.hpp>
 #include <memory>
 #include <optional>
 #include <variant>
 #include <utility>
 #include <vector>
+
+// The put() bookkeeping below relies on data_sink::put() being the sole caller
+// of data_sink_impl::put(), which only holds at API level 9 and above.
+#if defined(SEASTAR_DEBUG) && SEASTAR_API_LEVEL >= 9
+#define SEASTAR_DATA_SINK_PUT_DEBUG
+#endif
 
 namespace bi = boost::intrusive;
 
@@ -102,8 +109,23 @@ public:
 };
 
 class data_sink_impl {
+#ifdef SEASTAR_DATA_SINK_PUT_DEBUG
+    // Set while a put() issued through data_sink has not resolved yet.
+    // Implementations are allowed to hold per-sink state for the duration of a
+    // put, which makes two things errors: destroying the sink while this is set,
+    // which leaves the operation and the continuation that cleans up after it
+    // pointing at freed memory, and issuing a second put while it is set, which
+    // makes the two puts share that state.
+    bool _put_in_flight = false;
+
+    friend class data_sink;
+#endif
 public:
-    virtual ~data_sink_impl() {}
+    virtual ~data_sink_impl() {
+#ifdef SEASTAR_DATA_SINK_PUT_DEBUG
+        SEASTAR_DEBUG_ASSERT(!_put_in_flight);
+#endif
+    }
     virtual temporary_buffer<char> allocate_buffer(size_t size) {
         return temporary_buffer<char>(size);
     }
@@ -201,7 +223,18 @@ public:
 #if SEASTAR_API_LEVEL >= 9
     future<> put(std::span<temporary_buffer<char>> data) noexcept {
         try {
+#ifdef SEASTAR_DATA_SINK_PUT_DEBUG
+            auto* dsi = _dsi.get();
+            // this guard enforces the general contract of the data sink that
+            // (1) there are never > 1 puts in flight
+            // (2) we never destroy the sink with a put in flight
+            SEASTAR_DEBUG_ASSERT(!dsi->_put_in_flight);
+            dsi->_put_in_flight = true;
+            auto guard = defer([dsi] () noexcept { dsi->_put_in_flight = false; });
+            return dsi->put(data).finally([guard = std::move(guard)] {});
+#else
             return _dsi->put(data);
+#endif
         } catch (...) {
             return current_exception_as_future();
         }
--- a/include/seastar/util/assert.hh
+++ b/include/seastar/util/assert.hh
@@ -34,3 +34,11 @@ namespace seastar::internal {
                                            __PRETTY_FUNCTION__);      \
         }                                                             \
     } while (0)
+
+/// Like SEASTAR_ASSERT(), but only active in SEASTAR_DEBUG builds. For checks
+/// whose cost is only worth paying while debugging.
+#ifdef SEASTAR_DEBUG
+#define SEASTAR_DEBUG_ASSERT(x) SEASTAR_ASSERT(x)
+#else
+#define SEASTAR_DEBUG_ASSERT(x) do { } while (0)
+#endif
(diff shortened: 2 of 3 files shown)


travisdowns:
@scylladb/seastar-maint debug-only QoL improvement for finding data sink contract violations.

include/seastar/core/iostream.hh
```diff
@@ -204,7 +254,18 @@ public:
...
         try {
+#ifdef SEASTAR_DATA_SINK_PUT_DEBUG
+            auto* dsi = _dsi.get();
+            // Enforces the contract that a sink carries only one put at a time.
+            SEASTAR_DEBUG_ASSERT(!dsi->_put_in_flight);
+            // put() is sequenced before the guard is constructed, so a throwing
+            // put() leaves the flag clear, and a guard that never reaches a
+            // continuation is destroyed by the unwind.
+            return dsi->put(data).finally(
+                    [guard = data_sink_impl::put_in_flight_guard(dsi)] {});
```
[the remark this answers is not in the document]
travisdowns: Good catch, and correct. Fixed in `22f6f1004`, folded into the commit that introduces the guard rather than added on top, since the comment there gave the old ordering as a deliberate property and would otherwise now be wrong.

I confirmed the window rather than taking it on faith. A sink whose `put()` synchronously calls `data_sink::put` on the same sink:

- before: the test passes and prints "REENTRANCY WENT UNDETECTED" -- two puts genuinely overlapped and neither assert fired;
- after: `iostream.hh:260: data_sink::put(...): Assertion '!dsi->_put_in_flight' failed`.

Your parenthetical about destruction mattered too, in fact slightly more than the re-entry case. Because the flag is now set before the call, a sink destroyed inside its own `put()` reaches `~data_sink_impl` with the flag set, so the destructor assert catches it. The old ordering also dereferenced the impl pointer *after* `put()` returned, which in that scenario was itself a use-after-free.

The properties the old ordering was protecting all survive: a throwing `put()` unwinds the local guard, and on success it moves into the `finally` capture, where the move constructor nulls the source so the local's destructor is a no-op and the flag is not cleared twice. `test_throwing_put_leaves_sink_destructible` still passes, and the debug suite sweep (8 suites x {io_uring, asymmetric_io_uring}) is still 16/16 clean, so setting the flag earlier does not fire anywhere real.

One caveat for the record: I have no evidence any in-tree sink re-enters or self-destructs inside `put()`, so this widens the check's coverage rather than fixing a live bug.

travisdowns:
Addressing the two comments suppressed in the last automated review, since neither has an inline thread to reply on.

**The stale test comment: fixed.** `// A put that throws never reaches the in-flight bookkeeping` described the previous ordering and became wrong when the guard moved ahead of the call. It now reads "A put that throws unwinds the guard, which clears the flag again." Folded into the commit that introduced the test rather than added on top.

**The missing regression coverage: correct, and deliberately accepted rather than fixed.** The observation is right, and stated more plainly than I had it: the two committed tests would pass against unpatched master, so the bookkeeping has no automated coverage at all.

Covering the aborts needs out-of-process death tests, and seastar has no facility for them today -- `test.py` has no expected-failure support, and `spawn_test.cc` is the only test that spawns a subprocess. Building that infrastructure is a larger change than the check it would cover, so it is out of scope here rather than deferred to a follow-up I would not get to.

Both asserts were verified by hand instead, each fired at its own site, including a sink that synchronously re-enters `data_sink::put` on itself. The PR description now says all of this outright instead of describing it as merely "no positive test case".

Worth naming the trade behind it: this gap is the price of aborting rather than logging. An earlier revision used `on_internal_error_noexcept`, which was testable in-process through `scoped_no_abort_on_internal_error` and the `internal_errors` counter, and had exactly the positive test being asked for. Dying instead of logging was chosen on purpose, so the lost testability is a known cost, not an oversight. If a reviewer would rather have the coverage than the abort, that decision is the lever.

include/seastar/core/iostream.hh
```diff
@@ -105,8 +111,43 @@ public:
...
+    // which leaves the operation and the continuation that cleans up after it
+    // pointing at freed memory, and issuing a second put while it is set, which
+    // makes the two puts share that state.
+    bool _put_in_flight = false;
+
+    // Marks a put in flight for as long as it is alive. Constructed before the
+    // sink is called and then moved into the continuation that observes the put,
+    // which keeps the flag accurate on the success, failure and never-issued
+    // paths alike.
+    class put_in_flight_guard {
```
xemul: Since the user of this helper itself sits under SEASTAR_DATA_SINK_PUT_DEBUG, why introducing it at all? Plain `... = true`/`... = false` would (should) result in less code easier to read

include/seastar/core/iostream.hh
```diff
@@ -247,6 +247,8 @@ public:
         try {
 #ifdef SEASTAR_DATA_SINK_PUT_DEBUG
             auto* dsi = _dsi.get();
+            // Enforces the contract that a sink carries only one put at a time.
+            SEASTAR_DEBUG_ASSERT(!dsi->_put_in_flight);
```
xemul, line 241: Isn't the invariant_checker_sink from unit test the better place for that? Or your goal is to check how `put` is called directly, not necessarily by the output_stream?

include/seastar/core/iostream.hh
travisdowns: It's an RAII helper, guaranteed to DTRT on all paths including exceptional ones: let me assess if we can get rid of it.

include/seastar/core/iostream.hh
travisdowns: > Or your goal is to check how put is called directly, not necessarily by the output_stream?

Yes, but more importantly the goal is not to just test this in the unit test here: this check applies to _all_ code compiled in debug mode so is effectively being tested in every test run like that which uses put, including functional and end to end tests with the full application.

E.g., it is up to the application to satisfy the output_stream contract of no concurrent writes, and if they don't you get hard to catch bugs: so this is especially about failing hard in those cases always (in debug) no matter when/how they are hit.

`invariant_checker_sink` is only a narrow tool specifically for this test.

travisdowns:
Pushed bc006acd9, changes vs the previous version:

- replaced the `put_in_flight_guard` class with `ss::defer` (review feedback)
- squashed the two assert commits into one
- reworked the commit messages and the comment above the assert

include/seastar/core/iostream.hh
travisdowns: I used ss::defer here instead, which gets rid of this class entirely.

travisdowns:
Force push 2cd6f69ab was a pure rebase onto current master (one trivial include-context conflict, no content changes).

PR #3614, opened 2026-08-13
github: generate the test matrix with covertable, and check its coverage

travisdowns:
Fixes the bug where the existing CI matrix is not pairwise due to https://github.com/thombashi/allpairspy/issues/2.

Four pairs appear in no job today:

```
compiler=clang++-21 with mode=release
compiler=clang++-22 with mode=debug
compiler=g++-16    with reactor-backend=io_uring
mode=debug         with reactor-backend=linux-aio
```

Change to generate with [covertable](https://github.com/walkframe/covertable), which takes
the exclusions as a constraint and removes the pairs they make unreachable
*before* deciding which rows it needs, instead of discarding rows it has already
committed to. The rule is stated once, as "no gcc combined with debug or
sanitize", and both the constraint and the coverage check read it, so the pair
list and `filter_func`'s positional row indices are both gone.

Also adds `_validate_pairwise()`, which fails the run when a pair that may
appear does not. Worth having whichever generator is underneath, since the gap
this PR fixes was invisible for as long as it existed, and widened from two
pairs to four when the reactor-backend dimension was added without anyone
noticing. It deliberately checks the generated rows only: the special-purpose
jobs cover combinations of their own, but the cxx-modules job runs no tests at
all, so a pair carried only by those jobs is not exercised by anything.

The note on dimension order goes away with it. It was there because allpairspy
varies its earlier dimensions more systematically, so leading with the backends
was what kept the rows from collapsing onto fewer distinct configurations.
covertable needs no such help: the dimensions stay in the order the matrix keys
are written, and that order already produces the fewest rows covertable will
emit for these parameters, as do most others (14 rows for 84 of the 120
permutations, 16 at worst).

## Effect

|  | before | after |
|---|---|---|
| regular rows | 12 | 14 |
| total jobs | 16 | 18 |
| missing pairs | 4 | 0 |

Verified: the exclusions still hold (no g++ debug or sanitize job), an
independent check of all 73 permitted pairs finds none missing, all 14 rows are
distinct configurations, the validation does fail and exits non-zero when a row
is removed, and output is deterministic, so regenerating without changing the
inputs leaves `tests.yaml` untouched.

covertable is pure Python, Apache-2.0 like seastar itself, and was last released on 2026-08-02.


travisdowns:
@scylladb/seastar-maint - a script/config-only change which fixes the CI matrix so that all pairs are properly covered.

travisdowns:
@scylladb/seastar-maint gentle ping, CI stuff only

travisdowns:
@scylladb/seastar-maint please take a look, fixes a bug in CI coverage

travisdowns:
Rebased onto master (force push `b5bdae718`). The conflict was with 6f489302c "github: run the tests on every reactor backend", which added `reactor-backend` as a fifth all-pairs dimension and shifted `filter_func`'s positional row indices, both of which this PR removes.

New numbers: matrix is 12 -> 14 regular rows, 16 -> 18 jobs.

.github/workflows/gen-matrix.py
```diff
@@ -78,26 +78,70 @@
...
 
 
-def _keep_row(row: list[Any]) -> bool:
-    """allpairspy filter_func: drop rows hitting an EXCLUDED_PAIRS entry.
+def _is_excluded(key_a: str, value_a: Any, key_b: str, value_b: Any) -> bool:
+    """Is this pair of parameter values one CONSTRAINTS rules out?"""
+    if {key_a, key_b} != {"compiler", "mode"}:
+        return False
+    compiler, mode = (value_a, value_b) if key_a == "compiler" else (value_b, value_a)
+    return compiler in GCC_COMPILERS and mode in GCC_BROKEN_MODES
```
xemul: The description says

> The rule is stated once, as "no gcc combined with debug or sanitize", 

but it looks like it's stated twise -- here and in CONSTRAINTS above

.github/workflows/gen-matrix.py
```diff
@@ -171,19 +215,14 @@ def _build_include(items: list[dict[str, Any]]) -> CommentedSeq:
...
-        # distinct configurations; trailing it gives 12 rows covering only 8,
-        # the other 4 being a config already present under another backend.
-        # Either way pairwise coverage is complete, and 12 is the floor here
-        # (4 compilers x 3 modes, and 4 compilers x 3 backends).
-        for b, c, s, m, a in AllPairs(
-            [BACKENDS, COMPILERS, STANDARDS, MODES, ARCHS],
-            filter_func=_keep_row,
-        )
+        {key: row[key] for key in PARAMETERS}
+        for row in make(PARAMETERS, strength=2, constraints=CONSTRAINTS)
```
xemul: When the covertable.make() meets an unknown parameter in a rule, the rule doesn't rejects anything. Currently it's not a problem, but if someone misprints in CONSTRAINTS it will come "unnoticed". So maybe some extra check in _validate_pairwise() below to make sure it doesn't really contain _is_excluded rows would be good

xemul (approved):
ACK, but I found two minor issues. If you think they are truly minor, I'm OK to merge this now, just let me know

travisdowns:
https://github.com/scylladb/seastar/commit/cc1edf6b9b1e5fa84560029fe2dda4e91170a0a3 is a pure rebase

travisdowns:
> ACK, but I found two minor issues. If you think they are truly minor, I'm OK to merge this now, just let me know

I will assess and fix these.

.github/workflows/gen-matrix.py
travisdowns: Good catch, and you are right -- it was stated twice, once in `CONSTRAINTS` and again in `_is_excluded`. What the patch actually deduplicated was the data (the compiler and mode lists) and `filter_func`'s positional indices, not the rule, so that sentence in the description was wrong.

Fixed by dropping `_is_excluded` and `_validate_pairwise` entirely. The validator was only checking covertable's own output, and a constraint violation would fail loudly in CI anyway, so it is not worth carrying the second copy of the rule to have it. `CONSTRAINTS` is now the only place the rule appears, and the description says so.

The generated matrix is byte-identical, so this is a pure deletion on top of what you reviewed.

travisdowns: That's true.

I ended up just deleting _is_excluded machinery entirely, this was essentially a redundant self-check on the output, but we can just trust (I think) convertible to do what it says on the tin, so we don't need this "double check" of the matrix it spits out. If there are a problems down the road it will likely be evident (e.g., generating forbidden pairs which fail CI).

This simplifies the whole thing.

I also saw there is a mode where you don't need this awkward DSL to express the constraint, so it's simplified to a plain function now.

.github/workflows/gen-matrix.py
travisdowns: True. I ended up being able to delete the DSL constraints entirely and just use an "evaluate" rule which delegates to a python function, so errors/typos will be caught directly during execution now (except for a typo in the work "requires" but that's a one-off and doesn't have a typo in it now :) ).

PR #3615, opened 2026-08-14
RFC: rpc: opportunistically batch outgoing frames into one compressed blob

mykaul:
## Motivation
When compression is negotiated, every outgoing RPC message is compressed and flushed individually - one compress call and one flush/syscall per message, and small messages compress poorly in isolation.
Bursts of small messages (heartbeats, small mutations, one-way verbs) pay heavy per-message overhead. Specific example - Scylla repair.

## Change
When multiple messages are already sitting in the outgoing queue, coalesce them (bounded by a hard 256 KiB / 32-message cap) into a single compressed blob, written and flushed once.
No artificial delay is introduced - a lone message is sent exactly as before, with a zero-allocation fast path. The receive side drains a decompressed blob until exhausted, which is unconditionally safe for single-frame senders.
Batching is gated by a new negotiated `BATCH_FRAMES` protocol feature bit, so it is fully backward compatible: an old peer never advertises it and never receives a multi-frame blob.

## Tests
Seven new unit tests: coalescing behavior, backward compatibility with a non-advertising peer, timeout handling, stream flush semantics, hard byte-cap boundary, and two teardown-under-batch races (send-error mid-batch, stop racing a batch, each injected at 6 offsets, flake-checked 25 runs).
Full rpc suite passes; survived two independent adversarial reviews.

## Results
Burst of small `int(int)` calls with the real lz4_fragmented compressor:
- 1000 messages went from 1000 compress calls to 32, throughput 426k → 561k msgs/s (+32%); 
- 5000 messages: 446k → 664k msgs/s (+49%).
- Solo-message latency unchanged by design.

### DIFF
--- a/doc/rpc-compression.md
+++ b/doc/rpc-compression.md
@@ -23,4 +23,31 @@ This compressor uses LZ4 to compress and decompress RPC messages. It requires al
 
 This compressor uses LZ4 streaming interface to compress and decompress even large messages without linearising them. The LZ4 streaming routines tend to be slower than the basic ones and the general logic for handling buffers is more complex, so this compressor is best suited only when there is no clear upper bound on the message size or if the messages are expected to be fragmented.
 
-Internally, the compressor processes data in 32 kB chunks and tries to avoid unnecessary copies. Applications should therefore use memory-buffer fragment sizes that are an integral multiple of 32 kB.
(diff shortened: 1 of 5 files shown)


mykaul:
@asias - is that what you had in mind in our discussion as an alternative to https://github.com/scylladb/scylladb/pull/31041 ?

asias:
> @asias - is that what you had in mind in our discussion as an alternative to [scylladb/scylladb#31041](https://github.com/scylladb/scylladb/pull/31041) ?

Yes. However, it seems this batching can never engage for rpc::sink / rpc::source streams, in either direction. That is the workload the original discussion was about.

PR #3616, opened 2026-08-14
net: reorder tcb::send/receive fields for hot-path cache locality

mykaul:
## Motivation
`tcb::send` mixed the congestion-control fields hit on every ACK (`window`, `cwnd`, `ssthresh`, `dupacks`, `mss`, `unsent_len`) with `std::deque` members and one-shot `promise<>` fields, spreading the per-ACK working set across multiple cache lines and scattering padding around lone bool/small-int flags.

## Change
Reorder `tcb::send`/`tcb::receive` members into hot (per-ACK counters, sequence numbers, flags — now one cache line), semi-hot (loss-recovery counters), and cold (queues, signal promises, RTT stats) groups. Pure field reorder: both structs are constructor-less aggregates accessed only by name — no behavioral change possible. Size asserts are `<=` upper bounds so libc++ builds (smaller `std::deque`) are unaffected.

## Tests
Library, demos, and header self-containment build clean; `net_config`/`packet`/`ipv6` suites pass. Honest caveat: no in-tree unit test exercises the native-stack `tcb` directly; verification was static (no `offsetof`/`memcpy`/positional-init anywhere, init-order hazards ruled out) plus an independent review that recomputed both layouts by hand for x86-64 and aarch64.

## Results
`sizeof(send)`: 400 → 384 bytes per connection (libstdc++/LP64); `receive` unchanged at 208 but locality-grouped. A synthetic proxy of the `can_send()`/`update_cwnd()` hot loop over 8192 connections measured ~12-15% fewer ns/iteration; treat as indicative, since the real `tcb` cannot be instantiated standalone.

### DIFF
--- a/include/seastar/net/tcp.hh
+++ b/include/seastar/net/tcp.hh
@@ -328,20 +328,37 @@ private:
             unsigned nr_transmits;
             clock_type::time_point tx_time;
         };
+        // Grouped by access frequency for cache locality: hot (congestion
+        // control + sequence numbers, touched on every ACK), semi-hot
+        // (retransmit counters), cold (queues, promises, RTT stats).
         struct send {
+            // hot
+            uint32_t window;
+            uint32_t cwnd; // Congestion window
+            uint32_t ssthresh; // Slow start threshold
+            uint32_t unsent_len = 0;
+            uint16_t mss;
+            uint16_t dupacks = 0; // Duplicated ACKs
+            uint8_t window_scale;
+            bool closed = false;
+            bool first_rto_sample = true;
+            bool window_probe = false;
+            uint8_t zero_window_probing_out = 0;
             tcp_seq unacknowledged;
             tcp_seq next;
-            uint32_t window;
-            uint8_t window_scale;
-            uint16_t mss;
             tcp_seq urgent;
             tcp_seq wl1;
             tcp_seq wl2;
             tcp_seq initial;
+            tcp_seq recover;
+            // semi-hot
+            unsigned syn_retransmit = 0;
+            unsigned fin_retransmit = 0;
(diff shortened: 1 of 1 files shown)


include/seastar/net/tcp.hh
```diff
@@ -386,6 +394,10 @@ private:
             // Currently, it is the same as default receive window size when window scaling is enabled
             size_t max_receive_buf_size = 3737600;
         } _rcv;
+        // Unlike send, sizeof(receive) doesn't shrink -- its fixed portion
```
travisdowns, line 397: This "doesn't shrink" comment doesn't make sense as a permanent comment in the code, it's the LLM thing where it's talking about the current diff it's doing: it belongs in the commit message (or cover letter).

travisdowns:
LGTM except for my nit about the comment.

Question: does this TCP stack/DPDK actually see any use in scylla?

mykaul:
Closing this — investigated the cache-locality claim in detail and it doesn't hold up as merged: the reordered hot fields in `tcb::send`/`tcb::receive` still straddle two cache lines rather than fitting in one (confirmed via direct offsetof/sizeof measurement against this branch: `_snd` at offset 120, 56 mod 64; `_rcv` at offset 504, also 56 mod 64), so the PR doesn't achieve what it claims.

Beyond that, `tcb` belongs to Seastar's native userspace TCP/IP stack (`--network-stack native`), which isn't used by real consumers of Seastar in production (e.g. ScyllaDB runs over the POSIX stack) — it's exercised mainly by demos and DPDK experiments. Given the incomplete fix and the very limited real-world benefit, not enough value to merge as-is.

PR #3622, opened 2026-08-14
unit tests: cap test process memory at 1G by default, allow per-test override

### LINKED ISSUE #3619: Alpine Linux CI job intermittently fails all tests with "Out of memory" at startup
## Summary

The "Alpine Linux" CI workflow (`.github/workflows/alpinelinux.yaml`, job `build-and-test`) intermittently fails its "Run unit tests" step with every single unit test failing at process startup:

```
Test setup error: system_error produced by: exp: Out of memory
```

This happens on the very first line of execution for essentially every test binary (`abort_source_test`, `alloc_test`, `alien_test`, `checked_ptr_test`, `temporary_buffer_test`, `chunked_fifo_test`, `linux_perf_event_test`, `chunk_parsers_test`, ...) — completely unrelated tests, all failing in well under 0.3s, consistent with a memory-reservation failure at reactor startup rather than a runtime leak or a bug in any specific test.

## Evidence this is a CI/environment flake, not a code regression

Observed on PR #3616 (`net: reorder tcb::send/receive fields for hot-path cache locality`):
- Run `31780703344` (07:38): **passed**, 23m58s, no code changes since.
- Run `31791230438` (10:11), same commit content (only a force-push of a rebased/squashed history, no functional diff to the failing tests' code paths): **failed** with the OOM pattern above.
- The re-run (`gh run rerun --failed`) was queued immediately after.
- Sibling PRs (#3615, #3617) on the same day show the identical Alpine Linux workflow alternating between `success` and this failure across pushes with no relevant code changes.
- The 12+ jobs in the actual compiler/standard/mode/arch test matrix (`.github/workflows/tests.yaml`) are unaffected and consistently pass — this is isolated to the separate, simpler Alpine smoke-build workflow.

## Suspected mechanism (untested, worth investigating)

The job runs `runs-on: ubuntu-latest` with `container: image: alpine:latest` — i.e. GitHub's standard (memory-constrained, ~7GB, shared) hosted runner, executing job steps inside a nested Alpine container. The workflow configures with a raw `cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo` (not using `configure.py`/`seastar_cmake.py`'s named modes), then runs `ctest --test-dir build --output-on-failure` with no explicit `--smp`/memory constraints — yet the actual per-test invocations captured in CI logs include `-- -c 2` (2 shards) and `ASAN_OPTIONS=`/`UBSAN_OPTIONS=` environment prefixes.

Seastar has cgroup-aware memory detection (`src/core/resource.cc`, cgroups v1/v2 paths). Worth checking specifically whether:
1. cgroup memory-limit detection behaves correctly for the nested-container execution model GitHub Actions' `container:` job syntax produces (as opposed to a single top-level container), particularly under Alpine's musl libc.
2. The effective per-shard memory reservation seastar computes, combined with `-c 2` and whatever sanitizer overhead is actually present in this build, exceeds what's reliably available on a standard shared GitHub-hosted runner, making this a resource-starvation race rather than a strict bug — plausible given it doesn't fail every run.
3. Multiple test binaries are able to run concurrently (ctest default parallelism) each reserving memory independently, compounding the pressure.

## Ask

Reproduce locally (Alpine container matching the workflow's exact `apk add` list and cmake invocation), determine the actual cause of the intermittent OOM, and fix it — most likely at the CI workflow level (explicit memory/shard-count constraints, `ctest -j1`, or similar), but keep an open mind to a genuine cgroup-detection bug if the evidence points there.

mykaul:
## Summary
- Seastar's memory auto-detection (hwloc / cgroup / sysconf) makes each unit-test process reserve close to all the memory the host claims to have. On memory-constrained or over-committed hosts (e.g. the Alpine CI job, which runs in an unconstrained nested container on a shared runner) that reservation can fail with ENOMEM at startup.
- Add a `Seastar_UNIT_TEST_MEM` CMake cache variable, defaulting to `1G`, that caps how much memory each test process reserves via `--memory`. This is nicer for workstations and improves reproducibility.
- Add a per-test `MEM` argument to `seastar_add_test()` to override the cap for a single test, and use it to run `rpc_test` with `--memory 2G` (the LZ4 compressor sub-tests need more than 1G).

## Test plan
- [x] Reproduced the OOM-at-startup pattern in a matching Alpine container.
- [x] Verified the fix in the same container: 92/93 tests pass (the one failure is the pre-existing, unrelated `Seastar.unit.socket` flake), no more startup OOM.
- [x] Full unit test suite (93 runnable tests) passes locally with the 1G default; the `sstring` unit test fails to compile in this environment with fmt 11.2 (pre-existing, unrelated to this change).

Fixes #3619

### DIFF
--- a/.github/workflows/alpinelinux.yaml
+++ b/.github/workflows/alpinelinux.yaml
@@ -50,12 +50,22 @@ jobs:
           xfsprogs-dev       \
           yaml-cpp-dev
 
+    - name: Report available memory
+      run: |
+        # Diagnostic for #3619 / #3622: real free memory at test time.
+        cat /proc/meminfo | head -5
+        free -h || true
+
     - name: Configure build
       run: |
+        # Cap test process memory: this container has no cgroup limit, so
+        # seastar's auto-detection can try to reserve all the runner's RAM
+        # and OOM at startup under real memory pressure (#3619).
         cmake -B build -G Ninja            \
          -DCMAKE_BUILD_TYPE=RelWithDebInfo \
          -DSeastar_LTTNG=OFF               \
-         -DSeastar_DOCS=OFF
+         -DSeastar_DOCS=OFF                \
+         -DSeastar_UNIT_TEST_MEM=1G
 
     - name: Build Seastar
       run: |
@@ -63,4 +73,6 @@ jobs:
 
     - name: Run unit tests
       run: |
-        ctest --test-dir build --output-on-failure
+        ulimit -v unlimited  # mem_base()'s 32TB mmap probe fails under a finite RLIMIT_AS regardless of --memory (#3619)
+        # retry once: shared runners occasionally OOM under host-level memory pressure unrelated to this job
+        ctest --test-dir build --output-on-failure || ctest --test-dir build --output-on-failure --rerun-failed
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -29,6 +29,16 @@ set (Seastar_UNIT_TEST_SMP
   STRING
   "Run unit tests with this many cores.")
 
+# Caps memory each test process reserves via `--memory`, instead of seastar's
+# auto-detection (hwloc / cgroup / sysconf) grabbing most of the host's RAM.
+# Empty falls back to auto-detection. 512M: even 1G triggered genuine ENOMEM
+# on the Alpine CI job; may need further tuning against real CI numbers.
+set (Seastar_UNIT_TEST_MEM
+  "512M"
+  CACHE
+  STRING
+  "Run unit tests with this much memory (e.g. 1G). Empty means auto-detect.")
+
 # Check if the compiler supports -fsanitize=fuzzer (compile and link)
 include (CheckCXXSourceCompiles)
 include (CMakePushCheckState)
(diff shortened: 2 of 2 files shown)


avikivity:
I think it makes sense to cap the unit test memory by default, it's nicer for people's workstations, and improves reproducibility.

mykaul:
> I think it makes sense to cap the unit test memory by default, it's nicer for people's workstations, and improves reproducibility.

what's the default value should be?

avikivity:
> > I think it makes sense to cap the unit test memory by default, it's nicer for people's workstations, and improves reproducibility.
> 
> what's the default value should be?

1GB seems fine (as long as all tests pass)

mykaul:
Not enough? 
```
92/93 Test #92: Seastar.app.memcached.memcached ...............***Failed    4.16 sec
[1/1] cd /__w/seastar/seastar/build/apps/memcached/tests && /__w/seastar/seastar/apps/memcached/tests/test.py --memcached /__w/seastar/seastar/build/apps/memcached/memcached
WARNING: unable to mbind shard memory; performance may suffer: Operation not permitted
INFO  2026-08-17 09:14:25,212 seastar - Reactor backend: linux-aio
WARNING: unable to mbind shard memory; performance may suffer: Operation not permitted
WARN  2026-08-17 09:14:25,223 seastar - Creation of perf_event based stall detector failed: falling back to posix timer: std::system_error (error system:1, perf_event_open() failed: Operation not permitted)
WARN  2026-08-17 09:14:25,223 seastar - Creation of perf_event based stall detector failed: falling back to posix timer: std::system_error (error system:1, perf_event_open() failed: Operation not permitted)
INFO  2026-08-17 09:14:25,224 [shard 1:main] io - IO queue was unable to find a suitable maximum request length, the search was cut-off early at: 16MB
INFO  2026-08-17 09:14:25,224 [shard 1:main] io - IO queue was unable to find a suitable maximum request length, the search was cut-off early at: 16MB
ERROR 2026-08-17 09:14:25,225 [shard 1:main] seastar - : Out of memory
```

Need to look into this more.

PR #3626, opened 2026-08-14
[do not merge] rpc: avoid structured bindings over co_await results in loop()/process()

mykaul:
## Summary
- Works around a suspected GCC 16 coroutine codegen bug where a structured binding to a `co_await` result can have its `std::optional<rcv_buf>` member destroyed even when the `co_await` itself threw and the binding was never constructed, which trips UBSan on an uninitialized bool read.
- Replaces the structured bindings in `client::loop()` and `server::connection::process()` with a named variable plus `std::get<>()`, which sidesteps the problematic codegen path without changing behavior.
- Fixes seastar#3625.

## Test plan
- [x] `test_rpc_cancel` in isolation: 100/100 clean runs under `debug` (ASan+UBSan), vs. ~7.5% baseline failure rate before the fix.
- [x] Full `rpc_test` suite: 25/25 clean runs under `debug` (ASan+UBSan).
- [x] `ninja -C build/debug rpc_test` builds cleanly.

### DIFF
--- a/src/rpc/rpc.cc
+++ b/src/rpc/rpc.cc
@@ -1000,7 +1000,14 @@ future<> client::loop(client_options ops, const socket_address& addr, const sock
                 co_await handle_stream_frame();
                 continue;
             }
-            auto&& [msg_id, ht, data] = co_await read_response_frame_compressed(_connected->read_buf);
+            // Not a structured binding: GCC 16's coroutine codegen has been observed
+            // (seastar#3625) destroying the optional<rcv_buf> member of a structured
+            // binding to a co_await result even when the co_await itself threw and the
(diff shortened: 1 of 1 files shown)


mykaul:
this needs more attention (from me).

mykaul:
Closing per its own title: parked draft/workaround for a suspected, unconfirmed GCC 16 coroutine codegen issue, not intended to merge as-is.

mykaul:
Closing per its own title: parked draft/workaround for a suspected, unconfirmed GCC 16 coroutine codegen issue, not intended to merge as-is.

mykaul:
Closing per its own title: parked draft/workaround for a suspected, unconfirmed GCC 16 coroutine codegen issue, not intended to merge as-is.

PR #3629, opened 2026-08-17
tests: run unit tests in parallel

mykaul:
Motivation
- Reduce unit-test wall-clock time without allowing shared-resource races.

Changes
- Run unit tests in parallel with CTest resource protection.
- Use separate TLS provider ports with validated configuration.
- Split RPC tests into six parallel shards.
- Serialize the shared RPC build fixture and cap builds at 4 jobs.

Tests
- 9/9 focused RPC/TLS tests passed.
- RPC shards: 23.96s serial, 12.44s parallel.
- Full serial unit suite: 95/95 passed.
---

**Update:** A pre-existing `Seastar.unit.prometheus` flake was uncovered while enabling parallel ctest execution here (the exporter listens on `--smp=2`, and a `/metrics` request can land on a shard that hasn't seen a given metric's label, silently dropping the filter). This is a bug independent of the parallelism changes in this PR, so it has been split out into a standalone commit/PR: see the fix for review and merge separately. This PR's history has been reordered so the fix commit ("prometheus: build label filter from all shards, not just the local one") lands before "tests: run CI ctest in parallel", keeping bisectability correct.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -50,6 +50,8 @@ jobs:
     runs-on: ${{ inputs.arch == 'arm' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
     container:
       image: ubuntu:26.04
+      # --privileged: needed to raise fs.aio-max-nr (below) for -j4 test runs.
+      options: --privileged
     steps:
       - name: Show parameters
         run: |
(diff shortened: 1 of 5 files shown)


mykaul:
Need some work with the commit messages and so on.

PR #3630, opened 2026-08-18
prometheus: build label filter from all shards, not just the local one - if needed.

mykaul:
## Motivation

`Seastar.unit.prometheus` flakes intermittently when the metrics exporter runs with more than one shard.

## Root cause

`metrics_handler::make_filter()` used `mi::get_local_impl()->get_labels()` to decide which query parameters are valid label filters. This set is shard-local: a label is only known on a shard if some metric carrying it happens to be registered there.
Since the HTTP listener accepts connections on every shard, a request can be served by a shard that has never seen a given metric/label.

When that happens, the query param is treated as unknown, the filter matcher stays empty, and `make_filter()` falls back to the "accept everything" function, silently returning unfiltered data instead of applying the requested filter.

## Fix

Metric values are already aggregated across all shards (see `get_map_value()`), so a query param unknown locally isn't necessarily unknown cluster-wide.
`make_filter()` now falls back to checking the other shards' labels, but only for params the local shard doesn't recognize: the common case, where the handling shard already knows every requested label, stays a synchronous, allocation-free local lookup with no cross-shard fan-out.
This avoids adding any measurable overhead in the common case.

## Testing

- `tests/unit/prometheus_test.py` (standalone, 20 runs): OK, 5/5 tests each.
- `ctest -R "^Seastar.unit.prometheus$"` run 15x directly: 15/15 passed.
- `ctest -j4 -R "prometheus"` (parallel): all 3 prometheus tests (`prometheus_text`, `prometheus_http`, `prometheus`) passed every time across multiple runs.   ( see https://github.com/scylladb/seastar/pull/3629 )
- Timing comparison (fixed vs. unfixed) for the 3 prometheus tests showed no measurable difference, serial or parallel:

  | | Serial (`-j1`) | Parallel (`-j4`) |
  |---|---|---|
  | Unfixed | 0.95-1.39s | 0.52-0.56s |
  | Fixed | 0.95-1.32s | 0.51-0.56s |

### DIFF
--- a/include/seastar/core/metrics_api.hh
+++ b/include/seastar/core/metrics_api.hh
@@ -445,6 +445,7 @@ class impl {
     value_map _value_map;
     config _config;
     bool _dirty = true;
+    bool _propagate_new_labels = false;
     shared_ptr<metric_metadata> _metadata;
     std::set<sstring> _labels;
     std::vector<std::deque<metric_function>> _current_metrics;
@@ -461,6 +462,18 @@ public:
     }
 
     register_ref add_registration(const metric_id& id, const metric_type& type, metric_function f, const description& d, bool enabled, skip_when_empty skip, const std::vector<std::string>& aggregate_labels);
+    /*!
+     * \brief add label names to the set of labels known to this shard
+     */
+    void add_labels(const std::set<sstring>& labels);
+
+    /*!
+     * \brief share the label names known to this shard with the other shards
+     *
+     * Also enables propagating label names added from now on, as they are
+     * registered.
+     */
+    future<> propagate_labels();
     internalized_labels_ref internalize_labels(labels_type labels);
     void remove_registration(const metric_id& id);
     future<> stop() {
@@ -501,6 +514,7 @@ public:
     void update_aggregate(metric_family_info& mf) const noexcept;
 
 private:
+    void register_labels(const labels_type& labels);
     void gc_internalized_labels();
     bool apply_relabeling(const relabel_config& rc, metric_info& info);
 };
@@ -511,6 +525,11 @@ using values_reference = shared_ptr<values_copy>;
 foreign_ptr<values_reference> get_values();
 
 shared_ptr<impl> get_local_impl();
+
+/*!
+ * \brief add label names to the set of labels known to the current shard
+ */
+void add_labels(const std::set<sstring>& labels);
 
 void unregister_metric(const metric_id & id);
 
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -34,6 +34,8 @@
 #include <seastar/core/metrics_api.hh>
 #include <seastar/core/relabel_config.hh>
 #include <seastar/core/reactor.hh>
+#include <seastar/core/smp.hh>
+#include <seastar/util/log.hh>
 
 namespace seastar {
 extern seastar::logger seastar_logger;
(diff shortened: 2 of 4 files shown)


mykaul:
Perhaps a reasonable solution would be to only accept Prometheus HTTP queries on shard 0, and limit this? Unsure.

@amnonh ?

amnonh:
What about updating the label list on all shards when it changes?
 This is, first, static (only when registering a metric) and, second, rare, only when registering a metric and a new label is added.

mykaul:
> What about updating the label list on all shards when it changes? This is, first, static (only when registering a metric) and, second, rare, only when registering a metric and a new label is added.

I don't mind, but should it be part of this PR? Seems like an existing gap. Should I add another commit to this PR?

amnonh:
@mykaul I think there's a better implementation than this PR. If the labels list is always up to date, no need to check other shards.

mykaul:
> @mykaul I think there's a better implementation than this PR. If the labels list is always up to date, no need to check other shards.

Sure. How do I know that? I'd be happy to implement it. I just implemented that it consults the local shard first.

amnonh:
@mykaul in impl::add_registration we add the metrics labels to the _labels set. We can add a function that, if it's a new label, propagates the label set to the other shards.

mykaul:
> @mykaul in impl::add_registration we add the metrics labels to the _labels set. We can add a function that, if it's a new label, propagates the label set to the other shards.

@amnonh - in this PR?

amnonh:
@mykaul, if I understand the problem, not all shards know about all labels, this pr fix it by checking on other shards for missing labels.
I suggest an alternative solution that will not be part of the actual running path.

mykaul:
New version of the patch, rebased on `origin/master` and force-pushed.

- Change: propagate metric label names across shards during Prometheus route setup and runtime registration; preserve pending labels when propagation fails; register labels introduced by relabeling; keep filtering local and synchronous.
- Test: `ctest --test-dir build/debug -R "prometheus|metrics" --output-on-failure`
- Result: 4/4 tests passed, including Prometheus text, HTTP, Python, and metrics tests.

mykaul:
> New version of the patch, rebased on `origin/master` and force-pushed.
> 
> * Change: propagate metric label names across shards during Prometheus route setup and runtime registration; preserve pending labels when propagation fails; register labels introduced by relabeling; keep filtering local and synchronous.
> * Test: `ctest --test-dir build/debug -R "prometheus|metrics" --output-on-failure`
> * Result: 4/4 tests passed, including Prometheus text, HTTP, Python, and metrics tests.

@amnonh - I hope that's what you had in mind.

amnonh:
I think the implementation is too complicated. 
What about:
Add a ``` void add_labels(std::set<sstring> label);``` in metrics_api.hh for ```class impl``` and a global one.
Where the ```impl``` one's implementation adds the labels to _labels.

The global one's implementation is:
```
void add_labels(std::set<sstring> labels) {
    return impl::get_local_impl()->add_labels(labels);
}
```

Now, in ```impl::add_registration```, also check whether the set was modified; if so, call all other shrds with add_labels

mykaul:
Reimplemented along the lines you suggested, rebased on `origin/master` and force-pushed — now a single commit.

What it does now, per your sketch:

- `impl::add_labels(const std::set<sstring>&)` adds the names to `_labels`, plus a free `add_labels()` in `metrics_api.hh` that forwards to `get_local_impl()->add_labels()`.
- `impl::add_registration()` (and `set_relabel_configs()`, for labels that only relabeling introduces) checks whether the set changed and, if so, calls the other shards' `add_labels()` with just the new names.
- `add_prometheus_routes()` shares the names each shard already had at that point, once.

Gone from the previous version: the `_pending_labels` queue, the requeue-on-failure handling, and the cross-shard `shared_ptr`. Filtering is unchanged — still a local, synchronous lookup, and `make_filter()` no longer copies the label set on every request.

The propagation stays off until the metrics are exported, so nothing is submitted to other shards while metrics are being registered during startup (at that point the smp queues may not exist yet).

Testing, debug mode:

- `ctest -R "prometheus|metrics"`: 4/4 pass.
- `Seastar.unit.prometheus` (the flaky one) run 20+ times, serially and with `-j4`: all pass.
- New unit test covering all three paths: the initial full-set share, a label added by a later registration, and a label added by relabeling.
- `Seastar.unit.httpd` fails in this build both with and without the patch (a pre-existing UBSan `pollable_fd_state` complaint), so it is unrelated.

amnonh (approved):
LGTM

ScyllaPiotr:
@amnonh @mykaul  Please see https://github.com/scylladb/seastar/pull/3649#issuecomment-5444307128 and decide which solution (contract) you like more.

ScyllaPiotr:
> @amnonh @mykaul Please see [#3649 (comment)](https://github.com/scylladb/seastar/pull/3649#issuecomment-5444307128) and decide which solution (contract) you like more.

@amnonh do you have an opinion on this?

mykaul:
AI decided to merge this. @&@£#)#?£!!!!

PR #3632, opened 2026-08-20
coroutine: Remove deprecated exception helpers

xemul:
make_exception() and return_exception(exception_ptr) were deprecated in favor of return_exception_ptr() and coroutine::exception, by d43816702e1 (2024-02-08) and 32bf0618064 (2022-07-05) respectively. Neither has any in-tree callers left.

### DIFF
--- a/include/seastar/coroutine/exception.hh
+++ b/include/seastar/coroutine/exception.hh
@@ -80,35 +80,6 @@ struct exception {
 /// Allows propagating an exception from a coroutine directly rather than
 /// throwing it.
 ///
-/// `make_exception()` returns an object which must be co_returned.
-/// Co_returning the object will immediately resolve the current coroutine
-/// to the given exception.
-///
-/// \note Due to language limitations, this function doesn't work in coroutines
-/// which return future<>. Consider using return_exception instead.
-///
-/// Example usage:
-///
-/// ```
-/// co_return coroutine::make_exception(std::runtime_error("something failed miserably"));
-/// ```
-[[deprecated("Use co_await coroutine::return_exception_ptr or co_return coroutine::exception instead")]]
-[[nodiscard]]
-inline exception make_exception(std::exception_ptr ex) noexcept {
-    return exception(std::move(ex));
-}
-
-template<typename T>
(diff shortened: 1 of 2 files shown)


travisdowns (approved):

include/seastar/coroutine/exception.hh
```diff
@@ -143,12 +114,6 @@ inline exception return_exception_ptr(std::exception_ptr ex) noexcept {
 /// ```
 /// co_await coroutine::return_exception(std::runtime_error("something failed miserably"));
 /// ```
-[[deprecated("Use co_await coroutine::return_exception_ptr instead")]]
-[[nodiscard]]
-inline exception return_exception(std::exception_ptr ex) noexcept {
```
travisdowns, line 147: This one seems a bit risky, since any calls to `return_exception(eptr)` now bind to the remaining  `return_exception(T&&)` below, which will create an incorrectly nested exception_ptr (an expcetion pointer which itself points to an exception pointer). 

Perhaps the T&& t overload needs to be extended with a concept to reject exception pointers, or else we leave this overload here but `= delete`.

include/seastar/coroutine/exception.hh
```diff
@@ -77,35 +77,6 @@ struct exception {
...
-[[deprecated("Use co_await coroutine::return_exception_ptr or co_return coroutine::exception instead")]]
-[[nodiscard]]
-inline exception make_exception(std::exception_ptr ex) noexcept {
-    return exception(std::move(ex));
-}
-
-template<typename T>
-[[deprecated("Use co_await coroutine::return_exception_ptr or co_return coroutine::exception instead")]]
-[[nodiscard]]
-exception make_exception(T&& t) noexcept {
```
travisdowns, line 103: doc/tutorial.md:582 is out of date now as it it refers to this method

include/seastar/coroutine/exception.hh
xemul: Good catch, fixed

include/seastar/coroutine/exception.hh
xemul: Yup, updated

travisdowns (approved):

PR #3636, opened 2026-08-23
fixes for libc++

crackcomm:
Fixes for compilation under bazel, clang 22 with libc++, see commit messages for more details.

`operator[]` on circular buffer - [boost::sort in tests](https://github.com/crackcomm/seastar/blob/push-qwwsmuvuskzn/tests/unit/circular_buffer_fixed_capacity_test.cc#L148-L149).

### DIFF
--- a/include/seastar/core/circular_buffer_fixed_capacity.hh
+++ b/include/seastar/core/circular_buffer_fixed_capacity.hh
@@ -130,6 +130,9 @@ public:
             _idx -= n;
             return *this;
         }
+        ValueType& operator[](difference_type n) const {
+            return _start[mask(_idx + n)].data;
+        }
         bool operator==(const cbiterator& rhs) const {
             return _idx == rhs._idx;
         }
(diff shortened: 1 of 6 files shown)


tests/unit/exception_logging_test.cc
```diff
@@ -33,6 +33,13 @@
 using namespace seastar;
 
 #ifndef SEASTAR_BACKTRACE_UNIMPLEMENTED
+#ifdef _LIBCPP_VERSION
```
xemul: nit: better to keep it close to the only user

include/seastar/core/circular_buffer_fixed_capacity.hh
```diff
@@ -130,6 +130,9 @@ public:
             _idx -= n;
             return *this;
         }
+        ValueType& operator[](difference_type n) const {
```
avikivity, line 133: Should return const. And add the non-const counterpart too.

avikivity: Actually no, this is part of the iterator.

tests/unit/exception_logging_test.cc
crackcomm: done

PR #3638, opened 2026-08-23
treewide: cleanup unused includes

crackcomm:
It was done without any tools, only LLM in a loop.

Tests not included (maybe later if this gets merged).

### DIFF
--- a/include/seastar/net/api.hh
+++ b/include/seastar/net/api.hh
@@ -24,18 +24,14 @@
 #include <chrono>
 #include <memory>
 #include <vector>
-#include <cstring>
 #include <sys/types.h>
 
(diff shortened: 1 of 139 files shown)


include/seastar/net/api.hh
```diff
@@ -24,18 +24,14 @@
...
 #include <memory>
 #include <vector>
-#include <cstring>
 #include <sys/types.h>
 
 #include <seastar/core/future.hh>
-#include <seastar/net/byteorder.hh>
 #include <seastar/net/socket_defs.hh>
 #include <seastar/net/packet.hh>
-#include <seastar/core/internal/api-level.hh>
```
crackcomm, line 34: should this be exported?

PR #3639, opened 2026-08-24
util: move assert_fail to assert.cc

crackcomm:
It fixes linking issues when using parts of seastar in wasm, also, makes more sense than having it in log.cc.

### DIFF
--- a/src/util/assert.cc
+++ b/src/util/assert.cc
@@ -26,7 +26,7 @@
 namespace seastar::internal {
 [[noreturn]] void assert_fail(const char *msg, const char *file, int line,
                               const char *func) {
-    fprintf(stderr, "%s:%u: %s: Assertion `%s` failed.\n", file, line, func, msg);
+    fprintf(stderr, "%s:%d: %s: Assertion `%s` failed.\n", file, line, func, msg);
     std::fflush(stderr);
     std::terminate();
 }
--- a/src/util/assert.cc
+++ b/src/util/assert.cc
@@ -1,0 +1,33 @@
+/*
+ * This file is open source software, licensed to you under the terms
+ * of the Apache License, Version 2.0 (the "License").  See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership.  You may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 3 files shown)


src/util/assert.cc
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Copyright (C) 2015 Cloudius Systems, Ltd.
```
travisdowns: This is wrong, use your own copyright here (see Avi's prior comments on this) and fix the year.

src/util/assert.cc
```diff
@@ -0,0 +1,31 @@
...
+ */
+#include <cstdio>
+#include <exception>
+
```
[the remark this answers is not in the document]
travisdowns: @crackcomm this is valid, please fix.

src/util/assert.cc
```diff
@@ -0,0 +1,31 @@
...
+namespace seastar::internal {
+[[noreturn]] void assert_fail(const char *msg, const char *file, int line,
+                              const char *func) {
+    fprintf(stderr, "%s:%u: %s: Assertion `%s` failed.\n", file, line, func, msg);
```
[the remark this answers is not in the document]
travisdowns: existing issue

src/util/assert.cc
crackcomm: Not sure where are Avi's comments on this, but it's just a copy paste. It's not my code.

crackcomm: I updated the year and company name to ScyllaDB. I think Avi's comments you were referring to might be about the code you wrote, not copied.

src/util/assert.cc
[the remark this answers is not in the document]
crackcomm: Should I add a commit that fixes it while we're at it?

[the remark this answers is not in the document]
crackcomm: I pushed a fix for it too (ask for forgiveness not permission?).

src/util/assert.cc
travisdowns: Yes makes sense, we should use the copyright from the source file in that case.

dotnwat: do you have a reference to avi's comments that you are referring

travisdowns: Hey, @dotnwat , sure thing, some refs:

https://github.com/scylladb/seastar/pull/2646#discussion_r1958769227
https://github.com/scylladb/seastar/pull/3440#discussion_r3422781012
https://github.com/scylladb/seastar/pull/3146#discussion_r2611091790

PR #3641, opened 2026-08-25
tests: migrate websocket_test helpers to coroutines

crackcomm:
This change is the most straight-forward out of the series.

### DIFF
--- a/tests/unit/websocket_test.cc
+++ b/tests/unit/websocket_test.cc
@@ -5,6 +5,7 @@
 #include <seastar/core/future.hh>
 #include <seastar/websocket/server.hh>
 #include <seastar/websocket/client.hh>
+#include <seastar/core/coroutine.hh>
 #include <seastar/testing/test_case.hh>
 #include <seastar/testing/thread_test_case.hh>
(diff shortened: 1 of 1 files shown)


tests/unit/websocket_test.cc
```diff
@@ -108,74 +101,67 @@ SEASTAR_TEST_CASE(test_websocket_handshake_no_subprotocol) {
...
+
+        co_await output.write(ws_frame);
+        co_await output.flush();
+
+        auto response = co_await input.read_exactly(6);
         auto response_str = std::string(response.begin(), response.end());
         BOOST_REQUIRE_EQUAL(rs_frame, response_str);
-    });
+    };
+    co_await body().finally(seastar::coroutine::lambda([&]() -> future<> {
```
xemul: The 

```c++
    std::exception_ptr ex;
    try {
        co_await ...;
        co_await ...;
        ...
    } catch {
        ex = std::current_exception();
    }
    co_await ...; // deferred co_await-ed actions
    if (ex) {
        std::rethrow_exception(ex);
    }
````

patters is more common to seastar

PR #3645, opened 2026-08-26
Use a separate syscall thread for aio fallbacks

ballard26:
The lazy-time mode ctime patch does fix the AIO write fallbacks for most workloads in kernel 7.0+. However, for append-heavy workloads that often issue writes that span a written to unwritten extent boundary(from overwrites for alignment) this isn't the case. XFS currently considers writes across two extents to be potentially blocking as it requires a submission of multiple bio requests and subsequent retries if one fails. In internal testing we found that the ctime patch reduces fallbacks of our common workloads from ~67% to ~50%, with the remaining 50% of the fallbacks being due to the aforementioned issue.

We've tried a number of fixes for this:
- Submitting overwrites separately from newly written data ends up costing more than failing a unifed write request and retrying on the syscall thread.
- Submitting AIO write ops without the non-blocking flag. This was seastar's chosen solution before the ctime patch landed. However, the risk here is that an actually blocking AIO call can cause the reactor thread to stall.
- Support for the `FALLOC_FL_WRITE_ZEROES` fallocate flag landed in the v7.3 kernel for XFS. However, this solution suffers from having a span of kernel versions that doesn't support it.
- We're currently getting a kernel patch in that fixes this so that writes over multiple extents can be coalesced into a single bio request if the extents are logically and physically contiguous. However, same as the falloc flag above this too suffers from a span of kernel versions that don't support it.

From our testing failing a non-blocking aio call and having to fallback doesn't introduce noticeable overhead outside of microbenchmarks. Where it does introduce a lot of pain is when other non-aio syscalls on the thread actually block for an extended period of time(1ms+). In these cases tail latency for aio requests increases to the point that e2e latency of a client request also noticeably increases. Hence the solution to this issue that this PR implements is to have a separate syscall thread dedicated to only handling AIO requests. 

We ran this change in the `io_tester` with the following workload:
```yaml
  - name: appender
    shards: all
    type: append
    data_size: 4GB
    shard_info: { parallelism: 16, reqsize: 4kB, rps: 1000 }
    options: { fallocate: false }
  - name: slow_syscall
    shards: all
    type: blocking_sleep
    shard_info: { parallelism: 1, rps: 1, execution_time: 10ms }
```
`slow_syscall` request type was added to simulate blocking syscalls with high latency via a simple sleep. The 10ms is close to what we've seen in the wild. The implementation of this will be in a separate follow-up PR.

`io_tester` ran against a locally mounted XFS volume with the above workload and the following flags:
```
      --smp 2 \
      --duration 12 \
      --reactor-backend=linux-aio \
      --linux-aio-nowait 1
```
The `base` build is tip-of-dev seastar without any changes and the `change` build is with this PR applied. The raw results of the test are:

| build  | rep | p50 | p95  | p99  | p999  | max   | aio_retries |
  |--------|-----|-----|------|------|-------|-------|-------------|
  | base   | 1   | 94  | 397  | 5414 | 15828 | 17127 | 101488      |
  | base   | 2   | 82  | 1162 | 7268 | 15479 | 27675 | 120826      |
  | base   | 3   | 100 | 6611 | 7947 | 18112 | 21659 | 103460      |
  | base   | 4   | 74  | 4858 | 5950 | 17374 | 18761 | 104655      |
  | change | 1   | 89  | 137  | 1003 | 17816 | 19278 | 102577      |
  | change | 2   | 90  | 587  | 5617 | 19838 | 24720 | 102580      |
  | change | 3   | 65  | 111  | 895  | 24830 | 29686 | 99281       |
  | change | 4   | 70  | 94   | 902  | 19486 | 24464 | 103889      |
  
The p50 latency is unchanged and the p99 drops from ~5-8ms in the base to ~1ms with the change. The p999 and max stay roughly the same for both builds as at that point they're dominated by the reactor being saturated rather than the syscall thread.

### DIFF
--- a/src/core/thread_pool.hh
+++ b/src/core/thread_pool.hh
@@ -23,6 +23,10 @@
 
 #include "syscall_work_queue.hh"
 
+#include <atomic>
+#include <memory>
+#include <optional>
+
 namespace seastar {
 
 class file_desc;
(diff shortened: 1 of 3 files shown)


avikivity:
I don't object to this in principle (having a single thread doing everything was a first iteration than never received a second), but the best way to address this is io_uring.

avikivity:
"don't object" is badly worded. It's a good change; but io_uring would be better for you.

(and for us too)

src/core/thread_pool.hh
```diff
@@ -53,37 +57,70 @@ public:
...
     internal::submit_metrics metrics;
-    syscall_work_queue inter_thread_wq;
-    posix_thread _worker_thread;
-    std::atomic<bool> _stopped = { false };
-    std::atomic<bool> _main_thread_idle = { false };
+    // Drains every blocking syscall except AIO submission fallbacks.
+    std::unique_ptr<worker> _main_worker;
+    // Drains blocking AIO read/write (io_submit) fallbacks. Only created when
+    // the reactor uses the linux-aio backend.
+    std::unique_ptr<worker> _aio_worker;
```
avikivity, line 75: I don't understand the complication. Why not create a separate thread_pool object?

thread_pool: manages any number of symmetrical workers (with the restriction that the number must be one)
reactor: manages asymmetrical thread pools, each for a task, with each thread pool containing fully symmetrical threads (currently limited to exactly one thread per pool)

avikivity: I suppose the second pool would be optional, only enabled if the reactor model requires it.

travisdowns: (takes off maintainer hat, puts on Redpanda hat)

We tried both options internally, and I had suggested and was in originally in favor of the way you suggest. However, we ended up preferring this way because:

(v1 referring to this change, ALT referring to your suggestion of a separate `thread_pool` object for this purpose)

 1. Encapsulation/reactor. v1 is entirely inside thread_pool; the reactor only passes a flag. The ALT puts a second pool on the reactor and makes the reactor's syscall poller aware of a backend-specific concept. To avoid that leak, backends would need a hook to register 0..N pollers, moving "almost everything to the backend".
 2. Metrics. v1 needs no special handling: io_threaded_fallbacks keeps its reason label and keeps working because routing is by reason inside the pool. The ALT has to pass the pool explicitly at each submit site and hardcode the reason per pool. It gets messy.
 
... and the most important reason of all, the naming `pool` which suggests that this rightfully should manage > 1 threads to be worthy of the name (as you allude to).

That said, we are not dead set on this approach, it is more important that we are aligned with upstream, so we are putting this here first and if we can get any approach working will downstream that.

If it helps, we can put the ALT approach up for simultaneous review.

travisdowns:
> I don't object to this in principle (having a single thread doing everything was a first iteration than never received a second), but the best way to address this is io_uring.

Yes, but we are not using io_uring by default and neither are you, right?

We are highly interested in it when it's "ready" but currently it is slower. So we should still be open to fixes in aio backend in the meantime, I think (and in practice as older kernels are still common so even when (if) we switch to io_uring, aio will remain important for some subset of deployments).

src/core/thread_pool.hh
ballard26: > Why not create a separate thread_pool object?

I put up an alternate version of this PR that uses a separate thread pool in https://github.com/scylladb/seastar/pull/3656 . Let me know if you prefer that approach and I can close this PR.

avikivity:
> > I don't object to this in principle (having a single thread doing everything was a first iteration than never received a second), but the best way to address this is io_uring.
> 
> Yes, but we are not using io_uring by default and neither are you, right?

Right, but we should, and so should you.

> 
> We are highly interested in it when it's "ready" but currently it is slower. So we should still be open to fixes in aio backend in the meantime, I think (and in practice as older kernels are still common so even when (if) we switch to io_uring, aio will remain important for some subset of deployments).

Of course, and even after everyone switches, if someone sends a patch that improves linux-aio, it should be merged.

travisdowns:
closing as the alternate version merged: https://github.com/scylladb/seastar/pull/3656

PR #3647, opened 2026-08-27
test: futures_test: add test_when_all_exception

bhalevy:
Test that exceptional futures are wrapped properly in the variadic when_all version.


gleb-cloudius (approved):
Test does not hurt, but if this would be broken a lot of code would be broken too. `when_all_succeed` is the one that does not pack exceptions and was added exactly because this one does.

bhalevy:
> Test does not hurt, but if this would be broken a lot of code would be broken too.

It's good to cement the semantics.

> `when_all_succeed` is the one that does not pack exceptions and was added exactly because this one does.

Good point.  I'll add a test case for it too.

PR #3649, opened 2026-08-27
prometheus: make scrape output independent of the serving shard

ScyllaPiotr:
The prometheus scrape handler could produce different output depending on which shard served the connection: `make_filter()` consulted the serving shard's label set to decide whether a query parameter is a label filter, so a filter on a label registered only on other shards was silently ignored.

This went unnoticed because the conntrack load balancer virtually always hands the first connection of an idle server to shard 0 - the shard the metrics are registered on. It surfaced while testing #3526, which changes that tie-breaking: `prometheus_test.py` failed whenever the scrape landed on shard 1.

It can be reproduced on master independently of #3526: run `metrics_tester --smp 2` and issue two concurrent `/metrics?private=1` requests - the connection distributed to shard 1 returns unfiltered output, because the metrics (and the `private` label) are registered on shard 0 only.

The first patch fixes this by treating every query parameter not prefixed with the reserved `__` as a label filter. Note the deliberate behavior change, spelled out in the commit message: a filter on a label name that no metric uses is no longer silently ignored; each series is matched against its (empty) value for that label instead.

The second patch makes `prometheus_test.py` run its whole assertion suite twice: once against an exporter with the default load balancing (scrapes served by shard 0 - the previously exercised behaviour) and once with scrapes pinned, via a new `--serve-on-shard` metrics_tester option, to a shard holding no metrics state. The pinned variant fails its label-filtering tests without the fix.

Note for earlier reviewers: a previous revision carried a third change, setting the metric family configs on all shards in metrics_tester. It was dropped as having no effect - aggregation labels travel with the metric family metadata collected from the shard that registers the metrics, so the serving shard's config is never consulted and the plain shard-0 call was already correct.

### DIFF
--- a/src/core/prometheus.cc
+++ b/src/core/prometheus.cc
@@ -1056,10 +1056,16 @@ class metrics_handler : public httpd::handler_base  {
      */
     std::function<bool(const mi::labels_type&)> make_filter(const http::request& req) {
         std::unordered_map<sstring, std::regex> matcher;
-        auto labels = mi::get_local_impl()->get_labels();
         for (auto&& qp : req.get_query_params()) {
-            if (labels.find(qp.first) != labels.end()) {
-                matcher.emplace(qp.first, std::regex(qp.second.back().c_str()));
+            // Control parameters ("__name__", "__help__", ...) are not label
+            // filters; treat every other query parameter as one. Consulting
+            // the shard-local label registry instead would make the result
+            // depend on which shard serves the scrape, since labels do not
+            // have to be registered uniformly on all shards.
+            const auto& name = qp.first;
+            const bool control_param = name.size() >= 2 && name[0] == '_' && name[1] == '_';
+            if (!control_param) {
+                matcher.emplace(name, std::regex(qp.second.back().c_str()));
             }
         }
         return (matcher.empty()) ? _true_function : [matcher](const mi::labels_type& labels) {
--- a/tests/unit/metrics_tester.cc
+++ b/tests/unit/metrics_tester.cc
@@ -134,6 +134,7 @@ int main(int ac, char** av) {
         ("listen", bpo::value<sstring>()->default_value("0.0.0.0"), "address to start Prometheus server on")
         ("port", bpo::value<uint16_t>()->default_value(9180), "Prometheus port")
         ("conf", bpo::value<sstring>()->default_value("./conf.yaml"), "config with jobs and options")
+        ("serve-on-shard", bpo::value<unsigned>(), "pin serving HTTP connections to this shard; lets tests scrape from a shard other than the one the metrics are registered on")
     ;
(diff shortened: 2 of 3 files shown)


mykaul:
Reminds me of https://github.com/scylladb/seastar/pull/3630 ?

ScyllaPiotr:
> Reminds me of #3630 ?

Both #3630 and #3649 fix the same defect — `make_filter()` consults `get_local_impl()->get_labels()`, a per-shard set, so whether `?somelabel=regex` acts as a filter depends on which shard serves the scrape. Since the two fixes leave different contracts behind for parameters naming labels that no metric uses, I checked what contract we actually document.

`doc/prometheus.md`, section "Filtering by a label value":

▎ _"You can filter by any label using regular expressions. If you use multiple labels in your query, all conditions should be met. A missing label is considered an empty string. The expression should match the entire label value, to match a missing label, you can use `label=` or `label=^$`."_

Two things follow. The documented grammar is "any label", with only the `__`-prefixed parameters (`__name__`, `__help__`) documented as special — the doc never conditions filtering on the label being registered anywhere. And the missing-label rule already covers labels that exist on no metric, with no special case: if no series carries `foo`, every series has the empty string for it, so `?foo=1` must select nothing and `?foo=` must select everything.

Concretely, per request (with `private` registered on some shard, `foo` registered nowhere):

| request | doc says | master | #3630 | #3649 |
|---|---|---|---|---|
| `?private=1` | filtered | depends on serving shard (the bug) | filtered | filtered |
| `?foo=1` | empty output | full output (param ignored) | full output (param ignored) | empty output |
| `?foo=` | full output | full output — but by ignoring the param, not by matching | full output (same accident) | full output, by the documented empty-string match |

So both PRs fix the shard dependence, but against the documentation they differ:

- #3649 implements the documented semantics exactly, everywhere
- #3630 makes behavior shard-independent but globalizes the ignore-unknown fallback, with the new cross-shard machinery whose only purpose is to preserve that undocumented fallback.

**So let's first decide on the contract.** If the documented semantics are the intended ones, #3649's approach is the smaller change that lands us there. If silently ignoring unknown labels is the contract we actually want (it is the more forgiving behavior toward scrapers that append stray parameters), then #3630 is the right fix — but `doc/prometheus.md` should be amended in the same series to say so, because today it promises the other behavior.

Either way, the new test from #3649 (running the whole `prometheus_test.py` suite twice, once served from the metrics shard, and once from a shard with no metrics state) is independent of which fix is chosen and should land either way.

mykaul:
@amnonh - please review.

amnonh:
To my understanding, it does not work. Let's say that I want all the metrics related to a table: I would pass ```"cf=my-table"```, but instead of getting only that table's metrics, I would get everything else that just happens not to have the cf label (which is true for most metrics)

PR #3651, opened 2026-08-28
thread: allow a thread to switch its own scheduling group

nyh:
Code running in a seastar::thread had no way to change which scheduling group it is accounted and scheduled under once the thread was already running: thread_attributes::sched_group only sets the group at construction time, and task::set_scheduling_group() is protected and not reachable from outside.

This matters for the same reason coroutine::switch_to() exists: a long-running fiber may want part of its work billed to a different scheduling group than the rest (e.g. a background/low-priority group for a large chunk of work, switching back to the caller's group afterwards), without spawning a separate thread just to get a different group.

Add thread::switch_to(scheduling_group), the thread counterpart of coroutine::switch_to(): it moves the calling thread to the given group and returns the group it was previously in, so the caller can switch back later. Like coroutine::switch_to(), switching to the group the thread is already in is a no-op and doesn't yield.

The underlying machinery already generalizes to this with almost no new code: thread_context is itself a task, and yield() already does "schedule(this); switch_out();" to reschedule the task and hand control back to the reactor. The only missing piece was calling the already-available (but private to task) set_scheduling_group() before that reschedule, so the thread's task gets requeued under the new group instead of its old one, and exposing that through thread_context and thread as a public switch_to().

Add test_thread_switch_to to tests/unit/thread_test.cc, mirroring coroutines_test.cc's test_switch_to: switch a running thread across two freshly-created scheduling groups and back, checking current_scheduling_group() and the returned "previous group" at each step, including that switching to the current group is a no-op and that an unrelated yield() in between doesn't reset the group.

### DIFF
--- a/include/seastar/core/thread.hh
+++ b/include/seastar/core/thread.hh
@@ -121,6 +121,7 @@ public:
     bool should_yield() const;
     void reschedule();
     void yield();
+    scheduling_group switch_to(scheduling_group new_sg);
     task* waiting_task() noexcept override { return _done.waiting_task(); }
     friend class thread;
     friend void thread_impl::switch_in(thread_context*);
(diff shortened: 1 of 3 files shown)


nyh:
We needed this feature in a ScyllaDB patch (https://github.com/scylladb/scylladb/pull/31304), where we have a long-running worker thread which we realized was doing work for various scheduling groups but running this work in the wrong scheduling group. The way we solved this with existing Seastar features was to have the worker thread spawn a new seastar thread with the right scheduling group and wait for it to complete, but the downside is the overhead of another stack allocation. It's cleaner to just allow the original thread to change its scheduling group, and as this patch shows it's actually easy to add this feature to Seastar.

dkropachev (approved):
Looks great, don't see any problems with it

avikivity (approved):

PR #3656, opened 2026-08-30
[Alt] Use a separate syscall thread for aio fallbacks

ballard26:
Alternative implementation of https://github.com/scylladb/seastar/pull/3645 . This version uses a separate thread pool for the aio fallbacks instead of having the existing thread pool manage multiple threads. The new thread pool is owned by `aio_storage_context` as it's the sole user of the pool.

There is a third alternative implementation that involves the reactor owning the pool object directly. However, this introduces some odd/fragile interfaces between the reactor and `aio_storage_context`. I.e, there are two ways I can think of implementing it this way:
- The `reactor_backend` needs to expose a new virtual method `virtual bool uses_aio_syscall_thread() const` that every backend that uses `aio_storage_context` would need to overwrite to return true so that the reactor knows to construct a aio syscall thread in its constructor.
- The reactor would need to lazily construct the aio syscall thread on first use. Then the `aio_storage_context` would need to make sure that the aio syscall thread is constructed before the pollers were registered. 

Both seem to suffer from fragile contracts that aren't easily enforceable in code.

### DIFF
--- a/src/core/reactor.cc
+++ b/src/core/reactor.cc
@@ -2628,7 +2628,10 @@ void reactor::register_metrics() {
 
     auto io_fallback_counter = [this](const sstring& reason_str, internal::thread_pool_submit_reason r) {
         static auto reason_label = sm::label("reason");
-        return sm::make_counter("io_threaded_fallbacks", std::bind(&thread_pool::count, _thread_pool.get(), r),
+        thread_pool* aio_pool = _backend->aio_thread_pool();
+        thread_pool* pool = (r == internal::thread_pool_submit_reason::aio_fallback && aio_pool)
+                ? aio_pool : _thread_pool.get();
+        return sm::make_counter("io_threaded_fallbacks", std::bind(&thread_pool::count, pool, r),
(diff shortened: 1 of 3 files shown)


src/core/reactor.cc
```diff
@@ -2617,7 +2628,10 @@ void reactor::register_metrics() {
 
     auto io_fallback_counter = [this](const sstring& reason_str, internal::thread_pool_submit_reason r) {
         static auto reason_label = sm::label("reason");
-        return sm::make_counter("io_threaded_fallbacks", std::bind(&thread_pool::count, _thread_pool.get(), r),
+        thread_pool* aio_pool = _backend->aio_thread_pool();
+        thread_pool* pool = (r == internal::thread_pool_submit_reason::aio_fallback && aio_pool)
+                ? aio_pool : _thread_pool.get();
+        return sm::make_counter("io_threaded_fallbacks", std::bind(&thread_pool::count, pool, r),
```
avikivity, line 2634: Please follow up with a new counter for the original syscall thread pool.

PR #3660, opened 2026-09-01
memory: modernize cpu_mem initialization

avikivity:
memory.cc jumps through some hoops to avoid cpu_mem thread-local-storage guards.
But these hoops are now unneeded with constinit/constexpr. Massage the code to avoid
them.

### DIFF
--- a/include/seastar/util/sampler.hh
+++ b/include/seastar/util/sampler.hh
@@ -34,6 +34,9 @@
 
 #pragma once
 
+#include <cstdint>
+#include <limits>
+#include <optional>
 #include <random>
 
 // See also: https://perfetto.dev/docs/design-docs/heapprofd-sampling for more
@@ -41,9 +44,13 @@
 
 class sampler {
 public:
-    sampler() : random_gen(rd_device()) {
-        set_sampling_interval(0);
-    }
+    /// Constructs a sampler with sampling disabled.
+    ///
+    /// The constructor is constexpr (and the destructor trivial) so that a
+    /// thread_local object embedding a sampler can be constant-initialized,
+    /// which is what keeps the seastar allocator's thread_local state free of
+    /// TLS initialization guards.
+    constexpr sampler() = default;
     /// Sets the sampling interval in bytes. Setting it to 0 means to never sample
     void set_sampling_interval(uint64_t sampling_interval) {
         sampling_interval_ = sampling_interval;
@@ -148,19 +155,26 @@ private:
     }
 
     int64_t next_sampling_interval() {
+        if (!random_gen_) {
+            // Seeded on first use: std::random_device and std::mt19937_64 are
+            // not constant-initializable, and sampling is usually never enabled.
+            random_gen_.emplace(std::random_device()());
+        }
         std::exponential_distribution<double> dist(sampling_rate_);
-        int64_t next = static_cast<int64_t>(dist(random_gen));
+        int64_t next = static_cast<int64_t>(dist(*random_gen_));
         // We approximate the geometric distribution using an exponential
         // distribution.
         return next;
     }
 
-    uint64_t sampling_interval_; // Sample every N bytes ; 0 means off
-    double sampling_rate_; // 1 / sampling_interval_ ; used by the exp distribution
+    uint64_t sampling_interval_ = 0; // Sample every N bytes ; 0 means off
+    double sampling_rate_ = 0; // 1 / sampling_interval_ ; used by the exp distribution
     // How many bytes remain to be allocated before we take a sample.
     // Specifically, if this member has value N, a sample will be taken of the allocation
     // that allocates the Nth+1 byte.
-    int64_t interval_to_next_sample_;
-    std::random_device rd_device;
-    std::mt19937_64 random_gen;
+    //
+    // Sampling is off initially, so this is set very large: see
+    // set_sampling_interval().
+    int64_t interval_to_next_sample_ = std::numeric_limits<int64_t>::max();
+    std::optional<std::mt19937_64> random_gen_; // seeded on first use
 };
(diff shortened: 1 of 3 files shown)


avikivity:

include/seastar/util/sampler.hh
```diff
@@ -148,19 +155,26 @@ private:
...
     // Specifically, if this member has value N, a sample will be taken of the allocation
     // that allocates the Nth+1 byte.
-    int64_t interval_to_next_sample_;
-    std::random_device rd_device;
-    std::mt19937_64 random_gen;
+    //
+    // Sampling is off initially, so this is set very large: see
+    // set_sampling_interval().
+    int64_t interval_to_next_sample_ = std::numeric_limits<int64_t>::max();
+    std::optional<std::mt19937_64> random_gen_; // seeded on first use
```
nyh, line 179: When did we start using trailing underscores instead of leading underscores?
Oh, I guess you're just copying the pre-existing style of this header, right? And it was copied verbatim from perfetto?

src/core/memory.cc
```diff
@@ -572,13 +574,37 @@ struct cross_cpu_free_item {
...
 };
 
+/// Uninitialized storage for a single object of type \c T, whose lifetime is
+/// managed explicitly by the user.
+///
+/// The wrapper is trivially default constructible and trivially destructible,
+/// even if \c T is neither. This lets a thread_local aggregate containing one
+/// be constant-initialized, which in turn means the compiler needs no TLS
+/// initialization guard for accesses to it - something we cannot afford on the
+/// allocator fast path. The object must be construct()ed before use, and is
```
nyh, line 584: Cool. In the old days, we needed to use the ugly "__thread" for this.

src/core/memory.cc
```diff
@@ -572,13 +574,37 @@ struct cross_cpu_free_item {
...
+/// never destroyed.
```
nyh, line 585: How is it fine for it to be "never destroyed"?
Is it because we don't this part of Seastar to ever be destructed?

nyh (approved):
Looks good to me.
If I understand correctly, you are assuming that in Seastar, CPUs can never go down after being created? (if not, I think I missed something on why it is fine that all these objects can never be destructed).

nyh:
> Looks good to me. If I understand correctly, you are assuming that in Seastar, CPUs can never go down after being created? (if not, I think I missed something on why it is fine that all these objects can never be destructed).

Claude answered me:

the PR doesn't assume shards never go down. It actually preserves the "shard death" handling exactly, and the "leaked forever" objects are a separate, deliberate and pre-existing pattern that's safe for an unrelated reason. Here's the breakdown, with line references from [memory.cc](https://github.com/scylladb/seastar/pull/3660/files):

Shards do go down mid-process — confirmed by the existing code the PR doesn't touch:

```
void cpu_pages::free_cross_cpu(unsigned cpu_id, void* ptr) {
    if (!live_cpus[cpu_id].load(std::memory_order_relaxed)) {
        // Thread was destroyed; leak object
        // should only happen for boost unit-tests.
        return;
    }
```
`live_cpus[]` exists precisely to detect a dead shard so cross-shard frees targeting it get dropped instead of queued forever. This happens in unit tests, where each test spins up and tears down its own reactor threads within one process.

The PR keeps this mechanism intact, just relocates it. Previously `~cpu_pages()` (a normal thread_local destructor, run at OS thread exit) set `live_cpus[cpu_id] = false`. Now that cpu_pages is constinit/trivially-destructible (needed to drop the TLS guard on the allocator fast path), that one line of real behavior moves into a dedicated thread_local:

```
struct mark_shard_dead_on_thread_exit {
    ~mark_shard_dead_on_thread_exit() {
        if (cpu_mem.is_initialized()) {
            live_cpus[cpu_mem.cpu_id].store(false, ...);
        }
    }
};
static thread_local mark_shard_dead_on_thread_exit mark_shard_dead;
```
touched (and thus lazily registered for destruction) only from cpu_pages::initialize() — so only real shard threads pay for a non-trivial thread_local; alien/helper threads that never become a shard never do. Functionally, this is identical to the old destructor.

What actually becomes "leaked forever" is the shard's own bookkeeping (small_pool free-list trimming, reclaim_hook, reclaimers, alloc_sites), via manual_lifetime. That's safe for two independent reasons:

cpu_ids are handed out from a monotonic counter (cpu_id_gen.fetch_add) and never reused, and each thread_local cpu_mem is only ever initialize()d once (guarded by is_initialized()). So a dead shard's leaked state can never collide with a live one — a new shard thread gets a fresh pthread with fresh (constant-initialized) TLS and a brand-new cpu_id.
Part of this was already true before the PR: the old `small_pool::~small_pool()` was never actually invoked even pre-patch — it sat inside a union whose destructor was deliberately empty (~u() { /* cannot really call destructor... */ }), and alloc_sites was already in a union with `~asu(){} // alloc_sites live forever`. So for those two, the PR just deletes dead code, changing nothing. The only real behavior change is reclaim_hook/reclaimers, which used to be destroyed at thread exit and now leak too — a bounded, one-time leak per shard's lifetime (i.e., basically once per unit test), which is inconsequential.

So the short answer: shards going down is an anticipated, already-handled case (live_cpus + "leak object" cross-free path), and the PR preserves that handling faithfully. The "never destructed" objects are safe to leak because dead cpu_ids are never recycled, not because the PR assumes shards live forever.

include/seastar/util/sampler.hh
avikivity: Yes, Claude preserved the existing style. I don't remember where it came from.

avikivity:
Shards don't really go down mid-process, it's only during shutdown. It's practically impossible to have zero allocations lost (it's enough to have a thread_local std::vector somewhere to hold on to memory, without any way to order the cpu_mem and the variable destructors).

